PR #964, opened 2021-12-23
feat(stream): Add stream library

kriskowal:
Adds a small stream library that provides a foundation for streaming with async iterators. The types are a bit more rigid than AsyncIterator in order to mesh better in TypeScript.

The library includes a reductive makePromiseKit since the AsyncQueue requires a “soft” object and doesn’t require much else that `@agoric/promise-kit` might provide, but nevertheless, I took a cue from attempting to use `@agoric/promise-kit` and implemented the library atop hardened JavaScript.

### DIFF
--- a/packages/stream/index.js
+++ b/packages/stream/index.js
@@ -1,0 +1,147 @@
+/* `makeQueue`, `makeStream`, and `makePipe` are utilities for creating async
+ * iterator "streams". A Stream is compatible with AsyncIterator and Generator
+ * but differ in that every method and argument of both is required.
+ * For example, streams always have `return` and `throw` for closing the write
+ * side.
+ * The `Stream` interface is symmetric, but a stream that sends data and
+ * receives undefined is conventionally a `Writer` whereas a stream that
+ * receives data and sends undefined is conventionally a `Reader`.
+ */
+
+// @ts-check
+/// <reference types="ses"/>
+
+/**
+ * @template T
+ * @typedef {{
+ *   resolve(value?: T | Promise<T>): void,
+ *   reject(error: Error): void,
+ *   promise: Promise<T>
+ * }} PromiseKit
+ */
+
+/**
+ * @template T
+ * @returns {PromiseKit<T>}
+ */
+const makePromiseKit = () => {
+  let resolve;
+  let reject;
+  const promise = new Promise((res, rej) => {
+    resolve = res;
+    reject = rej;
+  });
+  assert(resolve !== undefined);
+  assert(reject !== undefined);
+  return harden({ promise, resolve, reject });
+};
+
+/**
+ * @template T
+ * @typedef {{
+ *   put(value: T | Promise<T>): void,
+ *   get(): Promise<T>
+ * }} AsyncQueue
+ */
+
+/**
+ * @template T
+ * @returns {AsyncQueue<T>}
+ */
+export const makeQueue = () => {
+  let { promise: tailPromise, resolve: tailResolve } = makePromiseKit();
+  return {
+    put(value) {
+      const { resolve, promise } = makePromiseKit();
+      tailResolve(Object.freeze({ value, promise }));
+      tailResolve = resolve;
+    },
+    get() {
+      const promise = tailPromise.then(next => next.value);
+      tailPromise = tailPromise.then(next => next.promise);
+      return harden(promise);
+    },
+  };
+};
+harden(makeQueue);
+
+/**
+ * @template T
+ * @template U
+ * @template V
+ * @typedef {{
+ *   next(value: U): Promise<IteratorResult<T>>,
+ *   return(value: V): Promise<IteratorResult<T>>,
+ *   throw(error: Error): Promise<IteratorResult<T>>,
+ *   [Symbol.asyncIterator](): Stream<T, U, V>
+ * }} Stream
+ */
+
+/**
+ * @template T
+ * @template V
+ * @typedef {Stream<T, undefined, V>} Reader
+ */
+
+/**
+ * @template U
+ * @template V
+ * @typedef {Stream<undefined, U, V>} Writer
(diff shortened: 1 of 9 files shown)


mhofman:
Gonna take a look as I've dealt quite a bit with streams lately on the node side.

But before I dive in, how closely aligned is this with WHATWG streams, and if differs substantially, any reason why we didn't want to align with that API?

kriskowal:
The WHATWG API trades iterator protocol for some helper methods on the prototype that I imagine they expected would court Node.js developers or that they expected writing to `next` to be weird. I find them hard to rationalize, but I’m sure we can bridge up or down.

packages/stream/index.js
```diff
@@ -0,0 +1,147 @@
...
+  let { promise: tailPromise, resolve: tailResolve } = makePromiseKit();
+  return {
+    put(value) {
+      const { resolve, promise } = makePromiseKit();
+      tailResolve(Object.freeze({ value, promise }));
+      tailResolve = resolve;
+    },
+    get() {
+      const promise = tailPromise.then(next => next.value);
+      tailPromise = tailPromise.then(next => next.promise);
```
erights, line 61: Are we trying to avoid depending on eventual-send / HandledPromise? I ask because I'd normally write this as
```suggestion
      const promise = E.get(tailPromise).value;
      tailPromise = E.get(tailPromise).promise;
```

packages/stream/index.js
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * @template T
+ * @template U
+ * @template V
+ * @typedef {{
+ *   next(value: U): Promise<IteratorResult<T>>,
+ *   return(value: V): Promise<IteratorResult<T>>,
+ *   throw(error: Error): Promise<IteratorResult<T>>,
+ *   [Symbol.asyncIterator](): Stream<T, U, V>
+ * }} Stream
```
erights, line 77: I see. It is an iterable iterator. Normally I try to separate the iterable from the iterator, but for streams I think I see why this is less natural. We should discuss.

packages/stream/index.js
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * @template T
+ * @template V
+ * @typedef {Stream<T, undefined, V>} Reader
+ */
+
+/**
+ * @template U
+ * @template V
+ * @typedef {Stream<undefined, U, V>} Writer
```
erights, line 89: Nice

erights (approved):
Looks Brilliant to me! And brilliantly simple!

Looking forward to building on this.

kriskowal:
It seems I can’t append to the conversations in @erights review. So, here inline.

I’ll set it up to use ses-ava. That is an error by omission and I’ll address that up the stack before landing.

I would like to take a dependency on eventual-send when eventual-send makes it into the endo repository, at least for the types.

kriskowal:
In an aside with @mhofman, we discussed why I chose not to use an async generator function to implement the JSON writer in #966. The position is that Node.js pipelines can be expressed as generator function, where the pipeline effectively inverts the writer and reader. This is possible because any abstraction on streams has this reflexive duality. So too with this stream library, it should be possible to express adapters to writers using reader adapters.

It’s clear that you can invert a stream from reader to writer using `makePipe()`, but a naïve first attempt to do that resulted in a difference of parity: the back-pressure signal of the returned writer was one-off from the back-pressure signal of the accepted writer. So, the existing tests failed.

I did eventually come up with a solution that worked, and it would justify adding a method to this library by a name that I haven’t found yet, tentatively `pump`. The insight here is that we have to use the iterator protocol to communicate with a generator, but the back-pressure signal continues to get transported by an async queue. This requires a slightly different entanglement of the async queues for the involved streams.

```js
/**
 * @template Input
 * @template Output
 * @template Return
 * @param {Writer<Input, Return>} acceptedWriter
 * @param {(reader: Reader<Input, Return>) => Reader<Output, Return>} transform
 * @returns {Writer<Output, Return>}
 */
export const pump = (acceptedWriter, transform) => {
  const input = makeQueue();
  const output = makeQueue();
  const acks = makeQueue();
  const inputReader = makeStream(input, output);
  const returnedWriter = makeStream(acks, input);
  const outputReader = transform(inputReader);

  /** @param {Promise<IteratorResult<Output>>} resultP */
  const crank = resultP => {
    resultP.then(
      ({ done, value }) => {
        if (done) {
          acks.put(acceptedWriter.return(value));
        } else {
          acks.put(acceptedWriter.next(value));
          crank(outputReader.next());
        }
      },
      error => {
        acks.put(acceptedWriter.throw(error));
        crank(outputReader.throw(error));
      },
    );
  };
  crank(outputReader.next());

  return returnedWriter;
};
harden(pump);
```

With this primitive, the JSON writer adapter looks like:

```js
// @ts-check
/// <reference types="ses"/>

/* Adapts a Writer<Uint8Array> to a JSON Writer<any>, where reader and writer
 * streams are modeled as hybrid async iterators + generators.
 */

import { pump } from '@endo/stream';

const textEncoder = new TextEncoder();

/**
 * @param {import('@endo/stream').Reader<ReturnType<JSON.parse>, void>} reader
 */
async function* transform(reader) {
  for await (const value of reader) {
    const json = JSON.stringify(value);
    const bytes = textEncoder.encode(json);
    yield bytes;
  }
}

/**
 * @param {import('@endo/stream').Writer<Uint8Array, void>} writer
 */
export const makeJsonWriter = writer => {
  return pump(writer, transform);
};
harden(makeJsonWriter);
```

This passes the same tests, but with the JSON adapters, implementing the async iterable iterator protocol directly results in a clearer program, so I defer this chunk of code to a future change, when it’s better motivated.

mhofman (commented):
I'm still wrapping my head around the composability of this, but here is some feedback on the implementation.

PR #965, opened 2021-12-23
docs(ses): Migrate guide and reference from Agoric docs site (ht dckc)

kriskowal:


packages/ses/docs/reference.md
```diff
@@ -0,0 +1,309 @@
+# Endo and Hardened JavaScript (SES) Programming Reference
+
+This document describes how Hardened JavaScript (formerly SES) affects writing
+Agoric JavaScript code.
+It is very much a "how to do something" document, with little explanation about why and
```
dckc, line 5: the howto and reference genres are essentially disjoint, no?

_I suppose such comments are orthogonal to this PR, which is just a bulk move operation._

dckc:
I appreciate the "ht @dckc" but putting that in the permanent git history produces a bunch of notifications that I can't opt out of, I think. I suggest not doing that. Refer to me as DC or something if you like.

kriskowal:
Ah, I’m taking the `@` off.

dckc (approved):

PR #966, opened 2021-12-24
feat(stream-json): Add JSON stream adapters

kriskowal:
This builds on #964, providing adapters for Uint8Array message streams to JSON streams, capturing the concerns of transcoding and marshaling. These are also hardened.

### DIFF
--- a/packages/stream-json/reader.js
+++ b/packages/stream-json/reader.js
@@ -1,0 +1,26 @@
+// @ts-check
+/// <reference types="ses"/>
+
+/* Adapts a Reader<Uint8Array> to a JSON object Reader<any> where reader and
+ * writer streams are modeled as hybrid async iterators + generators.
+ */
+const textDecoder = new TextDecoder();
+
+/**
+ * @param {import('@endo/stream').Reader<Uint8Array, undefined>} reader
+ */
+async function* makeJsonAsyncIterator(reader) {
+  for await (const bytes of reader) {
+    const text = textDecoder.decode(bytes);
+    yield JSON.parse(text);
+  }
+}
+
+/**
+ * @param {import('@endo/stream').Reader<Uint8Array, undefined>} reader
+ * @returns {import('@endo/stream').Reader<ReturnType<JSON.parse>, undefined>}
+ */
+export const makeJsonReader = reader => {
+  return harden(makeJsonAsyncIterator(reader));
+};
+harden(makeJsonReader);
--- a/packages/stream-json/writer.js
+++ b/packages/stream-json/writer.js
@@ -1,0 +1,27 @@
+// @ts-check
+/// <reference types="ses"/>
+
+/* Adapts a Writer<Uint8Array> to a JSON Writer<any>, where reader and writer
+ * streams are modeled as hybrid async iterators + generators.
+ */
+const textEncoder = new TextEncoder();
+
+/**
+ * @param {import('@endo/stream').Writer<Uint8Array, void>} writer
+ */
+export const makeJsonWriter = writer => {
+  return harden({
+    next(value) {
+      const json = JSON.stringify(value);
+      const bytes = textEncoder.encode(json);
+      return writer.next(bytes);
+    },
+    throw(error) {
+      return writer.throw(error);
+    },
+    return() {
+      return writer.return();
+    },
+  });
+};
+harden(makeJsonWriter);
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "packages/ses-types-test",
     "packages/static-module-record",
     "packages/stream",
+    "packages/stream-json",
     "packages/syrup",
     "packages/test262-runner",
     "packages/test262-runner",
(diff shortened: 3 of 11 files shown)


packages/stream-json/reader.js
```diff
@@ -0,0 +1,26 @@
...
+async function* makeJsonAsyncIterator(reader) {
+  for await (const bytes of reader) {
+    const text = textDecoder.decode(bytes);
+    yield JSON.parse(text);
+  }
+}
+
+/**
+ * @param {import('@endo/stream').Reader<Uint8Array, undefined>} reader
+ * @returns {import('@endo/stream').Reader<ReturnType<JSON.parse>, undefined>}
```
mhofman, line 21: Afaik `JSON.parse` is types to return `any`, right?

It might be more useful(but possibly unsafe) to provide a generic version here.

packages/stream-json/writer.js
```diff
@@ -0,0 +1,27 @@
...
+
+/* Adapts a Writer<Uint8Array> to a JSON Writer<any>, where reader and writer
+ * streams are modeled as hybrid async iterators + generators.
+ */
+const textEncoder = new TextEncoder();
+
+/**
+ * @param {import('@endo/stream').Writer<Uint8Array, void>} writer
+ */
+export const makeJsonWriter = writer => {
```
mhofman, line 12: Any reason not to write this as generator like reader?

kriskowal: Yeah, theoretically there are two kinds of generators: even and odd. An odd generator starts at the top of the function and even generator would advance immediately to the first yield. We only have odd generators, which means that the first value sent to a generator gets discarded. It’s possible to compensate by using a wrapped generator and inducing `next()` immediately, but the type is hard to pin down.

packages/stream-json/reader.js
kriskowal: I was not so sure because the signature is theoretically narrower than that. I figured stating this would give the types room to evolve.

packages/stream-json/reader.js
```diff
@@ -0,0 +1,26 @@
+// @ts-check
+/// <reference types="ses"/>
+
+/* Adapts a Reader<Uint8Array> to a JSON object Reader<any> where reader and
+ * writer streeams are modeled as hybrid async iterators + generators.
```
erights, line 5:
```suggestion
 * writer streams are modeled as hybrid async iterators + generators.
```

packages/stream-json/reader.js
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * @param {import('@endo/stream').Reader<Uint8Array, undefined>} reader
+ */
+async function* makeJsonAsyncIterator(reader) {
+  for await (const bytes of reader) {
+    const text = textDecoder.decode(bytes);
+    yield JSON.parse(text);
+  }
+}
```
erights, line 17: It is a shame we need to leave Jessie. Perhaps we can come up with a design rule for restricted async generators like we have for async functions. For example, if the `yield` is always at the top level of the body of an top level for-await-of, that would cover this case, probably many other related cases, and would still admit a simple implementation strategy short of CPS transforming.

packages/stream-json/reader.js
```diff
@@ -0,0 +1,26 @@
+// @ts-check
```
erights, line 1: Huh. I just noticed. Are we not following the `src` vs `test` directory convention? I think I like that better than having a bunch of source files at the package level, but I have not thought about it.

packages/stream-json/writer.js
erights: The old `function.next` tc39 proposal died because no one was willing to invest to move it forward. That would have allowed generators (and async generators) to read the argument of the first `next` call. Oh well.

I'll use this as an opportunity to lobby, separately, for an updater adaptor which would also rename the API surface. But these PRs should not block on that debate.

erights (approved):
LGTM

packages/stream-json/reader.js
kriskowal: The rule is generally the same, but I am proposing an amendment and I think I have @michaelfig’s blessing:

With Node.js implementing ESM they added `imports` and `exports` features to `package.json` that make it possible to expose more than `main` as the package’s exported API surface, and this is now *enforced*. But, to make a package that straddles the old world and the new, you have to keep the `.js` extension.

It has always been a faux pas to reach like `import "@org/package/src/internal.js`. The `src` directory should indicate that the modules within are not participating in the public API. It’s more proper to `import '@netstring/reader.js'`. It is possible with an `exports` directive to even make this `import '@netstring/reader'`, just not portably.

So I’m proposing that we surface all public API modules at the top-level. It would certainly be consistent with our existing rule to require each of these a thunk for an internal module, e.g., `export * from './src/internal.js'`, but I feel that would be wasteful and fragile.

kriskowal:
@erights What API and behavior do you propose for an updater?

packages/stream-json/reader.js
kriskowal: We certainly could implement this particular case with the iterator interface directly and avoid the need, but there are certainly other kinds of adapters that benefit from async generators.

PR #967, opened 2021-12-24
feat(stream-node): Node.js stream adapters

kriskowal:
This builds on #964 and provides adapters for Node.js Reader and Writer in terms of hardened JS streams (based on the async iterator protocol.)

### DIFF
--- a/packages/stream-node/reader.js
+++ b/packages/stream-node/reader.js
@@ -1,0 +1,66 @@
+// @ts-check
+/// <reference types="ses"/>
+
+// This module provided for sake of fewer head scratches.
+// Node.js readable streams satisfy the signature of an async iterable iterator.
+// They however iterate Node.js Buffer values and are not hardened, so this
+// implementation compensates for both.
+
+const { details: X, quote: q } = assert;
+
+/**
+ * @param {import('stream').Readable} input the source Node.js reader
+ * @returns {import('@endo/stream').Reader<Uint8Array>}
+ */
+export const makeNodeReader = input => {
+  assert(
+    !input.readableObjectMode,
+    X`Cannot convert Node.js object mode Reader to AsyncIterator<Uint8Array>`,
+  );
+  assert(
+    input.readableEncoding === null,
+    X`Cannot convert Node.js Reader with readableEncoding ${q(
+      input.readableEncoding,
+    )} to a AsyncIterator<Uint8Array>`,
+  );
+
+  const finalIteration = new Promise((resolve, reject) => {
+    input.on('error', reject);
+    input.on('close', () => {
+      resolve({ done: true, value: undefined });
+    });
+  });
+
+  const iterator = input[Symbol.asyncIterator]();
+  /** @type {import('@endo/stream').Reader<Uint8Array>} */
+  const reader = harden({
+    async next() {
+      const result = await iterator.next();
+      if (result.done) {
+        return result;
+      }
+      assert(typeof result.value !== 'string');
+      const { buffer, byteOffset, length } = result.value;
+      return {
+        done: false,
+        value: new Uint8Array(buffer, byteOffset, length),
+      };
+    },
+    async return() {
+      input.destroy();
+      return finalIteration;
+    },
+    /**
+     * @param {Error} error
+     */
+    async throw(error) {
+      input.destroy(error);
+      return finalIteration;
+    },
+    [Symbol.asyncIterator]() {
+      return reader;
+    },
+  });
+  return reader;
+};
+harden(makeNodeReader);
--- a/packages/stream-node/test/test-stream-node.js
+++ b/packages/stream-node/test/test-stream-node.js
@@ -1,0 +1,246 @@
+// @ts-check
+/* global setTimeout */
+
+import 'ses';
+import './lockdown.js';
+
+import rawTest from 'ava';
+import { wrapTest } from '@endo/ses-ava';
+
+import { fileURLToPath } from 'url';
+import { fork } from 'child_process';
+import { makeNodeReader, makeNodeWriter } from '../index.js';
+
+const test = wrapTest(rawTest);
+
+const catPath = fileURLToPath(new URL('cat.js', import.meta.url).toString());
+
+test('stream to and from Node.js reader/writer', async (/** @type {import('ava').ExecutionContext} */ t) => {
+  const scratch = new Uint8Array(1024 * 1024);
+  for (let i = 0; i < scratch.byteLength; i += 1) {
+    scratch[i] = i % 256;
+  }
+
+  const child = fork(catPath, {
+    stdio: ['pipe', 'pipe', 'inherit', 'ipc'],
+  });
+
+  // Call Symbol.asyncIterator in a barefaced attempt to get test coverage
+  // through a superfluous method call.
+  const writer = makeNodeWriter(child.stdin)[Symbol.asyncIterator]();
+  const reader = makeNodeReader(child.stdout)[Symbol.asyncIterator]();
+
+  // We send an IPC message to begin forwarding.
+  let flowing = false;
+
+  const exited = new Promise((resolve, reject) => {
+    child.on('error', reject);
+    child.on('exit', resolve);
+  });
+
+  const ipcReady = new Promise(resolve => {
+    child.on('message', resolve);
+  });
+
+  const makeProducer = async () => {
+    let chunkLength = 1;
+    for (let i = 0; i < scratch.byteLength; ) {
+      const j = i + chunkLength;
+      t.log('->', i, j);
+
+      const nextP = writer.next(scratch.subarray(i, j));
+      if (
+        !flowing &&
+        child.stdin.writableLength >= child.stdin.writableHighWaterMark
+      ) {
+        // eslint-disable-next-line no-await-in-loop
+        await ipcReady;
+        t.log('---');
+        flowing = true;
+        // Child will buffer stdin until called for, allowing us to verify and
+        // cover write stream pressure.
+        child.send({});
+      }
+      // eslint-disable-next-line no-await-in-loop
+      await nextP;
+
+      i = j;
+      chunkLength *= 2;
+    }
+    await writer.return();
+  };
+
+  const makeConsumer = async () => {
+    let i = 0;
+    for await (const chunk of reader) {
+      t.assert(flowing);
+      const j = i + chunk.byteLength;
+      t.log('<-', i, j);
+      t.deepEqual(chunk, scratch.subarray(i, j));
+      i = j;
+    }
+    t.is(i, scratch.byteLength);
+  };
+
+  await Promise.all([makeProducer(), makeConsumer(), exited]);
+});
+
+test('stream read return', async (/** @type {import('ava').ExecutionContext} */ t) => {
+  const scratch = new Uint8Array(1024 * 128);
+  for (let i = 0; i < scratch.byteLength; i += 1) {
+    scratch[i] = i % 256;
+  }
+
+  const child = fork(catPath, {
+    // Note: ignoring stderr. Node.js barfs as expected when we close its
+    // output stream in this test.
+    stdio: ['pipe', 'pipe', 'ignore', 'ipc'],
+  });
+  const reader = makeNodeReader(child.stdout);
+  const writer = makeNodeWriter(child.stdin);
+
+  const exited = new Promise((resolve, reject) => {
+    child.on('error', reject);
+    child.on('exit', resolve);
+  });
+
+  // Begin forwarding whenever ready:
+  child.on('message', () => {
+    child.send({});
+  });
+
+  const makeProducer = async () => {
+    await writer.next(scratch);
+    await writer.return();
+  };
+
+  const makeConsumer = async () => {
+    // Consumer aborts the stream.
+    await reader.return();
+  };
+
+  await Promise.allSettled([makeProducer(), makeConsumer(), exited]);
+  t.pass();
+});
+
+test('stream read throw', async (/** @type {import('ava').ExecutionContext} */ t) => {
+  const scratch = new Uint8Array(1024 * 128);
+  for (let i = 0; i < scratch.byteLength; i += 1) {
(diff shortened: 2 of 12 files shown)


erights:
> This builds on #964 and provides adapters for Node.js Reader and Writer in terms of hardened JS streams (based on the async iterator protocol.)

PR comment stale. It now builds on #966

packages/stream-node/reader.js
```diff
@@ -0,0 +1,39 @@
...
+// Node.js readable streams satisfy the signature of an async iterable iterator.
+// They however iterate Node.js Buffer values and are not hardened, so this
+// implementation compensates for both.
+
+/**
+ * @param {NodeJS.ReadableStream} input the source Node.js reader
+ * @returns {import('@endo/stream').Reader<Uint8Array, void>}
+ */
+export const makeNodeReader = input => {
+  const iterator = input[Symbol.asyncIterator]();
```
erights, line 14: Is input already an iterable-iterator? If so, why don't we just use it as `iterator` rather than calling its @@asyncIterator() method?

I note that our @@asyncIterator method does not call the underlying. I'm confused. Let's discuss.

packages/stream-node/writer.js
```diff
@@ -0,0 +1,36 @@
...
+ * @returns {import('@endo/stream').Writer<Uint8Array, void>}
+ */
+export const makeNodeWriter = writer => {
+  return harden({
+    async next(value) {
+      return new Promise(resolve => {
+        if (!writer.write(value)) {
+          writer.once('drain', resolve);
+        } else {
+          resolve();
```
erights, line 24: I don't understand the `else` case.

erights (commented):
Review so far

packages/stream-node/reader.js
mhofman: Node streams are async-iterable, but they're not async-iterators.

packages/stream-node/reader.js
```diff
@@ -0,0 +1,39 @@
+// @ts-check
+/// <reference types="ses"/>
+
+// This module provided for sake of fewer head scratches.
+// Node.js readable streams satisfy the signature of an async iterable iterator.
+// They however iterate Node.js Buffer values and are not hardened, so this
```
mhofman, line 6: Node streams iterate anything, their values may not be `Buffer` if they are in object mode. Also newer versions of Node actually have `Buffer` as a subclass of `UInt8Array`.

packages/stream-node/writer.js
mhofman: Node is a little peculiar when it comes to writable streams. The model is that you call `write`, and if it returns `false`, you should no longer call write until the `'drain'` event has been dispatched. By resolving right away it informs the writer that further writing calls can happen, preserving the node stream semantics.

mhofman (changes requested):
I thing coercing to `UInt8Array` is wrong, the reader should be a simple pass-through to the node stream iterator.

kriskowal:
I was inspired to coerce to Uint8Array because `t.deepEqual` didn’t think that the Uint8Array I put in was the same as the Buffer that came out. I can easily imagine using these adapters in situations that are equally sensitive, perhaps providing information about whether the program is running under Node.js or on the web. My preference is to present a stable type. There’s also the matter of needing to harden Buffer.prototype in addition to the Uint8Array prototype when this system is used in a hardened context.

packages/stream-node/reader.js
kriskowal: As Mathieu answered, Node.js streams are iterable but not iterators, so this is a necessary step.

mhofman:
I still maintain that a node stream is not automatically a stream of buffers. Maybe the adapter should check the `objectMode` and coerce only in those cases?

packages/stream-node/writer.js
kriskowal: Also, as Mathieu answers, writer.write tells the caller whether the underlying buffer is above a high water mark. It’s not a ring buffer, so it doesn’t have to return a partial success, but it does advise you to back off until drain.

I’m translating that to a system where `next` returns a promise that resolves when you can safely send more data. If `write` returns true, that’s immediate.

An alternate framing would return a promise for the drain even only if write returns false, but I vaguely suspect that always returning a promise is better.

```js
if (!writer.write(value)) {
  return new Promise(resolve => writer.once('drain', resolve));
}
```

kriskowal:
I think it would be reasonable to assert that the stream is not in objectMode.

mhofman:
I would still like an adapter that doesn't coerce, but arguably that's not specific to node streams and would simply be an adapter from async-iterable to a hardened async-iterable-iterator.

packages/stream-node/reader.js
```diff
@@ -0,0 +1,52 @@
...
+  const reader = harden({
+    async next() {
+      const result = await iterator.next();
+      if (result.done) {
+        return result;
+      }
+      assert(typeof result.value !== 'string');
+      return {
+        done: false,
+        value: new Uint8Array(result.value.buffer),
```
mhofman, line 36: Unfortunately that's not always right: https://nodejs.org/api/buffer.html#bufbyteoffset

I'm not sure what's the best way to do this while avoiding a copy.

packages/stream-node/writer.js
mhofman: That wouldn't help since `next` has to be an async function that always returns a Promise.

packages/stream-node/reader.js
kriskowal: Addressed. I believe in this case, it might be just working out because the underlying stream has to make a copy from a ring buffer so it’s always aligned, but in principle it could be offset and it’s easy to accommodate the possibility.

mhofman (commented):
I gave this more thought, and I'm a little worried we're being too lax with the drain mechanism, in particular not awaiting drainage when closing the node writable stream.

packages/stream-node/reader.js
```diff
@@ -0,0 +1,53 @@
...
+    async next() {
+      const result = await iterator.next();
+      if (result.done) {
+        return result;
+      }
+      assert(typeof result.value !== 'string');
+      const { buffer, byteOffset, byteLength } = result.value;
+      return {
+        done: false,
+        value: new Uint8Array(buffer, byteOffset, byteLength),
```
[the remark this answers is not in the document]
kriskowal: Addressed.

packages/stream-node/test/test-stream-node.js
```diff
@@ -0,0 +1,55 @@
...
+    let i = 0;
+    for await (const chunk of reader) {
+      const j = i + chunk.byteLength;
+      t.deepEqual(chunk, scratch.subarray(i, j));
+      i = j;
+    }
+    t.is(i, scratch.byteLength);
+  };
+
+  await Promise.all([makeProducer(), makeConsumer()]);
```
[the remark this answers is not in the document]
kriskowal: Addressed by adding an exited promise.

packages/stream-node/test/test-stream-node.js
```diff
@@ -0,0 +1,55 @@
...
+import { fileURLToPath } from 'url';
+import { fork } from 'child_process';
+import { makeNodeReader, makeNodeWriter } from '../index.js';
+
+const test = wrapTest(rawTest);
+
+const catPath = fileURLToPath(new URL('cat.js', import.meta.url).toString());
+
+test('stream to and from Node.js reader/writer', async t => {
+  const scratch = new Uint8Array(1024 * 16);
```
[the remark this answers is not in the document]
kriskowal: Addressed by instrumenting cat such that it buffers until sent an IPC message. Drain now has coverage.

packages/stream-node/writer.js
```diff
@@ -0,0 +1,43 @@
...
+        }
+      });
+    },
+    async return() {
+      writer.end();
+    },
+    async throw() {
+      writer.end();
+    },
```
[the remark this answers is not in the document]
kriskowal: Addressed these at least to the extent of my knowledge of how these work. I introduced promises for various events and set these methods up to race to their conclusion.

kriskowal:
This now has full test coverage.

packages/stream-node/writer.js
```diff
@@ -0,0 +1,72 @@
...
+          if (!writer.write(value)) {
+            writer.once('drain', resolve);
+          } else {
+            resolve(undefined);
+          }
+        }),
+        finished,
+        errored,
+        closed,
+      ]);
```
mhofman, line 56: I like the `race` approach. Wondering however about the order of these. In particular, should the `resolve` from `write` have precedence over the others?

packages/stream-node/writer.js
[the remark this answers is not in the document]
mhofman: Let me do a little digging on how stream errors are supposed to be reported into a writtable node stream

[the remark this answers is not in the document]
mhofman: Alright, it seems that both the readable async iterator [^1] and the pipeline logic [^2] in node simply calls `.destroy(err)` (falling back to `.close()`[^3] if not a function) when an error occurs, and assume that the destruction is synchronous.

I think we could change `throw(err)` to similarly call `.destroy(err)` with fallback, while keeping the racing logic.

[^1]: https://github.com/nodejs/node/blob/v14.17.0/lib/internal/streams/readable.js#L1108-L1167
[^2]: https://github.com/nodejs/node/blob/v14.17.0/lib/internal/streams/pipeline.js#L64-L69
[^3]: https://github.com/nodejs/node/blob/v14.17.0/lib/internal/streams/destroy.js#L181-L182

packages/stream-node/writer.js
kriskowal: This could probably be addressed by having `finish`, `error`, and `close` race to resolve a single promise, but I’m less sure whether that would be correct.

mhofman: I meant to simply place `finished, errored, closed` before `new Promise` since `race` will respect order

mhofman: Also not sure of the relative order between those, in particular, it feels like `errored` should come first.

packages/stream-node/writer.js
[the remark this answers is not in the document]
kriskowal: Addressed in follow-up, and adjusted the corresponding test. Separately reviewable commit appended. These I will squash before landing.

packages/stream-node/test/test-stream-node.js
```diff
@@ -0,0 +1,238 @@
...
+  });
+
+  // Call Symbol.asyncIterator in a barefaced attempt to get test coverage
+  // through a superfluous method call.
+  const writer = makeNodeWriter(child.stdin)[Symbol.asyncIterator]();
+  const reader = makeNodeReader(child.stdout)[Symbol.asyncIterator]();
+
+  // The child process will not forward stdin to stdout until unplugged.
+  // We later send an IPC message to unplug.
+  let plugged = true;
```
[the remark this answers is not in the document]
kriskowal: Done.

packages/stream-node/writer.js
kriskowal: I made the relative order moot by consolidating these three cases into a promise that gets resolved once by whichever event gets emitted first.

packages/stream-node/writer.js
```diff
@@ -22,17 +22,11 @@ export const makeNodeWriter = writer => {
     X`Cannot convert Node.js object mode Writer to AsyncIterator<undefined, Uint8Array>`,
   );
 
-  const finished = new Promise(resolve => {
+  const finalIteration = new Promise((resolve, reject) => {
```
mhofman, line 25: Unfortunately I think that means we lose the rejection in case the stream finishes because of an error.

mhofman (commented):
I'm not sure the switch to `finish` event only is right, as that means we no longer forward errors.

Thanks for fixing the determinism of the drain test.

packages/stream-node/reader.js
```diff
@@ -23,26 +23,42 @@ export const makeNodeReader = input => {
...
     },
-    /** @param {Error} error */
+    /**
+     * @param {Error} error
+     */
     async throw(error) {
-      return iterator.throw(error);
+      input.destroy(error);
+      return Promise.race([closed, errored]);
     },
```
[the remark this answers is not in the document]
kriskowal: I took your suggestion (from out-of-band) to try to reimplement Node.js reader in terms of mapReader, but the resulting reader would hang on the first test. I suspect that Node.js’s reader’s async generator isn’t handling `return` and `throw` the same way as this implementation.

packages/stream-node/writer.js
kriskowal: We concluded out-of-band that a stream should terminate with either finish or error exclusively and that close may follow after that, but is not guaranteed. We agreed to keep this block of code mostly the same (I’ve appended a commit to document the choice).

packages/stream-node/test/test-stream-node.js
```diff
@@ -39,14 +39,28 @@ test('stream to and from Node.js reader/writer', async (/** @type {import('ava')
...
+      const nextP = writer.next(scratch.subarray(i, j));
+      if (child.stdin.writableLength >= child.stdin.writableHighWaterMark) {
+        // eslint-disable-next-line no-await-in-loop
+        await ipcReady;
+        t.log('---');
+        plugged = false;
+        // Child will buffer stdin until called for, allowing us to verify and
+        // cover write stream pressure.
+        child.send({});
+      }
```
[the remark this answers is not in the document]
kriskowal: I’ve appended a commit that simply adds `!flowing &&` to the guard here.

packages/stream-node/writer.js
mhofman: Yeah apologies, I initially missed the `writer.on('error', reject);` in my previous review.

packages/stream-node/test/test-stream-node.js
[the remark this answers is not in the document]
mhofman: Well that's simpler than my silly suggestion.

mhofman (approved):
LGTM. Thanks for being patient with me on this back and forth!

PR #968, opened 2021-12-24
refactor(netstring)!: Eject stream library

kriskowal:
The netstring library was a proving ground for `@endo/stream`, validating the API and notion of async iterators as a streaming primitive, giving it exercise as a component of `@agoric/xsnap`. That experiment is successfully concluded and we’re making more stream libraries for the Endo browser.

This change ejects the original embedded stream library and makes use of the new `@endo/stream` package. This is however a breaking change because it introduces a dependency on Hardened JavaScript. I’ve taken the liberty to also align the API with the `make*` pattern used pervasively.


dckc (approved):
lgtm

dckc:
I wish you the best of luck in your battle with the ci bots. I'm unsubscribing. tag me again if this merits my attention.

PR #969, opened 2021-12-25
feat(lp32): Implement Chrome native message host streams

kriskowal:
This adds a stream library for communicating over Chrome’s native message host protocol.

https://developer.chrome.com/docs/apps/nativeMessaging/

### DIFF
--- a/packages/lp32/reader.js
+++ b/packages/lp32/reader.js
@@ -1,0 +1,87 @@
+// @ts-check
+/// <reference types="ses"/>
+
+// We use a DataView to give users choice over endianness.
+// But DataView does not default to host-byte-order like other typed arrays.
+
+import { hostIsLittleEndian } from './src/host-endian.js';
+
+const { details: X, quote: q } = assert;
+
+/**
+ * @param {Iterable<Uint8Array> | AsyncIterable<Uint8Array>} reader
+ * @param {Object} [opts]
+ * @param {string=} [opts.name]
+ * @param {number} [opts.maxMessageLength] - defaults to 1MB
+ * @param {boolean=} [opts.littleEndian]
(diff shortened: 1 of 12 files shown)


packages/lp32/reader.js
```diff
@@ -0,0 +1,88 @@
...
+    while (!drained && length >= 4) {
+      const messageLength = data.getUint32(0, littleEndian);
+      assert(
+        messageLength <= maxMessageLength,
+        X`Messages on ${q(
+          name,
+        )} must not exceed ${maxMessageLength} bytes in length`,
+      );
+      const envelopeLength = 4 + messageLength;
+      if (envelopeLength <= length) {
```
michaelfig, line 58: I'd prefer the following, but if it doesn't seem to make things clearer for you, feel free to disregard.

```suggestion
      drained = envelopeLength > length;
      if (!drained) {
```

packages/lp32/writer.js
```diff
@@ -0,0 +1,51 @@
...
+    },
+    /** @param {Error} error */
+    async throw(error) {
+      return output.throw(error);
+    },
+    [Symbol.asyncIterator]() {
+      return this;
+    },
+  });
+};
```
michaelfig, line 50: There's rarely a reason to use `this` (only time I've found is when providing a callback to a legacy API that sets `this` to something unobtainable by other means), and I'd prefer staying within Jessie.

Maybe instead:
```suggestion
  const writer = harden({
// the rest ...
    [Symbol.asyncIterator]() {
      return writer;
    },
  });
  return writer;
};
```

michaelfig (approved):
Looking good!  Take suggestions with a grain of salt.

PR #560, opened 2021-01-16
chore(base64): Migrate base64 from Agoric SDK

kriskowal:
This change subsumes the `@agoric/base64` package from the Agoric SDK into the SES shim repository as `@endo/base64`.

This is a move to practice migrating code while preserving its history using `git subtree`. The construction of this branch consisted of:

```js
agoric-sdk$ git subtree split --prefix packages/base64 HEAD --branch split-base64
agoric-sdk$ cd ../ses-shim
ses-shim$ git remote add agoric-sdk ../agoric-sdk
ses-shim$ git fetch agoric-sdk --no-tags
ses-shim$ git subtree join --prefix packages/base64 agoric-sdk/split-base64
```

Then, updates for the repository package.json workspaces, yarn.lock, and renaming the package to `@endo/base64`.

The significance of the rename is that it will allow us to establish legitimacy of our ownership of the `endo` organization on `npm` before it’s appropriated legitimately by their name-squatting policy.

PR #970, opened 2021-12-28
refactor(promise-kit): Migrate promise-kit from agoric-sdk

kriskowal:
This change uses git subtree (as directed in #560) to migrate `@agoric/promise-kit` to this repository then adapts it to the local idioms. The pull request consists of the history from agoric-sdk, a subtree merge commit, a touch-up refactor that touches the package name, descriptor, NEWS, and adds the missing LICENSE, then a commit to move `src/promiseKit.js` to `index.js`.


packages/promise-kit/index.js
```diff
@@ -0,0 +1,89 @@
...
+ * @property {Promise<T>} promise
+ */
+
+/**
+ * PromiseRecord is deprecated in favor of PromiseKit.
+ *
+ * @template T
+ * @typedef {PromiseKit<T>} PromiseRecord
+ */
+
```
erights, line 24: Seems like an opportune time to drop this completely.

packages/promise-kit/index.js
```diff
@@ -0,0 +1,89 @@
...
+/// <reference types="ses"/>
+
+/** @type {PromiseConstructor} */
+const BestPipelinablePromise = globalThis.HandledPromise || Promise;
+
+/**
+ * @template T
+ * @typedef {Object} PromiseKit A reified Promise
+ * @property {(value: ERef<T>) => void} resolve
+ * @property {(reason: any) => void} reject
```
erights, line 14: Since TypeScript is purposely unsound, in order to create a better user experience for the typical case, I wonder if we should 
```suggestion
 * @property {(reason: Error) => void} reject
```
to strongly suggest that rejecting with a non-error is probably a mistake.

packages/promise-kit/index.js
```diff
@@ -0,0 +1,89 @@
...
+    if (unsafe) {
+      const originalDomain = p.domain;
+      Object.defineProperty(p, 'domain', {
+        get() {
+          return originalDomain;
+        },
+      });
+    } else {
+      delete p.domain;
+    }
```
erights, line 74: Since
   * This never really provided any protection, because promises could always be created by other means
   * We now have real protection against domain
 
I think we should remove this wart completely.

```suggestion
```

packages/promise-kit/index.js
```diff
@@ -0,0 +1,89 @@
...
+ * @template T
+ * @returns {PromiseKit<T>}
+ */
+export function makePromiseKit() {
+  /** @type {(value: ERef<T>) => void} */
+  let res = NOOP_INITIALIZER;
+  /** @type {(reason: any) => void} */
+  let rej = NOOP_INITIALIZER;
+
+  /** @type {Promise<any> & {domain?: unknown}} */
```
erights, line 53:
```suggestion
  /** @type {Promise<any>} */
```
or perhaps this simpler type doesn't need to be present at all.

packages/promise-kit/index.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+  return harden({ promise: p, resolve: res, reject: rej });
+}
+harden(makePromiseKit);
+
+/**
+ * Determine if the argument is a Promise.
+ *
+ * @param {any} maybePromise The value to examine
+ * @returns {maybePromise is Promise} Whether it is a promise
```
erights, line 84: Gotta love these type guards!

erights (approved):
LGTM

Wow, 90% boilerplate. The actual content was 89/892.

packages/promise-kit/index.js
mhofman: Typescript seem pretty opiniated about typing caught things as `unknown`, so if any changes were made, I'd prefer `unknown` to `any`.

packages/promise-kit/index.js
mhofman: I believe the ` & {domain?: unknown}` wart was to allow accessing that property further down.

packages/promise-kit/package.json
```diff
@@ -1,61 +1,82 @@
...
+    "lint:js": "eslint .",
+    "lint:types": "tsc --build jsconfig.json",
+    "test": "ava",
+    "test:xs": "exit 0"
   },
-  "repository": {
-    "type": "git",
-    "url": "git+https://github.com/Agoric/agoric-sdk.git"
+  "dependencies": {
+    "@endo/ses-ava": "^0.2.11",
```
mhofman, line 38: why is `ses-ava` not a devDependency?

PR #971, opened 2021-12-29
refactor(stream): Give types unto TypeScript

kriskowal:
This change ejects certain type definitions from the stream library as they are not fully expressible as JSDoc. Particularly, this allows the reader and writer generic types to have default templates.

### DIFF
--- a/packages/stream-types-test/tsconfig.json
+++ b/packages/stream-types-test/tsconfig.json
@@ -1,0 +1,14 @@
+{
+  "compilerOptions": {
+    "noEmit": true,
+    "target": "esnext",
+    "module": "esnext",
+    "moduleResolution": "node",
+    "strict": true,
+    "strictNullChecks": true,
+    "noImplicitAny": true,
+    "downlevelIteration": true,
+    "strictNullChecks": true
+  },
+  "include": ["**/*.ts", "**/*.js"]
+}
--- a/packages/stream-types-test/validation.ts
+++ b/packages/stream-types-test/validation.ts
@@ -1,0 +1,60 @@
+/* eslint-disable no-unused-vars, no-underscore-dangle, no-empty */
+/// <reference types="ses"/>
+
+import { makeQueue, makeStream, makePipe } from '@endo/stream';
+// eslint-disable-next-line
+import type { Stream, Reader, Writer } from '@endo/stream';
+
+async () => {
+  const q = makeQueue<number>();
+  q.put(1);
+  q.put(Promise.resolve(1));
+  // @ts-expect-error
+  q.put('NaN'); // it's a string!
+  const a : number = await q.get();
+}
+
+async () => {
+  const q = makeQueue<IteratorResult<number>>();
+  const r = makeQueue<IteratorResult<undefined>>();
+  const s = makeStream(q, r);
+  for await (const value of s) {
+    const cast : number = value;
+  }
+
+  const t = makeStream(r, q);
+  // @ts-expect-error
+  for await (const value of t) {}
+};
+
+async () => {
+  const [reader, writer] = makePipe<number>();
+  for await (const _ of reader) {}
+  await writer.next(1);
+  // @ts-expect-error
+  await writer.next('NaN');
+  // @ts-expect-error
+  for await (const _ of writer) {}
+};
+
+async () => {
+  const [reader, writer] = makePipe<undefined, number>();
+  const r : IteratorResult<undefined> = await reader.next(1);
+  // @ts-expect-error
+  const s : IteratorResult<undefined> = await reader.next('NaN');
+  // @ts-expect-error
+  for await (const _ of reader) {}
+  for await (const value of writer) {
+    const cast : number = value;
+  }
+};
+
+async () => {
+  const [reader, writer] = makePipe<number, string, boolean, Array<number>>();
+  const a: IteratorResult<number, boolean> = await reader.next('A');
+  const b: IteratorResult<number, boolean> = await reader.return([1, 2, 3]);
+  const c: IteratorResult<number, boolean> = await reader.throw(new Error('Abort'));
+  const d: IteratorResult<string, Array<number>> = await writer.next(1);
+  const e: IteratorResult<string, Array<number>> = await writer.return(true);
+  const f: IteratorResult<string, Array<number>> = await writer.throw(new Error('Abort'));
+};
--- a/packages/stream/index.d.ts
+++ b/packages/stream/index.d.ts
@@ -1,2 +1,24 @@
-export { makeQueue, makeStream, makePipe } from './index.js';
-export type { Stream, Reader, Writer, AsyncQueue } from './index.js';
+export * from './types.js';
+import { AsyncQueue, Stream, Reader, Writer } from './types.js';
+
+export function makeQueue<TValue>(): AsyncQueue<TValue>;
+
+export function makeStream<
+  TRead,
+  TWrite = unknown,
+  TReadReturn = unknown,
+  TWriteReturn = unknown,
+>(
+  acks: AsyncQueue<IteratorResult<TRead, TReadReturn>>,
+  data: AsyncQueue<IteratorResult<TWrite, TWriteReturn>>,
+): Stream<TRead, TWrite, TReadReturn, TWriteReturn>;
+
+export function makePipe<
+  TRead,
+  TWrite = unknown,
+  TReadReturn = unknown,
+  TWriteReturn = unknown,
+>(): [
+  Stream<TRead, TWrite, TReadReturn, TWriteReturn>,
+  Stream<TWrite, TRead, TWriteReturn, TReadReturn>,
+];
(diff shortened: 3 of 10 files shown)


kriskowal:
cc @erights. I’ve tossed the review to @mhofman, but you may be interested in the shift.

packages/stream/types.d.ts
```diff
@@ -0,0 +1,23 @@
...
+  next(value: TWrite): Promise<IteratorResult<TRead, TReadReturn>>,
+  return(value: TWriteReturn): Promise<IteratorResult<TRead, TReadReturn>>,
+  throw(error: Error): Promise<IteratorResult<TRead, TReadReturn>>,
+  [Symbol.asyncIterator](): Stream<TRead, TWrite, TReadReturn, TWriteReturn>,
+}
+
+export type Reader<TRead> = Stream<TRead, undefined, undefined, undefined>;
+export type Writer<TWrite> = Stream<undefined, TWrite, undefined, undefined>;
+
+export type makePipeFn<T, U, TReturn, UReturn> = () => [
```
erights, line 20: How about
```suggestion
export type MakePipeFn<T, U, TReturn, UReturn> = () => [
```
?

packages/stream/index.js
```diff
@@ -20,6 +20,11 @@
  * }} PromiseKit
  */
 
+// TypeScript ReadOnly semantics are not sufficiently expressive to distinguish
+// a value one promises not to alter from a value one must not alter,
+// making it useless.
```
mhofman, line 25: I'm not sure I understand or agree with this statement. What is the harm in marking the result as `readonly`? Besides the fact that a lot of code is currently typed to only accept a non-readonly object.

packages/stream/index.js
```diff
@@ -131,11 +105,11 @@ export const makeStream = (acks, data) => {
 harden(makeStream);
 
 /**
- * @template T
- * @template U
+ * @template TRead
+ * @template TWrite
  * @template TReturn
  * @template UReturn
```
mhofman, line 111:
```suggestion
 * @template TReadReturn
 * @template TWriteReturn
```

packages/stream/index.js
```diff
@@ -131,11 +105,11 @@ export const makeStream = (acks, data) => {
...
- * @returns {[Stream<T, U, TReturn>, Stream<U, T, UReturn>]}
+ * @type {import('./types.js').makePipeFn<TRead, TWrite, TReturn, UReturn>}
```
mhofman, line 112: I'm thinking that we should instead type the `data` and `acks` queues, and let type inference work through. Combined with a `types.test.ts` file for types, it should be enough to ensure the return type is correct.

mhofman: Alternatively it might be possible to `declare function makePipe` in a `.d.ts` file and do a `@returns {ReturnType<typeof import('./index.js').makePipe<TRead, TWrite, TReturn, UReturn>>}`

packages/stream/index.d.ts
```diff
@@ -1,2 +1,2 @@
 export { makeQueue, makeStream, makePipe } from './index.js';
```
mhofman, line 1: I'm actually somewhat concerned this circular definition will cause these function to not be types properly.

mhofman (changes requested):
I'd feel a lot more confident if there was a `test-types.js` file which main purpose was to verify the types of the public exports are as expected. Or at the very least, have the regular test file be `// @ts-check`

packages/stream/index.d.ts
kriskowal: Would it be better to not have `index.d.ts` and not have `"types": "./index.d.ts"` in `package.json`?

packages/stream/index.js
kriskowal: Freezing the IterationResult objects makes them Readonly and they cannot be passed where IterationResult is expected, because the receiver hasn’t marked them as “WontTouch” (const& in C++ parlance), even though the receiver does not touch.

If TypeScript were useful for communicating Readonly, const reference marking would have to be pervasive. C++ does this and the cognitive burden on the developer is tremendous.

mhofman: Yeah so the problem is that most code, including official typings, are not strict enough in their readonly-ness. There might be an argument to help fix some of those. But I agree in the mean time, if it blows up too many things not in out control, then that's fine.

However I don't understand how `Readonly<IteratorResult<T>>` is different than `IteratorResult`. I guess declaring as an interface doesn't mean the methods are readonly, which is weird but understandable.

packages/stream/index.d.ts
mhofman: I'm honestly not sure. I probably would have gone with a fully manual `.d.ts` file that also includes `export declare function`. A test file should alleviate or confirm my concerns

kriskowal: Story turns out to be that we need an explicit external `index.d.ts`, an internal `types.d.ts` that is shared by `index.js` and `index.d.ts`, and an external validator package to verify `index.d.ts`. There is some possibility that `index.js` and `index.d.ts` might not agree, but this turns out to be a feature. For example, for the purposes of verifying internal integrity, it’s useful for the internal types to be more rigorous than the external types. The primer argument of pump and prime is almost always undefined and optional in practice, but requires unnecessarily verbose handling if it’s optional internally.

packages/stream/index.js
kriskowal: Please take a look again. It left a note about the error I ran into attempting to get the type system to recognize the entangled queues as producing complementary streams. When I added the external type validation, a bunch of things needed changing, but at the end of the day, the definition of `makePipe` in `index.d.ts` turned out to be useful, and the internal JSDoc definition could not be compelled to cooperate.

packages/stream/types.d.ts
kriskowal: I ended up needing to do this in a very different way and these types do not exist now.

packages/stream/index.d.ts
mhofman: I usually end up with a `.d.ts` that are more strict, and the implementation that handles the generic/wide case. Especially in the case of overloads (which can't always be expressed by generics only).

packages/stream-types-test/validation.ts
```diff
@@ -0,0 +1,48 @@
...
+/// <reference types="ses"/>
+
+import { makeQueue, makeStream, makePipe } from '@endo/stream';
+// eslint-disable-next-line
+import type { Stream, Reader, Writer } from '@endo/stream';
+
+async () => {
+  const q = makeQueue<number>();
+  q.put(1);
+  q.put(Promise.resolve(1));
```
mhofman, line 11: Can we add `put('foo')` with `@ts-expect-error` ?

packages/stream/index.js
```diff
@@ -130,18 +103,14 @@ export const makeStream = (acks, data) => {
...
-/**
- * @template T
- * @template U
- * @template TReturn
- * @template UReturn
- * @returns {[Stream<T, U, TReturn>, Stream<U, T, UReturn>]}
- */
+// JSDoc TypeScript seems unable to express this particular function's
+// entanglement of queues, but the definition in index.d.ts works for the end
+// user.
```
mhofman, line 108: I believe one way might be to give an explicit type to the `data` and `acks` variables.

packages/stream-types-test/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "noEmit": true,
+    "target": "esnext",
+    "module": "esnext",
+    "moduleResolution": "node",
+    "strict": true,
+    "strictNullChecks": true,
+    "noImplicitAny": true,
```
mhofman, line 9: Thanks for the strictness :)

packages/stream-types-test/validation.ts
```diff
@@ -0,0 +1,48 @@
...
+async () => {
+  const [reader, writer] = makePipe<number>();
+  for await (const _ of reader) {}
+  // @ts-expect-error
+  for await (const _ of writer) {}
+};
+
+async () => {
+  const [reader, writer] = makePipe<undefined, number>();
+  // @ts-expect-error
```
mhofman, line 35: I'm confused why there is an expected error here actually.

mhofman (approved):
Thanks for adding the test package. I'm not entirely sure I understand the stream reader expected error though.
Also we could probably add more expected error tests when feeding incompatible types.

packages/stream/index.js
kriskowal: I tried that to no avail.

packages/stream-types-test/validation.ts
kriskowal: The reader and writer types are inverted. The for loop is only able to send undefined forward, but the type of reader obliges it to send numbers (and receive undefined). The second loop works.

mhofman: It just feels so awkward to read from a writer! I realize that makePipe is actually completely reversible, but it still messes with my brain, and I'm wondering if there is a use case for this, or if we shouldn't instead be opiniated and build a pipe where the `reader` always discards the `value` argument, and where the `writer` always returns `undefined` (I know it wouldn't be a simple `makeStream` with the same underlying queues anymore).

PR #972, opened 2022-01-01
feat(stream): Add stream map functions

kriskowal:
This change introduces `mapReader` and `mapWriter`, which obviate the need for the stream-json package entirely. This became clear when I realized that I ought to factor `stream-text` out of `stream-json` and the resulting packages with more focused scope were nearly identical. Consequently this change includes a second commit that just removes the unpublished `@endo/stream-json` package.


packages/stream/test/test-map.js
```diff
@@ -0,0 +1,77 @@
...
+test('map reader', async (/** @type {import('ava').Assertions} */ t) => {
+  const [consumeBytesFrom, produceBytesTo] = makePipe();
+  const consumeFrom = mapReader(consumeBytesFrom, (
+    /** @type {Uint8Array} */ bytes,
+  ) => textDecoder.decode(bytes));
+  const produceTo = mapWriter(produceBytesTo, (/** @type {string} */ text) =>
+    textEncoder.encode(text),
+  );
+
+  const order = ['', 'Hello, World!', '\x00', '😇   '];
```
mhofman, line 24: I probably would have picked 😈 for mischievousness ;)

packages/stream/test/test-map.js
```diff
@@ -0,0 +1,77 @@
...
+  );
+
+  const makeProducer = async () => {
+    await produceTo.throw(new Error('Exit early'));
+  };
+
+  const makeConsumer = async () => {
+    try {
+      for await (const _ of consumeFrom) {
+        t.fail();
```
mhofman, line 69: Can we add a comment here clarifying this would be caught by the catch block but not match the `error.message` test?

mhofman (approved):
One minor nit. Looks great, much simpler and more generic!

packages/stream/test/test-map.js
kriskowal: I don’t follow. `t.fail()` doesn’t throw. This is to verify that no iterations transit before it throws.

mhofman: Oh it doesn't? Other test runners usually have their `.fail()` throw to interrupt immediately (aside from marking the test as failed).

PR #973, opened 2022-01-01
feat(stream): Add pump and prime

kriskowal:
This change introduces pump and prime functions to the stream package.

Pump moves iterations from a reader to a writer and returns a promise for when either the reader or writer closes. The implementation of pump was more concise before satisfying TypeScript because the `tick` and `tock` internal phases were possible to express with a single `push` function that alternated direction. I found the duplication and mutual recursion tolerable.

Prime is an adapter for a generator function when it is used as a writer, which “primes” the generator, causing it to advance to the first yield expression immediately. Some nuance follows to handle the cases that the generator returns or throws before reaching a yield.

### DIFF
--- a/packages/stream-types-test/validation.ts
+++ b/packages/stream-types-test/validation.ts
@@ -1,7 +1,7 @@
 /* eslint-disable no-unused-vars, no-underscore-dangle, no-empty */
 /// <reference types="ses"/>
 
-import { makeQueue, makeStream, makePipe, mapReader, mapWriter } from '@endo/stream';
+import { makeQueue, makeStream, makePipe, pump, prime, mapReader, mapWriter } from '@endo/stream';
 // eslint-disable-next-line
 import type { Stream, Reader, Writer } from '@endo/stream';
 
@@ -47,6 +47,39 @@ async () => {
   for await (const value of writer) {
     const cast : number = value;
   }
+};
+
+async () => {
+  const [r1, writer] = makePipe<string, number>();
+  const [reader, w2] = makePipe<string, number>();
+  await pump(writer, reader, 1);
+};
+
+async () => {
+  const s: Stream<string> =
+    prime(async function *generator() {
+      yield 'A';
+      return undefined;
+    }());
+  s.return(undefined);
+};
+
+async () => {
+  const s: Stream<string, number> =
+    prime(async function *generator() {
+      const n: number = yield 'A';
+      return undefined;
+    }(), 1);
+  s.return(undefined);
+};
+
+async () => {
+  const s: Stream<string, number, bigint, bigint> =
+    prime(async function *generator() {
+      const n: number = yield 'A';
+      return 1n;
+    }(), 1);
+  s.return(2n);
 };
 
 async () => {
--- a/packages/stream/README.md
+++ b/packages/stream/README.md
@@ -20,7 +20,7 @@ Awaiting the returned promise slows the writer to match the pace of the reader.
 
 ## Reading
 
-To read from a string, await the value returned by the next method.
+To read from a stream, await the value returned by the next method.
 
 ```js
 for await (const value of reader) {
@@ -68,6 +68,60 @@ An async queue ensures that the promises returned by `get` and accepted by
 which promises settle.
 A stream is consequently a pair of queues that transport iteration results,
 one to send messages forward and another to receive acknowledgements.
+
+## Pump
+
+The `pump` function pumps iterations from a reader to a writer.
+The pump must be primed with the first acknowledgement to send to the reader,
+typically `undefined`, as in `reader.next(undefined)`.
+This makes the parity of a pump "odd", because the reader needs a free
+acknowledgement to start.
+This is in contrast to a pipe, which has "even" parity, because the reader and
+writer can both proceed initially.
+
+So, for example, we can implement `cat` in Node.js by pumping stdin to stdout.
+
+```js
+import { makeNodeWriter, makeNodeReader } from '@endo/stream-node';
+
+const writer = makeNodeWriter(process.stdout);
+const reader = makeNodeReader(process.stdin);
+await pump(writer, reader);
+```
+
+## Prime
+
+Async generator functions are very useful for making reader adapters.
+
+```js
+async function *double(reader) {
+  for await (const value of reader) {
+    yield value * 2;
+  }
+  return undefined;
+}
+```
+
+However, async generator functions can also serve as writers, because `yield`
+evaluates to the argument passed to `next`.
+However, generator writers have odd parity, meaning the first value sent to a
+generator function has nowhere to go and gets discarded as the program counter
+proceeds from the beginning of the function to the first `yield`, `return`, or
+`throw`.
+
+The `prime` function compensates for this by sending a primer to the generator
+once.
+
+```js
+async function *logGenerator() {
+  for (;;) {
+    console.log(yield);
+  }
+}
+
+const writer = prime(logGenerator());
+await writer.next('First message is not discarded');
+```
 
 ## Hardening
 
--- a/packages/stream/README.md
+++ b/packages/stream/README.md
@@ -7,6 +7,10 @@ and consumer to producer.
 Streams are therefore symmetric.
 The same stream type serves for both a reader and a writer.
 
+These streams depend on full Endo environment initialization, as with `@endo/init`
+to ensure that they are run in Hardened JavaScript with remote promise support
+(eventual send).
+
 ## Writing
 
 To write to a stream, give a value to the next method.
--- a/packages/stream/index.d.ts
+++ b/packages/stream/index.d.ts
@@ -1,45 +1,69 @@
+/* eslint-disable no-unused-vars, import/no-unresolved */
+import { AsyncQueue, Stream } from './types.js';
+
 export * from './types.js';
-import { AsyncQueue, Stream, Reader, Writer } from './types.js';
 
-export function makeQueue<TValue>(): AsyncQueue<TValue>;
+export declare function makeQueue<TValue>(): AsyncQueue<TValue>;
 
-export function makeStream<
+export declare function makeStream<
   TRead,
   TWrite = undefined,
   TReadReturn = undefined,
-  TWriteReturn = undefined,
+  TWriteReturn = undefined
 >(
   acks: AsyncQueue<IteratorResult<TRead, TReadReturn>>,
   data: AsyncQueue<IteratorResult<TWrite, TWriteReturn>>,
 ): Stream<TRead, TWrite, TReadReturn, TWriteReturn>;
 
-export function makePipe<
+export declare function makePipe<
   TRead,
   TWrite = undefined,
   TReadReturn = undefined,
-  TWriteReturn = undefined,
+  TWriteReturn = undefined
 >(): [
   Stream<TRead, TWrite, TReadReturn, TWriteReturn>,
   Stream<TWrite, TRead, TWriteReturn, TReadReturn>,
 ];
 
-export function mapReader<
+export declare function pump<
+  TRead,
+  TWrite = unknown,
+  TReadReturn = unknown,
+  TWriteReturn = unknown
+>(
+  writer: Stream<TWrite, TRead, TWriteReturn, TReadReturn>,
+  reader: Stream<TRead, TWrite, TReadReturn, TWriteReturn>,
+  primer?: TWrite,
+): Promise<void>;
+
+export declare function prime<TRead>(
+  writer: AsyncGenerator<TRead, undefined, undefined>,
+): // primer is implicitly undefined for this overload.
+Stream<TRead, undefined, undefined, undefined>;
+// ESLint hasn't heard about overloads.
+// eslint-disable-next-line no-redeclare
+export declare function prime<TRead, TWrite = unknown, TReturn = unknown>(
+  writer: AsyncGenerator<TRead, TReturn, TWrite>,
+  primer: TWrite,
+): Stream<TRead, TWrite, TReturn, TReturn>;
+
+export declare function mapReader<
   TReadIn,
   TReadOut = TReadIn,
   TWrite = undefined,
   TReadReturn = undefined,
-  TWriteReturn = undefined,
+  TWriteReturn = undefined
 >(
   reader: Stream<TReadIn, TWrite, TReadReturn, TWriteReturn>,
   transform: (value: TReadIn) => TReadOut,
 ): Stream<TReadOut, TWrite, TReadReturn, TWriteReturn>;
 
-export function mapWriter<
+export declare function mapWriter<
   TWriteIn,
   TWriteOut = TWriteIn,
   TRead = undefined,
   TReadReturn = undefined,
-  TWriteReturn = undefined,
+  TWriteReturn = undefined
 >(
   writer: Stream<TRead, TWriteOut, TReadReturn, TWriteReturn>,
   transform: (value: TWriteIn) => TWriteOut,
--- a/packages/stream/index.js
+++ b/packages/stream/index.js
@@ -11,6 +11,8 @@
 // @ts-check
 /// <reference types="ses"/>
 
+import { E } from '@endo/eventual-send';
+
 /**
  * @template T
  * @typedef {{
@@ -127,7 +129,8 @@ harden(makePipe);
 export const pump = async (writer, reader, primer) => {
   /** @param {Promise<IteratorResult<TRead, TReadReturn>>} promise */
   const tick = promise =>
-    promise.then(
+    E.when(
+      promise,
       result => {
         if (result.done) {
           return writer.return(result.value);
@@ -143,7 +146,8 @@ export const pump = async (writer, reader, primer) => {
     );
   /** @param {Promise<IteratorResult<TWrite, TWriteReturn>>} promise */
   const tock = promise =>
-    promise.then(
+    E.when(
+      promise,
       result => {
         if (result.done) {
           return reader.return(result.value);
--- a/packages/stream/index.js
+++ b/packages/stream/index.js
@@ -116,6 +116,99 @@ export const makePipe = () => {
 harden(makePipe);
 
 /**
+ * @template TRead
+ * @template TWrite
+ * @template TReadReturn
+ * @template TWriteReturn
+ * @param {import('./types.js').Stream<TWrite, TRead, TWriteReturn, TReadReturn>} writer
+ * @param {import('./types.js').Stream<TRead, TWrite, TReadReturn, TWriteReturn>} reader
+ * @param {TWrite} primer
+ */
+export const pump = async (writer, reader, primer) => {
+  /** @param {Promise<IteratorResult<TRead, TReadReturn>>} promise */
+  const tick = promise =>
+    promise.then(
+      result => {
+        if (result.done) {
+          return writer.return(result.value);
+        } else {
+          // Behold: mutual recursion.
+          // eslint-disable-next-line no-use-before-define
+          return tock(writer.next(result.value));
+        }
+      },
+      (/** @type {Error} */ error) => {
+        return writer.throw(error);
+      },
+    );
+  /** @param {Promise<IteratorResult<TWrite, TWriteReturn>>} promise */
+  const tock = promise =>
+    promise.then(
+      result => {
+        if (result.done) {
+          return reader.return(result.value);
+        } else {
+          return tick(reader.next(result.value));
+        }
+      },
+      (/** @type {Error} */ error) => {
+        return reader.throw(error);
+      },
+    );
+  await tick(reader.next(primer));
+  return undefined;
+};
+harden(pump);
+
+/**
+ * @template TRead
+ * @template TWrite
+ * @template TReturn
+ * @param {AsyncGenerator<TRead, TReturn, TWrite>} generator
+ * @param {TWrite} primer
+ */
+export const prime = (generator, primer) => {
+  // We capture the first returned promise.
+  const first = generator.next(primer);
+  /** @type {IteratorResult<TRead, TReturn>=} */
+  let result;
+  const primed = harden({
+    /** @param {TWrite} value */
+    async next(value) {
+      if (result === undefined) {
+        result = await first;
+        if (result.done) {
+          return result;
+        }
+      }
+      return generator.next(value);
+    },
+    /** @param {TReturn} value */
+    async return(value) {
+      if (result === undefined) {
+        result = await first;
+        if (result.done) {
+          return result;
+        }
+      }
+      return generator.return(value);
+    },
+    /** @param {Error} error */
+    async throw(error) {
+      if (result === undefined) {
+        result = await first;
+        if (result.done) {
+          throw error;
+        }
+      }
+      return generator.throw(error);
+    },
+  });
+  return primed;
+};
+harden(prime);
+
+/**
  * @template TIn
  * @template TOut
  * @param {import('./types.js').Reader<TIn>} reader
--- a/packages/stream/test/test-prime.js
+++ b/packages/stream/test/test-prime.js
@@ -1,0 +1,101 @@
+// @ts-check
+/* eslint-disable require-yield, no-empty-function */
+
+import 'ses';
+import './lockdown.js';
+
+import rawTest from 'ava';
+import { wrapTest } from '@endo/ses-ava';
+import { prime } from '../index.js';
+
+const test = wrapTest(rawTest);
+
+test('prime single next', async (/** @type {import('ava').ExecutionContext} */ t) => {
+  async function* single() {
+    t.is(yield, 0);
+  }
+
+  const iterator = prime(single());
+  const { done, value } = await iterator.next(0);
+  t.is(done, true);
+  t.is(value, undefined);
+});
+
+test('prime single return', async (/** @type {import('ava').ExecutionContext} */ t) => {
+  async function* empty() {
+    return 'Z';
+  }
+
+  const iterator = prime(empty());
+  const { done, value } = await iterator.return(0);
+  t.is(done, true);
+  t.is(value, 'Z');
+});
+
+test('prime empty throw in', async (/** @type {import('ava').ExecutionContext} */ t) => {
+  async function* empty() {
+    return 'Z';
+  }
+
+  const iterator = prime(empty());
+  try {
+    await iterator.throw(new Error('Abort'));
+    t.fail();
+  } catch (error) {
+    t.is(error.message, 'Abort');
+  }
+});
+
+test('prime single throw in', async (/** @type {import('ava').ExecutionContext} */ t) => {
+  async function* single() {
+    try {
+      t.is(yield, 0);
+    } catch (error) {
+      t.is(error.message, 'Abort');
+      return 'Z';
+    }
+    return 'A';
+  }
+
+  const iterator = prime(single());
+  const { done, value } = await iterator.throw(new Error('Abort'));
+  t.is(done, true);
+  t.is(value, 'Z');
+});
+
+test('prime single throw', async (/** @type {import('ava').ExecutionContext} */ t) => {
+  async function* empty() {
+    throw new Error('Abort');
+  }
+
+  const iterator = prime(empty());
+  try {
+    await iterator.next(0);
+    t.fail('reached beyond end of generator');
+  } catch (error) {
+    t.is(error.message, 'Abort');
+  }
+});
+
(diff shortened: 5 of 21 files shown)


packages/stream/test/test-prime.js
```diff
@@ -0,0 +1,98 @@
...
+
+test('prime empty throw in', async (/** @type {import('ava').ExecutionContext} */ t) => {
+  async function* empty() {
+    return 'Z';
+  }
+
+  const iterator = prime(empty());
+  const { done, value } = await iterator.throw(new Error('Abort'));
+  t.is(done, true);
+  t.is(value, 'Z');
```
mhofman, line 43: This seems somewhat surprising. I'm actually wondering if the default behavior shouldn't be to rethrow the error if the result of priming is a close (`done: true`).

packages/stream/test/test-prime.js
```diff
@@ -0,0 +1,98 @@
...
+
+test('prime throw case', async (/** @type {import('ava').ExecutionContext} */ t) => {
+  async function* temperamental() {
+    throw new Error('Abort');
+  }
+
+  const iterator = prime(temperamental());
+  try {
+    await iterator.next();
+    t.pass();
```
mhofman, line 94: Why do we `pass` here?

mhofman (changes requested):
This is a partial review only for `prime`.

I'm not sure I understand the `generator.next()` call for empty generators. I also would expect the `throw` behavior to actually throw when priming an empty generator (it's kind of a lie but, it avoids the error to be swallowed).

packages/stream/index.d.ts
```diff
@@ -23,6 +23,28 @@ export function makePipe<
...
+export function prime<
+  TRead,
+  TWrite = unknown,
+  TReturn = unknown,
+>(
+  writer: AsyncGenerator<TRead,  TReturn, TWrite>,
+  // Omitted from type definition because undefined is
+  // nearly always adequate:
+  // primer: TWrite,
+): Stream<TRead, TWrite, TReturn, TReturn>;
```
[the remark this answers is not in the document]
kriskowal: Well, that’s neat.

packages/stream/index.js
```diff
@@ -115,6 +115,94 @@ export const makePipe = () => {
...
+ */
+export const prime = (generator, primer) => {
+  // We capture the first returned promise.
+  let first = generator.next(primer);
+  const primed = harden({
+    /** @param {TWrite} value */
+    async next(value) {
+      const result = await first;
+      if (result.done) {
+        first = generator.next(primer);
```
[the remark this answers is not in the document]
kriskowal: No test fails without these, so I’ll follow up with a change that makes `first` const.

packages/stream/test/test-prime.js
kriskowal: I concur. I’m following up with a change that causes throw to rethrow if the primer ran to the return.

packages/stream/test/test-prime.js
kriskowal: I misspelled `fail`. Thank you!

packages/stream/index.d.ts
```diff
@@ -40,11 +40,17 @@ export function prime<
...
 ): Stream<TRead, TWrite, TReturn, TReturn>;
 
+// Overload:
+export function prime<
+  TRead,
+>(
+  writer: AsyncGenerator<TRead, undefined, undefined>,
+  // primer is implicitly undefined for this overload.
+): Stream<TRead, undefined, undefined, undefined>;
+
```
mhofman, line 53: TS chooses the overload using top down, and uses the first one that matches. In this case, I'm not sure but the first one may actually match, and TS may pick it with `TWrite` defaulting to `unknown`.

packages/stream-types-test/validation.ts
```diff
@@ -56,13 +56,31 @@ async () => {
 };
 
 async () => {
-  const s: Stream<string, number, boolean, boolean> =
+  const s: Stream<string> =
     prime(async function *generator() {
-    const n: number = yield 'A';
-    return false;
-  }());
+      yield 'A';
```
mhofman, line 61: To make sure the type yielded is really `undefined` and not `unknown`:
```suggestion
      const r: undefined = yield 'A';
```

mhofman (changes requested):
I believe the overloaded is in the wrong order.

packages/stream/test/test-prime.js
mhofman: Did this fix get stacked on the wrong PR?

kriskowal: Yes, I’ve shifted the commit up to this PR.

packages/stream-types-test/validation.ts
kriskowal: Addressed.

packages/stream/index.d.ts
kriskowal: Addressed.

packages/stream/index.d.ts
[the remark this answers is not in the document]
kriskowal: For future Kris: Would be nice to have a failing test.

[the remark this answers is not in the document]
kriskowal: From my read of the tests in `stream-types-test`, I think that I already have coverage for any type error that might arise from these two overloads being in the wrong order. I have tests for both cases and they appear to pass both before and after the order fix.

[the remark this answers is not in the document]
mhofman: I am surprised that it passes before, but I am ok not understanding why

packages/stream/index.js
```diff
@@ -115,6 +115,91 @@ export const makePipe = () => {
...
+        } else {
+          return tick(reader.next(result.value));
+        }
+      },
+      (/** @type {Error} */ error) => {
+        return reader.throw(error);
+      },
+    );
+  await tick(reader.next(primer));
+  return undefined;
```
mhofman, line 159: Why not return the value of `tick` ?

I'm curious as to what the return type actually is of tick & tock. Seems like it would be `Promise<IteratorResult<never, TReadReturn | TWriteReturn>>` depending on which stream closed first?

packages/stream/test/test-pump.js
```diff
@@ -0,0 +1,323 @@
...
+});
+
+test('pump target closes early', async (/** @type {import('ava').ExecutionContext} */ t) => {
+  t.plan(5);
+
+  async function* source() {
+    for (let i = 0; i < 10; i += 1) {
+      yield i;
+    }
+    return 'X';
```
mhofman, line 48: can we add a `finally` with an assertion here too?

packages/stream/test/test-pump.js
```diff
@@ -0,0 +1,323 @@
...
+      t.deepEqual(collected, [0, 1, 2, 3, 4, 5, 6]);
+    } finally {
+      t.assert(true);
+    }
+  }
+
+  try {
+    await pump(prime(target()), source());
+  } catch (error) {
+    // Between Node.js 12 and 14, V8 stopped propagating the error from the
```
mhofman, line 90: ugh, that's sad

mhofman (approved):
Pump looks fine. I still need to compare it to my version, which was fairly different, and I need to figure out why (one reason was that I tried to parametrize the closing propagation).

packages/stream/test/test-pump.js
erights: What does this mean? I don't understand the comment. What did node stop propagating?

packages/stream/index.js
kriskowal: That is almost certainly the exact type and I think the only useful part of that is the possibility that it terminates on an error. It seems less regrettable for now to return `Promise<undefined>` and consider the union type only if and when we find a motivating use.

packages/stream/index.d.ts
```diff
@@ -23,6 +23,34 @@ export function makePipe<
...
 ];
 
+export function pump<
+  TRead,
+  TWrite = unknown,
+  TReadReturn = unknown,
+  TWriteReturn = unknown,
+>(
+  writer: Stream<TWrite, TRead, TWriteReturn, TReadReturn>,
+  reader: Stream<TRead, TWrite, TReadReturn, TWriteReturn>,
```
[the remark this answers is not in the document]
kriskowal: My brain concurs. The precedents I’m coming from are `copyWithin` and `encodeInto`. I could be convinced to turn this around.

packages/stream/test/test-pump.js
kriskowal: I will change the comment to this prose:
```
    // In Node.js 12, the error thrown on the last line of the `source`
    // generator does not propagate to the promise returned by `pump`.
    // In Node.js 14 and 16, the error does propagate.
```

packages/stream/index.d.ts
```diff
@@ -23,6 +23,34 @@ export function makePipe<
   Stream<TWrite, TRead, TWriteReturn, TReadReturn>,
 ];
```
erights, line 25: All the problems I'm about to mention are problem I would have expected `yarn lint` to have complained about. But I pulled the `kris-stream-pump` fork, and `yarn lint` succeeds silently. This is ALARMING! What alerted me to investigate was that vscode gave me a bunch of red underlines. These differences remain even after I tell vscode to use the same version of TS that endo uses.

This file imports './types.js'. But there is no './types.js'. What is clearly intended is 'types.d.ts'

After making that fix, I get a complaint that the `import` comes before the export. Quite right.

Are the function declarations in this file the correct TS way to declare a function type rather to define a function? Clearly it must be because there are no bodies. However, some of the complaints I get (again, vscode only) would only make sense for function definitions.

For parameter names, it complains that they aren't used in the (non-existent) body.

I know that TS supports function overload. However, for `prime` it instead complains about a duplicate definition.

packages/stream/README.md
```diff
@@ -69,6 +69,50 @@ which promises settle.
 A stream is consequently a pair of queues that transport iteration results,
 one to send messages forward and another to receive acknowledgements.
 
+## Pump
```
erights, line 72: For each of the other sections of this README, a trivial code example helped tremendously. Could we have one for `pump` too? Especially if it fits with the others?

packages/stream/README.md
```diff
@@ -69,6 +69,50 @@ which promises settle.
...
+```js
+async function *logGenerator() {
+  for (;;) {
+    console.log(yield);
+  }
+}
+
+const writer = prime(logGenerator());
+await writer.next('First message is not discarded');
+```
```
erights, line 114: Very nice example.

packages/stream/index.js
```diff
@@ -115,6 +115,91 @@ export const makePipe = () => {
...
+ * @template TReadReturn
+ * @template TWriteReturn
+ * @param {import('./types.js').Stream<TWrite, TRead, TWriteReturn, TReadReturn>} writer
+ * @param {import('./types.js').Stream<TRead, TWrite, TReadReturn, TWriteReturn>} reader
+ * @param {TWrite} primer
+ */
+export const pump = async (writer, reader, primer) => {
+  /** @param {Promise<IteratorResult<TRead, TReadReturn>>} promise */
+  const tick = promise =>
+    promise.then(
```
erights, line 130: Since we are writing this for a Hardened JS environment (thanks!), the abstractions should be more defensive. This synchronous call to `.then` enables the provider of the alleged promise to mount a reentrancy attack. Can we use `E.when`? Depends on how you want to layer endo, but I imagine eventual-send is lower level than stream.

Likewise for other calls to `.then`

Even an `await`, for all its flaws, is safer than a direct call to an unverified `.then`.

Btw, `E.when` accumulates deep stacks for debugging. Nether `.then` nor `await` do.

packages/stream/index.js
```diff
@@ -115,6 +115,91 @@ export const makePipe = () => {
...
+          // eslint-disable-next-line no-use-before-define
+          return tock(writer.next(result.value));
+        }
+      },
+      (/** @type {Error} */ error) => {
+        return writer.throw(error);
+      },
+    );
+  /** @param {Promise<IteratorResult<TWrite, TWriteReturn>>} promise */
+  const tock = promise =>
```
erights, line 145: Even though `tick` and `tock` are essentially identical, thanks for not arranging to use one function for both. Seeing the two next to each other with the names switched in place made this much easier to follow.

Elegant!

packages/stream/index.d.ts
kriskowal: I can fix the import/export order. Apart from that, cc @mhofman for clues. I’m not sure I can account for differences between the language server protocol and what `tsc` reports, but only `tsc` and `eslint` are shared objective reality with CI.

kriskowal: I can say that `ses-types-test` is considerably more useful for validating our type definitions than `tsc` on a package’s internals!

packages/stream/index.js
erights: Just verifying: With the code as is, the error case is still handled correctly because the `await` will turn a rejection into a thrown error, and the `async` will turn the thrown error back into a rejection.

packages/stream/index.js
```diff
@@ -115,6 +115,91 @@ export const makePipe = () => {
...
+  const first = generator.next(primer);
+  const primed = harden({
+    /** @param {TWrite} value */
+    async next(value) {
+      const result = await first;
+      if (result.done) {
+        return result;
+      }
+      return generator.next(value);
+    },
```
erights, line 181: The `await` here causes an extra turn *each time* even though it can only be relevant the first time.

True as well for `return` and `throw`, but not a big deal there because they're normally only called one anyway. Nevertheless, if you used local state to remember the fulfillment of `first`, you could avoid all these unnecessary turns.

erights (approved):
LGTM. Elegant! I have always been skeptical of using one API for both push and pull. The need for `prime` amplified that feeling. However, the elegance of `pump` and its agnosticism on direction is pretty compelling evidence. I am now open to removing Updater in favor of next/return/throw. We should discuss sometime. But obviously, the PR should not wait for that.

packages/stream/index.d.ts
erights: FWIW, the following diff comes from taking each of the vscode popup suggestions. It made all the red go away, except for the overload warning on `prime`. While making these changes, I ran `yarn lint` every time to see if it was ok. It was *always* ok, which is still alarming. But obviously, this PR need not wait on any of this. I am *not* suggesting you adopt the changes below. Though feel free to if any look like a genuine improvement.

```js
diff --git a/packages/stream/index.d.ts b/packages/stream/index.d.ts
index 37a19483..3e007f64 100644
--- a/packages/stream/index.d.ts
+++ b/packages/stream/index.d.ts
@@ -1,5 +1,6 @@
-export * from './types.js';
-import { AsyncQueue, Stream, Reader, Writer } from './types.js';
+import { AsyncQueue, Stream } from './types.d.ts';
+
+export * from './types.d.ts';
 
 export function makeQueue<TValue>(): AsyncQueue<TValue>;
 
@@ -7,17 +8,17 @@ export function makeStream<
   TRead,
   TWrite = undefined,
   TReadReturn = undefined,
-  TWriteReturn = undefined,
+  TWriteReturn = undefined
 >(
-  acks: AsyncQueue<IteratorResult<TRead, TReadReturn>>,
-  data: AsyncQueue<IteratorResult<TWrite, TWriteReturn>>,
+  _acks: AsyncQueue<IteratorResult<TRead, TReadReturn>>,
+  _data: AsyncQueue<IteratorResult<TWrite, TWriteReturn>>,
 ): Stream<TRead, TWrite, TReadReturn, TWriteReturn>;
 
 export function makePipe<
   TRead,
   TWrite = undefined,
   TReadReturn = undefined,
-  TWriteReturn = undefined,
+  TWriteReturn = undefined
 >(): [
   Stream<TRead, TWrite, TReadReturn, TWriteReturn>,
   Stream<TWrite, TRead, TWriteReturn, TReadReturn>,
@@ -27,28 +28,22 @@ export function pump<
   TRead,
   TWrite = unknown,
   TReadReturn = unknown,
-  TWriteReturn = unknown,
+  TWriteReturn = unknown
 >(
-  writer: Stream<TWrite, TRead, TWriteReturn, TReadReturn>,
-  reader: Stream<TRead, TWrite, TReadReturn, TWriteReturn>,
-  primer?: TWrite,
+  _writer: Stream<TWrite, TRead, TWriteReturn, TReadReturn>,
+  _reader: Stream<TRead, TWrite, TReadReturn, TWriteReturn>,
+  _primer?: TWrite,
 ): Promise<void>;
 
:...skipping...
diff --git a/packages/stream/index.d.ts b/packages/stream/index.d.ts
index 37a19483..3e007f64 100644
--- a/packages/stream/index.d.ts
+++ b/packages/stream/index.d.ts
@@ -1,5 +1,6 @@
-export * from './types.js';
-import { AsyncQueue, Stream, Reader, Writer } from './types.js';
+import { AsyncQueue, Stream } from './types.d.ts';
+
+export * from './types.d.ts';
 
 export function makeQueue<TValue>(): AsyncQueue<TValue>;
 
@@ -7,17 +8,17 @@ export function makeStream<
   TRead,
   TWrite = undefined,
   TReadReturn = undefined,
-  TWriteReturn = undefined,
+  TWriteReturn = undefined
 >(
-  acks: AsyncQueue<IteratorResult<TRead, TReadReturn>>,
-  data: AsyncQueue<IteratorResult<TWrite, TWriteReturn>>,
+  _acks: AsyncQueue<IteratorResult<TRead, TReadReturn>>,
+  _data: AsyncQueue<IteratorResult<TWrite, TWriteReturn>>,
 ): Stream<TRead, TWrite, TReadReturn, TWriteReturn>;
 
 export function makePipe<
   TRead,
   TWrite = undefined,
   TReadReturn = undefined,
-  TWriteReturn = undefined,
+  TWriteReturn = undefined
 >(): [
   Stream<TRead, TWrite, TReadReturn, TWriteReturn>,
   Stream<TWrite, TRead, TWriteReturn, TReadReturn>,
@@ -27,28 +28,22 @@ export function pump<
   TRead,
   TWrite = unknown,
   TReadReturn = unknown,
-  TWriteReturn = unknown,
+  TWriteReturn = unknown
 >(
-  writer: Stream<TWrite, TRead, TWriteReturn, TReadReturn>,
-  reader: Stream<TRead, TWrite, TReadReturn, TWriteReturn>,
-  primer?: TWrite,
+  _writer: Stream<TWrite, TRead, TWriteReturn, TReadReturn>,
+  _reader: Stream<TRead, TWrite, TReadReturn, TWriteReturn>,
+  _primer?: TWrite,
 ): Promise<void>;
 
-export function prime<
-  TRead,
-  TWrite = unknown,
-  TReturn = unknown,
->(
-  writer: AsyncGenerator<TRead,  TReturn, TWrite>,
-  primer: TWrite,
+export function prime<TRead, TWrite = unknown, TReturn = unknown>(
+  _writer: AsyncGenerator<TRead, TReturn, TWrite>,
+  _primer: TWrite,
 ): Stream<TRead, TWrite, TReturn, TReturn>;
 
 // Overload:
-export function prime<
-  TRead,
->(
-  writer: AsyncGenerator<TRead, undefined, undefined>,
+export function prime<TRead>(
   // primer is implicitly undefined for this overload.
+  _writer: AsyncGenerator<TRead, undefined, undefined>,
 ): Stream<TRead, undefined, undefined, undefined>;
 
 export function mapReader<
@@ -56,10 +51,10 @@ export function mapReader<
   TReadOut = TReadIn,
   TWrite = undefined,
   TReadReturn = undefined,
-  TWriteReturn = undefined,
+  TWriteReturn = undefined
 >(
-  reader: Stream<TReadIn, TWrite, TReadReturn, TWriteReturn>,
-  transform: (value: TReadIn) => TReadOut,
+  _reader: Stream<TReadIn, TWrite, TReadReturn, TWriteReturn>,
+  _transform: (_value: TReadIn) => TReadOut,
 ): Stream<TReadOut, TWrite, TReadReturn, TWriteReturn>;
 
 export function mapWriter<
@@ -67,8 +62,8 @@ export function mapWriter<
   TWriteOut = TWriteIn,
   TRead = undefined,
   TReadReturn = undefined,
-  TWriteReturn = undefined,
+  TWriteReturn = undefined
 >(
-  writer: Stream<TRead, TWriteOut, TReadReturn, TWriteReturn>,
-  transform: (value: TWriteIn) => TWriteOut,
+  _writer: Stream<TRead, TWriteOut, TReadReturn, TWriteReturn>,
+  _transform: (_value: TWriteIn) => TWriteOut,
 ): Stream<TRead, TWriteIn, TReadReturn, TWriteReturn>;
diff --git a/packages/stream/types.d.ts b/packages/stream/types.d.ts
index cbd2d0c0..4f2d4342 100644
--- a/packages/stream/types.d.ts
+++ b/packages/stream/types.d.ts
@@ -1,6 +1,6 @@
 export interface AsyncQueue<TValue> {
-  put(value: TValue | Promise<TValue>): void,
-  get(): Promise<TValue>,
+  put(_value: TValue | Promise<TValue>): void;
+  get(): Promise<TValue>;
 }
 
 // Stream is nearly identical to AsyncGenerator and AsyncGenerator should
@@ -11,13 +11,23 @@ export interface Stream<
   TRead,
   TWrite = undefined,
   TReadReturn = undefined,
-  TWriteReturn = undefined,
+  TWriteReturn = undefined
 > {
-  next(value: TWrite): Promise<IteratorResult<TRead, TReadReturn>>,
-  return(value: TWriteReturn): Promise<IteratorResult<TRead, TReadReturn>>,
-  throw(error: Error): Promise<IteratorResult<TRead, TReadReturn>>,
-  [Symbol.asyncIterator](): Stream<TRead, TWrite, TReadReturn, TWriteReturn>,
+  next(_value: TWrite): Promise<IteratorResult<TRead, TReadReturn>>;
+  return(_value: TWriteReturn): Promise<IteratorResult<TRead, TReadReturn>>;
+  throw(_error: Error): Promise<IteratorResult<TRead, TReadReturn>>;
+  [Symbol.asyncIterator](): Stream<TRead, TWrite, TReadReturn, TWriteReturn>;
 }
 
-export type Reader<TRead, TReadReturn = undefined> = Stream<TRead, undefined, TReadReturn, undefined>;
-export type Writer<TWrite, TWriteReturn = undefined> = Stream<undefined, TWrite, undefined, TWriteReturn>;
+export type Reader<TRead, TReadReturn = undefined> = Stream<
+  TRead,
+  undefined,
+  TReadReturn,
+  undefined
+>;
+export type Writer<TWrite, TWriteReturn = undefined> = Stream<
+  undefined,
+  TWrite,
+  undefined,
+  TWriteReturn
+>;
```

erights: Notice both the removed commas, and the commas that were replaced by semicolons.

erights:
Oh! What about NEWS.md ?

packages/stream/index.js
kriskowal: I’ve hesitated to introduce that dependency, but I see now that E doesn’t actually entrain CapTP or Marshal, just the HandledPromise/Promise.delegate shim.

Streams are going to have to entrain a dependency on eventual-send when we want streams to have hooks for the transport layer. So I’m okay with bringing that dependency in.

kriskowal: Changing `promise.then` to `E.when(promise` suggests we’re expecting an adversarial reader or writer. Do we need to arrange a similar defense for the stream methods, `next`, `throw`, `return`?

kriskowal:
> Oh! What about NEWS.md ?

First published version gets a freebie. No changes since never published! 😅

packages/stream/index.d.ts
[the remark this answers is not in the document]
mhofman: Oh yeah, and `Object.assign`...

packages/stream/index.js
mhofman: Agreed, we can always add later

packages/stream/index.js
erights: If cheap enough, yes. Most abstractions should be written defensively by default. It is much easier to understand how to use them as a component within adversarial code. Every abstraction that isn't defensive should be marked somehow as fragile, so that it is never released across a trust boundary. Ideally, we would even check and prevent such leakage automatically.

What would it take for these to be defensive?

packages/stream/index.js
kriskowal: Got this and managed to not make the mistake of calling `next` on concurrent cache misses. Just once up front!

packages/stream/index.js
kriskowal: Probably exactly the same mechanism as `E.when`. We could wrap any stream such that all its methods get called in turns that resolve the previously returned promise.

kriskowal: That would be expensive enough that I would like to defer the choice to the user. https://github.com/endojs/endo/issues/1003

kriskowal:
I’ve flattened the commits on this since I have stamps and I’m about to merge, but please note that making eventual-send a requirement for all streams has the side effect of making the existing `netstring` package have this requirement as well. This should not impact Agoric SDK since `netstring` is only used in full Endo environments, but it does require a version bump.

packages/stream/index.d.ts
mhofman: > This file imports './types.js'. But there is no './types.js'. What is clearly intended is 'types.d.ts'

@erights, no, this is a quirk of TypeScript, which assumes a `.ts` ends up compiled as `.js` and `.d.ts`, and it doesn't want to rewrite imports, so it expects you still import `.js` (when using extensions). That means a `.d.ts` file shadows `.js` for anything type related (and ignore the content of the `.js` file), even when the import is `.js` everwhere.

> But I pulled the `kris-stream-pump` fork, and `yarn lint` succeeds silently. This is ALARMING!

@kriskowal, same here. After digging in, eslint by default doesn't recurse in subdirectories so `eslint .` will not check `test/`. Also by default `.ts` files are not included. That leaves only `index.js` being checked by eslint. I checked my `package.json`, and use `eslint '**/*.{js,ts}'` instead.

That said, some of the eslint complaints I see in VS code in the `.d.ts` are non-sense. Not sure why eslint is busted regarding imports in this package (I have no such issue in my runner).

> Are the function declarations in this file the correct TS way to declare a function type rather to define a function?

Actually I missed this, but `.d.ts` file should all have the `declare` keyword in front of `function` or other JavaScript declaration (not needed for `interface` or `type`).
Technically to disambiguate `import`/`export`, you should use `import type`, but that runs the risk of being incompatible with fairly old TypeScript versions (I believe those are old enough by now to ignore)

> some of the complaints I get (again, vscode only) would only make sense for function definitions.
> 
> For parameter names, it complains that they aren't used in the (non-existent) body.

Checks notes, and see that most of my `.d.ts` files have `/* eslint-disable no-unused-vars,no-redeclare */`

I believe we may need a more holistic approach, but for now it should help.

packages/stream/index.js
mhofman: Agreed to defer discussion. It should be legal to call `.next` without waiting for a previous promise to resolve.

packages/stream/index.d.ts
```diff
@@ -1,46 +1,71 @@
...
   TRead,
   TWrite = undefined,
   TReadReturn = undefined,
-  TWriteReturn = undefined,
+  TWriteReturn = undefined
 >(
-  acks: AsyncQueue<IteratorResult<TRead, TReadReturn>>,
-  data: AsyncQueue<IteratorResult<TWrite, TWriteReturn>>,
+  _acks: AsyncQueue<IteratorResult<TRead, TReadReturn>>,
+  _data: AsyncQueue<IteratorResult<TWrite, TWriteReturn>>,
```
mhofman, line 14: I disagree with this change. The type definition file should have meaningfully named parameters as they serve the purpose of documentation. Satisfying inapplicable eslint rules shouldn't impact the quality of definitions.

mhofman (dismissed):
I believe some of the changes to the `.d.ts` are negative, please undo the `_` prefixing, and simply add a disable header for the rules eslint isn't happy about and which are not applicable to definition files.

Also I think we should add `declare` before `function` even if it's not strictly necessary. Also, let's make sure the `.d.ts` and `test/` files are included in the command line eslint.

packages/stream/test/test-pump.js
mhofman: Ok I re-read this, and now I'm thinking this was a bug in older v8 versions that was fixed? I find this error swallowing fairly surprising.

packages/stream/index.d.ts
kriskowal: Using `'**/*.{js,ts}'` in scripts runs us up against the problem that you can't use double quotes on unix and you can't use single quotes on Windows. It can be solved, but the cure is worse than the disease. But, `overrides: [{files: [...` seems to work and should get floated up to `@endo/eslint-config`. I’ve added it locally for now.

PR #976, opened 2022-01-10
chore: Relieve dependence on colors

kriskowal:
The compromised `colors` package is pervasive enough that both `http-server` and `node-static` depend upon it. I’m comfortable just saying informally that you need to run `python3 -m http.server` for the demo.


warner:
Wanna change the `demo:` line to just use that python snippet, in lieu of documentation? Anyone reading that should be able to figure out how to run some equivalent command if they don't have python around.

warner:
Oh wait should that be `python3 -m http.server demos` ? The special directory is probably not 100% obvious.

kriskowal:
Posted an update with the Python demo script. The demos depend on serving the SES directory, to get to the built lockdown script, so `-d demos` is not necessary.

warner (approved):

packages/ses/package.json
```diff
@@ -48,7 +48,7 @@
     "build": "node scripts/bundle.js",
     "clean": "rm -rf dist",
     "cover": "c8 ava",
-    "demo": "http-server -o /demos",
+    "demo": "python3 -m http.server",
```
mhofman, line 51: This is unfortunate, we shouldn't force a python dependency just for a static server. There ought to be a simple enough node based static http server! Doesn't [serve](https://www.npmjs.com/package/serve) fit the bill?

Issue #506, opened 2020-10-27
Migrate Agoric-SDK packages (Endo-0)


kriskowal:
- [x] ~acorn-eventual-send~
  - [x] ~assert~
  - [x] bundle-source
  - [x] captp
  - [x] eventual-send
  - [x] far
  - [x] import-bundle
  - [x] ~import-manager~
  - [ ] ~install-metering-and-ses~
  - [x] install-ses
  - [x] marshal
  - [ ] notifier
  - [x] promise-kit
  - [ ] same-structure
  - [ ] store
  - [x] ~tame-metering~
  - [x] ~transform-eventual-send~
  - [x] ~transform-metering~
  - [ ] weak-store
  - [x] base64

Per https://docs.google.com/spreadsheets/d/1-gSPlAqFZYw9QHyYuCYubSNjyLN00JxtWcqT9hjEvBc/edit#gid=0

dckc:
The current alpha plan clearly entails moving `bundle-source`, `import-bundle`, and their dependencies. It's not clear to me that it requires moving `@agoric/capttp`, `@agoric/store` and such. Perhaps moving `bundle-source` and `import-bundle` is a subgoal worth splitting out?

p.s. how did you generate the dependency diagram, @kriskowal ?

dckc:
@erights - for Lavamoat - is splitting out the distributed computing stuff useful?

erights:
I don't think splitting it out is useful. Rather, it is postponable because their need for it is less urgent than for local endo.

kriskowal:
I plan to migrate the following from the leaves to the roots in short order, in preparation for adding my endo daemon prototype to this repository.

- import-bundle
  - bundle-source
  - install-ses
- bundle-source
  - lockdown
- captp
  - install-ses
  - eventual-send (not handled-promise-shim)
  - marshal
- far
  - eventual-send (for E only)
  - install-ses (for handled-promise-shim)
  - marshal
- install-ses
  - eventual-send (just handled-promise-shim)
  - lockdown
- marshal
  - eventual-send (shim for tests, E otherwise)
  - lockdown
- eventual-send
  - lockdown
- lockdown

kriskowal:
[...]

PR #982, opened 2022-01-12
chore: Migrate marshal and nat from Agoric SDK

kriskowal:
Refs #506


kriskowal:
Looks like `marshal` can’t import `nat` on Node.js 12. cc @michaelfig 

Obvious option is to drop Node.js 12 from CI. Less obvious is addressing the issue with Nat, which is certainly because of named exports support in Node.js itself.

michaelfig:
> Looks like `marshal` can’t import `nat` on Node.js 12. cc @michaelfig

`@agoric/nat` is a CommonJS module with multiple `exports.something = foo;` lines.  What tools do you have at your disposal for CommonJS interop?

kriskowal:
> > Looks like `marshal` can’t import `nat` on Node.js 12. cc @michaelfig
> 
> `@agoric/nat` is a CommonJS module with multiple `exports.something = foo;` lines. What tools do you have at your disposal for CommonJS interop?

Source is ESM and Nat is using Rollup. I think integration in Endo is the right way to go, giving us an opportunity to kick off Rollup and RESM.

michaelfig (approved):
LGTM!  Nice to adopt the `@endo/nat` straggler.

PR #988, opened 2022-01-15
docs: Security claims and caveats

kriskowal:
- fix: Broken link to security at top level
- docs(ses): Security claims and caveats

### DIFF
--- a/packages/ses/README.md
+++ b/packages/ses/README.md
@@ -28,10 +28,9 @@ See https://github.com/Agoric/Jessie to see how SES fits into the various
 flavors of confined JavaScript execution. And visit
 https://ses-demo.agoric.app/demos/ for a demo.
 
-Derived from the Caja project, https://github.com/google/caja/wiki/SES.
-
-Still under development: do not use for production systems yet, there are
-known security holes that need to be closed.
+SES starts where the Caja project left off
+https://github.com/google/caja/wiki/SES, and goes on to introduce compartments
+and modernize the permitted JavaScript features.
 
 ## Install
 
@@ -502,6 +501,195 @@ taming hides error stacks, accumulating them in side tables. The `assert`
 system generated other diagnostic information hidden in side tables. The tamed
 console uses these side tables to output more informative diagnostics.
 [Logging Errors](./src/error/README.md) explains the design.
+
+## Security claims and caveats
+
+The `ses` shim concerns boundaries between programs in the same process and
+JavaScript realm.
+In terms of the [Taxonomy of Security Issues](https://agoric.com/blog/all/taxonomy-of-security-issues/),
+the `ses` shim creates a boundary that is finer than an operating system
+process or thread and facilitates boundaries as fine as individual objects.
+While `ses` can interpose at granularities where process isolation is not a
+viable boundary, as between an application and its dependencies or between a
+platform and a plugin, `ses` combines well with coarser boundaries for defense
+in depth.
+
+For the purposes of these claims and caveats, a "host program" is a program
+that arranges `ses`, calls `lockdown`, and orchestrates one or more "guest
+programs", providing limited access to its resources.
+
+Provided that the `ses` implementation and its trusted compute base are
+correct, we claim that a host program (Figure 1) can evaluate a guest program
+(`program`) in a compartment after `lockdown` and that the program:
+
+- will initially only have access to one mutable object, the compartment's
+  `globalThis`,
+- specifically cannot modify any shared primordial objects, which are part of
+  the default execution environment,
+- cannot initially perform any I/O (except I/O necessarily performed by the
+  trusted compute base like paging virtual memory),
+- and specifically cannot measure the passage of time at any resolution.
+
+However, such a program can:
+
+- execute for an indefinite amount of time,
+- allocate arbitrary amounts of memory,
+- detect the platform endianness,
+- in some JavaScript engines, observe the contents of the stack.
+  This may include sensitive information about the layout of files on the host
+  disk.
+  In cases where the stack is data-dependent, a guest can infer the data.
+  `ses` occludes the stack on V8 and SpiderMonkey, but cannot on
+  JavaScriptCore.
+
+```js
+// Claims, Figure 1
+lockdown();
+const compartment = new Compartment();
+compartment.evaluate(program);
+```
+
+If the host program (Figure 2) arranges for the compartment's `globalThis` to
+be frozen, we additionally claim that the host can evaluate any two guest
+programs (`program1` and `program2`) such that neither program will:
+
+- initially share *any* mutable objects.
+- be able to observe the relative passage of time of the other program,
+  as they would had they been given a reference to a working `Date.now()`.
+- be able to communicate, as they would if they had shared access to mutable
+  state like an unfrozen object, a hardened collection like a `Map`, or even
+  `Math.random()`.
+
+```js
+// Claims, Figure 2
+lockdown();
+const compartment = new Compartment();
+harden(compartment.globaThis);
+compartment.evaluate(program1);
+compartment.evaluate(program2);
+```
+
+However such programs (`program`, `program1`, or `program2`) are only
+as useful as a calculator.
+A host program is therefore responsible for maintaining any of the desired
+invariants above when "endowing" a compartment with any of its own objects.
+
+For example, a host program (Figure 3) may run two programs in separate
+compartments, giving one program the ability to resolve a promise and the other
+program the ability to observe the settlement (fulfillment or rejection) of
+that promise.
+The host program is responsible for hardening these objects.
+
+```js
+// Claims, Figure 2
+lockdown();
+
+const promise = new Promise(resolve => {
+  const compartmentA = new Compartment(harden({
+    resolve,
+  }));
+  compartment.evaluate(programA);
+});
+
+const compartmentB = new Compartment(harden({
+  promise,
+}));
+compartmentB.evaluate(programB);
+```
+
+With `ses`, guest programs are initially powerless.
+A host can explicitly share limited powers with guest programs
+and provide intentional communication channels between them.
+
+Host programs must maintain the `ses` boundary with care in what they present
+as endowments.
+A host program should take care not to share mutable state with guests,
+or distribute mutable state to multiple guests, such as an unfrozen object (use
+`harden`), direct read and write access to a collection, like a `Map` or `Set`,
+even if hardened.
+Furthermore, typed arrays are collections and cannot be hardened.
+
+For the purposes of sharing state, pseudo-random number generators (PRNG) like
+`Math.random()` are equivalent to read and write access to shared state, and
+any guest can use one to eavesdrop on other guests or the host that share one.
+
+If a guest program needs a high resolution timer to function, the host should
+only invite one guest to a single operating system process and limit the
+activity of the host program in the same process.
+
+Hosts must avoid exposing `SharedArrayBuffer` to guests.
+Any two JavaScript programs sharing a `SharedArrayBuffer` can use the shared
+buffer to construct a high resolution timer.
+
+The `ses` shim does not in itself isolate the stack of guest programs, even
+when evaluated in separate compartments.
+This is relevant when program created objects are shared between guest
+programs.
+
+When a program interacts with an object introduced by another program (as
+through the per-compartment `globalThis`, function arguments or returned
+values), there are potential risks due to the synchronous nature of object
+access.
+Even interactions that are not explicit function calls may cause code from
+another program, like property accessors or proxy traps, to execute on the same
+stack, which may be able to call back into the program (reentrancy), throw, or
+detect the current stack height.
+
+A host object can defend itself from reentrancy attacks by ensuring that it
+interacts with guest objects on a clean stack through the use of promises.
+
+Within these constraints, a host program can provide objects that grant limited
+I/O capabilities to guest programs, and even revoke or suspend those
+capabilities at runtime.
+
+The trusted compute base (TCB) for `ses` includes:
+
+- the host hardware,
+- the host operating system,
+- any intermediate virtual operating systems or hypervisors,
+- the process memory manager,
+- an implementation of JavaScript conforming to ECMAScript 262 as of
+  2021, providing no unspecified embedding host behavior like the introduction of syntax
+  that when evaluated reveals a mutable object.
+  `ses` accounts for one such host behavior provided by Node.js, namely the `domain`
+  property on promises, by preventing the use of `ses` in concert with the
+  `domain` module.
+- Also, any attached debugger, and
+- any JavaScript that has executed in the same realm before the host program calls
+  `lockdown`, including JavaScript that executes after `ses` initializes.
+
+## Audits
+
+In June 2021, `ses` underwent formal third party vulnerability assessment over a
+period of 4 weeks with 3 engineers and a dedicated project manager that
+surfaced no unknown security issues or vulnerabilities within the code. As a
+result of this assessment, [a single code change was
+made](https://github.com/endojs/endo/issues/126) to set a flag to disable the
+domain module in Node.js to mitigate a known issue identified in the code.  The
+code will be the subject of another round of intense application security
+review mid-2022 by a reputable application security firm renowned for their
+results in security reviews.
+
+In July 2021, `ses` was the target of an intensive collaborative bug hunt lead by
+the MetaMask team.
+No critical flaws in the code surfaced during the review.
+As a result of the search for flaws, deficiencies, and weaknesses in the code
+(which is currently a Stage 1 proposal with the ECMA TC39 committee), a series
+of small code changes and documentation improvements were made. There is a
+report available on the
+[Agoric blog](https://agoric.com/blog/technology/metamask-agoric-hardened-js-security-review/)
+that includes links to recordings of code walk-throughs and technical
+discussion, and issues are tagged
+[audit-SEStival](https://github.com/endojs/endo/labels/audit-sestival). 
+
+In addition to vulnerability assessments, active efforts to [formally verify
+the Agoric kernel](https://agoric.com/blog/technology/the-path-to-verified-blds-how-informal-systems-and-agoric-are-using-formal-methods-analysis-to-improve-software-integrity/)
+have found the object capability model that `ses` provides to be sound.
+
+Hardened JavaScript is also within the scope of the [Agoric bug bounty
+program](hackerone.com/agoric), which rewards researchers for surfacing valid
+bugs in our code. We welcome the opportunity to cooperate with researchers,
+whose efforts will undoubtedly yield stronger, more resilient code. 
 
 ## Bug Disclosure
 
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ pronunciations of `sudo`, "en-doh" and "en-doo" are both valid pronunciations of
 ### Bug Disclosure
 
 Please help us practice coordinated security bug disclosure, by using the
-instructions in our [security guide](./SECURITY.md) to report
+instructions in our [security guide](./packages/ses/SECURITY.md) to report
 security-sensitive bugs privately.
 
 For non-security bugs, please use the [regular Issues


kumavis (approved):
👍😁👍

packages/ses/README.md
```diff
@@ -503,6 +502,153 @@ system generated other diagnostic information hidden in side tables. The tamed
...
+  compartment.evaluate(programA);
+});
+
+const compartmentB = new Compartment(harden({
+  promise,
+}));
+compartmentB.evaluate(programB);
+```
+
+Any agent can also hide their own stack by causing client callbacks to
```
mhofman, line 600: Why are we switching the terminology to "agent" ?

packages/ses/README.md
```diff
@@ -503,6 +502,153 @@ system generated other diagnostic information hidden in side tables. The tamed
 console uses these side tables to output more informative diagnostics.
 [Logging Errors](./src/error/README.md) explains the design.
 
+## Security claims and caveats
+
+Provided that the `ses` implementation and its trusted compute base are
+correct, we claim that a host program (Figure 1) can evaluate a guest program
```
erights, line 508: Given the tc39 use of the term "host", if you want to use that here, you need to explain what you mean an do not mean early. Indeed, this text almost adequately explains it by example. But it doesn't impede the misunderstanding.

packages/ses/README.md
```diff
@@ -503,6 +502,153 @@ system generated other diagnostic information hidden in side tables. The tamed
...
+});
+
+const compartmentB = new Compartment(harden({
+  promise,
+}));
+compartmentB.evaluate(programB);
+```
+
+Any agent can also hide their own stack by causing client callbacks to
+run in a subsequent microtask.
```
mhofman, line 601: microtask is a little obscure, maybe clarify that's done through promises?

packages/ses/README.md
```diff
@@ -503,6 +502,153 @@ system generated other diagnostic information hidden in side tables. The tamed
...
 [Logging Errors](./src/error/README.md) explains the design.
 
+## Security claims and caveats
+
+Provided that the `ses` implementation and its trusted compute base are
+correct, we claim that a host program (Figure 1) can evaluate a guest program
+(`program`) in a compartment after `lockdown` and that the program:
+
+- will initially only have access to one mutable object, `globalThis`,
+- specifically cannot modify any primordial object that it shares with the host
```
erights, line 512:
```suggestion
- specifically cannot modify any primordial object that it shares with the host, except that compartment's `globalThis`, which the host has access to.
```

packages/ses/README.md
```diff
@@ -503,6 +502,153 @@ system generated other diagnostic information hidden in side tables. The tamed
...
+However, such a program can:
+
+- execute for an indefinite amount of time,
+- allocate arbitrary amounts of memory,
+- detect the platform endianness,
+- in some JavaScript engines, observe the contents of the stack.
+  This may include sensitive information about the layout of files on the host
+  disk.
+  In cases where the stack is data-dependent, a guest can infer the data.
+  SES occludes the stack on V8 and SpiderMonkey, but can not on
```
erights, line 527:
```suggestion
  SES occludes the stack on V8 and SpiderMonkey, but cannot on
```

packages/ses/README.md
mhofman, line 601: To be pedantic, the promise helps more in isolating the stack than hiding its info. Error taming is what hides the stack, and you could very well imagine that a JS engine exposes all sorts of async stack info through error stacks without taming.

Where isolation comes in is that an error thrown by `programB` will not bubble into and interrupt `programA`.
Also something something about stack overflows and promise resolution protecting `programB` from `programA` abuses.

packages/ses/README.md
```diff
@@ -503,6 +502,153 @@ system generated other diagnostic information hidden in side tables. The tamed
...
+const compartment = new Compartment();
+compartment.evaluate(program);
+```
+
+The trusted compute base includes:
+
+- the host hardware,
+- the host operating system,
+- the process memory manager,
+- a minimal implementation of JavaScript conforming to ECMAScript 262 as of
```
erights, line 542:
```suggestion
- a implementation of JavaScript conforming to ECMAScript 262 as of
```

Your remaining text says more clearly what the restriction we the spec actually is. Given that, I'd remove "minimal" here as that suggests a tighter restriction than we mean.

packages/ses/README.md
```diff
@@ -503,6 +502,153 @@ system generated other diagnostic information hidden in side tables. The tamed
...
+Any agent can also hide their own stack by causing client callbacks to
+run in a subsequent microtask.
+
+Among the ways a host program can invalidate these claims, a host can provide
+
+- either directly or by introduction through the transitive closure over
+  properties, property descriptors, arguments or returned values of functions
+  or methods:
+  - any shared mutable object,
+  - any resolution timer,
```
mhofman, line 609: Or feature performing I/O

packages/ses/README.md
```diff
@@ -503,6 +502,153 @@ system generated other diagnostic information hidden in side tables. The tamed
...
+  - any undesired communication channel, for example a PRNG,
+  - any `SharedArrayBuffer` that is indeed shared,
```
mhofman, line 611: To be pedantic, a regular `ArrayBuffer` would be sufficient in this case

packages/ses/README.md
```diff
@@ -503,6 +502,153 @@ system generated other diagnostic information hidden in side tables. The tamed
...
+- any JavaScript that has executed in the same realm before the host calls
+  `lockdown`, including JavaScript that executes after `ses` initializes.
+
+If the host program (Figure 2) arranges for the compartment's `globalThis` to
+be frozen, we additionally claim that the host can evaluate any two programs
+(`program1` and `program2`) such that neither program will:
+
+- initially share *any* mutable objects.
+- be able to observe the relative passage of time of the other program,
+- specifically because they do not have access to any shared mutable state as
```
erights, line 558: The phrasing of this bullet point is confusing. Either clarify or drop it. I think it's ok to drop it.

packages/ses/README.md
```diff
@@ -503,6 +502,153 @@ system generated other diagnostic information hidden in side tables. The tamed
...
+Among the ways a host program can invalidate these claims, a host can provide
+
+- either directly or by introduction through the transitive closure over
+  properties, property descriptors, arguments or returned values of functions
+  or methods:
+  - any shared mutable object,
+  - any resolution timer,
+  - any undesired communication channel, for example a PRNG,
+  - any `SharedArrayBuffer` that is indeed shared,
+  - the `Atomics` namespace object.
```
mhofman, line 612: Actually atomics is pretty useless without `SharedArrayBuffer`

packages/ses/README.md
```diff
@@ -503,6 +502,153 @@ system generated other diagnostic information hidden in side tables. The tamed
...
+- any function from from one agent that calls a callback from another agent on
```
mhofman, line 613: Agent terminology again here.

mhofman, line 613: I usually prefer to use "invoke" with callback to lessen the "call" visual overload.

packages/ses/README.md
```diff
@@ -503,6 +502,153 @@ system generated other diagnostic information hidden in side tables. The tamed
...
+// Claims, Figure 2
+lockdown();
+const compartment = new Compartment();
+harden(compartment.globaThis);
+compartment.evaluate(program1);
+compartment.evaluate(program2);
+```
+
+However no such programs (`program`, `program1`, nor `program2`) can be useful
+as such.
```
erights, line 571: Nice! But not quite true. The host has access to the second one (completion value!), and it could be a useful pure function.

packages/ses/README.md
```diff
@@ -503,6 +502,153 @@ system generated other diagnostic information hidden in side tables. The tamed
...
+## Audits
+
+In June 2021, SES underwent formal third party vulnerability assessment over a
+period of 4 weeks with 3 engineers and a dedicated project manager that
+surfaced no unknown security issues or vulnerabilities within the code. As a
+result of this assessment, [a single code change was
+made](https://github.com/endojs/endo/issues/126)to set a flag to disable the
+domain module in node.js to mitigate a known issue identified in the code.  The
+code will be the subject of another round of intense application security
+review mid-2022 by a reputable application security form reknowned for their
```
mhofman, line 628:
```suggestion
review mid-2022 by a reputable application security firm renowned for their
```

packages/ses/README.md
```diff
@@ -503,6 +502,153 @@ system generated other diagnostic information hidden in side tables. The tamed
...
+```
+
+However no such programs (`program`, `program1`, nor `program2`) can be useful
+as such.
+A host program is therefore responsible for maintaining any of the desired
+invariants above when "endowing" a compartment with any of its own objects.
+
+For example, a host program (Figure 3) may run two programs in separate
+compartments, giving one program the ability to resolve a promise and the other
+program the ability to observe the resolution.
```
erights, line 577:
```suggestion
compartments, giving one program the ability to resolve a particular promise and the other
program the ability to observe the settlement (fulfillment or rejection) of that promise.
```
resolution by forwarding to another promise is not observable.

mhofman (changes requested):

packages/ses/README.md
```diff
@@ -503,6 +502,153 @@ system generated other diagnostic information hidden in side tables. The tamed
...
+  - any shared mutable object,
+  - any resolution timer,
+  - any undesired communication channel, for example a PRNG,
+  - any `SharedArrayBuffer` that is indeed shared,
+  - the `Atomics` namespace object.
+- any function from from one agent that calls a callback from another agent on
+  its own stack (reentrancy hazards), including property descriptor functions
+  and proxy traps.
+
+These responsibilities for hosts are not exhaustive.
```
erights, line 617: I don't like the phrasing of this entire section. It sounds like we're saying these are all bad things for a host to do. Rather, I'd phrase this all in terms of enabling the host to express security policy by explicitly granting access to objects. In order to maximize expressiveness (among other things) the evaluate code starts out powerless, i.e., without any powers the host did not decide to grant.

packages/ses/README.md
```diff
@@ -503,6 +502,153 @@ system generated other diagnostic information hidden in side tables. The tamed
...
+
+## Audits
+
+In June 2021, SES underwent formal third party vulnerability assessment over a
+period of 4 weeks with 3 engineers and a dedicated project manager that
+surfaced no unknown security issues or vulnerabilities within the code. As a
+result of this assessment, [a single code change was
+made](https://github.com/endojs/endo/issues/126)to set a flag to disable the
```
erights, line 625:
```suggestion
made](https://github.com/endojs/endo/issues/126) to set a flag to disable the
```

erights (approved):
LGTM

packages/ses/README.md
mhofman: Yeah I'm conflicted. While it's good to make the distinction with guest programs below, the host terminology is a bit overloaded.

packages/ses/README.md
kriskowal: Hard to find a word that isn’t overloaded, neither agent is necessarily the host, both might be guests connected by introduction. Maybe I can work a clarification in.

packages/ses/README.md
kriskowal: I mean for the purposes of leaking a high resolution timer. In the sense that an `ArrayBuffer` is a communication channel, so is a `Map`. I can clarify.

mhofman: SAB is only a high resolution timer when combined with workers / threads

erights: SAB + Atomics is a timer though.

packages/ses/README.md
```diff
@@ -503,6 +502,153 @@ system generated other diagnostic information hidden in side tables. The tamed
...
+
+Provided that the `ses` implementation and its trusted compute base are
+correct, we claim that a host program (Figure 1) can evaluate a guest program
+(`program`) in a compartment after `lockdown` and that the program:
+
+- will initially only have access to one mutable object, `globalThis`,
+- specifically cannot modify any primordial object that it shares with the host
+  program
+- cannot perform any I/O, apart from I/O performed within the trusted compute
+  base, directly or indirectly,
```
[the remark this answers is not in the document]
kriskowal: Rephrasing this like (apart from I/O performed by the trusted compute base like paging virtual memory).

packages/ses/README.md
```diff
@@ -503,6 +502,153 @@ system generated other diagnostic information hidden in side tables. The tamed
 console uses these side tables to output more informative diagnostics.
 [Logging Errors](./src/error/README.md) explains the design.
 
+## Security claims and caveats
```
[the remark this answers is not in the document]
kriskowal: All claims should be for the `ses` shim so I’m revising to attempt to signal that typographically.

packages/ses/README.md
```diff
@@ -503,6 +502,153 @@ system generated other diagnostic information hidden in side tables. The tamed
...
+  `lockdown`, including JavaScript that executes after `ses` initializes.
+
+If the host program (Figure 2) arranges for the compartment's `globalThis` to
+be frozen, we additionally claim that the host can evaluate any two programs
+(`program1` and `program2`) such that neither program will:
+
+- initially share *any* mutable objects.
+- be able to observe the relative passage of time of the other program,
+- specifically because they do not have access to any shared mutable state as
+  they would had they been given a reference to `Math.random()`.
```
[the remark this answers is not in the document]
kriskowal: Clarifying elsewhere how `Math.random()` is shared state. Adding `Date.now()` to bullet above about timers.

packages/ses/README.md
erights: subsystem

packages/ses/README.md
erights: microtask isn't even a JS concept. It's a browser concept.

packages/ses/README.md
```diff
@@ -503,6 +502,153 @@ system generated other diagnostic information hidden in side tables. The tamed
...
+A host program is therefore responsible for maintaining any of the desired
+invariants above when "endowing" a compartment with any of its own objects.
+
+For example, a host program (Figure 3) may run two programs in separate
+compartments, giving one program the ability to resolve a promise and the other
+program the ability to observe the resolution.
+The host program is responsible for hardening these objects.
+
+With such an arrangement, the host and both guest programs are vulnerable
+to denial of service.
```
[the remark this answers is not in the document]
kriskowal: I’m assuming that no arrangement would have been made previously where host and guest are in the same JavaScript realm.

packages/ses/README.md
kriskowal: I seem to have found a way to use “any other guest or the host” instead of “agent”.

packages/ses/README.md
kriskowal: Adding: promise events.

kriskowal: I’ve adopted text you’ve suggested out-of-band.

packages/ses/README.md
kriskowal: Reworked this section entirely.

packages/ses/README.md
mhofman: Only if the agent can block. But yeah I see your point.

packages/ses/README.md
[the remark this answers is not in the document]
mhofman: I meant in the previous example. That's what I had understood.

erights:
Some of https://agoric.com/blog/all/taxonomy-of-security-issues/ may be helpful. If nothing else, you could use some its terminology and link to the doc for more detailed definitions.

packages/ses/README.md
mhofman: Promise reactions?

packages/ses/README.md
kriskowal: Adding a terminology sentence above.

packages/ses/README.md
kriskowal: I’m reading `globalThis` as not a shared primordial, for not being shared between compartments and also not primordial since it’s compartment-specific. I think the prior bullet makes that clear.

packages/ses/README.md
kriskowal: Posting a rewrite shortly.

packages/ses/README.md
```diff
@@ -504,16 +504,31 @@ console uses these side tables to output more informative diagnostics.
...
 
+The `ses` shim concerns boundaries between programs in the same process and
+JavaScript realm.
+In terms of the [Taxonomy of Security Issues](https://agoric.com/blog/all/taxonomy-of-security-issues/),
+the `ses` shim creates a boundary that is finer than an operating system
+process or thread and facilitates boundaries as fine as individual objects.
+While `ses` can interpose at granularities where process isolation is not a
+viable boundary, as between an application and its dependencies or between a
+platform and a plugin, `ses` combines well with coarser boundaries for defense
+in depth.
```
mhofman, line 515: Technically it's SES in this paragraph, right?

packages/ses/README.md
```diff
@@ -567,19 +569,17 @@ compartment.evaluate(program1);
 compartment.evaluate(program2);
 ```
 
-However no such programs (`program`, `program1`, nor `program2`) can be useful
-as such.
+However such programs (`program`, `program1`, nor `program2`) are only
```
mhofman, line 572:
```suggestion
However such programs (`program`, `program1`, or `program2`) are only
```

packages/ses/README.md
```diff
@@ -597,54 +597,95 @@ const compartmentB = new Compartment(harden({
...
+buffer to construct a high resolution timer.
+
+The `ses` shim does not in itself isolate the stack of guest programs, even
+when evaluated in separate compartment.
+This is relevant when program created objects are shared between guest
+programs.
+
+Any object endowed to a program (either through the per-compartment
+`globalThis`, function arguments or returned values) represents a potential
+risk, such as reentrancy, when it interacts with a guest object.
```
mhofman, line 631: This may need some wordsmithing. It's not clear what `it` references to here. It should be program but the original subject is the endowed object.

packages/ses/README.md
```diff
@@ -597,54 +597,95 @@ const compartmentB = new Compartment(harden({
...
+Even interactions that are not explicit function calls may cause code from
+another program, like property accessors or proxy traps, to execute on the same
+stack, which may be able to call back into the program (reentrancy), throw, or
+detect the current stack height.
+
+A host object can defend itself from reentrancy attacks by ensuring that it
+interacts with guest objects in fresh events, as in microtasks: promise events.
```
mhofman, line 638: I'm not a major fan of the event terminology without a more obvious reference to an event-loop. The ecma262 spec talk of jobs, which is probably even less obvious. Wondering if we can simply say something like "interacts with guest objects on a clean stack through the use of promises".

packages/ses/README.md
```diff
@@ -597,54 +597,95 @@ const compartmentB = new Compartment(harden({
...
+- any intermediate virtual operating systems or hypervisors,
+- the process memory manager,
+- an implementation of JavaScript conforming to ECMAScript 262 as of
+  2021, providing no unspecified host behavior like the introduction of syntax
+  that when evaluated reveals a mutable object.
+  `ses` accounts for one such host behavior provided by Node.js, namely the `domain`
+  property on promises, by preventing the use of `ses` in concert with the
+  `domain` module.
+- Also, any attached debugger, and
+- any JavaScript that has executed in the same realm before the host calls
```
mhofman, line 657:
```suggestion
- any JavaScript that has executed in the same realm before the host program calls
```

packages/ses/README.md
```diff
@@ -597,54 +597,95 @@ const compartmentB = new Compartment(harden({
...
+capabilities at runtime.
+
+The trusted compute base (TCB) for `ses` includes:
+
+- the host hardware,
+- the host operating system,
+- any intermediate virtual operating systems or hypervisors,
+- the process memory manager,
+- an implementation of JavaScript conforming to ECMAScript 262 as of
+  2021, providing no unspecified host behavior like the introduction of syntax
```
mhofman, line 651:
```suggestion
  2021, providing no unspecified embedding host behavior like the introduction of syntax
```

packages/ses/README.md
```diff
@@ -597,54 +597,95 @@ const compartmentB = new Compartment(harden({
...
 period of 4 weeks with 3 engineers and a dedicated project manager that
 surfaced no unknown security issues or vulnerabilities within the code. As a
 result of this assessment, [a single code change was
-made](https://github.com/endojs/endo/issues/126)to set a flag to disable the
-domain module in node.js to mitigate a known issue identified in the code.  The
+made](https://github.com/endojs/endo/issues/126) to set a flag to disable the
+domain module in Node.js to mitigate a known issue identified in the code.  The
 code will be the subject of another round of intense application security
-review mid-2022 by a reputable application security form reknowned for their
+review mid-2022 by a reputable application security form renown for their
```
mhofman, line 669:
```suggestion
review mid-2022 by a reputable application security firm renowned for their
```

mhofman (commented):
A few nits on a first pass. Will redo another full pass.

packages/ses/README.md
```diff
@@ -503,6 +502,194 @@ system generated other diagnostic information hidden in side tables. The tamed
...
+If a guest program needs a high resolution timer to function, the host should
+only invite one guest to a single operating system process and limit the
+activity of the host program in the same process.
+
+Hosts must avoid exposing `SharedArrayBuffer` to guests.
+Any two JavaScript programs sharing a `SharedArrayBuffer` can use the shared
+buffer to construct a high resolution timer.
+
+The `ses` shim does not in itself isolate the stack of guest programs, even
+when evaluated in separate compartment.
```
mhofman, line 625:
```suggestion
when evaluated in separate compartments.
```

packages/ses/README.md
kriskowal: For the purposes of the claims, I’m limiting the scope to this implementation specifically. Elsewhere, we’d refer to Hardened JavaScript. The SES Proposal is too aspirational to be useful here.

packages/ses/README.md
```diff
@@ -503,6 +502,194 @@ system generated other diagnostic information hidden in side tables. The tamed
...
+This is relevant when program created objects are shared between guest
+programs.
+
+Any object endowed to a program (either through the per-compartment
+`globalThis`, function arguments or returned values) represents a potential
+risk, such as reentrancy, when it interacts with a guest object.
```
mhofman:
```suggestion
When a program interacts with an object introduced by another program
(e.g. through the per-compartment `globalThis`, function arguments or
returned values), there are potential risks due to the synchronous nature
of object access.
```

kriskowal: Adopted, but avoided Latin `e.g.,`.

mhofman: We lost the second sentence which I believe is still valuable.

kriskowal: Oh, my mistake again. Pushed.

mhofman (approved):
🎉
Since there's been quite a few changes since the initial version, wondering if @erights wants to give it another look.

PR #990, opened 2022-01-16
fix(ses): Fix mistaken this binding example

kriskowal:
The example in question claims that the per-compartment `Function` constructor arranges for `this` to be bound to its particular `globalThis`. However, because compartments only reveal strict mode functions, `this` is `undefined`. So, to support the claim documented above the example, we amend the example to grab `globalThis` from the scope of the function body.


erights (approved):
LGTM

PR #992, opened 2022-01-18
feat(endo): Initial daemon

kriskowal:
This is a starting point for the Endo daemon.

The Endo daemon is a per-user background process that communicates with its
user over a Unix domain socket (eventually also a named pipe on Windows).

This change includes three new packages:
- where: Identifies the user-specific location of runtime files for Endo.
  Having a separate package for this allows multiple packages to rendezvous,
  which will eventually include demo packages.
- daemon: The daemon process and controller API.
- cli: A command line for the controller API.

The factoring allows for the daemon to be eventually be operated from an Endo
Browser application, instead of or in addition to the CLI.

As of this change, the Endo daemon has no interesting API, except a private
(user only) method to shut down the daemon.
The daemon will eventually be able to manage child worker processes,
some of which will persist and restart when Endo restarts.

The Endo UI will be implemented in terms of the daemon's private facet,
so this is where the browser or CLI will find methods for observing and
responding to requests from Endo applications.

### DIFF
--- a/packages/daemon/test/test-endo.js
+++ b/packages/daemon/test/test-endo.js
@@ -1,0 +1,28 @@
+// @ts-check
+/* global process */
+import test from 'ava';
+import url from 'url';
+import path from 'path';
+import { start, stop, restart, clean } from '../index.js';
+
+const { raw } = String;
+
+const dirname = url.fileURLToPath(new URL('.', import.meta.url)).toString();
+
+const locator = {
+  endoPath: path.join(dirname, 'endo'),
+  sockPath:
+    process.platform === 'win32'
+      ? raw`\\?\pipe\endo-test.sock`
+      : path.join(dirname, 'endo.sock'),
+  logPath: path.join(dirname, 'endo.log'),
+};
+
+test.serial('lifecycle', async t => {
+  await clean(locator);
+  await start(locator);
+  await stop(locator);
+  await restart(locator);
+  await stop(locator);
+  t.pass();
+});
--- a/packages/where/index.js
+++ b/packages/where/index.js
@@ -1,0 +1,90 @@
+// @ts-check
+
+/* Infers the rendezvous path for the endo.sock file and apps from the platform
+ * and environment.
+ */
+
+const { raw } = String;
+
+/**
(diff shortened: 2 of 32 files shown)


kriskowal:
Draft until tests.

packages/daemon/scratch.bak
```diff
@@ -0,0 +1,68 @@
+// Well-known-objects
```
michaelfig, line 1: Is this supposed to be in the PR?

packages/daemon/test/test-endo.js
```diff
@@ -0,0 +1,26 @@
...
+import path from 'path';
+import { start, stop, restart, clean } from '../index.js';
+
+const dirname = url.fileURLToPath(new URL('.', import.meta.url)).toString();
+
+const locator = {
+  endoPath: path.join(dirname, 'endo'),
+  sockPath:
+    process.platform === 'win32'
+      ? '\\\\?\\pipe\\endo-test.sock'
```
michaelfig, line 14:
```suggestion
      ? String.raw`\\?\pipe\endo-test.sock`
```

packages/where/index.js
```diff
@@ -0,0 +1,88 @@
...
+  // Favoring local app data over roaming app data since I don't expect to be
+  // able to listen on one host and connect on another.
+  if (env.LOCALAPPDATA !== undefined) {
+    return `${env.LOCALAPPDATA}\\Endo`;
+  }
+  if (env.APPDATA !== undefined) {
+    return `${env.APPDATA}\\Endo`;
+  }
+  if (env.USERPROFILE !== undefined) {
+    return `${env.USERPROFILE}\\AppData\\Endo`;
```
michaelfig, line 20: If you like the `String.raw` suggestion, you could apply it here and below.
```suggestion
    return String.raw`${env.USERPROFILE}\AppData\Endo`;
```

michaelfig (approved):
LGTM!

I'm not sure if you like the `String.raw` suggestions.  If so, maybe destructure with `const { raw } = String;` and use just ``raw`idet\aith\aid\` ``

PR #993, opened 2022-01-18
chore(bundle-source): Migrate from Agoric SDK

kriskowal:
Refs #506


michaelfig (approved):
LGTM!  Feel free to merge after CI passes.

kriskowal:
@michaelfig This looks like another case where we’re not able to use named exports from CommonJS on Node.js 12. I think this time, we have no option but to push forward into Node.js 14 and beyond.

PR #994, opened 2022-01-20
marshal's Hilbert processing should only freeze rest, not harden it.

erights:
Fixes https://github.com/Agoric/agoric-sdk/issues/4313


kriskowal (changes requested):
Looks good to me. Please keep off `master` until a release commit appears before ff68d336ddd963c681431fbd6d1d014526593842 in the log in chronological order. I will lift “request changes” at that time.

mhofman (approved):

kriskowal (approved):
`master` is released and clear for changes.

PR #996, opened 2022-01-20
Assert that unserialize makes only passables

erights:
Fixes https://github.com/Agoric/agoric-sdk/issues/4337.

Fixes https://github.com/Agoric/agoric-sdk/issues/4302 since thenables are not passable.

### DIFF
--- a/packages/marshal/src/marshal.js
+++ b/packages/marshal/src/marshal.js
@@ -4,7 +4,7 @@
 /// <reference types="ses"/>
 
 import { Nat } from '@endo/nat';
-import { passStyleOf } from './passStyleOf.js';
+import { assertPassable, passStyleOf } from './passStyleOf.js';
 
 import './types.js';
 import { getInterfaceOf } from './helpers/remotable.js';
@@ -483,7 +483,10 @@ export function makeMarshal(
     );
     const rawTree = harden(JSON.parse(data.body));
     const fullRevive = makeFullRevive(data.slots);
-    return harden(fullRevive(rawTree));
+    const result = harden(fullRevive(rawTree));
+    // See https://github.com/Agoric/agoric-sdk/issues/4337
+    assertPassable(result);
+    return result;
   };
 
   return harden({


kriskowal (changes requested):
This looks good to me. Please keep off master until `master` contains a release commit after ff68d336ddd963c681431fbd6d1d014526593842. I’ll lift my “request changes” block at that time.

packages/marshal/src/marshal.js
```diff
@@ -484,6 +484,7 @@ export function makeMarshal(
     const rawTree = harden(JSON.parse(data.body));
     const fullRevive = makeFullRevive(data.slots);
     const result = harden(fullRevive(rawTree));
+    // See https://github.com/Agoric/agoric-sdk/issues/4302
```
mhofman, line 487: I'm actually still writing a more general issue for this.

mhofman: https://github.com/Agoric/agoric-sdk/issues/4337

erights: Done. Linked to https://github.com/Agoric/agoric-sdk/issues/4337 . But this PR only solves the specific #4302 problem.

mhofman: How so? Doesn't it prevent all things that produce non passable, like buggy convertSlotToVal ?

mhofman (approved):

packages/marshal/src/marshal.js
erights: ***OOPS***. I just merged https://github.com/Agoric/agoric-sdk/pull/4304 which contained changes to two files in the `marshal` package of agoric-sdk. Thus, I made them out of sync.

@kriskowal How should we handle this?

erights: I noticed this in preparing to answer @mhofman 's question about #4337

In answer to that question, yes, the assert will catch that specific problem. But mysteriously. I'd like to put more checking directly around the convertSlotToVal call so I can complain with a more specific diagnostic. That same check can ensure that the result is only remotables and promises, which this PR does not enforce.

mhofman: Oh yes! Which is why I also wrote https://github.com/Agoric/agoric-sdk/issues/4336 for convertSlotToVal

https://github.com/Agoric/agoric-sdk/issues/4337 is the meta issue listing all known occurrences of non passables that this PR prevents.

kriskowal (approved):
`master` is released and clear for changes.

PR #999, opened 2022-01-20
chore: Synchronize internal versions

kriskowal:
This change adds some scripts we can use to keep dependency versions in sync
across packages, then applies them to our current setup.
The get, set, and sync scripts are borrowed from Agoric SDK.
The all script is new.

- chore: Add version synchronization scripts
- chore: Synchronize versions
- chore: Update yarn.lock


michaelfig (approved):
LGTM!

PR #1002, opened 2022-01-22
feat(ses): Detect corrupted feral eval early

### LINKED ISSUE #343: lockdown() should bail if it cannot get a direct eval
Agoric/agoric-sdk#1182 fixed a problem in the metering transform which broke the direct `eval` that SES-shim needs to build the proxy-scoped evaluator (turning it into an indirect eval). The symptom was subtle: added endowments went missing, but evaluation worked fine otherwise.

Should we add code to `lockdown()` to assert that it has access to a functional direct eval, and abort if it does not? The SES shim is, of course, entirely dependent upon the environment that loads it, but if a vetted shim were to unintentionally break direct eval, it might be better to halt than to produce a subtly-broken SES environment.

_Originally posted by @warner in https://github.com/Agoric/agoric-sdk/pull/1182#issuecomment-644253927_

kriskowal:
- test(ses): Test dynamic eval detection
- feat(ses): Check early for dynamic evalability

Fixes #343

### DIFF
--- a/packages/ses/src/lockdown-shim.js
+++ b/packages/ses/src/lockdown-shim.js
@@ -15,6 +15,8 @@
 // @ts-check
 
 import {
+  FERAL_FUNCTION,
+  FERAL_EVAL,
   TypeError,
   arrayFilter,
   arrayMap,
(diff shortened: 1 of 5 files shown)


kriskowal:
cc @warner FYI

packages/ses/src/lockdown-shim.js
```diff
@@ -87,6 +89,27 @@ const harden = makeHardener();
...
+  );
+};
+
+const assertDirectEvalAvailable = () => {
+  try {
+    const changed = FERAL_FUNCTION(
+      'eval',
+      'changed',
+      'eval("changed = true"); return changed',
+    )(FERAL_EVAL, false);
```
erights, line 104: Awesome test!

erights (approved):
LGTM!

kriskowal:
I’ve added an error description doc and updated the NEWS. Will land when CI is green.

PR #1004, opened 2022-01-22
fix(ses): Direct eval check should not preclude no-eval under CSP

kriskowal:
This change adds nuance to the code that checks for whether eval is
    direct-eval. We do want to screen environments where eval is indirect.
    But, if eval throws an exception instead, that would indicate a
    content-security-policy that excludes unsafe-eval, and SES should be
    usable for this case in a future version.

    Currently, this case fails for other reasons.
    So this change adds a known-failing test.

### DIFF
--- a/packages/ses/src/lockdown-shim.js
+++ b/packages/ses/src/lockdown-shim.js
@@ -89,24 +89,34 @@ const harden = makeHardener();
 // only ever need to be called once and that simplifying lockdown will improve
 // the quality of audits.
 
-const directEvalUnavailable = () => {
-  throw new TypeError(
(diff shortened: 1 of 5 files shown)


packages/ses/src/lockdown-shim.js
```diff
@@ -89,24 +89,31 @@ const harden = makeHardener();
...
 const assertDirectEvalAvailable = () => {
+  let allowed = false;
   try {
-    const changed = FERAL_FUNCTION(
+    allowed = FERAL_FUNCTION(
       'eval',
-      'changed',
-      'eval("changed = true"); return changed',
+      'SES_changed',
+      `eval("SES_changed = true"); return SES_changed`,
```
erights, line 98:
```suggestion
      `\
eval("SES_changed = true");
return SES_changed;`,
```

erights (approved):
LGTM

PR #1007, opened 2022-01-22
chore: release

kriskowal:
This release includes many new packages. Most are migrations from Agoric SDK.  The remaining packages include initial versions of Endo daemon, its locator, and its CLI, as well as stream libraries needed for Endo Browser and native message host.

The netstring library includes a principled breaking change, which should require no additional programming to integrate in Agoric SDK: all stream libraries now require eventual-send in addition to hardened JavaScript.


michaelfig (approved):
LGTM!

PR #1010, opened 2022-01-23
Retore edits from agoric-sdk #4304

erights:
Restore edits from https://github.com/Agoric/agoric-sdk/pull/4304 to the marshal package.

Fixes https://github.com/Agoric/agoric-sdk/issues/4312 again ;)

Have recognition of passable Error obtain name from prototype, skipping any own name property on the error instance.

Note several of the places we need to revisit when we admit error.cause and error.errors.

Inline TODOs reminding us to revisit issues identified by zestival


kriskowal (approved):
Commit message should be `fix(marshal): ...` and be sure to squash merge to preserve the commit message, not the PR title.

PR #1012, opened 2022-01-24
feat: HandledPromise hardening

michaelfig:
refs: Agoric/agoric-sdk#4362
closes: Agoric/agoric-sdk#4363

## Description

Move hardening around HandledPromises to the shim so that it can't be bypassed with a malicious user-level `E`.

The motivation in Agoric/agoric-sdk#4362 still applies; when we're under SES, we should harden on the boundaries of the system.  HandledPromise's eventual send and eventual get support is such a thing.

### DIFF
--- a/packages/eventual-send/src/handled-promise.js
+++ b/packages/eventual-send/src/handled-promise.js
@@ -63,6 +63,7 @@ const coerceToObjectProperty = specimen => {
  */
 export const makeHandledPromise = () => {
   const presenceToHandler = new WeakMap();
+  /** @type {WeakMap<any, any>} */
   const presenceToPromise = new WeakMap();
   const promiseToPendingHandler = new WeakMap();
   const promiseToPresence = new WeakMap();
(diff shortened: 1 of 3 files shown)


kriskowal (approved):

erights (approved):
LGTM -- Very nice!

packages/eventual-send/src/handled-promise.js
```diff
@@ -477,7 +478,11 @@ export const makeHandledPromise = () => {
     applyMethodSendOnly: makeForwarder('applyMethodSendOnly', localApplyMethod),
   };
 
-  handle = (p, operation, opArgs, ...dispatchArgs) => {
+  handle = (...handleArgs) => {
+    // We're in SES mode, so we should harden.
+    harden(handleArgs);
```
mhofman, line 483: I thought we wanted to avoid implicitly hardening values, and consider it an error if the argument wasn't hardened already?

erights: @michaelfig and I had a long discussion in which he convinced me of the opposite. @michaelfig captured the rationale at https://github.com/Agoric/agoric-sdk/issues/4362

We should discuss.

PR #1013, opened 2022-01-25
fix(promise-kit): Publish code

kriskowal:
The current published version does not include the code (index.js).
This change ensures it will get picked up next time.


michaelfig (approved):
Oops.

PR #1016, opened 2022-01-25
refactor: Consolidate Endo CLI

kriskowal:
The previous Endo prototype relied on the assumption that Endo might be something like RESM (`node -r esm`) with a toolkit for archive management.  The archive management is probably useful but the executor is a dead end since it has no powerbox mechanism.  The new Endo CLI needs address this in a different way.

This change preserves the archive management commands by moving them to the new `cli` and then removes the old `endo` package.

- refactor: Move archive management commands to Endo CLI
- refactor: Remove former Endo prototype
- refactor: Remove former Endo prototype from workspace


michaelfig (approved):
Yay for a negative lines-of-code PR!

PR #1021, opened 2022-01-26
fix: Publish all materials consistently

kriskowal:
Well, this is awkward.


michaelfig (approved):
Like herding cats.  Glad you did it!

PR #1025, opened 2022-01-28
fix(marshal): remove ambient types, generate declarations from JSDoc

michaelfig:
Experiment to remove dependency on ambient types.  The process:

1. Put `export {};` near the top of every JSDoc `types.js` file.  This switches it from an ambient file to a module with exported types.
2. Remove `exported.js` so that ambient type users break hard.
3. Add `export * from './src/types.js';` to `index.js` so that upstream consumers can import types from the package identifier.  Season with eslint-ignores to taste.
4. Remove any `import './types.js';`, `import './internal-types.js';` etc.
5. Add `/** @typedef {import('./types.js').MemberName} MemberName */` or for template types `/** @template T @typedef {import('./types.js').MemberName2<T>} MemberName2 */`. as needed to the top of `.js` sources.
6. Chase any other Typescript errors.
7. Add build step to generate declarations, modify `package.json` to include generated files.
7. In consumers of the package, add JSDoc typedef imports like `/** @typedef {import('@agoric/marshal').MyType} MyType */`, and remove any `import '@agoric/marshal/exported';`.
8. When all cross project dependencies have transitioned to explicitly exported type declarations, remove any `--maxNodeModulesJsDepth` hack that used to be needed with JSDoc ambient type in `node_modules`.

### DIFF
--- a/packages/marshal/exported.js
+++ b/packages/marshal/exported.js
@@ -1,1 +1,0 @@
-import './src/types.js';
--- a/packages/marshal/package.json
+++ b/packages/marshal/package.json
@@ -9,6 +9,9 @@
   },
   "scripts": {
     "build": "exit 0",
+    "clean": "tsc --build jsconfig.build.json --clean",
+    "prepack": "tsc --build jsconfig.build.json",
+    "postpack": "yarn clean",
     "test": "ava",
     "test:c8": "c8 $C8_OPTIONS ava --config=ava-nesm.config.js",
     "test:xs": "exit 0",
@@ -47,9 +50,11 @@
   "files": [
     "LICENSE*",
     "SECURITY*",
+    "NEWS*",
     "src",
     "*.js",
-    "*.ts"
+    "*.ts",
+    "*.map"
   ],
   "eslintConfig": {
     "extends": [
--- a/packages/marshal/src/deeplyFulfilled.js
+++ b/packages/marshal/src/deeplyFulfilled.js
@@ -3,14 +3,13 @@
 /// <reference types="ses"/>
 
 import { E } from '@endo/eventual-send';
-/**
- * @template T
- * @typedef {import('@endo/eventual-send').ERef<T>} ERef
- */
 import { isPromise } from '@endo/promise-kit';
 import { getTag, isObject } from './helpers/passStyle-helpers.js';
 import { makeTagged } from './makeTagged.js';
 import { passStyleOf } from './passStyleOf.js';
+
+/** @typedef {import('./types.js').Passable} Passable */
+/** @template T @typedef {import('@endo/eventual-send').ERef<T>} ERef */
 
 const { details: X, quote: q } = assert;
 const { ownKeys } = Reflect;
--- a/packages/marshal/src/helpers/internal-types.js
+++ b/packages/marshal/src/helpers/internal-types.js
@@ -2,6 +2,12 @@
 
 /// <reference path="../extra-types.d.ts" />
 
+export {};
+
+/** @typedef {import('../types.js').Checker} Checker */
+/** @typedef {import('../types.js').PassStyle} PassStyle */
+/** @typedef {import('../types.js').PassStyleOf} PassStyleOf */
+
 /**
  * The PassStyleHelper are only used to make a `passStyleOf` function.
  * Thus, it should not depend on an ambient one. Rather, each helper should be
--- a/packages/marshal/src/helpers/passStyle-helpers.js
+++ b/packages/marshal/src/helpers/passStyle-helpers.js
@@ -2,8 +2,8 @@
 
 /// <reference types="ses"/>
 
-import '../types.js';
-import './internal-types.js';
+/** @typedef {import('../types.js').Checker} Checker */
+/** @typedef {import('../types.js').PassStyle} PassStyle */
 
 const { details: X, quote: q } = assert;
 const {
--- a/packages/marshal/src/make-far.js
+++ b/packages/marshal/src/make-far.js
@@ -10,6 +10,8 @@ import {
 } from './helpers/remotable.js';
 import { pureCopy } from './pureCopy.js';
 
+/** @typedef {import('./types.js').InterfaceSpec} InterfaceSpec */
+
 const { quote: q, details: X } = assert;
 
 const { prototype: functionPrototype } = Function;
--- a/packages/marshal/src/make-far.js
+++ b/packages/marshal/src/make-far.js
@@ -83,8 +83,8 @@ const assertCanBeRemotable = candidate =>
  * Carol's `iface` as misrepresented by VatA.
  * @param {undefined} [props=undefined] Currently may only be undefined.
  * That plan is that own-properties are copied to the remotable
- * @param {object} [remotable={}] The object used as the remotable
- * @returns {object} remotable, modified for debuggability
+ * @param {any} [remotable={}] The object used as the remotable
+ * @returns {any} remotable, modified for debuggability
  */
 export const Remotable = (
   iface = 'Remotable',
--- a/packages/marshal/src/marshal-justin.js
+++ b/packages/marshal/src/marshal-justin.js
@@ -5,10 +5,11 @@
 import { Nat } from '@endo/nat';
 import { QCLASS } from './marshal.js';
 
-import './types.js';
 import { getErrorConstructor } from './helpers/error.js';
 import { isObject } from './helpers/passStyle-helpers.js';
 import { AtAtPrefixPattern, passableSymbolForName } from './helpers/symbol.js';
+
+/** @typedef {import('./types.js').Encoding} Encoding */
 
 const { ownKeys } = Reflect;
 const { isArray } = Array;
--- a/packages/marshal/src/marshal.js
+++ b/packages/marshal/src/marshal.js
@@ -5,7 +5,6 @@
 import { Nat } from '@endo/nat';
 import { assertPassable, passStyleOf } from './passStyleOf.js';
 
-import './types.js';
 import { getInterfaceOf } from './helpers/remotable.js';
 import { ErrorHelper, getErrorConstructor } from './helpers/error.js';
 import { makeTagged } from './makeTagged.js';
@@ -15,6 +14,15 @@ import {
   nameForPassableSymbol,
   passableSymbolForName,
 } from './helpers/symbol.js';
+
+/** @typedef {import('./types.js').MakeMarshalOptions} MakeMarshalOptions */
+/** @template Slot @typedef {import('./types.js').ConvertSlotToVal<Slot>} ConvertSlotToVal */
+/** @template Slot @typedef {import('./types.js').ConvertValToSlot<Slot>} ConvertValToSlot */
+/** @template Slot @typedef {import('./types.js').Serialize<Slot>} Serialize */
+/** @template Slot @typedef {import('./types.js').Unserialize<Slot>} Unserialize */
+/** @typedef {import('./types.js').Passable} Passable */
+/** @typedef {import('./types.js').InterfaceSpec} InterfaceSpec */
+/** @typedef {import('./types.js').Encoding} Encoding */
 
 const { ownKeys } = Reflect;
 const { isArray } = Array;
@@ -34,14 +42,18 @@ const { details: X, quote: q } = assert;
 const QCLASS = '@qclass';
 export { QCLASS };
 
+/** @type {ConvertValToSlot<any>} */
 const defaultValToSlotFn = x => x;
+/** @type {ConvertSlotToVal<any>} */
 const defaultSlotToValFn = (x, _) => x;
 
 /**
  * @template Slot
- * @type {MakeMarshal<Slot>}
+ * @param {ConvertValToSlot<Slot>} [convertValToSlot]
+ * @param {ConvertSlotToVal<Slot>} [convertSlotToVal]
+ * @param {MakeMarshalOptions} [options]
  */
-export function makeMarshal(
+export const makeMarshal = (
   convertValToSlot = defaultValToSlotFn,
   convertSlotToVal = defaultSlotToValFn,
   {
@@ -55,7 +67,7 @@ export function makeMarshal(
     marshalSaveError = err =>
       console.log('Temporary logging of sent error', err),
   } = {},
-) {
+) => {
   assert.typeof(marshalName, 'string');
   assert(
     errorTagging === 'on' || errorTagging === 'off',
@@ -67,7 +79,6 @@ export function makeMarshal(
   };
 
   /**
-   * @template Slot
    * @type {Serialize<Slot>}
    */
   const serialize = root => {
@@ -482,7 +493,6 @@ export function makeMarshal(
   };
 
   /**
-   * @template Slot
    * @type {Unserialize<Slot>}
    */
   const unserialize = data => {
@@ -507,4 +517,4 @@ export function makeMarshal(
     serialize,
     unserialize,
   });
-}
+};
--- a/packages/marshal/src/types.js
+++ b/packages/marshal/src/types.js
@@ -3,6 +3,8 @@
 
 /// <reference path="extra-types.d.ts" />
 
+export {};
+
 /**
  * @typedef { "undefined" | "null" |
  *   "boolean" | "number" | "bigint" | "string" | "symbol"
@@ -180,7 +182,7 @@
 
 /**
  * @template Slot
- * @typedef CapData
+ * @typedef {Object} CapData
  * @property {string} body A JSON.stringify of an Encoding
  * @property {Slot[]} slots
  */
@@ -201,22 +203,13 @@
 
 /**
  * @template Slot
- * @typedef Marshal
+ * @typedef {Object} Marshal
  * @property {Serialize<Slot>} serialize
  * @property {Unserialize<Slot>} unserialize
  */
 
 /**
- * @template Slot
- * @callback MakeMarshal
- * @param {ConvertValToSlot<Slot>=} convertValToSlot
- * @param {ConvertSlotToVal<Slot>=} convertSlotToVal
- * @param {MakeMarshalOptions=} options
- * @returns {Marshal<Slot>}
- */
-
-/**
- * @typedef MakeMarshalOptions
+ * @typedef {Object} MakeMarshalOptions
  * @property {'on'|'off'=} errorTagging controls whether serialized errors
  * also carry tagging information, made from `marshalName` and numbers
  * generated (currently by counting) starting at `errorIdNum`. The
--- a/packages/marshal/test/test-marshal-far-obj.js
+++ b/packages/marshal/test/test-marshal-far-obj.js
@@ -59,9 +59,10 @@ test('Remotable/getInterfaceOf', t => {
   t.is(p2.birthYear(2020), 1956, `birthYear() works`);
 
   // Remotables and Fars can be serialized, of course
-  function convertValToSlot(_val) {
+  /** @type {import('../src/types.js').ConvertValToSlot<string>} */
+  const convertValToSlot = _val => {
     return 'slot';
-  }
+  };
   const m = makeMarshal(convertValToSlot);
   t.deepEqual(m.serialize(p2), {
     body: JSON.stringify({
--- a/packages/captp/src/captp.js
+++ b/packages/captp/src/captp.js
@@ -1,6 +1,9 @@
 // @ts-check
 /// <reference types="ses"/>
 
+/** @template Slot @typedef {import('@endo/marshal').ConvertValToSlot<Slot>} ConvertValToSlot */
+/** @template Slot @typedef {import('@endo/marshal').ConvertSlotToVal<Slot>} ConvertSlotToVal */
+
 // Your app may need to `import '@endo/eventual-send/shim.js'` to get HandledPromise
 
 // This logic was mostly lifted from @agoric/swingset-vat liveSlots.js
--- a/packages/captp/src/types.js
+++ b/packages/captp/src/types.js
@@ -17,7 +17,7 @@
  */
 
 /**
- * @typedef {[boolean, CapData<CapTPSlot>]} TrapCompletion The head of the pair
+ * @typedef {[boolean, import('@endo/marshal').CapData<CapTPSlot>]} TrapCompletion The head of the pair
  * is the `isRejected` value indicating whether the sync call was an exception,
  * and tail of the pair is the serialized fulfillment value or rejection reason.
  * (The fulfillment value is a non-thenable.  The rejection reason is normally
--- a/packages/marshal/NEWS.md
+++ b/packages/marshal/NEWS.md
@@ -1,5 +1,17 @@
 User-visible changes in `@endo/marshal`:
 
+# Next release
+
+Switch from ambient to exported types.
+Include type declarations (`.d.ts`) generated from JSDoc to avoid requiring
+dependents to parse `.js` files in their `node_modules`.
+
+In order to use the types from `@endo/marshal` you now need to import them
+explicitly. For example, to make them available in scope, use the following:
+- JSDoc: `/** @typedef {import('@endo/marshal').PassStyle} PassStyle */`
+- TypeScript: `import type { PassStyle } from '@endo/marshal'`
+
+
 # v0.5.3 (2021-01-27)
 
 Includes TypeScript definitions in published artifact.
--- a/packages/marshal/index.js
+++ b/packages/marshal/index.js
@@ -39,3 +39,6 @@ export {
   isRecord,
   isCopyArray,
 } from './src/typeGuards.js';
+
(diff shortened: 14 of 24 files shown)


packages/marshal/src/helpers/passStyle-helpers.js
```diff
@@ -13,6 +13,10 @@ const {
 const { apply } = Reflect;
 const { isArray } = Array;
 
+/** @type {Checker} */
+export const conditionChecker = (cond, _details = undefined) => cond;
```
erights, line 17: I don't love the name `conditionChecker` but I don't have a better suggestion. I do see why you needed to introduce something.

packages/marshal/src/marshal-justin.js
```diff
@@ -402,6 +403,7 @@ const decodeToJustin = (encoding, shouldIndent = false) => {
       } else {
         out.open('{');
         for (const name of names) {
+          // @ts-ignore Type 'symbol' cannot be used as an index type.ts(2538)
```
erights, line 406: Interesting. Without the `@ts-ignore`, were you not getting the error before this PR but getting it with these changes? Do you know why?

packages/marshal/src/deeplyFulfilled.js
```diff
@@ -3,15 +3,14 @@
...
- * @template T
- * @typedef {import('@endo/eventual-send').ERef<T>} ERef
- */
 import { isPromise } from '@endo/promise-kit';
 import { getTag, isObject } from './helpers/passStyle-helpers.js';
 import { makeTagged } from './makeTagged.js';
 import { passStyleOf } from './passStyleOf.js';
 
+/** @typedef {import('./types').Passable} Passable */
+/** @template T @typedef {import('@endo/eventual-send').ERef<T>} ERef */
```
mhofman, line 12: Does that actually work? I remember running into typing issues with typedef of generics

packages/marshal/src/marshal.js
```diff
@@ -34,16 +42,20 @@ const { details: X, quote: q } = assert;
...
  * @template Slot
- * @type {MakeMarshal<Slot>}
+ * @param {ConvertValToSlot<Slot>} [convertValToSlot]
+ * @param {ConvertSlotToVal<Slot>} [convertSlotToVal]
+ * @param {MakeMarshalOptions} [options]
  */
-export function makeMarshal(
-  convertValToSlot = defaultValToSlotFn,
-  convertSlotToVal = defaultSlotToValFn,
+export const makeMarshal = (
```
erights, line 56: I missed that one. Thanks!

packages/marshal/src/deeplyFulfilled.js
```diff
@@ -3,15 +3,14 @@
...
-/**
- * @template T
- * @typedef {import('@endo/eventual-send').ERef<T>} ERef
- */
 import { isPromise } from '@endo/promise-kit';
 import { getTag, isObject } from './helpers/passStyle-helpers.js';
 import { makeTagged } from './makeTagged.js';
 import { passStyleOf } from './passStyleOf.js';
 
+/** @typedef {import('./types').Passable} Passable */
```
mhofman, line 11: Nit: while tsc doesn't care as much, I think we should keep the habit of using fully qualified names, aka with extensions, for imports

packages/marshal/src/marshal.js
```diff
@@ -67,7 +79,6 @@ export function makeMarshal(
   };
 
   /**
-   * @template Slot
    * @type {Serialize<Slot>}
```
erights, line 82: Why don't you need the `@template Slot` here?

packages/marshal/src/helpers/internal-types.js
```diff
@@ -2,6 +2,12 @@
 
 /// <reference path="../extra-types.d.ts" />
```
mhofman, line 3: Can we make this a typedef import as well ?

packages/marshal/src/marshal.js
```diff
@@ -382,7 +393,7 @@ export function makeMarshal(
           }
           case 'symbol': {
             const { name } = rawTree;
-            return passableSymbolForName(name);
+            return passableSymbolForName(`${name}`);
```
erights, line 396: Is this just a drive by improvement (thanks!), or did you notice this because of the enhanced type checking? Do you understand why?

packages/marshal/src/helpers/passStyle-helpers.js
mhofman, line 17: Nit: can we name this "default" or something?

packages/marshal/src/types.js
```diff
@@ -180,7 +182,7 @@
 
 /**
  * @template Slot
- * @typedef CapData
+ * @typedef {Object} CapData
```
erights, line 185: Good catch!

packages/marshal/test/test-marshal-far-obj.js
```diff
@@ -59,9 +59,10 @@ test('Remotable/getInterfaceOf', t => {
   t.is(p2.birthYear(2020), 1956, `birthYear() works`);
 
   // Remotables and Fars can be serialized, of course
-  function convertValToSlot(_val) {
+  /** @type {import('../src/types').ConvertValToSlot<string>} */
+  const convertValToSlot = _val => {
```
erights, line 63: Thanks.

packages/marshal/src/marshal.js
```diff
@@ -34,16 +42,20 @@ const { details: X, quote: q } = assert;
...
+ * @param {ConvertValToSlot<Slot>} [convertValToSlot]
+ * @param {ConvertSlotToVal<Slot>} [convertSlotToVal]
+ * @param {MakeMarshalOptions} [options]
  */
-export function makeMarshal(
-  convertValToSlot = defaultValToSlotFn,
-  convertSlotToVal = defaultSlotToValFn,
+export const makeMarshal = (
+  convertValToSlot = defaultSlotToValFn,
+  convertSlotToVal = defaultValToSlotFn,
```
mhofman, line 58: These got flipped (doesn't make much of a difference given their behavior).

mhofman (changes requested):
Cool stuff! I'd be so much more confident in this kind of changes if we had `noImplicitAny`.

Requesting change as I believe the default convert slot got flipped.

erights (approved):
A worthy experiment. It is much less painful than I expected. Enough so that, assuming this is representative, I'm ok moving in this direction in general.

vscode has a very nice feature that when I use a name I have not yet imported but should, it often (not always) prompts me with the correct choice and writes the `import` itself. If it did so for this `/** @typedef {import('../types').Foo} Foo */` pattern as well, then I wouldn't even hesitate. Maybe eventually we'll have such tooling.

Thanks for pushing ahead with this. A very pleasant surprise. LGTM!!!

packages/marshal/src/marshal.js
erights: Good catch! Makes a huge difference, just not to the computer ;)

mhofman:
> vscode has a very nice feature that when I use a name I have not yet imported but should, it often (not always) prompts me with the correct choice and writes the `import` itself.

It does for `import type` in `.ts` files, but unfortunately I'm not aware of a similar jsdoc style.

One drawback I believe is that going to the type definition would take 2 clicks, first one going to the `typedef` alias.

packages/marshal/src/deeplyFulfilled.js
turadg: Reasons to leave it this way:
- There should never be any ambiguity what file it refers to. (There must not be a `types.ts` next to a `types.js`.)
- We can change the syntax of the imported file between `.ts` and `.js` without touching everywhere it's imported.

packages/marshal/exported.js
```diff
@@ -1 +1,2 @@
-import './src/types.js';
+// eslint-disable-next-line import/export
+export * from './src/types.js';
```
turadg, line 2: +1 to this pattern of explicit type exports over implicit.

We shouldn't have to disable the linter to do the right thing. https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/export.md says it prevents "funny business"; doesn't TypeScript cover that? I propose we consider disabling this rule in our eslint config

packages/marshal/src/helpers/passStyle-helpers.js
```diff
@@ -2,8 +2,8 @@
 
 /// <reference types="ses"/>
 
-import '../types.js';
-import './internal-types.js';
+/** @typedef {import('../types').Checker} Checker */
```
turadg, line 5: If we're sticking with JSDoc, maybe we can submit a PR for TypeScript to have a nicer syntax for this.

In regular TS this is clean:
```
import {type Checker} from '../types`
```

In JSDoc TS maybe,
```
/**
 * @typeimport '../types' {Checker}
 */
```

packages/marshal/src/helpers/passStyle-helpers.js
```diff
@@ -2,8 +2,8 @@
...
+/** @typedef {import('../types').PassStyle} PassStyle */
```
turadg, line 6: If we're sticking with JSDoc, maybe we can submit a PR for TypeScript to have a nicer syntax for this.

In regular TS this is clean:
```
import type {Checker, PassStyle} from '../types`
```

In JSDoc TS maybe,
```
/**
 * @typeimport '../types' {Checker, PassStyle}
 */
```

packages/marshal/src/deeplyFulfilled.js
mhofman: Even if there is a `.d.ts` or `.ts` file, the import should always be of `.js` file per TypeScript conventions. TypeScript will internally use the `.d.ts` file if it exists. We use explicit `.js` extensions for our imports as automatic extension resolution is something we left behind when switching to native ESM.

packages/marshal/src/types.js
turadg: Curious why `tsc` doesn't report this. Is there a valid interpretation? I suppose it's more lenient with JSDoc than code syntax.

turadg (approved):
LGTM. I left some non-blocking comments.

Glad to see support for type imports, which came up in https://github.com/Agoric/agoric-sdk/pull/4410#discussion_r794093519

packages/marshal/src/deeplyFulfilled.js
turadg: Thanks for the correction on the second bullet about changing extensions. I didn't know that TypeScript conventions are that if the file is `foo.ts` you should still import it as `foo.js`.

I neglected to include what I think is the most compelling reason to leave it this way: less to type and read. I think simplicity wins unless there is a compelling reason to add the extension. Is it to be consistent with the JSM module imports? I think consistency is important when it reduces confusion or misunderstanding, but I have a hard time imagining someone being confused by the absence of `.js` here in the JSDoc.

mhofman: Right I was looking for consistency. I've never found terseness to be very compelling. We require extensions for ESM imports (static, and in the future dynamic), and since TS style jsdoc type imports use a dynamic-like ESM notion, I figure we should stay consistent.

packages/marshal/src/marshal.js
mhofman: I fixed this in https://github.com/endojs/endo/pull/1025/commits/b55a2bed166dbf8308a89d95a7c9ef2de41ee988

packages/marshal/src/marshal-justin.js
mhofman: I addressed this issue in #1043 (the new base of this PR) by addressing why the `names` had the wrong types in the first place.

packages/marshal/src/marshal.js
mhofman: `Slot` is already the template of the outer function,

mhofman:
I took over the PR and extended it with generation of declaration files for marshal, and rebased on latest master + preliminary TS 4.5 compat PR I opened in #1043. I've checked that with the appropriate updates to https://github.com/Agoric/agoric-sdk/pull/4413, all seem to be working well. This shows how we can move away from ambient types, and stop requiring TS from trying to parse the JS of `node_modules` to figure out typing. With this we should be able to publish packages that are more compatible with the ecosystem.

packages/marshal/src/make-far.js
```diff
@@ -81,8 +83,8 @@ const assertCanBeRemotable = candidate =>
  * Carol's `iface` as misrepresented by VatA.
  * @param {undefined} [props=undefined] Currently may only be undefined.
  * That plan is that own-properties are copied to the remotable
- * @param {object} [remotable={}] The object used as the remotable
- * @returns {object} remotable, modified for debuggability
+ * @param {any} [remotable={}] The object used as the remotable
+ * @returns {any} remotable, modified for debuggability
```
mhofman, line 87: This was necessary to fix a type inference issue in `zcfZygote`: https://github.com/Agoric/agoric-sdk/blob/mfig-marshal-explicit-jsdoc/packages/zoe/src/contractFacet/zcfZygote.js#L242

The inferred type when parsing JS was actually `any`, but somehow when switching to emitted declaration, it carried the `object` type, which is incompatible in some usages.

packages/marshal/src/helpers/passStyle-helpers.js
turadg: I'm not clear on why we all the `check = conditionChecker` can't be simply `check = v => v`. Is it because we need that default function to be hardened?

If so, wouldn't it still be better to name this function by what it does instead of how it's used (as yet)? E.g. `identity`. I imagine it would be useful to have a hardened identity function as a general utility.

packages/marshal/src/make-far.js
turadg: Unfortunate to regress in type specificity but solving it would be a rabbit hole that's out of scope.

packages/marshal/src/marshal.js
turadg: @michaelfig what's the reason for this change?

packages/marshal/src/types.js
turadg: Confirmed, the earlier expression worked the same.

packages/marshal/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "./jsconfig.json",
```
turadg, line 2: why keep both files? I'm pretty sure `tsconfig` alone can do the job these are in concert.

turadg (approved):
Left some non-blocking suggestions and questions.

packages/marshal/src/helpers/passStyle-helpers.js
mhofman: I believe that's because of the way type inference works now if no explicit `@param` have been defined for `check`. If you just use a `check = v => v`, then the type of check will be inferred as `(v: any) => any`, and thus will only have 1 allowed arg instead of 2.

packages/marshal/tsconfig.json
mhofman: tsconfig is for generating the `.d.ts` file. I don't want to generate those and clutter the workspace when type checking, and simply forcing `noEmit true` on the command line for `lint:types` didn't work (conflicting value with `emitDeclarationOnly`)

mhofman: Happy to rename to `jsconfig.build.json` if you believe that's a better name.

michaelfig: > Happy to rename to `jsconfig.build.json` if you believe that's a better name.

I like this better, assuming you mean: rename `tsconfig.json` to `jsconfig.build.json`.

packages/marshal/src/marshal.js
michaelfig: Not yet, but maybe sometime.

packages/marshal/src/marshal.js
michaelfig: It's a cast needed by the newer typescript: `name` could be a `number` or something like that.

mhofman: Ha I think I fixed the root cause of that with a different cast. Will see if I can undo this one

packages/marshal/src/helpers/passStyle-helpers.js
turadg: The type is defined by `PassStyleHelper` so it's not a type problem. Maybe it was before and something improved since then? 

If there's no runtime need for the `conditionChecker` function then I'd suggest including 9aba0c028bd501d943b8b83c1ec0e4d0d6d6e05d in this  PR

packages/marshal/tsconfig.json
turadg: Yes, I think if this config file is to be used only for building then it would help to have that conveyed in the filename.

Separately (and not now), I'd like to consider using `tsconfig` instead of `jsconfig` so that `yarn tsc` to typecheck doesn't require an option. My understanding is that with the right options a tsconfig can work the same as a jsconfig.

mhofman: Done

packages/marshal/src/marshal.js
mhofman: I indeed did, reverted this change.

packages/marshal/src/helpers/passStyle-helpers.js
mhofman: Reverted as I found no need for this anymore.

mhofman:
Reverted some of the changes so this PR is now purely type changes (and a couple `function foo() {}` -> `const foo = () => {}`)

kriskowal (approved):

kriskowal (changes requested):
Moving away from ambient types is a good but breaking change. This will require some ceremony.

In each affected `NEWS.md`, we need a note for `# Next release` of what end-users will need to do when they upgrade.

At least one commit must have `!` in the conventional commit and a `*BREAKING CHANGE*: ...`  paragraph that will get rolled into the changelog and bump the version appropriately.

When integrating into Agoric SDK, I use `scripts/sync-versions.sh ../endo` and that rolls over all the semver bumps, so all of the alignment will have to occur in the Sync Endo PR.

packages/marshal/src/marshal.js
erights: I would love to see this done automatically!

packages/marshal/exported.js
mhofman: @michaelfig, could we not include this `export * from './src/types.js';` in the `index.js`, so that we can get rid of the `exported.js`? I believe the only purpose it served was for making sure marhsal ambient types were loaded, which we want to get rid of, right?

mhofman:
@kriskowal I believe I added the required documentation. Happy to improve it. I left this stacked for now and will squash most of these appropriately once I have a sense everyone had time to review the new changes. (cc @michaelfig )

kriskowal:
In the commit message, `BREAKING CHANGE:` must be exactly `*BREAKING CHANGE*:`for Lerna’s conventional commit changelog thing to pick it up.

kriskowal (approved):
Please address my comment about the commit message having `*BREAKING CHANGE:*` verbatim. Otherwise, this is great progress.

mhofman:
> Please address my comment about the commit message having `*BREAKING CHANGE:*` verbatim. Otherwise, this is great progress.

Done.

@michaelfig could you re-review as a comment? I'll carry your review as mine (since you are the original author...)

packages/marshal/exported.js
mhofman: I have gone ahead and done that. It reads much better.

packages/marshal/package.json
```diff
@@ -9,6 +9,7 @@
   },
   "scripts": {
     "build": "exit 0",
+    "prepack": "tsc -p jsconfig.build.json",
```
michaelfig, line 12: Can you please add:
```suggestion
    "prepack": "tsc -p jsconfig.build.json",
    "clean": "tsc-emit-clean",
    "postpack": "yarn clean",
```

and take a dependency on `tsc-emit-clean` here?

packages/marshal/src/types.js
```diff
@@ -201,22 +203,16 @@
...
  */
 
 /**
- * @template Slot
- * @callback MakeMarshal
- * @param {ConvertValToSlot<Slot>=} convertValToSlot
- * @param {ConvertSlotToVal<Slot>=} convertSlotToVal
- * @param {MakeMarshalOptions=} options
- * @returns {Marshal<Slot>}
  */
```
michaelfig, line 212: Some leftover JSDoc here.

michaelfig (commented):
Looking good!  Just a few minor cleanups I'd recommend, but not blocking.

packages/marshal/package.json
mhofman: Apparently `yarn tsc --build jsconfig.build.json  --clean` works!

mhofman (approved):
Carrying @michaelfig's approval

packages/marshal/src/marshal.js
turadg: an attempt: https://github.com/Agoric/agoric-sdk/pull/4608

turadg:
I just noticed that building the d.ts files (e.g. by running prepack) spits them out adjacent to the .js files, which aren't gitignored and could end up being checked in.

Any reasons not to output the build into a `dist` directory that is ignored?

mhofman:
Because we'd need to also copy the .js source, which would be incompatible with non-publish same repo usage

turadg:
> Because we'd need to also copy the .js source, which would be incompatible with non-publish same repo usage

Incompatible how? If we pointed `"main":` at dist then that yeah we'd have to wait on a build step for same-repo usage, which is undesirable. However if you just point `"types": "dist/…"` how does that affect same-repo usage? 

A problem with how it is now (besides the gitignore) is that `src/foo.js` and `src/foo.ts` can be out of sync. If build output went to a `dist/` then you'd have a guarantee that `.ts` adjacent to its source `.js` matched.

packages/marshal/src/helpers/passStyle-helpers.js
turadg: Not the first to want this :) https://github.com/microsoft/TypeScript/issues/22160

mhofman:
I am not ok with anything that requires same repo build step to have anything work. It's a foot gun.

How can they be out of sync? They're only generated when publishing a package, and not needed same repo as tsc will happily parse the JS to get type definitions when the dependency is not in node_modules

turadg:
> I am not ok with anything that requires same repo build step to have anything work. It's a foot gun.

I'm not familiar with the foot injuries you have in mind. In my experience watch mode suffices to keep things up to date.

> How can they be out of sync? They're only generated when publishing a package

They're generated when publishing a package, but don't go away on their own. Then the source .js can be edited and is out of sync. Not "expected", but perhaps a "foot gun".

mhofman:
> They're generated when publishing a package, but don't go away on their own

True, we need a clean step after publish, and / or move publishing to CI.

Watch step requires developer environment setup. I think the fact these build artifacts are not ignored is probably a good thing as it'll remind a clean is needed.

kriskowal:
I would very much prefer not to need a watch mode.

michaelfig:
> > They're generated when publishing a package, but don't go away on their own

@turadg, there's a `postpack` that does a clean.  Are you saying that doesn't work for you?

kriskowal:
> > > They're generated when publishing a package, but don't go away on their own
> 
> @turadg, there's a `postpack` that does a clean. Are you saying that doesn't work for you?

After my last publish, the `.d.ts` and `.map` files were left behind.

michaelfig:
I find with the following, things work as they should (at least wrt. clean).  Explain again why we're including `*.ts` in our `jsconfig.json`s?

```diff
diff --git a/packages/marshal/jsconfig.json b/packages/marshal/jsconfig.json
index b455fcf0d..0611838b4 100644
--- a/packages/marshal/jsconfig.json
+++ b/packages/marshal/jsconfig.json
@@ -15,5 +15,5 @@
     "strictNullChecks": true,
     "moduleResolution": "node",
   },
-  "include": ["*.js", "*.ts", "src/**/*.js", "src/**/*.ts", "test/**/*.js"]
+  "include": ["*.js", "src/**/*.js", "test/**/*.js"]
 }
```

kriskowal:
That’s good to know. That patch is acceptable. I made the change to cover `.ts` files in general across the repository, since I found that hand-written definitions were not covered by tests, which caused failures to only be observable when integrated in a dependee.

michaelfig:
> I made the change to cover `.ts` files in general across the repository, since I found that hand-written definitions were not covered by tests, which caused failures to only be observable when integrated in a depended.

Maybe the correct solution then is to do:

```diff
diff --git a/packages/marshal/jsconfig.build.json b/packages/marshal/jsconfig.build.json
index 13018509f..ea9465e90 100644
--- a/packages/marshal/jsconfig.build.json
+++ b/packages/marshal/jsconfig.build.json
@@ -6,5 +6,5 @@
     "emitDeclarationOnly": true,
     "declarationMap": true
   },
-  "exclude": ["test/"]
+  "exclude": ["test/", "**/*.ts"]
 }
```

PR #1026, opened 2022-01-28
fix(compartment-mapper): Needless genericity considered harmful

kriskowal:
The has utility function in the compartment mapper did not need to be generic.  Because it was generic, at a vast distance down the dependency chain, TypeScript proposed bizarre combinations of types for each of the uses, probably sampling the first ambient type that emitted a valid function, then failed type check in compartment-mapper.

### DIFF
--- a/packages/compartment-mapper/src/link.js
+++ b/packages/compartment-mapper/src/link.js
@@ -45,10 +45,8 @@ const defaultCompartment = Compartment;
 const q = JSON.stringify;
 
 /**
- * @template {string | number} Key
- * @template Value
- * @param {Record<Key, Value>} object
- * @param {Key} key
+ * @param {Record<string, unknown>} object
+ * @param {string} key
  * @returns {boolean}
  */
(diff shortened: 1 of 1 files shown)


packages/compartment-mapper/src/link.js
```diff
@@ -45,10 +45,8 @@ const defaultCompartment = Compartment;
 const q = JSON.stringify;
 
 /**
- * @template {string | number} Key
- * @template Value
```
mhofman, line 49: There laid your problem!

It should have been `@template {any} Value`

`Value` is a deprecated alias for `AmountValue`.

You should be able to keep it generic if you want.

mhofman (approved):
Fine fix, but the problem was a malformed `template`

packages/compartment-mapper/src/link.js
mhofman: After double checking, adding a type constraint wouldn't have helped, only renaming to something that didn't conflict with a type in scope would have, e.g. `TValue`

PR #1027, opened 2022-01-28
fix(ses): Globals have vars, not consts

kriskowal:
Evidently, in a recent change to appease eslint, SES stopped exporting a type for globalThis.assert et al.

### DIFF
--- a/packages/ses/index.d.ts
+++ b/packages/ses/index.d.ts
@@ -1,4 +1,4 @@
-/* eslint-disable no-restricted-globals */
+/* eslint-disable no-restricted-globals, vars-on-top, no-var */
 
 // It's academically tempting to define a hardened type, but TypeScript doesn't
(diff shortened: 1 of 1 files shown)


packages/eslint-config/eslint-config.json
```diff
@@ -94,7 +94,9 @@
       ],
       "rules": {
         "import/no-unresolved": "off",
-        "no-unused-vars": "off"
+        "no-unused-vars": "off",
+        "vars-on-top": "off",
+        "no-var": "off"
```
mhofman, line 99: that's pretty extreme, can we just disable those rules in the `.d.ts` file

mhofman (changes requested):
Ah yeah the quirks of `.d.ts` global definition files. However the eslint change should not be global, but scoped to the file.

mhofman (approved):

PR #1028, opened 2022-01-28
feat(stream)!: Protect streams against reentrance

kriskowal:
This is a proposed change that would by default protect stream methods from reentrance by defering method invocations to subsequent turns.  However, it does not work because eventual-send hardens arguments, and for streams, sending a typed array must be a valid interaction.

I propose that harden needs to treat typed arrays as a special case of collections like Map and Set, which can be hardened. The typed arrays are equally mutable, in that their content is mutable. Hardening a typed array could instead prevent extensions and make its non-content properties non-configurable.


erights:
> make its non-content properties non-configurable

did you mean

make its non-content properties non-configurable non-writable

?

kriskowal:
> > make its non-content properties non-configurable
> 
> did you mean
> 
> make its non-content properties non-configurable non-writable
> 
> ?

yes

erights (commented):
If we successfully do the `harden` change you suggest, then this will LGTM.

And I am hopeful about that `harden` change.

kriskowal:
Blocked by #1029

kriskowal:
Hopes in the `harden` change have proven fruitful without modification here so I’ve opened this up for review.

erights (approved):
LGTM!

PR #1030, opened 2022-01-29
chore: release

kriskowal:
Needed to reintegrate into Agoric SDK.

### DIFF
--- a/packages/base64/CHANGELOG.md
+++ b/packages/base64/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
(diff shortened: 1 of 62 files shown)


michaelfig (approved):
LGTM!  Is there any way to do better integration testing with agoric-sdk so that you won't have to crank releases as frequently?

kriskowal:
Yeah, in the past I’ve added all of the Endo packages temporarily to the Agoric SDK workspace for integration testing, or installed individual packages from packs. I imagine I will soon know whether that is more or less expensive than cranking out releases at high frequency!

In this case, I did the fixes in-place in Agoric SDK modules and vetted them there on https://github.com/Agoric/agoric-sdk/pull/4384 using patch-package until the build was green, then applied them here.

PR #1032, opened 2022-01-29
feat(ses): Harden typed arrays

### LINKED ISSUE #1029: Hardening TypedArray
Currently, freezing a TypedArray throws, so harden throws. But

- typed arrays really ought to be directly marshalable for ergonomic reasons. The alternative is to wrap all typedarrays in a Blob wrapper of some kind
- hardening a Map or Set is morally equivalent to preventing extensions and making all properties non-writable on a TypedArray

Making a special case for typed array in harden poses a couple challenges:

Differentiating a typed array appears to require try/catch, and try/catch in a hot loop is expensive. try/catch deoptimizes the entire surrounding function, so the usual trick to making it go fast is to have an intermediate function on the stack between two optimizable functions. We could amortize the cost by making a freeze queue and a trampoline that may resume flushing the queue if freeze throws and verifying the type of the typed array does not.

The current best idea for inducing the isTypedArray check is:

```js
Reflect.apply(Object.getOwnPropertyDescriptor(Object.getPrototypeOf(Uint8Array.prototype), Symbol.toStringTag).get, target, [])
```

as of 2022-01-28, by @mhofman

kriskowal:
Harden can now handle typed arrays.  Although typed arrays cannot be frozen because their integer properties cannot be made non-writable by any mechanism, these properties are analogous to the entries in a map, which can be hardened.  This change makes harden treat typed arrays as the moral equivalent of a map.  Hardening a typed array seals the object and makes all other properties non-writable and non-configurable.

Fixes #1029

### DIFF
--- a/packages/ses/src/make-hardener.js
+++ b/packages/ses/src/make-hardener.js
@@ -38,6 +38,7 @@ import {
   isObject,
   objectHasOwnProperty,
   ownKeys,
+  preventExtensions,
   seal,
   setAdd,
   setForEach,
@@ -80,6 +81,8 @@ export const isTypedArray = object => {
 const freezeTypedArray = array => {
   const descs = getOwnPropertyDescriptors(array);
 
+  preventExtensions(array);
+
   // Downgrade writable expandos to readonly, even if non-configurable.
   arrayForEach(ownKeys(descs), (/** @type {string | symbol} */ name) => {
     const desc = descs[/** @type {string} */ (name)];
@@ -88,6 +91,13 @@ const freezeTypedArray = array => {
     // This is a strange behavior intrinsic to TypedArrays, but no more harmful
     // than the mutability of properties of a hardened Map or Set,
     // so we carve out this exceptional behavior.
+    //
+    // TypedArrays are integer-indexed exotic objects, so indexed properties
+    // outside the range of 0 to the typed array's length are disallowed.
+    // Assignment to these indexes silently fails.
+    // So, we only need to make non-index properties non-writable and
+    // non-configurable.
+    // https://tc39.es/ecma262/#sec-integer-indexed-exotic-objects
     const number = +String(name);
     if (!isInteger(number)) {
       defineProperty(array, name, {
@@ -97,8 +107,6 @@ const freezeTypedArray = array => {
       });
     }
   });
-
-  seal(array);
 };
 
 /**
--- a/packages/ses/src/make-hardener.js
+++ b/packages/ses/src/make-hardener.js
@@ -39,7 +39,6 @@ import {
   objectHasOwnProperty,
   ownKeys,
   preventExtensions,
-  seal,
   setAdd,
   setForEach,
   setHas,
--- a/packages/ses/src/make-hardener.js
+++ b/packages/ses/src/make-hardener.js
@@ -27,25 +27,79 @@ import {
   TypeError,
   WeakMap,
   WeakSet,
+  apply,
   arrayForEach,
+  defineProperty,
   freeze,
+  getOwnPropertyDescriptor,
   getOwnPropertyDescriptors,
   getPrototypeOf,
+  isInteger,
   isObject,
   objectHasOwnProperty,
   ownKeys,
+  seal,
   setAdd,
   setForEach,
   setHas,
+  toStringTagSymbol,
+  typedArrayPrototype,
   weakmapGet,
   weakmapSet,
   weaksetAdd,
   weaksetHas,
 } from './commons.js';
+import { assert } from './error/assert.js';
 
 /**
  * @typedef {import('../index.js').Harden} Harden
  */
+
+// Obtain the string tag accessor of of TypedArray so we can indirectly use the
+// TypedArray brand check it employs.
+const typedArrayToStringTag = getOwnPropertyDescriptor(
+  typedArrayPrototype,
+  toStringTagSymbol,
+);
+assert(typedArrayToStringTag);
+const getTypedArrayToStringTag = typedArrayToStringTag.get;
+assert(getTypedArrayToStringTag);
+
+// Exported for tests.
+/** @param {unknown} object */
+export const isTypedArray = object => {
+  // The object must pass a brand check or toStringTag will return undefined.
+  const tag = apply(getTypedArrayToStringTag, object, []);
+  return tag !== undefined;
+};
+
+/**
+ * @template T
+ * @param {ArrayLike<T>} array
+ */
+const freezeTypedArray = array => {
+  const descs = getOwnPropertyDescriptors(array);
+
+  // Downgrade writable expandos to readonly, even if non-configurable.
+  arrayForEach(ownKeys(descs), (/** @type {string | symbol} */ name) => {
+    const desc = descs[/** @type {string} */ (name)];
+    // The numbered properties are writable and non-configurable,
+    // and cannot be made non-writable by defineProperty.
+    // This is a strange behavior intrinsic to TypedArrays, but no more harmful
+    // than the mutability of properties of a hardened Map or Set,
+    // so we carve out this exceptional behavior.
+    const number = +String(name);
+    if (!isInteger(number)) {
+      defineProperty(array, name, {
+        ...desc,
+        writable: false,
+        configurable: false,
+      });
+    }
+  });
+
+  seal(array);
+};
 
 /**
  * Create a `harden` function.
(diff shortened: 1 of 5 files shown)


kriskowal:
cc @mhofman

kriskowal:
I took a run at using an error trampoline instead of brand checking every single object. The code fit pretty well, but whenever freeze threw an error for a typed array, the queue would have to be resumed flushing from the beginning. Deleting entries from `toFreeze` is not an option: they all need to be added to the WeakMap of hardened objects later. So, the alternative was to mark the object and reflush the queue from the beginning. So, there ended up being a brand check before freeze regardless of the mechanism. Just calling `isTypedArray` on every turns out to be considerably simpler, and probably performs no worse than checking for the object’s existence in a WeakSet.

packages/ses/src/make-hardener.js
```diff
@@ -27,26 +27,80 @@ import {
...
+);
+assert(typedArrayToStringTag);
+const getTypedArrayToStringTag = typedArrayToStringTag.get;
+assert(getTypedArrayToStringTag);
+
+// Exported for tests.
+/** @param {unknown} object */
+export const isTypedArray = object => {
+  // The object must pass a brand check or toStringTag will return undefined.
+  const tag = apply(getTypedArrayToStringTag, object, []);
```
mhofman, line 72: How bad is the overhead of empty array allocation these days?

packages/ses/src/make-hardener.js
```diff
@@ -27,26 +27,80 @@ import {
...
+  // Downgrade writable expandos to readonly, even if non-configurable.
+  arrayForEach(ownKeys(descs), (/** @type {string | symbol} */ name) => {
+    const desc = descs[/** @type {string} */ (name)];
+    // The numbered properties are writable and non-configurable,
+    // and cannot be made non-writable by defineProperty.
+    // This is a strange behavior intrinsic to TypedArrays, but no more harmful
+    // than the mutability of properties of a hardened Map or Set,
+    // so we carve out this exceptional behavior.
+    const number = +String(name);
+    if (!isInteger(number)) {
```
mhofman, line 92: Add a comment that the spec doesn't allow defining own props outside of the range `0..length` because typed arrays are ["Integer-Indexed Exotic Objects"](https://tc39.es/ecma262/#sec-integer-indexed-exotic-objects), so we're guaranteed that indexed props are the ones from the TypedArray instrinsic.

packages/ses/src/make-hardener.js
```diff
@@ -27,26 +27,80 @@ import {
...
+      defineProperty(array, name, {
+        ...desc,
+        writable: false,
+        configurable: false,
+      });
+    }
+  });
+
+  seal(array);
```
mhofman, line 101: Should seal/prevent extensions before walking the props. The object shouldn't be a proxy, so it should not be possible, but I'd feel more confident that own props don't magically appear after walking them.

packages/ses/src/make-hardener.js
```diff
@@ -27,26 +27,80 @@ import {
...
+  // The object must pass a brand check or toStringTag will return undefined.
+  const tag = apply(getTypedArrayToStringTag, object, []);
+  return tag !== undefined;
+};
+
+/**
+ * @template T
+ * @param {ArrayLike<T>} array
+ */
+const freezeTypedArray = array => {
```
mhofman, line 80: A typed array might be a subclass of the intrinsics typed array, so we need to freeze the prototype chain.

mhofman (changes requested):
Subclasses of types array are not handled (custom protos would be left mutable).

Small nit on order of seal / "freezing" extra own props.

packages/ses/src/make-hardener.js
kriskowal: That should be covered by `freezeAndTraverse`, which enqueues the prototype.

packages/ses/src/make-hardener.js
kriskowal: Sealing before breaks some of the expando tests, but I can use preventExtensions instead. I think that the loop would make sealing redundant. All the tests pass with preventExtensions first and seal occurring nowhere.

packages/ses/src/make-hardener.js
kriskowal: This is hot enough to care, but it’s not the long pole on this line. `uncurryThis` is in play, and it allocates profusely.

kriskowal: (I imagine that when we begin to care earnestly about SES performance, we will be compelled to examine how much care we need to give toward protection from flaws in the TCB like vetted shims, and how much SES can lean on Lockdown to protect its own invariants instead of these performance-expensive measures.)

packages/ses/src/make-hardener.js
kriskowal: Having thought about this more, I now understand the case you’re describing and I will add a test. I believe we’re covered, but best to be sure.

packages/ses/src/make-hardener.js
mhofman: Ugh? That shouldn't. We should always be able to make a non-configurable property non-writable.

packages/ses/src/make-hardener.js
mhofman: Ah yeah my bad. Thanks for adding the test tho :)

packages/ses/src/make-hardener.js
```diff
@@ -88,6 +91,13 @@ const freezeTypedArray = array => {
     // This is a strange behavior intrinsic to TypedArrays, but no more harmful
     // than the mutability of properties of a hardened Map or Set,
     // so we carve out this exceptional behavior.
+    //
+    // TypedArrays are integer-indexed exotic objects, so indexed properties
+    // outside the range of 0 to the typed array's length are disallowed.
+    // Assignment to these indexes silently fails.
```
mhofman, line 96: and defineOwnProp throws ;)

mhofman (approved):

erights:
@phoddie @patrick-soquet , we are thinking of changing the definition of `harden` as shown in this PR. However, before we do, we need to coordinate with you. We want to be sure that our shim continues to agree with your native implementation and with what we'd both like to propose to tc39 and tc53. What do you think of this change?

Obviously, we also need to talk about how this change interacts with our joint plans for frozen/pure/rommable typed arrays. AFAICT, this change is friendly to those other issues, but I may be missing something.

erights (changes requested):
I cannot add @phoddie or @patrick-soquet as official reviewers, so I'm filing this vacuous "Request changes" to block until we get feedback from them.

@phoddie or @patrick-soquet , please consider yourselves unofficially added as reviewers. Awaiting your feedback. Thanks!

phoddie:
We'll take a look at this.

kriskowal:
I’ve sent Peter and Patrick invitations to the org so they can participate in reviews going forward.

packages/ses/src/make-hardener.js
erights: I'm seeing bizarre behavior on v8:
```
> Object.defineProperty(ta, 88, { value: 99 })
Uncaught TypeError: Invalid typed array index
    at Function.defineProperty (<anonymous>)
> Object.defineProperty(ta, 88.1, { value: 99 })
Uncaught TypeError: Invalid typed array index
    at Function.defineProperty (<anonymous>)
> Object.defineProperty(ta, -88.1, { value: 99 })
Uncaught TypeError: Invalid typed array index
    at Function.defineProperty (<anonymous>)
> Object.defineProperty(ta, '-88.1', { value: 99 })
Uncaught TypeError: Invalid typed array index
    at Function.defineProperty (<anonymous>)
> Object.defineProperty(ta, '-88.10', { value: 99 })
Uint8Array(0) []
> ta['-88.10']
99
```
Is this a v8 bug? I don't think it makes the code above incorrect, but it does mean that the spec may not be an accurate picture of what engines do at runtime.

erights (commented):
Once we've coordinated with Peter and Patrick, this will LGTM.

packages/ses/src/make-hardener.js
mhofman: That's indeed weird!

packages/ses/src/make-hardener.js
```diff
@@ -27,26 +27,88 @@ import {
...
 } from './commons.js';
+import { assert } from './error/assert.js';
 
 /**
  * @typedef {import('../index.js').Harden} Harden
  */
 
+// Obtain the string tag accessor of of TypedArray so we can indirectly use the
+// TypedArray brand check it employs.
+const typedArrayToStringTag = getOwnPropertyDescriptor(
```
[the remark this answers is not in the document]
mhofman: I guess because typed arrays are weirdly meta, and the actual kind of typed array is saved as a slot in the spec?

phoddie:
After some internal discussion, we're fine with the approach you've taken here. We have implemented the behavior for the native version of `harden` in xsnap (which we'll get to you via PR shortly). Thank you for the opportunity to review ahead of these changes landing.

> Obviously, we also need to talk about how this change interacts with our joint plans for frozen/pure/rommable typed arrays. AFAICT, this change is friendly to those other issues, but I may be missing something.

Agreed – this change does not interfere fully immutable TypedArrays (where the main challenge is making the backing buffer immutable).

erights (approved):
Based on feedback from @phoddie and @patrick-soquet , LGTM. Thanks!

PR #1033, opened 2022-02-01
feat(daemon): Add private makeWorker

kriskowal:
Refs: #1036


Tartuffo:
Since you moved this to In Progress, I assume you are the assignee.  Please assign someone else if that is not the case!

kriskowal:
Moved to #1067 for the more specific branch name on the `endo` branch stack.

Issue #1046, opened 2022-02-02
Drop Node.js 12 support in CI


kriskowal:
Node.js 12 is becoming a burden. We need to drop it to upgrade TypeScript and ESLint. This would also give us an opportunity to drop some hacks around CommonJS named exports. This would also let us drop a work-around for old V8 generator function behavior in streams. Neither of those changes need to be done to close this issue. We need:

- [x] Remove 12 from CI matrixes.
- [ ] Update Node.js `engines` properties in `package.json` where applicable. Soft requirement.
- [ ] NEWS.md entries.
- [ ] This might be a whole-repo breaking change. Dropping an old Node.js version is breaking at least in principle.

mhofman:
No longer necessary for TS 4.5 update so reassigning.

kriskowal:
@Tartuffo No, this is orthogonal maintenance. The longer we put this off, the more time we waste on compatibility.

Tartuffo:
@kriskowal Let me put this another way - do you plan to do this in the Mainnet-1 timeframe?  If so, the only way to track that you are going to spend time on this is to put it into the "Mainnet 1" release.

kriskowal:
Roger, all debts paid in MN-1 are MN-1 debts. This somewhat changes how I classified issues in backlog grooming, so we’ll need to move some things from product backlog to the MN-1 column.

Tartuffo:
Yes, please do!

rekmarks:
As of today's date, even Node 16 is EOL. Is there any appetite for dropping all Node versions <18?
<img width="757" alt="image" src="https://github.com/endojs/endo/assets/25517051/ed33da74-fd38-4b87-9719-dbf13460e170">

ref: https://nodejs.org/en/about/previous-releases

kriskowal:
I would be open to entertaining a change that generally shifted our testing regimen to 18 and dropped support for 16. I would also like to begin a discipline of updating the `engines` field for `ses` at least to have both a minimum and maximum version.

kriskowal:
We definitely need to be that explicit for `ses`, but perhaps can be more open for others. I’m fine with forcing us to expressly opt in for new node versions. It does mean we become instantly irrelevant if we stop working, but that’s the safe option.

mhofman:
Apparently in most sane package managers the `engines` field is advisory only anyway

PR #1043, opened 2022-02-02
Update TypeScript to 4.5

mhofman:
Update TS to 4.5 to align with upcoming `agoric-sdk` change (see https://github.com/Agoric/agoric-sdk/pull/4420).

Make some types explicit any to prevent inference. Sprinkle some jsdoc here and there.

PR #1047, opened 2022-02-02
Bump minimum node 12 version

mhofman:
Typescript ESLint parser (see #1043) requires `node >=12.22`, and we're still testing with an old pinned node version (12.17) which is no longer necessary as `agoric-sdk` has moved on.

Also add node 16.x to the test matrix since `agoric-sdk` now tests on that as well.

Does not drop node 12 entirely as that's a breaking change. Refs #1046

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "packages/zip"
   ],
   "engines": {
-    "node": ">=12.17.0"
+    "node": ">=12.22.0"
   },
   "devDependencies": {
     "@octokit/core": "^3.4.0",
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,7 +72,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [12.17.0, 12.x, 14.x]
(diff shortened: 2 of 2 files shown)


michaelfig (approved):
LGTM!

package.json
```diff
@@ -37,7 +37,7 @@
     "packages/zip"
   ],
   "engines": {
-    "node": ">=12.17.0"
+    "node": ">=12.22.0"
```
kriskowal, line 40: This seems oddly specific and still in the 12 range in a package that only impacts developers. Maybe ^14 here?

kriskowal (commented):
We should probably also bump the engines support range in all-the-packages and maybe even mention it in the NEWS since this is a user-facing change.

kriskowal:
I see the limit to the scope in the description. I’m comfortable with a more aggressive stance. If we’re not testing 12, we should expressly drop support.

kriskowal (approved):
…and I will let you make the call as to whether to carry this change all the way to dropping support entirely and noting the breaking change.

mhofman:
Dropping this PR in favor of a limited bump included in #1043

Issue #1045, opened 2022-02-02
Seeing "impossible" error on master for typed array tests


erights:
A `yarn test` in endojs/endo master in `packages/ses`. Looks like a `configurable: true` somehow snuck in there for an indexed property on a typed array, which should be impossible. Or I'm reading the test wrong.

```
  make-hardener › harden typed arrays

  Difference:

    {
  -   configurable: false,
  +   configurable: true,
      enumerable: true,
      value: 0n,
      writable: true,
    }
```
and
```
  make-hardener › harden typed arrays and their expandos

  Difference:

    {
  -   configurable: false,
  +   configurable: true,
      enumerable: true,
      value: 0,
      writable: true,
    }
```

kriskowal:
@erights What version of Node.js are you using?

kriskowal:
You must be using 16. There’s a behavior difference in 14 and 12 that is probably explained by a V8 bugfix. I can for now relax the test to not check this property.

erights:
I'm using Node v16.13.1

PR #1048, opened 2022-02-02
fix(ses): Relax hardened typed array test to be insensitive to bugfix between Node.js 14 and 16

kriskowal:
Fixes: #1045


erights (approved):
Woof. Glad to see this was a v8 bug and that it was already fixed. Can we put a note or issue or TODO somewhere to fix the test once we're no longer dependent on Node 14?

LGTM

kriskowal:
I’ve left a note here https://github.com/endojs/endo/issues/1053

Issue #1056, opened 2022-02-03
[cjs] Avoid exposing the decision to move exports to `default` to the cjs module itself


naugtur:
see https://github.com/endojs/endo/blob/master/packages/compartment-mapper/src/parse-cjs.js#L33 

The frozen module object, when `exports` are set, puts them on `default` key meanwhile still remaining readable for the underlying cjs module. As a result, referencing exports may not always work as expected.

```
const a = {foo:1}
module.exports = a
module.exports.foo //undefined
module.exports.default.foo //1
``` 

Frozen `module` should probably have the public and internal facets or be wrapped before passing over to whoever imported it.

PR #1057, opened 2022-02-03
test(compartment-mapper): Failing test for CommonJS compatibility

naugtur:
Includes:

- test to cover #1056
- test for a package that happens to export something called 'default'
- test for using dynamic import inside a cjs module to get an esm module

   
----

To avoid clicking 5+ times on a PR with fixtures etc. add this as a bookmark to your bookmark toolbar and click the bookmark to load all:
```
javascript:void Array.from(document.querySelectorAll('.load-diff-button')).map(e=>e.click())
```


packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/app/index.js
```diff
@@ -0,0 +1,6 @@
+const callself = require('callself')
+try {
+  callself.makeTheCall()
+} catch (e) {
+  throw new Error('Referencing exported functions should be possible', { cause: e });
```
naugtur, line 5: BTW `cause` is supported starting node 16.9.0 but didn't show up for me in the output. Is Error here not the real Error?

packages/compartment-mapper/test/test-cjs-compat.js
```diff
@@ -0,0 +1,23 @@
+// import "./ses-lockdown.js";
+import 'ses';
```
naugtur, line 2: not sure why these are here, I cargo-culted them in.

kriskowal: This could be:

```js
import '@endo/init';
```

If we take a dependency on init, which we probably should now.

naugtur: switching the import explodes for me, I might look into it more.

For now, I got the best results doing 
```js
import '../../ses/index.js';
``` 
because I get nicer stacktraces

naugtur:
Closing, all tests from here, with improvements, are added to #1064

PR #1059, opened 2022-02-03
Fix CI to actually run type checks

### LINKED ISSUE #1058: Types linting not happening in CI
When #1011 landed, it broke types linting locally but all checks passed CI. While #1051 fixes the root cause, we still need to address the lack of CI failure.

The problem seem to be that the `lint:types` scripts use `tsc --build` for the incremental build, and for some reason that doesn't run correctly in CI, but does locally (possibly something to do with timestamps of checked out files). The fix is to switch to the `-p` option instead that doesn't do incremental build.

When trying this locally, there is type error in `lp32/reader`'s, regarding the generator function requiring a `return` statement because of the explicit return type. The problem is that the `Stream` return is parametrized with a `TReturn` of `undefined` instead of `void`.

mhofman:
Run lint type checks with `-p` instead of `--build`. Closes #1058 
Fix a type issue that arose in `lp32`

Remove Node 12.17 from the CI matrix. Bump minimum node version to `12.22` as that will be the minimum version supported by the upcoming TS 4.5 change (see #1043)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,7 +72,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [12.17.0, 12.x, 14.x]
+        node-version: [12.x, 14.x]
         platform: [ubuntu-latest, windows-latest]
 
     steps:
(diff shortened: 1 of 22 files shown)


kriskowal (approved):

kriskowal:
This may have to be rolled in atop your other TypeScript changes, as I’m seeing a type failure in CI around marshal as checked in daemon.

kriskowal (approved):

PR #1061, opened 2022-02-04
Remove pureCopy, ALLOW_IMPLICIT_REMOTABLES

erights:
We are now allowing ourselves to *assume* a protection we have not yet implemented --- that Passable pass-by-copy composites (CopyArray, CopyRecord, Tagged) cannot be proxies. Enforcing this would mirror the security properties of the proposed Records and Tuples, ensuring that these are only passive pure data that cannot cause side effects or interleave with user code, for example, to mount reentrancy attacks. (Need issue number where this assumption is documented.)

Given that assumption, we no longer need `pureCopy` nor the contract between `OnlyData` and `PureData`. The name `PureData` better suggests the strong property that we're allowing ourselves to assume. So we delete all uses of `pureCopy` and occurrences of `OnlyData` or replace them with `PureData`.

This PR also removes the long deprecated `ALLOW_IMPLICIT_REMOTABLES` flag / environment variable. We have long ago switched it off by default and have gone for a long time without any need to turn it back on. At this point, we won't break anything by removing it.

https://github.com/Agoric/agoric-sdk/pull/4458 should go first, deleting uses of `pureCopy` which endo will no longer export.

The only use of `sameValueZero` was in agoric-sdk, so between the two PRs, we're moving it there. The other PR should go first for this reason as well. This PR stops exporting it. https://github.com/Agoric/agoric-sdk/pull/4458 stops importing, instead defining its own.


erights:
https://github.com/Agoric/agoric-sdk/pull/4458 is merged, so this one should no longer be blocked on that one.

michaelfig (approved):
This is great!  Nice simplification.

PR #1062, opened 2022-02-04
fix(nat): declare `allegedNum` as `unknown`, then assert

michaelfig:
This change is necessary to allow callers to use `Nat` on data whose type they are unsure of.


erights (approved):
LGTM

I'm a bit surprised that you didn't need to change any of its uses in order to stay green.

kriskowal (approved):

michaelfig:
> I'm a bit surprised that you didn't need to change any of its uses in order to stay green.

This change was actually making `Nat` more tolerant of its inputs, so no existing uses had problems.

PR #1064, opened 2022-02-07
fix(compartment-mapper): change how parse-cjs.js treats exports to align with behavior of cjs in Node.js

naugtur:
Fix based on tests proposed in https://github.com/endojs/endo/pull/1057

Fixes https://github.com/endojs/endo/issues/1056  and some more.

----
- [x] Review fixes,
- [x] Clean up history,
- [x] CI green
- [x] Is this enough of a change to warrant a different commit message than `fix` for the semantic release?


packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/app/_generateExpected.mjs
```diff
@@ -0,0 +1,16 @@
+// run me with:
```
naugtur, line 1: this file doesn't run as a test. I can clean it up if documenting how I got the result is not worth keeping a separate file. 

Couldn't decide.

packages/compartment-mapper/test/test-cjs-compat.js
```diff
@@ -0,0 +1,42 @@
...
+  const fixture = new URL(
+    'fixtures-cjs-compat/node_modules/app/index.js',
+    import.meta.url,
+  ).toString();
+
+  const application = await loadLocation(read, fixture);
+  const {
+    namespace: { assertions },
+  } = await application.import({});
+  assertions.packageReferencingItself();
```
naugtur, line 24: :question: Would you rather see the assertions made inside the test itself? 

I didn't want to go through another layer of exporting and importing the same thing before checking, so I implemented cheching inside fixtures ([there's precedent!](https://github.com/endojs/endo/blob/3bbe1802cb789c975f3e190e234c061b6613742b/packages/compartment-mapper/test/fixtures-0/node_modules/app/main.js))

packages/compartment-mapper/src/parse-cjs.js
```diff
@@ -28,32 +28,42 @@ export const parseCjs = async (
   const execute = async (moduleExports, compartment, resolvedImports) => {
     const functor = compartment.evaluate(
       `(function (require, exports, module, __filename, __dirname) { ${source} //*/\n})\n//# sourceURL=${location}`,
+      { __evadeImportExpressionTest__: true },
```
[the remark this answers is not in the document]
naugtur: undone

packages/compartment-mapper/src/parse-cjs.js
```diff
@@ -25,35 +25,44 @@ export const parseCjs = async (
    * @param {Compartment} compartment
    * @param {Record<string, string>} resolvedImports
    */
-  const execute = async (moduleExports, compartment, resolvedImports) => {
```
naugtur, line 28: There doesn't seem to be a need for this to be async anywhere and we're risking unhandled rejections leaving it as-is. Removed async.

naugtur:
@kriskowal This is ready for a final review. I have no intention to add more scope to this PR.

packages/compartment-mapper/src/parse-cjs.js
```diff
@@ -25,35 +25,44 @@ export const parseCjs = async (
...
-
+    let exportsReferenceCopy = moduleExports;
     const module = freeze({
       get exports() {
-        return moduleExports;
+        return exportsReferenceCopy;
       },
       set exports(value) {
-        moduleExports.default = value;
+        exportsReferenceCopy = value;
```
naugtur, line 38: This will not make the `module.exports=something` become available as the default to other modules synchronously. 
Doesn't seem to be how they're using it though. I could try to come up with a test that breaks it.

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/app/_generateExpected.mjs
kriskowal: I like this here. The comment helps. More comment would be awesome.

kriskowal: The underscore in the name is unnecessary to my eye.

kriskowal: And I personally favor hyphens over camels in paths and URLs.

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/app/index.js
```diff
@@ -0,0 +1,48 @@
...
+    What we got     : ${whatWeGot}
+    What we expected: ${expect}`)
+  }
+}
+
+function stableOrderJson(obj) {
+  return JSON.stringify(obj, Object.keys(obj).sort());
+}
+
+
```
kriskowal, line 25: Our prettifier would probably collapse these lines. It might not be reaching here because of the `node_modules` path component. Maybe we can work-around that somehow without interfering with the test fixture shape, but I don’t know. Fixing the linter is out of scope for this change.

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/app/moduleinterops.js
```diff
@@ -0,0 +1,5 @@
+
+const a = { nameCollision:1 }
+module.exports.notDefault = 2
+module.exports.nameCollision = 3
+module.exports.default = a
```
kriskowal, line 5: `yarn lint-fix` would add semicolons and space after colon if it were playing along.

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/app/package.json
```diff
@@ -0,0 +1,12 @@
...
+  "name": "app",
+  "version": "1.0.0",
+  "main": "./main.js",
+  "dependencies": {
+    "callself": "^1.0.0",
+    "interops": "^1.0.0"
+  },
+  "devDependencies": {
+  
+  }
```
kriskowal, line 11: Can be omitted, I think.

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/callself/README.md
```diff
@@ -0,0 +1 @@
+This is a package that represents various ways in which `module.exports` and `exports` references could be used
```
kriskowal, line 1:
```suggestion
This is a package that represents various ways in which `module.exports` and `exports` references could be used.
```

packages/compartment-mapper/test/test-cjs-compat.js
kriskowal: This is fine.

kriskowal: For the back of the mind: it may begin to make sense to write these tests as test262 fixtures instead.

packages/compartment-mapper/test/test-cjs-compat.js
```diff
@@ -0,0 +1,42 @@
...
+  t.pass();
+});
```
kriskowal, line 26:
```suggestion
});

```

packages/ses/index.d.ts
```diff
@@ -194,6 +194,14 @@ export interface Assert {
   makeAssert: MakeAssert;
 }
```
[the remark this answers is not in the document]
kriskowal: As long as this is in a separate commit when it merges, it will have my blessing. I use a workflow like https://abhinavg.net/posts/restacking-branches/. If you prefer to squash this through the Github button, I would prefer this to be a separate PR.

kriskowal (approved):
The build is green and this looks like forward progress to me. Please consider my proposed nits, amendments, and advice below.

packages/compartment-mapper/src/parse-cjs.js
```diff
@@ -28,32 +28,42 @@ export const parseCjs = async (
...
     });
 
     const require = freeze((/** @type {string} */ importSpecifier) => {
       const namespace = compartment.importNow(resolvedImports[importSpecifier]);
       if (namespace.default !== undefined) {
-        return namespace.default;
+        // return namespace.default;
+
+        if (Object.keys(namespace).length > 1) {
+          return { ...namespace.default, ...namespace }; // this resembles Node's behavior more closely
```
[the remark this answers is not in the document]
naugtur: I looked at what node does and it's actually covered by the test case with the name collision between export and default field. This is the closest I got to reproducing node's behavior.
How would order of export affect this? this code seems to only be running once per namespace, so the content is final. the only thing that happens is top-level fields override the fields from default if they exist.

I might be wrong, let's see how this works on more realistic examples in tests.

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/app/_generateExpected.mjs
```diff
@@ -0,0 +1,16 @@
...
+console.log(
+    stableOrderJson(interops)
+)
+console.log(
+    stableOrderJson(moduleinterops)
+)
+
+
+function stableOrderJson(obj) {
+    return JSON.stringify(obj, Object.keys(obj).sort());
```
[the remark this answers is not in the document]
naugtur: I thought so too, but it happens to sort the contents of obj.default too. Probably be because the keys are exactly the same.
My first reflex was to use substack's `json-stable-stringify` but none of the tests seem to pull in real dependencies.

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/app/index.js
```diff
@@ -0,0 +1,48 @@
+const callself = require('callself')
+const interops = require('interops')
+const moduleinterops = require('./moduleinterops')
+
+// generated by running this file with console.error(stableOrderJson(interops))
```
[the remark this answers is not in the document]
naugtur: definitely!

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/app/_generateExpected.mjs
naugtur: `_` is there to indicate the file is not part of the fixture. After years of using almost opposite conventions for work and private projects I just agree to anything, so consider it done.

 I'm upset that there's a file to generate this, but the other example is only documented with a comment in index.js - I'll unify that one way or the other.

packages/ses/index.d.ts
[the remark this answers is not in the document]
naugtur: My plan was to do an interactive rebase and squash some commits (especially the back-and-forth changes) to leave a cleaner history and then merge, no squash

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/app/_generateExpected.mjs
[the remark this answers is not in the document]
naugtur: Got a decent version done in acceptable time.

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/app/index.js
naugtur: It was less work than finding what to fix

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/app/moduleinterops.js
naugtur: I got eslint running on fixtures too, but it had some destructive opinions. Also, didn't fix much. Prettier will do for now.

what it took to run eslint:
```
    "lint-fixtures": "find ./test/fixtures-*/**/*.js | xargs eslint --no-ignore",
    "lint-fixtures": "find ./test/fixtures-*/** -name '*.js' -exec eslint --no-ignore {} \\;",
```

packages/ses/index.d.ts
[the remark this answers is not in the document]
naugtur: Done, down to 3 meaningful commits

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/app/assertion-generators/json.cjs
```diff
@@ -0,0 +1,16 @@
+// Not guaranteed to sort all array elements and weird keys properly, but definitely good enough for the usecase
+module.exports = function stableOrderJson(obj) {
+  return JSON.stringify(obj, (_, v) => {
+    if (v && v.constructor === Object) {
```
naugtur, line 4: couldn't do `v?.` because node 12. my golf score suffers

naugtur:
This is now ready to merge.  
I'm leaving it open to confirm `fix` is the right conventional commit tag for the change.  
@kriskowal It could be considered a breaking change for someone, but did we make any guarantees about cjs support?

packages/compartment-mapper/src/parse-cjs.js
```diff
@@ -25,35 +25,44 @@ export const parseCjs = async (
...
-        moduleExports.default = value;
+        exportsReferenceCopy = value;
       },
     });
 
     const require = freeze((/** @type {string} */ importSpecifier) => {
       const namespace = compartment.importNow(resolvedImports[importSpecifier]);
       if (namespace.default !== undefined) {
-        return namespace.default;
+        // return namespace.default;
```
kumavis, line 45: commented code

naugtur: how did that get there? I blame git ;)

kriskowal:
CommonJS support is sufficiently broken that I can be certain that nobody is depending on its current brand of brokenness, so we can just call this a fix.

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/app/assertion-generators/json.cjs
kriskowal: I’ve dropped Node.js 12 from CI if that makes a difference.

kriskowal: The idiom we favor here is `Array.isArray(v)` then `Object(v) === v` for the isObject test.

https://github.com/endojs/endo/blob/master/packages/ses/src/commons.js#L233-L250

packages/compartment-mapper/src/parse-cjs.js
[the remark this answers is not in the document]
kriskowal: I’ll buy the argument and keep an eye open for a counter-example.

Issue #1036, opened 2022-02-02
Endo Daemon Worker


kriskowal:
Create a preliminary private API for spawning a worker in Endo. Whomever spawns the worker needs a handle to the worker control facet, which will at least have a `terminate` method and a place to put other methods or properties for getting at the rest of the worker public API, which remains up for design. This will be the basis for Endo Application workers, and will get there either through refactoring or layering.

kriskowal:
This has landed in the `endo` feature branch.

PR #1067, opened 2022-02-09
feat(daemon): Add worker with lifecycle

kriskowal:
This change takes a number of steps toward spawning a worker which currently does nothing except support graceful termination. Commits are individually reviewable.

Refs #1036

The first change reverts eventual-messages for streams. #1028 because this unexpectedly causes data-lock for the Endo daemon. This might be because of a layering issue: the stream in question is carrying CapTP. It might be possible to recover the feature manually, but was probably unwise to lean on eventual-send anyway, on account of the layering.

The subsequent commits each make some surgical change necessary to explore outward to endo daemon workers.

In subsequent changes, this initial worker will grow the ability to run arbitrary programs under hardened JavaScript with limited access to user powers. But first, each new worker will need to initialize some plugins that provide those powers.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,7 +70,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [12.x, 14.x]
+        node-version: [14.x]
         platform: [ubuntu-latest, windows-latest]
 
     steps:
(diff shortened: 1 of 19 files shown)


michaelfig (approved):
That was a lot!  I have enough of a glimmer that I think I could fix problems if they come up, so LGTM!

kriskowal:
A lot of moving things around for very little impact, indeed. Onward!

PR #1068, opened 2022-02-09
chore: Drop Node.js 12 in CI

### LINKED ISSUE #1046: Drop Node.js 12 support in CI

kriskowal:
Refs #1046

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,7 +70,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [12.x, 14.x]
+        node-version: [14.x]
         platform: [ubuntu-latest, windows-latest]
 
     steps:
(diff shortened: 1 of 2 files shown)


kriskowal:
I’ve elected, for now, not to poke the `"engines"` bear and formally decree a breaking change.

mhofman (approved):
> I’ve elected, for now, not to poke the `"engines"` bear and formally decree a breaking change.

Let's create an issue for that so we don't forget.

kriskowal:
Changed fixes to refs #1046 for tracking the engines update.

PR #1070, opened 2022-02-10
test(cjs-module-analyzer): Failing tests

naugtur:


naugtur:
These capabilities of the parser are not necessary for exports to work

PR #1071, opened 2022-02-12
style(eslint-config): allow globals that SES makes powerless

turadg:


packages/eslint-config/eslint-config.json
```diff
@@ -14,17 +14,59 @@
...
     "Compartment": "readonly",
     "StaticModuleRecord": "readonly",
     "TextDecoder": "readonly",
     "TextEncoder": "readonly",
     "URL": "readonly",
-    "URLSearchParams": "readonly"
+    "URLSearchParams": "readonly",
+
+    // Allow what SES makes powerless, copied from its whitelist
+    // *** Constructor Properties of the Global Object
```
dckc, line 27: JS comments in a JSON file? If this is read by something other than a JSON parser, note that at the top, maybe?

kriskowal (approved):
I too am a JSON comments skeptic. Perhaps we can describe the design choices in README.md. Looks good when lint is green.

turadg:
*Config comments*

I also prefer not to put comments in `.json` files but I see the value of having comments on config choices as far outweighing that. ESLint lets you express the rules in JS, JSONC or YAML. I assert YAML is the best because:
- most readable (vs JS, JSON)
- allows comments (vs JSON)
- does not allow arbitrary code (vs JS)

I'd be happy to follow up with a PR to change the config format to YAML.

*Maintenance of the sync between esilnt config and the SES whitelist*

I agree it's a good idea. I see it as out of scope for this change and something that could be prioritized, though I don't anticipate it being a high priority because the list changes very rarely and the consequences for it getting out of sync are very small.

kriskowal:
Though YAML is awful, I’m okay with YAML for this purpose. @michaelfig is also a stake-holder in that decision.

michaelfig:
> I’m okay with YAML for this purpose. @michaelfig is also a stake-holder in that decision.

As long as I'm not maintaining it, I don't have a preference.

turadg:
YAML conversion PR ^

While looking for examples of ESLint configs, I noticed ESlint projects defines globals like this https://github.com/eslint/eslint/blob/main/conf/globals.js

Also with Lerna on this monorepo we have a build step so maybe building the whitelist from SES wouldn't be as much work as I thought. Still putting this down for now. :)

PR #1074, opened 2022-02-14
fix(ses): make import * and default from cjs wire up correctly

naugtur:
TODO:
- [ ] increase coverage of module-instance.js at least back to the original


packages/ses/src/module-instance.js
```diff
@@ -73,6 +73,41 @@ export const makeThirdPartyModuleInstance = (
         update(value);
       };
     });
+    // Default was not in exports, need an implicit default
```
naugtur, line 76: imports function wants it, but I'm not sure if adding the default will not produce artifacts elsewhere. Tests seem to be fine.

naugtur:
I was hoping to improve coverage more, but don't know how to reach those yet:
Screenshot 2022-03-02 at 14-50-13 Code coverage report for ses src module-instance js

packages/ses/src/module-instance.js
```diff
@@ -73,6 +73,41 @@ export const makeThirdPartyModuleInstance = (
...
+      defineProperty(proxiedExports, 'default', {
+        get,
+        set,
+        enumerable: true,
+        configurable: false,
+      });
+
+      notifiers.default = update => {
+        arrayPush(updaters, update);
+        update(get()); // We want it to default to proxiedExports until value is set. Furhter updates will only happen if value was set, so get() only needs to be used on initial update
```
kriskowal, line 104:
```suggestion
        update(get()); // We want it to default to proxiedExports until value is set. Further updates will only happen if value was set, so get() only needs to be used on initial update
```

packages/ses/src/module-instance.js
```diff
@@ -73,6 +73,41 @@ export const makeThirdPartyModuleInstance = (
...
+      };
+    }
+    // This is enough to support import * from cjs - the '*' field doesn't need to be in exports nor proxiedExports because import will only ever access it via notifiers
+    notifiers['*'] = update => {
+      update(proxiedExports);
+    };
```
kriskowal, line 110: This is good.

packages/ses/src/module-instance.js
```diff
@@ -73,6 +73,41 @@ export const makeThirdPartyModuleInstance = (
         update(value);
       };
     });
+    // Default was not in exports, need an implicit default
+    if (!notifiers.default) {
```
kriskowal, line 77:
```suggestion
    // For third-party module records that do not export a `default` name,
    // we infer that the default export is the namespace object itself.
    if (!notifiers.default) {
```

As I write this comment, I wonder whether it’s better or possible to push this concern to the CommonJS module record, forcing it to provide a `default` name, but only if there’s no explicit `default` export. I believe it’s valid for a module to not have an exported variable named `default`.

I suppose the litmus test for whether to include this special behavior here would be: is this implementation equally valid for WASM? Could it possibly interfere with intentional behavior of a module? Could it encourage dependees to depend on the non-existence of an export named `default`, and if they did, would that make their code more accidentally fragile? Is adding a `default` export to a WASM or CommonJS module a breaking change?

kriskowal (commented):
Tentatively approving this change. Apart from nits, please feel free to land the change if you have satisfying answers to my questions about the implicit `default` export.

kriskowal (approved):

packages/ses/src/module-instance.js
naugtur: In import from cjs running prettier can be a breaking change thanks to lexer. I get the concern and I think the ecosystem is going to deal with it in certain ways we'll need to accommodate. When I posted about it people brought up " the infamous triplet" which makes sure the consumer gets the right reference no matter how they import it. It makes a mess, adds a default field visible in cjs exports and so on.

Moving default adding to parse-cjs is problematic because to make it wire up correctly of have to declare it early on because at execution time it's already to late. And that would make the synthetic default appear in cjs require cjs too. At least my mental model says so.

The way esm cjs interop works AFAIR there's no way to not have a default so adding an explicit default +/- __esModule is something to dive into via the compat table.

Looking at e2e tests this seems to be working fine most of the time. I think I saw one that maybe produces an unexpected default where it wasn't supposed to be, I'll check it more thoroughly.  


Having said that, I'll happily experiment with wasm to see if this has potential to interfere.
Any prior at on loading wasm in Endo?

naugtur: Ok, so I implemented wasm support in compartment mapper and the implicit default does show up unnecessarily. 
Let's see if I can fix it today.

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/parser-struggles/main.js
```diff
@@ -10,6 +10,7 @@ function nothing() {
 
 // TODO: fix the export - .default should not be here
 function notMuch(item) {
+  //  throw Error(JSON.stringify(require('./is-ok')))
```
kriskowal, line 13: debug line?

packages/compartment-mapper/test/scaffold.js
```diff
@@ -73,7 +73,7 @@ export function scaffold(
     });
   };
 
-  wrap(test)(`${name} / loadLocation`, async t => {
+  wrap(test.only)(`${name} / loadLocation`, async t => {
```
kriskowal, line 76: debug line?

packages/compartment-mapper/test/fixtures-0/node_modules/typecommon/main.js
```diff
@@ -3,4 +3,4 @@ const mjs = require('./mjs.mjs');
 const cjs = require('./cjs.cjs');
 const json = require('./json.json');
 
-module.exports = [js, mjs, cjs, json];
+module.exports = [js, mjs, cjs.default, json];
```
kriskowal, line 6: Ah, the old test case was definitely invalid. CommonJS importing CommonJS should not treat `default` as special! Great change.

Do we also test `module.exports = 42`? Changing `cjs.cjs` to do this instead might be more in the spirit of this test, but also I want to capture the case of a CommonJS uses `default` as an exported property name like you do here.

kriskowal (approved):
Affirming without reservations. Thanks!

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/parser-struggles/main.js
naugtur: yup. the fastest way to get output from code running in tests in endo :D

packages/compartment-mapper/test/fixtures-0/node_modules/typecommon/main.js
naugtur: I just wanted to make it pass. I'll do both to cover more regression possibilities

packages/compartment-mapper/test/scaffold.js
naugtur: now you know why I did wrap(test) instead of creating my own test function that calls test inside ;)

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/app/assertion-generators/json.cjs
```diff
@@ -1,16 +1,16 @@
 // Not guaranteed to sort all array elements and weird keys properly, but definitely good enough for the usecase
 module.exports = function stableOrderJson(obj) {
   return JSON.stringify(obj, (_, v) => {
-    if (v && v.constructor === Object) {
+    if (v && v.constructor === Array) {
+      return v.sort();
+    }
+    if (v && typeof v === 'object') {
```
naugtur, line 7: in Endo:
```
const namespace = require('something')
namespace.constructor === undefined
```

I think it's a problem

naugtur:
This is now ready to merge. I'll hit the button in the evening.

packages/compartment-mapper/src/parse-cjs.js
```diff
@@ -31,16 +51,32 @@ export const parseCjs = async (
...
           return moduleEnvironmentRecord[prop] || target[prop];
         },
-        set(target, prop, value) {
+        set(_target, prop, value) {
           moduleEnvironmentRecord[prop] = value;
           return true;
         },
+        defineProperty(_target, prop, descriptor) {
+          // Object.defineProperty(moduleEnvironmentRecord, prop, descriptor) would work if someone in the npm
+          // community didn't come up with the idea to redefine property in the getter.
```
kriskowal, line 66: We can be aggressively specific here. 😇

PR #1076, opened 2022-02-15
fix: Type definitions cannot overshadow

kriskowal:
Evidently, a type definition file replaces the implementation JavaScript file in the types space, so any reference to the implementation is a cycle, and any value participating in that cycle silently degrades to the any type. Consequently, the type information disappears for any external dependency.

In this change, I’ve deleted superfluous type definition files that I had previously thought necessary to expressly export types from JSDoc modules.


packages/compartment-mapper/src/parse-archive-cjs.js
```diff
@@ -1,13 +1,13 @@
 // @ts-check
-
-/** @typedef {import('ses').ThirdPartyStaticModuleInterface} ThirdPartyStaticModuleInterface */
+/// <reference types="ses"/>
```
mhofman, line 2: Interesting. I'd love if we could switch the base `ses` away from ambient types.

packages/ses-ava/src/types.js
```diff
@@ -57,7 +57,7 @@
  * @param {ImplFunc} implFunc
  * @returns {void}
  *
- * @typedef {TesterFunc} TesterInterface
+ * @typedef {Object} TesterProperties
```
mhofman, line 60: Would it work with `@callback` ?

packages/zip/src/exported.js
```diff
@@ -1,3 +1,7 @@
+// @ts-check
+
+export {};
```
mhofman, line 3: If switching to exported types, can we try to add the following to `zip/index.js` so that we can use import the types directly using package name?
```js
export * from './src/exported.js';
```

mhofman (changes requested):
> Evidently, a type definition file replaces the implementation JavaScript file in the types space, so any reference to the implementation is a cycle, and any value participating in that cycle silently degrades to the any type. Consequently, the type information disappears for any external dependency.

Yes! That is the single most important thing to remember with `.d.ts` files.

Some nits mostly, but I'd like to review the situation about what types gets exported and frfom where in the `zip` package.

packages/netstring/reader.js
```diff
@@ -67,6 +67,8 @@ async function* makeNetstringIterator(
       `Unexpected dangling message at offset ${offset} of ${name}`,
     );
   }
+
+  return undefined;
```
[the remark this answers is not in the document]
kriskowal: I’m not sure about prevention, but I have encountered difficulty and backed off to `undefined` by default for all the templates of a stream. The problem arises from the symmetry of the type. Void is great for a type that will be returned and can be ignored, where unknown seems to work better for an argument, but with pipe and pump, args become returns and returns become args. But, this is something I want to improve upon when I understand the base principles and emergent effects better. For example, it would be nice to have overloads that allow an argument of type undefined to be omitted more generally.

packages/daemon/src/connection.js
```diff
@@ -30,7 +30,7 @@ const makeCapTPWithStreams = (name, writer, reader, cancelled, bootstrap) => {
     }
   })();
 
-  const closed = cancelled.finally(async () => {
+  const closed = cancelled.catch(async () => {
```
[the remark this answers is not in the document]
kriskowal: I lost track of the chain of effects that led me here. It did start with ironing out types, but while I was there, I realized I’d misused the cancellation pattern here. The `cancelled` promise must be rejected, not fulfilled, but the rejection shouldn’t propagate to `closed`.

packages/ses-ava/src/ses-ava-test.js
```diff
@@ -110,8 +110,10 @@ const wrapTester = (testerFunc, logger = defaultLogger) => {
  * @returns {TesterInterface}
  */
 const wrapTest = (avaTest, logger = defaultLogger) => {
-  /** @type {TesterInterface} */
-  const testerWrapper = wrapTester(avaTest, logger);
+  const testerWrapper = /** @type {TesterInterface} */ (wrapTester(
```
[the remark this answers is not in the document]
kriskowal: It’s not avoidable in this case. The only reason the former pattern was allowed was because this file wasn’t properly covered by the type checker.

packages/ses-ava/src/types.js
kriskowal: Oddly enough, it seems we can have `@callback` with both `@param` and `@property` fields and that does seem to work.

packages/ses-ava/src/ses-ava-test.js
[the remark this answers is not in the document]
erights: It's always avoidable by using `@ts-ignore` or `@ts-expect-error` instead, right?

Aside from that, if it is unavoidable, how come the before code (code in the left pane) doesn't need one? It locally looks otherwise identical.

packages/zip/src/writer.js
```diff
@@ -50,15 +50,15 @@ export class ZipWriter {
 }
 
 /**
- * @returns {ArchiveWriter}
+ * @returns {import('./exported.js').ArchiveWriter}
```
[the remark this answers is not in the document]
kriskowal: I elected to fold everything into types.js and export them all. I didn’t see any type that we should keep to ourselves.

packages/compartment-mapper/src/parse-archive-cjs.js
kriskowal: Base SES provides certain ambient types like `Compartment`, `harden`, and `lockdown`, but ThirdPartyStaticModuleInterface is not ambient. I was able to remove this line.

packages/ses-ava/src/ses-ava-test.js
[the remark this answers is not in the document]
mhofman: Because before this PR, the result of `wrapTester` was likely `any`, so assigning that to something of a stricter type was ok. Now I'm not sure what the return type of `wrapTester` is (it's possible it could be better typed to avoid this issue), but it seem to be incompatible with `TesterInterface`.

adding a `ts-ignore` would basically make `testerWrapper` typed as `any` so I consider it inferior. It also may silence multiple errors.

packages/zip/package.json
```diff
@@ -19,6 +19,7 @@
   },
   "type": "module",
   "main": "./index.js",
+  "types": "./index.d.ts",
```
mhofman, line 22: Why is this coming back? Can't we add a single `export * from './src/types.js';` in `index.js` (not `.d.ts`)? That's what I did for `marshal` and it seems to work fine.

packages/zip/src/format-reader.js
```diff
@@ -2,6 +2,8 @@
 /* eslint no-bitwise: ["off"], max-lines: ["off"] */
 
 /**
+ * @typedef {import('./types.js').ArchiveHeaders} ArchiveHeaders
```
mhofman, line 5: Can we add typedefs for the other `import('./types.js')` with duplicates?

packages/ses-ava/src/ses-ava-test.js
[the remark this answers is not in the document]
kriskowal: A `@ts-ignore` comment does suppress the error on the line, and I believe that makes `testerWrapper`’s type `any`, which disables any internal coherence checking from that point to the end of the function. If we did not also have an explicit `@returns` pragma, the `any` type would propagate to the return type.

I believe the difference between before and after here is that in the before case, `testerWrapper` is being labeled as a `TesterInterface` but the return type of `wrapTester` is `TestFunc`, which is not compatible. In the after case, we are forcing the result of `wrapTester` from `TestFunc` to `TesterInterface`, which makes the const `testerWrapper` a `TesterInterface` by inference.

mhofman (commented):
Wondering if we can't do away with the `index.d.ts` file which seems to be a footgun.

packages/ses-ava/src/types.js
kriskowal: Ah, action at a distance strikes again. Using a callback pragma works locally but fails in static-module-record where I’m using `test.failing`.

```
Error: test/test-static-module-record.js(201,6): error TS2339: Property 'failing' does not exist on type 'TesterInterface'.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Reverting.

mhofman: Bummer. I remember @dckc running into issues with `@callback` too.

kriskowal:
> Wondering if we can't do away with the `index.d.ts` file which seems to be a footgun.

Oh, great. I misunderstood your suggestion. I’ll try this out.

kriskowal:
Added a commit to make jsconfig.json’s more consistent and less brittle.

PR #1077, opened 2022-02-15
Adds some missing hardens

erights:
See
https://github.com/Agoric/agoric-sdk/blob/5e9c02a24cb3fadb8c67279f8183cac8d7cd03b1/packages/run-protocol/src/collect.js#L57-L67

Just fixes the ones I encountered. I did no systemic search of any kind to detect more such violations.

### DIFF
--- a/packages/bundle-source/src/index.js
+++ b/packages/bundle-source/src/index.js
@@ -151,7 +151,7 @@ async function bundleZipBase64(startFilename, dev, powers = {}) {
     },
   });
   const endoZipBase64 = encodeBase64(bytes);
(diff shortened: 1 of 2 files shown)


turadg (approved):

kriskowal (approved):
It would be worthwhile to also fix line 385 of the same file.

https://github.com/endojs/endo/pull/1077/files#diff-fdaf8fc23eb5f1cdbc9d1052c8b0d0c7b5084d378ffbd288eabf5e22c6b266b9L385

erights:
> It would be worthwhile to also fix line 385 of the same file.

Done.

PR #1080, opened 2022-02-16
fix: Make scripts/set-versions suitable for external usage

kriskowal:
This change makes it possible to, for example from an Agoric dapp, use `~/endo/scripts/get-versions.sh ~/endo | ~/endo/scripts/set-versions.sh`, to update the versions of Endo dependencies to their most up-to-date versions in a local copy of Endo in the dapp's workspace, including dependencies at the top-level of Endo, or `~/endo/scripts/all-versions.sh ~/endo | ~/endo/scripts/set-versions.sh` (*all-versions*) to sync transitive dependencies.


michaelfig (approved):
Don't see any obvious problems.

Issue #601, opened 2021-03-06
Compartment mapper: validate compartment map JSON


kriskowal:
Currently, the compartment mapper takes for granted that the compartment map is a valid instance of its purported schema. TypeScript calls us out on this, so there’s a forced cast in place in `import-archive.js` and a note to write a TS compliant validator to visit and assert the entire JSON shape.

https://github.com/Agoric/SES-shim/pull/599#discussion_r588521471

Tartuffo:
@kriskowal This does not have an area label that is covered by our weekly tech / planning meetings.  Can you assign the proper label?  We cover: agd, agoric-cosmos, amm, core economy, cosmic-swingset, endo, getrun, governance, installation-bundling, metering, run-protocol, staking, swingset, swingset-runner, token economy, wallet, zoe contract,

PR #1085, opened 2022-02-18
feat(compartment-mapper): Validate compartment maps

kriskowal:
Fixes: https://github.com/endojs/endo/issues/601

This change adds validation for compartment map input from archives.

This design sacrifices schema evolution for a tight correspondence between the content of a compartment map and its consistent hash. The validator allows for nothing extra. This may become a hindrance, in which case we would need to relax the extra property assertions and closed ranges for some types. For now, this is conservative in order to provide the maximum assurance to the authors.


packages/compartment-mapper/src/import-archive.js
```diff
@@ -164,12 +165,11 @@ export const parseArchive = async (
   }
 
   const compartmentMapText = textDecoder.decode(compartmentMapBytes);
-  const compartmentMap = /** @type {CompartmentMapDescriptor} */ (parseLocatedJson(
+  const compartmentMap = parseLocatedJson(
```
warner, line 168: I might have asked this before, but do our specs say anything about repeated keys in JSON? That would be a way for two different readers to interpret the same compartment map in divergent ways. If TC39 doesn't have a stance, we might want to add a note to the docs around this hashing/security stuff, that correctness depends upon consistent behavior of everybody's `JSON.parse`.

packages/compartment-mapper/src/compartment-map.js
```diff
@@ -0,0 +1,418 @@
...
+const assertScope = (allegedScope, path, url) => {
+  const scope = Object(allegedScope);
+  assert(
+    allegedScope === scope && !Array.isArray(scope),
+    `${path} must be an object, got ${allegedScope} in ${q(url)}`,
+  );
+
+  const { compartment, ...extra } = scope;
+  assert(
+    Object.keys(extra).length === 0,
```
warner, line 216: I'm tempted to suggest this `assertNoExtra(extra, explanation)` pattern be factored out so the calls fit on a single line, but I think I'm just feeling grumpy towards `prettier`'s verbosity today. The work this file is doing is completely straightforward.

warner (approved):
Looks good

packages/compartment-mapper/src/import-archive.js
kriskowal: JSON (ECMA 404) has no position on repeated keys, but JS (ECMA 262) takes the position that JSON gets evaluated as a JS expression if it parses according to 404, such that the last property defined with the one key wins. So, as long as everyone uses JS, they’ll arrive at the same interpretation.

PR #1086, opened 2022-02-18
feat(compartment-mapper): parseArchive return hash

kriskowal:

The compartment mapper will now compute and return the hash for an arbitrary archive when given a `computeSha512` option. Previously, parseArchive could only be used to validate the `expectedSha512`.


packages/compartment-mapper/src/import-archive.js
```diff
@@ -205,7 +208,7 @@ export const parseArchive = async (
     return compartment.import(moduleSpecifier);
   };
 
-  return { import: execute };
+  return { import: execute, sha512 };
```
warner, line 211: Am I correct in believing that when I call `{ import, sha512 } = await parseArchive(archiveBytes)`, that I'll learn the computed hash of the compartment map without running the risk of the archive executing any code or starting any infinite loops? (as long as I refrain from calling `import()`, that is).

If so, I'm not sure where the docs would live, but I'd love to be told about the safety properties I can rely upon.

warner (approved):
looks good

packages/compartment-mapper/src/import-archive.js
kriskowal: I’ve added a line to `README` to make that more explicit. It currently states that `parseArchive` and `loadArchive` defer execution until `import`, but did not make clear that you might never `import`.

PR #1089, opened 2022-02-19
chore: release

kriskowal:
- chore: lerna version
- chore: Update release notes


mhofman (approved):
LGTM

Out of curiosity, why does lerna bump every package version, even if only dev dependencies get updated

kriskowal:
> LGTM
> 
> Out of curiosity, why does lerna bump every package version, even if only dev dependencies get updated

I wonder the same. The bit we had to fix with a patch was getting it to pick the right version number in the `^0.x` range.

I think this behavior of making patches to raise the floor for transitive dependencies might cause dependees to tend to have coherent cross-tested versions of all the workspace dependencies, which probably papers over accidental breaking changes.

PR #1090, opened 2022-02-21
Improve `require` statement detection and postpone missing package errors

### LINKED ISSUE #1083: cjs-module-analyzer doesn't find requires if they're not top-level
```js
const { requires } = analyzeCommonJS(`
  mime(require('./types/standard'), require('./types/other'));
  `);
```

`requires` is an empty array.

Works fine if `require` call statements are not arguments to a function

Attempted fix:  
https://github.com/endojs/endo/commit/2703569cd663040c7cb1f3efb7b786ed1d6249d6

The change causes any `require()` call to produce a `requires` entry, which may lead to 
```
 Cannot find file for internal module "./niematakiegonumeru" (with candidates "./niematakiegonumeru", "./niematakiegonumeru.js", "./niematakiegonumeru/index.js") in package file:///...
 ```
 Which in turn could be solved with a change here:  https://github.com/endojs/endo/blob/a7b42ae2388b232f7daa099495ba11f385010fd1/packages/compartment-mapper/src/import-hook.js#L160-L175 
Instead of erroring out, returning a dummy static module record that throws the error when execute is called could work. That way it'd emulate cjs more closely and not fail if the `require` reference was not the actual require.

### PR DESCRIPTION
**Pull request by naugtur**
Fix #1083

### CODE REVIEW

File: packages/compartment-mapper/test/test-main.js [hunk 1]
```diff
@@ -92,7 +92,7 @@ test('no dev dependencies', async t => {
       });
     },
     {
-      message: /Cannot find external module "typecommon"/,
+      message: /Cannot find external module "typemodule"/,
```
naugtur, line 95: I doublechecked - it's just a matter of different order. If I remove other dev deps and their mentions, I do get the typecommon package error again.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -37,6 +37,26 @@ const has = (haystack, needle) => apply(hasOwnProperty, haystack, [needle]);
  */
 const resolveLocation = (rel, abs) => new URL(rel, abs).toString();
 
+/**
+ * @param {Error} error - error to throw on execute
+ * @returns {StaticModuleType}
+ */
+const postponeErrorToExecute = error => {
```
kriskowal, line 44: I like “postpone”. You might like “defer” better. Your choice, regardless.

packages/compartment-mapper/test/test-cjs-compat.js
```diff
@@ -33,11 +32,25 @@ test('CommonJS exports include a field named `default`', async t => {
...
+    'fixtures-cjs-compat/node_modules/app/index.js',
+    import.meta.url,
+  ).toString();
+
+  const {
+    namespace: { assertions },
+  } = await importLocation(read, fixture);
+  assertions.parserStruggles();
+  t.pass();
+});
```
kriskowal, line 56: Please frame this test in terms of `scaffold.js` as seen in `test-main.js` so we can verify that the same behaviors survive a round-trip through an archive.

kriskowal (changes requested):
The only important request I have is that we test a round-trip through an archive as well as import/load. That can be done using the `scaffold`, which generates tests for all of the different workflows starting with a fixture.

packages/compartment-mapper/test/test-main.js
naugtur: Sounded great at first, but then I thought - until we have much tighter feature coverage for cjs support, a test that may fail on changes in order of things happening could be valuable. It took me 5 minutes to figure out the behavior is still correct and I got a nice signal from it.

I'll add a TOO comment to relax this assertion later, if you don't mind.

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/parser-struggles/main.js
```diff
@@ -0,0 +1,16 @@
+function nothing() {
+  const require = () => {};
+  require('spam', 'asifitsdoingsomething');
+  require('./spam');
+//   require('/spam'); // TODO: prevent this from causing errors
```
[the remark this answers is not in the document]
naugtur: I mean to defer the error in a similar fashion to what I already did here. 
So that it happens, but in execution, not earlier. 

All I intend to prevent is a userspace function that happens to be called require tripping up loading. 
And I don't even know if that's a reasonable thing to fix yet, gotta find where the error originates from.

packages/compartment-mapper/src/import-hook.js
naugtur: Postpone feels to me like a more specific point in time than defer. I might be wrong. But I considered both and went for postpone as IMO better describing what's happening.

packages/compartment-mapper/test/test-cjs-compat.js
naugtur: It messes with import-archive. the staticModuleRecord stub returned by postpone is not enough to satisfy its import hook.

naugtur: I fixed bits missing in parse-pre-cjs to get archive working on basic tests, but I don't see how I could get the postponeError functionality to work for archive creation.  
Might have to build a better mental model for the archive build+import flow. Or postpone as far as archive runtime.

kriskowal: For postpone to work in an archive, we would need a module record that can be serialized to one of the valid formats and produce the same effect. It does not need to be stack-preserving.

naugtur: My current draft implementation that successfully postpones errors defines this:
```
 * @typedef {Object} ModuleSource
 * @property {string} [deferredError] - module loading error deferred to later stage
 ...
 `

File: packages/compartment-mapper/src/compartment-map.js [hunk 5]
```diff
@@ -140,13 +140,15 @@ const assertModule = (allegedModule, path, url) => {
...
-  const { compartment, module, location, parser, exit } = moduleDescriptor;
+  const { compartment, module, location, parser, exit, deferredError } = moduleDescriptor;
   if (compartment !== undefined || module !== undefined) {
     assertCompartmentModule(moduleDescriptor, path, url);
   } else if (location !== undefined || parser !== undefined) {
     assertFileModule(moduleDescriptor, path, url);
   } else if (exit !== undefined) {
     assertExitModule(moduleDescriptor, path, url);
+  } else if (deferredError !== undefined) {
+    //it's ok, trust me
```
kriskowal, line 151: Is it a string?

packages/compartment-mapper/src/archive.js
```diff
@@ -105,6 +105,10 @@ const translateCompartmentMap = (compartments, sources, renames) => {
           modules[name] = {
             exit,
           };
+        } else if (deferredError !== undefined) {
+          modules[name] = {
+            deferredError,
+          };
```
kriskowal, line 111: I’m okay with `error` for the name.

packages/compartment-mapper/src/import-archive.js
```diff
@@ -30,13 +31,38 @@ const q = JSON.stringify;
 
 const textDecoder = new TextDecoder();
 
+const freeze = Object.freeze;
```
kriskowal, line 34: Prevailing style is `const { freeze } = Object` unless you need a TypeScript JSDoc (which you might need to get tsc off your back about ReadOnly!)

packages/compartment-mapper/src/import-hook.js
```diff
@@ -38,23 +39,31 @@ const has = (haystack, needle) => apply(hasOwnProperty, haystack, [needle]);
 const resolveLocation = (rel, abs) => new URL(rel, abs).toString();
 
 /**
+ * @param {string} specifier
  * @param {Error} error - error to throw on execute
+ * @param {CompartmentSources} packageSources
  * @returns {StaticModuleType}
  */
-const postponeErrorToExecute = error => {
+const deferError = (specifier, error, packageSources) => {
```
kriskowal, line 47: Looks like only the message part of error is used here, so this could be just `specifier, message, packageSource`.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -99,12 +108,14 @@ export const makeImportHookMaker = (
...
           return freeze({ imports: [], exports: [], execute() {} });
         }
-        return postponeErrorToExecute(
+        return deferError(
+          moduleSpecifier,
           new Error(
             `Cannot find external module ${q(
               moduleSpecifier,
             )} in package ${packageLocation}`,
           ),
```
kriskowal, line 117: The Error constructor around the message appears to be superfluous.

packages/compartment-mapper/src/parse-pre-cjs.js
```diff
@@ -35,19 +35,24 @@ export const parsePreCjs = async (
   const execute = async (moduleExports, compartment, resolvedImports) => {
     const functor = compartment.evaluate(source);
 
+    let exportsReferenceCopy = moduleExports;
```
kriskowal, line 38: Did we not change these names to `originalExports` and `finalExports` in a different change?

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/app/index.js
```diff
@@ -47,9 +47,7 @@ module.exports.assertions = {
     try {
       callself.makeTheCall();
     } catch (e) {
-      throw Error('Referencing exported functions should be possible', {
-        cause: e,
-      });
+      throw Error('Referencing exported functions should be possible. \n' + e.stack);
```
kriskowal, line 50: We should not inline the error stack of the cause in the message.

```suggestion
      throw Error(`Referencing exported functions should be possible, because ${cause.message}`, {cause});
```

Also, prefer `cause` or `error` to `e` for the catch clause.

packages/compartment-mapper/test/test-cjs-compat.js
```diff
@@ -1,56 +1,32 @@
...
+  'fixtures-cjs-compat/node_modules/app/index.js',
+  import.meta.url,
+).toString();
 
-  const fixture = new URL(
-    'fixtures-cjs-compat/node_modules/app/index.js',
-    import.meta.url,
-  ).toString();
+const assertFixture = (t, { namespace }) => {
+  console.error(namespace)
```
kriskowal, line 13: `t.log` if you need a log in a test.

packages/compartment-mapper/src/compartment-map.js
naugtur: It's a string containing the error message. I'll add an assertion eventually.

packages/compartment-mapper/src/import-archive.js
naugtur: I'll solve this once the mechanics of firing the error are done. Copy-pasted this from where the typing was necessary.

packages/compartment-mapper/src/import-hook.js
naugtur: I kept it there in case we decide to also pass the stacktrace.

packages/compartment-mapper/src/archive.js
naugtur: I might change it, but at least for now it's been much easier to navigate the codebase with a more unique name for it. I see value in that.

packages/compartment-mapper/src/parse-pre-cjs.js
naugtur: yes. I'm gonna need to rebase this, they were created in parallel

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/app/index.js
naugtur: Node 16.something introduced support for passing cause to error, but it didn't propagate for me (might be ava or some other reason it's dropped) I needed it for debugging.

packages/compartment-mapper/test/test-cjs-compat.js
naugtur: this is a leftover.

packages/compartment-mapper/src/archive.js
```diff
@@ -19,20 +19,22 @@ import { compartmentMapForNodeModules } from './node-modules.js';
...
 const textEncoder = new TextEncoder();
 
-/** @type {Record<string, ParseFn>} */
+/** @type {Record<string, ParserImplementation>} */
 const parserForLanguage = {
-  mjs: parseArchiveMjs,
-  cjs: parseArchiveCjs,
-  json: parseJson,
+  mjs: parserArchiveMjs,
+  'pre-mjs-json': parserArchiveMjs,
```
naugtur, line 34: `Language` type is used to mean 2 things. One of them is the whole set - possible values for parser, the other is just cjs,mjs,json - the list of possible file types/languages.

I needed to cover all values in the mapping here. Otherwise I'd need two distinct types and a one way converter from larger type to the limited one.

packages/compartment-mapper/src/link.js
```diff
@@ -351,7 +353,10 @@ export const link = (
       parserForLanguage,
       moduleTransforms,
     );
-    const importHook = makeImportHook(location, name, parse);
+    /** @type {ShouldDeferError} */
+    const shouldDeferError = language =>
+      parserForLanguage[language || ''].heuristicImports; // TODO: make typescript happy without this empty string
```
naugtur, line 358: Haven't decided how to fix this yet. I don't think this is reachable with language===undefined, but makes sense to future proof it.

packages/compartment-mapper/src/types.js
```diff
@@ -168,12 +169,18 @@ export const moduleJSDocTypes = true;
...
+ * @param {Language | undefined} language
+ * @returns {boolean}
+ */
+
 /**
  * @callback ImportHookMaker
  * @param {string} packageLocation
  * @param {string} packageName
  * @param {ParseFn} parse
- * @param {Record<string, Object>} exitModules
```
naugtur, line 176: exitModules was added (and is used) as input elsewhere, This line seems to be a mistake.

packages/compartment-mapper/test/scaffold.js
```diff
@@ -1,4 +1,4 @@
-import 'ses';
+import '../../ses/index.js';
```
naugtur, line 1: Is there a reason not to keep it that way?

naugtur: Ok, found one and it was unexpected enough, I now expect many more.

packages/compartment-mapper/src/import-hook.js
naugtur: used both where applicable. see new changes.

packages/compartment-mapper/test/test-cjs-compat.js
naugtur: done

packages/compartment-mapper/src/import-hook.js
naugtur: and now it's used to rethrow depending on situation

packages/compartment-mapper/src/archive.js
naugtur: I propose keeping it. Much easier to track across multiple files than just `error`

kriskowal: I support this choice. Thanks for considering.

packages/compartment-mapper/src/parse-pre-cjs.js
```diff
@@ -28,37 +28,59 @@ export const parsePreCjs = async (
...
    */
-  const execute = async (moduleExports, compartment, resolvedImports) => {
+  const execute = (moduleEnvironmentRecord, compartment, resolvedImports) => {
     const functor = compartment.evaluate(source);
 
+    const originalExports = new Proxy(
+      Object.create(compartment.globalThis.Object.prototype),
+      {
+        get(target, prop) {
+          return moduleEnvironmentRecord[prop] || target[prop]; // this makes things like exports.hasOwnProperty() work.
```
kriskowal, line 42: Not a blocker: I generally favor wrapping long comments at 80 or 120 columns, so I also consequently tend to put inline comments on the line above, to avoid a hanging indent.

packages/compartment-mapper/src/parse-pre-cjs.js
```diff
@@ -28,37 +28,59 @@ export const parsePreCjs = async (
...
+        finalExports = value;
       },
     });
 
     const require = freeze((/** @type {string} */ importSpecifier) => {
       const namespace = compartment.importNow(resolvedImports[importSpecifier]);
       if (namespace.default !== undefined) {
-        return namespace.default;
+        if (Object.keys(namespace).length > 1) {
+          return { ...namespace.default, ...namespace }; // this resembles Node's behavior more closely
```
kriskowal, line 66: This comment could use more elaboration.

packages/compartment-mapper/src/types.js
kriskowal: You are correct that this code probably died in a refactor. Thanks.

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/app/index.js
```diff
@@ -57,9 +63,9 @@ module.exports.assertions = {
     try {
       callself.makeTheCall();
     } catch (e) {
-      throw Error('Referencing exported functions should be possible', {
-        cause: e,
-      });
+      throw Error(
+        'Referencing exported functions should be possible. \n' + e.stack,
```
kriskowal, line 67: I’m on the fence about committing investigative code. Your call.

packages/compartment-mapper/src/bundle.js
```diff
@@ -17,21 +17,23 @@ import { compartmentMapForNodeModules } from './node-modules.js';
...
 /** quotes strings */
 const q = JSON.stringify;
 
-/** @type {Record<string, ParseFn>} */
+/** @type {Record<string, ParserImplementation>} */
 const parserForLanguage = {
-  mjs: parseArchiveMjs,
-  cjs: parseArchiveCjs,
-  json: parseJson,
+  mjs: parserArchiveMjs,
```
kriskowal, line 32: This is a little weird because an archive could theoretically contain a mix of mjs and pre-mjs-json, but we explicitly don’t allow `mjs` because to do so would necessitate a dependency on Babel in parts of our system where 3MB of JS is prohibitively expensive.

Is it possible to suffer the type narrowing here with more grace? I could accept stub implementations of the mjs and cjs parsers that just throw if those types are found in archives.

kriskowal (approved):
This reminds me of the expression “a yeoman’s effort”.

I’m giving you a stamp so you can land this while I’m asleep, but please consider my recommendation for making `mjs` and `cjs` parser types explicitly disallowed in archives.

packages/compartment-mapper/src/bundle.js
naugtur: **TL;DR:**

I didn't add the `mjs`, I only renamed the variable parse->parse_**r**_.  
I've added `pre-mjs-json`. And they're both pointing to the same parser reference. I didn't change the list in import-archive. I don't see how that adds new error potential.

That said, I'd like to refactor this to provide more strict typing between how parsers identify themselves and the lparserForLanguage/mapParsers code. 

My plan is to merge this (I find it safe to merge) and proceed to refactor.

----
**All of it:**

Well, the issue here is that `Language` is really 2 types.   

The language of a file as declared in package.json or extension: `Language1` := mjs, cjs, json   
The `parser` field meaning "which parser produced this": `Language2` := `Language1` + pre-mjs-json, pre-cjs-json  

parserForLanguage maps from `Language1` to a parser that's supposed to be used for processing whatever's loaded in the current usecase (therefore it differs between load, bundle and archive)

Meanwhile the lookup to check if the parser is using a heuristic to come up with imports is only given access to `Language2` typed `parser` field.

I had the choice to:
- follow the `Language` type definition and fill the maps with missing items
- add a few lines in link.js that, instead of using the value that came from the parts of the implementation that understands esm/csj nuance, is coding the logic to understnad that  pre-mjs-json maps to mjs etc.

Allowing link.js to be free of cjs/esm knowledge seemed a better choice.

> but we explicitly don’t allow mjs
Note that the fields that I'm adding are `pre-*` and mjs was already there. Newly added fields refer to the same parser implementation as their original counterparts and the actual value of the parser field is `pre-*`
I did not change the keys in import-archive and the bundle implementation doesn't use the newly added fields when looking up to choose a parser. Ideally the `parser` field in the `parse` function output would correspond to the key under which it's found in the parserForLanguage map. 

I'd see it wrapped in a language util (not the final name hopefully) which is fed applicable parsers and returns the parserForLanguage map. I'd go ahead and implement it right away but I don't fully understand the concern of `mjs` showing up in an archive and what's the intended reaction if it's there. My mental model tells me it'd go to import-archive and fail there and it's not a concern for the archive maker. My mental model doesn't fully cover bundling though.

Overall I think this implementation doesn't open new paths to language format errors that didn't exist before. I'd like to refactor the parser selection next so that we can, and I can't believe it's _me_ saying that, strict type matching between what parser says it is and the parserForLanguage map somehow. 


Possible designs:

### A:
Explicitly define 2 types for Language and resolve the type mismatch where it manifests itself, add a utility to handle creating parserForLanguage and/or mapping between Language1 and Language2 where needed

### B:
Drop the declarative mapping from languages to parsers and instead let parser implementations provide a function isResponsibleFor(Language): boolean, make parsers a list and iterate the list calling the isResponsibleFor instead of lookup. With 3 parsers and not more than 5 in reasonably foreseeable future the performance impact is none and the responsibility in load/archive/bundle is limited to providing the right set of parsers

I'd go for B. *Debate me* :D

packages/compartment-mapper/src/parse-pre-cjs.js
naugtur: Ok. 

I've abandoned all code formatting opinions other than use a non-zero number of whitespace characters for indent working in many differently opinionated environments. Happy to adopt yours here as I go and discover them :)

packages/compartment-mapper/src/parse-pre-cjs.js
naugtur: Oh I love elaborating. Had to restrain myself to leave it at this.

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/app/index.js
naugtur: While node 16 supports passing cause, ava doesn't seem to. This was an attempt to replicate the default result of passing cause, not to keep the investigative code. Switching to e.message

packages/compartment-mapper/src/parse-pre-cjs.js
```diff
@@ -28,37 +28,67 @@ export const parsePreCjs = async (
...
       if (namespace.default !== undefined) {
-        return namespace.default;
+        if (Object.keys(namespace).length > 1) {
+          // This resembles Node's behavior more closely.
+          // When originalExports get overwritten with finalExports, all exports get collected in the default field.
+          // While it works fine for the import case, when you require a module with its exports handled that way, it
+          // doesn't behave like require would. Returning namespace.default would match the expected behavior exactly,
+          // but would break the packages actually exporting a default key as their only export. This is a compromise
+          // that decently handles the edge case. A more precise solution capable of differentiating between the actual
+          // default export and the syntetic default export is may be possible.
```
kriskowal, line 73:
```suggestion
          // default export and the syntetic default export may be possible.
```

packages/compartment-mapper/src/bundle.js
kriskowal: I grabbed this branch and removed just `pre-mjs-json` and `pre-cjs-json` in `bundle.js` and nothing broke. But, I now see the problem with removing these in `archive.js`. I’m not sure what solution I favor at this point because I don’t fully understand the bug in what I wrote, but I suspect that this should be solved by dividing the language type.

My vision for language parsers is to eventually make it possible to have `package.json` annotations to allow a per-package mapping from extension to language, and languages to custom parser implementations. So, the key type for this really ought to remain quite liberal (`string`) and we ought generally to check whether an entry is present anywhere we’re consulting it.

packages/compartment-mapper/src/link.js
```diff
@@ -102,7 +104,7 @@ const makeExtensionParser = (
         `Cannot parse module ${specifier} at ${location}, no parser configured for the language ${language}`,
       );
     }
-    const parse = parserForLanguage[language];
+    const parse = parserForLanguage[language].parse;
```
kriskowal, line 107: Please consider

```js
const { parse } = parserForLanguage[language];
```

packages/compartment-mapper/src/link.js
```diff
@@ -351,7 +353,11 @@ export const link = (
       parserForLanguage,
       moduleTransforms,
     );
-    const importHook = makeImportHook(location, name, parse);
+    /** @type {ShouldDeferError} */
+    const shouldDeferError = language =>
+      language && parserForLanguage[language].heuristicImports;
```
kriskowal, line 358: Please add a check for whether `has(parserForLanguage, language)` and an appropriate behavior for the case the language is missing.

packages/compartment-mapper/src/types.js
```diff
@@ -190,6 +197,12 @@ export const moduleJSDocTypes = true;
  * }>}
  */
 
+/**
+ * @typedef {Object} ParserImplementation
+ * @property {boolean} heuristicImports
```
kriskowal, line 202: Please add a description of `heuristicImports` and the motivation for its existence.

kriskowal (changes requested):
You’ve convinced me that the entries you added to `parserForLanguage` in `archive.js` were necessary because of a design bug in the existing code that I won’t ask you to fix in this change.

I’ve left some notes for other minor things, but we can’t add `parentSpecifier` (`referrerSpecifier`) to the `importHook` protocol, for reasons I elaborate below.

Thanks!

packages/ses/src/module-load.js
```diff
@@ -165,7 +168,7 @@ const loadWithoutErrorAnnotation = async (
     return mapGet(moduleRecords, moduleSpecifier);
   }
 
-  const staticModuleRecord = await importHook(moduleSpecifier);
+  const staticModuleRecord = await importHook(moduleSpecifier, parentSpecifier);
```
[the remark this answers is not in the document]
naugtur: Oops, this is the foundation of the decision to postpone the error or not. 

I look up the parser of the parent and if it was mjs, I throw the error right away, if not - defer it. It provides the appropriate difference in behavior between mjs and cjs.

If I can't use parentSpecifier to make that decision it's back to the drawing board.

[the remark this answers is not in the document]
naugtur: So I've been to the pit of despair and back to bring you the new solution which doesn't depend on the parentSpecifier and addresses the concern of the erroring package being imported by both esm and cjs. :D

I may not address the remaining comments before our today's meeting.

packages/compartment-mapper/test/scaffold.js
```diff
@@ -55,8 +55,25 @@ export function scaffold(
   fixture,
   assertFixture,
   fixtureAssertionCount,
+  { onError, shouldFailBeforeArchiveOperations = false } = {},
 ) {
-  test(`${name} / loadLocation`, async t => {
+  // wrapping each time allows for convenient use of test.only
+  const wrap = testFunc => (title, implementation) => {
```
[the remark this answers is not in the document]
naugtur: I'll fix it as soon as you convince prettier to let me. :P

packages/compartment-mapper/src/bundle.js
naugtur: > removed just pre-mjs-json and pre-cjs-json in bundle.js and nothing broke

IMHO it's more likely to mean I need to write more tests for bundle.js

>  quite liberal

Ok, so refactoring option B is off the table. That was a short debate.

packages/compartment-mapper/src/link.js
naugtur: but typescript said it was fine! xD

done

packages/compartment-mapper/test/test-cjs-compat.js
```diff
@@ -1,59 +1,32 @@
...
-    import.meta.url,
-  ).toString();
+const assertFixture = (t, { namespace }) => {
+  const { assertions } = namespace;
 
-  const application = await loadLocation(read, fixture);
-  const {
-    namespace: { assertions },
-  } = await application.import({});
+  assertions.packageReferencingItself();
```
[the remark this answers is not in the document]
naugtur: lovely, but harder to understand for a non-native speaker.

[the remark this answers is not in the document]
kriskowal: It’s also less likely to be understood among native speakers 👍

packages/compartment-mapper/src/parse-pre-cjs.js
```diff
@@ -28,37 +28,67 @@ export const parsePreCjs = async (
...
       if (namespace.default !== undefined) {
-        return namespace.default;
+        if (Object.keys(namespace).length > 1) {
+          // This resembles Node's behavior more closely.
+          // When originalExports get overwritten with finalExports, all exports get collected in the default field.
+          // While it works fine for the import case, when you require a module with its exports handled that way, it
+          // doesn't behave like require would. Returning namespace.default would match the expected behavior exactly,
+          // but would break the packages actually exporting a default key as their only export. This is a compromise
+          // that decently handles the edge case. A more precise solution capable of differentiating between the actual
+          // default export and the syntetic default export may be possible.
```
kriskowal, line 73:
```suggestion
          // default export and the synthetic default export may be possible.
```

kriskowal (approved):
Looks like a win. Thanks for playing through to the end!

PR #1091, opened 2022-02-22
feat(compartment-mapper)!: Pre-load for archive integrity checks

kriskowal:
*BREAKING CHANGE:* Previously, `loadArchive` and `parseArchive`, when given a `computeSha512`, would accept just about any archive. Hash integrity checks for any used module occurred only after a request to import them. With this new version, all archives must use every file they contain and must pass hash integrity checks during the load or parse phase.  Consequently, if an archive requires any built-in modules ("exits"), these must be mentioned with the `modules` option to `loadArchive` or `parseArchive`, as an object whose keys are the names of the expected modules.

- I am posting this PR as an early draft for a direction check while I begin producing test cases that cover invalid bundles.

### DIFF
--- a/packages/compartment-mapper/test/test-integrity.js
+++ b/packages/compartment-mapper/test/test-integrity.js
@@ -1,0 +1,78 @@
+// @ts-check
+import 'ses';
+import test from 'ava';
+import { ZipReader, ZipWriter } from '@endo/zip';
+import { makeArchive, parseArchive } from '../index.js';
+import { readPowers } from './scaffold.js';
+
+const fixture = new URL(
+  'fixtures-0/node_modules/app/main.js',
+  import.meta.url,
+).toString();
+
+test('extracting an archive with a superfluous file', async t => {
+  const validBytes = await makeArchive(readPowers, fixture, {
+    modules: {
+      builtin: null,
+    },
+    dev: true,
+  });
+
+  const reader = new ZipReader(validBytes);
+  const writer = new ZipWriter();
+  writer.files = reader.files;
+  writer.write('extraneous.txt', new TextEncoder().encode('Extra'));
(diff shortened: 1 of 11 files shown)


warner (approved):
The approach sounds sensible. The `compartment.load()` sure looks like it's going to execute archive code, but I trust your assurance that it will not (might be worth a reassuring comment though). I look forward to the tests that demonstrate this for certain (maybe parse/validate an archive with a module that has an infinite loop at the top level).

packages/compartment-mapper/test/test-integrity.js
```diff
@@ -0,0 +1,78 @@
...
+    {
+      message:
+        'Archive contains extraneous files: ["extraneous.txt"] in "extra.zip"',
+    },
+  );
+
+  t.pass();
+});
+
+test('extracting an archive with an inconsistent hash', async t => {
```
warner, line 44: This tests that a module/member file can't be corrupted.. do we already have a test that the compartment map cannot be corrupted in a similar way?

Also, is there a test that a module/member file cannot be missing? If the resulting behavior is that the module cannot be loaded, that's fail-safe, but I do want the error to be sensed early (e.g. when swingset is asking the question "does this bundle match this ID?"), rather than not appearing until we try to evaluate code.

warner (approved):
Looks good, thanks for the tests. Might still need an extra test or two, happy to see this land now and the tests added later, if you prefer.

packages/compartment-mapper/test/test-integrity.js
kriskowal: I’ve added two compartment map corruption tests. One compares the valid and invalid returned hashes. The other makes use of the `expectedSha512` option on `parseArchive`. I’ve also added a test for a missing file.

PR #1092, opened 2022-02-22
fix(compartment-mapper): Relax schema check for name and label

kriskowal:
The name and label properties of a compartment map should be optional.
This change relaxes the schema validator for compartment maps.

### DIFF
--- a/packages/compartment-mapper/README.md
+++ b/packages/compartment-mapper/README.md
@@ -386,6 +386,8 @@ type CompartmentName = string;
 // to modules in other compartments, or to built-in modules.
 type Compartment = {
   location: Location,
+  name?: string,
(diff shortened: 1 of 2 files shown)


naugtur (approved):
tested - works.

(could) add a package to fixtures-0 with `{}` for package.json to prevent this in the future

kriskowal:
> tested - works.
> 
> (could) add a package to fixtures-0 with `{}` for package.json to prevent this in the future

I did this and it broke because archives do in fact require a `name`. I think that means we should require a `name` in general.

naugtur:
>  I think that means we should require a name in general.
Fine with me. I detected it accidentally and don't see a usecase where it wouldn't be there.

PR #1094, opened 2022-02-25
feat(hash-bundle): Create bundler hashing package

kriskowal:
This is an initial version of the `@endo/hash-bundle` package that can generate “version strings” for all three kinds of Endo bundle. The `lite.js` module has no powerful built-in module dependencies.

Individually reviewable commits. The first is mostly scaffolding, though `package.json` is filled out.


packages/hash-bundle/README.md
```diff
@@ -0,0 +1,12 @@
+# hash-bundle
+
+`hashBundle` is a function that takes a bundle and returns a consistent hash of the bundle that can be used to perform an integrity check of the bundle.
+To that end, `hashBundle` verifies the internal consistency, completeness, coherence, and conciseness (no extra files) of the bundle.
+
+```js
+const bundle = await bundleSource('path/to/bundle.js');
+// 'bundle' is JSON-serializable
+const version = await hashBundle(bundle);
```
warner, line 9: Let's call it a `bundleID`, since `version` implies (to me) a scalar integer of a class, not an instance.

In my prototype, I used `b1-` as a prefix (for "bundle"). That speaks to a larger application in which this is embedded, where we're anticipating the need to distinguish between modules (`m1-`?) and full bundles, and maybe some other user-visible identifiers that we haven't named yet. I kind of want to ask `hash-bundle` to use `b1-`, but that might be imposing an application-specific requirement on an upstream library, which feels wrong.

If `hash-bundle` uses `v1-`, I might have swingset prepend a `b` (so users see `bv1-HASH`), or replace the `v1-` with `b1-`.. not sure.

This document seems as good a place as any to specify what the string really is, since anyone using this package will care about the details (in particular they'll care about the details not changing, since this is a compatibility guarantee). Maybe some text like:

* "getExport" -type bundles: the BundleID is a `v0-` prefix followed by the lowercase hex SHA512 hash of the full contents. Note that this includes platform-native line endings, so the same source code bundle into different IDs.
* "endoZipBase64"-type bundles (the current default): the BundleID is a `v1-` prefix followed by the lowercase hex SHA512 hash of the compartment map. This does not depend upon the bundling platform.

packages/hash-bundle/lite.js
```diff
@@ -0,0 +1,73 @@
...
+      d`Cannot determine hash of bundle with getExport moduleFormat because it is not consistent across platforms`,
+    );
+  } else {
+    assert.fail(
+      d`Cannot determine hash of bundle with unrecognized moduleFormat ${q(
+        moduleFormat,
+      )}`,
+    );
+  }
+  return `${prefix}-${sha512}`;
```
warner, line 72: Either `prefix` should be `v1`, or the concatenation should omit the hyphen, this produces `v1--HASH`, not `v1-HASH`.

packages/hash-bundle/test/test-hash-bundle.js
```diff
@@ -0,0 +1,47 @@
...
+  const hash = await hashBundle(bundle, computeSha512, 'fixture/main.js');
+  t.snapshot(hash);
+  t.assert(hash.startsWith('v0-'));
+});
+
+test('bundle and hash endo (v1) package', async t => {
+  const bundle = await bundleSource(fixture, 'endoZipBase64');
+  const hash = await hashBundle(bundle, computeSha512, 'fixture/main.js');
+  t.snapshot(hash);
+  t.assert(hash.startsWith('v1-'));
```
warner, line 39: maybe also assert the rest of the shape, in the swingset prototype I use:

```js
  const bundleIDRE = new RegExp('^b1-[0-9a-f]{128}$');
```

packages/hash-bundle/README.md
warner: Oh, and the README should make it clear that calling `hashBundle(bundle)` does not verify that the bundle is internally consistent, but merely computes an ID that might match the rest of the bundle. Maybe provide an example of what needs to be done to compute+validate the ID of an alleged bundle (which requires an additional package).

For swingset and chain deploy, the use cases are:
* contract/vat/bundle author writes their code, then pushes the "upload/deploy" button
  * their agent bundles the sources, computes the ID, performs an upload
  * author needs to be told the ID
  * to avoid surprises / guard against bugs, the agent should probably validate the bundle against the computed ID, but it's not strictly necessary
    * worst case (platform-sensitive line-ending bug??) a bug in the bundling/hashing process would cost the author fees when they attempt to upload and get rejected for a mismatch
* the chain receives an alleged bundle and ID, and wants to store it
  * chain must validate the complete bundle against the claimed ID
  * some environments might not require the client to supply an ID, so the chain may need to compute the ID first
* later, when loading the bundle, it's probably a good idea to re-validate it, just before/during the `eval`

warner: Hm, now I'm on the fence at to whether I want `hash-bundle` to also validate the whole thing, and/or if there is a use case for only hashing the compartment map. In general I want `hash-bundle` to actually look at the whole bundle, but making it do everything would preclude it being used in a place where, I dunno, maybe you want to quickly/cheaply tell that a given bundle is not a duplicate of something you already have, or you want to look up previously-calculated information about an ID.

Oh, ok, so when we get to uploading bundles efficiently (just the modules they don't already have), the uploading agent (running `agoric install-contract BUNDLE` or something) might have a previously-created bundle and is willing to rely upon its correctness. It wants to quickly/cheaply learn the ID so it can send the first message to the chain ("I'd like to upload bundle-ID, how much of it do you already have?"), and they don't need to validate anything else about their local bundle to do that.

That use case will also want a function that takes a reliable bundle and returns a set of module IDs, probably at the same time as returning the overall bundle (compartment map) ID. We might compute and stash that during bundling, but we'll probably need to apply it to a pre-made bundle, and then we're either willing to rely upon the integrity or not. If we are, then all we have to do is hash the compartment map and return both the bundle hash and the hashes from inside the compartment map. If not, we need to validate everything, and we should gather the hashes to return at the same time, instead of hashing twice.

warner (changes requested):
double-hyphen needs fixing, docs changes are recommended but not strictly necessary

kriskowal:
I’ve changed `hash-bundle` to `check-bundle`, and the hash is expressed with an `endoZipBase64Sha512` property on the bundle itself. The integrity checker rejects if the bundle is internally inconsistent. I changed `bundleSource` to generate the expected hash property too, which required a small enhancement to compartment-mapper. Commits are individually reviewable.

I discovered, late, that the reason Windows tests were failing was not because I had failed to normalize the inputs, but because `recast` was generating platform-appropriate line-endings. I rebased atop @michaelfig’s upcoming https://github.com/endojs/endo/pull/1096 which reverts recast in favor of babel’s own codegen and the problem went away.

packages/bundle-source/src/index.js
```diff
@@ -140,21 +140,29 @@ async function transformSource(
...
       },
     },
-  });
+  );
   const endoZipBase64 = encodeBase64(bytes);
-  return harden({ endoZipBase64, moduleFormat: 'endoZipBase64' });
+  return harden({
+    endoZipBase64,
+    endoZipBase64Sha512: sha512,
+    moduleFormat: 'endoZipBase64',
```
warner, line 164: nit: I'd put `moduleFormat` first, to emphasize that it's the first thing a reader needs to pay attention to ("don't use the presence of `.endoZipBase64` to discover the module format, we have a key to tell you that")

packages/check-bundle/index.js
```diff
@@ -0,0 +1,20 @@
...
+  const hash = crypto.createHash('sha512');
+  hash.update(bytes);
+  return hash.digest().toString('hex');
+};
+
+/**
+ * @param {string} path
+ * @param {string=} name
+ */
+export const checkBundle = async (path, name = undefined) => {
```
warner, line 17: Thanks for the batteries-included form. My use case has bytes in memory rather than a file on disk, so `controller.validateAndInstallBundle()` would need to define its own `computeSha512` and use `checkBundleBytes` from `lite.js`, in case that influences what API you want to give the non-lite version.

packages/check-bundle/lite.js
```diff
@@ -0,0 +1,79 @@
...
+      computeSha512,
+      expectedSha512,
+    });
+    assert(parsedSha512 !== undefined);
+  } else if (
+    moduleFormat === 'getExport' ||
+    moduleFormat === 'nestedEvaluate'
+  ) {
+    assert.fail(
+      `checkBundle cannot determine hash of bundle with ${moduleFormat} moduleFormat because it is not necessarilly consistent`,
```
warner, line 70: nit `necessarily`

packages/check-bundle/lite.js
```diff
@@ -0,0 +1,79 @@
...
+    const { endoZipBase64, endoZipBase64Sha512: expectedSha512 } = bundle;
+    assert.typeof(
+      endoZipBase64,
+      'string',
+      d`checkBundle cannot hash non-bundle, property 'endoZipBase64' must be a string, got ${typeof endoZipBase64}`,
+    );
+    assert.typeof(
+      expectedSha512,
+      'string',
+      d`checkBundle cannot bundle without the property 'endoZipBase64Sha512', which must be a string, got ${typeof expectedSha512}`,
```
warner, line 57: `property 'expectedSha512'`, not `endoZipBase64Sha512`

packages/check-bundle/package.json
```diff
@@ -0,0 +1,82 @@
+{
+  "name": "@endo/check-bundle",
+  "version": "0.1.0",
+  "description": "Checks the integrity of an Endo bundle and returns a version string suitable for integrity verification.",
```
warner, line 4: I think this no longer returns a version string

packages/check-bundle/test/test-check-bundle.js
```diff
@@ -0,0 +1,64 @@
...
+  });
+});
+
+test('bundle and check endo zip base64 package', async t => {
+  const bundle = await bundleSource(fixture, 'endoZipBase64');
+  await checkBundle(bundle, computeSha512, 'fixture/main.js');
+  t.pass();
+});
+
+test('bundle and check corrupt endo zip base64 package', async t => {
```
warner, line 41: Hm, should we also test a package with the `endoZipBase64Sha512` property missing entirely, since that's similar to the way a lot of JWT libraries get broken (`alg: 'none'`)? Such bundles are fine for importing, but should not claim to be validatable.

warner: It's paranoia, but let's also test against a syntactically-valid but almost-certainly wrong hash, just 512/4=128 zeros, to get the test coverage deep enough into the implementation to be comparing hashes against hashes.

warner (approved):
A tiny part of me wants to ask for APIs with `SHA512` instead of `sha512`/`Sha512`, since it's an acronym. But I think that ship (we could christen it the `Hms Acronym`?) has already sailed.

Mostly cosmetic and testing improvements, feel free to land once addressed to your satisfaction.

kriskowal:
I certainly deliberately and without malice but with forethought steered our ship onto camel case for initialisms and acronyms here because upper-case inevitably degenerate to upper-run-on-case, obscuring the boundary between adjacent terms.

packages/hash-bundle/README.md
kriskowal: We converged on a solution that avoids naming a “tagged hash” at this layer of our stack. A bundle will contain an `endoZipBase64Sha512` property if it has an associated hash. `bundleSource` will produce this property. `checkBundle` will verify its consistency with the content of `endoZipBase64`.

packages/check-bundle/index.js
kriskowal: I resolved this issue by adding a second method to the batteries-included interface.

packages/check-bundle/lite.js
kriskowal: I reärranged the names to make this more obviously right.

PR #1095, opened 2022-02-26
fix(eventual-send): stronger (compatible) typing of `E` and `Remote`

michaelfig:
First, a little change to propagate types from `@endo/eventual-send`'s `E` to the outside world.

Then, a more ambitious change to instate a proper `Remote<Primary, Local=DataOnly<Primary>>` type that enforces the rules of our distributed object system.  This helps prevent the `(await o).myMethod()` cheat that worked locally but failed when `o` was not the primary remotable.

### DIFF
--- a/packages/eventual-send/jsconfig.test.json
+++ b/packages/eventual-send/jsconfig.test.json
@@ -1,0 +1,4 @@
+{
+  "extends": "./jsconfig.json",
+  "include": ["test/test-types.js"]
+}
--- a/packages/eventual-send/src/E.js
+++ b/packages/eventual-send/src/E.js
@@ -23,7 +23,7 @@ const baseFreezableProxyHandler = {
  * A Proxy handler for E(x).
  *
  * @param {*} x Any value passed to E(x)
- * @param {*} HandledPromise
+ * @param {import('./index').HandledPromiseConstructor} HandledPromise
  * @returns {ProxyHandler} the Proxy handler
  */
 function EProxyHandler(x, HandledPromise) {
@@ -47,7 +47,7 @@ function EProxyHandler(x, HandledPromise) {
  * It is a variant on the E(x) Proxy handler.
  *
  * @param {*} x Any value passed to E.sendOnly(x)
- * @param {*} HandledPromise
+ * @param {import('./index').HandledPromiseConstructor} HandledPromise
  * @returns {ProxyHandler} the Proxy handler
  */
 function EsendOnlyProxyHandler(x, HandledPromise) {
@@ -70,6 +70,10 @@ function EsendOnlyProxyHandler(x, HandledPromise) {
   });
 }
 
+/**
+ * @param {import('./index').HandledPromiseConstructor} HandledPromise
+ * @returns {import('./index').EProxy}
+ */
 export default function makeE(HandledPromise) {
   function E(x) {
     const handler = EProxyHandler(x, HandledPromise);
--- a/packages/eventual-send/src/index.d.ts
+++ b/packages/eventual-send/src/index.d.ts
@@ -18,6 +18,20 @@ export type EOnly<T> = T extends (...args: infer P) => infer R
     >
   : ERef<T>;
 
+type FilteredKeys<T, U> = {
+  [P in keyof T]: T[P] extends U ? P : never;
+}[keyof T];
+
+type DataOnly<T> = Omit<T, FilteredKeys<T, Function>>;
+type FunctionOnly<T> = Pick<T, FilteredKeys<T, Function>> &
+  (T extends Function ? (...args: Parameters<T>) => ReturnType<T> : {});
+
+interface Remotable<T> {
+  __Remote__: T;
+}
+type Remote<Primary, Local = DataOnly<Primary>> = ERef<
+  Local & Remotable<Primary>
+>;
 type Unpromise<T> = T extends ERef<infer U> ? U : T;
 
 type Parameters<T> = T extends (...args: infer T) => any ? T : any;
@@ -59,7 +73,7 @@ type ResolveWithPresenceOptionsBag<T extends Object> = {
 };
 
 declare interface HandledPromiseStaticMethods {
-  resolve<T>(x: T): Promise<Unpromise<T>>;
+  resolve<T>(x: T): Promise<Awaited<T>>;
   resolve(): Promise<undefined>;
   applyFunction(target: unknown, args: unknown[]): Promise<unknown>;
   applyFunctionSendOnly(target: unknown, args: unknown[]): void;
@@ -95,14 +109,14 @@ declare function makeHandledPromise(): HandledPromiseConstructor;
 type ESingleMethod<T> = {
   readonly [P in keyof T]: (
     ...args: Parameters<T[P]>
-  ) => Promise<Unpromise<ReturnType<T[P]>>>;
+  ) => Promise<Awaited<ReturnType<T[P]>>>;
 };
 type ESingleCall<T> = T extends Function
-  ? ((...args: Parameters<T>) => Promise<Unpromise<ReturnType<T>>>) &
+  ? ((...args: Parameters<T>) => Promise<Awaited<ReturnType<T>>>) &
       ESingleMethod<Required<T>>
   : ESingleMethod<Required<T>>;
 type ESingleGet<T> = {
-  readonly [P in keyof T]: Promise<Unpromise<T[P]>>;
+  readonly [P in keyof T]: Promise<Awaited<T[P]>>;
 };
 
 /* Same types for send-only. */
@@ -117,7 +131,7 @@ type ESingleGetOnly<T> = {
 };
 
 interface ESendOnly {
-  <T>(x: T): ESingleCallOnly<Unpromise<T>>;
+  <T>(x: T): ESingleCallOnly<Awaited<T>>;
 }
 
 interface EProxy {
@@ -130,7 +144,13 @@ interface EProxy {
    * @param {*} x target for method/function call
    * @returns {ESingleCall} method/function call proxy
    */
-  <T>(x: T): ESingleCall<Unpromise<T>>;
+  <T>(x: T): ESingleCall<
+    T extends Remotable<infer U>
+      ? FunctionOnly<U>
+      : Awaited<T> extends Remotable<infer U>
+      ? FunctionOnly<U>
+      : Awaited<T>
+  >;
 
   /**
    * E.get(x) returns a proxy on which you can get arbitrary properties.
@@ -141,13 +161,21 @@ interface EProxy {
    * @param {*} x target for property get
    * @returns {ESingleGet} property get proxy
    */
-  readonly get: <T>(x: T) => ESingleGet<Unpromise<T>>;
+  readonly get: <T>(
+    x: T,
+  ) => ESingleGet<
+    T extends Remotable<infer U>
+      ? DataOnly<U>
+      : Awaited<T> extends Remotable<infer U>
+      ? DataOnly<U>
+      : Awaited<T>
+  >;
 
   /**
    * E.resolve(x) converts x to a handled promise. It is
    * shorthand for HandledPromise.resolve(x)
    */
-  readonly resolve: <T>(x: T) => Promise<Unpromise<T>>;
+  readonly resolve: <T>(x: T) => Promise<Awaited<T>>;
 
   /**
    * E.when(x, res, rej) is equivalent to
@@ -155,7 +183,7 @@ interface EProxy {
    */
   readonly when: <T, U>(
     x: T,
-    onfulfilled?: (value: Unpromise<T>) => ERef<U>,
+    onfulfilled?: (value: Awaited<T>) => ERef<U>,
     onrejected?: (reason: any) => ERef<U>,
   ) => Promise<U>;
 
--- a/packages/eventual-send/package.json
+++ b/packages/eventual-send/package.json
@@ -13,7 +13,7 @@
     "lint-fix": "yarn lint:eslint --fix && yarn lint:types",
     "lint-check": "yarn lint",
     "lint": "yarn lint:types && yarn lint:eslint",
-    "lint:types": "tsc -p jsconfig.json",
+    "lint:types": "tsc -p jsconfig.test.json",
     "lint:eslint": "eslint '**/*.js'"
   },
   "repository": {
--- a/packages/eventual-send/test/prepare-test-env-ava.js
+++ b/packages/eventual-send/test/prepare-test-env-ava.js
@@ -3,5 +3,7 @@ import '@endo/lockdown/commit-debug.js';
 import { wrapTest } from '@endo/ses-ava';
 import rawTest from 'ava';
 
+export * from 'ava';
+
 /** @type {typeof rawTest} */
 export const test = wrapTest(rawTest);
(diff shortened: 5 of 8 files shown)


packages/eventual-send/jsconfig.test.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": "./jsconfig.json",
+  "include": ["test/test-types.js"]
```
mhofman, line 3: I don't remember, do this extend or replace the list of included files?

packages/eventual-send/src/index.d.ts
```diff
@@ -18,6 +18,20 @@ export type EOnly<T> = T extends (...args: infer P) => infer R
     >
   : ERef<T>;
 
+type FilteredKeys<T, U> = {
+  [P in keyof T]: T[P] extends U ? P : never;
+}[keyof T];
```
mhofman, line 23: neat!

packages/eventual-send/src/index.d.ts
```diff
@@ -18,6 +18,20 @@ export type EOnly<T> = T extends (...args: infer P) => infer R
...
+
+type DataOnly<T> = Omit<T, FilteredKeys<T, Function>>;
```
mhofman, line 25: I'm always wary of using `Function` in these cases. Why not use `(...args: any[]) => any` ?

packages/eventual-send/src/index.d.ts
```diff
@@ -18,6 +18,20 @@ export type EOnly<T> = T extends (...args: infer P) => infer R
...
+type FunctionOnly<T> = Pick<T, FilteredKeys<T, Function>> &
+  (T extends Function ? (...args: Parameters<T>) => ReturnType<T> : {});
+
+interface Remotable<T> {
+  __Remote__: T;
+}
+type Remote<Primary, Local = DataOnly<Primary>> = ERef<
+  Local & Remotable<Primary>
+>;
 type Unpromise<T> = T extends ERef<infer U> ? U : T;
```
mhofman, line 35: Could we also alias `Unpromise` to `Awaited` or is there a functional difference ?

turadg (approved):
LGTM. These types are working in agoric-sdk.

mhofman (commented):
Cool stuff!

packages/eventual-send/src/index.d.ts
turadg: > wary

Why? What risk would it prevent in this case?

> Why not use `(...args: any[]) => any` ?
`Function` reads much more clearly and there's no benefit to the longer form.

packages/eventual-send/src/index.d.ts
```diff
@@ -18,6 +18,20 @@ export type EOnly<T> = T extends (...args: infer P) => infer R
...
+type FilteredKeys<T, U> = {
+  [P in keyof T]: T[P] extends U ? P : never;
+}[keyof T];
+
+type DataOnly<T> = Omit<T, FilteredKeys<T, Function>>;
+type FunctionOnly<T> = Pick<T, FilteredKeys<T, Function>> &
+  (T extends Function ? (...args: Parameters<T>) => ReturnType<T> : {});
+
+interface Remotable<T> {
+  __Remote__: T;
```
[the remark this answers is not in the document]
turadg: How would that work? I imagine it would make the apparent type harder to interpret.

packages/eventual-send/jsconfig.test.json
turadg: Good catch! It replaces according to https://www.typescriptlang.org/tsconfig#extends

So this change made it so _only_ test-types is checked. Regret I missed that.

packages/eventual-send/src/index.d.ts
mhofman: My understanding is that `Function` is anything that has the shape of the `Function` prototype, aka has `.call`, `.apply`, `.bind` etc. `(...args: any[]) => any` is anything that is callable.

This is similar to an object with properties being different than something that extends `Object`.

See this [playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBCBGArGBeGBvAUDHMwEMBbAUwC4YAiBRCgGm1wBtiwBzKAC3IAZ7cZ8AB0GMAngAoA+pwCWEcvjCiAlBgC+fXMHyNGU2fIFLV6DQxzwZYACb6Oc0opUZz-GACdiUAK7uwcJABuVzN+KBAAZSh3K1ZxE1d+Tx8-SgAxfABrYhg07zBgKBlwCmD+UNwAbQjRQngQRgA6DnwIAElIKEVgYgBdeJc3XGTffwAzHQhiMtwKnEF3EHCoUUEyXPzC4rBGhaWl1eJNHHx3Vm8SMChDSt7jmG1dYndyAZQAPnV6NUxMFbWAog2hA8gUiuBUDB-sQQGNATBiAAPKAsawQDZg7YwAD8UPc3hy5AmjCmwUwoE6MDG+XIoK2ELQ1ECQA).

packages/eventual-send/src/index.d.ts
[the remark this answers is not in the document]
mhofman: Something like 

```ts
export declare const remoteSymbol: unique symbol;

interface Remotable<T> {
  [remoteSymbol]: T;
}
```

It'd avoid making `__Remote__` a string property of the `Remotable`.

packages/eventual-send/src/index.d.ts
michaelfig: I'll remove it.

packages/eventual-send/src/index.d.ts
turadg: Right. In this case I think that's what we want. We're partitioning things that are callable from those that aren't.

packages/eventual-send/src/index.d.ts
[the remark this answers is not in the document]
turadg: Neat! I thought it would put some more confusing string into the type, but it seems to be strictly better :) 
```
export declare const remoteSymbol: unique symbol;

interface Remotable<T> {
  [remoteSymbol]: T;
  __Remote__: T;
}

const obj = {a: 1, b: 2};

const robj = obj as unknown as Remotable<typeof obj>;
const r1 = robj.__Remote__; // passes typecheck, which is undesired
const r2 = robj[remoteSymbol]; // also passes typecheck but at least it's not a member of robj
```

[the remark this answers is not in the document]
mhofman: Correct, and we may get away with not exporting the `remoteSymbol` declaration, so actual code wouldn't even be able to write `robj[remoteSymbol]`

packages/eventual-send/src/index.d.ts
```diff
@@ -18,6 +18,20 @@ export type EOnly<T> = T extends (...args: infer P) => infer R
...
+}
+type Remote<Primary, Local = DataOnly<Primary>> = ERef<
+  Local & Remotable<Primary>
+>;
 type Unpromise<T> = T extends ERef<infer U> ? U : T;
 
 type Parameters<T> = T extends (...args: infer T) => any ? T : any;
```
mhofman, line 37: These are extraneous from `lib.es5.d.ts`

mhofman: Also while we're at it, I believe `Property === PropertyKey`

packages/eventual-send/src/index.d.ts
erights, line 25: Could you describe in english what `DataOnly` or `FunctionOnly` mean? Since I didn't understand `FilteredKeys` I can't figure out these. Also btw, what's `Omit` and `Pick`?

packages/eventual-send/src/index.d.ts
```diff
@@ -18,6 +18,20 @@ export type EOnly<T> = T extends (...args: infer P) => infer R
...
+type DataOnly<T> = Omit<T, FilteredKeys<T, Function>>;
+type FunctionOnly<T> = Pick<T, FilteredKeys<T, Function>> &
+  (T extends Function ? (...args: Parameters<T>) => ReturnType<T> : {});
+
+interface Remotable<T> {
+  __Remote__: T;
+}
+type Remote<Primary, Local = DataOnly<Primary>> = ERef<
+  Local & Remotable<Primary>
+>;
```
erights, line 34: Sorry, I'm just not getting these. From the terminology we've been using, I have an expectation of what `Remote<T>` or `Remotable<T>` would mean. But I haven't been able to fit any hypothesis into the declarations written here.

packages/eventual-send/src/index.d.ts
erights: For each of `Promise<T>`, `Awaited<T>`, or `Unpromise<T>`, what's their treatment of non-promise thenables?

packages/eventual-send/src/index.d.ts
erights, line 35: If we're deprecating `Unpromise` in favor of `Awaited`, should we formally deprecate `Unpromise` here? Together with a note to use `Awaited` instead?

packages/eventual-send/src/index.d.ts
```diff
@@ -130,7 +144,13 @@ interface EProxy {
    * @param {*} x target for method/function call
    * @returns {ESingleCall} method/function call proxy
    */
-  <T>(x: T): ESingleCall<Unpromise<T>>;
+  <T>(x: T): ESingleCall<
+    T extends Remotable<infer U>
```
erights, line 148: This is the first mention of `U`. What is it inferred from?

Needless to say, since I don't understand the things this is based on, several levels deep, I am not able to decypher this. At least in this case I feel fairly confident of what `EProxy` is supposed to mean.

packages/eventual-send/src/index.d.ts
```diff
@@ -141,21 +161,29 @@ interface EProxy {
...
+      ? DataOnly<U>
+      : Awaited<T>
+  >;
 
   /**
    * E.resolve(x) converts x to a handled promise. It is
    * shorthand for HandledPromise.resolve(x)
    */
-  readonly resolve: <T>(x: T) => Promise<Unpromise<T>>;
+  readonly resolve: <T>(x: T) => Promise<Awaited<T>>;
```
erights, line 178: Btw, what are our plans for `resolve` when we fold `HandledPromise` into `Promise`?

erights (commented):
I had hoped that I could use the types created in our .ts files without becoming a TS expert well beyond what we've needed to use ts in jsdoc in js. As a reviewer I can understand needing to, because otherwise I cannot understand the implementation of these types. But as a user, using them from ts in jsdoc in js, I should be able to get by with a non-expert understanding combined with ***comments*** in the .ts file explaining anything non-obvious about what these types mean.

Here, I started out not being able to figure out what `FilteredKeys` is even supposed to mean. Since it is uncommented, and everything else of interest is built on it several layers deep mostly without comments, I got totally lost. Hence, at this stage my review is mostly vacuous.

packages/eventual-send/src/E.js
```diff
@@ -23,7 +23,7 @@ const baseFreezableProxyHandler = {
  * A Proxy handler for E(x).
  *
  * @param {*} x Any value passed to E(x)
- * @param {*} HandledPromise
+ * @param {import('./index').HandledPromiseConstructor} HandledPromise
```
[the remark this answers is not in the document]
michaelfig: > Can we factor this out into a single import at the top of the file?

There's only one reference to it, which is usually why I would avoid creating an entirely separate

> Why isn't the imported type named the same?

Typescript conventions.  The `XXXConstructor` type is the type of the function that constructs an `XXX`.  `XXX` is used as the type of an instance created by an `XXXConstructor`.  JS value convention is to name the constructor `XXX`, and have no name for the instance.

packages/eventual-send/src/index.d.ts
michaelfig: > Could you translate this into an english description? I'm having trouble figuring it out.

For the type `FilteredKeys<T, U>` return a union of property names/symbols/numbers P for which the record element `T[P]`'s type extends `U`.

Given `const x = { a: 123, b: 'hello', c: 42, 49: () => {}, 53: 67 }`,

`FilteredKeys<typeof x, number>` is the type `'a' | 'c' | 53`.
`FilteredKeys<typeof x, string>` is the type `'b'`.
`FilteredKeys<typeof x, 42 | 67>` is the type `'c' | 53`.
`FilteredKeys<typeof x, boolean>` is the type `never`.

packages/eventual-send/src/index.d.ts
michaelfig: `DataOnly<T>` means to return a record type `T2` consisting only of properties that are *not* functions.
`FunctionOnly<T>` means to return a record type `T2` consisting only of properties that are functions.

`Omit<T, U>` means to return a record type `T2` which has none of the properties whose keys are part of `U`.
`Pick<T, U>` means to return a record type `T2` which has only the properties whose keys are part of `U`.

`Omit<{a: 1, b: 2, c: 3}, 'b'>` is the type `{a: 1, c: 3}`.
`Pick<{a: 1, b: 2, c: 3}, 'b'>` is the type `{b: 2}`.

packages/eventual-send/src/index.d.ts
michaelfig: `Remotable<T>` is confusing and should probably be written `RemoteIntefaceOf<T>`.

`Remote<T>` creates a type that accepts both near and marshalled references that were returned from `Remotable` or `Far`.

packages/eventual-send/src/index.d.ts
michaelfig: I do that in the next PR (#1102).  Nobody else was using `Unpromise`, so I didn't leave a note.

packages/eventual-send/src/index.d.ts
michaelfig: `PromiseLike<T>` is a thenable which resolves to `T`.

`Promise<PromiseLike<T>>` doesn't handle recursion and is distinct from 'T'.

`Unpromise<PromiseLike<T>>` strips off just one layer and is just `T`.  `Unpromise<PromiseLike<PromiseLIke<T>>` is `PromiseLike<T>`.

`Awaited<PromiseLike<T>>` recurses, and is just `T`.
`Awaited<PromiseLike<PromiseLike<T>>>` is just `T` as well.

packages/eventual-send/src/index.d.ts
michaelfig: As above, this type should use `RemoteInterfaceOf` instead of `Remotable`:

> This is the first mention of `U`. What is it inferred from?

It is inferred as the type `U` such that `T extends RemoteInterfaceOf<U>`.  So, `T is the remote interface of some U`.

```js
T extends RemoteInterfaceOf<infer U> // if a given T is the remote interface of some U
   ? // then,
  FunctionOnly<U> // use all the function properties of that U
  : // otherwise,
  Awaited<T> extends RemoteInterfaceOf<infer U> // if the final resolution of T is the remote interface of some U
  ? // then,
  FunctionOnly<U> // use the function properties of that U
  : // otherwise,
  Awaited<T> // use the final resolution of that T
```

packages/eventual-send/src/index.d.ts
michaelfig: > Btw, what are our plans for `resolve` when we fold `HandledPromise` into `Promise`?

Yeah, I've been thinking about that on and off.  Maybe we will have `Promise.delegated.resolve(x)` to return something that always waits one turn before resolving as a delegated promise?

Or just some encompassing behaviour for `Promise.resolve`?

PR #1096, opened 2022-02-28
Recast begone!

michaelfig:
While recast was good at trying to preserve existing formatting, it just wasn't cut out for line- and column-number preservation.

Rather than make more fragile hacks to it, I made a small change to `@babel/generator`:

```diff
diff --git a/packages/babel-generator/src/printer.ts b/packages/babel-generator/src/printer.ts
index bbdcc5b1ba..8383e6cf31 100644
--- a/packages/babel-generator/src/printer.ts
+++ b/packages/babel-generator/src/printer.ts
@@ -349,6 +349,15 @@ class Printer {
         this._newline();
       }
     }
+
+    // catch up to this nodes column if we're behind
+    if (pos?.column != null) {
+      const count = pos.column - this._buf.getCurrentColumn();
+
+      if (count > 0) {
+        this._append(" ".repeat(count));
+      }
+    }
   }
 
   /**
```

Unfortunately, this change doesn't result in really pretty code (spacing of punctuation is off), it just results in code that gives the correct line and column number in exceptions.

It took a lot more work to land this in `@agoric/babel-standalone`, but the investment in upgrading to Babel 7.17.6 (without any polyfills) should pay off.

Once this is released by Endo, an update to `@endo/bundle-source` and `@endo/static-module-record` in Agoric SDK will close Agoric/agoric-sdk#4681

### DIFF
--- a/packages/bundle-source/package.json
+++ b/packages/bundle-source/package.json
@@ -22,9 +22,9 @@
     "ava": "^3.12.1"
   },
   "dependencies": {
-    "@babel/generator": "^7.14.2",
-    "@babel/parser": "^7.14.2",
-    "@babel/traverse": "^7.14.2",
+    "@agoric/babel-generator": "^7.17.3",
+    "@babel/parser": "^7.17.3",
+    "@babel/traverse": "^7.17.3",
     "@endo/base64": "^0.2.19",
     "@endo/compartment-mapper": "^0.6.7",
     "@rollup/plugin-commonjs": "^19.0.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,10 +2,19 @@
 # yarn lockfile v1
 
 
+"@agoric/babel-generator@^7.17.3":
+  version "7.17.3"
+  resolved "https://registry.yarnpkg.com/@agoric/babel-generator/-/babel-generator-7.17.3.tgz#bcd1c157e6e4daa49aa8100aada61da0a96c29ca"
+  integrity sha512-PuOwEFCy7Y9NTSGBEzF7BNG1cIaZxwFd2agPQ41bD/VOIsnq/Me4Bn4t5PGCCy74TyyQGVfh6YfF5rawizRPmQ==
+  dependencies:
+    "@babel/types" "workspace:^"
+    jsesc "condition: BABEL_8_BREAKING ? ^3.0.2 : ^2.5.1"
+    source-map "^0.5.0"
+
 "@agoric/babel-standalone@^7.14.3":
-  version "7.14.3"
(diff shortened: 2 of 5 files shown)


kriskowal (approved):

packages/static-module-record/src/babelPlugin.js
```diff
@@ -253,31 +239,33 @@ function makeModulePlugins(options) {
 
       // Create the export calls.
       const isConst = decl.kind === 'const';
-      const replacements = rewriteVars(
+      const replace = rewriteVars(
```
kriskowal, line 242: I did like your rename from replace to replacements, which looks like it’s caught in a revert here.

packages/bundle-source/package.json
```diff
@@ -22,9 +22,9 @@
     "ava": "^3.12.1"
   },
   "dependencies": {
-    "@babel/generator": "^7.14.2",
-    "@babel/parser": "^7.14.2",
-    "@babel/traverse": "^7.14.2",
+    "@agoric/babel-generator": "^7.17.3",
```
mhofman, line 25: Have we considered using a resolutions instead / in addition?

kriskowal: Resolutions are useful, but unless I’m mistaken, don’t get inherited up to the application. I am hoping that this change makes duplication less of an existential problem too!

mhofman: Yeah it's unfortunately hard to tell if a dependency is dev, prod or both in `yarn.lock` alone.

yarn.lock
```diff
@@ -2,10 +2,19 @@
 # yarn lockfile v1
 
 
+"@agoric/babel-generator@^7.17.3":
+  version "7.17.3"
+  resolved "https://registry.yarnpkg.com/@agoric/babel-generator/-/babel-generator-7.17.3.tgz#bcd1c157e6e4daa49aa8100aada61da0a96c29ca"
+  integrity sha512-PuOwEFCy7Y9NTSGBEzF7BNG1cIaZxwFd2agPQ41bD/VOIsnq/Me4Bn4t5PGCCy74TyyQGVfh6YfF5rawizRPmQ==
+  dependencies:
+    "@babel/types" "workspace:^"
+    jsesc "condition: BABEL_8_BREAKING ? ^3.0.2 : ^2.5.1"
```
mhofman, line 11: @michaelfig is this an artifact of the publish step of `@agoric/babel-generator`? It seems like some templating didn't get applied properly

kriskowal: Thank you for catching this.

PR #1097, opened 2022-03-01
chore: small yarn.lock dedup

mhofman:
I've been trying to keep duplicate versions of babel away from `yarn.lock`. Manually deduplicated some entries that appeared after https://github.com/endojs/endo/pull/1096#issuecomment-1055808798

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -89,16 +89,7 @@
     semver "^5.4.1"
     source-map "^0.5.0"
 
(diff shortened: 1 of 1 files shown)


kriskowal:
@mhofman Am I wrong that this indicates a `yarn` bug is in play?

kriskowal (approved):

mhofman:
No, yarn is conservative and attempts to not introduce changes to existing packages. So when introducing a package with semver constraints that cannot be satisfied by existing packages, it adds a new package, but it does not remove the existing package even if the semver constraints that required it are also satisfied by the new package. Over time this leads to a lot of duplicate packages that can in practice be deduplicated without issues.

PR #1101, opened 2022-03-03
Pure Babel

michaelfig:
Remove our reliance on `@agoric/babel-standalone`, and instead compose `@babel/parser`, `@babel/traverse`, and `@agoric/babel-generator`.


kriskowal (approved):
Wonderful.

PR #1103, opened 2022-03-05
test(ses): Reenable and update third-party static module record test

kriskowal:
This lost test should have been enabled. The role it serves is to demonstrate the low-level API of a third-party static module record, without the additional complications specific to CommonJS.

### DIFF
--- a/packages/ses/test/test-import-non-esm.js
+++ b/packages/ses/test/test-import-non-esm.js
@@ -1,4 +1,5 @@
 import test from 'ava';
+import { StaticModuleRecord } from '@endo/static-module-record';
 import { resolveNode } from './node.js';
 import '../index.js';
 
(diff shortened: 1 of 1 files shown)


naugtur (approved):

packages/ses/test/test-import-non-esm.js
```diff
@@ -1,4 +1,5 @@
 import test from 'ava';
+import { StaticModuleRecord } from '@endo/static-module-record';
```
naugtur, line 2: What's the difference between this and the one that was previously available as a global?

kriskowal: Long ago, we moved `StaticModuleRecord` out of SES because it entrained too much of Babel. This code predates that migration and had to be brought back into the modern era. The proposal to TC39 includes `StaticModuleRecord` as a global, so this move is a reduction in the fidelity of the shim.

PR #1104, opened 2022-03-05
chore: release

kriskowal:
This will publish the new StaticModuleRecord that does not depend on babel-standalone.


michaelfig (approved):
LGTM!

PR #1107, opened 2022-03-10
fix(check-bundle): Immutable bundles and conveniences

### LINKED ISSUE #1106: checkBundle should assert that all examined properties are values, not accessors
Per https://github.com/Agoric/agoric-sdk/pull/4751#discussion_r822377931

`checkBundle` must be able to ensure that the given bundle is not only tamper-proof but genuinely immutable such that the caller can be sure that its properties remain coherent ever after.

To this end, `checkBundle` already asserts that it is frozen. It should go on and also assert that all of its relevant properties have `value` property descriptors with primitive values. Out of an abundance of caution, it would be sufficient to ensure that all own properties have `value` descriptors with `typeof` `"string"`.

kriskowal:
Fixes #1106

- fix(check-bundle)!: Fix batteries-included API
- fix(check-bundle)!: Verify immutability of checked bundle
- docs(check-bundle): Update release notes


warner (approved):
cool, more defensive == more better

PR #1111, opened 2022-03-11
fix(compartment-mapper): Fix "module" property in package.json

### LINKED ISSUE #1110: Compartment mapper should interpret package `module` as implicit `type: module`
Currently, the compartment mapper makes an exception for the `"module"` property in `package.json`, marking just that one file as an ESM, but leaving the rest of the package alone. The correct behavior would be to infer that the entire package is ESM and that the one given file corresponds to `"main"`.

kriskowal:
Fixes #1110


packages/compartment-mapper/src/infer-exports.js
```diff
@@ -81,20 +81,18 @@ export const inferExportsEntries = function* inferExportsEntries(
...
     const spec = relativize(module);
     types[spec] = 'mjs';
-    yield [name, spec];
-  }
-  if (browser !== undefined && tags.has('browser')) {
+    yield [name, relativize(module)];
+  } else if (browser !== undefined && tags.has('browser')) {
     yield* interpretBrowserExports(name, browser);
+  } else if (main !== undefined) {
+    yield [name, relativize(main)];
```
naugtur, line 95: This is a radical change of behavior for packages with both module and main defined. 

I understand that if we treat `module` as a flag declaring the entire package is mjs, the `main` field couldn't possibly work in most reasonable cases, but here's a snippet from package.json of `punycode`
```
  "name": "punycode",
  "version": "2.1.1",
  "main": "punycode.js",
  "jsnext:main": "punycode.es6.js",
  "module": "punycode.es6.js",
```

Judging from the behavior of my e2e test case, node.js loads "main" if required from cjs and "module" if imported from mjs and decides to do cjs or mjs parsing/execution based on that, not the extension (ambiguous) or global state (module field present).

Endo will not exhibit behaviors identical to node, because node has 2 module loading systems with independent cache that get bridged together (citation needed) while in endo we load each module once. 

I'm not objecting to the current change, I'm highlighting the need to decide how exactly we want to diverge from node.js in behavior.

naugtur (approved):

packages/compartment-mapper/src/infer-exports.js
kriskowal: Yeah, we should be deliberate in how Endo diverges from the behavior of Node. It won’t be possible to make everything work in Endo, because it’s possible for ESM and CJS facades of a single package to produce different APIs.

I think a principled approach requires us to select the weighted majority opinion not just of Node.js but ESM to CJS translators. In a scenario where a package is:

1. implemented as CJS
2. changed to ESM but compiled down to an equivalent CJS facade, preserving compatibility with existing CJS dependees
3. changed so ESM source and CJS facade are used depending on whether the dependee is ESM or CJS

We want Endo generally to be able to tolerate the package at any stage of that adventure, as often as that is possible. I think that, if a package provides ESM sources, we’re obliged to use them (that is at least consistent with *this* change). And one of the difficulties we will run up against is that different down-to-cjs-translators do not all generate the same CJS. In those cases, we need to make mitigations possible, but not necessarily easy. I am thinking of the hacks like `import * as name from 'name'; const api = name.default ? name.default.name || name : name` or some such, which are horrible but live in the wild.

naugtur: I'm trying to get a combination of those in parse-cjs to return the expected shape in esm and cjs (which differ) and it's not easy. I'm still hoping to make the ?.default ugliness go away.

PR #1112, opened 2022-03-11
fix(compartment-mapper): Avoid some property override pitfalls

kriskowal:
Using `{__proto__: null}` allows assignment to any property except `__proto__` when `Object.prototype` is frozen. Otherwise, assigning to `hasOwnProperty`, `toString`, &c would all cause errors. No purpose is served by changing these to Maps since all keys need to survive a JSON round-trip, but using a null prototype for any object-as-dict that will grown by assignment helps considerably.


kriskowal:
cc @michaelfig for witnessing related issue #1109 and prompting this change.

michaelfig (approved):

erights:
> Using `{__proto__: null}` allows assignment to any property except `__proto__` when `Object.prototype` is frozen. Otherwise, assigning to `hasOwnProperty`, `toString`, &c would all cause errors. No purpose is served by changing these to Maps since all keys need to survive a JSON round-trip, but using a null prototype for any object-as-dict that will grown by assignment helps considerably.

@kriskowal 
I haven't tested, but I believe the "except `__proto__`" above is incorrect. In the JavaScript expression `{__proto__: null}`, the `__proto__` is syntax, not a property name, and determines only what object the new object inherits from. In this case, as intended, nothing. Whereas in an assignment `x.__proto__ = y`, the `__proto__` is only a property name. For an object inheriting from `Object.prototype`, it normally inherits the `__proto__` accessor and thereby invokes the setter. But in an object inheriting from nothing and without its own `__proto__` property, it should just create a new own property without anything unusual happening.

PR #1114, opened 2022-03-15
fix(ses): Handle well-known symbol look-alikes

mhofman:
While working on fixing the `async_hooks` issue, I realized that the whitelisting mechanism for well-known symbols can be tricked into accepting a unique symbol that just has the description of the expected well-known symbol, but is actually unique.

This PR changes the whitelist parsing logic to generate a map of well-known symbols from the intrinsics and the permit list (similar to marshal's logic), and checks that a symbol matches the well-known symbol when found in the permits of an object. It does so by returning a name in the shape of `UniqueSymbol(description)` when finding a symbol that isn't well-known, and only returning the `@@name` if the symbol is found in the map. This technically would allow unique symbols to be permitted on intrinsics if named similarly (there is technically a chance of collision with string props, but that was already the case with the `@@` prefix anyway).

Added a unit test for this (renaming the existing test, which seemed to be testing a different file). Verified manually that before the fix, the test would fail (didn't create separate commit with failing test)


packages/ses/src/whitelist-intrinsics.js
```diff
@@ -86,9 +72,48 @@ export default function whitelistIntrinsics(
...
+      return prop;
+    }
+
+    const wellKnownSymbol = mapGet(wellKnownSymbolNames, prop);
+
+    if (typeof prop === 'symbol') {
+      if (wellKnownSymbol) {
+        return wellKnownSymbol;
+      } else {
+        return `Unique${String(prop)}`;
```
gibson042, line 110: It would be nice to document the nature of `UniqueSymbol(…)` (and string representation of property keys in general) somewhere, although I'm not sure where.

packages/ses/test/test-whitelist-intrinsics.js
```diff
@@ -1,47 +1,41 @@
...
+    // Yes, this is clearly not inert
+    '%InertFunction%': Function.prototype.constructor,
+    '%ObjectPrototype%': Object.prototype,
+    Object,
+  });
+  whitelistIntrinsics(intrinsics, () => {});
+  t.is(
+    ArrayProto[RogueSymbolIterator],
+    undefined,
+    `Well-known Symbol look-alike should have been removed`,
```
gibson042, line 39: Should there also be an assertion about `SymbolIterator`, e.g. that it is not removed from an object where it is expected?

gibson042 (approved):

erights (approved):
LGTM, thanks!

packages/ses/test/test-whitelist-intrinsics.js
mhofman: done

mhofman:
@erights after our discussion I added support for Registered symbols in https://github.com/endojs/endo/pull/1114/commits/6f6fffb3875fb91708996333f51647a6a447dfad

packages/ses/src/whitelist-intrinsics.js
mhofman: done in https://github.com/endojs/endo/commit/6f6fffb3875fb91708996333f51647a6a447dfad#diff-a241982c4634b07275319f160e1d3cb9b89929217019708fae1b5ea3bb4e8e77

packages/ses/src/commons.js
```diff
@@ -72,6 +72,8 @@ export const {
   toStringTag: toStringTagSymbol,
   iterator: iteratorSymbol,
   matchAll: matchAllSymbol,
+  keyFor: SymbolKeyFor,
+  for: SymbolFor,
```
kriskowal, line 76: Elsewhere, as in `arrayPrototype`, there’s precedent for camelCase over PascalCase.

kriskowal (approved):

packages/ses/src/commons.js
mhofman: Actually thinking about this more, that means we don't differentiate between static and proto methods?

PR #1115, opened 2022-03-15
Handle symbols installed on Promise by Node's `async_hooks`

### LINKED ISSUE #1105: Node bug causes failures within the vscode Debug Console
https://github.com/nodejs/node/issues/42229

mhofman:
Closes #1105 
Layers on top of #1114 

Node 14.18 and 16.6[^0] introduced a change (https://github.com/nodejs/node/pull/39135) to `async_hooks` that enabled the fast-path for promise hooks even in the presence of a `destroy` hook. The fast-path logic removes the wrapper object which was previously used to hold the async hooks metadata for each promise instance, such as async ids and a property bag related to the liveness of the promise instance, causing these to be attached directly to the promise instance as own symbols.

[^0]: https://github.com/nodejs/node/issues/42229#issuecomment-1062145546

When a debugger enables async stack traces (`Debugger.setAsyncCallStackDepth({maxDepth})` with `maxDepth` > 0), node's inspector support internally enables `async_hooks` with a `destroy` hook [^1]  [^2] [^3], causing all promises to gain these symbols.

[^1]: https://github.com/nodejs/node/blob/d4d0b09122d91478881ee7fd36eb3c9f78f8452d/src/inspector_agent.cc#L411-L424
[^2]: https://github.com/nodejs/node/blob/d4d0b09122d91478881ee7fd36eb3c9f78f8452d/lib/internal/bootstrap/pre_execution.js#L303-L316
[^3]: https://github.com/nodejs/node/blob/master/lib/internal/inspector_async_hook.js

The `init` promise hooks is the main source of symbol attachments. It runs synchronously right after the constructor returns and before the caller gets the reference. While that would normally cause all promise instances to gain these symbols as own property before the program gets a chance to freeze them, other promise hooks enforce the presence of async ids on existing promise objects. If `async_hooks` is enabled after a promise has already been created and frozen, for example by attaching the debugger later, it will fail to attach these symbols, throwing a `TypeError` synchronously.

Until node removes these symbols as own property, or handles assignment failure (see https://github.com/nodejs/node/issues/42229), we workaround the issue by installing accessors for these symbols on the `Promise.prototype`, which defines them as an own property (non-enumerable to avoid polluting the console output), and falls-back to a `WeakMap` for non-extensible promise instances.

The accessors handle both async ids, and the `destroyed` symbol. While the latter is not strictly necessary as it's always installed during `init` and never late, it's more consistent to handle it similarly to async ids, and has the added benefit of removing enumerable pollution.

The `async_hooks` logic is added to `@endo/init` because of the dependency on a node import. `lockdown` is updated to whitelist these symbols on the promise prototype. It would be better to have `lockdown` remove these by default and require a taming option to leave them in place, but I didn't find a way to thread that into the whitelist logic to make it conditional.


packages/init/node-async_hooks.js
```diff
@@ -0,0 +1,226 @@
...
+  set(value) {
+    if (Object.isExtensible(this)) {
+      Object.defineProperty(this, symbol, {
+        value,
+        writable: false,
+        configurable: false,
+        enumerable: false,
+      });
+    } else {
+      // console.log('fallback set of async id', symbol, value, new Error().stack);
```
naugtur, line 178: tip. Especially if you're dealing with async_hooks, use `process._rawDebug` instead of `console.log/error`  
console API is triggering hooks, so if you end up using console in an async hook you get infinite loop.   
_rawDebug is the only really synchronous thing that writes output.

packages/init/node-async_hooks.js
```diff
@@ -0,0 +1,226 @@
...
+    if (Symbol.nodeAsyncHooksDestroyed) {
+      state[Symbol.nodeAsyncHooksDestroyed] = undefined;
+    }
+    promiseAsyncHookFallbackStates.set(promise, state);
+  }
+  return state;
+};
+
+const setAsyncIdFallback = (promise, symbol, value) => {
+  const state = getAsyncHookFallbackState(promise, true);
```
naugtur, line 155: (nitpick) 3 function signatures here have boolean arguments, each meaning different things. using an options object where the boolean is named in the caller as well would limit the lookups one has to make while reading the body of the current function to remember what the boolean did.

packages/init/node-async_hooks.js
```diff
@@ -0,0 +1,226 @@
...
+
+    // In some versions of Node, async_hooks don't give access to the resource
+    // itself, but to a "wrapper" which is basically hooks metadata for the promise
+    const promisesData = bootstrapData.filter(
+      ({ resource }) => Promise.resolve(resource) === resource,
+    );
+    bootstrapData.length = 0;
+    const { length } = promisesData;
+    if (length > 1) {
+      // console.warn('Found multiple potential candidates');
```
naugtur, line 82: Are you aware of a situation when this would happen? AFAIR it shouldn't be possible.

mhofman: I did see it happen at one point when I was exploring, and then couldn't reproduce. not sure why exactly, but I kept the check around.

packages/init/node-async_hooks.js
```diff
@@ -0,0 +1,226 @@
...
+    // Normally all promise hooks are disabled in a subsequent microtask
+    // That means Node versions that modify promises at init will still
+    // trigger our proto hooks for promises created in this turn
+    // The following trick will disable the internal promise init hook
+    // However, only do this for destroy modifying versions, since some versions
+    // only modify promises if no destroy hook is requested, and do not correctly
+    // reset the internal init promise hook in those case. (e.g. v14.16.2)
+    const resetHook = createHook({});
+    resetHook.enable();
+    resetHook.disable();
```
naugtur, line 105: This is

packages/init/node-async_hooks.js
mhofman: That explains the infinite loop I got when I had a logging call in an `after` hook!

packages/init/node-async_hooks.js
naugtur: It would mean the init hook was called twice. I'd consider that a bug.

mhofman: I think it created multiple promise instances for some reason.

packages/init/node-async_hooks.js
```diff
@@ -0,0 +1,229 @@
...
+    //   process._rawDebug(`Found ${length} symbols on promise:`, ...symbols);
+    // }
+    symbols.forEach(symbol => {
+      const value = resource[symbol];
+      let type;
+      if (value === asyncId) {
+        type = 'async_id_symbol';
+      } else if (value === triggerAsyncId) {
+        type = 'trigger_async_id_symbol';
+      } else if (typeof value === 'object' && 'destroyed' in value) {
```
gibson042, line 119: Should this be an own check?
```suggestion
      } else if (typeof value === 'object' && Reflect.getOwnPropertyDescriptor(value, 'destroyed') !== undefined) {
```

packages/init/test/test-async_hooks.js
```diff
@@ -0,0 +1,79 @@
...
+    gc = () => {
+      Array.from({ length: 2 ** 24 }, () => Math.random());
+    };
+  }
+  return gc;
+})();
+
+test('async_hooks Promise patch', async t => {
+  const hasSymbols =
+    Symbol.nodeAsyncHooksAsyncId && Symbol.nodeAsyncHooksTriggerAsyncId;
```
gibson042, line 20:
```suggestion
    Boolean(Symbol.nodeAsyncHooksAsyncId && Symbol.nodeAsyncHooksTriggerAsyncId);
```

packages/init/test/test-async_hooks.js
```diff
@@ -0,0 +1,79 @@
...
+      // after() {},
+      destroy() {},
+    });
+    testHooks.enable();
+
+    // Create a promise with symbols attached
+    const p3 = Promise.resolve();
+    if (hasSymbols) {
+      t.truthy(Reflect.ownKeys(p3));
+    }
```
gibson042, line 53: Isn't the array returned by `Reflect.ownKeys` _always_ truthy, even when empty?
```suggestion
    t.is(Reflect.ownKeys(p3).length > 0, hasSymbols);
```

packages/init/test/test-async_hooks.js
```diff
@@ -0,0 +1,79 @@
...
+      // async_hooks will attempt to get the asyncId from `p1`
+      // which was created and frozen before the symbols were installed
+      const ret = p1.then(() => {});
+      // Trigger attempting to get asyncId of `p1` again, which in current
+      // node versions will fail and generate a new one because of an own check
+      p1.then(() => {});
+
+      if (hasSymbols) {
+        t.truthy(Reflect.ownKeys(ret));
+      }
```
gibson042, line 67:
```suggestion
      t.is(Reflect.ownKeys(ret).length > 0, hasSymbols);
```

packages/ses/src/whitelist.js
```diff
@@ -501,6 +501,9 @@ export const whitelist = {
     keyFor: fn,
     match: 'symbol',
     matchAll: 'symbol',
+    nodeAsyncHooksAsyncId: 'symbol',
+    nodeAsyncHooksTriggerAsyncId: 'symbol',
+    nodeAsyncHooksDestroyed: 'symbol',
```
gibson042, line 506:
```suggestion
    // Non-standard, used in node to prevent async_hooks from breaking
    nodeAsyncHooksAsyncId: 'symbol',
    nodeAsyncHooksTriggerAsyncId: 'symbol',
    nodeAsyncHooksDestroyed: 'symbol',

```

gibson042 (approved):
LGTM modulo some nits

packages/init/node-async_hooks.js
```diff
@@ -0,0 +1,229 @@
...
+};
+
+const promiseAsyncHookFallbackStates = new WeakMap();
+
+const setAsyncSymbol = (description, symbol) => {
+  const wellKnownName = asyncHooksWellKnownNameFromDescription[description];
+  if (!wellKnownName) {
+    throw new Error('Unknown symbol');
+  } else if (!Symbol[wellKnownName]) {
+    Symbol[wellKnownName] = symbol;
```
erights, line 18: I am uncomfortable making these weird symbols into apparent well-known symbols by adding them as statics to the `Symbol` constructor. Let's discuss before we go forward with this.

erights (changes requested):
I am uncomfortable making these weird symbols into apparent well-known symbols by adding them as statics to the `Symbol` constructor. Let's discuss before we go forward with this.

packages/init/test/test-async_hooks.js
mhofman: Yup, I did a last minute change and missed some things.

packages/init/node-async_hooks.js
mhofman: I think that's overkill

packages/ses/src/whitelist.js
mhofman: No longer relevant

mhofman:
@erights I switched away from making the async hooks symbols registered and using the new ability to name unique symbols in the whitelist. PTAL

mhofman:
It was brought up that `@endo/init` may be used in non-Node environments, in which case we'll need either different entrypoints for each environment, or a shim for `async_hooks`, with the following shape:

```js
export class AsyncResource {};
export const createHook = () => ({ enable() {}, disable() {} });
```

mhofman:
I've updated the PR to only include the async_hooks patch on node based on conditional exports, and included a test, but I'm getting an apparently unrelated failure:

```
Cannot read property 'buildError' of undefined
```

attn @kriskowal

packages/init/pre-remoting.js
```diff
@@ -1,5 +1,7 @@
 // pre-remoting.js - shims necessary to use @endo/far
 
-import './pre.js';
+// Use a package self-reference to go through the "exports" resolution
+// eslint-disable-next-line import/no-extraneous-dependencies
+import '@endo/init/pre.js';
```
mhofman, line 5: Compartment mapper isn't happy about this and fails with:

```
Cannot find external module "@endo/init/pre.js" in package file:///workspaces/agoric/endo/packages/init/
```

mhofman: Filed https://github.com/endojs/endo/issues/1121

packages/ses/package.json
```diff
@@ -35,11 +35,11 @@
   "types": "./index.d.ts",
   "exports": {
     ".": {
-      "import": "./dist/ses.umd.js",
```
mhofman, line 38: Should there be a `"browser"` fallback (after import and require) ?

mhofman:
> @mhofman I’ve added a commit that should unblock the build with `makeArchive`.

Well except I've taken a reliance on self references. I think I'll move the async hooks logic in a separate taming package that endo/init takes a dependency on for now?

mhofman:
@erights PTAL

erights (changes requested):
I just noticed that all the normal source files of this `init` package are at top level in the package, rather than nested in a `src` directory. Please move the normal source files into a `src` directory.

Other than that, my review is still in progress, but looking good to me so far.

erights (approved):
Review done. Everything I understood LGTM except the directory structure. I did look at all the node/async_hook manipulation code, and it seemed fine to the extent I understood it. Mostly I was amazed at how many cases you had to wrestle with!

mhofman:
> I just noticed that all the normal source files of this `init` package are at top level in the package, rather than nested in a `src` directory. Please move the normal source files into a `src` directory.

I believe the files were in the root because they are all entrypoints. While `package.json` `"exports"` now allows us to remap this anywhere, I think it's best to keep the regular entrypoints in place for tools which don't support remapping. I have however moved all other/new files into `src`. @kriskowal does the structure look good by you?

kriskowal:
> > I just noticed that all the normal source files of this `init` package are at top level in the package, rather than nested in a `src` directory. Please move the normal source files into a `src` directory.
> 
> I believe the files were in the root because they are all entrypoints. While `package.json` `"exports"` now allows us to remap this anywhere, I think it's best to keep the regular entrypoints in place for tools which don't support remapping. I have however moved all other/new files into `src`. @kriskowal does the structure look good by you?

Yes, @michaelfig and I have agreed that, due to limitations of interoperability between Node.js versions that recognize `exports` and other systems that do not yet, we will place all entry-point modules (modules that are public) at the top level, and any other module under `src` as an indication that it is private.

packages/init/package.json
```diff
@@ -7,7 +7,10 @@
   "exports": {
     ".": "./index.js",
     "./debug.js": "./debug.js",
-    "./pre.js": "./pre.js",
+    "./pre.js": {
+      "node": "./src/pre-node.js",
```
kriskowal, line 11: This is a fine exception to the rule I mention, because it only applies in cases where `exports` are respected.

kriskowal (approved):
I approve in terms of general structure and willingness to maintain. I don’t begin to grasp the implications of `async_hooks`, but the effect of this change seems good to me.

PR #1120, opened 2022-03-16
Handle user code catching errors from require

### LINKED ISSUE #1116: Catching a missing package error leaves memoized parent package in a corrupted state in cjs
To reproduce, try getting the keccak package to load in Endo.

What happens there:  
- ./lib/api module is loaded as part of loading a module depending on streams and bindings
- loading fails
- catch falls back to a different implementation, which also loads ./lib/api but gets an empty uninitialized exports object from memoization

naugtur:
fixes #1116

I didn't observe negative consequences of calling activate() more than once, but that could be avoided.

ALso, it would be possible to hold on to the error and rethrow it instead of running execute each time, but I think that could get cached higher the dependency chain than needed.

### DIFF
--- a/packages/compartment-mapper/test/fixtures-error-handling/node_modules/catch/README.md
+++ b/packages/compartment-mapper/test/fixtures-error-handling/node_modules/catch/README.md
@@ -1,0 +1,1 @@
+Catches the error thrown by require and tries again
--- a/packages/compartment-mapper/test/fixtures-error-handling/node_modules/catch/getmissing.js
+++ b/packages/compartment-mapper/test/fixtures-error-handling/node_modules/catch/getmissing.js
@@ -1,0 +1,2 @@
+module.exports = 42;
+require('obviouslymissing');
--- a/packages/compartment-mapper/test/fixtures-error-handling/node_modules/catch/main.js
+++ b/packages/compartment-mapper/test/fixtures-error-handling/node_modules/catch/main.js
@@ -1,0 +1,5 @@
+try {
+  const a = require('./getmissing');
+} catch (e) {
+  const a = require('./getmissing');
+}
--- a/packages/compartment-mapper/test/fixtures-error-handling/node_modules/catch/package.json
+++ b/packages/compartment-mapper/test/fixtures-error-handling/node_modules/catch/package.json
@@ -1,0 +1,5 @@
+{
+  "name": "catch",
+  "version": "1.0.0",
+  "main": "./main.js"
+}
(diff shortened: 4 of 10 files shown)


packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/defineprop/example-elliptic.js
```diff
@@ -23,5 +23,5 @@ function defineCurve(name) {
   });
 }
 
-defineCurve('p192')
-defineCurve('secp256k1')
\ No newline at end of file
+defineCurve('p192');
+defineCurve('secp256k1');
```
naugtur, line 27: just prettier

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/defineprop/main.js
```diff
@@ -1,17 +1,19 @@
 const curves = require('./example-elliptic');
 const ethereum = require('./example-ethereum-util');
```
naugtur, line 3: just prettier

kriskowal (changes requested):
I think I would prefer to avoid the side-effects of attempting to initialize a module twice. As you suggest, capturing and rethrowing the error would suffice. Otherwise, this is a great fix.

kriskowal:
From our conversation out-of-band, consider dependency graph A imports B imports C. If you import A, you’re arranging for a working set that will initialize in order C, B, A. My expectation, which requires confirmation, is that if B were to throw an error during initialization, that error would contaminate A but not C. So, any subsequent import that depends on B or C would replay the same error, but A would be fine. No module should ever be initialized twice, or initialized at all if one of its dependencies failed to initialize.

erights:
> From our conversation out-of-band, consider dependency graph A imports B imports C. If you import A, you’re arranging for a working set that will initialize in order C, B, A. My expectation, which requires confirmation, is that if B were to throw an error during initialization, that error would contaminate A but not C. So, any subsequent import that depends on B or C would replay the same error, but A would be fine. No module should ever be initialized twice, or initialized at all if one of its dependencies failed to initialize.

I can confirm this. In addition, if a module that is within an import cycle throws an error, it contaminates everything in its cycle, even if some others had started earlier.

Put another way, any directed graph can be considered to be a directed acyclic graph of *strongly connected components*. A failure contaminates its own strongly connected component and all dependent strongly connected components.

erights (commented):
This seems like a fine change, and is certainly well motivated. However, I'm going to defer any approval decision to @kriskowal .

kriskowal (approved):

Issue #1123, opened 2022-03-17
Compartment Mapper parity for Node.js module ecosystem


kriskowal:
Compartment mapper’s inferExports.js gets a number of things inconsistent with Node.js 

1. done ~Self references https://github.com/endojs/endo/issues/1121~
2. done ~Arrays of conditions https://github.com/endojs/endo/issues/1122~
3. done ~Conditions are currently called `tags` both in the compartment-map.json schema and in the implementation. This will need to be gracefully migrated.~
4. done ~Priority: an object must either be entirely conditions or entirely mappings. The first must win, not the last.~
5. done ~There is a `default` condition that is implicitly present.~
6. done ~There must be an `endo` condition.~
7. see sub-issues

PR #1125, opened 2022-03-17
fix: Support default conditions and avoid SES double bundle

kriskowal:
Refs https://github.com/endojs/endo/issues/1123

- fix(ses): Do not bundle modules for use as modules
- fix(endo): Ensure conditions include default, import, and endo

### DIFF
--- a/packages/ses/package.json.md
+++ b/packages/ses/package.json.md
@@ -98,6 +98,9 @@ The variations differ only in file name extension.
 
 Node.js and other tools will use this file when importing `ses` as an ESM.
 (JavaScript module).
+We have in the past experimented with using the precompiled bundle of SES here
+(`./dist/ses.cjs`), but found that this interacted poorly with Endo,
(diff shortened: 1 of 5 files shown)


packages/ses/package.json.md
```diff
@@ -98,6 +98,9 @@ The variations differ only in file name extension.
 
 Node.js and other tools will use this file when importing `ses` as an ESM.
 (JavaScript module).
+We have in the past experimented with using the precompiled bundle of SES here
+(`./dist/ses.cjs`), but found that this interacted poorly with Endo,
```
mhofman, line 102: We were using the UMD version, not the CommonJS version.

```suggestion
(`./dist/ses.umd.js`), but found that this interacted poorly with Endo,
```

kriskowal: Addressed with “(`./dist/ses.cjs` or `./dist/ses.umd.js`)” since these files are identical)

mhofman (approved):

PR #1131, opened 2022-03-22
refactor(compartment-mapper): rebuild cjs exports - module.exports is the default export

naugtur:
This finally solves the need to cater for default export being implicit.


packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
```diff
@@ -0,0 +1,111 @@
...
+export const wrap = (moduleEnvironmentRecord, compartment, resolvedImports) => {
+  // This initial default value makes things like exports.hasOwnProperty() work in cjs.
+  moduleEnvironmentRecord.default = create(
+    compartment.globalThis.Object.prototype,
+  );
+
+  // Set all exported properties on the defult and on the namespace simultaneously for import *
+  // Root namespace is only accessible for imports. Requiring from cjs gets the default field of the namespace
+  const assignProp = (prop, value) => {
+    // TODO Consider __esModule handling
```
naugtur, line 35: There's a separate issue for this. Also, I don't see what we should do differently to handle it at this time.

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
```diff
@@ -0,0 +1,111 @@
+// @ts-check
```
naugtur, line 1: I got tired of copy-pasting between parse-cjs and parse-pre-cjs and disabling archive tests every time I work on the logic.

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
```diff
@@ -0,0 +1,111 @@
...
+    },
+    set exports(value) {
+      finalExports = value;
+    },
+  });
+
+  const require = freeze((/** @type {string} */ importSpecifier) => {
+    const namespace = compartment.importNow(resolvedImports[importSpecifier]);
+    // If you read this file carefully, you'll see it's not possible for a cjs module to not have the default anymore.
+    // It's currently possible to require modules that were not created by this file though.
```
naugtur, line 83: so now `require('a-package-that-has-esm-exports')` will get the default if present and the entire namespace if not. 

This may be controversial, but at this time it seems to be working great (at least for examples I've got in e2e tests)

If we want to fix it, we'd have to make `require` get the cjs variant of the package, parsed as such.

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
kriskowal: It is possible that `__esModule` was only helpful for ESM on CJS, not CJS on ESM. It would be good to rule that possibility out definitively.

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
kriskowal: This is fine. We could on the one hand create separate compartments for the CJS or ESM interpretation of a single package and thread them accordingly, but that would risk causing identity discontinuity between ESM and CJS instances.

This is an option we might present to developers, something they can express in an app’s package.json, but I’m content to avoid the problem for now.

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
kriskowal: Fantastic.

kriskowal (approved):

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
naugtur: I'll start working on that later this week and I'll keep digging until I find an example that differs between using __esModule and not.

kriskowal: It might be easier to page into memory what problem `__esModule` was created to exist and check whether that problem exists on ESM. I could be convinced that we can ignore it with a very short proof.

PR #1134, opened 2022-03-30
fix(compartment-mapper): add support for alternatives in exports defnitions

naugtur:
There's no point in iterating the array, because in the top 1000+ packages with exports field defined as an object, all examples using an array define the same export twice in the array - once as an object with default/import alternatives and another as string.

We support both ways, so taking `[0]` is enough of an alternative choice algorithm. 

Edit: Iterating the array now, for correctness ;)


packages/compartment-mapper/src/infer-exports.js
```diff
@@ -34,6 +35,14 @@ function* interpretBrowserExports(name, exports) {
  * @yields {[string, string]}
  */
 function* interpretExports(name, exports, tags) {
+  if (isArray(exports)) {
+    yield* interpretExports(
+      name,
+      // Find one that produces non-empty result, discard result and use again
+      exports.find(ex => !interpretExports(name, ex, tags).next().done),
+      tags,
+    );
```
naugtur, line 44: Alternative implementation that doesn't perform one unnecessary operation just feels clunky
```js
    let sample,
      alternative,
      i = 0;
    do {
      alternative = interpretExports(name, exports[i], tags);
      sample = alternative.next();
      i++;
    } while (sample.done);

    yield sample.value;
    yield* alternative;
```
This is not a heavy operation, so I decided to go with spawning 2 copies of the successful generator.

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/nested-export/package.json
```diff
@@ -0,0 +1,22 @@
...
+    ".": [
+      {
+        "default": "./index.js"
+      },
+      "./index.js"
+    ],
+    "./callBound": [
+      {
+        "deno": "./this-gets-skipped"
+      },
```
kriskowal, line 14: ✅

packages/compartment-mapper/src/infer-exports.js
kriskowal: It took me a moment to read, but I agree it’s better than the alternative you propose. I had imagined something like:

```js
for (const section of exports) {
  const results = [...interpretExports(name, section, tags)];
  if (results.length > 0) {
    yield* results;
    break;
  }
}
```

I’m convinced that your solution is equivalent.

kriskowal (approved):

packages/compartment-mapper/src/infer-exports.js
naugtur: I'll replace with this to avoid having to look at `.next().done`. Thanks.

PR #1137, opened 2022-04-03
XS `Base64.decode` returns an `ArrayBuffer`, not `Uint8Array`

michaelfig:
This PR includes some changes I needed to do in order to debug trying to `ZipWriter.write(name, decodeBase64(str))`.

The solution is simple; just wrap the XS native decoder.

### DIFF
--- a/packages/base64/src/decode.js
+++ b/packages/base64/src/decode.js
@@ -60,5 +60,18 @@ const jsDecodeBase64 = (string, name = '<unknown>') => {
   return data.subarray(0, j);
 };
 
+// The XS Base64.decode function is faster, but might return ArrayBuffer (not
+// Uint8Array).  Adapt it to our needs.
+const adaptDecoder = nativeDecodeBase64 => (...args) => {
+  const decoded = nativeDecodeBase64(...args);
+  if (decoded instanceof Uint8Array) {
+    return decoded;
+  }
+  return new Uint8Array(decoded);
+};
+
+/** @type {typeof jsDecodeBase64} */
 export const decodeBase64 =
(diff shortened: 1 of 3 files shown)


packages/base64/src/decode.js
```diff
@@ -16,7 +16,7 @@ import { monodu64, padding } from './common.js';
  * @param {string} string Base64-encoded string
  * @param {string} [name] The name of the string as it will appear in error
  * messages.
- * @returns {Uint8Array} decoded bytes
+ * @returns {Uint8Array} decoded bytes (array buffer on XS)
```
kriskowal, line 19:
```suggestion
 * @returns {Uint8Array} decoded bytes
```

packages/base64/src/decode.js
```diff
@@ -60,5 +60,18 @@ const jsDecodeBase64 = (string, name = '<unknown>') => {
   return data.subarray(0, j);
 };
 
+// The XS Base64.decode function is faster, but might return ArrayBuffer (not
+// Uint8Array).  Adapt it to our needs.
```
kriskowal, line 64: Does XS return ArrayBuffer, Uint8Array, or either depending? I would find it weird if the signature varied. The conditional short might not be necessary.

kriskowal (approved):
How did this work before this change? Is there an XS bug in Base64.decode that makes the return type vary by size?

packages/base64/src/decode.js
michaelfig: It currently returns ArrayBuffer.  I don't want to be surprised if/when it used to return Uint8Array or will do so in the future.

Issue #291, opened 2020-05-05
Compartment support for dynamic import and importMetaHook


kriskowal:
*[update April 20, 2022]:*

# Description

SES shim should support the features of dynamic import (`import`) and import metadata (`import.meta`). These must be implemented in a manner that is host neutral and afford the possibility of lazy, on-demand initialization of `import.meta` for use cases like `import.meta.resolve` that would be prohibitively expensive if constructed for modules that do not utter `import.meta` in their source.

# Design

In `@endo/static-module-record` we employ a source to source transformation that produces a “first-party static module record” that `ses` understands how to load, link, and initialize when returned through the `importHook`. In SES, the `module-instance.js` module is responsible for the runtime component for evaluating modules. We must extend this calling convention to support dynamic import and import metadata.

Also, because static module records are committed to archives and stored, the calling convention for modules generated by older versions of `@endo/static-module-record` must be preserved.

I propose that we do this by rewriting dynamic `import(specifier)` and `import.meta` in the module-to-program transformation to forms like `$h_import()` and `$h_import_meta`, including the ZWJ so the hidden symbol censorship still applies. The module-instance.js linker must pass suitable values for these into the generated module functor. The dynamic import function must be a closure that applies the `resolveHook` relative to the current `moduleSpecifier`.

Since creating `$h_import_meta` and `$h_import` are avoidable if they were not uttered by the source, or compiled by an earlier version of `@endo/static-module-record`, the newer `StaticModuleRecord` constructor should add new properties to the record, like `__usesImport` and `__usesImportMeta`, as an indication for whether these should be made.

`$h_import_meta` must be an object with a `null` prototype.

The `importHook` (to be called `loadHook` in a future revision #420) will need to be able to return a “module record envelope” that has an optional `meta` property. The `record.meta` property is an object to copy to the `$h_import_meta` object, either with the semantics of `Object.assign($h_import_meta, record.meta)` or `Object.defineProperties($h_import_meta, Object.getOwnPropertyDescriptors(record.meta))`, on which we need consensus among the champions of the [TC39 Compartment Proposal](https://github.com/tc39/proposal-compartments).

We would, either in this change or a future change to be factored out of this issue, also implement a `importMetaHook`. The `importMetaHook` would receive `$h_import_meta` and have an opportunity to augment it and freeze it before the object is revealed to guest code. The module machinery must only call the hook if `import.meta` is uttered in the module code. The host function must not be called on the stack of guest code.

[...]

PR #1141, opened 2022-04-05
feat(static-module-record): add support for import.meta

naugtur:
Implemented import.meta according to #291

The actual implementation of exposing import.meta.url to modules loaded by compartment-mapper is going to be a separate PR.

### DIFF
--- a/packages/compartment-mapper/src/bundle.js
+++ b/packages/compartment-mapper/src/bundle.js
@@ -332,9 +332,7 @@ ${importsCellSetter(__liveExportMap__, index)}\
     onceVar: {
 ${importsCellSetter(__fixedExportMap__, index)}\
     },
-    importMeta: {
-      url: '${packageLocation}'\
-    },
+    importMeta: {},
   });
 `,
   ),
--- a/packages/compartment-mapper/src/bundle.js
+++ b/packages/compartment-mapper/src/bundle.js
@@ -332,7 +332,7 @@ ${importsCellSetter(__liveExportMap__, index)}\
     onceVar: {
 ${importsCellSetter(__fixedExportMap__, index)}\
     },
-    meta: {
+    importMeta: {
       url: '${packageLocation}'\
     },
   });
--- a/packages/compartment-mapper/src/bundle.js
+++ b/packages/compartment-mapper/src/bundle.js
@@ -332,7 +332,7 @@ ${importsCellSetter(__liveExportMap__, index)}\
     onceVar: {
 ${importsCellSetter(__fixedExportMap__, index)}\
     },
-    metaVar: {
+    meta: {
       url: '${packageLocation}'\
     },
   });
--- a/packages/compartment-mapper/src/bundle.js
+++ b/packages/compartment-mapper/src/bundle.js
@@ -332,6 +332,9 @@ ${importsCellSetter(__liveExportMap__, index)}\
     onceVar: {
 ${importsCellSetter(__fixedExportMap__, index)}\
     },
+    metaVar: {
+      url: '${packageLocation}'\
+    },
   });
 `,
   ),
--- a/packages/ses/README.md
+++ b/packages/ses/README.md
@@ -419,7 +419,9 @@ A compiled static module record has the following shape:
       exports to notifier functions.
       A notifier function accepts an update function and registers
       to receive updates for the value exported by the other module.
-    - `importMeta` is a null-prototype object with keys transferred from `meta` property in the envelope returned by importHook and/or mutated by calling `importMetaHook(importMeta)`
+    - `importMeta` is a null-prototype object with keys transferred from `importMeta`
+      property in the envelope returned by importHook and/or mutated by
+      calling `importMetaHook(moduleSpecifier, importMeta)`
     - `liveVar` is a record that maps names exported by this module
       to a function that may be called to initialize or update
       the corresponding value in another module.
--- a/packages/ses/README.md
+++ b/packages/ses/README.md
@@ -412,13 +412,14 @@ A compiled static module record has the following shape:
   an initialization record and initializes the module.
   This property distinguishes this type of module record.
   The name implies a future record type that supports top-level await.
-  - An initialization record has the properties `imports`, `liveVar`, and
+  - An initialization record has the properties `imports`, `liveVar`, `importMeta` and
     `onceVar`.
     - `imports` is a function that accepts a map from partial import
       module specifiers to maps from names that the corresponding module
       exports to notifier functions.
       A notifier function accepts an update function and registers
       to receive updates for the value exported by the other module.
+    - `importMeta` is a null-prototype object with keys transferred from `meta` property in the envelope returned by importHook and/or mutated by calling `importMetaHook(importMeta)`
     - `liveVar` is a record that maps names exported by this module
       to a function that may be called to initialize or update
       the corresponding value in another module.
--- a/packages/ses/src/module-instance.js
+++ b/packages/ses/src/module-instance.js
@@ -127,14 +127,14 @@ export const makeModuleInstance = (
     compartment,
     moduleSpecifier,
     staticModuleRecord,
-    meta: moduleRecordMeta,
+    importMeta: moduleRecordMeta,
   } = moduleRecord;
   const {
     reexports: exportAlls = [],
     __syncModuleProgram__: functorSource,
     __fixedExportMap__: fixedExportMap = {},
     __liveExportMap__: liveExportMap = {},
-    __usesImportMeta__: usesImportMeta = false,
+    __needsImportMeta__: needsImportMeta = false,
   } = staticModuleRecord;
 
   const compartmentFields = weakmapGet(privateFields, compartment);
@@ -169,8 +169,8 @@ export const makeModuleInstance = (
   if (moduleRecordMeta) {
     assign(importMeta, moduleRecordMeta);
   }
-  if (usesImportMeta && importMetaHook) {
-    importMetaHook(importMeta);
+  if (needsImportMeta && importMetaHook) {
+    importMetaHook(moduleSpecifier, importMeta);
   }
 
   // {_localName_: [{get, set, notify}]} used to merge all the export updaters.
--- a/packages/ses/src/module-instance.js
+++ b/packages/ses/src/module-instance.js
@@ -165,12 +165,12 @@ export const makeModuleInstance = (
   // both initialize and update live bindings.
   const liveVar = create(null);
 
-  const meta = create(null);
+  const importMeta = create(null);
   if (moduleRecordMeta) {
-    assign(meta, moduleRecordMeta);
+    assign(importMeta, moduleRecordMeta);
   }
   if (usesImportMeta && importMetaHook) {
-    importMetaHook(meta);
+    importMetaHook(importMeta);
   }
 
   // {_localName_: [{get, set, notify}]} used to merge all the export updaters.
@@ -453,7 +453,7 @@ export const makeModuleInstance = (
             imports: freeze(imports),
             onceVar: freeze(onceVar),
             liveVar: freeze(liveVar),
-            meta,
+            importMeta,
           }),
         );
       } catch (e) {
--- a/packages/ses/src/module-instance.js
+++ b/packages/ses/src/module-instance.js
@@ -18,6 +18,7 @@ import {
   mapGet,
   weakmapGet,
   reflectHas,
+  assign,
 } from './commons.js';
 import { compartmentEvaluate } from './compartment-evaluate.js';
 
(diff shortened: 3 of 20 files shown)


kriskowal (commented):
Let’s talk requirements for this feature at Endo Sync tomorrow. The Compartment abstraction will be aware of meta but not meta.url. For that, we need to give the hooks system a way to populate meta. The current notion is that the import/loadHook will be able to return an object with a meta property and the compartment will copy the properties to the meta object. Because some of those properties are expensive, we would have an importMetaHook that takes a meta object and add things like resolve but would only be called if a module utters import.meta in its source text.

kriskowal:
See https://github.com/endojs/endo/issues/291

packages/compartment-mapper/src/bundle.js
```diff
@@ -332,6 +332,9 @@ ${importsCellSetter(__liveExportMap__, index)}\
     onceVar: {
 ${importsCellSetter(__fixedExportMap__, index)}\
     },
+    metaVar: {
+      url: '${packageLocation}'\
+    },
```
naugtur, line 337: What do we want to do about bundler?

naugtur:
Made some progress on the implementation. Looks like it's it.

packages/compartment-mapper/src/bundle.js
kriskowal: To support `import.meta` in the bundler, we’d inline the logic we’re introducing to `module-instance.js` in the generated code. `Object.assign($h_ort_meta, descriptor.meta)`. We’d need to capture the `meta` property and JSON serialize it into the bundle at that location.

For now, though, the bundler is very incomplete and can remain that way until we have a motivating use case, like replacing Browserify at MetaMask or replacing Rollup in Agoric SDK bootstrap. To get there, it needs a lot of work. For now, it’s sufficient for generating the SES bundles only. For example, it only works within a single package. It doesn’t support CommonJS (though, it probably could pretty easily).

kriskowal (commented):
Yeah, this is on-track. The comments I’ve left pertain to the leading half of this feature, which needs to allow for usage like:

```js
const c = new Compartment({}, {}, {
  importHook(specifier) {
    const requestUrl = locate(specifier);
    const response = await fetch(requestUrl);
    const responseText = await response.text();
    return {
      record: new StaticModuleRecord(responseText),
      meta: {url: responseURL};
    };
  },
  importMetaHook(meta) {
    import.meta.resolve = url => importMetaResolve(url, meta.url);
  },
});
```

packages/ses/src/module-instance.js
```diff
@@ -158,6 +160,11 @@ export const makeModuleInstance = (
   // both initialize and update live bindings.
   const liveVar = create(null);
 
+  const metaVar = create(null);
+  if (usesImportMeta) {
+    assign(metaVar, importMetaHook(moduleSpecifier));
+  }
```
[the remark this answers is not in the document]
naugtur: Ok, that looks very similar to what we saw from Modable too.

packages/compartment-mapper/src/bundle.js
```diff
@@ -332,6 +332,9 @@ ${importsCellSetter(__liveExportMap__, index)}\
     onceVar: {
 ${importsCellSetter(__fixedExportMap__, index)}\
     },
+    metaVar: {
```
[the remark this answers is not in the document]
naugtur: As a newcomer to the codebase I enjoyed the variables all sharing a naming convention. It was much easier to spot them in the more complex bits. Could we incorporate the term "a value for a hidden variable" into the name somehow?

packages/ses/src/module-instance.js
```diff
@@ -438,6 +445,7 @@ export const makeModuleInstance = (
             imports: freeze(imports),
             onceVar: freeze(onceVar),
             liveVar: freeze(liveVar),
+            metaVar: freeze(metaVar),
```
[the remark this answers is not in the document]
naugtur: ok, keeping it room temperature for now.

packages/compartment-mapper/src/bundle.js
naugtur: this is now potentially the only remaining todo unless we want to progress on changing the record envelope format

packages/ses/src/module-load.js
```diff
@@ -317,7 +321,7 @@ export const load = async (
       `Failed to load module ${q(moduleSpecifier)} in package ${q(
         compartmentName,
       )} (${errors.length} underlying failures: ${arrayJoin(
-        arrayMap(errors, error => error.message),
+        arrayMap(errors, error => error.stack),
```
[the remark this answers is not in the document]
naugtur: my bad. I always check if I committed it before the final merge, I change it so often. Thinking of automating it with a precommit script

packages/static-module-record/src/static-module-record.js
```diff
@@ -60,5 +61,8 @@ export function StaticModuleRecord(source, url) {
   this.__syncModuleProgram__ = functorSource;
   this.__liveExportMap__ = liveExportMap;
   this.__fixedExportMap__ = fixedExportMap;
+  this.__usesImportMeta__ = importMetaProperties.length > 0;
+  // useful for knowing if resolve or any other expensive features are used
+  this.__usesImportMetaProperties__ = importMetaProperties;
```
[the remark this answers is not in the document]
naugtur: ok. I might have fixated on the Node.js `resolve` being expensive.

packages/ses/README.md
```diff
@@ -412,13 +412,14 @@ A compiled static module record has the following shape:
...
   The name implies a future record type that supports top-level await.
-  - An initialization record has the properties `imports`, `liveVar`, and
+  - An initialization record has the properties `imports`, `liveVar`, `importMeta` and
     `onceVar`.
     - `imports` is a function that accepts a map from partial import
       module specifiers to maps from names that the corresponding module
       exports to notifier functions.
       A notifier function accepts an update function and registers
       to receive updates for the value exported by the other module.
+    - `importMeta` is a null-prototype object with keys transferred from `meta` property in the envelope returned by importHook and/or mutated by calling `importMetaHook(importMeta)`
```
naugtur, line 422: I suppose we're going to need to write more about importMetaHook

packages/static-module-record/src/transform-analyze.js
```diff
@@ -65,6 +64,8 @@ const makeCreateStaticRecord = transformSource =>
       hoistedDecls: [],
       importSources: Object.create(null),
       importDecls: [],
+      // enables passing import.meta usage hints up.
+      importMeta: { uttered: false },
```
naugtur, line 68: I don't like the looks of it, but how else do I pass a boolean by reference? 
At least it's future proof if we want more information out of it.

packages/compartment-mapper/src/bundle.js
naugtur: So should we agree on how to proceed for bundle.js or should I remove this change and add a new issue?

packages/static-module-record/test/test-static-module-record.js
```diff
@@ -556,6 +557,25 @@ test('import for side-effect', t => {
   t.deepEqual(record.__liveExportMap__, {});
   t.deepEqual(record.imports, ['module']);
 });
+test('import meta', t => {
+  t.notThrows(() => initialize(t, `const a = import.meta.url`));
+});
+test.failing('import meta in export', t => {
```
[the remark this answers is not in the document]
naugtur: Update.

`rewriteDeclaration` is somehow preventing visitor from reaching its results. 
AFAIK visitor should go into replacements and run on nodes matched there. I don't know why this is not happening here. 

Adding `path.traverse(...` inside `rewriteDeclaration` can fix the issue, but it halves the results of a quick benchmark I've set up.

One way to keep the performance unchanged is to rewrite on the first pass not the second. It works, but it's probably controversial.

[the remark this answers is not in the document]
naugtur: I've got n update where `rewriteDeclarations` no longer escapes the visitor. I managed to narrow it down to something that replacing declarations with themselves was doing. It seems ok to append instead.

I'll make a separate PR with it.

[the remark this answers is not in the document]
naugtur: separate PR merged

packages/compartment-mapper/src/bundle.js
naugtur: Agreed to not do anything and leave import.meta as an empty object.

naugtur:
@kriskowal Unless I need to put more contrnt into `packages/ses/README.md` I think this is ready for final review.

packages/ses/README.md
kriskowal: We probably also ought to make the signature `importMetaHook(fullSpecifier, importMeta)` in case that’s useful.

packages/ses/src/module-instance.js
```diff
@@ -158,6 +165,14 @@ export const makeModuleInstance = (
   // both initialize and update live bindings.
   const liveVar = create(null);
 
+  const importMeta = create(null);
+  if (moduleRecordMeta) {
+    assign(importMeta, moduleRecordMeta);
+  }
+  if (usesImportMeta && importMetaHook) {
+    importMetaHook(importMeta);
```
kriskowal, line 173: We probably won’t regret making this `importMetaHook(moduleSpecifier, importMeta)`

packages/static-module-record/src/babelPlugin.js
```diff
@@ -309,6 +310,22 @@ function makeModulePlugins(options) {
     };
 
     const moduleVisitor = (doAnalyze, doTransform) => ({
+      MetaProperty(path) {
+        if (
+          path.node.meta &&
+          path.node.meta.name === 'import' &&
+          path.node.property.name === 'meta'
+        ) {
```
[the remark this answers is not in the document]
kriskowal: I like the specificity.

packages/static-module-record/src/hidden.js
```diff
@@ -5,12 +5,14 @@ export const HIDDEN_IMPORT = `${HIDDEN_PREFIX}import`;
 export const HIDDEN_IMPORT_SELF = `${HIDDEN_PREFIX}importSelf`;
 export const HIDDEN_IMPORTS = `${HIDDEN_PREFIX}imports`;
 export const HIDDEN_ONCE = `${HIDDEN_PREFIX}once`;
+export const HIDDEN_META = `${HIDDEN_PREFIX}import_meta`;
```
kriskowal, line 8: One of us requested that we somehow make the apparent width of the replacement string the same as the original, so it doesn’t displace the column number of following text. I’m not sure how ZWJ generally appears in, say, VSCode, but maybe this should be something like`$h_____meta`.

packages/static-module-record/src/static-module-record.js
```diff
@@ -60,5 +61,6 @@ export function StaticModuleRecord(source, url) {
   this.__syncModuleProgram__ = functorSource;
   this.__liveExportMap__ = liveExportMap;
   this.__fixedExportMap__ = fixedExportMap;
+  this.__usesImportMeta__ = importMetaUttered;
```
kriskowal, line 64: The tentative proposed verbiage is “needs”. This is “pretend not to see” public. Which is to say, we don’t get to change it after it releases. So perhaps we should wait for a blessing on https://github.com/tc39/proposal-compartments/pull/46/ before we land this.

packages/static-module-record/src/transform-analyze.js
kriskowal: Passing an `importMetaSeen()` function instead might be clearer.

packages/static-module-record/src/transform-analyze.js
```diff
@@ -122,25 +85,25 @@ const makeCreateStaticRecord = transformSource =>
     if (url) {
       functorSource += `//# sourceURL=${url}\n`;
     }
-
     const moduleAnalysis = freeze({
       exportAlls: freeze(sourceOptions.exportAlls),
       imports: freeze(sourceOptions.imports),
       liveExportMap: freeze(sourceOptions.liveExportMap),
       fixedExportMap: freeze(sourceOptions.fixedExportMap),
+      importMetaUttered: sourceOptions.importMeta.uttered,
```
kriskowal, line 93: I believe we agree that we should pretend “utter” isn’t a word.

kriskowal (commented):
This is ready to land, though the specification it’s based on needs a little more time to bake. Until then, I’ve left some notes for your consideration.

kriskowal (approved):
…and I’m leaving a stamp so you don’t have to wait for my blessing when everything’s ready.

packages/static-module-record/src/hidden.js
naugtur: ZWJ can display as nothing or as a blocky box and a warning that  something's fishy about the file. I guess it's the number of characters that counts, so ZWJ counts as 1.

I'll adjust the length and add a comment explaining what happened.

packages/static-module-record/src/transform-analyze.js
naugtur: That was 2 weeks after I put it here based on the sophisticated-vocabulary-filled meeting ;)
Then again - this is an internal name visible to people who need to grow their vocabulary significantly to grok this codebase anyway.  
I think the best result would be to make it consistent with the public name we eventually land.
https://github.com/tc39/proposal-compartments/pull/46

packages/ses/README.md
naugtur: done

packages/static-module-record/src/hidden.js
naugtur: length including ZWJj should now match

PR #1144, opened 2022-04-06
accept require.resolve implementation as option to importLocation et al.

naugtur:
Receiving the implementation externally makes more sense than supporting require.resolve functionality endoEndo.

This PR accepts the implementation and, if not provided, defaults to a similar error Node would throw.


packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
```diff
@@ -131,7 +140,20 @@ export const wrap = (moduleEnvironmentRecord, compartment, resolvedImports) => {
...
+  };
+  if (requireResolve) {
+    require.resolve = freeze((specifier, options) =>
+      requireResolve(location, specifier, options),
+    );
+  } else {
+    require.resolve = freeze(specifier => {
+      const error = Error(`Cannot find module '${specifier}'`);
+      defineProperty(error, 'code', { value: 'MODULE_NOT_FOUND' });
+      throw error;
```
naugtur, line 152: Question: Should this be indistinguishable from an error Node would throw, or rather match it loosely but leave a hint somewhere to it resulting from a lack of requireResolve option being passed to Endo?

packages/compartment-mapper/test/scaffold.js
```diff
@@ -87,6 +88,8 @@ export function scaffold(
 
     const application = await loadLocation(readPowers, fixture, {
       dev: true,
+      requireResolve: (from, specifier, options) =>
+        createRequire(from).resolve(specifier, options),
```
naugtur, line 92: I like how easy it is to provide the implementation.
A more secure implementation would use `options.paths` to limit the search space and guard return values by checking with `String.startsWith`that it's not reaching beyond the project folder.

A cache lookup of known needed paths for specifiers is also an option.

packages/compartment-mapper/test/test-cjs-compat.js
```diff
@@ -26,10 +26,35 @@ const assertFixture = (t, { namespace }) => {
...
+      '/skipped/fixtures-cjs-compat/node_modules/require-resolve/nested/index.js',
+      '/skipped/fixtures-cjs-compat/node_modules/require-resolve/nested/file.js',
+      '/skipped/fixtures-cjs-compat/node_modules/require-resolve/nested/file.js.map',
+      "Cannot find module './nested/file.missing'",
+      '/skipped/fixtures-cjs-compat/node_modules/app/index.js',
+      'fs',
+      '/skipped/fixtures-cjs-compat/node_modules/nested-export/callBound.js',
+    ]);
+  } else {
+    t.deepEqual(results.requireResolvePaths, [
```
naugtur, line 42: This is what it looks like when requireResolve implementation is not provided.

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
naugtur: from a meeting: add the second line to the error mesage explaining requireResolve needs to be passed

kriskowal (commented):
I would like to fold requireResolve into the optional properties of readPowers. Otherwise, this looks good to me.

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
```diff
@@ -131,7 +140,22 @@ export const wrap = (moduleEnvironmentRecord, compartment, resolvedImports) => {
     } else {
       return namespace;
     }
-  });
+  };
+  if (readPowers && readPowers.requireResolve) {
+    require.resolve = freeze((specifier, options) =>
+      readPowers.requireResolve(location, specifier, options),
```
naugtur, line 146: Now all that's left to do is convince Typescript I can call this function if I check the field exists even thought it's optional on readPowers

mhofman:
I'm not sure I have much context to review this PR. Is there anything specific you'd like me to take a look at?

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
naugtur: it stopped complaining when I did this
```js
 if (typeof readPowers === 'object' && readPowers.requireResolve) {
    const { requireResolve } = readPowers;
    require.resolve = freeze((specifier, options) =>
      requireResolve(location, specifier, options),
    );
```
I don't appreciate what TS is trying to do for me here

naugtur:
@kriskowal This is ready IMHO, I need *the* final review.

mhofman:
> @mhofman Automatons said you had overlap with the codebase that this touches

Automatons are mistaken, I'll remove myself as reviewer.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -176,7 +176,9 @@ export const makeImportHookMaker = (
             candidateSpecifier,
             moduleLocation,
             packageLocation,
-            readPowers,
+            {
+              readPowers,
+            },
```
kriskowal, line 181: Is this envelope still necessary?

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
```diff
@@ -131,7 +140,23 @@ export const wrap = (moduleEnvironmentRecord, compartment, resolvedImports) => {
...
+  };
+  if (typeof readPowers === 'object' && readPowers.requireResolve) {
+    const { requireResolve } = readPowers;
+    require.resolve = freeze((specifier, options) =>
+      requireResolve(location, specifier, options),
+    );
+  } else {
+    require.resolve = freeze(specifier => {
+      const error = Error(
+        `Cannot find module '${specifier}'\nPass ReadPowers with a requireResolve function to provide require.resolve`,
```
kriskowal, line 152: Maybe `Add requireResolve to Endo Compartment Mapper readPowers.`

kriskowal (approved):

PR #1145, opened 2022-04-06
fix(compartment-mapper): there's more benefit to keeping __esModule flag than not 

naugtur:
compiled TS depends on it across packages. 

```
var __importDefault = (this && this.__importDefault) || function (mod) {
    return (mod && mod.__esModule) ? mod : { "default": mod };
};
```
https://www.typescriptlang.org/tsconfig#esModuleInterop


kriskowal (approved):

kriskowal:
Can you frame this requirement as a test?

naugtur:
> Can you frame this requirement as a test?

I did, but in e2e tests. Will do.

PR #1147, opened 2022-04-07
fix(compartment-mapper): infer types of nested exports

naugtur:
This PR implements looking for package.json files in the nested structure of the package so that they can change the "type" to "module" for a subfolder.

Too much back and forth here, I'll need to squash this.

----

Old text for reference:

I researched it more and this is a good heuristic but it's not what others are doing.

the field being named 'import' doesn't actually mean it should be interpreted as esm. It just so happens that all packages I've encountered to date point those fields at files with mjs extensions or locaded in folders with a type:module package.json file.

The implementation that'd be fully valid would need to collect all inferred types, check their folder trees for package.json files and decide the formats based on that.  
example:
https://github.com/jashkenas/underscore/issues/2954


packages/compartment-mapper/src/node-modules.js
```diff
@@ -269,6 +269,20 @@ const graphPackage = async (
...
+    do {
+      // eslint-disable-next-line no-await-in-loop
+      desc = await readDescriptor(
+        resolveLocation(pathSegments.join('/'), packageLocation),
+      );
+      pathSegments.pop();
+    } while (!desc && pathSegments.length !== 0);
+
+    return desc;
+  };
```
kriskowal, line 284: Please look at `search.js` and either borrow the technique or factor the code so you can use the same algorithm. This should be implemented in terms of proper URL math for platform independence.

There should be at least one test that would fail on Windows with this algorithm, by using back-slashes.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -279,6 +293,15 @@ const graphPackage = async (
     parsers: inferParsers(packageDescriptor, packageLocation),
   });
 
+  await Promise.all(
+    values(result.exports).map(async item => {
+      const descriptor = await readDescriptorUpwards(item);
```
kriskowal, line 298: I presume the performance of this is okay because `readDescriptor` is memoized?

kriskowal (changes requested):

packages/compartment-mapper/src/node-modules.js
naugtur: Yes. I'd gladly avoid the whole thing altogether, but that's the best we can do.
As for optimizing it, I've been thinking of priming the fs.stat with a lookup table of paths that exist to avoid expensive errors on reading paths that don't exist. But that'd essentially be a node contribution or a package wrapping fs

packages/compartment-mapper/src/node-modules.js
naugtur: done. refactored search.js a bit to expose the logic of searching but allow using the local readDescriptor function

packages/compartment-mapper/test/snapshots/test-infer-exports.js.md
```diff
@@ -0,0 +1,140 @@
...
+
+## infer-exports for @noble/hashes
+
+> Snapshot 1
+
+    {
+      exports: {
+        '@noble/hashes/_sha2': './_sha2.js',
+        '@noble/hashes/blake2b': './blake2b.js',
+        '@noble/hashes/blake2b.d.ts': 'blake2b.d.ts',
```
naugtur, line 29: Before I merged origin/master these were `./blake2b.d.ts` and now they're not relativized.
When I make a change to relativize them, I get a test failure elsewhere

```
retained › archives only contain compartments retained by modules

  Returned promise rejected with:

  Error {
    message: 'Cannot find file for internal module "./mk2" (with candidates "./mk2", "./mk2.js", "./mk2/index.js") in package file:///home/naugtur/work/metamask/endo/packages/compartment-mapper/test/fixtures-retained/node_modules/mk1/',
  }

  › file://test/test-retained.js:16:11

```

@kriskowal which is correct? I thought they should all be relativized

kriskowal: I recently made a change that makes it possible for a package to reëxport an external dependency, e.g., `"exports": {".": "fs"}`. It doesn’t seem likely to me that the `exports` feature intentionally limits that possibility by relativizing everything, but maybe it does. I think I will have to depend on your research to tell.

naugtur: These are not exports in package.json, these are results of running inferExports and are a flat key-value with values to use by compartment-mapper. 
I was hoping to get a "no, these won't work if not relativized" or "this is fine" ;)

kriskowal: No misunderstanding. These will not work if not relativized. That means you would have to revert https://github.com/endojs/endo/commit/5a8a893b9f9d4375661fae597c158a0bbf258785 to get this to work. That makes me sad, but is consistent with the usage in `@noble/hashes` and therefore indicates a mistake in the design of `"exports"` that we can’t fix.

naugtur: @kriskowal done. I even persisted some of the sadness from yesterday's meeting in a comment.
Let's merge this ;)

kriskowal (approved):
Consider adding a bullet to `NEWS.md`

naugtur:
TIL interactive rebase of commits surrounding a merge commit doesn't work. Had to take clean master and cherry-pick squashed work on top.

Also, @kriskowal tests failed on Windows after I pushed everything despite no changes to the final state of code. Retrying made it go away. Is this a known issue? Flakiness of some sort?

kriskowal:
Can you spot the failed CI among these? https://github.com/endojs/endo/actions

naugtur:
It's close to impossible to find ;) 
https://github.com/endojs/endo/actions/runs/2330647919/attempts/1

kriskowal:
Thanks. I’ve passed the report on to @michaelfig. Fixing flappy tests is the gift that keeps on giving.

kriskowal:
Will keep an eye for it. Thanks for looking!

Issue #927, opened 2021-11-03
Account for ESM/CJS compat mode __esModule


kriskowal:
The SES and the Endo Compartment Mapper have a frail and under-tested implementation of CommonJS as a third-party static module record. These need to be more extensively tested and the difference between a CommonJS module that sets `exports.__esModule` distinguished from one that does now. The setter can be detected after initialization and the behavior of the `default` getter on the proxied exports behavior varied accordingly.

Background on compatibility mode https://github.com/esnext/es6-module-transpiler/issues/85

naugtur:
A table from esbuild folks for reference https://github.com/evanw/esbuild/issues/532#issuecomment-727647005

naugtur:
Looking at the results for case 11 in the table [here](https://github.com/endojs/endo-e2e-tests/blob/naugtur-matrix-iteration/matrix/table.md)  unless we want to be more like tsc and some transpilers, no work needs to be done other than merging 
https://github.com/endojs/endo/pull/1145

kriskowal:
I think we’re close enough. If we can get parity with Parcel, Rollup &c, it would be nice, but not at the expense of Node.js parity or our sanity.

naugtur:
The overlap between parcel and sanity is not something I'm comfortable with https://github.com/endojs/endo-e2e-tests/blob/naugtur-matrix-iteration/matrix/table.md
I may need to configure it further, but default out of the box behavior is surprising.

Anyway - we've got node compatibility and a bit more.
All the functionality of __esModule I'm aware of is happening in node and endo even when __esModule is not there. It needs to be there for the bundled packages to work together correctly, that's all. (I thought I'm fine removing it entirely, but no)

naugtur:
Ok, so there's a PR now.  #1149 

I tested it and it seems just adding this in doesn't break nor fix anything in e2e on real packages.

It does create a behavior difference between endo and node getting endo closer to what bundlers do. 
The glitches this is potentially introducing are harder to figure out then not having it in, so I'm not sure if I want it. 

I could add in some failing tests to ilustrate those

kriskowal:
My preference is not to modify third-party-module support in such a way that WASM or any other language would interact with `__esModule` at all. I also doubt that TC39 would accept back-wash from the `__esModule` hack.

Failing tests would be illuminating!

[...]

PR #1149, opened 2022-04-11
feat(compartment-mapper): full support of __esModule

naugtur:
Fixes: #927


kriskowal:
1. This looks correct to me. 
2. This is horrible.

I’m content to follow your lead on this behavior. If that means holding this back until we have a motivating case, I’m fine with that.

naugtur:
I woke up with a case that's likely to break it. None of the transpilers would do it, but this won't work:
```js
module.exports=()=>{•••}
module.exports.__esModule=true
module.exports.default=something
``` 

The function becomes unreachable if required because we're gonna return the module record namespace because default is taken over by esmodule.

Transpiled code will not create this type of abomination though. (they're into other stuff 😉)


This is in theory a breaking change, so if we ever want to introduce it, we better introduce it early. But! I want to at least find proof it's needed for anything.

naugtur:
This solution doesn't stand up to the test of real ecosystem examples. Adding it has not improved test results on real packages and has broken a few.

PR #1150, opened 2022-04-11
fix(ses): Prevent hypothetical stack bumping for unsafe eval

### LINKED ISSUE #840: Rename localObject to globalLexicals in performEval
At Sestival 2021, in review we decided that globalLexicals would be more understandable in performEval. This will require some munging in the compartment evaluate function, which weaves globalLexicals from various options named globalLexicals and similar.

### LINKED ISSUE #956: OOM must fail fast (through unsafe eval switch)
If an attacker can arrange for a stack of a particular depth before calling an evaluator, on a platform with some specific combination of preferential inlining of functions, there is a chance, in the face of maintenance, that a program could cause an evaluator to leak unsafeEval. The implementation as written today, makes a single function call in a finally block to retract the unsafeEval that is temporarily granted to evaluate client code. To eliminate this risk, that finally block should make no function calls and merely reset the underlying flag.

https://github.com/endojs/endo/blob/9a9953bbf80fd4c35d857c999cfe2972f6891912/packages/ses/src/make-safe-evaluator.js#L91

To achieve this, we must integrate the compartment scope handler maker and the evaluator into a shared scope.

And we should leave a comment to the effect that the code in the finally block must never call functions in the future.

kriskowal:
- refactor(ses): Make evaluate factory validation less sensitive to white space
- refactor(ses): Rename localObject to globalLexicals
- fix(ses): Prevent hypothetical stack bumping for unsafe eval

Fixes #956
Fixes #840 (drive by fix)

### DIFF
--- a/packages/ses/src/scope-handler.js
+++ b/packages/ses/src/scope-handler.js
@@ -53,7 +53,7 @@ const alwaysThrowHandler = new Proxy(
  */
 export const createScopeHandler = (
   globalObject,
-  localObject = {},
+  globalLexicals = {},
   { sloppyGlobalsMode = false } = {},
 ) => {
   // This flag allow us to determine if the eval() call is an done by the
@@ -90,12 +90,12 @@ export const createScopeHandler = (
         // fall through
       }
 
-      // Properties of the localObject.
-      if (prop in localObject) {
+      // Properties of the globalLexicals.
+      if (prop in globalLexicals) {
         // Use reflect to defeat accessors that could be present on the
-        // localObject object itself as `this`.
+        // globalLexicals object itself as `this`.
         // This is done out of an overabundance of caution, as the SES shim
-        // only use the localObject carry globalLexicals and live binding
+        // only use the globalLexicals carry globalLexicals and live binding
         // traps.
         // The globalLexicals are captured as a snapshot of what's passed to
         // the Compartment constructor, wherein all accessors and setters are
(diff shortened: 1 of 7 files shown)


packages/ses/test/test-make-evaluate-factory.js
```diff
@@ -5,12 +5,18 @@ test('Intrinsics - values', t => {
   t.plan(2);
 
   t.is(
-    makeEvaluateFactory().toString(),
-    "function anonymous(\n) {\n\n    with (this) {\n      \n      return function() {\n        'use strict';\n        return eval(arguments[0]);\n      };\n    }\n  \n}",
+    makeEvaluateFactory()
+      .toString()
+      .replace(/\s/g, ' ')
+      .replace(/ +/g, ' '),
```
mhofman, line 11: Curious `replace(/\s+/g, ' ')` didn't work?

packages/ses/src/scope-handler.js
```diff
@@ -59,17 +60,15 @@ export const createScopeHandler = (
...
-    return wasSet;
+  // Using a flag on an object with a single mutable property allows a safe
+  // evaluator to signal to the scope proxy without consuming a stack frame.
+  // Consuming a stack frame could possibly allow an attacker to control the
+  // stack depth before calling `evaluate` to cause a RangeError before this
+  // flag can be reset, leaving the unsafe evaluator available.
+  const scopeController = {
+    allowNextEvalToBeUnsafe: false,
   };
+  preventExtensions(scopeController);
```
mhofman, line 71: Technically `seal` would work too, right?

mhofman (approved):

PR #1154, opened 2022-04-13
build(bundle-source): pin to a fixed rollup tarball

michaelfig:
Fix the following override mistake in Rollup used by `endo/packages/bundle-source` that rears its head when some apparently unrelated package uses `hasOwnProperty`:

```
2022-04-13T00:38:54.385Z ag-solo: solo [ 'start' ]
(TypeError#1)
TypeError#1: Override property hasOwnProperty
  at Object.setter (file:///private/var/folders/67/vylk7hzs6v72jwj0gjbq205w0000gp/T/agoric-cli-test--42195-9cOV8ApJw4EU/dapp-foo/node_modules/ses/dist/ses.umd.js:5943:27)
  at MagicString.overwrite (file:///private/var/folders/67/vylk7hzs6v72jwj0gjbq205w0000gp/T/agoric-cli-test--42195-9cOV8ApJw4EU/dapp-foo/node_modules/rollup/dist/es/shared/rollup.js:812:30)
  at Identifier.render (file:///private/var/folders/67/vylk7hzs6v72jwj0gjbq205w0000gp/T/agoric-cli-test--42195-9cOV8ApJw4EU/dapp-foo/node_modules/rollup/dist/es/shared/rollup.js:6469:22)
  at Property.render (file:///private/var/folders/67/vylk7hzs6v72jwj0gjbq205w0000gp/T/agoric-cli-test--42195-9cOV8ApJw4EU/dapp-foo/node_modules/rollup/dist/es/shared/rollup.js:11028:20)
[...]
```

Our friend @Jack-Works has already [landed a patch](https://github.com/Rich-Harris/magic-string/pull/194) upstream for this problem, but it has not yet made it into a published rollup version.


kriskowal (approved):
Please leave a note somewhere to look for the upstream fix in Rollup, so we know to remove the hack when the permanent fix is available.

PR #1155, opened 2022-04-13
fix(compartment-mapper): remove the file: protocol from __filename an…

naugtur:
…d __dirname for cjs compatibility

This makes browserify work.

### DIFF
--- a/packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
+++ b/packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
@@ -8,12 +8,46 @@ const { freeze, keys, create, hasOwnProperty, defineProperty } = Object;
  * @returns {boolean}
  */
 const has = (object, key) => apply(hasOwnProperty, object, [key]);
-
-export const dropFileProtocol = url => {
+const removeProtocol = url => {
   if (url.substring(0, 7) === 'file://') {
     url = url.substring(7);
   }
   return url;
+};
+const noTrailingSlash = path => path.replace(/[\\/]$/, '');
+
+export const getModulePaths = async (readPowers, url) => {
+  if (!(readPowers && readPowers.fileURLToPath)) {
+    return {
+      filename: null,
+      dirname: null,
+    };
+  }
+
+  let urlFilename = url;
+  let urlDirname;
+  try {
+    urlDirname = new URL('./', url).href;
+  } catch (_) {
+    return {
+      filename: null,
+      dirname: null,
+    };
+  }
+
+  if (readPowers && readPowers.fileURLToPath) {
+    urlFilename = readPowers.fileURLToPath(urlFilename).toString();
+    urlDirname = readPowers.fileURLToPath(urlDirname).toString();
+  } else {
+    urlFilename = removeProtocol(urlFilename);
+    urlDirname = removeProtocol(urlDirname);
+  }
+  urlDirname = noTrailingSlash(urlDirname);
+
+  return {
+    filename: urlFilename,
+    dirname: urlDirname,
+  };
 };
 
 /**
--- a/packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
+++ b/packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
@@ -1,4 +1,8 @@
 // @ts-check
+
+/** @typedef {import('./types.js').ReadFn} ReadFn */
+/** @typedef {import('./types.js').ReadPowers} ReadPowers */
+
 const { apply } = Reflect;
 const { freeze, keys, create, hasOwnProperty, defineProperty } = Object;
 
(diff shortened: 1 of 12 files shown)


packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
```diff
@@ -9,6 +9,13 @@ const { freeze, keys, create, hasOwnProperty } = Object;
  */
 const has = (object, key) => apply(hasOwnProperty, object, [key]);
 
+export const dropFileProtocol = url => {
+  if (url.substring(0, 7) === 'file://') {
+    url = url.substring(7);
+  }
+  return url;
+};
```
kriskowal, line 17: This isn’t quite sufficient, accounting for Windows. The `readPowers` threaded into all of the Compartment Mapper API’s capture `url` (`url.filePathToURL`, `url.pathToFileURL`) so it can convert between URL space and path space.

naugtur: When importLocation is used, read powers don't enter compartment-mapper and even if I threaded them down to the parser implementation I couldn't leverage them. I can use the `new URL … url.pathname` implementation locally but don't see means of drilling the original reference to it down to the parser reliably.

kriskowal: I think a reasonable fallback position is to provide `__dirname === undefined` if the compartment mapper doesn’t receive the `canonicalize` property on `ReadPowers`. You may have to change the signature to `ReadFn | ReadPowers` to thread it, but there are examples where I’ve already done this. There’s a utility function for normalizing that type laying around somewhere.

naugtur: readpowers is `{ read, canonical }`.   
`canonical` is not helpful for providing a file path on windows. I will expose existing path2url and url2path functions.

packages/compartment-mapper/src/types.js
```diff
@@ -190,6 +190,7 @@ export const moduleJSDocTypes = true;
  * @param {string} specifier
  * @param {string} location
  * @param {string} packageLocation
+ * @param {ReadFn | ReadPowers} [readPowers]
```
kriskowal, line 193: I believe you’re proposing that we add `fileURLToPath` to `ReadPowers`.

  I’m okay with `importLocation` using `fileURLToPath` but not okay with `makeArchive` or `parseArchive` using them because the behavior is platform-dependent, these functions must be consistent regardless of platform, and the values would not be useful for archives. For backward-compatibility, the new function would need to be optional and have a fallback. Since this is only used for `__filename` and `__dirname` in CommonJS, I’m okay with the fallback being to provide `null` and `null`. Providing a fallback that gives right answers only on some platforms would be sketchy.

packages/compartment-mapper/src/parse-cjs.js
```diff
@@ -25,6 +26,8 @@ export const parseCjs = async (
     exports.push('default');
   }
 
+  const lowlow = await getPaths(readPowers, location); // __filename
```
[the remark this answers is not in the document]
naugtur: It meant `__` :D 
I didn't have a name for this reference at the time.

packages/compartment-mapper/src/parse-pre-cjs.js
```diff
@@ -44,8 +40,8 @@ export const parsePreCjs = async (
       require,
       moduleExports,
       module,
-      location, // __filename
-      locationParent(location), // __dirname
+      lowlow.filename, // __filename
+      lowlow.dirname, // __dirname
```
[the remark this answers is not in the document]
naugtur: those comments were there before me ;)

packages/compartment-mapper/test/test-cjs-compat.js
```diff
@@ -32,3 +37,16 @@ scaffold(
...
   fixtureAssertionCount,
 );
+
+scaffold(
+  'fixtures-cjs-compat-__dirname',
+  test,
+  fixtureDirname,
+  (t, { namespace }) => {
+    const { __dirname, __filename } = namespace;
+    t.is(__filename, `${__dirname}/dirname.js`);
```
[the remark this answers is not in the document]
naugtur: because my implementation was also incorrect :D

naugtur:
@kriskowal this should now be ready. One controversy I can see is removing the trailing slash with regex - couldn't find a function that'd do that for me.

packages/compartment-mapper/src/parse-cjs.js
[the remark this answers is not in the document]
kriskowal: My favorite head canon for underscore, and low bar variants is:

- `_` underbar
- `__` dunderbar via the Python community
- `___` wunderbar 😉 via the Caja community h/t @erights

packages/compartment-mapper/test/test-cjs-compat.js
```diff
@@ -8,6 +9,10 @@ const fixture = new URL(
   'fixtures-cjs-compat/node_modules/app/index.js',
   import.meta.url,
 ).toString();
+const fixtureDirname = new URL(
+  'fixtures-cjs-compat/node_modules/app/dirname.js',
+  import.meta.url,
+).toString();
```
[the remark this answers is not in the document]
naugtur: been using it ever since

kriskowal:
> @kriskowal this should now be ready. One controversy I can see is removing the trailing slash with regex - couldn't find a function that'd do that for me.

This calls for an obscure reference. Barclay says:

image

```js
const directoryToFileURL = url => {
  if (url.lastIndexOf('/', url.length - 1) === -1) return url;
  return url.slice(0, -1);
};
```

naugtur:
@kriskowal 
> ```js
> const truncateIfFinalSlash = path => {
>   if (path.lastIndexOf('/', url.length - 1) === -1) return path;
>   return path.slice(0, -1);
> };
> ```

I prefer substring with negative values, which I could use too, but the thing is, you insisted on supporting both `/` and `\` which seemed most readable with a tiny regex.
Want me to switch to string manipulation?

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
```diff
@@ -8,6 +8,47 @@ const { freeze, keys, create, hasOwnProperty } = Object;
  * @returns {boolean}
  */
 const has = (object, key) => apply(hasOwnProperty, object, [key]);
+const removeProtocol = url => {
+  if (url.substring(0, 7) === 'file://') {
```
kriskowal, line 12: `indexOf('file://') !== -1` has the virtue of not allocating, but I don’t feel strongly.

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
```diff
@@ -8,6 +8,47 @@ const { freeze, keys, create, hasOwnProperty } = Object;
...
+    url = url.substring(7);
+  }
+  return url;
+};
+const noTrailingSlash = path => path.replace(/[\\/]$/, '');
```
kriskowal, line 17: I don’t feel strongly about avoiding a regex here, but I am not strong enough with regexen to tell you whether they’re ReDoS risks. I assume this compiles to a regex that doesn’t scan the whole length of the string.

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
```diff
@@ -8,6 +8,47 @@ const { freeze, keys, create, hasOwnProperty } = Object;
...
+
+export const getModulePaths = async (readPowers, url) => {
```
kriskowal, line 19: Do you not get TypeScript warning about these `any` typed args?

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
```diff
@@ -8,6 +8,47 @@ const { freeze, keys, create, hasOwnProperty } = Object;
...
+    urlDirname = readPowers.fileURLToPath(urlDirname).toString();
+  } else {
+    urlFilename = removeProtocol(urlFilename);
+    urlDirname = removeProtocol(urlDirname);
+  }
+  urlDirname = noTrailingSlash(urlDirname);
+
+  return {
+    filename: urlFilename,
+    dirname: urlDirname,
```
kriskowal, line 49: Slightly more concise with just `filename` instead of `urlFilename`. I do not feel strongly.

packages/compartment-mapper/src/parse-pre-cjs.js
```diff
@@ -1,31 +1,26 @@
...
   _packageLocation,
+  readPowers,
 ) => {
   const text = textDecoder.decode(bytes);
   const { source, imports, exports, reexports } = parseLocatedJson(
     text,
     location,
   );
 
+  const modulePaths = await getModulePaths(readPowers, location);
```
kriskowal, line 22: Nit, can destructure here.

kriskowal (approved):

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
naugtur: appreciated

PR #1158, opened 2022-04-14
fix(bundle-source): Point rollup patch at github

kriskowal:
- fix(bundle-source): Use a github reference instead of a bundled tarball for rollup patch
- chore: Update yarn.lock

The most recent release was not installable since yarn can't handle the tarball within a tarball approach.
Posting the patched version as a Github tag on this repository seems to work.


michaelfig (approved):
LGTM, thanks!

PR #1159, opened 2022-04-14
chore: release

kriskowal:
This release fixes a problem with the previous release of bundle-source, pertaining to a patched version of Rollup.


michaelfig (approved):
LGTM!

As an aside, what is causing all the unnecessary "Version bump"s?  It seems that many of these packages were nowhere near what should have been the blast radius.

kriskowal:
> LGTM!
> 
> As an aside, what is causing all the unnecessary "Version bump"s? It seems that many of these packages were nowhere near what should have been the blast radius.

Just looking at this diff, they might all be propagating from eslint-plugin. Why eslint-plugin gets bumped, I can’t fathom.

PR #1164, opened 2022-04-15
fix(eventual-send): fix unhandled rejection caused by rejection handling

mhofman:
If a result is rejected with a rejected promise as reason, `handle` would harden the promise, and return it, causing it to be adopted, and actually becoming the unhandled rejection we were trying to avoid in the first place.

### DIFF
--- a/packages/eventual-send/src/handled-promise.js
+++ b/packages/eventual-send/src/handled-promise.js
@@ -45,6 +45,11 @@ const coerceToObjectProperty = specimen => {
   return String(specimen);
 };
 
+const hardenVoid = obj => {
+  // Do not return the hardened value
+  harden(obj);
+};
+
 // the following method (makeHandledPromise) is part
 // of the shim, and will not be exported by the module once the feature
 // becomes a part of standard javascript
@@ -548,8 +553,9 @@ export const makeHandledPromise = () => {
 
     // Harden the fulfillment and rejection, as well as a workaround for
     // Node.js: silence "Unhandled Rejection" by default when using the static
-    // methods.
-    returnedP.then(harden, harden);
+    // methods. Use a void returning version of harden to prevent transforming
+    // a rejected promise used as reason into a new unhandled rejection.
+    returnedP.then(hardenVoid, hardenVoid);
 
     // We return a handled promise with the default pending handler.  This
     // prevents a race between the above Promise.resolves and pipelining.


packages/eventual-send/src/handled-promise.js
```diff
@@ -45,6 +45,11 @@ const coerceToObjectProperty = specimen => {
   return String(specimen);
 };
 
+const hardenVoid = obj => {
+  // Do not return the hardened value
+  harden(obj);
+};
```
kriskowal, line 51: It would amuse me to use `void` for its purpose just once.

```suggestion
const hardenVoid = obj => void(harden(obj));
```

kriskowal (approved):
Fantastic catch. It would be phenomenal to capture this in a test. Maybe we can trust `ava` to trap unhandled rejection?

packages/eventual-send/src/handled-promise.js
mhofman: I considered it, but decided it was too obscure for reader.

mhofman:
> It would be phenomenal to capture this in a test.

I would love to, but am a little preoccupied with my own tests right now. Will try to get back to this later. External contributions welcome :)

> Maybe we can trust `ava` to trap unhandled rejection?

It very much does, and fails the test file. That's how I got tripped in the first place.

packages/eventual-send/src/handled-promise.js
```diff
@@ -548,8 +553,9 @@ export const makeHandledPromise = () => {
 
     // Harden the fulfillment and rejection, as well as a workaround for
     // Node.js: silence "Unhandled Rejection" by default when using the static
-    // methods.
-    returnedP.then(harden, harden);
+    // methods. Use a void returning version of harden to prevent transforming
+    // a rejected promise used as reason into a new unhandled rejection.
+    returnedP.then(hardenVoid, hardenVoid);
```
mhofman, line 558: @michaelfig if the intent is to NOT silence unhandled rejections, then the body of the HandledPromise constructor above should call harden before passing to `resolve` or `reject`.

michaelfig: I see.  Maybe a separation of orthogonal concerns would be clearest?  `harden(reason)` and `harden(fulfillment)`, as well as just a `returnedP.catch(() => {})`?

WDYT?

michaelfig: Oh wait, I want to harden the settled value.  So yes, your code is probably best.

michaelfig: > if the intent is to NOT silence unhandled rejections, then the body of the HandledPromise constructor above should call `harden` before passing the value to `resolve` or `reject`.

I have a counterproposal along those lines in #1178.  Would appreciate your eyes on it.

PR #1167, opened 2022-04-20
test(check-bundle): Cover invalid archive error with location diagnostic

kriskowal:
This change improves an error diagnostic that has become common when we introduced a cache that sometimes vended out corrupt bundles.

### DIFF
--- a/packages/check-bundle/package.json
+++ b/packages/check-bundle/package.json
@@ -44,6 +44,7 @@
     "@endo/bundle-source": "^2.2.0",
     "@endo/eslint-config": "^0.4.10",
     "@endo/init": "^0.5.41",
+    "@endo/zip": "^0.2.25",
     "ava": "^3.12.1",
     "babel-eslint": "^10.0.3",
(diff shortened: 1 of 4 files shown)


warner (approved):
change looks sound to me, I'll leave it up to @dtribble to decide whether it provides enough information to fix the problem he observed.

kriskowal:
This is probably necessary but not sufficient to address @dtribble’s needs since the “location” threaded to this point by https://github.com/Agoric/agoric-sdk/pull/5164 would be the bundle ID (SHA-512) and the cache doesn’t leave a breadcrumb mapping hash to original entry module, yet. It may be necessary to incorporate an alleged name in the bundle itself for a genuinely good developer experience, like `{ moduleFormat: 'endoZipBase64', bundleName: '@agoric/zoe/contracts/means-of-production.js', endoZipBase64, endoZipBase64Sha512}` which would need to surface in the `checkBundle` error like `Invalid archive "allegedly @agoric/zoe/contracts/means-of-production.js"`.

PR #1169, opened 2022-04-23
feat(compartment-mapper): Support text and bytes asset modules

kriskowal:
Motivated only to provide an Endo alternative to needing hacks like the use of `import.meta.url` in combination with `fs` or `fetch` powers that cannot be decoupled cleanly for bundling or archiving applications, this change introduces the notion of binary and text asset modules, which behave spiritually in much the same way as `json` type modules.


kriskowal:
cc @erights @mhofman

packages/compartment-mapper/README.md
```diff
@@ -225,8 +225,12 @@ To overcome such obstacles, the compartment mapper will accept a non-standard
 `js` to the corresponding language name, one of `mjs` for ECMAScript modules,
 `cjs` for CommonJS modules, and `json` for JSON modules.
 All other language names are reserved and the defaults for files with the
-extensions `cjs`, `mjs`, and `json` default to the language of the same name
-unless overridden.
+extensions `cjs`, `mjs`, `json`, `text`, and `bytes` default to the language of
```
dckc, line 228: .txt is conventional. Did you consider it?

dckc: I guess folks can say .txt in package.json...

packages/compartment-mapper/NEWS.md
```diff
@@ -1,5 +1,13 @@
 User-visible changes to the compartment mapper:
 
+# Next release
+
+- Adds support by default for "text" and "bytes" as file types with epynomous
```
erights, line 5:
```suggestion
- Adds support by default for "text" and "bytes" as file types with eponymous
```

packages/compartment-mapper/README.md
kriskowal: Indeed, and also considered `.bin` or `.oct` as a nod to 8.3. I’m open to either design direction.

packages/compartment-mapper/test/fixtures-assets/main.js
```diff
@@ -0,0 +1,26 @@
+import text from './text.text';
+import bytes from './bytes.bytes';
+import uint32 from './uint32.uint32';
```
dckc, line 3: very nice.

I wonder if there's a straightforward way to link from docs to test examples such as this one.

dckc (approved):
I can confirm that the use case I'm interested in is tested.

The implementation looks formulaic, so it's likely to be OK, but I didn't review it in detail. Let me know if you'd like me to look closer.

packages/compartment-mapper/README.md
```diff
@@ -225,8 +225,12 @@ To overcome such obstacles, the compartment mapper will accept a non-standard
...
+the same name unless overridden.
+JSON modules export a default object resulting from the conventional JSON.parse
+of the module's UTF-8 encoded bytes.
+Text modules export a default string from the module's UTF-8 encoded bytes.
+Bytes modules export a default ArrayBuffer capturing the module's bytes.
```
naugtur, line 233: Is there prior art to this behavior? I recall using bundler plugins for pulling stuff in, but none I used were configurable really.

packages/compartment-mapper/src/types.js
```diff
@@ -1,7 +1,7 @@
 // @ts-check
 /// <reference types="ses"/>
 
-export const moduleJSDocTypes = true;
+export {};
```
naugtur, line 4: What's this?

naugtur (approved):

packages/compartment-mapper/README.md
kriskowal: Not to my knowledge apart from things I wrote myself:

- https://github.com/montagejs/mr
- https://github.com/gutentags/system

packages/compartment-mapper/src/types.js
kriskowal: TypeScript treats a JavaScript file that contains only JSDoc comments with type annotations as a source of “ambient types”, but we don’t want that, so we need one executable line of JavaScript. This is just an update for. the simpler idiom we’ve converged on at Agoric, effectively a noöp.

packages/compartment-mapper/test/fixtures-assets/main.js
naugtur: You mean like this?

https://github.com/endojs/endo/blob/acc828cf74308b7cdcf000f7492a53e3135bdfd3/packages/compartment-mapper/test/fixtures-assets/main.js#L1-L3

It should work in markdown files too. worth trying

PR #1170, opened 2022-04-23
test(ses): Verify assignability of array length over frozen prototype

kriskowal:
This is a sanity check that seems like a good one to carry forward into test 262 one day. The problem that motivated the sanity check was a hardened array instance. Hardening the prototype of all arrays doesn’t preclude using length assignment to truncate.


kriskowal:
cc @mhofman @naugtur

erights (approved):
LGTM

PR #1171, opened 2022-04-23
fix: all errors have stacks, even if empty

### LINKED ISSUE #1165: errorTaming: 'unsafe' does not work on Firefox
image

Test code:

```js
await import('https://cdn.jsdelivr.net/npm/ses@0.15.15/dist/lockdown.mjs')
lockdown({ errorTaming: 'unsafe' })
new Error().stack.toString()
```

erights:
Fixes #1165 

Implements https://github.com/tc39/proposal-error-stacks/ assuming the fix proposed at https://github.com/tc39/proposal-error-stacks/issues/46

### DIFF
--- a/packages/ses/src/error/tame-error-constructor.js
+++ b/packages/ses/src/error/tame-error-constructor.js
@@ -145,30 +145,28 @@ export default function tameErrorConstructor(
       errorTaming,
       stackFiltering,
     );
-  }
+  } else if (errorTaming === 'unsafe') {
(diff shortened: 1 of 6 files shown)


erights:
@Jack-Works , github won't let me list you as a reviewer, but I would like your review. Thanks.

erights:
How should test this across engines? Have we tried using the tool that test262 uses? Do we use it anywhere?

Jack-Works (approved):
LGTM

gibson042 (approved):

PR #1174, opened 2022-04-26
feat(compartment-mapper): Collect unused compartments in archives

kriskowal:
This change causes archives and bundles to have smaller compartment maps and to
create fewer compartments at runtime.
The motivation for this change is to reduce the probability that a vagary of
`yarn`, like the difference in behavior between `yarn install` and `yarn
install --production` on the hash of resulting bundles, by making the versions
and positions of of unused packages irrelevant.

- fix(compartment-mapper): Package exports may be absolute
- refactor(compartment-mapper): Avoid redundant relativize
- test(compartment-mapper): Test compartment map retention
- feat(compartment-mapper): Archives retain compartments only for retained modules
- docs(compartment-mapper): Update news for archive compartment retention

### DIFF
--- a/packages/compartment-mapper/NEWS.md
+++ b/packages/compartment-mapper/NEWS.md
@@ -7,6 +7,8 @@ User-visible changes to the compartment mapper:
   and bytes as exporting a default ArrayBuffer.
   The `"parsers"` directive in `package.json` can map additional extensions to
   either of these types, in the scope of the declaring package.
+- Compartment maps in archives now only retain compartment descriptors for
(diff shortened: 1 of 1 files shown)


naugtur (approved):
This will make dynamic imports a bit more difficult to pull off in archive if we ever want to and using neighboring packages for assets. None of that is a concern AFAIK.

LGTM

kriskowal:
That’s a good observation. We’ll need a mechanism for express retaining supplementary module graphs so they can be dynamically imported.

PR #1175, opened 2022-04-26
feat(compartment-mapper): Improve archive hash stability

kriskowal:
Before this change, the compartment map would contain a sequence number on every compartment to disambiguate compartments produced by packages with the same name and version but different physical locations.

After this change, compartments only have a disambiguating sequence number if there is a name collision, and only for each package after the first. The order in which colliding compartments appear is based on the lexical order of their respective dependency name paths, favoring the shortest of all the dependency paths that arrive at the same physical location, or as @kumavis calls it, the package’s ’a’a.


packages/compartment-mapper/src/archive.js
```diff
@@ -50,18 +50,73 @@ const resolveLocation = (rel, abs) => new URL(rel, abs).toString();
...
+  // We are very proud of you.
+  const compartmentsByPath = Object.entries(compartments)
+    .map(([name, compartment]) => ({
+      name,
+      path: compartment.path,
+      label: compartment.label,
+    }))
+    .sort((a, b) => {
+      if (a.label === b.label) {
+        return compare(a.path, b.path);
```
naugtur, line 106: This is not how `'a'a` compares paths, at least not anymore.
The preference is for human consumption so 
`zbzb/x` is better than `a/b/x`

Not sure how relevant that is here. I understand this sort is only meant to sort by label and avoid producing different results each time.

packages/compartment-mapper/src/archive.js
```diff
@@ -73,7 +128,7 @@ const renameCompartments = compartments => {
  */
 const translateCompartmentMap = (compartments, sources, renames) => {
   const result = Object.create(null);
-  for (const compartmentName of keys(compartments).sort()) {
+  for (const compartmentName of keys(renames)) {
```
naugtur, line 131: (could) renames->renamedCompartments

packages/compartment-mapper/src/node-modules.js
```diff
@@ -395,6 +396,40 @@ const graphPackages = async (
...
+ * layout tends to differ between installation with and without devopment-time
+ * dependencies.
+ *
+ * @param {Graph} graph
+ * @param {string} location
+ * @param {string} path
+ */
+const trace = (graph, location, path) => {
+  const node = graph[location];
+  if (node.path !== '' && node.path <= path) {
```
naugtur, line 417: This is not how 'a'a compares paths, at least not anymore.
The preference is for human consumption so
zbzb/x is better than a/b/x

Not sure how relevant that is here. I understand this is only meant to aid the sort by label to avoid producing different results each time.

naugtur (approved):

kriskowal:
@naugtur That’s a good observation. My preference is indeed to favor the shortest dependency name path, as an array of components, so I have put this back into a draft state and will follow up on that.

packages/compartment-mapper/src/compartment-map.js
```diff
@@ -13,6 +13,49 @@ const moduleLanguages = [
...
+ * @param {Array<string>} a
+ * @param {Array<string>} b
+ */
+export const pathCompare = (a, b) => {
+  // Prefer the shortest dependency path.
+  if (a.length !== b.length) {
+    return a.length - b.length;
+  }
+  // Otherwise, favor the shortest cumulative length.
+  const aSum = a.reduce(cumulativeLength, 0);
```
naugtur, line 38: my lazy brain: why not just `a.join('').length`
It's obviously less performant but I wanted to know how much. 10 fold.

PR #1178, opened 2022-05-02
Implement `unhandledRejectionTrapping`

michaelfig:
Closes: #1164

# XS approach to unhandled rejections

This PR was inspired by the unhandled rejection support in the XS engine.  Their trick is to keep a `Map<WeakRef<Promise>, reason>` for the rejections that were not handled synchronously.  Then, the map entry is removed if the rejection is handled before it is GCed or the process exits.

We accomplish the same behaviour with a FinalizationRegistry and a combination of a weak and strong map.  That removes the need for periodic scanning of WeakRefs.

# Impact on eventual-send

I also audited the eventual-send implementation to fix the times when a promise had a then handler attached (but no catch), which was the main source of unnecessary unhandled rejections.

The combination of these should provide more verbose errors for actual unhandled rejections when using SES, as well as minimising the amount of noise due to handlers that were attached anytime before the promise was GCed.

# Demo

Here is a demonstration of running under Node:

```js
/* global setTimeout */
import '@endo/init';

const test = async () => {
  const f = Promise.reject(Error('f'));
  const g = Promise.reject(Error('g'));
  // eslint-disable-next-line no-unused-vars
  const h = Promise.reject(Error('h'));

  f.catch(reason => console.log('@@@ handled', `${reason}`));

  console.log('@@@ delay a little bit');
  await new Promise(resolve => setTimeout(resolve, 1500));
  console.log('@@@ delay done');

  g.catch(reason => console.log('@@@ handled', `${reason}`));
};

test();
```

```console
$ node test-unhandled.js
@@@ delay a little bit
@@@ handled Error: f
@@@ delay done
@@@ handled Error: g
SES_UNHANDLED_REJECTION: (Error#1)
Error#1: h

  at test (packages/eventual-send/t.js:8:28)
  at packages/eventual-send/t.js:19:1
  at async Promise.all (index 0)
$
```

And here is the output from the browser (note that it only works in a same-origin script served by http:// or https://):

image


michaelfig:
Thanks to @mhofman for inspiring me to see this change through to a reasonable outcome.  My future self is grateful.

packages/ses/src/error/tame-console.js
```diff
@@ -53,21 +56,56 @@ export const tameConsole = (
...
+  /* eslint-disable @endo/no-polymorphic-call */
+
+  const reportReason = reason =>
+    causalConsole.warn('SES_UNHANDLED_REJECTION:', reason);
+
   // Node.js
   if (errorTrapping !== 'none' && globalThis.process !== undefined) {
-    // eslint-disable-next-line @endo/no-polymorphic-call
-    globalThis.process.on('uncaughtException', error => {
+    const sesDefaultPolicy = globalThis.Symbol.for('sesDefaultPolicy');
```
erights, line 66: I don't understand this yet, but why a symbol rather than a weakmap?

packages/ses/src/error/tame-console.js
```diff
@@ -76,17 +114,42 @@ export const tameConsole = (
...
       causalConsole.error(event.error);
       if (errorTrapping === 'exit' || errorTrapping === 'abort') {
         globalThis.window.location.href = `about:blank`;
       }
     });
+
+    // Maybe track unhandled promise rejections.
+    if (unhandledRejectionPeriodMs && globalThis.WeakRef !== undefined) {
+      const scheduleOnce = callback => {
+        globalThis.setTimeout(callback, unhandledRejectionPeriodMs);
```
erights, line 128: Ok, I just do not understand why timeout is the right tool for this.

erights (commented):
Just comments so far. Review not done.

packages/ses/src/error/tame-console.js
michaelfig: My mistake.  The timer was dead code that I just didn't realize could be entirely removed.

packages/ses/src/error/tame-console.js
michaelfig: Overly complex from a failed experiment.  I've simplified.

packages/eventual-send/src/handled-promise.js
```diff
@@ -546,11 +544,6 @@ export const makeHandledPromise = () => {
         .catch(lose);
     });
 
-    // Harden the fulfillment and rejection, as well as a workaround for
-    // Node.js: silence "Unhandled Rejection" by default when using the static
-    // methods.
-    returnedP.then(harden, harden);
```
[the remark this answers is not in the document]
michaelfig: Yes, we no longer need silencing because the catches are correct.

packages/eventual-send/src/handled-promise.js
```diff
@@ -401,7 +398,7 @@ export const makeHandledPromise = () => {
     },
     getSendOnly(target, prop) {
       prop = coerceToObjectProperty(prop);
-      handle(target, 'getSendOnly', [prop]);
+      handle(target, 'getSendOnly', [prop]).catch(() => {});
```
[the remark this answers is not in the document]
michaelfig: This is just a more surgical approach.  Our blanket approach worked before, but was overbroad.

packages/eventual-send/src/handled-promise.js
```diff
@@ -246,11 +247,6 @@ export const makeHandledPromise = () => {
         // Remove stale pending handlers, set to canonical form.
         shorten(handledP);
 
-        // Ensure our pendingHandler is cleaned up if not already.
-        if (promiseToPendingHandler.has(handledP)) {
-          handledP.then(_ => promiseToPendingHandler.delete(handledP));
```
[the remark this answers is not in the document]
michaelfig: `promiseToPendingHandler` is a WeakMap so it shouldn't be too leaky.

But if we attach just a `.then` handler and handledP is rejected, this expression produces an unhandled rejection.  If 
 we handle it with a `.catch`, then we silence legitimate warnings if `handledP` isn't ever handled by userspace.

packages/eventual-send/src/handled-promise.js
```diff
@@ -182,6 +182,7 @@ export const makeHandledPromise = () => {
       if (typeof amfn === 'function') {
         // Downlevel a missing applyFunction to applyMethod with undefined name.
         const [args] = opArgs;
+        console.log('applyFunction', amfn);
```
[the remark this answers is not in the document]
michaelfig: Removed.

packages/eventual-send/src/handled-promise.js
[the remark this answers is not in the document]
mhofman: An alternative would be:
```js
const cleanup = _ => { promiseToPendingHandler.delete(handledP); }
Promise.resolve().then(() => handledP).then(cleanup , cleanup);
```

That way you don't silence the original handledP, and you cleanup in both settlement paths.

packages/ses/src/commons.js
```diff
@@ -178,6 +179,8 @@ export const iterateArray = uncurryThis(arrayPrototype[iteratorSymbol]);
 export const mapSet = uncurryThis(mapPrototype.set);
 export const mapGet = uncurryThis(mapPrototype.get);
 export const mapHas = uncurryThis(mapPrototype.has);
+export const mapDelete = uncurryThis(mapPrototype.delete);
```
mhofman, line 182: I am befuddled as to why SES code never had a need to `delete` until now...

packages/eventual-send/src/handled-promise.js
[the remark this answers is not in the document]
mhofman: Where do we harden the fulfillment value now?

packages/ses/src/error/unhandled-rejection.js
```diff
@@ -0,0 +1,121 @@
...
+  /** @type {(() => void) | undefined} */
+  let cancelChecking;
+
+  const removeReasonId = reasonId => {
+    mapDelete(idToReason, reasonId);
+    if (cancelChecking && idToReason.size === 0) {
+      // No more unhandled rejections to check, just cancel the check.
+      cancelChecking();
+      cancelChecking = undefined;
+    }
```
mhofman, line 53: Is this leftover from the schedule hook?

packages/ses/src/error/unhandled-rejection.js
```diff
@@ -0,0 +1,121 @@
...
+  const promiseToReasonId = new WeakMap();
+
+  /**
+   * Clean up and report the reason for a GCed unhandled rejection.
+   *
+   * @param {ReasonId} heldReasonId
+   */
+  const finalizeDroppedPromise = heldReasonId => {
+    const reason = mapGet(idToReason, heldReasonId);
+    if (reason !== undefined) {
```
mhofman, line 66: Nit, technically you can reject with `undefined` value...
```suggestion
    if (mapHas(idToReason, heldReasonId)) {
      const reason = mapGet(idToReason, heldReasonId);
```

packages/ses/src/error/unhandled-rejection.js
```diff
@@ -0,0 +1,121 @@
...
+      removeReasonId(heldReasonId);
+      reportReason(reason);
+    }
+  };
+
+  /** @type {FinalizationRegistry<ReasonId>} */
+  const promiseToReason = new FinalizationRegistry(finalizeDroppedPromise);
+
+  /**
+   * Add a weakref to a rejected promise and its corresponding reason if there
```
mhofman, line 76: No longer using actual weakref

mhofman (commented):
This is looking great, mostly some nits, and a genuine question on the hardening of the fulfillment value of a handled promise.

kriskowal (changes requested):
This is fantastic. I’m glad the platforms allow us to do this now, because this is clearly the right way to deal with unhandled rejections. I’ll second @mhofman’s feedback and after they’re addressed, I’ll take another look, but I think this is on track.

packages/eventual-send/src/handled-promise.js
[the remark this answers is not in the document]
michaelfig: > An alternative would be:

Really cool!  Thanks for that.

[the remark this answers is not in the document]
michaelfig: Oh, but it doesn't work for me.  The following doesn't trigger the `unhandledRejection` callback when it should:

```js
const h = Promise.reject(Error('expected to be uncaught'));
Promise.resolve().then(() => h).then(() => {}, () => {});
```

packages/eventual-send/src/handled-promise.js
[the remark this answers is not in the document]
michaelfig: I no longer think that fulfilment needs to be hardened, since that feature is incompatible with avoiding silencing the unhandled rejections.

packages/eventual-send/src/handled-promise.js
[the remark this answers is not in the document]
mhofman: Oh right because the initial promise was actually handled in the `then` assimilation. Only option in that case is to re-create a promise which won't be handled:

```js
const h = Promise.reject(Error('expected to be uncaught'));
// Create a new promise which won't be handled
Promise.resolve().then(() => h);
h.then(() => {}, () => {});
```

packages/eventual-send/src/handled-promise.js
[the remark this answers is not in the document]
michaelfig: Yep, removed.

packages/eventual-send/src/handled-promise.js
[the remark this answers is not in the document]
mhofman: Incompatible how? What about doing:

```js
returnedP.then(hardenVoid).catch(() => {});
```

packages/ses/src/error/unhandled-rejection.js
michaelfig: Yes, I've removed it.

packages/eventual-send/src/handled-promise.js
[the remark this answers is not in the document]
michaelfig: I'm not satisfied with any of these answers.  Either they:
a) unconditionally suppress the unhandled rejection even if `handledP` doesn't have a `.catch` handler attached by the user; or
b) create an additional unhandled rejection that cannot be caught (because it falls out of scope), even if the original `handledP` has a `.catch` handler attached by the user.

I have tried but failed to find a solution that avoids both of these problems.  Am I missing something obvious?

The problem to solve is to come up with YOUR-CODE-HERE manipulating `h` such that:
```js
// Should result in "unhandled rejection" event
h = Promise.reject('unhandled'); ...YOUR-CODE-HERE
// Should NOT result in "unhandled rejection" event
h = Promise.reject('handled'); h.catch(() => {}); ...YOUR-CODE-HERE
```

It is my assertion that the only `YOUR-CODE-HERE` that can work is something that doesn't register any handlers on `h`--neither `.then`, `.catch`, or `.finally`.

packages/eventual-send/src/handled-promise.js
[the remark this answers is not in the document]
michaelfig: > Incompatible how?

See the challenge above.  Touching `returnedP` in any way breaks my requirements.

packages/eventual-send/src/handled-promise.js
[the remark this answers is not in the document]
mhofman: Yeah it looks like the reason `.finally` worked when I tried it is because it creates a new promise that was itself unhandled.

The requirement to not chain is indeed incompatible with unhandled rejection tracking. If we can't put a harden before resolving the promise to the user (as in before in the chain vs before in time relying on promise reactions ordering), then yes there is no way.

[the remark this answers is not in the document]
mhofman: I think there is a more general question too whether a generic handled promise should work in a non-lockdown environment, in which case the explicit hardening should only be done as a lockdown of HandledPromise.

packages/ses/src/error/tame-console.js
```diff
@@ -53,21 +56,56 @@ export const tameConsole = (
   // utter that. That unnecessary shim forces the whole bundle into sloppy mode,
   // which in turn breaks SES's strict mode invariant.
 
+  /* eslint-disable @endo/no-polymorphic-call */
```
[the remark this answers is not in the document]
michaelfig: It's not a global disable, it's a disable the rule on all following lines until a corresponding `eslint-enable`.  I specifically did not want to disable the rule for other parts of the function.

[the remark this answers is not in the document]
mhofman: With the new comment above I'm happy with this approach.

michaelfig:
@kriskowal and @erights this is R4R again.  I believe it cleanly implements `errorTrapping`-like semantics for `unhandledRejectionTrapping`.

packages/ses/docs/lockdown.md
```diff
@@ -26,7 +26,8 @@ Each option is explained in its own section below.
 | `localeTaming`   | `'safe'`         | `'unsafe'`     | `toLocaleString`           |
 | `consoleTaming`  | `'safe'`         | `'unsafe'`     | deep stacks                |
 | `errorTaming`    | `'safe'`         | `'unsafe'`     | `errorInstance.stack`      |
-| `errorTrapping`  | `'platform'`     | `'exit'` `'abort'` `'report'` | handling of uncaught exceptions |
+| `errorTrapping`  | `'platform'`     | `'exit'` `'abort'` `'report'` `'none'` | handling of uncaught exceptions |
+| `unhandledRejectionTrapping`  | `'none'`     | `'exit'` `'abort'` `'report'` | handling of finalized unhandled rejections |
```
erights, line 30: In the table, this differs from the `errorTrapping` option by omitting `platform`. But the documentation of this option below still mentions platform.

Having this default to `'none'` seems wrong. Why not at least `'report'`? (Below I also ask whether the actual platform behavior should be an option.)

packages/ses/docs/lockdown.md
```diff
@@ -367,6 +368,52 @@ the container to exit explicitly, and we highly recommend setting
...
+were not handled in the same turn in which they were created, and potentially a
+scary warning that says unhandled rejections may cause the process to exit in a
+future release.
+
+By setting a non-`'none'` value for `unhandledRejectionTrapping`, the event
+handler will only be triggered by unhandled rejections when they are finalized.
+This enables the program to attach rejection handlers asynchronously without
+triggering the SES trap handler.
+
+- `'platform'`: same as `'report'` for web and Node.js
```
erights, line 408: Should this `'platform'` bullet be removed? If it is just a way to say `'report'`, yes please remove it. However, should we also allow the actual platform behavior? If so, should it be a non-default option?

packages/ses/index.d.ts
```diff
@@ -24,6 +24,12 @@ export interface LockdownOptions {
   localeTaming?: 'safe' | 'unsafe';
   consoleTaming?: 'safe' | 'unsafe';
   errorTrapping?: 'platform' | 'exit' | 'abort' | 'report' | 'none';
+  unhandledRejectionTrapping?:
+    | 'platform'
```
erights, line 28: ?

packages/ses/src/lockdown-shim.js
```diff
@@ -160,6 +160,10 @@ export const repairIntrinsics = (options = {}) => {
   const {
     errorTaming = getenv('LOCKDOWN_ERROR_TAMING', 'safe'),
     errorTrapping = getenv('LOCKDOWN_ERROR_TRAPPING', 'platform'),
+    unhandledRejectionTrapping = getenv(
+      'LOCKDOWN_UNHANDLED_REJECTION_TRAPPING',
+      'platform',
```
erights, line 165: Is `'platform'` the default?

packages/ses/src/lockdown-shim.js
```diff
@@ -286,18 +290,23 @@ export const repairIntrinsics = (options = {}) => {
...
+   *
+   * @type {((error: any) => string | undefined) | undefined}
+   */
   let optGetStackString;
   if (errorTaming !== 'unsafe') {
     optGetStackString = intrinsics['%InitialGetStackString%'];
   }
   const consoleRecord = tameConsole(
-    // @ts-ignore tameConsole does its own input validation
+    // @ts-expect-error2 tameConsole does its own input validation
```
erights, line 306: typo?
```suggestion
    // @ts-expect-error tameConsole does its own input validation
```

packages/eventual-send/src/handled-promise.js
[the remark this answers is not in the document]
erights: What do you mean "touching it"?

@michaelfig , you and I talked since this thread and reached a decision about hardening. Do you remember what that was?

packages/ses/src/error/tame-console.js
```diff
@@ -17,11 +18,13 @@ const originalConsole = console;
  *
  * @param {"safe" | "unsafe"} consoleTaming
  * @param {"platform" | "exit" | "abort" | "report" | "none"} [errorTrapping]
+ * @param {"platform" | "exit" | "abort" | "report" | "none"} [unhandledRejectionTrapping]
  * @param {GetStackString=} optGetStackString
  */
 export const tameConsole = (
   consoleTaming = 'safe',
   errorTrapping = 'platform',
+  unhandledRejectionTrapping = 'platform',
```
erights, line 27: As mentioned elsewhere, do we want `'platform'` to be the default? In any case, glad that `'none'` isn't the default after all. That was scary ;)

Also mentioned elsewhere: An option named `'platform'` should provide the actual platform behavior.

packages/ses/src/error/tame-console.js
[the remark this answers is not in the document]
erights: my too. Resolving.

erights (commented):
Review not finished, but looking good to me so far.

kriskowal (approved):
My comments are for you to take or leave. I reviewed eventual-send lightly since it looks like @mhofman is reviewing that closely. In short, I think we could safely pare down the unhandled rejection taming options to just `report` (default) or `none`, and I will happily rely on your discretion.

packages/ses/src/lockdown-shim.js
michaelfig: Now it is `'report'`.

packages/ses/index.d.ts
michaelfig: Now just `'report' | 'none'`

packages/ses/docs/lockdown.md
michaelfig: Changed to default to `'report'`.

packages/ses/docs/lockdown.md
michaelfig: I've removed `'platform'`.

packages/ses/docs/lockdown.md
```diff
@@ -27,7 +27,7 @@ Each option is explained in its own section below.
 | `consoleTaming`  | `'safe'`         | `'unsafe'`     | deep stacks                |
 | `errorTaming`    | `'safe'`         | `'unsafe'`     | `errorInstance.stack`      |
 | `errorTrapping`  | `'platform'`     | `'exit'` `'abort'` `'report'` `'none'` | handling of uncaught exceptions |
-| `unhandledRejectionTrapping`  | `'none'`     | `'exit'` `'abort'` `'report'` | handling of finalized unhandled rejections |
+| `unhandledRejectionTrapping`  | `'report'`     | `'report'` `'none'` | handling of finalized unhandled rejections |
```
erights, line 30:
```suggestion
| `unhandledRejectionTrapping`  |  `'report'`  | `'none'` | handling of finalized unhandled rejections |
```
Third column is only other choices.

packages/ses/docs/lockdown.md
```diff
@@ -375,11 +375,7 @@ able to install platform-specific finalized (rather than just same-turn)
...
 
 ```js
-lockdown(); // defaults to 'none'
-// or
-lockdown({ unhandledRejectionTrapping: 'exit' }); // report and exit
-// or
-lockdown({ unhandledRejectionTrapping: 'abort' }); // report and drop a core dump
+lockdown(); // defaults to 'report'
 // or
 lockdown({ unhandledRejectionTrapping: 'report' }); // just report
```
erights, line 380:
```suggestion
lockdown({ unhandledRejectionTrapping: 'none' }); // a comment
```

packages/ses/docs/lockdown.md
```diff
@@ -405,12 +401,7 @@ handler will only be triggered by unhandled rejections when they are finalized.
...
 triggering the SES trap handler.
 
-- `'platform'`: same as `'report'` for web and Node.js
 - `'report'`: just report finalized unhandled rejections to the tamed console so stack traces appear.
-- `'exit'`: reports and exits on Node.js, reports and navigates away on the
-  web.
-- `'abort'`: reports and aborts a Node.js process, leaving a core dump for
-  postmortem analysis, reports and navigates away on the web.
 - `'none'`: do not install traps for finalized unhandled rejections. Errors are
   likely to appear as `{}` when they are reported by the default trap.
```
erights, line 406: > Errors are likely to appear as `{}` when they are reported by the default trap.

Uh oh. I am confused about the meaning of `'none'`. What default trap? The platform's original behavior? In that case it should be called `'platform'` not `'none'`. I thought `'none'` meant to suppress any reporting.

erights (changes requested):

packages/ses/docs/lockdown.md
michaelfig: I used the same meaning as `errorTrapping`: `'none'` means don't install a handler.

kriskowal: I agree with @michaelfig’s interpretation of `'none'`: do nothing, do not install handlers.

erights: Huh. I see that you're right, and it even uses the same text as its explanation. But `errorTrapping` also has a `'platform'` option. If `'none'` means "leave the platform behavior as is", what does `'platform'` mean? I cannot figure it out from the text in lockdown.md.

Maybe a question for @kriskowal ?

kriskowal: For `errorTrapping`, `'platform'` means to use the idiomatic behavior for the platform: either `'exit'` in Node.js or `'report'` on the web. The other options override the platform-specific idiom, forcing a non-idiomatic behavior if elected. For example, `'exit'` on the web navigates away rather than allow the program to continue running, rendering the error undiscoverable.

michaelfig: ``platform`` means to install a handler that has different behaviour depending on the platform.

erights: I still do not understand. Operationally, what is the observable difference between `'platform'` and `'none'`? Which is closer to what the raw platform itself does?

michaelfig: > I still do not understand. Operationally, what is the observable difference between `'platform'`

Print the deep stack trace.

> and `'none'`?

Don't print the deep stack trace.

erights: Thanks!

erights (commented):
not done yet, sorry

erights (approved):
LGTM

I have not fully reviewed the two key files: tame-console.js and unhandled-rejection.js, and so am deferring to @mhofman and @kriskowal 's approvals on those. But I did not spot any problems.

packages/ses/test/console-rejection-trap/hazard.js
```diff
@@ -0,0 +1,5 @@
+/* global setImmediate */
+setImmediate(() => {
+  Promise.reject(new Error('I am once again throwing an error'));
```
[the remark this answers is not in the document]
michaelfig: Will change the message.

packages/ses/docs/lockdown.md
```diff
@@ -367,6 +368,41 @@ the container to exit explicitly, and we highly recommend setting
 - `'none'`: do not install traps for uncaught exceptions. Errors are likely to
   appear as `{}` when they are reported by the default trap.
 
+## `unhandledRejectionTrapping` Options
+
+**Background**: Same concerns as `errorTrapping`, but in addition, SES may be
+able to install platform-specific finalized (rather than just same-turn)
+unhandled rejection trapping.
```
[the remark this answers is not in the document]
michaelfig: If SES can't, then we get whatever the platform does by default.  I'll rephrase.

PR #1183, opened 2022-05-19
fix: or vs vs convention

erights:
The early entries in lockdown.md followed a consistent convention when showing the different possible calls to `lockdown` within each section. More recent entries did not, probably because it was not clear that there was a convention to be followed. This PR is stacked on #1178 and repairs that convention throughout lockdown.md as of #1178 .

Also updates domainTaming to show that it now defaults to `'safe'`.

@michaelfig if you'd like to just fold this into #1178, please do. I would prefer that. But if you'd like me to hold this as a follow-on PR, that would be ok.

@kriskowal I included you because some of the repairs are, I think, to text you wrote so you should be aware of this.


michaelfig (approved):
LGTM

PR #1184, opened 2022-05-19
Fix StaticModuleRecord types in SES and static-module-record to be consistent with implementation

ritave:


kriskowal (approved):
Thank you!

PR #1186, opened 2022-05-20
fix: decodeToJustin accepts slot backrefs

### LINKED ISSUE #1185: decodeToJustin can break on valid input
https://github.com/Agoric/agoric-sdk/pull/5401 uses `decodeToJustin` to produce more readable diagnostic. But on the way to this PR we came across valid input that `decodeToJustin` fails to decode. The problem arises when the same slot appears more than once. After the first appearance, the `iface` field of the slot record does not appear on further appearances. But `decodeToJustin` asserts that `iface` is a string (and therefore present). This is doubly wrong if we still want to consider the `iface` field optional.

Attn @Chris-Hibbert

erights:
Fixes #1185 

As of https://github.com/Agoric/agoric-sdk/pull/5401 we'll use `decodeToJustin`, so export it as we always expected to.

Take @dtribble 's suggestion to rename `getSlotVal` to simply `slot`.

### DIFF
--- a/packages/marshal/index.js
+++ b/packages/marshal/index.js
@@ -24,8 +24,8 @@ export { Remotable, Far, ToFarFunction } from './src/make-far.js';
 
 export { QCLASS, makeMarshal } from './src/marshal.js';
 export { stringify, parse } from './src/marshal-stringify.js';
-// Works, but not yet used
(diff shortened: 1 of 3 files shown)


michaelfig (approved):
LGTM!

PR #1190, opened 2022-05-24
refactor(stream): Stand on promise-kit dependency

kriskowal:
This refactor is merely deduplication of makePromiseKit in the stream library.
Since writing stream, we moved promise kit to Endo, simplified, and committed to hardened-only streams.


erights (approved):
LGTM

PR #1192, opened 2022-05-24
WIP fix: guard against promise reentrancy

erights:
DO NOT MERGE: Based on an ancient master, and likely doesn't solve the problem anyway.

Completes the defense against known promise reentrancy by adding a constructor test.

### DIFF
--- a/packages/eventual-send/src/handled-promise.js
+++ b/packages/eventual-send/src/handled-promise.js
@@ -45,6 +45,24 @@ const coerceToObjectProperty = specimen => {
   return String(specimen);
 };
 
+/**
+ * This function assumes that `p` is already guaranteed to be a platform
+ * promise. If this function returns true, then both `p.then(...` and
+ * `await p` will be safe against reentrancy attacks.
+ *
+ * @param {Promise} p
+ * @returns {boolean}
+ */
+const isPromiseSafe = p => {
+  return (
+    isFrozen(p) &&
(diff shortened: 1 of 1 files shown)


mhofman (changes requested):
I don't believe this will work as-is. I think our approach should be:
- assume the object is not a proxy (later we can add a brand check as the first test)
- Check that the prototype is `Promise.prototype`
- Check that the object doesn't have own `then` or `constructor` props
- Only then check that the object is a native promise with `Promise.resolve(p) === p`, as that step can trigger any `constructor` or `then` getters, either own or on the prototype chain.

It's actually not strictly necessary to check that the object is frozen since after a proxy brand check, we know that no traps can be triggered while our code is on the stack, but we still should perform a frozen check since it'll guarantee that the promise won't add an own `then`/`constructor`, or change the proto chain, once our stack has unwound (I'd like to think about what it'd take to make `HandledPromise` not reliant on `lockdown` and frozen objects but that's another story).

The biggest complexity however is not the "safe promise" check, but first `Promise.resolve()` done before we even call this predicate. https://github.com/endojs/endo/blob/db2ac2c357522c26754a7e2d1fc882c0d21d1a1d/packages/eventual-send/src/handled-promise.js#L437

erights:
If we try to do this again, we should start a new PR. Closing this one as fatally stale and probably could not have worked anyway.

Issue #1195, opened 2022-05-25
Compartment options.name not convert to string


Jack-Works:
![image](https://user-images.githubusercontent.com/5390719/170204944-7511854d-35cc-4b16-b044-402d975a1ebb.png)

kriskowal:
Thank you for reporting. This should be a quick fix.

PR #1197, opened 2022-05-27
fix(ses): Compartment name from object with toString

kriskowal:
- test(ses): Test object with toString as compartment name
- fix(ses): Fix compartment with name from object with toString

Fixes: #1195


kriskowal:
cc @Jack-Works

Jack-Works (approved):

mhofman (approved):

PR #1198, opened 2022-05-30
feat(init): `@endo/init/legacy.js`

michaelfig:
For some scripts I was writing, I needed to write my own `lockdown.js` to do:

```js
import { lockdown } from '@endo/lockdown';
import './pre-remoting.js';

const options = {
  overrideTaming: 'severe',
  stackFiltering: 'verbose',
  errorTaming: 'unsafe',
};
lockdown(options);
```

This combination of options proved extremely useful for locking down one-off scripts with pervasive override mistakes, and that have no need to protect themselves against error stacks.

So, I created `@endo/init/legacy.js`.

## Node bug

I tried adding exports so that:

```js
import { lockdown } from '@endo/init/pre-remoting.js';
```

could be done to obtain the lockdown wrapper without piercing `@endo/lockdown`.  However, I ran into what looks like a bug in at least Node v16.15.0:

```
file:///Users/michael/agoric/endo/packages/init/legacy.js:2
import { lockdown } from './pre-remoting.js';
         ^^^^^^^^
SyntaxError: The requested module './pre-remoting.js' does not provide an export named 'lockdown'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:128:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:194:5)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:385:24)
    at async loadESM (node:internal/process/esm_loader:88:5)
    at async handleMainPromise (node:internal/modules/run_main:61:12)
```

That's weird, given the definition of `pre-remoting.js` in this PR.  I hope somebody with better module-fu can look into this.


packages/init/legacy.js
```diff
@@ -0,0 +1,12 @@
+// legacy.js - call tolerant lockdown with default Agoric shims
+import { lockdown } from '@endo/lockdown';
+import './pre-remoting.js';
+
+// The loosest, non-stack-hiding SES environment.
```
erights, line 5: Not really loosest. I do not suggest making it looser; just altering the comment. Your point is that it is loose enough for your purposes. It's good you're not being looser than that.

erights (commented):
Looks fine to me as far as I can tell. But I'll leave approval to someone with better module-fu than I. No further suggestions or objections.

mhofman (commented):
I'll need to defer to @kriskowal as I'm not the most familiar with the layering of files and exports in `lockdown` and `init`.

packages/init/pre-bundle-source.js
```diff
@@ -6,4 +6,4 @@
 
 // Use a package self-reference to go through the "exports" resolution
 // eslint-disable-next-line import/no-extraneous-dependencies
-import '@endo/init/pre.js';
+export * from '@endo/init/pre.js';
```
kriskowal, line 9: Please consider `export { lockdown } from ...`. The hazard I have in mind is that a name might be added or removed from the originating module and cause strange name collisions in distant dependees.

kriskowal (approved):
Riffing on names for legacy: migration, adoption, embrace, conversion, loose, introduction, experiment, welcome, step-1.

kriskowal:
@michaelfig In the description, you mention a mystery in the module system. Did changing `import` to `export * from` in those few places address that issue?

michaelfig:
> @michaelfig In the description, you mention a mystery in the module system. Did changing `import` to `export * from` in those few places address that issue?

No, the mystery continues to exist (Node.js's ESM implementation only, I believe).

PR #1201, opened 2022-06-01
refactor eslint config

turadg:
The `@endo` eslint config mixes concerns of how Endo is developed (e.g. jsdoc rules, prettier) with how to use Endo (e.g. state of globals in SES). Agoric SDK then gets them both when [it extends `@endo`](https://github.com/Agoric/agoric-sdk/blob/master/packages/eslint-config/eslint-config.json#L3). There are also multiple places that extend `@jessie.js/recommended`.

This PR is a step in a proposal to normalize how Jessie / Endo / Agoric eslint all fit together. 

- `@org/eslint-plugin` packages contain features to support the _use_ of artifacts from the org, including recommended rules.
- `@org/eslint-config` packages are strongly opinionated about style for _development_ within that org. 
- DX concerns that don't affect runtime (e.g. jsdoc, prettier, whitespace rules) are only included in the _development_ config and not the _use_ config

Once this lands, I would update Agoric SDK accordingly.


kriskowal:
I defer to @michaelfig for the stamp. Separating the concerns of “rules that apply in a Node.js start compartment” and “rules that apply in an Endo environment” seems like a good idea to me. Rules that apply only within the SES implementation should, of course, be package-scoped.

PR #1202, opened 2022-06-02
Implement providing the url value in import.meta

naugtur:
Implemented meta.url via module descriptors. 

Bundler complained, but I think I've fixed it correctly.


packages/compartment-mapper/src/import-hook.js
```diff
@@ -214,7 +220,7 @@ export const makeImportHookMaker = (
             sourceLocation: moduleLocation,
             parser,
             bytes: transformedBytes,
-            record,
+            record: concreteRecord,
```
naugtur, line 223: this is for bundling

packages/compartment-mapper/src/import-hook.js
```diff
@@ -196,10 +196,16 @@ export const makeImportHookMaker = (
...
             };
           } else {
-            record = concreteRecord;
+            record = {
+              record: concreteRecord,
+              specifier: moduleSpecifier,
+              meta: { url: moduleLocation },
+            };
           }
+          // ^ to be simplified, left unchanged for a more comprehensive diff
```
kriskowal, line 208: This and the “draft” status suggests to me you’re not looking for a stamp yet. I’ll leave comments for now.

packages/compartment-mapper/src/import-hook.js
kriskowal: I’m convinced this is the right way to go about this for now. Bundling will eventually need to be able to follow aliases through the module descriptor graph, either by shortening paths in pre-processing or threading descriptors into the bundle. Probably the former.

packages/ses/index.d.ts
```diff
@@ -67,6 +67,7 @@ export type FinalStaticModuleType =
 export interface RedirectStaticModuleInterface {
   record: FinalStaticModuleType;
   specifier: string;
+  meta?: any;
```
kriskowal, line 70: I think we’re leaning toward naming this `importMeta`.

packages/compartment-mapper/test/test-main.js
```diff
@@ -22,8 +25,16 @@ const assertFixture = (t, { namespace, globals, globalLexicals }) => {
...
     typemoduleImplied,
     typeparsers,
+    metaUrl,
   } = namespace;
 
+  if (testCategoryHint === 'Location') {
+    // matching any character where / or \ would be
+    t.regex(metaUrl, /fixtures-0.node_modules.app.main\.js$/);
+  } else {
+    t.is(metaUrl, null);
```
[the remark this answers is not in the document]
kriskowal: The more I see this, the more `import.meta.url` feels like a misfeature and that we should nudge the Endo community in the direction of text and binary asset modules, which we already support.

I think it’s a desirable property of Endo that, if a program runs with `importLocation`, it’ll also run with `makeArchive` and `parseArchive`, and in the latter case, `import.meta.url` and `__dirname` are not useful. However, with asset modules, we can preserve that guarantee, since the necessary assets are explicitly retained.

That being said, I can see value in the feature for Node.js portability, and we in fact could really use this feature real-soon-now™ in Agoric’s codebase. So, may I recommend that we add an option to `importLocation` to enable `needsImportMetaUrl`?

naugtur:
:warning: now I need to redo most of this to match the import.meta implementation we ended up merging.  
Expect a push -f

packages/ses/index.d.ts
```diff
@@ -67,6 +67,7 @@ export type FinalStaticModuleType =
 export interface RedirectStaticModuleInterface {
   record: FinalStaticModuleType;
   specifier: string;
+  importMeta?: any;
 }
 
 export type StaticModuleType =
```
naugtur, line 73: This part will need a rewrite to match current spec. That's going to happen alongside a wider rewrite to match shapes defined in the compartment proposal later.

packages/compartment-mapper/src/import-archive.js
```diff
@@ -151,7 +151,7 @@ const makeArchiveImportHookMaker = (
         sourceLocation,
         packageLocation,
       );
-      return record;
+      return { record, specifier: moduleSpecifier, importMeta: { url: null } };
```
naugtur, line 154: make it a 
```suggestion
      return { record, specifier: moduleSpecifier, importMeta: {  } };
```

kriskowal: Or rather, omit the `importMeta` property entirely.

naugtur:
@kriskowal All done, I need a rubber stamp.

packages/compartment-mapper/test/fixtures-0/node_modules/app/main.js
```diff
@@ -26,6 +26,8 @@ export { builtin } from 'builtin';
 export const receivedGlobalProperty = globalProperty;
 export const receivedGlobalLexical = globalLexical;
 
+export const metaUrl = import.meta.url;
```
kriskowal, line 29: Would buy `importMetaUrl` since that kind of explicitness is en vogue.

kriskowal (approved):

PR #1204, opened 2022-06-02
fix(eventual-send): use `!Object.is(a, b)` instead of `a !== b` for NaNs

michaelfig:
We need to be more defensive in the `HandledPromise` innards or else we get the following if `NaN` is shortened:

```
TypeError#18: Invalid value used as weak map key
  at WeakMap.set (<anonymous>)
  at shorten (file:///Users/michael/agoric/agoric-sdk/node_modules/@endo/eventual-send/src/handled-promise.js:114:35)
  at handledResolve (file:///Users/michael/agoric/agoric-sdk/node_modules/@endo/eventual-send/src/handled-promise.js:232:17)
  at resolveHandled (file:///Users/michael/agoric/agoric-sdk/node_modules/@endo/eventual-send/src/handled-promise.js:380:9)
  at win (file:///Users/michael/agoric/agoric-sdk/node_modules/@endo/eventual-send/src/handled-promise.js:510:11)
  at file:///Users/michael/agoric/agoric-sdk/node_modules/@endo/eventual-send/src/handled-promise.js:528:20
  at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

### DIFF
--- a/packages/eventual-send/src/handled-promise.js
+++ b/packages/eventual-send/src/handled-promise.js
@@ -26,6 +26,7 @@ const {
   getPrototypeOf,
   setPrototypeOf,
   isFrozen,
+  is: objectIs,
 } = Object;
 
 const { apply, construct } = Reflect;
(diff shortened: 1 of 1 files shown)


erights (approved):
LGTM. Fascinating!

PR #1205, opened 2022-06-03
`@endo/captp` client robustness features

michaelfig:
Added some new features:

- implement WeakRef/FinalizationRegistry-based garbage collection (refactoring for this tightened up the implementation considerably)
- allow optional `nearOptions` and `farOptions` for `makeLoopback`
- expose `isOnlyLocal(specimen)` to see if a passable specimen contains any nonlocal promises/presences
- add `importHook(val, slot)` and `exportHook(val, slot)` to allow interception of vals that are being imported/exported.  This allows warning about (or preventing) in-transit object references that may be severed if the client exits

### DIFF
--- a/packages/captp/src/captp.js
+++ b/packages/captp/src/captp.js
@@ -218,8 +218,31 @@ export const makeCapTP = (
     // associated slot number.
     const slot = valToSlot.get(val);
     assert.typeof(slot, 'string');
+
     return slot;
   }
+
+  /**
+   * @type {ConvertSlotToVal<CapTPSlot>}
+   */
+  const assertValIsLocal = val => {
+    const slot = valToSlot.get(val);
+    assert(
+      !(slot && slot[1] === '-'),
+      X`Value ${val} slot ${slot} indicates it is remote; we are expecting only local`,
+    );
+  };
+
+  const { serialize: assertOnlyLocal } = makeMarshal(assertValIsLocal);
+  const isOnlyLocal = specimen => {
+    // Try marshalling the object, but throw on references to remote objects.
+    try {
+      assertOnlyLocal(specimen);
+      return true;
+    } catch (e) {
+      return false;
+    }
+  };
 
   /**
    * Generate a new question in the questions table and set up a new
@@ -332,6 +355,7 @@ export const makeCapTP = (
     // side.
     const otherDir = theirSlot[1] === '+' ? '-' : '+';
     const slot = `${theirSlot[0]}${otherDir}${theirSlot.slice(2)}`;
+
     if (!slotToVal.has(slot)) {
       // Make a new handled promise for the slot.
       const pr = makeRemoteKit(slot);
@@ -618,6 +642,7 @@ export const makeCapTP = (
     abort,
     dispatch,
     getBootstrap,
+    isOnlyLocal,
     serialize,
     unserialize,
     makeTrapHandler,
--- a/packages/captp/src/captp.js
+++ b/packages/captp/src/captp.js
@@ -30,14 +30,15 @@ const isThenable = maybeThenable =>
   maybeThenable && typeof maybeThenable.then === 'function';
 
 /**
- * @typedef {object} CapTPOptions the options to makeCapTP
- * @property {(err: any) => void} onReject
- * @property {number} epoch an integer tag to attach to all messages in order to
+ * @typedef {Object} CapTPOptions the options to makeCapTP
+ * @property {(val: unknown, slot: CapTPSlot) => void} [exportHook]
+ * @property {(err: any) => void} [onReject]
+ * @property {number} [epoch] an integer tag to attach to all messages in order to
  * assist in ignoring earlier defunct instance's messages
- * @property {TrapGuest} trapGuest if specified, enable this CapTP (guest) to
+ * @property {TrapGuest} [trapGuest] if specified, enable this CapTP (guest) to
  * use Trap(target) to block while the recipient (host) resolves and
  * communicates the response to the message
- * @property {TrapHost} trapHost if specified, enable this CapTP (host) to serve
+ * @property {TrapHost} [trapHost] if specified, enable this CapTP (host) to serve
  * objects marked with makeTrapHandler to synchronous clients (guests)
  */
 
@@ -47,7 +48,7 @@ const isThenable = maybeThenable =>
  * @param {string} ourId our name for the current side
  * @param {(obj: Record<string, any>) => void} rawSend send a JSONable packet
  * @param {any} bootstrapObj the object to export to the other side
- * @param {Partial<CapTPOptions>} opts options to the connection
+ * @param {CapTPOptions} opts options to the connection
  */
 export const makeCapTP = (
   ourId,
@@ -58,6 +59,7 @@ export const makeCapTP = (
   const {
     onReject = err => console.error('CapTP', ourId, 'exception:', err),
     epoch = 0,
+    exportHook,
     trapGuest,
     trapHost,
   } = opts;
@@ -171,6 +173,9 @@ export const makeCapTP = (
         lastPromiseID += 1;
         const promiseID = lastPromiseID;
         slot = `p+${promiseID}`;
+        if (exportHook) {
+          exportHook(val, slot);
+        }
         // Set up promise listener to inform other side when this promise
         // is fulfilled/broken
         E.when(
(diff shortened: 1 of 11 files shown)


packages/captp/src/captp.js
```diff
@@ -311,29 +387,22 @@ export const makeCapTP = (
...
-      const pr = makeRemoteKit(slot);
+      const { promise, settler } = makeRemoteKit(slot);
       if (slot[0] === 'o' || slot[0] === 't') {
         if (iface === undefined) {
           iface = `Alleged: Presence ${ourId} ${slot}`;
         }
         // A new remote presence
         // Use Remotable rather than Far to make a remote from a presence
-        val = Remotable(iface, undefined, pr.resPres());
+        val = Remotable(iface, undefined, settler.resolvePresence());
```
erights, line 401: You declared the `resolvePresence` method as void return, but the implementation returns a presence that you seem to be using here intentionally.

packages/captp/src/captp.js
```diff
@@ -137,14 +174,18 @@ export const makeCapTP = (
...
   // question key
-  let lastQuestionID = 0;
 
-  /** @type {Map<string, any>} */
-  const questions = new Map(); // chosen by us
+  /**
+   * @typedef {object} Settler
+   * @property {(result?: any) => void} resolve
+   * @property {(reason: any) => void} reject
+   * @property {(handler?: any) => void} resolvePresence
```
erights, line 182: I like Settlers! I have some other uses for them.

But I'm confused about the type of `resolvePresence` method. Clearly it is meant to be like `resolveWithPresence`. But the `handler` parameter here is never used. When would it be used?

Also, this is declared as returning void. However, the implementation returns the presence that `resolveWithPresence` returns, and the only call to it makes use of that presence!

Perhaps

```suggestion
   * @property {() => object} resolvePresence
```
?

If you do want to pass the `handler`, which would sometimes be forwarded to `resolveWithPresence`, what about the full type of `resolveWithPresence` including `options`? If you do go that far, perhaps `Settler` is synonymous with `HandledExecutor`?

But then the `presenceHandler` parameter of `resolveWithPresence` would need to be optional, which does not seem to make sense for `HandledExecutor`. `reason` also differs in optionality. But come to think of it, I don't understand why the `value` and `reason` of `HandledExecutor` are optional anyway?

packages/captp/src/captp.js
```diff
@@ -249,59 +318,66 @@ export const makeCapTP = (
...
-      pr.res = res;
-    }, handler);
+    /** @type {Settler | undefined} */
+    let settler;
+    const promise = new HandledPromise(
+      (resolve, reject, resolveWithPresence) => {
+        settler = Far('settler', {
+          resolve,
+          reject,
+          resolvePresence: () => resolveWithPresence(handler),
```
erights, line 369: If you did mean this to be void return, how about

```suggestion
          resolvePresence: () => { resolveWithPresence(handler); },
```
?

This is another example where we need to stop checking `void` and start checking that the function *implicitly or explicitly* returns only `undefined`. Otherwise, we'll have accidental capability leaks, as this one would be if the type were intentional.

packages/captp/src/captp.js
```diff
@@ -137,14 +174,18 @@ export const makeCapTP = (
...
-  const questions = new Map(); // chosen by us
+  /**
+   * @typedef {object} Settler
+   * @property {(result?: any) => void} resolve
+   * @property {(reason: any) => void} reject
+   * @property {(handler?: any) => void} resolvePresence
+   */
+
+  /** @type {Map<CapTPSlot, Settler>} */
+  const settlers = new Map();
```
erights, line 186: Does this substitute for both `questions` and `imports`? That's surprising.

packages/captp/src/captp.js
```diff
@@ -167,8 +208,11 @@ export const makeCapTP = (
         // and use that to construct the slot name.  Promise slots are prefaced
         // with 'p+'.
         lastPromiseID += 1;
-        const promiseID = lastPromiseID;
-        slot = `p+${promiseID}`;
+        slot = `p+${lastPromiseID}`;
+        const promiseID = reverseSlot(slot);
+        if (exportHook) {
+          exportHook(val, slot);
```
erights, line 214: Yes, I can see how this would be very useful. But why no `importHook`? I'm happy to omit it but would like to understand the asymmetry.

packages/captp/src/captp.js
```diff
@@ -167,8 +208,11 @@ export const makeCapTP = (
...
+        }
         // Set up promise listener to inform other side when this promise
         // is fulfilled/broken
         val.then(
```
erights, line 218: How safe is this from reentrancy?

packages/captp/src/captp.js
```diff
@@ -206,21 +254,45 @@ export const makeCapTP = (
...
+  const { serialize: assertOnlyLocal } = makeMarshal(assertValIsLocal);
+  const isOnlyLocal = specimen => {
+    // Try marshalling the object, but throw on references to remote objects.
+    try {
+      assertOnlyLocal(specimen);
+      return true;
+    } catch (e) {
+      return false;
+    }
+  };
```
erights, line 281: OMG I hate to see a try/catch used to convert an assert into a predicate, where predicate failure is considered non-exceptional. I use your `isFoo/assertFoo/checkFoo` pattern alot and love it! Is there anything we can or should do here that is not too much trouble for the short term?

packages/captp/src/captp.js
erights: Does this mean we can finally stop spelling this `resPres` ;) ?

packages/captp/src/captp.js
```diff
@@ -249,59 +318,66 @@ export const makeCapTP = (
...
+    );
+    assert(settler);
 
     // Silence the unhandled rejection warning, but don't affect
     // the user's handlers.
-    pr.p.catch(e => quietReject(e, false));
+    promise.catch(e => quietReject(e, false));
 
-    return harden(pr);
+    return harden({ promise, settler });
```
erights, line 380: If you did declare a `RemoteKit` type to be this record's type, I'd probably find myself using this elsewhere. Just sayin.

packages/captp/src/captp.js
```diff
@@ -538,11 +614,9 @@ export const makeCapTP = (
         // Deliver the object, even though we're unplugged.
         rawSend(obj);
       }
-      for (const pr of questions.values()) {
-        pr.rej(reason);
-      }
-      for (const pr of imports.values()) {
-        pr.rej(reason);
+      slotToVal.clear();
+      for (const settler of settlers.values()) {
```
erights, line 618: Well, unifying `questions` and `imports` does pay off. I suppose I'm less surprised now.

(No change suggested)

erights (changes requested):
You need to incorporate at least the bug fixes from https://github.com/endojs/endo/pull/1216

(Unless I'm mistaken about those, of course, in which case I'd like to understand)

packages/captp/test/engine-gc.js
```diff
@@ -0,0 +1,16 @@
+import v8 from 'v8';
```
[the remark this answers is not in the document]
turadg: We do have an Ava clone for XS: https://github.com/Agoric/agoric-sdk/blob/3e4d66428e2b4cad3ab65045b805e20ff7cc746c/packages/xsnap/src/avaXS.js

Not saying this PR needs to make an XS test too, but it would be helpful for this module to document why this isn't done and perhaps what it would take to pull it off.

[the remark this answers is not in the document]
michaelfig: Done and pushed.

packages/captp/src/finalize.js
```diff
@@ -0,0 +1,131 @@
...
+ * timing-independent side channel.
+ *
+ * This non-determinism also enables code to escape deterministic replay. In a
+ * blockchain context, this could cause validators to differ from each other,
+ * preventing consensus, and thus preventing chain progress.
+ *
+ * JS standards weakrefs have been carefully designed so that operations which
+ * `deref()` a weakref cause that weakref to remain stable for the remainder of
+ * that turn. The operations below guaranteed to do this derefing are `has`,
+ * `get`, `set`, `delete`. Note that neither `clear` nor `getSize` are
```
erights, line 42:
```suggestion
 * `get`, `set`, `delete`. Note that neither `clearWithoutFinalizing` nor `getSize` are
```

erights (approved):
LGTM

I did not review from scratch this pass, and thus do not have a full understanding. Instead, I'm relying on my previous review, when I knew more than I still remember.

PR #1210, opened 2022-06-09
Implement and shim `atob` and `btoa`

michaelfig:
These are useful and standard Web/Node.js APIs for Base64 support.

At least `protobufjs` relies on having them in the global scope (Web-compatible) when `Buffer` is unavailable.

### DIFF
--- a/packages/base64/shim.js
+++ b/packages/base64/shim.js
@@ -1,0 +1,12 @@
+/* global globalThis */
+// @ts-check
+import { atob } from './atob.js';
+import { btoa } from './btoa.js';
+
+if (globalThis.atob === undefined) {
+  globalThis.atob = atob;
+}
+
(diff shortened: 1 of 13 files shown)


packages/base64/shim.js
```diff
@@ -0,0 +1,12 @@
+/* global globalThis */
+// @ts-check
+import { atob as ourAtob } from './src/atob.js';
+import { btoa as ourBtoa } from './src/btoa.js';
+
+if (typeof atob === 'undefined') {
```
kriskowal, line 6: Would also accept `globalThis.atob === undefined` and not need the `as`.

kriskowal (approved):
This appears to cover all the bases. Please add a section to `NEWS.md` to advertise the new `shim.js`. Consider moving `./src/?to?.js` up a level and adding them to `exports` too.

PR #1211, opened 2022-06-10
feat(marshal): display slot values if supplied

michaelfig:
Allow an optional `slots` argument to `decodeToJustin` that allows printing of `slotToVal(slot, "iface")`.

### DIFF
--- a/packages/marshal/src/marshal-justin.js
+++ b/packages/marshal/src/marshal-justin.js
@@ -8,6 +8,7 @@ import { isObject } from './helpers/passStyle-helpers.js';
 import { AtAtPrefixPattern, passableSymbolForName } from './helpers/symbol.js';
 
 /** @typedef {import('./types.js').Encoding} Encoding */
+/** @template T @typedef {import('./types.js').CapData<T>} CapData */
 
 const { ownKeys } = Reflect;
 const { isArray } = Array;
@@ -115,9 +116,10 @@ const identPattern = /^[a-zA-Z]\w*$/;
 /**
  * @param {Encoding} encoding
  * @param {boolean=} shouldIndent
+ * @param {any[]} [slots]
  * @returns {string}
  */
-const decodeToJustin = (encoding, shouldIndent = false) => {
+const decodeToJustin = (encoding, shouldIndent = false, slots = []) => {
   /**
    * The first pass does some input validation.
    * Its control flow should mirror `recur` as closely as possible
@@ -254,7 +256,7 @@ const decodeToJustin = (encoding, shouldIndent = false) => {
   };
 
   const makeIndenter = shouldIndent ? makeYesIndenter : makeNoIndenter;
-  const out = makeIndenter();
+  let out = makeIndenter();
 
   /**
    * This is the second pass recursion after the first pass `prepare`.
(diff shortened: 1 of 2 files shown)


erights (commented):
Could you provide a link to where you (plan to) use it? Otherwise, I'm having trouble imagining the motivation.

michaelfig:
> Could you provide a link to where you (plan to) use it? Otherwise, I'm having trouble imagining the motivation.

It's in the `agoric follow` command for consuming off-chain subscriptions (`@agoric/casting`):
https://github.com/Agoric/agoric-sdk/blob/02c97e9f25cda307825a9a8e9dae8b8995b86d73/packages/agoric-cli/src/follow.js#L44

Each such subscription usually consists of marshalled data, for which the slots are meaningful data that can indicate the path of how to obtain a live reference to an on-chain object given a marshalling context.  (The slots essentially indicate clist entries, and are only informational unless you hold a live unserialization ocap.)

Here is an illustrative (partly aspirational) example based on what I'm trying to achieve with this PR.

## Before

```
$ agoric follow :wallet.agoric1xyzzy.purses --ojustin
[{
  id: 99n,
  currentBalance: {
    value: 123n,
    brand: slot(0, 'Alleged: public brand')
  },
  actions: slot(1, 'Alleged: private purse actions')
}]
^C
```

see how the `slot` rendering doesn't contain enough information to understand how to reconstitute the ocaps just given this output (even if you know which Agoric instance it came from)?

## After

```
$ agoric follow :wallet.agoric1xyzzy.purses --ojustin
[{
  id: 99n,
  currentBalance: {
    value: 123n,
    brand: slotToVal('board/board01929', 'Alleged: public brand')
  },
  actions: slotToVal('wallet/agoric1xyzzy/o+11', 'Alleged: private purse actions')
}]
^C
```

With that information (and knowing which Agoric instance it came from), a consumer with access to the ocap corresponding to `board` can reconstitute the public brand.  A consumer with access to the ocap corresponding to `wallet/agoric1xyzzy` can reconstitute the private purse actions.

packages/marshal/src/marshal-justin.js
```diff
@@ -358,12 +360,18 @@ const decodeToJustin = (encoding, shouldIndent = false) => {
         case 'slot': {
           let { index, iface } = rawTree;
           index = Number(Nat(index));
-          if (iface === undefined) {
-            return out.next(`slot(${index})`);
-          } else {
+          if (index < slots.length) {
+            const slot = quote(slots[index]);
```
erights, line 364: The other places in this file we call `quote` (aka `JSON.stringify`) we know that we're providing it data that can be faithfully JSON.stringified. From your example, I suspect you have in mind only cases where the quotes slot value is a string. Would it be appropriate to limit to those cases with an assert? If not, can you think of any cheap test that a JSON.stringify will be full fidelity?

erights (approved):
LGTM

Without this, the justin test sets up a compartment environment so that the justin string is tested for round tripping through eval. If you modify that test to provide a `slotToVal` binding in addition to its `slot` binding, you could extend those to test round tripping for these too.

packages/marshal/src/marshal-justin.js
michaelfig: I implemented full fidelity (nested decode instead of JSON)!

michaelfig:
> you could extend those to test round tripping for these too.

I tested round tripping.  Please take another look before I merge.

PR #1213, opened 2022-06-10
fix: tests use debug settings

erights:
I normally run with 
```sh
export LOCKDOWN_ERROR_TAMING=unsafe
```
This caused a failure in a test case comparing error messages, that were made more explanatory with this setting. The problem is that our tests were run with a neutral `lockdown` call that was letting this setting be set by the environment variable, but the test itself happened to depend on this option being `'safe'`. I think our best practice should be that `test-*.js` ava test files[1] should use the `@endo/init/debug.js` settings rather than the `@endo/init` settings. Then the error messages during testing will be more informative without endangering info leakage in production.

[1] aside from those whose purpose is to test the lockdown options, or whose purpose is to test that sensitive info in error messages get properly redacted.

### DIFF
--- a/packages/check-bundle/test/test-check-bundle.js
+++ b/packages/check-bundle/test/test-check-bundle.js
@@ -1,6 +1,6 @@
 // @ts-check
 import '@endo/init/pre-bundle-source.js';
-import '@endo/init';
+import '@endo/init/debug.js';
 import test from 'ava';
 import * as fs from 'fs';
 import * as url from 'url';
@@ -76,7 +76,7 @@ test('bundle and check endo zip base64 package absent hash', async t => {
     checkBundle(lightBundle, computeSha512, 'fixture/main.js'),
     {
       message:
-        "checkBundle cannot bundle without the property 'endoZipBase64Sha512', which must be a string, got (a string)",
+        'checkBundle cannot bundle without the property \'endoZipBase64Sha512\', which must be a string, got "undefined"',
     },
   );
 });
--- a/packages/init/test/test-async_hooks.js
+++ b/packages/init/test/test-async_hooks.js
@@ -2,7 +2,7 @@
 
 // Use a package self-reference to go through the "exports" resolution
 // eslint-disable-next-line import/no-extraneous-dependencies
-import '@endo/init';
+import '@endo/init/debug.js';
 import test from 'ava';
 import { createHook } from 'async_hooks';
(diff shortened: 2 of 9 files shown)


packages/check-bundle/test/test-check-bundle.js
```diff
@@ -76,7 +76,7 @@ test('bundle and check endo zip base64 package absent hash', async t => {
     checkBundle(lightBundle, computeSha512, 'fixture/main.js'),
     {
       message:
-        "checkBundle cannot bundle without the property 'endoZipBase64Sha512', which must be a string, got (a string)",
+        'checkBundle cannot bundle without the property \'endoZipBase64Sha512\', which must be a string, got "undefined"',
```
erights, line 79: This was the case causing the failure.

erights:
I left one `@endo/init` unchanged because it is in `packages/init` and so might be testing `init` itself. I'm not sure. That is `test-async-hooks.js`. I suspect this should be changed as well. Please let me know if I should change it as well.

erights:
> I left one `@endo/init` unchanged because it is in `packages/init` and so might be testing `init` itself. I'm not sure. That is `test-async-hooks.js`. I suspect this should be changed as well. Please let me know if I should change it as well.

I have since changed my mind, and the PR no longer makes a special case for `test-async-hooks.js`. However, I still want feedback on this before merging.

packages/init/test/test-async_hooks.js
```diff
@@ -2,7 +2,7 @@
 
 // Use a package self-reference to go through the "exports" resolution
 // eslint-disable-next-line import/no-extraneous-dependencies
-import '@endo/init';
+import '@endo/init/debug.js';
```
erights, line 5: Reviewers, since this is in `packages/init`, I'd like feedback on whether this change is appropriate before merging.

kriskowal: I defer to @mhofman’s opinion on this test.

kriskowal (approved):

packages/init/test/test-async_hooks.js
mhofman: If debug transitively imports pre throught the package.json module map, i think it's ok

mhofman: Thinking about it, it should since it's a public entry point itself.

mhofman (approved):

PR #1215, opened 2022-06-12
feat(marshal): add optional input/output validators

michaelfig:
Intended to support Agoric/agoric-sdk#5579, by allowing the abrupt termination of serialization/unserialization if memory budgets are exceeded.


erights (commented):
(As mentioned verbally)

I need to understand the motivations for this better. Both immediate and anticipated future uses. If the motivations are only instrumentation, to gather diagnostic info, then the API should only enable gathering info, not changing output. POLA FTW!

packages/marshal/src/marshal.js
```diff
@@ -168,15 +183,25 @@ export const makeMarshal = (
...
+     * @param {string} key
      * @returns {Encoding}
      */
-    const encode = val => {
+    const encode = (val, key) => {
       if (ErrorHelper.canBeValid(val)) {
+        // FIXME: We can't protect against the validator permuting the error,
+        // since it is in our contract not to harden it.
+        if (rawEncodingValidator) {
+          rawEncodingValidator(val);
```
erights, line 194: Should this be
```suggestion
          rawEncodingValidator(key, val);
```
?

packages/marshal/src/marshal.js
```diff
@@ -275,15 +302,27 @@ export const makeMarshal = (
       }
     };
 
-    const encoded = encode(root);
+    // Harden to protect against the validator.
```
erights, line 305: Is this Harden comment out of place?

packages/marshal/src/types.js
```diff
@@ -169,17 +169,33 @@ export {};
...
+/**
+ * @callback Validator
+ * @param {string} key
+ * @param {Passable} value
+ */
+
+/**
+ * @callback MakeValidator
+ * @param {boolean} jsonPass - true if the validator is to be used in the JSON
+ * replacer/reviver.  Otherwise, used when encoding/decoding the contents of the
```
erights, line 181: This surprises me. Without looking yet at the actual code which uses it, I'd expect either
   * these two cases are sufficiently similar that no flag is needed, or
   * these cases are different enough that they should be represented as distinct functions.

Of course, some cases fall in the middle ground where a flag is the right compromise. But I don't understand if this is one of those.

erights (commented):
If we just limit the size of the overall message, would any of this mechanism still be needed?

michaelfig:
> If we just limit the size of the overall message, would any of this mechanism still be needed?

Yes.  A message may come in under the total byte counts, but still have contents that should not validate (like JSON object keys that are large enough to cause problems, bigints that are out of the acceptable range, or strings that aren't a specific encoding).

packages/marshal/src/types.js
michaelfig: I choose distinct functions, an encoding pass and a JSON pass.  Will implement that, if you agree.

michaelfig: Actually, just `serialize(Passable, inputValidator)` and `unserialize(capdata, outputValidator)`.  The JSON passes can be done before or after serialization without losing  basic functionality.

packages/marshal/src/marshal.js
```diff
@@ -81,7 +81,7 @@ export const makeMarshal = (
   /**
    * @type {Serialize<Slot>}
    */
-  const serialize = root => {
+  const serialize = (root, inputValidator) => {
```
erights, line 84:
```suggestion
  const serialize = (root, inputValidator = undefined) => {
```

erights: I generally like to see optionality explicitly in the parameter list declaration. When the types are not local, this is even more important.

packages/marshal/src/marshal.js
```diff
@@ -168,15 +168,25 @@ export const makeMarshal = (
      * a canonical-json stringify of the encoded form.
      *
      * @param {Passable} val
+     * @param {string} key
```
erights, line 171: Why is the `key` second? I'd expect either
```suggestion
     * @param {Passable} val
     * @param {string} [key]
```
or
```suggestion
     * @param {string} key
     * @param {Passable} val
```

packages/marshal/src/marshal.js
```diff
@@ -168,15 +168,25 @@ export const makeMarshal = (
...
      * @returns {Encoding}
      */
-    const encode = val => {
+    const encode = (val, key) => {
       if (ErrorHelper.canBeValid(val)) {
+        // FIXME: We can't protect against the validator permuting the error,
+        // since it is in our contract not to harden it.
```
erights, line 177: "FIXME" implies that it is fixable. Is it?
```suggestion
        // Note: We can't protect against the validator permuting the error,
        // since it is in our contract not to harden it.
```

packages/marshal/src/marshal.js
```diff
@@ -275,15 +287,15 @@ export const makeMarshal = (
...
+    const encoded = encode(root, '');
 
     return harden({
       body: JSON.stringify(encoded),
       slots,
     });
   };
 
-  const makeFullRevive = slots => {
+  const makeFullRevive = (slots, validator) => {
```
erights, line 298:
```suggestion
  const makeFullRevive = (slots, outputValidator) => {
```
Is this right?

packages/marshal/src/marshal.js
```diff
@@ -333,11 +345,22 @@ export const makeMarshal = (
      * representation. Rather, fullRevive must validate that.
      *
      * @param {Encoding} rawTree must be hardened
+     * @param {string} key
```
erights, line 348: Likewise, I'd expect either
```suggestion
     * @param {Encoding} rawTree must be hardened
     * @param {string} [key]
```
or
```suggestion
     * @param {string} key
     * @param {Encoding} rawTree must be hardened
```

packages/marshal/src/marshal.js
```diff
@@ -333,11 +345,22 @@ export const makeMarshal = (
...
      */
-    function fullRevive(rawTree) {
+    function fullRevive(rawTree, key) {
```
erights, line 350: If we do swap the order of params, we should probably pick a name other than `fullRevive`. We probably should anyway.

packages/marshal/src/marshal.js
```diff
@@ -475,9 +498,9 @@ export const makeMarshal = (
             'string',
             X`Property ${name} of ${rawTree} must be a string`,
           );
-          result[name] = fullRevive(rawTree[name]);
+          result[name] = fullRevive(rawTree[name], String(name));
```
erights, line 501: Given the `assert` above, this could be just
```suggestion
          result[name] = fullRevive(rawTree[name], name);
```

erights (approved):
LGTM

Yes, much better!

michaelfig:
Closing, but can still be used for future reference if the memory budget is implemented in Agoric SDK.

PR #1216, opened 2022-06-13
Suggested changes to #1205's weak value map semantics and implementation.

erights:
My suggestions on #1205 's finalize.js were better expressed as a separate PR.

However, this suggestion as is breaks something. I'm not sure what. So be careful in acting on these suggestions.

Take the inline comments as commentary for you, not as comments to include. 

We do not need to iterate on this PR. Take anything you think adds value, or merge it into your's if you'd like.

### DIFF
--- a/packages/captp/src/finalize.js
+++ b/packages/captp/src/finalize.js
@@ -114,9 +114,9 @@ export const makeFinalizingMap = () => {
       //       they would call `delete`.
     },
     getSize: () => keyToRef.size,
-    // `entries` is omitted both from the type and from the fakeFinalizingMap,
-    // and so should be empty here as well. If it is present, since it needs to
-    // deref anyway, it should filter out those that were already collected.
+    // `entries` is omitted both from the type and from the `fakeFinalizingMap`,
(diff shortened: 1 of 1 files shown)


packages/captp/src/captp.js
```diff
@@ -431,6 +423,7 @@ export const makeCapTP = (
     async CTP_DROP(obj) {
       const { slotID } = obj;
       slotToVal.delete(slotID);
+      send({ type: 'CTP_DROP', slotID: reverseSlot(slotID), epoch });
```
michaelfig, line 426: Adding this just ping-pongs `CTP_DROP` calls back and forth between endpoints.  That's not the desired behaviour.

michaelfig (commented):
Here's how you broke the implementation.

packages/captp/src/captp.js
erights: good to know. thanks.

erights:
#1205 incorporated the changes this PR was about, so closing this one.

PR #1218, opened 2022-06-15
fix: trivial changes to make lint happy

erights:
For some reason, when a newline is placed between a function jsdoc-type declaration and the function itself, ts type checking makes fewer bad rejections while otherwise seeming to stay as accurate. Discovered by accident. I have no idea why.

Also removing extraneous spaces within the outer curlies of jsdoc type decls, as our style is generally to omit them. By omitting them consistently, it is easier to find things with trivial greps.

See https://github.com/Agoric/agoric-sdk/pull/5597

### DIFF
--- a/packages/compartment-mapper/src/compartment-map.js
+++ b/packages/compartment-mapper/src/compartment-map.js
@@ -467,6 +467,7 @@ const assertEntry = (allegedEntry, url) => {
  * @param {string} [url]
  * @returns {asserts compartmentMap is import('./types.js').CompartmentMapDescriptor}
  */
+
 export const assertCompartmentMap = (
   allegedCompartmentMap,
   url = '<unknown-compartment-map.json>',
@@ -488,5 +489,4 @@ export const assertCompartmentMap = (
   assertTags(tags, url);
   assertEntry(entry, url);
   assertCompartments(compartments, url);
-  return undefined;
 };
--- a/packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
+++ b/packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
@@ -20,6 +20,7 @@ const noTrailingSlash = path => {
 
 /**
  * Generates values for __filename and __dirname from location
+ *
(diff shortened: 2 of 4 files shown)


packages/compartment-mapper/src/compartment-map.js
```diff
@@ -467,6 +467,7 @@ const assertEntry = (allegedEntry, url) => {
  * @param {string} [url]
  * @returns {asserts compartmentMap is import('./types.js').CompartmentMapDescriptor}
  */
+
```
kriskowal, line 470: This newline seems out-of-place. I’m sure it’s omissible nearly everywhere and I’m not sure why this one’s different.

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
```diff
@@ -20,6 +20,7 @@ const noTrailingSlash = path => {
 
 /**
  * Generates values for __filename and __dirname from location
+ *
```
kriskowal, line 23: This newline is unsurprising to me. Thanks.

kriskowal (approved):

packages/compartment-mapper/src/compartment-map.js
erights: Without both the `return undefined;` I removed at line 491, and this newline, I get
```
$ yarn lint
yarn run v1.22.0
$ yarn lint:types && yarn lint:js
$ tsc -p jsconfig.json
$ eslint .

/Users/markmiller/src/ongithub/endojs/endo/packages/compartment-mapper/src/compartment-map.js
  465:1  warning  JSDoc @returns declaration present but return expression not available in function  jsdoc/require-returns-check

✖ 1 problem (0 errors, 1 warning)
```

See the similar example at https://github.com/Agoric/agoric-sdk/pull/5597

and speculation at https://github.com/Agoric/agoric-sdk/pull/5597#issuecomment-1155936997

> looking at the shell trace of the failure I record above, it looks like typescript itself doesn't fail mysteriously, but eslint does, based on eslint also trying to interpret jsdoc type declarations.

PR #1222, opened 2022-06-22
fix: workaround leak through promise resolvers

mhofman:
The investigation into https://github.com/Agoric/agoric-sdk/issues/5507 lead us into discovering a leak caused by promises, in particular when using [`Promise.race` with a long pending promise](https://github.com/nodejs/node/issues/17469). The root cause really seem that the promise resolve functions hold onto the promise strongly (instead of just an internal record representing the reactions registered on the promise), and through that, the result of said promise. `Promise.race` is implemented as adding the result's promise's resolver functions to each of the contenders reactions, causing the long pending promise to keep the raced result alive.

This PR re-implements `Promise.race` based on the solution in https://github.com/nodejs/node/issues/17469#issuecomment-685216777. It also updates `makePromiseKit` to return resolvers that will sever their hold on the original resolvers (and thus the promise) once called. That latter approach, while simpler, is not sufficient for `Promise.race` as reactions would keep queueing on the long pending promise (engines are not sufficiently smart to realize a reaction is a native resolver that has become inert).

Finally it installs the new race implementation as a `Promise.race` vetted shim in `@endo/init`.

(I'm half considering re-writing promises in userland with proper support for short-circuiting, which would avoid most of these issues, but the few spec places creating primordial promises would damper that. A long term solution might be to rewrite the spec  with proper short-circuiting, maybe as part of [tc39/proposal-faster-promise-adoption](https://github.com/tc39/proposal-faster-promise-adoption))

### DIFF
--- a/packages/promise-kit/index.js
+++ b/packages/promise-kit/index.js
@@ -79,11 +79,11 @@ harden(isPromise);
  * Unlike `Promise.race` it cleans up after itself so a non-resolved value doesn't hold onto
  * the result promise.
  *
- * @template {readonly unknown[] | []} T
- * @param {T} values An array of Promises.
- * @returns {Promise<Awaited<T[number]>>} A new Promise.
+ * @template T
+ * @param {Iterable<T>} values An iterable of Promises.
+ * @returns {Promise<Awaited<T>>} A new Promise.
  */
 export function racePromises(values) {
-  return harden(memoRace(values));
+  return harden(memoRace.call(BestPipelinablePromise, values));
 }
 harden(racePromises);
--- a/packages/promise-kit/index.js
+++ b/packages/promise-kit/index.js
@@ -3,6 +3,8 @@
 
 /// <reference types="ses"/>
 
+import { memoRace } from './src/memo-race.js';
+
 /** @type {PromiseConstructor} */
 const BestPipelinablePromise = globalThis.HandledPromise || Promise;
 
@@ -69,3 +71,19 @@ export function isPromise(maybePromise) {
   return Promise.resolve(maybePromise) === maybePromise;
 }
 harden(isPromise);
+
+/**
+ * Creates a Promise that is resolved or rejected when any of the provided Promises are resolved
+ * or rejected.
+ *
+ * Unlike `Promise.race` it cleans up after itself so a non-resolved value doesn't hold onto
+ * the result promise.
+ *
+ * @template {readonly unknown[] | []} T
+ * @param {T} values An array of Promises.
+ * @returns {Promise<Awaited<T[number]>>} A new Promise.
+ */
+export function racePromises(values) {
+  return harden(memoRace(values));
+}
+harden(racePromises);
--- a/packages/promise-kit/index.js
+++ b/packages/promise-kit/index.js
@@ -3,6 +3,7 @@
 
 /// <reference types="ses"/>
 
+import { makeReleasingExecutorKit } from './src/promise-executor-kit.js';
 import { memoRace } from './src/memo-race.js';
 
 export * from './src/is-promise.js';
@@ -13,11 +14,6 @@ export * from './src/types.js';
 const BestPipelinablePromise = globalThis.HandledPromise || Promise;
 
 /**
- * Needed to prevent type errors where functions are detected to be undefined.
- */
-const NOOP_INITIALIZER = harden(() => {});
-
-/**
  * makePromiseKit() builds a Promise object, and returns a record
  * containing the promise itself, as well as separate facets for resolving
  * and rejecting it.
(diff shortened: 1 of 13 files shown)


kriskowal (approved):
We could alternately move this into `ses` as a taming of the existing `Promise.race`. It would only be a slight feature creep.

michaelfig:
> We could alternately move this into `ses` as a taming of the existing `Promise.race`. It would only be a slight feature creep.

Why not just as a separate "vetted shim" package, loaded by `@endo/init`?

kriskowal:
> > We could alternately move this into `ses` as a taming of the existing `Promise.race`. It would only be a slight feature creep.
> 
> Why not just as a separate "vetted shim" package, loaded by `@endo/init`?

That’s even better.

mhofman:
Updated to use a vetted shim approach, and reverted the explicit uses to rely on the shim.

Also updated the implementation to follow the spec more closely. It'd be nice to test against test262, but I don't believe we have the infra for that.

packages/promise-kit/shim.js
```diff
@@ -0,0 +1,4 @@
+import { memoRace } from './memo-race.js';
+
+// Unconditionally replace with a non-leaking version
+Promise.race = memoRace;
```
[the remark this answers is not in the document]
mhofman: It is and is one of the thing I fixed in https://github.com/endojs/endo/pull/1222/commits/8cfe2a1b106468402f11e00c079987b0ed5e0f61 to preserve the behavior.

packages/promise-kit/index.js
```diff
@@ -3,17 +3,26 @@
...
- * @typedef {Object} PromiseKit A reified Promise
+ * @typedef {object} PromiseKit A reified Promise
  * @property {(value: ERef<T>) => void} resolve
  * @property {(reason: any) => void} reject
  * @property {Promise<T>} promise
  */
 
+/**
+ * @template T
+ * @callback PromiseExecutor The promise executor
```
erights, line 21: How aware is the promiseKit package of eventual-send / HandledPromise? Should the executor type include `resolveToPresence`?

packages/promise-kit/memo-race.js
```diff
@@ -0,0 +1,154 @@
...
+   * @template {PromiseConstructor} [P=PromiseConstructor]
+   * @this {P | void}
+   * @param {Iterable<T>} values An iterable of Promises.
+   * @returns {Promise<Awaited<T>>} A new Promise.
+   */
+  race(values) {
+    let deferred;
+    /** @type {T[]} */
+    const cachedValues = [];
+    const C = this || Promise;
```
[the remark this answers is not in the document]
erights: Yeah, as a shim, I think we should be as strict as possible about spec compliance. Thanks.

packages/promise-kit/index.js
mhofman: It is not and ignoring it, like the existing `makePromiseKit`

packages/promise-kit/index.js
```diff
@@ -3,17 +3,26 @@
 
 /// <reference types="ses"/>
```
erights, line 4: Now that index.js is growing and has other *.js peers, we need to follow our normal convention of having index.js just be a pass-through, and have all the substantive code in non-index `src/*.js` files.

erights (changes requested):
Not done yet. Just giving you bits as they occur to minimize latency.

kriskowal (approved):

packages/promise-kit/index.js
kriskowal: In my opinion, this is still small enough not to need additional ceremony. Both of the modules at this level need entry points in `package.json` because they are exposed API.

kriskowal: Which is to say, moving code into `src`  would not eliminate these two files, but hollow both out and leave `export * from './src/x.js` ceremony.

mhofman: I just took a stab at splitting into different files and it's not straightforward, mainly because of the types intermingling and the `BestPipelinablePromise` stuff. Best I could do is move the promise executor stuff into `src`, maybe even the `isPromise` too, but leave the others behind.

packages/promise-kit/src/promise-executor-kit.js
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * makeReleasingExecutorKit() builds resolve/reject functions which drop references
+ * to the resolve/reject functions gathered from an executor to be used with a
+ * promise constructor.
+ *
+ * @template T
+ * @returns {Pick<import('./types.js').PromiseKit<T>, 'resolve' | 'reject'> & { executor: PromiseExecutor<T>}}
+ */
+export const makeReleasingExecutorKit = () => {
```
erights, line 20: Elegant!

packages/promise-kit/src/memo-race.js
```diff
@@ -0,0 +1,154 @@
...
+
+For more information, please refer to <http://unlicense.org/>
+*/
+
+const isObject = value => Object(value) === value;
+
+/**
+ * @template [T=any]
+ * @typedef {object} Deferred
+ * @property {(value?: T | PromiseLike<T> ) => void} resolve
```
erights, line 37: Are you trying to stay close to the original source, or trying to adapt to our style? If the first, ignore this suggestion.

```suggestion
 * @property {(value?: ERef<T> ) => void} resolve
```

Also import `isObject` rather than redefine it?

erights (approved):
I did not do a thorough review of memo-race.js . But nothing seemed wrong, and you already have @kriskowal 's approval. I'll defer to that. Please proceed.

packages/promise-kit/src/memo-race.js
mhofman: I was trying to stay close to the original, but at this point, I've made so many changes this would be warranted.

Regarding `isObject`, it's actually not exported by any package, at least not any that would make sense taking a dependency on.

mhofman:
Added a blurb to `NEWS.md` and addressed @erights type feedback in https://github.com/endojs/endo/commit/0377da065fd1af4012797121c6f59f9223b86f40 then rebase the whole thing cleaning up the history: [0377da0..219e49d](https://github.com/endojs/endo/compare/0377da065fd1af4012797121c6f59f9223b86f40..219e49dcdb61bddad9bb6ff8662aed38b6d89b8e)

PR #1230, opened 2022-07-01
fix(compartment-mapper): avoid mislabeling cjs files as esm based on type field 2

naugtur:
Introduce assumptions about language vs extension into the makeExtensionParser function to prioritize resolving by extension and fall back to configured type only for ambiguous extensions (currently only js)

This looks nicer than version 1 but leaves invalid entries in `types` being ignored, but risking something else uses them.

An alternative: https://github.com/endojs/endo/pull/1231


packages/compartment-mapper/src/link.js
```diff
@@ -54,6 +54,14 @@ const q = JSON.stringify;
  */
 const has = (object, key) => apply(hasOwnProperty, object, [key]);
 
+/**
+ * Decide if extension is not clearly indicating a parser for a file
+ *
+ * @param {string} extension
+ * @returns {boolean}
+ */
+const isAmbiguousExtension = extension => extension === 'js';
```
kriskowal, line 63: The opposite might be easier to read, `extensionImpliesLanguage`

kriskowal (approved):

PR #1231, opened 2022-07-01
fix(compartment-mapper): avoid mislabeling cjs files as esm based on type field 1

naugtur:
A version of the fix that doesn't introduce assumptions about language vs extension into the makeExtensionParser function, but at a cost of parsing extensions in graphPackage where it aggregates types.


kriskowal (commented):
I like #1230 more but trust your judgement if stacking these changes helps.

naugtur:
I'm looking at a choice between stacking this onto the other solution so that `types` field doesn't contain lies OR renaming `types` to indicate it's not the final verdict on the module type. 

Renaming would be an obvious choice but the rename propagates across the codebase further than I'd want it to - which made me hesitate. 
Not closing this for now

PR #1241, opened 2022-07-27
fix: more hardens

erights:
Some hardens that seem to be missing, and seem harmless to add.


kriskowal (approved):
Great. Now we just need better, faster, and stronger, because our work is never over.

packages/ses/src/tame-function-constructors.js
```diff
@@ -49,7 +49,7 @@ export default function tameFunctionConstructors() {
     FERAL_FUNCTION.prototype.constructor('return 1');
   } catch (ignore) {
     // Throws, no need to patch.
-    return {};
+    return harden({});
```
Jack-Works, line 52: This is wrong. It crashes in our environment.

image

Because when this function is called, there is no `harden` defined yet.

PR #1243, opened 2022-08-04
fix: remove dead environment-options module

erights:
The environment-options.js file in @endo/marshal is similar to the one in @endo/ses. The biggest difference is that the one in ses is designed to be used before lockdown, whereas the one marshal is designed to be used after lockdown. But no one does. This module is unmentioned within its package and its exports are not reexported from its package. Thus, deleting it should have no effect.

We should have something like this that can be used after lockdown. But I'd much rather just use the one in @endo/ses, and have lockdown change it so that, after lockdown, it can be used safely with the same meaning. That hypothetical change is not part of this PR.

Since this PR is only removing dead code, there's no security implication and nothing to test or document.


erights:
> We should have something like this that can be used after lockdown. But I'd much rather just use the one in @endo/ses, and have lockdown change it so that, after lockdown, it can be used safely with the same meaning. That hypothetical change is not part of this PR.

On second thought, it looked like the ses environment-options module was already fine both pre and post lockdown, once one missing `freeze` is added. So I'm now including that in this PR.

kriskowal (approved):

PR #1247, opened 2022-08-07
chore: Neutralize test fixtures

kriskowal:
Security researcher @nvk0x pointed out to us that anyone attempting to run `npm install` or `yarn install` in one of the Compartment Mapper test fixtures would be vulnerable to an attacker who had published one of its fake dependencies.  Although this would not occur during normal development of this project, to entirely deny the possibility of such an attack, I've added a preinstall script to each of these fixtures to place a guard rail and preclude such an accident. Thank you to @nvk0x for pointing out the hazard.


naugtur (approved):
There's a stronger solution, where we put all packages in an `@scope/package` under a scope we control and nest the fixtures one more level. 
But let's merge this first and think if we want to upgrade later.

PR #1249, opened 2022-08-12
fix(ses)!: remove __allowUnsafeMonkeyPatching__

erights:
Well after https://github.com/Agoric/agoric-sdk/pull/5922 removes uses of this unsafe kludgy option, this PR removes the option itself. Good riddance!


kriskowal:
@samsiegart How confident are you that we’ve solved this problem once and for all, that our UIs can run under lockdown going forward?

samsiegart (approved):

erights:
@kriskowal this is technically a breaking change, although we've already updated agoric-sdk, which is likely the only one who cares. Do we need a deprecation dance, or can I go ahead and merge this?

kriskowal:
The breaking-change version bump induced by a conventional commit message
will suffice. We do not need to do a two phase deprecation and removal
migration.

kriskowal (approved):
Please change the commit message prefix to fix(ses)!: and use the rebase merge.

kriskowal:
Oh, and update NEWS.md

PR #1250, opened 2022-08-14
WIP: passStyleOf full input validation

erights:
Will fix https://github.com/Agoric/agoric-sdk/issues/4333

### DIFF
--- a/packages/marshal/src/helpers/safe-promise.js
+++ b/packages/marshal/src/helpers/safe-promise.js
@@ -1,0 +1,62 @@
+// @ts-check
+
+/// <reference types="ses"/>
+
+import { isPromise } from '@endo/promise-kit';
+import { assertChecker } from './passStyle-helpers.js';
+
+/** @typedef {import('../types.js').Checker} Checker */
+
+const { details: X, quote: q } = assert;
+const { isFrozen, getPrototypeOf } = Object;
+const { ownKeys } = Reflect;
+
+/**
+ * Under Hardened JS a promise is "safe" if its `then` menthod can be called
+ * syncronously without giving the promise an opportunity for a
+ * reentrancy attack during that call. No matter what `pr` is
+ * `checkSafePromise(pr)` also should not give it such an opportunity.
+ *
+ * We first test whether `pr` is a promise using `isPromise`,
+ * which currently is not safe against reentrancy.
+ * TODO: Make `isPromise` safe against reentrancy.
+ *
+ * Once we know `pr` is a promise, we know `pr` is not a proxy, so
+ * pr should not be able to sense the rest of these tests.
+ *
+ * @param {unknown} pr The value to examine
+ * @param {Checker} [check]
+ * @returns {pr is Promise} Whether it is a safe promise
+ */
+const checkSafePromise = (pr, check = x => x) => {
+  let keys;
+  return (
+    check(isPromise(pr), X`${pr} - Must be a promise`) &&
+    check(isFrozen(pr), X`${pr} - Must be frozen`) &&
+    check(
+      getPrototypeOf(pr) === Promise.prototype,
+      X`${pr} - Must inherit from Promise.prototype: ${q(getPrototypeOf(pr))}`,
+    ) &&
+    check(
+      // Suppressing prettier for the following line because it wants to
+      // remove the "extra" parens around `pr`. However, these parens are
+      // required for the TypeScript case syntax. We know this case is safe
+      // because we only get here if `ifPromise(pr)` already passed.
+      // eslint-disable-next-line prettier/prettier
+      (keys = ownKeys(/** @type {Promise} pr */(pr))).length === 0,
+      X`{pr} - Must not have any own properties: ${q(keys)}`,
+    )
+  );
+};
+harden(checkSafePromise);
+
+/**
+ * Determine if the argument is a Promise.
+ *
+ * @param {unknown} pr The value to examine
+ * @returns {pr is Promise} Whether it is a promise
+ */
+export const isSafePromise = pr => checkSafePromise(pr);
+harden(isSafePromise);
+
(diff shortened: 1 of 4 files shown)


erights:
Ready for review

@mhofman Regarding https://github.com/Agoric/agoric-sdk/issues/4333, I verified that I failed to check that the tagRecord is frozen. On remotable recognition, I made no other changes, because it now seems complete. Could you verify that there is no remaining laxity in remotable recognition? If there is, This PR should not close out https://github.com/Agoric/agoric-sdk/issues/4333 . Thanks.

mhofman (commented):
I did a quick pass through, and given this is not trying to guard against attacks through `isPromise` checks, this looks fine. We'll have to review this and other promise checks once we can have a safe Promise brand check, depending on which layer that check is implemented.

packages/marshal/src/helpers/passStyle-helpers.js
```diff
@@ -110,6 +111,7 @@ harden(getTag);
  */
 export const checkTagRecord = (tagRecord, passStyle, check = x => x) => {
   return (
+    check(isFrozen(tagRecord), X`A tagRecord must be frozen: ${tagRecord}`) &&
     check(
       typeof tagRecord === 'object',
       X`A non-object cannot be a tagRecord: ${tagRecord}`,
```
gibson042, line 117: These checks feel like they're in the wrong order; I think we should only check frozen status after confirming that we've got an object. It would probably also make sense to move the `null` check up, and possibly combine it with `typeof` (e.g., ```check(typeof tagRecord === 'object' && tagRecord !== null, X`A non-object cannot be a tagRecord: ${tagRecord}`)```)

packages/marshal/src/helpers/safe-promise.js
```diff
@@ -0,0 +1,61 @@
...
+
+/** @typedef {import('../types.js').Checker} Checker */
+
+const { details: X, quote: q } = assert;
+const { isFrozen, getPrototypeOf } = Object;
+const { ownKeys } = Reflect;
+
+/**
+ * Under Hardened JS a promise is "safe" if its `then` menthod can be called
+ * syncronously without giving the promise an opportunity for a
```
gibson042, line 16:
```suggestion
 * Under Hardened JS a promise is "safe" if its `then` method can be called
 * synchronously without giving the promise an opportunity for a
```

packages/marshal/src/helpers/safe-promise.js
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @param {unknown} pr The value to examine
+ * @param {Checker} [check]
+ * @returns {pr is Promise} Whether it is a safe promise
+ */
+const checkSafePromise = (pr, check = x => x) => {
+  let keys;
+  return (
+    check(isPromise(pr), X`${pr} - Must be a promise`) &&
+    check(isFrozen(pr), X`${pr} - Must be frozen`) &&
```
gibson042, line 34: As a matter of hygiene, I'd like the frozen status check to come first. And I think our normal pattern with assert/check messages is to put the variable parts last.
```suggestion
    check(isFrozen(pr), X`A safe promise must be frozen: ${pr}`) &&
    check(isPromise(pr), X`A safe promise must be a promise: ${pr}`) &&
```

gibson042 (approved):

packages/marshal/src/helpers/passStyle-helpers.js
erights: Nice improvements, thanks. Done.

packages/marshal/src/helpers/safe-promise.js
erights: Done

packages/marshal/src/helpers/safe-promise.js
erights: Agreed. Done.

packages/marshal/src/helpers/safe-promise.js
```diff
@@ -0,0 +1,61 @@
...
+      X`${pr} - Must inherit from Promise.prototype: ${q(getPrototypeOf(pr))}`,
+    ) &&
+    check(
+      // Suppressing prettier for the following line because it wants to
+      // remove the "extra" parens around `pr`. However, these parens are
+      // required for the TypeScript case syntax. We know this case is safe
+      // because we only get here if `ifPromise(pr)` already passed.
+      // eslint-disable-next-line prettier/prettier
+      (keys = ownKeys(/** @type {Promise} pr */(pr))).length === 0,
+      X`{pr} - Must not have any own properties: ${q(keys)}`,
```
mhofman, line 46:
```suggestion
      X`${pr} - Must not have any own properties: ${q(keys)}`,
```

packages/marshal/src/helpers/safe-promise.js
```diff
@@ -0,0 +1,61 @@
...
+      getPrototypeOf(pr) === Promise.prototype,
+      X`${pr} - Must inherit from Promise.prototype: ${q(getPrototypeOf(pr))}`,
+    ) &&
+    check(
+      // Suppressing prettier for the following line because it wants to
+      // remove the "extra" parens around `pr`. However, these parens are
+      // required for the TypeScript case syntax. We know this case is safe
+      // because we only get here if `ifPromise(pr)` already passed.
+      // eslint-disable-next-line prettier/prettier
+      (keys = ownKeys(/** @type {Promise} pr */(pr))).length === 0,
```
mhofman, line 45: This check ends up too strict in the presence of the node async_hooks symbols added to some promise instances.

erights:
Sorry, @mhofman , your comments arrived after #1250 was merged. Also, you'd know better than I how to fix the safe-promise test. Would like like to send me a PR with your suggestions, that I can review? Thanks.

mhofman:
Yup, this was noticed by @kriskowal while merging into agoric-sdk. Working on a PR right now.

PR #1252, opened 2022-08-16
test: add a test to check if arguments prototype is frozen under lockdown

naugtur:


packages/ses/test/test-frozen-primordials.js
```diff
@@ -2,13 +2,20 @@ import test from 'ava';
...
+});
+test('check if override-protected primordials are frozen', t => {
   // After removing the detachedProperties mechanism and without
   // the originalValue mechanism, this test failed.
   t.truthy(Object.isFrozen(Object.prototype.toString));
 
   const desc = getOwnPropertyDescriptor(Object.prototype, 'toString');
   t.is(desc.get.originalValue, Object.prototype.toString);
 });
+test('check if arguments prototype is frozen', t => {
```
kriskowal, line 16: nit: newline between tests, please.

packages/ses/test/test-frozen-primordials.js
```diff
@@ -2,13 +2,20 @@ import test from 'ava';
...
+  (function() {
```
kriskowal, line 17: lint evidently wants a throw-away name here.

kriskowal (approved):

PR #1255, opened 2022-08-16
fix(eventual-send)!: Disallow using E proxy methods as functions

### LINKED ISSUE #1253: Consider "unbinding" `E` methods
The implementation of the `E` proxy allows to use the methods as free function as it implicitly binds them, preventing future changes that may require to revert this behavior.

https://github.com/endojs/endo/blob/fe9c78414eab5d1bce73cdb16e1455c1c4307e98/packages/eventual-send/src/E.js#L33

A similar issue arose in the [agoric-sdk for local kind representatives](https://github.com/Agoric/agoric-sdk/pull/5970) which is planned to switch to a more "class" based implementation which is incompatible with bound methods.

While this would be a breaking change to the `E` proxies as some code currently relies on this auto-binding behavior, we could add a check in the E proxy handlers to disallow this bound usage.

```js
    get(_target, p, receiver) {
      return harden({
        [p](...args) {
          assert(this === receiver);
          return HandledPromise.applyMethod(x, p, args);
        },
      }[p]);
    },
```

mhofman:
Closes #1253 

I decided to reject with an error instead of using a plain `assert`ion which would have thrown synchronously for an expected "async function". Cannot use a real `async function` of course since that would absorb the `HandledPromise` into a platform `Promise`.

@kriskowal please let me know if I got the dance of breaking change documentation right.

One this is approved, I will incorporate the changed in a patch on `agoric-sdk` along with the fixes from https://github.com/Agoric/agoric-sdk/commit/ab3b5900fd0845b9ed8081f61cadf81704891ba2

### DIFF
--- a/packages/eventual-send/NEWS.md
+++ b/packages/eventual-send/NEWS.md
@@ -1,5 +1,11 @@
 User-visible changes in eventual-send:
 
+# Next release
+
+- _BREAKING_: Disallow using E proxy methods as functions.
+  Enforces the `E(x).foo()` calling convention and disallows using as bound
+  methods. Constructs like `const foo = E(x).foo; foo()` now cause a rejection.
+
 # v0.14.2 (2022-01-25)
 
 - Eventual send now hardens arguments and results (values and errors).
--- a/packages/eventual-send/NEWS.md
+++ b/packages/eventual-send/NEWS.md
@@ -2,7 +2,7 @@ User-visible changes in eventual-send:
 
 # Next release
 
-- _BREAKING_: Disallow using E proxy methods as functions.
+- *BREAKING*: Disallow using E proxy methods as functions.
   Enforces the `E(x).foo()` calling convention and disallows using as bound
   methods. Constructs like `const foo = E(x).foo; foo()` now cause a rejection.
 
--- a/packages/eventual-send/src/E.js
+++ b/packages/eventual-send/src/E.js
@@ -3,6 +3,8 @@ import { trackTurns } from './track-turns.js';
 
 /// <reference path="index.d.ts" />
 
+const { details: X, quote: q } = assert;
+
 /** @type {ProxyHandler<any>} */
 const baseFreezableProxyHandler = {
   set(_target, _prop, _value) {
@@ -29,8 +31,23 @@ const baseFreezableProxyHandler = {
 function EProxyHandler(x, HandledPromise) {
   return harden({
     ...baseFreezableProxyHandler,
-    get(_target, p, _receiver) {
-      return harden((...args) => HandledPromise.applyMethod(x, p, args));
+    get(_target, p, receiver) {
+      return harden(
+        {
+          [p](...args) {
+            if (this !== receiver) {
+              return HandledPromise.reject(
+                assert.error(
+                  X`Unexpected receiver for "${p}" method of E(${q(x)})`,
+                ),
+              );
+            }
+
+            return HandledPromise.applyMethod(x, p, args);
+          },
+          // @ts-expect-error https://github.com/microsoft/TypeScript/issues/50319
+        }[p],
+      );
     },
     apply(_target, _thisArg, argArray = []) {
       return HandledPromise.applyFunction(x, argArray);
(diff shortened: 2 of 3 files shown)


kriskowal:
Your conventional commit message is correct. I believe the body must _also_ have a footer of the form:

```
*BREAKING CHANGE*: Lorem ipsum caveat emptor.
```

That gets rolled into the changelog when we cut releases.

packages/eventual-send/src/E.js
```diff
@@ -29,8 +31,23 @@ const baseFreezableProxyHandler = {
 function EProxyHandler(x, HandledPromise) {
   return harden({
     ...baseFreezableProxyHandler,
-    get(_target, p, _receiver) {
-      return harden((...args) => HandledPromise.applyMethod(x, p, args));
+    get(_target, p, receiver) {
+      return harden(
+        {
+          [p](...args) {
```
erights, line 37:
```suggestion
          // This function purposely uses `this` for diagnostic purposes
          // (link to explanation?)
          // In order to be `this` sensitive it is defined using concise method
          // syntax rather than as an arrow function. To ensure the function
          // is not constructable, it also avoids the `function` syntax.
          [p](...args) {
```

At some point we will bulk covert all our normal functions to arrow functions. We need to be sure that any such automatic conversion skips all functions that mention `this`.

packages/eventual-send/src/E.js
```diff
@@ -53,11 +70,16 @@ function EProxyHandler(x, HandledPromise) {
...
     ...baseFreezableProxyHandler,
-    get(_target, p, _receiver) {
-      return (...args) => {
-        HandledPromise.applyMethodSendOnly(x, p, args);
-        return undefined;
-      };
+    get(_target, p, receiver) {
+      return harden(
+        {
+          [p](...args) {
```
erights, line 76:
```suggestion
          // (somehow refer to the above comment. 
          //  Or put the point in a common place and refer to
          //  it from both places)
          [p](...args) {
```

erights (approved):
LGTM, thanks!!!!

erights:
> I will incorporate the changed in a patch on agoric-sdk along with the fixes from https://github.com/Agoric/agoric-sdk/commit/ab3b5900fd0845b9ed8081f61cadf81704891ba2

I only patch agoric-sdk with an upcoming change if agoric-sdk breaks without the patch. In these cases, it is only stricter error behavior and better diagnostics, both of which can wait for agoric-sdk to depend on this PR.

packages/eventual-send/src/E.js
```diff
@@ -53,11 +70,16 @@ function EProxyHandler(x, HandledPromise) {
...
   return harden({
     ...baseFreezableProxyHandler,
-    get(_target, p, _receiver) {
-      return (...args) => {
-        HandledPromise.applyMethodSendOnly(x, p, args);
-        return undefined;
-      };
+    get(_target, p, receiver) {
+      return harden(
+        {
```
kriskowal, line 75: Does Prettier not allow you to consolidate the matching `({` and `})` brackets and spare an indent?

packages/eventual-send/src/E.js
```diff
@@ -53,11 +70,16 @@ function EProxyHandler(x, HandledPromise) {
...
+          [p](...args) {
+            assert(this === receiver);
```
kriskowal, line 77: This assertion appears to be the heart of this change.

kriskowal (approved):

packages/eventual-send/src/E.js
mhofman: Correct

packages/eventual-send/NEWS.md
```diff
@@ -1,5 +1,11 @@
 User-visible changes in eventual-send:
 
+# Next release
+
+- _BREAKING_: Disallow using E proxy methods as functions.
+  Enforces the `E(x).foo()` calling convention and disallows using as bound
+  methods. Constructs like `const foo = E(x).foo; foo()` now cause a rejection.
```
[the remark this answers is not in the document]
mhofman: Prettier plugin in VS Code auto-corrects Markdown files.

packages/eventual-send/src/E.js
mhofman: Correct, and the corresponding one above.

mhofman (commented):
Thanks for the feedback, PTAL if the added comments clarify things enough.

packages/eventual-send/src/E.js
```diff
@@ -53,11 +84,25 @@ function EProxyHandler(x, HandledPromise) {
...
+      return harden(
+        {
+          // This function purposely checks the `this` value (see above)
+          // In order to be `this` sensitive it is defined using concise method
+          // syntax rather than as an arrow function. To ensure the function
+          // is not constructable, it also avoids the `function` syntax.
+          [p](...args) {
+            // Throw since the function returns nothing
+            assert(
+              this === receiver,
```
michaelfig, line 97:
```suggestion
            assert.equal(
              this,
              receiver,
```

michaelfig (approved):

PR #1256, opened 2022-08-16
fix(bundle-source): assert that the entrypoint exists at all

### LINKED ISSUE #1206: `bundleSource` emits a bundle when entrypoint does not exist
The following program should throw an error because bogus.js does not exist. Instead, it generates a bundle.

```js
import '@endo/init';
import bundleSource from '@endo/bundle-source';
(async () => {
  const bundle = await bundleSource(new URL('bogus.js', import.meta.url).pathname);
  console.log(bundle);
})();
```

warner:
closes #1206

### DIFF
--- a/packages/bundle-source/src/index.js
+++ b/packages/bundle-source/src/index.js
@@ -402,6 +402,7 @@ export default async function bundleSource(
   options = {},
   powers = undefined,
 ) {
+  assert(fs.existsSync(startFilename));
   if (typeof options === 'string') {
     options = { format: options };
   }
--- a/packages/bundle-source/test/test-missing-entrypoint.js
+++ b/packages/bundle-source/test/test-missing-entrypoint.js
@@ -1,0 +1,9 @@
+import { lockdown } from '@endo/lockdown';
+
+import test from 'ava';
+import bundleSource from '../src/index.js';
+
+test(`missing entrypoint`, async t => {
+  await t.throwsAsync(() => bundleSource('___missing.js', 'endoZipBase64'),
+                      { message: 'xxx' });
+});


warner:
@kriskowal this is "draft" because the patch is insufficient in at least the following ways:

* the `fs.existsSync()` call should probably be coming from `powers`, if something suitable exists therein
* `existsSync` would be a funny power to provided when `read` is the actually useful one
* that call does not understand e.g. `../demo/bigint` as meaning `../demo/bigint.js`

but I'm hoping you might be able to point me in the right direction so I can make it better. What's the right sort of way to build the assertion? Would it be easier to build in `bundle{ZipBase64,NestedEvaluateAndGetExports)`, rather than up front in `bundleSource`?

(incidentally, patching my agoric-sdk copy with this code was enough to help figure out where I had a typo in my `bundleSource()` invocation, which was my most immediate goal)

packages/bundle-source/src/index.js
```diff
@@ -402,6 +402,7 @@ export default async function bundleSource(
   options = {},
   powers = undefined,
 ) {
+  assert(fs.existsSync(startFilename));
```
mhofman, line 405: un-requested feedback: I still would prefer avoiding `Sync` APIs where possible. In this case `stat` and checking for an `ENOENT` error is the alternative

kriskowal:
I’m sure there’s a more correct way to ensure that the file exists within compartment-mapper. `exists` is one of those “considered harmful” syscalls because it isn’t proof against race conditions on a non-transactional filesystem. `existsSync` is “considered harmful” because it limits portability to the web (perhaps not a concern specifically for `bundleSource` which has a higher hill to climb before it is portable).

The mystery is how it’s possible for `compartment-mapper` to construct a bundle at all if it cannot find the entrypoint module. This is probably due to the `.catch(() => undefined)` in the loop that searches for a file to satisfy an import specifier in the archiver’s `importHook`, and probably some missing logic to construct an appropriate error if the loop falls through. But, we see that error pretty frequently in other cases, so it must be a problem specific to the entrypoint module, specific to `makeArchive`, that allows it to produce an empty zip instead of throwing.

PR #1258, opened 2022-08-18
test: Replace misleading test

gibson042:
arguments objects don't have a special prototype

Ref 5c41bdc0628d69a83a1c9378e986573444f2bb1d

### DIFF
--- a/packages/ses/test/test-frozen-primordials.js
+++ b/packages/ses/test/test-frozen-primordials.js
@@ -15,9 +15,6 @@ test('check if override-protected primordials are frozen', t => {
   t.is(desc.get.originalValue, Object.prototype.toString);
 });
 
-test('check if arguments prototype is frozen', t => {
-  (function gimmeArguments() {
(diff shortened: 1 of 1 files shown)


kriskowal (approved):

naugtur (approved):

kriskowal:
Perhaps it would be instructive to capture an assertion that the arguments
prototype is in fact null. It is frozen by virtue of Object.isFrozen(null)
=== true.

gibson042:
The prototype isn't null, it's `Object.prototype`.

kriskowal:
Pardon, as it were!

naugtur:
^ I'd say that adds grounds to my claim that we need to demonstrate arguments.__proto__ is frozen regardless of what you think it is.

naugtur:
The only thing that's special is the number of people knowing JS very well that were unsure what's the prototype of arguments.
I have this occasional itch to document surprising stuff with tests if not too time consuming. But I'm not orthodox about it ;)

PR #1260, opened 2022-08-21
fix: Start migrating encodePassable and rankOrder from agoric-sdk to endo

erights:
For our passable data types, mostly, we have a clean abstraction layering between the "marshal" layer (here in the endo repository) and the "store" layer (currently in the agoric-sdk repository). However, two modules (rankOrder and encodePassable) currently at the store layer really belong at the marshal layer.

The `encodePassable` API and marshal's `encodeToCapData` and `encodeToSmallcaps` APIs have converged to the degree that they should. They can be understood together. Before smallcaps, I originally wanted to do this in the hope that `encodePassable` would become an alternate to capData for general use, but @warner 's criticisms below are likely fatal. The remaining purpose of `encodePassable` is specialized uses for which a sort-order-preserving string encoding is useful. Nevertheless, because of the conceptual similarity and layering, it is clearer for these to live here.

See also https://github.com/Agoric/agoric-sdk/pull/4435 which we should make progress on as part of a larger project to explain these two abstraction levels and how they relate.

See also https://github.com/Agoric/agoric-sdk/pull/6260 for keeping these two in sync until the situation simplifies.

### DIFF
--- a/packages/marshal/src/encodePassable.js
+++ b/packages/marshal/src/encodePassable.js
@@ -8,14 +8,37 @@ import {
   nameForPassableSymbol,
   passableSymbolForName,
 } from './helpers/symbol.js';
-import { recordParts } from './rankOrder.js';
 import { ErrorHelper } from './helpers/error.js';
 
+/** @typedef {import('./types.js').PassStyle} PassStyle */
 /** @typedef {import('./types.js').Passable} Passable */
 /** @typedef {import('./types.js').Remotable} Remotable */
+/** @template T @typedef {import('./types.js').CopyRecord<T>} CopyRecord */
+/** @typedef {import('./types.js').RankCover} RankCover */
 
 const { details: X, quote: q } = assert;
-const { is, fromEntries } = Object;
+const { fromEntries, setPrototypeOf, is } = Object;
+const { ownKeys } = Reflect;
+
+/**
+ * @template T
+ * @param {CopyRecord<T>} record
+ * @returns {[string[], T[]]}
+ */
+export const recordParts = record => {
+  assertRecord(record);
+  // https://github.com/endojs/endo/pull/1260#discussion_r1003657244
+  // compares two ways of reverse sorting, and shows that this way
+  // is currently faster on Moddable XS while the other way,
+  // `.sort(reverseComparator)`, is faster on v8. We currently care more about
+  // XS performance, so we reverse sort this way.
+  const names = /** @type {string[]} */ (ownKeys(record)
+    .sort()
+    .reverse());
+  const vals = names.map(name => record[name]);
+  return harden([names, vals]);
+};
+harden(recordParts);
 
 export const zeroPad = (n, size) => {
   const nStr = `${n}`;
@@ -229,8 +252,8 @@ const decodeRecord = (encoded, decodePassable) => {
     keys.length === vals.length &&
     keys.every(key => typeof key === 'string')) ||
     assert.fail(X`not a valid record encoding: ${encoded}`);
-  const entries = keys.map((key, i) => [key, vals[i]]);
-  const record = harden(fromEntries(entries));
+  const mapEntries = keys.map((key, i) => [key, vals[i]]);
+  const record = harden(fromEntries(mapEntries));
   assertRecord(record, 'decoded record');
   return record;
 };
@@ -437,3 +460,49 @@ harden(makeDecodePassable);
 
 export const isEncodedRemotable = encoded => encoded.charAt(0) === 'r';
 harden(isEncodedRemotable);
+
+// /////////////////////////////////////////////////////////////////////////////
+
+/**
+ * @type {[PassStyle, RankCover][]}
+ */
+const PassStyleRankAndCover = harden([
+  /* !  */ ['error', ['!', '!~']],
+  /* (  */ ['copyRecord', ['(', '(~']],
+  /* :  */ ['tagged', [':', ':~']],
+  /* ?  */ ['promise', ['?', '?~']],
+  /* [  */ ['copyArray', ['[', '[~']],
+  /* b  */ ['boolean', ['b', 'b~']],
+  /* f  */ ['number', ['f', 'f~']],
+  /* np */ ['bigint', ['n', 'p~']],
+  /* r  */ ['remotable', ['r', 'r~']],
+  /* s  */ ['string', ['s', 't']],
+  /* v  */ ['null', ['v', 'v~']],
+  /* y  */ ['symbol', ['y', 'z']],
+  /* z  */ ['undefined', ['z', '{']],
+  /* | remotable->ordinal mapping prefix: This is not used in covers but it is
+       reserved from the same set of strings. Note that the prefix is > any
+       prefix used by any cover so that ordinal mapping keys are always outside
+       the range of valid collection entry keys. */
+]);
+
+export const PassStyleRank = fromEntries(
+  PassStyleRankAndCover.map(([passStyle, _range], i) => [passStyle, i]),
+);
+setPrototypeOf(PassStyleRank, null);
+harden(PassStyleRank);
+
+/**
+ * Associate with each passStyle a RankCover that may be an overestimate,
+ * and whose results therefore need to be filtered down. For example, because
+ * there is not a smallest or biggest bigint, bound it by `NaN` (the last place
+ * number) and `''` (the empty string, which is the first place string). Thus,
+ * a range query using this range may include these values, which would then
+ * need to be filtered out.
+ *
+ * @param {PassStyle} passStyle
+ * @returns {RankCover}
+ */
+export const getPassStyleCover = passStyle =>
+  PassStyleRankAndCover[PassStyleRank[passStyle]][1];
+harden(getPassStyleCover);
--- a/packages/marshal/src/encodePassable.js
+++ b/packages/marshal/src/encodePassable.js
@@ -1,0 +1,434 @@
+/* eslint-disable no-bitwise */
+// @ts-check
+import { getTag } from './helpers/passStyle-helpers.js';
+import { makeTagged } from './makeTagged.js';
+import { passStyleOf } from './passStyleOf.js';
+import { assertRecord } from './typeGuards.js';
+import {
+  nameForPassableSymbol,
+  passableSymbolForName,
+} from './helpers/symbol.js';
+import { recordParts } from './rankOrder.js';
+import { ErrorHelper } from './helpers/error.js';
+
+/** @typedef {import('./types.js').Passable} Passable */
+/** @typedef {import('./types.js').Remotable} Remotable */
+
+const { details: X, quote: q } = assert;
+const { is, fromEntries } = Object;
+
+export const zeroPad = (n, size) => {
+  const nStr = `${n}`;
+  assert(nStr.length <= size);
+  const str = `00000000000000000000${nStr}`;
+  const result = str.substring(str.length - size);
+  assert(result.length === size);
+  return result;
+};
+harden(zeroPad);
+
+// This is the JavaScript analog to a C union: a way to map between a float as a
+// number and the bits that represent the float as a buffer full of bytes.  Note
+// that the mutation of static state here makes this invalid Jessie code, but
+// doing it this way saves the nugatory and gratuitous allocations that would
+// happen every time you do a conversion -- and in practical terms it's safe
+// because we put the value in one side and then immediately take it out the
+// other; there is no actual state retained in the classic sense and thus no
+// re-entrancy issue.
+const asNumber = new Float64Array(1);
+const asBits = new BigUint64Array(asNumber.buffer);
+
+// JavaScript numbers are encoded by outputting the base-16
+// representation of the binary value of the underlying IEEE floating point
+// representation.  For negative values, all bits of this representation are
+// complemented prior to the base-16 conversion, while for positive values, the
+// sign bit is complemented.  This ensures both that negative values sort before
+// positive values and that negative values sort according to their negative
+// magnitude rather than their positive magnitude.  This results in an ASCII
+// encoding whose lexicographic sort order is the same as the numeric sort order
+// of the corresponding numbers.
+
+// TODO Choose the same canonical NaN encoding that cosmWasm and ewasm chose.
+const CanonicalNaNBits = 'fff8000000000000';
+
+const encodeBinary64 = n => {
+  // Normalize -0 to 0 and NaN to a canonical encoding
+  if (is(n, -0)) {
+    n = 0;
+  } else if (is(n, NaN)) {
+    return `f${CanonicalNaNBits}`;
+  }
+  asNumber[0] = n;
+  let bits = asBits[0];
+  if (n < 0) {
+    bits ^= 0xffffffffffffffffn;
+  } else {
+    bits ^= 0x8000000000000000n;
+  }
+  return `f${zeroPad(bits.toString(16), 16)}`;
+};
+
+const decodeBinary64 = encoded => {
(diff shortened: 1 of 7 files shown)


packages/marshal/src/encodePassable.js
```diff
@@ -0,0 +1,423 @@
...
+  // Normalize -0 to 0 and NaN to a canonical encoding
+  if (is(n, -0)) {
+    n = 0;
+  } else if (is(n, NaN)) {
+    return `f${CanonicalNaNBits}`;
+  }
+  asNumber[0] = n;
+  let bits = asBits[0];
+  if (n < 0) {
+    // XXX Why is the no-bitwise lint rule even a thing??
```
kriskowal, line 62: `&` is a common `&&` typo that produces very confusing effects for the majority of web developers who have never needed nor likely know about bitwise operators. I find that it’s sufficient to disable no-bitwise at the top of a file that uses bitwise math at all, since it’s the kind of thing you either don’t use at all or use everywhere in a module.

kriskowal (commented):
I’m excited to see this move up and generally excited about separating the JSON/binary representation layers. I also see a lot of similarity to Syrup in this binary encoding. Similar but not the same, of course.

I can give this a closer read when it gets higher in priority.

packages/marshal/src/encodePassable.js
erights: Done

packages/marshal/src/encodePassable.js
```diff
@@ -0,0 +1,419 @@
...
+
+// `'\u0000'` is the terminator after elements.
+// `'\u0001'` is the backslash-like escape character, for
+// escaping both of these characters.
+
+const encodeArray = (array, encodePassable) => {
+  const chars = ['['];
+  for (const element of array) {
+    const enc = encodePassable(element);
+    for (const c of enc) {
```
warner, line 179: That's a lot of iterating strings, one character at a time, repeated for every depth of the input object graph, and double that for records. (`{ foo: {bar: 'baz'} }` is going to see `foo` walked twice and `bar` walked four times. Let's be sure to benchmark this: I suspect this string walk will dominate the encoding time.

The alternative might be a length-prefix on each component string, which also avoids the exponential escape-the-escape-character penalty. Even if the caller (and the other layers of protocol they might be using) manages to avoid calling `marshal.serialize` twice on the same data, the fact that `encodeArray` needs to escape the output of `encodePassable` means that even e.g. `[ [ 1, 2 ] ]` is going to invoke the `\u0001` escape, and `[ [...Array(101)] ]` is going to have 100 `\u0001\u0000` sequences, and `[ [ [...Array(101)] ] ]` will have 200, and `[ [ [ [...Array(101)] ] ] ]` will have 400, etc. Every record in the nesting will cost double.

erights: Attn @gibson042

gibson042: I believe this could be fixed rather conveniently by moving the escaping logic from _array_ encoding/decoding to _string_ encoding/decoding and refactoring array decoding to handle depth. For example, assuming the array sigil is updated from "[" to "~" for control-character adjacency:
* Define string escapes like `const escapes = { __proto__: null, "\0": "\u{0001}0", "\x01": "\u{0001}1", "~": "\u{007F}0", "\x7F": "\u{007F}1" }` (or to squeeze out every bit of performance, replace that object with an array like `escapes = Array(128).fill().map((_, codePoint) => escapes[String.fromCodePoint(codePoint)])` or its sparse equivalent).
* Encode strings like ``` `s${passable.replaceAll(/[\0\x01~\x7F]/g, c => escapes[c])}` ```.
* Decode arrays like
  ```js
  const decodeArray = (encoded, decodePassable) => {
    encoded.startsWith('~') || assert.fail(X`invalid array encoding: ${encoded}`);
    const tail = encoded.slice(1);
    let elements = [];
    let depth = 0;
    let nextIndex = 0;
    let currentElementIndex = 0;
    for (let { 0: c, index } of tail.matchAll(/[\0~]/g)) {
      if (c === '~') {
        // This is the start of a nested array.
        depth += 1;
      } else {
        // This is an element terminator.
        if (index === nextIndex) {
          // A terminator after ~ or another terminator ends the deepest array.
          depth -= 1;
          (depth >= 0) ||
            assert.fail(X`unexpected element terminator: ${encoded.slice(0, index + 2)}`);
        }
        if (depth === 0) {
          // We have a complete top-level element.
          elements.push(decodePassable(tail.slice(currentElementIndex, index)));
          currentElementIndex = index + 1;
        }
      }
      // Advance the index.
      nextIndex = index + 1;
    }
    (depth === 0) || assert.fail(X`unterminated encoded array: ${encoded}`);
    (nextIndex === tail.length) ||
      assert.fail(X`unterminated encoded array element: ${tail.slice(currentElementIndex)}`);
    return harden(elements);
  };
  ```
  (and note that a further optimization is possible by decoding nested arrays into a cache rather than discarding their contents after syntax validation and thereby forcing repeat work for each increased level)

Sample encodings:
* `[]`: "~"
* `[[]]`: "~~\u0000"
* `'foo'`: "sfoo"
* `'\0,\x01,~,\x7F'`: "s\u00010,\u00011,\u007F0,\u007F1"
* `['\0']`: "~s\u00010\u0000"
* `[['\0']]`: "~~s\u00010\u0000\u0000"
* `['~']`: "~s\u007F0\u0000"
* `[['~']]`: "~~s\u007F0\u0000\u0000"
* `{ foo: "bar" }`: "(~~sfoo\u0000\u0000~sbar\u0000\u0000"

gibson042 (commented):
I've made it through encodePassable.js, and will come back later for rankOrder.js.

packages/marshal/src/encodePassable.js
gibson042: And we can also reduce the size of JSON representations by abandoning characters between U+0000 and U+001F (inclusive), which require "\u…" escape sequences. This would look something like changing the array element terminator from U+0000 to "&" (updating string escapes to include "&" → "%1" and "%" → "%0") and—because there is a requirement that the terminator sorts before anything else—the error sigil from "!" to "*" (assuming we don't care about the relative sort order of types w.r.t. each other).
* `[]`: "~"
* `[[]]`: "~~&"
* `'foo'`: "sfoo"
* `'\0,\x01,%,&,~,\x7F'`: "s\u0000,\u0001,%0,%1,␡0,␡1"
* `['\0']`: "~s\u0000&"
* `[['\0']]`: "~~s\u0000&&"
* `['~']`: "~s␡0&"
* `[['~']]`: "~~s␡0&&"
* `{ foo: "bar" }`: "(~~sfoo&&~sbar&&"

packages/marshal/test/test-encodePassable.js
```diff
@@ -0,0 +1,129 @@
...
+    t.is(rankComp, 0);
+  } else {
+    t.assert(rankComp === 0 || rankComp === fullComp);
+  }
+};
+
+test('order invariants', t => {
+  for (let i = 0; i < sample.length; i += 1) {
+    for (let j = i; j < sample.length; j += 1) {
+      orderInvariants(t, sample[i], sample[j]);
```
gibson042, line 107: Shouldn't this be structured like `await fc.assert(fc.property(…))` rather than using a direct list?

gibson042 (approved):
I feel most strongly about `PassStyleRankAndCover` being out-of-place in rankOrder.js (at least while it hard-codes prefix sigils from encodePassable.js), but everything appears like it will function correctly. I support merging now or fixing first, whichever is least disruptive to the relationship between endo and agoric-sdk.

packages/marshal/src/rankOrder.js
```diff
@@ -0,0 +1,545 @@
...
+  assertRecord(record);
+  // TODO Measure which is faster: a reverse sort by sorting and
+  // reversing, or by sorting with an inverse comparison function.
+  // If it makes a significant difference, use the faster one.
+  const names = ownKeys(record)
+    .sort()
+    .reverse();
+  // @ts-expect-error It thinks name might be a symbol, which it doesn't like.
+  const vals = names.map(name => record[name]);
+  return harden([names, vals]);
```
gibson042, line 168: I think we can also avoid some unnecessary traversal.
```suggestion
  return freeze([names, vals]);
```

packages/marshal/src/rankOrder.js
```diff
@@ -0,0 +1,545 @@
...
+    // Calling isRankSorted gives it a chance to get memoized for
+    // this `compare` function even if it was already memoized for a different
+    // `compare` function.
+    if (isRankSorted(passables, compare)) {
+      return passables;
+    }
+  }
+  const unsorted = [...passables];
+  unsorted.forEach(harden);
+  const sorted = harden(unsorted.sort(compare));
```
gibson042, line 364:
```suggestion
  const sorted = freeze(unsorted.sort(compare));
```

packages/marshal/src/encodePassable.js
erights: @gibson042 I decided not to make these format changes as part of this PR. Instead I filed https://github.com/endojs/endo/issues/1349 and assigned both of us to remember to do it later.

Resolving.

packages/marshal/src/encodePassable.js
```diff
@@ -0,0 +1,428 @@
...
+  assert(encoded.startsWith('f'), X`Encoded number expected: ${encoded}`);
+  let bits = BigInt(`0x${encoded.substring(1)}`);
+  if (encoded[1] < '8') {
+    bits ^= 0xffffffffffffffffn;
+  } else {
+    bits ^= 0x8000000000000000n;
+  }
+  asBits[0] = bits;
+  const result = asNumber[0];
+  assert(!is(result, -0), X`Unexpected negative zero: ${encoded}`);
```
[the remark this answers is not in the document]
erights: I fixed it. Will worry about syncing as part of https://github.com/Agoric/agoric-sdk/pull/6260 . Done here.

packages/marshal/src/encodePassable.js
```diff
@@ -0,0 +1,428 @@
...
+      nDigits
+    }:${
+      // The digits.
+      n
+    }`;
+  }
+};
+
+const decodeBigInt = encoded => {
+  const typePrefix = encoded[0];
```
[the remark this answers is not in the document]
erights: Done

packages/marshal/src/rankOrder.js
```diff
@@ -0,0 +1,545 @@
...
+ */
+
+/**
+ * This is the equality comparison used by JavaScript's Map and Set
+ * abstractions, where NaN is the same as NaN and -0 is the same as
+ * 0. Marshal serializes -0 as zero, so the semantics of our distributed
+ * object system does not distinguish 0 from -0.
+ *
+ * `sameValueZero` is the EcmaScript spec name for this equality comparison,
+ * but TODO we need a better name for the API.
```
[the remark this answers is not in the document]
erights: IMO neither is enough better to be worth deviating from the standard. I'm leaving this as is but resolving.

packages/marshal/src/rankOrder.js
```diff
@@ -0,0 +1,545 @@
...
+const comparatorMirrorImages = new WeakMap();
+
+export const recordParts = record => {
+  assertRecord(record);
+  // TODO Measure which is faster: a reverse sort by sorting and
+  // reversing, or by sorting with an inverse comparison function.
+  // If it makes a significant difference, use the faster one.
+  const names = ownKeys(record)
+    .sort()
+    .reverse();
```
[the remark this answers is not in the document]
erights: Thanks for measuring! That is a bizarre little measuring tool you've built there ;)

Since we care more about XS performance, I'm leaving the code as is, but I changed the comment to cite your analysis. Done.

packages/marshal/src/rankOrder.js
erights: I prefer that we stay in the robust `harden` habit and only do otherwise when there's a good enough reason. Performance can be a good enough reason, but only with good enough evidence that it makes a significant overall difference.

Leaving as is. Resolving.

packages/marshal/src/rankOrder.js
erights: Likewise.

packages/marshal/src/rankOrder.js
```diff
@@ -0,0 +1,545 @@
...
+  /* ?  */ ['promise', ['?', '?~']],
+  /* [  */ ['copyArray', ['[', '[~']],
+  /* b  */ ['boolean', ['b', 'b~']],
+  /* f  */ ['number', ['f', 'f~']],
+  /* np */ ['bigint', ['n', 'p~']],
+  /* r  */ ['remotable', ['r', 'r~']],
+  /* s  */ ['string', ['s', 't']],
+  /* v  */ ['null', ['v', 'v~']],
+  /* y  */ ['symbol', ['y', 'z']],
+  /* z  */ ['undefined', ['z', '{']],
```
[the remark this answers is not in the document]
erights: It's actually supposed to be an inclusive, inclusive interval, though I should double check that. (It's been awhile)

Moved to encodePassable and updated the comment. The rest not done yet, so not yet resolving.

packages/marshal/src/rankOrder.js
```diff
@@ -0,0 +1,545 @@
...
+  /* | remotable->ordinal mapping prefix: This is not used in covers but it is
+       reserved from the same set of strings. Note that the prefix is > any
+       prefix used by any cover so that ordinal mapping keys are always outside
+       the range of valid collection entry keys. */
+]);
+
+const PassStyleRank = fromEntries(
+  entries(PassStyleRankAndCover).map(([i, v]) => [v[0], Number(i)]),
```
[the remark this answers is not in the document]
erights: Doh!

Done

packages/marshal/src/rankOrder.js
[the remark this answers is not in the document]
erights: Took your refinement below. Resolving.

packages/marshal/src/rankOrder.js
```diff
@@ -0,0 +1,545 @@
...
+ * there is not a smallest or biggest bigint, bound it by `NaN` (the last place
+ * number) and `''` (the empty string, which is the first place string). Thus,
+ * a range query using this range may include these values, which would then
+ * need to be filtered out.
+ *
+ * @param {PassStyle} passStyle
+ * @returns {RankCover}
+ */
+export const getPassStyleCover = passStyle =>
+  PassStyleRankAndCover[PassStyleRank[passStyle]][1];
```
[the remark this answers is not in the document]
erights: Great suggestions! I did exactly that, copying your code and carefully reviewing it. I made no further revisions. Done!

packages/marshal/test/test-encodePassable.js
erights: At https://github.com/endojs/endo/pull/1291#pullrequestreview-1169811433 I postponed it from this PR to #1291 , since it is about use of fast-check, that I'll be thinking about harder there. Resolving.

packages/marshal/src/rankOrder.js
```diff
@@ -0,0 +1,545 @@
...
+        // that if the names of record X are a subset of the names of record Y,
+        // then record X will have an earlier rank and sort to the left of Y.
+        const [leftNames, leftValues] = recordParts(left);
+        const [rightNames, rightValues] = recordParts(right);
+
+        const result = comparator(leftNames, rightNames);
+        if (result !== 0) {
+          return result;
+        }
+        return comparator(leftValues, rightValues);
```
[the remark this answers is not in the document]
erights: Done, with just a pinch more abstraction. Good suggestion, thanks.

PR #1261, opened 2022-08-22
WIP: make many multiline asserts into ifs

erights:
Same as https://github.com/Agoric/agoric-sdk/pull/6019 , but for Endo

When tracing code in the debugger, I'm always surprised about how many steps it takes to walk through an assert is I single step all the way. For multiline `assert(` calls, the `if` form with an `assert.fail(` is about the same number of lines, sometimes a bit more, sometimes a bit less. (Altogether, this PR is a net increase in lines of code, though https://github.com/Agoric/agoric-sdk/pull/6019 is a net reduction.) It is *less* readable than the simple `assert(` because of the inversion of the condition. But it does skip all the 
```js
X`...${q(expr)}...`
```
in the success case, which is the only case where we might care about performance. Would it make a significant difference? A difference big enough to pay the cost of reduced readability? We won't know until we measure. Hence this WIP draft PR.

### DIFF
--- a/packages/captp/src/atomics.js
+++ b/packages/captp/src/atomics.js
@@ -23,10 +23,11 @@ const STATUS_FLAG_REJECT = 4;
  * @param {SharedArrayBuffer} transferBuffer the backing buffer
  */
 const splitTransferBuffer = transferBuffer => {
-  assert(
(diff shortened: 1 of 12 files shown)


erights:
Problem solved instead with `||`. Closing as obsolete.

PR #1262, opened 2022-08-22
feat(compartment-mapper): Add policy enforcement

naugtur:
**notes:**
- policy is built into `compartmentMap` in graph translation. Compartment descriptors for items not covered by policy are not preserved.
- package location abbreviations are used to differentiate packages that might otherwise have the same (or confusingly similar) names. If a malicious package declares a bundled dependency called `request` it won't steal the policy to access network from the real package
- app uses policy to manage what its dependencies can have access to and limit supply chain risks; user endows the app with globals and can still decide how much to give it, but doesn't decide the access internal bits of the app get. 
- if an endowment granted by policy is missing, it's simply undefined. It seems like it'd make sense to throw an error about a missing endowments, but in real life globals are often used to detect platform capability and so need to be allowed by policy even if they may not exist.

**requirements:**
- [x] enforce policy for builtins (exit modules)
- [x] enforce policy for importing across package boundaries
- [x] enforce policy for top-level globals
- [x] separate attenuation compartment
- [x] deny policy bypass with aliases
- [x] expand tests of policy enforcement


packages/compartment-mapper/src/policy.js
```diff
@@ -0,0 +1,64 @@
+// @ts-check
+
+const { keys, create, entries } = Object;
+
+const { has } = Reflect;
+
+const copyGlobals = (globals, list) => {
```
naugtur, line 7: To be replaced by https://github.com/LavaMoat/LavaMoat/blob/main/packages/core/src/makeGetEndowmentsForConfig.js

packages/compartment-mapper/test/test-policy.js
```diff
@@ -0,0 +1,98 @@
...
+
+const fixture = new URL(
+  'fixtures-policy/node_modules/app/index.js',
+  import.meta.url,
+).toString();
+
+const globals = {
+  redPill: 42,
+  bluePill: 2,
+  yellowPill: 3,
```
naugtur, line 19: Yes, yellow pill is worse than blue pill

packages/compartment-mapper/src/policy.js
naugtur: `__isKnownScopeProxy__` should no longer be an issue.

packages/compartment-mapper/src/compartment-map.js
```diff
@@ -356,6 +356,28 @@ const assertTypes = (allegedTypes, path, url) => {
...
+  path,
+  url = '<unknown-compartment-map.json>',
+) => {
+  const policy = Object(allegedPolicy);
+  assert(
+    allegedPolicy === undefined ||
+      (allegedPolicy === policy && !Array.isArray(policy)),
+    `${path}.policy must be undefined or an object, got ${allegedPolicy} in ${q(
+      url,
+    )}`,
```
naugtur, line 377: Further assertions on policy shape, while possible, might not make a lot of sense as in the future we might decouple policy implementation from hooks and make policy pluggable. That means a higher level, more approachable policy text could be possible. Stating things like "network access" or "local network access" or "same origin network access" etc.

packages/compartment-mapper/src/import.js
```diff
@@ -80,6 +87,7 @@ export const loadLocation = async (readPowers, moduleLocation, options) => {
       parserForLanguage,
       globals,
       modules,
+      attenuations,
```
naugtur, line 90: Should we call them attenuations or attenuators? My initial choice may have been wrong.

packages/compartment-mapper/src/link.js
```diff
@@ -14,6 +14,11 @@
 
 import { resolve } from './node-module-specifier.js';
 import { parseExtension } from './extension.js';
+import {
+  getAllowedGlobals,
+  gatekeepModuleAccess,
+  attenuateModuleHook,
+} from './policy.js';
```
naugtur, line 21: It is possible that I did not identify all places where these functions need to be called in this file.   
I plan to address that with more tests.   
Hints are welcome though.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -228,6 +285,8 @@ const inferParsers = (descriptor, location) => {
  * @param {Object} packageDetails.packageDescriptor
  * @param {Set<string>} tags
  * @param {boolean} dev
+ * @param {Array<string>} logicalPath
+ * @param {Map<string, Array<string>>} preferredPackageLogicalPathMap
  * @returns {Promise<undefined>}
  */
 const graphPackage = async (
```
naugtur, line 292: Changes to graphPackage are temporary.   
I've ported the algorithm for identifying dependencies from lavamoat/aa for comparison.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -609,6 +704,7 @@ const translateGraph = (
       scopes,
       parsers,
       types,
+      policy: packagePolicy,
```
naugtur, line 707: policy specified upon loading is introduced into compartments piece by piece. Thus getting successfully introduced in archives too.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -125,6 +126,9 @@ export const makeImportHookMaker = (
       // The `moduleMapHook` captures all third-party dependencies.
       if (moduleSpecifier !== '.' && !moduleSpecifier.startsWith('./')) {
         if (has(exitModules, moduleSpecifier)) {
+          gatekeepModuleAccess(moduleSpecifier, compartmentDescriptor.policy, {
+            exit: true,
+          });
```
naugtur, line 131: I'm not sure about this. I don't see how execution would get to here. @kriskowal ?

naugtur:
Example differences between compartment-mapper's graph and @lavamoat/aa

legend:
```
descriptor.path
descriptor.logicalPath (implemented within the single pass of graph
result from lavamoat/aa
location (key in graph)
descriptor.label
^ explanation
```
results"
```
abc
browser-abc
browser-abc
file:///home/naugtur/repo/endo/packages/compartment-mapper/test/fixtures-resolve/node_modules/browser-abc/
browser-abc-v1.0.0
^ browser field treated as source of truth or not

@endo/cjs-module-analyzer
@endo/cjs-module-analyzer
undefined
file:///home/naugtur/repo/endo/packages/cjs-module-analyzer/
@endo/cjs-module-analyzer-v0.2.28
^ going up from root

@endo/static-module-record
@endo/static-module-record
undefined
file:///home/naugtur/repo/endo/packages/static-module-record/
@endo/static-module-record-v0.7.15
^ going up from root

@endo/zip
@endo/zip
undefined
file:///home/naugtur/repo/endo/packages/zip/
@endo/zip-v0.2.28
^ going up from root

mallory>dan
alice>hackity
alice>hackity
file:///home/naugtur/repo/endo/packages/compartment-mapper/test/fixtures-policy/node_modules/hackity/
hackity-v1.0.0
^ browser field treated as source of truth or not

qwe
browser-qwe
browser-qwe
file:///home/naugtur/repo/endo/packages/compartment-mapper/test/fixtures-resolve/node_modules/browser-qwe/
browser-qwe-v1.0.0
^ browser field treated as source of truth or not

ses
ses
undefined
file:///home/naugtur/repo/endo/packages/ses/
ses-v0.18.0
^ going up from root

symlink>symlink-peer
symlink>symlink-peer
undefined
file:///home/naugtur/repo/endo/packages/compartment-mapper/test/fixtures-symlink/deps/node_modules/symlink-peer/
symlink-peer
^ symlink support

symlink
symlink
undefined
file:///home/naugtur/repo/endo/packages/compartment-mapper/test/fixtures-symlink/deps/node_modules/symlink/
symlink
^ symlink support
```

packages/compartment-mapper/src/link.js
```diff
@@ -231,30 +239,54 @@ const makeModuleMapHook = (
...
+            attenuators,
+          );
         }
       }
       if (foreignModuleSpecifier !== undefined) {
+        if (!moduleSpecifier.startsWith('./')) {
+          // archive goes through foreignModuleSpecifier for local modules too
+          gatekeepModuleAccess(moduleSpecifier, compartmentDescriptor, {
+            exit: false,
+          });
```
naugtur, line 255: This check should be redundant.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -531,33 +558,6 @@ const graphPackages = async (
...
- * The shortest path serves as a suitable sort key for generating archives that
- * are consistent even when the package layout on disk changes, as the package
- * layout tends to differ between installation with and without devopment-time
- * dependencies.
- *
- * @param {Graph} graph
- * @param {string} location
- * @param {Array<string>} path
- */
-const trace = (graph, location, path) => {
```
naugtur, line 550: `path` is now computed in the same pass as graph building. 
Added benefit is: it uses actual package names and not the overrides (from browser field in package.json)

packages/compartment-mapper/src/policy.js
```diff
@@ -0,0 +1,304 @@
...
+};
+
+/**
+ * Generates an ID for the package for policy purposes.
+ * Some arguments here are for future proofing - I could imagine using them for IDs instead
+ *
+ * @param {PolicyIdentityComponents} identityComponents
+ * @returns {string}
+ */
+const generatePolicyId = ({ isEntry = false, name, path }) => {
```
naugtur, line 56: I'm considering adding location here as bikeshedding for hupothetical future use

packages/compartment-mapper/test-manual/policy/app.js
```diff
@@ -0,0 +1,18 @@
+/* global require Buffer module */
+const Poet = require('entropoetry');
```
naugtur, line 2: Leaving the manual test here as illustration for the review and to make it easier to play around with an example.

Not sure if it's welcome to stay in the repository, but I'm ok removing it.

packages/compartment-mapper/test-manual/policy/lavamoat/node/policy.json
```diff
@@ -0,0 +1,38 @@
+{
```
naugtur, line 1: provided for comparison

packages/compartment-mapper/test/snapshots/test-policy.js.md
```diff
@@ -0,0 +1,113 @@
+# Snapshot report for `test/test-policy.js`
```
naugtur, line 1: Informative errors for policy problems are still possible despite it's being applied as early as building the compartment map

kumavis (commented):
very nice.
lavamoat compatibility:
- we call `<root>` as `$root$` because we use `>` as a separator.
- LM `$root$` does not have a policy and can have everything. goal is good DX
- in LM endowment globals come from the `$root$` compartment instead of the start compartment
- in LM we transform function endowments to be called with the start compartment globalThis instead of their compartment's globalThis

naugtur:
Responding to points from @kumavis below. I'll add numbers for ease of reference.

#### K1
> * we call `<root>` as `$root$` because we use `>` as a separator.

Initially I wanted to keep it consistent with `<ATTENUATORS_COMPARTMENT>` and that one was using `<>` because they are not allowed in file locations. Now that I've eliminated the need to surface  `<ATTENUATORS_COMPARTMENT>` in the policy, this is no longer needed.  
We could have something even more familiar as the root key in policy - **maybe we should call it `"."` by convention from package.json->exports?**

#### K2
> * LM `$root$` does not have a policy and can have everything. goal is good DX

When LM builds on top of compartment-mapper I believe we'd maintain that top compartment gets it all. We could generate it from the endowments we prepare or add a feature to specify "allow all" in the policy.  
For the wider Endo usecase I believe this is valuable, because **it allows an app/archive to come with a specification of what it needs**. In the grand scheme of things this would give us a better understanding of what the app wants to get and open up ways to improve UX for the person granting those permissions.

#### K3
> * in LM endowment globals come from the `$root$` compartment instead of the start compartment

I'm not sure what the consequences of this difference would be. Compartment-mapper requires explicit endowments, but if we endow all globals to the app I think it'd be equivalent to how LM currently works.

#### K4
> * in LM we transform function endowments to be called with the start compartment globalThis instead of their compartment's globalThis

Yes, this would have to be added on. My assumption was I'd implement attenuators support for globals and allow an attenuator to decide what to do. Unless there's **reasons to rebind all functions by default?**

packages/compartment-mapper/src/node-modules.js
```diff
@@ -699,13 +729,23 @@ export const compartmentMapForNodeModules = async (
     commonDependencies,
   );
 
-  trace(graph, packageLocation, []);
+  if (policy) {
+    // Instead of worrying if ATTENUATORS_COMPARTMENT is unique enough, we're overwriting whatever tried to use it anyway
```
michaelfig, line 733: Would it be worthwhile to add an assertion that `graph[ATTENUATORS_COMPARTMENT]` is not used?  Or do you really want to silently overwrite?

packages/compartment-mapper/src/policy.js
```diff
@@ -0,0 +1,305 @@
...
+
+const { create, entries, assign, keys, freeze } = Object;
+const { stringify } = JSON;
+
+const copyGlobals = (globals, list) => {
+  const g = create(null);
+  if (list && list.length > 0) {
+    for (let k = 0; k < list.length; k += 1) {
+      const key = list[k];
+      // If an endowment is missing, global value is undeinfed.
```
michaelfig, line 13:
```suggestion
      // If an endowment is missing, global value is undefined.
```

packages/compartment-mapper/src/policy.js
```diff
@@ -0,0 +1,305 @@
...
+};
+
+function attenuateModule({ attenuators, name, params, originalModule }) {
+  if (!attenuators) {
+    // TODO: figure out what to pass here or where to move this check
+    throw Error(
+      `Attenuation '${name}' in policy doesn't have a corresponding implementation.`,
+    );
+  }
+  const attenuationCompartment = new Compartment(
```
michaelfig, line 201: Is it the case that attenuation is incompatible with the speak-no-eval bundling @kumavis is working on?  If so, would that be worth a comment in one of the two PRs?

packages/compartment-mapper/src/policy.js
```diff
@@ -0,0 +1,305 @@
...
+  compartmentDescriptor,
+  foreignModuleSpecifier,
+  foreignCompartmentName,
+}) => {
+  const { policy, name, scopes } = compartmentDescriptor;
+
+  if (policy) {
+    if (!policy.packages) {
+      return padDiagnosis(
+        `There were no allowed packages specified in policy for "${name}"`,
```
michaelfig, line 281: Please use a `q` function for robust quoting of values here and below.  It can be as simple as `const q = JSON.stringify;` if the normal `const { quote: q } = assert;` convention doesn't work.

```suggestion
        `There were no allowed packages specified in policy for ${q(name)}`,
```

packages/compartment-mapper/src/types.js
```diff
@@ -60,6 +61,7 @@ export {};
  * @typedef {Object} ModuleDescriptor
  * @property {string=} [compartment]
  * @property {string} [module]
+ * @property {string} [policyId]
```
michaelfig, line 64: It seems this property isn't used anymore.  If so, please delete.

```suggestion
```

packages/compartment-mapper/test-manual/policy/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "endo-sample",
```
michaelfig, line 2: We try to use `@endo/...` scoping so that those who are watching don't get nervous about NPM hijacking.

```suggestion
  "name": "@endo/compartment-mapper-demo-policy",
```

packages/compartment-mapper/test-manual/policy/app.js
michaelfig: I'm happy with leaving it here, but would you consider moving it to a `demo` subdirectory instead of `test-manual`?

packages/compartment-mapper/test-manual/policy/att1/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "att1",
```
michaelfig, line 2:
```suggestion
  "name": "@endo/compartment-mapper-demo-policy-att1",
```

packages/compartment-mapper/test/fixtures-policy/node_modules/eve/README.md
```diff
@@ -0,0 +1 @@
+this package is called eve not mallory, because it's attempting to ensure that if identifying packages for the sake of policy enfocement is not tricked by the override `"dan": "hackity"` it specifies. Just in case we don't get a false negative from the test, both 'eve' and 'dan' are shorter names than the alternative path that's the correct one: `alice>hackity`
```
michaelfig, line 1: "that if identifying packages" is a little unclear to me.  Can you please rephrase?

michaelfig (commented):
It's looking good to me.  I won't yet explicitly approve until I'm more confident I understand things correctly, but that'll happen soon enough.

packages/compartment-mapper/src/types.js
naugtur: thanks for spotting this! leftover from uglier implementation

packages/compartment-mapper/src/policy.js
```diff
@@ -0,0 +1,305 @@
...
+const { stringify } = JSON;
+
+const copyGlobals = (globals, list) => {
+  const g = create(null);
+  if (list && list.length > 0) {
+    for (let k = 0; k < list.length; k += 1) {
+      const key = list[k];
+      // If an endowment is missing, global value is undeinfed.
+      // This is an expected behavior if globals are used for platform feature detection
+      g[key] = globals[key];
```
naugtur, line 15: needs to bind. here or by the compartment-mapper caller?

packages/compartment-mapper/src/node-modules.js
naugtur: There's no way for a valid package to happen to have a `location` starting with `<` instead of `file://` so I assumed silent failure would give away less details to an attacker. 

Throwing would be fine too. Do you think we should?

packages/compartment-mapper/src/policy.js
naugtur: There's no going from source text to execution involved here. If a mock StaticModuleRecord-shaped output is returned from importHook would `eval` be called anywhere?

packages/compartment-mapper/src/node-modules.js
kriskowal: I think we should throw. It’s easier than reasoning through the ways an attacker can exploit silent failure and manipulate the attenuator interface.

packages/compartment-mapper/src/policy.js
kriskowal: Consensus from the Endo call today is that we can defer the responsibility to bind endowments to the user of the compartment mapper.

packages/compartment-mapper/src/link.js
```diff
@@ -391,13 +441,21 @@ export const link = (
...
       archiveOnly,
     );
     const resolveHook = resolve;
     resolvers[compartmentName] = resolve;
 
-    // TODO also thread powers selectively.
-    const compartment = new Compartment(globals, undefined, {
+    let personalGlobals = Object.create(null);
+    if (!archiveOnly) {
+      personalGlobals = getAllowedGlobals(
```
kriskowal, line 452: If the person in question is the compartment, consider naming this `compartmentGlobals`.

packages/compartment-mapper/src/policy.js
```diff
@@ -0,0 +1,305 @@
+// @ts-check
+
+/** @typedef {import('./types.js').PolicyIdentityComponents} PolicyIdentityComponents */
+
+const { create, entries, assign, keys, freeze } = Object;
+const { stringify } = JSON;
+
+const copyGlobals = (globals, list) => {
+  const g = create(null);
```
kriskowal, line 9: Let’s not lose moral ground over mathematicians and instead use descriptive variable names.

packages/compartment-mapper/src/policy.js
```diff
@@ -0,0 +1,305 @@
...
+ * Goes through policy and lists all attenuator specifiers used.
+ * Memoization keyed on policy object reference
+ *
+ * @param {*} policy
+ * @returns {Array<string>} attenuators
+ */
+export const detectAttenuators = policy => {
+  if (!attenuatorsCache.has(policy)) {
+    const attenuators = [];
+    // a free recursive visitor implementation for simple objects
```
kriskowal, line 37: 😐 Mixed feelings.

packages/compartment-mapper/src/policy.js
```diff
@@ -0,0 +1,305 @@
...
+
+/**
+ * Generates a string identifying a package for policy lookup purposes.
+ *
+ * @param {PolicyIdentityComponents} identityComponents
+ * @returns {string}
+ */
+const generateCanonicalName = ({ isEntry = false, name, path }) => {
+  if (isEntry) {
+    return '<root>';
```
kriskowal, line 57: May as well consolidate magic string like the attenuators.

packages/compartment-mapper/src/policy.js
```diff
@@ -0,0 +1,305 @@
...
+
+/**
+ * Returns the policy applicable to the canonicalName - either by taking from user
+ * supplied policy or returning localPolicy if user didn't specify one at runtime.
+ *
+ * @param {PolicyIdentityComponents} identityComponents - a key in the policy resources spec
+ * @param {Object|undefined} policy - user supplied policy
+ * @returns {Object|undefined} policy fragment if policy was specified
+ */
+export const getPolicyFor = (identityComponents, policy) => {
```
kriskowal, line 108: Consider something like:
```suggestion
export const policyForCanonicalName = (identityComponents, policy) => {
```

packages/compartment-mapper/src/policy.js
```diff
@@ -0,0 +1,305 @@
...
+    );
+    return undefined;
+  }
+};
+
+const getGlobalsList = myPolicy => {
+  if (!myPolicy.globals) {
+    return [];
+  }
+  // TODO: handle 'write' policy
```
kriskowal, line 136: If this TODO survives to the end of the review, please link a tracking issue.

packages/compartment-mapper/src/policy.js
```diff
@@ -0,0 +1,305 @@
...
+};
+
+/**
+ * Throws if importing of the specifier is not allowed by the policy
+ *
+ * @param {string} specifier
+ * @param {Object} compartmentDescriptor
+ * @param {Object} [info]
+ */
+export const gatekeepModuleAccess = (
```
kriskowal, line 164:
```suggestion
export const assertModulePolicy = (
```

packages/compartment-mapper/src/policy.js
```diff
@@ -0,0 +1,305 @@
...
+    params: policy.builtins[specifier].params,
+    originalModule,
+  });
+};
+
+const padDiagnosis = text => ` (${text})`;
+/**
+ * Provide dignostic information for a missing compartment error
+ *
+ * @param {Object}  params
```
kriskowal, line 263: I don’t know why we are heading in this stylistic direction, but spelling `object` in the lesser case is the prevailing trend. I have favored `args` and `opts` depending on whether this is a bag of required or optional values.
```suggestion
 * @param {object}  args
```

packages/compartment-mapper/src/policy.js
```diff
@@ -0,0 +1,305 @@
...
+    }
+    if (name === ATTENUATORS_COMPARTMENT) {
+      return padDiagnosis(
+        `Attenuator "${moduleSpecifier}" was imported but there is no policy resources entry defined for it.`,
+      );
+    }
+
+    const scopeNames = entries(scopes)
+      .filter(([_name, scope]) => scope.compartment === foreignCompartmentName)
+      .map(([n]) => n);
```
kriskowal, line 292:
```suggestion
      .map(([name]) => name);
```

packages/compartment-mapper/test-manual/policy/package.json
kriskowal: And/or a `scripts.preinstall` that dies loudly.

packages/compartment-mapper/test/fixtures-policy/node_modules/eve/README.md
kriskowal, line 1: Nits
```suggestion
This package is called `eve` not `mallory`, because it's attempting to ensure that if identifying packages for the sake of policy enfocement is not tricked by the override `"dan": "hackity"` it specifies. Just in case we don't get a false negative from the test, both `eve` and `dan` are shorter names than the alternative path that's the correct one: `alice>hackity`
```

packages/compartment-mapper/test/fixtures-policy/node_modules/myattenuator/index.js
```diff
@@ -0,0 +1,7 @@
+export const attenuate = (params, originalModuleNamespace) => {
+  const ns = params.reduce((acc, k) => {
+    acc[k] = originalModuleNamespace[k];
+    return acc;
+  }, {});
+  return ns;
```
kriskowal, line 6: Consider freezing.

packages/compartment-mapper/test/fixtures-policy/node_modules/myattenuator/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "myattenuator",
+  "version": "1.0.0",
+  "main": "./index.js",
+  "type": "module",
+  "dependencies": {
```
kriskowal, line 6: Suspect lint-fix would collapse the {}

packages/compartment-mapper/test/scaffold.js
```diff
@@ -17,6 +17,25 @@ import { makeReadPowers } from '../src/node-powers.js';
 
 export const readPowers = makeReadPowers({ fs, crypto, url });
 
+const CompartmentInstrumentation = (() => {
```
kriskowal, line 20:
```suggestion
const compartmentInstrumentation = (() => {
```

kriskowal: Ava runs tests concurrently. This should be a maker so tests do not have shared state.

packages/compartment-mapper/src/policy.js
```diff
@@ -0,0 +1,305 @@
...
+    if (scopeNames.length === 1 && scopeNames[0] === moduleSpecifier) {
+      return padDiagnosis(
+        `Package "${moduleSpecifier}" is missing. Are you sure there is an entry in policy resources specified for it?`,
+      );
+    } else {
+      return padDiagnosis(
+        `Package "${moduleSpecifier}" resolves to "${foreignModuleSpecifier}" in "${foreignCompartmentName}" which seems disallowed by policy. There is likely an override defined that causes another package to be imported as "${moduleSpecifier}".`,
```
kriskowal, line 299: Also `q` here.

packages/compartment-mapper/src/compartment-map.js
```diff
@@ -30,10 +30,20 @@ const cumulativeLength = (length, term) => {
...
+  // Undefined is not preferred
+  if (a === undefined && b === undefined) {
+    return 0;
+  }
+  if (a === undefined) {
+    return 1;
+  }
+  if (b === undefined) {
+    return -1;
+  }
```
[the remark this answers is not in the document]
naugtur: I love it. A bit too clever for production use where I grew up so I'm hesitating ;)

packages/compartment-mapper/src/import-hook.js
naugtur: There are no tests covering it. 
How about we keep it as is for now and I'll create a new PR replacing this with an error that this should never happen?

packages/compartment-mapper/src/link.js
```diff
@@ -216,10 +225,9 @@ const makeModuleMapHook = (
         exit,
       } = moduleDescriptor;
       if (exit !== undefined) {
-        // TODO Currenly, every package can connect to built-in modules.
-        // Policies should be able to allow third-party modules to exit to
-        // built-ins explicitly, or have built-ins subverted by modules from
-        // specific compartments.
+        gatekeepModuleAccess(moduleSpecifier, compartmentDescriptor, {
```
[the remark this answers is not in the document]
naugtur: If I say "gatekeep" is easier to find with grep, does it make you like it more? :) 

assertModulePolicy sounds good.

packages/compartment-mapper/src/policy.js
naugtur: That's the last one I missed. Noticed it in the meeting yesterday.

packages/compartment-mapper/src/policy.js
naugtur: canonicalName is internal to policy.js - I'll try to come up with something.

packages/compartment-mapper/test/fixtures-policy/node_modules/myattenuator/package.json
naugtur: unfortunately, it refuses to go past a `node_modules` folder. I think I had a workaround for that.

packages/compartment-mapper/test/fixtures-policy/node_modules/myattenuator/index.js
naugtur: Ok, I'm freezing. It's not necessary though, because it's not the final namespace - it geets Object.assign-ed to exportsProxy in execute.

packages/compartment-mapper/src/policy.js
naugtur: Typescript enthusiasts told me to use Object not object. I no longer remember why tho. Cargo culting it ever since.
It's also 3x more popular in this codebase. We can make it consistent in a separate PR

packages/compartment-mapper/src/policy.js
naugtur: renamed to getPolicyForPackage

packages/compartment-mapper/src/policy.js
naugtur: I like using it because it's terse and unlikely to contain bugs. Can revert to regular recursive function.

packages/compartment-mapper/src/node-modules.js
naugtur: I made it throw.

packages/compartment-mapper/test/fixtures-policy/node_modules/eve/README.md
naugtur: I think I fixed it.

naugtur:
I ran linter on fixtures for the first time after a while apparently and some rules changed. Gonna have to separate it out.

naugtur:
the controversy around what to name the entry compartment in policy has been solved by removing the need for it to have a canonicalName that doesn't accidentally overlap with something else. Now all we can debate is what to name the top level field where policy for entry is defined.
Compatibility with LavaMoat is further expanded if policy.entry becomes optional or optionally accepts a value that allows everything.

michaelfig (approved):
LGTM!  Feel free to wait on other approvals, as they may have more critique.

packages/compartment-mapper/src/policy.js
```diff
@@ -0,0 +1,328 @@
...
+    collectAttenuators(attenuators, policy);
+    attenuatorsCache.set(policy, attenuators);
+  }
+  return attenuatorsCache.get(policy);
+};
+
+/**
+ * Generates a string identifying a package for policy lookup purposes.
+ *
+ * @param {PolicyIdentityComponents} identityComponents
```
naugtur, line 60: rename it to something else - naming kit?

PR #1265, opened 2022-08-25
feat(eventual-send): De-risk track-turns

### LINKED ISSUE #1245: track-turns / HandledPromise seems to retain function arguments longer than expected
https://github.com/Agoric/agoric-sdk/pull/5892/files?diff=split&w=1#r937349609 noticed that disabling `trackTurns` (https://github.com/Agoric/agoric-sdk/issues/5886) caused a GC-sensitive SwingSet unit test to fail, because an object (a Presence delivered into a vat method) was retained too long when `trackTurns` was enabled. By disabling `trackTurns`, the object is released earlier, which makes a GC syscall get emitted at a different time.

I believe this is a consequence of "overeager lexical scope retention" (I'm sure there's a more common name), where a JS engine doesn't perform a fine-grained analysis of which variables are being closed over by the definition of a new function, and instead just holds on to all of them. In this case, invocation of a HandledPromise method could cause the arguments to that invocation to be held until the result promise resolves, maybe.

I'm experimenting with hoisting some of the internal functions of `track-turns.js` to become top-level functions, so that they don't have `funcs` or `args` in their lexical scope. This appears to cause the same behavior changes as disabling `trackTurns` entirely did.

My plan is to file a PR that hoists both, to minimize the potential for accidental argument retention.

cc @erights

kriskowal:
This change includes work from myself and @warner.

1. Reduces the closure retention of functions wrapped for tracking event-loop turns.
2. Supports a DEBUG=track-turns flag that enables verbose logging, previously only available by editing track-turns.js
3. Supports a TRACK_TURNS=enabled flag that enables the feature. The intention is to reverse the flag’s polarity when we are confident that the feature can be used in production without retaining memory indefinitely.

Fixes #1245

### DIFF
--- a/packages/eventual-send/src/track-turns.js
+++ b/packages/eventual-send/src/track-turns.js
@@ -18,6 +18,49 @@ let hiddenCurrentEvent = 0;
 // Turn on if you seem to be losing error logging at the top of the event loop
 const VERBOSE = false;
 
+// We hoist these functions out of trackTurns() to discourage the
+// closures from holding onto 'args' or 'func' longer than necessary,
+// which we've seen cause HandledPromise arguments to be retained for
+// a surprisingly long time.
+
+const addRejectionNote = detailsNote => reason => {
+  if (reason instanceof Error) {
+    assert.note(reason, detailsNote);
+  }
+  if (VERBOSE) {
+    console.log('REJECTED at top of event loop', reason);
+  }
+};
+
+const wrapFunction = (func, sendingError, X) => (...args) => {
+  hiddenPriorError = sendingError;
+  hiddenCurrentTurn += 1;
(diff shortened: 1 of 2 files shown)


packages/eventual-send/src/track-turns.js
```diff
@@ -15,8 +15,58 @@ let hiddenPriorError;
 let hiddenCurrentTurn = 0;
 let hiddenCurrentEvent = 0;
 
+const env = globalThis?.process?.env;
```
erights, line 18: Please add a TODO to use environment-options.js instead, once it has been moved to its own package.

erights (approved):
LGTM, thanks! It is exactly the kind of environment variable control I was hoping for.

PR #1269, opened 2022-08-26
make README links navigable locally

turadg:
I encountered some broken links to packages that had moved from agoric-sdk. While updating them I made the links relative and include README.md so they'd be be navigable locally in the IDE. I went ahead and did that for other links I could find to READMEs in the repo.


kriskowal (approved):

kriskowal:
Thank you!

PR #1270, opened 2022-08-26
fix(eventual-send): Remove ?. and ?? operators for RESM

kriskowal:
This allows eventual-send to be used in -r esm style Agoric deploy scripts.

### DIFF
--- a/packages/eventual-send/src/track-turns.js
+++ b/packages/eventual-send/src/track-turns.js
@@ -17,10 +17,10 @@ let hiddenCurrentEvent = 0;
 
 // TODO Use environment-options.js currently in ses/src after factoring it out
 // to a new package.
-const env = globalThis?.process?.env;
+const env = (globalThis.process || {}).env || {};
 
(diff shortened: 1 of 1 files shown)


packages/eventual-send/src/track-turns.js
```diff
@@ -17,10 +17,10 @@ let hiddenCurrentEvent = 0;
 
 // TODO Use environment-options.js currently in ses/src after factoring it out
 // to a new package.
-const env = globalThis?.process?.env;
+const env = ((globalThis.process || {}).env ||{});
```
mhofman, line 20: Is this linting?

kriskowal: No, pushed update.

mhofman: Still seeing the same, bad push?

kriskowal: Yes, addressed.

mhofman (approved):
Simple enough

PR #1272, opened 2022-08-26
fix(eventual-send): Remove lingering ?. operator

kriskowal:
I missed one.


mhofman (approved):

mhofman:
Wondering if we should have a test for preventing recurrences.

PR #1275, opened 2022-08-28
fix: tolerate more from async_hooks

erights:
Our attempt to safely tolerate Node async_hooks was safe, but missed a case we could handle safely and which is provoked at least by the vscode debugger.

Node's async_hooks currently contains the following code, which we can also safely tolerate

```js
function destroyTracking(promise, parent) {
  trackPromise(promise, parent);
  const asyncId = promise[async_id_symbol];
  const destroyed = { destroyed: false };
  promise[destroyedSymbol] = destroyed;
  registerDestroyHook(promise, asyncId, destroyed);
}
```

However, before this PR we did not know to expect this. I just encountered this failure when trying to debug some of our code in the vscode debugger. The additional record that Node adds here is one we can tolerate safely. This PR's expectations includes this additional record, so that we can safely tolerate it in environments, like the vscode debugger, that provoke it.

Reviews: How do I test this? I have already tried it experimentally on the debugger experience that was blocking me. This code got me past that problem.

### DIFF
--- a/packages/marshal/src/helpers/safe-promise.js
+++ b/packages/marshal/src/helpers/safe-promise.js
@@ -23,25 +23,52 @@ const checkPromiseOwnKeys = (pr, check) => {
     key => typeof key !== 'symbol' || !hasOwnPropertyOf(Promise.prototype, key),
   );
 
+  const checkSafeEnoughKey = key => {
+    const val = pr[key];
+    if (val === undefined || typeof val === 'number') {
+      return true;
+    }
+    if (
+      typeof val === 'object' &&
+      val !== null &&
+      isFrozen(val) &&
+      getPrototypeOf(val) === Object.prototype
+    ) {
+      const subKeys = ownKeys(val);
+      if (subKeys.length === 0) {
+        return true;
+      }
+
+      // At the time of this writing, Node's async_hooks contains the
+      // following code, which we can also safely tolerate
+      //
+      // function destroyTracking(promise, parent) {
+      // trackPromise(promise, parent);
+      //   const asyncId = promise[async_id_symbol];
+      //   const destroyed = { destroyed: false };
+      //   promise[destroyedSymbol] = destroyed;
+      //   registerDestroyHook(promise, asyncId, destroyed);
+      // }
+
+      if (
+        subKeys.length === 1 &&
+        subKeys[0] === 'destroyed' &&
+        val.destroyed === false
+      ) {
+        return true;
+      }
+    }
+    return check(
+      false,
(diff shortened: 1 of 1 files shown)


packages/marshal/src/helpers/safe-promise.js
```diff
@@ -23,25 +23,52 @@ const checkPromiseOwnKeys = (pr, check) => {
     key => typeof key !== 'symbol' || !hasOwnPropertyOf(Promise.prototype, key),
   );
 
+  const checkSafeEnoughKey = key => {
```
mhofman, line 26: Nit: maybe we should be explicit in the function name that we're checking async hooks values?

packages/marshal/src/helpers/safe-promise.js
```diff
@@ -23,25 +23,52 @@ const checkPromiseOwnKeys = (pr, check) => {
...
+      if (
+        subKeys.length === 1 &&
+        subKeys[0] === 'destroyed' &&
+        val.destroyed === false
+      ) {
+        return true;
+      }
+    }
+    return check(
+      false,
```
mhofman, line 62: Same here at this point, the values should only be for async hook keys.

packages/marshal/src/helpers/safe-promise.js
mhofman: Unrelated, I don't know how I feel about allocating a closure for something which really should seldomly be encountered. I kinda would prefer if this was either inline in the `every`, or factored out of the checkPromiseOwnKeys function, but I'm probably prematurely optimizing.

mhofman (approved):
This is safe and can be merged as is, my review is mostly stylistic nits.

Sorry for not writing the correct check in the first place. Could you please open issue for me to look into unit testing this somehow?

packages/marshal/src/helpers/safe-promise.js
erights: `checkSafeAsyncHooksKey`.

Kept it factored out but refactored the context so you get your premature optimization anyway. I think a bit better even ;)

packages/marshal/src/helpers/safe-promise.js
erights: `Unexpected Node async_hooks additions to promise`

Done

packages/marshal/src/helpers/safe-promise.js
erights: Done

erights:
> Could you please open issue for me to look into unit testing this somehow?

https://github.com/endojs/endo/issues/1276

PR #1277, opened 2022-09-06
Avoid buffer copies in netstring

mhofman:
Rewrite the reader and writer of the netstring stream transform to avoid buffer copies and allocations as much as possible.
- The reader delays allocation of a buffer until the length prefix is parsed.
- The reader re-uses the input buffer if the whole data is a subset of the incoming chunk
  - possible point of improvement is to handle the trailing separator (comma) in a separate chunk, as may happen went sent chunked in the first place
- The writer can output in chunked mode, where the prefix, data and separator are sent to the output as separate writes/next calls.
- The writer can handle an array of buffers as input for producers which would otherwise concatenate their data.
- The writer no longer re-uses an internal buffer for its output as a consumer that doesn't fully process the buffer (e.g. caches it) before resolving would get corrupted data. While Node.js stream sinks should be safe since our adapter doesn't resolve until drain / write completion, not all sinks may be as robust.

Fix a bug discovered in node-stream where the writer's `next` didn't return an iterator result.

### DIFF
--- a/packages/netstring/reader.js
+++ b/packages/netstring/reader.js
@@ -10,8 +10,15 @@ const NINE = '9'.charCodeAt(0);
  * @param {Iterable<Uint8Array> | AsyncIterable<Uint8Array>} input
  * @param {Object} [opts]
  * @param {string} [opts.name]
+ * @param {number} [opts.maxMessageLength]
  */
-async function* makeNetstringIterator(input, { name = '<unknown>' } = {}) {
+async function* makeNetstringIterator(
+  input,
+  { name = '<unknown>', maxMessageLength = 999999999 } = {},
+) {
+  // eslint-disable-next-line no-bitwise
+  const maxPrefixLength = `${maxMessageLength | 0}:`.length;
+
   // byte offset of data consumed so far in the input stream
   let offset = 0;
 
@@ -41,6 +48,13 @@ async function* makeNetstringIterator(input, { name = '<unknown>' } = {}) {
           i += 1;
           if (c >= ZERO && c <= NINE) {
             lengthBuffer.push(c);
+            if (lengthBuffer.length === maxPrefixLength) {
+              throw new Error(
+                `Too long netstring length prefix ${JSON.stringify(
+                  String.fromCharCode(...lengthBuffer),
+                )}... at offset ${offset} of ${name}`,
+              );
+            }
           } else if (c === COLON && lengthBuffer.length) {
             lengthBuffer.push(c);
             break;
@@ -62,6 +76,10 @@ async function* makeNetstringIterator(input, { name = '<unknown>' } = {}) {
           if (Number.isNaN(remainingDataLength)) {
             throw new Error(
               `Invalid netstring prefix length ${prefix} at offset ${offset} of ${name}`,
+            );
+          } else if (remainingDataLength > maxMessageLength) {
+            throw new Error(
+              `Netstring message too big (length ${remainingDataLength}) at offset ${offset} of ${name}`,
             );
           }
           offset += lengthBuffer.length + 1;
@@ -121,6 +139,7 @@ async function* makeNetstringIterator(input, { name = '<unknown>' } = {}) {
  * @param {Iterable<Uint8Array> | AsyncIterable<Uint8Array>} input
  * @param {Object} [opts]
  * @param {string} [opts.name]
+ * @param {number} [opts.maxMessageLength]
  * @returns {import('@endo/stream').Reader<Uint8Array, undefined>} input
  */
 export const makeNetstringReader = (input, opts) => {
--- a/packages/netstring/reader.js
+++ b/packages/netstring/reader.js
@@ -95,7 +95,7 @@ async function* makeNetstringIterator(input, { name = '<unknown>' } = {}) {
           lengthBuffer = [];
           yield data;
         } else if (buffer.length) {
-          dataBuffer ??= new Uint8Array(remainingDataLength);
+          dataBuffer = dataBuffer || new Uint8Array(remainingDataLength);
           dataBuffer.set(buffer, dataBuffer.length - remainingDataLength);
           remainingDataLength -= buffer.length;
           buffer = buffer.subarray(buffer.length);
--- a/packages/netstring/reader.js
+++ b/packages/netstring/reader.js
@@ -95,8 +95,12 @@ async function* makeNetstringIterator(input, { name = '<unknown>' } = {}) {
           lengthBuffer = [];
           yield data;
         } else if (buffer.length) {
-          dataBuffer = dataBuffer || new Uint8Array(remainingDataLength);
-          dataBuffer.set(buffer, dataBuffer.length - remainingDataLength);
+          if (!dataBuffer && buffer.length === remainingDataLength) {
+            dataBuffer = buffer;
+          } else {
+            dataBuffer = dataBuffer || new Uint8Array(remainingDataLength);
+            dataBuffer.set(buffer, dataBuffer.length - remainingDataLength);
+          }
           remainingDataLength -= buffer.length;
           buffer = buffer.subarray(buffer.length);
         }
--- a/packages/netstring/reader.js
+++ b/packages/netstring/reader.js
@@ -2,67 +2,109 @@
 /// <reference types="ses"/>
 
 const COLON = ':'.charCodeAt(0);
-
-const decoder = new TextDecoder();
+const COMMA = ','.charCodeAt(0);
+const ZERO = '0'.charCodeAt(0);
+const NINE = '9'.charCodeAt(0);
 
 /**
  * @param {Iterable<Uint8Array> | AsyncIterable<Uint8Array>} input
  * @param {Object} [opts]
  * @param {string} [opts.name]
- * @param {number} [opts.capacity]
  */
-async function* makeNetstringIterator(
-  input,
-  { name = '<unknown>', capacity = 1024 } = {},
-) {
-  let length = 0;
-  let buffer = new Uint8Array(capacity);
+async function* makeNetstringIterator(input, { name = '<unknown>' } = {}) {
+  // byte offset of data consumed so far in the input stream
   let offset = 0;
 
-  for await (const chunk of input) {
-    if (length + chunk.byteLength >= capacity) {
-      while (length + chunk.byteLength >= capacity) {
-        capacity *= 2;
-      }
-      const replacement = new Uint8Array(capacity);
-      replacement.set(buffer, 0);
-      buffer = replacement;
-    }
-    buffer.set(chunk, length);
-    length += chunk.byteLength;
+  // The iterator can be in 2 states: waiting for the length, or waiting for the data
+  // - When waiting for the length, the lengthBuffer is an array containing
+  //   digits charCodes for the length prefix
+  // - When waiting for the data, the dataBuffer is either:
+  //   - null to indicate no data has been received yet.
+  //   - A newly allocated buffer large enough to accommodate the whole expected data.
+  //   In either case, remainingDataLength contains the length of the data to read.
+  //   If the whole data is received in one chunk, no copy is made.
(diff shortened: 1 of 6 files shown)


mhofman:
Best reviewed commit-by-commit

packages/stream-node/writer.js
```diff
@@ -57,7 +57,7 @@ export const makeNodeWriter = writer => {
           } else {
             resolve(undefined);
           }
-        }),
+        }).then(() => ({ done: false, value: undefined })),
```
kriskowal, line 60: We could harden and intern the medial iteration result.

We could also avoid some promise and closure allocations by changing `resolve(undefined)` to `resolve(nonFinalIteration)` and creating a `() => resolve(nonFinalIteration)` closure only if `writer.write` is false.

packages/netstring/writer.js
```diff
@@ -1,51 +1,54 @@
...
+      const prefix = getLengthPrefixCharCodes(messageLength);
 
-      buffer[colonAt] = COLON;
-      buffer.set(message, messageAt);
-      buffer[commaAt] = COMMA;
+      if (chunked) {
+        return Promise.all([
+          output.next(new Uint8Array(prefix)),
+          ...messageChunks.map(async chunk => output.next(chunk)),
+          output.next(COMMA_BUFFER),
```
kriskowal, line 35: I’m not sure whether calling `next` after a final iteration has well-defined behavior. I don’t think this is wrong, but it might be good to document somewhere that this writer might write after the target closes.

packages/netstring/writer.js
```diff
@@ -1,51 +1,54 @@
...
-// It also must be long enough to fit an ASCII decimal representation of the
-// length of the longest possible message.
-// This is a guess:
-const initialBufferLength = 128;
-
-const encoder = new TextEncoder();
+/** @param {number} length */
+const getLengthPrefixCharCodes = length =>
+  // eslint-disable-next-line no-bitwise
+  [...`${length | 0}:`].map(char => char.charCodeAt(0));
```
kriskowal, line 9: Switching from `map` to a `for` loop may yield a measurable improvement.
```suggestion
const getLengthPrefixCharCodes = length => {
  // eslint-disable-next-line no-bitwise
  const text = String(length | 0);
  const codes = new Uint8Array(text.length + 1);
  let i;
  for (i = 0; i < codes.length; i += 1) {
    codes[i] = text.charCodeAt(i);
  }
  codes[i] = COLON;
}
```
Can also avoid the intermediate Array and go straight to Uint8Array.

packages/netstring/writer.js
```diff
@@ -1,51 +1,54 @@
...
-        buffer,
+    async next(messageChunks) {
+      if (!Array.isArray(messageChunks)) {
+        messageChunks = [messageChunks];
+      }
+
+      const messageLength = messageChunks.reduce(
+        (acc, { length }) => acc + length,
+        0,
       );
```
kriskowal, line 27: This could have been a for loop.

packages/netstring/test/test-netstring.js
```diff
@@ -20,81 +20,115 @@ async function read(source) {
...
+test('fails reading no colon', readErroneousChunkedMessage, ['1A,']);
 
 function delay(ms) {
   return new Promise(resolve => {
     setTimeout(resolve, ms);
   });
 }
 
-const makeArrayWriter = () => {
+const makeArrayWriter = opts => {
```
kriskowal, line 106: Very ignorable nit: does our style guide have anything to say about abbreviations? My personal preference is to spell these out.

packages/netstring/test/test-netstring.js
```diff
@@ -119,9 +155,25 @@ test('concurrent writes', async t => {
...
-});
+};
+test('concurrent writes', concurrentWrites);
+test('concurrent writes (chunked)', concurrentWrites, { chunked: true });
 
-test('round-trip varying messages', async t => {
+const chunkedWrite = async (t, opts) => {
+  const { array, writer } = makeArrayWriter(opts);
+  const strChunks = ['hello', ' ', 'world'];
+  await writer.next(strChunks.map(strChunk => encoder.encode(strChunk)));
```
kriskowal, line 165: 👀 Apparent signature change from bytes to array of byte arrays.

packages/netstring/reader.js
```diff
@@ -2,67 +2,113 @@
...
  * @param {string} [opts.name]
- * @param {number} [opts.capacity]
  */
-async function* makeNetstringIterator(
-  input,
-  { name = '<unknown>', capacity = 1024 } = {},
-) {
-  let length = 0;
-  let buffer = new Uint8Array(capacity);
+async function* makeNetstringIterator(input, { name = '<unknown>' } = {}) {
```
kriskowal, line 14: I’ve heard that generators are slow and that state machines are the way to go for speed. I used a generator in hopes that a the readability would be more important than speed. This appears to be the time to revisit that choice. But, I like testing the fast one in comparison to the slow-but-readable one for correctness, especially in a fuzzer.

packages/netstring/reader.js
```diff
@@ -2,67 +2,113 @@
...
+        let i = 0;
+        while (i < buffer.length) {
+          const c = buffer[i];
+          i += 1;
+          if (c >= ZERO && c <= NINE) {
+            lengthBuffer.push(c);
+          } else if (c === COLON && lengthBuffer.length) {
+            lengthBuffer.push(c);
+            break;
+          } else {
```
kriskowal, line 47: Should throw an error of the lengthBuffer length exceeds the maximum length necessary to express the maximum message length, so the reader can’t be induced to OOM by a long length prefix.

kriskowal: Similarly below. Can punt to an issue since it would be a new level of hardening unrelated to this performance push.

packages/netstring/reader.js
```diff
@@ -2,67 +2,113 @@
...
+          if (Number.isNaN(remainingDataLength)) {
+            throw new Error(
+              `Invalid netstring prefix length ${prefix} at offset ${offset} of ${name}`,
+            );
+          }
+          offset += lengthBuffer.length + 1;
+          lengthBuffer = null;
+        }
+      }
+
```
kriskowal, line 71: I unfortunately need to lean heavily on tests to verify the logic in here. I don’t see any flaws but from experience writing similar code, I’m sure that’s no indication. A fuzzer would make me more comfortable. A boundary searcher would be even better. A boundary searcher would be a test that visits various permutations of message sizes +/-1 around powers of two or other likely boundary cases, in various orders, trying to tease out edge cases.

kriskowal (changes requested):
Good call on making {chunked} writing opt-in.

We need:

- [ ] fuzzer (a test that runs a producer and consumer using a PRNG to generate a very long sequence of messages)
- [ ] benchmark (before and after figures)
- [ ] confirmation that agoric-sdk still passes with this dependency pinned

 I could be convinced to run with scissors except for the last bit.

kriskowal:
Good catch re the undefined final iteration result!

mhofman:
> * [ ]  fuzzer (a test that runs a producer and consumer using a PRNG to generate a very long sequence of messages)
Doesn't one of the unit test already do this?
> * [ ]  benchmark (before and after figures)
I tried to get perf before/after in agoric-sdk, but the end-to-end timings are very noisy. Didn't have time to get better data
> * [ ]  confirmation that agoric-sdk still passes with this dependency pinned
I did run agoric-sdk tests including integration with this PR as a patch-package

packages/stream-node/writer.js
mhofman: Yeah I considered those as well, but went for this for simplicity's sake. If you have a preference I'm happy to go any other way.

packages/netstring/writer.js
mhofman: A simple accumulator is the one case where I feel like `reduce` is acceptable but if it hurts too much to read, I'm happy to make it a loop.

packages/netstring/writer.js
mhofman: Given the size of the iteration I don't think the performance impact is measurable. Plus v8 can optimize these quite a bit these days.

I also wanted to avoid allocating an `ArrayBuffer` in the case where I just need to set the charCodes in an existing buffer. Not sure what's better, a temporary `ArrayBuffer` + `UInt8Array` or a temporary `Array`

packages/netstring/reader.js
mhofman: Yeah since we already have a state machine inside the generator, you're right going all the way might be better.

packages/netstring/reader.js
mhofman: What do you mean by maximum message length? I believe the netstring spec doesn't have such a notion (even though the example implies a max length of 999999999 bytes). The tagged netstring spec does have that length as specified limit.

Would we consider 1GiB-1 as the upper limit, or would it be a config option of the reader?

packages/netstring/reader.js
mhofman: I'm not sure what powers of 2 have to do with the length prefix logic here. I believe the unit tests already cover most edge cases, which are:
- break in the middle of the prefix
- break between the number prefix and the colon
- invalid character in the prefix

The one case not covered is an empty prefix before the colon, and I can definitely fix that.

Can you think of any other cases to test for?

packages/netstring/test/test-netstring.js
mhofman: Arguably I also prefer spelling out but used the term `opts` because I copied it from the JSDoc `@param` of `makeNetstringWriter`, which I itself copied from the existing reader one.

packages/netstring/test/test-netstring.js
mhofman: Correct, `chunkedWrite` tests the new chunks receiving mode.

packages/netstring/writer.js
mhofman: Yeah it's definitely an aggressive writer in chunk writing mode. Good thing it's not the default, and the node-stream writer can take it. I can add a note.

packages/netstring/test/test-netstring.js
kriskowal: _mea culpa_!

packages/netstring/reader.js
kriskowal: We should definitely have a config option and highly encourage a reasonable limit. Protobuf has a hard limit at 2GB, which is close enough to Infinity to be a reasonable default but not adequate for limiting the sender’s ability to induce an OOMkill remotely.

kriskowal (approved):
- [x] round-trip messages of varying lengths is good enough to cover the fuzzer case. Thanks for reminding me of its existence. I grepped for fuzz.
- [x] I’m satisfied that you got an end-to-end test passing
- [ ] Would be nice to have a benchmark to confirm this moves performance in the right direction in isolation. Benchmarks are surprising. I’m imagining something lightweight, like a generalization of the round-trip test that runs until some minimum time threshold and reports the iteration count over the total time.

packages/netstring/reader.js
mhofman: Decided to not do as consuming the input without a `for await of` loop makes things more complicated.

packages/netstring/reader.js
mhofman: Done, made configurable with the default being what the netstring specs imply.

packages/netstring/reader.js
mhofman: Added an empty prefix test

packages/netstring/writer.js
mhofman: Note added in the description of the maker.

packages/stream-node/writer.js
mhofman: Fixed. Also hardened the `finalIteration` result as well.

mhofman (commented):
@kriskowal PTAL

kriskowal (approved):

PR #1278, opened 2022-09-08
feat(bundle-source): Add JSON cache mode

kriskowal:
- Adds a `--cache-json` mode that:
  - Generates `.json` bundle files in JSON format.
  - Generates `-json-meta.json` meta files in JSON format.
- Adds a `--cache-js` mode that should replace `--to`:
  - Generates `.js` bundle files in JSON format.
    Adds a UNIX newline absent when using `--to`.
  - Generates `-json-meta.json` meta files in JSON format instead of
    `-meta.js`.

### DIFF
--- a/packages/bundle-source/NEWS.md
+++ b/packages/bundle-source/NEWS.md
@@ -1,3 +1,14 @@
+# Next release
+
+- Adds a `--cache-json` mode that:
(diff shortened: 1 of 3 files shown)


dckc (approved):
yes, let's

PR #1279, opened 2022-09-09
fix(compartment-mapper): Harden bundles

kriskowal:
This converges the style of generated bundles on our known best practices, ensuring that emulated module namespace objects would be frozen and that hardening the internal cells would be a no-op after lockdown.


erights:
What is the status of this? Is it still relevant and needed, and still awaiting my review?

erights (approved):
LGTM. Sorry this dropped off my radar. Since it has been sitting around awhile, I don't know how stale it is. If rebasing and repairing result in significant changes, please ask me to review again.

PR #1280, opened 2022-09-10
fix: alt syntax for positive but faster assertions

erights:
Like https://github.com/Agoric/agoric-sdk/pull/6174

Replace some occurrences of
```js
assert(..condExpr.., X`..detailsString..`);
```
with
```js
..condExpr.. || assert.fail(X`..detailsString..`);
```

I only made the changes that I was easily able to do via regexp search and replace in vscode
```
^\s*assert\(\n\s*(.*?),\n\s*X`(.*?)`,\n\s*\);
($1) || assert.fail(X`$2`);
```
and similar, followed by a global `yarn lint-fix` and then repair until everything tested green again. No doubt there are more cases that I missed. But from @dtribble 's measurements, these are already enough to show significant perf improvements.

### DIFF
--- a/packages/captp/src/atomics.js
+++ b/packages/captp/src/atomics.js
@@ -23,10 +23,10 @@ const STATUS_FLAG_REJECT = 4;
  * @param {SharedArrayBuffer} transferBuffer the backing buffer
  */
 const splitTransferBuffer = transferBuffer => {
-  assert(
-    transferBuffer.byteLength >= MIN_TRANSFER_BUFFER_LENGTH,
-    X`Transfer buffer of ${transferBuffer.byteLength} bytes is smaller than MIN_TRANSFER_BUFFER_LENGTH ${MIN_TRANSFER_BUFFER_LENGTH}`,
-  );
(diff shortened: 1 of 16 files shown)


erights:
Ready for review!

@arirubinstein for some reason, I cannot add you as an official reviewer. Please consider yourself a reviewer anyway. Thanks.

kriskowal:
If this is intended to be comprehensive, this does not appear to include compartment-mapper, which does rely on assertions extensively for type narrowing.

erights:
> If this is intended to be comprehensive, this does not appear to include compartment-mapper, which does rely on assertions extensively for type narrowing.

It is intended to be comprehensive. Everything that should have been covered that I instead skipped, I skipped only because
   * it wasn't caught by the regexp search-and-replaces that I tries, or
   * the auto-replacement had some other problem, as regexp-based transforms often do, so I backed those out.

In any case, after this PR we should indeed hunt down and convert the others. Fortunately, there's no coupling so we can do these in separate PRs.

PR #1282, opened 2022-09-12
fix: Provide smallcaps encoding in addition to capData

erights:
Based on
https://github.com/endojs/endo/blob/38078aec70975d87a0a76918d0ba788e87420581/packages/marshal/src/marshal.js

This key thing about this PR is that the unserializer understands both new smallcaps and old capData protocols, switching between them based on a leading `#`. The serialized defaults to the old protocol, based on the `useSmallcaps` option to `makeMarshal`. With this defaulting to off, everything is trivially green, but that puts us in a position to switch different subsystems on separately as we put in the work to do so. In this PR itself, only `test-marshal-smallcaps.js` does so.

See https://github.com/endojs/endo/pull/1283 for an experiment of changing the default `useSmallcaps` to true, to see what breaks. However, be aware that some tests and even some modules in this PR explicitly set `useSmallcaps` to `false`, evading this experiment, because they're known to not yet work with smallcaps.

# SECURITY CONCERN

Even though our serializer, as of this PR, does not send smallcaps by default, others can, including malformed smallcaps. For capData, we've reviewed the code for robustness in the face of malformed capData. ***REVIEWERS, please*** review the unserializer for robustness against malformed smallcaps. If we cannot be confident that we're at least as robust against malformed smallcaps as we were against malformed capData, that would be a reason to delay this PR until we are.


erights:
Ready for review!

@arirubinstein for some reason, I cannot add you as an official reviewer. Please consider yourself a reviewer anyway. Thanks.

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,461 @@
...
+    const passStyle = passStyleOf(passable);
+    switch (passStyle) {
+      case 'null':
+      case 'boolean': {
+        return passable;
+      }
+      case 'string': {
+        if (passable !== '' && SpecialChars.includes(passable[0])) {
+          return `'${passable}`;
+        }
```
mhofman, line 138: Wondering if problems may not arise in the future from the lexical order of conditionally hilbert-ed strings? E.g. `"*a" < "+a"` but their encoding has a different lexical order: `"'+a" < "*a"`. 
I guess the keys of copyRecords are not subject to such hilbert treatment so it should not interfere with a potential canonicalizing JSON encoder, but it does raise the question of how we would handle non string keys in the future.

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,461 @@
...
+          return '#-Infinity';
+        }
+        return passable;
+      }
+      case 'bigint': {
+        const str = String(passable);
+        return passable < 0n ? `${str}n` : `+${str}n`;
+      }
+      case 'copyRecord': {
+        if (hasOwnPropertyOf(passable, SCLASS)) {
```
mhofman, line 169: Why not use the existing helper?

```suggestion
        if (hasSClass(passable)) {
```

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,461 @@
...
+   * @returns {SmallcapsEncoding} except that `encodeToSmallcaps` does not generally
+   * `harden` this result before returning. Rather, `encodeToSmallcaps` is not
+   * directly exposed.
+   * What's exposed instead is a wrapper that freezes the output before
+   * returning. If this turns out to impede static analysis for `harden` safety,
+   * we can always put the (now redundant) hardens back in. They don't hurt.
+   */
+  const encodeToSmallcapsRecur = passable => {
+    // First we handle all primitives. Some can be represented directly as
+    // JSON, and some must be encoded as [SCLASS] composites.
```
mhofman, line 128: This comment is no longer accurate as SCLASS is not not the only encoding for non-JSON serializable data.

mhofman (commented):
I only did a quick pass to understand the smallcaps encoding. If I understand it correctly (it would be nice to have a description to the design and/or comparisons between capdata and smallcaps), the main difference with capdata is the way BigInt, remotables/promises and certain specific values are encoded as strings instead of being encoded as qclass "objects".

While I see how this may result in less verbose (and potentially more readable) output, I'm somewhat surprised it'd lead to such an improvement in performance.

erights:
> I'm somewhat surprised it'd lead to such an improvement in performance.

We have not measured this, and so do not know. I can explain why I expect a big reduction in allocations, but my pre-measurement intuitions are often wrong. @dtribble , we need to measure this!

The measured increase in performance was https://github.com/Agoric/agoric-sdk/pull/6174 + https://github.com/endojs/endo/pull/1280 . @dtribble measured this, but only for pattern matching. I suspect that at least the endo-level changes generalize, but again we should be skeptical until we measure. We need to more realistically measure this!

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,461 @@
...
+            );
+          }
+          return result;
+        }
+        default: {
+          return encoding;
+        }
+      }
+    }
+    if (!isObject(encoding)) {
```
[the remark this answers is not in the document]
erights: Care to suggest a code pattern?

packages/marshal/src/encodeToSmallcaps.js
erights: > Wondering if problems may not arise in the future from the lexical order of conditionally hilbert-ed strings? E.g. `"*a" < "+a"` but their encoding has a different lexical order: `"'+a" < "*a"`. I guess the keys of copyRecords are not subject to such hilbert treatment so it should not interfere with a potential canonicalizing JSON encoder,

exactly.

> but it does raise the question of how we would handle non string keys in the future.

Do you suggest we do anything now in anticipation of that? Or should we cross that bridge when it happens?

FWIW, Records and Tuples have no non-string property names, so that would not be the source of such a change. I doubt we'll ever make such a change because it makes the protocol *much* less friendly to non-js endpoints.

packages/marshal/src/encodeToSmallcaps.js
erights: Good catch. How about
```js
    // First we handle all primitives. Some can be represented directly as
    // JSON, and some must be encoded into smallcaps strings.
```
?
For all primitives, smallcaps uses only strings.

erights:
>  If I understand it correctly (it would be nice to have a description to the design and/or comparisons between capdata and smallcaps),

Yes. (FWIW, I did not expect us to decide today to put this into review immediately, or I would have been more ready. But yes, this needs to be written.)

 the main difference with capdata is the way BigInt, remotables/promises and certain specific values are encoded as strings instead of being encoded as qclass "objects".

Yes. This means that strings themselves do not always encode as the same string, which also requires that we hilbert the string dimension.

packages/marshal/src/encodeToSmallcaps.js
erights: Done

packages/marshal/src/encodeToSmallcaps.js
erights: Resolving.

packages/marshal/src/encodeToSmallcaps.js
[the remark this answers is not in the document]
mhofman: I was thinking something like:
```js
if (typeof encoding === 'number' || typeof encoding === 'boolean' || encoding == null) {
  return encoding;
}
if (!isObject(encoding)) {
  assert.fail(X`invalid encoding, unexpected type ${typeof encoding)}`);
}
```

packages/marshal/src/encodeToSmallcaps.js
mhofman: > Do you suggest we do anything now in anticipation of that? Or should we cross that bridge when it happens?

It may be safe enough to Hilbert all `@` starting keys if we ever believe symbols may become acceptable? But yeah maybe there will never be a strong motivation to support non-string keys.

mhofman: Got an idea: why not use `&` for promises and `%` for symbols, and hilbert all the string starting with a character between 33 and 47 (to give us some room for the future)?

That would mean that strings before and after Hilbert would always have the same lexical order.

packages/marshal/src/encodeToSmallcaps.js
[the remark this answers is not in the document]
erights: Done

packages/marshal/src/encodeToSmallcaps.js
erights: You got me to look at an ascii table!

Not quite so simple, but something like this is still worth doing. Besides `'?'` (63) for promises, there's also `'@'` (64) for symbols. And we have to deal with our special property name `@qclass`. 

But there's one way things become simpler: If we do hilbert property names, then an incoming `"@qclass":` becomes `"'@qclass":`, which AFAICT, fully takes care of hilberting objects. We can drop all the object-based hilbert logic, which was kinda subtle https://github.com/endojs/endo/blob/246c0b2a51097646d04b64caeb4a309b6afb7ccc/packages/marshal/src/encodeToSmallcaps.js#L403

erights: > Got an idea: why not use ... and % for symbols, 

Doh! Sorry I read too quickly before responding.

packages/marshal/test/test-marshal-smallcaps.js
```diff
@@ -0,0 +1,219 @@
...
+const { freeze, isFrozen, create, prototype: objectPrototype } = Object;
+
+// this only includes the tests that do not use liveSlots
+
+const makeTestMarshal = () =>
+  makeMarshal(undefined, undefined, {
+    errorTagging: 'off',
+    useSmallcaps: true,
+  });
+
```
dckc, line 18: I'd like to see a test that serves as a succinct design description of smallcaps so that in less than a page of code we would clearly see:

>  - `$` is slot
>  - `+` or `-` is bigint
> - `#` is constant (e.g., #NaN)
> - `'` is quote so that any of the above can be escaped

and perhaps another showing that smallcaps saves 40% in size in examples such as the one in

https://github.com/Agoric/agoric-sdk/issues/2589#issuecomment-1238686087

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,469 @@
...
+          return '#Infinity';
+        }
+        if (passable === -Infinity) {
+          return '#-Infinity';
+        }
+        return passable;
+      }
+      case 'bigint': {
+        const str = String(passable);
+        return passable < 0n ? `${str}n` : `+${str}n`;
```
dtribble, line 166: There is no reason to redundantly waste a character on the trailing n. Anything encoded this way is a bigint, and the  JS syntax reminder doesn't add any value.

packages/marshal/src/encodeToSmallcaps.js
erights: Done! from https://github.com/endojs/endo/pull/1284 successful experiment

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,434 @@
...
+ * @param {string} encodedStr
+ * @returns {boolean}
+ */
+const startsSpecial = encodedStr => {
+  if (encodedStr === '') {
+    return false;
+  }
+  const c = encodedStr[0];
+  // eslint-disable-next-line yoda
+  return '!' <= c && c <= '-';
```
mhofman, line 71: I believe `.charCodeAt(0)` and number comparison is faster than `[0]` and string comparison.

Some "old" thread about it, I'm sure there are others:
https://www.reddit.com/r/webdev/comments/e2eft5/til_that_charcodeat_is_faster_than_accessing_a/

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,434 @@
...
+    // JSON, and some must be encoded into smallcaps strings.
+    const passStyle = passStyleOf(passable);
+    switch (passStyle) {
+      case 'null':
+      case 'boolean': {
+        // pass through to JSON
+        return passable;
+      }
+      case 'string': {
+        if (startsSpecial(passable[0])) {
```
mhofman, line 144:
```suggestion
        if (startsSpecial(passable)) {
```

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,434 @@
...
+            encodeToSmallcapsRecur(passable[name]),
+          ]),
+        );
+      }
+      case 'copyArray': {
+        return passable.map(encodeToSmallcapsRecur);
+      }
+      case 'tagged': {
+        return {
+          '#tagged': getTag(passable),
```
mhofman, line 203: Curious why the conjugation. `#tag` would be as readable and save 3 chars, no?

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,434 @@
...
+              }
+            }
+          }
+          case '+':
+          case '-': {
+            const last = encoding.length - 1;
+            assert(
+              encoding[last] === 'n',
+              X`Encoded bigint must start with "+" or "-" and end with "n": ${encoding}`,
+            );
```
mhofman, line 341: I believe @dtribble suggested dropping the trailing `n`, which would allow us to avoid the `slice` call,

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,434 @@
...
+            const result = decodePromiseFromSmallcaps(encoding);
+            if (passStyleOf(result) !== 'promise') {
+              assert.fail(
+                X`internal: decodePromiseFromSmallcaps option must return a promise: ${result}`,
+              );
+            }
+            return result;
+          }
+          default: {
+            return encoding;
```
mhofman, line 363: Should we enforce that the string doesn't start with one of the other currently unused hilberted chars? Maybe we can add `case` statements for those ?

mhofman (commented):
Copy over from https://github.com/endojs/endo/pull/1284#pullrequestreview-1106150594

packages/marshal/src/encodeToSmallcaps.js
erights: Done

packages/marshal/test/test-marshal-smallcaps.js
erights: Between @dtribble and I, the logical tests are now done. I'm gonna resolve this without the size test, but please unresolve if you think I need to do that before merging.

packages/marshal/src/encodeToSmallcaps.js
erights: See conversation at https://github.com/endojs/endo/pull/1284#pullrequestreview-1106150594

I still don't want to do this unless measurements show that it is needed on XS. Resolving.

packages/marshal/src/encodeToSmallcaps.js
erights: Done

packages/marshal/src/encodeToSmallcaps.js
erights: Done

packages/marshal/src/encodeToSmallcaps.js
erights: Done

packages/marshal/src/encodeToSmallcaps.js
erights: Done

packages/marshal/test/test-marshal-smallcaps.js
```diff
@@ -0,0 +1,327 @@
...
+  // Hilbert record
+  assertSer(
+    harden({
+      '#tag': 'what',
+      '#error': 'me',
+      '#huh': 'worry',
+      '': 'empty',
+      '%sym': 'not a symbol',
+    }),
+    '#{"":"empty","\'#error":"me","\'#huh":"worry","\'#tag":"what","\'%sym":"not a symbol"}',
```
mhofman, line 323: Oh I didn't realize that `'` was getting escaped in JSON encoding. That's unfortunate...

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -32,37 +28,48 @@ import {
...
-
-/**
- * `'` - escaped string
- * `+` - non-negative bigint
- * `-` - negative bigint
- * `#` - constant
- * `@` - symbol
- * `$` - remotable
- * `?` - promise
+ * Smallcaps considers the characters between `!` (ascii code 33)
```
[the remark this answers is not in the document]
erights: Very glad you asked. Definitely worth writing.

```js
/**
 * An `encodeToSmallcaps` function takes a passable and returns a tree
 * of plain JSON objects, i.e., those that (except for identity) round
 * trip through a plain `JSON.stringify` and `JSON.parse` with no
 * replacers or resolvers. We call this encoding of a passable into a
 * tree of plain JSON objects a Smallcaps Encoding
 *
 * A `decodeFromSmallcaps` function takes as argument what it
 * *assumes* is the result of a plain `JSON.parse` with no resolver. It then
 * must validate that it is a valid Smallcaps Encoding, and if it is,
 * return a corresponding passable.
```

Done

packages/marshal/test/test-marshal-smallcaps.js
mhofman: I'm an idiot, this is because the test uses ` ' ` for the outer quotes. Please update the test to use `` ` `` instead.

erights: I think your looking at the wrong metalevel. To print the string as a single-quote quoted string, the embedded single quotes are escaped. But they're not actually escaped in JSON string. In any case, I expect these hilbert cases to be rare enough that it shouldn't make much difference. Resolving.

mhofman: Yes I read it wrong, but I think to avoid confusion, we can very well use a template / backticks in this case. Reopened.

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,452 @@
...
+      // valid Passables. For example, because they're not frozen.
+      // The special case can only ever apply at the root, and therefore
+      // outside the recursion, since an error could only be deeper in
+      // a passable structure if it were passable.
+      //
+      // We pull out this special case because, for these errors, we're much
+      // more interested in reporting whatever diagnostic information they
+      // carry than we are about reporting problems encountered in reporting
+      // this information.
+      return harden(encodeErrorToSmallcaps(passable));
```
gibson042, line 265:
```suggestion
      const result = encodeErrorToSmallcaps(passable);
      if (
        typeof result === 'object' &&
        hasOwnPropertyOf(result, '#error') &&
        typeof result['#error'] === 'string'
      ) {
        return harden(result);
      }
      assert.fail(
        X`internal: Error encoding must have "#error" property: ${result}`,
      );
```
Or alternatively and probably better, `encodeErrorToSmallcaps` and the other input encoders could be replaced up front with a wrapper that performs the assertions:
```patch
 export const makeEncodeToSmallcaps = ({
   encodeRemotableToSmallcaps = dontEncodeRemotableToSmallcaps,
   encodePromiseToSmallcaps = dontEncodePromiseToSmallcaps,
   encodeErrorToSmallcaps = dontEncodeErrorToSmallcaps,
 } = {}) => {
+  encodeRemotableToSmallcaps = enforceGuards(
+    encodeRemotableToSmallcaps,
+    // TODO? M.startsWith/M.matchesRegExp/etc.
+    M.call(M.any()).returns(M.and(M.string(), {length: M.gte(1), 0: '$'})),
+  );
+  encodePromiseToSmallcaps = enforceGuards(
+    encodePromiseToSmallcaps,
+    M.call(M.any()).returns(M.and(M.string(), {length: M.gte(1), 0: '&'})),
+  );
+  encodeErrorToSmallcaps = enforceGuards(
+    encodeErrorToSmallcaps,
+    M.call(M.any()).returns({'#error': M.string()}),
+  );
```

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,452 @@
...
+          const result = [];
+          const { length } = encoding;
+          for (let i = 0; i < length; i += 1) {
+            result[i] = decodeFromSmallcaps(encoding[i]);
+          }
+          return result;
+        }
+
+        if (hasOwnPropertyOf(encoding, '#tag')) {
+          const { '#tag': tag, payload, ...rest } = encoding;
```
gibson042, line 400: I haven't looked at the tests yet, but we will want to have one verifying that an object with both "#tag" and "#error" properties is rejected to ensure that a refactoring here does not break anything.

packages/marshal/src/marshal.js
```diff
@@ -142,24 +147,90 @@ export const makeMarshal = (
...
+        return serializeSlotToSmallcaps('&', promise);
+      };
+
+      const encodeErrorToSmallcaps = err => {
+        // Not the most elegant way to reuse code. TODO refactor.
+        const capDataErr = encodeErrorToCapData(err);
+        const { [QCLASS]: _, message, ...rest } = capDataErr;
+        return harden({
+          '#error': message,
+          ...rest,
```
gibson042, line 197: Should `rest` itself be encoded? If it isn't, then an error object with e.g. a "#tag" property will be encoded into an invalid structure.

packages/marshal/src/marshal.js
```diff
@@ -216,28 +287,84 @@ export const makeMarshal = (
...
-    const fullRevive = makeDecodeFromCapData({
+    const decodeRemotableFromSmallcaps = stringEncoding => {
+      assert(stringEncoding.startsWith('$'));
+      // slots: $slotIndex.iface or $slotIndex
+      const i = stringEncoding.indexOf('.');
+      const index = Number(stringEncoding.slice(1, i));
+      // assert(i > 0, X`${rawTree} is not a valid slot encoding`);
+      // i < 0 means there was no iface included
+      const iface = i < 0 ? undefined : stringEncoding.slice(i + 1);
+      return unserializeSlot(index, iface);
```
gibson042, line 298: Won't `stringEncoding.slice(1, i)` chop off the last digit when `i` is -1 because `stringEncoding` doesn't contain a "."?
```suggestion
      const i = stringEncoding.indexOf('.');
      const slotIndex = Number(stringEncoding.slice(1, i < 0 ? Infinity : i));
      // i < 0 means there was no iface included.
      const iface = i < 0 ? undefined : stringEncoding.slice(i + 1);
      return unserializeSlot(slotIndex, iface);
```
Are we missing a slot deserialization test, or am I just wrong about the behavior of this code?

packages/marshal/src/marshal.js
```diff
@@ -216,28 +287,84 @@ export const makeMarshal = (
...
+    };
+
+    const decodePromiseFromSmallcaps = stringEncoding => {
+      assert(stringEncoding.startsWith('&'));
+      // slots: $slotIndex.iface or $slotIndex
+      const i = stringEncoding.indexOf('.');
+      const index = Number(stringEncoding.slice(1, i));
```
gibson042, line 305: This seems to suffer the same problem. Suggestion:
```js
const makeDecodeSlotFromSmallcaps = prefix => {
  return stringEncoding => {
    assert(stringEncoding.startsWith(prefix));
    // slots: $slotIndex.iface or $slotIndex
    const i = stringEncoding.indexOf('.');
    const slotIndex = Number(stringEncoding.slice(1, i < 0 ? Infinity : i));
    // i < 0 means there was no iface included.
    const iface = i < 0 ? undefined : stringEncoding.slice(i + 1);
    return unserializeSlot(slotIndex, iface);
  };
};
const decodeRemotableFromSmallcaps = makeDecodeSlotFromSmallcaps('$');
const decodePromiseFromSmallcaps = makeDecodeSlotFromSmallcaps('&');
```

packages/marshal/src/marshal.js
```diff
@@ -216,28 +287,84 @@ export const makeMarshal = (
...
       assert.fail(X`unserialize() given non-capdata (.slots are not Array)`);
-    const rawTree = harden(JSON.parse(data.body));
-    const fullRevive = makeFullRevive(data.slots);
-    const result = harden(fullRevive(rawTree));
+    const { reviveFromCapData, reviveFromSmallcaps } = makeFullRevive(slots);
+    let result;
+    // JSON cannot begin with a '#', so this is an unambiguous signal.
+    if (body.startsWith('#')) {
+      const smallcapsBody = body.slice(1);
+      const encoding = harden(JSON.parse(smallcapsBody));
```
gibson042, line 359: There's probably a future optimization that skips this `harden` via e.g. a revive option. It's currently necessary for the deep `passStyleOf` calls, but things might be noticeably faster if we didn't walk the data structure twice (once to recursively freeze and once to actually decode data).

packages/marshal/src/marshal.js
```diff
@@ -216,28 +287,84 @@ export const makeMarshal = (
...
+      result = harden(reviveFromSmallcaps(encoding));
```
gibson042, line 360: Likewise here, although we'd need to be careful about not violating guarantees if this were moved back into the reviver. If we introduce something like `enforceGuards` as suggested above, that might be usable (e.g., wrapping `decodeFromSmallcaps` and `decodeFromCapData`).

And how should we keep track of such potential optimizations? TODO comment, GitHub issue, other...?

packages/marshal/src/encodeToSmallcaps.js
gibson042: I also second @mhofman's suggestion of `EXCLAMATION` and `DASH` constants from https://github.com/endojs/endo/pull/1284#discussion_r969959237

kriskowal: Call and raise `const BANG = '!'.charCodeAt(0)`

erights: Measurements, thanks!

Within `startsSpecial`, done.

packages/marshal/test/test-marshal-smallcaps.js
mhofman: I had meant to quote the expected value with backticks to avoid escapes, but I think I do find the bang more readable as an escape anyway.

```js
    `#{"":"empty","'#error":"me","'#huh":"worry","'#tag":"what","'%sym":"not a symbol"}`,
```

packages/marshal/src/types.js
```diff
@@ -214,6 +214,7 @@ export {};
  * that error with its errorId. Thus, if `marshalSaveError` in turn logs
  * to the normal console, which is the default, then the console will
  * show that note showing the associated errorId.
+ * @property {boolean} [useSmallcaps]
```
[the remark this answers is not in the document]
erights: Good point. Done.

packages/marshal/src/encodeToCapData.js
```diff
@@ -396,11 +390,15 @@ export const makeDecodeFromCapData = ({
           return result;
         }
 
+        // @ts-expect-error This is the error case we're testing for
+        case 'ibid': {
+          assert.fail(
+            X`The capData protocol no longer supports [${q(QCLASS)}]: ${q(
+              qclass,
+            )} encoding: ${jsonEncoded}.`,
+          );
```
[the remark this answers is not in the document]
erights: Done

packages/captp/src/captp.js
```diff
@@ -611,10 +611,11 @@ export const makeCapTP = (
         assert.fail(X`Trap(${target}) target cannot be a promise`);
 
       const slot = valToSlot.get(target);
-      (slot && slot[1] === '-') ||
+      if (!(slot && slot[1] === '-')) {
+        // TypeScript confused about `||` control flow so use `if` instead
+        // https://github.com/microsoft/TypeScript/issues/50739
```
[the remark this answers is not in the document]
erights: Done (only for endo)

packages/marshal/src/marshal.js
```diff
@@ -142,24 +147,90 @@ export const makeMarshal = (
...
+      const encodeToSmallcaps = makeEncodeToSmallcaps({
+        encodeRemotableToSmallcaps,
+        encodePromiseToSmallcaps,
+        encodeErrorToSmallcaps,
+      });
+
+      const encoded = encodeToSmallcaps(root);
+      const smallcapsBody = JSON.stringify(encoded);
+      return harden({
+        // Valid JSON cannot begin with a '#', so this is a valid signal
```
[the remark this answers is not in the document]
erights: The simple suggestion, done.

The more ambitious suggestion would be more in @FUDCo and @warner 's wheelhouse, and is well beyond this PR regardless.

@FUDCo @warner , what do you think?

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,441 @@
...
+
+const { ownKeys } = Reflect;
+const { isArray } = Array;
+const { is, fromEntries } = Object;
+const { details: X, quote: q } = assert;
+
+/**
+ * Smallcaps considers the characters between `!` (ascii code 33)
+ * and `-` (ascii code 45) to be special. These characters, in order, are
+ * `!"#$%&'()*+,-` Of these, smallcaps currently uses the following:
```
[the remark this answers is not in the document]
erights: Done

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,441 @@
...
+ *  * `+` - non-negative bigint
+ *  * `-` - negative bigint
+ *  * `#` - manifest constant
+ *  * `%` - symbol
+ *  * `$` - remotable
+ *  * `&` - promise
+ *
+ * All other special characters (`!"()*,`) are reserved for future use.
+ *
+ * The menifest constants that smallcaps currently uses for values:
```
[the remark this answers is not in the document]
erights: Thought I already did that, but it was still there. In any case, done.

packages/marshal/src/marshal.js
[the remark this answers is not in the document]
mhofman: Well often we process the slots without touching the body. The question is whether we should parse and stringify the body every time we do that. I don't think it'd make sense, however that may not be necessary if we had custom revivers and serializers in those places?

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,441 @@
...
+        }
+        assert.fail(
+          X`internal: Promise encoding must start with "&": ${result}`,
+        );
+      }
+      case 'error': {
+        const result = encodeErrorToSmallcaps(passable);
+        if (
+          typeof result === 'object' &&
+          typeof result['#error'] === 'string'
```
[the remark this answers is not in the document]
erights: Done

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,441 @@
...
+        ) {
+          return result;
+        }
+        assert.fail(
+          X`internal: Error encoding must have "#error" property: ${result}`,
+        );
+      }
+      default: {
+        assert.fail(X`unrecognized passStyle ${q(passStyle)}`, TypeError);
```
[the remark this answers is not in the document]
erights: Yes, good to promulgate the distinction. Done.

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,452 @@
...
+const { isArray } = Array;
+const { is, fromEntries } = Object;
+const { details: X, quote: q } = assert;
+
+/**
+ * An `encodeToSmallcaps` function takes a passable and returns a tree
+ * of plain JSON objects, i.e., those that (except for identity) round
+ * trip through a plain `JSON.stringify` and `JSON.parse` with no
+ * replacers or resolvers. We call this encoding of a passable into a
+ * tree of plain JSON objects a Smallcaps Encoding
```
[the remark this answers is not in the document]
erights: Yes, better. Done.

packages/marshal/src/encodeToSmallcaps.js
erights: Did the first simple suggestion.

A fly in the ointment of the second suggestion is that `M` is defined by `@agoric/store`, and so difficult for marshal to import ;). Nevertheless, I agree that something along those lines would be better. But not doing it in this PR.

packages/marshal/src/marshal.js
erights: Skipping this one.

packages/marshal/src/marshal.js
erights: Skipping this one too. Issue would be most useful. If you file one, please link it here.

packages/marshal/src/marshal.js
dtribble: You are correct that this is wrong. I was originally always including the "." I think.  I woudl use `undefined` rather than `Infinity`, but that's minor

packages/marshal/src/marshal.js
dtribble: As above, agreed

packages/marshal/src/marshal.js
erights: Took your next suggestion.

packages/marshal/src/marshal.js
erights: Done

packages/marshal/src/marshal.js
erights: This `rest` is not about the rest of the error object. It is a way for `encodeErrorToSmallcaps` to delegate most of the job to `encodeErrorToCapData`, and then to say (by destructuring and restructuring) "My encoding is just like that encoding except for these differences."

However, thinking about this reveals a bigger problem: the `message`, `rest`, and optional `errorId` property names are fine. Encoding them or not wouldn't make any difference. However, the string value of these properties are not encoded, which is technically "correct" because the decoder doesn't decode them. However, we designed to be more human readable than capData. To someone reading smallcaps, they will habituate to running the decoder in their head. They will do so here as well. Enabling users to provide unencoded strings that might look like encoded strings enables them to mislead readers.

Fixing this should also fix a related fragility: The ability of `encodeErrorToSmallcaps` to delegate some of the encoding to `encodeErrorToCapData` relies on capData *not* encoding these strings to capData, which smallcaps should not be burdened with understanding. Will update soon with a solution to these problems.

erights: Done, by two significant changes:
   * The parts of errors (`name`, `message`, and optional `errorId`) now recursively encoded.
   * To enable this, I changed the interface between marshal and its optional plugin encoders/decoders to take the overall recursive encoder/decoder as an argument.

packages/marshal/src/encodeToSmallcaps.js
erights: Done

erights:
All done. PTAL

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,495 @@
...
+              X`Property name ${q(
+                encodedName,
+              )} of ${encoding} must be a string`,
+            );
+          }
+          !encodedName.startsWith('#') ||
+            assert.fail(
+              X`Unrecognized record type ${q(encodedName)}: ${encoding}`,
+            );
+          const name = decodeFromSmallcaps(encodedName);
```
gibson042, line 479:
```suggestion
          const name = decodeFromSmallcaps(encodedName);
          typeof name === 'string' ||
            assert.fail(
              X`Decoded property name ${name} from ${encoding} must be a string`,
            );
```
This is probably worth a test case as well.

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,495 @@
...
+ *
+ * All other special characters (`"'()*,`) are reserved for future use.
+ *
+ * The manifest constants that smallcaps currently uses for values:
+ *  * `#undefined`
+ *  * `#NaN`
+ *  * `#Infinity`
+ *  * `#-Infinity`
+ *
+ * and for property names:
```
gibson042, line 71:
```suggestion
 * and for property names analogous to capdata @qclass:
```

gibson042 (changes requested):
Looks good except for verification that each decoded object property name is a string.

We should also create a ticket to update [agoric-sdk get-flattened-publication.sh](https://github.com/Agoric/agoric-sdk/blob/HEAD/scripts/get-flattened-publication.sh) for decoding smallcaps after this lands.

packages/marshal/src/encodeToSmallcaps.js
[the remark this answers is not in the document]
gibson042: Do we also want to enforce that `result['#error']` _decodes_ to a string?

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,495 @@
...
+            );
+          ownKeys(rest).length === 0 ||
+            assert.fail(
+              X`#tag record unexpected properties: ${q(ownKeys(rest))}`,
+            );
+          return makeTagged(tag, decodeFromSmallcaps(payload));
+        }
+
+        if (hasOwnPropertyOf(encoding, '#error')) {
+          const result = decodeErrorFromSmallcaps(
```
gibson042, line 453: Should there be an assertion that `encoding` has no extra keys, either here or in `decodeErrorFrom{Smallcaps,CapData}`, to mirror the `ownKeys(rest).length === 0 || assert.fail(…)` line for decoding tagged objects? Not having such a guard seems like a future hazard, although having it means more work if we ever do want to go beyond `name`, `message`, and `errorId` (and we don't currently have such a guard for either tagged objects or errors in decoding capdata).

packages/marshal/src/marshal.js
```diff
@@ -124,42 +136,116 @@ export const makeMarshal = (
...
-    const encodeRemotableToCapData = val => {
-      const iface = getInterfaceOf(val);
-      // console.log(`serializeSlot: ${val}`);
-      return serializeSlot(val, iface);
-    };
+    if (serializeBodyFormat === 'capdata') {
+      const encodeRemotableToCapData = (val, _encodeRecur) => {
+        const iface = getInterfaceOf(val);
+        // console.log(`serializeSlot: ${val}`);
+        return serializeSlot(val, iface);
```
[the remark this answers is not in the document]
erights: Such renamings done as part of your larger suggestions below.

packages/marshal/src/marshal.js
```diff
@@ -111,9 +118,14 @@ export const makeMarshal = (
...
      */
-    const encodeErrorToCapData = err => {
+    const encodeErrorToCapData = (err, encodeRecur) => {
+      // capData doesn't transform strings, so this is for others to
+      // reuse, such as smallcaps.
+      const message = encodeRecur(`${err.message}`);
+      const name = encodeRecur(`${err.name}`);
       // Must encode `cause`, `errors`.
       // nested non-passable errors must be ok from here.
       if (errorTagging === 'on') {
```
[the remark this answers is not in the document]
erights: Subsumed by doing your larger suggestion below.

packages/marshal/src/marshal.js
```diff
@@ -178,66 +264,116 @@ export const makeMarshal = (
...
+    const decodeErrorFromSmallcaps = (encoding, decodeRecur) => {
+      const { '#error': message, ...rest } = encoding;
+      // Not the most elegant way to reuse code. TODO refactor
+      const rawTree = harden({
+        [QCLASS]: 'error',
+        message,
+        ...rest,
+      });
+      return decodeErrorFromCapData(rawTree, decodeRecur);
+    };
```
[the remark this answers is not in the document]
erights: I made the decode side more parallel to the encode side.

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,495 @@
...
+            encodeToSmallcapsRecur(passable[name]),
+          ]),
+        );
+      }
+      case 'copyArray': {
+        return passable.map(encodeToSmallcapsRecur);
+      }
+      case 'tagged': {
+        return {
+          '#tag': getTag(passable),
```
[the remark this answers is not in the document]
erights: The #error property value is the `message`, which was already recursively encoded and decoded, and still is.

But good catch on the #tag. It was not.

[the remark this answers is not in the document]
erights: Done

packages/marshal/src/encodeToSmallcaps.js
erights: FWIW, I added the guard. But I am on the fence as well about whether it is the right decision.

packages/marshal/src/encodeToSmallcaps.js
erights: Done

packages/marshal/src/encodeToSmallcaps.js
erights: Done

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,495 @@
...
+        }
+
+        if (isArray(encoding)) {
+          const result = [];
+          const { length } = encoding;
+          for (let i = 0; i < length; i += 1) {
+            result[i] = decodeFromSmallcaps(encoding[i]);
+          }
+          return result;
+        }
```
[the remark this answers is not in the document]
erights: I have no idea why I didn't do that. Done.

erights:
> Looks good except for verification that each decoded object property name is a string.

Note that `passStyleOf(x) === 'tagged'` and `makeTagged(tag, payload)` both validate that the tag is a string, so marshal doesn't need to.

packages/marshal/src/encodeToSmallcaps.js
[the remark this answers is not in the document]
erights: Done.

erights:
Great suggestions! Done. PTAL.

packages/marshal/src/marshal.js
gibson042: Followup issue: #1295

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,503 @@
...
+        if (
+          typeof result === 'object' &&
+          hasOwnPropertyOf(result, '#error') &&
+          typeof result['#error'] === 'string'
+        ) {
+          return result;
+        }
+        assert.fail(
+          X`internal: Error encoding must have "#error" property: ${result}`,
+        );
```
gibson042, line 268: Consistency with the `encodeToSmallcaps` special case:
```suggestion
        if (typeof result === 'object' && hasOwnPropertyOf(result, '#error')) {
          const message = result['#error'];
          if (
            typeof message === 'string' &&
            // check that it decodes to a string
            (!startsSpecial(message) || message.startsWith('!'))
          ) {
            return result;
          }
          assert.fail(
            X`internal: Error encoding must string message: ${q(message)}`,
          );
        }
        assert.fail(
          X`internal: Error encoding must have "#error" property: ${q(result)}`,
        );
```
Also, when is `q` appropriate? Some calls to `X` use it but others don't... should it be limited to values that are expected to be strings?

packages/marshal/src/marshal.js
```diff
@@ -124,49 +123,130 @@ export const makeMarshal = (
...
+        body,
+        slots,
+      });
+    } else if (serializeBodyFormat === 'smallcaps') {
+      /**
+       * @param {string} prefix
+       * @param {Passable} passable
+       * @param {string} [iface]
+       */
+      const serializeSlotToSmallcaps = (
```
gibson042, line 191: Consistency with `encodeSlotToCapData`:
```suggestion
      /**
       * @param {string} prefix
       * @param {Passable} passable
       * @param {InterfaceSpec} [iface]
       * @returns {string}
       */
      const serializeSlotToSmallcaps = (
```

packages/marshal/src/marshal.js
```diff
@@ -176,68 +256,111 @@ export const makeMarshal = (
...
-        'string',
-        X`invalid error message typeof ${q(typeof message)}`,
-      );
-      const EC = getErrorConstructor(`${name}`) || Error;
+    /**
+     * @param {{errorId?: string, message: string, name: string}} errData
+     * @param {(e: unknown) => Passable} decodeRecur
+     * @returns {Error}
+     */
+    const decodeErrorCommon = (errData, decodeRecur) => {
```
gibson042, line 266: Should `decodeRecur` have a default value so `decodeErrorFromCapData` can ignore its second argument like `decodeRemotableOrPromiseFromCapData` does?
```suggestion
     * @param {{errorId?: string, message: string, name: string}} errData
     * @param {(e: unknown) => Passable} [decodeRecur]
     * @returns {Error}
     */
    const decodeErrorCommon = (errData, decodeRecur = val => val) => {
```

packages/marshal/src/marshal.js
```diff
@@ -176,68 +256,111 @@ export const makeMarshal = (
...
+        const iface = i < 0 ? undefined : stringEncoding.slice(i + 1);
+        return decodeSlotCommon({ iface, index });
+      };
+    };
+    const decodeRemotableFromSmallcaps = makeDecodeSlotFromSmallcaps('$');
+    const decodePromiseFromSmallcaps = makeDecodeSlotFromSmallcaps('&');
+
+    const decodeErrorFromSmallcaps = (encoding, decodeRecur) => {
+      const { '#error': message, ...restErrData } = encoding;
+      return decodeErrorFromCapData({ message, ...restErrData }, decodeRecur);
```
gibson042, line 327:
```suggestion
      !hasOwnPropertyOf(restErrData, 'message') ||
        assert.fail(X`unexpected encoded error property ${q('message')}`);
      return decodeErrorCommon({ message, ...restErrData }, decodeRecur);
```

gibson042 (approved):
A few final nits, but I trust your judgment on all of them. Looks like this is it!

packages/marshal/src/marshal.js
erights: Done. Also renamed `serializeSlotToSmallcaps` to `encodeSlotToSmallcaps`

packages/marshal/src/marshal.js
erights: Good catch! I completely missed that `restErrData` might have a `message` property that would then overwrite the message text obtained from `#error`. Done.

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -0,0 +1,503 @@
...
+          // preserving.
+          return `!${passable}`;
+        }
+        // All other strings pass through to JSON
+        return passable;
+      }
+      case 'symbol': {
+        // At the smallcaps level, we prefix symbol names with `%`.
+        // By "symbol name", we mean according to `nameForPassableSymbol`
+        // which does some further escaping. See comment on that function.
```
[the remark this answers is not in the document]
erights: I thought about this and I am not inclined to do it. From talking to Spritely about what concepts are sufficiently language neutral to include in this level of protocol, to facility cross-language interoperation, we can agree that there are strings and symbols, and that symbols have a string name, such that two symbols with the same string name are the same symbol. But registered vs well-known seems too specific to JS to include in the definition of this level of the protocol. Instead, we keep the mapping from strings to passable symbols a separate concern encapsulated by `nameForPassableSymbol` and its inverse.

Resolving.

packages/marshal/src/encodeToSmallcaps.js
erights: Nice symmetry. Done.

Using `q` or not expresses a security policy. Within a details (`X`) substitution, the contents are normally redacted, because the default assumption is that it may convey detailed data that is useful for debugging, but inappropriate to leak to callers. But when the information is relatively static an unlikely to convey a secret that callers should not see, use `q` to disclose it rather than redact it. 

A recent example illustrates this well: Redact amounts because they contain quantities. But brands and keywords should both be disclosed using `q`.

packages/marshal/src/marshal.js
erights: No. Slots are low enough level that not passing the *recur on to the common function is fine. Error handling should not encourage that assumption in too many places. I'd rather make the callers of the error function pass the *recur function explicitly. This also prepares us better to extend the error common functions to handle `cause` and `errors`.

Resolving.

PR #1283, opened 2022-09-12
WIP: smallcaps, see what breaks

erights:
Starts from https://github.com/endojs/endo/pull/1282 by changing only the default choice from the capData encoding to the smallcaps encoding, in order to see what breaks under CI.


erights:
Closing as this diagnostic is no longer how we'll probe the issue

PR #881, opened 2021-09-08
add __makeScopeProxy__ to Compartment.prototype

kumavis:
this allows the ses shim consumer to construct compartments normally and then run code wrapped in the magic lines of code so that they execute as if they were in the compartment but without use of `eval`. This enables ses/endo like behavior in environments where eval is not allowed such a restrictive CSP environments like "WebExtension Manifest v3 Background Service Workers"

prolly some better factoring to be had here

PR #1285, opened 2022-09-13
feat: expose scope proxy for lavamoat

naugtur:
# Not intended for merging into master. 
 
 - based on #881 but refactored for simplicity of updates from upstream
 - will no longer be needed once someone lands a quadruple backflip, and I don't mean https://www.youtube.com/watch?v=OFPmjxkNRmA


kumavis:
thanks. this makes it possible to do the non-eval "templated Compartment eval" / "precompiled module" thing that we need to do for metamask

`__makeScopeProxy__` and `__isKnownScopeProxy__` are to be removed with the introduction of the "quadruple backflip" refactor

naugtur:
no longer needed :tada:

PR #1287, opened 2022-09-14
chore: release

kriskowal:
- chore: lerna version
- chore: Update release notes


dckc:
This morning I noticed that https://github.com/endojs/endo/releases thinks the latest release was March 2020.

dckc (approved):
I gather this is entirely mechanical.

kriskowal:
> This morning I noticed that https://github.com/endojs/endo/releases thinks the latest release was March 2020.

I don’t even recall the process we had in 2020 to publish these two releases. It is probably better for me to remove them to give a false sense that we use this system.

PR #1288, opened 2022-09-16
feat(ses): uncurryThis uses bind

mhofman:
From micro benchmarks, it looks like the bind form is 2x on v8 JIT, 1.5x on v8 jitless, and 1.3x on xsnap.

While looking at this I found a bug in the type definition, which when fixing, uncovered a couple real bugs (including one in the unhandled rejection logic).

### DIFF
--- a/packages/ses/src/error/unhandled-rejection.js
+++ b/packages/ses/src/error/unhandled-rejection.js
@@ -63,7 +63,7 @@ export const makeRejectionHandlers = reportReason => {
    * @param {ReasonId} heldReasonId
    */
   const finalizeDroppedPromise = heldReasonId => {
-    if (mapHas(idToReason)) {
+    if (mapHas(idToReason, heldReasonId)) {
       const reason = mapGet(idToReason, heldReasonId);
       removeReasonId(heldReasonId);
       reportReason(reason);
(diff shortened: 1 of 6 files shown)


kriskowal (approved):
Please add a note to NEWS.md for any user-facing effects under the heading `# Next release`

mhofman:
@kriskowal do you believe perf improvement is a user facing constraint that warrants a news item, or did you mean the unhandled rejection promise?

kriskowal:
The unhandled rejection fix is newsworthy.

packages/ses/src/error/unhandled-rejection.js
```diff
@@ -63,7 +63,7 @@ export const makeRejectionHandlers = reportReason => {
    * @param {ReasonId} heldReasonId
    */
   const finalizeDroppedPromise = heldReasonId => {
-    if (mapHas(idToReason)) {
+    if (mapHas(idToReason, heldReasonId)) {
```
erights, line 66: That's quite a bug!

Why did the old code seem to work?

erights (approved):
LGTM, thanks!

packages/ses/src/error/unhandled-rejection.js
mhofman: Old code probably didn't when the promise would get collected. `map.has(undefined)` was always `false` so we never reported then. However we also had a report path on exit.

mhofman:
PTAL. I will run this with agoric-sdk see if there are any unintended effects.

PR #1290, opened 2022-09-17
fix: add a do-nothing SharedError.prepareStackTrace

erights:
This PR does what I thought I understood to solve your `depd` problem. Does it?

If so, perhaps you should rebase https://github.com/endojs/endo/pull/1286 on this and adapt your test so it passes. At the moment, this PR does not include any test for the added property, so your test would be most welcome

If not, please help me correct my misunderstanding.

### DIFF
--- a/packages/ses/src/error/tame-error-constructor.js
+++ b/packages/ses/src/error/tame-error-constructor.js
@@ -124,12 +124,31 @@ export default function tameErrorConstructor(
   // The default SharedError much be completely powerless even on v8,
   // so the lenient `stackTraceLimit` accessor does nothing on all
   // platforms.
+  //
+  // `SharedError.prepareStackTrace`, if it exists, must also be
+  // powerless. However, from what we've heard, depd expects to be able to
+  // assign to it without the assignment throwing. It is normally a function
+  // that returns a stack string to be magically added to error objects.
+  // However, as long as we're adding a lenient standin, we may as well
+  // accommodate any who expect to get a function they can call and get
+  // a string back. This prepareStackTrace is a do-nothing function that
+  // always returns the empty string.
   defineProperties(SharedError, {
     stackTraceLimit: {
       get() {
         return undefined;
       },
       set(_newLimit) {
+        // do nothing
+      },
+      enumerable: false,
+      configurable: true,
+    },
+    prepareStackTrace: {
+      get() {
+        return () => '';
+      },
+      set(_prepareFn) {
         // do nothing
       },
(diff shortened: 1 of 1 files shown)


erights (commented):
(I'd request this change of myself if github allowed it)

One thing that's too weird about this PR as is, is that the do-nothing `SharedError.prepareStackTrace` would appear on all platforms, but `InitialError.prepareStackTrace` would not exist on non-v8 platforms. This needs to be more consistent, whether additively or subtractively. Waiting to understand the `depd` problem better before deciding.

naugtur:
I've pushed some tests, but they're failing on `Error.captureStackTrace is not a function` now. Would need to address that and I don't yet understand why it's not there.

erights:
> I've pushed some tests, but they're failing on `Error.captureStackTrace is not a function` now. Would need to address that and I don't yet understand why it's not there.

It was not there because I had not understood that you needed it. It's there now.

See the changes I made to your test cases. Your first two passed, but I made them more precise. Your third one is impossible with a powerless Error object. But by creating a compartment endowed with the start compartment's powerful Error object, you should be able to do anything you can do with it in the start compartment.

In the endowed test case, `captureStackTrace` does work, but I could not make the substitute `prepareStackTrace` do anything. This is odd because the immediately following test case does, in order to support some depd functionality! There's something about it that I'm missing.

PTAL

erights:
The two new emulated properties:
   * `SharedError.prepareStackTrace` accessor for silently ignoring assignment
   * `SharedError.captureStackTrace` for assigning an empty string to its first argument.stack

are now added to SharedError only on v8. If we encounter code assuming these on other platforms, we can revisit.

The `SharedError.stackTraceLimit` accessor for silently ignoring assignment remains available everywhere. Due to Google's bad advice, this one does need to be available on all platforms.

naugtur:
In practice the assumption `prepareStackTrace` can only be available for writing in v8 may not be enough for Endo in particular, because the only way to invoke the version of `depd` that doesn't attempt setting it is using a bundler that prioritizes package.json's "browser" field to look for the module main. That's where `depd` exposes a compatible API that does nothing.

If we run codebase with depd in Endo we'll get into trouble again. For depd in particular a simple patch-package change to the package.json to define an export for Endo would do. Not sure about other packages.

Anyway, this is now a great improvement to ecosystem compatibility.

naugtur (approved):

mhofman (approved):
The approach seems sane

PR #1291, opened 2022-09-21
fix: tools arb passable

erights:
Follows https://github.com/Agoric/agoric-sdk/pull/6296 , to move those enhancements to test-rankOrder to endo as well.

### DIFF
--- a/packages/marshal/test/test-encodePassable.js
+++ b/packages/marshal/test/test-encodePassable.js
@@ -11,35 +11,75 @@ import {
 } from '../src/encodePassable.js';
 import { compareRank, makeComparatorKit } from '../src/rankOrder.js';
 import { sample } from './test-rankOrder.js';
+import { arbPassable } from '../tools/arb-passable.js';
 
-const { Fail } = assert;
+const { Fail, quote: q } = assert;
 
-const r2e = new Map();
-const e2r = [];
-
-const encodeRemotable = r => {
-  if (r2e.has(r)) {
-    return r2e.get(r);
-  }
-  const result = `r${e2r.length}`;
-  r2e.set(r, result);
-  e2r.push(r);
-  return result;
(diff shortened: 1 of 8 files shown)


kriskowal:
I presume `s/arb/are` in the PR title and branch name, but I’m prepared to be surprised by a new acronym!

erights:
It is fast-check speak for "arbitrary"

kriskowal:
I was almost as prepared for a new abbreviation! Thanks.

gibson042 (approved):
LGTM!

packages/marshal/test/test-encodePassable.js
```diff
@@ -0,0 +1,156 @@
...
+  const encComp = compareRank(ex, ey);
+  if (fullComp !== 0) {
+    t.is(encComp, fullComp);
+  }
+};
+
+test('order invariants', t => {
+  for (let i = 0; i < sample.length; i += 1) {
+    for (let j = i; j < sample.length; j += 1) {
+      orderInvariants(t, sample[i], sample[j]);
```
erights, line 122: I postponed addressing https://github.com/endojs/endo/pull/1260#discussion_r1003711830 from #1290 to this PR, since it was about usage of fast-check. So this is a note to address it here.

erights: At [xxx](https://github.com/endojs/endo/pull/1260#discussion_r1003711830) @gibson042 wrote

> Shouldn't this be structured like `await fc.assert(fc.property(…))` rather than using a direct list?

erights: In discussion, we realized this suggestion is already covered by the last test case. However, this non-fuzzing version still serves a purpose. It ensures that at least these always get tested. Moved and added explanation. Resolving.

Issue #912, opened 2021-09-30
Module shims lexicals leak internal live bindings object


mhofman:
# Description of the problem

Dynamic lexicals are used to emulate live bindings of transformed modules. From a high level, a transformed module has
- its `import`s rewritten into a `let` binding with an updater registered the get notified of value changes
- `export`s trigger the notifier when first initialized
- live binding exports (e.g. `let`) rewrite the initial identifier and setup a scope lexical object property with `get` and `set` functions to return or update the current value of the export.

The `get` is used when a module reads the value of the exported binding, which may be any value, primitive or not, generated by the local module or not. The case of interest here for this issue is exported live bindings which values are functions generated by another module.

A consequence of https://github.com/endojs/endo/issues/31 is that when an exported function is called, it would receive the caller module's scope object as its `this`, which has all kind of interesting powers. Amongst others, the called function would gain direct and non-lexical access to set and get the calling module's exported bindings.

Even with the proposed split `with` (https://github.com/endojs/endo/issues/905), this issue wouldn't go away, as its an intrinsic consequence of using `with` blocks. It actually would become worse if we used a plain object for the `dynamicGlobalLexicals` as it would allow the function to list all the properties representing the exported live bindings of the caller.

The call from a `with` environment is actually problematic because of the [`WithBaseObject` logic](https://tc39.es/ecma262/#sec-object-environment-records-withbaseobject) in [`EvaluateCall`](https://tc39.es/ecma262/#sec-evaluatecall). The `get` property operation performed on the scope object is not itself problematic, but because of `with`, whatever value is returned will be called with the scope object as `this`. A spec fix in the form of [an `@@hiddenWith` symbol](https://github.com/endojs/endo/issues/31#issuecomment-546662631) would be great, but the problem can be solved differently in this case.

## Example

```js
// a.js
function doEvil() {
  // Do something terrible
}

export function foo() {
  if (this) {
    this.foo = doEvil;
  }
}

// b.js
import {foo as aFoo} from './a.js';

export let foo;

export function setup(...args) {
  foo = aFoo;
  foo(...args); // leaks our scope object to `a.js#foo`
}
```

# Impact

The danger is pretty esoteric. Only code following a fairly non common pattern would be impacted:
- Assign an untrusted function declared in another module to an exported live binding
- Call that function using the exported lexical binding

In another scenario, a module could use this mechanism to escape any syntactic audit that its exports are only modified lexically (aka following specified means).

# Mitigations

## Linting

[...]

Issue #490, opened 2020-10-12
How do we wire up a panic() at the SES level.


erights:
At https://github.com/Agoric/SES-shim/pull/447/files/b24a650a1bebec1c8a0b9abfad7690becb4f8d50#diff-03e7a59379df2f84382dd5a0b76e3d76 we have an example where we need a stronger `assert`, that immediately terminates the [unit of preemptive termination](https://github.com/tc39/proposal-oom-fails-fast), which for our usage is almost certainly the vat.

At https://github.com/Agoric/SES-shim/blob/b24a650a1bebec1c8a0b9abfad7690becb4f8d50/packages/ses/src/error/assert.js#L378-L387 is a commented out implementation of `insist` which, if made from a capability for terminating this unit, is such a more powerful `assert`.

On SwingSet we can wire up a capability for sudden vat termination. However, what is a host-neutral mechanism?

Btw, the contrasting terms `assert` vs `insist` don't work as the labels for their respective sides of the distinction. If one didn't know other JS `assert` libraries, then from the two names, once would guess they had the swapped meanings: That `insist` is doing input validation, in which case throwing on failure makes sense. The first widespread use of `assert` in a programming language that I am aware of is C's stdlib, where failure terminates the process. But we cannot swap the names, because entrenched JS expectations are that `assert` throws on failure. So we need at least a new name for `insist`.

erights:
I think the most that SES itself *can* provide for this is what it does provide (or will as of the next release) -- a `makeAssert` function for making scoped asserts for shutting down larger units. I do have an active PR in progress that uses it to create a `zcf.assert` so a contract can fail-fast when it thinks it is confused. So, at the SES level, this is done. Closing.

mhofman:
Let's provide a `panic()` shim, which uses platform functions or falls-back to an infinite loop.

PR #1293, opened 2022-09-21
refactor(ses)!: Divide safe evaluator scope stack into four layers

kumavis:
#### description
instead of `with(complicatedScopeProxy){ ... }`, we now have the following scope stack (outermost to innermost, lowest takes precedence):
- `scopeTerminator` {**security**, **fidelity**} (always return undefined, if `sloppyGlobalsMode` is true, sets happen on `globalObject`)
- `globalObject` {**fidelity**} (bare)
- `globalLexicals` {**fidelity**} (bare)
- `evalScope` {**security**} (obj with `FERAL_EVAL` exposed as `eval` only once)

#### behavioural changes
this refactor introduces some changed behavior:
- `globalLexicals` object can be leaked if a special lexical is crafted (should not be craftable by evaluated code)
- setting a correctly formatted `Symbol.unscopables` property on the evaluator globalObject will cause the globals configured by `Symbol.unscopables` to not be in scope. this is a shim fidelity issue and not a security issue as the `scopeTerminator` is not affected.
- fixes a bug where sloppyGlobal assignment is allowed in non-sloppyGlobalsMode when a property does not exist on the compartment but does exist on the realm globalThis

#### other notes
- attempts to adapt `test-scope-handler` test to use `makeSafeEvaluator` and test the whole scope stack, but its not a great match and a lot had to be changed
- adds a failing test `test-scope-handler-pollution` because the test's prototype pollution broke the `makeSafeEvaluator` internals
- changed a couple tests to match behavioral changes

#### things to revisit after this PR
- [ ] leak of presence of properties on globalThis with strict-scope-terminator https://github.com/endojs/endo/issues/1305
- [ ] consider adding another lexical layer to differentiate between `moduleScope` and `globalLexicals` to limit impact of a leak https://github.com/endojs/endo/issues/912


packages/ses/src/make-safe-evaluator.js
```diff
@@ -56,7 +56,7 @@ export const makeSafeEvaluator = ({
     if (!evaluate) {
       const constants = getScopeConstants(globalObject, globalLexicals);
       const evaluateFactory = makeEvaluateFactory(constants);
-      evaluate = apply(evaluateFactory, scopeProxy, []);
+      evaluate = apply(evaluateFactory, { scopeProxy, evalScope }, []);
```
kumavis, line 59: add a freeze just for fun?

kriskowal: For sure.

packages/ses/test/test-scope-handler.js
```diff
@@ -11,7 +11,7 @@ import { createScopeHandler } from '../src/scope-handler.js';
 const FERAL_EVAL = eval;
 
 test('scopeHandler - has trap', t => {
-  t.plan(7);
+  t.plan(8);
```
kriskowal, line 14: almost there…

packages/ses/src/scope-terminator.js
```diff
@@ -0,0 +1,103 @@
...
+      return undefined;
+    },
+
+    set(_shadow, prop, value) {
+      if (sloppyGlobalsMode) {
+        // Redirect set properties to the globalObject.
+        return reflectSet(globalObject, prop, value);
+      } else {
+        return false;
+      }
```
mhofman, line 57: The fact this check didn't exist before seems like a bug to me, but I don't have enough historical background.

mhofman (commented):
Some preliminary observations.

I'm having a hard time reviewing the tests because of the combined rename + modifications past the diff threshold. Would appreciate a minimal rename and separate modification to the tests

kumavis:
my biggest regret is that the diff is a bit difficult to review,  but I think the result is easy to review and audit

@mhofman the `test-scope.js` file (previously `test-scope-handler.js`) was changed pretty dramatically bc it tested behavior about the full scope stack but in very low level ways. I did my best to rework the tests against the `safeEvaluator` level but observing scope behavior at this level meant different checks and results. The new test file is somewhat haphazard and very different from the original so reviewing a diff from the previous test is perhaps not particularly useful. However I still thought having a test tracking the behavior was useful. Additionally all observed changes to behavior  are enumerated in the top comment of this pr.

packages/ses/src/scope-terminator.js
kumavis: Yeah, same suspicion

packages/ses/src/make-safe-evaluator.js
```diff
@@ -56,7 +47,24 @@ export const makeSafeEvaluator = ({
...
       const constants = getScopeConstants(globalObject, globalLexicals);
       const evaluateFactory = makeEvaluateFactory(constants);
-      evaluate = apply(evaluateFactory, { scopeProxy, evalScope }, []);
+      const optimizerObject = defineProperties(
+        create(null),
+        assign(
+          getOwnPropertyDescriptors(globalObject),
+          getOwnPropertyDescriptors(globalLexicals),
+        ),
+      );
```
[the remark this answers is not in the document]
kumavis: > It looks like currently getScopeConstants excludes any prop from globalLexicals

that doesnt seem correct. it returns optimizer names from both
```
[...globalConstants, ...localConstants]
```

packages/ses/src/scope-terminator.js
kumavis: ok so its about the combination of the `has` and `set` implementation.

if `has` returns false, it will not hit the `set` hook and instead just naturally throw a `ReferenceError` (if it doesnt exist on realm globalThis).

if `has` returns true then we hit `set` where we [previously](https://github.com/endojs/endo/blob/705aef24f34bb9794f0aa807d567b3efbf0c23af/packages/ses/src/scope-handler.js#L128) just forwarded the set to the `globalObject`. in this refactor, we only hit `scopeTerminator` set hook if the property is missing from the globalObject at which point we should deny the set / throw a ReferenceError (strict mode) or redirect the set to the globalObject (sloppyGlobals)

kumavis:
split the `scopeTerminator` into a reuseable strict and per-evaluator sloppyGlobals version

packages/ses/src/make-safe-evaluator.js
```diff
@@ -76,7 +86,11 @@ export const makeSafeEvaluator = ({
       mandatoryTransforms,
     ]);
 
-    scopeController.allowNextEvalToBeUnsafe = true;
+    // Allow next reference to eval produce the unsafe FERAL_EVAL.
+    // We avoid defineProperty because it consumes an extra stack frame taming
+    // its return value.
+    defineProperties(evalScope, oneTimeEvalProperties);
```
naugtur, line 92: Does it need to happen here? What if `createEvalScope`  returned the scope and a function to call if you want a new eval? 

Also, I wonder if deleting the eval and defining it again toggles the object between two shapes (aka hidden classes) or nests them further with each operation.

packages/ses/src/strict-scope-terminator.js
```diff
@@ -0,0 +1,92 @@
...
+
+  set(_shadow, prop, _value) {
+    // we should only hit this if the has() hook returned true
+    // matches the v8 ReferenceError message
+    // "Uncaught ReferenceError: xyz is not defined"
+    throw new ReferenceError(`${String(prop)} is not defined`);
+  },
+
+  has(_shadow, prop) {
+    // we must at least return true for all properties on the realm globalThis
```
naugtur, line 54: Is that still true though? 
All of them are already handled in the previous backflip. 
This is really only needed for sloppy mode, which could be overridden in there anyway, right? 
Strict scope terminator shouldn't need a reference to any globals I think.

packages/ses/test/test-confinement.js
```diff
@@ -72,9 +72,10 @@ test('confinement evaluation Symbol.unscopables with-statement escape', t => {
   const c = new Compartment({ flag: 'safe' });
 
   t.is(c.evaluate('Symbol.unscopables'), Symbol.unscopables);
+  // this modification causes a loss of shim fidelity, but not a loss of security
   t.is(
     c.evaluate('globalThis[Symbol.unscopables] = { flag: true }; flag'),
-    'safe',
+    undefined,
```
naugtur, line 78: What's preventing us from adding a noop setter on `globalThis[Symbol.unscopables] `?
Would that be too much intervention?

kumavis: hmm we could do that

packages/ses/src/strict-scope-terminator.js
kumavis: still true. in order to block access to the **realm globalThis** properties we need to return true in this has hook

packages/ses/src/make-safe-evaluator.js
kumavis: > What if createEvalScope returned the scope and a function to call if you want a new eval?

i would prefer that approach for legibility but it is already intentionally not a function https://github.com/endojs/endo/issues/956

kumavis: > Also, I wonder if deleting the eval and defining it again toggles the object between two shapes (aka hidden classes) or nests them further with each operation.

idk what this means but it sounds interesting

kriskowal: I know about hidden classes, and am also curious. You’d have to inspect disassembly to know for sure. My suspicion is that eval scope simply doesn’t benefit from hidden classes. I don’t believe that deletion can cause a return to a previous hidden class.

@mhofman has an idea for using Symbol.unscopables instead of this defineProperties mechanism. It would be delightful if we could find a mechanism that doesn’t bump the stack. That might be more likely with simple assignment and deletion on the unscopables object. Do `with` blocks and Symbol.unscopables interact in a way that bumps the stack?

packages/ses/src/strict-scope-terminator.js
kriskowal: Herein lies the leak of information about the properties of the underlying realm’s globalThis, which we found to be less bad than breaking `typeof`. I would feel most confident about our security claims if this simply returned `true`. I like that this new formulation of the scope proxy makes this compromise more obvious!

packages/ses/test/test-confinement.js
kriskowal: I think it’s worth the loss in fidelity to prevent leaking the terminal scope proxy. What do you think, @erights?

kumavis: no known way to leak the terminal scope proxy

packages/ses/src/strict-scope-terminator.js
kumavis: yes! i think we can `return true` here to no change in behavior

packages/ses/test/test-confinement.js
erights: What loss of fidelity are we talking about? The 
> noop setter on `globalThis[Symbol.unscopables] `?

?
And in exchange we would no longer leak a scope proxy? Yes, worth it!

kumavis:
> I'm having a hard time reviewing the tests because of the combined rename + modifications past the diff threshold. Would appreciate a minimal rename and separate modification to the tests

@mhofman makes sense - though im not sure how to accomplish this. can someone provide some suggestions on how to improve the legibility of the diff? my gitfoo is not amazing. and i dont think its possible to make a minimal modification to the `test-scope.js` tests

packages/ses/src/make-safe-evaluator.js
kumavis: > Also, I wonder if deleting the eval and defining it again toggles the object between two shapes (aka hidden classes) or nests them further with each operation.

is this a performance concern?

packages/ses/test/test-confinement.js
kumavis: there is no known way to leak a scope proxy in this PR's implementation

packages/ses/src/make-safe-evaluator.js
kriskowal: Yes. Hidden classes are a performance concern specific to V8.

Using unscopables on `evalScope` may also frustrate “stack bumping attacks” (a name inspired by “lock bumping” I have made up for when an attacker prepares for a stack overflow at a specific call depth during a synchronous call to host behavior in hopes of forcing it down an exception path that leaves the underlying API in a vulnerable state)

kriskowal:
@kumavis One way to satisfy @mhofman’s concern would be to either create or refer to tests that cover all the security behaviors of the collective scopes that pass both before and after this refactor, since the unit tests cut too closely to the particulars of the implementations before and after. If that requires new tests, we could land them in another PR before so we’re sure they passed before this change, and also highlight any deliberate behavior changes in this PR.

packages/ses/src/make-safe-evaluator.js
kumavis: would a try-catch gain us any leverage?

kriskowal: In this specific case, perhaps not. We `delete evalScope.eval` in the `finally` block around our use of `safeEval`, which is safe against stack bumping. If this `defineProperties` call fails, the worst it can do is succeed.

packages/ses/src/scope-terminator.js
kumavis: added a test for this and recorded in pr top comment

packages/ses/test/test-scope.js
```diff
@@ -329,12 +325,15 @@ test('scope behavior - Symbol.unscopables fidelity test', t => {
     globalObject,
   });
 
-  t.is(evaluate('typeof localProp'), 'object');
+  t.is(evaluate('typeof localProp'), 'undefined');
```
kumavis, line 328:
```suggestion
  // Known compromise in fidelity of the emulated script environment:
  t.is(evaluate('typeof localProp'), 'undefined');
```

packages/ses/test/test-scope.js
```diff
@@ -370,19 +373,27 @@ test('scope behavior - Symbol.unscopables fidelity test', t => {
   );
   // after property is created on globalObject, assignment is evaluated to
   // test if it is affected by the Symbol.unscopables configuration
+  // Known compromise in fidelity of the emulated script environment:
   globalObject.eventuallyAssignedLocalProp = null;
-  evaluate('eventuallyAssignedLocalProp = {}');
+  t.throws(() => evaluate('eventuallyAssignedLocalProp = {}'), {
```
kumavis, line 378:
```suggestion
  // Known compromise in fidelity of the emulated script environment:
  t.throws(() => evaluate('eventuallyAssignedLocalProp = {}'), {
```

packages/ses/test/test-scope.js
mhofman, line 329: Please add a comment this is a fidelity compromise

packages/ses/test/test-scope.js
```diff
@@ -370,19 +373,27 @@ test('scope behavior - Symbol.unscopables fidelity test', t => {
   );
   // after property is created on globalObject, assignment is evaluated to
   // test if it is affected by the Symbol.unscopables configuration
+  // Known compromise in fidelity of the emulated script environment:
```
mhofman, line 377: How is it a fidelity compromise in this case? This would only be possible in sloppy mode, no?

packages/ses/src/make-safe-evaluator.js
```diff
@@ -56,7 +49,24 @@ export const makeSafeEvaluator = ({
     if (!evaluate) {
       const constants = getScopeConstants(globalObject, globalLexicals);
       const evaluateFactory = makeEvaluateFactory(constants);
-      evaluate = apply(evaluateFactory, scopeProxy, []);
+      const optimizerObject = defineProperties(
```
kumavis, line 52: no doubt theres a more efficient way to create the optimizer object

packages/ses/test/test-scope.js
kumavis: seems like this comment was added here by accident

packages/ses/test/test-scope.js
```diff
@@ -370,19 +374,28 @@ test('scope behavior - Symbol.unscopables fidelity test', t => {
...
```
kumavis, line 377:
```suggestion
```

packages/ses/src/strict-scope-terminator.js
kumavis: this breaks a lot of tests, including tests outside of ses
> 17 tests failed
for example where it previously threw an error
```js
t.throws(() => c.evaluate('XYZ'), { instanceOf: ReferenceError });
```
it now just returns `undefined`
```js
t.is(c.evaluate('XYZ'), undefined);
```

i think we should consider this change outside of this PR

mhofman (commented):
Intermediate review

packages/ses/src/make-evaluate-factory.js
```diff
@@ -60,12 +60,18 @@ export const makeEvaluateFactory = (constants = []) => {
...
   return FERAL_FUNCTION(`
-    with (this) {
-      ${optimizer}
-      return function() {
-        'use strict';
-        return eval(arguments[0]);
-      };
+    with (this.scopeTerminator) {
+      with (this.globalObject) {
+        with (this.globalLexicals) {
```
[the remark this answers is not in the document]
kumavis: we're trying to determine if this PR makes the [module lexicals](https://github.com/endojs/endo/issues/912) leak worse. if it doesnt, i would consider it out of scope of this pr and revisit it separately

my understanding is it does not make it worse

kriskowal:
@mhofman @erights I’d like to land this without the fifth with block, without fixing the global lexicals leak (#912), and then proceed to work on that fix myself after this has landed.

packages/ses/src/make-evaluate-factory.js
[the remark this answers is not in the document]
erights: > my understanding is it does not make it worse

Mine too.

erights:
> @mhofman @erights I’d like to land this without the fifth with block, without fixing the global lexicals leak (#912), and then proceed to work on that fix myself after this has landed.

Fine with me.

mhofman:
Since the leak is no worse than what we have today, find by me too.

packages/ses/test/test-make-eval-function.js
```diff
@@ -21,7 +21,7 @@ test('makeEvalFunction - leak', t => {
   t.is(safeEval('none'), undefined);
   t.is(safeEval('this.none'), undefined);
 
-  safeEval('none = 8');
+  safeEval('this.none = 8');
 
   t.is(safeEval('none'), 8);
   t.is(safeEval('this.none'), 8);
```
kriskowal, line 27: If assignment to the global as a free variable that is an existing property of the global object isn’t supposed to work, these three lines can probably be deleted since they’re redundant with the next three.

@erights Was this test valid previously? @kumavis Was this changed because the refactor made this behavior throw?

packages/ses/src/strict-scope-terminator.js
kriskowal: Agreed. We’ll revisit this behavior change. #1305

packages/ses/src/eval-scope.js
```diff
@@ -0,0 +1,25 @@
...
+    },
+  });
+
+  return {
+    evalScope,
+    allowNextEvalToBeUnsafe: () => {
+      // Allow next reference to eval produce the unsafe FERAL_EVAL.
+      // We avoid defineProperty because it consumes an extra stack frame taming
+      // its return value.
+      defineProperties(evalScope, oneTimeEvalProperties);
```
naugtur, line 22: I moved the logic here so it's easier to follow how and when it gets exposed.

kriskowal:
@erights @mhofman This should be ready for review again. We bifurcated the optimizer at today’s SES meeting and the remaining follow-ups have tracking issues.

packages/ses/test/test-make-eval-function.js
kumavis: line 24 was indeed changed because in this version it throws. it is a sloppyGlobals assignment. the previous version accidently allowed sloppyGlobals if the property was on the realm globalThis.

this change in behavior is recorded in the PR top comment

packages/ses/src/eval-scope.js
```diff
@@ -0,0 +1,20 @@
+import { FERAL_EVAL, create, freeze } from './commons.js';
+
+export const createEvalScope = () => {
+  const evalScope = create(null);
+  const oneTimeEvalProperties = freeze({
+    eval: {
+      get() {
+        delete evalScope.eval;
```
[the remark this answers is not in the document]
kumavis: this is now done locally

packages/ses/src/scope-terminator.js
```diff
@@ -0,0 +1,103 @@
...
+ * createScopeTerminator()
+ * scopeTerminatorHandler manages a scopeTerminator Proxy which serves as
+ * the final scope boundary that will always return "undefined" in order
+ * to prevent access to "start compartment globals". When "sloppyGlobalsMode"
+ * is true, the Proxy will perform sets on the "globalObject".
+ */
+export const createScopeTerminator = (
+  globalObject,
+  { sloppyGlobalsMode = false } = {},
+) => {
```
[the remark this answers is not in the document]
kumavis: this change was made

erights (commented):
TIL "apotheosis"

erights:
> adds a failing test `test-scope-handler-pollution` because the test's prototype pollution broke the `makeSafeEvaluator` internals

This sounds like a severe bug and a possible vulnerability to code (like vetted shims) that runs between repair and lockdown. Is it?

packages/ses/src/scope-terminator.js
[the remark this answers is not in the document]
erights: Nice!

packages/ses/test/test-confinement.js
erights: I reread the thread and understand that my previous response did not make sense.

I think the fidelity issue here cuts both ways, with no choice being strictly better fidelity than the other. The fidelity issue is also minor, as it is unlikely to be encountered by accident, and it is unlikely to be exploitable by an attacker. When faced with such a choice, the next question is: ***Which lack of fidelity fails safe?***

I think giving the global object a `Symbol.unscopables` non-configurable accessor without a setter, or whose setter always throws, means that any program that does not cause an error to be thrown will work the same under the shim and under a native implementation.

Unresolving.

packages/ses/src/make-evaluate-factory.js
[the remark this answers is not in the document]
erights: Is there an issue so we don't forget to get back to this? Unresolving only pending the issue link.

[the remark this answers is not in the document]
erights: Ah. That seems to be #912 cited immediately below. Reresolving.

packages/ses/test/test-confinement.js
mhofman: If we decide to install a `Symbol.unscopables` on the `globalThis`, it should be a configurable (to allow a program to change it through `defineProperty` if really needed) inert accessor (to avoid assignment errors).

packages/ses/src/strict-scope-terminator.js
```diff
@@ -0,0 +1,92 @@
...
+  create(
+    alwaysThrowHandler,
+    getOwnPropertyDescriptors(scopeProxyHandlerProperties),
+  ),
+);
+
+export const strictScopeTerminator = new Proxy(
+  immutableObject,
+  strictScopeTerminatorHandler,
+);
```
mhofman, line 92: Wondering if it wouldn't make more sense to encapsulate / DRY up the scope proxy making logic a little more.

```suggestion
export const makeScopeTerminatorProxy = customHandlerProperties =>
  new Proxy(
    immutableObject,
    freeze(
      create(
        alwaysThrowHandler,
        getOwnPropertyDescriptors({
          ...scopeProxyHandlerProperties,
          ...customHandlerProperties,
        }),
      ),
    ),
  );

export const strictScopeTerminator = makeScopeTerminatorProxy();
```

packages/ses/test/test-sloppy-globals-scope-terminator.js
```diff
@@ -0,0 +1,87 @@
...
+  t.plan(4);
+
+  const globalObject = {};
+  const scopeTerminator = createSloppyGlobalsScopeTerminator(globalObject);
+
+  t.is(Reflect.has(scopeTerminator, 'eval'), true);
+  t.is(Reflect.get(scopeTerminator, 'eval'), undefined);
+
+  t.is(Reflect.has(scopeTerminator, 'xyz'), true);
+  t.is(Reflect.get(scopeTerminator, 'xyz'), undefined);
```
mhofman, line 14: Can we add a test for another prop `abc` (pick a name) that exists on the `globalObject`, so it's showing that `scopeTerminator` doesn't leak it?

I guess that behavior is tested in the `set` test below, but it'd be good to have an explicit `get` only test.

packages/ses/test/test-sloppy-globals-scope-terminator.js
```diff
@@ -0,0 +1,87 @@
...
+  t.is(Reflect.has(scopeTerminator, 'eval'), true);
+  t.is(didWarn, 0);
+  t.is(Reflect.getOwnPropertyDescriptor(scopeTerminator, 'eval'), undefined);
+  t.is(didWarn, 1);
+  t.is(Reflect.has(scopeTerminator, 'xyz'), true);
+  t.is(didWarn, 1);
+  t.is(Reflect.getOwnPropertyDescriptor(scopeTerminator, 'xyz'), undefined);
+  t.is(didWarn, 2);
+
+  console.warn = originalWarn;
```
mhofman, line 62: Please use `t.shutdown` to restore to original.

packages/ses/test/test-sloppy-globals-scope-terminator.js
```diff
@@ -0,0 +1,87 @@
...
+test('sloppyGlobalsScopeTerminator/getPrototypeOf - has null prototype', t => {
+  t.plan(1);
+
+  const globalObject = {};
+  const scopeTerminator = createSloppyGlobalsScopeTerminator(globalObject);
+
+  t.is(Reflect.getPrototypeOf(scopeTerminator), null);
+});
+
+test('sloppyGlobalsScopeTerminator/getOwnPropertyDescriptor - always has start compartment properties but provides no prop desc', t => {
```
mhofman, line 41: The `getOwnPropertyDescriptor`'s trap use of `console.warn` seem to be there to debug an obscure issue. I'm not sure it's worth testing that `console.warn` actually gets called. @erights you added this console logging I believe, do we care about testing it?

mhofman (approved):
This is looking good!

Some comments with a few potential further improvements but nothing blocking.

packages/ses/test/test-sloppy-globals-scope-terminator.js
mhofman: After discussing this in the meeting today, let's drop the `console.warn` part of this test.

packages/ses/src/make-safe-evaluator.js
```diff
@@ -104,7 +106,6 @@ export const makeSafeEvaluator = ({
         // variable resolution via the scopeHandler, and throw an error with
         // diagnostic info including the thrown error if any from evaluating the
         // source code.
-        revokeScopeProxy();
```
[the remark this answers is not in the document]
mhofman: Not needed for this PR, but we should look into a `panic()` call here instead [once we have one](https://github.com/endojs/endo/issues/490), as documented in comments below.

packages/ses/src/make-safe-evaluator.js
```diff
@@ -26,37 +22,42 @@ const { details: d } = assert;
...
+        freeze({
+          evalScope,
+          globalLexicals,
+          globalObject,
+          scopeTerminator,
+        }),
+        [],
+      );
     }
   };
```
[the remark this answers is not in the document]
mhofman: I will be pushing a commit with this refactor.

packages/ses/test/test-confinement.js
mhofman: I have erred in my judgement, and have now come around to @erights's suggestion that we install a property with `{ configurable: false, enumerable: false, set() { assert.fail('Some meaningful message'); } }`

packages/ses/src/strict-scope-terminator.js
mhofman: Here is what it would look like: https://github.com/endojs/endo/commit/dd156de59c4f862d99ed73ba298a4ca902ec4bdd

packages/ses/src/make-evaluate.js
```diff
@@ -0,0 +1,108 @@
...
+  // inside the `evaluate` closure. While we could potentially nest an
+  // intermediate layer of `() => {'use strict'; ${optimizers}; ...`, it
+  // doesn't seem worth the overhead and complexity.
+  const evaluateFactory = FERAL_FUNCTION(`
+    with (this.scopeTerminator) {
+      with (this.globalObject) {
+        with (this.globalLexicals) {
+          with (this.evalScope) {
+            ${globalObjectOptimizer}
+            ${globalLexicalOptimizer}
```
naugtur, line 96: The one meaningful change I see here compared to previous behavior is lexical optimizer is overshadowing/redefining variables of the same name from globals. There used to be a bit of logic that removed the ones we'd override anyway and the optimizer was therefore smaller.  
I don't know if it's an issue.

naugtur: Oh, you're still using getScopeConstants. Nothing changed then.

packages/ses/src/make-evaluate.js
```diff
@@ -48,35 +47,46 @@ export const makeEvaluate = context => {
   // eval, and thus in strict mode in the same scope. We must be very careful
   // to not create new names in this scope
 
-  // 1: we use 'with' (around a Proxy) to catch all free variable names. The
-  // `this` value holds the Proxy which safely wraps the safeGlobal
-  // 2: 'optimizer' catches constant variable names for speed.
-  // 3: The inner strict function is effectively passed two parameters:
+  // 1: we use multiple nested 'with' to catch all free variable names. The
```
kriskowal, line 50: I suggest “`with` blocks”.

kriskowal:
I’ve reviewed @mhofman’s 3 amendments and am convinced they’re a desirable refactor with no behavior change.

packages/ses/test/test-scope.js
```diff
@@ -169,39 +169,34 @@ test('scope behavior - this-value', t => {
...
+  t.is(evaluate('globalObjectProtoFn()'), globalObject);
+  t.is(evaluate('globalObjectFn()'), globalObject);
+  t.is(evaluate('globalObjectProtoFnImmutable()'), globalObject);
   t.is(evaluate('globalObjectFnImmutable()'), undefined);
-  t.true(knownScopeProxies.has(evaluate('globalLexicalsProtoFn()')));
-  t.true(knownScopeProxies.has(evaluate('globalLexicalsFn()')));
-  t.true(knownScopeProxies.has(evaluate('globalLexicalsProtoFnImmutable()')));
+  t.is(evaluate('globalLexicalsProtoFn()'), globalLexicals);
+  t.is(evaluate('globalLexicalsFn()'), globalLexicals);
+  t.is(evaluate('globalLexicalsProtoFnImmutable()'), globalLexicals);
```
kriskowal, line 199: This is the thesis, right here.

packages/ses/src/eval-scope.js
```diff
@@ -1,17 +1,18 @@
-import { FERAL_EVAL, create, freeze, defineProperties } from './commons.js';
+import { FERAL_EVAL, assign, create, freeze, defineProperties } from './commons.js';
 
 export const createEvalScope = () => {
-  const evalScope = create(null);
-  const oneTimeEvalProperties = freeze({
-    eval: {
+  const oneTimeEvalProperties = freeze(assign(create(null), {
```
mhofman: Why not `create(null, {` ?

packages/ses/test/test-confinement.js
kriskowal: Appended commit https://github.com/endojs/endo/pull/1293/commits/fc8f84b25512190b7bec36de8840011a47f98fa7

packages/ses/src/eval-scope.js
kriskowal: Equivalent but a bit more cumbersome, since there are more descriptor objects. It’s arguably equally reasonable to ask, why not just `{ __proto__: null, ... }`.

kriskowal:
> > adds a failing test `test-scope-handler-pollution` because the test's prototype pollution broke the `makeSafeEvaluator` internals
> 
> This sounds like a severe bug and a possible vulnerability to code (like vetted shims) that runs between repair and lockdown. Is it?

I add a change that fixed the test by using a null prototype on a property descriptor. We don’t do this nearly enough to pervasively avoid `Object.prototype.value` pollution from poorly-vetted shims. Filed https://github.com/endojs/endo/issues/1329

kriskowal (approved):
I intend to perform the merge myself when @erights is satisfied, so that I can ensure we get a clean rebase and include a commit with the breaking change ceremony. Thank you, @kumavis, @naugtur, & @mhofman for your commits.

packages/ses/src/commons.js
```diff
@@ -73,6 +73,7 @@ export const {
   toStringTag: toStringTagSymbol,
   iterator: iteratorSymbol,
   matchAll: matchAllSymbol,
+  unscopables: unscopablesSymbol,
   keyFor: symbolKeyFor,
   for: symbolFor,
 } = Symbol;
```
erights, line 79: Huh. I don't remember noticing this section before.

No change suggested.

packages/ses/src/eval-scope.js
```diff
@@ -0,0 +1,36 @@
...
+  defineProperties,
+} from './commons.js';
+
+export const createEvalScope = () => {
+  const evalScope = create(null);
+
+  const oneTimeEvalProperties = freeze(
+    assign(create(null), {
+      eval: freeze(
+        assign(create(null), {
```
erights, line 15: What is the double `assign(create(null), ...)` thing about? Please explain in a comment in the code.

erights: Is the part of the defense against `Object.prototype.value = something`? Given our current stance, well, I can see why we might want to at least preserve the protection right here. But `assign(create(null), ...)` seems overly complicated. Why not `{__proto__: null, ...}`?

erights: Or, given our general stance, I'd also be ok with removing the protection here as well, unless there's an additional reason to worry that I'm missing.

packages/ses/src/global-object.js
```diff
@@ -1,8 +1,47 @@
...
+ */
+export const setGlobalObjectSymbolUnscopables = globalObject => {
+  defineProperty(
+    globalObject,
+    unscopablesSymbol,
+    freeze(
+      assign(create(null), {
+        set: freeze(() => {
+          throw new TypeError(
+            `Cannot set Symbol.unscopables of a Compartment's globalThis`,
```
erights, line 35: Thanks!

packages/ses/src/global-object.js
```diff
@@ -1,8 +1,47 @@
...
+  freeze,
+  objectHasOwnProperty,
+  unscopablesSymbol,
+} from './commons.js';
 import { makeEvalFunction } from './make-eval-function.js';
 import { makeFunctionConstructor } from './make-function-constructor.js';
 import { constantProperties, universalPropertyNames } from './whitelist.js';
 
+/**
+ * The host's ordinary global object is not provided by a with block,
```
erights, line 16:
```suggestion
 * The host's ordinary global object is not provided by a `with` block,
```
and likewise elsewhere. My preference is to do this for all embedded code text, but we should do it at least for code that also looks like a normal standalone english word.

packages/ses/src/global-object.js
```diff
@@ -1,8 +1,47 @@
...
+ * so assigning to Symbol.unscopables has no effect.
+ * Since this shim uses with blocks to create a confined lexical scope for
+ * guest programs, we cannot emulate the proper behavior.
+ * With this shim, assigning Symbol.unscopables causes the given lexical
+ * names to fall through to the terminal scope proxy.
+ * But, we can install this setter to prevent a program from proceding on
+ * false assumption.
```
erights, line 23:
```suggestion
 * this false assumption.
```
or
```suggestion
 * such a false assumption.
```

packages/ses/src/eval-scope.js
erights: Whatever you do here wrt the `assign(create(null), ...)`, please do it for all occurrences that serve the same purpose.

erights (changes requested):
Just comments so far. This review pass not done yet.

packages/ses/src/eval-scope.js
mhofman: Yeah I think `__proto__: null` may be more readable.

```js
  const oneTimeEvalProperties = freeze({
    __proto__: null,
    eval: freeze({
      __proto__: null,
      get() {
        delete evalScope.eval;
        return FERAL_EVAL;
      },
      enumerable: false,
      configurable: true,
    }),
  });
```

(and I guess same for the `setGlobalObjectSymbolUnscopables` property descriptor.

packages/ses/test/test-sloppy-globals-scope-terminator.js
```diff
@@ -0,0 +1,82 @@
...
+
+  const globalObject = {};
+  const scopeTerminator = createSloppyGlobalsScopeTerminator(globalObject);
+
+  // Mute warnings
+  const originalWarn = console.warn;
+  t.teardown(() => {
+    console.warn = originalWarn;
+  });
+  console.warn = () => {};
```
mhofman, line 52: @kriskowal I believe we decided to simply remove the testing of `console.warn` being called, which would remove all this.

packages/ses/src/eval-scope.js
kriskowal: Per #1329, we’ve elected to actually leave the original in place, no `__proto__` at all, since only a poorly vetted shim can invalidate this code. I’ve added a commit that removes the test that motivated this dance.

erights (commented):
I'm missing something very basic. Why does `delete evalScope.eval` appear in both eval-scope.js and make-safe-evalutor.js ? Surely this should be done in only one place, yes?

mhofman:
The one in safe-evaluator.js is the backup one in case something went extremely wrong and it didn't get deleted.

kriskowal:
> I'm missing something very basic. Why does `delete evalScope.eval` appear in both eval-scope.js and make-safe-evalutor.js ? Surely this should be done in only one place, yes?

The occurrence in `evalScope` is the one we’re sure will work. The one in `makeSafeEvaluator` _should be redundant_ and I can’t imagine a scenario where it would do anything except delete the property after it has already been deleted, but is here as a hedge in case there’s a way for the `evalScope.eval` getter manages to throw before it deletes itself.

kriskowal:
And, notably, we’re using `delete evalScope.eval` instead of a meaningfully named method of the “eval scope kit” in order to avoid consuming a stack frame, in order to frustrate stack bumping.

kriskowal:
@erights I’ve appended a commit, 79e77b9fb, that makes the evalScope revocable. I chose to use a property of the evalScopeKit to communicate the revocation bit (and a wrapper object to guard against `throw null`) for reasons I explain in a new comment. I’m asking for a review of just this change.

packages/ses/src/eval-scope.js
```diff
@@ -1,6 +1,29 @@
-import { FERAL_EVAL, create, freeze, defineProperties } from './commons.js';
+import { FERAL_EVAL, create, defineProperties, freeze } from './commons.js';
+
+import { assert } from './error/assert.js';
+
+const { details: d } = assert;
+
+export const makeEvalScopeKit = () => {
+  // In makeSafeEvaluator, we ensure that we consume at least as many stack
+  // frames when granting access to FERAL_EVAL than we consume when we are
```
erights, line 9:
```suggestion
  // In makeSafeEvaluator, we ensure that we consume at least as many stack
  // frames when granting access to FERAL_EVAL as we consume when we are
```

packages/ses/src/eval-scope.js
```diff
@@ -1,6 +1,29 @@
...
+  // immediately withdrawing access, such that the safe evaluator can reach it
+  // in the lexical scope of the confined program once (to evaluate the
+  // confined code) and confined code cannot access it thereafter.
+  // We do this to frustrate stack-bumping attacks, where the caller arranges
+  // for the stack to be a number of frames short of inducing a `RangeError` in
+  // order to interrupt a program that may not be able to clean up safely.
+  // In this case, there is a plausible but very narrow window of opportunity
+  // between defining `oneTimeEvalProperties` on the `evalScope` and when the
+  // confined program reaches for the eval property on the `evalScope` through
+  // a lexical reference where a program could induce a RangeError that would
```
erights, line 19:
```suggestion
  // In this case, there is a plausible but very narrow window of 
  // opportunity---between defining `oneTimeEvalProperties` on the `evalScope` and when the
  // confined program reaches for the eval property on the `evalScope` through
  // a lexical reference---where a program could induce a RangeError that would
```

packages/ses/src/eval-scope.js
```diff
@@ -1,6 +1,29 @@
...
+  // prevent the program from reaching `delete evalScope.eval` either here or
```
erights, line 20:
```suggestion
  // prevent the program from reaching `delete evalScope.eval`, either here or
```

Even with the extra punctuation, this is an amazing run-on sentence. Better to rephrase using more smaller sentences ;)

packages/ses/src/eval-scope.js
```diff
@@ -1,6 +1,29 @@
...
+  // in the fallback position in `makeSafeEvaluator` where we attempt to delete
```
erights, line 21:
```suggestion
  // in the fallback position in `makeSafeEvaluator`, where we attempt to delete
```
How long is that sentence? ;)

packages/ses/src/eval-scope.js
```diff
@@ -1,6 +1,29 @@
...
+  // it again in the finally clause.
+  // By using `delete` and simple property assignment for
+  // `evalScopeKit.revoked`, we avoid pushing to the stack while withdrawing
+  // access to the `FERAL_EVAL`.
```
erights, line 25: Worth noting the non-guaranteed implementation properties we assume in this reasoning that may well be wrong on particular platforms:
   * That if stack A has more frames than stack B, where none of these frames are unusual, then stack B will not exceed the stack if stack A did not. May be false depending on the sizes of each frame.
   * That function calls in the semantics of the language is an adequate predictor of stack frames. May be false when some of these function calls are inlined or otherwise optimized away --- as we know happens with JITs.
   * Have we reexamined our frame-count reasoning in light of the possibility of implementation tail-calling? The ECMAScript std mandates it and JSC actually implements it.

The reason we're ok with this fragile reasoning is that it is only about a backstop that we'd need if all the other relevant mechanisms failed. Nothing will protect against any possible bug we overlooked. Rather, we're only trying for defense in depth, so that multiple improbably things would need to go wrong to create an *exploitable* vulnerability.

packages/ses/src/make-safe-evaluator.js
```diff
@@ -95,6 +97,7 @@ export const makeSafeEvaluator = ({
         // variable resolution via the scopeHandler, and throw an error with
         // diagnostic info including the thrown error if any from evaluating the
         // source code.
+        evalScopeKit.revoked = { err };
```
erights, line 100: Very nice communicating the error through the revocation channel!

erights (approved):
The "Revocable evalScope" commit LGTM

PR #1296, opened 2022-09-26
feat(bundle-source): normalized `getExport` and `nestedEvaluate` pathnames

michaelfig:
Remove pathname sensitivity from the `getExport` and `nestedEvaluate` bundle formats by trimming the longest common prefix from emitted output names. 

Also, make the `nestedEvaluate` format backwards compatible with `getExport` when evaluating code in the start compartment.

### DIFF
--- a/packages/bundle-source/src/index.js
+++ b/packages/bundle-source/src/index.js
@@ -325,11 +325,8 @@ ${sourceMap}`;
     // This function's source code is inlined in the output bundle.
     // It figures out the exports from a given module filename.
     const nsBundle = {};
-    const nestedEvaluate = _src => {
-      throw Error('need to override nestedEvaluate');
-    };
     function computeExports(filename, exportPowers, exports) {
-      const { require: systemRequire, _log } = exportPowers;
+      const { require: systemRequire, systemEval, _log } = exportPowers;
       // This captures the endowed require.
       const match = filename.match(/^(.*)\/[^/]+$/);
       const thisdir = match ? match[1] : '.';
@@ -399,7 +396,8 @@ ${sourceMap}`;
       }
 
       // log('evaluating', code);
-      return nestedEvaluate(code)(contextRequire, exports);
+      // eslint-disable-next-line no-eval
+      return (systemEval || eval)(code)(contextRequire, exports);
     }
 
     source = `\
@@ -420,7 +418,8 @@ function getExportWithNestedEvaluate(filePrefix) {
 
   // Evaluate the entrypoint recursively, seeding the exports.
   const systemRequire = typeof require === 'undefined' ? undefined : require;
-  return computeExports(entrypoint, { require: systemRequire }, {});
+  const systemEval = typeof nestedEvaluate === 'undefined' ? undefined : nestedEvaluate;
+  return computeExports(entrypoint, { require: systemRequire, systemEval }, {});
 }
 ${sourceMap}`;
   }
--- a/packages/bundle-source/src/index.js
+++ b/packages/bundle-source/src/index.js
@@ -32,6 +32,23 @@ const textEncoder = new TextEncoder();
 const textDecoder = new TextDecoder();
 const readPowers = makeReadPowers({ fs, url, crypto });
 
+// Find the longest common prefix of an array of strings.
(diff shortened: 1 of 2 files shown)


packages/bundle-source/src/index.js
```diff
@@ -192,6 +209,17 @@ async function bundleNestedEvaluateAndGetExports(
   });
   // console.log(output);
 
+  // Find the longest common prefix of all the source file names.
+  // We shorten the fileNames to make the bundle output deterministic.
```
turadg, line 213: wasn't it "deterministic" before, just dependent on the host machine path?

Consider,
```suggestion
  // We shorten the fileNames to make the bundle output consistent between build environments.
```

packages/bundle-source/src/index.js
```diff
@@ -192,6 +209,17 @@ async function bundleNestedEvaluateAndGetExports(
...
+  const fileNameToUrlPath = fileName =>
+    readPowers.pathToFileURL(pathResolve(startFilename, fileName)).pathname;
+  const pathnames = output.map(({ fileName }) => fileNameToUrlPath(fileName));
+  const longestPrefix = longestCommonPrefix(pathnames);
+
+  // Ensure the prefix ends with a slash.
+  const pathnameEndPos = longestPrefix.lastIndexOf('/');
+  const pathnamePrefix = longestPrefix.slice(0, pathnameEndPos + 1);
```
kriskowal, line 221: I reckon that this is correct even for pairs of paths like `common/common-left` and `common/common-right`. My intuition would have been to implement `longestCommonPrefix` as `longestCommonPath` such that it guarantees that the prefix occurs at a path delimiter, but this works too.

Also, doing the math in the URL domain would be more portable, since it normalizes delimiters.

packages/bundle-source/src/index.js
```diff
@@ -250,7 +282,7 @@ async function bundleNestedEvaluateAndGetExports(
   let sourceMap;
   let source;
   if (moduleFormat === 'getExport') {
-    sourceMap = `//# sourceURL=${resolvedPath}\n`;
+    sourceMap = `//# sourceURL=${DEFAULT_FILE_PREFIX}/${entrypoint}\n`;
```
kriskowal, line 285: I think the default file prefix should be `...`, but I see it’s `/bundled-source/...`. I presume you verified this works for debuggers?

kriskowal (approved):

packages/bundle-source/src/index.js
michaelfig: > Also, doing the math in the URL domain would be more portable, since it normalizes delimiters.

Not sure what you mean by this... the math is currently done on the normalised `url.pathname`.  Are you suggesting that something different would be better?

packages/bundle-source/src/index.js
michaelfig: Yes, for VSCode at least.

packages/bundle-source/src/index.js
kriskowal: No, I misread. You’re already doing what I suggested. Thanks!

packages/bundle-source/src/index.js
warner: I like that phrasing

warner:
To make sure I understand, if we `bundleSource('.../src/entrypoint.js')` and that file only transitively imports siblings or deeper, the resulting bundle's pathnames would omit the `src/` prefix, right? Because this just finds all the files in the bundle and then trims off any common prefix in their fully-qualified pathnames? And then given the way npm and `node_modules` and rollup and monorepos all work, everything in our e.g. `supervisor` bundle will either come from the monorepo's top-level `node_modules/`, or from something like `packages/$PACKAGE/src/foo.js`, so we'll expose the structure of the monorepo but not the directory in which it lives?

If so, good, I was sort of expecting `bundleSource()` to require an extra argument to say "this is the top of the tree", but that would be a nuisance to use.

michaelfig:
> If so, good, I was sort of expecting bundleSource() to require an extra argument to say "this is the top of the tree", but that would be a nuisance to use.

Yeah, I considered both possibilities and went with the first one.

PR #1297, opened 2022-09-26
fix: safe-promise had a false negative

erights:
A debugger may be activated after lockdown, in which case `Promise.prototype` may be unaffected, and therefore not a good source for the symbols we wish to tolerate.

I was getting a throw because of this when running code under the vscode debugger. After this fix, I was able to resume my debugging.


mhofman (commented):
Opened an issue for solving the root cause of this observed behavior: https://github.com/endojs/endo/issues/1298

erights:
Closing this as misdiagnosed, and therefore solving the wrong problem. See https://github.com/endojs/endo/issues/1298

PR #1302, opened 2022-09-27
Add general scope tests

kumavis:
tests extracted from https://github.com/endojs/endo/pull/1293 and fixed for current implementation

### DIFF
--- a/packages/ses/test/test-scope.js
+++ b/packages/ses/test/test-scope.js
@@ -158,6 +158,7 @@ test('scope behavior - this-value', t => {
   t.is(globalObjectProtoSetterValue, globalObject);
   evaluate('globalObjectSetter = 123');
   t.is(globalObjectSetterValue, globalObject);
+  // bug: properties in prototype of globalLexicals error on set
   t.throws(() => evaluate('globalLexicalsProtoSetter = 123'), {
     instanceOf: Error,
   });
--- a/packages/ses/test/test-scope.js
+++ b/packages/ses/test/test-scope.js
@@ -1,0 +1,199 @@
+/* global globalThis */
+
+import test from 'ava';
+import { makeSafeEvaluator } from '../src/make-safe-evaluator.js';
+
+test('scope behavior - lookup behavior', t => {
+  t.plan(7);
+
+  const globalObject = { foo: {} };
+  const globalLexicals = { foobar: {} };
+
+  const { safeEvaluate: evaluate } = makeSafeEvaluator({
+    globalObject,
+    globalLexicals,
+  });
+
+  globalThis.bar = {};
+
+  t.is(evaluate('globalThis'), undefined);
+  t.is(evaluate('eval'), undefined);
+  t.is(evaluate('bar'), undefined);
+  t.throws(() => evaluate('dummy'), { instanceOf: ReferenceError });
+
+  t.is(evaluate('foo'), globalObject.foo);
+  t.is(evaluate('foobar'), globalLexicals.foobar);
+  t.deepEqual(evaluate('arguments'), ['arguments']);
+
+  delete globalThis.bar;
+});
+
+test('scope behavior - lookup in sloppyGlobalsMode', t => {
+  t.plan(7);
+
+  const globalObject = {};
+  const globalLexicals = {};
+  const { safeEvaluate: evaluate } = makeSafeEvaluator({
+    globalObject,
+    globalLexicals,
+    sloppyGlobalsMode: true,
+  });
+
+  globalThis.bar = {};
+
+  t.is(evaluate('globalThis'), undefined);
+  t.is(evaluate('eval'), undefined);
+  t.is(evaluate('foo'), undefined);
+  t.is(evaluate('bar'), undefined);
+  t.is(evaluate('foobar'), undefined);
+  t.is(evaluate('dummy'), undefined);
+
+  t.deepEqual(evaluate('arguments'), ['arguments']);
+
+  delete globalThis.bar;
+});
+
+test('scope behavior - this-value', t => {
+  t.plan(6);
+
+  let hogeValue;
+  let fugaValue;
+
+  const globalObject = {
+    get foo() {
+      return this;
+    },
+    set hoge(_value) {
+      hogeValue = this;
+    },
+    quux() {
+      return this;
+    },
+  };
+  const globalLexicals = {
+    get bar() {
+      return this;
+    },
+    set fuga(_value) {
+      fugaValue = this;
+    },
+    garply() {
+      return this;
+    },
+  };
+  const { safeEvaluate: evaluate } = makeSafeEvaluator({
+    globalObject,
+    globalLexicals,
+  });
+
+  t.is(evaluate('foo'), globalObject);
+  t.is(evaluate('bar'), globalLexicals);
+
+  evaluate('hoge = 123');
+  evaluate('fuga = 456');
+  t.is(hogeValue, globalObject);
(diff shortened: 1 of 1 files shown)


packages/ses/test/test-scope.js
```diff
@@ -0,0 +1,199 @@
...
+  globalThis.bar = {};
+
+  t.is(evaluate('globalThis'), undefined);
+  t.is(evaluate('eval'), undefined);
+  t.is(evaluate('bar'), undefined);
+  t.throws(() => evaluate('dummy'), { instanceOf: ReferenceError });
+
+  t.is(evaluate('foo'), globalObject.foo);
+  t.is(evaluate('foobar'), globalLexicals.foobar);
+  t.deepEqual(evaluate('arguments'), ['arguments']);
```
kriskowal, line 26: Let’s call this case out as a deliberate aberration in a comment.

```
// Known compromise in fidelity of the emulated script environment:
```

packages/ses/test/test-scope.js
```diff
@@ -0,0 +1,199 @@
...
+  const originalFoobar = globalLexicals.foobar;
+  doAssignment('foobar', foobar);
+  t.is(globalLexicals.foobar, foobar);
+  t.not(globalObject.foo, originalFoobar);
+  t.is(globalObject.foobar, undefined);
+  t.is(globalThis.foobar, undefined);
+
+  t.is(Object.keys(globalObject).length, 3);
+  t.is(Object.keys(globalLexicals).length, 1);
+
```
kriskowal, line 149: `doAssignment('eval', evil)` would be an interesting case, and of course verifying that this does not break the machinery of the safe evaluator afterward.

packages/ses/test/test-scope.js
```diff
@@ -0,0 +1,199 @@
...
+  t.is(evaluate('typeof bar'), 'undefined');
+  t.throws(() => evaluate('foo'), { instanceOf: ReferenceError });
+  t.throws(() => evaluate('bar'), { instanceOf: ReferenceError });
+
+  globalThis.bar = {};
+
+  t.is(evaluate('typeof foo'), 'undefined');
+  t.is(evaluate('typeof bar'), 'undefined');
+  t.throws(() => evaluate('foo'), { instanceOf: ReferenceError });
+  t.is(evaluate('bar'), undefined);
```
kriskowal, line 196: Perhaps also mark this as an aberration on the fidelity of the emulation.

kriskowal (approved):

packages/ses/test/test-scope.js
kumavis: assigning to `eval` without first doing an assignment to `this.eval` will throw a `ReferenceError` when `sloppyGlobalsMode` is not set

kumavis: as for testing if that assignment breaks the evaluator, we are already indirectly testing that

kriskowal: Works for me as long as it’s not the last assertion.

packages/ses/test/test-scope.js
```diff
@@ -0,0 +1,341 @@
...
+  t.throws(() => evaluate('dummy'), { instanceOf: ReferenceError });
+
+  t.is(evaluate('foo'), globalObject.foo);
+  t.is(evaluate('foobar'), globalLexicals.foobar);
+
+  // Known compromise in fidelity of the emulated script environment:
+  t.deepEqual(evaluate('arguments'), ['arguments']);
+
+  delete globalThis.bar;
```
mhofman, line 30: Please setup as a [`teardown`](https://github.com/avajs/ava/blob/main/docs/02-execution-context.md#tteardownfn) cleanup in case the test fails to avoid polluting other tests.

packages/ses/test/test-scope.js
```diff
@@ -0,0 +1,341 @@
...
+  t.is(globalObjectProtoSetterValue, globalObject);
+  evaluate('globalObjectSetter = 123');
+  t.is(globalObjectSetterValue, globalObject);
+  // bug: properties in prototype of globalLexicals error on set
+  t.throws(() => evaluate('globalLexicalsProtoSetter = 123'), {
+    instanceOf: Error,
+  });
+  t.is(globalLexicalsProtoSetterValue, undefined);
+  evaluate('globalLexicalsSetter = 123');
+  t.is(globalLexicalsSetterValue, globalObject);
```
mhofman, line 196: I'm confused, I would have expected the scope proxy here, not the `globalObject`.

Is that because we use an explicit `Reflect.get` / `Reflect.set` with the `globalObject`?

packages/ses/test/test-scope.js
```diff
@@ -0,0 +1,341 @@
...
+
+  t.true(knownScopeProxies.has(evaluate('globalObjectProtoFn()')));
+  t.true(knownScopeProxies.has(evaluate('globalObjectFn()')));
+  t.true(knownScopeProxies.has(evaluate('globalObjectProtoFnOptimizable()')));
```
mhofman, line 200: A bit of a misnomer, on "own" properties are optimizable, never proto props. How about `globalObjectProtoImmutableFn`

packages/ses/test/test-scope.js
```diff
@@ -0,0 +1,341 @@
...
+  // Known compromise in fidelity of the emulated script environment:
+  t.is(evaluate('bar'), undefined);
+
+  evaluate('this[Symbol.unscopables] = { bar: true }');
+
+  t.is(evaluate('typeof foo'), 'undefined');
+  t.is(evaluate('typeof bar'), 'undefined');
+  t.throws(() => evaluate('foo'), { instanceOf: ReferenceError });
+  // Known compromise in fidelity of the emulated script environment:
+  t.is(evaluate('bar'), undefined);
```
mhofman, line 338: I think the overload of a `bar` test is confusing here. I'd expect the aberration to be:

```js
evaluate('this.baz = {}; this[Symbol.unscopables] = { baz: true }; ');
// Known compromise in fidelity of the emulated script environment:
t.throws(() => evaluate('baz'), { instanceOf: ReferenceError });
```

mhofman (commented):
Thanks for adding all these scope tests!

packages/ses/test/test-scope.js
kumavis: yes, the aberration is very different in the current version vs the triple backflip refactor. this test was mostly added to show off the aberration the backflip introduces.

packages/ses/test/test-scope.js
kumavis: I was also surprised by the results from this test!

you are correct about the reason for `globalLexicalsSetterValue` https://github.com/endojs/endo/blob/c603c5aa4a1ba271cf17d754df789a52aa7debfb/packages/ses/src/scope-handler.js#L122-L124

packages/ses/test/test-scope.js
```diff
@@ -289,53 +291,84 @@ test('scope behavior - realm globalThis property info leak', t => {
...
   // Known compromise in fidelity of the emulated script environment:
-  t.is(evaluate('bar'), undefined);
+  t.is(evaluate('eventuallyAssignedRealmGlobalProp'), undefined);
+
+  evaluate(
+    'this[Symbol.unscopables] = { eventuallyAssignedRealmGlobalProp: true, localProp: true }',
+  );
 
-  delete globalThis.bar;
+  t.is(evaluate('typeof localProp'), 'object');
```
mhofman, line 364: Oh yeah I realize now the current scopeProxy shields us from this footgun.

mhofman (approved):

Issue #1303, opened 2022-09-28
Unserialization creates copyRecord with assignment rather than define semantics


erights:
This fails in two ways:
   * If a copyRecord contains a property that would override a non-enabled frozen data property inherited from `Object.prototype`.
   * If a copyRecord contains a property named `__proto__` is a system in which the `Object.prototype.__proto__` accessor property still exists. See https://github.com/Agoric/agoric-sdk/pull/6349

erights:
Being fixed by #1304 . The new test case there fails before the fix in #1304. However, it fails too late!
```js
test('proto problems', t => {
  ...
  const exampleAlice = Far('Alice', {});
  ...
  const wrongProto = harden({ ['__proto__']: exampleAlice });
  const wrongProtoEnc = ser(wrongProto);
  ...
  t.deepEqual(unser(wrongProtoEnc), wrongProto);
});
```

The `unser` (unserializer) call itself succeeds, returning a passable that inherits from exampleAlice. In fact, in that pre-fix-state, the following bogus test passes instead:
```js
  t.deepEqual(unser(wrongProtoEnc), exampleAlice);
```
apparently because Ava's `deepEqual` considers an empty object that inherits from Alice to look deeply like Alice. But in any case, this establishes that before #1304 a malicious message can use this bug to serialize an object that inherits from a remotable that they have. However, this does not violate ocap rules, and can happen locally anyway between mutually suspicious subgraphs within the same vat.

PR #1306, opened 2022-09-30
fix: delete broken objectFromEntries

erights:
commons.js contains its own ponyFill (aka shim, polyfill) of `Object.fromEntries` because it was relevant at the time. It no longer is. Fortunately it was dead code anyway because it was only exported (as `fromEntries`) when on a platform on which it is absent. Thus deleting it causes no transition cost. We can avoid the deprecation dance. By this criteria of course, fixing this is also low priority.

The reason to bother getting rid of it is that it uses assignment semantics to initialize the object it creates and returns. The resulting problems explained at https://github.com/endojs/endo/issues/1303 this has some terrible problems do apply to this code as well.

### DIFF
--- a/packages/ses/src/commons.js
+++ b/packages/ses/src/commons.js
@@ -66,6 +66,7 @@ export const {
   preventExtensions,
   setPrototypeOf,
   values,
+  fromEntries,
 } = Object;
 
 export const {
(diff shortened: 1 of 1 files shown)


erights:
Hmmm. https://github.com/endojs/endo/actions/runs/3162163428/jobs/5148508233 shows a platform compat failure, presumably because it is testing a platform only enough to still not have its own `fromEntries`. If we do decide to support such platforms, then the code is not dead and its assignment problems (like #1303) should be fixed.

kriskowal:
@erights Consider review and rebase on https://github.com/endojs/endo/pull/1308

I’m in favor of allowing Node.js 10 vanish into history.

erights:
> @erights Consider review and rebase on #1308
> 
> I’m in favor of allowing Node.js 10 vanish into history.

Rebased, thanks!

kriskowal (approved):
Thank you. I’m surprised that we no longer use `fromEntries` anywhere, but I trust the tests.

erights:
Thank you. I’m surprised that we no longer use `fromEntries` anywhere, but I trust the tests.

Look again. I'm still exporting `fromEntries`, but now doing so like a normal static method.

erights:
Huh. But maybe we don't actually import it anyway. I'm just as surprised.

erights:
It is imported by `console.js` so I'm going to leave the new export as is.

Issue #1254, opened 2022-08-16
export explicit types


turadg:
Most @endo packages export types ambiently. This leads to problems downstream like https://github.com/Agoric/agoric-sdk/issues/4560

Instead @endo packages should export types explicitly as in https://github.com/endojs/endo/pull/1025/

Here is a prioritized partial list:
1. [x] marshall
1. [ ] captp
1. [ ] promise-kit
1. _add more here as they are prioritized or completed_

Issue #1235, opened 2022-07-18
Adopt simplified ses-ava


turadg:
https://github.com/Agoric/agoric-sdk/pull/5774/

**Motivation**

Hacks around side-effects of prepare-test-env-ava such as https://github.com/endojs/endo/pull/1439#discussion_r1070430689

**Acceptance criteria**
- ~~module imports either have side-effects OR bindings~~ moved to #1467
- Ava macros work

PR #1307, opened 2022-09-30
more explicit type exports

turadg:
Progress on https://github.com/endojs/endo/issues/1254


packages/captp/jsconfig.json
```diff
@@ -3,17 +3,17 @@
...
-/*
-    // The following flags are for creating .d.ts files:
-    "noEmit": false,
-    "declaration": true,
-    "emitDeclarationOnly": true,
-*/
     "downlevelIteration": true,
     "strictNullChecks": true,
     "moduleResolution": "node",
+    "types": []
```
mhofman, line 11: Curious what motivated this. It seems about right, but I would not have thought of changing this.

packages/promise-kit/package.json
```diff
@@ -40,6 +43,7 @@
   "devDependencies": {
     "@endo/eslint-config": "^0.5.1",
     "@endo/ses-ava": "^0.2.33",
+    "@types/node": ">=11.0",
```
mhofman, line 46: I'd be ok with bumping this up to `14.0`

packages/promise-kit/test/test-promise-kit.js
```diff
@@ -22,6 +22,8 @@ if (typeof globalThis.gc !== 'function') {
   engineGC = globalThis.gc;
 }
 
+/** @type {typeof rawTest} */
+// @ts-expect-error cast
```
mhofman, line 26: I wish we could type `wrapTest` correctly to avoid this.

packages/promise-kit/test/test-promise-kit.js
```diff
@@ -58,7 +60,7 @@ function testRacePromise(t, candidate) {
     const result = Promise.resolve(candidate);
 
     bothResults = Promise.allSettled([raced, result]);
-    fr.register(raced, collected.resolve);
+    fr.register(raced, () => collected.resolve(undefined));
```
mhofman, line 63: Nothing to do with this really, but it's one of my biggest nit with tsc that `() => void` is incompatible with `() => any`. If it wasn't a test, I'd have asked for correcting the type of `collected` to `PromiseKit<void>` instead as I dislike changing runtime for types.

mhofman (approved):
Thanks for this!

packages/captp/package.json
```diff
@@ -30,6 +30,9 @@
   },
   "scripts": {
     "build": "exit 0",
+    "clean": "tsc --build jsconfig.build.json --clean",
```
kriskowal, line 33: This is fine, but it may be good to manually verify that this works. The last Endo publish didn’t clean up after itself and I’m suspicious. I did not look deeply into why. It may be a problem higher up the foodchain, like lerna not running postpack.

kriskowal (approved):
Typing _and_ typing!

packages/promise-kit/test/test-promise-kit.js
turadg: yeah. part of https://github.com/endojs/endo/issues/1235

packages/promise-kit/package.json
turadg: done. and since that still resolves to 14 that was available in yarn.lock, I also did a yarn upgrade.

packages/captp/jsconfig.json
turadg: Without it there are [errors from `node_modules/events`](https://github.com/endojs/endo/actions/runs/3162294420/jobs/5148757446),
```
../../node_modules/events/events.js:53:3 - error TS2323: Cannot redeclare exported variable 'defaultMaxListeners'.

 53   Object.defineProperty(EventEmitter, 'defaultMaxListeners', {
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 54     enumerable: true,
    ~~~~~~~~~~~~~~~~~~~~~
...
 64     }
    ~~~~~
 65   });
    ~~~~

../../node_modules/events/events.js:67:3 - error TS2323: Cannot redeclare exported variable 'defaultMaxListeners'.

67   EventEmitter.defaultMaxListeners = defaultMaxListeners;
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

../../node_modules/events/events.js:165:11 - error TS2339: Property 'context' does not exist on type 'Error'.

165       err.context = er;
              ~~~~~~~
```

packages/captp/package.json
turadg: when I ran `yarn clean` it didn't remove the generated files. blocker? seems out of scope to me.

kriskowal: Let’s remove the ineffective script and proceed. The work around for now is an additional step in the release process that involves nuking files that are not checked in. That carries some risk that I’ll bear going forward, until we find a better way.

turadg: I haven't removed it yet because, from what I've read, it sometimes works. It's just very cautious to never delete if some files have been touched.

kriskowal: This is not a blocker.

PR #1308, opened 2022-10-01
chore(ses): Advance Node.js compatibility test from version 10 to 12

kriskowal:
Per #1306, we have an invalid `Object.fromEntries` shim that is only needed before Node.js 12 and analogous browsers. This change raises the _tested_ lockdown compatibility floor up to Node.js 12.

- fix(ses): Fix incompatible spelling
- chore(ses): Advance lowest supported Node.js compatibility test from 10 to 12

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -349,10 +349,10 @@ jobs:
     - name: 'build'
       run: yarn run build
 
-    - name: 'switch to node v10'
+    - name: 'switch to node v12'
       uses: actions/setup-node@v1
       with:
-        node-version: '10.x'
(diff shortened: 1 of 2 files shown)


erights (approved):
LGTM, thanks!

erights:
ping. You've already enables auto-merge, but it seems stuck in CI.

pinging because https://github.com/endojs/endo/pull/1306 waits for this one.

turadg:
The problem was the branch protection rules were looking for the mispelling. 
Screen Shot 2022-10-03 at 9 21 56 AM
I've updated to match the new correct spelling.

PR #1310, opened 2022-10-07
perf(marshal): Extend new assert style to passStyleOf `check` calls

gibson042:
Ref dc24f2f2c3cac7ce239a64c503493c41a2334315

```sh
  import "../init/index.js";
  import { passStyleOf } from "./src/passStyleOf.js";
  import { makeTagged } from "./src/makeTagged.js";
  const inputs = [
    undefined,
    "foo",
    true,
    33,
    33n,
    Symbol.for("foo"),
    Symbol.iterator,
    null,
    Promise.resolve(null),
    [3, 4],
    { foo: 3 },
    { then: "non-function then ok" },
    makeTagged("unknown", undefined),
    Error("ok"),
  ].map(v => harden(v));
  const badInputs = [
    Symbol("unique"),
    {},
    harden(Object.setPrototypeOf(Promise.resolve(), { __proto__: Promise.prototype })),
    harden(Object.assign(Promise.resolve(), { extra: "unexpected own property" })),
    harden(Object.defineProperties(Promise.resolve(), { then: { value: () => "bad then" } })),
    harden({ then: () => "thenable" }),
  ];
  let result;
  (new Benchmark.Suite())
    .add("valid inputs", () => {
      for (const v of inputs) result = passStyleOf(v);
    })
    .add("invalid inputs", () => {
      let caught;
      for (const v of badInputs) {
        caught = undefined;
        try {
          result = passStyleOf(v);
        } catch (err) {
          caught = err;
        } finally {
          if (!caught) throw new Error(`unexpected acceptance: ${v}`);
        }
      }
    })
    .add("big input", () => {
      result = passStyleOf(harden(Array(1000).fill().map(() => [...inputs])));
    })
    .on("complete", () => result)
    .run({ async: true })
'
```

</details>

before:
```
valid inputs x 831,564 ops/sec ±0.71% (91 runs sampled)
invalid inputs x 3,645 ops/sec ±0.86% (88 runs sampled)
big input x 10.31 ops/sec ±3.22% (32 runs sampled)
```
after:
```
valid inputs x 828,877 ops/sec ±0.60% (91 runs sampled)
invalid inputs x 3,780 ops/sec ±0.76% (85 runs sampled)
big input x 66.53 ops/sec ±1.74% (67 runs sampled)
```

(and we could probably get even more improvement by replacing `check` parameters with `reject` to avoid creating the latter functions inline)

### DIFF
--- a/packages/marshal/src/helpers/copyRecord.js
+++ b/packages/marshal/src/helpers/copyRecord.js
@@ -24,9 +24,10 @@ export const CopyRecordHelper = harden({
   styleName: 'copyRecord',
 
   canBeValid: (candidate, check) => {
+    const reject = details => check(false, details);
     const proto = getPrototypeOf(candidate);
     if (proto !== objectPrototype && proto !== null) {
-      return check(false, X`Unexpected prototype for: ${candidate}`);
+      return reject(X`Unexpected prototype for: ${candidate}`);
     }
     const descs = getOwnPropertyDescriptors(candidate);
     const descKeys = ownKeys(descs);
@@ -34,15 +35,13 @@ export const CopyRecordHelper = harden({
     for (const descKey of descKeys) {
       if (typeof descKey !== 'string') {
         // Pass by copy
-        return check(
-          false,
+        return reject(
           X`Records can only have string-named own properties: ${candidate}`,
         );
       }
       const desc = descs[descKey];
       if (canBeMethod(desc.value)) {
-        return check(
-          false,
+        return reject(
           X`Records cannot contain non-far functions because they may be methods of an implicit Remotable: ${candidate}`,
         );
       }
--- a/packages/marshal/src/helpers/copyArray.js
+++ b/packages/marshal/src/helpers/copyArray.js
@@ -17,7 +17,7 @@ export const CopyArrayHelper = harden({
   styleName: 'copyArray',
 
   canBeValid: (candidate, check) =>
-    check(isArray(candidate), X`Array expected: ${candidate}`),
+    isArray(candidate) || check(false, X`Array expected: ${candidate}`),
 
   assertValid: (candidate, passStyleOfRecur) => {
     CopyArrayHelper.canBeValid(candidate, assertChecker);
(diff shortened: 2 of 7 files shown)


erights (commented):
Should I review this assuming that you intend it to be a pure refactoring --- with no observable differences? FWIW, I have not spotted any yet.

gibson042:
> Should I review this assuming that you intend it to be a pure refactoring --- with no observable differences? FWIW, I have not spotted any yet.

Correct. :+1:

erights (approved):
All nice simplifications, thanks!

LGTM

packages/marshal/src/helpers/copyRecord.js
```diff
@@ -24,25 +24,24 @@ export const CopyRecordHelper = harden({
...
     const descKeys = ownKeys(descs);
 
     for (const descKey of descKeys) {
       if (typeof descKey !== 'string') {
         // Pass by copy
-        return check(
-          false,
+        return reject(
           X`Records can only have string-named own properties: ${candidate}`,
         );
```
kriskowal, line 40: @erights I’m reasonably certain that the X/describe tagged literal does not have the same effect on `reject` or even as an argument to the `Error` constructor as it does on `assert` methods. You may need `assert.error` here for the details to be extractable by the console.

kriskowal (commented):
Looks mechanical to me, but please wait for confirmation about using details/quote (X/q) with `reject`. I’m pretty sure it doesn’t do what it says on the tin.

packages/marshal/src/helpers/copyRecord.js
gibson042: `assertValid` calls `canBeValid(candidate, assertChecker)`, where `assertChecker` (and therefore `check` here) is `(cond, details) => { assert(cond, details); return true; }`, so this `reject` ultimately does send output from ``` X`…` ``` to `assert` in that case. Or are you saying that there's some dependence upon e.g. static analysis detecting literal use of "assert"?

kriskowal: No, my review was too shallow. I see now you’re plumbing through `assert` and that `reject` is not `Promise.reject`.

kriskowal (approved):
Please dismiss my prior review.

PR #1311, opened 2022-10-07
fix(marshal): Increase consistency for encoding/decoding CapData

gibson042:
Should not merge ahead of fixing #1312 unless confirmed safe by @FUDCo.

* Add assertions mirroring Smallcaps
* Consistently order special cases: remotable, promise, error
* Consistently quote "@qclass" in error messages

PR #1313, opened 2022-10-11
feat(marshal): Allow promise stand-ins

### LINKED ISSUE #1312: marshal: Allow promise stand-ins
The [agoric-sdk](https://github.com/Agoric/agoric-sdk) kernel needs to transform marshaled messages, preferably while respecting their encoding by using standard marshal interfaces. But doing so effectively requires opaque treatment of remotables and promises, which is not currently possible with smallcaps because it maintains an encoding-level distinction between the two classes and requires that the latter correspond with actual ECMAScript Promise instances.

Proposed solution: Update `passStyleOf` to support ersatz promises like `{ [PASS_STYLE]: 'promise' }`, allowing `convertValToSlot` and `convertSlotToVal` functions to utilize such objects in place of native promises. How much state they should be allowed to carry is an open question, but my initial inclination is to either prohibit it entirely or to have it match what can be carried by a native promise (e.g., the `{ status: 'fulfilled', value: any } | { status: 'rejected', reason: any }` objects from `Promise.allSettled`.

gibson042:
Stacked on #1311

Fixes #1312

This is the most restrictive solution, in which a pseudo-promise is not
allowed to have any properties or methods (although that may be relaxed
in the future to support a `then` method and/or settlement/resolution
data).

TODO:
- [ ] combine promise and safe-promise helpers (after #1310 )

### DIFF
--- a/packages/marshal/src/helpers/promise.js
+++ b/packages/marshal/src/helpers/promise.js
@@ -64,11 +64,17 @@ const checkPromiseOwnKeys = (pr, check) => {
    * @param {string|symbol} key
    */
   const checkSafeAsyncHooksKey = key => {
-    const val = pr[key];
-    if (val === undefined || typeof val === 'number') {
+    const desc = getOwnPropertyDescriptor(pr, key);
+    if (desc === undefined) {
+      return true;
+    }
+    const isDataDesc = hasOwnPropertyOf(desc, 'value');
+    const val = isDataDesc && desc.value;
+    if (isDataDesc && (val === undefined || typeof val === 'number')) {
       return true;
     }
     if (
+      isDataDesc &&
       typeof val === 'object' &&
       val !== null &&
       isFrozen(val) &&
@@ -92,7 +98,7 @@ const checkPromiseOwnKeys = (pr, check) => {
       reject(
         X`Unexpected Node async_hooks additions to promise: ${pr}.${q(
           String(key),
-        )} is ${val}`,
+        )} is ${desc}`,
       )
     );
   };


packages/marshal/src/helpers/promise.js
```diff
@@ -0,0 +1,45 @@
...
+    );
+  },
+
+  assertValid: candidate => {
+    if (isPromise(candidate)) {
+      assertSafePromise(candidate);
+      return;
+    }
+
+    PromiseHelper.canBeValid(candidate, assertChecker);
```
mhofman, line 34: Seems like we're duplicating the `isPromise` check if the candidate is not a native promise, which means really we're creating 2 throw away `Promise` objects.

packages/marshal/src/helpers/promise.js
```diff
@@ -0,0 +1,45 @@
...
+
+    const proto = getPrototypeOf(candidate);
+    proto === objectPrototype ||
+      proto === null ||
+      assert.fail(X`Unexpected prototype for: ${candidate}`);
+
+    const descKeys = ownKeys(candidate);
+    (descKeys.length === 1 && descKeys[0] === PASS_STYLE) ||
+      assert.fail(X`Must not have any own properties: ${candidate}`);
```
mhofman, line 43: I thought that for @FUDCo's use case, we needed at least a `toString()` method? When discussing this with @erights we didn't see a need to disallow any more methods than for remotables.

FUDCo: Being able to specify the `toString()` method is literally the only thing that matters to me here. Everything else is just machinery to enable that to work.

packages/marshal/src/helpers/promise.js
```diff
@@ -0,0 +1,45 @@
...
+ * @type {import('./internal-types.js').PassStyleHelper}
+ */
+export const PromiseHelper = harden({
+  styleName: 'promise',
+
+  canBeValid: (candidate, check) => {
+    return (
+      (isPromise(candidate) ||
+      candidate[PASS_STYLE] === 'promise' ||
+      check(false, X`Missing ${q(PASS_STYLE)}: ${candidate}`))
```
erights, line 24: The `assertValid` below is not close to adequate input validation. For example, `candidate[PASS_STYLE]` might be an accessor that changes its value each time it is read. However, if we replace these two lines in `canBeValid` with

```suggestion
      checkTagRecord(candidate, 'promise', check)
```

then we get a more accurate check with better diagnostics.

erights (changes requested):
Not approving for two reasons:

  * The requested changes, in particular using `checkTagRecord` for better validation.
  * In light of subsequent discussions, I am queasy about introducing something `passStyleOf` recognizes as a promise type, but which cannot yet work with `E(` nor `E.when(`.

Despite the second bullet, please do respond to requested changes, so we'll be closer to moving ahead if/when the second bullet is resolved. There are two needs this PR would be a step towards:
   * @FUDCo 's immediate need for his encode/decode manipulation in the post-smallcaps SwingSet tests,
   * @mhofman 's plans for virtual persistent promises.

@FUDCo already has something working. I'd like to see that before judging that use to be urgent. Assuming it is not urgent, I'd like to hold this off until we find something that works with `E` and `E.when`, probably as part of doing virtual and durable promises as the same time.

packages/marshal/src/helpers/promise.js
erights, line 43: Please revise to also insist on a `candidate[Symbol.toStringTag] === 'virtual Promise'` or something like that, so that it appears in printed and console output in the same places we currently see an iface or tag. For genuine promises, `p[Symbol.toStringTag] === 'Promise'`, which is why their printed form is already informative.

erights: > ... `toString()` method ...

Cannot do that in any case, and this PR wisely does not attempt. What do you really need? What are you doing now, so that your current state works? My sense is that you only need a pair of functions which can be local to a context, where

```
getSomething(makePromiseWithSomething(something)) === something
```
for some appropriate somethings. We can do that trivially with real promises and a weakmap, yes?

packages/marshal/src/helpers/promise.js
erights: But only one extra one, since we gotta do at least one. Unless we wanna invert the order of tests, which is likely to be net more expensive.

packages/marshal/src/helpers/promise.js
erights: > When discussing this with @erights we didn't see a need to disallow any more methods than for remotables.

True. And what I said then contradicts what I said above. There are two simple extremes, and we should do one or the other.

* Allow mostly-arbitrary unvalidatable methods, like we do for remotables. This is why remotables can have an arbitrary `toString` method.
* Data-only, which is where this PR currently is, and why I'm trying to avoid any unvalidatable behavior.

The problem is that we are still operating under the "any module might be multiply instantiated" restriction, and so any objective predicate, like `passStyleOf`, cannot brand things to remember that they are safe by construction. It can only tell that it is safe by inspection. `passStyleOf` does use a memo table to avoid redundant work, which is much like a brand table. Except that if absent, it always falls back to inspection. The behavior of non-primordial non-global methods cannot be inspected to be safe.

packages/marshal/src/helpers/promise.js
gibson042: It's actually even more than that, because `passStyleOf` itself uses `isPromise` on every [frozen] object, and then every one that fails the check and does not have an explicit PASS_STYLE is subject to iteration over each helper's `canBeValid` in priority order, with promise's using `isPromise` and then each object accepted by that `canBeValid` is passed to `assertValid` (which also uses `isPromise`). I'll reorder to check PASS_STYLE first, but that will still leave two calls for ersatz promises, and if we really care then I think we'll need a WeakMap.

packages/marshal/src/helpers/promise.js
gibson042: This comes down to a tension between practicality and the principle that unmarshaling should recover full state. We can relax this constraint if the use case is in fact sufficiently strong, but we can't go in the other direction.

packages/marshal/src/passStyleOf.js
```diff
@@ -195,6 +197,7 @@ export const passStyleOf = makePassStyleOf([
   CopyRecordHelper,
   TaggedHelper,
   RemotableHelper,
+  PromiseHelper,
```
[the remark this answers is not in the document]
gibson042: Only if we also move the prohibition of a `then` method. It seemed cleaner to keep it at the expense of some redundancy.

packages/marshal/test/test-marshal-capdata.js
```diff
@@ -383,6 +384,35 @@ test('records', t => {
...
 
+test('promises', t => {
+  const { serialize, unserialize } = makeMarshal(
+    val => val,
+    slot => slot,
+    {
+      serializeBodyFormat: 'capdata',
+    },
+  );
+  const assertRoundTrip = (val, body, slots, description) => {
```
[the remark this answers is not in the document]
gibson042: Absolutely!

packages/marshal/src/helpers/promise.js
erights: `passStyleOf` memoizes

packages/marshal/src/helpers/promise.js
gibson042: Should we require it to be an own key, or from the prototype chain, or not care?

packages/marshal/src/helpers/promise.js
gibson042: I have adopted this in `assertValid`, although it does make construction rather cumbersome:
```js
Object.create(Object.prototype, {
  [PASS_STYLE]: { value: 'promise' },
  { [Symbol.toStringTag]: value: 'Pseudo-promise' },
});
```

packages/marshal/src/helpers/promise.js
```diff
@@ -22,27 +27,40 @@ export const PromiseHelper = harden({
...
-    PromiseHelper.canBeValid(candidate, assertChecker);
     if (isPromise(candidate)) {
       assertSafePromise(candidate);
       return;
     }
 
-    const proto = getPrototypeOf(candidate);
-    proto === objectPrototype ||
-      proto === null ||
+    checkTagRecord(candidate, 'promise', assertChecker);
```
erights, line 49: Yes, good choice moving `checkTagRecord` from `canBeValid` to `assertValid`.

No change suggested.

packages/marshal/src/helpers/promise.js
```diff
@@ -22,27 +27,40 @@ export const PromiseHelper = harden({
...
+
+    // XXX Should this (and TaggedHelper.assertValid) support a null prototype?
+    getPrototypeOf(candidate) === objectPrototype ||
       assert.fail(X`Unexpected prototype for: ${candidate}`);
```
erights, line 53: Good question. Yes, I think both should. But here's a counter-argument that says we shouldn't support it at all, i.e., that we should remove such support even from copyRecords:

Any such non-standard `null` prototype is not pass invariant. If such a copyRecord is passed through any marshal barrier, the copyRecord that gets unserialized on the other side will always inherit directly from `Object.prototype'.

So, I am still on the fence. Opinions?

packages/marshal/src/helpers/promise.js
```diff
@@ -22,27 +27,40 @@ export const PromiseHelper = harden({
...
-      assert.fail(X`Must not have any own properties: ${candidate}`);
+    const tagDesc = getOwnPropertyDescriptor(candidate, toStringTag);
+    (tagDesc && tagDesc.value === 'Pseudo-promise') ||
+      assert.fail(
+        X`Pseudo-promise must be an object with ${q(toStringTag)} ${q(
+          'Pseudo-promise',
+        )}: ${candidate}`,
+      );
+    const keys = ownKeys(candidate);
+    (keys.filter(k => k !== PASS_STYLE && k !== toStringTag).length === 0) ||
```
erights, line 63:
```suggestion
    keys.every(k => k === PASS_STYLE || k === toStringTag) ||
```

packages/marshal/src/passStyleOf.js
```diff
@@ -20,7 +21,7 @@ import { assertSafePromise } from './helpers/safe-promise.js';
 /** @typedef {import('./types.js').PassStyleOf} PassStyleOf */
 /** @typedef {import('./types.js').PrimitiveStyle} PrimitiveStyle */
 
-/** @typedef {Exclude<PassStyle, PrimitiveStyle | "promise">} HelperPassStyle */
```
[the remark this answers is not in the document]
erights: No change suggested.

packages/marshal/src/passStyleOf.js
```diff
@@ -195,7 +199,10 @@ export const passStyleOf = makePassStyleOf([
   CopyRecordHelper,
   TaggedHelper,
   RemotableHelper,
+  // TODO: Move ErrorHelper above RemotableHelper
+  // so an Error with no own keys is not rejected as a bad remotable.
   ErrorHelper,
+  PromiseHelper,
```
erights, line 205: Postponing this TODO would be fine. But this PR does move it in `HelperTable`. I prefer these stay in the same order, so do it in both or neither. I prefer both.

packages/marshal/test/test-marshal-capdata.js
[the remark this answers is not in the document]
erights: ;)

No change suggested for this PR.

packages/marshal/test/test-marshal-capdata.js
```diff
@@ -383,6 +384,43 @@ test('records', t => {
...
+        [PASS_STYLE]: { value: 'promise' },
+        [Symbol.toStringTag]: { value: 'Pseudo-promise' },
+      },
+    ),
+  );
+  assertRoundTrip(
+    fakeP,
+    `{"@qclass":"slot","index":0}`,
+    [fakeP],
+    'ersatz promise',
```
erights, line 420: This PR generally calls these 'pseudo-promises', including immediately above, but also immediately above suggests that they are "fake" and "ersatz".

I'm not sure what to do about this terminology problem. According to `passStyleOf` and therefore our distributed objects system, there are promises. When passed across a marshal barrier, they unserialize to genuine promises. Thus, from the perspective of the distributed object system, their promise nature is pass invariant. This is all as it should be.

Normally I'd suggest perhaps "virtual promise", but we should now use that only for the ones that can be paged out. I'm ok with "pseudo", but we should not additionally imply that they are fake or esratz.

erights: Also elsewhere

packages/marshal/test/test-passStyleOf.js
```diff
@@ -55,6 +77,89 @@ test('some passStyleOf rejections', t => {
     message: /\[Promise\]" - Must not have any own properties: \["then"\]/,
   });
 
+  // Pseudo-promise
+  for (const proto of [null, {}]) {
+    const fakeprbadProto = makePseudoPromiseVariant(proto);
```
erights, line 82: This one is genuinely fake. No change suggested.

packages/marshal/test/test-marshal-capdata.js
mhofman: > When passed across a marshal barrier, they unserialize to genuine promises.

They may not.


> I'm ok with "pseudo", but we should not additionally imply that they are fake or esratz.

I'm also ok with pseudo, but would be happy to find another term that can later be exposed in types (since the non native promise will be visible to user-land)

packages/marshal/src/passStyleOf.js
```diff
@@ -39,20 +39,22 @@ const makeHelperTable = passStyleHelpers => {
...
+    styleName in HelperTable ||
+      assert.fail(X`Unrecognized helper: ${q(styleName)}`);
+    HelperTable[styleName] === undefined ||
+      assert.fail(X`conflicting helpers for ${q(styleName)}`);
     HelperTable[styleName] = helper;
   }
+  // "promise" was a late addition, so we tolerate its absence.
+  if (HelperTable.promise === undefined) {
+    delete HelperTable.promise;
+  }
```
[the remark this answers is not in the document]
gibson042: If any code outside of this package calls `makePassStyleOf`, which is exported. However, there are no examples of that anywhere in endo or agoric-sdk AFAICT... should I remove this leeway?

packages/marshal/src/helpers/promise.js
gibson042: I am in favor of removing support for all null-prototype passables unless we have a clear scenario benefiting from them, in which case every check for Object.prototype should be expanded accordingly.

packages/marshal/src/helpers/promise.js
gibson042: D'oh! Thanks.

packages/marshal/src/passStyleOf.js
gibson042: Done: https://github.com/endojs/endo/pull/1313/commits/3dfbfbeacfe8b38bc92b69fd2ce663b03f9697a5

packages/marshal/test/test-marshal-capdata.js
gibson042: I went through a bit of churn before settling on "pseudo-promise", and remain open to alternatives. But in the meantime, I've replaced references to "ersatz promise", although variables in test files still use a "fake" prefix because "pseudo" sucks to type.

packages/marshal/src/passStyleOf.js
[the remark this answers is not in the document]
erights: Where is `makePassStyleOf` exported? I should not be.

packages/marshal/src/helpers/promise.js
erights: I agree. So no change suggested here. Anytime you'd like to propagate this reform (assuming it doesn't break anything but tests), I'll happily review. Thanks!

packages/marshal/test/test-passStyleOf.js
```diff
@@ -45,6 +45,12 @@ test('passStyleOf basic success cases', t => {
   t.is(passStyleOf(harden({ then: 'non-function then ok' })), 'copyRecord');
   t.is(passStyleOf(makeTagged('unknown', undefined)), 'tagged');
   t.is(passStyleOf(harden(Error('ok'))), 'error');
+
+  const hairlessError = Error('hairless');
```
erights, line 49: Cool name!

No change suggested.

erights (commented):
Were we to go ahead with this, its current state would LGTM

packages/marshal/src/helpers/promise.js
gibson042: #1324

packages/marshal/src/passStyleOf.js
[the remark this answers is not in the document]
gibson042: Oh, my mistake! I guess I was looking elsewhere. Fixed.

packages/marshal/test/test-passStyleOf.js
gibson042: https://en.wikipedia.org/wiki/No-hair_theorem :nerd_face:

packages/marshal/src/helpers/promise.js
```diff
@@ -97,32 +112,70 @@ const checkPromiseOwnKeys = (pr, check) => {
...
+      assert.fail(X`Unexpected prototype for: ${candidate}`);
 
-export const assertSafePromise = pr => checkSafePromise(pr, assertChecker);
+    const tagDesc = getOwnPropertyDescriptor(candidate, toStringTag);
+    (tagDesc && tagDesc.value === 'Pseudo-promise') ||
+      assert.fail(
+        X`Pseudo-promise must be an object with ${q(toStringTag)} ${q(
+          'Pseudo-promise',
+        )}: ${candidate}`,
+      );
```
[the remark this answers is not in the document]
gibson042: I figure we can relax this constraint if that need arises.

packages/marshal/src/helpers/promise.js
```diff
@@ -97,32 +112,70 @@ const checkPromiseOwnKeys = (pr, check) => {
...
+    const keys = ownKeys(candidate);
+    keys.every(k => k === PASS_STYLE || k === toStringTag) ||
+      assert.fail(X`Unexpected properties on pseudo-promise ${keys}`);
```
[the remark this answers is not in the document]
gibson042: Done with the `...rest` pattern used elsewhere.

erights:
@mhofman I just created the label `pseudo-promise` and labeled this PR with that label, to keep track of PRs and issues relevant for thinking about possible pseudo-promise semantics.

packages/marshal/src/helpers/promise.js
mhofman: Necroing this thread but an argument could be made that a copyRecord should be unmarshalled with a null prototype instead.

erights: Given that we agree on preserving pass-invariance, there are only two solutions:
   1. All CopyRecords inherit from `Object.prototype` (status quo)
   2. All CopyRecords inherit from `null` (what we’d all prefer if we could have it everywhere at reasonable cost)

Yes, we could change `passStyleOf` and `unserialize` to be consistent with (2). What we cannot reasonably change is requiring everyone to write
```js
harden({
  __proto__: null,
  foo: 3
})
```
rather than the status quo
```js
harden({foo: 3})
```
Yes, there are approaches to this notational constraint, like using something instead of `harden` that either
   * modifies the argument record in place (like `Far` does), or
   * produces a new object derived from the argument object
Altogether, for the way we use CopyRecords everywhere, I don’t think either of these coping strategies are at reasonable notational and explanatory cost. Such a cure would be worse than the disease.

erights:
Is this PR still relevant as a possible step towards virtual and durable promises?

PR #1314, opened 2022-10-12
refactor(marshal): Simplify checkNormalProperty

gibson042:
The `nameType` argument is mostly noise and can only fail at one call site, so this PR moves the relevant check to it.

### DIFF
--- a/packages/marshal/src/helpers/passStyle-helpers.js
+++ b/packages/marshal/src/helpers/passStyle-helpers.js
@@ -55,9 +55,10 @@ export const assertChecker = (cond, details) => {
 harden(assertChecker);
 
 /**
+ * Checks for the presence and enumerability of an own data property.
+ *
  * @param {Object} candidate
  * @param {string|number|symbol} propertyName
- * @param {string} nameType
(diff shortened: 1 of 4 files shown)


packages/marshal/src/helpers/passStyle-helpers.js
```diff
@@ -75,14 +75,6 @@ export const checkNormalProperty = (
...
   }
   return (
-    (nameType === undefined ||
-      // eslint-disable-next-line valid-typeof
-      typeof propertyName === nameType ||
-      reject(
-        X`${q(propertyName)} must be a ${q(
-          nameType,
-        )}-named property: ${candidate}`,
-      )) &&
```
erights, line 85: So was all this simply unneeded?

erights (approved):
LGTM, thanks!

packages/marshal/src/helpers/passStyle-helpers.js
gibson042: Yes, because `propertyName` comes from a literal or number loop iteration variable at all but one call site, and making assertions about its type in those cases is pointless.

PR #1315, opened 2022-10-12
perf(marshal): Skip unnecessary assert.details calls

gibson042:
Ref 11d9f976944b4faaea06037252cfd54e3dd91c37

A modest but significant speedup, especially for big invalid input:

```sh
  import "../init/index.js";
  import { makeTagged } from "./src/makeTagged.js";
  import { makeMarshal } from "./src/marshal.js";
  const { serialize } = makeMarshal(x => x, x => x, { errorTagging: "off", serializeBodyFormat: "smallcaps" });
  const rProto = Object.defineProperties({}, { [Symbol.for("passStyle")]: { value: "remotable" }, [Symbol.toStringTag]: { value: "Remotable" } });
  const inputs = [
    undefined,
    "foo",
    true,
    33,
    33n,
    Symbol.for("foo"),
    Symbol.iterator,
    null,
    Promise.resolve(null),
    [3, 4],
    { foo: 3 },
    { then: "non-function then ok" },
    makeTagged("unknown", undefined),
    Error("ok"),
  ].map(v => harden(v));
  const badInputs = [
    Symbol("unique"),
    {},
    harden(Object.setPrototypeOf(Promise.resolve(), { __proto__: Promise.prototype })),
    harden(Object.assign(Promise.resolve(), { extra: "unexpected own property" })),
    harden(Object.defineProperties(Promise.resolve(), { then: { value: () => "bad then" } })),
    harden({ then: () => "thenable" }),
  ];
  let result;
  (new Benchmark.Suite())
    .add("valid inputs", () => {
      for (const v of inputs) result = serialize(v);
    })
    .add("invalid inputs", () => {
      let caught;
      for (const v of badInputs) {
        caught = undefined;
        try {
          result = serialize(v);
        } catch (err) {
          caught = err;
        } finally {
          if (!caught) throw new Error(`unexpected acceptance: ${v}`);
        }
      }
    })
    .add("big input", () => {
      const input = Array(999).fill().map(() => [...inputs]).concat(Object.create(rProto));
      result = serialize(harden(input));
    })
    .add("big invalid input", () => {
      const input = Array(999).fill().map(() => [...inputs]).concat(Object.create(rProto, { own: {} }));
      let caught;
      caught = undefined;
      try {
        result = serialize(harden(input));
      } catch (err) {
        caught = err;
      }
      if (!caught) throw new Error(`unexpected acceptance`);
    })
    .on("complete", () => result)
    .run({ async: true })
'
```

</details>

before:
```
valid inputs x 5,179 ops/sec ±1.26% (87 runs sampled)
invalid inputs x 2,507 ops/sec ±1.30% (83 runs sampled)
big input x 14.41 ops/sec ±2.08% (37 runs sampled)
big invalid input x 49.72 ops/sec ±3.46% (65 runs sampled)
```
after:
```
valid inputs x 5,908 ops/sec ±1.18% (88 runs sampled)
invalid inputs x 2,565 ops/sec ±1.13% (90 runs sampled)
big input x 14.85 ops/sec ±2.01% (40 runs sampled)
big invalid input x 59.31 ops/sec ±1.16% (61 runs sampled)
```

### DIFF
--- a/packages/marshal/src/helpers/copyArray.js
+++ b/packages/marshal/src/helpers/copyArray.js
@@ -17,15 +17,13 @@ export const CopyArrayHelper = harden({
   styleName: 'copyArray',
 
   canBeValid: (candidate, check) =>
-    isArray(candidate) || check(false, X`Array expected: ${candidate}`),
+    isArray(candidate) ||
+    (!!check && check(false, X`Array expected: ${candidate}`)),
 
   assertValid: (candidate, passStyleOfRecur) => {
     CopyArrayHelper.canBeValid(candidate, assertChecker);
-    assert(
-      getPrototypeOf(candidate) === arrayPrototype,
-      X`Malformed array: ${candidate}`,
-      TypeError,
-    );
+    getPrototypeOf(candidate) === arrayPrototype ||
+      assert.fail(X`Malformed array: ${candidate}`, TypeError);
     // Since we're already ensured candidate is an array, it should not be
     // possible for the following test to fail
     checkNormalProperty(candidate, 'length', false, assertChecker);
@@ -33,12 +31,9 @@ export const CopyArrayHelper = harden({
     for (let i = 0; i < len; i += 1) {
       checkNormalProperty(candidate, i, true, assertChecker);
     }
-    assert(
-      // +1 for the 'length' property itself.
-      ownKeys(candidate).length === len + 1,
-      X`Arrays must not have non-indexes: ${candidate}`,
-      TypeError,
-    );
+    // +1 for the 'length' property itself.
+    ownKeys(candidate).length === len + 1 ||
+      assert.fail(X`Arrays must not have non-indexes: ${candidate}`, TypeError);
     // Recursively validate that each member is passable.
     candidate.every(v => !!passStyleOfRecur(v));
   },
--- a/packages/marshal/src/helpers/copyRecord.js
+++ b/packages/marshal/src/helpers/copyRecord.js
@@ -24,10 +24,10 @@ export const CopyRecordHelper = harden({
   styleName: 'copyRecord',
 
   canBeValid: (candidate, check) => {
-    const reject = details => check(false, details);
+    const reject = !!check && (details => check(false, details));
     const proto = getPrototypeOf(candidate);
     if (proto !== objectPrototype && proto !== null) {
-      return reject(X`Unexpected prototype for: ${candidate}`);
+      return reject && reject(X`Unexpected prototype for: ${candidate}`);
     }
     const descs = getOwnPropertyDescriptors(candidate);
     const descKeys = ownKeys(descs);
@@ -35,14 +35,20 @@ export const CopyRecordHelper = harden({
     for (const descKey of descKeys) {
       if (typeof descKey !== 'string') {
         // Pass by copy
-        return reject(
-          X`Records can only have string-named own properties: ${candidate}`,
+        return (
+          (reject &&
+          reject(
+            X`Records can only have string-named own properties: ${candidate}`,
+          ))
         );
       }
       const desc = descs[descKey];
       if (canBeMethod(desc.value)) {
-        return reject(
-          X`Records cannot contain non-far functions because they may be methods of an implicit Remotable: ${candidate}`,
+        return (
+          (reject &&
+          reject(
+            X`Records cannot contain non-far functions because they may be methods of an implicit Remotable: ${candidate}`,
+          ))
         );
       }
     }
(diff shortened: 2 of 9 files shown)


packages/marshal/src/helpers/copyRecord.js
```diff
@@ -24,25 +24,32 @@ export const CopyRecordHelper = harden({
...
       if (typeof descKey !== 'string') {
         // Pass by copy
-        return reject(
-          X`Records can only have string-named own properties: ${candidate}`,
+        return (
+          (!!check &&
+          check(
+            false,
+            X`Records can only have string-named own properties: ${candidate}`,
+          ))
```
gibson042, line 42: This pattern would arguably be better with ES2020 optional chaining:
```suggestion
          !!check?.(
            false,
            X`Records can only have string-named own properties: ${candidate}`,
          )
```

packages/marshal/src/helpers/copyArray.js
```diff
@@ -17,28 +17,23 @@ export const CopyArrayHelper = harden({
   styleName: 'copyArray',
 
   canBeValid: (candidate, check) =>
-    isArray(candidate) || check(false, X`Array expected: ${candidate}`),
+    isArray(candidate) ||
+    (!!check && check(false, X`Array expected: ${candidate}`)),
```
erights, line 21: How do you feel about dropping the `!!`?

packages/marshal/src/helpers/copyRecord.js
```diff
@@ -24,25 +24,32 @@ export const CopyRecordHelper = harden({
   styleName: 'copyRecord',
 
   canBeValid: (candidate, check) => {
-    const reject = details => check(false, details);
     const proto = getPrototypeOf(candidate);
     if (proto !== objectPrototype && proto !== null) {
-      return reject(X`Unexpected prototype for: ${candidate}`);
+      return !!check && check(false, X`Unexpected prototype for: ${candidate}`);
```
erights, line 29: Interesting that this causes you to revert the `reject` refactoring.

No change suggested.

packages/marshal/src/helpers/remotable.js
```diff
@@ -105,7 +109,10 @@ const checkRemotableProtoOf = (original, check) => {
   // would always return true.
   // @ts-ignore TypeScript assumes what we're trying to check
   if (proto === objectPrototype) {
-    return reject(X`Remotables must be explicitly declared: ${q(original)}`);
+    return (
+      reject &&
+      reject(X`Remotables must be explicitly declared: ${q(original)}`)
```
erights, line 114: Weird that you rescue your `reject` refactoring here but revert it elsewhere. How about rescuing it everywhere? That'll bring us closer to your other refactoring suggestion, where we parameterize these with an optional `reject` directly.

erights (approved):
LGTM, thanks!

erights (commented):
I am confused about the title of this PR. I wouldn't call what you're avoiding "serialization"

packages/marshal/src/helpers/copyArray.js
gibson042: `canBeValid` is typed to return boolean, so dropping that upsets TypeScript:
```sh
$ tsc -p jsconfig.json
src/helpers/copyArray.js:16:14 - error TS2322: Type '{ styleName: "copyArray"; canBeValid: (candidate: any, check: Checker | undefined) => boolean | undefined; assertValid: (candidate: any, passStyleOfRecur: PassStyleOf) => void; }' is not assignable to type 'PassStyleHelper'.
  The types returned by 'canBeValid(...)' are incompatible between these types.
    Type 'boolean | undefined' is not assignable to type 'boolean'.
      Type 'undefined' is not assignable to type 'boolean'.

16 export const CopyArrayHelper = harden({
                ~~~~~~~~~~~~~~~


Found 1 error in src/helpers/copyArray.js:16
```

packages/marshal/src/helpers/copyRecord.js
gibson042: I thought so too. There's a lesson here regarding tagged templates which I cannot fully articulate, although it's also interesting to note that we could alternatively have something like
```js
const reject = (strings, ...args) => !!check && check(false, X(strings, ...args))
…
if (!ok) {
  return reject`Unexpected prototype for: ${candidate}`;
}
```
for the loss of call-site-local `X` and the (hopefully minor) cost of array construction/spread.

packages/marshal/src/helpers/remotable.js
gibson042: Good point; I don't know why I didn't go back to the other files! Done.

gibson042:
> I am confused about the title of this PR. I wouldn't call what you're avoiding "serialization"

Serialization seems to be what dominates the cost of ``` X`…` ```, but it's true that I'm actually avoiding the latter rather than the former. Rephrasing would be welcome.

Issue #1309, opened 2022-10-05
Smallcaps decode chokes on positive bigints under XS


FUDCo:
If you invoke `serialize(0n)` using the marshal package's smallcaps encoding it yields `{ body: '#"+0"', slots: [] }`.  However, if you try to unserialize this while running XS you get
```
SyntaxError: BigInt: Cannot coerce string to bigint
 at BigInt ()
 at unserialize (/Users/chip/Agoric/agoric-sdk/packages/SwingSet/src/supervisors/subprocess-xsnap/supervisor-subprocess-xsnap.js:7284)
 ...
```
This appears to work OK under Node/v8, however.

It appears that the XS implementation of the `BigInt` constructor does not handle the leading `+` character (whereas v8's implementation does).

Looking at the `marshal` package source, the function `decodeFromSmallcaps` contains the code fragment:
```
          case '+':
          case '-': {
            return BigInt(encoding);
          }
```
If this is changed to:
```
          case '+': {
            return BigInt(encoding.slice(1));
          }
          case '-': {
            return BigInt(encoding);
          }
```
the problem goes away. 

Depending on what the ES spec says, this is either a bug in XS or a bug in v8, but, either way, we need a change such as the above in the smallcaps decoder in order for it to work in both places.

mhofman:
If it's a bug in XS I'd like Moddable to fix instead to avoid slices.

I'll run an eshost test and check the spec.

mhofman:
This is confirmed to be an XS parsing bug.
The spec says to try parsing as [`StringIntegerLitteral`](https://tc39.es/ecma262/#prod-StringIntegerLiteral), which includes [`SignedInteger`](https://tc39.es/ecma262/#prod-SignedInteger) which includes [`+ DecimalDigits`](https://tc39.es/ecma262/#prod-DecimalDigits).

eshost also confirms this is an XS only issue.
```
$ eshost -se 'BigInt("+42")'
#### engine262, JavaScriptCore, SpiderMonkey, V8
42

#### Moddable XS

SyntaxError: BigInt: Cannot coerce string to bigint
```

PR #1325, opened 2022-10-13
fix: parse positive bigints correctly despite XS parsing them wrong

FUDCo:
Fixes: #1309

### DIFF
--- a/packages/marshal/src/encodeToSmallcaps.js
+++ b/packages/marshal/src/encodeToSmallcaps.js
@@ -399,7 +399,9 @@ export const makeDecodeFromSmallcaps = ({
               }
             }
           }
-          case '+':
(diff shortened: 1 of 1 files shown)


erights (approved):
LGTM

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -399,7 +399,9 @@ export const makeDecodeFromSmallcaps = ({
               }
             }
           }
-          case '+':
+          case '+': {
+            return BigInt(encoding.slice(1));
```
mhofman, line 403: I'm a little sad we have to do a `.slice` here as that can't be free performance-wise. Let's consider reverting once we land the XS update

PR #1328, opened 2022-10-14
fix(marshal)!: Reject null as a substitute for Object.prototype

### LINKED ISSUE #1324: marshal: null should not be accepted in place of Object.prototype
Per https://github.com/endojs/endo/pull/1313#discussion_r993852290 , a `null` prototype is not pass-invariant—a marshal–unmarshal round-trip will convert it into `Object.prototype`. Accordingly, acceptance of a `null` prototype should be removed from copyRecord until and unless there is a clear need, at which point it should also be added to tagged record/pseudo-promise/etc.

gibson042:
Fixes #1324


packages/marshal/src/helpers/copyRecord.js
```diff
@@ -25,31 +21,27 @@ export const CopyRecordHelper = harden({
...
+      const valid =
+        (typeof key === 'string' ||
+          (!!reject &&
+            reject(
+              X`Records can only have string-named properties: ${candidate}`,
+            ))) &&
+        (!canBeMethod(candidate[key]) ||
+          (!!reject &&
+            reject(
+              X`Records cannot contain non-far functions because they may be methods of an implicit Remotable: ${candidate}`,
```
erights, line 41: Error message needs revision because there is no longer any such thing as an "implicit Remotable". For this PR, a TODO is adequate.

packages/marshal/src/helpers/copyRecord.js
```diff
@@ -25,31 +21,27 @@ export const CopyRecordHelper = harden({
...
-          ))
-        );
+    for (const key of ownKeys(candidate)) {
+      const valid =
+        (typeof key === 'string' ||
+          (!!reject &&
+            reject(
+              X`Records can only have string-named properties: ${candidate}`,
+            ))) &&
+        (!canBeMethod(candidate[key]) ||
```
erights, line 38: Before this PR, `canBeMethod` tests `desc.value`. This PR tests `candidate[key]`. Where does the new code ensure that `candidate` is frozen and has no own accessor properties? Were we previously being reflective unnecessarily? Isn't reflective examination the safer style anyway? Is a reflective check significantly slower on XS?

packages/marshal/src/helpers/passStyle-helpers.js
```diff
@@ -102,36 +105,64 @@ harden(checkNormalProperty);
...
-    checkNormalProperty(tagRecord, Symbol.toStringTag, false, check) &&
-    (typeof getTag(tagRecord) === 'string' ||
-      (reject &&
-        reject(
-          X`A [Symbol.toStringTag]-named property must be a string: ${tagRecord}`,
-        )))
+    actual === expectedPassStyle ||
+    (!!check &&
+      check(
+        false,
```
erights, line 114: `reject` style?

packages/marshal/src/helpers/passStyle-helpers.js
```diff
@@ -102,36 +105,64 @@ harden(checkNormalProperty);
...
+};
+
+export const checkTagRecord = makeCheckTagRecord(
+  (val, proto, check) =>
+    proto === objectPrototype ||
+    (!!check &&
+      check(false, X`A tagRecord must inherit from Object.prototype: ${val}`)),
+);
+
+export const checkFunctionTagRecord = makeCheckTagRecord(
```
erights, line 159: What is our stance on builtin `Function` subclasses, such as async functions, generator functions, or async generators? Can these be far functions? These do not *directly* inherit from `Function.prototype`, but they do inherit from known primordial prototypes.

erights (commented):
Just comments so far. Review not done yet.

packages/marshal/src/helpers/remotable.js
```diff
@@ -182,16 +150,12 @@ const checkRemotable = (val, check) => {
...
-const isRemotable = val => checkRemotable(val, x => x);
-
 /** @type {MarshalGetInterfaceOf} */
 export const getInterfaceOf = val => {
-  const typestr = typeof val;
   if (
-    (typestr !== 'object' && typestr !== 'function') ||
-    val === null ||
+    !isObject(val) ||
     val[PASS_STYLE] !== 'remotable' ||
```
erights, line 157: Is this line still necessary? Isn't it covered by the next?

erights (approved):
LGTM, thanks!

packages/marshal/src/helpers/remotable.js
erights: ok

packages/marshal/src/helpers/copyRecord.js
gibson042: Done.

packages/marshal/src/helpers/passStyle-helpers.js
gibson042: This preserves the existing behavior, which allows far async functions if their prototype chain has been suitably manipulated (e.g., `setPrototypeOf(async () => {}, makeTagishRecord('Remotable', getPrototypeOf(async () => {})))`) while rejecting generators and async generators—and classes, for that matter—by dint of a non-removable "prototype" property ("_Far functions unexpected properties besides .name and .length_"). But an (async)generator is basically a callable factory for (async)iterator objects which at their core are a bag of inherited methods, so I could definitely see extending support in that direction, probably with a helper like `ExoGenerator(async function* () { … })` to alter the prototype chain and harden responses.

PR #1330, opened 2022-10-16
fix(marshal): Return a special error message from passStyleOf(typedArray)

### LINKED ISSUE #1326: TypedArrays are not passable, but not because they need to be hardened
At time of writing, attempting to send a message that contains a `TypedArray` produces an error to the effect `Cannot pass non-frozen objects like (an object). Use harden()`. This is confusing because `harden` has a special case for `TypedArrays`, the result of which is that the TypedArray becomes non-extensible, all its properties become non-configurable, but its content remains mutable, much like a hardened `Map`. But, like a `Map`, it should not be passable because of that content-mutability.

The expected behavior is that sending a message containing a TypedArray should produce a rejection with an Error that states that TypedArrays are not passable.

- [Alternate hardening behavior for TypedArrays](https://github.com/endojs/endo/blob/ab31d51fc2df0355b45da8bbeff892c45d81c98a/packages/ses/src/make-hardener.js#L76-L110)
- [Exceptionally non-obvious defensive brand-check for typed arrays](https://github.com/endojs/endo/blob/ab31d51fc2df0355b45da8bbeff892c45d81c98a/packages/ses/src/make-hardener.js#L58-L74)

We might avoid duplicating the brand check by producing an `isHardened` global that generalizes `isFrozen`. We’ve avoided this in the past because most places where you’d reach for `isHardened` are places that would be better served by calling `harden` regardless of whether it was previously hardened.

gibson042:
Fixes #1326

### DIFF
--- a/packages/marshal/src/passStyleOf.js
+++ b/packages/marshal/src/passStyleOf.js
@@ -3,7 +3,11 @@
 /// <reference types="ses"/>
 
 import { isPromise } from '@endo/promise-kit';
-import { isObject, PASS_STYLE } from './helpers/passStyle-helpers.js';
+import {
+  isObject,
+  isTypedArray,
+  PASS_STYLE,
+} from './helpers/passStyle-helpers.js';
 
 import { CopyArrayHelper } from './helpers/copyArray.js';
 import { CopyRecordHelper } from './helpers/copyRecord.js';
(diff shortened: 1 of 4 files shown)


packages/marshal/src/passStyleOf.js
```diff
@@ -141,10 +145,15 @@ const makePassStyleOf = passStyleHelpers => {
...
             return 'null';
           }
-          (isFrozen(inner)) ||
+          if (!isFrozen(inner)) {
             assert.fail(
-              X`Cannot pass non-frozen objects like ${inner}. Use harden()`,
+              // TypedArrays get special treatment in harden()
+              // and a corresponding special error message here.
+              isTypedArray(inner)
+                ? X`Cannot pass mutable typed arrays like ${inner}.`
```
kriskowal, line 153:
```suggestion
                ? X`Cannot pass typed arrays like ${inner}`
```

This will be a good place to recommend CopyBytes when they’re an option.

kriskowal (approved):

erights (approved):
LGTM

PR #1332, opened 2022-10-19
core: release

kriskowal:
- chore: Update release notes
- chore(release): publish


mhofman (approved):
Unfortunate dirty history, but besides that LGTM

Issue #1281, opened 2022-09-11
Uncaught ReferenceError: harden is not defined


Jack-Works:
This is wrong. It crashes in our environment.

![image](https://user-images.githubusercontent.com/5390719/189535478-707a6938-c148-40a9-8537-398d22280b1c.png)

Because when this function is called, there is no `harden` defined yet.

_Originally posted by @Jack-Works in https://github.com/endojs/endo/pull/1241#discussion_r967845817_

Jack-Works:
hello 👀

kriskowal:
👋 I’m working on this. Thanks for your patience.

I found that I need to bypass all function repairs if eval is disabled, but I don’t feel good about doing that without also verifying that the other intrinsic evaluators are inert. So, I’ve done that, and now I’m bolstering the no-eval emulation environment to make all the evaluators inert, not just `eval` proper. So at this point, I’m improving the tests.

plusminushalf:
It crashes for us as well, do we have any update on this issue?

Environment: Chrome extension, Manifest version v3, loading SES in an iframe in one of the HTML pages with the page being part of the v3 sandbox environment.

plusminushalf:
This isn't an issue with version 0.12.4 but rather with version 0.18.0

damianofalcioni:
Sorry to reopen this issue but the same error still occur in Node for latest SES from npm (1.9.0):
```
import 'ses';

const compartment = new Compartment({
  globals: {
    print: harden(console.log)
  },
  __options__: true
});
compartment.evaluate('print("Hello World");');
```
Return:

```
file:///...
    print: harden(console.log)
           ^

ReferenceError: harden is not defined
    at file:///....
    at ModuleJob.run (node:internal/modules/esm/module_job:222:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:316:24)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:123:5)

Node.js v20.16.0
```

kriskowal:
@damianofalcioni In the above snippet, call `lockdown` after importing `ses` and it will expose `harden`. Using `harden` before lockdown would prevent subsequent `lockdown` from succeeding and using `harden` without `lockdown` and `Compartment` provides no guarantee of isolation.

damianofalcioni:
Thanks, that work!
As a suggestion, please fix the readme code (https://github.com/endojs/endo/tree/master/packages/ses#compartment) that propose this sample without lockdown()

Issue #903, opened 2021-09-21
Fail-safe for environments without `unsafe-eval` CSP


mhofman:
SES relies on dynamic eval for its own initialization and virtualizing Compartment evaluators.

Some environments prevent execution of such dynamic eval, such as a browser with [`Content-Security-Policy`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src) not including `'unsafe-eval'` in its `script-src` policy.

SES's initialization does some dynamic eval to [tame the Function constructors](https://github.com/endojs/endo/blob/1ee0a1f46f549710aa6ab1bb5c9f6af7f52c89d2/packages/ses/src/tame-function-constructors.js). However this currently includes an idempotency test which is too lax, and fails open if `'unsafe-eval'` is not allowed. Also, a dynamic eval is considered for addition to make SES more resilient against transpilers (#902).

https://github.com/endojs/endo/blob/1ee0a1f46f549710aa6ab1bb5c9f6af7f52c89d2/packages/ses/src/tame-function-constructors.js#L46-L53

While this is not really a security issue since the Compartment API doesn't work at all without `'unsafe-eval'`, changes like the one proposed in #881 are looking into using the Compartment feature without requiring dynamic eval of code. The actual taming doesn't have to rely on dynamic eval anymore than the current way anonymous intrinsics are looked up by direct syntax. For that matter, the logic for looking up the intrinsic could probably be deduplicated between `repairFunction` and `getAnonymousIntrinsics`.

It could also be argued that in an environment with no dynamic eval, the intrinsic `eval` and `Function` constructors are not feral and thus safe to leave in place. However making this behavior an explicit opt-in would be preferable, if at all allowed, instead of the current fail-open situation.

PR #1333, opened 2022-10-20
fix: crash under csp

Jack-Works:
close https://github.com/endojs/endo/issues/1281

### DIFF
--- a/packages/ses/src/tame-function-constructors.js
+++ b/packages/ses/src/tame-function-constructors.js
@@ -49,7 +49,7 @@ export default function tameFunctionConstructors() {
     FERAL_FUNCTION.prototype.constructor('return 1');
   } catch (ignore) {
     // Throws, no need to patch.
-    return {};
+    return freeze({});
   }
 
   const newIntrinsics = {};
--- a/packages/ses/src/tame-function-constructors.js
+++ b/packages/ses/src/tame-function-constructors.js
@@ -49,7 +49,7 @@ export default function tameFunctionConstructors() {
     FERAL_FUNCTION.prototype.constructor('return 1');
   } catch (ignore) {
     // Throws, no need to patch.
-    return harden({});
+    return {};
   }
 
   const newIntrinsics = {};
--- a/packages/ses/src/tame-function-constructors.js
+++ b/packages/ses/src/tame-function-constructors.js
@@ -5,6 +5,7 @@ import {
   defineProperties,
   getPrototypeOf,
   setPrototypeOf,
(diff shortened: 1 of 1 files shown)


kriskowal (commented):
Freeze would suffice here

packages/ses/src/tame-function-constructors.js
```diff
@@ -49,7 +49,7 @@ export default function tameFunctionConstructors() {
     FERAL_FUNCTION.prototype.constructor('return 1');
   } catch (ignore) {
     // Throws, no need to patch.
-    return harden({});
+    return {};
```
kriskowal, line 52:
```suggestion
    return Object.freeze({});
```

kriskowal: Will also need to import freeze from commons.js

mhofman:
Not in scope for this PR, but wondering what kind of test we could have to make sure we don't break no-eval environments. The setup would likely look like replacing the evaluators with functions that throw before importing ses. The question is how to scope a subset of the tests with this environment.

Jack-Works:
since Compartment is not available under no-eval environment, test lockdown is enough.

kriskowal:
> Not in scope for this PR, but wondering what kind of test we could have to make sure we don't break no-eval environments. The setup would likely look like replacing the evaluators with functions that throw before importing ses. The question is how to scope a subset of the tests with this environment.

As it happens, you wrote one already. https://github.com/endojs/endo/blob/master/packages/ses/test/test-no-eval.js

It’s marked as `failing`. I’d like this PR to remove the `failing` bit from that test. I looked into it and this change isn’t sufficient. Thankfully, @mhofman left notes: https://github.com/endojs/endo/issues/903

```
  no-eval › lockdown must not throw when eval is forbidden

  Error thrown in test:

  TypeError {
    message: 'no unsafe-eval, as if by content-security-policy',
  }

  › globalThis.eval (file://test/no-eval.js:6:9)
  › repairFunction (file://src/tame-function-constructors.js:71:35)
  › tameFunctionConstructors (file://src/tame-function-constructors.js:118:3)
  › repairIntrinsics (file://src/lockdown-shim.js:276:17)
  › lockdown (file://src/lockdown-shim.js:410:28)
  › file://test/test-no-eval.js:9:30
```

kriskowal (approved):

kriskowal:
I’m landing this as it’s incremental progress. Further work will be necessary to avoid eval with a lockdown flag without compromising lockdown invariants.

PR #1334, opened 2022-10-20
fix: fail template

erights:
See https://github.com/Agoric/agoric-sdk/pull/6482

Prior to this PR, our new assertion style let to many statements of the form
```js
condition || assert.fail(X`...`);
```
with this PR, we can now write
```js
condition || Fail`...`;
```
which is more compact and more readable.

### DIFF
--- a/packages/captp/src/atomics.js
+++ b/packages/captp/src/atomics.js
@@ -1,6 +1,6 @@
 /// <reference types="ses"/>
 
-const { details: X } = assert;
+const { details: X, Fail } = assert;
 
 // This is a pathological minimum, but exercised by the unit test.
 export const MIN_DATA_BUFFER_LENGTH = 1;
@@ -23,9 +23,7 @@ const STATUS_FLAG_REJECT = 4;
  */
 const splitTransferBuffer = transferBuffer => {
   transferBuffer.byteLength >= MIN_TRANSFER_BUFFER_LENGTH ||
-    assert.fail(
-      X`Transfer buffer of ${transferBuffer.byteLength} bytes is smaller than MIN_TRANSFER_BUFFER_LENGTH ${MIN_TRANSFER_BUFFER_LENGTH}`,
-    );
+    Fail`Transfer buffer of ${transferBuffer.byteLength} bytes is smaller than MIN_TRANSFER_BUFFER_LENGTH ${MIN_TRANSFER_BUFFER_LENGTH}`;
   const lenbuf = new BigUint64Array(transferBuffer, 0, 1);
 
   // The documentation says that this needs to be an Int32Array for use with
@@ -40,9 +38,7 @@ const splitTransferBuffer = transferBuffer => {
   );
   const databuf = new Uint8Array(transferBuffer, overheadLength);
   databuf.byteLength >= MIN_DATA_BUFFER_LENGTH ||
-    assert.fail(
-      X`Transfer buffer of size ${transferBuffer.byteLength} only supports ${databuf.byteLength} data bytes; need at least ${MIN_DATA_BUFFER_LENGTH}`,
-    );
+    Fail`Transfer buffer of size ${transferBuffer.byteLength} only supports ${databuf.byteLength} data bytes; need at least ${MIN_DATA_BUFFER_LENGTH}`;
   return harden({ statusbuf, lenbuf, databuf });
 };
 
--- a/packages/marshal/src/encodeToCapData.js
+++ b/packages/marshal/src/encodeToCapData.js
@@ -38,7 +38,7 @@ const {
   fromEntries,
   freeze,
 } = Object;
-const { details: X, quote: q } = assert;
+const { details: X, Fail, quote: q } = assert;
 
 /**
  * Special property name that indicates an encoding that needs special
@@ -80,14 +80,11 @@ const qclassMatches = (encoded, qclass) =>
  * ) => Encoding} [encodeErrorToCapData]
  */
 
-const dontEncodeRemotableToCapData = rem =>
-  assert.fail(X`remotable unexpected: ${rem}`);
+const dontEncodeRemotableToCapData = rem => Fail`remotable unexpected: ${rem}`;
 
-const dontEncodePromiseToCapData = prom =>
-  assert.fail(X`promise unexpected: ${prom}`);
+const dontEncodePromiseToCapData = prom => Fail`promise unexpected: ${prom}`;
 
-const dontEncodeErrorToCapData = err =>
-  assert.fail(X`error object unexpected: ${err}`);
+const dontEncodeErrorToCapData = err => Fail`error object unexpected: ${err}`;
 
 /**
  * @param {EncodeToCapDataOptions} encodeOptions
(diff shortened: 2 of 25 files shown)


erights:
automated by regexp search-replace in vscode using
```
assert\.fail\((?:\n|\s)*X`((?:\n|[^`])*?)`,?(?:\n|\s)*\);$

Fail`$1`;
```

erights:
Oops. I violated our recent process agreement and squashed. But it was a trivial squash, and AFAIK no one has yet started to review. Speaking of which, ping ;)

(Low priority, but should be an easy review and does make our code prettier.)

packages/marshal/src/encodeToCapData.js
```diff
@@ -221,33 +218,30 @@ export const makeEncodeToCapData = ({
         if (qclassMatches(encoded, 'slot')) {
           return encoded;
         }
-        assert.fail(
-          X`internal: Remotable encoding must be an object with ${q(
-            QCLASS,
-          )} ${q('slot')}: ${encoded}`,
-        );
+        // `throw` is noop since `Fail` throws. But linter confused
+        throw Fail`internal: Remotable encoding must be an object with ${q(
```
turadg, line 222: Not a blocker, but I'd like to see if we can fix this before merge. If not the at least having an issue and a caveat comment on Fail.

packages/ses/src/error/assert.js
```diff
@@ -375,6 +375,9 @@ const makeAssert = (optRaise = undefined, unredacted = false) => {
   };
   freeze(fail);
 
+  /** @type {FailTag} */
```
turadg, line 378: Will anything else implement FailTag? If not I think we're better off inlining the type here and when the type of this needs to be referenced using "typeof"

packages/marshal/src/encodeToCapData.js
turadg: I tried this in [a playground](https://www.typescriptlang.org/play?target=99&noFallthroughCasesInSwitch=true&ts=5.0.0-dev.20221103#code/GYVwdgxgLglg9mABFAFgJzgdwM4AoCUAXImAKYBupaiA3gFCKPLpYmmaICiaGaBdAXzqhIsBMlLYoAVTBpSAQwgoFAIwA2pAIwFaDJqgw4CAbn2NziAPRXk2RMHUKA5vdQx7C++HlKVG0ksIBGw4TQA6dThnXAAiD0QfRWU1TQBCWPxBOjooAE8AB1JEABFSKAUYdXsAXkSwAGswLDAzYLApRAATcsrqxDrcKFIAWwKnYeJwJpaAGkRwxYU0V2IFMDyAbQBdfAGAPkRN4bGJ0nnF8OXXbcQvUt6q7DM6G2RC4tweiqfiMp-qkQ2JQ0MJwNB4EhgH0AGJgL6Pap-RHYIFkEG0ZhGNgcbi8BEAzyeDb4EyIIQiCHiYZSWS+FIBABMunoTAcsPhAHJfKEwJzSZZLG8oPZHC43CgEvckn5UoE2W8etgYPIuncwGrgDARd1JDBnEgUFAoAVsIQbM5tSgQKpwsERlYRjAIBhQsAoFYACofADKLpgBQ9HmwIEkVgAzABWAAMjIAbABiYOh+0jUhgKAAWkjkcZjOjcYAHHmgiEwqRItE4gkZQz0plsq9bPkiogCX0zQ9CXsaod0VQ6O1OtCqgBhEJQAZt74d5Hd4j96i9zGGVhkXE8OB8GdPO7EvKk8lg0SQiS0uTJfyacMsywj9TjjpQXDcxS8-lmNlC5v2LpwSR8pOYrOMwUreBesoBOE36ICMCh5KoxTYKQkiINqe6IPIUAgGgYAwGAIEKIg5AKOooaIHAwDvK2i4wUaJpmhaVo2nacAOk6LpwG6HrekUfpoAGQbYCGYYACwAJziTGpYdOWlYxPE4H0lepAZFkQjHlSSAwn0QyjOMCiTIgnr6WcPpQAJBHYAAgjw8GIAAPogUiWc4OxohQVB6Gyq4cOuXCbnwJwGcM4QAFZwPhBACgIiCaWISA0jIEF1qQom3myOlVAABjyCDZZ+TAwTq95uHADhOCB7hEvUylyjBcEIcUUDlX+iCYFaiD0aa5pWJaqAsfajrOq6lE8b6-qBlYyZhjG4ZxuGSbCSmbFphmmZaIyWjiYWWhaKJokAOwyaEERRApNYpSpanZEAA) and took some notes. It appears to be a Design Limitation. 
https://github.com/microsoft/TypeScript/issues/50363#issuecomment-1219811447

We could file a request for support for never templates so there's a record of this case. WDYT?

Either way, I think the jsdoc for `Fail` should document the limitation and explain when to use `throws` or not.

erights: > We could file a request for support for never templates so there's a record of this case. WDYT?

I think that's a good idea. Would you like to?

erights: https://github.com/microsoft/TypeScript/issues/50363never mentions templates per se. Is it possible to adapt the workaround explained at https://github.com/microsoft/TypeScript/issues/50363#issuecomment-1219811447 (which I did not understand)?

packages/ses/src/error/assert.js
erights: But assert.js imports types.js, and types.js uses `FailTag`, but if we inlined it, we'd inline it in assert.js. Advice? How do you want this layered?

packages/marshal/src/encodeToCapData.js
turadg: I've filed it here https://github.com/microsoft/TypeScript/issues/51426

Please include it in the jsdoc for the `Fail` function.

packages/ses/src/error/assert.js
turadg: Ah. Transitively we need an interface, `Assert`, so this is appropriate

erights: > so this is appropriate

Which "this"? Do you mean that the status quo FailTag of this PR is fine as is?

turadg: status quo

erights: Cool. Resolving. Thanks.

packages/captp/src/atomics.js
```diff
@@ -1,6 +1,6 @@
 /// <reference types="ses"/>
 
-const { details: X } = assert;
+const { details: X, Fail } = assert;
```
kriskowal, line 3: I would favor a lower-case name since this is not a constructor. I feel the same way about `X` but it’s at least a local alias. I recognize that’s a difficult proposition since we already have an `assert.fail`. Since this throws, perhaps consider `raise` (nodding to Python) or `failWithDetails` (aliased locally to `fail` or `Fail`).

I might be convinced if we’re going to establish a convention to reserve BactrianCase for both constructors and template functions, but precedent of `String.raw` seems like a fixed point that would never be consistent.

kriskowal (approved):
Looks mechanical to me. I register a naming preference below, which I will hold weakly because it’s a naming preference.

packages/captp/src/atomics.js
erights: TIL "BactrianCase". When googling it just now, my favorite comment (at https://news.ycombinator.com/item?id=965405 ) is

> I've known it as lowerCamel and UpperCamel

erights: OTOH, https://en.wikipedia.org/wiki/Bactrian_camel#/media/File:2011_Trampeltier_1528.JPG is memorable ;)

packages/marshal/src/encodeToCapData.js
erights: Done

packages/captp/src/atomics.js
erights: > which I will hold weakly because it’s a naming preference.

Thanks for noting that. I have weak agreements with your criticism, counter-arguments for your concrete suggestions, and no better alternative to offer.

   * `failWithDetails` - way too long. For this to be pleasant to use for what we want people to use it for, the template tag name must remain small. Indeed, I don't find `failWithDetails` more pleasant than the `assert.fail(X...)` pattern it would replace.
   * `raise` - we're already using this term to mean something related but different. We're not using it a lot, so we could change this too. But then we'd need a new name for that concept. Frankly, the concept currently called `raise` is the better meaning for `raise`.

So, since your naming preference is weak, I'm resolving this without renaming `Fail`.

kriskowal: For what it’s worth, people who like to be right instead of awesome just call them camelCase and PascalCase.

PR #1335, opened 2022-10-20
perf(marshal): Memoize checkRemotable

gibson042:
Without this, marshaling a remotable incurs a redundant verification in `getInterfaceOf`.

```sh
node --input-type=module -e '
  import "../init/index.js";
  import { makeMarshal } from "./src/marshal.js";
  const { serialize } = makeMarshal(x => x, x => x, { errorTagging: "off", serializeBodyFormat: "smallcaps" });
  let result;
  (new Benchmark.Suite())
    .add("shallow remotable", {
      setup: () => {
	const rProto = Object.defineProperties({}, { [Symbol.for("passStyle")]: { value: "remotable" }, [Symbol.toStringTag]: { value: "Remotable" } });
        shallow = Object.create(rProto);
        Array(20).fill().forEach((_, i) => { shallow[`method${i}`] = () => {}; });
      },
      fn: () => { result = serialize(harden(shallow)); },
    })
    .add("deep remotable", {
      setup: () => {
	const rProto = Object.defineProperties({}, { [Symbol.for("passStyle")]: { value: "remotable" }, [Symbol.toStringTag]: { value: "Remotable" } });
        deep = rProto;
        for (let i = 0; i < 6; i++ ) {
          deep = Object.create(deep);
          Array(20).fill().forEach((_, i) => { deep[`method${i}`] = () => {}; });
        }
      },
      fn: () => { result = serialize(harden(deep)); },
    })
    .on("complete", () => result)
    .run({ async: true })
'
```

</details>

before:
```
shallow remotable x 32,886 ops/sec ±0.96% (89 runs sampled)
deep remotable x 10,695 ops/sec ±1.45% (86 runs sampled)
```

after:
```
shallow remotable x 59,592 ops/sec ±1.61% (84 runs sampled)
deep remotable x 56,173 ops/sec ±2.12% (79 runs sampled)
```

### DIFF
--- a/packages/marshal/src/helpers/remotable.js
+++ b/packages/marshal/src/helpers/remotable.js
@@ -134,11 +134,25 @@ const checkRemotableProtoOf = (original, check) => {
 };
 
 /**
+ * Keep a weak set of confirmed remotables for marshal performance
+ * (without which we would incur a redundant verification in
+ * getInterfaceOf).
+ * We don't remember rejections because they are possible to correct
+ * with e.g. `harden`.
+ *
+ * @type {WeakSet<Remotable>}
+ */
+const confirmedRemotables = new WeakSet();
+
+/**
  * @param {Remotable} val
  * @param {Checker} [check]
  * @returns {boolean}
  */
 const checkRemotable = (val, check) => {
+  if (confirmedRemotables.has(val)) {
+    return true;
+  }
   const reject = !!check && (details => check(false, details));
   if (!isFrozen(val)) {
     return reject && reject(X`cannot serialize non-frozen objects like ${val}`);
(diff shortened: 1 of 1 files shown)


erights (changes requested):
Locally this is all fine. But this memo's contents are redundant with passStyleOf's memo map from passable -> passStyle. 

I do understand why `getInterfaceOf` cannot simply call `passStyleOf` --- because it rejects only by throwing, it has no identCheck mode. But still, please find a way to reuse that one memo rather than creating a new one.

gibson042:
There are some obstacles to that...
* **module structure**: passStyleOf.js imports from helpers/remotable.js, and IIRC we try to avoid cycles.
* **scope**: the passStyle memo is part of the `makePassStyleOf` closure rather than being a top-level binding.
* **API alignment**: `getInterfaceOf` does not accept a `passStyleOf`/`passStyleOfRecur` argument, and it may be presumptuous to assume use of the one defined in passStyleOf.js.

None of these are insurmountable, but they do give me pause. How important is it to avoid a remotable-specific memoization?

erights (approved):
> ... How important is it to avoid a remotable-specific memoization?

All good points. And yes, not so important that this PR needs to wait on it. If we want to get rid of the redundant representation, then we can fix that separately later. Could you file a bug recording your points? You can assign it to me. With that bug filed, this PR LGTM.

gibson042:
> Could you file a bug recording your points?

#1339

PR #1336, opened 2022-10-20
test(marshal): Clean up encoding tests

gibson042:
Eliminate magic strings and an unnecessary helper function.


erights (approved):
LGTM, I am really glad to see that cleaned up, thanks!

PR #1338, opened 2022-10-20
fix: restore capdata leniency

erights:
capdata is being deprecated. It exists only to avoid breaking old clients that have not yet updated to smallcaps. However, I accidentally introduced an error check that I did not realize would be a breaking change. This removes that extra check, enabling old code to continue working the old way until it upgrades to smallcaps.


gibson042 (approved):
😢

Issue #904, opened 2021-09-22
Scope proxy leak allow constructive access to global lexicals


mhofman:
A property of the `globalLexicals` option of the `Compartment` constructor is that they can only be accessed lexically, which can be denied to evaluated code by simple code transforms. However combined with the leak of scope proxy objects (#31), evaluate code can constructively reach the global lexicals.

```js
const c = new Compartment({}, { globalLexicals: { foo() { return 'secret'; } });
c.evaluate(`
  globalThis.getThis = function() { return this; };
  const hiddenFoo = getThis()[['f', 'o', 'o'].join('')];
  hiddenFoo()
`);
```

Unlike #876 this affects all evaluators.

mhofman:
Now that #1293 has landed, the global lexical leak can only happen through the lexicals created for module bindings which currently use the same object. #1341 should fix that.

PR #1341, opened 2022-10-26
fix(ses): Patch leak of globalLexicals thru moduleLexicals

### LINKED ISSUE #912: Module shims lexicals leak internal live bindings object

kriskowal:
Fixes #912

The global lexicals are intended to provide a namespace that guest code cannot enumerate for arbitrary properties.  By employing a transform for all evaluation in a compartment and also transforming all module code provided to a compartment's import hooks, a host program can introduce a name like `meter` into the global lexical scope of a program and also deny that program access to the lexical name through censorship of the original source.  Previously, a crafted module could use a function in module lexical scope to capture the global lexicals object or the scope proxy.  This change creates another intermediate layer in the safe evaluator that separates the global lexicals from the module lexicals, plugging this leak.

Individually reviewable commits.

- test(ses): Leak globalLexicals through moduleLexicals
- refactor(ses): Decouple moduleLexicals from globalLexicals
- test(ses): Cover moduleLexicals optimizer overshadowing

### DIFF
--- a/packages/ses/src/compartment-evaluate.js
+++ b/packages/ses/src/compartment-evaluate.js
@@ -3,7 +3,6 @@ import {
   TypeError,
   arrayPush,
   create,
-  defineProperties,
   getOwnPropertyDescriptors,
 } from './commons.js';
 import {
@@ -30,8 +29,11 @@ export const provideCompartmentEvaluator = (compartmentFields, options) => {
     let { globalTransforms } = compartmentFields;
     const { globalObject, globalLexicals } = compartmentFields;
 
-    let localObject = globalLexicals;
-    if (__moduleShimLexicals__ !== undefined) {
+    let moduleLexicals;
+    if (
+      __moduleShimLexicals__ !== undefined &&
+      __moduleShimLexicals__ !== null
+    ) {
       // When using `evaluate` for ESM modules, as should only occur from the
       // module-shim's module-instance.js, we do not reveal the SES-shim's
       // module-to-program translation, as this is not standardizable behavior.
@@ -42,16 +44,16 @@ export const provideCompartmentEvaluator = (compartmentFields, options) => {
       // and `import`, at the expense of being tightly coupled to SES-shim.
       globalTransforms = undefined;
 
-      localObject = create(null, getOwnPropertyDescriptors(globalLexicals));
-      defineProperties(
-        localObject,
+      moduleLexicals = create(
+        null,
         getOwnPropertyDescriptors(__moduleShimLexicals__),
       );
     }
 
     ({ safeEvaluate } = makeSafeEvaluator({
       globalObject,
-      globalLexicals: localObject,
+      globalLexicals,
+      moduleLexicals,
       globalTransforms,
       sloppyGlobalsMode,
     }));
--- a/packages/ses/src/make-evaluate.js
+++ b/packages/ses/src/make-evaluate.js
@@ -26,14 +26,20 @@ function buildOptimizer(constants, name) {
  *
  * @param {object} context
  * @param {object} context.evalScope
- * @param {object} context.globalLexicals
  * @param {object} context.globalObject
+ * @param {object} context.globalLexicals
+ * @param {object} context.moduleLexicals
  * @param {object} context.scopeTerminator
  */
 export const makeEvaluate = context => {
-  const { globalObjectConstants, globalLexicalConstants } = getScopeConstants(
+  const {
+    globalObjectConstants,
+    globalLexicalConstants,
+    moduleLexicalConstants,
+  } = getScopeConstants(
     context.globalObject,
     context.globalLexicals,
+    context.moduleLexicals,
   );
   const globalObjectOptimizer = buildOptimizer(
     globalObjectConstants,
@@ -42,6 +48,10 @@ export const makeEvaluate = context => {
   const globalLexicalOptimizer = buildOptimizer(
     globalLexicalConstants,
     'globalLexicals',
+  );
+  const moduleLexicalOptimizer = buildOptimizer(
+    moduleLexicalConstants,
+    'moduleLexicals',
   );
 
   // Create a function in sloppy mode, so that we can use 'with'. It returns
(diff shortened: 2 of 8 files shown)


mhofman:
> Fixes #912

It would fix #904, not #912, which is a problem self contained within module lexicals

packages/ses/src/compartment-evaluate.js
```diff
@@ -30,8 +29,11 @@ export const provideCompartmentEvaluator = (compartmentFields, options) => {
     let { globalTransforms } = compartmentFields;
     const { globalObject, globalLexicals } = compartmentFields;
 
-    let localObject = globalLexicals;
-    if (__moduleShimLexicals__ !== undefined) {
+    let moduleLexicals;
+    if (
+      __moduleShimLexicals__ !== undefined &&
+      __moduleShimLexicals__ !== null
```
mhofman, line 35: This is really one of those cases where I prefer to write `__moduleShimLexicals__ != null`

packages/ses/src/make-evaluate.js
```diff
@@ -26,14 +26,20 @@ function buildOptimizer(constants, name) {
  *
  * @param {object} context
  * @param {object} context.evalScope
- * @param {object} context.globalLexicals
  * @param {object} context.globalObject
+ * @param {object} context.globalLexicals
+ * @param {object} context.moduleLexicals
```
mhofman, line 31: wondering if we should keep the order the same as in the nested scopes

packages/ses/src/make-evaluate.js
```diff
@@ -43,6 +49,10 @@ export const makeEvaluate = context => {
     globalLexicalConstants,
     'globalLexicals',
   );
+  const moduleLexicalOptimizer = buildOptimizer(
+    moduleLexicalConstants,
+    'moduleLexicals',
+  );
```
mhofman, line 55: Can this ever actually happen? I mean this is probably the safest way, and doesn't cost much.

mhofman (commented):
Preliminary observations, not a full review yet, and nothing actionable, just stream of thoughts.

So I'm wondering, should `moduleLexicals` be allowed to shadow `globalLexicals`, I suppose a global lexical should come with a transform denying normal lexical access to these bindings, but I don't quite remember if that global transform would apply before or after the module transforms which introduce the corresponding module lexical usages.

Then, if no shadowing is allowed between global and module lexical, and given the use case of `moduleLexicals` which do not get optimized (not immutable since they're getter / setters), I'm wondering if the removal of the evaluate factory wasn't premature. If we had kept the evaluate factory, we technically could have cached it on the compartment, and re-used it between `evaluate` calls with different `__moduleShimLexicals__`, but that kinda feels like we'd want another kind of 2 stage process where only the `moduleLexicals` change instead of the whole `context` object.

Finally, I really hope we get to a point where we can build the evaluate factory more dynamically if `globalLexicals` or `moduleLexicals` are missing / empty, removing layers of unnecessary `with` blocks siting between the user code and lookups on the `globalObject`.

packages/ses/src/make-evaluate.js
kriskowal: Not in any correct arrangement. Any potential overlap would need to be censored by a transform.

kriskowal:
In out-of-band group review, we elected to instead remove support for `globalLexicals` entirely #1343

PR #1342, opened 2022-10-26
Bump types deps

turadg:
Bumps TypeScript and tsd deps.

Also enables `checkJs` in `@endo/marshal`.


kriskowal (approved):
🙏

mhofman (commented):
I am somewhat concerned about the spread of `ts-expect-error`, which in IMO can be worse than an `any` cast.

My concern with them is that they're not targeted, and can hide multiple independent errors on the same line. If there was a way to target the error code ([open issue](https://github.com/microsoft/TypeScript/issues/19139)), I wouldn't be so concerned, but sprinkling them in the method declaration here has the risk of hiding refactor bugs.

I'd much prefer giving in and changing the JSDoc type in this case if we must update the TS version.

mhofman (approved):

PR #1343, opened 2022-10-26
feat(ses)!: Remove support for globalLexicals

### LINKED ISSUE #904: Scope proxy leak allow constructive access to global lexicals

kriskowal:
Fixes #904

*BREAKING CHANGE*: Removes support for `globalLexicals`.  To our knowledge, there are no production uses for `globalLexicals`.  They currently could leak because `moduleLexicals` and `globalLexicals` used the same scope object, so properties of one would leak to the other with crafted modules.  We had an opportunity to plug the leak at the cost of a fifth scope in all evaluators, but elected to remove the unnecessary complexity instead.


kriskowal:
In lieu of #1341

mhofman:
@warner do we still use the meter transform in agoric-sdk, or is this safe to land?

mhofman (approved):
Looking good.

Layering question, should the second commit removing `globalLexicals` from compartment-mapper not come first before the removal from the shim?

kriskowal:
@mhofman It’s perfectly sensible to reverse the commit order. Note that I’ve added another one to propagate the change to `@endo/import-bundle`.

packages/import-bundle/src/compartment-wrapper.md
```diff
@@ -45,11 +45,12 @@ function getOdometer() {
 }
 
 const inescapableTransforms = [milageTransform];
-const inescapableGlobalLexicals = { getOdometer };
-const WrappedCompartment = wrapInescapableCompartment(Compartment,
-                                                      inescapableTransforms,
-                                                      inescapableGlobalLexicals);
-const endowments = {};
+const WrappedCompartment = wrapInescapableCompartment(
```
warner, line 48: note: the prose at the beginning of this file references the feature being removed, search for `provide additional global lexicals to its environment` and maybe remove it

warner: In fact, I think this example needs rethought.. the imposed transform will emit code that still works (it tries to call `getOdometer()`, which still works, but as a global global, not a global-lexical), but the user code can use `globalThis['get'+'Odometer']().read()`, which it could not do before

We don't need that restriction in agoric-sdk anymore, and I totally support removing it, but maybe the example we use here should emphasize the transform without also mentioning a confinement behavior that no longer exists.

warner (approved):
LGTM, but consider the docs example

PR #1350, opened 2022-11-08
fix: assert touchups

erights:
Now that https://github.com/endojs/endo/pull/1334 has been merged, this PR converts remaining stragglers in endo from the old `assert(` style to the new `||` style. Convert to `|| Fail...` when possible; otherwise still to `|| assert.fail(X...)`.

### DIFF
--- a/packages/check-bundle/lite.js
+++ b/packages/check-bundle/lite.js
@@ -4,7 +4,7 @@
 import { decodeBase64 } from '@endo/base64/decode.js';
 import { parseArchive } from '@endo/compartment-mapper/import-archive.js';
 
-const { details: d, quote: q } = assert;
+const { Fail, details: d, quote: q } = assert;
 
 /**
  * Verifies that a bundle passes its own integrity checks or rejects the
@@ -29,13 +29,13 @@ export const checkBundle = async (
       bundle,
     )}`,
   );
-  assert(bundle !== null, d`checkBundle expects a bundle object`);
-  assert(
-    Object.isFrozen(bundle),
-    `checkBundle cannot vouch for the ongoing integrity of an unfrozen object, got ${q(
+  if (bundle === null) {
+    throw Fail`checkBundle expects a bundle object`;
+  }
+  Object.isFrozen(bundle) ||
+    Fail`checkBundle cannot vouch for the ongoing integrity of an unfrozen object, got ${q(
       bundle,
-    )}`,
-  );
+    )}`;
   const properties = Object.entries(Object.getOwnPropertyDescriptors(bundle));
   const nonValues = properties.filter(
     ([, property]) => typeof property.get === 'function',
@@ -43,16 +43,14 @@ export const checkBundle = async (
   const nonStrings = properties.filter(
     ([, property]) => typeof property.value !== 'string',
   );
-  assert(
-    nonValues.length === 0 && nonStrings.length === 0,
-    `checkBundle cannot vouch for the ongoing integrity of a bundle ${q(
+  (nonValues.length === 0 && nonStrings.length === 0) ||
+    Fail`checkBundle cannot vouch for the ongoing integrity of a bundle ${q(
       bundleName,
     )} with getter properties (has ${nonValues.map(
       ([name]) => name,
     )}) or non-string value properties (has ${nonStrings.map(
       ([name]) => name,
-    )})`,
-  );
+    )})`;
 
   const { moduleFormat } = bundle;
   assert.typeof(
@@ -83,14 +81,12 @@ export const checkBundle = async (
     moduleFormat === 'getExport' ||
     moduleFormat === 'nestedEvaluate'
   ) {
-    assert.fail(
-      `checkBundle cannot determine hash of bundle with ${moduleFormat} moduleFormat because it is not necessarily consistent`,
-    );
+    Fail`checkBundle cannot determine hash of bundle with ${q(
+      moduleFormat,
+    )} moduleFormat because it is not necessarily consistent`;
   } else {
-    assert.fail(
-      d`checkBundle cannot determine hash of bundle with unrecognized moduleFormat ${q(
-        moduleFormat,
-      )}`,
-    );
+    Fail`checkBundle cannot determine hash of bundle with unrecognized moduleFormat ${q(
+      moduleFormat,
+    )}`;
   }
 };
--- a/packages/compartment-mapper/src/compartment-map.js
+++ b/packages/compartment-mapper/src/compartment-map.js
@@ -1,6 +1,10 @@
 // @ts-check
 /// <reference types="ses"/>
 
+// TODO Since this file pervasively uses simple template strings rather than
+// template strings tagged with `assert.details` (aka `X`), and since it uses
+// this definition of `q` rather than `assert.quote`, we have not yet
+// converted it to the new `||` assert style.
 const q = JSON.stringify;
 
 const moduleLanguages = [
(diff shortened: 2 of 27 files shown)


erights:
A bad search led me to miss many stragglers. Putting into draft until more of these are fixed.

packages/check-bundle/lite.js
```diff
@@ -29,30 +29,28 @@ export const checkBundle = async (
...
     ([, property]) => typeof property.get === 'function',
   );
   const nonStrings = properties.filter(
     ([, property]) => typeof property.value !== 'string',
   );
-  assert(
-    nonValues.length === 0 && nonStrings.length === 0,
-    `checkBundle cannot vouch for the ongoing integrity of a bundle ${q(
+  (nonValues.length === 0 && nonStrings.length === 0) ||
+    Fail`checkBundle cannot vouch for the ongoing integrity of a bundle ${q(
```
erights, line 47: Reviewers, note that the original did not say `X` and some of the substitutions do not have `q`, so this change is not equivalence preserving. Is this a problem? Should the remaining substitutions be unredacted with `q`?

What the correct answer is depends on the intention of the code. Please look for this possible issue everywhere the original did not have an `X` and advise. Thanks.

packages/ses/src/lockdown-shim.js
```diff
@@ -191,26 +191,23 @@ export const repairIntrinsics = (options = {}) => {
...
-  assert(
-    priorLockdown === undefined,
-    `Already locked down at ${priorLockdown} (SES_ALREADY_LOCKED_DOWN)`,
-    TypeError,
-  );
+  priorLockdown === undefined ||
+    // eslint-disable-next-line @endo/no-polymorphic-call
+    assert.fail(
+      d`Already locked down at ${priorLockdown} (SES_ALREADY_LOCKED_DOWN)`,
+      TypeError,
```
erights, line 209: Reviewers,
I left this as `assert.fail(d...)` rather than converting to `Fail...` because of the additional `TypeError`. This occurs several places. For each of these, if the `TypeError` is extraneous and you would prefer that I convert to the `Fail...` form, please let me know.

erights:
This is now ready for review.

packages/compartment-mapper/src/compartment-map.js
```diff
@@ -1,6 +1,10 @@
 // @ts-check
 /// <reference types="ses"/>
 
+// TODO Since this file pervasively uses simple template strings rather than
+// template strings tagged with `assert.details` (aka `X`), and since it uses
+// this definition of `q` rather than `assert.quote`, we have not yet
+// converted it to the new `||` assert style.
```
turadg, line 7: nit: say on the TODO line what there is to do.
```suggestion
// TODO convert to the new `||` assert style.
// Deferred because this file pervasively uses simple template strings rather than
// template strings tagged with `assert.details` (aka `X`), and uses
// this definition of `q` rather than `assert.quote`
```

packages/eventual-send/src/handled-promise.js
```diff
@@ -208,23 +208,23 @@ export const makeHandledPromise = () => {
...
   function baseHandledPromise(executor, pendingHandler = undefined) {
-    assert(new.target, X`must be invoked with "new"`);
+    new.target || Fail`must be invoked with "new"`;
     let handledResolve;
     let handledReject;
     let resolved = false;
     let resolvedTarget = null;
     let handledP;
     let continueForwarding = () => {};
+    const assertNotYetForwarded = () => {
```
turadg, line 218: nice added improvement

packages/marshal/src/encodeToCapData.js
```diff
@@ -435,12 +433,10 @@ export const makeDecodeFromCapData = ({
...
           if (hasOwnPropertyOf(jsonEncoded, 'rest')) {
-            assert(
-              typeof rest === 'object' &&
-                rest !== null &&
-                ownKeys(rest).length >= 1,
-              X`Rest encoding must be a non-empty object: ${rest}`,
-            );
+            (typeof rest === 'object' &&
+              rest !== null &&
+              ownKeys(rest).length >= 1) ||
```
turadg, line 438: up above you had `if (rest === null) {` instead of `||`. That made it a little easier to read.

Consider,
```
const isNonEmptyObject = (typeof rest === 'object' &&
…
isNonEmptyObject || Fail`Rest encoding must be a non-empty object: ${rest}`;
```

packages/marshal/src/helpers/symbol.js
```diff
@@ -12,13 +12,11 @@ const wellKnownSymbolNames = new Map(
       name => typeof name === 'string' && typeof Symbol[name] === 'symbol',
     )
     .filter(name => {
-      assert(
-        // @ts-ignore It doesn't know name cannot be a symbol
-        !name.startsWith('@@'),
-        X`Did not expect Symbol to have a symbol-valued property name starting with "@@" ${q(
+      // @ts-ignore It doesn't know name cannot be a symbol
```
turadg, line 15:
```suggestion
      // @ts-expect-error It doesn't know name cannot be a symbol
```

packages/ses/index.test-d.ts
```diff
@@ -206,4 +206,4 @@ expectType<Error>(assert.error(X`details are ${stringable}`, TypeError));
 
 expectType<Error>(assert.error(X`details are ${stringable}`, TypeError, { errorName: 'Nom de plum' }));
 
-expectType<never>(assert.fail(X`details are ${stringable}`));
```
turadg, line 209: we shouldn't lose this test about `assert.fail`

packages/compartment-mapper/src/compartment-map.js
erights: Done

packages/marshal/src/encodeToCapData.js
erights: I like it. Done.

packages/marshal/src/helpers/symbol.js
erights: Done

packages/ses/index.test-d.ts
erights: If I include both, lint says the second is unreachable. Since, IIUC, the point of this test is to statically test what typescript says and the unreachability error is lint, not ts, I'm suppressing the lint error. But I'm not really sure I understand `expectType` enough, so I'm leaving this conversation unresolved until you PTAL. Thanks.

turadg: That's fine. Another option would be to put it into a function body.

turadg (approved):

packages/check-bundle/lite.js
gibson042: Should we update `assert` to detect when the output from `quote` is serialized by anything other than `details` and surface the problem by something like `console.error` and/or creating an unhandled rejection and/or forcing a nonzero process exit status?

erights: I don't understand the question. Let's be sure to discuss in our upcoming meeting.

erights: Oops. We forgot to discuss this. Could you expand or clarify a bit here?

gibson042: I was brainstorming ways to detect and address accidental omission of `X` in assertion arguments, but the real risk is placeholder expressions that _don't_ use `q`/`assert.quote`, and it's now basically moot anyway because we no longer encourage or expect direct use of `X`/`assert.details` to construct arguments. So I consider this resolved.

PR #1352, opened 2022-11-09
test(static-module-record): add failing test for "immer@9.0.6"

kumavis:
encountered this while attempting to bundle the metamask-extension code base


kumavis:
this call should always fail (?)
https://github.com/endojs/endo/blob/725b987ffa812a070ff45fcd496cf8fd88df6963/packages/static-module-record/src/babelPlugin.js#L197
bc `topLevelExported[name]` is falsey
but the first thing we do is `topLevelExported[name].forEach`

packages/static-module-record/test/test-static-module-record.js
```diff
@@ -30,6 +32,14 @@ function assertDefaultExport(t, record) {
   t.deepEqual(record.__liveExportMap__, {});
 }
 
+function readFixture(filename) {
+  return fs.readFileSync(
+    new URL(filename, `${new URL(path.dirname(import.meta.url))}/fixtures/`)
+      .pathname,
```
kriskowal, line 38: For the benefit of Windows in CI, something like:
```suggestion
    url.fileURLToPath(new URL(filename, new URL('fixtures/', import.meta.url)))
```

With
```js
import url from 'url';
```

kriskowal: Added a commit to this effect.

PR #1353, opened 2022-11-09
Static module record: error with module url on transform fail

kumavis:
before:
```
TypeError#1: Cannot read properties of undefined (reading 'forEach')
```

after:
```
SyntaxError#1: Error transforming source in "file:///home/xyz/Development/metamask-extension4/node_modules/immer/dist/immer.esm.js": Cannot read properties of undefined (reading 'forEach')
```

### DIFF
--- a/packages/static-module-record/src/transform-analyze.js
+++ b/packages/static-module-record/src/transform-analyze.js
@@ -32,7 +32,16 @@ const makeCreateStaticRecord = transformSource =>
       // Comment out the shebang lines.
       moduleSource = `//${moduleSource}`;
     }
-    const scriptSource = transformSource(moduleSource, sourceOptions);
+    let scriptSource;
+    try {
+      scriptSource = transformSource(moduleSource, sourceOptions);
+    } catch (err) {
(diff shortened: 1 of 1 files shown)


packages/static-module-record/src/transform-analyze.js
```diff
@@ -32,7 +32,16 @@ const makeCreateStaticRecord = transformSource =>
       // Comment out the shebang lines.
       moduleSource = `//${moduleSource}`;
     }
-    const scriptSource = transformSource(moduleSource, sourceOptions);
+    let scriptSource;
+    try {
+      scriptSource = transformSource(moduleSource, sourceOptions);
+    } catch (err) {
+      const moduleLocation = url ? `"${url}"` : '<unknown>';
```
kriskowal, line 39: Quoting strings, generally.
```suggestion
      const moduleLocation = url ? JSON.stringify(url) : '<unknown>';
```

kriskowal (approved):
Please preserve your commit message in the git log, not the PR title.

PR #1354, opened 2022-11-09
WIP experiment only: regen yarn.lock experiment

erights:
Just seeing what happens if I delete yarn.lock and regenerate it from scratch using `yarn && yarn build`

similar: https://github.com/Agoric/agoric-sdk/pull/6549

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -72,10 +72,10 @@
   dependencies:
     "@babel/highlight" "^7.0.0"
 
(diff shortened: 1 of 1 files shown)


erights:
Under CI, only lint failed, and only in `node_modules/events/events.js` as used by ses-ava

erights:
Too stale to be relevant. Closing.

PR #1355, opened 2022-11-09
compartment-mapper: add support for package.json exotic dependency fields

### LINKED ISSUE #1177: Compartment mapper support for peerDependencies
Compartment mapper does not yet provide any behavior for `peerDependencies`. If I understand the purpose of that feature correctly, the compartment mapper should treat these the same way as `dependencies` and that the distinction between the two kinds of dependencies is otherwise only meaningful to the package manager.

If you can understand https://nodejs.org/es/blog/npm/peer-dependencies/, you’re smarter than me.

kumavis:
Fixes #1177 and related compatibility issues

### DIFF
--- a/packages/compartment-mapper/src/node-modules.js
+++ b/packages/compartment-mapper/src/node-modules.js
@@ -252,16 +252,37 @@ const graphPackage = async (
   graph[packageLocation] = /** @type {Node} */ (result);
 
   /** @type {Record<string, string>} */
-  const dependencies = {};
+  const dependencyLocations = {};
   const children = [];
-  const predicates = packageDescriptor.dependencies || {};
-  assign(predicates, packageDescriptor.peerDependencies || {});
-  assign(predicates, packageDescriptor.bundleDependencies || {});
+  const optionals = new Set();
+  const {
+    dependencies = {},
+    peerDependencies = {},
+    peerDependenciesMeta = {},
+    bundleDependencies = {},
+    optionalDependencies = {},
+    devDependencies = {},
+  } = packageDescriptor;
+  const allDependencies = {};
+  assign(allDependencies, dependencies);
+  assign(allDependencies, peerDependencies);
+  for (const [name, { optional }] of Object.entries(peerDependenciesMeta)) {
+    if (optional) {
+      optionals.add(name);
+    }
+  }
+  assign(allDependencies, bundleDependencies);
+  assign(allDependencies, optionalDependencies);
+  Object.keys(optionalDependencies).forEach(name => {
+    optionals.add(name);
+  });
 
   if (dev) {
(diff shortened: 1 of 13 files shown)


packages/compartment-mapper/src/node-modules.js
```diff
@@ -252,24 +252,49 @@ const graphPackage = async (
...
+    devDependencies = {},
+  } = packageDescriptor;
+  const allDependencies = {};
+  assign(allDependencies, dependencies);
+  assign(allDependencies, peerDependencies);
+  Object.keys(peerDependenciesMeta).forEach(name => {
+    if (peerDependenciesMeta[name].optional) {
+      optionals.add(name);
+    }
+  });
```
kriskowal, line 273: Please consider, in keeping with the local rustic æsthetic:

```suggestion
  for (const [name, { optional }] of Object.entries(peerDependenciesMeta)) {
    if (optional) {
      optionals.add(name);
    }
  }
```

If we went more toward the array functional style, I’d be suggesting `entries`, `filter` and `forEach(set.add, set)`, but my prevailing preference is to use loops.

The contents of `package.json` are user data to us. We probably should be validating the package descriptor with something similar to what we use to validate compartment maps, before we even see it here. I don’t think we’re validating package descriptors at all.

Filed https://github.com/endojs/endo/issues/1357

packages/compartment-mapper/src/node-modules.js
```diff
@@ -252,24 +252,49 @@ const graphPackage = async (
   graph[packageLocation] = /** @type {Node} */ (result);
 
   /** @type {Record<string, string>} */
-  const dependencies = {};
+  const dependenciesObj = {};
```
kriskowal, line 255: Please consider a meaningful distinction in the name.
```suggestion
  const dependencyLocations = {};
```

packages/compartment-mapper/test/fixtures-optional/node_modules/optional-cjs/index.js
```diff
@@ -0,0 +1,24 @@
...
+  } catch (err) {
+    results.errors.one = err;
+  }
+  try {
+    results.exports.two = require('missing-two');
+  } catch (err) {
+    results.errors.two = err;
+  }
+  return results
+}
```
kriskowal, line 24: Such Windows file without newline on the end.

kriskowal (approved):
Yeah, that ought to do the trick.

packages/compartment-mapper/test/fixtures-optional/node_modules/optional-cjs/index.js
kumavis: tell it to the linter rules :tongue:

PR #1356, opened 2022-11-09
feat(compartment-mapper): Bundle the unarchiver

kriskowal:
- feat(compartment-mapper): Bundles evaluate to entrypoint namespace
- test(compartment-mapper): Verify that importArchive can be bundled by makeBundle
- test(compartment-mapper): Verify bundles can transit package boundaries

### DIFF
--- a/packages/compartment-mapper/src/import-archive.js
+++ b/packages/compartment-mapper/src/import-archive.js
@@ -304,7 +304,8 @@ export const parseArchive = async (
       __shimTransforms__,
       Compartment,
     });
-    return compartment.import(moduleSpecifier);
+    // eslint-disable-next-line dot-notation
+    return compartment['import'](moduleSpecifier);
   };
 
   return { import: execute, sha512 };
--- a/packages/compartment-mapper/test/test-main.js
+++ b/packages/compartment-mapper/test/test-main.js
@@ -1,6 +1,6 @@
 import 'ses';
 import test from 'ava';
-import { loadLocation, importArchive } from '../index.js';
+import { loadLocation, importArchive, makeBundle } from '../index.js';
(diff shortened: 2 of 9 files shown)


packages/compartment-mapper/src/import-archive.js
```diff
@@ -308,7 +308,8 @@ export const parseArchive = async (
       __shimTransforms__,
       Compartment,
     });
-    return compartment.import(moduleSpecifier);
+    // eslint-disable-next-line dot-notation
+    return compartment['import'](moduleSpecifier);
```
kumavis, line 312: whats this change for?

packages/compartment-mapper/test/test-main.js
```diff
@@ -83,6 +83,42 @@ with the current test fixture.`);
...
+  const archiverBundle = await makeBundle(readPowers.read, archiverLocation);
+  const archiverCompartment = new Compartment({
+    TextEncoder,
+    TextDecoder,
+    URL,
+    assert,
+  });
+  const evasiveArchiverBundle = archiverBundle
+    .replace(/(?<!\.)\bimport\b(?![:"'])/g, 'IMPORT')
+    .replace(/\beval\b/g, 'EVAL');
```
kumavis, line 103: can we reuse the transforms or would that setup a dep weird topology

kriskowal: The evasive transform (not the censor) is Babel based and in import-bundle, this the layering concern, and why I had to manually evade the compartment.import call.

packages/compartment-mapper/src/import-archive.js
kriskowal: Manual evasive transform for the `import` censor.

kumavis (approved):

PR #1360, opened 2022-11-11
fix(ses): Lockdown under no-unsafe-eval Content-Security-Policy

kriskowal:
This change adds a test that verifies that Lockdown runs to completion even in the absence of a usable evaluator.


kriskowal:
cc @Jack-Works. I’m leaving this in draft because it does not address your concern about Lockdown generating excessive warnings. But, this should subsume all of #1333. I have not rigorously accounted for @mhofman’s concerns in #903. Posted as draft to show progress.

Jack-Works (approved):

packages/ses/src/tame-function-constructors.js
```diff
@@ -45,11 +50,40 @@ import {
...
+          // eslint-disable-next-line no-empty-function
+          async function unsafeAsyncFunction() {},
+          // eslint-disable-next-line no-empty-function
+          async function* unsafeAsyncGeneratorFunction() {},
+        ],
+        f => f.constructor,
+      ),
+      constructor => {
+        try {
+          constructor('');
```
mhofman, line 70: I may be paranoid, but wouldn't there technically be a way to bypass this by allowing the "empty" / null sha in CSP?

I guess it'd really be the embedder shooting itself in the foot in that case.

mhofman (commented):
> I have not rigorously accounted for @mhofman’s concerns in #903.

I'll probably have to page this back in memory as I do not currently recall my previous investigation and work in this area.

kriskowal:
This remains in draft because we need to reframe the solution in terms of a lockdown option that expressly avoids any spelling of `eval`, and attempts to preserve the invariants of lockdown with the _assumption_ that any attempt by guest code to use any spelling of `eval` will be thwarted by the host.

cc @erights

PR #1365, opened 2022-11-15
chore(import-bundle): Not yet TypeScriptable

kriskowal:
In order to obviate a patch in Agoric SDK, this change marks `importBundle` as explicitly not yet ready to be mechanically verified by TypeScript.

### DIFF
--- a/packages/import-bundle/src/index.js
+++ b/packages/import-bundle/src/index.js
@@ -1,3 +1,7 @@
+// XXX Omit from typecheck for TypeScript packages depending upon
+// import-bundle.
+// TODO https://github.com/endojs/endo/issues/1254
+// @ts-nocheck
 /* global globalThis */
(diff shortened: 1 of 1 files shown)


packages/import-bundle/src/index.js
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
turadg, line 1: imo, this is the concern of agoric-sdk since this package doesn't have any `tsc`.

I'll approve per your inter-repo maintenance judgement but I think there should be a comment like,
```
XXX omit from typecheck by consumers
TODO https://github.com/endojs/endo/issues/1254
```

turadg (approved):

kriskowal:
Holding in abeyance. May be obviated by https://github.com/endojs/endo/pull/1366

PR #1366, opened 2022-11-15
1254 import bundle

turadg:
Part of https://github.com/endojs/endo/issues/1254


packages/import-bundle/src/index.js
```diff
@@ -52,6 +53,7 @@ export async function importBundle(bundle, options = {}) {
     const { namespace } = await archive['import']({
       globals: endowments,
       __shimTransforms__: transforms,
+      // @ts-expect-error TS2740 missing properties from type
```
kriskowal, line 56: I’m curious why `Compartment` is not more specifically typed here, since it should be in the `/// <reference types="ses"/>` above.

kriskowal (approved):
Looks like progress to me.

packages/import-bundle/src/index.js
turadg: The error message without it:
> Type 'typeof Compartment' is missing the following properties from type 'Compartment': globalThis, evaluate, import, load, and 2 more.

Issue #1364, opened 2022-11-14
compartment-mapper/importHook - npm package exports field defines both '.' and './'


kumavis:
compatibility issue with [tslib@1.14.1](https://npmfs.com/package/tslib/1.14.1/)

package.json defines unusually and maybe invalid exports field that defines both '.' and './' https://npmfs.com/package/tslib/1.14.1/package.json
```json
    "exports": {
        ".": {
            "module": "./tslib.es6.js",
            "import": "./modules/index.js",
            "default": "./tslib.js"
        },
        "./": "./"
    }
```

kumavis:
[`tslib@2.4.0`](https://npmfs.com/package/tslib/2.4.0/package.json#L29) has a similar issue
```json
    "exports": {
        ".": {
            "module": "./tslib.es6.js",
            "import": "./modules/index.js",
            "default": "./tslib.js"
        },
        "./*": "./*",
        "./": "./"
    }
```

kriskowal:
@kumavis I don’t recall whether we ever added support for wildcards, much less support `imports`. Adding this issue to the `review` label.

PR #1369, opened 2022-11-15
bundle - metamask extension bundling compatibility (unrefined)

kumavis:
hack and slash monkey patching to enable loading and linking of metamask dependency graph

will be split into a handful of small PRs

**to be broken out:**
- [x] **importHook**: support resolution candidates when extension omitted but "." present
https://github.com/endojs/endo/pull/1370
- [x] **infer-exports**: handle browser field more correctly
https://github.com/endojs/endo/pull/1373
- [x] **node-modules/digest**: uses wrong packageLocation
https://github.com/endojs/endo/pull/1372
- [x] **bundler**: handle dev mode https://github.com/endojs/endo/pull/1400
- [x] **link**: fallback to extension as language and dont error until after transform
https://github.com/endojs/endo/pull/1371
- [x] **importHook**: support resolution candidates when '.' in packageDir means main
    which involves threading the packageDescriptor or less
https://github.com/endojs/endo/pull/1374
- [x] **bundler**: handle tags https://github.com/endojs/endo/pull/1404
- [x] **bundler**: handle deferredErrors https://github.com/endojs/endo/pull/1403
- [x] **importHook** allow alternate candidateSuffixes https://github.com/endojs/endo/pull/1405
- [x] **bundler**: handle pre-cjs-json modules https://github.com/endojs/endo/pull/1383
- [x] **bundler**: handle exitModules or "global internal aliases" or "dependency injections" https://github.com/endojs/endo/pull/1406
- [ ] **importHook**: include parentSpecifier for debugging
- [ ] **node-modules/graphPackage**: wrap error for more clarity

getting the loading and linking working also required a number of changes to metamask-extension, including many patches to dependencies' `package.json`s to fix missing or incorrect specified dependencies. some of these may be fixable inside `compartment-mapper`, for example: https://github.com/endojs/endo/issues/1364
see https://github.com/MetaMask/metamask-extension/pull/16520 for the changes necessary in metamask-extension

### DIFF
--- a/packages/compartment-mapper/src/bundle.js
+++ b/packages/compartment-mapper/src/bundle.js
@@ -46,6 +46,7 @@ const parserForLanguage = {
  * @param {Record<string, ResolveHook>} compartmentResolvers
  * @param {string} entryCompartmentName
  * @param {string} entryModuleSpecifier
+ * @param {Record<string, object>} exitModules
  */
 const sortedModules = (
   compartmentDescriptors,
@@ -53,6 +54,7 @@ const sortedModules = (
   compartmentResolvers,
   entryCompartmentName,
   entryModuleSpecifier,
+  exitModules,
 ) => {
   const modules = [];
   const seen = new Set();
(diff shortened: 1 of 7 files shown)


kriskowal:
@kumavis Just from reviewing titles, I’m only worried about the one where we recognize extensions for languages that require translation. For that, I’m expecting to need nothing short of a new public API for translation, and I’ll be wanting to take the opportunity to confine translators, as well as provide alternative browser and non-browser translation environments. Looking forward to seeing what that change actually does!

kumavis:
also see https://github.com/LavaMoat/LavaMoat/issues/421

kumavis:
:vulcan_salute:

PR #1370, opened 2022-11-16
fix(compartment-mapper): allow specifier to include period and omit extension

kumavis:
specifier `./util.thing` may refer to `./util.thing.js`


kriskowal (changes requested):
Needs a test. Looks good.

kriskowal (approved):
🙏

PR #1371, opened 2022-11-16
fix(compartment-mapper): allow transforms to run on unknown languages

kumavis:
this allowed me to add transforms for typescript files, that transform them into cjs modules.

The extension is used when the language is unknown. It still enforces that there is a parser for the language as a result of the transformations.

```js
const moduleTransforms = {
  // ...
    'extension:ts': async (bytes, specifier, location, packageLocation) => {
      let parserLanguage = 'cjs'
      const moduleRecord = {
        file: location.toString(),
        source: bytes.toString(),
      }
      await runMetaMaskTransforms(moduleRecord)
      const transformedSource = Buffer.from(moduleRecord.source, 'utf8')
      return { bytes: transformedSource, parser: parserLanguage }
    },
 }
 ```

### DIFF
--- a/packages/compartment-mapper/src/link.js
+++ b/packages/compartment-mapper/src/link.js
@@ -95,12 +95,7 @@ const makeExtensionParser = (
     ) {
       language = languageForModuleSpecifier[specifier];
     } else {
-      if (!has(languageForExtension, extension)) {
-        throw new Error(
-          `Cannot parse module ${specifier} at ${location}, no parser configured for extension ${extension}`,
-        );
-      }
-      language = languageForExtension[extension];
+      language = languageForExtension[extension] || extension;
     }
 
     if (has(moduleTransforms, language)) {
--- a/packages/compartment-mapper/src/link.js
+++ b/packages/compartment-mapper/src/link.js
@@ -76,14 +76,14 @@ const extensionImpliesLanguage = extension => extension !== 'js';
  * the type of a module is implied by package.json and should not be inferred
  * from its extension.
  * @param {Record<string, ParserImplementation>} parserForLanguage
- * @param {ModuleTransforms} transforms
+ * @param {ModuleTransforms} moduleTransforms
  * @returns {ParseFn}
  */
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by kriskowal — CHANGES_REQUESTED** [review]
Needs a test that exercises the case where a module exists with an extension that is neither supported by the compartment-mapper nor provided by the user.

File: packages/compartment-mapper/src/link.js [hunk 1]
```diff
@@ -95,12 +95,7 @@ const makeExtensionParser = (
...
-      if (!has(languageForExtension, extension)) {
-        throw new Error(
-          `Cannot parse module ${specifier} at ${location}, no parser configured for extension ${extension}`,
-        );
-      }
-      language = languageForExtension[extension];
+      language = languageForExtension[extension] || `extension:${extension}`;
     }
 
     if (has(transforms, language)) {
```
kumavis, line 101: this could be a `while` loop instead

kumavis: >   107:38  error  Unexpected `await` inside a loop  no-await-in-loop

kumavis: hmm actually this would break for things that dont change the language type but expect to run only once and not infinite times... :skull:

kumavis:
alternate prefix could be `unknown:ts`

packages/compartment-mapper/src/link.js
```diff
@@ -95,16 +95,11 @@ const makeExtensionParser = (
     ) {
       language = languageForModuleSpecifier[specifier];
     } else {
-      if (!has(languageForExtension, extension)) {
-        throw new Error(
-          `Cannot parse module ${specifier} at ${location}, no parser configured for extension ${extension}`,
-        );
-      }
-      language = languageForExtension[extension];
+      language = languageForExtension[extension] || `extension:${extension}`;
```
kriskowal, line 98: I would also accept `|| extension` based on our out-of-band conversation yesterday.

PR #1372, opened 2022-11-16
fix(compartment-mapper): fix reflexive packageLocation in node-modules/translateGraph

kumavis:
`translateGraph` accidentally set packages' own `packageLocation` as the `entryPackageLocation`

further naming improvements could happen here


kriskowal (approved):
This is enough. If you follow-thru, I’d like to divide variables named `packageLocation` into `dependeeLocation` and `dependencyLocation`. If you don’t, please leave me a reminder.

PR #1374, opened 2022-11-16
fix(compartment-mapper): resolving to package dir should include candidate for package main

### LINKED ISSUE #1363: compartment-mapper/importHook - npm package expects '.' to resolve to main
compatibility issue with [`superagent@3.8.3`](https://npmfs.com/package/superagent/3.8.3/)

a module specifies `../..` here https://npmfs.com/package/superagent/3.8.3/lib/node/agent.js#L10
which resolves to the package root, and expects that to resolve to the main https://npmfs.com/package/superagent/3.8.3/package.json#L62

kumavis:
Fixes #1363


packages/compartment-mapper/src/import-hook.js
```diff
@@ -144,6 +146,10 @@ export const makeImportHookMaker = (
       const candidates = [];
       if (moduleSpecifier === '.') {
         candidates.push('./index.js');
+        // https://github.com/endojs/endo/issues/1363
+        if (packageMainExport) {
+          candidates.push(packageMainExport);
+        }
```
kriskowal, line 152: I think that, if there is a `main` in `package.json`, it should be used instead of `index.js`, not as a fallback when `index.js` does not exist. We should have tests, and it would be good for that test to verify parity with Node.js.

kriskowal (changes requested):
I suspect a slight change is needed (noted below) and of course tests.

kumavis:
moved default package resolution to aliases instead of candidates :tada:

packages/compartment-mapper/test/fixtures-0/node_modules/excalibur/src/sword.js
```diff
@@ -0,0 +1,2 @@
+const { sword } = require('..');
+exports.sword = sword;
```
kumavis, line 2: this test was way better when i first wrote it as `export { sword } from (stone)` but alas its only relevant for commonjs behavior

packages/compartment-mapper/src/infer-exports.js
```diff
@@ -157,12 +157,27 @@ export const inferExportsAndAliases = (
...
+
+  // collect externalAliases from exports and main/module
   assign(
     externalAliases,
     fromEntries(inferExportsEntries(descriptor, tags, types)),
   );
+
+  // expose default module as package root
+  // may be overwritten by browser field
+  // see https://github.com/endojs/endo/issues/1363
```
kriskowal, line 170: Optional: I try to consistently use capitalization and punctuation to delimit sentences in comments. I also break sentences between lines so that changes to one sentence do not rewrap subsequent sentences.

kriskowal (approved):

PR #1375, opened 2022-11-17
fix(ses): fix for reexports confusion

kumavis:


packages/ses/test/test-import-gauntlet.js
```diff
@@ -256,3 +256,73 @@ test('export name as default', async t => {
...
+    'https://example.com/xyz.js': `
+      export const answer = 1337;
+    `,
+    'https://example.com/qwe.js': `
+      export default 3791;
+    `,
+    'https://example.com/reexport.js': `
+      export { answer as answer1 } from './abc.js';
+      export { answer as answer2 } from './xyz.js';
+      export { answer as answer3 } from './xyz.js';
```
naugtur, line 276: cover two possible sources of duplication

packages/static-module-record/test/test-static-module-record.js
```diff
@@ -588,7 +588,8 @@ test('import meta present', t => {
   t.is(record.__needsImportMeta__, true);
 });
 
-test('export names', t => {
+// FIXME: this test depends on published SES.
+test.failing('export names', t => {
```
naugtur, line 592: Sadly two tests here include stable SES behavior in the test. I tried replacing SES locally, but didn't succeed.

packages/static-module-record/src/babelPlugin.js
```diff
@@ -536,18 +537,33 @@ function makeModulePlugins(options) {
...
+            if (source) {
+              // If source is defined, it's a reexport from a different module.
+              // In this case we want to collect the local and reexported name
+              // If there's no `as newName`, the ExportNamedDeclaration will have the `exported` field anyway,
+              // representing the same identifier as `local`
+
+              if (!reexportMap[source.value]) {
+                reexportMap[source.value] = [];
+              }
+              reexportMap[source.value].push([local.name, exported.name]);
```
naugtur, line 553: Used the source name as key here instead of exported name, because I need to iterate the reexportMap once to get all exports and I'd need to keep iterating it in module-instance if it was keyed on exports.

naugtur:
Pushed out a fix.

Now compartment-mapper's bundler is failing, because of its intimate relationship with static-module-record's internals.
I think I might have solved half of the problem here: https://gist.github.com/naugtur/c592ee319275333a4957e3db983a1b57  
But TBH I don't get what bundler is doing yet.

packages/static-module-record/test/test-static-module-record.js
kumavis: `compartment-mapper` tests have a thing where they build ses before running.
might be complicated dep topology bc it uses the code from compartment-mapper

kumavis: does just going into ses and running `yarn build` not resolve the issue locally?

packages/compartment-mapper/src/bundle.js
```diff
@@ -271,22 +283,45 @@ ${__syncModuleProgram__},
...
+            return renames.map(
+              ([localName, exportedName]) =>
+                `${q(exportedName)}: { value: cells[${
+                  indexedImports[specifier]
+                }][${q(localName)}] }`,
+            );
+          },
+        )} });
+          `);
+      }
```
naugtur, line 322: I'm considering printing the 5 lines of closing braces as a poster or something.   
Spent non-trivial time matching all of them :sweat_smile:

naugtur:
Fixed the bundler.

packages/static-module-record/test/test-static-module-record.js
naugtur: Nope, I tried that.

naugtur: Ok, some experience from the bundler helped me notice the test is not using notifiers mapping from ses, but instead implementing its own. fixing.

naugtur: aaaand fixed

packages/compartment-mapper/src/bundle.js
kriskowal: Code generation is fun. We need more Lisp macros.

packages/compartment-mapper/src/bundle.js
```diff
@@ -271,22 +283,45 @@ ${__syncModuleProgram__},
...
-    ),
+      );
+      // Create references for export name as newname
+      const namedReexportsToProcess = Object.entries(__reexportMap__);
+      if (namedReexportsToProcess.length > 0) {
+        mappings.push(`
+  Object.defineProperties(cells[${index}], {${namedReexportsToProcess.map(
+          ([specifier, renames]) => {
+            return renames.map(
+              ([localName, exportedName]) =>
```
kriskowal, line 314: Optional nit, I would buy `importedName` to mirror `exportedName`.

packages/ses/test/test-import-gauntlet.js
```diff
@@ -256,3 +256,73 @@ test('export name as default', async t => {
...
+    {
+      resolveHook: resolveNode,
+      importHook: makeImportHook('https://example.com'),
+    },
+  );
+
+  await compartment.import('./main.js');
+});
+
+// throws SyntaxError: This experimental syntax requires enabling the parser plugin: "exportDefaultFrom". (4:13)␊
```
kriskowal, line 301: Windows line ending

packages/ses/test/test-import-gauntlet.js
```diff
@@ -256,3 +256,73 @@ test('export name as default', async t => {
...
+test.failing('reexport with implicit default syntax', async t => {
+  t.plan(2);
+  const makeImportHook = makeNodeImporter({
+    'https://example.com/qwe.js': `
```
kriskowal, line 305: I did enjoy this. 🥚

kriskowal (approved):
Needs entries in `NEWS.md` for each of these three packages. Otherwise, looks like a pure win from here. Thank you for finding and fixing.

packages/compartment-mapper/src/bundle.js
naugtur: I mirrored the names from AST, didn't think twice. I'd gladly call them renamedFrom renamedTo.

PR #1376, opened 2022-11-18
refactor(marshal): Increase internal consistency

gibson042:
* Fix typos
* Work around [microsoft/TypeScript#50286](https://github.com/microsoft/TypeScript/issues/50286) without `@ts-ignore`
* Increase similarity between encodeTo{CapData,Smallcaps}
* DRY out encodeToSmallcaps
* Use more standard patterns in helpers
* Move detailed ErrorHelper checks from canBeValid to assertValid
* Update ErrorHelper to use `assert.Fail`

### DIFF
--- a/packages/marshal/src/encodePassable.js
+++ b/packages/marshal/src/encodePassable.js
@@ -298,10 +298,10 @@ const decodeTagged = (encoded, decodePassable) => {
 
 /**
  * @param {EncodeOptions} encodeOptions
- * @returns {(passable: Passable) => string}
- *
  * encodeOptions is actually optional, but not marked as such to work around
  * https://github.com/microsoft/TypeScript/issues/50286
+ *
+ * @returns {(passable: Passable) => string}
  */
 export const makeEncodePassable = ({
   encodeRemotable = (rem, _) => Fail`remotable unexpected: ${rem}`,
--- a/packages/marshal/src/encodePassable.js
+++ b/packages/marshal/src/encodePassable.js
@@ -297,16 +297,12 @@ const decodeTagged = (encoded, decodePassable) => {
  */
 
 /**
- * @param {EncodeOptions} [encodeOptions]
+ * @param {EncodeOptions} encodeOptions
  * @returns {(passable: Passable) => string}
+ *
+ * encodeOptions is actually optional, but not marked as such to work around
+ * https://github.com/microsoft/TypeScript/issues/50286
  */
-// `yarn lint` complains here but not for equivalent code in agoric-sdk.
-// Also, vscode does not complain. Hence we're using at-ts-ignore rather than
-// at-ts-expect-error. Using at-ts-ignore should also generate a complaint
-// that we should be using at-expect-error, where we would normally need
-// to suppress that error as well. However, perhaps that second error currently
-// happens only in agoric-sdk, but not yet in endo. TODO figure out and fix.
-// @ts-ignore
 export const makeEncodePassable = ({
   encodeRemotable = (rem, _) => Fail`remotable unexpected: ${rem}`,
   encodePromise = (prom, _) => Fail`promise unexpected: ${prom}`,
--- a/packages/marshal/src/encodeToCapData.js
+++ b/packages/marshal/src/encodeToCapData.js
@@ -86,10 +86,10 @@ const dontEncodeErrorToCapData = err => Fail`error object unexpected: ${err}`;
 
 /**
  * @param {EncodeToCapDataOptions} encodeOptions
- * @returns {(passable: Passable) => Encoding}
- *
  * encodeOptions is actually optional, but not marked as such to work around
  * https://github.com/microsoft/TypeScript/issues/50286
+ *
+ * @returns {(passable: Passable) => Encoding}
  */
 export const makeEncodeToCapData = ({
   encodeRemotableToCapData = dontEncodeRemotableToCapData,
--- a/packages/marshal/src/encodeToCapData.js
+++ b/packages/marshal/src/encodeToCapData.js
@@ -87,6 +87,9 @@ const dontEncodeErrorToCapData = err => Fail`error object unexpected: ${err}`;
 /**
  * @param {EncodeToCapDataOptions} encodeOptions
  * @returns {(passable: Passable) => Encoding}
+ *
+ * encodeOptions is actually optional, but not marked as such to work around
+ * https://github.com/microsoft/TypeScript/issues/50286
  */
 export const makeEncodeToCapData = ({
   encodeRemotableToCapData = dontEncodeRemotableToCapData,
--- a/packages/marshal/src/encodeToCapData.js
+++ b/packages/marshal/src/encodeToCapData.js
@@ -134,30 +134,26 @@ export const makeEncodeToCapData = ({
     // JSON, and some must be encoded as [QCLASS] composites.
     const passStyle = passStyleOf(passable);
     switch (passStyle) {
-      case 'null': {
-        return null;
+      case 'null':
+      case 'boolean':
+      case 'string': {
+        // pass through to JSON
+        return passable;
       }
       case 'undefined': {
         return { [QCLASS]: 'undefined' };
       }
-      case 'string':
-      case 'boolean': {
-        return passable;
-      }
       case 'number': {
+        // Special-case numbers with no digit-based representation.
         if (Number.isNaN(passable)) {
           return { [QCLASS]: 'NaN' };
-        }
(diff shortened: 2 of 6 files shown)


packages/marshal/src/helpers/error.js
```diff
@@ -47,18 +47,19 @@ export const ErrorHelper = harden({
...
     const { name } = proto;
     const EC = getErrorConstructor(name);
-    if (!EC || EC.prototype !== proto) {
-      const note = X`Errors must inherit from an error class .prototype ${candidate}`;
-      // Only terminate if check throws
-      reject && reject(note);
-      assert.note(candidate, note);
-    }
+    (EC && EC.prototype === proto) ||
+      Fail`Errors must inherit from an error class .prototype ${candidate}`;
```
gibson042, line 62: Removing these `assert.note` invocations from `ErrorHelper.canBeValid` might be running afoul of [Chesterson's fence](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence)—they will no longer be encountered in the special "accommodate errors that are not valid Passables" `if (ErrorHelper.canBeValid(passable)) { return $encodeError(passable, $encoder) }` tests at the top of encode{Passable,ToCapData,ToSmallCaps} functions. If they _should_ be encountered in those cases, I will move the checks back into `ErrorHelper.canBeValid` along with explanatory comments.

packages/marshal/src/encodeToCapData.js
```diff
@@ -87,6 +87,9 @@ const dontEncodeErrorToCapData = err => Fail`error object unexpected: ${err}`;
 /**
  * @param {EncodeToCapDataOptions} encodeOptions
  * @returns {(passable: Passable) => Encoding}
+ *
+ * encodeOptions is actually optional, but not marked as such to work around
+ * https://github.com/microsoft/TypeScript/issues/50286
```
erights, line 92:
```suggestion
 * @param {EncodeToCapDataOptions} encodeOptions
 * encodeOptions is actually optional, but not marked as such to work around
 * https://github.com/microsoft/TypeScript/issues/50286
 *
 * @returns {(passable: Passable) => Encoding}
```

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -113,12 +113,27 @@ const dontEncodeErrorToSmallcaps = err =>
 /**
  * @param {EncodeToSmallcapsOptions} encodeOptions
  * @returns {(passable: Passable) => SmallcapsEncoding}
+ *
+ * encodeOptions is actually optional, but not marked as such to work around
+ * https://github.com/microsoft/TypeScript/issues/50286
```
erights, line 118:
```suggestion
 * @param {EncodeToSmallcapsOptions} encodeOptions
 * encodeOptions is actually optional, but not marked as such to work around
 * https://github.com/microsoft/TypeScript/issues/50286
 *
 * @returns {(passable: Passable) => SmallcapsEncoding}
```

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -175,37 +189,30 @@ export const makeEncodeToSmallcaps = ({
         // All other strings pass through to JSON
         return passable;
       }
-      case 'symbol': {
-        // At the smallcaps level, we prefix symbol names with `%`.
-        // By "symbol name", we mean according to `nameForPassableSymbol`
-        // which does some further escaping. See comment on that function.
```
erights, line 181: Is this comment worth preserving?

packages/marshal/src/helpers/copyRecord.js
```diff
@@ -52,6 +49,8 @@ export const CopyRecordHelper = harden({
       checkNormalProperty(candidate, name, true, assertChecker);
     }
     // Recursively validate that each member is passable.
-    values(candidate).every(v => !!passStyleOfRecur(v));
+    for (const val of values(candidate)) {
+      passStyleOfRecur(val);
+    }
```
erights, line 54: Weird that in the same file you replace one old `for` loop with an `.every` and you replace one old `.every` with a `for` loop. Should we instead use the same style in both places?

erights (commented):
Still thinking about the handling of non-passable errors, so this review is not yet done. But in the meantime...

packages/marshal/src/encodePassable.js
```diff
@@ -304,23 +304,19 @@ const decodeTagged = (encoded, decodePassable) => {
  */
 
 /**
- * @param {EncodeOptions} [encodeOptions]
+ * @param {EncodeOptions} encodeOptions
  * @returns {(passable: Passable) => string}
+ *
+ * encodeOptions is actually optional, but not marked as such to work around
+ * https://github.com/microsoft/TypeScript/issues/50286
```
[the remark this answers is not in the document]
gibson042: Huh, I didn't know that kind rearrangement was allowed. Suggestion accepted.

As for the "why is this an improvement?" question, the most important thing to me is correcting the current divergence between makeEncodePassable documenting optional `encodeOptions` and using `@ts-ignore` vs. makeEncodeTo{CapData,Smallcaps} documenting it as required. The latter seemed slightly preferable (and all current call sites do provide it), but we could just as easily go the other way.

packages/marshal/src/encodeToSmallcaps.js
gibson042: I don't think so, because it is covered by the package comment and the direct use of `nameForPassableSymbol` in the same way that e.g. 'tagged' doesn't re-explain use of `{ '#tag': …, payload: … }`.

packages/marshal/src/helpers/copyRecord.js
gibson042: Yeah, I meant to call that out myself. The `.every(…)` sites make use of the return value, while this site just needs to call `passStyleOfRecur` on each value in case that throws an exception. We could arguably make that even more clear here:
```suggestion
    for (const val of values(candidate)) {
      void passStyleOfRecur(val);
    }
```

erights:
@gibson042 Sorry this dropped off my radar and I have not yet completed my review. Is this blocked only on awaiting my approval? (Which itself was blocked only on my looking deeper at the serializing of non-passable errors.)

Because it's been sitting for so long, please rebase to current master and let me know when you've done so. At that point I will resume my review.

Btw, beware of collisions between this and https://github.com/endojs/endo/pull/1385 , which I've also asked you to review.

gibson042:
@erights Huh, I lost track of it too. Rebased and confirmed compatible with #1385.

packages/marshal/src/encodePassable.js
```diff
@@ -304,23 +304,19 @@ const decodeTagged = (encoded, decodePassable) => {
...
-// happens only in agoric-sdk, but not yet in endo. TODO figure out and fix.
-// @ts-ignore
 export const makeEncodePassable = ({
   encodeRemotable = (rem, _) => assert.fail(X`remotable unexpected: ${rem}`),
   encodePromise = (prom, _) => assert.fail(X`promise unexpected: ${prom}`),
   encodeError = (err, _) => assert.fail(X`error unexpected: ${err}`),
 } = {}) => {
   const encodePassable = passable => {
-    if (ErrorHelper.canBeValid(passable, x => x)) {
+    if (ErrorHelper.canBeValid(passable)) {
```
erights, line 319: Aha. I knew I saw that somewhere but forgot where. Already fixed by https://github.com/endojs/endo/pull/1385 , which is now merged. Sorry for the miscoordination.

packages/marshal/src/encodeToSmallcaps.js
erights: Thanks. Resolving.

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -112,13 +112,28 @@ const dontEncodeErrorToSmallcaps = err =>
...
+  const assertEncodedError = encoding => {
+    (typeof encoding === 'object' && hasOwnPropertyOf(encoding, '#error')) ||
+      Fail`internal: Error encoding must have "#error" property: ${q(
+        encoding,
+      )}`;
+    // Assert that the #error property decodes to a string.
+    const message = encoding['#error'];
+    (typeof message === 'string' &&
+      (!startsSpecial(message) || message.startsWith('!'))) ||
+      Fail`internal: Error encoding must string message: ${q(message)}`;
```
erights, line 134:
```suggestion
      Fail`internal: Error encoding must have string message: ${q(message)}`;
```

erights (approved):
LGTM

PR #1378, opened 2022-11-18
feat(ses): support RedirectStaticModuleInterface without specifying a record

kumavis:
allow `importHook` to return an alias record that `ses/src/module-load.js` can attempt to import
this allows `importHook` to resolve to something that may be provided by a different part of the import phase, like the `moduleMapHook`

### DIFF
--- a/packages/ses/index.d.ts
+++ b/packages/ses/index.d.ts
@@ -64,9 +64,10 @@ export type FinalStaticModuleType =
   | ThirdPartyStaticModuleInterface;
 
 export interface RedirectStaticModuleInterface {
-  record: FinalStaticModuleType;
   specifier: string;
+  record?: FinalStaticModuleType;
   importMeta?: any;
+  compartment?: Compartment;
 }
 
 export type StaticModuleType =
--- a/packages/ses/src/module-load.js
+++ b/packages/ses/src/module-load.js
@@ -175,27 +175,64 @@ const loadWithoutErrorAnnotation = async (
     )} in compartment ${q(compartment.name)}`;
   }
 
-  if (staticModuleRecord.record !== undefined) {
-    const {
-      compartment: aliasCompartment = compartment,
-      specifier: aliasSpecifier = moduleSpecifier,
-      record: aliasModuleRecord,
-      importMeta,
-    } = staticModuleRecord;
+  // check if record is a RedirectStaticModuleInterface
+  if (staticModuleRecord.specifier !== undefined) {
+    // check if this redirect with an explicit record
+    if (staticModuleRecord.record !== undefined) {
+      // ensure expected record shape
+      if (staticModuleRecord.compartment !== undefined) {
+        throw new TypeError(
+          'Cannot redirect to an explicit record with a specified compartment',
+        );
+      }
+      const {
+        compartment: aliasCompartment = compartment,
+        specifier: aliasSpecifier = moduleSpecifier,
+        record: aliasModuleRecord,
+        importMeta,
+      } = staticModuleRecord;
 
-    const aliasRecord = await loadRecord(
-      compartmentPrivateFields,
-      moduleAliases,
-      aliasCompartment,
-      aliasSpecifier,
-      aliasModuleRecord,
-      pendingJobs,
-      moduleLoads,
-      errors,
-      importMeta,
+      const aliasRecord = await loadRecord(
+        compartmentPrivateFields,
+        moduleAliases,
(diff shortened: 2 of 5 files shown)


packages/ses/index.d.ts
```diff
@@ -69,8 +69,14 @@ export interface RedirectStaticModuleInterface {
   importMeta?: any;
 }
 
+export interface AliasModuleInterface {
+  alias: string;
+  compartment: Compartment;
```
naugtur, line 74: why would the alias come from a different compartment if the goal for this is to support the `browser` field? 
is this the part where browser field defines a different package name for an import? I was thinking of the browser field pointing to a different export of a package. 
Why would an importHook know of a different compartment than its own?

TL;DR: what's the point of having `compartment` here?

packages/ses/index.d.ts
```diff
@@ -69,8 +69,14 @@ export interface RedirectStaticModuleInterface {
...
+}
+
 export type StaticModuleType =
   | RedirectStaticModuleInterface
+  | AliasModuleInterface
```
naugtur, line 79: would it be possible to use the Redirect type with slight modifications? Alias and Redirect seem conceptually close.
That would require a way to tell if it should go through `loadRecord` or `memoizedLoadWithErrorAnnotation` and potentially some simplification of `Redirect...`?   
I want to make sure that we're intentionally making sure the same module instance gets cached under both specifiers. (I'm not sure this is what other engines or bundlers are doing when things get aliased, they probably are)
A Redirect that says "look me up again in moduleMap under this other name before moving forward" is potentially a viable alternative. 

I'm not sure if that's a good idea, just trying to go through the consequences and play the devil's advocate or checking if they're intentional.

kriskowal: I would also like to entertain the possibility that this could have the same shape as a redirect. The implementation is likely the same, but with separate handling for the cases where a redirect appears in the moduleMap/moduleMapHook or the importHook.

That is to say, if we rename `alias` to `specifier`, does this break? Does this require us to distinguish an alias/redirect record differently?

packages/ses/index.d.ts
kriskowal: This feature should be sufficiently general to also handle `imports` in `package.json`.

kriskowal (changes requested):
Requesting further research into aligning the shapes of module descriptors to make usage easier to learn and remember if not standardize.

packages/ses/src/module-load.js
```diff
@@ -175,27 +175,64 @@ const loadWithoutErrorAnnotation = async (
...
+      mapSet(moduleRecords, moduleSpecifier, aliasRecord);
+      return aliasRecord;
+    }
+
+    // check if this redirect with an explicit compartment
+    if (staticModuleRecord.compartment !== undefined) {
+      // ensure expected record shape
+      if (staticModuleRecord.importMeta !== undefined) {
+        throw new TypeError(
+          'Cannot redirect to an inexplicit record with a specified importMeta',
```
kriskowal, line 215:
```suggestion
          'Cannot redirect to an implicit record with a specified importMeta',
```

packages/ses/src/module-load.js
```diff
@@ -175,27 +175,64 @@ const loadWithoutErrorAnnotation = async (
...
+        aliasCompartment,
+        aliasSpecifier,
+        aliasModuleRecord,
+        pendingJobs,
+        moduleLoads,
+        errors,
+        importMeta,
+      );
+      mapSet(moduleRecords, moduleSpecifier, aliasRecord);
+      return aliasRecord;
```
kriskowal, line 207: This is fine. But, reviewing this has given me a set of fresh eyes and there’s an existing defect. Having the memo set after await makes suspicious that there might be a race condition here. 

But, `loadRecord`, on closer inspection, is synchronous! We should get rid of `await` in `await loadRecord` and remove `async` from the `loadRecord` declaration.

There remains a possibility here that concurrent calls to `importHook` that redirect to the same `specifier` with a `record` will race to define that module (the last one winning), and then an `importHook` for the aliased specifier might load it again. Fixing _that_ problem is definitely out of scope for this change. The proper solution to something like that is probably to thread a cancellation token thru `importHook` and ensure the _first_ attempt to define the aliased module wins, aborting and ignoring any that remain. (Way out of scope. I’ll make an issue.)

kriskowal (approved):
Thank you, this is good.

PR #1382, opened 2022-12-01
fix: include inherited methods in missing method error

erights:
Fixes https://github.com/Agoric/agoric-sdk/issues/6570

### DIFF
--- a/packages/eventual-send/src/handled-promise.js
+++ b/packages/eventual-send/src/handled-promise.js
@@ -4,7 +4,7 @@ import {
   localApplyFunction,
   localApplyMethod,
   localGet,
-  sortedOwnKeys,
+  getMethodNames,
 } from './local.js';
 import { makePostponedHandler } from './postponed.js';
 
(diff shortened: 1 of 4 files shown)


erights:
> The test shows the outcome I'm interested in. Thanks!
> 
> I don't know enough to approve the implementation. It looks fine to me, but I'm not aware of the constraints you need to maintain here, so someone more familiar with eventual-send should take a look.

@michaelfig I think that's you. I'll hold off merging until I get your review. Thanks.

erights:
Hi @kriskowal , I need your help to debug https://github.com/endojs/endo/actions/runs/3596352943/jobs/6056920559 . I tried to run it locally. But simply doing `yarn test:pre-release` gave 
`>>>  Failed to load resource: net::ERR_FILE_NOT_FOUND`

So I then tried to puzzle my way through https://github.com/endojs/endo/blob/master/packages/ses-integration-test/README.md but immediately got stuck on the non-existence of `./scripts/build-pre-release-test.sh`.

Trying to puzzle out what to do instead, I found I was out of my depth.

michaelfig (approved):
LGTM!

PR #1383, opened 2022-12-01
feat(compartment-mapper): support bundling cjs

kumavis:


packages/compartment-mapper/src/bundle.js
```diff
@@ -206,60 +223,24 @@ export const makeBundle = async (read, moduleLocation, options) => {
...
   for (const module of modules) {
     module.indexedImports = Object.fromEntries(
       Object.entries(module.resolvedImports).map(([importSpecifier, key]) => [
         importSpecifier,
         modulesByKey[key].index,
       ]),
     );
+    parsersInUse.add(module.parser);
+    // TODO: ask for a better name:
+    module.yummyChunks = processModule(module);
```
naugtur, line 236: Any ideas what a collective noun of these thingamabobs should be?
Most names I introduced here are horrible and need thinking.

packages/compartment-mapper/src/parse-archive-cjs.js
```diff
@@ -47,6 +51,7 @@ export const parseArchiveCjs = async (
         exports: freeze(exports),
         reexports: freeze(reexports),
         execute: noopExecute,
+        __syncModuleProgram__: syncModuleProgram,
```
naugtur, line 54: There's an issue with this change - when it's added here, archive no longer works. 
The presence of `__syncModuleProgram__` causes an error:
`'Property \'__fixedExportMap__\' of a precompiled module record must be an object, got "[undefined]", for module "./index.js"'`

packages/compartment-mapper/src/bundle.js
kumavis: `chonkyChunks`

packages/compartment-mapper/b.js
```diff
@@ -0,0 +1,347 @@
+'use strict';
```
kriskowal, line 1: Was this file intentionally checked-in?

packages/compartment-mapper/src/bundle.js
kriskowal: I’m going to go out on a limb with `module.processed`, and only better than that if we rename `processModule` around the name of the thing it returns instead of what it accepts. Maybe `module.generator` and `makeModuleGenerator(module)`.

packages/compartment-mapper/src/parse-archive-cjs.js
kriskowal: Be sure to remove the `execute` method if you’re changing this from being a third-party-static-module-record to a first-party-static-module-record.

Can you contrive a valid `__fixedExportMap__`, `__liveExportMap__`, and any other properties that a 1st-party module source are supposed to have? Are they reasonable defaults? If absent, can the reasonably be inferred from the other properties?

If we’ve got a lot of yes answers, I’m open to moving default beahvior into `ses/src/module-instance.js`.

packages/compartment-mapper/test/test-bundle.js
```diff
@@ -41,16 +41,25 @@ const expectedLog = [
...
+    fromMjs: 'foo',
+    zzz: 1,
+  },
 ];
 
-test('bundles work', async t => {
+test.only('bundles work', async t => {
   const bundle = await makeBundle(read, fixture);
-  // t.log(bundle);
+  fs.writeFileSync('./b.js', bundle);
```
kriskowal, line 56: Presumably temporary.

naugtur: very much so

packages/compartment-mapper/b.js
naugtur: yes, but temporarily, as the visualization of what progress/changes we're making to an example bundle.

it was very useful for understanding changes without having to run/debug the code.

packages/compartment-mapper/src/parse-archive-cjs.js
naugtur: I went with a better isPrecompiled duck typing option https://github.com/endojs/endo/pull/1383/commits/b1cc4c5bf36d9f04438c1de0df30f126c4d2476f#diff-4209ab82d2d256122a0e77a725980708c388be37712ffcd296561976778daa09R65

How do you like it?

I don't think we can come up with correct `__fixedExportMap__` and `__liveExportMap__` for a cjs module without executing it. If we could, we'd have done it instead of creating a thirdparty module concept in ses, right?

packages/ses/src/module-link.js
```diff
@@ -62,7 +62,10 @@ export const link = (
 };
 
 function isPrecompiled(staticModuleRecord) {
-  return typeof staticModuleRecord.__syncModuleProgram__ === 'string';
+  return (
+    typeof staticModuleRecord.__syncModuleProgram__ === 'string' &&
+    typeof staticModuleRecord.__fixedExportMap__ !== 'undefined'
+  );
```
naugtur, line 68: if it quacks like a duck... let's check if it also floats.

packages/compartment-mapper/src/parse-pre-cjs.js
```diff
@@ -50,6 +52,7 @@ export const parsePreCjs = async (
       reexports,
       exports,
       execute,
+      __syncModuleProgram__: syncModuleProgram,
```
naugtur, line 55: note: Name it `cjsFunctor` instead, so that it doesn't trigger SES to worry about it

packages/compartment-mapper/src/parse-archive-cjs.js
```diff
@@ -48,8 +52,9 @@ export const parseArchiveCjs = async (
         exports: freeze(exports),
         reexports: freeze(reexports),
         execute: noopExecute,
-      })
-    ),
+        cjsFunctor,
+      },
+    )),
```
naugtur, line 57: prettier consistency

packages/compartment-mapper/src/bundle-cjs.js
```diff
@@ -0,0 +1,55 @@
...
+const exportsCellRecord = exportsList =>
+  ''.concat(
+    ...exportsList.map(
+      exportName => `\
+      ${exportName}: cell(${q(exportName)}),
+`,
+    ),
+  );
+
+// It's more comfortable to write when not in a string, but can't get eslint to leave it be
```
kriskowal, line 13: This could be more clear, or omitted.

kriskowal (approved):
The refactoring into `bundle-*.js` is nice.

PR #1384, opened 2022-12-02
refactor: track comment changes from agoric-sdk

erights:
https://github.com/Agoric/agoric-sdk/blob/master/packages/assert/src/types.js says

```js
// Based on
// https://github.com/Agoric/SES-shim/blob/master/packages/ses/src/error/types.js
// Coordinate edits until we refactor to avoid this duplication
```

This PR borrows back some minor improvements from that other types.js , so that diffs will again show only real differences.

See https://github.com/Agoric/agoric-sdk/pull/6613

### DIFF
--- a/packages/ses/src/error/types.js
+++ b/packages/ses/src/error/types.js
@@ -12,7 +12,7 @@
  */
 
 /**
- * @typedef {Object} AssertMakeErrorOptions
+ * @typedef {object} AssertMakeErrorOptions
  * @property {string=} errorName
  */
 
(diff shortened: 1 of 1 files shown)


erights:
@kriskowal , I get the same bizarre CI error https://github.com/endojs/endo/actions/runs/3598853169/jobs/6061980754 here that I got https://github.com/endojs/endo/actions/runs/3596352943/jobs/6056920559 at https://github.com/endojs/endo/pull/1382

kriskowal (approved):

PR #1390, opened 2022-12-06
fix(eventual-send): E.resolve safer against constructor attack

erights:
If a genuine promise has an own `constructor` accessor property, the getter can mount a reentrancy attack on `Promise.resolve'. Call this the "constructor" attack. This is similar to the "then" attack of https://github.com/Agoric/agoric-sdk/issues/9 but more obscure.

Currently `HandledPromise.resolve` and its alias `E.resolve` protect against the "then" attack but not the "constructor" attack. This PR fixes `HandledPromise.resolve` and `E.resolve` to provide partial reentrancy defense against the "constructor" attack. The "constructor" getter can still run synchronously during the call. But the mere presence of the `constructor` own property will cause `HandledPromise.resolve` and `E.resolve` to coerce the malicious promise into a fresh benign one that will not cause a reentrancy attack.

### DIFF
--- a/packages/eventual-send/test/test-thenable.js
+++ b/packages/eventual-send/test/test-thenable.js
@@ -3,76 +3,75 @@ import { test } from './prepare-test-env-ava.js';
 
 import { E, HandledPromise } from './get-hp.js';
 
-test('E.resolve does not coerce simple promise', async t => {
+const verifySimplePromise = async (t, resolve) => {
   const p = new Promise(_ => {});
-  const p2 = E.resolve(p);
+  const p2 = resolve(p);
   t.is(p, p2, 'simple promise should not coerce');
-});
+};
+test(
+  'E.resolve does not coerce simple promise',
+  verifySimplePromise,
+  E.resolve,
+);
+test(
+  'HandledPromise.resolve does not coerce simple promise',
+  verifySimplePromise,
+  HandledPromise.resolve,
+);
 
-test('HandledPromise.resolve does not coerce simple promise', async t => {
+const verifyThenAttack = async (t, resolve) => {
   const p = new Promise(_ => {});
-  const p2 = HandledPromise.resolve(p);
-  t.is(p, p2, 'simple promise should not coerce');
-});
-
-test('E.resolve protects against "then" attack', async t => {
-  const p = new Promise(_ => {});
+  let getHappened = false;
+  let callHappened = false;
   // Work around the override mistake on SES.
-  Object.defineProperty(p, 'then', { value: (res, _rej) => res('done') });
-  let happened = false;
-  const p2 = E.resolve(p).then(ret => (happened = ret));
+  Object.defineProperty(p, 'then', {
+    get() {
+      getHappened = true;
+      return (res, _rej) => {
+        callHappened = true;
+        res('done');
+      };
+    },
+  });
+  const p2 = resolve(p).then(ret => ret);
   t.not(p, p2, 'an own "then" should cause coercion');
-  t.is(happened, false, `p2 is not yet resolved`);
+  t.is(getHappened, false, `getter not called too early`);
+  t.is(callHappened, false, `then not called too early`);
   t.is(await p2, 'done', `p2 is resolved`);
-});
+  t.is(getHappened, true, `getter not called too early`);
+  t.is(callHappened, true, `then not called too early`);
+};
+test('E.resolve protects against "then" attack', verifyThenAttack, E.resolve);
+test(
+  'HandledPromise.resolve protects against "then" attack',
+  verifyThenAttack,
+  HandledPromise.resolve,
+);
 
-test('HandledPromise.resolve protects against "then" attack', async t => {
+const verifyConstructorAttack = async (t, resolve) => {
   const p = new Promise(_ => {});
-  // Work around the override mistake on SES.
-  Object.defineProperty(p, 'then', { value: (res, _rej) => res('done') });
-  let happened = false;
-  const p2 = HandledPromise.resolve(p).then(ret => (happened = ret));
-  t.not(p, p2, 'an own "then" should cause coercion');
-  t.is(happened, false, `p2 is not yet resolved`);
-  t.is(await p2, 'done', `p2 is resolved`);
-});
-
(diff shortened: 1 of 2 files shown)


packages/eventual-send/test/test-thenable.js
```diff
@@ -3,22 +3,76 @@ import { test } from './prepare-test-env-ava.js';
...
 import { E, HandledPromise } from './get-hp.js';
 
-test('E.resolve is always asynchronous', async t => {
+test('E.resolve does not coerce simple promise', async t => {
+  const p = new Promise(_ => {});
+  const p2 = E.resolve(p);
+  t.is(p, p2, 'simple promise should not coerce');
+});
+
+test('HandledPromise.resolve does not coerce simple promise', async t => {
```
gibson042, line 12: Should we be handling these tests [macro style](https://github.com/avajs/ava/blob/main/docs/01-writing-tests.md#reusing-test-logic-through-macros)?
```js
const verifySimplePromise = async (t, resolve) => {
  const p = new Promise(_ => {});
  const p2 = resolve(p);
  t.is(p, p2, 'simple promise should not coerce');
};
test('E.resolve does not coerce simple promise', verifySimplePromise, E.resolve);
test('HandledPromise.resolve does not coerce simple promise', verifySimplePromise,
  HandledPromise.resolve);

const verifyThenAttack = async (t, resolve) => {
  const p = new Promise(_ => {});
  // Work around the override mistake on SES.
  Object.defineProperty(p, 'then', { value: (res, _rej) => res('done') });
  let happened = false;
  const p2 = resolve(p).then(ret => (happened = ret));
  t.not(p, p2, 'an own "then" should cause coercion');
  t.is(happened, false, `p2 is not yet resolved`);
  t.is(await p2, 'done', `p2 is resolved`);
};
test('E.resolve protects against "then" attack', verifyThenAttack, E.resolve);
test('HandledPromise.resolve protects against "then" attack', verifyThenAttack,
  HandledPromise.resolve);

const verifyConstructorAttack = async (t, resolve) => {
  const p = new Promise(_ => {});
  let happened = false;
  // Work around the override mistake on SES.
  Object.defineProperty(p, 'constructor', {
    get() {
      happened = true;
      return Promise;
    },
  });
  const p2 = resolve(p);
  t.not(p, p2, 'an own "constructor" should cause coercion');
  // This first 'true' demonstrates our remaining vulnerability to reentrancy.
  // But the fact that p coerced to a fresh p2 means that p2 cannot
  // cause a reentrancy attack
  t.is(happened, true, `same turn`);
  await null;
  t.is(happened, true, `later turn with p still unresolved`);
};
test('E.resolve protects against "constructor" attack', verifyConstructorAttack, E.resolve);
test('HandledPromise.resolve protects against "constructor" attack', verifyConstructorAttack,
  HandledPromise.resolve);
```

packages/eventual-send/test/test-thenable.js
```diff
@@ -3,22 +3,76 @@ import { test } from './prepare-test-env-ava.js';
...
+    },
+  });
+  const p2 = E.resolve(p);
+  t.not(p, p2, 'an own "constructor" should cause coercion');
+  // This first 'true' demonstrates our remaining vulnerability to reentrancy.
+  // But the fact that p coerced to a fresh p2 means that p2 cannot
+  // cause a reentrancy attack
+  t.is(happened, true, `same turn`);
+  await null;
+  t.is(happened, true, `later turn with p still unresolved`);
```
gibson042, line 57: I don't think I understand this assertion; what is its purpose/significance?

packages/eventual-send/test/test-thenable.js
erights: Done. Started with exactly your code.

@turadg , this code is using macro style and the unpatched ses-ava. Is it surprising that it seems to work?

packages/eventual-send/test/test-thenable.js
erights: Mostly vestigial from a failed experiment. Gone.

packages/eventual-send/test/test-thenable.js
turadg: Not surprising. It would be if you used the `test.macro` helper because ses-ava has a whiltelist of methods and `macro` isn't in it https://github.com/endojs/endo/blob/235931f48757fc8c2ba472a7e7c5f46fca6286fe/packages/ses-ava/src/ses-ava-test.js#L49-L58

packages/eventual-send/test/test-thenable.js
```diff
@@ -3,22 +3,63 @@ import { test } from './prepare-test-env-ava.js';
...
-});
+  Object.defineProperty(p, 'constructor', {
+    get() {
+      happened = true;
+      return Promise;
+    },
+  });
+  const p2 = resolve(p);
+  t.not(p, p2, 'an own "constructor" should cause coercion');
+  // This first 'true' demonstrates our remaining vulnerability to reentrancy.
```
gibson042, line 51:
```suggestion
  // This 'true' demonstrates our remaining vulnerability to reentrancy.
```

gibson042 (approved):
LGTM

mhofman (commented):
Is preventing re-entrancy during resolve wrapping out of scope for this PR?

packages/eventual-send/test/test-thenable.js
erights: Done

packages/eventual-send/test/test-thenable.js
```diff
@@ -3,22 +3,63 @@ import { test } from './prepare-test-env-ava.js';
...
+
+const verifyThenAttack = async (t, resolve) => {
   const p = new Promise(_ => {});
   // Work around the override mistake on SES.
   Object.defineProperty(p, 'then', { value: (res, _rej) => res('done') });
-  let thened = false;
-  const p2 = E.resolve(p).then(ret => (thened = ret));
-  t.is(thened, false, `p2 is not yet resolved`);
+  let happened = false;
+  const p2 = resolve(p).then(ret => (happened = ret));
```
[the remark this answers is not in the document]
erights: Both done.

Issue #1392, opened 2022-12-06
assert members lack JSDoc in IDE


turadg:
# Problem to be solved

members of the `Assert` type don't show their jsdoc on the member properties.

For example, [FailTag has comprehensive documentation](https://github.com/Agoric/agoric-sdk/blob/6e3bc14aa0a77d8bdfad72ca8f0d8893c1f80b5f/packages/assert/src/types.js#L282-L338). But when hovering on the symbol it's all elided:
<img width="604" alt="Screenshot 2022-12-06 at 9 41 44 AM" src="https://user-images.githubusercontent.com/21505/205983558-83482ec5-5ae3-40b0-b7cc-06d32d8670c9.png">

# Design

One option that works is to keep the docstrings in the `d.ts` files too. The downside is having to keep them in sync. To solve that we could build the .js from .ts. If we need .js without .ts, then we'll have to move the `Assert` type out of .d.ts, but I'm not sure we can then express it in JSDoc since it will be nested.

erights:
> that we could build the .js from .ts. 

That would either require us to
   * audit the ts-to-js compiler for preservation of our security properties, and re-audit when the ts compiler changes, or
   * manually review the js output of the compiler every time either the compiler or the ts sources change.

So let's continue looking for solutions elsewhere, like moving the explanations to .d.ts files.

turadg:
> audit the ts-to-js compiler for preservation of our security properties, and re-audit when the ts compiler changes

Whatever is in that audit could be automated, no? Do we have a test suite of security properties that must be maintained by any syntax transformer? If not, that seems valuable to have.

For the actual JS code generation we could use a "strip only" transformer like esbuild. https://esbuild.github.io/content-types/#typescript-caveats

But I like what I think you're proposing: moving the JSDoc to TypeScript files.

erights:
> Whatever is in that audit could be automated, no? Do we have a test suite of security properties that must be maintained by any syntax transformer? If not, that seems valuable to have.

Yes it could. On the one hand, that strikes me as a significant research project that would take much longer than the first audits, but would not need to be redone.

> For the actual JS code generation we could use a "strip only" transformer like esbuild. https://esbuild.github.io/content-types/#typescript-caveats

Yes, I would find a strip-only one acceptable. It would sidestep all my strong reasons for worry. But is the source to that truly TypeScript? Does the strip-only transform really preserve TypeScript's semantics? I haven't looked so I genuinely don't know. But this is interesting.

[...]

PR #1393, opened 2022-12-06
docs: include FailTag jsdoc on Assert type

turadg:
draft of how to solve https://github.com/endojs/endo/issues/1392

Copying the jsdoc as done in this PR makes it appear in the IDE as desired:
Screenshot 2022-12-06 at 9 48 12 AM

### DIFF
--- a/packages/ses/index.d.ts
+++ b/packages/ses/index.d.ts
@@ -195,6 +195,62 @@ export interface Assert {
     template: TemplateStringsArray | string[],
     ...args: any
   ): DetailsToken;
+  /**
+   * The `Fail` tamplate tag supports replacing patterns like
+   * ```js
+   * assert(cond, X`...complaint...`);
+   * ```
+   * or
+   * ```js
+   * cond || assert.fail(X`...complaint...`);
+   * ```
+   * with patterns like
+   * ```js
+   * cond || Fail`...complaint...`;
+   * ```
+   *
+   * However, due to [weakness in current
+   * TypeScript](https://github.com/microsoft/TypeScript/issues/51426), the `||`
+   * patterns are not as powerful as the `assert(...)` call at enabling static
+   * reasoning. Of the `||`, again due to weaknesses in current TypeScript,
+   * the
+   * ```js
+   * cond || Fail`...complaint...`
+   * ```
+   * pattern is not as powerful as the
+   * ```js
+   * cond || assert.fail(X`...complaint...`);
+   * ```
+   * at enabling static resoning. Despite these problems, we do not want to
+   * return to the
+   * ```js
+   * assert(cond, X`...complaint...`)
+   * ```
+   * style because of the substantial overhead in
+   * evaluating the `X` template in the typical `true` case where it is not
+   * needed. And we do not want to return to the
+   * ```js
+   * assert.fail(X`...complaint...`)`
+   * ```
+   * because of the verbosity and loss of readability. Instead, until/unless
+   * https://github.com/microsoft/TypeScript/issues/51426 is fixed, for those
+   * new-style assertions where this loss of static reasoning is a problem,
+   * instead express the assertion as
+   * ```js
+   *   if (!cond) {
+   *     Fail`...complaint...`;
+   *   }
+   * ```
+   * or, if needed,
+   * ```js
+   *   if (!cond) {
+   *     // `throw` is noop since `Fail` throws. But linter confused
+   *     throw Fail`...complaint...`;
+   *   }
+   * ```
+   * This avoid the TypeScript bugs that cause the loss of static reasoning,
+   * but with no loss of efficiency and little loss of readability.
+   */
   Fail(template: TemplateStringsArray | string[], ...args: any): never;
   quote(payload: any, spaces?: string | number): ToStringable;
   makeAssert: MakeAssert;


SMotaal:
@turadg from past experiences with TypeScript I know that there serious drawback to the expansive use of `@typedef` and `@type` which I see here. My take is that the pattern used here will not yield the desired outcomes.

**Questions**

- Is the intent of this PR to ensure that the documentation body is retained for a consumer facing `(await import(…)).assert.Fail` as defined in [packages/ses/src/error/types.js#L277-L333](https://github.com/endojs/endo/blob/abb07944a385c2b7eaf96dbb487ea07a899ccfe5/packages/ses/src/error/types.js#L277-L333)?
- Do we really need all the type coercion within the `makeAssert` function itself when ultimately from the consumer's viewpoint is that `makeAssert` is coerced to have the type `MakeAssert` which returns `Assert`?

<details>

My understanding of how the IDE works through the code relative to [this PR](https://github.com/endojs/endo/blob/abb07944a385c2b7eaf96dbb487ea07a899ccfe5):

1. The `FailTag` JSDoc type defines the tagged literal function signature along with the documentation body in [packages/ses/src/error/types.js#L277-L333](https://github.com/endojs/endo/blob/abb07944a385c2b7eaf96dbb487ea07a899ccfe5/packages/ses/src/error/types.js#L277-L333).
2. The `Fail` wrapped tagged template function is defined within `makeAssert` with `@type {FailTag}` in [packages/ses/src/error/assert.js#L378-L379](https://github.com/endojs/endo/blob/abb07944a385c2b7eaf96dbb487ea07a899ccfe5/packages/ses/src/error/assert.js#L378-L379)
4. The `Assert` JSDoc type declares a `{Fail: FailTag}` in [packages/ses/src/error/types.js#L335-L365](https://github.com/endojs/endo/blob/abb07944a385c2b7eaf96dbb487ea07a899ccfe5/packages/ses/src/error/types.js#L335-L365)
5. The `assert` function is defined within `makeAssert` using `assign(baseAssert, {Fail})` with `@type {Assert}` in [packages/ses/src/error/assert.js#L429-L442](https://github.com/endojs/endo/blob/abb07944a385c2b7eaf96dbb487ea07a899ccfe5/packages/ses/src/error/assert.js#L429-L442)
6. The `MakeAssert` JSDoc type defines the assert maker function with `@returns {Assert}` in [packages/ses/src/error/types.js#L233-L253](https://github.com/endojs/endo/blob/abb07944a385c2b7eaf96dbb487ea07a899ccfe5/packages/ses/src/error/types.js#L233-L253)
7. The `makeAssert` function which returns `freeze(assert)` is defined with `@type {MakeAssert}` in [packages/ses/src/error/assert.js#L360-L444](https://github.com/endojs/endo/blob/abb07944a385c2b7eaf96dbb487ea07a899ccfe5/packages/ses/src/error/assert.js#L360-L444)
8. The `assert` top-level function is defined by calling `makeAssert()` in [packages/ses/src/error/assert.js#L448-L450](https://github.com/endojs/endo/blob/abb07944a385c2b7eaf96dbb487ea07a899ccfe5/packages/ses/src/error/assert.js#L448-L450)
9. The consumer-facing `Fail` tagged template function is obtained using `(await import("./packages/ses/src/error/assert.js")).assert.Fail` in some arbitrary module assumed to be at the root of this PR for convenience.

</details>

---

**Insights**

- I am convinced that the current pattern will not yield the desired outcome.
- I can explore alternatives and let you know when I have a working example and then we can consider how we want to refactor things — just let me know if that is the direction to take and I will get on it!

<details>

To see how broken TypeScript (version 4.9.4 per my VSCode) is for our purposes all I need to do is make the following subtle correction:

```js
/**
 * @typedef {Object} AssertPrototype
 * @property {FailTag} Fail
 * @property {MakeAssert} makeAssert
 * 
 * assert that expr is truthy, with an optional details to describe
 * the assertion. It is a tagged template literal like
 * ```js
 * assert(expr, details`....`);`
 * ```
 *
 * The literal portions of the template are assumed non-sensitive, as
 * are the `typeof` types of the substitution values. These are
 * assembled into the thrown error message. The actual contents of the
 * substitution values are assumed sensitive, to be revealed to
 * the console only. We assume only the virtual platform's owner can read
 * what is written to the console, where the owner is in a privileged
 * position over computation running on that platform.
 *
 * The optional `optDetails` can be a string for backwards compatibility
 * with the nodejs assertion library.
 *
 * @typedef { BaseAssert & AssertPrototype & {
 *   typeof: AssertTypeof,
 *   error: AssertMakeError,
 *   fail: AssertFail,
 *   equal: AssertEqual,
 *   string: AssertString,
 *   note: AssertNote,
 *   details: DetailsTag,
 *   Fail: FailTag,
 *   quote: AssertQuote,
 *   makeAssert: MakeAssert,
 * } } Assert
 */
```

Now you will see the documentation body when you hover only for `assert.makeAssert` and not for `assert.Fail`. If you switch around those two lines in my code you will see the documentation body `assert.Fail` and not `assert.makeAssert. merely because somehow TypeScript is honouring the last `@property`.

So this does not give us a solution and I am merely demonstrating how TypeScript really lacks the necessary functionality for this pattern.

</details>

---

**Aside**

I also have some concerns regarding casing and naming conventions which are outside the scope of this PR.

<details>

My observations regarding conventions:

- Using `title-casing` for `callable types`: `MakeAssert`, `Assert`, `FailTag` … etc.
- Using `camel-casing` for `callable members`: `assert.makeAssert` … etc.
- Using `title-casing` or `camel-casing` for `tagged template members`: `assert.Fail`, `assert.details` … etc.


</details>

turadg:
> Is the intent of this PR to ensure that the documentation body is retained for a consumer facing (await import(…)).assert.Fail as defined in [packages/ses/src/error/types.js#L277-L333](https://github.com/endojs/endo/blob/abb07944a385c2b7eaf96dbb487ea07a899ccfe5/packages/ses/src/error/types.js#L277-L333)?

Yes, the intent is to solve https://github.com/endojs/endo/issues/1392.  Any instance of `Fail` should have a hoverable doc about what it does. 

> Do we really need all the type coercion within the makeAssert function itself when ultimately from the consumer's viewpoint is that makeAssert is coerced to have the type MakeAssert which returns Assert?

No. Of course that's out of scope of the problem at hand.

> drawback to the expansive use of @typedef and @type which I see here

I assume you mean by "here" you mean the repo because this PR has none. I expect the types will work better by making less use of @typedef, but that's not a requirement. 

> I can explore alternatives and let you know when I have a working example and then we can consider how we want to refactor things — just let me know if that is the direction to take and I will get on it!

@SMotaal you're more than welcome to submit another PR to solve #1392. This is a draft mostly to demonstrate an approach. I haven't had time to complete it and will probably hold off until I see what you propose.

If you want to tackle bigger changes, you're also welcome to propose some, though I can't promise they'll be supported.

SMotaal:
> This is a draft mostly to demonstrate an approach. I haven't had time to complete it and will probably hold off until I see what you propose.

I plan to open PR next week, not sure how that fits with project timelines.

> If you want to tackle bigger changes, you're also welcome to propose some, though I can't promise they'll be supported.

Yes, I think I will open up a separate issue regarding conventions depending on what I dig up when looking closer at the history.

> I assume you mean by "here" you mean the repo because this PR has none.

Here at this point when I checkout the PR as I was commenting 😊

> I expect the types will work better by making less use of @typedef, but that's not a requirement.

Yes that is one aspect, but also coerced `@type`, unless they are intentional to make people more cognizant as a requirement.

I guess it would be helpful if you know if that is the case?

Thanks!

turadg:
> if you know if that is the case

Most of the types here were written for earlier versions of TypeScript and by people very experienced with JavaScript but new to TypeScript. I think originally there was a desire to keep the type annotations in separate files, but that's no longer a requirement. In my opinion, there's no need for `@callback` to type functions that have one implementation and those can instead be on the function definition.

Issue #1395, opened 2022-12-06
Difference in prettier behavior between endo and agoric-sdk.


erights:
At https://github.com/endojs/endo/blob/5e111cd0a4dd869fe63e7eaed966f9b4be1e838e/packages/marshal/src/encodePassable.js#L37-41 is the code

```js
  harden(
    /** @type {string[]} */ (ownKeys(record))
      .sort()
      .reverse(),
  );
```
which is left as is by a `yarn lint-fix` in endo's packages/marshal.

At https://github.com/Agoric/agoric-sdk/blob/db5b218488f2b13275cd1bfee639cd9222656d04/packages/store/src/patterns/encodePassable.js#L33-37 is the exact same code, provoking the lint error at https://github.com/Agoric/agoric-sdk/actions/runs/3633839349/jobs/6131385788 . When I run `yarn lint-fix` in agoric-sdk's packages/store, it rewrites it to

```js
  harden(/** @type {string[]} */ (ownKeys(record)).sort().reverse());
```

I'm ok with either form. But I'm disturbed by the different in `lint-fix` behavior. Anyone what causes it, or what we need to do to realign these?

PR #1396, opened 2022-12-07
update Prettier to 2.8

turadg:
Fixes: #1395 

This leaves Prettier config up to each package, and repeats it in all of them. (Agoric SDK treats Prettier as a repo-level concern and has one place to configure the settings and one script for formatting over the repo, but I figured that workflow change is out of scope for solving 1395.)


erights (approved):
Aside from the comma-spacing one, were all the *.js changes just automated application of the new prettier? Assuming so, I only gave these the briefest of visual sanity checks. Please let me know if there's something I should pay more attention to.

Given this assumption, LGTM, thanks!

erights:
I noticed you enabled auto-merge and are passing all requirement but for a trivial lint-fix, so I'm about to push a fix to that now.

PR #1397, opened 2022-12-07
fix(ses): Remove superfluous tick in module loader

### LINKED ISSUE #1394: possible race condition in compartment-mapper
@kriskowal in this thread https://github.com/endojs/endo/pull/1378#discussion_r1040261896 says:

> But, reviewing this has given me a set of fresh eyes and there’s an existing defect. Having the memo set after await makes suspicious that there might be a race condition here.
> 
> But, loadRecord, on closer inspection, is synchronous! We should get rid of await in await loadRecord and remove async from the loadRecord declaration.
> 
> There remains a possibility here that concurrent calls to importHook that redirect to the same specifier with a record will race to define that module (the last one winning), and then an importHook for the aliased specifier might load it again. Fixing that problem is definitely out of scope for this change. The proper solution to something like that is probably to thread a cancellation token thru importHook and ensure the first attempt to define the aliased module wins, aborting and ignoring any that remain. 
>

kriskowal:
This removes a superfluous and possibly dangerous await in the module loader. Delimiting the event could cause a race to write to the module memo.

Fixes #1394

### DIFF
--- a/packages/ses/src/module-load.js
+++ b/packages/ses/src/module-load.js
@@ -54,7 +54,7 @@ const resolveAll = (imports, resolveHook, fullReferrerSpecifier) => {
   return freeze(resolvedImports);
 };
 
-const loadRecord = async (
+const loadRecord = (
   compartmentPrivateFields,
   moduleAliases,
   compartment,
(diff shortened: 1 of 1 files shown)


kumavis (approved):

kumavis:
i assume it was marked async due to an earlier version of this section that kicks of async jobs
https://github.com/endojs/endo/blob/181106502d85fdad4d5eb000b605ccaae5dc7d54/packages/ses/src/module-load.js#L91-L105

PR #1399, opened 2022-12-07
feat(cli): Add endo ping and log subcommands

kriskowal:
While fiddling with endo, these subcommands were useful. `ping` performs a trivial healthcheck, confirming that the daemon is running and connected to its named pipe or UNIX domain socket. `log` shells out to `tail` or `tail -f` for the user agent’s log file.


michaelfig (approved):
LGTM!

PR #1400, opened 2022-12-07
fix(compartment-mapper): Thread dev option thru bundler

kriskowal:
It seems I neglected to thread a flag. This allows `devDependencies` to be incorporated into a bundle.


naugtur:
seems fine

naugtur (approved):

PR #1401, opened 2022-12-07
fix: Use wildcard for workspace packages

kriskowal:
In order to reduce friction when adding or removing packages.


turadg (approved):
:happy-dance:

Issue #1248, opened 2022-08-09
marshal method names are misleading


gibson042:
This ship has probably already sailed, but the names of `serialize` and `unserialize` are misleading—they don't actually produce/consume a serialized representation, but rather a CapData structure that in practice is often serialized/deserialized (for real) by means of JSON.{stringify,parse}. They might fairly be called `marshal` and `unmarshal` (cf. https://en.wikipedia.org/wiki/Marshalling_(computer_science) , especially its [first reference](https://stackoverflow.com/a/770509/7864943)), but something like `toCapData` and `fromCapData` would be even more clear.

turadg:
I've been mislead in this way. =)  I think solving this is important for Mainnet 2.

turadg:
I also find myself wanting to express the shape of what's coming out of the CapData. We could solve both ergonomics problems at once with a wrapper that uses @erights 's pattern validation work. E.g.
```
const capData = JSON.parse(capDataStr);
const shapedData = capPresence(marshaller).from(capData, specShape);
```

erights:
> ```js
> const capData = JSON.parse(capDataStr);
> const shapedData = capPresence(marshaller).from(capData, specShape);
> ```

I don't understand. What is this code trying to say? What is it an example of?

gibson042:
`JSON.parse(capDataStr)` is **deserialization** (it converts an input sequence of UTF-16 code units to an output tree-shaped data structure, for CapData in particular one that is expected to be an object with a string-valued "body" property containing valid JSON text and an array-valued "slots" property), and `from(capData, specShape)` could fairly be called **unmarshalling** or **decoding** (it converts an input {body: string, slots: any[]} CapData structure to a data structure that potentially includes promises, remotables, and/or objects with multiple inbound references).

turadg:
>  What is it an example of?

Consider https://github.com/Agoric/agoric-sdk/blob/d042efce684176b68c4d8055564250c7fabf304c/packages/smart-wallet/src/smartWallet.js#L460-L463

That doesn't validate the shape before casting it. It could with a helper that does the unmarshallng (à la `fromCapData` née `unserialize`) that also checks the shape before returning it. And then it wouldn't have to be cast.

In the comment above, `capPresence` is that utility that takes a remote marshaller to do it.

erights:
[...]

PR #1402, opened 2022-12-08
clearer marshal method names

turadg:
Closes: #1248

### DIFF
--- a/packages/marshal/src/marshal.js
+++ b/packages/marshal/src/marshal.js
@@ -21,8 +21,8 @@ import {
 /** @typedef {import('./types.js').MakeMarshalOptions} MakeMarshalOptions */
 /** @template Slot @typedef {import('./types.js').ConvertSlotToVal<Slot>} ConvertSlotToVal */
 /** @template Slot @typedef {import('./types.js').ConvertValToSlot<Slot>} ConvertValToSlot */
-/** @template Slot @typedef {import('./types.js').Serialize<Slot>} Serialize */
-/** @template Slot @typedef {import('./types.js').Unserialize<Slot>} Unserialize */
+/** @template Slot @typedef {import('./types.js').ToCapData<Slot>} ToCapData */
+/** @template Slot @typedef {import('./types.js').FromCapData<Slot>} FromCapData */
 /** @typedef {import('./types.js').Passable} Passable */
 /** @typedef {import('./types.js').InterfaceSpec} InterfaceSpec */
 /** @typedef {import('./types.js').Encoding} Encoding */
@@ -71,7 +71,7 @@ export const makeMarshal = (
   };
 
   /**
-   * @type {Serialize<Slot>}
+   * @type {ToCapData<Slot>}
    */
   const serialize = root => {
     const slots = [];
@@ -340,7 +340,7 @@ export const makeMarshal = (
   };
 
   /**
-   * @type {Unserialize<Slot>}
+   * @type {FromCapData<Slot>}
    */
   const unserialize = data => {
     const { body, slots } = data;
--- a/packages/marshal/src/marshal.js
+++ b/packages/marshal/src/marshal.js
@@ -73,7 +73,7 @@ export const makeMarshal = (
   /**
    * @type {ToCapData<Slot>}
    */
-  const serialize = root => {
+  const toCapData = root => {
     const slots = [];
     // maps val (promise or remotable) to index of slots[]
     const slotMap = new Map();
@@ -342,7 +342,7 @@ export const makeMarshal = (
   /**
    * @type {FromCapData<Slot>}
    */
-  const unserialize = data => {
+  const fromCapData = data => {
     const { body, slots } = data;
     typeof body === 'string' ||
       Fail`unserialize() given non-capdata (.body is ${body}, not string)`;
@@ -367,7 +367,13 @@ export const makeMarshal = (
   };
 
   return harden({
-    serialize,
-    unserialize,
+    toCapData,
+    fromCapData,
+
+    // for backwards compatibility
+    /** @deprecated use toCapData */
+    serialize: toCapData,
+    /** @deprecated use fromCapData */
+    unserialize: fromCapData,
   });
 };
--- a/packages/marshal/src/types.js
+++ b/packages/marshal/src/types.js
@@ -99,8 +99,10 @@ export {};
 /**
  * @template Slot
  * @typedef {Object} Marshal
- * @property {ToCapData<Slot>} serialize
- * @property {FromCapData<Slot>} unserialize
+ * @property {ToCapData<Slot>} serialize use toCapData
+ * @property {FromCapData<Slot>} unserialize use fromCapData
+ * @property {ToCapData<Slot>} toCapData
+ * @property {FromCapData<Slot>} fromCapData
  */
 
 /**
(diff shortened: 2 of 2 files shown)


packages/marshal/src/marshal.js
```diff
@@ -69,9 +69,9 @@ export const makeMarshal = (
   };
 
   /**
-   * @type {Serialize<Slot>}
+   * @type {ToCapData<Slot>}
    */
-  const serialize = root => {
+  const toCapData = root => {
```
gibson042, line 74: This raises the interesting question of to what extent we consider "CapData" to describe just the `{ body: string, slots: any[] }` structure vs. that and/or a body string format that contains JSON text in which special values including those that cannot be directly represented in JSON are replaced with `{ [Symbol.for("@qclass")]: string, ... }` objects (as opposed to the "Smallcaps" body string format that inserts a "#" prefix before JSON text in which special values are represented as strings). Put another way, is `{ body: "{}", slots: [] }` something like CapData–CapData, or something like CapData–default/CapData–qclass?

erights: In choosing format names contrasting "capdata" vs "smallcaps", I feel like we've used up the term "capdata" for that narrower distinction. OTOH, we're still using the type name "CapData" for the record covering both these formats, and possibly future similar formats, when paired with a slots array. While I realize I caused this confusion, I think we need to repair it, and decide on one of these two meanings for the term "CapData".

I mildly favor resolving this by using "capdata" for the specific format and inventing another name for the record type. But I don't have a concrete suggestion. Also, we should expect the capdata format to be deprecated and eventually disappear, whereas the record type will continue, which perhaps argues for renaming the format rather than the type.

turadg: > perhaps argues for renaming the format rather than the type.

+1 to this. If the format is to be deprecated in favor of "smallcaps" then we could name the format something that contrasts it with smallcaps. "oldcaps", "caps1", etc. "largecaps"! ;-)

`CapData` makes more sense to me as the generic record for whatever capability data, in whatever format it is encoded.

dtribble: API-wise, don't you want an API where the smallcap marshaller and the largecap marshaller both use the same API to produce their output?  In that case, this should be the more generic method name. I can see `toCapData` being generic enough.

erights: Now that @FUDCo 's smallcaps PR is merged, it may be easier to simply delete support for the deprecated format, quickly even if not immediately, rather than go through the pain of renaming it. It would be especially nice to delete it before bulldozer.

erights: But in any case, yeah I am convinced: Keep CapData as the name of the type and eat the pain on the format-name side.

packages/marshal/src/marshal.js
```diff
@@ -359,7 +359,13 @@ export const makeMarshal = (
...
-    serialize,
-    unserialize,
+    toCapData,
+    fromCapData,
+
+    // for backwards compatibility
+    /** @deprecated use toCapData */
+    serialize: toCapData,
+    /** @deprecated use fromCapData */
+    unserialize: fromCapData,
```
erights, line 369: Do these `@deprecated`s propagate through the IDE?

erights (approved):
LGTM, thanks!

packages/marshal/src/marshal.js
turadg: a `bodyFormat` property is compelling. Would you still call that structure `CapData`?

gibson042: If we consider `{ body: string, slots: any[] }` to be "CapData", then I think we would keep that name for `{ bodyFormat: string, body: any, slots: any[] }` (with a modifier if distinguishing old vs. new is important, e.g. "CapData2" or "CapData2023" or "flexible CapData" or "self-describing CapData").

erights: `bodyFormat: string` in every one of these records? Compactness matters here, which is why we did smallcaps in the first place, and why I chose to identify the difference in format with a single `'#'` character.

gibson042: There are several places where even the smallcaps encoding could be more compact (some of which have been explicitly considered and rejected, e.g. [non-decimal bigints](https://github.com/Agoric/agoric-sdk/pull/4469#issuecomment-1043602399)), but note that even in that dimension the fixed overhead of `"bodyFormat":"sigil",` would pay for itself in eliminated backslashes against any encoding that recursively includes at least 10 object properties or _x_ properties and _y_ string values where _x_ + _y_ ≥ 10:
```diff
diff current vs. proposed serialization of { string: "foo", promise, set: makeCopySet(["a", "b"]) }
-{"body":"{\"string\":\"foo\",\"promise\":\"&0\",\"set\":{\"#tag\":\"copySet\",\"payload\":[\"b\",\"a\"]}}","slots":["o+10"]}
+{"bodyFormat":"sigil","body":{"string":"foo","promise":"&0","set":{"#tag":"copySet","payload":["b","a"]}},"slots":["o+10"]}
```
(and even fewer than that with a smaller but more opaque alternative—`"f":1,` would break even at just two strings, and automatic inference when `typeof body !== 'string'` starts off two characters _smaller_ [at the cost of additional complexity to encode an unwrapped string which would otherwise be ambiguous with the current double serialization]).

mhofman: I have to admit, since we already separated the transformation from a passable into a serialization friendly structure, and the actual string serialization of that structure, it would make sense having a way to carry this structure around without forcing its serialization, allowing for more natural nesting.

However I'd like to point out that a smallcap structure nested inside another smallcap structure will still require nested marshalling, as any string key or value starting with a reserved char (ascii 33-45) will need to be `!` prefixed, so the nested gains are not as straightforward to calculate (but worst case scenario we'd have 1 extra char per string where with nested serialization we have 2).

We might also be able to avoid JSON serialization altogether as the final decision to serialize would be left to the transport layer. For example we could decide to adopt [tagged netstring](https://tnetstrings.info/) for the XS worker <-> kernel communication channel.

erights: The double encoding is indeed awful. We should figure out how to avoid the double encoding. But still, please let's not add a `bodyFormat` property.

mhofman: An alternative would be to use a different property name than `body` to hold structured / non serialized data. The property name would be representative of the encoding it contains.

For example `{ smallCapsBody: { ... }, slots: [] }`

turadg: Anyone have reservations with landing this PR as is? https://github.com/Agoric/agoric-sdk/pull/6646 is sort of waiting on it.

gibson042 (approved):
This PR is implicitly defining **CapData** as a record structure that currently encompasses a `body` string and a `slots` array, in which the encoding of `body` is implicitly either "_Smallcaps_"[^1] if the first character is "#" or [_unnamed original encoding_][^2] otherwise. That works for me, and we can propagate those definitions in followup changes.

Thanks @turadg!

[^1]: [JSON text](https://www.rfc-editor.org/rfc/rfc8259#section-2) in which arbitrary [pass-by-copy](https://docs.agoric.com/guides/js-programming/far.html#pass-styles-and-harden) source data has been made representable mostly by translation of JSON-incompatible values into "#…" strings
[^2]: [JSON text](https://www.rfc-editor.org/rfc/rfc8259#section-2) in which arbitrary [pass-by-copy](https://docs.agoric.com/guides/js-programming/far.html#pass-styles-and-harden) source data has been made representable mostly by translation of JSON-incompatible values into objects with `@qclass` members

PR #1404, opened 2022-12-13
feat(compartment-mapper): bundle should support tags

kumavis:


packages/compartment-mapper/src/bundle.js
```diff
@@ -140,20 +140,18 @@ const sortedModules = (
  * @param {Object} [options]
  * @param {ModuleTransforms} [options.moduleTransforms]
  * @param {boolean} [options.dev]
+ * @param {Set<string>} [options.tags]
  * @returns {Promise<string>}
  */
 export const makeBundle = async (read, moduleLocation, options) => {
-  const { moduleTransforms, dev } = options || {};
+  const { moduleTransforms, dev, tags = new Set() } = options || {};
```
kriskowal, line 147: I would prefer to default `tags: tagsOption = undefined` and later `const tags = new Set(tagsOption)` so arrays and iterables get lifted, if provided. The TypeScript will complain, but not all is TypeScript.

kriskowal (approved):

PR #1405, opened 2022-12-13
feat(compartment-mapper): allow alternate candidateSuffixes

kumavis:
allows the specification of alternate suffixes when determining candidates

related to supporting features like typescript where files may import other typescript modules with the extension omitted or browserify's `opts.extensions` https://github.com/browserify/browserify#browserifyfiles--opts

### DIFF
--- a/packages/compartment-mapper/src/import-hook.js
+++ b/packages/compartment-mapper/src/import-hook.js
@@ -45,6 +45,9 @@ function getImportsFromRecord(record) {
   return (has(record, 'record') ? record.record.imports : record.imports) || [];
 }
 
+// Node.js default resolution allows for an incomplement specifier that does not include a suffix.
+const nodejsConventionSearchSuffixes = ['.js', '/index.js'];
+
 /**
  * @param {ReadFn|ReadPowers} readPowers
  * @param {string} baseLocation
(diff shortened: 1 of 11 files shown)


packages/compartment-mapper/src/import-hook.js
```diff
@@ -45,13 +45,27 @@ function getImportsFromRecord(record) {
   return (has(record, 'record') ? record.record.imports : record.imports) || [];
 }
 
+// Node.js resolution allows for an incomplement specifier.
```
kriskowal, line 48: “incomplete”? Can you add an example of one?

packages/compartment-mapper/src/import-hook.js
```diff
@@ -45,13 +45,27 @@ function getImportsFromRecord(record) {
...
+];
+
 /**
  * @param {ReadFn|ReadPowers} readPowers
  * @param {string} baseLocation
  * @param {Sources} sources
  * @param {Record<string, CompartmentDescriptor>} compartmentDescriptors
  * @param {Record<string, any>} exitModules
  * @param {HashFn=} computeSha512
+ * @param {Array<string>} candidateSuffixes
```
kriskowal, line 68: “Pass [] to emulate Node.js’s strict behavior. The default handles Node.js’s CommonJS behavior. Unlike Node.js, the Compartment Mapper lifts CommonJS up, more like a bundler, and does not attempt to vary the behavior of resolution depending on the language of the importing module.”

packages/compartment-mapper/test/scaffold.js
```diff
@@ -58,6 +58,7 @@ export function scaffold(
     shouldFailBeforeArchiveOperations = false,
     knownFailure = false,
     tags = undefined,
+    candidateSuffixes = undefined,
```
kriskowal, line 61: Perhaps for the public API, using the name `extensions` or just `suffixes` would be more discoverable. Or `searchSuffixes`?

kriskowal (approved):

packages/compartment-mapper/src/import-hook.js
kumavis: as in "lacking a suffix", like `./util`. maybe ill just write that

packages/compartment-mapper/test/scaffold.js
kumavis: i think its fine to use `searchSuffixes` throughout

PR #1406, opened 2022-12-14
feat(compartment-mapper): common dependencies

kumavis:
provides a means of exposing additional dependencies to every package

this is common when a transpiler adds additional dependencies (eg `@babel/runtime`) or when relying on polyfills of node builtins (eg browserify's replacement of `buffer`, etc)

this is somewhat related to `exitModules`, but they can otherwise be treated as normal packages that can be loaded from the file system

see related:
- browserify: `{ builtins: { [name]: location } }`
  see `opts.builtins` https://github.com/browserify/browserify#browserifyfiles--opts
- webpack: `{ resolve: { fallback: { [name]: location } } }`
  see https://webpack.js.org/configuration/resolve/#resolvefallback

### DIFF
--- a/packages/compartment-mapper/src/node-modules.js
+++ b/packages/compartment-mapper/src/node-modules.js
@@ -35,6 +35,10 @@
  * modules.
  */
 
+/**
+ * @typedef {Record<string, {spec: string, alias: string}>} CommonDependencyDescriptors
+ */
+
 import { inferExportsAndAliases } from './infer-exports.js';
 import { searchDescriptor } from './search.js';
 import { parseLocatedJson } from './json.js';
@@ -228,6 +232,7 @@ const inferParsers = (descriptor, location) => {
  * @param {Object} packageDetails.packageDescriptor
  * @param {Set<string>} tags
  * @param {boolean} dev
+ * @param {CommonDependencyDescriptors} commonDependencyDescriptors
  * @returns {Promise<undefined>}
  */
 const graphPackage = async (
@@ -238,6 +243,7 @@ const graphPackage = async (
   { packageLocation, packageDescriptor },
   tags,
   dev,
+  commonDependencyDescriptors,
 ) => {
   if (graph[packageLocation] !== undefined) {
     // Returning the promise here would create a causal cycle and stall recursion.
@@ -270,6 +276,10 @@ const graphPackage = async (
     devDependencies = {},
   } = packageDescriptor;
   const allDependencies = {};
+  assign(allDependencies, commonDependencyDescriptors);
+  for (const [name, { spec }] of Object.entries(commonDependencyDescriptors)) {
+    allDependencies[name] = spec;
+  }
   assign(allDependencies, dependencies);
   assign(allDependencies, peerDependencies);
   for (const [name, { optional }] of Object.entries(peerDependenciesMeta)) {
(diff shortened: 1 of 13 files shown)


packages/compartment-mapper/src/node-modules.js
```diff
@@ -383,6 +401,7 @@ const graphPackage = async (
  * @param {string} name - name of the package of interest.
  * @param {Set<string>} tags
  * @param {boolean} optional - whether the dependency is optional
+ * @param {Object} [commonDependencyDescriptors] - dependencies to be added to all packages
```
kriskowal, line 404: Let’s narrow this type. I believe this is `Record<string, {specifier: string, alias: string}>`. `Object` and `object` are aliases for `any` in JSDoc.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -436,6 +457,7 @@ const gatherDependency = async (
  * package.json, which was already read when searching for the package.json.
  * @param {boolean} dev - whether to use devDependencies from this package (and
  * only this package).
+ * @param {Object} [commonDependencies] - dependencies to be added to all packages
```
kriskowal, line 460: Also narrow here. `types.js` has common types to DRY up.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -469,6 +492,17 @@ const graphPackages = async (
       `Cannot find package.json for application at ${packageLocation}`,
     );
   }
+
+  // Resolve common dependencies.
+  const commonDependencyDescriptors = {};
+  for (const [alias, dependencyName] of Object.entries(commonDependencies)) {
+    const spec = packageDescriptor.dependencies[dependencyName];
```
kriskowal, line 499: This may be sensitive in packages that don’t describe their `dependencies`, unless we’ve normalized the package descriptor above. Worth verifying.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -469,6 +492,17 @@ const graphPackages = async (
...
+    commonDependencyDescriptors[dependencyName] = {
+      spec,
```
kriskowal, line 501: Would prefer to spell out `specifier` in keeping with the prevailing æsthetic.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -444,6 +466,7 @@ const graphPackages = async (
   tags,
   mainPackageDescriptor,
   dev,
+  commonDependencies = {},
```
kriskowal, line 469: We can default this higher up the call tree so the dependency isn’t optional, and thereby reduce throw-away allocations.

kriskowal (approved):
Please narrow the new types. Otherwise, merge at will.

packages/compartment-mapper/src/node-modules.js
kumavis: this is not the specifier but the install version spec[ification] eg: `^1.0.0`

packages/compartment-mapper/src/node-modules.js
kumavis: this is the default in the lowest level before recursion, so it should only be called once

kumavis:
at time of writing, `commonDependencies` must be listed in `dependencies`, other dependency category are not checked

PR #1409, opened 2022-12-15
fix(where): Fix XDG conventions

kriskowal:
1. There’s a typo for `XDG_CACHE_DIR` that should be `XDG_CACHE_HOME`
2. `XDG` prefixes should be respected everywhere, including Darwin and Windows, except (probably) for the named pipe on Windows.
3. State and Cache should be clearly delineated, except on Windows where there does not appear to be an (obvious) separation within its native conventions.

Based on feedback out of band, I intend to use whereEndoState to locate anything intended to persist between daemon restarts, including logs. I no longer have anything in mind for whereEndoCache. whereEndoSock is using the runtime directory for PID files and Unix Domain Sockets.

### DIFF
--- a/packages/where/index.js
+++ b/packages/where/index.js
@@ -1,33 +1,41 @@
 // @ts-check
 
-/* Infers the rendezvous path for the endo.sock file and apps from the platform
- * and environment.
- */
-
 const { raw } = String;
 
 /**
+ * Returns a path for local Endo application user data on Windows.
+ *
  * @param {{[name: string]: string}} env
  */
 const whereEndoHomeWindows = env => {
   // Favoring local app data over roaming app data since I don't expect to be
   // able to listen on one host and connect on another.
+  // TODO support roaming data for shared content addressable state and
+  // find a suitable mechanism for merging state that may change independently
+  // on separate roaming hosts.
   if (env.LOCALAPPDATA !== undefined) {
     return `${env.LOCALAPPDATA}\\Endo`;
   }
   if (env.APPDATA !== undefined) {
-    return `${env.APPDATA}\\Endo`;
+    return `${env.APPDATA}\\Local\\Endo`;
   }
   if (env.USERPROFILE !== undefined) {
-    return `${env.USERPROFILE}\\AppData\\Endo`;
+    return `${env.USERPROFILE}\\AppData\\Local\\Endo`;
   }
   if (env.HOMEDRIVE !== undefined && env.HOMEPATH !== undefined) {
-    return `${env.HOMEDRIVE}${env.HOMEPATH}\\AppData\\Endo`;
+    return `${env.HOMEDRIVE}${env.HOMEPATH}\\AppData\\Local\\Endo`;
   }
   return 'Endo';
 };
 
 /**
+ * Returns the most suitable path for Endo state with this platform and
+ * environment.
+ * Endo uses the state directory for saved files including applications,
+ * durable capabilities, and the user's pet names for them.
+ * Endo also logs here, per XDG's preference to persist logs even when caches
+ * are purged.
+ *
  * @type {typeof import('./types.js').whereEndoState}
  */
 export const whereEndoState = (platform, env) => {
@@ -46,6 +54,37 @@ export const whereEndoState = (platform, env) => {
 };
 
 /**
+ * Returns the most suitable location for storing state that ideally does not
+ * persist between restarts or reboots, specifically PID files.
+ * This should be coincident with the directory containing UNIX domain sockets,
+ * but is not suitable for Windows named pipes.
+ *
+ * @type {typeof import('./types.js').whereEndoEphemeralState}
+ */
+export const whereEndoEphemeralState = (platform, env) => {
+  if (env.XDG_RUNTIME_DIR !== undefined) {
+    return `${env.XDG_RUNTIME_DIR}/endo`;
+  } else if (platform === 'win32') {
+    return whereEndoHomeWindows(env);
+  } else if (platform === 'darwin') {
+    if (env.HOME !== undefined) {
+      return `${env.HOME}/Library/Application Support/Endo`;
+    }
+  }
+  if (env.USER !== undefined) {
+    // The XDG specification says that in this case, we should fall through to
+    // system specific behavior, but we are not in a position to assume which
+    // system we're using, other than it is probably a variety of Linux,
+    // in which case /tmp and /run might be viable base paths.
+    return `/tmp/endo-${env.USER}`;
+  }
+  return '/tmp/endo';
+};
+
+/**
+ * Returns the most suitable path for the Endo UNIX domain socket or Windows
+ * named pipe.
+ *
  * @type {typeof import('./types.js').whereEndoSock}
  */
 export const whereEndoSock = (platform, env) => {
@@ -69,6 +108,8 @@ export const whereEndoSock = (platform, env) => {
 };
 
 /**
+ * Returns the most suitable path for Endo caches.
+ *
  * @type {typeof import('./types.js').whereEndoCache}
  */
 export const whereEndoCache = (platform, env) => {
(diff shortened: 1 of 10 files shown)


kriskowal:
cc @jcorbin

packages/where/index.js
```diff
@@ -9,7 +9,7 @@ const { raw } = String;
 /**
  * @param {{[name: string]: string}} env
```
dckc, line 10: I wonder if this is consistent with `typeof process.env`, which, in node, seems to be `Record<string, string|undefined>`.

packages/where/index.js
```diff
@@ -24,28 +24,25 @@ const whereEndoStateWindows = env => {
...
+  return 'Endo';
 };
 
 /**
  * @type {typeof import('./types.js').whereEndoState}
  */
 export const whereEndoState = (platform, env) => {
-  if (platform === 'win32') {
-    return whereEndoStateWindows(env);
+  if (env.XDG_STATE_HOME !== undefined) {
```
dckc, line 34: The name `XDG_STATE_HOME` is not chosen by this package, right? It's intended to conform to [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) ? Please cite the external constraint.

packages/where/index.js
```diff
@@ -24,28 +24,25 @@ const whereEndoStateWindows = env => {
...
     return `${env.HOMEDRIVE}${env.HOMEPATH}\\AppData\\Endo`;
   }
-  return '.';
+  return 'Endo';
 };
 
 /**
  * @type {typeof import('./types.js').whereEndoState}
  */
 export const whereEndoState = (platform, env) => {
```
dckc, line 33: The comment atop this file is sort of disconnected. Is it better to add something to the exported functions?

If not, use [@file](https://jsdoc.app/tags-file.html)?

packages/where/index.js
```diff
@@ -59,32 +56,32 @@ export const whereEndoSock = (platform, env) => {
...
     return `/tmp/endo-${env.USER}/endo.sock`;
   }
-  return 'endo.sock';
+  return 'endo/endo.sock';
 };
 
 /**
  * @type {typeof import('./types.js').whereEndoCache}
  */
 export const whereEndoCache = (platform, env) => {
```
dckc, line 74: yay for explicit `platform, env` args rather than ambient access to `os.type()` and `process.env`.

dckc (approved):
Looks good.

We discussed a bit of prose for the test assertions; that's welcome but not critical.

packages/where/index.js
kriskowal: Added comments to individual functions and removed the vestigial file doc.

packages/where/index.js
kriskowal: Yay, testable!

packages/where/index.js
```diff
@@ -1,54 +1,90 @@
...
+      return `${env.HOME}/Library/Application Support/Endo`;
     }
   }
-  return 'endo';
+  if (env.USER !== undefined) {
+    // The XDG specification says that in this case, we should fall through to
+    // system specific behavior, but we are not in a position to assume which
+    // system we're using, other than it is probably a variety of Linux,
+    // in which case /tmp and /run might be viable base paths.
+    return `/tmp/endo-${env.USER}`;
```
jcorbin, line 79: - should respect `$TMPDIR` if set
- should be resilient to `$USER` not being set; fallback to however node binds `getent(1)`
- apply throughout to all other mentions of `env.USER`

packages/where/index.js
```diff
@@ -1,54 +1,90 @@
...
+ */
+export const whereEndoEphemeralState = (platform, env) => {
+  if (env.XDG_RUNTIME_DIR !== undefined) {
+    return `${env.XDG_RUNTIME_DIR}/endo`;
+  } else if (platform === 'win32') {
+    return whereEndoHomeWindows(env);
+  } else if (platform === 'darwin') {
     if (env.HOME !== undefined) {
-      return `${env.HOME}/.config/endo`;
+      return `${env.HOME}/Library/Application Support/Endo`;
```
jcorbin, line 71: - should be resilient to `$HOME` not being set; fallback to however node binds `getent(1)`
- apply throughout to all other mentions of `env.Home`

packages/where/index.js
```diff
@@ -59,32 +95,34 @@ export const whereEndoSock = (platform, env) => {
...
     if (env.HOME !== undefined) {
       return `${env.HOME}/Library/Application Support/Endo/endo.sock`;
     }
-  } else if (env.XDG_RUNTIME_DIR !== undefined) {
-    return `${env.XDG_RUNTIME_DIR}/endo/endo.sock`;
   } else if (env.USER !== undefined) {
     return `/tmp/endo-${env.USER}/endo.sock`;
   }
-  return 'endo.sock';
+  return 'endo/endo.sock';
```
jcorbin, line 107: - above comments imply this branch goes away, since:
  - $USER is always defined or definable via `getent(1)`
  - but do consult `$TMPDIR` here

packages/where/index.js
```diff
@@ -59,32 +95,34 @@ export const whereEndoSock = (platform, env) => {
...
     if (env.HOME !== undefined) {
       return `${env.HOME}/Library/Caches/Endo`;
     }
-  } else if (env.XDG_CACHE_HOME !== undefined) {
-    return `${env.XDG_CACHE_HOME}/endo`;
   } else if (env.HOME !== undefined) {
     return `${env.HOME}/.cache/endo`;
   }
-  return '';
+  return 'endo/cache';
```
jcorbin, line 127: likewise, this branch probably fades away; `$HOME` is always defined or definable

jcorbin: then a possible concern would be "homeless" users, like service accounts that sometimes set their home to something `/` or `/dev/null`

jcorbin (commented):
Progress :tm:

- can handle `$HOME` and `$USER` via `getent(1)` ; which I think node spells like `os.userInfo()` to be non-posix-specific ; may as well factor those out ala some kinda `getHome()` / `getUser()`
- should factor out all `/tmp` references into a "use $TMPDIR or /tmp" sort of `getTmp()` utility
- maybe spell all those suggested `getFoo()`s as `whereFoo()`s for maximum neologism

One consequence of all of this is elimination of any cwd-oriented fallthru/back branches... which is always a plus in my cold-old-sysadminy heart... but I know cwd-orientation is all the rage with kids these days...

jcorbin:
Further clarifying: various execution contexts or modalities can be very minimal with what env they populate, to include:
- docker
- sudo
- spawned by various init daemons
- ssh remote command execution sans shell

Basically: various paths to "production-ization" can reveal this lack ; idk what endo even is, or who/what/where wants to run it, but that's the deal from my pov :-)

kriskowal:
> Basically: various paths to "production-ization" can reveal this lack ; idk what endo even is, or who/what/where wants to run it, but that's the deal from my pov :-)

Thank you for these constructive notes. For now, at least at this layer, I think it’s helpful to say Endo is “Docker but JavaScript”. The Docker part implies confinement of the guest programs, and that’s intentional. As with all other “X but Y”, the framing doesn’t capture the whole, but I find that it helps inform the shape of the CLI, since there are analogues and UX patterns that we can usefully borrow to foster familiarity.

Endo is also “Browser but just Extensions”, “Browser but Decentralized”, “Chat but mostly with Bots”, or a number of other useful framings. Another useful framing is:

Microsoft unified the office productivity segment (word processor, spreadsheets, &c) with object linking and embedding (but mostly, copy and paste). Endo is (aims to be) that except improving the suite of (chat, web, apps, authentication, authorization, key management, and wallets) with object linking and embedding. You can `@` _anything_.

packages/where/index.js
kriskowal: I’m going to willfully ignore the users without home directories. Endo’s a user agent. No user, no agent.

packages/where/index.js
kriskowal: I’m using TMPDIR and falling back to `os.tmpdir()` now. Thanks.

PR #1417, opened 2022-12-23
fix(captp): Serialize serialization errors

kriskowal:
Wherein we observe that resolving a remote promise with a non-remotable causes an exception and for the receiver to hang indefinitely. This neat trick promotes the serialization error and stuffs it in the return path so the originator can observe the error.


kriskowal:
cc @michaelfig

kriskowal:
Reviewers, please note that there’s a leading commit that is a mechanical application of `yarn lint-fix` that was necessary to get the blessing of CI, but separate from the much more surgical commit that fixes the problem at hand.

gibson042 (approved):
:shipit:

PR #1418, opened 2022-12-23
chore: release

kriskowal:
- Bugfix from external contributor @Jack-Works that unbreaks MetaMask's
  Browserify-style bundles in a no-unsafe-eval Content Security Policy.
- @kumavis and @naugtur Improved Node.js and Browserify ecosystem
  compatibility to cover more of the behaviors of the `browser` field in
  `package.json`, which required work on SES Compartments to enable import hooks
  to effect module aliases, reflective imports (module imports self by own package name),
  a fix for differentiating module languages based on their extension.
  They also added CommonJS support to Endo's Compartment Mapper bootstrapping bundle
  format.
- @erights Eventual send now reflects prototype inheritance.
- @gibson042 Paid off some refactoring debt in marshalling.
- @kriskowal Endo Daemon now has debugging commands, `log`, `log -f`, and `ping`.
- @kriskowal Fix marshalling bug, such that serialization errors get
  communicated up the promise resolution chain.
- @turadg Upgraded Ava from version 3 to 5.
- @turadg Updated Prettier to 2.8 and addressed style changes.
- @turadg Made progress on homogenizing our approach to publishing TypeScript types.
- External contributor @smotaal upgrades Lerna to version 5.
- @erights Performance refactor for assertions, using Fail tag
- @kriskowal The SES bundle now uses a more hardened generated internal linkage runtime.

### DIFF
--- a/packages/marshal/NEWS.md
+++ b/packages/marshal/NEWS.md
@@ -1,5 +1,10 @@
 User-visible changes in `@endo/marshal`:
 
+# v0.8.1 (2022-12-23)
+
+- Remote objects now reflect methods present on their prototype chain.
+- Serialization errors now serialize.
+
 # v0.8.0 (2022-10-24)
 
 - Requires plain objects to inherit from Object.prototype, ensuring pass-invariance
--- a/packages/base64/CHANGELOG.md
+++ b/packages/base64/CHANGELOG.md
@@ -3,273 +3,167 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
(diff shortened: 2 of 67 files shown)


packages/marshal/NEWS.md
```diff
@@ -1,5 +1,10 @@
 User-visible changes in `@endo/marshal`:
 
+# v0.8.1 (2022-12-23)
+
+- Remote objects now reflect methods present on their prototype chain.
+- Serialization errors now serialize.
```
erights, line 6: Neither of these were changes to the marshal package. But I agree it's good to list them here too.

erights (approved):
LGTM!

PR #1421, opened 2022-12-29
update TypeScript to 4.9

turadg:
Release notes:
https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/

Most motivating feature: [the new `satisfies` operator](https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/#satisfies)

Updates the dep and others that were necessary to get back to green.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -12769,15 +12769,10 @@ typedarray@^0.0.6:
   resolved "https://registry.yarnpkg.com/typedarray/-/typedarray-0.0.6.tgz#867ac74e3864187b1d3d47d996a78ec5c8830777"
   integrity sha512-/aCDEGatGvZ2BIk+HmLf4ifCJFwvKFNb9/JeZPMulfgFracn9QFcAf5GO8B/mweUjSoblS5In0cWhqpfs/5PQA==
 
-"typescript@^3 || ^4":
+"typescript@^3 || ^4", typescript@~4.9.4:
   version "4.9.4"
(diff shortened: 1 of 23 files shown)


yarn.lock
```diff
@@ -2272,9 +2272,9 @@
   integrity sha512-jhuKLIRrhvCPLqwPcx6INqmKeiA5EWrsCOPhrlFSrbrmU4ZMPjj5Ul/oLCMDO98XRUIwVm78xICz4EPCektzeQ==
 
 "@types/node@*", "@types/node@>=14.0":
-  version "18.7.23"
-  resolved "https://registry.yarnpkg.com/@types/node/-/node-18.7.23.tgz#75c580983846181ebe5f4abc40fe9dfb2d65665f"
-  integrity sha512-DWNcCHolDq0ZKGizjx2DZjR/PqsYwAcYUJmfMWqtVU2MBMG5Mo+xFZrhGId5r/O5HOuMPyQEcM6KUBp5lBZZBg==
+  version "18.11.18"
```
mhofman, line 2275: I didn't realize we were using node 18 types. I would feel safer using node 16 types as that is more representative of the platform we support.

mhofman (approved):
LGTM!

I wish we could get the old events library out, but since it's a type only change, I'm ok with the fix taking the form of a `patch-package`

yarn.lock
turadg: True. Though IMO we can rely on the unit tests that run on 14, 15 and 18 for platform compatibility.

Regardless, out of scope of this PR but feel free to make one for that.

PR #1422, opened 2022-12-30
Endo support for unsafe Node.js plugins

kriskowal:
This change introduces a new type of sturdy reference builder/rebuilder to the Endo Daemon. Workers currently can only evaluate compartmentalized code in the presence of specified endowments. With this change, workers can now load arbitrary Node.js plugins using the `export const main0 = powers => exo` calling convention. If given a pet name, these services will be restored on demand after an Endo restart or system reboot.

This is a prelude to a future `import-bundle` command that will have the same calling convention but only operate on confined applications.

### DIFF
--- a/packages/cli/src/endo.js
+++ b/packages/cli/src/endo.js
@@ -407,6 +407,34 @@ export const main = async rawArgs => {
       }
     });
 
+  program
+    .command('import-unsafe0 <worker> <path>')
+    .option(
+      '-n,--name <name>',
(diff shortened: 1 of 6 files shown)


kriskowal:
cc @michaelfig This is intended to support your vision for arbitrary Node.js plugins. Endo workers can be induced to import arbitrary Node.js “services”. Not included in this change is the mechanism for introducing a confined application to an unconfined service API object.

PR #1423, opened 2022-12-30
WIP: Event-based encoding API based on Data-E

erights:


packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -216,13 +220,12 @@ export const makeEncodeToSmallcaps = (encodeOptions = {}) => {
         return `%${name}`;
       }
       case 'copyRecord': {
-        // Currently copyRecord allows only string keys so this will
-        // work. If we allow sortable symbol keys, this will need to
-        // become more interesting.
+        // copyRecord allows only string keys so this will
+        // work.
```
dckc, line 224: copyRecord keys can be symbols now, right?

@gibson042 and I just talked about that recently, I think.

dckc: oh... or is that Far object method names?

erights: No. Yes.

Cannot be copyRecord property names. Can be Far object method names.

erights: Though we discourage the latter until we see how receptive OCapN is to @kriskowal 's suggestions to no longer accept symbol method names per se. IOW, only string names transmitted as method names, where each side get to decide whether to translate it into a local string method name or a local symbol method name. Would require a bit of refactoring of abstraction levels, so method names have a special place, and so unlikely to be settled soon. Until then, the current symbols-also consensus stands. But we'd like to avoid unnecessarily entrenching that until this matter is settled.

Issue #1414, opened 2022-12-21
Marshal tests delegating to JSON.parse may result in unexpected SyntaxError message strings


SMotaal:
The following test will fail on newer versions of Node.js and likely other runtimes:

https://github.com/endojs/endo/blob/dbd498e30a5c3b0d2713d863bc7479ceef39cd79/packages/marshal/test/test-marshal-stringify.js#L57-L59

I would like to get some input from folks before drafting a PR from my end to reduce the friction of having to rely downgrade just to ensure this test passes.

---

#### Considerations

- It is not clear if it is the intent that marshal would delegate to `JSON.parse` for this specific case:
  - This was the only test in the `marshal parse errors` which failed on my end.
  - There are other tests that look for `{message}` strings that are obviously specific to marshal and not what `JSON.parse` itself would report.
- It it is the intent:
  - The error thrown by [`JSON.parse` per the ECMAScript specification](https://tc39.es/ecma262/#sec-json.parse) must be of the type `SyntexError`.
  - This would [explicitly determine its `{name}`](https://tc39.es/ecma262/#sec-nativeerror.prototype.name).
  - This would not determine the `{message}` which is left to the discretion of implementers.
  - This issue may also be relevant in other places as it relates to changes made recently to "improve error messages" in https://github.com/v8/v8/commit/718f743750135b3a3e643655a0406a90bf4bf010.

#### Alternatives

1. Keep revising matcher to make it work for Node.js and other specific platforms moving forward:
   ```js
     t.throws(() => parse('X'), {
       message: /\b[Un]expected\b/,
     });
   ```
2. Switch from `{message}` to `{name}` checks:
   ```js
     t.throws(() => parse('X'), {
       name: /^SyntaxError$/,
     });
   ```
3. ~~Revise the test to dynamically prepare the actual and expected values.~~ (counterintuitive)

mhofman:
Yeah it seems that checking for a `.name` to match `SyntaxError` would be the correct test to have. @mfig @erights WDYT ?

michaelfig:
Another alternative to consider:
4. Use `e instanceof SyntaxError`:
```js
t.throws(() => parse('X'), {
  instanceOf: SyntaxError,  
});
```

I'm unsure of the relative merits of `name:` vs `instanceOf:`, but either would be fine with me!

erights:
Huh. I didn't know that Ava had all these options. Neat.

[...]

PR #1425, opened 2023-01-03
test: expect `SyntaxError` instance only instead of "Unexpected token…" message

SMotaal:
Addresses: #1414

The ECMAScript spec explicitly states the types of error to be expected in the cases of [`JSON.parse`](https://tc39.es/ecma262/#sec-json.parse) and [`eval`](https://tc39.es/ecma262/#sec-performeval) leaving the message itself at the discretion of implementers.

As such, this PR removes assertions for the `message` string from across this repository, some of which fail today in newer Node.js versions tracking with https://github.com/v8/v8/commit/718f743750135b3a3e643655a0406a90bf4bf010.

Co-Authored-By: @michaelfig

### DIFF
--- a/packages/marshal/test/test-marshal-stringify.js
+++ b/packages/marshal/test/test-marshal-stringify.js
@@ -56,7 +56,7 @@ test('marshal parse errors', t => {
     message: /Marshal's parse must not encode any slot positions .*/,
   });
   t.throws(() => parse('X'), {
-    message: /Unexpected token X in JSON at position 0*/,
+    instanceOf: SyntaxError,
   });
   t.throws(() => parse('{"@qclass":8}'), {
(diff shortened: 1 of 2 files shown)


SMotaal:
@kriskowal Please see [log](https://github.com/endojs/endo/actions/runs/3831546420/jobs/6520805132)

Worth mentioning that I hit inconsistent errors on and off running `yarn test` locally today, both with and without docker, and so I kept some logs which may be useful.

erights (approved):
Assuming the intermittent CI error is unrelated (as seems likely), LGTM. Thanks!

kriskowal:
The CI error is certainly unrelated noise. I’m rerunning that test to verify that it flaps back to green.

michaelfig (approved):
LGTM!  Thanks for doing this.

Issue #1429, opened 2023-01-04
Error codes not discoverable


erights:
At https://github.com/endojs/endo/tree/master/packages/ses/error-codes is the directory of the current SES error code explanations, where the name of each file in the directory is the error code it explains.

I had initially proposed that the rendering of the error include a URL to the page explaining the error. This could be by manually writing the url as the error code where the error is defined, or by keeping the current conventions of error codes as identifiers, but then rendering the error by prepending a fixed URL base. Both of these were rejected on the theory that we did not want the source code to privilege a given host as the definitive explanation of the errors. By this theory, it would be better for users to do a web search for the error code identifier, to see all those providing explanations.

Empirically, this theory has failed. That directory and most of the files in it are now over a year old. I just did a Google search on each of these identifiers. Not one of these Google searches provided a useful result. In particular, not one produced a URL to the error explanation page itself. (Attn @dtribble )

@turadg just has an interesting observation: when investigating an error, after looking at the message, the next thing we generally look at are the source code positions captured in the error stack. A comma placed into the source code where the error is thrown from is thereby more naturally discoverable than are anything based on websearching error codes.

erights:
Mitigation in progress at https://github.com/endojs/endo/pull/1431

PR #1431, opened 2023-01-05
feat: Comment links error code errors to explanation

erights:
Mitigates https://github.com/endojs/endo/issues/1429

When seeing an error with an error code, it can be hard to find the page explaining that error code. As #1429 documents, Google searching on the error code is not helpful. However, @turadg noticed that he often uses the stack trace locations to look at the code that raised this error. This PR adds comments to the source code locations that raise errors we have already given error codes to, where these comments contain the URL of the error code explanation page we want the developer to discover.

### DIFF
--- a/packages/ses/index.js
+++ b/packages/ses/index.js
@@ -26,6 +26,7 @@ function getThis() {
 }
 
 if (getThis()) {
+  // See https://github.com/endojs/endo/blob/master/packages/ses/error-codes/SES_NO_SLOPPY.md
   throw new TypeError(`SES failed to initialize, sloppy mode (SES_NO_SLOPPY)`);
 }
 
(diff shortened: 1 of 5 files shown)


erights:
Reviewers, please review under the assumption that this PR would set a precedent for how we make error explanations discoverable in practice.

turadg (approved):
Since the files are in the same package, I thought a relative link might work, but I couldn't get VS Code to make it navigable.

This Markdown comment will render as clickable on hover, but that doesn't help with the problem at hand.
```
/** [SES_NO_EVAL](../error-codes/SES_NO_EVAL.md) */
```
This would be a little more JSDoc style and concise, but even less clickable:
```
/** {@link ../error-codes/SES_NO_EVAL.md} *
```

I expect we'll have a better way eventually. E.g. https://github.com/microsoft/TypeScript/issues/47718 . Or a little web build like https://endojs.github.io/errors/SES_NO_EVAL 

But this is already an improvement to discovery and all the above would be incremental improvements. So 👍 to merge.

kriskowal:
I approve and endorse the approach going forward.

PR #1432, opened 2023-01-07
feat(stream): Pubsub

kriskowal:
In the course of implementing Endo CLI inbox followers, I needed a minimum viable pubsub topic. Endo doesn’t have `@agoric/notifiers` yet, but it does have streams. Given that I expect notifiers to eventually make their way up to Endo, I’d accept a recommendation to instead inline this feature in the Endo Daemon in anticipation of replacing the implementation with a notifier. However, I believe this implementation stands on its own at least pedagogically.

This change introduces a succinct implementation of pubsub in terms of streams (async iterators). The `makeTopic` function is a pubsub analog for `makePipe`, but is broadcast instead of unicast, and consequently loses the ability to connect forward- and back-pressure signals. Where `makePipe` returns an entangled pair of streams, `makeTopic` produces a single `publisher` stream and a `subscribe` function for any number of subscriber streams.

The key insight that unlocks this change is that `makeQueue` is not a suitable foundation for pubsub, but is a suitable pattern to apply toward a new `makePubSub` variation on the theme. This function returns the producer side of a queue (which I’m calling a sink) and a function for generating consumers (springs). Like an async queue, a pub sub pair is backed by an async linked list, but the publisher side willfully advances the read head, such that new subscriptions only see future values and historical values are forgotten. Then, each subscriber captures a cursor so it doesn’t forget any of the links since its subscription was created.

In order to disentangle the pub and subs, `makeTopic` uses a `nullQueue`, so each pub and sub is paired with either the sink or spring side of the null queue. Consequently, I’ve found it useful to narrow the types accepted by `makeStream` to `Sink` and `Spring`, since pubs and subs only present these narrower types.

### DIFF
--- a/packages/daemon/src/pubsub.js
+++ b/packages/daemon/src/pubsub.js
@@ -1,0 +1,72 @@
+import { makePromiseKit } from '@endo/promise-kit';
+import { makeStream } from '@endo/stream';
+
+// TypeScript ReadOnly semantics are not sufficiently expressive to distinguish
+// a value one promises not to alter from a value one must not alter,
+// making it useless.
+const freeze = /** @type {<T>(v: T | Readonly<T>) => T} */ (Object.freeze);
+
+/**
+ * @template TValue TValue
+ * @param {TValue} value
+ * @returns {import('./types.js').AsyncQueue<TValue, unknown>}
+ */
+export const makeNullQueue = value =>
+  harden({
+    put: () => {},
+    get: async () => value,
+  });
+
+export const nullIteratorQueue = makeNullQueue(
+  harden({ value: undefined, done: false }),
+);
+
+/**
+ * @template TValue
+ */
+export const makeChangePubSub = () => {
+  // Request pubsub async queue internals
+  let { promise: tailPromise, resolve: tailResolve } = makePromiseKit();
+
+  const sink = {
+    /**
+     * @param {TValue} value
+     */
+    put: value => {
+      const { resolve, promise } = makePromiseKit();
+      tailResolve(freeze({ value, promise }));
+      tailResolve = resolve;
+      // Unlike a queue, advance the read head for future subscribers.
+      tailPromise = promise;
+    },
+  };
+
+  const makeSpring = () => {
+    // Capture the read head for the next published value.
+    let cursor = tailPromise;
+    return {
+      get: () => {
+        const promise = cursor.then(next => next.value);
+        cursor = cursor.then(next => next.promise);
+        return harden(promise);
+      },
+    };
+  };
+
+  return harden({ sink, makeSpring });
+};
+harden(makeChangePubSub);
+
+/**
+ * @template TValue
+ * @returns {import('./types.js').Topic<TValue>}
+ */
+export const makeChangeTopic = () => {
+  /** @type {ReturnType<makeChangePubSub<TValue>>} */
+  const { sink, makeSpring } = makeChangePubSub();
+  return harden({
+    publisher: makeStream(nullIteratorQueue, sink),
+    subscribe: () => makeStream(makeSpring(), nullIteratorQueue),
+  });
+};
+harden(makeChangeTopic);
--- a/packages/daemon/package.json
+++ b/packages/daemon/package.json
@@ -49,6 +49,8 @@
     "ses": "^0.18.4"
   },
   "devDependencies": {
+    "@endo/init": "^0.5.56",
+    "@endo/ses-ava": "^0.2.40",
     "ava": "^5.3.0",
     "babel-eslint": "^10.0.3",
     "c8": "^7.14.0",
--- a/packages/daemon/src/types.d.ts
+++ b/packages/daemon/src/types.d.ts
@@ -1,4 +1,4 @@
-import type { Reader, Writer } from '@endo/stream';
+import type { Reader, Writer, Stream } from '@endo/stream';
 
 export type Locator = {
   statePath: string;
@@ -114,3 +114,13 @@ export type Request = {
 };
 
 export type Message = Label & Request;
+
+export interface Topic<
+  TRead,
+  TWrite = undefined,
+  TReadReturn = undefined,
+  TWriteReturn = undefined,
+> {
+  publisher: Stream<TWrite, TRead, TWriteReturn, TReadReturn>;
+  subscribe(): Stream<TRead, TWrite, TReadReturn, TWriteReturn>;
+}
--- a/packages/daemon/test/test-pubsub.js
+++ b/packages/daemon/test/test-pubsub.js
@@ -1,0 +1,131 @@
+// @ts-check
+/* eslint-disable no-await-in-loop */
+
+import '@endo/init/debug.js';
+
+import rawTest from 'ava';
+import { wrapTest } from '@endo/ses-ava';
+import { makePromiseKit } from '@endo/promise-kit';
+import { makeChangeTopic } from '../src/pubsub.js';
+
+const test = wrapTest(rawTest);
+
+test('change topic supports parallel subscriptions', async (/** @type {import('ava').Assertions} */ t) => {
+  const { publisher, subscribe } = makeChangeTopic();
+
+  // Coordinated checkpoints.
+  const [p1, p2, p4, s1, s2, s3, s4] = (function* generatePromiseKits() {
+    for (;;) yield makePromiseKit();
+  })();
+
+  await Promise.all([
+    (async () => {
+      // This await ensures that the synchronous portion of the first, full
+      // subscription obtains a subscriber before it yields to the event loop
+      // the first time.
+      await null;
+      await publisher.next(1);
+      p1.resolve();
+
+      await s1.promise;
+      await publisher.next(2);
+      p2.resolve();
+
+      await Promise.all([s2.promise, s3.promise]);
+      await publisher.next(3);
+
+      await publisher.next(4);
+      p4.resolve();
+
+      await s4.promise;
+      await publisher.return('EOL');
+    })(),
+
+    // Full subscription.
+    (async () => {
+      let expected = 1;
+      const subscription = subscribe();
+      for await (const actual of subscription) {
+        t.is(actual, expected);
+        expected += 1;
+      }
+      t.is(expected, 5);
+    })(),
+
+    // Delayed subscriber.
+    (async () => {
+      await p1.promise;
+      const subscription = subscribe();
+      s1.resolve();
+
+      let expected = 2;
+      for await (const actual of subscription) {
+        t.is(actual, expected);
+        expected += 1;
+      }
+      t.is(expected, 5);
+    })(),
+
+    // Further delayed subscriber.
+    (async () => {
+      await p2.promise;
+      const subscription = subscribe();
+      s2.resolve();
+
+      let expected = 3;
+      for await (const actual of subscription) {
+        t.is(actual, expected);
+        expected += 1;
+      }
+      t.is(expected, 5);
+    })(),
+
+    // Same subscription timing as previous, but using the iterator protocol to
+    // see the final value.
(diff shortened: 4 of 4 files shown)


kriskowal:
cc @michaelfig, reductive pubsub!

packages/stream/README.md
```diff
@@ -73,6 +73,33 @@ which promises settle.
...
+```
+
+Topics are very similar to pipes, but instead of using `makeQueue`,
+topics use a very similar `makePubSub()`, which produces one sink and any
+number of springs encapsulating a shared async linked list.
+The topic writer is a stream constructed from the sink and a null spring.
+The null spring provides no forward-pressure.
+The topic readers are streams constructed from a null sink (so no
+back-pressure) and a subscriber spring that serves as an independent
+cursor starting with the next published value.
```
michaelfig, line 101: To avoid unnecessary incompatibility, subscribers should start iterating at the head of the async list (i.e. a promise for the latest published value, which is only the next published one if the async list is currently empty)

packages/stream/index.js
```diff
@@ -49,13 +49,57 @@ export const makeQueue = () => {
...
+  get: async () => {},
+});
+
+/**
+ * @template TValue
+ */
+export const makePubSub = () => {
+  // Request pubsub async queue internals
+  let { promise: tailPromise, resolve: tailResolve } = makePromiseKit();
+
```
michaelfig, line 66:
```suggestion
  let headPromise = tailPromise;

```

packages/stream/index.js
```diff
@@ -49,13 +49,57 @@ export const makeQueue = () => {
...
+  const pub = {
+    /**
+     * @param {TValue} value
+     */
+    put: value => {
+      const { resolve, promise } = makePromiseKit();
+      tailResolve(freeze({ value, promise }));
+      tailResolve = resolve;
+      // Unlike a queue, advance the read head for future subscribers.
+      tailPromise = promise;
```
michaelfig, line 76:
```suggestion
      headPromise = tailPromise;
      tailPromise = promise;
```

packages/stream/index.js
```diff
@@ -49,13 +49,57 @@ export const makeQueue = () => {
...
+    },
+  };
+
+  const sub = () => {
+    // Capture the read head for the next published value.
+    let cursor = tailPromise;
```
michaelfig, line 82:
```suggestion
    // Capture the read head for the latest published value.
    let cursor = headPromise;
```

packages/stream/README.md
```diff
@@ -73,6 +73,33 @@ which promises settle.
...
+The `makeTopic` function returns a kit with a `publisher` and a `subscribe`
+function.
+The `publisher` is a writer, but provides no back-pressure.
+The `subscribe` function returns readers.
+
+```js
+const { publisher, subscribe } = makeTopic();
+const subscriber = subscribe();
+publisher.next(value);
+for (const value of subscriber) {}
```
michaelfig, line 91:
```suggestion
for await (const value of subscriber) {}
```

michaelfig (commented):
Looks good!  I'm not approving yet until you have a chance to address my comments (possibly through rebuttal).

BTW, I don't think `@agoric/notifier` should become `@endo/notifier`.  I agree with providing something simple and robust in `@endo/stream` that could be used directly or later wrapped by `@agoric/notifier` to provide its current abstractions.

There seem to be some arbitrary gaps between `@agoric/notifier` Subscribers and `makeTopic` that could be smoothed over:
- the `subscribe` created by `makeTopic` should be remotable.  Because of that, we require adapter functions to handle the case of creating a near AsyncIterable from a far `subscribe`.  A near AsyncIterable can't be created by `E(subscribe)()` because we need near behaviour (not just passable data or remote behaviour).  `ai[Symbol.asyncIterator]()` is called immediately and cannot return a promise.
- `subscribeEach(subP)` would be the replacement I'm suggesting for the current `subscribe()` function.  In the future, we could have `subscribeLatest(subP)`

I made some suggestions to change `makePubSub` to also track the `headPromise`, not just the `tailPromise`.

kriskowal:
Very interesting observation about `subscribeLatest` and `subscribeEach`. The use case this factors out of definitely needs to follow the tail, so suggests we need another verb like `subscribeEachUpdate` (lossless). I don’t think `subscribeLatest` falls out of this kind of pubsub. Or rather, our experience with Casting suggests that `subscribeLatest` is best implemented as a lazy async iterator, rather than an eager async iterator that gathers but discards intermediate values. However, a pressure relief adapter on an async reader is definitely possible:.

packages/stream/README.md
kriskowal: What do you think of:

- `makeChangeTopic()` subscribers start with the next change.
- `makeUpdateTopic()` subscribers start with the last published value, or wait for the first published value.
- `makeLatestTopic()` subscribers start with the last published value and may miss intermediate updates.

`makeChangeTopic()` is useful for Endo’s inbox, which looks like:

```js
const requests = new Map();
const requestsTopic = makeTopic();
async function* followInbox() {
  const pendingRequests = Array.from(requests.values());
  const subsequentRequests = requestsTopic.subscribe();
  yield* pendingRequests;
  yield* subsequentRequests();
}
```

It doesn’t fit the `makeUpdateTopic` or `makeLatestTopic` idiom because the range is discrete changes instead of replacement values.

michaelfig: I'm thinking that the style of iteration should be determined entirely by the subscriber (implementation and names can be bikeshodden):

* `subscribeAllFuture(topic)` - get all future published values
* `subscribeSomeFuture(topic)` - get some (i.e. skip when possible) future published values
* `subscribeLatestThenAllFuture(topic)` - get the latest published value followed by all future published values
* `subscribeLatestThenSomeFuture(topic)` - get the latest published value, followed by some (skip when possible) future published values

As for the topics, I believe there are only push and pull modes needed (again, the implementation should be idiomatic to your other abstractions in `@endo/stream`):

* `const { pusher, topic } = makePushTopicKit();` - a topic where the publisher calls `pusher.next(valueP)` to append the corresponding value to `topic`'s async list
* `const topic = makePullTopic(puller);` - a topic which calls `puller.next()` to get a new value to append to the async list on demand (when the first subscriber reads past its end)

kriskowal: I disagree. I believe the creator of the topic stands in a position where they know the semantics of the value they are publishing and with those semantics there is a natural subscription pattern. For example, if I were to publish ETAs or snapshots, I should use `makeLatestTopic`. If I were to publish deltas, patches, or operational transforms, I should use `makeChangeTopic`, where some intermediary is in a position to synchronously grab a snapshot and a change subscription in the same term. Agoric notifiers are a third category, with behavior like I propose for `makeUpdateTopic`, which is suitable for updating snapshots but doesn’t shed load like `makeLatestTopic` would.

kriskowal:
This change now includes three flavors of pubsub, and documentation for what they do and what they’re suited for, and how they optimize transit. I’ve also added an independently reviewable commit that adds `Far` wrappers to all remoteable objects. To that end, it was necessary to rewrite `mapWriter` in terms of the iterator protocol since we can’t use `Far` on async iterators produced by async generator functions (because they have a funny prototype chain).

packages/stream/README.md
```diff
@@ -73,6 +73,70 @@ which promises settle.
...
+Consumers do not need to expressly dispose of their subscriber.
+Releasing the subscriber for garbage collection is sufficient.
+
+The `makeTopic` functions returns a kit with `publish` and `subscribe`
+functions.
+The `publish` function returns a writer but provides no back-pressure.
+The `subscribe` function returns a reader and also provides no back-pressure.
+
+```js
+const { publish, subscribe } = makeLatestTopic();
```
michaelfig, line 124: This example cannot work when Alice calls `makeLatestTopic()`, Bob uses `publisher` and Carol uses `subscriber1`.

It is alright with me if you know this and don't yet want to add the ability to have pubsub across vats or a captp loopback, but please add a TODO.

```suggestion
// TODO: demonstrate usage of topics across separate vats.
const { publish, subscribe } = makeLatestTopic();
```

packages/stream/index.js
```diff
@@ -34,7 +34,7 @@ const freeze = /** @type {<T>(v: T | Readonly<T>) => T} */ (Object.freeze);
  */
 export const makeQueue = () => {
   let { promise: tailPromise, resolve: tailResolve } = makePromiseKit();
-  return {
+  return Far('AsyncQueue', {
```
michaelfig, line 37: Kits are not useful when they are declared as `Far`, since that means the kit is a single ocap, and when sent to a different vat, it has no locally-separable methods.

I'm picking on `makeQueue` here, but it applies to the other changes in this PR.

Presences can be used like `E(queueP).get()` and `E(queueP).put(val)`, but `const { get, put } = await queueP;` just makes `get` and `put` `undefined`.

However, if you instead make it a hardened record of far functions (which doesn't change the local behaviour of your code), like:
```js
return harden({
   get: Far('AsyncQueueGet', () => { ... }),
   put: Far('AsyncQueuePut', (value) => { ... }),
});
```

then that's what I meant earlier by leaning into far functions.  The remote usage would be:
```
const { get: getP, put: putP } = await queueP;
await E(getP)();
await E(putP)(value);
```

Again, some TODOs would be sufficient, since without a captp test of these abstractions we can't verify their remote behaviour is correct.

michaelfig (approved):
My comments are nitpicky, but I'd be fine with you landing this knowing that it still needs work to be usable across vats or a loopback captp.

Doing that work without changing the local API  might lead to finding some bugs in the marshal+captp system, since we haven't worked much with records of far functions as opposed to far objects.  That would be a worthwhile thing to do, but tackling it now is probably out of scope.

michaelfig:
> we can’t use `Far` on async iterators produced by async generator functions (because they have a funny prototype chain).

I've used the following approach to paper over problems like this, if I understand what you're saying.

```js
const makeNearAsyncIterable = aitP => Far('AsyncIterable', {
  [Symbol.asyncIterator]: () => {
    const ait = E(aitP)[Symbol.asyncIterator]();
    return Far('AsyncIterator', {
      next: (...args) => E(ait).next(...args),
    };
  },
});
const ait = makeNearAsyncIterable(asyncGenerator());
```

kriskowal:
I’m holding this change off `master` until after @erights moves the core of Agoric SDK pubsub up to Endo so I can evaluate whether this layering fits within the bigger picture.

kriskowal:
I’ve trimmed this PR down to just an internal implementation of makeChangeTopic that we can substitute for a notifier without changing any public interfaces. My intention is to get this on master before notifier’s have moved. Please consider whether you’re on board with this direction @mfig @erights.

packages/daemon/src/pubsub.js
```diff
@@ -0,0 +1,72 @@
...
+};
+harden(makeChangePubSub);
+
+export const makeChangeTopic = () => {
+  const { pub, sub } = makeChangePubSub();
+  return Far('Topic', {
+    publish: () => makeStream(nullIteratorQueue, pub()),
+    subscribe: () => makeStream(sub(), nullIteratorQueue),
+  });
+};
```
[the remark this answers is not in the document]
kriskowal: Great, I grok the motivation for making the publish and subscribe powers more easily separable.

Can I talk you into `publishable` and `subscribable`, or `pubable` and `subable`, or `publication` and `subscription` (best words, weakest correspondence, alas)?

I take “topic” to mean the shared state of publishers and subscribers, the thing implied underlying a the kit in this case.

kriskowal:
Plan remains to substitute this pubsub with the light-client approach from Agoric-SDK. I’m fully convinced that this approach, while interesting, isn’t adequate. The Pet Daemon spike in `endo` has is tall with weeds so closing all these out with an expectation that I’ll have to retcon the spike.

PR #1435, opened 2023-01-08
Run confined services in Endo

kriskowal:
This change adds commands `endo bundle <path> -n <bundleName>` and `endo import-bundle0 <workerName> <bundleName> -n <resultName>` that enable the user to create a bundle of an Endo program and then run that program in a confined compartment of a particular worker.

### DIFF
--- a/packages/cli/src/endo.js
+++ b/packages/cli/src/endo.js
@@ -39,6 +39,7 @@ import {
   makeArchive,
   writeArchive,
 } from '@endo/compartment-mapper';
+import bundleSource from '@endo/bundle-source';
 import {
   makeReadPowers,
   makeWritePowers,
@@ -53,6 +54,8 @@ const { write } = writePowers;
 const packageDescriptorPath = url.fileURLToPath(
   new URL('../package.json', import.meta.url),
 );
+
+const textEncoder = new TextEncoder();
 
 const { username, homedir } = os.userInfo();
 const temp = os.tmpdir();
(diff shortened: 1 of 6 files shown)


packages/cli/src/endo.js
```diff
@@ -277,6 +280,30 @@ export const main = async rawArgs => {
       }
     });
 
+  program
+    .command('bundle <application-path>')
```
dckc, line 284: What's an application path? Perhaps I know enough to guess that it's the root module path that I would have given to `bundleSource()`, but how about the hapless `endo bundle --help` user?

Does `command` support extended descriptions with examples and such?

I'd like to see a short example, something like:

> Given `app1.js` that imports `mod1.js`, you can bundle them with `endo bundle index.js -n app1`.

Then... what object does `app1` get bound to? Again, I can perhaps guess that it's the module namespace exported by `app1.js`, but we should share that info with `endo bundle --help` users.

packages/daemon/src/worker.js
```diff
@@ -59,6 +59,17 @@ export const makeWorkerFacet = (getDaemonBootstrap, cancel) => {
       const namespace = await import(path);
       return namespace.main0(powerBox);
     },
+
+    importBundle0: async readable => {
+      const bundleText = await E(readable).text();
+      const bundle = JSON.parse(bundleText);
+
+      const { importBundle } = await import('@endo/import-bundle');
```
dckc, line 67: Why the dynamic import?

kriskowal: Possibly a premature optimization, but it seemed wasteful to bring in the import-bundle machinery for workers that don’t use it. That is, workers that just evaluate confined programs, for now.

packages/cli/src/endo.js
kriskowal: Yeah, `commander` does support extended usage. I need to do a `--help` documentation pass over all of them.

packages/daemon/src/worker.js
dckc: ah. makes sense.

perhaps worth a comment.

kriskowal: Sure thing. A commit will appear in my next push.

packages/daemon/src/worker.js
```diff
@@ -66,6 +66,17 @@ export const makeWorkerFacet = ({
...
+      const namespace = await importBundle(bundle, {
+        endowments,
```
dckc, line 76: it's pretty inconvenient to not have `console`

PR #1436, opened 2023-01-11
compartment-mapper bundler: ensure archive execution machinery can be bundled

kumavis:
relies on a potentially controversial feature: exposing evasive transforms from ses package

blocked by: https://github.com/endojs/endo/pull/1454

### DIFF
--- a/packages/ses/package.json
+++ b/packages/ses/package.json
@@ -42,6 +42,9 @@
       "import": "./index.js",
       "require": "./dist/ses.cjs"
     },
+    "./transforms": {
+      "import": "./src/transforms.js"
+    },
     "./package.json": "./package.json"
   },
   "scripts": {
--- a/packages/compartment-mapper/test/fixtures-bundle-self/index.js
+++ b/packages/compartment-mapper/test/fixtures-bundle-self/index.js
@@ -1,0 +1,6 @@
+import { parseArchive } from '../../src/import-archive.js';
+
+export async function executeArchive(archive, fixture, globals) {
+  const application = await parseArchive(archive, fixture);
+  return application.import({ globals });
+}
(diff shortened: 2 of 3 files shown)


packages/ses/package.json
```diff
@@ -42,6 +42,9 @@
       "import": "./index.js",
       "require": "./dist/ses.cjs"
     },
+    "./transforms": {
+      "import": "./src/transforms.js"
```
kriskowal, line 46: cc @michaelfig Elsewhere and elsewhen we’ve been disciplined to support versions of Node.js and tools that have not caught up and don’t support `exports`. So, I’d want a `./transforms.js` that reëxports `./src/transforms.js`.

While I hesitate to export internal SES utilities, since that makes them part of the supported API going forward, I’m leaning toward allowing this through.

But, toward that end, it would be good to leave a comment to that effect in `./src/transforms.js` so we can purposefully avoid breaking-changes to that module’s exported interface.

Or, we could also selectively reexport certain functions from `./transforms.js` to reduce risk.

Also, we can’t harden internal utility functions. Maybe `./transform.js` should be a post-lockdown repackaging of `./src/transform.js`.

kriskowal: Coming back from today’s Endo meeting, the approach above is approved, with the caveat that we must go into `./src/transforms.js` and freeze everything exposed up to any reference that `Lockdown` will eventually harden. So, if we only export arrow functions and those functions return primitives, a simple `freeze` of the exported function should suffice. We should only reexport the utilities that have a proven need.

https://www.youtube.com/watch?v=lvOTc_HN9Uk

kumavis: moved to https://github.com/endojs/endo/pull/1454

kumavis:
replaced by https://github.com/endojs/endo/pull/1449

PR #1437, opened 2023-01-11
feat(compartment-mapper): bundle with "magic lines wrapped functor"

kumavis:
rough draft of a secure bundler via "magic lines wrapped functor"

### DIFF
--- a/packages/compartment-mapper/src/bundle.js
+++ b/packages/compartment-mapper/src/bundle.js
@@ -12,6 +12,7 @@
 /** @typedef {import('./types.js').WriteFn} WriteFn */
 /** @typedef {import('./types.js').ArchiveOptions} ArchiveOptions */
 
+import fs from 'fs';
 import { resolve } from './node-module-specifier.js';
 import { compartmentMapForNodeModules } from './node-modules.js';
 import { search } from './search.js';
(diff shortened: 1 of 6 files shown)


packages/compartment-mapper/src/bundle.js
```diff
@@ -248,11 +255,155 @@ export const makeBundle = async (read, moduleLocation, options) => {
...
+      'use strict';
+      return (
+${functorSrc}
+      );
+    };
+  }
+  }
+}).call(getEvalKitForModule(${moduleIndex}))()`)
+  return wrappedSrc
+}
```
naugtur, line 276: We could reuse the original wrappings by importing the evaluator making funciton and calling .toString() on it, then splitting on the line with eval.
I'm not sure if it's not too much work to be worth it, but it'd guarantee compatibility between the two methods as SES develops.

kumavis: i was thinking that if we wanted to keep the copy pasta for someone reason but still wanted to ensure it matches ses that we could have a test that checks the ses version and fails if it doesnt match expectations

kumavis:
replaced by https://github.com/endojs/endo/pull/1449

PR #1438, opened 2023-01-12
chore(pass-style): Scaffold new package

kriskowal:
- chore(pass-style): Scaffold
- chore: Update skeleton and scaffolder

[Recording of process](https://youtu.be/SCOYsE5G7Xk)


packages/pass-style/package.json
```diff
@@ -0,0 +1,74 @@
+{
+  "name": "@endo/pass-style",
+  "version": "0.1.0",
+  "private": null,
+  "description": "Description forthcoming.",
```
kriskowal, line 5: 👆 Needs to be filled in.

skel/test/test-index.js
```diff
@@ -0,0 +1,5 @@
+import { test } from './prepare-test-env-ava.js';
+
+test('place holder', async t => {
```
kriskowal, line 3: `ava` complains about empty suites.

erights (changes requested):
Somehow you missed one of our lint settings. After experimentally moving the passStyleOf related sources to the new package, a `yarn lint` produces a ton of

```
  36:21  error  Replace `v` with `(v)`  prettier/prettier
```

kriskowal:
> Somehow you missed one of our lint settings. After experimentally moving the passStyleOf related sources to the new package, a `yarn lint` produces a ton of
> 
> ```
>   36:21  error  Replace `v` with `(v)`  prettier/prettier
> ```

`yarn lint-fix` did this to the scaffolded project too. @turadg Do you know why the prettier configuration could have drifted?

erights:
I just uploaded https://github.com/endojs/endo/pull/1439 , first, to test this PR. CI run https://github.com/endojs/endo/actions/runs/3908177607/jobs/6678096216 demonstrates the lint problem with the scaffold I mention above.

turadg:
> why the prettier configuration could have drifted?

Looks like the new package is missing `    "arrowParens": "avoid",` in its `prettier` config in its package.json.  The root cause IMO is that each package has its own Prettier config, making it possible for them to differ. 

agoric-sdk treats code formatting as a repo-wide concern,
https://github.com/Agoric/agoric-sdk/blob/f4143fe13363a763e37f163225f4684028786663/package.json#L48-L50

kriskowal:
I’ll amend the scaffold skeleton for now to include the new prettier override. I would very much like to get to a place where lint configuration is centralized (and preserve localized commands to drive it). I’d also like to get to a place where the scaffold is itself a package that lerna will update and lint will verify, but there are some minor obstacles.

turadg:
> I would very much like to get to a place where lint configuration is centralized

Cool, filed as https://github.com/endojs/endo/issues/1440

kriskowal:
I’ve pushed fixes for the problems surfaced above. The prettier config is synchronized, the eslint overrides are captured, both in the scaffold and the skeleton from which it’s generated.

erights (approved):
All my issues are addressed. My test PR #1439 now passes CI except for remaining lint problems which obviously are not due to any weakness in this PR.

LGTM, thanks!

PR #1439, opened 2023-01-13
feat(pass-style): Extract passStyleOf and friends from marshal into the new pass-style package

erights:
Staged on #1438 
Fixes: #1441 
Fixes: #1442
Fixes: #1443.

To test https://github.com/endojs/endo/pull/1438 , I moved the passStyleOf level of abstraction from marshal into this new pass-style scaffold. But I did it as a separate PR so we can keep https://github.com/endojs/endo/pull/1438 as a record of what the scaffold looks like. 

This test has so far revealed a linting problem, but otherwise seems to work. As https://github.com/endojs/endo/pull/1438 is fixed, I'll continually rebase this PR on it to test the newer candidates.

I started out planning to only do a little of the passStyleOf migration, just to test https://github.com/endojs/endo/pull/1438 . But once I started I couldn't stop. Seems to be close to complete.


packages/marshal/test/prepare-test-env-ava.js
```diff
@@ -1,6 +1,8 @@
 // @ts-nocheck
+// eslint-disable-next-line import/no-extraneous-dependencies
```
erights, line 2: Why do I now need these?

packages/pass-style/test/prepare-test-env-ava.js
```diff
@@ -1,5 +1,7 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
erights, line 1: I was curious why I seemed to need these in marshal but not in pass-style. So I looked at it in the IDE and see that I do seem to need it here too. `yarn lint` isn't the problem. Rather, it is vscode:

image

erights: If we truly do need these lines into into prepare-test-env-ava.js, they should come from the scaffold, not this PR.

turadg: VSCode is showing an eslint error. `yarn lint` should show the same. It isn't?

erights: > VSCode is showing an eslint error. yarn lint should show the same. It isn't?

It was not.

packages/marshal/test/prepare-test-env-ava.js
```diff
@@ -1,6 +1,8 @@
...
 import '@endo/init/debug.js';
 
+// eslint-disable-next-line import/no-extraneous-dependencies
 import { wrapTest } from '@endo/ses-ava';
```
kriskowal, line 6: It is especially unclear to me how this dependency is extraneous. It deliberately eschew’s the ordering requirements, but is not extraneous by any measure.

packages/pass-style/src/error.js
```diff
@@ -24,34 +24,44 @@ const errorConstructors = new Map([
 export const getErrorConstructor = name => errorConstructors.get(name);
 harden(getErrorConstructor);
 
+const checkErrorLike = (candidate, check = undefined) => {
+  const reject = !!check && (details => check(false, details));
+  // TODO: Need a better test than instanceof
```
kriskowal, line 29: Do you hope for a brand check?

This may be a good opportunity to add an issue link to the TODO comment.

kriskowal (approved):

packages/marshal/test/prepare-test-env-ava.js
erights: IIRC, the actual complaint is that the dependency should be in `dependencies` rather than `devDependencies`.

packages/pass-style/src/error.js
erights: In @ljharb's meaning of the term, yes. The error stack proposal would let us do that. In fact, our partial shimming of that proposal may already do that on v8, firefox, and xs, depending on what their corresponding primitives do. Worth looking at! But not in this PR.

In this PR, the code within `checkErrorLike`, including this comment, was just moved from code within a method of the ErrorHelper. This PR is not trying to improve the semantics of any of that.

packages/marshal/src/marshal-justin.js
```diff
@@ -112,6 +113,9 @@ const makeNoIndenter = () => {
 };
 
 const identPattern = /^[a-zA-Z]\w*$/;
+harden(identPattern);
+const AtAtPrefixPattern = /^@@(.*)$/;
```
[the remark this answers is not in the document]
erights: Let's leave this conversation unresolved, so that anyone visiting this PR sees this ;)

packages/pass-style/test/prepare-test-env-ava.js
erights: No longer a problem. Resolving.

packages/marshal/test/prepare-test-env-ava.js
erights: Resolving.

packages/marshal/test/prepare-test-env-ava.js
kriskowal: That is an odd failure because our eslint-config is set up like:

```js
    "import/no-extraneous-dependencies": [
      "error",
      {
        "devDependencies": [
          "**/*.config.js",
          "**/*.config.*.js",
          "*test*/**/*.js",
          "demo*/**/*.js",
          "scripts/**/*.js"
        ]
      }
    ],
```

So anything under `test` should be running in the `dev` context and not require full-fledged `dependencies`.

erights:
After @kriskowal merged #1438 , I got this PR working in all ways except for lint errors, as represented by the first commit.

Of course I want to be green under CI before merging. But for many of the type errors reported by lint, I could not figure them out, so I silenced them instead, as shown in the second commit https://github.com/endojs/endo/pull/1439/commits/1046a4e5b8e63edb71dd8d13706ac185ee6e0e46 . Advice? Do you mind if I merge this to unblock @kriskowal, while filing an issue to undo the damage of that second commit?

@turadg @kriskowal , I hit the "Re-request review" button on both of you because of this question. Thanks.

erights:
Oh damn. We are indeed using `Remotable` as both a type name in some places and as a function name in other places. For some reason we happened to be getting away with this before, which is kinda terrible. After extraction, for some equally mysterious reason we are no longer getting away with this. But fixing either presents compat dangers we'd need to think about. After all the fixes of the second commit, captp now fails to check due to this collision.

packages/pass-style/src/error.js
ljharb: Incidentally, pattern matching would also provide such a check for errors.

packages/marshal/test/test-dot-membrane.js
```diff
@@ -1,6 +1,8 @@
 import { test } from './prepare-test-env-ava.js';
+
+// eslint-disable-next-line import/order
```
turadg, line 3: why is this suppression necessary?

turadg (commented):
>  I hit the "Re-request review" button on both of you because of this question.

The type suppressions seem appropriate given the scope of this PR. The comments leave a good trail for investigation. It's not totally surprising that changing the dependency graph would yield different resolution of the ambient types. 

I would like to see the import/order suppressions removed or justified with a comment.

For that reason and because I'm not fully acquainted with this code or the PR motivation I'll make this just a Comment and not Approve.

packages/marshal/test/test-dot-membrane.js
erights: Because of the side effects performed by the `./prepare-test-env-ava.js` import, it must come first. But we apparently have a lint rule that effectively says that `./...` imports must come after `@...` imports. When I remove this suppression and `yarn lint` I get

```
/Users/markmiller/src/ongithub/endojs/endo/packages/marshal/test/test-dot-membrane.js
  3:1  error  `@endo/pass-style` import should occur before import of `./prepare-test-env-ava.js`  import/order

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

In this case ;) vscode gives feedback that agrees with `yarn lint`

image

erights: Even before this PR, we seem to have a lot of these in our system

image

so it seems out of place to add an inline comment for this one in particular.

erights: See https://github.com/endojs/endo/issues/1446 . With that filed, resolving.

PR #1448, opened 2023-01-15
fix: move arb-passable to pass-style

erights:
Move `tools/arb-passable.js` from `@endo/marshal` to `@endo/pass-style`

`tools/arb-passable.js` mainly exports `arbPassable`, which is a so-called fast-check "arbitrary" for generating Passable values for use in fuzz testing. Even though it is currently only used by tests in `@endo/marshal`, it is defined only in terms of concepts at the `@endo/pass-style` level of abstraction. Any consumer of `@endo/pass-style` may wish to use it to write fuzz tests over all possible Passables.

For example, `test-amountProperties.js` currently defines an `arbAmount` which is supposed to generate arbitrary amounts, for fuzz testing code that operates on amounts. But the current form is too trivial to be useful. 

https://github.com/Agoric/agoric-sdk/pull/6769 moves `arbAmount` to its own similar `tools/arb-amount.js` file, but does not yet repair the triviality of `arbAmount`. To define a good `arbAmount`, it should be built on `arbPassable`. After this PR, it would be able to import `arbPassable` from `@endo/pass-style`


packages/pass-style/tools/arb-passable.js
```diff
@@ -1,19 +1,22 @@
 // @ts-check
 import '../src/types.js';
```
erights, line 2: Uniquely for this js file, github is not doing any syntax highlighting for me. Do you see it highlighted? Any idea why I am not?

packages/pass-style/tools/arb-passable.js
```diff
@@ -107,3 +110,14 @@ const { arbDag } = fc.letrec(tie => {
  * A factory for arbitrary passables
  */
 export const arbPassable = arbDag.map(x => harden(x));
+
+// NOTE: Not hardened because the arbs it contains cannot generally be hardened.
+// TODO: Why not? What would be needed so that arbs can be hardened?
```
erights, line 115: @gibson042 any idea why arbs cannot be hardened? I originally wrote this as `harden` rather than `freeze` and testing failed with a failed assignment.

packages/marshal/test/test-rankOrder.js
```diff
@@ -5,6 +5,7 @@ import { test } from './prepare-test-env-ava.js';
 // eslint-disable-next-line import/no-extraneous-dependencies
 import { fc } from '@fast-check/ava';
 import { makeTagged } from '@endo/pass-style';
+import { arbPassableKit } from '@endo/pass-style/tools.js';
```
kriskowal, line 8: You could revert to exporting the items by name in this new form, to avoid the separate destructure.

packages/pass-style/tools/arb-passable.js
kriskowal: I’m seeing highlights.

kriskowal (approved):

packages/marshal/test/test-rankOrder.js
erights: Good point. Done

PR #1449, opened 2023-01-18
Secure bundler hybrid

kumavis:


packages/compartment-mapper/src/archive.js
```diff
@@ -204,27 +205,38 @@ const renameSources = (sources, compartmentRenames) => {
 };
 
 /**
- * @param {ArchiveWriter} archive
  * @param {Sources} sources
  */
-const addSourcesToArchive = async (archive, sources) => {
+export const locationsForSources = function* (sources) {
```
kumavis, line 210: are generators ok? I don't see a lot of them around the code base

packages/compartment-mapper/src/bundle.js
```diff
@@ -298,6 +340,164 @@ ${''.concat(...modules.map(m => m.bundlerKit.getFunctorCall()))}\
...
+${makeStrictScopeHandler}
+${getEvalKitForCompartment}
+const moduleFunctors = ${renderFunctorTable(moduleFunctors)}
+// END MODULE REGISTRY ==================================
+
+const { compartments, execute } = loadApplication(
+  compartmentMap,
+  moduleRegistry,
+  moduleFunctors,
+  'xxx',
```
kumavis, line 486: fix

packages/compartment-mapper/src/bundle.js
```diff
@@ -298,6 +340,164 @@ ${''.concat(...modules.map(m => m.bundlerKit.getFunctorCall()))}\
...
+const compartmentMap = ${JSON.stringify(archiveCompartmentMap, null, 2)};
+const moduleRegistry = ${JSON.stringify(moduleRegistry, null, 2)}
+
+const strictScopeTerminator = makeStrictScopeHandler();
+${makeStrictScopeHandler}
+${getEvalKitForCompartment}
+const moduleFunctors = ${renderFunctorTable(moduleFunctors)}
+// END MODULE REGISTRY ==================================
+
+const { compartments, execute } = loadApplication(
```
kumavis, line 482: move into runtime?

packages/compartment-mapper/test/test-bundle.js
```diff
@@ -75,3 +82,103 @@ test('equivalent archive behaves the same as bundle', async t => {
...
+  vmEval('lockdown()');
+  const appExecPromise = vmEval(bundle);
+  const {
+    namespace: { myGlobalThis, myEval },
+  } = await appExecPromise;
+  // ensure the modules compartment global is not the realm global or context object
+  // nodejs vm implements a half-baked membrane with the vm realm globalThis and the context object
+  // wrapping in an array defeats the membrane
+  t.not(vmEval('globalThis'), myGlobalThis);
+  t.not(vmEval('[globalThis][0]'), myGlobalThis);
```
kumavis, line 181: i think that was a mistake
```suggestion
  t.not(vmEval('[globalThis]')[0], myGlobalThis);
```

packages/compartment-mapper/src/bundle-runtime.js
```diff
@@ -0,0 +1,164 @@
...
+  reference cycle:
+    - `getEvalKitForCompartment` needs `compartments`
+    - `compartments` is created by `link()`
+    - `link()` needs `makeImportHook`
+    - `makeImportHook` needs `getEvalKitForCompartment`
+  */
+  const { compartment: entryCompartment, compartments } = link(compartmentMap, {
+    makeImportHook,
+    globals: globalThis,
+    // transforms,
```
naugtur, line 75: observation: if we had a reference to a record of modules from another bundle or chunk, all we need to do here is to pass them as `modules` and current app can import them.

packages/compartment-mapper/src/bundle.js
```diff
@@ -252,7 +295,7 @@ export const makeBundle = async (read, moduleLocation, options) => {
 'use strict';
 (() => {
   const functors = [
-${''.concat(...modules.map(m => m.bundlerKit.getFunctor()))}\
+${''.concat(modules.map(m => m.bundlerKit.getFunctor()).join(','))}\
```
naugtur, line 298: you don't need concat anymore - join returns an empty string for empty array

packages/compartment-mapper/src/bundle.js
```diff
@@ -298,6 +341,225 @@ ${''.concat(...modules.map(m => m.bundlerKit.getFunctorCall()))}\
...
 };
 
+function wrapFunctorInPrecompiledModule(functorSrc, compartmentName) {
+  const wrappedSrc = `() => (function(){
+  with (this.scopeTerminator) {
+  with (this.globalThis) {
+    return function() {
+      'use strict';
+      return (
+${functorSrc}
```
naugtur, line 351: I'm not sure when's the good time to do it, but we should probably check `functorSrc` parses as JavaScript.

```js
try {
  eval(`throw Error('valid');
  ${functorSrc}`);
} catch(e) {
 assert( e.message === 'valid')
}
```

It'd be the safest to do at runtime in case there's differences in parsing, but that's not something we want to do with end-users' CPUs so earlier in build process would have to suffice.

packages/compartment-mapper/test/test-bundle.js
naugtur: Or we could create a secure and insecure bundle and compare results.

legobeat: The first option sounds like the simplest way to get what we need?

naugtur: That requires 3 folders with package.json and index.js in each. That's why I thought of other suggestions to make the fixture smaller. It is though the closest thing to the actual subject of the test.

packages/compartment-mapper/src/archive.js
naugtur: The one place I interacted with one it was annoying to debug, but no other complaints form me. Especially here, where they're used in a loop not passed to Array.from or Object.fromEntries

packages/compartment-mapper/src/bundle.js
```diff
@@ -298,6 +341,225 @@ ${''.concat(...modules.map(m => m.bundlerKit.getFunctorCall()))}\
...
+    }
+    // other module types?
+  }
+
+  const bundleRuntimeLocation = new URL(
+    './bundle-runtime.js',
+    import.meta.url,
+  ).toString();
+  const runtimeBundle = evadeImportExpressionTest(
+    await makeBundle(read, bundleRuntimeLocation),
```
naugtur, line 416: This will not work - read can be anything. It only has access to the bundle-runtime.js if it's coming from node-powers.js

Temporary fix:
```suggestion
      await makeBundle(makeReadPowers({ fs }).read, bundleRuntimeLocation),
```

But I think ultimately it'd be best if we could read the runtime from the archive (backward compatibility wins?)

packages/compartment-mapper/src/archive.js
```diff
@@ -243,6 +255,44 @@ const captureSourceLocations = async (sources, captureSourceLocation) => {
...
+    compartments: archiveCompartments,
+  };
+
+  // Cross-check:
+  // We assert that we have constructed a valid compartment map, not because it
+  // might not be, but to ensure that the assertCompartmentMap function can
+  // accept all valid compartment maps.
+  assertCompartmentMap(archiveCompartmentMap);
+
+  return { archiveCompartmentMap, archiveSources, compartmentRenames };
```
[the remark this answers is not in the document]
kumavis: idk, it represents work performed and seemed relevant. can be removed and re-added as needed.

packages/compartment-mapper/src/bundle-runtime.js
```diff
@@ -0,0 +1,164 @@
...
+  });
+
+  function getCompartmentByName(name) {
+    let compartment = compartments[name];
+    if (compartment === undefined) {
+      compartment = new Compartment();
+      compartments[name] = compartment;
+    }
+    return compartment;
+  }
```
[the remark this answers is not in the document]
kumavis: this is a style / performance question I'll delegate to kris or someone

packages/compartment-mapper/src/bundle.js
kumavis: > This will not work - read can be anything. It only has access to the bundle-runtime.js if it's coming from node-powers.js
yeah you're right, we're expecting this to read our internal file

> Alternatively - importing the whole pre-bundled runtime as a function and running a toString on it might work when concatenated with the right invocation.

thats interesting, not sure i understand. are you suggesting pulling it out of the functor table?

packages/compartment-mapper/src/archive.js
```diff
@@ -206,27 +207,38 @@ const renameSources = (sources, compartmentRenames) => {
...
   for (const compartment of keys(sources).sort()) {
     const modules = sources[compartment];
     const compartmentLocation = resolveLocation(`${compartment}/`, 'file:///');
     for (const specifier of keys(modules).sort()) {
-      const { bytes, location } = modules[specifier];
+      const module = modules[specifier];
+      const { location } = module;
       if (location !== undefined) {
         const moduleLocation = resolveLocation(location, compartmentLocation);
         const path = new URL(moduleLocation).pathname.slice(1); // elide initial "/"
```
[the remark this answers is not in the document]
kumavis: not my code, not my word

packages/compartment-mapper/src/bundle-runtime.js
kumavis: this is in reference to a code splitting feature?

packages/compartment-mapper/src/archive.js
```diff
@@ -206,27 +207,38 @@ const renameSources = (sources, compartmentRenames) => {
 };
 
 /**
- * @param {ArchiveWriter} archive
  * @param {Sources} sources
  */
-const addSourcesToArchive = async (archive, sources) => {
+export const locationsForSources = function* locationsForSources(sources) {
```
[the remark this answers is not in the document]
kumavis: this is for generating module names not the modules themselves

packages/compartment-mapper/test/test-bundle.js
kumavis: opted to create an insecure bundle and compare behavior

packages/compartment-mapper/src/bundle.js
kumavis: consider :smiling_imp: 
```js
  eval(`throw Error('valid');  ${functorSrc}`);
```

with `functorSrc`:
```js
function Error () { /* attack payload */ }
```

browserify uses `throw 'STOP';`

packages/compartment-mapper/src/bundle.js
kumavis: fixed

packages/compartment-mapper/src/bundle.js
kumavis: - src is validated when creating a bundle by entrypoint, somewhere in `prepareToBundle`
-  aded a failing test showing `makeSecureBundleFromArchive` not validating invalid module sources

naugtur: Ok, that looks solid, the only remaining concern is someone finding a way to make unziping+parsing in the runtime host result in a different module. 
eg. we build an archive in Node and then run it in Hermes or some bespoke JS engine with an ASI implementation quirk or parsing bug that changes the AST resulting from the original text. That could be exploited fairly easily (after it's been found, which is the hard part) using the method I used when exploiting the comment in LavaMoat.

This is the only reason to consider the browserify trick at runtime. It depends on eval though and is therefore unusable for us. Unless we have a non-eval alternative the work I put into typing all this up is for nothing anyway ;)

packages/compartment-mapper/src/bundle-runtime.js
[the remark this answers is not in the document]
kriskowal: I don’t have a strong feeling one way or the other but maybe I can tap my friend @romanco to make a ruling based on his recent performance observations of a similar pattern in nested maps.

packages/compartment-mapper/src/bundle.js
```diff
@@ -298,6 +344,226 @@ ${''.concat(...modules.map(m => m.bundlerKit.getFunctorCall()))}\
...
+  );
+  return `{\n${lines.map(line => `  ${line}`).join(',\n')}\n};`;
+}
+
+/**
+ * @param {CompartmentMapDescriptor} compartmentMap
+ * @param {Sources} sources
+ * @returns {Promise<string>}
+ */
+export const makeSecureBundleFromAppContainer = async (
```
kriskowal, line 378: On the principle of importing-what-you-paid-for, this could be factored into `secure-bundle.js`.

packages/compartment-mapper/src/bundle.js
kriskowal: Have we censored the functorSrc at this point?

kriskowal (commented):
Partial review for today.

packages/compartment-mapper/src/archive.js
```diff
@@ -206,27 +207,38 @@ const renameSources = (sources, compartmentRenames) => {
...
+      const module = modules[specifier];
+      const { location } = module;
       if (location !== undefined) {
         const moduleLocation = resolveLocation(location, compartmentLocation);
-        const path = new URL(moduleLocation).pathname.slice(1); // elide initial "/"
-        if (bytes !== undefined) {
-          // eslint-disable-next-line no-await-in-loop
-          await archive.write(path, bytes);
-        }
+        const path = new URL(moduleLocation).pathname.slice(1); // skip initial "/"
```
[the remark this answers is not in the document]
kumavis: this line was not introduced in this PR, it was simply moved / refactored

line was introduced in 2020
https://github.com/endojs/endo/commit/406e24958007bac5de2100dbaa8d122720d979a4#diff-037d65c9f5e8bccd16957f7741d978790c098bc8dfa068eb64714ab92802b7ddR157

packages/compartment-mapper/src/bundle.js
```diff
@@ -5,13 +5,18 @@
...
+/** @typedef {import('./types.js').CompartmentMapDescriptor} CompartmentMapDescriptor */
 /** @typedef {import('./types.js').CompartmentSources} CompartmentSources */
 /** @typedef {import('./types.js').ReadFn} ReadFn */
 /** @typedef {import('./types.js').ModuleTransforms} ModuleTransforms */
 /** @typedef {import('./types.js').Sources} Sources */
 /** @typedef {import('./types.js').WriteFn} WriteFn */
 /** @typedef {import('./types.js').ArchiveOptions} ArchiveOptions */
 
+import fs from 'fs';
+import url from 'url';
```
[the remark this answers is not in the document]
kumavis: relevant comment 
https://github.com/endojs/endo/blob/16d0c2fbdd230bc7186880f0fc0f52ed5b033193/packages/compartment-mapper/src/bundle.js#L416-L419

unsure how to proceed - ask the bundle for two different read powers?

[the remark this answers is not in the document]
kriskowal: Probably. What are the two file systems in this case?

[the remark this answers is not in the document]
kumavis: 1. this repo's file system so we can bundle the secure bundler runtime
2. the user's file system for their app

packages/compartment-mapper/src/bundle.js
kumavis: we had not censored -- this has been fixed by running the "mandatory transforms"

Issue #1450, opened 2023-01-18
TS `never` confused in endo but not in agoric-sdk


erights:
The `throwLabeled` function defined at https://github.com/Agoric/agoric-sdk/blob/bfbb808fa9f4151844516ef0ca8af1a8b6bf98ae/packages/internal/src/utils.js#L103-L120 has no lint problems.

Currently I'm in progress in migrating the patterns abstractions from `@agoric/store` into a separate `@endo/patterns`. Because that code depends on `throwLabeled` I made a copy there. However, there is causes a lint error:

/Users/markmiller/src/ongithub/endojs/endo/packages/patterns/src/utils.js
  70:1  warning  JSDoc @returns declaration present but return expression not available in function  jsdoc/require-returns-check

This difference shows up in vscode as well. The left pane is my copy in my new `@endo/patterns` package. The right pane is the one in `@agoric/internal`. As you can see, locally there is no difference.
<img width="1530" alt="image" src="https://user-images.githubusercontent.com/273868/213059761-64877084-fd76-4592-8bb4-b4e364b0fe0b.png">

The lint error makes no sense to me. The function can only complete via a `throw` statement. TS is supposed to be happy with `@returns {never}` for that, which makes a lot of sense.

erights:
See https://github.com/endojs/endo/pull/1451

turadg:
That Eslint rule can get confused by TS types. How I've been working around that is a newljne after the JSdoc comment.

turadg:
Not sure. Maybe changes in eslint-plugin-jsdoc. https://github.com/endojs/endo/pull/1453 to bump to latest.

kriskowal:
It’s been a nearly year. Has this problem persisted?

erights:
Don't close it yet. Likely trivial PR coming up shortly.

erights:
At https://github.com/endojs/endo/pull/1930

PR #1451, opened 2023-01-18
feat(patterns): Start migrating patterns from @agoric/store to new @endo/patterns

erights:


packages/patterns/package.json
```diff
@@ -32,21 +28,25 @@
...
+    "eslint": "^7.32.0",
     "eslint-config-airbnb-base": "^14.0.0",
     "eslint-config-prettier": "^6.9.0",
     "eslint-plugin-eslint-comments": "^3.1.2",
-    "eslint-plugin-import": "^2.19.1",
+    "eslint-plugin-import": "^2.26.0",
     "eslint-plugin-prettier": "^3.4.1",
     "prettier": "^2.8.0",
-    "typescript": "~4.8.4"
+    "typescript": "~4.9.4"
```
erights, line 49: @kriskowal 
Thanks for https://www.youtube.com/watch?v=SCOYsE5G7Xk. It is how I found out how to update the version numbers, and is what clued me in to put this tilda back in afterwards.

packages/patterns/package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "@endo/patterns",
   "version": "0.1.0",
-  "private": null,
```
erights: Please look after commit 1 to see the changes I made after repackage. Did I delete exactly what I should have deleted?

erights:
Attn @Tyrosine22

erights:
Aside from the one mystery at #1450 , it was amazingly painless to get to CI green! Yeah, repackage!

FUDCo:
Since this is a cross-repo move, the git diffs make it entirely unclear which things are simply relocations and which are changes and/or additions.  I presume it's the latter that are of interest in reviewing?

erights:
> Since this is a cross-repo move, the git diffs make it entirely unclear which things are simply relocations and which are changes and/or additions. I presume it's the latter that are of interest in reviewing?

At this point, my intention is to make it as blindly imitative as possible while still having the resulting package be a good endo citizen. This is almost all 
   * deleting things not relevant to the pattern portion of the old `@agoric/store` package
   * repairing imports
   * copying some functions from `@agoric/internal` utils.js to the new `utils.js`, until they can find their proper home.

The attached .patch file is from diffing the two and manually removing everything that obviously does not need to be looked at. Look at in an editor that's friendly for looking at diffs.

[pattern-migration.patch](https://github.com/endojs/endo/files/10441557/pattern-migration.patch)

erights:
From diffing only the respective utils.js files. As you can see, it is all deletions aside from 
   * `identChecker` which is copied literally from `@agoric/assert`
   * import repair
   * #1450 repair.

[utils.patch](https://github.com/endojs/endo/files/10441623/utils.patch)

packages/patterns/package.json
kriskowal: All good.

kriskowal (approved):
Trusting that the code is largely the same as what integrated in Agoric SDK, this seems fine. So you have a PR coming to refactor Agoric SDK in place to use these?

erights:
> So you have a PR coming to refactor Agoric SDK in place to use these?

I will.

packages/patterns/package.json
```diff
@@ -0,0 +1,77 @@
+{
+  "name": "@endo/patterns",
+  "version": "0.1.0",
+  "description": "Description forthcoming.",
```
turadg, line 4: @erights looks like this is still needed

packages/patterns/package.json
```diff
@@ -0,0 +1,77 @@
...
+  "keywords": [],
+  "author": "Endo contributors",
+  "license": "Apache-2.0",
+  "homepage": "https://github.com/endojs/endo/tree/master/packages/patterns#readme",
```
turadg, line 8: the fragment only works for projects, not files in the tree.

should be: https://github.com/endojs/endo/blob/master/packages/patterns/README.md

Incidentally, it's empty and needs contents.

erights: See https://github.com/endojs/endo/pull/1480

(Separately, still needs contents)

erights: The empty READMEs for pass-style, patterns, and exo will be minimally fixed by https://github.com/endojs/endo/pull/1480 with initial text that's much better than nothing. A first step.

PR #1453, opened 2023-01-18
bump eslint and other deps

turadg:
- chore(deps): bump eslint-plugin-jsdoc to 39.6.6
- style: object for jsdoc/check-types
- chore(deps): bump eslint-config-jessie to 0.0.6
- chore(deps): rm unused eslint-plugin-jsx-a11y
- chore(deps): rm unused @ava/babel
- chore(deps): bump ava to 5.1.1
- build: require Node >=14.18
- chore: package.json name must be lowercase


turadg:
This is failing on [the test-async-hooks test against Node 14.17](https://github.com/endojs/endo/actions/runs/4027289695/jobs/6922832134) but passes on 14.18. @mhofman do we need to keep supporting 14.17?

[AVA now only supports >=14.19](https://github.com/avajs/ava/pull/3043/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f).

mhofman:
That's a weird error, but I think it's ok to drop. There was a change of behavior between 14.17 and 14.18, but it's ok to drop the test. Maybe we should find a way to execute these backwards compat tests in a way that relies less on scaffolding that changes over time.

michaelfig (approved):
TIL `object` instead of `Object` matters and is preferred (I thought I read the opposite once but don't recall where).

PR #1454, opened 2023-01-19
feat(ses): publish utils bundle

kumavis:
makes ses utils available which is useful when working with ses

### DIFF
--- a/packages/ses/package.json
+++ b/packages/ses/package.json
@@ -42,6 +42,7 @@
       "import": "./index.js",
       "require": "./dist/ses.cjs"
     },
+    "./tools.js": "./tools.js",
     "./package.json": "./package.json"
   },
   "scripts": {
@@ -85,6 +86,7 @@
     "dist",
     "index.d.ts",
     "index.js",
+    "tools.js",
     "src"
   ],
   "private": false,
--- a/packages/ses/package.json.md
+++ b/packages/ses/package.json.md
@@ -115,6 +115,10 @@ is deprecated.
 
 The value is the same as for `"."` above.
 
+## "./tools.js": ...
+
+Exposes tools that are useful for creating hardened JavaScript environments that move certain responsibilities from `eval` to generated code.
+
 ## "./package.json": "./package.json"
 
 Tools like Svelte need to access the `package.json` of every package in an
--- a/packages/ses/src/sloppy-globals-scope-terminator.js
+++ b/packages/ses/src/sloppy-globals-scope-terminator.js
@@ -51,3 +51,4 @@ export const createSloppyGlobalsScopeTerminator = globalObject => {
 
   return sloppyGlobalsScopeTerminator;
 };
+freeze(createSloppyGlobalsScopeTerminator);
--- a/packages/ses/tools.js
+++ b/packages/ses/tools.js
@@ -1,0 +1,25 @@
+// Copyright (C) 2018 Agoric
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { freeze } from './src/commons.js';
+import { transforms } from './src/transforms.js';
+import { strictScopeTerminator } from './src/strict-scope-terminator.js';
+import { createSloppyGlobalsScopeTerminator } from './src/sloppy-globals-scope-terminator.js';
+
+export { transforms };
+
+export const scopeTerminators = freeze({
+  strictScopeTerminator,
+  createSloppyGlobalsScopeTerminator,
+});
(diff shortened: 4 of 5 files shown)


packages/ses/package.json
```diff
@@ -42,6 +42,10 @@
       "import": "./index.js",
       "require": "./dist/ses.cjs"
     },
+    "./util": {
+      "import": "./dist/util.js",
+      "require": "./dist/util.cjs"
+    },
```
kriskowal, line 48: I don’t think we have to bundle the utils, but because some tools don’t recognize the `"exports"` field, we do need the key and value to match. We also have an emerging precedent for a file like this having the name `./tools.js` (`@endo/pass-style` will have one soon if it hasn’t landed yet). So my overall recommendation is for this to be:

```suggestion
    "./tools.js": "./tools.js",
```

And just to rename `utils.js` to `tools.js`.

packages/ses/scripts/bundle.js
```diff
@@ -39,6 +39,30 @@ const main = async () => {
...
+  const { code: terse } = terser.minify(bundle, {
+    mangle: false,
+    keep_classnames: true,
+  });
+
+  console.log(`Util Bundle size: ${bundle.length} bytes`);
+  console.log(`Util Minified bundle size: ${terse.length} bytes`);
+
+  await fs.promises.mkdir('dist', { recursive: true });
+  // we only write out cjs bundles due to compartment-mapper bundler limitation
```
kriskowal, line 56: I don’t understand this limitation. Compartment mapper should be able to slurp up the original ESM.

kriskowal (changes requested):
Let me know if these simplifications are intractable.

packages/ses/scripts/bundle.js
kumavis: it does handle consuming the esm just fine but bundles are not currently able to export as esm. if you evaluate the bundle, it will return the namespace, but will not create export statements. we want to be able to import the utilities normally. we're able to do it in commonjs by a sort of hack. this will result in the bundle importable as a commonjs module  https://github.com/endojs/endo/blob/558a3c3d8d3a8ae7518ed1dbb9e23e7c0fb8e8c0/packages/ses/util.js#L26-L31

kumavis: if we dont need to bundle, we dont need this workaround.

kumavis:
we determined bundling was a good way to ensure that we did not expose ses-shim internals

packages/ses/util.js
```diff
@@ -0,0 +1,31 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as transforms from './src/transforms.js';
+
+import { strictScopeTerminator } from './src/strict-scope-terminator.js';
+import { createSloppyGlobalsScopeTerminator } from './src/sloppy-globals-scope-terminator.js';
+
+export { transforms };
+export const scopeTermainators = {
```
michaelfig, line 21:
```suggestion
export const scopeTerminators = {
```

packages/ses/scripts/bundle.js
michaelfig: Can you at least file an issue for how we could fix the inability to produce ESM exports?  Then, please reference it here (and other places that would ideally be unnecessary) in a comment as `TODO: solve this workaround by fixing #XXXX`.

kumavis:
while we moved away from bundles for this, I created an issue for the "bundles with esm exports" feature https://github.com/endojs/endo/issues/1460

packages/ses/package.json.md
```diff
@@ -115,6 +115,10 @@ is deprecated.
 
 The value is the same as for `"."` above.
 
+## "./tools.js": ...
+
+Exposes some utilities that are convenient when working with SES.
```
kriskowal, line 120:
```suggestion
Exposes tools that are useful for creating hardened JavaScript environments that move certain responsibilities from `eval` to generated code.
```

packages/ses/src/sloppy-globals-scope-terminator.js
```diff
@@ -18,7 +18,7 @@ import {
  * to prevent access to "start compartment globals". When "sloppyGlobalsMode"
  * is true, the Proxy will perform sets on the "globalObject".
  */
-export const createSloppyGlobalsScopeTerminator = globalObject => {
+export const createSloppyGlobalsScopeTerminator = freeze(globalObject => {
```
kriskowal, line 21: This may interfere with certain debuggers’ ability to infer the name of the function. That can be addressed by freezing on the subsequent line.

packages/ses/tools.js
```diff
@@ -0,0 +1,26 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { freeze } from './src/commons.js';
+import * as transforms from './src/transforms.js';
+import { strictScopeTerminator } from './src/strict-scope-terminator.js';
+import { createSloppyGlobalsScopeTerminator } from './src/sloppy-globals-scope-terminator.js';
+
+freeze(transforms);
```
kriskowal, line 20: This `freeze` can occur in the originating module.

kriskowal (approved):

naugtur (approved):
LGTM

packages/ses/tools.js
kumavis: created a new `transforms` export

PR #1455, opened 2023-01-20
fix: make deeplyFulfilled an async function

erights:
I already manually checked all uses in the endo and agoric-sdk repository. It looks like this change will be harmless. The advantage (discussed in private chat) is that it normalized the outcome of `deeplyFulfilled` to always be a promise, even if
   * the argument was already deeply fulfilled, i.e., a Passable without promises
   * an error needs to be reported, even if the error is immediate, e.g., the argument is not Passable

This normalization of the outcome, especially for the error case, should reduce the case analysis burden on callers.

### DIFF
--- a/packages/marshal/src/deeplyFulfilled.js
+++ b/packages/marshal/src/deeplyFulfilled.js
@@ -34,7 +34,7 @@ const { fromEntries } = Object;
  * defines the "key" notion in question.
  *
  * // TODO: That higher level is in the process of being migrated from
- * // `@agoric/store` to `@endo/passterns`. Once that is far enough along,
+ * // `@agoric/store` to `@endo/patterns`. Once that is far enough along,
  * // revise the above comment to match.
  * // See https://github.com/endojs/endo/pull/1451
  *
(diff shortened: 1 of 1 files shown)


erights:
See https://github.com/Agoric/agoric-sdk/pull/6816

dckc (approved):
This would have saved me some head-scratching (though I did learn something in the process).

packages/marshal/src/deeplyFulfilled.js
```diff
@@ -24,15 +24,24 @@ const { fromEntries } = Object;
...
+ * Passable, or if any of the transitive promises fulfill to something
+ * that is not Passable, then the returned promise rejects.
+ *
+ * If `val` or its parts are non-key Passables only *because* they contains
  * promises, the deeply fulfilled forms of val or its parts may be keys. This
  * is for the higher "store" level of abstraction to determine, because it
  * defines the "key" notion in question.
  *
+ * // TODO: That higher level is in the process of being migrated from
+ * // `@agoric/store` to `@endo/passterns`. Once that is far enough along,
```
kriskowal, line 37: passterns / patterns + pass-style?

kriskowal (approved):

packages/marshal/src/deeplyFulfilled.js
erights: Fixed, thanks.

PR #1456, opened 2023-01-20
Fix SECURITY.md vulnerability rewards program link

leotm:
`hackerone.com/agoric` -> `https://hackerone.com/agoric` so it links correctly from GitHub


kriskowal:
Thank you for this submission. Since `SECURITY.md` exists in every package, I will need to follow-up on this change to synchronize all of them and also the scaffold for new packages.

kriskowal:
Following thru at https://github.com/endojs/endo/pull/1471

PR #1459, opened 2023-01-22
feat(exo): start migrating exo from @agoric/store

erights:
Migrated the exo support from the `@agoric/store` package. Changed to depend on `@endo/patterns` which was previously migrated from `@agoric/store`.

This directly supports only the heap implementations of exo objects and exo classes, but also provides the common implementation machinery used by `@agoric/vat-data` (via `@agoric/swing-set`) to implement virtual and durable exos. See doc/exo-taxonomy.md in this PR for these relationships.


erights:
Reviewers, I added a commit that renamed from *HeapExo* to *Exo*, in coordination with https://github.com/Agoric/agoric-sdk/pull/6838 . See explanation there.

FUDCo (approved):
Ya shoor.

PR #1461, opened 2023-01-23
feat(ses): allow module-instance to use a precompiledFunctor

kumavis:
the "precompiled functor" is a function potentially present on the `staticModuleRecord`. If present, it is used as a replacement for string evaluation of the module code, useful in environments where eval is not available. Confinement of the functor is the responsibility of the creator of the precompiled functor and `staticModuleRecord`.

`ThirdPartyModuleInstances` can already do this by implementing a custom `execute` fn https://github.com/endojs/endo/blob/7de3ef8e9d5bcecc09545f5782eeb647a742ba51/packages/ses/src/module-instance.js#L99-L103

This is intended to be used by the `compartment-mapper`'s secureBundler "no eval" output format

### DIFF
--- a/packages/ses/src/module-instance.js
+++ b/packages/ses/src/module-instance.js
@@ -136,6 +136,7 @@ export const makeModuleInstance = (
     __liveExportMap__: liveExportMap = {},
     __reexportMap__: reexportMap = {},
     __needsImportMeta__: needsImportMeta = false,
+    __syncModuleFunctor__,
   } = staticModuleRecord;
 
   const compartmentFields = weakmapGet(privateFields, compartment);
@@ -444,11 +445,16 @@ export const makeModuleInstance = (
     activate();
   }
 
-  let optFunctor = compartmentEvaluate(compartmentFields, functorSource, {
-    globalObject: compartment.globalThis,
-    transforms: __shimTransforms__,
-    __moduleShimLexicals__: moduleLexicals,
-  });
+  let optFunctor;
+  if (__syncModuleFunctor__ !== undefined) {
+    optFunctor = __syncModuleFunctor__;
+  } else {
+    optFunctor = compartmentEvaluate(compartmentFields, functorSource, {
+      globalObject: compartment.globalThis,
+      transforms: __shimTransforms__,
+      __moduleShimLexicals__: moduleLexicals,
+    });
+  }
   let didThrow = false;
   let thrownError;
   function execute() {
--- a/packages/ses/NEWS.md
+++ b/packages/ses/NEWS.md
@@ -1,5 +1,10 @@
 User-visible changes in SES:
 
+# next
+
+- Introduces the `__syncModuleFunctor__` property of static module record
+  to replace evauluation of `__syncModuleProgram__` for environments without eval.
+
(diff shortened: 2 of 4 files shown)


packages/ses/src/module-instance.js
```diff
@@ -136,6 +136,7 @@ export const makeModuleInstance = (
     __liveExportMap__: liveExportMap = {},
     __reexportMap__: reexportMap = {},
     __needsImportMeta__: needsImportMeta = false,
+    __precompiledFunctor__,
```
kriskowal, line 139: The terms of `__syncModuleProgram__` mean:

- `sync` because it doesn’t support top-level-await, but someday `async` should
- `module` because it was a [[Module]]
- `program` because it is a [[Script]]
- `__` because it isn’t going to work without the shim

I’d like to consider naming this `__syncModuleFunctor__`
- `sync` because `async` will support top-level-await maybe someday
- `module` because reasons
- `functor` instead of `program` because the grammar production is irrelevant, the calling convention is.

packages/ses/src/module-instance.js
```diff
@@ -444,11 +445,16 @@ export const makeModuleInstance = (
...
   }
 
-  let optFunctor = compartmentEvaluate(compartmentFields, functorSource, {
-    globalObject: compartment.globalThis,
-    transforms: __shimTransforms__,
-    __moduleShimLexicals__: moduleLexicals,
-  });
+  let optFunctor;
+  if (__precompiledFunctor__ !== undefined) {
+    optFunctor = __precompiledFunctor__;
```
kriskowal, line 450: Our theory is that this is safe because guest code can only create confined functors and host code is necessarily responsible for participating in the enforcement of confinement. cc @erights because this is an area where we’re likely to miss something.

The other half of the theory is that this violates Hyrum’s law but only in a way that we’re already breaking Hyrum’s law. Code can be written to depend on this behavior that will necessarily break on a standard implementation of Compartment sans shim. That doesn’t preclude the possibility of the shim standing atop a genuine implementation and providing a backward compatibility layer, though.

kriskowal (dismissed):
This needs to be documented in `README.md` for our future selves and a note in `NEWS.md` would be nice. I will resign as a reviewer when my feedback is addressed since I consider @erights feedback necessary for this change.

packages/ses/src/module-instance.js
kriskowal: And let’s capture this reasoning in a comment for posterity.

mhofman: I suppose the main thing that needs to be evaluated is whether the argument passed to the functor reveals anything the code running in the compartment may not have had access to otherwise?

kriskowal: I believe that’s the nature of the question.

This path skips `evaluate` so I do not believe we have concerns about lifting text to behavior safely. The functor receives an `imports` object from the module system which grants the module the ability to get and set values of its environment. If that provides excess authority, it does so in a way that’s already achievable by fabricating a static module record duck type for a child compartment.

erights (approved):
Pleasingly simple! I understand the security argument and it is solid. LGTM!

PR #1463, opened 2023-01-25
docs: missed documentation suggestion

erights:
The change I made to the agoric-sdk copy of this package also has a doc fix in response to https://github.com/Agoric/agoric-sdk/pull/6838#discussion_r1085515755 that I foolishly omitted from https://github.com/endojs/endo/pull/1459 . This PR repairs that, bringing the copies back into alignment.


turadg (approved):
Commit change is good. Not a code "fix" (which bumps the version), just a "docs" change. I changed the PR title so you can just squash merge.

PR #1464, opened 2023-01-25
Rename 'fit' to 'mustMatch'

FUDCo:
Per https://github.com/Agoric/agoric-sdk/issues/6844


turadg:
What's the motivation?

If we're going to make it so long, why not "assertMatch"?

kriskowal:
I’ve seen and like the `mustX` convention for functions whose behavior is to throw if `X` returns false. I’ve seen it used like `mustFind` or `mustQuerySelector`. I first saw the pattern used in Go, then in JavaScript among friends who also wrote Go. I don’t object to `assertX`.

FUDCo:
I think this is one problem with cross-repo changes.  The relevant issue is https://github.com/Agoric/agoric-sdk/issues/6844

turadg:
That issue provides the context I needed. I've put it in the PR description.

turadg (approved):
Thanks @kriskowal for the explanation of the "must" naming pattern.

PR #1465, opened 2023-01-25
feat(ses-ava): Support the full ava API

### LINKED ISSUE #1235: Adopt simplified ses-ava

gibson042:
Fixes #1235
Ref https://github.com/Agoric/agoric-sdk/pull/5774


packages/ses-ava/src/ses-ava-test.js
```diff
@@ -23,21 +14,48 @@ const isPromise = maybePromise =>
...
+ * from directly calling `func(...args)` but will be equivalent enough for most
+ * purposes.
+ *
+ * TODO This function is useful independent of Ava, so consider moving it
+ * somewhere and exporting it for general reuse.
+ *
+ * @param {(...unknown) => unknown} func
+ * @param {unknown[]} args
+ * @param {string} name
+ * @param {Logger=} logError
```
erights, line 44:
```suggestion
 * @param {Logger} [logError]
```
This has become our std way of saying optional at the jsdoc/ts level.

packages/ses-ava/src/ses-ava-test.js
```diff
@@ -23,21 +14,48 @@ const isPromise = maybePromise =>
...
  */
-const logErrorFirst = (func, args, name, logger = defaultLogger) => {
+const logErrorFirst = (func, args, name, logError = console.error) => {
```
erights, line 46:
```suggestion
const logErrorFirst = (func, args, name, logError = (...args) => console.error(...args)) => {
```
just in case on a platform in which `console.error` is this-sensitive.

packages/ses-ava/src/ses-ava-test.js
```diff
@@ -100,27 +161,27 @@ const wrapTester = (testerFunc, logger = defaultLogger) => {
...
-    wrapTester(avaTest, logger)
-  );
-  for (const methodName of testerMethodsWhitelist) {
-    if (methodName in avaTest) {
-      /** @type {TesterFunc} */
-      const testerMethod = (title, implFunc, ...otherArgs) =>
-        avaTest[methodName](title, implFunc, ...otherArgs);
-      testerWrapper[methodName] = wrapTester(testerMethod);
-    }
+const wrapTest = (avaTest, logError = console.error) => {
```
erights, line 172:
```suggestion
const wrapTest = (avaTest, logError = (...args) => console.error(...args)) => {
```

erights (commented):
Just comments so far. Review not yet finished.

packages/ses-ava/src/ses-ava-test.js
turadg: The only platform Ava runs on is Node. They don't even intend to support Deno and if they did it woudl be using the Node-compatibility API https://github.com/avajs/ava/discussions/2743

gibson042:
Good suggestions; done.

packages/ses-ava/src/ses-ava-test.js
```diff
@@ -100,27 +163,27 @@ const wrapTester = (testerFunc, logger = defaultLogger) => {
...
+  harden(sesAvaTest);
+  return sesAvaTest;
 };
-// harden(wrapTest);
+// Most instantiations of this module are pre-lockdown,
+// in which case harden is not ready here.
+try {
+  harden(wrapTest);
+} catch (err) {
+  freeze(wrapTest);
```
turadg, line 187: If `freeze` is sufficient, why bother with `harden`?

If we need `harden` before lockdown, couldn't this just have its own?
```
const harden = makeHardener();
```

packages/ses-ava/src/ses-ava-test.js
```diff
@@ -23,21 +14,50 @@ const isPromise = maybePromise =>
...
- * @type {LogCallError}
+ * @typedef {(...args: unknown[]) => void} Logger
  */
-const logErrorFirst = (func, args, name, logger = defaultLogger) => {
+
+const defaultLogError = (...args) => console.error(...args);
+
+/**
+ * Calls `func(...args)` passing back approximately its outcome, but first
+ * logging any erroneous outcome to the `logger`, which defaults to
```
turadg, line 24: since this only logs to one place, I think the earlier `logger` name is more appropriate. That it's logging an error is implied by it being the `logger` param of `logErrorFirst`.

Regardless,
1. the name here in the docs should match the param
2. the default is `console.error` not `console.log`

packages/ses-ava/src/ses-ava-test.js
```diff
@@ -23,21 +14,50 @@ const isPromise = maybePromise =>
...
+ *         even if the fulfillment was an error.
+ *       * If the first promise rejects, log the rejection reason and then
+ *         reject the returned promise with the same reason.
+ *
+ * The delayed rejection of the returned promise is an observable difference
+ * from directly calling `func(...args)` but will be equivalent enough for most
+ * purposes.
+ *
+ * TODO This function is useful independent of Ava, so consider moving it
+ * somewhere and exporting it for general reuse.
```
turadg, line 41: Now is a good time to make a decision on this. Should it be extracted? If not, let's remove this TODO.

packages/ses-ava/src/ses-ava-test.js
```diff
@@ -46,45 +66,88 @@ const logErrorFirst = (func, args, name, logger = defaultLogger) => {
...
   'only',
-  'skip',
 ];
 
 /**
- * @param {TesterFunc} testerFunc
- * @param {Logger} [logger]
- * @returns {TesterFunc} Not yet frozen!
+ * @callback BaseImplFunc
+ * This is the function that invariably starts `t => {`.
```
turadg, line 81: sometimes it starts `async t => {`

packages/ses-ava/src/ses-ava-test.js
```diff
@@ -46,45 +66,88 @@ const logErrorFirst = (func, args, name, logger = defaultLogger) => {
...
+ * Ava's types call this `Implementation`, but that's just too confusing.
```
turadg, line 82: It wouldn't be if we imported it from Ava. That would be better than redefining it. I'll propose a change below.

packages/ses-ava/src/ses-ava-test.js
```diff
@@ -46,45 +66,88 @@ const logErrorFirst = (func, args, name, logger = defaultLogger) => {
...
+ *
+ * @callback TesterFunc
+ * @param {string} title
+ * @param {ImplFunc} [implFunc]
+ * @returns {void}
  */
-const wrapTester = (testerFunc, logger = defaultLogger) => {
+
+/**
+ * @template {TesterFunc} T
```
turadg, line 96: I think with this we could erase the three custom typedefs above. We shouldn't have to define any types for an Ava wrapper; they should all match Ava's.
```suggestion
 * @template {import('ava').TestFn} T
```

packages/ses-ava/src/ses-ava-test.js
```diff
@@ -46,45 +66,88 @@ const logErrorFirst = (func, args, name, logger = defaultLogger) => {
...
+  // re-use other properties (e.g. `.always`)
+  // https://github.com/endojs/endo/issues/647#issuecomment-809010961
+  Object.assign(augmented, testerFunc);
+  // @ts-expect-error cast
+  return augmented;
 };
 
 /**
- * The ava `test` function takes a callback argument of the form
+ * The Ava `test` function takes a callback argument of the form
```
turadg, line 145: +1 to proper noun. Same two lines below.

turadg (commented):
Glad to see this in Endo. I don't think it will complete #1235 as written because that has an acceptance criterion,
> module imports either have side-effects OR bindings

That maybe deserves its own issue. For hygiene at least let's strike that from the issue claiming to be fixed by this PR.

packages/ses-ava/src/ses-ava-test.js
gibson042: I don't think so, because that would traverse and mutate intrinsics in precisely the way this is trying to avoid.

packages/ses-ava/src/ses-ava-test.js
gibson042: Fair enough, I think we can drop the TODO and export it if/when there is need.

packages/ses-ava/src/ses-ava-test.js
gibson042: Ha, didn't realize that was removed and then subsequently restored! Fixed.

packages/ses-ava/src/ses-ava-test.js
gibson042: Love it! Done.

packages/ses-ava/src/ses-ava-test.js
gibson042: Except that the proper noun is "AVA": https://github.com/avajs/ava/tree/main#how-is-the-name-written-and-pronounced

I updated to consistently use lowercase, with the (somewhat weak) rationale being reference to the npm package rather than the project.

packages/ses-ava/src/ses-ava-test.js
turadg: Given that a call to `wrapTest` has `harden`, I don't understand what is gained by `harden` here vs `freeze`. Not a blocker but please consider simplifying this to `freeze` (if appropriate) or leaving a comment on what will be different if only `freeze` is run.

turadg (approved):
LGTM. Thanks for the issue hygiene.

Looking forward to dropping the patch in agoric-sdk!

packages/ses-ava/src/ses-ava-test.js
gibson042: Any thoughts, @erights? This stems from https://github.com/Agoric/agoric-sdk/pull/5774#discussion_r922724162 .

gibson042: I've tentatively opted for unconditional `freeze` with preceding justification.

erights:
@gibson042 I have not yet looked at this, nor reviewed my objections at the old conversation at https://github.com/Agoric/agoric-sdk/pull/5774#discussion_r922722599 . Have you look at that conversation, and are those objections properly avoided in this PR?

I will still do a real review if you want to wait for that. But answering my question here would help me get oriented. Thanks.

packages/ses-ava/src/ses-ava-test.js
erights: @michaelfig @kriskowal , are you two were exploring a pre-lockdown `harden` that just registers things to be hardened once `lockdown` is called? I seem to recall that on some thread somewhere. What's the status of that?

In the absence of a pre-lockdown registering `harden`, I have taken to manually freezing everything that a post-lockdown `harden` would have frozen. Another good thing about arrow functions is that is only the function itself. (Also true for concise methods, but not for `function` functions nor classes.)

So yes, if `wrapTest` is an arrow function, then just a `freeze` is enough.

> If `freeze` is sufficient, why bother with `harden`?

- fragility. Were `wrapTest` changed from an arrow function to a `function` function. Or were more properties added to it before freezing/hardening, `freeze` would fail to freeze more things than `harden` would fail to freeze.
- memoizing. `harden` remembers what's hardened so it can avoid redundant work on future `harden`s.

gibson042: @erights keeping in mind that this is code to be imported by test files, is that an argument for preferring `try { harden(wrapTest) }` over an unconditional `freeze(wrapTest)`?

gibson042:
@erights I believe your concerns are addressed by the use of ava's `TestFn` type and the omission of "skip" and "todo" from our list of properties to wrap (which is fine since they don't have any assertions to produce opaque log output anyway).

packages/ses-ava/src/ses-ava-test.js
erights: I don't see why. Why would use by test files suggest this?

turadg: I think Richard's asking if you object to the use of `freeze` instead of `harden` for the `wrapTest` function. I.e. are you requesting changes to this PR.

I asked a sort of academic question about the difference between `freeze` and `harden` and you contrasted them, but it wasn't clear whether you assert that those differences imply the need for the (conditional) use of `harden` here.

I'll make the case that they don't. Memoization is irrelevant because this happens once at module load. "Fragility" is only a risk that a change could be made to wrapTest that would freeze fewer things. As Richard points out, this is a testing util. Test environments are always more malleable than production. `wrapTest` is meant to test code that runs in SES, but we don't need to defend against mutations to wrapTest because it's our tool. If we don't trust our own tests then we've got bigger problems :)  Is this about a coding discipline or to defend against some real risk? Perhaps I'm just not being creative enough… is the attack that some package gets into our import graph and itself imports `ses-ava` and mutates wrapTest to deceive our other tests into satisfying some important security property when they don't?

erights: @turadg 100%. Sorry I was unclear. I was indeed saying three things:
    * approving the use of `freeze` rather than `harden` specifically here in this PR
    * that I see no need for a condition harden or a harden in a try/catch here in this PR
    * why, in general, we should avoid freeze and use harden instead. Where that observation was not intended to apply to this case.

packages/ses-ava/src/ses-ava-test.js
```diff
@@ -100,27 +145,26 @@ const wrapTester = (testerFunc, logger = defaultLogger) => {
...
-  for (const methodName of testerMethodsWhitelist) {
-    if (methodName in avaTest) {
-      /** @type {TesterFunc} */
-      const testerMethod = (title, implFunc, ...otherArgs) =>
-        avaTest[methodName](title, implFunc, ...otherArgs);
-      testerWrapper[methodName] = wrapTester(testerMethod);
-    }
+  const sesAvaTest = augmentLogging(avaTest, logger);
+  for (const methodName of overrideList) {
+    sesAvaTest[methodName] = augmentLogging(avaTest[methodName], logger);
```
erights, line 160: While this assignment is my mistake and fixing it is not relevant to the purpose of this PR, please fix it anyway as a drive by, while you can still fit it into this PR. We should generally define properties using define semantics rather assignment semantics.

erights (approved):
> @erights I believe your concerns are addressed by the use of ava's `TestFn` type and the omission of "skip" and "todo" from our list of properties to wrap (which is fine since they don't have any assertions to produce opaque log output anyway).

Agreed, so LGTM.

However, I did look and ava;s `TestFn` is not a perfect fit either, as it says that a testFn must have all the enumerated TestFn-like properties. However, rather than stating a function-type-only for what these TestFn-like properties have in common, they define an entire new type for each one. Their type definitions are not DRY, and none of them perfectly fit what you're trying to say here.

However, `TestFn` is close enough, and is the best you can do importing one of their types. Even with the mismatch, I agree it is better to do that than to define our own function type.

No change suggested re `TestFn`.

PR #1466, opened 2023-01-25
fix(static-module-record): ESM compatibility

mhofman:
There is no `default` export on `@babel/types`. Not sure how this worked in the first place for Node.js but it makes `esm` happy for the few places that still use RESM.

See https://github.com/babel/babel/pull/14674 and https://github.com/babel/babel/pull/13414

### DIFF
--- a/packages/static-module-record/src/transformSource.js
+++ b/packages/static-module-record/src/transformSource.js
@@ -1,7 +1,7 @@
 import * as babelParser from '@babel/parser';
 import babelGenerate from '@agoric/babel-generator';
 import babelTraverse from '@babel/traverse';
-import babelTypes from '@babel/types';
+import * as babelTypes from '@babel/types';
 
 const parseBabel = babelParser.default
   ? babelParser.default.parse
--- a/packages/static-module-record/test/fixtures/small.js
+++ b/packages/static-module-record/test/fixtures/small.js
@@ -3,7 +3,7 @@ console.error("This is a code sample for trying out babel transforms, it's not m
 import * as babelParser from '@babel/parser';
 import babelGenerate from '@agoric/babel-generator';
 import babelTraverse from '@babel/traverse';
-import babelTypes from '@babel/types';
+import * as babelTypes from '@babel/types';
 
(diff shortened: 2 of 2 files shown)


packages/static-module-record/src/transformSource.js
```diff
@@ -1,7 +1,7 @@
 import * as babelParser from '@babel/parser';
 import babelGenerate from '@agoric/babel-generator';
 import babelTraverse from '@babel/traverse';
-import babelTypes from '@babel/types';
+import * as babelTypes from '@babel/types';
 
 const parseBabel = babelParser.default
   ? babelParser.default.parse
```
kriskowal, line 7: We probably need to cargo this pattern for `babelTypes` for it to work everywhere `parseBabel` works. It’s possible we’re not testing that far back anymore.

kriskowal (approved):

packages/static-module-record/src/transformSource.js
mhofman: I'm not sure I follow what you mean here. Would bumping the minimum required version of `@babel/types` to `^7.18.6` be sufficient? That's when babel fixed some ESM compat stuff I believe

kriskowal: I actually do not recall what combination of platforms and Babel versions required each these variations. I do recall that neither `import * as name` nor `import name` were sufficient for a Babel on ESM and Node.js and that the behavior varied between Node.js 12 and 14. If upgrading Babel makes a single import pattern work on every platform we currently support, that’s fine by me. We have presumably moved on from 14 so maybe the reverse advice applies: maybe we can get rid of the hack on line 6.

mhofman: Given we're not sure about the coverage of the `parseBabel` backwards compat logic here, I don't feel like removing it.

I don't think we need to have a `const actualBabelTypes = babelTypes.default || babelTypes` equivalent since tests pass here and in `agoric-sdk`, but we can always add later if we get breakage report.

PR #1468, opened 2023-01-26
feat(eventual-send): Harden HandledPromise modulo intrinsic exit points

gibson042:
Prompted by conversation at https://github.com/endojs/endo/pull/1465#discussion_r1086887864 .

### DIFF
--- a/packages/eventual-send/src/handled-promise.js
+++ b/packages/eventual-send/src/handled-promise.js
@@ -569,7 +569,7 @@ export const makeHandledPromise = () => {
 
   // We're a vetted shim which runs before `lockdown` allows
   // `harden(HandledPromise)` to function, but single-level `freeze` is a
-  // suitable replacement because all objects reachable from the result are
+  // suitable replacement because all mutable objects reachable afterwards are
   // intrinsics hardened by lockdown.
   freeze(HandledPromise);
   for (const key of ownKeys(HandledPromise)) {
--- a/packages/eventual-send/src/handled-promise.js
+++ b/packages/eventual-send/src/handled-promise.js
@@ -564,11 +564,8 @@ export const makeHandledPromise = () => {
 
   // FIXME: This is really ugly to bypass the type system, but it will be better
   // once we use Promise.delegated and don't have any [[Constructor]] behaviours.
-  /** @type {unknown} */
-  const unknownBaseHandledPromise = baseHandledPromise;
-  HandledPromise = /** @type {typeof HandledPromise} */ (
-    unknownBaseHandledPromise
-  );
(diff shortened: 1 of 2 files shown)


packages/eventual-send/src/handled-promise.js
```diff
@@ -563,14 +564,19 @@ export const makeHandledPromise = () => {
...
-
-  // We cannot harden(HandledPromise) because we're a vetted shim which
-  // runs before lockdown() allows harden to function.  In that case,
-  // though, globalThis.HandledPromise will be hardened after lockdown.
+  // @ts-expect-error cast
+  HandledPromise = baseHandledPromise;
+
+  // We're a vetted shim which runs before `lockdown` allows
+  // `harden(HandledPromise)` to function, but single-level `freeze` is a
+  // suitable replacement because all objects reachable from the result are
```
erights, line 572:
```suggestion
  // suitable replacement because all objects reachable from `HandledPromise` are
```
may as well be explicit.

packages/eventual-send/src/handled-promise.js
```diff
@@ -563,14 +564,19 @@ export const makeHandledPromise = () => {
...
+  // intrinsics hardened by lockdown.
+  freeze(HandledPromise);
+  for (const key of ownKeys(HandledPromise)) {
+    // prototype is the intrinsic Promise.prototype to be hardened by lockdown.
+    if (key !== 'prototype') {
+      freeze(HandledPromise[key]);
```
erights, line 578:
```suggestion
      // These will be the staticMethods added above.
      freeze(HandledPromise[key]);
```
I had to look to see that these properties were each things that should be frozen but not transitively frozen. What do you think about instead freezing `staticMethods` before adding them to `baseHandledPromise`?

erights (approved):
LGTM

packages/eventual-send/src/handled-promise.js
gibson042: My first thought was also `freeze(staticMethods)`, but that doesn't work because `Object.freeze` is shallow and so the methods theirselves remain mutable. We could instead change the _definitions_ there (e.g., `get(target, prop) { … }` → `get: freeze((target, prop) => { … }),`, but this after-the-fact freezing with preceding explanation and test coverage seemed like a better approach.

erights: I was just thinking of you'd write the same loop, but over the staticMethods before making them static properties of HandledPromise. Not a big deal either way.

In any case, please *do not* change them into arrow functions. Because this is still supposed to be a shim of something we propose, the methods should remain written in concise method syntax.

PR #1471, opened 2023-01-27
fix(cli): Fix hash, map, and hash-archive commands

kriskowal:
Either these never worked or something changed in an upgrade of Commander (the command line flag parser we use) between when these commands last worked and present day.


kriskowal:
Please merge if you approve, @michaelfig. This defect spotted by @dckc.

michaelfig (approved):

PR #1474, opened 2023-01-28
refactor(exo): split exo tools

erights:
The `@endo/exo` package needs to separate its normal exo makers from the esoteric interface tools (like `defendPrototype`) for use by others (like `@agoric/vat-data`) to make exo makers.

We use the package.json "exports" feature to explicitly export just these two entry points.

See https://github.com/Agoric/agoric-sdk/pull/6865 for the tracking PR until we can remove the redundancy.


FUDCo (approved):
Exo echo.

PR #1480, opened 2023-01-29
docs: fix readme urls

erights:
See https://github.com/endojs/endo/pull/1451#pullrequestreview-1263981440

@kriskowal does this reflect a problem in repackage.sh ?

### DIFF
--- a/packages/exo/package.json
+++ b/packages/exo/package.json
@@ -5,7 +5,7 @@
   "keywords": [],
   "author": "Endo contributors",
   "license": "Apache-2.0",
-  "homepage": "https://github.com/endojs/endo/tree/master/packages/exo#readme",
+  "homepage": "https://github.com/endojs/endo/blob/master/packages/exo/README.md",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/endojs/endo.git"
--- a/packages/lp32/package.json
+++ b/packages/lp32/package.json
@@ -16,7 +16,7 @@
   ],
   "author": "Endo contributors",
   "license": "Apache-2.0",
-  "homepage": "https://github.com/endojs/endo/tree/master/packages/netstring#readme",
+  "homepage": "https://github.com/endojs/endo/blob/master/packages/netstring/README.md",
   "repository": {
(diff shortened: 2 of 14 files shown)


packages/exo/package.json
```diff
@@ -5,7 +5,7 @@
   "keywords": [],
   "author": "Endo contributors",
   "license": "Apache-2.0",
-  "homepage": "https://github.com/endojs/endo/tree/master/packages/exo#readme",
+  "homepage": "https://github.com/endojs/endo/blob/master/packages/exo/README.md",
```
gibson042, line 8: Is the [exo README](https://github.com/endojs/endo/blob/master/packages/exo/README.md) intentionally empty?

packages/lp32/package.json
```diff
@@ -16,7 +16,7 @@
   ],
   "author": "Endo contributors",
   "license": "Apache-2.0",
-  "homepage": "https://github.com/endojs/endo/tree/master/packages/netstring#readme",
+  "homepage": "https://github.com/endojs/endo/blob/master/packages/netstring/README.md",
```
gibson042, line 19:
```suggestion
  "homepage": "https://github.com/endojs/endo/blob/master/packages/lp32/README.md",
```

gibson042 (approved):
Looks good after fixing lp32/package.json.

packages/exo/package.json
erights: Was not intentional, just not yet written. So I added the text for my recent empty ones: pass-style, patterns, exo. Minimal but much better than nothing. It's a start. Done.

packages/lp32/package.json
erights: Done

kriskowal:
Yes, to use this style going forward, `repackage.sh` needs to be adjusted.

It would be an overstatement to describe these as problems, since the links work. The `#readme` anchor advances the scroll down the fold to the description but also allows the reader to scroll up to inspect the files in the root of the repository. With these changes, the files are a click away. I read either as valid intentions.

PR #1486, opened 2023-02-06
refactor: expand out some export star

erights:
Expand out `export * ...` into their explicit exports. A bit harder to write, but helps reading and reasoning about modularity.

See https://github.com/Agoric/agoric-sdk/pull/6932

### DIFF
--- a/packages/captp/src/index.js
+++ b/packages/captp/src/index.js
@@ -4,4 +4,10 @@ export * from '@endo/marshal';
 
 export * from './captp.js';
 export { makeLoopback } from './loopback.js';
-export * from './atomics.js';
+export {
+  MIN_DATA_BUFFER_LENGTH,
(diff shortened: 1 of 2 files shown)


turadg (commented):
> > A bit harder to write

Yeah and remember to. E.g. adding an export to `atomics.js` now requires repeating it in `index.js`.

> but helps reading and reasoning about modularity

How so? By my read, "all exports from atomics.js are exported by the package" is pretty clear to read and reason about.

mhofman:
What is the motivation for this change? One effect is that any types that may be defined in the previously `export *` is no longer re-exported on the entrypoint. In these cases, I'd argue the explicitness is a negative.

erights:
Based on review comments, closing. Thanks!

PR #1487, opened 2023-02-07
feat(compartment-mapper): wildcard policy specified as "any"

naugtur:
This change introduces an option to specify a more permissive policy so that building a more human-friendly policy format is not blocked but the impossibility of listing all possible values in the policy.

I choose the word `"any"` for the magic value because it already has a reputation because of what it means in Typescript and thus makes it more likely that a developer specifying it will get triggered to think twice.

### DIFF
--- a/packages/compartment-mapper/src/policy.js
+++ b/packages/compartment-mapper/src/policy.js
@@ -70,6 +70,36 @@ const generateCanonicalName = ({ isEntry = false, name, path }) => {
   return path.join('>');
 };
 
+const POLICY_FIELDS = ['builtins', 'globals', 'packages'];
+
+/**
+ *
+ * @param {Object} packagePolicy
+ * @param {string} field
+ * @param {string} itemName
+ * @returns {boolean|Object}
+ */
+const policyLookupHelper = (packagePolicy, field, itemName) => {
+  if (!POLICY_FIELDS.includes(field)) {
+    throw Error(`Invalid field ${q(field)}`);
+  }
+  if (
+    typeof packagePolicy !== 'object' ||
+    packagePolicy === null ||
+    !packagePolicy[field]
+  ) {
+    return false;
+  }
+
+  if (packagePolicy[field] === 'any') {
+    return true;
+  }
+  if (packagePolicy[field][itemName]) {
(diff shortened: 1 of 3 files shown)


packages/compartment-mapper/src/policy.js
```diff
@@ -70,6 +70,34 @@ const generateCanonicalName = ({ isEntry = false, name, path }) => {
...
+  }
+
+  if (packagePolicy[field]) {
+    if (packagePolicy[field] === 'any') {
+      return true;
+    }
+    if (packagePolicy[field][itemName]) {
+      return packagePolicy[field][itemName];
+    }
+  }
```
michaelfig, line 97: I'd rather see this broken out.  I'll defer to your judgement, though:
```suggestion
  if (packagePolicy[field] === 'any') {
    return true;
  }
  if (packagePolicy[field]) {
    if (packagePolicy[field][itemName]) {
      return packagePolicy[field][itemName];
    }
  }
```

packages/compartment-mapper/test/test-policy.js
```diff
@@ -63,38 +63,80 @@ const policy = {
...
-  await t.throwsAsync(
-    () => compartments.find(c => c.name.includes('alice')).import('hackity'),
-    { message: /Failed to load module "hackity" in package .*alice/ },
-    'Attempting to import a package into a compartment despite polict should fail.',
-  );
-};
+    await t.throwsAsync(
+      () => compartments.find(c => c.name.includes('alice')).import('hackity'),
+      { message: /Failed to load module "hackity" in package .*alice/ },
+      'Attempting to import a package into a compartment despite polict should fail.',
```
michaelfig, line 112:
```suggestion
      'Attempting to import a package into a compartment despite policy should fail.',
```

michaelfig (approved):
Nice and straightforward.  I like it.

packages/compartment-mapper/src/policy.js
naugtur: I need to make sure packagePolicy[field] exists first. I can make it an early return maybe.

naugtur: :+1:

PR #1491, opened 2023-02-08
feat(compartment-mapper): globals attenuation enabling LavaMoat feature parity

naugtur:
The method allowing link to remain synchronous while globals attenuators are asynchronous was inspired by module-load in SES.  

Open issue: LavaMoat style attenuator I made as a demo works around compartment.globalThis being frozen with a prototype trick. LavaMoat may demand that we make freezing globalThis of the entry compartment optional.

- [x] Implement global attenuators
- [x] cover with tests
- [x] aggregate errors from attenuators
- [x] ability to specify default attenuator to avoid repetition in policy
- [x] policy shape validation and test coverage of policy validation

Open questions:
- Do we want to future-proof the policy shape by adding an options key where defaultAttenuator would be specified?
- Do we want a debug mode where stack traces from attenuator errors would propagate up to the user?
- Do we want to version the policy shape?
- I intend to not make breaking changes to the policy format lightly, so if there are any concerns about naming things, please raise them.


packages/compartment-mapper/demo/policy/index.mjs
```diff
@@ -36,9 +39,15 @@ const options = {
       },
     },
     resources: {
```
naugtur, line 41: diff failed, look at the whole policy

kumavis:
i like it

packages/compartment-mapper/src/policy.js
```diff
@@ -1,28 +1,89 @@
...
+
+/**
+ *
+ * @param {unknown} item
+ * @param {string} path
+ * @param {string} [url]
+ * @param {object} [options]
+ * @returns {void}
+ */
+export const assertPolicyItem = (item, path, url, options = {}) => {
```
naugtur, line 35: I just realized this is not correct yet, we'll need separate validation logic for globals builtins and packages because they differ in how attenuation works.

michaelfig: This may be related to a question I have: would it make more sense to have multiple single-purpose exports from the attenuator module?  Instead of just `attenuate`, we would have `attenuateGlobals` and `attenuateModule`.

I was thinking of bringing this up in our sync, but we ran out of time.

naugtur: It can work and it does help when defining a single default attenuator. On the other hand, it prevents me from using any simple attenuator implementation for whichever purpose I wish to use it. We could use attenute export as a default for both, but that feels like bikeshedding.

packages/compartment-mapper/test/test-policy.js
```diff
@@ -100,53 +109,60 @@ const anyExpectations = {
...
-      { message: /Failed to load module "hackity" in package .*alice/ },
-      'Attempting to import a package into a compartment despite policy should fail.',
-    );
   };
 
+const assertNoPolicyBypassImport = async (t, { compartments }) => {
+  await t.throwsAsync(
+    () => compartments.find(c => c.name.includes('alice')).import('hackity'),
+    { message: /Failed to load module "hackity" in package .*alice/ },
+    'Attempting to import a package into a compartment despite polict should fail.',
```
michaelfig, line 122:
```suggestion
    'Attempting to import a package into a compartment despite policy should fail.',
```

packages/compartment-mapper/src/types.js
```diff
@@ -328,3 +328,35 @@ export {};
...
+
+/**
+ * @typedef {FullAttenuationDefinition | ImplicitAttenuationDefinition} AttenuationDefinition
+ */
+
+/**
+ * @typedef {object} UnifiedAttenuationDefinition
+ * @property {string} displayName
+ * @property {string | null} specifier
+ * @property {Array<any> | undefined} params
```
michaelfig, line 350: If this is an optional property (not just one that is explicitly `undefined`), you can express that more idiomatically with:

```suggestion
 * @property {Array<any>} [params]
```

michaelfig (approved):
Really polished stuff!  I had just a few little nits, maybe easily addressable when you rebase to merge, but feel free to ignore.

packages/compartment-mapper/test/test-policy.js
naugtur: The number of times I already fixed this typo...  :sweat_smile:

PR #1492, opened 2023-02-08
A .then detection added.

zarutian:
This is usefull for detecting intrest in what a remote promise resolves to in CapTP of Spritely Goblins, for example. In that case an op:listen or such can be sent over the connection and, crucially, skip that iff no interest is shown and the handled promise is only used as an target or an argument of an eventual send.

See [this ocapn issue](https://github.com/ocapn/ocapn/issues/15) for more details.


zarutian:
I probably need to add Ava tests and jsdoc typedef for the thenDetected trap in the pendingHandler.

zarutian:
Had not thought of that. Closing.

PR #1493, opened 2023-02-13
refactor: Tighten pattern-related type constraints

gibson042:
* Update JSDoc descriptions for pattern-related types.
* Consistently capitalize types Key, Passable, and Pattern.
* Constrain template base types for CopyArray, CopyRecord, CopySet, CopyBag, CopyMap, and Matcher.
* Approximate `[PASS_STYLE]: 'tagged'` in the CopyTagged typedef.
* Update typo in an error message.

cc @Tyrosine22

### DIFF
--- a/packages/pass-style/src/types.js
+++ b/packages/pass-style/src/types.js
@@ -21,7 +21,8 @@ export {};
  * @typedef {*} Passable
  *
  * A Passable is acyclic data that can be marshalled. It must be hardened to remain
- * stable even if some components are proxies, and is classified by PassStyle:
+ * stable (even if some components are proxies; see PureData restriction below),
+ * and is classified by PassStyle:
  *   * Atomic primitive values have a PrimitiveStyle (PassStyle
  *     "undefined" | "null" | "boolean" | "number" | "bigint" | "string" | "symbol").
  *   * Containers aggregate other Passables into
--- a/packages/pass-style/src/types.js
+++ b/packages/pass-style/src/types.js
@@ -78,14 +78,14 @@ export {};
  */
 
 /**
- * @template T
+ * @template {Passable} T
  * @typedef {T[]} CopyArray
  *
  * A Passable sequence of Passable values.
  */
 
 /**
- * @template T
+ * @template {Passable} T
  * @typedef {Record<string, T>} CopyRecord
  *
  * A Passable dictionary in which each key is a string and each value is Passable.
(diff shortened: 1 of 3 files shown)


packages/pass-style/src/types.js
```diff
@@ -20,22 +20,22 @@ export {};
 /**
  * @typedef {*} Passable
  *
- * A Passable value that may be marshalled. It is classified as one of
- * PassStyle. A Passable must be hardened.
+ * A Passable is acyclic data that can be marshalled. It must be hardened to remain
+ * stable even if some components are proxies, and is classified by PassStyle:
```
erights, line 24:
```suggestion
 * stable (even if some components are currently proxies; see restriction below), and is classified by PassStyle:
```

gibson042: Done, but is it possible to clean this up further? I believe every aspect of a Passable proxy except CopyRecord property iteration order **is** guaranteed to be stable, although proxies are obviously not inert and thus capable of _receiving_ information (e.g., encoded in an observable sequence of property accesses from code with a direct local reference). Is that the concern hinted at in PureData, and if so should we rephrase it to be less subtle?
```diff
- * This restriction assures purity *given* that none of these pass-by-copy composites
- * can be a Proxy. TODO SECURITY BUG we plan to enforce this, giving these
- * pass-by-copy composites much of the same security properties as the
- * proposed Records and Tuples (TODO need link).
+ * This restriction assures absence of side effects and interleaving risks *given*
+ * that none of the containers can be a Proxy instance.
+ * TODO SECURITY BUG we plan to enforce this, giving PureData the same security
+ * properties as the proposed [Records and Tuples](https://github.com/tc39/proposal-record-tuple).
  *
  * Given this (currently counter-factual) assumption, a PureData value cannot
  * be used as a communications channel,
  * and can therefore be safely shared with subgraphs that should not be able
  * to communicate with each other.
+ * Without that assumption, such a guarantee requires a marshal-unmarshal round
+ * trip (as exists between vats) to produce data structures disconnected from
+ * any potential proxies.
```

erights: Your suggested text above is good, and an improvement over the text it replaces. No change suggested. And anytime you can make security warnings less subtle and more explicit, please do! 

For a long time, I focused only on the threat of illicitly receiving information, and was not greatly concerned. It took me too long to realize that the bigger threat is of illicitly receiving synchronous control at moments when the innocent provoking code may be vulnerable to reentrancy attacks. Once I realized how target-rich reentrancy attacks were by providing proxies-as-hardened-data, I finally appreciated the severity of the problem.

We really really do need to ensure that `passStyleOf` does not certify a proxy as a CopyRecord, CopyArray, or Tagged. The requirement is hard enough that we should elsewhere assume that it will be enforced.

erights (approved):
LGTM, thanks!

PR #1494, opened 2023-02-16
WIP feat(ses): suppress implicit coercion of objects to numbers

erights:
This is currently for diagnostic purposes only. DO NOT MERGE.

Probably, this should never merge. But it can still be used for diagnostic purposes when needed. 

@warner , is this useful to you right now?

For this to become a merge candidate, the suppression should be moved to a separate repair file, it should be off by default, and it should only be enabled by yet another new `lockdown` option. Such an option should still be understood as "for diagnostic purposes only". The deviation from standard JS is too great to consider. 

Also, it makes v8 incredibly slower. I expect it will make most engines incredibly slower. But I haven't actually measured anything.

### DIFF
--- a/packages/ses/src/lockdown-shim.js
+++ b/packages/ses/src/lockdown-shim.js
@@ -25,6 +25,8 @@ import {
   ownKeys,
   stringSplit,
   noEvalEvaluate,
+  objectPrototype,
+  getPrototypeOf,
 } from './commons.js';
 import { enJoin } from './error/stringify-utils.js';
 import { makeHardener } from './make-hardener.js';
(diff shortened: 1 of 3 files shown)


packages/ses/src/lockdown-shim.js
```diff
@@ -57,6 +59,37 @@ import { makeCompartmentConstructor } from './compartment-shim.js';
...
+Object.prototype[TO_PRIMITIVE_SYM] = {
+  [TO_PRIMITIVE_SYM](hint) {
+    if (
+      typeof this === 'object' &&
+      this &&
+      getPrototypeOf(this) === objectPrototype &&
+      hint === 'number'
+    ) {
+      // eslint-disable-next-line @endo/no-polymorphic-call
+      Fail`Suppressing conversion of ${q(this.toString())} to number`;
```
mhofman, line 75: Slightly worried about an own `toString()` that would recursively try trigger `@@toPrimitive`.
If the intent is to used a polymorphic call, why not add a reentrancy guard around the whole method?

```suggestion
      Fail`Suppressing conversion of ${q(Object.prototype.toString.call(this))} to number`;
```

PR #1497, opened 2023-02-18
feat: Add general type-aware linting

gibson042:
This PR adds to the eslint-plugin package a new "restrict-comparison-operands" rule (inspired by and roughly analogous to [@typescript-eslint/restrict-plus-operands](https://typescript-eslint.io/rules/restrict-plus-operands)) that uses TypeScript-derived type information to reject relational comparisons between mixed types such as `number <= string` (and much more importantly for the agoric-sdk conversion of timestamps from bigints to records, **`object > bigint`**). The rule currently disallows use of unknown type for either operand by default and unconditionally allows mixed numeric comparisons such as `number >= bigint`, but could easily be tweaked and/or extended for different behavior.

It also updates the eslint-config package (relied upon by other packages) to dynamically configure type-aware linting including that new rule in the presence of a nonempty ENDO_LINT_TYPES environment variable. There are two distinct type-aware modes:
* `ENDO_LINT_TYPES=SRC` limits type-aware linting to "src" directories, increasing runtime slightly (on my slow machine, from about 1.50 minutes to 2.25 minutes [+50%] for the entire repository—which should be even further reduced if we eliminate the probably-now-redundant `tsc` invocations in "lint-types" scripts of various packages in the repository).
* `ENDO_LINT_TYPES=FULL` includes all files, increasing time _greatly_ (and making me thing that most of the slowdown motivating https://github.com/Agoric/agoric-sdk/commit/4a84690f226c2e20bce25629757de4113f79d931 was due to files in "tests").

Finally, it updates files throughout the repository to fix newly-reported lint issues.

Next steps are to try using this in agoric-sdk, where I suspect it **will** be practical to use type-aware linting when restricted to "src" directories. If so, we could choose to make that the default and drop type-oblivious "src" linting altogether.


packages/marshal/src/encodePassable.js
```diff
@@ -115,18 +128,23 @@ const decodeBinary64 = encoded => {
...
-// To ensure that the lexicographic sort order of encoded values matches the
-// numeric sort order of the corresponding numbers, the characters of the unary
-// prefix are different for negative values (type "n" followed by any number of
-// "#"s [which sort before decimal digits]) vs. positive and zero values (type
-// "p" followed by any number of "~"s [which sort after decimal digits]) and
-// each decimal digit of the encoding for a negative value is replaced with its
-// ten's complement (so that negative values of the same scale sort by
-// *descending* absolute value).
+/**
+ * Encode a JavaScript bigints using a variant of Elias delta coding, with an
```
erights, line 132:
```suggestion
 * Encode JavaScript bigints using a variant of Elias delta coding, with an
```

packages/eslint-plugin/lib/rules/restrict-comparison-operands.js
```diff
@@ -0,0 +1,161 @@
...
+        `Missing full type information. Make sure eslint configuration "parser" uses @typescript-eslint/parser parser and "parserOptions" specifies a "project" TypeScript configuration that includes each file subject to this rule.`,
+      );
+    }
+
+    const comparableTypeOf = type => {
+      if (type.isNumberLiteral()) {
+        return 'number';
+      } else if (
+        type.isStringLiteral() ||
+        getTypeFlags(type) & ts.TypeFlags.TemplateLiteral
```
erights, line 82: Does this exclude tagged template literals?

erights (commented):
29/30 files LGTM. The remaining file restrict-comparison-operands.js does not look bad to me. But I do not know enough about the internals of either TS or eslint to review.

jsconfig.eslint-base.json
```diff
@@ -0,0 +1,12 @@
...
+  "compilerOptions": {
+    "target": "esnext",
+    "module": "esnext",
+    "checkJs": false,
+    "noEmit": true,
+    "downlevelIteration": true,
+    "strictNullChecks": true,
+    "moduleResolution": "node"
+  },
+  "include": []
```
turadg, line 11: do we need to have three config files? I wonder if "src" could be the base and "full" extends from that. Would redundant globs slow things down?

package.json
```diff
@@ -10,6 +10,7 @@
   },
   "devDependencies": {
     "@octokit/core": "^3.4.0",
+    "@typescript-eslint/eslint-plugin": "^5.52.0",
```
turadg, line 13: fwiw there was a new release yesterday but the release notes don't list anything relevant: https://github.com/typescript-eslint/typescript-eslint/releases/tag/v5.53.0

packages/bundle-source/src/tool.js
```diff
@@ -143,6 +143,7 @@ export const makeBundleCache = (wr, cwd, readPowers, opts) => {
         return { relativePath, mtime: mtime.toISOString() };
       }),
     );
+    // eslint-disable-next-line @endo/restrict-comparison-operands
```
turadg, line 146: when we suppress an error there should be a comment with the excuse :) 

but even better don't suppress. wouldn't this pass if `meta` were annotated? I take it `bundleTime` is a string like `mtime.toISOString()`.

packages/bundle-source/src/tool.js
```diff
@@ -196,7 +197,7 @@ export const makeBundleCache = (wr, cwd, readPowers, opts) => {
 };
 
 export const main = async (args, { fs, url, crypto, path }) => {
-  const [to, dest, ...pairs] = args;
+  const [to, dest, ...pairs] = /** @type {[string, string, string[]]} */ (args);
```
turadg, line 200: please type the function signature than cast. fine with any on the second param: `{*} io`

packages/compartment-mapper/src/infer-exports.js
```diff
@@ -57,6 +57,7 @@ function* interpretBrowserField(name, browser, main = 'index.js') {
 function* interpretExports(name, exports, tags) {
   if (isArray(exports)) {
     for (const section of exports) {
+      /** @type {Array} */
```
turadg, line 60: what happens without this? is TS not picking up the `yields` return type?

packages/eslint-config/NEWS.md
```diff
@@ -0,0 +1,4 @@
+# Next release
+* Support type-aware linting by environment variable
+  * `ENDO_LINT_TYPES=SRC`: Limit to "src" directories (increases time slightly).
```
turadg, line 3: since you have the data, how about quantifiying?
```suggestion
  * `ENDO_LINT_TYPES=SRC`: Limit to "src" directories (increases time ~50%).
```

packages/eslint-config/eslint-config.cjs
```diff
@@ -1,14 +1,73 @@
-// ESLint reads this as a JSONC file so comments are allowed.
```
turadg, line 1: 👍 to cjs

packages/eslint-config/eslint-config.cjs
```diff
@@ -1,14 +1,73 @@
...
+    : "./{js,ts}config.eslint-src.json";
+  const parserOptions = {
+    tsconfigRootDir: path.join(__dirname, "../.."),
+    project: [rootTsProjectGlob, "packages/*/{js,ts}config.eslint.json"],
+  };
+  const rules = {
+    "@typescript-eslint/restrict-plus-operands": "error",
+  };
+  if (isFull) {
+    dynamicConfig.extends.push("plugin:@endo/recommended-requiring-type-checking");
```
turadg, line 42: this reads at first blush as if `isFull` has different rules. Downsides of just using `overrides` in both cases? That would be much easier to follow.

packages/eslint-plugin/lib/configs/recommended-requiring-type-checking.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  extends: ['plugin:@endo/recommended'],
+  rules: {
+    '@endo/restrict-comparison-operands': 2,
```
turadg, line 4:
```suggestion
    '@endo/restrict-comparison-operands': 'error',
```

packages/eslint-plugin/lib/rules/restrict-comparison-operands.js
```diff
@@ -0,0 +1,161 @@
...
+    ],
+  },
+  defaultOptions: [
+    {
+      allowUnknown: false,
+    },
+  ],
+  create(context, [{ allowUnknown }]) {
+    // Loosely follows
+    // https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/restrict-plus-operands.ts
```
turadg, line 61: I'm going to assume that code works and not review this file closely.

packages/marshal/src/marshal-justin.js
```diff
@@ -92,6 +92,7 @@ const makeNoIndenter = () => {
     line: () => strings.length,
     next: token => {
       if (strings.length >= 1) {
+        /** @type {string} */
```
turadg, line 95: better to type `const strings` than cast.

packages/marshal/src/marshal-justin.js
```diff
@@ -353,8 +354,9 @@ const decodeToJustin = (encoding, shouldIndent = false, slots = []) => {
         }
 
         case 'slot': {
-          let { index, iface } = rawTree;
-          index = Number(Nat(index));
+          // eslint-disable-next-line prefer-const
+          let { index: untypedIndex, iface } = rawTree;
+          const index = Number(Nat(untypedIndex));
```
turadg, line 359:
```suggestion
          let { iface } = rawTree;
          const index = Number(Nat(rawTree.index));
```

packages/marshal/src/rankOrder.js
```diff
@@ -99,7 +99,7 @@ const { entries, fromEntries, setPrototypeOf, is } = Object;
 const sameValueZero = (x, y) => x === y || is(x, y);
 
 const trivialComparator = (left, right) =>
-  // eslint-disable-next-line no-nested-ternary
+  // eslint-disable-next-line no-nested-ternary, @endo/restrict-comparison-operands
```
turadg, line 102: wouldn't it be good to type the function signature so there are type errors if uncomparable are passed in? could be worth naming the type `number | bigint | string`.

packages/marshal/src/rankOrder.js
```diff
@@ -272,6 +275,7 @@ export const makeComparatorKit = (compareRemotables = (_x, _y) => 0) => {
         throw Fail`Unrecognized passStyle: ${q(leftStyle)}`;
       }
     }
+    /* eslint-enable */
```
turadg, line 278: does this re-enable everything that's been disabled? consider being more explicit

packages/pass-style/src/copyArray.js
```diff
@@ -7,25 +7,32 @@ const { getPrototypeOf } = Object;
...
   assertValid: (candidate, passStyleOfRecur) => {
-    CopyArrayHelper.canBeValid(candidate, assertChecker);
+    canBeValid(candidate, assertChecker);
     getPrototypeOf(candidate) === arrayPrototype ||
       assert.fail(X`Malformed array: ${candidate}`, TypeError);
     // Since we're already ensured candidate is an array, it should not be
     // possible for the following test to fail
     checkNormalProperty(candidate, 'length', false, assertChecker);
-    const len = candidate.length;
+    const len = /** @type {unknown[]} */ (candidate).length;
```
turadg, line 35: better at the function signature

packages/patterns/src/keys/checkKey.js
```diff
@@ -430,7 +430,7 @@ export const getCopyMapEntries = m => {
   const {
     payload: { keys, values },
   } = m;
-  const { length } = keys;
+  const { length } = /** @type {Array} */ (keys);
```
turadg, line 433: can the param be typed? Why isn't `keys` inferred from CopyMap? Oh, it's an alias for `CopyTagged` which says `payload` is just Passable/any.  Might be a good time to have a proper CopyMap type.

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -958,7 +966,8 @@ const makePatternKit = () => {
       const { stringLengthLimit } = limit(limits);
       return (
         checkKind(specimen, 'string', check) &&
-        (specimen.length <= stringLengthLimit ||
+        // eslint-disable-next-line prettier/prettier -- JSDoc requires these parentheses!
```
turadg, line 969: bumping Prettier 2.8.0 to 2.8.4 might fix this. I don't see that in the [changelog](https://github.com/prettier/prettier/blob/main/CHANGELOG.md) but I feel like I've stopped running into this problem in agoric-sdk since a recent bump

packages/zip/src/buffer-reader.js
```diff
@@ -3,8 +3,21 @@
...
+ * @property {number} length
+ * @property {number} index
+ * @property {number} offset
+ */
+
+/** @type {WeakMap<BufferReader, BufferReaderState>} */
 const privateFields = new WeakMap();
 
+/** @type {(bufferReader: BufferReader) => BufferReaderState} */
+const privateFieldsGet = privateFields.get.bind(privateFields);
```
turadg, line 19: why this change?

packages/zip/src/buffer-writer.js
```diff
@@ -24,7 +24,7 @@ const getPrivateFields = self => {
 };
 
 const assertNatNumber = n => {
-  if (Number.isSafeInteger(n) && n >= 0) {
+  if (Number.isSafeInteger(n) && /** @type {number} */ (n) >= 0) {
```
turadg, line 27: /me disappointed that TS not inferring from `Number.isSafeInteger`.

yarn.lock
```diff
@@ -2288,6 +2332,33 @@
     semver "^7.3.7"
     tsutils "^3.21.0"
 
+"@typescript-eslint/typescript-estree@5.52.0":
```
turadg, line 2335: some new dupes here. try `npx yarn-deduplicate`

turadg:
> Next steps are to try using this in agoric-sdk, where I suspect it will be practical to use type-aware linting when restricted to "src" directories. If so, we could choose to make that the default and drop type-oblivious "src" linting altogether.

 👍 Excellent!

packages/patterns/src/keys/checkKey.js
erights: I'm abstain from whether now is a good time. But it would indeed be wonderful to have proper typing of CopyMap<K,V>, CopySet<K>, and CopyBag<K>. The other main category of Tagged are the Matchers. Would also be nice to have proper typing of them, but less important.

erights:
> If so, we could choose to make that the default and drop type-oblivious "src" linting altogether.

To programmers like me with only a loose understanding of TS type rules, and who often needs to write code that starts off violating these rules, what are the practical implications of dropping type-oblivious linting? What additional work does such a programmer need to do before their code lints cleanly?

packages/marshal/src/rankOrder.js
gibson042: Do you know how to do that? The following is not enough:
```js
/**
 * @template {number | bigint | string} T
 * @param {T} left
 * @param {T} right 
 * @returns {RankComparison}
 */ 
const trivialComparator = (left, right) =>
  // eslint-disable-next-line no-nested-ternary
  left < right ? -1 : left === right ? 0 : 1;
```

packages/marshal/src/rankOrder.js
gibson042: It does ("[_`/* eslint-enable */` without any specific rules listed causes all disabled rules to be re-enabled._](https://eslint.org/docs/latest/use/configure/rules#using-configuration-comments-1)"), and being more explicit is bad here because it can enable rules that were never on.

packages/pass-style/src/copyArray.js
gibson042: But not appropriate here, because `candidate` really could have any type.

packages/zip/src/buffer-reader.js
gibson042: To get the typing right in one location (`privateFields.get(this).length` etc. are rejected by TypeScript because it doesn't know that `this` is guaranteed to be a key of `privateFields`).

jsconfig.eslint-base.json
gibson042: It has a big enough impact for me to prefer this approach.

packages/compartment-mapper/src/infer-exports.js
gibson042: No, the issue seems to be with `[...val]` (and `Array.from(val)`, for that matter) not being recognized as an array.

packages/eslint-config/eslint-config.cjs
gibson042: Implementing this was trickier than expected, but uncovered some nice improvements. Thanks!

packages/eslint-plugin/lib/rules/restrict-comparison-operands.js
gibson042: It does, but your comment prompted me to dig deeper and ultimately ignore literal vs. non-literal in favor of TypeFlags.{Number,String,BigInt}Like bitmasks (in which e.g. `'foo'`, ``` `bar${suffix}` ```, and `'baz' + String(getSuffix())` are all equally well covered).

yarn.lock
gibson042: TIL about `yarn-deduplicate`, but running it made no changes.

packages/bundle-source/src/tool.js
gibson042: Added an explanation.

packages/zip/src/buffer-writer.js
gibson042: Me too.

gibson042:
> > If so, we could choose to make that the default and drop type-oblivious "src" linting altogether.
> 
> To programmers like me with only a loose understanding of TS type rules, and who often needs to write code that starts off violating these rules, what are the practical implications of dropping type-oblivious linting? What additional work does such a programmer need to do before their code lints cleanly?

Basically just disabling the rules. There are a couple examples in [packages/ses/types.test-d.ts](https://github.com/endojs/endo/pull/1497/files#diff-1c85b197ae6044ea121bfdc4cf572798bb261d2a7e9e72df3e33a9e90951c623) to verify errors when adding bigint to number and number to string:
```js
(n: number | bigint) => {
  // @ts-expect-error
  // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
  return n + 10n;
};
```
```js
(n: number | string) => {
  // @ts-expect-error
  // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
  return n + 10;
};
```

packages/compartment-mapper/src/infer-exports.js
turadg: That was because `val` was `any`, because `interpretExports` was returning `any`, because https://github.com/Microsoft/TypeScript/issues/23857

I've got a fix in https://github.com/endojs/endo/tree/ta/gibson-2023-02-type-aware-linting

packages/marshal/src/rankOrder.js
turadg: That should work, yeah. I suspect it's a limitation of how `restrict-comparison-operands` is implemented. Maybe it doesn't handle subtypes of the comparable types? `T extends string | number | bigint`

The suppression is fine in that case, but the limitation is worth documenting.

packages/pass-style/src/copyArray.js
turadg: I'm confused. You're saying with the cast that it's `unknown[]`. So I'm saying to just put that in the function signature.

packages/zip/src/buffer-reader.js
turadg: Ok. Probably better than trying to make it work with `ThisType`. But consider a source code comment with what you explained here

packages/patterns/src/keys/checkKey.js
turadg: Looks like @gibson042 landed a working type yesterday: https://github.com/endojs/endo/pull/1493/files#diff-a81b0292a3e5e26fea27548664397a190fdf3b8f407031f508fa1c460dce77efR112

turadg (approved):
Thanks for addressing the feedback.

When I tried running the tool I got a warning,
```
❯ j bundle-source
❯ yarn lint
yarn run v1.22.0
$ eslint '**/*.js'
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <4.9.0

YOUR TYPESCRIPT VERSION: 4.9.4

Please only submit bug reports when using the officially supported version.

=============
```

So I bumped the versions in 0bd1303c8033230d2cffaf1578e4d481c65773f5

I also did some more annotating to reduce the suppressions: 67dafb5ffc2569fee57afd5d1b88d5d7d7fd68bf

Please consider including those but if not I'll just PR them after this lands.
 https://github.com/endojs/endo/compare/gibson-2023-02-type-aware-linting...ta/gibson-2023-02-type-aware-linting

gibson042:
Thanks! I incorporated the changes and am merging now!

Issue #1489, opened 2023-02-07
`assertKeyPattern(M.kind("null"))` fails with "null keys are not supported")


gibson042:
I observed when exploring https://github.com/Agoric/agoric-sdk/pull/6903#discussion_r1099042215 that a `'null'` case appears to be missing from matchKindHelper's [`checkKeyPattern`](https://github.com/endojs/endo/blob/686ab61802c0e5206b14f4b91d499e3873555cd4/packages/patterns/src/patterns/patternMatchers.js#L887) (e.g., `matches(undefined, M.kind("undefined"))` and `matches(null, M.kind("null"))` are both true, but `assertKeyPattern(M.kind("undefined"))` passes while `assertKeyPattern(M.kind("null"))` fails with "_null keys are not supported_"), but then realized that `kind` has no documentation and its purpose isn't quite clear. Do we really want to be conflating passStyleOf results with tagged record tags, and if so then which of them count as "key" patterns and which do not?

erights:
First, on the narrow issue:

> "null keys are not supported"

Looks like a bug to me. 

@FUDCo , 
IIRC, `assertKeyPattern` was added to support something you were doing in collectionManager.js . Do you remember what this was about, and if `'null'` was excluded on purpose? Would anything break if we stopped excluding null?

erights:
On the broader issue:

> Do we really want to be conflating passStyleOf results with tagged record tags, and if so then which of them count as "key" patterns and which do not?

I initially felt very uncomfortable with this, and included a TODO or similar comment warning that we really need to revisit this mixing of namespaces. Since then, I have become more comfortable with it for reasons I can explain, and I know of no alternative I'd like better. We should certainly talk about it, and see if we can invent something better that I missed.

erights:
The name "passStyleOrRecordTag" should certainly be changed, as it no longer reflects how I think of the kind taxonomy.

erights:
There is no `passStyleOrRecordTag`. Renaming this bug to the problem being reported.

But @gibson042 , I owe you an explanation of the "kind" namespace, and why I became comfortable with it. Let's talk.

erights:
Fixed by https://github.com/Agoric/agoric-sdk/pull/7035

PR #1498, opened 2023-02-19
fix: omitted null case

erights:
Tracks https://github.com/Agoric/agoric-sdk/pull/7034

Fixes https://github.com/endojs/endo/issues/1489

The `checkKeyPattern` test seems to have forgotten to include the `'null':` case. This PR adds it. (It also rearranges the cases into the order of the `PassStyle` type declaration.)

@FUDCo , `checkKeyPattern` came from you, so I need you to verify that this omission was not intentional.

### DIFF
--- a/packages/patterns/src/patterns/patternMatchers.js
+++ b/packages/patterns/src/patterns/patternMatchers.js
@@ -886,13 +886,14 @@ const makePatternKit = () => {
 
     checkKeyPattern: (kind, check) => {
       switch (kind) {
+        case 'undefined':
+        case 'null':
         case 'boolean':
         case 'number':
         case 'bigint':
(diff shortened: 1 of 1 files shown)


erights:
@FUDCo it looks like `assertKeyPattern`'s purpose is to check a pattern to see if that pattern would only match scalars. If so, then it does look like the omission of `null` was accidental. Yes?

erights:
That purpose of `assertKeyPattern` seems inconsistent with https://github.com/Agoric/agoric-sdk/issues/6953 , so this is still mysterious.

erights:
Closing in favor of https://github.com/Agoric/agoric-sdk/pull/7035

PR #1500, opened 2023-02-23
ci: Update to use type-aware linting

gibson042:
...but limit enhanced linting to "src" files (for now).

Ref #1497

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,7 @@ jobs:
       run: yarn build
 
     - name: Run yarn lint
-      run: ENDO_LINT_TYPES=SRC yarn lint
+      run: yarn lint
 
   test:
     name: test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,7 @@ jobs:
       run: yarn build
 
     - name: Run yarn lint
-      run: yarn lint
+      run: ENDO_LINT_TYPES=SRC yarn lint
 
   test:
     name: test
--- a/packages/captp/src/captp.js
+++ b/packages/captp/src/captp.js
@@ -53,7 +53,7 @@ const reverseSlot = slot => {
 };
 
 /**
- * @typedef {Object} CapTPOptions the options to makeCapTP
+ * @typedef {object} CapTPOptions the options to makeCapTP
  * @property {(val: unknown, slot: CapTPSlot) => void} [exportHook]
  * @property {(val: unknown, slot: CapTPSlot) => void} [importHook]
  * @property {(err: any) => void} [onReject]
@@ -80,7 +80,9 @@ export const makeCapTP = (
   bootstrapObj = undefined,
   opts = {},
 ) => {
+  /** @type {Record<string, number>} */
   const sendCount = {};
+  /** @type {Record<string, number>} */
   const recvCount = {};
   const getStats = () =>
     harden({
--- a/packages/compartment-mapper/src/link.js
+++ b/packages/compartment-mapper/src/link.js
@@ -12,6 +12,7 @@
 /** @typedef {import('./types.js').CompartmentMapDescriptor} CompartmentMapDescriptor */
 /** @typedef {import('./types.js').DeferredAttenuatorsProvider} DeferredAttenuatorsProvider */
 /** @typedef {import('./types.js').LinkOptions} LinkOptions */
+/** @template T @typedef {import('@endo/eventual-send').ERef<T>} ERef */
 
 import { resolve } from './node-module-specifier.js';
 import { parseExtension } from './extension.js';
@@ -28,6 +29,11 @@ const { entries, fromEntries } = Object;
 const { hasOwnProperty } = Object.prototype;
 const { apply } = Reflect;
 const { allSettled } = Promise;
+
+/**
+ * @template T
+ * @type {(iterable: Iterable<ERef<T>>) => Promise<Array<PromiseSettledResult<T>>>}
+ */
 const promiseAllSettled = allSettled.bind(Promise);
 
 const inertStaticModuleRecord = {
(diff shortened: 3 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -59,7 +59,7 @@ jobs:
       run: yarn build
 
     - name: Run yarn lint
-      run: yarn lint
+      run: ENDO_LINT_TYPES=SRC yarn lint
```
turadg, line 62: I think people may be surprised for `yarn lint` in their terminal to have different results than CI.

How much slower is it? If it's tolerable I think it would be better to make `SRC` the default and have `NONE` be optional (along with `FULL`)

gibson042:
Increased GitHub actions lint time from about 2.25 minutes ([2m22s for #1497](https://github.com/endojs/endo/actions/runs/4244731219/jobs/7379263050), [2m13s for #1499](https://github.com/endojs/endo/actions/runs/4256226922/jobs/7404829924)) to about 3.5 minutes ([3m26s for this PR](https://github.com/endojs/endo/actions/runs/4256229553/jobs/7404835927)), right in line with the expected 50% hit.

That seems perfectly acceptable to me, and I bet it renders redundant all the [`lint:types` automation targets](https://github.com/endojs/endo/blob/b08f5e0e62812257f17aa54bba170822232c6ca3/scripts/repackage.sh#L83-L85) (or at least most of them), allowing for further speedup in the future.

.github/workflows/ci.yml
gibson042: Yes, I have similar thoughts but just wasn't sure how far to take things in this step.

turadg: I've seen people get confused by CI having different results for `yarn lint` than CLI does. So I think they should stay in sync. But I also think the 50% hit is acceptable so I'd say to make `SRC` the default in this PR. Maybe run it by team-engineering for approval of the workflow impact.

turadg:
> That seems perfectly acceptable to me
👍 

> renders redundant all the [lint:types automation targets](https://github.com/endojs/endo/blob/b08f5e0e62812257f17aa54bba170822232c6ca3/scripts/repackage.sh#L83-L85) (or at least most of them), allowing for further speedup in the future.

I think not. IIIUC TS for Eslint just includes types in the AST so Eslint rules can use them. It won't do what tsc does finding type errors.

gibson042:
> I think not. IIIUC TS for Eslint just includes types in the AST so Eslint rules can use them. It won't do what tsc does finding type errors.

I'd like to test this hypothesis. Can you think of some code that would do so?

gibson042:
Thanks, and confirmed. I've updated eslint configuration to a `ENDO_LINT_TYPES=SRC` default.

turadg (approved):
Changes LTGM, assuming the workflow impact is acceptable to the team.

SMotaal:
@gibson042 I did a check to see which files are un/claimed, and this lead me to a question.

**Findings**:

- All nested files in `jsconfig.eslint-src.json` are claimed by source files:
  ```json
  { "fileCount": { "scoped": 0, "nested": 247, "unclaimed": 0 } }
  ```
- Not all files matched by `jsconfig.eslint-full.json` are claimed:
  ```json
  { "fileCount": { "scoped": 4, "nested": 562, "unclaimed": 332 } }
  ```

I can share the bulk of the sets if this would be helpful.

**Thoughts**:

- Worth noting that the expected behaviour of Visual Studio Code is that files which are not claimed by either a `jsconfig.json` or `tsconfig.json` are treated differently from ones which are claimed.
- The specific precedence for which `jsconfig.json` or `tsconfig.json` applies to a file is escaping me, suppose it is either the first one matching the file top-down, or the first one matched in the parent folders for a given file.

**Question**:

- I am getting a little lost trying to figure out how this plays out in the editor, can you elaborate please?

Thanks!

gibson042:
I'm not sure what you're asking, can you rephrase? Whatever method you used to identify claimed vs. unclaimed should apply in the context of a specific tsconfig, and I'm wondering about that process. As for VS Code, I'd be surprised if it gave special treatment to a jsconfig.eslint-$suffix.json file, but #1499 updated package-level {jsconfig,tsconfig}.json files to explicitly extend from the top-level jsconfig.eslint-base.json (which does not itself specify any included files, although those would be overridden anyway per [tsconfig#extends](https://www.typescriptlang.org/tsconfig#extends)).

SMotaal:
Okay, let me try to split things out. Also to clarify, I am appreciating this is extending outside the context of the PR and that it was imperative to harmonize the trickle down in configs.

---

My question is more about what comes next, knowing that when VS Code does not find a respective `jsconfig.json` or `tsconfig.json` such a file will be treated in ways that may or may not be aligned with the linter, which aligns with what you stated about `jsconfig.eslint-$suffix.json` files.

With this, we now end up with following logical possibilities:

- Files claimed by a linting only config.
- Files claimed by a VS Code only config.
- Files claimed by both linting and VS Code with measures to ensure they match.
- Entirely orphaned files which VS code treats per user configuration and which are ignored by the linter without any such measures at least for now.

Does this sound this track or did I miss something?

Now, the question is more about the experience in the editor and the concerns that this could lead to cases where VS Code is treating a file differently from the linter?

I was wondering if you may have had some thoughts on this moving forward.

---

The analysis I am doing uses `ts.parseJsonConfigFileContent(…)` to parse all relevant configs (ie un-suffixed or indicated in package scripts) while honouring `extends`… etc then compares the resulting `fileNames` arrays to determine:

- Scoped or nested counts reflect files are those matched by a config including top-level and nested configs relative to package scopes (ie `package.json`) to align with the `nodejs` module resolution.
- Claimed or unclaimed counts reflect the subset of files matched by a top-level config which are in fact also in nested packages and are either matched by a respective nested config or not.

In that sense:

1. A file that is not matched by any relevant configs can still exist and is not counted at this point.
2. Unclaimed counts are explicitly indicative of nested files that were only matched by a top-level config but not by a package-level config and I felt I needed to ask for more clarity.

I appreciate this may take a bit back and fourth until there are mutually framings, but that's just part of the process, which I appreciate in order to refine the analysis around the most practical insights possible.

---

Thanks again!

PR #1505, opened 2023-03-02
a scaffold setup that runs tests with secure bundle

naugtur:


packages/compartment-mapper/test/scaffold.js
```diff
@@ -287,6 +289,53 @@ export function scaffold(
...
+      });
+      const bundle = await makeSecureBundleFromArchive(
+        {
+          read: async _path => {
+            return archive;
+          },
+        },
+        'fakeLocation',
+        {
+          modules: Object.fromEntries(
```
kumavis, line 314: what does specifying the modules here do

naugtur: nothing yet. It doesn't work, so I hacked them away in the fixture. I have a feeling I can fix it

PR #1506, opened 2023-03-03
exitModuleImportHook alongside synchronous exit module linking

naugtur:
### Abandoned in favor of https://github.com/endojs/endo/pull/1507

This is a PoC exitModuleImportHook.

It works, but not necessarily the way we want it to work. 

Notes:

- It seems to make sense for the exitModuleImportHook to return a `ThirdPartyStaticModuleInterface`
- I am exploring replacing exitModules with the exitModuleImportHook entirely, which removes a lot of complications from linking, changes module attenuation slightly (might be for the better, not sure yet) but it changes how modules from external assemblies or compartments could get included. I don't see a lot of use for that though. 
- I don't yet see a way to properly attenuate on top of exitModuleImportHook - I just discovered that attenuation would only run once per builtin and then it gets cached by the specifier, so attenuation would have to return a unique alias record at least. I'm not sure if that's enough.

### DIFF
--- a/packages/compartment-mapper/src/import-archive.js
+++ b/packages/compartment-mapper/src/import-archive.js
@@ -14,6 +14,7 @@
 /** @typedef {import('./types.js').ComputeSourceLocationHook} ComputeSourceLocationHook */
 /** @typedef {import('./types.js').LoadArchiveOptions} LoadArchiveOptions */
 /** @typedef {import('./types.js').ExecuteOptions} ExecuteOptions */
+/** @typedef {import('./types.js').ExitModuleImportHook} ExitModuleImportHook */
 
 import { ZipReader } from '@endo/zip';
 import { link } from './link.js';
@@ -78,6 +79,7 @@ const postponeErrorToExecute = errorMessage => {
  * @param {string} archiveLocation
  * @param {HashFn} [computeSha512]
  * @param {ComputeSourceLocationHook} [computeSourceLocation]
+ * @param {ExitModuleImportHook} [exitModuleImportHook]
  * @returns {ArchiveImportHookMaker}
  */
 const makeArchiveImportHookMaker = (
@@ -86,6 +88,7 @@ const makeArchiveImportHookMaker = (
   archiveLocation,
   computeSha512 = undefined,
   computeSourceLocation = undefined,
+  exitModuleImportHook = undefined,
 ) => {
   // per-assembly:
   /** @type {ArchiveImportHookMaker} */
@@ -97,6 +100,23 @@ const makeArchiveImportHookMaker = (
       // per-module:
       const module = modules[moduleSpecifier];
       if (module === undefined) {
+        if (exitModuleImportHook) {
+          console.error('#################x');
+          const ns = await exitModuleImportHook(moduleSpecifier);
+          if (ns) {
+            return {
+              record: freeze({
+                imports: [],
+                exports: Object.keys(ns),
+                execute: moduleExports => {
+                  moduleExports.default = ns;
+                  Object.assign(moduleExports, ns);
+                },
+              }),
+              specifier: moduleSpecifier,
+            };
+          }
+        }
         throw new Error(
           `Cannot find module ${q(moduleSpecifier)} in package ${q(
             packageLocation,
@@ -190,6 +210,7 @@ const makeFeauxModuleExportsNamespace = Compartment => {
  * @param {string} [options.expectedSha512]
  * @param {HashFn} [options.computeSha512]
  * @param {Record<string, unknown>} [options.modules]
+ * @param {ExitModuleImportHook} [options.exitModuleImportHook]
  * @param {Compartment} [options.Compartment]
  * @param {ComputeSourceLocationHook} [options.computeSourceLocation]
  * @returns {Promise<Application>}
@@ -205,6 +226,7 @@ export const parseArchive = async (
     computeSourceLocation = undefined,
     Compartment = DefaultCompartment,
     modules = undefined,
+    exitModuleImportHook = undefined,
   } = options;
 
   const archive = new ZipReader(archiveBytes, { name: archiveLocation });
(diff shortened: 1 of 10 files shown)


packages/compartment-mapper/src/import-archive.js
```diff
@@ -97,6 +100,23 @@ const makeArchiveImportHookMaker = (
...
+          console.error('#################x');
+          const ns = await exitModuleImportHook(moduleSpecifier);
+          if (ns) {
+            return {
+              record: freeze({
+                imports: [],
+                exports: Object.keys(ns),
+                execute: moduleExports => {
+                  moduleExports.default = ns;
+                  Object.assign(moduleExports, ns);
```
naugtur, line 113: This uses the `ThirdPartyStaticModuleInterface` style, which might lead to compatibility issues, but until Module work in TC39 gets into the language and SES becomes compatible with it, I don't see a more transparent way to interface any module here regardless of it being imported or required.

We could burden the exitModuleImportHook implementer with returning a record though, in which case they get more freedom to figure out what the default and named exports should be.

I think both versions could work.

packages/compartment-mapper/src/import-archive.js
```diff
@@ -264,6 +285,7 @@ export const parseArchive = async (
       archiveLocation,
       computeSha512,
       computeSourceLocation,
+      exitModuleImportHook,
```
naugtur, line 288: I don't enjoy passing the hook to the archive making part of the archive flow. 
It should not be necessary before runtime. 

Solutions:

1. A boolean configuration value saying to "expectExitModuleImportHook" would be enough to omit the archive-time errors where exit import hook would kick in.   
pros: no unnecessary importing and hook mechanics before runtime  
cons: no early errors if exitModuleImportHook doesn't provide a module, another config item

2. exitModuleImportHook accepts options, including `archiveOnly` and if that's truthy, it can avoid importing actual modules while still returning undefined where it originally would (eg. if checking against a list of known builtins it wishes to support)  
pros: no additional option, no need to import actual modules during archiving  
cons: more complexity in exitModuleImportHook

packages/compartment-mapper/src/types.js
```diff
@@ -239,6 +245,7 @@ export {};
  * @property {Array<Transform>} [transforms]
  * @property {Array<Transform>} [__shimTransforms__]
  * @property {Record<string, object>} [modules]
+ * @property {ExitModuleImportHook} [exitModuleImportHook]
```
naugtur, line 248: I'm wondering if 
```
* @property {Record<string, object> | ExitModuleImportHook} [modules]
```
would be frowned upon.

packages/compartment-mapper/test/scaffold.js
```diff
@@ -78,6 +78,7 @@ export function scaffold(
     tags = undefined,
     searchSuffixes = undefined,
     commonDependencies = undefined,
+    additionalOptions = {},
```
naugtur, line 81: Can't extend scaffold with new minor bits of logic forever...

packages/compartment-mapper/src/import-archive.js
michaelfig: I would rename this parameter to `isExitModuleImportAllowed`, since the logic for it may be distinct from the logic of `exitModuleImportHook`, and it only needs to return truthy or falsy (default: `() => false`).  Using `myExitModuleImportHook` for the parameter is a degenerate case that works, but isn't ideal (it does too much work, as you point out above).

packages/compartment-mapper/src/types.js
michaelfig: I'd prefer not to mix types in this way in the API (policy description type mixing as you've done in the past is a different matter, for reasons of needing to be concise).  I'd rather have just an assertion around `!modules || !exitModuleImportHook`.

packages/compartment-mapper/src/import-archive.js
```diff
@@ -97,6 +100,16 @@ const makeArchiveImportHookMaker = (
       // per-module:
       const module = modules[moduleSpecifier];
       if (module === undefined) {
+        if (exitModuleImportHook) {
+          console.error('#################x');
```
michaelfig, line 104: Leftover debug message.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -156,6 +158,21 @@ export const makeImportHookMaker = (
           // Archived compartments are not executed.
           return freeze({ imports: [], exports: [], execute() {} });
         }
+        if (exitModuleImportHook) {
+          console.error('#################i');
```
michaelfig, line 162: Here again.

packages/compartment-mapper/src/types.js
michaelfig: I'm willing to remove the `modules` option entirely if you mark it as a breaking change, and update any uses within endo that depend on it to use the `exitModuleImportHook` instead.

naugtur: I see you're assuming this is more ready than it really is. 
I experimented with removing the modules option entirely and it does improve a bunch of things, but also has some interesting consequences. I'll talk about it in the meeting.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -156,6 +158,21 @@ export const makeImportHookMaker = (
...
+          if (record) {
+            // It'd be nice to check the policy before importing it, but we can only throw a policy error if the
+            // hook returns something. Otherwise, we need to fall back to the 'cannot find' error below.
+            enforceModulePolicy(moduleSpecifier, compartmentDescriptor, {
+              exit: true,
+            });
+            // note it's not being marked as exit in sources
+            // it could get marked and the second pass, when the archive is being executed, would have the data
+            // to enforce which exits can be dynamically imported
+            return record;
```
naugtur, line 173: Introducing attenuation by policy in here is going to be difficult, because each time a builtin is imported it'd have to get a new instance for different attenuations and here it'll get memoized.

packages/compartment-mapper/src/import-hook.js
naugtur: yes, it's still early in the implementation and these are useful for showing when and what happens. I'll need them for the meeting.

PR #1507, opened 2023-03-07
Sketching what it'd look like if we removed exitModules in favor of exitModuleImportHook 

naugtur:


packages/compartment-mapper/src/link.js
```diff
@@ -274,24 +247,6 @@ const makeModuleMapHook = (
...
-      // there isn't a module descriptor for every module.
-      moduleDescriptors[moduleSpecifier] = { exit: moduleSpecifier };
-      if (archiveOnly) {
-        return inertModuleNamespace;
-      } else {
-        return attenuateModuleHook(
-          moduleSpecifier,
-          exitModules[moduleSpecifier],
-          compartmentDescriptor.policy,
-          attenuators,
```
naugtur, line 292: module attenuation is gone from link, so pending jobs can be removed too.

packages/compartment-mapper/src/policy.js
```diff
@@ -381,57 +381,51 @@ export const enforceModulePolicy = (specifier, compartmentDescriptor, info) => {
...
-        const staticModuleRecord = freeze({
-          imports: [],
-          exports: keys(ns),
-          execute: moduleExports => {
-            assign(moduleExports, ns);
-          },
-        });
-        return staticModuleRecord;
-      },
+  // No need for a new compartment anymore, since instead of a module namespace, we're now producing a module record.
```
naugtur, line 392: No need for a new compartment for each attenuation, since instead of a module namespace, we're now producing a module record!

packages/compartment-mapper/test/snapshots/test-error-handling.js.md
```diff
@@ -78,27 +78,15 @@ Generated by [AVA](https://avajs.dev).
...
     `Error: Cannot find external module "missing" in package file://.../compartment-mapper/test/fixtures-error-handling/node_modules/cjs/␊
         at importHook (file://.../compartment-mapper/src/import-hook.js:…)␊
-        at loadWithoutErrorAnnotation (file://.../ses/src/module-load.js:…)␊
-        at memoizedLoadWithErrorAnnotation (file://.../ses/src/module-load.js:…)␊
-        at loadRecord (file://.../ses/src/module-load.js:…)␊
-        at loadWithoutErrorAnnotation (file://.../ses/src/module-load.js:…)␊
-        at async load (file://.../ses/src/module-load.js:…)␊
-        at async file://.../compartment-mapper/test/scaffold.js:…␊
-        at async file://.../compartment-mapper/test/scaffold.js:…`
+        at async loadWithoutErrorAnnotation (file://.../ses/src/module-load.js:…)`
```
naugtur, line 89: Not yet sure why these two stack traces changed.

packages/compartment-mapper/test/test-optional.js
```diff
@@ -31,15 +31,13 @@ scaffold(
       ['one', 'two'],
       'expected errors',
     );
-    t.true(
-      result.errors.one.message.startsWith(
-        `Cannot find external module "missing-one" in package`,
-      ),
+    t.regex(
+      result.errors.one.message,
+      /^Cannot find external module "missing-one" in package.*/,
```
naugtur, line 36: No behavioral change in this test, I just made failure messages more helpful (while these tests were not yet passing)

packages/compartment-mapper/test/test-error-handling.js
```diff
@@ -30,6 +30,12 @@ const onError = (t, { error, title }) => {
   } else {
     t.fail();
   }
+
+  // Skip snapshotting the 'both' case because esm/csj order is not deterministic 
+  // which causes the error to contain one or two causes.
+  if(title.match(/both/i)) {
+    return t.pass();
+  }
```
naugtur, line 38: The inconsistent behavior happens when importing the same specifier from both cjs and esm modules when there's no package matching the specifier.
The order in which we attempt importing dependencies is not guaranteed.

    If we require the missing package from cjs first, the error gets postponed to runtime, but it doesn't get cached so importing from an esm module errors out before the application runs and that's the single error.
    If we import esm first, the error happens right away and gets cached so the subsequent cjs require of the missing package gets the error right away instead of postponing to runtime.

packages/compartment-mapper/src/link.js
```diff
@@ -36,29 +35,6 @@ const { allSettled } = Promise;
...
-  {
-    resolveHook() {
-      return '';
-    },
-    async importHook() {
-      return inertStaticModuleRecord;
-    },
-  },
-).module('');
-
```
kriskowal, line 61: This is a nice deletion.

packages/compartment-mapper/src/import.js
```diff
@@ -70,22 +73,32 @@ export const loadLocation = async (readPowers, moduleLocation, options) => {
...
+      exitModuleImportHook,
+    });
     const makeImportHook = makeImportHookMaker(
       readPowers,
       packageLocation,
       undefined,
       compartmentMap.compartments,
-      undefined,
+      internalExitModuleImportHook,
+      false,
```
kriskowal, line 94: boolean trap

packages/compartment-mapper/test/test-error-handling.js
naugtur: It is fluctuating when running the same exact application via the same exact means. In both cases the error will prevent the application from working because an error is thrown and the only difference is in aggregation of errors - we either get one or two.  
An app with that issue does not reach runtime. 
The non-error case of this was a known issue when I approached the CJS support improvements - that Endo differs from other engines in unifying the module loaded from CJS and ESM into one. 
That unification has a gap - CJS require deferred error is wrapped in a module record and it makes it not-the-same as the errored record for the missing module that we got from importing.  
When we discussed deferredError, applying it to ESM was deemed a waste of good fail-fast and I agree with that.

I'll give it a second thought, but other than confusing the user with 1 or 2 "underlying errors" interchangeably I don't think this behavior hurts. Any result of it should not be accessible at runtime, because the ESM import errors out. The only thing I have to think about is if catching the error from require in CJS and reacting to it is possible.  
At worst, if all is spawned from CJS and both paths to the missing module are wrapped in separate try-catch it might be a source of 1 bit of randomness.

This PR did not introduce it though, I just added snapshotting to a test that was already reproducing the situation, but we didn't check in enough detail.
deferredError being applied to only the CJS and not ESM is the root cause.

packages/compartment-mapper/src/archive.js
```diff
@@ -293,19 +297,24 @@ const digestLocation = async (powers, moduleLocation, options) => {
...
+  });
+
   const makeImportHook = makeImportHookMaker(
     read,
     packageLocation,
     sources,
     compartments,
-    exitModules,
+    internalExitModuleImportHook,
+    true,
```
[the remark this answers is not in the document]
naugtur: I was very much tempted to refactor makeImportHookMaker to accept an object (I changed arguments list and structure 5 times before landing here) and I see I needlessly resisted. 
I'll happily change it.

packages/compartment-mapper/test/test-error-handling.js
naugtur: Quick note from today's meeting (recorded) 

- The same non-deterministic behavior could be observed in all-ESM case by inspecting the message from a rejected dynamic import promise for one of two nested missing packages to see which one failed first. Assuming the time-inconsistency comes from file load times or the host itself being non-deterministic in promise resolution at runtime.
- The current behavior is success-confluent
- I still intend to try to pin down the exact reason why the non-determinism is introduced
- Decreasing the likelihood of the condition occurring to less than now, but not zero is not worth pursuing.

packages/compartment-mapper/src/archive.js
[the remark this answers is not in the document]
naugtur: updated with a commit that fixes most issues bar one - with archive import hook maker. 
Today in the Endo meeting we decided to remove the ArchiveImportHookMaker type as it can't successfully be used to guard against anything.

packages/compartment-mapper/test/scaffold.js
```diff
@@ -84,6 +84,7 @@ export function scaffold(
     tags = undefined,
     searchSuffixes = undefined,
     commonDependencies = undefined,
+    additionalOptions = {},
```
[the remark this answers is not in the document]
naugtur: The problem this tries to solve is I wanted to avoid exposing more and more options from load/import functions up to the scaffold. I'm trying to make it possible to pass any options to all calls scaffold makes without adding too many new options to an already complex scaffold API.

packages/compartment-mapper/demo/policy/index.mjs
```diff
@@ -97,14 +93,21 @@ const options = {
     Buffer: ApiSubsetOfBuffer,
     console,
     process,
+    setTimeout,
+  },
+  exitModuleImportHook: async specifier => {
```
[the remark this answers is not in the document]
naugtur: Renamed in public API. Attempting to also rename references used internally was confusing, kept the mouthful of a name to differentiate from the importHook that compartment-mapper is creating as the scopes where the references to these two are handled overlap a bunch,

packages/compartment-mapper/src/archive.js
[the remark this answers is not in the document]
naugtur: That's also been updated now. The refactoring is done.

packages/compartment-mapper/src/policy.js
```diff
@@ -389,57 +395,60 @@ export const enforceModulePolicy = (specifier, compartmentDescriptor, info) => {
...
+  return freeze({
+    imports: originalModuleRecord.imports,
+    // It seems ok to declare the exports but then let the attenuator trim the values.
+    // Seems ok for attenuation to leave them undefined - accessing them is malicious behavior.
+    exports: originalModuleRecord.exports,
+    execute: (moduleExports, compartment, resolvedImports) => {
+      const ns = {};
+      originalModuleRecord.execute(ns, compartment, resolvedImports); // TODO: fix typing
+
+      // TODO: attenuator being async forces us to call original execute out of order before returning the record.
```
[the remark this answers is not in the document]
naugtur: this comment was outdated, there's a comment in the higher scope with a PR link

packages/compartment-mapper/src/link.js
naugtur: The observable result of this deletion and using an importHook is that modules that we pass by reference no longer need to pass the brand check in SES as they're wrapped in a record. 
As a result, passing the same exit module reference to two apps/assemblies result in them receiving the module namespace that's not the same reference. 
Note it would also be the case if using policies.

naugtur:
I believe this can be merged as-is. All feedback has been addressed. The issue with non-deterministic error has been explained away by Mark in the 2023/04/12 Endo meeting.

@kriskowal @kumavis @michaelfig Who wants to give me the rubber stamp?

michaelfig (approved):
LGTM!

I'm willing to approve on the condition that @kriskowal and @kumavis aren't requesting changes.

packages/compartment-mapper/src/import-archive.js
```diff
@@ -183,14 +225,20 @@ const makeFeauxModuleExportsNamespace = Compartment => {
   return compartment.module('.');
 };
 
+// Have to give it a name to capture the external meaning of Compartment
+// Otherwise @param {typeof COmpartment} takes the Compartment to mean
```
kriskowal, line 229: sp COmpartment

packages/compartment-mapper/src/import-archive.js
```diff
@@ -291,14 +346,26 @@ export const parseArchive = async (
...
+    const {
+      globals,
+      modules,
+      transforms,
+      __shimTransforms__,
+      Compartment,
+      importHook: exitModuleImportHook,
+    } = options || {};
+
+    const internalExitModuleImportHook = exitModuleImportHookMaker({
```
kriskowal, line 358: `internal` is less specific than naming the boundary we’re inside, which I believe is `compartment`.

packages/compartment-mapper/test/test-error-handling.js
kriskowal: I’m satisfied, so resolving.

kriskowal (approved):

PR #1512, opened 2023-03-17
fix(exo): fix exo-tools TSC problem

warner:
Running tsc in some other packages causes this unrelated line in packages/exo to start complaining. Removing the ts-expect-error caused it to fail tsc when run in packages/exo. Thanks to @michaelfig for a fix that appeased both cases.

Copied from https://github.com/Agoric/agoric-sdk/pull/7177


erights (approved):
LGTM, thanks.

PR #1513, opened 2023-03-17
feat(ses): finite deep stacks, on by default

erights:
The deep stacks continue to use the `assert.note` mechanism to associate an error with previous stack-capturing errors. The unbounded leak that caused us to turn deep stacks off was really the unbounded memory of error-note-graphs.

Prior to this PR, the line disabling deep stacks by default was commented
```js
// ... We intend to change the default after verifying that having
// the feature enabled in production does not cause memory to leak.
```

This PR makes the error-note storage into a finite budget LRU cache, defaulting to a budget of `1000`. With the memory leak fixed, `track-turns.js` is changed to enable deep stacks by default. Measurements will still be needed to see what the cost is, and what is a reasonable budget.

### DIFF
--- a/packages/ses/src/error/note-log-args.js
+++ b/packages/ses/src/error/note-log-args.js
@@ -8,6 +8,7 @@ const { isSafeInteger } = Number;
 
 /**
  * @typedef {object} DoublyLinkedCell
+ * A cell of a doubly-linked ring, i.e., a doubly-linked circular list.
  * DoublyLinkedCells are not frozen, and so should be closely encapsulated by
  * any abstraction that uses them.
  * @property {DoublyLinkedCell} next
@@ -18,7 +19,7 @@ const { isSafeInteger } = Number;
 
 /**
  * Makes a new self-linked cell. There are two reasons to do so:
- *    * To make the head sigil of a new initially-empty list
+ *    * To make the head sigil of a new initially-empty doubly-linked ring.
  *    * To make a non-sigil cell to be `spliceAfter`ed.
  *
  * @param {object} key
@@ -26,25 +27,24 @@ const { isSafeInteger } = Number;
  * @returns {DoublyLinkedCell}
  */
 const makeSelfCell = (key, value) => {
-  /** @type {DoublyLinkedCell} */
-  const selfCell = {
-    // @ts-expect-error will be fixed before return
+  /** @type {Partial<DoublyLinkedCell>} */
+  const incompleteCell = {
     next: undefined,
-    // @ts-expect-error will be fixed before return
     prev: undefined,
     key,
     value,
   };
-  selfCell.next = selfCell;
-  selfCell.prev = selfCell;
+  incompleteCell.next = incompleteCell;
+  incompleteCell.prev = incompleteCell;
+  const selfCell = /** @type {DoublyLinkedCell} */ (incompleteCell);
   // Not frozen!
   return selfCell;
 };
 
 /**
- * Splices a self-linked non-sigil cell into a list after `prev`.
+ * Splices a self-linked non-sigil cell into a ring after `prev`.
  * `prev` could be the head sigil, or it could be some other non-sigil
- * cell within a list.
+ * cell within a ring.
  *
  * @param {DoublyLinkedCell} prev
  * @param {DoublyLinkedCell} selfCell
@@ -70,7 +70,7 @@ const spliceAfter = (prev, selfCell) => {
 
 /**
  * @param {DoublyLinkedCell} cell
- * Must be a non-sigil part of a list, and therefore non-self-linked
+ * Must be a non-sigil part of a ring, and therefore non-self-linked
  */
 const spliceOut = cell => {
   const { prev, next } = cell;
@@ -103,7 +103,7 @@ export const makeLRUCacheMap = budget => {
   }
   /** @type {Map<object, DoublyLinkedCell>} */
   const map = new Map();
-  let count = 0; // count must remain <= budget
+  let size = 0; // `size` must remain <= `budget`
   // As a sigil, `head` uniquely is not in the `map`.
   const head = makeSelfCell(undefined, undefined);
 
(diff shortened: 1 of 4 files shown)


erights:
@kriskowal if you could have a look at this, that'd be great. CI is giving me a jackpot of errors, all of which seem rooted in

```
Error running main: TypeError: Cannot read property 'length' of undefined
    at main (file:///home/runner/work/endo/endo/packages/ses/scripts/bundle.js:28:46)
```
where the immediately relevant code in bundle.js is
```js
  const { code: terse } = terser.minify(bundle, {
    mangle: false,
    keep_classnames: true,
  });

  console.log(`Bundle size: ${bundle.length} bytes`);
  console.log(`Minified bundle size: ${terse.length} bytes`);
```

However, this PR itself is clearly completely independent of bundling. I don't see how it could be affected. Any ideas?

packages/ses/src/error/note-log-args.js
```diff
@@ -0,0 +1,216 @@
...
+    let cell = touchCell(key);
+    if (cell !== undefined) {
+      cell.value = value;
+    }
+    if (count >= budget) {
+      const condemned = head.prev;
+      spliceOut(condemned); // Drop least recently used
+      map.delete(condemned.key);
+      count -= 1;
+    }
```
gibson042, line 137: This won't work when `budget` is 0, which there are a few ways to fix.
* Require `budget` to be positive.
* Silently promote `budget` to be positive.
* Honor 0 by making `set` noöp.

packages/ses/src/error/note-log-args.js
```diff
@@ -0,0 +1,216 @@
...
+  const has = key => touchCell(key) !== undefined;
+  freeze(has);
+
+  const get = key => touchCell(key)?.value;
+  freeze(get);
+
+  const set = (key, value) => {
+    let cell = touchCell(key);
+    if (cell !== undefined) {
+      cell.value = value;
```
gibson042, line 130: This seems to be missing either an early `return` or a following `else`, without which a new cell will be created and added.

gibson042 (commented):
> The unbounded leak that caused us to turn deep stacks off was really the unbounded memory of error-note-graphs.

Huh, really? I would have expected `takeNoteLogArgsArray` to free that... was nothing calling it?

packages/ses/src/error/note-log-args.js
```diff
@@ -0,0 +1,216 @@
...
+  const head = makeSelfCell(undefined, undefined);
+
+  const touchCell = key => {
+    const cell = map.get(key);
+    if (cell === undefined) {
+      return undefined;
+    }
+    // Becomes most recently used
+    spliceOut(cell);
+    spliceAfter(head, cell);
```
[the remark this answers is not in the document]
erights: constant time LRU. Is there a way to do that with an array?

packages/ses/src/error/note-log-args.js
erights: Good point! I missed that completely.

packages/ses/src/error/note-log-args.js
erights: Good catch. Fixed.

packages/ses/src/error/note-log-args.js
```diff
@@ -0,0 +1,216 @@
...
+ */
+export const makeLRUCacheMap = budget => {
+  if (!isSafeInteger(budget) || budget < 0) {
+    throw new TypeError('budget must be a safe non-negative integer number');
+  }
+  /** @type {Map<object, DoublyLinkedCell>} */
+  const map = new Map();
+  let count = 0; // count must remain <= budget
+  // As a sigil, `head` uniquely is not in the `map`.
+  const head = makeSelfCell(undefined, undefined);
```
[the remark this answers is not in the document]
erights: Good point. Fixed.

erights:
> > The unbounded leak that caused us to turn deep stacks off was really the unbounded memory of error-note-graphs.
> 
> Huh, really? I would have expected `takeNoteLogArgsArray` to free that... was nothing calling it?

Only for errors that are actually reported. The fatal leak were for errors that could eventually be potentially reported, but were not.

erights:
PTAL

packages/ses/src/error/note-log-args.js
```diff
@@ -0,0 +1,219 @@
...
+ * @param {number} budget
+ * @returns {WeakMap<object,any>}
+ */
+export const makeLRUCacheMap = budget => {
+  if (!isSafeInteger(budget) || budget < 0) {
+    throw new TypeError('budget must be a safe non-negative integer number');
+  }
+  /** @type {Map<object, DoublyLinkedCell>} */
+  const map = new Map();
+  let count = 0; // count must remain <= budget
```
gibson042, line 106: Nit:
```suggestion
  let size = 0; // size must remain <= budget
```

packages/ses/src/error/note-log-args.js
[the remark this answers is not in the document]
gibson042: > constant time LRU. Is there a way to do that with an array?

Since the maximum size of the array would be bounded by `budget`, the answer to this question is _technically_ "yes". And it's possible that **`cache = [cache[i], ...cache.slice(0, i), ...cache.slice(i + 1)]`**[^1] or **`cache.splice(0, i + 1, cache[i], ...cache.slice(0, i))`**[^2] is faster than **`spliceOut(cell); spliceAfter(head, cell)`** in absolute terms for the budget we set even though it scales in a worse algorithmic time complexity class, or that it's slower but the difference doesn't matter so we could still go with simpler code.

[^1]: suitably protected against environment manipulation, e.g. `cache = [cache[i], ...arrayIterator(arraySlice(cache, 0, i)), ...arrayIterator(arraySlice(cache, i + 1))]`
[^2]: likewise (e.g., `arraySplice(cache, 0, i + 1, cache[i], ...arrayIterator(arraySlice(cache, 0, i)))`), and also with batching if `budget` can exceed maximum argument count

gibson042 (approved):

packages/ses/src/error/note-log-args.js
erights: Done

packages/ses/src/error/note-log-args.js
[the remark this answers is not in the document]
erights: I agree that is possible. Since it is a fully encapsulated decision, if measurements justify a change, we can always change it later.

Resolving.

erights:
I also changed the default budget from `10_000` to `1000` because, in the absence of measurement, I'd really like to avoid guessing too high. If too low, it is still better than the status quo, with deep stacks turned off. And we can always increase it later.

We should also consider configuring the budget with an environment option.

erights:
Worked around stale tooling. Bug at https://github.com/endojs/endo/issues/1514 . (@kriskowal , thanks!)

erights:
Re https://github.com/endojs/endo/pull/1513#issuecomment-1474560497

I reduced the note-args budget again, this time to `100`. At `1000` I ran into one problem:

```js
// Bizarrely, if set to 1000 or 10_000, captp/test/test-gc.js
// test 'test loopback gc' fails ONLY ON WINDOWS.
// TODO: Figure out why, and either special case or fix.
// TODO: Once captp/test/test-gc.js doesn't stop us,
// set this default back to a more reasonable number.
// TODO: make this default configurable by an environment option.
const defaultLogArgsBudget = 100;
```

erights:
Given the nature of this PR, before merging, it was checking the effect on performance. The following first crude test is consistent enough to show we're ok, or even a bit better!

https://github.com/endojs/endo/pull/1515 , identical to current master, has the following timings from CI

image
image

Compare with the timings of this PR under CI. Bizarrely, this PR is always the same or faster. (ignoring the failed test)

image
image

erights:
At `100` the same test now fails or passes intermittently, which is even worse!

erights:
Re https://github.com/endojs/endo/pull/1513#issuecomment-1474626631 , @1516 fixed the immediate problem, the OS-dependent behavior. However, this PR still changes the counts, for reasons we (@michaelfig and I) do not yet precisely understand but believe are fine for now.

@mfig, I changed those lines as we discussed but also added a TODO

```js
  // TODO(mfig,erights): explain why #1513 changed these counts from 3 to 2
  t.is(getFarStats().sendCount.CTP_DROP, 2);
  t.is(getNearStats().recvCount.CTP_DROP, 2);
```

PR #1519, opened 2023-03-21
Weak LRU Cache Map

michaelfig:
Refs #1513 

I tracked down the object leak to [where cause stacks are being attached to a given eventual-send](https://github.com/endojs/endo/blob/aae0e57f7a6bdcc898396c65ec22616a33672d32/packages/eventual-send/src/track-turns.js#L116). This pointed to a strongly-held chain of errors and their details arguments.

This weakifying of the LRUCacheMap uses an outer WeakMap instead of the strong map, and then a *unique* WeakMap instance **per non-condemned entry in the cache**.  That is, given N entries in the cache, 1+N WeakMaps in total (!).  This is because WeakMaps don't support composite keys, and we are avoiding using WeakRefs in the Endo implementation.

The benefit is that we no longer have a leak.  Now the CapTP GC drop counts are the same as before #1513 (3 drop messages instead of 2 for `packages/captp/test/test-gc.js`).

### DIFF
--- a/packages/ses/src/error/note-log-args.js
+++ b/packages/ses/src/error/note-log-args.js
@@ -1,3 +1,4 @@
+// @ts-check
 /* eslint-disable @endo/no-polymorphic-call */
 /* eslint-disable no-restricted-globals */
 
@@ -7,14 +8,14 @@ const { freeze } = Object;
 const { isSafeInteger } = Number;
 
 /**
+ * @template Data
  * @typedef {object} DoublyLinkedCell
  * A cell of a doubly-linked ring, i.e., a doubly-linked circular list.
  * DoublyLinkedCells are not frozen, and so should be closely encapsulated by
  * any abstraction that uses them.
- * @property {DoublyLinkedCell} next
- * @property {DoublyLinkedCell} prev
- * @property {object} key
- * @property {any} value;
+ * @property {DoublyLinkedCell<Data>} next
+ * @property {DoublyLinkedCell<Data>} prev
+ * @property {Data} data
  */
 
 /**
@@ -22,21 +23,20 @@ const { isSafeInteger } = Number;
  *    * To make the head sigil of a new initially-empty doubly-linked ring.
  *    * To make a non-sigil cell to be `spliceAfter`ed.
  *
- * @param {object} key
- * @param {any} value
- * @returns {DoublyLinkedCell}
+ * @template Data
+ * @param {Data} data
+ * @returns {DoublyLinkedCell<Data>}
  */
-const makeSelfCell = (key, value) => {
-  /** @type {Partial<DoublyLinkedCell>} */
+const makeSelfCell = data => {
+  /** @type {Partial<DoublyLinkedCell<Data>>} */
   const incompleteCell = {
     next: undefined,
     prev: undefined,
-    key,
-    value,
+    data,
   };
-  incompleteCell.next = incompleteCell;
-  incompleteCell.prev = incompleteCell;
-  const selfCell = /** @type {DoublyLinkedCell} */ (incompleteCell);
+  const selfCell = /** @type {DoublyLinkedCell<Data>} */ (incompleteCell);
+  selfCell.next = selfCell;
+  selfCell.prev = selfCell;
   // Not frozen!
   return selfCell;
 };
@@ -46,8 +46,9 @@ const makeSelfCell = (key, value) => {
  * `prev` could be the head sigil, or it could be some other non-sigil
  * cell within a ring.
  *
- * @param {DoublyLinkedCell} prev
- * @param {DoublyLinkedCell} selfCell
+ * @template Data
+ * @param {DoublyLinkedCell<Data>} prev
+ * @param {DoublyLinkedCell<Data>} selfCell
  */
 const spliceAfter = (prev, selfCell) => {
   if (prev === selfCell) {
@@ -69,14 +70,12 @@ const spliceAfter = (prev, selfCell) => {
 };
 
 /**
- * @param {DoublyLinkedCell} cell
- * Must be a non-sigil part of a ring, and therefore non-self-linked
+ * @template Data
+ * @param {DoublyLinkedCell<Data>} cell
+ * No-op if the cell is self-linked.
  */
 const spliceOut = cell => {
   const { prev, next } = cell;
-  if (prev === cell || next === cell) {
-    throw TypeError('Expected non-self-linked cell');
-  }
   prev.next = next;
   next.prev = prev;
   cell.prev = cell;
@@ -87,15 +86,19 @@ const spliceOut = cell => {
  * The LRUCacheMap is used within the implementation of `assert` and so
  * at a layer below SES or harden. Thus, we give it a `WeakMap`-like interface
  * rather than a `WeakMapStore`-like interface. To work before `lockdown`,
- * the implementation must use `freeze` manually, but still exhausively.
+ * the implementation must use `freeze` manually, but still exhaustively.
  *
- * It does not hold onto anything weakly. The only sense in which it is
- * WeakMap-like is that it does not support enumeration.
+ * It implements the WeakMap interface, and holds its keys weakly.  Cached
+ * values are only held while the key is held by the user and the key/value
+ * bookkeeping cell has not been pushed off the end of the cache by `budget`
+ * number of more recently referenced cells.  If the key is dropped by the user,
+ * the value will no longer be held by the cache, but the bookkeeping cell
+ * itself will stay in memory.
  *
- * TODO: Make parameterized `Key` and `Value` template types
- *
+ * @template {{}} K
+ * @template {unknown} V
  * @param {number} keysBudget
- * @returns {WeakMap<object,any>}
+ * @returns {WeakMap<K,V>}
  */
 export const makeLRUCacheMap = keysBudget => {
   if (!isSafeInteger(keysBudget) || keysBudget < 0) {
@@ -103,15 +106,18 @@ export const makeLRUCacheMap = keysBudget => {
       'keysBudget must be a safe non-negative integer number',
     );
   }
-  /** @type {Map<object, DoublyLinkedCell>} */
-  const map = new Map();
+  /** @typedef {DoublyLinkedCell<WeakMap<K, V> | undefined>} LRUCacheCell */
+  /** @type {WeakMap<K, LRUCacheCell>} */
+  const keyToCell = new WeakMap();
   let size = 0; // `size` must remain <= `keysBudget`
-  // As a sigil, `head` uniquely is not in the `map`.
-  const head = makeSelfCell(undefined, undefined);
+  // As a sigil, `head` uniquely is not in the `keyToCell` map.
+  /** @type {LRUCacheCell} */
+  const head = makeSelfCell(undefined);
 
   const touchCell = key => {
-    const cell = map.get(key);
-    if (cell === undefined) {
+    const cell = keyToCell.get(key);
+    if (cell === undefined || cell.data === undefined) {
+      // Either the key was GCed, or the cell was condemned.
       return undefined;
     }
     // Becomes most recently used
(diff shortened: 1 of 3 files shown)


packages/ses/src/error/note-log-args.js
```diff
@@ -89,26 +84,35 @@ const spliceOut = cell => {
  * rather than a `WeakMapStore`-like interface. To work before `lockdown`,
  * the implementation must use `freeze` manually, but still exhausively.
  *
- * It does not hold onto anything weakly. The only sense in which it is
- * WeakMap-like is that it does not support enumeration.
+ * It implements the WeakMap interface, and holds its keys and values only
```
mhofman, line 87: I think "and values" is erroneous ? A WeakMap only holds the key weakly, and simulates a strong link between the key and the value.

michaelfig:
@mhofman, I implemented your suggested "composite-key-like" solution.  Tested successfully locally that deep stacks work, and enabling/disabling them does not change CapTP's GC behaviour.

PTAL when you have time.  This is not urgent.

packages/ses/src/error/note-log-args.js
```diff
@@ -7,36 +8,37 @@ const { freeze } = Object;
...
  * A cell of a doubly-linked ring, i.e., a doubly-linked circular list.
  * DoublyLinkedCells are not frozen, and so should be closely encapsulated by
  * any abstraction that uses them.
- * @property {DoublyLinkedCell} next
- * @property {DoublyLinkedCell} prev
- * @property {object} key
- * @property {any} value;
+ * @property {DoublyLinkedCell<K, V>} next
+ * @property {DoublyLinkedCell<K, V>} prev
+ * @property {WeakMap<object, V>} keyValue
```
mhofman, line 18: I assume `K` is for key?

```suggestion
 * @property {WeakMap<K, V>} keyValue
```

packages/ses/src/error/note-log-args.js
```diff
@@ -7,36 +8,37 @@ const { freeze } = Object;
...
 
 /**
  * Makes a new self-linked cell. There are two reasons to do so:
  *    * To make the head sigil of a new initially-empty doubly-linked ring.
  *    * To make a non-sigil cell to be `spliceAfter`ed.
  *
- * @param {object} key
- * @param {any} value
- * @returns {DoublyLinkedCell}
+ * @template K,V
```
mhofman, line 26: Might be cleaner to constraint the `K` parameter?

```suggestion
 * @template {object} K
 * @template V
```

packages/ses/src/error/note-log-args.js
```diff
@@ -89,26 +93,31 @@ const spliceOut = cell => {
...
- * It does not hold onto anything weakly. The only sense in which it is
- * WeakMap-like is that it does not support enumeration.
+ * It implements the WeakMap interface, and holds its keys only weakly.
+ * However, the spine of the doubly-linked ring (up to `budget` cells in length)
+ * is held strongly.
  *
- * TODO: Make parameterized `Key` and `Value` template types
+ * It may pin more than `budget` entries in memory if the user of the
+ * LRUCacheMap holds a key strongly, or the garbage collector has not yet
+ * emptied the unreferenced keys from the internal weakmaps.
```
mhofman, line 102: I'd be more precise here and say that some internal data may be left associated with a key even after it falls off the cache, but the values will not be held unless both the key is held by the user and the key hasn't fallen off the cache.

packages/ses/src/error/note-log-args.js
```diff
@@ -118,33 +127,42 @@ export const makeLRUCacheMap = budget => {
...
   const set = (key, value) => {
     if (budget >= 1) {
       let cell = touchCell(key);
       if (cell !== undefined) {
-        cell.value = value;
+        cell.keyValue.set(key, value);
       } else {
         if (size >= budget) {
           const condemned = head.prev;
           spliceOut(condemned); // Drop least recently used
```
mhofman, line 160: We need to also null out the `keyValue` property of the condemned cell. If the key is kept around, so will the cell through `keyToCell`, even though the cell is unlinked from others.

It might then make sense to let `set` create this `WeakMap` and set it on the cell on insertion below (and keep `makeSelfCell` agnostic of the data kept in the cell.

packages/ses/src/error/note-log-args.js
```diff
@@ -118,33 +127,42 @@ export const makeLRUCacheMap = budget => {
...
+   * @param {K} key
+   */
   // TODO Change to the following line, once our tools don't choke on `?.`.
   // See https://github.com/endojs/endo/issues/1514
-  // const get = key => touchCell(key)?.value;
+  // const get = key => touchCell(key)?.keyValue.get(key);
   const get = key => {
     const cell = touchCell(key);
-    return cell && cell.value;
+    return cell && cell.keyValue.get(key);
```
mhofman, line 144:
```suggestion
    return cell && cell.keyValue && cell.keyValue.get(key);
```

mhofman (changes requested):
There is still a leak of an associated value if the cell falls off the cache, but the key is kept around.

packages/ses/src/error/note-log-args.js
michaelfig: Can't do this, because...

mhofman: Yeah given the comment on `set` below, We probably should rename this to `data` with a type of `T` type param solely defined by the user of the linked list.

packages/ses/src/error/note-log-args.js
michaelfig: I'd love to constrain, but this syntax (which is the only one that could conceivably do the right thing) does not do the right thing.  Without `noImplicitAny: true`, it does *not* produce `<K extends object, V>`, it results in just `<K extends unknown, V>` which causes all kinds of issues with trying to use it as `WeakMap<K, V>`.

So, I won't tighten up the types, but I will add a comment to that effect.

mhofman: ah yeah I wish we could switch to  `noImplicitAny`...

Anyway, probably moot given my other suggestion to drop this altogether.

packages/ses/src/error/note-log-args.js
```diff
@@ -110,11 +109,12 @@ export const makeLRUCacheMap = budget => {
   if (!isSafeInteger(budget) || budget < 0) {
```
mhofman, line 107: Add a typedef here for the Cell

```js
  /** @typedef {DoublyLinkedCell<WeakMap<K, V> | undefined>} LRUCacheCell */
```

packages/ses/src/error/note-log-args.js
```diff
@@ -151,20 +151,24 @@ export const makeLRUCacheMap = budget => {
...
           size -= 1;
         }
         size += 1;
-        cell = makeSelfCell(key, value);
+        cell = makeSelfCell(new WeakMap());
         keyToCell.set(key, cell);
         spliceAfter(head, cell); // start most recently used
       }
+      if (!cell.data) {
+        throw TypeError('Internal error; expected cell.data');
```
mhofman, line 167: Actually this could totally happen when a key is pushed out then later re-introduced. We should simply reset `keyToCell` like if the cell was not found.

mhofman (changes requested):
Some typing nits, but I think we need to handle the `set` operation where the `cell` exists but `cell.data` was nulled out.

We probably need unit test for all this eviction logic.

packages/ses/src/error/note-log-args.js
```diff
@@ -118,49 +126,69 @@ export const makeLRUCacheMap = budget => {
...
   freeze(get);
 
+  /**
+   * @param {K} key
+   * @param {V} value
+   */
   const set = (key, value) => {
     if (budget >= 1) {
+      /** @type {LRUCacheCell | undefined} */
       let cell = touchCell(key);
```
mhofman, line 154: Actually I just realized `spliceOut` will also throw if the key was previously pushed out since the cell will be self-linked.

mhofman (changes requested):

packages/ses/src/error/note-log-args.js
```diff
@@ -118,49 +123,69 @@ export const makeLRUCacheMap = budget => {
...
-    const cell = map.get(key);
+    const cell = keyToCell.get(key);
     if (cell === undefined) {
       return false;
     }
     spliceOut(cell);
-    map.delete(key);
+    cell.data = undefined;
+    keyToCell.delete(key);
     size -= 1;
```
mhofman, line 189: if `cell.data` was already undefined, we should not decrement the size as it was already condemned.

mhofman (changes requested):
Sorry I found one last issue on deletion: a budget underflow.

packages/ses/src/error/note-log-args.js
```diff
@@ -118,49 +123,69 @@ export const makeLRUCacheMap = budget => {
...
           size -= 1;
         }
         size += 1;
-        cell = makeSelfCell(key, value);
-        map.set(key, cell);
+        cell = makeSelfCell(undefined);
         spliceAfter(head, cell); // start most recently used
       }
+      if (!cell.data) {
+        // Either a new cell or a condemned cell.  Add its data.
```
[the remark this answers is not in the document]
mhofman: I think we technically can just make the cell and insert if missing above, and move the size check in this section.

The only remaining edge case which the previous version doesn't handle either is if `budget === 0`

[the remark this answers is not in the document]
michaelfig: `budget < 1` is handled by the topmost `if` condition in this function, making `.set` a no-op.  I've made that code clearer now.

packages/ses/src/error/note-log-args.js
```diff
@@ -118,49 +123,78 @@ export const makeLRUCacheMap = budget => {
...
+      cell.data = new WeakMap();
+      // Advertise the cell for this key.
+      keyToCell.set(key, cell);
+      while (size >= budget) {
+        const condemned = head.prev;
+        spliceOut(condemned); // Drop least recently used
+        condemned.data = undefined;
+        size -= 1;
       }
+      size += 1;
```
mhofman, line 170: This is totally on me, but it would make more sense to move `size += 1` at the top of this `if` block since we've already inserted the cell, and change the `size >= budget` check to `size > budget`.

erights (commented):
> Would love to see a review from @erights to know if we got it right.

If you two come to be confident that this solution is right, don't want long for me to chime in. Feel free to proceed anyway. But do become confident first ;)

My only hurry is that, if this PR is right, I'd like to see it in the next endo release rather than my unfixed change.

packages/ses/src/error/note-log-args.js
```diff
@@ -1,3 +1,4 @@
+// @ts-check
```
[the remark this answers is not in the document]
michaelfig: It's opt-in, and `packages/ses/tsconfig.json` doesn't opt in.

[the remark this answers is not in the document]
erights: Hard to develop habits if we're not consistent. Any reason we don't opt in everywhere?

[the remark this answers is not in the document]
michaelfig: The only reason is that we'd have to fix all the type errors before we can land a PR to opt in everywhere.  I don't know the scope of that, nor how long that would take.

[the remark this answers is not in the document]
erights: Or to start by opting out everywhere that got broken by implicitly opting in.

mhofman:
> If you two come to be confident that this solution is right, don't want long for me to chime in. Feel free to proceed anyway. But do become confident first ;)

I am pretty confident it is sound. I will be 100% confident once we have unit tests for the edge cases that I raised during review: cache eviction then re-insertion or deletion, and making sure these do not unexpectedly affect the cache size.

Testing cache size / eviction behavior is hard to test, as it can only be done by inference: check that inserting a new entry results in the tail entry to be evicted or not (if at capacity), or that a further insertion causes an eviction or not).

Issue #1514, opened 2023-03-18
Some parsers in our toolchain choke on "?."


erights:
See https://github.com/endojs/endo/pull/1513#issuecomment-1473061658

Separately, @kriskowal sent me 

> The length undefined error means minify failed and returned a result struct with error info instead of the code. Probably a syntax error, probably a stale version of the minifier.

That was the clue I needed. The new code in this PR contained a "?.". Once refactored to avoid any such newer syntax, CI is much happier.

turadg:
Terser didn't support optional chaining until 5.2: https://github.com/terser/terser/blob/master/CHANGELOG.md#v520

The repo currently imports 4.8.1, the latest without hitting the breaking changes of 5.0: https://github.com/terser/terser/blob/master/CHANGELOG.md#v500-beta0

michaelfig:
Notable culprits:
  - Node 12 in the platform-compatibility-test,
  - webpack 4.46.0 in the integration-test

kriskowal:
We have since added a lint rule forbidding optional chaining. We discovered that `esm` in the Agoric SDK testnet load generator used a `node -r esm` deploy script and failed with optional chaining.

Closing this issue as addressed, though we hope to obviate the need and drop support for some of these tests and tools.

PR #1520, opened 2023-03-21
chore(deps): bump Terser 4.8 to 5.16

turadg:
closes: #1514

# Description

https://github.com/terser/terser/blob/master/CHANGELOG.md#v500-beta0 shows these breaking changes,
- BREAKING: minify() is now async and rejects a promise instead of returning an error.
- BREAKING: Internal AST is no longer exposed, so that it can be improved without releasing breaking changes.
- BREAKING: Lowest supported node version is 10
- BREAKING: There are no more warnings being emitted

# Test plan

Anything more than CI?

### DIFF
--- a/packages/ses/package.json
+++ b/packages/ses/package.json
@@ -74,7 +74,7 @@
     "eslint-plugin-import": "^2.27.5",
     "prettier": "^2.8.5",
     "sinon": "8.0.4",
-    "terser": "^4.8.0",
+    "terser": "^5.16.6",
     "tsd": "^0.24.1",
     "typescript": "~4.9.5"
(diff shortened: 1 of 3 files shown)


mhofman:
I would expect https://github.com/endojs/endo/blob/7be31c30a20f4203f72b3f0b4cc5e6b8384569fc/packages/ses/scripts/bundle.js#L22 would need an `await` at the very least. I suppose that file is not tested in endo itself (it should be used in agoric-sdk). It's the only one I could find that relied on `terser`.

erights (approved):
LGTM, thanks!

PR #1523, opened 2023-03-22
doc: add issue link for bogus proxy trap observation

### LINKED ISSUE #1510: Still getting bogus scope proxy traps under debugger (vscode, node, v8)
The following screenshot is what happened when I was in the debugger and hovered over the variable `policyOutput`. It is unclear whether this should be considered a v8 bug because the spec must be largely silent on what can happen only during debugging. But this does indicate that we should should make some scope proxy more tolerant of surprising traps.

What's curious is that the stack reported for this error is the suspended stack of the program being debugged, not a stack within the debugger. I can see how this makes sense, but it is surprising. If I had not noticed when this happened, I could well have mistaken this to indicate a bug in the program being debugged.

image

erights:
Investigating https://github.com/endojs/endo/issues/1510 , I see that the code already had a proxy trap to deal with this case, and the trace I observed was correctly generated by a `console.warn` in that code. The issue adds a more specific observation of this anomaly. With this PR, that issue can be closed but remain linked as an explanation.

Closed #1510

### DIFF
--- a/packages/ses/src/strict-scope-terminator.js
+++ b/packages/ses/src/strict-scope-terminator.js
@@ -58,6 +58,7 @@ const scopeProxyHandlerProperties = {
   },
 
   // Chip has seen this happen single stepping under the Chrome/v8 debugger.
+  // See also https://github.com/endojs/endo/issues/1510
   // TODO record how to reliably reproduce, and to test if this fix helps.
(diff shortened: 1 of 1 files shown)


naugtur (approved):
I think the writeup in 1510 fulfills or partially fulfills the first TODO and makes the comment in line 60 less relevant.

PR #1529, opened 2023-03-29
test(compartment-mapper): snapshot errors from missing module loading test

naugtur:
The purpose of this is to document the change in stack trace for import failures in esm and cjs in more detail.
It should allow us to be more careful with errors happening at the right stage of archive->import->execute staying where they belong.


packages/compartment-mapper/test/test-error-handling.js
```diff
@@ -30,6 +30,7 @@ const onError = (t, { error, title }) => {
   } else {
     t.fail();
   }
```
naugtur, line 32: Wile the assertion above is redundant now, I kept it as explanation of the purpose of the test. Could be replaced with a long comment block potentially.

packages/compartment-mapper/test/scaffold.js
```diff
@@ -17,6 +17,12 @@ import { makeReadPowers } from '../src/node-powers.js';
 
 export const readPowers = makeReadPowers({ fs, crypto, url });
 
+export const sanitizePaths = (text = '', tolerateLineChange = false) => {
+  if (tolerateLineChange) {
+    text = text.replace(/:[0-9]+:[0-9]+/g, ':…');
+  }
+  return text.replace(/file:\/\/[^'"\n]+\/endo\/packages\//g, 'file://.../');
```
michaelfig, line 24: Please don't assume the repository has been checked out in a directory called `endo`, but since the `packages` directory is within the repo, I think it's fine to use it as the indicator of a path.

```suggestion
  return text.replace(/file:\/\/[^'"\n]+\/packages\//g, 'file://.../');
```

michaelfig (approved):
LGTM, modulo the little change in path sanitisation.

PR #1531, opened 2023-03-31
Make latest Endo usable by Agoric SDK

michaelfig:
(Only exists to provide a basis for Agoric/agoric-sdk#7273 to be tested before the pieces have fully landed.)

This PR won't be merged, but has been split into separate independent pieces for easier review:

- [x] #1519
- [x] #1544 
- [x] #1545 
- [x] #1546

### DIFF
--- a/packages/captp/test/test-gc.js
+++ b/packages/captp/test/test-gc.js
@@ -21,6 +21,35 @@ test('test loopback gc', async t => {
 
   await isolated(t, makeFar);
   await gcAndFinalize();
-  t.is(getFarStats().sendCount.CTP_DROP, 3);
-  t.is(getNearStats().recvCount.CTP_DROP, 3);
+
+  // Check the GC stats.
+  const nearStats = getNearStats();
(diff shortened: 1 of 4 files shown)


packages/captp/test/test-gc.js
```diff
@@ -21,6 +21,12 @@ test('test loopback gc', async t => {
 
   await isolated(t, makeFar);
   await gcAndFinalize();
-  t.is(getFarStats().sendCount.CTP_DROP, 3);
-  t.is(getNearStats().recvCount.CTP_DROP, 3);
+
+  // Check the GC stats.
+  const { sendCount: nearSend, recvCount: nearRecv } = getNearStats();
```
turadg, line 26: here's a way to test this without more bindings:
```js
t.like(getNearStats(), {
  sendCount: {
    CTP_DROP: 4,
  },
  recvCount: {
    DROPPED: 2,
  },
};
```

michaelfig: Thanks!

PR #1532, opened 2023-04-01
fix: sync with shadows in agoric-sdk

erights:
Some code from the `@agoric/store` package was unbundled and migrated to the `@endo/patterns` and `@endo/exo`  packages. However, until agoric-sdk is upgraded to depend on the endo with these migrated packages, it continues to use its own deprecated copies, which we be removed after this upgrade happens. To smooth the way for this, we try to keep the code in these packages in sync with their shadows in agoric-sdk.

However, since the last sync there has been version drift. This PR brings them back into alignment.

Some of this may be redundant with https://github.com/endojs/endo/pull/1531 . This PR is draft until these redundancies or possible conflicts are examined.

As explained at https://github.com/Agoric/agoric-sdk/pull/7138#issuecomment-1482387387 , https://github.com/Agoric/agoric-sdk/pull/7138 makes an incompat change to the exo-tools API. Since this is not yet merged, there will need to be a separate PR to sync `@endo/exo` with it.


warner (commented):
I don't have strong opinions on this one.

dckc (commented):
I'm struggling to find the beginning of this story.

@turadg maybe you're more familiar?

packages/exo/src/exo-makers.js
```diff
@@ -118,24 +140,24 @@ export const defineExoClassKit = (
...
-      if (finish) {
-        finish(context);
-      }
+    if (finish) {
+      // @ts-expect-error `facets` was added
+      finish(context);
     }
     return facets;
   };
+  // eslint-disable-next-line @typescript-eslint/prefer-ts-expect-error -- different per package https://github.com/Agoric/agoric-sdk/issues/4620
```
turadg, line 149: does the Endo repo suffer from this too?

erights:
> I'm struggling to find the beginning of this story.
> 
> @turadg maybe you're more familiar?

#1532 is just repairing version drift that happened because the agoric-sdk shadow copies were updated without coordinated updates to the primary copies in endo. (edited)

turadg (approved):

packages/exo/src/exo-tools.js
```diff
@@ -235,6 +235,7 @@ export const defendPrototype = (
       methodGuards && methodGuards[prop],
     );
   }
+
```
dckc, line 238: This doesn't look right. When I compare endo and agoric-sdk, I get substantive diffs in this file:

```
~/projects/endo
17:21 connolly@bldbox$ diff -u packages/exo/src/exo-tools.js ~/projects/agoric-sdk/packages/store/src/patterns/exo-tools.js 
--- packages/exo/src/exo-tools.js       2023-04-03 17:19:54.297144138 -0500
+++ /home/connolly/projects/agoric-sdk/packages/store/src/patterns/exo-tools.js 2023-04-01 14:03:32.978861672 -0500
@@ -1,18 +1,12 @@
 import { E, Far } from '@endo/far';
-import { listDifference, objectMap, mustMatch, M } from '@endo/patterns';
+import { listDifference, objectMap } from '@agoric/internal';
+
+import { MinMethodGuard, mustMatch, M } from './patternMatchers.js';
 
 const { quote: q, Fail } = assert;
 const { defineProperties } = Object;
 
-/**
- * A method guard, for inclusion in an interface guard, that enforces only that
- * all arguments are passable and that the result is passable. (In far classes,
- * "any" means any *passable*.) This is the least possible enforcement for a
- * method guard, and is implied by all other method guards.
- */
-const MinMethodGuard = M.call().rest(M.any()).returns(M.any());
-
 const defendSyncArgs = (args, methodGuard, label) => {
   const { argGuards, optionalArgGuards, restArgGuard } = methodGuard;
   const paramsPattern = M.splitArray(
@@ -235,6 +229,7 @@
       methodGuards && methodGuards[prop],
     );
   }
+
   return Far(tag, /** @type {T} */ (prototype));
 };
 harden(defendPrototype);
```

packages/patterns/src/types.js
```diff
@@ -493,10 +493,17 @@
 
 /**
  * @typedef {any} MethodGuardMaker
- * a parameter list like foo(a, b, c = d, …e) => f should be guarded by
- * something like
- * foo: M.call(AShape, BShape).optional(CShape).rest(EShape).returns(FShape)
- * optional is for optional (=) params. rest is for … (varargs) params
+ * A method name and parameter/return signature like:
+ * ```js
```
dckc, line 497: I'm not sure what agoric-sdk file the corresponds to, so I'm not sure how to verify that this is just syncing up.

dckc (commented):
doesn't look like just syncing up, to me

packages/exo/src/exo-makers.js
erights: Just tested and it does not. I have no idea why they differ and I find it disturbing that they do. Nevertheless, in response I removed it, so I'll count this suggestion as "fixed". Resolving.

packages/exo/src/exo-tools.js
erights: In agoric-sdk, `MinMethodGuard` is created and exported by `patternMatchers.js` and imported into `exo-tools.js`.

In endo, `MinMethodGuard` is still created and exported by `patternMatchers.js`, and it is still needed by `exo-tools.js`. However, now these are in separate packages. `patternMatchers.js` is in the `@endo/patterns` package, which does not export `MinMethodGuard`, so it would be harder for `exo-tools.js` to import it from there.

Repairs could be
   * revise `@endo/patterns` to directly export `MinMethodGuard`. But it seems to specialized to treat it as a general export of the package
   * revive `@endo/patterns` so that it is separately exported, via an `"exports":` in the package.json.
   * just redundantly define it is `exo-tools.js`, which is the *only* place it is needed.

I elected to do the last. But, with your question provoking me to discover that *only* exo-tools.js uses it, I am about to remove its redundant definition from the endo `patternMatchers.js`.

To keep the repositories in sync, I will also create a separate PR for agoric-sdk that moves `MinMethodGuard` from `patternMatchers.js` to `exo-tools.js`.

erights: Done here, and opened https://github.com/Agoric/agoric-sdk/pull/7320 . Resolving.

packages/patterns/src/types.js
erights: https://github.com/Agoric/agoric-sdk/blob/841a7012831166f6c29f00bdffc907f2b35a7b0c/packages/store/src/types.js#L754-L767

Resolving.

PR #1534, opened 2023-04-01
fix: sync with exo API change

erights:
Sync with https://github.com/Agoric/agoric-sdk/pull/7304 

To cope with the https://github.com/Agoric/agoric-sdk/pull/7138 API change, so that the next published endo can be compat with the agoric-sdk upgraded to depend on it.


warner (approved):
sounds good to me.. I don't know the broader Endo environment, but this seems to match what I need from the new swingset PR

erights:
Since I got this approved before https://github.com/endojs/endo/pull/1532 , I unstaged on #1532 and am merging separately.

Issue #1331, opened 2022-10-18
Support passable immutable ArrayBuffer as copy-data


kriskowal:
*2024-10-18:*

The plan of record for passable immutable binary data is an immutable `ArrayBuffer` per [our proposal to TC39](https://github.com/tc39/proposal-immutable-arraybuffer).

---
*2022-10-18:*

Our current best solution for carrying bytes is a Uint8Array, which we use pervasively for hardened streams. This is one of the reasons that we treat Typed Arrays like Map for the purposes of hardening. The content is mutable but the API surface is tamper-proof. However, the content is readable and writable by both the sender and receiver over time, so like a Map, TypedArrays are not passable. That makes them a poor foundation for remote streams in an ocap discipline.

I propose that we need an analogous solution to the `CopyMap` and `MapStore` for transmission of immutable byte strings, tentatively `CopyBytes` and `BytesStore`, which would be passable and provide a firmer foundation for streams (async iterators of byte strings). `CopyBytes` should mesh well with TypedArrays, particularly Uint8Array and ArrayBuffer. To that end, it may be necessary for them to be Proxies, such that `uint8Array.set(copyBytes)`  is sensible. Constructing an immutable copy of the content of an array-like should be possible, like `copyBytes(uint8Array)` or `makeCopyBytes([0, 1, 2, …, 127])`. `CopyBytes` would be backed by a TypedArray but would attenuate mutability. Slicing `CopyBytes` should be trivial. As with glyph strings, ropes are a possible optimization in the fullness of time.

As a critical optimization, we often use ring buffers for byte streams. We could conceivably also support another type for revocable copy bytes, that would permit a stream to reclaim the underlying byte range when the consumer signals it’s ready for another subrange.

mhofman:
I'm extremely concerned by the performance overhead of such a user-land abstraction. We should pursue a standard approach which allows engines to avoid array buffer copies, and minimal overhead on read access.

One such proposal at stage 1 is @Jack-Works's [Limited ArrayBuffer](https://github.com/tc39/proposal-limited-arraybuffer)

kriskowal:
Shimming a standards track solution would be good.

Ah, [memories](https://wiki.commonjs.org/wiki/Binary/D). (My abandoned hobbies include a decades old fantasy of JavaScript having byte strings.)

erights:
Attn @phoddie

phoddie:
[...]

PR #1538, opened 2023-04-04
feat(pass-style,marshal): ByteArray, a new binary Passable type

erights:
Closes: #1331 
Refs: https://github.com/ocapn/ocapn/issues/5#issuecomment-1492778252 #2414 

## Description

Recognize Immutable ArrayBuffers, when also frozen, as the new pass-style `byteArray`. Add a branch for each of the dispatches on pass-style that must now understand byte-arrays as well. As of this PR, many of those additional branches are stubbed out with "not yet implemented" errors as placeholders.

### Security Considerations

The underlying shim implementation, if we're still using that, uses a normal mutable ArrayBuffer internally, but safely encapsulates it so nothing outside the shim can mutate it.

### Scaling Considerations

The underlying shim implementation helps us handle bulk binary data in a largely no-copy or minimal-copy manner.

Will add a new pass-style, so everywhere we enumerate all the pass-styles, we'll need to add byte-array.

### Testing Considerations

- [ ] TODO tests. For each stubbed out case, there instead should be a `test.failing` test.

Old code that dispatched only on the pass-styles it knew about will not correctly handle passables that contain byte-arrays.

The shim cannot reasonably support TypedArrays of DataViews backed by Immutable ArrayBuffers.

### Upgrade Considerations

- [ ] TODO BREAKING.md
- [ ] TODO NEWS.md

### DIFF
--- a/packages/marshal/src/rankOrder.js
+++ b/packages/marshal/src/rankOrder.js
@@ -46,12 +46,11 @@ const { entries, fromEntries, setPrototypeOf, is } = Object;
 const sameValueZero = (x, y) => x === y || is(x, y);
 
 /**
- * @deprecated use type-specific `compareNumerics` etc.
  * @param {any} left
  * @param {any} right
  * @returns {RankComparison}
  */
-export const trivialComparator = (left, right) =>
+const trivialComparator = (left, right) =>
   // eslint-disable-next-line no-nested-ternary, @endo/restrict-comparison-operands
   left < right ? -1 : left === right ? 0 : 1;
 
@@ -88,9 +87,8 @@ const passStyleRanks = /** @type {PassStyleRanksRecord} */ (
       // thus contain only code units that are equivalent to code points.
       // In practice, they are entirely printable ASCII
       // (0x20 SPACE through 0x7E TILDE).
-      .sort(([_aStyle, aPrefixes], [_bStyle, bPrefixes]) => {
-        // eslint-disable-next-line no-nested-ternary
-        return aPrefixes < bPrefixes ? -1 : aPrefixes > bPrefixes ? 1 : 0;
+      .sort(([_leftStyle, leftPrefixes], [_rightStyle, rightPrefixes]) => {
+        return trivialComparator(leftPrefixes, rightPrefixes);
       })
       .map(([passStyle, prefixes], index) => {
         // Verify that `prefixes` is sorted, and cover all strings that start
@@ -227,6 +225,41 @@ export const makeComparatorKit = (compareRemotables = (_x, _y) => NaN) => {
         // If all matching elements were tied, then according to their lengths.
         // If array X is a prefix of array Y, then X has an earlier rank than Y.
         return comparator(left.length, right.length);
+      }
+      case 'byteArray': {
+        // ByteArrays compare by shortlex.
+        // - first, if they are of unequal length, then the shorter is less.
+        // - then, among byteArrays of equal length, by lexicographic comparison
+        //   of their bytes in ascending order.
+        const { byteLength: leftLen } = left;
+        const { byteLength: rightLen } = right;
+        if (leftLen < rightLen) {
+          return -1;
+        }
+        if (leftLen > rightLen) {
+          return 1;
+        }
+
+        // Account for gaps in the @endo/immutable-arraybuffer shim.
+        const leftArray =
+          Object.getPrototypeOf(left) === ArrayBuffer.prototype
+            ? new Uint8Array(left)
+            : new Uint8Array(left.slice(0));
+        const rightArray =
+          Object.getPrototypeOf(right) === ArrayBuffer.prototype
+            ? new Uint8Array(right)
+            : new Uint8Array(right.slice(0));
+        for (let i = 0; i < leftLen; i += 1) {
+          const leftByte = leftArray[i];
+          const rightByte = rightArray[i];
+          if (leftByte < rightByte) {
+            return -1;
+          }
+          if (leftByte > rightByte) {
+            return 1;
+          }
+        }
+        return 0;
       }
       case 'tagged': {
         // Lexicographic by `[Symbol.toStringTag]` then `.payload`.
--- a/packages/marshal/test/marshal-stringify.test.js
+++ b/packages/marshal/test/marshal-stringify.test.js
@@ -38,11 +38,13 @@ test('marshal stringify errors', t => {
     t.throws(() => stringify({}), {
       message: /Cannot pass non-frozen objects like .*. Use harden()/,
     });
-    // @ts-expect-error intentional error
+    // at-ts-ignore rather than at-expect-error because of disagreement
+    // @ts-ignore intentional error
     t.throws(() => stringify(harden(new Uint8Array(1))), {
       message: 'Cannot pass mutable typed arrays like "[Uint8Array]".',
     });
-    // @ts-expect-error intentional error
+    // at-ts-ignore rather than at-expect-error because of disagreement
+    // @ts-ignore intentional error
     t.throws(() => stringify(harden(new Int16Array(1))), {
       message: 'Cannot pass mutable typed arrays like "[Int16Array]".',
     });
--- a/packages/pass-style/src/byteArray.js
+++ b/packages/pass-style/src/byteArray.js
@@ -1,0 +1,49 @@
+import { X } from '@endo/errors';
+
+/**
+ * @import {PassStyleHelper} from './internal-types.js';
+ */
+
+const { getPrototypeOf, getOwnPropertyDescriptor } = Object;
+const { ownKeys, apply } = Reflect;
+
+// @ts-expect-error TODO How do I add it to the ArrayBuffer type?
+const AnImmutableArrayBuffer = new ArrayBuffer(0).transferToImmutable();
+
+/**
+ * As proposed, this will be the same as `ArrayBuffer.prototype`. As shimmed,
+ * this will be a hidden intrinsic that inherits from `ArrayBuffer.prototype`.
+ * Either way, get this in a way that we can trust it after lockdown, and
+ * require that all immutable ArrayBuffers directly inherit from it.
+ */
+const ImmutableArrayBufferPrototype = getPrototypeOf(AnImmutableArrayBuffer);
+
+const immutableGetter = /** @type {(this: ArrayBuffer) => boolean} */ (
+  // @ts-expect-error We know the desciptor is there.
+  getOwnPropertyDescriptor(ImmutableArrayBufferPrototype, 'immutable').get
+);
+
+/**
+ * @type {PassStyleHelper}
+ */
+export const ByteArrayHelper = harden({
+  styleName: 'byteArray',
+
+  canBeValid: (candidate, check = undefined) =>
+    (candidate instanceof ArrayBuffer &&
+      // @ts-expect-error TODO How do I add it to the ArrayBuffer type?
+      candidate.immutable) ||
+    (!!check && check(false, X`Immutable ArrayBuffer expected: ${candidate}`)),
+
+  assertRestValid: (candidate, _passStyleOfRecur) => {
+    getPrototypeOf(candidate) === ImmutableArrayBufferPrototype ||
+      assert.fail(X`Malformed ByteArray ${candidate}`, TypeError);
+    apply(immutableGetter, candidate, []) ||
+      assert.fail(X`Must be an immutable ArrayBuffer: ${candidate}`);
+    ownKeys(candidate).length === 0 ||
+      assert.fail(
+        X`ByteArrays must not have own properties: ${candidate}`,
+        TypeError,
+      );
+  },
+});
--- a/packages/pass-style/src/deeplyFulfilled.js
+++ b/packages/pass-style/src/deeplyFulfilled.js
@@ -6,7 +6,7 @@ import { passStyleOf } from './passStyleOf.js';
 import { makeTagged } from './makeTagged.js';
 
 /**
- * @import {Passable, Primitive, CopyRecord, CopyArray, CopyTagged, RemotableObject} from '@endo/pass-style'
+ * @import {Passable, ByteArray, CopyRecord, CopyArray, CopyTagged, RemotableObject} from '@endo/pass-style'
  */
 
 const { ownKeys } = Reflect;
@@ -104,6 +104,11 @@ export const deeplyFulfilled = async val => {
       const valPs = arr.map(p => deeplyFulfilled(p));
       // @ts-expect-error not assignable to type 'DeeplyAwaited<T>'
       return E.when(Promise.all(valPs), vals => harden(vals));
+    }
+    case 'byteArray': {
+      const bytes = /** @type {ByteArray} */ (val);
+      // @ts-expect-error not assignable to type 'DeeplyAwaited<T>'
+      return bytes;
     }
     case 'tagged': {
       const tgd = /** @type {CopyTagged} */ (val);
--- a/packages/pass-style/src/passStyleOf.js
+++ b/packages/pass-style/src/passStyleOf.js
@@ -12,6 +12,7 @@ import {
 } from './passStyle-helpers.js';
 
 import { CopyArrayHelper } from './copyArray.js';
+import { ByteArrayHelper } from './byteArray.js';
 import { CopyRecordHelper } from './copyRecord.js';
 import { TaggedHelper } from './tagged.js';
 import {
@@ -31,23 +32,20 @@ import { assertPassableString } from './string.js';
 /** @import {CopyArray, CopyRecord, CopyTagged, Passable} from './types.js' */
 /** @import {PassStyle} from './types.js' */
 /** @import {PassStyleOf} from './types.js' */
-/** @import {PrimitiveStyle} from './types.js' */
-
-/** @typedef {Exclude<PassStyle, PrimitiveStyle | "promise">} HelperPassStyle */
 
 const { ownKeys } = Reflect;
 const { isFrozen, getOwnPropertyDescriptors, values } = Object;
 
 /**
+ * @template {Record<PassStyle, PassStyleHelper>} HelpersRecord
  * @param {PassStyleHelper[]} passStyleHelpers
- * @returns {Record<HelperPassStyle, PassStyleHelper> }
+ * @returns {HelpersRecord}
  */
-
 const makeHelperTable = passStyleHelpers => {
-  /** @type {Record<HelperPassStyle, any> & {__proto__: null}} */
   const HelperTable = {
     __proto__: null,
     copyArray: undefined,
+    byteArray: undefined,
     copyRecord: undefined,
     tagged: undefined,
     error: undefined,
@@ -65,7 +63,9 @@ const makeHelperTable = passStyleHelpers => {
       Fail`missing helper for ${q(styleName)}`;
   }
 
-  return harden(HelperTable);
+  return /** @type {HelpersRecord} */ (
+    /** @type {unknown} */ (harden(HelperTable))
+  );
 };
 
 /**
@@ -236,6 +236,7 @@ export const passStyleOf =
   (globalThis && globalThis[PassStyleOfEndowmentSymbol]) ||
   makePassStyleOf([
     CopyArrayHelper,
+    ByteArrayHelper,
     CopyRecordHelper,
     TaggedHelper,
     ErrorHelper,
--- a/packages/pass-style/src/types.d.ts
+++ b/packages/pass-style/src/types.d.ts
@@ -3,15 +3,18 @@ import { PASS_STYLE } from './passStyle-helpers.js';
 
 /**
  * Matches any [primitive value](https://developer.mozilla.org/en-US/docs/Glossary/Primitive).
+ * TODO This now represents passable primitives, some of which are represented
+ * in JS as JS objects, not JS primitives.
  */
 export type Primitive =
-  | null
   | undefined
-  | string
-  | number
+  | null
   | boolean
-  | symbol
-  | bigint;
+  | number
+  | bigint
+  | string
+  | ByteArray
+  | symbol; // TODO we'll need to stop using the JS symbol type here.
 
 export type PrimitiveStyle =
   | 'undefined'
@@ -20,6 +23,7 @@ export type PrimitiveStyle =
   | 'number'
   | 'bigint'
   | 'string'
+  | 'byteArray'
   | 'symbol';
 
 export type ContainerStyle = 'copyRecord' | 'copyArray' | 'tagged';
@@ -116,16 +120,16 @@ export type PassStyleOf = {
 /**
  * A Passable is PureData when its entire data structure is free of PassableCaps
  * (remotables and promises) and error objects.
- * PureData is an arbitrary composition of primitive values into CopyArray
- * and/or
- * CopyRecord and/or CopyTagged containers (or a single primitive value with no
- * container), and is fully pass-by-copy.
+ * PureData is an arbitrary composition of primitive values into CopyArray,
+ * CopyRecord, and/or CopyTagged containers
+ * (or a single primitive value with no container), and is fully pass-by-copy.
  *
- * This restriction assures absence of side effects and interleaving risks *given*
- * that none of the containers can be a Proxy instance.
+ * This restriction assures absence of side effects and interleaving risks
+ * *given* that none of the containers can be a Proxy instance.
  * TODO SECURITY BUG we plan to enforce this, giving PureData the same security
  * properties as the proposed
  * [Records and Tuples](https://github.com/tc39/proposal-record-tuple).
+ * (TODO update to point at the non-trapping shim)
  *
  * Given this (currently counter-factual) assumption, a PureData value cannot
  * be used as a communications channel,
@@ -155,6 +159,11 @@ export type PassableCap = Promise<any> | RemotableObject;
  * A Passable sequence of Passable values.
  */
 export type CopyArray<T extends Passable = any> = Array<T>;
+
+/**
+ * A hardened immutable ArrayBuffer.
+ */
+export type ByteArray = ArrayBuffer;
 
 /**
  * A Passable dictionary in which each key is a string and each value is Passable.
--- a/packages/patterns/src/keys/checkKey.js
+++ b/packages/patterns/src/keys/checkKey.js
@@ -108,6 +108,13 @@ const keyMemo = new WeakSet();
  */
 export const checkKey = (val, check) => {
   if (!isObject(val)) {
+    // TODO!!! Once we switch `symbol` to be represented by an object type,
+    // it will no longer fall into the `!isObject` camp, just as byteArrays
+    // are "primitive" in the distributed object system but are still
+    // objects at the JS level.
+    // TODO: look for other uses of `isObject` at the pass-style level
+    // or above that might be misclassifying this way.
+    //
     // TODO There is not yet a checkPassable, but perhaps there should be.
     // If that happens, we should call it here instead.
     assertPassable(val);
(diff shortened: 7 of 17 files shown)


packages/marshal/src/rankOrder.js
```diff
@@ -263,6 +263,26 @@ export const makeComparatorKit = (compareRemotables = (_x, _y) => 0) => {
...
+        for (let i = 0; i < byteLen; i += 1) {
+          const leftByte = leftArray[i];
+          const rightByte = rightArray[i];
+          if (leftByte < rightByte) {
+            return -1;
+          }
+          if (leftByte > rightByte) {
+            return 1;
+          }
+        }
```
dckc, line 279: This really isn't in the stdlib, optimized down to `memcmp`?! (looks... nope...) Wow.

packages/pass-style/src/copyBytes.js
```diff
@@ -0,0 +1,82 @@
...
+    arrayBuffer = sliceOf(arrayBuffer);
+  } catch {
+    Fail`Expected genuine ArrayBuffer" ${arrayBuffer}`;
+  }
+  /** @type {CopyBytes} */
+  const copyBytes = {
+    // Can't say it this way because it confuses TypeScript
+    // __proto__: ArrayBuffer.prototype,
+    byteLength: arrayBuffer.byteLength,
+    slice(start, end) { return sliceOf(arrayBuffer, start, end); },
```
dckc, line 46: a slice of a `CopyBytes` is a mutable `ArrayBuffer`? am i reading that right? that seems weird.

mhofman (commented):
I think we should try to align copyBytes with the prior art of the Web platform's `Blob`.

packages/pass-style/src/copyBytes.js
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * A CopyBytes is much like an ArrayBuffer, but immutable.
+ * It cannot be used as an ArrayBuffer argument when a genuine ArrayBuffer is
+ * needed. But a `copyBytes.slice()` is a genuine ArrayBuffer, initially with
+ * a copy of the copyByte's data.
+ *
+ * On platforms that support freezing ArrayBuffer, like perhaps a future XS,
+ * (TODO) the intention is that `copyBytes` could hold on to a single frozen
+ * one and return it for every call to `arrayBuffer.slice`, rather than making
+ * a fresh copy each time.
```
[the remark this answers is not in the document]
dckc: > whether there should be a JS API to create an immutable ArrayBuffer from another one

We've proposed [snapshot](https://github.com/tc39/proposal-readonly-collections#snapshotdivergereadonlyview-methods-for-all-collections) for that, yes?

odd... https://github.com/tc39/proposal-readonly-collections#status says stage 1 but I don't see it inhttps://github.com/tc39/proposals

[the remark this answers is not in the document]
mhofman: That proposal seem targeted at collections, not ArrayBuffer. There is however a related proposal for ArrayBuffer: https://github.com/tc39/proposal-limited-arraybuffer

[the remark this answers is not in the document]
dckc: Arrays aren't collections? hm.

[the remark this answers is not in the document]
erights: ArrayBuffers aren't arrays.

packages/marshal/test/marshal-stringify.test.js
```diff
@@ -38,11 +38,9 @@ test('marshal stringify errors', t => {
     t.throws(() => stringify({}), {
       message: /Cannot pass non-frozen objects like .*. Use harden()/,
     });
-    // @ts-expect-error intentional error
```
erights, line 41: drive-by.

packages/marshal/test/marshal-stringify.test.js
```diff
@@ -38,11 +38,9 @@ test('marshal stringify errors', t => {
...
     t.throws(() => stringify(harden(new Uint8Array(1))), {
       message: 'Cannot pass mutable typed arrays like "[Uint8Array]".',
     });
-    // @ts-expect-error intentional error
```
erights, line 45: drive-by

packages/patterns/src/keys/copyBag.js
```diff
@@ -100,7 +100,6 @@ export const checkBagEntries = (bagEntries, check) => {
       );
     }
   }
-  // @ts-expect-error XXX Key types
```
erights, line 103: drive-by. Why wasn't this already giving an error forcing it to be removed?

packages/patterns/src/keys/keycollection-operators.js
```diff
@@ -27,7 +27,6 @@ import { q, Fail } from '@endo/errors';
  * @returns {IterableIterator<[Key, V]>}
  */
 const generateFullSortedEntries = (entries, rankCompare, fullCompare) => {
-  // @ts-expect-error XXX Key types
```
erights, line 30: drive-by

packages/patterns/test/patterns.test.js
```diff
@@ -532,7 +532,6 @@ const runTests = (t, successCase, failCase) => {
     t.throws(
       () => {
         copyMapComparison || Fail`No CopyMap comparison support`;
-        // @ts-expect-error XXX Key types
```
erights, line 535: drive-by

packages/patterns/test/types.test-d.ts
```diff
@@ -4,7 +4,6 @@ import { isKey, isScalarKey } from '../src/keys/checkKey.js';
 import { M } from '../src/patterns/patternMatchers.js';
 import type { Key, ScalarKey } from '../src/types.js';
 
-// @ts-expect-error M.any missing parens
```
erights, line 7: drive-by

packages/pass-style/src/byteArray.js
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * @param {unknown} candidate
+ * @param {import('./types.js').Checker} [check]
+ * @returns {boolean}
+ */
+const canBeValid = (candidate, check = undefined) =>
+  (candidate instanceof ArrayBuffer &&
+    // @ts-expect-error TODO How do I add it to the ArrayBuffer type?
+    candidate.immutable) ||
```
mhofman, line 31: Remind me, what are the semantics of `canBeValid`, and why are the checks much loser?

mhofman (commented):
I actually like the direction this is going. It's relying on a global to have certain semantics, which we can shim, either in ses or in a separate independent shim. We mainly have to hope the committee would be accepting this or we'll end up with some weird compat issues

packages/pass-style/src/byteArray.js
erights: Both better error messages and reducing search. First, what kind of passable does it seem the candidate is trying to be? This first checking pass should be quick. But the check ***must*** be tight enough to ensure that the candidate cannot be valid as any of the later categories. (Logic programming calls this committed choice.) Once we know what bucket it's in, we then try to see if it is a valid case of that bucket. If not, we can emit more relevant complaints, rather than continue searching.

packages/marshal/src/rankOrder.js
erights: Hi @dckc feel free to propose something to tc39. But I'm resolving here ;)

packages/pass-style/src/copyBytes.js
erights: Now, with https://github.com/endojs/endo/pull/2311 following @phoddie's suggestion at https://github.com/endojs/endo/pull/2309#issuecomment-2155513240 , this will be proposed as merely an ArrayBuffer.p.slice producing a normal mutable ArrayBuffer, independent of whether the original was mutable or immutable. Nothing could be less surprising. Resolving.

packages/pass-style/src/copyBytes.js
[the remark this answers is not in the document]
erights: We're no longer trying to base this on https://github.com/tc39/proposal-readonly-collections#snapshotdivergereadonlyview-methods-for-all-collections but rather a `transferToImmutable` to be proposed, as shimmed at #2311 . Resolving.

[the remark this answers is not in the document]
erights: @mhofman , thanks for the pointer to https://github.com/tc39/proposal-limited-arraybuffer !

@Jack-Works , what do you think of `transferToImmutable` as an alternate way to achieve these goals? (See shim at #2311 of suggestion by @phoddie at https://github.com/endojs/endo/pull/2309#issuecomment-2155513240 .) Since proposal-limited-arraybuffer is at stage 1, if you like this alternate route to these goals, I'd consider it within stage 1 to shift that proposal in this direction. If you agree, should we work together to revise proposal-limited-arraybuffer in this way? I'd certainly be happy to co-champion.

packages/pass-style/src/copyBytes.js
```diff
@@ -0,0 +1,82 @@
...
+ * unprivileged re-validation of open data, which is incompat with our
+ * need to encapsulate `arrayBuffer`, the genuinely mutable ArrayBuffer.
+ *
+ * @param {unknown} candidate
+ * @param {import('./types.js').Checker} [check]
+ * @returns {boolean}
+ */
+const canBeValid = (candidate, check = undefined) =>
+  // @ts-expect-error `has` argument can actually be anything.
+  genuineCopyBytes.has(candidate);
```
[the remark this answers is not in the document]
erights: > I'm wondering if we shouldn't go all the way and use a class to define CopyBytes, in which case the backing buffer could be stored in a private field, and this could be a simple brand check on the private field.

Done, in the #2311 shim.

> Which does raise the question of whether we actually need CopyBytes or can simply rely on a platform Blob, with a shim for platforms without native support. 

After discussions in various threads resulting in #2311, we agreed that Blob itself in unsuitable. Given that something new was needed anyway, #2311 bets on the immutable ArrayBuffer approach. Resolving.

packages/pass-style/src/byteArray.js
erights: Resolving

packages/patterns/test/types.test-d.ts
erights: restored

packages/patterns/test/patterns.test.js
erights: restored

packages/patterns/src/keys/keycollection-operators.js
erights: restored

packages/patterns/src/keys/copyBag.js
erights: restored

packages/marshal/test/marshal-stringify.test.js
erights: restored, but as an @ts-ignore because of inconsistent reports

packages/marshal/test/marshal-stringify.test.js
erights: restored, but as an @ts-ignore because of inconsistent reports

packages/pass-style/src/copyBytes.js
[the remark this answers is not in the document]
mhofman: Limited ArrayBuffer has stated goals that go beyond a simple view on an immutable ArrayBuffer. There are likely orthogonal aspects.

packages/marshal/src/rankOrder.js
```diff
@@ -209,6 +211,26 @@ export const makeComparatorKit = (compareRemotables = (_x, _y) => 0) => {
...
+        // If all corresponding bytes are the same,
+        // then according to their lengths.
+        // Thus, if the data of ByteArray X is a prefix of
+        // the data of ByteArray Y, then X is smaller than Y.
+        return comparator(left.byteLength, right.byteLength);
```
erights, line 232: Change to shortlex order. See https://en.wikipedia.org/wiki/Shortlex_order

Shortlex order would enable encodePassable (aka compactOrdered) to encode buffers with a prefix followed by the unescaped bytes. Avoiding escaping would be great.

packages/patterns/src/keys/compareKeys.js
```diff
@@ -147,6 +147,30 @@ export const compareKeys = (left, right) => {
...
+        if (leftByte > rightByte) {
+          return 1;
+        }
+      }
+      // If all corresponding bytes are the same,
+      // then according to their lengths.
+      // Thus, if the data of ByteArray X is a prefix of
+      // the data of ByteArray Y, then X is smaller than Y.
+      // @ts-expect-error narrowed
+      return compareRank(left.byteLength, right.byteLength);
```
erights, line 172: If both are byteArrays, then we should delegate the whole thing to `compareRank`. Which, btw, we'll be switching to shortlex order.

packages/marshal/src/rankOrder.js
mhofman: So this makes ByteArray different than string for ordering?

erights: Yes. For strings, normal lexicographic ordering is overwhelmingly more expected than shortlex. For binaries of different length, I don't think there is any one overwhelming expectation.

FWIW, if I did think we could get away with shortlex for strings too, I would go for it, for the same reason. Needing to insert escapes into bulk data sucks. Attn @gibson042

erights: Agreed. That's another reason we cannot get away with it.

erights: Indeed, because of this breaking change issue, we haven't even moved forward on https://github.com/endojs/endo/pull/2008

packages/marshal/src/rankOrder.js
erights: @dckc this is now more urgent. It would be nice to have a builtin that could compare Immutable ArrayBuffer contents at builtin speeds. If ImmutableArrayBuffers had their own class, I might propose `ImmutableArrayBuffer.shortLexCompare`. But they simply share the `ArrayBuffer` class, and a compare over all ArrayBuffer contents might hit more resistance.

Suggestions? (Whatever those suggestion are, they can come after this PR and after the Immutable ArrayBuffer proposal.)

packages/pass-style/src/copyBytes.js
erights: Noting that there's also now a `sliceToImmutable` that works on all ArrayBuffers and, like `toImmutable`, always produces an Immutable ArrayBuffer.

packages/marshal/src/rankOrder.js
erights: In fact, we are moving forward on the purpose of #2008 as https://github.com/tc39/proposal-compare-strings-by-codepoint . But that's not relevant here. `compareRank` and its brethren will compare byteArrays by shortlex --- first by length, and then lexicographically by bytes about byteArrays of the same length.

Resolving.

packages/marshal/src/rankOrder.js
dckc: > Suggestions?

not really.

For any stuff that hasn't been proposed to tc39 yet, the time it takes to get widely deployed seems well beyond my attention-span.

My "Wow" comment was more of a TIL thing.

packages/pass-style/src/byteArray.js
```diff
@@ -0,0 +1,50 @@
...
+ * Either way, get this in a way that we can trust it after lockdown, and
+ * require that all immutable ArrayBuffers directly inherit from it.
+ */
+const ImmutableArrayBufferPrototype = getPrototypeOf(AnImmutableArrayBuffer);
+
+// @ts-expect-error ok to implicitly assert the access is found
+const immutableGetter = getOwnPropertyDescriptor(
+  ImmutableArrayBufferPrototype,
+  'immutable',
+).get;
```
gibson042, line 24:
```suggestion
const immutableGetter = /** @type {(this: ArrayBuffer) => boolean} */ (
  getOwnPropertyDescriptor(ImmutableArrayBufferPrototype, 'immutable').get
);
```

packages/pass-style/src/deeplyFulfilled.js
```diff
@@ -105,6 +105,13 @@ export const deeplyFulfilled = async val => {
       // @ts-expect-error not assignable to type 'DeeplyAwaited<T>'
       return E.when(Promise.all(valPs), vals => harden(vals));
     }
+    case 'byteArray': {
+      const bytes = /** @type {ByteArray} */ (val);
+      // @ts-expect-error Why
+      // "Type 'ArrayBuffer' is not assignable to type 'DeeplyAwaited<T>'."?
+      // TODO fix.
```
gibson042, line 112: Aligning with the other clauses:
```suggestion
      // @ts-expect-error not assignable to type 'DeeplyAwaited<T>'
```

packages/pass-style/src/types.d.ts
```diff
@@ -22,7 +22,11 @@ export type PrimitiveStyle =
   | 'string'
   | 'symbol';
 
-export type ContainerStyle = 'copyRecord' | 'copyArray' | 'tagged';
+export type ContainerStyle =
+  | 'copyRecord'
+  | 'copyArray'
+  | 'byteArray'
```
gibson042, line 28: https://endojs.github.io/endo/types/_endo_pass_style.Passable.html documents that Containers aggregate other Passables, so I don't think it should include "byteArray", which should instead be either a special case or (if we're taking a data model perspective) a primitive (similar to "string"). Which reminds me that packages/patterns/README.md also needs updating.

packages/pass-style/src/types.d.ts
```diff
@@ -67,6 +72,7 @@ export type PassByRef =
  *     | 'string' | 'symbol'). (Passable considers `void` to be `undefined`.)
  *   * Containers aggregate other Passables into
  *     * sequences as CopyArrays (PassStyle 'copyArray'), or
+ *     * sequences of 8-bit bytes (PassStyle 'byteArray'), or
```
gibson042, line 75: As noted above, 8-bit bytes are not "other Passables".

packages/pass-style/src/types.d.ts
```diff
@@ -116,10 +123,9 @@ export type PassStyleOf = {
 /**
  * A Passable is PureData when its entire data structure is free of PassableCaps
  * (remotables and promises) and error objects.
- * PureData is an arbitrary composition of primitive values into CopyArray
- * and/or
- * CopyRecord and/or CopyTagged containers (or a single primitive value with no
- * container), and is fully pass-by-copy.
+ * PureData is an arbitrary composition of primitive values into CopyArray,
+ * ByteArray, CopyRecord, and/or CopyTagged containers
+ * (or a single primitive value with no container), and is fully pass-by-copy.
```
gibson042, line 128: Likewise here.

packages/pass-style/src/types.d.ts
```diff
@@ -156,6 +162,11 @@ export type PassableCap = Promise<any> | RemotableObject;
  */
 export type CopyArray<T extends Passable = any> = Array<T>;
 
+/**
+ * A `ByteArray` is a normal hardened immutable `ArrayBuffer`
```
gibson042, line 166: Styling nits:
```suggestion
 * A hardened immutable ArrayBuffer.
```

packages/patterns/src/keys/compareKeys.js
```diff
@@ -147,6 +147,30 @@ export const compareKeys = (left, right) => {
       // @ts-expect-error narrowed
       return compareRank(left.length, right.length);
     }
+    case 'byteArray': {
```
gibson042, line 150: Given the non-recursive nature of pass style "byteArray", I wonder if it belongs earlier in these `switch` cases. But we're not fully consistent anyway, so no harm if you want to leave it after "copyArray" (which is recursive).

packages/patterns/src/keys/compareKeys.js
```diff
@@ -147,6 +147,30 @@ export const compareKeys = (left, right) => {
...
+      for (let i = 0; i < byteLen; i += 1) {
+        const leftByte = leftArray[i];
+        const rightByte = rightArray[i];
+        if (leftByte < rightByte) {
+          return -1;
+        }
+        if (leftByte > rightByte) {
+          return 1;
+        }
+      }
```
gibson042, line 166: Small optimization: copy only when necessary.
```suggestion
      // Account for gaps in the @endo/immutable-arraybuffer shim.
      const leftArray =
        Object.getPrototypeOf(left) === ArrayBuffer.prototype
          // @ts-expect-error narrowed
          ? new Uint8Array(left)
          // @ts-expect-error narrowed
          : new Uint8Array(left.slice(0));
      const rightArray =
        Object.getPrototypeOf(right) === ArrayBuffer.prototype
          // @ts-expect-error narrowed
          ? new Uint8Array(right)
          // @ts-expect-error narrowed
          : new Uint8Array(right.slice(0));
      // @ts-expect-error narrowed
      const byteLen = Math.min(left.byteLength, right.byteLength);
      for (let i = 0; i < byteLen; i += 1) {
        const leftByte = leftArray[i];
        const rightByte = rightArray[i];
        if (leftByte < rightByte) {
          return -1;
        }
        if (leftByte > rightByte) {
          return 1;
        }
      }
```
And also, why not implement shortlex ordering for ByteArrays from the start?

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -381,6 +382,9 @@ const makePatternKit = () => {
         // patterns
         return patt.every(checkIt);
       }
+      case 'byteArray': {
+        return true;
+      }
```
gibson042, line 387: ByteArrays are Keys, so this code should be unreachable as it is for e.g. "string".
```suggestion
```

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -637,6 +642,10 @@ const makePatternKit = () => {
         // ]);
         break;
       }
+      case 'byteArray': {
+        // TODO implement
+        throw Fail`getCover of byteArray not yet implemented`;
+      }
```
gibson042, line 648: Likewise here.
```suggestion
```

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1207,6 +1216,34 @@ const makePatternKit = () => {
...
+    checkIsWellFormed: (payload, check) =>
+      checkIsWellFormedWithLimit(
+        payload,
+        harden([]),
+        check,
+        'match:bytes payload',
+      ),
+
+    getRankCover: (_matchPayload, _encodePassable) =>
+      getPassStyleCover('string'),
```
gibson042, line 1244:
```suggestion
      getPassStyleCover('byteArray'),
```

gibson042 (changes requested):
This looks mostly good, but I have reservations about treating ByteArray as a Container [of Passables].

packages/patterns/src/keys/checkKey.js
```diff
@@ -551,6 +551,9 @@ const checkKeyInternal = (val, check) => {
       // A copyArray is a key iff all its children are keys
       return val.every(checkIt);
     }
+    case 'byteArray': {
+      return true;
+    }
```
erights: I noticed you did not suggest deleting this one, which is correct. My investigating why led me to realize a bug we would have had once passable symbols are represented as js objects. `isObject` is no longer a good classifiers for whether something is a passable-primitive value. I added a TODO to the code.

Resolving.

erights:
Wow, @gibson042 , those suggestions were extremely helpful, thanks!

PTAL

packages/pass-style/src/passStyleOf.js
```diff
@@ -33,7 +34,11 @@ import { assertPassableString } from './string.js';
 /** @import {PassStyleOf} from './types.js' */
 /** @import {PrimitiveStyle} from './types.js' */
 
-/** @typedef {Exclude<PassStyle, PrimitiveStyle | "promise">} HelperPassStyle */
+/**
+ * TODO will need a symbol helper too.
+ * @typedef { "byteArray" | Exclude<PassStyle, PrimitiveStyle | "promise">
+ * } HelperPassStyle
+ */
```
gibson042, line 41:
```suggestion
/**
 * Non-primitive pass styles (except "promise") and special primitive pass
 * styles ("byteArray", and in the future also "symbol") are supported by
 * helpers.
 * @typedef {Exclude<PassStyle, PrimitiveStyle | "promise"> | "byteArray"} HelperPassStyle
 */
```
but for that matter, I don't think this HelperPassStyle type is carrying its own weight, and would prefer dropping it:
```js
/**
 * @template {Record<PassStyle, PassStyleHelper> & {remotable: PassStyleHelper}} HelpersRecord
 * @param {Array<PassStyleHelper & {styleName: keyof HelpersRecord}>} passStyleHelpers
 * @returns {HelpersRecord}
 */
const makeHelperTable = passStyleHelpers => {
```

packages/patterns/src/keys/checkKey.js
```diff
@@ -543,6 +550,12 @@ const checkKeyInternal = (val, check) => {
 
   const passStyle = passStyleOf(val);
   switch (passStyle) {
+    case 'symbol': // anticipating a JS object representation of a passable symbol
+    case 'byteArray': // passable primitive, with js object representation.
+    case 'remotable': {
+      // All remotables are keys.
+      return true;
+    }
```
gibson042, line 558:
```suggestion
    case 'symbol': // anticipating a JS object representation of a passable symbol
    case 'byteArray': // passable primitive, with js object representation.
    case 'remotable': {
      // Any Primitive or Remotable is a Key, and `checkKey` covers those that
      // are represented in JS as *language* primitives while this covers the
      // remainder.
      return true;
    }
```

gibson042 (approved):
Thanks! LGTM modulo updating packages/patterns/README.md.

gibson042 (commented):
Oh, I think this PR should also update [arb-passable.js](https://github.com/endojs/endo/blob/master/packages/pass-style/tools/arb-passable.js), although accommodations will be necessary to exclude ByteArrays in [encodePassable.test.js](https://github.com/endojs/endo/blob/master/packages/marshal/test/encodePassable.test.js#L18) until they have a defined encoding, e.g.:
```patch
-/** @param {typeof import('@fast-check/ava').fc} fc */
-export const makeArbitraries = fc => {
+/**
+ * @param {typeof import('@fast-check/ava').fc} fc
+ * @param {Array<'byteArray'>} [exclusions]
+ */
+export const makeArbitraries = (fc, exclusions = []) => {
   const arbString = fc.oneof(fc.string(), fc.fullUnicodeString());
 
   const keyableLeaves = [
     fc.constantFrom(null, undefined, false, true),
     arbString,
     arbString.map(s => Symbol.for(s)),
     // primordial symbols and registered lookalikes
     fc.constantFrom(
       ...Object.getOwnPropertyNames(Symbol).flatMap(k => {
         const v = Symbol[k];
         if (typeof v !== 'symbol') return [];
         return [v, Symbol.for(k), Symbol.for(`@@${k}`)];
       }),
     ),
     fc.bigInt(),
     fc.integer(),
+    ...[fc.uint8Array().map(arr => arr.buffer.transferToImmutable())].filter(
+      () => !exclusions.includes('byteArray'),
+    ),
     fc.constantFrom(-0, NaN, Infinity, -Infinity),
     fc.record({}),
     fc.constantFrom(exampleAlice, exampleBob, exampleCarol),
   ];
```

erights:
> Oh, I think this PR should also update [arb-passable.js](https://github.com/endojs/endo/blob/master/packages/pass-style/tools/arb-passable.js), although accommodations will be necessary to exclude ByteArrays in [encodePassable.test.js](https://github.com/endojs/endo/blob/master/packages/marshal/test/encodePassable.test.js#L18) until they have a defined encoding,

Done

packages/pass-style/src/types.d.ts
erights: I just read patterns/README.md carefully, and did not spot the flaw you have in mind. Is this indeed the document you have in mind? What needs updating?

erights:
> Thanks! LGTM modulo updating packages/patterns/README.md.

I just read patterns/README.md carefully, and did not spot the flaw you have in mind. Is this indeed the document you have in mind? What needs updating?

gibson042 (approved):
> I just read patterns/README.md carefully, and did not spot the flaw you have in mind. Is this indeed the document you have in mind? What needs updating?

Ah, no change is needed now that we acknowledge ByteArray as primitive.

packages/pass-style/src/types.d.ts
erights: > Ah, no change is needed now that we acknowledge ByteArray as primitive

Resolving.

PR #1541, opened 2023-04-06
fix(ses): limit logged args per error

erights:
While @michaelfig and I were trying to track down a different bug, we noticed that the LRU budget on error notes, while correctly limiting the number of logged errors according to budget, failed to limit the number of logged args per error.

The latter is just kept as an ordered array. For the errors, there's a chronology of use after the initial entry was created, so an LRU was needed. For the logged args per error, they are only entered, and then not used until all of them are taken at once. Thus, dropping the front of the array if over budget is an adequate LRU.

This PR adds and enforces this separate budget of logged args per error.

### DIFF
--- a/packages/ses/src/error/note-log-args.js
+++ b/packages/ses/src/error/note-log-args.js
@@ -94,16 +94,18 @@ const spliceOut = cell => {
  *
  * TODO: Make parameterized `Key` and `Value` template types
  *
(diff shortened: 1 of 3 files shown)


erights:
(Self-request changes) Needs tests

michaelfig (approved):
Tidy, and LGTM!

PR #1542, opened 2023-04-07
Refactoring `@endo/eventual-send` types from `.d.ts` to `jsdoc` tags.

SMotaal:
This PR gradually refactors types from [src/index.d.ts] into respect `.js` modules.

**Motivation**

- Moving towards `jsdoc` being the single source of truth.

**Considerations**

- Figure out which types belong in each `.js` module.
- Opt for patterns that would prevent breakage of existing types.

**Chores**

- [x] Ensure consistent `import('.')` references
- [ ] Move all `E` related types into [src/E.js] or other `.js` modules.
  - [x] Move `EProxy` and related types (`FarRef`, `DataOnly` and `RemotableBrand`) into `makeE()`
  - [ ] Move other `E`-prefixed and related types into `makeE`
  - [x] Ensure all refactored types are exposed via `src/index.d.ts`
- [ ] Reassess the need for [src/index.d.ts] and plan next steps

[src/index.d.ts]: https://github.com/endojs/endo/blob/fc744665d2e6b8bb4715221b1c0ec500aeeb4416/packages/eventual-send/src/index.d.ts
[src/E.js]: https://github.com/endojs/endo/blob/fc744665d2e6b8bb4715221b1c0ec500aeeb4416/packages/eventual-send/src/E.js


packages/eventual-send/src/index.d.ts
```diff
@@ -26,9 +44,6 @@
 
 export type Callable = (...args: any[]) => any;
```
michaelfig, line 45: Internal-only.  Please move to `types.d.ts`.

packages/eventual-send/src/index.d.ts
```diff
@@ -26,9 +44,6 @@
...
 
-// Same as https://github.com/microsoft/TypeScript/issues/31394
-export type ERef<T> = PromiseLike<T> | T;
-
 export declare const EmptyObj: {};
```
michaelfig, line 47: Internal-only.

packages/eventual-send/src/index.d.ts
```diff
@@ -55,28 +70,6 @@ export type FilteredKeys<T, U> = {
   [P in keyof T]: T[P] extends U ? P : never;
```
michaelfig, line 70: FilteredKeys is internal-only.

packages/eventual-send/src/index.d.ts
```diff
@@ -55,28 +70,6 @@ export type FilteredKeys<T, U> = {
...
- * Creates a type that accepts both near and marshalled references that were
- * returned from `Remotable` or `Far`, and also promises for such references.
- */
-export type FarRef<Primary, Local = DataOnly<Primary>> = ERef<
-  Local & RemotableBrand<Local, Primary>
->;
-
 /**
  * `PickCallable<T>` means to return a single root callable or a record type
  * consisting only of properties that are functions.
```
michaelfig, line 75: More internal-only types:
- `PickCallable`
- `RemoteFunctions`
- `LocalRecord`

The following should be reexported from `handled-promise.js`
- ResolveWithPresenceOptionsBag
- HandledExecutor
- HandledPromiseConstructor
- HandledPromiseStaticMethods - internal to `handled-promise.js`
- Settler

`RemoteKit` is a bit trickier.  I'd like it defined in `E.js` as `EPromiseKit`, reexport it in `index.d.ts`, and change its only existing reference (in `@endo/captp/src/captp.js`)

michaelfig (commented):
I'm really liking this direction!

I audited the new `index.d.ts` and figured out specific suggestions for what I'd like to do to organise what remains.  Let's talk about whether it makes sense for you to take that on, or whether it is better to land this PR mostly as-is for a first step.

packages/eventual-send/src/E.js
```diff
@@ -101,49 +101,179 @@ function EsendOnlyProxyHandler(x, HandledPromise) {
...
+         * value will be the property fetched from whatever 'x' designates (or
+         * resolves to) in a future turn, not this one.
+         *
+         * @template T
+         * @param {T} x target for property get
+         * @returns {EGetters<import('.').LocalRecord<T>>} property get proxy
+         * @readonly
+         */
+        get: x =>
+          new Proxy(
```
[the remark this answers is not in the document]
SMotaal: I thought that was the intent but forgot to even ask, good that you caught it here 😅

mhofman:
> * Ensure consistent `import('.')` references

why remove the explicit `index`? I would prefer avoiding node resolution style as much as possible, even in type. I actually prefer indicating an extension name in type imports, but that would be weird in this case since the file is not an actual `.js`.

Maybe an alternative would be `import('@endo/eventual-send')` instead?

michaelfig:
> > * Ensure consistent `import('.')` references
> Maybe an alternative would be `import('@endo/eventual-send')` instead?

I think this will be worked out, and then will ping you for review later.  I'm putting the PR back into draft for now since it's mostly a place for @SMotaal and me to come to some agreement before attracting more attention from the rest of the team.

packages/eventual-send/src/index.d.ts
SMotaal: There are no more references to `EmptyObj` and unless I am missing something, I suppose this was substituted with `{}`.

@michaelfig Do you want me to check or just drop it?

packages/eventual-send/src/index.d.ts
SMotaal: `RemoteFunctions` and `LocalRecord` are only referenced in `E.js` while `PickCallable` is only referenced by `RemoteFunctions`. 

@michaelfig They seem to belong inside `E.js` however keeping them in `types.d.ts` would be needed if we want to preserve the line comments, thoughts?

packages/eventual-send/src/index.d.ts
michaelfig: Drop, please.

packages/eventual-send/src/index.d.ts
michaelfig: I think moving them to `E.js` is fine.  The comments are useful, but not essential for users of the types.

michaelfig (commented):
Looks good!  I'll postpone approval until I can solicit feedback from some other folks, but I think this is a great improvement in clarity.

SMotaal:
> **Note**: A number of `yarn lint` warnings showed up after rebasing from master in places that do not seem to be related to changes in this PR, but I could be mistaken — see [annotations](https://github.com/endojs/endo/actions/runs/5236596318).

michaelfig:
> > Note: A number of `yarn lint` warnings showed up after rebasing from master in places that do not seem to be related to changes in this PR, but I could be mistaken — see [annotations](https://github.com/endojs/endo/actions/runs/5236596318).

I checked the annotations, and it looks like these warnings were the result of endojs/endo#1622 .  I wouldn't sweat fixing them, at least until there's clear direction as to what idioms we should use in their place.

packages/eventual-send/src/E.js
```diff
@@ -31,13 +32,13 @@ const baseFreezableProxyHandler = {
...
+ * @param {import('./types').HandledPromiseConstructor} HandledPromise
  * @returns {ProxyHandler} the Proxy handler
  */
-function EProxyHandler(x, HandledPromise) {
-  return harden({
+const makeEProxyHandler = (x, HandledPromise) =>
+  harden({
     ...baseFreezableProxyHandler,
-    get(_target, p, receiver) {
+    get: (_target, p, receiver) => {
```
kungfooman, line 41: What's the narrative here? Arrow functions introduce more ascii soup than object function style.

packages/eventual-send/src/E.js
```diff
@@ -101,49 +101,285 @@ function EsendOnlyProxyHandler(x, HandledPromise) {
...
+ *     : T
+ * )} LocalRecord
+ */
+
+/**
+ * @template {unknown} R
+ * @typedef {{
+ *   promise: Promise<R>;
+ *   settler: import('./types').Settler<R>;
+ * }} EPromiseKit
```
kungfooman, line 366: Isn't this style more common in the codebase, or is there any decision on this?

```js
/**
 * @template {unknown} R - the advantage is that we can comment e.g. templates too
 * @typedef {object} EPromiseKit
 * @property {Promise<R>} promise - or comment on anything else really
 * @property {import('./types').Settler<R>} settler
 */
```

packages/eventual-send/src/E.js
```diff
@@ -101,49 +101,285 @@ function EsendOnlyProxyHandler(x, HandledPromise) {
...
+ *     : Awaited<T> extends RemotableBrand<infer L, infer R>
+ *     ? L
+ *     : T extends PromiseLike<infer U>
+ *     ? Awaited<T>
+ *     : T
+ * )} LocalRecord
+ */
+
+/**
+ * @template {unknown} R
```
SMotaal:
```suggestion
 * @template [R = unknown]
```

packages/eventual-send/src/handled-promise.js
```diff
@@ -581,5 +575,60 @@ export const makeHandledPromise = () => {
...
+ *   proxy?: {
+ *     handler: ProxyHandler<T>;
+ *     target: unknown;
+ *     revokerCallback?(revoker: () => void): void;
+ *   };
+ * }} ResolveWithPresenceOptionsBag
+ */
+
+/**
+ * @template {unknown} R
```
SMotaal:
```suggestion
 * @template [R = unknown]
```

packages/eventual-send/src/handled-promise.js
```diff
@@ -581,5 +575,60 @@ export const makeHandledPromise = () => {
...
+ * @typedef {(
+ *   resolveHandled: (value?: R) => void,
+ *   rejectHandled: (reason?: unknown) => void,
+ *   resolveWithPresence: (presenceHandler: Handler<{}>, options?: ResolveWithPresenceOptionsBag<{}>) => object,
+ * ) => void} HandledExecutor
+ */
+
+/**
+ * @template {unknown} R
```
SMotaal:
```suggestion
 * @template [R = unknown]
```

packages/eventual-send/src/E.js
michaelfig: This change reflects an aspiration to conform to the (evolving) [https://github.com/endojs/Jessie](Jessie syntax).

Ideally, as much Endo code as possible would be written in Jessie (or TinySES) to facilitate porting it to platforms that choose not to implement all of Javascript's foibles.

SMotaal: @michaelfig 

> This change reflects an aspiration to conform to the (evolving) [Jessie syntax](https://github.com/endojs/Jessie#readme%5D).
> 
> Ideally, as much Endo code as possible would be written in Jessie (or TinySES) to facilitate porting it to platforms that choose not to implement all of Javascript's foibles.

I appreciate having this here for reference.

SMotaal: @kungfooman did you have something in mind to add or is this resolved?

packages/eventual-send/src/E.js
SMotaal: @kungfooman resolved?

> **Aside**: Thanks for bringing attention to `@template {unknown} R` which was supposed to be `@template [R = unknown]` per later revisions 🤘

packages/eventual-send/src/E.js
kungfooman: Nothing to add, thank you both, I didn't know it was due to the limitation of the Jessie parser.

SMotaal: Thanks @kungfooman!

PR #1543, opened 2023-04-07
chore(exo): Define narrow Exo-related context types

gibson042:
Mirrors https://github.com/Agoric/agoric-sdk/pull/7360


erights (approved):
LGTM

PR #1544, opened 2023-04-08
feat(bundle-source): separate the `cache.js` library from the tool

michaelfig:
Please review commit-by-commit to avoid changes caused only by splitting the files.

- Make `@endo/bundle-source/cache.js` importable by other packages that want a bundle cache
- Separate the `bundle-source` script from the `cache.js` library, to make it clearer as a `tool.js` entrypoint
- Add filesystem-based locking to prevent bundling update collisions (seen when running AVA tests in parallel)


packages/bundle-source/cache.js
```diff
@@ -31,42 +29,49 @@ export const makeFileReader = (fileName, { fs, path }) => {
   });
 };
 
+/**
+ * @param {string} fileName
+ * @param {{ fs: import('fs'), path: import('path') }} io
```
dckc, line 34: I recently learned about [Pick<Type, Keys>](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys). It's pretty handy for narrowing down what `io` is really needed.

packages/bundle-source/cache.js
```diff
@@ -81,57 +86,91 @@ export const makeBundleCache = (wr, cwd, readPowers, opts) => {
...
+    // Check the bundle/meta file write lock.
+    try {
+      await lockWr.writeText('', { flag: 'wx' });
+    } catch (oops) {
+      if (oops.code !== 'EEXIST') {
+        throw oops;
+      }
+      // The lock exists, so something is already writing the bundle.
+      // eslint-disable-next-line no-use-before-define
+      return validate(targetName, rootPath);
```
dckc, line 108: so... we win if what we're trying to write is already there?

packages/bundle-source/cache.js
```diff
@@ -81,57 +86,91 @@ export const makeBundleCache = (wr, cwd, readPowers, opts) => {
...
+    }
   };
 
-  const validate = async targetName => {
+  const validate = async (targetName, rootOpt, log = defaultLog) => {
     const metaRd = wr.readOnly().neighbor(toBundleMeta(targetName));
+    const lockRd = wr.readOnly().neighbor(toBundleLock(targetName));
+
+    // Wait for the bundle to be written.
+    // eslint-disable-next-line no-await-in-loop
```
dckc, line 162: adding an `await null;` above the loop would make this await-safe, per recent discussion, yes?

dckc: oh... the loop starts with an `await`. fair enough.
I wonder if a `for await ...` would be more clear.

dckc (approved):
yeah... clobbering is not good.

Adding this much code without tests makes me uncomfortable. But... I started it :)

packages/bundle-source/cache.js
```diff
@@ -0,0 +1,346 @@
...
+  return makeBundleCache(destWr, cwd, readPowers, options);
+};
+
+/**
+ * @param {[string, string, string[]]} args
+ * @param {object} powers
+ * @param {(spec: string) => any} powers.loadModule
+ * @returns {void}
+ */
+export const main = async (args, { loadModule }) => {
```
turadg, line 320: command line handling shouldn't be a concern of this module. please move into `src/index.js`

turadg (commented):
Please retain the improvements in the agoric-sdk version:
https://github.com/Agoric/agoric-sdk/commit/609ab33410aacca6dc3f83d3d0ac1ddbffb3eb04 from 7103
https://github.com/Agoric/agoric-sdk/pull/7109
https://github.com/Agoric/agoric-sdk/pull/7120

turadg (changes requested):
Please retain the improvements in the agoric-sdk version:
- [x] https://github.com/Agoric/agoric-sdk/commit/609ab33410aacca6dc3f83d3d0ac1ddbffb3eb04 from 7103
- [x] https://github.com/Agoric/agoric-sdk/pull/7109
- [x] https://github.com/Agoric/agoric-sdk/pull/7120

packages/bundle-source/cache.js
michaelfig: Clarified with a better comment:

```js
      // The lock exists, so something is already writing the bundle on our
      // behalf.
      //
      // All we need to do is try validating the bundle, which will first wait
      // for the lock to disappear before reading the freshly-written bundle.
```

michaelfig:
> Please retain the improvements in the agoric-sdk version:

I'm planning on keeping this behaviour in the Agoric SDK rather than making Endo's implementation less generic.

The code I ended up with configures the logger as follows:

```js
export const makeNodeBundleCache = async (dest, options, loadModule) => {
  const log = (...args) => {
    const flattened = args.map(arg =>
      // Don't print stack traces.
      arg instanceof Error ? arg.message : arg,
    );
    console.log(
      // Make all messages prefixed and dim.
      `${styles.dim.open}[bundleTool]`,
      ...flattened,
      styles.dim.close,
    );
  };
  return makeNodeBCache(dest, { log, ...options }, loadModule);
};
```

turadg (approved):

PR #1545, opened 2023-04-08
build: consistently produce `yarn pack` type declarations

michaelfig:
- `prepack` now consistently builds Typscript types, and `postpack` removes them
- Sort out the difficulties between `@endo/marshal` and `@endo/pass-style`

### DIFF
--- a/packages/pass-style/src/types.js
+++ b/packages/pass-style/src/types.js
@@ -69,14 +69,14 @@ export {};
  */
 
 /**
- * @typedef {Passable} Remotable
+ * @typedef {Passable} RemotableObject
  *
  * An object marked as remotely accessible using the `Far` or `Remotable`
  * functions, or a local presence representing such a remote object.
  */
 
 /**
- * @typedef {Promise | Remotable} PassableCap
+ * @typedef {Promise | RemotableObject} PassableCap
  *
  * The authority-bearing leaves of a Passable's pass-by-copy superstructure.
  */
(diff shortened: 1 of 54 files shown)


turadg (approved):
Yay! I'd ask for a test plan to check against sdk, but I think you've already been doing that.

For the many new build configs this feature in 5.0 will be handy: https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#supporting-multiple-configuration-files-in-extends

packages/pass-style/src/types.js
```diff
@@ -69,14 +69,14 @@ export {};
  */
 
 /**
- * @typedef {Passable} Remotable
+ * @typedef {Passable} RemotableObject
```
mhofman, line 72: Ugh the conflict with the also exported `Remotable` runtime value from `make-far.js` is unfortunate. Maybe we can fix this mess once we have more precise types for `Passable`.

mhofman (approved):

packages/pass-style/src/types.js
michaelfig: Agreed,  I'm just happy to have a workaround until then. :)

michaelfig:
> Yay! I'd ask for a test plan to check against sdk, but I think you've already been doing that.
> 
> For the many new build configs this feature in 5.0 will be handy: https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#supporting-multiple-configuration-files-in-extends

Interesting!  Yes, the Typescript-related changes are a neverending firehose of busywork. :)

PR #1546, opened 2023-04-08
feat(captp): consistent garbage collection

michaelfig:
- Uses in-transit `exportID` reference counts to prevent races when dropping references
- Split the `slotToVal` map into `slotToImported`(strong)  and `slotToExported` (weak values with finalizers)
- Remove `async` methods wherever possible to better distinguish and propagate CapTP errors vs. userspace errors

### DIFF
--- a/packages/captp/src/captp.js
+++ b/packages/captp/src/captp.js
@@ -60,6 +60,7 @@ const reverseSlot = slot => {
  * communicates the response to the message
  * @property {TrapHost} [trapHost] if specified, enable this CapTP (host) to serve
  * objects marked with makeTrapHandler to synchronous clients (guests)
+ * @property {boolean} [gcImports] if true, aggressively garbage collect imports
  */
 
 /**
@@ -98,6 +99,7 @@ export const makeCapTP = (
     importHook,
     trapGuest,
     trapHost,
+    gcImports = false,
   } = opts;
 
   // It's a hazard to have trapGuest and trapHost both enabled, as we may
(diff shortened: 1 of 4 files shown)


packages/captp/src/captp.js
```diff
@@ -110,7 +119,7 @@ export const makeCapTP = (
 
   /** @type {any} */
   let unplug = false;
-  const quietReject = async (reason = undefined, returnIt = true) => {
+  const quietReject = (reason = undefined, returnIt = true) => {
```
erights, line 122: Why remove the `async`? AFAICT, it should only make a difference if the body synchronously throws. In that case, would you prefer to propagate a throw or a rejection?

It is a common convention that a function whose only successful results are promises indicates problems as rejections rather than throws, which `async` reliably enforces.

mhofman: If I had to guess it's to make sure we don't create an unhandled rejection. The promise that we return is handled, but it can't be if this function is async, because the rejection would propagate to the different promise created intrinsically by the async function.

michaelfig: Yes, @mhofman is correct.  That's what the "quiet" in "quietReject" refers to.

erights: resolving

erights (approved):
Since gcImports is turned off by default in this PR, and since our immediate short term plans are to use it with gcImports off, I did not do a thorough review of the code that would only be relevant to this case. Nevertheless, I feel comfortable enough to approve. 

I did not see any problems with that code. I just don't understand it well enough yet to be confident in it. (distributed gc remains hard to think about.) Before making gcImports on by default, or using it in agoric-sdk with gcImports on, I'd like to review again.

That said, LGTM. Thanks!

PR #1552, opened 2023-04-11
feat(init): add `@endo/init/unsafe-fast.js`

michaelfig:
As discussed with @erights in a private meeting:

Create `@endo/init/unsafe-fast.js` to use a performant, but less-than-safe SES lockdown, currently

`lockdown({ __hardenTaming__: 'unsafe' })`


erights (approved):
LGTM, thanks!

PR #1553, opened 2023-04-12
Default `TRACK_TURNS=disabled`

michaelfig:
As discussed at Agoric.

Also, fix a layering mistake in which hardened results required `TRACK_TURNS=enabled`.


erights (approved):
I'm not sure I understand why this is a more correct layering, but I'm certainly ok with it. LGTM, thanks!

PR #1554, opened 2023-04-13
feat(pass-style,exo): label remotable instances

erights:
Replaces https://github.com/Agoric/agoric-sdk/pull/6218 on the endo side. Similar LABEL_INSTANCE logic will then be needed on the agoric-sdk side using `baseRef` and `facetNames` as https://github.com/Agoric/agoric-sdk/pull/6218 does.

Stay in sync with https://github.com/Agoric/agoric-sdk/pull/7409

### DIFF
--- a/packages/exo/src/exo-makers.js
+++ b/packages/exo/src/exo-makers.js
@@ -1,8 +1,31 @@
+/* global globalThis */
 import { objectMap } from '@endo/patterns';
 
 import { defendPrototype, defendPrototypeKit } from './exo-tools.js';
 
-const { seal, freeze } = Object;
+const { seal, freeze, defineProperty } = Object;
+
+// TODO Use environment-options.js currently in ses/src after factoring it out
+// to a new package.
+const env = (globalThis.process || {}).env || {};
+
+// Turn on to give each exo instance its own toStringTag value.
+const LABEL_INSTANCES = (env.DEBUG || '')
+  .split(':')
+  .includes('label-instances');
+
+const makeSelf = (proto, instanceCount) => {
+  const self = { __proto__: proto };
+  if (LABEL_INSTANCES) {
+    defineProperty(self, Symbol.toStringTag, {
+      value: `${proto[Symbol.toStringTag]}#${instanceCount}`,
+      writable: false,
+      enumerable: false,
+      configurable: false,
+    });
+  }
+  return harden(self);
+};
 
 const emptyRecord = harden({});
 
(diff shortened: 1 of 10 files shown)


packages/exo/src/exo-makers.js
```diff
@@ -133,8 +155,18 @@ export const defineExoClassKit = (
     const state = seal(init(...args));
     // Don't freeze context until we add facets
     const context = { state };
+    instanceCount += 1;
     const facets = objectMap(prototypeKit, (prototype, facetName) => {
-      const self = harden({ __proto__: prototype });
+      const self = { __proto__: prototype };
+      if (LABEL_INSTANCES) {
+        defineProperty(self, Symbol.toStringTag, {
+          value: `${prototype[Symbol.toStringTag]}#${instanceCount}`,
```
turadg, line 163: consider DRYing these two blocks with something like,
```
const makeSelf = (prototype, instanceCount) {
  const self = { __proto__: prototype };
  if (LABEL_INSTANCES) {
    defineProperty(self, Symbol.toStringTag, {
      value: `${prototype[Symbol.toStringTag]}#${instanceCount}`,
      writable: false,
      enumerable: false,
      configurable: false,
    });
  }
  return harden(self);
}
```

packages/exo/test/test-label-instances.js
```diff
@@ -0,0 +1,117 @@
...
+  makeExo,
+} from '../src/exo-makers.js';
+
+const { quote: q } = assert;
+
+const UpCounterI = M.interface('UpCounter', {
+  incr: M.call()
+    // TODO M.number() should not be needed to get a better error message
+    .optional(M.and(M.number(), M.gte(0)))
+    .returns(M.number()),
```
turadg, line 17: type signature not relevant to this test.

consider simplifying 
```suggestion
  incr: M.call().returns(undefined),
```

or simply omitting the interface guards

turadg (approved):

turadg:
Note that for the changelog this is a new `feat` not a `fix`.

packages/exo/test/test-label-instances.js
```diff
@@ -0,0 +1,117 @@
+// eslint-disable-next-line import/order
+import { test } from './prepare-test-env-ava-label-instances.js';
+import { passStyleOf } from '@endo/far';
```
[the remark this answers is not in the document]
erights: TIL that newlines are significant to import order linting, thanks.

Done

packages/exo/test/test-label-instances.js
erights: Got rid of the optional arg as suggested, but kept the return, so the incremented value would still be observable. That observation isn't actually relevant, but without it the code is confusingly vacuous.

Resolving.

erights:
> Note that for the changelog this is a new `feat` not a `fix`.

Done

PR #1556, opened 2023-04-17
feat(ses): use the start compartment's `globalThis.harden` if available

michaelfig:
Allows the start compartment to provide `globalThis.harden` to override the `harden` used by the SES shim.
This enables some hosts (like Moddable's XS) to provide their own (likely more performant) `harden` implementations.


erights (approved):
LGTM, thanks!

Issue #1478, opened 2023-01-28
Formally define and consistently use "CapData [record]"/"Smallcaps [encoding]"/etc.


gibson042:
Now that we have an implicit definition of "CapData" as a `{ body: string, slots: copyable[] }` structure with constraints on `body` (that will hopefully be [relaxed](https://github.com/endojs/endo/pull/1402#discussion_r1044617758)), we should make that **explicit** (which may involve naming the `@qclass`-based encoding that preceded smallcaps).

From https://github.com/endojs/endo/pull/1402#pullrequestreview-1273083220 :
> This PR is implicitly defining **CapData** as a record structure that currently encompasses a `body` string and a `slots` array, in which the encoding of `body` is implicitly either "_Smallcaps_"<sup>1</sup> if the first character is "#" or [_unnamed original encoding_]<sup>2</sup> otherwise.
> 
> ## Footnotes
> 1. [JSON text](https://www.rfc-editor.org/rfc/rfc8259#section-2) in which arbitrary [pass-by-copy](https://docs.agoric.com/guides/js-programming/far.html#pass-styles-and-harden) source data has been made representable mostly by translation of JSON-incompatible values into "#…" strings
> 2. [JSON text](https://www.rfc-editor.org/rfc/rfc8259#section-2) in which arbitrary [pass-by-copy](https://docs.agoric.com/guides/js-programming/far.html#pass-styles-and-harden) source data has been made representable mostly by translation of JSON-incompatible values into objects with `@qclass` members

mhofman:
I still believe that even `{body: Copyable, slot: string[]}` is a too restrictive definition for future CapData. When relaxing from using a serialized body, I'm not convinced we won't want to pick a different property name than `body` to hold the encoded but not serialized body.

I think however that `slot` will not change, and often some routing layers will be able to inspect/update slots while passing through the other properties, whatever their encoding / serialization is. We should however make it clear that any other property could hold "copyable" data, not just strings.

If we want we can place a restriction that any future body related property will have its name start with `body`, and that properties with other names represent invalid CapData.

[...]

PR #1558, opened 2023-04-19
feat(marshal): cheap capData stringify and parse

erights:
Given that `capData` is a CapData whose `slots` are themselves simple strings (see def in PR), and whose `body` is a smallcaps string, then `stringifyCapData(capData)` cheaply stringifies the capData as a whole to a JSON-parsable string, while avoiding the double encoding of the body. This should make the body considerable shorter and more readable.

`parseCapData` cheaply does the reverse of course.

Example:
```js
  roundTrips(t, { body: '#[]', slots: ['a'] }, '{"slots":["a"],"#body":[]}');
```

@arirubinstein this uses a regexp. Are both directions safe against malicious input? @gibson042 , does this seem fastcheck-able?

Everyone, assume this is not urgent. But it is small and, if robust, would be pleasant.

### DIFF
--- a/packages/marshal/src/stringify-capdata.js
+++ b/packages/marshal/src/stringify-capdata.js
@@ -1,0 +1,54 @@
+/** @template S @typedef {import('./types.js').CapData<S>} CapData */
+
+const { Fail, quote: q } = assert;
+
+/** @typedef {string} SimpleString */
+
+/**
+ * A SimpleString is one where JSON.strigify-ing the string just
+ * puts quotes around the contents, and it contains no close square
+ * brackets. Many slot encodings use slot strings that
+ * obey these rules, which is why they are useful.
+ *
+ * @param {unknown} str
+ * @returns {asserts str is SimpleString}
+ */
+export const assertSimpleString = str => {
+  assert(typeof str === 'string');
+  const stringified = JSON.stringify(str);
+  const expected = `"${str}"`;
+  stringified === expected ||
+    Fail`Expected to stringify to ${q(expected)}, not ${q(stringified)}}`
+  assert(str.indexOf(']') === -1);
+};
+harden(assertSimpleString);
+
+/**
+ * @param {CapData<SimpleString>} capData
+ * @returns {string}
+ */
+export const stringifyCapData = capData => {
+  const { body, slots: [...slots] } = capData;
+  assert(typeof body === 'string');
+  assert(body.startsWith('#'));
+  const scBody = body.slice(1);
+  JSON.parse(scBody); // Just asserts that it parses as JSON
+  slots.forEach(assertSimpleString);
+  return `{"slots":${JSON.stringify(slots)},"#body":${scBody}}`;
+}
+harden(stringifyCapData);
+
+const DSL = /^\{"slots":(\[[^\]]*\]),"#body":(.*)\}$/;
+
+export const parseCapData = str => {
+  assert(typeof str === 'string');
+  const matches = DSL.exec(str);
+  assert(matches && matches.length === 3);
+  const slots = JSON.parse(matches[1]);
+  slots.forEach(assertSimpleString);
+  const scBody = matches[2];
+  JSON.parse(scBody); // Just asserts that it parses as JSON
+  const body = `#${scBody}`;
+  return harden({ body, slots });
+};
+harden(parseCapData);
--- a/packages/marshal/src/stringify-capdata.js
+++ b/packages/marshal/src/stringify-capdata.js
@@ -18,7 +18,7 @@ export const assertSimpleString = str => {
   const stringified = JSON.stringify(str);
   const expected = `"${str}"`;
   stringified === expected ||
-    Fail`Expected to stringify to ${q(expected)}, not ${q(stringified)}}`
+    Fail`Expected to stringify to ${q(expected)}, not ${q(stringified)}}`;
   assert(str.indexOf(']') === -1);
 };
 harden(assertSimpleString);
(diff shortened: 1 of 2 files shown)


erights:
Won't really be cheap unless we can remove the calls to `JSON.parse(scBody)`, which is there just to validate. I conjecture that it is robust even with these checks removed. But I am not yet confident of that.

packages/marshal/src/stringify-capdata.js
```diff
@@ -0,0 +1,57 @@
...
+  assert(typeof body === 'string');
+  assert(body.startsWith('#'));
+  const scBody = body.slice(1);
+  JSON.parse(scBody); // Just asserts that it parses as JSON
+  slots.forEach(assertSimpleString);
+  return `{"slots":${JSON.stringify(slots)},"#body":${scBody}}`;
+};
+harden(stringifyCapData);
+
+const DSL = /^\{"slots":(\[[^\]]*\]),"#body":(.*)\}$/;
```
arirubinstein, line 44: I would recommend to move to a less greedy search and string boundaries if possible
`\A\{"slots":(\[[^\]]+?\]),"#body":(.+?)\}\Z`

packages/marshal/src/stringify-capdata.js
```diff
@@ -0,0 +1,57 @@
...
+  const stringified = JSON.stringify(str);
+  const expected = `"${str}"`;
+  stringified === expected ||
+    Fail`Expected to stringify to ${q(expected)}, not ${q(stringified)}}`;
+  assert(str.indexOf(']') === -1);
+};
+harden(assertSimpleString);
+
+/**
+ * @param {CapData<SimpleString>} capData
```
turadg, line 27: since `SimpleString = String` this won't give a type error for plain string.

if you want to ensure that `assertSimpleString` was called first, the SimpleString type could have a tag so it's a superset of a regular string: The [`Opaque` utility from type-fest](https://github.com/sindresorhus/type-fest/blob/main/source/opaque.d.ts) makes this easy

packages/marshal/src/stringify-capdata.js
```diff
@@ -0,0 +1,57 @@
+/** @template S @typedef {import('./types.js').CapData<S>} CapData */
+
+const { Fail, quote: q } = assert;
+
+/** @typedef {string} SimpleString */
+
+/**
+ * A SimpleString is one where JSON.strigify-ing the string just
```
turadg, line 8:
```suggestion
 * A SimpleString is one where JSON.stringify-ing the string just
```

packages/marshal/src/stringify-capdata.js
erights: I had never seen `\A` or `\Z` before. Searching for it in JS RegExp documentation finally found it at https://docstore.mik.ua/orelly/webprog/jscript/ch10_01.htm in a section titled 
"Perl RegExp Features Not Supported in JavaScript"

But good point about newlines. Where the `(.*)` appears, we should accept newlines along with everything else. IIUC, adding the `s` flag (aka "dotAll") on the end will give "." that meaning. Of course there are other ways to get the same effect by changing the pattern rather than adding a flag. What do you recommend?

mhofman:
While the approach is interesting, before we merge this I'd like to have a discussion about the use cases.

FUDCo:
While getting rid of the double encoding is definitely a benefit, I'm not sure I buy the "this is cheap" assertion.  All the pre-flight and post-flight checking seem to me to add up to considerably more overhead than what this replaces (though of course we'd need to do some benchmarking to be sure).  We might decide that the gain in clarity and simplicity in the encoding is worth it, but if so we should make that decision with our eyes open.

I'd also second @mhofman's concern about use cases.  This smells to me like it may be overly fixated on one particular use case (albeit an important one) -- Note that I don't as yet have a good justification for that sensation, I just have it.

erights:
Converted to draft until we understand 
   * if this is needed
   * if it is safe with the extra `JSON.parse(scBody)` validation step
   * if it is cheap when that check is absent.

We do not yet have answers for any of those.

In the meantime, I will continue to keep this warm -- to revise and respond to reviewer comments as if this is an Open PR.

Thanks.

erights:
https://github.com/Agoric/agoric-sdk/pull/6646 has the following code
```js
          return JSON.stringify(readonly.serialize(val));
...
          const data = JSON.parse(str);
          return readonly.unserialize(data);
```
Aside from a name change, IIUC, this would also be part of the plan at https://github.com/Agoric/agoric-private/issues/47

Would these cases be better without the double encoding? Should we care?

gibson042 (commented):
This definitely looks possible to cover with fastcheck, and `#body` seems like a fine way to differentiate a non-serialized payload (thereby reducing the size and increasing the human-friendliness of messages) but I share the thoughts of others regarding this approach to getting there.

mhofman:
> If we do this, I hope it's in a way that expands the definition of "CapData" from `{ body: string, slots: Copyable[] }` to e.g. `{ body: string, slots: Copyable[] } | { "#body": Copyable<encoding: 'smallcaps'>, slots: Copyable[] }` and allows for direct production of the latter (Passable → CapData<format: 'smallcapsObject'> from e.g. `marshaller.toCapData(passable)`, with subsequent `JSON.stringify` as warranted [and possibly after embedding in another structure]) rather than imposing a multi-step and non-embeddable process (Passable → CapData<bodyFormat: 'smallcapsString'> → JSONText).

Mark and I discussed this PR yesterday. The disconnect was indeed how CapData is usually nested into a larger structure, and where the serialization through JSON parse is usually disconnected and in a completely unrelated module (e.g. just before sending it over a socket).

However this serialization effectively is a regular JSON serialization of an object with a differently named `#body` property, and is actually really close to the generalized CapData definition described above and in https://github.com/endojs/endo/issues/1478

erights:
In light of @mhofman's compelling arguments that this will not be needed, closing.

PR #1560, opened 2023-04-20
Release 2023-04-20 19:45:20Z

michaelfig:
- @michaelfig - exo: Parse `$DEBUG` as comma-separated
- @michaelfig - ses: Pass through the start compartment's `globalThis.harden` if defined
- @michaelfig - Enable JS type checking by default in the monorepo, and fix types accordingly


dckc (approved):
looks straightforward and mechanical. I presume this follows established/documented norms.

PR #1562, opened 2023-04-22
refactor: Error constructor does not need new

erights:
A step towards turning on more `@jessie-check` directives. See https://github.com/Agoric/agoric-sdk/pull/3876 and https://github.com/Agoric/agoric-sdk/pull/5497

Jessie prohibits `new`. One trivial case of revising code to conform is to remove the `new` on calls to the `*Error(` constructors. For these specifically, their construct behavior is the same as their call behavior. This also gets rid of a bit of visual noise, make code a bit more readable.

### DIFF
--- a/packages/base64/btoa.js
+++ b/packages/base64/btoa.js
@@ -8,7 +8,7 @@ export const btoa = stringToEncode => {
   const bytes = stringToEncode.split('').map(char => {
     const b = char.charCodeAt(0);
     if (b > 0xff) {
-      throw new Error(`btoa: character out of range: ${char}`);
+      throw Error(`btoa: character out of range: ${char}`);
     }
     return b;
   });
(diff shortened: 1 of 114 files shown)


erights:
As discovered from looking at CI errors on https://github.com/Agoric/agoric-sdk/pull/7481 , not all `*Error`s mentioned by our sources are builtin errors, or adequately like builtin errors. Specifically, `CommanderError` is a JS subclass of `Error` and `InvalidArgumentError` is a JS subclass of `CommanderError`. These are both apparently from https://github.com/tj/commander.js/ , which is used by both endo cli and agoric-sdk agoric-cli.

However, neither of these classes is mentioned in a problematic way in endo, and so this PR succeeded under CI the first time.

FUDCo (approved):
It's like deja vu all over again.

PR #1567, opened 2023-04-29
doc(pass-style): property name cases table

erights:
Adapt @warner's table from https://github.com/Agoric/agoric-sdk/issues/7446 as replacement for nested list.

Eventually, see also http://arejsobjectscursed.com/

Note to reviewers: Low urgency


FUDCo (approved):
Loverly.

PR #1571, opened 2023-05-01
fix(patterns): mismatch errors should not redact the pattern

erights:
If we accept this PR, we also need to make it clear, *somewhere findable*, that secrets should not be placed into patterns (other than keys-as-patterns), since they will not be redacted in error messages. But where?

### DIFF
--- a/packages/patterns/src/patterns/patternMatchers.js
+++ b/packages/patterns/src/patterns/patternMatchers.js
@@ -273,6 +273,8 @@ const makePatternKit = () => {
     }
     return (
       check !== identChecker &&
+      // When the mismatch occurs against a key used as a pattern,
+      // the pattern should still be redacted.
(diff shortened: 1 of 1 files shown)


packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -273,6 +273,8 @@ const makePatternKit = () => {
     }
     return (
       check !== identChecker &&
+      // When the mismatch occurs agains a key used as a pattern,
```
FUDCo, line 276: s/agains/against/

FUDCo (approved):
Seems fine modulo a comment typo.
I totally hate when prettier puts line breaks in the middle of a template string.

packages/patterns/src/patterns/patternMatchers.js

PR #1575, opened 2023-05-03
test(marshal): ava deepEqual on remotables works for testing

erights:
Demonstrates how to use ava `t.deepEqual` on remotables (and therefore on structures containing remotables).


dckc (approved):
Nifty!

PR #1579, opened 2023-05-04
feat(ses): tame Symbol so whitelist works

### LINKED ISSUE #1577: Shared Symbol intrinsic
Currently, Compartments share `globalThis.Symbol` with the start compartment, the realm global environment. Consequently, if the host has any symbols that are not recognized by SES’s permits, Lockdown will fail.

We should instead create a separate `Symbol` shared intrinsic, like the repaired `Function` or `Date`, that shares the global prototype of `Symbol`, disables the shared prototype constructor, and only includes the symbols that SES will permit in shared compartments.

erights:
Fixes #1577 

If the ses shim initializes on a platform in which Symbol contains non-configurable properties absent from the whitelist, `lockdown` will effectively succeed at deleting them anyway.

This is a concern because later versions of the language, as well as bespoke variations, may add symbols whose meaning is currently unknown. The symbols themselves would be safe, be we should know what behavior they unlock before making them generally available.

### DIFF
--- a/packages/ses/src/lockdown-shim.js
+++ b/packages/ses/src/lockdown-shim.js
@@ -264,10 +264,6 @@ export const repairIntrinsics = (options = {}) => {
 
   tameDomains(domainTaming);
 
-  const SharedSymbol = tameSymbolConstructor();
-  // Must happen before `makeIntrinsicsCollector()`
-  globalThis.Symbol = SharedSymbol;
-
   const { addIntrinsics, completePrototypes, finalIntrinsics } =
     makeIntrinsicsCollector();
 
@@ -280,6 +276,7 @@ export const repairIntrinsics = (options = {}) => {
   addIntrinsics(tameErrorConstructor(errorTaming, stackFiltering));
   addIntrinsics(tameMathObject(mathTaming));
   addIntrinsics(tameRegExpConstructor(regExpTaming));
+  addIntrinsics(tameSymbolConstructor());
 
   addIntrinsics(getAnonymousIntrinsics());
 
--- a/packages/ses/src/lockdown-shim.js
+++ b/packages/ses/src/lockdown-shim.js
@@ -53,6 +53,7 @@ import { makeEnvironmentCaptor } from './environment-options.js';
 import { getAnonymousIntrinsics } from './get-anonymous-intrinsics.js';
 import { makeCompartmentConstructor } from './compartment-shim.js';
 import { tameHarden } from './tame-harden.js';
+import { tameSymbolConstructor } from './tame-symbol-constructor.js';
 
 /** @typedef {import('../types.js').LockdownOptions} LockdownOptions */
 
@@ -262,6 +263,10 @@ export const repairIntrinsics = (options = {}) => {
    */
 
   tameDomains(domainTaming);
+
+  const SharedSymbol = tameSymbolConstructor();
+  // Must happen before `makeIntrinsicsCollector()`
+  globalThis.Symbol = SharedSymbol;
 
   const { addIntrinsics, completePrototypes, finalIntrinsics } =
     makeIntrinsicsCollector();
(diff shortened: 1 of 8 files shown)


packages/ses/src/lockdown-shim.js
```diff
@@ -263,6 +264,10 @@ export const repairIntrinsics = (options = {}) => {
 
   tameDomains(domainTaming);
 
+  const SharedSymbol = tameSymbolConstructor();
+  // Must happen before `makeIntrinsicsCollector()`
+  globalThis.Symbol = SharedSymbol;
```
kriskowal, line 269: I think this has the unintended effect of erasing powerful symbols from the start compartment, so they can’t be delegated selectively by an application. Should we leave the start compartment’s `Symbol` in tact?

kriskowal (commented):
Should we leave the `globalThis.Symbol` in tact so that the application, being in a position of maximum host powers, has the option of delegating any symbol to a compartment and only provide a different, shared `Symbol` in compartments?

packages/ses/src/lockdown-shim.js
erights: Good suggestion. This would also make this taming more consistent with our other tamings. Will do.

erights: Fixed. PTAL

packages/ses/src/tame-symbol-constructor.js
```diff
@@ -0,0 +1,65 @@
...
+ * Starting these properties as configurable assumes two succeeding phases of
+ * processing: A whitelisting phase, that
+ * removes all properties not on the whitelist (which requires them to be
+ * configurable) and a global hardening step that freezes all primordials,
+ * returning these properties to their expected non-configurable status.
+ *
+ * The ses shim is constructed to eventually enable vetted shims to run between
+ * repair and global hardening. However, such vetted shims would normally
+ * run in the start compartment, which continues to use the original unmodified
+ * `Symbol`, so they should not normally be affected by the remporary
```
gibson042, line 26:
```suggestion
 * `Symbol`, so they should not normally be affected by the temporary
```

erights: Done. PTAL

packages/ses/src/tame-symbol-constructor.js
```diff
@@ -0,0 +1,65 @@
...
+ * the start compartment's global. The tamed alternative is used as the shared
+ * `Symbol` constructor on constructed compartments.
+ *
+ * Starting these properties as configurable assumes two succeeding phases of
+ * processing: A whitelisting phase, that
+ * removes all properties not on the whitelist (which requires them to be
+ * configurable) and a global hardening step that freezes all primordials,
+ * returning these properties to their expected non-configurable status.
+ *
+ * The ses shim is constructed to eventually enable vetted shims to run between
```
kriskowal, line 23: Implies that we’ll need to expose our intrinsics as well as the mechanism for altering the permits.

packages/ses/test/test-whitelist-intrinsics.js
```diff
@@ -7,7 +7,11 @@ if (!eval.toString().includes('native code')) {
   throw TypeError('Module "esm" enabled: aborting');
 }
 
-test('whitelistIntrinsics - Well-known symbols', t => {
+// This test as stated no longer makes sense, now that `whitelistIntrinsics`
+// decodes symbol strings using the real `globalThis.Symbol` rather than
+// `intrinsics.Symbol`. But I left this in as a `test.skip` in case anyone
+// wants to rescure some of the purpose of this test.
```
kriskowal, line 13:
```suggestion
// wants to rescue some of the purpose of this test.
```

To your credit, I looked it up first.

kriskowal (approved):

packages/ses/src/tame-symbol-constructor.js
erights: I think so

PR #1585, opened 2023-05-10
feat(env-options): env-options as separate importable package

erights:
Lets us do the many `// TODO Use environment-options.js`, rather than hand rolling `process.env[name]` handling each time.

### DIFF
--- a/packages/env-options/README.md
+++ b/packages/env-options/README.md
@@ -1,0 +1,75 @@
+# Parameterizing Modules with Environment Options
+
+JavaScript module semantics resist attempts to parameterize a module's
+initialization behavior. A module initializes in order according to
+the path by which it is first imported, and then the initialized module
+is reused by all the other times it is imported. Compartments give us
+the opportunity to bind the same import name to different imported
+modules, depending on the package/compartment doing the import. Compartments
+also address the difficulty of parameterizing a module's initialization
+logic, but not in a pleasant manner.
+
+A pleasant parameterization would be for a static module to be function-like
+with explicit parameters, and for the parameterization to be like
+calling the static module with parameters in order to derive from it a
+module instance. Compartments instead lets us parameterize the meaning
+of a module instance derived from a static module according to the
+three namespaces provided by the JavaScript semantics, affecting the
+meaning of a module instance.
+   * The global variable namespaces.
+      * The global scope, aliased to properties of the global object.
+        This is necessarily compartment-wide. In our
+        recommened usage pattern of one compartment per package,
+        each global would be package-wide. (See LavaMoat)
+      * The global lexical scope. The SES-shim compartments support
+        these both compartment-wide as well as per-module. But it is
+        not yet clear what we will propose in the Compartment proposal.
+   * The import namespace.
+   * The host hooks.
+
+This `@endo/env-options` package follows the Node precedent for
+finding Unix environment variable settings: looking for a
+global `process` object holding an `env` object,
+optionally holding a property with the same name as the option,
+whose value is the configuration setting of that option.
+
+```js
+import { makeEnvironmentCaptor } from '@endo/env-options';
+const { getEnvironmentOption } = makeEnvironmentCaptor(globalThis);
+const FooBarOption = getEnvironmentOption('FOO_BAR', 'absent');
+```
+
+The first argument to `getEnvironmentOption` is the name of the option.
+The value of `FooBarOption` would then be the value of
+`globalThis.process.env.FOO_BAR`, if present.
+If setting is either absent or `undefined`, the default `'absent'`
+would be used instead.
+
+In either case, reflecting Unix environment variable expectations,
+the resulting setting must be a string.
+This restriction also helps ensure that this channel is used only to pass data,
+not authority beyond the ability to read this global state.
+
+The `makeEnvironmentCaptor` function also returns a
+`getCapturedEnvironmentOptionNames` function for use to give feedback about
+which environment variables were actually read, for diagnostic purposes. The
+ses-shim `lockdown` contains code such as the following, to explain which
+environment variables were read to provide `lockdown` settings.
+
+```js
+import { makeEnvironmentCaptor } from '@endo/env-options';
+const {
+  getEnvironmentOption,
+  getCapturedEnvironmentOptionNames,
+} = makeEnvironmentCaptor(globalThis);
+...
+const capturedEnvironmentOptionNames = getCapturedEnvironmentOptionNames();
+if (capturedEnvironmentOptionNames.length > 0) {
+  console.warn(
+    `SES Lockdown using options from environment variables ${enJoin(
+      arrayMap(capturedEnvironmentOptionNames, q),
+      'and',
+    )}`,
+  );
+}
+```
--- a/packages/env-options/package.json
+++ b/packages/env-options/package.json
@@ -1,11 +1,12 @@
 {
-  "name": "@endo/exo",
-  "version": "0.2.2",
-  "description": "exo: remotable objects protected by interface guards.",
+  "name": "@endo/env-options",
+  "version": "0.1.0",
+  "private": null,
+  "description": "Reading environment options.",
   "keywords": [],
   "author": "Endo contributors",
   "license": "Apache-2.0",
-  "homepage": "https://github.com/endojs/endo/blob/master/packages/exo/README.md",
+  "homepage": "https://github.com/endojs/endo/tree/master/packages/env-options#readme",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/endojs/endo.git"
@@ -16,38 +17,36 @@
   "type": "module",
   "main": "./index.js",
   "module": "./index.js",
+  "browser": null,
+  "unpkg": null,
+  "types": null,
   "exports": {
     ".": "./index.js",
-    "./tools.js": "./tools.js",
     "./package.json": "./package.json"
   },
   "scripts": {
     "build": "exit 0",
-    "prepack": "tsc --build jsconfig.build.json",
-    "postpack": "git clean -f '*.d.ts*'",
     "lint": "yarn lint:types && yarn lint:js",
     "lint-fix": "eslint --fix .",
     "lint:js": "eslint .",
     "lint:types": "tsc -p jsconfig.json",
+    "postpack": "git clean -f '*.d.ts*'",
+    "prepack": "tsc --build jsconfig.build.json",
     "test": "ava"
   },
-  "dependencies": {
-    "@endo/env-options": "^0.1.0",
-    "@endo/far": "^0.2.18",
-    "@endo/patterns": "^0.2.2"
-  },
+  "dependencies": {},
   "devDependencies": {
-    "@endo/init": "^0.5.56",
-    "@endo/ses-ava": "^0.2.40",
+    "@endo/init": "^0.5.53",
+    "@endo/ses-ava": "^0.2.37",
     "ava": "^5.2.0",
     "babel-eslint": "^10.0.3",
-    "eslint": "^7.32.0",
+    "eslint": "^7.23.0",
     "eslint-config-airbnb-base": "^14.0.0",
     "eslint-config-prettier": "^8.8.0",
     "eslint-plugin-eslint-comments": "^3.1.2",
-    "eslint-plugin-import": "^2.27.5",
+    "eslint-plugin-import": "^2.19.1",
     "prettier": "^2.8.5",
-    "typescript": "~4.9.5"
+    "typescript": "~4.8.4"
   },
   "files": [
     "*.js",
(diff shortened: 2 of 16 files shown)


packages/env-options/README.md
gibson042, line 1: Could we get something in the README?

packages/env-options/package.json
```diff
@@ -0,0 +1,73 @@
+{
+  "name": "@endo/env-options",
+  "version": "0.1.0",
+  "private": null,
+  "description": "Description forthcoming.",
```
gibson042, line 5: I wonder if we should have the package generation script detect an interactive terminal to prompt for this kind of data.

packages/env-options/src/environment-options.js
```diff
@@ -85,7 +91,7 @@ export const makeEnvironmentCaptor = aGlobal => {
       const globalEnv = globalProcess.env;
       if (globalEnv && typeof globalEnv === 'object') {
         if (optionName in globalEnv) {
-          arrayPush(capturedEnvironmentOptionNames, optionName);
+          capturedEnvironmentOptionNames.push(optionName);
```
gibson042, line 94: This and the above `Fail` introduce late property access on built-ins; is that acceptable?

packages/env-options/test/test-index.js
```diff
@@ -0,0 +1,5 @@
+import { test } from './prepare-test-env-ava.js';
+
+test('place holder', async t => {
+  t.pass();
```
gibson042, line 4: We probably want `yarn test` to fail in the initial state: #1586

gibson042 (commented):
Aren't there more files that should make use of this?

packages/env-options/README.md
erights: Done

packages/env-options/package.json
erights: @kriskowal , @gibson042 is also making a suggestion here for the script.

In any case, fixed here, so resolving.

packages/env-options/src/environment-options.js
erights: @kriskowal , what do you think? Leaving this conversation open waiting for your answer.

erights:
> Aren't there more files that should make use of this?

Done: track-turns.js and exo-makers.js

@michaelfig , please let me know if you'd like me to back out the changes to eventual-send because you want it to remain dependency-free.

packages/env-options/src/environment-options.js
kriskowal: There is a layering problem. We impose mutually exclusive requirements of libraries that operate above, within, and below SES in the layers of the dependency graph. This change as written uses `env-options` as both within and above SES. That is, it’s used within SES and also above SES.

We currently require anything within SES to defend itself from mutations to the global environment that may occur in the start compartment between the initialization of SES and execution of lockdown. This is not strictly necessary, but we have maintained that high bar within SES using the `no-polymorphic-call` eslint rule. By taking `env-options` as a dependency of `ses`, it must meet that bar.

In accordance with being below SES, this code also rightly duplicates `Fail`, which it would not need to do if it were a library that assumes the pre-arrangement of a SES environment.

I appreciate avoiding duplication in source, but as a practical matter, the SES shim gets bundled in some environments, so this library would necessarily be duplicated in production regardless. That somewhat weakens the case for having a single shared implementation between SES internals and for use in SES environments. Not entirely, though.

In any case, acceptable outcomes in my opinion are:
1. `env-options` is largely duplicative of similar code in `ses` but not held to the `no-polymorphic-call` rule.
2. `env-options` continues to be used both within and above SES, in which case it must include `no-polymorphic-call` in its `eslint` config.

kriskowal (approved):

packages/env-options/src/environment-options.js
erights: Done

packages/env-options/src/env-options.js
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * `makeEnvironmentCaptor` provides a mechanism for getting environment
+ * variables, if they are needed, and a way to catalog the names of all
+ * the environment variables that were captured.
+ *
+ * @param {object} aGlobal
+ */
+export const makeEnvironmentCaptor = aGlobal => {
+  const capturedEnvironmentOptionNames = [];
```
gibson042, line 42: I think we also want to eagerly capture the `env` object.
```suggestion
  const globalProcess = aGlobal.process;
  const globalEnv = globalProcess && typeof globalProcess === 'object'
    ? globalProcess.env
    : undefined;
  const capturedEnvironmentOptionNames = [];
```

packages/env-options/src/env-options.js
```diff
@@ -0,0 +1,98 @@
...
+
+    /** @type {string} */
+    let setting = defaultSetting;
+    const globalProcess = aGlobal.process;
+    if (globalProcess && typeof globalProcess === 'object') {
+      const globalEnv = globalProcess.env;
+      if (globalEnv && typeof globalEnv === 'object') {
+        if (optionName in globalEnv) {
+          arrayPush(capturedEnvironmentOptionNames, optionName);
+          const optionValue = globalEnv[optionName];
```
gibson042, line 70: Performing this lookup against `globalEnv` in every call creates the possibility of results changing—should we instead memoize them by `optionName` (e.g., replacing the `capturedEnvironmentOptionNames` array into a `capturedEnvironmentOptions` map)?

packages/env-options/src/env-options.js
```diff
@@ -0,0 +1,98 @@
...
+            )}`;
+          setting = optionValue;
+        }
+      }
+    }
+    setting === undefined ||
+      typeof setting === 'string' ||
+      Fail`Environment option value ${q(
+        setting,
+      )}, if present, must be a string.`;
```
gibson042, line 86: I believe this section is unnecessary, since `setting` is initialized to `defaultSetting` after the latter is confirmed to be a string and is only ever reassigned to `optionValue` after _that_ is confirmed to be a string.
```suggestion
```

packages/env-options/src/env-options.js
erights: See answer below to your "should we instead memoize" question.

Resolving.

packages/env-options/src/env-options.js
erights: Options setting code, to be tolerant of the absence of a `globalThis.process.env` object, currently looks like
```js
const env = (globalThis.process || {}).env || {};
env.TRACK_TURNS = 'enabled';
```
This kinda defeats one of the purposes of this package, which is to raise the abstraction level.
In a later PR I intend to provide an API for setting options. That one will determine our stance on these temporal issues. In the meantime, the current code is a good starting point for what I expect to do.

Resolving.

packages/env-options/src/env-options.js
erights: Nice catch! Done.

packages/env-options/README.md
```diff
@@ -0,0 +1,75 @@
+# Parameterizing Modules with Environment Options
+
+JavaScript module semantics resists attempts to parameterize a module's
```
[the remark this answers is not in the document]
kriskowal: Semantics is indeed a singular, but only because English adopted Greek -ikos suffix inconsistently (-ic or -ics) and we should be more angry about it.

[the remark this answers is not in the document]
kriskowal: The principled way to express our anger is to deliberately spell it “semantology” but this is not the time or place 😉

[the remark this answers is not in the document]
erights: So should this be `resist` or `resists`?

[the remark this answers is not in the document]
erights: Unresolving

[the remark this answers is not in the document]
kriskowal: `semantics resist` is correct.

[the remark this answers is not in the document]
erights: Cool. Reresolving.

packages/env-options/src/env-options.js
```diff
@@ -0,0 +1,93 @@
...
+const arrayPush = uncurryThis(Array.prototype.push);
+const arrayJoin = uncurryThis(Array.prototype.join);
+
+const q = JSON.stringify;
+
+const Fail = (template, ...args) => {
+  const msg = arrayFlatMap(args, (arg, i) => [template[i], arg]);
+  arrayPush(msg, template[args.length]);
+  throw Error(arrayJoin(msg, ''));
+};
```
gibson042, line 29: How much do we care about performance here? Assuming "_not much because this only applies for generating exceptions_", I think the code could be a bit more clear. Some options:

```js
const Fail = (literals, ...args) => {
  const msg = arrayReduce(literals, (head, str, i) => `${head}${args[i - 1]}${str}`);
  throw Error(msg);
};
```


```js
const Fail = (literals, ...args) => {
  const msgParts = arrayMap(literals, (str, i) =>
    i === 0 ? str : `${args[i - 1]}${str}`
  );
  throw Error(arrayJoin(msgParts, ""));
};
```

```js
const Fail = (literals, ...args) => {
  let msg = literals[0];
  for (let i = 0; i < args.length; i += 1) {
    msg = `${msg}${args[i]}${literals[i + 1]}`;
  }
  throw Error(msg);
};
```

gibson042 (approved):
One small code organizational question, but LGTM no matter how it's resolved.

packages/env-options/src/env-options.js
kriskowal: I believe the last of these options to be likely both the fastest and clearest.

erights:
Reviewers, at https://github.com/endojs/endo/actions/runs/5034078931/jobs/9028682745?pr=1585 I get

```
npm ERR! code E404
npm ERR! 404 Not Found - GET https://registry.npmjs.org/@endo%2fenv-options - Not found
npm ERR! 404 
npm ERR! 404  '@endo/env-options@^0.1.0' is not in the npm registry.
npm ERR! 404 You should bug the author to publish it (or use the name yourself!)
npm ERR! 404 It was specified as a dependency of 'ses'
npm ERR! 404 
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.
```

This prevents me from merging. What should I do about this?

gibson042:
I am able to reproduce the issue locally with Node.js v14. Basically, the script tries to `npm pack` the `ses` package and then install the resulting tarball, which works on master (where the `ses` package has no dependencies) but not on this PR (where the `ses` package has a dependency on `@endo/env-options@^0.1.0`, which does not exist in the npm registry because this PR just created it. It would probably be fixed by introducing the package in one PR and updating the `ses` package to use it in a second PR, but would recur with every relevant change to the `env-options` package. In some ways, the `ses-integration-test` package seems like an indirect and incomplete test that the `ses` package has no dependencies, which impacts efforts such as this to share robust code across `ses` and other Endo packages.

I think we should keep the `ses` package free of dependencies, and either tolerate duplication of env-options code (cf. `isTypedArray` in both [`ses`](https://github.com/endojs/endo/blob/0ab46f79216e9802136e8bc4365e402dad1d796b/packages/ses/src/make-hardener.js#L70-L79) and [`pass-style`](https://github.com/endojs/endo/blob/0ab46f79216e9802136e8bc4365e402dad1d796b/packages/pass-style/src/passStyle-helpers.js#L35-L44)) or keep the code in `ses` for deep import from other packages (as it seems may have [previously been the case for `ses/lockdown`](https://github.com/endojs/endo/blob/master/packages/ses/NEWS.md#0130-2021-06-01)?).

packages/env-options/src/env-options.js
erights: Done

kriskowal:
I think in this case, the test might be holding us back. @michaelfig Do you recall what kind of regression the SES integration test is protecting us from, since it appears to test the published sources and not what is in the working copy? Is there a possibility that the test is misbehaving only because it uses `npm` instead of `yarn` (Was this intended to verify that `npm` continued to work even tho we’re using `yarn`?)

michaelfig:
> I think in this case, the test might be holding us back.

I think in this case the test is revealing that the current PR is making a breaking change that may affect our downstream users.  We should not take that lightly.

> @michaelfig Do you recall what kind of regression the SES integration test is protecting us from,

From `packages/ses-integration-test/README.md`:
> This test folder creates a tarball of the main repository and runs the main tests in a browser context with multiple bundlers and tools. The goal is to ensure that someone downstream of the package will not have an issue when importing it.

> since it appears to test the published sources and not what is in the working copy?

It certainly is testing what is in the working copy.  I agree with @gibson042's assessment: it's only attempting to download an unpublished package because `ses/package.json` takes a hard dependency on env-options.  Besides Richard's suggested workarounds we could do one of:

- maybe make the `env-options` dep into a peerDependency
- `npm pack` the env-options package during the ses-integration-test and use it just like we use the `ses` pack

> Is there a possibility that the test is misbehaving only because it uses `npm` instead of `yarn` (Was this intended to verify that `npm` continued to work even tho we’re using `yarn`?)

That's a red herring.  Yarn's behaviour does not differ in this regard.

erights:
I'm going with 

>  It would probably be fixed by introducing the package in one PR and updating the ses package to use it in a second PR, but

at least as far as this first PR, where I have reverted all changes to the `ses` package. Will open a second PR with only these postponed changes where we can discuss what to do with them.

Issue #1582, opened 2023-05-09
wish: single-file version of makeMarshal


dckc:
I have re-implemented makeMarshal badly several times now due to friction around npm packages and modules and stuff vs just copy-and-paste of functions.

as long as using makeMarshal is harder than copy-and-paste-from-stack-overflow, we should expect folks to reimplement it. badly

could be as simple as running a bundler that supports tree shaking on an entrypoint consisting of  `makeMarshal()` .
A UMD file would probably suffice.

dckc:
I tried the bundler approach and ran into lack of support for `0n` literals and such.

In discussion with @erights , I'm learning that perhaps what I want is a specification of marshal consisting of something other than the existing code... something that facilitates high quality re-implementation.

I might expand the scope of this to "marshal specification for interop" or the like. A new issue might be better, but as long as I'm the only one using this issue, I might economize.

kriskowal:
> I tried the bundler approach and ran into lack of support for `0n` literals and such.

Which bundler? If it’s one of ours, that’s a bug.

dckc:
[inter-protocol/test/liq-bid-sheet](https://github.com/Agoric/agoric-sdk/tree/8c53193/packages/inter-protocol/test/liq-bid-sheet) is some code I kludged together to read the state of Inter Protocol contracts in a Google sheet.

[boardSlottingMarshaller](https://github.com/Agoric/agoric-sdk/blob/8c53193/packages/inter-protocol/test/liq-bid-sheet/board-utils.js#L21) is the 3rd or 4th "poor man's marshal" I have written because of friction around `import '@endo/marshal'` vs. copy-and-paste-from-stack-overflow.

This issue was originally a wish for a single script to avoid module issues (Apps Script is not along in having friction, if not complete lack of support, for modules). But after experimenting with `esbuild` to bundle `makeMarshal` I discovered that the 
[V8 Runtime in  Apps Script](https://developers.google.com/apps-script/guides/v8-runtime) doesn't support bigint literals... among the [esbuild JS targets](https://esbuild.github.io/content-types/#javascript) it seems to correspond to something around `es2017`.

I'm narrowing this issue to be about that specific use case and making a new issue for marshal specification for interop.

I think it's likely that this one will get closed as "out of scope" / "wontfix".

dckc:
## webpack and google-apps-script-starter: no joy

I tried using

[...]

PR #1589, opened 2023-05-20
fix(marshal): Avoid bigint literals for Google Apps Script compatibility

gibson042:
Ref #1582

This doesn't fix the redefined issue, but hopefully does remove the incompatibility that originally prompted it.

### DIFF
--- a/packages/marshal/src/encodeToSmallcaps.js
+++ b/packages/marshal/src/encodeToSmallcaps.js
@@ -208,7 +208,7 @@ export const makeEncodeToSmallcaps = (encodeOptions = {}) => {
       }
       case 'bigint': {
         const str = String(passable);
-        return /** @type {bigint} */ (passable) < 0n ? str : `+${str}`;
+        return str.startsWith('-') ? str : `+${str}`;
       }
       case 'symbol': {
         assertPassableSymbol(passable);


dckc (commented):
Would this change be visible in XS heap snapshots? I hesitate to approve until we're quite clear on how we're dealing with endo/agoric-sdk compatibility going forward.

mhofman:
> Would this change be visible in XS heap snapshots? I hesitate to approve until we're quite clear on how we're dealing with endo/agoric-sdk compatibility going forward.

Changes in Endo will only be reflected once the vat is upgraded, either the lockdown / supervisor bundle and / or the contract itself. I think it's safe to merge changes in Endo, unless they are dealing with some kind of protocol that isn't backwards compatible (e.g. a new marshal encoding/serialization)

dckc:
> Changes in Endo will only be reflected once the vat is upgraded, either the lockdown / supervisor bundle and / or the contract itself.

Right... but suppose we decide later that we want to make a change in endo... call it B... one that does _not_ impact the supervisor bundle. But we had already merged this change--call it A--which _does_ impact the supervisor bundle. And we want to start new contracts/vats that take advantage of B. Are we forced to start them with a supervisor bundle that includes A? How do we even do that? Suppose it's not automatic, but we forget/neglect to do it... does the vat get launched with the pre-A supervisor bundle? Is it compatible?

I'm not going to stand in the way / request changes. But Google Apps Script compatibility is not a clear product requirement. Until we have changes motivated by a clear product requirement and we have a clear story on how they're going to be deployed, I'm not inclined to approve.

erights (commented):
FWIW the change looks good to me. But I'm not approving to leave @dckc 's concern in place.

Separately, do we have any idea what other impediments there are to using marshal in Google Sheets?

It seems weird to me that there is a version of JS that does have bigints and bigint arithmetic, but does not parse bigint literals. How confident are we that this parsing issue is a limitation of the JS engine and not some earlier tool in their internal toolchain?

PR #1592, opened 2023-05-20
test(eventual-send): Variation of ses-ava test with deep stacks

erights:
Initially just a port of https://github.com/Agoric/agoric-sdk/pull/2582 to endo


erights:
With the addition of test/test-deep-stacks-ses-ava.js and test/test-deep-send-ses-ava.js to https://github.com/endojs/endo/pull/2109 , closing this PR as subsumed by https://github.com/endojs/endo/pull/2109

PR #1594, opened 2023-05-23
fix(marshal): Add a smaller and more efficient passable encoding

### LINKED ISSUE #1349: Change ordered binary encoding to fix bad nesting complexity.
At https://github.com/endojs/endo/pull/1260#discussion_r960369826 @warner explains that the current ordered binary encoding has a bad complexity measure for nested structures, due to the doubling of bracketing escapes with each additional level.

At https://github.com/endojs/endo/pull/1260#discussion_r1002155752 @gibson042 suggests a much better binary encoding without this bad complexity problem. Rather than do this as part of https://github.com/endojs/endo/pull/1260 , I'm filing this bug so we can remember to do it later.

Either way, for any such encoding change we will need to manage the transition, much like we managed the transition from capdata to smallcaps.

gibson042:
Fixes #1349

Open questions:
- [x] [**REJECTED**] Should we push for extra compactness, e.g. `bfalse` → `bf`/`btrue` → `bt` (booleans) and/or `p1:0` → `o` (0n)?
- [x] [**REJECTED**] Should we allow custom remotable/promise/error encoding to use sequences and/or characters that would not otherwise be allowed, such as raw control characters U+0000 through U+001F or raw string escape characters `!`/`_`?
- [x] [**REPLACE ALL C0 CONTROLS**] For that matter, should we replace the full U+0000 through U+001F range in strings, or just a subset? The full range is necessary if U+0020 SPACE is the array element separator, but that could alternatively be U+0009 TAB and escapes could be limited to code points at or before that. This is really a tradeoff around expectations of character frequency inside strings, limiting blowup when embedding encodings in JSON-serialized containers such as current CapData communication, and (to some extent) human comprehensibility. There's not really a performance difference and at most a linear compactness difference, so it's really a question of aesthetics—for example, JSON-serializing the encoding of `{ "metasyntactic variables": ["foo", "bar", "baz"], tab: "\t", null: "\0" }` would change as follows if we used tab-termination rather than space-termination:
  ```diff
  -"~(^^stab snull smetasyntactic!_variables  ^s!* s!! ^sfoo sbar sbaz   "
  +"~(^^stab\tsnull\tsmetasyntactic variables\t\t^s\n9\ts\n0\t^sfoo\tsbar\tsbaz\t\t\t"
  ```

### DIFF
--- a/packages/marshal/src/encodePassable.js
+++ b/packages/marshal/src/encodePassable.js
@@ -561,8 +561,8 @@ export const makeEncodePassable = (encodeOptions = {}) => {
     }
   };
   const encodePassable = xxx
-    ? // A leading "#" indicates the v2 encoding (with escaping in strings rather than arrays).
-      passable => `#${innerEncode(passable)}`
+    ? // A leading "~" indicates the v2 encoding (with escaping in strings rather than arrays).
+      passable => `~${innerEncode(passable)}`
     : innerEncode;
   return harden(encodePassable);
 };
@@ -654,8 +654,8 @@ export const makeDecodePassable = (decodeOptions = {}) => {
     return innerDecode;
   };
   const decodePassable = encoded => {
-    // A leading "#" indicates the v2 encoding (with escaping in strings rather than arrays).
-    if (encoded.startsWith('#')) {
+    // A leading "~" indicates the v2 encoding (with escaping in strings rather than arrays).
+    if (encoded.startsWith('~')) {
       const innerDecode = makeInnerDecode(
         decodeStringWithEscapes,
         decodeArrayWithoutEscapes,
--- a/packages/marshal/src/encodePassable.js
+++ b/packages/marshal/src/encodePassable.js
@@ -272,7 +272,7 @@ const encodeCompactStringSuffix = str =>
  * @type {(encoded: string) => string}
  */
 const decodeCompactStringSuffix = encoded => {
-  return encoded.replace(/([!_])(.|\n)?/g, (esc, prefix, suffix) => {
+  return encoded.replace(/([\0-!_])(.|\n)?/g, (esc, prefix, suffix) => {
     switch (esc) {
       case '!_':
         return ' ';
@@ -284,7 +284,7 @@ const decodeCompactStringSuffix = encoded => {
         return '_';
       default: {
         const ch = /** @type {string} */ (suffix);
-        // The range of valid escapes is [(0x00+0x21)..(0x1F+0x21)], i.e.
+        // The range of valid `!`-escape suffixes is [(0x00+0x21)..(0x1F+0x21)], i.e.
         // [0x21..0x40] (U+0021 EXCLAMATION MARK to U+0040 COMMERCIAL AT).
         (prefix === '!' && suffix !== undefined && ch >= '!' && ch <= '@') ||
           Fail`invalid string escape: ${q(esc)}`;
--- a/packages/marshal/src/encodePassable.js
+++ b/packages/marshal/src/encodePassable.js
@@ -464,6 +464,21 @@ const decodeTagged = (encoded, decodeArray, decodePassable) => {
   return makeTagged(tag, payload);
 };
 
+const assertEncodedRemotable = encoding => {
+  encoding.charAt(0) === 'r' ||
+    Fail`internal: Remotable encoding must start with "r": ${encoding}`;
+};
+
+const assertEncodedPromise = encoding => {
+  encoding.charAt(0) === '?' ||
+    Fail`internal: Promise encoding must start with "?": ${encoding}`;
+};
+
+const assertEncodedError = encoding => {
+  encoding.charAt(0) === '!' ||
+    Fail`internal: Error encoding must start with "!": ${encoding}`;
+};
+
 /**
  * @typedef {object} EncodeOptions
  * @property {(
@@ -500,7 +515,19 @@ export const makeEncodePassable = (encodeOptions = {}) => {
 
   const innerEncode = passable => {
     if (isErrorLike(passable)) {
-      return encodeError(passable, innerEncode);
+      // We pull out this special case to accommodate errors that are not
+      // valid Passables. For example, because they're not frozen.
+      // The special case can only ever apply at the root, and therefore
+      // outside the recursion, since an error could only be deeper in
+      // a passable structure if it were passable.
+      //
+      // We pull out this special case because, for these errors, we're much
+      // more interested in reporting whatever diagnostic information they
+      // carry than we are about reporting problems encountered in reporting
+      // this information.
+      const result = encodeError(passable, innerEncode);
+      assertEncodedError(result);
+      return result;
     }
     const passStyle = passStyleOf(passable);
     switch (passStyle) {
@@ -524,20 +551,17 @@ export const makeEncodePassable = (encodeOptions = {}) => {
       }
       case 'remotable': {
         const result = encodeRemotable(passable, innerEncode);
-        result.charAt(0) === 'r' ||
-          Fail`internal: Remotable encoding must start with "r": ${result}`;
+        assertEncodedRemotable(result);
         return result;
       }
       case 'error': {
         const result = encodeError(passable, innerEncode);
-        result.charAt(0) === '!' ||
-          Fail`internal: Error encoding must start with "!": ${result}`;
+        assertEncodedError(result);
         return result;
       }
       case 'promise': {
         const result = encodePromise(passable, innerEncode);
-        result.charAt(0) === '?' ||
-          Fail`internal: Promise encoding must start with "?": ${result}`;
+        assertEncodedPromise(result);
         return result;
       }
       case 'symbol': {
--- a/packages/marshal/src/encodePassable.js
+++ b/packages/marshal/src/encodePassable.js
@@ -224,16 +224,24 @@ const decodeBigInt = encoded => {
   return n;
 };
 
-// Escape all characters from U+0000 NULL to U+001F INFORMATION SEPARATOR ONE
-// like `!<character offset by 0x21>` to avoid JSON.stringify expansion to
-// `\uHHHH`, and specially escape U+0020 SPACE (the array element terminator)
-// as `!_` and U+0021 EXCLAMATION MARK (the escape prefix) as `!|`.
-// Relative lexicographic ordering is preserved by this mapping of any character
-// at or before `!` in the contiguous range [0x00..0x21] to a respective
-// character in [0x21..0x40, 0x5F, 0x7C] preceded by `!` (which is itself in the
-// replaced range).
-// Similarly, escape `^` as `_@` and `_` as `__` because `^` indicates the start
-// of an encoded array.
+/**
+ * A sparse array for which every present index maps a code point in the ASCII
+ * range to a corresponding escape sequence.
+ *
+ * Escapes all characters from U+0000 NULL to U+001F INFORMATION SEPARATOR ONE
+ * like `!<character offset by 0x21>` to avoid JSON.stringify expansion as
+ * `\uHHHH`, and specially escapes U+0020 SPACE (the array element terminator)
+ * as `!_` and U+0021 EXCLAMATION MARK (the escape prefix) as `!|` (both chosen
+ * for visual approximation).
+ * Relative lexicographic ordering is preserved by this mapping of any character
+ * at or before `!` in the contiguous range [0x00..0x21] to a respective
+ * character in [0x21..0x40, 0x5F, 0x7C] preceded by `!` (which is itself in the
+ * replaced range).
+ * Similarly, escapes `^` as `_@` and `_` as `__` because `^` indicates the
+ * start of an encoded array.
+ *
+ * @type {Array<string>}
+ */
 const stringEscapes = Array(0x22)
   .fill(undefined)
   .map((_, cp) => {
@@ -264,7 +272,9 @@ const decodeStringWithEscapes = encoded => {
         return '_';
       default: {
         const ch = /** @type {string} */ (suffix);
-        (prefix === '!' && ch >= '!' && ch <= '@') ||
+        // The range of valid escapes is [(0x00+0x21)..(0x1F+0x21)], i.e.
+        // [0x21..0x40] (U+0021 EXCLAMATION MARK to U+0040 COMMERCIAL AT).
+        (prefix === '!' && suffix !== undefined && ch >= '!' && ch <= '@') ||
           Fail`invalid string escape: ${q(esc)}`;
         return String.fromCharCode(ch.charCodeAt(0) - 0x21);
       }
@@ -531,6 +541,8 @@ export const makeEncodePassable = (encodeOptions = {}) => {
         return result;
       }
       case 'symbol': {
+        // Strings and symbols share encoding logic.
+        // Encode the name as a string, then replace the `s` prefix.
         const encName = encodeString(nameForPassableSymbol(passable));
         return `y${encName.slice(1)}`;
       }
@@ -620,7 +632,8 @@ export const makeDecodePassable = (decodeOptions = {}) => {
           return decodeError(encoded, innerDecode);
         }
         case 'y': {
-          const name = decodeString(`s${encoded.slice(1)}`);
+          // Strings and symbols share decoding logic.
+          const name = decodeString(encoded);
           return passableSymbolForName(name);
         }
         case '[':
--- a/packages/marshal/src/encodePassable.js
+++ b/packages/marshal/src/encodePassable.js
@@ -24,6 +24,9 @@ const { isArray } = Array;
 const { fromEntries, is } = Object;
 const { ownKeys } = Reflect;
 
+// eslint-disable-next-line no-control-regex
+const rC0 = /[\x00-\x1F]/;
+
 /**
  * Assuming that `record` is a CopyRecord, we have only
  * string-named own properties. `recordNames` returns those name *reverse*
@@ -492,7 +495,7 @@ const makeEncodeRemotable = (unsafeEncodeRemotable, verifyEncoding) => {
   const encodeRemotable = (r, innerEncode) => {
     const encoding = unsafeEncodeRemotable(r, innerEncode);
     (typeof encoding === 'string' && encoding.charAt(0) === 'r') ||
-      Fail`internal: Remotable encoding must start with "r": ${encoding}`;
+      Fail`Remotable encoding must start with "r": ${encoding}`;
     verifyEncoding(encoding, 'Remotable');
     return encoding;
   };
@@ -503,7 +506,7 @@ const makeEncodePromise = (unsafeEncodePromise, verifyEncoding) => {
   const encodePromise = (p, innerEncode) => {
     const encoding = unsafeEncodePromise(p, innerEncode);
     (typeof encoding === 'string' && encoding.charAt(0) === '?') ||
-      Fail`internal: Promise encoding must start with "?": ${encoding}`;
+      Fail`Promise encoding must start with "?": ${encoding}`;
     verifyEncoding(encoding, 'Promise');
     return encoding;
   };
@@ -514,7 +517,7 @@ const makeEncodeError = (unsafeEncodeError, verifyEncoding) => {
   const encodeError = (err, innerEncode) => {
     const encoding = unsafeEncodeError(err, innerEncode);
     (typeof encoding === 'string' && encoding.charAt(0) === '!') ||
-      Fail`internal: Error encoding must start with "!": ${encoding}`;
+      Fail`Error encoding must start with "!": ${encoding}`;
     verifyEncoding(encoding, 'Error');
     return encoding;
   };
@@ -745,13 +748,22 @@ export const makePassableKit = (options = {}) => {
       decodeCompactArray,
       liberalDecoders,
     );
+    /**
+     * @param {string} encoding
+     * @param {string} label
+     * @returns {void}
+     */
     const verifyEncoding = (encoding, label) => {
+      !encoding.match(rC0) ||
+        Fail`${b(
+          label,
+        )} encoding must not contain a C0 control character: ${encoding}`;
       const decoded = decodeCompactArray(`^s ${encoding} s `, liberalDecode);
       (isArray(decoded) &&
         decoded.length === 3 &&
         decoded[0] === '' &&
         decoded[2] === '') ||
-        Fail`internal: ${b(label)} encoding must be embeddable: ${encoding}`;
+        Fail`${b(label)} encoding must be embeddable: ${encoding}`;
     };
     const encodeCompact = makeInnerEncode(
       encodeCompactStringSuffix,
--- a/packages/marshal/src/encodePassable.js
+++ b/packages/marshal/src/encodePassable.js
@@ -250,28 +250,26 @@ stringEscapes['^'.charCodeAt(0)] = '_@';
 stringEscapes['_'.charCodeAt(0)] = '__';
 
 const encodeStringWithEscapes = str =>
-  `s${str.replaceAll(/[\0-!^_]/g, ch => stringEscapes[ch.charCodeAt(0)])}`;
+  `s${str.replace(/[\0-!^_]/g, ch => stringEscapes[ch.charCodeAt(0)])}`;
 const decodeStringWithEscapes = encoded => {
-  return encoded
-    .slice(1)
-    .replaceAll(/([!_])(.|\n)?/g, (esc, prefix, suffix) => {
-      switch (esc) {
-        case '!_':
-          return ' ';
-        case '!|':
-          return '!';
-        case '_@':
-          return '^';
-        case '__':
-          return '_';
-        default: {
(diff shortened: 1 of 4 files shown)


packages/marshal/src/encodePassable.js
```diff
@@ -216,11 +216,133 @@ const decodeBigInt = encoded => {
...
+ *
+ * @param {unknown[]} array
+ * @param {(p: Passable) => string} encodePassable
+ * @returns {string}
+ */
+const encodeArrayWithoutEscapes = (array, encodePassable) => {
+  const chars = ['^'];
+  for (const element of array) {
+    const enc = encodePassable(element);
+    chars.push(enc, ' ');
```
gibson042, line 284: I chose space as the new array element terminator both because it is adjacent to the 0x00 through 0x1F C0 escape characters (and can therefore participate in the same escaping scheme) and because it acts as a visual separator. That said, there is a potential drawback in trailing spaces to close final array elements being important but invisible (e.g., `[["foo", "bar"]]` [an array containing a single element that is a `["foo", "bar"]` array] encodes to `^^sfoo␠sbar␠␠` [in which spaces have been made visible through replacement with U+2420 SYMBOL FOR SPACE "␠"]).

packages/marshal/src/encodePassable.js
```diff
@@ -328,11 +451,17 @@ export const makeEncodePassable = (encodeOptions = {}) => {
     encodeRemotable = (rem, _) => Fail`remotable unexpected: ${rem}`,
     encodePromise = (prom, _) => Fail`promise unexpected: ${prom}`,
     encodeError = (err, _) => Fail`error unexpected: ${err}`,
+    xxx = false,
```
gibson042, line 454: We'll need a name for the smaller encoding and this property opting in to it, and ideally also a name for the current encoding. I don't have a good suggestion at the moment, so :bike::house: away!

packages/marshal/src/encodePassable.js
```diff
@@ -355,40 +484,45 @@ export const makeEncodePassable = (encodeOptions = {}) => {
...
-        const result = encodePromise(passable, encodePassable);
+        const result = encodePromise(passable, innerEncode);
         result.startsWith('?') ||
           Fail`internal: Promise encoding must start with "?": ${result}`;
         return result;
       }
       case 'symbol': {
-        return `y${nameForPassableSymbol(passable)}`;
+        const encName = encodeString(nameForPassableSymbol(passable));
+        return `y${encName.slice(1)}`;
```
gibson042, line 506: This part was the holdup today—I didn't realize that symbol encoding essentially duplicates string encoding but with a different prefix and must therefore also be updated for the non-trivial string encoding.

packages/marshal/src/encodePassable.js
```diff
@@ -355,40 +484,45 @@ export const makeEncodePassable = (encodeOptions = {}) => {
...
       }
       default: {
         throw Fail`a ${q(passStyle)} cannot be used as a collection passable`;
       }
     }
   };
+  const encodePassable = xxx
+    ? // A leading "#" indicates the v2 encoding (with escaping in strings rather than arrays).
+      passable => `#${innerEncode(passable)}`
+    : innerEncode;
```
gibson042, line 525: This leading `#` to indicate the new encoding is analogous to indication of smallcaps in a CapData `body`, but I'm open to other ideas.

packages/marshal/src/encodePassable.js
```diff
@@ -479,11 +633,13 @@ harden(isEncodedRemotable);
...
- * `bigint` is a two character string because each of those characters
- * individually is a valid bigint prefix. `n` for "negative" and `p` for
- * "positive". The ordering of these prefixes is the same as the
- * rankOrdering of their respective PassStyles. This table is imported by
- * rankOrder.js for this purpose.
+ * `bigint` is a two-character string because each of those characters
+ * individually is a valid bigint prefix (`n` for "negative" and `p` for
+ * "positive"), and copyArray is a two-character string because one encoding
+ * prefixes arrays with `[` while the other uses `^` (which is prohibited from
+ * appearing in an encoded string).
```
gibson042, line 640: I didn't want to continue use of `[` to avoid asymmetric escaping inside strings, but considered it important to maintain the relative rank order of copyArrays in both encodings (the only characters between `[` and `^` are `\` and `]`, neither of which are currently used). Alternatively, we could escape all four such that `specials: [^\]` would be encoded as e.g. `sspecials:!_^{^~^|}^` via `[` → `^{`, `\` → `^|`, `]` → `^}`, and `^` → `^~`.

erights:
Hopefully fixes https://github.com/Agoric/agoric-sdk/issues/2589 , yes?

packages/marshal/src/encodePassable.js
kriskowal: Some thoughts:

- `oldAndBusted` vs `newHawtness`
- `tired` vs `wired`

packages/marshal/src/encodePassable.js
kriskowal: I believe syrup is intended to ignore whitespace, so there’s a plausible objection.

packages/marshal/src/encodePassable.js
kriskowal: * `escapism` vs `realism`

packages/marshal/src/encodePassable.js
```diff
@@ -216,11 +216,131 @@ const decodeBigInt = encoded => {
...
+ */
+const decodeArrayWithoutEscapes = (encoded, decodePassable) => {
+  encoded.startsWith('^') || Fail`Encoded array expected: ${encoded}`;
+  const tail = encoded.slice(1);
+  const elements = [];
+  let depth = 0;
+  let nextIndex = 0;
+  let currentElementStart = 0;
+  for (const { 0: ch, index: i } of tail.matchAll(/[\^ ]/g)) {
+    const index = /** @type {number} */ (i);
```
erights, line 300: Just idly curious why `RegExpMatchArray` type declares its `.index` as optional. Your following unchecked cast relies, seeming correctly, on the knowledge that any match will indeed produce this property. Which would seem to make the optionality unnecessary.

packages/marshal/src/encodePassable.js
```diff
@@ -216,11 +216,131 @@ const decodeBigInt = encoded => {
...
+    if (ch === '^') {
+      // This is the start of a nested array.
+      // TODO: Since the syntax of nested arrays must be validated as part of
+      // decoding the outer one, consider decoding them here into a shared cache
+      // rather than discarding information about their contents until the later
+      // decodePassable.
```
erights, line 306: Since the point of this PR is to fix the bad complexity measure of my previous `encodePassable` attempt, I think this TODO is needed. Do I understand correctly that this additional TODO, we still have bad time complexity, and have only fixed bad space complexity?

If this is hard, feel free to file an issue and postpone to another PR. But IIUC, it doesn't look hard, in which case I'd prefer it in this PR.

packages/marshal/src/encodePassable.js
erights: The way it currently is in this PR is fine.

packages/marshal/src/encodePassable.js
erights: assert what initial character it is that you're assuming you're dropping?

erights (changes requested):
Sorry this took so long!

Looking forward to using this! I think I can start making use of this soon.

erights:
Btw @gibson042 , I reviewed it as if it is Ready For Review. Is there a reason it is still Draft?

erights:
Any idea why CI didn't complete?

packages/marshal/src/encodePassable.js
gibson042: It's also true that trailing whitespace must already be preserved, it just isn't _added_ (e.g., the string consisting only of U+0020 SPACE encodes to `s␠`).

gibson042: Other possible array element terminators would be adjacent to the higher range: ``` ] ``` or ``` ` ```, with the additional cost of requiring underscore-prefixed escaping for actual use of _those_ characters.
* If ``` ] ```, ``` "/[^`_]/" ``` encodes to e.g. ``` s/[_@`___// ``` (`^`, `_`, and `]` being underscore-escaped, the latter to `_/`) and ``` [["foo", "/[^`_]/", "bar"]] ``` encodes to ``` ^^sfoo]s/[_@`___//]sbar]] ```. A bit asymmetric, but workable.
* If ``` ` ```, ``` "/[^`_]/" ``` encodes to e.g. ``` s/[_@_~__]/ ``` (`^`, `_`, and ``` ` ``` being underscore-escaped, the latter to `_~`) and ``` [["foo", "/[^`_]/", "bar"]] ``` encodes to ``` ^^sfoo`s/[_@_~__]/`sbar`` ```.

However, limiting relevant trailing whitespace to standalone strings isn't necessarily good; it might be difficult to notice and potentially possible to exploit an incorrect replacement of "foo " with "foo".

packages/marshal/src/encodePassable.js
gibson042: I agree, but `lint` doesn't:
```sh
marshal$ yarn lint
yarn run v1.22.19
$ yarn lint:types && yarn lint:eslint
$ tsc -p jsconfig.json
src/encodePassable.js:330:31 - error TS18048: 'index' is possibly 'undefined'.

330         currentElementStart = index + 1;
                                  ~~~~~

src/encodePassable.js:334:17 - error TS18048: 'index' is possibly 'undefined'.

334     nextIndex = index + 1;
```

Even the latest version of TypeScript seems to complain, subject to  `strictNullChecks`: [TS playground](https://www.typescriptlang.org/play?strictNullChecks=true&target=99&ts=5.2.2&filetype=js&useJavaScript=truee=4#code/PQKhAIAEBcE8AcCm4DeBnaAnAlgOwOYC+4IwAUAMYD2uG4iu1AJok+ALzgBEAZlVVwDcZMth7gAFAEIGzVgDoMAQ0zQ0AdWzQAFhIDkAPT0BKY6nA7MVAO7gABgFFGVFmxWYlsegA8kFaKwAXOAAJCiyLqyEdoLghJQ0dNBK2AA2HPTOroqp2BSIEgCMxsLUtND0qYgAtgxqGQDaALrCVRUs8DoZAAytiBW4iN7QAJK4LN49fRUUAK6YmHUOVbW40ADKyapTZHyYkmV0KODdwRTaADTgeBNx4FTiyWny1UrQ5wCCqakSwFTA+FMqDI4Gu4yG4AA1OBCsJCEA)

packages/marshal/src/encodePassable.js
gibson042: I'd like to keep that separate from the encoding changes, because it's a transparent optimization on top of them. But you're right that the difficulty is low, so I can keep a branch and PR on top of this one.

packages/marshal/src/encodePassable.js
gibson042: Should we pick a different indication? It may be worth observing that a CapData smallcaps body always follows the initial "#" with JSON text, while this marshal encoding always follows it with a character that cannot start JSON text (although the old encoding uses "[" for arrays, which would require further lookahead if we wanted that level of compatibility).

packages/marshal/src/encodePassable.js
gibson042: For what it's worth, I don't think there's a fundamental aspect to the new encoding other than translation of rank ordering into lexicographic ordering (a feature which it shares with the current encoding). And the key detail differentiating the two is that the current encoding applies escaping to array elements (causing blowup for deep structures) while the new one applies escaping to strings and transitively to symbols, which are the only types that can contain characters confusable with array encoding delimiters and as scalars are immune to the blowup issue.

"Compact" is accurate, but keep in mind that we're still using text rather than raw octets.

gibson042:
> Hopefully fixes [Agoric/agoric-sdk#2589](https://github.com/Agoric/agoric-sdk/issues/2589) , yes?

Yes, I think we can make that claim.

> Btw @gibson042 , I reviewed it as if it is Ready For Review. Is there a reason it is still Draft?

I consider it a draft until we settle on a name.

> Any idea why CI didn't complete?

Pushed to kick CI, and I can also rebase on master if that won't disrupt review.

erights:
> and I can also rebase on master if that won't disrupt review

Please do rebase to master, thanks.

packages/marshal/src/encodePassable.js
erights: > It's also true that trailing whitespace must already be preserved

I'm satisfied with this answer. Let's leave this as you have it. Resolving.

packages/marshal/src/encodePassable.js
erights: So does this support naming these `legacyOrdered` vs `compactOrdered` ? If not, then I don't think I got the thrust of your point.

packages/marshal/src/encodePassable.js
erights: > Should we pick a different indication?

Yes, I think so. There are plenty of ascii candidate characters left ;)

packages/marshal/src/encodePassable.js
```diff
@@ -420,52 +554,72 @@ export const makeDecodePassable = (decodeOptions = {}) => {
...
+        }
+        default: {
+          throw Fail`invalid database key: ${encoded}`;
+        }
       }
+    };
+    return innerDecode;
+  };
+  const decodePassable = encoded => {
+    // A leading "#" indicates the v2 encoding (with escaping in strings rather than arrays).
```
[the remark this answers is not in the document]
erights: Not in this PR, but I would like to be able to choose this encoding as another alternative to smallcaps within marshal, which is why I'd like to make it trivially distinguishable from both smallcaps and its predecessor. When I pick https://github.com/endojs/endo/pull/1423 back up, I hope to try.

packages/marshal/src/encodePassable.js
erights: Since there doesn't seem to be any controversy or open questions here, resolving.

packages/marshal/src/encodePassable.js
erights: In that case, no change suggested. Resolving.

packages/marshal/src/encodePassable.js
erights: Later PR is fine. File an issue so we're sure to get around to it?

packages/marshal/src/encodePassable.js
erights: Looking at this again, I would not add the assertion or comment that I asked for. Fine as is. Resolving.

packages/marshal/src/encodePassable.js
```diff
@@ -216,11 +216,131 @@ const decodeBigInt = encoded => {
...
+      case ' ':
+        return '!_';
+      case '!':
+        return '!|';
+      default:
+        return `!${String.fromCharCode(cp + 0x21)}`;
+    }
+  });
+stringEscapes['^'.charCodeAt(0)] = '_@';
+stringEscapes['_'.charCodeAt(0)] = '__';
```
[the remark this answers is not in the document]
erights: I see this is done. Resolving.

packages/marshal/src/encodePassable.js
```diff
@@ -216,11 +216,131 @@ const decodeBigInt = encoded => {
...
+        return ' ';
+      case '!|':
+        return '!';
+      case '_@':
+        return '^';
+      case '__':
+        return '_';
+      default: {
+        const ch = /** @type {string} */ (suffix);
+        (prefix === '!' && ch >= '!' && ch <= '@') ||
```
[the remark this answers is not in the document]
erights: I see this is done. Resolving.

packages/marshal/src/encodePassable.js
gibson042: I was documenting my thoughts in case that sparked something. "legacy" vs. "compact" seems fine, and we can cross the "moreCompact" bridge if and when we come to it. I do like the implicit value judgment that wouldn't be present in an alternative pair like e.g. "recursiveEscaping" vs. "scalarEscaping".

packages/marshal/src/encodePassable.js
gibson042: OK, looking at https://github.com/endojs/endo/blob/7f1f3319acbd06fa28c08aa12f276ce3532ef6f0/packages/marshal/src/encodePassable.js#L494-L508 and also keeping in mind https://github.com/endojs/endo/blob/7f1f3319acbd06fa28c08aa12f276ce3532ef6f0/packages/marshal/src/encodeToSmallcaps.js#L47-L59 to avoid gratuitous divergence, how would you feel about prefixing with `~`?

packages/marshal/src/encodePassable.js
erights: So `legacyOrdered` vs `compactOrdered` is fine?

packages/marshal/src/encodePassable.js
erights: > how would you feel about prefixing with `~` ?

Good.

packages/marshal/src/encodePassable.js
gibson042: #1747

erights: Thx. Resolving.

packages/marshal/src/encodePassable.js
gibson042: I would think just `"legacy"` vs. `"compact"`, e.g. `makeEncodePassable({ format: "compact" })`. What significance of "ordered" am I missing?

erights: Neither smallcaps nor its predecessor are order preserving. The main feature of both the formats we're naming here is that they are.

gibson042: Ah, so IIUC you're thinking about something like `makeMarshal(convertValToSlot, convertSlotToVal, { serializeBodyFormat: "compactOrdered" })` and wanting "compactOrdered" to be used everywhere. :+1:

erights: yes, exactly.

gibson042:
Updated to use "legacyOrdered" and "compactOrdered" names—so this PR is no longer a draft!

erights (approved):
LGTM, thanks!

(If possible, would be nice to see this merged before tomorrow's ocapn mtg)

gibson042:
Returning to draft because I found that the new format needs more handling for functions provided to `makeEncodePassable`:
```js
import "@endo/init";
import { Far } from "@endo/far";
import { makeMarshal, makeEncodePassable } from "@endo/marshal";
const r0 = Far("Foo");
const data = harden([r0, r0, "."]);
const safeEncodeRemotable = r => "r0";
const unsafeEncodeRemotable = r => "r 0";
console.log({
  safeEncoding: makeEncodePassable({ format: "compactOrdered", encodeRemotable: safeEncodeRemotable })(data),
  unsafeEncoding: makeEncodePassable({ format: "compactOrdered", encodeRemotable: unsafeEncodeRemotable })(data),
});
/* =>
{
  safeEncoding: '~^r0 r0 s. ',
  unsafeEncoding: '~^r 0 r 0 s. '
}
*/
```

gibson042:
> Returning to draft because I found that the new format needs more handling for output from functions provided to `makeEncodePassable`

Resolved now; @erights please review the updates at https://github.com/endojs/endo/pull/1594/commits/9c9d1e3baf20fdb667156bf6d98c43d68433c6c4 .

packages/marshal/src/encodePassable.js
```diff
@@ -346,7 +567,7 @@ export const makeEncodePassable = (encodeOptions = {}) => {
         return encodeBinary64(passable);
       }
       case 'string': {
-        return `s${passable}`;
+        return `s${encodeStringSuffix(passable)}`;
       }
       case 'boolean': {
         return `b${passable}`;
```
gibson042, line 573: While we're making changes, should we also replace "btrue" and "bfalse" with e.g. "bt" and "bf"?

packages/marshal/src/encodePassable.js
```diff
@@ -325,14 +509,51 @@ const decodeTagged = (encoded, decodePassable) => {
...
+  }
+  const encodeRemotable = makeEncodeRemotable(
+    unsafeEncodeRemotable,
+    encodeStringSuffix,
+  );
+  const encodePromise = makeEncodePromise(
+    unsafeEncodePromise,
+    encodeStringSuffix,
+  );
+  const encodeError = makeEncodeError(unsafeEncodeError, encodeStringSuffix);
```
gibson042, line 542: This double-escapes the contents of any embedded string (such as for the error message), but I don't see a good alternative. In particular, asserting that `unsafeEncodeError` produced an independently valid encoding that needs no further escaping would be awkward because this code doesn't have access to any caller-provided decode functions for embedded remotables/promises/errors.

erights:
Could we provide it such a decode function?

gibson042:
Yes, with a refactoring to be more like `makeMarshal`.

erights:
> Yes, with a refactoring to be more like `makeMarshal`.

This seems attractive. Is there any reason not to?

erights:
ping?

gibson042:
@erights I've avoided the double-escaping by instead requiring caller-provided encode{Remotable,Promise,Error} functions to return valid encodings, which is trivial for the legacy format (because escaping is not applied until a value is wrapped in an array) but not for the compact format, and introduced `makePassableKit` to hold the shared encode/decode functionality for such verification (reimplementing `makeEncodePassable` and `makeDecodePassable` to rely upon it). This means a heavier reliance upon the callback for encoding values with substructure (such as errors), but on balance seems like an improvement to me.

I've also documented the remaining interesting questions above:
> - [ ] Should we push for extra compactness, e.g. `bfalse` → `bf`/`btrue` → `bt` (booleans) and/or `p1:0` → `o` (0n)?
> - [ ] Should we allow custom remotable/promise/error encoding to use sequences and/or characters that would not otherwise be allowed, such as raw control characters U+0000 through U+001F or raw string escape characters `!`/`_`?
> - [ ] For that matter, should we replace the full U+0000 through U+001F range in strings, or just a subset? The full range is necessary if U+0020 SPACE is the array element separator, but that could alternatively be U+0009 TAB and escapes could be limited to code points at or before that. This is really a tradeoff around expectations…

gibson042:
My take:
* Extra compactness of booleans and 0n probably doesn't matter, so basically "don't care" and status quo would stand by default.
* To what extent custom remotable/promise/error encodings should be allowed to color outside the lines is interesting. I consider the `!`- and `_`-prefixed escapes to be details of the built-in string encoding and therefore not mandatory (although convenient to adopt), but would like to reject any use of U+0000 through U+001F (U+0020 SPACE and 	U+005E CIRCUMFLEX ACCENT `^` are already effectively disallowed by inescapable interpretation as array syntax)... there's just too much risk of being a vector to produce e.g. misleading console output to justify the small benefit of naïveté in encode functions.
* I predict that element terminators will be more common than space characters inside strings, and also that use of JSON will persist (among other reasons, for its performance benefits). Therefore I like the use of U+0020 SPACE as the element terminator, despite the loss of readability inside encoded strings (replacing spaces with `!_`) and the loss of compactness in a non-JSON binary container or with especially space-heavy strings. That is, I _don't_ think we should pursue this kind of change:
  ```diff
  -"~(^^stab snull smetasyntactic!_variables  ^s!* s!! ^sfoo sbar sbaz   "
  +"~(^^stab\tsnull\tsmetasyntactic variables\t\t^s\n9\ts\n0\t^sfoo\tsbar\tsbaz\t\t\t"
  ```

erights (approved):
I am happy with all these decisions. Please proceed, thanks!

packages/marshal/src/encodePassable.js
warner: `b1` and `b0`?

warner:
Any chance we could get a neighboring `.md` document that explains the format?

This document would also be a good place to mention that the `format:` option limits the decoder's expectations.. if I read this correctly, there's no option to build a decoder that can handle both new- and old- format data. To use this in liveslots, we'll either need:

* 1: Set a vat-wide flag to say what format we're using for all virtual collection keys. If we're launching a brand new vat, we can set it to use the new format forever. If the new-format-capable liveslots is loaded in a pre-existing vat (ie vat upgrade), and it doesn't see this flag, it must stick with the old format forever.
* or 2: Introduce some sort of prefix in the vatstore records to indicate whether each record is new- or old- format. The current vatstore key is like `vc.5.sfoo` (where `vc` means "virtual collection", `5` means collection number 5, and `sfoo` is the passable-encoded string key `foo`). If there's a format key (`s/n/p/b/r`/etc) that's guaranteed to not appear in the old format, we could use that as a per-record indicator, but it would break sort order between old- and new- format keys of the same category.  So that doesn't leave us much/any room for a per-record format indicator, but maybe we could do a per-collection indicator, and at least allow collections created after the upgrade to use the new format.

mhofman:
> * Introduce some sort of prefix in the vatstore records to indicate whether each record is new- or old- format.

I expect this to be per collection to preserve sort order between entries, so it could be in the vc metadata.

PR #1599, opened 2023-05-30
docs(daemon): Preliminary design

kriskowal:
This is a preliminary design for the Endo “Pet Daemon”. My hope is to settle whether we would like to pursue an architecture where “pet names refer to renewable resources” between restarts.


packages/daemon/DESIGN.md
```diff
@@ -0,0 +1,90 @@
+
+Endo is a user agent for running guest programs.
+These programs receive capabilities from the user in the form of references to
+potentially powerful remote objects.
+To that end, Endo exhibits a system called Eventual Send that stretches
```
erights, line 5: "exhibits" is an odd choice. I might have said "provide". But no change suggested.

packages/daemon/DESIGN.md
```diff
@@ -0,0 +1,90 @@
...
+references to remote objects over one or more capability transfer protocols
+(CapTP).
+
+A Promise represents a result from either the past or the future, and Eventual
```
erights, line 9:
```suggestion
A Promise represents a result from either the past or the future that is either local or remote, and Eventual
```
Or better, find some opportunity to say "space and time" ;)

packages/daemon/DESIGN.md
```diff
@@ -0,0 +1,90 @@
...
+Send enables programs to interact with those eventual results, even if they are
+in another worker or on the far side of a network.
+Furthermore, Eventual Send allows messages to be pipelined to these eventual
+results to overcome any latency that might otherwise accrue if the program
```
erights, line 13: FYI just noticing word choices that jump out to me as odd. "accrue". No change suggested.

erights: Please link to something on promise pipelining, since you don't (and should not) explain it in this document. Some suggestions:

http://erights.org/elib/distrib/pipeline.html
https://capnproto.org/rpc.html
https://en.wikipedia.org/wiki/Futures_and_promises#Promise_pipelining
https://www.proposals.es/proposals/Support%20for%20Distributed%20Promise%20Pipelining

packages/daemon/DESIGN.md
```diff
@@ -0,0 +1,90 @@
...
+waited for each intermediate result before sending a follow-up message.
+
+Much hinges on what happens when a program restarts.
+Like most programs, Endo workers start without any memory from their previous
```
erights, line 17: Do you want to be saying "Endo workers" rather than "Endo vats"? Are these distinct concepts in Endo?

packages/daemon/DESIGN.md
```diff
@@ -0,0 +1,90 @@
...
+run.
+
+Endo applications ultimately receive all their powers from the user.
+They receive a "powerbox" object from the user they can use to petition the
```
erights, line 21: "petition". I would have said "ask" or "request". No change suggested.

erights: The best explanation of "powerbox" is at https://www.hpl.hp.com/techreports/2006/HPL-2006-116.html , which you should both read and link to. Lots of similarities here.

erights: "They receive a ... object from the user"

Even though I think I know what you mean, I'm having a hard time understanding this phrase. Objects are in the computers. Users are not. Users therefore do things through user interfaces. For most purposes we can leave that implicit. But when we're talking about that very user interface by which users get access to objects, we need to be clearer.

packages/daemon/DESIGN.md
```diff
@@ -0,0 +1,90 @@
...
+user for access to additional references including capability-bearing objects,
+including remote objects, including remote objects from other peers.
+If the program requests a reference with a name, Endo records a formula for
+recreating the reference such that any subsequent request with the same
+name will provide the result of applying the same formula without interrogating
+the user.
+
+Likewise, the Endo user holds their own powerbox and can assign pet names to
```
erights, line 29: Ok, I am genuinely confused at this point. The program receives a powerbox from the user, which means that the user in some sense had that powerbox. But this is a distinct powerbox held by that same user?

erights (changes requested):
Stopped when I got genuinely confused. I look forward to understanding!

packages/daemon/DESIGN.md
kriskowal: That depends entirely on whether Endo workers meet your criteria for vats. I’m not clear whether the bar for calling a worker a “vat” is above or below orthogonal persistence, but I do know that “worker” requires no explanation and I prefer “worker” over “vat” if there isn’t a material distinction in the context of Endo. Our developer community will already know what a web worker, cloud worker, edge worker, service worker, and node worker are. Vats are certainly a species of the worker genus.

erights: E did manual persistence, but still had vats.

A vat is internally synchronous, with a stack and a queue, and only asynchronously coupled to other vats.

Before Shared Array Buffers, JS agents were vats. Hardened JS Agents, having (mostly) prohibited Shared Array Buffers, are vats.

Regardless, "worker" is fine. Glad to understand the distinction. Resolving.

kriskowal: I’ll make a note that Endo workers qualify as vats. Thank you!

packages/daemon/DESIGN.md
kriskowal: I’ve posted an update that clarifies how programs request powers, how users review requests, and how both have pet name spaces.

packages/daemon/DESIGN.md
```diff
@@ -0,0 +1,215 @@
...
+These programs receive capabilities from the user in the form of references to
+potentially powerful local or remote objects.
+Endo uses a system called Eventual Send that that stretches references to
+remote objects over one or more capability transfer protocols (CapTP).
+Endo is an example of the Powerbox Pattern[^powerbox].
+
+[^powerbox]: See _The Powerbox Pattern_ from [How Emily Tamed the
+    Caml][powerbox].
+
+[powerbox]: https://www.hpl.hp.com/techreports/2006/HPL-2006-116.html
```
erights, line 12: TIL a better pattern markdown pattern for academic-style citations. Thanks!

packages/daemon/DESIGN.md
```diff
@@ -0,0 +1,215 @@
...
+    restarts, albeit it by replaying every message it recieved since the last
+    snapshot or inception.  The architecture of Endo allows for Orthogonal
+    Persistence to be built on top of a manually persisted worker and a durable
+    reference to external storage for snapshots, message log, and virtual
+    objects.
+
+Endo applications ultimately receive all their data and powers from the Endo
+daemon with the user's consent.
+Endo persists **data** between restarts with a content address store in the
+user's home directory.
```
erights, line 49: We need to beware of a terminology ambiguity that's gonna bite us. In most contexts, when we say "Endo", we mean the endo repository, or the set of packages defined by that repo, or the level of abstraction brought about by those packages. When we talk about an "Endo application" I think we mean an application built on those packages.

Here, when you say "Endo", you seem to mean the same thing you're also calling "the Endo daemon". When you say "Endo applications", you state that you mean something with the special relationship to the Endo daemon that this doc is about to explain. I think this shift in meaning of "Endo application" is probably fine. In E terms, perhaps your "Endo application" is more like a capDesk caplet than like an E application.

But repurposing "Endo" to mean specifically the Endo daemon would be painful. OTOH, "the Endo daemon" is a exactly the kind of mouthful that drives people to synecdoche. Perhaps we need a new term?

packages/daemon/DESIGN.md
```diff
@@ -0,0 +1,215 @@
...
+Endo persists **powers** between restarts by storing formulas in the user's
+home directory.
+A formula describes how Endo should recreate a value, including powerful
+capability bearing objects if it restarts.
+
+When Endo runs a confined program, it provides a Powers object that the
```
erights, line 55: > When Endo runs a confined program

Does "Endo application" as used here === "A confined program that Endo runs" === "A confined program that the Endo daemon runs"?

packages/daemon/DESIGN.md
```diff
@@ -0,0 +1,215 @@
...
+  ```
+
+- The program can use the Powers object to request a durable value
+  and give it a pet name.
+  The Powers will memoize the value and retain the resulting formula.
+  Endo will honor subsequent requests for the same pet name from this Powers
+  object by reconstructing the value from its formula if necessary.
+
+  ```js
+  const file = E(powers).request('a file, 'fileName');
```
erights, line 81:
```suggestion
  const file = E(powers).request('a file', 'fileName');
```

packages/daemon/DESIGN.md
```diff
@@ -0,0 +1,215 @@
...
+refer to formulas with pet names, and have separate pet name spaces.
+Changing a pet name does not alter or delete the underlying formula, but a
+formula that is unreachable from the user's pet names may be garbage collected.
+
+- The program can use the Powers object to request an ephemeral value
+  from the user.
+  The Powers will not retain the resulting formula.
+
+  ```js
+  const file = E(powers).request('a file');
```
erights, line 71: As I read this, I don't know what `'a file'` is an example of, or what a realistic form of this example might be. A bit of such suspense is fine, because you cannot explain everything before everything else. But as I read ahead, this was not resolved within my "what does this really mean?" horizon. What does this mean? When the program makes a request like this, what is it actually asking for? What does the user see?

erights:
```suggestion
  const fileP = E(powers).request('a file');
```
to make it clear that the expression returns a promise?

or possibly
```suggestion
  const file = await E(powers).request('a file');
```
though I prefer the first form over this ;)

packages/daemon/DESIGN.md
erights: I am still confused that the program has a petname namespace that is distinct from the user's petname namespace.

packages/daemon/DESIGN.md
```diff
@@ -0,0 +1,215 @@
...
+  Formula Identifier:
+
+  ```
+  readableSha512/$SHA512
+  ```
+
+  Endo stores the blob in the user's home directory, tentatively in a file with
+  the same name as the formula identifier.
+
+- A worker formula denotes the unique identifier for an Endo worker.
```
erights, line 126: A node or browser worker is ephemeral. For example, a node or browser worker never survives a crash. What about an Endo worker? Can a worker formula be dereferenced to a later incarnation of the "same" worker?

packages/daemon/DESIGN.md
```diff
@@ -0,0 +1,215 @@
...
+  Workers can execute confined and unconfined programs and have a `terminate`
+  method.
+
+  Formula Identifier:
+
+  ```
+  workerUuid/$UUID
+  ```
+
+  Endo stores the worker's last known PID and a combined stdout and stderr log
```
erights, line 137: This seems to indicate that Endo workers do persist, in that it would be non-sensical to talk about a browser worker's or an node worker's "last known PID".

Are Endo workers also portable? Can they be moved between machines? Can the be migrated from one Endo daemon to another? (I am not suggesting that you should try answering these in the text here. But I would like to know.)

erights: Seems strange to designate a worker, rather than designate some object exported by a worker.

packages/daemon/DESIGN.md
```diff
@@ -0,0 +1,215 @@
...
+  in a directory with the same name.
+
+- For an evaluation formula, Endo provides the completion value of a JavaScript
+  program to be run in the identified worker, confined to an Endo Compartment,
+  endowed with the values for the identified formulas, bound to the given
+  names.
+  Endo Compartments are endowed with `E` and `Far` for sending and receiving
+  messages.
```
erights, line 145: Aha! Now I appreciate why you want to make these global rather than imports.

packages/daemon/DESIGN.md
```diff
@@ -0,0 +1,215 @@
...
+
+  Formula Identifier:
+
+  ```
+  value/$UUID
+  ```
+
+  Formula Description:
```
erights, line 153: What is a "Formula Description" as distinct from a "Formula Identifier"? Is this the first time that the phrase "Formula Description" is used? I see above mention of a "JSON description" of something stored in the key value store. This description is obviously JSON. Same thing?

packages/daemon/DESIGN.md
```diff
@@ -0,0 +1,215 @@
...
+    "type": "evaluate",
+    "source": "E(worker).evaluate(E(file).text())",
+    "names": ["worker", "file"],
+    "values": [0, 1],
+    "worker": 0,
+    "slots": [
+      "readableSha512/b0b5c0ffeefacade...",
+      "workerUuid/b0b5-c0ffee-facade...",
+    ]
+  }
```
erights, line 166: Meaning of field explained somewhere? I can guess some, but not all.

packages/daemon/DESIGN.md
```diff
@@ -0,0 +1,215 @@
...
+  ```
+
+- For a connection formula, Endo provides a stream of references to a remote
+  connection bootstrap object.
+  The formula describes the capability transfer protocol, address, and public
+  key of the interlocutor, as well as the bootstrap value for the near
```
erights, line 172: Chains have no public key :(

packages/daemon/DESIGN.md
```diff
@@ -0,0 +1,215 @@
...
+  side of the connection.
+
+- The inbox identifier denotes a pet name store and corresponds to an
+  Inbox object.
+  In a user interface, requests will be identified by the inbox's pet name for
+  the powers object.
```
erights, line 178: Boggle point reached here. Stopping. Will try again once I'm more oriented.

packages/daemon/DESIGN.md
kriskowal: I’ll review this to make sure Endo Daemon is explicit. If I say `endo`, I mean the CLI for the Endo Daemon.

I would like to entertain “Endo caplet” or “Endo extension” since “Endo application” might invite ambiguity with a GUI Application.

I don’t know the distinction between E application and capDesk caplet.

In the reactions, cc @dckc @michaelfig 
- 🎉 means Endo extension
- 🚀 means Endo caplet
😉

packages/daemon/DESIGN.md
kriskowal: The nesting looks like:

E. Daemon > E. Worker > E. Compartment > E. Application

As above, I would like to entertain Extension or Caplet instead of Application.

packages/daemon/DESIGN.md
kriskowal: In the Pet Daemon spike, I’ve just made the PetStore type explicit https://github.com/endojs/endo/pull/1635/files#diff-89066b5795eb4c834b09314c6f49442ccfbfd6ea1fbc91c757cca1a6a29c88a6R117-R122

I’m open to giving this a more generic name like FormulaNameStore or Apothecary since only one instance of it is what you would properly call the user’s “pet name store” or just “pet store”.

The Endo Daemon arranges for every guest program (application, extension, caplet) to have its own “formula name store” so that:
- If the Endo Daemon restarts, the guest program (application, extension, caplet) can reconnect to powers the user had previously granted. The pet store tracks (and retains!) the formula tree needed by the guest.
- Formula garbage collection: the Endo Daemon can collect any formula that’s not reachable started at the user’s pet store, through any of the formulas transitively retained, and for any _pet store formula_, any of the formula’s retained by that pet store.

packages/daemon/DESIGN.md
erights: > I don’t know the distinction between E application and capDesk caplet.

We should talk about this verbally sometime. It is interesting and relevant.

packages/daemon/DESIGN.md
erights: What does "X > Y" mean here? I tried on
   * cardinality: there are multiple Xs per Y
   * taxonomy: every X is also a Y

and neither seemed to fit.

packages/daemon/DESIGN.md
```diff
@@ -0,0 +1,215 @@
...
+
+- The user can create a new Powers object by calling the `makePowers`
+  method of their Inbox.
+  In a user interface, requests will be attributed to their pet name for the
+  Powers object.
+
+Every formula has a string identifier.
+If the behavior of the formula cannot be inferred from the identifier,
+Endo stores a corresponding JSON description in a key value store (tentatively
+just files) indexed on the identifier.
```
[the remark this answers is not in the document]
kriskowal: No, the scheme for formula storage, and even formula identifiers, is invisible to confined programs.

It’s possible that formula identifiers might be invisible to other Endo daemon peers as well.

That is, formulas and formula identifiers are fully encapsulated implementation details at this time, and likely to remain so.

packages/daemon/DESIGN.md
kriskowal: If an Endo daemon restarts, it can reconstruct a value for any formula that is still retained starting from the user’s pet store, including any workers and confined programs, including any of the values they were previously granted and gave a durable name.

packages/daemon/DESIGN.md
kriskowal: Endo daemon retains the PID and logs for debugging purposes. These are not reachable from within a confined program.

packages/daemon/DESIGN.md
kriskowal: These do not need to be global for confined programs, but they do need to be endowed for `eval` formulas since dynamic import would be unwieldy. I should note that eval scripts are very useful for ad hoc orchestration at the command line or a REPL, but are not intended to represent normal usage by end users.

packages/daemon/DESIGN.md
kriskowal: This is from a recent refactor that greatly improved the correctness and legibility of the spike. Some formula identifiers encode enough information about the formula so that it’s not necessary to store any additional information, and having more information in the formula identifier makes it easier to read the descriptions.

For example, a formula identifier was previously just a long number. Now, it’s a tag and a number, like `readable-blob-sha512:cf83e1357eefb8bdf1542850d66d8007d620e4050b5715dc83f4a921d36ce9ce47d0d13c5d85f2b0ff8318d2877eec2f63b931bd47417a81a538327af927da3e`. There is no formula description file for this, but there is a corresponding blob in the content address store. Eval and import formulas have more dependencies, so the JSON file is necessary to encode those. For the formula of a name-to-formula-store, there’s no JSON formula description either.

packages/daemon/DESIGN.md
kriskowal: Added field descriptions.

packages/daemon/DESIGN.md
kriskowal: Generally, “contains”

- Endo Daemon spawns Endo workers
- Endo Workers construct Endo Compartments
- Endo Compartments confine Endo Applications (extensions, caplets, guest programs)

packages/daemon/DESIGN.md
kriskowal: Indeed, a chain connection would necessarily be framed in terms of the sender’s wallet address instead. I’m not sure what the usual practices is for establishing the authenticity of the chain itself, since we usually just use an RPC node to find a sample of RPC nodes. In any case, chain connections would be framed as a kind of formula.

I’ll try to make more clear that this is a family of formula types with varying means of establishing connectivity and authenticity.

PR #1600, opened 2023-05-31
docs(patterns): Limit M.nat() to bigints

### LINKED ISSUE #1597: M.nat() allows safe number?
The docs say yes:

https://github.com/endojs/endo/blob/0ab46f79216e9802136e8bc4365e402dad1d796b/packages/patterns/src/types.js#L305-L308

but the implementation says no:

https://github.com/endojs/endo/blob/0ab46f79216e9802136e8bc4365e402dad1d796b/packages/patterns/src/patterns/patternMatchers.js#L837

the tests don't seem to take a side.

gibson042:
Fixes #1597

Also improves pattern matcher test coverage in general.

### DIFF
--- a/packages/patterns/test/test-patterns.js
+++ b/packages/patterns/test/test-patterns.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-continue */
 import { test } from './prepare-test-env-ava.js';
 // eslint-disable-next-line import/order
 import { makeTagged } from '@endo/marshal';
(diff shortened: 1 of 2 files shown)


packages/patterns/test/test-patterns.js
```diff
@@ -1,27 +1,70 @@
...
+    record: (repr, kind) => `${kind} ${repr} - Must be a copyRecord`,
+    array: (repr, kind) => `${kind} ${repr} - Must be a copyArray`,
+    set: (repr, kind) => `${kind} ${repr} - Must be a copySet`,
+    bag: (repr, kind) => `${kind} ${repr} - Must be a copyBag`,
+    map: (repr, kind) => `${kind} ${repr} - Must be a copyMap`,
+    remotable: (repr, kind) => `${kind} ${repr} - Must be a remotable`,
+    error: (repr, kind) => `${kind} ${repr} - Must be a error`,
+    promise: (repr, kind) => `${kind} ${repr} - Must be a promise`,
+    undefined: (repr, kind) => `${kind} ${repr} - Must be a undefined`,
+    null: repr => `${repr} - Must be: null`,
```
gibson042, line 47: It's interesting to note that `M.undefined()` is implemented as `makeKindMatcher('undefined')` (and therefore `makeTagged('match:kind', 'undefined')`) while `M.null()` is implemented as just `null`, resulting in the need for special-casing negative tests of the latter.

erights: That seems like an adequate list of irregularities that you should probably switch `M.null()` to be more like `M.undefined()`. This PR seems like a fine time to do that.

erights: > because it results in exceptions rather than false results when checking against a non-Key specimen such as a Matcher

The fact that this case acts differently itself seems like a bug. Agree?

packages/patterns/test/test-patterns.js
```diff
@@ -368,46 +518,120 @@ const runTests = (successCase, failCase) => {
...
+          specimen,
+          M[method](),
+          makeMessage('"[mysteryTag]"', 'mysteryTag', 'tagged'),
+        );
+      } else if (method === 'null') {
+        failCase(
+          specimen,
+          M[method](),
+          simpleMethods.key('"[mysteryTag]"', 'mysteryTag', 'tagged'),
+          true,
```
erights, line 540: Oh yeah, this is just too damn ugly. Please reform `M.null()`

erights (approved):
Thanks for the improved tests!

Btw, have you thought about how we could use fast-check for this?

dckc (commented):
This addresses my concerns.

gibson042:
I agree that the `M.null()` divergence feels like a bug (now #1601), but that's a behavioral change so I'd like to fix it in a followup and keep this PR limited to docs and tests.

As for fast-check, property-based testing was in the back of my mind here... a lot of it seems manual, but we could definitely assert agreement such as `kindOf` response values over arbitrary Passables corresponding with the same-named `M` method to the exclusion of all other type-based `M` methods (possibly with predictable error messages, which are in general the biggest hurdle), and also trivial properties like `M.any()` and `M.and()` and `M.or(self, arbitrary)` always matching any Passable.

PR #1603, opened 2023-06-01
fix(ses): ses depends on env-options

erights:
Changes the `ses` package to depend on the `env-options` package, rather than using its own module.

See https://github.com/endojs/endo/pull/1585#issuecomment-1571154487


michaelfig:
> Changes the `ses` package to depend on the `env-options` package, rather than using its own module.

<details><summary>I believe that's disregarding @gibson042's <a href="https://github.com/endojs/endo/pull/1585#issuecomment-1561353429">"I think we should..." suggestions</a>.</summary>

> I think we should keep the `ses` package free of dependencies, and either tolerate duplication of env-options code (cf. `isTypedArray` in both [`ses`](https://github.com/endojs/endo/blob/0ab46f79216e9802136e8bc4365e402dad1d796b/packages/ses/src/make-hardener.js#L70-L79) and [`pass-style`](https://github.com/endojs/endo/blob/0ab46f79216e9802136e8bc4365e402dad1d796b/packages/pass-style/src/passStyle-helpers.js#L35-L44)) or keep the code in `ses` for deep import from other packages (as it seems may have [previously been the case for `ses/lockdown`](https://github.com/endojs/endo/blob/master/packages/ses/NEWS.md#0130-2021-06-01)?).
</details>

<details><summary>If that's the case, I'd strongly recommend against the "It probably would be fixed by..." buried higher up in that comment, as it (correctly) states we will face this same problem every time we haven't yet published the latest `env-options` to NPM.  That will cause the integration test to fail in CI, just as it (currently) is in this PR.</summary>

> It would probably be fixed by introducing the package in one PR and updating the `ses` package to use it in a second PR, but would recur with every relevant change to the `env-options` package.
</details>

<details><summary>What do you think of my <a href="https://github.com/endojs/endo/pull/1585#issuecomment-1570497334">"we could do one of..." suggestions</a>?</summary>

> Besides Richard's suggested workarounds we could do one of:
>
> - maybe make the `env-options` dep into a peerDependency
> - `npm pack` the env-options package during the ses-integration-test and use it just like we use the `ses` pack
</details>

erights:
I do not understand either peerDependency or `npm pack` well enough to have an opinion, or to know how to actually do either of these. Concretely ELI5 what do you advise?

Since I don't know enough to object, I'll also ask: why do you think I might object to one of these or the other? ;)

kriskowal:
I believe a peerDependency might leave ses in a broken state if no peers in
the node_modules tree took a hard dependency on env-options. When running
under Node.js, we don’t use the version if env-options that is embedded in
the ses bundle. But, I don’t understand peerDependencies well either.

michaelfig:
> Concretely ELI5 what do you advise?

I just now pushed an update to this PR's `ses-integration-test` that makes it aware of `ses/package.json`'s `dependencies`.  Running through CI now.

kriskowal:
@michaelfig I like your change since it moves us closer to being able to evolve the `ses` dependency graph without having to worry about breaking this test. I’m also fine being a steward for the institutional knowledge that the test has to be kept in sync with `ses` for the `npm install "$(npm pack)` list, and maybe I can improve upon that in time.

kriskowal (approved):

michaelfig (approved):
LGTM!  (But you probably should take a look yourself, or have another reviewer for the code I wrote outside of `packages/ses` and `packages/env-options`).

PR #1605, opened 2023-06-01
fix(patterns): Implement M.null() and M.undefined() as Key Patterns

### LINKED ISSUE #1601: patterns: `M.null()` diverges from `M.undefined()` in a way that introduces exceptions
Observed at https://github.com/endojs/endo/pull/1600#discussion_r1212050616
> `M.undefined()` is implemented as `makeKindMatcher('undefined')` (and therefore the Matcher `makeTagged('match:kind', 'undefined')`) while `M.null()` is implemented as just the self-matching non-Matcher Pattern `null`, resulting in the need for special-casing negative tests of the latter (both because the error messages differ and because it results in exceptions rather than `false` results when checking against a non-Key specimen such as a Matcher).

`M.null()` should instead be brought into alignment with the other type-based Matchers, presumably as `makeKindMatcher('null')`.

gibson042:
Fixes #1601

stacked on #1600: https://github.com/endojs/endo/compare/gibson-1597-nat-matcher-requires-bigint...gibson-1601-null-undefined-pattern

### DIFF
--- a/packages/patterns/src/patterns/patternMatchers.js
+++ b/packages/patterns/src/patterns/patternMatchers.js
@@ -270,7 +270,7 @@ const makePatternKit = () => {
    * @returns {boolean}
    */
   const checkAsKeyPatt = (specimen, keyAsPattern, check) => {
-    if (keyEQ(specimen, keyAsPattern)) {
+    if (isKey(specimen) && keyEQ(specimen, keyAsPattern)) {
       return true;
     }
     return (
--- a/packages/patterns/src/patterns/patternMatchers.js
+++ b/packages/patterns/src/patterns/patternMatchers.js
@@ -167,6 +167,12 @@ const makePatternKit = () => {
    * that the store level associates with that kind.
    */
 
+  /** @type {Map<Kind, unknown>} */
+  const singletonKinds = new Map([
+    ['null', null],
+    ['undefined', undefined],
+  ]);
+
   /**
    * @type {WeakMap<CopyTagged, Kind>}
    * Only for tagged records of recognized kinds whose store-level invariants
@@ -251,6 +257,12 @@ const makePatternKit = () => {
    * @returns {boolean}
    */
   const checkKind = (specimen, kind, check) => {
+    // check null and undefined as Keys
+    if (singletonKinds.has(kind)) {
+      // eslint-disable-next-line no-use-before-define
+      return checkAsKeyPatt(specimen, singletonKinds.get(kind), check);
+    }
+
     const realKind = kindOf(specimen, check);
     if (kind === realKind) {
       return true;
--- a/packages/patterns/test/test-patterns.js
+++ b/packages/patterns/test/test-patterns.js
@@ -43,9 +43,11 @@ const runTests = (successCase, failCase) => {
     remotable: (repr, kind) => `${kind} ${repr} - Must be a remotable`,
     error: (repr, kind) => `${kind} ${repr} - Must be a error`,
     promise: (repr, kind) => `${kind} ${repr} - Must be a promise`,
-    undefined: (repr, kind) => `${kind} ${repr} - Must be a undefined`,
+    // M.undefined() and M.null() match as exact Keys rather than kinds.
+    undefined: repr => `${repr} - Must be: "[undefined]"`,
     null: repr => `${repr} - Must be: null`,
   };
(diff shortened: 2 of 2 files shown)


erights:
Since I'm sole reviewer, I took the liberty of changing the PR so I can see the diffs from #1600

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -249,6 +249,14 @@ const makePatternKit = () => {
    * @returns {boolean}
    */
   const checkKind = (specimen, kind, check) => {
+    // check null and undefined as Keys
+    const key =
+      // eslint-disable-next-line no-nested-ternary
+      kind === 'null' ? null : kind === 'undefined' ? undefined : kind;
+    if (key == null) {
```
erights, line 256: With `@jessie-check` I get

image

as I should. Also, my subjective experience is that I always misread this first as `=== null` and think "What about `undefined`?" Only then do I backtrack and look harder. This has happened to me already on three different occasions just looking at this line of code in this PR. Twice on github. Once in vscode.

We're not about to turn `@jessie-check` on for this file anytime soon because of other violations. But we should not gratuitously violate Jessie rules either. Nor should we make Jessie rules more complicated by allowing this weird exception.

erights: cc @mhofman who I know has a contrary view and may want to contribute to this review conversation.

@mhofman ,
I do understand that it is a widespread JS convention. And if one gets used to it (as I have never been able to do), that it is just as safe. But: From the rest of our code, my eyes are trained under the *assumption* that there are no `==` in our code, which is why I always find myself doing this double take.

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -268,7 +276,7 @@ const makePatternKit = () => {
    * @returns {boolean}
    */
   const checkAsKeyPatt = (specimen, keyAsPattern, check) => {
-    if (keyEQ(specimen, keyAsPattern)) {
+    if (isKey(specimen) && keyEQ(specimen, keyAsPattern)) {
```
erights, line 279: Good!

packages/patterns/test/test-patterns.js
```diff
@@ -525,20 +527,13 @@ const runTests = (successCase, failCase) => {
...
           M[method](),
           makeMessage('"[mysteryTag]"', 'mysteryTag', 'tagged'),
         );
-      } else if (method === 'null') {
-        failCase(
-          specimen,
-          M[method](),
-          simpleMethods.key('"[mysteryTag]"', 'mysteryTag', 'tagged'),
-          true,
-        );
```
erights, line 541: Huh. I'd forgotten that I had already made this special case. Nice to see that the rest of this PR causes this to melt away.

Wow, in four places. Nice!

erights (changes requested):
LGTM except for the gratuitous Jessie violation of `== null`. Comments there.

packages/patterns/src/patterns/patternMatchers.js
mhofman: Well you could always bypass it in this case by changing the check to
```suggestion
    if (key !== kind) {
```

But in my experience, `== null` is well understood by most JS programmer.

erights: The view from Jessie gives an interesting counterpoint: Jessie is simple enough to be teachable to non-JS programmers. For them, saying "There is no `==` in Jessie" is a lot simpler than making the special case.

erights (approved):
I'm approving to clear my objection, because I don't want to block this PR on this minor controversy. But I would still appreciate it if this were changed to be within Jessie. Thanks.

packages/patterns/src/patterns/patternMatchers.js
gibson042: Not that it would affect things here, but I'd like to avoid **all** comparisons of a variable to itself because `name === name` and `name !== name` give counterintuitive results when the value is `NaN` (and to that effect, I just opened #1631).

However, I did replace this check in https://github.com/endojs/endo/pull/1605/commits/6a77c8873b25e2f5262369aa6453245ba4c03a36 with use of a `singletonKinds` Map\<Kind, unknown> containing only `'null' => null, 'undefined' => undefined`.

PR #1606, opened 2023-06-02
feat(ses): Add assert.raw for embedding unquoted strings in details

gibson042:
When working on #1600 and #1605, I noticed patterns like
```js
// quoting without quotes
const details = X([`${realKind} `, ` - Must be a ${kind}`], specimen);
```

Rather than manually manipulating the input of a function intended for use in tagged templates (and mixing up the order in which parts appear), it seems more clear to have a function that is like `quote` but does not add quotes (replacing the above with ```const details = X`${r(realKind)} ${specimen} - Must be a ${r(kind)}`;```). Also tagging @FUDCo, who IIRC has asked for this in the past.

### DIFF
--- a/packages/ses/src/error/assert.js
+++ b/packages/ses/src/error/assert.js
@@ -24,6 +24,7 @@ import {
   globalThis,
   is,
   isError,
+  regexpTest,
   stringIndexOf,
   stringReplace,
   stringSlice,
@@ -55,6 +56,23 @@ const quote = (payload, spaces = undefined) => {
   return result;
 };
 freeze(quote);
+
+// We may relax this later, but for now only space-separated sequences of
+// sufficiently word-like parts may be embedded into details without quoting.
+const canBeRaw = freeze(/^[\w:-]( ?[\w:-])*$/);
+
+/** @type {AssertQuote} */
+const raw = (payload, spaces = undefined) => {
+  if (typeof payload !== 'string' || !regexpTest(canBeRaw, payload)) {
+    return quote(payload, spaces);
+  }
+  const result = freeze({
+    toString: freeze(() => payload),
+  });
+  weakmapSet(declassifiers, result, payload);
+  return result;
+};
+freeze(raw);
 
 // /////////////////////////////////////////////////////////////////////////////
 
@@ -403,13 +421,9 @@ const makeAssert = (optRaise = undefined, unredacted = false) => {
     typeof typename === 'string' || Fail`${quote(typename)} must be a string`;
 
     if (optDetails === undefined) {
-      // Like
-      // ```js
-      // optDetails = details`${specimen} must be ${quote(an(typename))}`;
-      // ```
-      // except it puts the typename into the literal part of the template
-      // so it doesn't get quoted.
-      optDetails = details(['', ` must be ${an(typename)}`], specimen);
+      // Embed the type phrase without quotes.
+      const typeWithDeterminer = an(typename);
+      optDetails = details`${specimen} must be ${raw(typeWithDeterminer)}`;
     }
     fail(optDetails, TypeError);
   };
@@ -431,6 +445,7 @@ const makeAssert = (optRaise = undefined, unredacted = false) => {
     details,
     Fail,
     quote,
+    raw,
     makeAssert,
   });
   return freeze(assert);
--- a/packages/ses/src/error/assert.js
+++ b/packages/ses/src/error/assert.js
@@ -57,13 +57,25 @@ const quote = (payload, spaces = undefined) => {
 };
 freeze(quote);
 
-// We may relax this later, but for now only space-separated sequences of
-// sufficiently word-like parts may be embedded into details without quoting.
-const canBeRaw = freeze(/^[\w:-]( ?[\w:-])*$/);
+const canBeBare = freeze(/^[\w:-]( ?[\w:-])*$/);
 
-/** @type {AssertQuote} */
-const raw = (payload, spaces = undefined) => {
-  if (typeof payload !== 'string' || !regexpTest(canBeRaw, payload)) {
+/**
+ * Embed a string directly into error details without wrapping punctuation.
+ * To avoid injection attacks that exploit quoting confusion, this must NEVER
+ * be used with data that is possibly attacker-controlled.
+ * As a further safeguard, we fall back to quoting any input that is not a
+ * string of sufficiently word-like parts separated by isolated spaces (rather
+ * than throwing an exception, which could hide the original problem for which
+ * explanatory details are being constructed---i.e., ``` assert.details`...` ```
+ * should never be the source of a new exception, nor should an attempt to
+ * render its output, although we _could_ instead decide to handle the latter
+ * by inline replacement similar to that of `bestEffortStringify` for producing
+ * rendered messages like `(an object) was tagged "[Unsafe bare string]"`).
+ *
+ * @type {AssertQuote}
+ */
+const bare = (payload, spaces = undefined) => {
+  if (typeof payload !== 'string' || !regexpTest(canBeBare, payload)) {
     return quote(payload, spaces);
   }
   const result = freeze({
@@ -72,7 +84,7 @@ const raw = (payload, spaces = undefined) => {
   weakmapSet(declassifiers, result, payload);
   return result;
 };
-freeze(raw);
+freeze(bare);
 
 // /////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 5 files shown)


packages/ses/src/error/assert.js
```diff
@@ -56,6 +57,23 @@ const quote = (payload, spaces = undefined) => {
 };
 freeze(quote);
 
+// We may relax this later, but for now only space-separated sequences of
+// sufficiently word-like parts may be embedded into details without quoting.
+const canBeRaw = freeze(/^[\w:-]( ?[\w:-])*$/);
```
FUDCo, line 62: I don't think this is necessary.  If you're deliberately choosing to output the string without quotation marks, I'm happy to presume you know why you're doing that.  IIUC, the reason `q` outputs the quotation marks is because it's often used in contexts where you might not be entirely sure what's going to get spit out, whereas you'd use `r` in contexts where you (supposedly) know what you're doing.  In other words, the idea is to avoid accidentally generating confusing output, not to protect against an adversarial output source.

FUDCo (approved):
I'm happy with this if Mark is, but please do wait for his blessing before landing.

packages/ses/src/error/assert.js
erights: I think I agree with @FUDCo , but let's investigate before dropping a restriction motivated by security.

I know I can state the threat model addressed by the quoting of `quote`, as well as the threat model addressed by the default redaction vs the quoted non-redaction of `quote`. Can anyone state a threat model that the `canBeRaw` test addresses that would not be addressed if `raw` did not do a `canBeRaw` test?

erights (changes requested):
Besides my `canBeRaw` question, I also wonder about the choice of the name `raw`. If it is supposed to suggest some analogy to `String.raw` I don't see it. Why `raw`?

I'm hitting "Request changes" but so far I'm only asking for clarification first. No actual changes yet requested.

packages/ses/src/error/assert.js
erights: FWIW, leaving aside the `canBeRaw` question, I can also state the threat model difference addressed by the distinction between `quote` and `raw`.

erights:
Also wonder about the decision to fall back on `quote` if the payload is not a string (or whatever additional test) vs error if not a string. I think I like the fallback better too, but would like to hear your thoughts.

FUDCo:
I also like the fallback.  The whole reason to omit the quotes is if you are trying to control the format of the output, in which case if you need a non-quoted form of a non-string you can coerce to a string yourself.

FUDCo:
W.r.t. the "raw" question: I often use `p` (for `print`) in my debug code as a shorthand for `console.log` if I'm doing a lot of them.  So maybe `raw` becomes `print` and `r` becomes `p` (`p` just by convention, in the same way that `quote` becomes `q` just by convention)?

erights:
I like that `p` is the mirror image of `q`. Mnemonic!

erights:
> I like that `p` is the mirror image of `q`. Mnemonic!

FWIW, the opposite of "redact" is "disclose". `d` is an upside down `q`.

kriskowal:
> > I like that `p` is the mirror image of `q`. Mnemonic!
> 
> FWIW, the opposite of "redact" is "disclose". `d` is an upside down `q`.

We were using `d` for details and I miss that. We need to find a way to use `b` so we can complete the set.

kriskowal:
Please ignore the following message:

```js
const { q: quote, d: details, p: pretty, b: bare } = assert;
```

packages/ses/src/error/assert.js
gibson042: I added `canBeRaw` to make sure that e.g. future expansion of `kindOf` output to propagate arbitrary tag strings cannot cause rendering error details to produce misleading text.

gibson042:
> Also wonder about the decision to fall back on `quote` if the payload is not a string (or whatever additional test) vs error if not a string. I think I like the fallback better too, but would like to hear your thoughts.

The reason for the fallback to `quote` is pragmatic—this code supports error visibility, and inappropriate construction of details (or a successful attempt to put a details-relevant variable into an unexpected state) should not preclude best-effort rendering of the original details. For example, I'd rather have ``` X`bad input: ${value} of tag type ${raw(tagType)}` ``` render like «_bad input: (an object) of tag type "$#*!"_» than see another error in place of the "bad input" message, similar to how [marshal makes an exceptional allowance for non-hardened Error instances](https://github.com/endojs/endo/blob/7a685f59a8f61c63da9dbec5ef78cec0d1b17d3b/packages/marshal/src/marshal.js#L101-L104).

packages/ses/src/error/assert.js
erights: I don't understand. Can you give an example?

gibson042:
> I also wonder about the choice of the name `raw`. If it is supposed to suggest some analogy to `String.raw` I don't see it. Why `raw`?

No particular reason other than difficulty of naming a function similar to `quote` but that doesn't actually add quotation marks, as a method on an object where methods are generally expected to perform assertions rather than return output (`assert.fail`, `assert.equal`, `assert.typeof`, etc.). Of the alternatives suggested here, my favorite is `bare` from a comment that @kriskowal must have deleted.

erights:
I also like `bare` best. Good you saw @kriskowal 's comment before he deleted it!

packages/ses/src/error/assert.js
gibson042: Sure. [`checkTagged`](https://github.com/endojs/endo/blob/7a685f59a8f61c63da9dbec5ef78cec0d1b17d3b/packages/patterns/src/patterns/patternMatchers.js#L184) currently rejects tagged records in which the tag is not recognized with ``` X`cannot check unrecognized tag ${q(tag)}: ${tagged}` ```, but we might decide in the future to pass through such values, in which case downstream code that relies on the current rejection of unknown tags and e.g. complains about bad input like ``` X`unacceptable tag: ${raw(tag)} must start with "match:"` ``` might instead encounter a tagged record in which the tag contains line feeds or other control characters, in which case we'd probably like to see
```
unacceptable tag: "foo must start with \"match:\"\nim in ur base, killing ur d00ds\nunacceptable tag: bar" must start with "match:"
```
rather than
```
unacceptable tag: foo must start with "match:"
im in ur base, killing ur d00ds
unacceptable tag: bar must start with "match:"
```

erights: This is the attacker-controlled data case, i.e., any string determined by anything outside the current code's reliance set. No such string should ***ever*** be handled by `raw` or `bare`. The lack of quotes leaves the quoting vulnerable to an injection attack regarding human interpretation of the resulting diagnostic string. IOW, all injection attacks are actually quoting confusion attacks. Splicing attacker controlled data into trusted strings should always safely quote the attacker controlled data.

This is actually much of the reasoning that initially led me to bundle quoting with unredaction.

In some systems, they use taint tracking to try to avoid such quoting confusions. Since we don't have that tooling (yet?), a case could still be made that your `canBeRaw` check helps reduce the damage of accidents. That possible argument has me on the fence. If failing the `canBeRaw` check caused an error, we could remove it later. However, since it falls back to quote, for the very good reason you explain, it would be harden to relax the constraint later.

Whether we keep the `canBeRaw` (or `canBeBare`) check or not, we need anyway to document that one should never use `raw`/`bare` on data that is possibly attacker controlled, as the risk of attacker-induced quoting confusion. We should also remember that we would benefit from better tooling anyway, to track attacker controlled data and how it mixes with relied-upon data.

All that said, I still weakly prefer that we remove the `canBeRaw`/`canBeBare` check. But if you strongly prefer to keep it, do so.

That answers all my previous questions. But it did clarify for me that the explanation of `raw` needs to warn about this hazard, so that's my only remaining requested change.

packages/ses/src/error/types.js
```diff
@@ -341,6 +341,7 @@
  *   details: DetailsTag,
  *   Fail: FailTag,
  *   quote: AssertQuote,
+ *   raw: AssertQuote,
```
erights, line 344: My only remaining requested change is to document the need to avoid using `raw`/`bare` on possibly attacker controlled data, lest it give the attacker the opportunity to introduce a quoting confusion in the interpretation of the error string. I leave it to you to determine where this explanation should go.

erights (changes requested):

packages/ses/src/error/types.js
gibson042: Done; please re-review.

erights (approved):
LGTM, thanks!

erights:
See https://github.com/Agoric/agoric-sdk/issues/8005

PR #1612, opened 2023-06-06
docs(ses): Provide guidance/coordination for SES incompatibility tracking

### LINKED ISSUE #1611: Documentation/correction of SES-incompatible code should be coordinated
#576 tracks SES incompatibilities, but can be hard to discover. The [SES README](https://github.com/endojs/endo/blob/master/packages/ses/README.md) should reference it, and possibly also provide a reporting template.

gibson042:
Fixes #1611

### DIFF
--- a/packages/ses/README.md
+++ b/packages/ses/README.md
@@ -749,7 +749,29 @@ Object.defineProperties(<lhs>, {
 ```
 
 Upon encountering an incompatibility, we recommend that you add a comment to
-[issue #576][incompatibility tracking] and file an issue with the external project referencing this section.
+[issue #576][incompatibility tracking] and file an issue with the external
+project referencing this section.
+Projects often have their own unique issue reporting templates, but generally
+provide some place to include text like
+
+> ```
+> This project has some assignments that break in an environment with frozen
+> intrinsic objects, such as
+> [Hardened JS (a.k.a. SES)](https://github.com/endojs/endo/blob/master/packages/ses#ecosystem-compatibility)
+> or Node.js with the
+> [`--frozen-intrinsics`](https://nodejs.org/docs/latest/api/cli.html#--frozen-intrinsics)
+> option.
+> Specifically, [link to source in the project] does not work correctly in such
+> an environment.
+> 
+> Please consider increasing support by replacing assignments to object
+> properties inherited from intrinsics with use of `Object.defineProperties`
+> (thereby working around the JavaScript "override mistake"), and if applicable
+> also by avoiding mutation of intrinsic objects.
+> If you don't have the capacity but would accept a PR, please comment to that
+> effect so that a volunteer knows their efforts would be welcomed.
+> ```
+
 We find that library authors are generally amenable to making these small changes to increase
 compatibility with any environment that protects itself from prototype pollution attacks by freezing
 intrinsics, including `ses`.
--- a/packages/ses/README.md
+++ b/packages/ses/README.md
@@ -726,11 +726,42 @@ instructions in [SECURITY.md][] to report security-sensitive bugs privately.
 
 For non-security bugs, please use the [regular Issues page][SES Issues].
 
+## Ecosystem Compatibility
+
+Most ordinary JavaScript can run without issues in a realm locked down by SES.
+Exceptions are tracked at [issue #576][incompatibility tracking], and almost
+always take the form of assignments that fail because the
+"[override mistake][override mistake]" prevents overriding properties inherited
+from a frozen intrinsic object in the prototype chain. When that is the case,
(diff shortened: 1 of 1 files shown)


packages/ses/README.md
```diff
@@ -726,11 +726,39 @@ instructions in [SECURITY.md][] to report security-sensitive bugs privately.
...
+    value: <rhs>,
+    writable: true,
+    enumerable: <true | false>,
+    configurable: true,
+  },
+});
+```
+
+Upon encountering an incompatibility, we recommend that you add a comment to
+[issue #576][incompatibility tracking] and file an issue with the external project referencing this section.
```
kriskowal, line 752:
```suggestion
[issue #576][incompatibility tracking] and file an issue with the external project referencing this section.
We find that library authors are generally amenable to making these small changes to increase
compatibility with any environment that protects itself from prototype pollution attacks by freezing
intrinsics, including `ses`.
```

kriskowal (approved):

packages/ses/README.md
```diff
@@ -726,11 +726,39 @@ instructions in [SECURITY.md][] to report security-sensitive bugs privately.
 
 For non-security bugs, please use the [regular Issues page][SES Issues].
 
+## Ecosystem Compatibility
+
+Most ordinary JavaScript can run without issues in a realm locked down by SES.
+Exceptions are tracked at [issue #576][incompatibility tracking], and almost
+always take the form of assignments that fail because the
+"[override mistake][override mistake]" prevents overwriting properties inherited
```
erights, line 734:
```suggestion
"[override mistake][override mistake]" prevents overriding properties inherited
```
"overwriting" encourages a common misunderstanding that the issue is assignment to the property itself.

packages/ses/README.md
```diff
@@ -726,11 +726,39 @@ instructions in [SECURITY.md][] to report security-sensitive bugs privately.
...
+objects are frozen (such as in Node.js with the
+[`--frozen-intrinsics`][Node frozen intrinsics] option) and can be fixed by
+replacing `<lhs>.<propertyKey> = <rhs>;` or `<lhs>[<propertyKey>] = <rhs>;` with
+```js
+Object.defineProperties(<lhs>, {
+  [<propertyKey>]: {
+    value: <rhs>,
+    writable: true,
+    enumerable: <true | false>,
+    configurable: true,
```
erights, line 746: A detail worth explaining is that if the intent is to emulate the original assignment, than all of these attributes should be true. Otherwise, any of them may be changed according to the preferences of the maintainers of the system being fixed.
```suggestion
    writable: true,
    enumerable: true,
    configurable: true,
```

erights (approved):
Love it!

LGTM, thanks!

packages/ses/README.md
```diff
@@ -726,11 +726,42 @@ instructions in [SECURITY.md][] to report security-sensitive bugs privately.
...
+  [<propertyKey>]: {
+    value: <rhs>,
+    writable: true,
+    enumerable: true,
+    configurable: true,
+  },
+});
+```
+
+Upon encountering an incompatibility, we recommend that you add a comment to
```
turadg, line 751: why should someone do this?

packages/ses/README.md
```diff
@@ -726,11 +726,42 @@ instructions in [SECURITY.md][] to report security-sensitive bugs privately.
...
+[issue #576][incompatibility tracking] and file an issue with the external project referencing this section.
```
turadg, line 752: we could reduce the load for filing the issue by providing a template. e.g. 
> Some applications defend themselves by freezing their intrinsics. This project has some assignments or mutations that break in such an environment.
> Specifically,
> [link to source in the project]
>
> Please consider making the project compatible with these environments by avoiding mutations to intrinsics.
> If you don't have the capacity but would accept a PR, please comment to that effect so that a volunteer knows their efforts would be welcomed.

gibson042: Good idea, and thank you for the starting point. PTAL.

erights: I think we can narrow this from "assignment or mutation" to just "assignment".

> by avoiding mutations to intrinsics

Is sometimes the issue and should be mentioned when relevant. But by far the most common case is using assignment to override a property inherited from a frozen intrinsic, i.e., the override mistake. We should always lead with the override mistake.

erights: And for the override mistake, always first suggest doing the mutation by `defineProperty` rather than assignment, because that is much more likely to be correctness preserving and least effort.

erights: Or rather `defineProperties` since that's what is already shown in this PR's example code.

turadg (approved):

PR #1613, opened 2023-06-06
docs(ses): Improve SES README diction/grammar/etc.

gibson042:


packages/ses/README.md
```diff
@@ -593,31 +596,28 @@ compartment.evaluate(program1);
 compartment.evaluate(program2);
 ```
 
-However such programs (`program`, `program1`, or `program2`) are only
-as useful as a calculator.
+However, such guest programs (`program`, `program1`, or `program2`) are only
+useful as glorified calculators.
```
kriskowal, line 600: 🐸

packages/ses/README.md
```diff
@@ -647,17 +646,16 @@ buffer to construct a high resolution timer.
...
 When a program interacts with an object introduced by another program (as
 through the per-compartment `globalThis`, function arguments or returned
 values), there are potential risks due to the synchronous nature of object
 access.
 Even interactions that are not explicit function calls may cause code from
 another program, like property accessors or proxy traps, to execute on the same
-stack, which may be able to call back into the program (reentrancy), throw, or
-detect the current stack height.
+stack, which may be able to detect the stack height, throw an exception, or call
+back into the program in pursuit of a reentrancy attack.
```
kriskowal, line 658: Why do we use “reentrancy” when “reentrance” is right there? (I know why we don’t use “reënter”, mwahah)

kriskowal (approved):

packages/ses/README.md
```diff
@@ -140,8 +141,8 @@ console.log(Object.isFrozen(capability.inc));
 
 Note that although the **surface** of the capability is frozen, the capability
 still closes over the mutable counter.
-Hardening an object graph makes the surface immutable, but does not make
-methods pure.
+Hardening an object graph makes the surface immutable, but does not guarantee
+that methods are free of side effects.
```
erights, line 145: Worth working in here that the purpose of `freeze` and `harden` is not primarily about immutability, but primarily about tamper proofing API surfaces.

packages/ses/README.md
```diff
@@ -573,9 +575,10 @@ const compartment = new Compartment();
 compartment.evaluate(program);
 ```
 
-If the host program (Figure 2) arranges for the compartment's `globalThis` to
+If the host program arranges for the compartment's `globalThis` to
```
erights, line 578: Why remove the "Figure" references by leave the labeling in the figures themselves?

packages/ses/README.md
```diff
@@ -647,17 +646,16 @@ buffer to construct a high resolution timer.
...
 
 When a program interacts with an object introduced by another program (as
 through the per-compartment `globalThis`, function arguments or returned
 values), there are potential risks due to the synchronous nature of object
 access.
 Even interactions that are not explicit function calls may cause code from
 another program, like property accessors or proxy traps, to execute on the same
-stack, which may be able to call back into the program (reentrancy), throw, or
-detect the current stack height.
+stack, which may be able to detect the stack height, throw an exception, or call
```
erights, line 657: detect the stack height how? (assuming only safe lockdown settings)

erights (approved):
LGTM, thanks!

packages/ses/README.md
kriskowal: Catch a range error of an infinitely recursive function that increments a counter.

erights: Not under XS or any system that implements https://github.com/tc39/proposal-oom-fails-fast , because stack overflow is terminating. No one around to observe the error.

However, this does still leave open the anthropic side channel https://esdiscuss.org/topic/the-anthropic-side-channel-was-how-would-shallow-generators-compose-with-lambda . I don't think there is anything we can do about that.

packages/ses/README.md
gibson042: Good point. I just added https://github.com/endojs/endo/pull/1613/commits/9a22e01c10d431775fe14bdfdde7aac60aef61ea to replace the figure numbering with linkable section headings; PTAL.

PR #1615, opened 2023-06-06
Endo Daemon CLI Subcommand Descriptions

kriskowal:
This change adds descriptions to all the `endo` CLI subcommands so far implemented.

### DIFF
--- a/packages/cli/src/endo.js
+++ b/packages/cli/src/endo.js
@@ -116,54 +116,87 @@ export const main = async rawArgs => {
     packageDescriptorPath,
   );
   const packageDescriptor = JSON.parse(packageDescriptorBytes);
-  program.name(packageDescriptor.name).version(packageDescriptor.version);
+  program.name('endo').version(packageDescriptor.version);
 
-  const where = program.command('where');
+  const where = program
+    .command('where')
+    .description('prints paths for state, logs, caches, socket, pids');
 
-  where.command('state').action(async _cmd => {
-    process.stdout.write(`${statePath}\n`);
-  });
+  where
+    .command('state')
+    .description('prints the state directory path')
+    .action(async _cmd => {
+      process.stdout.write(`${statePath}\n`);
+    });
 
-  where.command('run').action(async _cmd => {
-    process.stdout.write(`${ephemeralStatePath}\n`);
-  });
+  where
+    .command('run')
+    .description('prints the daemon PID file path')
+    .action(async _cmd => {
+      process.stdout.write(`${ephemeralStatePath}\n`);
+    });
 
-  where.command('sock').action(async _cmd => {
-    process.stdout.write(`${sockPath}\n`);
-  });
+  where
+    .command('sock')
+    .description('prints the UNIX domain socket or Windows named pipe path')
+    .action(async _cmd => {
+      process.stdout.write(`${sockPath}\n`);
+    });
 
-  where.command('log').action(async _cmd => {
-    process.stdout.write(`${logPath}\n`);
-  });
+  where
+    .command('log')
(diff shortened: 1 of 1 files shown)


packages/cli/src/endo.js
```diff
@@ -324,80 +380,95 @@ export const main = async rawArgs => {
...
-      const pet = await E(bootstrap).provide(name);
-      console.log(pet);
-    } catch (error) {
-      console.error(error);
-      cancel(error);
-    }
-  });
+  program
+    .command('show <name>')
+    .description('prints a representation of the named value')
```
dckc, line 403: what does "representation" add?

```suggestion
    .description('prints the named value')
```

I tried the [console.log docs](https://nodejs.org/api/console.html#consolelogdata-args); they're no help: "Prints to stdout with newline. ... the arguments are all passed to `util.format()`)"

Another angle: "carries out the named formula and prints the result".

I see `provide` in the implementation, which suggests something like: "looks the name up in a cache; carries out the named formula on cache miss; prints the result". The relationship between cache entry lifetimes and the daemon lifetime also seems relevant.

packages/cli/src/endo.js
```diff
@@ -324,80 +380,95 @@ export const main = async rawArgs => {
...
-    try {
-      const bootstrap = getBootstrap();
-      const iterable = await E(bootstrap).provide(name);
-      const iterator = await E(iterable)[Symbol.asyncIterator]();
-      for await (const iterand of makeRefIterator(iterator)) {
-        console.log(iterand);
+  program
+    .command('follow <name>')
+    .description(
+      'prints a representation of each value from the named async iterable as it arrives',
```
dckc, line 423:
```suggestion
      'prints each value from the named async iterable as it arrives',
```

dckc (approved):
some of my comments are more relevant to a spec; I suppose command descriptions don't need to be a complete specification to be useful

kriskowal:
Per this feedback, I took a run at making these descriptions much more concise at the end of the `endo` branch. The spike continues but I’ve left these PRs behind.

```
  open [options] <name> [filePath]                      opens a web page (weblet)
  run [options] [<file>] [<args>...]                    runs a program (runlet)
  make [options] [file]                                 make a plugin or a worker caplet (worklet)
  inbox [options]                                       read messages
  request [options] <informal-description>              ask someone for something
  resolve [options] <request-number> <resolution-name>  grant a request
  reject [options] <request-number> [message]           deny a request
  send [options] <party> <messag>                       send a message with @named-values @for-you:from-me
  adopt [options] <message-number> <name-in-message>    accept a @value from a message
  dismiss [options] <message-number>                    delete a message
  list [options]                                        show names
  remove [options] [names...]                           forget a named value
  rename [options] <from> <to>                          change the name for a value
  show [options] <name>                                 prints the named value
  follow [options] <name>                               subscribe to a stream of values
  cat [options] <name>                                  dumps a blob
  store [options] <path>                                stores a blob
  eval [options] <source> [names...]                    creates a value
  spawn [options] [names...]                            creates a worker
  bundle [options] <application-path>                   stores a program
  mkhost [options] <name>                               makes a separate mailbox and storage for you
  mkguest [options] <name>                              makes a mailbox and storage for a guest (peer or program)
  where                                                 prints paths for state, logs, caches, socket, pids
  start                                                 start the endo daemon
  stop                                                  stop the endo daemon
  restart                                               stop and start the daemon
  clean                                                 erases ephemeral state
  reset                                                 erases persistent state and restarts if running
  log [options]                                         writes out the daemon log, optionally following updates
  ping                                                  prints ok if the daemon is responsive
  help [command]                                        display help for command
```

PR #1617, opened 2023-06-07
fix(ses)! make `LOCKDOWN_...` options less noisy

erights:
Remove the `console.warn` logic for reporting which `LOCKDOWN_...` options were used. The extra noise does not seem to serve any purpose.


FUDCo (approved):
Yay!

PR #1618, opened 2023-06-08
feat(ses): whitelist some recent >= stage 3 additions

erights:


packages/ses/src/whitelist.js
```diff
@@ -932,14 +935,22 @@ export const whitelist = {
       values: 'boolean',
       // Failed tc39 proposal
       // Seen on FF Nightly 88.0a1
-      at: false,
+      at: 'boolean',
```
erights, line 938: Saw the false, and even though `at` is a withdrawn experiment, I wonder if we should always err on unscopables towards inclusion?

mhofman:
In the latest XS 14.1, I note the following removals:

```
Removing intrinsics.%SharedSymbol%.dispose
Removing intrinsics.%ArrayPrototype%.toReversed
Removing intrinsics.%ArrayPrototype%.toSorted
Removing intrinsics.%ArrayPrototype%.toSpliced
Removing intrinsics.%ArrayPrototype%.with
Removing intrinsics.%ArrayPrototype%.@@unscopables.toReversed
Removing intrinsics.%ArrayPrototype%.@@unscopables.toSorted
Removing intrinsics.%ArrayPrototype%.@@unscopables.toSpliced
Removing intrinsics.%ArrayBufferPrototype%.detached
Removing intrinsics.%TypedArrayPrototype%.toReversed
Removing intrinsics.%TypedArrayPrototype%.toSorted
Removing intrinsics.%TypedArrayPrototype%.with
```

Which seem to all be covered by this PR. 👍

packages/ses/src/whitelist.js
```diff
@@ -932,14 +936,26 @@ export const whitelist = {
...
       // See https://github.com/tc39/proposal-array-find-from-last
       findLast: 'boolean',
       findLastIndex: 'boolean',
+      toReversed: 'booleab',
```
mhofman, line 943:
```suggestion
      toReversed: 'boolean',
```

erights: Done

packages/ses/src/whitelist.js
kriskowal: If the property is present on Array.prototype, we certainly want it to be unscopable. But, lockdown will remove it if it is present.

kriskowal (approved):

packages/ses/src/whitelist.js
erights: I just checked, and we do list `Array.prototype.at` as not to be removed. Given your remarks, it seems that not removing it on unscopables was indeed a correct fix, so Resolving. However, I'm curious: Given

> If the property is present on Array.prototype, we certainly want it to be unscopable.

Shouldn't all methods on Array.prototype already be unscopable in plain JS? Node JS does not even come close:

```
> gopn = Object.getOwnPropertyNames
[Function: getOwnPropertyNames]
> gopn(Array.prototype).length
34
> gopn(Array.prototype[Symbol.unscopables]).length
11
```

kriskowal: My understanding is that methods of arrays that predates the introduction of unscopables must not be unscopable, and all that followed must be unscopable, such that all legacy code that uses arrays as with block contexts continue to work, both in the case where they intentionally invoke a method of the array and all the cases where a new array method would overshadow a lexical function. 

with (arr) push(10)

function flatMap() {}
with (arr) flatMap()

erights: Makes sense, thanks!

Issue #835, opened 2021-07-10
Possible bug in allowlist visitor getSubPermit


kriskowal:
The function `getSubPermit` could possibly fail to mark a virtualized native function if that case gets overshadowed by preceding case. We need to answer what kind of arrangement in the allowlist and global environment could induce invalid behavior.

```js
  function getSubPermit(obj, permit, prop) {
    const permitProp = prop === '__proto__' ? '--proto--' : prop;
    if (objectHasOwnProperty(permit, permitProp)) {
      return permit[permitProp];
    }

    // TODO maybe above and below should be swapped
    // so that makeVirtualizedNativeFunction gets called in all proper cases

    if (typeof obj === 'function') {
      markVirtualizedNativeFunction(obj);
      if (objectHasOwnProperty(FunctionInstance, permitProp)) {
        return FunctionInstance[permitProp];
      }
    }

    return undefined;
  }
```

erights:
OMG @kriskowal , the bug I found and fixed at https://github.com/endojs/endo/pull/1619 is one you already reported!

In any case, closing as fixed by https://github.com/endojs/endo/pull/1619

PR #1619, opened 2023-06-08
fix(ses): missing native function markings

### LINKED ISSUE #1228: Function.toString is not marked as native code
image

erights:
Fixes #1228 
Fixes https://github.com/endojs/endo/issues/835

The replacement `Function.prototype.toString`, which uses the synthetic native function marking, should itself have been marked as a native function, so that it would recognize itself as such. First, I saw I needed to install it early. But it still wasn't being marked, which was quite puzzling. Turns out there was a bug in the marking logic in the whitelist-intrinsics code, where functions could both skip being marked, or could be pointlessly marked redundantly.

@Jack-Works , please consider yourself a reviewer of this PR. I look forward to your comments. And thanks for the original report! If I hadn't investigated I would not have found this deeper bug.

Turns out that this deeper bug was already reported by @kriskowal at #835

### DIFF
--- a/packages/ses/src/lockdown-shim.js
+++ b/packages/ses/src/lockdown-shim.js
@@ -250,6 +250,10 @@ export const repairIntrinsics = (options = {}) => {
 
   tameDomains(domainTaming);
 
+  // Replace Function.prototype.toString with one that recognizes
+  // shimmed functions as honorary native functions.
+  const markVirtualizedNativeFunction = tameFunctionToString();
+
   const { addIntrinsics, completePrototypes, finalIntrinsics } =
     makeIntrinsicsCollector();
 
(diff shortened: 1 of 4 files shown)


Jack-Works (approved):

packages/ses/src/whitelist-intrinsics.js
```diff
@@ -261,6 +260,10 @@ export default function whitelistIntrinsics(
     const protoName = permit['[[Proto]]'];
     visitPrototype(path, obj, protoName);
 
+    if (typeof obj === 'function') {
+      markVirtualizedNativeFunction(obj);
+    }
```
kriskowal, line 265: Yes!

erights:
@kriskowal FYI I still need your approval

kriskowal (approved):

Issue #1540, opened 2023-04-06
Node 14 sunset


naugtur:
https://nodejs.dev/en/about/releases/

Node 14 maintenance ends this month. We might want to switch everything to newer versions.

erights:
This would let us assume `Object.hasOwn` and phase out `hasOwnPrototypeOf`. What else?

naugtur:
String.replaceAll
Promise.any
WeakRef
Logical Assignment

Got them all from:
https://node.green/#ES2021
https://node.green/#ES2022

PR #1623, opened 2023-06-08
update Node versions

turadg:
closes: https://github.com/endojs/endo/issues/1540

Node 14 is EOL and Node 18 is now the Active LTS. 

This updates engines, types deps, and CI accordingly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,14 +115,10 @@ jobs:
       fail-fast: false
       matrix:
         node-version:
-          # 14.2 not longer tested because the test's runner, Ava, now requires 14.19+ and fails with:
-          #   Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only file and data URLs are supported by the default ESM loader
-          # - '14.2' # last version before promise fast-path without destroy
-          # - '14.17' # last version before unconditional promise fast-path
-          - '14.18' # first version after unconditional promise fast-path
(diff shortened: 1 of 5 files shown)


kriskowal:
Also, thank you.

turadg:
> expecting CI to tell you something broke

It did a couple times to get here :)

turadg:
I also had to update branch protection from…
Screenshot 2023-06-08 at 2 58 30 PM

… to …
Screenshot 2023-06-08 at 3 11 36 PM

I tinkered with simplifying the CI config so the jobs that don't need a matrix don't include the Node version in their job title, but punted for another day.

erights:
Note https://github.com/endojs/endo/pull/1627#issuecomment-1583613078

mhofman:
We should probably adopt the same engine version checks that agoric-sdk has, which is more specific about LTS versions only.

erights:
I don't understand how CI was able to use Node 16.5 at all. On that Node, when I `yarn` at the top, I get
```
$ yarn
yarn install v1.22.0
[1/5] 🔍  Validating package.json...
[2/5] 🔍  Resolving packages...
[3/5] 🚚  Fetching packages...
error ava@5.2.0: The engine "node" is incompatible with this module. Expected version ">=14.19 <15 || >=16.15 <17 || >=18". Got "16.5.0"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

PR #1627, opened 2023-06-08
refactor: hasOwn replaces hasOwnPropertyOf

erights:
Enabled by https://github.com/endojs/endo/pull/1623 , since Node 14 did not have `Object.hasOwn`.

`Object.hasOwn` replaces `hasOwnPropertyOf` and `objectHasOwnProperty`. However, to avoid breaking external dependencies, the old names are still exported, but now deprecated.

See https://github.com/Agoric/agoric-sdk/pull/7902

### DIFF
--- a/packages/static-module-record/test/fixtures/exportheavy.js
+++ b/packages/static-module-record/test/fixtures/exportheavy.js
@@ -7,7 +7,6 @@ export {
   isObject,
   assertChecker,
   getTag,
-  hasOwnPropertyOf,
 } from './src/helpers/passStyle-helpers.js';
 
 export { getErrorConstructor, toPassableError } from './src/helpers/error.js';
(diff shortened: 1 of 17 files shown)


packages/static-module-record/test/fixtures/exportheavy.js
```diff
@@ -42,9 +41,6 @@ export {
 // eslint-disable-next-line import/export
 export * from './src/types.js';
 
-
-const { details: X, Fail } = assert;
```
erights, line 46: This seems like dead code, so I deleted it as a drive by. Nothing to do with the purpose of this PR

erights:
Converting to draft because `Object.hasOwn` was introduced by Node 16.9 but #1623 allows earlier versions of Node 16. In particular, some tests seem to be using 16.5

PR #1628, opened 2023-06-09
DO NOT MERGE: @endo/daemon @endo/cli ongoing documentation

tgrecojs:
*This PR is intended to stay in draft state*

### @endo/daemon & @endo/cli documentation
Branch has been created for the purpose of having somewhere to stash my questions as well as additional, more extensive docs related to my dealings with endo.

@kriskowal When you have a moment I'd love to get any feedback you may have. In the event that my notes are confusing, or something is incorrect please let me know! There's no rush on this but I thought I had enough material to warrant pushing up. That said, I definitely expect to add more to this page before we reconvene next week.


kriskowal (commented):
Partial review.

packages/daemon/daemon-docs.md
```diff
@@ -0,0 +1,112 @@
...
+  // @ts-expect-error zero-argument promise resolve
+  cancel();
+  await closed.catch(() => {});
+};
+```
+
+### `clean`
+
+- recursively and forcibly removes any directories that have been created throughout the lifecycle of the previous Endo daemon.
+- takes one argument, `locator`. If nothing is passed in, this defaults to [`defaultLocator`](#defaultlocator).
```
kriskowal, line 47: Resets the Endo daemon’s persistent state to “factory defaults”. Actually an open question whether this will destroy the user’s identity. It might in fact do so, but maybe we’d require some “yes I mean it” incantation.

The purpose of the locator is to afford some flexibility for testing.

packages/daemon/daemon-docs.md
```diff
@@ -0,0 +1,112 @@
...
+
+### `stop`
+
+- terminates existing Endo daemon processess before calling [`clean`](#clean).
```
kriskowal, line 51: The cleaning step just gets rid of the domain socket, in case the Endo daemon didn’t stop gracefully and clean up after itself.

packages/daemon/daemon-docs.md
```diff
@@ -0,0 +1,112 @@
...
+- takes one argument, `locator`. If nothing is passed in, this defaults to [`defaultLocator`](#defaultlocator).
+
+### `makeEndoClient`
+
+- Used to spin up a new application within the context of an Endo daemon.
+- The processess created with makeEndoClient come pre-assembled with the ability to communicate over CapTP using Node.js-based messages.
+- Utilizes the global `net` package to create a socket connection for an Endo process. (`net.connect(sockPath)`)
```
kriskowal: This actually just establishes a CapTP connection over the Endo daemon’s UNIX domain socket or Windows named pipe, so a client like the CLI can send and receive messages.

packages/daemon/daemon-docs.md
```diff
@@ -0,0 +1,112 @@
...
+const defaultLocator = {
+  statePath: whereEndoState(process.platform, process.env, info),
+  ephemeralStatePath: whereEndoEphemeralState(
+    process.platform,
+    process.env,
+    info,
+  ),
+  sockPath: whereEndoSock(process.platform, process.env, info),
+  cachePath: whereEndoCache(process.platform, process.env, info),
+};
```
kriskowal: Best to refer to `@endo/where` for details about what’s going on here, and maybe even move the concept of the locator to that package. The `@endo/where` package touches a lot of details about the conventional locations for application support files across all modern operating systems, starting with XDG as a baseline. The locator captures all these decisions so that the daemon, its clients, and people debugging can rendezvous with the relevant files.

packages/daemon/daemon-docs.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+# Questions
+
+- the `main` function seems is where “all of it” starts.
+- where does `main0` come from?
```
kriskowal: An Endo extension is a JavaScript module and its transitive dependencies that has been captured in a bundle. Rather than endowing the entire extension with direct access to the “powers”, “outbox”, or “guest program’s permission management facet”, Endo conventionally expects the bundle’s entry module to export a function that can be used to construct the extension’s API object and receive the extension’s powers. This function can then delegate authorities piecemeal to each of its components according to the principle of least-authority. That is: if a component doesn’t need it, it doesn’t get it.

packages/daemon/daemon-docs.md
```diff
@@ -0,0 +1,112 @@
...
+- Regarding the locator...
+  - what is the difference between:
+    - statePath
+    - ephemeralStatPath
+    - sockPath
+    - cachePath
+  - after coming across the term "noncelocator" a number of times while reading erights.org, I'm wondering what is the locator within the context of SwingSet?
```
kriskowal: I don’t know about these things but there might be a convergent reason for them. My impression is that these are different locators. The Endo pet daemon locator is just for allowing various components to rendezvous and communicate with each other.

packages/daemon/daemon-docs.md
```diff
@@ -0,0 +1,112 @@
...
+- What is the significance of `child.on` within the endo daemon's `start` function? What exactly occurs when a “message” event occurs?
```
kriskowal, line 93: This is one of the wonders of daemonization. Daemonization is the process of spawning a child process, removing it from the parent process’s session, and handing it over to process 0 (init) to manage until it dies. On UNIX, by default, a child process remains in the parent process’s session, sees all of the same signals, and exits when the parent exits. Daemons are longer-lived.

When a client like the CLI attempts to open a connection to the Endo daemon, there’s a possibility that the daemon is not running. So, the client will spawn and daemonize the daemon. But, owing to the vagaries of timing, the daemon will not be ready to receive connections from the client until it has run a while and started listening.

This is a problem for the client which wants to connect to the daemon the moment it becomes available. We’re using the Node.js IPC port to send a message from the child process (daemon) to the parent process (cli) that it can connect.

The alternative is a retry loop, which will always be slower than this approach.

packages/daemon/daemon-docs.md
```diff
@@ -0,0 +1,112 @@
...
+- What facets does `@endo/daemon` make accessible upon calling the `start` function?
+- Confusion around the `bootstrap` object.
+  - `makeMessagesCapTP`
+    - This function takes a `bootstrap` object as an argument, which is then passed into `makeCapTP` in exchange for a few values, one of which being `getBootstrap`.
+    - I'm trying to figure out how imports/exports of bootstrap objects work.
```
kriskowal: CapTP connections are initially between the local bootstrap object and the remote bootstrap object. CapTP is bidirectional and there are potentially two bootstrap objects. Both are optional.

packages/daemon/daemon-docs.md
```diff
@@ -1,26 +1,29 @@
...
-- shorthand method for executing `endo stop`
+Upon launch of the "main" module, the Endo daemon is comprised of the following powers:
 
-# `@endo/daemon`
+- `crypto`
+- `net`
+- `fs`
+- `path`
+- `popen`
+- `url`
```
kriskowal, line 22: This is a very Node.js specific list of powers needed in order to construct the powers actually needed by the daemon. Those are defined by `DaemonicPowers` in `types.d.ts`.

packages/daemon/daemon-docs.md
```diff
@@ -10,9 +10,9 @@ Endo makes use of the Powerbox pattern[^1]. By employing this pattern, Endo acts
 
 > **The powerbox is a composition of objects that grants, revokes, negotiates, and in general manages, the authorities granted to another object.** - _How Emily Tamed the Caml_
 
-## Methods
+At the heart of the Endo Powerbox (and all programs that implement the powerbox pattern) is a `main` module for ensuring that each object is given the permissions that it needs to function properly. If implemented correctly, then the `main` module will hand out to each module **just enough authority to do the job needed by that particular module.**
 
-Upon launch of the "main" module, the Endo daemon is comprised of the following powers:
+The following powers made accessible through Endo's `main` module:
```
kriskowal, line 15: Again, to clarify, these Node.js modules provide more than the Endo daemon needs. The `daemon-powers-node.js` module attenuates / trims these down to the `DaemonPowers` it needs, and the `DaemonPowers` type avoids closing over Node.js implementation details. So, theoretically, the powers could be constructed from other builtin modules on other environments.

packages/daemon/daemon-docs.md
```diff
@@ -0,0 +1,112 @@
...
+
+### `endo restart`
+
+- shorthand method for executing `endo stop`
+
+# `@endo/daemon`
+
+### `start`
+
+- handles setup of the initial process for creating Endo applications, and communicating across
```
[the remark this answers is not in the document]
tgrecojs: 👍

packages/daemon/daemon-docs.md
```diff
@@ -0,0 +1,112 @@
...
+
+- executes [stop](#start) function definined within the `@endo/daemon`.
+
+### `endo stop`
+
+- executes [stop](#stop) function definined within the `@endo/daemon`.
+
+### `endo clean`
+
+- executes the [`clean`](#clean) function definined within the `@endo/daemon`.
```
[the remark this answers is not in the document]
tgrecojs: To clarify - "resetting to factory defaults" is to reset all settings *that haven't been saved*? I'm inferring that it must be or otherwise this metaphor would apply to `endo reset`.

packages/daemon/daemon-docs.md
tgrecojs: 👍

packages/daemon/daemon-docs.md
tgrecojs: 👍  this has been removed in the latest version.

packages/daemon/daemon-docs.md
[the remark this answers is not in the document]
kriskowal: I was mistaken. `clean` only deletes the domain socket or named pipe.

I expect that we’ll just get rid of this command. I used it while debugging.

packages/daemon/daemon-docs.md
```diff
@@ -1,26 +1,29 @@
-# `@endo/cli`
+# `@endo/daemon`
 
-### `endo start`
+The Endo Daemon allows us to bootstrap systems in which POLA confinement can be accomplished at the deepest level of granularity. In the case of an Endo application, the deepest level of granularity refers to the lexical environment of each object that exists within the system. This is otherwise known as the "object level".
```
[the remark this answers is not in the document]
tgrecojs: 👍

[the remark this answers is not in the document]
tgrecojs: Ok, good to know. For now, I've swapped out use of the term "application" for "extension" when referring to Endo programs.

PR #1636, opened 2023-06-14
parameterize InterfaceGuard

turadg:
Replaces https://github.com/Agoric/agoric-sdk/pull/7375/ now that it's in Endo.

Parameterizes the `InterfaceGuard` type.

The immediate benefit is getting static errors for mismatch between interface guards and behavior methods.


erights (approved):
LGTM, thanks!

PR #1638, opened 2023-06-15
test(patterns): revert error msg improvement that breaks old tests

erights:
Fixes https://github.com/endojs/endo/issues/1640

https://github.com/endojs/endo/pull/1606 improved at least one error message, which is great. However, some tests in current agoric-sdk currently depend on the old error message, which this PR restores. 

Note that this PR restores the message, but not by reverting to the code before #1606. Rather, it continues to use the mechanisms introduced in #1606 to output the earlier message.

TODO: Upgrade the tests in agoric-sdk that depend on the old message to also tolerate the improvement.

### DIFF
--- a/packages/patterns/src/patterns/patternMatchers.js
+++ b/packages/patterns/src/patterns/patternMatchers.js
@@ -953,8 +953,19 @@ const makePatternKit = () => {
         false,
         // `label` can be embedded without quotes because it is provided by
         // local code like `M.remotable("...")`.
-        X`${specimen} - Must be a remotable ${b(label)}, not ${kindDetails}`,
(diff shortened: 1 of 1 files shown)


erights:
Example at https://github.com/Agoric/agoric-sdk/actions/runs/5273958254/jobs/9538040586?pr=7926#step:8:23

gibson042:
Ugh, I was hoping not to see such tests outside of endo. But fortunately, there didn't seem to be many of them—alternative PR that loosens the error message assertions within agoric-sdk is now up at https://github.com/Agoric/agoric-sdk/pull/7932 .

erights:
Closing in favor of https://github.com/Agoric/agoric-sdk/pull/7932

PR #1639, opened 2023-06-15
refactor(ses): tolerate core-js better

erights:
At https://github.com/zloirock/core-js#ecmascript-symbol we see that core-js adds the non-standard `Symbol.useSimple` and `Symbol.useSetter`. Before this PR, our whitelisting mechanism, having never heard of these, safely removes them during `lockdown`. But because they're unexpected, ses also emits a console diagnostic.

This PR enhances our whitelist so that these two properties are silently removed.

### DIFF
--- a/packages/ses/src/permits.js
+++ b/packages/ses/src/permits.js
@@ -518,6 +518,10 @@ export const permitted = {
     toPrimitive: 'symbol',
     toStringTag: 'symbol',
     unscopables: 'symbol',
+    // Seen at core-js https://github.com/zloirock/core-js#ecmascript-symbol
+    useSimple: false,
+    // Seen at core-js https://github.com/zloirock/core-js#ecmascript-symbol
+    useSetter: false,
   },
 
   '%SymbolPrototype%': {


erights:
@samsiegart I added you as a reviewer because, AFAIK, we only encounter core-js in UI code. While this removal preserves SES safety, does it hurt core-js? Is it possible we could stop depending on core-js?

samsiegart:
@erights I've never had to deal with core-js directly so I'm a bit confused, where do we depend on it?

erights:
Hi @samsiegart 

At https://github.com/Agoric/agoric-sdk/actions/runs/5283428596/jobs/9559785872?pr=7926#step:19:33 I see

```
test/agoric-wallet-connection.test.js:

 🚧 Browser logs:
      ...
      Removing intrinsics.%SharedSymbol%.useSetter
      Removing intrinsics.%SharedSymbol%.useSimple
      ...
```

These console messages are from the SES whitelisting mechanism, when they remove entries they encounter that they've never heard of. Searching, I only see these two as coming from core-js. Other than that clue, I do not know why we're loading core-js.

What browser or browser emulation does CI use? Perhaps core-js was pulled in by that browser emulation?

kriskowal (approved):

samsiegart:
@erights We did have chrome webdriver tests, but we actually removed that code this week, so maybe the issue has solved itself as well. If @kriskowal approves then I have no objections to this PR

PR #1643, opened 2023-06-15
fix(ses): permits for new proposal problems

erights:
When doing https://github.com/endojs/endo/pull/1618

I had misread @gibson042 's proposal as adding a `toRawJSON` rather than `isRawJSON`.
I had missed that ArrayBuffer `detached` is an accessor.

I also added citations for the #1618 additions. While doing so, I remain confused about where `group` and `groupToMap` came from. I also noticed `groupBy` missing.

### DIFF
--- a/packages/ses/src/permits.js
+++ b/packages/ses/src/permits.js
@@ -437,6 +437,8 @@ export const permitted = {
     seal: fn,
     setPrototypeOf: fn,
     values: fn,
+    // https://github.com/tc39/proposal-array-grouping
+    groupBy: 'boolean',
   },
 
   '%ObjectPrototype%': {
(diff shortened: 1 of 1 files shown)


packages/ses/src/whitelist.js
```diff
@@ -940,22 +944,28 @@ export const whitelist = {
...
     toReversed: fn,
     toSorted: fn,
     toSpliced: fn,
     with: fn,
-    group: fn,
-    groupToMap: fn,
+    // https://github.com/tc39/proposal-array-grouping
+    group: fn, // Not in proposal? Where?
+    groupToMap: fn, // Not in proposal? Where?
+    groupBy: fn,
```
kriskowal, line 968: I believe these have been withdrawn. Vendors unshipped `group` at least on account of web compat concerns. Judging by reactions in TC39’s Matrix chat, Mozilla and Google have both lost their appetite for any further properties on `Array.prototype`.

kriskowal (approved):
I see no harm in permitting the `group*` features through the SES shim, but I think they’ve been unshipped and may already be “unspecified”.

PR #1644, opened 2023-06-15
permits naming

turadg:
Adopt more specific terminology for the permitted properties.


kriskowal (approved):
This turned out to conjugate better than I hoped.

Please coördinate with @erights who has pending PRs that overlap permits. This tends to be the kind of change Git is good at reconciling.

turadg:
> This tends to be the kind of change Git is good at reconciling.

Indeed. I checked out https://github.com/endojs/endo/pull/1643 and rebased it onto this branch. It went cleanly without manual edits.

So I'm going to merge this now.

PR #1649, opened 2023-06-23
fix(static-module-record): Do not sort imports

kriskowal:
Sorting imports alters the intended execution order. This composes poorly with programs that carefully order their imports for side-effects, for example, `ses`.

There remains an issue that all `reexports` will be topologically sorted after all `imports`, and the fix for that is to adopt the `bindings` design in XS, since that preserves all ordering concerns.


kriskowal:
cc @naugtur when you’re back. cc @kumavis until then.

naugtur:
How was the order of imports affecting anything? In CJS I'm pretty sure order is preserved by the virtue of `require` being synchronous. I'm assuming this is only relevant for ESM, right?

kriskowal:
> How was the order of imports affecting anything? In CJS I'm pretty sure order is preserved by the virtue of `require` being synchronous. I'm assuming this is only relevant for ESM, right?

This affects execution order for both ESM and CJS for every loader variant we support.

For example, the bundler linearizes execution order using a depth first search / topological sort based on the order of appearance of imports (and then reexports). That should be reframed to occur in order of appearance of bindings, but that is a project for another day.

For both bare Compartment and the other Compartment Mapper loaders, the toposort emerges from the behavior of the Compartment in the link and instance modules of SES.

michaelfig (approved):
Thanks for the comprehensive tests for this little change.

PR #1650, opened 2023-06-23
feat(compartment-mapper): Bundler support for aliases

kriskowal:
The Compartment Map contains a bundler that is entirely sufficient for bundling SES proper. Since its original design, @naugtur expanded it to handle CommonJS. So, the missing feature that would make it suitable for bundling a larger bootstrapping system, say, one that could bootstrap eventual-send and import-bundle, was module aliases.

This change introduces an alias map, which fast-forwards every import resolution thru aliases to the concrete implementation of the imported module.

### DIFF
--- a/packages/compartment-mapper/src/bundle.js
+++ b/packages/compartment-mapper/src/bundle.js
@@ -29,6 +29,8 @@ import cjsSupport from './bundle-cjs.js';
 
 const textEncoder = new TextEncoder();
 
+const { quote: q } = assert;
+
 /** @type {Record<string, ParserImplementation>} */
 const parserForLanguage = {
   mjs: parserArchiveMjs,
@@ -55,6 +57,7 @@ const sortedModules = (
   entryModuleSpecifier,
 ) => {
   const modules = [];
+  const aliases = new Map();
   const seen = new Set();
 
   /**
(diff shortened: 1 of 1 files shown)


kriskowal:
cc @erights because this greatly expands the reach of our bundler. We should be able to bundle a bootstrapping layer that uses endo/init, endo/eventual-send, endo/captp, and endo/import-bundle.

cc @kumavis @naugtur because this is probably also relevant for the secure bundler you’re working on.

michaelfig (approved):
LGTM!  Straightforward.

packages/compartment-mapper/src/bundle.js
```diff
@@ -239,10 +244,21 @@ export const makeBundle = async (read, moduleLocation, options) => {
...
   for (const module of modules) {
     module.indexedImports = Object.fromEntries(
-      Object.entries(module.resolvedImports).map(([importSpecifier, key]) => [
-        importSpecifier,
-        modulesByKey[key].index,
-      ]),
+      Object.entries(module.resolvedImports).map(([importSpecifier, key]) => {
+        key = aliases.get(key) ?? key;
+        const module = modulesByKey[key];
+        if (module === undefined) {
```
naugtur, line 250: So the original implementation would fail to read the index if an alias was used?

naugtur (approved):

packages/compartment-mapper/src/bundle.js
kriskowal: Indeed, the modules array lacks entries for the aliases.

Issue #1490, opened 2023-02-08
Debugging in Safari is not working for script executing inside SES


argp1:
Steps to reproduce:
1. Run the demo mentioned [here](https://github.com/endojs/endo/blob/686ab61802c0e5206b14f4b91d499e3873555cd4/packages/ses/demos/console/index.html#L18),
2. Put the `debugger` statement anywhere in script code.
3. Open the Safari developer tools.
4. Click on `Execute` button

**Result**: The script doesn't break on debugger statement in Safari. However in Chrome and Firefox, debugging works fine.
**Workaround**: If we add following function [here](https://github.com/endojs/endo/blob/686ab61802c0e5206b14f4b91d499e3873555cd4/packages/ses/src/strict-scope-terminator.js#L59), the issue gets fixed and we can debug in Safari.

```
  // Note: This is a bug of Safari.
  // In Safari, scope handlers call the ownKeys trap, which is not defined. 
  // This causes the alwaysThrowHandler proxy handler to be called and an error will be thrown.
  // In order to create a scope handler, buildScopeHandler must handle ownKeys
  ownKeys(target) {
    return Reflect.ownKeys(target);
  },
```

Please let me know if you need any additional information.

argp1:
Hi @erights , wanted to check if there is any update on this. Please let me know if I can be of any assistance here. Thanks!

erights:
Hi @argp1 , 
thanks for reporting! And sorry for the terribly late response.
I verified that Safari still has this problem. Has the bug been reported to Safari yet?

Hi @mhofman , 
I'm happy to do a PR adding the `ownKeys` method to some proxy handler, as @argp1 suggests. But which proxy handler do you think is best for this?

argp1:
@erights , Nope, I haven't reported this to Safari. Even if Safari fixes this, the older versions of Safari will still have this problem. So I think we should handle it in SES itself.

mhofman:
@erights we likely want to add an `ownKeys` trap to the `scopeProxyHandlerProperties` as @argp1 suggests, but instead of calling `Reflect`, it likely should just `return [];`

erights:
Both the `scopeProxyHandlerProperties` in strict-scope-terminator.js and the `scopeProxyHandlerProperties` in sloppy-globals-scope-terminator.js ?

erights:
Nevermind. I see that one includes the other. Proceeding...

erights:
See https://github.com/endojs/endo/pull/1651

PR #1651, opened 2023-06-23
fix(ses): better safari debugging

erights:
Fixes https://github.com/endojs/endo/issues/1490

See that issue for explanation.

@argp1 I cannot add you as an official reviewer, but please review anyway. Thanks.


erights:
Given the nature of the bug this fixes, I have no idea how to write an automated test for it.

mhofman:
@erights looks like you need to update the `all other handlers throw errors` test, which is now failing.

erights:
> @erights looks like you need to update the `all other handlers throw errors` test, which is now failing.

Done

mhofman (approved):
Looks good. I don't think we need more tests than this.

PR #1654, opened 2023-06-27
feat(memoize): safe memoization

erights:


packages/ses/docs/memoize.md
```diff
@@ -0,0 +1,151 @@
...
+
+By "contingent safety", we mean the safety guarantees that follow given that
+certain requirements are met. Before we examine these, let's first understand
+the non-contingent semantics of this code.
+
+## Base semantics
+
+Given a function `fn`, `memoize(fn)` returns `memoFn`,
+a memoizing form of that `fn`
+as a one-argument function of an `arg` that is a valid WeakMap key. When `memoFn(arg)` is
```
erights, line 36: Terrible runon sentence. Really awful. I will revise.

packages/ses/src/memoize.js
```diff
@@ -0,0 +1,30 @@
...
+ * Given a one-argument function `fn` or a WeakMap-key compatible
+ * argument `arg`, returns `memoFn`, a memoizing form of that function
+ * that will only call `fn(arg)` for a given `arg` the first time.
+ * The `memoFn` function remembers the result and reuses it on subsequent
+ * calls with the same `arg`.
+ *
+ * See memoize.md for the Memoization Safety properties of `memoize`.
+ * (TODO turn into link once there's a URL)
+ *
+ * @template {any} A=[any]
```
turadg, line 13: Simpler to have one param which is itself the function that takes one argument.

mhofman:
> Either the object always has the same identity for all `fn(arg)` calls, as with reproducibility, or it has a fresh identity per call, as with fresh allocation by the call. This weaker guarantee is "determinism", which we define to allow such always-fresh but otherwise equivalent.

> But even meeting all these requirements, the
resulting memoization is observable because it turns observably distinct identities into observably shared identities.

Oh this is a good call. I hadn't though about the memoization being observable if the return value is equivalent but newly allocated every time. That seem to indicate that a generic memoization feature would not only need to prevent side effects of the memoized function, but also capture which allocations are observable in the results, and replicate those for every cached result.

One tempting naive approach to accomplish this replication in our environment might be to marshal encode then decode the result for every memoized call, however because of promises you effectively have to apply a membrane-like transform to any promises within the result.

packages/ses/src/memoize.js
erights: That's what the code does. You mean the type param and function type?

erights:
I just realized that we don't export things from the 'ses' package, so I'm using repackage.sh to create a new `@endo/memoize` package. Seems awkward to create a package for a single export though. But I don't see any existing package that would be a natural home for `memoize`. Please let me know if you have an alternate suggestion. Thanks.

erights:
@kris I generated this `@endo/memoize` package using `repackage.sh memoize` and have made only minimal changes from there.
At https://github.com/endojs/endo/actions/runs/5393031850/jobs/9792198279?pr=1654#step:9:120 I'm getting a lint error about the undeclared global use of `harden`. The lint error also repros locally for me, but from `yarn lint` and with red IDE squiggles.

I don't remember what the configuration magic is to let lint know that this is in the SES language in which `harden` is already a recognized global. Whatever it is, I'm surprised it wasn't automatically set up by repackage.sh. I looked at `@endo/promise-kit`, which does use `harden` as an undeclared global without problem, to see if I could spot the relevant configuration difference. I did not. What am I missing?

erights:
Looking at `@endo/promise-kit` for a similar example that works, @kriskowal spotted the 
```js
/// <reference types="ses"/>
```
at the top of index.js but not is-promise.js. @kris suggested maybe that's how ambient types work, so I tried it. It works here too, so immediate problem solved.

However, this case makes me even more befuddled about what the static typing rules are. cc @turadg @michaelfig

packages/ses/src/memoize.js
erights: Hi @turadg , I tried to figure this out and could not. Could you provide an example? Thanks.

erights: First, I tried

```js
/**
 * @template {(arg: any) => any} F function type
 * @param {F} fn
 * @returns {F}
 */
  const memo = new WeakMap();

  /** @type {F} */
  const memoFn = arg => {
```
but got 
```
src/memoize.js:19:9 - error TS2322: Type '(arg: any) => any' is not assignable to type 'F'.
  '(arg: any) => any' is assignable to the constraint of type 'F', but 'F' could be instantiated with a different subtype of constraint '(arg: any) => any'.

19   const memoFn = arg => {
           ~~~~~~


Found 1 error in src/memoize.js:19
```

erights: Then I tried
```js
/**
 * @template {(arg: any) => any} F function type
 * @param {F} fn
 * @returns {F}
 */
  const memo = new WeakMap();

  /**
   * @param {Parameters<F>[0]} arg
   * @returns {ReturnType<F>}
   */
  const memoFn = arg => {
    ...
  return harden(memoFn);
};
```
and got
```
src/memoize.js:30:10 - error TS2322: Type '(arg: Parameters<F>[0]) => ReturnType<F>' is not assignable to type 'F'.
  '(arg: Parameters<F>[0]) => ReturnType<F>' is assignable to the constraint of type 'F', but 'F' could be instantiated with a different subtype of constraint '(arg: any) => any'.

30   return harden(memoFn);
            ~~~~~~~~~~~~~~

  src/memoize.js:30:10
    30   return harden(memoFn);
                ~~~~~~~~~~~~~~
    Did you mean to call this expression?


Found 1 error in src/memoize.js:30
```
IIUC the two error messages, they seem to have the same root cause. I often run into this problem, but I have previously had a compact capture to report. I would love to understand what to about this.

erights: cc @michaelfig

erights:
If my reviewers do not mind, I'd like to squash the current commits because their combination is still small and their separation is not really helpful.

packages/ses/src/memoize.js
michaelfig: To thwart these kind of "is assignable to the constraint of type 'F', but 'F' could be instantiated with..." errors, you need to do an explicit typecast.  The only syntax to do so in JSDoc is to use `/** @type {sometype} */ (expr)`.  The parentheses are *not* optional, and all together they form an expression with value `expr` and type `sometype`.

```js
  /**
   * @param {Parameters<F>[0]} arg
   * @returns {ReturnType<F>}
   */
  const memoFn = arg => {
    ...
  };
  harden(memoFn);
  // This cast beats the type system into submission so that it accepts memoFn really is type F. 
  return /** @type {F} */ (memoFn);
```

michaelfig:
> ```js
> /// <reference types="ses"/>
> ```

> @kris suggested maybe that's how ambient types work, so I tried it.

SES's types are not ambient, but that triple-slash-reference statement explicitly imports the contents of `ses/types.d.ts` (as named by SES's `packageJson.types` field).  You're getting the type declarations for the actual values it adds to globalThis (which are in the `declare global` section in `ses/types.d.ts`), and that is leaking into the rest of your sources as global state tracked by the TS compiler.

Ambient types would be leakage of the type namespace itself into other files, which doesn't happen with triple-slash-reference.

packages/ses/src/memoize.js
erights: Given that I would need a type case, would this change still be a net simplification?

packages/ses/docs/memoize.md
erights: Done

packages/ses/src/memoize.js
erights: @turadg @michaelfig , after the above two attempts that IMO made things worse, how do you feel about proceeding with the current type declarations?

packages/memoize/src/memoize.js
```diff
@@ -0,0 +1,28 @@
...
+ * argument `arg`, returns `memoFn`, a memoizing form of that function
+ * that will only call `fn(arg)` for a given `arg` the first time.
+ * The `memoFn` function remembers the result and reuses it on subsequent
+ * calls with the same `arg`.
+ *
+ * See memoize.md for the Memoization Safety properties of `memoize`.
+ * (TODO turn into link once there's a URL)
+ *
+ * @template {any} A=[any]
+ * @template {any} R=[any]
```
michaelfig, line 12: If you want to constrain the type of `A` to be a WeakMap-compatible key (and unconstrain R), you can do:
```suggestion
 * @template {{}} A
 * @template R
```

michaelfig (commented):
Looks good.  Now that I see what you're trying to do with the simpler typing in the `memoize` function, I suggest tweaking the template parameters to more directly express the requirements (weakmap key as first argument, no constraints on return value).

michaelfig (approved):

packages/memoize/src/memoize.js
erights: Done

PR #1655, opened 2023-06-28
feat(ses): anticipate iterator helpers

### LINKED ISSUE #1289: Anticipate new hidden intrinsics: Iteration helpers
As proposals advance in tc39, some will introduce new hidden intrinsics, starting with Iteration helpers. At the Sep 2022 tc39 meeting, we agreed on a compromise where the proposal provides code for discovering these new hidden intrinsics. The ses-shim would use this code in order to add these hidden intrinsics to the collection of intrinsics to be checked against the whitelist and hardened. This compromise is only needed until the getIntrinsics proposal lands, which will provide a unified mechanism for discover all intrinsics.

https://github.com/tc39/proposal-iterator-helpers/pull/235/files is an adequate approximation of this agreement. We need to modify it so it doesn't fail if run on a system in which these are not present. With that modification, we should adopt it into the ses-shim well ahead of anyone shipping an engine that implements the Iterator helpers proposal. 

(During the Google era of SES, we had a bad vulnerability do to failure to add code to discover the async generator intrinsics until after it was shipping. Fortunately, it was responsibly disclosed and we have no reason to believe it was ever maliciously exploited.)

Attn @michaelficarra @ljharb @kriskowal

erights:
Fixes #1289 

Anticipate both [Iterator Helpers](https://github.com/tc39/proposal-iterator-helpers) and [Async Iterator Helpers](https://github.com/tc39/proposal-async-iterator-helpers).

Note that Iterator Helpers are already as stage 3, so we can expect engines to start shipping these at any time.

Reviewers, please check it against the text of the proposals.

### DIFF
--- a/packages/ses/src/permits.js
+++ b/packages/ses/src/permits.js
@@ -76,6 +76,10 @@ export const universalPropertyNames = {
   URIError: 'URIError',
   WeakMap: 'WeakMap',
   WeakSet: 'WeakSet',
+  // https://github.com/tc39/proposal-iterator-helpers
+  Iterator: 'Iterator',
+  // https://github.com/tc39/proposal-async-iterator-helpers
+  AsyncIterator: 'AsyncIterator',
 
   // *** Other Properties of the Global Object
 
@@ -246,7 +250,7 @@ export const FunctionInstance = {
 };
 
 // AsyncFunction Instances
-const AsyncFunctionInstance = {
+export const AsyncFunctionInstance = {
   // This property is not mentioned in ECMA 262, but is present in V8 and
   // necessary for lockdown to succeed.
   '[[Proto]]': '%AsyncFunctionPrototype%',
@@ -1233,14 +1237,90 @@ export const permitted = {
 
   // *** Control Abstraction Objects
 
+  // https://github.com/tc39/proposal-iterator-helpers
+  Iterator: {
+    // Properties of the Iterator Constructor
+    '[[Proto]]': '%FunctionPrototype%',
+    prototype: '%IteratorPrototype%',
+    from: fn,
+  },
+
   '%IteratorPrototype%': {
     // The %IteratorPrototype% Object
     '@@iterator': fn,
+    // https://github.com/tc39/proposal-iterator-helpers
+    constructor: 'Iterator',
+    map: fn,
+    filter: fn,
+    take: fn,
+    drop: fn,
+    flatMap: fn,
+    reduce: fn,
+    toArray: fn,
+    forEach: fn,
+    some: fn,
+    every: fn,
+    find: fn,
+    '@@toStringTag': 'string',
+    // https://github.com/tc39/proposal-async-iterator-helpers
+    toAsync: fn,
+  },
+
+  // https://github.com/tc39/proposal-iterator-helpers
+  '%WrapForValidIteratorPrototype%': {
+    '[[Proto]]': '%IteratorPrototype%',
+    next: fn,
+    return: fn,
+  },
+
+  // https://github.com/tc39/proposal-iterator-helpers
+  '%IteratorHelperPrototype%': {
+    '[[Proto]]': '%IteratorPrototype%',
+    next: fn,
+    return: fn,
+    '@@toStringTag': 'string',
+  },
+
+  // https://github.com/tc39/proposal-async-iterator-helpers
+  AsyncIterator: {
+    // Properties of the Iterator Constructor
+    '[[Proto]]': '%FunctionPrototype%',
+    prototype: '%AsyncIteratorPrototype%',
+    from: fn,
   },
 
   '%AsyncIteratorPrototype%': {
     // The %AsyncIteratorPrototype% Object
     '@@asyncIterator': fn,
+    // https://github.com/tc39/proposal-async-iterator-helpers
+    constructor: 'AsyncIterator',
+    map: fn,
+    filter: fn,
+    take: fn,
+    drop: fn,
+    flatMap: fn,
+    reduce: fn,
+    toArray: fn,
+    forEach: fn,
+    some: fn,
+    every: fn,
+    find: fn,
+    '@@toStringTag': 'string',
+  },
+
+  // https://github.com/tc39/proposal-async-iterator-helpers
+  '%WrapForValidAsyncIteratorPrototype%': {
+    '[[Proto]]': '%AsyncIteratorPrototype%',
+    next: fn,
+    return: fn,
+  },
+
+  // https://github.com/tc39/proposal-async-iterator-helpers
+  '%AsyncIteratorHelperPrototype%': {
+    '[[Proto]]': 'Async%IteratorPrototype%',
+    next: fn,
+    return: fn,
+    '@@toStringTag': 'string',
   },
 
   '%InertGeneratorFunction%': {
--- a/packages/ses/test/test-anticipate-async-iterator-helpers-shimmed.js
+++ b/packages/ses/test/test-anticipate-async-iterator-helpers-shimmed.js
@@ -1,0 +1,38 @@
+// KLUDGE HAZARD The core-js shims are written as sloppy code
+// and so introduce sloppy functions.
+import 'core-js/actual/async-iterator/index.js';
+import test from 'ava';
+import '../index.js';
+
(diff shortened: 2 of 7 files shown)


packages/ses/src/permits.js
```diff
@@ -1233,14 +1237,90 @@ export const permitted = {
...
+    // https://github.com/tc39/proposal-async-iterator-helpers
+    toAsync: fn,
+  },
+
+  // https://github.com/tc39/proposal-iterator-helpers
+  '%WrapForValidIteratorPrototype%': {
+    '[[Proto]]': '%IteratorPrototype%',
+    next: fn,
+    return: fn,
+  },
```
erights, line 1274: Is it odd that this one doesn't have an `@@toStringTag` but the next one does? Likewise for the async cases below.

mhofman: I think the wrap iterator is still one expected to not be revealed? I'm not sure, going by memory here.

erights:
Earlier I wrote

> TODO Will remain draft until it actually fixes #1289 . I have not yet added the recognition code.
>
> TODO to test this ahead of engine release, we would need to use or adapt a shim such as https://github.com/zloirock/core-js#iterator-helpers and https://github.com/zloirock/core-js#asynciterator-helpers

Both are now done. This is Ready for Review.

packages/ses/src/permits.js
mhofman: I believe this is indeed and oversight, and posted about it on the I believe erroneously closed issue: https://github.com/tc39/proposal-iterator-helpers/issues/42#issuecomment-1622659282

packages/ses/test/test-anticipate-async-iterator-helpers-shimmed.js
```diff
@@ -0,0 +1,38 @@
+// KLUDGE HAZARD The core-js shims are written as sloppy code
+// and so introduce sloppy functions.
```
mhofman, line 2: Afaik, core-js is compatible with strict mode, or at least most parts of it are. I actually never thought about CommonJS support in ES Modules loaded by node as being able to introduce sloppy functions (besides through dynamic eval).

Wondering if we could force the tests to be run with a node CommonJS module loader that forces `'use strict'`

packages/ses/test/test-anticipate-async-iterator-helpers-shimmed.js
```diff
@@ -0,0 +1,38 @@
...
+import { AsyncFunctionInstance } from '../src/permits.js';
+
+AsyncFunctionInstance.arguments = {};
+AsyncFunctionInstance.caller = {};
+
+// Skipped because the core-js shim seems to miss the
+// actual %AsyncIteratorPrototype%,
+// so it creates a new one, causing us to fail because lockdown correctly
+// detects the conflicting definitions.
+// TODO report the bug to core-js
```
mhofman, line 21: Let's resolve this TODO before merging, and include a link to the issue?

packages/ses/test/test-anticipate-async-iterator-helpers-shimmed.js
```diff
@@ -0,0 +1,38 @@
...
+
+// KLUDGE HAZARD only for testing with the sloppy modules of the
+// core-js iterator shim.
+// We mutate the permits to tolerates the sloppy functions for testing
+// by sacrificing security. The caller and arguments properties of
+// sloppy functions violate ocap encapsulation rules.
+import { AsyncFunctionInstance } from '../src/permits.js';
+
+AsyncFunctionInstance.arguments = {};
+AsyncFunctionInstance.caller = {};
```
mhofman, line 15: Wait, why do we have a sloppy `AsyncFunctionInstance`?

kriskowal (approved):

packages/ses/test/test-anticipate-async-iterator-helpers-shimmed.js
erights: > Wondering if we could force the tests to be run with a node CommonJS module loader that forces 'use strict'

@kriskowal Is there such a thing?

packages/ses/src/permits.js
erights: Thanks! Resolving here.

packages/ses/test/test-anticipate-async-iterator-helpers-shimmed.js
erights: (Deleted my comment as confused. I'll investigate)

erights: Trying this out on `test-anticipate-iterator-helpers-shimmed.js`. Commenting out the monkey patching of the corresponding two items from the permits list, 
```js
// FunctionInstance.arguments = {};
// FunctionInstance.caller = {};
```
and rerunning the test, I get
```
Removing intrinsics.Iterator.from.arguments
  ✘ [fail]: shimmed iterator helpers Error thrown in test
failed to delete intrinsics.Iterator.from.arguments (TypeError#1)
TypeError#1: Cannot delete property 'arguments' of function from(O) {
    var iteratorRecord = getIteratorFlattenable(typeof O == 'string' ? toObject(O) : O);
   ...<omitted>... }
    var iteratorRecord = getIteratorFlattenable(typeof O == 'string' ? toObject(O) : O);
   ...<omitted>... }
  at visitProperties (file:///Users/markmiller/src/ongithub/endojs/endo/packages/ses/src/permits-intrinsics.js:289:22)
```

What happens is that core-js adds `Iterator.from` as you'd expect. But the function it adds is a sloppy function that, on v8/node, has *undeletable* `arguments` and `caller` properties, which is not surprising. Our permit processing correctly fails safe when it finds it cannot delete them. Fortunately, the value of these properties at the time of permit processing is null.

By adding these two properties to the permits record that applies to most functions, we allow most functions in general to have these extra properties, including the ones added by the core-js shim.

packages/ses/test/test-anticipate-async-iterator-helpers-shimmed.js
erights: Filed https://github.com/endojs/endo/issues/1669 on this. If this PR becomes otherwise ready to merge, I'll let that issue speak for this conversation. So resolving here.

packages/ses/test/test-anticipate-async-iterator-helpers-shimmed.js
erights: Done! At https://github.com/zloirock/core-js/issues/1267

packages/ses/test/test-anticipate-async-iterator-helpers-shimmed.js
erights: Resolving

packages/ses/test/test-anticipate-async-iterator-helpers-shimmed.js
naugtur: Context on core-js and its single maintainer https://github.com/zloirock/core-js/blob/master/docs/2023-02-14-so-whats-next.md 
TL;DR: all attempts to sustain himself while working on core-js full time have been unsuccessful. He might stop at any point and he's in trouble. 
I checked and he does seem active, so it's possible the last attempt proposed in the doc is working for now.

packages/ses/test/test-anticipate-iterator-helpers-shimmed.js
```diff
@@ -0,0 +1,42 @@
...
+// and so introduce sloppy functions.
+import 'core-js/actual/iterator/index.js';
+import test from 'ava';
+import '../index.js';
+
+// KLUDGE HAZARD only for testing with the sloppy modules of the
+// core-js iterator shim.
+// We mutate the permits to tolerates the sloppy functions for testing
+// by sacrificing security. The caller and arguments properties of
+// sloppy functions violate ocap encapsulation rules.
```
zloirock, line 11: `core-js` use strict mode, but not in all cases - for example, in cases where `this` depends on it. `.arguments` and `.caller` are also interesting cases, I didn't think that it's interesting for anyone. I can consider adding the strict mode in all affected cases. You could open an issue in the `core-js` repo.

erights: https://github.com/zloirock/core-js/issues/1269

Thanks!

PR #1660, opened 2023-06-29
refactor(daemon): Rename provide to endow

kriskowal:
This change toys with a different name for the caplet and plugin calling convention. Where before, we were experimenting with:

```js
export const provide0 = async powers => Far('Caplet', {});
```

we are now trying out:

```js
export const endow = async powers => Far('Caplet', {});
```

Consider that we’re gradually reaching out to host four kinds of program:

- a web app caplet (weblet)
- a worker caplet (worklet)
- a CLI command caplet (runlet)
- a plugin (which is not a caplet because it is not confined)

I’m evaluating this change against the alternative `make`, in the expectation that a “runlet” would use `main` and would not make a public API object, since it is disposable.

```js
export const main = async powers => {}
```

The verb `endow` makes sense for the others (weblet, worklet, and plugin), but `make` doesn’t fit weblet, since its return value is also disposable.

### DIFF
--- a/packages/cli/src/endo.js
+++ b/packages/cli/src/endo.js
@@ -809,9 +809,9 @@ export const main = async rawArgs => {
     });
 
   program
-    .command('import-unsafe0 <path> <outbox>')
+    .command('import-unsafe <path> <guest>')
     .description(
-      'imports the module at the given path and runs its provide0 function with all of your authority',
+      'imports the module at the given path and runs its endow function with all of your authority',
     )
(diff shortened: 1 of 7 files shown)


dckc:
the name seems a little backwards: to my mind, endow is what the caller does, not what the confined function does.

kriskowal:
I’ve come to the same conclusion and I’m toying with ‘make’ now, to go with
‘endo make’. Also ‘main’ to go with ‘endo run’

PR #1661, opened 2023-06-29
Make `AsyncLocalStorage` compatible with `@endo/pass-style`

### LINKED ISSUE #1657: `AsyncLocalStorage` from `node:async_hooks` is broken
@mhofman, yet another instance of random async hooks symbols causing problems under SES.

I'm working with a library that is advertised to use `AsyncLocalStorage` under the covers, and it throws:

```
Error#1: Object [Promise] {
  [Symbol(kResourceStore)]: BaseContext {
    _currentContext: Map(0) {},
    getValue: {},
    setValue: {},
    deleteValue: {}
  }
} - Must not have any own properties: [ Symbol(kResourceStore) ]

  at Array.every (<anonymous>)
  at deeplyFulfilledObject (packages/internal/src/utils.js:186:10)
```

I hope we can also patch this in `node-async_hooks-patch.js`.  I have a workaround for now.

mhofman:
Fixes #1657 

Monkey patch the `AsyncLocalStorage` implementation to rewrite the handling of stores, replacing the use of symbol expando properties on promises with a WeakMap per instance.

`@endo/pass-style` and more specifically the `safePromise` check asserts that no extra own keys exist on promise instances. Furthermore hardening a promise would also harden any store object added as expando to promise instances.

### DIFF
--- a/packages/init/src/node-async-local-storage-patch.js
+++ b/packages/init/src/node-async-local-storage-patch.js
@@ -1,0 +1,95 @@
+/* eslint-disable no-underscore-dangle */
+
+import { AsyncLocalStorage, executionAsyncResource } from 'node:async_hooks';
+
+const { is: ObjectIs } = Object;
+const { apply: ReflectApply } = Reflect;
+
+/** @type {WeakMap<AsyncLocalStorageInternal, WeakMap>} */
+const resourceStoreMaps = new WeakMap();
+
+/** @type {(key: AsyncLocalStorageInternal) => WeakMap} */
+const getStoreMap = resourceStoreMaps.get.bind(resourceStoreMaps);
+
+/**
+ * @typedef {object} AsyncLocalStorageInternal
+ * @property {boolean} enabled
+ * @property {typeof _propagate} _propagate
+ * @property {(this: AsyncLocalStorage) => void} _enable
+ */
+
+Object.defineProperty(AsyncLocalStorage.prototype, 'kResourceStore', {
+  configurable: true,
+  /**
+   * @this {AsyncLocalStorage & AsyncLocalStorageInternal}
+   */
+  set() {
+    resourceStoreMaps.set(this, new WeakMap());
+  },
+});
+
+/**
+ * @this {AsyncLocalStorage & AsyncLocalStorageInternal}
+ * @param {object} resource
+ * @param {object} triggerResource
+ * @param {string} [type]
+ */
+function _propagate(resource, triggerResource, type) {
+  if (!this.enabled) return;
+
+  const storeMap = getStoreMap(this);
+  storeMap.set(resource, storeMap.get(triggerResource));
+}
+
+// @ts-expect-error propagate is internal
+AsyncLocalStorage.prototype._propagate = _propagate;
+
+/**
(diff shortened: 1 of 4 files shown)


mhofman:
cc @erights

michaelfig (approved):
Nicely done!  I don't fully comprehend, but don't see any problems with the implementation.

mhofman:
Actually thought of a way to avoid the lazy init. Updated!

packages/init/src/node-async-local-storage-patch.js
```diff
@@ -0,0 +1,90 @@
+/* eslint-disable no-underscore-dangle */
+
+import { AsyncLocalStorage, executionAsyncResource } from 'node:async_hooks';
+
+const { is: ObjectIs } = Object;
+const { apply: ReflectApply } = Reflect;
```
kriskowal, line 6: Precedent for `objectIs` and `reflectApply` in `ses`’s `commons.js`.

kriskowal, line 6: Precedent for `objectIs` and `reflectApply` in `ses`’s `commons.js`.

kriskowal (approved):

packages/init/src/node-async-local-storage-patch.js
mhofman: Ack, but I wanted to keep the Node style for these intrinsics. Please see the original implementation this is monkey patching.

erights:
I'm not sure I understand the title of the PR. What's the coupling to `pass-style` specifically? `safe-promise.js`? That makes sense, but if true, should be mentioned somewhere.

mhofman:
https://github.com/endojs/endo/issues/1657 describes the issue, but I updated the description of the PR to be more specific.

packages/init/src/node-async-local-storage-patch.js
```diff
@@ -0,0 +1,90 @@
...
+ * @param {object} resource
+ * @param {object} triggerResource
+ * @param {string} [type]
+ */
+function _propagate(resource, triggerResource, type) {
+  if (!this.enabled) return;
+
+  const storeMap = this._resourceStoreMap;
+  storeMap.set(resource, storeMap.get(triggerResource));
+}
```
gibson042, line 37: Might as well follow our normal lexical conventions rather than exposing a new property.
```suggestion
const resourceStoreMaps = new WeakMap();

Object.defineProperty(AsyncLocalStorage.prototype, 'kResourceStore', {
  configurable: true,
  /**
   * @this {AsyncLocalStorage & AsyncLocalStorageInternal}
   */
  set() {
    resourceStoreMaps.set(this, new WeakMap());
  },
});

/**
 * @this {AsyncLocalStorage & AsyncLocalStorageInternal}
 * @param {object} resource
 * @param {object} triggerResource
 * @param {string} [type]
 */
function _propagate(resource, triggerResource, type) {
  if (!this.enabled) return;

  const storeMap = resourceStoreMaps.get(this);
  if (storeMap) storeMap.set(resource, storeMap.get(triggerResource));
}
```
etc.

gibson042 (approved):

PR #1662, opened 2023-06-29
refactor: convert some forEach to for/of loops

erights:
Inspired by https://github.com/Agoric/agoric-sdk/pull/7983 to try to catch and fix most of these by regexp. I thought it would be quick, but it was not. Need more automation (or maybe better regexps?) before I would recommend trying on agoric-sdk

As part of my semi-automated approach I also ran `yarn format` and `yarn lint-fix`. Since we have stopped running `yarn lint-fix`, it also made some other improvements that it could have made earlier. The one I spotted is removing default values from ts-in-jsdoc-comment optional parameter declaration notation.

### DIFF
--- a/packages/static-module-record/test/fixtures/large.js
+++ b/packages/static-module-record/test/fixtures/large.js
@@ -125,9 +125,9 @@ function makeModulePlugins(options) {
      * @param {string} name - the local name of the exported variable.
      */
     const markLiveExport = name => {
-      topLevelExported[name].forEach(importTo => {
+      for (const importTo of (topLevelExported[name])) {
         liveExportMap[importTo] = [name, true];
-      });
+      }
       return hLiveId;
     };
 
@@ -143,9 +143,9 @@ function makeModulePlugins(options) {
      * @param {string} name - the local name of the exported variable.
      */
     const markFixedExport = name => {
-      topLevelExported[name].forEach(importTo => {
+      for (const importTo of (topLevelExported[name])) {
         fixedExportMap[importTo] = [name];
-      });
+      }
       return hOnceId;
     };
 
@@ -366,7 +366,7 @@ function makeModulePlugins(options) {
           if (!specs) {
             return;
           }
-          specs.forEach(spec => {
+          for (const spec of specs) {
             const importTo = spec.local.name;
             importDecls.push(importTo);
             let importFrom;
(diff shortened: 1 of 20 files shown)


packages/static-module-record/test/fixtures/large.js
```diff
@@ -366,7 +366,7 @@ function makeModulePlugins(options) {
           if (!specs) {
             return;
           }
-          specs.forEach(spec => {
+          for (const spec of (specs)) {
```
kriskowal, line 369: seemingly superfluous paren artifact
```suggestion
          for (const spec of specs) {
```

packages/static-module-record/test/fixtures/large.js
```diff
@@ -511,9 +511,9 @@ function makeModulePlugins(options) {
           collectPatternIdentifiers(path, id),
         );
         if (doAnalyze) {
-          vids.forEach(({ name }) => {
+          for (const { name } of (vids)) {
```
kriskowal, line 514:
```suggestion
          for (const { name } of vids) {
```

packages/static-module-record/test/fixtures/large.js
```diff
@@ -570,17 +570,17 @@ function makeModulePlugins(options) {
             const vids = declarations.flatMap(({ id }) =>
               collectPatternIdentifiers(path, id),
             );
-            vids.forEach(({ name }) => {
+            for (const { name } of (vids)) {
```
kriskowal, line 573: etc

packages/static-module-record/test/fixtures/large.js
```diff
@@ -570,17 +570,17 @@ function makeModulePlugins(options) {
...
                 tle = [];
                 topLevelExported[name] = tle;
               }
               tle.push(name);
-            });
+            }
           }
 
-          specs.forEach(spec => {
+          for (const spec of (specs)) {
```
kriskowal, line 583: etc

kriskowal (approved):

packages/static-module-record/test/fixtures/large.js
erights: Using the regexp 
```
 of \((\w+)\)\) \{
```
I think the four you found were all of them. In any case, done.

PR #1663, opened 2023-06-30
style: warning free lint

erights:


turadg (approved):
This is a "style:" change that doesn't change behavior.

PR #1664, opened 2023-06-30
fix: stub censored properties

### LINKED ISSUE #910: {[inert function]} from toString() for Math.random etc.
In a discussion of API evolution, @erights noted that for new features, testing for new properties by name works well, but for retroactive changes such as making `Math.random()` inert, it's unrealistic to expect code to test for absence of a property.

 - [ ] code changes, unit tests as usual
 - [ ] 262 test

_IOU more explanation, but I have a meeting..._

cc @phoddie

erights:
Fixes #910 

Staged on https://github.com/endojs/endo/pull/1677

@phoddie I'm starting with a more severe behavior: Making the replacement property be a poisoned access instead. What do you think of this approach? I like this better than a throwing method because the censored property may not be a function-valued property.

### DIFF
--- a/packages/ses/src/permits-intrinsics.js
+++ b/packages/ses/src/permits-intrinsics.js
@@ -55,6 +55,7 @@ import {
   entries,
   getOwnPropertyDescriptor,
   getPrototypeOf,
+  defineProperty,
   isObject,
   mapGet,
   objectHasOwnProperty,
@@ -248,6 +249,68 @@ export default function whitelistIntrinsics(
     return undefined;
   }
 
+  const removeProperty = (path, obj, prop, permit) => {
+    // Either the object lacks a permit or the object doesn't match the
+    // permit.
+    // If the permit is specifically false, not merely undefined,
+    // this is a property we expect to see because we know it exists in
+    // some environments and we have expressly decided to exclude it.
+    // Any other disallowed property is one we have not audited and we log
+    // that we are removing it so we know to look into it, as happens when
+    // the language evolves new features to existing intrinsics.
+    if (permit !== false) {
+      // This call to `console.warn` is intentional. It is not a vestige of
+      // a debugging attempt. See the comment at top of file for an
+      // explanation.
+      // eslint-disable-next-line @endo/no-polymorphic-call
+      console.warn(`Removing ${path}`);
+    }
+    try {
+      defineProperty(obj, prop, {
+        get() {
+          throw TypeError(`property ${path} removed from Hardened JS`);
+        },
+        set(_) {
+          throw TypeError(`property ${path} removed  from Hardened JS`);
+        },
+        // Whether or not the original property was enumerable,
+        // always make poisoned properties non-enumerable.
+        // Thus, poisoned properties invisible to casual discovery by reflective
+        // operations that only see enumerable properties,
+        // like `Object.assign`, `Object.entries`, or `...`.
+        // These are the only reflective operations that also
+        // GET the values generically, which would have broken on poisoned
+        // properties.
+        //
+        // The reflective operations that see non-enumerable properties,
+        // like `Object.getOwnPropertyDescriptors`, generally only get
+        // the property descriptor, which obtains an accessor property's
+        // getter rather than calling it.
+        //
+        // See https://github.com/endojs/endo/blob/master/packages/pass-style/doc/copyRecord-guarantees.md#how-do-i-enumerate-thee-let-me-list-the-ways
+        enumerable: false,
+      });
+    } catch (err) {
+      if (prop in obj) {
+        if (typeof obj === 'function' && prop === 'prototype') {
+          obj.prototype = undefined;
+          if (obj.prototype === undefined) {
+            // eslint-disable-next-line @endo/no-polymorphic-call
+            console.warn(`Tolerating undeletable ${path} === undefined`);
+            // eslint-disable-next-line no-continue
+            return;
+          }
+        }
+        // eslint-disable-next-line @endo/no-polymorphic-call
+        console.error(`failed to delete ${path}`, err);
+      } else {
+        // eslint-disable-next-line @endo/no-polymorphic-call
+        console.error(`deleting ${path} threw`, err);
+      }
+      throw err;
+    }
+  };
+
   /*
    * visitProperties()
    * Visit all properties for a permit.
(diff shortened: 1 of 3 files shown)


phoddie:
It is more severe! Seems fine. I understand that this is more developer-friendly for non-function-valued properties. For function-valued properties, is there a benefit to this approach over the previous one?

erights:
> It is more severe! Seems fine. I understand that this is more developer-friendly for non-function-valued properties. For function-valued properties, is there a benefit to this approach over the previous one?

Only two things:
   * consistency with non-function valued properties, if we agree that this severe approach is right for those.
   * sometimes code picks up a function to call later. This detects the error as soon as possible.

OTOH this level of severity may break that reflectively discovers all the properties of an object, like `Object.assign`. To mitigate this I made the poisoned properties non-enumerable. See the extended comment in the code about why this may work surprisingly well.

packages/ses/src/permits-intrinsics.js
```diff
@@ -248,6 +249,68 @@ export default function whitelistIntrinsics(
...
+      // This call to `console.warn` is intentional. It is not a vestige of
+      // a debugging attempt. See the comment at top of file for an
+      // explanation.
+      // eslint-disable-next-line @endo/no-polymorphic-call
+      console.warn(`Removing ${path}`);
+    }
+    try {
+      defineProperty(obj, prop, {
+        get() {
+          throw TypeError(`property ${path} removed from Hardened JS`);
```
kriskowal, line 271: This might cause code that tests for the existence of the property, in order to provide an alternate behavior or ponyfill, to throw instead of proceeding. Have we accommodated these cases elsewhere?

mhofman: That's a very good point! The usual pattern is to do something like `Foo.bar = Foo.bar || polyfilledFooBar`, which would throw. I'm having a hard time imagining a case where we want to throw on that statement.

erights: This PR is based on the insight that the property testing style
`Foo.bar = Foo.bar || polyfilledFooBar`
is useful in one time direction --- when `Foo.bar` is added to a system that did not previously contain `Foo.bar`. In the other time direction, it is not useful. If `Foo.bar` was universally present in the relevant past, then we should assume that old code that uses it does so unconditionally, since it had no reason to expect its absence. If a new transition (such as Hardened JavaScript) censors it, it is less useful to censor it by removing it, since the old code that breaks will break with a harder to diagnose error. Instead, since the relevant old code uses it unconditionally, providing a more informative diagnostic is more valuable to let someone know *why* the old code is broken in the new environment.

PTAL

kriskowal: I’m more interested in the ponyfill pattern, which would also throw:

```js
export const fooBar = Foo.bar || virtualFooBar;
```

Are we censoring properties that could be ponyfilled?

erights: To answer this question, I just looked at occurrences of `: false,` in https://github.com/endojs/endo/blob/master/packages/ses/src/permits.js . This convinced me that this PR in its current state is ill-conceived, in that it replaces all of these with throwing accessors, whereas some should still be censored by removal instead. There are two reasons why removal would be better for some:
   * The censored property should be removed because it shadows a property that should instead be available, such as v8's `*Error.prototype.toString`, for all the subclasses of `Error`, where this extra non-standard `toString` property uselessly shadows the standard `Error.prototype.toString` https://github.com/endojs/endo/blob/db35cc7cbeb90ddd8fe56f4c526aac60209a8ab8/packages/ses/src/permits.js#L302 . Removal would stop the silly shadowing. Censoring by accessor just breaks things. BAD. This PR should also add a test ensuring that normal use is not broken in that way.
   * The censored property was added by a sufficiently recent version of the ecma262 standard that the property testing style still makes sense.

To do this, the permits.js file will need an additional marker to distinguish how the property should be censored. At least removal vs throwing accessor. As long as we're going that far, would could make it a three-way distinction, to censor some method data properties with throwing-method data property as XS currently does.

Once we make any of these distinctions, there remains the additional matter of how to censor unexpected properties, i.e., properties not mentioned in permits.js. I think it is clear that if it is unexpected, then uses of it should be the property testing pattern anyway, so removal is appropriate, rather than a throwing accessor or a throwing method.

@phoddie , part of the point of this PR is to align the SES shim better with the XS implementation of Hardened JS. Does the above reasoning make sense for XS? If so, what would remain is to coordinate agreement on which `permits.js` properties should go into which category.

In light of this, I'm putting this PR into Draft until we agree on how to proceed.

phoddie: > @phoddie , part of the point of this PR is to align the SES shim better with the XS implementation of Hardened JS. Does the above reasoning make sense for XS? If so, what would remain is to coordinate agreement on which permits.js properties should go into which category.

For Hardened JavaScript, XS replaces `Date.now()` and `Math.random()` with stubs that always throw. It also uses a version of the Date constructor that throws in cases that would provide the current time. We consider all of that to be specified by Hardened JavaScript and so it is implemented in XS without consulting an external file (e.g. `permits.js`) at runtime.

The behavior of `Date.now()` and `Math.random()` is consistent with what XS does when stripping unused functions and methods to minimize the engine for embedded builds. Any stripped function or method is replaced with a function that always throws. XS does not strip unused data properties (e.g. `Math.PI`) since there is generally no size benefit.

So, if I understand the question correctly, there might not be anything for XS to do for `permits.js` unless we eventually need to migrate the handling of `permits.js` into XS. But... I think that can be handled by scripts prior to lockdown.

erights: Do you poison anything other than `Date.now()` and `Math.random()`? If not, I may replace this PR with a much simpler one ;)

phoddie: > Do you poison anything other than Date.now() and Math.random()? 

No. (Apart from the note on the Date constructor above)

I do want to be careful about who is leading here. XS installs stubs that throw for Date.now() and Math.random() to follow your specification of Hardened JavaScript. That behavior is the result of discussions we had a together a while back.

erights: Ok, I just looked at all 45 occurrences of `: false,` at https://github.com/endojs/endo/blob/master/packages/ses/src/permits.js . I think I'm happy to leave all of these as censored by removal.

As I mention above, I'm also happy that all unexpected properties, i.e., those not mentioned by permits.js, be censored by removal. Thus, I am closing this PR and will start a separate one just to align on the tiny number of properties we do want to censor by poisoning. I'm going to do this manually in the repairs, rather than generically by extending the permits processing system.

Everyone, Please do look at these 45 occurrences and let me know if I'm missing something. Should all these remain censored by removal?

erights:
Closing for reasons stated above at https://github.com/endojs/endo/pull/1664#discussion_r1289099470

packages/ses/src/permits-intrinsics.js
erights: Discussion continuing at https://github.com/endojs/endo/issues/910#issuecomment-1672369082

PR #1666, opened 2023-07-04
feat(exo): revocables

erights:
Background: Exos already have a level of indirection between the exposed method with the protective method guard vs the raw method. In order to enable prototype-based method sharing of the protective methods, they lookup the context from their `this`, which is the exo object itself. The context is then passed as the `this` to the raw method. Having paid for this level of indirection, we already use it for many benefits, like enforcing that the exposed methods cannot be applied to unrelated instances.

An additional benefit we can obtain from this indirection is revocation. However, creating a revocable exo is the odd case. We don't want the normal case to pay an undue notational cost to accommodate the odd case, even if that raises the notational cost of the odd case a bit. We do so by allowing an additional `getRevoker` option, which is called with a `revoke` function to be called with the exo to be revoked as an argument. 

This seems unduly higher order. But the `getRevoker` option is directly analogous to the `executor` argument of the promise constructor, which is called with a `resolve` function to be called with a resolution.

This is motivated by the benefits of direct support for revocation for exos at the liveslots level, which needs a parallel implementation of exactly the same options with the same observable semantics. At the liveslots level, the benefits would be
   * A step towards dropping exports of used-up exo objects such as payments, relieving distributed gc pressure.
   * A step towards Zoe2 use objects that can be turned (back) into transferable invitations.

Once this support is uniform across heap exos (from this PR) and virtual and durable exos ( https://github.com/Agoric/agoric-sdk/pull/8020 , https://github.com/Agoric/agoric-sdk/pull/8008 ), then perhaps zones can mirror this universal support. Further, zones themselves could conceivably absorb the revocation support, to provide batch revocation according to creating zone.

### DIFF
--- a/packages/exo/src/exo-makers.js
+++ b/packages/exo/src/exo-makers.js
@@ -5,7 +5,7 @@ import { objectMap } from '@endo/patterns';
 
 import { defendPrototype, defendPrototypeKit } from './exo-tools.js';
 
-const { create, seal, freeze, defineProperty } = Object;
+const { create, seal, freeze, defineProperty, values } = Object;
 
 const { getEnvironmentOption } = makeEnvironmentCaptor(globalThis);
 const DEBUG = getEnvironmentOption('DEBUG', '');
@@ -63,10 +63,23 @@ export const initEmpty = () => emptyRecord;
  */
 
 /**
+ * @callback Revoker
+ * @param {any} exo
+ * @returns {boolean}
+ */
+
+/**
+ * @callback ReceiveRevoker
+ * @param {Revoker} revoke
+ * @returns {void}
+ */
+
+/**
  * @template C
  * @typedef {object} FarClassOptions
  * @property {(context: C) => void} [finish]
  * @property {StateShape} [stateShape]
+ * @property {ReceiveRevoker} [receiveRevoker]
  */
 
 /**
@@ -79,9 +92,15 @@ export const initEmpty = () => emptyRecord;
  * @param {FarClassOptions<ClassContext<ReturnType<I>, M>>} [options]
  * @returns {(...args: Parameters<I>) => (M & import('@endo/eventual-send').RemotableBrand<{}, M>)}
  */
-export const defineExoClass = (tag, interfaceGuard, init, methods, options) => {
+export const defineExoClass = (
+  tag,
+  interfaceGuard,
+  init,
+  methods,
+  options = {},
+) => {
   harden(methods);
-  const { finish = undefined } = options || {};
+  const { finish = undefined, receiveRevoker = undefined } = options;
   /** @type {WeakMap<M,ClassContext<ReturnType<I>, M>>} */
   const contextMap = new WeakMap();
   const proto = defendPrototype(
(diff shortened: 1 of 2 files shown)


packages/exo/src/exo-makers.js
```diff
@@ -62,11 +62,28 @@ export const initEmpty = () => emptyRecord;
...
+ * @returns {boolean}
+ */
+
+/**
+ * @typedef {Record<FacetName, Revoker>} RevokerKit
+ */
+
+/**
+ * @callback GetRevoker
+ * @param {Revoker | RevokerKit} revoke
```
michaelfig, line 77: Maybe consider always returning a `RevokerKit`, potentially with just a single entry for `self`, so that it parallels the `this.facets` record?

I think that would be worthwhile to avoid needing to complicate this parameter's type.

michaelfig (commented):
Looking good so far.

packages/exo/src/exo-makers.js
erights: I'm gonna try a different strategy: Just have a one argument `revoke` function for both cases. For the kit case, the revoke function has to do a trivial search over all the facets. 
   * Downside: This does decrease the separation of powers: Rather than needing both the instance and a facet-specific revoke function to revoke, all you'd need is the instance and a class-kit-specific revoke function that covers all the facets.
   * Upside: The API is simpler both statically and dynamically.
   * Upside: While working on https://github.com/Agoric/agoric-sdk/pull/8008 , I discovered that virtual storage is organized differently, with a single cache map across all facets, with facet cross-talk suppressed by other means. This would make my original API harder to implement than the one I'm suggesting here.

Please let me know what you think.

erights: Even more radical simplification at https://github.com/endojs/endo/pull/1668

erights:
Closing in favor of https://github.com/endojs/endo/pull/1668

erights:
Reopening. Having tried to use the API of #1668, I now think this one may be the better approach. But also leaving #1668 open while investigating.

erights:
@michaelfig @FUDCo @warner , now that Chip is back, we should talk about this. In particular, the major design choice: should the unit of revocation be the facet or the cohort of facets?

erights:
@FUDCo and I discussed the main controversy, and decided that revoking a facet revokes just the facet, not the cohort. This PR does not yet reflect that decision.

erights:
> @FUDCo and I discussed the main controversy, and decided that revoking a facet revokes just the facet, not the cohort. This PR does not yet reflect that decision.

It does now.

packages/exo/src/exo-makers.js
erights: After talking to both @michaelfig and @FUDCo , the design we settled on is a `receiveRevoker` function (just a renaming of the confusingly named `getRevoker` function discussed above) that receives a `revoke` function with a single exo argument. When called with an exo instance made by this class, that instance is revoked. When called with an exo facet make by a class kit, only that facet is revoked.

I know of no remaining controversy about this design, so I'm resolving this conversation as Done.

erights:
Reviewers, ping. 

@kriskowal is gearing up to make a new endo release. I'd like to get this in there, because agoric-sdk needs to be able to depend on it in order to proceed with the work there to support the same feature for virtual/durable exos. Thanks.

packages/exo/src/exo-makers.js
```diff
@@ -62,11 +62,24 @@ export const initEmpty = () => emptyRecord;
  * Each property is distinct, is checked and changed separately.
  */
 
+/**
+ * @callback Revoker
+ * @param {object} exo
```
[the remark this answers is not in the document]
erights: I went with `any`. Done.

PR #1670, opened 2023-07-06
fix(ses): fixup async iterator helpers

### LINKED ISSUE #1289: Anticipate new hidden intrinsics: Iteration helpers

### LINKED ISSUE #1669: How to run .cjs shims strict?
To write tests for https://github.com/endojs/endo/pull/1655 , I import some shims from core-js. However, core-js is written in common lisp style modules, which by default are evaluated as sloppy code. As a result, even if the shims in question were otherwise safe and compatible with Hardened JS, this aspect makes them grossly unsafe. Fortunately, the SES shim by default will fail to initialize if it encounters apparently primordial sloppy functions. This fails safe.

At https://github.com/endojs/endo/pull/1655#discussion_r1253755342 @mhofman suggests that we run these as strict code. Presumably, to support LavaMoat, we already know how to do that. How should I modify the test cases from that PR?

mhofman:
#1655 introduced support for iterator helpers, but due to a limitation in the core-js shim usage, the async iterator helper test was disabled, and the PR missed a couple errors in the implementation of permits for the async iterator helpers.

This PR fixes these errors, and addresses the testing limitations so that the iterator helpers tests (both sync and async) can behave more like normal tests.

Closes #1669 
Closes #1289

### DIFF
--- a/packages/ses/src/get-anonymous-intrinsics.js
+++ b/packages/ses/src/get-anonymous-intrinsics.js
@@ -146,7 +146,7 @@ export const getAnonymousIntrinsics = () => {
     );
   }
 
-  if (globalThis.AsyncInterator) {
+  if (globalThis.AsyncIterator) {
     intrinsics['%AsyncIteratorHelperPrototype%'] = getPrototypeOf(
       // eslint-disable-next-line @endo/no-polymorphic-call
       globalThis.AsyncIterator.from([]).take(0),
--- a/packages/ses/src/permits.js
+++ b/packages/ses/src/permits.js
@@ -1317,7 +1317,7 @@ export const permitted = {
 
   // https://github.com/tc39/proposal-async-iterator-helpers
   '%AsyncIteratorHelperPrototype%': {
-    '[[Proto]]': 'Async%IteratorPrototype%',
+    '[[Proto]]': '%AsyncIteratorPrototype%',
     next: fn,
     return: fn,
     '@@toStringTag': 'string',
--- a/packages/ses/test/enforce-cjs-strict.js
+++ b/packages/ses/test/enforce-cjs-strict.js
@@ -1,0 +1,3 @@
+import module from 'node:module';
+
+module.wrapper[0] += '"use strict";';
--- a/packages/ses/test/core-js-configuration.js
+++ b/packages/ses/test/core-js-configuration.js
@@ -1,0 +1,3 @@
+import configurator from 'core-js/configurator.js';
+
+configurator({ USE_FUNCTION_CONSTRUCTOR: true });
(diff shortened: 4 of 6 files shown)


packages/ses/src/get-anonymous-intrinsics.js
```diff
@@ -146,7 +146,7 @@ export const getAnonymousIntrinsics = () => {
     );
   }
 
-  if (globalThis.AsyncInterator) {
+  if (globalThis.AsyncIterator) {
```
erights, line 149: Oops! Thanks for catching this.

packages/ses/src/permits.js
```diff
@@ -1317,7 +1317,7 @@ export const permitted = {
 
   // https://github.com/tc39/proposal-async-iterator-helpers
   '%AsyncIteratorHelperPrototype%': {
-    '[[Proto]]': 'Async%IteratorPrototype%',
+    '[[Proto]]': '%AsyncIteratorPrototype%',
```
erights, line 1320: Oops! Thanks for catching this too.

packages/ses/test/enforce-cjs-strict.js
```diff
@@ -0,0 +1,3 @@
+import module from 'node:module';
+
+module.wrapper[0] += '"use strict";';
```
erights, line 3: Frankly, much less painful than I feared ;)

erights (approved):
LGTM THANKS!

mhofman:
> Oops! Thanks for catching this.

Well the tests caught it, but I have to admit, it was hard to diagnose because ava caught the exception and didn't print out the details (I assume because we're not using ses-ava here)

PR #1674, opened 2023-07-11
stricter TypeScript checks

turadg:
Enables stricter TypesScript config. Punts on `noImplicitAny` and `useUnknownInCatchVariables` (so they can continue to be any). Before adding `useUnknownInCatchVariables` I fixed some cases.

Removes some `= undefined` so the checker doesn't get a faulty type hint.
**before**
Screenshot 2023-07-11 at 11 52 21 AM
**after**
Screenshot 2023-07-11 at 11 55 27 AM

Suppresses errors I didn't know how to fix.

### DIFF
--- a/packages/marshal/src/dot-membrane.js
+++ b/packages/marshal/src/dot-membrane.js
@@ -110,7 +110,7 @@ const makeConverter = (mirrorConverter = undefined) => {
   };
   // We need to pass this while convertYoursToMine is still in temporal
   // dead zone, so we wrap it in convertSlotToVal.
-  const convertSlotToVal = (slot, optIface = undefined) =>
+  const convertSlotToVal = (slot, optIface) =>
     convertYoursToMine(slot, optIface);
   const { serialize: mySerialize, unserialize: myUnserialize } = makeMarshal(
     convertMineToYours,
--- a/packages/promise-kit/index.js
+++ b/packages/promise-kit/index.js
@@ -45,6 +45,9 @@ harden(makePromiseKit);
  * @returns {Promise<Awaited<T>>} A new Promise.
  */
 export function racePromises(values) {
-  return harden(memoRace.call(BestPipelinablePromise, values));
+  const p = /** @type {Promise<Awaited<T>>} */ (
+    memoRace.call(BestPipelinablePromise, values)
+  );
+  return harden(p);
 }
 harden(racePromises);
--- a/jsconfig.eslint-base.json
+++ b/jsconfig.eslint-base.json
@@ -5,9 +5,14 @@
     "checkJs": true,
     "noEmit": true,
     "downlevelIteration": true,
+    "strict": true,
     "strictNullChecks": true,
-    "moduleResolution": "node"
+    "useUnknownInCatchVariables": false,
+    "moduleResolution": "node",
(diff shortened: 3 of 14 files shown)


packages/marshal/src/dot-membrane.js
```diff
@@ -110,7 +110,7 @@ const makeConverter = (mirrorConverter = undefined) => {
   };
   // We need to pass this while convertYoursToMine is still in temporal
   // dead zone, so we wrap it in convertSlotToVal.
-  const convertSlotToVal = (slot, optIface = undefined) =>
+  const convertSlotToVal = (slot, optIface) =>
```
erights, line 113: This removal of `= undefined` doesn't even satisfy your stylistic preferences, let alone mine. It is, as you say, "only" documentary in that its removal does not change the observable behavior of the program. However, even your argument against using this documentary signal is that it is redundant with the ts-declaration-in-jsdoc-comment documentary signal, which is absent in this case. Removing this one in the absence of the other one leaves none, or rather, leaves only the remaining weak third signal of the parameter name itself.

```suggestion
  /** @type {import('./types.js').ConvertSlotToVal<any>} */
  const convertSlotToVal = (slot, optIface = undefined) =>
```

I verified that in my vscode, with the addition of this ts-declaration-in-jsdoc-comment, the vscode red squiggle goes away and `optIface` is more accurately typed

image

This example is also an example of the other point in that thread. The documentary signal in the ts-declaration-in-jsdoc-comment is distant -- in the `@callback` declaration in `./types.js`, and so it not obviously visible at the site of the definition of this `convertSlotToVal` in `dot-membrabe.js`. With the `= undefined`, someone looking at the code locally in `dot-membrane.js` sees that this parameter is supposed to be optional. With *only* the `/** @type {import('./types.js').ConvertSlotToVal<any>} */` as a local documentary signal, they do not.

erights (changes requested):

packages/marshal/src/dot-membrane.js
erights: "that thread" being the internal thread that provoked https://github.com/Agoric/agoric-sdk/issues/8035 . I think this example serves as an adequate example of this point from that thread.

erights: Btw, since you're not turning on `noImplicitAny` (thank you!), I don't understand what rule change caused the code here to become a type error. Could you explain? I'd love to have a better predictive model.

packages/promise-kit/index.js
```diff
@@ -45,6 +45,9 @@ harden(makePromiseKit);
  * @returns {Promise<Awaited<T>>} A new Promise.
  */
 export function racePromises(values) {
-  return harden(memoRace.call(BestPipelinablePromise, values));
+  const p = /** @type {Promise<Awaited<T>>} */ (
+    memoRace.call(BestPipelinablePromise, values)
+  );
+  return harden(p);
```
kungfooman, line 51: One line of nice JS turns into 4 ugly lines just to please TypeScript, why isn't it possible that TS properly infers typs from `.call` :disappointed: 

I'm all for strict types, but TypeScript just doesn't support it itself... so it feels kinda pointless to ts-ignore everything?

PR #1690, opened 2023-07-21
chore(eventual-send): standardize TS/JSDoc typing

michaelfig:
Make this package more of an example of how to do typings, with the intent of eventually extracting a skeleton for documentation.

Also upgrade some Typescript dependencies to avoid warnings when running `@typescript-eslint/eslint-parser`


packages/eventual-send/test/test-types.js
```diff
@@ -2,10 +2,10 @@ import { test } from './prepare-test-env-ava.js';
 
 import { E } from './get-hp.js';
 
-/** @template T @typedef {import('../src/index').ERef<T>} ERef */
+/** @template T @typedef {import('../src/exports').ERef<T>} ERef */
```
mhofman, line 5: Should this test use the fully qualified name of the package? 
```suggestion
/** @template T @typedef {import('@endo/eventual-send').ERef<T>} ERef */
```

packages/eventual-send/src/exports.d.ts
```diff
@@ -1,4 +1,4 @@
-import type { HandledPromiseConstructor, EProxy } from './types.d';
+import type { HandledPromiseConstructor } from './types.d';
```
mhofman, line 1: Nit: still prefer identifiers with extension
```suggestion
import type { HandledPromiseConstructor } from './types.d.ts';
```

yarn.lock
```diff
@@ -10823,13 +10834,20 @@ semver@^6.0.0, semver@^6.3.0:
...
   dependencies:
     lru-cache "^6.0.0"
 
+semver@^7.3.7:
+  version "7.5.4"
+  resolved "https://registry.yarnpkg.com/semver/-/semver-7.5.4.tgz#483986ec4ed38e1c6c48c34894a9182dbff68a6e"
+  integrity sha512-1bCSESV6Pv+i21Hvpxp3Dx+pSD8lIPt8uVjRrxAUt/nbswYc+tK6Y2btiULjd4+fnq15PX+nqQDC7Oft7WkwcA==
+  dependencies:
+    lru-cache "^6.0.0"
+
```
mhofman, line 10850: Can we dedup?

packages/eventual-send/test/test-types.js
michaelfig: While I was at it, I avoided the typedef-import antipattern.

jsconfig.eslint-base.json
```diff
@@ -9,5 +9,5 @@
     "moduleResolution": "node"
   },
   "include": [],
-  "exclude": ["packages/test262-runner/test262"]
+  "exclude": ["packages/test262-runner/test262", "**/*.d.ts"]
```
mhofman, line 12: Wouldn't that prevent linting of explicit / manual `.d.ts` files?

michaelfig: Ah, yes.  I've narrowed it to `**/exports.d.ts`.

michaelfig: Oh, no.  I don't know how to solve this.  `exports.d.ts` needs to be included in order for its manual types to take effect, but if it *is* included, then `exports.js` has the unsilenceable eslint error.

Where do we go from here?

mhofman: I'm confused because I think @warner did not experience this issue in https://github.com/Agoric/agoric-sdk/pull/8075

michaelfig: > I'm confused because I think @warner did not experience this issue in [Agoric/agoric-sdk#8075](https://github.com/Agoric/agoric-sdk/pull/8075)

Swing-store does not include `*.ts` in its tsconfig files, so as you say:

> Wouldn't that prevent linting of explicit / manual `.d.ts` files?

Yes, yes it would.

```
snow:swing-store michael$ cat jsconfig.json 
// This file can contain .js-specific Typescript compiler config.
{
  "extends": "../../tsconfig.json",
  "include": [
    "src/**/*.js",
    "test/**/*.js",
  ],
}
snow:swing-store michael$ cat ../../tsconfig.json 
{
  "compilerOptions": {
    "target": "esnext",
    "module": "esnext",
    "moduleResolution": "node",
    "skipLibCheck": true,

    "allowJs": true,
    "checkJs": true,
    "strict": false,
    "downlevelIteration": true,
    "strictNullChecks": true,
    "noImplicitThis": true,

    "noEmit": true
  },
  "include": [
    ".eslintrc.cjs"
  ]
}
snow:swing-store michael$
```

michaelfig: Until I've tested against the latest typescript-eslint and/or filed a bug against that repository, I've opted for limiting the damage just to `exports.d.ts`.

.prettierignore
```diff
@@ -2,6 +2,7 @@
 bundles
 dist
 fixtures
+output
```
mhofman, line 5: Out of curiosity, what generates output?

packages/eventual-send/src/no-shim.js
```diff
@@ -1,6 +1,10 @@
-// @ts-nocheck
+/* global globalThis */
 import makeE from './E.js';
 
-const hp = HandledPromise;
-export const E = makeE(HandledPromise);
+/** @type {import('./exports.js').HandledPromiseConstructor} */
```
mhofman, line 4: I have mixed feelings about the explicit type. We likely could get an implicit type working with a `import from './exports.js'` at the top.

One one hand, it's safer to have an explicit type, on the other we rely on an implicit type for `E`

mhofman (approved):
Thanks for working that out!

jsconfig.eslint-base.json
mhofman: I'm confused now, what did we settle on for the exclusion?

packages/eventual-send/src/no-shim.js
michaelfig: I had incorrect `declare namespace global` instead of `declare global` syntax in `./exports.d.ts`.  With that fixed, the explicit typing was not required, so I removed it.

.prettierignore
michaelfig: Narrowed to `ses-integration-test`.

jsconfig.eslint-base.json
michaelfig: > what did we settle on for the exclusion?

[Omitting `src/**/exports.js`](https://github.com/endojs/endo/pull/1690/files#diff-ff983a2f596fa583a661fbfceba5f89dc87af77d8a3ac76a42c3356972ee085eR52) from the typescript-eslint configuration.

PR #1691, opened 2023-07-22
docs(bundle-source): Refresh README

### LINKED ISSUE #1645: Stale documentation for Endo bundler packages
The README for `bundle-source` is old enough to mistakenly assert that `getExports` is the only supported `moduleFormat`. We’re up to 3 including `nestedEvaluate` and `endoZipBase64` (with potential additional formats like `endoScript` around the corner)

Furthermore, we need documentation of the known limitations of bundle hash consistency. Currently, bundle hashes are consistent for a given `node_modules` solution, but the production (no dev deps) solution is _not always_ a strict supergraph of the dev (default) solution. Most divergences have been addressed through compartment map garbage collection, but to do better than we’re doing now, requires changes to the package manager to ensure the above invariant: prod is a strict supergraph of dev; that deduplication after introducing dev dependencies must not frustrate the prod solution.

I am also contemplating moving more of the bundler code into compartment mapper so that the compartment mapper can be more of a one-stop-shop for all the modes of bundling, including making the evasive transforms more accessible, particularly for `importLocation`, which isn’t threaded thru to `bundle-source` or `import-bundle` since it bypasses bundling.

kriskowal:
Fixes #1645. This documents the `endoZipBase64` format and includes answers to common questions about inspecting these bundles directly.


packages/bundle-source/README.md
```diff
@@ -3,37 +3,78 @@
...
+
+…to get a promise for a source bundle, that resolves after reading the
 named sources and bundling them into a form that vats can load, as indicated
-by the `moduleFormat` below. Currently, the only supported module format
-is `getExport`. Note that this way of loading external modules is likely to
-change.
+by the `moduleFormat` below.
+
+The resulting bundle is suitable for use with `@endo/import-bundle`.
+The default format is `moduleFormat` of a bundle is `"endoZipBase64"`.
```
Chris-Hibbert, line 30:
```suggestion
The default format is of a bundle is `"endoZipBase64"`.
```

packages/bundle-source/README.md
```diff
@@ -3,37 +3,78 @@
...
+> yarn add -D @endo/cli
+> yarn endo where cache
 ```
 
+Use the `XDG_CACHE_HOME` environment variable to override the default location
+of caches in general.
+The caches will be in `endo/source-map` and `endo/source-map-track`.
+The former is a content-address-store keyed on the SHA-512 of each bundled
+module file.
+The the latter is a location-address-store keyed on the SHA-512 of the fully
```
Chris-Hibbert, line 67:
```suggestion
The latter is a location-address-store keyed on the SHA-512 of the fully
```

Chris-Hibbert (approved):
lgtm

packages/bundle-source/README.md
Chris-Hibbert: Oops!  supposed to be *The default format of a bundle is `"endoZipBase64"`.*

packages/bundle-source/README.md
```diff
@@ -60,3 +101,67 @@ the bundle in the event of any stack traces.
...
+To inspect the contents of a bundle in a JSON file:
+
+```
+jq -r .endoZipBase64 | base64 -d | xxd | less
+```
+
+To extract the contents:
+
+```
+jq -r .endoZipBase64 | base64 -d > bundle.json
```
dckc, line 127: calling it `bundle.json` seems odd. I would have thought `bundle.zip`

kriskowal: Thanks. Queuing up https://github.com/endojs/endo/pull/1693

PR #1694, opened 2023-07-25
docs(ses): Clarify hardening

kriskowal:
Clarify the limits of hardening and the special carve-out for typed arrays.

### DIFF
--- a/packages/ses/README.md
+++ b/packages/ses/README.md
@@ -144,6 +144,29 @@ still closes over the mutable counter.
 Hardening an object graph makes the surface immutable, but does not guarantee
 that methods are free of side effects.
 
+Hardening an object ensures that an attacker cannot alter the API surface of an
+object.
+Hardening "passable" data *does* ensure that a pair of attackers cannot use
+the data to communicate.
+That is, if two guests have access to the same data, they can't manipulate some
+property of the data to send signals.
+
+So, for example, a `Map` instance does not qualify as passable data because
+two guests sharing a map can use it to send messages, by setting and getting
+entries of the map.
+Maps are a capability to read and write, so can be used to communicate.
+
+Typed Arrays including `Uint8Array` do not qualify to be passable data, since
+their content is mutable.
+However, their mutability limits a pair of holders to communicating
+information, not objects bearing capabilities.
+Typed Arrays receive special treatment from `harden` because they cannot be
+frozen.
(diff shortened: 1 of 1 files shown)


packages/ses/README.md
```diff
@@ -144,6 +144,29 @@ still closes over the mutable counter.
 Hardening an object graph makes the surface immutable, but does not guarantee
 that methods are free of side effects.
 
+Hardening an object ensures that an attacker cannot alter the API surface of an
+object.
+Hardening "passable" data *does* ensure that a pair of attackers cannot use
+the data to communicate.
+That is, if two guests have access to the same data, they can't manipulate some
+property of the data to send signals.
```
erights, line 152: Not true. Remotables and Promises are both Passables and are used to communicate. Indeed, we could in the future consider hardened types arrays to be Remotables, alongside MapStores and SetStores.

packages/ses/README.md
```diff
@@ -144,6 +144,29 @@ still closes over the mutable counter.
...
+
+So, for example, a `Map` instance does not qualify as passable data because
+two guests sharing a map can use it to send messages, by setting and getting
+entries of the map.
+Maps are a capability to read and write, so can be used to communicate.
```
erights, line 157: MapStores are Remotable, and so are Passable. The reasons we don't allow Maps to be Remotables are messy, but mainly because they allow storage of non-Passable elements. hardened TypedArrays as Remotables would not have this problem.

packages/ses/README.md
```diff
@@ -144,6 +144,29 @@ still closes over the mutable counter.
...
+
+Typed Arrays including `Uint8Array` do not qualify to be passable data, since
+their content is mutable.
+However, their mutability limits a pair of holders to communicating
+information, not objects bearing capabilities.
+Typed Arrays receive special treatment from `harden` because they cannot be
+frozen.
+Instead, `harden` makes all of their properties, including [expandos][expando],
+non-writable and non-configurable, then makes the TypedArray non-exensible.
+All of the indexed data properties remain writable.
```
erights, line 167: All this would be consistent with treating hardened TypedArrays as Remotables.

erights (changes requested):

kriskowal:
This will require another from-scratch attempt with a clearer model of our vocabulary. Filed #1695

PR #1696, opened 2023-07-25
Read carefully

### LINKED ISSUE #1593: Bundle tools should handle EMFILE gracefully
Bundling opens a lot of concurrent file descriptors, sometimes more than are available, especially when running the bundler concurrently. We should address this in general, but specifically for the `bundle-source` tool.

Tools like `bundle-source` currently construct their file system capabilities from `'fs'` module, which does not gracefully handle EMFILE or EAGAIN. The `graceful-fs` package in `npm` takes care of these concerns but does not provide a promise abstraction. We would need to lower the `compartment-mapper` `node-powers` module a layer of abstraction and embrace `fs` callbacks directly to use `graceful-fs` as a drop-in replacement for `fs`.

Alternately, we should implement graceful handling of these errors ourselves in the compartment mapper. This would implicitly improve other existing applications, since the concern of attenuating an `fs` power is currently deferred to the individual application entrypoints.

The author of this issue notes that we could have the best of all worlds had `graceful-fs` been implemented as a powerless adapter and did not import `fs` directly.

kriskowal:
Fixes #1593, the bundler and compartment-mapper read many files to build a bundle and these reads are by default aggressively concurrent. This can lead to the exhaustion of file descriptors. This is generally silly because a file system does not tend to improve throughput with any amount of concurrency.

However, the compartment mapper is sufficiently general to run on any read powers, so it would be poor form to use a synchronous read function.

With this change, the compartment mapper classifies read errors in Node.js:

- those that indicate that a read was interrupted, as can occur if reading when the process receives a signal
- those that indicate that a read failed because file descriptors were exhausted
- those that indicate that a read failed because a file did not exist, or an entry existed but was not a file

…and monitors the throughput of reads. The governor implements an hybrid control loop based on AIMD (additive increase, multiplicative decrease) and CoDel (control delay).

- if throughput increases, we increase the concurrency limit by 1
- if throughput decreases, we apply a proportional decrease to the concurrency limit
- if we exhaust file descriptors, we halve the concurrency limit
- if we are interrupted, we try again and do not apply the delay against the throughput

And we also incidentally are more careful to distinguish these errors from missing entries when searching for a file (`maybeRead`).

A worthy critique of this PR is: perhaps instead of this level of sophistication you could just limit the Node.js read powers to a single concurrent read.

### DIFF
--- a/packages/bundle-source/cache.js
+++ b/packages/bundle-source/cache.js
@@ -1,3 +1,4 @@
+/* global performance */
 // @ts-check
 import { makePromiseKit } from '@endo/promise-kit';
 import { makeReadPowers } from '@endo/compartment-mapper/node-powers.js';
@@ -350,7 +351,7 @@ export const makeNodeBundleCache = async (dest, options, loadModule) => {
   ]);
 
   const readPowers = {
-    ...makeReadPowers({ fs, url, crypto }),
+    ...makeReadPowers({ fs, url, crypto, now: performance.now }),
     delay: ms => new Promise(resolve => timers.setTimeout(resolve, ms)),
     basename: path.basename,
   };
--- a/packages/bundle-source/src/index.js
+++ b/packages/bundle-source/src/index.js
@@ -1,4 +1,4 @@
-/* global process */
+/* global process performance */
 
 import fs from 'fs';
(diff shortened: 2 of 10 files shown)


kriskowal:
Closing in favor of #1697.

PR #1697, opened 2023-07-25
One concurrent read

kriskowal:
This change limits read concurrency for the bundler to just one.

#1696 illustrates that a control loop governing read concurrency to optimize throughput will effectively limit concurrency to one. This confirms the conventional wisdom that concurrent file system reads do not improve throughput, but massively concurrent reads risk depleting the reservoir of available file descriptors. Perhaps we should just not do that.


packages/compartment-mapper/NEWS.md
```diff
@@ -5,6 +5,13 @@ User-visible changes to the compartment mapper:
 - Introduces support for source map generation.
   Look for `computeSourceMapLocation` and `sourceMapHook` in
   [`README.md`](README.md).
+- This version reads more carefully, adding a control loop around concurrent
+  reads and classifying read errors caused by interrupts, resource exhaustion,
+  non-existant files, and existence of non-files.
+  To preserve backward compatibility and object capability principles, the
+  governor uses a low resolution timer by default and can be endowed with
+  `performance.now` for a better signal.
```
michaelfig, line 13: This seems stale.

packages/compartment-mapper/src/node-powers.js
```diff
@@ -40,18 +40,43 @@ const makeReadPowersSloppy = ({ fs, url = undefined, crypto = undefined }) => {
...
     url === undefined ? fakePathToFileURL : url.pathToFileURL;
 
+  let readMutex = Promise.resolve(undefined);
+
   /**
    * @param {string} location
    */
   const read = async location => {
+    const promise = readMutex;
+    let release = Function.prototype;
```
michaelfig, line 50: Weird.  I had to stare at this to see the value wasn't used for anything except typing.

packages/compartment-mapper/test/scaffold.js
michaelfig, line 1: Are these changes also stale?

michaelfig (approved):
It's simple, and that WFM!

Just a little apparently-dead code to clean up.

packages/compartment-mapper/src/node-powers.js
kriskowal: Indeed and it avoids an allocation of an otherwise pointless noop function, or a pointless runtime assertion. But, there isn’t a winning move here, as far as I can tell.

PR #1701, opened 2023-07-27
chore(transforms): Scaffold new package

kriskowal:
To be the eventual new home for the evasive transforms currently found only in `@endo/bundle-source` that are more generally useful.


kriskowal:
Please join me in celebrating 1701 Github issues.🖖

image

naugtur (approved):

PR #1702, opened 2023-07-27
chore: Update package skeleton

kriskowal:
This moves the bulk of the project skeleton into `packages/skel`. The script `scripts/create-package.sh` takes over for the old `scripts/repackage.sh`. It loses the ability to update an existing package in-place (which has always been dubious) and instead allows the skeleton to more closely resemble an ordinary package, and allows `lerna` to keep the workspace dependencies of the skeleton up-to-date.


kriskowal:
@erights I’m altering the process for generating new Endo packages.

image

> Pray I don’t alter it further.

michaelfig (approved):

Issue #1665, opened 2023-07-01
FYI: typescript-eslint@v6 will remove parserServices.hasFullTypeInformation


JoshuaKGoldberg:
👋 Hi! I'm one of the maintainers of [typescript-eslint](https://typescript-eslint.io). We're getting ready to release a new v6 major version of the tooling, which includes some small breaking changes.

Coming over from https://github.com/typescript-eslint/typescript-eslint/issues/7124: I see this repo contains references to `hasFullTypeInformation`. That property is something we've never recommended using (see that issue & https://github.com/typescript-eslint/typescript-eslint/issues/7158). We recommend instead using `ESLintUtils.getParserServices` per https://typescript-eslint.io/developers/custom-rules#typed-rules.

If you _really_ want to "progressively enhance" lint rules, `parserServices.program` will be `null` if type information isn't available. But we definitely recommend against that kind of rule behavior changing.

Overall, I'd suggest trying out typescript-eslint@v6 and making sure internal builds & downstream repos still compile & lint nicely with all the changes. Let me know if there's anything we can do to help with the v6 upgrade! 💖

turadg:
Thanks for the note. We'll be sure to address this before upgrading. https://typescript-eslint.io/blog/announcing-typescript-eslint-v6-beta

PR #1704, opened 2023-07-28
always lint types

turadg:
closes: #1665

v6 of typescript-eslint made it fast enough that we don't need to be selective about where/when we run it.

This updates to that version and cuts out what we no longer need.

### DIFF
--- a/packages/base64/test/test-main.js
+++ b/packages/base64/test/test-main.js
@@ -2,6 +2,10 @@ import test from 'ava';
 import { atob as origAtob, btoa as origBtoa } from './capture-atob-btoa.js';
 import { encodeBase64, decodeBase64, atob, btoa } from '../index.js';
 
+/**
+ * @param {string} string
+ * @returns {Uint8Array}
+ */
 function stringToBytes(string) {
   const data = new Uint8Array(string.length);
   for (let i = 0; i < string.length; i += 1) {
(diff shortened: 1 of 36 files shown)


packages/base64/test/test-main.js
```diff
@@ -2,6 +2,10 @@ import test from 'ava';
 import { atob as origAtob, btoa as origBtoa } from './capture-atob-btoa.js';
 import { encodeBase64, decodeBase64, atob, btoa } from '../index.js';
 
+/**
+ * @param {string} string
+ * @returns {Uint8Array}
+ */
```
kriskowal, line 8: 🙏

packages/cjs-module-analyzer/index.js
```diff
@@ -7,6 +7,7 @@
 /* eslint no-continue: ["off"] */
 /* eslint no-shadow: ["off"] */
 /* eslint no-unreachable-loop: ["off"] */
+/* eslint @endo/restrict-comparison-operands: ["off"] */
```
kriskowal, line 10: cc @naugtur Would like to know whether we can turn this one on maybe someday.

kriskowal (approved):

packages/cjs-module-analyzer/index.js
naugtur: I don't see any `==` comparisons in there. What am I missing?

turadg: >=

PR #1705, opened 2023-07-29
feat(ses): Vetted shims

### LINKED ISSUE #422: Vetted shims
To improve compatibility between Hardened JavaScript and shims, programs that upgrade the underlying language, some shims need to be in a position where they can introduce new properties to shared intrinsics and for those properties to remain visible after lockdown.

## Enable Vetted Shims

In the first phase of this design, we would introduce a separate `repair(options)` such that the following programs are equivalent:

```js
lockdown(options);
```

```js
repair(options);
lockdown();
```

In the latter form, shims can be executed between `repair` and `lockdown` that alter the shared intrinsics in ways that will not be erased by calling `lockdown`. These are vetted shims.

`repair()` will introduce the global `harden()`, but between `repair()` and `lockdown()`, `harden` will simply mark objects that must be hardened by `lockdown()`. Currently, initializing the `ses` module introduces `harden()`, and it throws an error until after calling `lockdown()`. Introducing `harden` at time of `repair` is consistent with the idea that `repair` is effectively a shim that upgrades the base JavaScript to Almost-Hardened JavaScript, such that shims can run, then `lockdown` is effectively a shim that completes the job.

In the following program, `lockdown()` will throw an error because repair options were already applied.

```js
repair();
lockdown(options);
```

Allowing `lockdown(options)` to serve as both `repair` and `lockdown` if `repair` has not already been called gives us a graceful migration path from the status quo, and may even be desirable as a feature of the Hardened JavaScript standard since it simplifies usage in scenarios that do not have vetted shims.

## Enable Hardened Vetted Shims

The above design is consistent with further work, to be tracked in subsequent issues:

`repair` introduces new shared intrinsics but does not expose them in a way that vetted shims would be able to alter them. For this to become possible, `repair` would need to upgrade `getIntrinsic` and `getIntrinsics` in place such that, for example, `getIntrinsic('%SharedMath%')` could be upgraded in place. This would presumably also introduce a `%sharedGetIntrinsic%` and `%sharedGetIntrinsics%` function that could be safely shared with compartments, providing access only to the shared intrinsics and no others. Further intrinsics would be introduced to compartments by either writing a wrapper for these or possibly additional compartment constructor options with the same effect.

## Eject Internal Shims

Separating `repair` and `lockdown` will allow us to refactor `ses` such that some of the internal layers can become separate shims. For example, the internal compartment-shim could be refactored into a layers: compartment-as-module-loader and compartment-as-confinement. We can remove `HandledPromise` from the permits, since the `handled-promise-shim` can be applied between `repair` and `lockdown` instead. The `console`, `assert`, and `Error` shims could be separated, and alternate implementations can be used to provide the same services.

## Alternatives Considered

The other major design family that we considered and rejected was a variant on this design that lacked a separable `repair()` phase, also allowed but deferred `harden` until after `lockdown()`, and had an additional API for submitting plans for changes to the permits structure that `lockdown()` would apply. The crux of the design decision is the emergent behavior of shims that must have a follow-up shim to accommodate Hardened JavaScript. In both design spaces, these shims would need to use `getIntrinsic` to make changes to shared shims and mark new intrinsics with `harden`. The only material difference is that these hardening-shims would also need to use permit objects to plan for behaviors to be applied during lockdown instead of immediately. We elected to preserve the current convention, that shims modify the environment in place, at the expense of having a separate `repair()` call that serves as a shim for taming and repairing the JavaScript environment, upgrading it to Almost-Hardened JavaScript.

---

**[original description, 2020-08-18 by kriskowal:]**

Endo needs the ability to promote packages and modules that serve as vetted shims, such that they execute after repairs and before lockdown, before any non-shim modules execute. This is a form of dependency “hoisting”. We will need to incorporate the allow-list metadata into the `package.json` of either the shim, the module that depends upon the shim, or create shim envelope packages to host the metadata. The metadata would be collected by Endo and captured in the compartment map.

kriskowal:
Closes #422 

This change separates out the internal phases of `lockdown`: `repairIntrinsics()` and `hardenIntrinsics()`.


kriskowal:
Developer log (draft):

I’m attempting to make it clearer that `ses` is itself a stack of shims, so I’ve refactored `index.js` to reflect that and intend. I’ve also attempted to factor `x.js` and `x-shim.js` modules such that `x.js` implements the behavior in a way that might be used without shimming.

In the course of doing this, it became clear that the `errors` sub-section should not be using ambient types. I’ve attempted to correct this. I’m stopping for now at this yak:

```
$ tsc
types.test-d.ts:91:1 - error TS2775: Assertions require every name in the call target to be declared with an explicit type annotation.

91 assert.typeof(10.1, 'number');
...
```

kriskowal:
I managed to getting vetted shims implemented without refactoring ambient types. Or rather, I managed to keep that work out of this pull request for purposes of limiting the scope of review and will submit that cleanup separately.

kriskowal:
For reviewers: I’ve created an initial commit that renames `lockdown-shim.js` and `compartment-shim.js` to `lockdown.js` and `compartment.js` so that the subsequent commit’s diff clearly illustrates that all of the shim behavior was extracted from those modules and moved into the (new) adjacent `-shim.js` layering.

packages/ses/index.js
```diff
@@ -12,34 +12,6 @@
...
-
-/** getThis returns globalThis in sloppy mode or undefined in strict mode. */
-function getThis() {
-  return this;
-}
-
-if (getThis()) {
-  // See https://github.com/endojs/endo/blob/master/packages/ses/error-codes/SES_NO_SLOPPY.md
-  throw TypeError(`SES failed to initialize, sloppy mode (SES_NO_SLOPPY)`);
-}
```
kriskowal, line 31: Moves to a new `assert-sloppy-mode.js`, as imported by `lockdown-shim.js` to ensure it runs first and fails fast.

packages/ses/index.js
```diff
@@ -12,34 +12,6 @@
...
-
-const Compartment = makeCompartmentConstructor(
-  makeCompartmentConstructor,
-  getGlobalIntrinsics(globalThis),
-  markVirtualizedNativeFunction,
-);
-
-assign(globalThis, {
-  lockdown,
-  Compartment,
```
kriskowal, line 43: Moves to compartment-shim.js

packages/ses/index.js
```diff
@@ -12,34 +12,6 @@
...
-const markVirtualizedNativeFunction = tameFunctionToString();
-
-const Compartment = makeCompartmentConstructor(
-  makeCompartmentConstructor,
-  getGlobalIntrinsics(globalThis),
-  markVirtualizedNativeFunction,
-);
-
-assign(globalThis, {
-  lockdown,
```
kriskowal, line 42: Moves to lockdown-shim.js

packages/ses/index.js
```diff
@@ -12,34 +12,6 @@
...
-  Compartment,
-  assert,
```
kriskowal, line 44: Moves to assert-shim.js

packages/ses/index.js
```diff
@@ -12,34 +12,6 @@
...
-  throw TypeError(`SES failed to initialize, sloppy mode (SES_NO_SLOPPY)`);
-}
-
-const markVirtualizedNativeFunction = tameFunctionToString();
-
-const Compartment = makeCompartmentConstructor(
-  makeCompartmentConstructor,
-  getGlobalIntrinsics(globalThis),
-  markVirtualizedNativeFunction,
-);
```
kriskowal, line 39: Moves to compartment-shim.js

packages/ses/src/lockdown-shim.js
```diff
@@ -0,0 +1,35 @@
...
+    // Reveal harden after hardenIntrinsics.
+    // Harden is dangerous before hardenIntrinsics because hardening just
+    // about anything will inadvertently render intrinsics irreparable.
+    // Also, for modules that must work both before or after lockdown (code
+    // that is portable between JS and SES), the existence of harden in global
+    // scope signals whether such code should attempt to use harden in the
+    // defense of its own API.
+    // @ts-ignore harden not yet recognized on globalThis.
+    globalThis.harden = hardenIntrinsics();
+  };
```
kriskowal: I’ve moved all of the `globalThis` assignment behaviors into the shim to decontaminate the non-shim modules.

So, note that the global `repairIntrinsics` doesn’t return `hardenIntrinsics` and the global `hardenIntrinsics` does not return `harden`, but the internal versions perform this cascade.

`hardenIntrinsics` did not previously return `harden`.

leotm (commented):
we should update docs too right (mb better as follow-up)
- https://github.com/endojs/endo/blob/master/packages/ses/docs/guide.md
- https://github.com/endojs/endo/blob/master/packages/ses/docs/reference.md

so `lockdown` and `harden` relevant parts then add `repair` section

can prob reuse bits from the well-written issue description too

kriskowal:
I’ve added a commit with documentation changes and the update for the change log. Thank you!

packages/ses/docs/reference.md
```diff
@@ -32,6 +32,22 @@ To use SES as a script on the web, use the UMD build.
...
+intrinsics, calling `lockdown(options)` is equivalent to running
+`repairIntrinsics(options)` then `hardenIntrinsics()` and vetted shims can run
+in between.
+```js
+import './non-ses-code-before-lockdown.js';
+import './ses-repair-intrinsics.js'; // calls repairIntrinsics.
+import './vetted-shim.js';
+import './ses-harden-intrinsics.js'; // calls repairIntrinsics.
+import './ses-code-after-lockdown.js';
+```
```
erights, line 45: It is even more important that this example import sequence appear in the guide. This is the use pattern we should generally encourage.

erights (commented):
just doc comments so far. rest coming but will take longer.

erights (commented):
What about https://github.com/endojs/endo/blob/master/packages/ses/docs/lockdown.md ?

Should "Lockdown options" be renamed? What about `LOCKDOWN_*` environment variables?

IMO the env vars do not need to be renamed. The doc should at least be clear about what these options mean in the new world.

kriskowal:
In the type definition, I took the liberty of renaming `LockdownOptions` to `RepairOptions` and leaving an alias for backward compatibility. I agree that we do not need to change the names of the environment variables, or even really call attention to `repairIntrinsics()` unless it’s needed.

packages/ses/docs/guide.md
```diff
@@ -239,15 +285,29 @@ makes those global objects available.
...
+- [`lockdown()`](#lockdown) is a shorthand for `repairIntrinsics` and `hardenIntrinsics`.
+
+- [`Compartment`](https://github.com/endojs/endo/tree/SES-v0.8.0/packages/ses#compartment)
+  Code runs inside a `Compartment` and can create sub-compartments to host
+  other code (with different globals or code transforms).
+  The globals in a child compartment include the shared intrinsics including
+  `harden` and a batch of evaluators that run programs that will also be
+  confined to the compartment including `eval`, `Function`, and `Compartment`
+  itself.
+  Compartments also load and confine modules.
```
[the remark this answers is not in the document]
kriskowal: `Compartment` exists before `repairIntrinsics` and the only current difference is that the intrinsics are not hardened. `Compartments` constructed before `repairIntrinsics` do not guarantee confinement even after `repairIntrinsics`.

kriskowal:
Question for reviewers: do we need to tackle making `harden` available after `repairIntrinsics` but before `hardenIntrinsics` with the queuing behavior? The heart of this question is whether it’s necessary and whether we can expect any environment to repair but not harden indefinitely.

On the question of necessity, a vetted shim could be divided into layers that run before and after `hardenIntrinsics()`. So, I’d argue it’s not necessary but might make for better ergonomics once we start implementing hardened shims.

I also don’t think repairing but not hardening intrinsics is coherent, but I might be forgetting something.

mhofman:
Agreed with @erights. Let's not introduce some queuing behavior for `harden` that may not be part of the final design. I am ok with requiring shims to run a part post `hardenIntrinsics` for now. Maybe we should consider for `harden` to throw before `hardenIntrinsics` is called (which I believe is currently the behavior of XS's `harden` implementation before `lockdown` is called)

erights:
> Maybe we should consider for `harden` to throw before ...

Also fine with me. Even better because the diagnostic can help set expectations. My favorite word for such messages is "yet" ;)

erights:
OTOH, absence would make feature testing pleasant. I'm ok either way.

kriskowal:
I’m settling on the current state: presence of `globalThis.harden` is a reliable signal that it’s available for use. I think this is more likely to be consistent with the `@endo/harden` ponyfill @mhofman and I have been discussing, and is the more reversible option.

erights:
> For reviewers: I’ve created an initial commit that renames lockdown-shim.js and compartment-shim.js to lockdown.js and compartment.js so that the subsequent commit’s diff clearly illustrates that all of the shim behavior was extracted from those modules and moved into the (new) adjacent -shim.js layering.

Thank you! I was reviewing the overall differences until I got to one of those. I'm surprised that github diff loses the mv behavior for the overall diffs, but it does. Your separation is a lifesaver.

packages/ses/types.d.ts
```diff
@@ -36,6 +36,10 @@ export interface LockdownOptions {
   __hardenTaming__?: 'safe' | 'unsafe';
 }
 
+export type LockdownOptions = RepairOptions;
```
erights, line 39: Please comment that we preserve `LockdownOptions` only for legacy compat. Perhaps deprecate?

erights (approved):
LGTM!!!!

I love that we can get this into the next endo release. Awesome!

PR #1708, opened 2023-08-01
feat(bundle-source): type `bundleSource` more comprehensively

michaelfig:
## Description

Recently, #1704 was a good step forward to introduce typing for the return value of `@endo/bundle-source`.  However, it came at the cost of requiring all callers to narrow the return type using assertions, knowing the moduleFormat they were using.

This PR discriminates between the different ModuleFormats to return an appropriate type that requires no assertions or narrowing (unless the supplied ModuleFormat cannot be inferred in the calling context).

### Security Considerations

n/a

### Scaling Considerations

n/a

n/a

### Testing Considerations

The types could be tested more thoroughly, but they are at least exercised in `@endo/bundle-source` unit tests.

### Upgrade Considerations

n/a

### DIFF
--- a/packages/bundle-source/src/index.js
+++ b/packages/bundle-source/src/index.js
@@ -219,22 +219,23 @@ async function bundleZipBase64(startFilename, options = {}, powers = {}) {
         .readFile(sourceMapTrackerPath, 'utf-8')
         .then(async oldSha512 => {
           oldSha512 = oldSha512.trim();
-          if (oldSha512 !== sha512) {
-            const oldSha512Head = oldSha512.slice(0, 2);
-            const oldSha512Tail = oldSha512.slice(2);
-            const oldSourceMapDirectory = pathResolve(
-              sourceMapsCacheDirectory,
-              oldSha512Head,
-            );
-            const oldSourceMapPath = pathResolve(
-              oldSourceMapDirectory,
-              `${oldSha512Tail}.map.json`,
-            );
-            await fs.promises.unlink(oldSourceMapPath);
-            const entries = await fs.promises.readdir(oldSourceMapDirectory);
-            if (entries.length === 0) {
-              await fs.promises.rmdir(oldSourceMapDirectory);
-            }
+          if (oldSha512 === sha512) {
+            return;
+          }
+          const oldSha512Head = oldSha512.slice(0, 2);
+          const oldSha512Tail = oldSha512.slice(2);
+          const oldSourceMapDirectory = pathResolve(
+            sourceMapsCacheDirectory,
+            oldSha512Head,
+          );
+          const oldSourceMapPath = pathResolve(
+            oldSourceMapDirectory,
+            `${oldSha512Tail}.map.json`,
+          );
+          await fs.promises.unlink(oldSourceMapPath);
+          const entries = await fs.promises.readdir(oldSourceMapDirectory);
+          if (entries.length === 0) {
+            await fs.promises.rmdir(oldSourceMapDirectory);
           }
         })
         .catch(error => {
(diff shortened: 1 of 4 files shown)


packages/bundle-source/src/index.js
```diff
@@ -407,6 +416,7 @@ ${sourceBundle[entrypoint]}
 return module.exports;
 }
 ${sourceMap}`;
+    return /** @type {const} */ (harden({ moduleFormat, source, sourceMap }));
```
turadg, line 419: Not enough is known for these to be literals. What difference does the const make? (I suspect none.)

packages/bundle-source/src/index.js
```diff
@@ -530,31 +540,48 @@ function getExportWithNestedEvaluate(filePrefix) {
   return computeExports(entrypoint, { require: systemRequire, systemEval }, {});
 }
 ${sourceMap}`;
+    return /** @type {const} */ (harden({ moduleFormat, source, sourceMap }));
```
turadg, line 543: ditto, suspect superfluous const

packages/bundle-source/src/index.js
```diff
@@ -276,15 +276,22 @@ async function bundleZipBase64(startFilename, options = {}, powers = {}) {
...
     endoZipBase64,
     endoZipBase64Sha512: sha512,
   });
 }
 
+/**
+ * @template {'nestedEvaluate' | 'getExport'} T
+ * @param {string} startFilename
+ * @param {T} moduleFormat
+ * @param {*} powers
```
turadg, line 293: annotating out of scope. out of curiosity: do you know it's destructured in the body instead of params?

packages/bundle-source/src/index.js
```diff
@@ -276,15 +276,22 @@ async function bundleZipBase64(startFilename, options = {}, powers = {}) {
       sourceMapJobs.add(writeSourceMap(sourceMap, sourceDescriptor));
     },
   });
+  assert(sha512);
   await Promise.all(sourceMapJobs);
   const endoZipBase64 = encodeBase64(bytes);
   return harden({
-    moduleFormat: 'endoZipBase64',
+    moduleFormat: /** @type {const} */ ('endoZipBase64'),
```
turadg, line 283: I wish we had enums. Consider having something up top like,
```
export const ModuleFormat = /** @type {const} */ ({
  getExport: 'getExport',
  nestedEvaluate: 'nestedEvaluate',
  endoZipBase64: 'endoZipBase64',
});

const DEFAULT_MODULE_FORMAT = ModuleFormat.endoZipBase64;
const DEFAULT_FILE_PREFIX = '/bundled-source/...';
const SUPPORTED_FORMATS = Object.values(ModuleFormat);
```
then the string wouldn't need a const cast when used,
```suggestion
    moduleFormat: ModuleFormat.endoZipBase64,
```

I'm on the fence about it myself since we don't have enums and TS will detect string literal errors. But it does make it read a little easier

turadg (approved):
Some suggestions and non-blocking concerns

packages/bundle-source/src/index.js
kriskowal: `moduleFormat` is a union of constant strings. I don’t know the effect of capturing the rest.

packages/bundle-source/src/types.js
```diff
@@ -3,23 +3,60 @@
  */
 
 /**
- * @callback BundleSource
+ * @typedef {BundleSourceEndoZipBase64 & BundleSourceGetExport & BundleSourceNestedEvaluate} BundleSource
```
kriskowal, line 6: I’m surprised that this is intersection instead of union but this is clearly where the money is made.

kriskowal (approved):

packages/bundle-source/src/index.js
michaelfig: Yeah, the `const` was needed for an intermediate refactor, but I forgot to remove it after.

packages/bundle-source/src/index.js
michaelfig: I want to narrow the return type, otherwise the returned `moduleFormat` doesn't match that of the supplied `moduleFormat`.

packages/bundle-source/src/index.js
michaelfig: I'm not entirely sold on it either.  So many of our callers traffic in string moduleFormats, and I don't want to add abstractions unless they're motivated by the callers.

packages/bundle-source/src/types.js
michaelfig: Intersection means the result has characteristics of all the types.  Union means it might be any of the types (and you need to sniff it at runtime or cast it to determine which one you actually have).

PR #1710, opened 2023-08-02
feat(env-options): env-options conveniences for common cases

erights:
refs: https://github.com/Agoric/agoric-sdk/pull/8136
refs: https://github.com/Agoric/agoric-sdk/issues/8096
refs: https://github.com/Agoric/agoric-sdk/blob/master/docs/env.md#DEBUG
refs: https://github.com/endojs/endo/pull/1922

The indirection of `getEnvironmentCaptor` is only needed to support `getCapturedEnvironmentOptionNames`
`getCapturedEnvironmentOptionNames` is not currently used in either `endo` or `agoric-sdk`. Thus, we have enhanced
`@endo/env/options` to also directly export non-capturing forms of the other functions, like `getEnvironmentOption`, that are made and returned by `getEnvironmentCaptor`.

The value of an option is often supposed to be a list of strings separated by commas (`','`). However
https://github.com/Agoric/agoric-sdk/issues/8096 explains that sometimes our code (both in endo and in agoric-sdk) accidentally split on colon (`':'`) instead due to confusion about what the convention was supposed to be.

So that this PR would only be a compat `feat` PR, this PR does not fix any of those legacy manual uses that got the convention wrong. (Though it does add a TODO that links to https://github.com/Agoric/agoric-sdk/issues/8096.) Rather,
this PR does provide additional conveniences, `getEnvironmentOptionsList` and `environmentOptionsListHas` based on the correct comma option.

Later PRs (https://github.com/endojs/endo/pull/1922) that do take the compat hit of switching to comma should also switch to using these new conveniences.

### DIFF
--- a/packages/env-options/README.md
+++ b/packages/env-options/README.md
@@ -64,9 +64,9 @@ const DEBUG_VALUES = getEnvironmentOptionsList('DEBUG');
 const DEBUG_AGORIC = environmentOptionsListHas('DEBUG', 'agoric');
 ```
 
-Another common Unix convention is for the value of an option to be a
+Another common convention is for the value of an option to be a
 comma (`','`) separated list of strings. `getEnvironmentOptionsList` will
-return this list, or an empty list of the option is absent.
+return this list, or an empty list if the option is absent.
 `environmentOptionsListHas` will test if this list contains a specific
 value, or return false if the option is absent.
 
@@ -79,9 +79,10 @@ than comma. Once these are fixed, then these uses can be switched to use
 
 The `'@endo/env-options'` module also exports a lower-level
 `makeEnvironmentCaptor` that you can apply to whatever object you wish to treat
-as a global, such as the global of another compartment. It returns an entagled
+as a global(having a "process" property with its own "env" record),
+such as the global of another compartment. It returns an entagled
 pair of a `getEnvironmentOption` function as above, and a
-`getCapturedEnvironmentOptionNames` function for that returns an array of
+`getCapturedEnvironmentOptionNames` function that returns an array of
 the option names used by that `getEnvironmentOption` function. This is
 useful to give feedback about
 which environment variables were actually read, for diagnostic purposes.
--- a/packages/env-options/README.md
+++ b/packages/env-options/README.md
@@ -34,24 +34,56 @@ optionally holding a property with the same name as the option,
 whose value is the configuration setting of that option.
 
 ```js
-import { makeEnvironmentCaptor } from '@endo/env-options';
-const { getEnvironmentOption } = makeEnvironmentCaptor(globalThis);
+import { getEnvironmentOption } from '@endo/env-options';
 const FooBarOption = getEnvironmentOption('FOO_BAR', 'absent');
 ```
 
 The first argument to `getEnvironmentOption` is the name of the option.
 The value of `FooBarOption` would then be the value of
 `globalThis.process.env.FOO_BAR`, if present.
-If setting is either absent or `undefined`, the default `'absent'`
-would be used instead.
+If value is either absent or `undefined`, the second argument,
+such as `'absent'`, would be used instead.
 
 In either case, reflecting Unix environment variable expectations,
 the resulting setting must be a string.
 This restriction also helps ensure that this channel is used only to pass data,
 not authority beyond the ability to read this global state.
 
-The `makeEnvironmentCaptor` function also returns a
-`getCapturedEnvironmentOptionNames` function for use to give feedback about
+```js
+const ENABLED =
+  getEnvironmentOption('TRACK_TURNS', 'disabled', ['enabled']) === 'enabled';
+```
+
+`getEnvironmentOption` also takes an optional third argument, which if present
+is an exhaustive list of allowed strings other than the default. If present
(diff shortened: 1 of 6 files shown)


erights:
See also https://github.com/nodejs/node/pull/48890

packages/env-options/src/env-options.js
```diff
@@ -78,15 +88,49 @@ export const makeEnvironmentCaptor = aGlobal => {
...
   };
   freeze(getEnvironmentOption);
 
+  /**
+   * @param {string} optionName
+   * @returns {string[]}
+   */
+  const getEnvironmentOptionsList = optionName => {
+    const option = getEnvironmentOption(optionName, '');
+    return option === '' ? [] : stringSplit(option, ',');
```
michaelfig, line 107: I prefer (but do not require) the equivalent:

```suggestion
    return stringSplit(option, ',');
```

michaelfig (approved):
LGTM!

packages/env-options/README.md
```diff
@@ -34,24 +34,56 @@ optionally holding a property with the same name as the option,
...
+is an exhaustive list of allowed strings other than the default. If present
+and the actual environment option is neither the default nor one of these
+allowed strings, then an error is thrown explaining the problem.
+
+```js
+const DEBUG_VALUES = getEnvironmentOptionsList('DEBUG');
+const DEBUG_AGORIC = environmentOptionsListHas('DEBUG', 'agoric');
+```
+
+Another common Unix convention is for the value of an option to be a
```
gibson042, line 67:
```suggestion
Another common convention is for the value of an option to be a
```

packages/env-options/README.md
```diff
@@ -34,24 +34,56 @@ optionally holding a property with the same name as the option,
...
+comma (`','`) separated list of strings. `getEnvironmentOptionsList` will
+return this list, or an empty list of the option is absent.
```
gibson042, line 69:
```suggestion
return this list, or an empty list if the option is absent.
```

packages/env-options/README.md
```diff
@@ -34,24 +34,56 @@ optionally holding a property with the same name as the option,
...
+than comma. Once these are fixed, then these uses can be switched to use
+`getEnvironmentOptionsList` or `environmentOptionsListHas`.)
+
+## Tracking used option names
+
+The `'@endo/env-options'` module also exports a lower-level
+`makeEnvironmentCaptor` that you can apply to whatever object you wish to treat
+as a global, such as the global of another compartment. It returns an entagled
+pair of a `getEnvironmentOption` function as above, and a
+`getCapturedEnvironmentOptionNames` function for that returns an array of
```
gibson042, line 84:
```suggestion
`getCapturedEnvironmentOptionNames` function that returns an array of
```

packages/env-options/README.md
```diff
@@ -34,24 +34,56 @@ optionally holding a property with the same name as the option,
...
+(Compat note: https://github.com/Agoric/agoric-sdk/issues/8096 explains that
+for `DEBUG` specifically, some existing uses split on colon (`':'`) rather
+than comma. Once these are fixed, then these uses can be switched to use
+`getEnvironmentOptionsList` or `environmentOptionsListHas`.)
+
+## Tracking used option names
+
+The `'@endo/env-options'` module also exports a lower-level
+`makeEnvironmentCaptor` that you can apply to whatever object you wish to treat
+as a global, such as the global of another compartment. It returns an entagled
```
gibson042, line 82:
```suggestion
as a global (having a "process" property with its own "env" record),
such as the global of another compartment. It returns an entangled
```

packages/env-options/src/env-options.js
```diff
@@ -47,13 +52,16 @@ export const makeEnvironmentCaptor = aGlobal => {
    *
    * @param {string} optionName
    * @param {string} defaultSetting
+   * @param {string[]} [optOtherNames]
```
gibson042, line 55: Suggestion:
```suggestion
   * @param {string[]} [otherValues] If provided, the option value must be included or match `defaultSetting`
```

packages/env-options/src/env-options.js
gibson042: `stringSplit(option, ',')` would not be equivalent because `''.split(',')` is `['']` rather than `[]`. But we probably _should_ be freezing the output.
```suggestion
    return freeze(option === '' ? [] : stringSplit(option, ','));
```

gibson042 (approved):
LGTM after minor fixes.

packages/env-options/src/env-options.js
erights: Good point about freezing the output. Done.

packages/env-options/src/env-options.js
erights: Done

packages/env-options/README.md
erights: Done

packages/env-options/README.md
erights: Done

packages/env-options/README.md
erights: Done

packages/env-options/README.md
erights: Done

PR #1712, opened 2023-08-08
fix(patterns)!: tag and retype guards

erights:
Staged on https://github.com/endojs/endo/pull/1771

Make guards a distinct category of taggeds, on par with matchers.

### DIFF
--- a/packages/patterns/src/types.js
+++ b/packages/patterns/src/types.js
@@ -509,10 +509,10 @@ export {};
 
 /**
  * @template {Record<PropertyKey, MethodGuard>} [T=Record<PropertyKey, MethodGuard>]
- * @typedef {{ klass: 'Interface' } & InterfaceGuardPayload<T> } InterfaceGuard
- *
- * TODO https://github.com/endojs/endo/pull/1712 to make it into a genuine
- * guard that is distinct from a copyRecord
+ * @typedef {CopyTagged & {
+ *   [Symbol.toStringTag]: 'guard:interfaceGuard',
+ *   payload: InterfaceGuardPayload,
+ * }} InterfaceGuard
  */
 
 /**
@@ -581,14 +581,10 @@ export {};
  */
 
 /**
- * @typedef {{ klass: 'methodGuard' } & MethodGuardPayload} MethodGuard
- *
- * TODO https://github.com/endojs/endo/pull/1712 to make it into a genuine
- * guard that is distinct from a copyRecord.
- * Once we're ready for such a compat break, we might also take the
- * opportunity to rename `restArgGuard` and `returnGuard`
- * to reflect that their value must be a Pattern rather that a
- * non-pattern guard.
+ * @typedef {CopyTagged & {
+ *   [Symbol.toStringTag]: 'guard:methodGuard',
+ *   payload: MethodGuardPayload,
+ * }} MethodGuard
  */
 
 /**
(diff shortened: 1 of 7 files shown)


packages/patterns/src/types.js
```diff
@@ -574,37 +573,25 @@ export {};
...
+ *     returnGuard?: Pattern
+ *   }
  * }} MethodGuard
- *
- * TODO https://github.com/endojs/endo/pull/1712 to make it into a genuine
- * guard that is distinct from a copyRecord.
- * Once we're ready for such a compat break, we might also take the
- * opportunity to rename `restArgGuard` and `returnGuard`
- * to reflect that their value must be a Pattern rather that a
- * non-pattern guard.
```
michaelfig, line 591: Any desire to take on this change as well?

michaelfig (commented):
Looking good!

packages/patterns/src/types.js
erights: I'd like to. Depends on how we cope with the compat issue shown at https://github.com/endojs/endo/pull/1715#issuecomment-1682929677

erights: If all we do to smooth the transition is https://github.com/endojs/endo/pull/1745 , then we would not enable this renaming. I think that's ok.

erights:
Reviewers,
This is actually ready for review, but not marked as such because it cannot be merged until agoric-sdk uses #1745 to be compat with the transition to this PR.

packages/patterns/src/types.js
erights: All we did is #1745

erights:
https://github.com/Agoric/agoric-sdk/pull/8339 is where we're altering agoric-sdk to pave the way for this PR.

erights:
Hi @michaelfig , with https://github.com/Agoric/agoric-sdk/pull/8339 merged, this is now ready for review.

erights:
> Hi @michaelfig , with [Agoric/agoric-sdk#8339](https://github.com/Agoric/agoric-sdk/pull/8339) merged, this is now ready for review.

And with https://github.com/Agoric/agoric-sdk/pull/8340 showing that agoric-sdk should now be compat with this PR

packages/patterns/src/types.js
erights: Resolving, with these renamings not done.

packages/patterns/src/types.js
```diff
@@ -509,10 +509,10 @@ export {};
...
 /**
  * @template {Record<PropertyKey, MethodGuard>} [T=Record<PropertyKey, MethodGuard>]
- * @typedef {{ klass: 'Interface' } & InterfaceGuardPayload<T> } InterfaceGuard
- *
- * TODO https://github.com/endojs/endo/pull/1712 to make it into a genuine
- * guard that is distinct from a copyRecord
+ * @typedef {CopyTagged & {
+ *   [Symbol.toStringTag]: 'guard:interfaceGuard',
+ *   payload: InterfaceGuardPayload,
+ * }} InterfaceGuard
```
michaelfig, line 515: I would somewhat prefer a definition of CopyTagged in `packages/pass-style/src/types.js` as a parameterized type:
```js
/**
 * @template {string} [Tag=string]
 * @template {Passable} [Payload=Passable]
 * @typedef {{
 *   [Symbol.toStringTag]: Tag,
 *   payload: Payload,
 *   [passStyle: symbol]: 'tagged' | Tag,
 * }} CopyTagged
 */
```

And then using here as:
```suggestion
 * @typedef {CopyTagged<'guard:InterfaceGuard', InterfaceGuardPayload>} InterfaceGuard
```

And ditto, for the other uses of CopyTagged.

michaelfig (approved):
Approval on the condition that you clearly decide whether to parameterize CopyTagged, or to leave that out of this PR.

packages/patterns/src/types.js
erights: Trying now. But to clarify, changing the `[passStyle: symbol]` property to have a `| Tag` instead of a `| string` is wrong. Ideally it should have neither, since the only correct value is `'tagged'`. However, I cannot figure out how to do that without ts complaint. The `| string` seems to appease the checker while being least misleading of the options I tried. The comments says

```js
/**
 * ...
 * It must have a property with key equal to the `PASS_STYLE` export and value 'tagged'
 * and no other properties except `[Symbol.toStringTag]` and `payload`,
 * but TypeScript complains about a declaration like `[PASS_STYLE]: 'tagged'`
 * because importing packages do not know what `PASS_STYLE` is
 * so we appease it with a looser but less accurate definition
 * using symbol index properties.
 */
```

If you could help me figure out how to make this more accurate while still appeasing ts, that would be great, thanks.

erights: I like this *much* better, thanks!

Regarding the `| string`, while I would still like to clean this up, there's no reason to do so in this PR. Instead, I just clarified this comment and added a TODO.

Done

PR #1713, opened 2023-08-08
fix(bundle-source): more typing and export refinements

michaelfig:
refs: #1708

## Description

- Don't use ambient types
- Improve typing of `BundleSource` to reflect the actual defaults and overloads permitted by the code

### Security Considerations

n/a

### Scaling Considerations

n/a

n/a

### Testing Considerations

Manually tested against Agoric SDK and produced the expected bundle types.

### Upgrade Considerations

n/a


packages/bundle-source/src/types.js
```diff
@@ -1,62 +1,65 @@
...
+ * @typedef {'endoZipBase64' | 'nestedEvaluate' | 'getExport'} ModuleFormat
+ */
+
 /**
- * @typedef {'getExport' | 'nestedEvaluate' | 'endoZipBase64'} ModuleFormat
+ * @typedef { &
+ *  BundleSourceSimple &
+ *  BundleSourceWithFormat &
+ *  BundleSourceWithOptions &
+ *  BundleSourceFallback} BundleSource
```
michaelfig, line 13: It may not be obvious to reviewers, but the order of these intersections matters, insofar as Typescript treats the last one as the "most generic" version of the overloads.  That annoyed me for quite a while.

Using the generics was the best way I could find to produce the correct result type regardless of which overload you used.

packages/bundle-source/exported.js
```diff
@@ -1 +1,2 @@
-import './src/types.js';
+/* eslint-disable import/export */
```
turadg, line 1: we should try to avoid lint suppressions. How do we make the rule not complain about this case? Maybe disable it for `exported.js`.

turadg: I see `exports.js` has all the types via `exports.d.ts`. is this file just here for backwards compatibility? Please leave a comment explanation of why they export the same.

packages/bundle-source/src/index.js
```diff
@@ -547,7 +545,7 @@ ${sourceMap}`;
   throw Error(`unrecognized moduleFormat ${moduleFormat}`);
 }
 
-/** @type {BundleSource} */
+/** @type {import('./types.js').BundleSource} */
```
turadg, line 548: I know I'm in the minority on this but I'd really be fine with,
```suggestion
/** @type {import('./types').BundleSource} */
```

packages/bundle-source/src/types.js
turadg: Thanks for the important note. Mind including as a `//` comment?

packages/bundle-source/src/types.js
```diff
@@ -1,62 +1,65 @@
...
  */
 
 /**
- * @typedef {BundleSourceEndoZipBase64 & BundleSourceGetExport & BundleSourceNestedEvaluate} BundleSource
+ * @template {ModuleFormat} T
+ * @typedef {T extends 'endoZipBase64' ? {
```
turadg, line 18: 👏

turadg (approved):
All looks like an improvement to me.

The only thing I'd ask for is types regression tests, but not a blocker.

packages/bundle-source/exported.js
michaelfig: Yes, just for backwards compat.

michaelfig: I don't see how to disable it for `exported.js` unless we turn it off entirely.  Unfortunately, it's a rule that encompasses a buch of things we probably *do* want to enforce, so I think having a suppression is much clearer.  Comment added.

mhofman (commented):
I'm confused by this change.

`exports.js` doesn't seem to be referenced anywhere.

The semantics of `exported.js` is changing to no longer pull ambient types. Why are we keeping that file around, or exporting types through it?

I thought our decision was to export types through the entrypoint of the package?

packages/bundle-source/src/index.js
```diff
@@ -17,8 +17,6 @@ import { encodeBase64 } from '@endo/base64';
 import SourceMaps from 'source-map';
 import { whereEndoCache } from '@endo/where';
```
mhofman, line 19:
```suggestion

export * from './exports.js';
```

packages/bundle-source/src/index.js
kriskowal: I think I see your point. TypeScript is treating all the files with the same base name as a compilation unit.

packages/bundle-source/src/types.js
kriskowal: 🤯

michaelfig: 🤣

packages/bundle-source/src/index.js
michaelfig: I moved the guts of this file to `src/bundle-source.js`, leaving `src/index.js` to export the types and the implementation.

packages/bundle-source/src/index.js
michaelfig: I used `import('./types')` in eventual-send, so I'm amenable to doing it here as well... and done.

michaelfig:
> The only other thing I'd ask for is types regression tests, but not a blocker.

I toyed with regression tests, but had utterly no luck getting type assertions to report in my IDE (Typescript LSP in VSCode).  I suspect something in the `jsconfig.json` needs to change, or in a `tsconfig.json` for that matter.  I hope sometime we can converge on best-practices for `tsd` or something like @SMotaal's anno-type.

Issue #1703, opened 2023-07-28
provide an import for members of the `assert` global


turadg:
## What is the Problem Being Solved?

When coding in an IDE, it's common to begin typing a function and rely on the IDE to autocomplete it. Most IDEs are smart enough to even import the function if it's not in scope. But they don't know to create a local binding by destructuring a property of a certain object that is in global scope, such as functions on the `assert` global.

E.g. if I type `Fail` in my IDE and ask it to resolve that, it doesn't know how to find it unless there's a module export.

We have such a module in agoric-sdk but it belongs in Endo. https://github.com/Agoric/agoric-sdk/issues/5672 . Having them separated creates extra maintenance burden such as https://github.com/Agoric/agoric-sdk/issues/8005 ) 

## Description of the Design

Publish a module that re-exports the `assert` global (type Assert) so that it can be auto-imported in IDEs.

PR #1717, opened 2023-08-10
'assert' package

turadg:
closes: #1703

## Description

A new package to provide the properties of the `assert` global such that they can be auto-imported in an IDE.

### Security Considerations

If this dependency were compromised it could lead a user to thinking they're using the SES's assertions when they're not.

### Scaling Considerations

n/a

n/a

### Testing Considerations

New tests. Lightly covered yet since the idea is that eventually this package becomes the shim.

### Upgrade Considerations

n/a


packages/assert/index.js
```diff
@@ -0,0 +1,78 @@
...
+  Fail,
+  makeAssert,
+  note,
+  quote,
+  // properties with syntax collision
+  assertString,
+  assertTypeof,
+  // convenient short names
+  details as X,
+  quote as q,
```
kriskowal, line 77: My preference is to export the full name and alias at the import site.

kriskowal: My rationale is that abbreviations are for local ergonomics and the exported names need to communicate their purpose. A reviewer unfamiliar with the assert module should be able to infer the meaning from usage and name without leaving the file.

kriskowal (changes requested):

packages/assert/index.js
turadg: I agree they don't belong in Endo. I'll take them out.

I don't necessarily agree the aliasing should happen at the call site. To make imports easy a repo might have a utils package that re-exports them under abbreviations that are normal in the repo.

kriskowal (approved):

packages/assert/index.js
kriskowal: Thanks. Good enough for this change. We can converge on the broader tension between legibility and ergonomics elsewhere.

packages/assert/README.md
gibson042, line 1: Empty README. :sob:

packages/assert/index.js
```diff
@@ -0,0 +1,73 @@
...
+  details,
+  equal,
+  error,
+  Fail,
+  makeAssert,
+  note,
+  quote,
+  // properties with syntax collision
+  assertString,
+  assertTypeof,
```
gibson042, line 72: Why rename the exports?
```suggestion
  assertString as string,
  assertTypeof as typeof,
```

turadg: I figured it saves the importer from aliasing them. I'm open to making the change you suggest. I could add a README at that point. Also be my guest if you like to.

kriskowal: I do not know how I missed this, and I have no strong feelings one way or the other.

kriskowal: Actually, yes, I do feel strongly that all these names should be consistently exported either `assertX` or `x` to favor either the `import * as assert; assert.x()` usage or the `import { assert, assertX, assertY }` usage.

kriskowal: More broadly, these are the design families:

1.
    ```js
    import { assert, assertTypeof, assertString, assertEtc } from '@endo/assert';
    assert();
    assertTypeof();
    assertString();
    ```
   Vote 🎉
2.
    ```js
    import { assert } from '@endo/assert';
    assert();
    assert.typeof();
    assert.string();
    assert.etc();
    ```
    Vote ❤️
3.
    ```js
    import * as assert from '@endo/assert';
    assert.ok(); // unprecedented
    assert.string();
    assert.typeof();
    assert.etc();
    ```
    Vote 🚀

My suspicion is that we disfavor option 3(🚀) collectively. @erights’s position seems consistent with blessing one of these forms and herding usage. @turadg’s motivation for the creation of this package is having an anchor for import inference which is probably satisfied by any of these forms.

I believe that @erights’s position of having a single consistent form actually implies that option 1(🎉) should preclude 2(❤️) by exporting an `assert` function that unlike `globalThis.assert` _lacks_ the `assert.xyz()` properties.

dckc: my preference is 2 :heart: cuz it seems the least different from existing code. Changing `assert.typeof` to `assertTypeof` looks like a bunch of churn and I haven't seen benefit that comes anywhere near the cost.

michaelfig: I would add something like 2b: `import { assert, fail as Fail, details as X } from '@endo/assert';` such that the template literal tags aren't stuffed onto the `assert` export.  I've been ever annoyed by the uppercasing of `assert.Fail` just to avoid the conflict with `assert.fail`.

Chris-Hibbert: I'm confused. The uses of assert that I'm interested in seeing made consistent are `quote`, `bare`, `details` and `Fail`. Are these not included in this assert package? Where will they be instead?

kriskowal: As @michaelfig points out, `assert.etc()` and `assertEtc()` are standing in for the rest of the exports, and whether to partition names that pertain to assertions from names that pertain to message encoding is an orthogonal concern not represented in this loose poll. I like @michaelfig’s suggestion.

turadg: I also support 2b. I think this formulation is compatible:
- `assert` is exported top-level
- functions that assert are not exported top-level (they're accessible as `assert.string`, etc)
- functions on the `globalThis.assert` are exported top level

We also need to answer: are the non-asserting exports omitted from the exported `assert`? E.g. is this allowed…
```
import { assert, Fail } from `@endo/assert';
Fail`badness`;
assert.Fail`badness`;
```
I've always been troubled by the `globalThis.assert` having functions that don't assert. This is an opportunity to clean it up. So I'd propose that the exported `assert` has only assertion functions.

erights: For the third bullet, did you mean
> * functions on the globalThis.assert that do not assert are exported top level

?

If so, then I like and agree with all of this, including omitting these non-asserting functions from the exported `assert`. Unfortunate that this makes the exported `assert` !== `globalThis.assert`, but I agree that on balance it is worth it.

This is not strictly compat with existing code. Any idea how painful the transition? Or would we keep `@agoric/assert` with its current behavior but deprecated until we can delete it?

kriskowal: I agree, @turadg. `@endo/assert` should export an `assert` function that owns only assertion methods. All remaining API of the `globalThis.assert` should be be exported separately.

I believe @michaelfig’s stated preference is to distinguish `assert.fail()` from `fail` and not distinguish them by case.
```js
import { assert, fail } from '@endo/assert';
assert.fail('badness'); // and
fail`badness`
```

I do not have any preference, but acknowledge that the difference in capitalization was only necessary because we used `globalThis.assert` as a namespace that contained both and _something_ had to be different about the names. Not sure why we didn’t borrow from Perl…oh, no, it itches! They’re everṭ̶̽h̷̼͌e̶͖͋ ̵̖̅d̷͚̊e̴̩͝ŵ̸̗ȩ̶̓o̷̲͘m̵͕̆e̵̡̅r̷̝̆s̴̯̒,̶̙̈́ ̴̠̋T̸̰͛H̶̲͊Ȅ̸͖ ̴̠̏P̸̩͋Á̴͕I̸͖̕N̵̬͆

michaelfig: > This is not strictly compat with existing code.

I've been operating under the assumption that `@agoric/XXX` is our best-shot, needed for production version of a module.  When we introduce `@endo/XXX`, we have no need to make them compatible.  And then the goal is to move features from `@agoric/XXX`, clean them up, and put them into `@endo/XXX`, until we can remove `@agoric/XXX`.

This resembles the typical open-source "fork the parent, work without constraints on your fork, then try to clean up your changes and get them upstreamed into the parent."

erights: Please let us not rename `Fail` to `fail`. I like keeping template literal tags visually very distinct, like `X` and `Fail`.

That said, I agree that the exported `assert` should continue to have an `assert.fail` with its current meaning, and to omit `assert.Fail`.

michaelfig: > I like keeping template literal tags visually very distinct, like `X` and `Fail`.

If we're writing a library that we want other JS devs to adopt, I think they will be struck by the capital letters that aren't referring to constructors/classes.  Rather than using the tersest, most opaque, and least conventional names in the exports, why not just do something like:

```
import { errorMessage as X, throwError as Fail } from '@endo/assert';
```

in your own code?  That preserves the convention you like, while still making it clear to the reader what you are expressing.

erights: Have we resolved that we're doing likewise with
`import { quote as q, bare as b, something as X }`
with no separate renaming reexport anywhere?

I'm not in love with `errorMessage` and `throwError`, though I don't hate them either. But if we're agreed we're pervasively renaming to get to `q`, `b`, and `X`, I have no objection to adding
`import { something as Fail }`. 
If we're all agreed on this much, then lets proceed to bikeshed the two unresolved somethings.

kriskowal: My mind appears to be entangled at a quantum level with @michaelfig. I prefer to avoid abbreviation in public API and to scope abbreviations lexically when that improves local legibility. I propose `redact` for `X` and `throwRedact` for `Fail`.

erights: I like `redact` and `throwRedact` !!

Chris-Hibbert: Not that I like making names longer, but how about `throwRedacted`?

(and `redacted` if that seems more consistent  to y'all.)

erights: After sleeping on it, I do like `redacted` and `throwRedacted` better.

PR #1718, opened 2023-08-10
fix(ses): align with XS property censorship agreement

### LINKED ISSUE #910: {[inert function]} from toString() for Math.random etc.

erights:
Fixes #910 

@phoddie, I seem unable to add you as an official reviewer, but please consider yourself one anyway. I'm eager for your comments.

Technically this is a compat break, which I currently mark with a `fix(ses)!:`. Reviewers, please let me know whether you consider this change a compat break, or whether I should demote it to `fix(ses):`. Thanks.

### DIFF
--- a/packages/ses/src/permits.js
+++ b/packages/ses/src/permits.js
@@ -326,7 +326,7 @@ function TypedArrayPrototype(constructor) {
 }
 
 // Without Math.random
-const SharedMath = {
+const CommonMath = {
   E: 'number',
   LN10: 'number',
   LN2: 'number',
@@ -649,12 +649,16 @@ export const permitted = {
   },
 
   '%InitialMath%': {
-    ...SharedMath,
-    // random is standard but omitted from SharedMath
+    ...CommonMath,
+    // `%InitialMath%.random()` has the standard unsafe behavior
     random: fn,
   },
 
-  '%SharedMath%': SharedMath,
+  '%SharedMath%': {
+    ...CommonMath,
+    // `%SharedMath%.random()` is tamed to always throw
+    random: fn,
+  },
 
   '%InitialDate%': {
     // Properties of the Date Constructor
@@ -668,6 +672,7 @@ export const permitted = {
   '%SharedDate%': {
     // Properties of the Date Constructor
     '[[Proto]]': '%FunctionPrototype%',
+    // `%SharedDate%.now()` is tamed to always throw
     now: fn,
     parse: fn,
     prototype: '%DatePrototype%',
--- a/packages/ses/src/tame-date-constructor.js
+++ b/packages/ses/src/tame-date-constructor.js
@@ -17,26 +17,38 @@ export default function tameDateConstructor(dateTaming = 'safe') {
 
   // Use concise methods to obtain named functions without constructors.
   const tamedMethods = {
+    /**
+     * `%SharedDate%.now()` throw a `TypeError` starting with "secure mode".
+     * See https://github.com/endojs/endo/issues/910#issuecomment-1581855420
+     */
     now() {
-      return NaN;
+      throw TypeError('secure mode Calling %SharedDate%.now() throws');
     },
   };
 
-  // Tame the Date constructor.
-  // Common behavior
-  //   * new Date(x) coerces x into a number and then returns a Date
-  //     for that number of millis since the epoch
-  //   * new Date(NaN) returns a Date object which stringifies to
-  //     'Invalid Date'
-  //   * new Date(undefined) returns a Date object which stringifies to
-  //     'Invalid Date'
-  // OriginalDate (normal standard) behavior
-  //   * Date(anything) gives a string with the current time
-  //   * new Date() returns the current time, as a Date object
-  // SharedDate behavior
-  //   * Date(anything) returned 'Invalid Date'
-  //   * new Date() returns a Date object which stringifies to
-  //     'Invalid Date'
+  /**
+   * Tame the Date constructor.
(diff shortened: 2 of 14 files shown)


packages/ses/src/tame-date-constructor.js
```diff
@@ -17,26 +17,38 @@ export default function tameDateConstructor(dateTaming = 'safe') {
 
   // Use concise methods to obtain named functions without constructors.
   const tamedMethods = {
+    /**
+     * `%SharedDate%.now()` throw a `TypeError` starting with "secure mode".
+     * See https://github.com/endojs/endo/issues/910#issuecomment-1581855420
+     */
     now() {
-      return NaN;
+      throw TypeError('secure mode Calling %SharedDate%.now() throws');
```
kriskowal, line 25: This is an improvement.

packages/ses/src/tame-date-constructor.js
```diff
@@ -17,26 +17,38 @@ export default function tameDateConstructor(dateTaming = 'safe') {
...
+   *     'Invalid Date'
+   *
+   * OriginalDate (normal standard) behavior preserved by
+   * `%InitialDate%`.
+   *   * `Date(anything)` gives a string with the current time
+   *   * `new Date()` returns the current time, as a Date object
+   *
+   * `%SharedDate%` behavior
+   *   * `Date(anything)` throws a TypeError starting with "secure mode"
+   *   * `new Date()` throws a TypeError starting with "secure mode"
```
kriskowal, line 48: Seems like a missed opportunity.
- Error: No time
- Error: No time for you
- Error: Too late
- Error: Time is an illusion
- Error: Time is a flat circle
- Error: Time flies like an arrow, fruit flies like a banana

packages/ses/src/tame-date-constructor.js
```diff
@@ -69,13 +85,13 @@ export default function tameDateConstructor(dateTaming = 'safe') {
         configurable: false,
       },
       parse: {
-        value: Date.parse,
+        value: OriginalDate.parse,
```
kriskowal, line 88: There is no original date. The day after New Year’s Eve, 1 BCE was New Year’s Day, 1 CE.

packages/ses/src/permits.js
```diff
@@ -649,12 +649,16 @@ export const permitted = {
...
   '%InitialMath%': {
-    ...SharedMath,
-    // random is standard but omitted from SharedMath
+    ...CommonMath,
+    // `%InitialMath%.random()` has the standard unsafe behavior
     random: fn,
   },
 
-  '%SharedMath%': SharedMath,
+  '%SharedMath%': {
```
kriskowal, line 657: Or, as I like to call it, `%NewMath%`.

kriskowal (changes requested):
This is a breaking change to the extent that anyone relied on the old behavior. My feeling is that the old behavior sufficiently deterred relying upon the old behavior that this is not a breaking change. But, if you wish to play it safe, we can bump the 0.minor version and you’ll need a conventional commit to that effect.

Otherwise, this just needs an entry in `NEWS.md` under `# Next release` and an update to the docs, particularly the doc that outlines the differences in the Hardened JavaScript realm and compartment.

erights:
> This is a breaking change to the extent that anyone relied on the old behavior. My feeling is that the old behavior sufficiently deterred relying upon the old behavior that this is not a breaking change. But, if you wish to play it safe, we can bump the 0.minor version and you’ll need a conventional commit to that effect.

Based on this, I just changed from `fix(ses)!:` to `fix(ses):`. Do I understand correctly that this is consistent with your answer above?

packages/ses/src/tame-date-constructor.js
erights: What about the big bang? Or the Unix Epoch?

erights:
> Otherwise, this just needs an entry in `NEWS.md` under `# Next release` and an update to the docs, particularly the doc that outlines the differences in the Hardened JavaScript realm and compartment.

There are indeed stale *.md files that need updating. Working on it. By "docs", do you mean anything besides *.md files in the endo repo?

kriskowal:
> > Otherwise, this just needs an entry in `NEWS.md` under `# Next release` and an update to the docs, particularly the doc that outlines the differences in the Hardened JavaScript realm and compartment.
> 
> There are indeed stale *.md files that need updating. Working on it. By "docs", do you mean anything besides *.md files in the endo repo?

Correct, just *.md in the ses package. I think just docs/reference.md and docs/guide.md. There are other stale docs that are out of scope.

kriskowal:
> > This is a breaking change to the extent that anyone relied on the old behavior. My feeling is that the old behavior sufficiently deterred relying upon the old behavior that this is not a breaking change. But, if you wish to play it safe, we can bump the 0.minor version and you’ll need a conventional commit to that effect.
> 
> Based on this, I just changed from `fix(ses)!:` to `fix(ses):`. Do I understand correctly that this is consistent with your answer above?

Yes, we can treat this as non-breaking, unless you can think of a realistic working program that would break because of these changes.

erights:
> Correct, just *.md in the ses package. I think just docs/reference.md and docs/guide.md. There are other stale docs that are out of scope.

Done. The second has these. A lot more than either of us expected. PTAL.

kriskowal (approved):

Issue #1716, opened 2023-08-10
bundle-source incompatible with pnpm


turadg:
## Describe the bug

Installing the "bundle-source" package using [pnpm](https://pnpm.io/) fails.

## Steps to reproduce

1. `echo '{  "name": "depend-on-bundle-source",   "dependencies": {    "@endo/bundle-source": "^2.7.0"  } }' > package.json`
2. `pnpm install`
3. witness failure as such,

```
│ > rollup@2.70.1-endo.1 prepare
│ > husky install && npm run build
│ fatal: not a git repository (or any of the parent directories): .git
│ > rollup@2.70.1-endo.1 build
│ > shx rm -rf dist && node scripts/update-git-commit.js && rollup --config rollup.config.ts --configPlugin typescript && shx cp src/rollup/types.d.ts dist/rollup.d.ts && shx chmod a+x dist/bin/rollup
│ node:internal/modules/cjs/loader:998
│   throw err;
│   ^
│ Error: Cannot find module '/Users/turadg/Library/pnpm/store/v3/tmp/_tmp_50811_4fd916dad5ed14e3199246b9948a1f03/scripts/update-git-commit.js'
│     at Function.Module._resolveFilename (node:internal/modules/cjs/loader:995:15)
│     at Function.Module._load (node:internal/modules/cjs/loader:841:27)
│     at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
│     at node:internal/main/run_main_module:22:47 {
│   code: 'MODULE_NOT_FOUND',
│   requireStack: []
│ }
```

## Expected behavior

Package can be installed using pnpm. 

## Platform environment

PNPM 8.6.12
Node 16.19.0

## Additional context

Problem may also manifest with Yarn Berry (v2+) when using hard links like pnpm does.

PR #1719, opened 2023-08-12
bump rollup

turadg:
closes: #1716 

## Description

There was some problem with `magic-string` dependency in Rollup, which was circumvented by making a patched copy of Rollup at https://github.com/endojs/endo/tree/rollup-2.7.1-patch-1 and having `bundle-source` depend on that. That caused a problem for PNPM installs (#1716 ).

This PR updates Rollup to the latest non-breaking version, which is confirmed to have the version of magic-string that the patch file says is sufficient.

It also bumps `yarn` which had printed a warning while I was doing the other change.

### Testing Considerations

CI


.yarn/releases/yarn-1.22.19.js
michaelfig, line 1: Why not delete the old version from `.yarn/releases/yarn-1.22.0.js`?

packages/bundle-source/package.json
```diff
@@ -39,7 +39,7 @@
     "@rollup/plugin-node-resolve": "^13.0.0",
     "acorn": "^8.2.4",
     "jessie.js": "^0.3.2",
-    "rollup": "endojs/endo#rollup-2.7.1-patch-1",
+    "rollup": "^2.77.4-1",
```
michaelfig, line 42: It's a little strange to leave the `-1` suffix.  Is there a reason why you did that?

```suggestion
    "rollup": "^2.77.4",
```

yarn.lock
```diff
@@ -10625,9 +10625,10 @@ rollup@1.31.0:
     "@types/node" "*"
     acorn "^7.1.0"
 
-rollup@endojs/endo#rollup-2.7.1-patch-1:
-  version "2.70.1-endo.1"
-  resolved "https://codeload.github.com/endojs/endo/tar.gz/54060e784a4dbe77b6692f17344f4d84a198530d"
+rollup@^2.77.4-1:
+  version "2.79.1"
```
michaelfig, line 10629: Did you mean to resolve to 2.79?  If not, maybe `rollup@~2.77.4` is what you meant?

.yarn/releases/yarn-1.22.19.js
turadg: I wasn't sure if someone might get an error if their resolution was stale somehow. And it's not really in the way.

I'll delete it since I have to refresh this anyway.

packages/bundle-source/package.json
turadg: It was the latest version according to whatever I looked up, but the yarn.lock shows a newer version so I'll change this to that.

michaelfig (approved):

PR #1721, opened 2023-08-13
pattern types

turadg:
refs: #2392

## Description

Make `mustMatch` assert type implied by matcher.

TODO: rebase onto master. I have a working branch for that but it needs more work before overwriting this remote.


packages/pass-style/src/types.js
```diff
@@ -96,8 +96,9 @@ export {};
  */
 
 /**
+ * @template {string} [T=string] tag
  * @typedef {{
- *   [Symbol.toStringTag]: string,
+ *   [Symbol.toStringTag]: T,
  *   payload: Passable,
  *   [passStyle: symbol]: 'tagged' | string,
  * }} CopyTagged
```
gibson042, line 104: Won't this equate CopyTagged with CopyTagged\<string>, which is novel and unexpected because "string" is not a known tag?

packages/pass-style/test/test-passStyleOf.js
```diff
@@ -41,7 +41,7 @@ test('some passStyleOf rejections', t => {
     message:
       /Only registered symbols or well-known symbols are passable: "\[Symbol\(unique\)\]"/,
   });
-  if (harden.isFake) {
+  if ('isFake' in harden) {
```
gibson042, line 44: What motivates these changes?

turadg: the line started erroring because `harden` doesn't have an `isFake` property. I figure something in my changes fixed a resolution that had been falling back to `any`

packages/pass-style/src/types.js
turadg: hmm, I don't think `CopyTagged<string>` would satisfy the template because the parameter wouldn't be a string itself. It would have to by `CopyTagged<'string'>`. That's not a known tag but it was valid before, right? (i.e. not a regression).

turadg:
Done in #3133

PR #1725, opened 2023-08-16
feat(exo): unprotected exos

erights:
@michaelfig , does this do what you need at this stage, i.e., without yet addressing the zone-POLA issue?

Not at all yet tested.

### DIFF
--- a/packages/patterns/src/types.js
+++ b/packages/patterns/src/types.js
@@ -480,10 +480,16 @@ export {};
  * @typedef {{
  * <M extends Record<any, MethodGuard>>(interfaceName: string,
  *             methodGuards: M,
- *             options?: {sloppy?: false}): InterfaceGuard<M>;
+ *             options?: {
+ *     sloppy?: false,
+ *     unprotected?: boolean,
+ *   }): InterfaceGuard<M>;
  * (interfaceName: string,
  *             methodGuards: any,
- *             options?: {sloppy?: true}): InterfaceGuard<any>;
+ *             options?: {
+ *     sloppy?: true,
+ *     unprotected?: boolean,
+ *   }): InterfaceGuard<any>;
  * }} MakeInterfaceGuard
  */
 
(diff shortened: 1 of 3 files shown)


packages/patterns/src/types.js
```diff
@@ -480,10 +480,16 @@ export {};
  * @typedef {{
  * <M extends Record<any, MethodGuard>>(interfaceName: string,
  *             methodGuards: M,
- *             options?: {sloppy?: false}): InterfaceGuard<M>;
+ *             options?: {
+ *     sloppy?: false,
```
erights, line 484: @turadg , since I'm unsure why you split these cases on `sloppy` true vs false, I may be doing the wrong thing in not splitting on unprotected true vs false. Truly there are three cases: (false,false), (true,false), and (false,true). If I had not already introduced a boolean option, I would have made it a three-way. Oh well.

erights:
Successor to https://github.com/endojs/endo/pull/1724

packages/patterns/src/types.js
turadg: The reason for the overloading here is that `sloppy:true` implies `InterfaceGuard<any>` (everything matches) and `sloppy:false` parameterizes InterfaceGuard with the type of `methodGuards`.

I take it that "unprotected" should be `InterfaceGuard<any>` so this type needs more work. But let's settle on the API first.

> I would have made it a three-way

It's not too late! We could make it `sloppiness: `none' | 'sloppy' | 'unprotected'` and still retain backwards compatibility with `true` ('sloppy') and `false` ('none')

turadg (changes requested):

erights:
Converted to Draft in anticipation of successor from @michaelfig . Once that's pushed, I'll close this one.

michaelfig:
In experimenting with this API, I've come to feel that `unprotected` should be even more granular (explicitly annotating what is unprotected).  I think we should flesh that out before deciding whether to implement `defaultGuards` after all, because it may be orthogonal to `sloppy: true`.

So, I've created #1772.  Please take a look.

PR #1729, opened 2023-08-22
clean up assert exports

turadg:
refs: #1717

## Description

Fewer exports per https://github.com/endojs/endo/pull/1717#discussion_r1295130523
Clearer names per https://github.com/endojs/endo/pull/1717#discussion_r1297908833

These are breaking changes to master, but not to what's been published.

### DIFF
--- a/packages/assert/index.js
+++ b/packages/assert/index.js
@@ -42,32 +42,17 @@ if (missing.length > 0) {
   );
 }
 
-const {
-  bare,
-  details,
-  equal,
-  error,
-  Fail,
-  note,
-  quote,
-  makeAssert,
-  string: assertString,
-  typeof: assertTypeof,
-} = globalAssert;
+const { bare, details, error, Fail, note, quote, makeAssert } = globalAssert;
 
 export {
-  // the global
+  // the global, with assertions
   globalAssert as assert,
-  // properties
+  // related utilities that aren't assertions
   bare,
   details,
-  equal,
   error,
   Fail,
   makeAssert,
   note,
   quote,
-  // properties with syntax collision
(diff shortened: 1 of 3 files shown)


packages/assert/index.js
```diff
@@ -42,32 +42,18 @@ if (missing.length > 0) {
...
-  equal,
-  error,
-  Fail,
-  note,
-  quote,
-  makeAssert,
-  string: assertString,
-  typeof: assertTypeof,
-} = globalAssert;
+const { bare, details, error, Fail, note, quote, makeAssert } = globalAssert;
```
erights, line 45: I thought part of the consensus was to have the `assert` exported by this package contain only the assertions, i.e., the methods not separately exported by name.

```suggestion
const { bare, details, error, Fail, note, quote, makeAssert, ...assertions } = globalAssert;

Object.freeze(assertions);

export { assertions as assert };
```

erights (changes requested):

packages/assert/index.js
turadg: Thanks, I forgot to include that change.

kriskowal: And the exported `assert` needs to be callable as well.

```js
const { bare, details: redacted, error, Fail: throwRedacted, note, quote,  makeAssert, ...assertions } = globalAssert;
const assert = (...args) => globalAssert(...args);
Object.assign(assert, assertions);
export { assert, redacted, throwRedacted, quote, bare, note, makeAssert };
```

kriskowal (approved):

erights (approved):
LGTM, thanks!

PR #1730, opened 2023-08-23
fix(assert): mistyped assert.fail

erights:
`assert.fail` was correctly on the newly exported `assert` object and not in the exported utility functions. But the types seemed to imply otherwise.


turadg:
Agree that the types and implementation are out of sync, but I'm not sure the which should change.

In my mind,
- `fail` and `throwRedacted` should be in the same place (both take a message and throw an error)
- neither should be on `assert` because they don't make assertions _about_ anything

erights:
Agree that it is not clear which way to resolve this, and that your points for moving `fail` into the exported utility functions make sense. Counter points

   * Even though this assert package never mentions the term `Fail` in anything it exports, we will always rename `throwRedacted` to `Fail` on import. Putting `fail` and `Fail` into the same namespace still seems unpleasant. This package, where we separate out the assertions on `assert` from the exported assert utility functions is our opportunity to put them in separate namespaces.
   * `fail` is kinda the degenerate case of the various assertion functions: `assert.string` -> `assert.typeof` -> `assert` -> `assert.fail`. Not compelling by itself, but an adequate rationalization given the first bullet.

erights:
In light of internal discussions that did not resolve the question above, I added a commit that 
   * withdraws `makeAssert`, as we have not found it useful in practice, and mostly have not used it.
   * renames the exported `error` utility function to `makeError`, which is much more understandable as a standalone function (and which actually happens to undo the internal rename of this function in ses' `assert.js` module).

I kept this as a separate commit so that we could easily back out of these additional changes if we decide to before merging this PR.

erights:
This new commit is neutral on the earlier open question.

kriskowal (approved):
I’ll note that we can try to gradually converge `ses` closer to this package.

I’m amenable to additionally renaming the package `@endo/errors`.

packages/ses/types.d.ts
```diff
@@ -203,10 +203,15 @@ export interface AssertionUtilities {
   Fail(template: TemplateStringsArray | string[], ...args: any): never;
   quote(payload: any, spaces?: string | number): ToStringable;
   bare(payload: any, spaces?: string | number): ToStringable;
+}
+
+export interface DeprecatedAssertionUtilities {
```
turadg, line 208: Nicely done

turadg (approved):
+1 to renaming the package

I'm on the fence about 'fail' location but I defer

packages/ses/types.d.ts
```diff
@@ -186,6 +186,7 @@ export interface AssertionFunctions {
     errorConstructor?: ErrorConstructor,
   ): void;
   string(specimen: any, details?: Details): asserts specimen is string;
+  fail(details?: Details, errorConstructor?: ErrorConstructor): never;
```
kriskowal, line 189: This is consistent with my preference.

erights:
> I’m amenable to additionally renaming the package `@endo/errors`.

> +1 to renaming the package

Done

turadg:
FYI, it's confusing to have the package rename in the commit history as `[fix(assert): mistyped assert.fail](https://github.com/endojs/endo/pull/1730/commits/e1ebe75845e21470b2b732a6417d35c4106df6b8)` I think it should have been another commit.

And more generally to me it's a reason to move away from producing changelogs from commit history.

PR #1731, opened 2023-08-23
feat(ses): Freeze evaluators, Compartment constructor and Symbol

weizman:
## Description

This is a small PR to enhance security by:
1. Freezing some Compartment-endowed capabilities that aren't currently frozen due to them being non-identical instances of their original version from the start Compartment
	* Being:
		* evaluators: `Compartment`, `eval` and `Function`
2. Hardening some start-Compartment capabilities that aren't currently hardened
	*  Being:
		* evaluators: `Compartment`, `eval` and `Function`
		* `initialGlobalPropertyNames`: `Date`, `Error`, `RegExp`, `Math`, (?`getStackString`)
			* Also `Symbol` although not belonging officially to the `initialGlobalPropertyNames` list

### Security Considerations

Not as far as I can tell

### Scaling Considerations

Not as far as I can tell

* Some Intrinsics are now being hardened on the globalThis of the start Compartment, meaning this PR isn't backward compatible with programs that attempt to modify their behaviour by redefining properties of them
	* See [section #2](https://github.com/endojs/endo/pull/1731#Description) for the list
* Some Intrinsics are now being hardened on the globalThis of each new Compartment, meaning this PR isn't backward compatible with programs that attempt to modify their behaviour by redefining properties of them
	* See [section #1](https://github.com/endojs/endo/pull/1731#Description) for the list

### Testing Considerations

Added tests

### Upgrade Considerations

Not as far as I can tell

### DIFF
--- a/packages/ses/src/lockdown.js
+++ b/packages/ses/src/lockdown.js
@@ -25,6 +25,7 @@ import {
   ownKeys,
   stringSplit,
   noEvalEvaluate,
+  getOwnPropertyNames,
 } from './commons.js';
 import { makeHardener } from './make-hardener.js';
 import { makeIntrinsicsCollector } from './intrinsics.js';
@@ -384,6 +385,18 @@ export const repairIntrinsics = (options = {}) => {
     // Finally register and optionally freeze all the intrinsics. This
     // must be the operation that modifies the intrinsics.
     tamedHarden(intrinsics);
+
+    // Harden evaluators
+    tamedHarden(globalThis.Function);
+    tamedHarden(globalThis.eval);
+    // @ts-ignore Compartment does exist on globalThis
+    tamedHarden(globalThis.Compartment);
+
+    // Harden Symbol and properties for initialGlobalPropertyNames in the host realm
+    tamedHarden(globalThis.Symbol);
+    for (const prop of getOwnPropertyNames(initialGlobalPropertyNames)) {
+      tamedHarden(globalThis[prop]);
+    }
 
     return tamedHarden;
   };
--- a/packages/ses/test/test-compartment-instance.js
+++ b/packages/ses/test/test-compartment-instance.js
@@ -2,9 +2,15 @@ import test from 'ava';
 import '../index.js';
 
 test('Compartment instance', t => {
-  t.plan(9);
+  t.plan(12);
 
   const c = new Compartment();
+
+  t.truthy(c.globalThis.eval && Object.isFrozen(c.globalThis.eval));
+  t.truthy(c.globalThis.Function && Object.isFrozen(c.globalThis.Function));
+  t.truthy(
+    c.globalThis.Compartment && Object.isFrozen(c.globalThis.Compartment),
+  );
 
   t.is(typeof c, 'object', 'typeof');
   t.truthy(c instanceof Compartment, 'instanceof');
--- a/packages/ses/test/test-repair-intrinsics.js
+++ b/packages/ses/test/test-repair-intrinsics.js
@@ -3,6 +3,8 @@
 
 import test from 'ava';
 import { repairIntrinsics } from '../src/lockdown.js';
+import { getOwnPropertyNames } from '../src/commons.js';
+import { initialGlobalPropertyNames } from '../src/permits.js';
 
 // eslint-disable-next-line no-eval
 if (!eval.toString().includes('native code')) {
@@ -20,6 +22,13 @@ test('permitted prototypes - on', t => {
   Object.prototype.foo = 1;
   Object.prototype.hasOwnProperty.foo = 1;
 
+  // eslint-disable-next-line no-eval
+  t.truthy(globalThis.eval && !Object.isFrozen(globalThis.eval));
+  t.truthy(globalThis.Function && !Object.isFrozen(globalThis.Function));
+  for (const prop of getOwnPropertyNames(initialGlobalPropertyNames)) {
+    globalThis[prop] && t.truthy(!Object.isFrozen(globalThis[prop]));
+  }
+
   console.time('Benchmark repairIntrinsics()');
   const hardenIntrinsics = repairIntrinsics();
   console.timeEnd('Benchmark repairIntrinsics()');
@@ -27,6 +36,14 @@ test('permitted prototypes - on', t => {
   console.time('Benchmark hardenIntrinsics()');
   hardenIntrinsics();
   console.timeEnd('Benchmark hardenIntrinsics()');
+
+  // eslint-disable-next-line no-eval
+  t.truthy(globalThis.eval && Object.isFrozen(globalThis.eval));
+  t.truthy(globalThis.Function && Object.isFrozen(globalThis.Function));
+  t.truthy(globalThis.Compartment && Object.isFrozen(globalThis.Compartment));
+  for (const name of getOwnPropertyNames(initialGlobalPropertyNames)) {
+    t.assert(globalThis[name] && Object.isFrozen(globalThis[name]));
+  }
 
   t.is(globalThis.foo, 1);
   t.is(Object.foo, undefined);
--- a/packages/ses/src/global-object.js
+++ b/packages/ses/src/global-object.js
@@ -110,10 +110,12 @@ export const setGlobalObjectMutableProperties = (
     globalThis: globalObject,
   };
 
-  perCompartmentGlobals.Compartment = makeCompartmentConstructor(
-    makeCompartmentConstructor,
-    intrinsics,
-    markVirtualizedNativeFunction,
+  perCompartmentGlobals.Compartment = freeze(
+    makeCompartmentConstructor(
+      makeCompartmentConstructor,
+      intrinsics,
+      markVirtualizedNativeFunction,
+    ),
   );
 
   // TODO These should still be tamed according to the whitelist before
@@ -145,7 +147,7 @@ export const setGlobalObjectEvaluators = (
   markVirtualizedNativeFunction,
 ) => {
   {
-    const f = makeEvalFunction(evaluator);
+    const f = freeze(makeEvalFunction(evaluator));
     markVirtualizedNativeFunction(f);
     defineProperty(globalObject, 'eval', {
       value: f,
@@ -155,7 +157,7 @@ export const setGlobalObjectEvaluators = (
     });
   }
   {
-    const f = makeFunctionConstructor(evaluator);
+    const f = freeze(makeFunctionConstructor(evaluator));
     markVirtualizedNativeFunction(f);
     defineProperty(globalObject, 'Function', {
       value: f,
--- a/packages/ses/test/test-tame-symbol-constructor.js
+++ b/packages/ses/test/test-tame-symbol-constructor.js
@@ -23,8 +23,8 @@ lockdown();
 test('Symbol cleaned by permits', t => {
   t.true('dummy' in Symbol);
   t.false(gopd(Symbol, 'iterator').configurable);
-  t.true(isExtensible(Symbol));
-  t.false(isFrozen(Symbol));
+  t.false(isExtensible(Symbol));
+  t.true(isFrozen(Symbol));
   t.not(Symbol.constructor, Symbol);
 
   const c = new Compartment();


packages/ses/src/compartment-shim.js
```diff
@@ -27,9 +27,12 @@ import { getDeferredExports } from './module-proxy.js';
 import { assert } from './error/assert.js';
 import { compartmentEvaluate } from './compartment-evaluate.js';
 import { makeSafeEvaluator } from './make-safe-evaluator.js';
+import {makeHardener} from './make-hardener.js';
```
kriskowal, line 30: Running `yarn format` at the base of the repo will get you through CI. Your favorite IDE can be configured to do this automatically when you save a file. My favorite IDE can too, but I haven’t figured it out.

packages/ses/src/compartment-shim.js
```diff
@@ -27,9 +27,12 @@ import { getDeferredExports } from './module-proxy.js';
...
 
 const { quote: q } = assert;
 
+const safeHarden = makeHardener();
```
kriskowal, line 34: There should be a tamed harden laying around or available to be threaded through to here from `lockdown.js` (tho be aware it moves in #1705)

packages/ses/src/compartment-shim.js
```diff
@@ -269,6 +272,11 @@ export const makeCompartmentConstructor = (
 
     assign(globalObject, endowments);
 
+    // Harden special APIs
+    safeHarden(globalObject.eval);
+    safeHarden(globalObject.Function);
+    safeHarden(globalObject.Compartment);
```
kriskowal, line 278: `if (intrinsicsHardened()) {`

kriskowal (commented):
As discussed at Endo weekly sync.

packages/ses/src/lockdown.js
mhofman, line 1: I believe what needed to be hardened as well is the `%InitialSymbol%`, present as `globalThis.Symbol` after repair?

packages/ses/src/compartment.js
```diff
@@ -269,6 +270,12 @@ export const makeCompartmentConstructor = (
 
     assign(globalObject, endowments);
 
+    // freeze Symbol, Compartment constructor and evaluators in new Compartments
+    freeze(globalObject.Symbol);
```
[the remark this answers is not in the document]
weizman: You are correct, freezing this has no meaning, because `%SharedSymbol%` is being frozen already on https://github.com/endojs/endo/blob/master/packages/ses/src/lockdown.js#L386.
Meaning, this attempt should be removed completely, not moved over, since again, it's already being taken care of.

packages/ses/src/lockdown.js
weizman: 83c41f8

packages/ses/src/compartment.js
```diff
@@ -269,6 +270,12 @@ export const makeCompartmentConstructor = (
...
+    freeze(globalObject.Function);
+    freeze(globalObject.eval);
+    freeze(globalObject.Compartment);
```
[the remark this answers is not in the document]
weizman: 83c41f8 (accomplished for Compartment as well as your request of evaluators)

packages/ses/src/compartment.js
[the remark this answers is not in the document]
weizman: 83c41f8724b9d7b51e16237d22b580ad83612327

packages/ses/src/lockdown.js
```diff
@@ -385,7 +385,8 @@ export const repairIntrinsics = (options = {}) => {
     // must be the operation that modifies the intrinsics.
     tamedHarden(intrinsics);
 
-    // Harden Compartment constructor and evaluators in start Compartment
+    // Harden Symbol, Compartment constructor and evaluators in start Compartment
```
mhofman, line 388: Small nit. I think we should visually separate the hardening of `Symbol` from the evaluators.

My reasoning is that `Symbol` inside a compartment is shared between all compartments, unlike evaluators which all have their own distinct ones.

I'm also wondering if there are other "initial" only intrinsics that should similarly be hardened. `Error` comes to mind, and I'm wondering if it's already hardened, and if so, why `Symbol` is not handled similarly to `Error`.

weizman: 1. In that ctx, should `Compartment` be separated as well as `Symbol`? We don't consider it to be an evaluator right?
2. `Error` isn't hardened on the start compartment, before deciding on that might be useful to list those - is there a reference to the group of intrinsics that a Compartment gets that are new instances of their original instance on the start Compartment rather than identical instances? (e.g. `Function`, `eval`, `Error`, etc)

mhofman: `Compartment` is definitely an evaluator. `new Compartment().evaluate()` will eval code.

Regarding things that are only on the initial compartment's global, see `initialGlobalPropertyNames` in `permits.js`. `Symbol` is excluded there, for reasons I actually don't fully understand. The comment seem to indicate these intrinsics are somehow repaired, but I'm wondering why. Regardless, it's possible we want all these initial only intrinsics to be hardened during `hardenIntrinsics`. Paging @erights

mhofman: What I don't follow is how adding something to `initialGlobalPropertyNames` ends up having any kind of implication besides it being assigned to the global object.

mhofman: Oh there is no `%InitialSymbol%` defined anywhere. Why don't we capture the original `Symbol` as `%InitialSymbol%` instead of this weird dance? Then all these initial intrinsics can be given the same treatment.

erights: Because then we'd need to list all its properties that we do not want removed, which we assume we cannot know ahead of time.

Or, if you have something in mind you think will work, try it. The PR should either be trivial or impossible ;)

weizman: @mhofman fix d3163f722bfe8df507cbf86132c04d4af35545b5

Also, I noticed the list of evaluators we wish to harden is almost identical to `uniqueGlobalPropertyNames` in `permits.js` (except for `globalThis`), perhaps we should use it instead of naming all three by name?

mhofman: I have to admit I don't fully grasp the purpose of `uniqueGlobalPropertyNames` as it doesn't seem used anywhere. I suppose it's similar to `initialGlobalPropertyNames`, which is from permits, but seem only used to install some properties onto the initial compartment's global object.

I'd say for now, we can keep the explicit hardening of the evaluators.

packages/ses/src/lockdown.js
```diff
@@ -385,6 +386,19 @@ export const repairIntrinsics = (options = {}) => {
...
+
+    // Harden evaluators
+    tamedHarden(globalThis.Function);
+    tamedHarden(globalThis.eval);
+    tamedHarden(globalThis.Compartment);
+
+    // Harden initialGlobalPropertyNames
+    for (const prop of getOwnPropertyNames(initialGlobalPropertyNames)) {
+      tamedHarden(globalThis[prop]);
+    }
```
mhofman, line 400: Nit: given the status of `Symbol` as effectively an `initialGlobalPropertyNames`, move its standalone hardening here

weizman: 38a92c087

weizman:
I'm a bit confused regarding where to implement tests for this proposal, should this be a new file or is there an already existing file that matches that I might be missing IYO? @mhofman

erights:
Why still Draft? Is this ready for review?

mhofman:
> I'm a bit confused regarding where to implement tests for this proposal, should this be a new file or is there an already existing file that matches that I might be missing IYO? @mhofman

Your guess is as good as mine at this point. `test-global-object.js` looks like a good contender for the `Compartment` related changes. `test-repair-intrinsics` is likely the spot for the `hardenIntrinsics` changes. Neither seem to have closely matching tests though.

mhofman:
> Why still Draft? Is this ready for review?

@weizman please remove the WIP from the PR title too.

mhofman (commented):
Looks good so far but I'd like to see the tests you mention added.

weizman:
[fixed](https://github.com/endojs/endo/pull/1731#pullrequestreview-1631448385) 3db693ebf @mhofman

packages/ses/test/test-repair-intrinsics.js
```diff
@@ -28,6 +31,10 @@ test('permitted prototypes - on', t => {
   hardenIntrinsics();
   console.timeEnd('Benchmark hardenIntrinsics()');
 
+  t.truthy(globalThis.eval && Object.isFrozen(globalThis.eval));
+  t.truthy(globalThis.Function && Object.isFrozen(globalThis.Function));
+  t.truthy(globalThis.Compartment && Object.isFrozen(globalThis.Compartment));
+
   t.is(globalThis.foo, 1);
```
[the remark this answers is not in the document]
weizman: oops forgot we did that too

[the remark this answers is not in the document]
weizman: fixed 7e9d7cf29

packages/ses/test/test-compartment-instance.js
```diff
@@ -2,10 +2,14 @@ import test from 'ava';
...
 test('Compartment instance', t => {
-  t.plan(9);
+  t.plan(12);
 
   const c = new Compartment();
 
+    t.truthy(c.globalThis.eval && Object.isFrozen(c.globalThis.eval));
+    t.truthy(c.globalThis.Function && Object.isFrozen(c.globalThis.Function));
+    t.truthy(c.globalThis.Compartment && Object.isFrozen(c.globalThis.Compartment));
+
```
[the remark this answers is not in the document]
weizman: 58e902d3c

mhofman:
This looks good by me, I'll let @kriskowal help bringing this over the line.

packages/ses/src/lockdown.js
```diff
@@ -385,6 +386,17 @@ export const repairIntrinsics = (options = {}) => {
     // must be the operation that modifies the intrinsics.
     tamedHarden(intrinsics);
 
+    // Harden evaluators
+    tamedHarden(globalThis.Function);
+    tamedHarden(globalThis.eval);
+    tamedHarden(globalThis.Compartment);
+
+    // Harden initialGlobalPropertyNames and Symbol
```
kriskowal, line 394:
```suggestion
    // Harden Symbol and properties for initialGlobalPropertyNames in the host realm
```

packages/ses/test/test-repair-intrinsics.js
```diff
@@ -3,6 +3,8 @@
 
 import test from 'ava';
 import { repairIntrinsics } from '../src/lockdown.js';
+import {getOwnPropertyNames} from '../src/commons.js';
+import {initialGlobalPropertyNames} from '../src/permits.js';
```
kriskowal, line 7: Run `yarn format` at the top of the repository to clear out the CI errors for these lines.

packages/ses/test/test-repair-intrinsics.js
```diff
@@ -28,6 +37,14 @@ test('permitted prototypes - on', t => {
   hardenIntrinsics();
   console.timeEnd('Benchmark hardenIntrinsics()');
 
+  // eslint-disable-next-line no-eval
+  t.truthy(globalThis.eval && Object.isFrozen(globalThis.eval));
+  t.truthy(globalThis.Function && Object.isFrozen(globalThis.Function));
+  t.truthy(globalThis.Compartment && Object.isFrozen(globalThis.Compartment));
+  for (const prop of getOwnPropertyNames(initialGlobalPropertyNames)) {
+    globalThis[prop] && t.truthy(Object.isFrozen(globalThis[prop]));
+  }
```
kriskowal, line 46: The iterand is the name of the property, whereas prop could just as well imply the corresponding property descriptor or property value. I moved the outer condition so that the assertion count is consistent regardless of whether `&&` short-circuits evaluation.

```suggestion
  for (const name of getOwnPropertyNames(initialGlobalPropertyNames)) {
    t.ok(globalThis[name] && Object.isFrozen(globalThis[name]));
  }
```

kriskowal (changes requested):
Just a couple nits. Ping me for a quick stamp.

packages/ses/test/test-repair-intrinsics.js
weizman: fixed 4cb35cb6e

packages/ses/test/test-repair-intrinsics.js
weizman: fixed 4bef5ee26

kriskowal (approved):

Issue #1714, opened 2023-08-08
refactor(bundle-source): Use atomic rename instead of lock for bundle cache writer


kriskowal:
## What is the Problem Being Solved?

The `bundle-source` tool will occasionally fail to make progress if a previous run wrote a `.lock` file and did not clear it before exiting.

## Description of the Design

Instead of using a `.lock` file to coördinate concurrent runs of `bundle-source`, opportunistically create a new bundle with a name like `$X.$PID` and rename `$X.$PID` to `$X` when it is complete.

This will have the disadvantage that it will occasionally leave a trash `$X.$PID` file if it’s interrupted without an opportunity to clean up, in exchange for the advantage that concurrent runs will make progress regardless of whether a previous run was interrupted.

PR #1732, opened 2023-08-23
Bundle refactoring

kriskowal:
closes: #1714 

Individually reviewable commits.

## Description

This is a stack of changes to improve the experience of using the `bundle-source` command.

* Increases the sensitivity of the cache key: any deviation from the last known mtime (even in regress) as well as a change in size will cause a bundle to be regenerated.
* Replaces the lockfile mechanism with atomic renames. POSIX systems guarantee that `rename` is atomic if the file remains on the same volume and we can guarantee that the file remains on the same volume by keeping it in the same directory. So we create a scratch file including the PID of the writing process (and a consistent extension so they’re easy to delete if the bundler exits prematurely.) Then, we capture the mtime and size of the scratch and rename the extension off. This ensures that we capture the mtime and size of the file we wrote, even if there are concurrent processes generating the same bundle. The new failure mode is duplication of effort. Stray scratch files do not interfere with progress on subsequent runs.
* The bundler only loads the dependencies necessary for the requested formats. So, if you only generate Endo Zip Base 64 bundles, we no longer import Rollup. Likewise, if you only generate Get Exports or Nested Evaluate, you do not import the Compartment Mapper.
* Debug logs will appear only if the `DEBUG` environment variable is a comma separated list including `bundle-source`.
* Factors out `transform.js` and `fs.js` in expectation that these will be the bases of new packages. The evasive transforms are useful for other bundling mechanisms like WebPack and the file system module is a good place to start for a hardened file system for the Pet Daemon.

### Security Considerations

No known implications

### Scaling Considerations

This should improve performance of the bundler.

Usage of the bundler does not change, but the `DEBUG=bundle-source` directive may be useful to end users in the event that the bundler misbehaves.

### Testing Considerations

I’ve manually tested the bundler tool on `demo/circular/a.js` and `demo/circular/b.js`, inducing cache misses by touching or adding newlines to these files, and manually invalidating the `meta` file. In all cases, the tool converged on cache hits.

### Upgrade Considerations

This should not impact contract upgrade.


packages/bundle-source/src/fs.js
```diff
@@ -0,0 +1,81 @@
...
+ */
+export const makeFileWriter = (fileName, { fs, path }, pid) => {
+  const make = there => makeFileWriter(there, { fs, path });
+  return harden({
+    toString: () => fileName,
+    writeText: (txt, opts) => fs.promises.writeFile(fileName, txt, opts),
+    atomicWriteText: async (txt, opts) => {
+      const scratchName = `${fileName}.${pid}.scratch`;
+      await fs.promises.writeFile(scratchName, txt, opts);
+      const stats = await fs.promises.stat(scratchName);
```
kriskowal, line 72: I’m not going to lie: I’m really chuffed that we can reliably capture the mtime and size of the file we wrote by capturing them before removing the extension. Mid-way through this refactor, I noticed there was a race after writing the scratch bundle, removing the extension, (race here), and then getting the mtime and size of what we’d written. Moving the stat here eliminates the race.

packages/bundle-source/src/fs.js
```diff
@@ -0,0 +1,81 @@
...
+  const maybeReadText = () =>
+    readText().catch(error => {
+      if (
+        error.message.startsWith('ENOENT: ') ||
+        error.message.startsWith('EISDIR: ')
+      ) {
+        return undefined;
+      }
+      throw error;
+    });
```
kriskowal, line 37: Using `maybeReadText` instead of `exists` eliminates the race to read the `meta` file after checking that it exists. This was not a problem under the lock, but becomes a problem when we rely on atomic writes instead.

packages/bundle-source/src/fs.js
```diff
@@ -0,0 +1,81 @@
...
+  const exists = fn =>
+    fs.promises.stat(fn).then(
+      () => true,
+      e => {
+        if (e.code === 'ENOENT') {
+          return false;
+        }
+        throw e;
+      },
+    );
```
kriskowal, line 24: This is now dead code but I figure we’ll want to decide whether to keep it after we start building out a public `@endo/fs` API.

packages/bundle-source/src/transform.js
```diff
@@ -0,0 +1,121 @@
+import * as babelParser from '@babel/parser';
```
kriskowal, line 1: fyi @naugtur, I factored this transform out in preparation for it to move to `@endo/transforms`.

packages/bundle-source/src/transform.js
```diff
@@ -0,0 +1,121 @@
...
+      // then do it for this location.
+      if (unmapLoc) {
+        unmapLoc(loc);
+      }
+      (trailingComments || []).forEach(node => rewriteComment(node, unmapLoc));
+    },
+  });
+}
+
+export async function transformSource(
```
dckc, line 94: Seems like a prime candidate for a unit test.

packages/bundle-source/cache.js
```diff
@@ -238,15 +241,26 @@ export const makeBundleCache = (wr, cwd, readPowers, opts) => {
...
-        /** @type {import('fs').Stats} */
-        const { mtime } = await itemRd.stat();
-        return { relativePath, mtime: mtime.toISOString() };
-      }),
+    const actualStats = await Promise.all(
+      contents.map(
+        async ({ relativePath, mtime: priorMtime, size: priorSize }) => {
+          const itemRd = moduleRd.neighbor(relativePath);
+          /** @type {import('fs').Stats} */
+          const { mtime, size } = await itemRd.stat();
```
dckc, line 134: Does this consume an unbounded number of file descriptors? Worth fixing?

packages/bundle-source/cache.js
```diff
@@ -65,12 +79,20 @@ export const makeFileReader = (fileName, { fs, path }) => {
...
  * }} io
+ * @param {number} pid
  */
-export const makeFileWriter = (fileName, { fs, path }) => {
+export const makeFileWriter = (fileName, { fs, path }, pid) => {
   const make = there => makeFileWriter(there, { fs, path });
   return harden({
     toString: () => fileName,
     writeText: (txt, opts) => fs.promises.writeFile(fileName, txt, opts),
+    atomicWriteText: async (txt, opts) => {
```
dckc: do we need the `pid` for the life of the file writer? or could it be an arg to `atomicWriteText`?

dckc (approved):
looks great!

This review is just a little rushed... dinner bell is ringing.

I'm a little nervous about lack of tests, but that pre-dates this PR.

packages/bundle-source/cache.js
kriskowal: I considered making thee `pid` a parameter of `atomicWriteText` but, given that it is guaranteed to be consistent, it’s the only value guaranteed to work for the lifetime of the process, even if there’s trash on the file system from a previous boot with a coïncident pid, and that threading `pid` to all the call sites is a little invasive, I went with capturing it early.

packages/bundle-source/cache.js
kriskowal: Thankfully, `stat` syscall doesn’t allocate a descriptor. `writeFile` certainly does. I can follow up with a mutex like we did with Compartment Mapper’s read powers.

kriskowal: Added a commit to synchronize fs.

packages/bundle-source/cache.js
dckc: Maybe for later when fs graduates to a pkg... It seems burdensome to provide a pid to make a file writer. Maybe a maker without pid and without atomicWrite.

kriskowal: Agreed. Can be another layer.

kriskowal: Posted a commit that makes `makeAtomicFileWriter` a separable layer that extends `makeFileWriter`.

I committed a minor crime in applying a pattern for maker extensibility (the same crime committed in `makeCompartmentConstructor`).

packages/bundle-source/src/fs.js
```diff
@@ -73,40 +73,72 @@ export const makeFileReader = (fileName, { fs, path }) => {
...
+ * @param {(there: string) => ReturnType<makeFileWriter>} make
  */
-export const makeFileWriter = (fileName, { fs, path }, pid) => {
-  const make = there => makeFileWriter(there, { fs, path }, pid);
-
-  const writeText = async (specificName, txt, opts) => {
+export const makeFileWriter = (
+  fileName,
+  { fs, path },
+  make = there => makeFileWriter(there, { fs, path }, make),
```
kriskowal, line 81: *higher order maniacal cackling*

packages/bundle-source/src/fs.js
```diff
@@ -73,40 +73,72 @@ export const makeFileReader = (fileName, { fs, path }) => {
...
+ *   path: Pick<import('path'), 'resolve' | 'relative' | 'normalize'>,
+ * }} io
+ * @param {number} pid
+ * @param {(there: string) => ReturnType<makeAtomicFileWriter>} make
+ */
+export const makeAtomicFileWriter = (
+  fileName,
+  { fs, path },
+  pid,
+  make = there => makeAtomicFileWriter(there, { fs, path }, pid, make),
```
kriskowal, line 130: This ensures that the `pid` gets retained even as `makeFileWriter` uses this maker to create neighbors of this extended maker.

PR #1738, opened 2023-08-26
docs(patterns): Better explain key-order/rank-order invariants

gibson042:


erights (approved):
This is really good! No suggested changes. LGTM!

PR #1740, opened 2023-08-26
fix(exo): Extend InterfaceGuard to support symbol-keyed methods

### LINKED ISSUE #1728: Inconsistency between Remotable rules and InterfaceGuard pattern language
(Originally reported at https://github.com/Agoric/agoric-sdk/issues/8231 )

https://github.com/Agoric/agoric-sdk/blob/61b6c39a7c8cedf7c7b11a5f5352d82f351e5657/packages/notifier/src/publish-kit.js#L48-L53
```js
 export const IterableEachTopicI = M.interface('IterableEachTopic', { 
   subscribeAfter: SubscriberI.methodGuards.subscribeAfter, 
   [Symbol.asyncIterator]: M.call().returns( 
     M.remotable('ForkableAsyncIterableIterator'), 
   ), 
 }); 
```

Remotables are allowed to have symbol-named methods. (The particular case that is already important for us is `Symbol.asyncIterator` as we see here.)

InterfaceGuards are supposed to be Passable descriptions of the API of an exo, including the guards and methods used do the first level of input validation of the exo's methods. The InterfaceGuard uses a CopyRecord of method guards, where each property name of that CopyRecord is supposed to be the name of the method being described/guarded.

However, a CopyRecord can only have string-named properties. Therefore, there is currently no way for an InterfaceGuard to provide a MethodGuard for describing/guarding a symbol-named method.

gibson042:
Fixes #1728

## Description

Extend InterfaceGuard with a `symbolMethodGuards` field whose value is a CopyMap\<symbol, MethodGuard> when the interface has symbol-keyed methods such as Symbol.asyncIterator. The current implementation expresses it as completely optional and does not create it when there are no symbol-keyed methods, but we could also set it to an empty CopyMap in that case.

### Security Considerations

None AFAICT.

### Scaling Considerations

None AFAICT, even if we always add the CopyMap.

None.

### Testing Considerations

Just unit tests.

### Upgrade Considerations

None special; the new functionality should be fully backwards compatible.

### DIFF
--- a/packages/exo/README.md
+++ b/packages/exo/README.md
@@ -11,12 +11,20 @@ When an exo is defined with an InterfaceGuard, the exo is augmented by default w
 ```js
 // `GET_INTERFACE_GUARD` holds the name of the meta-method
 import { GET_INTERFACE_GUARD } from '@endo/exo';
+import { getCopyMapEntries } from '@endo/patterns';
 
 ...
    const interfaceGuard = await E(exo)[GET_INTERFACE_GUARD]();
    // `methodNames` omits names of automatically added meta-methods like
    // the value of `GET_INTERFACE_GUARD`.
    // Others may also be omitted if `interfaceGuard.partial`
-   const methodNames = Reflect.ownKeys(interfaceGuard.methodGuards);
+   const methodNames = [
+     ...Reflect.ownKeys(interfaceGuard.methodGuards),
+     ...(interfaceGuard.symbolMethodGuards
+       ? [...getCopyMapEntries(interfaceGuard.symbolMethodGuards)].map(
+           entry => entry[0],
+         )
+       : []),
+   ];
 ...
 ```
--- a/packages/exo/src/exo-tools.js
+++ b/packages/exo/src/exo-tools.js
@@ -8,6 +8,7 @@ import {
   isAwaitArgGuard,
   assertMethodGuard,
   assertInterfaceGuard,
+  getCopyMapEntries,
 } from '@endo/patterns';
 
 /** @typedef {import('@endo/patterns').Method} Method */
@@ -15,7 +16,7 @@ import {
 
 const { quote: q, Fail } = assert;
 const { apply, ownKeys } = Reflect;
-const { defineProperties } = Object;
+const { defineProperties, fromEntries } = Object;
 
 /**
  * A method guard, for inclusion in an interface guard, that enforces only that
@@ -263,8 +264,17 @@ export const defendPrototype = (
   let methodGuards;
   if (interfaceGuard) {
     assertInterfaceGuard(interfaceGuard);
-    const { interfaceName, methodGuards: mg, sloppy = false } = interfaceGuard;
-    methodGuards = mg;
+    const {
+      interfaceName,
+      methodGuards: mg,
+      symbolMethodGuards,
+      sloppy = false,
+    } = interfaceGuard;
+    methodGuards = harden({
+      ...mg,
+      ...(symbolMethodGuards &&
+        fromEntries(getCopyMapEntries(symbolMethodGuards))),
+    });
     {
       const methodNames = ownKeys(behaviorMethods);
       const methodGuardNames = ownKeys(methodGuards);
--- a/packages/exo/test/test-heap-classes.js
+++ b/packages/exo/test/test-heap-classes.js
@@ -2,7 +2,7 @@
 import { test } from './prepare-test-env-ava.js';
 
 // eslint-disable-next-line import/order
-import { M } from '@endo/patterns';
+import { getCopyMapEntries, M } from '@endo/patterns';
 import {
   defineExoClass,
   defineExoClassKit,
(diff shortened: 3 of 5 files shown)


erights (commented):
First quick reaction: Awesome choice to use a CopyMap rather than an array of pairs! In thinking about this issue, that "natural" choice never occurred to me.

(Good you're also fixing CopyMap patterns, just in time for the first production use of CopyMaps!)

erights:
> expresses it as completely optional and does not create it when there are no symbol-keyed methods, but we could also set it to an empty CopyMap in that case.

Since it arrives late and optional is coherent. I think it should remain optional. But if it does happen to be present with an empty map, that should be fine too. I don't see any reason to insist on a canonical representation for this case. (I think you're already doing that, in which case, no change suggested.)

packages/exo/README.md
```diff
@@ -10,13 +10,19 @@ When an exo is defined with an InterfaceGuard, the exo is augmented by default w
...
 // `GET_INTERFACE_GUARD` holds the name of the meta-method
-import { GET_INTERFACE_GUARD } from `@endo/exo`;
+import { GET_INTERFACE_GUARD } from '@endo/exo';
+import { getCopyMapKeys } from '@endo/patterns';
 
 ...
    const interfaceGuard = await E(exo)[GET_INTERFACE_GUARD]();
    // `methodNames` omits names of automatically added meta-methods like
    // the value of `GET_INTERFACE_GUARD`.
    // Others may also be omitted if `interfaceGuard.partial`
```
erights, line 20: Attn @michaelfig 

My mistake! In anything that saw the light of day, it was never called `interfaceGuard.partial`. It became `interfaceGuard.sloppy`, but I forgot to rename this occurrence. 

But @michaelfig is in process with a PR that will make changes in this area, so leave that to him. For your PR, no change suggested.

packages/exo/README.md
```diff
@@ -10,13 +10,19 @@ When an exo is defined with an InterfaceGuard, the exo is augmented by default w
...
-   const methodNames = Reflect.ownKeys(interfaceGuard.methodGuards);
+   const methodNames = Reflect.ownKeys(interfaceGuard.methodGuards)
+     .concat(
+       interfaceGuard.symbolMethodGuards
+         ? getCopyMapKeys(interfaceGuard.symbolMethodGuards)
+         : []
+     );
```
erights, line 26: I like the following style a bit better, but up to you.

```suggestion
   const methodNames = [
     ...Reflect.ownKeys(interfaceGuard.methodGuards),
     ...interfaceGuard.symbolMethodGuards
         ? getCopyMapKeys(interfaceGuard.symbolMethodGuards)
         : [],
   ];
```

packages/exo/src/exo-tools.js
```diff
@@ -263,8 +264,17 @@ export const defendPrototype = (
...
+      interfaceName,
+      methodGuards: mg,
+      symbolMethodGuards,
+      sloppy = false,
+    } = interfaceGuard;
+    methodGuards = harden({
+      ...mg,
+      ...(symbolMethodGuards &&
+        fromEntries(getCopyMapEntries(symbolMethodGuards))),
+    });
```
erights, line 277: Nice!

Given this, my preference to use [...x, ...y] above is stronger. Now I will request that change ;)

packages/exo/test/test-heap-classes.js
```diff
@@ -53,6 +53,18 @@ test('test defineExoClass', t => {
   });
   t.deepEqual(upCounter[GET_INTERFACE_GUARD](), UpCounterI);
   t.deepEqual(ownKeys(UpCounterI.methodGuards), ['incr']);
+  t.is(UpCounterI.symbolMethodGuards, undefined);
+
+  const symbolic = Symbol.for('symbolic');
+  const FooI = M.interface('Foo', {
```
erights, line 59: Please also add a test with an exo implementing FooI and then call the symbol-named method to test that the raw method is guarded by the methodGuard.

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1804,12 +1805,17 @@ const makeMethodGuardMaker = (
...
-  interfaceName: M.string(),
-  methodGuards: M.recordOf(M.string(), MethodGuardShape),
-  sloppy: M.boolean(),
-});
+const InterfaceGuardShape = M.splitRecord(
+  {
+    klass: 'Interface',
+    interfaceName: M.string(),
+    methodGuards: M.recordOf(M.string(), MethodGuardShape),
+    sloppy: M.boolean(),
```
erights, line 1813: Just be aware that as of @michaelfig 's upcoming PR, `sloppy` will move into that optional section, and another will appear there as well. No change suggested.

packages/patterns/src/types.js
```diff
@@ -514,8 +514,9 @@ export {};
  * @typedef {{
  *   klass: 'Interface',
  *   interfaceName: string,
- *   methodGuards: T
- *   sloppy?: boolean
+ *   methodGuards: { [K in keyof T]: K extends symbol ? never : T[K] },
+ *   symbolMethodGuards?: CopyMap<symbol, MethodGuard>,
```
erights, line 518: Nice typing of `methodGuards`. I wondered how to do that.

Could you use the same trick to type `symbolMethodGuard` with the opposing side of that conditional?

Attn @turadg

I think doing so will help @turadg 's efforts to statically inter TS types from patterns and guards.

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1824,11 +1830,29 @@ harden(assertInterfaceGuard);
...
+  }
   /** @type {InterfaceGuard} */
   const result = harden({
     klass: 'Interface',
     interfaceName,
-    methodGuards,
+    methodGuards: stringMethodGuards,
+    ...(symbolMethodGuardsEntries.length
+      ? { symbolMethodGuards: makeCopyMap(symbolMethodGuardsEntries) }
+      : {}),
```
erights, line 1855: That `... (x ? {y} : {})` is quite cute. No change suggested.

erights (approved):
LGTM, thanks! For the sake of agoric-sdk, it's great that we can get this into the next endo release.

packages/exo/README.md
gibson042: Done.

packages/exo/test/test-heap-classes.js
gibson042: Done.

packages/patterns/src/types.js
gibson042: I got it to type the CopyMap K as a symbol-keyed entry of T, but don't know how to establish the pair relationship. It will be open for followup.

PR #1743, opened 2023-08-27
doc: fix wrong quote in import example code

erights:
Saw https://github.com/endojs/endo/pull/1740/files?short_path=12f0146#diff-12f0146121857891eaea02008892eb51e98883364bc9ed4dfc544b7676e96c43 and wondered if it was a common mistake. Not too common, but not unique either.

See also https://github.com/Agoric/agoric-sdk/pull/8260


gibson042 (approved):
Nice; good thinking to check on that!

PR #1744, opened 2023-08-27
chore(daemon): endow worker with console

dckc:
### Security Considerations

Was this left out on purpose?

cc @michaelfig


michaelfig (approved):
I'm approving based on common understanding that console is harmless to endow.

Since this is on experimental `endo`, it can be excised especially easily if its effect is egregious.

PR #1748, opened 2023-08-28
Endo: Corrections for demo instructions and refactoring

kriskowal:
- Moves the CLI commands `start`, `stop`, `restart`, `reset`, &c into a `daemon` subcommand to keep the common usage obvious with `endo --help`.
- Updates the demo instructions to track the current state of things.
- Refactors the daemon’s current two servers into separate modules with some common interface affordances, in preparation for additional services (protected and public TCP and WebSocket servers.)


dckc (approved):
Looks plausible. I don't see anything wrong.

I'd be more confident that it's right with tests. And I'm not sure how to read the static types.

Issue #1750, opened 2023-08-29
compartment-mapper/policy: allows "resources" to be an array


boneskull:
## Steps to reproduce

I'm not sure the most straightforward way to reproduce this in the _real world_--since I discovered it while poking around in the codebase--but ostensibly one could edit the policy in `packages/compartment-mapper/demo/policy/index.mjs` and change `resources` to an array, eliminating all of the keys.  I am not sure if that itself will break, but per @naugtur this was unintended behavior.

## Expected behavior

An error is thrown if `resources` is an array.

## Platform environment

Ubuntu / v18.17.1 / `master`

## Additional context

I am creating this issue because when I started to open a PR to fix it, it said I should create an issue.  I am good at following directions!

kriskowal:
@boneskull You are a role model for us all. Thank you :-)

PR #1751, opened 2023-08-29
fix(compartment-mapper/policy): do not allow "resources" to be an array

boneskull:
closes: #1750

## Description

- Disallow `resources` prop of a policy to be an `Array`
- Fix error message when `defaultAttenuator` is not a `string` (it was referencing the wrong field)
- Add tests for this and other expected behavior in `assertPolicy()`

### Security Considerations

No

### Scaling Considerations

No

I am working on some sort of proper type declaration or schema for policies, so this will be forthcoming.

### Testing Considerations

There may be more things in `assertPolicy()` that I missed, but I at least wanted to cover the things I changed.

### Upgrade Considerations

Any policies currently in use which have a `resources` prop that is an `Array` will break.


kriskowal (approved):
This looks good to me and defer to @naugtur for final-stamp for all-things-policy.

naugtur (approved):

PR #1758, opened 2023-09-03
typescript 5.2

turadg:
## Description

https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/

Also fixes some typing problems that 5.2 uncovered.

### Security Considerations

n/a

### Scaling Considerations

n/a

not necessary

### Testing Considerations

CI sufficient

### Upgrade Considerations

Negligible runtime changes (default params)

### DIFF
--- a/packages/compartment-mapper/src/types.js
+++ b/packages/compartment-mapper/src/types.js
@@ -250,6 +250,7 @@ export {};
  * @param {string} packageLocation
  * @param {object} [options]
  * @param {string} [options.sourceMap]
+ * @param {SourceMapHook} [options.sourceMapHook]
  * @param {string} [options.sourceMapUrl]
  * @param {ReadFn | ReadPowers} [options.readPowers]
  * @returns {Promise<{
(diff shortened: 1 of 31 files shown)


packages/compartment-mapper/src/types.js
```diff
@@ -214,20 +215,6 @@ export {};
...
- * @property {string} location
- * @property {string} sha512
- */
-
-/**
- * @callback SourceMapHook
- * @param {string} sourceMap
- * @param {SourceMapHookDetails} details
- */
-
```
kriskowal, line 230: Deduplicating this interface implies a tighter coupling than necessarily exists.

turadg: Motivation was to fix a type error in their usage. The one kept fits the code in use.

Do you prefer duplicating that one instead of importing it?

turadg: Went with duplicating. PTAAL.

kriskowal (approved):

PR #1759, opened 2023-09-05
fix: restore compatibility with Agoric SDK master

michaelfig:
## Description

This fixes some of Endo's drifting away from Agoric SDK's requirements.  Other changes can be made in the Agoric SDK directly, so they are being done there.

Notably:
1. `@endo/bundle-source` typing and cache semantics (partly because `cache.js` was omitted from typechecking)
2. `?.` and `??` causing problems when attempting to bundle-source modules that use them.
3. Improve typing of `@endo/exo` to align with Agoric SDK's `packages/zone` and `packages/base-zone`.

### Security Considerations

n/a

### Scaling Considerations

n/a

n/a

### Testing Considerations

We really need CI somehow to trigger an integration test on Agoric SDK (#1761), to prevent landing Endo PRs that break things unless we know what we're doing, but that's for a later date.

### Upgrade Considerations

n/a

### DIFF
--- a/jsconfig.eslint-full.json
+++ b/jsconfig.eslint-full.json
@@ -1,7 +1,9 @@
 {
   "extends": "./jsconfig.eslint-base.json",
   "include": [
+    "*.js",
     "**/*.js",
+    "*.ts",
     "**/*.ts"
   ]
 }
--- a/packages/bundle-source/cache.js
+++ b/packages/bundle-source/cache.js
@@ -45,6 +45,7 @@ export const makeBundleCache = (wr, cwd, readPowers, opts) => {
     const statsByPath = new Map();
 
     const loggedRead = async loc => {
+      await null;
       if (!loc.match(/\bpackage.json$/)) {
         try {
           const itemRd = cwd.neighbor(new URL(loc).pathname);
@@ -107,13 +108,38 @@ export const makeBundleCache = (wr, cwd, readPowers, opts) => {
   };
 
   /**
-   * @param {BundleMeta} meta
+   * @param {string} targetName
+   * @param {Logger} _log
+   * @returns {Promise<string>}
+   */
+  const loadMetaText = (targetName, _log = defaultLog) =>
+    wr
+      .readOnly()
+      .neighbor(toBundleMeta(targetName))
+      .maybeReadText()
+      .catch(
+        ioError =>
+          Fail`${targetName}: cannot read bundle metadata: ${q(ioError)}`,
+      );
+
+  /**
    * @param {string} targetName
    * @param {*} rootOpt
    * @param {Logger} [log]
+   * @param {BundleMeta} [meta]
    * @returns {Promise<BundleMeta>}
    */
-  const validate = async (meta, targetName, rootOpt, log = defaultLog) => {
+  const validate = async (
+    targetName,
+    rootOpt,
+    log = defaultLog,
+    meta = undefined,
+  ) => {
+    meta = await (meta ||
+      loadMetaText(targetName, log).then(metaText => JSON.parse(metaText)));
+    if (!meta) {
+      throw Fail`no metadata found for ${q(targetName)}`;
+    }
     const {
       bundleFileName,
       bundleTime,
@@ -168,21 +194,14 @@ export const makeBundleCache = (wr, cwd, readPowers, opts) => {
    * @returns {Promise<BundleMeta>}
    */
   const validateOrAdd = async (rootPath, targetName, log = defaultLog) => {
-    const metaText = await wr
-      .readOnly()
-      .neighbor(toBundleMeta(targetName))
-      .maybeReadText()
-      .catch(
-        ioError =>
-          Fail`${targetName}: cannot read bundle metadata: ${q(ioError)}`,
-      );
+    const metaText = await loadMetaText(targetName, log);
 
     /** @type {BundleMeta | undefined} */
     let meta = metaText ? JSON.parse(metaText) : undefined;
 
     if (meta !== undefined) {
       try {
-        meta = await validate(meta, targetName, rootPath, log);
+        meta = await validate(targetName, rootPath, log, meta);
         const { bundleTime, bundleSize, contents } = meta;
         log(
           `${wr}`,
@@ -256,7 +275,7 @@ export const makeBundleCache = (wr, cwd, readPowers, opts) => {
 
 /**
  * @param {string} dest
- * @param {{ format?: string, dev?: boolean }} options
+ * @param {{ format?: string, dev?: boolean, log?: Logger }} options
  * @param {(id: string) => Promise<any>} loadModule
  * @param {number} [pid]
  */
--- a/packages/bundle-source/cache.js
+++ b/packages/bundle-source/cache.js
@@ -67,11 +67,6 @@ export const makeBundleCache = (wr, cwd, readPowers, opts) => {
     const bundleWr = wr.neighbor(bundleFileName);
     const metaWr = wr.neighbor(toBundleMeta(targetName));
 
-    // Prevent other processes from doing too much work just to see that we're
-    // already on it.
-    await metaWr.rm({ force: true });
-    await bundleWr.rm({ force: true });
-
     const bundle = await bundleSource(rootPath, bundleOptions, {
       ...readPowers,
       read: loggedRead,
@@ -110,7 +105,7 @@ export const makeBundleCache = (wr, cwd, readPowers, opts) => {
   /**
    * @param {string} targetName
    * @param {Logger} _log
-   * @returns {Promise<string>}
+   * @returns {Promise<string | undefined>}
    */
   const loadMetaText = (targetName, _log = defaultLog) =>
     wr
@@ -135,8 +130,20 @@ export const makeBundleCache = (wr, cwd, readPowers, opts) => {
     log = defaultLog,
     meta = undefined,
   ) => {
-    meta = await (meta ||
-      loadMetaText(targetName, log).then(metaText => JSON.parse(metaText)));
+    await null;
+    if (!meta) {
+      const metaJson = await loadMetaText(targetName, log);
+      if (metaJson) {
+        try {
+          meta = JSON.parse(metaJson);
+        } catch (error) {
+          if (error instanceof SyntaxError) {
+            throw SyntaxError(`Cannot parse JSON from ${targetName}, ${error}`);
+          }
+          throw error;
+        }
+      }
+    }
     if (!meta) {
       throw Fail`no metadata found for ${q(targetName)}`;
     }
--- a/packages/compartment-mapper/src/bundle.js
+++ b/packages/compartment-mapper/src/bundle.js
@@ -245,7 +245,12 @@ export const makeBundle = async (read, moduleLocation, options) => {
   for (const module of modules) {
     module.indexedImports = Object.fromEntries(
       Object.entries(module.resolvedImports).map(([importSpecifier, key]) => {
-        key = aliases.get(key) ?? key;
+        // UNTIL https://github.com/endojs/endo/issues/1514
+        // Prefer: key = aliases.get(key) ?? key;
+        const alias = aliases.get(key);
+        if (alias != null) {
+          key = alias;
+        }
         const module = modulesByKey[key];
         if (module === undefined) {
           throw new Error(
--- a/packages/exo/src/exo-makers.js
+++ b/packages/exo/src/exo-makers.js
@@ -86,7 +86,9 @@ export const initEmpty = () => emptyRecord;
  * @template {(...args: any[]) => any} I init function
  * @template {Methods} M methods
  * @param {string} tag
- * @param {any} interfaceGuard
+ * @param {import('@endo/patterns').InterfaceGuard<{
+ *   [K in keyof M]: import("@endo/patterns").MethodGuard
+ * }> | undefined} interfaceGuard
  * @param {I} init
  * @param {M & ThisType<{ self: M, state: ReturnType<I> }>} methods
  * @param {FarClassOptions<ClassContext<ReturnType<I>, M>>} [options]
@@ -147,7 +149,9 @@ harden(defineExoClass);
  * @template {(...args: any[]) => any} I init function
  * @template {Record<FacetName, Methods>} F facet methods
  * @param {string} tag
- * @param {any} interfaceGuardKit
+ * @param {{ [K in keyof F]: import("@endo/patterns").InterfaceGuard<{
+ *   [M in keyof F[K]]: import("@endo/patterns").MethodGuard;
+ * }> } | undefined} interfaceGuardKit
  * @param {I} init
  * @param {F & ThisType<{ facets: F, state: ReturnType<I> }> } methodsKit
  * @param {FarClassOptions<KitContext<ReturnType<I>,F>>} [options]
@@ -213,7 +217,9 @@ harden(defineExoClassKit);
 /**
  * @template {Methods} T
  * @param {string} tag
- * @param {import('@endo/patterns').InterfaceGuard<{ [M in keyof T]: import('@endo/patterns').MethodGuard }> | undefined} interfaceGuard CAVEAT: static typing does not yet support `callWhen` transformation
+ * @param {import('@endo/patterns').InterfaceGuard<{
+ *   [M in keyof T]: import('@endo/patterns').MethodGuard
+ * }> | undefined} interfaceGuard CAVEAT: static typing does not yet support `callWhen` transformation
  * @param {T} methods
  * @param {FarClassOptions<ClassContext<{},T>>} [options]
  * @returns {T & import('@endo/eventual-send').RemotableBrand<{}, T>}
--- a/packages/patterns/src/types.js
+++ b/packages/patterns/src/types.js
@@ -491,15 +491,20 @@ export {};
  */
 
 /**
- * @typedef {{
- * <M extends Record<any, MethodGuard>>(interfaceName: string,
- *             methodGuards: M,
- *             options?: {sloppy?: false}): InterfaceGuard<M>;
- * (interfaceName: string,
- *             methodGuards: any,
- *             options?: {sloppy?: true}): InterfaceGuard<any>;
- * }} MakeInterfaceGuard
+ * @typedef {(
+ *   interfaceName: string,
+ *   methodGuards: any,
+ *   options: {sloppy: true}) => InterfaceGuard<Record<PropertyKey, MethodGuard>>
+ * } MakeInterfaceGuardSloppy
  */
+/**
+ * @typedef {<M extends Record<PropertyKey, MethodGuard>>(
+ *   interfaceName: string,
+ *   methodGuards: M,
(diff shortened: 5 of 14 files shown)


packages/compartment-mapper/src/bundle.js
```diff
@@ -245,7 +245,13 @@ export const makeBundle = async (read, moduleLocation, options) => {
   for (const module of modules) {
     module.indexedImports = Object.fromEntries(
       Object.entries(module.resolvedImports).map(([importSpecifier, key]) => {
-        key = aliases.get(key) ?? key;
+        // TODO Change to the following line, once our tools don't choke on `?.`.
+        // See https://github.com/endojs/endo/issues/1514
```
turadg, line 249: that issue is closed. What was choking this time? We should have a new issue or at least re-open that one.

packages/compartment-mapper/src/bundle.js
```diff
@@ -245,7 +245,13 @@ export const makeBundle = async (read, moduleLocation, options) => {
   for (const module of modules) {
     module.indexedImports = Object.fromEntries(
       Object.entries(module.resolvedImports).map(([importSpecifier, key]) => {
-        key = aliases.get(key) ?? key;
+        // TODO Change to the following line, once our tools don't choke on `?.`.
```
turadg, line 248: consider an `UNTIL <issue>` comment https://github.com/Agoric/agoric-sdk/wiki/Coding-Style#code-comments

The issue provides enough documentation

packages/exo/src/exo-makers.js
```diff
@@ -86,7 +86,9 @@ export const initEmpty = () => emptyRecord;
  * @template {(...args: any[]) => any} I init function
  * @template {Methods} M methods
  * @param {string} tag
- * @param {any} interfaceGuard
+ * @param {import('@endo/patterns').InterfaceGuard<{
```
turadg, line 89: excellent

turadg (approved):
I didn't review the cache.js changes carefully.

The rest LGTM. Especially the annotations replacing `any`!

> We really need CI somehow to trigger an integration test on Agoric SDK, to prevent landing Endo PRs that break things unless we know what we're doing, but that's for a later date.

Worth an issue, yes? Please include a link.

packages/patterns/src/types.js
```diff
@@ -492,12 +492,12 @@ export {};
...
- *             options?: {sloppy?: false}): InterfaceGuard<M>;
  * (interfaceName: string,
  *             methodGuards: any,
- *             options?: {sloppy?: true}): InterfaceGuard<any>;
+ *             options: {sloppy: true}): InterfaceGuard<Record<PropertyKey, MethodGuard>>;
+ * <M extends Record<PropertyKey, MethodGuard>>(interfaceName: string,
+ *             methodGuards: M,
+ *             options?: {sloppy?: boolean}): InterfaceGuard<M>;
  * }} MakeInterfaceGuard
  */
```
gibson042, line 502: This is quite hard to read; can it be improved by naming the alternatives? Untested:
```js
/**
 * @typedef {(interfaceName: string, methodGuards: object, options: {sloppy: true}) => InterfaceGuard<Record<PropertyKey, MethodGuard>>} SloppyInterfaceGuard
 */
/**
 * @typedef {<M extends Record<PropertyKey, MethodGuard>>(interfaceName: string, methodGuards: M, options?: {sloppy?: boolean}) => InterfaceGuard<M>} StrictInterfaceGuard
 */
/**
 * @typedef {SloppyMakeInterfaceGuard | StrictMakeInterfaceGuard} MakeInterfaceGuard
 */
```

packages/patterns/src/types.js
```diff
@@ -524,17 +524,17 @@ export {};
 /** @typedef {(...args: any[]) => any} Method */
 
 /**
- * @template {Record<string | symbol, MethodGuard>} [T=Record<string | symbol, MethodGuard>]
+ * @template {Record<PropertyKey, MethodGuard>} [T=Record<PropertyKey, MethodGuard>]
  * @typedef {{
  *   interfaceName: string,
- *   methodGuards: { [K in keyof T]: K extends symbol ? never : T[K] },
+ *   methodGuards: { [K in keyof T as Exclude<K, symbol>]: T[K] },
```
gibson042, line 530: This _looks_ better, but I don't think it actually works correctly: [TypeScript Playground](https://www.typescriptlang.org/play?target=7&filetype=js&useJavaScript=truee=4#code/PQKhCgAIUgBAXAngBwKYBNUDNIG8CuAdgNaED2A7oQL6QCyq8AFmegOL4CGATulCMHDhQEaHHioAtsgA2nCXgBKqAMZleAHgAK3Mmm5IA0qkQAaeoxbsuvAHy0A2gBUAvMrWade1AcTGzFsysHDzotgC6-OIoGNh4uFGQkACWhBLcWJwqqABynJKoAFyQAM7w3KkA5qaJkAVB1qElxbiQDoYphJDEJmQ4TuHFHagAHhKE6CWliJIARmQykAD8kISoAG4+kMVO7eGQ1DVi1LRanIgyZJzoAPo5Gz78gsJgUQhSsvKoSqrq6Nq6fRGEzmBgNEJ2Ryudx-AHeXz+UGWYI2MKRMQIGKYHC4BJiJKpdKZbJ5ArFMoVQjVWr1KwQyYtNodVLdXr9SCcKYAURGKhk+EwGkM5hKM3mMgiOz2ByOMBOkDOFyutx5fIFqCeQhE0TQeAAFMhzpdrsVFcbbvdNtwAJSQFy2BVG5V3B7cWgCcBqQhlDkqbLIeBm52WrYuSCGpXXO0OiPmgDcWrAOu+uANTpNjsjKt5-MwtvtmfNN1VudQ7sEXp9WX9gfT2bVmDt4br0ebWYTnrI3vgkFm1wAkmkfMTvmG8UlOkSsrl8kVIAByABiZDI86OE9pKKaLSgE73kmKevzDtwh13e6SDgAymKFgA6JicwdlTiEbKDSBH1uZGQlVDrpJqHAIDwGrVAAyDa4XStPU+3QQcp2ya0EzAiC62LHN1VggchwyadkKAA)
```suggestion
 *   methodGuards: { [K in keyof T]: K extends symbol ? never : T[K] },
```

jsconfig.eslint-full.json
```diff
@@ -1,7 +1,9 @@
 {
   "extends": "./jsconfig.eslint-base.json",
   "include": [
+    "*.js",
     "**/*.js",
+    "*.ts",
     "**/*.ts"
   ]
```
gibson042, line 8: :+1:

packages/bundle-source/cache.js
```diff
@@ -107,13 +107,38 @@ export const makeBundleCache = (wr, cwd, readPowers, opts) => {
...
    */
-  const validate = async (meta, targetName, rootOpt, log = defaultLog) => {
+  const validate = async (
+    targetName,
+    rootOpt,
+    log = defaultLog,
+    meta = undefined,
+  ) => {
+    meta = await (meta ||
+      loadMetaText(targetName, log).then(metaText => JSON.parse(metaText)));
```
gibson042, line 138: Doesn't this violate our `await` rules in spirit, if not in fact? `loadMetaText` and `then` are both notionally capable of synchronous reentrancy/errors/etc.
```suggestion
    await null;
    if (!meta) {
      meta = await loadMetaText(targetName, log).then(metaText => JSON.parse(metaText));
    }
```

gibson042 (changes requested):
I like the `string | symbol` → `PropertyKey` changes, and am also leaving review of the bundling changes to @kriskowal unless directed to dive in more deeply.

packages/compartment-mapper/src/bundle.js
turadg: could it be this? https://github.com/terser/terser/blob/master/CHANGELOG.md#v5193

Endo is on terser 5.16.6 presently

packages/bundle-source/cache.js
kriskowal: We should also catch the JSON.parse error and annotate it with the source path.

packages/bundle-source/cache.js
```diff
@@ -107,13 +107,38 @@ export const makeBundleCache = (wr, cwd, readPowers, opts) => {
...
+
+  /**
    * @param {string} targetName
    * @param {*} rootOpt
    * @param {Logger} [log]
+   * @param {BundleMeta} [meta]
    * @returns {Promise<BundleMeta>}
    */
-  const validate = async (meta, targetName, rootOpt, log = defaultLog) => {
+  const validate = async (
```
kriskowal, line 131: Ah, a wiser me would have recognized that `validate` was public API. Thanks for back-filling.

packages/bundle-source/cache.js
```diff
@@ -69,7 +70,6 @@ export const makeBundleCache = (wr, cwd, readPowers, opts) => {
     // Prevent other processes from doing too much work just to see that we're
     // already on it.
     await metaWr.rm({ force: true });
-    await bundleWr.rm({ force: true });
```
kriskowal, line 72: Did this change obviate the need to address PID collisions?

kriskowal (approved):

packages/patterns/src/types.js
michaelfig: In the playground, you show the example of getting a desirable type error when passing a badly-constructed InterfaceGuard to a function.  However, introducing that behaviour leads to type errors when passing around well-formed InterfaceGuards if they have any symbol methods on them.

An [addendum to your playground](https://www.typescriptlang.org/play?target=7&filetype=js&useJavaScript=truee=4#code/PQKhCgAIUgBAXAngBwKYBNUDNIG8CuAdgNaED2A7oQL6QCyq8AFmegOL4CGATulCMHDhQEaHHioAtsgA2nCXgBKqAMZleAHgAK3Mmm5IA0qkQAaeoxbsuvAHy0A2gBUAvMrWade1AcTGzFsysHDzotgC6-OIoGNh4uFGQkACWhBLcWJwqqABynJKoAFyQAM7w3KkA5qaJkAVB1qElxbiQDoYphJDEJmQ4TuHFHagAHhKE6CWliJIARmQykAD8kISoAG4+kMVO7eGQ1DViSSUz8zIMDSG8JUvF7uroGgAUPYh9kE4AlJAAZNNzBbmS5Wa5hI4wai0LScRAyMicdAAfRyGx8-EEwjAUQQUlk8lQSlUj20un0RhMwMswRsYUcrgenjJPgpARBNNCERxSDQmBwuASx066Uy2TyBWKZQqhGqtXqoNpzTwbQ6qW6vX6kE4UwAoiMVDJ8JgNIZzKdATIIjs9gcIUkAed2Y0bndIIynq8NZ8fv9zecqVdabY7VDIDC4QjkXqDUbUBihCJomg8M9kLD4YjiuGM8jUZtuD8XLYw+nIyi0dxaAJwGpCGUtSpssh4Nmy3mti5IGmI4jIEWu6XEQBuBNgJOE3CpwfoLPTpHRw2YQvF1uI+f6xeoKuCWv1rJNltzhexvsDnvoPvF7s5kc1sh1+CQWaIgCSaR8osJncF9tSIqyuT5EUkAAOQAGJkGQIFHPa8ocjcLRQPayGSMUzzLnghxIchSQOAAymcCwAHRMNqb5lJwhDZIMkDoZekCZDIJSoDBSTUOA7HgPuqDNquuYVs8z7oG+-7ZF8I7cbxR4brGgmvu+GQAeJo4wAgMR4Hx5b5horSoas+BzD45j4YRMgkWRD6UdRxSEAZsxbNQ9jQDu971pUkHCQpn5aR2eDYcKH4AeKwHgZB0H+XBzqTIhOGwcUACMABMADM-lYchfoLE6YJKj+sUmRa5klOR8BWagNFJalyHsdQt6Jmpya4Jpx7GrpNl2UZbQEYVpHFZZVHle1hncAcTnVruj7uawImBdk64xpgp55QFilikBxShVBrF1NSUW5f5KEJSlaXbZlFy7TlMWxV1plFSVZUVcd1UcSOQA) shows how the `never` type prevents a legitimate value from being cast to an InterfaceGuard.

So that addendum shows the `never` typing raises the same error even for a well-formed InterfaceGuard.   I want to be able to carry along the methodGuards (all PropertyKeys) as a type parameter so that we can manipulate it in the exo makers (as I've done for the checking of guards against methods).

I think it's related to covariance/contravariance, but I'm not familiar with those terms enough to use them.  If you have a better solution you can demonstrate in the playground, please do let me know.

packages/bundle-source/cache.js
michaelfig: I don't think it violates the `await` spirit.  The subexpression is in the synchronous prelude, and the first `await` is still unconditional.

However, I'll rewrite it to be clearer anyway.  Thanks both!

packages/patterns/src/types.js
gibson042: Alright, I think I've got it: [TypeScript Playground](https://www.typescriptlang.org/play?target=7&ssl=10&ssc=6&pln=10&pc=115&filetype=js&useJavaScript=truee=4#code/PQKhCgAIUgBAXAngBwKYBNUDNIG8CuAdgNaED2A7oQL6QCyq8AFmegOL4CGATulCMHBDQEaHHioAtsgA2nCXgBKqAMZleAHgAK3Mmm5IA0qkQAaeoxbsuvAHy0A2gBUAvMrWade1AcTGzFsysHDzotgC6-OIoGNh4uFGQkACWhBLcWJwqqABynJKoAFyQAM7w3KkA5qaJkAVB1qElxbiQDoYphJAAogAeKjL4mBrEJmQ4TuYliJIARmQyEcVO7eGQtABkkFo88MmcMhqt7Z09veVZ8CNjE1Mz84vhxYSoAG4+67Y1YknTcwsMBohXglAD8LTaHVSZwuKiuo0Q40gk1K9wWS2Rq3W3xg1FoO0QMjInD40EEwjA0TQeAAFMhOITiehigSiSSAJSQFy2bYMtnoWgCcBqQhlSBZbLIeCsplcyD0xkkrk8hX8gDcQnAIipqDwMpJgsEIrFlTIrAAkmkfJlsnKEkkkql0jbcvkipAAOSm1ge74O+pWYHoZp4KAO8OSFrUP1JaNh35omSAwM2YMtePhtoAZUTADomJwSpaypxCNknng4+G49QNRLUFL9egad70JbnVlUOyNRSYAgYnq+UyjnVnvg5j5zA4c-8ZPnC8X4KXy2OJ9xPobhWRRfBIK329bOwB9Qg8XQUO0Zp2H7J5ArFL1m9Ans+UX0ZgPBVMh+2ZpKRyAAEYACYAGYY2xDM-geZMvyadM-ySac8wLIsd2XVAKxA8CMxrOsVElaUhxJFsnwPDJj1Pbhz27TVtX7alcCbTdjV3WYSXIl0j2ghZL3Da8KNvN0H3Y58eJkd8I0sOCQQQxCANwKtEOQ2d5zQksy0wqMILw8B60bYjm1Ezjj3E2jex1QdFXQEcAMIcdZknbMUIXdDNIrey1w3Mktx3SBjKtQTUCPSRkhKEoqj4h0BJdO93Q9USQrCiLCEqST-WkxpZMrCDxNgrK0xy8Baz0giGyI6yaQCjtsiS8KqnMrVKQY3UmMM2zV0c7gpxnB41MXDCPIcj5qHsHzWP8jjAq41Bzm4TgosdabOzikSSSPWaLnS-9MqDH8MwjYpsIg8NNvm7TcNyxN8r2uS-xUvrUIG9yjrAk6HQehZcyIEo1HpWYZFQEoK0UnSahK-SKv5Kqppq4Kzs4bsgA)

```
methodGuards: { [K in Exclude<keyof T, symbol>]: T[K] } & Partial<{ [K in Extract<keyof T, symbol>]: never }>
```

michaelfig: > Alright, I think I've got it

Excellent!  Thank you for your tenacity. 😄

Is there anything else you'd like me to do as part of your Request Changes?

packages/bundle-source/cache.js
michaelfig: IOW, I think we still need to account for intra-process collisions.

kriskowal: This might suffice https://github.com/endojs/endo/pull/1760

It might make sense to pull the now-optional `pid` argument from the maker. But, it might also make sense to keep the PID (so another tool can clean up dead scratch detritus) and add a nonce to the scratch file names. I’ll take that conversation there.

packages/patterns/src/types.js
```diff
@@ -524,17 +529,23 @@ export {};
...
 
 /**
- * @template {Record<string | symbol, MethodGuard>} [T=Record<string | symbol, MethodGuard>]
+ * @template {Record<PropertyKey, MethodGuard>} [T=Record<PropertyKey, MethodGuard>]
  * @typedef {{
  *   interfaceName: string,
- *   methodGuards: { [K in keyof T]: K extends symbol ? never : T[K] },
- *   symbolMethodGuards?: CopyMap<(keyof T) & symbol, MethodGuard>,
+ *   methodGuards:
+ *     { [K in Exclude<keyof T, symbol>]: T[K] } &
```
gibson042, line 536: This also has a shorter spelling, albeit one that [TypeScript documentation](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys) does not claim _official_ support for (which is probably accidental: https://github.com/microsoft/TypeScript-Website/issues/2937 ).
```suggestion
 *     Omit<T, symbol> &
```

Either is fine with me.

packages/patterns/src/types.js
```diff
@@ -524,17 +529,23 @@ export {};
...
+ *     Partial<{ [K in Extract<keyof T, symbol>]: never }>,
+ *   symbolMethodGuards?:
+ *     CopyMap<
+ *       Extract<keyof T, symbol>,
+ *       T[Extract<keyof T, symbol>]
+ *     >,
```
gibson042, line 542: Huh, interesting. I'm surprised that something like `T[keyof T]` works, but indeed it does.

gibson042 (approved):

packages/bundle-source/cache.js
kriskowal: I think we should delete the previous `metaWr.rm` as well. These are dross from the revision before atomic renames and in the new model, removing the old files is not necessary.

packages/patterns/src/types.js
```diff
@@ -491,15 +491,20 @@ export {};
...
+ * } SloppyMakeInterfaceGuard
+ */
+/**
+ * @typedef {<M extends Record<PropertyKey, MethodGuard>>(
+ *   interfaceName: string,
+ *   methodGuards: M,
+ *   options?: {sloppy?: boolean}) => InterfaceGuard<M>
+ * } FallbackMakeInterfaceGuard
  */
+/** @typedef {SloppyMakeInterfaceGuard & FallbackMakeInterfaceGuard} MakeInterfaceGuard */
```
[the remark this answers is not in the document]
michaelfig: The different method signatures in an intersection type are presented in order as overloads, with the last one being the most general.  In this case, the last one is not "Strict"er than the "Sloppy" one.  I thought of "Fallback" or "Default" for this, but maybe "General" would be the best?

PR #1762, opened 2023-09-08
fix(bundle-source): Add a nonce to bundle-source scratch files

kriskowal:
## Description

The bundle-source tool writes out to temporary scratch files to avoid race conditions between concurrent runs. The scratch file currently includes the PID of the bundle-source process so they do not collide and to make it easy to determine whether the scratch file can be deleted if bundle-source is interrupted before renaming the scratch file.

This does not account for use of the bundle cache as a library, where multiple bundle caches may exist concurrently in a single process, making the PID insufficient to disambiguate. Also, the PID argument is optional, so a nonce can disambiguate if a PID is not available. The new scratch file has both the PID and nonce, like `x.100.1000.json` or `x.no-pid.2000.json`.

### Security Considerations

Not to my knowledge.

### Scaling Considerations

This change includes concurrency tests for both the tool and library usage of the bundler.

This change does not alter the usage of the `bundle-source` command. `NEWS.md` includes a suggestion to use the new `pid` argument of the cache maker, which should otherwise be discovered on first use. We do not otherwise currently have documentation for the cache library in `README.md`, so discoverability is limited. This is probably fine for now.

### Testing Considerations

This adds testing for the cache command and library usage.

### Upgrade Considerations

This should not impact upgrades.


kriskowal:
@dckc FYI, cache maker options `pid` and `nonce` exist and have reasonable defaults.

packages/bundle-source/NEWS.md
```diff
@@ -1,4 +1,12 @@
 
+# Next
+
+- The `@endo/bundle-source/cache.js` maker now accepts additional optional
+  arguments: `pid` and `nonce`, with reasonable defaults.
+  Providing the `pid` can make it easier to clean up scratch files if
+  the bundler is interrupted, since deleting any scatch file with the PID of a
```
michaelfig, line 7:
```suggestion
  the bundler is interrupted, since deleting any scratch file with the PID of a
```

packages/bundle-source/src/fs.js
```diff
@@ -121,19 +121,23 @@ export const makeFileWriter = (
...
+  make = there => makeAtomicFileWriter(there, { fs, path }, pid, nonce, make),
 ) => {
+  const pidText = `${pid || 'no-pid'}`;
+  const nonceText = `${nonce || 'no-nonce'}`;
   const writer = makeFileWriter(fileName, { fs, path }, make);
   return harden({
     ...writer,
     atomicWriteText: async (txt, opts) => {
-      const scratchName = `${fileName}.${pid}.scratch`;
+      const scratchName = `${fileName}.${pidText}.${nonceText}.scratch`;
```
michaelfig, line 140: Consider the following, to make globs easier `rm -rf *.19837.scratch`:

```suggestion
      const scratchName = `${fileName}.${nonceText}.${pidText}.scratch`;
```

michaelfig (approved):
My approval is conditional on the consideration of my comments.

michaelfig:
Can you please also adjust the behaviour of Agoric/agoric-sdk#8314 to provide `process.pid` instead of `crypto.randomInt` to the bundler (letting the bundler generate the nonce)?

PR #1764, opened 2023-09-11
WIP fix(patterns,exo)!: args, splitArray, and splitRecord now default to …

erights:
Experiment re surprise that method guards with no declared rest by default accept extra args. This PR represents the most extreme and least compat "fix" to this surprise -- changing the semantics of `splitArray` and `splitRecord` so they also reject rest parts by default, i.e., if no rest argument is provided. Because of the way method guards delegate to splitArray, this takes care of the method guard "problem" as well.

Other options:
- Fix only method guards, so that they reject extra args by default, by changing how it delegates to splitArray. This "solution" would not change the semantics of splitArray at all.
- Grandfather in exactly the current tolerant behavior of all three: method guards, splitArray, and splitRecord. This tolerant behavior of methodGuards can be rationalized as reflecting the tolerant behavior of JS positional function parameters.
- Preserve the tolerant behavior, but not pass the values of the extra args through to the protected raw behavior method. This would preserve the input validation intent of method guards while still accommodating API growth and thus version slippage. (suggested by @turadg )


mhofman:
> * Preserve the tolerant behavior, but not pass the values of the extra args through to the protected raw behavior method. This would preserve the input validation intent of method guards while still accommodating API growth and thus version slippage. (suggested by @turadg )

I'm in favor of this approach as it's the one I had initially suggested as the behavior I'd expect.

PR #1765, opened 2023-09-12
fix(exo)!: extra undeclared args dropped

erights:
Alternative to https://github.com/endojs/endo/pull/1764 experiment.

Of the options listed there, this one

> Preserve the tolerant behavior, but not pass the values of the extra args through to the protected raw behavior method. This would preserve the input validation intent of method guards while still accommodating API growth and thus version slippage.

As suggested by @mhofman and @turadg

### DIFF
--- a/packages/exo/src/exo-tools.js
+++ b/packages/exo/src/exo-tools.js
@@ -32,6 +32,13 @@ const { defineProperties, fromEntries } = Object;
  */
 const MinMethodGuard = M.call().rest(M.any()).returns(M.any());
 
+/**
+ * @param {Passable[]} args
+ * @param {MethodGuardPayload} methodGuardPayload
+ * @param {string | undefined} label
+ * @returns {Passable[]} Returns the args that should be passed to the
+ * raw method
+ */
 const defendSyncArgs = (args, methodGuardPayload, label) => {
   const { argGuards, optionalArgGuards, restArgGuard } = methodGuardPayload;
   const paramsPattern = M.splitArray(
@@ -40,6 +47,15 @@ const defendSyncArgs = (args, methodGuardPayload, label) => {
     restArgGuard,
   );
   mustMatch(harden(args), paramsPattern, label);
+  if (restArgGuard !== undefined) {
+    return args;
+  }
+  const declaredLen =
+    argGuards.length + (optionalArgGuards ? optionalArgGuards.length : 0);
+  if (args.length <= declaredLen) {
+    return args;
+  }
+  return args.slice(0, declaredLen);
 };
 
 /**
@@ -53,8 +69,8 @@ const defendSyncMethod = (method, methodGuardPayload, label) => {
   const { syncMethod } = {
     // Note purposeful use of `this` and concise method syntax
     syncMethod(...args) {
-      defendSyncArgs(harden(args), methodGuardPayload, label);
-      const result = apply(method, this, args);
+      const realArgs = defendSyncArgs(harden(args), methodGuardPayload, label);
+      const result = apply(method, this, realArgs);
       mustMatch(harden(result), returnGuard, `${label}: result`);
       return result;
     },
@@ -83,9 +99,9 @@ const desync = methodGuardPayload => {
   return {
     awaitIndexes,
     rawMethodGuardPayload: {
+      ...methodGuardPayload,
       argGuards: rawArgGuards.slice(0, argGuards.length),
       optionalArgGuards: rawArgGuards.slice(argGuards.length),
-      restArgGuard,
     },
   };
 };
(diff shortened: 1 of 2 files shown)


erights:
Reviewers,

I made this a `fix(exo)!:` rather than a `fix(exo):` because it technically is a compat break. However, it is unlikely to currently be a compat break in practice, so let me know if you think I should remove the `!`.

packages/exo/src/exo-tools.js
```diff
@@ -32,6 +32,13 @@ const { defineProperties, fromEntries } = Object;
  */
 const MinMethodGuard = M.call().rest(M.any()).returns(M.any());
 
+/**
+ * @param {Passable[]} args
+ * @param {MethodGuardPayload} methodGuardPayload
+ * @param {string | undefined} label
```
turadg, line 38: consider JSDoc convention:
```suggestion
 * @param {string} [label]
```

packages/exo/src/exo-tools.js
```diff
@@ -40,6 +47,15 @@ const defendSyncArgs = (args, methodGuardPayload, label) => {
...
   mustMatch(harden(args), paramsPattern, label);
+  if (restArgGuard !== undefined) {
+    return args;
+  }
+  const declaredLen =
+    argGuards.length + (optionalArgGuards ? optionalArgGuards.length : 0);
+  if (args.length <= declaredLen) {
+    return args;
+  }
+  return args.slice(0, declaredLen);
```
turadg, line 58: consider documenting,
```suggestion
  // ignore extraneous arguments, as a JS function call would do
  return args.slice(0, declaredLen);
```

packages/exo/src/exo-tools.js
```diff
@@ -53,8 +69,8 @@ const defendSyncMethod = (method, methodGuardPayload, label) => {
   const { syncMethod } = {
     // Note purposeful use of `this` and concise method syntax
     syncMethod(...args) {
-      defendSyncArgs(harden(args), methodGuardPayload, label);
-      const result = apply(method, this, args);
+      const realArgs = defendSyncArgs(harden(args), methodGuardPayload, label);
```
turadg, line 72: please also update the signature to qualify the args
```suggestion
      const realArgs = defendSyncArgs(harden(rawArgs), methodGuardPayload, label);
```

packages/exo/test/test-heap-classes.js
```diff
@@ -10,6 +10,19 @@ import {
...
+  foo: M.call().returns(),
+});
+
+test('what happens with extra arguments', t => {
+  const exo = makeExo('WithExtra', NoExtraI, {
+    foo(x) {
+      t.is(x, undefined);
+    },
+  });
+  exo.foo(8);
```
turadg, line 23: takes a bit to know this value is irrelevant.
```suggestion
  exo.foo('an extra arg');
```

turadg (commented):
I agree that a) this is technically a breaking change and b) it's unlikely to break anything.

Since the package version hasn't reached 1.0, there's little downside to making a breaking change. So I think keep the `!` to err on the side of being loud about it.

packages/exo/src/exo-tools.js
erights: I had not intended to make it optional, but rather, merely allow its value to be `undefined`. It isn't exported and its two callers in this file provide the `label` argument explicitly.

However, I don't mind making it optional, so done.

packages/exo/src/exo-tools.js
erights: Done.

packages/exo/src/exo-tools.js
erights: I don't understand. There is no `rawArgs` in scope here. Do you suggest that I rename the syncMethod rest parameter to `rawArgs`?

> please also update the signature to qualify the args

I don't understand this either. The returned `syncMethod` as a whole is declared as `Method`. What do you have in mind?

packages/exo/test/test-heap-classes.js
erights: Done

packages/exo/src/exo-tools.js
mhofman: I believe the request is to rename `args` -> `rawArgs` to make it clearer what args represent

mhofman (approved):
LGTM

packages/exo/src/exo-tools.js
turadg: right. I would have edited both lines but the "suggestion" tool only works on lines modified in the PR.
```js
    syncMethod(...rawArgs) {
      const realArgs = defendSyncArgs(harden(rawArgs), methodGuardPayload, label);
```

erights: Gotcha. This rename makes a different distinction than I had in mind, but a better one. Accordingly, I renamed this one and the existing occurrences of `rawArgs`, and one more, to `syncArgs`. Reads much better. Done.

PR #1766, opened 2023-09-12
feat(pass-style): far class instances

erights:
Support use of JS classes to define far objects as class instances.

This was already possible, as demonstrated by prior tests in test-passStyleOf.js . This PR merely provides a base class to make this easier, and tests it.

### DIFF
--- a/packages/pass-style/src/far-class-instances.js
+++ b/packages/pass-style/src/far-class-instances.js
@@ -1,13 +1,13 @@
 import { Far } from './make-far.js';
 
 /**
- * Classes whose instances should be Far objects should inherit from
- * this convenient base class. Note that the constructor of this base class
+ * Classes whose instances should be Far objects may find it convenient to
+ * inherit from this base class. Note that the constructor of this base class
  * freezes the instance in an empty state, so all is interesting attributes
  * can only depend on its identity and what it inherits from.
  * This includes private fields, as those are keyed only on
  * this object's identity. However, we discourage (but cannot prevent) such
- * use of private fields, as they cannot easily be refactored into exo state.
+ * use of private fields, as they cannot easily be refactored into Exo state.
  */
 export const FarBaseClass = class FarBaseClass {
   constructor() {
--- a/packages/pass-style/src/far-class-instances.js
+++ b/packages/pass-style/src/far-class-instances.js
@@ -1,0 +1,19 @@
+import { Far } from './make-far.js';
+
+/**
+ * Classes whose instances should be Far objects should inherit from
+ * this convenient base class. Note that the constructor of this base class
+ * freezes the instance in an empty state, so all is interesting attributes
+ * can only depend on its identity and what it inherits from.
+ * This includes private fields, as those are keyed only on
+ * this object's identity. However, we discourage (but cannot prevent) such
+ * use of private fields, as they cannot easily be refactored into exo state.
+ */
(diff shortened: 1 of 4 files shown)


packages/pass-style/src/far-class-instances.js
```diff
@@ -0,0 +1,19 @@
+import { Far } from './make-far.js';
+
+/**
+ * Classes whose instances should be Far objects should inherit from
+ * this convenient base class. Note that the constructor of this base class
+ * freezes the instance in an empty state, so all is interesting attributes
+ * can only depend on its identity and what it inherits from.
+ * This includes private fields, as those are keyed only on
+ * this object's identity. However, we discourage (but cannot prevent) such
+ * use of private fields, as they cannot easily be refactored into exo state.
```
turadg, line 10: should it be "Exo"?

packages/pass-style/src/far-class-instances.js
```diff
@@ -0,0 +1,19 @@
+import { Far } from './make-far.js';
+
+/**
+ * Classes whose instances should be Far objects should inherit from
```
turadg, line 4: "should" … why? Please expand with what risk the developer is taking if they don't heed the advice.

packages/pass-style/test/test-far-class-instances.js
```diff
@@ -0,0 +1,80 @@
...
+}
+
+test('far class instances', t => {
+  const fb = new FarBaseClass();
+  t.is(passStyleOf(fb), 'remotable');
+  t.deepEqual(getMethodNames(fb), ['constructor']);
+
+  t.assert(new fb.constructor() instanceof FarBaseClass);
+  t.throws(() => fb.constructor(), {
+    // TODO message depends on JS engine, and so is a fragile golden test
```
turadg, line 37: how can this TODO be solved?

isn't it fair to assume all of Endo's tests run in whatever its CI matrix covers?

turadg (commented):
Stoked to see this working

packages/pass-style/src/far-class-instances.js
erights: Good point. Fixed to

> Classes whose instances should be Far objects may find it convenient to inherit from this base class.

Done.

packages/pass-style/src/far-class-instances.js
erights: Fixed

packages/pass-style/test/test-far-class-instances.js
erights: Attn @kriskowal @warner 

To truly not rely on any unspecified JS, all I may test is the error type. But I do like ensuring that the test case is failing for the reason it should. I don't know how to do that without this fragile dependency.

Recommendations?

mhofman (commented):
My main concern with this PR is `this`, and the difference with `ExoClass`. Since there is no similar way to handle state, users will simply rely on private fields, which we can't prevent. Also they are free to add accessors, which won't satisfy `passStyleOf` checks.

I would prefer supporting class syntax in a way compatible with our encapsulation paradigm (instance private state, not class scope private state), and making sure we fail early for class definitions that are not actually remotable.

erights:
See also https://github.com/endojs/endo/pull/1773

erights:
Closing in favor of https://github.com/endojs/endo/pull/1809

PR #1770, opened 2023-09-14
test: additional tests of splitRecord

turadg:
More test coverage of `M.splitRecord`

### DIFF
--- a/packages/patterns/test/test-patterns.js
+++ b/packages/patterns/test/test-patterns.js
@@ -330,10 +330,17 @@ const runTests = (t, successCase, failCase) => {
       '{"bar":4,"foo":3} - Must be >= {"baz":3}',
     );
 
+    successCase(specimen, M.splitRecord({}, undefined, undefined));
+    successCase(specimen, M.splitRecord({}, { unused: M.string() }));
     failCase(
       specimen,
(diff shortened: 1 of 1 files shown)


packages/patterns/test/test-patterns.js
```diff
@@ -330,11 +330,18 @@ const runTests = (t, successCase, failCase) => {
       '{"bar":4,"foo":3} - Must be >= {"baz":3}',
     );
 
+    successCase(specimen, M.splitRecord({}, undefined, undefined));
+    successCase(specimen, M.splitRecord({}, { unused: M.string() }));
```
erights, line 334: Whether it was a good or bad convention, I seem to have followed a convention of grouping success cases together, followed by fail cases. I don't feel strongly about this, but I'd prefer these success cases be moved up to be in the success case group.

erights (approved):
LGTM, thanks!

packages/patterns/test/test-patterns.js
turadg: As a reader, whether something is a success or fail case is starkly apparent from the function call.

What's not so apparent is what tests are aspects of the same behavior. I've used the ordering here to group that and I think it's wortwhile.

PR #1771, opened 2023-09-16
fix(patterns): pass type parameter from interface guard to payload

erights:
The previous endo release added `getInterfaceGuardPayload` so that agoric-sdk could start to use it, rather than accessing the fields of an interface guard directly. This paves the way for https://github.com/endojs/endo/pull/1712 . However, because the type of `getInterfaceGuardPayload` did not propagate the type parameter from the guard argument to the returned payload, thehttps://github.com/Agoric/agoric-sdk/pull/8339  attempt to switch to it causes type errors due to the loss of this type information.


michaelfig (approved):
Straightforward!  Sorry for the omission.

PR #1773, opened 2023-09-18
fix(exo): methods might be non-enumerable

erights:
Steps towards using a JS class' `.prototype` as behavior methods for exos. This first step mostly just corrects an error of missing non-enumerable methods, which of course is the normal case for actual class `.prototype`s

Staged on https://github.com/endojs/endo/pull/1766


erights:
Closing in favor of https://github.com/endojs/endo/pull/1809

PR #1775, opened 2023-09-18
fix(static-module-record): Bind `this` and `arguments` more correctly

kriskowal:
## Description

Currently, the `this` value in module scope is `globalThis`. This is a gap in fidelity, since it should be `undefined`, but is not a leak. This change alters the signature of the module functor generated by `StaticModuleRecord` such that `this` is bound to `undefined` with the existing instantiation code in `ses`. This also ensures that the undeniable `arguments` lexical name reveals nothing about the evaluator or module calling convention to the module.

This will break any program that depends on `this` being bound to `globalThis`. Since all such programs were already invalid and would not work in other environments like Node.js, I am choosing to treat this as a non-breaking-change, even though it may require bug-fixes elsewhere to reconcile versions.

### Security Considerations

This change does not add or remove any capabilities from the scope of modules, just requires that `globalThis` to be spelled out in full.

### Scaling Considerations

This should not affect scale.

This makes behavior more consistent with expectations documented for ESM generally.

### Testing Considerations

I’ve included a `ses` test that was breaking before the change to `static-module-record`.

### Upgrade Considerations

This change will alter the hash of the Agoric kernel modules and will get picked up when bundles get regenerated. Some contracts may have bugs where `this` is assumed to be `globalThis`.

### DIFF
--- a/packages/static-module-record/src/transform-analyze.js
+++ b/packages/static-module-record/src/transform-analyze.js
@@ -82,23 +82,24 @@ const makeCreateStaticRecord = transformSource =>
       })
       .join('');
 
-    // The functor captures the SES `arguments`, which is definitely
-    // less bad than the functor's arguments (which we are trying to
-    // hide).
-    //
-    // It must also be strict to enforce strictness of modules.
-    // We use destructuring parameters, so 'use strict' is not allowed
-    // but the function actually is strict.
+    // The outer function destructures the module calling convention's internal
+    // variables into hidden lexical variables.
+    // The inner function binds `this` to `undefined` and overshadows the
+    // evaluator's `arguments` with a completely empty `arguments` object.
+    // There is no avoiding the overshadowing of `globalThis.arguments` if it
+    // exists in this emulation of ESM since the evaluator binds `arguments` as
+    // well.
+    // Relies on the evaluator to ensure these functions are strict.
     let functorSource = `\
-(({ \
+({ \
   imports: ${h.HIDDEN_IMPORTS}, \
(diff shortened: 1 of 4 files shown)


packages/static-module-record/src/transform-analyze.js
```diff
@@ -82,23 +82,25 @@ const makeCreateStaticRecord = transformSource =>
...
+    // `undefined`.
+    // The inner function ensures that `arguments` is empty instead of
+    // revealing the module system internals or the evaluator's arguments.
+    // There is no avoiding the overshadowing of `globalThis.arguments` if it
+    // exists in this emulation of ESM since the evaluator binds `arguments` as
+    // well.
+    // Relies on the evaluator to ensure these functions are strict.
     let functorSource = `\
-(({ \
+(function ({ \
```
erights, line 95: For shedding, `this`, only the inner function matters. The outer function can remain an arrow function.

```suggestion
(({ \
```

packages/static-module-record/src/transform-analyze.js
```diff
@@ -82,23 +82,25 @@ const makeCreateStaticRecord = transformSource =>
...
   imports: ${h.HIDDEN_IMPORTS}, \
   liveVar: ${h.HIDDEN_LIVE}, \
   onceVar: ${h.HIDDEN_ONCE}, \
   importMeta: ${h.HIDDEN_META}, \
- }) => { \
+ }) { (function () { \
```
erights, line 100:
```suggestion
 }) => { (function () { \
```

erights (approved):
LGTM, thanks!

Issue #1776, opened 2023-09-18
compartment-mapper: add policy types


boneskull:
## What is the Problem Being Solved?

Policies are essentially untyped (considered to be arbitrary objects).  It would be helpful for consumers to have access to type information so that they do not construct invalid policies.

## Description of the Design

A policy has a shape defined already; it's just a matter of formalizing that shape into types.

## Security Considerations

none

## Scaling Considerations

none

## Test Plan

I'd probably want to add `tsd` tests for this; while it looks like `tsd` is used in other workspaces, I'll need to add the `tsd` package to `compartment-mapper`.

dckc:
See also:

 - https://github.com/Agoric/agoric-sdk/issues/6160

kriskowal:
Closed by #1869

PR #1778, opened 2023-09-18
feat(compartment-mapper): add policy-related types

boneskull:
closes: #1776

## Description

- Add some types for the policy and export them from `compartment-mapper`.
- Remove some "alias"-style typedefs
- Change `assertPackagePolicy` to be a type guard.

### Security Considerations

none

### Scaling Considerations

none

This could cause compiler errors in consumers already using policies (if those consumers exist), since the types have been narrowed.  

### Testing Considerations

This adds tests using `tsd`, which requires adding that dev dependency to `compartment-mapper` and modifying its `test` script.

### Upgrade Considerations

This should not impact live production systems; it may impact builds.  Because it may impact builds, this could be construed as a breaking change, depending on Endo's policy or lack thereof.  

* * *

This is currently a **DRAFT** and should be considered in-progress.

### DIFF
--- a/packages/compartment-mapper/demo/policy/index.mjs
+++ b/packages/compartment-mapper/demo/policy/index.mjs
@@ -8,7 +8,11 @@ lockdown({
 
 import fs from 'fs';
 
-import { importLocation, makeArchive, parseArchive } from '../../index.js';
+import {
+  importLocation,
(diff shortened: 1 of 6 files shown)


boneskull:
Closed in lieu of #1839

PR #1779, opened 2023-09-18
fix(eslint-plugin): fix overzealous jsdoc typechecks

### LINKED ISSUE #1755: @endo/eslint-plugin: something should be done w/ jsdoc/check-types
I know someone put a lot of time into the issue templates, so I apologize that I'm not using them. It wasn't clear which one I should use.

**TL;DR: plz disable `jsdoc/check-types`**

I see warnings coming out of [eslint-plugin-jsdoc](https://npm.im/eslint-plugin-jsdoc) of this variety:

```js
/**
 * @typedef {Object} MyTypedef
 * @property {string} whatever
 */
```

causing:

```
Invalid JSDoc @typedef "MyTypedef" type "Object"; prefer "object"
```

I'm sure there's a reason for this (it appears to be the default behavior), but the plugin _doesn't have enough context_ to make a helpful recommendation.  This assumes that Endo even wants to let this plugin tell us there's something wrong with our types (instead of e.g., TypeScript itself).

> I _do_ think this is reasonable in JSDoc proper or closure-compiler-land, perhaps, but those are not applicable here.

I kind of _implicitly understood_ this before, but I have now explicitly tested my assumption. _There is no difference_ between these typedefs, as far as TypeScript is concerned:

```js
/**
 * @typedef {Object} MyTypedef
 * @property {string} whatever
 */

/**
 * @typedef MyTypedef
 * @property {string} whatever
 */

// this last one triggers a warning:

/**
 * @typedef {object} MyTypedef
 * @property {string} whatever
 */
```

Generally, it appears TS will (internally or otherwise) turn typedefs into `interface`s when they "extend" `object`, `Object`, or nothing; everything else is a `type`s. The above all compile to:

```ts
interface MyTypedef {
  whatever: string;
}
```

So in the _above_ context, any warning from `eslint-plugin-jsdoc` is just hand-wringing, because it doesn't matter.

There's a distinction between `object` and `Object` in TypeScript, of course, which can be seen when using typedefs _as type aliases_:

```js
// jsdoc won't complain about this

/**
 * @typedef {object} MyNotPrimitive
 */

// jsdoc will complain about this 

/**
 * @typedef {Object} MyActualObject
 */
```

The above compiles to:

```ts
type MyNotPrimitive = object;

type MyActualObject = Object;
```

...which is expected (`object` can be thought of as "not a primitive" and `Object` is more akin to the intrinsic--but that's about the limit of my understanding).

This distinction is also respected in function signatures; `@param {Object} foo` is not `@param {object} foo`; `@returns {Object}` is not `@returns {object}`.

* * *

All that above is to write that _I_ don't trust the `jsdoc/check-types` rule and think it should be disabled.  Failing that, it should not warn about `Object` since perhaps it's helpful otherwise.

boneskull:
closes: #1755

## Description

**TL;DR:** ; ESLint will no longer coerce `Object` to `object` in `@typedef` tags.

This does two things:

- Uses the `recommended-typescript-flavor` config from `eslint-plugin-jsdoc`, which is recommended for the types-in-JavaScript set.  This only disables `jsdoc/no-undefined-types`, which can be safely removed from the rule overrides.
- Configures rule `jsdoc/check-types` to allow `Object` in `@typedef` tags _only_; see #1755 for explanation of why coercing `Object` to `object` has potentially unintended consequences but otherwise can be used interchangeably in the vast majority of circumstances.

You may read [the rule docs](https://github.com/gajus/eslint-plugin-jsdoc/blob/3d052ba/docs/rules/check-types.md#readme) about why it is the way it is.  I find the rationale to be entirely reasonable _excepting_ in the case of `Object` vs `object`--which is why I did not disable the rule entirely.

### Security Considerations

none

### Scaling Considerations

none

none

### Testing Considerations

none

### Upgrade Considerations

none

### DIFF
--- a/packages/eslint-plugin/lib/configs/style.js
+++ b/packages/eslint-plugin/lib/configs/style.js
@@ -1,5 +1,9 @@
 module.exports = {
-  extends: ['airbnb-base', 'plugin:jsdoc/recommended', 'prettier'],
+  extends: [
+    'airbnb-base',
+    'plugin:jsdoc/recommended-typescript-flavor',
+    'prettier',
+  ],
   rules: {
     quotes: [
       'error',
@@ -38,7 +42,6 @@ module.exports = {
     'no-inner-declarations': 'off',
 
     'jsdoc/no-multi-asterisks': ['warn', { allowWhitespace: true }],
-    'jsdoc/no-undefined-types': 'off',
     'jsdoc/require-jsdoc': 'off',
     'jsdoc/require-property-description': 'off',
     'jsdoc/require-param-description': 'off',
@@ -47,6 +50,10 @@ module.exports = {
     'jsdoc/require-yields': 'off',
     'jsdoc/tag-lines': 'off',
     'jsdoc/valid-types': 'off',
+    'jsdoc/check-types': [
+      'warn',
+      { exemptTagContexts: [{ tag: 'typedef', types: ['Object'] }] },
+    ],
 
     'no-unused-vars': [
       'error',


packages/eslint-plugin/lib/configs/style.js
```diff
@@ -47,6 +50,10 @@ module.exports = {
     'jsdoc/require-yields': 'off',
     'jsdoc/tag-lines': 'off',
     'jsdoc/valid-types': 'off',
+    'jsdoc/check-types': [
+      'warn',
+      { exemptTagContexts: [{ tag: 'typedef', types: ['Object'] }] },
```
turadg, line 55: Your issue makes a good case for allowing `Object`. However, TypeScript [discourages `Object`](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#:~:text=%E2%9D%8C%20Don't%20ever%20use,used%20appropriately%20in%20JavaScript%20code.) 

Moreover, this is a "style" config so it's meant to be have opinions. Even insofar as they're interchangeable, using one consistently is worthwhile.

turadg (commented):
Thanks for your contribution!

packages/eslint-plugin/lib/configs/style.js
boneskull: It's true that TypeScript discourages `Object` _in TypeScript_, but the rules for JS usage are different, because TS maintains some modicum of compatibility with JSDoc.

If I had my way, all object-based `@typedefs` would just _omit_ `{Object}` or `{object}`, since it's implied.  In fact, we can do that without changing any configuration; it's not enforced consistently as-is.

This is likely a situation in which I am the only one really bothered by the way it is currently.  If the change seems unreasonable to others, I will close this.

kriskowal:
I’m closing to clean the shop floor and don’t foresee us scheduling time to settle a disagreement on this narrow of a concern.

PR #1785, opened 2023-09-21
feat(pass-style): Safe promises can override @@toStringTag with a string

erights:
https://github.com/endojs/endo/pull/1781 demonstrates that the current endo (and it just so happens, the already-released endo) enables a passable promise to carry the string that @FUDCo needs to support testing. However, its way of carrying a string was terribly obtuse. 

This PR implements part of the plan explained at https://github.com/endojs/endo/pull/1781#issuecomment-1727080735 . Specifically, to have passable promises follow the precedent of passable far objects and exos, which also allow them to override the inherited @@toStringTag property with an own data property carrying a string value. However, agoric-sdk would not be able to use this more pleasant way of carrying a string until it is upgraded to depend on an endo containing this PR.

### DIFF
--- a/packages/marshal/test/test-encode-slot-in-promise.js
+++ b/packages/marshal/test/test-encode-slot-in-promise.js
@@ -1,0 +1,54 @@
+import { test } from './prepare-test-env-ava.js';
+
+// eslint-disable-next-line import/order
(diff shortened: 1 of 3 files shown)


FUDCo (approved):
Yay!

PR #1787, opened 2023-09-21
Endo Pet Daemon Catch-up

kriskowal:
This change consists of various fixes I’ve found while continuing to work in connection invitations. Commits are individually reviewable. This is targeted to merge in the `endo` branch.

### DIFF
--- a/packages/cli/demo/cat.js
+++ b/packages/cli/demo/cat.js
@@ -193,7 +193,6 @@ const followNamesComponent = async ($parent, $end, powers) => {
 
   const $names = new Map();
   for await (const change of makeRefIterator(E(powers).followNames())) {
-    console.log(change);
     if ('add' in change) {
       const name = change.add;
 
--- a/packages/cli/src/endo.js
+++ b/packages/cli/src/endo.js
@@ -7,42 +7,15 @@ import '@endo/eventual-send/shim.js';
 import '@endo/lockdown/commit.js';
 
 import fs from 'fs';
-import path from 'path';
 import url from 'url';
-import os from 'os';
 
 import { Command } from 'commander';
-import { start, stop, restart, clean, reset } from '@endo/daemon';
-import {
-  whereEndoState,
-  whereEndoEphemeralState,
-  whereEndoSock,
-  whereEndoCache,
-} from '@endo/where';
 
 const collect = (value, values) => values.concat([value]);
 
 const packageDescriptorPath = url.fileURLToPath(
   new URL('../package.json', import.meta.url),
 );
-
-const { username, homedir } = os.userInfo();
-const temp = os.tmpdir();
-const info = {
-  user: username,
-  home: homedir,
-  temp,
-};
-
-const statePath = whereEndoState(process.platform, process.env, info);
-const ephemeralStatePath = whereEndoEphemeralState(
-  process.platform,
-  process.env,
-  info,
-);
-const sockPath = whereEndoSock(process.platform, process.env, info);
-const cachePath = whereEndoCache(process.platform, process.env, info);
-const logPath = path.join(statePath, 'endo.log');
 
 export const main = async rawArgs => {
   const program = new Command();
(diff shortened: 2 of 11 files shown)


packages/cli/demo/cat.js
```diff
@@ -193,7 +193,6 @@ const followNamesComponent = async ($parent, $end, powers) => {
 
   const $names = new Map();
   for await (const change of makeRefIterator(E(powers).followNames())) {
-    console.log(change);
```
kriskowal, line 196: oops

packages/cli/src/endo.js
```diff
@@ -241,7 +213,7 @@ export const main = async rawArgs => {
     });
 
   program
-    .command('send <party> <message with embedded references>')
+    .command('send <party> <message-with-embedded-references>')
```
kriskowal, line 216: consistency is the hobgoblin

packages/cli/src/endo.js
```diff
@@ -516,69 +488,79 @@ export const main = async rawArgs => {
     .command('state')
     .description('prints the state directory path')
     .action(async _cmd => {
+      const { statePath } = await import('./locator.js');
```
kriskowal, line 491: With this change, the CLI no longer depends on anything of substance so it starts very quickly. So, fast commands can be fast.

packages/cli/src/message-format.js
```diff
@@ -10,11 +10,11 @@ export const formatMessage = (strings, edgeNames) => {
     index < Math.min(strings.length, edgeNames.length);
     index += 1
   ) {
-    message += strings[0].replace(/@/g, '\\@');
+    message += strings[index].replace(/@/g, '\\@');
     message += `@${edgeNames[index]}`;
   }
   if (strings.length > edgeNames.length) {
-    message += strings[index];
+    message += strings[index].replace(/@/g, '\\@');
```
kriskowal, line 17: Obvious error in hindsight. Tests for the win. This will become more complicated when we support dereferencing dotted paths through name hubs.

packages/cli/src/mkguest.js
```diff
@@ -3,13 +3,7 @@ import os from 'os';
...
 import { withEndoParty } from './context.js';
 
-export const mkguest = async ({
-  cancel,
-  cancelled,
-  sockPath,
-  name,
-  partyNames,
-}) =>
+export const mkguest = async ({ name, partyNames }) =>
```
kriskowal, line 6: Dross from a prior refactor.

packages/daemon/src/daemon-node-powers.js
```diff
@@ -310,6 +310,20 @@ export const makePowers = ({
...
+  const maybeReadFileText = async path =>
+    readFileText(path).catch(error => {
+      if (
+        error.message.startsWith('ENOENT: ') ||
+        error.message.startsWith('EISDIR: ')
+      ) {
+        return undefined;
+      }
+      throw error;
+    });
```
kriskowal, line 325: This has been a useful pattern in other tools. Testing existence with `stat` is an anti-pattern, so I favor attempting to read and handling failure if the file didn’t exist, passing the data forward if the test succeeds. But, classifying the errors is necessary to avoid swallowing file descriptor limit errors.

packages/daemon/src/daemon-node-powers.js
```diff
@@ -310,6 +310,20 @@ export const makePowers = ({
...
   };
 
+  /**
+   * @param {string} path
+   */
+  const maybeReadFileText = async path =>
+    readFileText(path).catch(error => {
+      if (
+        error.message.startsWith('ENOENT: ') ||
+        error.message.startsWith('EISDIR: ')
```
kumavis, line 320: is there a subprop or code instead of inspecting the message?

kumavis (approved):

packages/daemon/src/daemon-node-powers.js
kriskowal: `error.code` should be available but my experiments with it did not pan out.

PR #1790, opened 2023-09-25
fix(netstring): Early return during chunked write

kriskowal:
## Description

@kumavis identified in review a defect in the chunked mode for netstring writer. This change includes tests for the case where the target writer closes while we attempt to write netstring chunks, and provides the fix. Both tests previously failed.

### Security Considerations

Previously, writing to a netstring writer could mistakenly indicate that the stream was still open after the target closed, possibly reducing availability.

### Scaling Considerations

This should not have a material impact on scale, but might plug a memory leak.

No changes.

### Testing Considerations

The behaviors changed are covered by the unit tests.

### Upgrade Considerations

This should not create an obstacle for upgrades.

### DIFF
--- a/packages/netstring/writer.js
+++ b/packages/netstring/writer.js
@@ -1,6 +1,8 @@
 // @ts-check
 /// <reference types="ses"/>
 
+import { makePromiseKit } from '@endo/promise-kit';
+
 const COMMA_BUFFER = new Uint8Array([','.charCodeAt(0)]);
 
 /** @param {number} length */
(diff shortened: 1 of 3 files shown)


kumavis (approved):

packages/netstring/writer.js
```diff
@@ -39,14 +41,28 @@ export const makeNetstringWriter = (output, { chunked = false } = {}) => {
...
+            }
+          });
+        }
+
+        return Promise.race([
+          writerClosedEarly.promise,
+          Promise.all(partsWritten).then(results => ({
+            done: results.some(({ done }) => done),
+            value: undefined,
+          })),
```
mhofman, line 64: It feels like we're racing ourselves here, and the whole "check done again" is to make sure we resolve with `false` if one of the stream has closed?

packages/netstring/writer.js
```diff
@@ -39,14 +41,28 @@ export const makeNetstringWriter = (output, { chunked = false } = {}) => {
...
+        ]);
```
mhofman, line 65: Ok thinking about this more, how about we simply use the promise kit for the result and forego the race altogether?

```suggestion
        Promise.all(partsWritten).then(results => {
          if (results.every(({ done }) => !done)) {
            writerClosedEarly.resolve({ done: false, value: undefined });
          }
        }).catch(writerClosedEarly.reject);
        return writerClosedEarly.promise;
```

(there are different ways to slice this one of course, but the idea is to use the promise kit to race as the racer, since that's effectively what we're doing)

mhofman (commented):
I don't like the structure of the current race. I think we can clarify it a little.

packages/netstring/writer.js
kriskowal: That works for me.

mhofman (approved):

PR #1791, opened 2023-09-25
fix(compartment-mapper): Root alias search

kriskowal:
I’ve so far been unable to reproduce a problem @kumavis found in compartment-mapper with an obvious solution.

### DIFF
--- a/packages/compartment-mapper/src/import-hook.js
+++ b/packages/compartment-mapper/src/import-hook.js
@@ -251,10 +251,8 @@ export const makeImportHookMaker = (
       // Collate candidate locations for the moduleSpecifier,
       // to support Node.js conventions and similar.
       const candidates = [moduleSpecifier];
-      if (moduleSpecifier !== '.') {
-        for (const candidateSuffix of searchSuffixes) {
-          candidates.push(`${moduleSpecifier}${candidateSuffix}`);
-        }
+      for (const candidateSuffix of searchSuffixes) {
+        candidates.push(`${moduleSpecifier}${candidateSuffix}`);
       }
 
       const { maybeRead } = unpackReadPowers(readPowers);


kumavis:
### history
- originally [no special case](https://github.com/endojs/endo/blob/c81ad12d5e3e05466a41514d4907f592e18b3748/packages/endo/src/archive.js#L44) for "." 
- "." special case added here, adding candidate "./index.js"
https://github.com/endojs/endo/commit/7e7a98e39e4ddb06a5f955ad1407e36888247b16#diff-7de3acb49671b7b9010b69fa9b7c2f3f6bb669f65f6c72ac9c8a5044628fffb5R44 in https://github.com/endojs/endo/pull/442
- "." :scream_cat:  **special case incorrectly removed** :scream_cat: and main candidate handled in infer-exports as `externalAlias` https://github.com/endojs/endo/pull/1374/commits/128eb406afd09e9c6c806a97796a0cebd74d5fd0
- for commonjs, "." main candidate also set in infer-exports via `internalAlias` https://github.com/endojs/endo/pull/1374/commits/2fd471d2dcb2b4ea366e5d0e37d0a2292184e0cc

upon review looks like that was my fault, should have removed the condition instead of just the '.' case

packages/compartment-mapper/test/test-main.js
```diff
@@ -227,7 +227,21 @@ scaffold(
...
+  ).toString(),
+  (t, { namespace: { sword } }) => {
+    t.is(typeof sword, 'string');
+    t.is(sword.length, 30);
+  },
+  2,
+);
+
+scaffold(
+  'package root as main via internalAlias (..)',
```
kumavis, line 244: this path (`..`) first gets resolved to package root (`.`) before searching for candidates, so these tests should be equivalent

kumavis (approved):
i approve of the removal of the condition, the test may be redundant

PR #1794, opened 2023-09-28
feat(pass-style): Use endowed passStyleOf if at global symbol

erights:
Implement @gibson042 's suggestion at https://github.com/endojs/endo/pull/1676/files#r1264339663 as agreed in the thread starting at https://github.com/Agoric/agoric-sdk/pull/8051#issuecomment-1638454456

An advantage of using a symbol-named property on the global instead of a string-named property is that it is not aliased to a global variable name.

### DIFF
--- a/packages/pass-style/package.json
+++ b/packages/pass-style/package.json
@@ -19,6 +19,7 @@
   "exports": {
     ".": "./index.js",
     "./tools.js": "./tools.js",
+    "./endow.js": "./endow.js",
     "./package.json": "./package.json"
   },
   "scripts": {
--- a/packages/pass-style/src/passStyleOf.js
+++ b/packages/pass-style/src/passStyleOf.js
@@ -190,10 +190,13 @@ const makePassStyleOf = passStyleHelpers => {
   return harden(passStyleOf);
 };
 
+export const PassStyleOfEndowmentSymbol = Symbol.for('@endo passStyleOf');
+
 /**
- * If there is already a `VataData` global containing a `passStyleOf`,
- * then presumably it was endowed for us by liveslots, so we should use
- * and export that one instead. Other software may have left it for us here,
+ * If there is already a PassStyleOfEndowmentSymbol property on the global,
+ * then presumably it was endowed for us by liveslots with a `passStyleOf`
+ * function, so we should use and export that one instead.
+ * Other software may have left it for us here,
  * but it would require write access to our global, or the ability to
  * provide endowments to our global, both of which seems adequate as a test of
  * whether it is authorized to serve the same role as liveslots.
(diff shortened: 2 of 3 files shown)


packages/pass-style/package.json
```diff
@@ -19,6 +19,7 @@
   "exports": {
     ".": "./index.js",
     "./tools.js": "./tools.js",
+    "./endow.js": "./endow.js",
```
mhofman, line 22: What is the purpose of splitting the symbol in its own external path? From what I gather it doesn't help with any kind of module shaking because all modules are still necessarily imported.

packages/pass-style/src/passStyleOf.js
```diff
@@ -205,9 +208,7 @@ const makePassStyleOf = passStyleHelpers => {
  * @type {PassStyleOf}
  */
 export const passStyleOf =
-  // UNTIL https://github.com/endojs/endo/issues/1514
-  // Prefer: globalThis?.VatData?.passStyleOf ||
-  (globalThis && globalThis.VatData && globalThis.VatData.passStyleOf) ||
+  (globalThis && globalThis[PassStyleOfEndowmentSymbol]) ||
   makePassStyleOf([
```
mhofman, line 212: It would be great if we could somehow avoid entraining all of passStyleOf's code into contract bundles, but that will likely require another approach in the future (I suppose something like an override of the package's import with a module that only looks up the global and fails otherwise)

packages/pass-style/package.json
erights: Only separation of concerns. Clients of the package, in the normal sense, would import the API in `index.js`. Only providers of (some of) the API would import the API is `endow.js`. This does not help security, since it is a registered symbol and therefore forgeable. This does not help space as you say. The intention is only separation of concerns.

packages/pass-style/src/passStyleOf.js
erights: Not the intent of this PR at all. Willing to entertain at a later time, but seems low priority.

packages/pass-style/package.json
mhofman: I'm not convinced of the need. The endower need to create `passStyleOf` in the first place, so I would expect them to import from `index.js`

packages/pass-style/src/passStyleOf.js
mhofman: Oh totally. Sorry that wasn't clear, it was more a random observation for the future.

erights: It's not to keep the endower's view simpler. It is to keep the normal client's view simpler.

packages/pass-style/package.json
erights: It's not to keep the endower's view simpler. It is to keep the normal client's view simpler.

erights: The normal client should be unaware of the existence of this symbol. The normal client should see no change if we switch to a different endowing mechanism.

packages/pass-style/src/passStyleOf.js
erights: Thanks. Resolving.

packages/pass-style/package.json
mhofman: I see. I had missed that index.js was not wholesale re-exporting passStyleOf.js

mhofman (approved):

PR #1797, opened 2023-09-29
refactor(daemon): rework power structure for easier persistence + platform abstraction

kumavis:
trying to abstract over
1) persistence mechanism
2) runtime platform

### DIFF
--- a/packages/daemon/src/daemon-node-powers.js
+++ b/packages/daemon/src/daemon-node-powers.js
@@ -1,14 +1,22 @@
 // @ts-check
-/* global process, setTimeout, clearTimeout */
 /* eslint-disable no-void */
 
 import { makePromiseKit } from '@endo/promise-kit';
 import { makePipe } from '@endo/stream';
 import { makeNodeReader, makeNodeWriter } from '@endo/stream-node';
+import { makeReaderRef } from './reader-ref.js';
+import { makePetStoreMaker } from './pet-store.js';
+import { servePrivatePortHttp } from './serve-private-port-http.js';
+import { servePrivatePath } from './serve-private-path.js';
+
+const { quote: q } = assert;
 
 const textEncoder = new TextEncoder();
 const medialIterationResult = harden({ done: false });
 const finalIterationResult = harden({ done: false });
+
+const zero512 =
+  '0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000';
 
 /**
  * @param {object} modules
@@ -83,6 +91,13 @@ export const makeHttpPowers = ({ http, ws }) => {
     if (connect) {
       const wss = new WebSocketServer({ server });
       wss.on('connection', (socket, req) => {
+        if (req.method === undefined) {
+          throw Error('expected method in request');
+        }
+        if (req.url === undefined) {
+          throw Error('expected url in request');
+        }
+
         const {
           promise: closed,
           resolve: close,
@@ -164,54 +179,13 @@ export const makeHttpPowers = ({ http, ws }) => {
 
 /**
  * @param {object} modules
- * @param {typeof import('crypto')} modules.crypto
  * @param {typeof import('net')} modules.net
- * @param {typeof import('fs')} modules.fs
- * @param {typeof import('path')} modules.path
- * @param {typeof import('child_process')} modules.popen
- * @param {typeof import('url')} modules.url
- * @param {typeof import('ws')} modules.ws
  * @param {typeof import('http')} modules.http
- * @param {Record<string, string | undefined>} modules.env
- * @param {(pid: number) => void} modules.kill
- * @returns {import('./types.js').DaemonicPowers}
+ * @param {typeof import('ws')} modules.ws
+ * @returns {import('./types.js').NetworkPowers}
  */
-export const makePowers = ({
-  crypto,
-  net,
-  fs,
-  path: fspath,
-  popen,
-  url,
-  http,
-  ws,
-  kill,
-  env,
-}) => {
-  /** @param {Error} error */
-  const sinkError = error => {
-    console.error(error);
-  };
-
-  const makeSha512 = () => {
-    const digester = crypto.createHash('sha512');
-    return harden({
-      update: chunk => digester.update(chunk),
-      updateText: chunk => digester.update(textEncoder.encode(chunk)),
-      digestHex: () => digester.digest('hex'),
-    });
-  };
-
-  const randomHex512 = () =>
-    new Promise((resolve, reject) =>
-      crypto.randomBytes(64, (err, bytes) => {
-        if (err) {
-          reject(err);
-        } else {
-          resolve(bytes.toString('hex'));
-        }
-      }),
-    );
+export const makeNetworkPowers = ({ http, ws, net }) => {
+  const { servePortHttp } = makeHttpPowers({ http, ws });
 
   const serveListener = async (listen, cancelled) => {
     const [
@@ -267,18 +241,73 @@ export const makePowers = ({
       cancelled,
     );
 
-  const informParentWhenReady = () => {
-    if (process.send) {
-      process.send({ type: 'ready' });
+  const connectionNumbers = (function* generateNumbers() {
+    let n = 0;
+    for (;;) {
+      yield n;
+      n += 1;
     }
+  })();
+
+  /**
+   * @param {import('@endo/far').FarRef<unknown>} endoBootstrap
+   * @param {string} sockPath
+   * @param {Promise<never>} cancelled
+   * @param {(error: Error) => void} exitWithError
+   * @returns {{ started: () => Promise<void>, stopped: Promise<void> }}
+   */
+  const makePrivatePathService = (
+    endoBootstrap,
+    sockPath,
+    cancelled,
+    exitWithError,
+  ) => {
+    const privatePathService = servePrivatePath(sockPath, endoBootstrap, {
+      servePath,
+      connectionNumbers,
+      cancelled,
+      exitWithError,
+    });
+    return privatePathService;
   };
 
-  const reportErrorToParent = message => {
-    if (process.send) {
-      process.send({ type: 'error', message });
-    }
+  /**
+   * @param {import('@endo/far').FarRef<unknown>} endoBootstrap
+   * @param {number} port
+   * @param {(port: Promise<number>) => void} assignWebletPort
+   * @param {Promise<never>} cancelled
+   * @param {(error: Error) => void} exitWithError
+   * @returns {{ started: () => Promise<void>, stopped: Promise<void> }}
+   */
+  const makePrivateHttpService = (
+    endoBootstrap,
+    port,
+    assignWebletPort,
+    cancelled,
+    exitWithError,
+  ) => {
+    const privateHttpService = servePrivatePortHttp(port, endoBootstrap, {
+      servePortHttp,
+      connectionNumbers,
+      cancelled,
+      exitWithError,
+    });
+
+    assignWebletPort(privateHttpService.started);
+
+    return privateHttpService;
   };
 
+  return harden({
+    servePortHttp,
+    servePort,
+    servePath,
+    makePrivatePathService,
+    makePrivateHttpService,
+  });
+};
+
+export const makeFilePowers = ({ fs, path: fspath }) => {
   /**
    * @param {string} path
    */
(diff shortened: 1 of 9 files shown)


packages/daemon/src/daemon-node-powers.js
```diff
@@ -279,111 +375,208 @@ export const makePowers = ({
     }
   };
 
-  /**
-   * @param {string} path
-   */
-  const makeFileReader = path => {
-    const nodeReadStream = fs.createReadStream(path);
-    return makeNodeReader(nodeReadStream);
+  const makeHashedContentReadeableBlob = (statePath, sha512) => {
```
kumavis, line 378: naming is hard

packages/daemon/src/daemon-node-powers.js
```diff
@@ -279,111 +375,208 @@ export const makePowers = ({
...
   };
 
-  /**
-   * @param {string} path
-   */
-  const makeFileWriter = path => {
-    const nodeWriteStream = fs.createWriteStream(path);
-    return makeNodeWriter(nodeWriteStream);
-  };
+  const makeHashedContentWriter = async (statePath) => {
```
kumavis, line 397: naming is hard

packages/daemon/src/daemon-node-powers.js
```diff
@@ -279,111 +375,208 @@ export const makePowers = ({
...
+    const getSha512Hex = () => {
+      return sha512Kit.promise;
+    }
 
-  /**
-   * @param {string} path
-   */
-  const readFileText = async path => {
-    return fs.promises.readFile(path, 'utf-8');
+    return { writer: hashedFileWriter, getSha512Hex };
```
kumavis, line 447: better interface for getting the hash on completion?

packages/daemon/src/daemon.js
```diff
@@ -733,17 +605,7 @@ export const main = async (powers, locator, pid, cancel, cancelled) => {
...
-  await powers
-    .readFileText(pidPath)
-    .then(pidText => {
-      const oldPid = Number(pidText);
-      powers.kill(oldPid);
-    })
-    .catch(() => {});
-
-  await powers.writeFileText(pidPath, `${pid}\n`);
+  await powers.updateDaemonPid(locator, pid);
```
kumavis, line 608: i originally called this `finalizeInitialization` as a sort of generic hook into the platform. might make more sense than this specific name

packages/daemon/src/types.d.ts
```diff
@@ -267,3 +218,63 @@ export type EndoWebBundle = {
...
+    locator: Locator,
+    cancelled: Promise<never>,
+  ) => Promise<Worker>;
+  fileURLToPath: (url: string) => string;
+  makeHashedContentWriter: (statePath: string) => Promise<{
+    writer: Writer<Uint8Array>;
+    getSha512Hex: () => Promise<string>;
+  }>;
+  makeHashedContentReadeableBlob: (statePath: string, sha512: string) => {
+    stream: () => Promise<FarRef<Reader<Uint8Array>>>;
```
kumavis, line 271: is `FarRef` correct?

packages/daemon/src/daemon-node-powers.js
```diff
@@ -162,6 +166,96 @@ export const makeHttpPowers = ({ http, ws }) => {
...
+  const renamePath = async (source, target) =>
+    fs.promises.rename(source, target);
+
+  const joinPath = (...components) => fspath.join(...components);
+
+  return harden({
+    makeFileReader,
+    makeFileWriter,
+    writeFileText,
+    readFileText,
```
kriskowal, line 381: I suppose it’s worth noting that I chose to frame these methods with `Text` in the name as opposed to only having byte-oriented APIs at the platform level. That was certainly possible, but this gives the platform latitude to take a shortcut, like Node.js. I also did not want to have an overloaded signature for both bytes and strings in the future.

packages/daemon/src/daemon-node-powers.js
kriskowal: It is. How about `makeReadableBlobForContentHash`?

packages/daemon/src/daemon-node-powers.js
kriskowal: Yeah, how about `makeContentHashWriter`? `makeContentAddressWriter`? Maybe step back to `makeContentAddressStore` with `store` and `fetch` instead?

packages/daemon/src/daemon-node-powers.js
kriskowal: Yeah, I’m feeling better about abstracting this as “makeContentSha512Store” with `store(readable) => Promise<hash: string>` and `fetch(hash: string)`

packages/daemon/src/daemon-node-powers.js
```diff
@@ -279,111 +375,208 @@ export const makePowers = ({
...
+    const head = formulaId512.slice(0, 2);
+    const tail = formulaId512.slice(3);
+    const directory = diskPowers.joinPath(
+      statePath,
+      'formulas',
+      formulaType,
+      head,
+    );
+    const file = diskPowers.joinPath(directory, `${tail}.json`);
+    return { directory, file };
```
kriskowal: I like that this moves to a platform concern. That’s the right encapsulation.

packages/daemon/src/daemon-node-powers.js
```diff
@@ -279,111 +375,208 @@ export const makePowers = ({
...
-    fs.promises.rename(source, target);
+  const initializePersistence = async locator => {
+    const statePathP = diskPowers.makePath(locator.statePath);
+    const ephemeralStatePathP = diskPowers.makePath(locator.ephemeralStatePath);
+    const cachePathP = diskPowers.makePath(locator.cachePath);
+    await Promise.all([statePathP, cachePathP, ephemeralStatePathP]);
+  }
 
-  const joinPath = (...components) => fspath.join(...components);
+  const updateDaemonPid = async (locator, pid) => {
```
kriskowal: I like this.

packages/daemon/src/daemon.js
```diff
@@ -93,33 +78,12 @@ const makeEndoBootstrap = (
...
-    // but also because we won't know the name we will use until we've
-    // completed the hash.
-    const digester = powers.makeSha512();
-    const storageId512 = await powers.randomHex512();
-    const temporaryStoragePath = powers.joinPath(
-      storageDirectoryPath,
-      storageId512,
-    );
-    const writer = powers.makeFileWriter(temporaryStoragePath);
+    const { writer, getSha512Hex } = await powers.makeHashedContentWriter(locator.statePath);
```
kriskowal: There should only be one content address store instance for the lifetime of a daemon process.

packages/daemon/src/types.d.ts
kriskowal: I think `Promise<FarRef<T>>` simplifies to `FarRef<T>` and it seems right to me. cc @michaelfig

packages/daemon/src/daemon-node-powers.js
```diff
@@ -517,7 +517,7 @@ export const makePowers = ({
     await Promise.all([statePathP, cachePathP, ephemeralStatePathP]);
   }
 
-  const updateDaemonPid = async (locator, pid) => {
+  const finalizeInitialization = async (locator, pid) => {
```
kriskowal: I like this too.

packages/daemon/src/daemon.js
```diff
@@ -541,7 +541,7 @@ export const main = async (powers, locator, pid, cancel, cancelled) => {
 
   await powers.initializePersistence(locator);
 
-  const requestedWebletPortText = powers.env.ENDO_HTTP_PORT;
+  const requestedWebletPortText = powers.endoHttpPort;
```
kriskowal: I look forward to making the daemon’s “services” platform specific to the extent that is possible. That might mean that some power facet methods like `requestOpen` or `open` for weblets might have to get rejected early (not supported) or initialized late depending (requires user intervention to enable, like selecting an HTTP port).

packages/daemon/src/daemon.js
```diff
@@ -337,15 +337,8 @@ const makeEndoBootstrap = (
...
-      } else if (formulaIdentifier === 'web-page-js') {
-        return makeValueForFormula('web-page-js', zero512, {
-          type: /** @type {'import-unsafe'} */ ('import-unsafe'),
-          worker: `worker-id512:${zero512}`,
-          powers: 'host',
-          importPath: powers.fileURLToPath(
-            new URL('web-page-bundler.js', import.meta.url).href,
-          ),
-        });
+      } else if (powers.webPageFormula && formulaIdentifier === 'web-page-js') {
```
kriskowal: Would prefer to mux on `formulaIdentifier` and reject with an informative error if the power is absent.

packages/daemon/src/pet-store.js
```diff
@@ -10,228 +10,237 @@ const validFormulaPattern =
...
-    if (!validFormulaPattern.test(formulaIdentifier)) {
-      throw new Error(
-        `Invalid formula identifier ${q(formulaIdentifier)} for pet name ${q(
-          petName,
-        )}`,
-      );
-    }
-    return formulaIdentifier;
-  };
+export const makePetStoreMaker = (diskPowers) => {
```
kriskowal: Running `prettier` will change this line.

packages/daemon/src/types.d.ts
```diff
@@ -136,12 +138,13 @@ export interface Topic<
 }
 
 export interface PetStore {
-  get(petName: string): string | undefined;
-  write(petName: string, formulaIdentifier: string): Promise<void>;
+  lookup(petName: string): string | undefined;
```
kriskowal, line 141: Thank you for catching this misalignment.

packages/daemon/src/daemon-node-powers.js
```diff
@@ -625,20 +651,75 @@ export const makePowers = ({
...
+      informParentWhenReady();
+    },
+    error => {
+      reportErrorToParent(error.message);
+      throw error;
+    })
+
+    const servicesStopped = Promise.all(services.map(({ stopped }) => stopped)).then(() => {});
+
+    return { servicesStopped };
```
kriskowal: Wouldn’t hurt to `harden`.

kriskowal (commented):
I am feeling good about the progress in this stack. Please feel free to squash these and assign the review to me when you think it’s ready for review. I’ve made some naming suggestions. Given that the project is far from stone, I’m happy to merge progress as long as it’s in a working state (as defined by the tests and a run-through of the demo (until we have a test for that!)).

kumavis:
should rework crypto API to be compatible with web crypto (needs more async)

kumavis:
need to pull in some commits from `-web`

kumavis:
`filePowers.readDirectory` return type is a node-ism

petstore and control both rely on `filePowers` instead of `persistencePowers`

packages/daemon/src/types.d.ts
michaelfig: If you intend for `stream` to be potentially remote but returned `async`, then you indeed need `Promise<FarRef<T>>`.  If it's potentially remote but synchronously returned, you can get away with `FarRef<T>`.

kumavis:
types are lacking and sometimes wrong frequently

packages/daemon/src/pet-store.js
```diff
@@ -10,228 +12,238 @@ const validFormulaPattern =
...
+        );
+      }
 
-  const follow = async () =>
-    makeIteratorRef(
-      (async function* currentAndSubsequentNames() {
-        const changes = changesTopic.subscribe();
-        for (const name of [...petNames.keys()].sort()) {
-          yield { add: name };
+      const fromPath = filePowers.joinPath(petNameDirectoryPath, fromName);
```
kriskowal, line 163: I look forward to abstracting file powers even more thoroughly. This might mean that the daemon powers consist of a content store, and a “tables” for pet names, mailboxes, and formulas.

packages/daemon/src/types.d.ts
```diff
@@ -267,3 +223,94 @@ export type EndoWebBundle = {
...
+  makePrivateHttpService: (
+    endoBootstrap: FarRef<unknown>,
+    port: number,
+    assignWebletPort: (portP: Promise<number>) => void,
+    cancelled: Promise<never>,
+    exitWithError: (error: Error) => void,
+  ) => { started: () => Promise<void>; stopped: Promise<void> };
+};
+
+// The return type here is almost an EndoReadable, but not quite. Should fix.
```
kriskowal, line 283: 👀

kriskowal (approved):

Issue #1798, opened 2023-09-30
v8 error taming should leverage `prepareStackTrace`


mhofman:
## What is the Problem Being Solved?

`Error.prepareStackTrace` can be [used by Node.js to apply source maps](https://nodejs.org/docs/latest-v16.x/api/cli.html#--enable-source-maps) to the captured stack trace. However SES sets its own `prepareStackTrace` during v8 Error taming. Instead of completely discarding it. Maybe instead it should leverage as much as possible the existing `prepareStackTrace` to prepare its own stack strings?

Furthermore, if the start compartment set a `prepareStackTrace` after lockdown, it seems the returned value will be used for the `.stack` property instead of the empty string, even if `errorTaming` is set to `'safe'`: https://github.com/endojs/endo/blob/2a85f8aa6684b4d4a92fea48ccc63d01ccddc85e/packages/ses/src/error/tame-v8-error-constructor.js#L301-L304

## Description of the Design

Use the original `prepareStackTrace` to generate the stack string.

Alternatively implement source map support in endo itself (maybe once part of the error handling logic has been ejected from SES as a trusted shim?, see #945). While [node's `prepareStackTrace` implementation is not very modular](https://github.com/nodejs/node/blob/a86c2caea9c329442bbbd8eeec6d5385b594b7ca/lib/internal/source_map/prepare_stack_trace.js#L73-L102), it should be possible to rebuild it on top of [`findSourceMap`](https://nodejs.org/api/module.html#modulefindsourcemappath).

## Security Considerations

None particularly. The start compartment is already in position to decide how lockdown should tame errors, and can also set `prepareStackTrace` after lockdown

## Scaling Considerations

None

## Test Plan

load some minified source which throws errors in Node.js with `--use-source-maps` and verify stack traces are mapped.

## Upgrade Considerations

None particular

erights:
Hi @mhofman , I've co-assigned myself to this to.

mhofman:
> Use the original `prepareStackTrace` to generate the stack string.

I believe I explored this, and that didn't work

PR #1799, opened 2023-09-30
fix(ses): use prepareStackTrace for error stack string

mhofman:
closes: #1798

## Description

Fixes `Error.prepareStackTrace` with the following behavior changes:
- If it generates a stack trace string, use it instead of generating our own
- do not leak the stack string if `errorTaming` is not `'unsafe'` 
  - but keep the workaround for `depd` and similar that return the callSites array
- Restore a made safe `Error.prepareStackTrace` that existed before SES loaded (e.g. Node.js uses it to enable source maps)

### Security Considerations

This change defers to a `prepareStackTrace` set in the start compartment to generate SES stack strings. The assumption is that such overrides are vetted.

### Scaling Considerations

The new implementation uses `try`/`catch` which is often more expensive, but it likely dwarfs in comparison to the call sites wrapping we are already doing.

How should this change be documented? I somewhat expected this to be the behavior in the first place

### Testing Considerations

Passes existing unit test.

### Upgrade Considerations

None

### DIFF
--- a/packages/ses/src/error/tame-v8-error-constructor.js
+++ b/packages/ses/src/error/tame-v8-error-constructor.js
@@ -8,10 +8,12 @@ import {
   arraySlice,
   create,
   defineProperties,
+  errorToString,
   fromEntries,
   reflectSet,
   regexpExec,
   regexpTest,
+  stringReplace,
   weakmapGet,
   weakmapSet,
   weaksetAdd,
@@ -260,7 +262,7 @@ export const tameV8ErrorConstructor = (
       if (errorTaming === 'unsafe') {
         const stackString = stackStringFromSST(error, sst);
         weakmapSet(stackInfos, error, { stackString });
-        return `${error}${stackString}`;
+        return `${errorToString(error)}${stackString}`;
       } else {
         weakmapSet(stackInfos, error, { callSites: sst });
         return '';
@@ -300,9 +302,10 @@ export const tameV8ErrorConstructor = (
     const systemMethods = {
       prepareStackTrace(error, sst) {
         const stackInfo = {};
+        const safeCallSites = safeV8SST(sst);
+        let preparedStack;
         try {
-          const safeCallSites = safeV8SST(sst);
-          const preparedStack = inputPrepareFn(error, safeCallSites);
+          preparedStack = inputPrepareFn(error, safeCallSites);
           if (preparedStack === safeCallSites) {
             // Handle depd and similar prepareStackTrace which return the structured stack trace
             stackInfo.callSites = sst;
@@ -311,7 +314,7 @@ export const tameV8ErrorConstructor = (
             typeof preparedStack === 'string' &&
             preparedStack !== ''
           ) {
-            stackInfo.stackString = preparedStack;
+            stackInfo.stackString = stringReplace(preparedStack, /^[\n]+/, '');
           } else {
             // We just need to get to the `catch` clause
             // eslint-disable-next-line no-throw-literal
@@ -320,6 +323,7 @@ export const tameV8ErrorConstructor = (
         } catch (_err) {
           if (errorTaming === 'unsafe') {
             stackInfo.stackString = stackStringFromSST(error, sst);
+            preparedStack = `${errorToString(error)}${stackInfo.stackString}`;
           } else {
             stackInfo.callSites = sst;
           }
@@ -327,7 +331,7 @@ export const tameV8ErrorConstructor = (
           weakmapSet(stackInfos, error, stackInfo);
         }
 
-        return errorTaming === 'unsafe' ? stackInfo.stackString : '';
+        return errorTaming === 'unsafe' ? preparedStack : '';
       },
     };
     weaksetAdd(systemPrepareFnSet, systemMethods.prepareStackTrace);
--- a/packages/ses/src/error/tame-v8-error-constructor.js
+++ b/packages/ses/src/error/tame-v8-error-constructor.js
@@ -285,7 +285,7 @@ export const tameV8ErrorConstructor = (
 
   const defaultPrepareFn = tamedMethods.prepareStackTrace;
 
-  OriginalError.prepareStackTrace = defaultPrepareFn;
+  const originalPrepareStackTrace = OriginalError.prepareStackTrace;
 
   // A weakset branding some functions as system prepareFns, all of which
   // must be defined by this module, since they can receive an
@@ -299,8 +299,35 @@ export const tameV8ErrorConstructor = (
     // Use concise methods to obtain named functions without constructors.
     const systemMethods = {
       prepareStackTrace(error, sst) {
-        weakmapSet(stackInfos, error, { callSites: sst });
-        return inputPrepareFn(error, safeV8SST(sst));
+        const stackInfo = {};
+        try {
+          const safeCallSites = safeV8SST(sst);
+          const preparedStack = inputPrepareFn(error, safeCallSites);
+          if (preparedStack === safeCallSites) {
+            // Handle depd and similar prepareStackTrace which return the structured stack trace
+            stackInfo.callSites = sst;
+            return safeCallSites;
+          } else if (
+            typeof preparedStack === 'string' &&
+            preparedStack !== ''
+          ) {
+            stackInfo.stackString = preparedStack;
+          } else {
+            // We just need to get to the `catch` clause
+            // eslint-disable-next-line no-throw-literal
+            throw undefined;
+          }
+        } catch (_err) {
+          if (errorTaming === 'unsafe') {
+            stackInfo.stackString = stackStringFromSST(error, sst);
+          } else {
+            stackInfo.callSites = sst;
+          }
+        } finally {
+          weakmapSet(stackInfos, error, stackInfo);
+        }
+
+        return errorTaming === 'unsafe' ? stackInfo.stackString : '';
       },
     };
     weaksetAdd(systemPrepareFnSet, systemMethods.prepareStackTrace);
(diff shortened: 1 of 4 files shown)


mhofman:
For reference, I think v8 does its own changes to `Error` for `prepareStackTrace`:
- https://github.com/nodejs/node/blob/v20.8.0/lib/internal/errors.js
- https://github.com/nodejs/node/blob/v20.8.0/lib/internal/bootstrap/realm.js#L435-L458

erights:
> For reference, I think v8 does its own changes to Error for prepareStackTrace:

Did you mean Node?

packages/ses/src/error/tame-v8-error-constructor.js
```diff
@@ -285,7 +287,7 @@ export const tameV8ErrorConstructor = (
 
   const defaultPrepareFn = tamedMethods.prepareStackTrace;
 
-  OriginalError.prepareStackTrace = defaultPrepareFn;
+  const originalPrepareStackTrace = OriginalError.prepareStackTrace;
```
erights, line 290: The old code attenuated the `OriginalError.prepareStackTrace`. The new code does not. But `OriginalError` *should* be encapsulated within this taming. The powerful dangerous error exposed on the start compartment is `InitialError`, whose taming should equivalently attenuate its `InitialError.prepareStackTrace` access property. Yes?

Additionally, the old code made no use of what happened to already be the value of `OriginalError.prepareStackTrace` before SES started. Instead, the taming implicitly assumed either that it would be as v8 left it. If the host had modified it before SES started (as it turns out Node does), that we would ignore that replacement behavior. Yes?

This PR captures this original `prepareStackTrace` function, whatever it is, and then initializes the `InitialError.prepareStackTrace` accessor property by reusing the attenuation it already implements in its setter. Yes?

packages/ses/src/error/tame-v8-error-constructor.js
```diff
@@ -333,5 +364,7 @@ export const tameV8ErrorConstructor = (
     },
```
erights, line 364: The getter above on line 351 returns `OriginalError.prepareStackTrace`. This initially looked less safe to me, because you're not attenuating it above. But actually, the original `prepareStackTrace` remains encapsulated within this taming, and cannot escape, because the assignment on line 367 calls the setter, which replaces it with an attenuation. Yes?

packages/ses/src/error/tame-v8-error-constructor.js
```diff
@@ -299,8 +301,37 @@ export const tameV8ErrorConstructor = (
...
+          preparedStack = inputPrepareFn(error, safeCallSites);
+          if (preparedStack === safeCallSites) {
+            // Handle depd and similar prepareStackTrace which return the structured stack trace
+            stackInfo.callSites = sst;
+            return safeCallSites;
+          } else if (
+            typeof preparedStack === 'string' &&
+            preparedStack !== ''
+          ) {
+            stackInfo.stackString = stringReplace(preparedStack, /^[\n]+/, '');
```
erights, line 317: This stringReplace does not seem to correspond to anything in the original. Is it an observable behavior change? What is its purpose?

It seems to only remove an initial run of newlines and not do anything else. If this is a behavioral change, it is a good one! When does this happen? (Seem to ring a bell, but I don't remember)

packages/ses/src/error/tame-v8-error-constructor.js
```diff
@@ -299,8 +301,37 @@ export const tameV8ErrorConstructor = (
...
+            // We just need to get to the `catch` clause
+            // eslint-disable-next-line no-throw-literal
+            throw undefined;
+          }
+        } catch (_err) {
+          if (errorTaming === 'unsafe') {
+            stackInfo.stackString = stackStringFromSST(error, sst);
+            preparedStack = `${errorToString(error)}${stackInfo.stackString}`;
+          } else {
+            stackInfo.callSites = sst;
```
erights, line 328: I had to read very carefully before I was satisfied that the original `sst` still cannot leak. That invariant seems less local now. Worth a comment?

packages/ses/test/test-v8-callsite-properties.js
```diff
@@ -8,10 +8,14 @@ test('callSite properties', t => {
...
-      const pst = (_err, sst0) => sst0;
+      const pst = (_err, sst0) => {
+        sst = sst0;
+        return '';
+      };
       Error.prepareStackTrace = pst;
       const e = Error();
-      sst = e.stack;
+      // eslint-disable-next-line no-void
+      void e.stack;
```
erights, line 18: Please help me understand this change and how it relates to this PR.

packages/ses/src/error/tame-v8-error-constructor.js
```diff
@@ -260,7 +262,7 @@ export const tameV8ErrorConstructor = (
       if (errorTaming === 'unsafe') {
         const stackString = stackStringFromSST(error, sst);
         weakmapSet(stackInfos, error, { stackString });
-        return `${error}${stackString}`;
+        return `${errorToString(error)}${stackString}`;
```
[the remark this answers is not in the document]
mhofman: Not really more safe because it still defers to the `name` and `message` properties, but more aligned to what I understand is the behavior of the `.stack` property on Node and v8 which ignore an overridden `toString()`.

packages/ses/src/error/tame-v8-error-constructor.js
mhofman: The new code does attenuate the original `OriginalError.prepareStack` when it uses it with the `InitialError.prepareStack` setter below. I can add a comment to this effect to make this more clear.

Indeed the original code completely discarded the original `prepareStack`. It's a stated goal of this PR to no longer do so as the host, in particular Node.js, may have set it up to something that helps handle stack traces.

packages/ses/src/error/tame-v8-error-constructor.js
mhofman: a `stack` property is of the form `ErrorName: error message(\n  at location:12:3)+`.
`prepareStack` is supposed to return a full `stack` property.

`stackString` seems to be only the formatted location lines, missing the error name and message. This reconciliates the 2 formats.

This behavior wasn't in the original code because the taming never used the return value of the user specified `prepareStack` to compute its own stack info.

packages/ses/src/error/tame-v8-error-constructor.js
mhofman: Correct. I'll add a comment to clarify.

packages/ses/test/error/test-tame-v8-error-unsafe.js
```diff
@@ -94,7 +94,7 @@ test('SES compartment error compatibility - endow w Error power', t => {
   const c1 = new Compartment({ t, Error });
   const result = c1.evaluate(`
     const obj = {
-      toString: () => 'Pseudo Error',
+      name: 'Pseudo Error',
     };
```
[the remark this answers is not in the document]
mhofman: As I mentioned, not a safety change, but more of a fidelity change. I could add a `message` and keep the overridden `toString` to show the expected behavior.

packages/ses/test/test-v8-callsite-properties.js
mhofman: This test was abusing the `depd` hack of `prepareStack` returning the sst directly. I didn't want to perpetrate it, so I refactored the test to instead capture the sst in another way.

packages/ses/src/error/tame-v8-error-constructor.js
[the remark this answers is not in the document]
erights: Nice. Resolving.

packages/ses/src/error/tame-v8-error-constructor.js
erights: > I can add a comment to this effect to make this more clear.

Yes please. Resolving.

packages/ses/src/error/tame-v8-error-constructor.js
erights: I see the 👍 . Resolving.

packages/ses/src/error/tame-v8-error-constructor.js
erights: Cool. Resolving.

packages/ses/test/error/test-tame-v8-error-unsafe.js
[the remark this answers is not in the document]
erights: No need. For this, I'd prefer just a comment anyway. Resolving.

packages/ses/test/test-v8-callsite-properties.js
erights: Oh! Resolving.

erights (approved):
With the expected added commentary, LGTM, thanks!

If it makes more sense to explain in a separate docs/*.md file, feel free to do that instead.

kriskowal:
@mhofman You have a stamp and a request from @turadg to include this in the next release.

mhofman:
This does not do what it's supposed to do, back to draft

erights:
Hi @mhofman , after this had several approvals, you wrote:

> This does not do what it's supposed to do, back to draft

Do you remember what the problems are?

As a draft this PR was not assigned. I went ahead and assigned myself, since I want to look at reviving it. I also left myself on as a reviewer FWIW.

mhofman:
> Do you remember what the problems are?

Yeah it didn't solve the problem. I think because Node.js verified that the `prepareStackTrace` function was not modified from its own.

turadg:
I think https://github.com/endojs/endo/pull/2355/ obviated this

PR #1801, opened 2023-10-01
docs(marshal): Smallcaps Cheatsheet

erights:
As promised at the last ocapn meeting, here is a compact summary of the Smallcaps encoding of passable JS values into JSON.

### DIFF
--- a/packages/marshal/docs/smallcaps-cheatsheet.md
+++ b/packages/marshal/docs/smallcaps-cheatsheet.md
@@ -1,0 +1,32 @@
+# Smallcaps Cheatsheet
+
+An example-based summary of the Smallcaps encoding
+
+| Passable value  | OCapn name    | JS example          | JSON encoding        |
+| ----------------|---------------|---------------------|----------------------|
+| bigint          | SignedInteger | `7n`<br>`-7n`       | `"+7"`<br>`"-7"`     |
+| manifest constant | Undefined<br><br>Float64<br><br><br> | `undefined`<br>`Infinity`<br>`-Infinity`<br>`NaN`<br>`-0` | `"#undefined"`<br>`"#Infinity"`<br>`"#-Infinity"`<br>`"#NaN"`<br>`"#-0"` // unimplemented |
+| passable symbol | Symbol        | `Symbol.for('foo')`<br>`Symbol.asyncIterator` | `"%foo"`<br>`"%@@asyncIterator"` |
+| remotable       | Capability    | `Far('foo', {})`    | `"$0.foo"`           |
+| promise         | Capability    | `Promise.resolve()` | `"&1"`               |
+| special string  | String        | `'#foo'`            | `"!#foo"`            |
+| other string    | String        | `'foo'`             | `"foo"`              |
+| other JSON scalar | Null<br>Boolean<br><br>Float64 | `null`<br>`true`<br>`false`<br>`7.1` | `null`<br>`true`<br>`false`<br>`7.1` |
+| copyArray       | Sequence      | `[a,b]`             | `[<a>,<b>]`          |
+| copyRecord      | Struct        | `{x:a,y:b}`         | `{<x>:<a>,<y>:<b>}`  |
+| error           | Error         | `TypeError(msg)`    | `{"#error":<msg>,"name":"TypeError"}` |
+| tagged          | Tagged        | `makeTagged(t,p)`   | `{"#tag":<t>,"payload":<p>}` |
+| ? | ByteString | ? | ? |
+
+* The `-0` encoding is defined as above, but not yet implemented in JS.
+* In JS, only registered and well-known symbols are passable.
+* The number after `"$"` or `"&"` is an index into a separate slots array.
+* Special strings begin with any of the `!"#$%&'()*+,-` characters.
+* `<expr>` is nested encoding of `expr`.
+* To be passable, arrays, records, and errors must also be hardened.
+* Structs [can only have string-named properties](https://github.com/endojs/endo/blob/master/packages/pass-style/doc/copyRecord-guarantees.md).
+* Errors can also carry an optional `errorId` string property.
+* We expect to expand the optional error properties over time.
+* The ByteString encoding is not yet designed or implemented.
+
+Every JSON encoding with no special strings anywhere decodes to itself.


erights:
@cwebber @tsyesika Although I cannot officially add you as reviewers, I look forward to your comments. Thanks!

packages/marshal/docs/smallcaps-cheatsheet.md
```diff
@@ -0,0 +1,29 @@
...
+| passable symbol | Symbol        | `Symbol.for('foo')`<br>`Symbol.asyncIterator` | `"%foo"`<br>`"%@@asyncIterator"` |
+| remotable       | Capability    | `Far('foo', {})`    | `"$0.foo"`           |
+| promise         | Capability    | `Promise.resolve()` | `"&1"`               |
+| special string  | String        | `'#foo'`            | `"!#foo"`            |
+| other string    | String        | `'foo'`             | `"foo"`              |
+| other JSON scalar | Null<br>Boolean<br><br>Float64 | `null`<br>`true`<br>`false`<br>`7.1` | `null`<br>`true`<br>`false`<br>`7.1` |
+| copyArray       | Sequence      | `[a,b]`             | `[<a>,<b>]`          |
+| copyRecord      | Struct        | `{x:a,y:b}`         | `{<x>:<a>,<y>:<b>}`  |
+| error           | Error         | `TypeError(msg)`    | `{"#error":<msg>,"name":"TypeError"}` |
+| tagged          | Tagged        | `makeTagged(t,p)`   | `{"#tag":<t>,"payload":<p>}` |
```
kriskowal, line 18: I recommend also showing the JS example for a tagged that does not rely on the `makeTagged` utility, since this illustrates why tagged values do not overlap with the representations of any other value, and clarifies why we can’t support structs with symbol keys.

packages/marshal/docs/smallcaps-cheatsheet.md
```diff
@@ -0,0 +1,29 @@
...
+| manifest constant | Undefined<br><br>Float64<br><br><br> | `undefined`<br>`Infinity`<br>`-Infinity`<br>`NaN`<br>`-0` | `"#undefined"`<br>`"#Infinity"`<br>`"#-Infinity"`<br>`"#NaN"`<br>`"#-0"` // unimplemented |
+| passable symbol | Symbol        | `Symbol.for('foo')`<br>`Symbol.asyncIterator` | `"%foo"`<br>`"%@@asyncIterator"` |
+| remotable       | Capability    | `Far('foo', {})`    | `"$0.foo"`           |
+| promise         | Capability    | `Promise.resolve()` | `"&1"`               |
+| special string  | String        | `'#foo'`            | `"!#foo"`            |
+| other string    | String        | `'foo'`             | `"foo"`              |
+| other JSON scalar | Null<br>Boolean<br><br>Float64 | `null`<br>`true`<br>`false`<br>`7.1` | `null`<br>`true`<br>`false`<br>`7.1` |
+| copyArray       | Sequence      | `[a,b]`             | `[<a>,<b>]`          |
+| copyRecord      | Struct        | `{x:a,y:b}`         | `{<x>:<a>,<y>:<b>}`  |
+| error           | Error         | `TypeError(msg)`    | `{"#error":<msg>,"name":"TypeError"}` |
```
kriskowal, line 17: It might be illustrative to make the message concrete and show how or whether it gets escaped.

packages/marshal/docs/smallcaps-cheatsheet.md
```diff
@@ -0,0 +1,29 @@
+# Smallcaps Cheatsheet
+
+An example-based summary of the Smallcaps encoding
+
+| Passable value  | OCapn name    | JS example          | JSON encoding        |
+| ----------------|---------------|---------------------|----------------------|
+| bigint          | SignedInteger | `7n`<br>`-7n`       | `"+7"`<br>`"-7"`     |
+| manifest constant | Undefined<br><br>Float64<br><br><br> | `undefined`<br>`Infinity`<br>`-Infinity`<br>`NaN`<br>`-0` | `"#undefined"`<br>`"#Infinity"`<br>`"#-Infinity"`<br>`"#NaN"`<br>`"#-0"` // unimplemented |
+| passable symbol | Symbol        | `Symbol.for('foo')`<br>`Symbol.asyncIterator` | `"%foo"`<br>`"%@@asyncIterator"` |
```
kriskowal, line 9: It might be good to also illustrate how registered symbols get escaped so they do not overlap well-known symbols.

packages/marshal/docs/smallcaps-cheatsheet.md
kriskowal: I assume our handling of other properties is tentative, but it might be useful to illustrate how we handle a collision between the struct field name `#error` and the special field names `#error` or `#tag`.

kriskowal (approved):

packages/marshal/docs/smallcaps-cheatsheet.md
erights: Based on feedback from the ocapn meeting, and also a private conversation with @cwebber , I'm trying to keep this as small as possible while still conveying enough info that they can get started. (That's why I called it only a "Cheatsheet".)

@cwebber , did I understand what you need correctly? For your (and ocapn's) purposes, would you rather see these details added or omitted? It would certainly be little trouble to add. Likewise for @kriskowal 's suggestions below.

packages/marshal/docs/smallcaps-cheatsheet.md
erights: Awaiting feedback from @cwebber , but even for JS this seems too detailed for a cheetsheet. OTOH, the whole thing is so close to being correct and complete that it might be worth the expansion, and then relabeling the document from "cheetsheet" to something else.

packages/marshal/docs/smallcaps-cheatsheet.md
erights: > It might be illustrative to make the message concrete and show how or whether it gets escaped.

For this doc, my inclination is to let the angle brackets do the work.

> I assume our handling of other properties is tentative, 

Added two bullets to clarify that we expect errors to grow more optional properties over time.

> but it might be useful to illustrate how we handle a collision between the struct field name #error and the special field names #error or #tag.

Again, I think the angle brackets already do all that work.

packages/marshal/docs/smallcaps-cheatsheet.md
erights: > clarifies why we can’t support structs with symbol keys.

Added a bullet point stating that Structs can only have string keys.

erights: @kriskowal @cwebber , with the added bullet points, I feel close enough to @kriskowal 's suggestions to make use of his approval to merge this, which I will proceed to do.

@kriskowal,  @cwebber, other reviewers, I'd still like feedback. It is simple enough to do post-merge adjustments on this doc.

dckc (commented):
seems reasonable to me

PR #1803, opened 2023-10-02
fix: enable compatibility with node16/nodenext module resolution

### LINKED ISSUE #1802: types.d.ts inaccessible when using recommended module resolution algorithm
## Describe the bug

A consumer using [`module` option](https://www.typescriptlang.org/tsconfig#module) `node16/nodenext` and/or or [`moduleResolution` option](https://www.typescriptlang.org/tsconfig#moduleResolution) `node16/nodenext` cannot access `types.d.ts` from the `types` field of any package which provides it.

### Case A

This won't be obvious unless `types.d.ts` exports something that the main entry point `index.js` doesn't.  Assuming that's the case:

1. Create a package with `module: node16` in its `tsconfig.json`
2. Import a type from a package w/ a `types` field where the type only exists in `types.d.ts` but not `index.js`
3. Observe compilation error

### Case B

Otherwise, it's more subtle:

1. Create a package with `module: node16` in its `tsconfig.json`
2. Import a type from a package with a `types` field
3. Use your editor to navigate to the reference of the type.
4. You will navigate to `index.js`.

## Expected behavior

Case A: no error
Case B: should have navigated to `types.d.ts`

## Platform environment

n/a

## Additional context

This is not a problem when using the legacy `node` module resolution algorithm.

The solution is to add a `types` conditional export, as you will read in the documentation linked above.

boneskull:
Closes #1802 

## Description

TypeScript's `node16` and `nodenext` module resolution algorithms attempt to adhere more closely to the behavior of the `exports` field.  To that end, if an `exports` field is present in a package, the `types` field is ignored; instead, a `types` conditional export is expected.

Without this change, any consumer using one of these newer--_officially recommended_--module resolution algorithms will be unable to load any `types.d.ts` file from any package.  We retain the `types` field because of backwards compatibility with the legacy `node` module resolution algorithm.

[Read more about how I came to this conclusion](https://gist.github.com/boneskull/ae1faac253379a7d6f24749c9ba12c24)

### Should the `types` field be used at all?

_It depends._  While this file is not _truly_ needed for type support if type declarations are shipped alongside each `.js` source file, it's awkward to export any types created via `@typedef` without it.

For instance, say we have a function `foo(opts: FooOpts)`.  `foo` lives in `a.js`, and `FooOpts` is a `@typedef` in `b.js`. The parameter `opts` is tagged as `@param {import('./b.js').FooOpts} opts`, as it should.  Entry point `index.js` re-exports `foo()` from `a.js`.  However, the definition for `FooOpts` _will not be available to consumers_--which will cause compilation errors.  We must, then, re-export this type from the entry point.  To avoid this, we have precious few tools at our disposal:

1. Create a reference of `FooOpts` in `index.js` via `@typedef`, which will export the `FooOpts` type.  If `FooOpts` is generic, this becomes more of a headache, as we have to _copy_ the generics from the original.  This is a hazard, because the generics could become out-of-sync.
2. Create a `.d.ts` (it could also be a `.ts`, which we can assume will generate the associated `.d.ts`; this doesn't matter for our purposes) which re-exports `FooOpts` from `b.js`.  This file will be the titular `types.d.ts`.

The question we need to ask, then, is "do we need to export any `@typedef`s?" If the answer is _yes_, then a `types.d.ts` makes sense.

I didn't actually check to see if `types.d.ts` is appropriate in all of these packages; I just fixed the problem I was encountering.  For `@endo/compartment-mapper`, some work I'm doing will mean re-exporting some `@typedef`s, even if that's not happening currently.  Furthermore, I ignored the packages with `"types": null`, since I don't understand what that's supposed to mean.

### Testing Considerations

I did not add any tests, as the test needs to happen on the result of `npm pack`, which is more of a fixture than I wanted to build out.  However, I am working on a [tool](https://github.com/boneskull/midnight-smoker) which will (hopefully soon) automatically check for problems like this.  Once that's ready, I can open a new PR to add the checks and you can see what you think.

### DIFF
--- a/packages/compartment-mapper/package.json
+++ b/packages/compartment-mapper/package.json
@@ -22,7 +22,10 @@
   "main": "./index.js",
   "types": "./types.d.ts",
   "exports": {
-    ".": "./index.js",
+    ".": {
+      "types": "./types.d.ts",
+      "default": "./index.js"
+    },
     "./import.js": "./import.js",
     "./archive.js": "./archive.js",
     "./import-archive.js": "./import-archive.js",
--- a/packages/daemon/package.json
+++ b/packages/daemon/package.json
@@ -21,7 +21,10 @@
   "module": "./index.js",
   "types": "./types.d.ts",
   "exports": {
-    ".": "./index.js",
+    ".": {
+      "types": "./types.d.ts",
+      "default": "./index.js"
+    },
     "./package.json": "./package.json"
   },
   "scripts": {
--- a/packages/lp32/package.json
+++ b/packages/lp32/package.json
@@ -29,7 +29,10 @@
   "module": "./index.js",
   "types": "./types.d.ts",
   "exports": {
-    ".": "./index.js",
+    ".": {
+      "types": "./types.d.ts",
+      "default": "./index.js"
+    },
     "./reader.js": "./reader.js",
     "./writer.js": "./writer.js",
     "./package.json": "./package.json"
--- a/packages/ses/package.json
+++ b/packages/ses/package.json
@@ -35,11 +35,13 @@
   "types": "./types.d.ts",
   "exports": {
     ".": {
+      "types": "./types.d.ts",
       "import": "./index.js",
       "require": "./dist/ses.cjs",
       "types": "./types.d.ts"
     },
     "./lockdown": {
+      "types": "./types.d.ts",
       "import": "./index.js",
       "require": "./dist/ses.cjs",
       "types": "./types.d.ts"
--- a/packages/stream/package.json
+++ b/packages/stream/package.json
@@ -25,7 +25,10 @@
   "module": "./index.js",
   "types": "./types.d.ts",
   "exports": {
-    ".": "./index.js",
+    ".": {
+      "types": "./types.d.ts",
+      "default": "./index.js"
+    },
     "./package.json": "./package.json"
   },
   "scripts": {
--- a/packages/where/package.json
+++ b/packages/where/package.json
@@ -18,7 +18,10 @@
   "module": "./index.js",
   "types": "./types.d.ts",
   "exports": {
-    ".": "./index.js",
+    ".": {
+      "types": "./types.d.ts",
+      "default": "./index.js"
+    },
     "./package.json": "./package.json"
   },
   "scripts": {


mhofman (commented):
As a stop gap measure, this seems fine.

> Should the types field be used at all?

I am still of the opinion that the answer is no, and that we should migrate away from these manual type definition files. We don't have to use JSDoc `@typedefs` to migrate away, just need to "runtime" re-export the types file from the entrypoint, but that does require generating type definition files in a prepublish build step, which these packages do not do yet.

boneskull:
> We don't have to use JSDoc @typedefs to migrate away, just need to "runtime" re-export the types file from the entrypoint, but that does require generating type definition files in a prepublish build step, which these packages do not do yet.

I don't follow.  Say we wanted to expose `PackageNamingKit` from `types.js` in `@endo/component-mapper`.  How would we do that, exactly?

mhofman:
> @mhofman I don't follow. Say we wanted to expose `PackageNamingKit` from `types.js` in `@endo/component-mapper`. How would we do that, exactly?

- Remove the current `types.d.ts` which seems to only be a "types" copy of runtime exports present in `index.js`.
- Create an `exports.js` with only `export {};` as content
- Create an `exports.d.ts`, with the following as content: `export { PackageNamingKit } from './src/types.js';`
- Add `export * from './exports.js';` in `index.js`
- Remove any `"types"` field from `package.json`
- Add a `prepublish` step to generate `.d.ts` files from `.js` files like some other packages are already doing

boneskull:
@mhofman  I'll try that.

But it begs the question: why use a `types.js` anyway, when we could just rewrite it as `types.ts` using the convenient syntax?  To me, a `.js` file containing only types seems like it should be a `.ts` file.  Is there a practical/technical concern?  Or a preference / philosophical one?

boneskull:
(btw, `compartment-mapper` already compiles declarations in its `prepack` script then blasts them in its `postpack` script)

boneskull:
@mhofman By "I'll try that", I mean I will try that in another forthcoming PR.  Care to approve?

mhofman:
> In this case, `exports.js` and `export.d.ts` would just become `exports.ts`, and the `.d.ts` would be generated in the `prepack` lifecycle script, then removed after.

Because that requires the generation of the output not just for publish but also for authoring inside the repo, which I am against.

> TS finds no types whatsoever unless the `.d.ts` files have been generated. e.g. If I _link_ `compartment-mapper` into my own project, I have to run its `prepack` script to have type information. I don't think this is unusual behavior--though it is awkward for my use case.

That is exactly the behavior I find unacceptable. There should be no generation step needed when working within the repo (which effectively also uses links through yarn workspace)

boneskull:
@mhofman 

> Because that requires the generation of the output not just for publish but also for authoring inside the repo, which I am against.

It doesn't.

> That is exactly the behavior I find unacceptable. There should be no generation step needed when working within the repo (which effectively also uses links through yarn workspace)

There isn't!

It's only when working _outside_ of the repo.  By my own project I mean "a package somewhere in the lavamoat monorepo which links `@endo/compartment-mapper`.

I'm confident this all works the way you think it should.

mhofman:
if there is an `export * from './src/export.js'` in `index.js` that means an `export.js` must exist or else the package will fail to load at runtime. If that file is generated, it means a generation step is required. Note this is not a concern to make types work, but a concern to keep runtime imports of the package working when the package is imported by other packages in the same repo.

boneskull:
@mhofman Ah, you're right.  Anyway: the code in question doesn't exist.  I am trying to get _this_ PR merged, though.  ;) 

cc @naugtur @kriskowal

naugtur (approved):
improvement. no risks. let's merge

PR #1804, opened 2023-10-04
feat(marshal): encode capData in 1 level of JSON

dckc:
refs: #1558 , https://github.com/Agoric/agoric-sdk/issues/7999

## Description

encode capData to 1 level of JSON, much like #1558, but

 - using `lastIndexOf` rather than a regex
 - fastcheck testing

motivation: senders pay by the byte etc.
 - https://github.com/Agoric/agoric-sdk/issues/7999

### Security Considerations

careful review for confusion vulnerability is in order

### Scaling Considerations

double-backslashes cost storage space

This flatter format is easier to read, and so easier to document in some senses, though there's a mixing of levels that's somewhat subtle.

### Testing Considerations

This has unit tests for specific examples plus fastcheck tests. Whether I stated the property exactly quite right is worth careful review.

### Upgrade Considerations

This is a DRAFT, pending:

 - [ ] figure out the whole upgrade story

cc @erights @gibson042

### DIFF
--- a/packages/marshal/src/capDataJSON.js
+++ b/packages/marshal/src/capDataJSON.js
@@ -1,0 +1,38 @@
+// @ts-check
+
+const { Fail } = assert;
+
+/** @param {unknown} x */
+const assertJSON = x => {
+  assert.typeof(x, 'string');
+  void JSON.parse(x);
+};
+
+/**
+ * @param {import('./types').CapData<unknown>} capData - with "simple" slots;
+ *   that is: slots whose JSON form has no occurrence of `:[`.
+ * @returns {string}
+ */
+export const capDataToJSON = ({ body, slots }) => {
+  assert(Array.isArray(slots));
+  const slotj = JSON.stringify(slots);
+  slotj.indexOf(':[') < 0 || Fail`expected simple slots`;
+  const body1 = body.replace(/^#/, '');
+  assertJSON(body1);
+  const json = `{"$body":${body1},"slots":${slotj}}`;
+  assertJSON(json);
+  return json;
+};
+
+export const JSONToCapData = json => {
+  assert.typeof(json, 'string');
+  json.startsWith('{"$body":') || Fail`expected $body`;
+  json.endsWith('}') || Fail`expected }`;
+  const pos = json.lastIndexOf(':[');
+  pos > 0 || Fail`expected slots`;
+  const body = `#${json.slice('{"$body":'.length, pos - ',"slots"'.length)}`;
+  const slotj = json.slice(pos + 1, -1);
+  const slots = JSON.parse(slotj);
+  Array.isArray(slots) || Fail`expected slots to be Array`;
+  return { body, slots };
+};
--- a/packages/marshal/test/test-marshal1.js
+++ b/packages/marshal/test/test-marshal1.js
@@ -35,10 +35,54 @@ const makeTestMarshal = () => {
   return m;
 };
 
+const brands = {
+  IST: Far('IST Brand', {}),
+  ATOM: Far('ATOM Brand', {}),
+};
+
 const suite = [
   { obj: null, json: '{"$body":null,"slots":[]}' },
   { obj: [1, 2, undefined], json: '{"$body":[1,2,"#undefined"],"slots":[]}' },
   { obj: { slots: [] }, json: '{"$body":{"slots":[]},"slots":[]}' },
+  // example from https://github.com/Agoric/agoric-sdk/issues/7999
+  {
+    obj: {
+      method: 'executeOffer',
+      offer: {
+        id: 'bid-1688229012779',
+        invitationSpec: {
+          callPipe: [['makeBidInvitation', [brands.ATOM]]],
+          instancePath: ['auctioneer'],
+          source: 'agoricContract',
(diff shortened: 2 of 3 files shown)


packages/marshal/src/capDataJSON.js
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * @param {import('./types').CapData<unknown>} capData - with "simple" slots;
+ *   that is: slots whose JSON form has no occurrence of `:[`.
+ * @returns {string}
+ */
+export const capDataToJSON = ({ body, slots }) => {
+  assert(Array.isArray(slots));
+  const slotj = JSON.stringify(slots);
+  slotj.indexOf(':[') < 0 || Fail`expected simple slots`;
+  const body1 = body.replace(/^#/, '');
```
mhofman, line 20: why not check body[0] === '#' and do `body.slice(1)`, I think that's a lot more efficient.

packages/marshal/src/capDataJSON.js
```diff
@@ -0,0 +1,38 @@
...
+  assertJSON(body1);
+  const json = `{"$body":${body1},"slots":${slotj}}`;
```
mhofman, line 22: In https://github.com/endojs/endo/issues/1478#issuecomment-1407500895 I suggest `body#`

packages/marshal/src/capDataJSON.js
```diff
@@ -0,0 +1,38 @@
...
+  assertJSON(json);
+  return json;
+};
+
+export const JSONToCapData = json => {
+  assert.typeof(json, 'string');
+  json.startsWith('{"$body":') || Fail`expected $body`;
```
mhofman, line 29: I guess this only works when this body is first in the serialized JSON, not second?

mhofman (commented):
I would so much rather we properly split encoding from serialization for marshal, as discussed in https://github.com/endojs/endo/issues/1478.

Also I would really prefer if we could find a way to partial parse JSON instead of relying of undocumented serialization constraints (body first, no spaces, etc.) I remember have a discussion with @gibson042 about what API we would need from JS to allow this.

gibson042 (commented):
This is just far too brittle for comfort, and doesn't feel like the right way to solve a "too much escaping" problem (assuming that is in fact what motivates it).

> I would so much rather we properly split encoding from serialization for marshal, as discussed in #1478.

I agree.

> Also I would really prefer if we could find a way to partial parse JSON instead of relying of undocumented serialization constraints (body first, no spaces, etc.) I remember have a discussion with @gibson042 about what API we would need from JS to allow this.

Yeah, but I don't know if we wrote it down (https://github.com/Agoric/agoric-private/issues/31#issuecomment-1494853056 is related but definitely distinct). At any rate, it's not _difficult_, although it would require going beyond the standard library.

packages/marshal/src/capDataJSON.js
dckc: `body.replace(/^#/, '')` handles both smallCaps and qclass, no? (I haven't tested it, though).

Why is `.slice(1)` significantly more efficient?

gibson042: > `body.replace(/^#/, '')` handles both smallCaps and qclass, no? (I haven't tested it, though).

I guess that depends upon what this function is expected to return. Regardless of the answer to that, though, CapData like ```{ body: `{"@qclass":"bigint","digits":"0"}`, slots: [] }``` and ```{ body: `#"+0"`, slots: [] }``` represent exactly the same data (`0n`) but would have distinct `String('{"$body":{"@qclass":"bigint","digits":"0"},"slots":[]}')` and `String('{"$body":"+0","slots":[]}')` return values (respectively) from the current implementation—which seems like a problem because there's no remaining signal differentiating smallcaps from the legacy encoding.

> Why is `.slice(1)` significantly more efficient?

The answer is implementation-specific, but basically comes down to being zero-copy.
```sh
$ esbench --eshost-option '-h V8,*XS*' \
'const unprefixed="a".repeat(1000), prefixed = "#" + unprefixed' '{
  "unprefixed.replace": `result = unprefixed.replace(/^#/, "")`,
  "unprefixed.slice": `result = unprefixed.startsWith("#") ? unprefixed.slice(1) : unprefixed`,
  "prefixed.replace": `result = prefixed.replace(/^#/, "")`,
  "prefixed.slice": `result = prefixed.startsWith("#") ? prefixed.slice(1) : prefixed`,
}'
#### Moddable XS
unprefixed.replace: 0.06 ops/ms
unprefixed.slice: 3.56 ops/ms
prefixed.replace: 0.07 ops/ms
prefixed.slice: 0.95 ops/ms

#### V8
unprefixed.replace: 29.41 ops/ms
unprefixed.slice: 100.00 ops/ms
prefixed.replace: 19.61 ops/ms
prefixed.slice: 62.50 ops/ms
```

dckc: ...
> > Why is `.slice(1)` significantly more efficient?
> 
> The answer is implementation-specific, but basically comes down to being zero-copy.

I guess I trained my regex intuitions in perl where such things are optimized out the wazoo.

Thanks for the `esbench` details.

PR #1808, opened 2023-10-05
fix(exo): allow non-enumerable methods

erights:
TODO explain

closes: #XXXX
refs: #XXXX

PR #1809, opened 2023-10-06
fix(exo): allow richer behaviorMethods

erights:
closes: #1817 
refs: #1815 #1808 #1773 #1766 

## Description

The intention of the exo-making methods was to be able to use a JS class' `.prototype` as a source of raw methods. This PR
* fixes the bugs reported in #1817 that prevented use of JS classes as a source of raw methods for an exo class.
* additionally accommodates JS class inheritance so that superclasses can also provide raw classes.
* provides tests that show use of JS classes to construct Exo classes.
* these tests also provide hypothetical helper abstractions for better supporting such use of classes. 
   * These helpers remain in tests because this PR merely demonstrates that such support is possible, without committing to the particular helpers to actually provide. We need at least to bikeshed from the currently awful names (`defineExoClassFromJSClass`).
* these tests demonstrate the use of both `super` and late-bound `self` in inheritance among these methods, with `super` calls bypassing guards, while `self` calls are guarded.

This PR also provides tests demonstrating the use of JS classes to directly provide unguarded behavior at the `@endo/pass-style` / `Far` level of abstraction. This PR makes no changes to the `src/` code at this level, demonstrating what support is already possible. Unlike the above exo-level support, we do *not* recommend the coding style shown in these pass-style level tests. However, neither can we reasonably prevent it, so we need to understand it.

### Security Considerations

Corresponding examples at 
https://github.com/endojs/endo/blob/d50a60bc3e5f477afad92b05d03335205e4b9451/packages/pass-style/test/test-far-wobbly-point.js#L132-L144

and
https://github.com/endojs/endo/blob/d50a60bc3e5f477afad92b05d03335205e4b9451/packages/exo/test/test-exo-wobbly-point.js#L183-L193

demonstrates that the demonstrated non-recommended use of class inheritance at the far level of abstraction has a security vulnerability absent from the use of class inheritance at the exo level of abstraction. This is one reason why we do not recommend the former but may recommend the latter.

### Scaling Considerations

If classes had large number of methods *and* class inheritance chains were deep, this has somewhat worse scaling than a JS class programmer would normally expect. However, if either of these are reasonably small, there should not be any significant scaling problem.

If we decide that we like and wish to promote the use of JS classes demonstrated by this PR, we should move some support (`defineExoClassFromJSClass`) into exported sources and we should document the whole approach. The purpose of this PR is not to establish such practice or decide on its exact shape, but merely to demonstrate that this PR makes such JS class-based practice possible.

In particular, we may be able to improve of the usability shown in this PR by using JS class decorators. It is a worthy experiment. (attn @mhofman )

### Testing Considerations

The `src` changes that this PR makes are only to `exo-tools.js` code shared among heap, virtual, and durable exos. However, the tests only test the behavior of heap exos. Once agoric-sdk is upgraded to depend on the changes from this PR, we should reproduce these tests for virtual and durable exos. Ideally, we would use Zones to write the tests once and then reuse it to test all three types of zone.

This PR makes no effort to test JS classes used for facets of exo kits. Nor does it provide any hypothetical helper functions to help write such exo kits. If we decide that this code pattern for using JS classes is attractive, a later PR should indeed extend the approach to exo kits.

### Upgrade Considerations

This PR makes no changes to data. For existing exo-making code that only uses object literals for own enumerable methods, this PR should not result in any behavior change or even internal representation change. The only code it should make a difference for is code that previously did not work because it needed the fixes and features provided by this PR. We are not aware of any such code prior to this PR.

### DIFF
--- a/packages/eventual-send/package.json
+++ b/packages/eventual-send/package.json
@@ -18,6 +18,12 @@
     "lint:types": "tsc",
     "lint:eslint": "eslint '**/*.js'"
   },
+  "exports": {
+    "./package.json": "./package.json",
+    ".": "./src/no-shim.js",
+    "./shim.js": "./shim.js",
+    "./utils.js": "./utils.js"
+  },
   "repository": {
     "type": "git",
     "url": "git+https://github.com/endojs/endo.git"
--- a/packages/exo/package.json
+++ b/packages/exo/package.json
@@ -33,6 +33,7 @@
   },
   "dependencies": {
     "@endo/env-options": "^0.1.4",
+    "@endo/eventual-send": "^0.17.6",
     "@endo/far": "^0.2.22",
     "@endo/pass-style": "^0.1.7",
     "@endo/patterns": "^0.2.6"
--- a/packages/exo/src/exo-tools.js
+++ b/packages/exo/src/exo-tools.js
@@ -1,5 +1,5 @@
+import { getMethodNames } from '@endo/eventual-send/src/local.js';
 import { E, Far } from '@endo/far';
-import { hasOwnPropertyOf } from '@endo/pass-style';
 import {
   listDifference,
   objectMap,
@@ -12,7 +12,6 @@ import {
   getInterfaceGuardPayload,
   getCopyMapEntries,
 } from '@endo/patterns';
-
 import { GET_INTERFACE_GUARD } from './get-interface.js';
 
 /** @typedef {import('@endo/patterns').Method} Method */
@@ -378,21 +377,6 @@ const bindMethod = (
 };
 
 /**
- *
- * @template {Record<PropertyKey, CallableFunction>} T
- * @param {T} behaviorMethods
- * @param {InterfaceGuard<{ [M in keyof T]: MethodGuard }>} interfaceGuard
- * @returns {T & import('./get-interface.js').GetInterfaceGuard<T>}
- */
-const withGetInterfaceGuardMethod = (behaviorMethods, interfaceGuard) =>
-  harden({
-    [GET_INTERFACE_GUARD]() {
-      return interfaceGuard;
-    },
-    ...behaviorMethods,
-  });
-
-/**
  * @template {Record<PropertyKey, CallableFunction>} T
  * @param {string} tag
  * @param {ContextProvider} contextProvider
@@ -408,13 +392,20 @@ export const defendPrototype = (
   interfaceGuard = undefined,
 ) => {
   const prototype = {};
-  if (hasOwnPropertyOf(behaviorMethods, 'constructor')) {
-    // By ignoring any method named "constructor", we can use a
+  const methodNames = getMethodNames(behaviorMethods).filter(
+    // By ignoring any method that seems to be a constructor, we can use a
     // class.prototype as a behaviorMethods.
-    const { constructor: _, ...methods } = behaviorMethods;
-    // @ts-expect-error TS misses that hasOwn check makes this safe
-    behaviorMethods = harden(methods);
-  }
+    key => {
+      if (key !== 'constructor') {
+        return true;
+      }
+      const constructor = behaviorMethods.constructor;
+      return !(
+        constructor.prototype &&
+        constructor.prototype.constructor === constructor
+      );
+    },
+  );
   /** @type {Record<PropertyKey, MethodGuard> | undefined} */
   let methodGuards;
   /** @type {import('@endo/patterns').DefaultGuardType} */
@@ -432,10 +423,9 @@ export const defendPrototype = (
       ...(symbolMethodGuards &&
         fromEntries(getCopyMapEntries(symbolMethodGuards))),
     });
+    assert(methodGuards !== undefined);
     defaultGuards = dg;
     {
-      const methodNames = ownKeys(behaviorMethods);
-      assert(methodGuards);
       const methodGuardNames = ownKeys(methodGuards);
       const unimplemented = listDifference(methodGuardNames, methodNames);
       unimplemented.length === 0 ||
@@ -446,12 +436,9 @@ export const defendPrototype = (
           Fail`methods ${q(unguarded)} not guarded by ${q(interfaceName)}`;
       }
     }
-    behaviorMethods = withGetInterfaceGuardMethod(
-      behaviorMethods,
-      interfaceGuard,
-    );
   }
-  for (const prop of ownKeys(behaviorMethods)) {
+
+  for (const prop of methodNames) {
     prototype[prop] = bindMethod(
       `In ${q(prop)} method of (${tag})`,
       contextProvider,
@@ -460,6 +447,22 @@ export const defendPrototype = (
       // TODO some tool does not yet understand the `?.[` syntax
       methodGuards && methodGuards[prop],
       defaultGuards,
+    );
+  }
+
+  if (!hasOwnPropertyOf(prototype, GET_INTERFACE_GUARD)) {
+    const getInterfaceGuardMethod = {
+      [GET_INTERFACE_GUARD]() {
+        // Note: May be `undefined`
+        return interfaceGuard;
+      },
+    }[GET_INTERFACE_GUARD];
+    prototype[GET_INTERFACE_GUARD] = bindMethod(
+      `In ${q(GET_INTERFACE_GUARD)} method of (${tag})`,
+      contextProvider,
+      getInterfaceGuardMethod,
+      thisfulMethods,
+      undefined,
     );
   }
 
--- a/packages/exo/src/exo-tools.js
+++ b/packages/exo/src/exo-tools.js
@@ -1,4 +1,5 @@
-import { getMethodNames } from '@endo/eventual-send/src/local.js';
+import { getMethodNames } from '@endo/eventual-send/utils.js';
+import { hasOwnPropertyOf } from '@endo/pass-style';
 import { E, Far } from '@endo/far';
 import {
   listDifference,
@@ -423,7 +424,6 @@ export const defendPrototype = (
       ...(symbolMethodGuards &&
         fromEntries(getCopyMapEntries(symbolMethodGuards))),
     });
-    assert(methodGuards !== undefined);
     defaultGuards = dg;
     {
       const methodGuardNames = ownKeys(methodGuards);
--- a/packages/exo/test/test-exo-class-js-class.js
+++ b/packages/exo/test/test-exo-class-js-class.js
@@ -1,0 +1,77 @@
+/* eslint-disable max-classes-per-file */
+/* eslint-disable class-methods-use-this */
+// eslint-disable-next-line import/order
+import { test } from './prepare-test-env-ava.js';
+
+import { passStyleOf } from '@endo/pass-style';
+import { M, getInterfaceGuardPayload } from '@endo/patterns';
+import { makeExo, defineExoClass } from '../src/exo-makers.js';
+
+// Based on FarSubclass1 in test-far-class-instances.js
+class DoublerBehaviorClass {
+  double(x) {
+    return x + x;
+  }
+}
+
+const DoublerI = M.interface('Doubler', {
+  double: M.call(M.lte(10)).returns(M.number()),
+});
+
+const doubler = makeExo('doubler', DoublerI, DoublerBehaviorClass.prototype);
+
+test('exo doubler using js classes', t => {
+  t.is(passStyleOf(doubler), 'remotable');
+  t.is(doubler.double(3), 6);
+  t.throws(() => doubler.double('x'), {
+    message: 'In "double" method of (doubler): arg 0: "x" - Must be <= 10',
+  });
+  t.throws(() => doubler.double(), {
+    message:
+      'In "double" method of (doubler): Expected at least 1 arguments: []',
+  });
+  t.throws(() => doubler.double(12), {
+    message: 'In "double" method of (doubler): arg 0: 12 - Must be <= 10',
+  });
+});
+
+// Based on FarSubclass2 in test-far-class-instances.js
+class DoubleAdderBehaviorClass extends DoublerBehaviorClass {
+  doubleAddSelfCall(x) {
+    const {
+      state: { y },
+      self,
+    } = this;
+    return self.double(x) + y;
+  }
+
+  doubleAddSuperCall(x) {
+    const {
+      state: { y },
+    } = this;
+    return super.double(x) + y;
+  }
+}
+
+const DoubleAdderI = M.interface('DoubleAdder', {
+  ...getInterfaceGuardPayload(DoublerI).methodGuards,
+  doubleAddSelfCall: M.call(M.number()).returns(M.number()),
+  doubleAddSuperCall: M.call(M.number()).returns(M.number()),
+});
+
+const makeDoubleAdder = defineExoClass(
+  'doubleAdderClass',
+  DoubleAdderI,
+  y => ({ y }),
+  DoubleAdderBehaviorClass.prototype,
+);
+
+test('exo inheritance self vs super call', t => {
+  const da = makeDoubleAdder(5);
+  t.is(da.doubleAddSelfCall(3), 11);
+  t.throws(() => da.doubleAddSelfCall(12), {
+    message:
+      'In "double" method of (doubleAdderClass): arg 0: 12 - Must be <= 10',
+  });
+  t.is(da.doubleAddSuperCall(12), 29);
+});
--- a/packages/exo/test/test-exo-wobbly-point.js
+++ b/packages/exo/test/test-exo-wobbly-point.js
@@ -10,10 +10,9 @@
 /* eslint-disable-next-line import/order */
 import { test } from './prepare-test-env-ava.js';
 
-// TODO enable import of getMethodNames without deep import
 // eslint-disable-next-line import/order
-import { getMethodNames } from '@endo/eventual-send/src/local.js';
-import { passStyleOf, Far } from '@endo/pass-style';
+import { getMethodNames } from '@endo/eventual-send/utils.js';
+import { passStyleOf, Far, GET_METHOD_NAMES } from '@endo/pass-style';
 import { M } from '@endo/patterns';
 import { defineExoClass } from '../src/exo-makers.js';
 import { GET_INTERFACE_GUARD } from '../src/get-interface.js';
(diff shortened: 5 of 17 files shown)


erights:
@michaelfig , reviewers,

I rebased and resolved merge conflicts with https://github.com/endojs/endo/pull/1831

packages/exo/src/exo-tools.js
```diff
@@ -439,12 +426,22 @@ export const defendPrototype = (
...
+    prototype[GET_INTERFACE_GUARD] = bindMethod(
+      `In ${q(GET_INTERFACE_GUARD)} method of (${tag})`,
+      contextProvider,
+      getInterfaceGuardMethod,
+      thisfulMethods,
+      undefined,
+    );
+  }
+
+  return Far(tag, /** @type {T} */ (prototype));
```
michaelfig, line 444: Is there a particular reason to make the presence of the `GET_INTERFACE_GUARD` method conditional?  Shouldn't it return the actual interface used in defendPrototype rather than the user-supplied value?

I ask because omitting `GET_INTERFACE_GUARD` instead of always defining it and making it return *something* makes it hard for remote callers to use.  They currently have no way to sniff the presence or absence of the method on a remote object.

I would argue that the original typing of this function is the appropriate one, and would encourage you to find a way to satisfy it instead of using a (sometimes) inaccurate return type.

packages/exo/package.json
```diff
@@ -33,6 +33,7 @@
   },
   "dependencies": {
     "@endo/env-options": "^0.1.4",
+    "@endo/eventual-send": "^0.17.6",
```
michaelfig, line 36: This seems like incorrect layering.  Can you please put `getMethodNames` somewhere better, like `patterns` or `pass-style`?  If you can't remove it from `@endo/eventual-send`, at least deprecate it there, but make a new version in the better place.

packages/exo/test/test-exo-class-js-class.js
```diff
@@ -0,0 +1,79 @@
+/* eslint-disable max-classes-per-file */
+/* eslint-disable class-methods-use-this */
+// eslint-disable-next-line import/order
+import { test } from './prepare-test-env-ava.js';
+
+import { passStyleOf } from '@endo/pass-style';
+// import { M, getInterfaceGuardPayload } from '@endo/patterns';
+// import { defineExoClass, makeExo } from '../src/exo-makers.js';
```
michaelfig, line 8: Stray comments?

packages/exo/test/test-exo-wobbly-point.js
```diff
@@ -0,0 +1,205 @@
...
+ * and
+ * test-far-wobbly-point.js
+ */
+
+/* eslint-disable class-methods-use-this */
+/* eslint-disable max-classes-per-file */
+/* eslint-disable-next-line import/order */
+import { test } from './prepare-test-env-ava.js';
+
+// TODO enable import of getMethodNames without deep import
```
michaelfig, line 13: Agreed.  `eventual-send` is a layer atop `pass-style` so maybe best to put it there?

packages/exo/test/test-non-enumerable-methods.js
```diff
@@ -0,0 +1,129 @@
...
+  Object;
+
+const { ownKeys } = Reflect;
+
+/**
+ * Borrowed from https://github.com/endojs/endo/pull/1815 to avoid
+ * depending on it being merged. TODO If it is merged, then delete this
+ * copy and import `objectMetaMap` instead.
+ *
+ * Like `objectMap`, but at the reflective level of property descriptors
```
michaelfig, line 19: Pretty neat!

packages/pass-style/package.json
```diff
@@ -37,6 +37,7 @@
     "@fast-check/ava": "^1.1.5"
   },
   "devDependencies": {
+    "@endo/eventual-send": "^0.17.6",
```
michaelfig, line 40: Yeah.  `@endo/pass-style` really does seem like it should be home for `getMethodNames`.

packages/exo/src/exo-tools.js
michaelfig: I should clarify:
> the original typing of this function (defendPrototype) is the appropriate one

What I mean is to keep the deleted `return Far(.... /** @type {T & import('./get-interface.js').GetInterfaceGuard<T>} 
 */ ...);` intact, and define `GET_INTERFACE_GUARD` before it, unconditionally, so that no remote-sniffing needs to happen in order to invoke it.  Then, for bonus points (or at least worth a TODO), change `GET_INTERFACE_GUARD`'s return value always to be something meaningful.

packages/exo/test/test-exo-wobbly-point.js
erights: Actually, pass-style devDepends on eventual-send, not the other way.

michaelfig: > Actually, pass-style devDepends on eventual-send, not the other way.

Oh, I was confused. This actually makes sense: eventual-send doesn't rely on the pass-style or marshal rules.

Sorry for the mixup.

packages/exo/src/exo-tools.js
erights: > instead of always defining it and making it return something makes it hard for remote callers to use

@michaelfig 
I agree this would be an improvement. However, I started on it, and this improvement is a more significant refactor than I anticipated. (Either that, or I missed how to do it simply.) Could we move ahead with this PR without this improvement, together with my filing an issue, and then revisit this improvement in a later PR?

> actual interface ... than the user-supplied value

Currently, there is no actual interface distinct from the user supplied value. The refactor would be to always build it, including in the thisless case where there isn't really even an implicit an interface guard.

Even the improvement as stated doesn't quite obviate the sniffing, since simple `Far` objects would still not have a `GET_INTERFACE_GUARD` method. But addressing that as well is seems impossible without grossly violating our layering.

> keep the deleted ...

That part is now done.

erights: https://github.com/endojs/endo/issues/1844

I assigned it to both of us.

erights: It just occurred to me that a more pleasant interim solution while waiting for #1844 is to simply have GET_INTERFACE_METHOD() return undefined in that cases where it currently is absent. I just made this change. It still doesn't cover the `Far` case, but it does somewhat cover all the other exo/kind cases by at least making it clear that it isn't yet helping with further sniffing.

erights: It does occur to me that if we are content with the possibility of GET_INTERFACE_GUARD() permanently being allowed to return `undefined`, then it would not be a horrible layering violation to have `Far` add only this. Please advise.

michaelfig: > `GET_INTERFACE_GUARD()` permanently being allowed to return `undefined`, [and] have `Far` add only this.

💯 I really like this option.  Brilliant!

packages/eventual-send/package.json
```diff
@@ -18,6 +18,12 @@
     "lint:types": "tsc",
     "lint:eslint": "eslint '**/*.js'"
   },
+  "exports": {
+    "./package.json": "./package.json",
+    ".": "./src/no-shim.js",
+    "./shim.js": "./shim.js",
+    "./utils.js": "./utils.js"
+  },
```
erights, line 26: Reviewers, is it ok to add this? It will break any deep imports to other names.

packages/exo/src/exo-tools.js
erights: I think you'll like my new variation even better: Have `Far` introduce a `GET_METHOD_NAMES` meta method that applies to all objects made with `Far` or any object they inherit from. This is only a miranda method -- it is only installed if the user does not provide one.

Then, at the cost of a round trip, you can use that to determine whether there is a method named `GET_INTERFACE_GUARD`. Thus, I was able to restore the conditional. I did go further and add a further conjunct to ensure that there is not already a `GET_INTERFACE_GUARD` method, so it is also a miranda method.

Given this, I propose that it is now the correct decision for `GET_INTERFACE_GUARD` to return the interface guard the user provided, since the user can now combine it with the result of `GET_METHOD_NAMES` to calculate the effective interface guard.

PTAL

erights: I did make that variation as a separate `feat(pass-style)` commit, so we could back out.

erights: This new feature is somewhat tested by the tests I had to modify. I'll wait on writing more tests pending your feedback on this approach.

packages/exo/test/test-non-enumerable-methods.js
erights: See https://github.com/endojs/endo/pull/1815

packages/eventual-send/package.json
michaelfig: Looks like there are no deep imports in agoric-sdk, so I would tentatively say yes.  Maybe ensure that it's in a separate commit, and mark it as a breaking change like `chore(eventual-send)!: restrict exports`.

packages/exo/src/exo-tools.js
```diff
@@ -439,6 +427,21 @@ export const defendPrototype = (
     );
   }
 
+  if (interfaceGuard && !hasOwnPropertyOf(prototype, GET_INTERFACE_GUARD)) {
```
michaelfig, line 430: I still don't like the first part of this conditional (`if (interfaceGuard &&`), since it means the return type of defendPrototype is not consistent with the static type of the return value (`/** @type {T & import('./get-interface.js').GetInterfaceGuard<T> */`).  This is no longer a sniffing problem, since you solved that with GET_METHOD_NAMES, but it is a Typescript problem (inconsistent return value that makes the static type incorrect).

Could you just do this condition based on `hasOwnPropertyOf`, but not based on the truthiness of `interfaceGuard`?

packages/pass-style/src/make-far.js
```diff
@@ -128,16 +129,57 @@ export const Remotable = (
 };
 harden(Remotable);
 
+/**
+ * The name of the automatically added default meta-method for obtaining a
+ * list of all methods of an object declared with `Far`, or an object that
+ * inherits from an object declared with `Far`.
+ *
+ * Modeled on `GET_INTERFACE_GUARD` from
```
michaelfig, line 137: Sentence to complete after "from"?

michaelfig (commented):
Looking good, just some quibbles.  I'd approve based on their resolution, but I have a feeling that some of the other reviewers you tagged should weigh in first.

packages/eventual-send/package.json
erights: Done

packages/pass-style/src/make-far.js
erights: Done

kriskowal:
Aside: If I convince the OCapN folks that method invocations should be case-normalized, then a method named “get interface guard” would only be addressable as `get-interface-guard` in Guile, `get_interface_guard` in Python, `getInterfaceGuard` in JavaScript, `GetInterfaceName` in Go &c. This is sufficiently far over the horizon and addressable with a non-breaking change (duplicate the method) that it doesn’t make sense to belabor now.

packages/exo/src/exo-tools.js
erights: Frankly, I'm not sure this is better. The runtime behavior has weaker invariants, since the method may exist but return `undefined`. Either way, we needed to weaken the static types. But had we weakened the static type of `defendPrototype` instead, we would not have needed to weaken the runtime invariants.

But in any case, done.

packages/exo/src/exo-tools.js
erights: Tests done.

erights:
At https://github.com/endojs/endo/pull/1809#pullrequestreview-1714037916 @michaelfig writes:

> Looking good, just some quibbles. I'd approve based on their resolution, 

All quibbles resolved. PTAL.


> but I have a feeling that some of the other reviewers you tagged should weigh in first.

@turadg @mhofman , Care to? Thanks.

erights:
Reviewers,
In case @kriskowal is able, before the hackathon, to do a new endo release & upgrade agoric-sdk to depend on it, I'd really like to be able to get this in so I could build on it. Given the reviewing @michaelfig has already done, you may find the remaining review work could be done quickly, perhaps? Just sayin ;) . Thanks.

packages/exo/test/test-exo-class-js-class.js
```diff
@@ -0,0 +1,77 @@
...
+const DoubleAdderI = M.interface('DoubleAdder', {
+  ...getInterfaceGuardPayload(DoublerI).methodGuards,
+  doubleAddSelfCall: M.call(M.number()).returns(M.number()),
+  doubleAddSuperCall: M.call(M.number()).returns(M.number()),
+});
+
+const makeDoubleAdder = defineExoClass(
+  'doubleAdderClass',
+  DoubleAdderI,
+  y => ({ y }),
```
turadg, line 65: out of scope but what would it take to have the class constructor define the init? making `this.state` available as it is in methods?

packages/pass-style/src/make-far.js
```diff
@@ -128,16 +129,57 @@ export const Remotable = (
...
 
+/**
+ * The name of the automatically added default meta-method for obtaining a
+ * list of all methods of an object declared with `Far`, or an object that
+ * inherits from an object declared with `Far`.
+ *
+ * Modeled on `GET_INTERFACE_GUARD` from `@endo/exo`.
+ *
+ * TODO Name to be bikeshed. Perhaps even whether it is a
+ * string or symbol to be bikeshed.
```
turadg, line 140: after this is released?

What will the impact be of changing it then?

packages/pass-style/src/make-far.js
```diff
@@ -128,16 +129,57 @@ export const Remotable = (
...
+ *
+ * TODO Beware that an exo's interface can change across an upgrade,
```
turadg, line 142: what is the action TO DO ?

turadg (commented):
I raised some issues at the periphery with assertions and documentation.

> We need at least to bikeshed from the currently awful names (`defineExoClassFromJSClass`).

I think "bikeshed" here is facetious but I want to point out that API naming in Endo merits careful consideration. They will stick around a long time and we want to avoid developers being confused by them or having misconceptions about what they do.

That said, I don't think `defineExoClassFromJSClass` is awful. It's long but it says what it does. It's probably also not the end state for the dev-facing API because I expect defining the class with decorators will be the most common.

erights:
> We need at least to bikeshed from the currently awful names (`defineExoClassFromJSClass`).
> 
> > I think "bikeshed" here is facetious but I want to point out that API naming in Endo merits careful consideration. They will stick around a long time and we want to avoid developers being confused by them or having misconceptions about what they do.

I agree completely, which is why these awful names --- as well as the entirety of the `defineExoClassFromJSClass` helper abstractions --- are in a `test/` file rather than a `src/` file. I want to put a lot more thought into them before providing them for actual usage. Their purpose in the test is to establish the that `src/` changes of this PR enable such things to be possible.

That said, even as a `test/` POC of possibility, I'm happy to consider suggested improvements.

packages/exo/src/exo-tools.js
```diff
@@ -408,10 +400,9 @@ export const defendPrototype = (
       ...(symbolMethodGuards &&
         fromEntries(getCopyMapEntries(symbolMethodGuards))),
     });
+    assert(methodGuards !== undefined);
```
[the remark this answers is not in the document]
erights: > Though how could it possibly be undefined here after the assignment above?

Obvious to you and I, but at some point in the past it seemed not to be obvious to TypeScript. The only reason for the `assert` was to let TypeScript know that within the immediately following code `methodGuards` had only the non-`undefined` portion of its type.

In light of your comment, I wanted to see again what TypeScript complaint I'd get if I deleted it. Happy to report that it no longer causes a type error. I don't know what changed, but in any case: Gone. Resolving.

packages/exo/test/test-exo-class-js-class.js
erights: Not sure. @mhofman has raised the possibility of providing `this.state` to the `init` function. But the programmer would still manually write the `init` function. If the constructor were to define it, what information would it figure it out from?

Since "out of scope", resolving. But please continue the conversation in any venue you'd like. Thanks.

packages/pass-style/src/make-far.js
erights: It won't be pleasant. This being a non-`test/` name, if you have a suggestion for a better name, the earlier we switch to it the better.

The symbol issue is tied to an ongoing ocapn controversy about whether symbols should be Passable at all. The current consensus ocapn agreement does include them, but @kriskowal has good arguments about why we shouldn't. We're got very few, but this PR does add to that. OTOH, if we are keeping symbols as Passable at all, I find it very attractive that the meta methods at stake in this PR should be symbols.

At least this PR exports these method-names as the value of importable variables. If all code using these do so only via the imported variable, that will ease the transition code. But it'll still be substantial because these become protocol between code linked against different versions.

packages/pass-style/src/make-far.js
erights: Good point. I replaced "TODO" with "HAZARD:". Resolving.

packages/pass-style/src/make-far.js
erights: Leaving this conversation unresolved for now. Reviewers + @kriskowal , what do you think we should do about naming these methods?

turadg: until there's a durable name, you could indicate that it shouldn't be relied upon (even though it's exported). e.g. by prefixing with underscores. Same for `__GET_METHOD_GUARD`.

turadg (approved):
No blocking concerns. I do think the underscores are worth considering before merging.

I'd also suggest cleaning up the commits to reduce noise in the changelog. Most of the "fix" are really "fixup" that should be squashed into noteworthy commits. 

Btw, if a commit shouldn't appear in the changelog, it shouldn't have conventional commit message. Noise in the commit history isn't as bad as noise in the changelog.

packages/pass-style/src/make-far.js
michaelfig: > you could indicate that it shouldn't be relied upon
IIUC the naming controversy is not for `export const GET_METHOD_NAMES`, it's for its value, `Symbol.for('getMethodNames')`.  The `GET_METHOD_NAMES` exported binding can be used widely in importing code without any problems, but if the value (the `Symbol.for`) is changed, there will be an incompatibility.

Having said that, @erights, this is a case where a Symbol is used to avoid unintentional conflicts with an object's existing properties.  Interestingly, leading underscores have performed this task in other situations.  I would rather not add to the ocapn controversy to keep our options open.  Why not:

```js
export const GET_METHOD_NAMES = '__getMethodNames';
export const GET_INTERFACE_GUARD = "__getInterfaceGuard';
```

erights: > IIUC the "bikeshed" is not for export const GET_METHOD_NAMES, it's for its value,

Correct.

> ```js
> export const GET_METHOD_NAMES = '__getMethodNames';
> export const GET_INTERFACE_GUARD = '__getInterfaceGuard';
> ```

Our more common convention is both a prefix and a suffix double underbar

```js
export const GET_METHOD_NAMES = '__getMethodNames__';
export const GET_INTERFACE_GUARD = '__getInterfaceGuard__';
```

Done.

erights:
> I'd also suggest cleaning up the commits to reduce noise in the changelog. Most of the "fix" are really "fixup" that should be squashed into noteworthy commits.
> 
> Btw, if a commit shouldn't appear in the changelog, it shouldn't have conventional commit message. Noise in the commit history isn't as bad as noise in the changelog.

Manually squashed those away before merging. Done.

PR #1812, opened 2023-10-06
feat(transforms): add SES censorship evasion function

boneskull:
## Description

Extracts the `rewriteComment` function from `@endo/bundle-source/src/transform.js` and exposes it as `transformComment` in the `evade-censor` export. This should be considered an experimental API.

## Motivation

We could use this functionality in LavaMoat itself, so extracting it into a module lighter than `@endo/bundle-source` seems appropriate.

The function is not exported from the main entry point, nor documented publicly, as to avoid committing to a API we may not be ready to commit to.  I can change this, if desired!

## Notes

- `@babel/types` is added as a production dependency, because the "public" API surface depends on types from this module.
- Did not export the new function from the main entry point.
- Added some unit tests.
- Used homoglyphs.
- Removed the dummy test index.
- Added barebones `README.md`

## Questions

- Why do we make everything a `CommentBlock`?
- Why do we replace the end-of-comment marker?
- Is it preferable to use character escapes or no?


packages/transforms/src/evade-censor.js
```diff
@@ -0,0 +1,55 @@
...
+  // Within comments...
+  node.value = node.value
+    // ...strip extraneous comment whitespace
+    .replace(/^\s+/gm, ' ')
+    // ...replace HTML comments with a defanged version to pass SES restrictions.
+    .replace(HTML_COMMENT_START_RE, '<!X-')
+    .replace(HTML_COMMENT_END_RE, '-X>')
+    // ...replace import expressions with a defanged version to pass SES restrictions
+    // (featuring homoglyphs for @kriskowal)
+    .replace(IMPORT_RE, '\u{406}\u{16D6}\u{420}\u{39F}\u{13D2}\u{422}$2')
```
kriskowal, line 42: summoning @mhofman, resident homoglyph idea haver. (I do not need credit)

packages/transforms/src/evade-censor.js
```diff
@@ -0,0 +1,55 @@
...
+ */
+export function transformComment(node, unmapLoc) {
+  node.type = 'CommentBlock';
+  // Within comments...
+  node.value = node.value
+    // ...strip extraneous comment whitespace
+    .replace(/^\s+/gm, ' ')
+    // ...replace HTML comments with a defanged version to pass SES restrictions.
+    .replace(HTML_COMMENT_START_RE, '<!X-')
+    .replace(HTML_COMMENT_END_RE, '-X>')
```
kriskowal, line 39: Also potential homoglyph, e.g., `\u2010`.

kriskowal (commented):
What do you think of exporting `evadeCensor(source, sourceMap) => {source, sourceMap}` for less public coupling?

kriskowal (commented):
Also would like:

- [x] unit test
- [ ] integration test in compartment-mapper
- [ ] refactor bundle-source in place to take on this dependency

Let’s be sure to separate the commit that changes the behavior to use homoglyphs so it can be easily reverted if necessary.

kriskowal:
I defer to @michaelfig or @erights regarding the comment end marker escape and why we replace `//` with `/*` style comments. We should capture the reasoning in the code!

I have no strong feelings about how to escape homoglphys.

Also worth noting that there are other evasive transforms we hope to add one day to provide better compatibility with the ecosystem, particularly transforms that break strings that have censored terms.

boneskull:
@kriskowal Which unit test do you mean?

packages/transforms/src/evade-censor.js
boneskull: Is there a reason to use homoglyphs in addition to, uh, _lulz_?

erights:
> why we replace // with /* style comments. We should capture the reasoning in the code!

I do not remember being aware that we changed comments this way, so I also do not remember why. If I was aware at the time, my apologies for not capturing or at least leaving more clues.

kriskowal:
> @kriskowal Which unit test do you mean?

There is none as such and we are relying on integration tests inside bundle-source for the effect. I’m willing to continue in that vein.

packages/transforms/src/evade-censor.js
kriskowal: They have the benefit of preserving the column numbers for subsequent code on the same line. But mostly lulz.

boneskull:
> There is none as such and we are relying on integration tests inside bundle-source for the effect. I’m willing to continue in that vein.

For posterity: this PR _does_ include some unit tests.

boneskull:
> What do you think of exporting `evadeCensor(source, sourceMap) => {source, sourceMap}` for less public coupling?

@kriskowal 

This sounds like I'd need to copy `makeLocationUnmapper()` and other functions from `bundle-source/src/transform.js` here as well.

Was the intention to rehome the entirety of `bundle-source/src/transform.js` into the `transform` pkg?  Apologies if that wasn't clear.  And if so, you want it all in this PR (with multiple, logical commits)?

kriskowal:
> > What do you think of exporting `evadeCensor(source, sourceMap) => {source, sourceMap}` for less public coupling?
> 
> @kriskowal
> 
> This sounds like I'd need to copy `makeLocationUnmapper()` and other functions from `bundle-source/src/transform.js` here as well.
> 
> Was the intention to rehome the entirety of `bundle-source/src/transform.js` into the `transform` pkg? Apologies if that wasn't clear. And if so, you want it all in this PR (with multiple, logical commits)?

I had not thought that many steps ahead, but that would seem to be necessary in order to decouple the public interface from the encapsulated dependencies, so I’m in favor.

boneskull:
this is getting closer, but I need to add some more tests.  I'm not sure how unit-y I can make them, though...

boneskull:
@kriskowal Please let me know if my test fixtures in `@endo/transforms` are overkill

boneskull:
I suppose I should update `README.md` with API docs too

packages/transforms/README.md
```diff
@@ -1 +1,55 @@
...
 
+> Source transforms for [SES](https://github.com/endojs/endo/tree/master/packages/ses)-enabled applications
+
+This package currently provides a transform to modify comments which would otherwise be rejected outright by SES.
+
+## Example
+
+```js
+// ESM example
+import { transformSource } from '@endo/transforms';
```
kriskowal, line 11: I’m imagining that the scope of the `transforms` package allows for additional transforms unrelated to SES censorship evasion. Do you think it might make sense to narrow the scope of the package by making its name more specific, or have a more specific sub-path for SES censorship evasion? Can we avoid a situation where a dependent module has to entrain all of the package’s transforms when it only needs one?

packages/transforms/README.md
```diff
@@ -1 +1,55 @@
...
+await fs.writeFile('./dist/index.ses.js.map', JSON.stringify(map));
+```
+
+The censorship-evasion function can also be used directly, if you have parsed
+your source into an AST via Babel:
+
+```ts
+// TypeScript example
+import { evadeCensor } from '@endo/transforms';
+import { type Comment } from '@babel/types';
```
kriskowal, line 35: Supposing we switched the internal implementation from Babel to Esprima, removing the Babel API surface would be a breaking change. I’m thinking it might be possible to anticipate that kind of migration, allowing users to continue upgrading minor versions during the transition, shifting from the Babel version to the Esprima on their own time, but only entraining the dependency on one or the other in any particular release. Perhaps “babel” could appear in the module specifier when the user explicitly elects that level of coupling.

packages/transforms/README.md
```diff
@@ -1 +1,55 @@
...
+const node: Comment = getCommentNodeFromSomewhere();
+
+/**
+ * After this, the comment will become `@property {ІᛖРΟᏒТ('foo').Bar} bar` which
+ * is no longer understandable by TypeScript, but will be allowed by SES. 
+ * 
+ * Note that  this could be avoided entirely by stripping comments during, say,
+ * a bundling phase.
+ */
+evadeCensor(node);
```
kriskowal, line 50: I prefer `evadeCensor` over `transformSource` in general. Can we use this root for both the stringy and the ASTy APIs?

packages/bundle-source/src/nested-evaluate-and-get-exports.js
```diff
@@ -96,8 +96,8 @@ export async function bundleNestedEvaluateAndGetExports(
         moduleFormat === 'nestedEvaluate' && !fileName.startsWith('_virtual/');
 
       const { code: transformedCode } = await transformSource(code, {
-        sourceMapUrl: pathname,
         sourceMap: chunk.map,
+        sourceUrl: pathname,
```
kriskowal, line 100: Thank you for spotting this defect before it became difficult to fix!

kriskowal (commented):
The substance of this change is perfect. I challenge you to consider my comments about anticipating future changes in the names and paths of functions and modules.

packages/transforms/README.md
boneskull: Renaming this to e.g., `@endo/evasive-transform`  and limiting scope seems reasonable--though it potentially means adding more pkgs in the future.  At first glance, Endo doesn't seem like a project that fears this sort of tradeoff. 😄

packages/transforms/README.md
boneskull: I think I'd prefer:

1. _not_ exporting the underlying `evadeCensor` function, thus
2. decoupling the public API from these implementations

As it currently stands, LavaMoat shouldn't need anything except the "main" function--which is the same as current usage in `@endo/bundle-source`.

If, in the future, someone needs a lower-level, implementation-specific API, we could provide it by splitting this package in two: expose these APIs in the new package then consume it from _this_ higher-level package.

packages/transforms/README.md
kriskowal: For the record, I laughed, and my laugh was audible in an adjacent room, but not audible to our neighbors. Thank you.

boneskull:
@kriskowal OK, I've

1. Renamed the package (I did not change the version number, however)
2. Updated the README
3. Narrowed the API so that only a single function is exported. Renamed it
4. Created a couple additional adapter functions so that, ideally, swapping out the underlying AST implementation won't be so painful
5. Accidentally discovered how to implement function overloading via JSDoc

boneskull:
linefeeds. lol

packages/evasive-transform/src/transform-comment.js
```diff
@@ -0,0 +1,48 @@
...
+  // Within comments...
+  node.value = node.value
+    // ...strip extraneous comment whitespace
+    .replace(/^\s+/gm, ' ')
+    // ...replace HTML comments with a defanged version to pass SES restrictions.
+    .replace(HTML_COMMENT_START_RE, '<!X-')
+    .replace(HTML_COMMENT_END_RE, '-X>')
+    // ...replace import expressions with a defanged version to pass SES restrictions
+    // (featuring homoglyphs for @kriskowal)
+    .replace(IMPORT_RE, '\u{406}\u{16D6}\u{420}\u{39F}\u{13D2}\u{422}$2')
```
kriskowal, line 42: Could be more subtle:

```suggestion
    .replace(IMPORT_RE, 'im\u{440}ort$2')
```

packages/evasive-transform/src/transform-comment.js
```diff
@@ -0,0 +1,48 @@
...
+ * @param {import('./location-unmapper.js').LocationUnmapper} [unmapLoc]
+ */
+export function transformComment(node, unmapLoc) {
+  node.type = 'CommentBlock';
+  // Within comments...
+  node.value = node.value
+    // ...strip extraneous comment whitespace
+    .replace(/^\s+/gm, ' ')
+    // ...replace HTML comments with a defanged version to pass SES restrictions.
+    .replace(HTML_COMMENT_START_RE, '<!X-')
```
kriskowal, line 38:
```suggestion
    .replace(HTML_COMMENT_START_RE, '<!\u{2010}-')
```

packages/evasive-transform/src/transform-comment.js
```diff
@@ -0,0 +1,48 @@
...
+    .replace(HTML_COMMENT_END_RE, '-X>')
```
kriskowal, line 39:
```suggestion
    .replace(HTML_COMMENT_END_RE, '-\u{2010}>')
```

kriskowal (approved):
Left some homoglyph suggestions that should make comments look the same to a developer in a debugger without tripping the censor.

boneskull:
OK. Don't touch anything. @kriskowal can I be done please

boneskull:
8d9d2b7 was committed from windows where I am evidently not setup properly with gpg.

kriskowal:
Thanks, @boneskull. You’re green. Would you like to squash some or all of the commits or shall I command Github to do so on your behalf?

boneskull:
@kriskowal Squashed

Issue #1805, opened 2023-10-05
Bundler transform issue (?)


kumavis:
[ Issue description in progress ]

I encountered an issue when executing a compartment-mapper bundle of the following file

before transform
```js
// src/ApiError.ts
var ErrorCode = /* @__PURE__ */ ((ErrorCode2) => {
  ErrorCode2[ErrorCode2["EPERM"] = 1] = "EPERM";
  // ...
  return ErrorCode2;
})(ErrorCode || {});
// ...
export {
  // ...
  ErrorCode,
  // ...
}
```

after transform
```js
// src/ApiError.ts
var $c‍_ErrorCode=/* @__PURE__ */((ErrorCode2)=> {
  ErrorCode2[ErrorCode2["EPERM"]=  1]=  "EPERM";
  // ...
  return ErrorCode2;
 })(ErrorCode|| {});ErrorCode=$c‍_ErrorCode;
 ```
 
causing
`ReferenceError: ErrorCode is not defined`

kumavis:
Here's a failing test for `static-module-record` https://github.com/endojs/endo/pull/1806/commits/efabcda3d6ad9fefb7aeb8f320ad0795e555ed21

PR #1813, opened 2023-10-07
test(static-module-record): Validate export var

kriskowal:
closes: #1805

This change adds tests that validate the `StaticModuleRecord` treatment for exported vars.

The search to isolate the defect @kumavis found continues.


kumavis (approved):

kumavis:
I think the error I saw was in a compartment-mapper bundle

PR #1820, opened 2023-10-12
fix: Support absence of TextEncoder/TextDecoder/console

### LINKED ISSUE #1819: lockdown() fails in `eshost` environments
## Describe the bug

`lockdown()` throws errors in `eshost` environments.

## Steps to reproduce

```sh
cd /path/to/endo
printf 'import "@endo/init";\n' | npx rollup -p @rollup/plugin-node-resolve -f iife > bug.js
eshost bug.js
```

## Actual behavior

Errors in every engine.
```console
$ eshost bug.js
#### GraalJS

ReferenceError: TextEncoder is not defined

#### JavaScriptCore

TypeError: baseConsole is not an Object. (evaluating 'name in baseConsole')

#### Moddable XS

ReferenceError: (anonymous-586): get console: undefined variable

#### SpiderMonkey

TypeError: baseConsole[level] is not a function

#### V8
console.warn: Removing intrinsics.Promise.withResolvers
ReferenceError: TextEncoder is not defined
```

## Expected behavior

Successful evaluation.

gibson042:
Fix #1819

## Description

Tolerate the absence of non-ECMA-262 web platform APIs.

### Security Considerations

This follows existing patterns and hardens all introduced new non-primitive values. It will result in calls to a global `print` function that would not have otherwise happened, but only when it exists and `console` is undefined (but I believe any code capable of creating those circumstances could have overridden `console` anyway).

### Scaling Considerations

n/a

The non-web-platform environments of `eshost` invocations are sufficiently esoteric that I don't think documentation needs updating.

### Testing Considerations

For similar reasons, I have not introduced new tests (although I suspect that _could_ be done).

### Upgrade Considerations

None.

### DIFF
--- a/packages/ses/src/error/console.js
+++ b/packages/ses/src/error/console.js
@@ -174,6 +174,10 @@ freeze(ErrorInfo);
 
 /** @type {MakeCausalConsole} */
 const makeCausalConsole = (baseConsole, loggedErrorHandler) => {
+  if (!baseConsole) {
+    return undefined;
+  }
+
   const { getStackString, tagError, takeMessageLogArgs, takeNoteLogArgsArray } =
     loggedErrorHandler;
 
--- a/packages/ses/src/error/tame-console.js
+++ b/packages/ses/src/error/tame-console.js
@@ -1,18 +1,47 @@
 // @ts-check
 
-import { TypeError, globalThis } from '../commons.js';
+import {
+  TypeError,
+  apply,
+  defineProperty,
+  freeze,
+  globalThis,
+} from '../commons.js';
 import { loggedErrorHandler as defaultHandler } from './assert.js';
 import { makeCausalConsole } from './console.js';
 import { makeRejectionHandlers } from './unhandled-rejection.js';
 import './types.js';
 import './internal-types.js';
 
+const wrapLogger = (logger, thisArg) =>
+  freeze((...args) => apply(logger, thisArg, args));
+
 // eslint-disable-next-line no-restricted-globals
 const originalConsole = /** @type {VirtualConsole} */ (
+  // eslint-disable-next-line no-nested-ternary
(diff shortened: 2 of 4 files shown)


packages/ses/src/error/console.js
```diff
@@ -174,6 +174,10 @@ freeze(ErrorInfo);
 
 /** @type {MakeCausalConsole} */
```
erights, line 175: Please adjust the type `MakeCausalConsole` to `@returns {VirtualConsole | undefined}`

packages/ses/src/error/console.js
```diff
@@ -174,6 +174,10 @@ freeze(ErrorInfo);
...
 const makeCausalConsole = (baseConsole, loggedErrorHandler) => {
+  if (!baseConsole) {
+    return baseConsole;
```
erights, line 178: For the sake of being sounder wrt the type I suggest above, what do you think of changing this to

```suggestion
    return undefined;
```

? Or do I miss the intent of this code?

packages/ses/src/error/tame-console.js
```diff
@@ -1,14 +1,45 @@
...
+
 // eslint-disable-next-line no-restricted-globals
-const originalConsole = console;
+const originalConsole = /** @type {typeof console} */ (
+  // eslint-disable-next-line no-nested-ternary
+  typeof console !== 'undefined'
+    ? console
+    : typeof print === 'function'
+    ? // Make a good-enough console for eshost.
+      (p => freeze({ debug: p, log: p, info: p, warn: p, error: p }))(
```
erights, line 26: Is a console with only these methods good enough for eshost? Is eshost likely to be the only case this code needs to accommodate?

erights (approved):
LGTM, thanks!

Issue #1772, opened 2023-09-17
Comprehensive `unprotected` interface guards


michaelfig:
## What is the Problem Being Solved?

#1725 is one way to provide the ability to opt out of strict Exo object method guards.  It would be nice for even finer granularity to be able to express the code's intentions more accurately.

## Description of the Design

As a jumping-off point for better designs, here is a (probably suboptimal, may not be easily implemented), but conceptually simple one:

1. For convenience of a library user, provide an "unprotected: true" Zone option that does not enforce any exo interfaces for exos created in that Zone.
2. For patterns, method guards, and interface guards, provide an operator that designates any such context as "unprotected", such as:

```js
// Return a pattern of the same type as T but always matches its value, regardless of passability.
M.unprotected(pattern: T);
// Return a method guard that passes through all values, uninspected.
M.unprotected.call(...args).optional(...opts).rest(dots).returns(ret);
// For every method in methodGuards, mark that method's arguments and return type as unprotected.
M.unprotected.interface('NameI', { ...methodGuards }, flags);
```

As a possible enhancement, do the work so that individual `M.unprotected(patt)` annotations can be used to mark specific arguments (or the return value) as unprotected.

## Security Considerations

The use of `unprotected` needs to be clearly visible to prevent accidentally overlooking the absence of guards for a given method or argument.

## Scaling Considerations

May somewhat complicate the implementation, but may also save time checking `passStyleOf`.

## Test Plan

n/a

## Upgrade Considerations

n/a

PR #1831, opened 2023-10-12
feat: support `M.raw()` in method guards

michaelfig:
closes: #1725 closes: #1772
refs: #1724

## Description

Complete the #1724 experiment, as well as using the `defaultGuards` option (originally discussed in https://github.com/endojs/endo/pull/1725#discussion_r1296249277) instead of boolean flags for `sloppy` and `raw`.

### Security Considerations

The new `M.raw()` guard is only honoured by arguments and return values in `defendSyncMethod`, so it should not leak elsewhere as a pattern.

### Scaling Considerations

Some additional overhead in redacting raw arguments (in `defendSyncArgs`), but work is minimized by calculating relevant data structures while wrapping methods before any exos are created (in `defendSyncMethod`).

### Upgrade Considerations

Backward compatibility is implemented for interfaceGuards created using the legacy `sloppy` parameter.


packages/exo/src/exo-tools.js
```diff
@@ -24,41 +26,131 @@ const { quote: q, Fail } = assert;
...
 /**
  * @param {Passable[]} syncArgs
- * @param {MethodGuardPayload} methodGuardPayload
+ * @param {object} argConfig
+ * @param {number} argConfig.declaredLen
+ * @param {boolean} argConfig.hasRestArgGuard
+ * @param {boolean} argConfig.needRedaction
+ * @param {Pattern} argConfig.paramsPattern
+ * @param {boolean} argConfig.rawRestArgs
+ * @param {number[]} argConfig.redactedIndices
```
erights, line 54: If you pull this out into a separate ArgConfig (or ArgsConfig or RedactConfig) type, then you could add a 
```
@returns {ArgConfig)
```
to `redactRawValues` below, which I think helps readability.

packages/exo/src/exo-tools.js
```diff
@@ -67,16 +159,15 @@ const defendSyncArgs = (syncArgs, methodGuardPayload, label = undefined) => {
  */
 const defendSyncMethod = (method, methodGuardPayload, label) => {
   const { returnGuard } = methodGuardPayload;
+  const isRawReturn = isRawValueGuard(returnGuard);
+  const redactConfig = redactRawValues(methodGuardPayload);
```
erights, line 163: Since you get the `returnGuard` from the `methodGuardPayload` anyway, and since you call the value it returns `redactConfig`, does it read slightly better to move line 162 into `redactRawValues` and include `isRawReturn` as a property of the `redactConfig`? (If so, then for my type name suggestions above, only consider `RedactConfig`.)

erights: Oh, I see. Not as much nicer as I thought, because the return value processing must still be separate here. Nevermind. (But left original comment as record)

packages/exo/src/exo-tools.js
```diff
@@ -184,13 +275,15 @@ const defendMethod = (method, methodGuard, label) => {
  * @param {CallableFunction} behaviorMethod
  * @param {boolean} [thisfulMethods]
  * @param {MethodGuard} [methodGuard]
+ * @param {import('@endo/patterns').DefaultGuardType} [defaultGuards]
```
erights, line 278: We have multiple styles on this and have not worked towards convergence. I'll just mention that I prefer to pull all such `{import('...` into separate typedefs at the top of the file. YMMV.

packages/exo/src/exo-tools.js
```diff
@@ -226,12 +319,23 @@ const bindMethod = (
...
         },
       };
+  if (!methodGuard && thisfulMethods) {
+    switch (defaultGuards) {
+      case 'never':
+      case 'passable':
+        methodGuard = PassableMethodGuard;
+        break;
+      case 'raw':
+        methodGuard = RawMethodGuard;
```
erights, line 329: Nice!

packages/exo/test/test-heap-classes.js
```diff
@@ -203,6 +203,53 @@ test('sloppy option', t => {
   );
 });
 
+const RawGreeterI = M.interface('greeter', {}, { defaultGuards: 'raw' });
```
erights, line 206: Please also test `defaultGuards: 'passable'`

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1788,25 +1792,52 @@ const makeAwaitArgGuard = argPattern => {
...
 };
 
-const PatternListShape = M.arrayOf(M.pattern());
+// M.rawValue()
+
+// TODO does not need to be a singleton, and would not be if we added a
+// parameter, like a description string.
+/** @type {RawValueGuard} */
+const TheRawValueGuard = harden({
+  klass: 'rawValueGuard',
```
erights, line 1801: We have completely moved away from `klass: '...',`. Please follow the new coding pattern. The hazard with the `klass: '...'` coding pattern is that if it happens to appear in another position where a pattern is expected, it will be interpreted as a key pattern that matches only itself. This unsoundness was actually the unsoundness flaw that pushed us away from `klass:`.

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1788,25 +1792,52 @@ const makeAwaitArgGuard = argPattern => {
...
+export const assertRawValueGuard = specimen =>
+  mustMatch(specimen, RawValueGuardShape, 'rawValueGuard');
+
+/**
+ * @returns {RawValueGuard}
+ */
+const makeRawValueGuard = () => TheRawValueGuard;
+
+// M.call(...)
+// M.callWhen(...)
```
erights, line 1818: Is the appearance of these two comment lines here a copy-paste error?

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1788,25 +1792,52 @@ const makeAwaitArgGuard = argPattern => {
...
 const SyncMethodGuardPayloadShape = harden({
   callKind: 'sync',
-  argGuards: PatternListShape,
-  optionalArgGuards: M.opt(PatternListShape),
-  restArgGuard: M.opt(M.pattern()),
-  returnGuard: M.pattern(),
+  argGuards: SyncValueGuardListShape,
+  optionalArgGuards: M.opt(SyncValueGuardListShape),
+  restArgGuard: M.opt(SyncValueGuardShape),
+  returnGuard: SyncValueGuardShape,
```
erights, line 1832: Ironically, it is now good that we failed to rename these properties from `*Guards` to `*Patterns`!

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1886,7 +1917,7 @@ const InterfaceGuardPayloadShape = M.splitRecord(
   {
     interfaceName: M.string(),
     methodGuards: M.recordOf(M.string(), MethodGuardShape),
-    sloppy: M.boolean(),
+    defaultGuards: M.or('never', 'passable', 'raw'),
```
erights, line 1920: Since this property is new, consider whether it should be moved into the optional part of this `splitRecord` pattern. But if you are confident we don't need to, then don't.

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1940,11 +1971,12 @@ harden(getInterfaceMethodKeys);
  * @template {Record<PropertyKey, MethodGuard>} [M = Record<PropertyKey, MethodGuard>]
  * @param {string} interfaceName
  * @param {M} methodGuards
- * @param {{ sloppy?: boolean }} [options]
+ * @param {{ sloppy?: boolean, defaultGuards?: import('../types.js').DefaultGuardType }} [options]
  * @returns {InterfaceGuard<M>}
  */
 const makeInterfaceGuard = (interfaceName, methodGuards, options = {}) => {
-  const { sloppy = false } = options;
+  const { sloppy = false, defaultGuards = sloppy ? 'passable' : 'never' } =
```
erights, line 1978: Nice use of JS's guarantee of left-to-right evaluation of param default values.

packages/patterns/src/types.js
```diff
@@ -456,36 +456,75 @@ export {};
...
 
+/**
+ * @typedef {'never' | 'passable' | 'raw'} DefaultGuardType
+ */
+
+/**
+ * @typedef {<M extends Record<PropertyKey, MethodGuard>>(
+ *   interfaceName: string,
+ *   methodGuards: M,
+ *   options: {defaultGuards?: 'never', sloppy?: false }) => InterfaceGuard<M>
```
erights, line 467: Should this be

```suggestion
 *   options?: {defaultGuards?: 'never', sloppy?: false }) => InterfaceGuard<M>
```

packages/patterns/src/types.js
```diff
@@ -456,36 +456,75 @@ export {};
...
+ * is itself a `CopyRecord`. If used nested within a pattern, rather than
+ * at top level, it may be mistaken for a CopyRecord pattern that would match
+ * only a specimen shaped like an `AwaitArgGuard`.
+ *
+ * @property {(() => RawValueGuard)} rawValue
+ * In parameter position, pass this argument through without any checking.
+ * In rest position, pass the rest of the arguments through without any checking.
+ * In return position, return the result without any checking.
+ *
+ * HAZARD: Until https://github.com/endojs/endo/pull/1712 a `RawValueGuard`
```
erights, line 524: Exactly. https://github.com/endojs/endo/pull/1712 has since been merged, so you should avoid reintroducing this flaw.

packages/patterns/src/types.js
```diff
@@ -590,4 +630,17 @@ export {};
  * @typedef {CopyTagged<'guard:awaitArgGuard', AwaitArgGuardPayload>} AwaitArgGuard
  */
 
-/** @typedef {AwaitArgGuard | Pattern} ArgGuard */
+/**
+ * @typedef {{
+ *   klass: 'rawValueGuard'
```
erights, line 635: Fix

erights (changes requested):
Looks great! But the `klass: '...'` issue must first be fixed. See the `AwaitArgGuard` as an example guide.

packages/exo/test/test-heap-classes.js
```diff
@@ -203,6 +203,53 @@ test('sloppy option', t => {
...
+    sayHello(mutable) {
+      mutable.x = 3;
+      return 'hello';
+    },
+  });
+  t.deepEqual(greeter[GET_INTERFACE_GUARD](), RawGreeterI);
+  testGreeter(t, greeter, 'raw defaultGuards');
+
+  const Greeter2I = M.interface('greeter2', {
+    sayHello: M.call(M.rawValue()).returns(M.string()),
```
turadg, line 227: I'm new to the motivating issues and don't grok them yet. "It would be nice for even finer granularity" why?

Here it's not clear to me how the behavior of `M.rawValue()` is different than `M.any()`. Please include a test contrasting them specifically.

michaelfig: I'll include more explicit tests.

As to the difference: matching `M.any()` automatically hardens its input, then enforces it is Passable (e.g. is compatible with inter-vat marshalling).  `M.rawValue()` avoids that processing: it allows designated arguments or return values through verbatim.

michaelfig: I'll include more explicit tests.

As to the difference: matching `M.any()` automatically hardens its input, then enforces it is Passable (e.g. is compatible with inter-vat marshalling).  `M.rawValue()` avoids that processing: it allows designated arguments or return values through verbatim.

turadg: Ah, I didn't realize:
- the matchers do modify the argument (hardening)
- M.any() doesn't really alllow _any_; it requires Passable. I think it's a good name but maybe some jsdoc is in order: "any Passable"

Given that the other matchers are for values, why should "Value" be included in the method name? iow, why not `M.raw()` ?

packages/exo/src/exo-tools.js
michaelfig: I was one of the instigators of the "separate typedefs for import" pattern.  Since then, I've learned that doing that doesn't really import the types, it creates a JSDoc-less structural copy of the imported type.  So, I now prefer import-on-use, which directly references both the JSDoc and the type.

So, I think I'll leave this one intact.

packages/patterns/src/patterns/patternMatchers.js
michaelfig: No, just an artifact of the diff algorithm finding the inserted `M.rawValue()` lines.

packages/patterns/src/patterns/patternMatchers.js
michaelfig: Thanks, I reintroduced sloppy and moved defaultGuards down to remain compatible.

packages/patterns/src/types.js
michaelfig: Actually, no.  These two overloads are only used if the options bag is specified.

packages/exo/test/test-heap-classes.js
michaelfig: "even finer granularity" is what this PR provides.  The issue that mentioned the need was addressed at an older PR that only did interface-granularity.

`M.rawValue()` doesn't do any processing of the designated argument or return value.  OTOH, `M.any()` automatically hardens the designated value, and asserts that it is Passable (i.e. can be marshalled using `@endo/marshal`).

I'll add some more specific tests.

turadg: >M.rawValue() doesn't do any processing of the designated argument or return value. OTOH, M.any() automatically hardens the designated value, and asserts that it is Passable (i.e. can be marshalled using @endo/marshal).

I got that in the previous message. 

When I said "maybe some jsdos is in order: 'any Passable'" I meant to make it more clear that M.any doesn't match "any value" but any _passable_. It turns out that's already in the jsdoc,
https://github.com/endojs/endo/blob/54ff59f9e44a19a4ef7c9d3f72b01df8c9059626/packages/patterns/src/types.js#L239-L241

However, none of those JSDocs indicate that the argument will be hardened. That seems important to say. Other APIs complain when the argument isn't already hardened so this seems inconsistent and likely to surprise people. Not something that has to be solved in this PR, of course.

I don't see an answer in your reply to this question:
> Given that the other matchers are for values, why should "Value" be included in the method name? iow, why not `M.raw()` ?

michaelfig: > I got that in the previous message.

Yeah, as discussed offline I thought my phone lost my first comment (when it actually posted it twice), so from-memory recapitulation probably obscured what I was trying to say.

> However, none of those JSDocs indicate that the argument will be hardened. That seems important to say. Other APIs complain when the argument isn't already hardened so this seems inconsistent and likely to surprise people. Not something that has to be solved in this PR, of course.

What I didn't state correctly is that `M.any()` doesn't actually do the hardening.  It's the default behaviour of `M.call()` or `M.callWhen()` that hardens arguments and return values before matching patterns, and `M.rawValue()` opts out of this behaviour.  I will add some JSDoc to `.call` and `.callWhen` to make this clearer.
 
> > Given that the other matchers are for values, why should "Value" be included in the method name? iow, why not `M.raw()` ?

I spoke with @erights, and we agreed that `M.raw()` would be better.  I'll make that change.

michaelfig:
All comments are addressed, and CI is passing.  Ready for rereview!

packages/exo/src/exo-makers.js
```diff
@@ -130,9 +149,7 @@ export const defineExoClass = (
     if (finish) {
       finish(context);
     }
-    return /** @type {M & import('@endo/eventual-send').RemotableBrand<{}, M>} */ (
-      self
-    );
+    return self;
```
turadg, line 152: is this still typed? consider adding a type regression test

packages/exo/src/exo-makers.js
```diff
@@ -222,7 +238,6 @@ harden(defineExoClassKit);
  * }> | undefined} interfaceGuard CAVEAT: static typing does not yet support `callWhen` transformation
  * @param {T} methods
  * @param {FarClassOptions<ClassContext<{},T>>} [options]
- * @returns {T & import('@endo/eventual-send').RemotableBrand<{}, T>}
```
turadg, line 225: while this may not be necessary for the type checker or IDE, it's helpful for reading the code. Is this inaccurate now or just deemed superfluous?

packages/exo/src/exo-tools.js
```diff
@@ -6,59 +6,156 @@ import {
...
 
 const { quote: q, Fail } = assert;
 const { apply, ownKeys } = Reflect;
 const { defineProperties, fromEntries } = Object;
 
+/**
+ * A method guard, for inclusion in an interface guard, that does not
+ * enforce any constraints of incoming arguments or return results.
+ */
+const RawMethodGuard = M.call().rest(M.raw()).returns(M.raw());
```
turadg, line 34: not much of a _guard_  ;-)

consider `RawMethodCall` or `UnguardedCall`

packages/exo/src/exo-tools.js
```diff
@@ -6,59 +6,156 @@ import {
...
+const redactRawArgs = methodGuardPayload => {
+  const {
+    argGuards,
+    optionalArgGuards = [],
+    restArgGuard,
+  } = methodGuardPayload;
+
+  const rawArgGuards = [...argGuards, ...optionalArgGuards];
+
+  let needRedaction = false;
```
turadg, line 117: this state is implied by `redactedIndices`. consider omitting this and down below returning,
```js
    needsRedaction: redactedIndicies.length > 0,
```
Though since it's implied by the length of that array, why pass it around at all?

packages/exo/src/exo-tools.js
```diff
@@ -67,16 +164,15 @@ const defendSyncArgs = (syncArgs, methodGuardPayload, label = undefined) => {
...
-      const realArgs = defendSyncArgs(
-        harden(syncArgs),
-        methodGuardPayload,
-        label,
-      );
+      // Only harden args and return value if not dealing with a raw value guard.
+      const realArgs = defendSyncArgs(syncArgs, redactConfig, label);
       const result = apply(method, this, realArgs);
-      mustMatch(harden(result), returnGuard, `${label}: result`);
+      isRawReturn || mustMatch(harden(result), returnGuard, `${label}: result`);
```
turadg, line 175: we usually use the `||` like this for an assertion, suggesting that left expression is supposed to be true. that's not the case here so consider,
```suggestion
      if (!isRawReturn) {
        mustMatch(harden(result), returnGuard, `${label}: result`);
      }
```

packages/exo/src/exo-makers.js
michaelfig: Yes, still typed.  The types of these `define*` methods are exercised in `test-heap-classes.js`.

packages/exo/src/exo-makers.js
michaelfig: Okay.  It was both inaccurate and deemed superfluous.  I've added correct `@returns` types instead.

packages/exo/src/exo-tools.js
michaelfig: It is still a `MethodGuard`, so I don't want to obscure that with a different name.

packages/exo/src/exo-tools.js
michaelfig: Omitted and using `redactedIndices` in its place.

packages/exo/src/exo-tools.js
```diff
@@ -6,59 +6,150 @@ import {
...
+    rawRestArgs,
+    redactedIndices,
+  } = redactConfig;
+
+  // Only manipulate the original args if our params pattern didn't contain any
+  // raw values.
+  let redactedArgs = syncArgs;
+  if (rawRestArgs) {
+    // Don't harden the rest args.
+    redactedArgs = syncArgs.slice(0, Math.min(syncArgs.length, declaredLen));
```
gibson042, line 77: I don't think this `Math.min` is doing anything; [`slice`](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype.slice) already clamps its second argument to the length of the array.
```suggestion
    redactedArgs = syncArgs.slice(0, declaredLen);
```

packages/exo/src/exo-tools.js
```diff
@@ -6,59 +6,150 @@ import {
...
+  }
+
+  for (const i of redactedIndices) {
+    if (i >= redactedArgs.length) {
+      break;
+    }
+    redactedArgs[i] = REDACTED_RAW_ARG;
+  }
+
+  mustMatch(harden(redactedArgs), paramsPattern, label);
```
gibson042, line 91: The `redactedArgs` name seems a little misleading, implying e.g. "those arguments that have been redacted". Maybe instead `matchableArgs`?

Also, I question the value of completely ignoring raw rest args here—it's probably better to preserve their presence:
```js
  // Use syncArgs if possible, but copy it when necessary to implement redactions.
  let matchableArgs = syncArgs;
  if (rawRestArgs && syncArgs.length > declaredLen) {
    const restLen = syncArgs.length - declaredLen;
    const redactedRest = Array(restLen).fill(REDACTED_RAW_ARG);
    matchableArgs = [...syncArgs.slice(0, declaredLen), ...redactedRest];
  } else if (redactedIndices.length > 0 && redactedIndices[0] < syncArgs.length) {
    matchableArgs = [...syncArgs];
  }

  for (const i of redactedIndices) {
    if (i >= matchableArgs.length) {
      break;
    }
    matchableArgs[i] = REDACTED_RAW_ARG;
  }

  mustMatch(harden(redactedArgs), paramsPattern, label);
```

packages/exo/src/exo-tools.js
```diff
@@ -6,59 +6,150 @@ import {
...
-    argGuards.length + (optionalArgGuards ? optionalArgGuards.length : 0);
   if (syncArgs.length <= declaredLen) {
     return syncArgs;
   }
   // Ignore extraneous arguments, as a JS function call would do.
   return syncArgs.slice(0, declaredLen);
 };
 
+/**
+ * @param {MethodGuardPayload} methodGuardPayload
```
gibson042, line 104: The purpose and behavior of this function is non-obvious and could use a description.

packages/exo/src/exo-tools.js
```diff
@@ -6,59 +6,150 @@ import {
...
+      rawArgGuards[i] = REDACTED_RAW_ARG;
+      redactedIndices.push(i);
+    }
+  }
+
+  // Always match the redacted args if there is a raw rest that applies.
+  let rawRestArgGuard = restArgGuard;
+  let rawRestArgs = false;
+  if (isRawGuard(rawRestArgGuard)) {
+    rawRestArgGuard = undefined;
```
gibson042, line 127:
```suggestion
    rawRestArgGuard = REDACTED_RAW_ARG;
```

packages/exo/src/exo-tools.js
```diff
@@ -67,16 +158,17 @@ const defendSyncArgs = (syncArgs, methodGuardPayload, label = undefined) => {
  */
 const defendSyncMethod = (method, methodGuardPayload, label) => {
   const { returnGuard } = methodGuardPayload;
+  const isRawReturn = isRawGuard(returnGuard);
+  const redactConfig = redactRawArgs(methodGuardPayload);
```
gibson042, line 162:
```suggestion
  const redactionConfig = toRedactionConfig(methodGuardPayload);
```

packages/patterns/src/types.js
```diff
@@ -456,36 +456,67 @@ export {};
...
+ * fulfillment does not satisfy the nested pattern, then the call is rejected
+ * without ever calling the raw method.
+ *
+ * Any `AwaitArgGuard` may not appear as a rest pattern or a result pattern,
+ * only a top-level single parameter pattern.
+ *
+ * @property {() => RawGuard} raw
+ * In parameter position, pass this argument through without any hardening or checking.
+ * In rest position, pass the rest of the arguments through without any hardening or checking.
+ * In return position, return the result without any hardening or ßchecking.
```
gibson042, line 519:
```suggestion
 * In return position, return the result without any hardening or checking.
```

gibson042 (commented):
Even reading the linked issues, I'm not clear on the motivation for introducing this non-Passable "raw" argument/return and unknown-method functionality, and am wondering if it might be better to instead establish a convention for e.g. conveying mutable objects via thunks and/or introduce specific configuration for supporting unknown methods. Can you explain the need?

packages/exo/src/exo-tools.js
michaelfig: This needed some finessing, after replacing `undefined` with the only pattern that could possibly work: `M.arrayOf(REDACTED_RAW_ARG)`.

michaelfig:
> Even reading the linked issues, I'm not clear on the motivation for introducing this non-Passable "raw" argument/return 

Exos used to be tightly coupled to `@endo/pass-style`, so that it was impossible to define an exo method that either took a non-Passable as an argument or returned a non-Passable.  The usage I am working towards is an exo-promise that allows `settler.resolve(anyValueAtAll)`.  There is no way to express such a method on an exo class, since the old most lenient argument guard (`M.any()`) was to `harden(anyValueAtAll)` and then assert it was passable.

> and unknown-method functionality,

We already had the unknown-method functionality called `sloppy`, and @erights proposed extending it to allow the specification of sloppy-passable methods, as well as sloppy-raw methods.  I did the implementation in this PR, even though I have no particular use for the functionality at this point, and prefer the "raw" argument/return annotations since they are stricter.

> and am wondering if it might be better to instead establish a convention

I want to be able to create exo objects or facets whose methods can be just as flexible as a vanilla JS function, but have that be an opt-out so that the Exo model of guards and patterns is the default behaviour.

> for e.g. conveying mutable objects via thinks

Funny you should mention that.  Thunks are non-passable, too!

> and/or introduce specific configuration for supporting unknown methods.

The specific configuration is to explicitly specify `M.interface('Foo', { meth: M.call().rest(M.raw()).returns(M.raw()) })`.  Arguably, we could lean in further and do something like `M.interface('Foo', { meth: M.raw() })` too, but that starts to blur  the distinction between MethodGuard and SyncArgGuard.

> Can you explain the need?

I don't want to make breaking design changes to contracts or libraries that I'm converting to be Exo-objects.  Especially when so much of their behaviour fits into the Exo-object model already.  `M.raw()` makes that possible.  The unknown method support is tangential but related, and I don't really care as much about it.

packages/patterns/src/patterns/patternMatchers.js
erights: Actually, you introduced `defaultGuards` in the mandatory section. If we get away with that, then we know this new code never encounters any `InterfaceGuardPayload`s created by code before this PR. That's great! If we believe that, then I think you can simply delete the optional `sloppy:` below. Otherwise, I think both `sloppy:` and `defaultGuards:` need to go into the optional section.

packages/exo/src/exo-makers.js
```diff
@@ -3,7 +3,14 @@
...
 
-import { defendPrototype, defendPrototypeKit } from './exo-tools.js';
+import {
+  GET_INTERFACE_GUARD,
+  defendPrototype,
+  defendPrototypeKit,
+} from './exo-tools.js';
+
+// Used in typing.
+GET_INTERFACE_GUARD;
```
erights: What does this no-effect statement do?

packages/exo/src/exo-makers.js
```diff
@@ -186,8 +218,8 @@ export const defineExoClassKit = (
     // Be careful not to freeze the state record
     const state = seal(init(...args));
     // Don't freeze context until we add facets
-    /** @type {KitContext<ReturnType<I>,F>} */
-    const context = { state, facets: {} };
+    /** @type {{ state: ReturnType<I>, facets: unknown }} */
+    const context = { state, facets: null };
```
erights, line 208: Why is null here correct? The original
```suggestion
    const context = { state, facets: {} };
```
seems more correct.

erights (approved):
LGTM!

But especially curious about `facets: null`. Am I missing something?

gibson042:
> > Even reading the linked issues, I'm not clear on the motivation for introducing this non-Passable "raw" argument/return
> 
> Exos used to be tightly coupled to `@endo/pass-style`, so that it was impossible to define an exo method that either took a non-Passable as an argument or returned a non-Passable. The usage I am working towards is an exo-promise that allows `settler.resolve(anyValueAtAll)`. There is no way to express such a method on an exo class, since the old most lenient argument guard (`M.any()`) was to `harden(anyValueAtAll)` and then assert it was passable.

I think that just pushes the question back a level... what is the underlying motivation for non-Passable promise resolution? `M.raw()` seems like a reasonable way to get a pattern representing literally any value, and `defaultGuards` "passable" vs. "raw" seems good for its propagation through `defineExo{Class,ClassKit}`, but I'm still wondering why we need to do anything at all here.

> > for e.g. conveying mutable objects via thinks
> 
> Funny you should mention that. Thunks are non-passable, too!

Are they?
```js
import "@endo/init";
import { Far } from "@endo/far";
import { M, matches } from "@endo/patterns";

const fn = Far("fn", () => "returned");
fn();
// => "returned"
matches(fn, M.remotable());
// => true
```

> I don't want to make breaking design changes to contracts or libraries that I'm converting to be Exo-objects.

Ah, this seems like it's heading in the direction I care about. What is being converted to exo, what is the benefit, and how do they currently interact with non-Passable values?

packages/patterns/src/types.js
```diff
@@ -456,36 +456,67 @@ export {};
...
+ * parameter position of an `M.callWhen`.
+ * `M.await(M.nat())`, for example, with `await` the corresponding argument,
+ * check that the fulfillment of the `await` satisfies the `M.nat()`
+ * pattern, and only then proceed to call the raw method with that fulfillment.
+ * If the argument already passes the `M.nat()` pattern, then the result of
+ * `await`ing it will still pass, and the `M.callWhen` will still delay the
+ * raw method call to a future turn.
+ * If the argument is a promise that rejects rather than fulfills, or if its
+ * fulfillment does not satisfy the nested pattern, then the call is rejected
+ * without ever calling the raw method.
```
gibson042, line 511: :+1: :+1: for adding this.
```suggestion
 * Guard a positional parameter in `M.callWhen`, awaiting it and matching its
 * fulfillment against the provided pattern.
 * For example, `M.await(M.nat())` will await the first argument, check that its
 * fulfillment satisfies `M.nat()`, and only then call the guarded method with
 * that fulfillment. If the argument is a non-promise value that already
 * satisfies `M.nat()`, then the result of `await`ing it will still pass, and
 * `M.callWhen` will still delay the guarded method call to a future turn.
 * If the argument is a promise that rejects rather than fulfills, or if its
 * fulfillment does not satisfy the nested pattern, then the call is rejected
 * without ever invoking the guarded method.
```

packages/patterns/src/types.js
```diff
@@ -546,8 +580,12 @@ export {};
...
  * }
  * ```
- * @property {(rArgGuard: Pattern) => MethodGuardMaker2} rest
- * @property {(returnGuard?: Pattern) => MethodGuard} returns
+ * @property {(restArgGuard: SyncValueGuard) => MethodGuardMaker2} rest
+ * If the rest argument guard is not `M.raw()`, all rest arguments are
+ * automatically hardened and must be Passable.
+ * @property {(returnGuard?: SyncValueGuard) => MethodGuard} returns
+ * If the return guard is not `M.raw()`, the return value is automatically
+ * hardened and must be Passable.
```
gibson042, line 588: Is it possible to define these types in a way that doesn't require so much repetition?
```js
/**
 * @typedef {MethodGuardOptMaker & MethodGuardRestMaker & MethodGuardFinalizer} MethodGuardMaker
 * A method name and parameter/return signature like:
 * ```js
 *   foo(a, b, c = d, ...e) => f
 * ```
 * should be guarded by something like:
 * ```js
 * {
 *   ...otherMethodGuards,
 *   foo: M.call(AShape, BShape).optional(CShape).rest(EShape).returns(FShape),
 * }
 * ```
 */

/**
 * @typedef {object} MethodGuardOptMaker
 * @property {(...optArgGuards: ArgGuard[]) => (MethodGuardRestMaker & MethodGuardFinalizer)} optional
 * Optional arguments not guarded with `M.raw()` are automatically hardened and
 * must be Passable.
 */

/**
 * @typedef {object} MethodGuardRestMaker
 * @property {(restArgGuard: SyncValueGuard) => MethodGuardFinalizer} rest
 * If the rest argument guard is not `M.raw()`, all rest arguments are
 * automatically hardened and must be Passable.
 */

/**
 * @typedef {object} MethodGuardFinalizer
 * @property {(returnGuard?: SyncValueGuard) => MethodGuard} returns
 * If the return guard is not `M.raw()`, the return value is automatically
 * hardened and must be Passable.
 */
```

gibson042 (approved):

packages/patterns/src/types.js
```diff
@@ -590,4 +630,14 @@ export {};
  * @typedef {CopyTagged<'guard:awaitArgGuard', AwaitArgGuardPayload>} AwaitArgGuard
  */
 
-/** @typedef {AwaitArgGuard | Pattern} ArgGuard */
+/**
+ * @typedef {{}} RawGuardPayload
```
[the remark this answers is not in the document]
michaelfig: This time, I'm specifying that `RawGuardPayload` is the empty object, not just anything of type object.

packages/exo/src/exo-tools.js
michaelfig: If we're going through the effort to create rest arguments that are each `REDACTED_RAW_ARG`, I'd rather do the stricter `arrayOf` pattern.  Without being strict, I see no compelling reason to do anything other than just drop raw rest arguments rather than doing this whole match-but-bypass dance.

michaelfig:
> ```js
> const fn = Far("fn", () => "returned");
> fn();
> // => "returned"
> matches(fn, M.remotable());
> // => true
> ```

Oh, I misspoke when I said "thunks are non-passable".  I meant to say "passable thunks are not consistently supported in SwingSet and other random parts of our system, so I wouldn't count on using them for anything important until they have had significantly more testing."

> Ah, this seems like it's heading in the direction I care about. What is being converted to exo, what is the benefit, and how do they currently interact with non-Passable values?

IBC/Network vats and Pegasus are being made durable and upgradable.  I'm trying to do a minimal implementation of upgrade-tolerant promise-like remotables, and that's the last in a long line of trying to do promise-like remotables and bashing my head into the ceiling imposed by the passable restrictions.  This PR is the result of trying to raise that ceiling so that I can use Zones and Exo-objects for durability as advertised.

PR #1832, opened 2023-10-12
fix(daemon): Strengthen type checks

kriskowal:
Evidently type checks have drifted because `@ts-check` comments are for some reason not redundant with `"checkJs": true` as advertised. This change gets ts-check to pass for all the daemon sources.

### DIFF
--- a/packages/daemon/src/daemon-node-powers.js
+++ b/packages/daemon/src/daemon-node-powers.js
@@ -223,6 +223,12 @@ export const makeNetworkPowers = ({ http, ws, net }) => {
     return readFrom;
   };
 
+  /**
+   * @param {object} args
+   * @param {number} args.port
+   * @param {string} [args.host]
+   * @param {Promise<never>} args.cancelled
+   */
   const servePort = async ({ port, host = '0.0.0.0', cancelled }) =>
     serveListener(
       server =>
@@ -232,6 +238,11 @@ export const makeNetworkPowers = ({ http, ws, net }) => {
       cancelled,
     );
 
+  /**
+   * @param {object} args
+   * @param {string} args.path
+   * @param {Promise<never>} args.cancelled
+   */
   const servePath = async ({ path, cancelled }) =>
     serveListener(
       server =>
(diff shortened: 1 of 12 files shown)


packages/stream/types.d.ts
```diff
@@ -40,8 +47,8 @@ export declare function makeStream<
   TReadReturn = undefined,
   TWriteReturn = undefined,
 >(
-  acks: AsyncQueue<IteratorResult<TRead, TReadReturn>>,
-  data: AsyncQueue<IteratorResult<TWrite, TWriteReturn>>,
+  acks: AsyncSpring<IteratorResult<TRead, TReadReturn>>,
+  data: AsyncSink<IteratorResult<TWrite, TWriteReturn>>,
```
kriskowal, line 51: I imagine @mhofman will be interested in this relaxation of the signature of `makeStream`. This is a remnant of the pubsub work. Although we’re not pursuing that variety of pubsub, this does make the type signature of `makeStream` no more specific than necessary, which I think is good POLA.

packages/daemon/src/daemon-node-powers.js
```diff
@@ -232,6 +238,11 @@ export const makeNetworkPowers = ({ http, ws, net }) => {
       cancelled,
     );
 
+  /**
+   * @param {object} args
+   * @param {string} args.path
+   * @param {Promise<unknown>} args.cancelled
```
kumavis, line 244: inconsistent types (see comment above)
here `Promise<unknown>`

packages/daemon/src/daemon-node-powers.js
```diff
@@ -223,6 +223,12 @@ export const makeNetworkPowers = ({ http, ws, net }) => {
     return readFrom;
   };
 
+  /**
+   * @param {object} args
+   * @param {number} args.port
+   * @param {string} [args.host]
+   * @param {Promise<never>} args.cancelled
```
kumavis, line 230: inconsistent types (see comment below)
here `Promise<never>`

kumavis (approved):
:tada:

PR #1833, opened 2023-10-13
fix(compartment-mapper): fix for bundling of appended cjs exports

kumavis:
bug fix for bundler for supporting the following commonjs exports patterns 
```js
module.exports.abc = 123
exports.xyz = 456
```

Originally tried to fix it with https://github.com/endojs/endo/commit/7daba15b88752898ab1f9a1895d5cf479ea3575f
but this was insufficient, need to add missing cells and update the `*` cell

not sure if this is the correct change or just a bandaid but it got tests passing

### DIFF
--- a/packages/compartment-mapper/src/bundle-cjs.js
+++ b/packages/compartment-mapper/src/bundle-cjs.js
@@ -5,7 +5,9 @@ const exportsCellRecord = exportsList =>
   ''.concat(
     ...exportsList.map(
       exportName => `\
-      ${exportName}: cell(${q(exportName)}),
+      ${q(exportName)}: cell(${q(exportName)}${
+        exportName !== 'default' ? '' : `, {}`
+      }),
 `,
     ),
   );
@@ -14,15 +16,28 @@ const exportsCellRecord = exportsList =>
 const runtime = function wrapCjsFunctor(num) {
   /* eslint-disable no-undef */
   return ({ imports = {} }) => {
+    const moduleCells = cells[num];
     const cModule = Object.freeze(
-      Object.defineProperty({}, 'exports', cells[num].default),
+      Object.defineProperty({}, 'exports', moduleCells.default),
     );
     // TODO: specifier not found handling
     const requireImpl = specifier => cells[imports[specifier]].default.get();
     functors[num](Object.freeze(requireImpl), cModule.exports, cModule);
-    Object.keys(cells[num])
+    // Update all named cells from module.exports.
+    Object.keys(moduleCells)
       .filter(k => k !== 'default' && k !== '*')
-      .map(k => cells[num][k].set(cModule.exports[k]));
+      .map(k => moduleCells[k].set(cModule.exports[k]));
+    // Add new named cells from module.exports.
+    Object.keys(cModule.exports)
+      .filter(k => k !== 'default' && k !== '*')
+      .filter(k => moduleCells[k] === undefined)
+      .map(k => (moduleCells[k] = cell(k, cModule.exports[k])));
+    // Update the star cell from all cells.
+    const starExports = Object.create(null);
+    Object.keys(moduleCells)
+      .filter(k => k !== '*')
+      .map(k => Object.defineProperty(starExports, k, moduleCells[k]));
+    moduleCells['*'].set(Object.freeze(starExports));
   };
   /* eslint-enable no-undef */
 }.toString();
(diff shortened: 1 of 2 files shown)


packages/compartment-mapper/test/test-bundle.js
```diff
@@ -77,3 +77,49 @@ test('equivalent archive behaves the same as bundle', async t => {
...
+    import.meta.url,
+  ).toString();
+
+  const bundle = await makeBundle(read, cjsFixture);
+  fs.writeFileSync('bundle.js', bundle);
+  const compartment = new Compartment();
+  let namespace;
+  t.notThrows(() => {
+    namespace = compartment.evaluate(bundle);
+  });
```
kriskowal, line 111: I suspect that awaiting here is sufficient since the intent is to stop the test if it throws. The subsequent errors will be less direct than failing here.

packages/compartment-mapper/test/test-bundle.js
```diff
@@ -77,3 +77,49 @@ test('equivalent archive behaves the same as bundle', async t => {
...
+});
+
+test('bundle cjs-compat default-difficulties', async t => {
+  const cjsFixture = new URL(
+    'fixtures-cjs-compat/node_modules/default-difficulties/index.mjs',
+    import.meta.url,
+  ).toString();
+
+  const bundle = await makeBundle(read, cjsFixture);
+  fs.writeFileSync('bundle.js', bundle);
```
kriskowal, line 106: As a matter of tending toward web-portable ocap style, you could `await writePowers.write` here.

kriskowal (commented):
I defer to @naugtur and I see nothing to block this from landing. Consider adding a note to `NEWS.md`.

packages/compartment-mapper/test/test-bundle.js
kumavis: oh this was not supposed to make it in, thanks for catching it

packages/compartment-mapper/test/test-bundle.js
kumavis: `compartment.evaluate(bundle)` returns synchronously and does not return a promise, so i'm confused about the purpose of the `await`

kriskowal: my bad, just call :-)

packages/compartment-mapper/src/bundle-cjs.js
```diff
@@ -14,15 +16,28 @@ const exportsCellRecord = exportsList =>
 const runtime = function wrapCjsFunctor(num) {
   /* eslint-disable no-undef */
   return ({ imports = {} }) => {
+    const moduleCells = cells[num];
```
naugtur, line 19: improvement for bundle size when minified :+1:

packages/compartment-mapper/src/bundle-cjs.js
```diff
@@ -14,15 +16,28 @@ const exportsCellRecord = exportsList =>
...
+    Object.keys(moduleCells)
       .filter(k => k !== 'default' && k !== '*')
-      .map(k => cells[num][k].set(cModule.exports[k]));
+      .map(k => moduleCells[k].set(cModule.exports[k]));
+    // Add new named cells from module.exports.
+    Object.keys(cModule.exports)
+      .filter(k => k !== 'default' && k !== '*')
+      .filter(k => moduleCells[k] === undefined)
+      .map(k => (moduleCells[k] = cell(k, cModule.exports[k])));
+    // Update the star cell from all cells.
```
naugtur, line 35: I mentioned earlier that in regular endo apps the `*` behavior is implemented elsewhere, but that elsewhere doesn't exist in bundler, so I think this is the right place to do this.

packages/compartment-mapper/test/test-bundle.js
```diff
@@ -77,3 +77,44 @@ test('equivalent archive behaves the same as bundle', async t => {
...
+  ).toString();
+
+  const bundle = await makeBundle(read, cjsFixture);
+  const compartment = new Compartment();
+  const { results } = compartment.evaluate(bundle);
+  const resultExports = results.map(result => {
+    return Object.keys(result).sort();
+  });
+  t.deepEqual(resultExports, [
+    ['default', 'even'],
```
naugtur, line 112: if required in cjs they should not expose the default unless it was explicitly assigned by the module. I think that's true with the current implementation. Not sure if there's a fixture for that, but there might already be.

naugtur (approved):

packages/compartment-mapper/test/test-bundle.js
kumavis: im surprised `default` is not like a Symbol or something, but I dont know the desired behavior for compat

PR #1835, opened 2023-10-18
fix(daemon): Various bug fixes

kriskowal:
These are various bug fixes. There was a problem for services reporting that they started too early, due to a type mismatch for the `started` promise. Also, domain sockets evidently have a path length limit, so certain configurations will fail. The change here makes those errors more transparent to the user.


kumavis (approved):
gr00vey

PR #1836, opened 2023-10-21
feat(daemon): Terminator DAG

kriskowal:
This stack of changes gives us a `terminate(petName)` method on the Endo Host Powers object. This will cause the daemon to forget the current incarnation of that named value and all named values that depend upon it, including parties that  received the value through `request` or `adopt`. This also terminates workers.

The implications of termination are far-reaching and this is only the beginning. New formulas may self-destruct, like a formula that uses a file watcher to invalidate itself when its on-disk dependencies change. Connections may self-destruct if they disconnect. Weblets might auto-reload if their powers are invalidated.

Also, because the formulas form an acyclic DAG, simple ref-counting should suffice for garbage collecting formulas. Formulas can be retained by pet name paths, but also live values, so using a WeakRef in the future should allow us to automatically terminate and collect provably unreachable formulas.

To make this work, the internal memo now envelopes values in a controller object to track the corresponding terminator (synchronously) and the internal and external facets of the formula. This cleans up some side tables previously held for formula-specific internals like the worker bootstrap and the message bus between parties.

### DIFF
--- a/packages/cli/src/endo.js
+++ b/packages/cli/src/endo.js
@@ -481,6 +481,21 @@ export const main = async rawArgs => {
       return mkguest({ name, partyNames });
     });
 
+  program
+    .command('kill <name>')
+    .option(
+      '-a,--as <party>',
+      'Pose as named party (as named by current party)',
+      collect,
+      [],
+    )
+    .description('terminate a value and its deps, recovering resources')
+    .action(async (name, cmd) => {
+      const { as: partyNames } = cmd.opts();
+      const { killCommand } = await import('./kill.js');
+      return killCommand({ name, partyNames });
+    });
+
   const where = program
     .command('where')
     .description('prints paths for state, logs, caches, socket, pids');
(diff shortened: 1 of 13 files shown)


kumavis (approved):

kumavis:
might be nice to be able to review termination dependencies, but not a blocker for this pr

kriskowal:
Building upon this I’ve realized that cancellation, termination, and kill should all just be cancel. Consolidating under a single abstraction will require further work. Notably, terminate() doesn’t yet take a reason argument like cancel(reason), terminated is Promise<void> whereas cancelled is a Promise<never>. `endo cancel` works for me, just as a `Cancel` button in the “familiar“ inventory works.

Viewing termination dependencies is closely related to a discussion I started on Matrix today, but for revealing the dependencies of a formula in a pet name path. I think that will look like:

```console
> endo eval 10 -n ten
10
> endo list META.ten
worker
source
names
values
> endo show META.ten.worker
[alleged Worker]
```

Where META is an indelible reference to the party’s meta-pet-store. This would happen to reflect all the dependencies of a pet named value, all of the objects that would cause this value to get cancelled if they were cancelled. I don’t have a ready idea for inspecting the inverse.

PR #1838, opened 2023-10-26
Fix the mismatch between specifier and resource identifier in attenuators policy AND allow skiping powerless packages

naugtur:
Update:

The problem identified when fixing attenuators policy turns out to not exist when we allow skipping setting empty values for policy resources for packages we want to endow no powers to.  
I decided to solve that right away.

The current state of the PR is a work in progress, but it's only missing some test coverage for error hints and some hint improvements.

----

The issue is a mismatch between a specifier and a resource id AKA cannonical name in:
https://github.com/endojs/endo/blob/master/packages/compartment-mapper/src/policy.js#L158
the right policy value for `@a/b/custom-export` is `@a/b` and I need to add a proper conversion there.


packages/compartment-mapper/src/archive.js
```diff
@@ -145,13 +145,16 @@ const translateCompartmentMap = (compartments, sources, compartmentRenames) => {
...
+            module.compartment !== undefined &&
+            compartmentRenames[module.compartment] !== undefined
+          ) {
             modules[name] = {
               ...module,
               compartment: compartmentRenames[module.compartment],
             };
           } else {
-            modules[name] = module;
+            // If compartment is not defined, the module must not be part of the compartment map.
```
naugtur, line 157: Other parts of endo claim leaving a `compartments[name].modules[name].compartment` undefined is not acceptable. 
I think this is the right way to go about fixing because:
1. validation has been in place to guarantee a compartment map with this issue could not be used
2. even if there's code running afterwards that could populate the field, it has not been the case with the error I saw, so it's definitely possible for it to be overlooked (also, I didn't identify such code at first glance)

kriskowal (approved):
It’s not clear to me at a glance how the change effects the fix described but we can talk that thru at next sync.

naugtur:
> It’s not clear to me at a glance how the change effects the fix described but we can talk that thru at next sync.

The fix for attenuators is to allow importing all into the attenuators compartment instead of trying to resolve a specifier into a policy resource identifier, which is a lot of work and implemented far away :) 

The fix in archive is pretty much independent, but the switch to `packages: 'any'` in attenuators compartment policy caused a pre-existing bug to surface. 
The line I removed and replaced with a comment in archive.js was never reached in testing. Otherwise it would fail, because soon after the compartment map translation is performed, validation rejects a compartment map which has an undefined compartment for a module.
My fixtures for policy testing happen to trigger that error now, while previously the list of modules for the attenuators compartment was trimmed by the policy to not include the one that doesn't get retained in an archive.

I'm looking forward to discussing my assumptions and decisions. Don't mind keeping this open till Wed, but I'm not sure if I'm gonna be able to attend while at NodeConfEU. We'll see.

kriskowal:
@naugtur Good enough for me!

packages/compartment-mapper/src/node-modules.js
```diff
@@ -611,11 +611,6 @@ const translateGraph = (
       },
       policy,
     );
-    // do not include compartments for packages not covered by policy
-    if (policy && !packagePolicy) {
-      // eslint-disable-next-line no-continue
-      continue;
-    }
```
naugtur, line 618: this was a premature optimization, we could introduce a strict mode in the future but for real applications with npm dependencies it doesn't make sense to generate  and maintain policy entries for powerless packages. 

The only alternative approach would be to process the policy and assume empty entries exist for all packages allowed for import. 

Looking at the positive outcomes from this change here, I'd rather avoid optimizing this.

packages/compartment-mapper/src/policy.js
```diff
@@ -113,29 +113,6 @@ export const dependencyAllowedByPolicy = (namingKit, packagePolicy) => {
   return !!policyLookupHelper(packagePolicy, 'packages', canonicalName);
 };
 
-const validateDependencies = (policy, canonicalName) => {
```
naugtur, line 116: all this is no longer necessary as this was alerting early about missing policy entries for known dependencies.

packages/compartment-mapper/src/policy.js
```diff
@@ -154,20 +131,14 @@ export const getPolicyForPackage = (namingKit, policy) => {
   if (canonicalName === ATTENUATORS_COMPARTMENT) {
     return {
       defaultAttenuator: policy.defaultAttenuator,
-      packages: detectAttenuators(policy).reduce((packages, specifier) => {
-        packages[specifier] = true;
-        return packages;
-      }, {}),
+      packages: 'any',
```
naugtur, line 134: this removed section had the bug where we mistook specifier for a resource id. and it was all unnecessary.

packages/compartment-mapper/src/policy.js
```diff
@@ -459,61 +438,3 @@ export const attenuateModuleHook = async (
     originalModuleRecord,
   });
 };
-
-const padDiagnosis = text => ` (${text})`;
```
naugtur, line 463: With compartments for disallowed dependencies being defined in the compartment map, loading them will reach the point where policy is checked and we now can avoid all the guesswork below.

packages/compartment-mapper/src/link.js
```diff
@@ -279,20 +274,19 @@ const makeModuleMapHook = (
...
-              foreignModuleSpecifier,
-              foreignCompartmentName,
-            })}`,
+            )}`,
           );
         }
 
-        // Despite all non-exit modules not allowed by policy being dropped
-        // while building the graph, this check is necessary because module
-        // is written back to the compartment map below.
```
naugtur, line 293: this comment is no longer relevant

packages/compartment-mapper/src/node-modules.js
boneskull: And [here](https://github.com/boneskull/endo/blob/ba77c6644962cf25510554fc9318d8a8094fd39a/packages/compartment-mapper/src/import-hook.js#L236)

boneskull: And [here](https://github.com/boneskull/endo/blob/ba77c6644962cf25510554fc9318d8a8094fd39a/packages/compartment-mapper/src/import-archive.js#L109) too 😄

packages/compartment-mapper/src/policy.js
```diff
@@ -154,20 +131,14 @@ export const getPolicyForPackage = (namingKit, policy) => {
...
-    validateDependencies(policy, canonicalName);
+  if (policy.resources && policy.resources[canonicalName] !== undefined) {
     return policy.resources[canonicalName];
   } else {
-    console.warn(
-      `No policy for '${canonicalName}', omitting from compartment map.`,
-    );
-    return undefined;
+    // Allow skipping policy entries for packages with no powers.
+    return {};
```
naugtur, line 141: create(null)

packages/compartment-mapper/test/test-policy.js
```diff
@@ -403,3 +414,18 @@ scaffold(
     },
   },
 );
+
+scaffold(
+  'policy - nested export in attenuator',
+  test,
+  fixture,
+  combineAssertions(
```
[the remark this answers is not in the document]
naugtur: Counting the expected number of assertions remains annoying tho.

packages/compartment-mapper/src/policy.js
```diff
@@ -357,25 +328,37 @@ export const attenuateGlobals = (
...
+  }
+  return ` (info: ${errorHint})`;
+};
 /**
  * Throws if importing of the specifier is not allowed by the policy
  *
  * @param {string} specifier
- * @param {object} compartmentDescriptor
+ * @param {import('./types.js').CompartmentDescriptor} compartmentDescriptor
  * @param {object} [info]
```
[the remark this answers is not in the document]
naugtur: I'll unwrap errorHint

packages/compartment-mapper/src/policy.js
naugtur: packages is normally a key/value map. there's only one string value allowed and that's the "any". There's no room for a mistake here

PR #1839, opened 2023-10-26
feat(compartment-mapper): add policy-related types

boneskull:
## Description

This adds/narrows types for policies and exports them from the `compartment-mapper` entry point. It also updates some usages of existing and new types (e.g., removing "`@typedef` references", squashing some null-reference bugs, etc.).

> [!NOTE]
> This does not change the fact that `@endo/compartment-mapper` does not export types properly to `node16`/`nodenext`-resolution consumers (see #1803).

**Reviewers:** please scrutinize the types added to `src/types.js` closely. The generics allow an attenuator implementation to define the types of custom params which it expects. For example:

```js
// policy-converter.js

/**
 * @typedef {'$root'} RootPolicy
 * @typedef {'write'} WritePolicy
 */

/**
 * Extends Endo's `PolicyItem` with the special {@link RootPolicy} and {@link WritePolicy}
 * @typedef {RootPolicy | WritePolicy} LavaMoatGlobalPolicyItem
 */

/**
 * Extends Endo's `PolicyItem` with the special {@link RootPolicy}
 * @typedef {RootPolicy} LavaMoatPackagePolicyItem
 */

/**
 * An Endo policy tailored to LavaMoat's default attenuator
 * @typedef {import('@endo/compartment-mapper').Policy<LavaMoatPackagePolicyItem, LavaMoatGlobalPolicyItem>} LavaMoatEndoPolicy
 */

/**
 * Package policy based on {@link LavaMoatPackagePolicyItem} and {@link LavaMoatGlobalPolicyItem}.
 *
 * Member of {@link LavaMoatEndoPolicy}
 * @typedef {import('@endo/compartment-mapper').PackagePolicy<LavaMoatPackagePolicyItem, LavaMoatGlobalPolicyItem>} LavaMoatPackagePolicy
 */
```

And in the attenuator:

```js
// my-attenuator.js

export default {
  /**
   * @type {import('@endo/compartment-mapper').GlobalAttenuatorFn<[import('./policy-converter.js').LavaMoatPackagePolicy['globals']]>}
   */
  attenuateGlobals(params, originalGlobalThis, packageCompartmentGlobalThis) {
    // ...
  },
  /**
   * @type {import('@endo/compartment-mapper').ModuleAttenuatorFn<[import('./policy-converter.js').LavaMoatPackagePolicy['packages']]>}
   */
  attenuateModule(params, originalObject) {
    // ...
  }
}
```

Does Endo consider type changes to be breaking?  These types will likely change in the future, and am curious about whether that should be a major bump. 

### Testing Considerations

I could use `tsd` to test some of these to avoid regressions, if desired.

### Upgrade Considerations

none

### DIFF
--- a/packages/compartment-mapper/src/node-modules.js
+++ b/packages/compartment-mapper/src/node-modules.js
@@ -631,7 +631,7 @@ const translateGraph = (
               name,
               path,
             },
-            packagePolicy,
+            /** @type {import('./types.js').PackagePolicy} */ (packagePolicy),
           )
         ) {
           moduleDescriptors[localPath] = {
--- a/packages/compartment-mapper/src/node-modules.js
+++ b/packages/compartment-mapper/src/node-modules.js
@@ -612,6 +612,12 @@ const translateGraph = (
       policy,
     );
 
+    /* c8 ignore next */
+    if (policy && !packagePolicy) {
+      // this should never happen
+      throw new TypeError('Unexpectedly falsy package policy');
+    }
+
     /**
      * @param {string} dependencyName
      * @param {string} packageLocation
--- a/packages/compartment-mapper/src/node-modules.js
+++ b/packages/compartment-mapper/src/node-modules.js
@@ -42,18 +42,18 @@
  * @typedef {Record<string, {spec: string, alias: string}>} CommonDependencyDescriptors
  */
 
+import { pathCompare } from './compartment-map.js';
 import { inferExportsAndAliases } from './infer-exports.js';
-import { searchDescriptor } from './search.js';
 import { parseLocatedJson } from './json.js';
-import { unpackReadPowers } from './powers.js';
+import { join } from './node-module-specifier.js';
+import { assertPolicy } from './policy-format.js';
 import {
-  getPolicyForPackage,
   ATTENUATORS_COMPARTMENT,
   dependencyAllowedByPolicy,
+  getPolicyForPackage,
 } from './policy.js';
-import { join } from './node-module-specifier.js';
-import { pathCompare } from './compartment-map.js';
-import { assertPolicy } from './policy-format.js';
+import { unpackReadPowers } from './powers.js';
+import { searchDescriptor } from './search.js';
 
 const { assign, create, keys, values } = Object;
 
@@ -566,7 +566,7 @@ const graphPackages = async (
  * @param {Graph} graph
  * @param {Set<string>} tags - build tags about the target environment
  * for selecting relevant exports, e.g., "browser" or "node".
- * @param {object|undefined} policy
+ * @param {import('./types.js').Policy} [policy]
  * @returns {CompartmentMapDescriptor}
  */
 const translateGraph = (
@@ -626,13 +626,14 @@ const translateGraph = (
         const localPath = join(dependencyName, exportPath);
         if (
           !policy ||
-          dependencyAllowedByPolicy(
-            {
-              name,
-              path,
-            },
-            /** @type {import('./types.js').PackagePolicy} */ (packagePolicy),
-          )
+          (packagePolicy &&
+            dependencyAllowedByPolicy(
+              {
+                name,
+                path,
+              },
+              packagePolicy,
+            ))
         ) {
           moduleDescriptors[localPath] = {
             compartment: packageLocation,
--- a/packages/compartment-mapper/src/policy-format.js
+++ b/packages/compartment-mapper/src/policy-format.js
@@ -60,10 +60,11 @@ export const isAllowingEverything = policyValue =>
  * @returns {allegedDefinition is import('./types.js').AttenuationDefinition}
  */
 export const isAttenuationDefinition = allegedDefinition => {
-  return (
-    (typeof allegedDefinition === 'object' &&
-      typeof allegedDefinition?.[ATTENUATOR_KEY] === 'string') || // object with attenuator name
-    isArray(allegedDefinition) // params for default attenuator
+  return Boolean(
+    (allegedDefinition &&
+      typeof allegedDefinition === 'object' &&
+      typeof allegedDefinition[ATTENUATOR_KEY] === 'string') || // object with attenuator name
+      isArray(allegedDefinition), // params for default attenuator
   );
 };
 
--- a/packages/compartment-mapper/src/policy-format.js
+++ b/packages/compartment-mapper/src/policy-format.js
@@ -133,11 +133,11 @@ const isPolicyItem = item =>
  * @param {unknown} allegedPackagePolicy - Alleged `PackagePolicy` to test
  * @param {string} path - Path in the `Policy` object; used for error messages only
  * @param {string} [url] - URL of the policy file; used for error messages only
- * @returns {allegedPackagePolicy is import('./types.js').PackagePolicy}
+ * @returns {asserts allegedPackagePolicy is import('./types.js').PackagePolicy|undefined}
  */
 export const assertPackagePolicy = (allegedPackagePolicy, path, url) => {
   if (allegedPackagePolicy === undefined) {
-    return true;
+    return;
   }
   const inUrl = url ? ` in ${q(url)}` : '';
 
@@ -191,7 +191,6 @@ export const assertPackagePolicy = (allegedPackagePolicy, path, url) => {
         builtins,
       })}${inUrl}`,
     );
-  return true;
 };
 
 /**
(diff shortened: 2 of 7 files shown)


boneskull:
cc @naugtur

packages/compartment-mapper/src/node-modules.js
```diff
@@ -636,7 +636,7 @@ const translateGraph = (
               name,
               path,
             },
-            packagePolicy,
+            /** @type {import('./types.js').PackagePolicy} */ (packagePolicy),
```
naugtur, line 639: note:
When we rebase this on top of #1838 the `packagePolicy` is always an object if `policy` is truthy. 
Currently passing undefined is being prevented by an early `continue`  on line 617

packages/compartment-mapper/src/policy-format.js
```diff
@@ -90,32 +95,49 @@ export const getAttenuatorFromDefinition = attenuationDefinition => {
...
- * @returns {void}
+ * @param {unknown} allegedPackagePolicy - Alleged `PackagePolicy` to test
+ * @param {string} path - Path in the `Policy` object; used for error messages only
+ * @param {string} [url] - URL of the policy file; used for error messages only
+ * @returns {allegedPackagePolicy is import('./types.js').PackagePolicy}
  */
 export const assertPackagePolicy = (allegedPackagePolicy, path, url) => {
   if (allegedPackagePolicy === undefined) {
-    return;
+    return true;
```
naugtur, line 140: This says undefined is PackagePolicy - I mean it's a correct value of an input where packagePolicy might be, but this is not strictly true I think.

packages/compartment-mapper/src/policy.js
```diff
@@ -113,8 +111,14 @@ export const dependencyAllowedByPolicy = (namingKit, packagePolicy) => {
...
 
-const validateDependencies = (policy, canonicalName) => {
-  const packages = policy.resources[canonicalName].packages;
+/**
+ * Asserts dependencies in a policy are defined
+ * @param {import('./types.js').Policy} policy
+ * @param {string} canonicalName
+ * @returns {void}
+ */
+const assertDependencies = (policy, canonicalName) => {
```
naugtur, line 120: This function got removed in #1838

packages/compartment-mapper/src/policy.js
```diff
@@ -171,21 +175,27 @@ export const getPolicyForPackage = (namingKit, policy) => {
...
+ * @param {import('./types.js').PackagePolicy} [packagePolicy]
+ * @returns {Array<string>}
+ */
 const getGlobalsList = packagePolicy => {
-  if (!packagePolicy.globals) {
+  if (!packagePolicy?.globals) {
     return [];
   }
-  return entries(packagePolicy.globals)
+  return entries(packagePolicy?.globals)
```
naugtur, line 187: (should) avoid the `?` here, there's an early return that handles it already, and passing `null` to `entries` would throw anyway.

packages/compartment-mapper/src/types.js
```diff
@@ -420,12 +423,84 @@ export {};
...
+ */
+
+/**
+ * @template {[any, ...any[]]} [Params=[any, ...any[]]]
+ * @callback GlobalAttenuatorFn
+ * @param {Params} params
+ * @param {Record<PropertyKey, any>} originalObject
+ * @param {Record<PropertyKey, any>} globalThis
+ * @returns {void}
+ * @todo Unsure if we can do much typing of `originalObject` and `globalThis` here.
```
naugtur, line 440: `originalObject` is the same as the value passed in as `global` to `importLocation` options. I don't think it's reasonable to thread it all the way if TS can't infer.

packages/compartment-mapper/src/types.js
```diff
@@ -420,12 +423,84 @@ export {};
...
+ * @typedef {Record<string, AttenuationDefinition | boolean>} NestedAttenuationDefinition
+ */
+
+/**
+ * An object representing a base package policy.
+ * @template [PackagePolicyItem=void]
+ * @template [GlobalsPolicyItem=void]
+ * @template [BuiltinsPolicyItem=void]
+ * @typedef {object} PackagePolicy
+ * @property {string} [defaultAttenuator] - The default attenuator.
```
naugtur, line 485: (nit) There's no defaultAttenuator field on the individual resource in policy in the public API. 
I'm guessing this is the consequence of defaultAttenuator being saved in the package policy of the attenuators compartment. Should we document it here? Should we sub-type internally? Should we implement a valid usecase for it?

naugtur:
I wanted to rebase this and add cleanup to one of mine functions but noticed this is from a fork.
@boneskull want me to fold this into a branch on the endo repo?


patch I wish to apply after rebase on master:
```
diff --git a/packages/compartment-mapper/src/policy.js b/packages/compartment-mapper/src/policy.js
index 8604bcb2a..c646c37c3 100644
--- a/packages/compartment-mapper/src/policy.js
+++ b/packages/compartment-mapper/src/policy.js
@@ -339,26 +339,28 @@ const diagnoseModulePolicy = errorHint => {
  *
  * @param {string} specifier
  * @param {import('./types.js').CompartmentDescriptor} compartmentDescriptor
- * @param {object} [info]
+ * @param {object} [options]
+ * @param {boolean} [options.exit] - Whether it is an exit module
+ * @param {string} [options.errorHint] - Error hint message
  */
 export const enforceModulePolicy = (
   specifier,
   compartmentDescriptor,
-  info = {},
+  { exit, errorHint } = {},
 ) => {
   const { policy, modules, label } = compartmentDescriptor;
   if (!policy) {
     return;
   }
 
-  if (!info.exit) {
+  if (!exit) {
     if (!modules[specifier]) {
       throw Error(
         `Importing ${q(specifier)} in ${q(
           label,
         )} was not allowed by packages policy ${q(
           policy.packages,
-        )}${diagnoseModulePolicy(info.errorHint)}`,
+        )}${diagnoseModulePolicy(errorHint)}`,
       );
     }
     return;
@@ -368,7 +370,7 @@ export const enforceModulePolicy = (
     throw Error(
       `Importing ${q(specifier)} was not allowed by policy builtins:${q(
         policy.builtins,
-      )}${diagnoseModulePolicy(info.errorHint)}`,
+      )}${diagnoseModulePolicy(errorHint)}`,
     );
   }
 };
```

boneskull:
@naugtur Resolved

packages/compartment-mapper/src/types.js
boneskull: I saw it referenced [here](https://github.com/endojs/endo/blob/master/packages/compartment-mapper/src/policy-format.js#L132), but I can't answer your other questions.  If it's intended to be public, then great, but if not, we should subtype.

boneskull:
@naugtur OK, so, it's _slightly_ different than the diff you gave me, since the diff you gave me did not match the snapshots; I retained the use of the `label` prop from the compartment descriptor

Please take another look at the changes I made based on your comments.

packages/compartment-mapper/src/types.js
boneskull: Unsure if there's something actionable here

packages/compartment-mapper/src/types.js
boneskull: I'd lean towards subtyping it (though I'm not sure where that type assignment needs to happen) for now unless there's an obvious use-case.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -626,13 +626,14 @@ const translateGraph = (
...
         if (
           !policy ||
-          dependencyAllowedByPolicy(
-            {
-              name,
-              path,
-            },
-            /** @type {import('./types.js').PackagePolicy} */ (packagePolicy),
-          )
+          (packagePolicy &&
```
naugtur, line 629: if `policy` is truthy and `packagePolicy` is falsy, we're under attack. should we introduce an assert before the if to make sure we explode immediately?

packages/compartment-mapper/src/policy.js
```diff
@@ -149,7 +166,7 @@ const getGlobalsList = packagePolicy => {
   if (!packagePolicy?.globals) {
     return [];
   }
-  return entries(packagePolicy?.globals)
+  return entries(packagePolicy.globals)
```
naugtur, line 169: valid point! :)

packages/compartment-mapper/src/node-modules.js
boneskull: @naugtur I've added an assertion to this effect

naugtur (approved):

PR #1840, opened 2023-10-27
fix(ses): fix types export for newer module resolutions

boneskull:
Much like #1803, this adds `types` conditional exports where appropriate.

Additionally, I fixed a type in `ThirdPartyStaticModuleInterface` which I was bumping up against.

### DIFF
--- a/packages/ses/package.json
+++ b/packages/ses/package.json
@@ -36,11 +36,13 @@
   "exports": {
     ".": {
       "import": "./index.js",
-      "require": "./dist/ses.cjs"
+      "require": "./dist/ses.cjs",
+      "types": "./types.d.ts"
     },
     "./lockdown": {
       "import": "./index.js",
-      "require": "./dist/ses.cjs"
+      "require": "./dist/ses.cjs",
+      "types": "./types.d.ts"
     },
     "./tools.js": "./tools.js",
     "./package.json": "./package.json"
--- a/packages/ses/types.d.ts
+++ b/packages/ses/types.d.ts
@@ -58,8 +58,11 @@ export interface PrecompiledStaticModuleInterface {
 export interface ThirdPartyStaticModuleInterface {
   imports: Array<string>;
   exports: Array<string>;
+  /**
+   * Note that this value does _not_ contain any numeric or symbol property keys, which can theoretically be members of `exports` in a CommonJS module.
+   */
   execute(
-    proxiedExports: Object,
+    proxiedExports: Record<string, any>,
     compartment: Compartment,
     resolvedImports: Record<string, string>,
   ): void;
--- a/packages/ses/package.json.md
+++ b/packages/ses/package.json.md
@@ -3,7 +3,7 @@
 This is an explainer for the module system configuration of the SES shim
 package through some properties of `package.json`.
 
-## "type": "module",
+## "type": "module"
 
 For Node.js, this means that any `.js` file will be interpreted as if it were
 `.mjs`, meaning a JavaScript module / ESM, as opposed to `.cjs` which is
@@ -14,7 +14,7 @@ and translates all `.js` files down from ESM to CommonJS.
 A patch to `esm` that is in the Agoric SDK repository allows `.js`
 files in dependencies to break through to the underlying Node.js ESM.
 
-## "main": "./dist/ses.cjs",
+## "main": "./dist/ses.cjs"
 
 SES provides its own translation from its own ESM sources to CommonJS, emitted
 by `yarn build`, specifically `scripts/bundle.js`.
@@ -35,7 +35,7 @@ The `main` property has been supported by the npm ecosystem since the
 earliest versions, so every version of Node.js and every tool will look
 here if nothing else in `package.json` overrides it.
 
-## "module": "./index.js",
+## "module": "./index.js"
 
 Some tools like WebPack, Parcel and the ESM emulation provided by `node -r esm`
 use this instead of `main` if it is present.
@@ -49,7 +49,7 @@ non-ASCII characters (we use zero-width-joiner to avoid collisions with other
 names in scope, then censor the use of zero-width-joiner in source).
 Most tools tolerate this, but WebPack does not.
 
-## "unpkg": "./dist/ses.umd.js",
+## "unpkg": "./dist/ses.umd.js"
 
 The [Unpkg][] CDN uses this property to direct usage of SES to a precompiled
 module in "Universal Module Definition" format.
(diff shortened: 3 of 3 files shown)


packages/ses/types.d.ts
```diff
@@ -59,7 +59,7 @@ export interface ThirdPartyStaticModuleInterface {
   imports: Array<string>;
   exports: Array<string>;
   execute(
-    proxiedExports: Object,
+    proxiedExports: Record<PropertyKey, any>,
```
mhofman, line 62: I'll defer the review of this change to @kriskowal

mhofman (approved):
The changes to `package.json` is fine. It's necessary given the way the "exports" and "types" fields  work.

packages/ses/types.d.ts
boneskull: The problem with `Object` is that you cannot assign to it.  You can't assign to `object`, either.

kriskowal: And I defer to the submitter since we can rely on @boneskull to answer support questions on Matrix going forward :-). I have no idea. Did not know `PropertyKey` was a TS built-in.

packages/ses/package.json
```diff
@@ -36,11 +36,13 @@
   "exports": {
     ".": {
       "import": "./index.js",
-      "require": "./dist/ses.cjs"
+      "require": "./dist/ses.cjs",
+      "types": "./types.d.ts"
```
kriskowal, line 40: Can you add a line to `package.json.md` explaining this?

kriskowal (approved):

packages/ses/package.json
boneskull: @kriskowal Done.  Also fixed another path in there that was wrong.

packages/ses/types.d.ts
boneskull: Yep! It's `string | number | symbol` from the `es5` internal lib.

boneskull:
@kriskowal Looks like CI needs to be burped

packages/ses/types.d.ts
mhofman: Is this object parameter meant to be mutated? That seem contrary to our programming model of hardening what is passed.

kriskowal: The `proxiedExports` is the internal view of the namespace and `exportsProxy` is the external view of the namespace. The internal view is mutated by bindings during evaluation of the module functor, and the external view needs to be consistent with namespace objects, which are not mutable.

The names are bad. Since I wrote this, I learned about “module environment record” which is approximately the 262 name for the internal object I’m calling `proxiedExports`. Not precisely. And it’s an awful name too.

boneskull: that's all over my head, but this is what prompted the change:

```js
export const importHook = async (specifier) => {
  const ns = await import(specifier)
  return Object.freeze(
    /** @type {import('ses').ThirdPartyStaticModuleInterface} */ ({
      imports: [],
      exports: Object.keys(ns),
      execute: (moduleExports) => {
        moduleExports.default = ns // <--- TYPE ERROR
        Object.assign(moduleExports, ns)
      },
    })
  )
}
```

boneskull: @kriskowal Also, it shouldn't be breaking anybody to change the name of that argument.  I can change it to something else..?

kriskowal: Let’s reserve renaming to another change, which you’re welcome to propose! I am dissatisfied with the name but don’t have a better idea yet. Ideas welcome! `namespace` is a good name for `proxiedExports` if there’s a dual that doesn’t have `namespace` in its name. `privateNamespace` vs `publicNamespace` or `internalNamespace` vs `externalNamespace` would be clearer albeit tortured. I like `bindings` except that I want to use that word for a property of a `ModuleSource` object that reflects the “bindings” between the internal and external namespace objects.

kriskowal: I should note that `bindings` is the name that our friends at Moddable suggested and it was such a good name that it made clear in my mind how these two objects are related. The `as` keyword does a lot of work in a “binding” like `export { name as rename }`. That would create a “binding” between `internal.name` and `external.rename` for a pair of module namespaces. It would be reasonable to call both namespaces and distinguish them by context in most places.

I might buy (external) `namespace` and (internal) `environment`.

mhofman: Yeah if the argument is meant to be assigned to, `Record` is the right type. It is just a surprising API shape.

kriskowal: Oh, also worth noting that `PropertyKey` might be too broad. I don’t think modules can import or export symbol-named or nominal properties. I _do_ know that the strings aren’t not limited to valid JavaScript identifiers!

boneskull: > Oh, also worth noting that PropertyKey might be too broad. I don’t think modules can import or export symbol-named or nominal properties. I do know that the strings aren’t limited to valid JavaScript identifiers!

That appears to be true.  It's not true about CJS, of course.  If we narrow the type further to `Record<string, any>`, this _could_ impact something working with a CJS package, though perhaps unlikely.

How about this: let's do `Record<string, any>` for now, and if it becomes a problem, we change it to `Record<PropertyKey, any>` later. That won't be a breaking change, but narrowing the type arguably would.

boneskull: I've updated the code to do that.

kriskowal: Summoning @naugtur to rule on the ESM<->CJS namespace bindings behavior for symbols. My perception is that they would appear on `namespace.default` but not be replicated as `namespace[symbol]`.

naugtur: I'm gonna need to look into it to make sure, but off the top of my head I think there's no way to define a symbol key in the record's exports but on the other hand my implementation of lifting them from defaults is not explicitly skipping symbols. Gotta dig deeper.

boneskull: @naugtur Understood.  This begs the question, though, _should_ `Object.keys()` be used there, and not, say, `Object.getOwnPropertyDescriptors()`?

boneskull: I've updated the comment.

boneskull:
Please note that https://github.com/endojs/endo/pull/1803 is the same sort of thing but for `@endo/compartment-mapper`

packages/ses/types.d.ts
naugtur: I'm not sure if it's possible for whatever Object.keys would skip to be detected as named exports by the lexer and only the pre-defined top-level fields will be able to populate from cjs to esm. But I can't say you're not right pointing that out.

One case I can think of is having a getter for lazy loading. I wonder if copying a getter over would even work as intended.

PR #1841, opened 2023-10-27
fix(evasive-transform): replace homoglyphs with boring ascii

boneskull:
This eschews homoglyphs in the comment evasion transform because they are "too clever" and not "tamper-evident" enough.  But I think it's just envy.

cc @kriskowal


kriskowal (approved):

boneskull:
@kriskowal Not in a hurry with this one, but curious if we are waiting for more eyes--or how exactly that works 'round these parts.

kriskowal:
In this case, you’ve got my stamp and I think I speak for the Agoric team that this is a preferable behavior. We generally prefer for the submitter to land if they’re a member of the org, so they have an opportunity to groom the history with a local rebase on `origin/master` before merging thru the Github UI.

We like to keep the git history looking like a single railway line with non-overlapping sidings for PRs, though we understand that’s not always possible. We like to have individually reviewable commits, but also understand there are many cases where a single commit will suffice. We still like to use the merge commit from the Github UI to preserve the signature of the individual commits, if there was a signature.

I’ve sent you an invitation to join the @endojs org which should make it possible for you to land this.

For Agoric SDK, this is our policy for commit history management https://github.com/Agoric/agoric-sdk/wiki/Commit-Management. However, we do not use a merge queue on Endo yet. We don’t have enough concurrent work to require it yet.

Issue #812, opened 2021-07-06
Use eshost to run tests in multiple engines


mhofman:
https://github.com/bterlson/eshost allows running code in different agents / engines. It would be great to run endo tests in all engines to make sure no implementations exhibit unexpected behavior.

mhofman:
@kriskowal I believe this would be subsumed by moving to test262 and a runner based on eshost?

kriskowal:
Certainly overlaps:
- https://github.com/endojs/endo/issues/870
- https://github.com/endojs/endo/issues/1001

Those issues could be reframed as this one. We need something parody of eshost that runs on every engine with SES shim pre-initialized.

kriskowal:
With https://github.com/endojs/endo/pull/1847 we have a place to stand for adding these tests.

Issue #870, opened 2021-08-25
SES test262 conformance suite


kriskowal:
In order to provide a foundation for validating parity with the XS implementation of SES, particularly the module system changes both proposed and not-yet-proposed, we should extract a SES conformance suite from the SES unit tests such that it runs under test262. This would also prime the pump for proposing tests for SES proposals for ECMA 262.

PR #1847, opened 2023-11-03
Revitalize test262 runner

kriskowal:
refs: https://github.com/endojs/endo/issues/870
closes: #812 

## Description

We have not been running test262 since we migrated from tape to ava for lack of a suitable runner. We have deeper needs for a test262 runner now since we would like to verify parity between node+ses and XS. This change pivots the runner from a bespoke implementation to bterlsen’s test262-harness which supports node and xs. Then, by way of demonstration, we bring in XS tests and get one of them passing in SES.


packages/test262-runner/test262/test/built-ins/Compartment/prototype/Symbol.toStringTag.js
```diff
@@ -1,7 +1,7 @@
 /*---
 description:
 includes: [propertyHelper.js]
-features: [Symbol.toStringTag,Compartment]
+features: [Symbol.toStringTag,Compartment,ses-xs-parity]
```
kriskowal, line 4: This is the one XS test that I’ve so far attempted to get working for both Node.js and XS.

packages/test262-runner/test262/test/built-ins/Compartment/prototype/Symbol.toStringTag-lockdown.js
```diff
@@ -0,0 +1,13 @@
...
+features: [Symbol.toStringTag,Compartment,lockdown,ses-xs-parity]
+---*/
+
+lockdown();
+
+assert.sameValue(Compartment.prototype[Symbol.toStringTag], 'Compartment');
+
+verifyNotEnumerable(Compartment.prototype, Symbol.toStringTag);
+verifyNotWritable(Compartment.prototype, Symbol.toStringTag);
+verifyNotConfigurable(Compartment.prototype, Symbol.toStringTag);
```
kriskowal, line 13: This is a new test262 style test for the post-lockdown appearance of Compartment.prototype.@@toStringTag.

kriskowal (commented):
Individually reviewable commits. Please ignore the large, mechanical commits that sync test262 and XS test262 suites, and the test262 harness sync. Only the two tests marked `features=[ses-xs-parity]` currently pass. Also note, I’ve done nothing to ensure a working version of `xst` is on your path when you run `yarn test262:xs`. That’s an exercise I’ll undertake as part of the CI script for test262-runner when I reënable it in a future change.

kriskowal:
attn @phoddie, this is the PR that’s introducing eshost based SES parity testing. It requires running our shim in a prelude to every SES test, so it’s slow, but a place to start. This also gives us a place to stand to carve out a leaner SES shim specific to XS.

packages/test262-runner/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "../../tsconfig.eslint-base.json",
+  "compilerOptions": {
+    "allowJs": true
+  },
```
erights, line 5: For this PR, no change suggested.

But I've been meaning to ask about this. Some packages have a `tsconfig.json` without this section. Some have this section. And some have it with an additional 
```json
    "maxNodeModuleJsDepth": 1, // or 2
```
At #2042 I found I had to add one to @endo/pass-style to get the types to show up in my IDE as expected. I gave it a max depth of 1, but not for any particular reason. Can we be uniform? Is there a default config for new packages?

erights (approved):
LGTM. Full speed ahead!

packages/test262-runner/tsconfig.json
kriskowal: I’m alarmed that this was necessary within Endo. My read of how this feature works is that TSC literally counts occurrences of the term `node_modules` in the canonical path to the module, not the shortest dependency path. Within the Endo repository, the physical locations of each package contain _no_ path components of `node_modules` so TSC should reach indefinitely. This was why we didn’t encounter the need for this feature until packages started moving out of Agoric SDK, increasing the depth.

But, we may have just been lucky until now. I find it alarming that we would ever need this for local development in Endo.

packages/ses/package.json
```diff
@@ -46,6 +46,9 @@
       "require": "./dist/ses.cjs"
```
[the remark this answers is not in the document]
kriskowal: https://nodejs.org/api/packages.html#package-entry-points

The `exports` and `imports` fields are a sort of DSL where each node of the tree can _either_ switch on a predicate (like `node`, `types`, `import`, `require`, or `default` and our Endo allows arbitrary others like `endo`), _OR_ map exported names to internal names. Node.js distinguishes which kind of node it is based on whether _every_ key either starts with `.` or every key does not start with `.`.

We started with just one export (`"."`). Some tools like Parcel expected Node.js to allow them to import `ses/package.json`, so we generally always also export `./pakage.json`. We made a concession for `./tools.js` so they could be imported by LavaMoat avoiding a proliferation of small packages. In #1981, I felt that was a case where we needed a separate package to keep APIs loosely coupled. That brings us to this PR.

Now that we have `repairIntrinsics` and `hardenIntrinsics` separated, our design tensions have shifted. Now, it’s possible to expose the internal layers of SES. This gives us an opportunity to make a more focused SES for XS. We also may in time be able to factor out the assert/errors, console, and platform error unwrapping shims.

packages/test262-runner/README.md
```diff
@@ -1,10 +1,18 @@
 # test262-runner
 
-Run ECMAScript compliance tests on `node` using native ESM support.
+Run ECMAScript compliance tests on Node.js and XS with a prelude that shims
```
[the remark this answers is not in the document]
kriskowal: Yes.

packages/test262-runner/tsconfig.json
erights: @turadg , do you understand what the effects are of including vs omitting a `"compilerOptions":` section in tsconfig.js files? Are you surprised that without them I was at least missing type info in my vscode IDE?

turadg: The effect of `compilerOptions` from a tsconfig is to use the TS defaults, or those inherited from another config.

The effect of each option is described in https://www.typescriptlang.org/tsconfig#compilerOptions

In Endo, the options for "build" are different from normal dev. The dev tsconfig in every package inherits from these this file:
- https://github.com/endojs/endo/blob/75bd2d2f71038bc720af7d927b0052041f07f199/tsconfig.eslint-base.json#L2-L12

The options used in various packages:
- `checkJs` - whether to check all `.js` files or just the ones with `// @ts-check`
- `allowSyntheticDefaultImports` - whether to allow `import x from y` when the module doesn't have a default export (like CJS)
- `noImplicitAny` - a strictness check to prevent unannotated values from silently being `any`.
- [maxNodeModuleJsDepth](https://www.typescriptlang.org/tsconfig#maxNodeModuleJsDepth) - The maximum dependency depth to search under node_modules and load JavaScript files.

> Are you surprised that without them I was at least missing type info in my vscode IDE?

Not surprising that that state was encountered. When refactoring packages, the search depth may have to be increased because we don't build `.d.ts` files in development.

> Can we be uniform? Is there a default config for new packages?

The default is zero (i.e. when not specified). The only way to be uniform is to adopt the smallest number that works for all packages, but that would make some packages slower and possibly introduce new errors as they resolve something new that then needs more coverage to continue.

Note that this is only an issue while working in an IDE. The typedoc build in CI ensures that the types of the entire repo resolve sanely.

If we want to also ensure that type coverage never falls on a per-package basis, we can add a check to CI that uses compares its `typeCoverage` value in its package.json with the coverage after the PR changes.

Issue #1849, opened 2023-11-07
`Promise.withResolvers` is unexpectedly removed


doodlewind:
## Describe the bug

We are using SES for simulating sandbox, and recently we found that the newly added native API `Promise.withResolvers` ([proposal](https://github.com/tc39/proposal-promise-with-resolvers)) is not recognized and removed by SES.

## Steps to reproduce

Call `lockdown()` in latest Chrome environment.

## Screenshots

<img width="901" alt="image" src="https://github.com/endojs/endo/assets/7312949/40da2c6c-8e59-4312-90d1-d15340738dc0">

kriskowal:
Thanks, @doodlewind. Always appreciate planned work like this. Please do find us in [Matrix](https://matrix.to/#/!MYcRZqNhDImDvJSOCX:matrix.org?via=matrix.org&via=luxaritas.com&via=alteck.club) and tell us more about your application. We always appreciate such stories and having folks to engage about emerging changes.

PR #1850, opened 2023-11-07
Add `Promise.withResolvers` to permits list

doodlewind:
closes: #1849

## Description

This allows `Promise.withResolves` to the permits list.

### DIFF
--- a/packages/ses/src/permits.js
+++ b/packages/ses/src/permits.js
@@ -1414,6 +1414,7 @@ export const permitted = {
     race: fn,
     reject: fn,
     resolve: fn,
+    withResolvers: fn,
     '@@species': getter,
   },
 


kriskowal (approved):

kriskowal:
Tapping @erights. I am almost 100% confident that this is complete and consistent with our design. Hardened JavaScript does not ensure that all objects obtained through shared intrinsics are themselves hardened, just that their prototypes are hardened. We do not harden instances returned by intrinsics unless they themselves are implicitly shared with other compartments. I’m merging and commit to fix before release if not.

erights:
>  commit to fix before release if not

Please add a comment above entry linking to https://github.com/tc39/proposal-promise-with-resolvers , consistent with other such comments in permits.js

erights:
Other than that, LGTM

> does not ensure that all objects obtained through shared intrinsics are themselves hardened, just that their prototypes are hardened. We do not harden instances returned by intrinsics unless they themselves are implicitly shared with other compartments.

Of those two sentences, the second is correct. The first sentence by itself
* mistakenly doesn't include shared intrinsics other than prototypes.
* mistakenly includes prototypes, even if they are freshly made and therefore unshared, IIUC such as the prototypes of new async generator functions.

PR #1852, opened 2023-11-08
docs(ses): Annotate Promise.withResolvers permit

kriskowal:
Follow-up from https://github.com/endojs/endo/pull/1850. I was hasty.


erights (changes requested):
Copy-n-paste error? Stale clipboard?

kriskowal:
I’ve had better days. Thanks for spotting.

erights (approved):
LGTM

Issue #1853, opened 2023-11-08
Add version number to SES lockdown shim


leotm:
## What is the Problem Being Solved?

After downloading a lockdown shim, it's hard to remember which version it is later

e.g. `curl -O https://npmfs.com/download/ses/0.18.8/dist/ses.cjs`
but it's difficult to tell the version purely from inspecting the file

## Description of the Design

Start the bundled lockdown file with a comment containing the version number

## Test Plan

```bash
npm run build
```

```bash
head dist/ses.cjs # ensure correct version present as comment on line 1
```

```bash
npm run test
```

PR #1854, opened 2023-11-08
feat(ses): add SES version to lockdown shim

leotm:
closes: #1853

## Description

Prepends the `ses` _package.json > version_ as a comment to the bundled lockdown file

Possibly worth mentioning the version number is now included in the SES lockdown shim

### Testing Considerations

Possibly add a unit test to check the version number is present at the head of the file

Manually tested locally:

```bash
git clean -fdx
```

```bash
nvm use 16 # or 18 or 20
```

```bash
npm i --force # https://github.com/endojs/endo/assets/1881059/72a4079c-fe00-4de8-ac2d-fb13406edbad
```

```bash
cd packages/ses
npm run build
```

```bash
head dist/ses.cjs # ensure correct version present as comment on line 1
```

```bash
npm run test
```

on Node
- v14.20.1
  - ❌ [CI errors](https://github.com/endojs/endo/pull/1854#discussion_r1387159574)
  - now [fixed](https://github.com/endojs/endo/pull/1854#discussion_r1388081581)
- v16.20.0
- v18.17.1

### DIFF
--- a/packages/ses/scripts/bundle.js
+++ b/packages/ses/scripts/bundle.js
@@ -4,7 +4,8 @@ import fs from 'fs';
 import { makeBundle } from '@endo/compartment-mapper/bundle.js';
 import { minify } from 'terser';
 import { fileURLToPath, pathToFileURL } from 'url';
-import packageJson from '../package.json' assert { type: 'json' };
+
+const textDecoder = new TextDecoder();
 
 const resolve = (rel, abs) => fileURLToPath(new URL(rel, abs).toString());
 const root = new URL('..', import.meta.url).toString();
@@ -16,7 +17,13 @@ const write = async (target, content) => {
 };
 
 const main = async () => {
+  const bytes = await fs.promises.readFile(
+    fileURLToPath(`${root}/package.json`),
+  );
+  const text = textDecoder.decode(bytes);
+  const packageJson = JSON.parse(text);
   const version = packageJson.version;
+
   const bundle = `// v${version}\n${await makeBundle(
     read,
     pathToFileURL(resolve('../index.js', import.meta.url)).toString(),
--- a/packages/ses/scripts/bundle.js
+++ b/packages/ses/scripts/bundle.js
@@ -38,15 +38,21 @@ const main = async () => {
   console.log(`Minified bundle size: ${terse.length} bytes`);
 
   await fs.promises.mkdir('dist', { recursive: true });
-  await write('dist/ses.cjs', versionedBundle);
-  await write('dist/ses.mjs', versionedBundle);
-  await write('dist/ses.umd.js', versionedBundle);
-  await write('dist/ses.umd.min.js', terse);
 
-  await write('dist/lockdown.cjs', versionedBundle);
-  await write('dist/lockdown.mjs', versionedBundle);
-  await write('dist/lockdown.umd.js', versionedBundle);
-  await write('dist/lockdown.umd.min.js', terse);
+  const bundles = [
+    'dist/ses.cjs',
+    'dist/ses.mjs',
+    'dist/ses.umd.js',
+    'dist/lockdown.cjs',
+    'dist/lockdown.mjs',
+    'dist/lockdown.umd.js',
+  ];
+  const terses = ['dist/ses.umd.min.js', 'dist/lockdown.umd.min.js'];
+
+  await Promise.all([
+    ...bundles.map(dest => write(dest, versionedBundle)),
+    ...terses.map(dest => write(dest, terse)),
+  ]);
 };
 
 main().catch(err => {
--- a/packages/ses/scripts/bundle.js
+++ b/packages/ses/scripts/bundle.js
@@ -24,28 +24,30 @@ const main = async () => {
   const packageJson = JSON.parse(text);
   const version = packageJson.version;
 
-  const bundle = `// ses@${version}\n${await makeBundle(
+  const bundle = await makeBundle(
     read,
     pathToFileURL(resolve('../index.js', import.meta.url)).toString(),
-  )}`;
-  const { code: terse } = await minify(bundle, {
+  );
+  const versionedBundle = `// ses@${version}\n${bundle}`;
+
+  const { code: terse } = await minify(versionedBundle, {
     mangle: false,
     keep_classnames: true,
   });
   assert.string(terse);
 
-  console.log(`Bundle size: ${bundle.length} bytes`);
+  console.log(`Bundle size: ${versionedBundle.length} bytes`);
   console.log(`Minified bundle size: ${terse.length} bytes`);
 
   await fs.promises.mkdir('dist', { recursive: true });
-  await write('dist/ses.cjs', bundle);
-  await write('dist/ses.mjs', bundle);
-  await write('dist/ses.umd.js', bundle);
+  await write('dist/ses.cjs', versionedBundle);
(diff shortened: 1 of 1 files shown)


packages/ses/scripts/bundle.js
```diff
@@ -15,10 +16,11 @@ const write = async (target, content) => {
 };
 
 const main = async () => {
-  const bundle = await makeBundle(
+  const version = packageJson.version;
+  const bundle = `// v${version}\n${await makeBundle(
```
leotm, line 20: or

```suggestion
  const bundle = `// ses@${version}\n${await makeBundle(
```

packages/ses/scripts/bundle.js
```diff
@@ -4,6 +4,7 @@ import fs from 'fs';
 import { makeBundle } from '@endo/compartment-mapper/bundle.js';
 import { minify } from 'terser';
 import { fileURLToPath, pathToFileURL } from 'url';
+import packageJson from '../package.json' assert { type: 'json' };
```
leotm, line 7: TODO: rewrite to fix `test-async-hooks (16.x, ubuntu-latest)` failing tests

```console
import packageJson from '../package.json' assert { type: 'json' };
                                          ^^^^^^

SyntaxError: Unexpected identifier
    at Loader.moduleStrategy (node:internal/modules/esm/translators:146:18)
    at async link (node:internal/modules/esm/module_job:67:21)
```

leotm, line 7: TODO: rewrite to fix `test-async-hooks (16.x, ubuntu-latest)` failing tests

```console
import packageJson from '../package.json' assert { type: 'json' };
                                          ^^^^^^

SyntaxError: Unexpected identifier
    at Loader.moduleStrategy (node:internal/modules/esm/translators:146:18)
    at async link (node:internal/modules/esm/module_job:67:21)
```

leotm: fixed: https://github.com/endojs/endo/pull/1854/commits/c14ff1c2ea1490558cb8073a0158708558d683ae

packages/ses/scripts/bundle.js
```diff
@@ -15,10 +17,17 @@ const write = async (target, content) => {
 };
 
 const main = async () => {
-  const bundle = await makeBundle(
+  const bytes = await fs.promises.readFile(
```
leotm, line 20: or `fs.readFileSync` (prob not)

kriskowal: No change requested. Here’s some feedback on how I think about this trade-off:

This is a place where it doesn’t matter one way or the other. Synchronous is actually faster. The reason to favor async in Node.js is to reduce variation in latency on a system serving multiple concurrent requests, to mitigate head-of-line blocking in the event loop. Since this is not a server, that’s not a win.

The other good reason to avoid the sync functions is that sync I/O is not portable to the web and embracing async early can avoid an expensive and backward-incompatible refactor in the future if the code needs to be portable to the web later. That is also not a concern here.

The only remaining good reason is to pretend that the sync functions don’t even exist so you don’t have to think about any of the above subtle trade-offs, since the performance win is not worth thinking about.

packages/ses/scripts/bundle.js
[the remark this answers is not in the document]
kriskowal: Optional: I would like to move the await into its own statement and name the intermediate variable. Maybe this can be `versionedBundle`.

Agoric style encourages `await` to only be used as a top level expression or the right-side of an assignment in order to make the event delimiter obvious to reviewers. We sometimes even put large comments above the await so it doesn’t go unnoticed. A lot of the Agoric code operates in an adversarial environment and uses events as low-weight transactions. All code is responsible for maintaining its invariants between events so that events from an attacker don’t get interleaved between partial events. This is not a concern for this script, but a habit we like to encourage.

kriskowal: These are both good options. I have a slight preference for matching the git/npm tag.

kriskowal (approved):
Feedback for your consideration. If you are unable to merge this, please reply when it’s ready to merge.

packages/ses/scripts/bundle.js
boneskull: This is my stance too.  I almost never write server code, but I avoid sync methods for the above reasons.  Mostly the "don't have to think" reason, which is truly underrated.

packages/ses/scripts/bundle.js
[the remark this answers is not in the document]
boneskull: @kriskowal What's the "event delimiter"?

packages/ses/scripts/bundle.js
```diff
@@ -15,10 +17,17 @@ const write = async (target, content) => {
...
+    fileURLToPath(`${root}/package.json`),
+  );
+  const text = textDecoder.decode(bytes);
```
boneskull, line 23: Can we not do this:

```suggestion
  const test = await fs.promises.readFile(
    fileURLToPath(`${root}/package.json`),
    'utf8'
  );
```

and drop the `TextDecoder`?  Is this an Endo convention?

boneskull (approved):
Looks fine, but had a question or two

packages/ses/scripts/bundle.js
leotm: nice the 2nd arg w `utf8` cleans up the need for `TextDecoder` ^
spotted `utf-8` is also used, but don't think it matters? (same result)
if not, then RE convention i favour our suggestion (the default)

Screenshot 2023-11-20 at 4 39 04 pm

packages/ses/scripts/bundle.js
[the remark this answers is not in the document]
leotm: i initially split it to `versionedBundle` too, but reverted thinking too many invasive changes 😅 updated in https://github.com/endojs/endo/pull/1854/commits/bccf39aea1b0414102bbf13cce05fac6a545f28d

> @kriskowal What's the "event delimiter"?

`await` delimiter i believe (as opposed to template var backtick or dollar sign delimiters)

packages/ses/scripts/bundle.js
```diff
@@ -15,28 +15,37 @@ const write = async (target, content) => {
...
+  await write('dist/ses.umd.js', versionedBundle);
   await write('dist/ses.umd.min.js', terse);
 
-  await write('dist/lockdown.cjs', bundle);
-  await write('dist/lockdown.mjs', bundle);
-  await write('dist/lockdown.umd.js', bundle);
+  await write('dist/lockdown.cjs', versionedBundle);
+  await write('dist/lockdown.mjs', versionedBundle);
+  await write('dist/lockdown.umd.js', versionedBundle);
   await write('dist/lockdown.umd.min.js', terse);
```
boneskull, line 49: Make go faster?

```js
const bundles = [
  'dist/ses.cjs',
  'dist/ses.mjs',
  'dist/ses.umd.js',
  'dist/lockdown.cjs',
  'dist/lockdown.mjs',
  'dist/lockdown.umd.js'
];
const terses = ['dist/ses.umd.min.js', 'dist/lockdown.umd.min.js'];

await Promise.all([
  ...bundles.map(dest => write(dest, bundle)),
  ...terses.map(dest => write(dest, terse))
]);
```

boneskull (approved):

packages/ses/scripts/bundle.js
leotm: updated 🏎️

PR #1856, opened 2023-11-09
docs: Formalize next-release label in process

kriskowal:
This change enshrines the role of the `next-release` label in our process for cutting Endo releases, ensuring that the releaser is reminded to check.


turadg (approved):
🙏

Issue #1858, opened 2023-11-10
`M.callWhen().optional(...)` causes rejections


michaelfig:
## Steps to reproduce

Many Agoric SDK unit tests are failing with the same kind of error.

```
not ok 36 - name-hub › nameAdmin provideChild %ava-dur=40ms
  ---
    name: AssertionError
    message: Rejected promise returned by test
    values:
      'Rejected promise returned by test. Reason:': |-
        TypeError {
          message: 'Cannot read properties of undefined (reading \'length\')',
        }
    at: 'async packages/vats/test/test-name-hub.js:185:17'
  ...
```

Investigating [`packages/vats/test/test-name-hub.js:185:17`](https://github.com/Agoric/agoric-sdk/blob/40edb74f0d41db8fdde74f55b09cf2eb09bceb06/packages/vats/test/test-name-hub.js#L185) leads eventually to the [following InterfaceGuard](https://github.com/Agoric/agoric-sdk/blob/40edb74f0d41db8fdde74f55b09cf2eb09bceb06/packages/vats/src/nameHub.js#L30).

Furthermore, this error could be reproduced by calling a method guarded by `M.callWhen().optional(M.arrayOf(M.any())).returns()`, namely a `callWhen` with an `optional` `arrayOf`.

## Expected behavior

No error.

## Platform environment

Agoric SDK CI

PR #1859, opened 2023-11-10
fix(exo): update `M.callWhen` broken by `M.raw()`

michaelfig:
closes: #1858

## Description

Correct the outdated implementation of `M.callWhen` that didn't account for changes made in #1831.

### Security Considerations

Needs careful examination, since method guards provide an essential barrier to protect against misuse.

### Scaling Considerations

n/a

n/a

### Testing Considerations

A small unit test to demonstrate the problem now passes with the updated code.

### Upgrade Considerations

n/a

### DIFF
--- a/packages/exo/test/test-heap-classes.js
+++ b/packages/exo/test/test-heap-classes.js
@@ -24,6 +24,19 @@ test('what happens with extra arguments', t => {
   exo.foo('an extra arg');
 });
 
+const OptionalArrayI = M.interface('OptionalArray', {
+  foo: M.callWhen().optional(M.arrayOf(M.any())).returns(),
+});
+
+test('callWhen-guarded method called without optional array argument', async t => {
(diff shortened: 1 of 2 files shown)


packages/exo/test/test-heap-classes.js
```diff
@@ -24,6 +24,19 @@ test('what happens with extra arguments', t => {
   exo.foo('an extra arg');
 });
 
+const OptionalArrayI = M.interface('OptionalArray', {
+  foo: M.callWhen().optional(M.arrayOf(M.any())).returns(),
+});
+
+test('what happens with callWhen undefined optional array', async t => {
```
gibson042, line 31:
```suggestion
test('callWhen-guarded method called without optional array argument', async t => {
```

gibson042 (approved):

PR #1860, opened 2023-11-11
feat(eventual-send): breakpoint on delivery by env-options

erights:
closes: #XXXX
refs: #XXXX

## Description

Given an environment variable setting of

```console
$ export ENDO_DELIVERY_BREAKPOINTS="quatloos issuer.getAmountOf=*"
```

then when `E` invokes a `getAmountOf` method on an object with @@toStringTag of `'Alleged: quatloos issuer'`, it should first execute a JavaScript `debugger;` statement. When run normally, i.e., not under a debugger, the `debugger;` statement has no effect. But when run under a debugger, this statement acts as a breakpoint, causing execution to pause there.

If the "*" at the end is replaced by a number, then each time execution matches, the number is decremented by 1. When it reaches zero, it executes `debugger;`.

If the "getAmountOf" is replaced with "*", then it breakpoints on any method of a quatloos issuer.

If the "quatloos issuer" is replaced with "*", then it breakpoints on calling a method named "getAmountOf" on anything. 

Similarly

```console
$ export ENDO_SEND_BREAKPOINTS="quatloos issuer.getAmountOf=*"
```

will breakpoint when the message is eventually sent, so that you can look up the stack in the debugger to understand why it was sent. The name only matches if the object itself, or a remote presence for the object, is the target and so has the correct @@toStringTag. This does not work when eventually sending to a promise. To catch those cases, use "Promise" or "*" for the target name.

### Security Considerations

As with all use of env-options, we should worry about whether an adversary can read or write these options, and so interfere with code that is supposed to use these options.

### Scaling Considerations

If the `ENDO_DELIVERY_BREAKPOINTS` env-option is not even set, this implementation takes a fast path which should have essentially no penalty compared to the absence of these mechanisms.

When `ENDO_DELIVERY_BREAKPOINTS` is present but has no "*" on the method name, we still try to minimize the overhead on non-matching method names. But once we have a method name match, then we do a linear search among these directives on each such call.

Likewise for `ENDO_SEND_BREAKPOINTS`

Let's get some road experience ourselves first. But if it proves useful, it'll be a great debugging aid to explain.

It will be interesting to figure out how to show how to use it interactively from inside a debugger well, including dynamically changing breakpoints. Probably a short video?

### Testing Considerations

Since this alters only the behavior seen from a debugger, it is unclear how to write an automated test.

### Upgrade Considerations

None.

### DIFF
--- a/packages/eventual-send/src/E.js
+++ b/packages/eventual-send/src/E.js
@@ -130,7 +130,6 @@ const makeESendOnlyProxyHandler = (recipient, HandledPromise) =>
       );
     },
     apply: (_target, _thisArg, argsArray = []) => {
-      HandledPromise.applyFunctionSendOnly(recipient, argsArray);
       if (onSend && onSend.shouldBreakpoint(recipient, undefined)) {
         // eslint-disable-next-line no-debugger
         debugger; // LOOK UP THE STACK
@@ -138,6 +137,7 @@ const makeESendOnlyProxyHandler = (recipient, HandledPromise) =>
         // so that you can walk back on the stack to see how we came to
         // make this eventual-send
       }
+      HandledPromise.applyFunctionSendOnly(recipient, argsArray);
       return undefined;
     },
     has: (_target, _p) => {
--- a/packages/eventual-send/src/E.js
+++ b/packages/eventual-send/src/E.js
@@ -1,7 +1,10 @@
 import { trackTurns } from './track-turns.js';
+import { makeMessageBreakpointTester } from './message-breakpoints.js';
 
 const { details: X, quote: q, Fail } = assert;
 const { assign, create } = Object;
+
+const onSend = makeMessageBreakpointTester('ENDO_SEND_BREAKPOINTS');
 
 /** @type {ProxyHandler<any>} */
 const baseFreezableProxyHandler = {
@@ -31,38 +34,55 @@ const baseFreezableProxyHandler = {
 /**
  * A Proxy handler for E(x).
  *
- * @param {any} x Any value passed to E(x)
+ * @param {any} recipient Any value passed to E(x)
  * @param {import('./types').HandledPromiseConstructor} HandledPromise
  * @returns {ProxyHandler} the Proxy handler
  */
-const makeEProxyHandler = (x, HandledPromise) =>
+const makeEProxyHandler = (recipient, HandledPromise) =>
   harden({
     ...baseFreezableProxyHandler,
-    get: (_target, p, receiver) => {
+    get: (_target, propertyKey, receiver) => {
       return harden(
         {
           // This function purposely checks the `this` value (see above)
           // In order to be `this` sensitive it is defined using concise method
           // syntax rather than as an arrow function. To ensure the function
           // is not constructable, it also avoids the `function` syntax.
-          [p](...args) {
+          [propertyKey](...args) {
             if (this !== receiver) {
               // Reject the async function call
               return HandledPromise.reject(
                 assert.error(
-                  X`Unexpected receiver for "${p}" method of E(${q(x)})`,
+                  X`Unexpected receiver for "${propertyKey}" method of E(${q(
+                    recipient,
+                  )})`,
                 ),
               );
             }
 
-            return HandledPromise.applyMethod(x, p, args);
+            if (onSend && onSend.shouldBreakpoint(recipient, propertyKey)) {
+              // eslint-disable-next-line no-debugger
+              debugger; // LOOK UP THE STACK
+              // Stopped at a breakpoint on eventual-send of a method-call
+              // message,
+              // so that you can walk back on the stack to see how we came to
+              // make this eventual-send
+            }
+            return HandledPromise.applyMethod(recipient, propertyKey, args);
           },
           // @ts-expect-error https://github.com/microsoft/TypeScript/issues/50319
-        }[p],
+        }[propertyKey],
       );
     },
     apply: (_target, _thisArg, argArray = []) => {
-      return HandledPromise.applyFunction(x, argArray);
+      if (onSend && onSend.shouldBreakpoint(recipient, undefined)) {
+        // eslint-disable-next-line no-debugger
+        debugger; // LOOK UP THE STACK
+        // Stopped at a breakpoint on eventual-send of a function-call message,
+        // so that you can walk back on the stack to see how we came to
+        // make this eventual-send
+      }
+      return HandledPromise.applyFunction(recipient, argArray);
     },
     has: (_target, _p) => {
       // We just pretend everything exists.
@@ -74,35 +94,50 @@ const makeEProxyHandler = (x, HandledPromise) =>
  * A Proxy handler for E.sendOnly(x)
  * It is a variant on the E(x) Proxy handler.
  *
- * @param {any} x Any value passed to E.sendOnly(x)
+ * @param {any} recipient Any value passed to E.sendOnly(x)
  * @param {import('./types').HandledPromiseConstructor} HandledPromise
  * @returns {ProxyHandler} the Proxy handler
  */
-const makeESendOnlyProxyHandler = (x, HandledPromise) =>
+const makeESendOnlyProxyHandler = (recipient, HandledPromise) =>
   harden({
     ...baseFreezableProxyHandler,
-    get: (_target, p, receiver) => {
+    get: (_target, propertyKey, receiver) => {
       return harden(
         {
           // This function purposely checks the `this` value (see above)
           // In order to be `this` sensitive it is defined using concise method
           // syntax rather than as an arrow function. To ensure the function
           // is not constructable, it also avoids the `function` syntax.
-          [p](...args) {
+          [propertyKey](...args) {
             // Throw since the function returns nothing
             this === receiver ||
-              Fail`Unexpected receiver for "${q(p)}" method of E.sendOnly(${q(
-                x,
-              )})`;
-            HandledPromise.applyMethodSendOnly(x, p, args);
+              Fail`Unexpected receiver for "${q(
+                propertyKey,
+              )}" method of E.sendOnly(${q(recipient)})`;
+            if (onSend && onSend.shouldBreakpoint(recipient, propertyKey)) {
+              // eslint-disable-next-line no-debugger
+              debugger; // LOOK UP THE STACK
+              // Stopped at a breakpoint on eventual-send of a method-call
+              // message,
+              // so that you can walk back on the stack to see how we came to
+              // make this eventual-send
+            }
+            HandledPromise.applyMethodSendOnly(recipient, propertyKey, args);
             return undefined;
           },
           // @ts-expect-error https://github.com/microsoft/TypeScript/issues/50319
-        }[p],
+        }[propertyKey],
       );
     },
     apply: (_target, _thisArg, argsArray = []) => {
-      HandledPromise.applyFunctionSendOnly(x, argsArray);
+      HandledPromise.applyFunctionSendOnly(recipient, argsArray);
+      if (onSend && onSend.shouldBreakpoint(recipient, undefined)) {
+        // eslint-disable-next-line no-debugger
+        debugger; // LOOK UP THE STACK
+        // Stopped at a breakpoint on eventual-send of a function-call message,
+        // so that you can walk back on the stack to see how we came to
+        // make this eventual-send
+      }
       return undefined;
     },
     has: (_target, _p) => {
--- a/packages/eventual-send/src/local.js
+++ b/packages/eventual-send/src/local.js
@@ -1,9 +1,13 @@
+import { makeMessageBreakpointTester } from './message-breakpoints.js';
+
 const { details: X, quote: q, Fail } = assert;
 
 const { getOwnPropertyDescriptors, getPrototypeOf, freeze } = Object;
 const { apply, ownKeys } = Reflect;
 
 const ntypeof = specimen => (specimen === null ? 'null' : typeof specimen);
+
+const onDelivery = makeMessageBreakpointTester('ENDO_DELIVERY_BREAKPOINTS');
 
 /**
  * TODO Consolidate with `isObject` that's currently in `@endo/marshal`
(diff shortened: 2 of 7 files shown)


dckc:
This looks like a new product feature. I'd like us to bring @sufyaankhan up to speed and get his take on it.

> Documentation Considerations
> Let's get some road experience ourselves first.

Before merging it? Or before documenting it?

erights:
> This looks like a new product feature. I'd like us to bring @sufyaankhan up to speed and get his take on it.

Absolutely! @sufyaankhan would love your thoughts.

> 
> > Documentation Considerations
> > Let's get some road experience ourselves first.
> 
> Before merging it? Or before documenting it?

Before documenting. We won't get enough road experience ourselves if we don't merge it first.

packages/eventual-send/src/local.js
```diff
@@ -13,6 +18,55 @@ const ntypeof = specimen => (specimen === null ? 'null' : typeof specimen);
...
+  if (bp !== '') {
+    const match = BP_REGEXP.exec(bp);
+    if (match === null) {
+      throw Fail`Expected "tagName.methodName=count" ${q(bp)}`;
+    }
+    const [_, tag, verb, countStr] = match;
+    const patts = hasOwn(BREAKS, verb) ? BREAKS[verb] : (BREAKS[verb] = []);
+    patts.push([tag, countStr === '*' ? countStr : +countStr]);
+  }
+}
```
dckc, line 41: `harden(BREAKS)` after this for loop?

dckc (commented):
I looked it over. I don't see any critical problems, but I haven't studied `makeEnvironmentCaptor` well enough to be comfortable with this yet. Nor have I tested it manually.

turadg (commented):
Since this is a tentative name/feature, please indicate that. E.g. `EXPERIMENTAL_BREAKPOINTS`. That makes clear that it isn't supported and may break.

When/if it is advanced from experimentation, the name should be scoped to what it does. It's not a way to express any breakpoint, just ones for remote calls. E.g. `E_CALL_BREAKPOINTS`. Or for more harmony with the extant `DEBUG=` option, `DEBUG_E_CALLS`.

packages/eventual-send/src/local.js
gibson042, line 41: Grouping primarily by method name/wildcard in appearance order and secondarily by object tag/wildcard in appearance order implies a particular specificity resolution such that e.g. `BREAKPOINTS=Alleged: Foo.*=1,Alleged: Foo.m=2,*.m=3,Alleged: Bar.m=4,*.m2=5` will result in `E(fooInstance).m()` using the 2 while `E(barInstance).m()` and `E(bazInstance).m()` use the 3 (which in particular completely supersedes the 4), and `E(fooInstance).m2()` using the 5. Prioritizing method name over object tag seems fine, but we probably want wildcards to be considered last in each group. For example:
```suggestion
    // eslint-disable-next-line no-nested-ternary
    const count = countStr === '*'
      ? countStr
      : countStr === '0'
        ? 0
        : (+countStr || Fail`invalid skip count in delivery breakpoint filter ${q(bp)}`);
    const patts = hasOwn(BREAKS, verb) ? BREAKS[verb] : (BREAKS[verb] = []);
    const i = tag === '*' ? -1 : patts.findIndex(patt => patt[0] === '*');
    if (i === -1) {
      patts.push([tag, count]);
    } else {
      patts.splice(i, 0, [tag, count]);
    }
  }
}
if (BREAKS['*']) {
  const wildcard = BREAKS['*'];
  delete BREAKS['*'];
  BREAKS['*'] = wildcard;
}
```

gibson042 (commented):
I think the concept has merit, but I agree that we'll want some experience with it to really decide.

packages/eventual-send/src/local.js
```diff
@@ -13,6 +18,55 @@ const ntypeof = specimen => (specimen === null ? 'null' : typeof specimen);
...
+                  // eslint-disable-next-line no-debugger
+                  debugger;
+                  break;
+                }
+                patt[1] = count - 1;
+              }
+            }
+          }
+        }
+        return apply(f, obj, args);
```
gibson042, line 67: Is there a benefit to this `Record<Verb, Array<[Tag, Count]>>` structure over `Record<Verb, Record<Tag, Count>>`?
```suggestion
        const countsByTag = BREAKS[verb] || BREAKS['*'];
        if (countsByTag !== undefined) {
          const objTag = obj[Symbol.toStringTag];
          const matchTag = hasOwn(countsByTag, objTag) ? objTag : '*';
          const count = countsByTag[matchTag];
          if (count === '*' || count === 0) {
            // eslint-disable-next-line no-debugger
            debugger;
          } else if (count !== undefined) {
            countsByTag[matchTag] = count - 1;
          }
        }
        return apply(f, obj, args);
```

packages/eventual-send/src/local.js
```diff
@@ -13,6 +18,55 @@ const ntypeof = specimen => (specimen === null ? 'null' : typeof specimen);
  */
 const isObject = val => Object(val) === val;
 
+const BREAKPOINTS = getEnvironmentOption('BREAKPOINTS', '');
```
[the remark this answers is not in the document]
michaelfig: I appreciate @turadg's concern over the naming of the `BREAKPOINTS` environment variable.  I think as long as it is prefixed with `ENDO_`, we can use whatever name we want without fear of confusion or stomping on somebody else's variables.

```suggestion
const BREAKPOINTS = getEnvironmentOption('ENDO_E_BREAKPOINTS', '');
```

packages/eventual-send/src/local.js
```diff
@@ -13,6 +18,55 @@ const ntypeof = specimen => (specimen === null ? 'null' : typeof specimen);
...
+    if (match === null) {
+      throw Fail`Expected "tagName.methodName=count" ${q(bp)}`;
+    }
+    const [_, tag, verb, countStr] = match;
+    const patts = hasOwn(BREAKS, verb) ? BREAKS[verb] : (BREAKS[verb] = []);
+    patts.push([tag, countStr === '*' ? countStr : +countStr]);
+  }
+}
+
+const debugApply =
```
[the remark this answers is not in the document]
erights: Better, I've removed the extra function call from the stack. Rather, I replaced it with a predicate call 
```js
if (shouldBreakpointDelivery(...) {
  breakpoint;
}
``` 
immediately before the real apply, which also creates a much more pleasant experience in the debugger.

As part of the refactoring, I did also give all these better names. Resolving.

packages/eventual-send/src/local.js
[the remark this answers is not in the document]
erights: The rename to `ENDO_DELIVERY_BREAKPOINTS` is done. This also made it easy to add `ENDO_SEND_BREAKPOINTS` for an additional feature which may be even more useful.

Not resolving this conversation yet, as I haven't yet made the suggested representation change

erights:
I have not yet made the suggested representation changes, so I reverted this PR to Draft.

packages/eventual-send/src/local.js
```diff
@@ -13,6 +18,55 @@ const ntypeof = specimen => (specimen === null ? 'null' : typeof specimen);
...
+ * @type {Record<string, [string, (number | '*')][]>}
+ */
+// @ts-expect-error doesn't like null __proto__
+const BREAKS = { __proto__: null };
+
+const BP_REGEXP = /(.+?)\.(\w+?|\*)=(\d+?|\*)/;
+
+for (const bp of BREAKPOINTS.split(',')) {
+  if (bp !== '') {
+    const match = BP_REGEXP.exec(bp);
```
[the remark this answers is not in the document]
erights: I took @michaelfig suggestion below to just go with a JSON string as the value of the option, completely avoiding the need for regexp matching. Resolving.

packages/eventual-send/src/local.js
erights: The data structure previously name `BREAKS` is now the encapsulated `breakpointsTable`. Its internal representation still needs reform, but this has not yet happened. With the current representation, I countdown by mutating the part containing the `count` number in place. So, not yet resolving, but noting to clarify further review.

erights: @gibson042 this internal representation is now encapsulated but is not yet reformed. So, not yet resolving, but noting to clarify further review.

@dckc , I refactored the decision about whether to breakpoint into a separate predicate function that does not itself do the breakpointing, but instead leaves that to its caller. As a result, I was able to write unit tests testing that predicate. PTAL.

packages/eventual-send/src/local.js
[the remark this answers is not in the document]
erights: I have made this suggested external representation change. But I have not yet made the suggested internal one. Nor have I yet done anything to support symbol-named methods. I like the idea of using a manually implemented tiny subset of smallcaps for this one case, so I will examine that.

Thus, not yet resolving.

erights:
Note to reviewers: Enough changed that it wasn't worth preserving old commits, so I squashed since you last reviewed.

packages/eventual-send/src/local.js
erights: Since the internal breakpoints table is encapsulated, I didn't freeze so I could continue to do this trick.

Separately, to support interactive change of breakpoints, I now make the external representation available from within the debugger. To ease modification, I separately do not freeze it, so it can be modified in place.

@dckc since your suggestion ended in a "?", I'll consider this suggestion done. But please unresolve this conversation if you disagree.

erights: With the new nested-record internal representation, I was more easily able to consistently always check the '*' cases last. Resolving.

packages/eventual-send/src/local.js
erights: Done

packages/eventual-send/src/local.js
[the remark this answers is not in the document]
erights: Shift to internal representation done.

I have not yet done anything about symbol-named methods. But I don't think that needs to happen in this PR. Resolving.

erights:
Reviewers,
Now that this is ready for review, I'll avoid further squashes without first asking you.

packages/eventual-send/src/E.js
```diff
@@ -74,35 +94,50 @@ const makeEProxyHandler = (x, HandledPromise) =>
...
             return undefined;
           },
           // @ts-expect-error https://github.com/microsoft/TypeScript/issues/50319
-        }[p],
+        }[propertyKey],
       );
     },
     apply: (_target, _thisArg, argsArray = []) => {
-      HandledPromise.applyFunctionSendOnly(x, argsArray);
+      HandledPromise.applyFunctionSendOnly(recipient, argsArray);
```
michaelfig, line 133: Any reason this `HandledPromise...` call appears *before* the `onSend` checking?  All the other calls to `HandledPromise` static methods appear to be *after* the `onSend`.

packages/eventual-send/src/message-breakpoints.js
```diff
@@ -0,0 +1,148 @@
...
+
+const { hasOwn, freeze, entries } = Object;
+
+/**
+ * This is the external JSON representation, in which
+ * - the outer property name is the class-like tag or '*',
+ * - the inner property name is the method name or '*',
+ * - the value is a non-negative integer countdown or '*'.
+ *
+ * @typedef {Record<string, Record<string, number | '*'>>} MessageBreakpoints
```
michaelfig, line 13: Maybe document in terms of types, like.

```js
/** @typedef {string | '*'} MatchStringTag docs... */
/** @typedef {string | '*'} MatchMethodName docs... */
/** @typedef {number | '*'} MatchCountdown docs... */ 
```

```suggestion
 * @typedef {Record<MatchStringTag, Record<MatchMethodName, MatchCountdown>>} MessageBreakpoints
```

packages/eventual-send/src/message-breakpoints.js
```diff
@@ -0,0 +1,148 @@
...
+ */
+
+/**
+ * This is the internal JSON representation, in which
+ * - the outer property name is the method name or '*',
+ * - the inner property name is the class-like tag or '*',
+ * - the value is a non-negative integer countdown or '*'.
+ *
+ * @typedef {Record<string, Record<string, number | '*'>>} BreakpointTable
```
michaelfig, line 22:
```suggestion
 * @typedef {Record<MatchMethodName, Record<MatchStringTag, MatchCountdown>>} BreakpointTable
```

packages/eventual-send/src/message-breakpoints.js
```diff
@@ -0,0 +1,148 @@
...
+ */
+const isJSONRecord = val =>
+  typeof val === 'object' && val !== null && !Array.isArray(val);
+
+/**
+ * @param {string} tag
+ * @returns {string}
+ */
+const simplifyTag = tag => {
+  for (const prefix of ['Alleged: ', 'DebugName: ']) {
```
michaelfig, line 47: Does it ever happen that a string tag has both prefixes?  If not, can you please document some examples of them in the jsdoc above the function?

packages/eventual-send/src/message-breakpoints.js
```diff
@@ -0,0 +1,148 @@
...
+        classBPs[tag] = count;
+      }
+    }
+    breakpoints = newBreakpoints;
+    breakpointsTable = newBreakpointsTable;
+  };
+  freeze(setBreakpoints);
+
+  const shouldBreakpoint = (recipient, methodName) => {
+    if (methodName === undefined) {
```
michaelfig, line 111: I think this should be:

```suggestion
    if (methodName === undefined || methodName === null) {
```

packages/eventual-send/src/message-breakpoints.js
michaelfig: Also, could you please describe explicitly what `MatchCountdown` does?  It is not obvious from reading the other jsdocs whether this is the number of noops until entering the debugger, or the number of debugger traps until it becomes a noop.

packages/eventual-send/src/E.js
erights: My mistake. Thanks for catching! Done.

packages/eventual-send/src/message-breakpoints.js
erights: Done.

packages/eventual-send/src/message-breakpoints.js
erights: Done

packages/eventual-send/src/message-breakpoints.js
erights: Done

packages/eventual-send/src/message-breakpoints.js
erights: Done

packages/eventual-send/src/message-breakpoints.js
```diff
@@ -0,0 +1,179 @@
+import { getEnvironmentOption } from '@endo/env-options';
+
+const { quote: q, Fail } = assert;
+
+const { hasOwn, freeze, entries } = Object;
+
+/**
+ * @typedef {string | '*'} MatchStringTag
+ *   A star `'*'` matches any recipient. Otherwise, the string is
+ *   matched against the value of a recipient's` @@toStringTag`
```
michaelfig, line 10:
```suggestion
 *   matched against the value of a recipient's `@@toStringTag`
```

packages/eventual-send/src/message-breakpoints.js
```diff
@@ -0,0 +1,179 @@
...
+ * @typedef {string | '*'} MatchMethodName
+ *   A star `'*'` matches any method name. Otherwise, the string is
+ *   matched against the method name. Currently, this is only an exact match.
+ *   However, beware that we may introduce a string syntax for
+ *   symbol method names.
+ */
+/**
+ * @typedef {number | '*'} MatchCountdown
+ *   A star `'*'` will always breakpoint. Otherwise, the string
+ *   must be a non-negative integer. Once zero, that always breakpoint.
```
michaelfig, line 25:
```suggestion
 *   must be a non-negative integer. Once that is zero, always breakpoint.
```

michaelfig (approved):
Some trivial doc fixes, and you may want a second opinion before merging, but LGTM!

packages/eventual-send/src/message-breakpoints.js
erights: Done

packages/eventual-send/src/message-breakpoints.js
erights: Done

PR #1864, opened 2023-11-22
fix: Import types explicitly throughout endo

kriskowal:
## Description

In the current `yarn pack` workflow, some type information for `export * from '@endo/top-level-name'` do not propagate from `.js` to `.d.ts` files. To compensate for this, in the publishing process, we must generate all `.d.ts` files for all packages in order from the least-dependent to most-dependent package, without deleting the generated `.d.ts` files between packages. This in turn locally breaks the current behavior of `import './types.js'`, where all the types from an exclusively-jsdoc module can be brought into another module’s namespace. This change makes all such imports explicit.

---

I am working toward getting a working integration with Agoric SDK. We are currently observing TypeScript errors as a consequence of recent work toward eliminating the need to configure TypeScript to traverse an arbitrary depth of `node_modules` trees to discover type information of transitive dependencies.

The ethos of TypeScript and the performance of type validation depends on each published artifact including `.d.ts` files that capture the type information gleaned from JSDoc on sister `.js` files. To that end, recent changes to Endo cause packages to generate these artifacts.

In this workflow, we lose the local ability to import all the type names from a `.js` file that only contains JSDoc and this is not evident until attempting to generate types from a dependent package. To compensate, we will need to also adjust our publish workflow. All of our `prepack` and `postpack` scripts will need to be renamed `prepublish` and `postpublish`, such that:

```console
> yarn lerna run prepbublish
> EXISTING PUBLISH WORKFLOW
> yarn lerna run postpublish
```

This will ensure that all necessary types will be generated correctly across the Endo workspace.

### Testing Considerations

With the new publish workflow, we can no longer rely entirely on `yarn lint` to discover the problem that this change addresses. To verify this change, one must run `yarn lerna run prepack` and this must succeed. I’ve altered CI to verify this workflow. We will also need to adjust the agoric-sdk endo sync workflow.


kriskowal:
This landed through a different PR. Closing to clean up.

PR #1868, opened 2023-11-30
chore: Improve the explanation of PassStyle "tagged"

gibson042:
_Originally posted by @erights in https://github.com/Agoric/documentation/pull/876#discussion_r1411146371_

> > _CopyTaggeds_ representing higher-order types
> 
> Not necessarily higher order in general. Probably the intended point is that these concepts are defined at a higher abstraction layer.


erights (approved):
LGTM thanks

PR #1871, opened 2023-12-01
fix(daemon): pet-store removes petname when overwriting with write

kumavis:
when overwriting a value in pet-store, we should remove the previous value.
This ensures we run proper cleanup:
- delete petname files
- remove known pet name for formula
- report deletion to subscribers

### DIFF
--- a/packages/daemon/src/pet-store.js
+++ b/packages/daemon/src/pet-store.js
@@ -77,6 +77,15 @@ export const makePetStoreMaker = (filePowers, locator) => {
         throw new Error(`Invalid formula identifier ${q(formulaIdentifier)}`);
       }
 
+      if (petNames.has(petName)) {
+        // Perform cleanup on the overwritten pet name.
+        const formulaPetNames = formulaIdentifiers.get(petName);
+        if (formulaPetNames !== undefined) {
+          formulaPetNames.delete(petName);
+        }
+        changesTopic.publisher.next({ remove: petName });
+      }
+
       petNames.set(petName, formulaIdentifier);
 
       const formulaPetNames = formulaIdentifiers.get(formulaIdentifier);


packages/daemon/src/pet-store.js
```diff
@@ -132,6 +107,35 @@ export const makePetStoreMaker = (filePowers, locator) => {
...
+     * @param {string} formulaIdentifier
+     */
+    const write = async (petName, formulaIdentifier) => {
+      assertValidName(petName);
+      if (!validFormulaPattern.test(formulaIdentifier)) {
+        throw new Error(`Invalid formula identifier ${q(formulaIdentifier)}`);
+      }
+
+      if (petNames.has(petName)) {
+        await remove(petName);
```
kriskowal, line 121: This `await` probably introduces a race condition. Can we avoid that by inlining the synchronous behavior of `remove`?

kriskowal: Or rather, an interleaving hazard for concurrent writes.

packages/daemon/src/pet-store.js
```diff
@@ -88,6 +99,9 @@ export const makePetStoreMaker = (filePowers, locator) => {
 
       const petNamePath = filePowers.joinPath(petNameDirectoryPath, petName);
       const petNameText = `${formulaIdentifier}\n`;
+      if (isOverwrite) {
+        await filePowers.removePath(petNamePath);
+      }
       await filePowers.writeFileText(petNamePath, petNameText);
```
kriskowal, line 105: Will `writeFileText` not also incidentally overwrite?

kriskowal (approved):
This should work. I think the `removePath` can be removed since the `writeFileText` will do the job. That suggests `isOverwrite` doesn’t need to be tracked.

PR #1872, opened 2023-12-01
fix(Daemon): node file powers write lock

kumavis:
use mutex to ensure serial file operations

all updates happen synchronously in memory
all disk operations are queued to ensure that the previous operation completed before continuing

### DIFF
--- a/packages/daemon/src/mutex.js
+++ b/packages/daemon/src/mutex.js
@@ -1,0 +1,30 @@
+import { makeQueue } from '@endo/stream';
+
+/**
+ * @returns {{ lock: () => Promise<void>, unlock: () => void, enqueue: (asyncFn: () => Promise<any>) => Promise<any> }}
+ */
+export const makeMutex = () => {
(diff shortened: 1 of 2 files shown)


kumavis:
it was said that we should push this into file powers

kriskowal:
> it was said that we should push this into file powers

It was me. I might have been wrong, and if I was wrong, it would be because doing so might complicate ejecting the powers into other Endo packages someday. But, since these powers are currently all internal, refactoring our way to brilliance is still fine and on the path of least regret.

packages/daemon/src/mutex.js
```diff
@@ -0,0 +1,49 @@
...
+      tailResolve({ value, promise });
+      tailResolve = next.resolve;
+    },
+    get() {
+      const promise = tailPromise.then(next => next.value);
+      tailPromise = tailPromise.then(next => next.promise);
+      return promise;
+    },
+  };
+};
```
kriskowal, line 21: Can be imported from `@endo/stream`

kriskowal (approved):

PR #1885, opened 2023-12-05
npx typedoc

turadg:
## Description

The job is running now but failing because the `typedoc` dep isn't installed: https://github.com/endojs/endo/actions/runs/7107676143/job/19349573778

This solves it with `npx typedoc`, instead of doing the full yarn install.

### DIFF
--- a/.github/workflows/typedoc-gh-pages.yml
+++ b/.github/workflows/typedoc-gh-pages.yml
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/configure-pages@v4
       # Generate the TypeDoc site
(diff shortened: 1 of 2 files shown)


kriskowal (approved):
For your consideration: I suspect that this means that new releases to `typedoc` can break our CI, since we’re no longer going to be pinned in `yarn.lock`.

turadg:
>For your consideration: I suspect that this means that new releases to typedoc can break our CI, since we’re no longer going to be pinned in yarn.lock.

They can break this publishing job, which is not a Required one to pass.

I'm more worried about putting garbage up at the web page, but if that ever happens it will be easy enough to clean up.

PR #1889, opened 2023-12-09
Agoric SDK integration fixes

kriskowal:
refs: https://github.com/Agoric/agoric-sdk/pull/8514

## Description

Various changes to Endo are necessary to fix integration with Agoric SDK such that Endo packages can be upgraded there.

The largest necessary change is that we must generate the types for each package based on the generated types of all dependencies. To do this, we replace prepack and postpack scripts with build:types and clean:types, then drive them from the top level before publishing.

Then, certain ambient types must be explicitly imported and the `.js` extension is required throughout.

We also needed to relax some types in Exo and Patterns.


packages/pass-style/src/make-far.js
```diff
@@ -179,7 +186,7 @@ export const Far = (farName, remotable = undefined) => {
     // * does further mutations,
     // * hardens the mutated object before returning it.
     // so this mutation is not unprecedented. But it is surprising!
-    r[GET_METHOD_NAMES] = getMethodNamesMethod;
+    Object.defineProperty(r, GET_METHOD_NAMES, getMethodNamesDescriptor);
```
kriskowal, line 189: attn @erights

packages/exo/src/get-interface.js
```diff
@@ -18,7 +18,7 @@ export const GET_INTERFACE_GUARD = '__getInterfaceGuard__';
 /**
  * @template {Record<PropertyKey, CallableFunction>} M
  * @typedef {{
- *   [GET_INTERFACE_GUARD]: () =>
+ *   [GET_INTERFACE_GUARD]?: () =>
```
kriskowal, line 21: attn @erights

kriskowal:
Tagged @michaelfig for the review so at least you know about the fix for publish. We should do the same in SDK to ensure dapps get correct types from npm.

CONTRIBUTING.md
```diff
@@ -156,6 +166,12 @@ https://github.com/endojs/endo/labels/next-release
   git tag -l | egrep -e '@[0-9]+\.[0-9]+\.[0-9]+$' | xargs git push origin
   ```
 
+* Clean up generated types.
+
+  ```sh
+  yarn lerna run glean:types
```
erights, line 172: Should this be

```suggestion
  yarn lerna run clean:types
```
?

packages/exo/src/get-interface.js
erights: As discussed, I think this is good enough. Resolving.

packages/pass-style/src/make-far.js
erights: Yes, non-enumerable is better. Resolving.

erights (approved):
LGETM -- looks good enough to me.

I will note that the change to importing all types inline at the point of use leads to the type notation being much less usable than before. However, I (almost) understand why this is necessary, and that we have no other reasonable choice.

You should probably still wait for @michaelfig 's review. My approval is just FWIW.

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -36,15 +36,13 @@ import {
...
-import './internal-types.js';
-
 /// <reference types="ses"/>
 
 const { quote: q, bare: b, details: X, Fail } = assert;
 const { entries, values } = Object;
 const { ownKeys } = Reflect;
 
-/** @type {WeakSet<Pattern>} */
+/** @type {WeakSet<import('./internal-types.js').Pattern>} */
```
michaelfig, line 45: I see you were trying to avoid intrusive changes, but this is too much verbose noise.  Can you please rename `./internal-types.js` to `./types.js` to follow the same filename convention as `eventual-send`  (internal=types.{js,d.ts}, external=exports.{js,d.ts})?

.github/workflows/ci.yml
```diff
@@ -401,6 +401,14 @@ jobs:
...
+        run: yarn lerna run generate:types
+
       - name: pack
         # Concurrency is 1 to avoid bizarre cross-package contamination.
-        run: yarn run lerna exec --concurrency=1 yarn pack
+        run: yarn lerna run pack
+
+      - name: clean:types
+        # Concurrency is 1 to avoid bizarre cross-package contamination.
+        run: yarn lerna run clean:types
```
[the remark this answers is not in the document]
michaelfig: Oh, and `yarn lerna run pack` does absolutely nothing if there is no explicit `packageJson.scripts.pack` lifecycle script:
```
$ yarn lerna run pack
yarn run v1.22.19
$ /Users/michael/agoric/endo/node_modules/.bin/lerna run pack
lerna notice cli v5.6.2
lerna info versioning independent
lerna success run No packages found with the lifecycle script 'pack'
✨  Done in 0.76s.
$
```
Please revert to `yarn lerna exec yarn pack` or simply just `yarn pack`.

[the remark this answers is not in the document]
kriskowal: Ahah! Thanks for the tip. I’m going to continue using `yarn lerna run` for `build:types` and `clean:types` since ignoring the step on packages that provide no such script is a useful behavior, and I’m reverting to your original lines for `pack`.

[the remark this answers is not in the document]
kriskowal: Also, I don’t understand the concurrency concern. It’s definitely the case that `build:types` needs to run from the bottom up, but it could do so concurrently. With `build:types` occurring in a separate step, we could probably safely pack concurrently as well. To that end, I’m going to remove the concurrency limit and comment; please let me know if they remain necessary.

packages/patterns/src/patterns/patternMatchers.js
kriskowal: Sure, I can line this up in this change with @erights’s blessing. It’ll be a separate commit.

kriskowal: I do not foresee an objection, so I’ve merged for expedience. Do let me know if I should revert the refactor of internal-types.js to merely types.js.

Issue #1846, opened 2023-11-02
Babel versions circa 7.20 are incompatible with Hardened JavaScript


kriskowal:
## Describe the bug

Refs: https://github.com/Agoric/agoric-sdk/issues/8458
Refs: https://github.com/endojs/endo/pull/1842

New contract packages are encountering a new Hardened JavaScript incompatibility (property override mistake) when attempting to use `@endo/bundle-source` to generate contract bundles.

One of the following errors appears:

```
(TypeError#1)
TypeError#1: Cannot assign to read only property 'constructor' of object '[object Object]'
  at _interopRequireWildcard (/home/x/projects/x/contract/node_modules/@babel/highlight/lib/index.js:12:710)
  at Object.<anonymous> (/home/x/projects/x/contract/node_modules/@babel/highlight/lib/index.js:10:14)
  at Module._compile (node:internal/modules/cjs/loader:1256:14)
  at module.exports.o.a._extensions..js (/home/x/projects/xcontract/node_modules/esm/esm.js:1:102412)
  at Module.load (node:internal/modules/cjs/loader:1119:32)
  at Module._load (node:internal/modules/cjs/loader:960:12)
  at Module.require (node:internal/modules/cjs/loader:1143:19)
```

Or:

```
TypeError#1: Cannot assign to read only property 'constructor' of object '[object Object]'
  at _interopRequireWildcard (/home/x/projects/x/contract/node_modules/@babel/code-frame/lib/index.js:11:710)
  at Object.<anonymous> (/home/x/projects/x/contract/node_modules/@babel/code-frame/lib/index.js:9:14)
  at Module._compile (node:internal/modules/cjs/loader:1256:14)
  at module.exports.o.a._extensions..js (/home/x/projects/x/contract/node_modules/esm/esm.js:1:102412)
  at Module.load (node:internal/modules/cjs/loader:1119:32)
  at Module._load (node:internal/modules/cjs/loader:960:12)
  at Module.require (node:internal/modules/cjs/loader:1143:19)
  at require (node:internal/modules/cjs/helpers:121:18)
  at Object.<anonymous> (/home/x/projects/x/contract/node_modules/@babel/traverse/lib/path/replacement.js:12:18)
  at Module._compile (node:internal/modules/cjs/loader:1256:14)
  at module.exports.o.a._extensions..js (/home/x/projects/x/contract/node_modules/esm/esm.js:1:102412)
  at Module.load (node:internal/modules/cjs/loader:1119:32)
[...]

PR #1893, opened 2023-12-11
Upgrade Babel 7.23

kriskowal:
closes: #1846

## Description

This should put #1846 to bed. The latest patch releases for Babel have addressed the property override mistake observed while bundling under lockdown.

As evidence, integrating this branch into Agoric SDK https://github.com/Agoric/agoric-sdk/pull/8514 obviates the need to run the bundler with overrideTaming: severe.

### DIFF
--- a/packages/ses/package.json
+++ b/packages/ses/package.json
@@ -37,14 +37,12 @@
     ".": {
       "types": "./types.d.ts",
       "import": "./index.js",
-      "require": "./dist/ses.cjs",
-      "types": "./types.d.ts"
+      "require": "./dist/ses.cjs"
     },
     "./lockdown": {
       "types": "./types.d.ts",
       "import": "./index.js",
-      "require": "./dist/ses.cjs",
-      "types": "./types.d.ts"
+      "require": "./dist/ses.cjs"
     },
     "./tools.js": "./tools.js",
     "./package.json": "./package.json"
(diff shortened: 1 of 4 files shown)


kriskowal:
> Nit title: 7.23 not 2.23

Fixed

image

packages/ses/package.json
```diff
@@ -37,14 +37,12 @@
     ".": {
       "types": "./types.d.ts",
       "import": "./index.js",
-      "require": "./dist/ses.cjs",
-      "types": "./types.d.ts"
```
mhofman, line 41: The "types" changes seem unrelated

mhofman (changes requested):
Please run a `npx yarn-deduplicate`

packages/ses/package.json
kriskowal: It is. This was a byproduct of running `./scripts/set-versions.sh`, which went ahead and removed the duplicate key from this `package.json`.

kriskowal: (and I’m retaining an individual commit for this change so it’s easy to identify/remove/move. I’ll accommodate your preference. Mine is to preserve the commit and save time.)

mhofman: Oh I missed the duplication!

mhofman (approved):
After dedup, the diff of the yarn.lock looks as expected. If tests pass, we should be good to go.

PS: to double check, you ran `yarn install` after `npx yarn-deduplicate` to make sure the `yarn.lock` was stable?

kriskowal:
> After dedup, the diff of the yarn.lock looks as expected. If tests pass, we should be good to go.
> 
> PS: to double check, you ran `yarn install` after `npx yarn-deduplicate` to make sure the `yarn.lock` was stable?

I did not and it was not stable. Please see the new commit.

PR #1894, opened 2023-12-11
chore: Primary CI on Node.js 18

kriskowal:
closes: #2035 

## Description

Move most CI to Node.js LTS 18.

Removes bundler integration tests. These have become stale enough that it will be easier to reconstruct them on modern foundations, using a modern sample of bundlers, than it is to maintain the existing ones, which make extensive use of `node -r esm`.

I would like to defer reconstruction of the bundler integration tests per https://github.com/endojs/endo/issues/2036


.github/workflows/browser-test.yml
```diff
@@ -0,0 +1,31 @@
...
+        with:
+          node-version: 18
+      - name: Install dependencies
+        run: yarn install
+      - name: Build artifacts
+        run: yarn build
+      - name: Install browser test dependencies
+        run: cd browser-test; npm ci
+      - name: Install Playwright Browsers
+        run: cd browser-test; npx playwright install --with-deps
```
turadg, line 23: more readable than cd
```suggestion
        run: playwright install --with-deps
        working-directory: browser-test
```

turadg (approved):
Changes LGTM. Glad to see EOL dropped. I defer to your judgement on deferring the bundler integration tests.

Left a minor CI suggestion. If you rebase commits, consider using `style:` for style changes so they're more obvious to ignore.

PR #1895, opened 2023-12-12
chore: Release 2023-12

kriskowal:
Please consider the following update in preparation to publish new versions of all public Endo packages. This release bumps the major version on all packages due to changes in the way types are generated and published that we found to have far-reaching implications for compatibility. Upgrading will likely require some adjustments to TypeScript annotations in dependent packages.

We also patched a defect in interface guards that causes previously-unenforced invariants to be enforced. This is likely to reveal defects in existing interface guards, and is consequently a breaking change (although it _fixes_ the feature!)


patches/@lerna+conventional-commits+5.6.2.patch
michaelfig, line 1: Why is this patch being dropped?  Are you guaranteeing we will never add another package with a `0.x.y` version?

michaelfig (approved):
Ready to land after my questions are addressed.

patches/@lerna+conventional-commits+5.6.2.patch
kriskowal: I’ll land this without the commit that removes the patch and with a commit that reverts the hack.

PR #1899, opened 2023-12-14
refactor: Change unsafe import formula from path to specifier

kriskowal:
In the pet daemon, workers currently require a power to convert paths to URL specifiers. This creates undue coupling to the underlying platform. A module specifier should suffice in this place. This generalizes well between daemon-on-web and daemon-on-node.

### DIFF
--- a/packages/cli/src/make.js
+++ b/packages/cli/src/make.js
@@ -2,6 +2,7 @@
 
 import os from 'os';
 import path from 'path';
+import url from 'url';
 
 import bundleSource from '@endo/bundle-source';
 import { makeReaderRef } from '@endo/daemon';
@@ -68,7 +69,7 @@ export const makeCommand = async ({
       importPath !== undefined
         ? E(party).importUnsafeAndEndow(
             workerName,
-            path.resolve(importPath),
+            url.pathToFileURL(path.resolve(importPath)).href,
             powersName,
             resultName,
           )
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -219,14 +219,14 @@ const makeEndoBootstrap = (
    * @param {string} valueFormulaIdentifier
    * @param {string} workerFormulaIdentifier
    * @param {string} guestFormulaIdentifier
-   * @param {string} importPath
+   * @param {string} specifier
    * @param {import('./types.js').Terminator} terminator
    */
   const makeControllerForUnsafePlugin = async (
     valueFormulaIdentifier,
     workerFormulaIdentifier,
     guestFormulaIdentifier,
-    importPath,
+    specifier,
     terminator,
   ) => {
     terminator.thisDiesIfThatDies(workerFormulaIdentifier);
@@ -249,7 +249,7 @@ const makeEndoBootstrap = (
       provideValueForFormulaIdentifier(guestFormulaIdentifier)
     );
     const external = E(workerDaemonFacet).importUnsafeAndEndow(
-      importPath,
+      specifier,
       guestP,
     );
     return { external, internal: undefined };
@@ -329,7 +329,9 @@ const makeEndoBootstrap = (
         formulaIdentifier,
         formula.worker,
         formula.powers,
-        formula.importPath,
+        formula.specifier ??
+          // @ts-expect-error
+          formula.importPath, // (TODO deprecated)
         terminator,
       );
     } else if (formula.type === 'import-bundle') {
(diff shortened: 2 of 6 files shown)


packages/cli/src/make.js
```diff
@@ -68,7 +69,7 @@ export const makeCommand = async ({
       importPath !== undefined
         ? E(party).importUnsafeAndEndow(
             workerName,
-            path.resolve(importPath),
+            url.pathToFileURL(path.resolve(importPath)).href,
```
dckc, line 72: Does this change behavior? Do any of the tests show how?

packages/daemon/src/worker.js
```diff
@@ -55,12 +50,11 @@ export const makeWorkerFacet = ({
...
     /**
-     * @param {string} path
+     * @param {string} specifier
      * @param {unknown} powersP
      */
-    importUnsafeAndEndow: async (path, powersP) => {
-      const url = pathToFileURL(path);
-      const namespace = await import(url);
+    importUnsafeAndEndow: async (specifier, powersP) => {
+      const namespace = await import(specifier);
```
dckc, line 57: Here I suppose you're putting `specifier` into `import`, so you don't need to resolve it first, so this isn't all that relevant, but...

Based on a clue from @michaelfig , I have migrated from using `resolve` on `path` to `createRequire(...)`.
For example:

https://github.com/Agoric/dapp-offer-up/blob/58d5ee78e302bfbcb2696375be57d9a19a0fca41/contract/test/test-bundle-source.js#L9-L15

dckc (commented):
just a question and an observation...

packages/cli/src/make.js
kriskowal: No behavior change, but if this were released, it would be a breaking change since the CLI and daemon would have to be updated together for it to continue working. This also will eventually require an `endo reset` since it changes the formula schema.

This is a coördinated change between the CLI and the daemon, such that the daemon is portable between web and node, both of which can dynamically import URLs. The CLI does the path to URL conversion instead of the daemon.

packages/daemon/src/worker.js
kriskowal: Interesting. By way of comparison:

- `path.resolve` will take a path and create a fully qualified path, which Node.js will accept as a module specifier for purposes of `require` or `import`.
- `createRequire(import.meta.url)` will give you a `require.resolve` which accepts CommonJS module specifiers and so distinguishes `./local-path.js` from 'in-your-node-modules` and uses the CommonJS module loader semantics.
- `import.meta.resolve` is intended to replace `require.resolve` and surface the Node.js 16+ ESM module resolution, respecting `"exports"` and `"imports"` in `package.json`. It’s shimmed with the `import-meta-resolve` package in npm, but that’s async. When `import.meta.resolve` is broadly supported, it will be synchronous because it benefits from the `package.json` cache Node.js builds while loading the surrounding module.

For this particular case, the Endo Pet Dæmon needs a module specifier as specific as a fully qualified path `file:///` URL since the Dæmon and the plugin are installed in separate locations and should have completely disjoint `package.json` graphs. So, there’s no expectation that the pet dæmon will be able to figure out where `@dckc/plugin` was installed.

packages/daemon/src/daemon.js
```diff
@@ -329,7 +329,9 @@ const makeEndoBootstrap = (
         formulaIdentifier,
         formula.worker,
         formula.powers,
-        formula.importPath,
+        formula.specifier ??
+          // @ts-expect-error
+          formula.importPath, // (TODO deprecated)
```
kriskowal, line 334: This is me trying to be gentle about the need for `endo reset`. All new formulas will use the new format, and the new version can read old formulas. I am not promising forward-compatibility yet, but it was easy to be backward-compatible for this round.

packages/daemon/test/test-endo.js
```diff
@@ -444,7 +445,13 @@ test('direct termination', async t => {
   await E(host).provideWorker('worker');
 
   const counterPath = path.join(dirname, 'test', 'counter.js');
-  await E(host).importUnsafeAndEndow('worker', counterPath, 'NONE', 'counter');
+  const counterLocation = url.pathToFileURL(counterPath).href;
```
kriskowal, line 448: @dckc This test demonstrates the shift in responsibility for resolving a path to a URL from the server to the client.

packages/cli/src/make.js
dckc: > The CLI does the path to URL conversion instead of the daemon.

hm. That sounds like a behavior change to me.

packages/daemon/test/test-endo.js
dckc: ah. good to see test coverage.

Perhaps my question wasn't phrased clearly, but since it's observable from a test, it's certainly a change in behavior, from my way of looking at things.

kumavis:
merged in https://github.com/endojs/endo/pull/2067

PR #1904, opened 2023-12-18
fix: commit auto-generated yarn.lock

erights:
closes: #XXXX
refs: #XXXX

## Description

The yarn.lock changes in this PR were purely automatically generated from a clean master by running `yarn`. No other changes.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -743,6 +743,41 @@
   resolved "https://registry.yarnpkg.com/@bcoe/v8-coverage/-/v8-coverage-0.2.3.tgz#75a2e8b51cb758a7553d6804a5932d7aace75c39"
   integrity sha512-0hYQ8SB4Db5zvZB4axdMHGwEaQjkZzFjQiN9LVYvIFB2nSUHW9tYpxWriPrWDASIxiaXax83REcLxuSdnGPZtw==
 
+"@endo/env-options@^0.1.4":
+  version "0.1.4"
+  resolved "https://registry.yarnpkg.com/@endo/env-options/-/env-options-0.1.4.tgz#e516bc3864f00b154944e444fb8996a9a0c23a45"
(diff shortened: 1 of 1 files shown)


yarn.lock
```diff
@@ -743,6 +743,41 @@
   resolved "https://registry.yarnpkg.com/@bcoe/v8-coverage/-/v8-coverage-0.2.3.tgz#75a2e8b51cb758a7553d6804a5932d7aace75c39"
   integrity sha512-0hYQ8SB4Db5zvZB4axdMHGwEaQjkZzFjQiN9LVYvIFB2nSUHW9tYpxWriPrWDASIxiaXax83REcLxuSdnGPZtw==
 
+"@endo/env-options@^0.1.4":
+  version "0.1.4"
+  resolved "https://registry.yarnpkg.com/@endo/env-options/-/env-options-0.1.4.tgz#e516bc3864f00b154944e444fb8996a9a0c23a45"
+  integrity sha512-Ol8ct0aW8VK1ZaqntnUJfrYT59P6Xn36XPbHzkqQhsYkpudKDn5ILYEwGmSO/Ff+XJjv/pReNI0lhOyyrDa9mg==
```
kriskowal, line 749: This is suspicious. It suggests that something has a `package.json` that is pinned to a version of this package that is not in the workspace.

I would like to try this alternative and see whether `package.json`s change or `yarn.lock` changes.

```console
> ./scripts/get-versions.sh . | ./scripts/set-versions.sh
> yarn
```

kriskowal (changes requested):
I would like to investigate these changes more deeply.

yarn.lock
erights: Looks the same after starting over from a fresh master and then running those two commands.

kriskowal:
I’ve reproduced these `yarn.lock` changes locally and also executed the commands I recommended above and found _no changes_ to `package.json`, so I am _perplexed_. Stay tuned.

kriskowal:
Figured it out. The problem is that Jessie.js is out of sync.

kriskowal (approved):
Let’s land this since it at least makes sense that Jessie needs older versions. I’m separately getting Jessie in sync, which will undo these changes. https://github.com/endojs/Jessie/pull/113

PR #1905, opened 2023-12-19
Avoid null coalescing and optional chaining

kriskowal:
Agoric’s load generator test uses `node -r esm` which doesn’t support the `??` nor `?.` operators.

This change introduces a lint rule to help us avoid this trip-fall and then ensures that existing packages pass the new rules.

### DIFF
--- a/packages/bundle-source/src/is-entrypoint.js
+++ b/packages/bundle-source/src/is-entrypoint.js
@@ -3,7 +3,14 @@ import url from 'url';
 import process from 'process';
 import fs from 'fs';
 
+// Agoric still uses Endo dependencies under an emulation of ESM we call RESM
+// because it is invoked with `node -r esm`.
+// RESM does not support ?? nor ?. operators, so we must avoid them expressly.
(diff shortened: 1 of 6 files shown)


packages/bundle-source/src/is-entrypoint.js
```diff
@@ -3,7 +3,10 @@ import url from 'url';
 import process from 'process';
 import fs from 'fs';
 
+const favor = (primary, secondary) =>
+  typeof primary === 'undefined' ? secondary : primary;
```
erights, line 7:
```suggestion
  primary === undefined ? secondary : primary;
```

Since we statically see that `primary` is a variable in lexical scope, your code and my suggestion must be equivalent. The only reason to use the `typeof name === 'undefined'` instead is if `name` might not be a variable name in scope and we wish to avoid throwing a `ReferenceError`.

erights (approved):
Please leave TODOs or some such behind in these locations, as markers for what code should be restored to a more readable state when we can.

LGTM, thanks!

PR #1907, opened 2023-12-20
fix: Expressly forbid deep imports through captp, far, lockdown, marshal

kriskowal:
To limit exposure to Hyrum’s Law, this change introduces an `"exports"` property to the four packages in Endo that currently lack one. Node.js 16+ and Endo’s bundler forbid deep imports to modules not expressly mentioned in `"exports"`.


erights (approved):
LGTM, thanks!

PR #1908, opened 2023-12-20
fix(eslint-plugin): Relax null coalescing and optional chaining to wa…

kriskowal:
…rning.

In #1905 we introduced lint rules to appease the one use we still have for `node -r esm` in Agoric’s stack (the testnet load generator). We found that these rules were useful for identifying usage that would not work under `node -r esm` but that the rule caused to many false positives in Agoric SDK. This change relaxes these rules to issue warnings.


erights (approved):
LGTM, thanks!

PR #1909, opened 2023-12-20
fix(evasive-transform): RESM treatment

kriskowal:
Agoric still uses an ESM emulation through the pattern `node -r esm` in its testnet load generator, which entrains Endo’s bundler. This change addresses “RESM” compatibility regressions with the recent creation of the `@endo/evasive-transform`, which is not tested under RESM.

### DIFF
--- a/packages/evasive-transform/src/generate.js
+++ b/packages/evasive-transform/src/generate.js
@@ -4,13 +4,17 @@
  * @module
  */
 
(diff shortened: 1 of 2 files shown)


kriskowal:
fyi @boneskull, we rediscovered why we have these tortured patterns in bundle-source.

erights (approved):
LGTM, thanks!

PR #1911, opened 2023-12-20
chore: release

kriskowal:
This release includes fixes to support RESM, and to lint with warning for patterns RESM does not support.

The version of the eslint plugin is going from 1 to 2. There is a commit that introduces two new lint rules and reports them as errors. This was appropriately marked as a breaking change. In a later commit, I backed it down to warn, retroactively making this new release a minor upgrade. Our workflow does not accommodate manually fixing conventional commits. I’m not inclined to fix this at this time. I’m investigating changesets, which provide a mechanism for intervening. Attn @michaelfig


kriskowal:
I also removed a dependency on jessie.js in bundle-source, which was intended for a lint job that has gotten stale, instead of releasing a new version of the jessie stdlib. This eliminates a circular dependency and cleans up our yarn.lock. We should take a run at Jessie linting again some time.

erights (approved):
LGTM!!!!

Issue #1686, opened 2023-07-19
Terminate CapTP in non-hardened Realm without SES shim


kriskowal:
@danfinlay raises an issue we have struggled with: it’s currently not possible to terminate CapTP in a realm that does not also use the SES shim, because these layers depend upon `harden`. We have yet to find a nice compromise that allows us to write these libraries in a way that works well in both hardened and unhardened JavaScript without compromising the readability or audit-ability of that code in the hardened case. Some of the difficulty lies in the dichotomy:

- `harden` walks up prototype chains, so hardening an object before lockdown would cause the repair phase to throw errors
- `harden` that does not walk up the prototype chain does not provide assurance that the object is transitively hardened

However, it is clearly the responsibility of the creator of instances to harden the instance and also clearly the responsibility of a library to harden its classes, constructors, prototypes, return values, and so on. Currently, `harden` does both as a safeguard against a library that failed to harden itself.

So, I propose we can break this logjam with the following strategy:

- We recreate an `@endo/harden` user-space implementation of `harden` that is a ponyfill for `globalThis.harden` and otherwise provides an implementation that freeze an object by walking transitive properties and _not_ prototype chains. This is a version of `harden` that is suitable for use in both unhardened JavaScript and hardened JavaScript.
- We also change `globalThis.harden` so that it freezes exactly the same objects as the ponyfill, and asserts that every hardened object is transitively frozen over both property and prototype, throwing an error that indicates that the hardened instance must stand on top of only hardened prototypes.

This would allow us to refactor CapTP and all its dependencies to depend on the `@endo/harden` ponyfill, making CapTP usable without shims, equally safe with shims, and otherwise identical verbatim to the current implementation.

kriskowal:
For review by @erights and @michaelfig

mhofman:
So there is the question of getters and setters in this split world: who is responsible for hardening them?

I believe that in most cases these are not shared with other objects, and that a "surface" harden should thus freeze them as part of the "walking transitive properties" step.

Btw in this world, it's not really the `harden` implementation that changes , but whether `isHardened()` recursively checks the prototype chain or not. In fact `isHardened()` could skip the prototype walk before `lockdown()`, but enforce it after. Taken an other way, `lockdown()` replaces `isHardened()` to add a recursive hardening check for the prototype.

gibson042:
[...]

PR #1914, opened 2023-12-29
[WIP] Shallow harden

mhofman:
refs: #1686

### DIFF
--- a/packages/ses/src/lockdown.js
+++ b/packages/ses/src/lockdown.js
@@ -27,8 +27,9 @@ import {
   noEvalEvaluate,
   getOwnPropertyNames,
   getPrototypeOf,
+  deleteProperty,
 } from './commons.js';
-import { makeHardener } from './make-hardener.js';
+import { makeHardenerKit } from './make-hardener.js';
 import { makeIntrinsicsCollector } from './intrinsics.js';
 import whitelistIntrinsics from './permits-intrinsics.js';
 import tameFunctionConstructors from './tame-function-constructors.js';
@@ -52,7 +53,7 @@ import tameErrorConstructor from './error/tame-error-constructor.js';
 import { assert, makeAssert } from './error/assert.js';
 import { getAnonymousIntrinsics } from './get-anonymous-intrinsics.js';
 import { makeCompartmentConstructor } from './compartment.js';
-import { tameHarden } from './tame-harden.js';
+import { tameHardenKit } from './tame-harden.js';
 import { tameSymbolConstructor } from './tame-symbol-constructor.js';
 
 /** @typedef {import('../types.js').LockdownOptions} LockdownOptions */
@@ -65,14 +66,15 @@ let priorRepairIntrinsics;
 /** @type {Error=} */
 let priorHardenIntrinsics;
 
+// @ts-expect-error harden always defined
+if (globalThis.harden) {
+  deleteProperty(globalThis, 'harden');
+}
+
 // Build a harden() with an empty fringe.
 // Gate it on lockdown.
-/**
- * @template T
- * @param {T} ref
- * @returns {T}
- */
-const safeHarden = makeHardener();
+/** @type {import('./make-hardener.js').HardenerKit} */
+const safeHardenKit = makeHardenerKit();
 
 /**
  * @callback Transform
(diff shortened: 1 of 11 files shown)


packages/ses/src/lockdown.js
```diff
@@ -64,14 +66,15 @@ let priorRepairIntrinsics;
 /** @type {Error=} */
 let priorHardenIntrinsics;
 
+// @ts-expect-error harden always defined
+if (globalThis.harden) {
+  deleteProperty(globalThis, 'harden');
```
gibson042, line 71: Hmm, what happens with captured references to the pre-lockdown `harden`?

packages/ses/src/lockdown.js
```diff
@@ -300,6 +315,16 @@ export const repairIntrinsics = (options = {}) => {
     optGetStackString,
   );
   globalThis.console = /** @type {Console} */ (consoleRecord.console);
+  // eslint-disable-next-line no-underscore-dangle
+  if (typeof (/** @type {any} */ (globalThis.console)._times) === 'object') {
+    hostIntrinsics.SafeMap = getPrototypeOf(
```
gibson042, line 320: This probably needs more explanation.

packages/ses/src/make-hardener.js
```diff
@@ -123,28 +123,41 @@ const freezeTypedArray = array => {
...
-     * @returns {T}
-     */
+  /** @type {HardenerKit} */
+  const { harden, isHardened, hardenIntrinsics } = {
+    isHardened(value) {
+      return weaksetHas(hardened, value);
+    },
+    hardenIntrinsics(root) {
+      intrinsicsWereHardened = true;
+      return harden(root);
```
gibson042, line 155: Can `harden` throw? Might need a`finally` here.

gibson042 (commented):
Quick review on mobile, I like where this is headed!

packages/ses/src/lockdown.js
mhofman: Yeah, we need to figure out the story for this. Right now I'm simply removing any harden found, like in the case of our xsnap-worker.

packages/ses/src/lockdown.js
mhofman: Oh yeah, just getting it to pass first. But basically NodeJS has hidden intrinsics, supposed to be internal only but one of them leaks through an "underscore" property of their console.

packages/ses/src/make-hardener.js
mhofman: Yes it can, and no need for finally. The goal for now is to leave harden permanently broken. We can discuss whether we want this situation to be resolvable somehow.

packages/ses/src/make-hardener.js
```diff
@@ -225,23 +260,65 @@ export const makeHardener = () => {
...
+                `Not all expected prototypes were hardened (expected=${hardenedAtLockdownSize}, got=${actuallyHardenedIntrinsicsSize})`,
+              );
+            } else {
+              hardenedAtLockdownSize = 0;
+              setForEach(actuallyHardenedIntrinsics, value => {
+                setDelete(hardenedAtLockdown, value);
+              });
+            }
+          }
+        } else if (setGetSize(protosToCheck) > 0) {
```
mhofman, line 299: If intrinsics were note hardened, we should fail for any objects we intend to harden, but that were already in the `hardenedAtLockdown` set. (enforcing the rule that a prototype must be hardened before its instances)

erights (commented):
The vast majority of of changes in this PR are just generic improvements in use of `harden`. If you could separate these into a distinct PR, we could probably review and merge that quickly. After that, this PR would also be much smaller and easier to review.

mhofman:
@erights the use of harden are already in separate commits, but hold off anyway, I'm more thoroughly rewriting the harden implementation to add hooks.

erights:
What about https://github.com/Agoric/agoric-sdk/pull/8700 ? It, likewise, seems to be almost all improvements in the use of harden which would be valuable regardless, and could get merged early. Yes?

mhofman:
> The vast majority of of changes in this PR are just generic improvements in use of `harden`. If you could separate these into a distinct PR, we could probably review and merge that quickly.

@erights I have now done that in https://github.com/endojs/endo/pull/1939. While I have rebased this PR on top of that change (roughly), I do expect to rewrite it significantly to add hooks to fully separate `harden` from `lockdown`

erights (commented):
Now that #1939 is merged, please rebase this one for review brevity. Thx.

mhofman:
> Now that #1939 is merged, please rebase this one for review brevity. Thx.

It's stacked by apparently GH is struggling and didn't update the base ...

erights:
I think if you `rebase -i` onto current (post merge) master and `push --force`, that should clean it up. yes?

erights:
Is this subsumed by @kriskowal 's recent PR along the same lines? Should we close this PR in favor of that one?

PR #1915, opened 2023-12-29
feat(daemon): Support pet name path lookup

kriskowal:
This change allows dot delimited pet name paths to follow references through `lookup` and establishes a protocol where any object can serve as a name hub if it implements simple pet name `lookup`.

### DIFF
--- a/packages/daemon/src/mail.js
+++ b/packages/daemon/src/mail.js
@@ -1,5 +1,6 @@
 // @ts-check
 
+import { E } from '@endo/far';
 import { makePromiseKit } from '@endo/promise-kit';
 import { makeChangeTopic } from './pubsub.js';
 import { makeIteratorRef } from './reader-ref.js';
@@ -41,16 +42,21 @@ export const makeMailboxMaker = ({
     };
 
     /**
-     * @param {string} petName
+     * @param {string} petNamePath
      */
-    const lookup = async petName => {
-      const formulaIdentifier = lookupFormulaIdentifierForName(petName);
+    const lookup = async petNamePath => {
+      const [headName, ...tailNames] = petNamePath.split('.');
+      const formulaIdentifier = lookupFormulaIdentifierForName(headName);
       if (formulaIdentifier === undefined) {
-        throw new TypeError(`Unknown pet name: ${q(petName)}`);
+        throw new TypeError(`Unknown pet name: ${q(headName)}`);
       }
       // Behold, recursion:
       // eslint-disable-next-line no-use-before-define
-      return provideValueForFormulaIdentifier(formulaIdentifier);
+      const value = provideValueForFormulaIdentifier(formulaIdentifier);
+      return tailNames.reduce(
+        (directory, petName) => E(directory).lookup(petName),
+        value,
+      );
     };
 
     const terminate = async petName => {


dckc:
I'd expect this to work in `endo eval` CLI bindings. It doesn't seem to. Is that by design?

This works:
```
$ endo eval "E(wk.query).lookup('agoricNames', 'brand', 'IST')" wk:alice-wk
Object [Alleged: IST brand#board0257] {}
```

But using a dotted path doesn't:

```
$ endo eval "E(wk.query).lookup('agoricNames', 'brand')" -n brand1 wk:alice-wk
Object [Alleged: NameHub] {}

$ endo eval 'x' x:brand1.IST
CapTP cli exception: (RemoteError(error:captp:Endo#20001)#1)
RemoteError(error:captp:Endo#20001)#1: Invalid pet name "brand1.IST"

  at decodeErrorCommon (packages/marshal/src/marshal.js:281:28)
  at decodeErrorFromCapData (packages/marshal/src/marshal.js:297:14)
  at decodeFromCapData (packages/marshal/src/encodeToCapData.js:404:27)
  at fromCapData (packages/marshal/src/marshal.js:359:23)
  at CTP_RETURN (packages/captp/src/captp.js:697:24)
  at dispatch (packages/captp/src/captp.js:776:7)
  at packages/daemon/src/connection.js:34:7

(RemoteError(error:captp:Endo#20001)#1)
```

rekmarks:
@dckc I think you're seeing that error because it's hitting the pet store `lookup` instead of the mailbox lookup modified in this PR: https://github.com/endojs/endo/blob/kriskowal-daemon-pet-name-path-lookup/packages/daemon/src/pet-store.js/#L70-L74

@kriskowal, should the pet store lookup also be updated or are they serving different purposes?

kriskowal:
I have not gone so far in this PR to support both dot delimited pet paths and also variadic lookup arguments. We could do both. I feel that we must support the dot delimited paths since a lot of the Endo API uses string arguments for pet names / pet paths and it would be awkward to make them all arrays or variadic.

kriskowal:
> The dotted paths should just be a CLI / UI feature, right? They should get parsed before use in other parts of the API. In agoric-sdk, a `PetName` is `string | string[]`.

That would work and I do like moving the concern of pet name language out to the UI layers.

kriskowal:
(Then the question is whether to flatten both variadic arguments and array arguments here)

dckc:
> (Then the question is whether to flatten both variadic arguments and array arguments here)

I think so, yes... but I'm not sure about the question. Give me an example, please?

kriskowal:
> > (Then the question is whether to flatten both variadic arguments and array arguments here)
> I think so, yes... but I'm not sure about the question. Give me an example, please?

Should this work: `E(hub).lookup(['a', 'b'], ['c', 'd'])`?

Should this work: `E(hub).lookup('a', ['b', 'c'])`?

dckc:
> > > (Then the question is whether to flatten both variadic arguments and array arguments here)
> > > I think so, yes... but I'm not sure about the question. Give me an example, please?
> 
> Should this work: `E(hub).lookup(['a', 'b'], ['c', 'd'])`?
> 
> Should this work: `E(hub).lookup('a', ['b', 'c'])`?

No.

```js
const KeyShape = M.string();
const PathShape = M.arrayOf(KeyShape);

export const NameHubIKit = harden({
  nameHub: M.interface('NameHub', {
...
    lookup: M.call().rest(PathShape).returns(M.promise()),
```

https://github.com/Agoric/agoric-sdk/blob/master/packages/vats/src/nameHub.js

```ts
export type NameHub = {
...
  /**
   * Look up a path of keys starting from the current NameHub. Wait on any
   * reserved promises.
   */
  lookup: (...path: string[]) => Promise<any>;
```
https://github.com/Agoric/agoric-sdk/blob/master/packages/vats/src/types.d.ts

kriskowal:
> > Should this work: `E(hub).lookup(['a', 'b'], ['c', 'd'])`?
> > Should this work: `E(hub).lookup('a', ['b', 'c'])`?
> No.

So, only `E(hub).lookup('a', 'b', 'c', 'd')`. It would follow `E(hub).lookup(petName)`, `E(hub).lookup(...petNamePath)`, or `E(hub).lookup(...petNamePathFrom(petNameOrPetNamePath)`.

```js
/**
 * @typedef {string} PetName
 * @typedef {string[]} PetNamePath
 */

/**
 * @param {string | string[]} petNameOrPetNamePath
 */
const petNamePathFrom = arg => typeof arg === 'string' ? [ arg ] : arg;

E(hub).lookup(...petNamePathFrom(petNameOrPetNamePath);
```

kriskowal:
Attn @rekmarks I’m in favor of reworking this PR so that the Daemon API uses arrays for pet name paths instead of dot delimited strings. The CLI would be responsible for splitting pet name path strings to make those arrays. The Daemon API might accept `string | string[]` in some positions, except the `lookup` method would become variadic and not accept arrays.

rekmarks:
Superseded by #2022.

PR #1916, opened 2023-12-29
feat(daemon): Undeniable host INFO tool

kriskowal:
Problem: Commands like `endo make counter.js --name counter` implicitly create a unique worker for the counter. This makes it impossible to address the worker for follow-up commands, like an `eval` co-located with the `counter`.

This change introduces an undeniable `INFO` name to all hosts that allows the host to address the values used to create pet named values. For example, `INFO -> counter -> worker` addresses the worker that was used to create `counter`. Until we have pet name path lookup (#1915), you need multiple invocations of `E()` to perform of depth greater than 1, for example:

```text
> endo make counter.js --name counter
  Object [Alleged: Counter] {}
> endo eval 'E(INFO).lookup("counter")' INFO --name inspector
  Object [Alleged: Inspector (make-bundle ... )]
> endo eval 'E(inspector).lookup("worker")' inspector
  Object [Alleged: EndoWorker] {}
```

This facility is intended to compose well with follow-up changes for inspecting a worker’s log or copy one of these “unnamed” capabilities to a name.

### DIFF
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -37,7 +37,15 @@ const delay = async (ms, cancelled) => {
   });
 };
 
-const makeInfo = (type, number, record) =>
+/**
+ * Creates an inspector object for a formula.
+ *
+ * @param {string} type - The formula type.
+ * @param {string} number - The formula number.
+ * @param {Record<string, unknown>} record - A mapping from special names to formula values.
+ * @returns {import('./types.js').EndoInspector} The inspector for the given formula.
+ */
+const makeInspector = (type, number, record) =>
   Far(`Inspector (${type} ${number})`, {
     lookup: async petName => {
       if (!Object.hasOwn(record, petName)) {
@@ -385,18 +393,18 @@ const makeEndoBootstrap = (
     } else if (formulaIdentifier === 'pet-inspector') {
       // Behold, unavoidable forward-reference:
       // eslint-disable-next-line no-use-before-define
-      const external = makeIdentifiedInspector('pet-store');
+      const external = makePetStoreInspector('pet-store');
       return { external, internal: undefined };
     } else if (formulaIdentifier === 'host') {
       const storeFormulaIdentifier = 'pet-store';
-      const infoFormulaIdentifier = 'pet-inspector';
+      const inspectorFormulaIdentifier = 'pet-inspector';
       const workerFormulaIdentifier = `worker-id512:${zero512}`;
       // Behold, recursion:
       // eslint-disable-next-line no-use-before-define
       return makeIdentifiedHost(
         formulaIdentifier,
         storeFormulaIdentifier,
-        infoFormulaIdentifier,
+        inspectorFormulaIdentifier,
         workerFormulaIdentifier,
         terminator,
       );
@@ -428,7 +436,7 @@ const makeEndoBootstrap = (
     } else if (formulaType === 'pet-inspector-id512') {
       // Behold, unavoidable forward-reference:
       // eslint-disable-next-line no-use-before-define
-      const external = makeIdentifiedInspector(
+      const external = makePetStoreInspector(
         `pet-store-id512:${formulaNumber}`,
       );
       return { external, internal: undefined };
@@ -440,14 +448,14 @@ const makeEndoBootstrap = (
       return { external, internal: undefined };
     } else if (formulaType === 'host-id512') {
       const storeFormulaIdentifier = `pet-store-id512:${formulaNumber}`;
-      const infoFormulaIdentifier = `pet-inspector-id512:${formulaNumber}`;
+      const inspectorFormulaIdentifier = `pet-inspector-id512:${formulaNumber}`;
       const workerFormulaIdentifier = `worker-id512:${formulaNumber}`;
       // Behold, recursion:
       // eslint-disable-next-line no-use-before-define
       return makeIdentifiedHost(
         formulaIdentifier,
         storeFormulaIdentifier,
-        infoFormulaIdentifier,
+        inspectorFormulaIdentifier,
         workerFormulaIdentifier,
         terminator,
       );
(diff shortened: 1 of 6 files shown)


packages/daemon/src/daemon.js
```diff
@@ -611,6 +638,108 @@ const makeEndoBootstrap = (
...
+    /** @param {string} petName */
+    const lookup = async petName => {
+      const formulaIdentifier = petStore.lookup(petName);
+      if (formulaIdentifier === undefined) {
+        throw new Error(`Unknown pet name ${petName}`);
+      }
+      const delimiterIndex = formulaIdentifier.indexOf(':');
+      // eslint-disable-next-line @endo/restrict-comparison-operands
+      if (delimiterIndex < 0) {
+        return undefined;
```
dckc, line 655: the lookup protocol in the nameHub in `@agoric/vats` throws on a key-not-found situation.

packages/daemon/src/daemon.js
```diff
@@ -611,6 +638,108 @@ const makeEndoBootstrap = (
...
+              formula.names.map((name, index) => {
+                return [
+                  name,
+                  provideValueForFormulaIdentifier(formula.values[index]),
+                ];
+              }),
+            ),
+          }),
+        );
+      } else if (formula.type === 'import-unsafe') {
```
dckc, line 691: we're not fans of `switch`?

dckc (commented):
looks plausible. I don't see any correctness issues.

without tests, I defer to others to approve

packages/daemon/src/daemon.js
kriskowal: Doug maligned switch in `jslint` and The Good™ Parts and I haven’t looked back.

rekmarks: IIRC the most pernicious problems with `switch` are solved by the [no-case-declarations](https://eslint.org/docs/latest/rules/no-case-declarations#rule-details) and [no-fallthrough](https://eslint.org/docs/latest/rules/no-fallthrough) ESLint rules. There are a couple of others related to `default` cases that are also nice.

Academic point for present purposes: In TypeScript you can even do fancy things like [switch-exhaustiveness-check](https://typescript-eslint.io/rules/switch-exhaustiveness-check/).

packages/daemon/src/daemon.js
```diff
@@ -611,6 +638,108 @@ const makeEndoBootstrap = (
...
+        return makeInfo(
+          formula.type,
+          formulaNumber,
+          harden({
+            SPECIFIER: formula.specifier,
```
kumavis, line 696: lint error here https://github.com/endojs/endo/actions/runs/7360309483/job/20036265775?pr=1916#step:9:147

kumavis (changes requested):

kumavis (changes requested):
I'd like to see types for these interfaces. I'd also suggest that because we control all these keys (no user specified keys), they don't need to be capitalized

```
eval
  SOURCE
  WORKER
  ENDOWMENTS
import-unsafe
  SPECIFIER
  WORKER
  POWERS
import-bundle
  WORKER
  BUNDLE
  POWERS
guest
  HOST
web-bundle
  BUNDLE
  POWERS
```

packages/daemon/src/daemon.js
rekmarks: I'll be using the `parseFormulaIdentifier` introduced in #1997, which does throw an error in the relevant sub-cases of this. See [here](https://github.com/endojs/endo/blob/dfa60718d97f311c6fd05c087cc55af5a52abe82/packages/daemon/src/formula-identifier.js).

packages/daemon/test/test-endo.js
```diff
@@ -694,3 +694,93 @@ test('make a host', async t => {
...
+    ['inspector'],
+  );
+  t.regex(String(worker), /Alleged: EndoWorker/u);
+
+  await stop(locator);
+});
+
+// TODO: This test verifies existing behavior when pet-naming workers.
+// This behavior is undesirable.
+test('eval-mediated worker name', async t => {
```
rekmarks, line 741: I went for a passing test rather than `test.failing()` because the test could otherwise start to fail for an unexpected reason.

packages/daemon/src/daemon.js
```diff
@@ -583,6 +618,117 @@ const makeEndoBootstrap = (
...
+    const petStore = await provideValueForFormulaIdentifier(
+      petStoreFormulaIdentifier,
+    );
+
+    /**
+     * @param {string} petName - The pet name to inspect.
+     * @returns {Promise<import('./types').KnownEndoInspectors[string]>} An
+     * inspector for the value of the given pet name.
+     */
+    const lookup = async petName => {
```
rekmarks, line 640: We have types for the individual known inspectors (see `types.d.ts`), but due to the indirection between the `petName` argument and the `formulaType`, it is difficult (impossible?) to map the `petName` to the "correct" return type.

packages/daemon/test/test-endo.js
rekmarks: Issue for this: #2021

packages/daemon/src/daemon.js
kriskowal: Agreed. We can’t be more specific than a union of all known inspector types. Lookup in general has this limitation. For some Endo APIs, we could overload a formula type argument and provide guarantees of specific types, to make some casts unnecessary.

packages/daemon/src/formula-identifier.js
```diff
@@ -1,10 +1,11 @@
 const { quote: q } = assert;
 
 const numberlessFormulasIdentifiers = new Set([
-  'pet-store',
-  'host',
   'endo',
```
kriskowal, line 4: Could add a comment that these are sorted, to maintain that invariant.

packages/daemon/test/test-endo.js
kriskowal: We like embedding links to issues in the code, especially for something like this.

kriskowal (commented):
I approve, but as the author can’t authorize a merge. Need a second opinion, possibly from coauthor @rekmarks, better from @kumavis.

kumavis (approved):
whee

packages/daemon/src/formula-identifier.js
rekmarks: I'll defer this to when I look into our lint configs.

PR #1918, opened 2023-12-30
feat(daemon): List special names

kriskowal:
Hosts and guests have a pet store and also special names. The special names are distinguished by being in `ALL_CAPS` like `SELF` or `HOST` so that they do not collide with `kebab-case` pet names. The special names cannot be deleted or overwritten. Guests and hosts have different special names.

This change makes the list of special names observable by adding `listSpecial()` and `listAll()` methods to the host and guest facets and revealing them in the CLI with `--special` and `--all` flags.

This change also alters the `endo list` command to accept an optional dot-delimited pet name path to the store the user wishes to list. The object at the end of that dot delimited path must implement `list()`, `listAll()`, or `listSpecial()` but can otherwise be an arbitrary object implementing this protocol. This completely subsumes the behavior of the `--as` flag for this command, freeing `-a` up for `--all`.

### DIFF
--- a/packages/cli/src/endo.js
+++ b/packages/cli/src/endo.js
@@ -288,18 +288,11 @@ export const main = async rawArgs => {
     });
 
   program
-    .command('list')
-    .option(
-      '-a,--as <party>',
-      'Pose as named party (as named by current party)',
-      collect,
-      [],
-    )
+    .command('list [directory]')
     .description('show names')
-    .action(async cmd => {
-      const { as: partyNames } = cmd.opts();
+    .action(async (directoryName, cmd) => {
       const { list } = await import('./list.js');
-      return list({ partyNames });
+      return list({ directoryName });
     });
 
   program
--- a/packages/cli/src/endo.js
+++ b/packages/cli/src/endo.js
@@ -289,10 +289,13 @@ export const main = async rawArgs => {
 
   program
     .command('list [directory]')
-    .description('show names')
+    .description('show names (dot delimited pet name path)')
+    .option('-s,--special', 'show special names')
+    .option('-a,--all', 'show all names')
     .action(async (directoryName, cmd) => {
+      const { special, all } = cmd.opts();
       const { list } = await import('./list.js');
-      return list({ directoryName });
+      return list({ directoryName, special, all });
     });
 
   program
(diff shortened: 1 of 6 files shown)


dckc:
these naming conventions (starting with kebab-case) make my spidey-sense tingle.

These names are supposed to be user-chosen, no? It feels very strange to not let the user choose whatever name they like.

Distinguishing system names by SHOUTING also seems fragile. something like `$ystem` would appeal to me.

Are the tests of what kind of name a string is localized? that would help in case we change our minds.

kriskowal:
> these naming conventions (starting with kebab-case) make my spidey-sense tingle.
> 
> These names are supposed to be user-chosen, no? It feels very strange to not let the user choose whatever name they like.
> 
> Distinguishing system names by SHOUTING also seems fragile. something like `$ystem` would appeal to me.
> 
> Are the tests of what kind of name a string is localized? that would help in case we change our minds.

This is a good time to consider alternative designs. The parameters of the design are like:

* A pet name path is a sequence of pet names and special (un-losable) names.
* For some user interfaces, it will be necessary to distinguish a pet name path from surrounding text, as in `endo send friend 'Hello, please enjoy this @edge-name:pet-name.pet-name.pet-name.'`
* For the API, a pet name path could be an array of tagged pet/special names, like `[{special: 'INFO'}, {name: 'counter'}, {special: 'WORKER'}]` or a string in a language for expressing pet name paths like `INFO.counter.WORKER`
* Special names could be distinguished by a sigil like `$info` and that invites a conversation about Hilbert Hoteling / escaping the sigil.
* Pet names could be arbitrary strings including dots and sigils, and for the `endo send` example, that invites a conversation about the grammar. Kebab-case could be special and braces could be used to envelope more expressive names, e.g., `endo send CPA 'Please download your @{Tax Signature Pages.pdf}, sign, and return.'` That invites a conversation about escaping / Hilbert Hoteling.

The current design is definitely optimized to minimize the scope of design conversations during this spike, but we’re definitely getting close to the end of the spike.

dckc:
> ... Kebab-case could be special and braces could be used to envelope more expressive names, e.g., `endo send CPA 'Please download your @{Tax Signature Pages.pdf}, sign, and return.'`

That one appeals to me, but...

> That invites a conversation about escaping / Hilbert Hoteling.

Yes. Lots to think about. Hm.

rekmarks:
Superseded by #2068.

Issue #1901, opened 2023-12-17
`endo mkhost` fails to properly create an `EndoHost` object.


tgrecojs:
## Describe the bug
any attempt to create a host object via `endo mkhost <host-name>` fails to properly create an `EndoHost` object and instead, the command returns an error which is stored inside the pet store under the host name used when attempting to make the host.

The entire sequence can be seen in the screenshot at the bottom of this issue.

## Steps to reproduce

1. make sure `endo` command is accessible
2. execute `endo mkhost thomas`

## Expected behavior

A new `EndoHost` object should be created an added to my pet store so that when I execute `endo show thomas` I will see `Object [Alleged: EndoHost] {}`.

## Screenshots

![endo mkhost behavior](https://github.com/endojs/endo/assets/6646552/2ed8fa76-6022-4bf9-8241-6a8ed21201f4)

PR #1919, opened 2023-12-30
fix(daemon): makeIdentifiedHost wrapped controller bug

kriskowal:
closes: #1901

## Description

A flaw in the Endo pet daemon prevented the user from creating new user profiles. The flow crept in through the refactor that separated external and internal facets for every entity that endo can create.

### DIFF
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -445,13 +445,12 @@ const makeEndoBootstrap = (
       const workerFormulaIdentifier = `worker-id512:${formulaNumber}`;
       // Behold, recursion:
       // eslint-disable-next-line no-use-before-define
-      const external = makeIdentifiedHost(
+      return makeIdentifiedHost(
         formulaIdentifier,
         storeFormulaIdentifier,
         workerFormulaIdentifier,
         terminator,
       );
-      return { external, internal: undefined };
     } else if (
       [
         'eval-id512',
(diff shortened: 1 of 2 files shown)


packages/daemon/test/test-endo.js
```diff
@@ -670,3 +670,27 @@ test('terminate because of requested capability', async t => {
...
+    'client',
+    locator.sockPath,
+    cancelled,
+  );
+  const bootstrap = getBootstrap();
+  const host = E(bootstrap).host();
+  const host2 = E(host).provideHost('fellow-host');
+  await E(host2).makeWorker('w1');
+  const ten = await E(host2).evaluate('w1', '10', [], []);
+  t.is(10, ten);
```
dckc, line 693: `t.is` calls its args actual, expected, so this looks backward. Not that it matters much.

```suggestion
  t.is(ten, 10);
```

packages/daemon/src/daemon.js
```diff
@@ -445,13 +445,12 @@ const makeEndoBootstrap = (
       const workerFormulaIdentifier = `worker-id512:${formulaNumber}`;
       // Behold, recursion:
       // eslint-disable-next-line no-use-before-define
-      const external = makeIdentifiedHost(
+      return makeIdentifiedHost(
```
dckc, line 448: How did this pass static type-checking previously? I guess type coverage isn't thorough?

dckc (approved):
good stuff

packages/daemon/src/daemon.js
kriskowal: I’m betting we have to disable implicit any to get more juice from ts here.

PR #1922, opened 2024-01-01
fix(eventual-send): switch DEBUG option to be comma separated

erights:
refs: https://github.com/Agoric/agoric-sdk/pull/8136
refs: https://github.com/Agoric/agoric-sdk/issues/8096
refs: https://github.com/Agoric/agoric-sdk/blob/master/docs/env.md#DEBUG

## Description

As explained at https://github.com/Agoric/agoric-sdk/issues/8096 , prior to this PR, eventual-send's use of the `DEBUG` option mistakenly split it on colons (`':'`). This PR switches that one use to split on commas (`','`) instead. 

### Security Considerations

Following widespread conventions leads to fewer unpleasant surprises.

### Scaling Considerations

none

https://github.com/Agoric/agoric-sdk/blob/master/docs/env.md#DEBUG already speaks in terms of the value of `DEBUG` being a comma separated list. Prior to this PR, the endo implementation does not conform to that document. But the agoric-sdk implementation would still not conform until fixed by https://github.com/Agoric/agoric-sdk/pull/8136

### Testing Considerations

none

### Upgrade Considerations

none

### DIFF
--- a/packages/eventual-send/src/track-turns.js
+++ b/packages/eventual-send/src/track-turns.js
@@ -1,5 +1,8 @@
 /* global globalThis */
-import { getEnvironmentOption } from '@endo/env-options';
+import {
+  getEnvironmentOption,
+  environmentOptionsListHas,
+} from '@endo/env-options';
 
 // NOTE: We can't import these because they're not in scope before lockdown.
 // import { assert, details as X, Fail } from '@agoric/assert';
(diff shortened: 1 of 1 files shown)


michaelfig (approved):

mhofman:
While I agree it's technically a breaking change, I'm wondering if this particular instance warrant the churn associated with a major version bump. @kriskowal opinions?

kriskowal (approved):
Sounds good to me. I do not have strong feelings about the breaking change sigil, but if it’s kept, the body of the commit must have prose for the changelog.

```
fix(eventual-send)!: switch DEBUG option to be comma separated

*BREAKING CHANGE*: Use `DEBUG=track-turns,etc` instead of `DEBUG=track-turns:etc`.
```

erights:
>  I do not have strong feelings about the breaking change sigil,

I got rid of it. Thanks.

PR #1924, opened 2024-01-03
fix(exo): reform exo amplifier API

erights:
closes: #XXXX
refs: #1902 

## Description

At https://github.com/endojs/endo/pull/1902#issuecomment-1874451253 @mhofman rightly points out

> In the kernel meeting a couple weeks ago, we discussed having the shape of this rights amplification function be a simple function that accepts a facet, and returns the cohort. I think I preferred that shape vs taking a facet name, but I do realize I didn't capture that feedback in this pull request.

This PR fixes that oversight.

### Upgrade Considerations

I flagged this PR with a `fix(exo)!: ...` because this fix is not a compat change from #1902. OTOH, it is unlikely that anyone is using the #1902 API yet, in which case this PR is a compat change from the prior endo state. The last endo release was before #1902, so this PR is also a compat change from the last release.

Should I downgrade to `fix(exo): ...` ?

### DIFF
--- a/packages/exo/src/exo-makers.js
+++ b/packages/exo/src/exo-makers.js
@@ -70,29 +70,47 @@ export const initEmpty = () => emptyRecord;
  */
 
 /**
- * @callback Revoker
+ * Template for function-valued options for exo class or exo class kit
+ * definitions, for receiving powers back at definition time. For example,
+ * ```js
+ * let amplify;
+ * const makeFoo = defineExoClassKit(
+ *   tag,
+ *   interfaceGuardKit,
+ *   initFn,
+ *   behaviorKit,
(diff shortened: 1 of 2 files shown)


mhofman:
> Should I downgrade to `fix(exo): ...`

IMO Yes

mhofman (approved):
Besides some typing and comment nits, LGTM. Please downgrade the "breaking" conventional commit down to a regular fix given that no release has been made since the feature was introduced.

erights:
> Besides some typing and comment nits, LGTM. Please downgrade the "breaking" conventional commit down to a regular fix given that no release has been made since the feature was introduced.

Done

PR #1925, opened 2024-01-04
feat(exo): instance testing

erights:
Now staged on #1964 

closes: #XXXX
refs: #1924 #1964 

## Description

This PR adds a third meta-level rights amplification power: the power to tell if an object is a live instance of an exo class, or a live facet instance of an exo class kit.

After revoking and amplifying, the remaining obvious meta-level rights amplification power normally associated with classes is unforgeable instance testing. (In some ocap literature, this is sometimes referred to as "brand testing". But given ERTP's conflicting use of the term "brand", this is best avoided.)

Given revocability and our desire that all these meta-level powers be essentially free, these test only that an object is a live instance of the associated class, or a live facet instance of the associated class kit. In neither case do we distinguish between a revoked former instance vs any other non-live-instance value.

### Security Considerations

In many OO languages, such as Java, access to a class bundles the ability to instantiate the class with the ability to reliably test for membership in the class. For exo classes and exo class kits, to follow the principles of least authority and information hiding, we decided to separate the two. As with the other meta-level rights amplification powers, these are not provided by default, and it is easy for a code review to see it they are provided at all, and where these powers might be used.

The obvious JavaScript test, `instanceof`, is not reliable. However, JavaScript classes do provide a reliable test via
private fields. Like the design in this PR, this power is initially present only within the lexical scope of the class. It is up to code within that lexical scope whether to make that power available to code elsewhere.

### Scaling Considerations

Should be free, as it uses the `contextMap` indirection already present in exos.

For all these meta-level rights amplification powers, the inline doc-comments now contains adequate text to seed real documentation, especially when consulting the tests for illustrative examples. However, none of these are yet implemented for virtual or durable exos, and cannot be until agoric-sdk is upgraded to depend on an endo release incorporating this PR. Until then, we should probably hold back on more public documentation of these APIs.

### Testing Considerations

This PR provided only tests for the heap-based exo implementations, which are all that the @endo/exo package provided. Once base-zone is migrated from agoric-sdk to endo, we should rewrite these tests using Zones so they are reusable for virtual and durable exos as well.

### Upgrade Considerations

For all these meta-level rights amplification powers, which powers are provided might vary per incarnation. Since amplification and live instance testing are query only powers against representation that exists anyway, having these vary per incarnation should not be a problem. It remains to be seen whether this will also be true for revocation.

### DIFF
--- a/packages/exo/src/exo-makers.js
+++ b/packages/exo/src/exo-makers.js
@@ -97,6 +97,19 @@ export const initEmpty = () => emptyRecord;
  * @returns {F}
  */
 
+/**
+ * The power to test if a value is a live instance of the
+ * associated exo class, or a live facet instance of the
+ * associated exo class kit. In the later case, if a `facetName` is provided,
+ * then it tests only whether the argument is a facet instance of that
+ * facet of the associated exo class kit.
+ *
+ * @callback IsLiveInstance
+ * @param {any} exo
+ * @param {string} [facetName]
+ * @returns {boolean}
+ */
+
 // TODO Should we split FarClassOptions into distinct types for
 // class options vs class kit options? After all, `receiveAmplifier`
 // makes no sense for normal exo classes.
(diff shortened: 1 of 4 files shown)


FUDCo (approved):
LGTM.  Still wish we had a better word than "kit" though.

PR #1928, opened 2024-01-08
more typedoc

turadg:
## Description

Enables typedoc for these packages:
- ses
- @endo/lockdown
- @endo/bundle-source - this required fixing a `realpathSync` argument
- @endo/marshal - this required hard-coding the `any` and `string` types that it was trying to import from `pass-style` but failing

### DIFF
--- a/packages/bundle-source/src/is-entrypoint.js
+++ b/packages/bundle-source/src/is-entrypoint.js
@@ -1,7 +1,8 @@
+// @ts-check
 // Detect if this is run as a script.
-import url from 'url';
-import process from 'process';
-import fs from 'fs';
+import * as url from 'url';
+import * as process from 'process';
+import * as fs from 'fs';
 
 // Agoric still uses Endo dependencies under an emulation of ESM we call RESM
 // because it is invoked with `node -r esm`.
@@ -13,4 +14,4 @@ const favor = (primary, secondary) =>
 // FIXME: Should maybe be exported by '@endo/something'?
 export const isEntrypoint = href =>
   String(href) ===
-  url.pathToFileURL(fs.realpathSync(favor(process.argv[1]), '/')).href;
+  url.pathToFileURL(fs.realpathSync(favor(process.argv[1]))).href;
--- a/packages/marshal/src/types.js
+++ b/packages/marshal/src/types.js
@@ -4,7 +4,7 @@ export {};
 /**
  * @template Slot
  * @callback ConvertValToSlot
- * @param {import('@endo/pass-style').PassableCap} val
+ * @param {any} val a PassableCap
  * @returns {Slot}
  */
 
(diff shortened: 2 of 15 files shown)


kriskowal (approved):

packages/marshal/src/types.js
```diff
@@ -4,16 +4,16 @@ export {};
 /**
  * @template Slot
  * @callback ConvertValToSlot
- * @param {import('@endo/pass-style').PassableCap} val
+ * @param {any} val a PassableCap
```
mhofman, line 7: That seems like a terrible regression, possibly due to types not being built!

mhofman (commented):
I am concerned something marked as "typedoc" is regressing types

packages/marshal/test/test-marshal-capdata.js
```diff
@@ -6,6 +6,11 @@ import { makeMarshal } from '../src/marshal.js';
 
 const { freeze, isFrozen, create, prototype: objectPrototype } = Object;
 
+const harden = /** @type {import('ses').Harden & {isFake: boolean}} */ (
```
erights, line 9: Would the following also work? If so, I think it better documents the runtime reality it describes. Likewise for the other occurrences of course

```suggestion
const harden = /** @type {import('ses').Harden & {isFake?: boolean}} */ (
```

packages/marshal/src/types.js
turadg: It's not a regression. `PassableCap` in master is `any`. 
Screenshot 2024-01-08 at 1 10 49 PM

packages/bundle-source/src/is-entrypoint.js
```diff
@@ -1,7 +1,8 @@
+// @ts-check
 // Detect if this is run as a script.
-import url from 'url';
-import process from 'process';
-import fs from 'fs';
+import * as url from 'url';
+import * as process from 'process';
+import * as fs from 'fs';
```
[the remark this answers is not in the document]
turadg: These Node modules have no default exports. Under CommonJS the set of named exports could be treated as an object but under ESM you have to bundle the named exports into an object if you want to access them that way.

packages/marshal/test/test-marshal-capdata.js
turadg: Yes it would. I'll see about adding tacking that onto another PR.

packages/marshal/src/types.js
mhofman: Then the issue is with the `PassableCap` definition, not this param definition.

turadg: There was indeed an issue with the PassableCap definition. It was `any`. The type is still `any` so there's no regression here.

Once we fix those types, I expect to return to this.

packages/marshal/test/test-marshal-capdata.js
turadg: https://github.com/endojs/endo/pull/1929/commits/19dbdd8ee06ce82d674ac0d961f920883292c4d0

packages/marshal/src/types.js
turadg: Fix in https://github.com/endojs/endo/pull/1933

Issue #1488, opened 2023-02-07
missing Remotable and PassByRef types


turadg:
The description is correct but its type is aliased to `Passable` (which is wrong). Moreover since `Passable` is any, `Passable` is any.

https://github.com/endojs/endo/blob/686ab61802c0e5206b14f4b91d499e3873555cd4/packages/pass-style/src/types.js#L20-L74

https://docs.agoric.com/guides/js-programming/far.html

`Remotable` = `Far` object or local presence for a Far object (can be done with a tag b/c only `Far()` and `unserialize` return it)

`PassByCopy` needs a def.

PureData is deeply PassByCopy. (currently `Passable`/any)

PassByRef = `Remotable | Promise<Remotable> | Promise<PassByCopy>` (in snippet as PassableCap)

Passable = `PassByRef  | PassByCopy` (but probably not feasible to fully define)

@gibson042 points out adjustments may be needed for error handling

erights:
PassByCopy !== PureData

PassByCopy is shallow. PureData is deeply PassByCopy. `harden([Far('x',{})])` is a PassByCopy CopyArray, but is not `PureData` because it contains a `Remotable`.

erights:
> @gibson042 points out adjustments may be needed for error handling

Points out where? Or what?

erights:
> Remotable = Far object or local presence for a Far object (can be done with a tag b/c only Far() and serialize return it)

Did you mean "unserialize"?

turadg:
For context, this issue body was from a huddle (@gibson042,  @dckc and me)

> PassByCopy !== PureData

Thanks. I've updated the body.

> Points out where? Or what?

It was merely a verbal statement like "adjustments may be needed for error handling" (paraphrase)

> Did you mean "unserialize"?

Yes, corrected.

> What does the third disjunct mean?

Editing error. Updated in body.

kriskowal:
What’s the status of this change?

turadg:
> What’s the status of this change?

This is an issue so no changes have yet been made. 

I'd still like to see this problem solved. The `any` cropped up for me again in https://github.com/endojs/endo/pull/1928 just now. 

I'll see if I can make some progress today

mhofman:
Wondering if this is linked, and possibly was worked around by the manual type generation step before publish

aj-agoric:
Decided this is high priority in the endo project meeting cc @turadg

PR #1933, opened 2024-01-08
1488 Passable

turadg:
closes: #1488

## Description

- Makes `Passable` a generic type instead of `any`.
- Defines overloads for `passStyleOf` to return the actual style.
- Defines the `Key` in terms of `Passable`
- Makes a ton of fixes and suppressions in places that relied on the previous `any`'s

### Security Considerations

n/a

### Scaling Considerations

n/a

Better types are better documentation.

### Testing Considerations

I improved the test coverage. Additional suggestions welcomed.

### Upgrade Considerations

These changes may cause type errors downstream. I don't think we should call those breaking change à la semver because they don't affect the runtime.


packages/pass-style/src/types.js
```diff
@@ -6,19 +6,21 @@ export {};
...
  */
 
-// TODO declare more precise types throughout this file, so the type system
-// and IDE can be more helpful.
+/** @typedef {import('type-fest').Primitive | CopyArray | CopyRecord | CopyTagged } PassByCopy */
+/** @typedef {RemotableObject | Promise<RemotableObject> | Promise<PassByCopy> } PassByRef */
 
 /**
- * @typedef {any} Passable
+ * @typedef {PassByCopy | PassByRef} Passable
```
dckc, line 23: I can't fit this recursive definition into JSDoc, but things like `PureData` suggest that `Passable` should be parameterized...

```suggestion
 * @template [C=PassableCap]
 * @template [E=Error]
 * @typedef {import('type-fest').Primitive | Container<Passable<C, E>> | C | E} Passable
```

where...

```js
/**
 * @template [C=PassableCap]
 * @template [E=Error]
 * @typedef {CopyArray<Passable<C, E>> | CopyRecord<Passable<C, E>> | CopyTagged<Passable<C, E>>} Container
 */
```

and then:

```js
 /**
  * @typedef {Passable<never, never>} PureData
  */
```

dckc (commented):
is a recursive definition of `Passable` feasible?

packages/pass-style/src/types.js
turadg: I ran into circular definition errors. I'll try more. https://stackoverflow.com/a/70501728/46040 has a tip.

I also want to try https://2ality.com/2020/06/type-guards-assertion-functions-typescript.html#using-overloading for the switch-based narrowing

dckc: A while back I managed to do a similar recursive type in https://github.com/ocapn/ocapn/blob/dc-codec-play/packages/ocapn-value/src/ocapn-value.d.ts

I wish I had cited a source for the "gymnastics"

packages/pass-style/src/types.js
```diff
@@ -42,10 +57,26 @@ export {};
...
+ * (p: number): 'number';
+ * (p: bigint): 'bigint';
+ * (p: symbol): 'symbol';
+ * (p: null): 'null';
+ * (p: Promise): 'promise';
+ * (p: Error): 'error';
+ * (p: CopyTagged): 'tagged';
+ * (p: Array): 'copyArray';
+ * (p: Iterable): 'remotable';
+ * (p: Iterator): 'remotable';
```
dckc, line 75: I don't understand why `Iterable` and `Iterator` are here. I wonder what I'm missing.

packages/pass-style/src/types.js
```diff
@@ -74,7 +105,14 @@ export {};
  */
 
 /**
- * @typedef {Passable} RemotableObject
+ * @template {string} S pass style
+ * @template {InterfaceSpec} I interface tag
+ * @typedef {PassStyled<S> & {[Symbol.toStringTag]: I}} TaggedRecord
```
dckc, line 110: This use of "Tagged" is different from the `tagged` passStyle, right? That seems like a source of confusion.

packages/pass-style/src/types.js
turadg: hacks along the way. I'll see about removing them

packages/pass-style/src/types.js
turadg: Different, yes. This is as in "toString**Tag**".  Naming suggestions welcomed

packages/pass-style/src/types.js
turadg: I didn't know how else to handle these cases https://github.com/endojs/endo/blob/3d927c30dec4786d16aa1e988e86d0ea6617f2e2/packages/patterns/test/test-copyMap.js#L131-L133

dckc: oh. wild. I didn't realize `getCopyMapEntries()` returned a passable value.

I _did_ know that `tsc` thinks `getCopyMapEntries()` returns `any` before this PR.
https://github.com/Agoric/dapp-offer-up/blob/8da2191af4a3333e0ec0f0cd884f739fa021950a/contract/src/gameAssetContract.js#L21

packages/pass-style/src/types.js
dckc: I see.... :thinking:

packages/marshal/src/encodeToCapData.js
```diff
@@ -20,7 +20,7 @@ import {
 
 /** @typedef {import('@endo/pass-style').Passable} Passable */
 /** @typedef {import('./types.js').Encoding} Encoding */
-/** @typedef {import('@endo/pass-style').Remotable} Remotable */
+/** @typedef {import('@endo/pass-style').RemotableObject} RemotableObject */
```
gibson042, line 23: Hmm, what is RemotableObject?

packages/marshal/src/marshal.js
```diff
@@ -236,7 +238,7 @@ export const makeMarshal = (
 
     /**
      * @param {{iface?: string, index: number}} slotData
-     * @returns {Remotable | Promise}
+     * @returns {Passable}
```
gibson042, line 241: Why is this return type being loosened? A slot really should decode to only a Remotable or Promise.

packages/marshal/src/marshal.js
```diff
@@ -255,7 +257,7 @@ export const makeMarshal = (
 
     /**
      * @param {{errorId?: string, message: string, name: string}} errData
-     * @param {(e: unknown) => Passable} decodeRecur
+     * @param {(e: unknown) => string} decodeRecur
```
gibson042, line 260: `decodeRecur` is the general decoder function and is not limited to returning strings.

packages/pass-style/src/passStyleOf.js
```diff
@@ -81,28 +81,28 @@ const makePassStyleOf = passStyleHelpers => {
    * structures, so without this cache, these algorithms could be
    * O(N**2) or worse.
    *
-   * @type {WeakMap<Passable, PassStyle>}
+   * @type {WeakMap<WeakKey, PassStyle>}
```
gibson042, line 84: Do we have a TypeScript type corresponding with our Key concept (Passable structure in which each leaf is a Passable primitive value or a Remotable)?

packages/pass-style/src/passStyleOf.js
```diff
@@ -81,28 +81,28 @@ const makePassStyleOf = passStyleHelpers => {
...
     const passStyleOfRecur = inner => {
       const innerIsObject = isObject(inner);
       if (innerIsObject) {
-        if (passStyleMemo.has(inner)) {
-          // @ts-ignore TypeScript doesn't know that `get` after `has` is safe
-          return passStyleMemo.get(inner);
+        const innerMemo = passStyleMemo.get(inner);
+        if (innerMemo) {
+          // sanity check on falsy values. Do 'get' first to get the type narrowing of the `if`
+          assert(passStyleMemo.has(inner));
```
gibson042, line 104: What value is this sanity check providing, given that `get` is guaranteed to return undefined for absent keys?
```suggestion
```

packages/pass-style/src/remotable.js
```diff
@@ -167,15 +166,26 @@ const checkRemotable = (val, check) => {
   return result;
 };
 
-/** @type {MarshalGetInterfaceOf} */
+/**
+ * Simple semantics, just tell what interface (or undefined) a remotable has.
+ * @type {{
+ * <T extends string>(val: import('./types.js').TaggedRecord<any, T>): T;
+ * (val: any): any;
+ * }}
```
gibson042, line 174: Cool! Can we narrow further?
```suggestion
/**
 * Simple semantics, just tell what interface (or undefined) a remotable has.
 * @type {{
 * <T extends string>(val: import('./types.js').TaggedRecord<any, T>): T;
 * (val: any): string | undefined;
 * }}
```

packages/pass-style/src/types.js
```diff
@@ -6,19 +9,31 @@ export {};
...
+ * @typedef {{ [PASS_STYLE]: S }} PassStyled
+ */
 
 /**
- * @typedef {any} Passable
+ * @template {PassStyled<any>} P
+ * @typedef {P[typeof PASS_STYLE]} ExtractStyle
+ */
+
+/** @typedef {import('type-fest').Primitive | CopyArray | CopyRecord | CopyTagged } PassByCopy */
```
gibson042, line 32: Is this intentionally omitting errors?

packages/pass-style/src/types.js
```diff
@@ -74,7 +105,14 @@ export {};
...
 /**
- * @typedef {Passable} RemotableObject
+ * @template {string} S pass style
+ * @template {InterfaceSpec} I interface tag
+ * @typedef {PassStyled<S> & {[Symbol.toStringTag]: I}} TaggedRecord
+ */
+
+/**
+ * @template {InterfaceSpec} [I=string]
+ * @typedef {TaggedRecord<'remotable', I>} RemotableObject
```
gibson042, line 115: Why RemotableObject rather than Remotable? What differentiates them?

packages/pass-style/src/types.js
```diff
@@ -87,26 +125,24 @@ export {};
  */
 
 /**
- * @template {Passable} [T=Passable]
+ * @template {Passable} [T=object]
```
gibson042, line 128: `object` isn't Passable, is it? I thought that type allows mixing methods and non-function properties. Same applies to later `@template {Passable} [T=object]` patterns as well.

packages/pass-style/src/types.js
gibson042: "PassStyled" seems like a good name... do we need both types, or can they be collapsed together?

packages/patterns/src/keys/copySet.js
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck FIXME relies heavily on dynamic typing.
```
gibson042, line 1: Are these ts-nocheck directives going to preclude useful checks that would otherwise be possible?

packages/patterns/src/keys/merge-set-operators.js
```diff
@@ -25,7 +25,7 @@ const { quote: q, Fail } = assert;
  *  to `fullOrder`. However, it optimizes for the case where these contiguous
  * runs that need to be resorted are either absent or small.
  *
- * @template T
+ * @template {import('../types').Passable} T
```
gibson042, line 28: I don't think the functions in this file need to be constrained to collections of Passables.

packages/patterns/src/types.js
```diff
@@ -22,7 +22,7 @@ export {};
 /** @typedef {import('@endo/marshal').RankCover} RankCover */
 
 /**
- * @typedef {Passable} Key
+ * @typedef {Exclude<Passable, Error | Promise>} Key
```
gibson042, line 25: This is still overly broad; we should probably have a TODO.

packages/patterns/src/types.js
```diff
@@ -61,7 +61,7 @@ export {};
  */
 
 /**
- * @typedef {Passable} Pattern
+ * @typedef {Exclude<Passable, Error | Promise>} Pattern
```
gibson042, line 64: Likewise here.

gibson042 (commented):
This is very nice!!!

packages/pass-style/src/passStyleOf.js
dckc: That reminds me: Key is another motivation for parameterizing Passable: `Passable<Remotable, never>`

PureData is `Passable<never, never>` as discussed earlier

packages/pass-style/src/remotable.js
dckc: Since remotable tags are only for debugging, I don't think we should use them for static typing.

packages/patterns/src/types.js
dckc: Exclude at the top level doesn't address promises and errors nested further down. Again, I hope we can parameterize Passable.

packages/pass-style/src/remotable.js
turadg: > Can we narrow further?

I had tried that and it caused some problems. I'll document.

> Since remotable tags are only for debugging, I don't think we should use them for static typing.

That's what this doc says,
https://github.com/endojs/endo/blob/40f6cf8794f9c67b3f5a65f01b7037ab960149bc/packages/patterns/src/types.js#L310-L313

but when we changed the label it failed rehydration,
https://github.com/Agoric/agoric-sdk/blob/62a7c9c42007020994f6036cf4090590e4811b8e/packages/zoe/src/zoeService/startInstance.js#L65

Is that a bug?

packages/pass-style/src/passStyleOf.js
turadg: indeed

packages/pass-style/src/types.js
turadg: I wasn't sure. I thought I saw an object with the `[PASS_STYLE]` property and not the `toStringTag` or vice-versa. Are they supposed to be always present together?

packages/pass-style/src/types.js
turadg: nope. I'll add that

packages/pass-style/src/types.js
turadg: `Remotable` is a value 
https://github.com/endojs/endo/blob/master/packages/pass-style/src/make-far.js#L80C1-L82

I'll document

packages/pass-style/src/types.js
turadg: Documented now,
```
// TODO a better default type. Ideally it's Passable, but that causes a circular definition.
```

packages/patterns/src/keys/copySet.js
turadg: Possibly. When the rest of this settles I can come back and change them to local suppressions.

packages/patterns/src/keys/merge-set-operators.js
turadg: They compare T using `RankCompare` whose documentation says,
https://github.com/endojs/endo/blob/40f6cf8794f9c67b3f5a65f01b7037ab960149bc/packages/marshal/src/types.js#L163-L166

packages/pass-style/src/remotable.js
gibson042: > when we changed the label it failed rehydration, https://github.com/Agoric/agoric-sdk/blob/62a7c9c42007020994f6036cf4090590e4811b8e/packages/zoe/src/zoeService/startInstance.js#L65
> 
> Is that a bug?

Plausibly, although not related to this PR. The fundamental issue is that [agoric-sdk doesn't allow a new vat incarnation to change `stateShape`](https://github.com/Agoric/agoric-sdk/issues/7337#issuecomment-1514195810), and adding/removing/altering the label of a Matcher\<'remotable'> _is_ a change (albeit a benign one from internal structure like `{ [Symbol.for("passStyle")]: "tagged", [Symbol.toStringTag]: "match:kind", payload: "remotable" }` to `{ [Symbol.for("passStyle")]: "tagged", [Symbol.toStringTag]: "match:remotable", payload: { label: "$label" } }`). This will need to be allowed in the fix for https://github.com/Agoric/agoric-sdk/issues/7337 / https://github.com/Agoric/agoric-sdk/issues/7407 .

packages/pass-style/src/passStyleOf.js
gibson042: I don't think so, because `passStyleMemo` values are always PassStyle strings. And if we do care, I guess the place for such an assertion would be adjacent to `set` rather than `get`.

mhofman:
I am only following from a distance, but I believe you can have circular type definitions in `.d.ts` files, but not in JSDoc

packages/pass-style/src/types.js
gibson042: > Are they supposed to be always present together?

Yes, AFAIK. A Tagged has own `[PASS_STYLE]: "tagged", [Symbol.toStringTag]: $tag`, and a Remotable has a prototype chain in which the penultimate object has own `[PASS_STYLE]: "remotable", [Symbol.toStringTag]: $iface` (where both `$tag` and `$iface` must be strings, and the latter must either be "Remotable" or start with "Alleged: " or "DebugName: ").

packages/patterns/src/keys/merge-set-operators.js
gibson042: Ah, I forgot about that.

turadg:
> I am only following from a distance, but I believe you can have circular type definitions in `.d.ts` files, but not in JSDoc

I'm not sure about that but I do know the solution is more complicated than changing to .ts. Even with .ts,
```
export type CopyArray<T extends Passable = Passable> = T[];
```
causes
```
Type alias 'PassByCopy' circularly references itself.
```

packages/pass-style/src/types.d.ts
```diff
@@ -0,0 +1,163 @@
...
+  (p: number): 'number';
+  (p: bigint): 'bigint';
+  (p: symbol): 'symbol';
+  (p: null): 'null';
+  (p: Promise<any>): 'promise';
+  (p: Error): 'error';
+  (p: CopyTagged): 'tagged';
+  (p: any[]): 'copyArray';
+  (p: Iterable<any>): 'remotable';
+  (p: Iterator<any, any, undefined>): 'remotable';
```
erights, line 69: I don't understand these two lines. Remotables don't have any du jure relationship with `Iterable` or `Iterator`. It so happens that Remotables can implement the `Iterable` or `Iterator` interface, i.e., they can also be Interables or Iterators, but that's true for many possible such interface types.

erights: Nevermind. I see this point is already under active discussion. Removing.

erights:
Attn @michaelfig re https://github.com/endojs/endo/pull/1933#issuecomment-1885583987 circular types mysteries

dckc:
#1952 shows how to address the circular type issues, I think.

It's based on [earlier ocapn work](https://github.com/ocapn/ocapn/pull/51), but at the time I learned tsc's limitations around self-referential types, I neglected to take good notes.

@mhofman gave a particularly useful pointer earlier today:

 - https://github.com/microsoft/TypeScript/pull/33050

packages/marshal/src/encodeToCapData.js
```diff
@@ -117,7 +117,7 @@ export const makeEncodeToCapData = (encodeOptions = {}) => {
    * Readers must not care about this order anyway. We impose this requirement
    * mainly to reduce non-determinism exposed outside a vat.
    *
-   * @param {Passable} passable
+   * @param {any} passable
```
erights, line 120: Curious: Why did this "regress" from `Passable` to `any`?

(scare quotes because the old `Passable` meant `any`. So the real question is: Why couldn't this become the new `Passable`?)

Likewise for similar changes elsewhere.

packages/marshal/src/encodeToCapData.js
```diff
@@ -269,11 +269,11 @@ harden(makeEncodeToCapData);
  * @property {(
  *   encodedRemotable: Encoding,
  *   decodeRecur: (e: Encoding) => Passable
- * ) => (Promise|Remotable)} [decodeRemotableFromCapData]
+ * ) => (Promise|RemotableObject)} [decodeRemotableFromCapData]
```
erights, line 272: Thanks for doing this type renaming. Having a type name collide with a value (function) name was really terrible. Good to see this straightened out. No change suggested.

packages/pass-style/src/passStyleOf.js
erights: Unfortunately, simpy knowledge that it is a string does not by itself imply that it is truthy, since the empty string is falsy. In this case, we know that it is always a pass-style string, which is always a non-empty string and thus always truthy. But still, to avoid even needing to worry about this edge case, I suggest rewriting

```js
if (inner) {
```
to
```js
if (inner !== undefined) {
```
or
```js
if (typeof inner === 'string') {
```
FWIW, I prefer the `!== undefined` .

erights: In any case, given that we statically know we're only putting strings in there, I'd say no further sanity check is needed.

turadg:
> #1952 shows how to address the circular type issues, I think. I discussed it with @turadg today. It's not clear how much work it would be to finish that direction.

I think that solved it, @dckc . PTAL [5b8b349](https://github.com/endojs/endo/pull/1933/commits/5b8b349ab4651b1f401b6bf0f8aecd4d48f9683b) [9aac68e](https://github.com/endojs/endo/pull/1933/commits/9aac68e5aad82c24ebd3eca3262178faa606d94f).

packages/patterns/src/keys/checkKey.js
```diff
@@ -100,11 +100,13 @@ harden(assertScalarKey);
...
 // ////////////////////////////// Keys /////////////////////////////////////////
 
+// @ts-expect-error Key does not satisfy WeakKey
 /** @type {WeakSet<Key>} */
+// @ts-expect-error Key does not satisfy WeakKey
 const keyMemo = new WeakSet();
 
 /**
- * @param {Passable} val
+ * @param {Key} val
```
erights, line 109: Excellent!

No change suggested.

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1,3 +1,5 @@
+// @ts-nocheck So many errors that the suppressions hamper readability.
+// TODO parameterize MatchHelper which will solve most of them
```
erights, line 2: Interesting.

No change suggested.

erights (commented):
Very enthusiastic to see this PR be merged!

Looks Good To Me so far, but review not yet completed.

packages/pass-style/src/types.d.ts
```diff
@@ -0,0 +1,182 @@
...
+  | ContainerStyle
+  | 'remotable'
+  | 'error'
+  | 'promise';
+export type PassStyled<S extends string> = {
+  [PASS_STYLE]: S;
+};
+export type ExtractStyle<P extends PassStyled<any>> = P[typeof PASS_STYLE];
+export type PassByCopy =
+  | import('type-fest').Primitive
```
erights, line 24: TIL `'type-fest'`. Does everyone else already know about this? Is it worth a comment or URL?

packages/pass-style/src/types.d.ts
```diff
@@ -0,0 +1,182 @@
...
+  | Error
+  | any[]
+  | CopyRecord
```
erights, line 27: Curious why we do use `CopyRecord` rather than `Record` but we use `any[]` rather than `CopyArray`?

Also, shouldn't this at least be `Passable[]`, or does this encounter recursive type problems again?

packages/marshal/src/encodeToCapData.js
turadg: All the string replacements of old-Passable to `any` is because the new (non-any) Passable caused too many errors and I judged solving them to not be worth it. Usually because the following code relied heavily on dynamic type inspection and would not benefit from the static type information. This was mostly in `switch` statements where TS cannot statically infer the type narrowing being used.

packages/pass-style/src/types.d.ts
```diff
@@ -0,0 +1,182 @@
...
+  | CopyTagged;
+export type PassByRef =
+  | RemotableObject
+  | Promise<RemotableObject>
+  | Promise<PassByCopy>;
```
erights, line 32: I was going to suggest

```suggestion
export type PassByRef =
  | RemotableObject
  | Promise<Passable>;
```
but then I realized that your's correctly excludes `Promise<Promise>`, which is great! No change suggested.

packages/pass-style/src/types.d.ts
```diff
@@ -0,0 +1,182 @@
...
+ *     Passable (PassStyle 'error').
+ *
+ * A Passable is essentially a pass-by-copy superstructure with a
+ * pass-by-reference
+ * exit point at the site of each PassableCap (which marshalling represents
+ * using 'slots').
+ */
+export type Passable<
+  PC extends PassableCap = PassableCap,
+  E extends Error = Error,
```
erights, line 57: Took me awhile to understand what's going on here. This is awesome! No change suggested.

packages/pass-style/src/types.d.ts
```diff
@@ -0,0 +1,182 @@
...
+  | CopyTaggedI<PC, E>;
+interface CopyArrayI<PC extends PassableCap, E extends Error>
+  extends CopyArray<Passable<PC, E>> {}
+interface CopyRecordI<PC extends PassableCap, E extends Error>
+  extends CopyRecord<Passable<PC, E>> {}
+interface CopyTaggedI<PC extends PassableCap, E extends Error>
+  extends CopyTagged<string, Passable<PC, E>> {}
+
+export type PassStyleOf = {
+  (p: undefined): 'undefined';
```
erights, line 72: I don't understand the type notation here. What does it mean?

What especially confuses me is `{ ... : ...`
suggests a record type where the thing to the left of the colon is a property key. But that does not look consistent with the rest. From the rest, I'd guess that it is a type overload, but overloaded on what? Is `p` a parameter?

packages/pass-style/src/types.d.ts
```diff
@@ -0,0 +1,182 @@
...
+  (p: bigint): 'bigint';
+  (p: symbol): 'symbol';
+  (p: null): 'null';
+  (p: Promise<any>): 'promise';
+  (p: Error): 'error';
+  (p: CopyTagged): 'tagged';
+  (p: any[]): 'copyArray';
+  (p: Iterable<any>): 'remotable';
+  (p: Iterator<any, any, undefined>): 'remotable';
+  <T extends PassStyled<any>>(p: T): ExtractStyle<T>;
```
erights, line 85: Is this too open ended? Does it loose the static knowledge that the extracted style can only be one of a small number of enumerated strings?

packages/pass-style/src/types.d.ts
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * Given this (currently counter-factual) assumption, a PureData value cannot
+ * be used as a communications channel,
+ * and can therefore be safely shared with subgraphs that should not be able
+ * to communicate with each other.
+ * Without that assumption, such a guarantee requires a marshal-unmarshal round
+ * trip (as exists between vats) to produce data structures disconnected from
+ * any potential proxies.
+ */
+export type PureData = Passable<never, never>;
```
erights, line 111: Awesome! No change suggested.

packages/pass-style/src/types.d.ts
```diff
@@ -0,0 +1,182 @@
...
+ */
+export type CopyTagged<
+  Tag extends string = string,
+  Payload extends Passable = any,
+> = TaggedRecord<'tagged', Tag> & {
+  payload: Payload;
+};
+/**
+ * This is an interface specification.
+ * For now, it is just a string, but will eventually be `PureData`. Either
```
erights, line 158: No change suggested.

But noting that I doubt we ever would change this from a string. But the time we'd want to, it will be too great a compat hazard to bother with. Unless we encounter a compelling need, which I do not expect.

packages/pass-style/src/types.d.ts
```diff
@@ -0,0 +1,182 @@
...
+ *    would have returned `false`, and simply returns `undefined` normally
+ *    when the predicate would have returned `true`.
+ *  * The internal checking function that they share is parameterized by a
+ *    `Checker` that determines how to proceed with a failure condition.
+ *    Predicates pass in an identity function as checker. Validators
+ *    pass in `assertChecker` which is a trivial wrapper around `assert`.
+ *
+ * See the various uses for good examples.
+ */
+export type Checker = (
```
erights, line 179: No change suggested.

Just FYI as of https://github.com/endojs/endo/pull/1935 I expect I'll migrate `Checker` to a lower level utils package. But there's no reason for that to affect this PR, which will merge first.

erights (approved):
LGTM!!

packages/pass-style/src/types.d.ts
turadg: I figure it's intuitive that it's an NPM package and those are easy to find. But I have to clean up the commits anyway so I'll include this link https://github.com/sindresorhus/type-fest/blob/main/source/primitive.d.ts

packages/pass-style/src/types.d.ts
turadg: Yep, it's defines overloads for the function signature. If the call matches a function that has one parameter, of type `undefined`, then the return type is the string literal `'undefined'`. Each case defines another signature. They're matched in order so the most general is last.

packages/pass-style/src/types.d.ts
turadg: I'll revise the text to something less certain than "will eventually be"

packages/pass-style/src/types.d.ts
mhofman: This is a weird dependency to take on for non test types. It's simple enough, can we just inline it?

turadg: Ok, I'll inline it.

packages/pass-style/src/types.d.ts
```diff
@@ -0,0 +1,182 @@
...
+  | 'string'
+  | 'symbol';
+export type ContainerStyle = 'copyRecord' | 'copyArray' | 'tagged';
+export type PassStyle =
+  | PrimitiveStyle
+  | ContainerStyle
+  | 'remotable'
+  | 'error'
+  | 'promise';
+export type PassStyled<S extends string> = {
```
mhofman, line 19: Why not `extends PassStyle`?

packages/pass-style/src/types.d.ts
turadg: changed to CopyArray

packages/pass-style/src/types.d.ts
mhofman: You could likely do `Promise<Awaited<Passable>>`

packages/pass-style/src/types.d.ts
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * A Passable is essentially a pass-by-copy superstructure with a
+ * pass-by-reference
+ * exit point at the site of each PassableCap (which marshalling represents
+ * using 'slots').
+ */
+export type Passable<
+  PC extends PassableCap = PassableCap,
+  E extends Error = Error,
+> = import('type-fest').Primitive | Container<PC, E> | PC | E;
```
mhofman, line 58: `type-fest` here too

mhofman (commented):
Did a quick pass. Lots of great changes. I'm a little surprised some type definitions drop down to not plumbing through the PassableCap restriction, and I would need to look closer where that loosening would have an impact, but we can always iterate on that later

packages/pass-style/src/types.d.ts
turadg: tightened up with,
```ts
export type TaggedOrRemotable = 'tagged' | 'remotable';
```

packages/pass-style/src/types.js
turadg: Thanks! 1b30d600cf5084b061590c12ca39294328bb69c6

erights:
@turadg @kriskowal all, 

Now that the endo-agoric-sdk sync is behind us, I would love to see this PR move forward again! What is needed?

At https://github.com/Agoric/agoric-sdk/pull/8771 I have a PR of postponed changes on the agoric-sdk side that were waiting on that sync. Would that one help or hurt the effort to move forward on this PR?

turadg:
> What is needed?

In Endo, a PR re-applying these changes.

And in agoric-sdk a PR that integrates with that PR, demonstrating agoric-sdk CI passing. That way when land the Endo PR we can follow up quickly with a release and make the SDK PR the basis of an Endo upgrade.

I've started in
- https://github.com/Agoric/agoric-sdk/pull/8774 


> At [Agoric/agoric-sdk#8771](https://github.com/Agoric/agoric-sdk/pull/8771) I have a PR of postponed changes on the agoric-sdk side that were waiting on that sync. Would that one help or hurt the effort to move forward on this PR?

I'm not sure. I think 8771 should be prioritized, though, because it doesn't require an Endo release. I'll try to get 8774 passing and I expect by the time it's R4R that master will have 8771 so I'll solve whatever else comes up.

PR #1934, opened 2024-01-08
fix(scripts): Update create-package comment doc

kriskowal:
I renamed `repackage.sh` to `create-package.sh` when it stopped being idempotent. This updates its internal documentation to reflect the change.

### DIFF
--- a/scripts/create-package.sh
+++ b/scripts/create-package.sh
@@ -1,12 +1,11 @@
 #!/bin/bash
 
-# Updates or creates a package with the given name (idempotent).
+# Creates a package with the given name.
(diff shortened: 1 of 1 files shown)


erights (approved):
LGTM

@kriskowal @michaelfig @dckc what do you think should happen to the remaining repackage.sh in Jessie?

kriskowal:
> LGTM
> 
> @kriskowal @michaelfig @dckc what do you think should happen to the remaining repackage.sh in Jessie?

My feeling about these kinds of tools is that we should update them when we use them next.

Issue #1754, opened 2023-08-30
Links are broken in npm package pages


gibson042:
## Describe the bug

For example, the types.js link at the bottom of https://www.npmjs.com/package/@endo/patterns _should_ point at https://github.com/endojs/endo/blob/master/packages/patterns/src/types.js like the one in its source at https://github.com/endojs/endo/blob/master/packages/patterns/README.md , but instead points at the nonexistent https://github.com/endojs/endo/blob/HEAD/src/types.js .

We can probably address this in the publication process.

kriskowal:
We need to add a `directory` entry to the `repository` in `package.json`. That should fix the npm rendering of relative links.

PR #1936, opened 2024-01-09
fix: Add repository directory to all package descriptors

kriskowal:
fixes: #1754 

This adds a repository directory to every package’s descriptor `package.json` so that `npmjs.com` can render relative links correctly.


gibson042 (approved):
Nice!

PR #1939, opened 2024-01-09
feat: explicitly harden some shared prototypes

mhofman:
refs: #1914

## Description

While working on #1686, I found out that some shared prototypes are currently hardened as a side effect of an instance being hardened. This PR adds explicit harden for these.

It also contains a small refactor to harden all intrinsics in one shot instead of multiple calls to `harden`, as there are cycles in those, and some intrinsics were effectively hardened in other calls.

### Security Considerations

More explicit hardening brings more security

### Scaling Considerations

None

In Node.js, the `Buffer` and when the console is untamed, the internal `SafeMap` are now explicitly hardened. These would likely get hardened as side-effect of hardening instances before.

### Testing Considerations

Tested in #1914 and https://github.com/Agoric/agoric-sdk/pull/8700 that these are all prototypes found through instances by existing tests. There may be others lurking, but until we have shallow hardening implemented, we won't find them.

### Upgrade Considerations

This is a non breaking change that should be mostly unobservable to consumers.

### DIFF
--- a/packages/captp/src/atomics.js
+++ b/packages/captp/src/atomics.js
@@ -56,7 +56,7 @@ export const makeAtomicsTrapHost = transferBuffer => {
 
   const te = new TextEncoder();
 
-  return async function* trapHost([isReject, serialized]) {
+  return harden(async function* trapHost([isReject, serialized]) {
     // Get the complete encoded message buffer.
     const json = JSON.stringify(serialized);
     const encoded = te.encode(json);
@@ -94,7 +94,7 @@ export const makeAtomicsTrapHost = transferBuffer => {
         yield;
       }
     }
-  };
+  });
 };
 
 /**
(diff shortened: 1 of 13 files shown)


mhofman:
Apparently some changes run before lockdown, back to drawing board

packages/ses-ava/src/ses-ava-test.js
```diff
@@ -75,6 +75,16 @@ const overrideList = [
   'only',
 ];
 
+// Helper to explicitely harden an object including its prototype chain
```
erights, line 78: Should the changes in this file be moved into #1914 ? They don't really make sense in this PR.

erights (approved):
Except for the one misplaced set of changes to ses-ava-test.js, LGTM.

Thanks for this!

Thanks especially for figuring out and taking care of that Node `console` weirdness.

packages/ses-ava/src/ses-ava-test.js
mhofman: Yeah I debated this. Happy to move to #1914

Issue #1345, opened 2022-10-31
Remove or suppress warning: Tolerating undeletable ... === undefined


Jack-Works:
https://github.com/endojs/endo/blob/4da7742e8017d07094ed9b9336a85b6a4d3ee7b6/packages/ses/src/whitelist-intrinsics.js#L292

![image](https://user-images.githubusercontent.com/5390719/198927559-f8a974fc-6e33-4413-9a48-427ae1ab2414.png)

This warning is annoying and does not help too much because we know what we're doing (using core-js)

kriskowal:
We could alternately bundle all the Lockdown warnings under an error group to be less loud and still reveal all these details. @erights?

erights:
Error group is good.

PR #1941, opened 2024-01-09
refactor(ses): refactor console init and test cleanup

erights:
closes: #XXXX
refs: #1345 #1942 

## Description

A flaw in tame-console.js made it hard to use throws-and-logs.js to test the console output that happens during lockdown. The flaw is that tame-console.js read the value of the global `console` variable at initialization, rather than when `tameConsole()` is called. This PR moves that initialization code into `tameConsole`, which is highly unlikely to break any working code.

Using that fix, this PR also tests the current behavior that #1345 complains about. A later PR #1942 will fix that problem, building on this PR so that the difference is itself tested.

### Security Considerations

Prior to this PR, tame-console.js used `console` and `print` as ambient globals without a `/* global ... */` annotation. For `console` this is not surprising, as we generally assume this is ambiently available. But for `print` it is surprising. Do we have a bug in our lint diagnostic framework? Attn @kriskowal 

In any case, this PR changes those specifically to `globalThis.console` and `globalThis.print`, where that `globalThis` is itself obtained from `commons.js`.

Moving the sampling of these globals into `tameConsole()` enables them to be changed before `tameConsole()` samples them, as we now intend. This should have zero security implications as these will still be sampled before `lockdown()` returns.

### Scaling Considerations

none

none

### Testing Considerations

Moving the sampling of `console` this way should enable other uses of `throws-and-logs.js` for testing console output that would not work before this PR.

### Upgrade Considerations

none.

### DIFF
--- a/packages/ses/src/error/tame-console.js
+++ b/packages/ses/src/error/tame-console.js
@@ -21,33 +21,6 @@ const failFast = message => {
 const wrapLogger = (logger, thisArg) =>
   freeze((...args) => apply(logger, thisArg, args));
 
-// eslint-disable-next-line no-restricted-globals
-const originalConsole = /** @type {VirtualConsole} */ (
-  // eslint-disable-next-line no-nested-ternary
-  typeof console !== 'undefined'
-    ? console
-    : typeof print === 'function'
-    ? // Make a good-enough console for eshost (including only functions that
-      // log at a specific level with no special argument interpretation).
-      // https://console.spec.whatwg.org/#logging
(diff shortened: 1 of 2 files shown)


kriskowal (approved):
I do not know whether the base lint environment implies `print`, but I would not be surprised if it did. I also do not know where to look.

https://github.com/eslint/eslint/blob/main/conf/globals.js provides neither console nor print, so is clearly not our source of truth

I did learn there are some interesting environment names, like `shared-node-browser`, that we might try out. https://eslint.org/docs/latest/use/configure/language-options#specifying-environments

PR #1942, opened 2024-01-09
feat(ses): group removal cleanup diagnostics

erights:
Staged on #1941 

closes: #1345 
refs: #1941 

## Description

Places all intrinsics removal cleanup diagnostics into a collapsed console group.

Attn @Jack-Works 

### Security Considerations

none

### Scaling Considerations

none

We should mention somewhere what this diagnostic means. Even though it is less visible, because it is now in a collapsed group, it is still purposely visible on the console output and people should be able to find out what it means --- and to know when to bring it to our attention.

### Testing Considerations

none

### Upgrade Considerations

none

### DIFF
--- a/packages/ses/src/permits-intrinsics.js
+++ b/packages/ses/src/permits-intrinsics.js
@@ -74,7 +74,18 @@ export default function whitelistIntrinsics(
   intrinsics,
   markVirtualizedNativeFunction,
 ) {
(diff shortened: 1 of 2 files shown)


kriskowal (approved):
Thank you for taking this ticket.

For the record, I see that [`console.groupCollapsed`](https://caniuse.com/?search=console.groupCollapsed) is pretty universal.

PR #1949, opened 2024-01-10
docs(ses): Clarify that SES reveals stacks to console on uncaught exception

### LINKED ISSUE #887: Stale docs: Logging uncaught exceptions
At https://github.com/endojs/endo/blob/master/packages/ses/src/error/README.md

>  * SES hides annotations and stack traces by default. To reveal them, use a mechanism like `process.on("uncaughtException")` in Node.js to catch the error and log it back to the `console` tamed by `lockdown`.

@kriskowal could you update this in light of your recent tooling improvements? Thanks.

kriskowal:
Closes #887 updating docs to reflect current implementation, which does reveal stacks on uncaught exception in most platforms without additional effort.


erights (approved):
LGTM

PR #1952, opened 2024-01-10
define Key, PureData using parameterized Passable (WIP)

dckc:
suggestions for #1933

I think the pass-style stuff is complete. The patterns stuff is just a start.

The marshal stuff is probably superfluous.


turadg:
@dckc I rebased https://github.com/endojs/endo/pull/1933 on top of master so I rebased this PR on top of that. https://github.com/endojs/endo/tree/ta/dc-Passable-types has the result. 

I don't totally get what it buys. Does it make sense to try getting #1933 into master and coming back to this?

turadg (commented):
This was in my review queue so I took a look at getting this into master and most of it is already, including constraining PureData.

The only functional change I see is:
```diff
- * @typedef {Exclude<Passable<RemotableObject, never>, Error | Promise>} Key
+ * @typedef {Passable<RemotableObject, never>} Key
```

@dckc is that more accurate? What's a good test that fails now and will pass?

dckc:
I think that's not even a functional change. The Exclude<...> is redundant: the never already says that. 

So it's just a refactor for clarity/conciseness. 

A few type tests to show that the types are equivalent might be worthwhile.

PR #1957, opened 2024-01-10
fix(bundle-source): Accommodate Windows lack of atomic rename

### LINKED ISSUE #1955: Windows broke atomic rename
## Describe the bug

Per https://github.com/nodejs/node/issues/29481, Windows does not support atomic swap (POSIX rename semantics) by default when the target file exists. Since CI caches the target for some tests, we’re seeing CI errors on Windows only like https://github.com/endojs/endo/actions/runs/7471064596. To address this, we may need an environment variable or to back off atomic swap on Windows only.

kriskowal:
Fixes #1955

Windows rename is not atomic and fails if the target exists. Our CI revealed this to us because it cached a bundle.

### DIFF
--- a/packages/bundle-source/cache.js
+++ b/packages/bundle-source/cache.js
@@ -296,12 +296,13 @@ export const makeNodeBundleCache = async (
   pid,
   nonce,
 ) => {
-  const [fs, path, url, crypto, timers] = await Promise.all([
+  const [fs, path, url, crypto, timers, os] = await Promise.all([
     await loadModule('fs'),
     await loadModule('path'),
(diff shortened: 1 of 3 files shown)


mhofman (approved):
I'm confused. Why not get the stats after the rename? would that not alleviate the risk of race?

kriskowal:
> I'm confused. Why not get the stats after the rename? would that not alleviate the risk of race?

The race occurs if there are concurrent reads and writes for the same cache entry. The reader can encounter a partial bundle. But, I won’t pretend that I’ve thought this through completely.

PR #1958, opened 2024-01-11
fix(pass-style): remove redundant vestigial @returns declaration

erights:
closes: #XXXX
refs: #1933 

## Description

The declaration site had both a correct @type declaration and a redundant vestigial partial @returns declaration. Because the latter also was not properly formed, it caused the warning

```
.../endojs/endo/packages/pass-style/src/remotable.js
  175:1  warning  Missing JSDoc @returns type  jsdoc/require-returns-type

✖ 1 problem (0 errors, 1 warning)
```

This PR removes it, and does a bit of trivial polishing of the remainder.

### Security Considerations

none

### Scaling Considerations

none

more accurate type declarations should eventually help documentation

### Testing Considerations

none

### Upgrade Considerations

none

### DIFF
--- a/packages/pass-style/src/remotable.js
+++ b/packages/pass-style/src/remotable.js
@@ -164,12 +164,13 @@ const checkRemotable = (val, check) => {
 };
 
 /**
- * Simple semantics, just tell what interface (or undefined) a remotable has.
+ * Simple semantics, just tell what interface spec a Remotable has,
+ * or undefined if not deemed to be a Remotable.
+ *
  * @type {{
  * <T extends string>(val: PassStyled<any, T>): T;
- * (val: any): string | undefined;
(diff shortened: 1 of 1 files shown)


turadg (approved):
I'd suggest removing the `jsdoc/require-returns-type` rule, since it is legitimate to define a type for the return and separately a JSDoc describing it.

erights:
@turadg , you approved this long ago, but I held it back waiting for endo typing to settle down. Which it has -- thanks!

I'm still getting the same warning, and the same fix still seems correct, so I'm merging this. Also, I have not done anything about the `jsdoc/require-returns-type` that you suggest removing.

Letting you know because so much has changed since then, even though this still seems good as is.

PR #1960, opened 2024-01-11
chore(monorepo): Make placeholder tests fail by default

### LINKED ISSUE #1586: New packages should be initialized with test files that fail
Right now, [skel/test/test-index.js](https://github.com/endojs/endo/blob/591b985bedd1dd2ffa3bf3e510d25c1dd5531c72/skel/test/test-index.js) includes a "place holder" test with a passing assertion, but it should probably instead include contents that will cause `yarn test` etc. to fail, as an indication that there is still work to be done.

rekmarks:
Makes the placeholder test in the placeholder package (`packages/skel/test/test-index.js`) fail by default.

Closes #1586

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "format": "yarn prettier --write .github packages",
     "lint": "yarn prettier --check .github packages && lerna run lint",
     "lint-fix": "lerna run --no-bail lint-fix",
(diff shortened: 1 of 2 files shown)


kriskowal (approved):
Very nice, skipping `skel`.

Issue #676, opened 2021-04-19
Replace toString overrides with @toStringTag symbol overrides


kriskowal:
This is in better keeping with the language’s aesthetic. Notably, Compartment.prototype.toString should be replaced with a tag.

rekmarks:
The module namespace object now also has a `@@toStringTag`. See #1972.

kriskowal:
Closed by https://github.com/endojs/endo/pull/1962

PR #1962, opened 2024-01-13
refactor(ses): Replace Compartment toString property with @@toStringTag

rekmarks:
Replaces the `Compartment` `toString` function with a `@@toStringTag` string property. This necessitated updating the permit for `%CompartmentPrototype%`, and some related tests. Some adjacent cleanup is also performed.

Ref: #676

### DIFF
--- a/packages/ses/test/error/throws-and-logs.js
+++ b/packages/ses/test/error/throws-and-logs.js
@@ -1,13 +1,11 @@
 /* global globalThis */
 
-import { freeze, getPrototypeOf, is } from '../../src/commons.js';
-import { loggedErrorHandler, assert } from '../../src/error/assert.js';
+import { freeze, getPrototypeOf } from '../../src/commons.js';
+import { loggedErrorHandler } from '../../src/error/assert.js';
 import {
   makeLoggingConsoleKit,
   makeCausalConsole,
 } from '../../src/error/console.js';
-
-const { quote: q } = assert;
 
 // For our internal debugging purposes
 // const internalDebugConsole = console;
@@ -19,11 +17,6 @@ const compareLogs = freeze((t, log, goldenLog) => {
   t.is(log.length, goldenLog.length, 'wrong log length');
   log.forEach((logRecord, i) => {
     const goldenRecord = goldenLog[i];
-    t.is(
-      logRecord.length,
-      goldenRecord.length,
-      `wrong length of log record ${i}`,
-    );
     logRecord.forEach((logEntry, j) => {
       const goldenEntry = goldenRecord[j];
       if (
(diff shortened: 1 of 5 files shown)


packages/ses/test/error/throws-and-logs.js
```diff
@@ -19,11 +17,6 @@ const compareLogs = freeze((t, log, goldenLog) => {
   t.is(log.length, goldenLog.length, 'wrong log length');
   log.forEach((logRecord, i) => {
     const goldenRecord = goldenLog[i];
-    t.is(
-      logRecord.length,
-      goldenRecord.length,
-      `wrong length of log record ${i}`,
-    );
```
rekmarks, line 26: This is uselessly redundant with the assertions later in the function, and makes the failure from `ava` less informative.

packages/ses/test/error/throws-and-logs.js
```diff
@@ -33,14 +26,7 @@ const compareLogs = freeze((t, log, goldenLog) => {
...
         t.assert(logEntry instanceof goldenEntry, 'not the right error');
       } else {
-        // tap uses `===` instead of `Object.is`.
-        // Assuming ava does the right thing, switch back to this when
-        // switching back to ava.
-        // t.is(logEntry, goldenEntry);
-        t.assert(
-          is(logEntry, goldenEntry),
-          `${q(logEntry)} not same as ${q(goldenEntry)}`,
-        );
```
rekmarks, line 43: `ava` does the right thing: https://github.com/avajs/ava/blob/0a050243ea13df959e6c1536b2513b5fe3557690/docs/03-assertions.md#isactual-expected-message

packages/ses/test/test-compartment-instance.js
```diff
@@ -28,22 +28,7 @@ test('Compartment instance', t => {
...
-      'globalThis',
-      'import',
-      'importNow',
-      'load',
-      'module',
-      'name',
-      'toString',
-    ].sort(),
-    'prototype properties',
-  );
```
rekmarks, line 48: This appears to be redundant with the prototype properties test performed in `test-compartment-prototype`. Checking that the prototype of the Compartment instance is the expect prototype ought to be enough for this test?

kriskowal: I buy it.

packages/ses/test/error/throws-and-logs.js
kriskowal: Let’s instead move the length check after the individual record check. Removing this check entirely would hide a defect where the actual log is shorter than the expected log, since we’re iterating the actual log.

kriskowal (changes requested):
Needs a small adjustment to the test to maintain coverage, and otherwise looks good to me.

packages/ses/test/error/throws-and-logs.js
rekmarks: Done in [cf6d268](https://github.com/endojs/endo/pull/1962/commits/cf6d2684000b216353e638b5dfb76e402f9cdd4f)

kriskowal (approved):

PR #1935, opened 2024-01-08
refactor(common): move low level utilities to @endo/common

erights:
closes: #XXXX
refs: #1815

## Description

Successor to https://github.com/endojs/endo/pull/1815 with utilities moved to a new @endo/utils package.

Note: I did not label this `refactor(utils)!: ...` because I intend to add reexports to keep compat before declaring this ready for review.

PR #1963, opened 2024-01-13
fix(common): fix @endo/common integration breakage

erights:
closes: #XXXX
refs: #1935 

## Description

#1935 broke the agoric-sdk integration tests, i.e., would have broke integration with agoric-sdk . To repair the problem, this PR 
- gets rid of the `index.js` file which rolled up all the exports together
- moves each of the source files -- each of which currently represents a separately exported utility -- to the top level.
- changes the corresponding `"exports":` to use the same export name as the source file it exports, so that imports are the same whether processed by a tool that understands `"exports":` or not.
- change all importers to do the deep import of the particular exports it needs.

### Security Considerations

none

### Scaling Considerations

By forcing importers to import only from the exporting file they need, some implementations (bundlers, packagers) can do tree-shaking, omitting code not reachable from imports. For `@endo/common` specifically, the savings are tiny. But the pattern could bring substantial savings elsewhere.

nothing interesting

### Testing Considerations

none

### Upgrade Considerations

The old exports prior to #1935 that were exporting some of these, for compat, still reexports these. But it deprecates those reexports, with advice that importers do a deep import directly from `@endo/common/...`. Thus, this PR should not have any compat issues compared to the state preceding #1935.

This PR is technically a compat break with #1935 itself. But #1935 was just merged and never released, so that is not a source of worry.


packages/common/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@endo/common",
   "version": "1.0.1",
-  "private": true,
+  "private": false,
```
kriskowal, line 4: The field can be omitted. Harmless if present and false.

kriskowal (approved):
That covers it. It might or might not save time to propose a no-op PR to agoric-sdk pinned to this branch to verify that CI settles.

erights:
> That covers it. It might or might not save time to propose a no-op PR to agoric-sdk pinned to this branch to verify that CI settles.

https://github.com/Agoric/agoric-sdk/pull/7937

packages/common/package.json
erights: Done

erights:
For the record, https://github.com/Agoric/agoric-sdk/pull/7937 CI is currently green with `#endo-branch: markm-fix-common-package`, so merging. Afterwards, I'll switch https://github.com/Agoric/agoric-sdk/pull/7937 back to `master`.

Issue #629, opened 2021-03-22
Naming Integrity in compartment maps, lavamoat policies


dckc:
@kriskowal plans to write up what we talked about last week.

cc @erights @kumavis

dckc:
`yarn why` has an interesting naming scheme...

```
=> Found "@agoric/compartment-mapper#ses@0.11.0"
info This module exists because "_project_#@agoric#bundle-source#@agoric#compartment-mapper" depends on it.
```

kriskowal:
See also #619

kumavis:
@dckc theres a few other ways a "non-canonical" (wrt to the npm registry) package name can be used, such as [`bundledDependencies`](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#bundleddependencies). I have a poc in my [`bad-ideas-collection`](https://github.com/LavaMoat/bad-idea-collection/blob/main/packages/non-canonical-bundled/README.md)

kriskowal:
@kumavis @naugtur Can we consider this resolved with the use of A’a for naming packages in compartment map policies?

naugtur:
I've created a test case specifically for that bundled dependency scenario.
Our malicious package example, eve, now has a dependency named carol, which is a name used elsewhere.
Observations:
- the way the nested package has to be identified for policy is `eve>carol` instead of `carol` https://github.com/endojs/endo/pull/1967/files#diff-ddb1847c929bf61b1fd65b8b3c34028d5be8ec2b3d932e98271ab62e94be2c8bR237
- it does not get confused for any other carol and permissions aren't granted (as seen in the test result matching the assertion here https://github.com/endojs/endo/pull/1967/files#diff-ddb1847c929bf61b1fd65b8b3c34028d5be8ec2b3d932e98271ab62e94be2c8bR143)


We do have a place in the codebase that seems to be affected by this trick though. Not sure where, but the test is failing with `The bundler and importer should agree on source map count` and it seems like the sourcemap for the attenuator is not being loaded in this case, for which I have no idea how it could be affected by the fixture.

@kriskowal I might need some help figuring this out.

PR #1967, opened 2024-01-15
test(compartment-mapper): check for resistance to bundled dependency name collision

naugtur:
WIP test for #629


packages/compartment-mapper/test/scaffold.js
```diff
@@ -291,13 +291,15 @@ export function scaffold(
       };
 
       const sourceMaps = new Set();
+      const sourceMapLog = [];
       const sourceMapHook = (sourceMap, { sha512 }) => {
         sourceMaps.add(sha512);
-        t.log(sha512, sourceMap);
+        sourceMapLog.push(['+', sha512, sourceMap]);
```
naugtur, line 297: Made the log silent until it's needed because with even more lines added below it was getting annoying.

PR #1968, opened 2024-01-16
fix(ses): fix properties already override enabled

erights:
Seems to work but too complicated. To be closed once we see what CI does.

Back to the drawing board, with this one for reference.


erights:
Closed in favor of https://github.com/endojs/endo/pull/1969

PR #1969, opened 2024-01-16
fix(ses): handle properties that are already override protected

erights:
Fixes a bug initially reported by @FrederikBolding.
image

Staged on #1974 

closes: #1971 
refs: #1655 #1968 https://github.com/tc39/proposal-iterator-helpers https://github.com/endojs/endo/issues/1973

## Description

The iterators-helpers proposal https://github.com/tc39/proposal-iterator-helpers is now shipping in Chrome Canary. As per spec, it makes two properties into accessors that act like data properties that protect against the override mistake, i.e., that enable the property to be overridden by assignment. However, our permits for these properties were written when we knew them to be data properties, are they currently are on the other engines we've tested on to date. Because of this discrepancy, SES failed to initialize on Chrome Canary.

The purpose of this PR is to provide a mechanism to tolerate such methods, so that SES initializes when it encounters an accessor method of this form.

During the `repairIntrinsics` phase of lockdown, before even vetted shims are supposed to run, for each property we said may be of this form, validate as best we can whether it indeed seems to be an accessor that protects from the override mistake, but otherwise emulates a data property. For those, *temporarily* turn it into the plain data property it seems to emulate. But writable, so it won't trigger the override mistake in that form. Note that this JS is not a conformant initial JS state, and is observable by vetted shims.

Then, during the `enable-property-override` portion of the `hardenIntrinsics` phase, after all vetted shims ran, those properties listed in `enablements.js` will turn them (back) into accessors, but ones that we control. This PR also adds these two properties to all phases of enablements.

It turned out that the `enablements` language could not use symbols as property names. This PR fixed that because it needed that feature.

As a drive-by, the accessor functions installed by `enable-property-overrides.js` were `function` functions. This PR changes them to concise methods, so they'll still have the `this` sensitivity they need, but omit the `prototype` property and the `[[Construct]]` behavior.

### Security Considerations

ses did exactly the right thing here. When the initial pre-ses state deviates from what the ses authors knew about as of that version of ses, we divide those changes into two categories:
- Simple addition of stuff we did not know about. JS additions are usually additive, so we assume it is safe enough to delete these that ses init still results in a secure state. But even for these, we issue a warning (now made a bit less noisy https://github.com/endojs/endo/pull/1942 ) so that we become aware that there are changes to the initial state beyond what ses authors knew about as of that version of ses.
- Any change to stuff we already knew about, where the change is something we did not know about. For these, that version of ses must fail to initialize with a decent diagnostic (unless overridden by an explicitly “unsafe” configuration setting). Such changes should not be assumed to simply be additive. Ses must fail to initialize because the initial state changed in ways we don’t know if we can still secure. Ses init should NOT blindly do an intervention to restore the initial state to something it expects, because this surprising change may be symptomatic of coordinated semantic changes to other things that we cannot detect. The only way to get back to secure is exactly what we’ll do today: the ses authors must manually ensure they understand the actual meaning of the surprise, and only then, alter ses to arrive at a secure state starting from this now-understood initial state.

So,
- SES did exactly what it should do in refusing to initialize with a good enough diagnostic for us to pinpoint the problem quickly.
- Chrome did exactly what it should do, in releasing an agreed change to the JS spec first in Canary, so others (like us) get an early warning if JS changes in a way we need to adjust to.
- @FrederikBolding did what we SHOULD periodically do ourselves, which is run SES on the latest canaries of all the platforms we care about. @ivan we need such a process. In fact, we still need to try it against the canaries of the other browsers.
- We did in tc39 exactly what we should have done, which is lobby, successfully in this case, that a change that was going to happen somehow happened in the way least damaging to us.
- Having gotten the change in at tc39, I should have better anticipated it and adjusted ses init to expect this change without breakage.

### Scaling Considerations

none

none

### Testing Considerations

None of the existing tests broke, and I manually verified that things work as expected both on Chrome Canary, and on an earlier Chrome browser that has not implemented the iterators-helpers proposal.

I also exported the core logic as a function that I unit test.

### Upgrade Considerations

This JS change was just encountered in Chrome Canary for now, but it is accepted into the language and will shortly ship on all platforms. Before then, all users of SES should upgrade to a SES incorporating this PR.

### DIFF
--- a/packages/ses/src/enable-property-overrides.js
+++ b/packages/ses/src/enable-property-overrides.js
@@ -11,7 +11,6 @@ import {
   defineProperty,
   getOwnPropertyDescriptor,
   getOwnPropertyDescriptors,
-  getOwnPropertyNames,
   isObject,
   objectHasOwnProperty,
   ownKeys,
@@ -156,12 +155,11 @@ export default function enablePropertyOverrides(
     }
     // TypeScript does not allow symbols to be used as indexes because it
     // cannot recokon types of symbolized properties.
-    // @ts-ignore
     arrayForEach(ownKeys(descs), prop => enable(path, obj, prop, descs[prop]));
   }
 
   function enableProperties(path, obj, plan) {
-    for (const prop of getOwnPropertyNames(plan)) {
+    for (const prop of ownKeys(plan)) {
       const desc = getOwnPropertyDescriptor(obj, prop);
       if (!desc || desc.get || desc.set) {
         // No not a value property, nothing to do.
@@ -169,10 +167,8 @@ export default function enablePropertyOverrides(
         continue;
       }
 
-      // Plan has no symbol keys and we use getOwnPropertyNames()
-      // so `prop` cannot only be a string, not a symbol. We coerce it in place
-      // with `String(..)` anyway just as good hygiene, since these paths are just
-      // for diagnostic purposes.
+      // In case `prop` is a symbol, we first coerce it with `String`,
+      // purely for diagnostic purposes.
       const subPath = `${path}.${String(prop)}`;
       const subPlan = plan[prop];
 
--- a/packages/ses/src/enablements.js
+++ b/packages/ses/src/enablements.js
@@ -1,3 +1,5 @@
+import { toStringTagSymbol } from './commons.js';
+
 /**
  * @file Exports {@code enablements}, a recursively defined
  * JSON record defining the optimum set of intrinsics properties
@@ -73,6 +75,13 @@ export const minEnablements = {
 
   '%ErrorPrototype%': {
     name: true, // set by "precond", "ava", "node-fetch"
+  },
+  '%IteratorPrototype%': {
+    toString: true,
+    // https://github.com/tc39/proposal-iterator-helpers
+    constructor: true,
+    // https://github.com/tc39/proposal-iterator-helpers
+    [toStringTagSymbol]: true,
   },
 };
 
@@ -152,6 +161,10 @@ export const moderateEnablements = {
 
   '%IteratorPrototype%': {
     toString: true,
+    // https://github.com/tc39/proposal-iterator-helpers
+    constructor: true,
+    // https://github.com/tc39/proposal-iterator-helpers
+    [toStringTagSymbol]: true,
   },
 };
 
--- a/packages/ses/src/tame-faux-data-properties.js
+++ b/packages/ses/src/tame-faux-data-properties.js
@@ -1,0 +1,131 @@
+import {
+  getOwnPropertyDescriptor,
+  apply,
+  defineProperty,
+  toStringTagSymbol,
+} from './commons.js';
+
+const throws = thunk => {
+  try {
+    thunk();
+  } catch (er) {
+    return true;
+  }
+  return false;
+};
+
+/**
+ * Exported for convenience of unit testing. Harmless, but not expected
+ * to be useful by itself.
+ *
+ * @param {any} obj
+ * @param {string|symbol} prop
+ * @param {any} expectedValue
+ * @returns {boolean}
+ */
+export const tameFauxDataProperty = (obj, prop, expectedValue) => {
+  if (obj === undefined) {
+    return false;
+  }
+  const desc = getOwnPropertyDescriptor(obj, prop);
+  if (!desc || 'value' in desc) {
+    return false;
+  }
+  const { get, set } = desc;
+  if (typeof get !== 'function') {
+    return false;
+  }
+  if (typeof set !== 'function') {
+    return false;
+  }
+  const observedValue = get();
+  if (observedValue !== expectedValue) {
+    return false;
+  }
+  if (apply(get, obj, []) !== expectedValue) {
+    return false;
+  }
+  const testValue = 'Seems to be a setter';
+  const subject1 = { __proto__: null };
+  apply(set, subject1, [testValue]);
+  if (subject1[prop] !== testValue) {
+    return false;
+  }
+  const subject2 = { __proto__: obj };
+  apply(set, subject2, [testValue]);
+  if (subject2[prop] !== testValue) {
+    return false;
+  }
+  if (!throws(() => apply(set, obj, [testValue]))) {
+    return false;
+  }
+  if ('originalValue' in get) {
+    return false;
+  }
+
+  // We assume that this code runs before any untrusted code runs, so
+  // we do not need to worry about the above conditions passing because of
+  // malicious intent. In fact, it runs even before vetted shims are supposed
+  // to run, between repair and hardening. Given that, after all these tests
+  // pass, we have adequately validated that the property in question is
+  // an accessor function whose purpose is suppressing the override mistake,
+  // i.e., enabling a non-writable property to be overridden by assignment.
+  // In that case, here we *temporarily* turn it into the data property
+  // it seems to emulate, but writable so that it does not trigger the
+  // override mistake while in this temporary state.
+
+  // For those properties that are also listed in `enablements.js`,
+  // that phase will re-enable override for these properties, but
+  // via accessor functions that ses controls, so we know what they are
+  // doing. In addition, the getter functions installed by
+  // `enable-property-overrides.js` have an `originalValue` field
+  // enabling meta-traversal code like harden to visit the original value
+  // without calling the getter.
+
+  if (desc.configurable === false) {
+    // Even though it seems to be an accessor, we're unable to fix it.
+    return false;
+  }
+
+  // Many of the `return false;` cases above plausibly should be turned into
+  // errors, or an least generate warnings. However, for those, the checks
+  // following this phase are likely to signal an error anyway.
+
+  defineProperty(obj, prop, {
+    value: expectedValue,
+    writable: true,
+    enumerable: desc.enumerable,
+    configurable: true,
+  });
+
+  return true;
+};
+
+/**
+ * For each, see it the property is an accessor property that
+ * seems to emulate a data property with this expectedValue,
+ * and seems to be an accessor whose purpose is to protect against
+ * the override mistake, i.e., enable these properties to be overridden
+ * by assignment. If all these expected conditions are met, then
+ * *temporarily* turn it into the data property it emulated.
+ *
+ * Coordinate with `enablements.js` so the appropriate ones are
+ * turned back to accessor that protect against override mistake,
+ * but accessors we know.
+ *
+ * @param {Record<any,any>} intrinsics
+ */
+export const tameFauxDataProperties = intrinsics => {
+  // https://github.com/tc39/proposal-iterator-helpers
+  tameFauxDataProperty(
+    intrinsics['%IteratorPrototype%'],
+    'constructor',
+    intrinsics.Iterator,
+  );
+  // https://github.com/tc39/proposal-iterator-helpers
+  tameFauxDataProperty(
+    intrinsics['%IteratorPrototype%'],
+    toStringTagSymbol,
+    'Iterator',
+  );
+};
--- a/packages/ses/src/tame-faux-data-properties.js
+++ b/packages/ses/src/tame-faux-data-properties.js
@@ -22,44 +22,75 @@ const throws = thunk => {
  * @param {string|symbol} prop
  * @param {any} expectedValue
  * @returns {boolean}
+ * Returns whether `tameFauxDataProperty` turned the property in question
+ * from an apparent faux data property into the actual data property it
+ * seemed to emulate.
+ * If it returns `false`, then we hope no effects happened. However, the whole
+ * point of the validation tests here are to sniff out if an accessor
+ * property seems to be a faux data property. To do we, it invokes the
+ * getter and setter functions that might possibly have side effects.
+ * `tameFauxDataProperty` is not in a position to tell.
  */
 export const tameFauxDataProperty = (obj, prop, expectedValue) => {
   if (obj === undefined) {
+    // The object does not exist in this version of the platform
     return false;
   }
   const desc = getOwnPropertyDescriptor(obj, prop);
   if (!desc || 'value' in desc) {
+    // The property either doesn't exist, or is already an actual data property.
     return false;
   }
   const { get, set } = desc;
   if (typeof get !== 'function') {
+    // A faux data property has both a getter and a setter
     return false;
   }
   if (typeof set !== 'function') {
     return false;
   }
-  const observedValue = get();
-  if (observedValue !== expectedValue) {
+  if (get() !== expectedValue) {
+    // The getter called by itself should produce the expectedValue
     return false;
   }
   if (apply(get, obj, []) !== expectedValue) {
+    // The getter called with `this === obj` should also return the
+    // expectedValue.
     return false;
   }
   const testValue = 'Seems to be a setter';
   const subject1 = { __proto__: null };
   apply(set, subject1, [testValue]);
   if (subject1[prop] !== testValue) {
+    // The setter called with an unrelated object as `this` should
+    // set the property on the object.
     return false;
   }
   const subject2 = { __proto__: obj };
   apply(set, subject2, [testValue]);
   if (subject2[prop] !== testValue) {
+    // The setter called on an object that inherits from `obj` should
+    // override the property from `obj` as if by assignment.
     return false;
   }
-  if (!throws(() => apply(set, obj, [testValue]))) {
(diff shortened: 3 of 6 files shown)


FrederikBolding:
Can confirm this makes lockdown run successfully in Chrome Canary again! 🎉

packages/ses/src/enable-property-overrides.js
```diff
@@ -88,42 +87,46 @@ export default function enablePropertyOverrides(
...
-      }
+      const isDebug = setHas(debugProperties, prop);
+
+      // We use concise method syntax to be `this` sensitive but still
+      // omit a prototype property of [[Construct]] behavior.
+      const { getter, setter } = {
+        getter() {
+          return value;
+        },
+        setter(newValue) {
```
mhofman, line 98: If we're going there, I'd rather go all the way and do
```suggestion
      const { get: getter, set: setter } = getOwnPropertyDescriptor({
        get [prop]() {
          return value;
        },
        set [prop](newValue) {
```

At least the `name` of each accessor would be correct. I know it'd be yet another observable change however.

packages/ses/src/enablements.js
```diff
@@ -152,6 +161,10 @@ export const moderateEnablements = {
 
   '%IteratorPrototype%': {
     toString: true,
+    // https://github.com/tc39/proposal-iterator-helpers
+    constructor: true,
+    // https://github.com/tc39/proposal-iterator-helpers
+    [toStringTagSymbol]: true,
```
mhofman, line 167: Unrelated to this, but I kinda wished our enablements logic really represented "levels" and extended lower layers instead of redefining them.

packages/ses/src/tame-faux-data-properties.js
```diff
@@ -0,0 +1,131 @@
...
+  if (typeof set !== 'function') {
+    return false;
+  }
+  const observedValue = get();
+  if (observedValue !== expectedValue) {
+    return false;
+  }
+  if (apply(get, obj, []) !== expectedValue) {
+    return false;
+  }
```
mhofman, line 47: This feels redundant. I would be fine just testing the second one.

packages/ses/src/tame-faux-data-properties.js
```diff
@@ -0,0 +1,131 @@
...
+  const subject1 = { __proto__: null };
+  apply(set, subject1, [testValue]);
+  if (subject1[prop] !== testValue) {
+    return false;
+  }
+  const subject2 = { __proto__: obj };
+  apply(set, subject2, [testValue]);
+  if (subject2[prop] !== testValue) {
+    return false;
+  }
```
mhofman, line 58: Do we care about the setter behavior?

packages/ses/src/tame-faux-data-properties.js
```diff
@@ -0,0 +1,131 @@
...
+  if (!throws(() => apply(set, obj, [testValue]))) {
+    return false;
+  }
```
mhofman, line 61: Won't this change the target object !?

mhofman (commented):
Preliminary review

As a matter of process, I would really prefer if independent changes were in separate commits. The change to the property override logic is great, but it seems it could easily stand in its own commit.

> None of the existing tests broke, and I manually verified that things work as expected both on Chrome Canary, and on an earlier Chrome browser that has not implemented the iterators-helpers proposal.

You should be able to test this in the [node-v8](https://github.com/nodejs/node-v8) nightly build.

packages/ses/src/tame-faux-data-properties.js
erights: Since we're sniffing, to see if the observed accessor seems to be an attempt at similarly enabling property overrides, yes. If the setter does something different, then the accessor is unlikely to be emulating a data property, so we should not assume we can preserve intended semantics if we turn it into an actual data property. Better to fail in permits-intrinsics.js until a developer can look and understand the anomaly.

Resolving.

packages/ses/src/tame-faux-data-properties.js
erights: Done

packages/ses/src/tame-faux-data-properties.js
```diff
@@ -0,0 +1,124 @@
...
+  const subject2 = { __proto__: obj };
+  apply(set, subject2, [testValue]);
+  if (subject2[prop] !== testValue) {
+    return false;
+  }
+
+  // TODO We would like to test that the setter throws when attempting to
+  // set the target object. However, since we don't already know what it
+  // does, the following test could cause a persistent side effect.
```
mhofman, line 49: So I just read the proposal steps and I'm actually somewhat surprised at the specified behavior for [SetterThatIgnoresPrototypeProperties](https://tc39.es/proposal-iterator-helpers/#sec-SetterThatIgnoresPrototypeProperties). I would expect that if `this === home` it would attempt a define property. Arguably `Iterator.prototype.contstructor = Iterator` should never fail (but also possibly not change the property), and direct assignment to another value shouldn't fail either if `constructor` is configurable.

I know the spec there is exactly how we implemented our property override taming, but we assumed lockdown (which may be a mistake?), but the spec shouldn't assume that the intrinsic prototypes are frozen.

erights:
Ready for review again

packages/ses/test/test-tame-faux-data-properties.js
```diff
@@ -0,0 +1,120 @@
+import test from 'ava';
+import '../index.js';
+
+import { tameFauxDataProperty as tfdp } from '../src/tame-faux-data-properties.js';
+
+const { freeze, defineProperty, getOwnPropertyDescriptor } = Object;
+
+test('unit test tameFauxDataProperty', t => {
```
mhofman, line 8: Can we get messages added to all these test assertions?

packages/ses/src/tame-faux-data-properties.js
mhofman: Given the above, it might be safe to test the setter with the expected value, and restore the throw test? If the setter is not compliant, worst case it overrides with the expected value as a side effect, which should maintain semantics. FWIW, the other setter tests might have side effects too if non compliant. And if we want to be pedantic, calling the getter might have side effects too.

mhofman (approved):
I would be ok with this PR landing as-is but please consider adding messages to the test assertions.

packages/ses/src/tame-faux-data-properties.js
mhofman: Regarding the behavior when assigning to a non-writable property, I forgot that it throws for same value (unlike define to same desc, such a weird discrepancy). It doesn't change my surprise that the spec implements `constructor` as if it was a non-writable property. It also doesn't change my acceptance that we could test the throwing behavior with the expected value. It just changes my understanding of what the spec should do here.

packages/ses/src/tame-faux-data-properties.js
```diff
@@ -0,0 +1,124 @@
...
+  if (typeof get !== 'function') {
+    return false;
+  }
+  if (typeof set !== 'function') {
+    return false;
+  }
+  if (apply(get, obj, []) !== expectedValue) {
+    return false;
+  }
+  const testValue = 'Seems to be a setter';
```
kriskowal, line 35: No change recommended here, but reasoning aloud:

Proposition: maybe this value should be non-forgeable, like `Symbol()` or `{}` so we cannot be confused by a clever implementation.

However, the clever implementation must necessarily be part of the TCB because it runs before SES. We are not looking to avoid confusion from an adversary, only accidental coïncidence.

Also, however, “Seems to be a getter” is very unlikely to collide accidentally.

Also, however, by sending the value to the setter, being non-forgeable is irrelevant: the clever implementation will have the value immediately.

packages/ses/src/tame-faux-data-properties.js
```diff
@@ -0,0 +1,124 @@
...
+ */
+export const tameFauxDataProperty = (obj, prop, expectedValue) => {
+  if (obj === undefined) {
+    return false;
+  }
+  const desc = getOwnPropertyDescriptor(obj, prop);
+  if (!desc || 'value' in desc) {
+    return false;
+  }
+  const { get, set } = desc;
```
kriskowal, line 25: Perhaps a comment: “Here we attempt to discover whether the getter and setter behave like a value property, modulo side-effects we cannot observe and do not care about.”

packages/ses/src/tame-faux-data-properties.js
```diff
@@ -0,0 +1,124 @@
...
+} from './commons.js';
+
+/**
+ * Exported for convenience of unit testing. Harmless, but not expected
+ * to be useful by itself.
+ *
+ * @param {any} obj
+ * @param {string|symbol} prop
+ * @param {any} expectedValue
+ * @returns {boolean}
```
kriskowal, line 15: Consider documenting the meaning of the return value. “The return value is a diagnostic for tests and ignored in the course of `repairIntrinsics()` or `lockdown()`. Returns whether all tests indicated that the property was a getter/setter pair intended to make the property writable objects that inherit the property on their prototype chain, and that it was (possibly temporarily) converted back to a data property.”

Possibly also: “Throws if repairs or lockdown should throw.”

packages/ses/src/tame-faux-data-properties.js
erights: > would attempt a define property

define the property to what? Remember, by assumption, it is an accessor property. If it redefined the property to the data property that would normally be the result of a successful assignment to a data property, then that property, if subsequently made non-writable, will trigger the override mistake. I don't think attempting to define a property on home is an option here.

> direct assignment to another value shouldn't fail either if `constructor` is configurable

Neither assignment failure nor override mistake failures depend at all on configurability. They both depend only on writability. However, as an accessor property emulating a data property storing only the emulated value off to the side, it has no state to track whether it should emulate writable or non-writable. Extending the spec to burden the emulation with tracking emulated writability is, I suppose, possible. But I think it is great we did not do that.

> FWIW ... side effects too.

Good point! I'm going to restore the setter test. But I really like your suggestion to invoke the setter with the already-validated expected value. That way, if it fails by performing the most obvious emulated assignment, it will not actually have changed any observable state. Or so one can hope.

> It just changes my understanding of what the spec should do here

I still think the spec is doing the least bad thing it is possible for the spec to do, given all the constraints. If you still disagree, we should discuss.

packages/ses/src/tame-faux-data-properties.js
```diff
@@ -0,0 +1,124 @@
...
+  // without calling the getter.
+
+  if (desc.configurable === false) {
+    // Even though it seems to be an accessor, we're unable to fix it.
+    return false;
+  }
+
+  // Many of the `return false;` cases above plausibly should be turned into
+  // errors, or an least generate warnings. However, for those, the checks
+  // following this phase are likely to signal an error anyway.
```
kriskowal, line 85: I’m convinced that this is correct. There might be cases where we can throw a more informative error here, that this repair would make the later error confusing, but I can’t imagine what they would be.

packages/ses/src/tame-faux-data-properties.js
```diff
@@ -0,0 +1,124 @@
...
+ * by assignment. If all these expected conditions are met, then
+ * *temporarily* turn it into the data property it emulated.
+ *
+ * Coordinate with `enablements.js` so the appropriate ones are
+ * turned back to accessor that protect against override mistake,
+ * but accessors we know.
+ *
+ * @param {Record<any,any>} intrinsics
+ */
+export const tameFauxDataProperties = intrinsics => {
```
kriskowal, line 111: What distinguishes an “enablement” from a “faux data property”? Are they the same?

packages/ses/src/tame-faux-data-properties.js
kriskowal: I suspect that this sniff is neither helpful nor harmful: not consequential. Given: we are using this repair only on properties we know are either:

1. not implemented
2. implemented as a data property per prior specification (these were all unshipped)
3. implemented as a getter or setter per to specification
4. implemented to a future specification likely to resemble 2 or 3
5. not implemented to specification

The sniff pertains to the last case. In many places, we assume that if a function exists, it behaves according to specification, like `Map.prototype.set`. All such functions are in the TCB. There are many ways that these functions could misbehave, including:

1. ignoring the subject and entangling the getter and setter over otherwise shared state
2. ignoring the set
3. inconsistent get
4. observable side-effects
5. un-observable side-effects

kriskowal (approved):

packages/ses/src/tame-faux-data-properties.js
erights: I mostly accept this reasoning. The point is not to catch malice or purposeful misbehavior. The point is to catch cases where the underlying platform changed since that version of SES was written, and we want to catch if it seems to have only been changed in ways we expect. If we're confident enough of that, then we let ses succeed at initializing. We purposely tolerate all additions that we can remove, because we're confident enough that this is symptomatic of purely additive changes, that don't break our safety once removed.

Right now, having the spec define accessor functions is quite novel for tc39. I do not yet feel confident that tc39 will follow the (IMO good) precedent set by the current state of the iterators-helpers proposal. So I want us to continue to fail early when there is a platform change we need to think about before concluding that it does not harm SES's safety.

> nor harmful: not consequential

I interpret this as I'm clear to proceed ;)

packages/ses/src/enable-property-overrides.js
```diff
@@ -156,23 +155,20 @@ export default function enablePropertyOverrides(
     }
     // TypeScript does not allow symbols to be used as indexes because it
     // cannot recokon types of symbolized properties.
-    // @ts-ignore
     arrayForEach(ownKeys(descs), prop => enable(path, obj, prop, descs[prop]));
   }
 
   function enableProperties(path, obj, plan) {
-    for (const prop of getOwnPropertyNames(plan)) {
+    for (const prop of ownKeys(plan)) {
```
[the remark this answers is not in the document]
erights: Yes, that is the purpose of the change: to add an enablement for `Iterator.prototype[Symbol.toStringTag]` which has now become an accessor.

packages/ses/src/tame-faux-data-properties.js
erights: From this, I infer you're ok with the current strategy? Resolving. But if I misunderstood, please unresolve.

packages/ses/src/tame-faux-data-properties.js
erights: From this, I infer you're ok with the current strategy? Resolving. But if I misunderstood, please unresolve.

kriskowal:
Also: consider adding an entry to `NEWS.md`.

packages/ses/src/tame-faux-data-properties.js
erights: In light of how this thread has gone, I added it back in. But I added a comment to each to explain what it tests.

packages/ses/src/tame-faux-data-properties.js
erights: In light of how this thread has gone, I restored the stricter setter behavior testing. Resolving.

packages/ses/src/tame-faux-data-properties.js
erights: I restored this test, but as you suggested I set it to `expectedValue` to minimize the chance of a harmful inadvertent side effect. Added a comment explaining why.

erights:
> I would be ok with this PR landing as-is but please consider adding messages to the test assertions.

All test assertions are well commented.

packages/ses/src/tame-faux-data-properties.js
erights: In light of this thread, I restored these stricter checks but added comments. Resolving.

packages/ses/test/test-tame-faux-data-properties.js
erights: Done

packages/ses/src/tame-faux-data-properties.js
erights: I added a bunch of comments that I think makes this point well enough with other words. Done.

packages/ses/src/tame-faux-data-properties.js
erights: Your first paragraph, I think I've done with other words.

> Possibly also: “Throws if repairs or lockdown should throw.”

I don't understand this. Optimistically I'm going to resolve anyway. Please let me know if I'm missing something here, even if I've merged by the time you see this.

erights:
> Also: consider adding an entry to `NEWS.md`.

Done. After merge though, please do look it over consistency with your style. Thanks.

mhofman:
Thanks for addressing all the feedback!

PR #1978, opened 2024-01-17
test(ses): add a smoke test for SES lockdown in chrome dev

naugtur:
Update: 
- reworked the setup for browser tests, 
- made it possible to run more than smoke tests, 
- added a proof-of-concept test of lockdown being inconsistent across browsers (skipped because I don't know if we want to do anything about it)
- removed a dependency that was only used for statically serving one file.

[update]
I made chrome nightly available for playwright by using a docker image provided by Selenium org. The ugly consequence is I had to adapt it to GH Actions a bit.


browser-test/tests/repair.spec.js
```diff
@@ -0,0 +1,30 @@
...
+  // Expect a title "to contain" a substring.
+  // await expect(page).toHaveTitle(/Pass/);
+  const result = await page.evaluate(() => {
+    Object.defineProperty(Object.prototype, 'toString', {
+      value: function () {},
+    });
+    try {
+      lockdown();
+    } catch (e) {
+      return e.message;
```
naugtur, line 25: Fails with ` "Cannot delete property 'arguments' of function () {}"` only in chromium

mhofman:
> Fails with ` "Cannot delete property 'arguments' of function () {}"` only in chromium

> @erights @kriskowal - when experimenting, I stumbled upon a cross-browser inconsistency in lockdown

Is there possibly any transform of the code? That smells like the code is running in sloppy mode.

erights:
> > Fails with ` "Cannot delete property 'arguments' of function () {}"` only in chromium
> 
> > @erights @kriskowal - when experimenting, I stumbled upon a cross-browser inconsistency in lockdown
> 
> Is there possibly any transform of the code? That smells like the code is running in sloppy mode.

Agreed. This looks like a sloppy-mode symptom. Would be great to know where it comes from.

erights:
Attn @frazarshad @toliaqat

browser-test/tests/repair.spec.js
kriskowal: attn @erights. My understanding of the above code is that `page.evaluate` calls `code.toString()` and evaluates that in the page. I assume that means this is running in sloppy mode so the overridden `toString` has an `arguments` object. It’s unclear to me why only Chromium would reäct to this code in this way. Maybe a `"use strict"` would make all engines perform uniformly. Is it possible that this is a variation in behavior for the Chromium harness?

kriskowal: @erights I see your reaction from an hour ago. Thanks.

browser-test/server.js
```diff
@@ -0,0 +1,30 @@
...
+const path = require('path');
+const http = require('http');
+const fs = require('fs');
+
+const server = http.createServer((req, res) => {
+  console.log(req.url);
+
+  if (req.url === '/') {
+    res.writeHead(200, { 'Content-Type': 'text/html' });
+    return res.end('<!doctype html><script src="ses.umd.js"></script>');
```
naugtur, line 11: Re slopy mode: this is all the HTML there is in the browser.

browser-test/tests/repair.spec.js
naugtur: Adding `"use strict"` in front of the function does make it consistent again.

naugtur:
Ok, so we've identified that the sloppy mode comes from the way playwright runs a function in the context of the page. It's not that interesting as a result. I'm dropping the test case now, but the conversation should stay intact.

erights:
> Ok, so we've identified that the sloppy mode comes from the way playwright runs a function in the context of the page. It's not that interesting as a result. I'm dropping the test case now, but the conversation should stay intact.

Can/should we send a fix upstream to playwright?

.github/workflows/browser-test.yml
```diff
@@ -4,22 +4,35 @@ on:
     branches: [master]
   pull_request:
     branches: [master]
+  schedule:
+    - cron: '0 0 * * *' # Runs every day at midnight
```
kriskowal, line 8: Nice. I need to figure out how to subscribe to these reports.

.github/workflows/browser-test.yml
```diff
@@ -4,22 +4,35 @@ on:
...
 jobs:
   browser-tests:
     timeout-minutes: 30
     runs-on: ubuntu-latest
+    container:
+      image: selenium/standalone-chrome:nightly
+      options: --user root # reintroducing a GH Actions default expectation
+      ## https://docs.github.com/en/actions/creating-actions/dockerfile-support-for-github-actions#user
+      ## the elenium image is configured to not run as root, so we have to undo some of it
```
kriskowal, line 17:
```suggestion
      ## the Selenium image is configured to not run as root, so we have to undo some of it
```

kriskowal (approved):

.github/workflows/browser-test.yml
naugtur: You get an email when it fails. It's going to be one of the 100s of github notification emails you're getting daily.

naugtur:
> > Ok, so we've identified that the sloppy mode comes from the way playwright runs a function in the context of the page. It's not that interesting as a result. I'm dropping the test case now, but the conversation should stay intact.
> 
> Can/should we send a fix upstream to playwright?

I think they might want to keep the sloppy mode as a possibility and let people who care opt-in, but I'm willing to ask.

erights:
> but I'm willing to ask.

Please do. Thanks.

PR #1982, opened 2024-01-18
perf(base64): Replace random access with iteration

gibson042:
## Description

We recently discovered that in XS, random access to string characters is O(n). This PR updates the JS base64 implementation to instead use iterators. It improves the performance in XS as expected, but irrelevantly so because XS provides a native implementation which is _much_ faster still. There's a slight penalty in Node.js, which may or may not be tolerable (_but note that apples-to-apples comparison across implementations is necessary because the JS input is substantially shorter_). I'm comfortable with rejection of this PR (or at least everything after the first commit), but felt that opening it was important for transparency.

### Security Considerations

n/a

### Scaling Considerations

As noted above. Here are the measured effects:
```diff
 Node.js
-encodes 10775.74909090909 characters per millisecond
+encodes 9801.535832414553 characters per millisecond
-JS encodes 23130.176470588234 characters per millisecond
+JS encodes 18329.573806881242 characters per millisecond
-decodes 48752.36084452975 bytes per millisecond
+decodes 45848.339350180504 bytes per millisecond
-JS decodes 43007.671875 bytes per millisecond
+JS decodes 45448.94117647059 bytes per millisecond
 XS
-encodes 577272.2727272727 characters per millisecond
+encodes 578303.1160714285 characters per millisecond
-JS encodes 1335.143596377749 characters per millisecond
+JS encodes 1700.2734761120264 characters per millisecond
-decodes 396777.0553505535 bytes per millisecond
+decodes 393870.26373626373 bytes per millisecond
-JS decodes 1093.7360890302066 bytes per millisecond
+JS decodes 1946.3762376237623 bytes per millisecond
```

n/a

### Testing Considerations

n/a

### Upgrade Considerations

n/a

### DIFF
--- a/packages/base64/src/decode.js
+++ b/packages/base64/src/decode.js
@@ -22,15 +22,11 @@ import { monodu64, padding } from './common.js';
  * @returns {Uint8Array} decoded bytes
  */
 export const jsDecodeBase64 = (string, name = '<unknown>') => {
-  // eslint-disable-next-line no-nested-ternary
-  const paddingLength = string.endsWith('==')
-    ? 2
-    : string.endsWith('=')
-    ? 1
-    : 0;
-  const data = new Uint8Array(
-    Math.ceil((string.length * 3) / 4) - paddingLength,
-  );
+  // String operations can be O(n) in XS as of 2024-01, so cache the length
+  // and use an iterator rather than sequential access by index.
+  const stringLength = string.length;
+  let paddingLength = 0;
+  const data = new Uint8Array(Math.ceil((stringLength * 3) / 4));
   let register = 0;
   let quantum = 0;
   let i = 0; // index in string
@@ -41,6 +37,7 @@ export const jsDecodeBase64 = (string, name = '<unknown>') => {
     if (done || ch === padding) {
       if (ch !== padding) break;
       done = true;
+      paddingLength += 1;
       quantum -= 2;
       i += 1;
       // Padding is only expected to complete a short chunk of two or three data characters
@@ -69,12 +66,14 @@ export const jsDecodeBase64 = (string, name = '<unknown>') => {
 
   if (quantum !== 0) {
     throw Error(`Missing padding at offset ${i} of string ${name}`);
-  } else if (i < string.length) {
+  } else if (i < stringLength) {
     throw Error(
       `Base64 string has trailing garbage ${string.slice(i)} in string ${name}`,
     );
   }
-  return data;
+  return paddingLength === 0
+    ? data
+    : data.subarray(0, data.length - paddingLength);
 };
 
 // The XS Base64.decode function is faster, but might return ArrayBuffer (not
--- a/packages/base64/src/decode.js
+++ b/packages/base64/src/decode.js
@@ -22,45 +22,59 @@ import { monodu64, padding } from './common.js';
  * @returns {Uint8Array} decoded bytes
  */
 export const jsDecodeBase64 = (string, name = '<unknown>') => {
-  const data = new Uint8Array(Math.ceil((string.length * 4) / 3));
+  // eslint-disable-next-line no-nested-ternary
+  const paddingLength = string.endsWith('==')
+    ? 2
+    : string.endsWith('=')
+    ? 1
+    : 0;
+  const data = new Uint8Array(
+    Math.ceil((string.length * 3) / 4) - paddingLength,
+  );
   let register = 0;
   let quantum = 0;
   let i = 0; // index in string
   let j = 0; // index in data
+  let done = false;
 
-  while (i < string.length && string[i] !== padding) {
-    const number = monodu64[string[i]];
+  for (const ch of string) {
+    if (done || ch === padding) {
+      if (ch !== padding) break;
+      done = true;
+      quantum -= 2;
+      i += 1;
+      // Padding is only expected to complete a short chunk of two or three data characters
+      // (i.e., the last two in the `quantum` cycle of [0, 6, 12 - 8 = 4, 10 - 8 = 2]).
+      if (quantum === 4 || quantum === 2) {
+        // We MAY reject non-zero padding bits, but choose not to.
+        // https://datatracker.ietf.org/doc/html/rfc4648#section-3.5
+        // eslint-disable-next-line no-continue
+        continue;
+      }
+      break;
+    }
+    const number = monodu64[ch];
     if (number === undefined) {
-      throw Error(`Invalid base64 character ${string[i]} in string ${name}`);
+      throw Error(`Invalid base64 character ${ch} in string ${name}`);
     }
     register = (register << 6) | number;
     quantum += 6;
     if (quantum >= 8) {
       quantum -= 8;
-      data[j] = register >>> quantum;
+      data[j] = register >> quantum;
       j += 1;
-      register &= (1 << quantum) - 1;
     }
     i += 1;
   }
 
-  while (i < string.length && quantum % 8 !== 0) {
-    if (string[i] !== padding) {
-      throw Error(`Missing padding at offset ${i} of string ${name}`);
-    }
-    i += 1;
-    quantum += 6;
-  }
-
-  if (i < string.length) {
+  if (quantum !== 0) {
+    throw Error(`Missing padding at offset ${i} of string ${name}`);
+  } else if (i < string.length) {
     throw Error(
-      `Base64 string has trailing garbage ${string.substr(
-        i,
-      )} in string ${name}`,
+      `Base64 string has trailing garbage ${string.slice(i)} in string ${name}`,
     );
   }
-
-  return data.subarray(0, j);
+  return data;
 };
 
 // The XS Base64.decode function is faster, but might return ArrayBuffer (not
--- a/packages/base64/src/encode.js
+++ b/packages/base64/src/encode.js
@@ -27,8 +27,7 @@ export const jsEncodeBase64 = data => {
   let register = 0;
   let quantum = 0;
 
-  for (let i = 0; i < data.length; i += 1) {
-    const b = data[i];
+  for (const b of data) {
     register = (register << 8) | b;
     quantum += 8;
     if (quantum === 24) {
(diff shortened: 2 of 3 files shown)


packages/base64/src/decode.js
```diff
@@ -18,46 +18,55 @@ import { monodu64, padding } from './common.js';
...
-const jsDecodeBase64 = (string, name = '<unknown>') => {
-  const data = new Uint8Array(Math.ceil((string.length * 4) / 3));
+export const jsDecodeBase64 = (string, name = '<unknown>') => {
+  // eslint-disable-next-line no-nested-ternary
+  const paddingLength = string.endsWith('==')
+    ? 2
+    : string.endsWith('=')
+    ? 1
+    : 0;
+  const data = new Uint8Array((string.length * 3) / 4 - paddingLength);
```
mhofman, line 28: I'm confused, why check the end of the string? `new Uint8Array(Math.ceil((string.length * 3) / 4))` should be correct.

Checking the end of the string has 2 issues: it disallows trimming the trailing `=` which is a very common case, and I suspect `endWith` may be `O(n)` as well.

On that note. `string.length` is `O(n)` and looked up twice in this implementation.

packages/base64/src/decode.js
```diff
@@ -18,46 +18,55 @@ import { monodu64, padding } from './common.js';
...
 
-  while (i < string.length && quantum % 8 !== 0) {
-    if (string[i] !== padding) {
-      throw Error(`Missing padding at offset ${i} of string ${name}`);
-    }
-    i += 1;
-    quantum += 6;
-  }
-
   if (i < string.length) {
```
mhofman, line 62: Cache `string.length` at the beginning. But then in which case can there ever be garbage at the end?

packages/base64/src/decode.js
```diff
@@ -18,46 +18,55 @@ import { monodu64, padding } from './common.js';
...
     throw Error(
-      `Base64 string has trailing garbage ${string.substr(
-        i,
-      )} in string ${name}`,
+      `Base64 string has trailing garbage ${string.slice(i)} in string ${name}`,
     );
+  } else if (quantum !== 0) {
+    throw Error(`Missing padding at offset ${i} of string ${name}`);
```
mhofman, line 67: This doesn't need to throw, it just needs to output the register.

packages/base64/test/bench-main.js
```diff
@@ -28,13 +32,32 @@ const data = encodeBase64(string);
...
-  console.log(
+  log(
     'encodes',
     (operations * string.length) / duration,
     'characters per millisecond',
   );
 }
 
+{
+  const start = Date.now();
```
mhofman, line 43: I wish we used a slightly more precise `performance.now()` if available.

mhofman (commented):
Something seems fishy with the bench results. From what I gather, under node.js, the `encodes` and `JS encodes` should be using the same implementation, yet they have drastically different results.

> ```diff
>  Node.js
> -encodes 10775.74909090909 characters per millisecond
> +encodes 9801.535832414553 characters per millisecond
> -JS encodes 23130.176470588234 characters per millisecond
> +JS encodes 18329.573806881242 characters per millisecond
> ```

kriskowal:
Thank you for digging into this and validating the hypothesis that `string[i]` was `O(i)` on XS. I’m inclined to take the performance win on Node.js with the existing code and the performance win on XS with the native implementation, archiving this PR as evidence for future reference.

packages/base64/src/decode.js
gibson042: > I'm confused, why check the end of the string? `new Uint8Array(Math.ceil((string.length * 3) / 4))` should be correct.

"Cg==", "QQo=", and "QUIK" all have length 4, but encode different-length data (respectively, 1, 2, and 3 octets). Trailing padding is necessary to differentiate them.

> Checking the end of the string has 2 issues: it disallows trimming the trailing `=` which is a very common case,

...but not one supported by this code.

> and I suspect `endWith` may be `O(n)` as well.
> 
> On that note. `string.length` is `O(n)` and looked up twice in this implementation.

Hrm, I thought not but I just double-checked and you're right about both. Updated.

packages/base64/src/decode.js
gibson042: When a non-alphabet character is encountered, or anything after expected padding.

packages/base64/src/decode.js
gibson042: No, this remains a strict implementation that requires padding.

packages/base64/test/bench-main.js
gibson042: Yeah, that bugged me too. Done.

gibson042:
> Something seems fishy with the bench results. From what I gather, under node.js, the `encodes` and `JS encodes` should be using the same implementation, yet they have drastically different results.

Updated the benchmark script to clarify with better output:
```console
$ node test/bench-main.js
[pass 1] encodes 8378.88124410933 characters per millisecond
[pass 2] encodes 8523.48322147651 characters per millisecond
[pass 1] JS short-string encodes 18514.513452914798 characters per millisecond
[pass 2] JS short-string encodes 18514.513452914798 characters per millisecond
[pass 1] decodes 45357.107142857145 bytes per millisecond
[pass 2] decodes 51212.96780487805 bytes per millisecond
[pass 1] JS short-string decodes 43996.02797202797 bytes per millisecond
[pass 2] JS short-string decodes 43261.155206286836 bytes per millisecond
```

> I’m inclined to take the performance win on Node.js with the existing code and the performance win on XS with the native implementation, archiving this PR as evidence for future reference.

Makes sense. Should I open a new PR for the benchmarking script improvements?

kriskowal:
> Makes sense. Should I open a new PR for the benchmarking script improvements?

That would be excellent.

packages/base64/src/decode.js
mhofman: Don't we throw if an invalid character is encountered?

One case where the number of characters iterated would differ from the string length is if any surrogate pairs are encountered, but again these would be invalid characters, and we'd have thrown already.

I maintain that this condition can never be true.

mhofman:
> ```
> [pass 1] encodes 8378.88124410933 characters per millisecond
> [pass 2] encodes 8523.48322147651 characters per millisecond
> ```

I don't understand how the new changes you made dropped the performance in node even more.

packages/base64/src/decode.js
gibson042:
```console
  import "@endo/init";
  import { jsDecodeBase64 } from "./src/decode.js";
  console.log("valid", jsDecodeBase64("QQ=="));
  console.log("invalid");
  jsDecodeBase64("QQ===");
'
valid Uint8Array(1) [ 65 ]
invalid
(Error#1)
Error#1: Base64 string has trailing garbage = in string <unknown>

  at jsDecodeBase64 (packages/base64/src/decode.js:62:11)
  at packages/base64/[eval1]:6:3
```

packages/base64/test/bench-main.js
```diff
@@ -1,16 +1,24 @@
...
 import { encodeBase64, decodeBase64 } from '../index.js';
 import { jsEncodeBase64 } from '../src/encode.js';
 import { jsDecodeBase64 } from '../src/decode.js';
 
 // eslint-disable-next-line no-undef,no-restricted-globals
 const log = (typeof console !== 'undefined' && console.log) || print;
+const now = (() => {
+  try {
+    // eslint-disable-next-line no-undef,global-require
+    return require('perf_hooks').performance.now || Date.now;
```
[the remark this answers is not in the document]
gibson042: Eh, I just went ahead with moving to async.

gibson042:
> Should I open a new PR for the benchmarking script improvements?

#1986

packages/base64/src/decode.js
mhofman: Oh right I missed the new break logic from the loop. I believe there's actually a bug allowing input like `QQ=*`

mhofman: Oh and it doesn't seem to check the register is empty once padding is reached

erights:
I am confused about the separation between this PR and https://github.com/endojs/endo/pull/1986 , Both contain all three files. I would have expected one (that one) to define the benchmarking framework, and the other (this one) to be staged on it and use it.

packages/base64/src/decode.js
mhofman: It looks like [according to the RFC](https://datatracker.ietf.org/doc/html/rfc4648#section-3.5), a decoder "MAY" reject if the padding bits are not 0, but doesn't have to. I still believe this implementation will accept `QQ=*` for example, which is not compliant.

gibson042: > I believe there's actually a bug allowing input like `QQ=*`

Nice catch! Opened #1991 to add coverage and rebased this PR on top of its test changes.

gibson042:
> I am confused about the separation between this PR and #1986 , Both contain all three files. I would have expected one (that one) to define the benchmarking framework, and the other (this one) to be staged on it and use it.

@erights Good point; this is now the case.

erights:
Now that #1986 is merged, if it is ok with the other reviewers, I'd prefer that this PR be rebased on master if it is not already based on a state that includes #1986

***Reviewers***, is that ok with you?

gibson042:
#1986 is already incorporated, but it would be a good idea to land #1991 and then rebase this PR (which already shares its packages/base64/test/test-main.js) on top of the resulting master.

erights:
> #1986 is already incorporated, but it would be a good idea to land #1991 and then rebase this PR (which already shares its packages/base64/test/test-main.js) on top of the resulting master.

Sounds good to me

kriskowal:
@gibson042 Are you shooting to improve performance on both Node.js and XS (and assuming this will still fall thru to the native version on XS)?

gibson042:
@kriskowal Since the two are at odds and the JS implementation isn't even used by XS, I'm comfortable with getting this PR clean and then closing it without merge as you suggested in https://github.com/endojs/endo/pull/1982#issuecomment-1899159764
> Thank you for digging into this and validating the hypothesis that `string[i]` was _O(i)_ on XS. I’m inclined to take the performance win on Node.js with the existing code and the performance win on XS with the native implementation, archiving this PR as evidence for future reference.

PR #1983, opened 2024-01-18
refactor(ses): Move makeLRUCacheMap to a dedicated file

gibson042:
## Description

`makeLRUCacheMap` is only _used_ by the SES error system, but is not really specific to it. It should eventually be an endo export, possibly in its own package (just not in `common`, which itself depends upon `ses`), but this PR just promotes it up one level into an easily-movable file for when that time comes.

### Security Considerations

This code is depended upon by SES assert code, and therefore cannot itself assume a hardened environment. All of the existing commentary and `freeze` calls were preserved.

### Scaling Considerations

n/a

None yet.

### Testing Considerations

Testing was consolidated into packages/ses/test/test-lru-cache-map.js.

### Upgrade Considerations

n/a


erights (approved):
LGTM, much better, thanks!

Given that it has to work before lockdown, it itself does not to be exported in that form. Rather, a trivial but post-lockdown wrapper of this could be what it exported.

OTOH, given adequate use of freeze to emulate full hardening, exporting from a lower layer package is fine. As you say, it cannot be `@endo/common` because of this layering. In any case, this PR is the right first step.

PR #1986, opened 2024-01-19
chore(base64): Improve benchmark script

gibson042:
## Description

* Capture results
* Make two passes per scenario
* Clarify full vs. short-input scenarios and run both against the same loop body
* Use performance.now() for benchmarking where available (e.g. Node.js)

### Security Considerations

n/a

### Scaling Considerations

n/a, except for potentially providing data to guide scaling improvements.

n/a

### Testing Considerations

js{Encode,Decode}Base64 are now exported internally within the package so they may be benchmarked.

### Upgrade Considerations

n/a

### DIFF
--- a/packages/base64/src/decode.js
+++ b/packages/base64/src/decode.js
@@ -13,12 +13,15 @@ import { monodu64, padding } from './common.js';
  * arranges a native binding on the global object.
  * We use that if it is available instead.
  *
+ * This function is exported from this *file* for use in benchmarking,
+ * but is not part of the *module*'s public API.
+ *
  * @param {string} string Base64-encoded string
  * @param {string} [name] The name of the string as it will appear in error
  * messages.
  * @returns {Uint8Array} decoded bytes
  */
-const jsDecodeBase64 = (string, name = '<unknown>') => {
+export const jsDecodeBase64 = (string, name = '<unknown>') => {
   const data = new Uint8Array(Math.ceil((string.length * 4) / 3));
   let register = 0;
   let quantum = 0;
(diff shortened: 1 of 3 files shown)


packages/base64/test/bench-main.js
```diff
@@ -1,55 +1,106 @@
...
+  const log = (typeof console !== 'undefined' && console.log) || print;
+  /** @type {typeof Date.now} */
+  const now = await (async () => {
+    try {
+      const { performance } = await requireOrImport('perf_hooks');
+      if (performance.now) {
+        return performance.now.bind(performance);
+      }
+    } catch (_err) {
+      // eslint-disable-next-line no-empty
```
kumavis, line 37: maybe ignore a more specific error here

kumavis (approved):

packages/base64/src/decode.js
```diff
@@ -18,7 +18,7 @@ import { monodu64, padding } from './common.js';
  * messages.
  * @returns {Uint8Array} decoded bytes
  */
-const jsDecodeBase64 = (string, name = '<unknown>') => {
+export const jsDecodeBase64 = (string, name = '<unknown>') => {
```
kriskowal, line 21: It might be worth leaving a comment because this justifies the existence of the thunk at `packages/base64/decode.js` that just exports the portion of this module that is public.

packages/base64/test/bench-main.js
```diff
@@ -1,55 +1,106 @@
...
+ * @param {'perf_hooks'} specifier
+ */
+const requireOrImport = specifier => {
+  try {
+    // eslint-disable-next-line no-undef,global-require,import/no-dynamic-require
+    return require(specifier);
+  } catch (err) {
+    return import(specifier);
+  }
+};
```
kriskowal, line 25: What environments are we targeting? Which one needs `require`?

kriskowal (approved):

Issue #1987, opened 2024-01-19
`@endo/cli`: Consider renaming `restart` and/or `reset` commands


rekmarks:
The situation:
- `endo restart` reboots the daemon, destroying all ephemeral state (i.e. memory).
- `endo reset` performs a "factory reset" of the daemon, destroying all ephemeral _and_ persistent state.

I find "restart" and "reset" to be both phonetically and semantically similar words, and suggest that we rename one or both. In particular, `reset` is a pretty innocuous name for the most destructive command of the CLI.

Note that there is also a command `clean` which destroys all ephemeral state without restarting the daemon.

Here are some alternatives:
`restart`:
- `reboot`
- ~`reload`~ (IMO, inappropriate because does not imply turning it off and on again)

`reset`:
- `wipe`
- `purge`
- ~`clear`~ (too similar to `clean`)

kriskowal:
I agree that `reset` is too similar to `restart`.

I would like to keep `restart` since it’s in the `stop` and `start` family.

I am hoping that `clean` has outlived its usefulness and that it can just be deleted.

I would like to reserve `clear` for caches.

I like the absoluteness of `wipe` and `purge`.

I could also buy `factory-reset` by analogy, forgetting all the current user’s state so it can be passed to a new person.

I could also buy `endo clear [--state] [--cache]` etc.

rekmarks:
After ruminating on this some more, I'm partial to `purge`. After all, this is spiritually the "nuke it from orbit" command.

> I would like to keep `restart` since it’s in the `stop` and `start` family.
I find this to be a strong argument for keeping `restart`.

dckc:
I kinda like `purge`, but IME `--purge` is an option to modify something like `uninstall`.
That makes me lean toward `endo clear [--state] [--cache]`.

Hm... it seems that `apt-get purge packagename` now means the same as `apt-get remove --purge packagename`.

Either of these is **vastly** better than the hazard of `reset`.

Even `purge` should probably have an "are you sure?" thing and a `--force` option to skip the confirmation a la [docker system prune](https://docs.docker.com/engine/reference/commandline/system_prune/).

I mean: the premise of this thing is that users record all sorts of policy / authorization info in here. To have it all go "poof!" with one command would be like having a password manager that looses all your passwords if you accidentally click a button... a button which is on the screen during normal usage.

[...]

PR #1989, opened 2024-01-22
refactor(cli): Rename `reset` command to `purge`

rekmarks:
Closes: #1987 

## Description

Renames the CLI `reset` command to `purge` in order to better distinguish it from `restart`. Also adds a confirmation prompt to the command, which can be overridden with the `-f`/`--force` flag.

### Security Considerations

N/A

### Scaling Considerations

N/A

N/A

### Testing Considerations

Would it be helpful to start adding unit tests for anything where that's straightforward? cc: @kriskowal 

### Upgrade Considerations

This is breaking for `@endo/cli`, but I'm not sure what we're doing about breaking changes at this stage, if anything.

### DIFF
--- a/packages/cli/src/endo.js
+++ b/packages/cli/src/endo.js
@@ -599,7 +599,9 @@ export const main = async rawArgs => {
       const doPurge =
         force ||
         /^y(es)?$/u.test(
-          await prompt('Are you sure you want to erase all state? (y/n)'),
+          await prompt(
+            'Are you sure you want to erase all state? This irreversible action will permanently sever all peer connections. Continue? (y/n)',
+          ),
         );
       if (doPurge) {
         const { reset } = await import('@endo/daemon');
--- a/packages/cli/src/endo.js
+++ b/packages/cli/src/endo.js
@@ -10,6 +10,7 @@ import fs from 'fs';
 import url from 'url';
 
 import { Command } from 'commander';
+import { prompt } from './prompt.js';
 
 const collect = (value, values) => values.concat([value]);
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/prompt.js
```diff
@@ -0,0 +1,23 @@
+import { stdin, stdout } from 'process';
+import readline from 'readline';
```
kriskowal, line 2: We are catching a lucky break here. Most commands retain `ses` because it is necessary to establish a CapTP connection to the daemon, but this one does not. `readline` is not compatible with SES because it retains `domain`. This will work for now, but it might be worth finding an alternative.

packages/cli/src/endo.js
```diff
@@ -590,11 +591,20 @@ export const main = async rawArgs => {
...
     .description('erases persistent state and restarts if running')
-    .action(async _cmd => {
-      const { reset } = await import('@endo/daemon');
-      await reset();
+    .action(async cmd => {
+      const { force } = cmd.opts();
+      const doPurge =
+        force ||
+        /^y(es)?$/u.test(
+          await prompt('Are you sure you want to erase all state? (y/n)'),
```
kriskowal, line 602: Might be worth noting that this is not reversible and will permanently sever any connections to previously known peers. The formula identifiers are (or soon will be, working copy) rooted at a random seed that is regenerated after a purge.

kriskowal (approved):
I started looking into CLI unit tests using `execa` but they were too flappy. Some work needs to be done to make the CLI tests as reliable as the daemon tests (and the daemon tests don’t clean up properly—needs investigation.)

My current idiom for local debugging for a daemon test is `pkill node ; rm -rf tmp ; yarn test -m 'broken test'` and that is *not ideal*.

packages/cli/src/endo.js
rekmarks: Like this, perhaps?
```suggestion
          await prompt(
            'Are you sure you want to erase all state? This irreversible action will permanently sever all peer connections and destroy all formulae. Continue? (y/n)'
          ),
```

packages/cli/src/prompt.js
rekmarks: I'll leave this alone for now, but we may roll our own SES-compatible `readline` later, or hope that Node removes the `domain` dependency from their implementation.

PR #1991, opened 2024-01-22
fix(base64): Reject too-short encodings

### LINKED ISSUE #1990: jsDecodeBase64 fails to reject short encodings
## Describe the bug

`jsDecodeBase64` incorrectly accepts short encodings that leave bits hanging in the register, and the incorrectness is visible in Node.js.

## Steps to reproduce

```console
  import "@endo/init";
  import { decodeBase64 } from "@endo/base64";
  console.log(decodeBase64("Z"));
  console.log(decodeBase64("V1RGZ"));
'
```

## Expected behavior

An error.

## Actual behavior

```
Uint8Array(0) []
Uint8Array(3) [ 87, 84, 70 ]
```

gibson042:
Fixes #1990

## Description

Test invalid encodings, and reject input that is too short.

### Security Considerations

None known.

### Scaling Considerations

n/a

Should this have a CHANGELOG entry?

### Testing Considerations

New and improved!

### Upgrade Considerations

n/a

### DIFF
--- a/packages/base64/src/decode.js
+++ b/packages/base64/src/decode.js
@@ -44,14 +44,14 @@ export const jsDecodeBase64 = (string, name = '<unknown>') => {
     i += 1;
   }
 
-  while (quantum % 8 !== 0) {
+  while (quantum > 0) {
     if (i === string.length || string[i] !== padding) {
       throw Error(`Missing padding at offset ${i} of string ${name}`);
     }
     // We MAY reject non-zero padding bits, but choose not to.
     // https://datatracker.ietf.org/doc/html/rfc4648#section-3.5
     i += 1;
-    quantum += 6;
+    quantum -= 2;
   }
 
   if (i < string.length) {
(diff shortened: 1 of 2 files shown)


kriskowal:
With this change, do we accept all strings we previously generated? That is, do we generate the appropriate padding in all cases?

kriskowal (approved):

gibson042:
@kriskowal This is good without a CHANGELOG.md entry, right?

kriskowal:
> @kriskowal This is good without a CHANGELOG.md entry, right?

This is fine without a `NEWS.md` entry (`CHANGELOG.md` gets generated from commit messages at time of release.)

I prefer (and do not insist) on redundantly capturing messages to upgraders in `NEWS.md` if the dependant package needs to make some changes to compensate for an upgrade. Like, if a method is deprecated, there should be a recommendation to use a different equivalent behavior (or stop that). Likewise for removal of deprecated methods.

kriskowal (approved):

PR #1997, opened 2024-01-23
feat(daemon): pet store list entries

kumavis:
- Adds `parseFormulaIdentifier` utility
- Adds  `{type, number}` formula identifier pair as `value` to petstore `follow()` entries
- Adds `listEntries` to petstore in `Object.fromEntries` with `{type, number}` formula identifier pair as value
- Provides `followEntries` as an alias to `follow`
- Exposes `listEntries` and `followEntries` on host and guest

I'm sending the `{type, number}` formula identifier pair instead of the string formula identifier. Alternatively, we could just the formula identifier.

### DIFF
--- a/packages/daemon/src/formula-identifier.js
+++ b/packages/daemon/src/formula-identifier.js
@@ -1,0 +1,29 @@
+const { quote: q } = assert;
+
+const numberlessFormulasIdentifiers = new Set([
+  'pet-store',
+  'host',
+  'endo',
+  'least-authority',
+  'web-page-js',
+]);
+
+/**
+ * @param {string} formulaIdentifier
+ * @returns {import("./types").FormulaIdentifierRecord}
+ */
+export const parseFormulaIdentifier = formulaIdentifier => {
+  const delimiterIndex = formulaIdentifier.indexOf(':');
+  if (delimiterIndex < 0) {
+    if (numberlessFormulasIdentifiers.has(formulaIdentifier)) {
+      return { type: formulaIdentifier, number: '' };
+    } else {
+      throw new TypeError(
+        `Formula identifier must have a colon: ${q(formulaIdentifier)}`,
+      );
+    }
+  }
+  const type = formulaIdentifier.slice(0, delimiterIndex);
+  const number = formulaIdentifier.slice(delimiterIndex + 1);
+  return { type, number };
+};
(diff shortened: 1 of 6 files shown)


packages/daemon/src/formula-identifier.js
```diff
@@ -0,0 +1,29 @@
+const { quote: q } = assert;
+
+const numberlessFormularIdentifiers = new Set([
+  'pet-store',
+  'host',
+  'endo',
+  'least-authority',
+  'web-page-js',
+]);
```
kriskowal, line 9: Aside: All of these will have numbers in order to make their formula identifiers unforgeable.

packages/daemon/src/pet-store.js
```diff
@@ -104,22 +105,59 @@ export const makePetStoreMaker = (filePowers, locator) => {
       const petNamePath = filePowers.joinPath(petNameDirectoryPath, petName);
       const petNameText = `${formulaIdentifier}\n`;
       await filePowers.writeFileText(petNamePath, petNameText);
-      changesTopic.publisher.next({ add: petName });
+      const formularIdentifierPair = parseFormulaIdentifier(formulaIdentifier);
+      changesTopic.publisher.next({
+        add: petName,
+        value: formularIdentifierPair,
```
kriskowal, line 111: Let’s call it a `FormulaIdentifierRecord` or `FormulaIdentifierEntry`. Pair is fragile if we add a `subtype` field or whatnot.

packages/daemon/src/pet-store.js
```diff
@@ -104,22 +105,59 @@ export const makePetStoreMaker = (filePowers, locator) => {
...
+      });
     };
 
-    const list = () => harden([...petNames.keys()].sort());
+    /**
+     * @param {string} petName
+     * @returns {import('./types.js').FormulaIdentifierPair}
+     */
+    const formulaIdentifierPairForPetname = petName => {
```
kriskowal, line 119: I’ve elsewhere used `Name` consistently to mean pet name in method names. I’m not against expanding these all to `PetName` but would want to do it in a holistic style change.

```suggestion
    const formulaIdentifierPairForName = petName => {
```

kriskowal (changes requested):
Requesting some small changes to make my next rebase easier.

kriskowal (approved):
Cool. Maybe we’ll someday reduce `FormulaIdentifierRecord` to `FormulaDescriptor`.

Issue #1984, opened 2024-01-18
Refactor code for XS O(n) string operations


gibson042:
## Describe the bug

As observed in #1982, many if not all XS string operations are O(n), affecting in particular code that performs successive random access like e.g. `for (let i = 0; i < string.length; i += 1) { const ch = string[i]; … }`. We should therefore avoid such patterns in favor of iterators in any potentially performance-sensitive code:
* [encodePassable.js `decodeArray`](https://github.com/endojs/endo/blob/98d6ebed7d8f202458ed11041d16a6b4a8f71a20/packages/marshal/src/encodePassable.js#L246)
* [compartment-map.js `pathCompare`](https://github.com/endojs/endo/blob/98d6ebed7d8f202458ed11041d16a6b4a8f71a20/packages/compartment-mapper/src/compartment-map.js#L63)
* [maybe?] [nested-evaluate-and-get-exports.js `longestCommonPrefix`](https://github.com/endojs/endo/blob/98d6ebed7d8f202458ed11041d16a6b4a8f71a20/packages/bundle-source/src/nested-evaluate-and-get-exports.js#L23)
* [cjs-module-analyzer `parseSource`](https://github.com/endojs/endo/blob/98d6ebed7d8f202458ed11041d16a6b4a8f71a20/packages/cjs-module-analyzer/index.js#L102)

We should also evaluate successive `replace` calls vs. a single call with a replacer function:
* [evasive-transform `transformComment`](https://github.com/endojs/endo/blob/98d6ebed7d8f202458ed11041d16a6b4a8f71a20/packages/evasive-transform/src/transform-comment.js#L31)

PR #2000, opened 2024-01-23
perf(syrup): Minimize expensive operations when encoding

gibson042:
refs: #1984

## Description

`encodeString`:
* Avoid redundant encoding.
* Cache `length` to avoid re-reading.
* Anticipate the likely count of prefix bytes to avoid moving data.

`encodeArray`:
* Account for element minimum size in encodeArray buffer expansion

`encodeAny`:
* Pass diagnostic path data as an array rather than a string 
   (In the absence of errors, the associated string concatenation doesn't even take place.)

<details><summary>Effect</summary>

```shell
esbench -n 100 --repetitions 10 --eshost-option '-h V8,*XS' --eshost-option '-m' \
--init-file <(printf '%s\n' '
    import { smuggle } from "data:text/javascript,delete globalThis.harden; if(typeof TextEncoder === `undefined`) Object.defineProperties(globalThis, { TextDecoder: { writable: true, enumerable: false, configurable: true, value: class TextDecoder {} }, TextEncoder: { writable: true, enumerable: false, configurable: true, value: class TextEncoder { encodeInto(str, buf){ const bufLen = buf.length; let read = 0, written = 0; if (bufLen > 0) for(const ch of str){ const cp = ch.codePointAt(0), n = cp >= 0x10000 ? 4 : cp >= 0x800 ? 3 : cp >= 0x80 ? 2 : 1; if(written + n > bufLen) break; read += ch.length; written += n; if(written === bufLen) break; } return { read, written }; } } }, }); globalThis.console ||= Object.fromEntries(`debug log info warn error groupCollapsed groupEnd`.split(` `).map(m => [m, print])); /* abuse console.clear to smuggle values */ let smuggled; export const smuggle = val => { smuggled = val; }; { globalThis.process ||= { env: {} }; process.env.LOCKDOWN_CONSOLE_TAMING = `unsafe`; console.clear = () => smuggled; }";
    import "@endo/init";
    import { encodeSyrup } from "@endo/syrup";
    smuggle({ encodeSyrup });
  ' | \
  npx rollup@4.9.5 -p @rollup/plugin-node-resolve -p ~/lib/rollup-plugin-js-data-url.js -f iife | \
  npx terser@v5.26.0 -cm --toplevel
) \
'
  const { encodeSyrup } = console.clear(), capacity = 1;
  const str = "A".repeat(42), short = Array(5).fill(str), mid = Array(100).fill(str), long = Array(1000).fill(str);
' '{
  encodeShort: `result = encodeSyrup(short, { length: capacity });`,
  encodeMid: `result = encodeSyrup(mid, { length: capacity });`,
  encodeLong: `result = encodeSyrup(long, { length: capacity });`,
}'
```

Before:
```console
#### Moddable XS
encodeShort: 2.08 ops/ms
encodeMid: 0.22 ops/ms
encodeLong: 0.02 ops/ms

#### V8
encodeShort: 8.33 ops/ms
encodeMid: 1.14 ops/ms
encodeLong: 0.12 ops/ms
```

After:
```console
#### Moddable XS
encodeShort: 3.03 ops/ms
encodeMid: 0.23 ops/ms
encodeLong: 0.02 ops/ms

#### V8
encodeShort: 10.00 ops/ms
encodeMid: 1.35 ops/ms
encodeLong: 0.16 ops/ms
```

</details>

### Security Considerations

None known.

### Scaling Considerations

For large strings, performance is still dominated by currently-unavoidable O(n) operations.

n/a

### Testing Considerations

n/a

### Upgrade Considerations

n/a

### DIFF
--- a/packages/syrup/src/encode.js
+++ b/packages/syrup/src/encode.js
@@ -56,24 +56,34 @@ function grow(buffer, increaseBy) {
  * @param {string} value
  */
 function encodeString(buffer, value) {
-  const start = buffer.length;
-  for (;;) {
-    const { read = 0, written = 0 } = textEncoder.encodeInto(
-      value,
-      buffer.bytes.subarray(start),
+  const stringLength = value.length;
+  const likelyPrefixLength = `${stringLength}`.length + 1;
+  // buffer.length will be incorrect until we fix it before returning.
+  const start = grow(buffer, likelyPrefixLength + stringLength);
+  const likelyDataStart = start + likelyPrefixLength;
+
+  for (let remaining = value, read = 0, written = 0; ; ) {
+    const chunk = textEncoder.encodeInto(
+      remaining,
+      buffer.bytes.subarray(likelyDataStart + written),
     );
-    if (read === value.length) {
+    written += chunk.written || 0;
+    read += chunk.read || 0;
+    if (read === stringLength) {
       const prefix = `${written}"`; // length prefix quote suffix
+      const prefixLength = prefix.length;
       buffer.length = start;
-      grow(buffer, prefix.length + written);
-      buffer.bytes.copyWithin(start + prefix.length, start); // shift right
+      grow(buffer, prefixLength + written);
+      if (prefixLength !== likelyPrefixLength) {
+        buffer.bytes.copyWithin(start + prefixLength, likelyDataStart); // shift right
+      }
       textEncoder.encodeInto(
         prefix,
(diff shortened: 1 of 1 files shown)


kriskowal (approved):

kriskowal:
(I don’t advise spending a great deal of time on `@endo/syrup`. I suspect that the next run at Syrup, if one should occur, is to turn it into a streaming codec to avoid allocating an intermediate representation.)

PR #2001, opened 2024-01-23
perf(marshal): Replace XS-expensive string operations

gibson042:
Ref #1982

## Description

Speeds up by 21% decoding a representation of the following CopyRecord:
```
{
  primitives: [undefined, null, true, false, 0, 1, 0n, 1n, "foo", Symbol.for("bar")],
  arrays: [[], [{}]],
  records: { empty: {}, singular: { array: [] } },
}
```
(and speeds up decoding a CopyArray containing 10 copies thereof by about 100%)

No significant effects in Node.js.

### Security Considerations

n/a

### Scaling Considerations

As stated on the tin.

n/a

### Testing Considerations

n/a

n/a

### Upgrade Considerations

n/a

### DIFF
--- a/packages/marshal/src/encodePassable.js
+++ b/packages/marshal/src/encodePassable.js
@@ -295,9 +295,9 @@ const decodeRecord = (encoded, decodePassable) => {
   assert(encoded.startsWith('('));
   // Skip the "(" inside `decodeArray` to avoid slow `substring` in XS.
   // https://github.com/endojs/endo/issues/1984
-  const keysvals = decodeArray(encoded, decodePassable, 1);
-  keysvals.length === 2 || Fail`expected keys,values pair: ${encoded}`;
-  const [keys, vals] = keysvals;
+  const unzippedEntries = decodeArray(encoded, decodePassable, 1);
+  unzippedEntries.length === 2 || Fail`expected keys,values pair: ${encoded}`;
+  const [keys, vals] = unzippedEntries;
 
   (passStyleOf(keys) === 'copyArray' &&
     passStyleOf(vals) === 'copyArray' &&
@@ -317,9 +317,9 @@ const decodeTagged = (encoded, decodePassable) => {
   assert(encoded.startsWith(':'));
   // Skip the ":" inside `decodeArray` to avoid slow `substring` in XS.
   // https://github.com/endojs/endo/issues/1984
-  const tagpayload = decodeArray(encoded, decodePassable, 1);
-  tagpayload.length === 2 || Fail`expected tag,payload pair: ${encoded}`;
-  const [tag, payload] = tagpayload;
+  const taggedPayload = decodeArray(encoded, decodePassable, 1);
+  taggedPayload.length === 2 || Fail`expected tag,payload pair: ${encoded}`;
+  const [tag, payload] = taggedPayload;
   passStyleOf(tag) === 'string' ||
     Fail`not a valid tagged encoding: ${encoded}`;
   return makeTagged(tag, payload);
--- a/packages/marshal/src/encodePassable.js
+++ b/packages/marshal/src/encodePassable.js
@@ -53,6 +53,10 @@ export const recordValues = (record, names) =>
   harden(names.map(name => record[name]));
 harden(recordValues);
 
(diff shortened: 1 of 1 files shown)


packages/marshal/src/encodePassable.js
```diff
@@ -53,18 +53,21 @@ export const recordValues = (record, names) =>
   harden(names.map(name => record[name]));
 harden(recordValues);
 
+const zeroes = Array(16)
+  .fill(undefined)
+  .map((_, i) => '0'.repeat(i));
```
erights, line 58: Why is this better than a literal?

erights: Nevermind

packages/marshal/src/encodePassable.js
```diff
@@ -298,7 +315,9 @@ const encodeTagged = (tagged, encodePassable) =>
 
 const decodeTagged = (encoded, decodePassable) => {
   assert(encoded.startsWith(':'));
-  const tagpayload = decodeArray(encoded.substring(1), decodePassable);
+  // Skip the ":" inside `decodeArray` to avoid slow `substring` in XS.
+  // https://github.com/endojs/endo/issues/1984
+  const tagpayload = decodeArray(encoded, decodePassable, 1);
```
erights, line 320:
```suggestion
  const tagPayload = decodeArray(encoded, decodePassable, 1);
```

Or do you have a convention for when you avoid midcaps? Why? I also notice `keysvals` above.

packages/marshal/src/encodePassable.js
```diff
@@ -241,32 +244,44 @@ const encodeArray = (array, encodePassable) => {
...
     } else if (c === '\u0001') {
-      i += 1;
-      i < encoded.length || Fail`unexpected end of encoding ${encoded}`;
-      const c2 = encoded[i];
-      c2 === '\u0000' ||
-        c2 === '\u0001' ||
-        Fail`Unexpected character after u0001 escape: ${c2}`;
-      elemChars.push(c2);
+      inEscape = true;
+      // eslint-disable-next-line no-continue
```
erights, line 275: Not a comment on this PR, but curious: Do you think we should turn off warning on `continue`? I don't see the value in it.

packages/marshal/src/encodePassable.js
```diff
@@ -241,32 +244,44 @@ const encodeArray = (array, encodePassable) => {
...
+      continue;
     } else {
       elemChars.push(c);
     }
+    inEscape = false;
   }
-  elemChars.length === 0 || Fail`encoding terminated early: ${encoded}`;
+  !inEscape || Fail`unexpected end of encoding ${encoded.slice(skip)}`;
```
erights, line 282: Good point! Expensive operations in that position are fine.

erights (approved):
LGTM, thanks!

decodePassable with need these same changes, yes?

packages/marshal/src/encodePassable.js
gibson042: No, it was preexisting (and ~~blame~~ praise) identifies introduction in the landmark https://github.com/Agoric/agoric-sdk/pull/4468). But worth fixing here, which I have done for both sites.

packages/marshal/src/encodePassable.js
gibson042: I would _love_ to disable that warning; continue is as useful as `break` and early `return` and I don't see any value in discouraging it.

gibson042:
> decodePassable with need these same changes, yes?

Both are in the same file and included in this PR (and in fact most of the changes affect decoding).

packages/marshal/src/encodePassable.js
erights: (FWIW I'm surprised I did that and have no idea why I did. Glad to see it fixed, thanks!)

erights:
> > decodePassable with need these same changes, yes?
> 
> Both are in the same file and included in this PR (and in fact most of the changes affect decoding).

Oops. I mean `compactOrdered`.

gibson042:
Ah, yes. I'll do that in the next round of updates to #1594 this week.

PR #1998, opened 2024-01-23
Pet Daemon: Merge Master (January, 2024)

kriskowal:
This just brings the `endo` branch back into sync with `master`.

PR #2003, opened 2024-01-25
fix(bundle-source): apply eval evasion for cjs modules too

kumavis:
encountered the incompatibility in either 1kce demo or cat ui in https://github.com/endojs/endo/pull/1877

this applies the evasive eval transform to cjs in addition to mjs as we do currently

### DIFF
--- a/packages/bundle-source/src/zip-base64.js
+++ b/packages/bundle-source/src/zip-base64.js
@@ -107,6 +107,32 @@ export async function bundleZipBase64(
     };
   }
 
+  /**
+   *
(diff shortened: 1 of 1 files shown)


kriskowal (changes requested):
This will not merge cleanly with when `master` gets merged into `endo` (see https://github.com/endojs/endo/pull/1998) because `evadeCensor` was factored out of `bundleSource` by @boneskull in https://github.com/endojs/endo/pull/1812

This is the right tactic, but it should be much cleaner if built on `master`.

PR #2004, opened 2024-01-25
feat(daemon): mail is delivered to both inboxes + includes formula identifiers

kumavis:
this powers the chat feature in https://github.com/endojs/endo/pull/1877
including the package message formula type-aware context menu

### DIFF
--- a/packages/daemon/src/mail.js
+++ b/packages/daemon/src/mail.js
@@ -100,21 +100,27 @@ export const makeMailboxMaker = ({
     const dubMessage = message => {
       const { type } = message;
       if (type === 'request') {
-        const { who: senderFormulaIdentifier, ...rest } = message;
+        const { who: senderFormulaIdentifier, dest: recipientFormulaIdentifier, ...rest } = message;
         const [senderName] = reverseLookupFormulaIdentifier(
           senderFormulaIdentifier,
         );
+        const [recipientName] = reverseLookupFormulaIdentifier(
+          recipientFormulaIdentifier,
+        );
         if (senderName !== undefined) {
-          return { who: senderName, ...rest };
+          return { who: senderName, dest: recipientName, ...rest };
         }
         return undefined;
       } else if (type === 'package') {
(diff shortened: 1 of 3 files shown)


kriskowal (commented):
This is progress. It could use a test in the pattern of `test/test-endo.js`.

The `who`, `what`, `when` convention was cute but it’s no longer going to cut it. Can I bother you to change that to `to` and `from` or `sender` and `receiver`? I’d welcome changing `when` to `date`.

kumavis:
Can add a test.

I'm considering refactoring messages to cover both requests and packages.
I'd prefer to revisit the property rename then

packages/daemon/src/mail.js
```diff
@@ -421,29 +454,35 @@ export const makeMailboxMaker = ({
...
+      // Send to own inbox.
+      const selfResponseP = respond(
         what,
         responseName,
         selfFormulaIdentifier,
         petStore,
+        recipientFormulaIdentifier,
       );
-      responses.set(responseName, newResponseP);
+      const newResponseP = Promise.race([recipientResponseP, selfResponseP]);
```
kriskowal, line 480: I’m convinced of this logic. I think we should make a note that in the future, we should probably project the message back to the sender’s inbox with a different form, since the UI and the powers conferred should be different for the sender than the receiver. The sender should be able to cancel the request. Maybe a case can be made that they should be able to preempt the resolution, but I’m not sure.

kriskowal (approved):

kumavis:
we'll likely need to significantly redo mail to support multiplayer right? im unclear about that part

kriskowal:
> we'll likely need to significantly redo mail to support multiplayer right? im unclear about that part

There’ll be changes. I’m not sure how extensive yet. I do know that we’ll be putting “undubbed” messages on the wire (edge names and formula identifiers, neither the sender nor the receiver’s pet names), and the protocol object exposed to the wire will be some kind of dropbox for which the sender and receiver are either implied or enforced. The internal mailbox API might be the same or similar.

PR #2007, opened 2024-01-25
fix(bundle-source): apply eval evasion for cjs modules too

kumavis:
encountered the incompatibility in either 1kce demo or cat ui in https://github.com/endojs/endo/pull/1877

this applies the evasive eval transform to cjs in addition to mjs as we do currently

replaces https://github.com/endojs/endo/pull/2003

### DIFF
--- a/packages/bundle-source/src/zip-base64.js
+++ b/packages/bundle-source/src/zip-base64.js
@@ -108,6 +108,35 @@ export async function bundleZipBase64(
     };
   }
 
+  /**
+   * @param {import('@endo/compartment-mapper/src/types.js').Language} parser
+   * @param {Uint8Array} sourceBytes
+   * @param {string} specifier
+   * @param {string} location
+   * @param {string|import('source-map').RawSourceMap} sourceMap
+   */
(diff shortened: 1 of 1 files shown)


kriskowal (approved):
Thank you. No notes.

packages/bundle-source/src/zip-base64.js
```diff
@@ -108,6 +108,35 @@ export async function bundleZipBase64(
...
+    parser,
+    sourceBytes,
+    specifier,
+    location,
+    sourceMap,
+  ) => {
+    if (!['mjs', 'cjs'].includes(parser)) {
+      throw Error(`Parser ${parser} not supported in evadeEvalCensor`);
+    }
+    const babelSourceType = parser === 'mjs' ? 'module' : 'script';
```
kriskowal, line 128: It occurs to me that you could pass `'module'` and `'script'` without this translation.

kumavis: we need to indicate both parser lang and `babelSourceType`, so i decided to abstract over babel

PR #2008, opened 2024-01-25
fix(marshal)!: compare strings by codepoint

erights:
closes: #2113 
refs: #2002  https://github.com/Agoric/agoric-sdk/pull/10299 https://github.com/Agoric/agoric-sdk/pull/10165/files#diff-b51392c299617fe43392dd0bcf11b9ed89b619cd4e5051fcffc9bfedef1c4d15 https://github.com/endojs/endo/pull/2873

## Description

- JavaScript's relational comparison operations as exposed in methods like `Array.prototype.sort` and operators `<`/`<=`/`>=`/`>` compare strings by lexicographic UTF-16 code unit order, which is exposes an internal representational detail not relevant to the string's meaning as a Unicode string. Previously, `compareRank` and associated functions compared strings using this JavaScript-native comparison. Now `compareRank` and associated functions compare strings by lexicographic Unicode _code point_ order. ***This change only affects comparison of characters in the range U+E000 through U+FFFF vs. supplementary-plane characters starting at U+10000 [i.e., those whose code point does not fit in 16 bits], and therefore only collections of strings including characters from both of those ranges***.
  - This release does not change the `encodePassable` encoding. But now, when we say it is order preserving, we need to be careful about which order we mean. `encodePassable` is rank-order preserving when the encoded strings are compared using `compareRank`.
  - The key order of strings defined by the @endo/patterns module is still defined to be the same as the rank ordering of those strings. So this release changes key order among strings to also be lexicographic comparison of Unicode code points. To accommodate this change, you may need to adapt applications that relied on key-order being the same as JS native order. This could include the use of any patterns expressing key inequality tests, like `M.gte(string)`.

### Security Considerations

The fact that the string ordering is closer to the Unicode semantics of the strings probably minimizes some surprises in ways that help security. OTOH, this difference from JS native string ordering probably causes other surprises that hurt security. Altogether, we do not expect much effect.

### Scaling Considerations

As a comparison written in JS, will be slower that the JS native string comparison. On XS at least, we expect to have a native code point comparison function available eventually. Altogether, we do not expect much effect.

Most developers will not care. But it needs to be explained somewhere carefully so that developers that do care can easily find out.

### Testing Considerations

@gibson042 , in a later PR, could you expand the property-based-testing to generate test cases sensitive to this change?

  - These string ordering changes brings Endo into conformance with any string ordering components of the OCapN standard.
  - To accommodate these change, you may need to adapt applications that relied on rank-order or key-order being the same as JS native order. You may need to resort any data that had previously been rank sorted using the prior `compareRank` function. You may need to revisit any use of patterns like `M.gte(string)` expressing inequalities over strings.

### Upgrade Considerations

If we currently have any persistent data, especially on chain, sorted according to JS native order (by UTF-16 code unit), then we cannot accept this PR until we have a plan to resort that data, or somehow continue to live with mis-sorted. (Historical note: This is how Oracle came to permanently rely on UTF-16 code unit order, because of the impracticality of resorting all that data.)

### DIFF
--- a/packages/marshal/NEWS.md
+++ b/packages/marshal/NEWS.md
@@ -1,5 +1,13 @@
 User-visible changes in `@endo/marshal`:
 
+# Next release
+
+- JavaScript's relational comparison operators like `<` compare strings by lexicographic UTF16 code unit order, which exposes an internal representational detail not relevant to the string's meaning as a Unicode string. Previously, `compareRank` and associated functions compared strings using this JavaScript-native comparison. Now `compareRank` and associated functions compare strings by lexicographic Unicode Code Point order. ***This change only affects strings containing so-called supplementary characters, i.e., those whose Unicode character code does not fit in 16 bits***.
+  - This release does not change the `encodePassable` encoding. But now, when we say it is order preserving, we need to be careful about which order we mean. `encodePassable` is rank-order preserving when the encoded strings are compared using `compareRank`.
+  - The key order of strings defined by the @endo/patterns module is still defined to be the same as the rank ordering of those strings. So this release changes key order among strings to also be lexicographic comparison of Unicode Code Points. To accommodate this change, you may need to adapt applications that relied on key-order being the same as JS native order. This could include the use of any patterns expressing key inequality tests, like `M.gte(string)`.
+  - These string ordering changes brings Endo into conformance with any string ordering components of the OCapN standard.
+  - To accommodate these change, you may need to adapt applications that relied on rank-order or key-order being the same as JS native order. You may need to resort any data that had previously been rank sorted using the prior `compareRank` function. You may need to revisit any use of patterns like `M.gte(string)` expressing inequalities over strings.
+
 # v1.7.0 (2025-06-02)
 
 - `@endo/marshal` now also exports a `qp` function meaning "quote passable"
--- a/packages/marshal/src/rankOrder.js
+++ b/packages/marshal/src/rankOrder.js
@@ -53,6 +53,38 @@ const sameValueZero = (x, y) => x === y || is(x, y);
 const trivialComparator = (left, right) =>
   // eslint-disable-next-line no-nested-ternary, @endo/restrict-comparison-operands
   left < right ? -1 : left === right ? 0 : 1;
+harden(trivialComparator);
+
+// Apparently eslint confused about whether the function can ever exit
+// without an explicit return.
+// eslint-disable-next-line jsdoc/require-returns-check
+/**
+ * @param {string} left
+ * @param {string} right
+ * @returns {RankComparison}
+ */
+export const compareByCodePoints = (left, right) => {
+  const leftIter = left[Symbol.iterator]();
+  const rightIter = right[Symbol.iterator]();
+  for (;;) {
+    const { value: leftChar } = leftIter.next();
+    const { value: rightChar } = rightIter.next();
+    if (leftChar === undefined && rightChar === undefined) {
+      return 0;
+    } else if (leftChar === undefined) {
+      // left is a prefix of right.
+      return -1;
+    } else if (rightChar === undefined) {
+      // right is a prefix of left.
+      return 1;
+    }
+    const leftCodepoint = /** @type {number} */ (leftChar.codePointAt(0));
+    const rightCodepoint = /** @type {number} */ (rightChar.codePointAt(0));
+    if (leftCodepoint < rightCodepoint) return -1;
+    if (leftCodepoint > rightCodepoint) return 1;
+  }
+};
+harden(compareByCodePoints);
 
 /**
  * Compare two same-type numeric values, returning results consistent with
@@ -170,8 +202,7 @@ export const makeComparatorKit = (compareRemotables = (_x, _y) => NaN) => {
         return 0;
       }
       case 'boolean':
-      case 'bigint':
-      case 'string': {
+      case 'bigint': {
         // Within each of these passStyles, the rank ordering agrees with
         // JavaScript's relational operators `<` and `>`.
         if (left < right) {
@@ -180,6 +211,9 @@ export const makeComparatorKit = (compareRemotables = (_x, _y) => NaN) => {
           assert(left > right);
           return 1;
         }
+      }
+      case 'string': {
+        return compareByCodePoints(left, right);
       }
       case 'symbol': {
         return comparator(
--- a/packages/marshal/index.js
+++ b/packages/marshal/index.js
@@ -16,6 +16,7 @@ export {
 
 export {
   compareNumerics,
+  compareByCodePoints,
   assertRankSorted,
   compareRank,
   isRankSorted,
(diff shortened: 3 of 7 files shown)


kriskowal (commented):
Excellent.

For this change, I do not think we can avoid the breaking change marker. That might render my argument for leaving it out of `pass-style`, moot.

erights:
> Excellent.
> 
> For this change, I do not think we can avoid the breaking change marker. That might render my argument for leaving it out of `pass-style`, moot.

Let me be sure I understand:

You're saying that this PR should keep the "!". Given that, we may as well keep the "!" on #2002 as well. Right?

kriskowal:
Yes

erights:
Just noting here for curiosity. In the UTF16 portion of https://icu-project.org/docs/papers/utf16_code_point_order.html

> This opens the door for a "fix-up" of code unit values that is faster than assembling 21-bit code point values. 

OMG

packages/marshal/NEWS.md
```diff
@@ -1,5 +1,13 @@
 User-visible changes in `@endo/marshal`:
 
+# next release
+
+- JavaScript's relational comparison operators like `<` compare strings by lexicographic UTF16 code unit order, which is exposes an internal representational detail not relevant to the string's meaning as a Unicode string. Previously, `compareRank` and associated functions compared strings using this JavaScript-native comparison. Now `compareRank` and associated functions compare strings by lexicographic Unicode Code Point order. ***This change only affects strings containing so-called supplementary characters, i.e., those whose Unicode character code does not fit in 16 bits***.
+  - This release does not change the `encodePassable` encoding. But now, when we say it is order preserving, we need to be careful about which order we mean. `encodePassable` is rank-order preserving when the encoded strings are compared using `compareRank`.
```
erights, line 6: @gibson042 is this true? It was true for my small test case, which proves very little. Will the same property also be true for `compactOrdered`? For either, does restricting these strings to well-ordered have any effect on whether their encoding is order preserving?

packages/marshal/src/rankOrder.js
```diff
@@ -46,9 +46,46 @@ const { entries, fromEntries, setPrototypeOf, is } = Object;
  */
 const sameValueZero = (x, y) => x === y || is(x, y);
 
+/**
+ * @param {any} left
+ * @param {any} right
+ * @returns {RankComparison}
+ */
 export const trivialComparator = (left, right) =>
```
kriskowal, line 54: This could be simply called `compare` since it is the default JavaScript comparator made flesh.

packages/marshal/test/test-encodePassable.js
```diff
@@ -67,7 +67,7 @@ const encodePassableInternal = makeEncodePassable({
   encodeError: er => encodeThing('!', er),
 });
 
-const encodePassable = passable => {
+export const encodePassable = passable => {
```
kriskowal, line 70: I do not feel strongly about this, but I think a utility for multiple tests should be ejected to a non `test-` prefix module. I suspect that Ava might do something weird if a `test-` imports another `test-`, or does something extraordinary to avoid doing something weird.

kriskowal (approved):
Thanks for the detailed `NEWS.md`.

packages/marshal/src/rankOrder.js
```diff
@@ -46,9 +46,46 @@ const { entries, fromEntries, setPrototypeOf, is } = Object;
...
   // eslint-disable-next-line no-nested-ternary, @endo/restrict-comparison-operands
   left < right ? -1 : left === right ? 0 : 1;
```
gibson042, line 56: Now that this is no longer meant for strings, we should probably rename it to something like `numericComparator` (possibly with an interim deprecation period) and tighten up the JSDoc types.

packages/marshal/test/test-encodePassable.js
dckc: yes, if a imports b, then `ava a` will run all of b too.

> modules should not export powerful objects
> Note that tests don't export anything.
-- https://github.com/Agoric/agoric-sdk/wiki/OCap-Discipline

packages/marshal/NEWS.md
erights: Good point, and bad news!

Grepping for `.sort()` specifically with nothing between the parens, I see 96 occurrences in agoric-sdk and 26 in endo. Some may not be or contain strings. But still, fixing all that do will be disruptive. And the longer we wait, the more disruptive it'll be.

I'm putting this back into Draft until we decide what our plan is. Attn @ivanlei

packages/marshal/src/rankOrder.js
erights: Already exported from package before this PR, so I'd rather not introduce a compat problem for a minor improvement to this name. That said, were it not for the (also minor) compat pain, I do like `compare` better.

erights: Resolving.

packages/marshal/test/test-encodePassable.js
erights: Good suggestion. Nice to have the common stuff factored out. Done.

packages/marshal/NEWS.md
```diff
@@ -1,5 +1,13 @@
 User-visible changes in `@endo/marshal`:
 
+# next release
+
+- JavaScript's relational comparison operators like `<` compare strings by lexicographic UTF16 code unit order, which is exposes an internal representational detail not relevant to the string's meaning as a Unicode string. Previously, `compareRank` and associated functions compared strings using this JavaScript-native comparison. Now `compareRank` and associated functions compare strings by lexicographic Unicode Code Point order. ***This change only affects strings containing so-called supplementary characters, i.e., those whose Unicode character code does not fit in 16 bits***.
```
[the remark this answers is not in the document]
erights: Done

packages/marshal/src/rankOrder.js
erights: Again, since it is already exported from the package, I'm electing not make this minor name improvement. The old name is good enough. Resolving.

packages/marshal/NEWS.md
erights: Is there any practical way to scan a recent snapshot of our chain and ***somehow*** see how many persistent strings are
- non-ascii,
- non-well-formed, or
- have supplementary characters (those whose code is > 16 bits)

? 
How hard would it be?
Attn @mhofman 

NOT URGENT.

erights:
Ideally, I would like to assign @gibson042 and @mhofman, to unassign myself, and then to list myself as a reviewer. However, for some #^*)$ reason github won't allow the creator to be a reviewer even after they are no longer assigned. So I'm staying on as assignee just as a proxy for being a reviewer.

@gibson042 and @mhofman , all yours! Thanks!

packages/marshal/src/rankOrder.js
```diff
@@ -181,6 +212,9 @@ export const makeComparatorKit = (compareRemotables = (_x, _y) => NaN) => {
           return 1;
         }
       }
+      case 'string': {
+        return compareByCodePoints(left, right);
+      }
```
erights, line 217: Switch between old and new behavior based on an environment var

erights:
Closing in favor of #2875

PR #2010, opened 2024-01-25
refactor(daemon,cli): Rename capitalized names and arguments

rekmarks:
- Uncapitalizes all "special names" in the daemon, since we have determined we no longer need this.
- Renames the `--UNSAFE` CLI flag to `--UNCONFINED`, which is more specific. We maintain capitalization here to draw attention to it.

### DIFF
--- a/packages/cli/demo/cat.js
+++ b/packages/cli/demo/cat.js
@@ -4,7 +4,7 @@
 // This command will set up the cat page, create a URL,
 // and open it.
 //
-// > endo open familiar-chat cat.js --powers HOST
+// > endo open familiar-chat cat.js --powers host
 //
 // Thereafter,
 //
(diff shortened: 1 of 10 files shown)


kriskowal (changes requested):
I’ve miscommunicated again!

We do actually need special names to be capitalized to distinguish them from pet names in the pet store. For example, the name `HOST` is not deletable whereas `host` can be assigned.

There is no need for the third term in `INFO.host.worker` to be capitalized under a pet inspector.

kriskowal:
(Renaming `--UNSAFE` to `--UNCONFINED` is good and @dckc will thank you!)

rekmarks:
Alas dear PR, I barely knew ye...

kriskowal:
I should also clarify that the design requirement is that special names and pet names need to be distinguishable. This is just the convention of the day and can be changed. We want pet names to have a much more broad range. The distinguisher could be something else. A dot prefix (`.self`, `guest..self`)?

PR #2011, opened 2024-01-26
feat(cli)!: Rename --UNSAFE to --UNCONFINED

rekmarks:
Renames the `UNSAFE` flag to `UNCONFINED`, which is more specific. Also renames internals using "unsafe" (like `import-unsafe`) to use "unconfined" instead.

Will require an `endo purge -f`.

### DIFF
--- a/packages/daemon/src/types.d.ts
+++ b/packages/daemon/src/types.d.ts
@@ -73,7 +73,7 @@ type EvalFormula = {
 };
 
 type ImportUnsafeFormula = {
-  type: 'import-unsafe';
+  type: 'make-unconfined';
   worker: string;
   powers: string;
   importPath: string;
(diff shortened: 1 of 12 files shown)


kriskowal (approved):
Cool, to cut all the way down to the bone, I would like to rename `importUnsafeAndEndow` to `makeUnconfined` and `importBundleAndEndow` to merely `makeBundle` throughout the `daemon.js`. I would also like to rename the formula types, `import-unsafe` and `import-bundle` to merely `make-unconfined` and `make-bundle`. These changes will necessitate an `endo purge --force`, but c’est la mode.

packages/daemon/src/types.d.ts
```diff
@@ -73,15 +73,15 @@ type EvalFormula = {
 };
 
 type ImportUnsafeFormula = {
```
kriskowal, line 75: Could hit the type name too.

kriskowal (approved):

dckc:
an lo! there was great rejoicing!

thanks!

PR #2013, opened 2024-01-26
refactor(errors): support our assert conventions

erights:
closes: #XXXX
refs: https://github.com/Agoric/agoric-sdk/pull/8781#pullrequestreview-1835371445

## Description

At https://github.com/Agoric/agoric-sdk/pull/8781#pullrequestreview-1835371445 @turadg reacts to

```js
import { quote as q, throwRedacted as Fail } from '@endo/errors';
```
with

> This pattern is what I was trying to get away from with making a package at all. There's no export of `Fail` so there's no autocomplete when someone wants to import `Fail`.

IMO quite right. At https://github.com/Agoric/agoric-sdk/pull/8781 I tried following the agreement from which the current @endo/errors API was born, where the exported explanatory names would be renamed to the conventionally used names on import. Even though I did not notice the autocomplete problem myself, I found https://github.com/Agoric/agoric-sdk/pull/8781 unpleasant enough that I consider the current @endo/errors API to be a failed experiment.

This PR in its current state splits the difference. It does not withdraw the explanatory names. It simply exports aliases to their conventional names. ***Reviewers: Should I instead delete the explanatory names, since we don't intend to use them anyway?*** If so, it is ok if that deletion happens in a separate PR?

I needed to make two additional changes to switch things to @endo/errors. Unlike the `assert` global, the `assert` exported by @endo/errors does not include `error` (aka `makeError`) and `note`, instead making them separate exports of the module. This is good. But is requires changing the affected occurrences of `assert.error` to `makeError` and `assert.note` to something. For the latter, I first tried just using the name @endo/errors already exports, `note`, but found it too mysterious by itself. So I added an alias to `errorNote` and universally used that instead.

Finally, I had to omit completely those packages that were too low level to get `assert` by importing it from @endo/assert. I left them continuing to use the `assert` global.

### Security Considerations
none
### Scaling Considerations
none

We should document the new @endo/assert convention as we use it, to encourage others to adopt the same conventions.

### Testing Considerations
none, I think
If this PR continues to not withdraw the explanatory names, then none. Otherwise we'd need to fix existing uses, of which there appears to be none.
### Upgrade Considerations

Without withdrawing the explanatory names, there should be no breaking change. I still need to write a NEWS.md entry and check that box. But I'll wait until I get feedback on this PR before writing that.

### DIFF
--- a/packages/captp/src/atomics.js
+++ b/packages/captp/src/atomics.js
@@ -1,6 +1,6 @@
 /// <reference types="ses"/>
 
-const { details: X, Fail } = assert;
+import { X, Fail } from '@endo/errors';
 
 // This is a pathological minimum, but exercised by the unit test.
 export const MIN_DATA_BUFFER_LENGTH = 1;
@@ -154,7 +154,7 @@ export const makeAtomicsTrapGuest = transferBuffer => {
     //
     // TODO: It would be nice to use an error type, but captp is just too
     // noisy with spurious "Temporary logging of sent error" messages.
-    // it.throw(assert.error(X`Trap host has not finished`));
+    // it.throw(makeError(X`Trap host has not finished`));
     it.throw(null);
 
     // eslint-disable-next-line no-bitwise
--- a/packages/captp/src/captp.js
+++ b/packages/captp/src/captp.js
@@ -12,7 +12,7 @@ import { Remotable, Far, makeMarshal, QCLASS } from '@endo/marshal';
 import { E, HandledPromise } from '@endo/eventual-send';
 import { isPromise, makePromiseKit } from '@endo/promise-kit';
 
-import { X, Fail, errorNote } from '@endo/errors';
+import { X, Fail, annotateError } from '@endo/errors';
 import { makeTrap } from './trap.js';
 
 import { makeFinalizingMap } from './finalize.js';
@@ -669,7 +669,7 @@ export const makeCapTP = (
           if (!e) {
             Fail`trapGuest expected trapHost AsyncIterator(${questionID}) to be done, but it wasn't`;
           }
-          errorNote(e, X`trapHost AsyncIterator(${questionID}) threw`);
+          annotateError(e, X`trapHost AsyncIterator(${questionID}) threw`);
           throw e;
         }
       };
--- a/packages/captp/src/captp.js
+++ b/packages/captp/src/captp.js
@@ -12,13 +12,12 @@ import { Remotable, Far, makeMarshal, QCLASS } from '@endo/marshal';
 import { E, HandledPromise } from '@endo/eventual-send';
 import { isPromise, makePromiseKit } from '@endo/promise-kit';
 
+import { X, Fail, errorNote } from '@endo/errors';
 import { makeTrap } from './trap.js';
 
 import { makeFinalizingMap } from './finalize.js';
 
 export { E };
-
-const { details: X, Fail } = assert;
 
 const WELL_KNOWN_SLOT_PROPERTIES = harden(['answerID', 'questionID', 'target']);
 
@@ -670,7 +669,7 @@ export const makeCapTP = (
           if (!e) {
             Fail`trapGuest expected trapHost AsyncIterator(${questionID}) to be done, but it wasn't`;
           }
-          assert.note(e, X`trapHost AsyncIterator(${questionID}) threw`);
+          errorNote(e, X`trapHost AsyncIterator(${questionID}) threw`);
           throw e;
         }
       };
--- a/packages/errors/index.js
+++ b/packages/errors/index.js
@@ -76,5 +76,5 @@ export {
   quote as q,
   redacted as X,
   throwRedacted as Fail,
-  note as errorNote,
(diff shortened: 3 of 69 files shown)


packages/errors/index.js
```diff
@@ -71,4 +71,10 @@ export {
   quote,
   redacted,
   throwRedacted,
+  // conventional abbreviations and aliases
+  bare as b,
+  quote as q,
+  redacted as X,
+  throwRedacted as Fail,
+  note as errorNote,
```
erights, line 79: This is the substantive and possibly controversial change.

packages/eventual-send/src/E.js
```diff
@@ -52,8 +52,8 @@ const makeEProxyHandler = (recipient, HandledPromise) =>
             if (this !== receiver) {
               // Reject the async function call
               return HandledPromise.reject(
-                assert.error(
-                  X`Unexpected receiver for "${propertyKey}" method of E(${q(
+                makeError(
+                  X`Unexpected receiver for "${q(propertyKey)}" method of E(${q(
```
erights, line 56: @michaelfig 
This is a drive by. IMO the property key should not be redacted.

packages/eventual-send/src/track-turns.js
```diff
@@ -1,11 +1,12 @@
 /* global globalThis */
+import { X, errorNote } from '@endo/errors';
 import {
   getEnvironmentOption,
   environmentOptionsListHas,
 } from '@endo/env-options';
 
 // NOTE: We can't import these because they're not in scope before lockdown.
-// import { assert, details as X, Fail } from '@agoric/assert';
+// import { assert, X, Fail } from '@agoric/assert';
```
erights, line 9: Either this is still true, in which case we should back out this change, or it is false, in which case we should remove this comment. If we pass CI with this change, that is weak evidence that this statement is false.

erights: CI did fail with
```
Error: Cannot initialize @endo/errors, missing globalThis.assert, import 'ses' before '@endo/errors'
```
which clearly indicates that this comment is still true. Will back out changes to the eventual send package (except the drive-by)

erights: Done. Resolving.

packages/captp/src/captp.js
```diff
@@ -670,7 +669,7 @@ export const makeCapTP = (
           if (!e) {
             Fail`trapGuest expected trapHost AsyncIterator(${questionID}) to be done, but it wasn't`;
           }
-          assert.note(e, X`trapHost AsyncIterator(${questionID}) threw`);
+          errorNote(e, X`trapHost AsyncIterator(${questionID}) threw`);
```
kriskowal, line 672: A thought: `annotateError` would read more descriptively.

kriskowal (commented):
I dislike this change and recognize that my hard line (that a reviewer should not have to look beyond the file they’re reading to expand an abbreviation to a name that describes the function’s behavior) is mutually exclusive with @turadg’s hard line (that the name used at the call site in our style should allow the IDE to automatically draw forth the correct import statement.) Using descriptive names at the call sites would break that log jam but is mutually exclusive with @erights preference for brevity (and mine, for `q` and `b` in particular).

So, I would appeal to arbitration (by any member of the team) and will abide whatever choice is favored by the arbiter.

turadg:
> mutually exclusive with @turadg’s hard line (that the name used at the call site in our style should allow the IDE to automatically draw forth the correct import statement.)

I don't have a position on how it's factored in Endo. My position is just for consumer of `@agoric` packages.

I'm fine with leaving `@endo/errors` as it is and putting these convenience exports only into `@agoric/assert`.

kriskowal:
> > mutually exclusive with @turadg’s hard line (that the name used at the call site in our style should allow the IDE to automatically draw forth the correct import statement.)
> 
> I don't have a position on how it's factored in Endo. My position is just for consumer of `@agoric` packages.
> 
> I'm fine with leaving `@endo/errors` as it is and putting these convenience exports only into `@agoric/assert`.

I think it would be a pity to have an adapter package in `@agoric` to settle a disagreement over policy about concerns that apply equally to both `@endo` and `@agoric` packages.

kriskowal (approved):
Out of band, @erights has made the value judgement that abbreviating the exports and using consistent abbreviations between Endo and Agoric is the better value for cost than making those conventional abbreviations explicit in every module that uses them. I respect @turadg’s preference and this choice.

I would like to continue exporting the descriptive names that these abbreviations alias.

packages/captp/src/atomics.js
```diff
@@ -1,6 +1,6 @@
 /// <reference types="ses"/>
 
-const { details: X, Fail } = assert;
+import { X, Fail } from '@endo/errors';
```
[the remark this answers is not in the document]
erights: In that file, only occurrence was within a comment. No `assert.error` occurrences at all.

packages/captp/src/captp.js
erights: Done

PR #2015, opened 2024-01-27
feat(cli): Add install --open flag

kriskowal:
Buzz a sharp corner off of the `install` and `open` workflow by allowing them to be combined into a single command.

### DIFF
--- a/packages/cli/demo/README.md
+++ b/packages/cli/demo/README.md
@@ -258,11 +258,16 @@ _Familiar Chat_ is an example application that provides a web app for
 interacting with your pet daemon.
 
 ```
-> endo open familiar-chat cat.js --powers HOST
+> endo install familiar-chat cat.js --powers SELF
 ```
 
 This command creates a web page named familiar-chat and endows it with the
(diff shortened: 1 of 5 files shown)


packages/cli/demo/README.md
```diff
@@ -258,11 +258,16 @@ _Familiar Chat_ is an example application that provides a web app for
 interacting with your pet daemon.
 
 ```
-> endo open familiar-chat cat.js --powers HOST
+> endo install familiar-chat cat.js --powers HOST
```
kriskowal, line 261: I could see an argument that this might be a more predictable usage if shaped more like other commands: `endo install cat.js --powers HOST --name familiar-chat`, but the `--name` flag is not optional for `install`. If that’s preferable to others, I’d consider doing that in a follow-up change.

rekmarks (changes requested):
Docs should be updated to say `SELF` instead of `HOST` where appropriate (e.g. for `--powers`).

rekmarks (approved):
LGTM!

PR #2016, opened 2024-01-27
fix(daemon): Address off-by-nybble formula path error

kriskowal:
When splitting a hash into a single byte prefix and suffix, I accidentally trimmed four bits from the head of the suffix. Nothing broke because the mistake was consistent over a consistent hash but it is technically incorrect.


rekmarks (approved):
LGTM.

Reminder for readers to `endo purge -f` after pulling this.

PR #2020, opened 2024-01-30
chore: Specify node engine compatibility range

### LINKED ISSUE #1046: Drop Node.js 12 support in CI

kriskowal:
Closes #1046

The engines field is advisory, but this should ensure that folks get a warning if they’re using a version of Node.js for which SES and its libraries were not vetted. SES at least does not guarantee confinement for future versions of JavaScript, particularly in the unlikely but possible case syntax reveals more undeniable intrinsics.

This change was generated mechanically:

```bash
ls packages/*/package.json | while read -r file; do
  jq '. + { engines: {node: "^16 || ^18 || ^20" }}' $file | pbcopy
  pbpaste > $file
done
```


kriskowal:
I’ve removed the Node.js 12 platform compatibility test and removed it from required checks in Github CI. We can now rest easy that `esm` is gone (gone enough) and `globalThis` is broadly supported.

kriskowal:
I’m closing this since the value is low and the risk of interfering with a working system is high. Also easy to revisit.

PR #2022, opened 2024-01-31
feat(daemon): Add petname path lookup

rekmarks:
Supersedes: #1915 

## Description

Implements mailbox petname path lookup by making the mailbox `lookup()` method variadic. Adds tests to verify how this composes with the inspector and caplets with their own `lookup()` method.

This enables us to later add dot-separated paths for petnames in the CLI.

## Note to reviewers

Did I overlook any places that should accept `string | string[]` where currently only a single `petName` is expected?

### DIFF
--- a/packages/daemon/src/mail.js
+++ b/packages/daemon/src/mail.js
@@ -1,5 +1,6 @@
 // @ts-check
 
+import { E } from '@endo/far';
 import { makePromiseKit } from '@endo/promise-kit';
 import { makeChangeTopic } from './pubsub.js';
 import { makeIteratorRef } from './reader-ref.js';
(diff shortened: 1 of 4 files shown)


packages/daemon/test/lookup.js
```diff
@@ -0,0 +1,9 @@
+import { Far } from '@endo/far';
+
+export const make = () => {
+  return Far('Lookup', {
```
kriskowal, line 4: Shirley, there’s a noun for that verb.

```suggestion
  return Far('Well', {
```

packages/daemon/src/mail.js
```diff
@@ -41,16 +42,20 @@ export const makeMailboxMaker = ({
     };
 
     /**
-     * @param {string} petName
+     * @param {string[]} petNamePath - A sequence of pet names.
```
kriskowal, line 45: Does the tsdoc need an ellipses somewhere?

kriskowal (approved):
Thank you 🙏

packages/daemon/test/lookup.js
rekmarks: Well, according to my sources it's also "lookup". And stop calling me Shirley!

packages/daemon/src/mail.js
rekmarks: TIL you do the ellipses in the JSDoc type annotations. (I also forgot to add them to `types.d.ts`)

Issue #2024, opened 2024-01-31
@endo/daemon: Add `makeBundle()` test 


rekmarks:
The `daemon` unit tests currently all use `makeUnconfined` where caplets are used. We should add a test using `makeBundle`, because confinement should be tested.

PR #2027, opened 2024-02-01
test(daemon): Add `makeBundle` test

rekmarks:
Closes: #2024 

Tests confined bundles using `makeBundle`. Adds a utility to make using `makeBundle` significantly easier in tests for the daemon.

### DIFF
--- a/packages/daemon/test/test-endo.js
+++ b/packages/daemon/test/test-endo.js
@@ -11,6 +11,7 @@ import url from 'url';
 import path from 'path';
 import { E } from '@endo/far';
 import { makePromiseKit } from '@endo/promise-kit';
+import bundleSource from '@endo/bundle-source';
 import {
   start,
   stop,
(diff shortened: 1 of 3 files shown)


packages/daemon/test/test-endo.js
```diff
@@ -41,6 +42,36 @@ const makeLocator = (...root) => {
   };
 };
 
+// The id of the next bundle to make.
+let bundleId = 0;
+const textEncoder = new TextEncoder();
+
+// TODO: We should be able to use {import('../src/types').EndoHost} for `host`,
+// but when laundered through `E()` it becomes `never`.
```
rekmarks, line 50: @kriskowal @kumavis is this something you've encountered before? Is it user error?

kriskowal: No. The `.js` extension may be required (even though it is a `.d.ts`)

kriskowal (approved):
Very interested in obviating the need for temporary names. Maybe we can eventually use dotted paths into mailbox messages and log commands.

packages/daemon/test/test-endo.js
rekmarks: This does not appear to make a difference on my system. The error is visible in my editor (VSCode) using the repo version of TypeScript (`5.3.3`). ESLint does not appear to be configured to validate JSDoc types (although we could make it so).

PR #2030, opened 2024-02-02
feat(ses): enablements needed by (old?) mobx

### LINKED ISSUE #2029: Cosmos-kit incompatible with lockdown
## Describe the bug

When trying to use https://github.com/cosmology-tech/cosmos-kit in a React app with lockdown, it crashes with an error like:

```
TypeError: Cannot assign to read only property 'concat' of object '[object Object]'
```

## Steps to reproduce

1. Go to https://github.com/agoric-labs/cosmos-kit-ses?tab=readme-ov-file
2. Clone the repo
3. `yarn && yarn dev` to run locally
4. Reload the url outputted by `yarn dev`
5. Observe the app crashes to a blank page and the error appears in the console

## Expected behavior

There should be some way to use cosmos-kit alongside endo

## Platform environment

 - What OS are you using? MacOS 13.3.1 (22E261)
 - What version of Node.js? v18.17.1
 - Is there anything special/unusual about your platform? No
 - What version of Endo are you using? `"ses": "^1.1.0"`

## Additional context

I tried using `@endo/init/unsafe-fast.js` to circumvent the issue in this [branch](https://github.com/agoric-labs/cosmos-kit-ses/tree/endo-init-unsafe?tab=readme-ov-file) unsuccessfully. It throws a different error `TypeError#2: Cannot perform 'ownKeys' on a proxy that has been revoked` described in the README.

## Screenshots

Screenshot 2024-02-01 at 10 24 59 AM

erights:
closes: #XXXX
refs: #2029 

## Description

This does not close #2029 , but it does close the first hurdle, the 

```
TypeError: Cannot assign to read only property 'concat' of object '[object Object]'
```

problem. I have yet to diagnose or fix the 

```
TypeError#2: Cannot perform 'ownKeys' on a proxy that has been revoked
```

The problem fixed by this PR is interesting. The override mistake is in mobx generated code, where the corresponding mobx source code is a TypeScript class definition. The generated code is not a JS class definition, but rather, seems to target pre-class JS. The surprising mistake is that the generated code is initializing the methods on the "class" .prototype using assignment rather than `defineProperty` in blatant violation of the JS class semantics. Presumably, this is also a blatant violation of the current TS class semantics, but I don't know. Is some ancient TS compiler at play here? Or some non-standard TS or TS-like compiler? I don't know.

### Security Considerations

The errant TS class compilation is itself worrisome, both from general correctness and from security. But that issue is only revealed by this bug and PR, not affected by it. Nevertheless, we should try to understand what's going on, and to send upstream a mobx fix that repairs their class semantics. Since mobx is only a very indirect dependency of ours to which we do not impute any particular security properties, this is likely low urgency.

In the absence of this PR, as explained at #2029, much more violent inhibitions of SES's security were attempted (`unsafe-fast.js`) and might have been used in practice, resulting is a great loss of hardening and integrity protection. This PR makes that unnecessary, at least as far as this first #2029 problem which raised that possibility.

### Scaling Considerations
none
none
### Testing Considerations
none
none
### Upgrade Considerations
none. Nothing breaking. Nothing that warrants a NEWS.md update.

### DIFF
--- a/packages/ses/src/enablements.js
+++ b/packages/ses/src/enablements.js
@@ -1,4 +1,4 @@
-import { toStringTagSymbol } from './commons.js';
+import { toStringTagSymbol, iteratorSymbol } from './commons.js';
 
 /**
  * @file Exports {@code enablements}, a recursively defined
@@ -97,6 +97,8 @@ export const moderateEnablements = {
   '%ArrayPrototype%': {
     toString: true,
     push: true, // set by "Google Analytics"
+    concat: true, // set by mobx generated code (old TS compiler?)
+    [iteratorSymbol]: true, // set by mobx generated code (old TS compiler?)
   },
 
   // Function.prototype has no 'prototype' property to enable.
--- a/packages/ses/test/test-enable-property-overrides-default.js
+++ b/packages/ses/test/test-enable-property-overrides-default.js
@@ -8,7 +8,7 @@ test('enablePropertyOverrides - on', t => {
   overrideTester(t, 'Object', {}, ['toString', 'valueOf']);
   // We allow 'length' *not* because it is in enablements; it is not;
   // but because each array instance has its own.
-  overrideTester(t, 'Array', [], ['toString', 'length', 'push']);
+  overrideTester(t, 'Array', [], ['toString', 'length', 'push', 'concat']);
   // eslint-disable-next-line func-names, prefer-arrow-callback
   overrideTester(t, 'Function', function () {}, [
     'constructor',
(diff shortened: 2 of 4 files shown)


samsiegart (approved):
This seems to fix the issue in https://github.com/agoric-labs/cosmos-kit-ses, thanks!

I checkout out this branch locally and did:
```
yarn
cd packages/ses
yarn build
yarn link
```

Then in https://github.com/agoric-labs/cosmos-kit-ses I did:
```
yarn link "ses"
yarn build
yarn preview
```
And was able to run the app without changing any of the source code in `main` branch. The lockdown options it uses are:
```

lockdown({
  errorTaming: 'unsafe',
  overrideTaming: 'severe',
  consoleTaming,
});
```
...which seem to be necessary still.

For using `yarn dev`, vite seems finicky when doing `yarn link`, and I had to delete `node_modules` first so that it would use the linked version of `ses` instead of the published version that I previously installed.

Issue #2023, opened 2024-01-31
@endo/cli: Support dot-delimited petname paths


rekmarks:
Following #2022, we should support dot-delimited petname paths in the CLI such that wherever† a petname for an existing value is provided, a path can be provided instead. This path will then be passed to the mailbox `lookup` method, which will attempt to resolve the value.

For example, this should work:
```text
$ endo eval '10' --name ten
10
$ endo eval 'foo' foo:INFO.ten.source
Object [Alleged: Inspector (eval <id>)] {}
```

-----
† In all places? Maybe not `endo rename <from> <to>`.

kriskowal:
`endo rename` should be replaced with `endo mv|move` when we can support dot delimited paths, and we should support dot delimited paths for both the source and target. We should also guarantee that a move occurs in a single JavaScript event _unless_ the dot-delimited path reaches outside the local dæmon. That will require the `move` operation to walk from formula to formula for both the source and target path before falling through to the `lookup` protocol. We will need a corresponding protocol for “depositing” a formula identifier to the target “directory” and allow that protocol to work remotely.

`E(remoteDirectory).deposit(name, nonce, addresses)`. See https://github.com/ocapn/ocapn/pull/105

Issue #1921, opened 2023-12-31
macOS Sonoma leaves `*.localhost` handling at the mercy of ISPs


michaelfig:
## Describe the bug

refs #1006

_Reported by @SMotaal, and we investigated it together._

Using `*.endo.localhost` is not adequately supported under Safari for macOS 16.2.

Unlike Firefox and Chrome, which always use their own HTTPS DNS servers that support `*.localhost` in an RFC-compliant way, Safari uses the system default DNS resolver, which is probably backed by whichever Internet Service Provider is in use.

Not all such providers support `*.localhost`, which makes Safari unusable for the current Endo Daemon.

## Steps to reproduce

1. on macOS 16.2, `ping example.endo.localhost`
2. it may work, depending on your ISP, or it may give `ping: cannot resolve example.endo.localhost`
3. If the ping failed, try the 1kce demo, and attempt to open it in Safari.
4. Safari will display a failure as well.

## Expected behavior

There are workaround [public wildcard localhost DNS providers](https://gist.github.com/tinogomes/c425aa2a56d289f16a1f4fcb8a65ea65).  It would be useful to be able to specify a command line argument or environment variable to generate `*.endo${localhostSuffix}` URLs, where the default `localhostSuffix = '.localhost'`.

With that being supported, the Endo [daemon host match regexp](https://github.com/endojs/endo/blob/7c64fc28c4679f9bc83cba4a34f9ddffc6bd3df2/packages/daemon/src/daemon.js#L949) should be more tolerant of the specified URL suffices, something like:

```js
    const match =
          /^([0-9a-f]{32})\.endo(\.[^:/#?]+):([1-9][0-9]{0,4})$/.exec(host);
```
## Platform environment

 - What OS are you using? macOS 16.2
 - What version of Node.js? Node v20.6.1
 - Is there anything special/unusual about your platform? no
 - What version of Endo are you using? `kumavis-1kce-demo`

## Additional context

This works under Safari on macOS 15.

michaelfig:
I added more details to an [inconclusive WebKit bug that's been open since 2016](https://bugs.webkit.org/show_bug.cgi?id=160504#c16)

kriskowal:
What we know so far is that this `*.endo.localhost` strategy is unsuitable for confinement for a number of reasons including this issue. There are a number of alternatives, and we might employ some or all of them as they have varying pros and cons.

[...]

PR #2034, opened 2024-02-05
feat(cli,daemon): Support dot-delimited petname paths for CLI `eval` command

rekmarks:
Ref: #2023 

## Description

Adds support for dot-delimited petname paths to the CLI `eval` command. This required modifications to the `evaluate` method of the daemon's `host.js`, and the introduction of a new formula type, `lookup`.

The `lookup` formula type is necessary because the names in a lookup path may have neither petnames nor formula identifiers associated with them. Consider:

```text
> endo eval '10' --name ten
10
> endo eval 'foo' foo:INFO.ten.source
10
```

The only way retrieve the value for `source` is to call `E(HOST).lookup('INFO', 'ten', 'source')`, and since the `eval` formula expects its values to be mediated by formula identifiers, the lookup of `INFO.ten.source` must itself be stored as a formula.

The `lookup` formula is an `-id512` formula with the following interface:
```typescript
type LookupFormula = {
  type: 'lookup';

  /**
   * The formula identifier of the naming hub to call lookup on.
   * A "naming hub" is an object with a variadic `lookup()` method.
   */
  hub: string;

  /**
   * The pet name path.
   */
  path: Array<string>;
};
```

The `lookup` formula introduces the language of "naming hubs" into the daemon codebase. At present the only such objects are guests and hosts, which are also agents, but the category of naming hubs is expected to grow to include non-agent objects.

### DIFF
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -680,7 +680,7 @@ const makeEndoBootstrap = (
       if (
         ![
           'eval-id512',
-          'lookup',
+          'lookup-id512',
           'make-unconfined-id512',
           'make-bundle-id512',
           'guest-id512',
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -231,22 +231,22 @@ const makeEndoBootstrap = (
    * Creates a controller for a `lookup` formula. The external facet is the
    * resolved value of the lookup.
    *
-   * @param {string} agentFormulaIdentifier
+   * @param {string} hubFormulaIdentifier
    * @param {string[]} path
    * @param {import('./types.js').Terminator} terminator
    */
   const makeControllerForLookup = async (
-    agentFormulaIdentifier,
+    hubFormulaIdentifier,
     path,
     terminator,
   ) => {
-    terminator.thisDiesIfThatDies(agentFormulaIdentifier);
+    terminator.thisDiesIfThatDies(hubFormulaIdentifier);
 
     // Behold, recursion:
     // eslint-disable-next-line no-use-before-define
-    const agent = provideValueForFormulaIdentifier(agentFormulaIdentifier);
+    const hub = provideValueForFormulaIdentifier(hubFormulaIdentifier);
 
-    const external = E(agent).lookup(...path);
+    const external = E(hub).lookup(...path);
     return { external, internal: undefined };
   };
 
@@ -349,7 +349,7 @@ const makeEndoBootstrap = (
         terminator,
       );
     } else if (formula.type === 'lookup') {
-      return makeControllerForLookup(formula.agent, formula.path, terminator);
+      return makeControllerForLookup(formula.hub, formula.path, terminator);
     } else if (formula.type === 'make-unconfined') {
       return makeControllerForUnconfinedPlugin(
         formula.worker,
@@ -715,7 +715,7 @@ const makeEndoBootstrap = (
           formula.type,
           formulaNumber,
           harden({
-            agent: provideValueForFormulaIdentifier(formula.agent),
+            hub: provideValueForFormulaIdentifier(formula.hub),
             path: formula.path,
           }),
         );
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -228,6 +228,9 @@ const makeEndoBootstrap = (
   };
 
   /**
+   * Creates a controller for a `lookup` formula. The external facet is the
+   * resolved value of the lookup.
+   *
    * @param {string} agentFormulaIdentifier
    * @param {string[]} path
    * @param {import('./types.js').Terminator} terminator
@@ -517,7 +520,7 @@ const makeEndoBootstrap = (
    * @param {string} formulaType - The type of the formula.
    * @param {string} formulaNumber - The number of the formula.
    * @param {import('./types').Formula} formula - The formula.
-   * @returns {Promise<unknown>} The value of the formula.
+   * @returns {Promise<{ formulaIdentifier: string, value: unknown }>} The value of the formula.
    */
   const provideValueForNumberedFormula = async (
     formulaType,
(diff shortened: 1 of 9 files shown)


packages/daemon/src/daemon.js
```diff
@@ -227,6 +227,26 @@ const makeEndoBootstrap = (
...
+  const makeControllerForLookup = async (
+    agentFormulaIdentifier,
+    path,
+    terminator,
+  ) => {
+    terminator.thisDiesIfThatDies(agentFormulaIdentifier);
+
+    // Behold, recursion:
+    // eslint-disable-next-line no-use-before-define
+    const agent = provideValueForFormulaIdentifier(agentFormulaIdentifier);
```
kriskowal, line 244: Do we want `lookup` to work for all the formula types that contain a pet store? That’d include `host`, `guest`, and eventually `directory` if not others. We could insist that the formula identifier corresponds to the underlying `pet-store` (which would simplify the controller maker for watching the pet store for changes and propagating cancellation. That would put the responsibility for looking up the pet store higher up.

packages/daemon/src/daemon.js
```diff
@@ -676,6 +705,15 @@ const makeEndoBootstrap = (
             worker: provideValueForFormulaIdentifier(formula.worker),
           }),
         );
+      } else if (formula.type === 'lookup') {
+        return makeInspector(
+          formula.type,
+          formulaNumber,
+          harden({
+            agent: provideValueForFormulaIdentifier(formula.agent),
```
kriskowal, line 713: I think this should be `store`.

packages/daemon/src/pet-store.js
```diff
@@ -9,7 +9,7 @@ const { quote: q } = assert;
 
 const validIdPattern = /^[0-9a-f]{128}$/;
 const validFormulaPattern =
-  /^(?:host|pet-store|pet-inspector|(?:readable-blob-sha512|worker-id512|pet-store-id512|eval-id512|make-unconfined-id512|make-bundle-id512|host-id512|guest-id512):[0-9a-f]{128}|web-bundle:[0-9a-f]{32})$/;
+  /^(?:host|pet-store|pet-inspector|(?:readable-blob-sha512|worker-id512|pet-store-id512|eval-id512|make-unconfined-id512|make-bundle-id512|host-id512|guest-id512):[0-9a-f]{128}|(?:lookup|web-bundle):[0-9a-f]{32})$/;
```
kriskowal, line 12: In the fullness of time, I expect all of these to have the same CID breadth. `web-bundle` has the shorter CID because of the subdomain length restriction, but we’re going to move away from subdomains #1921. This should be an `id-512` like the rest. We will also, in the fullness of time, remove the `-id-512` suffix from all these and then make CIDs _something else entirely_ but at least homogeneous.

rekmarks: That makes sense. In that case, I'll just use all 64 bytes of the digest.

packages/daemon/src/host.js
```diff
@@ -203,24 +204,34 @@ export const makeHostMaker = ({
...
+            if (formulaIdentifier === undefined) {
+              throw new Error(`Unknown pet name ${q(petNamePath[0])}`);
+            }
+            return formulaIdentifier;
           }
-          return formulaIdentifier;
+
+          const { formulaIdentifier: lookupFormulaIdentifier } =
+            await provideLookupFormula(petNamePath);
+          return lookupFormulaIdentifier;
```
kriskowal, line 230: Would be good to factor out a `lookupFormulaIdentifierForPath` function. This won’t be the last place we use it.

packages/daemon/src/mail.js
```diff
@@ -58,11 +60,32 @@ export const makeMailboxMaker = ({
...
+      } else {
+        // eslint-disable-next-line no-use-before-define
+        const { value } = await provideLookupFormula(petNamePath);
+        // eslint-disable-next-line no-use-before-define
+        [formulaIdentifier] = await reverseLookup(value);
+        if (formulaIdentifier === undefined) {
+          throw new TypeError(
+            `Value at ${q(petNamePath.join(', '))} cannot be terminated.`,
+          );
+        }
```
kriskowal, line 86: Can `provideLookupFormula` return the `formulaIdentifier`? There are other functions like that.

packages/daemon/src/daemon.js
rekmarks: The property name, that is. `hub` for "name hub" would also be acceptable.

packages/daemon/src/daemon.js
rekmarks: Use either `hub` or `store`, not `agent`. Revert use of `pet-store` formula identifier to create `lookup` formula identifier.

packages/daemon/src/mail.js
rekmarks: This is already the case.

packages/daemon/src/daemon.js
```diff
@@ -227,6 +227,29 @@ const makeEndoBootstrap = (
...
+   * @param {string} agentFormulaIdentifier
+   * @param {string[]} path
+   * @param {import('./types.js').Terminator} terminator
+   */
+  const makeControllerForLookup = async (
+    agentFormulaIdentifier,
+    path,
+    terminator,
+  ) => {
+    terminator.thisDiesIfThatDies(agentFormulaIdentifier);
```
rekmarks, line 243: In the future, we should also have a watcher on the store such that any change in the names along the lookup path causes this formula to be terminated.

rekmarks: Now: Add terminate hook that does a reverse lookup of the resolved `lookup` formula value and calls `terminate()` on it. Return early if the reverse lookup returns an empty list.

packages/daemon/src/host.js
rekmarks: Decision: We will do this later if necessary.

packages/daemon/src/daemon.js
rekmarks: Ref: #2041

kriskowal (approved):

PR #2039, opened 2024-02-06
test(patterns,exo): pre-tests of upcoming #2038

erights:
closes: #XXXX
refs: https://github.com/Agoric/agoric-sdk/issues/8826 #2038 

## Description

Tests the compat failure behavior prior to #2038 that we believe is the cause behind our current understanding of https://github.com/Agoric/agoric-sdk/issues/8826 . Sets things up so that #2038 's differences in behavior, to give us the compat https://github.com/Agoric/agoric-sdk/issues/8826 needs, can be demonstrated by the changes it must make to these tests so they will pass again.

### Security Considerations

Since this PR is only adding tests, none. But see considerations in #2038 

### Scaling Considerations

Since this PR is only adding tests, none. But see considerations in #2038 

Since this PR is only adding tests, none. But see considerations in #2038 

### Testing Considerations

The whole point of this PR. See the description. But these are just for unit testing compat problems that we believe cause https://github.com/Agoric/agoric-sdk/issues/8826 . It does not even think about doing the needed integration tests.

Since this PR is only adding tests, none. But the point of #2038 is to repair some of these incompatibilities.

### Upgrade Considerations

Since this PR is only adding tests, none. But the point of #2038 is to repair some of these incompatibilities, which we believe are breaking the upgrade needed by https://github.com/Agoric/agoric-sdk/issues/8826 .

### DIFF
--- a/packages/exo/test/test-legacy-guard-tolerance.js
+++ b/packages/exo/test/test-legacy-guard-tolerance.js
@@ -1,0 +1,140 @@
+// @ts-check
+// eslint-disable-next-line import/order
+import { test } from './prepare-test-env-ava.js';
+
+import { passStyleOf } from '@endo/pass-style';
+import {
+  M,
+  getAwaitArgGuardPayload,
+  getInterfaceGuardPayload,
+  getMethodGuardPayload,
+} from '@endo/patterns';
+import {
+  makeLegacyAwaitArgGuard,
+  makeLegacyMethodGuard,
+  makeLegacyInterfaceGuard,
+} from './make-legacy-guards.js';
+import { makeExo } from '../src/exo-makers.js';
+
+test('legacy guard tolerance', async t => {
+  const aag = M.await(88);
+  const laag = makeLegacyAwaitArgGuard({
+    argGuard: 88,
+  });
+  const mg1 = M.callWhen(77, aag).returns(M.any());
+  const mg2 = M.callWhen(77, laag).returns(M.any());
+  const lmg = makeLegacyMethodGuard({
+    callKind: 'async',
+    argGuards: [77, laag],
+    returnGuard: M.any(),
+  });
+  const ig1 = M.interface('Foo', {
+    mg1,
+    mg2,
+  });
+  const lig = makeLegacyInterfaceGuard({
+    interfaceName: 'Foo',
+    methodGuards: {
+      mg1,
+      mg2,
+      // @ts-expect-error Legacy adaptor can be ill typed
+      lmg,
+    },
+  });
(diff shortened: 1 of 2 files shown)


packages/exo/test/test-legacy-guard-tolerance.js
```diff
@@ -0,0 +1,140 @@
...
+  getMethodGuardPayload,
+} from '@endo/patterns';
+import {
+  makeLegacyAwaitArgGuard,
+  makeLegacyMethodGuard,
+  makeLegacyInterfaceGuard,
+} from './make-legacy-guards.js';
+import { makeExo } from '../src/exo-makers.js';
+
+test('legacy guard tolerance', async t => {
```
mhofman, line 19: Why not create a `test.failing` case, instead of testing for an expected error?

erights: If #2038 was going to take awhile, I would. But since it is about to overwrite the test output, I wanted to leave behind an accurate record of what the state before #2038 was.

dckc: `test.failing` can also hide failure modes other than the one you want to test for.

erights: Stronger reason (that I think was at the back of my mind somewhere):

The test output in this PR is the correct one, i.e., without the kludge adaptors of #2038 that will become irrelevant once we no longer need to support code prior to #1712 . At that point, we should not exactly revert #2038 , but we should revert the effect of #2038 by simplifying getGuardPayloads.js, and we should exactly revert #2038's changes to this PR's test-legacy-guard-tolerance.js, i.e., restore the version in this PR.

erights: In any case, seeing the thumbs up, resolving.

packages/exo/test/test-legacy-guard-tolerance.js
```diff
@@ -0,0 +1,140 @@
...
+
+  t.deepEqual(getInterfaceGuardPayload(ig1), {
+    interfaceName: 'Foo',
+    methodGuards: {
+      mg1,
+      mg2,
+    },
+    defaultGuards: undefined,
+  });
+  // would have been ig2 is it could be made
```
Chris-Hibbert, line 85: Hard to parse. Is this what was meant?
```suggestion
  // would have been ig2 if it could be made
```

erights: yes, thanks. Done.

Chris-Hibbert (approved):

PR #2040, opened 2024-02-08
build: temporarily allow Node.js 16

michaelfig:
## Description

Temporarily fixate on `ts-api-utils@1.0.x` not to force downstream simultaneous mass migration from Node.js 16.  Those users can take migration into their own hands, rather than being a breaking change.

### Upgrade Considerations


kriskowal (approved):
🐤, How does this fix work mechanically? I think this is narrowing the range of eligible versions (from the broader constraints mentioned in some transitive dependency)

mhofman:
I think this would only work for the endo repo, right? The top level `package.json` and the `yarn.lock` has no influence on the repos consuming endo packages.

michaelfig:
> I think this would only work for the endo repo, right? The top level `package.json` and the `yarn.lock` has no influence on the repos consuming endo packages.

It's the `@typescript-eslint` devDependencies pulling in `ts-api-utils`, so if Node.js 16 is used to run `yarn` in Endo (as it is in the breaking dapp integration tests), it fails without these changes.

But you're right; I've just now pinned the versions in `eslint-plugin`, `stream-node` and `stream`, which also had references to `@typescript-eslint`.

PR #2043, opened 2024-02-09
feat(cli,daemon): Support pet name paths in `kill` command

rekmarks:
Ref: #2023 

Adds support for dot-delimited pet name paths to the `kill` command. Makes requisite changes to handle paths and lookup formulas in the mailbox `terminate()` method.

### DIFF
--- a/packages/cli/src/commands/kill.js
+++ b/packages/cli/src/commands/kill.js
@@ -2,8 +2,9 @@
 import os from 'os';
 import { E } from '@endo/far';
 import { withEndoParty } from '../context.js';
+import { parsePetNamePath } from '../pet-name.js';
 
 export const killCommand = async ({ name, partyNames }) =>
   withEndoParty(partyNames, { os, process }, async ({ party }) => {
-    await E(party).terminate(name);
+    await E(party).terminate(parsePetNamePath(name));
   });
--- a/packages/daemon/src/mail.js
+++ b/packages/daemon/src/mail.js
@@ -4,7 +4,7 @@ import { E } from '@endo/far';
 import { makePromiseKit } from '@endo/promise-kit';
 import { makeChangeTopic } from './pubsub.js';
 import { makeIteratorRef } from './reader-ref.js';
-import { assertPetName } from './pet-name.js';
+import { assertPetName, petNamePathFrom } from './pet-name.js';
 
 const { quote: q } = assert;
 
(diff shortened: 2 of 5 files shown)


packages/daemon/src/mail.js
```diff
@@ -60,11 +60,32 @@ export const makeMailboxMaker = ({
       );
     };
 
-    const terminate = async petName => {
-      const formulaIdentifier = lookupFormulaIdentifierForName(petName);
```
kumavis, line 64: i would expect `lookupFormulaIdentifierForNamePath` or just `lookupFormulaIdentifierForName` to support paths and not require any dance in implementation land

packages/cli/src/commands/kill.js
```diff
@@ -2,8 +2,9 @@
 import os from 'os';
 import { E } from '@endo/far';
 import { withEndoParty } from '../context.js';
+import { parsePetNamePath } from '../pet-name.js';
 
 export const killCommand = async ({ name, partyNames }) =>
   withEndoParty(partyNames, { os, process }, async ({ party }) => {
-    await E(party).terminate(name);
+    await E(party).terminate(parsePetNamePath(name));
```
kumavis, line 9: can we just support paths natively in `terminate`

rekmarks: In my understanding, the decision was to only make the "hub" (i.e. agent (guest / host) and, in the future, directories) `lookup` variadic.

I could change the way we call `parsePetNamePath` such that its part of how `commander` coerces options / arguments, though. (See e.g. usage of `collect` in the CLI `endo.js`.)

packages/daemon/src/mail.js
rekmarks: As part of introducing petname paths, the design decision was that only the naming hub `lookup()` would be made variadic.

Currently, "lookup" is an overloaded word in the codebase. @kriskowal and I discussed renaming pet store-associated methods that include the word "lookup" to use some other word, like "locate", instead. Would that help?

packages/cli/src/commands/kill.js
kriskowal: Parsing pet name paths must be a UI (CLI or GUI) concern because they’re likely to handle that concern differently in the long run. GUI will be in a position to support other kinds of tokenization.

rekmarks:
Deemed unsalvageable due to age. Rewrite from scratch.

PR #2046, opened 2024-02-09
feat: daemon can specify introducedNames when making guests or hosts

kumavis:
Allows host or guest to be created with petnames populated by creator

### DIFF
--- a/packages/cli/src/commands/mkguest.js
+++ b/packages/cli/src/commands/mkguest.js
@@ -3,8 +3,8 @@ import os from 'os';
 import { E } from '@endo/far';
 import { withEndoParty } from '../context.js';
 
-export const mkguest = async ({ name, partyNames }) =>
+export const mkguest = async ({ name, partyNames, introducedNames }) =>
   withEndoParty(partyNames, { os, process }, async ({ party }) => {
-    const newGuest = await E(party).provideGuest(name);
+    const newGuest = await E(party).provideGuest(name, { introducedNames });
     console.log(newGuest);
   });
--- a/packages/daemon/src/host.js
+++ b/packages/daemon/src/host.js
@@ -466,7 +466,7 @@ export const makeHostMaker = ({
       provideWebPage,
     });
 
-    const internal = harden({ receive, respond });
+    const internal = harden({ receive, respond, petStore });
 
     await provideValueForFormulaIdentifier(mainWorkerFormulaIdentifier);
 
--- a/packages/daemon/src/host.js
+++ b/packages/daemon/src/host.js
@@ -9,6 +9,7 @@ export const makeHostMaker = ({
   provideValueForFormulaIdentifier,
   provideValueForFormula,
   provideValueForNumberedFormula,
+  provideControllerForFormulaIdentifier,
   formulaIdentifierForRef,
   storeReaderRef,
   makeSha512,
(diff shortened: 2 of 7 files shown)


packages/daemon/src/host.js
```diff
@@ -380,11 +395,22 @@ export const makeHostMaker = ({
...
+      if (introducedNames !== undefined) {
+        const { petStore: newHostPetStore } = await newHostController.internal;
+        for (const [parentName, childName] of Object.entries(introducedNames)) {
+          const introducedFormulaIdentifier =
+            lookupFormulaIdentifierForName(parentName);
+          if (introducedFormulaIdentifier === undefined) {
+            continue;
+          }
+          await newHostPetStore.write(childName, introducedFormulaIdentifier);
+        }
```
kriskowal, line 411: Opportunity for `Promise.all` and `map`. Not to imply it makes much difference since writes will be serialized.

packages/cli/src/commands/mkguest.js
```diff
@@ -3,8 +3,8 @@ import os from 'os';
 import { E } from '@endo/far';
 import { withEndoParty } from '../context.js';
 
-export const mkguest = async ({ name, partyNames }) =>
+export const mkguest = async ({ name, partyNames, introducedNames }) =>
   withEndoParty(partyNames, { os, process }, async ({ party }) => {
-    const newGuest = await E(party).provideGuest(name);
+    const newGuest = await E(party).provideGuest(name, introducedNames);
```
kriskowal, line 8: This seems fine. Elsewhere, I’ve used parallel arrays. Parallel arrays might adapt better if we allow pet name paths on the key side.

kumavis: should this be an options bag?

kriskowal: It’s worth considering. None of these APIs are future-proofed yet.

packages/daemon/src/host.js
```diff
@@ -68,10 +69,32 @@ export const makeHostMaker = ({
...
+      introducedNames,
+    ) => {
+      const { petStore: newGuestPetStore } = await newController.internal;
+      await Promise.all(
+        Object.entries(introducedNames).map(async ([parentName, childName]) => {
+          const introducedFormulaIdentifier = identifyLocal(parentName);
+          if (introducedFormulaIdentifier === undefined) {
+            return;
+          }
+          await newGuestPetStore.write(childName, introducedFormulaIdentifier);
```
rekmarks, line 87:
```suggestion
      const { petStore: newPetStore } = await newController.internal;
      await Promise.all(
        Object.entries(introducedNames).map(async ([parentName, childName]) => {
          const introducedFormulaIdentifier = identifyLocal(parentName);
          if (introducedFormulaIdentifier === undefined) {
            return;
          }
          await newPetStore.write(childName, introducedFormulaIdentifier);
```

kumavis: nice catch

packages/daemon/src/host.js
```diff
@@ -68,10 +69,32 @@ export const makeHostMaker = ({
       terminator,
     });
 
+    /**
+     * @param {any} newController
```
rekmarks, line 73: I believe there's an importable controller type in `types.d.ts`.

kumavis: cool, i have a few type improvements ill add

rekmarks (approved):
LGTM

PR #2048, opened 2024-02-10
fix: just "yarn format" on master

erights:
closes: #XXXX
refs: #XXXX

## Description

Just ran `yarn format` on master, and was surprised to see these differences. I see that CI below doesn't like this, so my local setup must somehow be misconfigured.

### Upgrade Considerations


erights:
Nevermind. Deleting all the node_modules and rebuilding made the discrepancy go away. Closing.

PR #2051, opened 2024-02-10
Daemon: identify

kumavis:


rekmarks:
There are pet name path utilities in `cli/src/pet-name.js` and `daemon/src/pet-name.js` that looks like they could be used in a variety of places in this PR.

kumavis:
while this approach can yield some potential optimization, the shortcut needs to be implemented per formula type with hosts/guests being the only candidates at present.

It's not clear that this optimization is worth the limitation that lookup values must have a formulaId. This limitation might be avoidable by replacing `identify` with a
`resolve: (namePath: string[]) => Promise<{ formulaId: string|undefined, remainingPath: string }>` that identifies as far as it can, and then `lookup` is run on the remainder.

other design considerations:
this implementation does a single "fully delegated identify with recursion". You could replace this approach with an "iterated identify". fully delegated may have better performance when a long path section is across a network boundary. iterated could have better performance when paths cross many network boundaries (analogy; iterated: series of HTTP redirects, delegated: establishing a tor circuit). 

delegated and resolve both allow multiple path parts to be consumed in one step where possible. this is seen in this implementation for inspector.identify for eval record endowments. For the example path `INFO.someEval.endowments.someEndowment`, the inspector will consume three parts (petName, propertyName, endowmentName)

packages/cli/src/commands/identify.js
```diff
@@ -0,0 +1,14 @@
...
+
+export const identify = async ({ cancel, cancelled, sockPath, namePath }) =>
+  withEndoBootstrap({ os, process }, async ({ bootstrap }) => {
+    const defaultHostFormulaIdentifier = 'host';
+    const formulaId = await E(bootstrap).identifyFrom(
+      defaultHostFormulaIdentifier,
+      namePath,
+    );
+    console.log(formulaId);
+  });
```
kriskowal, line 14: This is very close to what I imagined and a step in the right direction.

High level:
- We need an API and a command that can reveal the formula type. I had envisioned that `endo identify` would do that, but perhaps that should have a name like `describe`. I think `endo describe` should also be able to print the nonce/swissnum/identifier like `endo describe --id <name>`
- Whatever API and command reveals the nonce/swissnum/identifier for a pet name path should probably be producing a URL so maybe that should be called `endo locate <pet-name-path>` and produce the URL by default, including the connection hints for the listening networks. Maybe `endo locate --json <path>` gives you the JSON representation. Maybe `endo locate --id <path>` trims that back to just the nonce.
- Maybe `locate` and `describe` are the same command or two commands with the same behavior but different defaults.

Nits:
- Let’s get rid of `cancel`, `cancelled`, and `sockPath` here and where this refactor-cruft was copied from.
- `host` is not a formula identifier anymore. We should always start looking stuff up at `E(bootstrap).host()`.
- If we move some functions from the endo bootstrap object to host objects, maybe we can just present the primary host facet as the bootstrap object.
- Maybe we should use ``process.stdout.write(`${id}\n`)`` instead of `console.log` just to break the bad habit of confusing the debugger for an output printer. ymmv.

packages/cli/src/commands/show.js
```diff
@@ -1,10 +1,14 @@
...
-import { withEndoParty } from '../context.js';
+import { withEndoBootstrap } from '../context.js';
 
-export const show = async ({ cancel, cancelled, sockPath, name, partyNames }) =>
-  withEndoParty(partyNames, { os, process }, async ({ party }) => {
-    const pet = await E(party).lookup(name);
+export const show = async ({ cancel, cancelled, sockPath, namePath }) =>
+  withEndoBootstrap({ os, process }, async ({ bootstrap }) => {
+    const defaultHostFormulaIdentifier = 'host';
+    const pet = await E(bootstrap).lookupFrom(
```
kriskowal, line 9: Ah, I see. In my head, `lookupFrom` is equivalent to `E(E(host).locate(identifier)).lookup(path)` but consolidated. The consolidation might be important since we might want to locate and lookup locally in a single event.

I think we should move the method to Host in order to stay one step away from folding the Endo Bootstrap object into Host objects. They are equally powerful since you can’t get one without the other and there’s little reason to keep them separate.

I’m not attached to the `locate(nonce) ~=> identifier` function name, and I think I at least temporarily need the name of this method to be explicitly `lookupFromFormulaNumber`, with goal of reducing that to `lookupFromIdentifier` when we no longer need to distinguish formula identifier with type from formula number.

packages/daemon/src/daemon.js
```diff
@@ -107,6 +107,51 @@ const makeEndoBootstrap = (
...
+    if (formulaType === 'host' || formulaType === 'host-id512') {
+      let storeFormulaIdentifier = `pet-store`;
+      if (formulaType === 'host-id512') {
+        storeFormulaIdentifier = `pet-store-id512:${formulaNumber}`;
+      }
+      // eslint-disable-next-line no-use-before-define
+      const petStore = await provideValueForFormulaIdentifier(
+        storeFormulaIdentifier,
+      );
+      const result = await E(petStore).identify(namePath);
```
kriskowal, line 129: We should definitely rename `petStore.locate` to `petStore.identify` as this suggests. It should be synchronous, though.

packages/daemon/src/daemon.js
```diff
@@ -107,6 +107,51 @@ const makeEndoBootstrap = (
...
+    const { type: formulaType, number: formulaNumber } = parseFormulaIdentifier(
+      originFormulaIdentifier,
+    );
+    if (formulaType === 'host' || formulaType === 'host-id512') {
+      let storeFormulaIdentifier = `pet-store`;
+      if (formulaType === 'host-id512') {
+        storeFormulaIdentifier = `pet-store-id512:${formulaNumber}`;
+      }
+      // eslint-disable-next-line no-use-before-define
+      const petStore = await provideValueForFormulaIdentifier(
```
kriskowal, line 126: I suppose we can’t make this synchronous!

packages/daemon/src/daemon.js
```diff
@@ -107,6 +107,51 @@ const makeEndoBootstrap = (
...
+      // Use result if present, otherwise fallback to identifying from the host.
+      if (result !== undefined) {
+        return result;
+      }
+    }
+    // eslint-disable-next-line no-use-before-define
+    const origin = await provideValueForFormulaIdentifier(
+      originFormulaIdentifier,
+    );
+    return E(origin).identify(namePath);
```
kriskowal, line 139: Given this is async tail recursive and has to at least reïfy pet stores (an unavoidably async operation), I withdraw my hope that `identify` can even be locally synchronous or transactional.

Given that it can’t be synchronous, there’s no reason to bundle up `locate(identifier)~.lookup(path)`. They can be peanut and butter: two separate methods that are better together.

packages/daemon/src/daemon.js
```diff
@@ -651,6 +699,24 @@ const makeEndoBootstrap = (
...
+    'eval-id512',
+    'lookup-id512',
+    'make-unconfined-id512',
+    'make-bundle-id512',
+    'guest-id512',
+    'web-bundle',
+  ];
+
+  const allowedInspectorFormulaIdentificationsByType = {
+    eval: ['endowments', 'worker'],
```
kriskowal, line 712: I’m not sure that inspectors generalize this well. `endowments` is an array of identifiers and we want the corresponding values, whereas `worker` is a single identifier. This does not mention `source` which is just a string.

In short, I think we should have a boring identify function for each formula type rather than a generalization because it will be more legible and easier to debug.

packages/daemon/src/daemon.js
```diff
@@ -665,28 +731,81 @@ const makeEndoBootstrap = (
...
+        : [maybeNamePath];
+      const [petName, propertyName, ...namePathRest] = namePath;
+      assertPetName(petName);
+      assertPetName(propertyName);
+      const entryFormulaIdentifier = petStore.identifyLocal(petName);
+      if (entryFormulaIdentifier === undefined) {
+        return undefined;
+      }
+      const { type: formulaType, number: formulaNumber } =
+        parseFormulaIdentifier(entryFormulaIdentifier);
```
kriskowal, line 776: By the time we get here, we may have removed the type from the formula identifier, but it can be trivially recovered from the controller object for the number.

kriskowal: When we have a formula for every formula identifier (just the number), it might make sense to just generally store that on the controller too, so we benefit from the memo and avoid redundant reads from the formula store.

packages/daemon/src/host.js
```diff
@@ -170,7 +171,7 @@ export const makeHostMaker = ({
      * @param {string | 'NONE' | 'SELF' | 'ENDO'} partyName
      */
     const providePowersFormulaIdentifier = async partyName => {
-      let guestFormulaIdentifier = lookupFormulaIdentifierForName(partyName);
+      let guestFormulaIdentifier = identifyLocal(partyName);
```
kriskowal, line 174: I like abbreviating `lookupFormulaIdentifierForName` to `identifyLocal` generally. That could be one easy-to-review mechanical refactor on its own.

Since the difference between `identify` and `identifyLocal` is more like `identifyPetNamePath` vs `identifyPetName` respectively, I could buy `identify` (implied pet name path since it’s public API) and `identifyName` instead.

packages/daemon/src/mail.js
```diff
@@ -35,33 +35,40 @@ export const makeMailboxMaker = ({
...
+      const namePath = Array.isArray(maybeNamePath)
+        ? maybeNamePath
+        : [maybeNamePath];
+      const [headName, ...namePathRest] = namePath;
+      const formulaIdentifier = identifyLocal(headName);
+      if (formulaIdentifier === undefined) {
+        return undefined;
+      } else {
+        return identifyFrom(formulaIdentifier, namePathRest);
+      }
```
kriskowal, line 58: No notes.

packages/daemon/src/pet-store.js
```diff
@@ -67,11 +72,27 @@ export const makePetStoreMaker = (filePowers, locator) => {
...
+      const namePath = Array.isArray(maybeNamePath)
+        ? maybeNamePath
+        : [maybeNamePath];
+      const [headName, ...namePathRest] = namePath;
+      const formulaIdentifier = identifyLocal(headName);
+      if (formulaIdentifier === undefined) {
+        return undefined;
+      } else {
+        return identifyFrom(formulaIdentifier, namePathRest);
+      }
```
kriskowal, line 95: I don’t think this should exist at this level of abstraction but we should factor out an intermediate layer for name hubs between mail and pet store whenever we introduce directories.

PR #2042, opened 2024-02-08
feat(ses,pass-style,marshal): permit Promise.any, AggregateError, cause

erights:
closes: #550 
refs: https://github.com/endojs/endo/issues/550#issuecomment-1935013037 https://github.com/cosmology-tech/cosmos-kit/blob/b95e2f92585fd12d7532e846b98cd3b757c1acb8/packages/core/src/utils/endpoint.ts#L28 https://github.com/Agoric/agoric-sdk/pull/7937

## Description

We had previously omitted `AggregateError` from `permits.js`, effectively not permitting it as part of SES, because of concern with it not following the `NativeError` patterns we assume for the other errors. On further examination just now, the only non-uniformity I see are the addition of the own property `errors`. As an own property, it is irrelevant to the job of `permits.js`.

Likewise we have postponed dealing with Error `cause`. But it is an own property as well, and so besides the point of `permits.js`.

Having omitted `AggregateError`, we also declined to permit `Promise.any` since it returns an `AggregateError` instance. Thus, this PR simultaneously permits `Promise.any`.

- [x] Still need to adapt the ses console
- [x] Still need to adapt pass-style
- [x] Still need to adapt marshal
- [x] Still need to test

I am pleasantly surprised that it does not look like I need to adapt the taming of the `Error` constructor itself to add a `cause` argument to the replacement safe constructors. The replacement constructors already take and pass through `...rest` arguments. But

- [x] Still need to test `cause` argument for the faux `Error` constructors, as well as a representative set of the permitted error constructors.

### Security Considerations

Unlikely. But anything that permits more of the engine's surface area does in theory increase vulnerability.

### Scaling Considerations

None

### Documentation Considerations

If we currently document this weird exception to normal JS expectations, we can delete that explanation.

Otherwise, no docs needed since this PR just removes a surprising difference from normal JS developer expectations.

### Testing Considerations

Currently, to test the handling of `*Error` subclasses of `Error`, we just test with a representative subclass such as `UriError`. This PR duplicates some of these tests for `AggregateError`.

### Compatibility Considerations

We need to examine what happens if a ses that permits `AggregateError` talks to a ses that does not. This is mostly a `marshal` issue.

Some agoric-sdk compat concerns noted at https://github.com/endojs/endo/issues/550#issuecomment-1935013037

[...]

Issue #550, opened 2020-12-23
Add AggregateError and Promise.any to whitelist


erights:
#549 updates the whitelist to ES2021 including `Promise.any`. Thus it needs to include `AggregateError` as well, since `Promise.any` makes instances of this. However #549 doesn't do this yet since it doesn't fit into the `NativeError` pattern and will need a bit more thought.

mhofman:
A little investigation uncovers https://bugs.chromium.org/p/chromium/issues/detail?id=1379054.

I suspect this issue is likely fixed already in v8.

erights:
Note that @agoric/internal defines and exports a `makeAggregateError` function that, because this issue is unfixed, makes a faux AggegateError instance which is actually a direct instance of `Error` with an extra own `errors` field to emulate an AggregateError instance. This export is used elsewhere in agoric-sdk. 

Fortunately, since we don't treat the exports from @agoric/internal as public API we need to continue to support, after this bug is fixed, we should eventually be able to remove `makeAggregateError` and fix all sites that call it.

erights:
I also see other existing uses of `AggregateError` itself in agoric-cli/src/install.js. I do not understand how this can currently work at all. Obviously, by permitting actual `AggregateError`, we need to be careful not to break this case.

erights:
It works because it branches on the existence of the global. This explains why it currently works. It also means that fixing #550 will cause that code to take a different branch that should work but is probably not yet testable.

Attn @michaelfig who wrote the relevant uses of `AggregateError` in install.js

erights:
Needed due to https://github.com/cosmology-tech/cosmos-kit/blob/b95e2f92585fd12d7532e846b98cd3b757c1acb8/packages/core/src/utils/endpoint.ts#L28

Although @samsiegart currently has a workaround. So less urgent.

michaelfig:
> Attn @michaelfig who wrote the relevant uses of `AggregateError` in install.js

Where is this "install.js" of which you speak?

erights:
https://github.com/Agoric/agoric-sdk/blob/b9ffe482411b63acfaffbbfbbc1eba6bb241808e/packages/agoric-cli/src/install.js#L177

michaelfig:
> https://github.com/Agoric/agoric-sdk/blob/b9ffe482411b63acfaffbbfbbc1eba6bb241808e/packages/agoric-cli/src/install.js#L177

[...]

PR #2052, opened 2024-02-11
feat(marshal): tolerate extra error props

erights:
closes: #XXXX
refs: #2042 #550 

## Description

To enable future versions of marshal to send extra error properties, like `cause` and `error`, marshal must tolerate receiving them.

Sets the stage for #2024 

### Security Considerations
by tolerating the reception of extra error props, we no longer inhibit senders from sending them, including misspelled or otherwise mistaken ones. This weaken the degree to which unmarshaling validates. This is the necessary price of accommodating expected future growth.

### Scaling Considerations
none
none
### Testing Considerations
tests included also serve as a baseline for #2042, which needs to change them in order to pass again. This provides a record of the difference between this PR and #2042.
The whole point, is to accommodate an expected future in which some extra properties are sent, such as `cause` and `errors`.
### Upgrade Considerations

This change, to tolerate receiving extra properties, needs to roll out everywhere, before we can roll out a future change sending extra properties such as `cause` or `errors`. (#2042, though it is a future change that builds on this one, is not that future change because it still does not send any extra properties.)

This PR is not breaking. Nor is it combined with #2042. The future change where extra error properties are sent will not be a breaking change compared to code incorporating this PR (or this PR + #2042). However, that future change will be breaking compared to code prior to this PR.

No news update needed.

### DIFF
--- a/packages/marshal/NEWS.md
+++ b/packages/marshal/NEWS.md
@@ -1,5 +1,9 @@
 User-visible changes in `@endo/marshal`:
 
+# next
+
+- Tolerates receiving extra error properties (https://github.com/endojs/endo/pull/2052). Once pervasive, this tolerance will eventually enable additional error properties to be sent. The motivating examples are the JavaScript standard properties `cause` and `errors`. This change also enables smoother interoperation with other languages with their own theories about diagnostic information to be included in errors.
+
 # v0.8.1 (2022-12-23)
 
 - Remote objects now reflect methods present on their prototype chain.
--- a/packages/marshal/package.json
+++ b/packages/marshal/package.json
@@ -41,6 +41,7 @@
   },
   "homepage": "https://github.com/endojs/endo#readme",
   "dependencies": {
+    "@endo/common": "^1.0.2",
     "@endo/errors": "^1.0.2",
     "@endo/eventual-send": "^1.1.0",
     "@endo/nat": "^5.0.2",
--- a/packages/marshal/src/marshal-justin.js
+++ b/packages/marshal/src/marshal-justin.js
@@ -390,6 +390,10 @@ const decodeToJustin = (encoding, shouldIndent = false, slots = []) => {
 
         case 'error': {
           const { name, message } = rawTree;
+          // TODO cause, errors, AggregateError
+          // See https://github.com/endojs/endo/pull/2052
+          name !== `AggregateError` ||
+            Fail`AggregateError not yet implemented in marshal-justin`;
           return out.next(`${name}(${quote(message)})`);
         }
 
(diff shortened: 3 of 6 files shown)


erights:
@gibson042 , as far as I could tell, I did not need to make a corresponding change to `encodePassable.js`. Is this true, or did I miss something? What about tests of `encodePassable`?

erights:
@kriskowal , although #2042 is also ready for review, I do not expect to get it into the next endo release. It is also large, and therefore has more design issues we may need to argue through.

But because of the stage setting nature of this PR and well as its minimal effect, I would very much like to get this PR into the next endo release if possible.

erights:
> > ### Upgrade Considerations
> > None
> 
> I wonder if it's more subtle than that. This one (call it n) is not a breaking change, and a future change (n+m) to take advantage of this won't look like a breaking change compared to this one, but n+m will be a breaking change w.r.t. n-1.
> 
> So this looks like one of those changes that has to roll out ~everywhere before anything can take advantage of it.

Ok, I updated the PR "Upgrade Considerations" section to explain this. But it still doesn't seem to be that this change should be considered breaking or NEWSworthy. Reviewers, please let me know if you disagree, and what you suggest that I do. Thanks.

dckc (commented):
this looks plausible to me, but the history goes back to #550 , so I'm inclined to defer to someone with more context. LMK if they're not available in a useful timeframe.

kriskowal (approved):

kriskowal:
This is a relaxation and a new feature. I do think a note in NEWS about the new capability to transmit errors with properties is worth mentioning and will be useful for future investigations.

kriskowal:
> So this looks like one of those changes that has to roll out ~everywhere before anything can take advantage of it.

This seems accurate to me.

In isolation, this allows a node to receive messages that no node can yet produce. When they can produce such messages, a new node will be able to send messages that will be rejected by an old node. There is some hazard that new programs will pass tests locally and discover they fail in integration with an old node. When we broach that, we need a migration plan.

erights:
> This is a relaxation and a new feature. I do think a note in NEWS about the new capability to transmit errors with properties is worth mentioning and will be useful for future investigations.

Done

gibson042 (commented):
> @gibson042 , as far as I could tell, I did not need to make a corresponding change to `encodePassable.js`. Is this true, or did I miss something? What about tests of `encodePassable`?

Confirmed; passable decoders already support decoding errors into [whatever comes from a caller-provided `decodeError`](https://github.com/endojs/endo/blob/70cdcd7f314d459f9a53cf5cd76da2e1e00691c4/packages/marshal/src/encodePassable.js#L714) (and don't even assert that the results are errors, although they probably should) and encoders [require only that the result has the right prefix](https://github.com/endojs/endo/blob/70cdcd7f314d459f9a53cf5cd76da2e1e00691c4/packages/marshal/src/encodePassable.js#L534-L543).

But a dedicated test in [test-encodePassable.js](https://github.com/endojs/endo/blob/master/packages/marshal/test/test-encodePassable.js) might make sense.

PR #2054, opened 2024-02-12
fix(daemon): use naming convention "identifyLocal"

kumavis:
unify naming of methods in `petStore` and `mail` that synchronously return a `formulaIdentifier` for a non-path petname

### DIFF
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -671,7 +671,7 @@ const makeEndoBootstrap = (
      * inspector for the value of the given pet name.
      */
     const lookup = async petName => {
-      const formulaIdentifier = petStore.lookup(petName);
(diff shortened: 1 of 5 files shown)


rekmarks (commented):
Should the pet store's `reverseLookup` be renamed `reverseIdentify`?

Also, should we just call it `identify` instead of `identifyLocal`? If the signature of the eventual non-local `identify` will be the same, we can just call them all the same thing.

kumavis:
@rekmarks id opt to keep them separate at this time. by being local we know it will be sync and thus never return a promise

kumavis:
> Should the pet store's reverseLookup be renamed reverseIdentify?

probably

rekmarks (approved):
TODO: Rename `reverseLookup` to `reverseIdentify` in `pet-store.js`.

PR #2057, opened 2024-02-13
feat(daemon): expose teardown for resetting without restarting

kumavis:
this enabled cleaning up test daemon dirs (but not outer `tmp` dir) after tests


kumavis:
`rimraf` package has included a fair amount of workarounds for windows wonkiness around deletion (better called "mark for delete")

kumavis:
idk/

Issue #1828, opened 2023-10-12
Pet Daemon: Prototype of remote connections between daemons using Tailscale, November 2023


aj-agoric:
## What is the Problem Being Solved?
Pet Daemon: Prototype of remote connections between daemons using Tor, November 2023
From [MetaMask Requirements](https://docs.google.com/document/d/16u_JpaMNc-5Xp4l0fQn3OPAWFUIX-5gxtPWuIu6-fQ0/edit)

aj-agoric:
Triaged in today's meeting and decided this is the best guess for now

kriskowal:
@danfinlay @kumavis @rekmarks I’ve revised this issue to reflect that we achieved pet daemon connectivity using Tailscale instead of Tor for the purposes of our prototype and closing. This ends the prototype spike and I believe adequately closes out our planned deliverables for the 2023 grant, with your blessing. See also https://github.com/endojs/endo/issues/1829

PR #2060, opened 2024-02-13
feat(daemon): Add TCP net layer

kumavis:
- breaks out `SocketPowers` from `NetworkPowers`
- adds `connectPort` for tcp net layer to be utilized later

Ref: https://github.com/endojs/endo/pull/1993

### DIFF
--- a/packages/daemon/src/daemon-node-powers.js
+++ b/packages/daemon/src/daemon-node-powers.js
@@ -179,13 +179,9 @@ export const makeHttpPowers = ({ http, ws }) => {
 /**
  * @param {object} modules
  * @param {typeof import('net')} modules.net
- * @param {typeof import('http')} modules.http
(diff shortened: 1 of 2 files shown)


rekmarks (approved):
LGTM

dckc:
I'm interested to try this out. How is it used? Are there relevant tests?

kriskowal:
Relevant tests https://github.com/endojs/endo/blob/983900db60e4f388614a4003c8237102d3806236/packages/daemon/test/test-endo.js#L1179-L1251

However, it might be good to wait for my signal. We’ll have a user-level workflow for inviting peers to connect and message delivery between connected peers within a week. I’ll post an update at https://github.com/endojs/endo/issues/1828, maybe even carve out the narrower issue we’ll hit with TCP.

PR #2064, opened 2024-02-13
refactor(daemon): Make least-authority formula numbered

rekmarks:
Converts the `least-authority` formula to a numbered, id512 equivalent. The formula number used will always be the zero id512, but this can be changed in the future.


kumavis (approved):
gang gang

PR #2067, opened 2024-02-13
Daemon: Change unsafe import formula from path to specifier

kumavis:
if we want to add backwards compat we can attempt to resolve the old path, but the attempt at backwards compat that was in the cherry-picked commit looked broken (didnt re-resolve)

update of https://github.com/endojs/endo/pull/1899

### DIFF
--- a/packages/daemon/src/worker.js
+++ b/packages/daemon/src/worker.js
@@ -14,6 +14,21 @@ const endowments = harden({
   URL,
 });
 
+const normalizeFilePath = path => {
+  // Check if the path is already a file URL.
+  if (path.startsWith('file://')) {
+    return path;
+  }
+  // Windows path detection and conversion (look for a drive letter at the start).
+  const isWindowsPath = /^[a-zA-Z]:/.test(path);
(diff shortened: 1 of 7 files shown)


kumavis:
windows berg:
> Only URLs with a scheme in: file, data are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'd:'

packages/daemon/src/worker.js
```diff
@@ -14,16 +14,30 @@ const endowments = harden({
...
 });
 
+const normalizeFilePath = path => {
+  // Check if the path is already a file URL.
+  if (path.startsWith('file://')) {
+    return path;
+  }
+  // Windows path detection and conversion (look for a drive letter at the start).
+  const isWindowsPath = /^[a-zA-Z]:/.test(path);
+  if (isWindowsPath) {
```
rekmarks, line 24: We should probably replace this with `os.platform` instead.

rekmarks (approved):
This will do for now.

PR #2068, opened 2024-02-13
Daemon: List special names

kumavis:
Replaces https://github.com/endojs/endo/pull/1918

dropped shorthand flag for `--all` bc its already pretty short and didnt want to confuse with `--as`

### DIFF
--- a/packages/cli/src/commands/list.js
+++ b/packages/cli/src/commands/list.js
@@ -3,12 +3,16 @@ import os from 'os';
 import { E } from '@endo/far';
 import { withEndoHost } from '../context.js';
 
-export const list = async ({ directoryName }) =>
-  withEndoHost({ os, process }, async ({ party }) => {
+export const list = async ({ directoryName, special, all }) =>
+  withEndoHost({ os, process }, async ({ host: party }) => {
     if (directoryName !== undefined) {
       party = E(party).lookup(directoryName);
     }
-    const petNames = await E(party).list();
+    const petNames = await (() => {
+      if (all) return E(party).listAll();
+      if (special) return E(party).listSpecial();
+      return E(party).list();
+    })();
     for await (const petName of petNames) {
       console.log(petName);
     }
--- a/packages/cli/src/commands/list.js
+++ b/packages/cli/src/commands/list.js
@@ -1,10 +1,13 @@
 /* global process */
 import os from 'os';
 import { E } from '@endo/far';
-import { withEndoParty } from '../context.js';
+import { withEndoHost } from '../context.js';
 
-export const list = async ({ partyNames }) =>
-  withEndoParty(partyNames, { os, process }, async ({ party }) => {
+export const list = async ({ directoryName }) =>
+  withEndoHost({ os, process }, async ({ party }) => {
+    if (directoryName !== undefined) {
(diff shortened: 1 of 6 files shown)


kumavis:
ill fix it, but feel free to review

packages/cli/src/endo.js
```diff
@@ -254,13 +254,14 @@ export const main = async rawArgs => {
     });
 
   program
-    .command('list')
-    .option(...commonOptions.as)
-    .description('show names')
-    .action(async cmd => {
-      const { as: partyNames } = cmd.opts();
+    .command('list [directory]')
+    .description('show names (dot delimited pet name path)')
```
rekmarks, line 258:
```suggestion
    .command('list [naming hub]')
    .description('show names known to the current or specified naming hub')
```

rekmarks (changes requested):
Elsewhere we have started to use the language of "naming hub" to refer to "thing that manages a namespace". I think we should replace references to "directory" with "hub" or "namingHub". If the latter, we should update existing references to "hub" with "namingHub" (that's probably what we should do).

rekmarks (approved):
LGTM

packages/cli/src/commands/list.js
```diff
@@ -1,11 +1,18 @@
...
-  withEndoParty(partyNames, { os, process }, async ({ party }) => {
-    const petNames = await E(party).list();
+export const list = async ({ directoryName, special, all }) =>
+  withEndoHost({ os, process }, async ({ host: party }) => {
+    if (directoryName !== undefined) {
+      party = E(party).lookup(directoryName);
+    }
+    const petNames = await (() => {
+      if (all) return E(party).listAll();
+      if (special) return E(party).listSpecial();
```
kriskowal, line 13: Do the `--all` and `--special` flags work? My preference is to eliminate them and just list special names with the usual `list()`.

packages/daemon/src/mail.js
```diff
@@ -75,6 +75,13 @@ export const makeMailboxMaker = ({
       return controller.terminator.terminate();
     };
 
+    const list = () => harden(petStore.list());
+
+    const listSpecial = () => harden(Object.keys(specialNames).sort());
+
+    const listAll = () =>
+      harden([...Object.keys(specialNames).sort(), ...petStore.list()]);
```
kriskowal, line 83: Ah, they do work. Let’s just make `list()` have the behavior of `listAll()` and not have `listSpecial()`.  Sound good?

packages/daemon/test/test-endo.js
```diff
@@ -1071,3 +1071,34 @@ test('evaluate name resolved by lookup path', async t => {
...
   await stop(locator);
 });
+
+test('list special names', async t => {
+  const { promise: cancelled, reject: cancel } = makePromiseKit();
+  t.teardown(() => cancel(Error('teardown')));
+  const locator = makeLocator('tmp', 'list-names');
+
+  await stop(locator).catch(() => {});
+  await reset(locator);
```
kriskowal, line 1081: @rekmarks Are you on reset-rename-purge?

rekmarks: Aye, I will sort out the remaining references tomorrow.

PR #2069, opened 2024-02-13
fix(bundle-source): Export types properly

kriskowal:
## Description

Using dynamic import in bundleSource made it possible to use `bundleSource` for one format and not entrain all the tooling for other formats, like Rollup for `getExport` and Babel for `endoZipBase64`. But, it broke the flow of type information. This change proposed by @michaelfig rectifies the internal and external types.

### Security Considerations

None.

### Scaling Considerations

None.

None.

### Testing Considerations

This PR does not adequately test the exported type signature.
It would pair well with a change that relies on the exported types.

None.

### Upgrade Considerations

None.

- ~~[ ] Includes `*BREAKING*:` in the commit message with migration instructions for any breaking change.~~
- ~~[ ] Updates `NEWS.md` for user-facing changes.~~

### DIFF
--- a/packages/bundle-source/src/bundle-source.js
+++ b/packages/bundle-source/src/bundle-source.js
@@ -2,8 +2,9 @@
 const DEFAULT_MODULE_FORMAT = 'endoZipBase64';
 const SUPPORTED_FORMATS = ['getExport', 'nestedEvaluate', 'endoZipBase64'];
 
-/** @type {import('./types').BundleSource} */
-// @ts-ignore cast
+/**
+ * @type {import('./types').BundleSourceGeneral}
+ */
 const bundleSource = async (
   startFilename,
   options = {},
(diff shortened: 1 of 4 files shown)


kriskowal:
I’ve tacked on a commit for @dckc to relax the bundle cache option to be actually optional.

packages/bundle-source/src/bundle-source.js
```diff
@@ -52,4 +51,4 @@ const bundleSource = async (
   }
 };
 
-export default bundleSource;
+export default /** @type {import('./types').BundleSource} */ (bundleSource);
```
dckc, line 54: this seems redundant w.r.t. the type annotation where `bundleSource` is declared.

kriskowal: I’ve changed this to `@satisfies` and it seems to work. But, @michaelfig recast the type from `BundleSourceGeneral` to `BundleSource` and that appears to have been necessary. I will confess to not fully understanding the `bundleSource` overloads.

packages/bundle-source/src/zip-base64.js
```diff
@@ -17,9 +17,12 @@ const textEncoder = new TextEncoder();
 const textDecoder = new TextDecoder();
 const readPowers = makeReadPowers({ fs, url, crypto });
 
+/**
+ * @type {import('./types').BundleSourceWithOptions}
+ */
 export async function bundleZipBase64(
   startFilename,
-  options = {},
```
kumavis, line 22: options mandatory now?

kumavis (approved):

PR #2071, opened 2024-02-14
fix(daemon): enhance types for pet-store

kumavis:
I'm not sure what the ideal pattern is.
This change is an improvement because it gives us a single source of truth for the types, whereas before we have repeated and unaligned definitions.
Its also worse because now the definitions are not readable above the functions implementations.
Apparently this setup also affects jsdoc generation

im not sure what a better way is, but i think this is worth merging for now


rekmarks (approved):
LGTM!

PR #2072, opened 2024-02-14
refactor(daemon): Make all formula numbers unique

rekmarks:
Ensures that all formula numbers unique by either deriving them from a root "seed" value or randomly generating them. The seed value exists until the next `purge`.

We in some cases rely on formula number derivation paths to reference other values. For example, the pet store number is derived from the inspector number, so that the inspector factory function can easily reference the pet store.


kumavis (approved):

kumavis (commented):
works for me. while we might want just randomness for uniformity, this works and is a fine source of pseudorandomness for our needs

PR #2073, opened 2024-02-14
refactor(daemon,cli): Rename `terminate`/`kill` to `cancel`

rekmarks:
Renames the daemon mailbox `terminate` method and CLI `kill` command to `cancel`. Also renames the daemon's `terminator` abstraction to `context`.


packages/daemon/src/context.js
```diff
@@ -2,75 +2,78 @@
...
       controllerForFormulaIdentifier.delete(formulaIdentifier);
-      for (const dependentTerminator of dependents.values()) {
-        dependentTerminator.terminate(` ${prefix}`);
+      for (const dependentContext of dependents.values()) {
+        dependentContext.cancel(reason, ` ${prefix}`);
       }
       dependents.clear();
 
-      resolveTerminated(Promise.all(hooks.map(hook => hook())).then(() => {}));
+      resolveDisposed(Promise.all(hooks.map(hook => hook())).then(() => {}));
```
kumavis, line 41: any idea what the final then is for

packages/daemon/src/daemon.js
```diff
@@ -131,26 +131,25 @@ const makeEndoBootstrap = async (
...
     const { workerTerminated, workerDaemonFacet } =
       await controlPowers.makeWorker(
         workerId512,
         daemonWorkerFacet,
-        Promise.race([workerCancelled, gracePeriodElapsed]),
+        Promise.race([hardCancelled, gracePeriodElapsed]),
       );
 
-    const terminate = async () => {
+    const softCancel = async () => {
```
kumavis, line 152: what's softCancel

kumavis (approved):
I know it's not your word choice but context is so general I can't guess what it's use is related to. Terminator was at least closer. but yeah words

packages/daemon/src/context.js
rekmarks: It cause the promise returned by `Promise.all()` to resolve to `undefined` as opposed to the resolved values of the hook functions. Since the hooks are about creating side effects, I think swallowing their return values in this fashion is the correct behavior.

packages/daemon/src/daemon.js
rekmarks: I renamed the couplet `gracefulCancel`/`forceCancel`. I think the former just means "respect the worker's termination grace period", but I won't add more documentation for right now.

packages/daemon/src/context.js
rekmarks: I added a comment for this.

kriskowal:
Regarding the admittedly dubious name choice, context is a carrier from parent to child contexts, in particular for cancellation but also potentially other things.

- https://pkg.go.dev/context
- https://github.com/tc39/proposal-async-context

And relatedly of course [Abort](https://developer.mozilla.org/en-US/docs/Web/API/AbortController)

PR #2076, opened 2024-02-14
fix(cli): Improve log behavior when stopped

rekmarks:
Doing `endo log` when the daemon is stopped would previously provide unhelpful results. Now, the user is notified that the daemon is not running instead. `endo log -f` will not exit but wait for the daemon to start, then operate normally.


kriskowal (approved):
Maybe this should work on Windows someday.

PR #2077, opened 2024-02-15
chore: Release 2024-02-14

kriskowal:
Regular release of Endo packages.


erights (changes requested):
Please scan all NEWS.md files for `# next` or `# next release` or similar that still needs updating to the explicit version. I know of at least one: in @endo/marshal

erights (approved):
No need for another round trip thru me. LGTM

PR #2079, opened 2024-02-15
feat(daemon): Expose cancellation context in caplets

rekmarks:
Updates #2086 

Exposes the `context` object to confined and unconfined caplets by passing it as a second argument to their `make()` function.

### DIFF
--- a/packages/daemon/src/daemon-node-powers.js
+++ b/packages/daemon/src/daemon-node-powers.js
@@ -234,7 +234,7 @@ export const makeSocketPowers = ({ net }) => {
     );
 
   /** @type {import('./types.js').SocketPowers['connectPort']} */
-  const connectPort = ({ port, host, cancelled }) =>
+  const connectPort = ({ port, host }) =>
     new Promise((resolve, reject) => {
       const conn = net.connect(port, host, err => {
         if (err) {
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -46,6 +46,10 @@ const makeInspector = (type, number, record) =>
     list: () => Object.keys(record),
   });
 
+/**
+ * @param {import('./types.js').Context} context - The context to make far.
+ * @returns {import('./types.js').FarContext} The far context.
+ */
 const makeFarContext = context =>
   Far('Context', {
     cancel: context.cancel,
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -46,6 +46,14 @@ const makeInspector = (type, number, record) =>
     list: () => Object.keys(record),
   });
 
+const makeFarContext = context =>
+  Far('Context', {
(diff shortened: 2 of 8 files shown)


packages/daemon/src/daemon-node-powers.js
```diff
@@ -234,7 +234,7 @@ export const makeSocketPowers = ({ net }) => {
     );
 
   /** @type {import('./types.js').SocketPowers['connectPort']} */
-  const connectPort = ({ port, host, cancelled }) =>
```
rekmarks, line 237: Unrelated unused argument cleanup.

packages/daemon/src/daemon.js
```diff
@@ -284,7 +296,12 @@ const makeDaemonCore = async (
       // eslint-disable-next-line no-use-before-define
       provideValueForFormulaIdentifier(guestFormulaIdentifier)
     );
-    const external = E(workerDaemonFacet).makeUnconfined(specifier, guestP);
+    const external = E(workerDaemonFacet).makeUnconfined(
+      specifier,
+      guestP,
+      // TODO fix type
+      /** @type {any} */ (makeFarContext(context)),
```
rekmarks, line 303: `yarn lerna run build:types` errors without this.

packages/daemon/src/mutex.js
```diff
@@ -21,7 +28,7 @@ export const makeMutex = () => {
     enqueue: async asyncFn => {
       await lock();
       try {
-        return await asyncFn();
+        return await (asyncFn ? asyncFn() : null);
```
kriskowal, line 31: Better to default the ayncFn to noop.

kriskowal: I also don’t see this getting used. Is it necessary?

packages/daemon/src/mutex.js
```diff
@@ -1,7 +1,14 @@
 import { makeQueue } from '@endo/stream';
 
 /**
- * @returns {{ lock: () => Promise<void>, unlock: () => void, enqueue: (asyncFn: () => Promise<any>) => Promise<any> }}
+ * @typedef {object} Mutex
+ * @property {() => Promise<void>} lock
+ * @property {() => void} unlock
+ * @property {(asyncFn?: () => Promise<any>) => Promise<any>} enqueue
```
kriskowal, line 7:
```suggestion
 * @template T
 * @property {() => Promise<void>} lock
 * @property {() => void} unlock
 * @property {(asyncFn?: () => Promise<T>) => Promise<T>} enqueue
```

kriskowal (approved):
Please consider the questions. I do not need to review again.

packages/daemon/src/mutex.js
rekmarks: It is used at the top of the added `cancelValue` method. I did change the implementation to use a default noop function, however.

PR #2082, opened 2024-02-15
chore: Merge Pet Dæmon and CLI

kriskowal:
The `daemon` and `cli` packages will evolve on `master` going forward, but are marked `private` so they do not escape to `npm` until we are prepared to make forward compatibility warranties.


rekmarks (approved):
Avancez!

PR #2084, opened 2024-02-16
daemon: enhance types for mail

kumavis:
types for the type gods :moyai:

### DIFF
--- a/packages/daemon/src/mail.js
+++ b/packages/daemon/src/mail.js
@@ -8,13 +8,29 @@ import { assertPetName } from './pet-name.js';
 
 const { quote: q } = assert;
 
+/**
+ * @param {object} args
+ * @param {import('./types.js').ProvideValueForFormulaIdentifier} args.provideValueForFormulaIdentifier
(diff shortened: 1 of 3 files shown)


packages/daemon/src/mail.js
```diff
@@ -97,37 +113,31 @@ export const makeMailboxMaker = ({
...
-     */
+    /** @type {import('./types.js').Mail['provideLookupFormula']} */
     const provideLookupFormula = async petNamePath => {
       // The lookup formula identifier consists of the hash of the associated
       // naming hub's formula identifier and the pet name path.
-      // A "naming hub" is an objected with a variadic lookup method. At present,
-      // the only such objects are guests and hosts.
+      // A "naming hub" is an objected with a variadic lookup method. It includes
+      // objects such as guests and hosts.
       const hubFormulaIdentifier = identifyLocal('SELF');
```
kriskowal, line 131: Aside: I think this identifier is a lexical variable and doesn’t require `identifyLocal`.

kriskowal (approved):

packages/daemon/src/mail.js
kumavis: https://github.com/endojs/endo/pull/2085

Issue #2087, opened 2024-02-17
daemon: All formulas should be associated with a JSON file


rekmarks:
Currently, some formulas (e.g. `host`) do not have a JSON object or file associated with them. In order to, among other things, remove the formula type prefix from formula identifiers (leaving only the number), all formulas must be associated with an object storing their `type`, and be written to disk as JSON files.

Given a formula number alone, it should be possible to retrieve this file from disk. This ought to be possible as of #2075.

Issue #2088, opened 2024-02-17
daemon: Remove formula type prefix from formula identifiers


rekmarks:
Following #2087, we should remove the formula type prefix from formula identifiers, leaving only the formula number.

PR #2089, opened 2024-02-17
feat(daemon): store endo bootstrap and default host formulas

kumavis:
- Factors out endo configuration ( eg default host formulaIdentifier ) from newly named `daemonCore`.
- Introduces `incarnateXyz` methods to be used when creating a new thing for the first time.
- Persists all formulas and requires them to be incarnated before they can be reified via `provideValueForFormulaId`
- Adds Handles as a generic pointer to another formula, where the pointer is exposed on the internal facet only. This is currently used to give a guest a reference to the host without exposing host endowments. mail uses `provideControllerForFormulaIdentifierAndResolveHandle` to resolve the handle.

Updates: #1993 
Closes: #2087

### DIFF
--- a/packages/daemon/src/daemon-node-powers.js
+++ b/packages/daemon/src/daemon-node-powers.js
@@ -504,6 +504,12 @@ export const makeDaemonicPersistencePowers = (
     await Promise.all([statePathP, cachePathP, ephemeralStatePathP]);
   };
 
+  const isRootInitialized = async () => {
+    const noncePath = filePowers.joinPath(locator.statePath, 'nonce');
+    const nonce = await filePowers.maybeReadFileText(noncePath);
+    return nonce !== undefined;
+  };
+
   const provideRootNonce = async () => {
     const noncePath = filePowers.joinPath(locator.statePath, 'nonce');
     let nonce = await filePowers.maybeReadFileText(noncePath);
@@ -636,6 +642,7 @@ export const makeDaemonicPersistencePowers = (
   return harden({
     initializePersistence,
     provideRootNonce,
+    isRootInitialized,
     makeContentSha512Store,
     readFormula,
     writeFormula,
--- a/packages/daemon/src/daemon-node-powers.js
+++ b/packages/daemon/src/daemon-node-powers.js
@@ -512,8 +512,7 @@ export const makeDaemonicPersistencePowers = (
       nonce = await cryptoPowers.randomHex512();
       await filePowers.writeFileText(noncePath, `${nonce}\n`);
     }
-    const rootNonce = nonce.trim();
-    return { value: rootNonce, isNewlyCreated };
+    return { rootNonce: nonce.trim(), isNewlyCreated };
   };
 
   const makeContentSha512Store = () => {
--- a/packages/daemon/src/daemon-node-powers.js
+++ b/packages/daemon/src/daemon-node-powers.js
@@ -504,20 +504,16 @@ export const makeDaemonicPersistencePowers = (
     await Promise.all([statePathP, cachePathP, ephemeralStatePathP]);
   };
 
-  const isRootInitialized = async () => {
-    const noncePath = filePowers.joinPath(locator.statePath, 'nonce');
-    const nonce = await filePowers.maybeReadFileText(noncePath);
-    return nonce !== undefined;
-  };
-
   const provideRootNonce = async () => {
     const noncePath = filePowers.joinPath(locator.statePath, 'nonce');
     let nonce = await filePowers.maybeReadFileText(noncePath);
+    const isNewlyCreated = nonce === undefined;
     if (nonce === undefined) {
       nonce = await cryptoPowers.randomHex512();
       await filePowers.writeFileText(noncePath, `${nonce}\n`);
     }
-    return nonce.trim();
+    const rootNonce = nonce.trim();
+    return { value: rootNonce, isNewlyCreated };
   };
 
   const makeContentSha512Store = () => {
@@ -642,7 +638,6 @@ export const makeDaemonicPersistencePowers = (
   return harden({
     initializePersistence,
     provideRootNonce,
-    isRootInitialized,
     makeContentSha512Store,
     readFormula,
     writeFormula,
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -467,6 +467,11 @@ const makeDaemonCore = async (
         },
       });
       return { external: leastAuthority, internal: undefined };
+    } else if (formula.type === 'pet-inspector') {
+      // Behold, unavoidable forward-reference:
+      // eslint-disable-next-line no-use-before-define
+      const external = makePetStoreInspector(formula.petStore);
+      return { external, internal: undefined };
     } else {
       throw new TypeError(`Invalid formula: ${q(formula)}`);
     }
@@ -483,14 +488,7 @@ const makeDaemonCore = async (
     context,
   ) => {
     const formulaIdentifier = `${formulaType}:${formulaNumber}`;
-    if (formulaType === 'pet-inspector') {
-      const storeFormulaNumber = derive(formulaNumber, 'pet-store');
-      const storeFormulaIdentifier = `pet-store:${storeFormulaNumber}`;
-      // Behold, unavoidable forward-reference:
-      // eslint-disable-next-line no-use-before-define
-      const external = makePetStoreInspector(storeFormulaIdentifier);
-      return { external, internal: undefined };
-    } else if (formulaType === 'pet-store') {
+    if (formulaType === 'pet-store') {
       const external = petStorePowers.makeIdentifiedPetStore(
         formulaNumber,
         assertPetName,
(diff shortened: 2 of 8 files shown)


kumavis:
once witnessed an unreproduceable EPIPE in CI :thinking:

kumavis:
EPIPE seen on windows CI https://github.com/endojs/endo/actions/runs/7955058144/job/21713421990?pr=2089

packages/daemon/src/daemon-node-powers.js
```diff
@@ -504,6 +504,12 @@ export const makeDaemonicPersistencePowers = (
     await Promise.all([statePathP, cachePathP, ephemeralStatePathP]);
   };
 
+  const isRootInitialized = async () => {
+    const noncePath = filePowers.joinPath(locator.statePath, 'nonce');
+    const nonce = await filePowers.maybeReadFileText(noncePath);
+    return nonce !== undefined;
+  };
+
```
rekmarks, line 512: We could avoid attempting to read the root nonce file twice by merging this into `provideRootNonce` by making the latter return `{ value: string, isNewlyCreated: boolean }` (alternatively `wasRestored: boolean`).

packages/daemon/src/types.d.ts
```diff
@@ -385,6 +448,15 @@ export type EndoWebBundle = {
   powers: ERef<unknown>;
 };
 
+export type EndoBootstrap = FarRef<{
```
rekmarks, line 451: Although there is no `EndoBootstrap`, let's do this for consistency with `FarEndoReadable`?
```suggestion
export type FarEndoBootstrap = FarRef<{
```

packages/daemon/src/types.d.ts
```diff
@@ -188,6 +241,15 @@ export type ProvideValueForFormulaIdentifier = (
...
   formulaIdentifier: string,
 ) => Controller;
+export type ProvideControllerForFormulaIdentifierAndResolveHandle = (
+  formulaIdentifier: string,
+) => Promise<Controller>;
+
+export interface Handle {}
+export interface InternalHandle {
+  targetFormulaIdentifier: string;
+}
```
rekmarks, line 251: We should document what handles are somewhere, and docstrings on these interfaces is as good a place as any. The distinction between `Handle` and `InternalHandle` and when to use one vs. the other would also be good.

packages/daemon/src/mail.js
```diff
@@ -486,20 +467,18 @@ export const makeMailboxMaker = ({
...
-          await provideControllerForFormulaIdentifier(
-            recipientFormulaIdentifier,
-          )
+        await provideControllerForFormulaIdentifierAndResolveHandle(
+          recipientFormulaIdentifier,
         );
-
       const recipientInternal = await recipientController.internal;
-      if (recipientInternal === undefined) {
+      if (recipientInternal === undefined || recipientInternal === null) {
```
rekmarks, line 474: Why can this be `null` and not just `undefined`?

packages/daemon/src/guest.js
```diff
@@ -4,40 +4,36 @@ import { Far } from '@endo/far';
...
-      throw new Error(
-        `panic: a host request function must exist for every host`,
+    const hostController =
+      await provideControllerForFormulaIdentifierAndResolveHandle(
+        hostHandleFormulaIdentifier,
       );
-    }
-    const { respond: deliverToHost } = hostPrivateFacet;
+    const hostPrivateFacet = await hostController.internal;
+    const { respond: deliverToHost } = /** @type {any} */ (hostPrivateFacet);
```
rekmarks, line 36: Should this be a `ts-expect-error` instead?

packages/daemon/src/guest.js
```diff
@@ -67,7 +63,7 @@ export const makeGuestMaker = ({
       selfFormulaIdentifier: guestFormulaIdentifier,
       specialNames: {
         SELF: guestFormulaIdentifier,
-        HOST: hostFormulaIdentifier,
+        HOST: hostHandleFormulaIdentifier,
```
rekmarks, line 66: We should add a test verifying that the guest actually can't do anything as the host now that this change is made.

packages/daemon/src/types.d.ts
```diff
@@ -57,9 +57,35 @@ type FormulaIdentifierRecord = {
   number: string;
 };
 
+type EndoFormula = {
+  type: 'endo';
+  host: string;
+  leastAuthority: string;
+  webPageJs?: string;
```
rekmarks, line 64: It would be neat to document what the `string` properties on all of our formulas are supposed to be. Names? Identifiers if the real thing? Identifiers of handles?

packages/daemon/src/daemon.js
```diff
@@ -404,6 +396,82 @@ const makeEndoBootstrap = async (
...
+      return {
+        external: {},
+        internal: {
+          targetFormulaIdentifier: formula.target,
+        },
+      };
+    } else if (formula.type === 'endo') {
+      /** @type {import('./types.js').EndoBootstrap} */
+      const endoBootstrap = Far('Endo private facet', {
+        // TODO for user named
```
rekmarks, line 410: Qué?

packages/daemon/src/daemon.js
```diff
@@ -420,84 +488,23 @@ const makeEndoBootstrap = async (
...
         'make-bundle',
+        'host',
         'guest',
+        'least-authority',
         'web-bundle',
+        'web-page-js',
+        'handle',
+        'pet-inspector',
+        'pet-store',
+        'lookup',
```
rekmarks, line 507: There's an outdated comment ~20 lines below this that begins `// The two functions provideValueForFormula and provideValueForFormulaIdentifier`. It should be updated to reflect the removal of `provideValueForFormula`.

packages/daemon/src/daemon.js
```diff
@@ -624,33 +622,383 @@ const makeEndoBootstrap = async (
...
+      // eslint-disable-next-line no-constant-condition
+      while (true) {
+        const controller = provideControllerForFormulaIdentifier(
+          currentFormulaIdentifier,
+        );
+        // eslint-disable-next-line no-await-in-loop
+        const internalFacet = await controller.internal;
+        if (internalFacet === undefined || internalFacet === null) {
+          return controller;
+        }
```
rekmarks, line 638: Why can this be `null` and not just `undefined`?

rekmarks (changes requested):
Looks promising!

packages/daemon/src/mail.js
kumavis: just appeasing type checker

packages/daemon/src/guest.js
kumavis: good q, idk

packages/daemon/src/types.d.ts
kumavis: how to make subtypes of string? so you can't give it a string unless you cast it to a formulaId? or not a good idea?

packages/daemon/src/daemon.js
kumavis: idk it was already there

rekmarks:
One more question: what's the difference between incarnating something and reifying it? We should perhaps document this.

kumavis:
> One more question: what's the difference between incarnating something and reifying it? We should perhaps document this.

this PR introduces "incarnate" as a verb to mean to create a new thing for the first time. it follows that we can use "reincarnate" to reify things. fun. But would not be surprised if we land on different language.

rekmarks (approved):
LGTM!

packages/daemon/src/types.d.ts
rekmarks: Ah, I don't think it's worth doing anything with [template literal types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) here (especially in light of #2088). I just meant documenting what they're supposed to be on the interfaces would be nice.

I think we're OK for now since that information is there when the formulas are instantiated, so I won't block on it.

kumavis:
thanks for the improvements/

packages/daemon/src/daemon.js
kriskowal: The comment can be deleted. It’s a note to myself to consider adding the user’s name to the string above in the first argument of `Far` but that itself raises the question: “what name?”. It could be the first name assigned, but that raises the question: “by whom?”. The first agent to give them a name? Not sure it’s a good idea anyway.

packages/daemon/src/daemon.js
```diff
@@ -404,6 +411,82 @@ const makeEndoBootstrap = async (
...
+        },
+      });
+      return { external: leastAuthority, internal: undefined };
+    } else if (formula.type === 'pet-store') {
+      const external = petStorePowers.makeIdentifiedPetStore(
+        formulaNumber,
+        assertPetName,
+      );
+      return { external, internal: undefined };
+    } else if (formula.type === 'pet-inspector') {
```
kriskowal, line 485: We could probably turn this would-be-switch block into a JavaScript map from type to behavior now.

Issue #2021, opened 2024-01-30
daemon: `eval`-mediated worker names do not resolve


rekmarks:
## Describe the bug

When pet-naming a worker that is returned from an `eval` formula, you cannot later use that name to specify a `--worker` to the daemon. Doing so will result in the following error: `Cannot deliver "evaluate" to target; typeof target is "undefined"`.

## Steps to reproduce
Using the `INFO` feature of #1916:
```text
> endo make counter.js --name counter
  Object [Alleged: Counter] {}
> endo eval 'E(INFO).lookup("counter")' INFO --name inspector
  Object [Alleged: Inspector (make-bundle ... )]
> endo eval 'E(inspector).lookup("worker")' inspector --name cworker
  Object [Alleged: EndoWorker] {}
> endo eval 'E(counter).incr()' counter --worker cworker
  CapTP cli exception: (RemoteTypeError(error:captp:Endo#20001)#1)
  RemoteTypeError(error:captp:Endo#20001)#1: Cannot deliver "evaluate" to target; typeof target is "undefined"
```

## Expected behavior
As above except:
```text
...
> endo eval 'E(counter).incr()' counter --worker cworker
  1
```

kriskowal:
Design: we should go back to keeping an internal a WeakMap from worker handles to their corresponding bootstrap objects and stop using the internal facet trick. The WeakMap will work for workers even if laundered through an eval formula.

rekmarks:
I believe this is another instance of the same problem (originally reported in #2074):

### Steps to reproduce
```
> endo make counter.js --name counter

> endo eval 'E(counter).incr()' counter
1

> endo eval 'E(counter).incr()' counter
2

> endo eval 'foo' foo:INFO.counter.worker -n main-worker
Object [Alleged: EndoWorker] {}

// This should cancel the worker, but actually cancels a lookup formula instead.
// The bug is that this should, but does not, propagate to the worker (and then the counter).
> endo cancel main-worker

> endo eval 'E(counter).incr()' counter
3
```

### Expected Behavior
```
> endo make counter.js --name counter

> endo eval 'E(counter).incr()' counter
1

> endo eval 'E(counter).incr()' counter
2

> endo eval 'foo' foo:INFO.counter.worker -n main-worker
Object [Alleged: EndoWorker] {}

> endo cancel main-worker

> endo eval 'E(counter).incr()' counter
1
```

rekmarks:
[...]

Issue #2086, opened 2024-02-17
daemon: Formula graph must not be mutated concurrently


rekmarks:
Value incarnations, removals, and cancellations (collectively, "formula graph mutations") must be synchronized per #2092. This will necessitate refactoring a number of methods in `host.js` and (to some extent) `mail.js` that use these methods. Wherever possible, we should replace existing `incarnateFoo()` methods with their `incarnateNumberedFoo()` equivalents (again see #2092 for details).

- [x] `host.js`: `evaluate` #2092
- [x] `mail.js`: `cancel()` #2079 
- [x] `host.js`: `makeGuest()` #2120 
  - `host.js`: `makeNewHandleForSelf()`
  - `daemon.js`: `incarnateGuest()`
  - `daemon.js`: `incarnateHandle()`
  - `daemon.js`: `incarnateWorker()`
  - `daemon.js`: `incarnatePetStore()` 
- [x] `host.js`: `makeWorker()` #2123 
- [x] `host.js`: `makeUnconfined()` #2124 
  - `daemon.js`: `incarnateUnconfined()`
  - `host.js`: `provideWorkerFormulaIdentifier()`
- [x] `host.js`: `makeBundle()` #2126 
  - `daemon.js`: `incarnateBundle()`
- [x] `host.js`: `makeHost()` #2138 
  - `daemon.js`: `incarnateHost()`
  - `daemon.js`: `incarnatePetInspector()`
- [x] `host.js`: `provideWorker()` #2139 
- [x] `daemon.js`: `incarnateLeastAuthority()` #2141 
- [x] `daemon.js`: `incarnateReadableBlob()` #2142
- [x] `host.js`: `provideWebPage()` #2148 
  - `daemon.js`: `incarnateBundler()`
  - `daemon.js`: `incarnateWebBundle()`
- [ ] `daemon.js`: `incarnateEndoBootstrap()` #2149 

## What is the Problem Being Solved?

In the course of implementing #2079, it was discovered that two commands delivered serially to the daemon were fulfilled out of order, leading to unexpected behavior. In particular, the final `await` in this test code hangs indefinitely:

```javascript
await E(host).makeUnconfined(
  'worker',
  capletLocation,
  'NONE',
  'context-consumer',
);

// `result` is a promise that resolves with 'cancelled' once the
// caplet is cancelled.
const result = E(host).evaluate(
  'worker',
  'E(caplet).awaitCancellation()',
  ['caplet'],
  ['context-consumer'],
);
[...]

Issue #2074, opened 2024-02-14
`endo cancel` does not propagate properly under some circumstances


rekmarks:
## Describe the bug

As of the current `endo` feature branch (#2073), `endo cancel` does not propagate as expected under some cirumstances. See reproduction steps for two such cases. These cases may be ultimately the same, or distinct.

### Case 1
```
> endo purge -f

> endo make counter.js --name counter
Object [Alleged: Counter] {}

> endo mkguest doubler-agent
Object [Alleged: EndoGuest] {}

> endo make doubler.js -n doubler -p doubler-agent
Object [Alleged: Doubler] {}

> endo eval 'E(doubler).incr()' doubler
2

> endo eval 'E(doubler).incr()' doubler
4

> endo eval 'E(doubler).incr()' doubler
6

> endo kill counter

> endo eval 'E(doubler).incr()' doubler
8

> endo eval 'E(counter).incr()' counter
1

> endo eval 'E(counter).incr()' counter
2

> endo eval 'E(doubler).incr()' doubler
10

> endo kill doubler

> endo eval 'E(doubler).incr()' doubler
12

> endo kill doubler-agent

> endo eval 'E(doubler).incr()' doubler
6
```

### Case 2
```
> endo make counter.js --name counter

> endo eval 'E(counter).incr()' counter
1

> endo eval 'E(counter).incr()' counter
2

> endo eval 'foo' foo:INFO.counter.worker -n main-worker
Object [Alleged: EndoWorker] {}

> endo kill main-worker

> endo eval 'E(counter).incr()' counter
3
```
### Case 1
```
> endo purge -f

> endo make counter.js --name counter
Object [Alleged: Counter] {}

> endo mkguest doubler-agent
Object [Alleged: EndoGuest] {}

> endo make doubler.js -n doubler -p doubler-agent
Object [Alleged: Doubler] {}

> endo eval 'E(doubler).incr()' doubler
2

> endo eval 'E(doubler).incr()' doubler
4

> endo eval 'E(doubler).incr()' doubler
6

> endo kill counter

> endo eval 'E(doubler).incr()' doubler
2

> endo eval 'E(counter).incr()' counter
2

> endo eval 'E(counter).incr()' counter
3

> endo eval 'E(doubler).incr()' doubler
8

> endo kill doubler

> endo eval 'E(doubler).incr()' doubler
10

> endo kill doubler-agent

> endo eval 'E(doubler).incr()' doubler
12
```

### Case 2
```
[...]

PR #2092, opened 2024-02-21
refactor(daemon): Synchronize host `evaluate` method

rekmarks:
Progresses: #2086 
Fixes: #2021

Synchronizes the host's `evaluate()` method by delegating all incarnations to the daemon via `incarnateEval()`. The latter is responsible for incarnating its dependents as necessary, and for the generation of their formula numbers. To facilitate this, the synchronous methods `incarnateLookupSync()` and `incarnateWorkerSync()` have been added. These methods synchronously mutate the formula graph, and return promises that resolve when the formulas have been written to disk. The result is that the formula graph is mutated within a single turn of the event loop.

To achieve this, the implementation introduces new constraints on the daemon and its dependents. https://github.com/endojs/endo/pull/2089 introduced the notion of "incarnating" values. By the current definition, incarnating a value consists of the following steps:
1. Collecting dependencies (async)
    1. Generating requisite formula numbers (async)
        - We use the asynchronous signature of `crypto.randomBytes` to do
        this for performance reasons.
    2. Incarnating any dependent values (recursion!)
2. Updating the in-memory formula graph (sync)
3. Writing the resulting formula to disk (async)
4. Reifiying the resulting value (async)

In order to make formula graph mutations mutually exclusive, we introduce a "formula graph mutex" under which step 1 must be performed. This mutex is currently only used by `incarnateEval()`, and must be expanded to its sibling methods in the future.

`incarnateEval()` also introduces the notion of "incarnation hooks" to the codebase. Originators of incarnations that are exogenous to the daemon may themselves have asynchronous work perform. For example, `petName -> formulaIdentifier` mappings are the responsibility of the host and its pet store, and pet names must be associated with their respective formula identifiers the moment that those identifiers are observable to the consumer. To handle sych asynchronous side effects, the implementation introduces a notion of "hooks" to `incarnateEval()`, with the intention of spreading this to other incarnation methods as necessary. These hooks receive as an argument all formula identifiers created by the incarnation, and are executed under the formula graph mutex. This will surface IO errors to the consumer, and help us uphold the principle of "death before confusion".

Also of note, `provideValueForNumberedFormula` has been modified such that the formula is written to disk _after_ the controller has been constructed. This is critical in order to synchronize formula graph mutations.

Finally, it appears that the implementation incidentally fixed https://github.com/endojs/endo/issues/2021. We may still wish to adopt the more robust solution proposed in that issue.

### DIFF
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -49,7 +49,6 @@ const makeInspector = (type, number, record) =>
  * @param {import('./types.js').DaemonicPowers} powers
  * @param {Promise<number>} webletPortP
  * @param {object} args
- * @param {Promise<never>} args.cancelled
  * @param {(error: Error) => void} args.cancel
  * @param {number} args.gracePeriodMs
  * @param {Promise<never>} args.gracePeriodElapsed
@@ -57,7 +56,7 @@ const makeInspector = (type, number, record) =>
 const makeDaemonCore = async (
   powers,
   webletPortP,
-  { cancelled, cancel, gracePeriodMs, gracePeriodElapsed },
+  { cancel, gracePeriodMs, gracePeriodElapsed },
 ) => {
   const {
     crypto: cryptoPowers,
@@ -336,7 +335,7 @@ const makeDaemonCore = async (
    * @param {import('./types.js').Formula} formula
    * @param {import('./types.js').Context} context
    */
-  const makeControllerForFormula = async (
+  const makeControllerForFormula = (
     formulaIdentifier,
     formulaNumber,
     formula,
@@ -622,15 +621,16 @@ const makeDaemonCore = async (
   };
 
   /** @type {import('./types.js').ProvideValueForFormulaIdentifier} */
-  const provideValueForFormulaIdentifier = async formulaIdentifier => {
+  const provideValueForFormulaIdentifier = formulaIdentifier => {
     const controller = /** @type {import('./types.js').Controller<>} */ (
       provideControllerForFormulaIdentifier(formulaIdentifier)
     );
-    const value = await controller.external;
-    if (typeof value === 'object' && value !== null) {
-      formulaIdentifierForRef.set(value, formulaIdentifier);
-    }
-    return value;
+    return controller.external.then(value => {
+      if (typeof value === 'object' && value !== null) {
+        formulaIdentifierForRef.set(value, formulaIdentifier);
+      }
+      return value;
+    });
   };
 
   /** @type {import('./types.js').ProvideControllerForFormulaIdentifierAndResolveHandle} */
@@ -1010,7 +1010,6 @@ const makeDaemonCore = async (
     incarnateWebBundle,
     incarnateHandle,
     storeReaderRef,
-    randomHex512,
     makeMailbox,
   });
 
@@ -1159,7 +1158,6 @@ const makeDaemonCore = async (
  * @param {import('./types.js').DaemonicPowers} powers
  * @param {Promise<number>} webletPortP
  * @param {object} args
- * @param {Promise<never>} args.cancelled
  * @param {(error: Error) => void} args.cancel
  * @param {number} args.gracePeriodMs
  * @param {Promise<never>} args.gracePeriodElapsed
@@ -1168,12 +1166,11 @@ const makeDaemonCore = async (
 const provideEndoBootstrap = async (
   powers,
   webletPortP,
-  { cancelled, cancel, gracePeriodMs, gracePeriodElapsed },
+  { cancel, gracePeriodMs, gracePeriodElapsed },
 ) => {
   const { persistence: persistencePowers } = powers;
 
   const daemonCore = await makeDaemonCore(powers, webletPortP, {
-    cancelled,
     cancel,
     gracePeriodMs,
     gracePeriodElapsed,
@@ -1228,7 +1225,6 @@ export const makeDaemon = async (powers, daemonLabel, cancel, cancelled) => {
     powers,
     assignedWebletPortP,
     {
-      cancelled,
       cancel,
       gracePeriodMs,
       gracePeriodElapsed,
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -551,13 +551,10 @@ const makeDaemonCore = async (
 
     // Memoize for lookup.
     console.log(`Making ${formulaIdentifier}`);
-    const {
-      promise: partial,
-      resolve: resolvePartial,
-      reject: rejectPartial,
-    } = /** @type {import('@endo/promise-kit').PromiseKit<import('./types.js').InternalExternal<>>} */ (
-      makePromiseKit()
-    );
+    const { promise: partial, resolve: resolvePartial } =
+      /** @type {import('@endo/promise-kit').PromiseKit<import('./types.js').InternalExternal<>>} */ (
+        makePromiseKit()
+      );
 
     // Behold, recursion:
     // eslint-disable-next-line no-use-before-define
@@ -575,7 +572,7 @@ const makeDaemonCore = async (
     });
     controllerForFormulaIdentifier.set(formulaIdentifier, controller);
 
-    // We _must not_ await before the controller value is constructed.
+    // The controller _must_ be constructed in the synchronous prelude of this function.
     const controllerValue = makeControllerForFormula(
       formulaIdentifier,
       formulaNumber,
@@ -583,14 +580,16 @@ const makeDaemonCore = async (
       context,
     );
 
-    try {
-      await persistencePowers.writeFormula(formula, formulaType, formulaNumber);
-    } catch (error) {
-      rejectPartial(error);
-      throw error;
-    }
+    // Ensure that failure to flush the formula to storage
+    // causes a rejection for both the controller and the incarnation value.
+    const written = persistencePowers.writeFormula(
+      formula,
+      formulaType,
+      formulaNumber,
+    );
+    resolvePartial(written.then(() => /** @type {any} */ (controllerValue)));
+    await written;
 
-    resolvePartial(controllerValue);
     return harden({
       formulaIdentifier,
       value: controller.external,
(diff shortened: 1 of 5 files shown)


packages/daemon/src/daemon.js
```diff
@@ -689,55 +704,94 @@ const makeDaemonCore = async (
...
   /**
+   * @param {string} [specifiedFormulaNumber] - The formula number of the pet store.
    * @returns {Promise<{ formulaIdentifier: string, value: import('./types').PetStore }>}
    */
-  const incarnatePetStore = async () => {
-    const formulaNumber = await randomHex512();
+  const incarnatePetStore = async specifiedFormulaNumber => {
+    const formulaNumber =
+      specifiedFormulaNumber ??
+      (await formulaGraphLock.enqueue(() => randomHex512()));
```
kumavis, line 713: do we need to enqueue the number generation?

rekmarks: tl;dr: Yes. I just updated the description of this PR, which explains why.

packages/daemon/src/daemon.js
```diff
@@ -775,19 +803,60 @@ const makeDaemonCore = async (
...
   ) => {
-    const formulaNumber = await randomHex512();
+    const {
+      workerFormulaIdentifier,
+      endowmentFormulaIdentifiers,
+      evalFormulaNumber,
+    } = await formulaGraphMutex.enqueue(async () => {
+      const ownFormulaNumber = await randomHex512();
+      const workerFormulaNumber = await (specifiedWorkerFormulaIdentifier ??
+        randomHex512());
```
rekmarks, line 829: @kriskowal, let me know if I'm violating the letter and/or spirit of the law with this (strictly speaking) _non-conditional_ `await`.

packages/daemon/src/daemon.js
```diff
@@ -775,19 +803,60 @@ const makeDaemonCore = async (
...
+   * @param {string} [specifiedWorkerFormulaIdentifier]
    * @returns {Promise<{ formulaIdentifier: string, value: unknown }>}
    */
   const incarnateEval = async (
-    workerFormulaIdentifier,
+    hostFormulaIdentifier,
     source,
     codeNames,
-    endowmentFormulaIdentifiers,
+    endowmentFormulaPointers,
```
kriskowal, line 818: What is a “pointer”?

packages/daemon/src/daemon.js
```diff
@@ -775,19 +803,60 @@ const makeDaemonCore = async (
...
+        ).formulaIdentifier,
+        endowmentFormulaIdentifiers: await Promise.all(
+          endowmentFormulaPointers.map(async formulaIdOrPath => {
+            if (typeof formulaIdOrPath === 'string') {
+              return formulaIdOrPath;
+            }
+            return (
+              /* eslint-disable no-use-before-define */
+              (
+                await incarnateLookupSync(
```
kriskowal, line 843: await sync is a nominal contradiction.

packages/daemon/src/daemon.js
```diff
@@ -797,26 +866,33 @@ const makeDaemonCore = async (
...
+   * @param {string} formulaNumber - The lookup formula's number.
+   * @param {string} hubFormulaIdentifier - The formula identifier of the naming
+   * hub to call `lookup` on. A "naming hub" is an objected with a variadic
+   * lookup method. It includes objects such as guests and hosts.
+   * @param {string[]} petNamePath - The pet name path to look up.
+   * @returns {Promise<{ formulaIdentifier: string, value: import('./types').EndoWorker }>}
    */
-  const incarnateLookup = async (hubFormulaIdentifier, petNamePath) => {
-    const formulaNumber = await randomHex512();
+  const incarnateLookupSync = (
```
kriskowal, line 883: I see. Elsewhere I used the naming trick like “incarnate lookup” vs “incarnate numbered lookup” to distinguish whether the number was passed forward or internally constructed.

packages/daemon/src/host.js
```diff
@@ -282,23 +313,26 @@ export const makeHostMaker = ({
             return formulaIdentifier;
           }
 
-          const { formulaIdentifier: lookupFormulaIdentifier } =
-            await provideLookupFormula(petNamePath);
-          return lookupFormulaIdentifier;
-        }),
+          // TODO:lookup Check if a formula already exists for the path. May have to be
```
kriskowal, line 316: Consider an issue link for TODOs.

packages/daemon/src/host.js
```diff
@@ -282,23 +313,26 @@ export const makeHostMaker = ({
...
+          // done in the daemon itself.
+          return petNamePath;
```
kriskowal, line 318: This function appears to either return a pet name path or a formula identifier. Is a “pointer” the union of pet name path and identifier? If so, how are they distinguished?

packages/daemon/test/test-endo.js
```diff
@@ -880,8 +880,8 @@ test('name and reuse inspector', async t => {
   await stop(locator);
 });
 
-// TODO: This test verifies existing behavior when pet-naming workers.
-// This behavior is undesirable. See: https://github.com/endojs/endo/issues/2021
+// This tests behavior that previously failed due to a bug.
+// See: https://github.com/endojs/endo/issues/2021
```
kriskowal, line 884:
```suggestion
// Regression test for: https://github.com/endojs/endo/issues/2021
```

kriskowal (commented):
This looks like it closes the particular bug. Does this synchronize all formula writes or is the intention to capture those in follow-ups?

packages/daemon/src/daemon.js
```diff
@@ -49,15 +50,14 @@ const makeInspector = (type, number, record) =>
  * @param {import('./types.js').DaemonicPowers} powers
  * @param {Promise<number>} webletPortP
  * @param {object} args
- * @param {Promise<never>} args.cancelled
```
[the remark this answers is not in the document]
rekmarks: This was unrelated cleanup; `cancelled` was never used.

packages/daemon/src/daemon.js
```diff
@@ -775,19 +803,60 @@ const makeDaemonCore = async (
...
+      const ownFormulaNumber = await randomHex512();
+      const workerFormulaNumber = await (specifiedWorkerFormulaIdentifier ??
+        randomHex512());
+
+      const identifiers = harden({
+        workerFormulaIdentifier: (
+          await incarnateWorkerSync(workerFormulaNumber)
+        ).formulaIdentifier,
+        endowmentFormulaIdentifiers: await Promise.all(
+          endowmentFormulaPointers.map(async formulaIdOrPath => {
```
kriskowal, line 836: I think `formulaIdOrPath` is a misnomer here. I recommend just using the name `petNamePath` and just fall through to `petName` if it’s a string. It should never be a nonce / formula identifier.

packages/daemon/src/daemon.js
```diff
@@ -622,15 +632,16 @@ const makeDaemonCore = async (
...
-    if (typeof value === 'object' && value !== null) {
-      formulaIdentifierForRef.set(value, formulaIdentifier);
-    }
-    return value;
+    return controller.external.then(value => {
+      if (typeof value === 'object' && value !== null) {
+        formulaIdentifierForRef.set(value, formulaIdentifier);
+      }
+      return value;
+    });
```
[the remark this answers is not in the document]
rekmarks:
```suggestion
    return controller.external.then(value => {
      // Release the value to the public only after ensuring
      // we can reverse-lookup its nonce.
      if (typeof value === 'object' && value !== null) {
        formulaIdentifierForRef.set(value, formulaIdentifier);
      }
      return value;
    });
```

packages/daemon/src/host.js
rekmarks: I'm getting rid of the comment, but did create #2098 because we should have an issue anyway.

packages/daemon/src/host.js
rekmarks: I'm renaming the property / array to `endowmentFormulaIdsOrPaths`. In this particular array, strings are formula identifiers and arrays are pet name paths. They are distinguished using `typeof value === 'string'`.

packages/daemon/src/daemon.js
rekmarks: See: https://github.com/endojs/endo/pull/2092/files#r1498644209

packages/daemon/src/daemon.js
rekmarks: I'm going with your suggestion of `incarnateNumberedXyz`.

packages/daemon/src/daemon.js
rekmarks: See: https://github.com/endojs/endo/pull/2092/files#r1498644209

rekmarks:
@kriskowal 
> Does this synchronize all formula writes or is the intention to capture those in follow-ups?

The latter.

rekmarks:
ATTN reviewers: [e285f4c](https://github.com/endojs/endo/pull/2092/commits/e285f4c0f8354fcda390f4f2837eac88f7b2595d)

> `incarnateEval()` was accidentally passing worker formula identifiers instead of formula numbers when it received a specified worker formula identifier. This caused `eval` formulas to always fail on Windows due to a `:` in the path.
> 
> Surprisingly, fixing this caused the "indirect termination" test to fail. This is likely just surfacing a deeper problem, see https://github.com/endojs/endo/issues/2074.

packages/daemon/src/host.js
kriskowal: Ah, the plural agreement is confusing, but there’s perhaps nothing for it. I would like to consider (in follow-up) whether we can normalize these to formula identifiers up front.

packages/daemon/src/daemon.js
```diff
@@ -797,26 +868,33 @@ const makeDaemonCore = async (
...
+   * @param {string} formulaNumber - The lookup formula's number.
+   * @param {string} hubFormulaIdentifier - The formula identifier of the naming
+   * hub to call `lookup` on. A "naming hub" is an objected with a variadic
+   * lookup method. It includes objects such as guests and hosts.
+   * @param {string[]} petNamePath - The pet name path to look up.
+   * @returns {Promise<{ formulaIdentifier: string, value: import('./types').EndoWorker }>}
    */
-  const incarnateLookup = async (hubFormulaIdentifier, petNamePath) => {
-    const formulaNumber = await randomHex512();
+  const incarnateNumberedLookup = (
```
kriskowal, line 885: This will be tricky to replicate if we go on as I suggest to make lookup formulas single-hop.

kriskowal (approved):

PR #2093, opened 2024-02-21
fix(errors): Restore support for SES 0.18.3 and prior

kriskowal:
## Description

Agoric’s chain bootstrap vat runs with a version of SES that predates the `bare` function. Consequently, any library including `@endo/errors` running in a contract bundle in that vat must be prepared to shim any features absent on the global `assert`.

### Security Considerations

I do not believe this comes with direct security implications, but the situation emphasizes the need for extra scrutiny for programs admitted thru governance to run in the Agoric bootstrap vat given that it runs on a fixed version of SES.

### Scaling Considerations

None.

None.

### Testing Considerations

This includes a test that fails before the fix and passes after the fix. To review this PR, rebase incrementally and run tests before and after the fix.

The `bare` method falls through to the behavior of `quote`. This is consistent with the behavior of programs were written before the addition of `bare`, so I expect them to converge if that is indeed important.

### Upgrade Considerations


packages/errors/package.json
```diff
@@ -43,6 +43,7 @@
     "@endo/ses-ava": "^1.1.1",
     "ava": "^5.3.0",
     "c8": "^7.14.0",
+    "ses0_18_3": "npm:ses@0.18.3",
```
erights, line 46: OMFG

packages/errors/test/test-ses0_18_3.js
```diff
@@ -0,0 +1,8 @@
+import test from 'ava';
+
+test('shim bare on ses 0.18.3', async t => {
+  // SES v0.18.3 predates the `bare` export.
+  await import('ses0_18_3');
```
erights, line 5: This is my first encounter with this dependency magic. Why does this need to be a dynamic import of the magic name rather than a static import? What happens if you just statically import it instead?

If you could statically import it, you could statically import the next one too, right?

packages/static-module-record/test/test-static-module-record.js
```diff
@@ -792,3 +792,12 @@ test('should handle package "immer" source', t => {
     setUseProxies: ['vn', true],
   });
 });
+
+// https://github.com/endojs/endo/issues/2094
```
erights, line 796: Weird!

Is this a drive-by, or is this new test in this PR for a deeper reason?

packages/errors/test/test-ses0_18_3.js
kriskowal: The initialization of `@endo/errors` throws an exception. I could move `ses` to a static import but `@endo/errors` has to be dynamic for the error to be catchable in the test.

packages/static-module-record/test/test-static-module-record.js
kriskowal: I’ve bundled this test in here to explain why I did not use this pattern in the fix. My first attempt was:

```js
const {
  // ...
  quote,
  bare = quote,
  // ...
} = globalThis.assert;

export {
  // ...
  quote,
  bare,
  quote as q,
  bare as b,
  // ...
};
```

kriskowal: Notably, I detected this bug because a pet daemon test tripped on it. It manifests as a deeply nested Babel “Can’t get property buildError of undefined” exception.

packages/errors/package.json
kriskowal: This is, incidentally, one of the tools I’ve intimated to @warner and @mhofman that we have available to us should it ever be necessary to depend on multiple versions of `xsnap`.

packages/errors/test/test-ses0_18_3.js
```diff
@@ -0,0 +1,8 @@
+import test from 'ava';
+
+test('shim bare on ses 0.18.3', async t => {
```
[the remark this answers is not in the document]
kriskowal: Addressed.

packages/errors/index.js
```diff
@@ -61,18 +60,23 @@ const assert = (value, optDetails, optErrorContructor) =>
   globalAssert(value, optDetails, optErrorContructor);
 Object.assign(assert, assertions);
 
+// The Agoric chain's bootstrap vat runs with a version of SES that predates
+// the addition of the 'bare' method, so we must fall back to quote behavior
+// for that environment. (2024)
```
gibson042, line 65: Is this 2024 a year reference?

packages/errors/package.json
erights: Thx. Resolving.

packages/errors/index.js
```diff
@@ -31,7 +31,6 @@ const missing = /** @type {const} */ ([
   'details',
   'Fail',
   'quote',
-  'bare',
   'makeAssert',
```
gibson042, line 34: I think this omission could be made more clearly intentional.
```suggestion
  // As of 2024-02, the Agoric chain's bootstrap vat runs with a version of SES that
  // predates addition of the 'bare' method, so we must tolerate its absence and fall
  // back to quote behavior in that environment (see below).
  // 'bare',
  'makeAssert',
```

packages/errors/test/test-ses0_18_3.js
erights: Makes sense. Resolving.

packages/static-module-record/test/test-static-module-record.js
erights: Ok. Resolving.

erights (approved):
If this were a long lived anomaly, I might desire a more accurate `bare` emulation. By I don't buy the "in perpetuity". We cannot ever get stuck needing to support an infinite number of prior versions, because some of those will have fatal problems.

So this PR LGTM, thanks!

packages/errors/index.js
```diff
@@ -61,18 +60,23 @@ const assert = (value, optDetails, optErrorContructor) =>
...
   // assertions
   assert,
   // related utilities that aren't assertions
-  bare,
+  bareOrQuote as bare,
   makeError,
   note,
   quote,
   redacted,
   throwRedacted,
```
gibson042, line 77: Alternative possibility:
```suggestion
  makeError,
  note,
  quote,
  bare = quote,
  redacted,
  throwRedacted,
```

gibson042 (approved):

packages/errors/index.js
kriskowal: As noted with the failing test, our module transform does not support this pattern. Per the convergence of great minds, this was the shape of my first attempt!

kriskowal:
> If this were a long lived anomaly, I might desire a more accurate `bare` emulation. By I don't buy the "in perpetuity". We cannot ever get stuck needing to support an infinite number of prior versions, because some of those will have fatal problems.

For sure. Hopefully, the unstoppable force can move the immovable object. I suspect the only way out of this situation is for the chain bootstrap vat to eval a final contract that bequeaths its authority to a replacement chain bootstrap vat. We would presumably prevent any further core evals to execute in the original chain bootstrap vat via governance.

packages/errors/index.js
mhofman: To clarify, we are planning on replacing the environment in which core evals run, and retire the bootstrap vat, at which point old agoric chains will be able to evaluate core evals without the need for this backwards compatibility. Technically replacement is not an upgrade (something something ship of Theseus)

PR #2095, opened 2024-02-21
refactor(daemon): Retain multiple formula numbers per value

rekmarks:
Replaces the `formulaIdentifierForRef` WeakMap with a WeakMap-backed multimap, enabling the maintenance of one-to-many relationships between values and formula identifiers.


packages/daemon/src/multimap.js
```diff
@@ -0,0 +1,35 @@
+/**
+ * @returns {import('./types.js').Multimap<WeakKey, any>}
+ */
+export const makeMultimap = () => {
```
kriskowal, line 4: `makeWeakMultimap`

kriskowal (approved):

PR #2097, opened 2024-02-21
style(pass-style): Reduce nesting depth for input validation

gibson042:
## Description

Reduce the nesting depth for input validation by defining `reject`s as tag functions rather than traditional wrappers.

### Security Considerations

n/a

### Scaling Considerations

Constructing details is already in the slow path exercised only with an explicit check function, and introducing rest/spread to the wrapper imposes minor additional cost (the `reject` call itself gets about 6% slower on XS and 20% slower on V8). We could probably do better by passing `reject` functions rather than constructing them at each level, but that would affect the public API so I'm not contemplating that here.

n/a

### Testing Considerations

n/a

n/a

### Upgrade Considerations

n/a

- [ ] ~~Includes `*BREAKING*:` in the commit message with migration instructions for any breaking change.~~
- [ ] ~~Updates `NEWS.md` for user-facing changes.~~

### DIFF
--- a/packages/pass-style/src/copyRecord.js
+++ b/packages/pass-style/src/copyRecord.js
@@ -18,27 +18,23 @@ export const CopyRecordHelper = harden({
   styleName: 'copyRecord',
 
   canBeValid: (candidate, check = undefined) => {
-    const reject = !!check && (details => check(false, details));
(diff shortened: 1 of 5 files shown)


kriskowal (approved):
Looks mechanical to me.

PR #2099, opened 2024-02-22
test(marshal,ses): Prefer test.skip over conditional early return

gibson042:
Ref #2042
Ref #550

## Description

Some missed improvements related to https://github.com/endojs/endo/pull/2042#pullrequestreview-1894338425 .


erights (approved):
LGTM, thanks.

That's a nice way of abstracting things. Would be worth keeping if the underlying problem were long term. But `AggregateError` has been standard for awhile. Please at least insert TODOs to remove all this extra indirection and simplify, once we no longer need to support pre-AggregateError platforms.

Issue #1467, opened 2023-01-25
ses-ava should not mix exports and side-effects


gibson042:
(split from #1235)

https://github.com/Agoric/agoric-sdk/pull/5774/

**Motivation**

Hacks around side-effects of prepare-test-env-ava such as https://github.com/endojs/endo/pull/1439#discussion_r1070430689

**Acceptance criteria**
- module imports either have side-effects OR bindings

turadg:
closing per https://github.com/endojs/endo/pull/2133#issuecomment-1992647735

Issue #891, opened 2021-09-14
Ses-ava console virtualization: report to t.log instead of console


kriskowal:
SES hides error stacks from callers and reveals them to the global console. SES-ava captures errors thrown in tests and reports them to the console to reveal their stacks, regardless of whether the test passes or fails, generating a great deal of decontextualized noise.

We would like to reveal the error to the test log buffer (t.log) instead. Since SES interacts thru the console API, we would need to virtualize the console for tests.

This console virtualization might be adaptable across other purposes.

dckc:
see also #731

dckc:
#701 is in this direction, yes?

kriskowal:
Yes, very much so.

kriskowal:
To get here, we still need to tease out the compartment shim from `ses`.

erights:
May be fixed by #2109

erights:
Hi @kriskowal , I said in #2109 that it would close this, and so co-assigned myself. But I don't understand your comments about needing compartment work, which #2109 does not do. If #2109 would not close this, please let me know what I'm missing.

kriskowal:
I misunderstood what was necessary to accomplish this. Please disregard.

Issue #647, opened 2021-03-29
Some ses-ava static `test` methods  don't work


erights:
Changing test-message-pattern.js to use ses-ava causes the `test.serial` at
https://github.com/Agoric/agoric-sdk/blob/201cd2d4ed41a41bb9b751a714f99fcd3f6bb303/packages/SwingSet/test/test-message-patterns.js#L123

to fail at
https://github.com/Agoric/agoric-sdk/pull/2740/checks?check_run_id=2211691217#step:7:220
```
  Uncaught exception in test/test-message-patterns.js

  test/test-message-patterns.js:123

   122:   }                                                     
   123:   test.serial('local patterns', testLocalPattern, name);
   124: }                                                       

  Error: Duplicate test title: local patterns
```

erights:
Also `test.after.always`. See https://github.com/Agoric/agoric-sdk/pull/2740/checks?check_run_id=2214861371#step:10:10

erights:
I'm listing methods in this issue as I encounter them in trying to switch to ses-ava.
If it is easier to fix our code to not use these methods then it is to get ses-ava to virtualize these correctly, feel free. I'd rather define ses-ava to implement a sensible subset of the ava API if possible.

FUDCo:
I'm very confused.  First, there's no `test.repeat` there.  Second, my change to `ses-ava` last week makes this work.  Something has gotten tangled up in the plumbing somewhere.

erights:
Doh! I remembered the wrong name. `test.serial` is the problem. I revised the first message in this thread to correct that.

FUDCo:
OK, that takes care of my first point, but why did this work correctly for me the other day?  The change I made to `ses-ava` should (did!) fix this.

erights:
It seems to be only those two uses of `test.serial`. Several others do fine with ses-ava. Some others have not yet been tried with ses-ava.

FUDCo:
That specific test file (`test-message-patterns.js`) was the thing I was going on about wanting to run under `ses-ava` and after my changes to `ses-ava` to support the Ava "macro" facilitiy, it worked.  So if it doesn't work now, something else you did changed something to make it not work.  I'm trying to figure out what that might have been.

erights:
The left pane of

https://github.com/Agoric/agoric-sdk/pull/2740/files#diff-8845f7584737c882a225a4b88eee42f5064626370ad4b222d3e0319fd6e4cb68

[...]

PR #2101, opened 2024-02-24
test(ses-ava): option to use t.log

erights:
Staged on https://github.com/endojs/endo/pull/2107

closes: #XXXX
refs: https://github.com/Agoric/agoric-sdk/pull/8965#discussion_r1501192561 https://github.com/endojs/endo/issues/611 https://github.com/endojs/endo/issues/647 https://github.com/endojs/endo/issues/891 https://github.com/endojs/endo/issues/1467 https://github.com/endojs/endo/pull/2107 https://github.com/Agoric/agoric-sdk/issues/8662 https://github.com/endojs/endo/issues/1798

## Description

Alter ses-ava so that it ultimately uses `t.log` to log errors or rejections from the `t => {...}` function.

### Security Considerations

As of this PR, ses-ava will import from `console-tools.js`, as exported by `ses` in https://github.com/endojs/endo/pull/2107 . We must ensure that this arrangement does not accidentally make these powers available to non-privileged code executing in constructed (non-start) compartments. See "Security Considerations" at https://github.com/endojs/endo/pull/2107

### Scaling Considerations

none

none

### Testing Considerations

The whole point. Since *this* PR changes only the logging of errors or rejections from the `t => {...}` function, it will not by itself clean up console output in general. But at least it will cause these error reports to appear at the "right" place in the overall console output.

Mostly none.

If there are programs post-processing our voluminous console output, they might need to change. However,
- I am not aware of any such programs
- the console output was never stable anyway, so any such programs should be adaptable enough to tolerate this change as well.

### Upgrade Considerations

Nothing breaking. Nothing newsworthy as of this PR by itself. 

### DIFF
--- a/packages/ses-ava/src/ses-ava-test.js
+++ b/packages/ses-ava/src/ses-ava-test.js
@@ -7,6 +7,19 @@ import {
   consoleOtherMethods,
 } from 'ses/console-tools.js';
 
+/**
+ * TODO For some reason, the following declaration (with "at-" as "@")
+ * doesn't work well for either TS or typedoc. For TS it seems to type
+ * `VirtualConsole` as `any` in a vscode hover. For typedoc it results in
+ * errors.
+ *
+ * at-typedef {import('ses/console-tools.js').VirtualConsole} VirtualConsole
+ *
+ * so instead, for now, we just declare it as `any`. TODO is to repair this.
+ *
+ * @typedef {any} VirtualConsole
+ */
+
 const { stringify } = JSON;
 const { defineProperty, freeze } = Object;
 const { apply } = Reflect;
@@ -50,7 +63,7 @@ const isPromise = maybePromise =>
  * @param {unknown[]} args
  * @param {string} source
  * @param {Logger} logger
- * @param {import('ses/console-tools.js').VirtualConsole} [optConsole]
+ * @param {VirtualConsole} [optConsole]
  */
 const logErrorFirst = (func, args, source, logger, optConsole = undefined) => {
   const originalConsole = globalThis.console;
@@ -98,7 +111,7 @@ const overrideList = [
  * are output in the right place.
  *
  * @param {Logger} tlogger
- * @returns {import('ses/console-tools.js').VirtualConsole}
+ * @returns {VirtualConsole}
  */
 const makeConsoleFromLogger = tlogger => {
   const baseConsole = {};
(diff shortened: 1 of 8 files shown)


packages/ses-ava/src/ses-ava-test.js
```diff
@@ -75,13 +75,29 @@ const overrideList = [
...
 /**
  * @template {import('ava').TestFn} T
  * @param {T} testerFunc
- * @param {Logger} logger
+ * @param {Logger | 'tlog'} loggerOrOption
  * @returns {T} Not yet frozen!
  */
-const augmentLogging = (testerFunc, logger) => {
+const augmentLogging = (testerFunc, loggerOrOption) => {
   const testerFuncName = `ava ${testerFunc.name || 'test'}`;
```
gibson042, line 101: I'd probably deal with the logger argument here, to validate early and simplify `wrapImplFunc` to `return logErrorFirst(fn, [t, ...args], source, logger || t.log)`.
```suggestion
  const testerFuncName = `ava ${testerFunc.name || 'test'}`;
  let logger;
  if (typeof loggerOrOption === 'function') {
    logger = loggerOrOption;
  } else if (loggerOrOption !== 'tlog') {
    throw Error(`Unrecognized loggerOrOption ${loggerOrOption}`);
  }
```

gibson042 (approved):
Nice; this is definitely an improvement!

packages/ses-ava/test/prepare-test-env-ava.js
```diff
@@ -4,4 +4,4 @@ import rawTest from 'ava';
 import { wrapTest } from '../src/ses-ava-test.js';
 
 /** @type {typeof rawTest} */
-export const test = wrapTest(rawTest);
+export const test = wrapTest(rawTest, 'tlog');
```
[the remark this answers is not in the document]
erights: Done, but seeking a descriptive name for the option

erights:
This is now actually ready for review.

Enough has changed since you approved that I request new approvals. (I have no idea how to clear the old ones, or even if that's possible.)

Note that this PR now depends on, and is staged on, #2107, so that would need to get reviewed and merged before this one could. Please review that one too. It should be straightforward.

packages/ses-ava/src/ses-ava-test.js
erights: @gibson042 , now that I've got the case where the causal console wraps on `t.log`, I no longer see the opportunity for this simplification. Please PTAL and let me know if I missed it. Thanks.

erights:
Converting to Draft. See https://github.com/endojs/endo/pull/2107#issuecomment-1972239076

erights:
Closing in favor of https://github.com/endojs/endo/pull/2109

erights:
#2109 is sufficiently different that it needs a from-scratch review.

packages/ses-ava/src/ses-ava-test.js
erights: Successor #2109 is no longer options driven. It provides one built in behavior. Thus, on behalf of #2109 , I'm resolving this conversation.

PR #2102, opened 2024-02-26
refactor(daemon,cli): Consolidate pet name list commands

kriskowal:
This change reverses a choice I made earlier, merged by #2068, where I elected to simplify the daemon’s list API to just have one `list()` method for each type of directory, name hub, or agent, guest, or host. I think if we want filters in the future, we can recover them in the CLI alone.


packages/cli/src/commands/list.js
```diff
@@ -3,16 +3,12 @@ import os from 'os';
 import { E } from '@endo/far';
 import { withEndoHost } from '../context.js';
 
-export const list = async ({ directoryName, special, all }) =>
+export const list = async ({ directoryPath }) =>
   withEndoHost({ os, process }, async ({ host: party }) => {
-    if (directoryName !== undefined) {
-      party = E(party).lookup(directoryName);
+    if (directoryPath !== undefined) {
+      party = E(party).lookup(...directoryPath.split('.'));
```
kriskowal, line 9: I tacked on this fix for dot delimited lookup at the CLI.

kumavis (approved):

PR #2106, opened 2024-02-28
docs(ses): Add section links

gibson042:
## Description

Add section links to the options summary table (placed at the end to minimize disruption when looking at the Markdown source).


erights (approved):
LGTM, thanks! Makes the page much more usable.

kriskowal (approved):

PR #2107, opened 2024-02-28
feat(ses): expose console tools support

erights:
closes: #XXXX
refs: https://github.com/endojs/endo/pull/2101

## Description

Exports `console-tools.js` from `ses`, so PRs like https://github.com/endojs/endo/pull/2101 can use these to manipulate consoles ***ONLY*** for testing and debugging purposes. Some of our other packages export testing support API with an export named "tools.js". But ses already exports a "tools.js" with a different meaning.

Other than arranging this new availability, this PR itself should have no observable change whatsoever. In that sense, we can consider it a pure refactoring.

### Security Considerations

The powers exported from `ses` as `console-tools.js` should only be available to privileged code, such as ses-ava and instrumentation that runs in the start compartment. But as an export from the "ses" package, can we prevent others from importing it within the way we're currently using our tooling? @kriskowal , I think this is a question mostly for you.

### Scaling Considerations

none

none

### Testing Considerations

The whole point is to enable testing tools like ses-ava to manipulate the console for better testing and debugging.

none

### Upgrade Considerations

none. Nothing breaking. Nothing newsworthy.

### DIFF
--- a/packages/ses/console-tools.js
+++ b/packages/ses/console-tools.js
@@ -1,0 +1,8 @@
+export { loggedErrorHandler } from './src/error/assert.js';
+export {
+  makeCausalConsole,
+  consoleWhitelist,
+  makeLoggingConsoleKit,
+  filterConsole,
+} from './src/error/console.js';
+export { assertLogs, throwsAndLogs } from './src/error/throws-and-logs.js';
--- a/packages/ses/console-tools.js
+++ b/packages/ses/console-tools.js
@@ -1,8 +1,10 @@
 export { loggedErrorHandler } from './src/error/assert.js';
 export {
   makeCausalConsole,
-  consoleWhitelist,
+  consoleLevelMethods,
+  consoleOtherMethods,
   makeLoggingConsoleKit,
   filterConsole,
+  pumpLogToConsole,
 } from './src/error/console.js';
 export { assertLogs, throwsAndLogs } from './src/error/throws-and-logs.js';
--- a/packages/ses/package.json
+++ b/packages/ses/package.json
@@ -46,6 +46,7 @@
       "require": "./dist/ses.cjs"
     },
     "./tools.js": "./tools.js",
+    "./console-tools.js": "./console-tools.js",
     "./assert-shim.js": "./assert-shim.js",
     "./lockdown-shim.js": "./lockdown-shim.js",
     "./compartment-shim.js": "./compartment-shim.js",
--- a/packages/ses/src/error/console.js
+++ b/packages/ses/src/error/console.js
@@ -46,8 +46,17 @@ import './internal-types.js';
 
 /** @typedef {keyof VirtualConsole | 'profile' | 'profileEnd'} ConsoleProps */
 
-/** @type {readonly [ConsoleProps, LogSeverity | undefined][]} */
-const consoleLevelMethods = freeze([
+/**
+ * Those console methods whose actual parameters are `(fmt?, ...args)`
+ * even if their TypeScript types claim otherwise.
+ *
+ * Each is paired with what we consider to be their log severity level.
+ * This is the same as the log severity of these on other
+ * platform console implementations when they all agree.
+ *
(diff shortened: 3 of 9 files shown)


kriskowal:
In advance of my full review, some things to consider when deciding whether to export a module or a global:

What is the intended outcome for the usage pattern we encourage on the web, where `ses` is not bundled but injected with an early `<script>`? Would the outcome be acceptable if the bundled application imports an eval twin from`ses/console-tools.js`? My intuition is that, for this feature, there is important shared state and an eval twin would not be desirable.

Is the intention to treat consumers that import as having a higher level of privilege than other code that runs in the original realm / start/initial compartment? My intuition is that imports and globals in the start compartment have equal privilege.

How does using a module for some facilities and globals for other facilities impact the narrative for standardization? Is the intention that anything imported by a module would not participate in the standard? I don’t think console tools are standards track, but they could be factored in such a way that the essential primitive (error unboxing) would be standard.

Is the intention of using a module to communicate a lower level of commitment to backward compatibility within the `^1` semver range? That’s not consistent with my view of our obligations. This is why I’m encouraging a refactoring of `@endo/permits` out of `ses`, so that `ses` can depend on any major version of `@endo/permits` internally without any impact on its own obligations for backward compatibility. That particular case is also not affected by eval twins and ideally would use “nested dependencies”, if such a feature existed.

erights:
> In advance of my full review, some things to consider when deciding whether to export a module or a global:
> 
> What is the intended outcome for the usage pattern we encourage on the web, where `ses` is not bundled but injected with an early `<script>`? Would the outcome be acceptable if the bundled application imports an eval twin from`ses/console-tools.js`? My intuition is that, for this feature, there is important shared state and an eval twin would not be desirable.

Correct. But the problem is not the reexports from `console.js`, which are rather stateless makers. The problem is the reexport of `loggedErrorHandler`, which is state maintained by the `assert.js` module on behalf of causal consoles. Remaining state (`errorsLogged`) is only in the console instance, not in the module.

We know we're already making assumptions about shared `assert` state, and using the global `assert` to coordinate. I see that there's a separate `assert-shim.js` that makes the global from the module. But I don't see anything preventing eval twin or `assert.js` or even `assert-shim.js`. If `assert-shim.js` is run a second time, it will overwrite the global. Is this what you intend?

So I think the current behavior means we already have an eval twin problem with the `assert.js` module. But maybe not because, prior to this PR, nothing from `assert.js` nor `assert-shim.js` is ever exported. Further, `@endo/errors reexports the global, not imports from the module.

Ok, I am confused and am placing this PR back into draft status until we discuss this and work it out.

> Is the intention to treat consumers [of] that import as having a higher level of privilege than other code that runs in the original realm / start/initial compartment? My intuition is that imports and globals in the start compartment have equal privilege.

Good! My intention is that only privileged code can import from `console-tools.js`. But any code running in the start compartment has enough privilege. For that code, the issue is separation to avoid accidental POLA violations, which this PR does.

> 
> How does using a module for some facilities and globals for other facilities impact the narrative for standardization? Is the intention that anything imported by a module would not participate in the standard? I don’t think console tools are standards track, but they could be factored in such a way that the essential primitive (error unboxing) would be standard.

I am not currently thinking of proposing the causal console as standard. I would not be upset if it needed major revisions before it could be proposed. The current purpose is only to support some shenanigans that I want ses-ava to do, in development at https://github.com/endojs/endo/pull/2101 . I certainly do not expect to propose ses-ava for standardization ;) .

> Is the intention of using a module to communicate a lower level of commitment to backward compatibility within the `^1` semver range? That’s not consistent with my view of our obligations. This is why I’m encouraging a refactoring of `@endo/permits` out of `ses`, so that `ses` can depend on any major version of `@endo/permits` internally without any impact on its own obligations for backward compatibility. That particular case is also not affected by eval twins and ideally would use “nested dependencies”, if such a feature existed.

I do not understand what the implications are for this PR. Let's remember to include this in our discussion.

erights:
Closing in favor of #2109

Issue #611, opened 2021-03-15
ses-ava support for t.log


dckc:
from https://github.com/Agoric/agoric-sdk/pull/2582#discussion_r594347491 ...

[our tests aren't supposed to console.log](https://github.com/Agoric/agoric-sdk/wiki/agoric-sdk-unit-testing#patterns). `t.log` would be silent unless `ava -v` were used...

```js
    t.log('expected failure', reason);
```

but it's not integrated with the SES console, so we get the boring...

```
  ✔ ses-ava-reject-deep-stacks › ses-ava reject console output
    ℹ expected failure TypeError {
        message: 'msg (a string)',
      }
```

erights:
The boringness should be fixed by https://github.com/endojs/endo/pull/2109

Whatever Ava's properties around quieting or verbosifying `t.log`, #2109 should preserve them. However, from my own testing of raw Ava I am not seeing that silence.

dckc:
The "silent unless -v" aspect of ava went away in some version a while ago, IIRC.

gibson042:
> The "silent unless -v" aspect of ava went away in some version a while ago, IIRC.

Yeah, looks like it was dropped as part of version 4: https://github.com/avajs/ava/commit/0a6e2df467f35256bbf9d855a4cc9ebda0646909

PR #701, opened 2021-04-28
WIP: Better errors from t.fail, t.notThrows, t.notThrowsAsync

erights:
DO NOT MERGE: Still based on an ancient master

Draft until demo-tests. But by using `yarn link` to link agoric-sdk to ses-ava, I see that it already improved the error reporting problem reported today by @Chris-Hibbert . This improvement is merely that these three methods also report the error to our console, where we have more control. What it not yet addressed is actually reporting deep stacks.

PR #2109, opened 2024-03-01
test(ses-ava): use causal console

erights:
closes: https://github.com/endojs/endo/issues/611 
closes: https://github.com/endojs/endo/issues/891
refs: https://github.com/Agoric/agoric-sdk/pull/8965#discussion_r1501192561 https://github.com/endojs/endo/issues/647 https://github.com/endojs/endo/issues/1467 https://github.com/endojs/endo/pull/2107 https://github.com/Agoric/agoric-sdk/issues/8662 https://github.com/endojs/endo/issues/1798 https://github.com/endojs/endo/pull/2101 https://github.com/endojs/endo/pull/2107 https://github.com/Agoric/agoric-sdk/wiki/agoric-sdk-unit-testing#patterns

## Description

Replaces https://github.com/endojs/endo/pull/2101 and https://github.com/endojs/endo/pull/2107 , though unclosed conversations there probably apply here as well.

Alter ses-ava so that 
- The exposed virtual `t.log` first goes through our causal console logic, to show redacted information including nesting errors, but still ultimately directs the output to the original `t.log` so it retains all the ordering and verbosity switching associated with Ava's original `t.log`.
- When a test fails by throwing or rejecting, the error is also logged with the virtual `t.log`, so you see its redacted info as well, while staying associated with the error.

To get access to the SES's causal console logic, this PR also adds a new export for a privileged power, which ses-ava needs and imports using 
```js
import { makeCausalConsoleFromLogger } from 'ses/console-tools.js';
```

https://github.com/endojs/endo/pull/2107 review comments raise the issue of whether such special powers should be "exported" from ses and "imported" by privileged clients vs `import`, as this PR currently does, or as a global binding of the start compartment. Unlike https://github.com/endojs/endo/pull/2101 and https://github.com/endojs/endo/pull/2107, this PR currently bundles the ses changes and the ses-ava changes together so we can coordinate changing our minds on this. Reviewers, once that settles, we can split this up into two PRs again if you wish.

### Security Considerations

As of this PR, ses-ava will import `makeCausalConsoleFromLogger` from `console-tools.js`. We must ensure that this arrangement does not accidentally make these powers available to non-privileged code executing in constructed (non-start) compartments. 

### Scaling Considerations

none

none

### Testing Considerations

The whole point. Since *this* PR changes only the logging of errors or rejections from the `t => {...}` function, it will not by itself clean up console output in general. But at least it will cause these error reports to appear at the "right" place in the overall console output.

In addition, the "export" of `makeCausalConsoleFromLogger` from `ses/console-tools.js` enables other testing tools to build on our causal console the way ses-ava does.

Mostly none.

If there are programs post-processing our voluminous console output, they might need to change. However,
- I am not aware of any such programs
- the console output was never stable anyway, so any such programs should be adaptable enough to tolerate this change as well.

### Upgrade Considerations

Nothing breaking.


erights:
May want to redo https://github.com/endojs/endo/pull/701#issuecomment-1974066710 as part of this PR. It would be a natural fit.

packages/ses/console-tools.js
```diff
@@ -0,0 +1,21 @@
...
+//   filterConsole,
+//   pumpLogToConsole,
+// } from './src/error/console.js';
+// export { assertLogs, throwsAndLogs } from './src/error/throws-and-logs.js';
+
+/**
+ * `makeCausalConsoleFromLogger` is privileged because it exposes
+ * unredacted error info onto the `Logger` provided by the caller.
+ */
+export const makeCausalConsoleFromLogger =
```
turadg, line 20: what does "causal" mean in this context? I can't find it explained in the repo. 

Also it would help for this export to explain why one would use it.

packages/ses/src/error/console.js
```diff
@@ -362,12 +394,51 @@ const makeCausalConsole = (baseConsole, loggedErrorHandler) => {
...
 
-// /////////////////////////////////////////////////////////////////////////////
+/**
+ * @typedef {(...args: unknown[]) => void} Logger
+ */
+
+/**
+ * @param {LoggedErrorHandler} loggedErrorHandler
+ */
+export const makeCausalConsoleFromLoggerMaker = loggedErrorHandler => {
```
turadg, line 405: I think we've tried to avoid the "make…Maker" naming. @gibson042 had a better scheme though I don't recall whether it was written up. With Exos a function that produces a `make` is called a `define`.
```suggestion
export const defineCausalConsoleFromLogger = loggedErrorHandler => {
```

erights: Done

gibson042: > I think we've tried to avoid the "make…Maker" naming. @gibson042 had a better scheme though I don't recall whether it was written up. With Exos a function that produces a `make` is called a `define`.

I don't remember the details, but I think that resulted in https://github.com/Agoric/agoric-sdk/pull/5234 .

packages/ses/console-tools.js
erights: Added explanations to ses/error/README.md, which already said a lot about it, but had not introduced the "causal console" terminology to name it.

Added a brief explanation here and a link to that README, where the relevant explanation will appear one this PR is merged.

Done. PTAL.

packages/ses/src/error/console.js
erights: It's also used throughout the exo class making functions, and even the earlier deprecated virtual kind making functions.

packages/ses/src/error/console.js
```diff
@@ -362,12 +394,51 @@ const makeCausalConsole = (baseConsole, loggedErrorHandler) => {
...
+    for (const [name1, _] of consoleLevelMethods) {
+      baseConsole[name1] = tlogger;
+    }
+    for (const [name2, _] of [
+      ['group', 'log'],
+      ['groupCollapsed', 'log'],
+      ...consoleOtherMethods,
+    ]) {
+      baseConsole[name2] = (...args) => tlogger(name2, ...args);
+    }
```
gibson042, line 426: Suggestion: add indentation rather than "groupEnd" entries.
```suggestion
    const indents = [];
    const logWithIndent = (...args) => {
      if (indents.length > 0) {
        args = [...indents, ...args];
      }
      return tlogger(...args);
    };
    const makeNamed = (name, fn) =>
      ({ [name]: (...args) => fn(...args) })[name];
  
    const baseConsole = fromEntries([
      ...arrayMap(consoleLevelMethods, ([name]) => [
        name,
        makeNamed(name, logWithIndent),
      ]),
      ...arrayMap(consoleOtherMethods, ([name]) => [
        name,
        makeNamed(name, (...args) => logWithIndent(name, ...args)),
      ]),
    ]);
    // https://console.spec.whatwg.org/#grouping
    for (const name of ['group', 'groupCollapsed', 'groupEnd']) {
      if (!baseConsole[name]) continue;
      baseConsole[name] = makeNamed(name, (...args) => {
        if (name === 'groupEnd') {
          indents.pop();
          return;
        }
        // Prefix the logged data with "group" or "groupCollapsed".
        logWithIndent(name, ...args);
        // A single space is enough;
        // the host console will separate them with additional spaces.
        indents.push(' ');
      });
    }

```

gibson042 (commented):
Very nice!

packages/eventual-send/test/test-deep-stacks-ses-ava.js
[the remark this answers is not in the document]
gibson042: (and likewise for packages/eventual-send/test/test-deep-send-ses-ava.js)

erights:
Apologies to my reviewers, but in responding to @kriskowal 's request about how ses makes the special console-making-power available to `@endo/ses-ava` at this time, I went about making the change in a way that squashed. I hope this does not make further review too difficult. I have not yet responded to the reviewer comments still open above, and will do so from this baseline.

packages/ses/index.js
```diff
@@ -15,3 +15,4 @@
 import './src/lockdown-shim.js';
 import './src/compartment-shim.js';
 import './src/assert-shim.js';
+import './src/console-shim.js';
```
kriskowal, line 18: We may eventually need an entry in `package.json` and a top-level `./console-shim.js`, but not today.

packages/ses-ava/src/ses-ava-test.js
```diff
@@ -1,16 +1,59 @@
...
+  globalThis[MAKE_CAUSAL_CONSOLE_FROM_LOGGER_KEY_FOR_SES_AVA];
 
 /**
- * @typedef {(...args: unknown[]) => void} Logger
+ * TODO For some reason, the following declaration (with "at-" as "@")
+ * doesn't work well for either TS or typedoc. For TS it seems to type
+ * `VirtualConsole` as `any` in a vscode hover. For typedoc it results in
+ * errors.
+ *
+ * at-typedef {import('ses/console-tools.js').VirtualConsole} VirtualConsole
```
kriskowal, line 41: This type reference error may be because of the rename to `ses/src/console-shim.js`. We may need to export `ses/console-shim.js` in `package.json` to make this internal type addressable. We should avoid reaching into `src` from other packages in general.

packages/ses-ava/src/ses-ava-test.js
```diff
@@ -76,12 +92,65 @@ const overrideList = [
...
- * @template {import('ava').TestFn} T
+ * @param {import('ava').ExecutionContext} originalT
+ * @returns {import('ava').ExecutionContext}
+ */
+const makeVirtualExecutionContext = originalT => {
+  const causalConsole = makeCausalConsoleFromLogger(originalT.log);
+  const virtualT = {
+    log: /** @type {import('ava').LogFn} */ (causalConsole.error),
+    console: causalConsole,
+    withConsole: thunk => {
```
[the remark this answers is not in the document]
kriskowal: I think we should omit `withConsole` until we encounter a concrete case that can be solved no other way. Or, just wait for `AsyncContext` to give us better options.

kriskowal (commented):
I’m satisfied by this coupling at a global symbol between `ses` and `ses-ava` since this is a stopgap on a path and not where we want to land for error unsealing in general.

I defer to @gibson042 for the final stamp since he has pending feedback!

packages/ses/index.js
erights: Why? By "entry in `package.json`, do you mean in `"exports":`? If so, I'm surprised. I thought that's what you just talked me out of. Am I confused?

packages/ses-ava/src/ses-ava-test.js
erights: Is there an explanation anywhere of these type scoping issues? I remain perpetually confused about these, and trying to infer them by example has only dug me into a deeper hold of confusion ;)

packages/ses-ava/src/ses-ava-test.js
```diff
@@ -1,16 +1,59 @@
...
+ * Further, we consider this particular API choice to be experimental
+ * and may change in the future. It is currently only intended for use by
+ * `@endo/ses-ava`, with which it will be co-maintained.
+ *
+ * Thus, this `console-shim.js` makes `makeCausalConsoleFromLoggerForSesAva` for
+ * on `globalThis` which it *assumes* is the global of the start compartment,
+ * which is therefore allowed to hold powers that should not be available
+ * in constructed compartments. It makes it available as the value of a
+ * global property named by a registered symbol named
+ * `MAKE_CAUSAL_CONSOLE_FROM_LOGGER_KEY_FOR_SES_AVA`.
```
gibson042, line 22:
```suggestion
 * Thus, the SES console-shim.js makes `makeCausalConsoleFromLoggerForSesAva`
 * available on `globalThis` which it *assumes* is the global of the start
 * compartment and is therefore allowed to hold powers that should not be
 * available in constructed compartments. It makes it available as the value of
 * a global property named by a registered symbol named
 * `MAKE_CAUSAL_CONSOLE_FROM_LOGGER_KEY_FOR_SES_AVA`.
```

packages/ses/src/console-shim.js
```diff
@@ -0,0 +1,50 @@
...
+ * Further, we consider this particular API choice to be experimental
+ * and may change in the future. It is currently only intended for use by
+ * `@endo/ses-ava`, with which it will be co-maintained.
+ *
+ * Thus, this `console-shim.js` makes `makeCausalConsoleFromLoggerForSesAva` for
+ * on `globalThis` which it *assumes* is the global of the start compartment,
+ * which is therefore allowed to hold powers that should not be available
+ * in constructed compartments. It makes it available as the value of a
+ * global property named by a registered symbol named
+ * `MAKE_CAUSAL_CONSOLE_FROM_LOGGER_KEY_FOR_SES_AVA`.
```
gibson042, line 39:
```suggestion
 * Thus, this `console-shim.js` makes `makeCausalConsoleFromLoggerForSesAva`
 * available on `globalThis` which it *assumes* is the global of the start
 * compartment and is therefore allowed to hold powers that should not be
 * available in constructed compartments. It makes it available as the value of
 * a global property named by a registered symbol named
 * `MAKE_CAUSAL_CONSOLE_FROM_LOGGER_KEY_FOR_SES_AVA`.
```

dckc (commented):
I gather @gibson042 and @kriskowal have this review well in hand.

packages/eventual-send/test/test-deep-send-ses-ava.js
```diff
@@ -0,0 +1,30 @@
+// This file does not start with "test-" because it is useful as an
+// automated test. Rather, its purpose is just to run it to see what a
+// deep stack looks like.
+
```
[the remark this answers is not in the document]
erights: Done

packages/eventual-send/test/test-deep-stacks-ses-ava.js
```diff
@@ -0,0 +1,18 @@
+// This file does not start with "test-" because it is useful as an
+// automated test. Rather, its purpose is just to run it to see what a
+// deep stack looks like.
+
```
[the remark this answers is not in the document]
erights: Done

packages/eventual-send/test/test-deep-stacks-ses-ava.js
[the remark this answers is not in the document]
erights: Done, thanks!

packages/ses/console-tools.js
```diff
@@ -0,0 +1,26 @@
+import { defineCausalConsoleFromLogger } from './src/error/console.js';
+import { loggedErrorHandler } from './src/error/assert.js';
```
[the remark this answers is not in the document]
erights: The `loggedErrorHandler` itself could be at least a first approximation of such a collection of powers. It is sufficient, in that `defineCausalConsoleFromLogger` itself has no special privilege. All the privilege of `makeCausalConsoleFromLoggerForSesAva` derives only from `loggedErrorHandler`.

In this PR I explicitly decided to postpone this, making only `makeCausalConsoleFromLoggerForSesAva` available outside ses, and only as the value of a awkwardly symbol named property on the global object of what we assume is the start compartment. The comments in `console-shim.js` make clear this is likely only a temporary state of affairs until something more principled, like your suggestion, comes along.

Resolving.

packages/ses-ava/src/ses-ava-test.js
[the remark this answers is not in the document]
erights: `withConsole` is now gone.

packages/ses-ava/src/ses-ava-test.js
erights: Done

packages/ses/src/console-shim.js
erights: Done

packages/ses/src/error/console.js
```diff
@@ -161,9 +179,23 @@ const makeLoggingConsoleKit = (
...
 };
 freeze(makeLoggingConsoleKit);
-export { makeLoggingConsoleKit };
 
-// /////////////////////////////////////////////////////////////////////////////
+/**
+ * Makes the same calls on a `baseConsole` that were made on a
+ * `loggingConsole` to produce this `log`. It this way, a logging console
+ * can be used as a buffer to delay the application of these calls to a
+ * `baseConsole`.
```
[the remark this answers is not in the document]
erights: Done

packages/ses-ava/test/test-raw-ava-reject.js
```diff
@@ -1,57 +1,38 @@
...
-  consoleTaming: 'unsafe', // Doesn't make a difference here
-  errorTaming: 'unsafe', // Redacts entire `error.stack`
+  // Commenting out all settings for a given switch defaults to using
+  // the current relevant environment variable setting. To get results
+  // independent of that, always uncomment one setting for each switch.
+  //
+  stackFiltering: 'concise', // Default. Hide infrastructure, shorten paths
+  // stackFiltering: 'verbose', // Include `assert` infrastructure
+  consoleTaming: 'safe', // Default. Console with access to redacted info
+  // consoleTaming: 'unsafe', // Console without access to redacted info
```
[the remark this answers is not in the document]
erights: Done

packages/ses-ava/test/test-ses-ava-reject.js
```diff
@@ -1,57 +1,48 @@
...
+  // stackFiltering: 'concise', // Default. Hide infrastructure, shorten paths
+  stackFiltering: 'verbose', // Include `assert` infrastructure
+  consoleTaming: 'safe', // Default. Console with access to redacted info
+  // consoleTaming: 'unsafe', // Console without access to redacted info
+  errorTaming: 'safe', // Default. Hide redacted info on error
+  // errorTaming: 'unsafe', // Disclose redacted info on error
 });
 
-const test = wrapTest(rawTest);
+const test = wrapTest(rawTest, { tlog: true, pushConsole: true });
```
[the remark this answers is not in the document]
erights: Thanks for catching. Done

packages/ses-ava/src/ses-ava-test.js
```diff
@@ -76,12 +92,65 @@ const overrideList = [
...
+      }
+    },
+  };
+  const originalProto = getPrototypeOf(originalT);
+  const descs = {
+    ...getOwnPropertyDescriptors(originalProto),
+    ...getOwnPropertyDescriptors(originalT),
+  };
+  for (const [name, desc] of entries(descs)) {
+    if (!(name in virtualT)) {
```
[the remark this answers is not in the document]
erights: The latter ... dominates in an object literal. So I kept the original order and added your comment with correction. Done.

packages/ses-ava/src/ses-ava-test.js
erights: Talked to @kriskowal . This is fine. Resolving.

packages/ses/index.js
erights: Talked to @kriskowal . This is fine. Resolving.

packages/ses-ava/src/ses-ava-test.js
erights: Given my continuing ignorance of type scoping rules + our joint aversion to unnecessary "exports":, I’m planning to proceed with https://github.com/endojs/endo/pull/2109#discussion_r1513654976 unfixed. I expect future waves of type improvement to catch such things, so I think the TODO comment is adequate for now. Resolving.

packages/ses-ava/src/ses-ava-test.js
```diff
@@ -76,12 +92,65 @@ const overrideList = [
...
+      if ('get' in desc) {
```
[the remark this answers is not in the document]
erights: I tried, but it doesn't fit. Resolving with no change.

packages/ses-ava/src/ses-ava-test.js
kriskowal: Let’s merge without an entry in `"exports"`. It’s not needed for this feature and I like how you’ve structured the feature so that this would be easy to extend if we find a need.

packages/ses/src/error/console.js
erights: Oh geez there was a fly in the ointment on this one. A stack trace is multiple lines, i.e., with internal newlines. The logic above left all but the first line unindented.

Rather than find a way to fix the well, I just kludged until I got something acceptable, with a TODO explaining the problems to be fixed. I was tempted to just skip this suggestion to avoid the kludge. But the indented form is so much more pleasant to look at that it was worth it, thanks!

Kludgily done.

erights:
Ping. I addressed all reviewer suggestions. PTAL. thanks.

kriskowal (approved):

PR #2111, opened 2024-03-01
Daemon: add directories

kumavis:
- adds pet-sitter utility for dealing with specialNames
- factors non-mail methods out of mail
- introduces directories and nameHub interface
- removes `listEntries` and `followEntries` from petStore, host, guest
- breaks host and guest API slightly

Updates https://github.com/endojs/endo/pull/1993

### DIFF
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -122,9 +122,7 @@ const makeDaemonCore = async (
     });
   };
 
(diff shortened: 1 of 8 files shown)


kriskowal (approved):
I see this is in draft. Please feel free to dismiss this stamp when it’s r4r, or use it if it’s draft in error.

PR #2114, opened 2024-03-03
feat(daemon): multiplayer

kumavis:
- adds NETS network directory which are automatically started on boot, used for generating network addresses, and attempting connections to peers.
- adds a nodeIdentifier to formula identifiers that specifies where (what peer) a formula/value lives
- adds an ownNodeIdentifier for identifying local formula/values
- adds a peer store that stores the peers by nodeIdentifier
- `provideValueForFormulaIdentifier` looks up non-local values via the formulaIdentifier's nodeIdentifiers when peer is known
- adds `getPeerInfo`, `addPeerInfo` for introducing peers
- adds a gateway (used by network caplets) for exposing value requests from peers
- peers are local formulas that expose the remote gateway via the matching network caplet


packages/daemon/src/pet-store.js
```diff
@@ -7,7 +7,22 @@ const { quote: q } = assert;
 
 const validIdPattern = /^[0-9a-f]{128}$/;
 const validFormulaPattern =
-  /^(?:(?:readable-blob|worker|pet-store|pet-inspector|eval|lookup|make-unconfined|make-bundle|host|guest|handle):[0-9a-f]{128}|web-bundle:[0-9a-f]{32})$/;
+  /^(?:(?:readable-blob|worker|pet-store|pet-inspector|eval|lookup|make-unconfined|make-bundle|host|guest|handle|peer|remote|loopback-network):[0-9a-f]{128}:[0-9a-f]{128}|web-bundle:[0-9a-f]{32}:[0-9a-f]{128})$/;
```
rekmarks, line 10: This should not include `remote`:
```suggestion
  /^(?:(?:readable-blob|worker|pet-store|pet-inspector|eval|lookup|make-unconfined|make-bundle|host|guest|handle|peer|loopback-network):[0-9a-f]{128}:[0-9a-f]{128}|web-bundle:[0-9a-f]{32}:[0-9a-f]{128})$/;
```

packages/daemon/src/directory.js
```diff
@@ -95,6 +95,21 @@ export const makeDirectoryMaker = ({
...
+      const identities = new Set();
+      await Promise.all(
+        names.map(async name => {
+          const formulaIdentifier = await identify(...petNamePath, name);
+          if (formulaIdentifier !== undefined) {
+            identities.add(formulaIdentifier);
+          }
+        }),
+      );
+      return Array.from(identities);
```
kriskowal, line 110:
```suggestion
      return harden(Array.from(identities).sort());
```

packages/daemon/src/networks/loopback.js
```diff
@@ -0,0 +1,25 @@
...
+export const makeLoopbackNetwork = ({ provideValueForFormulaIdentifier }) => {
+  return Far(
+    'Loopback Network',
+    /** @type {import('../types.js').EndoNetwork} */ ({
+      addresses: () => [],
+      supports: address => new URL(address).protocol === 'loop:',
+      connect: address => {
+        if (address !== 'loop:') {
+          throw new Error(
+            'Failed invariant: loopback only supports loop: address',
```
rekmarks, line 18:
```suggestion
            'Failed invariant: loopback only supports "loop:" address',
```

packages/daemon/src/formula-identifier.js
```diff
@@ -12,7 +12,13 @@ export const parseFormulaIdentifier = formulaIdentifier => {
     );
   }
 
-  const type = formulaIdentifier.slice(0, delimiterIndex);
-  const number = formulaIdentifier.slice(delimiterIndex + 1);
-  return { type, number };
+  const [type, number, location] = formulaIdentifier.split(':');
```
kriskowal, line 15: I’m guessing this eventually evolves to the `endo://NODE/VALUE?type=TYPE` subset of `endo://NODE/VALUE?type=TYPE&subtype=&address=&address=`

packages/daemon/src/host.js
```diff
@@ -21,8 +21,10 @@ const { quote: q } = assert;
  * @param {import('./types.js').DaemonCore['incarnateWebBundle']} args.incarnateWebBundle
  * @param {import('./types.js').DaemonCore['incarnateHandle']} args.incarnateHandle
  * @param {import('./types.js').DaemonCore['storeReaderRef']} args.storeReaderRef
+ * @param {import('./types.js').DaemonCore['getAllNetworkAddresses']} args.getAllNetworkAddresses
```
kriskowal, line 24: I’m fine abbreviating this to `addresses`

packages/daemon/src/host.js
```diff
@@ -37,24 +39,28 @@ export const makeHostMaker = ({
   incarnateWebBundle,
   incarnateHandle,
   storeReaderRef,
+  getAllNetworkAddresses,
   makeMailbox,
   makeDirectoryNode,
+  ownLocation,
```
kriskowal, line 45: Maybe `nodeIdentifier`?

packages/daemon/src/host.js
```diff
@@ -326,7 +334,7 @@ export const makeHostMaker = ({
 
       if (resultName !== undefined) {
         addHook(identifiers =>
-          petStore.write(resultName, `eval:${identifiers.evalFormulaNumber}`),
+          petStore.write(resultName, identifiers.evalFormulaIdentifier),
```
kriskowal, line 337: huzzah

packages/daemon/src/host.js
```diff
@@ -514,12 +523,39 @@ export const makeHostMaker = ({
       return cancelValue(formulaIdentifier, reason);
     };
 
+    /** @type {import('./types.js').EndoHost['gateway']} */
+    const gateway = async () => {
```
kriskowal, line 527: I like where this is headed. Maybe leave a note that this corresponds to the nonce locator from E. Similar names include entrance, threshold, door, welcome.

packages/daemon/src/host.js
```diff
@@ -514,12 +523,39 @@ export const makeHostMaker = ({
...
+      // TODO: we should only materialize local objects this way
+      return Far('Gateway', { provide: provideValueForFormulaIdentifier });
```
kriskowal, line 529: This should probably be a singleton that the `gateway` method returns but does not construct.

packages/daemon/src/host.js
```diff
@@ -514,12 +523,39 @@ export const makeHostMaker = ({
...
+    };
+
+    /** @type {import('./types.js').EndoHost['addPeerInfo']} */
+    const addPeerInfo = async peerInfo => {
+      const endoBootstrap =
+        /** @type {import('./types.js').FarEndoBootstrap} */ (
+          await lookup('ENDO')
+        );
```
kriskowal, line 537: We could thread the bootstrap object or a getter for the bootstrap object into the host maker args for looser coupling to the special names.

packages/daemon/src/host.js
```diff
@@ -514,12 +523,39 @@ export const makeHostMaker = ({
...
+      await E(endoBootstrap).addPeerInfo(peerInfo);
+    };
+
+    /** @type {import('./types.js').EndoHost['getPeerInfo']} */
+    const getPeerInfo = async () => {
+      const addresses = await getAllNetworkAddresses(
+        networksDirectoryFormulaIdentifier,
+      );
+      const peerInfo = {
+        location: ownLocation,
```
kriskowal, line 547: Location is going to become quickly ambiguous since we will have locations of varying specificity:

- `endo://node` node location
- `endo://node/value` value location
- `endo://node/value.type` typed value location
- `endo://node/value.type?address=&address=` value location (née offline capability) with connection hints
- `endo://node/value.blob-reader-base64/application/javascript?address=&address=` MIME-typed downloadable blob location over base64 until CapTP maybe someday can do byte streams

This can perhaps just be `peer` implying “peer identifier” or “peer public key” someday

kriskowal (approved):
Only notes.

rekmarks (changes requested):

packages/daemon/src/daemon.js
```diff
@@ -82,6 +88,23 @@ const makeDaemonCore = async (
...
+    'location',
+    rootEntropy,
+    cryptoPowers.makeSha512(),
+  );
+  // eslint-disable-next-line no-use-before-define
+  const peersFormulaNumber = getDerivedId(
+    'peers',
+    rootEntropy,
+    cryptoPowers.makeSha512(),
+  );
```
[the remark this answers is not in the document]
kumavis: I made a refactor that made sense at the time that separated endoCore generic methods from the configured Endo. this looked good at the time but now provideValueforFormulaIdentifier (and everything that uses it) rely on the configuration of the ownLocation. I made the smallest change here, but it invalidated the previous refactor.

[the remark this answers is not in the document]
rekmarks: So we can just use random numbers instead?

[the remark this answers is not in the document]
kumavis: after arm wrestling `daemonCore` a bit, yes

rekmarks (approved):

kumavis:
:tada: :tada: :tada:

PR #2115, opened 2024-03-06
refactor(daemon): Index formulas by number

kriskowal:
This is a sequence of refactors that get us closer to indexing controllers by formula numbers (without type).

### DIFF
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -669,8 +669,8 @@ const makeDaemonCore = async (
     context,
   ) => {
     const {
-      type: formulaType,
       number: formulaNumber,
+      type: allegedFormulaType,
       node: formulaNode,
     } = parseId(formulaIdentifier);
     const isRemote = formulaNode !== ownNodeIdentifier;
@@ -683,8 +683,16 @@ const makeDaemonCore = async (
       // eslint-disable-next-line no-use-before-define
       return provideRemoteValue(peerIdentifier, formulaIdentifier, context);
     }
+    const formula = await persistencePowers.readFormula(formulaNumber);
     if (
-      [
+      allegedFormulaType !== undefined &&
+      formula.type !== allegedFormulaType
+    ) {
+      assert.Fail`Alleged formula type ${q(
+        allegedFormulaType,
+      )} is not the actual formula type {formula.type}`;
+    } else if (
+      ![
         'endo',
         'worker',
         'eval',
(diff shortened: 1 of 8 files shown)


packages/daemon/src/formula-identifier.js
```diff
@@ -1,24 +1,68 @@
+/// <ref types="ses">
+
 const { quote: q } = assert;
 
+const idPattern =
+  /^(?:(?<type>endo|readable-blob|worker|pet-store|pet-inspector|eval|lookup|make-unconfined|make-bundle|host|guest|handle|peer|directory|least-authority|loopback-network):)?(?<number>[0-9a-f]{128}):(?<node>[0-9a-f]{128})|(?<type2>web-bundle):(?<number2>[0-9a-f]{32})(?<node2>:[0-9a-f]{128})$/;
```
rekmarks, line 6: The last `web-bundle` capture group includes the final `:`.
```suggestion
const idPattern =
  /^(?:(?<type>endo|readable-blob|worker|pet-store|pet-inspector|eval|lookup|make-unconfined|make-bundle|host|guest|handle|peer|directory|least-authority|loopback-network):)?(?<number>[0-9a-f]{128}):(?<node>[0-9a-f]{128})|(?<type2>web-bundle):(?<number2>[0-9a-f]{32}):(?<node2>[0-9a-f]{128})$/;
```

packages/daemon/src/formula-identifier.js
```diff
@@ -1,24 +1,68 @@
...
+
+/**
+ * @param {string} formulaIdentifier
```
rekmarks, line 9: Would we welcome a PR of `s/formulaIDentifier/formulaId/g` after this is merged?

packages/daemon/src/formula-identifier.js
```diff
@@ -1,24 +1,68 @@
...
-  `${type}:${number}:${node}`;
+export const formatId = ({ type, number, node }) => {
+  const parts = [number];
+  if (type !== undefined) {
+    parts.unshift(type);
+  }
+  if (node !== undefined) {
+    parts.push(node);
+  }
+  return parts.join(':');
```
rekmarks, line 67: As a defensive measure, should we call `assertValidId` on the constructed id before returning it? If so, we should perhaps change the error message to differentiate this case from that of an invalid already-constructed id.

```suggestion
  const id = parts.join(':');
  assertValidId(id);
  return id;
```

rekmarks (changes requested):
One RegEx-related bug, then just optional questions and suggestions.

packages/daemon/src/formula-identifier.js
kriskowal: I would like to do this and we should book a quiet day to do it on.

packages/daemon/src/daemon.js
```diff
@@ -698,24 +708,19 @@ const makeDaemonCore = async (
...
-      );
-    } else {
-      throw new TypeError(
-        `Invalid formula identifier, unrecognized type ${q(formulaIdentifier)}`,
-      );
+      assert.Fail`Invalid formula identifier, unrecognized type ${q(
+        allegedFormulaIdentifier,
+      )}`;
     }
+    // TODO further validation
```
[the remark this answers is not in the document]
kriskowal: We would presumably validate the shape farther down, in each individual formula type reviver. For the moment, we assume an absence of corruption, either automatic or manual. We can probably get away with that for a while.

packages/daemon/src/pet-store.js
```diff
@@ -248,7 +238,7 @@ export const makePetStoreMaker = (filePowers, locator) => {
    * @returns {Promise<import('./types.js').PetStore>}
    */
   const makeIdentifiedPetStore = (id, assertValidName) => {
-    if (!validIdPattern.test(id)) {
+    if (!validNumberPattern.test(id)) {
```
rekmarks, line 241: The `id` param here should have been renamed `formulaNumber` in a previous PR. Likewise in `types.d.ts`.

rekmarks (approved):
One suggestion, otherwise LGTM!

PR #2119, opened 2024-03-08
feat(daemon): locate based lookup with three party handoff

kumavis:
rework recursive lookup to use identify/locate. this achieves third party handoff with automatic introductions to new peers.

what do we do when things support `lookup` but not `locate`?


kriskowal:
attn @dckc We have a dilemma with the pet daemon’s implementation of pet name path lookup. Changing the protocol for for following pet-name paths to internally call `locate` instead of `lookup` (producing location of the value instead of the value) and a final call to `provide(location)` to obtain the value trivially solves three-party hand-off and introduction to remote peers, but loses us the ability to follow chains of name hubs that do not implement `locate`.

This might be a case where having some interface hints on the location would help us decide whether to dispatch `locate` or `lookup`, leaning on `lookup` by default. Consider a location like `endo://node-id/id?has=locate&address=...&address=...`.

dckc:
> attn @dckc We have a dilemma with the pet daemon’s implementation of pet name path lookup...

hm. um... I don't see a way out. I would like to phone a friend. @michaelfig ?

cc @gibson042 @erights

kriskowal:
The answer is probably that we should have one method that supports hand-off and another that explicitly does not, such that you can choose whether your node masks the route to the target or hands the target node off to the source. Maybe that should even be expressible with a different kind of dot in dotted paths.

PR #2120, opened 2024-03-08
refactor(daemon): Synchronize makeGuest

rekmarks:
Progresses: #2086 

Synchronizes the host's `makeGuest()` and its dependencies. See #2086 for details.


packages/daemon/src/pet-store.js
```diff
@@ -85,12 +85,14 @@ export const makePetStoreMaker = (filePowers, locator) => {
...
       assertValidFormulaIdentifier(formulaIdentifier, petName);
 
+      // TODO: Return early if the formula identifier is the same.
       if (petNames.has(petName)) {
         // Perform cleanup on the overwritten pet name.
         const formulaPetNames = formulaIdentifiers.get(petName);
         if (formulaPetNames !== undefined) {
           formulaPetNames.delete(petName);
         }
+        // TODO: Should this only happen if something is actually deleted?
```
rekmarks, line 95: I noticed these couple oddities in the pet store's `write()`. I'll create an issue if we agree that there's something to do.

packages/daemon/src/host.js
```diff
@@ -472,13 +468,12 @@ export const makeHostMaker = ({
           )}`,
         );
       }
+
+      introduceNamesToParty(formulaIdentifier, introducedNames);
```
rekmarks, line 472: Updated to match the new location `makeGuest()`. Also, `introducedNames` is never `undefined` on account of the default parameter value.

packages/daemon/src/async-hooks.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * @returns {import('./types.js').AsyncHooks<any>}
+ */
+export const makeAsyncHooks = () => {
```
kriskowal, line 4: There’s a Node.js module by the same name that we don’t want to confuse this with. This is similar to a `WaitGroup` in Go. I think `DeferredSet` might be good.

packages/daemon/src/daemon.js
```diff
@@ -78,6 +78,11 @@ const getDerivedId = (path, rootNonce, digester) => {
   return nonce;
 };
 
+/**
+ * @template T
+ * @typedef {import('./types.js').IncarnateResult<T>} IncarnateResult
+ */
```
kriskowal, line 84: Let’s inline these. Typedefs create new types and interrupt “go to definition” chains.

packages/daemon/src/daemon.js
```diff
@@ -972,6 +973,14 @@ const makeDaemonCore = async (
     );
   };
 
+  /**
+   * @type {import('./types.js').DaemonCore['incarnateWorker']}
+   */
+  const incarnateWorker = async () => {
+    const formulaNumber = await formulaGraphMutex.enqueue(randomHex512);
+    return incarnateNumberedWorker(formulaNumber);
```
kriskowal, line 981: I’m unsure whether this is sufficiently synchronized since `randomHex512` will complete and release the lock before `incarnateNumberedWorker` has an opportunity to add the number to the table.

packages/daemon/src/daemon.js
```diff
@@ -1006,12 +1015,48 @@ const makeDaemonCore = async (
...
+        randomHex512(),
+        randomHex512(),
+        randomHex512(),
+        randomHex512(),
+      ]).then(([own, handle, store, worker]) =>
+        Promise.all([
+          own,
+          incarnateNumberedHandle(handle, hostFormulaIdentifier),
+          incarnateNumberedPetStore(store),
+          incarnateNumberedWorker(worker),
```
kriskowal, line 1040: This parallelism is probably more elaborate than necessary given that the incarnations are going to be serialized.

packages/daemon/src/host.js
```diff
@@ -132,48 +132,48 @@ export const makeHostMaker = ({
...
-        const { formulaIdentifier: hostHandleFormulaIdentifier } =
-          await makeNewHandleForSelf();
-        const { value, formulaIdentifier: guestFormulaIdentifier } =
-          // Behold, recursion:
-          // eslint-disable-next-line no-use-before-define
-          await incarnateGuest(hostHandleFormulaIdentifier);
-        if (petName !== undefined) {
-          assertPetName(petName);
-          await petStore.write(petName, guestFormulaIdentifier);
+          // TODO: Should this be awaited?
```
kriskowal, line 146: If it returns a promise, yes. And it does.

packages/daemon/src/host.js
```diff
@@ -132,48 +132,48 @@ export const makeHostMaker = ({
...
+
+      const { value, formulaIdentifier } =
+        // Behold, recursion:
+        // eslint-disable-next-line no-use-before-define
+        await incarnateGuest(hostFormulaIdentifier, hooks);
+      // TODO: Should this be awaited?
+      introduceNamesToParty(formulaIdentifier, introducedNames);
+
       return {
+        value: Promise.resolve(value),
```
kriskowal, line 175: Is the type system demanding we wrap these? Maybe we should use `Awaited<T>` somewhere.

kriskowal (approved):

packages/daemon/src/host.js
kumavis: yes

packages/daemon/src/daemon.js
rekmarks: Hm, I think that concern applies to all extant uses of `formulaGraphMutex`. Looking at the implementation of the mutex, we synchronously call `unlock()` in a `finally` block after a `return await asyncFn()` (in this case it would be `randomHex512`). Shouldn't control be relinquished to L980 here before anything else that was waiting on the lock?

I could try to write a test for this.

packages/daemon/src/host.js
rekmarks: The type system does demand this. Since this was already here, I'll leave it for now. Should we create an issue?

packages/daemon/src/host.js
rekmarks: #2121

packages/daemon/src/async-hooks.js
rekmarks: I went with `deferredTasks`.

packages/daemon/src/daemon.js
rekmarks: #2122

packages/daemon/src/pet-store.js
rekmarks: @kriskowal thoughts on these?

packages/daemon/src/host.js
kriskowal: Meh. Not big enough to track.

packages/daemon/src/pet-store.js
kriskowal: It would greatly simplify client code if they are assured that they only see `delete` messages for equal but opposite prior `add`.

PR #2123, opened 2024-03-08
refactor!(daemon): Remove host `makeWorker()`

rekmarks:
Progresses: #2086 

Removes the host's `makeWorker()` and replaces its use everywhere with the equivalent `provideWorker()`. The latter is essentially like the former but with more validation.

### DIFF
--- a/packages/daemon/test/test-endo.js
+++ b/packages/daemon/test/test-endo.js
@@ -96,7 +96,7 @@ test('lifecycle', async t => {
   );
   const bootstrap = getBootstrap();
   const host = E(bootstrap).host();
-  await E(host).makeWorker('worker');
+  await E(host).provideWorker('worker');
   await E(host).cancel('worker');
   cancel(new Error('Cancelled'));
   await closed.catch(() => {});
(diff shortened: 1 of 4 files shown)


packages/daemon/test/test-endo.js
```diff
@@ -150,6 +150,31 @@ test('anonymous spawn and evaluate', async t => {
...
+    locator.sockPath,
+    cancelled,
+  );
+  const bootstrap = getBootstrap();
+  const host = E(bootstrap).host();
+  const ten = await E(host).evaluate('MAIN', '10', [], [], 'ten');
+  t.is(ten, 10);
+  await t.throwsAsync(() => E(host).provideWorker('ten'), {
+    message: 'Not a worker "ten"',
+  });
```
rekmarks, line 173: Rather than throwing, `makeWorker()` would resolve with a worker.

kumavis:
"provide" usually means get me the thing and if it doesn't exist yet, create it

kumavis (approved):
seems fine

noticed this identifier type check, it should use parse identifier or something else
https://github.com/endojs/endo/blob/6516b25aae304618ed1c7830b9b18cb2df2d4162/packages/daemon/src/host.js#L210

PR #2124, opened 2024-03-08
refactor(daemon): Synchronize host `makeUnconfined()`

rekmarks:
Progresses: #2086

Synchronizes the host's `makeUnconfined()` per #2086. Refactoring `daemon.js` in support of this goal fixed one bug while revealing another.

In particular, #2074 is progressed by enabling indirect cancellation of caplets via their workers. The issue is not resolved since indirect cancellation of caplets via their caplet dependencies still does not work as intended. A new, failing regression test has been added for this specific case.

The revealed bug is #2021, which we believed to be fixed by #2092. Rather than fixing the bug, that PR concealed it by always creating a new incarnation of `eval` formula workers, even if they already existed. The regression test for #2021 has been marked as failing, and we will have to find a different solution for it.

### DIFF
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -1007,41 +1007,38 @@ const makeDaemonCore = async (
     );
   };
 
-  /** @type {import('./types.js').DaemonCore['incarnateGuest']} */
-  const incarnateGuest = async (hostFormulaIdentifier, deferredTasks) => {
+  /**
+   * Helper for callers of `incarnateNumberedGuest`.
+   * @param {string} hostFormulaIdentifier - The formula identifier of the host.
+   */
+  const incarnateGuestDependencies = async hostFormulaIdentifier =>
+    harden({
+      guestFormulaNumber: await randomHex512(),
+      hostHandleFormulaIdentifier: (
+        await incarnateNumberedHandle(
+          await randomHex512(),
+          hostFormulaIdentifier,
+        )
+      ).formulaIdentifier,
+      storeFormulaIdentifier: (
+        await incarnateNumberedPetStore(await randomHex512())
+      ).formulaIdentifier,
+      workerFormulaIdentifier: (
+        await incarnateNumberedWorker(await randomHex512())
+      ).formulaIdentifier,
+    });
+
+  /**
+   *
+   * @param {ReturnType<any>} identifiers
+   */
+  const incarnateNumberedGuest = identifiers => {
     const {
       guestFormulaNumber,
       hostHandleFormulaIdentifier,
       storeFormulaIdentifier,
       workerFormulaIdentifier,
-    } = await formulaGraphMutex.enqueue(async () => {
-      const formulaNumber = await randomHex512();
-      const hostHandle = await incarnateNumberedHandle(
-        await randomHex512(),
-        hostFormulaIdentifier,
-      );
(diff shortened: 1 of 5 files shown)


kumavis (commented):
looks a little complicated, but idk how to make it simpler yet

rekmarks:
@kumavis I hope to simplify things as the synchronization effort continues, but the simplest way of doing things may not reveal itself until late in that process.

packages/daemon/src/host.js
```diff
@@ -283,6 +287,21 @@ export const makeHostMaker = ({
       return guestFormulaIdentifier;
     };
 
+    /**
+     * @param {string | 'NONE' | 'SELF' | 'ENDO'} partyName
+     * @param {import('./types.js').DeferredTasks<{ powersFormulaIdentifier: string }>['push']} deferTask
+     * @returns {string | undefined}
+     */
+    const providepowersFormulaIdentifierSync = (partyName, deferTask) => {
```
kriskowal, line 295:
```suggestion
    const providePowersFormulaIdentifierSync = (partyName, deferTask) => {
```

packages/daemon/src/host.js
```diff
@@ -356,24 +375,36 @@ export const makeHostMaker = ({
...
+      /** @type {import('./types.js').DeferredTasks<import('./types.js').MakeUnconfinedDeferredTaskParams>} */
+      const tasks = makeDeferredTasks();
+
+      const workerFormulaIdentifier = provideWorkerFormulaIdentifierSync(
         workerName,
+        tasks.push,
       );
 
-      const powersFormulaIdentifier = await providePowersFormulaIdentifier(
+      const powersFormulaIdentifier = providepowersFormulaIdentifierSync(
```
kriskowal, line 388:
```suggestion
      const powersFormulaIdentifier = providePowersFormulaIdentifierSync(
```

packages/daemon/test/test-endo.js
```diff
@@ -979,8 +1034,8 @@ test('name and reuse inspector', async t => {
   await stop(locator);
 });
 
-// Regression test for: https://github.com/endojs/endo/issues/2021
-test('eval-mediated worker name', async t => {
+// Regression test for https://github.com/endojs/endo/issues/2021
+test.failing('eval-mediated worker name', async t => {
```
kriskowal, line 1038: So, we have whacked a mole?

kriskowal (approved):
With a nit.

packages/daemon/test/test-endo.js
rekmarks: I am consulting my exterminator as to our next steps.

packages/daemon/src/daemon.js
```diff
@@ -1007,41 +1007,38 @@ const makeDaemonCore = async (
     );
   };
 
-  /** @type {import('./types.js').DaemonCore['incarnateGuest']} */
-  const incarnateGuest = async (hostFormulaIdentifier, deferredTasks) => {
+  /**
+   * Helper for callers of `incarnateNumberedGuest`.
+   * @param {string} hostFormulaIdentifier - The formula identifier of the host.
+   */
+  const incarnateGuestDependencies = async hostFormulaIdentifier =>
```
kriskowal, line 1014: Should be `provideGuestDependencies`. Doesn’t create formulas. Renaming `formulate*` functions should make that distinction clearer.

PR #2132, opened 2024-03-11
build: .d.ts for nat

turadg:
## Description

The types weren't being built in the NPM package. Downstream would get them sometimes by node modules search, but only if within sufficient search depth.

This removes the need for searching node_modules

### Security Considerations

none

### Scaling Considerations

none

no, more like other packages now

### Testing Considerations

CI

fully back compat

### Upgrade Considerations

none

### DIFF
--- a/packages/nat/package.json
+++ b/packages/nat/package.json
@@ -10,9 +10,11 @@
   },
   "scripts": {
     "build": "exit 0",
+    "build:types": "tsc --build tsconfig.build.json",
+    "clean:types": "git clean -f '*.d.ts*'",
     "test": "ava test/**/test-*.js",
     "lint-fix": "eslint --fix",
-    "lint-check": "eslint"
+    "lint": "tsc"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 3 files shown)


kriskowal (approved):

packages/nat/package.json
```diff
@@ -10,9 +10,11 @@
   },
   "scripts": {
     "build": "exit 0",
+    "build:types": "tsc --build tsconfig.build.json",
+    "clean:types": "git clean -f '*.d.ts*'",
     "test": "ava test/**/test-*.js",
     "lint-fix": "eslint --fix",
-    "lint-check": "eslint"
+    "lint": "tsc"
```
kriskowal, line 17: Do we lose eslint here?

turadg: It wasn't working. It wasn't erroring in CI because the script wasn't called "lint"

kriskowal: Cool story, tracking https://github.com/endojs/endo/issues/2134

PR #2133, opened 2024-03-11
feat(ses-ava): import test from @endo/ses-ava/prepare-endo.js

erights:
closes: #XXXX
refs: #XXXX

## Description

At https://github.com/Agoric/agoric-sdk/pull/9026#discussion_r1511360711 @turadg suggested that rather than scatter redundant `prepare-test-env-ava.js` test files in each test directory, that @endo/ses-ava should just export a `test` that most ses-ava users can just use directly. This PR attempts to provide that.

To do so, I had to change the layering so that the new `@endo/ses-ava/prepare-test-env-ava.js` can do the endo initialization that most clients expect, which includes setting up both `ses` and `@endo/eventual-send` with various options set up appropriately for debugging. Thus, for all those packages that @endo/ses-ava now depends on, directly or indirectly, I changed them to use `'ava'` rather than `@endo/ses-ava` to avoid circular build dependencies.

I also had to move `@endo/ses-ava`'s dependency on `'ava'` from `"devDependencies":` `"dependencies":`. This creates a hazard that bundlers might try to bundle `'ava'`, which would be bad. To avoid that, all packages using `@endo/ses-ava` must depend on it as a `"devDependencies":`, which they should do anyway.

From a pair debugging session with @Chris-Hibbert just now, I think the options set up by this new `@endo/ses-ava/prepare-test-env-ava.js` are adequate for him to get the stack traces he needed.

### Security Considerations
none
### Scaling Considerations
none
easier to document how to set up tests in a test directory.
### Testing Considerations

The point. This should both make it easier to set up a test directory, and to get more options set right to support debugging.

This PR mades a coordinated change between various endo packages, without trying to tolerate version slippage among them. This is all in preparation for a new endo release, and effects on tests, so should be fine. Should have no compat impact on any existing clients of anything in endo.

### Upgrade Considerations
none

Nothing breaking. NEWSworthy.


packages/bundle-source/test/test-bigint-transform.js
```diff
@@ -1,6 +1,6 @@
 // @ts-check
 import url from 'url';
-import { test } from './prepare-test-env-ava.js';
+import { test } from '@endo/ses-ava/prepare-test-env-ava.js';
```
turadg, line 3: would be nice for it to match Ava:
```suggestion
import test from '@endo/ses-ava`;
```
The only existing export, `wrapTest`, is longer necessary.

turadg (changes requested):

packages/bundle-source/test/test-bigint-transform.js
erights: Removing `wrapTest` would be a breaking change to existing clients, especially agoric-sdk.

packages/ses-ava/package.json
```diff
@@ -36,12 +37,12 @@
     "test": "ava"
   },
   "dependencies": {
+    "@endo/env-options": "^1.1.1",
+    "@endo/init": "^1.0.4",
+    "ava": "^5.3.0",
```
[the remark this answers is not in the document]
erights: I have no experience with peer dependencies and don't understand them. Are you ok with this peer dependencies change not happening in this PR, given that placing ses-ava in devDependencies solved the immediate problem?

Any such deeper change in our dependency structure for endo I'd like to leave in @kriskowal 's hands anyway. @kriskowal ?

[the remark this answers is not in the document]
turadg: > placing ses-ava in devDependencies solved the immediate problem

I expect there will be other problems from using `dependencies`. I'm willing to defer to @kriskowal .

Chris-Hibbert:
I've looked at the changes. I approve of the goal, but I leave it to others to approve the PR.

packages/bundle-source/test/test-bigint-transform.js
erights: Currently, this PR still leaves some endo uses of `wrapTest` in place. Most of these look like candidate to use the new import directly. I'll try in out is separate commits and see how it goes.

kriskowal: It’s not necessary to remove `wrapTest` in order to also export a default `test`. I would like to keep the ball moving forward.

kriskowal: Reminder that this weird trick is possible in the unlikely event it is necessary:
```js
import test, { wrapTest } from '@endo/ses-ava';
```

packages/ses-ava/package.json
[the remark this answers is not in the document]
kriskowal: I prefer adding `ava` to the `dependencies` of `@endo/ses-ava` over making it a peer dependency.

Using `peerDependencies` means that a application package has to retain a `devDependencies` entry for both `@endo/ses-ava` and a compatible version range of `ava`. I like `peerDependencies` for packages that might or might not be retained by the application package depending on whether they’re actually used, but in this case, Ava is always necessary for the functioning of `@endo/ses-ava`.

Please note that by “application”, I mean the package that ultimately retains the `devDependencies` tree.

packages/bundle-source/test/test-bigint-transform.js
erights: DONE!

Turns out, JS supports a module have both named exports and a default export.

- All endo uses of `wrapTest` are now gone. 
- I left `wrapTest` in as a named export and commented it as deprecated (though not `@deprecated`). 
- I made `test` a default export of `@endo/ses-ava` 
- I removed the new package export `@endo/ses-ava/prepare-test-env-ava.js` and replaced all uses with
```js
import test from '@endo/ses-ava';
```
exactly as @turadg suggested.

***HOWEVER***, even leaving `wrapTest` in as a deprecated export, doing this does introduce a version slippage incompatibility:

Anyone doing 
```js
import { wrapTest } from '@endo/ses-ava';
```
had to do their own separate import to initialize ses and call `lockdown`. In order to export the new `test` from the same `index.js`, I had to make it do that initialization itself. That means that some existing clients will fail when these two initializations of ses collide.

erights: @kriskowal , I see you beat me to the punch ;)

But please see the ***HOWEVER*** in the immediately preceding comment. Advice appreciated.

packages/ses-ava/package.json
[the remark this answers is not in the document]
erights: Thanks. Resolving.

erights:
At https://github.com/Agoric/agoric-sdk/pull/9026#discussion_r1511357998 @turadg suggested removing the unneeded lint suppression. With this change, I was able to remove all unneeded
```js
// eslint-disable-next-line import/order
```
Good riddance! Done.

packages/bundle-source/test/test-bigint-transform.js
kriskowal: Ah, thank you for pointing out the caveat. I might have otherwise missed it.

I think we need a new entry-point for the convenience of importing a pre-wrapped `test`, both because of the backward-incompatibility you mention (which I’m sure will be caught with an integration branch check PR), but also because this precludes using `@endo/ses-ava` with any other combination of shims. Consider:

It might even be useful to have separate convenience entry-points for other layerings. That would let us continue to use `@endo/ses-ava` below the eventual-send layer.

```js
import test from '@endo/ses-ava/prepare-ses.js';
```

```js
import test from '@endo/ses-ava/prepare-endo.js';
```

This would allow folks to continue using `wrapTest` from `@endo/ses-ava` to compose other preparations at higher layers.

erights: Got it. I will reverse by bundling of the two into `index.js` and remove the deprecation comment on `wrapTest`.

`prepare-endo.js` seems like a fine name, even though it is still below the level of most of endo, since most of endo uses it.

packages/ses-ava/package.json
```diff
@@ -22,7 +22,8 @@
   "exports": {
     ".": "./index.js",
     "./exported.js": "./exported.js",
-    "./package.json": "./package.json"
+    "./package.json": "./package.json",
+    "./prepare-endo.js": "./prepare-endo.js"
```
kriskowal, line 26: ✅ Got it in one.

Nit: we put `package.json` at the end of the list in all other packages.

kriskowal (approved):

packages/ses-ava/package.json
erights: Done

packages/bundle-source/test/test-bigint-transform.js
erights: Separating it back out and renaming to prepare-endo.js, Done!

erights:
@turadg , I acted on all suggestions as well as possible and git @kriskowal 's approval. PTAL.

Thanks for the great suggestions! It really is much better now.

turadg (approved):
Much better!

We had planned,
- https://github.com/endojs/endo/issues/1467 

but I think this is a good solution to the side-effect surprises and boilerplate. @gibson042 should this PR close that issue?

packages/ses-ava/package.json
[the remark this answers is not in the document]
turadg: One of my concerns was version coupling and duplicate deps. I agree that on balance `dependencies` is the better choice.

But to reduce the coupling, mind making the version range accept Ava 6 too? I don't see any incompatible changes in https://github.com/avajs/ava/releases/tag/v6.0.0 (though you would know better) And in the other direction, I don't know that it necessarily requires 5.3 vs 5.0.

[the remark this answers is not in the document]
erights: @kriskowal just noting:

While looking into this, I notice that at 
https://github.com/endojs/endo/blob/5186a5040689f27719f79f1f8f808dbc46b7af0d/packages/init/package.json#L34

@endo/init currently depends on Ava "^5.1.1" whereas all other endo packages currently depend on Ava "^5.3.0". I am proceeding assuming this is a mistake unless I hear otherwise.

What tooling do we have for checking such consistency among package dependencies?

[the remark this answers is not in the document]
kriskowal: `./scripts/all-versions.sh` gives you the most recent of all transitive dependencies and piping that to `./scripts/set-verison.sh` will give you a diff for any discrepancies. Some of those discrepancies are deliberate, e.g., `typescript` is pinned with `~` instead of `^` because of differences in how they interpret their semver obligations.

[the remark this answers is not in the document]
erights: Ava 6.0.0 does not work with Node 16. Endo still seems to require Node 16, or at least endo's CI tests still do:

image

So, reverting the Ava dependencies back to "^5.3.0". If it does not cause CI problems, I will keep the change of @endo/init from Ava "^5.1.1" to "^5.3.0".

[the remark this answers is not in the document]
erights: Done!

gibson042:
> We had planned,
> 
> * [ses-ava should not mix exports and side-effects #1467](https://github.com/endojs/endo/issues/1467)
> 
> but I think this is a good solution to the side-effect surprises and boilerplate. @gibson042 should this PR close that issue?

Yes, I would say so. This is great!

PR #2135, opened 2024-03-12
upgrade TypeScript to 5.4

turadg:
## Description

https://devblogs.microsoft.com/typescript/announcing-typescript-5-4/

### Security Considerations

No change

### Scaling Considerations

n/a

Unnecessary

### Testing Considerations

CI suffices

none

### Upgrade Considerations

none

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "ts-api-utils": "~1.0.1",
     "type-coverage": "^2.26.3",
     "typedoc": "^0.25.7",
-    "typescript": "~5.3.3"
+    "typescript": "~5.4.2"
   },
   "scripts": {
     "clean": "lerna clean",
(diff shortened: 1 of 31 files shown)


kriskowal (approved):
Can I ask for an Agoric SDK PR pinned to this branch to verify this leaves integration in a good place?

kriskowal:
I see nothing in here that can possibly break Agoric SDK but I’ve guessed wrong before. Your discretion. I think there are some bundleSource type issues on `endo/master` that need to be resolved anyway and you might see a false negative.

turadg:
> I see nothing in here that can possibly break Agoric SDK but I’ve guessed wrong before. Your discretion.

Any changes to the types that are output will be correcting problems. While I don't expect it to cause any work for agoric-sdk if it does that will be fixing bugs. If it's onerous for the sync process I will do it.

PR #2139, opened 2024-03-13
refactor(daemon): Coordinate host `provideX` implementations

rekmarks:
Progresses: #2086 

> coordinate, verb, "to combine in harmonious relation or action."

Synchronizes the host's `provideWorker()`. In addition, coordinates the implementations of the host's `provideX` methods such that they all:
1. Attempt to get the existing formula id for the provided name, if any.
2. If there is no existing formula id, incarnate and return a new value.

All type checks during step 1 have been removed. In other words, if you attempt to provide something under an existing name that resolves to a different value type, you're on your own. We may revisit this decision in the future.

### DIFF
--- a/packages/daemon/src/host.js
+++ b/packages/daemon/src/host.js
@@ -125,27 +125,27 @@ export const makeHostMaker = ({
      * @param {string} workerName
      */
     const provideWorker = async workerName => {
-      if (typeof workerName !== 'string') {
-        throw new Error('worker name must be string');
-      }
-      let workerFormulaIdentifier = petStore.identifyLocal(workerName);
-      if (workerFormulaIdentifier === undefined) {
-        ({ formulaIdentifier: workerFormulaIdentifier } =
-          await incarnateWorker());
-        assertPetName(workerName);
-        await petStore.write(workerName, workerFormulaIdentifier);
-      } else if (!workerFormulaIdentifier.startsWith('worker:')) {
-        throw new Error(`Not a worker ${q(workerName)}`);
-      }
-      return /** @type {Promise<import('./types.js').EndoWorker>} */ (
-        // Behold, recursion:
-        // eslint-disable-next-line no-use-before-define
-        provideValueForFormulaIdentifier(workerFormulaIdentifier)
+      /** @type {import('./types.js').DeferredTasks<import('./types.js').WorkerDeferredTaskParams>} */
+      const tasks = makeDeferredTasks();
+      // eslint-disable-next-line no-use-before-define
+      const workerFormulaIdentifier = prepareWorkerFormulaIdentifier(
+        workerName,
+        tasks.push,
       );
+
+      if (workerFormulaIdentifier !== undefined) {
+        return /** @type {Promise<import('./types.js').EndoWorker>} */ (
+          // Behold, recursion:
+          provideValueForFormulaIdentifier(workerFormulaIdentifier)
+        );
+      }
+
+      const { value } = await incarnateWorker(tasks);
+      return value;
     };
 
     /**
-     * @param {string | 'MAIN' | 'NEW'} workerName
+     * @param {string} workerName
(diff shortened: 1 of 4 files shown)


packages/daemon/src/host.js
```diff
@@ -159,6 +159,11 @@ export const makeHostMaker = ({
         deferTask(identifiers =>
           petStore.write(workerName, identifiers.workerFormulaIdentifier),
         );
+      } else {
+        const { type } = parseId(workerFormulaIdentifier);
+        if (type !== undefined && type !== 'worker') {
+          throw new Error(`Not a worker ${q(workerName)}`);
+        }
```
rekmarks, line 166: Since we no longer store the formula type in the identifier, this refactor causes the test `'cannot spawn worker with existing non-worker name'` to fail. We can either mark that test as intentionally failing or attempt to address the matter here.

kriskowal: Your discretion. If the test doesn’t work, delete it. I’m not entirely convinced it’s a feature.

rekmarks: I think this needs to be resolved separately. I created #2147 and added a regression test for it in this PR.

rekmarks: Resolution for this PR: We should simply fail if `foo` is not a worker. Whether `provideWorker()` overwrites the existing name is a separate question. We can create a regression test for the latter for now.

kriskowal:
> > coordinate, verb, "to combine in harmonious relation or action."

and etymologically, “order with”.

packages/daemon/src/host.js
```diff
@@ -435,7 +415,7 @@ export const makeHostMaker = ({
      * @returns {Promise<{formulaIdentifier: string, value: Promise<import('./types.js').EndoGuest>}>}
      */
     const makeGuest = async (petName, { introducedNames = {} } = {}) => {
-      let guest = getNamedAgent('guest', petName);
+      let guest = getNamedAgent(petName);
```
kriskowal, line 418: I think I had to do this in #2148, so the merge conflict might resolve itself.

packages/daemon/test/test-endo.js
```diff
@@ -166,10 +167,15 @@ test('cannot spawn worker with existing non-worker name', async t => {
...
-    message: 'Not a worker "ten"',
+  const foo = await E(host).evaluate('MAIN', '10', [], [], 'foo');
+  t.is(foo, 10);
+
+  // This resolves with the existing value of 'foo' rather than overwriting it
+  // with a new worker.
+  await E(host).provideWorker('foo');
+  await t.throwsAsync(() => E(host).evaluate('foo', '20', [], [], 'bar'), {
+    message:
+      'Cannot deliver "evaluate" to target; typeof target is "undefined"',
```
kriskowal, line 178: This is almost certainly a symptom of there being no entry in the worker weak map, or the internal facet for this formula being undefined anyway. I regret the internal/external separation and look forward to eliminating it.

kriskowal (approved):
I’m not entirely sold on the deferred tasks pattern but I’m going to take ownership of coming up with something better. Good to go!

packages/daemon/src/host.js
```diff
@@ -256,15 +245,15 @@ export const makeHostMaker = ({
       /** @type {import('./types.js').DeferredTasks<import('./types.js').MakeCapletDeferredTaskParams>} */
       const tasks = makeDeferredTasks();
 
-      const workerFormulaIdentifier = prepareWorkerFormulaIdentifier(
-        workerName,
-        tasks.push,
-      );
+      const workerFormulaIdentifier = tryGetWorkerFormulaIdentifier(workerName);
+      prepareWorkerIncarnation(workerName, workerFormulaIdentifier, tasks.push);
```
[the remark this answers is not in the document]
rekmarks: 💀

PR #2140, opened 2024-03-13
chore: Drop Node.js 16 from CI entirely

kriskowal:
closes: #1046

## Description

This removes the last two vestiges of testing on Node.js 16. Depcheck was on 16 by omission. The remaining CI matrix tests were for preventing regressions in `async_hooks` handling, which are leaving firmly behind us. We may at this point remove the promise-kit accommodations for these earlier versions, but that is not in scope for this change.

### Security Considerations

none

### Scaling Considerations

none

We may need to document somewhere that we no longer test Node.js. We have elected not to use the `engines` field in `package.json` since that would be brittle.

### Testing Considerations

fewer

none

### Upgrade Considerations

none to my knowledge

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,9 +119,9 @@ jobs:
       fail-fast: false
       matrix:
         node-version:
-          - '16.1' # last version before some significant promise hooks changes
-          - '16.5' # last version before unconditional promise fast-path
-          - '16.6' # first version after unconditional promise fast-path
+          # - '16.1' # last version before some significant promise hooks changes
+          # - '16.5' # last version before unconditional promise fast-path
+          # - '16.6' # first version after unconditional promise fast-path
           - '18' # Active LTS
           # '20.6' not viable due to https://github.com/nodejs/node/issues/49497
           # '20.3' to '20.6' not viable due to https://github.com/nodejs/node/pull/49211
--- a/.github/workflows/depcheck.yml
+++ b/.github/workflows/depcheck.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: ['16.x']
+        node-version: ['18.x']
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -119,9 +119,9 @@ jobs:
...
         node-version:
-          - '16.1' # last version before some significant promise hooks changes
-          - '16.5' # last version before unconditional promise fast-path
-          - '16.6' # first version after unconditional promise fast-path
+          # - '16.1' # last version before some significant promise hooks changes
+          # - '16.5' # last version before unconditional promise fast-path
+          # - '16.6' # first version after unconditional promise fast-path
           - '18' # Active LTS
           # '20.6' not viable due to https://github.com/nodejs/node/issues/49497
           # '20.3' to '20.6' not viable due to https://github.com/nodejs/node/pull/49211
```
LuqiPan, line 127: _question_

Worth a revisit to these versions? Look like both issues are closed

.github/workflows/depcheck.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: ['16.x']
```
LuqiPan, line 16: _question_

Why not add node 20 here?

erights (commented):
FWIW, LGTM, but I'll leave the approval decision in @LuqiPan 's hands.

erights:
See also https://github.com/Agoric/agoric-sdk/pull/8365

.github/workflows/ci.yml
turadg: We only need one Node 20. It [turned out to be 20.9](https://nodejs.org/en/blog/release/v20.9.0) (not 20.7 that the line below enables)

.github/workflows/depcheck.yml
kriskowal: I would like to attack Node.js 20 in another change.

turadg: I can't tell, what is your opinion on running the cycles and depcheck in multiple versions of Node?

kriskowal: I agree that we only need to run these checks in one version of Node.js and we could drop the matrix.

LuqiPan (approved):
LGTM

LuqiPan (approved):
LGTM :white_check_mark:

PR #2142, opened 2024-03-13
refactor(daemon): Synchronize incarnateReadableBlob()

rekmarks:
Progresses: #2086 

Merges the daemon's `storeReaderRef()` into `incarnateReadableBlob()` and synchronizes the implementation. Also converts `makeReadableBlob()` into `makeControllerForReadableBlob()` to align it with the family of functions that already exist for most other formulas.

Contrary to its predecessor, `incarnateReadableBlob()` returns the value for stored value, namely a `FarEndoReadable`. A unit test is added for the case of immediately using a stored value without naming it.

### DIFF
--- a/packages/daemon/src/host.js
+++ b/packages/daemon/src/host.js
@@ -24,7 +24,7 @@ const assertPowersName = name => {
  * @param {import('./types.js').DaemonCore['incarnateEval']} args.incarnateEval
  * @param {import('./types.js').DaemonCore['incarnateUnconfined']} args.incarnateUnconfined
  * @param {import('./types.js').DaemonCore['incarnateBundle']} args.incarnateBundle
- * @param {import('./types.js').DaemonCore['storeReaderRef']} args.storeReaderRef
+ * @param {import('./types.js').DaemonCore['incarnateReadableBlob']} args.incarnateReadableBlob
  * @param {import('./types.js').DaemonCore['getAllNetworkAddresses']} args.getAllNetworkAddresses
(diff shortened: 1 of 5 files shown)


packages/daemon/src/host.js
```diff
@@ -107,15 +107,19 @@ export const makeHostMaker = ({
...
+        );
       }
 
-      const formulaIdentifier = await storeReaderRef(readerRef);
-
-      if (petName !== undefined) {
-        await petStore.write(petName, formulaIdentifier);
-      }
+      // TODO: Should we not return the value?
+      await incarnateReadableBlob(readerRef, tasks);
```
rekmarks, line 122: ATTN: Two questions in the form of TODOs here.

rekmarks: Resolution: Leave value optional, return live value.

packages/daemon/src/host.js
```diff
@@ -107,15 +107,17 @@ export const makeHostMaker = ({
...
+          petStore.write(petName, identifiers.readableBlobFormulaIdentifier),
+        );
       }
 
-      const formulaIdentifier = await storeReaderRef(readerRef);
-
-      if (petName !== undefined) {
-        await petStore.write(petName, formulaIdentifier);
-      }
+      return (await incarnateReadableBlob(readerRef, tasks)).value;
```
kriskowal, line 120: Easier to read with a named intermediate variable. Not necessary, but in general, we like to avoid deeply embedding `await` in expressions, in order to make its presence more obvious.

kriskowal (approved):

Issue #2125, opened 2024-03-08
daemon: Tests do not clean up after themselves


rekmarks:
Tests leave dangling Node processes and temporary files behind, definitively on failure (expected or otherwise), and potentially on success as well. We should rewrite the test suite such that cleanup will always successfully occur.

rekmarks:
We should try bumping `ava` to `^6`. Ref: https://github.com/avajs/ava/pull/3283

rekmarks:
I am still observing dangling Node processes on my system after the `ava@6` update. Going to try to fiddle with `ava`'s teardown APIs next.

PR #2144, opened 2024-03-14
chore: Upgrade to ava ^6

kriskowal:
Progresses: #2125


kriskowal:
> @turadg This is a wholesale upgrade to ava@^6 in endo and it looks great except there appears to be a bad interaction between `ses`, `execa`, and the new `ava`. I’m investigating this now.

Root cause of this is in `mimic-fn`, retained by `onetime`, retained by `execava`, which does a hasty `toString` patch which does not compose well with our “enablements” on `toString` (replacing the `value` with a `get` descriptor). I’ve pushed a patch and seek to get it upstream.

rekmarks (approved):
LGTM.

turadg (approved):
LGTM. Leaving to your discretion when to drop the patch (since your https://github.com/sindresorhus/mimic-function/pull/51 is released in https://www.npmjs.com/package/mimic-function/v/5.0.1 )

kriskowal:
> LGTM. Leaving to your discretion when to drop the patch (since your [sindresorhus/mimic-function#51](https://github.com/sindresorhus/mimic-function/pull/51) is released in https://www.npmjs.com/package/mimic-function/v/5.0.1 )

Wow, we’ve never gotten one of these upstreamed overnight.

PR #2150, opened 2024-03-16
refactor(daemon): Contract formulate and provide

kriskowal:
This change blesses `formulate` and `provide` with short names to convey their significance and reduce noise.


rekmarks (approved):
We hereby claim these verbs for the daemon.

PR #2151, opened 2024-03-16
refactor(daemon): Id means formula identifier

kriskowal:
We have decided that there is one meaning of “identifier”, that it is “formula identifier”, that it is what “identify” returns, and that we can consequently consistently abbreviate it as “Id” or “id”.

While we’re here, abbreviate `provideControllerForFormulaIdentifier` to `provideController`.


rekmarks (approved):
Fabulous.

PR #2154, opened 2024-03-18
fix(daemon): Handle server response error

rekmarks:
Completes our error handling logic for the daemon's web server following #2148.

### DIFF
--- a/packages/daemon/src/web-server-node-powers.js
+++ b/packages/daemon/src/web-server-node-powers.js
@@ -36,39 +36,56 @@ export const makeHttpPowers = ({ http, ws }) => {
     });
 
     if (respond) {
+      const sendResponse = async (req, res) => {
+        const response = await respond(
+          harden({
+            method: req.method,
+            url: req.url,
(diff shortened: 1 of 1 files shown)


packages/daemon/src/web-server-node-powers.js
```diff
@@ -36,39 +36,56 @@ export const makeHttpPowers = ({ http, ws }) => {
...
-            for await (const chunk of response.content) {
-              res.write(chunk);
-            }
-            res.end();
+
+          try {
+            await sendResponse(req, res);
+          } catch (_error) {
+            // TODO: Log this error locally.
+            tryRespondWithError(res);
```
rekmarks, line 88: Should we log something here?

rekmarks:
In draft awaiting #2155.

packages/daemon/src/web-server-node-powers.js
kriskowal: No, this is fine.

kriskowal (approved):

PR #2155, opened 2024-03-18
fix(daemon): Fix provision of new, unnamed workers

rekmarks:
Fixes the provision of new (i.e. `NEW`) unnamed workers after #2139 broke it. Adds a regression test for this case.

### DIFF
--- a/packages/daemon/src/host.js
+++ b/packages/daemon/src/host.js
@@ -128,9 +128,7 @@ export const makeHostMaker = ({
       /** @type {import('./types.js').DeferredTasks<import('./types.js').WorkerDeferredTaskParams>} */
       const tasks = makeDeferredTasks();
       // eslint-disable-next-line no-use-before-define
-      const workerId = tryGetWorkerId(workerName);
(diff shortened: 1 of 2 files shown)


kriskowal (commented):
Can has test?

packages/daemon/src/host.js
```diff
@@ -128,9 +128,7 @@ export const makeHostMaker = ({
       /** @type {import('./types.js').DeferredTasks<import('./types.js').WorkerDeferredTaskParams>} */
       const tasks = makeDeferredTasks();
       // eslint-disable-next-line no-use-before-define
-      const workerId = tryGetWorkerId(workerName);
-      // eslint-disable-next-line no-use-before-define
-      prepareWorkerFormulation(workerName, workerId, tasks.push);
+      const workerId = prepareWorkerFormulation(workerName, tasks.push);
```
kriskowal, line 131: An amusing twist of naming I didn’t catch in the big rename.

kriskowal (approved):

PR #2157, opened 2024-03-18
fix(ses-ava): revert typing "improvement"

erights:
closes: #XXXX
refs: #XXXX

### Upgrade Considerations


erights:
From recent comments by @kriskowal , it sounds like this is not needed and would not help. Closing.

PR #2160, opened 2024-03-19
feat(daemon): Add `locate()` facility

rekmarks:
Adds locator strings and a `locate()` method to the daemon's directory abstraction, and thereby hosts and guests. A locator string is a URL of the following form:
```text
endo://{nodeIdentifier}/?id={formulaNumber}&type={formulaType}
```

Formula identifiers are of the form `{formulaNumber}:{nodeIdentifier}`. Therefore, we can say that a locator is the URL form of the formula identifier, annotated with the `formulaType`. If a value is remote, i.e. belonging to a different peer, we set the `type` query param to `remote` to indicate this. `remote` is not a valid formula type in the general case—i.e. there are no formula objects with that type—but perhaps it should be.

Note that the `id` query param is just the `formulaNumber`, not the formula identifier. We may address this internal discrepancy in naming by using locators and formula numbers instead of the `{formulaNumber}:{nodeIdentifier}` construction.

### DIFF
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -183,18 +183,36 @@ const makeDaemonCore = async (
 
   /**
    * The two functions "formulate" and "provide" share a responsibility for
-   * maintaining the memoization tables "controllerForId" and
+   * maintaining the memoization tables "controllerForId", "typeForId", and
    * "idForRef".
    * "formulate" is used for creating and persisting new formulas, whereas
    * "provide" is used for "reincarnating" the values of stored formulas.
    */
 
   /**
-   * Reverse look-up, for answering "what is my name for this near or far
-   * reference", and not for "what is my name for this promise".
+   * Forward look-up, for answering "what is the value of this id".
    * @type {Map<string, import('./types.js').Controller>}
    */
   const controllerForId = new Map();
+
+  /**
+   * Forward look-up, for answering "what is the formula type of this id".
+   * @type {Map<string, string>}
+   */
+  const typeForId = new Map();
+
+  /** @param {string} id */
+  const getTypeForId = async id => {
+    await null;
+
+    const formulaType = typeForId.get(id);
+    if (formulaType !== undefined) {
+      return formulaType;
+    }
+
+    const formula = await persistencePowers.readFormula(parseId(id).number);
+    return formula.type;
+  };
 
   /**
    * Reverse look-up, for answering "what is my name for this near or far
@@ -667,14 +685,18 @@ const makeDaemonCore = async (
     if (isRemote) {
       // eslint-disable-next-line no-use-before-define
       const peerIdentifier = await getPeerIdForNodeIdentifier(formulaNode);
+      typeForId.set(id, 'remote');
       // Behold, forward reference:
       // eslint-disable-next-line no-use-before-define
       return provideRemoteValue(peerIdentifier, id);
     }
+
     const formula = await persistencePowers.readFormula(formulaNumber);
     console.log(`Making ${formula.type} ${formulaNumber}`);
     assertValidFormulaType(formula.type);
     // TODO further validation
+    typeForId.set(id, formula.type);
+
     return makeControllerForFormula(id, formulaNumber, formula, context);
   };
 
@@ -707,6 +729,7 @@ const makeDaemonCore = async (
       internal: E.get(partial).internal,
     });
     controllerForId.set(id, controller);
+    typeForId.set(id, formula.type);
 
     // The controller _must_ be constructed in the synchronous prelude of this function.
     const controllerValue = makeControllerForFormula(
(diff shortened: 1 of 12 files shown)


packages/daemon/src/daemon.js
```diff
@@ -182,19 +183,43 @@ const makeDaemonCore = async (
...
+   * maintaining the memoization tables "controllerForId", "typeForId", and
    * "idForRef".
    * "formulate" is used for creating and persisting new formulas, whereas
    * "provide" is used for "reincarnating" the values of stored formulas.
    */
 
   /**
-   * Reverse look-up, for answering "what is my name for this near or far
-   * reference", and not for "what is my name for this promise".
+   * Forward look-up, for answering "what is the value of this id".
```
kriskowal, line 193: Good fix.

packages/daemon/src/daemon.js
```diff
@@ -182,19 +183,43 @@ const makeDaemonCore = async (
...
    * @type {Map<string, import('./types.js').Controller>}
    */
   const controllerForId = new Map();
 
+  /**
+   * Forward look-up, for answering "what is the formula type of this id".
+   * @type {Map<string, string>}
+   */
```
kriskowal, line 201: Perhaps we can get a leg up on building a full formula memo here instead of just formula type, then use that anywhere we previously read formulas.

packages/daemon/src/daemon.js
```diff
@@ -670,35 +691,18 @@ const makeDaemonCore = async (
...
-        'handle',
-        'pet-inspector',
-        'pet-store',
-        'lookup',
-        'directory',
-      ].includes(formula.type)
-    ) {
-      assert.Fail`Invalid formula identifier, unrecognized type ${q(id)}`;
-    }
+    assertValidFormulaType(formula.type);
```
kriskowal, line 702: Nice.

packages/daemon/src/daemon.js
```diff
@@ -1454,6 +1451,33 @@ const makeDaemonCore = async (
...
+    };
+
+    return harden({
+      has: nodeIdentifier => petStore.has(getNameFor(nodeIdentifier)),
+      identify: nodeIdentifier =>
+        petStore.identifyLocal(getNameFor(nodeIdentifier)),
+      write: (nodeIdentifier, peerId) =>
+        petStore.write(getNameFor(nodeIdentifier), peerId),
+    });
+  };
```
kriskowal, line 1479: Another angle for this would be to introduce a `known-peers-store` formula type that is just like `pet-store` except that it is constructed with a different `assertValidName`. The same trick will probably be necessary for message persistence.

packages/daemon/src/directory.js
```diff
@@ -84,6 +87,17 @@ export const makeDirectoryMaker = ({
...
 
+    /** @type {import('./types.js').EndoDirectory['locate']} */
+    const locate = async (...petNamePath) => {
+      const id = await identify(...petNamePath);
+      if (id === undefined) {
+        return undefined;
+      }
+
+      const formulaType = await getTypeForId(id);
+      return formatLocator(id, formulaType);
```
kriskowal, line 98: Wherein the protagonist verifies that the L in URL is for Locator as opposed to Location.

packages/daemon/src/locator.js
```diff
@@ -0,0 +1,94 @@
...
+
+import { parseId, isValidNumber } from './formula-identifier.js';
+import { isValidFormulaType } from './formula-type.js';
+
+const { quote: q } = assert;
+
+/**
+ * The endo locator format:
+ * ```
+ * endo://{nodeIdentifier}/?number={formulaNumber}&type={formulaType}
```
kriskowal, line 11: Recommend
```suggestion
 * endo://{nodeIdentifier}/?id={formulaNumber}&type={formulaType}
```

packages/daemon/test/test-endo.js
```diff
@@ -1390,95 +1432,116 @@ test('read unknown nodeId', async t => {
...
+  const hostA = await makeHostWithTestNetwork(locatorA, cancelled);
+  const hostB = await makeHostWithTestNetwork(locatorB, cancelled);
+
   // introduce nodes to each other
   await E(hostA).addPeerInfo(await E(hostB).getPeerInfo());
   await E(hostB).addPeerInfo(await E(hostA).getPeerInfo());
 
   // create value to share
-  await E(hostB).evaluate('MAIN', '`haay wuurl`', [], [], 'salutations');
+  await E(hostB).evaluate('MAIN', '"hello, world!"', [], [], 'salutations');
```
kriskowal, line 1536: Noted.

kriskowal (approved):

packages/daemon/src/daemon.js
rekmarks: #2165

packages/daemon/src/daemon.js
rekmarks: #2166

PR #2162, opened 2024-03-19
chore: Release 2024-03-16

kriskowal:
Regular releases. This release includes improvements to ses-ava and pass-style. SES+Ava can now capture unredacted traces in the test-specific log (see the notes for how to adapt your tests for this feature) and we now have a feature flag in pass-style for testing the performance impact of constraining strings to well-formed Unicode.


packages/bundle-source/CHANGELOG.md
```diff
@@ -3,6 +3,22 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [3.2.0](https://github.com/endojs/endo/compare/@endo/bundle-source@3.1.0...@endo/bundle-source@3.2.0) (2024-03-19)
+
+
+### Features
+
+* **ses-ava:** import test from @endo/ses-ava/prepare-endo.js ([#2133](https://github.com/endojs/endo/issues/2133)) ([9d3a7ce](https://github.com/endojs/endo/commit/9d3a7ce150b6fd6fe7c8c4cc43da411e981731ac))
```
turadg, line 11: not a consumer-facing feature

turadg (commented):
I think the commit history is misleading as to what's a new feature. Does Lerna's history-by-commit feature support ignoring one? 

https://github.com/endojs/endo/commit/9d3a7ce150b6fd6fe7c8c4cc43da411e981731ac was a feature in ses-ava but not in the packages that were updated to use it. (I regret not reviewing the commit messages in https://github.com/endojs/endo/pull/2133, though it was squashed anyway)

packages/bundle-source/CHANGELOG.md
kriskowal: I think this is a user-facing feature, in that the user needs to do a minor refactor to take advantage of this new public export. Once they have done this refactor, no prior version will suffice.

kriskowal:
> I think the commit history is misleading as to what's a new feature. Does Lerna's history-by-commit feature support ignoring one?
> 
> [9d3a7ce](https://github.com/endojs/endo/commit/9d3a7ce150b6fd6fe7c8c4cc43da411e981731ac) was a feature in ses-ava but not in the packages that were updated to use it. (I regret not reviewing the commit messages in #2133, though it was squashed anyway)

Lerna conventional commits aren’t particularly flexible. There’s no intermediate step in the publish workflow where I can edit the generated changelogs or propose a different version number than the one it infers from commit history. These are reasons I’m intrigued by changesets, though I’m sympathetic to @dckc’s case that this means that detailed commit messages may be redundant with the curated copy that arrives in the changelog.

packages/bundle-source/CHANGELOG.md
turadg: It's user-facing for ses-ava but not for bundle-source. There are no code changes in bundle-source. It should have a message like base64 has:
> **Note:** Version bump only for package @endo/base64

But maybe this comment is superseded by your PR comment.

turadg (approved):
> There’s no intermediate step in the publish workflow where I can edit the generated changelogs or propose a different version number than the one it infers from commit history.

I take it you find the version and changelog impact acceptable. You're the repo owner so that's your call. I approve because I don't see anything problematic besides that.

> These are reasons I’m intrigued by changesets, though I’m sympathetic to @dckc’s case that this means that detailed commit messages may be redundant with the curated copy that arrives in the changelog.

I think that would be solved by tooling. To me the inability to correct change history (demonstrated here) makes logs generated by commit messages untenable.

gibson042 (approved):
I think @turadg has a point, but ultimately that's something to address with tooling changes. The details here look good to me in the current paradigm.

PR #2170, opened 2024-03-22
test(daemon): Refactor daemon test suite

### LINKED ISSUE #2125: daemon: Tests do not clean up after themselves

rekmarks:
Fixes #2125

Refactors `test-endo.js` by introducing utilities to absorb the necessary boilerplate to get from an empty state to a reference to a live endo host. Also addresses #2125 by ensuring that test teardown will always run, even in the case of test failures. There should no longer be any dangling node processes after running `yarn test`, and it may even be possible to forego the use of `test:clean` and its `rm -rf /tmp` prelude in the general case.

### DIFF
--- a/packages/daemon/index.js
+++ b/packages/daemon/index.js
@@ -149,15 +149,14 @@ export const clean = async (locator = defaultLocator) => {
   }
 };
 
-export const restart = async (locator = defaultLocator) => {
-  await terminate(locator).catch(() => {});
-  await clean(locator);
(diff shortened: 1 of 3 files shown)


kriskowal (approved):
🏜️

PR #2171, opened 2024-03-23
refactor(daemon): Memoize formulas

### LINKED ISSUE #2165: daemon: Memoize entire formulas instead of just their type
Following #2160, we should memoize entire formulas as opposed to merely their type in the memoization table introduced thereby.

kriskowal:
Closes #2165

Once a formula has been written and until it’s provably unreachable, we need not consult storage.


rekmarks (approved):
You love to see it!

PR #2179, opened 2024-03-25
refactor(daemon): Add known-peers-store formula

### LINKED ISSUE #2166: daemon: Add `known-peers-store` formula type
Following #2160, we should create a dedicated formula type for the known peers store, replacing the "pet store in a trench coat" approach used in #2160.

rekmarks:
Closes #2166 

Replace the known peers store adapter with a dedicated `known-peers-store` formula type, instantiated the same way as other "special" formulas such as `least-authority`. The adapter used a `pet-store` formula under the hood. Giving it a different formula entirely is probably less likely to confuse us in the future.

### DIFF
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -15,7 +15,7 @@ import { makeGuestMaker } from './guest.js';
 import { makeHostMaker } from './host.js';
 import { assertPetName } from './pet-name.js';
 import { makeContextMaker } from './context.js';
-import { parseId, formatId } from './formula-identifier.js';
+import { assertValidNumber, parseId, formatId } from './formula-identifier.js';
 import { makeSerialJobs } from './serial-jobs.js';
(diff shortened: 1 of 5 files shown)


packages/daemon/src/daemon.js
```diff
@@ -687,9 +691,19 @@ const makeDaemonCore = async (
...
         assertPetName,
       );
       return { external, internal: undefined };
+    } else if (formula.type === 'known-peers-store') {
+      const external = petStorePowers.makeIdentifiedPetStore(
+        formulaNumber,
+        'known-peers-store',
+        // The known peers store is just a pet store that only accepts node identifiers
+        // (i.e. formula numbers) as "names".
+        assertValidNumber,
```
kriskowal, line 704: # 🪄

kriskowal (approved):
Pending CI, LGTM.

PR #2181, opened 2024-03-26
docs: demote some packages from developer API product

dckc:
refs: https://github.com/Agoric/documentation/issues/1031

In https://github.com/Agoric/documentation/issues/1031 , we aim to define "the developer API product" and to supply corresponding reference docs.

This is a proposal to exclude whole packages from consideration, leaving them as implementation details or some such.

@sufyaankhan 

### Security / Scaling / Testing / Upgrade Considerations

n/a

This could lower some compatibility burdens.

### DIFF
--- a/typedoc.json
+++ b/typedoc.json
@@ -3,6 +3,19 @@
         "packages/*"
     ],
     "exclude": [
+        "packages/cjs-module-analyzer",
+        "packages/compartment-mapper",
+        "packages/cli",
+        "packages/daemon",
+        "packages/evasive-transform",
+        "packages/lp32",
+        "packages/memoize",
+        "packages/netstring",
+        "packages/syrup",
+        "packages/skel",
+        "packages/static-module-record",
+        "packages/stream-types-test",
+        "packages/where",
         "packages/test262-runner/"
     ],
     "name": "Endo API documentation",


erights:
The windows CI test seems to be flaxy, so I restarted your failed one.

typedoc.json
```diff
@@ -3,6 +3,20 @@
         "packages/*"
     ],
     "exclude": [
+        "packages/cjs-module-analyzer",
+        "packages/compartment-mapper",
+        "packages/daemon",
+        "packages/env-options",
```
erights, line 9: um

typedoc.json
```diff
@@ -3,6 +3,20 @@
...
+        "packages/errors",
```
erights, line 10: um

erights (approved):
All the exclusions except `env-options` and `errors` LGTM. But these two, especially `errors` is something that IMO should be exposed and explained to developers. At least `errors`.

typedoc.json
dckc: https://endojs.github.io/endo/modules/_endo_errors.html is blank.

I took that to mean that there wasn't anything anybody wanted to say to a larger audience.

typedoc.json
dckc: My impression is that env-options is an implementation detail. It's useful for _building_ endo, but not so much for _using_ it. Not so?

erights: The only issue is if we want to empower the user to develop with conditionals on environment variables they define. If so, we should expose this. Certainly, we don't need to expose it immediately, as none of their use of anything else hinges on their ability to use the env-options API. But I think it is a nice development convenience. 

(Note: until recently I hated the notion because environment variables leak across boundaries in a security hostile manner. So if we do encourage, we should be clear that it is for development-time uses, like switching instrumentation on or off, or changing verbosity level, or...)

typedoc.json
erights: Sigh. We do need to explain this API. Searching for `assert` in `*.md`, a lot of its mentions are in https://github.com/endojs/endo/blob/master/packages/ses/src/error/README.md and https://github.com/endojs/endo/blob/master/packages/ses/docs/lockdown.md . 

There are also the doc-comments in https://github.com/endojs/endo/blob/master/packages/ses/src/error/types.js and https://github.com/endojs/endo/blob/master/packages/ses/src/error/assert.js . Not sure how useful these are as starting points though.

erights: On env-options above, I make the point

> Certainly, we don't need to expose it immediately, as none of their use of anything else hinges on their ability to use the env-options API. But I think it is a nice development convenience.

I must reluctantly observe that this is true for `errors` too. So if you want to exclude both at this stage, the PR would still LGTM. Your choice.

dckc: The docs for the ses `assert` are... not straightforward to deal with.

https://github.com/endojs/endo/blob/master/packages/ses/src/error/types.js has all sorts of detailed doc on stuff like `DetailsTag`. But ... it's all clobbered by https://github.com/endojs/endo/blob/master/packages/ses/types.d.ts

erights: That clobbering is interesting. Does that mean that, in general, when both files exist none of the doc-comments in the `types.js` file will make it to the typedoc?

dckc: evidently so.

escalated / reified:

 - #2183

typedoc.json
```diff
@@ -3,6 +3,20 @@
         "packages/*"
     ],
     "exclude": [
+        "packages/cjs-module-analyzer",
+        "packages/compartment-mapper",
+        "packages/daemon",
```
dckc, line 8: @kriskowal I'm suggesting / proposing that the daemon package is an implementation detail for the cli package. I'd like to know your take on it. Does that seem right? Or are there docs in the daemon package intended for users / devs?

kriskowal: The daemon is definitely an implementation detail of something else and the CLI is the only example so far. However, the daemon does provide the API for caplets.

dckc: I don't see anything about caplets on https://endojs.github.io/endo/modules/_endo_daemon.html
I guess its `tsdoc.json` needs some love?

dckc: I don't see docs on an API for caplets in this package.

https://github.com/endojs/endo/tree/master/packages/cli/demo is the only docs I've seen for the endo daemon. I'm struggling to get `typedoc` to include it in the output. @turadg , clues?

turadg: add an `entrypoint` in the typedoc.json for the package

dckc: I should have mentioned that I tried that; got:

```
[warning] The entry point ./packages/cli/demo/README.md is not referenced by the 'files' or 'include' option in your tsconfig.
```

tried adding that to the `include` key of `packages/cli/tsconfig.json` but the symptoms persist.

kriskowal: More important: Neither daemon nor cli are stable and published, so it would be good to withhold docs for now.

kriskowal (approved):

typedoc.json
dckc: I took env-options out of the exclude list. It's not really hurting anything.

typedoc.json
dckc: took errors out too

PR #2184, opened 2024-03-26
feat(daemon,cli): One handle for every agent

kriskowal:
Toward sending messages between mailboxes in separate nodes, this change gives every agent (both hosts and guests) a powerless handle that can be used to mention the agent without leaking agency.

This is somewhat tricky because this is the first cycle in the formula graph. The agent _contains_ a handle and the handle identifier and formula are just ways of referring to the facet held by the agent. So, writing an agent formula generates a handle formula as a side-effect and populates the agent id for handle table.

This also means that `mkguest` and `mkhost` need to designate the handle (so that messages from the other agent are visible) and might optionally also designate a pet name for the agent.

### DIFF
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -558,11 +558,11 @@ const makeDaemonCore = async (
         context,
       );
     } else if (formula.type === 'handle') {
-      context.thisDiesIfThatDies(formula.target);
+      context.thisDiesIfThatDies(formula.agent);
       return {
         external: {},
         internal: {
-          targetId: formula.target,
+          agentId: formula.agent,
         },
       };
     } else if (formula.type === 'endo') {
@@ -885,13 +885,13 @@ const makeDaemonCore = async (
         return controller;
       }
       // @ts-expect-error We can't know the type of the internal facet.
-      if (internalFacet.targetId === undefined) {
+      if (internalFacet.agentId === undefined) {
         return controller;
       }
       const handle = /** @type {import('./types.js').InternalHandle} */ (
         internalFacet
       );
-      currentId = handle.targetId;
+      currentId = handle.agentId;
     }
   };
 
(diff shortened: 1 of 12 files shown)


packages/daemon/src/daemon.js
```diff
@@ -875,24 +892,12 @@ const makeDaemonCore = async (
...
-        return controller;
-      }
-      // @ts-expect-error We can't know the type of the internal facet.
-      if (internalFacet.targetId === undefined) {
-        return controller;
-      }
-      const handle = /** @type {import('./types.js').InternalHandle} */ (
-        internalFacet
-      );
-      currentId = handle.targetId;
```
kriskowal, line 894: cha-ching

packages/daemon/src/types.d.ts
```diff
@@ -725,6 +726,9 @@ export type DeferredTasks<T extends Record<string, string | string[]>> = {
 
 type FormulateNumberedGuestParams = {
   guestFormulaNumber: string;
+  handleId: string;
+  guestId: string;
```
rekmarks, line 730: For consistency with `hostAgentId` and `hostHandleId` in the same interface:
```suggestion
  guestHandleId: string;
  guestAgentId: string;
```

Would also accept `agentId` / `handleId`.

packages/daemon/src/types.d.ts
```diff
@@ -738,6 +742,8 @@ type FormulateHostDependenciesParams = {
 
 type FormulateNumberedHostParams = {
   hostFormulaNumber: string;
+  hostId: string;
+  handleId: string;
```
rekmarks, line 746: See comment on `FormulateNumberedGuestParams` above.

packages/daemon/src/daemon.js
```diff
@@ -759,7 +776,7 @@ const makeDaemonCore = async (
     formulaForId.set(id, formula);
 
     // Memoize for lookup.
-    console.log(`Making ${id}`);
+    console.log(`Making ${formula.type} ${id}`);
```
rekmarks, line 779: 👌

rekmarks (changes requested):
I have some optional suggestions in here, but the familiar chat example is broken on this branch. No names are listed, the console is full of errors, and mailbox requests do not appear. Also, the instructions in `cli/demo/README.md` are not update (see e.g. `endo mkguest`).

packages/daemon/src/types.d.ts
kriskowal: I would like to keep these as present, or instead `ownHandleId` and `ownAgentId`. The role of `handleId` and `agentId` is the same between Host and Guest so I would like to use the same names. Whereas a guest has a host, but a host does not. (I would imagine adding `PARENT` for hosts created by other hosts. If we explored that path, we’d have to consider how to accommodate the specialness of the root host. We could borrow the `/..` == `/` trick from *ix file systems.)

kriskowal:
@rekmarks This is ready for another review. I’ve appended a fixup commit with the necessary changes to the demo script to get it on the right path again.

packages/daemon/src/types.d.ts
rekmarks: That makes sense. I'd prefer the more explicit `ownHandleId` / `ownAgentId` in that case, but I leave it up to you.

rekmarks (approved):
LGTM. The remaining optional suggestion left as an exercise to the author.

packages/daemon/src/types.d.ts
kriskowal: Sure. I will circle back for this.

PR #2185, opened 2024-03-26
927 typedoc markdown

LuqiPan:
closes: #XXXX
refs: Agoric/documentation#927

## Description

- add `typedoc-plugin-markdown` to dev dependencies and add a `typedoc-markdown` script target to convert docs to markdown files
- fix a raw <T> tag in comment which will get converted into <T> in `.md` file and as a result confuse vitepress in Agoric/documentation#1038
    - to verify this indeed resolves the issue, I manually copied the generated `.md` file into documentation repo, added quotes and verified that vitepress could process the `.md` file successfully

This is one of the blockers for Agoric/documentation#927

### Upgrade Considerations


packages/patterns/src/types.js
```diff
@@ -449,7 +449,7 @@ export {};
  *
  * @property {(subPatt: Pattern) => Pattern} eref
  * Matches any Passable that is either matched by `subPatt` or is a promise object.
- * Note that validation is immediate, so (unlike the TypeScript ERef<T>
+ * Note that validation is immediate, so (unlike the TypeScript `ERef<T>`
```
LuqiPan, line 452: <T> if left unquoted, will be confused with a HTML tag by vitepress, and will throw an error that looks like `SyntaxError: Element is missing end tag.`

We can see a similar error being thrown for something else in agoric-sdk repo before vitepress can process this file here: https://github.com/Agoric/documentation/actions/runs/8442634353/job/23124473938?pr=1038

erights (commented):
FWIW LGTM, but leaving it to @dckc or @turadg to approve or request.

package.json
```diff
@@ -31,6 +32,7 @@
     "cover": "lerna run cover",
     "depcheck": "node scripts/check-mismatched-dependencies.js",
     "docs": "typedoc",
+    "typedoc-markdown": "typedoc --plugin typedoc-plugin-markdown",
```
turadg, line 35: When would this be run? 

Does it need to run at different times than the existing `docs` script?

Should it be in CI to prevent errors like the one you find in `types.js`?

LuqiPan: This would actually be run in Agoric/documentation repo, which then publish the artifacts to docs.agoric.com via Cloudflare pages.

There should be CI to prevent errors like this, the challenge is that running `yarn typedoc-markdown` itself wouldn't surface the error, it's when we convert `.md` files to html, css, etc using vitepress in `Agoric/documentation` repo where errors will surface. I've summarized where we're at now and what still need to be done in https://github.com/Agoric/documentation/issues/1046

kriskowal (approved):

package.json
kriskowal: I’d be happy if we simulate the whole workflow in our CI so we can see the markdown errors before we merge them.

LuqiPan:
Hmm, just noticed that I'm missing permission to merge PRs in `endojs` org...
Screenshot 2024-03-26 at 5 05 51 PM

package.json
turadg: > This would actually be run in Agoric/documentation repo

Seems like an important thing to document in this repo. Perhaps, `docs:markdown-for-documentation-repo`

turadg (approved):

.github/workflows/ci.yml
```diff
@@ -64,6 +64,10 @@ jobs:
       - name: build API docs
         run: yarn docs
 
+      # build the API docs in markdown for agoric/documentation repo to verify it works
+      - name: build API docs in markdown
+        run: yarn docs-markdown-for-agoric-documentation-repo
```
turadg, line 69:
```suggestion
        run: yarn docs:markdown-for-agoric-documentation-repo
```

turadg, line 69:
```suggestion
        run: yarn docs:markdown-for-agoric-documentation-repo
```

PR #2186, opened 2024-03-27
refactor(daemon): Move mail delivery to agent public facets

kriskowal:
Toward delivering mail over the network, this change introduces a `deliver` method on guest and host agents that accepts arbitrary messages into the agent’s inbox. Any other agent can send mail to any other local agent for whom they have a handle and cannot spoof their sender handle.

This is a net simplification, since it reduces our dependence on the internal facet of agents and requires less machinery because it treats messages as capabilities on the wire, freely exposing and carrying identifiers. We no longer have separate internal and external representations of messages and pet name “dubbing” is deferred to the UI. This in turn enables us to defer to the UI the question of whether to query the name for each identifier once or watch for changes.

Stacked on #2184

### DIFF
--- a/packages/cli/src/commands/show.js
+++ b/packages/cli/src/commands/show.js
@@ -5,6 +5,6 @@ import { withEndoAgent } from '../context.js';
 
 export const show = async ({ name, agentNames }) =>
   withEndoAgent(agentNames, { os, process }, async ({ agent }) => {
-    const pet = await E(agent).lookup(name);
+    const pet = await E(agent).lookup(...name.split('.'));
     console.log(pet);
   });
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -890,8 +890,8 @@ const makeDaemonCore = async (
     });
   };
 
-  /** @type {import('./types.js').DaemonCore['provideControllerAndResolveHandle']} */
-  const provideControllerAndResolveHandle = async id => {
+  /** @type {import('./types.js').DaemonCore['provideAgentControllerForHandleId']} */
+  const provideAgentControllerForHandleId = async id => {
     const handle = /** @type {{}} */ (await provide(id));
     const agentId = agentIdForHandle.get(handle);
     if (agentId === undefined) {
@@ -1580,12 +1580,12 @@ const makeDaemonCore = async (
 
   const makeMailbox = makeMailboxMaker({
     provide,
-    provideControllerAndResolveHandle,
+    provideAgentControllerForHandleId,
   });
 
   const makeIdentifiedGuestController = makeGuestMaker({
     provide,
-    provideControllerAndResolveHandle,
+    provideAgentControllerForHandleId,
     makeMailbox,
     makeDirectoryNode,
   });
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -890,18 +890,6 @@ const makeDaemonCore = async (
     });
   };
 
-  /** @type {import('./types.js').DaemonCore['provideAgentForHandle']} */
-  const provideAgentForHandle = async id => {
-    const handle = /** @type {{}} */ (await provide(id));
-    const agentId = agentIdForHandle.get(handle);
-    if (agentId === undefined) {
-      throw assert.error(assert.details`No agent for handle ${id}`);
-    }
-    return /** @type {Promise<import('./types.js').EndoAgent>} */ (
-      provide(agentId)
-    );
-  };
-
   /** @type {import('./types.js').DaemonCore['formulateReadableBlob']} */
   const formulateReadableBlob = async (readerRef, deferredTasks) => {
     const { formulaNumber, contentSha512 } = await formulaGraphJobs.enqueue(
@@ -1580,10 +1568,7 @@ const makeDaemonCore = async (
     formulateDirectory,
   });
 
-  const makeMailbox = makeMailboxMaker({
-    provide,
-    provideAgentForHandle,
-  });
+  const makeMailbox = makeMailboxMaker({ provide });
 
   const makeIdentifiedGuestController = makeGuestMaker({
     provide,
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -890,14 +890,16 @@ const makeDaemonCore = async (
     });
   };
 
-  /** @type {import('./types.js').DaemonCore['provideAgentControllerForHandleId']} */
-  const provideAgentControllerForHandleId = async id => {
+  /** @type {import('./types.js').DaemonCore['provideAgentForHandle']} */
+  const provideAgentForHandle = async id => {
     const handle = /** @type {{}} */ (await provide(id));
     const agentId = agentIdForHandle.get(handle);
     if (agentId === undefined) {
       throw assert.error(assert.details`No agent for handle ${id}`);
     }
-    return provideController(agentId);
+    return /** @type {Promise<import('./types.js').EndoAgent>} */ (
+      provide(agentId)
+    );
   };
 
   /** @type {import('./types.js').DaemonCore['formulateReadableBlob']} */
@@ -1580,7 +1582,7 @@ const makeDaemonCore = async (
 
   const makeMailbox = makeMailboxMaker({
     provide,
-    provideAgentControllerForHandleId,
+    provideAgentForHandle,
   });
 
   const makeIdentifiedGuestController = makeGuestMaker({
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -1585,7 +1585,6 @@ const makeDaemonCore = async (
 
   const makeIdentifiedGuestController = makeGuestMaker({
     provide,
-    provideAgentControllerForHandleId,
     makeMailbox,
     makeDirectoryNode,
   });
--- a/packages/daemon/src/guest.js
+++ b/packages/daemon/src/guest.js
@@ -8,13 +8,13 @@ import { makePetSitter } from './pet-sitter.js';
 /**
  * @param {object} args
  * @param {import('./types.js').DaemonCore['provide']} args.provide
- * @param {import('./types.js').DaemonCore['provideControllerAndResolveHandle']} args.provideControllerAndResolveHandle
+ * @param {import('./types.js').DaemonCore['provideAgentControllerForHandleId']} args.provideAgentControllerForHandleId
  * @param {import('./types.js').MakeMailbox} args.makeMailbox
  * @param {import('./types.js').MakeDirectoryNode} args.makeDirectoryNode
  */
 export const makeGuestMaker = ({
   provide,
-  provideControllerAndResolveHandle,
+  provideAgentControllerForHandleId,
   makeMailbox,
   makeDirectoryNode,
 }) => {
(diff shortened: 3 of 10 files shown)


packages/cli/src/commands/show.js
```diff
@@ -5,6 +5,6 @@ import { withEndoAgent } from '../context.js';
 
 export const show = async ({ name, agentNames }) =>
   withEndoAgent(agentNames, { os, process }, async ({ agent }) => {
-    const pet = await E(agent).lookup(name);
+    const pet = await E(agent).lookup(...name.split('.'));
```
kriskowal, line 8: Oh, by the way™

packages/daemon/src/guest.js
```diff
@@ -49,16 +43,6 @@ export const makeGuestMaker = ({
...
-    const hostController =
-      /** @type {import('./types.js').EndoHostController} */
-      (await provideControllerAndResolveHandle(hostHandleId));
-    const hostPrivateFacet = await hostController.internal;
-    const { respond: deliverToHost } = hostPrivateFacet;
-    if (deliverToHost === undefined) {
-      throw new Error(
-        `panic: a host request function must exist for every host`,
-      );
-    }
```
kriskowal, line 61: Dead code that was just laying around demanding to be refactored. It got the shovel instead.

packages/daemon/src/guest.js
```diff
@@ -68,6 +52,7 @@ export const makeGuestMaker = ({
     const { petStore } = mailbox;
     const directory = makeDirectoryNode(petStore);
 
+    const { reverseIdentify } = specialStore;
```
kriskowal, line 55: Incidentally, I needed the base case. I hope this doesn’t overlap too much with `followReverseLocate`. I am in favor of consolidating this into `reverseLocate` eventually.

packages/daemon/src/mail.js
```diff
@@ -1,174 +1,112 @@
...
+      {},
+      {
+        defaultGuards: 'passable',
+      },
+    ),
+    {
+      respondId: resolve,
+    },
+  );
+  const request = {
```
kriskowal, line 37: Should harden.

packages/daemon/src/mail.js
```diff
@@ -1,174 +1,112 @@
...
-        const [recipientName] = petStore.reverseIdentify(recipientId);
-        if (senderName !== undefined) {
-          return { who: senderName, dest: recipientName, ...rest };
-        }
-        return undefined;
-      }
-      throw new Error(
-        `panic: Unknown message type ${/** @type {any} */ (message).type}`,
-      );
-    };
```
kriskowal, line 60: This will not be missed.

packages/daemon/src/mail.js
```diff
@@ -1,174 +1,112 @@
...
-        }
-      }
-      for await (const message of subsequentRequests) {
-        const dubbedMessage = dubMessage(message);
-        if (dubbedMessage !== undefined) {
-          yield dubbedMessage;
-        }
-      }
+      yield* messages.values();
+      yield* subsequentRequests;
```
kriskowal, line 71: This makes the pattern much more obvious.

packages/daemon/src/mail.js
```diff
@@ -180,69 +118,43 @@ export const makeMailboxMaker = ({
...
-    ) => {
-      deliver({
-        type: /** @type {const} */ ('package'),
-        strings,
-        names: edgeNames,
-        formulas: ids,
-        who: senderId,
-        dest: receiverId,
-      });
-    };
```
kriskowal, line 226: Subsumed `receive` into `deliver`. It wasn’t doing much and `deliver` serves for `send` as well.

packages/daemon/src/mail.js
```diff
@@ -180,69 +118,43 @@ export const makeMailboxMaker = ({
...
-        recipientId,
-      );
-      const recipientInternal = await recipientController.internal;
-      if (recipientInternal === undefined || recipientInternal === null) {
-        throw new Error(`Recipient cannot receive messages: ${recipientName}`);
-      }
-      // @ts-expect-error We check if its undefined immediately after
-      const { receive: agentReceive } = recipientInternal;
-      if (agentReceive === undefined) {
-        throw new Error(`Recipient cannot receive messages: ${recipientName}`);
```
kriskowal, line 244: This code will not be missed.

packages/daemon/src/mail.js
```diff
@@ -266,17 +178,20 @@ export const makeMailboxMaker = ({
         }
         return id;
       });
-      // add to recipient mailbox
-      agentReceive(selfId, strings, edgeNames, ids);
-      // add to own mailbox
-      receive(
-        selfId,
+
+      const message = {
```
kriskowal, line 182: Should harden.

packages/daemon/src/mail.js
```diff
@@ -1,174 +1,112 @@
...
-        // eslint-disable-next-line no-use-before-define
-        return provide(id);
-      }
-      // The reference is not named nor to be named.
-      const id = await requestId(what, senderId, recipientId);
-      // TODO:
-      // context.thisDiesIfThatDies(id);
-      // Behold, recursion:
-      // eslint-disable-next-line no-use-before-define
-      return provide(id);
```
kriskowal, line 171: The fact that the recipient needed the sender’s pet store diffused a pretty awful smell. The pet store wasn’t even transmissible (not a Far/Exo). The `request` function uses `requestId` directly now and handles all the pet name preservation.

packages/daemon/src/mail.js
```diff
@@ -331,93 +246,47 @@ export const makeMailboxMaker = ({
...
+      if (responseName !== undefined) {
+        await petStore.write(responseName, responseId);
       }
-    };
-
-    /** @type {import('./types.js').PetStore['remove']} */
-    const remove = async petName => {
-      await petStore.remove(petName);
-      responses.delete(petName);
-    };
```
kriskowal, line 405: In the new framing, the `responses` memo turns out to be duplicative. Since we all see the same identifiers now, we can lean on the memo beneath `provide`. Because we don’t need to coördinate caches, the mailbox no longer needs to intermediate these pet store methods. Concerns: separated.

packages/daemon/src/types.d.ts
```diff
@@ -232,35 +232,41 @@ export type Specials = {
...
+}
 
-export type Message = Label & Payload;
-export type InternalMessage = InternalLabel & InternalPayload;
+export type StampedMessage = EnvelopedMessage & {
+  number: number;
+  date: string;
+  dismissed: Promise<void>;
+  dismisser: ERef<Dismisser>;
+};
```
kriskowal, line 269: These are the new degrees of annotation for messages. Message (when proposed) inside an Envelope (on the wire) with a Stamp (in an inbox).

kriskowal:
Just added a change to get Familiar Chat back into working order, since this changes the public type of messages. Now, both the CLI and WUI use `reverseIdentify` to translate the identifier-on-the-wire to their agent’s pet names. The WUI will need a follow-up to utilize `followLocatorNames`.

packages/daemon/src/mail.js
rekmarks: ibid.

packages/daemon/src/daemon.js
```diff
@@ -927,22 +934,31 @@ const makeDaemonCore = async (
...
     };
-    return /** @type {import('./types').FormulateResult<import('./types').ExternalHandle>} */ (
-      formulate(formulaNumber, formula)
-    );
+    await persistencePowers.writeFormula(formulaNumber, formula);
+    const id = formatId({
+      number: formulaNumber,
+      node: ownNodeIdentifier,
+    });
+    formulaForId.set(id, formula);
```
rekmarks, line 960: We should perhaps extract the first ~10 lines of `formulate()` into a new function and replace this inlined formula graph mutation with that.

I don't believe it matters in our usage of `formulateNumberedHandle()`, but `formulate()` also does not call `writeFormula()` until after mutating `formulaForId`.

rekmarks (commented):
Big fan of this PR 👍 

Some optional suggestions in my first pass. Conflicts need to be resolved, then I'll come back for a second and likely final look.

packages/daemon/src/mail.js
kriskowal: Breadcrumb.

packages/daemon/src/daemon.js
kriskowal: This is interesting and I’d like to revisit it. There might be a case for awaiting on waitFormula before this. It might help to test writeFormula failures and recovery from those failures on restart.

kriskowal: https://github.com/endojs/endo/issues/2207

kriskowal:
> Big fan of this PR 👍
> 
> Some optional suggestions in my first pass. Conflicts need to be resolved, then I'll come back for a second and likely final look.

Good for another pass. I’ve responded to your feedback (and my breadcrumbs) and also appended one commit that has turned out to be necessary: [refactor(daemon): Handles are mail delivery facets](https://github.com/endojs/endo/pull/2186/commits/4fe370edeed0bd12d961bab2c52324089941fe33). This is a pretty big deal. Handles become public capabilities to both deliver mail and also to verify the authenticity of received mail. I used a sealer/unsealer pattern.

In testing this, I discovered that mail delivery over the network will require us to ensure that we only have one live CapTP session between any pair of nodes so we can do grant-matching (so that an envelope that takes a round trip through CapTP comes back as the identical reference to the one sent). To do that, we need to negotiate crossed hellos (so one node knows to drop either its outgoing or incoming connection if nodes attempt to connect to each other concurrently). And once we do that, we need to have application level key pair cryptography to ensure an attacker can’t induce a node to drop a legitimate connection or impersonate another node’s handles.

That is to say, I had hoped that this would be as much scope as we needed to take for a demo. We will have to do at least grant-matching for a demo. We’ll have to do cryptography to be standing on a firm security model, even if we have redundant cryptography relying on Tor or TLS for confidentiality.

packages/daemon/src/types.d.ts
```diff
@@ -444,20 +444,8 @@ export interface Mail {
...
-  ): Promise<unknown>;
-  receive(
-    senderId: string,
-    strings: Array<string>,
-    edgeNames: Array<string>,
-    ids: Array<string>,
-    receiverId: string,
-  ): void;
+  // TODO attenuate
+  deliver(message: EnvelopedMessage): void;
```
rekmarks, line 438: Attenuate in what sense? I'm not sure what this means in the context of the type definition.

packages/daemon/src/mail.js
```diff
@@ -281,7 +302,51 @@ export const makeMailboxMaker = ({ provide, provideAgentForHandle }) => {
...
     };
 
+    /**
+     * @param {import('./types.js').Envelope} envelope
+     */
+    const open = envelope => {
+      const message = outbox.get(envelope);
+      if (message === undefined) {
+        throw new Error('Mail fraud'); // TODO
+      }
```
rekmarks: What exactly is the TODO here?

rekmarks (approved):
LGTM! Just a couple of questions.

packages/daemon/src/types.d.ts
kriskowal: Left over. I’ll delete the comment.

PR #2187, opened 2024-03-28
refactor(daemon): Refactor pet store id to name mapping

rekmarks:
Refactors the pet store's id to name mapping by replacing its multiple internal maps with a bidirectional multimap. Also fixes what is probably a bug in the `follow()` functionality, where `add` and `remove` events were published for redundant `write()` calls. Tests of the existing behavior were added before the refactor, then updated for the new behavior.

Note: The bidirectional multimap _may_ also be used for `followReverseLocate()`, which will be introduced in a follow-up.


kriskowal (approved):
Some naming thoughts for your consideration.

PR #2188, opened 2024-03-28
fix(compartment-mapper): fix ReadPowers-related types

boneskull:
## Description

This change:

1. Creates a minimal interface for the `fs`, `url`, and `crypto` objects as passed into `makeReadPowers()`. This makes it easier to duck-type the objects.
2. Fixes the invalid type of `MaybeReadPowers`; properties (defined thru `@property`) are ignored in a `@typedef` of that `@typedef` does not extend `object`/`Object`.
3. Added necessary type assertion in `powers.js`
4. Adds return type to `makeReadPowersSloppy()`

### Upgrade Considerations

This is a type-only change.

### DIFF
--- a/packages/compartment-mapper/src/node-powers.js
+++ b/packages/compartment-mapper/src/node-powers.js
@@ -30,9 +30,10 @@ const fakePathToFileURL = path => {
  * but `makeReadPowers` presents a type that requires `url`.
  *
  * @param {object} args
- * @param {typeof import('fs')} args.fs
- * @param {typeof import('url')} [args.url]
- * @param {typeof import('crypto')} [args.crypto]
+ * @param {import('./types.js').FsAPI} args.fs
+ * @param {import('./types.js').UrlAPI} [args.url]
+ * @param {import('./types.js').CryptoAPI} [args.crypto]
+ * @returns {import('./types.js').MaybeReadPowers}
  */
 const makeReadPowersSloppy = ({ fs, url = undefined, crypto = undefined }) => {
   const fileURLToPath =
--- a/packages/compartment-mapper/src/powers.js
+++ b/packages/compartment-mapper/src/powers.js
@@ -18,7 +18,8 @@ export const unpackReadPowers = powers => {
   if (typeof powers === 'function') {
     read = powers;
   } else {
-    ({ read, maybeRead, canonical } = powers);
+    ({ read, maybeRead, canonical } =
+      /** @type {import('./types.js').MaybeReadPowers} */ (powers));
   }
 
   if (canonical === undefined) {
--- a/packages/compartment-mapper/src/types.js
+++ b/packages/compartment-mapper/src/types.js
@@ -165,8 +165,7 @@ export {};
  */
 
 /**
- * @typedef {ReadPowers | object} MaybeReadPowers
- * @property {MaybeReadFn} maybeRead
+ * @typedef {ReadPowers & {maybeRead: MaybeReadFn}} MaybeReadPowers
  */
 
 /**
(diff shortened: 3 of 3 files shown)


boneskull:
* **#2188** <a href="https://app.graphite.dev/github/pr/endojs/endo/2188?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @boneskull and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/compartment-mapper/src/node-powers.js
```diff
@@ -30,9 +30,10 @@ const fakePathToFileURL = path => {
  * but `makeReadPowers` presents a type that requires `url`.
  *
  * @param {object} args
- * @param {typeof import('fs')} args.fs
- * @param {typeof import('url')} [args.url]
- * @param {typeof import('crypto')} [args.crypto]
+ * @param {import('./types.js').FsAPI} args.fs
+ * @param {import('./types.js').UrlAPI} [args.url]
+ * @param {import('./types.js').CryptoAPI} [args.crypto]
```
kriskowal, line 35: I’m a tiny bit worried that, if we need more powers from these modules, reaching them going forward will constitute a breaking change. Is it possible to address your concerns without narrowing these types?

packages/compartment-mapper/src/powers.js
```diff
@@ -18,7 +18,8 @@ export const unpackReadPowers = powers => {
   if (typeof powers === 'function') {
     read = powers;
   } else {
-    ({ read, maybeRead, canonical } = powers);
+    ({ read, maybeRead, canonical } =
+      /** @type {import('./types.js').MaybeReadPowers} */ (powers));
```
kriskowal, line 22: Cool.

packages/compartment-mapper/src/types.js
```diff
@@ -165,8 +165,7 @@ export {};
  */
 
 /**
- * @typedef {ReadPowers | object} MaybeReadPowers
- * @property {MaybeReadFn} maybeRead
+ * @typedef {ReadPowers & {maybeRead: MaybeReadFn}} MaybeReadPowers
```
kriskowal, line 168: The old pattern broken in general? Used to be vogue.

packages/compartment-mapper/src/node-powers.js
boneskull: I don't see it as possible to do without breaking things, no.  But currently, we're asking for way more than we need to.  

I'd even be amenable to ditching these objects entirely, and asking for the specific functions in isolation--i.e., instead of discrete `fs`, `url` and `crypto` objects, we'd ask for a bucket of functions matching what we need.

packages/compartment-mapper/src/types.js
boneskull: I don't recall that ever working.  To combine types in a `@typedef`, I've always needed to use intersections or something akin to `type-fest`'s `Merge`.

As written, it looks like it _might_ work, but for the wrong reasons--afaict `ReadPowers | object` simplifies to `object`.

packages/compartment-mapper/src/powers.js
boneskull: I think this is actually `MaybeReadPowers | ReadPowers`, but we don't know until L29

packages/compartment-mapper/src/node-powers.js
kriskowal: I’m with you on principle of least authority but not aligned on resilience to change. This is a tool for attenuating the Node.js builtins down to the powers the compartment mapper already needs. The powers space is very narrowly scoped and any new powers have to be optional and have backward-compatible fallbacks. The attenuator however benefits from being able to reach for more powers as needed to upgrade gracefully. Last time we needed this, it was to reach for fs.realpath in order to canonicalize. I _hope_ the dependencies for compartment mapper are stable, but I’m not sure.

PR #2189, opened 2024-03-28
feat(compartment-mapper): expose loadCompartmentForArchive

boneskull:
## Description

This extracts a function, `loadCompartmentForArchive()`, from `digestLocation()` in `archive.js`, and exposes it. This function returns archival-ready data structures, in addition to a record of the compartment renames which `makeArchiveCompartmentMap()` performs.

The intent is to leverage the module resolution performed by the function, which is reflected in its output. A consumer can use the information therein to perform additional operations on the tree (e.g., parse ASTs).

### Upgrade Considerations

This is a backwards-compatible change

PR #2190, opened 2024-03-28
feat(compartment-mapper): expose extraParsers option

boneskull:
## Description

The `extraParsers` option is an optional mapping of file extension (without leading `.`) to parser name. Consumers can use it to associate a file extension with a specific parser, e.g., `{"node": "bytes"}` or override the default mappings.

The specific use-case is enabling creation of compartment map descriptors for Node.js native modules (e.g., `index.node`) _without_ needing a `parsers: {node: 'bytes'}` property in the `package.json` of the package containing the native module.

Because we are unlikely to be the maintainers of any given package exposing a native module, we do not have control over its `package.json`; to add the needed `parsers` field would require forking or using `patch-package` or something else. With this change, the consumer can add `extraParsers: {node: 'bytes'}` to any supported API call.

The API calls supporting the new option should be any that end up calling `link()` in `link.js`; this is where the mapping is concatenated with the default set of parsers.

### Upgrade Considerations

This is a backwards-compatible change.


boneskull:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/endojs/endo/2190?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#2190** <a href="https://app.graphite.dev/github/pr/endojs/endo/2190?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2189** <a href="https://app.graphite.dev/github/pr/endojs/endo/2189?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2188** <a href="https://app.graphite.dev/github/pr/endojs/endo/2188?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @boneskull and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

kriskowal:
I think my choice of `"parsers"` in `package.json` was a misnomer. I changed it to `languageForExtension` internally, but can’t bring myself to be that verbose in `package.json`. I could bring myself to be that verbose for the option, though. Would you consider `fallbackLanguageForExtension`? That’d clarify that it shouldn’t override a local choice.

packages/compartment-mapper/src/link.js
```diff
@@ -374,7 +376,7 @@ export const link = (
     compartmentDescriptor.modules = modules;
 
     const parse = mapParsers(
-      languageForExtension,
+      { ...languageForExtension, ...extraParsers },
```
kriskowal, line 379: Are we sure we want this to `override`? Could we install `fallback`? I could buy both.

boneskull: _fallback_ would work for my use case, but _override_ may be more useful in the general case (i.,e. some other use-case that I can't be bothered to define precisely).  I think we can go with _fallback_ for now.

PR #2191, opened 2024-03-29
docs(pass-style): don't rely on Remotable in Far docs

dckc:
refs:

 - https://github.com/Agoric/documentation/issues/1031

The 1st bit of SDK used in the 1st sample app in our docs is `Far`.

https://github.com/Agoric/dapp-offer-up/blob/cbf3836dff05e95220a201f930430a10ce960ce3/contract/src/offer-up.contract.js#L22

likewise the "hello world" contract in [smart contract basics](https://docs.agoric.com/guides/zoe/contract-basics.html).

Far's JSDoc was in terms of Remotable. I don't think folks using `Far` need to know about `Remotable` so much.

There's momentum to skip `Far` and teach exo's from the start (https://github.com/Agoric/documentation/issues/1033) but we're not there yet.

### Security / Scaling / Testing / Compatibility / Upgrade Considerations

n/a

### DIFF
--- a/packages/pass-style/src/make-far.js
+++ b/packages/pass-style/src/make-far.js
@@ -164,16 +164,24 @@ const getMethodNamesDescriptor = harden({
 });
 
 /**
- * A concise convenience for the most common `Remotable` use.
+ * Mark an object to be exposed for remote interaction
+ * and give it a suggestive interface name for debugging.
  *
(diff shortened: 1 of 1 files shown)


erights (approved):
I didn't know about `@example`. Cool!

LGTM, thanks.

PR #2194, opened 2024-03-30
feat(daemon): Add bidirectional map

rekmarks:
Adds a (non-multi) bidirectional map alongside the bidirectional multimap implementation.


rekmarks:
As it turns out, I don't actually need this for anything right now, but I'm putting this up and closing it For the Record<sup>TM</sup>.

PR #2195, opened 2024-03-30
feat(daemon): Add `reverseLocate()` and `followLocatorNameChanges()`

rekmarks:
Adds `reverseLocate()` and `followLocatorNameChanges()` to the daemon's directory. To match this new method, the directory's `followChanges()` and the pet store's underlying `follow()` are renamed to `followNameChanges()`.

The directory's `reverseLocate(locator)` and `followLocatorNameChanges(locator)` are backed up by the pet store's `reverseIdentify(id)` and `followIdNameChanges(id)`, respectively. The implementation of the former is very straightforward. The implementation of the latter required the introduction of a new pubsub topic, `idChangesTopic`, and a map from ids to `idChangesTopic` subscriptions (one for each id).

Since the pet store is unaware of whether ids exist, `followIdNameChanges()` will naively create a subscription that first publishes an empty array of names, and then any subsequent names that are added for that id. `followLocatorNameChanges()` inherits this property.

The current implementation suffers from the limitation that locator name change subscriptions cannot be ended. A follow-up will introduce the ability for subscribers to cancel their subscription. Locator subscriptions will also be a target for our future garbage collector, which will delete subscriptions for unreachable locators. Cancelling a locator should not end the subscription, because we consider the liveness of a value to be orthogonal to the subscriber's interest in changes to its names.


packages/daemon/src/directory.js
```diff
@@ -100,6 +100,28 @@ export const makeDirectoryMaker = ({
...
     };
 
+    /** @type {import('./types.js').EndoDirectory['reverseLocate']} */
+    const reverseLocate = async locator => {
+      const id = idFromLocator(locator);
+      return petStore.reverseIdentify(id);
+    };
+
+    /** @type {import('./types.js').EndoDirectory['followLocatorNames']} */
+    const followLocatorNames = async function* followLocatorNames(locator) {
```
kriskowal, line 110: name’s growing on me.

packages/daemon/src/directory.js
```diff
@@ -126,16 +148,16 @@ export const makeDirectoryMaker = ({
       return harden(Array.from(identities).sort());
     };
 
-    /** @type {import('./types.js').EndoDirectory['followChanges']} */
-    const followChanges = async function* followChanges(...petNamePath) {
+    /** @type {import('./types.js').EndoDirectory['followNames']} */
+    const followNames = async function* followNames(...petNamePath) {
```
kriskowal, line 152: I like this change, but will note that @kumavis made the opposite change, so we should get on the same page. I suppose @kumavis liked having `Changes` in the name because `followNames` implies snapshots. I could meet up at `followNameChanges`, to leave room for the other form.

packages/daemon/src/locator.js
```diff
@@ -28,48 +28,50 @@ const isValidLocatorType = allegedType =>
  */
 const assertValidLocatorType = allegedType => {
   if (!isValidLocatorType(allegedType)) {
-    assert.Fail`Unrecognized locator type ${q(allegedType)}`;
+    throw assert.error(`Unrecognized locator type ${q(allegedType)}`);
```
kriskowal, line 31: Linter trouble you about type narrowing? These should be equivalent.

kriskowal (commented):
So far, so good!

packages/daemon/src/locator.js
rekmarks: They are equivalent, I just found older uses of `throw assert.error()` in `formula-identifier.js` and decided to standardize on that more familiar (to me) idiom.

As of this change, there are only two other uses of `assert.Fail` in the daemon. We could make all of these cases use the same thing?

packages/daemon/src/directory.js
rekmarks: While pretty clear about the format, I find `followChanges` too non-specific regarding _which_ changes are being followed. I'd be okay with `followNameChanges` / `followLocatorNameChanges`, but I also think it's fine to force consumers to look at the docs to figure out the format and go with the terser `followNames` / `followLocatorNames`. Personally, I wouldn't expect anything about the format with the latter formulation.

I've pinged @kumavis for feedback, and we can await his word to form a consensus.

kumavis: I like having "changes" when the iterated entries are diffs rather than values

rekmarks: > but I also think it's fine to force consumers to look at the docs to figure out the format

To be clear: I think this because consumers will have to look at the docs anyway. There is more than one way to format both snapshots and incremental changes.

rekmarks: @kriskowal wanna break the tie?

kriskowal: Let’s run with `followNameChanges`.

packages/daemon/src/locator.js
kriskowal: Agoric style favors `assert.Fail` and I just haven’t gone back to make the dæmon consistent with the emerging æsthetic.

rekmarks: Ah, there was one other reason: TypeScript does not appear to understand that `assert.Fail` throws, and so some of my type guards only actually narrow the types when using `throw assert.error()` as opposed to `assert.Fail`.

packages/daemon/src/pet-store.js
```diff
@@ -168,9 +194,6 @@ export const makePetStoreMaker = (filePowers, locator) => {
         );
       }
       assertValidId(formulaIdentifier, fromName);
-      if (overwrittenId !== undefined) {
-        assertValidId(overwrittenId, toName);
-      }
```
rekmarks, line 151: If this id exists we just got it from the memo, and we validate ids before they are memoized.

packages/daemon/src/locator.js
kriskowal: I’m seeing the same. We can circle back to fix it.

packages/daemon/src/directory.js
```diff
@@ -97,6 +97,30 @@ export const makeDirectoryMaker = ({
       return formatLocator(id, formulaType);
     };
 
+    /** @type {import('./types.js').EndoDirectory['reverseLocate']} */
+    const reverseLocate = async locator => {
```
kriskowal, line 101: Silly idea:
- “locate” means “get locator for name”
- “nominate” means “get name from locator”

packages/daemon/src/directory.js
```diff
@@ -97,6 +97,30 @@ export const makeDirectoryMaker = ({
...
+      const id = idFromLocator(locator);
+      return petStore.reverseIdentify(id);
+    };
+
+    /** @type {import('./types.js').EndoDirectory['followLocatorNameChanges']} */
+    const followLocatorNameChanges = async function* followLocatorNameChanges(
+      locator,
+    ) {
+      const id = idFromLocator(locator);
+      for await (const idDiff of petStore.followIdNameChanges(id)) {
```
kriskowal, line 111: I will digest this more easily if we only use “change” over synonyms like “diff” and “patch”. I take synonyms mean a deliberate difference and distinction.

packages/daemon/src/pet-sitter.js
```diff
@@ -47,16 +47,35 @@ export const makePetSitter = (petStore, specialNames) => {
...
+    yield* petStore.followNameChanges();
+  };
+
+  /** @type {import('./types.js').PetStore['followIdNameChanges']} */
+  const followIdNameChanges = async function* currentAndSubsequentIds(id) {
+    const subscription = petStore.followIdNameChanges(id);
+
+    const [idSpecialName] = Object.entries(specialNames)
+      .filter(([_, specialId]) => specialId === id)
+      .map(([specialName, _]) => specialName);
```
kriskowal, line 68: We don’t currently assign multiple special names to an identifier, but I think we can eliminate the `if` block below if we assume there can be zero or more entries in this array.

kriskowal (approved):
Worth considering whether we might just factor a light-client out of `followLocatorNameChanges` that is based on a sharable subscription to `followNameChanges`. That might put the tracking burden closer to where it ought to be and relax the need to perform timely identifier GC.

Issue #2198, opened 2024-04-03
Marshal fails to rehydrate errors in Chrome, Firefox, Safari


kriskowal:
## Describe the bug

Scene: a web page CapTP client connects over WebSocket to a remote CapTP endpoint. The client invokes a remote method. The server throws an error. The server logs that error. The client receives the error. However, instead of reporting an error with the same message as observed on the server, we see an error indicating that Marshal was unable to hydrate the error.

Chrome:
```
Error#2: Passable Error stack own property must be a data property: Object
```

Firefox:
```
Error#2: Passable Error has extra unpassed property fileName
```

Safari
```
Error#2:"Passable Error has extra unpassed property""line"
```

## Steps to reproduce

To reproduce in the Endo repository:

In `endo/packages/cli`, create `err.js`

```js
export const make = async powers => {
  await E(powers).bogus();
};
```

Then, also in `endo/packages/cli`:

```
bin/endo purge --force
bin/endo install err.js --name err --listen 3333 --powers NONE --open
```

This will open your system browser to a weblet. There will be one of the above errors in the JavaScript console.

## Expected behavior

An error should be reported that has the same message as the error reported on the server.

For the isolated reproduction above, you can find the server’s error in one of the Endo worker logs. I apologize ahead of time that this is not yet polished: `bin/endo log` will show you the most recent daemon logs and the error should be in there. `bin/endo where log` will tell you where that log file is so you can open it elsewhere. `bin/endo where state` will show you where all the logs are.

## Platform environment

Any system, any browser.

erights:
Reopening because the previous fix #2200 did not update regular non-browser tests so they would work on a platform with v8's problematic stack-accessor behavior. I accidentally discovered this when testing locally with Node 21, which our CI does not yet test.

erights:
See https://chromium-review.googlesource.com/c/v8/v8/+/4459251 where the problematic v8 behavior started

erights:
At https://chromium-review.googlesource.com/c/v8/v8/+/4459251 v8 changed their error `stack` property into an own accessor property with per-instance getters and setters. Because these getters and setters are fresh per error instance, on seeing such a property, we have no way to know whether the getter and setter is the platform-provided one, or one provided by an attacker.

erights:
@kriskowal , @frazarshad , since this is in the same ballpark, I went ahead and co-assigned this to the three of us.

erights:
[...]

PR #2200, opened 2024-04-05
fix(ses): makeError defaults to making passable errors

erights:
closes: #2198 
refs: #2156 

## Description

The symptoms of #2156 reveal that some hosts add "extraneous" or harmful own properties to errors
- Chrome, Brave (v8) adds an own `stack` accessor property, whose getter and setter functions can be used for mischief.
- Firefox (SpiderMonkey) adds own `fileName` and `lineNumber` properties that are useful for diagnostics but should be redacted from public view.
- Safari (JSC) adds an own `line` property, that likewise is useful for diagnostics, but should be redacted from public view.

Note that the contents of the stack should also be redacted from public view, but SES already deals with that separately.

Prior to this PR, some code assumed that the error that `makeError` returns would be passable once frozen. But these extra own properties, or own accessor properties, added by the host before `makeError` returns the error, caused it not to be passable, causing #2198 . As of this PR, `makeError` *by default* makes reasonable effort to return a passable error despite such host-added own properties. However, some code took advantage of the mutability of the error that `makeError` was returning, so this PR adds a `sanitize: false` option to suppress this cleaning and freezing of the returned error.

`makeError` does *not* guarantee that the returned error is passable, no matter what `cause` or `errors` options are provided. The purpose of this new sanitization behavior is to protect against host mischief. To test or coerce an error into a passable error, use `toPassable` in @endo/pass-style.

### Security Considerations

Having `makeError`, by default, return errors that are already better behaved and more discreet helps security. However, because the underlying platform makes errors that are not sanitized in this manner, this PR doesn't help security significantly.

### Scaling Considerations

none

I added doc-comments to the `makeError` options doc-comment. However, the important added doccomment is on the new `sanitizeError` function which is encapsulated with SES's assert.js. I made a `TODO` that I don't know how one doccomment can link to another.

### Testing Considerations

This problem was caught by accident. We should do enough browser-based testing early enough that problems like this would be caught first by purposeful testing.

This PR is staged on #2156. Like #2156, it presents a compat problem in theory but likely not in practice. #2156 stops exporting `isPassableError` and `assertPassableError`, which it turns out were never correct but which no one imports or uses. 

This PR changes `makeError` to return a "sanitized" and frozen error, where it used to return a non-sanitized and non-frozen error. I looked through every call to `makeError` in endo and agoric-sdk, and found one site that made use of the non-frozen nature of that error. However, `makeError` is much older, so it is more plausible that there are other such uses outside these two repos. To accommodate that one known use case and mitigate problems at possible others, this PR adds a `sanitize: false` option to `makeError` to recover the old behavior. That one known site uses this option.

I leave it to my reviewers whether this PR as well as #2156 should be marked with a `!` or not. In theory both should be. IMO, we should mark neither with `!`. Please let me know.

### Upgrade Considerations

Since this bug manifested for #2156 on unmarshalling, i.e., unserializing use the `@endo/marshal` mechanisms, that bug raises the possibility of an upgrade hazard -- that an error saved to durable storage before an upgrade might cause another error on the attempt to read it from durable storage after the upgrade. This PR prevents that hazard.

### DIFF
--- a/packages/ses/src/error/assert.js
+++ b/packages/ses/src/error/assert.js
@@ -35,6 +35,11 @@ import {
   weakmapHas,
   weakmapSet,
   AggregateError,
+  getOwnPropertyDescriptors,
+  ownKeys,
+  create,
+  objectPrototype,
+  objectHasOwnProperty,
 } from '../commons.js';
 import { an, bestEffortStringify } from './stringify-utils.js';
 import './types.js';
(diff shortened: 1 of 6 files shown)


erights:
Following the repro instructions at #2198 , without this PR I get the bug it describes on the console output of Brave (v8), Firefox (SpiderMonkey) and Safari (JSC). With this PR, I get the following screenshots, showing that the desired error is now successfully reported

Brave:
image

Firefox
image

Safari
image

erights:
Now testing CI for https://github.com/Agoric/agoric-sdk/pull/9201 with `force:integration` label and `#endo-branch: markm-sanitize-errors`. We'll see how it goes.

packages/ses/src/error/types.js
```diff
@@ -27,9 +27,24 @@
...
+ *   Normally only used when the ErrorConstuctor is `AggregateError`, to
+ *   represent the set of prior errors aggregated together in this error,
+ *   typically by `Promise.any`. But `makeError` allows it on any error.
+ *   This is represented by a public `errors` data property on the error,
+ *   not a hidden annotation.
+ * @property {boolean} [sanitize]
+ *   Defaults to true. If true, `makeError` will apply `sanitizeError`
+ *   to the error before returning it. See the comments on `sanitizeError`.
+ *   (TODO what is the proper jsdoc manner to link to another function's
+ *   doc-comment?)
```
kriskowal, line 47: JSDoc of yore supported `@link` annotations. I think it depends on the doc generator. @LuqiPan was setting something up that appeared to support Markdown, but that implies some coupling to the layout of the generated docs too.

kriskowal (approved):

packages/ses/src/error/types.js
erights: Well, I gave it a good try and learned a bunch, but did not succeed. For now I'll just leave the TODO in there and we can then figure this out separately.

Resolving.

packages/ses/src/error/assert.js
```diff
@@ -254,13 +259,80 @@ const tagError = (err, optErrorName = err.name) => {
...
+ *     by the host,
+ *     such as `fileName`,`lineNumber` on FireFox or `line` on Safari.
+ *   - If any such "extraneous" properties were removed, `sanitizeError` will
+ *     annotate
+ *     the error with them, so they still appear on the causal console
+ *     log output for diagnostic purposes, but not be otherwise visible.
+ *   - `sanitizeError` will ensure that any expected properties already
+ *     added by the host are data
+ *     properties, converting accessor properties to data properties as needed,
+ *     such as `stack` on v8 (Chrome, Brave, Edge?)
```
mhofman, line 274: It looks like stack own accessors are not modified by this helper, what am I missing?

Issue #2201, opened 2024-04-05
add importNowHook to SES Compartment


naugtur:
## What is the Problem Being Solved?
Any level of dynamic behavior for CJS require in endo implemented on top of Compartment requires either preloading all dynamically requested modules or adding importNowHook - the synchronous version of importHook.

## Description of the Design
Provide synchronous module loading facilities in SES by analogy to what already exists.

## Security Considerations
TBD, none that I'm aware of.

## Test Plan
- port some unit tests to synchronous flow
- test with dynamic require in compartment-mapper

## Compatibility Considerations
- This will never work with top-level await anywhere in the tree starting at the specifier being passed to importNow.

## Upgrade Considerations
- We might have depended on a missing specifier causing a specific error from importNow (seems unlikely)

PR #2202, opened 2024-04-05
feat(ses): importNowHook

naugtur:
see #2201 

TODO:
 - [x] Make sure importNowHook is only called if moduleMap doesn't already have an entry for that specifier
 - [x] Add error handling where error aggregation used to be
 - [x] find a way to reuse logic
 - [x] figure out how I've offended eslint this time
 - [x] wire up the verbose loading errors option for printing stacks of aggregated errors to an env variable

~~Currently ctrl+c;ctrl+v is used for code reuse.~~
generator + trampoline allowed reuse of the load logic.

For anyone who needs an intro to trampolines - here's the intro I got:
https://dev.to/naugtur/gpt-teaches-me-how-to-make-my-logic-sync-and-async-at-the-same-time-with-trampolines-4hnl


packages/ses/src/module-load-sync.js
```diff
@@ -0,0 +1,330 @@
+// For brevity, in this file, as in module-link.js, the term "moduleRecord"
```
naugtur, line 1: Diff of how this file became, for a slight improvement in convenience.
```diff
13d12
<   Set,
22,24d20
<   setAdd,
<   promiseCatch,
<   promiseThen,
32,33d27
< const noop = () => {};
< 
63d56
<   pendingJobs,
89,105c82,95
<     // Behold: recursion.
<     // eslint-disable-next-line no-use-before-define
<     const dependencyLoaded = memoizedLoadWithErrorAnnotation(
<       compartmentPrivateFields,
<       moduleAliases,
<       compartment,
<       fullSpecifier,
<       pendingJobs,
<       moduleLoads,
<       errors,
<     );
<     setAdd(
<       pendingJobs,
<       promiseThen(dependencyLoaded, noop, error => {
<         arrayPush(errors, error);
<       }),
<     );
---
>     try {
>       // Behold: recursion.
>       // eslint-disable-next-line no-use-before-define
>       memoizedLoadWithErrorAnnotation(
>         compartmentPrivateFields,
>         moduleAliases,
>         compartment,
>         fullSpecifier,
>         moduleLoads,
>         errors,
>       );
>     } catch (error) {
>       arrayPush(errors, error);
>     }
113c103
< const loadWithoutErrorAnnotation = async (
---
> const loadWithoutErrorAnnotation = (
118d107
<   pendingJobs,
122c111
<   const { importHook, moduleMap, moduleMapHook, moduleRecords } = weakmapGet(
---
>   const { importNowHook, moduleMap, moduleMapHook, moduleRecords } = weakmapGet(
153c142
<     const aliasRecord = await memoizedLoadWithErrorAnnotation(
---
>     const aliasRecord = memoizedLoadWithErrorAnnotation(
158d146
<       pendingJobs,
170c158
<   const staticModuleRecord = await importHook(moduleSpecifier);
---
>   const staticModuleRecord = importNowHook(moduleSpecifier);
173c161
<     Fail`importHook must return a promise for an object, for module ${q(
---
>     Fail`importNowHook must return an object, for module ${q(
201d188
<         pendingJobs,
220c207
<       const aliasRecord = await memoizedLoadWithErrorAnnotation(
---
>       const aliasRecord = memoizedLoadWithErrorAnnotation(
225d211
<         pendingJobs,
242d227
<     pendingJobs,
248c233
< const memoizedLoadWithErrorAnnotation = async (
---
> const memoizedLoadWithErrorAnnotation = (
253d237
<   pendingJobs,
273,274c257,258
<   moduleLoading = promiseCatch(
<     loadWithoutErrorAnnotation(
---
>   try {
>     moduleLoading = loadWithoutErrorAnnotation(
279d262
<       pendingJobs,
282,293c265,275
<     ),
<     error => {
<       // eslint-disable-next-line @endo/no-polymorphic-call
<       assert.note(
<         error,
<         d`${error.message}, loading ${q(moduleSpecifier)} in compartment ${q(
<           compartmentName,
<         )}`,
<       );
<       throw error;
<     },
<   );
---
>     )
>   } catch (error) {
>     // eslint-disable-next-line @endo/no-polymorphic-call
>     assert.note(
>       error,
>       d`${error.message}, loading ${q(moduleSpecifier)} in compartment ${q(
>         compartmentName,
>       )}`,
>     );
>     throw error;
>   }
301c283
<  * `load` asynchronously gathers the `StaticModuleRecord`s for a module and its
---
>  * `loadSnc` synchronously gathers the `StaticModuleRecord`s for a module and its
307c289
< export const load = async (
---
> export const loadSync = (
318,320c300
<   /** @type {Set<Promise<undefined>>} */
<   const pendingJobs = new Set();
<   /** @type {Map<object, Map<string, Promise<Record<any, any>>>>} */
---
>   /** @type {Map<object, Map<string, Record<any, any>>>} */
325,347c305,315
<   const dependencyLoaded = memoizedLoadWithErrorAnnotation(
<     compartmentPrivateFields,
<     moduleAliases,
<     compartment,
<     moduleSpecifier,
<     pendingJobs,
<     moduleLoads,
<     errors,
<   );
<   setAdd(
<     pendingJobs,
<     promiseThen(dependencyLoaded, noop, error => {
<       arrayPush(errors, error);
<     }),
<   );
< 
<   // Drain pending jobs queue.
<   // Each job is a promise for undefined, regardless of success or failure.
<   // Before we add a job to the queue, we catch any error and push it into the
<   // `errors` accumulator.
<   for (const job of pendingJobs) {
<     // eslint-disable-next-line no-await-in-loop
<     await job;
---
>   try {
>     memoizedLoadWithErrorAnnotation(
>       compartmentPrivateFields,
>       moduleAliases,
>       compartment,
>       moduleSpecifier,
>       moduleLoads,
>       errors,
>     );
>   } catch (error) {
>     arrayPush(errors, error);
349c317,318
< 
---
>  
>   
```

packages/ses/src/module-load.js
```diff
@@ -315,47 +382,80 @@ export const load = async (
...
+
+  // Drain pending jobs queue and throw an aggregate error
+  const errors = await drainQueue();
+
+  throwAggregateError({
     errors,
+    errorPrefix: `Failed to load ${q(moduleSpecifier)} in compartment ${q(
+      compartmentName,
+    )}`,
+    debug: true,
```
naugtur, line 408: What's the recommended way to pass a boolean into here? Is there a precedent for using env variables? I keep switching between message and stack in these aggregated errors whenever working with SES.

kriskowal: We do have a precedent for using `process.env` and some tooling around it from `@endo/env-options`. This is how we thread environment `LOCKDOWN_` options.

naugtur: `COMPARTMENT_LOAD_ERROR_VERBOSE` or `COMPARTMENT_LOAD_DEBUG` ?

packages/ses/src/commons.js
```diff
@@ -133,6 +133,8 @@ export const { prototype: weakmapPrototype } = WeakMap;
 export const { prototype: weaksetPrototype } = WeakSet;
 export const { prototype: functionPrototype } = Function;
 export const { prototype: promisePrototype } = Promise;
+// eslint-disable-next-line no-empty-function
+export const { prototype: iteratorPrototype } = getPrototypeOf(function*(){});
```
naugtur, line 137: I'd appreciate if someone could verify whether this is the iterator prototype or generator prototype. I managed to confuse myself while implementing this and while I think this is correct, there's doubt.

cc @mhofman

packages/ses/src/module-load.js
```diff
@@ -297,6 +322,63 @@ const memoizedLoadWithErrorAnnotation = async (
...
+  const pendingJobs = new Set();
+  /** @type {Array<Error>} */
+  const errors = [];
+
+  /**
+   * Enqueues a job that starts immediately but won't be awaited until drainQueue is called.
+   * 
+   * @template {any[]} T
+   * @param {(...args: T)=>Promise<*>} func
+   * @param {T} args
```
naugtur, line 336: @boneskull A few months ago I had trouble specifying what a template extends, but now it seems to work fine with this syntax.

packages/ses/src/module-load.js
```diff
@@ -315,47 +400,78 @@ export const load = async (
...
+    preferAsync,
     moduleLoads,
+  ]);
+
+  // Drain pending jobs queue and throw an aggregate error
+  const errors = await drainQueue();
+
+  throwAggregateError({
     errors,
+    errorPrefix: `Failed to load module ${q(moduleSpecifier)} in package ${q(
```
naugtur, line 423: This error message should say `in compartment`  to be consistent with the other error message in the same file, but the inconsistency was introduced ~3 years ago and I'm not taking that on in this PR

packages/ses/README.md
```diff
@@ -384,6 +384,41 @@ const odd = new Compartment({}, {}, {
 });
 ```
 
+### importNowHook
+
+Additionally, an `importNowHook` may be provided that the compartment will use as means to synchronously load modules not seen before in situations where calling out to asynchronous `importHook` is not possible.
+Specifically, when `compartmentInstance.importNow('specifier')` is called, the compartment will first look up module records it's already aware of and call `moduleMapHook` and if none of that is successful in finding a module record matching the specifier, it will call `importNowHook` expecting to synchronously receive the same record type as from `importHook`
```
kriskowal, line 390:
```suggestion
Specifically, when `compartmentInstance.importNow('specifier')` is called, the compartment will first look up module records it's already aware of and call `moduleMapHook` and if none of that is successful in finding a module record matching the specifier, it will call `importNowHook` expecting to synchronously receive the same record type as from `importHook` or throw if it cannot.
```

packages/ses/README.md
```diff
@@ -384,6 +384,41 @@ const odd = new Compartment({}, {}, {
...
+    // Platform specific synchronous read API can be used
+    const moduleText = fs.readFileSync(moduleLocation);
+    return new StaticModuleRecord(moduleText, moduleLocation);
+  },
+});
+//...                   | importHook | importNowHook
+await c1.import('a'); //| called     | not called
+c1.importNow('b');    //| not called | called
+c1.importNow('a');    //| not called | not called
+c1.importNow('c2');   //| not called | not called
```
kriskowal, line 419: Very nice.

packages/ses/src/commons.js
```diff
@@ -191,6 +193,10 @@ export const stringEndsWith = uncurryThis(stringPrototype.endsWith);
 export const stringIncludes = uncurryThis(stringPrototype.includes);
 export const stringIndexOf = uncurryThis(stringPrototype.indexOf);
 export const stringMatch = uncurryThis(stringPrototype.match);
+//
```
kriskowal, line 196:
```suggestion

```

packages/ses/src/commons.js
```diff
@@ -191,6 +193,10 @@ export const stringEndsWith = uncurryThis(stringPrototype.endsWith);
...
+export const iteratorNext = uncurryThis(iteratorPrototype.next);
+export const iteratorThrow = uncurryThis(iteratorPrototype.throw);
```
kriskowal, line 198: Might be more accurate to call these `generator*` since they do not work on arbitrary iterators.

packages/ses/src/module-load.js
```diff
@@ -20,17 +21,46 @@ import {
...
+  iteratorThrow,
 } from './commons.js';
 import { assert } from './error/assert.js';
 
 const { Fail, details: d, quote: q } = assert;
 
 const noop = () => {};
 
+async function asyncTrampoline(generatorFunc, args, errorWrapper) {
+  // TODO: add iterator prototype methods to commons
```
kriskowal, line 37: Noticed you did this.

```suggestion

```

packages/ses/src/module-load.js
```diff
@@ -297,6 +323,64 @@ const memoizedLoadWithErrorAnnotation = async (
   return moduleLoading;
 };
 
+function asyncOverseer() {
```
kriskowal, line 326: Recommend naming asyncJobQueue or some such. At least in my dialect, overseer is best known for its use in the Bible and a troubling history of forced labor.

packages/ses/src/module-load.js
```diff
@@ -297,6 +323,64 @@ const memoizedLoadWithErrorAnnotation = async (
...
+    const verbose =
+      getenv('COMPARTMENT_LOAD_ERRORS', '', ['verbose']) === 'verbose';
+    throw TypeError(
+      `${errorPrefix} (${errors.length} underlying failures: ${arrayJoin(
+        arrayMap(errors, error => error.message + (verbose ? error.stack : '')),
+        ', ',
+      )}`,
+    );
+  }
+}
```
kriskowal, line 379: We may now feature detect `AggregateError` and use it. It is admitted into permits now.

kriskowal (commented):
This deserves a mention in `NEWS.md`. I’m chuffed (_chuffed_, I say) that this solution was not terribly invasive!

kriskowal:
attn @erights, @naugtur found a way to introduce `compartment.loadNow` and `importNowHook` that works and shares the existing loader logic by converting it from async functions to generator functions and driving it with either a sync or async trampoline. This is relevant to both alignment with Moddable (tapping @phoddie), Module Harmony, and Node.js’s needs for an `import.now` to drive ESM from CJS `require`.

phoddie:
Very cool.. bordering on magic.

This is the first I've heard about Node.js having a potential need for synchronous import. That makes sense given require. All of a sudden our fringe embedded behavior doesn't seem quite so fringe.

packages/ses/src/commons.js
naugtur: I considered `generatorIterable*`, `generatedIterable*` or `generatee`, which would be more precise but weirder

packages/ses/src/module-load.js
naugtur: I remember removing this line.  :thinking:

packages/ses/src/module-load.js
naugtur: I know the word as construction site term and a StarCraft2 unit :see_no_evil: 
Will fix.

packages/ses/src/module-load.js
naugtur: I did put it in and it looks great. 
It affects so many tests, I want it to be a separate PR and we need to decide whether it makes for a breaking change.

naugtur:
There's some flakiness with an unhandled rejection https://github.com/endojs/endo/actions/runs/8812445711/job/24188150583#step:9:3131
And then as soon as I updated to main, Browser Tests started consistently failing.

packages/ses/NEWS.md
```diff
@@ -1,5 +1,9 @@
 User-visible changes in SES:
 
+# Next release
+
+- `importNowHook` has been added to the `Compartment` options. The hook will be invoked whenever a yet not loaded module is attempted to be synchronously imported via `compartmentInstance.importNow(specifier)`
```
kriskowal, line 5:
```suggestion
- Adds `importNowHook` to the `Compartment` options. The compartment will invoke the hook whenever it encounters a missing dependency during while running `compartmentInstance.importNow(specifier)`, which cannot use an asynchronous `importHook`.
```

kriskowal (approved):

PR #2204, opened 2024-04-06
feat(scripts): Add esbench microbenchmarking utility

gibson042:
TODO:
- Improve output to include aggregate statistics like standard deviation, min/median/max, etc.
- Support collecting performance data for flame graphs etc. (probably bypassing eshost).
- Support loading endoZipBase64 bundles?

### DIFF
--- a/scripts/esbench.mjs
+++ b/scripts/esbench.mjs
@@ -1,0 +1,725 @@
+#!/usr/bin/env node
+import { spawn } from 'node:child_process';
+import { randomBytes } from 'node:crypto';
+import { mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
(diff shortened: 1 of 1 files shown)


erights:
PR comment should first of all say with the PR does contribute. TODOs are nice but secondary.

Need explanations with examples.

I don't know how amenable this is to testing, but presumably we could at least have some smoke tests so we find out if we break it eggregiously.

erights (changes requested):
Just recording here what @gibson042 and I discussed awhile ago:

Rather than this heavy command-line orientation and separate process oriented implementation, we will shift to something that more closely resembles a simplified ses-ava-like framework, where writing bench cases feels much more like writing ses-ava test cases.

kriskowal:
@gibson042 Noting this is stale and could use some clarity of next steps.

Issue #1891, opened 2023-12-11
SES lockdown Hermes compat tracking issue


leotm:
Briefly demo'd https://github.com/MetaMask/metamask-mobile/pull/8009, a minimally transformed ses@0.18.8 wip that runs on Hermes (latest React Native default JS engine)

here's the omnibus of issues/mods below, while figuring which parts we definitely need (vs can skip/remove in a codemod) cc @erights would appreciate any thoughts

#### Bundling (Metro, release)
- [x] `error: async functions are unsupported`
  - [x] <s>`loadWithoutErrorAnnotation`</s>
  - [x] <s>`memoizedLoadWithErrorAnnotation`</s>
  - [x] `load`
  - [x] `drainQueue` (ses@1.5.0)
  - [x] `asyncTrampoline` (ses@master)
  - fixed in Hermes SES shim variant
    - https://github.com/endojs/endo/commit/d04d591ffd1b0da47d08e49ae2e295b3c5634ae1 https://github.com/leotm/RN07117SES/commit/ebc6ed6d22c44f6b66a49f94917f6c643a3d5c9d
- [x] `error: async generators are unsupported`
  - `async function* AsyncGeneratorFunctionInstance() {}`
  - fixed in Hermes SES shim variant
    - https://github.com/endojs/endo/commit/0421ca7eaf4c69c2343637f495c5e5738230cb8f https://github.com/leotm/RN07117SES/commit/8a64849a8bdba01369015bfe18d0d4a96b833e80

+ [x] SES shim variant for Hermes compiler
  - https://github.com/endojs/endo/pull/2334

#### Runtime
- [x] `// assertDirectEvalAvailable(); // SES TypeError`
  - https://github.com/facebook/hermes/issues/957
    - https://github.com/facebook/hermes/pull/1515
  - _https://github.com/facebook/hermes/issues/1056_
    - https://github.com/facebook/hermes/pull/1571
  - [x] DRAFT https://github.com/endojs/endo/compare/master...ses-hermes-p2
  - [x] https://github.com/endojs/endo/pull/2723
- [x] `addIntrinsics(tameRegExpConstructor(regExpTaming)); // SES TypeError`</s>
  - [x] https://github.com/endojs/endo/pull/2108
- [x] `addIntrinsics(tameSymbolConstructor()); // Metro exception, RN polyfill error guard Fatal Error`</s>
  - [x] https://github.com/endojs/endo/pull/2206
[...]

PR #2206, opened 2024-04-10
feat(ses): redefine SharedSymbol to bypass Hermes prototype bug on obj literal short-hand methods

leotm:
closes: #XXXX
refs: https://github.com/endojs/endo/issues/1891

## Description

Fix SES compat with Hermes when calling `addIntrinsics(tameSymbolConstructor());`
- https://github.com/endojs/endo/issues/1891

Follow-up to
- https://github.com/endojs/endo/pull/2108

### Upgrade Considerations


mhofman (commented):
I'm confused as to what the problem is this is trying to solve?

How is the just created `SharedSymbol` function born with own non-configurable properties, conflicting with the original `Symbol` properties?

packages/ses/src/tame-symbol-constructor.js
```diff
@@ -50,14 +50,20 @@ export const tameSymbolConstructor = () => {
     },
   });
 
+  const sharedSymbolDescs = getOwnPropertyDescriptors(SharedSymbol);
   const originalDescsEntries = entries(
     getOwnPropertyDescriptors(OriginalSymbol),
   );
   const descs = fromEntries(
```
leotm, line 57: consider adding a comment like
`// Hermes has many non-configurable Symbol properties. We should support such platforms.`

mhofman: The JS spec says that well known symbols are non-configurable properties of the `Symbol` global. How is hermes different exactly?

leotm:
> I'm confused as to what the problem is this is trying to solve?
> 
> How is the just created `SharedSymbol` function born with own non-configurable properties, conflicting with the original `Symbol` properties?

it's solving SES compatibility for React Native running on Hermes ^ when calling `addIntrinsics(tameSymbolConstructor());` in the shim

otherwise the output on Android emulators is `property is not configurable, no stack`
- [Android emu > screenshot](https://github.com/endojs/endo/assets/1881059/b9ac8c2d-ddd6-4bb8-ab26-3b0013f3492f) (older ses)
- [Android emu > screenshot 2](https://github.com/endojs/endo/assets/1881059/4b34b932-a6bf-4d4f-bd8b-21805736dd1b) (ses@1.4.0)

after calling `defineProperties(SharedSymbol, descs);`
- [Flipper > Hermes Debugger > screenshot](https://github.com/endojs/endo/assets/1881059/2519c891-e03c-4b44-82d4-6c34bdf1275b)
- [Flipper > Hermes Debugger > screenshot 2](https://github.com/endojs/endo/assets/1881059/c5d87168-ba6f-40db-89cc-917f76c6a13d)

the original `originalDescsEntries` on Hermes logged out are

```console
 LOG  length {"configurable": true, "enumerable": false, "value": 0, "writable": false}
 LOG  name {"configurable": true, "enumerable": false, "value": "Symbol", "writable": false}
 LOG  for {"configurable": true, "enumerable": false, "value": [Function for], "writable": true}
 LOG  keyFor {"configurable": true, "enumerable": false, "value": [Function keyFor], "writable": true}
 LOG  hasInstance {"configurable": false, "enumerable": false, "value": Symbol(Symbol.hasInstance), "writable": false}
 LOG  iterator {"configurable": false, "enumerable": false, "value": Symbol(Symbol.iterator), "writable": false}
 LOG  isConcatSpreadable {"configurable": false, "enumerable": false, "value": Symbol(Symbol.isConcatSpreadable), "writable": false}
 LOG  toPrimitive {"configurable": false, "enumerable": false, "value": Symbol(Symbol.toPrimitive), "writable": false}
 LOG  toStringTag {"configurable": false, "enumerable": false, "value": Symbol(Symbol.toStringTag), "writable": false}
 LOG  match {"configurable": false, "enumerable": false, "value": Symbol(Symbol.match), "writable": false}
 LOG  matchAll {"configurable": false, "enumerable": false, "value": Symbol(Symbol.matchAll), "writable": false}
 LOG  search {"configurable": false, "enumerable": false, "value": Symbol(Symbol.search), "writable": false}
 LOG  replace {"configurable": false, "enumerable": false, "value": Symbol(Symbol.replace), "writable": false}
 LOG  split {"configurable": false, "enumerable": false, "value": Symbol(Symbol.split), "writable": false}
```

- [Flipper > Hermes Debugger > screenshot](https://github.com/endojs/endo/assets/1881059/597dddf5-a8de-4349-ae95-54f595590a88)
- [Flipper > Hermes Debugger > screenshot 2](https://github.com/endojs/endo/assets/1881059/36b573ab-a3ec-44ff-9d56-aa1609406018)

so the conflict seems to be attempting to define 11 non-configurable properties to configurable

packages/ses/src/tame-symbol-constructor.js
leotm: looking further at the source

```h
...
/// Represents a JS Symbol (es6).  Movable, not copyable.
/// TODO T40778724: this is a limited implementation sufficient for
/// the debugger not to crash when a Symbol is a property in an Object
/// or element in an array.  Complete support for creating will come
/// later.
class JSI_EXPORT Symbol : public Pointer {
...
```
https://github.com/facebook/hermes/blob/main/API/jsi/jsi/jsi.h#L515-L540

https://github.com/facebook/hermes/blob/main/include/hermes/VM/PrimitiveBox.h#L310-L347

mhofman:
> otherwise the output on Android emulators is `property is not configurable, no stack`
> 
> * [Android emu > screenshot](https://github.com/endojs/endo/assets/1881059/b9ac8c2d-ddd6-4bb8-ab26-3b0013f3492f) (older ses)
> * [Android emu > screenshot 2](https://github.com/endojs/endo/assets/1881059/4b34b932-a6bf-4d4f-bd8b-21805736dd1b) (ses@1.4.0)
> 
> after calling `defineProperties(SharedSymbol, descs);`
> 
> * [Flipper > Hermes Debugger > screenshot](https://github.com/endojs/endo/assets/1881059/2519c891-e03c-4b44-82d4-6c34bdf1275b)
> * [Flipper > Hermes Debugger > screenshot 2](https://github.com/endojs/endo/assets/1881059/c5d87168-ba6f-40db-89cc-917f76c6a13d)
> 
> the original `originalDescsEntries` on Hermes logged out are

None of this tells me which non-configurable property exists on `SharedSymbol` that cannot be redefined. What is the actual trigger?

packages/ses/src/tame-symbol-constructor.js
```diff
@@ -50,14 +50,20 @@ export const tameSymbolConstructor = () => {
...
-    arrayMap(originalDescsEntries, ([name, desc]) => [
-      name,
-      { ...desc, configurable: true },
-    ]),
+    arrayMap(originalDescsEntries, ([name, desc]) => {
+      if (
+        sharedSymbolDescs[name] &&
+        sharedSymbolDescs[name].configurable === false
```
leotm, line 61: nb: condition `name === 'prototype'` is also sufficient on Hermes

mhofman: @erights any ideas on how to best handle this non-compliance? It will likely bite us in other places.

erights: I've never run Hermes before. I'm on a Mac. What's the best way for me to reproduce this? Possible to reproduce under a debugger, as I saw in your screenshots?

erights: >  Short-hand methods are not supposed to have a .prototype property.

Current hypothesis. The correct fix is likely to make `SharedSymbol` a `function` function rather than a concise method. This probably would have been somewhat better all along anyway, since the original `Symbol` does, in most ways, act more like a `function` function than a concise method.

erights: But I'd like to repro locally and see what happens.

erights: > condition name === 'prototype' is also sufficient on Hermes

I like the configurability test better. Perhaps, for aesthetics, we should do both.

mhofman: > since the original `Symbol` does, in most ways, act more like a `function` function than a concise method.

It does throw on construction. We'd have to explicitly do `if (new.target) throw TypeError()`

mhofman: > But I'd like to repro locally and see what happens.

Pretty primitive, but you have a REPL and CLI. Download https://github.com/facebook/hermes/releases/tag/v0.12.0

```
$ ./hermes 
>> ({foo(){}}).foo.prototype
foo { [constructor]: [Function foo] }
```

mhofman: > I like the configurability test better. Perhaps, for aesthetics, we should do both.

I think neither is valid. We need to copy over the `.prototype` or the Symbol "constructor" just isn't correct (`getPrototypeOf(Object(Symbol())) !== Symbol.prototype`)

mhofman: > > since the original `Symbol` does, in most ways, act more like a `function` function than a concise method.
> 
> It does throw on construction. We'd have to explicitly do `if (new.target) throw TypeError()`

As a matter of fidelity, the spec does say `Symbol` has a `[[Construct]]` which throws on construction. It is possible to detect using the following:

```js
const hasConstruct = (target) => { 
  const proxy = new Proxy(target, {
    construct() { return {}; }
  });
  try {
    new proxy();
    return true;
  } catch (e) {
    return false;
  }
}
hasConstruct(Symbol); //true
hasConstruct(({Symbol() {}}).Symbol); // false
```

mhofman: I am not sure why the re-definition of `.prototype` fails since it is marked as `writable`. Anyway, we should switch to function as mentioned above.

I have also filed an issue upstream for the conformance of short hand methods: https://github.com/facebook/hermes/issues/1371

mhofman (changes requested):
We should bypass the Hermes bug, and increase spec fidelity by only doing the following:

```js
  const SharedSymbol = functionBind(Symbol, undefined);
```

Where in `common.js`:
```js
export const functionBind = uncurryThis(bind);
```

The above will maintain the behavior of the original Symbol "constructor", including its native throwing behavior. It works because the receiver is not used by the `[[Call]]` of the primordial ``%Symbol%`.

The `name` property will be "fixed" by the later copy of own properties from OriginalSymbol, and the `toString()` will be censored like for the original. Special casing of own properties should no longer be necessary for Hermes.

packages/ses/src/tame-symbol-constructor.js
leotm: thanks again for looking into this and raising the issue
and sorry i didn't spot Hermes fixed this [last week](https://github.com/facebook/hermes/commit/00f18c89c720e1c34592bb85a1a8d311e6e99599) in the [static_h](https://github.com/facebook/hermes/tree/static_h) branch
since Static Hermes it's few months from shipping, will proceed with changes
perhaps with a comment like this?
```js
// This fix can be removed when Hermes (containing the mistake)
// is deprecated for Static Hermes (containing the fix)
// which is a backwards compatible drop-in replacement
// so object literal short-hand methods no longer have have a prototype
```

leotm: since the https://github.com/facebook/hermes/releases/tag/v0.12.0 CLI is shy of couple yrs old
i requested to help get an updated version (being discussed with team)

so i tested the commit fix by [building and running](https://github.com/facebook/hermes/blob/main/doc/BuildingAndRunning.md) Hermes, remembering to git checkout the [static_h](https://github.com/facebook/hermes/tree/static_h) branch after cloning / before building

leotm: i'll get back to you @erights with repro's on RN 71+ and running them with Flipper and Hermes Debugger
debugging works well for me on RN 71 but not above (if the same for you i'll file an upstream issue)

mhofman: > will proceed with changes
> perhaps with a comment like this?

The changes I suggested are actually an improvement over what's currently there, and should be kept regardless of Hermes.

leotm: updated ^ and tested working on RN 71 (containing Hermes bug) on Android
with SES shim built from current/suggested changes
(`dist/ses.cjs` generated from `yarn clean && yarn build`)

packages/ses/src/tame-symbol-constructor.js
```diff
@@ -37,11 +38,9 @@ export const tameSymbolConstructor = () => {
   const OriginalSymbol = Symbol;
   const SymbolPrototype = OriginalSymbol.prototype;
 
-  const SharedSymbol = {
-    Symbol(description) {
-      return OriginalSymbol(description);
-    },
-  }.Symbol;
+  // Bypass Hermes bug, fixed in: https://github.com/facebook/hermes/commit/00f18c89c720e1c34592bb85a1a8d311e6e99599
+  // Increase JS spec fidelity by no longer defining SharedSymbol as an object literal short-hand method
```
mhofman, line 42:
```suggestion
  // Make a "copy" of the primordial [Symbol "constructor"](https://tc39.es/ecma262/#sec-symbol-description) which maintains all observable behavior. The primordial explicitly throws on `[[Construct]]` and has a `[[Call]]` which ignores the receiver. Binding also maintains the `toString` source as a native function. The `name` is restored below when copying own properties.
```

mhofman (approved):
Looks good, but since I proposed the change, I'd like @erights to approve.

Also suggestion on the comment, since we don't currently use bind anywhere for this purpose.

packages/ses/src/tame-symbol-constructor.js
leotm: > I've never run Hermes before. I'm on a Mac. What's the best way for me to reproduce this? Possible to reproduce under a debugger, as I saw in your screenshots?

> i'll get back to you @erights with repro's on RN 71+ and running them with Flipper and Hermes Debugger on Mac debugging works well for me on RN 71, but not above versions (if the same for you i'll file an upstream issue)

@erights here's the repro with @mhofman's fix applied in a [commit ](https://github.com/leotm/RN07117SES/commit/7c4c043266f85978f3a86c33249920b886d6571e)
https://github.com/leotm/RN07117SES

and i've added a ReadMe instructions to run it on Mac, debug with Flipper and repro the issues tracked in
- https://github.com/endojs/endo/issues/1891

kriskowal (commented):
This also looks good to me and I defer to @erights for the final stamp.

erights:
(rerunning failed CI jobs)

leotm:
passing ^ looks good to approve then merge

erights (approved):
LGTM, thanks!

leotm:
sry @kriskowal looks like i added an extra mandatory approval required by you ^
i don't have merge auth so Zb's enabled auto-merge

mhofman:
I am honestly unsure what the unmet merge requirement is. Will update the branch just in case

PR #2208, opened 2024-04-11
refactor(daemon): Factor out internal facets

kriskowal:
This is a simplifying refactor that reduces {internal, external} controller facets to just the (external) value. In preparation, removes the last vestiges of internal facets on worker and agent. Then mops up by collapsing the controller logic and structures.

### DIFF
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -109,6 +109,8 @@ const makeDaemonCore = async (
   } = powers;
   const { randomHex512 } = cryptoPowers;
   const contentStore = persistencePowers.makeContentSha512Store();
+  /** @type {WeakMap<object, import('@endo/eventual-send').ERef<import('./worker.js').WorkerBootstrap>>} */
+  const workerDaemonFacets = new WeakMap();
   /**
    * Mutations of the formula graph must be serialized through this queue.
    * "Mutations" include:
@@ -313,9 +315,12 @@ const makeDaemonCore = async (
       {},
     );
 
+    // @ts-expect-error Evidently not specific enough.
+    workerDaemonFacets.set(worker, workerDaemonFacet);
+
     return {
       external: worker,
-      internal: workerDaemonFacet,
+      internal: undefined,
     };
   };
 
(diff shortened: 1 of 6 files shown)


packages/daemon/src/host.js
```diff
@@ -384,7 +384,7 @@ export const makeHostMaker = ({
      * @param {import('./types.js').MakeHostOrGuestOptions} [opts]
      * @returns {Promise<{id: string, value: Promise<import('./types.js').EndoHost>}>}
      */
-    const makeHost = async (
+    const makeChildHost = async (
```
rekmarks, line 387: +1

packages/daemon/test/test-endo.js
```diff
@@ -1011,7 +1011,7 @@ test('name and reuse inspector', async t => {
 });
 
 // Regression test for https://github.com/endojs/endo/issues/2021
-test.failing('eval-mediated worker name', async t => {
+test('eval-mediated worker name', async t => {
```
rekmarks, line 1014: 🎉

packages/daemon/src/daemon.js
```diff
@@ -501,35 +466,30 @@ const makeDaemonCore = async (
    * @param {import('./types.js').Formula} formula
    * @param {import('./types.js').Context} context
    */
-  const makeControllerForFormula = async (
-    id,
-    formulaNumber,
-    formula,
-    context,
-  ) => {
+  const evaluateFormula = async (id, formulaNumber, formula, context) => {
```
rekmarks, line 469: Consider: `makeFormula()`

packages/daemon/src/daemon.js
```diff
@@ -746,23 +687,25 @@ const makeDaemonCore = async (
    * @param {string} id
    * @param {import('./types.js').Context} context
    */
-  const makeControllerForId = async (id, context) => {
+  const evaluateFormulaForId = async (id, context) => {
```
rekmarks, line 690: Consider: `makeFormulaForId()` or something along those lines. I'm not sold on the `FormulaForId` suffix, which doesn't seem to illuminate what distinguishes this function from `evaluateFormula` / `makeFormula`, since both take the id as an argument.

rekmarks (approved):
LGTM. Some optional suggestions.

packages/daemon/src/daemon.js
kriskowal: Good points all around.

The first name I considered was `makeFormulaForId` (just removing the `Controller`) and I hesitated because it makes a value and does not make a formula. It uses the formula to make a value, which struck me as indistinguishable from evaluating a node in any other abstract syntax, which led me to `evaluateFormula`.

The distinction between `evaluateFormula` vs `evaluateFormulaForId` is going to haunt me. The latter presumes that it can find the formula for the identifier, and the former requires that to be spelled out, and all roads flow to `evaluateFormula`.

I’m going to merge for now and continue to think and discuss this with you.

PR #2209, opened 2024-04-12
refactor(daemon): Maker table

kriskowal:
Converts the maker if/else chain into a lookup table for faster dispatch and better type hints. The new maker table type identified the extraneous WebBundleFormula left over from an earlier refactor.

### DIFF
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -460,52 +460,44 @@ const makeDaemonCore = async (
     );
   };
 
-  /**
-   * @param {string} id
-   * @param {string} formulaNumber
-   * @param {import('./types.js').Formula} formula
-   * @param {import('./types.js').Context} context
-   */
-  const evaluateFormula = async (id, formulaNumber, formula, context) => {
-    if (formula.type === 'eval') {
-      return makeEval(
-        formula.worker,
-        formula.source,
-        formula.names,
-        formula.values,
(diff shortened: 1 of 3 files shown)


packages/daemon/src/daemon.js
```diff
@@ -816,21 +843,6 @@ const makeDaemonCore = async (
...
-      provideController(id)
-    );
-    return controller.value.then(value => {
-      // Release the value to the public only after ensuring
-      // we can reverse-lookup its nonce.
-      if (typeof value === 'object' && value !== null) {
-        idForRef.add(value, id);
-      }
-      return value;
-    });
```
kriskowal, line 831: This was duplicative. `provideController` does this already.

packages/daemon/src/types.d.ts
```diff
@@ -347,6 +340,17 @@ export interface Controller<Value = unknown> {
...
 
+export type FormulaMaker<F extends Formula> = (
+  formula: F,
+  context: Context,
+  id: string,
+  number: string,
+) => unknown;
+
+export type FormulaMakerTable = {
+  [T in Formula['type']]: FormulaMaker<{ type: T } & Formula>;
```
kriskowal, line 351: If you can read this, you do not need type goggles.

rekmarks (approved):
LGMT, no notes.

PR #2210, opened 2024-04-12
refactor(daemon): Parameterize provide type assertion

kriskowal:
This change adds a vestigial second argument to `provide` that just hints to TypeScript the expected return type, resulting is much less verbose type narrowing. I’ve also advanced `provide` higher in the flow so it can eat all the forward-reference lint comments.

### DIFF
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -252,6 +252,14 @@ const makeDaemonCore = async (
   /** @type {import('./types.js').WeakMultimap<Record<string | symbol, unknown>, string>['get']} */
   const getIdForRef = ref => idForRef.get(ref);
 
+  /** @type {import('./types.js').Provide} */
+  const provide = (id, _expectedType) =>
+    /** @type {any} */ (
+      // Behold, unavoidable forward-reference:
+      // eslint-disable-next-line no-use-before-define
+      provideController(id).value
+    );
+
   /**
    * @param {string} workerId512
    */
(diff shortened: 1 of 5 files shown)


packages/daemon/src/types.d.ts
```diff
@@ -718,6 +717,37 @@ type FormulateNumberedHostParams = {
...
+    : T extends 'worker'
+    ? EndoWorker
+    : T extends 'agent'
+    ? EndoAgent
+    : T extends 'guest'
+    ? EndoGuest
+    : T extends 'invitation'
+    ? Invitation
+    : unknown
+>;
```
rekmarks, line 749: A couple of things regarding this one. First, I propose we reimplement it as follows:

1. Establish a private type that only exists to map `Formula['type']` to the type of the value instantiated by a formula:

```ts
type FormulaForValue = {
  endo: EndoBootstrap,
  host: EndoHost,
  guest: EndoGuest,
  worker: EndoWorker,
  // ...
};
```

2. Replace the ternary chain above with:
```ts
export type Provide = <T extends Formula['type'], U extends FormulaForValue[T]>(
  id: string,
  type?: T,
) => Promise<U>;
```

This has the benefit of being more succinct and partially self-validating since the definition of `Provide` errors if `FormulaForValue` does not have a key for every value in the string literal union `Formula['type']`. (It's still possible for `keyof FormulaForValue` to superset `Formula['type']`, but that will cause type errors at the point of use.)

Second, as it happens, `Formula['type']` in fact **does not** include all of the string literals used as the second argument to `provide()`, specifically the following:
- `agent`
- `hub`
- `network`

I had thought that there was a bijection between "formula types" and "values that can be provided". I think we should attempt to eliminate this discrepancy if possible. If we don't want to do that now, we can implement my above suggestion with the following band-aid:
```ts
type Providables = Formula['type'] & 'agent' | 'hub' | 'network';
// FormulaForValue modified as necessary
export type Provide = <T extends Providables, U extends FormulaForValue[T]>( /* ... */
```

rekmarks (commented):
This is a great change. Below, I provide a suggestion and one finding concerning `Provide`.

packages/daemon/src/daemon.js
```diff
@@ -252,6 +252,14 @@ const makeDaemonCore = async (
   /** @type {import('./types.js').WeakMultimap<Record<string | symbol, unknown>, string>['get']} */
   const getIdForRef = ref => idForRef.get(ref);
 
+  /** @type {import('./types.js').Provide} */
+  const provide = id =>
```
rekmarks, line 256: Just for explicitness:
```suggestion
  const provide = (id, _formulaType) =>
```

rekmarks (approved):
LGTM!

PR #2212, opened 2024-04-15
feat(daemon): Rename Locator to Config, improve type exports

rekmarks:
Renames the daemon's `Locator`, used to specify its file and storage paths, to `Config` to get rid of the name collision with endo locator strings. Also adds return values to functions in exported types, and removes redundant definition of `Config` type.


kriskowal:
There is also a touch-point in `cli`: the file named `locator.js` exports effectively the config.

kriskowal (approved):

PR #2213, opened 2024-04-15
refactor(daemon): Remove extraneous worker process args

rekmarks:
Removes extraneous arguments to worker processes.


kriskowal (approved):
I changed redundant for extraneous in the PR title and description but not your commit. For your consideration.

PR #2217, opened 2024-04-17
feat(daemon): Ensure a single connection between peers

kriskowal:
This feature ensures that any pair of pet dæmons will only be connected by a single CapTP connection to ensure that identical remoteable objects round-trip between them. In this framework it is necessary to _partially_ solve the problem of “crossed hellos”, such that two nodes that simultaneous start up and suddenly wish to be connected to each other will race to create the first connection. One of these daemons must drop an established connection and invalidate incarnations that depend upon it. We settle this question by biasing the behavior based on the relative values of their node identifiers.

This does not fully solve the experience of crossed hellos. If hellos get crossed, the loser of the race will get cancelled and may not gracefully recover. Evidence suggests this can manifest as data-lock where it would ideally result in an explicit cancellation exception. More work is necessary.

This does not solve the more general problem that three-party-hand-off would solve: a remoteable that takes a trip through a third party will not return as an identical object. This is necessary toward that end, but there will need to be shared state among all open CapTP sessions.

This also raises another problem not solved in this change: when the loser of the race to cross hellos abandons an incarnation, it should not merely delete the incarnation but should immediately reïncarnate the formula subgraph that depends upon it.

### DIFF
--- a/packages/daemon/src/remote-control.js
+++ b/packages/daemon/src/remote-control.js
@@ -1,0 +1,304 @@
+// @ts-check
+
+/**
+ * @param {string} localNodeId
+ */
+export const makeRemoteControlProvider = localNodeId => {
+  /** @type {Map<string, import('./types.js').RemoteControl>} */
+  const remoteControls = new Map();
+
+  /** @param {string} remoteNodeId */
+  const makeRemoteControl = remoteNodeId => {
+    /** @type {import('./types.js').RemoteControlState} */
+    let state;
+
+    // In this state, we have received a remoteGateway from an ingress
+    // connection (and provided our local gateway to them.)
+    // We do not have a pending outbound connection attempt.
+    /**
+     * @type {(
+     *   remoteGateway: Promise<import('./types.js').EndoGateway>,
+     *   cancel: (error: Error) => void | Promise<void>,
+     *   cancelled: Promise<never>,
+     * ) => import('./types.js').RemoteControlState}
+     */
+    const accepted = (remoteGateway, cancelCurrent, currentCancelled) => {
+      return {
+        accept(
+          _proposedRemoteGateway,
+          proposedCancel,
+          _proposedCancelled,
+          proposedDispose,
+        ) {
+          // And we receive an inbound connection.
+          // There are two possibilities:
+          // The sender raced multiple outbound connections and this connection
+          // lost the race.
+          // Or, the sender restarted but left a prior connection half-open.
+          // We consider the race to be the common consideration and that
+          // replacing peers after establishing a connection would be far to
+          // disruptive.
+          // TODO: For the case where we leave a peer wedged half-open, we
+          // will need health checks.
+          Promise.resolve(
+            proposedCancel(new Error('Already accepted a connection.')),
+          ).then(proposedDispose);
+          return accepted(remoteGateway, cancelCurrent, currentCancelled);
+        },
+        connect(
+          _getProposedRemoteGateway,
+          proposedCancel,
+          proposedCancelled,
+          proposedDispose,
+        ) {
+          // Use the gateway we already have.
+          // Bind the fates of the current peer incarnation and the inbound
+          // connection.
+          Promise.all([
+            currentCancelled.catch(proposedCancel),
+            proposedCancelled.catch(cancelCurrent),
+          ])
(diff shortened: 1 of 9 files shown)


packages/daemon/src/remote-control.js
```diff
@@ -0,0 +1,227 @@
...
+
+    // In this state, we have received a remoteGateway from an ingress
+    // connection (and provided our local gateway to them.)
+    // We do not have a pending outbound connection attempt.
+    /**
+     * @type {(
+     *   remoteGateway: Promise<import('./types.js').EndoGateway>,
+     *   cancel: (error: Error) => void,
+     *   cancelled: Promise<never>,
+      ) => import('./types.js').RemoteControlState} */
```
rekmarks, line 23: Formatting issue:
```suggestion
     *   cancelled: Promise<never>,
     * ) => import('./types.js').RemoteControlState}
     */
```

packages/daemon/src/remote-control.js
```diff
@@ -0,0 +1,227 @@
...
+      };
+    };
+
+    // We have an active outbound connection.
+    /**
+     * @type {(
+     *   remoteGateway: Promise<import('./types.js').EndoGateway>,
+     *   cancel: (error: Error) => void,
+     *   cancelled: Promise<never>,
+      ) => import('./types.js').RemoteControlState} */
```
rekmarks, line 60: Same formatting issue as above:
```suggestion
     *   cancelled: Promise<never>,
     * ) => import('./types.js').RemoteControlState}
     */
```

packages/daemon/test/test-endo.js
rekmarks, line 1: All of these tests also pass on `master`. We should probably add a test for crossed hellos that passes here but fails on `master`.

rekmarks (changes requested):
Looks good, but we should probably add a test that conclusively demonstrates that we've solved crossed hellos (see inline comment). Also some nits.

packages/daemon/test/test-endo.js
```diff
@@ -1363,6 +1366,37 @@ test('read remote value', async t => {
...
+      const token = Far('Token', {});
+      E(echoer).echo(token).then(allegedlyIdenticalToken =>
+        token === allegedlyIdenticalToken
+      );
+    `,
+    ['echoer'],
+    ['echoer'],
+  );
+  t.assert(survivedEcho);
+});
```
kriskowal, line 1398: This test fails on master because master does not attempt to deduplicate the A->B and B<-A CapTP connections.

kriskowal: I was mistaken. This was passing. I created a more specific test that exercised a hello from the remote and integrated the remote control into the provider.

packages/daemon/test/test-endo.js
```diff
@@ -1348,9 +1348,8 @@ test('read remote value', async t => {
   const hostA = await prepareHostWithTestNetwork(t);
   const hostB = await prepareHostWithTestNetwork(t);
 
-  // introduce nodes to each other
+  // Introduce A to B (such that B becomes connected to A consequently.)
   await E(hostA).addPeerInfo(await E(hostB).getPeerInfo());
-  await E(hostB).addPeerInfo(await E(hostA).getPeerInfo());
```
kriskowal, line 1353: This explicitly aggravates a crossed-hello and results in data-lock. I would like to punt the general solution in hopes that daemons will rarely restart at the same time. I am aware that this is foolish in the long term, but crossed hellos are hard and time is running short for our proof-of-concept demo.

rekmarks (approved):
This is good for now 👍

packages/daemon/src/remote-control.js
```diff
@@ -0,0 +1,229 @@
...
+    // We have an active outbound connection.
+    /**
+     * @type {(
+     *   remoteGateway: Promise<import('./types.js').EndoGateway>,
+     *   cancel: (error: Error) => void,
+     *   cancelled: Promise<never>,
+     * ) => import('./types.js').RemoteControlState}
+     */
+    const connected =
+      remoteNodeId > localNodeId
```
rekmarks, line 64: @kriskowal did you say that the datalock happens about half the time? Observation: this will be true about half the time.

PR #2219, opened 2024-04-17
feat(daemon): Invitations

kriskowal:
This change introduces a user workflow for connecting daemons.

```
alice> endo invite bob > invitation.link
*alice sends bob invitation.link*
bob> endo accept alice < invitation.link
```

The two parties can now exchange mail.

```
alice> endo send bob 'Hi, here is @power'
bob> endo inbox
0. alice sent “Hi, here is @power”
bob> endo adopt 0 power 
```

Here's a transcript for testing this in `packages/cli/demo`:

```
> endo make counter.js --name counter
Object [Alleged: Counter] {}

> endo mkhost other other-agent
Object [Alleged: EndoHost] {}

> endo invite bob > invitation.link

> endo accept alice --as other-agent < invitation.link

> endo send bob 'Hi, here is @counter'

> endo show counter
Object [Alleged: Counter] {}

> endo show counter --as other-agent
CapTP cli exception: (RemoteTypeError(error:captp:Endo#20001)#1)
RemoteTypeError(error:captp:Endo#20001)#1: Unknown pet name: "counter"
...

> endo adopt 0 counter --as other-agent

> endo show counter --as other-agent
Object [Alleged: Counter] {}
```

### DIFF
--- a/packages/cli/src/commands/accept.js
+++ b/packages/cli/src/commands/accept.js
@@ -1,0 +1,20 @@
+/* global process */
+import os from 'os';
+import { E } from '@endo/far';
+import { withEndoAgent } from '../context.js';
+
+const fromAsync = async iterable => {
+  const all = [];
+  for await (const iterand of iterable) {
+    all.push(iterand);
+  }
+  return all;
+};
+
+export const accept = async ({ guestName, agentNames }) => {
+  process.stdin.setEncoding('utf-8');
+  const invitationLocator = (await fromAsync(process.stdin)).join('').trim();
+  return withEndoAgent(agentNames, { os, process }, async ({ agent }) => {
+    await E(agent).accept(invitationLocator.trim(), guestName);
+  });
+};
--- a/packages/daemon/src/host.js
+++ b/packages/daemon/src/host.js
@@ -1,10 +1,12 @@
 // @ts-check
+/// <reference types="ses"/>
 
 import { E } from '@endo/far';
 import { makeExo } from '@endo/exo';
 import { M } from '@endo/patterns';
 import { makeIteratorRef } from './reader-ref.js';
 import { assertPetName, petNamePathFrom } from './pet-name.js';
+import { parseId, formatId } from './formula-identifier.js';
 import { makePetSitter } from './pet-sitter.js';
 import { makeDeferredTasks } from './deferred-tasks.js';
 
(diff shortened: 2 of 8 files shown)


rekmarks:
I'm not sure if the demo is missing some steps or if I'm doing something incorrectly. When I do `endo send bob 'Hi, here is @power'` (`@counter` in my case), I get the following error:
```
CapTP cli exception: (RemoteTypeError(error:captp:Endo#20001)#1)
RemoteTypeError(error:captp:Endo#20001)#1: target has no method "receive", has ["__getInterfaceGuard__","__getMethodNames__","accept","locate"]
...
```

Because `bob` is at that point an invitation:
```
endo show bob
Object [Alleged: Invitation] {}
```

I first made a guest / host named `bob` / `bob-agent`. `endo invite bob > invitation.link` overwrites `bob` with the invitation, and `bob` still refers to the invitation after `endo accept alice --as bob < invitation.link`.

Two questions:
1. Is `endo invite bob` supposed to overwrite `bob`? Shouldn't there be a different pet name for the invitation?
2. Am I doing something wrong?

kriskowal:
> 1. Is `endo invite bob` supposed to overwrite `bob`? Shouldn't there be a different pet name for the invitation?
> 2. Am I doing something wrong?

`endo invite bob` is supposed to create the invitation named bob.

`endo accept alice` is supposed to cause alice to overwrite the bob name with alice’s handle.

I don’t think you’re doing something wrong, except perhaps waiting for the acceptance.

We could conceivably cause invitations to queue up messages destined for the to-be-identified recipient.

rekmarks:
My problem turned out to be completely endogenous.

packages/daemon/src/host.js
```diff
@@ -432,6 +436,84 @@ export const makeHostMaker = ({
...
+    /**
+     * @param {string} invitationLocator
+     * @param {string} guestName
+     */
+    const accept = async (invitationLocator, guestName) => {
+      const url = new URL(invitationLocator);
+      const nodeNumber = url.hostname;
+      const invitationNumber = url.searchParams.get('id');
+      const remoteHandleNumber = url.searchParams.get('from');
+      const addresses = url.searchParams.getAll('at');
```
rekmarks, line 474: Should we also validate this? All other extracted URL components appear to be validated. At present, the ultimately thrown error is "no supported networks", which is not very helpful if `addresses` were somehow empty.

rekmarks (approved):
LGTM! Two suggestions.

packages/cli/src/commands/accept.js
```diff
@@ -0,0 +1,20 @@
+/* global process */
+import os from 'os';
+import { E } from '@endo/far';
+import { withEndoAgent } from '../context.js';
+
+const fromAsync = async iterable => {
```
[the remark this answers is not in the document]
kriskowal: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fromAsync

packages/daemon/src/host.js
kriskowal: I think this code needs to be very temporary. A lot of this should be done automatically by lower-level mechanisms. Eg., `E(hub).write(name, locator)` should extract `{ node, formula, type, addresses }` from the locator, submit `{ node, addresses }` to the known peers store, capture `{node, formula, type }` in the pet store, and likewise `provide` should operate on only `{ node, formula }`.

PR #2223, opened 2024-04-17
feat(pass-style): generalize passable errors, throwables

### LINKED ISSUE #2891: self-identification for errors

`@endo/compartment-mapper` and `ses` both throw many exceptions (mostly `Error` and a dusting of other intrinsics).

As a consumer and/or a developer of these libraries, when an exception is thrown, I want to:

- Easily identify the subsystem(s) involved, whether debugging or no
- Have some notion of intent, at-a-glance
- Safely trap errors and programmatically take actions depending on the type of error
- Grep better

## Description of the Design

I know of two (2) strategies for this, which are not mutually exclusive:

1. Subclassing builtin `Error` classes
2. Setting a `code` (typically a string constant, e.g., `ENO_TEA`) and/or `errno` (integers; [probably negative](https://docs.libuv.org/en/v1.x/errors.html)) property on an `Error` object prior to throw.

I prefer using a subclass _and_ providing a `code` field.  `instanceof` has gotchas, but a subclass communicates intent for the Endo developer and enables further customization of behavior (URLs to docs; redaction; other formatting; additional metadata).  A `code` field is common in the ecosystem, but `errno` is mostly a node-ism.

The `name` field _should_ also be set for these subclasses (as intrinsic subclasses of `Error` do):

```js
class GooseError extends Error {}

try {
  throw new GooseError()
} catch (err) {
  err.name // Error
}

class GanderError extends Error { name = 'GanderError' }
try {
  throw new GanderError()
} catch (err) {
  err.name // GanderError
}
```

> In the weeds here: 
> 
> The signature constructor signatures should probably be compatible with the constructor signature of `Error`, for the sake of familiarity and ease of implementation:
>
> ```ts
> interface ErrorConstructor {
>   new (message?: string, options?: ErrorOptions): Error;
> }
> 
> interface ErrorOptions {
>   cause?: unknown;
> }
> ```
>
> If we wish to accept additional metadata, we can just stuff it in the second parameter by extending  `ErrorOptions`. 
> 
> I'm not sure if we want to set `code` on the prototype or just use a class field. It _is_ handy to have `code`` available as a static field as well (e.g. `GooseError.code === 'EHORRIBLE_GOOSE'`).
> 
> Unsure if we should be using getters or `Object.defineProperty` on these fields; it doesn't seem necessary to me, but I'm open to suggestions!

**Implementation does not need to happen all at once;** we _don't_ need to define all the possible errors and codes up-front _and then_ do a full sweep through the codebase looking for replacement candidates.  For this issue, I'm hoping we can settle on a design.

> [!IMPORTANT]
> Remember that `class` demands use of `new` (sorry @erights):
>
> ```js
> class GooseError extends Error {}
>
> try {
>   throw GooseError()
> } catch (err) {
>   err.name // TypeError
>   err.message // Class constructor GooseError cannot be invoked without 'new'
> }
> ```
> 
> I suppose we could avoid `class` and define them the _old-fashioned_ way, though!

## Security Considerations

I know there are some concerns in `ses` around the information exposed in `Error` objects. I don't know about it otherwise; for all I know, it could be the _reason_ that exceptions are somewhat murky.  That said, I don't _think_ `@endo/compartment-mapper` faces or should face the same restrictions.

## Scaling Considerations

Throwing in a hot code path has negative performance impacts, but I am not suggesting we throw _more_ exceptions—only that we throw _better_ ones.

The usual warnings about deep inheritance trees apply.

## Test Plan

Depends on implementation, but we could certainly update extant tests to inspect error codes instead of error messages.  

## Compatibility Considerations

AFAIK subclassing `Error` is pretty straightforward in modern JS.

## Upgrade Considerations

We may want to notify users that the _code_ or _class_ is part of the public API and thus will adhere to SemVer, but the error _message_ may not.  This also makes i18n/l10n more easily achievable, if that ever wants to happen.

erights:
Just placeholder for now. Not yet expected to work or almost work.

Staged on #2841

closes: #2891 
refs: #2429 

### Upgrade Considerations


mhofman:
We should consider only allowing a passable error as a throwable so that we maintain the ability to reconstitute distributed stack traces. Only the top level needs to be a passable error, nested values can be any copy data or passable error.

erights:
See also https://github.com/Agoric/agoric-sdk/issues/10375

PR #2224, opened 2024-04-18
refactor(daemon): Use @import

kriskowal:
Largely mechanical change to use `@import` instead of `@type {import(...`.


rekmarks (approved):
Gods be good, there is a better way.

There are a couple of extremely verbose `import()` types in `cli/install.js` and `cli/make.js` that may be worth addressing, otherwise LGTM.

PR #2225, opened 2024-04-19
refactor(daemon): Consolidate interface definitions and cleanup

kriskowal:
These individually reviewable commits consolidate some recurring patterns, move interface definitions into a separate module, adds more explanatory comments, and fixes some lingering errors.

### DIFF
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -4,7 +4,6 @@
 /* global setTimeout, clearTimeout */
 
 import { makeExo } from '@endo/exo';
-import { M } from '@endo/patterns';
 import { E, Far } from '@endo/far';
 import { makePromiseKit } from '@endo/promise-kit';
 import { q } from '@endo/errors';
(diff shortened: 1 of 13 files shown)


packages/daemon/src/interfaces/web.js
rekmarks, line 1: Hm, why not `interfaces/index.js` and `interfaces/web.js`? Indeed, why a separate file at all?

packages/daemon/src/daemon.js
```diff
@@ -186,46 +189,34 @@ const makeDaemonCore = async (
...
-  );
-  const knownPeersId = formatId({
-    number: knownPeersFormulaNumber,
-    node: localNodeId,
-  });
-  await persistencePowers.writeFormula(knownPeersFormulaNumber, {
-    type: 'known-peers-store',
-  });
+  // We generate formulas for some entities that are presumed to exist
+  // because they are parts of the daemon's root object.
```
rekmarks, line 193:
```suggestion
  // because they are part of the daemon's root object.
```

rekmarks (approved):
LGTM! A couple of inline comments. The `interfaces` refactor is most appreciated.

packages/daemon/src/interfaces/web.js
kriskowal: I loathe the Node.js `index.js` lookup because it assumes a search path that doesn’t translate to the web. Of course, that’s immaterial since we’re never going to be free of a bundler or import-map, but I prefer to live in the make-believe so as to be ready for change.

I separated `interfaces/web.js` because these interfaces are not canon to the daemon itself but a contract between the web server and client, and the web client should shake out the interfaces it doesn’t need in its bundle.

packages/daemon/src/daemon.js
kriskowal: I think the cardinality of “formulas”, “entities”, “are”, and “parts” should agree here.

PR #2226, opened 2024-04-19
feat(daemon): Store passable values

kriskowal:
This change adds a `E(host).storeValue(value, name)` API allowing us to store passable data, provided it consists of copy-data and references with known formulas.


rekmarks (commented):
This looks good (`@endo/marshal` is cool!), but I'm hesitant to approve because I'm anxious about introducing a formula type named "value", given our prolific use of that word in the sense of "incarnated formula". It might be worth the trouble because it is a good word choice for the use case, but have we exhausted alternatives, e.g. "data"?

kriskowal:
> This looks good (`@endo/marshal` is cool!), but I'm hesitant to approve because I'm anxious about introducing a formula type named "value", given our prolific use of that word in the sense of "incarnated formula". It might be worth the trouble because it is a good word choice for the use case, but have we exhausted alternatives, e.g. "data"?

I think this is worth some conversation. Data would be appropriate if these formulas couldn’t capture formula remoteables, e.g., just passStyle cap-data.

kriskowal:
> > This looks good (`@endo/marshal` is cool!), but I'm hesitant to approve because I'm anxious about introducing a formula type named "value", given our prolific use of that word in the sense of "incarnated formula". It might be worth the trouble because it is a good word choice for the use case, but have we exhausted alternatives, e.g. "data"?
> 
> I think this is worth some conversation. Data would be appropriate if these formulas couldn’t capture formula remoteables, e.g., just passStyle cap-data.

We settled on “marshal” for the formula type out-of-band.

rekmarks (approved):
Approved conditioned on resolving old comment about new unit test.

PR #2227, opened 2024-04-19
feat(cli,daemon): Add `move` CLI command

rekmarks:
Adds a command `move` corresponding to the daemon directory's `move()`, with pet name path support. Replaces the existing `rename` command, which had been broken for some time.

Also ensures that the directory's `move()` method immediately deletes the "from" name mapping. Adds a test for the same. This mapping was already removed from disk and would not persist after restarts.

### DIFF
--- a/packages/cli/src/endo.js
+++ b/packages/cli/src/endo.js
@@ -297,13 +297,14 @@ export const main = async rawArgs => {
     });
 
   program
-    .command('rename <from> <to>')
+    .command('move <from> <to>')
+    .alias('mv')
     .description('change the name for a value')
     .option(...commonOptions.as)
-    .action(async (fromName, toName, cmd) => {
+    .action(async (fromPath, toPath, cmd) => {
       const { as: agentNames } = cmd.opts();
-      const { rename } = await import('./commands/rename.js');
-      return rename({ fromName, toName, agentNames });
+      const { move } = await import('./commands/move.js');
+      return move({ fromPath, toPath, agentNames });
     });
 
   program
--- a/packages/daemon/src/directory.js
+++ b/packages/daemon/src/directory.js
@@ -174,6 +174,7 @@ export const makeDirectoryMaker = ({
       const { hub: fromHub, name: fromName } =
         await lookupTailNameHub(fromPath);
       const { hub: toHub, name: toName } = await lookupTailNameHub(toPath);
+
       if (fromHub === toHub) {
         // eslint-disable-next-line no-use-before-define
         if (fromHub === directory) {
@@ -183,13 +184,14 @@ export const makeDirectoryMaker = ({
         }
         return;
       }
+
       const id = await fromHub.identify(fromName);
       if (id === undefined) {
         throw new Error(`Unknown name: ${q(fromPath)}`);
       }
-      const removeP = fromHub.remove(fromName);
-      const addP = toHub.write([toName], id);
-      await Promise.all([addP, removeP]);
+      // First write to the "to" hub so that the original name is preserved on the
+      // "from" hub in case of failure.
+      await toHub.write([toName], id).then(() => fromHub.remove(fromName));
     };
 
     /** @type {EndoDirectory['copy']} */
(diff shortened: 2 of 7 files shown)


packages/cli/src/endo.js
```diff
@@ -297,13 +297,14 @@ export const main = async rawArgs => {
     });
 
   program
-    .command('rename <from> <to>')
+    .command('move <from> <to>')
+    .alias('mv')
```
kriskowal, line 301: Should follow up with analogous aliases on other commands.

packages/daemon/test/test-endo.js
```diff
@@ -354,6 +354,21 @@ test('store with name', async t => {
   }
 });
 
+test('move renames value', async t => {
```
kriskowal, line 357: Would enjoy tests that cover moving to and from other hubs. Should verify that we fall through to `rename` only if the to and from hubs fulfill to identical values and otherwise _carefully_ add _then_ delete, so that failure leaves the name intact at the source. And a test for failure to add to a bogus target hub.

kriskowal (changes requested):

kriskowal:
The design space for `move` interacts with the representation of identifiers in the pet dæmon network. Moving should be approximately equivalnt to `E(powers).locate(...toPath).write(toName, E(powers).locate(...fromPath))`. We currently use `identify` so this operation fails to introduce the addresses of the node housing the `from` entity to the node housing the `to` hub/directory/agent.

I’m imagining that `locate` and `write` will conspire at the level of a pet-sitter to take a “pet locator” and divide its (overlapping) parts between a “pet id” and “peer locator”, submitting the “pet id” to the underlying “pet store” and submitting the “peer locator” to the node’s “peer store”. So `locate` would combine these components on-the-fly and `write` would separate them on-the-fly. I think that would allow a reference to move from one node to another robustly.

packages/daemon/src/directory.js
```diff
@@ -174,22 +174,26 @@ export const makeDirectoryMaker = ({
...
         throw new Error(`Unknown name: ${q(fromPath)}`);
       }
-      const removeP = fromHub.remove(fromName);
-      const addP = toHub.write([toName], id);
-      await Promise.all([addP, removeP]);
+      // First write to the "to" hub so that the original name is preserved on the
+      // "from" hub in case of failure.
+      await E(toHub)
+        .write([toName], id)
+        .then(() => E(fromHub).remove(fromName));
```
kriskowal, line 196: Equivalent.
```suggestion
      await E(toHub).write([toName], id);
      await E(fromHub).remove(fromName);
```

packages/daemon/test/test-endo.js
```diff
@@ -439,6 +439,160 @@ test('store with name', async t => {
...
+
+  t.false(await E(host).has('ten'));
+  t.true(await E(host).has('zehn'));
+});
+
+test('move renames value, overwriting the "to" name', async t => {
+  const { host } = await prepareHost(t);
+
+  await E(host).evaluate('MAIN', '10', [], [], 'ten');
+  await E(host).evaluate('MAIN', '"X"', [], [], 'decimus');
```
kriskowal, line 459: These can be `E(host).storeValue(10, 'ten')` etc now!

packages/daemon/test/test-endo.js
```diff
@@ -439,6 +439,160 @@ test('store with name', async t => {
...
+  t.false(await E(host).has('ten'));
+  t.true(await E(host).has('decimus'));
+
+  const decimusValue = await E(host).lookup('decimus');
+  t.is(decimusValue, 10);
+});
+
+test('move moves value, from the host to a different name hub', async t => {
+  const { host } = await prepareHost(t);
+  const nets = await E(host).lookup('NETS');
```
kriskowal, line 475: We don’t have `makeDirectory` yet? `NETS` is probably okay, but I wouldn’t count on it remaining that way for non-network shaped things.

kriskowal (approved):

packages/daemon/test/test-endo.js
rekmarks: Thank you for reminding me that `makeDirectory()` exists!

PR #2228, opened 2024-04-19
feat(cli): Add support for `--as` to `list` command

rekmarks:
Adds support for the `--as` flag to the CLI's `list` command.


rekmarks:
We don't actually need this; `endo list <agent-name>` provides the intended functionality. Maybe we'll add `--as` to all relevant commands for consistency purposes in the future.

Issue #2230, opened 2024-04-19
Feature request: Need to habitually test on pre-release forms of supported target platforms


erights:
https://github.com/endojs/endo/issues/2198 was erroneously closed by #2200 because #2198 only manifested on browsers and we manually tested #2200 on those browsers to "verify" that it was fixed.

However, we are not set up to run our normal `yarn test` CI tests habitually (or under CI) on those browser engines. We current do our regular `yarn test` CI testing, where everything was coming up green, but only because the most recent Node we CI is Node 20, and Node 20 does not yet have v8's problematic stack accessor behavior. I separately reopened #2198 until those old tests are fixed, in progress at #2229. But currently I can only verify that #2229 fixes it by running `yarn test` locally with Node 21.

So, feature request: We should as much as possible run all our normal tests on the pre-release form of all target platforms we support, so we get early warning when we're about to stop working on a release.

Fortunately, in this case, it looks like the `src/` fixes from #2200 do still actually fix the #2198 problem. The only remaining problem, being addressed in #2229, is to update the tests. This would indeed have caused CI failures as soon as we would upgrade to support Node 22, but only because the tests are wrong. The code being tested already seems good for Node 21 and thus presumably for Node 22, at least as far as these issues are concerned.

erights:
See also https://github.com/Agoric/agoric-sdk/issues/9265

erights:
@kriskowal , @frazarshad , I co-assigned this to the three of us.

@kriskowal I see you merged https://github.com/endojs/endo/pull/2231 despite the two CI failures https://github.com/endojs/endo/actions/runs/8760680740/job/24046036596 . What’s failing there is exactly what this PR was adding

```
Unable to find Node version '22.x' for platform linux and architecture x64.
```

That’s why I tried https://github.com/endojs/endo/pull/2233 and https://github.com/endojs/endo/pull/2234 . #2234 brings this back to green CI with 21.x, which it finds, rather than 22.x, which it does not.

Also, how did you merge this despite the CI failures? Isn’t that supposed to be what CI protects us from?

Could/should you back out #2231 while we make progress towards getting the state of #2234 reviewed and merged?

erights:
See also 
https://github.com/Agoric/agoric-sdk/pull/9267
https://github.com/Agoric/agoric-sdk/pull/9273
https://github.com/Agoric/agoric-sdk/pull/9274
and
https://github.com/Agoric/agoric-sdk/pull/9267#issuecomment-2067874367
[...]

PR #2231, opened 2024-04-19
test(ci): Node.js 22 in CI

kriskowal:
refs: #2230

## Description

This change adds `latest` to the CI [node setup action](https://github.com/actions/setup-node) so we can see upcoming breaking changes to our platform. The corresponding test results will not be required for merge.

Testing canary did not pan out. You can get `22-v8-canary` out of `actions/setup-node`, and you can add `yarn install --ignore-engines`, but `yarn` will still die with error if any dependency specifies `engines` since the canary versions do not match other version ranges, even broad ones like `>= 16`. This may be worth revisiting with newer versions of `yarn` or whatever succeeds it.

### Security Considerations

This increases our visibility into breaking changes on the Node.js platform.

### Scaling Considerations

This may prolong CI runs by a multiplicative factor if CI workers are contended.

None.

### Testing Considerations

Increases visibility into our future.

None.

### Upgrade Considerations

None.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,7 +77,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [18.x, 20.x]
+        node-version: [18.x, 20.x, 22.x]
         platform: [ubuntu-latest, windows-latest]
 
     steps:


erights:
The CI failures for ubuntu-latest look like the ones fixed by #2229, which is what's supposed to happen! Great!

erights (approved):
LGTM thanks!

PR #2232, opened 2024-04-21
fix(ses): `harden` hacks v8 `stack` own accessor problem

erights:
closes: #2198 
refs: #2230 #2200  https://chromium-review.googlesource.com/c/v8/v8/+/4459251 #2229 #2231 

## Description

Alternative to #2229 that just hacks `harden` to directly repair a problematic error own stack accessor property, replacing it with a data property.

### Security Considerations

Both before and after this PR, `passStyleOf` will reject errors with the v8 problematic error own stack accessor property, preventing the unsafety at stake here. However, this would mean that much existing code that used to be correct will break when run on a v8 with this problem.

### Scaling Considerations

 Avoids any extra overhead on platforms without this problem, including all platforms other than v8.

probably none. This PR essentially avoids the need to document the v8 problem that it masks.

### Testing Considerations

Only needed to repair one test to use `harden` rather than `freeze`, in a case where `harden` was more natural anyway.

This PR enables more errors to pass that check without further changes to user code. #2229 had similar goals, but would still require more changes to user code than this PR. This is demonstrated by all the test code in #2229 that needed to be fixed that does not need to be fixed in this PR.

### Upgrade Considerations

none

### DIFF
--- a/packages/ses/src/commons.js
+++ b/packages/ses/src/commons.js
@@ -294,3 +294,58 @@ export const noEvalEvaluate = () => {
   // See https://github.com/endojs/endo/blob/master/packages/ses/error-codes/SES_NO_EVAL.md
   throw TypeError('Cannot eval with evalTaming set to "noEval" (SES_NO_EVAL)');
 };
+
+// ////////////////// FERAL_STACK_GETTER FERAL_STACK_SETTER ////////////////////
+
+const er1StackDesc = getOwnPropertyDescriptor(Error('er1'), 'stack');
+const er2StackDesc = getOwnPropertyDescriptor(TypeError('er2'), 'stack');
+
+let feralStackGetter;
+let feralStackSetter;
+if (er1StackDesc && er2StackDesc && er1StackDesc.get) {
+  // We should only encounter this case on v8 because of its problematic
+  // error own stack accessor behavior.
+  // Note that FF/SpiderMonkey, Moddable/XS, and the error stack proposal
+  // all inherit a stack accessor property from Error.prototype, which is
+  // great. That case needs no heroics to secure.
+  if (
+    // In the v8 case as we understand it, all errors have an own stack
+    // accessor property, but within the same realm, all these accessor
+    // properties have the same getter and have the same setter.
+    // This is therefore the case that we repair.
+    typeof er1StackDesc.get === 'function' &&
+    er1StackDesc.get === er2StackDesc.get &&
+    typeof er1StackDesc.set === 'function' &&
+    er1StackDesc.set === er2StackDesc.set
+  ) {
+    // Otherwise, we have own stack accessor properties that are outside
+    // our expectations, that therefore need to be understood better
+    // before we know how to repair them.
+    feralStackGetter = er1StackDesc.get;
+    feralStackSetter = er1StackDesc.set;
+  } else {
+    throw TypeError('Error own stack accessor functions not as expected');
+  }
+}
+
+/**
+ * If on a v8 with the problematic error own stack accessor behavior,
+ * `FERAL_STACK_GETTER` will be the shared getter of all those accessors
+ * and `FERAL_STACK_SETTER` will be the shared setter. On any platform
+ * without this problem, `FERAL_STACK_GETTER` and `FERAL_STACK_SETTER` are
+ * both `undefined`.
+ *
+ * @type {(() => any) | undefined}
+ */
+export const FERAL_STACK_GETTER = feralStackGetter;
+
+/**
+ * If on a v8 with the problematic error own stack accessor behavior,
+ * `FERAL_STACK_GETTER` will be the shared getter of all those accessors
+ * and `FERAL_STACK_SETTER` will be the shared setter. On any platform
+ * without this problem, `FERAL_STACK_GETTER` and `FERAL_STACK_SETTER` are
+ * both `undefined`.
+ *
+ * @type {((newValue: any) => void) | undefined}
+ */
+export const FERAL_STACK_SETTER = feralStackSetter;
--- a/packages/ses/src/commons.js
+++ b/packages/ses/src/commons.js
@@ -321,8 +321,8 @@ if (er1StackDesc && er2StackDesc && er1StackDesc.get) {
     // Otherwise, we have own stack accessor properties that are outside
     // our expectations, that therefore need to be understood better
     // before we know how to repair them.
-    feralStackGetter = er1StackDesc.get;
-    feralStackSetter = er1StackDesc.set;
+    feralStackGetter = freeze(er1StackDesc.get);
+    feralStackSetter = freeze(er1StackDesc.set);
   } else {
     // See https://github.com/endojs/endo/blob/master/packages/ses/error-codes/SES_UNEXPECTED_ERROR_OWN_STACK_ACCESSOR.md
     throw TypeError(
--- a/packages/ses/src/commons.js
+++ b/packages/ses/src/commons.js
@@ -324,7 +324,10 @@ if (er1StackDesc && er2StackDesc && er1StackDesc.get) {
     feralStackGetter = er1StackDesc.get;
     feralStackSetter = er1StackDesc.set;
   } else {
-    throw TypeError('Error own stack accessor functions not as expected');
+    // See https://github.com/endojs/endo/blob/master/packages/ses/error-codes/SES_UNEXPECTED_ERROR_OWN_STACK_ACCESSOR.md
+    throw TypeError(
+      'Unexpected Error own stack accessor functions (SES_UNEXPECTED_ERROR_OWN_STACK_ACCESSOR)',
+    );
   }
 }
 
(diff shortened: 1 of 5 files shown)


erights:
CI for #2233 fails on Node 21 due to the v8 problems that this PR hacks around. CI for #2234 succeeds, validating that for at least these occurrences of the problem, this PR fixes them.

packages/ses/src/commons.js
```diff
@@ -294,3 +294,58 @@ export const noEvalEvaluate = () => {
   // See https://github.com/endojs/endo/blob/master/packages/ses/error-codes/SES_NO_EVAL.md
   throw TypeError('Cannot eval with evalTaming set to "noEval" (SES_NO_EVAL)');
 };
+
+// ////////////////// FERAL_STACK_GETTER FERAL_STACK_SETTER ////////////////////
+
+const er1StackDesc = getOwnPropertyDescriptor(Error('er1'), 'stack');
+const er2StackDesc = getOwnPropertyDescriptor(TypeError('er2'), 'stack');
```
mhofman, line 301: Why use 2 different error types?

packages/ses/src/commons.js
```diff
@@ -294,3 +294,58 @@ export const noEvalEvaluate = () => {
...
+    typeof er1StackDesc.set === 'function' &&
+    er1StackDesc.set === er2StackDesc.set
+  ) {
+    // Otherwise, we have own stack accessor properties that are outside
+    // our expectations, that therefore need to be understood better
+    // before we know how to repair them.
+    feralStackGetter = er1StackDesc.get;
+    feralStackSetter = er1StackDesc.set;
+  } else {
+    throw TypeError('Error own stack accessor functions not as expected');
```
mhofman, line 327: Don't we have a special error format for these kind of errors?

packages/ses/src/make-hardener.js
```diff
@@ -218,21 +220,53 @@ export const makeHardener = () => {
...
+            baseFreezeAndTraverse
+          : obj => {
+              if (isError(obj)) {
+                // Only pay the overhead if it first passes this cheap isError
+                // check. Otherwise, it will be unrepaired, but won't be judged
+                // to be a passable error anyway, so will not be unsafe.
+                const stackDesc = getOwnPropertyDescriptor(obj, 'stack');
+                if (
+                  stackDesc &&
+                  stackDesc.get === FERAL_STACK_GETTER &&
```
mhofman, line 238: Test for either the getter or setter?
```suggestion
                  (stackDesc.get === FERAL_STACK_GETTER || stackDesc.set === FERAL_STACK_SETTER) &&
```

packages/ses/src/make-hardener.js
```diff
@@ -218,21 +220,53 @@ export const makeHardener = () => {
...
+                ) {
+                  // Can only repair if it is configurable. Otherwise, leave
+                  // unrepaired, in which case it will not be judged passable,
+                  // avoiding a safety problem.
+                  defineProperty(obj, 'stack', {
+                    // NOTE: Calls getter during harden, which seems dangerous.
+                    // But we're only calling the problematic getter whose
+                    // hazards we think we understand.
+                    // @ts-expect-error TS should know FERAL_STACK_GETTER
+                    // cannot be `undefined` here.
```
mhofman, line 249: Yeah type narrowing does not cross function boundaries in general.

mhofman (commented):
Besides the couple small nits, I am concerned about modifying harden to mitigate this. I understand the reasoning, but it breaks the layering inside the ses shim (and would prevent my hope to extract harden as its own shim)

packages/ses/src/commons.js
erights: More robust sanity test, to see that the getter is the same across two different ones. Even slightly more robust that testing that two different subclasses of Error agree. 

Resolving.

packages/ses/src/make-hardener.js
erights: > Yeah type narrowing does not cross function boundaries in general.

Any idea why not?

mhofman: Most likely because most people do not define functions conditionally, so it's a case not worth handling for the compiler.

While not the case here because it's a `const` (which TS might not track, I don't know), there are a lot of cases where TypeScript "resets" type inference at function call boundaries because it cannot know whether the function is immediately invoked, or if other calls may have mutated the narrowed variable.

erights: Perhaps the type inference doesn't distinguish between the temporal implications of `let` vs `const`? If `FERAL_STACK_GETTER` were a `let`, then, since a function may be invoked later, it might be invoked well after the `FERAL_STACK_GETTER === undefined` conditioned block. However, `const` guarantees stability, and therefore the absence of these temporal hazards. Does this sound right? (Attn @turadg @michaelfig ) Is this something they should fix? If so, worth filing an issue?

packages/ses/src/make-hardener.js
erights: Done, but with `&&` rather than `||` to be more conservative. But either they are both set or neither is set, so cannot actually make a difference.

Resolving.

mhofman: I believe I didn't express myself correctly. I was trying to handle the case where an error object would show up with either a getter or a setter own prop. Arguably someone would have to go muck with the prop descriptor before hardening the error object, so probably not worth bothering about.

packages/ses/src/make-hardener.js
mhofman: Plenty of reasons this might not maintain the narrowing, but given how edgy this case is, I don't believe it's worth getting TS to improve support for this. There is also a work around if one can avoid a ternary: assign to a `const` after narrowing.

erights: Thanks.

I added a link to this PR conversation, which I think is adequate for this PR. Resolving.

packages/ses/src/make-hardener.js
erights: This test here is only testing whether we're on a platform suffering from v8's misbehavior. 

Ok to reresolve?

packages/ses/src/commons.js
erights: You mean the error codes at https://github.com/endojs/endo/tree/master/packages/ses/error-codes , as used for example at https://github.com/endojs/endo/blob/a77a1cd8c8063e3bf84b0cb348177dbed7bf5ea5/packages/ses/src/lockdown.js#L207 ?

I'll go ahead and do that, but let me know if you meant something else.

erights: Done.

packages/ses/src/make-hardener.js
mhofman: My understanding is that we already know we're on an affected platform, but here we're testing whether the error object we found exhibit the bad behavior. It's possible the error object got modified already (e.g. stack prop removed, or replaced with data). The question is which kind of modifications we want to handle here (in particular whether we want to handle a setter only stack prop if we find one)

packages/ses/src/commons.js
```diff
@@ -294,3 +294,61 @@ export const noEvalEvaluate = () => {
...
+    typeof er1StackDesc.get === 'function' &&
+    er1StackDesc.get === er2StackDesc.get &&
+    typeof er1StackDesc.set === 'function' &&
+    er1StackDesc.set === er2StackDesc.set
+  ) {
+    // Otherwise, we have own stack accessor properties that are outside
+    // our expectations, that therefore need to be understood better
+    // before we know how to repair them.
+    feralStackGetter = er1StackDesc.get;
+    feralStackSetter = er1StackDesc.set;
```
mhofman, line 325: I just realized that we should freeze these feral accessors as they are undeniable intrinsics on v8.

mhofman:
> I want to hear more from @mhofman about how this change would preclude a separable `harden` shim.

This is a repair that harden does for the benefit other layers (pass style). IMO, it's conceptually hard to justify its presence in the ses shim itself. It wouldn't be justifiable at all in a standalone harden shim.

packages/ses/src/commons.js
erights: Oh hell, they certainly are. We should also consider whether we need to add them to the ses-maintained list of "hidden intrinsics". My hypothesis is that they do not, and that freezing is adequate. I'm going to proceed on that assumption, but please let me know quickly if I'm missing something.

erights: Thanks for catching this! I am embarrassed to have missed it.

Done.

erights:
> > I want to hear more from @mhofman about how this change would preclude a separable `harden` shim.
> 
> This is a repair that harden does for the benefit other layers (pass style). IMO, it's conceptually hard to justify its presence in the ses shim itself. It wouldn't be justifiable at all in a standalone harden shim.

It could be exposed as a configuration option in a standalone harden shim. I agree that it is unpleasant, but the underlying problem caused by v8 is the source of the unpleasantness. "Fixing" it in harden is, I think, the fix that's least unpleasant for users, as evidenced by all the test case changes that I no longer needed to do.

erights:
See also https://github.com/endojs/endo/blob/master/packages/ses/error-codes/SES_UNEXPECTED_ERROR_OWN_STACK_ACCESSOR.md

PR #2238, opened 2024-04-23
1488 Passable redux

turadg:
closes: #1488

## Description

Reattempting:
-  https://github.com/endojs/endo/pull/1933

Again:
- Makes `Passable` a generic type instead of `any`.
- Defines overloads for `passStyleOf` to return the actual style.
- Defines the `Key` in terms of `Passable`
- Makes a ton of fixes and suppressions in places that relied on the previous `any`'s

### Security Considerations

n/a

### Scaling Considerations

n/a

Better types are better documentation.

### Testing Considerations

- [ ] https://github.com/Agoric/agoric-sdk/pull/8774

### Upgrade Considerations

These changes may cause type errors downstream. I don't think we should call those breaking change à la semver because they don't affect the runtime.


erights:
(I just reran the windows-latest flake)

erights:
> These changes may cause type errors downstream. I don't think we should call those breaking change à la semver because they don't affect the runtime.

@kriskowal , I defer to you on this. I'll just say I hope we can adopt that precedent as a general rule ;)

packages/marshal/src/rankOrder.js
```diff
@@ -142,16 +142,20 @@ export const makeComparatorKit = (compareRemotables = (_x, _y) => 0) => {
       case 'string': {
         // Within each of these passStyles, the rank ordering agrees with
         // JavaScript's relational operators `<` and `>`.
+        // @ts-expect-error FIXME narrowed
```
erights, line 145: Just curious if you have a plan for how we can eventually get these dispatches to do the narrowing for us. Everything I thought of is worse that just leaving these `@ts-expect-error FIXME`s in place.

packages/marshal/src/rankOrder.js
```diff
@@ -284,9 +299,10 @@ harden(assertRankSorted);
  * function. This is a genuine bug for us NOW because sometimes we sort
  * in reverse order by passing a reversed rank comparison function.
  *
- * @param {Iterable<Passable>} passables
+ * @template {Passable} T
```
erights, line 302: What do you think of
```suggestion
 * @template {Passable} [T=Passable]
```
for all or most of these? It offloads the use site from needing to say `<Passable>` when that's all it would say anyway. I've been doing essentially this elsewhere and been happy with the results.

packages/pass-style/package.json
```diff
@@ -19,6 +19,7 @@
   "module": "./index.js",
   "exports": {
     ".": "./index.js",
+    "./src/types.js": "./src/types.js",
```
erights, line 22: Interesting. Do you expect this precedent to spread?

packages/pass-style/src/make-far.js
```diff
@@ -74,9 +75,10 @@ const assertCanBeRemotable = candidate =>
  * @param {undefined} [props] Currently may only be undefined.
  * That plan is that own-properties are copied to the remotable
  * @param {T} [remotable] The object used as the remotable
- * @returns {T & RemotableBrand<{}, T>} remotable, modified for debuggability
+ * @returns {T & import('./types.js').RemotableObject<I> & RemotableBrand<{}, T>}} remotable, modified for debuggability
```
erights, line 78: Because this type expression is already do large, factoring out the `@import` seems even more compelling here.

packages/pass-style/src/passStyle-helpers.js
```diff
@@ -30,6 +30,7 @@ export const hasOwnPropertyOf = (obj, prop) =>
   apply(objectHasOwnProperty, obj, [prop]);
 harden(hasOwnPropertyOf);
 
+/** @type {(val) => val is {}} */
```
erights, line 33: I've never seen this inline before. Nice.

packages/pass-style/src/passStyleOf.js
```diff
@@ -87,28 +87,26 @@ const makePassStyleOf = passStyleHelpers => {
...
     const passStyleOfRecur = inner => {
       const innerIsObject = isObject(inner);
       if (innerIsObject) {
-        if (passStyleMemo.has(inner)) {
-          // @ts-ignore TypeScript doesn't know that `get` after `has` is safe
-          return passStyleMemo.get(inner);
+        const innerMemo = passStyleMemo.get(inner);
+        if (innerMemo) {
+          return innerMemo;
         }
```
erights, line 110:
```suggestion
        const innerStyle = passStyleMemo.get(inner);
        if (innerStyle) {
          return innerStyle;
        }
```

packages/pass-style/src/typeGuards.js
```diff
@@ -33,13 +33,10 @@ const isRemotable = remotable => passStyleOf(remotable) === 'remotable';
 harden(isRemotable);
 
 /**
- * @callback AssertArray
```
erights, line 36: Yeah, I don't remember why I have these separate named @callable types. Good riddance, thanks.

packages/patterns/src/keys/checkKey.js
```diff
@@ -126,14 +128,15 @@ export const checkKey = (val, check) => {
 harden(checkKey);
 
 /**
- * @param {Passable} val
+ * @param {Key} val
```
erights, line 131: Shouldn't this be

```suggestion
/**
 * @param {any} val
 * @returns {val is Key}
 */
```
?

packages/patterns/test/types.test-d.ts
```diff
@@ -0,0 +1,6 @@
+import { expectType } from 'tsd';
+import { M } from '../src/patterns/patternMatchers.js';
+
+// @ts-expect-error M.any missing parens
+M.arrayOf(M.any);
```
erights, line 5: ;)

erights (approved):
LGTM, thanks!

erights:
Did we stop requiring conventional commits?

packages/marshal/src/encodeToCapData.js
```diff
@@ -117,7 +115,7 @@ export const makeEncodeToCapData = (encodeOptions = {}) => {
    * Readers must not care about this order anyway. We impose this requirement
    * mainly to reduce non-determinism exposed outside a vat.
    *
-   * @param {Passable} passable
+   * @param {any} passable
```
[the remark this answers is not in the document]
turadg: it was `any` before because `Passable` was `any`. Making it `Passable` now causes an error at just about every access in this function. The job of this function is to inspect what the object actually is, narrow the type, and delegate. So the type hint isn't hindering more than helping.

packages/marshal/src/rankOrder.js
turadg: There's something in TS for narrowing by `typeof` but I don't expect they'll our helper function for pass style. The pragmatic thing may be to drop to `any` in places where the code is doing its own type validation.

packages/marshal/src/rankOrder.js
turadg: that makes sense for parameterized types, but here the template slot's type is constrained by the first argument. Giving a default doesn't help.

packages/pass-style/package.json
turadg: I'd rather export the types at the top level but I think that will require either manually maintaining an `index.d.ts` next to the `index.js` or having a single `index.ts` that can export both JS code and TS types. But the latter would entail a build step and thus require some new workflows and buy-in from the repo maintainers

packages/patterns/src/keys/checkKey.js
turadg: indeed

packages/pass-style/package.json
mhofman: I'm confused, why can't `index.js` do `export * from './src/types.js'` ?

turadg: yeah, I realized that soon after my comment. it does require an eslint suppression but I'll do that

packages/marshal/src/encodeToCapData.js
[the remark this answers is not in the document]
erights: Thx. Resolving.

packages/marshal/src/rankOrder.js
erights: Thx. Resolving.

erights:
> Did we stop requiring conventional commits?

My mistake. All your commits are fine. I was misled by the title, as my habit is to have the title also follow this convention (and usually be identical to the first commit).

packages/exo/index.js
```diff
@@ -1,3 +1,6 @@
 export * from './src/exo-makers.js';
 
+// eslint-disable-next-line import/export -- ESLint not aware of type exports in types.d.ts
```
kriskowal, line 3: That’s unfortunate and might explain some of my experience.

packages/exo/src/types.js
```diff
@@ -0,0 +1,2 @@
+/** @file Empty twin for .d.ts */
```
kriskowal, line 1: 👍

packages/far/test/test-marshal-far-obj.js
```diff
@@ -5,6 +5,10 @@ import { q } from '@endo/errors';
 
 import { Far, passStyleOf, getInterfaceOf } from '../src/index.js';
 
+/**
+ * @import {PassStyled} from '@endo/pass-style';
```
kriskowal, line 9: Would I be right to assume that someday `eslint` will catch up with this notation and demand spaces inside the curlies?

packages/marshal/src/deeplyFulfilled.js
```diff
@@ -11,6 +11,7 @@ import { X, q } from '@endo/errors';
 const { ownKeys } = Reflect;
 const { fromEntries } = Object;
 
+// TODO return a type conditioned on the parameter as deeplyFullfilledObject from agoric-sdk does
```
kriskowal, line 14: Optional wordsmithery.
```suggestion
// TODO return a type contingent on the parameter as deeplyFullfilledObject from agoric-sdk does
```

packages/marshal/src/encodePassable.js
```diff
@@ -728,6 +728,7 @@ const makeInnerDecode = (decodeStringSuffix, decodeArray, options) => {
       }
     }
   };
+  // @ts-expect-error XXX Type 'unknown' is not assignable to type 'Passable<PassableCap, Error>'.
```
kriskowal, line 731:
```suggestion
  // @ts-expect-error Type 'unknown' is not assignable to type 'Passable<PassableCap, Error>'.
```
I think this is fine without elaboration.

packages/marshal/src/encodeToCapData.js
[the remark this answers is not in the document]
kriskowal: Would `unknown` work better than `any` here, or is it better to blanket-ignore the type-unsafety internally?

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -386,6 +386,7 @@ export const makeDecodeFromSmallcaps = (decodeOptions = {}) => {
               encoding,
               decodeFromSmallcaps,
             );
+            // @ts-ignore FIXME tsc thinks result is 'copyRecord' style
```
kriskowal, line 389: What’s the fix? Is this as good as it gets?

packages/marshal/src/marshal.js
```diff
@@ -364,7 +370,9 @@ export const makeMarshal = (
     };
 
     const reviveFromSmallcaps = makeDecodeFromSmallcaps({
+      // @ts-expect-error FIXME
```
kriskowal, line 373: 👆

packages/marshal/src/marshal.js
```diff
@@ -364,7 +370,9 @@ export const makeMarshal = (
...
       decodeRemotableFromSmallcaps,
+      // @ts-expect-error FIXME
```
kriskowal, line 375: 👆

packages/marshal/src/rankOrder.js
```diff
@@ -142,16 +142,20 @@ export const makeComparatorKit = (compareRemotables = (_x, _y) => 0) => {
       case 'string': {
         // Within each of these passStyles, the rank ordering agrees with
         // JavaScript's relational operators `<` and `>`.
+        // @ts-expect-error FIXME narrowed
         if (left < right) {
           return -1;
         } else {
+          // @ts-expect-error FIXME narrowed
```
kriskowal, line 149: 👆 Is the intention to address all FIXMEs before landing? These could be links to an issue, otherwise.

packages/marshal/src/rankOrder.js
```diff
@@ -142,16 +142,20 @@ export const makeComparatorKit = (compareRemotables = (_x, _y) => 0) => {
...
           assert(left > right);
           return 1;
         }
       }
       case 'symbol': {
         return comparator(
+          // @ts-expect-error FIXME narrowed
```
kriskowal, line 156: I would not object to ditto if these stick around.
```suggestion
          // @ts-expect-error ditto, type narrowed
```

packages/pass-style/src/passStyleOf.js
```diff
@@ -173,10 +168,12 @@ const makePassStyleOf = passStyleHelpers => {
           }
           for (const helper of passStyleHelpers) {
             if (helper.canBeValid(inner)) {
+              // @ts-expect-error XXX
```
kriskowal, line 171: 👆

packages/patterns/src/keys/compareKeys.js
```diff
@@ -133,20 +133,25 @@ export const compareKeys = (left, right) => {
       // rank order.
       // Because the invariants above apply to the elements of the array,
       // they apply to the array as a whole.
+      // @ts-expect-error FIXME narrowed
```
kriskowal, line 136: 👆 I’m fine with a link and dittos if these are intended to commit. (but don’t like `FIXME` because I like `FIXME` and `XXX` to be used as markers for work that needs to be ironed out before a merge.

kriskowal (changes requested):
I’m substantially excited about this change. I would like to break out a PR for the change to the release and build process so that it can suffer a revert separately if necessary. I’d also like to replace temporary markers `XXX` and `FIXME` with something but it doesn’t have to be tedious. It could be a single explanation or link to a tracking issue with a bunch of dittos or daggers for the repeated cases.

⚠️I will not die on this hill and I will approve this change regardless⚠️: My intuition (born from typed languages and sympathy for dependant projects that have to adapt when they upgrade this dependency) is that we should consider this a breaking change both in spirit and semver, and rush the necessary adaptations in Agoric SDK and dapps to minimize the pain. Since we are treading heavily over breaking changes with the “sync endo versions” workflow and `resolutions` farther out, there’s not much reason to spare the major version bump.

kriskowal:
Also, please add `NEWS.md` with a description of what you’ve done for https://github.com/Agoric/agoric-sdk/pull/8774 so we can refer developers to this when they upgrade their *apps.

packages/base64/package.json
```diff
@@ -32,7 +32,7 @@
   },
   "scripts": {
     "build": "exit 0",
-    "build:types": "tsc --build tsconfig.build.json",
+    "prepack": "tsc --build tsconfig.build.json",
```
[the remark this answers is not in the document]
turadg: 👍 https://github.com/endojs/endo/pull/2258

packages/captp/src/captp.js
```diff
@@ -323,8 +323,9 @@ export const makeCapTP = (
 
   const IS_REMOTE_PUMPKIN = harden({});
   /**
-   * @type {import('@endo/marshal').ConvertSlotToVal<import('./types.js').CapTPSlot>}
+   * @type {import('@endo/marshal').ConvertValToSlot<import('./types.js').CapTPSlot>}
    */
+  // @ts-expect-error intentional hack
```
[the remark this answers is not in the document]
turadg: bad comment. the type was wrong so I just removed it

packages/captp/src/captp.js
```diff
@@ -413,6 +414,7 @@ export const makeCapTP = (
           epoch,
           questionID,
           target,
+          // @ts-expect-error XXX
```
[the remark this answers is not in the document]
turadg: solved

packages/exo/index.js
turadg: yeah. I think we should consider disabling the rule. I think it has more false positives and fewer true positives than TS

packages/marshal/src/encodeToSmallcaps.js
turadg: clarified

packages/marshal/src/rankOrder.js
turadg: not before landing. maybe never. I've taken out the FIXME

turadg:
I had to rebase to resolve merge conflicts created by https://github.com/endojs/endo/pull/2256

I squashed the fixups and rolled in other changes requested by review

kriskowal (approved):
You may want to look over the remaining XXX and TODO cases.

PR #2222, opened 2024-04-17
upgrade to Yarn 4

turadg:
closes: #2245 
refs: https://github.com/Agoric/agoric-sdk/issues/451
refs: https://github.com/Agoric/ui-kit/pull/105

## Description

Move from Yarn 1 to Yarn 4. Some advantages,

- actively maintained
- [workspace protocol](https://yarnpkg.com/protocol/workspace) (so we don't have to maintain version numbers in all deps)
- [patch protocol](https://yarnpkg.com/protocol/patch) (so we don't need patch-package)
- [constraints](https://yarnpkg.com/features/constraints) (e.g. to enforce layering)
- path to adopt pnpm-style linker (without changing the UI) (see https://github.com/endojs/endo/pull/1722 )

However this defers workspace protocol until the publishing workflow can support it. 

### Security Considerations

This does a bulk update of `yarn.lock`. It was automated by Yarn 4.

### Scaling Considerations

n/a

### Documentation Considerations

I reviewed `yarn` commands in *.md and I think they're all accurate.

### Testing Considerations

This could interact with the publishing pipeline. @kriskowal may want to push a draft before we merge. If problems are found, depending on the severity, we could follow up in a separate PR to land this sooner reduce merge conflicts.

This was failing on the Windows tests, something about corepack not taking effect. I don't know whether Windows is officially supported by Endo. We've since disabled them. https://github.com/endojs/endo/issues/2243 is the issue restore.

### Compatibility Considerations

Some CLI commands are slightly different. We are adopting it across the org so we have to adjust sometime.

### Upgrade Considerations

n/a

PR #2240, opened 2024-04-24
test: Drop Windows CI

kriskowal:
We’ve seen a lot of EPIPE on write failures that are more common on Windows than Linux/Mac, though they do occur. For that reason along, it would be good to restore Windows CI when we track that issue down, but for now, the flakes are too hard to bear and also block us from upgrading yarn #2222


rekmarks (approved):
I support this for now, although we should create an issue to track re-enabling it.

PR #2246, opened 2024-04-27
DO NOT MERGE: automated updates to yarn.lock

erights:
DO NOT MERGE

This is what I got on a clean local master after deleting all node_modules and doing `yarn`.

I have not looked at the contents and do not currently have an opinion of the actual changes. But it causes massive CI failures, so obviously does not work.

closes: #XXXX
refs: #XXXX

### Upgrade Considerations


erights:
Nevermind. I have not updated to yarn 4. Closing.

But it would be good if we could somehow give a better diagnostic, so people know they need to upgrade to yarn 4.

PR #2247, opened 2024-04-28
refactor(exo): shorten exo-call fastpath

erights:
closes: #XXXX
refs: #XXXX

## Description

Should be a pure refactor without any[1] observable differences. Shorten the exo-call fastpath for the normal thisful case 
- primarily to make stepping into an exo method more pleasant with fewer clicks and visual context switches
- probably also has a performance advantage for the normal thisful case.

By "thisful", I mean the exo objects defined with the normal exo-making, exo-class-making, and exo-class-kit-making APIs, whether directly or via zones. The non-thisful case is only the deprecated virtual and durable kinds, which are not yet gone and so must still be supported. But have only the non-thisful case pay for the arg shifting, rather than making all calls pay for that.

The previous debug experience involved two layers of wrapping function to get from the external call to a method of an exo into the exo's own behavior method. This PR reduces that to one level of wrapper. The effectively step into from the call into the actual behavior method before and after this PR
- "step into" (new page) "step over", "step into" (new page) "step over" "step over" "step into" (new page)
- "step into" (new page) "step over" "step over" "step into" (new page)

There is also one less exo infrastructure frame on the call stack. In the debugger, the remaining stack frame is labeled

image

vs

image

Also less noise in verbose error stacks.

### Security Considerations

none. The safety checks should be exactly the same.
### Scaling Considerations

Probably has a performance advantage for the normal thisful case. But we won't know until we measure. We're putting this into review before measuring though, since the main motivation is the better debugging experience.

none
### Testing Considerations

[1] The refactoring did result in an error message changing, requiring a corresponding change in a golden error message test.

The tests in endo only test the normal thisful case. We depend on integration testing with agoric-sdk to test the non-thisful case.

none, given that the integration test with agoric-sdk confirms we did not break the non-thisful case.

### Upgrade Considerations

none.

### DIFF
--- a/packages/exo/src/exo-tools.js
+++ b/packages/exo/src/exo-tools.js
@@ -406,6 +406,7 @@ export const defendPrototype = (
     };
     const behaviorMethod = thisfulMethods ? originalMethod : shiftedMethod;
     // TODO some tool does not yet understand the `?.[` syntax
+    // See https://github.com/endojs/endo/pull/2247#discussion_r1583724424
     let methodGuard = methodGuards && methodGuards[prop];
     if (!methodGuard) {
       switch (defaultGuards) {
--- a/packages/exo/src/exo-tools.js
+++ b/packages/exo/src/exo-tools.js
@@ -401,7 +401,7 @@ export const defendPrototype = (
     const originalMethod = behaviorMethods[prop];
     const { shiftedMethod } = {
       shiftedMethod(...args) {
-        return originalMethod([this, ...args]);
+        return originalMethod(this, ...args);
       },
     };
     const behaviorMethod = thisfulMethods ? originalMethod : shiftedMethod;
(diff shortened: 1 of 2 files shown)


erights:
Integration test via pount-endo-branch at https://github.com/Agoric/agoric-sdk/pull/9298

patch for debugging use until then https://github.com/Agoric/agoric-sdk/pull/9297

erights:
Both agoric-sdk integration test PRs are green.

packages/exo/src/exo-tools.js
```diff
@@ -402,14 +398,43 @@ export const defendPrototype = (
...
 
   for (const prop of methodNames) {
+    const originalMethod = behaviorMethods[prop];
+    const { shiftedMethod } = {
+      shiftedMethod(...args) {
+        return originalMethod(this, ...args);
+      },
+    };
+    const behaviorMethod = thisfulMethods ? originalMethod : shiftedMethod;
+    // TODO some tool does not yet understand the `?.[` syntax
```
kriskowal, line 408: Specifically `node -r esm` which `agoric deploy` scripts might still use, but it’s not clear to me that any old deploy scripts depend on `@endo/exo`.

erights: God to know. Added note pointing here. Resolving.

PR #2260, opened 2024-04-30
fix(ses): fix broken types for CJS consumers

boneskull:
Due to changes in TS v5.5, CJS consumers cannot import types from a `type: module` package unless the declaration file has the `.d.cts` extension. This may or may not work the other way.

See [the v5.5. beta announcement](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5-beta/#respecting-file-extensions-and-packagejson-in-other-module-modes) for further explanation.


kriskowal (changes requested):
An entry in `NEWS.md` would be good for this.

kriskowal (approved):

PR #2261, opened 2024-05-01
feat(cli): Add chat to demo web UI

kriskowal:
This adds a chat and message log auto-scroll support to the demo Chat Familier web UI.

image

Click the cat.

image

Click chat.

image

`AGENT` isn’t a handle. We can’t filter on type yet.

image

But `SELF` is a handle.

image

Press Send.

image

### DIFF
--- a/packages/cli/demo/cat.js
+++ b/packages/cli/demo/cat.js
@@ -21,11 +21,173 @@
 // In the Familiar Chat window, resolve the request with the pet name "ft" and
 // the request will exit with the number 42.
 
-/* global window document */
+/* global window document requestAnimationFrame */
 /* eslint-disable no-continue */
 
 import { E } from '@endo/far';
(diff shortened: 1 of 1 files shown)


packages/cli/demo/cat.js
```diff
@@ -173,26 +355,22 @@ const inboxComponent = async ($parent, $end, powers) => {
...
-        $as.type = 'text';
-        $message.appendChild($as);
-
-        $message.appendChild(document.createTextNode(' '));
-
-        const $adopt = document.createElement('button');
-        $adopt.innerText = 'Adopt';
-        $message.appendChild($adopt);
-        $adopt.onclick = () => {
+        const handleAdopt = () => {
```
dckc, line 373: the linter is very angry about `'$as' was used before it was defined `

dckc (approved):
works as advertised

The code is a little over my head in places, but it's fine for a demo.

PR #2263, opened 2024-05-01
feat(trampoline): add a trampoline module (@endo/trampoline)

boneskull:
we will probably need something like this.  based on work by @naugtur 

**UPDATE**: I plan to use this in #2310 and someone should eventually refactor `ses` to use it.


packages/trampoline/README.md
```diff
@@ -0,0 +1,9 @@
+# supertramp
+
+> Sync and async trampolines, for your pleasure
+
+## Wat
+
+From [Wikipedia](https://en.wikipedia.org/wiki/Trampoline_(computing)):
+
+> A trampoline is a loop that iteratively invokes [thunk](https://en.wikipedia.org/wiki/Thunk_(functional_programming))-returning functions ([continuation-passing style](https://en.wikipedia.org/wiki/Continuation-passing_style)).
```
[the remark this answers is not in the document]
boneskull: To be clear, you're _not_ saying, "don't call it a trampoline", yes?  There is [precedent](https://www.npmjs.com/search?q=trampoline) in the JS ecosystem for this meaning. 

I would prefer we name it something silly, of course.

Open to suggestions.  I like `rainbow-trampoline`, because it supports functions of any color 🌈.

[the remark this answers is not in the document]
mhofman: Function coloring is also a loaded term. Sync/async coloring is currently the main kind in JS, but coloring has been used in the context of a couple other proposals lately.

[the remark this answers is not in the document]
boneskull: I can, without rancor, use the more-widely-understood meaning of words within their context (in this case, the community of JavaScript developers) while disregarding those less-widely-understood meanings. Naming is hard and we do the best we can 😄 

But if you (or anyone else) has a suggestion for naming this package, please offer it!

packages/trampoline/package.json
```diff
@@ -0,0 +1,82 @@
...
+    "LICENSE*",
+    "SECURITY*",
+    "src",
+    "*.js",
+    "*.ts"
+  ],
+  "private": true,
+  "eslintConfig": {
+    "extends": [
+      "plugin:@endo/internal"
```
boneskull, line 70: I copy/pasted this. 

I don't understand why this is here and why we are pulling in all of those eslint dev deps.

boneskull: Also, I am not sure why we're pulling in all of these dev deps in _any_ workspace.  

Why not let them all live in the monorepo root?

packages/trampoline/README.md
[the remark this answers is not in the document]
naugtur: `optional-a-sync`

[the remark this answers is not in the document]
naugtur: @mhofman @boneskull need more razors for this yak shaving?

I tend to use GPT for seeding name ideas. gpt-4o is really bad at it tho, had to go back to gpt-4
Here are some not awful seeds which we could use as inspiration for coming up with something reasonable.
```
1. AsyncBridge
2. SyncAsyncHarmonizer
3. JumpSync
4. YieldBounce
5. GenTrampolinator
6. AsyncOrNot
8. CodeChameleon
12. SyncAsyncLeap
16. AsyncFlip
17. YieldFlex
18. BounceBetween
19. SyncAsyncFusion
```

I still stand by `optional-a-sync` but these could also work from above:
`async-or-not` `sync-async-chameleon` `sync-async-bridge`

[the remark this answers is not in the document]
kriskowal: I would like to name this `trampoline`.

packages/trampoline/src/trampoline.js
```diff
@@ -0,0 +1,49 @@
...
+ * @template {readonly any[]} [TArgs=unknown[]] Parameters for `generatorFn`
+ * @template [TResult=unknown] Type of the return value of the `generatorFn`
+ * @template {Generator<any, any, TResult>} [TGenerator=Generator] Type of the generator function
+ * @param {AsyncTrampolineGeneratorFn<TGenerator, TArgs, TResult>} generatorFn Generator-returning function accepting a thunk and optionally an initial value
+ * @param {TArgs} args Initial args
+ * @returns {Promise<TResult>}
+ */
+export async function trampoline(generatorFn, ...args) {
+  const iterator = generatorFn(...args);
+  let result = iterator.next();
```
naugtur, line 37: If we want to use this in SES we'll need to uncuryThis on the next and error methods and keep them local. (can steal the 3 lines for that from commons in SES)

packages/trampoline/README.md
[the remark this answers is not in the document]
boneskull: I'd name it "The Right Honourable Boris von Meija" if it were totally up to me, but I can live with `@endo/trampoline`.

packages/trampoline/src/trampoline.js
boneskull: @naugtur I don't understand if this is actionable by me in this PR?

packages/trampoline/README.md
```diff
@@ -0,0 +1,91 @@
...
+const syncResult = syncTrampoline(loadRecursive, readSync, './some.js');
+
+asyncResult === syncResult; // true
+```
+
+In the above example, **@endo/trampoline** allows us to re-use the recursive operations in `loadRecursive()` for _both_ sync and async execution. An implementation _without_ **@endo/trampoline** would need to duplicate the operations into two (2) discrete recursive functions—a synchronous-colored function and an asynchronous-colored function. Over time, this situation commonly leads to diverging implementations. If that _doesn't_ sound like a big deal for _whatever you're trying to do here_, then you probably don't need **@endo/trampoline**.
+
+## What is this?
+
+The pattern exposed by this library—known as [trampolining][]—helps manage control flow in a way that avoids deep recursion and potential stack overflows. It effectively "converts" recursive calls into a loop. This is especially helpful in a language like JavaScript [because reasons][proper-tail-calls].
```
naugtur, line 67: This line remains slightly confusing - first time reading it I didn't grok that it's describing what the library is NOT.
```suggestion
This library is not just an implementation of a trampoline that helps manage control flow in a way that avoids deep recursion and potential stack overflows (by turning recursive calls into a loop). 
```

packages/trampoline/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+**@endo/trampoline** provides the trampolining pattern, but in such a way that a consumer can execute _either_ synchronous _or_ asynchronous operations _paired with operations common to both_.
+
+In other words, **@endo/trampoline** can help _reduce code duplication_ when recursive operations must be executed _in both sync and async_ contexts.
```
naugtur, line 71:
```suggestion
In other words, **@endo/trampoline** can help _reduce code duplication_ when complex logic, potentially even with recursive operations must be executed _in both sync and async_ contexts with patterns like dependency injection used to provide matching sync or async utilities to the main logic.
```

packages/trampoline/src/types.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * A function type that represents a generator function for trampolining.
+ *
+ * @template TInitial - The type of the initial value.
+ * @template TArg - The type of the argument passed to the thunk function.
+ * Defaults to `TInitial`.
+ * @template TResult - The type of the result produced by the thunk function.
+ * Defaults to `TArg`.
+ * @param thunk - The thunk function to be used in the generator.
+ * @param initial - The initial value to start the generator.
```
naugtur, line 11: The implementation is not opinionated about the arguments of the generator function nor the way it gets the sync or async utilities.

It can be by dependency injection of sorts (including passing a thunk) but it could also be by choosing the right utility function from imported package based on the input arguments via any logic. 

I removed the opinionation from the trampoline params but wasn't exactly sure what the two params are up to here.

boneskull (commented):
> This line remains slightly confusing - first time reading it I didn't grok that it's describing what the library is NOT.

It's...not.  see updated text

boneskull (commented):
> ```suggestion
In other words, **@endo/trampoline** can help _reduce code duplication_ when complex logic, potentially even with recursive operations must be executed _in both sync and async_ contexts with patterns like dependency injection used to provide matching sync or async utilities to the main logic.
``` 

i've updated this too

**Review by naugtur — APPROVED** [review]
The only thing I dislike about this now is the `trampoline` export - `syncTrampoline` and `asyncTrampoline` are a perfect pair. There's not much special about the async one and  using only one of them doesn't make much sense.

File: packages/trampoline/package.json (hunk 2 above)
Comment (kriskowal) [reply packages/trampoline/package.json:70]: Let’s follow-up about this separately and solve it homogeneously if we can. The config for `ses` and below is a little different than above `ses`, where we can count on Hardened JavaScript for some invariants. But, otherwise, if we can hoist more, I’m in favor.

If we are going to use this package in SES, it requires the more-strict lint (no-polymorphic-call), and I assume you’re shooting for that. (to follow-up with a SES refactor). There’ll be a bit of duplication with uncurryThis and getting commons.

@naugtur did you review this with a close eye on whether this would interfere with Hermes?

File: packages/trampoline/src/trampoline.js [hunk 7]
```diff
@@ -0,0 +1,64 @@
+/* eslint-disable @jessie.js/safe-await-separator */
+/**
+ * @import {TrampolineResult, SyncTrampolineResult} from './types.js'
+ */
+const { getPrototypeOf } = Object;
+const { bind } = Function.prototype;
+const uncurryThis = bind.bind(bind.call); // eslint-disable-line @endo/no-polymorphic-call
```
kriskowal, line 7: Mind that the rule might not be enabled for this package (and probably should be enabled)

kriskowal (approved):
I like it. I agree with @naugtur that it should export `syncTrampoline` and `asyncTrampoline` without leaving the consumer to guess which of these `trampoline` aliases.

packages/trampoline/src/trampoline.js
boneskull: the eslint rule that warns if eslint directives are unused should be enabled 😄

boneskull:
> The only thing I dislike about this now is the `trampoline` export - `syncTrampoline` and `asyncTrampoline` are a perfect pair. There's not much special about the async one and using only one of them doesn't make much sense.

`trampoline` has been removed; now only `asyncTrampoline` and `syncTrampoline` exist

packages/trampoline/README.md
boneskull: for posterity: even if it is not an explicit goal, we get mitigation of the lack of proper TCO for free. right?

PR #2264, opened 2024-05-01
feat(cli): Add eval form to demo weblet

kriskowal:
This change adds an eval form to the demo weblet.

image

Click the cat then Eval or type the accelerator key “.”. Fill in the source and add an endowment.

image

Press thee Evaluate button and observe there’s a new `cat-id` in the pet store.

image

### DIFF
--- a/packages/cli/demo/cat.js
+++ b/packages/cli/demo/cat.js
@@ -25,11 +25,18 @@
 /* eslint-disable no-continue */
 
 import { E } from '@endo/far';
+import { passStyleOf } from '@endo/pass-style';
 import { makeRefIterator } from '@endo/daemon/ref-reader.js';
 import { parseMessage } from '../src/message-parse.js';
 
 const template = `
 <style>
+
+  :root {
+    --tint-color:   hsl(210, 71%, 72%);
+    --border-color: hsl(210, 60%, 52%);
+    --mask-color:   hsla(210, 38%, 29%, 0.5);
+  }
 
   * {
     box-sizing: border-box;
@@ -77,8 +84,8 @@ const template = `
     padding: 20px;
     overflow-y: auto;
     overflow-x: hidden;
-    background-color: hsl(210, 71%, 72%);
-    border-left: 1px solid hsl(210, 60%, 52%);
+    background-color: var(--tint-color);
+    border-left: 1px solid var(--border-color);
   }
 
   #pet-list {
@@ -155,7 +162,7 @@ const template = `
     right: 0;
     align-items: center;
     justify-content: center;
-    background-color: hsla(210, 38%, 29%, 0.5)
+    background-color: var(--mask-color);
   }
 
   .frame[data-show=true] {
@@ -166,7 +173,7 @@ const template = `
     background-color: white;
     margin: 20px;
     padding: 20px;
-    border: 1px solid hsl(210, 60%, 52%);
+    border: 1px solid var(--border-color);
     border-radius: 10px;
     overflow: auto;
     max-height: calc(100vh - 40px);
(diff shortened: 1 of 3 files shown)


packages/cli/demo/cat.js
```diff
@@ -175,20 +186,47 @@ const template = `
...
     <label for="chat-to">To:&nbsp;<select id="chat-to"></select></label>
     <p><input type="text" id="chat-message">
-    <p id="endowments">
     <p id="chat-error" class="error">
     <p><button id="chat-discard-button">Discard</button>
     <button id="chat-send-button">Send</button>
   </div>
 </div>
+
+<div id="eval-frame" class="frame">
```
rekmarks, line 202: This element is insufficiently reactive; when I open the console in Chrome on my system, the eval window adjusts left and its form elements become invisible.

packages/cli/demo/cat.js
```diff
@@ -585,6 +644,138 @@ const chatComponent = ($parent, powers, { dismissChat }) => {
...
+    const { key, repeat, metaKey } = event;
+    if (repeat || metaKey) return;
+    if (key === 'Enter') {
+      event.stopPropagation();
+      handleEval();
+    }
+    if (key === 'Escape') {
+      clearEval();
+      event.stopPropagation();
+    }
```
rekmarks, line 750: This does not appear to work on my system.

packages/cli/demo/cat.js
```diff
@@ -585,6 +644,138 @@ const chatComponent = ($parent, powers, { dismissChat }) => {
...
+      });
+  };
+
+  $frame.addEventListener('keyup', event => {
+    const { key, repeat, metaKey } = event;
+    if (repeat || metaKey) return;
+    if (key === 'Enter') {
+      event.stopPropagation();
+      handleEval();
+    }
```
rekmarks, line 746: This triggers `handleEval` too eagerly, namely whenever enter is pressed while the eval window is open. This causes errors when, for example, adding newlines in the source input or attempting to add an endowment using only the keyboard. IMO, form submission should only trigger on enter if the submission button is selected.

It's not part of this PR, but we should perhaps review `handleAdopt` for the same issue.

packages/cli/demo/cat.js
```diff
@@ -585,6 +644,138 @@ const chatComponent = ($parent, powers, { dismissChat }) => {
...
+      const $petName = $endowment.querySelector('.pet-name');
+      return {
+        petName: $petName.value,
+        codeName: $codeName.value,
+      };
+    });
+    const codeNames = names.map(({ codeName }) => codeName);
+    const petNames = names.map(({ petName }) => petName);
+    const resultName = $resultName.value.trim() || undefined;
+    E(powers)
```
rekmarks, line 729:
```suggestion
    const resultName = $resultName.value.trim() || undefined;
    console.log('eval', {
      source,
      workerName,
      codeNames,
      petNames,
      resultName,
    });

    E(powers)
```

rekmarks (changes requested):
Some inline change requests. In addition, should we add an element for displaying eval command results? If not, we could at least add a log statement.

packages/cli/demo/cat.js
rekmarks: Maybe we should log the result as well?

kriskowal: This entrains into scope the generation of a web representation of passable values, which we would need for other reasons, like revealing a representation of a pet-named value. It might be okay to just transition to a subsequent modal view.

kriskowal: I added a modal “value” viewer.

Screenshot 2024-05-03 at 5 10 35 PM

Screenshot 2024-05-03 at 5 10 40 PM

packages/cli/demo/cat.js
kriskowal: Tough call whether to oblige you to use shift+enter or navigate over to the submit button, but I’ve done the latter.

packages/cli/demo/cat.js
kriskowal: Works on my machine.

packages/cli/demo/cat.js
kriskowal: ☢️ irradiated

packages/cli/demo/cat.js
kriskowal: Now with responsive square brackets and curly braces made possibly by crimes.

image

dckc: Why not a plain text rendering with q?

kriskowal: RIP. I can certainly make it look like `JSON.stringify` (`q`), and copy-paste nicely, but would want to do better for other passables like bigint, promise, remotable, and tagged. Using a DOM representation allows for responsive layout. Named remotables might be labeled by their pet name. It’s conceivable that CSS could be tricked into pivoting long lines from columns to rows with indentation if they wrap, or just toggle on clicking the brace/bracket. Building out of DOM elements leaves the door open for more interactivity in future changes.

kriskowal: Reduced fanciness.

Screenshot 2024-05-06 at 7 31 10 PM

Screenshot 2024-05-06 at 7 31 24 PM

dckc: > Named remotables might be labeled by their pet name.

That's a _very_ interesting possibility.
IOU an issue.

packages/cli/demo/cat.js
rekmarks: Ah, I know what my problem is. I want this to trigger regardless of what I've selected on the page, i.e. I want a listener on `window` for the `Escape` accelerator only. We can consider this complication optional for purposes of this PR, however.

packages/cli/demo/cat.js
```diff
@@ -874,36 +755,59 @@ const render = value => {
...
-      $value.className = 'array';
+      const $value = document.createElement('span');
+      $value.appendChild(document.createTextNode('['));
+      const $entries = document.createElement('span');
+      $entries.className = 'entries';
+      $value.appendChild($entries);
+      let $entry;
       for (const child of value) {
+        $entry = document.createElement('span');
+        // $entry.className = 'entry';
```
rekmarks, line 766: There appears to be some leftover commented-out lines in the `copyArray` and `copyRecord` case blocks.

rekmarks (approved):
LGTM!

PR #2269, opened 2024-05-06
fix: Corepack with TypeDoc generation job

kriskowal:
The typedoc CI job failed because of a package manager mismatch that can be addressed by enabling corepack.

https://github.com/endojs/endo/actions/runs/8975075410/job/24648857762


turadg (approved):
Confirmed fixed

https://github.com/endojs/endo/actions/runs/8975455700

PR #2271, opened 2024-05-06
fix: Make pull request template guidance more evident if it persists

kriskowal:
The pull request template currently leaves invisible comments which have occasionally survived into commit messages. This changes those comments to block quotes to make their existence more evident to reviewers and to encourage change proposers to delete them as they fill the form.


erights (approved):
LGTM, thanks

PR #2281, opened 2024-05-09
feat(daemon,cli): Make and store at path

kriskowal:
Enables caplets to be made and values to be stored at deeper pet name paths.

```
endo make x.js -n dir.x
endo store --json 42 -n dir.forty-two
endo store --text-stdin --name dir.greeting < <(echo hi)
```

### DIFF
--- a/packages/cli/src/endo.js
+++ b/packages/cli/src/endo.js
@@ -12,6 +12,7 @@ import url from 'url';
 
 import { Command } from 'commander';
 import { prompt } from './prompt.js';
+import { parsePetNamePath } from './pet-name.js';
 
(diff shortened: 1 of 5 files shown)


rekmarks (approved):
Self-dealing approval.

packages/cli/src/endo.js
```diff
@@ -370,7 +371,7 @@ export const main = async rawArgs => {
         storeJson,
         storeJsonStdin,
         storeBigInt,
-        name,
+        name: parsePetNamePath(name),
```
kriskowal, line 374: Really ought to be done in `store.js` to keep `endo.js` light and loosely coupled.

PR #2282, opened 2024-05-10
feat(graph): also generate the transitive reduction

erights:
Closes: #XXXX
Refs: https://github.com/Agoric/agoric-sdk/pull/9359

## Description

Use the "tred" executable that seems to already be bundled with graphviz to

```
# Also generates visualizations of the transitive reduction (tred) of
# that graph, which is the minimal graph with the same *transitive*
# dependencies. Much more legible by itelf. Seeing the two side by side
# often helps to understand the full picture.
```

At the time of writing, the current graph of endo dependencies is

packages-graph

and the transitive reduction of those is

packages-graph-tred

Notice how the columns *mostly* align. Unfortunately, the columns don't fully align, and the vertical order of elements within each column is not the same. For viewing them side by side for better understanding, it would be better if both of these annoyances were fixed. But I have no idea how to do that, so this PR does not try.

### Security Considerations

none
### Scaling Considerations

none

potentially helps someone who wants to understand how our system is internally layered, which is the point. Otherwise, none
### Testing Considerations

none

none
### Upgrade Considerations

none

### DIFF
--- a/scripts/graph.sh
+++ b/scripts/graph.sh
@@ -1,5 +1,9 @@
 #!/bin/bash
 # Generates visualizations of the internal package dependency graph.
+# Also generates vistualizations of the transitive reduction (tred) of
+# that graph, which is the minimal graph with the same *transitive*
+# dependcies. Much more legible by itelf. Seeing the two side by side
+# often helps to understand the full picture.
 set -ueo pipefail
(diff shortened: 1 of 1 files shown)


turadg (approved):
helpful!

PR #2288, opened 2024-05-15
feat(cli,daemon): Add `mkdir` command

rekmarks:
Adds `mkdir` command to the CLI. Updates the daemon's directory abstraction to accept pet name paths at `makeDirectory()`.

```
> endo mkdir dir
> endo make counter.js -n dir.counter
> endo eval 'E(counter).incr()' counter:dir.counter
1
```

### DIFF
--- a/packages/daemon/src/types.d.ts
+++ b/packages/daemon/src/types.d.ts
@@ -453,7 +453,7 @@ export interface NameHub {
 }
 
 export interface EndoDirectory extends NameHub {
-  makeDirectory(petName: string): Promise<EndoDirectory>;
+  makeDirectory(...petNamePath: string[]): Promise<EndoDirectory>;
 }
 
 export type MakeDirectoryNode = (petStore: PetStore) => EndoDirectory;
--- a/packages/cli/src/commands/mkdir.js
+++ b/packages/cli/src/commands/mkdir.js
@@ -6,5 +6,5 @@ import { parsePetNamePath } from '../pet-name.js';
 
 export const mkdir = async ({ agentNames, directoryPath }) =>
   withEndoAgent(agentNames, { os, process }, async ({ agent }) => {
-    await E(agent).makeDirectory(parsePetNamePath(directoryPath));
+    await E(agent).makeDirectory(...parsePetNamePath(directoryPath));
   });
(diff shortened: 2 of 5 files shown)


packages/daemon/src/types.d.ts
```diff
@@ -453,7 +453,7 @@ export interface NameHub {
 }
 
 export interface EndoDirectory extends NameHub {
-  makeDirectory(petName: string): Promise<EndoDirectory>;
+  makeDirectory(...petNamePath: string[]): Promise<EndoDirectory>;
```
kriskowal, line 456: This is fine, but I have misgivings about throwing away the possibility of cleanly taking an options bag later that would be addressed by taking `string | string[]` here instead. I wouldn’t mind if `lookup` were very special and don’t love that it _is_ special.

kriskowal (approved):

packages/daemon/src/types.d.ts
rekmarks: Hm, we have actually accumulated a lot of these `...path` signature functions:
```ts
export interface NameHub {
  has(...petNamePath: string[]): Promise<boolean>;
  identify(...petNamePath: string[]): Promise<string | undefined>;
  locate(...petNamePath: string[]): Promise<string | undefined>;
  reverseLocate(locator: string): Promise<string[]>;
  followLocatorNameChanges(
    locator: string,
  ): AsyncGenerator<LocatorNameChange, undefined, undefined>;
  list(...petNamePath: string[]): Promise<Array<string>>;
  listIdentifiers(...petNamePath: string[]): Promise<Array<string>>;
  followNameChanges(
    ...petNamePath: string[]
  ): AsyncGenerator<PetStoreNameChange, undefined, undefined>;
  lookup(...petNamePath: string[]): Promise<unknown>;
  reverseLookup(value: unknown): Array<string>;
  write(petNamePath: string | string[], id: string): Promise<void>;
  remove(...petNamePath: string[]): Promise<void>;
  move(fromPetName: string[], toPetName: string[]): Promise<void>;
  copy(fromPetName: string[], toPetName: string[]): Promise<void>;
}

export interface EndoDirectory extends NameHub {
  makeDirectory(...petNamePath: string[]): Promise<EndoDirectory>;
}
```

PR #2291, opened 2024-05-16
docs(patterns): Comments for fixed-length heterogeneous collection patterns

kriskowal:
Until we have a rigorous treatment on the usage of patterns, this adds a bandaid to the code itself so anyone looking for how to make a pattern for a struct-like CopyRecord, tuple-like CopyArray, or fixed-length heterogeneous CopyMap will find their way.

### DIFF
--- a/packages/patterns/src/patterns/patternMatchers.js
+++ b/packages/patterns/src/patterns/patternMatchers.js
@@ -1651,13 +1651,33 @@ const makePatternKit = () => {
       limits ? makeLimitsMatcher('match:symbol', [limits]) : SymbolShape,
     record: (limits = undefined) =>
       limits ? M.recordOf(M.any(), M.any(), limits) : RecordShape,
+    // struct: A pattern that matches CopyRecords with a fixed quantity of
+    // entries where the values match patterns for corresponding keys is merely
+    // a hardened object with patterns in the places of values for
+    // corresponding keys.
+    // For example, a pattern that matches CopyRecords that have a string value
+    // for the key 'x' and a number for the key 'y' is:
+    // harden({ x: M.string(), y: M.number() }).
     array: (limits = undefined) =>
       limits ? M.arrayOf(M.any(), limits) : ArrayShape,
+    // tuple: A pattern that matches CopyArrays with a fixed quantity of values
+    // that match a heterogeneous array of patterns is merely a hardened array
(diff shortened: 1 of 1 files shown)


erights (changes requested):
Looks good so far. Do this symmetrically for CopyArrays, Taggeds, CopyMaps, and passable primitives. I think that covers all the cases, but you should double check whether I forgot anything.

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1651,13 +1651,26 @@ const makePatternKit = () => {
...
+    // harden([ M.string(), M.number() ]).
     set: (limits = undefined) => (limits ? M.setOf(M.any(), limits) : SetShape),
     bag: (limits = undefined) =>
       limits ? M.bagOf(M.any(), M.any(), limits) : BagShape,
     map: (limits = undefined) =>
       limits ? M.mapOf(M.any(), M.any(), limits) : MapShape,
+    // The pattern for a CopyMap with a fixed quantity of heterogeneous entry
+    // types is merely a CopyMap of corresponding key and value patterns, for
+    // example,
+    // makeCopyMap([[M.string(), M.number()], [M.number(), M.string()]]).
```
erights, line 1673: No. The keys must be keys. Only the values can be non-key patterns.

erights (changes requested):

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1651,13 +1651,26 @@ const makePatternKit = () => {
       limits ? makeLimitsMatcher('match:symbol', [limits]) : SymbolShape,
     record: (limits = undefined) =>
       limits ? M.recordOf(M.any(), M.any(), limits) : RecordShape,
+    // The pattern for a CopyRecord with a fixed quantity of heterogeneous
+    // entry types (a Struct) is merely a hardened object with the respective
```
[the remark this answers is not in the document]
kriskowal: I rewrote all sections to convey this clarification.

packages/patterns/src/patterns/patternMatchers.js
kriskowal: Rewrote record and map to clarify this distinction.

erights (approved):
LGTM, thanks!

PR #2293, opened 2024-05-20
adopt workspace protocol

turadg:
- **build: declare typescript devDependency**
- **build: emit types by pack hooks**
- **ci: prepack to check dep tree**
- **build: adopt workspace protocol**

Closes: #XXXX
Refs: #XXXX

## Description

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations


packages/compartment-mapper/demo/policy/package.json
```diff
@@ -8,9 +8,9 @@
     "preinstall": "echo DO NOT CALL INSTALL SCRIPTS; exit -1"
   },
   "dependencies": {
-    "@endo/compartment-mapper-demo-lavamoat-style-attenuator": "file:att2",
-    "@endo/compartment-mapper-demo-policy-attenuator1": "file:att1",
-    "@endo/compartment-mapper-demo-polyfill1": "file:polyfill",
```
kriskowal, line 13: This `package.json` should not change since it pertains to a demo.

kriskowal (changes requested):
Just one `package.json` that shouldn’t participate in the workspace. I added `force:integration` to the integration PR. This is all great otherwise.

kriskowal (approved):
Looks like the integration is green. I’m giving a preëmptive approval (contingent on the one `demo/package.json`) to unblock you.

michaelfig:
> Solved by @michaelfig in #2416

I really didn't search for the right terms when I was trying to check for prior work.  Sorry for being pushy. :

turadg:
> * Agoric integration PR: [Endo integration scratch PR Agoric/agoric-sdk#9385](https://github.com/Agoric/agoric-sdk/pull/9385) (it should pass `getting-started` which [failed last time](https://github.com/Agoric/agoric-sdk/issues/9288))

No problem! I'm happy for it to be done. I thought Lerna was the reason I dropped this back to draft but it might have been this:
> Agoric integration PR: https://github.com/Agoric/agoric-sdk/pull/9385 (it should pass getting-started which https://github.com/Agoric/agoric-sdk/issues/9288)

Sorry I didn't find that earlier. I'm afraid you might run into it. Of course it would be great if you also just solved,
- https://github.com/Agoric/agoric-sdk/issues/9288

Issue #400, opened 2020-08-03
Toward importing contract archives in Moddable’s XS


kriskowal:
There are two avenues to running contract archives on Moddable’s XS. Both involve creating an XS binary for hosting dynamically-loaded modules from an archive (presumed Zip), including both CommonJS and ECMAScript modules. In both scenarios, this binary would implement [Endo](https://github.com/Agoric/SES-shim/pull/336), ideally reüsing as much of its internals without modification as possible, so we have parity between running on Node.js with SES-Shim and running on XS without.

1. *Longshot*: Fill out the XS Compartment API (C) such that it can incorporate modules at run-time.
1. *Stopgap*: Emulate run-time modules using XS’s Compartment evaluator.

The Longshot is the most desirable long-term outcome. XS should in principle not need any part of the SES shim, since it’s a SES-only runtime. Also, our friends at Moddable are committed to implementing the Compartment specification, though the specification remains negotiable. To pursue this option, we may need to engage more closely with the XS project.

The Stopgap is possible (modulo risks) without major changes to XS, and largely reusing work we have already done on SES-Shim. We would need to implement the dynamic module loading feature as an `xs-compartment-shim`. This would reüse the `transform-module` Babel shim. We would need to factor some code out of the SES compartment-shim, presumably into the `make-importer` project Michael sketched.

The work break-down for the Longshot includes:

1. XS Compartment would need to implement the `StaticModuleRecord` constructor, so ECMAScript modules can be parsed and analyzed.
1. XS Compartment would need to implement `importHook`.
  1. On the first pass, `importHook` would need to be able to return a static module record provided by the above `StaticModuleRecord` and the Compartment and StaticModuleRecord can communicate behind the scenes with internal slots or weak maps to hide implementation details.
  1. On the second pass, `importHook` ould also support CommonJS, JSON, and potentially other modules by permitting the `importHook` to return  implementations of static module record with the currently unspecified `{ imports, execute(exports, compartment, resolvedImports) }` interface. This is possible as of #397 on Node.js and would be necessary for parity, though we could constrain the transitive dependencies of contracts to be limited to the currently rare ESM format.
[...]

PR #2294, opened 2024-05-21
feat(bundle-source): Zip original sources with --no-transforms

kriskowal:
Closes: #2295 
Refs: #400, #2252

## Description

This change adds a mode to the `bundle-source` command with the initial flag `--no-transforms` that generates “endo zip base64” style bundles without applying the module-to-program transform and SES shim censorship evasion transforms, such that the original files on disk appear in the zip file. This is a preparatory step, necessary for building test artifacts, in advance of full support for this bundle style.

### Security Considerations

`bundle-source` is part of the Endo and Agoric toolkit and it, or its surrogate, participate in the toolchain for generating content that can be confined by Hardened JavaScript, but is not trusted by Hardened JavaScript at runtime. It does however _currently_ run with all the authority of the developer in their development environment and its integrity must be carefully guarded.

### Scaling Considerations

No improvements expected at this time, but in pursuit of #400, it may be possible to move the heavy and performance sensitive JavaScript transform components from `bundle-source` to `import-bundle` and only suffer the performance cost of these transforms on Node.js, where those costs are more readily born by some runtimes. Precompiled bundles may continue to be the preferred medium for deployment to the web, for example.

We will need to advertise the `--no-transforms` flag eventually, since there will be a period where it is advisable if not necessary to generate contracts and caplets targeting the XS runtime.

### Testing Considerations

I have included a test that verifies the API behavior and manually run the following to verify behavior for the CLI:

```sh
rm -rf bundles
yarn bundle-source --no-transforms --cache-json bundles demo/circular/a.js circular-a
rm -rf circular-a
mkdir -p circular-a
jq -r .endoZipBase64 bundles/bundle-circular-a.json | base64 -d > circular-a/circular-a.zip
(cd circular-a; unzip circular-a.zip)
jq . circular-a/compartment-map.json
# verifying the final module entires have parser: 'mjs'
```

This flag is opt-in and breaks no prior behaviors. This introduces a new entry to the build cache meta-data and may cause some bundles to be regenerated one extra time after upgrading.

### Upgrade Considerations

This should not impact upgrade, though it participates in the greater #400 story which will require xsnap upgrades to come to bear.

### DIFF
--- a/packages/bundle-source/src/main.js
+++ b/packages/bundle-source/src/main.js
@@ -1,8 +1,30 @@
 // @ts-check
+import { parseArgs } from 'util';
 import { jsOpts, jsonOpts, makeNodeBundleCache } from '../cache.js';
 
 const USAGE =
   'bundle-source [--cache-js | --cache-json] cache/ module1.js bundleName1 module2.js bundleName2 ...';
+
+const options = /** @type {const} */ ({
+  'no-transforms': {
+    type: 'boolean',
+    short: 'T',
+    multiple: false,
+  },
+  'cache-js': {
+    type: 'string',
+    multiple: false,
+  },
+  'cache-json': {
+    type: 'string',
+    multiple: false,
+  },
+  // deprecated
+  to: {
+    type: 'string',
+    multiple: false,
+  },
+});
 
 /**
  * @param {[to: string, dest: string, ...rest: string[]]} args
@@ -13,20 +35,42 @@ const USAGE =
  * @returns {Promise<void>}
  */
 export const main = async (args, { loadModule, pid, log }) => {
-  const [to, dest, ...pairs] = args;
-  if (!(dest && pairs.length > 0 && pairs.length % 2 === 0)) {
+  const {
+    values: {
+      'no-transforms': noTransforms,
+      'cache-json': cacheJson,
+      'cache-js': cacheJs,
+      // deprecated
+      to: cacheJsAlias,
+    },
+    positionals: pairs,
+  } = parseArgs({ args, options, allowPositionals: true });
+
+  if (
+    !(
+      pairs.length > 0 &&
+      pairs.length % 2 === 0 &&
+      [cacheJson, cacheJs, cacheJsAlias].filter(Boolean).length === 1
+    )
+  ) {
     throw Error(USAGE);
   }
 
+  /** @type {string} */
+  let dest;
   let cacheOpts;
   // `--to` option is deprecated, but we now use it to mean `--cache-js`.
-  if (to === '--to') {
+  if (cacheJs !== undefined) {
+    dest = cacheJs;
     cacheOpts = jsOpts;
-  } else if (to === '--cache-js') {
+  } else if (cacheJsAlias !== undefined) {
+    dest = cacheJsAlias;
     cacheOpts = jsOpts;
-  } else if (to === '--cache-json') {
+  } else if (cacheJson !== undefined) {
+    dest = cacheJson;
     cacheOpts = jsonOpts;
   } else {
+    // unreachable
     throw Error(USAGE);
   }
 
@@ -41,6 +85,8 @@ export const main = async (args, { loadModule, pid, log }) => {
     const [bundleRoot, bundleName] = pairs.slice(ix, ix + 2);
 
     // eslint-disable-next-line no-await-in-loop
-    await cache.validateOrAdd(bundleRoot, bundleName);
+    await cache.validateOrAdd(bundleRoot, bundleName, undefined, {
+      noTransforms,
+    });
   }
 };
--- a/packages/bundle-source/src/zip-base64.js
+++ b/packages/bundle-source/src/zip-base64.js
@@ -7,7 +7,10 @@ import url from 'url';
 import fs from 'fs';
 import os from 'os';
 
-import { makeAndHashArchive } from '@endo/compartment-mapper/archive.js';
+import { defaultParserForLanguage as transformingParserForLanguage } from '@endo/compartment-mapper/archive-parsers.js';
+import { defaultParserForLanguage as transparentParserForLanguage } from '@endo/compartment-mapper/import-parsers.js';
+import { mapNodeModules } from '@endo/compartment-mapper/node-modules.js';
+import { makeAndHashArchiveFromMap } from '@endo/compartment-mapper/archive-lite.js';
 import { encodeBase64 } from '@endo/base64';
 import { whereEndoCache } from '@endo/where';
 import { makeReadPowers } from '@endo/compartment-mapper/node-powers.js';
@@ -17,12 +20,31 @@ const textEncoder = new TextEncoder();
 const textDecoder = new TextDecoder();
 const readPowers = makeReadPowers({ fs, url, crypto });
 
+/**
+ * @param {string} startFilename
+ * @param {object} [options]
+ * @param {boolean} [options.dev]
+ * @param {boolean} [options.cacheSourceMaps]
+ * @param {boolean} [options.noTransforms]
+ * @param {Record<string, string>} [options.commonDependencies]
+ * @param {object} [grantedPowers]
+ * @param {(bytes: string | Uint8Array) => string} [grantedPowers.computeSha512]
+ * @param {typeof import('path)['resolve']} [grantedPowers.pathResolve]
+ * @param {typeof import('os')['userInfo']} [grantedPowers.userInfo]
+ * @param {typeof process['env']} [grantedPowers.env]
+ * @param {typeof process['platform']} [grantedPowers.platform]
+ */
 export async function bundleZipBase64(
   startFilename,
   options = {},
   grantedPowers = {},
 ) {
-  const { dev = false, cacheSourceMaps = false, commonDependencies } = options;
+  const {
+    dev = false,
+    cacheSourceMaps = false,
+    noTransforms = false,
+    commonDependencies,
+  } = options;
   const powers = { ...readPowers, ...grantedPowers };
   const {
     computeSha512,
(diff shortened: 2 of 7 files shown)


kriskowal:
Looking for review buddies for work pursuant to XS native compartments, #400. This change touches `bundle-source` code I’ve shared maintainership with @dckc and compartment-mapper which has benefited tremendously from contributions from @naugtur and @boneskull, built on substantial work by @michaelfig. I would like to page @gibson042 into Compartment Mapper since it touches upon work we propose to TC39.

packages/bundle-source/src/main.js
```diff
@@ -13,6 +13,16 @@ const USAGE =
  * @returns {Promise<void>}
  */
 export const main = async (args, { loadModule, pid, log }) => {
+  let noTransforms = false;
```
boneskull, line 16: suggestion: if we're assuming Node.js, [util.parseArgs](https://nodejs.org/api/util.html#utilparseargsconfig) was created for the "just accept a couple options" use-case

packages/bundle-source/src/zip-base64.js
```diff
@@ -7,7 +7,9 @@ import url from 'url';
 import fs from 'fs';
 import os from 'os';
 
-import { makeAndHashArchive } from '@endo/compartment-mapper/archive.js';
+import { parserForLanguage as transformingParserForLanguage } from '@endo/compartment-mapper/archive-parsers.js';
+import { parserForLanguage as transparentParserForLanguage } from '@endo/compartment-mapper/import-parsers.js';
+import { makeAndHashArchive } from '@endo/compartment-mapper/archive-lite.js';
```
boneskull, line 12: suggestion: feels like we should add some more proper top-level exports to `@endo/compartment-mapper`, but that's just what _I'd_ do.

packages/bundle-source/src/zip-base64.js
```diff
@@ -22,7 +24,12 @@ export async function bundleZipBase64(
   options = {},
   grantedPowers = {},
 ) {
-  const { dev = false, cacheSourceMaps = false, commonDependencies } = options;
+  const {
+    dev = false,
+    cacheSourceMaps = false,
+    noTransforms = false,
+    commonDependencies,
+  } = options;
```
boneskull, line 32: please add types for this function's parameters

packages/bundle-source/test/no-transforms.test.js
```diff
@@ -0,0 +1,36 @@
...
+  const compartmentMapText = new TextDecoder().decode(compartmentMapBytes);
+  const compartmentMap = JSON.parse(compartmentMapText);
+  const { entry, compartments } = compartmentMap;
+  const compartment = compartments[entry.compartment];
+  const module = compartment.modules[entry.module];
+  // Alleged module type is not precompiled (pre-mjs-json)
+  t.is(module.parser, 'mjs');
+
+  const moduleBytes = zipReader.read(
+    `${compartment.location}/${module.location}`,
```
boneskull, line 30: ...I didn't see Windows in the build matrix, so not gonna complain about all this stuff that should probably be `path.join`'d or `path.normalize`'d

packages/compartment-mapper/src/archive-lite.js
```diff
@@ -0,0 +1,475 @@
+// @ts-check
```
boneskull, line 1: tbh I'd love to see a 
```js
/**
 * This module is for doing x, y, z and it is here for reasons.
 * @module
 */
```
at the top of each file.

the subtext being that I'm not sure how these low-calorie modules differ.

not really getting a great idea for how much duplication is going on, either. but that's just cause GH's PR UI won't let me look at two different files side-by-side at the same time.

packages/compartment-mapper/src/archive-lite.js
```diff
@@ -0,0 +1,475 @@
...
+import { parseLocatedJson } from './json.js';
+import { unpackReadPowers } from './powers.js';
+import {
+  assertCompartmentMap,
+  stringCompare,
+  pathCompare,
+} from './compartment-map.js';
+import { detectAttenuators } from './policy.js';
+
+const { details: X } = assert;
```
boneskull, line 34: where does this come from?

packages/compartment-mapper/src/archive-lite.js
boneskull: I'm wondering if it wouldn't be unreasonable to rename some of these subsystems (modules, functions) to something other than `*archive*` if they don't actually interface with archive files (or, generally speaking, just draw the line somewhere else).  I'm using some of the `*archive*`-named API surface just to get at `Sources`.  naming things is hard

packages/compartment-mapper/src/archive.js
```diff
@@ -1,404 +1,26 @@
...
-    archiveSha512 = computeSha512(archiveCompartmentMapBytes);
-  }
-
-  return {
-    compartmentMapBytes: archiveCompartmentMapBytes,
-    sources: archiveSources,
-    sha512: archiveSha512,
-  };
+// Add the default parserForLanguage option.
+const encumber = (options = {}) => {
```
boneskull, line 21:
```suggestion
const assignParserForLanguage = (options = {}) => {
```

or similar.  sorry

packages/compartment-mapper/src/import-parsers.js
```diff
@@ -0,0 +1,16 @@
...
+
+/** @import { ParserImplementation } from './types.js' */
+
+/** @type {Record<string, ParserImplementation>} */
+export const parserForLanguage = {
+  mjs: parserMjs,
+  cjs: parserCjs,
+  json: parserJson,
+  text: parserText,
+  bytes: parserBytes,
```
boneskull, line 15: in case this informs anything you're doing here: I have a branch that has exposed user-defined languages and parsers, which seems like the least-awful way to get native modules running.

boneskull (approved):
I had a vision... a _prophecy_. it was... merge conflicts

packages/compartment-mapper/src/import-parsers.js
kriskowal: I’m not going to land this before Tuesday next week. I do not mind rebuilding this on top of your more rigorous treatment. I’m not jazzed about exposing `parserForLanguage` to the public API but it was sufficient and minimal for the bundleSource needs.

packages/compartment-mapper/src/archive.js
kriskowal: I knew the risks.

packages/compartment-mapper/src/archive-lite.js
kriskowal: `import 'ses'` brings this in.

packages/compartment-mapper/src/archive-lite.js
kriskowal: I’ll circle back and add file docs.

I think it would be reasonable to factor out an API for the leading half of `archive.js` that doesn’t entrain a dependency on Zip.

The thing that makes these `-lite` is that they do not import any parsers, which in turn might entrain Babel.

packages/bundle-source/test/no-transforms.test.js
kriskowal: We recently removed Windows from the build matrix for reasons, but `@endo/zip` only deals in Unixy paths, regardless of the host platform, so we’re good here.

packages/bundle-source/src/zip-base64.js
kriskowal: Much of this is in service of manual tree shaking: allowing the consumer to depend only on what they need. XS in particular is sensitive about Babel, so being able to opt-out depending on which side of compartment-mapper you’re using is key.

packages/compartment-mapper/src/import-parsers.js
boneskull: I'm not going to land my code before Tuesday next week, so it's kind of moot.  I originally had a `parserForLanguage` option, but it turned out that I needed a new _language_, so it became an API that looks like:

```js
parsers: [
  {
    languages: ['native'], 
    parser: someParserImpl, 
    extensions: ['node']
  }
]
```

packages/bundle-source/src/main.js
kriskowal: Done.

packages/compartment-mapper/src/import-parsers.js
kriskowal: I think I can make this work on top of your change without too much trouble. My intent is to make these `-parsers.js` modules continue to export `parserForLanguage` for internal use, but also `languageForExtension` and `parsers` like above that merges the two for public use. I can use the stub module to hide the implementation detail from the public.

packages/bundle-source/test/no-transforms.test.js
```diff
@@ -0,0 +1,36 @@
...
+  );
+  const moduleText = new TextDecoder().decode(moduleBytes);
+  const originalModuleText = await fs.promises.readFile(entryPath, 'utf-8');
+  // And, just to be sure, the text in the bundle matches the original text.
+  t.is(moduleText, originalModuleText);
```
dckc, line 35: 👏

dckc (approved):

packages/compartment-mapper/src/import-parsers.js
kriskowal: For those tuning in, we landed @boneskull’s work with some modifications to compose well with this line of work, and I removed the overlapping commits from this PR in the rebase. I then reconstructed the middle commits of this PR as #2306, as that was common ground for our separate concerns, and then removed the overlapping commits from this PR. All that remains now is the final integration in bundle-source, with very little modification from the changes that have been already approved here. The only big difference is that I’ve additionally decoupled the function that constructs a preliminary compartment map through exploration of `node_modules` on a file system.

PR #2296, opened 2024-05-22
Additional dot-delimited petname path support

FUDCo:
Adds support for dot-delimited petname paths in the `bundle`, `cat`, `follow`, `open`, and `run` commands.  Cleans up handling in the `list` command.


packages/cli/src/commands/cat.js
```diff
@@ -4,10 +4,12 @@ import os from 'os';
 import { E } from '@endo/far';
 import { makeRefReader } from '@endo/daemon';
 import { withEndoAgent } from '../context.js';
+import { parsePetNamePath } from '../pet-name.js';
 
 export const cat = async ({ name, agentNames }) =>
   withEndoAgent(agentNames, { os, process }, async ({ agent }) => {
-    const readable = await E(agent).lookup(name);
+    const namePath = name && parsePetNamePath(name);
+    const readable = await E(agent).lookup(...namePath);
```
rekmarks, line 12: `name` is required for the `cat` command:
```suggestion
    const readable = await E(agent).lookup(...parsePetNamePath(name));
```

packages/cli/src/commands/bundle.js
```diff
@@ -22,11 +23,12 @@ export const bundleCommand = async ({
         format: 'endoZipBase64',
       })
     );
+  const bundlePath = bundleName && parsePetNamePath(bundleName);
```
rekmarks, line 26: Conventionally, we've only treated `undefined` values as absent:
```suggestion
  const bundlePath = bundleName === undefined ? bundleName : parsePetNamePath(bundleName);
```

packages/cli/src/commands/follow.js
```diff
@@ -4,10 +4,12 @@ import os from 'os';
 import { E } from '@endo/far';
 import { makeRefIterator } from '@endo/daemon';
 import { withEndoAgent } from '../context.js';
+import { parsePetNamePath } from '../pet-name.js';
 
 export const followCommand = async ({ name, agentNames }) =>
   withEndoAgent(agentNames, { os, process }, async ({ agent }) => {
-    const iterable = await E(agent).lookup(name);
+    const namePath = name && parsePetNamePath(name);
+    const iterable = await E(agent).lookup(...namePath);
```
rekmarks, line 12: `name` is required for the `follow` command:
```suggestion
    const iterable = await E(agent).lookup(...parsePetNamePath(name));
```

packages/cli/src/commands/open.js
```diff
@@ -6,10 +6,12 @@ import openWebPage from 'open';
 import { E } from '@endo/far';
 
 import { withEndoAgent } from '../context.js';
+import { parsePetNamePath } from '../pet-name.js';
 
 export const open = async ({ webletName, agentNames }) => {
   await withEndoAgent(agentNames, { os, process }, async ({ agent }) => {
-    const weblet = E(agent).lookup(webletName);
+    const webletNamePath = webletName && parsePetNamePath(webletName);
+    const weblet = E(agent).lookup(...webletNamePath);
```
rekmarks, line 14: `name` is required for the `open` command:
```suggestion
    const weblet = await E(agent).lookup(...parsePetNamePath(webletName));
```

packages/cli/src/commands/run.js
```diff
@@ -86,7 +87,8 @@ export const run = async ({
             args.unshift(filePath);
           }
 
-          const readableP = E(agent).lookup(bundleName);
+          const bundleNamePath = bundleName && parsePetNamePath(bundleName);
+          const readableP = E(agent).lookup(...bundleNamePath);
```
rekmarks, line 91: We're in an `if` clause where `bundleName !== undefined`:
```suggestion
          const readableP = E(agent).lookup(...parsePetNamePath(bundleName));
```

packages/cli/src/commands/bundle.js
FUDCo: That makes sense.  I was following the pattern of some of the other commands that had previously been converted to use paths, and it looked like the pattern
```
const fooPath = fooName && parsePetNamePath(fooName);
```
was an established idiom.  If we want to look for explicit `undefined` values then probably those older bits should be updated to match.

Operationally though, these get parsed from the command line by `commander`, which means that the value is always going to be either `undefined` or an actual string, so it's not obvious to me that the additional verbosity buys us much of value.  There is some suspenders-and-belt protection in the event that command implementations get used compositionally in the implementation of other commands, but that's not something I see in the current code and is probably a questionable practice.

packages/cli/src/commands/run.js
FUDCo: Nice catch.

packages/cli/src/commands/bundle.js
FUDCo: After taking a closer look, it turned out that there is only (now) one other occurrences of this pattern, in `run.js`.  However, I think my observation about `commander` processing the args is still apt, so I think a better pattern is to leave the current parse line unchanged but precede it with
```
assert(fooName === undefined || typeof fooName === 'string');
```
I think this will be a little easier on the eyes while still requiring that the invariant that would normally be ensured by `commander` remains true even if the command function is called directly without the command line parse via `commander` as an intermediary.

rekmarks (approved):
LGTM!

PR #2298, opened 2024-05-22
bump Yarn and misc deps

turadg:
evergreen

## Description

Bump some devDeps. While investigating https://github.com/endojs/endo/commit/53f64396204ec113e7d9fb364cc8c7ff56dab31d#commitcomment-142298182

### Security Considerations

none

### Scaling Considerations

none

none

### Testing Considerations

CI

none

### Upgrade Considerations

safe, just dev deps


package.json
```diff
@@ -7,23 +7,23 @@
   "engines": {
     "node": ">=16"
   },
-  "packageManager": "yarn@4.1.1",
+  "packageManager": "yarn@4.2.2",
```
leotm, line 10: was about to PR only this Yarn bump ^ from running `yarn` on `master`
remaining dep bumps lgtm

as an aside (on corepack 0.25.2 locally)

image

it's not obvious why `corepack enable` isn't enabling like suggested

image

wonder if it's worth mentioning somewhere in docs something like _you may need to set your global yarn version to v4 to run corepack`

leotm (approved):

package.json
turadg: You shouldn't have to change your global yarn version. Once corepack is enabled it ignores the global version.

The "Presence…" message you're seeing above is because corepack isn't enabled. It says "Corepack must currently be enabled by running `corepack enable` in your terminal`. I don't know why your command didn't achieve that but maybe try running it separately.

```
corepack enable
which yarn
```

leotm: repro round 2

Screenshot 2024-05-24 at 4 42 18 PM

but agree, shouldn't have to change the global yarn version, like i have to at the root level as a workaround, `corepack enable` should just work 🤔

kriskowal (approved):

PR #2300, opened 2024-05-24
chore(nat): Fix typo in Nat error message

gibson042:
## Description

Adds a missing "is" to an error message from `Nat`.

### Security Considerations

None.

### Scaling Considerations

n/a

n/a

### Testing Considerations

None.

n/a

### Upgrade Considerations

We expect error messages to change over time and require code to be tolerant of this.

### DIFF
--- a/packages/nat/src/index.js
+++ b/packages/nat/src/index.js
@@ -63,7 +63,7 @@ function Nat(allegedNum) {
 
   if (typeof allegedNum === 'number') {
     if (!Number.isSafeInteger(allegedNum)) {
-      throw RangeError(`${allegedNum} not a safe integer`);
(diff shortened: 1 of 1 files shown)


erights (approved):
LGTM thanks.

erights:
> We expect error messages to change over time and require code to be tolerant of this.

Just grepped for "not a safe integer" and verified that none of our endo or agoric-sdk code should have a problem with this change. This change also makes the error more consistent with other errors in agoric-sdk https://github.com/Agoric/agoric-sdk/blob/cc2bf0bb431aa64c601628f53cf191fcc84c5750/packages/ERTP/src/displayInfo.js#L26

PR #2301, opened 2024-05-24
cli: Add a `--verbose` option to the `endo list` command

FUDCo:
The `endo list` command now accepts a `--verbose` option that will cause it to print out a prettified version of the value that would be shown by `endo show` for each entry.

I kept finding myself wanting this during my explorations, so I just went ahead and implemented it.

### DIFF
--- a/packages/cli/src/commands/list.js
+++ b/packages/cli/src/commands/list.js
@@ -1,11 +1,40 @@
 /* global process */
 import os from 'os';
+import { inspect } from 'util';
 import { E } from '@endo/far';
 import { makeRefIterator } from '@endo/daemon';
 import { withEndoHost } from '../context.js';
 import { parsePetNamePath } from '../pet-name.js';
 
-export const list = async ({ directory, follow, json }) =>
+const prettyValue = val => {
+  let result;
+  const type = typeof val;
+  if (type === 'string') {
+    result = `'${val}'`;
+  } else if (type === 'object') {
+    result = `${val}`;
+    const noise = '[object Alleged: ';
+    if (result.startsWith(noise)) {
+      result = result.substring(noise.length);
+      result = result.substring(0, result.length - 1);
+    } else {
+      result = inspect(val);
+    }
+  } else {
+    result = inspect(val);
+  }
+  return result;
+};
+
+const pad = (fieldVal, width, minPad = 2) => {
+  let spaces = width - `${fieldVal}`.length;
+  if (spaces < minPad) {
+    spaces = minPad;
+  }
+  return ' '.repeat(spaces);
+};
+
+export const list = async ({ directory, follow, json, verbose }) =>
   withEndoHost({ os, process }, async ({ host: agent }) => {
     if (directory !== undefined) {
       const directoryPath = parsePetNamePath(directory);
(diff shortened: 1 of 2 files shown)


rekmarks (approved):
LGTM. A couple of inline suggestions.

packages/cli/src/commands/list.js
```diff
@@ -5,7 +5,27 @@ import { makeRefIterator } from '@endo/daemon';
...
+    result = `'${val}'`;
+  } else if (type === 'bigint') {
+    result = `${val}n`;
+  } else if (type === 'object') {
+    result = `${val}`;
+    const noise = '[object Alleged: ';
+    if (result.startsWith(noise)) {
+      result = result.substring(noise.length);
+      result = result.substring(0, result.length - 1);
+    }
```
kriskowal, line 21: This might be more gracefully accomplished with `object[Symbol.toStringTag]`.

packages/cli/src/commands/list.js
```diff
@@ -5,7 +5,27 @@ import { makeRefIterator } from '@endo/daemon';
 import { withEndoHost } from '../context.js';
 import { parsePetNamePath } from '../pet-name.js';
 
-export const list = async ({ directory, follow, json }) =>
+const prettyValue = val => {
+  let result;
+  const type = typeof val;
+  if (type === 'string') {
+    result = `'${val}'`;
```
kriskowal, line 12: I like using JSON.stringify in cases like this since it handles all the edge cases in a way that can be copy+pasted.

packages/cli/src/commands/list.js
```diff
@@ -30,7 +50,17 @@ export const list = async ({ directory, follow, json }) =>
     } else {
       const petNames = await E(agent).list();
       for await (const petName of petNames) {
-        console.log(petName);
+        if (verbose) {
+          let spaces = 20 - `${petName}`.length;
+          if (spaces < 2) {
+            spaces = 2;
+          }
+          const spacer = '                    '.substring(0, spaces);
```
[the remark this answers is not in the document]
FUDCo: I think I was overindexing on performance, which in this case really isn't significant enough to matter.  The motivation behind the original formulation was that the string of spaces was static, rather than rebuilding the string each pass through the loop.  But if you sign up for the cost of using `repeat` then I think it makes even more sense to instead just say:
```
const spacer = ' '.repeat(spaces);
```
which upon looking at it seems preferable to either alternative because the underlying intent is much clearer.

packages/cli/src/commands/list.js
FUDCo: That doesn't deal with the cases where the object in question doesn't have a string tag, plus the main motivation for me was getting rid of the "Alleged", which in spite of Mark's good intentions is in practice a noise word.

packages/cli/src/commands/list.js
FUDCo: What I really want is Node's `util.inspect`, which handles a bunch of stuff that chokes `JSON.stringify`.  You get that for free if you pass a naked value to `console.log`, but if you want to handle the string itself you need to be running in Node, which I don't think is an invariant we want here.  However, it occurs to me that since the value is the last thing on the line we can handle cases we don't know how to format but literally just passing the unformatted value as a second arg to `console.log`.

FUDCo: Very good.  That simplifies quite a few things.

Issue #2303, opened 2024-05-28
custom parser support


boneskull:
## What is the Problem Being Solved?

LavaMoat's use-case needs support for native modules (e.g., C++ code compiled to a `.node` file).

Endo, however, does not need to know anything about native modules.  To that end, we're proposing _custom parsers_.

## Description of the Design

To support this, APIs such as `importLocation` would need to accept a new option.  For example, given this interface:

```ts
interface CustomParser {
  /**
   * parser implementation; nothing special here
   */
  parser: ParserImplementation;

  /**
   * A new or known `Language`
   */
  language: Language;

  /**
   * One or more extensions which will be mapped to the value of `language`
   * no leading dot!
   */
  extensions: string[];
}
```

The `options` object could accept a value `parsers` of type `CustomParser[]`.

> To support this, the `Language` type changes to becomes a union of the known `Language` strings _and_ any `string`. Take care not to widen the type so that it is always inferred as `string`.

The parser-mapping code would then add these values to the pile of data structures it uses to determine which parser to use for what file.  

For now, the API should only allow _new_ parsers. It should not allow a builtin-parser to be replaced outright. If that feature is desired, it could be added in a future PR in a backwards-compatible manner.

### Package-Level Custom Policy Fields

Which packages can use which parsers can be controlled via policy.

The `Policy` type is flexible via type arguments, but not quite flexible enough. Currently, there's no way to add custom field to the policy at the package level (e.g., a new prop on `PackagePolicy`).

To avoid an explosion of custom properties (and to avoid Endo needing to know anything about native modules), `PackagePolicy` will now have an optional field `options` (of default type `unknown`) which can be narrowed via type argument.  For example:

```ts
import {type Policy} from '@endo/compartment-mapper';

type CustomOptions = {
  foo: string;
} | number;

// the `void`s are just default args for additional PolicyItem values (packages, globals, builtins)
type CustomPolicy = Policy<void, void, void, CustomOptions>;

const policy: CustomPolicy = {
  resources: {
    butts: {
      options: {
        foo: 'pants'
      }
    }
  }
};
```

`@endo/compartment-mapper`'s internal policy validator doesn't need to know anything about `options` other than it may exist.

## Security Considerations

Certainly, whoever wields a custom parser must accept full responsibility for its misuse.  The way we'd be using a custom parser is a full-blown escape hatch; a native module will throw any semblance of safety out the window.

## Scaling Considerations

None--_if_ the feature is unused.

## Test Plan

[...]

PR #2304, opened 2024-05-28
feat(compartment-mapper): Custom parser support

boneskull:
Closes: #2303

## Description

This PR adds custom parser support as discussed in #2303. To support the feature:

- A new prop `options` was added to the `PackagePolicy` type.
- Parsers now optionally accept `compartmentDescriptor` objects, so that parsers can enforce policy

### Security Considerations

See #2303

### Scaling Considerations

See #2303

- Add documentation for a new option `parsers` for e.g., `importLocation`
- Add documentation for a new prop `options` in `PackagePolicy`

### Testing Considerations

- [x] A round-trip test 
- [x] A round-trip test w/ policy enforcement 
- [x] Assert `options` is ignored by policy validator
- [x] Assert custom parsers cannot override builtin parsers

None

### Upgrade Considerations

- Update `NEWS.md`


packages/compartment-mapper/test/custom-parser.test.js
```diff
@@ -0,0 +1,67 @@
...
+          // This parser parses markdown files. Use your imagination
+          parse: async (
+            bytes,
+            specifier,
+            moduleLocation,
+            packageLocation,
+            { compartmentDescriptor },
+          ) => {
+            const execute = moduleEnvironmentRecord => {
+              moduleEnvironmentRecord.default = `${bytes}`;
```
kriskowal, line 32: My imagination requires some assistance (in order to continue happily not knowing what TypedArray.prototype.toString does instead of UTF-8)
```suggestion
              moduleEnvironmentRecord.default = new TextDecoder().decode(bytes);
```

packages/compartment-mapper/src/types.js
```diff
@@ -478,29 +499,71 @@ export {};
...
+/**
+ * Allows creation of a union of a primitive and a literal type which is not loosened to the primitive.
+ *
+ * Aids type hints in IDEs.
+ *
+ * Lifted from {@link https://npm.im/type-fest type-fest}
+ *
+ * @template LiteralType The literal type
+ * @template {Primitive} PrimitiveType The primitive type
+ * @typedef {LiteralType | (PrimitiveType & Record<never, never>)} LiteralUnion
```
kriskowal, line 564: Stunning.

packages/compartment-mapper/test/custom-parser.test.js
```diff
@@ -0,0 +1,67 @@
...
+            };
+
+            t.truthy(
+              compartmentDescriptor,
+              'compartmentDescriptor is passed to parser',
+            );
```
kriskowal, line 38: I could use a demonstration of a behavior that would switch on the compartment descriptor.

packages/compartment-mapper/src/import.js
```diff
@@ -1,7 +1,7 @@
 // @ts-check
 /* eslint no-shadow: "off" */
 
-/** @import {Application} from './types.js' */
+/** @import {Application, ExtraImportOptions} from './types.js' */
```
kriskowal, line 4: The “extra” qualifier here isn’t doing a bunch of work.
```suggestion
/** @import {Application, ImportOptions} from './types.js' */
```

packages/compartment-mapper/src/types.js
```diff
@@ -387,6 +396,18 @@ export {};
...
+ * @typedef CustomParser
+ * @property {ParserImplementation} parser Parser implementation
+ * @property {Language} language Language the parser applies to. Can be new or existing
+ * @property {string[]} extensions Extensions the language applies to
+ */
+
+/**
+ * @typedef {object} ExtraImportOptions
+ * @property {CustomParser[]} [parsers] Custom parser objects
+ */
```
kriskowal, line 409: What would this be named if we also supported custom parsers for `archive.js`, `bundle.js`, and `import-archive.js`? Perhaps we would be better served avoiding a name for it by defining the `ImportOptions`, `ImportArchiveOptions`, and `ArchiveOptions` and just including `parsers` on the ones that implement them so far.

kriskowal (changes requested):
Design direction seems fine. I like to do `import`, `archive`, and `import-archive` together for any feature that applies to all three, to make sure the design is sound. A `scaffold` test should verify that an application has the same behavior through `import` or `archive` + `import-archive` using custom parsers. Any commentary on why `archive` and `import-archive` should not support custom parsers would be welcome. Please request a review again when this PR is out of draft.

Thank you for sharing `LiteralUnion`. There are many places where I imagine this pattern should be applied. I welcome a comment from @turadg whether this should become integral to our style for enum literals.

packages/compartment-mapper/test/custom-parser.test.js
naugtur: Think policy enforcement - the parser is responsible for the security or the confinement of what it's processing and custom extension to policy could configure its behavior.

packages/compartment-mapper/src/types.js
turadg: Please lift the whole thing, since it explains why it's necessary and when it will not longer be necessary. 

Consider separating a `type-fest.d.ts` or `utils.d.ts` that has these utility types that aren't of compartment-mapper per se.

> I welcome a comment from @turadg whether this should become integral to our style for enum literals.

Union of literals is how in agoric-sdk we specify possible values for enums. But that doesn't require this new type. This one is only for "when a union type of a primitive type is combined with literal types".

boneskull: @turadg Lifting the whole thing from `type-fest` would involve several files.  What about: move this to `type-fest.d.ts`, and copy a bunch of the relevant comments over?

packages/compartment-mapper/src/import.js
boneskull: I will rename this.

packages/compartment-mapper/src/types.js
boneskull: Yeah, tbh the piecemeal definition of those options on the function definitions has been bothering me.  I'll do something here.

packages/compartment-mapper/src/types.js
boneskull: @turadg Also, would you still want this file in `compartment-mapper`?

packages/compartment-mapper/test/custom-parser.test.js
boneskull: Yes, that's one of the test cases I haven't implemented yet.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -169,23 +169,25 @@ const findPackage = async (readDescriptor, canonical, directory, name) => {
...
   cjs: 'cjs',
   mjs: 'mjs',
   json: 'json',
   text: 'text',
   bytes: 'bytes',
-};
-const commonParsers = { js: 'cjs', ...uncontroversialParsers };
-const moduleParsers = { js: 'mjs', ...uncontroversialParsers };
+});
+const COMMON_PARSERS = /** @type {const} */({ js: 'cjs', ...UNCONTROVERSIAL_PARSERS });
```
boneskull, line 180: I'll also undo the `SCREAMING_SNAKE_CASE` here

packages/compartment-mapper/src/node-modules.js
```diff
@@ -169,23 +169,25 @@ const findPackage = async (readDescriptor, canonical, directory, name) => {
...
+const MODULE_PARSERS = /** @type {const} */({ js: 'mjs', ...UNCONTROVERSIAL_PARSERS });
+// * @param {{languages?: readonly string[]|string[], uncontroversialParsers?: Record<string, string>, commonParsers?: Record<string, string>, moduleParsers?: Record<string, string>}} [options]
```
boneskull, line 182: looks removable

packages/compartment-mapper/src/node-modules.js
```diff
@@ -169,23 +169,25 @@ const findPackage = async (readDescriptor, canonical, directory, name) => {
...
 
 /**
  * @param {object} descriptor
  * @param {string} location
+
```
boneskull, line 187:
```suggestion
```

packages/compartment-mapper/src/types.js
turadg: > Lifting the whole thing from type-fest would involve several files

@boneskull , I just meant the whole jsdoc (basically https://github.com/sindresorhus/type-fest/blob/main/source/literal-union.d.ts)

> move this to type-fest.d.ts, and copy a bunch of the relevant comments over?

Works for me. Slight preference for `utils.d.ts` to allow for more like this. You might need an empty `.js` twin; our linter config has gotten tripped up before I think.

> would you still want this file in compartment-mapper?

It's so small I think it's fine to vendor in each package that uses it. TypeScript is structural so there won't be any difference in its type resolution.

boneskull:
@kriskowal 

> I like to do import, archive, and import-archive together for any feature that applies to all three, to make sure the design is sound. 

Is "import-archive" the "unarchive" to the "archive" case?  If so, I'm assuming that the same custom parser would need to be provided to both invocations. What happens (or should happen) if this is not true?  Would we need to do something like add the parser itself to the archive?

kriskowal:
> Is "import-archive" the "unarchive" to the "archive" case?

Indeed.

> If so, I'm assuming that the same custom parser would need to be provided to both invocations. What happens (or should happen) if this is not true? Would we need to do something like add the parser itself to the archive?

It’s good enough to require that the same custom parsers be present for `makeArchive` and `parseArchive` (and their analogues). I don’t care to fantasize about capturing the custom parser in the archive at this time, and that can be a different route if we do go that direction.

kriskowal:
Analogously, `makeArchive` (and `writeArchive` etc) has to receive compatible `exits` with `parseArchive` (and `importArchive` etc). Coördinating out of band is expected.

packages/compartment-mapper/src/types.js
boneskull: @turadg AFAICT the TS configuration won't consider the validity of a `.d.ts` file when type-checking.  I'm not sure of a way to force the checking of a single `.d.ts` file.  However, a `.ts` or `.js` file _would_ be type-checked.  To that end, it may be more straightforward to leave these definitions as `@typedef`s in a `.js` file.

turadg: You're right that it won't typecheck the `.d.ts` itself (because of skipLibCheck) but it would observe the type when imported into another module. We do this all the time in agoric-sdk. https://github.com/Agoric/agoric-sdk/blob/master/docs/typescript.md#dts-modules

But I don't want to slow you down. Factor it however you like. My only "please" was to include the full JSDoc comment. The other points were to "consider".

packages/compartment-mapper/src/link.js
```diff
@@ -372,6 +397,10 @@ export const link = (
     // The `moduleMapHook` writes back to the compartment map.
     compartmentDescriptor.modules = modules;
 
+    for (const [extension, language] of entries(customLanguageForExtension)) {
+      languageForExtension[extension] = language;
+    }
```
kriskowal, line 402: The `parsers` must take precedence over or override the default language for extension. We should have a test to confirm this.

packages/compartment-mapper/src/link.js
```diff
@@ -356,9 +357,33 @@ export const link = (
...
+        extension in customLanguageForExtension &&
+        customLanguageForExtension[extension] !== language
+      ) {
+        throw new Error(
+          `Extension ${q(extension)} already assigned language ${q(customLanguageForExtension[extension])}`,
+        );
+      }
+      customLanguageForExtension[extension] = language;
+    }
+  }
```
kriskowal, line 381: I’m beginning to suspect it’s better to thread separate “global” options for `parserForLanguage` and `languageForExtension` than to merge them into `parsers` and then unmerge them here. In https://github.com/endojs/endo/pull/2294, I would have to make `@endo/compartment-mapper/import-parsers.js` express an opinion about the language for each extension, and there is no clear opinion at that juncture.

boneskull:
@kriskowal OK, this is ready for Proper Review.  Please take a looksee

boneskull:
will update `NEWS.md` as a last step

packages/compartment-mapper/src/archive.js
```diff
@@ -39,16 +39,20 @@ import { detectAttenuators } from './policy.js';
...
+  /** @type {const} */ ({
+    mjs: parserArchiveMjs,
+    'pre-mjs-json': parserArchiveMjs,
+    cjs: parserArchiveCjs,
+    'pre-cjs-json': parserArchiveCjs,
+    json: parserJson,
+    text: parserText,
+    bytes: parserBytes,
+  }),
+);
```
boneskull, line 55: This (and the other one) shouldn't be mutated.  Calling it `const` isn't enough 😄

packages/compartment-mapper/src/import-archive.js
```diff
@@ -1,6 +1,23 @@
...
+/** @import {ComputeSourceMapLocationHook} from './types.js' */
+/** @import {ExecuteFn} from './types.js' */
+/** @import {ExitModuleImportHook} from './types.js' */
+/** @import {HashFn} from './types.js' */
+/** @import {ImportHookMaker} from './types.js' */
+/** @import {LoadArchiveOptions} from './types.js' */
+/** @import {ParserForLanguage} from './types.js' */
+/** @import {ReadFn} from './types.js' */
+/** @import {ReadPowers} from './types.js' */
+/** @import {SomeObject} from './types.js' */
```
boneskull, line 19: thank you, copilot

packages/compartment-mapper/src/node-modules.js
```diff
@@ -169,23 +169,45 @@ const findPackage = async (readDescriptor, canonical, directory, name) => {
...
+});
+const defaultModuleParsers = /** @type {const} */ ({
+  js: 'mjs',
+  ...defaultUncontroversialParsers,
+});
 
 /**
  * @param {object} descriptor
  * @param {string} location
+ * @param {{languages?: readonly string[]|string[], uncontroversialParsers?: Record<string, string>, commonParsers?: Record<string, string>, moduleParsers?: Record<string, string>}} [options]
```
boneskull, line 199: probably wants a typedef, but I'm not gonna bother unless someone asks

packages/compartment-mapper/test/policy-format.test.js
```diff
@@ -62,6 +62,12 @@ const q = JSON.stringify;
   {
     globals: ['a', {}],
   },
+  {
+    options: {
```
boneskull, line 66: Do we care that it's possible to put unserializable junk in here?  Should that be restricted via types or at runtime or both or neither?

boneskull:
@kriskowal Note: there's no explicit test for the archive use-case. I am not sure if it's needed?@k

packages/compartment-mapper/src/link.js
```diff
@@ -372,16 +411,42 @@ export const link = (
...
+      finalLanguageForExtension[extension] = language;
+    }
+    for (const [extension, language] of entries(languageForExtension)) {
+      finalLanguageForExtension[extension] = language;
+    }
+
+    /** @type {ParserForLanguage} */
+    const finalParserForLanguage = Object.create(null);
+
+    for (const [language, parser] of [
```
boneskull, line 431: I noticed that putting stuff in `parserForLanguage` seems to leak out of tests, so a) I stopped doing that, and b) I froze them.

packages/compartment-mapper/src/link.js
```diff
@@ -372,16 +411,42 @@ export const link = (
...
+    for (const [extension, language] of entries(customLanguageForExtension)) {
+      if (extension in languageForExtension) {
+        throw new Error(
+          `Extension ${q(extension)} already assigned language ${q(languageForExtension[extension])}`,
+        );
+      }
+      finalLanguageForExtension[extension] = language;
+    }
+    for (const [extension, language] of entries(languageForExtension)) {
+      finalLanguageForExtension[extension] = language;
```
boneskull, line 425: I'm not sure if I should write anything back to the `CompartmentDescriptor`.

packages/compartment-mapper/src/link.js
```diff
@@ -372,16 +411,42 @@ export const link = (
     // The `moduleMapHook` writes back to the compartment map.
     compartmentDescriptor.modules = modules;
 
+    /** @type {Record<string, Language>} */
+    const finalLanguageForExtension = Object.create(null);
+    for (const [extension, language] of entries(customLanguageForExtension)) {
+      if (extension in languageForExtension) {
```
boneskull, line 417: Should I be using `has` instead of this?  It's habit (because TypeScript seems to prefer it).

kriskowal: Since they’re equivalent for `Object.create(null)`, I’m inclined to overlook the question. Unless I’m wrong about them being equivalent. @erights for second opinion.

packages/compartment-mapper/src/link.js
kriskowal: No, this is fine.

packages/compartment-mapper/src/node-modules.js
kriskowal:
```suggestion
 * @param {object} [options]
 * @param {readonly string[]|string[]} [options.languages]
 * @param {Record<string, string>} [options.uncontroversialParsers]
 * @param {Record<string, string>} [options.commonParsers]
 * @param {Record<string, string>} [options.moduleParsers]
```

packages/compartment-mapper/src/policy-format.js
```diff
@@ -200,7 +205,7 @@ export const assertPackagePolicy = (allegedPackagePolicy, path, url) => {
  * It also moonlights as a type guard.
  *
  * @param {unknown} allegedPolicy - Alleged `Policy` to test
- * @returns {asserts allegedPolicy is import('./types.js').Policy|undefined}
+ * @returns {asserts allegedPolicy is SomePolicy|undefined}
```
kriskowal, line 208: Is there a precedent for `Some*` types already? Can we consider `Maybe*` instead?

packages/compartment-mapper/test/policy-format.test.js
kriskowal: I defer to @naugtur

packages/compartment-mapper/src/archive.js
kriskowal: You caught them all.

kriskowal (changes requested):
This is close to ready.

I investigated changing the `parsers` option to separate `parserForLanguage` and `languageForExtension` options and I like the results. I offer these commits for your consideration. I would like to subsume it into this PR in some fashion or another and I am not concerned about preserving my authorship metadata.  https://github.com/endojs/endo/compare/boneskull/native-parser...kriskowal-native-parser
We typically use a rebase and merge commit and preserve the narrative of the commit history. Let me know if it was your intention to squash+merge the PR or if you want me to review the narrative of the commit history.

packages/compartment-mapper/src/link.js
kriskowal: I did some additional work to ensure that `in` behaves as I would expect. In the case of `options` where a devious user might provide an object with a prototype chain that includes enumerable properties, the difference between `has` and `in` would be observable, but that can be avoided by capturing options with `freeze(create(null), option)`.

packages/compartment-mapper/src/link.js
kriskowal: This branch captures my investigation and I am convinced that threading `parserForLanguage` and `languageForExtension` as separate options (instead of consolidated custom `parsers`) is robust with less ceremony and will compose better with changes I have coming for #400 

https://github.com/endojs/endo/compare/boneskull/native-parser...kriskowal-native-parser

boneskull:
@kriskowal OK, I've cherry-picked your changes, then fixed a problem in `bundle.js`.  This is ready for final review.  Once approved, I'll squash and merge.

kriskowal (approved):
Great, thank you! And thanks for offering to squash+merge. No notes on commit structure or messages.

PR #2306, opened 2024-06-03
feat(compartment-map): Decouple discovery and languages

kriskowal:
Refs: #400 and #2294

## Description

In this change, we carve and export `-lite.js` and `-parsers.js` out of `import.js`, `archive.js`, and `import-archive.js`, as well revealing `mapNodeModules` through `@endo/compartment-mapper/node-modules.js` revealing hidden flexibility already present in the Compartment Mapper. This allows the Compartment Mapper to mix and match “language behaviors” with different workflows (e.g., using `import-parsers.js` with `archive-lite.js` instead of `archive-parsers.js` generates archives with original sources instead of precompiled sources.) We also decouple the process of discovering the physical compartments such that `node-modules.js` can be substituted for alternate compartment exploration algorithms, e.g., combining a custom lockfile and package cache from a specific package management solution.

### Security Considerations

Does not cross security considerations.

### Scaling Considerations

Exporting more narrowly scoped modules allows consumers to be avoid entraining Babel in a greater variety of scenarios.

Evolution of the reference documentation should suffice, though another pass at README.md to dig into these more surgical and composable APIs may be worthwhile.

### Testing Considerations

The existing scenarios are extensive and fully cover the behaviors that we’ve factored out. I expect #2294 tests to cover the new scenarios.

This change takes great care to preserve the existing behavior of the composite `import.js`, `archive.js`, and `import-archive.js` interfaces, only revealing existing behaviors that were previously internal, allowing more expressible scenarios.

### Upgrade Considerations

This change will not affect upgrade and paves a migration path forward for preserving backward-compatibility for `bundleSource` and `importBundle` as we draw closer to supporting XS native compartments.


kriskowal:
Advise reviewing commits individually. I took some care to rename and edit in separate steps.

packages/compartment-mapper/src/archive-parsers.js
```diff
@@ -0,0 +1,18 @@
...
+  /** @type {const} */ ({
+    mjs: parserArchiveMjs,
+    'pre-mjs-json': parserArchiveMjs,
+    cjs: parserArchiveCjs,
+    'pre-cjs-json': parserArchiveCjs,
+    json: parserJson,
+    text: parserText,
+    bytes: parserBytes,
+  }),
+);
```
boneskull, line 18: jeremiah-johnson-1972-v0-v4zsx95wn5xb1

packages/compartment-mapper/package.json
```diff
@@ -28,8 +28,14 @@
       "default": "./index.js"
     },
     "./import.js": "./import.js",
+    "./import-lite.js": "./import-lite.js",
```
boneskull, line 31: (suggestion/observation/trolling) there's obviously a convention at play here, but I'd omit the `.js` from the keys

packages/compartment-mapper/src/archive.js
```diff
@@ -0,0 +1,91 @@
...
+const { assign, create, freeze } = Object;
+
+/** @import {ArchiveOptions} from './types.js' */
+/** @import {ReadFn} from './types.js' */
+/** @import {ReadPowers} from './types.js' */
+/** @import {HashPowers} from './types.js' */
+/** @import {WriteFn} from './types.js' */
+
+// Add the default parserForLanguage option.
+const assignParserForLanguage = (options = {}) => {
```
boneskull, line 22: pls add `@param` type(s)

packages/compartment-mapper/src/import-archive.js
```diff
@@ -0,0 +1,80 @@
...
+import { defaultParserForLanguage } from './import-archive-parsers.js';
+import {
+  parseArchive as parseArchiveLite,
+  loadArchive as loadArchiveLite,
+  importArchive as importArchiveLite,
+} from './import-archive-lite.js';
+
+const { assign, create, freeze } = Object;
+
+/** @import {Application, ComputeSourceLocationHook, ComputeSourceMapLocationHook, ExecuteOptions, ExitModuleImportHook, HashFn, LoadArchiveOptions, ReadPowers} from './types.js' */
```
boneskull, line 12: probably should be one import per line

packages/compartment-mapper/src/import-archive.js
```diff
@@ -0,0 +1,80 @@
...
+/** @import {ParserForLanguage} from './types.js' */
+
+// Have to give it a name to capture the external meaning of Compartment
+// Otherwise @param {typeof Compartment} takes the Compartment to mean
+// the const variable defined within the function.
+//
```
boneskull: (suggestion) IMO it isn't necessary to explain this, but nbd.  I would probably even just zap the `@typedef` altogether and use `typeof Compartment` in the `parseArchive` signature's docstring

packages/compartment-mapper/src/import.js
```diff
@@ -0,0 +1,51 @@
...
+const { assign, create, freeze } = Object;
+
+/** @import {Application} from './types.js' */
+/** @import {ArchiveOptions} from './types.js' */
+/** @import {ExecuteOptions} from './types.js' */
+/** @import {ReadFn} from './types.js' */
+/** @import {ReadPowers} from './types.js' */
+
+// Add the default parserForLanguage option.
+const assignParserForLanguage = (options = {}) => {
```
boneskull: would it make sense to create a factory function for this?  idk how much DRY you're going for here.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -784,3 +784,39 @@ export const compartmentMapForNodeModules = async (
 
   return compartmentMap;
 };
+
+/**
+ * @param {ReadFn | ReadPowers} readPowers
+ * @param {string} moduleLocation
+ * @param {object} [options]
+ */
+export const mapNodeModules = async (
```
boneskull, line 793: public APIs ought to have `@returns`

boneskull (approved):
if I'm being fussy, I'd want to see, for each of these new files, a module-level comment describing the module's use-case. and descriptions for each of the exported functions. 

probably needs a `NEWS.md` update.

there's nothing here that I'd block about.

boneskull (approved):
if I'm being fussy, I'd want to see, for each of these new files, a module-level comment describing the module's use-case. and descriptions for each of the exported functions. 

probably needs a `NEWS.md` update.

there's nothing here that I'd block about.

packages/compartment-mapper/package.json
kriskowal: I very much want to. It’s not just convention, though. This makes the Compartment Mapper continue to work when used on some older foundations like `node -r esm` that do not read the `"exports"` directive, so we can’t use it for aliases, just to limit exposure. We have one lingering integration test that needs to be replaced before we can drop this pattern.

packages/compartment-mapper/src/import-archive.js
kriskowal: Using `typedef Compartment` doesn’t work for some reason, thus the comment. My preference as well.

packages/compartment-mapper/src/import.js
kriskowal: Meme with the normal curve where the developer is maximally DRY in the middle of their spiritual journey dot GIF.

packages/compartment-mapper/package.json
boneskull: Didn't realize that `esm` was a target

packages/compartment-mapper/src/archive-parsers.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * @file Provides a set of default language behaviors (parsers) suitable for
+ * creating archives (zip files with a `compartment-map.json` and a file for
+ * each module) with pre-compiled sources.
+ *
+ * This module entrains a dependency upon the core of Babel.
+ */
```
boneskull, line 7: You want [`@packageDocumentation`](https://typedoc.org/tags/packageDocumentation/) or [`@module`](https://typedoc.org/tags/module/) instead of `@file` (which is seemingly unsupported by TypeDoc).

It should appear below the text.

```suggestion
 * Provides a set of default language behaviors (parsers) suitable for
 * creating archives (zip files with a `compartment-map.json` and a file for
 * each module) with pre-compiled sources.
 *
 * This module entrains a dependency upon the core of Babel.
 * 
 * @packageDocumentation
 */
```

packages/compartment-mapper/src/link.js
```diff
@@ -494,6 +503,7 @@ export const link = (
 /**
  * @param {CompartmentMapDescriptor} compartmentMap
  * @param {LinkOptions} options
+ * @deprecated Use {@see link}.
```
boneskull, line 506:
```suggestion
 * @deprecated Use {@link link}.
```

packages/compartment-mapper/src/node-modules.js
```diff
@@ -784,3 +795,40 @@ export const compartmentMapForNodeModules = async (
 
   return compartmentMap;
 };
+
+/**
+ * @param {ReadFn | ReadPowers} readPowers
+ * @param {string} moduleLocation
+ * @param {object} [options]
```
boneskull, line 802: ...but public APIs should also have fully-typed params. 😄

boneskull (changes requested):
mo tags mo problems

packages/compartment-mapper/src/import-archive.js
boneskull: @kriskowal FWIW I cannot reproduce whatever the problem was that you ran into.  This compiles cleanly for me:

```js
/**
 * @typedef {object} Options
 * @property {string} [expectedSha512]
 * @property {HashFn} [computeSha512]
 * @property {Record<string, unknown>} [modules]
 * @property {ExitModuleImportHook} [importHook]
 * @property {typeof Compartment} [Compartment]
 * @property {ComputeSourceLocationHook} [computeSourceLocation]
 * @property {ComputeSourceMapLocationHook} [computeSourceMapLocation]
 * @property {ParserForLanguage} [parserForLanguage]
 */
```

kriskowal: Yeah, there are `Compartment` compatible constructors created by wrapping the `Compartment` constructor. Notably, `Compartment` before lockdown is different than after. Every `Compartment` has its own instance of `Compartment`. They’re structurally compatible, though. Unclear to me why that would make a difference to types, but they do have different values.

boneskull: even though TS is structural, because `Compartment` is a function, it will only ever allow the global `Compartment` from whichever version of `ses` provides it.  That may be fine, but also seems redundant.  Though it may be useful if we're using some mocked-out `Compartment` in tests?

kriskowal:
```
src/import-archive-lite.js:266:4 - error TS2502: 'Compartment' is referenced directly or indirectly in its own type annotation.

266  * @param {typeof Compartment} [options.Compartment]
```

packages/compartment-mapper/src/archive-parsers.js
kriskowal: Elected to leave these as ordinary comments until such a time as we take the trouble to set up typedoc for this package.

packages/compartment-mapper/src/import-archive.js
boneskull: ah, I see that now.  It's because of the `DefaultCompartment`.  `typeof DefaultCompartment` works

boneskull (approved):
:shipit:

PR #2310, opened 2024-06-07
feat(compartment-mapper): Support dynamic require/importNow

boneskull:
## Description

- This PR adds support for dynamic requires via `loadFromMap()` and `link()` (in `import-lite.js` and `link.js`, respectively).  `importLocation()`'s signature has also been modified for support.  

    To use this feature in either function, the following must be true:

    1. The `moduleTransforms` option (in the appropriate `options` parameter) must not be present. These are _asynchronous_ module transforms, which cannot be used by dynamic require.
    2. The `ReadPowers` param must be a proper `ReadPowers` object (not just a `ReadFn`) _and_ must contain both the _new_ `maybeReadSync`, _new_ `isAbsolute`, and `fileURLToPath` functions. The new type representing this is `SyncReadPowers`.
    
    If all of the above are true, then a compartment will be allowed to dynamically require something.  If that thing still cannot be found in the compartment map, the sync "fallback" exit hook (see next item) will be executed.  
- The new `importHookNow` property can be provided via options, which is a synchronous exit module import hook.
- About `SyncReadPowers`: 
    - `SyncReadPowers.maybeReadSync()` is necessary to read files synchronously which is necessary to load them synchronously. This fails gracefully if the file is not found; the implementation is platform-dependent.
    - `SyncReadPowers.isAbsolute()` is necessary to determine if the module specifier of a dynamic require is absolute. If it is, it could be just about anything, and must be loaded via the user-provided `importNowHook` (the sync exit module import hook).  
    - `SyncReadPowers.fileURLToPath()` is needed for `__filename` and `__dirname`, which is often used to create absolute paths for requiring dynamically.
- As an alternative to `moduleTransforms`, _synchronous_ module transforms may be provided via the _new_ `syncModuleTransforms` object.  In a non-dynamic-require use-case, if present, `syncModuleTransforms` are combined with the `moduleTransforms` option; all sync module transforms are module transforms, but not all module transforms are sync module transforms.
- **All builtin parsers are now synchronous**.  User-defined parsers _can_ be async, but this will disable dynamic require support.
- `@endo/evasive-transform` now exports `evadeCensorSync()` in addition to `evadeCensor()`. This is possible because I've swapped the async-only [source-map](https://npm.im/source-map) with [source-map-js](https://npm.im/source-map-js), which is a fork of the former before it went async-only.  `source-map-js` claims comparable performance.

### Security Considerations

Dynamically requiring an exit module (e.g., a Node.js builtin) requires a user-defined hook, which has the same security considerations as a user-defined exit module hook.  

Swapping out a dependency (`source-map-js` for `source-map`) incurs risk.

### Scaling Considerations

n/a

Should be announced as a user-facing feature

### Testing Considerations

I've added some fixtures and tested around the conditionals I've added, but am open to any suggestions for additional coverage.

This increases ecosystem compatibility considerably; use of dynamic require in the ecosystem is not rare.  

For example, most packages which ship a native module will be using dynamic require, because the filepath of the build artifact is dependent upon the platform and architecture.

### Upgrade Considerations

Everything else _should_ be backwards-compatible, as long as `source-map-js` does as it says on the tin.

Users of `@endo/evasive-transform` may note that native modules are neither downloaded/compiled (due to the switch from `source-map` to `source-map-js`).

* * *

## UPDATE: Aug 18 2024 

This PR now targets the `boneskull/supertramp` branch, which is PR #2263

### DIFF
--- a/packages/compartment-mapper/NEWS.md
+++ b/packages/compartment-mapper/NEWS.md
@@ -1,5 +1,12 @@
 User-visible changes to `@endo/compartment-mapper`:
 
+# Next release
+
+- Adds support for dynamic requires in CommonJS modules. This requires specific
+  configuration to be passed in (including new read powers), and is _not_
+  enabled by default. See the signature of `loadFromMap()` in `import-lite.js`
+  for details.
+
 # v1.2.0 (2024-07-30)
 
 - Fixes incompatible behavior with Node.js package conditional exports #2276.
--- a/packages/compartment-mapper/src/import-hook.js
+++ b/packages/compartment-mapper/src/import-hook.js
@@ -7,20 +7,41 @@
  */
 
 // @ts-check
+/**
+ * @import {
+ *   ImportHook,
+ *   ImportNowHook,
+ *   RedirectStaticModuleInterface,
+ *   StaticModuleType
+ * } from 'ses'
+ * @import {
+ *   CompartmentDescriptor,
+ *   ChooseModuleDescriptorOperators,
+ *   ChooseModuleDescriptorOptions,
+ *   ChooseModuleDescriptorYieldables,
+ *   ExitModuleImportHook,
+ *   FindRedirectParams,
+ *   HashFn,
+ *   ImportHookMaker,
+ *   ImportNowHookMaker,
+ *   MakeImportNowHookMakerOptions,
+ *   ModuleDescriptor,
+ *   ParseResult,
+ *   ReadFn,
+ *   ReadPowers,
+ *   SourceMapHook,
+ *   Sources,
+ *   ReadNowPowers
+ * } from './types.js'
+ */
 
-/** @import {ImportHook} from 'ses' */
-/** @import {StaticModuleType} from 'ses' */
-/** @import {RedirectStaticModuleInterface} from 'ses' */
-/** @import {ReadFn} from './types.js' */
-/** @import {ReadPowers} from './types.js' */
-/** @import {HashFn} from './types.js' */
-/** @import {Sources} from './types.js' */
-/** @import {CompartmentDescriptor} from './types.js' */
-/** @import {ImportHookMaker} from './types.js' */
-/** @import {ExitModuleImportHook} from './types.js' */
-
-import { attenuateModuleHook, enforceModulePolicy } from './policy.js';
+import { asyncTrampoline, syncTrampoline } from '@endo/trampoline';
 import { resolve } from './node-module-specifier.js';
+import {
+  attenuateModuleHook,
+  ATTENUATORS_COMPARTMENT,
+  enforceModulePolicy,
+} from './policy.js';
 import { unpackReadPowers } from './powers.js';
 
 // q, as in quote, for quoting strings in error messages.
@@ -35,6 +56,8 @@ const { apply } = Reflect;
  * @type {import('ses').Harden}
  */
 const freeze = Object.freeze;
+
+const { entries, keys, assign, create } = Object;
 
 const { hasOwnProperty } = Object.prototype;
 /**
@@ -69,9 +92,96 @@ const nodejsConventionSearchSuffixes = [
 ];
 
 /**
+ * Given a module specifier which is an absolute path, attempt to match it with
+ * an existing compartment; return a {@link RedirectStaticModuleInterface} if found.
+ *
+ * @throws If we determine `absoluteModuleSpecifier` is unknown
+ * @param {FindRedirectParams} params Parameters
+ * @returns {RedirectStaticModuleInterface|undefined} A redirect or nothing
+ */
+const findRedirect = ({
+  compartmentDescriptor,
+  compartmentDescriptors,
+  compartments,
+  absoluteModuleSpecifier,
+  packageLocation,
+}) => {
+  const moduleSpecifierLocation = new URL(
+    absoluteModuleSpecifier,
+    packageLocation,
+  ).href;
+
+  // a file:// URL string
+  let someLocation = new URL('./', moduleSpecifierLocation).href;
+
+  // we are guaranteed an absolute path, so we can search "up" for the compartment
+  // due to the structure of `node_modules`
+
+  // n === count of path components to the fs root
+  for (;;) {
+    if (
+      someLocation !== ATTENUATORS_COMPARTMENT &&
+      someLocation in compartments
+    ) {
+      const location = someLocation;
+      const someCompartmentDescriptor = compartmentDescriptors[location];
+      if (compartmentDescriptor === someCompartmentDescriptor) {
+        // this compartmentDescriptor wants to dynamically load its own module
+        // using an absolute path
+        return undefined;
+      }
+
+      // this tests the compartment referred to by the absolute path
+      // is a dependency of the compartment descriptor
+      if (compartmentDescriptor.compartments.has(location)) {
+        return {
+          specifier: absoluteModuleSpecifier,
+          compartment: compartments[location],
+        };
+      }
+
+      // this tests if the compartment descriptor is a dependency of the
+      // compartment referred to by the absolute path.
+      // it may be in scope, but disallowed by policy.
+      if (
+        someCompartmentDescriptor.compartments.has(
+          compartmentDescriptor.location,
+        )
+      ) {
+        enforceModulePolicy(
+          compartmentDescriptor.name,
+          someCompartmentDescriptor,
+          {
+            errorHint: `Blocked in import hook. ${q(absoluteModuleSpecifier)} is part of the compartment map and resolves to ${location}`,
+          },
+        );
+        return {
+          specifier: absoluteModuleSpecifier,
+          compartment: compartments[location],
+        };
+      }
+
+      throw new Error(`Could not import module: ${q(absoluteModuleSpecifier)}`);
+    } else {
+      // go up a directory
+      const parentLocation = new URL('../', someLocation).href;
+
+      // afaict this behavior is consistent across both windows and posix
+      if (parentLocation === someLocation) {
+        throw new Error(
+          `Could not import unknown module: ${q(absoluteModuleSpecifier)}`,
+        );
+      }
+
+      someLocation = parentLocation;
+    }
+  }
+};
+
+/**
  * @param {object} params
  * @param {Record<string, any>=} params.modules
- * @param {ExitModuleImportHook=} params.exitModuleImportHook
+ * @param {ExitModuleImportHook} [params.exitModuleImportHook]
  * @returns {ExitModuleImportHook|undefined}
  */
 export const exitModuleImportHookMaker = ({
@@ -84,12 +194,12 @@ export const exitModuleImportHookMaker = ({
   return async specifier => {
     if (modules && has(modules, specifier)) {
       const ns = modules[specifier];
-      return Object.freeze({
+      return freeze({
         imports: [],
-        exports: ns ? Object.keys(ns) : [],
+        exports: ns ? keys(ns) : [],
         execute: moduleExports => {
           moduleExports.default = ns;
-          Object.assign(moduleExports, ns);
+          assign(moduleExports, ns);
         },
       });
     }
@@ -101,6 +211,185 @@ export const exitModuleImportHookMaker = ({
 };
 
 /**
+ * Expands a module specifier into a list of potential candidates based on
+ * `searchSuffixes`.
+ *
+ * @param {string} moduleSpecifier Module specifier
+ * @param {string[]} searchSuffixes Suffixes to search if the unmodified
+ * specifier is not found
+ * @returns {string[]} A list of potential candidates (including
+ * `moduleSpecifier` itself)
+ */
+const nominateCandidates = (moduleSpecifier, searchSuffixes) => {
+  // Collate candidate locations for the moduleSpecifier,
+  // to support Node.js conventions and similar.
+  const candidates = [moduleSpecifier];
+  for (const candidateSuffix of searchSuffixes) {
+    candidates.push(`${moduleSpecifier}${candidateSuffix}`);
+  }
+  return candidates;
+};
+
+/**
+ * Returns a generator which applies {@link ChooseModuleDescriptorOperators} in
+ * `operators` using the options in options to ultimately result in a
+ * {@link StaticModuleType} for a particular {@link CompartmentDescriptor} (or
+ * `undefined`).
+ *
+ * Supports both {@link SyncChooseModuleDescriptorOperators sync} and
+ * {@link AsyncChooseModuleDescriptorOperators async} operators.
+ *
+ * Used by both {@link makeImportNowHookMaker} and {@link makeImportHookMaker}.
+ *
+ * @template {ChooseModuleDescriptorOperators} Operators Type of operators (sync
+ * or async)
+ * @param {ChooseModuleDescriptorOptions} options Options/context
+ * @param {Operators} operators Operators
+ * @returns {Generator<ChooseModuleDescriptorYieldables,
+ * StaticModuleType|undefined, Awaited<ChooseModuleDescriptorYieldables>>}
+ * Generator
+ */
+function* chooseModuleDescriptor(
+  {
+    candidates,
+    compartmentDescriptor,
+    compartmentDescriptors,
+    compartments,
+    computeSha512,
+    moduleDescriptors,
+    moduleSpecifier,
+    packageLocation,
+    packageSources,
+    readPowers,
+    sourceMapHook,
+    strictlyRequiredForCompartment,
+  },
+  { maybeRead, parse, shouldDeferError = () => false },
+) {
+  for (const candidateSpecifier of candidates) {
+    const candidateModuleDescriptor = moduleDescriptors[candidateSpecifier];
+    if (candidateModuleDescriptor !== undefined) {
+      const { compartment: candidateCompartmentName = packageLocation } =
+        candidateModuleDescriptor;
+      const candidateCompartment = compartments[candidateCompartmentName];
+      if (candidateCompartment === undefined) {
+        throw Error(
+          `compartment missing for candidate ${candidateSpecifier} in ${candidateCompartmentName}`,
+        );
+      }
+      // modify compartmentMap to include this redirect
+      const candidateCompartmentDescriptor =
+        compartmentDescriptors[candidateCompartmentName];
+      if (candidateCompartmentDescriptor === undefined) {
+        throw Error(
+          `compartmentDescriptor missing for candidate ${candidateSpecifier} in ${candidateCompartmentName}`,
+        );
+      }
+      candidateCompartmentDescriptor.modules[moduleSpecifier] =
+        candidateModuleDescriptor;
+      // return a redirect
+      /** @type {RedirectStaticModuleInterface} */
+      const record = {
+        specifier: candidateSpecifier,
+        compartment: candidateCompartment,
+      };
+      return record;
+    }
+
+    // Using a specifier as a location.
+    // This is not always valid.
+    // But, for Node.js, when the specifier is relative and not a directory
+    // name, they are usable as URL's.
+    const moduleLocation = resolveLocation(candidateSpecifier, packageLocation);
+
+    // "next" values must have type assertions for narrowing because we have
+    // multiple yielded types
+    const moduleBytes = /** @type {Uint8Array|undefined} */ (
+      yield maybeRead(moduleLocation)
+    );
+
+    if (moduleBytes !== undefined) {
+      /** @type {string | undefined} */
+      let sourceMap;
+      // must be narrowed
+      const envelope = /** @type {ParseResult} */ (
+        yield parse(
+          moduleBytes,
+          candidateSpecifier,
+          moduleLocation,
+          packageLocation,
+          {
+            readPowers,
+            sourceMapHook:
+              sourceMapHook &&
+              (nextSourceMapObject => {
+                sourceMap = JSON.stringify(nextSourceMapObject);
+              }),
+            compartmentDescriptor,
+          },
+        )
+      );
+      const {
+        parser,
+        bytes: transformedBytes,
+        record: concreteRecord,
+      } = envelope;
+
+      // Facilitate a redirect if the returned record has a different
+      // module specifier than the requested one.
+      if (candidateSpecifier !== moduleSpecifier) {
+        moduleDescriptors[moduleSpecifier] = {
+          module: candidateSpecifier,
+          compartment: packageLocation,
+        };
+      }
+      /** @type {StaticModuleType} */
+      const record = {
+        record: concreteRecord,
+        specifier: candidateSpecifier,
+        importMeta: { url: moduleLocation },
+      };
+
+      let sha512;
+      if (computeSha512 !== undefined) {
+        sha512 = computeSha512(transformedBytes);
+
+        if (sourceMapHook !== undefined && sourceMap !== undefined) {
+          sourceMapHook(sourceMap, {
+            compartment: packageLocation,
+            module: candidateSpecifier,
+            location: moduleLocation,
+            sha512,
+          });
+        }
+      }
+
+      const packageRelativeLocation = moduleLocation.slice(
+        packageLocation.length,
+      );
+      packageSources[candidateSpecifier] = {
+        location: packageRelativeLocation,
+        sourceLocation: moduleLocation,
+        parser,
+        bytes: transformedBytes,
+        record: concreteRecord,
+        sha512,
+      };
+      if (!shouldDeferError(parser)) {
+        for (const importSpecifier of getImportsFromRecord(record)) {
+          strictlyRequiredForCompartment(packageLocation).add(
+            resolve(importSpecifier, moduleSpecifier),
+          );
+        }
+      }
+
+      return record;
+    }
+  }
+  return undefined;
+}
+
+/**
  * @param {ReadFn|ReadPowers} readPowers
  * @param {string} baseLocation
  * @param {object} options
@@ -110,23 +399,23 @@ export const exitModuleImportHookMaker = ({
  * @param {HashFn} [options.computeSha512]
  * @param {Array<string>} [options.searchSuffixes] - Suffixes to search if the
  * unmodified specifier is not found.
- * Pass [] to emulate Node.js’s strict behavior.
- * The default handles Node.js’s CommonJS behavior.
+ * Pass [] to emulate Node.js' strict behavior.
+ * The default handles Node.js' CommonJS behavior.
  * Unlike Node.js, the Compartment Mapper lifts CommonJS up, more like a
  * bundler, and does not attempt to vary the behavior of resolution depending
  * on the language of the importing module.
  * @param {string} options.entryCompartmentName
  * @param {string} options.entryModuleSpecifier
  * @param {ExitModuleImportHook} [options.exitModuleImportHook]
- * @param {import('./types.js').SourceMapHook} [options.sourceMapHook]
+ * @param {SourceMapHook} [options.sourceMapHook]
  * @returns {ImportHookMaker}
  */
 export const makeImportHookMaker = (
   readPowers,
   baseLocation,
   {
-    sources = Object.create(null),
-    compartmentDescriptors = Object.create(null),
+    sources = create(null),
+    compartmentDescriptors = create(null),
     archiveOnly = false,
     computeSha512 = undefined,
     searchSuffixes = nodejsConventionSearchSuffixes,
@@ -168,11 +457,10 @@ export const makeImportHookMaker = (
   }) => {
     // per-compartment:
     packageLocation = resolveLocation(packageLocation, baseLocation);
-    const packageSources = sources[packageLocation] || Object.create(null);
+    const packageSources = sources[packageLocation] || create(null);
     sources[packageLocation] = packageSources;
     const compartmentDescriptor = compartmentDescriptors[packageLocation] || {};
-    const { modules: moduleDescriptors = Object.create(null) } =
-      compartmentDescriptor;
+    const { modules: moduleDescriptors = create(null) } = compartmentDescriptor;
     compartmentDescriptor.modules = moduleDescriptors;
 
     /**
@@ -209,8 +497,10 @@ export const makeImportHookMaker = (
 
     /** @type {ImportHook} */
     const importHook = async moduleSpecifier => {
-      await null;
       compartmentDescriptor.retained = true;
+
+      // for lint rule
+      await null;
 
       // per-module:
 
@@ -257,130 +547,31 @@ export const makeImportHookMaker = (
         );
       }
 
-      // Collate candidate locations for the moduleSpecifier,
-      // to support Node.js conventions and similar.
-      const candidates = [moduleSpecifier];
-      for (const candidateSuffix of searchSuffixes) {
-        candidates.push(`${moduleSpecifier}${candidateSuffix}`);
-      }
-
       const { maybeRead } = unpackReadPowers(readPowers);
 
-      for (const candidateSpecifier of candidates) {
-        const candidateModuleDescriptor = moduleDescriptors[candidateSpecifier];
-        if (candidateModuleDescriptor !== undefined) {
-          const { compartment: candidateCompartmentName = packageLocation } =
-            candidateModuleDescriptor;
-          const candidateCompartment = compartments[candidateCompartmentName];
-          if (candidateCompartment === undefined) {
-            throw Error(
-              `compartment missing for candidate ${candidateSpecifier} in ${candidateCompartmentName}`,
-            );
-          }
-          // modify compartmentMap to include this redirect
-          const candidateCompartmentDescriptor =
-            compartmentDescriptors[candidateCompartmentName];
-          if (candidateCompartmentDescriptor === undefined) {
-            throw Error(
-              `compartmentDescriptor missing for candidate ${candidateSpecifier} in ${candidateCompartmentName}`,
-            );
-          }
-          candidateCompartmentDescriptor.modules[moduleSpecifier] =
-            candidateModuleDescriptor;
-          // return a redirect
-          /** @type {RedirectStaticModuleInterface} */
-          const record = {
-            specifier: candidateSpecifier,
-            compartment: candidateCompartment,
-          };
-          return record;
-        }
+      const candidates = nominateCandidates(moduleSpecifier, searchSuffixes);
 
-        // Using a specifier as a location.
-        // This is not always valid.
-        // But, for Node.js, when the specifier is relative and not a directory
-        // name, they are usable as URL's.
-        const moduleLocation = resolveLocation(
-          candidateSpecifier,
+      const record = await asyncTrampoline(
+        chooseModuleDescriptor,
+        {
+          candidates,
+          compartmentDescriptor,
+          compartmentDescriptors,
+          compartments,
+          computeSha512,
+          moduleDescriptors,
+          moduleSpecifier,
           packageLocation,
-        );
-        // eslint-disable-next-line no-await-in-loop
-        const moduleBytes = await maybeRead(moduleLocation);
-        if (moduleBytes !== undefined) {
-          /** @type {string | undefined} */
-          let sourceMap;
-          // eslint-disable-next-line no-await-in-loop
-          const envelope = await parse(
-            moduleBytes,
-            candidateSpecifier,
-            moduleLocation,
-            packageLocation,
-            {
-              readPowers,
-              sourceMapHook:
-                sourceMapHook &&
-                (nextSourceMapObject => {
-                  sourceMap = JSON.stringify(nextSourceMapObject);
-                }),
-              compartmentDescriptor,
-            },
-          );
-          const {
-            parser,
-            bytes: transformedBytes,
-            record: concreteRecord,
-          } = envelope;
+          packageSources,
+          readPowers,
+          sourceMapHook,
+          strictlyRequiredForCompartment,
+        },
+        { maybeRead, parse, shouldDeferError },
+      );
 
-          // Facilitate a redirect if the returned record has a different
-          // module specifier than the requested one.
-          if (candidateSpecifier !== moduleSpecifier) {
-            moduleDescriptors[moduleSpecifier] = {
-              module: candidateSpecifier,
-              compartment: packageLocation,
-            };
-          }
-          /** @type {StaticModuleType} */
-          const record = {
-            record: concreteRecord,
-            specifier: candidateSpecifier,
-            importMeta: { url: moduleLocation },
-          };
-
-          let sha512;
-          if (computeSha512 !== undefined) {
-            sha512 = computeSha512(transformedBytes);
-
-            if (sourceMapHook !== undefined && sourceMap !== undefined) {
-              sourceMapHook(sourceMap, {
-                compartment: packageLocation,
-                module: candidateSpecifier,
-                location: moduleLocation,
-                sha512,
-              });
-            }
-          }
-
-          const packageRelativeLocation = moduleLocation.slice(
-            packageLocation.length,
-          );
-          packageSources[candidateSpecifier] = {
-            location: packageRelativeLocation,
-            sourceLocation: moduleLocation,
-            parser,
-            bytes: transformedBytes,
-            record: concreteRecord,
-            sha512,
-          };
-          if (!shouldDeferError(parser)) {
-            for (const importSpecifier of getImportsFromRecord(record)) {
-              strictlyRequiredForCompartment(packageLocation).add(
-                resolve(importSpecifier, moduleSpecifier),
-              );
-            }
-          }
-
-          return record;
-        }
+      if (record) {
+        return record;
       }
 
       return deferError(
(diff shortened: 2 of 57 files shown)


boneskull:
If anyone can point me in the direction of why the tests are failing, that'd be helpful; otherwise I'll just plug at it.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -398,3 +405,256 @@ export const makeImportHookMaker = (
...
+
+    packageLocation = resolveLocation(packageLocation, baseLocation);
+    const packageSources = sources[packageLocation] || Object.create(null);
+    sources[packageLocation] = packageSources;
+    const { modules: moduleDescriptors = Object.create(null) } =
+      compartmentDescriptor;
+    compartmentDescriptor.modules = moduleDescriptors;
+
+    let { policy } = compartmentDescriptor;
+    policy = policy || Object.create(null);
```
boneskull, line 479: this is for typescript

packages/compartment-mapper/src/import-hook.js
```diff
@@ -398,3 +405,256 @@ export const makeImportHookMaker = (
...
+        if (
+          !(pkgName in compartmentDescriptor.modules) &&
+          pkgName in compartmentDescriptor.scopes &&
+          policyItem
+        ) {
+          compartmentDescriptor.modules[pkgName] =
+            compartmentDescriptor.scopes[pkgName];
+        }
+      }
+    }
```
boneskull, line 494: I am unsure if this is correct, but it creates links between compartment descriptors based only on policy--where Endo would not have detected them otherwise.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -398,3 +405,256 @@ export const makeImportHookMaker = (
...
+        }
+
+        // Using a specifier as a location.
+        // This is not always valid.
+        // But, for Node.js, when the specifier is relative and not a directory
+        // name, they are usable as URL's.
+        const moduleLocation = resolveLocation(
+          candidateSpecifier,
+          packageLocation,
+        );
```
boneskull, line 551: copypasta

packages/compartment-mapper/src/import-hook.js
```diff
@@ -398,3 +405,256 @@ export const makeImportHookMaker = (
...
+        try {
+          moduleBytes = readSync(moduleLocation);
+        } catch (err) {
+          if (err && err.code === 'ENOENT') {
+            // might be an exit module. use the fallback `exitModuleImportNowHook` to import it
+            // eslint-disable-next-line no-continue
+            continue;
+          }
+          throw err;
+        }
```
boneskull, line 564: how do we know something is an exit module?  maybe we should check the compartment descriptor, and only use the fallback if the thing doesn't exist there?

packages/compartment-mapper/src/import-hook.js
```diff
@@ -398,3 +405,256 @@ export const makeImportHookMaker = (
...
+          };
+          for (const importSpecifier of getImportsFromRecord(record)) {
+            strictlyRequiredForCompartment(packageLocation).add(
+              resolve(importSpecifier, moduleSpecifier),
+            );
+          }
+
+          return record;
+        }
+      }
```
boneskull, line 638: copypasta

packages/compartment-mapper/src/link.js
```diff
@@ -175,6 +262,41 @@ export const mapParsers = (
...
 
+/**
+ * @param {Record<string, Language>} languageForExtension
+ * @param {Record<string, string>} languageForModuleSpecifier - In a rare case, the type of a module
+ * is implied by package.json and should not be inferred from its extension.
+ * @param {Record<string, ParserImplementation>} parserForLanguage
+ * @param {SyncModuleTransforms} moduleTransforms
+ * @returns {ParseFn}
+ */
+export const mapParsersSync = (
```
boneskull, line 273: should probably just change mapParsers to handle both `SyncModuleTransforms` and `ModuleTransforms`...?

packages/compartment-mapper/src/link.js
```diff
@@ -434,6 +648,34 @@ export const link = (
...
+          compartments,
+        });
+      } else {
+        // it's possible for this to still be called, but
+        // the error message would be cryptic otherwise.
+        importNowHook = () => {
+          throw new Error(
+            `Dynamic require not allowed in compartment ${q(compartmentDescriptor.name)}`,
+          );
+        };
```
boneskull, line 675: I'm not sure what else to do here.

Looking for the `dynamic` flag too early causes other tests to break, and would potentially need `dynamic: true` in the policy for attenuators (see [here](https://github.com/endojs/endo/blob/8f8f5897112520ae9ae48c83b1929c575183dc01/packages/compartment-mapper/src/policy.js#L164-L166))

packages/compartment-mapper/src/types.js
```diff
@@ -558,10 +704,40 @@ export {};
...
+/**
+ * @typedef UrlAPI
+ * @property {(location: string | URL) => string} fileURLToPath
+ * @property {(path: string) => URL} pathToFileURL
+ */
+
+/**
+ * @typedef CryptoAPI
+ * @property {typeof import('crypto').createHash} createHash
+ */
```
boneskull, line 729: not strictly necessary, but I found it helpful. YMMV

packages/compartment-mapper/src/import-hook.js
```diff
@@ -36,6 +41,8 @@ const { apply } = Reflect;
  */
 const freeze = Object.freeze;
 
+const entries = Object.entries;
```
kriskowal, line 44:
```suggestion
const { entries, freeze } = Object;
```

packages/compartment-mapper/src/import-hook.js
kriskowal, line 479: Does TypeScript not like this pattern:?
```suggestion
    const { policy = create(null) } = compartmentDescriptor;
```

kriskowal: Ah, thought if we’re coming off of JSON and using the `in` operator, it might be better to ensure that the policy has a null proto, something like:
```js
const policy = { __proto__: null, ...(compartment.descriptor.policy || {}) };
```
Or:
```js
const policy = assign(create(null), compartment.descriptor.policy || {});
```

packages/compartment-mapper/src/import-hook.js
```diff
@@ -398,3 +405,256 @@ export const makeImportHookMaker = (
...
+      compartmentDescriptor;
+    compartmentDescriptor.modules = moduleDescriptors;
+
+    let { policy } = compartmentDescriptor;
+    policy = policy || Object.create(null);
+
+    // associates modules with compartment descriptors based on policy
+    // which wouldn't otherwise be there
+    if ('packages' in policy && typeof policy.packages === 'object') {
+      for (const [pkgName, policyItem] of entries(policy.packages)) {
```
kriskowal, line 484: Style nit: I’ve generally elsewhere avoided abbreviating `package` to `pkg`. `policyItem` should probably be something like `compartmentPolicy` or `packakgePolicy` to increase the specificity of “item” to the scope it covers.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -398,3 +405,256 @@ export const makeImportHookMaker = (
...
+          candidateSpecifier,
+          packageLocation,
+        );
+        // eslint-disable-next-line no-await-in-loop
+        /** @type {Uint8Array} */
+        let moduleBytes;
+        try {
+          moduleBytes = readSync(moduleLocation);
+        } catch (err) {
+          if (err && err.code === 'ENOENT') {
```
kriskowal, line 558: This is coupled too closely to Node.js. The coupling should only exist in the Node.js powers. We have an analogous `maybeRead` that returns undefined if the file is not found. We will need a `maybeReadSync` power.

packages/compartment-mapper/src/import-hook.js
kriskowal: This is a lot of duplication. I think this is worth trying to bounce on a trampoline.

packages/compartment-mapper/src/import-hook.js
kriskowal: I have not arrived at a hard decision for this question. Falling through the bottom of the import hook should be enough, but might not be. An option we did not have when I first wrote this: we can identify “host modules” by the presence of a `prefix:` like `node:` or `endo:`. These would be guaranteed to escape the Node.js style mappings.

In the interest of keeping coupling to a specific specifier resolution strategy low, I am leaning heavily toward “pass through if nothing in the compartment map matches.”

packages/compartment-mapper/src/import-lite.js
```diff
@@ -55,6 +87,44 @@ export const loadFromMap = async (readPowers, compartmentMap, options = {}) => {
     assign(create(null), languageForExtensionOption),
   );
 
+  /**
+   * Object containing options and read powers which fulfills all requirements
```
kriskowal, line 91:
```suggestion
   * Object containing options and read powers that fulfills all requirements
```

packages/compartment-mapper/src/import-lite.js
```diff
@@ -55,6 +87,44 @@ export const loadFromMap = async (readPowers, compartmentMap, options = {}) => {
...
+   * for creation of a {@link ImportNowHookMaker}, thus enabling dynamic import
```
kriskowal, line 92:
```suggestion
   * for creation of a {@link ImportNowHookMaker}, thus enabling dynamic import.
```

packages/compartment-mapper/src/import-lite.js
```diff
@@ -55,6 +87,44 @@ export const loadFromMap = async (readPowers, compartmentMap, options = {}) => {
...
+   *
+   * @typedef SyncBehavior
+   * @property {SyncReadPowers} readPowers
+   * @property {SyncImportLocationOptions} options
+   * @property {'SYNC'} type
+   */
+
+  /**
+   * Object containing options and read powers which is incompatible with
+   * creation of a {@link ImportNowHookMaker}, thus disabling dynamic import
```
kriskowal, line 102:
```suggestion
   * creation of an {@link ImportNowHookMaker}, thus disabling dynamic import.
```

packages/evasive-transform/src/location-unmapper.js
```diff
@@ -4,7 +4,7 @@
  * @module
  */
 
-import { SourceMapConsumer } from 'source-map';
+import { SourceMapConsumer } from 'source-map-js';
```
kriskowal, line 7: I’d be happy to review this change to evasive-transform separately to make this one smaller and  land it faster.

packages/compartment-mapper/src/import-lite.js
```diff
@@ -85,16 +155,56 @@ export const loadFromMap = async (readPowers, compartmentMap, options = {}) => {
...
+        makeImportHook,
+        parserForLanguage,
+        languageForExtension,
+        globals,
+        transforms,
+        moduleTransforms,
+        __shimTransforms__,
+        Compartment,
+      }));
+    }
```
kriskowal, line 207: I need to think more about what it would take to make this work with a single call to `link`, deferring the problem of differentiating these cases to the `link` implementation. That might be by passing a `sync: true` option into `link`.

Maybe this would give us more options: SES currently accepts both `importHook` and `importNowHook` but only consults one or the other. We could change `Compartment#import` to consult `importNowHook` before `importHook`, proceeding to the later only if the former returns `undefined`. Maybe that would let us get more bang from `importNowHook`. Or perhaps we just need that trampoline.

kriskowal (commented):
Preliminary feedback.

packages/compartment-mapper/src/import-hook.js
boneskull: Yeah.  This is a good idea, but is also unfortunate because it will take me longer to land this.

packages/compartment-mapper/src/import-hook.js
boneskull: To answer your first question: no; it is inferred to be `{}` instead of `CompartmentDescriptor`.

I'll take a few putts at it, but if I bogey I'm just going to leave it as-is.

boneskull:
Going to extract the changes to `@endo/evasive-transform` into a separate PR.

boneskull:
Ref: https://github.com/endojs/endo/pull/2332

cc @kriskowal

boneskull:
Once #2332 is merged, I can rebase this onto `master`, which should eliminate the commit from this PR's history.  So let's sit on it until then.

packages/compartment-mapper/src/import-hook.js
naugtur: less code and better compatibility. Sounds good.
Also, explicit opt-in would allow an early failure if ReadPowers or transforms don't match the intent to enable dynamic require.

packages/compartment-mapper/src/import-hook.js
naugtur: The prefix option would be really unfortunate for ecosystem compatibility. I don't think we'll ever get all packages to switch form `fs` to `node:fs`.

For the sake of completion: if importHook for exit modules is not provided or returns nothing we still need to fail with an import error.

naugtur:
note on `source-map-js`

I don't see where the source-map-js is added. The documentation says it's a fork of source-map but with some optimizations added. One of the optimizations listed in the article they cite as source of the optimizations is using a `Function` constructor to create variants of sorting function. That's a potential concern (less of a security concern, more of a compatibility concern if Endo code is supposed to run correctly under lockdown)

packages/compartment-mapper/src/import-lite.js
naugtur: TL;DR: link was synchronous and didn't care about all the async work happening around much, how about we keep it that way.

I also have an intuition that link could work differently, but my idea for where to take it is the opposite - make link totally colorblind and have one implementation of link that is fed pre-populated constructs/functions to use. 

eg. mapParsers choice depending on isSync and passing corresponding transforms could be prepared earlier and passed in as one mapParsers implementation. 

In that case the if on behavior would be used to collect the right type of tools and we'd only get one link function. 

Link doesn't do anything asynchronous itself, so threading a trampoline through link would be an unnecessary detour - the pre-cooking of some utilities it uses would be the opportunity to trampoline some of the implementations if necessary but the link.js file shouldn't see any of it.

The only new behavior in link as of today that needs to understand `isSync` is the decision around entry module, but that could also be solved with dependency injection - `conditionalImportNowHook<VerbHere>er` would accept all ingredients for importNowHookMaker and decide whether to use it or provide one that throws an informative error. (could be added on to importNowHookMaker to skip using the new name I just half-made up)

I'm very self conscious about using lots of dependency injection vs just importing functionality as it requires jumping back and forth between files while preserving context to figure out which code actually runs, but here DI might actually concentrate context-dependent behavior more by removing that concern from link entirely and having a dense code block that makes all behavior decisions in one place. (note this is all trampoline-agnostic and the implementations of behaviors can be shared&bounced or not)

<details>
  <summary>alternatively</summary>
100% separate sync/async implementations exported from different modules sharing some of the internals and *C macros* in imports :see_no_evil: 
</details>

packages/compartment-mapper/src/link.js
```diff
@@ -411,13 +617,19 @@ export const link = (
...
     );
-
-    const parse = mapParsers(
-      languageForExtension,
-      languageForModuleSpecifier,
-      parserForLanguage,
-      moduleTransforms,
-    );
+    const parse = isSync
+      ? mapParsersSync(
```
naugtur, line 621: None of this concerns link really. (see my longer post in response to Kris)
```
const mapParsers = makeMapParsers[behavior]({
  syncModuleTransforms,
  moduleTransforms
})

link( ... mapParsers ... )
```

packages/compartment-mapper/src/node-powers.js
```diff
@@ -137,6 +165,53 @@ const makeReadPowersSloppy = ({ fs, url = undefined, crypto = undefined }) => {
...
+   * @type {MaybeReadSyncFn}
+   */
+  const maybeReadSync = location => {
+    const filepath = fileURLToPath(location);
+    try {
+      return fs.readFileSync(filepath);
+    } catch (error) {
+      if (
+        'code' in error &&
+        (error.code === 'ENOENT' || error.code === 'EISDIR')
```
naugtur, line 202: premature optimization thought: `EISDIR` is a hint to look at location/index.js and skip other candidates, but we have no good way to pass that information up to the loop that's iterating candidates.
I'm pretty sure it wouldn't be worth the complexity.

_will delete later_

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
```diff
@@ -141,13 +145,26 @@ export const wrap = ({
...
     if (!has(resolvedImports, importSpecifier)) {
-      throw new Error(
-        `Cannot find module "${importSpecifier}" in "${location}"`,
-      );
+      if (isSyncReadPowers(readPowers)) {
+        namespace = compartment.importNow(importSpecifier);
+      } else {
+        const invalidProps = findInvalidSyncReadPowersProps(readPowers).sort();
+        throw new Error(
+          `Synchronous readPowers required for dynamic import of ${assert.quote(importSpecifier)}; missing or invalid prop(s): ${invalidProps.join(', ')}`,
```
naugtur, line 161: nice

packages/compartment-mapper/src/import-hook.js
```diff
@@ -399,3 +411,292 @@ export const makeImportHookMaker = (
...
+  }) => {
+    const compartmentDescriptor = compartmentDescriptors[packageLocation] || {};
+
+    // this is necessary, because there's nothing that prevents someone from calling this function with the entry compartment.
+    assert(
+      compartmentDescriptor !== entryCompartmentDescriptor,
+      'Cannot create an ImportNowHook for the entry compartment',
+    );
+
+    // this is not strictly necessary, since the types should handle it.
```
naugtur, line 471: It's cool, I vote we keep it.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -8,18 +8,27 @@
...
 /** @import {CompartmentDescriptor} from './types.js' */
 /** @import {ImportHookMaker} from './types.js' */
 /** @import {ExitModuleImportHook} from './types.js' */
+/** @import {SourceMapHook} from './types.js' */
+/** @import {ImportNowHookMaker} from './types.js' */
 
-import { attenuateModuleHook, enforceModulePolicy } from './policy.js';
+import {
+  attenuateModuleHook,
+  ATTENUATORS_COMPARTMENT,
```
naugtur, line 29: I was able to contain  `ATTENUATORS_COMPARTMENT` reference to policy for the most part and from my experience refactoring my policy wiring in the initial PR many times, striving to hide it leads to good outcomes.

It'd be great if import hook didn't need to understand what  attenuators compartment is and the compartment "just worked" with the existing implementation. I'm for changing how attenuators are loaded if it's tripping up the import hook implementation, but it could also be an indicator that importNowHook relies on something it shouldn't.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -36,6 +45,8 @@ const { apply } = Reflect;
  */
 const freeze = Object.freeze;
 
+const { entries, keys, assign, create } = Object;
```
naugtur, line 48:
```suggestion
const { freeze, entries, keys, assign, create } = Object;
```

packages/compartment-mapper/src/import-hook.js
```diff
@@ -110,23 +121,23 @@ export const exitModuleImportHookMaker = ({
  * @param {HashFn} [options.computeSha512]
  * @param {Array<string>} [options.searchSuffixes] - Suffixes to search if the
  * unmodified specifier is not found.
- * Pass [] to emulate Node.js’s strict behavior.
- * The default handles Node.js’s CommonJS behavior.
+ * Pass [] to emulate Node.js' strict behavior.
+ * The default handles Node.js' CommonJS behavior.
```
naugtur, line 125: :heart_eyes:

packages/compartment-mapper/src/import-hook.js
```diff
@@ -212,6 +222,9 @@ export const makeImportHookMaker = (
       await null;
       compartmentDescriptor.retained = true;
 
+      // for lint rule
```
naugtur, line 225: seems odd, there's one above already. is the rule malfunctioning? 

(why is it mal- vs dis- with function?)

packages/compartment-mapper/src/import-hook.js
```diff
@@ -399,3 +411,292 @@ export const makeImportHookMaker = (
...
+    }
+
+    const { maybeReadSync, isAbsolute } = readPowers;
+
+    /** @type {ImportNowHook} */
+    const importNowHook = moduleSpecifier => {
+      if (isAbsolute(moduleSpecifier)) {
+        let shouldCallExitModuleImportNowHook = true;
+        for (const [someDescriptorName, someDescriptor] of entries(
+          compartmentDescriptors,
```
naugtur, line 513: not all of these must make sense, I'm "thinking aloud": 

1. `compartmentDescriptors` are not changing between calls to importNowHook, so most of this work could be hoisted out to the topmost maker function and only a lookup in a reindexed structure would be needed. On the other hand, the memory needed for that would be wasted most of the time as it's a very rare situation

2. I'd consider extracting it to a separate function `findCompartmentContainingAbsolutePath`

3. since this is not causing a module to be loaded, only redirecting - the further steps are going through a policy check anyway so skipping attenuators compartment in the lookup should be unnecessary as if any attenuators exist, a policy exists as well to prevent their code from being loaded by the compartment. 

4. I'm not sure if this makes sense, but could policy be used to narrow down the search here? Take only the packages allowed by policy and look in their descriptors only.

packages/compartment-mapper/src/import-hook.js
naugtur: I think we can bounce most of it out https://github.com/endojs/endo/commit/afa4e6354aa79de93307fbfd3eb8ea443a9e6e7f

packages/compartment-mapper/src/policy.js
```diff
@@ -20,18 +29,24 @@ export const ATTENUATORS_COMPARTMENT = '<ATTENUATORS>';
...
- * @param {object} to
- * @param {Array<string | symbol>} [list]
- * @returns {object}
+ * @template {Record<PropertyKey, any>} T
+ * @template {Record<PropertyKey, any>} U
+ * @template {Array<Partial<keyof T>>} [K=Array<keyof T>]
+ * @param {T} from
+ * @param {U} to
+ * @param {K} [list]
+ * @returns {Omit<U, K[number]> & Pick<T, K[number]>}
```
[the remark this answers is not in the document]
boneskull: It's just that it's ...not really necessary, as used. We can only do so much with types here, as TS doesn't understand enumerable properties. Still, it's more precise than before.

That said, given we know the types of `from` and `to` insofar as we have declared them (even if `object`) when calling `selectiveCopy`, I don't see how attenuation could change the type signature. The function does what it does, right? The onus is on the caller to provide correct types. So if you call this with `object`-typed parameters, you get an `object`-typed return value; no different than before.

packages/compartment-mapper/src/import-hook.js
boneskull: this might want a new issue

packages/compartment-mapper/src/import-hook.js
boneskull: `freeze` is separate due to its comment

packages/compartment-mapper/src/import-hook.js
boneskull: this is for the lint rule that says every `async` function, if it has an `await`, must have that `await` at the top level.  any `await` in the _closure_ does not satisfy the rule.

boneskull:
> note on `source-map-js`
> 
> I don't see where the source-map-js is added. The documentation says it's a fork of source-map but with some optimizations added. One of the optimizations listed in the article they cite as source of the optimizations is using a `Function` constructor to create variants of sorting function. That's a potential concern (less of a security concern, more of a compatibility concern if Endo code is supposed to run correctly under lockdown)

`source-map-js` was added in #2332

packages/compartment-mapper/src/node-powers.js
boneskull: I was just adapting the logic in `maybeRead`.  I don't know the reason for checking `EISDIR`; I wouldn't otherwise.

¯\\_(ツ)_/¯

packages/compartment-mapper/src/import-hook.js
```diff
@@ -399,3 +497,200 @@ export const makeImportHookMaker = (
...
+        if (
+          !(packageName in compartmentDescriptor.modules) &&
+          packageName in compartmentDescriptor.scopes &&
+          packagePolicyItem
+        ) {
+          compartmentDescriptor.modules[packageName] =
+            compartmentDescriptor.scopes[packageName];
+        }
+      }
+    }
```
naugtur, line 590: Feels like this would be better off as a function in policy.js called something like `addMissingRelationshipsToCompartmentModules` or something as meaningful but shorter.

packages/compartment-mapper/src/import-hook.js
naugtur: Please create an issue and link to it from a code comment here. It's an optimization we'll definitely want.

packages/compartment-mapper/src/link.js
```diff
@@ -328,37 +210,110 @@ const makeModuleMapHook = (
...
     transforms = [],
-    moduleTransforms = {},
     __shimTransforms__ = [],
     archiveOnly = false,
     Compartment = defaultCompartment,
-  },
-) => {
+  } = options;
+
+  const isSync = isSyncOptions(options);
```
naugtur, line 300: Ideally `isSync` should not be needed in link. We could merge as-is but I want to refactor it before I forget.

packages/compartment-mapper/src/map-parser.js
```diff
@@ -0,0 +1,294 @@
...
+    );
+  };
+
+  // if we have nothing in the moduleTransforms object, then we can use the syncParser.
+  if (keys(moduleTransforms).length === 0) {
+    transforms = syncModuleTransforms;
+    return syncParser;
+  }
+
+  transforms = moduleTransforms;
```
naugtur, line 233: AFAIR previeous iteration of this code was using both `[...moduleTransforms, ...syncModuleTransforms]` for the async case. Is this change intentional?

packages/compartment-mapper/src/map-parser.js
```diff
@@ -0,0 +1,294 @@
...
+    /** @type {string|undefined} */
+    let sourceMap;
+
+    if (has(transforms, language)) {
+      try {
+        ({
+          bytes,
+          parser: language,
+          sourceMap,
+        } = yield transforms[language](
```
naugtur, line 152: My first impression was trampolines might be too big of a gun for a function with just one await, but I don't see how to do it better.

packages/compartment-mapper/src/node-powers.js
naugtur: the logic is - if this fails for any other reason, it's a failure that we don't want to silence. These are the two errors that are expected in the process.

packages/compartment-mapper/src/import-hook.js
boneskull: yeah, I agree

packages/compartment-mapper/src/link.js
boneskull: yes, let me take another look at that

packages/compartment-mapper/src/map-parser.js
boneskull: it happens in `link`, here: https://github.com/endojs/endo/blob/476069c6df51edf2cfd089ed03e81403bf94751f/packages/compartment-mapper/src/link.js#L301-L310

packages/compartment-mapper/src/import-hook.js
boneskull: though I don't know why it'd be in `policy.js`. I'm leaving it here for now

packages/compartment-mapper/src/link.js
boneskull: the logic has to go somewhere. I got rid of the variable, though.

packages/compartment-mapper/src/import-hook.js
boneskull: please see changes from today

packages/compartment-mapper/src/import-hook.js
```diff
@@ -68,10 +91,67 @@ const nodejsConventionSearchSuffixes = [
...
+  compartments,
+  absoluteModuleSpecifier,
+  packageLocation,
+}) => {
+  let specifierIsArbitrary = true;
+  for (const [someDescriptorName, someDescriptor] of entries(
+    compartmentDescriptors,
+  )) {
+    if (someDescriptorName !== ATTENUATORS_COMPARTMENT) {
+      const moduleSpecifierUrl = `${new URL(absoluteModuleSpecifier, packageLocation)}`;
```
naugtur, line 114: (nit) we've had a convention introduced to use `new URL().href` AFAIR

packages/compartment-mapper/src/import-hook.js
```diff
@@ -68,10 +91,67 @@ const nodejsConventionSearchSuffixes = [
...
+          // can use that compartment
+          if (someDescriptor.name in compartmentDescriptor.modules) {
+            return {
+              specifier: absoluteModuleSpecifier,
+              compartment: compartments[someDescriptorName],
+            };
+          }
+          // compartmentDescriptor is a dependent of someSomeDescriptor;
+          // this is more dubious but a common pattern
+          if (compartmentDescriptor.name in someDescriptor.modules) {
```
naugtur, line 127: I don't understand why this is a case we want to handle.

If the absolute path import matches a compartment (someDescriptor) it makes sense to check if curernt compartment descriptor has that in the collection of modules, but why would we check the other way around?

naugtur: Maybe the thought process above would be easier to do if we added a comment explaining why the checks are there.

Also, the ifs can be merged.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -399,3 +560,158 @@ export const makeImportHookMaker = (
...
+          create(null)
+        ),
+    } = compartmentDescriptor;
+    compartmentDescriptor.modules = moduleDescriptors;
+
+    let { policy } = compartmentDescriptor;
+    policy = policy || create(null);
+
+    // associates modules with compartment descriptors based on policy
+    // which wouldn't otherwise be there
```
naugtur, line 634:
```suggestion
    // Associates modules with compartment descriptors based on policy
    // in cases where the association was not made when building the
    // compartment map but is indicated by the policy.
```

naugtur (approved):

packages/compartment-mapper/src/map-parser.js
```diff
@@ -0,0 +1,302 @@
...
+      packageLocation,
+      options,
+    );
+  };
+
+  // if we have nothing in the moduleTransforms object, then we can use the syncParser.
+  if (keys(moduleTransforms).length === 0) {
+    transforms = syncModuleTransforms;
+    return syncParser;
+  }
```
boneskull, line 233: This is incorrect.

Just because `moduleTransforms` is empty doesn't mean that the parsers are sync.  We don't have enough information in `mapParsers()` to make that determination.

We do, however, have the information in `link()` (which calls `mapParsers()`).  I'm not sure what's the best way to refactor this though. Send `mapParsers()` a `sync` bit?  idk 

@naugtur

packages/compartment-mapper/src/import-lite.js
```diff
@@ -85,16 +155,56 @@ export const loadFromMap = async (readPowers, compartmentMap, options = {}) => {
...
+        __shimTransforms__,
+        Compartment,
+      }));
+    } else {
+      // sync module transforms are allowed, because they are "compatible"
+      // with async module transforms (not vice-versa)
+      const moduleTransforms = /** @type {ModuleTransforms} */ ({
+        ...behavior.options.syncModuleTransforms,
+        ...behavior.options.moduleTransforms,
+      });
```
boneskull, line 196: I think this needs to be removed.  right @naugtur?

packages/compartment-mapper/src/import-lite.js
```diff
@@ -85,16 +155,56 @@ export const loadFromMap = async (readPowers, compartmentMap, options = {}) => {
...
+        behavior.options;
+      makeImportNowHook = makeImportNowHookMaker(
+        /** @type {SyncReadPowers} */ (readPowers),
+        entryCompartmentName,
+        {
+          compartmentDescriptors: compartmentMap.compartments,
+          searchSuffixes,
+          exitModuleImportNowHook,
+        },
+      );
```
boneskull, line 178: This should be the only difference between the "sync" and "async" calls to `link()`; the presence of a `makeImportNowHook`.

If it's not the conditional we're using here, it's a conditional we're using somewhere else.  I'm _pretty sure_ it should live in here, because we have access to the `options` and `readPowers` and we need both to determine whether we need to create an `makeImportNowHook`.

We need some information that says "enable import now hooks", and that information needs to get from **here**, then through **`link()`**, and finally into **`mapParsers()`**. Adding a `sync` flag seems bogus, but maybe a `makeLink()` and `makeMapParsers()` functions could be a more elegant way to express what we need to do.

@naugtur

naugtur: I'm in favor of `makeMapParsers` that accepts hints about sync/async situation and returns the util to pass down to link dependency injection style.  
That is if we won't find a simple correction of the current implementation. I'll look into it more tomorrow.

packages/compartment-mapper/src/import-lite.js
naugtur: yes, this was supposed to happen in map-parser.js only, not sure how it got here

packages/compartment-mapper/src/map-parser.js
naugtur: fixed it here: https://github.com/endojs/endo/pull/2464

packages/compartment-mapper/src/import-lite.js
boneskull: After pulling it out, a test failed, and I remember why it's here: if someone provides `syncTransforms` _but otherwise does not provide adequate configuration for dynamic requires_, we should merge it into `moduleTransforms`, otherwise `syncTransforms` will be completely ignored.  Ignoring it seems like poor UX; if we wanted to ban `syncTranforms` from non-dynamic-require use cases, then we should throw an exception.

boneskull: @naugtur 🔼 for your acknowledgement

packages/compartment-mapper/src/import-lite.js
```diff
@@ -85,16 +169,59 @@ export const loadFromMap = async (readPowers, compartmentMap, options = {}) => {
...
+        __shimTransforms__,
+        Compartment,
+      }));
+    } else {
+      // sync module transforms are allowed, because they are "compatible"
+      // with async module transforms (not vice-versa)
+
+      // this needs to be done _here_ in the case that `syncModuleTranforms` was provided
+      // against an otherwise-async operation.
+      const moduleTransforms = /** @type {ModuleTransforms} */ ({
```
naugtur, line 210: I'm pretty sure this is not right. 
The same operation is done right before they're being used. It's either a merge artifact or a mistake. syncModuleTransforms are now added to the list twice if I'm not mistaken. I'll verify that.

packages/compartment-mapper/src/import-hook.js
naugtur: It's become offtopic

packages/compartment-mapper/src/import-lite.js
naugtur: Can you create a test covering that? I don't see how they would be ignores as the mapParsers function is doing the same merging as we do here and I don't recall any other place where they're being consumed

boneskull: OK, this is addressed

packages/compartment-mapper/src/import.js
```diff
@@ -38,17 +43,36 @@ const assignParserForLanguage = (options = {}) => {
...
 export const loadLocation = async (
   readPowers,
   moduleLocation,
   options = {},
 ) => {
-  const { dev, tags, conditions = tags, commonDependencies, policy } = options;
+  const { dev, tags, commonDependencies, policy } = options;
+  // conditions are not present in SyncArchiveOptions
+  const conditions =
+    'conditions' in options ? options.conditions || tags : tags;
```
kriskowal, line 75:
```suggestion
  const { dev, tags, conditions = tags, commonDependencies, policy } = options;
```

packages/compartment-mapper/src/link.js
```diff
@@ -328,36 +211,57 @@ const makeModuleMapHook = (
   return moduleMapHook;
 };
 
+/**
+ * @type {ImportNowHookMaker}
+ */
+const impossibleImportNowHookMaker = () => {
```
kriskowal, line 217: Maybe consider `never` for `impossible`.

packages/compartment-mapper/src/map-parser.js
```diff
@@ -0,0 +1,339 @@
...
+
+// @ts-check
+
+import { syncTrampoline, asyncTrampoline } from '@endo/trampoline';
+import { parseExtension } from './extension.js';
+
+/**
+ * @import {
+ *   LanguageForExtension,
+ LanguageForModuleSpecifier,
```
kriskowal, line 16: *?

packages/compartment-mapper/src/node-powers.js
```diff
@@ -61,18 +88,18 @@ const makeReadPowersSloppy = ({ fs, url = undefined, crypto = undefined }) => {
     });
     await promise;
 
-    const path = fileURLToPath(location);
+    const filepath = fileURLToPath(location);
```
kriskowal, line 91: Consider `filePath` or merely `path`, since `path` is not otherwise in scope and `location` has no qualifier like `fileLocation`.

packages/compartment-mapper/src/node-powers.js
kriskowal: It’s a good point, @naugtur and I also don’t have any ideas on how to thread the hint, especially since some platforms have no way to communicate that hint. This way we at least have the same path regardless of backing storage.

packages/compartment-mapper/src/node-powers.js
```diff
@@ -146,8 +221,8 @@ const makeReadPowersSloppy = ({ fs, url = undefined, crypto = undefined }) => {
  * but `makeWritePowers` presents a type that requires `url`.
  *
  * @param {object} args
- * @param {typeof import('fs')} args.fs
- * @param {typeof import('url')} [args.url]
+ * @param {FsAPI} args.fs
```
kriskowal, line 224: I have a grudge against API where Interface is sufficient. At least consider `Api` if `Interface` is too bulky. That’s my other grudge: delimited initialisms like XMLHttpRequest should either be XMLHTTPRequest or XmlHttpRequest, and one of those loses the delimiter.

kriskowal (changes requested):
This has been a journey and I commend you, and apologize profusely for taking too long to review. That you have a green CI speaks volumes.

This is in the right shape. I have mostly annoying nits about naming and idioms below, most of which can justifiably be ignored at your discretion.

I only request you consider seriously the whether we can bear O(compartments) lookup and think about what it would take to do a more efficient prefix match. I do not want you to build a prefix trie, but if iteratively popping components off a module specifier until you find a matching compartment location works, please consider that, or explain why that won’t work or isn’t necessary.

There is a breaking change here now that parsers are part of the public API, but I am not worried about that since all usage of them is local to Endojs/Endo and is loosely coupled to the specific interface. It might be worth noting. It might not.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -68,10 +91,68 @@ const nodejsConventionSearchSuffixes = [
...
+      if (!compartmentDescriptor.location.startsWith(moduleSpecifierUrl)) {
+        if (moduleSpecifierUrl.startsWith(someDescriptor.location)) {
+          // compartmentDescriptor is a dependency of someDescriptor; we
+          // can use that compartment
+          // similarly, if the relationship is true the other way around.
+          // We could allow requiring any compartment contents by absolute path, but requiring
+          // a relationship between compartments is a reasonable limitation.
+          if (
+            someDescriptor.name in compartmentDescriptor.modules ||
+            compartmentDescriptor.name in someDescriptor.modules
```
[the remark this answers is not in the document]
naugtur: Sounds like something we want to have a fixture for in tests (not necessarily as part of this PR)

packages/compartment-mapper/src/import-hook.js
```diff
@@ -68,10 +91,68 @@ const nodejsConventionSearchSuffixes = [
...
+          ) {
+            return {
+              specifier: absoluteModuleSpecifier,
+              compartment: compartments[someDescriptorName],
+            };
```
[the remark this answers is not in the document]
naugtur: I was confused, then I found 
```
export type ModuleDescriptor =
  | SourceModuleDescriptor
  | NamespaceModuleDescriptor
  // To be deprecated:
  | RecordModuleDescriptor
  | ModuleExportsNamespace
  | VirtualModuleSource
  | PrecompiledModuleSource
  | string;
  ```
  and also:
```
namespace: string | ModuleExportsNamespace;
```
And I'm now confused on a different layer of abstraction. Which I've learned is a good thing.

File: packages/compartment-mapper/src/import-hook.js [hunk 45]
```diff
@@ -100,6 +181,185 @@ export const exitModuleImportHookMaker = ({
...
+    computeSha512,
+    moduleDescriptors,
+    moduleSpecifier,
+    packageLocation,
+    packageSources,
+    readPowers,
+    sourceMapHook,
+    strictlyRequiredForCompartment,
+  },
+  { maybeRead, parse, shouldDeferError = () => false },
```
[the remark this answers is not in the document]
naugtur: My bad. The reasoning was the first argument is emulating lexical context for the code now that it was extracted, but for that to make sense you'd need to know the history.

packages/compartment-mapper/src/import.js
naugtur: I bet typescript made him do it.

packages/compartment-mapper/src/link.js
naugtur: I wouldn't know how to make something that's not an adjective work here.

packages/compartment-mapper/src/node-powers.js
boneskull: `path` is in scope

packages/compartment-mapper/src/import-hook.js
```diff
@@ -68,10 +91,68 @@ const nodejsConventionSearchSuffixes = [
   '/index.node',
 ];
 
+/**
+ * Given a module specifier which is an absolute path, attempt to match it with
+ * an existing compartment; return a {@link RedirectStaticModuleInterface} if found.
```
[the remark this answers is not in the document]
boneskull: @kriskowal 

Neither `SourceModuleDescriptor` nor `NamespaceModuleDescriptor` is valid, in this case.  The former because we do not have a `source` property, and the latter because we do not have a `namespace` property; all we have is `{string} specifier` and `{Compartment} compartment`.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -7,20 +7,41 @@
...
+ *   MakeImportNowHookMakerOptions,
+ *   ModuleDescriptor,
+ *   ParseResult,
+ *   ReadFn,
+ *   ReadPowers,
+ *   SourceMapHook,
+ *   Sources,
+ *   SyncReadPowers
+ * } from './types.js'
+ */
```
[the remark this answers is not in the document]
boneskull: I think I asked a computer to format it this way.  There's some sort of prettier or eslint plugin that will do this sort of thing, I think.

packages/compartment-mapper/src/types.js
```diff
@@ -128,6 +135,14 @@ export {};
  * @returns {Promise<Uint8Array | undefined>} bytes
  */
 
+/**
+ * A resolution of `undefined` indicates `ENOENT` or the equivalent.
+ *
+ * @callback MaybeReadSyncFn
```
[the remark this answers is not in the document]
boneskull: Is this a blocker?  There's a lot of stuff to rename, if so.

boneskull:
@kriskowal I've applied all of the naming-related changes you've requested.

packages/compartment-mapper/src/import.js
boneskull: yes, it would require a type assertion otherwise

packages/compartment-mapper/src/import-hook.js
```diff
@@ -68,10 +91,136 @@ const nodejsConventionSearchSuffixes = [
...
+        throw new Error(
+          `Could not import unknown module: ${q(absoluteModuleSpecifier)}`,
+        );
+      }
+
+      someLocation = parentLocation;
+    }
+  }
+
+  // for (const [someDescriptorName, someDescriptor] of entries(
```
naugtur, line 180: leftover?

packages/compartment-mapper/src/import-hook.js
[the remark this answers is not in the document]
kriskowal: The difference is only that the the Moddable folks preferred to rename `specifier` to`source` so that it is clear that the referent (still a string) means that this compartment adopts the other compartment’s ModuleSource, vs `namespace` when it shares the other compartment’s module instance. The name `specifier` left that ambiguous.

packages/compartment-mapper/src/import-hook.js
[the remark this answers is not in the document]
kriskowal: There’s also an entry in [NEWS](https://github.com/endojs/endo/blob/f63db3a4706fa5758ca138a993618f5c8763c9a5/packages/ses/NEWS.md#v160-2024-07-30)

As mentioned above, Moddable noticed that `specifier` was ambiguous and pushed us to rename the property to `source` to distinguish a `namespace` specifier. The former adopts the source and the latter adopts the instance.

packages/compartment-mapper/src/import.js
kriskowal: In that case, disregard. I wonder why I got away with this pattern elsewhere.

kriskowal (approved):
This is in the right shape. I will not need another review. Please review prior comments and delete dead code. Naming nits are optional.

Please add an entry to NEWS.md for the new feature.

You should be able to merge when you are ready. Please rebase on `master` before merging, ensure that CI is green, and either squash with a merge commit or manually merge fixups and edits into the prior feature commits so that history appears linear. Also, review your merge commit description, which is generated from the PR description.

I would like to get this proven in Agoric SDK CI, so I’m going to produce a release for `@endo/trampoline` today and kick that CI job again to see what happens. I’ll let you know.

packages/compartment-mapper/src/import-hook.js
boneskull: yaeh, maybe that issue can be closed?  [#2466](https://github.com/endojs/endo/issues/2466)

packages/compartment-mapper/NEWS.md
```diff
@@ -1,5 +1,12 @@
 User-visible changes to `@endo/compartment-mapper`:
 
+# v1.3.0 (2024-??-??)
```
kriskowal, line 3: Please spell this `# Next release`. I have a grep in the release workflow to update all these.

naugtur:
https://github.com/endojs/endo/pull/2310#discussion_r1783324595
I'd like to elaborate on that, preferably synchronously

packages/compartment-mapper/src/node-modules.js
```diff
@@ -724,6 +734,7 @@ const translateGraph = (
       parsers,
       types,
       policy: /** @type {SomePackagePolicy} */ (packagePolicy),
+      compartments: compartmentNames,
```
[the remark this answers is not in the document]
naugtur: note for later: this could be narrowed down to what's allowed by policy. TODO: create an issue

PR #2311, opened 2024-06-08
feat(ses): ArrayBuffer.p.transferToImmutable

### LINKED ISSUE #1331: Support passable immutable ArrayBuffer as copy-data

erights:
Closes: #XXXX
Refs: #1538 #1331 #2309 

## Description

Alternative to #2309 as suggested by @phoddie at https://github.com/endojs/endo/pull/2309#issuecomment-2155513240

A step towards fixing #1331 , likely by restaging #1538 on this one and then fixing it.
By making `ArrayBuffer.p.immutable` as-if part of the language, in #1538 `passStyleOf` will be able to recognize an immutable ArrayBuffer even though it carries its own methods, avoiding the eval-twin problems that otherwise thwart such plans. This is based on one of the candidates explained at #1331 . That `passStyleOf` behavior, together with `ArrayBuffer.p.transferToImmutable` opens the door for `marshal` to serialize and unserialize these as additional ocapn Passables.

Additionally, by proposing it to tc39 as explained below, we'd enable immutable TypedArrays and DataViews as well, and XS could place all these in ROM cheaply, while conforming to the language spec. When also hardened, XS could judge these to be pure. Attn @phoddie @patrick-soquet

- [x] TODO Must treat the hidden prototype and its methods as hidden intrinsics, so they get hardened when they should, along with the others.

From the initial NEWS.md

- Adds `ArrayBuffer.p.immutable` and `ArrayBuffer.p.transferToImmutable` as a shim for a future proposal. It makes an ArrayBuffer-like object whose contents cannot be mutated. However, due to limitations of the shim
  - Unlike `ArrayBuffer` and `SharedArrayBuffer` this shim's ArrayBuffer-like object cannot be transfered or cloned between JS threads.
  - Unlike `ArrayBuffer` and `SharedArrayBuffer`, this shim's ArrayBuffer-like object cannot be used as the backing store of TypeArrays or DataViews.
  - On Node >= 21 we use the builtin `transferToFixed` to transfer exclusive access to the array buffer contents. On Node <= 20, we emulate `transferToFixedLength` with `structuredClone`. On platforms with neither `transferToFixedLength` nor `structuredClone`, we use `slice` to copy the contents, but have no way to detach the original.
  - Even after the upcoming `transferToImmutable` proposal is implemented by the platform, the current code will still replace it with the shim implementation, in accord with shim best practices. See https://github.com/endojs/endo/pull/2311#discussion_r1632607527 . It will require a later manual step to delete the shim, after manual analysis of the compat implications.

### Security Considerations

The eval-twin problem explained at #1331 is a security problem. This PR is one candidate for solving that problem, unblocking #1538 so it can fix #1331. Further, if accepted into a future version of the language, the immutability it provides will generally help security.

### Scaling Considerations

This shim implementation likely does more copying than even a naive native implementation would. A native implementation may even engage in copy-on-write tricks that this shim cannot. Use of the shim should beware of these "extra" copying costs. (Starting in Node 21, the shim's `ArrayBuffer.p.transferToImmutable` will no longer do an extra copy.)

Generally we've kept hardened JS close to standard JS, and we've kept the ses-shim close to hardened JS. With this PR, we'd need to explain `ArrayBuffer.p.transferToImmutable` and `ArrayBuffer.p.immutable` as part of the hardened JS implemented by the ses-shim, even though we have not ***yet*** proposed it to tc39.

### Testing Considerations

Ideally, we should identify the subset of test262 `ArrayBuffer` tests that should be applicable to immutable ArrayBuffers, and duplicate them for that purpose.

### Upgrade Considerations

Nothing breaking.

NEWS.md updated


erights:
@phoddie @patrick-soquet , please consider yourselves reviewers of this one instead. I look forward to your comments.

mhofman:
From https://github.com/endojs/endo/pull/2309#issuecomment-2155714787
> I just noticed another problem in shimming this though. Node 20, which we still support, has no way within the language to detach an ArrayBuffer. Therefore, on Node 20, `transferToImmutable` will copy (using `slice`) and leave the original undetached. Just another loss of fidelity for the shim that needs to be documented. Sigh. Still worth it to have a cleaner and more minimal proposal.

Actually Node.js [supports `structureClone` as a global since Node v17](https://nodejs.org/docs/latest-v18.x/api/globals.html#structuredclonevalue-options), so you definitely can [transfer an ArrayBuffer](https://stackoverflow.com/a/74082322) in all supported versions

packages/ses/NEWS.md
```diff
@@ -1,5 +1,12 @@
 User-visible changes in SES:
 
+# Next release
+
+- Adds `ArrayBuffer.p.immutable` and `ArrayBuffer.p.transferToImmutable` as a shim for a future proposal. It makes an ArrayBuffer-like object whose contents cannot be mutated. However, due to limitations of the shim
+  - Unlike `ArrayBuffer` and `SharedArrayBuffer` this ArrayBuffer-like object cannot be transfered or cloned between JS threads.
+  - Unlike `ArrayBuffer` and `SharedArrayBuffer`, this ArrayBuffer-like object cannot be used as the backing store of TypeArrays or DataViews.
+  - On Node 20, which lacks `transfer`, `transferToFixedLength`, or any other way within the language to detach an ArrayBuffer, `transferToImmutable` will copy the contents of the original, but leave the original undetached. Node 21 does not have this limitation.
```
gibson042, line 8: see https://github.com/endojs/endo/pull/2309#issuecomment-2156126666
```suggestion
  - On Node 20, we emulate `transferToFixedLength` with a combination of `slice` to copy the original and `structuredClone` to detach it. Node 21 does not have this inefficiency.
```

gibson042 (commented):
Comments to better emulate `transferToFixedLength`.

packages/ses/src/commons.js
```diff
@@ -174,6 +176,19 @@ export const arraySome = uncurryThis(arrayPrototype.some);
...
+  arrayBufferPrototype.transferToFixedLength
+    ? // @ts-expect-error absent from Node 20, which we still support
+      uncurryThis(arrayBufferPrototype.transferToFixedLength)
+    : (arrayBuffer, newLength = arrayBuffer.byteLength) =>
+        // There is no `transferToFixedLength` on Node 20, which we still support.
+        // In that case, there is no way just within the language to detach an
+        // ArrayBuffer, so the best we can do is emulate it using `slice`.
+        // Unfortunately, this leaves the original non-detached.
+        arrayBufferSlice(arrayBuffer, 0, newLength);
+//
```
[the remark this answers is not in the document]
erights: Done. But made tolerant on the absence of `structuredClone` since it is not part of official JS and does not necessarily exist on all conforming platforms.

packages/ses/NEWS.md
erights: Done

erights:
> From [#2309 (comment)](https://github.com/endojs/endo/pull/2309#issuecomment-2155714787)
> 
> > I just noticed another problem in shimming this though. Node 20, which we still support, has no way within the language to detach an ArrayBuffer. Therefore, on Node 20, `transferToImmutable` will copy (using `slice`) and leave the original undetached. Just another loss of fidelity for the shim that needs to be documented. Sigh. Still worth it to have a cleaner and more minimal proposal.
> 
> Actually Node.js [supports `structuredClone` as a global since Node v17](https://nodejs.org/docs/latest-v18.x/api/globals.html#structuredclonevalue-options), so you definitely can [transfer an ArrayBuffer](https://stackoverflow.com/a/74082322) in all supported versions

Done, using Richard's suggested code.

packages/ses/src/commons.js
```diff
@@ -19,6 +19,7 @@ export { universalThis as globalThis };
 
 export const {
   Array,
```
gibson042, line 21:
```suggestion
  structuredClone,
  Array,
```
(to avoid being affecting by post-hoc manipulation)

erights: To solve the stated problem, line 181 below already says

```js
const { structuredClone } = globalThis;
```

I do it in the ArrayBuffer section as a non-export because `structuredClone` is not part of JS, so I would rather not export it from commons.js

Resolving. But unresolve if you disagree.

packages/ses/NEWS.md
```diff
@@ -5,7 +5,7 @@ User-visible changes in SES:
 - Adds `ArrayBuffer.p.immutable` and `ArrayBuffer.p.transferToImmutable` as a shim for a future proposal. It makes an ArrayBuffer-like object whose contents cannot be mutated. However, due to limitations of the shim
   - Unlike `ArrayBuffer` and `SharedArrayBuffer` this ArrayBuffer-like object cannot be transfered or cloned between JS threads.
   - Unlike `ArrayBuffer` and `SharedArrayBuffer`, this ArrayBuffer-like object cannot be used as the backing store of TypeArrays or DataViews.
-  - On Node 20, which lacks `transfer`, `transferToFixedLength`, or any other way within the language to detach an ArrayBuffer, `transferToImmutable` will copy the contents of the original, but leave the original undetached. Node 21 does not have this limitation.
+  - On Node 20, we emulate `transferToFixedLength` with a combination of `slice` to copy the original and `structuredClone` to detach it. Node 21 does not have this inefficiency.
```
mhofman, line 8: `structuredClone` is supported in all LTS

```suggestion
  - On Node, we emulate `transferToFixedLength` with a combination of `slice` to copy the original and `structuredClone` to detach it. Node 21 does not have this inefficiency.
```

But I also don't understand why the `slice`? `transfer` methods should detach the original.

packages/ses/src/commons.js
```diff
@@ -177,18 +177,32 @@ export const arraySort = uncurryThis(arrayPrototype.sort);
...
+
+        // `slice` accepts negative arguments but `transferToFixedLength` does not...
+        // get at the underlying ToIndex operation through `BigInt.asUintN`
+        // (avoiding the redundant allocation of e.g. `ArrayBuffer(newLength)`)
+        // and ToNumber through unary `+` (rather than `Number(newLength)`,
+        // which fails to reject BigInts).
+        newLength = +newLength;
+        bigIntAsUintN(newLength, 0n);
+
+        const copied = arrayBufferSlice(arrayBuffer, 0, newLength);
```
mhofman, line 200: This slice should not be present at all if `structuredClone` doesn't exist:
- a `transfer` method is meant to detach the original
- a `slice` is not needed even after transfer since the original cannot be resizable

We should only use a `slice` if there is no way to detach the original, which becomes a deficiency of the shim.

packages/ses/src/immutable-array-buffer-shim.js
```diff
@@ -9,88 +8,99 @@ import {
...
- * Therefore, `ImmutableArrayBufferInternal.prototype` and all its methods
- * and accessor functions effectively become hidden intrinsics.
- *
- * TODO handle them as hidden intrinsics, so they get hardened when they should.
- */
-class ImmutableArrayBufferInternal {
-  /** @type {ArrayBuffer} */
-  #buffer;
+// If it already exists, don't replace it with the shim.
+if (!('transferToImmutable' in arrayBufferPrototype)) {
```
mhofman, line 12: Conditional shims for non stage 4 features have been seen as problematic by the community as there is a risk that the program would start relying on a shim behavior that does not match what the final spec says, resulting in breakage when the engine starts implementing the feature.

mhofman (changes requested):
`transferToImmutable` should detach the original buffer is possible

packages/ses/src/immutable-array-buffer-shim.js
erights: Makes sense. Will do. Is there something I can cite?

packages/ses/NEWS.md
erights: Why delete the "20" but still make a statement about "21"? Seems confusing to me. I'll revise to "<= 20" and ">= 21"

mhofman: Yeah sorry I didn't have a fully well formed suggestion here. I contend that with structured clone there is no "inefficiency", so the only thing that matters is "Node <17" which has a lack of fidelity.

packages/ses/src/commons.js
erights: > This slice should not be present at all if `structuredClone` doesn't exist:

I don't understand this. Seems inconsistent with 

> We should only use a `slice` if there is no way to detach the original, which becomes a deficiency of the shim.

which I do understand and agree with. In particular, if there is no `transfer`, `transferToFixed`, or `structuredClone`, then our shim should use `slice` and have this deficiency on those platforms. Yes?

erights: In a platform with `structuredClone` but without `transfer` or `transferToFixedLength`, I can transfer and detach without `slice` if the length doesn't need to change. But if the length does need to change, then I'd still need to use `slice`, yes?

mhofman: In what case does the length need to change?

packages/ses/src/immutable-array-buffer-shim.js
mhofman: I believe there is an article out there but I couldn't find a link

packages/ses/NEWS.md
erights: PTAL. I'm resolving because I think this is done, but please unresolve if you disagree.

packages/ses/src/commons.js
erights: PTAL. I'm resolving because I think this is done, but please unresolve if you disagree.

packages/ses/src/immutable-array-buffer-shim.js
erights: I simply cited your comment above. Done.

packages/ses/src/commons.js
erights: > In what case does the length need to change?

Argh. I see your point! Unresolving. I need to see if I can resize, and do that instead of `slice` when `resize` is possible.

erights: Note that there is no `resize` on Node <= 18.

erights: The ArrayBuffer made by `structuredClose` is not `resizable`, so that plan wouldn't have worked. I added a comment to that effect and continue to use `slice`. I don't see any other choice. Resolving again.

erights:
Converted to draft, because all the subtlety around resizing is
- currently still wrong
- needs tests.

mhofman:
> Converted to draft, because all the subtlety around resizing is

Sorry I think I expressed myself poorly in previous comments.

- If there is no "arraybuffer resize" feature in the platform, we shouldn't be in the business of providing a shim for it.
  - IMO it's acceptable to have `transferToImmutable` shimmed but not `transferToFixedLength`.
- We can use either `structuredClone` or `transferToFixedLength` to implement detachment of array buffer. If neither is available, we fallback to slice, which is a shim fidelity issue only (and performance penalty)

packages/ses/src/commons.js
```diff
@@ -174,6 +176,89 @@ export const arraySome = uncurryThis(arrayPrototype.some);
...
+          // On platforms like Node 20
+          // - without`tranferToFixedLength` or `transfer`
+          // - with `structuredClone`
+          // - with `resize`
+          //
+          // It might seem like we could avoid the extra copy by
+          // `newBuffer.resize(newLength)`. But `structuredClone`
+          // makes ArrayBuffers that are not resizable.
+          return arrayBufferSlice(newBuffer, 0, newLength);
+        }
```
gibson042, line 233:
```suggestion
        (arrayBuffer, newLength = arrayBufferByteLength(arrayBuffer)) => {
          // There is no `transferToFixedLength` on Node <= 20, but there
          // is web-standard `structuredClone` on Node >= 17, on all modern
          // browsers, and on many other JS platforms.
          // In those cases, we first use `structuredClone` to get a fresh
          // buffer with exclusive access to the underlying data, while
          // detaching it from the original `arrayBuffer`.

          // Before looking at actual arguments, verify that the input is an
          // ArrayBuffer.
          arrayBufferByteLength(arrayBuffer);

          // Calculate ToIndex(newLengthAsNumber) using `BigInt.asUintN`,
          // first getting newLengthAsNumber as ToNumber(newLength) using unary `+`
          // (rather than `Number(newLength)`, which fails to reject BigInts).
          newLength = +newLength;
          bigIntAsUintN(newLength, 0n);

          const newBuffer = /** @type {ArrayBuffer} */ (
            structuredClone(arrayBuffer, {
              transfer: [arrayBuffer],
            })
          );

          // We might already have what we need.
          // NOTE: The check for resizability is necessary to correctly emulate
          // ArrayBuffer.prototype.transferToFixedLength, but a more narrow
          // function specialized to e.g. transferToImmutable could skip it.
          if (newLength === arrayBufferByteLength(newBuffer) && !arrayBufferResizable(newBuffer)) {
            return newBuffer;
          }

          // We might be able to copy some or all of the contents.
          if (newLength <= arrayBufferByteLength(newBuffer)) {
            const copied = arrayBufferSlice(newBuffer, 0, newLength);
            const copiedLength = arrayBufferByteLength(copied);
            if (copiedLength !== newLength) {
              throw RangeError(`internal: length ${copiedLength} should have been ${newLength}`);
            }
            return copied;
          }

          // We need a bigger boat.
          // NOTE: Uint8Array must be extracted from globalThis.
          const view = new Uint8Array(newLength);
          typedArraySet(view, new Uint8Array(newBuffer));
          return typedArrayBuffer(view);
```

packages/ses/src/commons.js
```diff
@@ -174,6 +176,89 @@ export const arraySome = uncurryThis(arrayPrototype.some);
...
+          // get at the underlying ToIndex operation through `BigInt.asUintN`
+          // (avoiding the redundant allocation of e.g. `ArrayBuffer(newLength)`)
+          // and ToNumber through unary `+` (rather than `Number(newLength)`,
+          // which fails to reject BigInts).
+          newLength = +newLength;
+          bigIntAsUintN(newLength, 0n);
+
+          const newBuffer = arrayBufferSlice(arrayBuffer, 0, newLength);
+          return newBuffer;
+        };
```
gibson042, line 260:
```suggestion
        (arrayBuffer, newLength = arrayBuffer.byteLength) => {
          // There is no `transferToFixedLength` on Node <= 20,
          // and no `structuredClone` on Node <= 17 and possibly on some
          // non-browser JavaScript platforms.
          // In those cases,
          // and assuming the absence of `transfer`, we cannot detach
          // the original, but we must still produce a new fresh buffer with
          // exclusive mutability of its underlying state.

          // Before looking at actual arguments, verify that the input is an
          // ArrayBuffer.
          const srcLength = arrayBufferByteLength(arrayBuffer);

          // Calculate ToIndex(newLengthAsNumber) using `BigInt.asUintN`,
          // first getting newLengthAsNumber as ToNumber(newLength) using unary `+`
          // (rather than `Number(newLength)`, which fails to reject BigInts).
          newLength = +newLength;
          bigIntAsUintN(newLength, 0n);

          // We might be able to copy some or all of the contents.
          if (newLength <= srcLength) {
            const copied = arrayBufferSlice(arrayBuffer, 0, newLength);
            const copiedLength = arrayBufferByteLength(copied);
            if (copiedLength !== newLength) {
              throw RangeError(`internal: length ${copiedLength} should have been ${newLength}`);
            }
            return copied;
          }

          // We need a bigger boat.
          // NOTE: Uint8Array must be extracted from globalThis.
          const view = new Uint8Array(newLength);
          typedArraySet(view, new Uint8Array(arrayBuffer));
          return typedArrayBuffer(view);
        };
```

packages/ses/src/immutable-array-buffer-shim.js
```diff
@@ -0,0 +1,115 @@
...
+ * TODO handle them as hidden intrinsics, so they get hardened when they should.
+ */
+class ImmutableArrayBufferInternal {
+  /** @type {ArrayBuffer} */
+  #buffer;
+
+  constructor(buffer) {
+    // This also enforces that `buffer` is a genuine `ArrayBuffer`.
+    // This constructor is deleted from the prototype below.
+    this.#buffer = arrayBufferSlice(buffer, 0);
```
gibson042, line 44: We shouldn't be slicing an ArrayBuffer already came from detachment; full responsibility should fall on either `transferToImmutable` or `ImmutableArrayBufferInternal` but not both.

gibson042 (commented):
Comments to improve fidelity:
* support for expansion to a bigger size, e.g. `new ArrayBuffer(6).transferToImmutable(8)` (which should also be tested)
* proper receiver and argument checking
* extraction and use of ArrayBuffer.prototype `byteLength` and `resizable` getters and %TypedArray.prototype% `buffer` getter

mhofman:
> which is admittedly itself not necessary in the common case of a length-preserving transfer of non-resizable input

> * support for expansion to a bigger size, e.g. `new ArrayBuffer(6).transferToImmutable(8)`

Before we go on through another round of implementation changes, we should settle the proposed semantics. 

IMO, if someone want to change the length, they can do transferToFixedLength(newLength).transferToImmutable() No need to overload the methods. That only raises complexity of the API and the shim for no good reason.

gibson042:
> IMO, if someone want to change the length, they can do `transferToFixedLength(newLength).transferToImmutable()`. No need to overload the methods. That only raises complexity of the API and the shim for no good reason.

`transfer` and `transferToFixedLength` both support a _newLength_ parameter which can take any value that coerces to a valid ArrayBuffer length (i.e., an integer in [0, 2**53 - 1]). It would be surprising if `transferToImmutable` were to lack a parameter with identical behavior, and basically unacceptable for it to have a parameter supporting contraction but not expansion (as is currently the case in this PR).

packages/ses/src/immutable-array-buffer-shim.js
gibson042: > Conditional shims for non stage 4 features have been seen as problematic by the community as there is a risk that the program would start relying on a shim behavior that does not match what the final spec says, resulting in breakage when the engine starts implementing the feature... I believe there is an article out there but I couldn't find a link

https://ponyfill.com (which conveniently also links to http://svdictionary.com/words/ponyfill and https://ponyfoo.com/articles/polyfills-or-ponyfills and https://kikobeats.com/polyfill-ponyfill-and-prollyfill/ ).

Summary: for work like this, we can export a function for use like `transferToImmutable(arrayBuffer)` and `transferToImmutable(arrayBuffer, newLength)`, but SHOULD NOT add any new methods or properties to ArrayBuffer.prototype.

erights:
Closing in favor of https://github.com/endojs/endo/pull/2400/files

PR #2312, opened 2024-06-09
types(marshal): fix #1257 typing of `deeplyFulfilled`

### LINKED ISSUE #1257: deeplyFulfilled erases type information
https://github.com/endojs/endo/blob/master/packages/marshal/src/deeplyFulfilled.js#L36-L39

https://github.com/endojs/endo/blob/master/packages/marshal/src/types.js#L26

Because `Passable` is any, the type passed into `deeplyFulfilled` gets erased. The decision to adopt bare `deeplyFulfilled` in https://github.com/Agoric/agoric-sdk/pull/5891/commits/a8db35404777d1f9e385a7ce9ba1add827e4cff6 has erased type information there. Example impact: I just had a CI failure from a refactoring that the types would have alerted me to.

erights:
Closes: #1257 
Refs: https://github.com/Agoric/agoric-sdk/pull/5993 https://github.com/Agoric/agoric-sdk/pull/6816 https://github.com/endojs/endo/pull/1455

## Description

Low urgency. This is a minor reduction of tech debt.

https://github.com/Agoric/agoric-sdk/pull/5993 introduced `deeplyFulfilledObject` to work around the deficiencies in the typing of `deeplyFulfilled` explained at #1257 . This PR fixes those deficiencies. This does not necessary make `deeplyFulfilledObject` unneeded, but it should enable many callers of `deeplyFulfilledObject` to call `deeplyFulfilled` instead without loss of type fidelity.

Note: the code in this first commit probably does not use TS typing well, resulting in many internal `@ts-expect-error` directives in the implementation. This is because I still don't really understand what I'm doing with complex TS types. Reviewers, help appreciated.

But IIUC, the external type should now be as good as `deeplyFulfilledObject`. Hopefully, the only typing flaws are encapsulated by this implementation. Yes?

### Security Considerations

More accurate typing helps security. More precise but inaccurate unsound typing hurts security. Our security practices are already adapted to this dilemma, and this PR should not have much effect.

### Scaling Considerations

none

It would be nice if this PR eventually let us drop the need to explain `deeplyFulfilledObject` and why it co-exists with `deeplyFulfilled`

### Testing Considerations

- [ ] since this is all about changing static types, I need to write some static type tests.

Even though the code itself is also a bit different, these differences are with high confidence a pure refactor, so the existing dynamic tests should adequately test that.

Assuming that these changes are a pure refactor with no dynamic behavior changes, there should be no dynamic compat issues.

The tighter typing of `deeplyFulfilled` does raise the possibility that some call sites will no longer pass static type checks. But no such problem appears within the PR's CI, and therefore within the endo repo.

### Upgrade Considerations

none.

- [ ] Include `*BREAKING*:` in the commit message with migration instructions for any breaking change.

### DIFF
--- a/packages/marshal/src/deeplyFulfilled.js
+++ b/packages/marshal/src/deeplyFulfilled.js
@@ -66,7 +66,7 @@ const { fromEntries } = Object;
  * Passable, or if any of the transitive promises fulfill to something
  * that is not Passable, then the returned promise rejects.
  *
- * If `val` or its parts are non-key Passables only *because* they contains
+ * If `val` or its parts are non-key Passables only *because* they contain
  * promises, the deeply fulfilled forms of val or its parts may be keys. This
  * is for the higher "@endo/patterns" level of abstraction to determine,
  * because it defines the `Key` notion in question.
@@ -81,9 +81,7 @@ export const deeplyFulfilled = async val => {
   // https://github.com/endojs/endo/issues/1257 may be relevant.
 
   if (!isObject(val)) {
-    const prim = /** @type {Primitive} */ (val);
-    // @ts-expect-error not assignable to type 'DeeplyAwaited<T>'
-    return prim;
+    return /** @type {DeeplyAwaited<T>} */ (val);
   }
   if (isPromise(val)) {
     return E.when(val, nonp => deeplyFulfilled(nonp));
--- a/packages/marshal/src/deeplyFulfilled.js
+++ b/packages/marshal/src/deeplyFulfilled.js
@@ -1,12 +1,11 @@
-/// <reference types="ses"/>
-
+import { X, q } from '@endo/errors';
 import { E } from '@endo/eventual-send';
 import { isPromise } from '@endo/promise-kit';
 import { getTag, isObject, makeTagged, passStyleOf } from '@endo/pass-style';
 
-/** @import {Passable, Primitive, CopyRecord, CopyArray, CopyTagged, RemotableObject} from '@endo/pass-style' */
-
-import { X, q } from '@endo/errors';
+/**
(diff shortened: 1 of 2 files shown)


packages/marshal/src/deeplyFulfilled.js
```diff
@@ -31,51 +68,64 @@ const { fromEntries } = Object;
  *
  * If `val` or its parts are non-key Passables only *because* they contains
```
gibson042, line 69:
```suggestion
 * If `val` or its parts are non-key Passables only *because* they contain
```

packages/marshal/src/deeplyFulfilled.js
```diff
@@ -31,51 +68,64 @@ const { fromEntries } = Object;
...
 export const deeplyFulfilled = async val => {
+  // TODO Figure out why we need these at-expect-error directives below
+  // and fix if possible.
+  // https://github.com/endojs/endo/issues/1257 may be relevant.
+
   if (!isObject(val)) {
-    return val;
+    const prim = /** @type {Primitive} */ (val);
+    // @ts-expect-error not assignable to type 'DeeplyAwaited<T>'
+    return prim;
```
gibson042, line 86: Does it work if you just assert the maximally broad return type?
```suggestion
    return /** @type {DeeplyAwaited<T>} */ (val);
```

gibson042 (approved):

packages/marshal/src/deeplyFulfilled.js
erights: Done

packages/marshal/src/deeplyFulfilled.js
erights: Done

PR #2317, opened 2024-06-12
chore: Update Promise Kit README

### LINKED ISSUE #2318: Promise Kit Empty README

promise-kit package currently has an empty README file - this needs to be fixed.

mudassir-agoric:
Update Promise Kit README

fixes #2318

### DIFF
--- a/packages/promise-kit/README.md
+++ b/packages/promise-kit/README.md
@@ -1,6 +1,7 @@
 # Promise Kit
 
 The promise-kit package provides a simple abstraction for creating and managing a promise. It exports, `makePromiseKit` which is a utility function used to create a Promise and its associated resolver and rejector functions. This is particularly useful in asynchronous programming, where you might need to create a promise and resolve or reject it at a later point in time.
+Note that this serves as a "ponyfill" for `Promise.withResolvers`, making certain accommodations to ensure that the resulting promises can pipeline messages through `@endo/eventual-send`.
 
 ## Usage
 
--- a/packages/promise-kit/README.md
+++ b/packages/promise-kit/README.md
@@ -1,6 +1,6 @@
 # Promise Kit
 
-The promise-kit package provides a simple abstraction for creating a promise and its corresponding resolve and reject functions, allowing for easier promise management. It exports, `makePromiseKit` which is a utility function used to create a Promise and its associated resolver and rejector functions. This is particularly useful in asynchronous programming, where you might need to create a promise and resolve or reject it at a later point in time.
+The promise-kit package provides a simple abstraction for creating and managing a promise. It exports, `makePromiseKit` which is a utility function used to create a Promise and its associated resolver and rejector functions. This is particularly useful in asynchronous programming, where you might need to create a promise and resolve or reject it at a later point in time.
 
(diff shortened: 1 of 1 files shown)


packages/promise-kit/README.md
```diff
@@ -1,3 +1,71 @@
 # Promise Kit
 
-A helper for making promises
+The promise-kit package provides a simple abstraction for creating a promise and its corresponding resolve and reject functions, allowing for easier promise management. It exports, `makePromiseKit` which is a utility function used to create a Promise and its associated resolver and rejector functions. This is particularly useful in asynchronous programming, where you might need to create a promise and resolve or reject it at a later point in time.
+
```
kriskowal, line 4: @mhofman Would you contribute a word about why you would use this instead of `Promise.withResolvers` this year?

It may be worth stating up front that this is a “ponyfill” for `Promise.withResolvers` that makes certain accommodations to ensure that the resulting promises can pipeline messages through `@endo/eventual-send`.

kriskowal (approved):

packages/promise-kit/README.md
mudassir-agoric: @kriskowal I can append this text at the end of above if that is desirable:
**“ponyfill” for Promise.withResolvers that makes certain accommodations to ensure that the resulting promises can pipeline messages through @endo/eventual-send.**

kriskowal: This is good, thank you.

mhofman: It's actually more than this. It also fixes resolver leaks in common implementations. Once `withResolvers` get introduced later this year, we should consider making this a full shim (this package is already a shim for `race` which is similarly leaky)

mhofman: Oh yeah it also creates `HandledPromise`s, however I admit I still get confused about the layering there, @michaelfig might be more helpful.

mhofman: Thinking about this, a repaired `HandledPromise.withResolvers` would do what `makePromiseKit` does since it defers to the `this` to decide what the constructor is.

michaelfig: `HandledPromise` is an augmentation of `Promise` that supports eventual operations.  It does not rely on any other packages, but the `@endo/eventual-send` package provides a shim for it in addition to the `makeE` eventual send operator maker.  I'm not yet certain where `makeE` actually belongs; I'd like to have just one version of it that (like the version in `@agoric/vow/src/E.js`) can support a higher-than-eventual-send layer's implementation of pseudo-promises.

`makePromiseKit` uses `HandledPromise` if available; otherwise just `Promise`.

mudassir-agoric: So, should I add the following or is there a more accurate statement that needs to be added? Or leave it as is?

**This is a “ponyfill” for Promise.withResolvers that makes certain accommodations to ensure that the resulting promises can pipeline messages through @endo/eventual-send.**

kriskowal: I think this verbiage is fine and we can refine in follow-up. Please merge at your leisure.

Issue #2251, opened 2024-04-29
SES: Adapt Native XS Compartment to SES Compartment


kriskowal:
Refs: #400 

Add a Compartment adapter to `ses` such that the SES `Compartment` wraps the XS native `Compartment` providing parity with the current `ses` `Compartment` interface using the XS native implementation.

- One of:
      - SES: Introduce XS-specific Compartment interface-parity-adapter around native Compartment such that existing usage requires no changes.
      - Refactor SES Compartment to implement the XS Native Compartment API more directly and adjust Compartment Mapper to use these new forms.
      - Refactor Compartment Mapper such that it drives `ModuleSource` and `Module` constructors directly and subsumes some dependency analysis responsibilities from `Compartment`.

PR #2321, opened 2024-06-13
refactor(ses): Retool for module descriptors

kriskowal:
Refs: #400 #2251 

## Description

Pursuant to arriving at parity with XS, this change internally reörients the SES module loader around the concept of a “module descriptor”. This creates some clarity and removes some existing duplication of work for “alias” module descriptors, and prepares the material for a richer module descriptor schema to match those implemented by XS.

### Security Considerations

No changes.

### Scaling Considerations

No changes.

No changes.

### Testing Considerations

No changes.

No changes.

### Upgrade Considerations

No changes.

### DIFF
--- a/packages/ses/src/compartment.js
+++ b/packages/ses/src/compartment.js
@@ -10,6 +10,7 @@ import {
   assign,
   defineProperties,
   entries,
+  isObject,
   promiseThen,
   toStringTagSymbol,
   weakmapGet,
@@ -207,21 +208,36 @@ export const makeCompartmentConstructor = (
     // moduleMap can be invalid in ways that cannot be detected in the
     // constructor, but these checks allow us to throw early for a better
     // developer experience.
-    for (const [specifier, aliasNamespace] of entries(moduleMap || {})) {
-      if (typeof aliasNamespace === 'string') {
+    for (const [specifier, moduleDescriptor] of entries(moduleMap || {})) {
+      if (typeof moduleDescriptor === 'string') {
         // TODO implement parent module record retrieval.
         throw TypeError(
           `Cannot map module ${q(specifier)} to ${q(
-            aliasNamespace,
+            moduleDescriptor,
           )} in parent compartment`,
         );
-      } else if (weakmapGet(moduleAliases, aliasNamespace) === undefined) {
-        // TODO create and link a synthetic module instance from the given
-        // namespace object.
-        throw ReferenceError(
-          `Cannot map module ${q(
-            specifier,
-          )} because it has no known compartment in this realm`,
+      } else if (isObject(moduleDescriptor)) {
+        if (weakmapGet(moduleAliases, moduleDescriptor) !== undefined) {
+          // No further validation for module namespace objects.
+          // They do bear inspection before evaluation gracefully.
+        } else if (moduleDescriptor.specifier !== undefined) {
+          // TODO validate specifier module descriptors
+        } else if (moduleDescriptor.record !== undefined) {
+          // TODO validate record module descriptors
+        } else if (moduleDescriptor.__syncModuleProgram__ !== undefined) {
+          // TODO validate virtual module source descriptors
+        } else if (moduleDescriptor.execute !== undefined) {
+          // TODO validate module source descriptors
+        } else if (weakmapGet(moduleAliases, moduleDescriptor) === undefined) {
+          throw ReferenceError(
+            `Cannot map module ${q(
+              specifier,
+            )} because it has no known compartment in this realm`,
+          );
+        }
+      } else {
+        throw TypeError(
+          `Value for key ${q(specifier)} in module map must be a string, module descriptor object, module source object, or virtual module namespace`,
         );
       }
     }
--- a/packages/ses/src/compartment.js
+++ b/packages/ses/src/compartment.js
@@ -42,19 +42,6 @@ const moduleAliases = new WeakMap();
 // privateFields captures the private state for each compartment.
 const privateFields = new WeakMap();
 
-// Compartments do not need an importHook or resolveHook to be useful
-// as a vessel for evaluating programs.
-// However, any method that operates the module system will throw an exception
-// if these hooks are not available.
-const assertModuleHooks = compartment => {
-  const { importHook, resolveHook } = weakmapGet(privateFields, compartment);
-  if (typeof importHook !== 'function' || typeof resolveHook !== 'function') {
-    throw TypeError(
-      'Compartment must be constructed with an importHook and a resolveHook for it to be able to load modules',
-    );
-  }
-};
-
 export const InertCompartment = function Compartment(
   _endowments = {},
   _modules = {},
@@ -111,8 +98,6 @@ export const CompartmentPrototype = {
       throw TypeError('first argument of module() must be a string');
     }
 
-    assertModuleHooks(this);
-
     const { exportsProxy } = getDeferredExports(
       this,
       weakmapGet(privateFields, this),
@@ -127,8 +112,6 @@ export const CompartmentPrototype = {
     if (typeof specifier !== 'string') {
       throw TypeError('first argument of import() must be a string');
     }
-
-    assertModuleHooks(this);
 
     return promiseThen(
       load(privateFields, moduleAliases, this, specifier),
@@ -149,8 +132,6 @@ export const CompartmentPrototype = {
       throw TypeError('first argument of load() must be a string');
     }
 
-    assertModuleHooks(this);
-
     return load(privateFields, moduleAliases, this, specifier);
   },
 
@@ -159,7 +140,6 @@ export const CompartmentPrototype = {
       throw TypeError('first argument of importNow() must be a string');
     }
 
-    assertModuleHooks(this);
     loadNow(privateFields, moduleAliases, this, specifier);
     return compartmentImportNow(/** @type {Compartment} */ (this), specifier);
   },
--- a/packages/ses/src/compartment.js
+++ b/packages/ses/src/compartment.js
@@ -4,13 +4,10 @@
 
 import {
   Map,
-  ReferenceError,
   TypeError,
   WeakMap,
   assign,
   defineProperties,
-  entries,
-  isObject,
   promiseThen,
   toStringTagSymbol,
   weakmapGet,
@@ -26,11 +23,8 @@ import { sharedGlobalPropertyNames } from './permits.js';
 import { load, loadNow } from './module-load.js';
 import { link } from './module-link.js';
 import { getDeferredExports } from './module-proxy.js';
-import { assert } from './error/assert.js';
 import { compartmentEvaluate } from './compartment-evaluate.js';
 import { makeSafeEvaluator } from './make-safe-evaluator.js';
-
-const { quote: q } = assert;
 
 // moduleAliases associates every public module exports namespace with its
 // corresponding compartment and specifier so they can be used to link modules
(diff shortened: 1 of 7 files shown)


packages/ses/src/module-link.js
```diff
@@ -129,33 +123,31 @@ export const instantiate = (
...
       moduleAliases,
       moduleSpecifier,
       resolvedImports,
     );
   } else {
     throw TypeError(
-      `importHook must return a static module record, got ${q(
-        staticModuleRecord,
-      )}`,
+      `importHook must return a static module record, got ${q(moduleSource)}`,
```
erights, line 150: Did you mean?

```suggestion
      `importHook must return a static module source, got ${q(moduleSource)}`,
```

kriskowal: I’ve added a commit that updates all the terminology:

- “module records” now always refer to the internal record for a module instance
- “static module records” are now “module sources”
- “third party static module records” are now “virtual module sources“

This generally makes “module record” much less confusing, so a clarifying comment is no longer necessary at the head of `module-load.js` and `module-link.js`.

packages/ses/src/compartment.js
```diff
@@ -227,21 +208,36 @@ export const makeCompartmentConstructor = (
...
-      } else if (weakmapGet(moduleAliases, aliasNamespace) === undefined) {
-        // TODO create and link a synthetic module instance from the given
-        // namespace object.
-        throw ReferenceError(
-          `Cannot map module ${q(
-            specifier,
-          )} because it has no known compartment in this realm`,
+      } else if (isObject(moduleDescriptor)) {
+        if (weakmapGet(moduleAliases, moduleDescriptor) !== undefined) {
+          // No furthere validation for module namespace objects.
```
erights, line 221:
```suggestion
          // No further validation for module namespace objects.
```
unless it is an obscure pun I'm missing ;)

packages/ses/src/compartment.js
```diff
@@ -227,21 +208,36 @@ export const makeCompartmentConstructor = (
...
+          // They do bear inspection before evaluation gracefully.
```
erights, line 222: Did you mean?
```suggestion
          // They do bear inspection before evaluating gracefully.
```

erights (commented):
I probably need a live discussion before I'm knowledgeable enough to complete the review. But a lot already LGTM

packages/ses/src/module-link.js
erights: Resolving

packages/ses/src/compartment.js
kriskowal: I think this is grammatical as written, but always down for a second or third opinion. @gibson042 how’s your grammar?

kriskowal: (I went with evaluation because it agrees with inspection if that’s consequential.)

kriskowal: (Pardon, don’t waste too much time on this. I delete this entire section in a planned change. I might move that up to make that clear.)

packages/ses/src/module-load.js
```diff
@@ -145,83 +139,72 @@ function* loadWithoutErrorAnnotation(
   const { importHook, importNowHook, moduleMap, moduleMapHook, moduleRecords } =
     weakmapGet(compartmentPrivateFields, compartment);
 
+  if (mapHas(moduleRecords, moduleSpecifier)) {
```
kriskowal, line 142: The meat of this change begins here thru the end of `loadWithErrorAnnotation`, which preserves all prior behaviors, fixes a bug or two, and prepares to add new module descriptor shapes in order to be more compatible with usage of XS.

This is preparation to align SES with XS module descriptors https://github.com/Moddable-OpenSource/moddable/blob/public/documentation/xs/XS%20Compartment.md#-module-descriptor

packages/ses/src/module-load.js
```diff
@@ -145,83 +139,72 @@ function* loadWithoutErrorAnnotation(
...
-        ReferenceError,
-      );
+  } else if (isObject(moduleDescriptor)) {
+    // In this shim (and not in XS, and not in the standard we imagine), we
+    // allow a module namespace object to stand in for a module descriptor that
+    // describes its original {compartment, specifier} so that it can be used
+    // to create a link.
+    const aliasDescriptor = weakmapGet(moduleAliases, moduleDescriptor);
+    if (aliasDescriptor !== undefined) {
+      moduleDescriptor = aliasDescriptor;
```
kriskowal, line 185: This preserves legacy support for importHooks that return a namespace object via `compartment.module(specifier)`, translating that to the corresponding `{ specifier, compartment }` descriptor.

packages/ses/src/module-load.js
```diff
@@ -145,83 +139,72 @@ function* loadWithoutErrorAnnotation(
...
-          'Cannot redirect to an explicit record with a specified compartment',
-        );
-      }
+    // A (legacy) module descriptor for when we find the module source (record)
+    // but at a different specifier than requested.
+    // Providing this {specifier, record} descriptor serves as an ergonomic
+    // short-hand for stashing the record, returning a {compartment, specifier}
+    // reference, bouncing the module hook, then producing the source (record)
+    // when module hook receives the response specifier.
+    if (moduleDescriptor.record !== undefined) {
```
kriskowal, line 194: Descriptors with the `record` property are spiritually equivalent to the XS `source` module descriptor, to be introduced in the subsequent PR. The only behavior `record` descriptors support is the creation of a module that has a different “response specifier” than the “request specifier”. That is, `importHook(requestSpecifier)` returns `{ record: moduleSource, specifier: responseSpecifier }`, such that all the import specifiers in the source are resolved relative to responseSpecifier (e.g., `./src/index.js`) instead of the requestSpecifier (e.g., `.`).

XS `source` module descriptors will support more shapes and treat the `compartment` default differently. The `source` may be a string instead of a module source, in which case it’s a specifier to load the source from another compartment. The default compartment is the parent instead of the child.

packages/ses/src/module-load.js
```diff
@@ -231,21 +214,26 @@ function* loadWithoutErrorAnnotation(
...
-      // ensure expected record shape
-      if (staticModuleRecord.importMeta !== undefined) {
-        throw TypeError(
-          'Cannot redirect to an implicit record with a specified importMeta',
-        );
+    // A (legacy) module descriptor that describes a link to a module instance
+    // in a specified compartment.
+    if (
+      moduleDescriptor.compartment !== undefined &&
+      moduleDescriptor.specifier !== undefined
```
kriskowal, line 221: The legacy module descriptors with { compartment, specifier } correspond to { namespace, compartment } descriptors in XS, where the namespace is a string, the specifier. XS also allows the namespace to be a literal namespace or an ordinary object. If it’s an ordinary object, XS virtualizes a module to export its properties. We do not support any of the non-string cases in “specifier descriptors”.

packages/ses/src/module-load.js
```diff
@@ -254,19 +242,24 @@ function* loadWithoutErrorAnnotation(
       return aliasRecord;
     }
 
-    throw TypeError('Unnexpected RedirectStaticModuleInterface record shape');
+    // A (legacy) behavior: If we do not recognize the module descriptor as a
+    // module descriptor, we assume that it is a module source (record):
```
kriskowal, line 246: At some point, we’ll want to add options for strict portability and this case will become an exception in that mode. We would stop trying to differentiate module sources from module descriptors.

packages/ses/test/import.test.js
```diff
@@ -4,10 +4,264 @@
 /* eslint max-lines: 0 */
 
 import test from 'ava';
+import { StaticModuleRecord } from '@endo/static-module-record';
 import '../index.js';
 import { resolveNode, makeNodeImporter } from './node.js';
 import { makeImporter, makeStaticRetriever } from './import-commons.js';
 
+test('module map primed with module source', async t => {
```
kriskowal, line 12: The individual commits are framed to show which of these tests were failing before this refactor.

packages/ses/src/module-load.js
erights: I don't see a `loadWithErrorAnnotation`.

Did you mean `loadWithoutErrorAnnotation` or `memoizedLoadWithErrorAnnotation`?

kriskowal: Pardon, the surrounding function, `loadWithoutErrorAnnotation`.

packages/ses/src/compartment.js
erights: Resolving

packages/ses/src/compartment.js
erights: Resolving

packages/ses/src/module-load.js
erights: Resolving

packages/ses/src/module-load.js
erights: I'm fine with the code here in this PR. But I don't understand the comment. Why not continue to differentiate module sources from module descriptors?

erights (approved):
Sorry it took so long.

LGTM, thanks!

packages/ses/src/module-load.js
```diff
@@ -145,83 +139,72 @@ function* loadWithoutErrorAnnotation(
...
+      const aliasRecord = loadModuleSource(
         compartmentPrivateFields,
         moduleAliases,
         aliasCompartment,
         aliasSpecifier,
-        aliasModuleRecord,
+        moduleSource,
         enqueueJob,
         selectImplementation,
         moduleLoads,
```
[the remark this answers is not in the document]
kriskowal: Only that refactoring to use a named option bag has never been urgent. It’s been desirable for a while. Also adding TS comments to this (pretty old) module.

packages/ses/src/module-load.js
kriskowal: No, we’ve no intention of proposing any shape of `{specifier, compartment}` module descriptor. The motivating cases are fully subsumed into `{namespace, compartment?}` descriptors.

packages/ses/src/module-load.js
kriskowal: We’ve discriminated all the module descriptors up to this point, but to preserve backward compatibility with import hooks that simply returned a module source (not a module source wrapped in a module descriptor), we have to fall through here.

We do have a clear path to deprecate this path, though, which I hope to get to separately.

packages/ses/src/module-load.js
```diff
@@ -145,83 +139,72 @@ function* loadWithoutErrorAnnotation(
...
+    // promises.
+    if (!weakmapHas(moduleAliases, moduleDescriptor)) {
+      moduleDescriptor = yield moduleDescriptor;
+    }
   }
-  if (typeof aliasNamespace === 'string') {
+
+  if (typeof moduleDescriptor === 'string') {
     // eslint-disable-next-line @endo/no-polymorphic-call
     assert.fail(
```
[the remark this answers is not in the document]
kriskowal: This presumably means that the type isn’t getting properly threaded, since `fail` and `Fail` return `never`. I’ll change this to `throw assert.error` since throwing at an inevitably throwing call feels weird.

packages/ses/src/module-load.js
erights: Resolving

packages/ses/src/module-load.js
erights: > which I hope to get to separately

Worth filing an issue?

packages/ses/src/module-load.js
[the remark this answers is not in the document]
erights: It felt weird for me the first 100 times. But our codebase now does that pervasively. In `throw Fail...` I read the `Fail` as "throw an error" and the `throw` as "Let static analysis know that we're throwing an error".

packages/ses/src/module-load.js
```diff
@@ -254,19 +242,24 @@ function* loadWithoutErrorAnnotation(
...
-    compartment,
-    moduleSpecifier,
-    staticModuleRecord,
-    enqueueJob,
-    selectImplementation,
-    moduleLoads,
-  );
 }
 
 const memoizedLoadWithErrorAnnotation = (
```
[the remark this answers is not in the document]
erights: Resolving.

packages/ses/src/module-load.js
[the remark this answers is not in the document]
erights: Thanks. Since there was no change requested, just a question you answered, Resolving.

PR #2324, opened 2024-06-16
refactor: use @endo/errors when acyclic

erights:
Staged on #2323 

Closes: #XXXX
Refs: https://github.com/Agoric/agoric-sdk/issues/5672 https://github.com/Agoric/agoric-sdk/pull/9513

## Description

Does for endo what https://github.com/Agoric/agoric-sdk/pull/9513 does for agoric-sdk --- importing `assert` from @endo/errors --- when it can do so without causing dependency cycles. For the remaining packages that would cause dependency cycles, just move them towards more modern assert conventions while still using the unstructured global `assert`

### Security Considerations

none
### Scaling Considerations

none

We should document `assert` only as imported from @endo/errors, since our users will not generally contribute code within endo's internal dependency cycles.

### Testing Considerations

The CI run on this PR also serves to test #2323, since that one, by itself, does not cause a behavioral change. It's purpose is to enable this PR not to hit the problem described at https://github.com/Agoric/agoric-sdk/issues/9515

none
### Upgrade Considerations

none

### DIFF
--- a/packages/ses/types.d.ts
+++ b/packages/ses/types.d.ts
@@ -346,12 +346,10 @@ export interface AssertionUtilities {
    * ```js
    * assert(sky.isBlue(), details`${sky.color} should be "blue"`);
    * ```
-   * // TODO Update SES-shim to new convention, where `details` is
-   * // renamed to `X` rather than `d`.
-   * or following the normal convention to locally rename `details` to `d`
-   * and `quote` to `q` like `const { details: d, quote: q } = assert;`:
+   * or following the normal convention to locally rename `details` to `X`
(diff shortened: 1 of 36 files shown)


packages/ses/types.d.ts
```diff
@@ -346,12 +346,10 @@ export interface AssertionUtilities {
    * ```js
    * assert(sky.isBlue(), details`${sky.color} should be "blue"`);
    * ```
-   * // TODO Update SES-shim to new convention, where `details` is
```
erights, line 349: TODO done

kriskowal (approved):

PR #2325, opened 2024-06-20
Further endo daemon CLI support for dot-delimited paths

FUDCo:
Adds support for dot-delimited petname paths in the `mkdir`, `remove`, and `spawn` commands, plus everywhere there's a `--name` option (the `install`, `make`, `request`, `adopt`, `store`, `eval`, and `bundle` commands).

Ref: #2023

### DIFF
--- a/packages/cli/src/commands/adopt.js
+++ b/packages/cli/src/commands/adopt.js
@@ -2,6 +2,7 @@
 import os from 'os';
 import { E } from '@endo/far';
 import { withEndoAgent } from '../context.js';
+import { parsePetNamePath } from '../pet-name.js';
 
 export const adoptCommand = async ({
   messageNumberText,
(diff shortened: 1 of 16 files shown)


FUDCo:
Something for the TODO list that came up in the doing of this PR: The switchover to allowing paths where only simple names had previously been supported forced some of the functions that accept a path in the form of a variadic argument list to be converted to take the path in the form of an array of path elements, while other functions remain using the variadic form.  The resulting mixture is quite confusing.  For the sake of cleanliness, consistency, and legibility (not to mention making parameter types less mysterious), all the variadic functions should be converted to the array form.  The fact that this change is not completely trivial to execute highlights the need for it.

kriskowal:
I agree all methods should take the array form. Except `lookup` because I don’t make the rules for hub protocol (@dckc).

kriskowal (approved):
I’m fine with this as-is, but there’s a chance we’ll have to loosen up `lookup` to accept variadic pet name paths again because that’s the protocol of name hubs on the Agoric chain.

dckc:
The rules for the `lookup` protocol pre-date me. @erights , @michaelfig ? thoughts?

michaelfig:
> I’m fine with this as-is, but there’s a chance we’ll have to loosen up `lookup` to accept variadic pet name paths again because that’s the protocol of name hubs on the Agoric chain.

I don't see how that is a requirement: why would the current version of "name hubs on the Agoric chain" be relevant to the daemon's long-term API?  The daemon's pet name system should work with any other name system, using adapters as necessary to be agnostic.

PR #2327, opened 2024-06-21
use Yarn's pnpm linker

turadg:
- **build: bump yarn to 4.3.0**
- **build: use pnpm linker**

Closes: #XXXX
Refs: #XXXX

## Description

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,1 +1,1 @@
-nodeLinker: node-modules
+nodeLinker: pnpm
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "engines": {
     "node": ">=16"
   },
-  "packageManager": "yarn@4.2.2",
+  "packageManager": "yarn@4.3.0",
   "devDependencies": {
     "@jessie.js/eslint-plugin": "^0.4.1",
     "@octokit/core": "^3.4.0",


turadg:
@kriskowal any ideas on how to get `test262-runner` to build? I tried `yarn add eshost` but it still couldn't find the dependencies that that package brought in. Seems something specific to how `test262-harness` works.

kriskowal:
1. I love this change. I love everything about it, but especially revealing ghost dependencies. I hold Ryah personally accountable for allowing ghost dependencies. Hi, Ryah.
2. I might delete test262-runner soon. I have a hardened262 package in my stack for #400 which will be up for review two reviews currently in flight. So, maybe this goes away with no more effort from you.
3. The problem is not a ghost dependency from test262-harness on eshost https://github.com/tc39/test262-harness/blob/master/package.json
4. That’s odd. I’ll have to circle back and attempt to reproduce the issue locally, and poke around in the `node_modules` tree to see why it can’t find `node_modules/eshost` from any ancestor directory from the realpath for `test262-harness`.

.github/workflows/ci.yml
```diff
@@ -54,36 +54,6 @@ jobs:
...
-
-      # end macro
-
-      - name: Run yarn build
-        run: yarn build
-
-      # build the API docs in markdown for agoric/documentation repo to verify it works
-      - name: build API docs in markdown
-        run: yarn docs:markdown-for-agoric-documentation-repo
-
```
kriskowal, line 86: @turadg I’m assuming you know better than I why this was present and why we don’t need it anymore.

(I don’t know either answer. If you don’t, let’s find out who does.)

packages/captp/test/gc.test.js
```diff
@@ -29,31 +29,22 @@ test('test loopback gc', async t => {
...
+  t.true(nearSendDrop >= 3);
+  t.true(nearRecvDrop >= 2);
+  t.true(nearDropped >= 2);
+  t.true(farSendDrop >= 2);
+  t.true(farRecvDrop >= 3);
+  t.true(farDropped >= 3);
+  t.is(nearSendDrop, farRecvDrop);
+  t.is(nearRecvDrop, farSendDrop);
+  t.is(nearDropped, nearRecvDrop);
+  t.is(farDropped, farRecvDrop);
```
kriskowal, line 49: This is a plausible fix for the flakes. They will not be missed.

packages/captp/package.json
```diff
@@ -52,7 +52,9 @@
     "@endo/init": "workspace:^",
     "@endo/ses-ava": "workspace:^",
     "ava": "catalog:dev",
+    "c8": "^7.14.0",
```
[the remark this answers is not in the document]
kriskowal: Good suggestion.

packages/check-bundle/package.json
```diff
@@ -47,13 +47,20 @@
...
     "@endo/ses-ava": "workspace:^",
     "@endo/zip": "workspace:^",
     "ava": "catalog:dev",
     "babel-eslint": "^10.1.0",
     "c8": "catalog:dev",
     "eslint": "catalog:dev",
+    "eslint-config-airbnb-base": "^15.0.0",
+    "eslint-config-prettier": "^9.1.0",
+    "eslint-plugin-eslint-comments": "^3.2.0",
+    "eslint-plugin-import": "^2.31.0",
```
kriskowal, line 61: I don’t think these are in the catalog but maybe we should do that, but not necessarily in this change.

packages/compartment-mapper/test/_can-bundle-init.js
```diff
@@ -1,2 +1,2 @@
-// eslint-disable-next-line import/no-extraneous-dependencies
+// eslint-disable-next-line import/no-extraneous-dependencies,import/no-unresolved
```
kriskowal: What’s not resolved properly? Can we resolve it?

packages/eslint-plugin/lib/rules/no-assign-to-exported-let-var-or-function.js
```diff
@@ -200,7 +200,7 @@ module.exports = {
         }
 
         // Case 2: `export { local as exported }` (only when source == null; otherwise it's a re-export)
-        if (!node.source && node.specifiers && node.specifiers.length > 0) {
+        if (!node.source && node.specifiers && node.specifiers[0]) {
```
kriskowal, line 203: I probably would have been pedantic and checked the length before indexing, but maybe that’s because I wrote C that one time.

packages/init/test/async_hooks.test.js
```diff
@@ -3,7 +3,7 @@
 
 // Use a package self-reference to go through the "exports" resolution
 // Use the debug-async-hooks entrypoint which includes the async_hooks patch
-// eslint-disable-next-line import/no-extraneous-dependencies
+// eslint-disable-next-line import/no-extraneous-dependencies,import/no-unresolved
```
kriskowal: What didn’t resolve? Can we add it to `package.json` without introducing a cycle? If not, can we move this test to a package that doesn’t participate in the cycle?

packages/module-source/shim.js
```diff
@@ -3,7 +3,7 @@
 // We are using a reflexive import to make sure we pass through the conditional
 // export in package.json.
 // Eslint does not yet seem to have a carve-out for package-reflexive imports.
-// eslint-disable-next-line import/no-extraneous-dependencies
+// eslint-disable-next-line import/no-extraneous-dependencies,import/no-unresolved
```
kriskowal: This must be because of the reflexive import, and reflexive imports go through the `"exports"` table, so probably should remain reflexive.

Perhaps these are happening because of a difference in behavior, where pnpm link doesn’t present a reflexive symbolic link.

packages/pass-style/tsconfig.json
```diff
@@ -4,6 +4,7 @@
     "checkJs": true,
   },
   "include": [
+    "index.js",
     "*.js",
     "*.ts",
```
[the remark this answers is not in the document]
kriskowal: Good suggestion.

packages/pass-style/typedoc.json
```diff
@@ -2,7 +2,7 @@
     "extends": [
         "../../typedoc.base.json"
     ],
-    "entryPoints": [
-        "index.js"
-    ]
+  "entryPoints": [
+    "index.js"
+  ]
```
kriskowal: This is a formatting regression.

packages/ses-ava/package.json
```diff
@@ -46,14 +46,11 @@
     "@endo/env-options": "workspace:^",
     "@endo/harden": "workspace:^",
     "@endo/init": "workspace:^",
+    "ava": "catalog:dev",
     "ses": "workspace:^"
   },
```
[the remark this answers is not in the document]
kriskowal: This is a good decision. Packages that depend on `ses-ava` should also manually depend on their choice of `ava` version.

packages/ses/scripts/hermes-test.sh
```diff
@@ -27,8 +27,8 @@ case "$OS" in
         ;;
 esac
 
-HERMESC="../../node_modules/hermes-engine-cli/$OS_DIR/hermesc"
-HERMES="../../node_modules/hermes-engine-cli/$OS_DIR/hermes"
+HERMESC="node_modules/hermes-engine-cli/$OS_DIR/hermesc"
+HERMES="node_modules/hermes-engine-cli/$OS_DIR/hermes"
```
kriskowal, line 31: Cool.

kriskowal (changes requested):
I’m not sure what we can or should do about the no-unresolved-imports, but I would like to know more about them and try to fix them another way. If that proves impossible, I would like the lint exception to be a per-line annotation so we have a better grasp of why they’re there and so we don’t lose coverage of genuine missing entries in package.json.

I am hoping there’s a way to preserve the ses-ava peerDependency on ava, but if there isn’t I can be swayed to co-version ses-ava with the underlying ava. That just has some release management implications and might require us to maintain concurrent versions during an ava major version bump, which we’ve survived once with this mitigation. I think the solution is to just add explicit dependency on ava in packages where that is lacking.

Otherwise, excited to try the new format out.

.github/workflows/ci.yml
turadg: yeah I think this is obsolete now that we aren't updating the Agoric SDK docs site. cc @LuqiPan

turadg: we will have to remove the required check before merging. I take it you're ok with that.

turadg:
>I’m not sure what we can or should do about the no-unresolved-imports, but I would like to know more about them and try to fix them another way. If that proves impossible, I would like the lint exception to be a per-line annotation so we have a better grasp of why they’re there and so we don’t lose coverage of genuine missing entries in package.json.

It's just a deficiency of eslint-plugin-import. I've put a config level work-around so we don't need line suppressions.

> I am hoping there’s a way to preserve the ses-ava peerDependency on ava, but if there isn’t I can be swayed to co-version ses-ava with the underlying ava. That just has some release management implications and might require us to maintain concurrent versions during an ava major version bump, which we’ve survived once with this mitigation. I think the solution is to just add explicit dependency on ava in packages where that is lacking.

What failure mode are you trying to avoid? I think it’s real and more accurate that ses-ava depends on Ava. We can set the version range to whatever versions it can work with. I think the "catalog:" change was a mistake so I put it back to the original version range.

packages/module-source/package.json
```diff
@@ -76,6 +77,16 @@
...
       "plugin:@endo/internal"
     ],
+    "rules": {
+      "import/no-unresolved": [
+        "error",
+        {
+          "ignore": [
+            "^@endo/module-source(?:/.*)?$"
+          ]
+        }
```
kriskowal, line 87: Neat.

packages/ses-ava/package.json
```diff
@@ -46,14 +46,11 @@
     "@endo/env-options": "workspace:^",
     "@endo/harden": "workspace:^",
     "@endo/init": "workspace:^",
+    "ava": "^5.3.0 || ^6.1.2",
```
kriskowal, line 49: Works for me.

PR #2330, opened 2024-06-22
fix(compartment-mapper): Support ESM imports defined property from CJ…

kriskowal:
…S from set property of CJS

Refs: https://github.com/Agoric/agoric-sdk/issues/9408

## Description

If a modern JavaScript module imports a named binding from a CommonJS module, we can hardly guarantee that it will always work. Outrageous heroics make it work under most circumstances using a heuristic lexical analysis of the CommonJS module to determine which names should exist on the module’s internal namespace. Then, the CommonJS runtime must account for all variety of shenanigans that might cause the binding for that name to change. These include assignment to property names on `exports`, reassignment to `module.exports`, and `defineProperty` on the _original_ `exports` object. These changes must be reflected both on the module namespace object and on the `default` property thereof in _in every case where this is in fact possible_.

Consider a CJS module (2.cjs) that exports a name:

```js
exports.meaning = 42;
```

Then, consider a TypeScript module (1.ts) that reexports this name:

```ts
export { meaning } from './2.cjs';
```

Which gets compiled down to a CommonJS module (1.cjs):

```js
const universe = require('./1.cjs');
Object.defineProperty(exports, 'meaning', {
  enumerable: true,
  get() {
    return universe.meaning
  }
});
```

Which gets imported by an ESM by name (0.mjs):

```js
import { meaning } from './1.cjs';
```

This has no right to work. We can trivially copy the getter from 1.cjs to the `default` export but the internal namespace object we use to emulate ESM in SES emits live binding notifications for every named property and doesn’t suffer `defineProperty` well. So, in the CommonJS emulation, we must attempt to propagate the current value from the getter to the module environment record after the module initializes.

### Security Considerations

None.

### Scaling Considerations

None.

None. Folks already assume this works. It works in Node.js. Code exists that relies on it.

### Testing Considerations

We have an existing test that exercises the case that the getter under a defineProperty can’t be expected to succeed on the stack of defineProperty.

This increases ecosystem compatibility.

### Upgrade Considerations

None.

### PS

It didn’t have to be like this.

### DIFF
--- a/packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
+++ b/packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
@@ -103,6 +103,8 @@ export const wrap = ({
     }
   };
 
+  const redefined = new Set();
+
   const originalExports = new Proxy(moduleEnvironmentRecord.default, {
     set(_target, prop, value) {
       promoteToNamedExport(prop, value);
@@ -121,6 +123,9 @@ export const wrap = ({
       // descriptor.get in the trap will cause an error.
       // Object.defineProperty is used instead of Reflect.defineProperty for better error messages.
       defineProperty(target, prop, descriptor);
+      // Make a note to propagate the value from the getter to the module
+      // environment record when the module finishes executing.
+      redefined.add(prop);
       return true;
     },
   });
@@ -176,8 +181,13 @@ export const wrap = ({
     if (exportsHaveBeenOverwritten) {
       moduleEnvironmentRecord.default = finalExports;
       for (const prop of keys(moduleEnvironmentRecord.default || {})) {
-        if (prop !== 'default')
+        if (prop !== 'default') {
           moduleEnvironmentRecord[prop] = moduleEnvironmentRecord.default[prop];
+        }
+      }
+    } else {
+      for (const prop of redefined) {
+        moduleEnvironmentRecord[prop] = moduleEnvironmentRecord.default[prop];
       }
     }
   };
--- a/packages/compartment-mapper/test/esm-imports-cjs-define.test.js
+++ b/packages/compartment-mapper/test/esm-imports-cjs-define.test.js
@@ -1,0 +1,23 @@
+/* eslint-disable no-underscore-dangle */
+// import "./ses-lockdown.js";
+import 'ses';
+import test from 'ava';
+
+import { scaffold } from './scaffold.js';
+
+const fixture = new URL(
+  'fixtures-esm-imports-cjs-define/0.mjs',
+  import.meta.url,
+).toString();
+
+const assertFixture = t => t.pass();
+
+const fixtureAssertionCount = 1;
+
+scaffold(
+  'fixtures-esm-imports-cjs-define',
+  test,
+  fixture,
+  assertFixture,
+  fixtureAssertionCount,
+);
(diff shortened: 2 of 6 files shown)


naugtur:
I think we should add this case to the matrix in https://github.com/endojs/endo-e2e-tests (along with some new runtimes BTW) and see how it compares.

Also, I wonder if there's a noticeable change in the e2e test results form this. I remember at the time the cjs implementation was passing e2e with few exports missing and some exports being redundant. 

One thing we're guaranteed to never fix is redundant default. I hope that never becomes an issue.

Anyway, I think a run of e2e on main vs this could be informative.

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
```diff
@@ -176,10 +181,14 @@ export const wrap = ({
     if (exportsHaveBeenOverwritten) {
       moduleEnvironmentRecord.default = finalExports;
       for (const prop of keys(moduleEnvironmentRecord.default || {})) {
-        if (prop !== 'default')
+        if (prop !== 'default') {
```
boneskull, line 184: (suggestion) have a

```js
export const DEFAULT = 'default';
```

somewhere and use it

packages/compartment-mapper/test/esm-imports-cjs-define.test.js
```diff
@@ -0,0 +1,23 @@
...
+import test from 'ava';
+
+import { scaffold } from './scaffold.js';
+
+const fixture = new URL(
+  'fixtures-esm-imports-cjs-define/0.mjs',
+  import.meta.url,
+).toString();
+
+const assertFixture = t => t.pass();
```
boneskull, line 13: this means "if this fixture does not error, we're good"?

kriskowal:
@naugtur Can I beg a ruling on this change?

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
```diff
@@ -176,10 +181,14 @@ export const wrap = ({
...
           moduleEnvironmentRecord[prop] = moduleEnvironmentRecord.default[prop];
+        }
       }
     }
+    for (const prop of redefined) {
+      moduleEnvironmentRecord[prop] = moduleEnvironmentRecord.default[prop];
```
naugtur, line 190: Took me a while to get to this conclusion, but the only way this covers anything that was not covered before in the for loop above is if a prop is not enumerable. Otherwise it's already been looped through in the same synchronous run.
How is this changing anything?

(yes, I'm gonna go run the tests now)

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
```diff
@@ -176,10 +181,14 @@ export const wrap = ({
     if (exportsHaveBeenOverwritten) {
       moduleEnvironmentRecord.default = finalExports;
       for (const prop of keys(moduleEnvironmentRecord.default || {})) {
-        if (prop !== 'default')
+        if (prop !== 'default') {
           moduleEnvironmentRecord[prop] = moduleEnvironmentRecord.default[prop];
+        }
       }
     }
+    for (const prop of redefined) {
```
naugtur, line 189: there could be an `else` before this because redefined doesn't make a lot of sense when the default export was redefined. but it should also be an empty set in that case.

naugtur (approved):

PR #2333, opened 2024-06-24
ci: Mitigate Playwright failure in CI

kriskowal:
The Selenium Docker image with Chrome nightly builds is no longer sufficient for Playwright CI for lack of some system dependencies. This change mitigates the problem by walking back to the default container image until we can prepare a working Docker image for Chrome nightlies.

### DIFF
--- a/.github/workflows/browser-test.yml
+++ b/.github/workflows/browser-test.yml
@@ -10,13 +10,19 @@ jobs:
   browser-tests:
     timeout-minutes: 30
     runs-on: ubuntu-latest
-    container:
-      image: selenium/standalone-chrome:nightly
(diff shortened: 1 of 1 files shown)


naugtur:
Assuming we don't want to maintain our own image in a github-backed registry and making sure it stays on top of nightly automatically, I think the best option for us here is to do one of:
- find a better image that has chrome nightly and works
- add a step to the pipeline that installs the missing stuff 

This is good as a temporary workaround.

naugtur (approved):

naugtur:
Also, the whole issue might be a temporary misalignment between playwright and the OS distro.

PR #2337, opened 2024-06-28
feat(ses): call lockdown before bundling SES shim

leotm:
## Description

Call lockdown before bundling the SES shim

Discussion: https://youtu.be/qGGeACz4cyM

### Security Considerations

- bundle now fails (TypeError) to build if non-standard JS called after `lockdown()`
- `yarn build` now also logs to console (e.g. `Removing unpermitted intrinsics`) if non-standard JS called before `lockdown()`

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations

### DIFF
--- a/packages/ses/scripts/bundle.js
+++ b/packages/ses/scripts/bundle.js
@@ -2,6 +2,7 @@
 /* global process */
 
 import '../index.js';
+import '../lockdown.js';
 
 lockdown();
 
--- a/packages/ses/scripts/bundle.js
+++ b/packages/ses/scripts/bundle.js
@@ -2,10 +2,7 @@
 /* global process */
 
 import '../index.js';
-import '../lockdown.js';
-
-lockdown();
-
+import '../test/lockdown-safe.js';
 import fs from 'fs';
 import { makeBundle } from '@endo/compartment-mapper/bundle.js';
 import { minify } from 'terser';
--- a/packages/ses/scripts/bundle.js
+++ b/packages/ses/scripts/bundle.js
@@ -1,10 +1,7 @@
 /* eslint-disable import/first */
 /* global process */
-
-import '../index.js';
-
 lockdown();
-
+import '../index.js';
 import fs from 'fs';
 import { makeBundle } from '@endo/compartment-mapper/bundle.js';
 import { minify } from 'terser';
--- a/packages/ses/scripts/bundle.js
+++ b/packages/ses/scripts/bundle.js
@@ -1,5 +1,10 @@
+/* eslint-disable import/first */
 /* global process */
+
 import '../index.js';
+
+lockdown();
+
 import fs from 'fs';
 import { makeBundle } from '@endo/compartment-mapper/bundle.js';
 import { minify } from 'terser';
--- a/packages/ses/scripts/bundle.js
+++ b/packages/ses/scripts/bundle.js
@@ -1,7 +1,10 @@
 /* eslint-disable import/first */
 /* global process */
-lockdown();
+
 import '../index.js';
+
+lockdown();
+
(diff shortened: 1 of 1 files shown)


packages/ses/scripts/bundle.js
```diff
@@ -1,10 +1,7 @@
 /* eslint-disable import/first */
 /* global process */
-
-import '../index.js';
-
 lockdown();
-
+import '../index.js';
```
erights, line 4: IIUC, this does not actually change the order in which things happen, at least in the absence of cyclic imports. All imported modules are evaluated/initialized before the code in this module starts evaluation/initialization. So either way, `../index.js` initialization will happen before this call to `lockdown`.

Attn @kriskowal

erights (changes requested):

packages/ses/scripts/bundle.js
kriskowal: This is true. To interleave behavior relative to other imports, we have to use a module like `lockdown.js` that just calls `lockdown()`. So,

```js
import '../index.js'; // or maybe import 'ses' reflexively works now
import './lockdown.js'; // just lockdown()
```

This insures that `./index.js` initializes before you call `lockdown()`.

It’s unclear why the tests pass with this defect. It might be that we need to test `yarn build` under `yarn test`.

leotm: makes sense ^ reverted [feat(ses): call lockdown earlier](https://github.com/endojs/endo/pull/2337/commits/b5c04714ebdfa5a36598b0b21f9c3d31d0525c73), looks good now

packages/ses/scripts/bundle.js
```diff
@@ -1,5 +1,10 @@
+/* eslint-disable import/first */
 /* global process */
+
 import '../index.js';
```
leotm, line 4:
```suggestion
// eslint-disable-next-line import/no-extraneous-dependencies
import 'ses';
```

also seems ok here

packages/ses/scripts/bundle.js
leotm: > It’s unclear why the tests pass with this defect. It might be that we need to test `yarn build` under `yarn test`.

yeah the tests still pass calling lockdown() too early before the imports 🤔

https://github.com/endojs/endo/blob/7a2955080f34880c8028e1b2dff06bf7b969790f/packages/ses/package.json#L75

even when doing a clean build before `"test": "yarn prepare && tsd && ava",` if that's what we meant

erights (approved):
I'm approving to clear my previous "request changes". FWIW LGTM, but on this I defer to @kriskowal . Please wait for an approval from him.

packages/ses/scripts/bundle.js
```diff
@@ -1,5 +1,11 @@
...
+import '../lockdown.js';
+
+lockdown();
+
```
kriskowal, line 8: This has the effect of ensuring that `lockdown()` gets called before the following imports evaluate:
```suggestion
import '../test/lockdown-safe.js';
```

The proposed change causes `lockdown()` to be called after all imported modules have evaluated. This is one of surprising effects of ESM loader semantics. Imported modules do not evaluate on the stack of the `import` statement in the way that CJS modules _may_ evaluate on the stack of a `require` call. CJS bundlers generally behave like ESM, though.

Even with the above formulation, it is possible to force the subsequent imports to evaluate earlier (and thus in the pre-lockdown realm) by importing this module from another module with a different import order over the same modules. This is outside our concern because this is the entrypoint module for the program.

kriskowal (changes requested):
Posted the required change. Thank you for engaging!

kriskowal (approved):

PR #2338, opened 2024-07-01
docs(ses): Add and update Agoric and MetaMask collateral

kriskowal:
This change adds links and logos for Agoric and MetaMask as systems that rely on `ses` and updates surrounding collateral.


turadg (approved):
🎸

PR #2344, opened 2024-07-11
feat(bundle-source): Add format flags, add new endoScript format

kriskowal:
Refs: https://github.com/Agoric/agoric-sdk/issues/9686

## Description

Endo’s `bundle-source` currently generates bundles with the `endoZipBase64` format, which requires the `importBundle` runtime. We typically bootstrap `importBundle` using the `getExport` or `nestedEvaluate` bundle format, which only need `eval` or `compartment.evaluate` to run. These use Rollup internally.

Elsewhere, for distributing SES and toward bootstrapping XS and Hermes variants of SES, we have our own bundler. In this change, we expose this format as `endoScript`, an alternative to `getExport` and `nestedEvaluate` which do not depend on Rollup.

We introduce a `--format` flag to `bundle-source` so any of the old and the new bundle formats are addressable from the CLI.

For easy hacking, we introduce a no-cache mode to `bundle-source` so we can just write a bundle to `stdout`.

This required a minor fix to the Compartment Mapper.

### Security Considerations

This will introduce a bundle format that is somewhat easier to audit because it preserves the lexical encapsulation of each module.

### Scaling Considerations

This may not perform as well as `nestedEvaluate` for bootstrapping, or will at least perform differently.

With this change, we can treat `bundle-source` as a one-stop-shop for all bundling concerns.

### Testing Considerations

To manually verify these changes, in `packages/bundle-source`:

```
yarn bundle-source demo/circular/a.js --format endoScript | jq .source -r
node -e "$(yarn bundle-source demo/circular/a.js --format endoScript | jq .source -r)"
yarn bundle-source -f endoScript --cache-json bundles/ demo/circular/a.js a
jq . bundles/bundle-a.json
jq -r .source bundles/bundle-a.json | less
```

No changes to prior workflows.

### Upgrade Considerations

None.

### DIFF
--- a/packages/bundle-source/NEWS.md
+++ b/packages/bundle-source/NEWS.md
@@ -8,6 +8,9 @@
   they cannot be made to run on the web without further work on module
   virtualization in the platform without entraining a client-side
   dependency on a JavaScript parser framework (namely Babel).
+- Adds a `-f,--format` command flag to specify other module formats.
+- Adds a new `endoScript` module format.
+- Adds a no-cache, bundle-to-stdout mode.
 
 # v3.2.1 (2024-03-20)
 
--- a/packages/bundle-source/cache.js
+++ b/packages/bundle-source/cache.js
@@ -7,6 +7,8 @@ import { makeFileReader, makeAtomicFileWriter } from './src/fs.js';
 
 const { Fail, quote: q } = assert;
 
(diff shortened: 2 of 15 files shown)


erights (approved):
LGTM, thanks!

dckc (approved):
the usage / CLI looks good to me

I defer to @erights on the code itself, including test coverage.

PR #2347, opened 2024-07-12
refactor: insert `await null;`s needed for await safety

erights:
Closes: #XXXX
Refs: #XXXX

## Description

Prior to this PR, `yarn lint` warns about many violations of our `await` safety rules. In vscode, I was very pleasantly surprised to learn about the automated fixes offered interactively for these:

image

For each of these, I accepted the "insert `await null;` before..." option. Uniquely for the one shown in the above screenshot I needed to do an additional manual readjustment to move the `// eslint-disable-next-line ...` so it would still be about the next line.

Caveats:

- I applied the automated fix manually, without reading the code in question or in any other way validating that this change is correctness preserving. My only evidence at the moment is that all the tests still pass.
- I would have loved to mark this PR with `refactor:` or at least `fix:`. But I marked it with `fix!` because there's no reason yet to think these changes are either correctness preserving or compat.
- I applied the automated change blindly even to the test files, even though we're not normally concerned about `await` safety there. Indeed, we should also change our lint configuration so `await` safety is unenforced for those. After this PR, we should also change our `await` safety enforcement for non-test files from "warning" to "error", to keep these from creeping back in.
- I applied the automated fix to the test files anyway since test code still often serves as precedent-by-example for production code. Since the automated fix made it easy to do this, I decided it was better to improve our test-code-as-precedent.

Obviously, this should not be merged until we're confident that these changes are correctness preserving. I don't know of any way to do so other than careful manual review. Perhaps as part of the same review, we might also conclude that it is compat, in which case we can change back from `fix!` to `fix:` or even hopefully `refactor:`.

For the various "considerations" below, I'll answer under the assumption that we have already reviewed these changes for correctness, and deemed them all to be correctness preserving. Obviously, until that happens, this PR is fatal re all the considerations below.

Update: As suggested by review comments, I moved those inserted `await null;`s to the top of their respective functions.

Update: Between @kriskowal 's evaluation comments below and my own inspection, I'm reasonably confident that this PR should count as a refactor, so I'm changing the title from `fix!...` to `refactor:...`.

### Security Considerations

Assuming the changes are correctness preserving, `await` safety aids security. 

### Scaling Considerations

none

we already need to explain `await` safety. This PR doesn't change that, but repairs our code so that we practice what we will preach.
### Testing Considerations

A bit distressing that all tests pass without any further intervention. This perhaps indicates a failure to test for the semantics that are changed by these extra awaits.

If this change is correctness preserving, then it is likely also compat. But not necessarily. The extra awaits introduce extra turn delays and extra interleaving points.
### Upgrade Considerations

If this PR is correctness preserving and compat, then there are unlikely to be any upgrade considerations. But I don't know that for sure, so we should also worry about this during review.


packages/check-bundle/lite.js
```diff
@@ -59,6 +59,7 @@ export const checkBundle = async (
     X`checkBundle cannot hash non-bundle, moduleFormat must be a string, got ${typeof moduleFormat}`,
   );
 
+  await null;
```
kriskowal, line 62: would hoist to top

packages/cli/src/commands/accept.js
```diff
@@ -13,6 +13,7 @@ const fromAsync = async iterable => {
 
 export const accept = async ({ guestName, agentNames }) => {
   process.stdin.setEncoding('utf-8');
+  await null;
```
kriskowal, line 16: would hoist to top

packages/cli/src/commands/list.js
```diff
@@ -40,6 +40,7 @@ export const list = async ({ directory, follow, json, verbose }) =>
       const directoryPath = parsePetNamePath(directory);
       agent = E(agent).lookup(...directoryPath);
     }
+    await null;
```
kriskowal, line 43: would hoist to top

packages/cli/src/commands/log.js
```diff
@@ -39,6 +39,7 @@ export const log = async ({ follow, ping }) =>
 
     const logPath = path.join(statePath, 'endo.log');
 
+    await null;
```
kriskowal, line 42: would hoist to top

packages/cli/src/commands/make.js
```diff
@@ -46,6 +46,7 @@ export const makeCommand = async ({
   let bundleReaderRef;
   /** @type {string | undefined} */
   let temporaryBundleName;
+  await null;
```
kriskowal, line 49: would hoist to top

packages/cli/src/commands/make.js
```diff
@@ -62,6 +63,7 @@ export const makeCommand = async ({
 
   await withEndoAgent(agentNames, { os, process }, async ({ agent }) => {
     // Prepare a bundle, with the given name.
+    await null;
```
kriskowal, line 66: would hoist to top &c ad nauseam

packages/daemon/src/web-server-node.js
```diff
@@ -138,6 +138,7 @@ export const make = async (_powers, context) => {
         E.sendOnly(webletBootstrap).ping();
 
         // eslint-disable-next-line no-use-before-define
+        await null;
```
kriskowal, line 141: This should be interfering with the eslint-disable-next-line above it.

packages/stream/index.js
```diff
@@ -175,6 +177,7 @@ export const prime = (generator, primer) => {
     },
     /** @param {Error} error */
     async throw(error) {
+      await null;
```
kriskowal, line 180: For the changes in this file, I would prefer to disable the lint rule and keep this as-is. I don’t see a practical problem with the current formulation and do see an extra turn of the even loop in what might be a hot path.

kriskowal (approved):
For many of these cases, I think the automated refactor was very correctly conservative and placed the `await null` immediately before a block that introduces `await` order uncertainty, while preserving rejected-in-one-turn for any exceptions that precede it.

But, a consequence is that the `await null` is unnecessarily buried in many cases where `await null` at the beginning of the async function would more clearly communicate that no additional `await null` will ever be needed in that function.

Having looked at all of these, I believe floating `await null` to the first statement of every function that needs one preserves correctness. I see no case where we depend on the rejection to occur on the stack of the caller.

packages/daemon/src/daemon-node-powers.js
```diff
@@ -367,6 +367,7 @@ export const makeDaemonicPersistencePowers = (
           return filePowers.readFileText(storagePath);
         };
         const json = async () => {
+          await null;
           return JSON.parse(await text());
```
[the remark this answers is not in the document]
kriskowal: This is an odd case because the text I’m suggesting has equivalent behavior but avoids the semblance of naughtiness with nothing more than a tortuously named intermediate variable.

packages/check-bundle/lite.js
erights: Done

packages/cli/src/commands/accept.js
erights: Done

packages/cli/src/commands/list.js
erights: Done

packages/cli/src/commands/log.js
erights: Done

packages/cli/src/commands/make.js
erights: Done

packages/cli/src/commands/make.js
erights: Done

packages/daemon/src/daemon-node-powers.js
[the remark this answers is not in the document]
erights: Done

packages/daemon/src/web-server-node.js
erights: If it were still using before defining, which it isn't. So I removed this uneeded directive as well.

packages/stream/index.js
erights: Done

packages/eventual-send/test/e.test.js
```diff
@@ -115,7 +117,6 @@ test('E sendOnly call missing method', async t => {
 
   const result = E.sendOnly(counter).decr(210);
   t.is(result, undefined, 'return is undefined as expected');
-  await null;
```
erights: Oops. Got swept up in the hoisting. Need to restore.

erights: Done

packages/far/test/e.test.js
```diff
@@ -137,7 +139,6 @@ test('E sendOnly call missing method', async t => {
 
   const result = E.sendOnly(counter).decr(210);
   t.is(result, undefined, 'return is undefined as expected');
-  await null;
```
erights: mistakenly moved. restore

erights: Done

erights:
@kriskowal before hoisting, everything passed CI without any manual changes. After hoisting, I'm getting a snapshot divergence at https://github.com/endojs/endo/actions/runs/9915491154/job/27396418994?pr=2347#step:8:1224

```
  ✘ [fail]: error-handling › fixtures-error-handling / cjs / loadLocation Did not match snapshot
  ✘ [fail]: error-handling › fixtures-error-handling / cjs / importLocation Did not match snapshot
```

that I don't know how to evaluate. PTAL, and what do you suggest?

kriskowal:
> @kriskowal before hoisting, everything passed CI without any manual changes. After hoisting, I'm getting a snapshot divergence at https://github.com/endojs/endo/actions/runs/9915491154/job/27396418994?pr=2347#step:8:1224
> ```
>   ✘ [fail]: error-handling › fixtures-error-handling / cjs / loadLocation Did not match snapshot
>   ✘ [fail]: error-handling › fixtures-error-handling / cjs / importLocation Did not match snapshot
> ```
> that I don't know how to evaluate. PTAL, and what do you suggest?

You can regenerate these snapshots, `yarn ava --update-snapshots`. These two tests are over-sensitive to changes in the (expected) stack trace.

PR #2349, opened 2024-07-12
fix(ses): turn agoric #9700 into an Endo PR

erights:

Closes: #XXXX
Refs: #XXXX

## Description

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations


erights:
Having created this Draft PR form of the patch-PR https://github.com/Agoric/agoric-sdk/pull/9700 , I'm now closing this PR because it is not the approach I want to take anyway.

erights:
Superseded by https://github.com/endojs/endo/pull/2355

PR #2350, opened 2024-07-13
refactor(compartment-mapper): Pivot to module descriptors

kriskowal:
Refs: #400

## Description

XS does not support `compartment.module`. Rather than emulate it, I chose bring SES into closer alignment with XS module descriptors and use those instead. The Compartment Mapper is the only place we use this feature at the moment.

### Security Considerations

None.

### Scaling Considerations

None.

None.

### Testing Considerations

The existing Compartment Mapper tests exercise this path rigorously.

The new version of Compartment Mapper will not work with an older version of SES. Lerna will make sure the constraint gets updated when we release.

### Upgrade Considerations

None.

### DIFF
--- a/packages/compartment-mapper/src/link.js
+++ b/packages/compartment-mapper/src/link.js
@@ -259,7 +259,10 @@ const makeModuleMapHook = (
             )}}`,
           );
         }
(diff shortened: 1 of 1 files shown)


erights (approved):
LGTM, simple!

PR #2352, opened 2024-07-13
feat(compartment-mapper): JSON module support for makeBundle

kriskowal:
Refs: #400, https://github.com/Agoric/agoric-sdk/issues/9686

## Description

Endo’s zip bundle format and Rollup both support importing JSON modules. This change adds JSON module support to Endo’s script bundle format. We need this to have Rollup parity for Agoric Core Eval scripts and also to unblock test262 for XS native compartment parity tests.

### Security Considerations

Obviates Rollup, increases legibility of script bundles.

### Scaling Considerations

None.

None. JSON module support is assumed. No one will notice unless it doesn’t work.

### Testing Considerations

This change adds a JSON module to the existing bundle test fixture. This also verifies parity with the Zip bundle format.

This make Endo more compatible with Rollup, but Endo continues to not support `import / with { type: "json" }`, which would be necessary for parity with Node.js proper and newer bundlers.

### Upgrade Considerations

None.

### DIFF
--- a/packages/compartment-mapper/src/bundle-json.js
+++ b/packages/compartment-mapper/src/bundle-json.js
@@ -1,0 +1,24 @@
+/* Provides JSON support for `bundle.js`. */
+
+const textDecoder = new TextDecoder();
+
+export default {
+  runtime: '',
+  getBundlerKit({ index, indexedImports, bytes }) {
+    // Round-trip to revalidate JSON and squeeze out space.
+    const json = JSON.stringify(JSON.parse(textDecoder.decode(bytes)));
+    return {
+      getFunctor: () => `\
+// === functors[${index}] ===
+(set) => set(${json}),
+`,
+      getCells: () => `\
+    { default: cell('default') },
+`,
+      getReexportsWiring: () => '',
+      getFunctorCall: () => `\
+  functors[${index}](cells[${index}].default.set);
+`,
+    };
+  },
+};
--- a/packages/compartment-mapper/src/bundle.js
+++ b/packages/compartment-mapper/src/bundle.js
@@ -22,6 +22,7 @@ import { parseLocatedJson } from './json.js';
 
 import mjsSupport from './bundle-mjs.js';
 import cjsSupport from './bundle-cjs.js';
+import jsonSupport from './bundle-json.js';
 
 const textEncoder = new TextEncoder();
 
@@ -56,7 +57,7 @@ const sortedModules = (
 
     const source = compartmentSources[compartmentName][moduleSpecifier];
     if (source !== undefined) {
-      const { record, parser, deferredError } = source;
+      const { record, parser, deferredError, bytes } = source;
       if (deferredError) {
         throw Error(
           `Cannot bundle: encountered deferredError ${deferredError}`,
@@ -84,6 +85,7 @@ const sortedModules = (
           parser,
           record,
           resolvedImports,
+          bytes,
         });
 
         return key;
@@ -120,6 +122,7 @@ const sortedModules = (
 const implementationPerParser = {
   'pre-mjs-json': mjsSupport,
   'pre-cjs-json': cjsSupport,
+  json: jsonSupport,
 };
 
 function getRuntime(parser) {
--- a/packages/compartment-mapper/src/bundle.js
+++ b/packages/compartment-mapper/src/bundle.js
@@ -1,6 +1,8 @@
 // @ts-check
 /* eslint no-shadow: 0 */
 
+/** @import {StaticModuleType} from 'ses' */
+
 /** @import {ArchiveOptions} from './types.js' */
 /** @import {CompartmentDescriptor} from './types.js' */
 /** @import {CompartmentSources} from './types.js' */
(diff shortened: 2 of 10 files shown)


packages/compartment-mapper/test/bundle.test.js
```diff
@@ -52,6 +52,7 @@ const expectedLog = [
     fromMjs: 'foo',
     zzz: 1,
   }),
+  { widdershins: 'againshins' },
```
erights, line 55: Could you explain the joke? I don't get it and googling did not help.

packages/compartment-mapper/src/bundle-json.js
```diff
@@ -0,0 +1,23 @@
...
+      getFunctor: () => `\
+// === functors[${index}] ===
+(set) => set(${json}),
+`,
+      getCells: () => `\
+    { default: cell('default') },
+`,
+      getReexportsWiring: () => '',
+      getFunctorCall: () => `\
+  functors[${index}](cells[${index}].default.set);
```
erights, line 19: grepping for these four method names, I only see uses with bundle.js and a definition of a bundle with three of them. Where can I find an explanation of what they mean?

packages/compartment-mapper/test/bundle.test.js
kriskowal: I’m so glad you asked. It’s nonsense. Widdershins means counter-clockwise and its opposite is usually considered deosil, but widdershins has readily recognizable German cognates and deosil has Irish origins. So, I’m asking myself what the opposite of widdershins would be along a shared etymological path and I’m drawing a blank. But, "again", "against", "gainsay", and some other words are all cognates of "gegen" in German, so widdershins could have a Germanic cousin in `againshins`. I still haven’t figured out an opposite.

packages/compartment-mapper/src/bundle-json.js
kriskowal: There is no explanation except how they’re used in bundle.js.

packages/compartment-mapper/test/bundle.test.js
naugtur: Linguistic competence is a factor in why naming things is hard, just not in the way I thought.

packages/compartment-mapper/src/bundle-json.js
```diff
@@ -0,0 +1,23 @@
+/* Provides JSON support for `bundle.js`. */
+
+const textDecoder = new TextDecoder();
+
+export default {
+  runtime: '',
+  getBundlerKit({ index, indexedImports, bytes }) {
+    const json = textDecoder.decode(bytes).trim();
```
naugtur, line 8: (could) decode and encode back, in case it's not actually JSON

packages/compartment-mapper/src/bundle-json.js
naugtur: I'm wondering if it'd make sense to replace some of this with empty strings or no-ops and inline json in the cell

```
  getCells: () => `\
    { default: cell('default', ${json}) },
`,
```

packages/compartment-mapper/test/bundle.test.js
kriskowal: We aren’t allowed to use `"foo"` anymore. This is not better, but it is not `"foo"`.

packages/compartment-mapper/src/bundle-json.js
kriskowal: Oh, no. That doesn’t work because `cell` and `cells` are not in the lexical scope of functors.

packages/compartment-mapper/src/bundle-json.js
kriskowal: I’ve followed-up with a change that aligns with this reasoning. I hesitated because this eliminates whatever spacing choices the author had made.

packages/compartment-mapper/src/bundle-json.js
kriskowal: Oh, they were in scope! I’ve added a commit that adds docs, asserts this invariant, and fixes it. There should not be any lexical variables in scope that might mask a global.

packages/compartment-mapper/test/bundle.test.js
erights: There's a `'foo'` above on line 52 ;)

packages/compartment-mapper/src/bundle.js
```diff
@@ -11,6 +13,42 @@
...
+ * In the generated bundle runtime, the function will receive an environment
+ * record: a record mapping every name of the corresponding module's internal
+ * namespace to a "cell" it can use to get, set, or observe the linked
+ * variable.
+ * @property {() => string} getCells Produces a JavaScript string consisting of
+ * a JavaScript object and a trailing comma.
+ * The string is evaluated in a lexical context with a `cell` maker, the `cells`
+ * array of every module's internal environment record.
+ * @property {() => string} getFunctorCall Produces a JavaScript string with a
+ * single statement that will call this module's functor with the calling
```
erights, line 30: The implementation in this file returns a string consisting only of a comment, which therefore does not "call this module's functor"

packages/compartment-mapper/src/bundle.js
```diff
@@ -11,6 +13,42 @@
...
+ * a function expression followed by a comma delimiter that will be evaluated in
+ * a lexical scope with no free variables except the globals.
+ * In the generated bundle runtime, the function will receive an environment
+ * record: a record mapping every name of the corresponding module's internal
+ * namespace to a "cell" it can use to get, set, or observe the linked
+ * variable.
+ * @property {() => string} getCells Produces a JavaScript string consisting of
+ * a JavaScript object and a trailing comma.
+ * The string is evaluated in a lexical context with a `cell` maker, the `cells`
+ * array of every module's internal environment record.
```
erights, line 28: `getCells` implementation in this file has no trailing comma.

erights (approved):
Assuming the code is correct and the doc-comments are wrong but close, LGTM, thanks!

packages/compartment-mapper/src/bundle.js
```diff
@@ -11,6 +13,42 @@
...
 
+/**
+ * @typedef {object} BundlerKit
+ * @property {() => string} getFunctor Produces a JavaScript string consisting of
+ * a function expression followed by a comma delimiter that will be evaluated in
+ * a lexical scope with no free variables except the globals.
+ * In the generated bundle runtime, the function will receive an environment
+ * record: a record mapping every name of the corresponding module's internal
+ * namespace to a "cell" it can use to get, set, or observe the linked
+ * variable.
```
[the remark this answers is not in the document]
kriskowal: For this case, there needed to be a trailing comma and this was a bug that would surface with a broken language-support implementation (of which there are none, and it is not as yet extensible :-P)

Fixed.

kriskowal:
Ironed out the types.

PR #2353, opened 2024-07-13
feat(bundle-source): Add `tag` command-line flag

kriskowal:
Refs: #400

## Description

We’ve discovered that, to create an Endo script bundle that captures our `ModuleSource` constructor shim, it will need to entrain Babel. With the default build conditions, Babel entrains a `debug` package, which entrains `node:tty`. So, it is necessary to add the `"browser"` condition, which instead entrains a package with a JSON module (#2352). This change threads arbitrary conditions (build tags) through `bundle-source` to conveniently generate Endo script bundles with different tags like `browser` or `xs`.

### Security Considerations

None.

### Scaling Considerations

None.

The new flags are advertised in the command’s usage display.

### Testing Considerations

- [ ] test scaffold and manual testing instructions

None.

### Upgrade Considerations

None.

### DIFF
--- a/packages/bundle-source/NEWS.md
+++ b/packages/bundle-source/NEWS.md
@@ -11,6 +11,8 @@
 - Adds a `-f,--format` command flag to specify other module formats.
 - Adds a new `endoScript` module format.
 - Adds a no-cache, bundle-to-stdout mode.
+- Adds a `-t,--tag` command flag to specify export/import conditions like
+  `"browser"`.
 
 # v3.2.1 (2024-03-20)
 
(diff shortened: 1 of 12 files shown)


erights:
> With the default build conditions, Babel entrains a debug package, which entrains node:tty.

Are there better behaved non-default build conditions?

erights:
nevermind. I get it now.

erights (approved):
Simple! LGTM!

dckc:
> ... it will need to entrain Babel.

it = ?

I'm not sure I'm following.

kriskowal:
> an Endo script bundle that captures our `ModuleSource` constructor shim,

An Endo script bundle that captures our `ModuleSource` constructor shim (née `@endo/static-module-record`) such that a test262 “prelude” (shim script) can arrange for the same environment as XS, so we can write test262 tests that expect the same Hardened JavaScript environment.

Issue #2348, opened 2024-07-12
tests in TypeScript (.ts files) lack stack traces


erights:
See https://github.com/Agoric/agoric-sdk/issues/8662 

From https://github.com/Agoric/agoric-sdk/pull/9700

> file://test/facade.test.ts:1:1043

vs

> (test/facade.test.ts:52:9)

Why are we seeing this difference? Clearly, the second stack trace is more informative, and the tamed stack trace from Endo, whether 'safe' or 'unsafe', should show this more informative one. It looks like the second is relative to the source code the programmer wrote, which is ideal. But of course, the source is TS and what's executing is the JS that the TS compiles to. From the first, I'd guess that either the TS compiler is compiling into a single line of JS, or that some bundler is flattening the compiled JS into a single line of JS. But in either case, we should understand why the second stack trace is more informative. Is it based on a source map generated by the TS compiler? How does the native v8 stack get to be adjusted for that, but not the stack we're forming using the JS error API internally?

mhofman:
> I'd guess that either the TS compiler is compiling into a single line of JS

Correct


> we should understand why the second stack trace is more informative. Is it based on a source map generated by the TS compiler? How does the native v8 stack get to be adjusted for that, but not the stack we're forming using the JS error API internally?

TS does generate source maps. And Node.js uses `prepareStackTrace` to parse source maps and recreate the authored stack trace info. v8 itself doesn't know anything about source maps. However when SES overrides `prepareStackTrace`, we lose that Node.js behavior aware of source maps. Since we can't invoke that original behavior because of strange layering inside Node.js, we're left with having to re-implement the source mapping behavior ourselves in SES. See https://github.com/endojs/endo/issues/1798

PR #2355, opened 2024-07-14
fix(ses): work around #2348 linenumber bug

erights:
Closes: https://github.com/endojs/endo/issues/2348 

Refs: https://github.com/Agoric/agoric-sdk/pull/9711 https://github.com/endojs/endo/issues/1798 https://github.com/endojs/endo/pull/1799 https://github.com/Agoric/agoric-sdk/issues/8662 https://github.com/Agoric/agoric-sdk/pull/9700

## Description

Prior to this PR, when you ran on Ava on Node a test written in TypeScript, you'd see something like the following in your stack traces.

```
  boot/test/bootstrapTests/stack-linenumbers.test.ts:1:104
```

This is because the TypeScript compiler compiles a TypeScripy file into one line
of JavaScript with a sourceMap that should map back into original
source positions. Node specifically makes use of that sourceMap
to produce original line-numbers. However, Node does this in a way
that resists virtualization, so the normal SES error taming cannot use
this sourceMap info.

By default, this PR does not change this behavior. However it recognizes a new `SUPPRESS_NODE_ERROR_TAMING` environment variable.

With the `SUPPRESS_NODE_ERROR_TAMING` environment variable absent
or set to `'disabled'`, you should still see stack traces as shown above

However, if you also set  the `SUPPRESS_NODE_ERROR_TAMING` environment
variable `'enabled'`, for example by doing

```sh
$ export SUPPRESS_NODE_ERROR_TAMING=enabled
```
at a bash shell, then when you run this test you should instead see
something like
```
boot/test/bootstrapTests/stack-linenumbers.test.ts:40:32
```

At https://github.com/Agoric/agoric-sdk/pull/9711 I both 
- turn this PR into an agoric-sdk patch of endo, in order to emulate this fix until the next endo-release-agoric-sdk-sync cycle, and
- add a test case that emits an error stack trace from an Ava test case written in TypeScript, to test that it works.

### Security Considerations

This new behavior only applies when `errorTaming: 'unsafe'`, on v8, and with this new environment variable enabled. 

Setting `errorTaming: 'unsafe'` already flags to sacrifice some security for a better debugging experience. But the loss of security is moderate enough --- mostly confidentiality rather than integrity --- that some may chose this setting for some production purposes.

The new behavior is a more severe loss of security that really should be used ***only during development***, not production, when even a severe loss of security is usually not an issue.

### Scaling Considerations

none

The behavior prior to this PR or without this environment variable enabled is an unpleasant debugging experience. However, developers won't know how to repair it, or even that it can be repaired, without explanation. Even then, the difficultly of discovery in a problem.

The names `SUPPRESS_NODE_ERROR_TAMING` and the settings `'enabled'` and `'disabled'` are by no means clear expressions of what this does. Reviewers, ***better names would be appreciated!***

### Testing Considerations

The point. As developers write and run tests written in TypeScript, they need to iterate with problems revealed by the tests, for which they need good line numbers, including into the test code.

When the environment variable is enabled, the new behavior broke some SES tests written specifically to test the old behavior. This would not happen under CI because the environment variable is not set by default, and so may not have been noticed. But it was revealed in local testing. To repair this, this PR also sets those tests up to set `process.env.SUPPRESS_NODE_ERROR_TAMING` to `'disabled'` before lockdown, protecting those tests from the external environment variable setting.

Awkwardly, at the moment https://github.com/Agoric/agoric-sdk/pull/9711 serves as the only test of this PR. This is because I failed to figure out how to configure things so I can run TypeScript tests under Ava, like https://github.com/Agoric/agoric-sdk/pull/9711 does. I tried cargo culting the configs that seemed relevant, but it didn't work.

Reviewers, if you let me know how to do this, I'll duplicate the test case from https://github.com/Agoric/agoric-sdk/pull/9711 here, which would be good.

With the environment variable absent or disabled, there should be zero difference in behavior, so none.

In a development environment where this environment variable is enabled, some stack traces will be different. But outside of SES itself, nothing should depend on the contents of stack traces, so again none.

### Upgrade Considerations

No upgrade considerations.

Nothing BREAKING.


packages/ses/test/error/tame-v8-error.test.js
```diff
@@ -1,7 +1,7 @@
 import test from 'ava';
 import '../../index.js';
 
-lockdown();
+lockdown({ errorTaming: 'safe' });
```
erights, line 4: The absence of this was a previously unnoticed bug, since omitting it left it up to environment variables to determine, which ruins the purpose of the test. Since the relevant env variables would not be set in CI, the bug would not be noticed there either.

mhofman:
> The new behavior is a more severe loss of security that really should be used _**only during development**_, not production, when even a severe loss of security is usually not an issue.



> The names `SUPPRESS_NODE_ERROR_TAMING` and the settings `'enabled'` and `'disabled'` are by no means clear expressions of what this does. Reviewers, _**better names would be appreciated!**_

Why not re-use the existing `errorTaming` options, and a new value, like `extremely-unsafe` or `so-unsafe-you-will-be-fired-if-you-use-in-production` ([React background for the unaware](https://news.ycombinator.com/item?id=11447020))

turadg:
> > The new behavior is a more severe loss of security that really should be used _**only during development**_, not production, when even a severe loss of security is usually not an issue.
> 
> > The names `SUPPRESS_NODE_ERROR_TAMING` and the settings `'enabled'` and `'disabled'` are by no means clear expressions of what this does. Reviewers, _**better names would be appreciated!**_
> 
> Why not re-use the existing `errorTaming` options, and a new value, like `extremely-unsafe` or `so-unsafe-you-will-be-fired-if-you-use-in-production` ([React background for the unaware](https://news.ycombinator.com/item?id=11447020))

I concur. This new case is a kind of `errorTaming`. And it would be quite surprising to the user to set  `SUPPRESS_NODE_ERROR_TAMING: enabled` but that not happen because they didn't also set `errorTaming: unsafe`.

As a develop I would expect that `errorTaming: unsafe` is unsafe, even in production, but if there's reason to expect 'unsafe' to be used in production then I think this new level should indicate that it should not be. "extremely-unsafe" risks making "unsafe" seem like it has some safety. "you-will-be…" is clear but over the top, imo. `development` or `debugging` suffice.

erights:
I think I concur as well. I'll try it out and we'll see what we think.

kriskowal:
Approach looks good. I propose `errorTaming: 'unsafe-debug'`.

packages/ses/NEWS.md
```diff
@@ -1,5 +1,27 @@
 User-visible changes in SES:
 
+# Next release
+
+- When running Ava tests on Node written in TypeScript, the SES error taming
```
erights, line 5: There's no need to mention Ava here. This should be for all TypeScript on Node on v8.

packages/ses/docs/reference.md
```diff
@@ -179,9 +184,10 @@ below.
   </tr>
   <tr>
     <td><code>overrideTaming</code></td>
-    <td><code>'moderate'</code> (default) or <code>'min'</code></td>
+    <td><code>'moderate'</code> (default) or <code>'min'</code> or <code>'severe'</code></td>
     <td><code>'moderate'</code> moderates mitigations for legacy compatibility,<br>
-        <code>'min'</code> minimal mitigations for purely modern code</td>
+        <code>'min'</code> minimal mitigations for purely modern code, <br>
+        <code>'severe'</code> when moderate mitigations are inadequate</td>
```
erights, line 190: As a drive-by, I happened to notice that `'severe'` was accidentally omitted. Fixed.

erights:
> Approach looks good. I propose `errorTaming: 'unsafe-debug'`.

Done! PTAL

packages/ses/NEWS.md
```diff
@@ -4,6 +4,25 @@ User-visible changes in SES:
 
 - Adds support for module descriptors better aligned with XS.
 
+- When running Ava tests on Node written in TypeScript, the SES error taming
+  gives line-numbers into the generated JavaScript, in which all compiled code
+  is on line 1. This happens even with the normal development-time lockdown
+  options setting
```
turadg, line 10:
```suggestion
- When running transpiled code on Node, the SES error taming
  gives line-numbers into the generated JavaScript, which often don't match the 
  the original lines. This happens even with the normal development-time lockdown
  options setting,
```

packages/ses/NEWS.md
```diff
@@ -4,6 +4,25 @@ User-visible changes in SES:
...
+  ```js
+  errorTaming: 'unsafe'
+  ```
+  or setting the environment variable
+  ```sh
+  $ export LOCKDOWN_ERROR_TAMING=unsafe
+  ```
+  In addition to the normal `'safe'` and `'unsafe'` settings, this release
```
turadg, line 18:
```suggestion
  To get the original line numbers, this release
```

packages/ses/NEWS.md
```diff
@@ -4,6 +4,25 @@ User-visible changes in SES:
...
+  adds `'unsafe-debug'`. This `errorTaming: 'unsafe-debug'` setting
+  should be used ***during development only*** when you can
+  sacrifice more security for a better debugging experience, as explained at
+  [`errorTaming` Options](https://github.com/endojs/endo/blob/master/packages/ses/docs/lockdown.md#errortaming-options).
+  With this setting, when running Ava tests written in TypeScript on Node,
```
turadg, line 23:
```suggestion
  With this setting, when running transpiled code on Node (e.g. tests written in TypeScript),
```

packages/ses/NEWS.md
```diff
@@ -4,6 +4,25 @@ User-visible changes in SES:
...
+  the stacktrace line-numbers should point back into the original
+  TypeScript source, as they would on Node without SES.
```
turadg, line 25:
```suggestion
  the stacktrace line-numbers point back into the original
  source, as they do on Node without SES.
```

packages/ses/docs/lockdown.md
```diff
@@ -293,27 +293,48 @@ default error behavior is much more dangerous. The v8 `Error` constructor
...
+Node on v8.
 
 ```js
 lockdown(); // errorTaming defaults to 'safe'
 // or
 lockdown({ errorTaming: 'safe' }); // Deny unprivileged access to stacks, if possible
 // vs
 lockdown({ errorTaming: 'unsafe' }); // stacks also available by errorInstance.stack
+// vs
+lockdown({ errorTaming: 'unsafe-debug' }); // sacrifice more safety for TypeScript line numbers.
```
turadg, line 337:
```suggestion
lockdown({ errorTaming: 'unsafe-debug' }); // sacrifice more safety for sourcemapped line numbers.
```

packages/ses/docs/reference.md
```diff
@@ -258,18 +264,30 @@ In JavaScript the stack is only available via `err.stack`, so some
...
+this is usually the right tradeoff. Please do not use it in production.
+
 ```js
 lockdown(); // errorTaming defaults to 'safe'
 // or
 lockdown({ errorTaming: 'safe' }); // Deny unprivileged access to stacks, if possible
 // vs
 lockdown({ errorTaming: 'unsafe' }); // Stacks also available by errorInstance.stack
+// vs
+lockdown({ errorTaming: 'unsafe-debug' }); // sacrifice more safety for TypeScript line numbers.
```
turadg, line 290:
```suggestion
lockdown({ errorTaming: 'unsafe-debug' }); // sacrifice more safety for sourcemapped line numbers.
```

turadg (approved):
Approving modulo copy changes to remove references to Ava and TypeScript.

packages/ses/test/error/tame-v8-error.test.js
erights: Resolving.

packages/ses/NEWS.md
erights: Done

packages/ses/NEWS.md
erights: Done, thanks.

packages/ses/NEWS.md
erights: Done

packages/ses/NEWS.md
erights: Done

packages/ses/NEWS.md
erights: Done. Together, all this reads much better. Thanks!

packages/ses/docs/lockdown.md
erights: Done

packages/ses/docs/reference.md
erights: Done

kriskowal (approved):

Issue #2354, opened 2024-07-14
SES demos not working


GulgDev:
https://ses-demo.netlify.app/dist/lockdown.umd.js is not found, so SES fails to lockdown and https://ses-demo.netlify.app/demos/ don't work as expected.

liangyuqi:
same as you 

and: Uncaught ReferenceError: lockdown is not defined

<img width="1716" alt="截屏2024-07-15 17 57 20" src="https://github.com/user-attachments/assets/871617ea-ddc6-45a2-91a5-8062c7938467">

erights:
Thanks @GulgDev @liangyuqi for the reports! Verified. We're investigating. In the meantime, the investigation led to https://github.com/endojs/endo/pull/2357 , which fixes an unrelated problem that was also unnoticed. Thanks!

erights:
Sorry, that was automatically closed by accident from merging #2357. Reopening.

Assigning to @kriskowal who is already working towards fixing this.

kriskowal:
The demo has been moved to and fixed at https://hardenedjs.org/challenge/

PR #2357, opened 2024-07-15
fix(ses-demo): adapt to disabled `Date.now()` throwing

erights:
Closes: #XXXX
Refs: https://github.com/endojs/endo/issues/910#issuecomment-1581855420 https://github.com/endojs/endo/pull/1718 https://github.com/endojs/endo/issues/2354 https://github.com/endojs/endo/issues/910

## Description

While investigating https://github.com/endojs/endo/issues/2354 , I just tried it locally by visiting 

    file:///.../endojs/endo/packages/ses/demos/console/index.html

and

    file:///.../endojs/endo/packages/ses/demos/challenge/index.html

in my browser. The first is the SES demo console, which worked just fine. The second is the SES Escape Room, which still relied on the disabled `Date.now()` not throwing. Indeed, before https://github.com/endojs/endo/pull/1718 a disabled (secure mode) `Date.now()` returned `NaN`. But https://github.com/endojs/endo/pull/1718 changed it to throw.

What's strange is that in https://github.com/endojs/endo/pull/1718 I revise endojs/endo/packages/ses/demos/challenge/index.html to adjust the text to say that `Date.now()` is "disabled" rather than "NaN". But I didn't fix the escape room code.

Although I found this while investigating https://github.com/endojs/endo/issues/2354 , this is a completely distinct bug that is unrelated to https://github.com/endojs/endo/issues/2354 . This PR itself does nothing to fix https://github.com/endojs/endo/issues/2354 .

### Security Considerations

none
### Scaling Considerations

none

All our docs that link to an explain the Escape Room challenge need to be revisited, especially once https://github.com/endojs/endo/issues/2354 is fixed. For example, https://agoric.com/blog/technology/a-taxonomy-of-security-issues , which I just verified links to the broken page reported at https://github.com/endojs/endo/issues/2354

### Testing Considerations

It is frustrating that the Escape Room is broken at least since https://github.com/endojs/endo/pull/1718 , and also broken by https://github.com/endojs/endo/issues/2354 for an undetermined period of time, without anyone noticing until now. It would be good to bring that site under some kind of automated testing.

none
### Upgrade Considerations

none


packages/ses/demos/challenge/main.js
```diff
@@ -141,7 +141,14 @@ lockdown();
...
         return false;
       }
-      delayMS(10);
+      try {
+        delayMS(10);
+      } catch (err) {
+        if (err instanceof TypeError) {
+          // assume we cannot delay because of err
+        }
+        throw err;
```
kriskowal, line 150: This will throw regardless of the condition. I presume you mean to invert the condition and move the throw into the block.

kriskowal (changes requested):

packages/ses/demos/challenge/main.js
erights: Doh! Thanks for catching this. Done.

erights:
Done. PTAL

kriskowal (approved):

PR #2362, opened 2024-07-18
test: expand the browser smoke test to break when lockdown doesn't work under CSP denying eval

naugtur:
This test will make sure we don't start relying on direct eval for lockdown itself.

### DIFF
--- a/browser-test/server.js
+++ b/browser-test/server.js
@@ -7,7 +7,11 @@ const server = http.createServer((req, res) => {
   console.log(req.url);
 
   if (req.url === '/') {
-    res.writeHead(200, { 'Content-Type': 'text/html' });
+    res.writeHead(200, { 
+      'Content-Type': 'text/html', 
+      // ensure SES works when eval is firbidden by CSP (no `unsafe-eval` in script-src)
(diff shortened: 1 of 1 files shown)


erights (commented):
ELI5

kriskowal:
The test in question only runs `lockdown` and does not attempt to execute confined code. So, it may as well perform that test under a prohibitive Content-Security-Policy.

This will interact in interesting ways with using eval for feature-detection. We should go on to furthermore verify that Lockdown was effective for the questionable cases (like %AsyncGeneratorPrototype%)

erights (commented):
@naugtur , I know so little about CSP that you should ask someone else. I'm unassigning myself as a reviewer.

erights (commented):
Oh. I cannot unassign myself. Nevertheless...

naugtur:
The relevant information here is that CSP denies all evaluators by default (causes them to throw) and lockdown already supports that. If the feral Function cannot be used as an evaluator, the fragment of lockdown that uses `eval('async function*(){}')` to get the various function type constructors is skipped because there's no need to tame them.

This test change is intended as a regression check in case we use the same trick elsewhere - in places where it would be needed for lockdown under CSP as well.

PR #2366, opened 2024-07-18
ci: Pin Node 22.4.1

kriskowal:
Node.js 22.5.0 regressed CI with error:

```
Usage Error: Couldn't find the node_modules state file - running an install might help (findPackageLocation)
```

Pinning to 22.4.1 unblocks CI.

We should revisit reverting this with the next Node.js release.


rekmarks (approved):
Most reasonable.

Issue #2368, opened 2024-07-18
test cover cli demo


grypez:
## What is the Problem Being Solved?

The cli package is [not currently tested](https://github.com/endojs/endo/blob/3e3b8c6059c16c788223586ba01190d70963d61c/packages/cli/package.json#L31). It would be nice to write some tests, and the demo seems like a good place to start.

## Description of the Design

- [x] Activate ava test coverage in cli package
- [x] Create an ava entry point which tests each section in isolation after running `endo purge -f`.

For each runnable section of [the cli demo](https://github.com/endojs/endo/blob/master/packages/cli/demo/README.md), produce a bash script executing that functionality and a log of the expected output of the script, then assert equality of the expected and observed output of the script using diff (for an example see the description of [this issue](https://github.com/endojs/endo/issues/2074#issue-2135049457)).
- [x] A counter example
- [x] Doubler Agent
- [x] Sending Messages
- [x] Names in transit...
- [x] Mailboxes are symmetric
- [ ] Familiar Chat (note UI interaction)
- [x] Running a confined script

### Out of Scope
- Programmatically ensuring that the documented functionality aligns with the tested functionality

kriskowal:
You may have more luck ensuring portability if the tests are framed as Node.js shelling out to the CLI. Execa is popular at Agoric. https://www.npmjs.com/package/execa

PR #2372, opened 2024-07-23
[WIP] Test cli demo

grypez:
Refs: #2368 

## Description

A first pass at test covering the cli demo. Current weaknesses in the implementation include concurrency and independence. They are related and both should be resolved by a change to the daemon initialization in the test setups.

Note that despite these weaknesses, this setup is useful enough to reveal that certain cli commands do not behave as the demo supposes. Notably:
- `endo list --as alice-agent` is rejected because `--as` is not a supported option for `endo list`. This could have been caught with unit tests on the cli endpoints.
- `endo inbox` does not display messages as expected. In some cases the template string for the message has simply been altered, but the discrepancy in the mailboxes-are-symmetric section is unlikely to have been caught with a unit test.

#### Concurrency
Ideally each tested section would run concurrently, however as written the demo sections are implicitly serial. This is a nice UX for a demo and I don't suggest changing it.

Thus tests running on the same daemon which reference the same objects interfere with one another. Possible solutions:
1. Run a different endo daemon per test, configured to its own directory
1. Test a slightly modified version of each section which prefixes petnames with a per-test unique id
1. Run the tests serially
1. Some other solution of which I am unaware.

At the moment I have opted for (3.) _test the sections serially_. The logic to test each section is written in a distinct file bearing the name of that section, but the routines are imported into a common `index.test.js` and declared to ava as serial.

I intend to adopt (1.) in the near future, reusing setup from `daemon/test/endo.test.js`. I think [prepareConfig](https://github.com/endojs/endo/blob/8bebd9416cca3199dba5d5655d98046e2c7781e7/packages/daemon/test/endo.test.js#L209) accomplishes what I have in mind.

#### Independence
Although the tests are run serially, the test files export setup and teardown routines in a `withContext` wrapper which can be used to replay the state-changing commands which lead up to the given test. Although the tests are all declared to ava in the same `index.test.js` file, a test can be run independently with `yarn test --match <test-name>`, for example `yarn test --match doubler-agent`. However, the independence is weak, as any breakage in the state changing commands in the previous sections will cause the setup for subsequent tests to fail. Stronger test independence is ideal.

### Upgrade Considerations

* Adds the execa dev dependency.

### DIFF
--- a/packages/cli/test/demo/confined-script.js
+++ b/packages/cli/test/demo/confined-script.js
@@ -1,0 +1,10 @@
+import { withContext as withContextDaemon } from './daemon.js';
+
+export const section = async (execa, testLine) => {
+  await withContextDaemon(execa, async () => {
+    // If a runlet returns a promise for some value, it will print that value before exiting gracefully.
+    await testLine(execa`endo run runlet.js a b c`, {
+      stdout: "Hello, World! [ 'a', 'b', 'c' ]\n42",
+    });
+  });
+};
--- a/packages/cli/test/demo/index.test.js
+++ b/packages/cli/test/demo/index.test.js
@@ -1,0 +1,14 @@
+import test from '@endo/ses-ava/prepare-endo.js';
+import { $ } from 'execa';
+import { makeSectionTest } from './section.js';
+import { withContext as withContextDaemon } from './daemon.js';
+
+test.serial(
+  'trivial',
+  makeSectionTest($({ cwd: 'demo' }), async (execa, testLine) => {
+    await withContextDaemon(execa, async () => {
+      const maxim = 'a failing test is better than failure to test';
+      await testLine(execa`echo ${maxim}`, { stdout: maxim });
+    });
+  }),
+);
--- a/packages/cli/test/demo/names-in-transit.js
+++ b/packages/cli/test/demo/names-in-transit.js
@@ -1,0 +1,41 @@
+import { withContext as withContextSendingMessages } from './sending-messages.js';
+
+export const section = async (execa, testLine) => {
+  await withContextSendingMessages(execa, async () => {
+    // In this example, we send alice our "doubler" but let it appear...
+    await testLine(
+      execa`endo send alice ${'Please enjoy this @counter:doubler.'}`,
+    );
+    await testLine(execa`endo inbox --as alice-agent`, {
+      stdout: /^1\. \"HOST\" sent \"Please enjoy this \@counter\.\"/,
+    });
+    await testLine(
+      execa`endo adopt --as alice-agent 1 counter --name redoubler`,
+    );
+    await testLine(execa`endo list --as alice-agent`, {
+      stdout: 'redoubler',
+    });
+    await testLine(execa`endo dismiss --as alice-agent 1`);
+  });
+};
+
+/**
+ * Wraps a function with the setup and teardown of the endo object state from names-in-transit.
+ *
+ * @param {*} execa
+ * @param {() => Promise<void} implementation
+ * @returns {Promise<void>}
+ */
+export const withContext = async (execa, implementation) => {
+  await withContextSendingMessages(execa, async () => {
+    try {
+      await execa`endo send alice ${'Please enjoy this @counter:doubler.'}`;
+      await execa`endo inbox --as alice-agent`;
+      await execa`endo adopt --as alice-agent 1 counter --name redoubler`;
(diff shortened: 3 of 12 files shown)


packages/cli/test/demo/names-in-transit.js
```diff
@@ -0,0 +1,41 @@
...
+ * Wraps a function with the setup and teardown of the endo object state from names-in-transit.
+ *
+ * @param {*} execa
+ * @param {() => Promise<void} implementation
+ * @returns {Promise<void>}
+ */
+export const withContext = async (execa, implementation) => {
+  await withContextSendingMessages(execa, async () => {
+    try {
+      await execa`endo send alice ${'Please enjoy this @counter:doubler.'}`;
```
rekmarks, line 32: I see that Agoric's ESLint rule `@jessie.js/safe-await-separator` is triggered here. Usually we get around this problem like so:
```suggestion
  await withContextSendingMessages(execa, async () => {
    await null;
    try {
      await execa`endo send alice ${'Please enjoy this @counter:doubler.'}`;
```

Async functions execute synchronously until they encounter their first `await`. `await null` is of course semantically meaningless, but it has the side effect of ceding control of the event loop, thus ending the "synchronous prelude" of this `async` function. The lint rule exists because it can otherwise be hard to reason about which parts of an `async` function are executed synchronously vs. asynchronously.

IMO, this is a false positive since no expressions or statements are evaluated before the first `await`. Nevertheless, if you want to silence these warnings, you can.

packages/cli/test/demo/section.js
```diff
@@ -0,0 +1,26 @@
+/**
+ *
+ * @param {*} execa - the command execution environment
+ * @param {(execa: *, testCommand: (command: *, expectation: *) => Promise<true>) => Promise<void>} testRoutine - the test logic implementation
+ * @returns {(t: *) => Promise<void>}
```
rekmarks, line 5: You should be able to add comments like these at the top of the file to get the execa types, thereby getting rid of the `*` types:
```javascript
/** @import {Execa} from 'execa' */
/** @import {t} from 'ava' */
```

You can find examples of this working successfully elsewhere in the repo.

rekmarks (commented):
I like it!

Because we're actually using the daemon in these tests, I think they are more akin to integration or even e2e tests than unit tests. That being the case, it is completely fine that they run serially. It would certainly be better for each test case to be isolated and run concurrently, but we could also consider that complication a stretch goal for the time being. It's also worth noting that, even with the utilities in the `@endo/daemon` tests, we've had issues with running tests concurrently.

In addition to my in-line comments, I would prefer it if `test/demo` only contained actual test cases, and the utility files (AFAICT `daemon.js`, `section.js`) are moved to separate folder, like `test` or `test/utils`.

packages/cli/test/demo/confined-script.js
```diff
@@ -0,0 +1,10 @@
+import { withContext as withContextDaemon } from './daemon.js';
```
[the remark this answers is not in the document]
grypez: Your proposed word ordering properly respects the english compound noun convention. I really want `test ... withContext ... <implementation>` for readability. Combined with the `await null` gotcha you mentioned below I think I have one resolution to both suggestions: `test ... withContext(<context-name>) ... <implementation>`.

packages/cli/test/demo/index.test.js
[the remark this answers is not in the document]
grypez: At the moment, yes. As far as I can tell this pattern is necessary to get ava to run the tests serially w.r.t. one another. Ideally each `section-name.js` would become `section-name.test.js`. In that ideal world devs could test the whole demo with `yarn test` or a single section as simply as `yarn test path/to/demo/section-name.test.js`.

I'll commit to the serial approach per your general review. The isolation and concurrency can be resolved in a later PR. Lack of independence is annoying too but I think is quite a ways out.

packages/cli/test/demo/names-in-transit.js
grypez: interesting, thanks

packages/cli/test/demo/section.js
grypez: will do

packages/cli/test/demo/confined-script.js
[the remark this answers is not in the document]
rekmarks: Sounds reasonable!

PR #2374, opened 2024-07-23
chore(compartment-mapper): Restore backward compatibility for search

kriskowal:
Refs: 592a485e22 #2344

## Description

#2344 introduced a regression in backward compatibility which has not yet been released. The change improved our ability to differentiate ENOENT and EISDIR from other error classes when searching for package.json. This adjusts that change to tolerate the currently accepted `read` function and allow read powers that included a more precise `maybeRead` function instead.

### Security Considerations

None.

### Scaling Considerations

None.

None.

### Testing Considerations

Coverage test added for `search`.

This restores backward compatibility.

### Upgrade Considerations

Nnoe.

### DIFF
--- a/packages/compartment-mapper/src/search.js
+++ b/packages/compartment-mapper/src/search.js
@@ -4,10 +4,13 @@
 
 // @ts-check
 
-/** @import {MaybeReadFn} from './types.js' */
+/** @import {ReadFn} from './types.js' */
+/** @import {ReadPowers} from './types.js' */
+/** @import {MaybeReadPowers} from './types.js' */
 
 import { relativize } from './node-module-specifier.js';
 import { relative } from './url.js';
+import { unpackReadPowers } from './powers.js';
 
 // q, as in quote, for enquoting strings in error messages.
 const q = JSON.stringify;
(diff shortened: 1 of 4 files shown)


packages/compartment-mapper/src/search.js
```diff
@@ -4,10 +4,13 @@
 
 // @ts-check
 
-/** @import {MaybeReadFn} from './types.js' */
+/** @import {ReadFn} from './types.js' */
+/** @import {ReadPowers} from './types.js' */
+/** @import {MaybeReadPowers} from './types.js' */
```
michaelfig, line 9: Maybe import style?
```suggestion
/** @import {ReadFn, ReadPowers, MaybeReadPowers} from './types.js' */
```

packages/compartment-mapper/test/search.test.js
```diff
@@ -0,0 +1,56 @@
...
+test('search should fail to find package.json at system root with read power', async t => {
+  const readPowers = makeReadPowers({
+    fs,
+    crypto,
+    url,
+  });
+  const { read } = readPowers;
+  const root = new URL('/', import.meta.url);
+  await t.throwsAsync(() => search(read, root));
+});
```
michaelfig, line 46: Using the root directory makes me uncomfortable both in terms of accessing the filesystem outside of the endo checkout, as well as not being robust if somebody actually has a `/package.json`.  Can you instead fake the `fs` power provided to `makeReadPowers({ fs })`?

michaelfig (approved):
No blockers, but a suggestion for how to make `search.test.js` less reliant on files outside the endo checkout.

turadg:
> https://github.com/endojs/endo/pull/2344 introduced a regression in backward compatibility which has not yet been released

Incidentally, I think the `fix` commit message will be interpreted by the release pipeline as a "bug fixes" entry for the changelog, which won't be accurate because there was no such bug in a versioned release.

kriskowal:
> > #2344 introduced a regression in backward compatibility which has not yet been released
> 
> Incidentally, I think the `fix` commit message will be interpreted by the release pipeline as a "bug fixes" entry for the changelog, which won't be accurate because there was no such bug in a versioned release.

Switched to `chore` and feedback welcome if there’s a more specific. My vocabulary is limited to `fix`, `feat`, `chore`, `test`, `docs`, and `refactor`.

PR #2377, opened 2024-07-23
feat(import-bundle): accept symbol-named properties on inescapableGlobalProperties

### LINKED ISSUE #2376: import-bundle: allow symbol-named properties in inescapableGlobalProperties

As discussed in the comments on #1676, we'd like to convey `passStyleOf` into vat environments via a symbol-named property on the global, rather than a string-named property on VatData.

To support this, we need to change `import-bundle` to pay attention to symbol-named properties of `inescapableGlobalProperties`, because currently it uses `Object.keys`, which ignores symbols and only copies the string-named properties into the child Compartment's global. We'll accomplish this by using `Reflect.ownKeys`

While we're at it, we'll change `import-bundle` to look for *all* own properties of `inescapableGlobalProperties`, not just the enumerable ones (which matches the change from `Object.keys` to `Reflect.ownKeys`).

## Description of the Design

Add tests. Change it to use `Reflect.ownKeys`.

## Security Considerations

None I can think of, whoever is calling `importBundle` is in control of what properties they put on `inescapableGlobalProperties`, so if they want to make those properties unenumerable or Symbols, that's their business. And that party is in complete control over the guest/Compartment code's world.

## Scaling Considerations

none

## Test Plan

unit tests

## Compatibility Considerations

In the unlikely event that someone was already using Symbol-named properties, or unenumerable ones, in their `inescapableGlobalProperties` call, their guest code will start getting access to new properties, perhaps suprisingly. But that seems quite unlikely.

## Upgrade Considerations

To benefit from this, the host code must start using a new version of `import-bundle`. For the Agoric chain, that means a new version of `swingset-xsnap-supervisor`, so the process is:

* do a chain-halting upgrade that includes the new supervisor
* create a new vat, or upgrade an existing vat, both of which will then use the new supervisor code
* (also use a new liveslots, to get the `passStyleOf` endowment, and a new/current `@endo/pass-style`, which pulls that endowment)

warner:
Previously, the `inescapableGlobalProperties` option to `importBundle()` only accepted enumerable string-named properties. This commit changes that to accept all properties (even unenumerable ones), and to accept symbol-named properties (instead of only string-named ones).

We'll use this to provide `passStyleOf` to the vat environment built by liveslots, so it can use a real WeakMap, instead of the (slower) "virtual object -aware WeakMap".

closes #2376

This modifies the behavior of `options.inescapableGlobalProperties`, which is undocumented. We should probably write docs for it at some point.

### Upgrade Considerations

See #2376 for notes about compatibility considerations, probably nothing worth calling out in NEWS.md .


erights (approved):
LGTM, thanks!

packages/import-bundle/src/compartment-wrapper.js
```diff
@@ -42,11 +42,14 @@ export function wrapInescapableCompartment(
...
+    // string-named and symbol-named properties. Note that
+    // Reflect.ownKeys also includes non-enumerable keys.
+    for (const prop of Reflect.ownKeys(inescapableGlobalProperties)) {
       Object.defineProperty(c.globalThis, prop, {
         value: inescapableGlobalProperties[prop],
         writable: true,
-        enumerable: false,
+        enumerable: false, // TODO: really? why?
         configurable: true,
       });
```
mhofman, line 54: Should we just plainly copy the descriptor? Maybe verifying it's a data property first, but I'm not even convinced that's needed

erights: If we copy the descriptor, then we've potentially copied getter/setter functions written assuming a different `this`.

When we looked at this together, I at first assumed that before this PR it was `enumerable: true`, to go with `Object.keys`, in which case I was about to suggest changing to preserve the non-enumerability of newly revealed non-enumerable properties. But then when I saw that it was already `enumerable: false` for everything, I decided to leave it alone, since it is being added to a global object.

But, I suppose this should be aligned with the Compartment constructor endowment handling. Sigh. Can we make such adjustments in a distinct PR?

@kriskowal , please weigh in on Compartment constructor handling of endowed properties.

kriskowal: `Compartment` uses `Object.assign` for the endowments (`globals`). If we wanted property grafting, I’d create different option `globalProperties` since `assign` is a safer default and properties are useful elsewhere. Correspondingly, `inescapableGlobals` would assign and `inescapableGlobalProperties` would graft, and this is entirely coherent.

I like @mhofman’s suggestion, with the understanding that inescapable getters and setters are sensitive to `this` binding.

kriskowal (changes requested):
I would like to adopt @mhofman’s suggestion of fully grafting all `inescapableGlobalProperties`.

packages/import-bundle/src/compartment-wrapper.js
mhofman: If we have 2 options, what would be the precedence of one vs the other if both define the same property? or is that an error?

erights:
Ok. I made https://github.com/Agoric/agoric-sdk/pull/9762 back into a Draft until this one settles down.

But https://github.com/Agoric/agoric-sdk/pull/9762 should still be perfectly adequate for realistic performance testing of the point of this change!

packages/import-bundle/src/compartment-wrapper.js
kriskowal: It’s a question we would have to answer.

gibson042: In a fresh environment, most properties of the global object are non-enumerable (including, I believe, all properties defined at the language level). And the list of exceptions is idiosyncratic and—outside of web browsers—rather small (e.g., Node.js has `['global', 'clearImmediate', 'setImmediate', 'clearInterval', 'clearTimeout', 'setInterval', 'setTimeout', 'queueMicrotask', 'structuredClone', 'atob', 'btoa', 'performance', 'fetch', 'crypto', '__filename', 'module', 'exports', '__dirname', 'require']`, and XS for eshost has no enumerable global properties).

I therefore think the current default is reasonable, and does not preclude a future extension that opts in to different treatment such as preserving property attributes.

packages/import-bundle/src/compartment-wrapper.js
```diff
@@ -42,11 +42,14 @@ export function wrapInescapableCompartment(
...
-    for (const prop of Object.keys(inescapableGlobalProperties)) {
+    // Use Reflect.ownKeys, not Object.keys, because we want both
+    // string-named and symbol-named properties. Note that
+    // Reflect.ownKeys also includes non-enumerable keys.
+    for (const prop of Reflect.ownKeys(inescapableGlobalProperties)) {
       Object.defineProperty(c.globalThis, prop, {
         value: inescapableGlobalProperties[prop],
         writable: true,
-        enumerable: false,
+        enumerable: false, // TODO: really? why?
```
kriskowal, line 52:
```suggestion
        enumerable: false, // properties of globalThis are generally non-enumerable
```

kriskowal (approved):

gibson042:
This is in support of https://github.com/Agoric/agoric-sdk/issues/9781

packages/import-bundle/src/compartment-wrapper.js
erights: Resolved according to verbal agreement with @kriskowal stated in the extended comment in this PR's code, based on a variant of the meeting at https://www.youtube.com/watch?v=xlR21uDigGE

packages/import-bundle/src/compartment-wrapper.js
erights: Done

PR #2380, opened 2024-07-25
feat(ses): Opt out of namespace boxes

kriskowal:
Refs: #400 

## Description

To reduce mismatched behavior between SES and XS, this change adds an option (to be removed at the next major version) that changes the compartment import method such that it returns a promise for an exports namespace instead of a promise for a box containing the imports namespace. This opts-in to the same thenable behavior as dynamic import in the language.

### Security Considerations

With this option, we must be weary of thenable modules, as we must be for thenable modules across the rest of the JavaScript ecosystem.

### Scaling Considerations

None.

This option will necessarily appear in all portable code examples on hardenedjs.org once it is available.

### Testing Considerations

Tests have been adjusted to take advantage of the new option, except legacy tests which remain to test the behavior without the option enabled.

This is a backward compatible change. Namespace boxes are hereafter deprecated and we hope to remove this option on the next major release, making the new behavior mandatory or at least opt-out.

### Upgrade Considerations

None.

### DIFF
--- a/packages/ses/src/compartment.js
+++ b/packages/ses/src/compartment.js
@@ -172,7 +172,11 @@ export const makeCompartmentConstructor = (
   markVirtualizedNativeFunction,
   parentCompartment = undefined,
 ) => {
-  function Compartment(endowments = {}, moduleMap = {}, options = {}) {
+  function Compartment(
+    endowmentsOption = {},
+    moduleMapOption = {},
+    options = {},
+  ) {
     if (new.target === undefined) {
       throw TypeError(
         "Class constructor Compartment cannot be invoked without 'new'",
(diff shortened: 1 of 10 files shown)


packages/ses/src/compartment.js
```diff
@@ -189,8 +200,11 @@ export const makeCompartmentConstructor = (
       importNowHook,
       moduleMapHook,
       importMetaHook,
+      __noNamespaceBox__: noNamespaceBox = false,
     } = options;
     const globalTransforms = [...transforms, ...__shimTransforms__];
+    const endowments = { __proto__: null, ...endowmentsOption };
+    const moduleMap = { __proto__: null, ...moduleMapOption };
```
erights, line 207: Are these two lines drive-bys, or somehow related to this PR?

erights: Nevermind. Deleting.

erights (approved):
LGTM!

PR #2381, opened 2024-07-25
feat: add reflect metadata taming

Jack-Works:
Refs: #1950

## Description

reflect-metadata is a widely used package in the ecosystem. It is used to polyfill an outdated ECMAScript feature about decorators.

It modifies the global `Reflect` namespace object to add several methods. Although in https://github.com/rbuckton/reflect-metadata/pull/131, @rbuckton added a `/no-conflict` version of this package, it is impossible to upgrade every package in the ecosystem.

This PR adds an option called `reflectMetadataTaming`. It has 3 valid options:

- `none` (default): Current behavior, not try to fix it and cause `reflect-metadata` throw.
- `keep-and-inherit`: If the reflect-metadata methods are added **before** the lockdown is called, they will be preserved like they are safe primorials. They will also be inherited by any subcompartments.
- `mutable-per-global`: The Reflect namespace is an exotic object and unique per-global. Reflect-metadata can set (prior 0.2.0) or define methods as it wishes.

Example:

> `keep-and-inherit`, install polyfill before lockdown

```js
import './lockdown.umd.js'
import 'https://cdnjs.cloudflare.com/ajax/libs/reflect-metadata/0.2.2/Reflect.js'
lockdown({ reflectMetadataTaming: 'keep-and-inherit', errorTaming: 'unsafe', consoleTaming: 'unsafe' })
console.log(Reflect.metadata) // function

globalThis.c = new Compartment()
console.log(c.globalThis.Reflect === Reflect) // true
```

> `keep-and-inherit`, install polyfill after lockdown

```js
import './lockdown.umd.js'
lockdown({ reflectMetadataTaming: 'keep-and-inherit', errorTaming: 'unsafe', consoleTaming: 'unsafe' })
await import('https://cdnjs.cloudflare.com/ajax/libs/reflect-metadata/0.2.2/Reflect.js')
// Cannot define property decorate, object is not extensible
```

> `mutable-per-global`, install polyfill inside subcompartment, but not the initial one

```js
import './lockdown.umd.js'
lockdown({ reflectMetadataTaming: 'mutable-per-global', errorTaming: 'unsafe', consoleTaming: 'unsafe' })

globalThis.c = new Compartment()
console.log(c.globalThis.Reflect === Reflect) // false

var data = await fetch('https://cdnjs.cloudflare.com/ajax/libs/reflect-metadata/0.2.2/Reflect.js').then(x => x.text())
c.globalThis.eval(data)
console.log(c.globalThis.Reflect.metadata) // function
console.log(Reflect.metadata) // undefined
```

### Security Considerations

When using `keep-and-inherit`, all compartment and the initial compartment shares the Metadata Registry (which is viable by `Reflect[Symbol.for('@reflect-metadata:registry')]`). They may exchange objects via this registry and break the membrane if there is one.

### Scaling Considerations

No problem.

If you're hitting problems with `reflect-metadata`, you might need this.

### Testing Considerations

I have not added any tests yet. I hope I can iterate the PR design with the endo team and write the tests after the design is stable.

No breaking changes.

### Upgrade Considerations

No breaking changes.

TODO.

### DIFF
--- a/packages/ses/src/commons.js
+++ b/packages/ses/src/commons.js
@@ -90,6 +90,12 @@ export const { stringify: stringifyJson } = JSON;
 // Needed only for the Safari bug workaround below
 const { defineProperty: originalDefineProperty } = Object;
 
+/**
+ * @param {object} object
+ * @param {PropertyKey} prop
+ * @param {PropertyDescriptor} descriptor
+ * @returns
+ */
 export const defineProperty = (object, prop, descriptor) => {
   // We used to do the following, until we had to reopen Safari bug
   // https://bugs.webkit.org/show_bug.cgi?id=222538#c17
--- a/packages/ses/src/compartment-shim.js
+++ b/packages/ses/src/compartment-shim.js
@@ -12,4 +12,5 @@ globalThis.Compartment = makeCompartmentConstructor(
   makeCompartmentConstructor,
   getGlobalIntrinsics(globalThis),
   markVirtualizedNativeFunction,
+  'none',
 );
--- a/packages/ses/src/compartment.js
+++ b/packages/ses/src/compartment.js
@@ -25,6 +25,7 @@ import { link } from './module-link.js';
 import { getDeferredExports } from './module-proxy.js';
 import { compartmentEvaluate } from './compartment-evaluate.js';
 import { makeSafeEvaluator } from './make-safe-evaluator.js';
+import { setGlobalObjectReflectNamespace } from './make-reflect-namespace.js';
 
 // moduleAliases associates every public module exports namespace with its
 // corresponding compartment and specifier so they can be used to link modules
@@ -168,6 +169,7 @@ defineProperties(InertCompartment, {
  * @param {MakeCompartmentConstructor} targetMakeCompartmentConstructor
  * @param {Record<string, any>} intrinsics
  * @param {(object: object) => void} markVirtualizedNativeFunction
+ * @param {import('../types').RepairOptions['reflectMetadataTaming']} reflectMetadataTaming
  * @param {Compartment} [parentCompartment]
  * @returns {Compartment['constructor']}
  */
@@ -177,6 +179,7 @@ export const makeCompartmentConstructor = (
   targetMakeCompartmentConstructor,
   intrinsics,
   markVirtualizedNativeFunction,
+  reflectMetadataTaming,
   parentCompartment = undefined,
 ) => {
   function Compartment(
@@ -235,6 +238,7 @@ export const makeCompartmentConstructor = (
       newGlobalPropertyNames: sharedGlobalPropertyNames,
       makeCompartmentConstructor: targetMakeCompartmentConstructor,
       parentCompartment: this,
+      reflectMetadataTaming,
       markVirtualizedNativeFunction,
     });
 
@@ -244,6 +248,9 @@ export const makeCompartmentConstructor = (
       safeEvaluate,
       markVirtualizedNativeFunction,
     );
+
+    reflectMetadataTaming === 'mutable-per-global' &&
+      setGlobalObjectReflectNamespace(globalObject);
 
     assign(globalObject, endowments);
 
--- a/packages/ses/src/enablements.js
+++ b/packages/ses/src/enablements.js
@@ -4,7 +4,7 @@ import { toStringTagSymbol, iteratorSymbol } from './commons.js';
  * @file Exports {@code enablements}, a recursively defined
  * JSON record defining the optimum set of intrinsics properties
  * that need to be "repaired" before hardening is applied on
- * enviromments subject to the override mistake.
+ * environments subject to the override mistake.
  *
  * @author JF Paradis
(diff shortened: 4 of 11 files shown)


erights (commented):
Please add the word `'unsafe'` somewhere in the option currently named `'keep-and-inherit'`

Jack-Works:
> Please add the word `'unsafe'` somewhere in the option currently named `'keep-and-inherit'`

Done

erights (commented):
Hi @Jack-Works , I am reluctant to add this much complexity to the ses shim. So I'm wondering whether the following simpler approach would fulfill the actual need that has been encountered in practice:

- Have the `Reflect` in the start compartment be distinct from the `Reflect` in constructed compartments. Either like `Math` where all constructed compartments share a safe `Reflect`, or like `Function` where each constructed compartment gets its own `Reflect`. From the discussion at the last tc39 meeting, we may be pushed into a per-constructed-compartment safe `Reflect` anyway in order to have a per-compartment `Reflect.isTemplateObject`. Attn @littledan @ljharb

- Have even the `Reflect` in the start compartment get hardened, requiring any unsafe additions to happen before the `harden` phase of `lockdown`. For the particular unsafe additions of concern here, add them unconditionally to the permits list only for this unsafe `Reflect` only for the start compartment. 

- If code in the start compartment wants to grant code in a constructed compartment access to its own unsafe `Reflect`, it does so by endowment, not by lockdown option. This recapitulates the transition we made with `Math` and `Date`, where we initially propagated the unsafe one by lockdown option, and then switched to propagation only by endowment.

- permits.js remains as declarative as it is now. No imperative side effects to it based on lockdown options.

Attn @kriskowal

ljharb:
Reflect will likely house getIntrinsic as well, so i think a per compartment Reflect is already quite inevitable.

Jack-Works:
@erights 

For my use cases, `unsafe-keep-and-inherit` is enough, 

> permits.js remains as declarative as it is now. No imperative side effects to it based on lockdown options.

this is the best minimal change I can think of. If I patch this behavior in `makeIntrinsicsCollector`, I'll get more code changes than this.

> If code in the start compartment wants to grant code in a constructed compartment access to its own unsafe Reflect, it does so by endowment, not by lockdown option. This recapitulates the transition we made with Math and Date, where we initially propagated the unsafe one by lockdown option, and then switched to propagation only by endowment

Can you give an example what this might be look like in the user land?

erights:
> For my use cases, unsafe-keep-and-inherit is enough,

Excellent!


> Can you give an example what this might be look like in the user land?

```js
const compartment = new Compartment({
    globals: { Reflect },
    __options__: true, // only needed during a compatibility transition
  });
```

Whatever `Reflect` is already bound to in this calling code, it passes as the `Reflect` endowment for the globals of the newly constructed compartment.

kriskowal:
@erights If I read this correctly, this feature can be implemented with very little behavior contingent on a lockdown option.

1. Separating %Reflect% from %SharedReflect% is safe in general. We can visit a per-compartment Reflect when we realize a future need.
2. Leaving %Reflect% between repairIntrinsics and hardenIntrinsics is the normal behavior.

Adding `%Reflect%.metadata` to the permits might introduce a hazard in the event the `Reflect` is later hardened and endowed. Do we need an option?

erights:
> Adding %Reflect%.metadata to the permits might introduce a hazard in the event the Reflect is later hardened and endowed. Do we need an option?

If the `reflect-metadata` package runs after `repairIntrinsics` as a vetted shim, then it runs after the permit processing, so we do not need to add anything to permits.js to accommodate all the stuff it monkey patches in to the start compartment's `Reflect`.

Right now, we don't support (well?) clients of ses running outside libraries as vetted shims between `repairIntrinsics` and `hardenIntrinsics`. But if we can somehow arrange that the flag leaves the start compartment's `Reflect` mutable even after `hardenIntrinsics`, then `reflect-metadata` can run after `lockdown`, still without needing to accommodate any added property in `permits.js`.

As you say, we do need `permits.js` to distinguish the start compartment's `Reflect` from the `Reflect` shared by constructed compartments.

erights:
Hi @Jack-Works , What is the status of this?

Jack-Works:
> Hi @Jack-Works , What is the status of this?

Hi! I've been recently working on other projects, but this one is still on my TODO list. I'll keep it up to date after I come back to this.

Jack-Works:
Hi! I can now get back to this PR.

I reviewed the discussion before, and here is a summary:

1. We may have per-compartment Reflect objects because of `isTemplateObject` and `getIntrinsics`.
2. I need an unsafe option to keep additional properties on the Reflect object of the initial compartment.
3. If the initial compartment wants to share the metadata registry with a subcompartment, they can do it via endowments, which is a user-land solution. (That requires subcompartment.globalThis.Reflect is configurable).
4. We may need an option to allow a subcompartment to install its own Reflect metadata polyfill. This may be the same option as (2).

I wonder if there are big changes in endo.js/SES this 2 years, and do you still accept features like this.

Issue #621, opened 2021-03-17
Compatibility: regenerator-runtime


kumavis:
conflict is `Object.prototype.constructor`

metamask-extension3/node_modules/babel-runtime/node_modules/regenerator-runtime/runtime.js
https://npmfs.com/package/regenerator-runtime/0.11.1/runtime.js#L105
```js
function Generator() {}
function GeneratorFunction() {}
function GeneratorFunctionPrototype() {}

// This is a polyfill for %IteratorPrototype% for environments that
// don't natively support it.
var IteratorPrototype = {};
IteratorPrototype[iteratorSymbol] = function () {
  return this;
};

var getProto = Object.getPrototypeOf;
var NativeIteratorPrototype = getProto && getProto(getProto(values([])));
if (NativeIteratorPrototype &&
    NativeIteratorPrototype !== Op &&
    hasOwn.call(NativeIteratorPrototype, iteratorSymbol)) {
  // This environment has a native %IteratorPrototype%; use it instead
  // of the polyfill.
  IteratorPrototype = NativeIteratorPrototype;
}

var Gp = GeneratorFunctionPrototype.prototype = Generator.prototype = Object.create(IteratorPrototype);
GeneratorFunction.prototype = Gp.constructor = GeneratorFunctionPrototype; // < ----------- explodes here 
```

erights:
Closing as won't fix. But see https://github.com/endojs/endo/issues/576#issuecomment-800840188
We should report this bug upstream to regenerator-runtime.

erights:
Hi @Jack-Works , 
Reopening, thanks.
Do you have a taming in mind? Happy to review a PR.

PR #2383, opened 2024-07-26
feat: add regenerator runtime taming

Jack-Works:
Closes: #621
Refs: #1950

## Description

regenerator-runtime is a widely used package in the ecosystem. It is used to support generators and async functions transpiled to ES5.

This PR adds an option `legacyRegeneratorRuntimeTaming` to fix `regenerator-runtime` from 0.10.5 to 0.13.7. Although the newer version of the regenerator runtime package is compatible with lockdown, some libraries bundle old (hence "legacy") regenerator runtime in their code and it's not practical to get them all to upgrade.

- `legacyRegeneratorRuntimeTaming: 'safe'` do nothing.
- `legacyRegeneratorRuntimeTaming: 'unsafe-ignore'` turns `Iterator.prototype[@@iterator]` to a funky accessor that drops all assignments to it.

Note: `regenerator-runtime` is doing this:

```js
Gp[iteratorSymbol] = function () {
    return this;
}
```
which is effectively
```js
IteratorPrototype[Symbol.iterator] = function () {
    return this;
}
```

### Security Considerations

The replacement function from legacy regenerator runtime is the same as the native code, so it is "safe" to drop this assignment, in the sense that it does not cause any bad effects. However, this option drops the assignment by dropping any assignment to `IteratorPrototype[Symbol.iterator]`, since we have no practical way to ensure that the assignment it drops is exactly the one above. Thus, this option is not actual safe since it causes any other such assignment to be ignored silently. This echoes the unsafety of ES3 and of sloppy mode, where failed assignments were silently ignored. Such behavior is unsafe because it allows control flow to proceed into code that assumes the assignment succeeded. That's why ES5 strict mode changed failed assignments to throw. 

To emphasize the hazard, we have named this setting of the option `'unsafe-ignore'`.

### Scaling Considerations

Nothing

If you're hitting problems with an old version of regenerator-runtime (or any package that bundles it), you might need this.

### Testing Considerations

Tests added for the specific effect of this PR. However, to avoid introducing even a devDependency on a legacy version of regenerator runtime, no automated test has been added to test that compatibility. Instead, this PR can be tested like this:

```js
import './lockdown.umd.js'
lockdown({ legacyRegeneratorRuntimeTaming: 'unsafe-ignore', errorTaming: 'unsafe', consoleTaming: 'unsafe' })
const script = document.createElement('script')
script.src = 'https://cdn.jsdelivr.net/npm/regenerator-runtime@0.13.7/runtime.js'
document.head.appendChild(script)
```

Note: Some version of `regenerator-runtime` requires to be run in the sloppy mode. Thus, these are incompat with the ses-shim independent of this option.

### Upgrade Considerations

No

TODO.


packages/ses/types.d.ts
```diff
@@ -33,6 +33,12 @@ export interface RepairOptions {
   overrideTaming?: 'moderate' | 'min' | 'severe';
   overrideDebug?: Array<string>;
   domainTaming?: 'safe' | 'unsafe';
+  /**
+   * none (default): do nothing.
+   *
+   * safe: make %IteratorPrototype%[@@iterator] to a funky accessor which ignores all assignments.
+   */
```
kriskowal, line 40: I’m impressed that this is all that is necessary to compensate for the regenerator runtime. This looks a great deal like one of our “enablements” and maybe we can harness that.

I would like a test that embeds regenerator runtime as a shim to verify that this is sufficient.

kriskowal (commented):
I’d like to get @erights opinion. Given that the option is safe, I also want to consider automatically detecting the regenerator runtime and applying this mitigation without a feature flag.

packages/ses/types.d.ts
Jack-Works: I'm not clear about what is "enablements", would you like to explain?

> I would like a test that embeds regenerator runtime as a shim to verify that this is sufficient.

Yes I have given the code.

```js
import './lockdown.umd.js'
lockdown({ regeneratorRuntimeTaming: 'safe', errorTaming: 'unsafe', consoleTaming: 'unsafe' })
const script = document.createElement('script')
script.src = 'https://cdn.jsdelivr.net/npm/regenerator-runtime@0.13.7/runtime.js'
document.head.appendChild(script)
```

erights: > I'm not clear about what is "enablements", would you like to explain?

"Fixes" enumerated cases of the override mistake by turning those data properties into accessor properties that emulate how a data property would have acted without the override mistake.

Please see 
https://github.com/endojs/endo/blob/master/packages/ses/src/enablements.js
https://github.com/endojs/endo/blob/master/packages/ses/src/enable-property-overrides.js
and, for example
https://github.com/endojs/endo/blob/master/packages/ses/src/enable-property-overrides.js

At first I liked @kriskowal suggestion, which would amount to adding the line
```js
    [iteratorSymbol]: true,
```
after line 175 https://github.com/endojs/endo/blob/59b756eb896c72a75f4c0773f964903364bf7e71/packages/ses/src/enablements.js#L175

However, then I realized that the problem that this PR fixes is not an assignment to a mutable object that inherits the property from a frozen object. Rather, it is a pointless assignment to the frozen prototype object itself, which cannot succeed. The enablement mechanism would still make this fail by throwing, as it must in the general case. This PR makes *that* assignment fail silently rather than throwing. We cannot make this the default behavior, because making an assignment silently fail is ***not*** safe. It causes control flow to proceed on paths that assume the assignment succeeded. When we introduced strict mode in ES5, that's why we made failed assignments throw. Because the ES3 sloppy behavior was ***unsafe***.

> I also want to consider automatically detecting the regenerator runtime and applying this mitigation without a feature flag.

In @Jack-Works 's example, the regenerator runtime is not initialized until ***after*** `lockdown`. If we take that as a constraint on the problem we're solving, I don't see a way to get the effect by automatically detecting, at least without violating a harder taboo: Making behavior of the primordials vary depending on what code gets loaded after `lockdown`. We clearly cannot do that.

So I am ok with the overall shape of this PR.

packages/ses/src/lockdown.js
```diff
@@ -180,12 +181,20 @@ export const repairIntrinsics = (options = {}) => {
...
+      'LOCKDOWN_REGENERATOR_RUNTIME_TAMING',
+      'none',
+    ),
     __hardenTaming__ = getenv('LOCKDOWN_HARDEN_TAMING', 'safe'),
     dateTaming = 'safe', // deprecated
     mathTaming = 'safe', // deprecated
     ...extraOptions
   } = options;
 
+  regeneratorRuntimeTaming === 'safe' ||
```
erights, line 194: The behavior this flag causes is ***not*** safe, since it causes assignments to fail silently, allowing control flow to proceed into code that assumed the assignment succeeded. When we introduced strict mode in ES5, that's why we made failed assignments throw. Because the ES3 sloppy behavior was ***unsafe***.

I understand that silently ignoring the particular problematic assignment done by legacy regenerator runtime is safe, but I don't see any good way to make the behavior change without the hazard of allowing other innocent assignments to also fail silently.

I suggest renaming the `'none'` option to `'safe'` and renaming this `'safe'` option to `'unsafe'` or `'unsafe-something'`.

packages/ses/src/lockdown.js
```diff
@@ -180,12 +181,20 @@ export const repairIntrinsics = (options = {}) => {
       /** @param {string} debugName */
       debugName => debugName !== '',
     ),
+    regeneratorRuntimeTaming = getenv(
+      'LOCKDOWN_REGENERATOR_RUNTIME_TAMING',
+      'none',
+    ),
```
erights, line 187:
```suggestion
    legacyRegeneratorRuntimeTaming = getenv(
      'LOCKDOWN_LEGACY_REGENERATOR_RUNTIME_TAMING',
      'safe',
    ),
```

Since this is no longer necessary for the current regenerator runtime, I suggest including "legacy" somewhere in the name.

On changing `'none'` to `'safe'`, see comment on line 194 below.

packages/ses/src/tame-regenerator-runtime.js
```diff
@@ -0,0 +1,16 @@
...
+} from './commons.js';
+import { accessor, permitted } from './permits.js';
+
+export const tameRegeneratorRuntime = () => {
+  const iter = iteratorPrototype[iteratorSymbol];
+  permitted['%IteratorPrototype%']['@@iterator'] = accessor;
+  defineProperty(iteratorPrototype, iteratorSymbol, {
+    configurable: true,
+    get: () => iter,
+    set: () => true,
```
erights, line 14: This is more unsafe than it needs to be to solve the motivating problem. When the setter's `this` is exactly the expected prototype, this should fail silently. This does introduce an unsafety, but only to code that would otherwise have failed.

Aside from that case, it should do the remainder of the enable-property-override setter fix at
https://github.com/endojs/endo/blob/59b756eb896c72a75f4c0773f964903364bf7e71/packages/ses/src/enable-property-overrides.js#L108-L121

so that other assignment that should fail with a throw still do so, and that other assignments that should succeed still do.

packages/ses/types.d.ts
erights: I appreciate that the code snippet you gave above does serve as a manual test. But we should also have an automated ava test as part of the PR, so it becomes a perpetual part of CI. IMO, that test does not need to actually introduce a devDependency on a legacy version of regenerator runtime, though @kriskowal may disagree. (@kriskowal ?) The test I have in mind could just test the particular assignments in question, with a comment about which one is motivated by legacy regenerator runtime.

erights (changes requested):

packages/ses/src/tame-regenerator-runtime.js
erights: Since this is distinct from the general enablements mechanism. Feel free to omit the debug clause on lines 111..114 above

Jack-Works:
hi, @erights I have updated the option name with a `legacy` prefix and changed `none` to `safe`, and `safe` to `unsafe-ignore`. I also changed how the setter behaves. I'd like to add a test after the design is OK.

packages/ses/src/permits.js
```diff
@@ -1336,7 +1336,7 @@ export const permitted = {
 
   '%IteratorPrototype%': {
     // The %IteratorPrototype% Object
-    '@@iterator': fn,
+    '@@iterator': /** @type {any} */ (fn),
```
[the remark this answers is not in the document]
Jack-Works: This property is modified below otherwise it won't pass the check. Editing this is not good but I cannot think of how to do it under the current architecture.

https://github.com/endojs/endo/pull/2383/files#diff-8cc460b0bc4cf783d417fbe07122880176867c0b2c81878afd616f0903e13c56R11

packages/ses/src/commons.js
```diff
@@ -137,6 +137,9 @@ export const { prototype: generatorPrototype } = getPrototypeOf(
   // eslint-disable-next-line no-empty-function, func-names
   function* () {},
 );
+export const iteratorPrototype = getPrototypeOf(
+  getPrototypeOf(arrayPrototype.values()),
```
erights, line 141: For this lint error, you should feel free to just suppress it. `commons.js` is initialized early enough that for anything that happens during its initialization, this hazard is a non-issue.

```suggestion
  // eslint-disable-next-line @endo/no-polymorphic-call
  getPrototypeOf(arrayPrototype.values()),
```

packages/ses/src/tame-regenerator-runtime.js
```diff
@@ -0,0 +1,31 @@
...
+  permitted['%IteratorPrototype%']['@@iterator'] = accessor;
+  defineProperty(iteratorPrototype, iteratorSymbol, {
+    configurable: true,
+    get: () => iter,
+    set(value) {
+      // ignore the assignment on IteratorPrototype
+      if (this === iteratorPrototype) return true;
+      if (objectHasOwnProperty(this, iteratorSymbol)) {
+        this[iteratorSymbol] = value;
+        return true;
```
erights, line 20: Please double check me, but I don't think a setter's return result is of any significance, so this line should simply be omitted. The proxy set trap is, IIRC, a different story.

```suggestion
```

packages/ses/src/tame-regenerator-runtime.js
```diff
@@ -0,0 +1,31 @@
...
+      }
+      defineProperty(this, iteratorSymbol, {
+        value,
+        writable: true,
+        enumerable: true,
+        configurable: true,
+      });
+      return true;
```
erights, line 28: Likewise

erights (commented):
Hi Jack, everything here looks good to me (modulo minor suggestions). Once there's a test case, I'll be happy to approve. Thanks!

packages/ses/src/tame-regenerator-runtime.js
Jack-Works: the linter errors if I don't add one

```
 Warning:   15:10  warning  Expected to return a value at the end of method 'set'  consistent-return
```

Jack-Works:
> Hi Jack, everything here looks good to me (modulo minor suggestions). Once there's a test case, I'll be happy to approve. Thanks!

Hi! I added tests!

packages/ses/test/tame-legacy-regenerator-helper.test.js
```diff
@@ -0,0 +1,53 @@
...
+  t.is(
+    Function.prototype.toString.call(set),
+    'function set() { [native code] }',
+  );
+
+  const child = Object.create(IteratorProto);
+  child[Symbol.iterator] = 'foo'; // override test
+  t.is(child[Symbol.iterator], 'foo');
+
+  const native = get();
```
erights, line 29: Thanks for testing this. It surprised me that this worked. But then I looked at the getter installed by `enable-property-overrides.js` and realized that it would work for those too. So no change suggested.

packages/ses/test/tame-legacy-regenerator-helper.test.js
```diff
@@ -0,0 +1,53 @@
...
+
+  const { get, set } = desc;
+  t.is(
+    Function.prototype.toString.call(get),
+    'function get() { [native code] }',
+  );
+  t.is(
+    Function.prototype.toString.call(set),
+    'function set() { [native code] }',
+  );
```
erights, line 23:
```suggestion
  const { get } = desc;
```

Once you move  tameRegeneratorRuntime immediately below the call to enablePropertyOverrides as suggested above, these tests would no longer pass. At first I was about to suggest testing the new toString behavior, but then I remembered that ecma262 makes that optional and Moddable XS makes use of that to omit source strings from the compiled form. So I think you can just drop these. IIRC, we don't test the printed form of the getter and setters installed by enable-property-overrides.js, which would be the closest analogy.

packages/ses/test/tame-legacy-regenerator-helper.test.js
```diff
@@ -0,0 +1,53 @@
...
+
+  const native = get();
+  IteratorProto[Symbol.iterator] = function () {
+    return this;
+  };
+  t.is(get(), native);
+  t.is(
+    Function.prototype.toString.call(native),
+    'function [Symbol.iterator]() { [native code] }',
+  );
```
erights, line 37: This test is still appropriate and I'm glad to see it. No change suggested.

Jack-Works:
thanks for the review! I think I fixed everything you have mentioned.

erights (approved):
LGTM, thanks!

erights:
Hi @Jack-Works , I've LGTMed and am about to squash-and-merge. But before I do, I'm going to make some minor edits to the PR comment, which will then also be in the commit comment.

Thanks!

erights:
Hi @Jack-Works , I notice you wrote

> > Update NEWS.md for user-facing changes.
>
> TODO.

Quite right. This reminds me that https://github.com/endojs/endo/blob/master/packages/ses/docs/lockdown.md will also need updating. I'm going to go ahead and merge this without either, but please do take a look at these. (attn @kriskowal )

kriskowal:
Yes, please circle back and update `lockdown.md` and `NEWS.md`.

PR #2384, opened 2024-07-26
test(cli): Add demo doc alignment test, serial implementation

grypez:
Closes: #2368
Refs: #2372

## Description

A test suite which asserts that the cli demo produces its documented outputs.

Current weaknesses in the implementation include concurrency and independence. They are related and both should be resolved by a change to the daemon initialization in the test setups.

Note that despite these weaknesses, this setup is useful enough to reveal that certain cli commands do not behave as the demo supposes. Notably:
- `endo list --as alice-agent` is rejected because `--as` is not a supported option for `endo list`. This could have been caught with unit tests on the cli endpoints.
- `endo inbox` does not display messages as expected. In some cases the template string for the message has simply been altered, but the discrepancy in the mailboxes-are-symmetric section is unlikely to have been caught with a unit test.

### Testing Considerations

The `familiar-chat` section is not tested.
Three of the section tests are failing, which is correct test behavior.

#### Concurrency
Ideally each tested section would run concurrently, however as written the demo sections are implicitly serial. This is a nice UX for a demo and I don't suggest changing it.

Thus tests running on the same daemon which reference the same objects interfere with one another. Possible solutions:
1. Run a different endo daemon per test, configured to its own directory
2. Test a slightly modified version of each section which prefixes petnames with a per-test unique id
3. Run the tests serially
4. Some other solution of which I am unaware

In this PR I have opted for (3.) _test the sections serially_. The logic to test each section is written in a distinct file bearing the name of that section, but the routines are imported into a common `index.test.js` and declared to ava as serial.

In a future PR it may be nice to adopt (1.), reusing setup from `daemon/test/endo.test.js`. I think [prepareConfig](https://github.com/endojs/endo/blob/8bebd9416cca3199dba5d5655d98046e2c7781e7/packages/daemon/test/endo.test.js#L209) accomplishes what I have in mind.

#### Independence
Although the tests are run serially, each depended-upon test file `X` exports setup and teardown routines in a `context` object which can be passed to `withContext` to make a wrapper. This wrapper can be applied to a test routine `Y` to replay the state-changing commands from the test `X` before calling `Y`, then clean up the state changes from `X` afterward. 

The tests are all declared to ava in the same `index.test.js` file, but section tests can be selected independently with `yarn test --match <section-name-1> [--match <section-name-2> [...]]`, for example `yarn test --match doubler-agent --match messages-are-symmetric`. The independence is weak, as any breakage in the state changing commands in the previous sections will cause the setup for subsequent tests to fail. Stronger test independence is ideal, but probably a long way away.

### Upgrade Considerations

* Adds the execa dev dependency.

### DIFF
--- a/packages/cli/test/daemon-context.js
+++ b/packages/cli/test/daemon-context.js
@@ -1,0 +1,18 @@
+/** @import { Context } from './types' */
+
+/**
+ * Provides test setup and teardown hooks that purge the local endo
+ * daemon. In the future, we should create isolated daemon instances
+ * so that tests can be run in parallel.
+ *
+ * @type {Context}
+ */
+export const daemonContext = {
+  setup: async execa => {
+    await execa`endo purge -f`;
+    await execa`endo start`;
+  },
+  teardown: async execa => {
+    await execa`endo purge -f`;
+  },
+};
--- a/packages/cli/test/demo/index.test.js
+++ b/packages/cli/test/demo/index.test.js
@@ -3,6 +3,12 @@ import { $ } from 'execa';
 import { makeSectionTest } from '../section.js';
 import { withContext } from '../with-context.js';
 import { daemonContext } from '../daemon-context.js';
+import * as counterExample from './counter-example.js';
+import * as doublerAgent from './doubler-agent.js';
+import * as confinedScript from './confined-script.js';
+import * as sendingMessages from './sending-messages.js';
+import * as namesInTransit from './names-in-transit.js';
+import * as mailboxesAreSymmetric from './mailboxes-are-symmetric.js';
 
 test.serial(
   'trivial',
@@ -12,5 +18,68 @@ test.serial(
       const maxim = 'a failing test is better than failure to test';
       await testLine(execa`echo ${maxim}`, { stdout: maxim });
     }),
+  ),
+);
+
+test.serial(
+  'counter-example',
+  makeSectionTest(
+    $({ cwd: 'demo' }),
+    withContext(daemonContext)(counterExample.section),
+  ),
+);
+
+test.serial(
+  'doubler-agent',
+  makeSectionTest(
+    $({ cwd: 'demo' }),
+    withContext(daemonContext, counterExample.context)(doublerAgent.section),
+  ),
+);
+
+test.serial.failing(
+  'sending-messages',
+  makeSectionTest(
+    $({ cwd: 'demo' }),
+    withContext(
+      daemonContext,
+      counterExample.context,
+      doublerAgent.context,
+    )(sendingMessages.section),
+  ),
+);
+
+test.serial.failing(
+  'names-in-transit',
+  makeSectionTest(
+    $({ cwd: 'demo' }),
+    withContext(
+      daemonContext,
+      counterExample.context,
+      doublerAgent.context,
+      sendingMessages.context,
+    )(namesInTransit.section),
+  ),
+);
+
+test.serial.failing(
+  'mailboxes-are-symmetric',
+  makeSectionTest(
+    $({ cwd: 'demo' }),
+    withContext(
+      daemonContext,
+      counterExample.context,
+      doublerAgent.context,
+      sendingMessages.context,
+      namesInTransit.context,
+    )(mailboxesAreSymmetric.section),
+  ),
+);
+
+test.serial(
+  'confined-script',
+  makeSectionTest(
+    $({ cwd: 'demo' }),
+    withContext(daemonContext)(confinedScript.section),
   ),
 );
--- a/packages/cli/test/demo/index.test.js
+++ b/packages/cli/test/demo/index.test.js
@@ -1,0 +1,16 @@
+import test from 'ava';
+import { $ } from 'execa';
+import { makeSectionTest } from '../section.js';
+import { withContext } from '../with-context.js';
+import { daemonContext } from '../daemon-context.js';
+
+test.serial(
+  'trivial',
+  makeSectionTest(
+    $({ cwd: 'demo' }),
+    withContext(daemonContext)(async (execa, testLine) => {
+      const maxim = 'a failing test is better than failure to test';
+      await testLine(execa`echo ${maxim}`, { stdout: maxim });
+    }),
+  ),
+);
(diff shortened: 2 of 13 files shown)


packages/cli/test/demo/names-in-transit.js
```diff
@@ -0,0 +1,31 @@
...
+export const section = async (execa, testLine) => {
+  // In this example, we send alice our "doubler" but let it appear...
+  await testLine(
+    execa`endo send alice ${'Please enjoy this @counter:doubler.'}`,
+  );
+  await testLine(execa`endo inbox --as alice-agent`, {
+    stdout: /^1\. "HOST" sent "Please enjoy this @counter\."/,
+  });
+  await testLine(execa`endo adopt --as alice-agent 1 counter --name redoubler`);
+  await testLine(execa`endo list --as alice-agent`, {
```
rekmarks, line 13: I imagine this is the first thing that fails in this test. There's an asymmetry in the API here: rather than `endo list --as <agent>`, you simply do `endo list <agent>`. Does that change cause these test cases to succeed?

rekmarks (changes requested):

packages/cli/test/demo/index.test.js
```diff
@@ -0,0 +1,85 @@
...
+    $({ cwd: 'demo' }),
+    withContext(
+      daemonContext,
+      counterExample.context,
+      doublerAgent.context,
+      sendingMessages.context,
+      namesInTransit.context,
+    )(mailboxesAreSymmetric.section),
+  ),
+);
```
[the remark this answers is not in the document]
grypez: Acknowledged, but I prefer to leave as is.

My thought is that `endo purge -f` is an overzealous command for this situation, but I have used it here erring on the side of making the tests reliable, if hazardous. The `withContext` implementation suggests that the logic for tearing down a context be declared adjacent to the logic that set it up. I see the ease on cognitive burden as worth the slight runtime overhead, and while I agree with you that the annihilating daemon teardown renders redundant the other teardowns in this particular test, I point the finger at `endo purge -f` being not the teardown of least authority. A supporting claim for my finger point is that these tests are inconsistent under concurrency if two different tests' `endo purge -f` commands annihilate the same daemon.

packages/cli/test/demo/names-in-transit.js
grypez: I think your suggestion does what you think it does, but I want to limit the scope of this PR to correctly implementing a test of whether the expectations set in the demo's README.md match the observations from running the demo's commands. A future PR could make these tests pass either by changing the behavior of endo/cli or by altering the demo's expectations. In the former case it will be satisfying to associate the fixing change to the removal of `failing` in these tests. In the latter case it will be nice to have an example PR conjoining edits to the README.md with edits to these tests.

packages/cli/test/daemon-context.js
```diff
@@ -0,0 +1,18 @@
...
+ * Provides test setup and teardown hooks that purge the local endo
+ * daemon. In the future, we should create isolated daemon instances
+ * so that tests can be run in parallel.
+ *
+ * @type {Context}
+ */
+export const daemonContext = {
+  setup: async execa => {
+    await execa`endo purge -f`;
+    await execa`endo restart`;
```
rekmarks, line 13: nit (the daemon is already stopped):
```suggestion
    await execa`endo start`;
```

packages/cli/test/demo/index.test.js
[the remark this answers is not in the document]
rekmarks: My concern isn't so much about performance as simplicity, but fair enough! I'll note a couple of things just for context (no action necessary):
- `purge` certainly is not the "teardown of least authority" (nice), it's just the only teardown available to us.
- Even if we had a finer-grained teardown function, these tests would still be inconsistent under concurrency; definitely due to the mailbox usage in multiple test cases, and probably due to other, less obvious reasons as well.

packages/cli/test/demo/names-in-transit.js
rekmarks: I see, although in this case it's the README that's wrong; `--as` was omitted from `list` in a deliberate design decision. I noticed the discrepancy myself a while back and even put up a PR for it (#2228), but we decided against making the change at the time. I don't think the situation is likely to change unless we complete a more thorough audit of each command for `--as` support, and until then I think it makes the most sense to update the readme and these tests.

packages/cli/test/with-context.js
```diff
@@ -0,0 +1,50 @@
...
+ * @returns {(testRoutine: TestRoutine) => TestRoutine}
+ */
+const makeContextWrapper =
+  context => testRoutine => async (execa, testCommand) => {
+    await null;
+    try {
+      await context.setup(execa);
+      await testRoutine(execa, testCommand);
+    } finally {
+      (await context.teardown) ?? execa;
```
rekmarks, line 15: What's the purpose of `?? execa` here? Unless I'm misreading this, it does not appear to do anything.

packages/cli/test/demo/sending-messages.js
```diff
@@ -0,0 +1,29 @@
...
+  });
+  await testLine(execa`endo dismiss --as alice-agent 0`);
+};
+
+/** @type {Context} */
+export const context = {
+  setup: async execa => {
+    await execa`endo mkguest alice alice-agent`;
+    await execa`endo send alice ${'Please enjoy this @doubler.'}`;
+    // await execa`endo adopt --as alice-agent 0 doubler`;
```
rekmarks, line 25: Is this leftover? Please delete or add context.

packages/cli/test/demo/names-in-transit.js
rekmarks: I don't think we actually resolved this point, but I'll let it go in exchange for you resurrecting #2228 😉

packages/cli/test/with-context.js
grypez: silly me, I used the `??` rune where `?.` was required

packages/cli/test/demo/sending-messages.js
grypez: deleted

rekmarks (approved):
LGTM!

PR #2386, opened 2024-07-27
feat(ses): Compartment options migration

kriskowal:
Refs: #400

## Description

To converge with XS in the patterns accepted by SES for Compartment construction, this change introduces a `__options__` property to the first argument of the Compartment constructor that indicates that the constructor accepts a single options bag argument and does not receive separate endowments and module map arguments.

### Security Considerations

None.

### Scaling Considerations

None.

This change includes updates for NEWS (with migration instructions and a notice of intent to break deprecated usage patterns) and README (to encourage recommended usage). We intend to follow up with an update to hardenedjs.org upon release.

### Testing Considerations

All tests have been updated to use the new constructor pattern except those already marked as legacy, which remain to exercise support for the deprecated usage patterns.

There is a remote possibility that existing code uses `__options__` as the name of an endowment and will now be misinterpreted. We do not imagine any guest programs are in a position to influence their host compartment’s construction.

### Upgrade Considerations

None.

### DIFF
--- a/packages/ses/src/compartment.js
+++ b/packages/ses/src/compartment.js
@@ -19,12 +19,15 @@ import {
   setGlobalObjectMutableProperties,
   setGlobalObjectEvaluators,
 } from './global-object.js';
+import { assertEqual } from './error/assert.js';
 import { sharedGlobalPropertyNames } from './permits.js';
 import { load, loadNow } from './module-load.js';
 import { link } from './module-link.js';
 import { getDeferredExports } from './module-proxy.js';
 import { compartmentEvaluate } from './compartment-evaluate.js';
 import { makeSafeEvaluator } from './make-safe-evaluator.js';
+
+/** @import {ModuleDescriptor} from '../types.js' */
 
 // moduleAliases associates every public module exports namespace with its
 // corresponding compartment and specifier so they can be used to link modules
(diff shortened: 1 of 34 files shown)


packages/ses/src/error/assert.js
```diff
@@ -568,5 +568,15 @@ const assert = makeAssert();
...
+/** @type {AssertionFunctions['typeof']} */
+const assertTypeof = assert.typeof;
+
+export {
+  assertTypeof,
+  makeError,
+  note as annotateError,
+  redactedDetails as X,
+  quote as q,
+};
```
erights, line 582: Is this a drive-by?

packages/ses/src/compartment.js
```diff
@@ -172,18 +175,63 @@ defineProperties(InertCompartment, {
...
+  }
+  if (
+    args.length === 1 &&
+    typeof args[0] === 'object' &&
+    args[0] !== null &&
+    '__options__' in args[0]
+  ) {
+    const { __options__, ...options } = args[0];
+    assert(
+      __options__ === true,
```
erights, line 199: since you require it to have value `true`, you should be sure to mention that where you explain `__option__`, such as the doc-comment above. From the doc-comment alone, I would guest that you only tested whether the property is present.

packages/ses/src/compartment.js
```diff
@@ -172,18 +175,63 @@ defineProperties(InertCompartment, {
...
+    return options;
+  } else {
+    const [
+      globals = /** @type {Map<string, any>} */ ({}),
+      modules = /** @type {Map<string, ModuleDescriptor>} */ ({}),
+      options = {},
+    ] = args;
+    assertTypeof(
+      options.modules,
+      'undefined',
```
erights, line 211: Are you sure you want to check that the value is `undefined` rather than whether the property is present?

Why typeof? The above code should be equivalent to
```suggestion
    assertEqual(
      options.modules,
      undefined,
```
but the later says what it means better.

erights (approved):
LGTM!

packages/ses/src/error/assert.js
kriskowal: No, this was my preferred solution to avoid a polymorphic dispatch at `assertTypeof` elsewhere in this change. You had good feedback elsewhere which will change this to `assertEqual` if not a more systematic change.

PR #2393, opened 2024-08-01
fix(import-bundle): Adapt inescapable compartment mapper constructor

kriskowal:
## Description

Recent changes to align the Compartment constructor in SES with its XS counterpart changed the argument pattern and this broke the interface of all extant wrappers around the Compartment constructor, of which we only know of one. This change fixes that wrapper and restores continuity of compatibility.

### Security Considerations

None.

### Scaling Considerations

None.

None.

### Testing Considerations

Inescapable global properties should cover this failure. It’s not yet clear why it did not reveal the failure earlier.

Restores compatibility. Reveals that the Compartment change broke more than we anticipated, but not enough to roll back.

### Upgrade Considerations

None.

### DIFF
--- a/packages/import-bundle/src/compartment-wrapper.js
+++ b/packages/import-bundle/src/compartment-wrapper.js
@@ -1,3 +1,54 @@
+/** @import {ModuleDescriptor} from 'ses' */
+
+// In order to facilitate migration from the deprecated signature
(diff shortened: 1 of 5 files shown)


erights (approved):
How about a test?

Otherwise LGTM. No need for another round trip to approve the test.

PR #2396, opened 2024-08-01
chore: Release 2024-08-01

kriskowal:
Posts a release to address a bugfix in `importBundle` when using `inescapableGlobalProperties` and the `endoZipBase64` format.


erights (approved):
LGTM

PR #2398, opened 2024-08-02
feat(import-bundle): Support endoScript

kriskowal:
## Description

Adds support for `endoScript` to the `importBundle` runtime.

### Security Considerations

None.

### Scaling Considerations

None.

This change makes safe the assumption that all bundles produced by `bundleSource` can be consumed by `importBundle`.

- NEWS
- README

### Testing Considerations

Unit tests included.

None.

### Upgrade Considerations

None.

### DIFF
--- a/packages/import-bundle/src/index.js
+++ b/packages/import-bundle/src/index.js
@@ -54,6 +54,8 @@ export async function importBundle(bundle, options = {}, powers = {}) {
     );
   }
 
+  let compartment;
+
   const { moduleFormat } = bundle;
   if (moduleFormat === 'endoZipBase64') {
     const { endoZipBase64 } = bundle;
@@ -76,8 +78,8 @@ export async function importBundle(bundle, options = {}, powers = {}) {
     return namespace;
   }
 
-  let c;
-  const { source, sourceMap } = bundle;
+  let { source } = bundle;
+  const { sourceMap } = bundle;
   if (moduleFormat === 'getExport') {
     // The 'getExport' format is a string which defines a wrapper function
     // named `getExport()`. This function provides a `module` to the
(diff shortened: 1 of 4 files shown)


packages/import-bundle/src/index.js
```diff
@@ -98,17 +101,27 @@ export async function importBundle(bundle, options = {}, powers = {}) {
...
-  const actualSource = `(${source})\n${sourceMap || ''}`;
-  const namespace = c.evaluate(actualSource)(filePrefix);
-  // namespace.default has the default export
-  return namespace;
+  compartment = new CompartmentToUse(endowments, {}, { transforms });
+  harden(compartment.globalThis);
+  const result = compartment.evaluate(source);
+  if (moduleFormat === 'endoScript') {
+    // The completion value of an 'endoScript' is the namespace.
+    // This format does curry the filePrefix.
```
warner, line 117: s/does/does not/ ? or did you mean the format holds the filePrefix internally somehow?

packages/import-bundle/test/import-bundle.test.js
```diff
@@ -152,6 +161,20 @@ test('inescapable global properties, nested evaluate format', async t => {
...
+    url.fileURLToPath(new URL('export-inescapable-global.js', import.meta.url)),
+    'endoScript',
+  );
+
+  const ns = await importBundle(bundle, {
+    inescapableGlobalProperties: {
+      inescapableGlobalValue: 42,
+    },
+  });
+  t.is(ns.default, 42);
```
warner, line 175: Not really part of this PR, but should these tests exercise the "inescapability" of `inescapableGlobalProperties` ? Or did we decide that only the lexical-scope additions needed to be inescapable?

warner (approved):
LGTM, modulo the two questions

packages/import-bundle/test/import-bundle.test.js
warner: looks like transforms being inescapable might be important too

kriskowal: The test fixture is a module that runs in a Compartment and creates a child Compartment without options. So, all of these are verifying inescapability at least one level down.
```js
export default new Compartment().evaluate('inescapableGlobalValue');
```

PR #2399, opened 2024-08-06
feat(immutable-arraybuffer): transferToImmutable ponyfill and shim

### LINKED ISSUE #1331: Support passable immutable ArrayBuffer as copy-data

erights:
Closes: #XXXX
Refs:  #1538 #1331 #2309 #2311 

## Description

Introduces the `@endo/immutable-arraybuffer` package, the ponyfill exports of `@endo/immutable-arraybuffer`, and the shim obtained by importing `@endo/immutable-arraybuffer/shim.js`.

Alternative to #2309 as suggested by @phoddie at https://github.com/endojs/endo/pull/2309#issuecomment-2155513240

We plan to fix #1331 in a stack of PRs starting with this one
- This PR implements a ponyfill and shim for an upcoming *Immutable ArrayBuffer* proposal, along the lines suggested by @phoddie at https://github.com/endojs/endo/pull/2309#issuecomment-2155513240 + the suggestions on an earlier state of #2311 . This is a pure JavaScript ponyfill/shim, leaving it to #2311 to bring it into Hardened JavaScript.
- #2311 imports the #2399 shim, treating the new objects it introduces as if they are new primordials, to be permitted and hardened.
- #1538 uses the Hardened JavaScript Immutable ArrayBuffers to define a new `Passable` type, `ByteArray`, corresponding to the [OCapN](https://ocapn.org/) `ByteArray`.
- Some future PR extending the various marshal formats to encode and decode the `ByteArray` objects.

See the README.md in this PR for more.

### Security Considerations

Better support for immutability generally helps security. The imperfections of the shim are a leaky abstraction in all the ways explained in the Caveats section of the README.md. For example, objects that are purely artifacts of the emulation, like the `immutableArrayBufferPrototype`, are easily discoverable, revealing the emulation's mechanisms. 

As a pure JavaScript polyfill/shim, this `@endo/immutable-arraybuffer` package does not harden the objects it exposes. Thus, by itself it does not provide much security -- like the initial state of JavaScript does not by itself provide much security. Rather, both provide securability, depending on Hardened JavaScript to harden early as needed to provide the security. See #2311

Once hardened early, the abstraction will still be leaky as above, but the immutability of the buffer contents are robustly enforced.

### Scaling Considerations

This ponyfill/shim is a zero-copy implementation, meaning that it does no more buffer copying than expected of a native implementation. 

This ponyfill/shim implements zero-copy by relying on the platform to provide one of two primitives: `structuredClone` or `ArrayBuffer.prototype.transfer`. Neither exist on Node <= 16. Without either, this ponyfill/shim will fail to initialize.

This PR sets the stage for writing an Immutable ArrayBuffer proposal, proposing it to tc39, and including it in our own documentation.

### Testing Considerations

Ideally, we should identify the subset of test262 `ArrayBuffer` tests that should be applicable to immutable ArrayBuffers, and duplicate them for that purpose.

### Upgrade Considerations

Nothing breaking.


erights:
OCapN meeting coming up on Tuesday. So if there's no problem with this, I'd like to make progress before then. Ping ;)

packages/immutable-arraybuffer/test/index.test.js
```diff
@@ -0,0 +1,92 @@
...
+    instanceOf: TypeError,
+  });
+  // Unfortunately, unlike the immutable ArrayBuffer to be proposed,
+  // calling a TypedArray constructor with the shim implementation of
+  // an immutable ArrayBuffer as argument treats it as an unrecognized object,
+  // rather than throwing an error.
+  t.is(iab.byteLength, 2);
+  const ta3 = new Uint8Array(iab);
+  t.is(ta3.byteLength, 0);
+});
```
kriskowal, line 92: This might be more legible as a test for each limitation.

packages/immutable-arraybuffer/index.js
```diff
@@ -0,0 +1,152 @@
...
+    // This constructor is deleted from the prototype below.
+    this.#buffer = arrayBufferClone(buffer);
+  }
+
+  get byteLength() {
+    return this.#buffer.byteLength;
+  }
+
+  get detached() {
+    this.#buffer; // shim brand check
```
kriskowal, line 83: TIL this throws for knock-offs.

packages/immutable-arraybuffer/README.md
```diff
@@ -0,0 +1,66 @@
...
+- The `immutableArrayBufferPrototype` intermediate prototype is an artifact of the emulation, but it is not encapsulated. It is trivially discoverable as the object that emulated immutable buffers directly inherit from.
+- The shim's emulated immutable buffers are not real `ArrayBuffer` exotic objects. If they were, the shim would not be able to protect them from being written. Even though they implement the full proposed `ArrayBuffer` API, they cannot be plug compatible -- they cannot be used as the backing stores of `DataView`s or `TypedArray`s. Perhaps follow-on shims might modify `DataView` and `TypedArray` to emulate that as well, but that is hard and beyond the ambition of this ponyfill + shim.
```
kriskowal, line 63: My reading tripped on “plug compatible”. Maybe “plug-compatible”.

kriskowal (approved):

packages/immutable-arraybuffer/index.js
```diff
@@ -0,0 +1,152 @@
...
+    return false;
+  }
+
+  get immutable() {
+    this.#buffer; // shim brand check
+    return true;
+  }
+
+  slice(begin = 0, end = undefined) {
```
gibson042, line 102: We should probably just defer to the built-ins.
```suggestion
  slice(begin = undefined, end = undefined) {
```

packages/immutable-arraybuffer/index.js
```diff
@@ -0,0 +1,152 @@
...
+  try {
+    return apply(isImmutableGetter, buffer, []);
+  } catch (err) {
+    if (err instanceof TypeError) {
+      // TODO: Should this test if `buffer` is a genuine `ArrayBuffer` exotic
+      // object, and throw if not?
+      return false;
+    }
+    throw err;
+  }
```
gibson042, line 151: Looking at existing getters such as [`resizable`](https://tc39.es/ecma262/multipage/structured-data.html#sec-get-arraybuffer.prototype.resizable), we should throw on invalid input.
```suggestion
  try {
    return apply(isImmutableGetter, buffer, []);
  } catch (err) {
    if (err instanceof TypeError) {
      // Enforce that `buffer` is a genuine ArrayBuffer before returning.
      arrayBufferSlice(arrayBuffer, 0, 0);
      return false;
    }
    throw err;
  }
```

gibson042 (approved):
The code looks reasonable, but I feel pretty strongly that we should not have a polyfill at all... liveslots will need to use a new endo to take advantage of this anyway, and it IMO should explicitly propagate `transferBufferToImmutable` and `isBufferImmutable` as globals into all descendant compartments rather than manipulating a built-in prototype and creating compatibility concerns.

packages/immutable-arraybuffer/index.js
erights: Done

packages/immutable-arraybuffer/index.js
erights: Done, thanks.

packages/immutable-arraybuffer/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "@endo/immutable-arraybuffer",
+  "version": "0.1.0",
+  "private": true,
+  "description": null,
+  "keywords": [],
+  "author": "Endo contributors",
+  "license": "Apache-2.0",
+  "homepage": "https://github.com/endojs/endo/tree/master/packages/skel#readme",
```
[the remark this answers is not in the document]
erights: Done

packages/immutable-arraybuffer/test/index.test.js
erights: Indeed! Done.

PR #2400, opened 2024-08-07
feat(ses): ArrayBuffer.prototype.transferToImmutable

erights:
Closes: #XXXX
Refs: #XXXX

## Description

Now that the Immutable ArrayBuffer proposal is at 2.7 and looks likely to proceed to 3, and then on to 4 and the official language, we are now allowing ourselves to build on the shim to emulate it already being in the language. In this PR ses imports the shim (not just the ponyfill) and then permits it's additions as harmless for ses users.

### Security Considerations

We carefully designed the proposal to maintain our high security standards. Likewise, in our shim, an Immutable ArrayBuffer encapsulates a normal mutable ArrayBuffer. But it is careful not to mutate it or to let it escape. The shim is low fidelity in terms of what it must practically omit --- TypedArrays and DavaView backed by Immutable ArrayBuffers. But these omissions do not raise security hazards, only compat hazards.

### Scaling Considerations

Because Immutable ArrayBuffers obviate the need for defensive copying and support safe zero-copy sharing, it will help scalability.

- [ ] When on XS, we should somehow arrange to use its high-speed high-fidelity highly tested Immutable ArrayBuffer implementation. We should also figure out our migration strategy as other engines of interest ship their own standards compliant implementations. All these steps will help scaling immensely. But it is unclear how to handle the transition.

Mostly, we can just point at the docs for Immutable ArrayBuffer itself, currently in the proposal and soon in the JS standard.

### Testing Considerations

The XS native Immutable ArrayBuffer implementation is also highly tested by the test262 tests in the PR they submitted, which will also test other native implementations.

- [ ] Our shim is not highly tested. We should somehow test it against the relevant subset of those submitted test262 tests.

The shim is low fidelity in terms of what it omits from the standard. The shim is also low fidelity in that it cannot hide an internal shim implementation artifact, the hidden prototype that all and only Immutable ArrayBuffers share. A correct implementation of the proposal will have no such extra prototype object.

### Upgrade Considerations

### DIFF
--- a/packages/ses/NEWS.md
+++ b/packages/ses/NEWS.md
@@ -2,6 +2,13 @@ User-visible changes in `ses`
 
 # Next release
 
+- Uses the `@endo/immutable-arraybuffer` shim to add `ArrayBuffer.p.immutable`, `ArrayBuffer.p.transferToImmutable`, and `ArrayBuffer.p.sliceToImmutable` to ses, in order to emulate the [Immutable ArrayBuffer proposal](https://github.com/tc39/proposal-immutable-arraybuffer). These make an ArrayBuffer-like object whose contents cannot be mutated. However, due to limitations of the shim
+  - Unlike `ArrayBuffer` and `SharedArrayBuffer` this shim's ArrayBuffer-like object cannot be transfered or cloned between JS threads.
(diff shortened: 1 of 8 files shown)


gibson042 (approved):
Officially approving this PR, but it's also implicitly covered by https://github.com/endojs/endo/pull/1538#pullrequestreview-2810863184 .

PR #2402, opened 2024-08-10
feat(ses): `errorTrapping` reports prepend `"SES_UNCAUGHT_EXCEPTION:"`

michaelfig:
Closes: #XXXX
Refs: #XXXX

## Description

Use `"SES_UNCAUGHT_EXCEPTION: ...` report output for parity with `SES_UNHANDLED_REJECTION: ...` report output.  Without this feature, `ses` would write the naked exception to stderr, which could lead to confusion when trying to understand _why_ the output was written.  This was doubly confusing if the exception was not an `Error`, rather just some other thrown value.

### Security Considerations

Improves security by giving more context for error output.

### Scaling Considerations

n/a

n/a

### Testing Considerations

Just unit tests.

Since stderr is not generally parsed by programs, I am not concerned with breaking compatibility of the rendering of the uncaught exceptions.  This change should improve the user experience.

### Upgrade Considerations

If the stderr of your SES program is processed by another program, this change will be visible.  You may need to change your error processor to be aware of `SES_UNCAUGHT_EXCEPTION: ...` output on stderr.

### DIFF
--- a/packages/ses/NEWS.md
+++ b/packages/ses/NEWS.md
@@ -1,5 +1,20 @@
 User-visible changes in `ses`:
 
+# Unreleased
+
+- If lockdown's `errorTrapping: 'report'` mode is selected (possibly via the
+  `'platform'`, or `'exit'` or `'abort'` modes), uncaught exceptions will be
+  written to standard error with the new `SES_UNCAUGHT_EXCEPTION: ` prefix.
+  This is intended to give valuable context to users of the system, especially
+  when an uncaught exception is not an `Error` object, and therefore its origin
+  may be hard to find in source code.
+
+  This is not likely to affect most systems built with SES, as stderr is
+  generally reserved for user-only messages.  If your SES system sends its
+  stderr to a program which parses it, you may need to adapt that program to be
+  tolerant of the `SES_UNCAUGHT_EXCEPTION: ` prefix.  Even for such programs, it
+  is unlikely they are that sensitive to stderr formatting.
+
 # v1.6.0 (2024-07-30)
 
 - *NOTICE*: This version introduces multiple features to converge upon a
--- a/packages/ses/test/console-error-trap.test.js
+++ b/packages/ses/test/console-error-trap.test.js
@@ -24,6 +24,10 @@ const exitAssertions = (
       'exit error code',
     );
     t.assert(
+      stderr.includes('SES_UNCAUGHT_EXCEPTION:'),
+      'stderr should have SES_UNCAUGHT_EXCEPTION:',
+    );
+    t.assert(
       stderr.includes('(Error#1)'),
       'stderr should have an error marker',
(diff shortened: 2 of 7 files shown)


packages/ses/NEWS.md
```diff
@@ -1,5 +1,20 @@
 User-visible changes in `ses`:
 
+# Unreleased
+
+- If lockdown's `errorTrapping: 'report'` mode is selected (possibly via the
+  `'platform'`, or `'exit'` or `'abort'` modes), uncaught exceptions will
```
erights, line 6:
```suggestion
  `'platform'`, or `'exit'` or `'abort'` modes), uncaught exceptions will be
```

packages/ses/test/console-error-trap.test.js
```diff
@@ -44,47 +48,51 @@ const exitAssertions = (
 };
 
 test('errors reveal their stacks', async t => {
-  t.plan(5);
+  t.plan(6);
```
erights, line 51: What other change does this number change reflect? I didn't spot it.

michaelfig: The change is by adding a single new test to the `exitAssertions` utility function that all of these top-level `test('...')` blocks run, thus incrementing the number of planned assertions in every one of the top-level tests.

erights: Resolving.

erights (changes requested):
Given the way this is named, it deserves a sibling to https://github.com/endojs/endo/tree/master/packages/ses/error-codes . Please add to this directory an explanation in `SES_UNCAUGHT_EXCEPTION.md`. While your at it, please also add an explanation in `SES_UNHANDLED_REJECTION.md`. Follow the rough style of its siblings.

If these errors report stack traces at their point of origin, please follow the precedent at, for example, https://github.com/endojs/endo/blob/bec206bc6519abd95a7f4b143b1743dba4795fe7/packages/ses/src/assert-sloppy-mode.js#L9 by inserting a

```js
  // See https://github.com/endojs/endo/blob/master/packages/ses/error-codes/SES_UNCAUGHT_EXCEPTION.md
```

right above that origin, so that someone inspecting a stack traceback is likely to encounter this link to the explanation. Likewise for `SES_UNHANDLED_REJECTION` if possible.

Otherwise, LGTM!

michaelfig:
> Given the way this is named, it deserves a sibling to https://github.com/endojs/endo/tree/master/packages/ses/error-codes . Please add to this directory an explanation in `SES_UNCAUGHT_EXCEPTION.md`. While your at it, please also add an explanation in `SES_UNHANDLED_REJECTION.md`. Follow the rough style of its siblings.

Done, and ready for rereview.

erights (approved):
LGTM, thanks!

PR #2404, opened 2024-08-11
support destructuring in harden-exports

turadg:
- **chore: ignore Aider files**
- **feat: destructured exports in harden-exports**

Closes: #XXXX
Refs: #XXXX

## Description

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations

### DIFF
--- a/.changeset/host-module-exits.md
+++ b/.changeset/host-module-exits.md
@@ -1,0 +1,24 @@
+---
+'ses': minor
+'@endo/compartment-mapper': minor
+'@endo/import-bundle': minor
+---
+
+Add support for host module exits in bundled compartments.
+
+`ses` exports a new `StrictModuleDescriptor` type that consists only of the
+`NamespaceModuleDescriptor` and `SourceModuleDescriptor` shapes mutually
+supported by SES and XS.
+
+`compartment-mapper` lets arbitrary module descriptors pass through
+`importHook` when no policy is in effect for that edge (the
+policy-enforcement runtime was previously limited to virtual module sources).
+It also implicitly treats any module specifier with a URL-scheme prefix
+(like `node:fs`) as an exit module when bundling, removing the need for an
+additional bundler flag in the common case.
+
+`import-bundle` threads the `importHook` option through to the underlying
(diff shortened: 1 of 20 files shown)


packages/eslint-plugin/lib/rules/harden-exports.js
```diff
@@ -46,23 +46,39 @@ module.exports = {
...
+                      continue;
+                    }
+                    if (prop.value.type === 'Identifier') {
+                      exportNames.push(prop.value.name);
+                    } else if (
+                      prop.value.type === 'AssignmentPattern' &&
+                      prop.value.left.type === 'Identifier'
+                    ) {
+                      exportNames.push(prop.value.left.name);
+                    }
```
gibson042, line 64: This is missing the recursive case in which `prop.value.type` is itself an ObjectPattern or ArrayPattern (e.g., `export const { wrapper: { propName } } = objWithWrapper;` and `export const [{ wrapper: { propName: exportName } }] = [objWithWrapper];`). That was true before as well, but if we're making changes anyway then we ought to fix the gap.

(and likewise for the elements of an ArrayPattern below)

packages/eslint-plugin/lib/rules/harden-exports.js
```diff
@@ -73,17 +89,33 @@ module.exports = {
...
+                    )) ||
+                  // @ts-expect-error XXX non-overlapping
+                  (statement.expression.arguments[0].type === 'ArrayPattern' &&
+                    // @ts-expect-error XXX non-overlapping
+                    statement.expression.arguments[0].elements.some(
+                      element =>
+                        element &&
+                        element.type === 'Identifier' &&
+                        element.name === exportName,
+                    ))),
```
gibson042, line 117: This whole block could really use some commentary/substructure/etc.

gibson042 (approved):
Sorry about the long delay... I like the better coverage, but would like to see a few tweaks. And I've retagged other good reviewers.

michaelfig:
I agree with @gibson042's comments.  I would like to see them addressed before I can approve or offer more suggestions.

PR #2406, opened 2024-08-12
types misc, Passable, fix EProxy

turadg:
_incidental_

## Description

I was exploring why `tsc` is so slow in agoric-sdk and read that `"strictFunctionTypes": true` might offer a significant speed bump. With that I ran into some problems with `EProxy` that its methods were `never`. That was because the `Callable` type expect `...args: unknown[]` and real values aren't. `unknown` shouldn't be used as a test for extension. The fix was to make it `...args: any[]`.

This does that, and a bunch of other changes that came up while exploring.

The change for Aider is completely unrelated but didn't seem worth its own PR.

### Security Considerations

none

### Scaling Considerations

none

none

### Testing Considerations

CI

Fixes types bugs

### Upgrade Considerations

Will be noted in conventional commits changelog

### DIFF
--- a/packages/eventual-send/src/handled-promise.js
+++ b/packages/eventual-send/src/handled-promise.js
@@ -115,12 +115,13 @@ export const makeHandledPromise = () => {
    * @type {Required<Handler<any>>}
    */
   let forwardingHandler;
+  /** @type {(...args: any[]) => Promise<unknown>} */
   let handle;
 
   /**
    * @param {string} handlerName
    * @param {Handler<any>} handler
-   * @param {string} operation
+   * @param {keyof Handler<any>} operation
    * @param {any} o
    * @param {any[]} opArgs
    * @param {Promise<unknown>} [returnedP]
@@ -147,7 +148,9 @@ export const makeHandledPromise = () => {
 
     if (matchSendOnly && typeof handler[actualOp] !== 'function') {
       // Substitute for sendonly with the corresponding non-sendonly operation.
-      actualOp = matchSendOnly[1];
+      actualOp = /** @type {'get' | 'applyMethod' | 'applyFunction'} */ (
+        matchSendOnly[1]
+      );
     }
 
     // Fast path: just call the actual operation.
@@ -207,6 +210,7 @@ export const makeHandledPromise = () => {
     let handledReject;
     let resolved = false;
     let resolvedTarget = null;
+    /** @type {Promise<R>} */
     let handledP;
     let continueForwarding = () => {};
     const assertNotYetForwarded = () => {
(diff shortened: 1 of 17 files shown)


packages/eventual-send/src/handled-promise.js
```diff
@@ -439,6 +443,7 @@ export const makeHandledPromise = () => {
       }
       // Assimilate the `resolvedPromise` as an actual frozen Promise, by
       // treating `resolvedPromise` as if it is a non-promise thenable.
+      // this is the same type as the Promise constructor
```
michaelfig, line 446: I'm pretty sure you don't mean "the Promise constructor", which is `new Promise(executor)` and only takes a single argument.  Would the following typing be better?
```suggestion
      // @type {Promise['then']}
```

packages/eventual-send/src/types.d.ts
```diff
@@ -17,7 +17,7 @@ export type * from './track-turns.js';
 //   Types exposed to modules.
 //
 
-export type Callable = (...args: unknown[]) => any;
+export type Callable = (...args: any[]) => any;
```
michaelfig, line 20: Thank you!  This kept flipping back and forth so much that I lost track of the fact that `unknown[]` broke the typing.

michaelfig (approved):
Looks really good!  I had a few comments I'd appreciate you addressing, but I'm optimistically approving so that you don't have to do another round trip through me to land.

packages/eventual-send/src/handled-promise.js
turadg: oops that was me attempting to get an LLM to annotate the type. And yeah that was not accurate anyway.

I think your suggestion will work. I'll try it and if not I'll just remove any attempt

packages/promise-kit/src/memo-race.js
```diff
@@ -111,15 +111,16 @@ const { race } = {
    * Unlike `Promise.race` it cleans up after itself so a non-resolved value doesn't hold onto
    * the result promise.
    *
-   * @template T
+   * @template {readonly unknown[] | []} T
    * @template {PromiseConstructor} [P=PromiseConstructor]
    * @this {P}
-   * @param {Iterable<T>} values An iterable of Promises.
```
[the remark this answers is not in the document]
turadg: To maintain the type of each position in the tuple. This is the way Promise.race is defined. As we discussed, it's probably a pragmatic decision even though any iterable is supported.

packages/pass-style/src/types.test-d.ts
```diff
@@ -29,3 +29,30 @@ expectType<'copyRecord'>(passStyleOf({}));
...
+expectPassable('str');
+expectPassable(undefined);
+// void is really `undefined`, and thus Passable
+expectPassable(fn());
+
+expectPassable({});
+expectPassable({ a: {} });
+// @ts-expect-error not passable
+expectPassable(fn);
+// FIXME promise for a non-Passable is not Passable
```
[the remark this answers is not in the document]
turadg: 👍 https://github.com/endojs/endo/pull/2421

PR #2408, opened 2024-08-12
fix(compartment-mapper): fix #2407 include symbol-named properties

### LINKED ISSUE #2407: Why we still fail to endow liveSlots
https://github.com/endojs/endo/blob/59b756eb896c72a75f4c0773f964903364bf7e71/packages/compartment-mapper/src/policy.js#L31

Uses `Object.keys` which only includes string-named properties.

My first thought is to change it to `Reflect.ownKeys`, but this includes non-enumerable properties as well. What did we decide was correct in general for these `global` endowments? Should we change to `Reflect.ownKeys` and then filter out non-enumerables? Or should we just use `Reflect.ownKeys`?

erights:

Closes: #XXXX
Refs: #XXXX

## Description

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations


packages/compartment-mapper/src/policy.js
```diff
@@ -28,7 +30,12 @@ export const ATTENUATORS_COMPARTMENT = '<ATTENUATORS>';
...
   if (!list) {
-    list = keys(from);
+    const descs = getOwnPropertyDescriptors(from);
+    list = ownKeys(from).filter(
+      key =>
+        // @ts-expect-error TypeScript still confused about a symbol as index
+        descs[key].enumerable,
+    );
   }
   for (let index = 0; index < list.length; index += 1) {
```
kriskowal, line 40: @naugtur Might be worth you time to investigate how a policy would name a symbol to be preserved.

packages/compartment-mapper/src/policy.js
```diff
@@ -28,7 +30,12 @@ export const ATTENUATORS_COMPARTMENT = '<ATTENUATORS>';
  */
 const selectiveCopy = (from, to, list) => {
   if (!list) {
-    list = keys(from);
+    const descs = getOwnPropertyDescriptors(from);
+    list = ownKeys(from).filter(
+      key =>
+        // @ts-expect-error TypeScript still confused about a symbol as index
+        descs[key].enumerable,
+    );
```
kriskowal, line 38: could use a calming comment:
“by default, copies all enumerable, own, string- and symbol-named properties”

kriskowal (approved):

packages/compartment-mapper/src/policy.js
erights: Done

PR #2414, opened 2024-08-16
feat(pass-style,marshal): ByteArray, a binary Passable

erights:
Closes: #1331 
Refs: #1538 https://github.com/ocapn/ocapn/issues/5#issuecomment-1492778252

## Description

Use the `@endo/immutable-arraybuffer` ponyfill directly, rather than the shim `@endo/immutable-arraybuffer/shim.js`, to provide objects that act like the immutable ArrayBuffers that would result from the `transferToImmutable` proposal.

Recognize these objects, when also frozen, as the new pass-style `byteArray`. Add a branch for each of the dispatches on pass-style that must now understand byte-arrays as well. As of this PR, many of those additional branches are stubbed out with "not yet implemented" errors as placeholders.

### Security Considerations

Does correctly enforce immutability. But by depending only on the ponyfill rather than the shim, this PR inherits the eval twin problem of the ponyfill. Each instantiation of `passStyleOf` will only recognize the byte-arrays from the instantiation of `@endo/pass-style` that made that byte-array. For environment that, for example, use the `passStyleOf` endowed to them by liveslots, liveslots will also need to endow them with the corresponding byte-array maker.

In theory, this is not much worse than the need to co-endow `passStyleOf` and `Proxy`, so that `passStyleOf` can reject apparent copy structures that are also proxies.

### Scaling Considerations

The underlying ponyfill implementation helps us handle bulk binary data in a largely no-copy or minimal-copy manner.

Will add a new pass-style, so everywhere we enumerate all the pass-styles, we'll need to add byte-array.

### Testing Considerations

- [ ] TODO tests. For each stubbed out case, there should be a `test.failing` test.

Old code that dispatched only on the pass-styles it knew about will not correctly handle passables that contain byte-arrays.

### Upgrade Considerations

- [ ] TODO BREAKING.md
- [ ] TODO NEWS.md

### DIFF
--- a/packages/marshal/src/rankOrder.js
+++ b/packages/marshal/src/rankOrder.js
@@ -55,6 +55,8 @@ export const trivialComparator = (left, right) =>
 const passStyleRanks = /** @type {PassStyleRanksRecord} */ (
   fromEntries(
     entries(passStylePrefixes)
+      // TODO Until byteArray prefix is chosen
+      .filter(([_style, prefixes]) => prefixes.length >= 1)
       // Sort entries by ascending prefix.
       .sort(([_leftStyle, leftPrefixes], [_rightStyle, rightPrefixes]) => {
         return trivialComparator(leftPrefixes, rightPrefixes);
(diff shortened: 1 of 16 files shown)


erights:
Converting to draft until more of the stubbed out cases are implemented, and until more of both are tested.

packages/marshal/src/rankOrder.js
```diff
@@ -209,6 +211,26 @@ export const makeComparatorKit = (compareRemotables = (_x, _y) => 0) => {
...
+          }
+          if (leftByte > rightByte) {
+            return 1;
+          }
+        }
+        // If all corresponding bytes are the same,
+        // then according to their lengths.
+        // Thus, if the data of ByteArray X is a prefix of
+        // the data of ByteArray Y, then X is smaller than Y.
+        return comparator(left.byteLength, right.byteLength);
```
erights, line 232: Should switch to shortlex. See https://en.wikipedia.org/wiki/Shortlex_order

That way, encodePassable could encode as prefix followed by the *unescaped* content.

packages/patterns/src/keys/compareKeys.js
```diff
@@ -147,6 +147,30 @@ export const compareKeys = (left, right) => {
...
+        if (leftByte > rightByte) {
+          return 1;
+        }
+      }
+      // If all corresponding bytes are the same,
+      // then according to their lengths.
+      // Thus, if the data of ByteArray X is a prefix of
+      // the data of ByteArray Y, then X is smaller than Y.
+      // @ts-expect-error narrowed
+      return compareRank(left.byteLength, right.byteLength);
```
erights, line 172: Should delegate the entire compare to rankOrder, which, btw, will switch to shortlex order.

erights:
Closing in favor of #1538 , since we no longer need to use the ponyfill and avoid the shim.

PR #2417, opened 2024-08-19
feat(ses): shim ArrayBuffer.prototype.transfer

erights:
Staged on #2419 

Closes: #XXXX
Refs: https://github.com/endojs/endo/pull/2414 #2418 #2419 

## Description

#2414  by itself does not work on Node 18 and Node 20 because
- those platforms do not have `Array.prototype.transfer`, so #2414 must use `structuredClone` instead
- `structuredClone` does exist on Node >= 18, so it should be on supported platforms (though see #2418 ). However, `structuredClone` itself is dangerous and so must not be added to the shared intrinsics. As a result, in #2414 , when `@endo/pass-style` is initialized in a created compartment, it fails to find either `Array.prototype.transfer` and `structuredClone

To solve this, @kriskowal suggested that we also shim `Array.prototype.transfer` if needed during `lockdown`, along with other repairs. We are avoiding similarly shimming `Array.prototype.transferToImmutable` because it is not yet standard. But `Array.prototype.transfer` is standard, and so `lockdown` can globally shim it before hardening the shared intrinsics.

This PR implements @kriskowal 's suggestion.

### Security Considerations

none
### Scaling Considerations

by itself, none

nothing signicant.
### Testing Considerations

See #2418 . Aside from that, none

On platforms with neither `Array.prototype.transfer` nor a global `structuredClone`, the ses-shim will *currently* not install an emulation of `Array.prototype.transfer`. However, once we verify that endo is not intended to support platforms without both, we may change lockdown to throw, failing to lock down.

### DIFF
--- a/packages/ses/src/lockdown.js
+++ b/packages/ses/src/lockdown.js
@@ -285,7 +285,7 @@ export const repairIntrinsics = (options = {}) => {
   addIntrinsics(tameMathObject(mathTaming));
   addIntrinsics(tameRegExpConstructor(regExpTaming));
   addIntrinsics(tameSymbolConstructor());
-  shimArrayBufferTransfer();
+  addIntrinsics(shimArrayBufferTransfer());
 
   addIntrinsics(getAnonymousIntrinsics());
 
--- a/packages/ses/src/lockdown.js
+++ b/packages/ses/src/lockdown.js
@@ -56,6 +56,7 @@ import { tameHarden } from './tame-harden.js';
 import { tameSymbolConstructor } from './tame-symbol-constructor.js';
 import { tameFauxDataProperties } from './tame-faux-data-properties.js';
 import { tameRegeneratorRuntime } from './tame-regenerator-runtime.js';
+import { shimArrayBufferTransfer } from './shim-arraybuffer-transfer.js';
 
 /** @import {LockdownOptions} from '../types.js' */
 
@@ -284,6 +285,7 @@ export const repairIntrinsics = (options = {}) => {
   addIntrinsics(tameMathObject(mathTaming));
   addIntrinsics(tameRegExpConstructor(regExpTaming));
   addIntrinsics(tameSymbolConstructor());
+  shimArrayBufferTransfer();
 
   addIntrinsics(getAnonymousIntrinsics());
 
--- a/packages/ses/src/shim-arraybuffer-transfer.js
+++ b/packages/ses/src/shim-arraybuffer-transfer.js
@@ -1,6 +1,9 @@
 import {
+  ArrayBuffer,
   arrayBufferPrototype,
   arrayBufferSlice,
+  arrayBufferGetByteLength,
+  Uint8Array,
   globalThis,
   TypeError,
   defineProperty,
@@ -12,7 +15,9 @@ export const shimArrayBufferTransfer = () => {
     // Assume already exists so does not need to be shimmed.
     // Such conditional shimming is ok in this case since ArrayBuffer.p.transfer
     // is already officially part of JS.
-    return;
+    //
+    // Empty object because this shim has nothing for `addIntrinsics` to add.
+    return {};
   }
   const clone = globalThis.structuredClone;
   if (typeof clone !== 'function') {
@@ -30,10 +35,9 @@ export const shimArrayBufferTransfer = () => {
      * @param {number} [newLength]
      */
     transfer(newLength = undefined) {
-      // Hopefully, a zero-length slice is cheap, but still enforces that
-      // `this` is a genuine `ArrayBuffer` exotic object.
-      arrayBufferSlice(this, 0, 0);
-      const oldLength = this.byteLength;
+      // Using this builtin getter also ensures that `this` is a genuine
+      // ArrayBuffer.
+      const oldLength = arrayBufferGetByteLength(this);
       if (newLength === undefined || newLength === oldLength) {
         return clone(this, { transfer: [this] });
       }
@@ -41,13 +45,28 @@ export const shimArrayBufferTransfer = () => {
         throw new TypeError(`transfer newLength if provided must be a number`);
       }
       if (newLength > oldLength) {
-        // TODO support this case somehow
-        throw new TypeError(
-          `Cannot yet emulate transfer to larger ArrayBuffer ${newLength}`,
-        );
+        // TODO Is there any way to do this bulk move or copy other than by
+        // manually iterating over each position?
+        const result = new ArrayBuffer(newLength);
+        // TODO Should this use DavaViews rather than TypedArrays?
+        const taOld = new Uint8Array(this);
+        const taNew = new Uint8Array(result);
+        let i = 0;
+        for (; i < oldLength; i += 1) {
+          taNew[i] = taOld[i];
+        }
+        for (; i < newLength; i += 1) {
+          taNew[i] = 0;
+        }
+        // Using clone only to detach, and only after the copy succeeds
+        clone(this, { transfer: [this] });
+        return result;
+      } else {
+        const result = arrayBufferSlice(this, 0, newLength);
+        // Using clone only to detach, and only after the slice succeeds
+        clone(this, { transfer: [this] });
+        return result;
       }
-      const tmp = clone(this, { transfer: [this] });
-      return arrayBufferSlice(tmp, 0, newLength);
     },
   };
 
@@ -58,4 +77,7 @@ export const shimArrayBufferTransfer = () => {
     enumerable: false,
     configurable: true,
   });
+
+  // Empty object because this shim has nothing for `addIntrinsics` to add.
+  return {};
 };
--- a/packages/ses/src/shim-arraybuffer-transfer.js
+++ b/packages/ses/src/shim-arraybuffer-transfer.js
@@ -4,6 +4,7 @@ import {
   arrayBufferSlice,
   arrayBufferGetByteLength,
   Uint8Array,
+  typedArraySet,
   globalThis,
   TypeError,
   defineProperty,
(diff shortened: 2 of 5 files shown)


erights:
See https://github.com/endojs/endo/issues/2418

The platform-compatability-test tests are not marked as required, so solving #2418 does not actually block this PR. But it would be nice to get that fixed first.

packages/ses/src/lockdown.js
```diff
@@ -284,6 +285,7 @@ export const repairIntrinsics = (options = {}) => {
   addIntrinsics(tameMathObject(mathTaming));
   addIntrinsics(tameRegExpConstructor(regExpTaming));
   addIntrinsics(tameSymbolConstructor());
+  shimArrayBufferTransfer();
```
kriskowal, line 288: Does this shim need to wait for `lockdown()`? Would it make sense to instead make it the behavior of importing `arraybuffer-transfer-shim.js` and do that in `index.js` instead?

(A current invariant among `-shim.js` modules exported at the top level is that they effect their behavior at initialization time, importing the behavior from a module without the `-shim` suffix.)

I ask because I could imagine ejecting this shim into a package, as I can imagine ejecting other layers.

erights: I looked, and moving it to `index.js` means that it could be bypassed:
- `package.json` exports `./lockdown-shim.js`
- ./lockdown-shim.js` just forwards to `./src/lockdown-shim.js`
- `./src/lockdown-shim.js` imports and calls `lockdown` without importing `../index.js`

So, for the current arrangement of things, no.

> I ask because I could imagine ejecting this shim into a package, as I can imagine ejecting other layers.

Once we unbundle `lockdown` itself into lots of vetted shims between `repairIntrinsics` and `hardenIntrinsics`, we should revisit this one as well. It will probably participate in that great unbundling.

Resolving.

mhofman, line 288: Wondering if it would make sense for "consistency" to have `shimArrayBufferTransfer()` return an empty "newIntrinsics" object, and have this line call `addIntrinsics()` like the other taming calls. While not needed for this specific case, conceptually a vetted shim may need to define new intrinsics, and I expect that future work to support vetted shims will require some API to do this registration.

packages/ses/src/shim-arraybuffer-transfer.js
```diff
@@ -0,0 +1,61 @@
...
+   */
+  const methods = {
+    /**
+     * @param {number} [newLength]
+     */
+    transfer(newLength = undefined) {
+      // Hopefully, a zero-length slice is cheap, but still enforces that
+      // `this` is a genuine `ArrayBuffer` exotic object.
+      arrayBufferSlice(this, 0, 0);
+      const oldLength = this.byteLength;
```
mhofman, line 36: Why not use the [getter in `ArrayBuffer.prototype.byteLength`](https://tc39.es/ecma262/#sec-get-arraybuffer.prototype.bytelength) to do both the brand check and get the value here if we don't trust the `this` object?

```suggestion
      const oldLength = arrayBufferGetByteLength(this);
```

packages/ses/src/shim-arraybuffer-transfer.js
```diff
@@ -0,0 +1,61 @@
...
+        return clone(this, { transfer: [this] });
+      }
+      if (typeof newLength !== 'number') {
+        throw new TypeError(`transfer newLength if provided must be a number`);
+      }
+      if (newLength > oldLength) {
+        // TODO support this case somehow
+        throw new TypeError(
+          `Cannot yet emulate transfer to larger ArrayBuffer ${newLength}`,
+        );
```
mhofman, line 47: Why can't we construct a new buffer of a larger size and copy the data in?

packages/ses/src/shim-arraybuffer-transfer.js
```diff
@@ -0,0 +1,61 @@
...
+      }
+      const tmp = clone(this, { transfer: [this] });
+      return arrayBufferSlice(tmp, 0, newLength);
```
mhofman, line 50: Technically we should detach the source **after** the target has been allocated. If the allocation fails, we would find ourselves in a case where the source is no longer valid, yet don't have a new buffer with the data.

mhofman (changes requested):
Please consider removing the unnecessary `slice` call as brand check, and improve support for when the new length doesn't match.

packages/ses/src/lockdown.js
erights: Done.

For the same "consistency", should I also rename `shimArrayBufferTransfer` to `tameArrayBufferTransfer`, and rename `shim-arraybuffer-transfer*` to `tame-arraybuffer-transfer*`?

My mild preference is not to.

packages/ses/src/shim-arraybuffer-transfer.js
erights: Good idea! Done.

packages/ses/src/shim-arraybuffer-transfer.js
erights: I hoped there would be a way to bulk move or copy the contents, rather than copy each byte individually. But no matter. I resorted to that. Done.

packages/ses/src/shim-arraybuffer-transfer.js
erights: Good point. Done. Also done for the growth case you suggested I do above.

erights:
PTAL

packages/ses/src/lockdown.js
mhofman: No, it's still a shim, it's installing a new thing instead of modifying an existing thing,

erights: Good. In that case, done.

erights:
> Please consider removing the unnecessary `slice` call as brand check, and improve support for when the new length doesn't match.

Done

packages/ses/src/shim-arraybuffer-transfer.js
```diff
@@ -0,0 +1,83 @@
...
+        // TODO Should this use DavaViews rather than TypedArrays?
+        const taOld = new Uint8Array(this);
+        const taNew = new Uint8Array(result);
+        let i = 0;
+        for (; i < oldLength; i += 1) {
+          taNew[i] = taOld[i];
+        }
+        for (; i < newLength; i += 1) {
+          taNew[i] = 0;
+        }
```
mhofman, line 60:
```suggestion
        taNew.set(taOld)
```

packages/ses/src/shim-arraybuffer-transfer.js
```diff
@@ -0,0 +1,83 @@
...
+      const oldLength = arrayBufferGetByteLength(this);
+      if (newLength === undefined || newLength === oldLength) {
+        return clone(this, { transfer: [this] });
+      }
+      if (typeof newLength !== 'number') {
+        throw new TypeError(`transfer newLength if provided must be a number`);
+      }
+      if (newLength > oldLength) {
+        // TODO Is there any way to do this bulk move or copy other than by
+        // manually iterating over each position?
```
mhofman, line 49: Yes, [`TypedArray.set(otherTypedArray)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/set#typedarray)

packages/ses/test/shim-arraybuffer-transfer.test.js
```diff
@@ -0,0 +1,61 @@
...
+// correctly enough on platforms like Node 18 or Node 20 that don't yet have
+// it natively, and so are testing the shim on those. On platforms where
+// Array.prototype.transfer is present, like Node 22,
+// we also run the same tests.Thus,
+// this test only tests the intersection behavior of the standard and
+// the shim. The shim does not yet support a `newLength` argument
+// larger than the original.
+//
+// TODO once the shim supports transfering to a larger length, we must
+// test that as well.
```
mhofman, line 16: I think this needs to be updated now :)

mhofman (changes requested):
Thanks for addressing the previous feedback. There is indeed a better way to copy an array buffer,

packages/ses/src/shim-arraybuffer-transfer.js
erights: Glad to hear it! Done.

packages/ses/src/shim-arraybuffer-transfer.js
erights: Glad to hear it! Comment removed. Done.

packages/ses/test/shim-arraybuffer-transfer.test.js
erights: Was already done, but I had not noticed the stale comment. Thanks. Fixed.

mhofman (approved):

packages/ses/src/shim-arraybuffer-transfer.js
```diff
@@ -0,0 +1,75 @@
...
+     */
+    transfer(newLength = undefined) {
+      // Using this builtin getter also ensures that `this` is a genuine
+      // ArrayBuffer.
+      const oldLength = arrayBufferGetByteLength(this);
+      if (newLength === undefined || newLength === oldLength) {
+        return clone(this, { transfer: [this] });
+      }
+      if (typeof newLength !== 'number') {
+        throw new TypeError(`transfer newLength if provided must be a number`);
```
mhofman, line 46:
```suggestion
        throw TypeError(`transfer newLength if provided must be a number`);
```

packages/ses/src/shim-arraybuffer-transfer.js
```diff
@@ -0,0 +1,75 @@
...
+    //
+    // Empty object because this shim has nothing for `addIntrinsics` to add.
+    return {};
+  }
+  const clone = globalThis.structuredClone;
+  if (typeof clone !== 'function') {
+    // Indeed, Node <= 16 has neither.
+    throw TypeError(
+      `Can only shim missing ArrayBuffer.prototype.transfer on a platform with "structuredClone"`,
+    );
```
mhofman, line 28: I am wondering if it would make sense, instead of failing, to not install a shim if `structuredClone` is not available?

packages/ses/src/shim-arraybuffer-transfer.js
erights: Done

packages/ses/src/shim-arraybuffer-transfer.js
erights: Reluctantly, but it is the right thing to do at this time, so done.

Added a TODO explaining misgivings and the open questions to be revisited.

Issue #2418, opened 2024-08-19
The ses-shim tests on Node 12 despite our min support being Node 18


erights:
https://github.com/endojs/endo/pull/2417 as of this writing fails both "platform-compatibility-test"s, allegedly for Node 18 and Node 20. However, these tests also depend on Node 12, which https://github.com/endojs/endo/pull/2417 does not support.

The culprit seems to be
https://github.com/endojs/endo/blob/46ca503ccf58cf2a418b97813bc1f5d864f6d0f7/.github/workflows/ci.yml#L273-L276
where the Node v12 dependency seems to be intentional.

erights:
The Node 12 dependency comes from @kriskowal 's https://github.com/endojs/endo/pull/1308 , which upgraded it from Node 10 to Node 12.

@kriskowal can we just upgrade this again to 18 and call it a day? Since you authored #1308 , can I assign to you?

erights:
Experiment at #2419

PR #2419, opened 2024-08-19
fix: Delete obsolete platform-compatibility-test

erights:
Closes: #2418 
Refs: #2417 #1308 

## Description

Adapted from https://github.com/endojs/endo/pull/2419#pullrequestreview-2252979514 below

The platform compatibility test specifically validates that SES works on Node.js 12 and can be deleted since it has vanished into history. Node.js 12 required special consideration because of its experimental ESM support. Delete the `test:platform-compatibility` script in the ses `package.json`, as well as the `test/package` fixture in ci.yml

Immediate motivation explained in #2418 , #2417 broken the platform-compatibility-test tests because it depends on the platform providing either `Array.prototype.transfer` or `structuredClone`. Node supports `transfer` starting with Node 22, but supports `structuredClone` starting in Node 18. That should be fine, since that is our declared support floor. This PR changes our one known remaining dependence on Node 12.

### Security Considerations

none
### Scaling Considerations

none

none beyond the need to explain our platform requirements, which this PR does not change.
### Testing Considerations

The point. This PR only affects tests, not production code.

Reviewers, should this PR be labeled `test:` instead of `fix:`?

After this PR, we will no longer notice further breakage on Node < 18. That fits with our declared support floor.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -231,57 +231,6 @@ jobs:
       - name: Run yarn test262
         run: exit 0 # TODO remove test262 from required tests for CI
 
-  platform-compatibility-test:
-    name: platform-compatibility-test
-
-    # begin macro
-
-    runs-on: ${{ matrix.platform }}
-    strategy:
-      fail-fast: false
-      matrix:
-        node-version: [18.x, 20.x]
-        platform: [ubuntu-latest]
-
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v3
-
-      # without this, setup-node errors on mismatched yarn versions
-      - run: corepack enable
-
-      - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v3
-        with:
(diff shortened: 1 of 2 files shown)


kriskowal (changes requested):
The platform compatibility test specifically validates that SES works on Node.js 12 and can be deleted since it has vanished into history. Node.js 12 required special consideration because of its experimental ESM support. I would not mind following that down to the roots: The `test:platform-compatibility` script in the ses `package.json` can be deleted, as well as the `test/package` fixture.

erights:
> The platform compatibility test specifically validates that SES works on Node.js 12 and can be deleted since it has vanished into history. Node.js 12 required special consideration because of its experimental ESM support. I would not mind following that down to the roots: The `test:platform-compatibility` script in the ses `package.json` can be deleted, as well as the `test/package` fixture.

Done. PTAL

kriskowal (approved):

PR #2421, opened 2024-08-20
Promise for non-Passable is not Passable

turadg:
Refs: #2406

## Description

@mhofman [suggested a fix for the FIXME](https://github.com/endojs/endo/pull/2406#discussion_r1723844078) in #2406. As expected it trip on recursion:
```
packages/pass-style/src/deeplyFulfilled.js:74:32 - error TS2589: Type instantiation is excessively deep and possibly infinite.

74 export const deeplyFulfilled = async val => {
                                  ~~~~~~~~~~~~~~
```

Leaving this PR in draft until that's solved.  @mhofman @michaelfig feel free to push commits.

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations

### DIFF
--- a/packages/pass-style/src/deeplyFulfilled.js
+++ b/packages/pass-style/src/deeplyFulfilled.js
@@ -30,23 +30,17 @@ const { fromEntries } = Object;
  */
 
 /**
- * Currently copied from @agoric/internal utils.js.
- * TODO Should migrate here and then, if needed, reexported there.
- *
- * @template {{}} T
+ * @template {Record<string, any>} T
  * @typedef {{
  *   [K in keyof T]: T[K] extends Callable ? T[K] : DeeplyAwaited<T[K]>;
  * }} DeeplyAwaitedObject
  */
 
 /**
- * Currently copied from @agoric/internal utils.js.
- * TODO Should migrate here and then, if needed, reexported there.
- *
  * @template T
  * @typedef {T extends PromiseLike<any>
  *     ? Awaited<T>
- *     : T extends {}
+ *     : T extends Record<string, any>
  *       ? Simplify<DeeplyAwaitedObject<T>>
  *       : Awaited<T>} DeeplyAwaited
  */
@@ -73,7 +67,7 @@ const { fromEntries } = Object;
  * is for the higher "@endo/patterns" level of abstraction to determine,
  * because it defines the `Key` notion in question.
  *
- * @template {Passable} [T=Passable]
+ * @template {Passable} T
  * @param {T} val
  * @returns {Promise<DeeplyAwaited<T>>}
  */
--- a/packages/pass-style/src/deeplyFulfilled.js
+++ b/packages/pass-style/src/deeplyFulfilled.js
@@ -6,8 +6,9 @@ import { passStyleOf } from './passStyleOf.js';
 import { makeTagged } from './makeTagged.js';
 
 /**
- * @import {Passable, Primitive, CopyRecord, CopyArray, CopyTagged, RemotableObject} from '@endo/pass-style'
+ * @import {Passable, CopyRecord, CopyArray, CopyTagged, RemotableObject} from '@endo/pass-style'
  */
+/** @import {Callable, RemotableBrand} from '@endo/eventual-send' */
 
 const { ownKeys } = Reflect;
 const { fromEntries } = Object;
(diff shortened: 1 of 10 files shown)


erights:
> Promise for non-Passable is not Passable

When I first read this, I thought it was about promises and whether they were passable. Took me a bit to understand that this is actually about the `Promise` type and the `Passable` type. Would be good to rephrase to make clear this is not about what might actually happen at runtime.

The fact that a passable unresolved promise might later fulfill to a non-passable value is a significant hazard in our system we need to remain aware of, whatever the type system might claim to the contrary.

mhofman:
> The fact that a passable unresolved promise might later fulfill to a non-passable value is a significant hazard in our system we need to remain aware of, whatever the type system might claim to the contrary.

This is a good point, maybe this point to a mistake in the approach. The `Promise` (if an expected `PassableCap`) is `Passable`, but the result value may be `never` if it's itself non-`Passable`

mhofman:
I think I found a way to break the recursion issues, albeit with a theoretically breaking change in the type definition of Passable, but in practice I think it's backwards compatible.

I also took the opportunity to further narrow the type of `PassableCap`: when a promise, it's only a promise for a `RemotableObject`. I believe it matches actual explicit usages of the type.

I have not tested this change applied to agoric-sdk yet.

turadg:
> tested this change applied to agoric-sdk

We can use https://github.com/Agoric/agoric-sdk/pull/9385 . I've pointed it to this branch.

packages/pass-style/src/deeplyFulfilled.js
```diff
@@ -73,7 +58,7 @@ const { fromEntries } = Object;
  * is for the higher "@endo/patterns" level of abstraction to determine,
  * because it defines the `Key` notion in question.
  *
- * @template {Passable} [T=Passable]
+ * @template {Passable} T
```
erights, line 61: Is there a reason to remove the default?

In general when I write a type-constrained type parameter, I also provide that type constraint as a default binding of the type parameter. Should I stop doing so?

packages/pass-style/src/deeplyFulfilled.js
```diff
@@ -6,8 +6,9 @@ import { passStyleOf } from './passStyleOf.js';
 import { makeTagged } from './makeTagged.js';
 
 /**
- * @import {Passable, Primitive, CopyRecord, CopyArray, CopyTagged, RemotableObject} from '@endo/pass-style'
+ * @import {Passable, CopyRecord, CopyArray, CopyTagged, RemotableObject} from '@endo/pass-style'
  */
+/** @import {Callable, RemotableBrand} from '@endo/eventual-send' */
```
erights, line 11: Nice. I'lll just mention that I'm a bit surprised that `RemotableBrand` is defined by @endo/eventual-send but `RemotableObject` is defined by @endo/pass-style. No change suggested.

packages/pass-style/src/types.d.ts
```diff
@@ -80,20 +80,54 @@ export type PassByRef =
...
-  | CopyArrayI<PC, E>
-  | CopyRecordI<PC, E>
-  | CopyTaggedI<PC, E>;
-interface CopyArrayI<PC extends PassableCap, E extends Error>
-  extends CopyArray<Passable<PC, E>> {}
-interface CopyRecordI<PC extends PassableCap, E extends Error>
-  extends CopyRecord<Passable<PC, E>> {}
-interface CopyTaggedI<PC extends PassableCap, E extends Error>
-  extends CopyTagged<string, Passable<PC, E>> {}
+  AllowPromise extends boolean = any,
```
erights, line 85: What does it mean to both type-constrain the type variable `AllowPromise` to `boolean` but give it a default of `any`?

erights: Curious why you didn't adopt the convention of the previous two type variables and write

```suggestion
  P extends Promise = Promise,
```
?

erights: Or, given the other changes

```suggestion
  P extends PromiseLike = PromiseLike,
```

packages/pass-style/src/types.d.ts
```diff
@@ -80,20 +80,54 @@ export type PassByRef =
...
+  AllowTopLevelPromise extends boolean = AllowPromise,
```
erights, line 86: Where do you make use of this fourth parameter?

packages/pass-style/src/deeplyFulfilled.js
mhofman: For generic functions we want to rely on inference. I don't remember exactly why I dropped it, but I think a default doesn't have any advantages here, especially one where the default value is the constraint itself.

For type definitions it's a different story as it allows you to reference the bare type without type parameters.

packages/pass-style/src/deeplyFulfilled.js
mhofman: Yeah they're very different, somewhat orthogonal things. I've been discussing with @michaelfig, and from eventual-send's pov, the target or arguments don't require any specific pass-style. `RemotableBrand` helps the typing to capture the shape of remote methods (and the shape of local data) without exposing any actual property on the type itself (its storing them in the private fields of a "stamped" class).

`RemotableObject` is the pass-style brand, which includes the 2 symbol properties (`PASS_STYLE` and `Symbol.toStringTag`), denoting a remotable in passables.

packages/pass-style/src/types.d.ts
mhofman: It's a bit of a hack in the type system. I need to conditionally include a `PromiseLike` in the union, or a `never`. But the `PromiseLike` needs to be parametrized with a specific constraint. As such this type parameter cannot be provided as a type (or inferred), but instead is just a flag to say "please allow promises in the union".

The `any` default is also a hack to make TS bail out when trying to type match. Without it we wouldn't be able to assign a `Passable` without promises (`AllowPromise=false`) to ones with it (or really places where we don't care). I don't fully understand why, this is where I ended at experimentally.

packages/pass-style/src/types.d.ts
mhofman: Internally to avoid including `PromiseLike` in the union of acceptable types inside the `PromiseLike` when recursing the type.

PR #2422, opened 2024-08-20
feat(compartment-mapper): Host module exits

kriskowal:
## Description

This change closes the remaining gap between Endo as a Zip archive bundler and support for “exits” to host modules from those bundles at import-time.

This begins with a test to demonstrate the latent support for conditional exits to host modules through `importLocation`. The test fixture has a library that exports the host implementation of itself if the bundler specifies the exit (`endo:lib`) as a condition. This causes the module to be omitted from bundles and relies on the importer to provide the implementation. There remained a gap for a round-trip through a bundle.

The subsequent changes fix a bottleneck for exit modules in the compartment-mapper. The LavaMoat policy-enforcement runtime is limited to virtual module sources, which constrained support for other kinds of module descriptor. This change opens that up so arbitrary module descriptors pass-through the attenuating adapter if no policy is in effect for that edge. We can return to explore attenuation of other kinds of module-descriptor.

At this point, all exits have to be explicitly marked with an `importHook` that returns a module descriptor for the named exit module specifier. We then add a feature to the bundler that implicitly recognizes any module specifier that starts with a URI-scheme prefix is an exit, for convenience on the bundler side. This will obviate the need for an additional command-line flag in `bundle-source` in the common case.

Then, we trivially thread the `importHook` through `importBundle` options.

### Security Considerations

Host provided modules must be hardened and pure, to avoid being useful as a side-channel or mitm attack surface between guests.

### Scaling Considerations

This should allow the creation of smaller bundles.

Any module that implements this feature should document the condition that enables it for bundling and importing.

### Testing Considerations

Just a test.

None.

### Upgrade Considerations

None.

### DIFF
--- a/packages/compartment-mapper/src/import-archive-lite.js
+++ b/packages/compartment-mapper/src/import-archive-lite.js
@@ -136,15 +136,12 @@ const makeArchiveImportHookMaker = (
             // note it's not being marked as exit in sources
             // it could get marked and the second pass, when the archive is being executed, would have the data
             // to enforce which exits can be dynamically imported
-            return {
-              record: await attenuateModuleHook(
-                moduleSpecifier,
-                record,
-                compartmentDescriptor.policy,
-                attenuators,
-              ),
-              specifier: moduleSpecifier,
-            };
+            return attenuateModuleHook(
+              moduleSpecifier,
+              record,
+              compartmentDescriptor.policy,
+              attenuators,
+            );
           } else {
             // if exitModuleImportHook is allowed, the mechanism to defer
             // errors in archive creation is never used. We don't want to
--- a/packages/compartment-mapper/src/link.js
+++ b/packages/compartment-mapper/src/link.js
@@ -70,6 +70,13 @@ const has = (object, key) => apply(hasOwnProperty, object, [key]);
 // q, as in quote, for strings in error messages.
 const { quote: q } = assert;
 
+// See section 3.1 of RFC 3986 URI Scheme Generic Syntax
+// https://www.rfc-editor.org/rfc/rfc3986#section-3.1
+// Scheme names are case-insensitive per the cited section, so the regex
+// accepts both cases on the prefix. Bundlers that produce specifiers like
+// `HTTP:` or `File:` are unusual but valid per the spec.
+const urlish = /^[a-zA-Z][a-zA-Z0-9+\-.]*:/;
+
 /**
  * For a full, absolute module specifier like "dependency",
  * produce the module specifier in the dependency, like ".".
@@ -105,6 +112,7 @@ const trimModuleSpecifierPrefix = (moduleSpecifier, prefix) => {
  * @param {FileUrlString} compartmentName
  * @param {Record<string, FileModuleConfiguration|CompartmentModuleConfiguration>} moduleDescriptors
  * @param {Record<string, ScopeDescriptor<FileUrlString>>} scopeDescriptors
+ * @param {boolean} archiveOnly
  * @returns {ModuleMapHook | undefined}
  */
 const makeModuleMapHook = (
@@ -113,6 +121,7 @@ const makeModuleMapHook = (
   compartmentName,
   moduleDescriptors,
   scopeDescriptors,
+  archiveOnly,
 ) => {
   // Build pattern matcher once per compartment if patterns exist.
   const { patterns } = /** @type {Partial<PackageCompartmentDescriptor>} */ (
@@ -127,6 +136,25 @@ const makeModuleMapHook = (
    */
   const moduleMapHook = moduleSpecifier => {
     compartmentDescriptor.retained = true;
+
+    if (archiveOnly && urlish.test(moduleSpecifier)) {
+      // When creating an archive of an application that imports a platform
+      // module like node:fs, we implicitly expect these to be provided by the
+      // host's importHook on the target platform. Freeze the synthesized
+      // descriptor so the cross-boundary return cannot be tampered with by a
+      // downstream caller.
+      return freeze({
+        source: freeze({
+          imports: freeze([]),
+          exports: freeze([]),
+          execute() {
+            throw new Error(
+              'Cannot import an application loaded strictly for analysis',
+            );
+          },
+        }),
+      });
+    }
 
     const moduleDescriptor = moduleDescriptors[moduleSpecifier];
     if (moduleDescriptor !== undefined) {
@@ -438,6 +466,7 @@ export const link = (
       compartmentName,
       modules,
       scopes,
+      archiveOnly,
     );
 
     const compartment = new Compartment({
--- a/packages/compartment-mapper/src/policy.js
+++ b/packages/compartment-mapper/src/policy.js
@@ -19,7 +19,13 @@
  *   ModuleConfiguration,
  *   CanonicalName,
  * } from './types.js'
- * @import {ThirdPartyStaticModuleInterface} from 'ses'
+ * @import {
+ *   ModuleDescriptor,
+ *   SourceModuleDescriptor,
+ *   StrictModuleDescriptor,
+ *   ThirdPartyStaticModuleInterface,
+ *   VirtualModuleSource,
+ * } from 'ses'
  */
 
 import {
@@ -483,17 +489,18 @@ export const enforcePackagePolicyByCanonicalName = (
 };
 
 /**
- * Attenuates a module
+ * Attenuates a virtual module source.
+ *
  * @param {object} options
  * @param {DeferredAttenuatorsProvider} options.attenuators
  * @param {AttenuationDefinition} options.attenuationDefinition
- * @param {ThirdPartyStaticModuleInterface} options.originalModuleRecord
- * @returns {Promise<ThirdPartyStaticModuleInterface>}
+ * @param {VirtualModuleSource} options.moduleSource
+ * @returns {Promise<VirtualModuleSource>}
  */
-async function attenuateModule({
+async function attenuateVirtualModuleSource({
   attenuators,
   attenuationDefinition,
-  originalModuleRecord,
+  moduleSource,
 }) {
   const attenuate = await importAttenuatorForDefinition(
     attenuationDefinition,
@@ -504,15 +511,17 @@ async function attenuateModule({
   // An async attenuator maker could be introduced here to return a synchronous attenuator.
   // For async attenuators see PR https://github.com/endojs/endo/pull/1535
 
+  // Freeze the returned virtual module source so the attenuator boundary does
+  // not hand callers a mutable wrapper they could tamper with after the fact.
   return freeze(
-    /** @type {ThirdPartyStaticModuleInterface} */ ({
-      imports: originalModuleRecord.imports,
+    /** @type {VirtualModuleSource} */ ({
+      imports: moduleSource.imports,
       // It seems ok to declare the exports but then let the attenuator trim the values.
       // Seems ok for attenuation to leave them undefined - accessing them is malicious behavior.
-      exports: originalModuleRecord.exports,
+      exports: moduleSource.exports,
       execute: (moduleExports, compartment, resolvedImports) => {
         const ns = {};
-        originalModuleRecord.execute(ns, compartment, resolvedImports);
+        moduleSource.execute(ns, compartment, resolvedImports);
         const attenuated = attenuate(ns);
         moduleExports.default = attenuated;
         assign(moduleExports, attenuated);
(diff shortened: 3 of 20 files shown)


packages/compartment-mapper/test/fixtures-conditional-host-exports/node_modules/app/index.js
```diff
@@ -0,0 +1 @@
+export { feature } from 'lib';
```
dckc, line 1: why is `app` under `node_modules`?

packages/compartment-mapper/test/conditional-host-exports.test.js
```diff
@@ -0,0 +1,40 @@
...
+});
+
+test('conditional host exports should exit to import hook', async t => {
+  const readPowers = makeReadPowers({
+    url,
+    fs,
+    crypto,
+  });
+  const { namespace } = await importLocation(readPowers, fixture, {
+    conditions: new Set(['endo:lib']),
```
dckc, line 33: > This causes the module to be omitted from bundles ...

I'm trying to make such a bundle but losing:

```console
~/projects/endo/packages/compartment-mapper$ ../../node_modules/.bin/bundle-source -C endo:lib  ./test/fixtures-conditional-host-exports/node_modules/app/index.js 
(TypeError#1)
TypeError#1: Failed to load module "./index.js" in package "file:///home/connolly/projects/endo/packages/compartment-mapper/test/fixtures-conditional-host-exports/node_modules/app/" (1 underlying failures: Cannot find external module "endo:lib" in package file:///home/connolly/projects/endo/packages/compartment-mapper/test/fixtures-conditional-host-exports/node_modules/lib/

  at throwAggregateError (packages/ses/src/module-load.js:546:11)
  at load (packages/ses/src/module-load.js:594:3)
  at async digestFromMap (packages/compartment-mapper/src/archive-lite.js:364:3)
  at async makeAndHashArchiveFromMap (packages/compartment-mapper/src/archive-lite.js:415:52)
  at async bundleZipBase64 (packages/bundle-source/src/zip-base64.js:83:29)
  at async main (packages/bundle-source/src/main.js:107:20)
```

dckc (commented):
I don't think I get it yet.

packages/compartment-mapper/test/fixtures-conditional-host-exports/node_modules/app/index.js
kriskowal: Ah, this would mostly work at the directory containing `node_modules`, but the compartment mapper issues a warning if the `package.json` `name` does not match the parent directory name.

packages/compartment-mapper/test/conditional-host-exports.test.js
kriskowal: I’ve got some more wrangling to do to make this work through the `makeBundle` to `parseBundle`, and a tiny amount of work to thread through from `bundleSource` to `importBundle` after that. Retreating to draft.

kriskowal: This should work now. `bundle-source -C endo:lib -T` produced a bundle with a transparent exit import for me.

dckc: confirmed:

```console
$ cd packages/compartment-mapper
$ mkdir -p exit
$ mkdir -p no-exit

$ ../../node_modules/.bin/bundle-source  -T ./test/fixtures-conditional-host-exports/node_modules/app/index.js | jq -r .endoZipBase64| base64 -d | bsdtar -C no-exit -xvf-
x compartment-map.json
x app/index.js
x lib/index.js

$ ../../node_modules/.bin/bundle-source -C endo:lib -T ./test/fixtures-conditional-host-exports/node_modules/app/index.js | jq -r .endoZipBase64| base64 -d | bsdtar -C exit -xvf-
x compartment-map.json
x app/index.js
x lib/endo.js

$ cat exit/lib/endo.js 
export { feature } from 'endo:lib';

$ cat no-exit/lib/index.js 
export const feature = 'ponyfill';
```

packages/compartment-mapper/NEWS.md
```diff
@@ -1,5 +1,20 @@
 User-visible changes to `@endo/compartment-mapper`:
 
+# Next release
+
+- Archives will now automatically treat any module specifier that appears
+  to start with a URL scheme (like `node:fs`) and these "exits" will no
+  longer need to be marked in the `"modules"` option.
```
erights, line 7: Grammar? I'm having a hard time parsing this sentence.

packages/compartment-mapper/test/conditional-host-exports.test.js
dckc: ## POC: Eliding @endo/patterns etc. from a contract bundle

ooh! works on `orca.contract.js` from [dapp-orchestration-basics](https://github.com/Agoric/dapp-orchestration-basics) too!

Contract bundle in its original form is around 2.1M:

```console
$ cd ~/projects/dapp-orchestration-basics/contract

$ ~/projects/endo/node_modules/.bin/bundle-source -T src/orca.contract.js | jq -r .endoZipBase64| base64 -d >no-exit.zip && ls -sh no-exit.zip 
2.1M no-exit.zip
```

Then I tweaked `@endo/patterns`, `@endo/far`, and `@endo/eventual-send` something like:

```
$ git diff .
diff --git a/packages/patterns/package.json b/packages/patterns/package.json
index b2b9866ea..1a149a52f 100644
--- a/packages/patterns/package.json
+++ b/packages/patterns/package.json
@@ -18,7 +18,10 @@
   "main": "./index.js",
   "module": "./index.js",
   "exports": {
-    ".": "./index.js",
+    ".": {
+      "endo:1": "./endo-eventual-send-exit.js",
+    },
     "./package.json": "./package.json"
   },
```

And tweaked the dapp dependencies:

```
$ ls -l ../node_modules/@endo/{far,patterns,eventual-send}
lrwxrwxrwx 1 connolly connolly 51 Aug 26 21:32 ../node_modules/@endo/eventual-send -> /home/connolly/projects/endo/packages/eventual-send
lrwxrwxrwx 1 connolly connolly 41 Aug 26 21:25 ../node_modules/@endo/far -> /home/connolly/projects/endo/packages/far
lrwxrwxrwx 1 connolly connolly 47 Aug 26 21:37 ../node_modules/@endo/patterns -> /home/connolly/projects/endo/packages/patterns/
```

Contract bundle shrinks to 17.M:

```
$ ~/projects/endo/node_modules/.bin/bundle-source -C endo:1 -T src/orca.contract.js | jq -r .endoZipBase64| base64 -d >exit.zip &&
 ls -sh exit.zip 
1.7M exit.zip
```

packages/compartment-mapper/src/import-archive-lite.js
```diff
@@ -124,15 +124,12 @@ const makeArchiveImportHookMaker = (
...
-            return {
-              record: await attenuateModuleHook(
-                moduleSpecifier,
-                record,
-                compartmentDescriptor.policy,
-                attenuators,
-              ),
-              specifier: moduleSpecifier,
-            };
+            return attenuateModuleHook(
```
dckc, line 127: The scope of this PR has expanded beyond a `test:` now, right? Seems like a `feat:` or a `chore:` in preparation for a `feat:`.

packages/compartment-mapper/src/link.js
```diff
@@ -60,6 +60,8 @@ const has = (object, key) => apply(hasOwnProperty, object, [key]);
  */
 const extensionImpliesLanguage = extension => extension !== 'js';
 
+const urlish = /^[a-z][a-z0-9+\-.]*:/;
```
dckc, line 63: This syntax is subject to an external design constraint; let's please cite it:

```suggestion
const urlish = /^[a-z][a-z0-9+\-.]*:/; // cf. section 4.3 of RFC 3986 Scheme URI Generic Syntax
```

optionally include https://www.rfc-editor.org/rfc/rfc3986#section-4.3

packages/compartment-mapper/src/policy.js
```diff
@@ -440,39 +440,85 @@ async function attenuateModule({
...
  *
  * @param {string} specifier - exit module name
- * @param {import('ses').ThirdPartyStaticModuleInterface} originalModuleRecord - reference to the exit module
+ * @param {import('ses').VirtualModuleSource | import('ses').StrictModuleDescriptor} moduleDescriptor - reference to the exit module
  * @param {import('./types.js').PackagePolicy} policy - local compartment policy
  * @param {import('./types.js').DeferredAttenuatorsProvider} attenuators - a key-value where attenuations can be found
- * @returns {Promise<import('ses').ThirdPartyStaticModuleInterface>} - the attenuated module
+ * @returns {Promise<import('ses').ModuleDescriptor>} - the attenuated module descriptor
  */
 export const attenuateModuleHook = async (
```
dckc, line 513: To what extent is this an externally visible API change?

I don't see any corresponding changes in tests, which suggests it's _not_ an external API change.
But that could be just a lack of test coverage. Or maybe it's only tested indirectly.

packages/ses/NEWS.md
```diff
@@ -1,9 +1,14 @@
 User-visible changes in `ses`:
 
-# Unreleased
+# Next version
+
+- Exports a `StrictModuleDescriptor` type that consists only of the
+  `NamespaceModuleDescriptor` and `SourceModuleDescriptor` that are mutually
+  supported by SES and XS.
```
dckc, line 7: Yup. PR is clearly more than just `test:`.

dckc (commented):
Yay! This definitely establishes that we can shrink bundles!

It's more than just a test, though. I'm not confident about approving some of the non-test material.

packages/ses/NEWS.md
kriskowal: Thanks. The left of the PR description has now caught up with the right of the PR description. The tags on individual commits also capture the conventions.

packages/compartment-mapper/src/policy.js
kriskowal: As part of convergence with XS, VirtualModuleSource is an alias for ThirdPartyStaticModuleInterface and ModuleDescriptor is a new type that captures all of the new and old types importHooks may return (and the range of types that an importHook adapter is expected to handle). The change here allows the full range of module descriptors to pass through this adapter when a LavaMoat policy is absent, which allows us to surface the whole module descriptor range in an exit module importHook. This was a bottleneck before.

So, the input range expands from VirtualModuleSource to also accept StrictModuleDescriptor, the forward-compatible subset of ModuleDescriptor. I did not feel the need to create a new obligation for backward compatibility past the next SES major version and wasn’t able to get TypeScript to play along with reliably differentiating a ModuleExportsNamespace from a VirtualModuleSource. In SES, we use a brand check for that.

This catches us up with what I expected to be working behavior, so is more like a bugfix than a feature, so I didn’t capture this in the notes.

packages/compartment-mapper/NEWS.md
kriskowal: This should be more clear now.

packages/compartment-mapper/src/policy.js
dckc: Was this API in any previous release?

dckc (approved):
looks good.

I'm not confident that I know all of the invariants that should be maintained. This review presumes they're covered by tests. If you're not confident about that, please get review from someone who is more familiar with the detailed design of this code.

kriskowal:
Summoning @naugtur since this touches policy enforcement code with delicate invariants.

naugtur:
I'll need some time to dig through this. I'll see if I can go through it before the meeting tomorrow.

packages/compartment-mapper/src/policy.js
```diff
@@ -440,39 +440,85 @@ async function attenuateModule({
...
       const attenuated = attenuate(ns);
       moduleExports.default = attenuated;
       assign(moduleExports, attenuated);
     },
-  });
+  };
+}
+
+/**
+ * Attenuates a module
```
naugtur, line 459: It'd be nice to have descriptions differ between the two functions

packages/compartment-mapper/src/link.js
```diff
@@ -227,6 +233,23 @@ const makeModuleMapHook = (
...
+      // When creating an archive of an application that imports a platform
+      // module like node:fs, we implicitly expect these to be provided by the
+      // host's importHook on the target platform.
+      return {
+        source: {
+          imports: [],
+          exports: [],
+          execute() {
+            throw new Error(
+              'Cannot import an application loaded strictly for analysis',
```
naugtur, line 246: I'll give it another chance later, but I don't understand this error message.

packages/compartment-mapper/src/policy.js
```diff
@@ -440,39 +440,85 @@ async function attenuateModule({
...
+ * @returns {Promise<import('ses').StrictModuleDescriptor>}
+ */
+async function attenuateModule({
+  attenuators,
+  attenuationDefinition,
+  moduleDescriptor,
+}) {
+  await null;
+  if ('source' in moduleDescriptor) {
+    const { source: moduleSource } = moduleDescriptor;
```
naugtur, line 473:
```
const moduleSource = normalizeModuleDescriptorOrSource(moduleDescriptor);
```
and we don't need all the duplication

packages/compartment-mapper/src/link.js
```diff
@@ -219,6 +224,7 @@ const makeModuleMapHook = (
   compartmentName,
   moduleDescriptors,
   scopeDescriptors,
+  archiveOnly,
```
boneskull, line 124: This flag is contemptuous and I wish it did not propagate.

I suppose a refactor to annihilate it would be beyond the scope of this PR, however.

packages/compartment-mapper/src/link.js
boneskull: It is well over a year later and it still evades understanding

boneskull (approved):
This looks OK to me, but I am not a user of archives, so "looks OK" is the ceiling for me

PR #2432, opened 2024-08-27
test: tests live binding on a let (demonstrates: #2431)

naugtur:
Refs: #2431


kriskowal:
This is a known-limitation that we must at least document better.

erights:
> This is a known-limitation that we must at least document better.

I'm surprised. Not urgent, but I am curious:

I know we handle such live-binding assignments in general through the with-proxy mechanism, and the hidden callback functions registered by the transformed module code. But why doesn't it work in this case?

kriskowal:
> > This is a known-limitation that we must at least document better.
> 
> I'm surprised. Not urgent, but I am curious:
> 
> I know we handle such live-binding assignments in general through the with-proxy mechanism, and the hidden callback functions registered by the transformed module code. But why doesn't it work in this case?

We don’t use a `with` block in the calling convention for bundles generated with our `bundle.js`. We also generally don’t use live bindings for “bootstrap bundles”. This is the tool we use to make the `ses/dist/*` scripts to replace Rollup, which we are going to use for more cases down the road, but it is a trusted and not a confined bundle format. We have not needed live binding support yet and I would prefer to avoid the `with` block indefinitely.

naugtur:
Added a test documenting lack of live bindings support.

PR #2433, opened 2024-08-27
refactor(bundle-source): Remove vestigial internal dev option

kriskowal:
This removes dead code. The bundler uses the `dev` flag only for discovery of `devDependencies` in `mapNodeModules`, and makes no further use of it in the bundling kit.


erights (approved):
LGTM

PR #2435, opened 2024-08-27
chore: Release 2024-08-27

kriskowal:
# `ses` v1.8.0 (2024-08-27)

- New `legacyRegeneratorRuntimeTaming: 'unsafe-ignore'` lockdown option to tame
  old `regenerator-runtime` (from 0.10.5 to 0.13.7).
- If lockdown's `errorTrapping: 'report'` mode is selected (possibly via the
  `'platform'`, or `'exit'` or `'abort'` modes), uncaught exceptions will be
  written to standard error with the new `SES_UNCAUGHT_EXCEPTION: ` prefix.
  This is intended to give valuable context to users of the system, especially
  when an uncaught exception is not an `Error` object, and therefore its origin
  may be hard to find in source code.

  This is not likely to affect most systems built with SES, as stderr is
  generally reserved for user-only messages.  If your SES system sends its
  stderr to a program which parses it, you may need to adapt that program to be
  tolerant of the `SES_UNCAUGHT_EXCEPTION: ` prefix.  Even for such programs, it
  is unlikely they are that sensitive to stderr formatting.

# `@endo/bundle-source` v3.4.0 (2024-08-27)

- Adds support for `--elide-comments` (`-e`) that blanks out the interior of
  comments in `endoScript` and `endoZipBase64` formats to reduce bundle size
  without compromising cursor position correspondence between source and bundle
  code.

# `@endo/evasive-transform` v1.3.0 (2024-08-27)

- Adds an `elideComments` option to replace the interior of comments with
  minimal blank space with identical cursor advancement behavior.

### DIFF
--- a/packages/base64/CHANGELOG.md
+++ b/packages/base64/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+### [1.0.7](https://github.com/endojs/endo/compare/@endo/base64@1.0.6...@endo/base64@1.0.7) (2024-08-27)
+
+**Note:** Version bump only for package @endo/base64
+
+
+
+
+
 ### [1.0.6](https://github.com/endojs/endo/compare/@endo/base64@1.0.5...@endo/base64@1.0.6) (2024-07-30)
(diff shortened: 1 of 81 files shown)


erights (approved):
LGTM!

Issue #31, opened 2019-10-27
`with (scopeProxy) { foo() }` leaks scopeProxy if foo mentions `this`.


erights:
Typing `valueOf()` into the evaluator shim playground (demo/index.html) and hitting the Evaluate button causes the error

> please report internal shim error: unexpected scope handler trap called: ownKeys

Unlike other surprising scopeProxy traps, this does not look like an engine bug. We see identical symptoms on Brave, FF, and Safari.

[edit by @kriskowal 2020-08-12:]

Actions needed (to be independently tracked, and this issue closed)

- ~~Create a specification proposal that introduces a `hiddenWith` symbol that `with` would respect.~~
- [ ] Document this known-limitation of SES-shim, ~~barring eventual engine support for a `hiddenWith` symbol.~~
- [ ] Confirm that ~~absent `hiddenWith`,~~ revealing the `ScopeProxy` does not break containment.

dckc:
This is a _use_ of `this`, not a _mention_ of `this`, no?

> ```js
> ...
> function foo() { return this; }
> ...
> ```

erights:
Well, it is also a mention of `this`. But good point. I genuinely do not know what the least confusing short title would be. "Uses `this`" can be misunderstood in other ways (such as "uses the value that `this` evaluates to"). I'll just let these notes in this issue thread clarify, and leave the title as is. Thanks.

dckc:
> Well, it is also a mention of `this`.

Ah. I see. Thanks.

mhofman:
We've noticed that some Web APIs such as `fetch` are sensitive to the `this` / context. If it's null-ish or a real `window` object, it works fine, but fails with a `TypeError` if any other object: `Failed to execute 'fetch' on 'Window': Illegal invocation` (message on v8/Chrome) or `'fetch' called on an object that does not implement interface Window` (message in SpiderMonkey/Gecko).

This fidelity bug of the shim causes code that would otherwise work to break.

erights:
How has multi-backflip affected this?

mhofman:
It still leaks, but only the `globalThis` regular object.

Main problem left is for module lexical created by the shim, as detailed in #912.

Should we rename the title of this issue? It might be good to keep tracking an engine level change allowing to disable the current with context behavior.

erights:
> Should we rename the title of this issue?

Please do!

kriskowal:
Narrowed scope and posted for follow-up: https://github.com/endojs/endo/issues/1954

PR #2436, opened 2024-08-28
fix(module-source): Sub non-conforming `ZWJ` prefixes with `CGJ`

SMotaal:
## Description

This PR changes the `HIDDEN_PREFIX` of `ModuleSource` from the non-conforming `$h\u200D_` zero-width joiner (`ZWJ`) notation to the conforming `$h\u034F_` combining grapheme joiner (`CGJ`) notation.

A future PR may address further changes to a `$\u034F`-prefixed and `\u034F$`-suffixed format as was suggested by @michaelfig in discussions.

### Motivation

This change is motivated after encountering a parsing error when using `rollup` which was traced back to the `$h\u200D_`-prefixed identifier in an `endoScript` bundle. More importantly, this is also motivated by the subsequent discovery that `rollup`'s implementation was actually conforming to the ECMAScript Specification when it was throwing this error.

To elaborate, while runtimes today will accept the special identifier notation that is currently being introduced by the `ModuleSource` rewrites, the current `$h\u200D_` zero-width joiner (`ZWJ`) notation does not conform to the specifications defined in the [ECMAScript Lexical Grammar](https://tc39.es/ecma262/#sec-names-and-keywords). In essence, what the specifications entail is that the character sequence for [Identifier Names](https://tc39.es/ecma262/#sec-identifier-names) once unescaped would be expected to match the `/^[$_\p{ID_Start}][$_\p{ID_Continue}]*$/u` pattern, aside from the additional `#` character prefix required in the case of private fields.

As such, one can test this in the console by evaluating the following:

```js
Object.fromEntries([String.raw`$h\u200D_`, String.raw`$h\u034F_`].map(id => [id, /^[$_\p{ID_Start}][$_\p{ID_Continue}]*$/u.test(JSON.parse(`"${id}"`))]))
```

The above would yield the following object in a runtime where the unicode escape sequences are retained:

```js
{$h\u200D_: false, $h\u034F_: true}
```

Digging closer in the Unicode Standard, it seems that the zero-width joiner (`ZWJ`) may indeed be used in a conforming notation per [Emoji Profile in Annex #31 of the Unicode Standard](http://www.unicode.org/reports/tr31/#Emoji_Profile), however this is not applicable for this purpose as it would require the use of emojis.

At this point, my suggestion to instead use the combining grapheme joiner (`CGJ`) is best articulated with this excerpt that I am borrowing from its canonical Wikipedia entry:

> However, in contrast to the zero-width joiner and similar characters, the `CGJ` does not affect whether the two letters are rendered separately or as a ligature or cursively joined—the default behavior for this is determined by the font.[^1]
>
>
> [^1]: https://en.wikipedia.org/wiki/Combining_grapheme_joiner

The Wikipedia article offers additional nuances about the differences, while the [Proposal for addition of COMBINING GRAPHEME JOINER](https://www.unicode.org/L2/L2000/00274-N2236-grapheme-joiner.htm) offers the necessary context about its intent.

It is fair to note that there are many uses of the zero-width joiner (`ZWJ`) already in the wild, and in fact there are currently `test262` tests for its occurrence. That said, unless those uses are conforming to the ECMAScript Specification and the Unicode Standard, they will limit code portability and adoption by users who may end up confused by failures similar to the one encountered with `rollup`.

Ultimately, with the reasonable recommendations to exercise caution when it comes to bundling `ses` and related sources that are best bundled with `bundleSource` instead, those sources may still need to be parsed with tools like `rollup` for different purposes that would be aligned with the expectations that they are being handled safely.

#### Substituting the invisible joiner character

A search across the monorepo for `(?:\u200d|\\u200d)_` yields only 3 files of interest:

- `packages/module-source/TESTS.md`
- `packages/module-source/src/hidden.js`
- `packages/module-source/test/module-source.test.js`

While making changes to the 3 files of interest, a distinction is made between matching `\$h\\u200d_` and `\$h\u200d_` where the replacements are respectively `$h\\u034f_` and `$h\u034f_`, along with their `$c` equivalents.

The search across the monorepo for `(?:\u200d|\\u200d)_` yields another 978 files that are not of interest found in:

- `packages/test262-runner/test262/test/language/expressions/class/elements`
- `packages/test262-runner/test262/test/language/statements/class/elements`

All those files remain unchanged.

#### Ensuring generic wording is used

For testing and other purposes where descriptive phrases are used to refer to the use of `ZWJ`, `CGJ` or other characters for this same intent, the phrase *"invisible joiner character"* is suggested.

### Security Considerations

**Does not apply to my knowledge**

### Scaling Considerations

**Does not apply to my knowledge**

**Does not apply to my knowledge**

### Testing Considerations

**See**: https://github.com/endojs/endo/pull/2436#issuecomment-2317183697

While the changes do not affect compatibility when the generated code is evaluated at runtime, there can potentially be compatibility concerns with tools that have been specifically designed to work with the current notation.

### Upgrade Considerations

**Does not apply to my knowledge**

### DIFF
--- a/packages/module-source/test/module-source.test.js
+++ b/packages/module-source/test/module-source.test.js
@@ -442,22 +442,22 @@ export default async function (arg) { return arg; }
   t.is(await ret, 'foo', 'F returns correctly');
 });
 
-test('zero width joiner is reserved', t => {
+test('invisible joiner character is reserved', t => {
   t.throws(() => {
-    const _ = new ModuleSource(`const $h\u200d_import = 123; $h\u200d_import`);
+    const _ = new ModuleSource(`const $h\u034f_import = 123; $h\u034f_import`);
   });
 });
 
-test('zero width joiner in constified variable is reserved', t => {
+test('invisible joiner character in constified variable is reserved', t => {
   t.throws(() => {
-    const _ = new ModuleSource(`const $c\u200d_myVar = 123; $c\u200d_myVar`);
+    const _ = new ModuleSource(`const $c\u034f_myVar = 123; $c\u034f_myVar`);
   });
 });
 
-test('zero width joiner is allowed in non-reserved words', t => {
+test('invisible joiner character is allowed in non-reserved words', t => {
   const { namespace } = initialize(
     t,
-    `const $h\u200d_import2 = 123; export default $h\u200d_import2`,
+    `const $h\u034f_import2 = 123; export default $h\u034f_import2`,
   );
   const { default: name } = namespace;
   t.is(name, 123);
@@ -729,6 +729,7 @@ test('source map generation', t => {
   const { __syncModuleProgram__ } = new ModuleSource(`'Hello, World!'`, {
     sourceUrl: 'must-appear-in-source.js',
     sourceMapUrl: 'must-not-appear-in-source.js',
+    // @ts-expect-error SourceMapHookDetails do not have those properties!
     sourceMapHook(sourceMap, { sourceUrl, sourceMapUrl, source }) {
       t.log(sourceMap);
       t.is(sourceMapUrl, 'must-not-appear-in-source.js');
--- a/packages/module-source/DESIGN.md
+++ b/packages/module-source/DESIGN.md
@@ -77,7 +77,7 @@ the module needs to be linked to and how to link their imports and exports.
 
 The functor source, the module transformed into a program, has the following
 shape.
-The names are additionally obscured with Unicode zero-width-joiners to avoid
+The names are additionally obscured with invisible joiner characters to avoid
 collisions with sensibly constructed modules, and the transformation preserves
 line numbers.
 
--- a/packages/module-source/TESTS.md
+++ b/packages/module-source/TESTS.md
@@ -12,7 +12,7 @@ Pairs of module source vs functor source. The output was generated.
 ```
 
 ```js
- (({   imports: $h‍_imports,   liveVar: $h‍_live,   onceVar: $h‍_once,  }) => {   $h‍_imports(new Map([]), []);   const { default: $c‍_default } = { default: bb };$h‍_once.default($c‍_default);
+ (({   imports: $h͏_imports,   liveVar: $h͏_live,   onceVar: $h͏_once,  }) => {   $h͏_imports(new Map([]), []);   const { default: $c͏_default } = { default: bb };$h͏_once.default($c͏_default);
 })
 ```
 
@@ -24,7 +24,7 @@ Pairs of module source vs functor source. The output was generated.
 ```
 
 ```js
- (({   imports: $h‍_imports,   liveVar: $h‍_live,   onceVar: $h‍_once,  }) => {   $h‍_imports(new Map([]), []);   const { default: $c‍_default } = { default: class {valueOf() {return 45;}} };$h‍_once.default($c‍_default);
+ (({   imports: $h͏_imports,   liveVar: $h͏_live,   onceVar: $h͏_once,  }) => {   $h͏_imports(new Map([]), []);   const { default: $c͏_default } = { default: class {valueOf() {return 45;}} };$h͏_once.default($c͏_default);
 })
 ```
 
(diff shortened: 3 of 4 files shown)


kriskowal (dismissed):
I’m in favor of this change. Since this is a design passed down from @michaelfig and @erights, I would ask for them to sign-off as well.

There are some tests for "zero width joiner" invariants in `@endo/module-source`. We need to make sure that CGJ is disallowed for all `source` in `ModuleSource(source)`.

SMotaal:
@kriskowal To your point:

> There are some tests for "zero width joiner" invariants in @endo/module-source. We need to make sure that CGJ is disallowed for all source in ModuleSource(source).

I only came across the following `throw` statements in `packages/module-source/**/*.js`:

https://github.com/endojs/endo/blob/67141a177bef32e037f454a696a4532999560ff8/packages/module-source/src/babelPlugin.js#L268-L285

The code is structured to ensure there is a single source of truth, so this holds with `CGJ` all the same.

---

That said, I am also not sure if the following could prove problematic down the road:

```ts
ModuleSource<CGJ>(bundleSource<ZWJ>(sourcePath, { format: 'endoScript' }).source);
```

I don't foresee this being a problem, but I wonder if adding a test case could be reasonable while we're here.

kriskowal:
I’m not worried about mixing CGJ and ZWJ. The version of `ModuleSource` will censor and produce the same one and it doesn’t depend on version coördination with other packages.

I just want to make the names of the tests have "combining grapheme joiner" instead of "zero-width joiner" now, or some abstract term that would apply regardless of which is used. Otherwise, this is good to go.

packages/module-source/test/module-source.test.js
```diff
@@ -444,20 +444,20 @@ export default async function (arg) { return arg; }
 
 test('zero width joiner is reserved', t => {
   t.throws(() => {
-    const _ = new ModuleSource(`const $h\u200d_import = 123; $h\u200d_import`);
+    const _ = new ModuleSource(`const $h\u034f_import = 123; $h\u034f_import`);
   });
 });
 
 test('zero width joiner in constified variable is reserved', t => {
```
michaelfig, line 451:
```suggestion
test('joiner in constified variable is reserved', t => {
```

packages/module-source/test/module-source.test.js
```diff
@@ -444,20 +444,20 @@ export default async function (arg) { return arg; }
 
 test('zero width joiner is reserved', t => {
```
michaelfig, line 445:
```suggestion
test('joiner is reserved', t => {
```

packages/module-source/test/module-source.test.js
```diff
@@ -444,20 +444,20 @@ export default async function (arg) { return arg; }
...
 });
 
 test('zero width joiner in constified variable is reserved', t => {
   t.throws(() => {
-    const _ = new ModuleSource(`const $c\u200d_myVar = 123; $c\u200d_myVar`);
+    const _ = new ModuleSource(`const $c\u034f_myVar = 123; $c\u034f_myVar`);
   });
 });
 
 test('zero width joiner is allowed in non-reserved words', t => {
```
michaelfig, line 457:
```suggestion
test('joiner is allowed in non-reserved words', t => {
```

michaelfig (approved):
LGTM!

I marked a few cases where the test description could be loosened to just "joiner".  Up to you as to what you do with these descriptions.

packages/module-source/test/module-source.test.js
SMotaal:
```suggestion
test('invisible joiner character is reserved', t => {
```

packages/module-source/test/module-source.test.js
SMotaal:
```suggestion
test('invisible joiner character in constified variable is reserved', t => {
```

packages/module-source/test/module-source.test.js
SMotaal:
```suggestion
test('invisible joiner character is allowed in non-reserved words', t => {
```

PR #2439, opened 2024-08-29
fix(daemon): binding Promise breaks exo

kumavis:
this fixes Exo compatibility for Endo Daemon

### DIFF
--- a/packages/daemon/src/web-page.js
+++ b/packages/daemon/src/web-page.js
@@ -65,6 +65,8 @@ const hardenedEndowments = harden({
   TextEncoder,
   TextDecoder,
   URL,
+  // This one is important for exo to work.
+  Promise,
 });
 
 const globalProps = collectPropsAndBind(window);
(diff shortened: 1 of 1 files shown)


packages/daemon/src/web-page.js
```diff
@@ -65,6 +65,8 @@ const hardenedEndowments = harden({
   TextEncoder,
   TextDecoder,
   URL,
+  // This one is important for exo to work.
```
kriskowal, line 68: @mhofman or @michaelfig can I ask you to help @kumavis expand this comment to explain why and where to look for more? I think this is the work of the handled promise shim, but I did not know it replaced `globalThis.Promise`.

kriskowal (approved):

packages/daemon/src/web-page.js
michaelfig: `@endo/eventual-send` does not mutate `Promise` in any way, so I'm confused as to why this line is necessary.

mhofman: I really don't understand why this is necessary. Is there anything in your environment that might be replacing the global `Promise`? Maybe some telemetry library?

PR #2440, opened 2024-08-30
docs: add titles to relative path hrefs in markdown files

mujahidkay:
## Description

As part of keeping the [docs site up-to-date with API docs](https://docs.google.com/document/d/1h-wQzmGVtHVPGp8fNsSYormZQoRFRDMRNZksm37Nwm8/edit?usp=sharing) initiative, there are certain instances of hyperlinks in a few md files that generate deadlinks and are hence flagged by `vitepress`.  Since we use relative paths to reference md files within the same repo to make them [navigable in external IDEs](https://github.com/endojs/endo/pull/1269), simply replacing them with their github url would be a downgrade in terms of experience. 

This PR proposes that we add a **_title_** against each href which necessarily just equals the github url of that file. This helps us with two things:
 - maintain IDE navigation
 - do pre-processing on such links to replace href with the title for docs site

Currently, the plan is to use a `linkTransformer` of sorts to achieve this. On the docs site, these links will get replaced by their github URL thereby circumventing the 404s they would have led to (if we ignored the dead link warnings). Here's an initial version of code that I have tested to make it happen:

```js
export default (md) => {
    md.renderer.rules.link_open = (tokens, idx, options, _, self) => {
      const token = tokens[idx];
      const hrefIndex = token.attrIndex('href');
      const titleIndex = token.attrIndex('title');
  
      if (hrefIndex >= 0 && titleIndex >= 0) {
        const href = token.attrs[hrefIndex][1];
        const title = token.attrs[titleIndex][1];
  
        if (href.endsWith('.md') && title.startsWith('http')) {

          token.attrs[hrefIndex][1] = title;
          token.attrs.splice(titleIndex, 1);
        }
      }
  
      return self.renderToken(tokens, idx, options);
    };
  };
```


packages/pass-style/doc/copyArray-guarantees.md
```diff
@@ -17,8 +17,7 @@ The input validation check `assertCopyArray(arr)` asserts that `passStyleOf(arr)
 # How do I enumerate thee, let me list the ways
 
 Why these properties restrictions?
-JavaScript has a [tremendous number of different constructs for enumerating the
-properties](enumerating-properties.md) of an object, with different semantics
+JavaScript has a [tremendous number of different constructs for enumerating the properties](enumerating-properties.md "https://github.com/endojs/endo/blob/master/packages/pass-style/doc/enumerating-properties.md") of an object, with different semantics
```
turadg, line 20: the contents of the repo should not assume where the repo is located on the web or what branch it's in.

let's have a Slack convo to figure out an alternate solution to the problem

turadg (changes requested):

packages/pass-style/doc/copyArray-guarantees.md
mujahidkay: > the contents of the repo should not assume where the repo is located on the web or what branch it's in

fair enough

mujahidkay:
Closing this as it is de-prioritized for now in favor of https://github.com/Agoric/documentation/issues/1187

PR #2279, opened 2024-05-09
Endo Daemon Demo: multiplayer 1kce

kumavis:
building off of https://github.com/endojs/endo/pull/1877 and previous demos

### summary of changes

- [ ] daemon exposes formula types
f221f52 feat(daemon): directories expose formula types

- [ ] daemon namespace command (for multiplayer dev)
be4f610 feat(where): allow all paths to be set by env vars
4988cc7 feat(cli): load .env file from cwd
fb89f7b feat(cli): add namespace command for making new daemons

- [ ] daemon exposes pubusb for makeChangeTopic util
c85f86e feat(daemon): expose pubsub

- [ ] bug fix for installed weblets https://github.com/endojs/endo/pull/2439
b7804ce fix(daemon): binding Promise breaks exo

- [ ] import + update familiar-chat + 1kce game
1c88619 wip: add 1kce and grain packages from endo-familiar-chat-scifi
1fa0cde wip: add familiar-chat package from endo-familiar-chat-scifi
9a3d3ca fix(familiar-chat): get basic inventory working
e5832bb fix(familiar-chat): fix basic chat functionality
b878b22 fix(familiar-chat): fix app install
87cfb32 fix(1kce): fix make deck + include deck and game bundles
039c4c0 fix(1kce): game working again

PR #2441, opened 2024-08-30
feat(where): allow all paths to be set by env vars

kumavis:
from https://github.com/endojs/endo/pull/2279
part of a feature for easily creating new Endo daemons (eg for development)


kriskowal (commented):
This is good, and I think it would also be good to have a single environment variable to fall back to if one of these isn’t specified. I think it has to be a name like `export ENDO=alice` since the Windows named pipe, in particular, can’t be in a subdirectory of a shared root with the others. The same limitation applies to temporary files, which should be on a tmpfs mount (RAM filesystem).

PR #2442, opened 2024-09-03
docs(patterns): start with simple example

dckc:

Start the README with a simple example (taken from a test). Make the `PatternMatchers` interface that defines `M`'s methods easier to find.

 - Demote some material to the JSDoc of relevant types.
 - add various links
 - Demote some stuff to CONTRIBUTING.

### Security / Scaling / Testing / Compatibility / Upgrade Considerations

n/a


packages/patterns/README.md
```diff
@@ -1,103 +1,32 @@
...
-The set of all primitive values is a strict subset of Data, which is a strict subset of Keys.
+`M` also has {@link GuardMakers} methods to make {@link InterfaceGuard}s that use Patterns to characterize dynamic behavior such as method argument/response signatures and promise awaiting. The {@link @endo/exo!} package uses `InterfaceGuard`s (each of which maps a collection of method names to their respective method guards) as the first level of defense for Exo objects against malformed input.
 
-The set of all primitive values is also a strict subset of the set of Scalars (which is the union of Primitives and Capabilities [i.e., Remotables and Promises]). The union of primitive values and Remotables is a strict subset of Keys.
 
-Keys is a strict subset of Patterns, which is a strict subset of Passables.
+## Key Equality, Containers
 
-TODO: Include a diagram visually demonstrating the following.
+Builds on {@link @endo/pass-style!} as described in [`kindOf` and `passStyleOf` levels of abstraction](./docs/marshal-vs-patterns-level.md) to define higher level data types as individual refinements of Passable CopyTagged records (PassStyle "tagged"):
```
dckc, line 24: I don't suppose this sort of link markup works in the https://www.npmjs.com/package/@endo/patterns context.

We should have a link from there to https://endojs.github.io/endo/modules/_endo_patterns.html or at least to https://endojs.github.io/, I suppose.

packages/patterns/README.md
```diff
@@ -1,103 +1,32 @@
...
+import { M, mustMatch } from '@endo/patterns';
 
-Guards do not yet exist as distinct kinds, so we ignore them for now. TODO: Expand this if kinds expand to include guards.
+const specimen = { foo: 3, bar: 4 };
 
-As mentioned above, `keyEQ` is pass-invariant: if passing `xa` from vatA to vatB arrives as `xb`, and likewise `ya` and `yb`, then `keyEQ(xa,ya)` iff `keyEQ(xb,yb)`. And because we do not wish to give Promises, Errors, or unrecognized CopyTagged values any useful pass-invariant equality, a Key may not include any of those.
+const expected = M.splitRecord(
+   { foo: M.number() },
+   { bar: M.string(), baz: M.number() },
+);
```
warner, line 14: Would it be correct to have this say:
```js
  { foo: M.number() }, // required properties
  { bar: M.string(), baz: M.number() }, // optional
```

so a newbie to `M` like me gets a hint as to why there are two arguments?

packages/patterns/src/types.js
```diff
@@ -29,9 +29,38 @@ export {};
  * CopyTaggeds recognized as Keys.
  *
  * Distributed equality is location independent.
- * The same two Keys, passed to another location, will be `keyEQ` there iff
- * they are `keyEQ` here. (`keyEQ` tests equality according to the
+ * The same two Keys, passed to another location, will be {@link keyEQ} there iff
+ * they are {@link keyEQ} here. ({@link keyEQ} tests equality according to the
  * key distributed equality semantics.)
+ *
```
warner, line 35: I'm wondering if the rank/key order details would fit better in CONTRIBUTING.md along with the other details extracted from README. I don't think I'd look for details like that in a types.js file (although tbh I'm not sure "CONTRIBUTING.md" is where I'd look either, maybe something like IMPLEMENTATION-DETAILS.md).

warner (approved):
Looks like a good improvement! Note I have not verified that the moved text did not change, just that the additions to README are helpful to me.

packages/patterns/src/types.js
dckc: I had it in CONTRIBUTING for a bit, but rankCompare and keyCompare are part of the api, not an implementation detail.

The material could perhaps move to those functions, but I suspect the editorial work would exceed my timebox for this.

packages/patterns/README.md
```diff
@@ -1,103 +1,32 @@
 # `@endo/patterns`
 
-Builds on [`@endo/pass-style`](https://www.npmjs.com/package/@endo/pass-style) as described in [`kindOf` and `passStyleOf` levels of abstraction](./docs/marshal-vs-patterns-level.md) to define higher level data types as individual refinements of Passable CopyTagged records (PassStyle "tagged"):
-   - CopySet -- a collection of unique distinguishable Keys
-   - CopyBag -- a collection of entries associating a unique distinguishable Key with a positive integer count (see [Multiset](https://en.wikipedia.org/wiki/Multiset)).
-   - CopyMap -- a collection of entries associating a unique distinguishable Key with a Passable
-   - Matcher -- a predicate characterizing a subset of Passables, such as "strings" or "8-bit unsigned integer numbers" or "CopyArrays of Remotables"
+The main export from the package is an {@link M} namespace object, for making a variety of Matchers (hence "M"). See {@link PatternMatchers} for relevant methods.
```
erights, line 3: I'll guess that this `{@link M}` form is better for the generated documentation, which is great. But when viewed directly on github it does not render as a link or anything useful. Especially a README.md file should be pleasantly readable directly on github.

Can we get both somehow?

dckc: We can write some links one way and some the other. I can sort of imagine demoting the `{@link X}` things so that they don't get in the way in non-typedoc contexts.

packages/patterns/README.md
gibson042: `@link` is JSDoc: https://jsdoc.app/tags-inline-link

Issue #926, opened 2021-11-03
Lots of whitespace differences introduced by endo bundler


erights:
On the right is the original `lockdown-shim.js`. On the left is the corresponding portion of `lockdown.umd.js` aligned for a visual line by line comparison. There are lots of little divergences. I do not remember seeing these divergences when we were using some previous bundler (rollup?).

These differences will impede the debugging experience. It means that at least the columns are different, and so we cannot simply sourceURL the original source file to appear in the debugger in lieu of the bundled source.

![image](https://user-images.githubusercontent.com/273868/140021609-2fa99c34-99e3-45c4-8268-ee25e0e67d44.png)

Related issues:
* https://github.com/endojs/endo/issues/400

kriskowal:
We discussed this and resolved that this probably can be addressed by patching the babel-generator code that catches up with the cursor of the next written AST, such that it advances the column in addition to the line.

kriskowal:
We also discussed trying to preserve untouched subtrees of the AST by reusing the original source. This is possible but more invasive. It requires annotating which nodes were “touched” and modifying bundle-source to attend to that detail.

pyramation:
I would encourage the use of source maps vs. expecting babel to allow customization of whitespace:

https://github.com/evanw/node-source-map-support
https://github.com/chocolateboy/babel-plugin-source-map-support

If you integrate source maps correctly, all debugging should theoretically use the source code files instead of the transpiled code in stack traces.

If possible, I'd encourage and suggest finding a path that is not dependent on comments or whitespace output of babel. As far as I understand the deparser (babel/generate) or the transformer, whitespace (and comments) are two areas that I've empirically found that babel doesn't address perfectly.

Given the dependency on babel (which is the industry standard), if sourcemaps don't solve this issue, I'd suggest making an issue in babel. Looks like it has been brought up before as early as 2015, but has been closed https://github.com/babel/babel/issues/497

pyramation:
also, looking more deeply into the original babel issue I found a linked issue that helped me find this library: https://github.com/js-cst-tokens/cst-tokens 

they are not using ASTs, but rather, CSTs which appear to preserve whitespace. I would still think sourcemaps are the better path, but figured I'd share this link.

erights:
[...]

PR #2444, opened 2024-09-03
feat: Preserve formatting

kriskowal:
Closes: #926

## Description

This change leverages new Babel support for format preservation.

### Security Considerations

None

### Scaling Considerations

None

None

### Testing Considerations

This change includes a test to perform a narrow spot-check of the verbatim output of a module that should be largely preserved.

None

### Upgrade Considerations

None

- [ ] NEWS.md


packages/module-source/src/transformSource.js
```diff
@@ -26,22 +26,31 @@ export const makeTransformSource = (makeModulePlugins, babel = null) => {
...
-      });
+      generateBabel(
+        ast,
+        {
+          sourceFileName: sourceMapUrl,
+          sourceMaps,
+          inputSourceMap: sourceMap,
+          retainLines: true,
+          preserveFormat: true,
+          compact: true,
```
kriskowal, line 49: @nicoloribaudo Reviewing https://github.com/babel/babel/pull/16708/files#diff-ca2bda59eec9c35846e26c1c6247759c92c26357b80d4dc881fcdaf12df1d7e6R35-R39, it seems like `compact: true` here should be causing your code to throw, so I suspect this isn’t exercising the new path in my local testing.

kriskowal: Found it. Needed to swap out our fork `@agoric/babel-generator` back to `@babel/generator`.

kriskowal:
@nicolo-ribaudo,

I’ll have a push soon, but this looks correct, actually:

Pre-transform:
```js
function TokenType() {}
const beforeExpr = 0;

export function createBinop(name, binop) {
  return new TokenType(name, {
    beforeExpr,
    binop,
  });
}
```

Post-transform:
```js
function TokenType() {}
const beforeExpr = 0;

       function createBinop(name, binop) {
  return new TokenType(name, {
    beforeExpr,
    binop,
  });
}
```

I’ll note that there’s a _weird but explicable_ offset on the function declaration _because we have erased `export`_!

This was the previous (undesirable) effect:
```js
function TokenType() { }
const beforeExpr=  0;

function        createBinop(name, binop) {
  return new TokenType(name, {
    beforeExpr,
    binop});

 }
})()
```

And the test I’ve proposed is not yet passing because the generator doesn’t add a newline at the end of the file. That presumably means that it needs to catch up with the final line and column of the file to preserve trailing whitespace, or just add a single newline to the end to make sure the output is a valid UNIX text file.

kriskowal:
I’m mistaken above about the newlines issue. I’m updating the test fixture and it looks like the test will pass as-is, no changes needed from Babel.

packages/module-source/test/fixtures/unmangled.js
```diff
@@ -0,0 +1,10 @@
+({   imports: $h͏_imports,   liveVar: $h͏_live,   onceVar: $h͏_once,   importMeta: $h͏____meta, }) => (function () { 'use strict';   $h͏_imports([]);Object.defineProperty(createBinop, 'name', {value: "createBinop"});$h͏_once.createBinop(createBinop);   // deliberately offset
```
nicolo-ribaudo, line 1: I'm actually surprised by the spaces here. My expectation was that this would have been printed as compact as possible to increase the chances that the next node could be printed in the right place.

mhofman, line 1: Where is the closing paren for the function expression here?

nicolo-ribaudo: Mmm it should be at then end of the file right? It looks like I forgot to flush some tokens somewhere 🤔

nicolo-ribaudo: Actually, cloning the PR locally I get this additional line at the end of the file:
```js
})()
```

Probably @kriskowal accidentally deleted it.

nicolo-ribaudo: Oh I see. This code with spaces is generated by you with string concatenation after that Babel runs, and not by Babel itself.

kriskowal: Indeed. I also find the spacing arbitrary, emerging from some unnecessary space in the formatting of the code. I’ve pushed a commit to make that compact as expected, regardless of if not being generated by Babel.

kriskowal: Pushed the fix for this. It was present locally when I got the test to pass.

nicolo-ribaudo:
From the failure it seems like there is still a bug in Babel. I probably need to enforce semicolons in case where the printer doesn't put them, so that even if the next token is on the same line it works.

kriskowal:
> From the failure it seems like there is still a bug in Babel. I probably need to enforce semicolons in case where the printer doesn't put them, so that even if the next token is on the same line it works.

I’ve got good results for `@endo/module-source` and `@endo/evasive-transform` in their tests, which are not comprehensive. The remaining error is in composition with a sourcemap and rollup, and I need to spend more time isolating it. I’m not yet sure it’s a problem with Babel, but the problem does go away if we use compact instead of preserveFormat for the relevant case.

nicolo-ribaudo:
I found the root cause of the error, it's indeed a Babel bug.

```js
// input
const results = {
    answer1,
  answer2,
  answer3,
        answer4,
  answer5,
}
```
```js
// run babel
const out = babel.transformSync(code, {
  configFile: false,
  parserOpts: { tokens: true },
  generatorOpts: { retainLines: true, experimental_preserveFormat: true },
  plugins: [
    ({ template }) => ({
      visitor: {
        Program(path) {
          path.pushContainer("body", template.statement.ast`hello;`);
        },
      },
    }),
  ],
});
```

The VariableDeclaration is printed without a trailing semicolon to preserve its original format, but then the new injected `hello;` is printed on the same line as `}` (to avoid increasing the line number), causing a syntax error.

kriskowal:
We seem to still have some defects in a similar vein. We have a failing test that exists to ensure that error stacks report true line numbers when using one of our older bundle formats.

The source is:
```js
export const message = `You're great!`;
export const makeError = msg => Error(msg);
// Without an evasive transform, the following comment will trip the SES censor
// for dynamic imports. */
/** @type {import('./types.js').EncourageFn} */
export const encourage = nick => `Hey ${nick}!  ${message}`;
```

The current transform output preserves the line number.
```js
(function getExport(require, exports) {   'use strict';   const module = { exports };     'use strict';Object.defineProperty(exports,'__esModule',{value:true});const message=`You're great!`;
const makeError=(msg)=>Error(msg);
/* Without an evasive transform, the following comment will trip the SES censor*/
/* for dynamic imports. *X/*/
/** @type {IMPORT('./types.js').EncourageFn} */
const encourage=(nick)=>`Hey ${nick}!  ${message}`;exports.encourage=encourage;exports.makeError=makeError;exports.message=message;
  return module.exports;
})
//# sourceURL=/bundled-source/.../encourage.js
```

The new transformed output injects a lot of unexpected space:
```js
(function getExport(require, exports) {   'use strict';   const module = { exports };     'use strict';

Object.defineProperty(exports,'__esModule' , { value: true });

const                                                                                                                                                                                                                                                                                                                                                                                                                                                                message=`You're great!`;
const                                                                                                                                                                                                                                                                                                                                                                                                                                                                makeError=(msg)=>Error(msg);/* Without an evasive transform, the following comment will trip the SES censor*//* for dynamic imports. *X/*//** @type {IMPORT('./types.js').EncourageFn} */



const                                                                                                                                                                                                                                                                                                                                                                                                                                                                encourage=(nick)=>`Hey ${nick}!  ${message}`;

exports.                                                                                                                                                                                                                                                                                                                                                                                                                                                                encourage=encourage;
exports.                                                                                                                                                                                                                                                                                                                                                                                                                                                                makeError=makeError;
exports.                                                                                                                                                                                                                                                                                                                                                                                                                                                                message=message;
  return module.exports;
})
//# sourceURL=/bundled-source/.../encourage.js
```

This is the only failing test locally, but is also the only test that is sensitive to stack trace line numbers, so I should make more tests and also attempt to isolate this more. You will note that is a Babel transform that occurs after a prior Rollup transform, which is not present in our modern bundle formats.

nicolo-ribaudo:
I'll take a look. I'm also working on getting a large part of our transform tests to go through this printer to have a bigger corpus of test cases.

That's a funny output, I wonder how it's deciding to inject all those spaces there.

nicolo-ribaudo:
I started taking a look at that. Something I noticed is that if I disable the "source map unmap" functionality you have in `transformAst`, then the output looks ok.

It's not surprising that it has an effect, given that Babel relies on location info to preserve the format, but I don't know yet if it's a bug in Babel on in how you remap locations.

nicolo-ribaudo:
Ok so what's happening is that this is the input code given to Babel:
```js
'use strict';

Object.defineProperty(exports, '__esModule', { value: true });

const message = `You're great!`;
const makeError = msg => Error(msg);
// Without an evasive transform, the following comment will trip the SES censor
// for dynamic imports. */
/** @type {import('./types.js').EncourageFn} */
const encourage = nick => `Hey ${nick}!  ${message}`;

exports.encourage = encourage;
exports.makeError = makeError;
exports.message = message;
```

And Babel is trying to preserve the formatting of _this_ input code (and not of what the input code for Rollup was, since Babel doesn't know about it). That's why there are all those "extra" newlines.

You need to also re-map the locations `.tokens` array in the AST, since that's what Babel looks at to understand where the original tokens where in the original code. Or if you can, it would be even better to actually parse the original pre-rollup code and give those tokens and that original source code to `@babel/generator` (while still using the post-rollup AST with remapped locations).

The "fake locations" you generate are weird btw. In the test above, in ``const message=`You're great!`;``:
- ``const message = `You're great!`;`` has start and end loc 1:12
- ``message = `You're great!` `` has start and end loc 1:13

The extra horizontal space is due to a failing check in Babel. I was assuming it would have never happen, but it's happening due to the mismatch between the locations on the AST nodes and the locations on the tokens. If after fixing the tokens locations in your logic the problem persists, I can adjust the check in Babel.

kriskowal:
Thank you @nicolo-ribaudo, this implies a couple good mitigation options I can look into and I believe we’re unblocked.

@michaelfig My intention is to look into making Babel’s new preserveFormat mode work with the older getExport and nestedEvaluate in composition with Rollup by ensuring we return to text between stages. If that exceeds my (short) timebox, my intention is to just disable preserveFormat and live with the current experience. My intention is to (soon) migrate our remaining usage of getExport and nestedEvaluate to the new endoScript format, and maybe even introduce an endoNativeScript format to take advantage of XS Compartment and further improve debug experience for kernel bootstrap scripts.

kriskowal:
I’ve spoken to @michaelfig and we resolved that we have reimplemented enough of Rollup at this stage to pretty easily replace the internals of the legacy getExport and nestedEvaluator generators with ones we trust. So, we will not attempt to get Rollup and Babel to play well together. The end result is that this test will pass without modification, since it won’t be downgraded from ESM to CJS by Rollup, but will go through our ModuleSource shim instead, which will benefit from preserveFormat.

nicolo-ribaudo:
Perfect -- if you'll need anything else you know where to find me :)

packages/bundle-source/src/script.js
```diff
@@ -79,8 +81,9 @@ export async function bundleScript(
     },
   );
 
-  const source = await makeBundle(powers, entry, {
-    dev,
+  let source = await makeFunctor(powers, entry, {
+    dev:
+      dev || moduleFormat === 'nestedEvaluate' || moduleFormat === 'getExport',
```
gibson042, line 86: Add a comment indicating that this automatic implication of `dev` is limited to "nestedEvaluate" and "getExport", and will not be extended to any future format.

packages/bundle-source/src/script.js
```diff
@@ -89,13 +92,33 @@ export async function bundleScript(
     workspaceModuleLanguageForExtension,
     moduleTransforms,
     sourceMapHook,
+    useEvaluate: moduleFormat === 'nestedEvaluate',
+    sourceUrlPrefix: '/bundled-source/.../',
+    format:
+      moduleFormat === 'nestedEvaluate' || moduleFormat === 'getExport'
+        ? 'cjs'
+        : undefined,
```
gibson042, line 100: Maybe explain this as well.

packages/bundle-source/src/script.js
```diff
@@ -89,13 +92,33 @@ export async function bundleScript(
...
   });
 
+  if (moduleFormat === 'endoScript') {
+    source = `${source}()`;
```
gibson042, line 104: For parity and resiliency:
```suggestion
    source = `(${source})()`;
```

gibson042 (approved):
:party:

nicolo-ribaudo:
Given that this was a significant change in Babel, would it be possible for you to validate it with some of your internal packages that use Endo before actually releasing this PR?

kriskowal:
> Given that this was a significant change in Babel, would it be possible for you to validate it with some of your internal packages that use Endo before actually releasing this PR?

This is our Agoric SDK integration test pinned to this branch. https://github.com/Agoric/agoric-sdk/pull/10781

There are limitations to the reach of these integration tests, but if the new Babel would build a broken bundle for Agoric’s worker kernel, these integration tests would (and did) discover those problems. We hope to go farther and pin to the Endo branch in tests that run in Docker for much more integrated scenarios. (@mhofman reminded me that these tests have their own lockfiles.)

In short, I’m not expecting any Babel concerns to block a release at this stage, but I am expecting that we will find minor irregularities as more engineers begin debugging with contracts using the new bundler.

PR #2446, opened 2024-09-04
chore: Delete tsbuildinfo after pack

kriskowal:
Cleans up more dross after building packages for npm.

```sh
#!/bin/bash
set -ueo pipefail
while read -r PKGJSON; do
  echo "$PKGJSON"
  jq '.scripts.postpack = $postpack' \
    --arg postpack "git clean -f '*.d.ts*' '*.tsbuildinfo'" \
    "$PKGJSON" |
    sponge "$PKGJSON"
done < <(
  npm query '.workspace:not([private])' |
  jq -r '.[].location | "\(.)/package.json"'
)
```

### DIFF
--- a/packages/base64/package.json
+++ b/packages/base64/package.json
@@ -33,7 +33,7 @@
   "scripts": {
     "build": "exit 0",
     "prepack": "tsc --build tsconfig.build.json",
-    "postpack": "git clean -f '*.d.ts*'",
+    "postpack": "git clean -f '*.d.ts*' '*.tsbuildinfo'",
     "cover": "c8 ava",
     "lint": "yarn lint:types && yarn lint:eslint",
     "lint-fix": "eslint --fix .",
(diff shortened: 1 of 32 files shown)


packages/base64/package.json
```diff
@@ -33,7 +33,7 @@
   "scripts": {
     "build": "exit 0",
     "prepack": "tsc --build tsconfig.build.json",
-    "postpack": "git clean -f '*.d.ts*'",
+    "postpack": "git clean -f '*.d.ts*' tsconfig.build.tsbuildinfo",
```
turadg, line 36: might as well delete any future ones
```suggestion
    "postpack": "git clean -f '*.d.ts*' *.tsbuildinfo",
```

kriskowal: Any thoughts on whether to defer the glob expansion to git?

turadg: nope

turadg: I've developed an opinion: https://github.com/Agoric/agoric-sdk/pull/9795/commits/e119569b394e27cdee6718dfafddc04ba90cfaed

hazarded a guess that git might do it a little faster

kriskowal: Following your lead. The other effect is that it will match files in subdirectories.

turadg (approved):

PR #2449, opened 2024-09-06
feat(gems): add gems package

kumavis:
what if


kumavis:
https://github.com/endojs/endo/pull/2473

PR #2454, opened 2024-09-13
fix(cli): Improve number parsing

sirtimid:
## Description

This PR introduces a new `parseNumber` function that safely parses input as a number on cli commands.
Additional tests have been added to ensure this new functionality works correctly.

### Security Considerations

none
### Scaling Considerations

none

none
### Testing Considerations

none

none
### Upgrade Considerations

none

### DIFF
--- a/packages/cli/src/number-parse.js
+++ b/packages/cli/src/number-parse.js
@@ -1,0 +1,15 @@
+/**
+ * Parses the input and returns it as a number if it's valid, otherwise throws error.
+ *
+ * @param {string} input
+ * @returns {number}
+ */
+export const parseNumber = input => {
+  const result = /[0-9]/.test(input || '') ? Number(input) : NaN;
+
+  if (Number.isNaN(result)) {
+    throw new Error(`Invalid number: ${input}`);
+  }
+
+  return result;
+};
--- a/packages/cli/test/number-parse.test.js
+++ b/packages/cli/test/number-parse.test.js
@@ -1,0 +1,36 @@
+import test from 'ava';
+import { parseNumber } from '../src/number-parse.js';
+
+test('returns the number when input is a valid numeric string', t => {
+  t.is(parseNumber('42'), 42);
+  t.is(parseNumber('-40.0'), -40.0);
+  t.is(parseNumber('.5'), 0.5);
+  t.is(parseNumber('1.337e3'), 1337);
+  t.is(parseNumber(' +1 '), 1);
+  t.is(parseNumber('0B111'), 7);
+  t.is(parseNumber('0o040'), 32);
+  t.is(parseNumber('0xf00'), 3840);
+  t.is(parseNumber('0xF00'), 3840);
+  t.is(parseNumber('   -40.0   '), -40.0);
+  // Test for binary, octal and hexadecimal
+  t.is(parseNumber('0B111'), 7);
+  t.is(parseNumber('0o040'), 32);
+  t.is(parseNumber('0xf00'), 3840);
+});
+
+test('throws an error when input is not a valid numeric string', t => {
+  const badStrings = [
+    'f00',
+    'F00',
+    'NaN',
+    'Infinity',
+    '-Infinity',
+    '7up',
+    ' ', // U+1680 OGHAM SPACE MARK (which is whitespace)
+    '',
+    '   ',
+  ];
+  for (const bad of badStrings) {
+    t.throws(() => parseNumber(bad), { message: `Invalid number: ${bad}` });
+  }
+});
--- a/packages/cli/src/commands/adopt.js
+++ b/packages/cli/src/commands/adopt.js
@@ -3,6 +3,7 @@ import os from 'os';
 import { E } from '@endo/far';
 import { withEndoAgent } from '../context.js';
 import { parsePetNamePath } from '../pet-name.js';
+import { parseNumber } from '../number-parse.js';
 
 export const adoptCommand = async ({
   messageNumberText,
@@ -11,7 +12,9 @@ export const adoptCommand = async ({
   agentNames,
 }) =>
   withEndoAgent(agentNames, { os, process }, async ({ agent }) => {
-    // TODO less bad number parsing.
-    const messageNumber = Number(messageNumberText);
-    await E(agent).adopt(messageNumber, edgeName, parsePetNamePath(name));
+    await E(agent).adopt(
+      parseNumber(messageNumberText),
+      edgeName,
+      parsePetNamePath(name),
+    );
   });
--- a/packages/cli/src/commands/reject.js
+++ b/packages/cli/src/commands/reject.js
@@ -2,6 +2,7 @@
 import os from 'os';
 import { E } from '@endo/far';
 import { withEndoAgent } from '../context.js';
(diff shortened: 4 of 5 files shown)


packages/cli/src/number-parse.js
```diff
@@ -0,0 +1,17 @@
...
+export const parseNumber = input => {
+  if (typeof input === 'number' && !Number.isNaN(input)) {
+    return input;
+  }
+
+  if (typeof input === 'string' && /^-?\d+(\.\d+)?$/.test(input.trim())) {
+    return Number(input);
+  }
+
+  return undefined;
```
rekmarks, line 16: Would it not be better to throw an error here?

sirtimid: I also considered throwing an error, but given that we handle input validation later on, I opted for a more lenient approach to just extract numbers from strings. I can change it if you think it's better to throw instead.

rekmarks: Hmm, I appreciate that thinking, but for the sake of consistency with [this file](https://github.com/endojs/endo/blob/kumavis-gems/packages/cli/src/pet-name.js) and on the principle that the daemon's unit tests should cover our input validation, let's just throw an error early.

rekmarks (approved):
LGTM!

packages/cli/src/number-parse.js
```diff
@@ -0,0 +1,17 @@
...
+ * @returns {number | undefined}
+ */
+export const parseNumber = input => {
+  if (typeof input === 'number' && !Number.isNaN(input)) {
+    return input;
+  }
+
+  if (typeof input === 'string' && /^-?\d+(\.\d+)?$/.test(input.trim())) {
+    return Number(input);
+  }
```
kriskowal, line 14: A thing you can rely on is that `Number(string)` will return `NaN` if the string doesn’t parse as any valid JavaScript representation of a `number`.

You can also assume `parseNumber` receives `string` and will never see a `number` as input, given that this is not a general purpose utility but specifically for parsing numbers for the CLI.

kriskowal: (Notably, this will expand the range of expressible numbers to include `1e3` without requiring any additional complexity in a regex.)

gibson042: An alternative suggestion that I especially like for CLIs, from #2204: https://github.com/endojs/endo/blob/20b4d24ae9db95dfb8073090ca0ff572868ea5a0/scripts/esbench.mjs#L110

example valid inputs:
* "42"
* "-40.0"
* ".5"
* "1.337e3"
* "␣+1␣" (_where "␣" is a visible representation of whitespace_)
* "0B111" (_binary for `7`_)
* "0o040" (_octal for `32`_)
* "0xf00" (_hexadecimal for `3840`_)

example invalid inputs (_i.e., they return `NaN`_):
* "f00"
* "NaN"
* "Infinity"
* "7up"
* "␣"
* (empty string)

sirtimid: Thanks! I also added more test cases based on @gibson042 examples. @kriskowal What is weird is that `Number('')` returns `0` instead of `NaN`

packages/cli/test/number-parse.test.js
```diff
@@ -0,0 +1,67 @@
...
+
+// Test for valid number input
+test('returns the number itself if input is a valid number', t => {
+  t.is(parseNumber(42), 42);
+  t.is(parseNumber(-42), -42);
+  t.is(parseNumber(0), 0);
+  t.is(parseNumber(3.14), 3.14);
+  t.is(parseNumber(-3.14), -3.14);
+});
+
```
gibson042, line 12: I'd drop this; the JSDoc makes clear that only string input is valid.
```suggestion
```

packages/cli/test/number-parse.test.js
```diff
@@ -0,0 +1,67 @@
...
+  t.is(parseNumber('1.337e3'), 1337);
+  t.is(parseNumber(' +1 '), 1);
+  t.is(parseNumber('0B111'), 7);
+  t.is(parseNumber('0o040'), 32);
+  t.is(parseNumber('0xf00'), 3840);
+  t.is(parseNumber('   -40.0   '), -40.0);
+  // Test for binary, octal and hexadecimal
+  t.is(parseNumber('0B111'), 7);
+  t.is(parseNumber('0o040'), 32);
+  t.is(parseNumber('0xf00'), 3840);
```
gibson042, line 27:
```suggestion
  t.is(parseNumber('0xf00'), 3840);
  t.is(parseNumber('0xF00'), 3840);
```

packages/cli/test/number-parse.test.js
```diff
@@ -0,0 +1,67 @@
...
+  t.is(error4.message, 'Invalid number: 7up');
+
+  const error5 = t.throws(() => parseNumber(' ')); // Unicode character
+  t.is(error5.message, 'Invalid number:  ');
+
+  const error6 = t.throws(() => parseNumber(''));
+  t.is(error6.message, 'Invalid number: ');
+
+  const error7 = t.throws(() => parseNumber('   '));
+  t.is(error7.message, 'Invalid number:    ');
```
gibson042, line 51:
```suggestion
  const badStrings = [
    'f00',
    'F00',
    'NaN',
    'Infinity',
    '-Infinity',
    '7up',
    ' ', // U+1680 OGHAM SPACE MARK (which is whitespace)
    '',
    '   ',
  ];
  for (const bad of badStrings) {
    t.throws(() => parseNumber(bad), { message: `Invalid number: ${bad}` });
  }
```

packages/cli/src/number-parse.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Parses the input and returns it as a number if it's valid, otherwise throws error.
+ *
+ * @param {string} input
+ * @returns {number}
+ */
+export const parseNumber = (input = '') => {
+  const result = /[0-9]/.test(input) ? Number(input) : NaN;
```
gibson042, line 8: If we care about good error messages from `parseNumber(undefined)`:
```suggestion
export const parseNumber = input => {
  const result = /[0-9]/.test(input || '') ? Number(input) : NaN;
```

packages/cli/test/number-parse.test.js
```diff
@@ -0,0 +1,67 @@
...
+
+  const error2 = t.throws(() => parseNumber(undefined));
+  t.is(error2.message, 'Invalid number: ');
+
+  const error3 = t.throws(() => parseNumber({}));
+  t.is(error3.message, 'Invalid number: [object Object]');
+
+  const error4 = t.throws(() => parseNumber([]));
+  t.is(error4.message, 'Invalid number: ');
+});
```
gibson042, line 67: I'm not sure about this either, because e.g. `['42']` is also invalid input for `parseNumber` but is _not_ rejected by the current implementation because there is no type assertion.
```suggestion
```

gibson042 (approved):

packages/cli/test/number-parse.test.js
sirtimid: Alright, I removed them in the same spirit that the function accepts only a string input

Issue #1913, opened 2023-12-23
adopt Ava default `files` glob


erights:
Repeating https://github.com/Agoric/agoric-sdk/issues/8273

## What is the Problem Being Solved?

Our `package.json` has an `"ava":` section that configures a custom `"files"` glob. There are two problems with this:
1. It requires that every package.json repeat the same custom config. Customizing tool defaults should have a high bar because of the maintenance cost they incur.
2. The naming scheme chosen puts "test" at the front of what is obviously a test from its path context. To run a particular test from the CLI requires typing "test" three time to pick one: `yarn test test/test-something.js`. With Ava's default it would be `yarn test test/something.test.js` and typing the "so" can autocomplete. 

## Description of the Design

Remove all the custom globs.
Rename test files to `.test.js`.

## Security Considerations
none

## Scaling Considerations
none
## Test Plan
CI
## Upgrade Considerations
none

erights:
This should only be fixed for endo (this issue) in coordination with fixing it for agoric-sdk (https://github.com/Agoric/agoric-sdk/issues/8273)

See https://github.com/Agoric/agoric-sdk/pull/8653#pullrequestreview-1795505142

PR #2456, opened 2024-09-16
simpler test glob

turadg:
Closes: #1913

## Description

This set out to use the Ava default file glob (i.e. no custom setting) but ultimately doesn't for two reasons:
- requires all the non-test files to have a preceding underscore (this is both laborious to complete and not a unanimously well received outcome)
- would allow test files to have any file name, including `foo.fixture.js`. We like the `.test.js` consistency

Since I did move a bunch of files to use underscores, I left that change in here, along with the script that does it. I have no qualms about removing them at reviewer request.

This also includes a little CI change that saved me time, [ci: no-bail so all packages run lint](https://github.com/endojs/endo/pull/2456/commits/8618d95a8e22b63c4d754b0c45e3ee3aa5b03907)

### Security Considerations

n/a

### Scaling Considerations

n/a

Consistent with existing practices. There is a lot of underscore now, but it's not required.

### Testing Considerations

CI

Not exported

### Upgrade Considerations

none


package.json
```diff
@@ -35,9 +35,9 @@
     "docs:markdown-for-agoric-documentation-repo": "typedoc --plugin typedoc-plugin-markdown --tsconfig tsconfig.build.json",
     "update": "lernaupdate --dedupe",
     "format": "prettier --write .github packages",
-    "lint": "prettier --check .github packages && lerna run lint",
+    "lint": "prettier --check .github packages && lerna run --no-bail lint",
     "lint-fix": "lerna run --no-bail lint-fix",
-    "test": "lerna run --ignore @endo/skel test",
+    "test": "lerna run --ignore @endo/skel --no-bail test",
```
kriskowal, line 40: `--no-opportunity-for-parole`

kriskowal (approved):
Love kicking off the legacy test path pattern. Head’s up to @boneskull with possibly-impacted work in flight.

kumavis:
some weird behavior around `--no-bail`

doc https://github.com/lerna/lerna/tree/main/libs/commands/run#--no-bail
issue https://github.com/lerna/lerna/issues/3827

kumavis:
`yarn workspaces foreach` runs all without bailing https://github.com/endojs/endo/pull/2565

PR #2460, opened 2024-09-18
fix(ses): Expose ses/console-shim

kriskowal:
SES exposes `ses/console-shim.js` in `package.json`, but lacked the “thunk” module needed to entrain the actual implementation. Being able to compose all the layers of `ses` *except* `ses/assert-shim.js` proves to be necessary to make progress on parity testing with XS #2252

### DIFF
--- a/packages/ses/console-shim.js
+++ b/packages/ses/console-shim.js
@@ -1,0 +1,1 @@
+import './src/console-shim.js';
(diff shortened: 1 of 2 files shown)


gibson042 (approved):

packages/ses/console-shim.js
mhofman, line 1: Doesn't this need an entry in `exports` in package.json ?

PR #2461, opened 2024-09-18
fix(ses): Restrict global assert

kriskowal:
SES restricts access to globals in the lexical scope of SES itself, except for their initial capture in `commons.js`, in order to avoid subsequent modifications to the global environment from damaging the integrity of SES itself. This has the side-effect of making SES itself non-reliant on the side-effects of its own shimming behavior, especially when the application applies the shim layers selectively. This is about to become important because we must omit the `ses/assert-shim.js` layer when running SES under test262, since that conflicts with test262’s own `harness/assert.js`. Capturing `assert` lexically allows SES to rely on its own `assert` API without interfering with the different global.

I’ve added all the Hardened JavaScript API, but did not add `console`. We might add `console` in a future change, but the current behavior is to defer `console` to whatever is installed globally at the time of the log line, and that can be overridden by vetted shims.

### DIFF
--- a/packages/ses/package.json
+++ b/packages/ses/package.json
@@ -118,6 +118,7 @@
     "rules": {
       "no-restricted-globals": [
         "error",
+        "AggregateError",
         "Array",
         "ArrayBuffer",
         "Atomics",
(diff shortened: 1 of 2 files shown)


gibson042 (approved):
Nice!

PR #2463, opened 2024-09-20
feat(ses,module-source): Add ModuleSource shim

kriskowal:
Refs: #2252

## Description

To test feature the degree of compatibility between a version of XS, XS with Endo shims, and Node.js with Endo shims, we need a shim for ModuleSource proper. This lets us create an environment with a global `ModuleSource`, where `ModuleSource` is a shared intrinsic of all Compartments.

This change both introduces the shim and the necessary accommodations for the existence of a global ModuleSource shared intrinsic in SES.

### Security Considerations

The new ModuleSource is subject to hardening of shared intrinsics during SES lockdown. A failure to harden the shared intrinsic would lead to a potential for interference or communication between isolated compartments.

### Scaling Considerations

None.

None.

### Testing Considerations

This includes a test for SES that verifies that ModuleSource is properly propagated and hardened.

None.

### Upgrade Considerations

None.

### DIFF
--- a/packages/module-source/src/module-source.js
+++ b/packages/module-source/src/module-source.js
@@ -100,3 +100,32 @@ export function ModuleSource(source, opts = {}) {
   this.__needsImportMeta__ = needsImportMeta;
   freeze(this);
 }
+
+// AbstractModuleSource
+// https://github.com/tc39/proposal-source-phase-imports?tab=readme-ov-file#js-module-source
+//
+// We are attempting to ensure that a JavaScript shim (particularly ses) is
+// forward-compatible as the engine evolves beneath it, with or without this
+// ModuleSource shim, and with our without a native AbstractModuleSource which
+// remains undecided.
+// Lockdown does not gracefully handle the presence of an unexpected prototype,
+// but can tolerate the absence of an expected prototype.
+// So, we are providing AbstractModuleSource since we can better tolerate the
+// various uncertain futures.
+//
+// WebAssembly and ModuleSource are both in motion.
+// The Source Phase Imports proposal implies an additional AbstractModuleSource
+// layer above the existing WebAssembly.Module that would be shared by
+// the JavaScript ModuleSource prototype chains.
+// At time of writing, no version of WebAssembly provides the shared base
+// class, and the ModuleSource *shim* gains nothing from sharing one when that
+// prototype when it comes into being.
+// So, we do not attempt to entangle our AbstractModuleSource with
+// WebAssembly.Module.
+
+function AbstractModuleSource() {
+  // no-op, safe to super()
+}
+
+Object.setPrototypeOf(ModuleSource, AbstractModuleSource);
+Object.setPrototypeOf(ModuleSource.prototype, AbstractModuleSource.prototype);
--- a/packages/ses/src/get-anonymous-intrinsics.js
+++ b/packages/ses/src/get-anonymous-intrinsics.js
@@ -161,5 +161,19 @@ export const getAnonymousIntrinsics = () => {
     );
   }
 
+  if (globalThis.ModuleSource) {
+    const AbstractModuleSourcePrototype = getPrototypeOf(
+      globalThis.ModuleSource.prototype,
+    );
+    intrinsics['%AbstractModuleSourcePrototype%'] =
+      AbstractModuleSourcePrototype;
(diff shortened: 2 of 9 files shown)


packages/ses/src/get-anonymous-intrinsics.js
```diff
@@ -161,5 +161,19 @@ export const getAnonymousIntrinsics = () => {
...
 
+  if (globalThis.ModuleSource) {
+    const AbstractModuleSourcePrototype = getPrototypeOf(
+      globalThis.ModuleSource.prototype,
+    );
+    intrinsics['%AbstractModuleSourcePrototype%'] =
+      AbstractModuleSourcePrototype;
+    intrinsics['%AbstractModuleSource%'] =
+      AbstractModuleSourcePrototype.constructor;
+  }
```
kriskowal, line 172: A valid but annoying feedback reviewers might provide here is that we can’t anticipate whether `ModuleSource` will land in the language with or without an `AbstractModuleSource` on its prototype chain, so we could hedge our bets and add a repair for ModuleSource to force it to appear one way or the other, so that `lockdown()` doesn’t break if the AbstractModuleSource is absent. Or, we could go the other way and _not_ have AbstractModuleSource by default, in which case SES would just delete it and issue a warning if it showed up.

gibson042: You read my mind, but that's just the risk with front-running proposals. No suggestion.

packages/module-source/src/module-source.js
```diff
@@ -100,3 +113,5 @@ export function ModuleSource(source, opts = {}) {
   this.__needsImportMeta__ = needsImportMeta;
   freeze(this);
 }
+
+Object.setPrototypeOf(ModuleSource.prototype, AbstractModuleSource.prototype);
```
gibson042, line 117: Doesn't this approach also require setting the prototype of `ModuleSource` itself to `AbstractModuleSource`?
```js
class B { static name = "B"; }
class A extends B { static name = "A"; }
assert.equal(Object.getPrototypeOf(A.prototype), B.prototype);
assert.equal(Object.getPrototypeOf(A), B);
```

or are we anticipating an unusual relationship between the two?

gibson042 (approved):

packages/module-source/src/module-source.js
kriskowal: Negative. Pushed changes including tests for the property and prototype subgraph shape and hardening.

packages/ses/src/get-anonymous-intrinsics.js
kriskowal: Capturing here my reasoning for keeping AbstractModuleSource since it appears to be the path of least regret.

1. We shim AbstractModuleSource and include it in SES permits:
    1. The application includes module-source/shim.js:
        1. The VM implements AbstractModuleSource: :white_check_mark: Lockdown’s fine
        2. The VM does not implement AbstractModuleSource: :white_check_mark: Lockdown’s fine. The shim currently stomps the native implementation. If it didn’t stomp the native implementation, Lockdown would tolerate the incomplete intrinsics.
    2. The application excludes module-source/shim.js:
        1. The VM implements AbstractModuleSource: :white_check_mark: Lockdown’s fine
        2. The VM does not implement AbstractModuleSource: :white_check_mark: Lockdown’s fine, but intrinsics are incomplete
2. We do not shim AbstractModuleSource and exclude it in SES permits:
    1. The application includes module-source/shim.js:
        1. The VM implements AbstractModuleSource: :white_check_mark: :man-shrugging: The shim currently stomps the native implementation, so this works out. But, if the shim surfaced the native implementation (and perhaps feature-checks that the native implementation behaves just like the shim deigning to fall through to native)
        2. The VM does not implement AbstractModuleSource: :white_check_mark: Lockdown’s fine
    2. The application excludes module-source/shim.js:
        1. The VM implements AbstractModuleSource: :broken_heart: Lockdown is surprised to find AbstractModuleSource and chooses throw rather than attempt a repair. We don’t have the module source shim in play, so there’s nothing it can do to mitigate. We can’t require the module-source/shim.js because it entrains Babel.
        2. The VM does not implement AbstractModuleSource: :white_check_mark: Lockdown’s fine

gibson042 (approved):

PR #2465, opened 2024-09-20
test: Introduce XS CI workflow

kriskowal:
Refs: #2254

## Description

Rigs CI to run tests with `xst`. We can elaborate on this with a broader XS version vector and add XS tests to packages. This is in anticipation of a new harness262 package that will need `xst` in CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       # without this, setup-node errors on mismatched yarn versions
       - run: corepack enable
@@ -75,7 +75,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       # without this, setup-node errors on mismatched yarn versions
       - run: corepack enable
@@ -121,7 +121,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       # without this, setup-node errors on mismatched yarn versions
       - run: corepack enable
@@ -158,7 +158,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       # without this, setup-node errors on mismatched yarn versions
       - run: corepack enable
@@ -194,7 +194,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       # without this, setup-node errors on mismatched yarn versions
       - run: corepack enable
@@ -230,7 +230,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       # without this, setup-node errors on mismatched yarn versions
       - run: corepack enable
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -260,3 +260,106 @@ jobs:
...
+
+      - name: Install dependencies
+        run: yarn install --immutable
+
+      # end macro
+
+      - name: Run yarn build
+        run: yarn build
+
+      - name: Get XS
```
mhofman, line 298: Here's an idea:

```suggestion
      - name: Check release
        id: check-release
        run: |
          if curl -f -s -o /dev/null -I -L https://api.github.com/repos/Moddable-OpenSource/moddable/releases/tags/${{ matrix.moddable-version }} then
              echo "release=true"
          else
              echo "release=false"
          fi >> $GITHUB_OUTPUT
      
      - name: Get XS
        if: steps.check-release.outputs.release == 'true'
```

With the build step if it's not.

.github/workflows/ci.yml
```diff
@@ -260,3 +260,106 @@ jobs:
...
+  #     - name: Install dependencies
+  #       run: yarn install --immutable
+  #
+  #     # end macro
+  #
+  #     - name: Run yarn build
+  #       run: yarn build
+  #
+  #     - name: Clone XS
+  #       run: git clone https://github.com/moddable-OpenSource/moddable --branch "${{ matrix.moddable-version }}" --depth 1
```
mhofman, line 351: I'd recommend using `actions/checkout`, also avoid to checkout as a subfolder of the repo under testing, it tends to complicate things

kriskowal:
@mhofman I took your cue and reworked `test-xs` so it works with both versions or hashes and uses a precompiled release for the tag if one exists. I ran a CI job to verify both paths worked and settled on just the 5.0.0 release for now. I expect to update the Moddable version matrix to capture different versions in the fullness of time.

.github/workflows/ci.yml
```diff
@@ -260,3 +260,77 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        node-version: [20.x]
+        # may include release tags and hashes:
+        moddable-version: [5.0.0]
+        platform: [ubuntu-latest]
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
michaelfig, line 280: I'm not sure if you were aware of this, but `actions/checkout@v4` has the following default for `path':

```yaml
    # Relative path under $GITHUB_WORKSPACE to place the repository
    path: ''
```

So this means the checkouts you do later with `path: moddable` are checking out to `$GITHUB_WORKSPACE/moddable`.

If you meant for the endo and moddable checkouts to be siblings (as I understand @mhofman was suggesting) instead of `moddable` being a subdirectory of the `endo` checkout, you need something like:

```suggestion
        uses: actions/checkout@v4
        path: endo
```

I'll make further suggestions for changes along that line.  If the status quo is sufficient, feel free to ignore those suggestions.

.github/workflows/ci.yml
```diff
@@ -260,3 +260,77 @@ jobs:
...
+      - run: corepack enable
+
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: yarn
+
+      - name: Install dependencies
+        run: yarn install --immutable
```
michaelfig, line 292:
```suggestion
      - name: Install dependencies
        run: yarn install --immutable
        working-directory: endo
```

.github/workflows/ci.yml
```diff
@@ -260,3 +260,77 @@ jobs:
...
+
+      # end macro
+
+      - name: Run yarn build
+        run: yarn build
```
michaelfig, line 297:
```suggestion
      - name: Run yarn build
        run: yarn build
        working-directory: endo
```

.github/workflows/ci.yml
```diff
@@ -260,3 +260,77 @@ jobs:
...
+          if curl -f -s -o /dev/null -I -L https://api.github.com/repos/Moddable-OpenSource/moddable/releases/tags/${{ matrix.moddable-version }}
+          then
+            echo release=download
+          else
+            echo release=build
+          fi >> $GITHUB_OUTPUT
+
+      - name: Download XS
+        if: steps.check-release.outputs.release == 'download'
+        working-directory: ..
```
michaelfig, line 311:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -260,3 +260,77 @@ jobs:
...
+      - name: Build XS
+        if: steps.check-release.outputs.release == 'build'
+        working-directory: moddable/xs/makefiles/lin
+        run: |
+          make debug MODDABLE=$GITHUB_WORKSPACE/moddable CC='cc "-D__has_builtin(x)=1"' # give the syntax highlighter a hand: '
+
+      - name: Run XS tests
+        run: |
+          PATH=$PATH:$GITHUB_WORKSPACE/moddable/build/bin/lin/debug
+          yarn lerna run test:xs
```
michaelfig, line 336:
```suggestion
      - name: Run XS tests
        run: |
          PATH=$PATH:$GITHUB_WORKSPACE/moddable/build/bin/lin/debug
          yarn lerna run test:xs
        working-directory: endo
```

.github/workflows/ci.yml
```diff
@@ -260,3 +260,77 @@ jobs:
...
+        with:
+          repository: moddable-OpenSource/moddable
+          ref: ${{ matrix.moddable-version }}
+          path: moddable
+
+      - name: Build XS
+        if: steps.check-release.outputs.release == 'build'
+        working-directory: moddable/xs/makefiles/lin
+        run: |
+          make debug MODDABLE=$GITHUB_WORKSPACE/moddable CC='cc "-D__has_builtin(x)=1"' # give the syntax highlighter a hand: '
```
michaelfig, line 331:
```suggestion
          make debug MODDABLE=$GITHUB_WORKSPACE/moddable CC='cc "-D__has_builtin(x)=1"' # here's a nickle, kid... go get yourself a better syntax highlighter
```

michaelfig (approved):
Nice to see some Endo tests getting run by XS!  This'll be good to land.

I'm approving conditionally on you deciding to confirm or deny my suggested changes.  From that point a passing CI is all I'm looking for.

mhofman (approved):
Approved in principle, I'll let you decide whether to entertain my further suggestions.

.github/workflows/ci.yml
kriskowal: I see.

.github/workflows/ci.yml
```diff
@@ -260,3 +260,77 @@ jobs:
...
+            echo release=download
+          else
+            echo release=build
+          fi >> $GITHUB_OUTPUT
+
+      - name: Download XS
+        if: steps.check-release.outputs.release == 'download'
+        working-directory: ..
+        run: |
+          BINDIR=$GITHUB_WORKSPACE/moddable/build/bin/lin/debug
```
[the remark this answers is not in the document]
kriskowal: I considered both and flipped a coin. But, so long as I’m making `endo` and `moddable` peers under the workspace, I may as well make a `bin` peer.

.github/workflows/ci.yml
```diff
@@ -260,3 +260,77 @@ jobs:
...
+      - name: Install dependencies
+        run: yarn install --immutable
+
+      # end macro
+
+      - name: Run yarn build
+        run: yarn build
+
+      - name: Choose download or build
+        id: check-release
```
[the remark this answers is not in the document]
kriskowal: Very much entertaining suggestions. If you have a suggestion that would parameterize the architecture, that’ll make it easier to turn on darwin and the holes-in-sides-of-houses variants in CI maybe someday.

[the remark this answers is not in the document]
mhofman: [There are](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#runner-context), but not sure which are effectively needed, since GH cache is supposed to be different by default based on the OS (see [`enableCrossOsArchive`](https://github.com/actions/cache?tab=readme-ov-file#inputs)). Here is an [example in agoric-sdk](https://github.com/Agoric/agoric-sdk/blob/fd12d05b0f2242236e8526c4a4f4c3496b9c2e06/.github/actions/restore-node/action.yml#L134):
```
${{ runner.os }}-${{ runner.arch }}
```

A relevant PR/commit might be https://github.com/Agoric/agoric-sdk/pull/5512/commits/c9222a9d3c01b96c12d5b27c602e5c6a7ea088fc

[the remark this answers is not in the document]
kriskowal: For the record, I’ve got a good idea of how to add platforms to the matrix and am electing not to at this juncture.

.github/workflows/ci.yml
kriskowal: And that’s the difference between giving your 2¢ and giving your 5¢.

kriskowal:
@leotm This can be used as a template for making a CI job that will run `test:hermes` scripts in CI.

PR #2467, opened 2024-09-23
remove ts-api-utils

turadg:
hygiene

## Description

`ts-api-utils` was pinned to save consumers from upgrading to Node 16. 
- https://github.com/endojs/endo/pull/2040

It's been long enough that we can drop that concession.

### Security Considerations

none
### Scaling Considerations

none

none

### Testing Considerations

none

Endo devDependencies now really requires Node 16.

### Upgrade Considerations

TMK does not yet affect exports.


michaelfig (approved):
LGTM! :shipit:

PR #2469, opened 2024-09-24
feat(ses): Tame ModuleSource

kriskowal:
Refs: #2463, #2252

## Description

In #2463 we introduced ModuleSource to the SES permits, but this interacted catastrophically with the native XS ModuleSource. We reverted this change #2468 to unbreak Agoric SDK integration. This change reintroduces the ModuleSource permits, such that they are compatible with both XS and the `@endo/module-source/shim.js`, which anticipates the introduction of an AbstractModuleSource base class. Because SES can more gracefully tolerate the absence of an permitted [[Proto]] than the presence of a non-permitted [[Proto]], this adjusts the shim to ensure that the AbstractModuleSource shape exists as a side-effect of repairs/taming, before permits are applied.

### Security Considerations

None.

### Scaling Considerations

None.

This change reintroduces a note in NEWS.md for the next release.

### Testing Considerations

The prior regression went unnoticed because we did not yet have CI for XS #2465. With this change, `yarn test:xs` in SES validates the shim on XS. We also test `@endo/module-source/shim.js` in `ses/test/module-source.test.js` on Node.js.

This change is designed to tolerate either path forward for the language, whether or not it accepts an AbstractModuleSource base class.

### Upgrade Considerations

None.

### DIFF
--- a/packages/ses/src/error/console.js
+++ b/packages/ses/src/error/console.js
@@ -368,8 +368,10 @@ export const makeCausalConsole = (baseConsole, loggedErrorHandler) => {
     const levelMethod = (...logArgs) => {
       const subErrors = [];
       const argTags = extractErrorArgs(logArgs, subErrors);
-      // eslint-disable-next-line @endo/no-polymorphic-call
-      baseConsole[level](...argTags);
+      if (baseConsole[level]) {
+        // eslint-disable-next-line @endo/no-polymorphic-call
+        baseConsole[level](...argTags);
+      }
       // @ts-expect-error ConsoleProp vs LogSeverity mismatch
       logSubErrors(level, subErrors);
     };
@@ -476,12 +478,16 @@ export const defineCausalConsoleFromLogger = loggedErrorHandler => {
           // the host console will separate them with additional spaces.
           arrayPush(indents, ' ');
         });
+      } else {
+        baseConsole[name] = () => {};
       }
     }
     if (baseConsole.groupEnd) {
       baseConsole.groupEnd = makeNamed('groupEnd', (...args) => {
         arrayPop(indents);
       });
+    } else {
+      baseConsole.groupEnd = () => {};
     }
     harden(baseConsole);
     const causalConsole = makeCausalConsole(
--- a/packages/ses/src/permits.js
+++ b/packages/ses/src/permits.js
@@ -101,8 +101,12 @@ export const universalPropertyNames = {
 
   // ESNext
 
+  // https://github.com/tc39/proposal-source-phase-imports?tab=readme-ov-file#js-module-source
+  ModuleSource: 'ModuleSource',
+
   lockdown: 'lockdown',
   harden: 'harden',
+
   HandledPromise: 'HandledPromise', // TODO: Until Promise.delegate (see below).
 };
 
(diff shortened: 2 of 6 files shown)


kriskowal:
Agoric SDK integration probe https://github.com/Agoric/agoric-sdk/pull/10144

packages/ses/src/permits.js
```diff
@@ -1505,6 +1509,29 @@ export const permitted = {
...
 
+  // https://github.com/tc39/proposal-source-phase-imports?tab=readme-ov-file#js-module-source
+  '%AbstractModuleSourcePrototype%': {
+    constructor: '%AbstractModuleSource%',
+    '@@toStringTag': 'string',
+  },
+  '%AbstractModuleSource%': {
+    '[[Proto]]': '%FunctionPrototype%',
+    prototype: '%AbstractModuleSourcePrototype%',
+  },
```
erights, line 1520:
```suggestion
  '%AbstractModuleSource%': {
    '[[Proto]]': '%FunctionPrototype%',
    prototype: '%AbstractModuleSourcePrototype%',
  },

    '%AbstractModuleSourcePrototype%': {
    constructor: '%AbstractModuleSource%',
    '@@toStringTag': 'string',
  },

```

Please maintain the general pattern of the rest of the permits, where the `Foo` constructor comes before the `%FooPrototype%` prototype, and there is a blank line separating each permit.

erights: Likewise with `ModuleSource` below of course.

erights: Why does `%AbstractModuleSourcePrototype%` have a `@@toStringTag`? Is there precedent for other abstract superclasses having one, even though we always expect it to be overridden by concrete subclasses?

erights (commented):
One thing I don't understand. `lockdown` calls the repair, to repair things once we have a `ModuleSource`, whether natively or via the shim. But what brings in the shim? I expected `lockdown` to do that as well.

erights:
Why is everything failing under CI?

packages/ses/src/error/console.js
```diff
@@ -369,7 +369,9 @@ export const makeCausalConsole = (baseConsole, loggedErrorHandler) => {
       const subErrors = [];
       const argTags = extractErrorArgs(logArgs, subErrors);
       // eslint-disable-next-line @endo/no-polymorphic-call
-      baseConsole[level](...argTags);
+      if (baseConsole[level]) {
+        baseConsole[level](...argTags);
+      }
```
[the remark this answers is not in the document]
kriskowal: In the course of these changes, I added some XS-specific properties of ModuleSource.prototype to the permits, but without those permits, SES employs console.groupCollapsed() and console.endGroup() which are not present on the base console under XS. We now `test:xs` in CI.

packages/ses/src/permits.js
kriskowal: The `@@toStringTag` is my error. There is no such property even on my shim. There is not precedent for abstract base classes among the intrinsics. No way to know yet whether the spec will land with one and it’s easy to add later.

kriskowal:
I’ve restructured this change around addIntrinsics and tameModuleSource instead of any code in getAnonymousIntrinsics. I learned that we need to bind `globalThis.ModuleSource` at time of lockdown or we get too picky about whether the ModuleSource shim initializes before or after SES. This is ready for another review.

erights (approved):
LGTM!

PR #2471, opened 2024-09-27
feat(ses): Add XS variant of shim

kriskowal:
Closes: #2251

## Description

To take advantage of native XS compartments while retaining backward compatibility and parity with the SES shim, we introduce an `xs` specific Compartment adapter that requires two levels of opt-in.
1. The SES shim must be bundled with the `xs` package export/import condition.
2. The constructor of a Compartment must specify the `__native__` option to sacrifice the ability to use precompiled module sources (as generated by `@endo/module-source` _without_ the `xs` package export/import condition) and instead use adapted native `ModuleSource` (as generated by `@endo/module-source` _with_ the `xs` package export/import condition). This allows `@endo/import-bundle`, for example, to use the JSON serialization of a precompiled module source that is captured in a bundle and also opt-in for `__native__` treatment if the archive/bundle contains original sources.

This change introduces an XS-specific shim that is an adapter for `Compartment` and `lockdown`, and also papers over parity gaps like the XS `Object.freeze` second boolean argument. The adapter creates parallel native and shim (virtual) compartment trees, where any individual Compartment can elect to use the native or shim variant for a child Compartment.

We have not yet found a workable design that obviates the need for the `__native__` opt-in. Such a design would need to create an adapter from precompiled module sources to XS’s virtual module source protocol. To do that would require native module to emit notifications for the mutation of exported live bindings and also require the native Compartment evaluate method to accept an argument like the shim’s `__moduleGlobalLexicals__`.

### Security Considerations

Uncountably numerous. Among them, with the `__native__` option, censorship does not occur, so dynamic import and direct eval are possible.

### Scaling Considerations

The native ModuleSource makes it practical to defer module parsing to runtime, and should improve the performance of execution as well.

The NEWS.md qualifies these changes as under "incubation". When the shape of these changes settles, the NEWS will need to reiterate the final user facing API in README.md and NEWS.md. With the `__native__` option, censorship does not occur, so dynamic import and direct eval are possible.

### Testing Considerations

This change contains a token of `xst` testing that is exercised in CI with `test:xs`. This demonstrates the use of `@endo/compartment-mapper/bundle.js` to thread the `xs` package export/import condition and generate a script that can `xst` can run directly. This gives us some modest confidence that lockdown works and demonstrates the `__native__` feature but does not provide sufficient confidence of parity for the gamut of Compartment usage, both legacy and XS, for all of the accepted module descriptors and other Compartment features. This is an exercise that will begin with a subsequent change that introduces `hardened262`, a comprehensive parity checking framework for the full cross-product of [ SES on Node.js, SES on XS, and XS stand-alone ] ⨉ [ Lockdown, not Lockdown ] ⨉ [ Compartment, no Compartment ] ⨉ [ Sloppy, Strict, Module ].

This change preserves all existing usage and introduces an unstable alternate version of SES for XS that requires two layers of opt-in. The use of the `xs` condition introduces an adapter for `Compartment` that may not have full parity with the underlying implementation, and requires additional testing. The `__native__` option elects to break some usage (precompiled moduels) in favor of others (dynamic import, direct eval, top-level-await).

### Upgrade Considerations

In order to realize these changes on the Agoric chain will likely require a more recent version of XS and switching the bundle format for the lockdown/bootstrap script for xsnap swingset workers.


packages/ses/src-xs/index.js
```diff
@@ -0,0 +1,463 @@
+// @ts-nocheck
```
kriskowal, line 1: Github is hiding this file. This is the bulk of the review.

packages/ses/src-xs/index.js
```diff
@@ -0,0 +1,463 @@
...
+      ? fromEntries(
+          arrayMap(
+            entries(options.modules ?? {}),
+            ([specifier, descriptor]) => [
+              specifier,
+              adaptModuleDescriptor(descriptor, specifier, undefined),
+            ],
+          ),
+        )
+      : {};
```
kriskowal, line 331: Add comment explaining that this `module` is only used for `nativeOptions` and the shim options pass `options.modules` thru unaltered. Arguably, the `{}` at the end could be anything, but `{}` satisfies the type.

packages/import-bundle/NEWS.md
```diff
@@ -1,5 +1,16 @@
...
+# Next release
+
+Experimental:
+
+- Adds a `__native__: true` option that indicates that the application will
+  fall through to the native implementation of Compartment, currently only
+  available on XS, which lacks support for precompiled module sources (as exist
+  in many archived applications, particularly Agoric smart contract bundles)
+  and instead supports loading modules from original sources (which is not
+  possible at runtime on XS).
```
kriskowal, line 12: An alternative design that I am interested in considering is to instead create a new `moduleFormat` called `endoNativeZipBase64` that implies this `__native__` flag. Whether it can be overridden is worth consideration. This would imply a change in `bundle-source` such that the existing `--no-transforms` or a new `--native` flags would switch the `moduleFormat`. We can create bundles with `-T,--no-transforms` today that cannot be imported without this `__native__` flag on XS, but once this PR lands, can be imported on Node.js.

packages/compartment-mapper/NEWS.md
```diff
@@ -5,13 +5,43 @@ User-visible changes to `@endo/compartment-mapper`:
...
 
+Experimental:
+
+- Adds a `__native__: true` option to all paths to import, that indicates that
+  the application will fall through to the native implementation of
+  Compartment, currently only available on XS, which lacks support for
+  precompiled module sources (as exist in many archived applications,
+  particularly Agoric smart contract bundles) and instead supports loading
+  modules from original sources (which is not possible at runtime on XS).
+
```
erights, line 44: Did you mean to include this under "v1.3.0" rather than "Next version"?

packages/import-bundle/NEWS.md
erights: I don't understand

> loading modules from original sources (which is not possible at runtime on XS)

What about this is not possible on XS?

packages/ses/NEWS.md
```diff
@@ -36,8 +36,18 @@ User-visible changes in `ses`:
...
+and subject to breaking changes that will not be signaled by semver.
+
 - Adds permits for `ModuleSource`, either the native implementation or from
   `@endo/module-source/shim.js`.
+- Adds support for an XS-specific variant of the SES shim that is triggered
+  with the `xs` package export condition.
+  This version of SES preserves all the features of `Compartment` provided
+  uniquely by the SES shim, but with the `__native__` constructor option,
+  loses support for importing precompiled module records and gains support
+  for native `ModuleSource`.
```
erights, line 50: If you truly want to add/change an entry in an older NEWS section, could you add something to the "Next release" section to make clear what is new?

packages/compartment-mapper/src/bundle.js
```diff
@@ -113,8 +113,14 @@ const sortedModules = (
...
+        bytes !== undefined,
+        `No bytes for ${moduleSpecifier} in ${compartmentName}`,
+      );
+      assert(
+        parser !== undefined,
+        `No parser for ${moduleSpecifier} in ${compartmentName}`,
+      );
       if (deferredError) {
         throw Error(
           `Cannot bundle: encountered deferredError ${deferredError}`,
```
erights, line 126: Keep the errors connected re causal console log output

```suggestion
        Fail`Cannot bundle: encountered deferredError ${deferredError}`,
```

Also, consider for other asserts. Unless we're trying to avoid our assertion support and style in this package for some reason. Are we? Why?

erights (commented):
Review so far

packages/ses/src-xs/index.js
```diff
@@ -0,0 +1,463 @@
...
+/// <refs types="../types.js"/>
+
+import '../src/assert-shim.js';
+import '../src/console-shim.js';
+import { repairIntrinsics } from '../src/lockdown.js';
+
+import {
+  makeCompartmentConstructor as makeShimCompartmentConstructor,
+  compartmentOptions,
+} from '../src/compartment.js';
```
erights, line 11: Minor:

This file is intended to be an xs-flavored alternative to ses/index.js. Two minor changes would help a reader see and understand the parallelism:
- Move ses/index.js to ses/src/index.js, leaving behind a forwarding stub if appropriate.
- Rearrange the import order in that file to
   ```js
    import './assert-shim.js';
    import './console-shim.js';
    import './lockdown-shim.js';
    import './compartment-shim.js';
   ```
   so the reader doesn't need to worry about whether the difference in import order is significant.

erights: Might it even make sense to break this file up into a `ses/src-xs/lockdown-shim.js` and `ses/src-xs/compartment-shim.js`, and then change `ses/src-xs/index.js` to be that much more similar?

packages/ses/src-xs/index.js
```diff
@@ -0,0 +1,463 @@
...
+globalThis.Compartment = FERAL_EVAL(compartmentShim)(
+  compartmentShim,
+  commons,
+  NativeCompartment,
+  ShimCompartment,
+  noHarden,
+);
+
+globalThis.lockdown = options => {
+  const hardenIntrinsics = repairIntrinsics(options);
```
erights, line 445: I notice that this PR does not modify `lockdown.js` which defines the `repairIntrinsics` being called here. Nor does it modify `tame-harden.js` or `make-hardener.js` used by `repairIntrinsics` to make the `harden` function. But presumably, on XS we want to use only the XS native `harden`. How does that happen?

packages/ses/src-xs/index.js
```diff
@@ -0,0 +1,463 @@
...
+  hardenIntrinsics();
+  // Replace global Compartment with a version that is hardened and hardens
+  // transitive child Compartment.
+  globalThis.Compartment = FERAL_EVAL(compartmentShim)(
```
erights, line 449: This both reevaluates `compartmentShim` and reapplies its completion value, which is the big function defined by inline text above, to similar arguments, with the main (only?) difference being `noHarden` vs `harden`. While this is all fine, so no change necessarily suggested, it does make me wonder whether we could evaluate the function only once and have all needed differences come only from the application arguments?

erights (commented):
Review complete on everything but src-xs/index.js

This looks good to me in broad outline, but I have not yet found the time to review it in detail. @gibson042 , can you complete this? If you approve this one file, I'm willing to approve the whole thing. Thanks!

packages/ses/src-xs/index.js
kriskowal: I’ve posted a change that eliminates the eval entirely. It’s now just a function `adaptCompartmentConstructors(NativeCompartmentConstructor, ShimCompartmentConstructor, maybeHarden)`. I look forward to possibly remembering why the eval was necessary, but it probably wasn’t actually.

packages/ses/src-xs/index.js
kriskowal: We use native harden due to an existing fall through to globalThis.harden in `makeHarden`. We also continue to respect the lockdown option to make `harden` a no-op. I’ve elected to continue using our `lockdown` variant so-far, since it is parameterized, can conduct repairs to the XS intrinsics if any become necessary, and support enablements. I’m at least leaving native lockdown out of scope for this run at native Compartment and ModuleSource.

packages/ses/src-xs/index.js
kriskowal: I’ve taken this advice. You will find more files in `src-xs` on your next review.

packages/ses/NEWS.md
kriskowal: Unintentional: moved up.

packages/import-bundle/NEWS.md
kriskowal: SES on XS, both before this change and after this change (but without the `__native__` option set), cannot load a module from original sources without the assistance of the _optional_ `ModuleSource` shim.

In any case, this prose needs some work, but I’ve also moved this commit out of this PR since I intend to investigate the alternate design using an `"endoNativeZipBase64"` format. The MetaMask folks reminded me that the web will still benefit from pre-compiled bundles, so those should remain the default and we should refrain from changing the default even with a major version bump on `bundle-source`.

packages/ses/src-xs/compartment-shim.js
```diff
@@ -0,0 +1,128 @@
...
+const adapterFunctions = {
+  evaluate(source, options) {
+    const fields = weakmapGet(privateFields, this);
+    if (fields === undefined) {
+      return nativeEvaluate(source);
+    }
+    const { transforms, delegateNative, nativeEval } = fields;
+    for (const transform of transforms) {
+      source = transform(source);
+    }
```
gibson042, line 38: Do the imports from commons.js mean that this code needs to be robust against a tampered environment? If so, then it probably shouldn't use `of`.
```suggestion
    for (let i = 0; i < transforms.length; i += 1) {
      source = transforms[i](source);
    }
```

packages/ses/src-xs/compartment.js
```diff
@@ -0,0 +1,284 @@
...
+      moduleMapHook = () => {},
+      __native__: delegateNative = false,
+      __noNamespaceBox__: noNamespaceBox = false,
+    } = options;
+
+    const modules = delegateNative
+      ? fromEntries(
+          arrayMap(
+            entries(options.modules ?? {}),
+            ([specifier, descriptor]) => [
```
gibson042, line 179: Do the imports from commons.js mean that this code needs to be robust against a tampered environment? If so, then
```suggestion
            ({ 0: specifier, 1: descriptor }) => [
```

packages/ses/src-xs/compartment.js
```diff
@@ -0,0 +1,284 @@
...
+  };
+};
+
+const adaptModuleSource = source => {
+  if (source.execute) {
+    return adaptVirtualModuleSource(source);
+  }
+  // eslint-disable-next-line no-underscore-dangle
+  if (source.__syncModuleProgram__) {
+    throw new SyntaxError(
```
gibson042, line 87: I'm pretty sure @erights prefers universal avoidance of `new` with built-in error constructors.

packages/ses/src-xs/index.js
```diff
@@ -0,0 +1,26 @@
...
+import './compartment-shim.js';
+import '../src/assert-shim.js';
+import '../src/console-shim.js';
+
+// XS Object.freeze takes a second argument to apply freeze transitively, but
+// with slightly different effects than `harden`.
+// We disable this behavior to encourage use of `harden` for portable Hardened
+// JavaScript.
+/** @param {object} object */
+Object.freeze = object => freeze(object);
```
gibson042, line 26: Let's maintain `Object.freeze.name`.
```suggestion
Object.freeze = ({ freeze: object => freeze(object) }).freeze;
```

packages/ses/src-xs/lockdown-shim.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * @module Alters the XS implementation of Lockdown to be backward compatible
```
gibson042, line 2:
```suggestion
 * @module Alters the XS implementation of lockdown to be backward compatible
```

packages/ses/src-xs/lockdown-shim.js
```diff
@@ -0,0 +1,26 @@
...
+  const hardenIntrinsics = repairIntrinsics(options);
+  hardenIntrinsics();
+  // Replace global Compartment with a version that is hardened and hardens
+  // transitive child Compartment.
+  globalThis.Compartment = adaptCompartmentConstructors(
+    NativeStartCompartment,
+    ShimStartCompartment,
+    harden,
+  );
+};
```
gibson042, line 26:
```suggestion
const lockdown = options => {
  const hardenIntrinsics = repairIntrinsics(options);
  hardenIntrinsics();
  // Replace global Compartment with a version that is hardened and hardens
  // transitive child Compartment.
  globalThis.Compartment = adaptCompartmentConstructors(
    NativeStartCompartment,
    ShimStartCompartment,
    harden,
  );
};
globalThis.lockdown = lockdown;
```

gibson042 (commented):
This seems nicely threaded to me.

gibson042 (approved):

packages/ses/src-xs/commons.js
```diff
@@ -0,0 +1,22 @@
...
+ * are free to modify any intrinsic without risking the integrity of SES.
+ */
+
+import {
+  getOwnPropertyDescriptor,
+  globalThis,
+  uncurryThis,
+} from '../src/commons.js';
+
+export const NativeStartCompartment = globalThis.Compartment;
```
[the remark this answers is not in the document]
kriskowal: Making the Start distinction avoids a shadowing problem elsewhere, when a child compartment has its own constructors.

PR #2473, opened 2024-09-27
(experiment) feat(gems): gems goes multivat

kumavis:
highlight: **durable** **lazy** presences for remote objects via `ExtRefController` that can lazily trigger their underlying captp connection upon use

requires captp external table support: https://github.com/endojs/endo/pull/2564

see relevant test:  https://github.com/endojs/endo/blob/f1ea4a3a25b6363b3505037f51d7b3e717893735/packages/gems/test/extref.test.js#L22-L32


kumavis:
https://github.com/endojs/endo/pull/2449

PR #2477, opened 2024-10-01
chore(ci): Drop matrix for lint job

### LINKED ISSUE #2283: drop matrix from the lint job

The `lint` job does static analysis, but it's configured to run against multiple Node runtimes. Only one is needed and it shouldn't have a matrix because that necessitates updating its Required check when bumping Node versions.

## Description of the Design

1. Take the matrix out of https://github.com/endojs/endo/blob/b02f0e278c9fb19515cbab58635e6dbead2eb9bc/.github/workflows/ci.yml#L18-L28

2. A repo admin updates settings to require the plain "lint" job, not the runtime-scoped one

kriskowal:
Fixes #2283

In order to avoid unnecessary duplication of effort in CI and to keep the required Lint job name consistent as we upgrade the underlying Node.js version, this change drops the test matrix and hard-codes the platform and node version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,12 +19,9 @@ jobs:
 
     # begin macro
 
-    runs-on: ${{ matrix.platform }}
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
-      matrix:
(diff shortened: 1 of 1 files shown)


turadg (approved):
😃

PR #2478, opened 2024-10-01
configure dependabot

turadg:
incidental

## Description

Per https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates

Dependabot doesn't support wildcards so the file is generated with a script that adds a section for each package.json

### Security Considerations

Entrusts Github to modify package.json. PRs will be reviewed by humans. It's automating what we already do.

### Scaling Considerations

none

It's a new process. It's well documented by GH tho

### Testing Considerations

Let's just try it out and see how it works for us.

n/a

### Upgrade Considerations

There should be no downstream effects


kriskowal:
Let’s give it a shot. I’m hoping for a single weekly PR that covers the whole repository, but I’m guessing this will generate one for each package.

kriskowal (approved):

PR #2558, opened 2024-10-01
chore(dependabot): group minor and patch updates

turadg:
incidental

## Description

https://github.com/endojs/endo/pull/2478 yielded pages of new PRs: https://github.com/endojs/endo/pulls?page=1&q=is%3Apr+is%3Aopen+label%3Adependencies

This tries the `group` option to so they will be more consolidated. But I think each package will still get its own PR because each package has a different `updates` entry.

### Security Considerations

none
### Scaling Considerations

none

none

### Testing Considerations

Not sure how to test this other than merge and see what happens

none

### Upgrade Considerations

none


kriskowal (commented):
Worth a shot. I suspect that we can just dependabot / to get lockfile updates consolidated in a one-PR-per-week and one-concurrent-PR-ever arrangement.

kriskowal (approved):

PR #2334, opened 2024-06-25
feat(ses): Shim compatible with Hermes compiler

leotm:
## Description

- Refs: https://github.com/endojs/endo/issues/1891
  - SES shim compatible with Hermes compiler 
  - Fixes: https://github.com/endojs/endo/issues/2474

<s>Add lockdown shim compatible with Hermes</s>
Make current shim compatible with Hermes compiler
for building React Native (RN) prod apps with SES

Generating the release AAB ([Android App Bundle](https://reactnative.dev/docs/signed-apk-android#generating-the-release-aab))
i.e. `npx react-native build-android --mode=release` via RN CLI
calls Gradle's `bundleRelease` task under the hood (`bundle` build task on `release` variant)
which calls RNGP ([React Native Gradle Plugin](https://reactnative.dev/docs/react-native-gradle-plugin)) React task [`createBundleReleaseJsAndAssets`](https://github.com/facebook/react-native/blob/3af01a8a44853246d3deb957c4a1c65e43848bb5/packages/gradle-plugin/react-native-gradle-plugin/src/main/kotlin/com/facebook/react/TaskConfiguration.kt#L58) and fails
after Metro [finishes](https://github.com/facebook/react-native/blob/c25246044f51b770ff1e090d6b4ec78d41ec917f/packages/community-cli-plugin/src/commands/bundle/saveAssets.js#L109) writing the release bundle (bundle, sourcemaps, assets)

Gradle emits these Hermes errors
- [`async functions are unsupported`](https://github.com/facebook/hermes/blob/main/lib/IRGen/ESTreeIRGen-func.cpp#L213-L217)
  - https://github.com/facebook/hermes/issues/1395
- [`async generators are unsupported`](https://github.com/facebook/hermes/blob/main/lib/AST/SemanticValidator.cpp#L798-L800)

(at runtime we can see both are [`SyntaxError`](https://github.com/facebook/hermes/blob/main/tools/hermes-parser/js/hermes-parser/src/utils/createSyntaxError.js)s)
Resulting in vague `java.lang.StackOverflowError (no error message)`

The try/catch approach testing language ft support via a new fn works
since RNGP no longer emits the Hermes errors in the task after Metro bundles
and we're conditionally using parts of SES compatible with the Hermes engine

The initial approach involved building a new shim via an env var
which involved a lot of duplicates `/src` files

Nb: clean before bundling to see changes reflected (avoid cache)
i.e. `./gradlew clean :app:bundleRelease`

[...]

PR #2563, opened 2024-10-02
fix(ses): completePrototypes on Hermes

leotm:
## Description

- Refs: https://github.com/endojs/endo/issues/1891
  - completePrototypes
- Follow-up to
  - https://github.com/endojs/endo/pull/2334

Calling lockdown on Hermes, in the VM or React Native runtime, we reach

https://github.com/endojs/endo/blob/7b8c677b364bb076700f376362022fbd9d109c6e/packages/ses/src/intrinsics.js#L101-L104

```console
# Hermes VM
Uncaught TypeError: lockdown.prototype property not whitelisted
```

lockdown prototype property not whitelisted

checking the `lockdown` intrinsic descriptor

```json
{"length":{"value":1,"writable":false,"enumerable":false,"configurable":true},"name":{"value":"","writable":false,"enumerable":false,"configurable":true},"caller":{"enumerable":false,"configurable":false},"arguments":{"enumerable":false,"configurable":false},"prototype":{"value":{},"writable":true,"enumerable":false,"configurable":false}}
```

we see it's because the `prototype` is non-standard

https://github.com/endojs/endo/blob/7b8c677b364bb076700f376362022fbd9d109c6e/packages/ses/src/permits.js#L1644

https://github.com/endojs/endo/blob/7b8c677b364bb076700f376362022fbd9d109c6e/packages/ses/src/permits.js#L282-L283

https://github.com/endojs/endo/blob/7b8c677b364bb076700f376362022fbd9d109c6e/packages/ses/src/permits.js#L265-L273

so we should be terminating the current loop iteration instead of throwing the error, ideally at

https://github.com/endojs/endo/blob/7b8c677b364bb076700f376362022fbd9d109c6e/packages/ses/src/intrinsics.js#L93-L96

the [initial](https://github.com/endojs/endo/pull/2563/commits/415719fecdafd04e15ea487025dd75d85b1c2485) approach fixed our 3 problematic intrinsics on Hermes
by terminating the current loop iteration earlier
before throwing the error as intended
i.e. skip completing prototypes that shouldn't be there

Repro branch including proposed fix and detailed Hermes VM print output (no Babel)

- https://github.com/endojs/endo/pull/2334
  - https://github.com/endojs/endo/tree/ses-hermes-completePrototypes-repro
  - https://github.com/endojs/endo/tree/ses-hermes-whitelistIntrinsics-repro
    - `yarn clean && yarn build:hermes && test:hermes`

TODO: check [v0.13.0 for RN0.75.x](https://github.com/facebook/hermes/releases/tag/v0.13.0) ✅ partially fixed: `lockdown` intrinsic only
we can easily test this in https://github.com/endojs/endo/pull/2334 by downloading the .tar.gz locally then updating both [binary paths](https://github.com/endojs/endo/blob/7fc4678b9cd6639604dd7c88daf9a3ff22a96ca7/packages/ses/scripts/hermes.sh#L22-L23)

```js
// Hermes release v0.13.0 for RN0.75.x

// lockdown, fixed - no prototype
{"length":{"value":1,"writable":false,"enumerable":false,"configurable":true},"name":{"value":"","writable":false,"enumerable":false,"configurable":true},"caller":{"enumerable":false,"configurable":false},"arguments":{"enumerable":false,"configurable":false}}

// - Uncaught TypeError: harden.prototype property not whitelisted
{"length":{"value":1,"writable":false,"enumerable":false,"configurable":true},"name":{"value":"harden","writable":false,"enumerable":false,"configurable":true},"caller":{"enumerable":false,"configurable":false},"arguments":{"enumerable":false,"configurable":false},"prototype":{"value":{},"writable":true,"enumerable":false,"configurable":false}}

// - Uncaught TypeError: %InitialGetStackString%.prototype property not whitelisted
{"length":{"value":1,"writable":false,"enumerable":false,"configurable":true},"name":{"value":"getStackString","writable":false,"enumerable":false,"configurable":true},"caller":{"enumerable":false,"configurable":false},"arguments":{"enumerable":false,"configurable":false},"prototype":{"value":{},"writable":true,"enumerable":false,"configurable":false}}
```

nb:

https://github.com/endojs/endo/blob/7b8c677b364bb076700f376362022fbd9d109c6e/packages/ses/src/permits.js#L1644-L1647

TODO: build Static Hermes with cmake and ninja then test (i.e. is it fully fixed upstream?)
- https://github.com/facebook/hermes/tree/sh_stable
- https://github.com/facebook/hermes/tree/static_h

yes ✅ object literals short hand method
- https://github.com/facebook/hermes/issues/1371
- https://github.com/facebook/hermes/commit/00f18c89c720e1c34592bb85a1a8d311e6e99599

the current proposed change accounts for Hermes' non-standard fn instance prototype (fixed in Static Hermes), by setting it to `undefined` when completing the prototypes

another approach: https://github.com/endojs/endo/pull/2563#issuecomment-2420363381

### Security Considerations

- proposed pr change is minimal and less intrusive than alternate solution below (Compatibility Considerations)
- new assumptions: if the JS engine contains the 3 unexpected prototypes, tolerate them, if not - great!

### Scaling Considerations

- no further resources consumed

- the solution is backwards compatible with older versions of Hermes (v0.12.0) and React Native to be safe
- intrinsic `lockdown` is fixed on [v0.13.0 for RN0.75.x](https://github.com/facebook/hermes/releases/tag/v0.13.0), so this part of the condition can be sunset when we want to drop support for v0.12.0
- React Native [Releases Support Policy](https://github.com/reactwg/react-native-releases/blob/main/docs/support.md)

### Testing Considerations

- see repro to see completePrototypes finishing successfully on Hermes VM with detailed print output

Considered alternative solution

```js
      if (
        !objectHasOwnProperty(intrinsic, 'prototype') ||
        (typeof intrinsicPrototype === 'object' &&
          // eslint-disable-next-line
          Object.keys(intrinsicPrototype).length === 0)
      ) {
```

which accounts for `{ ... "prototype":{"value":{},"writable":true,"enumerable":false,"configurable":false} }`

but this breaks building the ses bundle when calling `whitelistIntrinsics` after `completePrototypes`

```log
SES_UNCAUGHT_EXCEPTION: (TypeError#1) Unexpected property prototype with permit %ArrayPrototype% at intrinsics.Array.prototype`
```

which appears to be a regression, so no bueno

(See also Documentation Considerations RE backward compatibility)

### Upgrade Considerations

(See also Documentation Considerations RE backward compatibility)

### DIFF
--- a/packages/ses/src/intrinsics.js
+++ b/packages/ses/src/intrinsics.js
@@ -98,25 +98,40 @@ export const makeIntrinsicsCollector = () => {
       if (typeof permit !== 'object') {
         throw TypeError(`Expected permit object at whitelist.${name}`);
       }
-      const namePrototype = permit.prototype;
-      if (!namePrototype) {
+      const permitPrototype = permit.prototype;
+
+      if (
+        typeof intrinsic === 'function' &&
+        intrinsic.prototype !== undefined &&
+        permitPrototype === 'undefined' // permits.js
+      ) {
+        // Set non-standard `.prototype` properties to `undefined` on Hermes.
+        // These include intrinsics that are additional properties of the global object,
+        // proposed by SES defined as function instances
+        // - arrow functions: lockdown, harden
+        // - concise methods: %InitialGetStackString%
+        intrinsic.prototype = undefined;
+      }
+
+      const intrinsicPrototype = intrinsic.prototype;
+
+      if (!permitPrototype) {
         throw TypeError(`${name}.prototype property not whitelisted`);
       }
       if (
-        typeof namePrototype !== 'string' ||
-        !objectHasOwnProperty(permitted, namePrototype)
+        typeof permitPrototype !== 'string' ||
+        !objectHasOwnProperty(permitted, permitPrototype)
       ) {
         throw TypeError(`Unrecognized ${name}.prototype whitelist entry`);
       }
-      const intrinsicPrototype = intrinsic.prototype;
-      if (objectHasOwnProperty(intrinsics, namePrototype)) {
-        if (intrinsics[namePrototype] !== intrinsicPrototype) {
-          throw TypeError(`Conflicting bindings of ${namePrototype}`);
+      if (objectHasOwnProperty(intrinsics, permitPrototype)) {
+        if (intrinsics[permitPrototype] !== intrinsicPrototype) {
+          throw TypeError(`Conflicting bindings of ${permitPrototype}`);
         }
         // eslint-disable-next-line no-continue
         continue;
       }
-      intrinsics[namePrototype] = intrinsicPrototype;
+      intrinsics[permitPrototype] = intrinsicPrototype;
     }
   };
   freeze(completePrototypes);
--- a/packages/ses/src/permits.js
+++ b/packages/ses/src/permits.js
@@ -270,6 +270,9 @@ export const FunctionInstance = {
   // Do not specify "prototype" here, since only Function instances that can
   // be used as a constructor have a prototype property. For constructors,
   // since prototype properties are instance-specific, we define it there.
+  // Older versions of Hermes contain a non-standard prototype, noted below in 'hermesFn'.
+  // This is fixed fully in Static Hermes: https://github.com/facebook/hermes/tree/static_h
+  // See: https://speakerdeck.com/tmikov2023/optimizing-with-static-hermes-chain-react-2024
 };
 
 // AsyncFunction Instances
@@ -281,6 +284,14 @@ export const AsyncFunctionInstance = {
 
 // Aliases
 const fn = FunctionInstance;
+// Bypass Hermes bugs, fixed in:
+// - https://github.com/facebook/hermes/commit/c42491de94aff479e5e83c073eff96a6261da080 (hermes: v0.13.0)
+// - https://github.com/facebook/hermes/commit/00f18c89c720e1c34592bb85a1a8d311e6e99599 (sh_stable, static_h)
+// Expect Additional Properties of the Global Object (Annex B) proposed by SES,
+// that are function instances defined as arrow functions ('lockdown' and 'harden')
+// and concise methods ('%InitialGetStackString%'), to have their non-standard
+// prototype properties set to `undefined` in intrinsics.js when completing prototypes.
+const hermesFn = { ...FunctionInstance, prototype: 'undefined' };
 const asyncFn = AsyncFunctionInstance;
 
 const getter = {
@@ -1643,8 +1654,8 @@ export const permitted = {
     '@@toStringTag': 'string',
   },
 
-  lockdown: fn,
-  harden: { ...fn, isFake: 'boolean' },
+  lockdown: hermesFn,
+  harden: { ...hermesFn, isFake: 'boolean' },
 
-  '%InitialGetStackString%': fn,
+  '%InitialGetStackString%': hermesFn,
 };


leotm:
draft pending testing changes from endo sync in VM and React Native

leotm:
we considered implementing

https://github.com/endojs/endo/blob/a13a879a8ea647d3e13bc6cb223de1b413e06632/packages/ses/src/intrinsics.js#L103-L110

instead in

```js
// packages/ses/src/tame-function-tostring.js
// ...
      if (
        !/^[A-Z]/.test(func.name) &&
        func.prototype !== undefined
      ) {
        func.prototype = undefined;
      }
// ...
```

but went for the simpler non-regex approach

leotm:
https://github.com/endojs/endo/actions/runs/11366289654/job/31616329403?pr=2563

```console
Run npx playwright install --with-deps
  npx playwright install --with-deps
  shell: /usr/bin/bash -e {0}
Installing dependencies...
Switching to root user to install dependencies...
Hit:1 http://azure.archive.ubuntu.com/ubuntu noble InRelease
Get:[2](https://github.com/endojs/endo/actions/runs/11366289654/job/31616329403?pr=2563#step:9:2) http://azure.archive.ubuntu.com/ubuntu noble-updates InRelease [126 kB]
Get:[3](https://github.com/endojs/endo/actions/runs/11366289654/job/31616329403?pr=2563#step:9:3) http://azure.archive.ubuntu.com/ubuntu noble-backports InRelease [126 kB]
Get:[4](https://github.com/endojs/endo/actions/runs/11366289654/job/31616329403?pr=2563#step:9:5) http://azure.archive.ubuntu.com/ubuntu noble-security InRelease [126 kB]
Hit:[5](https://github.com/endojs/endo/actions/runs/11366289654/job/31616329403?pr=2563#step:9:6) https://packages.microsoft.com/repos/azure-cli noble InRelease
Get:[6](https://github.com/endojs/endo/actions/runs/11366289654/job/31616329403?pr=2563#step:9:7) https://packages.microsoft.com/ubuntu/24.04/prod noble InRelease [3600 B]
Get:7 http://azure.archive.ubuntu.com/ubuntu noble-updates/main amd64 Packages [59[7](https://github.com/endojs/endo/actions/runs/11366289654/job/31616329403?pr=2563#step:9:8) kB]
Get:8 http://azure.archive.ubuntu.com/ubuntu noble-updates/main Translation-en [146 kB]
Get:9 http://azure.archive.ubuntu.com/ubuntu noble-updates/main amd64 c-n-f Metadata [10.3 kB]
Get:10 http://azure.archive.ubuntu.com/ubuntu noble-updates/universe amd64 Packages [706 kB]
Get:11 http://azure.archive.ubuntu.com/ubuntu noble-updates/universe Translation-en [209 kB]
Get:12 http://azure.archive.ubuntu.com/ubuntu noble-updates/universe amd64 c-n-f Metadata [19.[8](https://github.com/endojs/endo/actions/runs/11366289654/job/31616329403?pr=2563#step:9:9) kB]
Get:13 http://azure.archive.ubuntu.com/ubuntu noble-updates/restricted amd64 Packages [388 kB]
Get:14 http://azure.archive.ubuntu.com/ubuntu noble-updates/restricted Translation-en [74.8 kB]
Get:15 http://azure.archive.ubuntu.com/ubuntu noble-updates/multiverse amd64 Packages [14.8 kB]
Get:16 http://azure.archive.ubuntu.com/ubuntu noble-updates/multiverse Translation-en [3820 B]
Get:17 http://azure.archive.ubuntu.com/ubuntu noble-updates/multiverse amd64 c-n-f Metadata [552 B]
Get:18 http://azure.archive.ubuntu.com/ubuntu noble-backports/universe amd64 Packages [10.6 kB]
Get:1[9](https://github.com/endojs/endo/actions/runs/11366289654/job/31616329403?pr=2563#step:9:10) http://azure.archive.ubuntu.com/ubuntu noble-backports/universe Translation-en [10.8 kB]
Get:20 http://azure.archive.ubuntu.com/ubuntu noble-backports/universe amd64 c-n-f Metadata [1[10](https://github.com/endojs/endo/actions/runs/11366289654/job/31616329403?pr=2563#step:9:11)4 B]
Get:21 http://azure.archive.ubuntu.com/ubuntu noble-security/main amd64 Packages [410 kB]
Get:22 http://azure.archive.ubuntu.com/ubuntu noble-security/main Translation-en [90.4 kB]
Get:23 http://azure.archive.ubuntu.com/ubuntu noble-security/main amd64 c-n-f Metadata [5788 B]
Get:24 http://azure.archive.ubuntu.com/ubuntu noble-security/universe amd64 Packages [553 kB]
Get:25 http://azure.archive.ubuntu.com/ubuntu noble-security/universe Translation-en [147 kB]
Get:26 http://azure.archive.ubuntu.com/ubuntu noble-security/universe amd64 c-n-f Metadata [13.5 kB]
Get:27 https://packages.microsoft.com/ubuntu/24.04/prod noble/main armhf Packages [5057 B]
Get:28 https://packages.microsoft.com/ubuntu/24.04/prod noble/main amd64 Packages [12.2 kB]
Get:29 https://packages.microsoft.com/ubuntu/24.04/prod noble/main arm64 Packages [8451 B]
Fetched 3820 kB in 1s (7439 kB/s)
Reading package lists...
Reading package lists...
Building dependency tree...
Reading state information...
Package libicu70 is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

Package libasound2 is a virtual package provided by:
  liboss4-salsa-asound2 4.2-build2020-1ubuntu3
  libasound2t64 1.2.[11](https://github.com/endojs/endo/actions/runs/11366289654/job/31616329403?pr=2563#step:9:12)-1build2 (= 1.2.11-1build2)

E: Package 'libasound2' has no installation candidate
E: Package 'libicu70' has no installation candidate
E: Unable to locate package libffi7
E: Unable to locate package libx264-163
Failed to install browsers
Error: Installation process exited with code: 100
Error: Process completed with exit code 1.
```

flakey CI, re-run

gibson042:
Another thought: concise functions in Hermes may incorrectly have a `prototype` property, but bound functions do not:
```console
$ eshost -se '[()=>{}, (()=>{}).bind()].map(fn => Object.hasOwnProperty.call(fn, "prototype"))'
#### engine262, GraalJS, JavaScriptCore, Moddable XS, QuickJS, SpiderMonkey, V8
false,false

#### Hermes
true,false
```

So any receiver-insensitive function that needs be presented as non-newable (which I think describes all functions that we want to appear native) can be wrapped with `bind`.

legobeat:
> ```
> libasound2t64
> ```

I believe this is because GitHub runner images switched from `20.04` to `22.04`. Should re resolved by a prior `DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends libasound2t64`

leotm:
> Another thought: concise functions in Hermes may incorrectly have a `prototype` property, but bound functions do not:
> 
> ```
> $ eshost -se '[()=>{}, (()=>{}).bind()].map(fn => Object.hasOwnProperty.call(fn, "prototype"))'
> #### engine262, GraalJS, JavaScriptCore, Moddable XS, QuickJS, SpiderMonkey, V8
> false,false
> 
> #### Hermes
> true,false
> ```
> 
> So any receiver-insensitive function that needs be presented as non-newable (which I think describes all functions that we want to appear native) can be wrapped with `bind`.

good point ^ same with normal functions

```console
# Hermes release version: 0.12.0, HBC bytecode version: 89
>> [function(){}, (function(){}).bind()].map(fn => Object.hasOwnProperty.call(fn, "prototype"))
[ true, false, [length]: 2 ]
```

so binding the intrinsic works too after restoring the permits
https://github.com/endojs/endo/tree/ses-hermes-completePrototypes-bind
instead of setting the `prototype` to `undefined`

packages/ses/src/intrinsics.js
```diff
@@ -98,25 +98,36 @@ export const makeIntrinsicsCollector = () => {
       if (typeof permit !== 'object') {
         throw TypeError(`Expected permit object at whitelist.${name}`);
       }
-      const namePrototype = permit.prototype;
-      if (!namePrototype) {
+      const permitPrototype = permit.prototype;
+
+      // Bypass Hermes bug, fixed in: https://github.com/facebook/hermes/commit/00f18c89c720e1c34592bb85a1a8d311e6e99599
```
erights, line 103: Please explain the Hermes bug you're bypassing below, in addition to the commit link.

packages/ses/src/permits.js
```diff
@@ -281,6 +282,7 @@ export const AsyncFunctionInstance = {
 
 // Aliases
 const fn = FunctionInstance;
+const hermesFn = { ...FunctionInstance, prototype: 'undefined' }; // Bypass Hermes bug, fixed in: https://github.com/facebook/hermes/commit/00f18c89c720e1c34592bb85a1a8d311e6e99599
```
erights, line 285: Likewise. Is there an issue for this Hermes bug?

packages/ses/src/permits.js
```diff
@@ -1641,8 +1643,8 @@ export const permitted = {
     '@@toStringTag': 'string',
   },
 
-  lockdown: fn,
-  harden: { ...fn, isFake: 'boolean' },
+  lockdown: hermesFn,
+  harden: { ...hermesFn, isFake: 'boolean' },
 
-  '%InitialGetStackString%': fn,
+  '%InitialGetStackString%': hermesFn,
```
erights, line 1649: Why exactly these three?

leotm: it's these three intrinsics ^ that throw at 

https://github.com/endojs/endo/blob/7a991aa75b06fc418459b5e5d2d08f9fb73ab113/packages/ses/src/intrinsics.js#L101-L104

when calling `lockdown()` on Hermes VM (surfaced in React Native)

leotm: on Hermes [v0.12.0](https://github.com/facebook/hermes/releases/tag/v0.12.0) it's these three ^
on [v0.13.0 for RN0.75.x](https://github.com/facebook/hermes/releases/tag/v0.13.0) only the next two

otherwise on Static Hermes (unreleased) it's [fixed](https://github.com/facebook/hermes/issues/1371#issuecomment-2050775512)
which atm can only be tested by [building and running](https://github.com/facebook/hermes/blob/main/doc/BuildingAndRunning.md) either [sh_stable](https://github.com/facebook/hermes/tree/sh_stable) or [static_h](https://github.com/facebook/hermes/tree/static_h)

after https://github.com/endojs/endo/pull/2334 lands, adding v0.13.0 to CI is up next (it's [not](https://github.com/facebook/hermes/issues/1371#issuecomment-2341569115) on npm, so will need to wget the .tar.gz)
followed by Static Hermes next (no distro yet so in CI will need to build with cmake)

leotm: the [initial](https://github.com/endojs/endo/pull/2563/commits/415719fecdafd04e15ea487025dd75d85b1c2485) approach was minimal and to ignore these three, but it doesn't remove them

the proposed change expects `undefined` and sets `undefined` which seems better

or calling `.bind()` on these three instead which also works ([repro](https://github.com/endojs/endo/tree/ses-hermes-completePrototypes-bind))

erights: I'm confused. Why doesn't https://github.com/endojs/endo/pull/1221 make this a non-issue, setting the `.prototype` to `undefined` itself?

erights: Is it because the test in intrinsics.js is inconsistently stricter than the #1221 test in permit-intrinsics.js and runs first?

I also looked at the definition of these three functions to see what's different there. It looks like both `harden` and `getStackString` are defined as concise methods, so I'm inferring that the hermes bug is that concise methods have a `.prototype`. Does hermes also make these incorrectly constructable (i.e., with `new`)? Is this also fixed in more recent hermeses? (what's the plural of "hermes"?)

But AFAICT `lockdown` is defined as an arrow function. So why is it more problematic than a zillion other arrow functions? Or, perhaps, is it the only intrinsic defined by us as an arrow function?

If all this is true, then it makes sense to permit `prototype: undefined` on these three. So I'll approve on that basis. But before merging, please let us know if this is indeed all true.

erights (approved):
LGTM module suggestions, especially https://github.com/endojs/endo/pull/2563#discussion_r1807040061

packages/ses/src/permits.js
erights: Separately, it looks like a bug that

> the test in intrinsics.js is inconsistently stricter than the https://github.com/endojs/endo/pull/1221 test in permit-intrinsics.js and runs first

Could you file an issue on that and assign it to me? Thanks.

leotm: > I also looked at the definition of these three functions to see what's different there.

awesome ^ was about to examine how these three are defined next

> It looks like both harden and getStackString are defined as concise methods, so I'm inferring that the hermes bug is that concise methods have a .prototype.

exactly, but only a bug on Hermes v0.12.0 (and older)

```js
// Hermes release version: 0.12.0, HBC bytecode version: 89
var arrowFn = () => {}
arrowFn.prototype // { [constructor]: [Function] }
```

> Does hermes also make these incorrectly constructable (i.e., with new)?

fortunately not, so only a fake .prototype constructor on concise methods

```js
// Hermes release version: 0.12.0, HBC bytecode version: 89
var arrowFn = () => {}
arrowFn.prototype // { [constructor]: [Function] }
new arrowFn // Uncaught TypeError: Function is not a constructor
```

> Is this also fixed in more recent hermeses? (what's the plural of "hermes"?)

thought this was only fixed in Static Hermes, [mentioned](https://github.com/facebook/hermes/issues/1371#issuecomment-2050775512) in the issue and commit
but from testing it's been back ported to v0.13.0 too!

```js
// Hermes release version: 0.13.0, HBC bytecode version: 96 (bug: fixed)
var arrowFn = () => {}
arrowFn.prototype // undefined
```

```js
// Static Hermes custom version: 25d892b, HBC bytecode version: 98 (bug: fixed)
var arrowFn = () => {}
arrowFn.prototype // undefined
```

leotm: side note: looks like _all_ Hermes [releases](https://github.com/facebook/hermes/releases) have been published to GitHub now?! (but still not npmjs, if planned)

but it remains there's no need for SES to support these older versions
since metamask-mobile is now on React Native 0.72.15
whereas Hermes v0.11.0 for RN0.68.x has been [unsupported](https://github.com/reactwg/react-native-releases/blob/main/docs/support.md) for years

leotm: > But AFAICT `lockdown` is defined as an arrow function. So why is it more problematic than a zillion other arrow functions? Or, perhaps, is it the only intrinsic defined by us as an arrow function?

this part's puzzling been puzzling me too...
why is completing the `lockdown` prototype a problem on Hermes v0.12.0 (and older)...
but not on Hermes v0.13.0+ 🤔

> If all this is true, then it makes sense to permit `prototype: undefined` on these three. So I'll approve on that basis. But before merging, please let us know if this is indeed all true.

sure ^ i'll keep digging to find the answer to these last couple questions

leotm: > I'm confused. Why doesn't #1221 make this a non-issue, setting the `.prototype` to `undefined` itself?

> Is it because the test in intrinsics.js is inconsistently stricter than the https://github.com/endojs/endo/pull/1221 test in permit-intrinsics.js and runs first?

exactly the current order and strictness on `master` is

- intrinsics.js > completePrototypes
  - `isObject(intrinsic)`
  - `objectHasOwnProperty(intrinsic, 'prototype')`
  - `typeof permit === 'object'`
  - `!permit.prototype`
    - ```throw TypeError(`${name}.prototype property not whitelisted`);```
- permit-intrinsics.js > whitelistIntrinsics
  - `prop in obj`
  - `typeof obj === 'function' && prop === 'prototype'`
    - `obj.prototype = undefined` (not reached)
    - #1221

leotm: > Separately, it looks like a bug that
> 
> > the test in intrinsics.js is inconsistently stricter than the #1221 test in permit-intrinsics.js and runs first
> 
> Could you file an issue on that and assign it to me? Thanks.

- https://github.com/endojs/endo/issues/2598

packages/ses/src/intrinsics.js
leotm: TODO: add explanation pending investigation how `lockdown` / `harden` / `%InitialGetStackString%` are defined in SES
- if object literal concise method, commit link is correct
- if concise function, https://github.com/facebook/hermes/commit/c42491de94aff479e5e83c073eff96a6261da080

packages/ses/src/permits.js
leotm: yep the issue for concise methods is
- https://github.com/facebook/hermes/issues/1371

for concise (arrow) functions [no issue](https://github.com/facebook/hermes/issues?q=prototype+arrow+function) was raised only commit https://github.com/facebook/hermes/commit/c42491de94aff479e5e83c073eff96a6261da080

packages/ses/src/intrinsics.js
leotm: just clarifying the above (how these three are defined in SES)

- lockdown: arrow fn
  - harden: arrow fn
- getStackString: object literal concise (shorthand) method

https://github.com/endojs/endo/blob/4ca4028bb67ba9198060f30a7e5bbff21e8fa27f/packages/ses/src/lockdown-shim.js#L15-L18

https://github.com/endojs/endo/blob/4ca4028bb67ba9198060f30a7e5bbff21e8fa27f/packages/ses/src/lockdown.js#L408

https://github.com/endojs/endo/blob/4ca4028bb67ba9198060f30a7e5bbff21e8fa27f/packages/ses/src/error/tame-error-constructor.js#L21-L23

so added commit earlier updating documentation in permits.js and intrinsics.js https://github.com/endojs/endo/pull/2563/commits/575cd60c5d1322604430b89681d37989e2f34239

packages/ses/src/permits.js
leotm: > So why is it more problematic than a zillion other arrow functions?

for example checking: permits.js > Appendix B > Annex B

https://github.com/endojs/endo/blob/4ca4028bb67ba9198060f30a7e5bbff21e8fa27f/packages/ses/src/permits.js#L1613-L1650

these aren't problematic on Hermes
since they're not proposed by SES
so already implemented correctly in Hermes natively:

```js
function escape() { [native code] }
function unescape() { [native code] }
```

however our [three functions proposed by SES](https://github.com/endojs/endo/pull/2563#discussion_r1816606864) are problematic
since we are adding them to Hermes as [buggy](https://github.com/endojs/endo/pull/2563#discussion_r1811111552) arrow functions and concise methods:

```js
function (a0) { [bytecode] } // lockdown
function harden(a0) { [bytecode] } // harden
function getStackString(a0) { [bytecode] } // intrinsic: %InitialGetStackString%
```

packages/ses/src/permits.js
```diff
@@ -270,6 +270,7 @@ export const FunctionInstance = {
   // Do not specify "prototype" here, since only Function instances that can
   // be used as a constructor have a prototype property. For constructors,
   // since prototype properties are instance-specific, we define it there.
+  // The exception to this is Hermes, fixed since in Static Hermes.
```
boneskull, line 273: I don't know what "Static Hermes" is.  add reference to comment?

boneskull (commented):
I'm too unfamiliar with this to feel comfortable approving it. But here I am anyway!

@leotm Are we testing against Hermes in CI somewhere?  Can you point me to it?

packages/ses/src/permits.js
leotm: added here https://github.com/endojs/endo/pull/2563/commits/1d4bd35424d312e864109378f0ee0c8bab4f9ced

leotm: > @leotm Are we testing against Hermes in CI somewhere? Can you point me to it?

both proposed 2 changes (permits.js, intrinsics.js) can be verified working against Hermes CI in
https://github.com/endojs/endo/tree/ses-hermes-whitelistIntrinsics-repro
running `yarn build:hermes && yarn test:hermes` where completePrototypes is now fixed on Hermes

otherwise final changes to https://github.com/endojs/endo/pull/2334 nearly ready for official Hermes CI feedback

leotm:
> LGTM module suggestions, especially [#2563 (comment)](https://github.com/endojs/endo/pull/2563#discussion_r1807040061)

> If all this is true, then it makes sense to permit prototype: undefined on these three. So I'll approve on that basis. But before merging, please let us know if this is indeed all true.

all questions answered and addressed ^ in comments and commits

erights (changes requested):
Now that #2624 is merged, I'm changing to "Request changes".

Does #2624 indeed enable this to be simplified? What does it simplify down to?

packages/ses/src/intrinsics.js
```diff
@@ -98,25 +98,40 @@ export const makeIntrinsicsCollector = () => {
...
         throw TypeError(`Expected permit object at whitelist.${name}`);
       }
-      const namePrototype = permit.prototype;
-      if (!namePrototype) {
+      const permitPrototype = permit.prototype;
+
+      if (
+        typeof intrinsic === 'function' &&
+        intrinsic.prototype !== undefined &&
+        permitPrototype === 'undefined' // permits.js
```
boneskull, line 106: it's literally the string `undefined`?

boneskull (commented):
I wouldn't really understand this stuff without a considerable walkthrough.  Sorry I am useless!

erights:
Yes. When a permit is a string which is a `typeof` name, it means "this field must be a value whose `typeof` is that string. The only value `v` for which `typeof v === 'undefined'` is `undefined`.

packages/ses/src/intrinsics.js
erights: Yes. When a permit is a string which is a typeof name, it means "this field must be a value whose typeof is that string. The only value v for which typeof v === 'undefined' is undefined.

erights: But, IIUC, given #2624 , none of this may still be necessary, which is why I ask.

gibson042: Or the possibility of just forcing these functions to have no `prototype` property [even in Hermes](https://github.com/endojs/endo/pull/2563#issuecomment-2417858134).

leotm: after testing https://github.com/endojs/endo/pull/2624 with `git rebase master` on https://github.com/endojs/endo/pull/2334

can confirm https://github.com/endojs/endo/pull/2624 fixes _completePrototypes_ on Hermes and this PR's changes no longer necessary - thank you!

closing this PR with comment of the results

leotm:
> Now that #2624 is merged, I'm changing to "Request changes".
> 
> Does #2624 indeed enable this to be simplified? What does it simplify down to?

after testing https://github.com/endojs/endo/pull/2624 with `git rebase master` on https://github.com/endojs/endo/pull/2334

can confirm https://github.com/endojs/endo/pull/2624 fixes _completePrototypes_ on Hermes - i.e. this PR simplifies to no longer necessary - thank you!

results

```console
Removing lockdown.prototype
Tolerating undeletable lockdown.prototype === undefined

Removing harden.prototype
Tolerating undeletable harden.prototype === undefined

Removing %InitialGetStackString%.prototype
Tolerating undeletable %InitialGetStackString%.prototype === undefined
```

leotm:
> Now that #2624 is merged, I'm changing to "Request changes".
> 
> Does #2624 indeed enable this to be simplified? What does it simplify down to?

after testing https://github.com/endojs/endo/pull/2624 with `git rebase master` on https://github.com/endojs/endo/pull/2334

can confirm https://github.com/endojs/endo/pull/2624 fixes _completePrototypes_ on Hermes - i.e. this PR simplifies to no longer necessary - thank you!

results

```console
Removing lockdown.prototype
Tolerating undeletable lockdown.prototype === undefined

Removing harden.prototype
Tolerating undeletable harden.prototype === undefined

Removing %InitialGetStackString%.prototype
Tolerating undeletable %InitialGetStackString%.prototype === undefined
```

PR #2565, opened 2024-10-03
Use yarn for cross-workspace script running

kumavis:
Migrates cross-workspace script running to `yarn workspaces`.

`lerna` is still used for:
- ci prepack check after packing
- publishing
- cleaning (removing all `node_modules`)

etc:
- added test:xs script in root
- deprecated "yarn run update" in root


mhofman:
This might break agoric-sdk integrations with Endo. We'll need to run some integration on the agoric-sdk side before this can merge

kumavis:
@mhofman unsure how this would affect `agoric-sdk`. `lerna` is not removed and the change is primarily how the scripts are invoked. where can i learn more?

kumavis:
yarn did not solve the main issue i wanted it to, indicating which workspaces failed 🤕 
https://github.com/yarnpkg/berry/pull/6535

kumavis:
I encountered a "cover" CI issue twice in a row (once on first run, then on retry failed). Retrying all fixed it, so I suspect something to do with [the cache](https://github.com/endojs/endo/blob/master/.github/workflows/ci.yml#L167) (?)
unsure what else it could be. race condition?

scripts/migrate-test-names.mjs
```diff
@@ -21,9 +21,9 @@ const usage = `
 Run this script in a package to convert its test names and Ava config.
 
 To run it over all packages,
-  yarn lerna exec --concurrency 1 --no-bail '../../scripts/migrate-test-names.mjs'
+  yarn workspaces foreach -A exec '../../scripts/migrate-test-names.mjs'
```
turadg, line 24: we could also remove this file because the migration is done.

package.json
```diff
@@ -30,16 +30,17 @@
   },
   "scripts": {
     "clean": "lerna clean",
-    "cover": "lerna run cover",
+    "cover": "yarn workspaces foreach -A run cover",
     "docs": "typedoc",
     "docs:markdown-for-agoric-documentation-repo": "typedoc --plugin typedoc-plugin-markdown --tsconfig tsconfig.build.json",
-    "update": "lernaupdate --dedupe",
+    "update": "yarn upgrade-interactive",
```
turadg, line 36: I don't think anyone uses this, and if they do they can cope with it being removed. It's not run programmatically.

To lower the risk this could be,
```suggestion
    "update": "echo DEPRECATED use yarn's upgrade-interactive or dedupe",
```

.github/workflows/ci.yml
```diff
@@ -247,16 +247,15 @@ jobs:
...
+        run: yarn workspaces foreach -A exec yarn pack
 
       # Prepack (without cleanup per package) to ensure that type resolution in
       # dependent packages continues to work when the typedefs are generated by
       # their upstream packages. This helps avoid a situation in which the types
       # only resolve because of the state of the local filesystem, and fails
       # when imported in an NPM node_modules tree.
       - name: Prepack packages
-        run: yarn lerna run --reject-cycles --concurrency 1 prepack
+        run: yarn workspaces foreach -A run prepack
```
turadg, line 258: please leave this one. this change. loses cycle detection.

it also loses topological ordering, though `--topological` would restore that.

.github/workflows/ci.yml
```diff
@@ -247,16 +247,15 @@ jobs:
         run: yarn run build
 
       - name: pack
-        # Cannot use yarn lerna run pack
-        run: yarn lerna exec yarn pack
+        run: yarn workspaces foreach -A exec yarn pack
```
turadg, line 250: I think this needs topological ordering. Also consider using the long option names for readability (no need to abbreviate since nobody is typing it on the command line)
```suggestion
        run: yarn workspaces foreach --all --topological exec yarn pack
```

turadg (commented):
I don't see a way this could break agoric-sdk's Endo integration but here's a test to be sure: https://github.com/Agoric/agoric-sdk/pull/9385

scripts/migrate-test-names.mjs
kumavis: i considered it but since theres other similar migration files, ill leave removal to a separate PR

turadg (approved):
Thanks for reducing reliance on Lerna.

Code looks good, CI green here and in agoric-sdk integration branch

kumavis:
(rebased)

PR #2566, opened 2024-10-03
fix(ses): package test correctly rejects instead of uncaught exception

kumavis:
test assertion failure outside of the test function leads to uncaught exception


turadg (approved):
👍

PR #2575, opened 2024-10-05
tsconfig moduleResolution: NodeNext

turadg:
_incidental_

## Description

In an a3p test I tried disabling `skipLibCheck` and many errors were from Endo packages's omission of module filename extensions.

That's because Endo was using the default [moduleResolution](https://www.typescriptlang.org/tsconfig/#moduleResolution), `node10`. This changes it to `NodeNext` and also the `module` option for compatibility.

That resulted in new errors which I address mostly by switching to explicit `@import`.

### Security Considerations

none

### Scaling Considerations

none

none

### Testing Considerations

- https://github.com/Agoric/agoric-sdk/pull/9385

This might affect consuming packages that depend on ambient types. That's almost certainly only agoric-sdk which we can solve with https://github.com/Agoric/agoric-sdk/pull/9385

### Upgrade Considerations

If https://github.com/Agoric/agoric-sdk/pull/9385 needs changes then `NEWS.md` should indicate it. I don't think type changes live up to `BREAKING` which would be a semver major bump.

### DIFF
--- a/packages/evasive-transform/src/generate.js
+++ b/packages/evasive-transform/src/generate.js
@@ -37,7 +37,7 @@ const generator = /** @type {typeof import('@babel/generator')['default']} */ (
  * provided to the options.
  *
  * @template {string|undefined} [SourceUrl=undefined]
- * @typedef {{code: string, map: SourceUrl extends string ? import('source-map').RawSourceMap : never}} TransformedResult
+ * @typedef {{code: string, map: SourceUrl extends string ? import('source-map-js').RawSourceMap : never}} TransformedResult
  * @internal
  */
 
--- a/packages/evasive-transform/src/generate.js
+++ b/packages/evasive-transform/src/generate.js
@@ -4,6 +4,7 @@
  * @module
  */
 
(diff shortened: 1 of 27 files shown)


mhofman:
I would like to see @SMotaal chime in because from what I recall they did an investigation of module resolution and impact on types (especially exports which are well tested).

kriskowal (approved):
I am willing to chance this. I like that it catches missing extensions, which we already favor.

I think @boneskull might be able to fill in for @SMotaal on whether MetaMask recommends moving in this direction.

turadg:
The agoric-sdk integration PR passed without any changes so I don't think this breaks anything. Enabling auto-merge with rebase.

packages/evasive-transform/src/generate.js
```diff
@@ -4,6 +4,7 @@
  * @module
  */
 
+// @ts-ignore XXX no types defined
```
boneskull, line 7: nit: prefer `@ts-expect-error` over `@ts-ignore`.  the only time I use the latter is when I'm seeing flaky behavior

PR #2580, opened 2024-10-10
chore: Release 2024-10-10

kriskowal:
# `ses` v1.9.0

- On platforms without
  [`Array.prototype.transfer`](https://github.com/tc39/proposal-resizablearraybuffer)
  but with a global `structuredClone`, the ses-shim's `lockdown` will now
  install an emulation of `Array.prototype.transfer`. On platforms with neither,
  the ses-shim will *currently* not install such an emulation.
  However, once we verify that endo is not intended to support platforms
  without both, we may change `lockdown` to throw, failing to lock down.
  - XS and Node >= 22 already have `Array.prototype.transfer`.
  - Node 18, Node 20, and all browsers have `structuredClone`
  - Node <= 16 have neither, but are also no longer supported by Endo.
- Now exports separate layer for console shim: `ses/console-shim.js`.
- Adds permits for `ModuleSource`, either the native implementation or from
  `@endo/module-source/shim.js`.

# `@endo/compartment-mapper` v1.3.0

- Adds support for dynamic requires in CommonJS modules. This requires specific
  configuration to be passed in (including new read powers), and is _not_
  enabled by default. See the signature of `loadFromMap()` in `import-lite.js`
  for details.

# `@endo/import-bundle` v1.3.0

- Adds support for `endoScript` format bundles.

# `@endo/module-source` v1.1.0

- Adds `@endo/module-source/shim.js` to shim `globalThis.ModuleSource`.
  The shim currently replaces the native `globalThis.ModuleSource` if present.

# `@endo/captp` v4.4.0

- Add optional configuration `makeCapTPImportExportTables` for external management of import/export tables.

### DIFF
--- a/packages/base64/CHANGELOG.md
+++ b/packages/base64/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+### [1.0.8](https://github.com/endojs/endo/compare/@endo/base64@1.0.7...@endo/base64@1.0.8) (2024-10-10)
+
+**Note:** Version bump only for package @endo/base64
+
+
+
+
+
 ### [1.0.7](https://github.com/endojs/endo/compare/@endo/base64@1.0.6...@endo/base64@1.0.7) (2024-08-27)
 
 **Note:** Version bump only for package @endo/base64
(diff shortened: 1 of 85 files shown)


boneskull:
:shipit:

michaelfig (approved):
Whew!  Mostly mechanical and easy to skim.  LGTM!

PR #2582, opened 2024-10-11
fix(ses): reject unsupported lockdownOptions mathTaming + dateTaming

kumavis:
support for lockdown options `tameMath` and `tameDate` were removed so long ago I'm having trouble finding a proper reference (pre-2020? pre migration from agoric into endo?). Since these options don't do anything, lockdown should fail when these options are specified.

### DIFF
--- a/packages/ses/NEWS.md
+++ b/packages/ses/NEWS.md
@@ -1,5 +1,9 @@
 User-visible changes in `ses`:
 
+# Unreleased
+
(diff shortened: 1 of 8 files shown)


kumavis:
@kriskowal unrelated playwright install issue (?)

kumavis:
replaced by https://github.com/endojs/endo/pull/2583

PR #372, opened 2020-07-02
fix(ses): Distinguish intrinsics in start and non-start compartments

erights:
Whitelists no longer painfully redundant.
Placeholder for multiple Compartment constructors, including one which only throws.

Start compartment has initial authority. Others do not. Independent of *Taming options. 
`{ *Taming: 'unsafe' }` is no longer about initial authority in start compartment, as long as it obeys ocap rules.
'unsafe' is now about enabling violations of ocap rules, such as primordials that are no longer powerless.
Lockdown *Taming options have much milder meaning.
   * We still check that they are only set to 'safe' or 'unsafe', with default 'safe'. 
   * `dateTaming` is ignored. Maybe we should just get rid of this flag. Transition problem?
   * `mathTaming` is ignored. Maybe we should just get rid of this flag. Transition problem?
   * `regExpTaming` is effectively ignored. But we may decide that 'unsafe' suppresses the deletion of `RegExp.prototype.compile`.
   * The new `localeTaming`, if set to 'unsafe', leaves the original `*Locale*` methods in place.

Finally we get to Error taming. Independent of the `errorTaming` option, the tamed powerful v8 API emulation remains on the start compartment's `Error` constructor. None of these are on the `Error` constructor of other compartments. However, this shows that we need a different knob: whether to expose engine-specific (e.g., v8 specific) APIs.

In a step towards https://tc39.es/proposal-error-stacks/ , we leave on the start compartment the powerful `getStackString(error)` function, which gives the stacktrace string associated with that error. Because this special power is only on the start compartment, we do not consider it a material violation of ocap rules, so it is there independent of `errorTaming`.

The violation of ocap rules which is pervasively available is the legacy behavior of `error.stack`, since it takes no privilege but violates information hiding. Accordingly, on v8 where we have control, under `errorTaming: 'safe'` (or default) this is truncated to the empty string. The 'unsafe' setting leaves the legacy behavior alone, providing the same string that the new `getStackString(error)` provides.

Fixes #317 
Fixes #385 by getting rid of the realmRec altogether.

PR #2583, opened 2024-10-11
fix(ses): reject unsupported lockdownOptions mathTaming + dateTaming

kumavis:
This is a breaking change.

support for lockdown options `mathTaming` and `dateTaming` were removed so long ago I'm having trouble finding a proper reference (edit: july 2020 https://github.com/endojs/endo/pull/372). Since these options don't do anything, lockdown should fail when these options are specified.

Here is a non-breaking change that adds a warning when the options are specified https://github.com/endojs/endo/pull/2584

### DIFF
--- a/packages/ses/NEWS.md
+++ b/packages/ses/NEWS.md
@@ -1,5 +1,9 @@
 User-visible changes in `ses`:
 
+# Next major version
+
+- Specifying the long discontinued `mathTaming` or `dateTaming` options throws an Error.
+
 # v1.9.0 (2024-10-10)
 
 - On platforms without
--- a/packages/ses/src/lockdown.js
+++ b/packages/ses/src/lockdown.js
@@ -188,8 +188,6 @@ export const repairIntrinsics = (options = {}) => {
       'safe',
     ),
     __hardenTaming__ = getenv('LOCKDOWN_HARDEN_TAMING', 'safe'),
-    dateTaming = 'safe', // deprecated
-    mathTaming = 'safe', // deprecated
     ...extraOptions
   } = options;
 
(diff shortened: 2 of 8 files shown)


kumavis:
@kriskowal unrelated playwright install issue (?)

kriskowal:
> @kriskowal unrelated playwright install issue (?)

Definitely unrelated. We saw this once before when one of the browsers shifted underneath us. I reran the job and it persisted. It’s not a required job.

packages/ses/NEWS.md
```diff
@@ -1,5 +1,9 @@
 User-visible changes in `ses`:
 
+# Unreleased
```
kriskowal, line 3:
```suggestion
# Next version
```

There’s a note in CONTRIBUTING: in the release process, I grep for `# Next`. I (or whomever pulls the crank) could also grep for `# Unreleased` if you want to update the release process.

kriskowal (changes requested):
This is a breaking change. We should use a console warning instead, and leave a note somewhere for behaviors we would change to errors in the next-breaking version, maybe next year.

packages/ses/NEWS.md
kumavis: just adopting a previous pattern, will update

kumavis:
@kriskowal whats the process for queueing breaking changes?

boneskull (commented):
Is it unacceptable to just ignore the options?

EDIT:

My _personal_ opinion is that this should not throw, because shipping a breaking change should be a _last resort_.

kriskowal:
> @kriskowal whats the process for queueing breaking changes?

We do not have one yet, but I think a new label `next-major` is in order. The queue is not empty (`__options__` on `Compartment` options bag, deprecated module descriptor formats, &c.)

Closing in favor of https://github.com/endojs/endo/pull/2584

PR #2584, opened 2024-10-12
fix(ses): warn on unsupported lockdownOptions mathTaming + dateTaming

kumavis:
support for lockdown options `mathTaming` and `dateTaming` were [removed in July 2020](https://github.com/endojs/endo/pull/372). Since these options don't do anything, lockdown should **warn** when these options are specified.

Here is a PR for a breaking change that fully removes the options https://github.com/endojs/endo/pull/2583

### DIFF
--- a/packages/ses/src/lockdown.js
+++ b/packages/ses/src/lockdown.js
@@ -194,8 +194,8 @@ export const repairIntrinsics = (options = {}) => {
       'safe',
     ),
     __hardenTaming__ = getenv('LOCKDOWN_HARDEN_TAMING', 'safe'),
-    dateTaming = 'safe', // deprecated
-    mathTaming = 'safe', // deprecated
+    dateTaming, // deprecated
+    mathTaming, // deprecated
     ...extraOptions
   } = options;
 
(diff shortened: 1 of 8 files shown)


packages/ses/src/lockdown.js
```diff
@@ -188,11 +188,24 @@ export const repairIntrinsics = (options = {}) => {
...
-    dateTaming = 'safe', // deprecated
-    mathTaming = 'safe', // deprecated
+    dateTaming, // deprecated
+    mathTaming, // deprecated
     ...extraOptions
   } = options;
 
+  if (dateTaming !== undefined) {
+    // eslint-disable-next-line no-console
+    console.warn(
```
kriskowal, line 198: After rebase, this will need to be `const { warn } = reporter; …; warn(…)` since we added a layer of indirection for `lockdown({ reporting: 'platform' | 'console' | 'none' })`.

kriskowal (changes requested):
Thank you! Minor changes requested, and passing tests. Would not mind a nod from @erights, but we forgiveness-before-permission this one.

erights (approved):
> Would not mind a nod from @erights

Nod!

kriskowal (approved):

PR #2587, opened 2024-10-16
ci: fix browser-tests ubuntu 24.04 compat

legobeat:
## Description

- **ci: use npm-run instead of npx to run playwright**
- **ci: manually install playwright system dependencies**

### Security Considerations

N/A

### Scaling Considerations

N/A

### Testing Considerations

N/A

### Upgrade Considerations

N/A


.github/workflows/browser-test.yml
```diff
@@ -23,9 +24,20 @@ jobs:
...
       - name: Put unstable chrome where playwright would look for it
         run: mv /opt/google/chrome /opt/google/chrome-unstable
+      - name: Install browser dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
+            ca-certificates \
+            libatomic1 libflite1 libgles2 libharfbuzz-icu0 libhyphen0 libmanette-0.2-0 libicu74 libsecret-1-0 libxslt1.1 woff2 \
+            ca-certificates fonts-liberation libatk-bridge2.0-0 libatk1.0-0 libc6 libcairo2 libcups2 libdbus-1-3 libexpat1 libfontconfig1 libgbm1 libgcc1 libglib2.0-0 libgtk-3-0 libnspr4 libnss3 libpango-1.0-0 libpangocairo-1.0-0 libstdc++6 libx11-6 libx11-xcb1 libxcb1 libxcomposite1 libxcursor1 libxdamage1 libxext6 libxfixes3 libxi6 libxrandr2 libxrender1 libxss1 libxtst6 lsb-release wget xdg-utils \
+            libvpx9 libevent-2.1-7t64 libopus0 libgstreamer-plugins-base1.0-0 libgstreamer-gl1.0-0 libgstreamer-plugins-bad1.0-0 gstreamer1.0-libav
```
leotm, line 39: are all of these extra deps required? or can any be left out to be more minimal

leotm (approved):

.github/workflows/browser-test.yml
legobeat: They may not all be required. Most if not all were previously installed under the hood by the playwright util.

boneskull: - We should definitely install only the minimum needed
- This seems like a future maintenance burden
- Linebreaks seem inconsistent

.github/workflows/browser-test.yml
```diff
@@ -41,10 +53,10 @@ jobs:
         run: npm ci --ignore-scripts
       - name: Install Playwright Browsers
         working-directory: browser-test
-        run: npx playwright install --with-deps
+        run: npm exec playwright install --with-deps
```
boneskull, line 56: Doesn't `npx` do the same thing if `playwright` is local?

boneskull (commented):
I don't understand the motivation behind this change.  Can you update the PR description?

leotm:
> I don't understand the motivation behind this change. Can you update the PR description?

updated ^ it's to fix the recent failing Browser Tests happening in a few PRs

boneskull:
Is this still a problem?

.github/workflows/browser-test.yml
leotm: atm should be

```suggestion
        run: npm exec -- playwright install --with-deps
```

.github/workflows/browser-test.yml
turadg: Are you sure changes needed to the action runner? This looks like [a Playwright issue](https://github.com/microsoft/playwright/issues/30368) that was [fixed in 1.45](https://github.com/microsoft/playwright/releases/tag/v1.45.0).

leotm:
superseded currently by https://github.com/endojs/endo/pull/2686 i believe (the bump mentioned in https://github.com/endojs/endo/pull/2587#discussion_r1884215175)

naugtur:
> superseded currently by #2686 i believe (the bump mentioned in [#2587 (comment)](https://github.com/endojs/endo/pull/2587#discussion_r1884215175))

Yes, that was the minimal update needed to get it to work again.
We could consider pinning ubuntu version in case github starts A/B testing releases again - 24.04 is LTS so we're better off on it than updating to 25 whenever it becomes latest.

PR #2589, opened 2024-10-17
fix(ses-ava): wrap test deeply

mhofman:
Refs: https://github.com/Agoric/agoric-sdk/pull/10208

## Description

the `ava` test functions can be nested more than one level deep. This change recurses test functions with known names until none is found.

### Security Considerations

None

### Scaling Considerations

None

Relaxes the type of acceptable test functions as it only really cares about it being a test function, regardless of whatever chain expandos it may have.

### Testing Considerations

Manually verified that a `test.serial.only` does show errors details as expected. Not sure if we have any automated coverage for this.

None

### Upgrade Considerations

This is used by tests only, and can be consumed as a non breaking change simply by using the new NPM package including this fix.

### DIFF
--- a/packages/ses-ava/src/ses-ava-test.js
+++ b/packages/ses-ava/src/ses-ava-test.js
@@ -51,6 +51,7 @@ const {
   freeze,
   getPrototypeOf,
   getOwnPropertyDescriptors,
+  hasOwn,
   entries,
 } = Object;
 const { apply } = Reflect;
@@ -176,7 +177,7 @@ const makeVirtualExecutionContext = originalT => {
 };
 
 /**
- * @template {import('ava').TestFn} [T=import('ava').TestFn]
+ * @template {import('ava').OnlyFn} [T=import('ava').TestFn]
  * @param {T} testerFunc
  * @returns {T} Not yet frozen!
  */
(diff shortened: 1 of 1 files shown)


kriskowal (approved):
Neat.

packages/ses-ava/src/ses-ava-test.js
```diff
@@ -176,7 +177,7 @@ const makeVirtualExecutionContext = originalT => {
 };
 
 /**
- * @template {import('ava').TestFn} [T=import('ava').TestFn]
+ * @template {import('ava').OnlyFn} [T=import('ava').TestFn]
```
turadg, line 180: @mhofman why this change? it's breaking my Endo sync PR by making `wrapTest` erase the type:

Image

mhofman: This is 2 years old, and the type was working fine then. I think the reasoning was that this would accept any test fn, including possibly nested ones, not only the top level one that requires more expandos.

turadg: Yeah, it definitely worked. Maybe something in Ava 7 broke it.

PR #2597, opened 2024-10-21
fix(marshal)!: Update compareRank to terminate comparability at the first remotable

### LINKED ISSUE #2588: failure in elementsIsSuperset({x, y, z}, {z})
## Describe the bug

An ERTP withdraw of `{z}` from a purse with a balance of `{x, y, z}` failed.
 - https://github.com/Agoric/agoric-sdk/issues/10292

_I'm abbreviating `makeCopySet(...items)` using traditional `{...items}` set notation._

Investigation led to a failure in `elementsIsSuperset({x, y, z}, {z})`.

## Steps to reproduce

```js
test('{x, y, z} is superset of {z}', t => {
  const x = { instance: Remotable('Alleged: fooo'), description: 'fraz' };
  const y = { instance: Remotable('Alleged: fooo'), description: 'fraz' };
  const z = { instance: Remotable('Alleged: fooo'), description: 'alan' };

  const mainList = harden([x, y, z]);
  const withdraw = harden([z]);

  t.true(elementsIsSuperset(mainList, withdraw));
});
```

## Expected behavior

`elementsIsSuperset({x, y, z}, {z})` always returns true for keys `x`, `y`, and `z`.

## Platform environment

The original environment was [a ci log of an a3p test](https://github.com/Agoric/agoric-sdk/actions/runs/11365912632/job/31615121873?pr=10281).

It reproduces in 4406f5dde

## Additional context

 - https://github.com/Agoric/agoric-sdk/issues/10292

_some diagnosis to follow_

gibson042:
Fixes #2588

## Description

To be refinable into a total order that distinguishes remotables, `compareRank` must consider `[r1, 'x']` vs. `[r2, 'y']` as a tie rather than as equivalent to `[0, 'x']` vs. `[0, 'y']`, in case r1 vs. r2 ends up comparing differently than 'x' vs. 'y'.

### Security Considerations

None known.

### Scaling Considerations

None known.

This includes implicit fixup in types.js, but there may be some Markdown that needs updating as well.

### Testing Considerations

This expands property-based testing, but only narrowly. Expansion is expected in the future.

This affects storage of compound Keys in e.g. CopySets/CopyBags/CopyMaps, but any ordering written by the old `compareRank` should be accepted by the new one (which now reclassifies as ties what its predecessor considered to be distinct ranks). As such, we should be fine, but I welcome challenges to this assertion.

### Upgrade Considerations

This is breaking in the sense of changing `compareRank` output, but as far as I know, nothing depends upon that except in combination with `compareKeys` (which is fixed by this PR). I am not aware of any concerns that should hinder upgrade to a version of endo including this PR.

But I am wondering what if anything to put in NEWS.md file(s).


packages/marshal/src/rankOrder.js
```diff
@@ -101,15 +101,16 @@ const memoOfSorted = new WeakMap();
...
+ * It defaults to a function that always returns `NaN`, meaning that all
+ * remotables are incomparable and should tie for the same rank without further
+ * refinement (e.g., not only are `r1` and `r2` tied, but so are `[r1, 0]`
+ * and `[r2, "x"]`).
  * @returns {RankComparatorKit}
  */
-export const makeComparatorKit = (compareRemotables = (_x, _y) => 0) => {
-  /** @type {RankCompare} */
+export const makeComparatorKit = (compareRemotables = (_x, _y) => NaN) => {
+  /** @type {PartialCompare} */
```
kriskowal, line 113: Confirm or deny:

Previously:
- values from dispirit ranks are compared by the total order of ranks
- all values within atom ranks, e.g., numbers and strings, are mutually comparable
- all values within composite ranks, e.g., arrays, are mutually comparable provided
  - that their respective entries are all comparable, or
  - if they contain any respective pairs that are incomparable, those arrays have non-equivalent prefixes
  - and are otherwise incomparable
- within the rank of promises, all values are incomparable
- within the rank of remotables, all values are equivalent
- comparing incomparable values produces 0

After:
- within the rank of remotables, all values are incomparable
- comparing incomparable values produces NaN

kriskowal (commented):
Flushing this review before I get to the end, to keep the conversation lively.

packages/marshal/src/rankOrder.js
gibson042: That seems correct, although I would rephrase a bit:

Previously:
* …
* within the rank of remotables, all values are equivalent <ins>in the default comparison (used by `compareRank`)</ins>
* <del>comparing incomparable values produces 0</del> <ins>there is no concept of incomparable values within a rank (but encountering an unacceptable pass style such as "error" or "promise" results in a thrown exception)</ins>

After:
* within the rank of remotables, all values are incomparable
* comparing incomparable values produces NaN <ins>in the inner comparator and short-circuits the outer comparator result to 0</ins>

kriskowal: So, both before and after this change, comparing any pair of promises would by default return NaN?

Consequently, both before and after this change, the range of comparison was -1, 0, 1, and NaN?

Consequently, both before and after this change, rank compare is and was a partial order?

erights:
Just skimming, but saw the q at end. rankCompare is a full order. I cannot answer NaN. (Unless this PR is making a more radical change than I thought.

For all promises p,q, compareRank(p,q) === 0

packages/marshal/test/encodePassable.test.js
```diff
@@ -353,31 +337,58 @@ test('compact custom encoding validity constraints', t => {
...
+  const iter2 = str2[Symbol.iterator]();
+  let i = 0;
+  for (;;) {
+    const { value: char1 } = iter1.next();
+    const { value: char2 } = iter2.next();
+    if (char1 !== char2 || char1 === undefined) {
+      return str1.substring(0, i);
+    }
+    i += char1.length;
+  }
```
kriskowal, line 355: I’m sure you’re cognizant of the discrepancy here between iterating code points and indexing code units, but that might not be obvious under maintenance of these tests. Anything that compensates for that would be great, albeit accumulating common code points by concatenation or documenting a limitation to correctness under strings where code unit and code point are the same, but if documenting, consider using the dumber algorithm that co-iterates code units to be consistent with the substring call.

packages/marshal/src/rankOrder.js
```diff
@@ -301,7 +296,20 @@ export const sortByRank = (passables, compare) => {
...
+  if (compare(true, undefined) > 0) {
+    let i = sorted.length - 1;
+    while (i >= 0 && sorted[i] === undefined) i -= 1;
+    const n = sorted.length - i - 1;
+    if (n > 0 && n < sorted.length) {
+      sorted.copyWithin(n, 0);
+      sorted.fill(/** @type {T} */ (undefined), 0, n);
+    }
+  }
+  harden(sorted);
```
kriskowal, line 312: ✅ the undefined fix

packages/marshal/src/rankOrder.js
```diff
@@ -225,14 +225,18 @@ export const makeComparatorKit = (compareRemotables = (_x, _y) => 0) => {
   };
 
   /** @type {RankCompare} */
-  const antiComparator = (x, y) => comparator(y, x);
+  const outerComparator = (x, y) =>
+    /** @type {Exclude<PartialComparison, NaN>} */ (comparator(x, y) || 0);
```
kriskowal, line 229: ✅ narrowing the range in case of a compareRemotables that returns NaN, as it now will by default.

This does break any existing program that expects to distinguish NaN and provided a compareRemotables to produce that effect, but as you state you’re convinced nobody does that in practice, I’m fine with this. It may be worth noting in NEWS.md if only to build trust.

packages/marshal/src/rankOrder.js
kriskowal: I’m far enough along in my read to see with my own eyes that, both before and after this change:

- all exported comparators produce values in the range -1, 0, +1 _for the default comparator for remotables_.
- all promises are equivalent, not incoparable

packages/marshal/src/rankOrder.js
kriskowal: It might be worth explicitly reminding reviewers that `NaN || 0 === 0` because it’s not entirely obvious which JavaScript operators implicitly coerce down from the float to integer domain.

packages/patterns/test/copySet.test.js
```diff
@@ -119,6 +138,23 @@ test('operations', t => {
...
+    fc.property(
+      fc.uniqueArray(arbKey, { comparator: keyEQ }),
+      fc.infiniteStream(fc.boolean()),
+      (arr, keepSeq) => {
+        const sub = arr.filter(() => keepSeq.next().value);
+        setIsSuperset(makeCopySet(arr), makeCopySet(sub)) ||
+          Fail`${q(sub)} of ${q(arr)}`;
+      },
+    ),
+  );
```
kriskowal, line 152: I’m not a qualified reviewer for property tests yet.

kriskowal (approved):
I’m convinced about the correctness and practical forward-compatibility of the fix.

I note that the fix occurs almost entirely outside the diff as an emergent behavior of having `compareRemoatables` return NaN and coercing NaN back down to 0 at the border. The relevant behavior appears to be stopping early and propagating NaN on:

https://github.com/endojs/endo/blob/bf8b1f88759a0c0a1e3af0aa3911301645a4300d/packages/marshal/src/rankOrder.js#L199-L206

I am not qualified to read the property based tests yet.

gibson042:
> Just skimming, but saw the q at end. rankCompare is a full order. It cannot answer NaN. (Unless this PR is making a more radical change than I thought.

@erights Right, and that remains true... the default `compareRemotables` (i.e., the one used by `compareRank`) now returns NaN to avoid misclassification of e.g. `[r1, 'x']` vs. `[r2, 'y']` as being in different ranks (without disrupting such _correct_ classification of e.g. `[0, 'x']` vs. `[0, 'y']`), but that NaN is replaced with 0 on the way out—`compareRank` and indeed any function returned from `makeComparatorKit` continues to implement a total preorder in which all valid inputs are comparable but ties are possible.

Do you have further comments or reason(s) to delay merging now that @kriskowal has approved?

packages/patterns/src/types.js
```diff
@@ -143,61 +143,17 @@ export {};
  */
 
 /**
- * @typedef {RankCompare} FullCompare
```
dckc, line 146: Are static types considered part of the API? This `FullCompare` type is exported, right? So is removing it from this package a breaking change?

packages/patterns/test/copySet.test.js
dckc: I read this as: For an arbitrary array of distinct keys `arr` and subsequence `sub` of `arr`, `setIsSuperset(makeCopySet(arr), makeCopySet(sub))` holds, where a subsequence is formed by filtering using elements of an arbitrary infinite stream of booleans.

dckc: I thought it would be illustrative to show the output when it fails, before the fix is applied. But even when I check out 14fca7b49 that has the test and not the fix, this test seems to pass. I suppose the failing case is pretty obscure and the generator doesn't get that far?

When I crank `numRuns` up to 4000, I get a failure that might be showing another problem? `Got error: Error: ["[undefined]"] of ["[Symbol()]",null,"[undefined]"]`

dckc: oic... it's finding cases of the `undefined` problem.

erights:
> Do you have further comments or reason(s) to delay merging now that @kriskowal has approved?

I have not found time to review this in enough depth to approve it. But I saw no red flags and am satisfied from the conversations above that all the issues I thought to worry about have been extremely well thought through and reviewed. So please consider your existing approval adequate and do not wait for me. In this case, I'm happy to complete my review post merge.

Really looks great. A clever simple solution that I would not have thought of. Thanks much!!

packages/patterns/src/types.js
gibson042: Hmm, that's a good question. I've updated to reëxport FullCompare to preserve any remaining imports, on the assumption that dropping it _would_ be considered breaking.

PR #2601, opened 2024-10-21
docs(ses): Missing lockdown in Compartment doc

kriskowal:
This addresses a speed bump in our documentation of how to use SES, as seen at https://github.com/endojs/endo/issues/1281#issuecomment-2418717060


erights (approved):
LGTM

PR #2604, opened 2024-10-22
chore(patterns,pass-style): Recouple accidental 2.0.0 releases under 1.x train

kriskowal:
In https://github.com/endojs/endo/pull/2602, we inadvertently bumped the major versions for patterns and pass-style (while manually intervening in Lerna’s release process to avoid a major version bump for marshal)

In this change, we retroactively resume the 1.x train for these two packages. These two packages will be manually published to npm and assigned git tags.

### DIFF
--- a/packages/pass-style/CHANGELOG.md
+++ b/packages/pass-style/CHANGELOG.md
@@ -3,18 +3,13 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
-## [2.0.0](https://github.com/endojs/endo/compare/@endo/pass-style@1.4.4...@endo/pass-style@2.0.0) (2024-10-22)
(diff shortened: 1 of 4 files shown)


dckc (approved):
I would prefer to see an approval from someone else, but I do think this is a reasonable approach, and I'm willing to support it.

mhofman (approved):
From what I understand that'll do the trick, until we have the next breaking change, where lerna will try to use the `2.0.0` version. Let's make sure to keep the git tags in place to force an error in lerna when that happens, so we can likely manually bump to `2.0.1`. (without the git tags, lerna would likely only fail at NPM publish, which is arguably worse)

mhofman:
I might have approved too fast. Are these packages leafs or any workspace packages need their deps bumped accordingly?

kriskowal:
> I might have approved too fast. Are these packages leafs or any workspace packages need their deps bumped accordingly?

The workspace packages all use `workspace:^` for their `dependencies` and `devDependencies` within the workspace. But, Lerna automagically rolls those up, so all the published versions from this cohort are all probably pointing at the `^2` editions and need a bump. I think this means I need to use Lerna to do another full release.

PR #2605, opened 2024-10-22
chore: Release 2024-10-22 Supplemental

kriskowal:
In https://github.com/endojs/endo/pull/2602, we erroneously published 2.0 versions of pass-style and patterns. This release recouples these to their 1.x trains, and propagates that version selector to the dependent packages.

We are losing continuity of a published patch version number, but the computers and I are beyond caring.


dckc (approved):
whatever

PR #2610, opened 2024-10-23
fix(compartment-mapper): Defer all importHook errors when importing a…

naugtur:
… dependency if they're intended to be deferred

## Description

When importHook is used to load dependencies, they might be coming from `imports` that were identified by a parser that uses heuristics to find them. (e.g. cjs parser implementation uses a lexer to find all `require` calls even if they're not _the_ require function but a local require function)
In such cases we want to defer loading errors to runtime, so the error doesn't happen until actual `require` is _actually_ called to get it.

That was true for loading known dependency, but exit modules (via exitModuleImportHook) were not covered.

### Security Considerations

no changes security-wise

### Scaling Considerations

The way this is currently implemented - one big try-catch instead of returning the deferred error without throwing and catching, might be less performant. Should we consider trading aesthetics for a tiny performance gain here?
Note this only makes a difference for cases where non-dynamic `require` is called on a specifier that causes an error.

It's a fix of unintended behavior. No changes to what needs to be documented.

### Testing Considerations

Fix was added test-first

Compatibility is improved. Packages can now do:
```js
try {
  require('my-dev-dependency')
  } catch(){}
```
to silently attempt to load something that is only useful in their own development process. (typescript does that)

### Upgrade Considerations

While it's a fix, not a breaking change, it does change one specific behavior - an error resulting from calling exitModuleImportHook (the `importHook` passed to compartment-mapper, not the one in Compartment options) is now thrown at runtime, not during loading, which could be noteworthy for the Archive use-case.

[(insert mandatory xkcd about breaking changes)](https://xkcd.com/1172/)


packages/compartment-mapper/src/import-hook.js
```diff
@@ -502,89 +502,88 @@ export const makeImportHookMaker = (
...
-              compartmentDescriptor.policy,
-              attenuators,
-            );
-            return attenuatedRecord;
           }
-        }
-        return deferError(
-          moduleSpecifier,
-          Error(
+          throw Error(
```
naugtur, line 543: returning instead of throwing could be slightly more performant

packages/compartment-mapper/src/import-hook.js
```diff
@@ -502,89 +502,88 @@ export const makeImportHookMaker = (
...
             moduleSpecifier,
           )} (with candidates ${candidates
             .map(x => q(x))
             .join(', ')}) in package ${packageLocation}`,
-        ),
-      );
+        );
+      } catch (error) {
+        return deferError(moduleSpecifier, error);
+      }
```
boneskull, line 586: IMO this should probably just return the `deferError` instead of throwing and catching it immediately, but if there's a good reason for doing this then ok.

boneskull (approved):

packages/compartment-mapper/src/import-hook.js
naugtur: no good reason, just being DRY/consistent. Didn't feel good doing it.

kriskowal (changes requested):
Given that this would cause missing exits to be discovered much later, I would like this to be enabled with an option. Otherwise, this is well-motivated and I’m in favor.

boneskull:
@kriskowal Reconsider? This looks like a bugfix to me.

kriskowal (approved):
I see it now, thanks. More errors are caught, but those are released again for non-heuristic parsers. This only changes behavior for heuristic parsers.

kriskowal:
Consider adding to NEWS.

boneskull:
I've updated and added a blurb to `NEWS.md`.

packages/compartment-mapper/NEWS.md
```diff
@@ -4,6 +4,10 @@ User-visible changes to `@endo/compartment-mapper`:
 
 - Omits unused module descriptors from `compartment-map.json` in archived
   applications, potentially reducing file sizes.
+- Fixes an issue where errors thrown from exit module hooks (`importHook`) would
+  be thrown at parse-time instead of at runtime. Such errors will now be thrown
+  at runtime, as expected. To those who expected the previous behavior: if you
+  exist, please exercise caution when upgrading.
```
kriskowal, line 10: Should be qualified that the behavior only applies to parsers that have heuristic import analysis. The behavior for ESM should be to throw an error while loading if a static import dependency is unmet.

kriskowal (approved):
With request for qualification in NEWS.

packages/compartment-mapper/NEWS.md
boneskull: OK, I noted this

Issue #2608, opened 2024-10-23
Better platform/environment-dependent SES warning behavior


kriskowal:
## What is the Problem Being Solved?

Fixes: #2606 

SES warning behavior should:
1. On Node.js, be reported to `stderr` by any means. The Node.js `console.log`, `.info`, `.group`, and `.groupCollapsed` report to `stdout`. Grouping is not collapsible, so indenting on `.error` or `.warn` are indistinguishable.
2. On the Web, a collapsed group is useful and safe to fold up SES errors, and error levels are visually distinct.
3. XS only provides `print` and doesn’t seem to provide stderr IO without more work.
4. The default behavior should be platform dependent
5. A lockdown option or environment variable should be able to override the detected platform for the corresponding behavior
6. Or opt-out

## Description of the Design

I would like to thread a logging capability and ban direct use of the console. The logging capability probably does not need to concern grouping or levels, but can be a `{ log }` bag in case we need more. We should avoid dynamic dispatch.

## Security Considerations

None anticipated

## Scaling Considerations

None

## Test Plan

Existing tests sift stdout and stderr. A test must be contrived that forces a repair warning.

## Compatibility Considerations

Programs that use Lockdown should have sovereignty over their output channel. SES should not interfere.

PR #2613, opened 2024-10-24
feat(ses): Lockdown reporting option

kriskowal:
Closes: #2608

## Description

This change introduces a `"reporting"` option to `lockdown` and `repairIntrinsics` that determines the means by which SES will send warnings to diagnostic tools like the web or node `console`, `print`, or nowhere at all.

### Security Considerations

None

### Scaling Considerations

None

Relevant documentation added to lockdown options and NEWS.

### Testing Considerations

This adjusts the existing test to use the `"reporting": "console"` option, since it already verifies that behavior and consequently is suitable for verifying the web behavior while standing on Node.js.

This adds a new test that verifies that reporting with the `"platform"` default behavior on node generates no output on stderr and confirms the presence of expected indented and non-indented messages, while being resilient to additional intrinsics being added to the platform.

None.

### Upgrade Considerations

None.

### DIFF
--- a/packages/ses/src/lockdown.js
+++ b/packages/ses/src/lockdown.js
@@ -163,22 +163,24 @@ export const repairIntrinsics = (options = {}) => {
 
   const {
     errorTaming = getenv('LOCKDOWN_ERROR_TAMING', 'safe'),
-    errorTrapping = /** @type {"platform" | "none" | "report" | "abort" | "exit" | undefined} */ (
+    errorTrapping = /** @type {"platform" | "none" | "report" | "abort" | "exit"} */ (
       getenv('LOCKDOWN_ERROR_TRAPPING', 'platform')
     ),
     reporting = /** @type {"platform" | "console" | "none"} */ (
       getenv('LOCKDOWN_REPORTING', 'platform')
     ),
-    unhandledRejectionTrapping = /** @type {"none" | "report" | undefined} */ (
+    unhandledRejectionTrapping = /** @type {"none" | "report"} */ (
       getenv('LOCKDOWN_UNHANDLED_REJECTION_TRAPPING', 'report')
     ),
     regExpTaming = getenv('LOCKDOWN_REGEXP_TAMING', 'safe'),
     localeTaming = getenv('LOCKDOWN_LOCALE_TAMING', 'safe'),
 
-    consoleTaming = /** @type {'unsafe' | 'safe' | undefined} */ (
+    consoleTaming = /** @type {'unsafe' | 'safe'} */ (
       getenv('LOCKDOWN_CONSOLE_TAMING', 'safe')
     ),
-    overrideTaming = getenv('LOCKDOWN_OVERRIDE_TAMING', 'moderate'),
+    overrideTaming = /** @type {'moderate' | 'min' | 'severe'} */ (
+      getenv('LOCKDOWN_OVERRIDE_TAMING', 'moderate')
+    ),
     stackFiltering = getenv('LOCKDOWN_STACK_FILTERING', 'concise'),
     domainTaming = getenv('LOCKDOWN_DOMAIN_TAMING', 'safe'),
     evalTaming = getenv('LOCKDOWN_EVAL_TAMING', 'safeEval'),
--- a/packages/ses/src/lockdown.js
+++ b/packages/ses/src/lockdown.js
@@ -58,6 +58,7 @@ import { tameSymbolConstructor } from './tame-symbol-constructor.js';
 import { tameFauxDataProperties } from './tame-faux-data-properties.js';
 import { tameRegeneratorRuntime } from './tame-regenerator-runtime.js';
 import { shimArrayBufferTransfer } from './shim-arraybuffer-transfer.js';
+import { reportInGroup, chooseReporter } from './reporting.js';
 
 /** @import {LockdownOptions} from '../types.js' */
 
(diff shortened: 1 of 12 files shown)


packages/ses/src/permits-intrinsics.js
```diff
@@ -303,32 +298,22 @@ export default function whitelistIntrinsics(
...
-    visitProperties('intrinsics', intrinsics, permitted);
-  } finally {
-    if (groupStarted) {
-      // eslint-disable-next-line @endo/no-polymorphic-call
-      console.groupEnd();
-    }
-  }
+  // Start path with 'intrinsics' to clarify that properties are not
+  // removed from the global object by the whitelisting operation.
+  visitProperties('intrinsics', intrinsics, permitted);
```
erights, line 318: Why do we no longer need the try/finally? Or did we already not need it?

erights (changes requested):
Review not finished, but wanted to get these in front of you first.

packages/ses/src/console.js
```diff
@@ -0,0 +1,102 @@
+import { globalThis, TypeError } from './commons.js';
```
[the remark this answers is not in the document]
kriskowal: I’m inclined to rename this `reporting` rather than bundle it with console taming. I will probably continue to call them consoles, since a real console implements the needed interface. Sound good or no?

packages/ses/src/console.js
```diff
@@ -0,0 +1,102 @@
...
+    error(...args) {
+      indentLog(...args);
+    },
+    groupCollapsed(...args) {
+      indentLog(...args);
+      indent = true;
+    },
+    groupEnd() {
+      indent = false;
+    },
```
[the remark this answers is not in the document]
kriskowal: Given that this console is for internal use and we only ever indent once, I’m inclined to keep this simple and add an error check. I left the string manipulation on the shop floor since it’s weird and unnecessary. But, it looks like accumulating two spaces for each level, and always slicing one off for an indented log since the Node.js console itself always adds one space between arguments. In my opinion, it’s too weird to implement given that it’ll never be used.

packages/ses/src/permits-intrinsics.js
kriskowal: The `try/finally` moved into a utility function in `console.js` (to be renamed `reporting.js`, since it now gets used for this and separately enablements warnings, under a different group label.

Since try/finally deopts the surrounding function, this has the added benefit of deopting a smaller function.

packages/ses/src/lockdown.js
```diff
@@ -58,8 +58,10 @@ import { tameSymbolConstructor } from './tame-symbol-constructor.js';
 import { tameFauxDataProperties } from './tame-faux-data-properties.js';
 import { tameRegeneratorRuntime } from './tame-regenerator-runtime.js';
 import { shimArrayBufferTransfer } from './shim-arraybuffer-transfer.js';
+import { reportInGroup, chooseReporter } from './reporting.js';
 
 /** @import {LockdownOptions} from '../types.js' */
+/** @import {ExternalConsole} from './console-types.js' */
```
erights, line 64: I believe this line is vestigial and should simply be deleted.

```suggestion
```

packages/ses/src/reporting.js
```diff
@@ -0,0 +1,104 @@
...
+  if (
+    reporting === 'console' ||
+    globalThis.window === globalThis ||
+    globalThis.importScripts !== undefined
+  ) {
+    return console;
+  }
+  if (globalThis.console !== undefined) {
+    // On Node.js, we send all feedback to stderr, regardless of purported level.
+    return makeReportPrinter(globalThis.console.error);
```
erights, line 66: I'm just a bit surprised that `console.error` is thisless, i.e., that it doesn't need to be bound to `console` first. I tried it, and that is indeed the case in Node. Curious though if it is specified that way, by those who purport to standardize `console`?

packages/ses/src/reporting.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+  return makeReportPrinter(mute);
+};
+
+/**
+ * @param {string} groupLabel
+ * @param {GroupReporter} console
+ * @param {(internalConsole: Reporter) => void} callback
+ */
+export const reportInGroup = (groupLabel, console, callback) => {
```
erights, line 79: While I would agree it is overkill to assert that the `groupLabel` is the same every time, if you don't assert it, you should at least mention in the doc-comment that only the first `groupLabel` in a group matters, and any inconsistency between them is ignored.

erights (approved):
LGTM, thanks!

packages/ses/src/console.js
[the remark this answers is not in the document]
erights: (as we then later agreed verbally, the "them" we decided to call Reporters, where a Console is a structural subtype of Reporter)

packages/ses/src/reporting.js
kriskowal: Back in Safari 2, `console.error.call(null)` caused a segmentation fault for the whole browser.

packages/ses/src/reporting.js
kriskowal: I don’t grok this comment, but I can more rigorously describe `reportInGroup` above. Suffice it to say that it does not compose better than the underlying `console` group behavior and is very sensitive to context. We only ever have one reporter active on the stack at any given time, either on the stack of `repairIntrinsics` or `hardenIntrinsics` (for enablements).

packages/ses/src/reporting.js
erights: WOW! Please report. From previous history, I know they'll be interested.

packages/ses/src/reporting.js
erights: I was indeed confused. Resolving.

packages/ses/src/reporting.js
erights: Oh, Safari 2! Nevermind. Resolving.

Issue #2313, opened 2024-06-10
feat(compartment-mapper): Collect unused module descriptors


kriskowal:
## What is the Problem Being Solved?

When building an archive, the compartment mapper currently collects unused compartment descriptors, but the increased prevalence of package exports has created a lot of unused module descriptors.

## Description of the Design

 We should mark and sweep these too, as we do with the compartment “retained” flag.

## Security Considerations

None.

## Scaling Considerations

Should reduce bundle size.

## Compatibility Considerations

May frustrate hashes of existing applications when they upgrade bundle-source in their package lock. I hope we no longer support system-level agoric CLI since upgrading that will frustrate hashes without opt-in.

## Upgrade Considerations

None.

PR #2614, opened 2024-10-24
feat(compartment-mapper): Collect unused module descriptors

kriskowal:
Closes: #2313

## Description

This change adds instrumentation to the compartment mapper that allows it to omit module descriptors from archive compartment maps if they’re not needed to link the archived application.

### Security Considerations

None.

### Scaling Considerations

Should reduce bundle sizes.

Does not merit attention and one less surprise for bundle size investigators. Noted in NEWS.

### Testing Considerations

This change augments the existing test for compartment retention to cover module descriptor collection.

The change does not add the "retained" property to archived `compartment-map.json`, so new bundles should remain compatible with older versions of `checkBundle`. However, I relaxed the schema validator to get an unrelated test for `captureCompartmentMap` passing, so future deployments of `checkBundle` will be more lenient.

### Upgrade Considerations

Should not affect upgrades.

### DIFF
--- a/packages/compartment-mapper/NEWS.md
+++ b/packages/compartment-mapper/NEWS.md
@@ -1,5 +1,10 @@
 User-visible changes to `@endo/compartment-mapper`:
 
+# Next version
+
+- Omits unused module descriptors from `compartment-map.json` in archived
+  applications, potentially reducing file sizes.
+
 # v1.3.0 (2024-10-10)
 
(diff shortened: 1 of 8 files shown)


warner (approved):
Looks good to me. My understanding is that the generated archives will have exactly the same files as before (those are already minimal), but the compartment-map.json metadata will omit some linkage fields for unused exports, which are likely to be more common as we become more explicit about what each package exports (to discourage deep-imports).

PR #2615, opened 2024-10-24
misc tightening of linting

turadg:
incidental

## Description

Adopting `verbatimModuleSyntax` in agoric-sdk required a lot of changes so it thought it worthwhile to adopt here before more might be necessary.

I made several other lint tooling bumps and config tightening as well.

### Security Considerations

none

### Scaling Considerations

none

none

### Testing Considerations

CI

none

### Upgrade Considerations

none, dev deps


boneskull (approved):
There's a lint warning in the build, but that looks like an `eslint-plugin-jsdoc` bug to me.  workaround would be to rename the `array` parameter to something else

turadg:
> There's a lint warning in the build, but that looks like an `eslint-plugin-jsdoc` bug to me. workaround would be to rename the `array` parameter to something else

Yeah, I agree it's a bug. It was there before this PR so I'm going to overlook it.

> Unclear if we should split the dependency upgrades into a separate PR

I separated all the commits. What would be the advantage of also separating PRs? Having a different set of reviewers, but that's not necessary in this case.

PR #2622, opened 2024-11-04
refactor(compartment-mapper): Move types from jsdoc to d.ts

kriskowal:
This refactor converts the JSDoc types in Compartment Mapper with TypeScript definitions. It also takes liberty to use our new trick for omitting types from bundles, but leaving continuity through importing types from JavaScript sources.

Much of this conversion was automatic, except `internal.d.ts` and `external.d.ts` are manually converted. Much of the motivation for this change was to take advantage of the expressivity of `.d.ts` over jsdoc typescript documents, allowing option groups to extend each other without otherwise useless intermediate names. The motivation for that consolidation, in turn, was to ensure that certain symmetries are preserved. For example, execute options are threaded through all functions that ultimately execute. I was looking specifically at the `languageForExtensions` option, which we misplaced on the `import` path when it belonged down the `mapNodeModulesPath`. That fix comes separately.


packages/compartment-mapper/src/archive-lite.js
```diff
@@ -306,7 +310,7 @@ export const makeArchiveCompartmentMap = (compartmentMap, sources) => {
 /**
  * @param {ReadFn | ReadPowers} powers
  * @param {CompartmentMapDescriptor} compartmentMap
- * @param {ArchiveOptions} [options]
+ * @param {ArchiveLiteOptions} [options]
```
kriskowal, line 313: This is a bugfix. The type for ArchiveLiteOptions is narrower.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -726,11 +727,9 @@ export function makeImportNowHookMaker(
...
 
       if (exitModuleImportNowHook) {
-        // this hook is responsible for ensuring that the moduleSpecifier actually refers to an exit module
-        const exitRecord = exitModuleImportNowHook(
-          moduleSpecifier,
-          packageLocation,
-        );
+        // This hook is responsible for ensuring that the moduleSpecifier
+        // actually refers to an exit module.
+        const exitRecord = exitModuleImportNowHook(moduleSpecifier);
```
kriskowal, line 732: Bugfix revealed by typing.

packages/compartment-mapper/src/map-parser.js
```diff
@@ -19,14 +16,17 @@ import { parseExtension } from './extension.js';
  *   ModuleTransform,
  *   ModuleTransforms,
  *   ParseFn,
- *   ParseFnAsync,
+ *   AsyncParseFn,
```
kriskowal, line 19: I’m choosing to treat this as non-breaking since nothing in practice relies on these type names outside of the compartment mapper, although parsers are part of the public API. We tend to treat them as opaque handles and very much doubt there are parsers implemented outside Compartment Mapper.

packages/compartment-mapper/src/map-parser.js
```diff
@@ -19,14 +16,17 @@ import { parseExtension } from './extension.js';
...
  *   ParseResult,
  *   ParserForLanguage,
  *   SyncModuleTransform,
  *   SyncModuleTransforms
  * } from './types.js';
  */
 
+import { syncTrampoline, asyncTrampoline } from '@endo/trampoline';
+import { parseExtension } from './extension.js';
```
kriskowal, line 28: This moved in a fit of homogenization of the order of type import comments relative to imports.

packages/compartment-mapper/src/types/compartment-map-schema.d.ts
```diff
@@ -0,0 +1,116 @@
...
+export type ModuleDescriptor = {
+  compartment?: string | undefined;
+  module?: string | undefined;
+  location?: string | undefined;
+  parser?: Language | undefined;
+  /** in base 16, hex */
+  sha512?: string | undefined;
+  exit?: string | undefined;
+  deferredError?: string | undefined;
+  retained?: boolean | undefined;
```
kriskowal, line 76: Nit: tsc generated redundant `| undefined` here and I missed them in my first pass. Circle back.

packages/compartment-mapper/src/types/external.d.ts
```diff
@@ -0,0 +1,352 @@
...
+   */
+  dev: boolean;
+  /** Dependencies to make reachable from any package */
+  commonDependencies: Record<string, string>;
+  /** Maps extensions to languages for all packages, like `txt` to `text` */
+  languageForExtension: LanguageForExtension;
+  /** Maps extensions to languages for all type=module packages */
+  moduleLanguageForExtension: LanguageForExtension;
+  /** Maps extensions to languages for all type=commonjs packages (default) */
+  commonjsLanguageForExtension: LanguageForExtension;
```
kriskowal, line 67: Nit: this feature appears later in this stack.

packages/compartment-mapper/src/types/external.d.ts
```diff
@@ -0,0 +1,352 @@
...
+
+type SyncImportingOptions = ModulesOption &
+  SyncOrAsyncImportingOptions &
+  ExitModuleImportNowHookOption;
+
+type LinkingOptions = ParserForLanguageOption &
+  CompartmentOption &
+  LanguageForExtensionOption &
+  SyncModuleTransformsOption &
+  ModuleTransformsOption;
```
kriskowal, line 198: Much of what I want to achieve here is sensible groupings of options here, and to ensure that all of these options get threaded to their constituent subroutines. For example, `languageForExtension` should be threaded from `loadLocation`, `makeArchive`, and such all the way down to `mapNodeModules` where it’s used.

packages/compartment-mapper/src/types/external.d.ts
```diff
@@ -0,0 +1,352 @@
...
+};
+
+/**
+ * The compartment mapper can capture the Sources for all loaded modules
+ * for bundling, archiving, or analysis.
+ */
+export type Sources = Record<string, CompartmentSources>;
+export type CompartmentSources = Record<string, ModuleSource>;
+
+// TODO unionize:
```
kriskowal, line 218: For a future change, moving from jsdoc substantially improves our chances of turning the `ModuleSource` type from a `Partial` to a union of all valid combinations. That’ll probably complicate the code, but improve our understanding of what properties are expected on module sources of various shapes. They may need a vestigial discriminating property.

packages/compartment-mapper/src/types/node-powers.d.ts
```diff
@@ -0,0 +1,51 @@
...
+  createHash: (algorithm: 'sha512') => Hash;
+};
+
+/**
+ * Object returned by function in {@link CryptoInterface}
+ */
+type Hash = {
+  update: (data: Uint8Array | string) => Hash;
+  digest: () => {
+    toString: (radix: 'hex') => string;
```
kriskowal, line 49: This was `Buffer`, but we just need this one behavior and `Buffer` was hard to capture from Node.js types otherwise.

turadg: good choice. Buffer type has been changing in TS versions lately.

worth a comment tho

packages/compartment-mapper/src/types.d.ts
```diff
@@ -0,0 +1,11 @@
+// ESlint does not understand export *
+/* eslint-disable import/export */
+
+export * from './types/external.js';
```
turadg, line 4: Since you have this indirection, all the modules below can be `.ts`. Moreover they should be, because they are pure types rather than annotations about other `.js` files.

It would also prevent ignoring them when `skipLibCheck: true`

packages/compartment-mapper/src/import-hook.js
kriskowal: @boneskull I’ve posted a new commit to this PR that fixes the typing bug but not as reflected in this outdated comment. It turns out that `exitModuleImportHook` does require the second argument but did not provide it in all cases. So, the new commit includes the second argument in the hook type and threads it through the various call sites. Please take a look.

boneskull:
Types may be omitted from generated `.d.ts` files by using the `@internal` tag and adding `stripInternal: true` to `tsconfig.json`.

packages/compartment-mapper/src/import-archive-lite.js
```diff
@@ -405,7 +401,7 @@ export const parseArchive = async (
     await pendingJobsPromise;
 
     // eslint-disable-next-line dot-notation
-    return compartment['import'](moduleSpecifier);
+    return compartment['import'](entryModuleSpecifier);
```
boneskull, line 404: I assume this is for evasion

packages/compartment-mapper/src/link.js
```diff
@@ -19,13 +20,12 @@
  *   LinkResult,
  *   ModuleDescriptor,
  *   ParseFn,
- *   ParseFnAsync,
  *   ParserForLanguage,
  *   ParserImplementation,
  *   ShouldDeferError,
  * } from './types.js'
+ * @import {ERef} from '@endo/eventual-send'
```
boneskull, line 27: If the type consuming `ERef` is present in an _external_ type, then `@endop/eventual-send` it should be a production dependency of `@endo/compartment-mapper`.  If not, it _should_ be added as a dev dependency anyway, but might be OK to just leave it as-is given we're in a monorepo.

packages/compartment-mapper/src/map-parser.js
boneskull: I can live with it

packages/compartment-mapper/src/node-powers.js
```diff
@@ -138,10 +144,10 @@ const makeReadPowersSloppy = ({
         const realPath = await fs.promises.realpath(
           fileURLToPath(location).replace(/\/$/, ''),
         );
-        return `${pathToFileURL(realPath)}/`;
+        return `${pathToFileURL(realPath).href}/`;
       } else {
         const realPath = await fs.promises.realpath(fileURLToPath(location));
-        return pathToFileURL(realPath).toString();
+        return pathToFileURL(realPath).href;
```
boneskull, line 150: So I am curious about this.  I know that we (who?) settled on using `.href` to convert a `URL to a `string`.

`URL.prototype.toString()` just returns the `href` prop.  The advantage I see to using `toString()` (or wrapping in a template literal) is that `href` is writable and might be a source of subtle bugs (e.g., accidental assignment).

packages/compartment-mapper/src/search.js
```diff
@@ -4,9 +4,13 @@
...
-/** @import {ReadFn} from './types.js' */
-/** @import {ReadPowers} from './types.js' */
-/** @import {MaybeReadPowers} from './types.js' */
+/**
+ * @import {
+ *   ReadFn,
+ *   ReadPowers,
+ *   MaybeReadPowers,
+ * } from './types.js'
+ */
```
boneskull, line 13: I notice that in all of our `@import` blocks, we do not end each statement with a semicolon.  I am unsure if this is considered inconsistent given the presence of semicolons everywhere else

packages/compartment-mapper/src/types.d.ts
```diff
@@ -0,0 +1,11 @@
...
+/* eslint-disable import/export */
+
+export * from './types/external.ts';
+export * from './types/internal.ts';
+export * from './types/policy-schema.ts';
+export * from './types/compartment-map-schema.ts';
+export * from './types/policy.ts';
+export * from './types/powers.ts';
+export * from './types/node-powers.ts';
+export * from './types/typescript.ts';
```
boneskull, line 11: These should all end in `.js` afaik?

packages/compartment-mapper/src/types/compartment-map-schema.d.ts
boneskull, line 1: - We should probably normalize comment grammar
- None of these `.d.ts` files are type-checked, whereas `types.js` _was_ type-checked. 
 I don't like that. This can be resolved in one of two ways:
  - Set `skipLibCheck` to `false` (please don't)
  - Rename each `.d.ts` file to `.ts`. Associated `.d.ts` files will be generated for each at the same time as the `.js` files
- We should settle on `@file` vs `@module` vs `@packageDocumentation`; I think TypeDoc prefers one of the latter two

packages/compartment-mapper/src/types/external.d.ts
```diff
@@ -0,0 +1,350 @@
...
+export type SyncArchiveLiteOptions = SyncOrAsyncArchiveOptions &
+  SyncModuleTransformsOption &
+  SyncImportingOptions &
+  ExitModuleImportNowHookOption;
+
+export type ArchiveOptions = Omit<MapNodeModulesOptions, 'language'> &
+  ArchiveLiteOptions;
+
+export type SyncArchiveOptions = Omit<MapNodeModulesOptions, 'languages'> &
+  SyncArchiveLiteOptions;
```
boneskull, line 100: one of these is wrong

packages/compartment-mapper/src/types/external.d.ts
```diff
@@ -0,0 +1,350 @@
...
+type SearchSuffixesOption = {
+  /**
+   * Suffixes to search if the unmodified specifier is not found. Pass `[]` to
+   * emulate Node.js' strict behavior. The default handles Node.js' CommonJS
+   * behavior.
+   * Unlike Node.js, the Compartment Mapper lifts CommonJS up, more like a
+   * bundler, and does not attempt to vary the behavior of resolution depending
+   * on the language of the importing module.
+   */
+  searchSuffixes?: string[];
```
boneskull, line 127: `string[]` or `Array<string>`???

packages/compartment-mapper/src/types/external.d.ts
```diff
@@ -0,0 +1,350 @@
...
+  params: {
+    sourceMap?: string;
+  },
+];
+
+type ModuleTransformResult = {
+  bytes: Uint8Array;
+  parser: Language;
+  sourceMap?: string;
+};
```
boneskull, line 268: I think all of these should be exported, since the `moduleTransforms` etc. options are exposed

packages/compartment-mapper/src/types/external.d.ts
```diff
@@ -0,0 +1,350 @@
...
+  moduleSpecifier: string,
+  sourceLocation: string,
+) => void;
+
+export type ComputeSourceMapLocationHook = (details: {
+  compartment: string;
+  module: string;
+  location: string;
+  sha512: string;
+}) => string;
```
boneskull, line 304: Extract the options object to its own type?

packages/compartment-mapper/src/types/internal.d.ts
```diff
@@ -0,0 +1,218 @@
...
+}) => ImportHook;
+
+export type ImportNowHookMaker = (params: {
+  packageLocation: string;
+  packageName: string;
+  parse: ParseFn | AsyncParseFn;
+  compartments: Record<string, Compartment>;
+  // TODO why are these absent? Why this asymmetry?
+  // attenuators: DeferredAttenuatorsProvider,
+  // shouldDeferError: ShouldDeferError,
```
boneskull, line 92: they are ignored/unused

packages/compartment-mapper/src/types/node-powers.d.ts
```diff
@@ -0,0 +1,52 @@
+/**
+ * @file These interfaces describe the powers needed in `node-powers.js` to
+ * adapt host capabilities for the compartment mapper.
+ */
+
+/* eslint-disable no-use-before-define */
+
+/** For creating {@link ReadPowers} */
```
boneskull, line 8:
```suggestion
/** For creating `ReadPowers` */
```

The `@link` is now broken. There may be a few broken `@link`s in these files

turadg (approved):
🚀

packages/compartment-mapper/src/types/policy.d.ts
```diff
@@ -0,0 +1,42 @@
...
+  params: Params,
+  originalObject: Record<PropertyKey, any>,
+  globalThis: Record<PropertyKey, any>,
+) => void;
+
+export type ModuleAttenuatorFn<
+  Params extends [any, ...any[]] = [any, ...any[]],
+  T = SomeObject,
+  U = T,
+> = (params: Params, ns: T) => U;
```
boneskull, line 38: I noticed that all of these `X extends [any, ...any[]] = [any, ...any[]]` should be `X extends readonly [any, ...any[]] = [unknown, ...unknown[]]`.

Also, if it makes sense that e.g., a global attenuator function accepts no parameters whatsoever, then that should be `X extends readonly any[] = unknown[]`.

packages/compartment-mapper/src/types/powers.d.ts
```diff
@@ -0,0 +1,120 @@
...
+ * These are the platform-neutral types for those capabilities.
+ * For example, {@file node-powers.js} adapts Node.js how modules
+ * to various subsets of these capabilities.
+ */
+
+/* eslint-disable no-use-before-define */
+
+import { SomeObject } from './typescript.ts';
+
+// Read
```
boneskull, line 12: do audiobooks count?

packages/compartment-mapper/src/types/typescript.d.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * @file Helpers relevant to any TypeScript project.
```
boneskull, line 2: there might be a copyright/license notice we need to include from [type-fest](https://npm.im/type-fest), as these types were copied from it.

boneskull (commented):
I love this direction, as it's easier to define these types in TS sources. The problem -- and it's a sticking point for me -- is that the `.d.ts` files are not type-checked.

boneskull (changes requested):
I love this direction, as it's easier to define these types in TS sources. The problem -- and it's a sticking point for me -- is that the `.d.ts` files are not type-checked.  

**UPDATE:** I guess the `@endo/eventual-send` thing is kind of a sticking point as well.

packages/compartment-mapper/src/import-archive-lite.js
kriskowal: It was, though that is not necessary any more. It predates some added sophistication to the evasive transforms.

packages/compartment-mapper/src/node-powers.js
kriskowal: No decision has been made and this change is not necessary. I started with `toString()` but have been leaning in the direction of `href`. There isn’t a strong case either way.

packages/compartment-mapper/src/search.js
kriskowal: My understanding is that `@` pragmas are not generally terminated with a semicolon so I assumed they wouldn’t even be tolerated. Would have to check. My preference is to omit them given that they are not necessary.

packages/compartment-mapper/src/types.d.ts
kriskowal: It works either way and I don’t have a strong preference. I originally wrote these with `.js` and @turadg recommended `.ts`. I don’t have a principled position to stand on either way.

packages/compartment-mapper/src/types/compartment-map-schema.d.ts
kriskowal: These correspond more closely to `@module`. I expect `@packageDocumentation` to only exist once in a package. I had `@overview` and eslint suggested `@file`.

I’ll take a look.

kriskowal:
> I love this direction, as it's easier to define these types in TS sources. The problem -- and it's a sticking point for me -- is that the `.d.ts` files are not type-checked.

Is there some other way to enforce that these hand-written type definitions must contain only type definitions?

boneskull:
For type-checking, I'd strongly recommend just renaming `.d.ts` files to `.ts` files.

Using `skipLibCheck: false` makes us responsible for the correctness of the types of our dependencies--not just our own. To solve problems, we might have to do things like:
- Introduce a new dependency to fix a "missing" type only
- "Missing" types may have no actual impact on our own type fidelity (as we're not guaranteed to be actually _using_ any of those types)
- "Missing" types may have no actual impact on the type fidelity of our users (and probably won't)
- Incorrect types may need to be fixed by using module augmentation
- Incorrect types may need to be addressed via `patch-package` (or whatever `yarn` does here) . Those fixes will not cascade to our consumers unless we use lifecycle scripts

A concern is that using `.d.ts` files guarantees that the files contain _only_ types. That is virtuous--but there are other ways to enforce this. I can think of:

- It's likely ESLint can be bent in this direction (someone probably has had this problem already)
- Use `export type * from './foo'` in a barrel `index.d.ts` may work; this barrel won't be typechecked, but everything else would, I think (haven't tried this)
- PR reviews 😄 

We can also use [emitDeclarationsOnly](https://www.typescriptlang.org/tsconfig/#emitDeclarationOnly) to ensure no `.js` files are ever generated. This won't _enforce_ that the `.ts` files remain execution-free, but it will prevent compilation to JS.

boneskull:
Also, how else are we going to export `enum`s?  Ha ha ha.

packages/compartment-mapper/src/types/internal.d.ts
boneskull: ...and maybe we should consider adding `attenuators: never, shouldDeferError: never` to this type, just so whoever is using it cannot expect any behavior based on these two props

boneskull:
AFAIK, by default, _all_ extensions present in _any_ `.js`, `.ts`, or `.d.ts` file should be `.js` (unless we're using `.mjs`/`.cjs`). There's a `tsconfig.json` option to change this, but I haven't tried using it and don't understand the use-case.

kriskowal:
We have `emitDeclarationOnly` in the relevant config, so I will take that as sufficient. I can also keep the `.d.ts` for the barrel type exports files.

So, for @turadg’s and @boneskull’s consideration, I propose that I will:
1. rename `src/types/*.d.ts` to `src/types/*.ts`.
2. readdress `*.ts` to `*.js` in all `*.ts` and `*.d.ts`.

Please reactji.

packages/compartment-mapper/src/types.d.ts
kriskowal: I’ve changed all of these to `.js`.

packages/compartment-mapper/src/types.d.ts
kriskowal: For the record, we reversed this decision out of band. We are enforcing the limit with the build configuration and I’ve renamed these `.ts` so that `tsc` will complain about errors in them, per yours and @boneskull’s observation. The barrel `types.d.ts` (external) and `src/types.d.ts` (internal)  remain in `.d.ts`.

packages/compartment-mapper/src/types/compartment-map-schema.d.ts
kriskowal: Added a commit that converges all the existing `@file` and `@fileoverview` pragmas on `@module`.

packages/compartment-mapper/src/types/external.d.ts
kriskowal: I will change this to `Array` because that’s the prevailing convention, but really want to ignore this discrepancy unless there’s a meaningful difference and a lint rule to enforce it.

packages/compartment-mapper/src/types/internal.d.ts
kriskowal: I’m patching this as advised, and documenting the deliberate discrepancy, but is there a principled argument why they are not respected, to the extent that it would not make sense to add behavior for these properties as a future feature, or are they omitted for lack of a motivating use case?

packages/compartment-mapper/src/types/node-powers.d.ts
kriskowal: Did a holistic pass changing `@link` to bare words or ticked words if the referent was in a different file. Push forthcoming.

packages/compartment-mapper/src/types/policy.d.ts
kriskowal: I’m going to leave this for a subsequent fix since the recommended fix did not trivially pass and is beyond my ken.

packages/compartment-mapper/src/types/typescript.d.ts
kriskowal: Thank you for calling this out. `type-fest` is CC0, so no attribution is required. I’m inclined to add a citation regardless, but doesn’t need a copy of the license.

kriskowal:
This is ready for another pass @boneskull. I believe I’ve addressed all your feedback except for the bugfix you note for policy interfaces, which I would ask you to pursue in a follow-up since it’s orthogonal, delicate, and out of my wheelhouse.

packages/compartment-mapper/src/types.d.ts
mhofman: I still see the specifiers here using `.ts`

packages/compartment-mapper/types.d.ts
mhofman, line 1: This seems like a duplicate of `types-index.d.ts`. How do you propose to keep them in sync under maintenance?

I'm actually wondering if this file shouldn't just go away given that we export types through the runtime entrypoint now.

packages/compartment-mapper/src/types.d.ts
mhofman, line 1: This seems like a superset of `../types-index.d.ts`. Should we look into making that explicit somehow by having the internal file reexport the external file? (they would likely have to be moved around)

packages/compartment-mapper/src/types.js
mhofman, line 1: So we're positive all internal usages of `types.js` are through type imports and never runtime imports?

mhofman (commented):
I only reviewed the organization of the type files.

packages/compartment-mapper/types.d.ts
kriskowal: `./src/types.d.ts` exports everything under `./src/types/*.ts` whereas `./types.d.ts` does not export the types in `internal` or `typescript` since they do not participate in the public API. I _could_ eliminate the internal barrel and break up the type imports. I would like to defer that to a future refactor.

kriskowal: `./src/types.d.ts` exports everything under `./src/types/*.ts` whereas `./types.d.ts` does not export the types in `internal` or `typescript` since they do not participate in the public API. I _could_ eliminate the internal barrel and break up the type imports. I would like to defer that to a future refactor.

packages/compartment-mapper/src/types.js
kriskowal: I’m pretty sure that is the case. Can I delete this stub?

kriskowal: Yes.

packages/compartment-mapper/src/types.d.ts
kriskowal: This conversation is tied to an outdated delta.

packages/compartment-mapper/src/types/internal.d.ts
boneskull: I _thought_ I understood why `shouldDeferError` was omitted, but now I'm not so sure.  I _think_ it's essentially "too late" to defer an error.  @naugtur, can you explain this (again)?  

I'm assuming `DeferredAttenuatorsProvider` is not provided because a) there's only one attenuator compartment and b) that compartment is loaded via `importHook` instead of `importNowHook`.

packages/compartment-mapper/src/types/policy.d.ts
boneskull: I've put this on my List.

boneskull (approved):
LGTM. I've confirmed this works as linked against `@lavamoat/node`.  ship it

packages/compartment-mapper/package.json
```diff
@@ -21,10 +21,8 @@
   },
   "type": "module",
   "main": "./index.js",
-  "types": "./types.d.ts",
   "exports": {
     ".": {
-      "types": "./types.d.ts",
       "default": "./index.js"
     },
```
mhofman, line 27: Technically this can be simplified, no?

```suggestion
    ".": "./index.js",
```

mhofman (approved):
LGTM

PR #2624, opened 2024-11-04
fix(ses): fix #2598 with cauterizeProperty reuse

### LINKED ISSUE #2598: Function prototype set to undefined inconsistently too late
## Describe the bug

On Hermes we never reach (in whitelistIntrinsics)

https://github.com/endojs/endo/blob/b3a31b1559118dcfcd3811e9632c3f28fba99a3c/packages/ses/src/permits-intrinsics.js#L303-L304

but reach here first instead

https://github.com/endojs/endo/blob/b3a31b1559118dcfcd3811e9632c3f28fba99a3c/packages/ses/src/intrinsics.js#L87-L103

order of events and conditions

- intrinsics.js > completePrototypes
  - `isObject(intrinsic)`
  - `objectHasOwnProperty(intrinsic, 'prototype')`
  - `typeof permit === 'object'`
  - `!permit.prototype`
    - ```throw TypeError(`${name}.prototype property not whitelisted`);```
- permit-intrinsics.js > whitelistIntrinsics
  - `prop in obj`
  - `typeof obj === 'function' && prop === 'prototype'`
    - `obj.prototype = undefined` (not reached)
    - #1221 

## Steps to reproduce

- https://github.com/endojs/endo/pull/2334
  - `yarn build:hermes && yarn test:hermes`

## Expected behavior

The prototype to be set to `undefined` earlier
so we `continue` which skips throwing the TypeError

## Platform environment

- macOS Sequoia 15.0.1
- Node v20.12.2
- Hermes v0.12.0 (lockdown), v0.13.0 (harden)
- @endo/base64@1.0.8-10-gb3a31b155

## Additional context

- https://github.com/endojs/endo/pull/2563#discussion_r1807041929

Note: on [Static Hermes](https://speakerdeck.com/tmikov2023/hermes-better-performance-with-bytecode-translation-react-universe-2024?slide=9) this is a non-issue, since [both](https://github.com/endojs/endo/pull/2563#discussion_r1811111552) prototype bugs are fixed

## Screenshots

lockdown prototype property not whitelisted

erights:
Closes: #2598 
Refs: https://github.com/endojs/endo/pull/2563 https://github.com/endojs/endo/pull/2334 https://github.com/endojs/endo/pull/1221

## Description

#1221 was supposed to make ses tolerate undeletable `func.prototype` properties that should be absent, so long as they could be set to `undefined` instead, making them harmless. This tolerance came with a warning to flag the remaining non-conformance.

However #2598 explains why #1221 sometimes fails to do this. #1221 did come with a test, but it fell into the case where #1221 works, which is a non-toplevel function.

#2563 (and #2334 ?) fell into the trap explained by #2598 and untested by #1221, which is an undeletable `func.prototype` on a top-level instrinsic. As a result, #2563 currently contains a workaround for #2598 which this PR would make unnecessary.

This PR fixes the problem by factoring out the `func.prototype`-tolerant property deletion into a separate `cauterizeProperty` function which it calls from both places. This PR also adds the test that was missing from #1221 , having first checked that the test detects #2598 when run without the rest of this PR.

If this PR gets merged before #2563, then #2563's workaround for #2598 can first be removed before it is merged.

- [ ] TODO should pass a genuine reporter in to all calls to `cauterizeProperty`. @kriskowal , please advise how intrinsics.js should arrange to do so.

### Security Considerations

Allowing a `func.prototype` property that really shouldn't be there seems safe, so long as it is safely set to `undefined` first, which this PR does, and then checks that it has done so.

### Scaling Considerations

none

generally, this would be one less thing to worry about, and thus one less thing that needs to be documented for most users.

### Testing Considerations

Adds the test that was missing from #1221 that let #2598 go unnoticed until #2563

Should be none.

### Upgrade Considerations

Should be none.

### DIFF
--- a/packages/ses/src/cauterize-property.js
+++ b/packages/ses/src/cauterize-property.js
@@ -1,0 +1,51 @@
+import { objectHasOwnProperty } from './commons.js';
+
+/**
+ * @import {Reporter} from './reporting-types.js'
+ */
+
+/**
+ *
+ * @param {object} obj
+ * @param {PropertyKey} prop
+ * @param {boolean} known
+ * @param {string} subPath
+ * @param {Reporter} reporter
+ * @returns {void}
+ */
+export const cauterizeProperty = (
+  obj,
+  prop,
+  known,
+  subPath,
+  { warn, error },
+) => {
+  // Either the object lacks a permit or the object doesn't match the
+  // permit.
+  // If the permit is specifically false, not merely undefined,
+  // this is a property we expect to see because we know it exists in
+  // some environments and we have expressly decided to exclude it.
+  // Any other disallowed property is one we have not audited and we log
+  // that we are removing it so we know to look into it, as happens when
+  // the language evolves new features to existing intrinsics.
+  if (!known) {
+    warn(`Removing ${subPath}`);
+  }
+  try {
+    delete obj[prop];
+  } catch (err) {
+    if (objectHasOwnProperty(obj, prop)) {
+      if (typeof obj === 'function' && prop === 'prototype') {
+        obj.prototype = undefined;
+        if (obj.prototype === undefined) {
+          warn(`Tolerating undeletable ${subPath} === undefined`);
+          return;
+        }
+      }
+      error(`failed to delete ${subPath}`, err);
+    } else {
+      error(`deleting ${subPath} threw`, err);
+    }
+    throw err;
+  }
+};
--- a/packages/ses/src/cauterize-property.js
+++ b/packages/ses/src/cauterize-property.js
@@ -5,12 +5,30 @@ import { objectHasOwnProperty } from './commons.js';
  */
 
 /**
+ * Delete `obj[prop]` or at least make it harmless.
+ *
+ * If the property was not expected, then emit a reporter-dependent warning
+ * to bring attention to this case, so someone can determine what to do with it.
+ *
+ * If the property to be deleted is a function's `.prototype` property, this
+ * will normally be because the function was supposed to be a
+ * - builtin method or non-constructor function
+ * = arrow function
+ * - concise method
+ *
+ * all of whom are not supposed to have a `.prototype` property. Nevertheless,
+ * on some platforms (like older versions of Hermes), or as a result of
+ * some shim-based mods to the primordials (like core-js?), some of these
+ * functions may accidentally be more like `function` functions with
+ * an undeletable `.prototype` property. In add these case, if we can
+ * set the value of that bogus `.prototype` property to `undefined`,
+ * we do so, issuing a warning, rather than failing to initialize ses.
  *
  * @param {object} obj
  * @param {PropertyKey} prop
- * @param {boolean} known
- * @param {string} subPath
- * @param {Reporter} reporter
+ * @param {boolean} known If deletion is expected, don't warn
+ * @param {string} subPath Used for warning messages
+ * @param {Reporter} reporter Where to issue warning or error.
  * @returns {void}
  */
 export const cauterizeProperty = (
--- a/packages/ses/src/cauterize-property.js
+++ b/packages/ses/src/cauterize-property.js
@@ -13,14 +13,14 @@ import { objectHasOwnProperty } from './commons.js';
  * If the property to be deleted is a function's `.prototype` property, this
  * will normally be because the function was supposed to be a
  * - builtin method or non-constructor function
- * = arrow function
+ * - arrow function
  * - concise method
  *
  * all of whom are not supposed to have a `.prototype` property. Nevertheless,
  * on some platforms (like older versions of Hermes), or as a result of
  * some shim-based mods to the primordials (like core-js?), some of these
  * functions may accidentally be more like `function` functions with
- * an undeletable `.prototype` property. In add these case, if we can
+ * an undeletable `.prototype` property. In these cases, if we can
  * set the value of that bogus `.prototype` property to `undefined`,
  * we do so, issuing a warning, rather than failing to initialize ses.
  *
--- a/packages/ses/src/intrinsics.js
+++ b/packages/ses/src/intrinsics.js
@@ -77,9 +77,9 @@ function sampleGlobals(globalObject, newPropertyNames) {
 }
 
 /**
- * @param {Reporter} [reporter] defaults to `console`
+ * @param {Reporter} reporter
  */
-export const makeIntrinsicsCollector = (reporter = console) => {
+export const makeIntrinsicsCollector = reporter => {
   /** @type {Record<any, any>} */
   const intrinsics = create(null);
   let pseudoNatives;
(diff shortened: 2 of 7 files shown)


packages/ses/src/intrinsics.js
```diff
@@ -100,7 +101,15 @@ export const makeIntrinsicsCollector = () => {
       }
       const namePrototype = permit.prototype;
       if (!namePrototype) {
-        throw TypeError(`${name}.prototype property not permitted`);
+        cauterizeProperty(
+          intrinsic,
+          'prototype',
+          false,
+          `${name}.prototype`,
+          console, // TODO should be a proper Reporter
```
erights, line 109: @kriskowal , please advise on how to arrange to pass a proper Reporter here.

packages/ses/src/cauterize-property.js
```diff
@@ -0,0 +1,69 @@
...
+ * Delete `obj[prop]` or at least make it harmless.
+ *
+ * If the property was not expected, then emit a reporter-dependent warning
+ * to bring attention to this case, so someone can determine what to do with it.
+ *
+ * If the property to be deleted is a function's `.prototype` property, this
+ * will normally be because the function was supposed to be a
+ * - builtin method or non-constructor function
+ * = arrow function
+ * - concise method
```
gibson042, line 17:
```suggestion
 * If the property to be deleted is a function's `.prototype` property, this
 * will normally be because the function was supposed to be a
 * - builtin method or non-constructor function
 * - arrow function
 * - concise method
```

packages/ses/src/cauterize-property.js
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * all of whom are not supposed to have a `.prototype` property. Nevertheless,
+ * on some platforms (like older versions of Hermes), or as a result of
+ * some shim-based mods to the primordials (like core-js?), some of these
+ * functions may accidentally be more like `function` functions with
+ * an undeletable `.prototype` property. In add these case, if we can
+ * set the value of that bogus `.prototype` property to `undefined`,
+ * we do so, issuing a warning, rather than failing to initialize ses.
```
gibson042, line 25:
```suggestion
 * an undeletable `.prototype` property. In these cases, if we can
 * set the value of that bogus `.prototype` property to `undefined`,
 * we do so, issuing a warning, rather than failing to initialize ses.
```

packages/ses/src/cauterize-property.js
erights: Done

packages/ses/src/cauterize-property.js
erights: Done

packages/ses/src/cauterize-property.js
```diff
@@ -0,0 +1,69 @@
...
+  // the language evolves new features to existing intrinsics.
+  if (!known) {
+    warn(`Removing ${subPath}`);
+  }
+  try {
+    delete obj[prop];
+  } catch (err) {
+    if (objectHasOwnProperty(obj, prop)) {
+      if (typeof obj === 'function' && prop === 'prototype') {
+        obj.prototype = undefined;
```
mhofman, line 57: I assume this can throw too but we rely on that throw to interrupt the flow?

erights: It can throw, and it is fine if that throw interrupts the flow. But we don't rely on that in the sense that if we did suppress the throw and the value were not `undefined` afterwards, we'd still get an error report and the original throw. So I think it is fine as is, but let me know if you think it could be improved.

erights: Resolving.

mhofman: The only thing I can think of is to catch the error and annotate it. But yes regardless the behavior is correct.

packages/ses/src/intrinsics.js
erights: @kriskowal I threaded a proper reporter to here in the obvious way. But now I have the same mystery in compartment-shim.js

packages/ses/src/intrinsics.js
```diff
@@ -166,6 +182,7 @@ export const makeIntrinsicsCollector = () => {
  * @param {object} globalObject
  */
 export const getGlobalIntrinsics = globalObject => {
+  // TODO pass a proper reporter to `makeIntrinsicsCollector`
```
kriskowal, line 185: Threading `reporter` through here does seem appropriate to me. A way to avoid it would be to inject the intrinsics collector instead, and construct that in `lockdown` in the presence of the reporter.

packages/ses/src/cauterize-property.js
```diff
@@ -0,0 +1,69 @@
+import { objectHasOwnProperty } from './commons.js';
+
+/**
+ * @import {Reporter} from './reporting-types.js'
+ */
+
+/**
+ * Delete `obj[prop]` or at least make it harmless.
```
kriskowal, line 8: I do not have the details paged into active brain right now, but this looks like it might be relevant to a mitigation @leotm needs for Hermes support.

kriskowal (approved):
I have recommended options for threading the reporter, and any of those options are fine by me: we can either thread the `reporter` through all callers of `getGlobalIntrinsics`, or construct the intrinsics collector from the `reporter` available in `lockdown` and inject that instead.

packages/ses/src/cauterize-property.js
erights: @leotm , care to comment? Thanks.

packages/ses/src/intrinsics.js
erights: What about the call to `getGlobalIntrinsics` in compartment-shim.js? That's the problematic one.

packages/ses/src/permits-intrinsics.js
```diff
@@ -279,35 +280,10 @@ export default function removeUnpermittedIntrinsics(
...
-            error(`failed to delete ${subPath}`, err);
-          } else {
-            error(`deleting ${subPath} threw`, err);
-          }
-          throw err;
-        }
+        cauterizeProperty(obj, prop, subPermit === false, subPath, {
+          warn,
+          error,
+        });
```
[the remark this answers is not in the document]
erights: Done

packages/ses/src/intrinsics.js
erights: Done this much. But moved an equally broken `console` reference into compartment.shim.js' call to `getGlobalIntrinsics`, so still unsolved.

packages/ses/src/cauterize-property.js
erights: https://github.com/endojs/endo/pull/2563 (and https://github.com/endojs/endo/pull/2334 ?), right?

packages/ses/src/intrinsics.js
kriskowal: I think it might be sufficient to default `reporter` to `muteReporter` in `getIntrinsics`, because I assume we do not want to warn every time a `Compartment` gets constructed, and only when constructing intrinsics in service to `lockdown`. Does that sound right?

erights: > Does that sound right?

Yes. Done. Thanks!

packages/ses/src/cauterize-property.js
naugtur: MetaMask folks reporting back after a travel/vacation/conference period. Sorry for missing the notifications on this.
This does look like it addresses the issue we worked around. We'll rebase and report back.

leotm: sorry I missed this at Devcon then ill coming back (almost recovered)

just confirming this fixed _completePrototypes_ on Hermes after testing 🎉 thank you! closed https://github.com/endojs/endo/pull/2563 with the results

https://github.com/endojs/endo/pull/2334 still needed for CI feedback keeping SES compatible with Hermes
(after rebasing `master` five failing tests left before ready last for final re-review)

Issue #2415, opened 2024-08-17
support .ts in bundle-source


turadg:
## What is the Problem Being Solved?

TypeScript is very popular and the JsDoc syntax for it has many limitations compared to .ts. (Some described in https://github.com/Agoric/agoric-sdk/issues/5760) 

A developer who wants to write in `.ts` can transpile before bundling with Endo, but that increases work, room for error, and the transformation chain to audit. 

## Description of the Design

bundle-source would bundle `.ts` files directly by transparently transforming them to .js using something like https://bloomberg.github.io/ts-blank-space/ or `@swc/wasm-typescript` used in Node's new .ts support.

## Security Considerations

Transforming code can make it harder to audit the bundled code. But conversely, type annotations can mislead in an audit.

Given the complexities of audits and the trade-offs with runtime constraints, solving auditing will require purpose-built tooling: https://github.com/endojs/endo/issues/1656

mhofman:
Striping TS types and purely replacing with whitespace may cause semantic JS difference, and some cases need special handling. Any mechanism that support TS in bundle source need to correctly handle these cases and not change the runtime semantics.

Some pointers:
- [Effects of types on ASI](https://gist.github.com/acutmore/27444a9dbfa515a10b25f0d4707b5ea2)
- [TS syntax can also be valid JS syntax](https://github.com/microsoft/TypeScript/issues/33639), in particular the [3rd comment](https://github.com/microsoft/TypeScript/issues/33639#issuecomment-2291568328) showing that experimental type erasure in Node.js results in different runtime semantics.

kungfooman:
Pure ESM has so many advantages... just being able to do `<script type="module" src="..."></script>` with an import map... TS destroys nice ESM and also makes code unreadable. Simple one liner JS functions that are easy to document with JSDoc become type monsters where you can't even differ between real code and type sugar any longer without starring at ASCII soup for 5 minutes.

For what? The beginning of the end.

turadg:
This in TS 5.7 might help,
- https://github.com/microsoft/TypeScript/issues/59597

And this may be a better option than the WASM:
- https://bloomberg.github.io/ts-blank-space/

PR #2625, opened 2024-11-07
feat(compartment-mapper): Thread commonjs- and module- languageForExtension options (for TS)

kriskowal:
Refs: #2415

## Description

Toward support for type-erasure style TypeScript support in Endo, `.ts` files may correspond to `.mts` or `.cts` behavior in the same way that `.js` can correspond to either `.mjs` or `.cjs`, depending on `"type":` in the enveloping `package.json` (which is not necessarily the enveloping package, e.g., `src/package.json` beneath `package.json`!). We already implement this machinery for JavaScript, so this change threads additional options beside `languageForExtension` (which would map `.mts` and `.cts` to language behaviors) but also `commonjsLanguageForExtension` and `moduleLanguageForExtension`, which would get folded into the language-for-extension mapping on a package-by-package basis.

### Security Considerations

None. Absent from this design is any possibility that a file would be alternately implemented as TypeScript or JavaScript.

### Scaling Considerations

None.

Includes an update to README for the new features and the undocumented features it is based upon. Includes a mention in NEWS.

### Testing Considerations

TODO

By design error, `languageForExtension` threaded through `link` before this change, when it should have been threaded through `mapNodeModules`, since that is the machine that is aware of the `type` field in `package.json`. This change moves that internal mechanism and then adds `commonjsLanguageForExtension` and `moduleLanguageForExtension`. By common usage, this move is transparent since the option is taken at the top and simply diverted internally.

However, the `languageForExtension` property could be manually threaded through `makeArchiveLite` and omitted from the options for `mapNodeModules`. I’ve elected to treat this breaking change as a bug fix since I find it unlikely anyone has taken advantage of these relatively new “lite” functions. The “lite” functions are part of the refactoring necessary to approach native XS compartment and modules.

### Upgrade Considerations

None.


packages/compartment-mapper/test/language-for-extension.test.js
```diff
@@ -0,0 +1,111 @@
...
+import jsonp from './_parse-jsonp.js';
+import text from '../src/parse-text.js';
+
+// The JSONP parser uses harden, as a bit.
+lockdown({
+  errorTaming: 'unsafe',
+  errorTrapping: 'none',
+});
+
+scaffold(
```
kriskowal, line 13: For reviewers new to Compartment Mapper, `scaffold` generates a combination of test cases where a single fixture is expected to produce the same behavior for multiple flows:

- executed directly off a file system
- loaded off the file system then later executed
- archived off the file system, then parsed, then executed
&c

packages/compartment-mapper/src/node-modules.js
```diff
@@ -189,83 +196,73 @@ const findPackage = async (readDescriptor, canonical, directory, name) => {
...
-  'cjs',
-  'json',
-  'text',
-  'bytes',
-]);
-const defaultUncontroversialParsers = /** @type {const} */ ({
-  cjs: 'cjs',
+const defaultLanguageForExtension = /** @type {const} */ ({
   mjs: 'mjs',
+  cjs: 'cjs',
```
turadg, line 203: consider alphabetizing or leaving where it was

packages/compartment-mapper/src/node-modules.js
```diff
@@ -189,83 +198,83 @@ const findPackage = async (readDescriptor, canonical, directory, name) => {
...
+    packageLanguageForExtension === null
+  ) {
+    throw Error(
+      `Cannot interpret parser map ${JSON.stringify(
+        packageLanguageForExtension,
+      )} of package at ${location}, must be an object mapping file extensions to corresponding languages (for example, mjs for ECMAScript modules, cjs for CommonJS modules, or json for JSON modules`,
+    );
+  }
+  const invalidLanguages = values(packageLanguageForExtension).filter(
+    language => !languages.has(language),
```
turadg, line 256: `language` and `languages` look very similar. consider,
```suggestion
    l => !languages.has(l),
```

turadg (approved):
LGTM!

I'm new to this package so I don't have an opinion on the design decisions. The code changes all look good.

PR #2627, opened 2024-11-09
feat(bundle-source): Support TypeScript type erasure

kriskowal:
Closes: #2415 

## Description

This change introduces support for TypeScript through type-erasure, using ts-blank-space, which converts type annotations to equivalent blank space. As is consistent with `node --experimental-strip-types`, this only applies to modules with the `.ts`, `.mts`, or `.cts` extensions in packages that are not under `node_modules`, to discourage publishing TypeScript as a source language to npm.

### Security Considerations

The choice of `ts-blank-space` is intended to minimize runtime behavior difference between TypeScript and JavaScript, such that a reviewer or a debugger of the generated JavaScript aligns with the expected behavior and original text, to the extent that is possible. This should compose well with #2444.

### Scaling Considerations

None.

Contains README and NEWS.

### Testing Considerations

Contains spot check tests for TypeScript in the endoScript and endoZipBase64 formats. We stand on much more rigorous testing of the underlying workspace-language-for-extension feature in  Compartment Mapper #2625.

This does not break any prior usage.

### Upgrade Considerations

None.

### DIFF
--- a/packages/bundle-source/README.md
+++ b/packages/bundle-source/README.md
@@ -57,6 +57,25 @@ with `@preserve`, `@copyright`, `@license` pragmas or the Internet Explorer
 Comment elision does not strip comments entirely.
 The syntax to begin or end comments remains.
 
+## TypeScript type erasure
+
+TypeScript modules with the `.ts`, `.mts`, and `.cts` extensions in
+packages that are not under a `node_modules` directory are automatically
+converted to JavaScript through type erasure using
+[`ts-blank-space`](https://bloomberg.github.io/ts-blank-space/).
+
+This will not function for packages that are published as their original
+TypeScript sources, as is consistent with `node
+--experimental-strip-types`.
+This will also not function properly for TypeScript modules that have
+[runtime impacting syntax](https://github.com/bloomberg/ts-blank-space/blob/main/docs/unsupported_syntax.md),
+such as `enum`.
+
+This also does not support importing a `.ts` file using the corresponding
+imaginary, generated module with a `.js` extension.
+Use this feature in conjunction with
+[`--allowImportingTsExtensions`](https://www.typescriptlang.org/tsconfig/#allowImportingTsExtensions).
+
 ## Source maps
 
 With the `moduleFormat` of `endoZipBase64`, the bundler can generate source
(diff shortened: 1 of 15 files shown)


packages/bundle-source/README.md
```diff
@@ -57,6 +57,19 @@ with `@preserve`, `@copyright`, `@license` pragmas or the Internet Explorer
...
+TypeScript modules with the `.ts`, `.mts`, and `.cts` extensions in
+packages that are not under a `node_modules` directory are automatically
+converted to JavaScript through type erasure using
+[`ts-blank-space`](https://bloomberg.github.io/ts-blank-space/).
+
+This will not function for packages that are published as their original
+TypeScript sources, as is consistent with `node
+--experimental-transform-types`.
+This will also not function properly for TypeScript modules that generate
+JavaScript runtime code, like enums.
```
turadg, line 71: "enums" aren't necessarily TypeScript. (JSDoc has `@enum`). 
```suggestion
This will also not function properly for TypeScript modules that have [runtime impacting syntax](https://github.com/bloomberg/ts-blank-space/blob/main/docs/unsupported_syntax.md), such as `enum`.
```

packages/bundle-source/test/endo-script-format.test.js
```diff
@@ -4,28 +4,43 @@ import test from '@endo/ses-ava/prepare-endo.js';
...
-  const bundle = await generate();
+  const bundle = await generate('../demo/meaning.js');
   t.is(bundle.moduleFormat, 'endoScript');
   const { source } = bundle;
   const compartment = new Compartment();
   const ns = compartment.evaluate(source);
   t.is(ns.meaning, 42);
 });
 
+test('endo script format supports typescript type erasure', async t => {
```
turadg, line 29: I think we need more test cases to help make clear what to expect.

IIUC this these work:
```ts
// tsFromTs.ts
export { fortune } from './fortune.ts
```

```ts
// tsFromJs.ts
export { meaning } from './meaning.js
```

And this won't,
```ts
// jsFromTs.js
export { fortune } from './fortune.ts
```

kriskowal: I’ve posted tests alongside these that verify that the full cross-product of {js,ts} can import and export {js,ts}, and also an expected failing test that you can’t `import 'fortune.js'` in TypeScript, if there is no `fortune.js`, even if there _is_ `fortune.ts`.

turadg (approved):
🫶

PR #2639, opened 2024-11-20
feat(ses,module-source): Dynamic import

kriskowal:
Closes: #291

## Description

This change adds support for dynamic `import` to `ses` and `@endo/module-source`, such that `import(x)` in the scope of a module loaded from source (using `ModuleSource` in a `Compartment` module loading hook).

### Security Considerations

This change builds on prior work to ensure that dynamic import is facilitated by a hidden lexical name injected by `ses`. The dynamic import mechanism is isolated to the surrounding compartment and specifiers are resolved on the surrounding module.

### Scaling Considerations

This change introduces a static analysis for the presence of a dynamic `import` in the module source, which it communicates on the module source, even if marshaled through JSON, to `ses` that it should create an `import` closure bound to the calling module’s base specifier. This avoids an allocation for most modules.

Only news included. Dynamic import is a language feature that is expected to work in general and documented where JavaScript is documented as a language.

### Testing Considerations

This change includes a minimal happy path test that verifies that dynamic import produces a promise that settles on a module namespace object. Note that dynamic import does not box namespaces regardless of the `__noNamespaceBox__` compartment option.

Backward compatible.

### Upgrade Considerations

None.

### DIFF
--- a/packages/compartment-mapper/test/optional.test.js
+++ b/packages/compartment-mapper/test/optional.test.js
@@ -15,7 +15,8 @@ const fixtureOptionalDepsCjs = new URL(
 
 scaffold(
   'optionalDependencies/esm',
-  // this test fails because it relies on dynamic import
+  // fails for archives because dynamic import cannot reach modules not
+  // discovered during archival
   test,
   fixtureOptionalDepsEsm,
   async (t, { namespace }) => {
@@ -41,7 +42,7 @@ scaffold(
     );
   },
   4,
-  { knownFailure: true },
+  { knownArchiveFailure: true },
 );
 
 scaffold(
--- a/packages/module-source/NEWS.md
+++ b/packages/module-source/NEWS.md
@@ -2,6 +2,10 @@ User-visible changes in `@endo/module-source`:
 
 # Next release
 
+- Supports dynamic `import` within a `ModuleSource` in conjunction with
+  a related change in `ses`.
+  For example, `await import(specifier)` can now call through to the
+  surrounding compartment's `importHook` to load and evaluate further modules.
 - Provides an XS-specific variant of `@endo/module-source` that adapts the
   native `ModuleSource` instead of entraining Babel.
 
--- a/packages/ses/NEWS.md
+++ b/packages/ses/NEWS.md
@@ -2,7 +2,11 @@ User-visible changes in `ses`:
 
 # Next version
 
-- Specifying the long discontinued `mathTaming` or `dateTaming` options logs a warning.
+- Adds support for dynamic `import` in conjunction with an update to
+  `@endo/module-source`.
+
+- Specifying the long-discontinued `mathTaming` or `dateTaming` options logs a
+  warning.
 
 # v1.10.0 (2024-11-13)
 
--- a/packages/ses/src/compartment.js
+++ b/packages/ses/src/compartment.js
@@ -19,7 +19,7 @@ import {
   setGlobalObjectMutableProperties,
   setGlobalObjectEvaluators,
 } from './global-object.js';
-import { assert, assertEqual } from './error/assert.js';
+import { assert, assertEqual, q } from './error/assert.js';
 import { sharedGlobalPropertyNames } from './permits.js';
 import { load, loadNow } from './module-load.js';
 import { link } from './module-link.js';
@@ -27,7 +27,7 @@ import { getDeferredExports } from './module-proxy.js';
 import { compartmentEvaluate } from './compartment-evaluate.js';
 import { makeSafeEvaluator } from './make-safe-evaluator.js';
 
-/** @import {ModuleDescriptor} from '../types.js' */
+/** @import {ModuleDescriptor, ModuleExportsNamespace} from '../types.js' */
 
 // moduleAliases associates every public module exports namespace with its
 // corresponding compartment and specifier so they can be used to link modules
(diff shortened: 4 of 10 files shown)


packages/compartment-mapper/test/optional.test.js
```diff
@@ -41,7 +42,7 @@ scaffold(
     );
   },
   4,
-  { knownFailure: true },
+  { knownArchiveFailure: true },
```
naugtur, line 45: (not a polite way to say you should, just a note)
If you want to assert for both behaviors you could pass an assertion to scaffold that uses `testCategoryHint` to decide what to assert instead of this.

packages/compartment-mapper/test/optional.test.js
```diff
@@ -15,7 +15,8 @@ const fixtureOptionalDepsCjs = new URL(
 
 scaffold(
   'optionalDependencies/esm',
-  // this test fails because it relies on dynamic import
+  // fails for archives because dynamic import cannot reach modules not
+  // discovered during archival
```
boneskull, line 19: does dynamic require have the same limitation?

packages/module-source/NEWS.md
```diff
@@ -2,6 +2,8 @@ User-visible changes in `@endo/module-source`:
 
 # Next release
 
+- Supports dynamic `import` within a `ModuleSource` in conjunction with
+  a related change in `ses`.
```
boneskull, line 6:
```suggestion
- Supports dynamic `import` (i.e. `await import(specifier)`) within a `ModuleSource` in conjunction with
  a related change in `ses`.
```

packages/ses/NEWS.md
```diff
@@ -2,7 +2,11 @@ User-visible changes in `ses`:
 
 # Next version
 
-- Specifying the long discontinued `mathTaming` or `dateTaming` options logs a warning.
+- Adds support for dynamic `import` in conjunction with an update to
+  `@endo/module-source`.
+
+- Specifying the long discontinued `mathTaming` or `dateTaming` options logs a
```
boneskull, line 8:
```suggestion
- Specifying the long-discontinued `mathTaming` or `dateTaming` options logs a
```

packages/ses/src/compartment.js
```diff
@@ -297,6 +297,26 @@ export const makeCompartmentConstructor = (
 
     assign(globalObject, endowments);
 
+    /**
+     * @param {string} fullSpecifier
+     */
```
boneskull, line 302: - What is a `fullSpecifier`? 
- Please add description of function
- Please add return type of function

packages/ses/test/import.test.js
```diff
@@ -600,3 +601,21 @@ test('importMetaHook and meta from record', async t => {
   const { default: metaurl } = await compartment.import('./index.js');
   t.is(metaurl, 'https://example.com/index.js?foo');
 });
+
+test('dynamic import from source', async t => {
+  const c = new Compartment({
+    __options__: true,
+    __noNamespaceBox__: true,
```
boneskull, line 608: Out of curiosity: wtf are these?

packages/ses/test/import.test.js
```diff
@@ -600,3 +601,21 @@ test('importMetaHook and meta from record', async t => {
...
+    resolveHook: s => s,
+    modules: {
+      '-': {
```
boneskull, line 611: Does this have a specific meaning, or could it have been `foo`?

packages/ses/src/module-instance.js
```diff
@@ -171,6 +173,13 @@ export const makeModuleInstance = (
     importMetaHook(moduleSpecifier, importMeta);
   }
 
+  let dynamicImport;
```
boneskull, line 176: needs `/** @type {x} */`

packages/ses/test/import.test.js
boneskull, line 1: What if you added a test for CommonJS?  Or would that be of low value?

boneskull (approved):
🎉

packages/compartment-mapper/test/optional.test.js
kriskowal: Yes.

packages/ses/test/import.test.js
kriskowal: It could have been `'foo'`, `'.'`, or even `''`.

packages/ses/test/import.test.js
kriskowal: Middling value. We don’t test CJS directly at this layer of abstraction so much as we test the mechanisms for supporting it, which are virtual module sources and `compartment.importNow`, neither of which go through the dynamic async import path. The exercise of answering your question convinces me that CJS support is orthogonal to this feature.

packages/ses/src/compartment.js
kriskowal: Answering in comment, but for you:
- A *full specifier* is a key in the compartment's memo of known modules. When we add support for import attributes (`import ansiSys from 'autoexec.bat' with { type: 'bmp' }`), the full specifier will be a _part_ of the compartment memo key. Functions like `compartment.import` take full specifiers. In Node.js and the web, the full specifier happens to be a fully qualified URL, so `import.meta.url` is the same. In Compartment, the location and full specifier are not necessarily the same. With Compartment Mapper, the full specifier is the specifier you would see in a `package.json`, like `./src/compartment.js` (for internal linkage)` or `ses` (for external linkage).
- an *import specifier* by contrast is a specifier in the context of the surrounding module’s full specifier, like `./module-instance.js` is an import specifier in `./src/compartment.js` that resolves to a full specifier: `./src/module-instance.js`.
- you will note that a *relative specifier* like `./foo.js` can has different implications for full specifiers and import specifiers. The `./` prefix implies “internal” for a full specifier and full specifiers cannot have `..` path components. The `./` prefix for an import specifier implies that it should be resolved relative to the importer’s full specifier and `..` path components should eat the preceding path component of the full specifier.
- I don’t have active recall for the terms Node.js chose to distinguish these.

packages/compartment-mapper/test/optional.test.js
kriskowal: Yeah, this is a sign things are getting weedy. We might refactor `knowFailure` to be a callback in the future.

PR #2641, opened 2024-11-20
fix(compartment-mapper): Dev implied by development condition only for entry package

kriskowal:
I discovered this bug while reviewing Compartment Mapper: I’ve deprecated the `dev` option of `mapNodeModules` in favor of `development` in the `conditions` set option, since generally one should want both behaviors for development mode, so one should imply the other. However, the way this is implemented causes `devDependencies` to be linked for every package in the presence of the `development` condition, and should only be in the entry package.

### DIFF
--- a/packages/compartment-mapper/src/node-modules.js
+++ b/packages/compartment-mapper/src/node-modules.js
@@ -360,7 +360,7 @@ const graphPackage = async (
   for (const name of Object.keys(optionalDependencies)) {
     optionals.add(name);
   }
-  if (dev !== undefined && dev !== null ? dev : conditions.has('development')) {
+  if (dev) {
     assign(allDependencies, devDependencies);
   }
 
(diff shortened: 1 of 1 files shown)


packages/compartment-mapper/src/node-modules.js
```diff
@@ -858,22 +858,29 @@ const makeLanguageOptions = ({
...
+  // covers more intentional behaviors of the development mode.
+
   const graph = await graphPackages(
     maybeRead,
     canonical,
     packageLocation,
     conditions,
     packageDescriptor,
-    dev,
+    dev || (conditions && conditions.has('development')),
```
boneskull, line 883: `conditions` is never `false`.

boneskull (approved):
LGTM; the logic on L883 is unnecessarily verbose imo, but not gonna block on it.

Issue #2636, opened 2024-11-18
Compartment Mapper should gracefully include peerDependencies only if present


kriskowal:
## Describe the bug

The Compartment Mapper reaches for `peerDependencies` during `mapNodeModules`, and if the package manager did not arrange for them, the routine fails with an exception for a missing transitive package. We should instead include each peer dependency in the packages map only if it’s physically present.

## Steps to reproduce

Contrive a package graph that contains peerDependencies that are alternately present or absent and observe that the compartment mapper can link the peerDependencies only if they are present, regardless of whether they’re imported by the entry module.

## Expected behavior

`mapNodeModules` should tolerate the absence of a peer dependency, allowing a module graph to link as long as it doesn’t attempt to import the peer dependency.

## Platform environment

All at time of writing (2024-11-18)

kriskowal:
Reviewing our code for this case, it seems that `peerDependencies` are not in fact optional, unless specified as optional in `optionalDependencies` or `peerDependenciesMeta` with `optional: true`. So, the implementation is consistent and provides a useful diagnostic for invalid `node_modules` trees. We could instead ignore `peerDependenciesMeta` and treat all `peerDependencies` or even all `dependencies` as optional if there is no corresponding physical package, which would defer linkage errors to `load`. Those errors might be less informative.

PR #2642, opened 2024-11-21
feat(compartment-mapper): Relax package discovery

kriskowal:
Closes: #2636 

## Description

The compartment mapper is currently strict about the interpretation of `peerDependencies` and `peerDependenciesMeta.optional` as opposed to `optionalDependencies` and is technically correct, but that correctness and understanding is not evenly distributed among tools in the ecosystem.
If the Compartment Mapper instead tolerates the physical absence of expected packages, it defers an error that would be experienced under `mapNodeModules` to `link/load` since that might attempt to load a module from the missing package. This is not a particularly observable difference in behavior, but the error message will be slightly less informative.

In this change, we relax the behavior by default and provide a `strict` option to provide the more informative error and tolerate fewer ecosystem defficiencies.

### Security Considerations

None.

### Scaling Considerations

None.

News updated.

We do not yet publish API documentation for Compartment Mapper, but the necessary annotations are present when these devices are more public.

### Testing Considerations

Includes positive and negative tests for the `strict` behavior.

The nature of some errors during bundling may change. Bundling errors are not observed on Agoric’s chain, so that sensitivity should not cause problems.

### Upgrade Considerations

None.


FUDCo (approved):
Gosh, this seems swell.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -344,14 +346,18 @@ const graphPackage = async (
     devDependencies = {},
   } = packageDescriptor;
   const allDependencies = {};
-  assign(allDependencies, commonDependencyDescriptors);
-  for (const [name, { spec }] of Object.entries(commonDependencyDescriptors)) {
-    allDependencies[name] = spec;
+  for (const [name, descriptor] of Object.entries(
+    commonDependencyDescriptors,
+  )) {
+    if (Object(descriptor) === descriptor) {
```
boneskull, line 352: OK, `Object(x) === x` is new to me. what does it do?

boneskull (commented):
Does this have any effect on packages that are optionally `require`d but not otherwise mentioned in `package.json`?

An example from `typescript`:

```js
try {
  require('source-map-support')
} catch {}
```

...where `source-map-support` is a _dev_ dependency of `typescript`.  So this `require` would _not_ throw if a) someone has a local working copy of `Microsoft/typescript` or b) someone already has `source-map-support` in their `node_modules` (as they often do).

packages/compartment-mapper/src/node-modules.js
kriskowal: This is the terse way of checking `typeof x === 'object' && x !== null`. `Object(x)` returns `x` iff `x` is an object and not null, otherwise boxes primitives.

kriskowal:
> Does this have any effect on packages that are optionally `require`d but not otherwise mentioned in `package.json`?
> 
> An example from `typescript`:
> 
> ```js
> try {
>   require('source-map-support')
> } catch {}
> ```
> 
> ...where `source-map-support` is a _dev_ dependency of `typescript`. So this `require` would _not_ throw if a) someone has a local working copy of `Microsoft/typescript` or b) someone already has `source-map-support` in their `node_modules` (as they often do).

No, this just tolerates the absence in `.../node_modules` of packages that _are_ present in `package.json`, not tolerate the presence in `.../node_modules` of packages that are absent in `package.json`

kriskowal:
@boneskull We could conceivably add an even sloppier `mapNodeModules` that recursively lists `.../node_modules/{,@/}*` and disregards the `*dependencies` in `package.json` entirely. With a `link` hook, we might even be able to make discovery of `.../node_modules` lazy. In any case, not in scope for this PR, but so you know: that is a possibility.

PR #2645, opened 2024-11-22
refactor(pass-style): Make @fast-check/ava a devDependency

gibson042:
Fixes #1741

> [!CAUTION]
> While technically a breaking change, it is pragmatically not marked as such because we are not aware of anything outside of Endo that imports a fast-check arbitrary.

## Description

Rather than importing `fc` directly to export arbitraries, @endo/pass-style/tools.js now returns them from an exported `makeArbitraries` function that requires `fc` as an argument.

Best reviewed [with whitespace ignored](https://github.com/endojs/endo/pull/2645/files?w=1).

### Security Considerations

n/a

### Scaling Considerations

n/a

### Documentation Considerations

n/a

### Testing Considerations

n/a

### Compatibility Considerations

This is a breaking change, but only to @endo/pass-style/tools.js. All imports in this repository have been updated, and I didn't see any in [agoric-sdk](https://github.com/Agoric/agoric-sdk) or [agoric-3-proposals](https://github.com/Agoric/agoric-3-proposals), but if there are any then they will need similar updates.

### Upgrade Considerations

This only affects code intended for testing, and I'm disinclined to even mention it in NEWS.md.

PR #2647, opened 2024-11-22
chore(pass-style): Resolve eslint warnings

gibson042:
As seen at e.g. https://github.com/endojs/endo/pull/2645/files?w=1#files:~:text=Unchanged%20files%20with%20check%20annotations


erights (approved):
Sure

PR #2654, opened 2024-12-05
fix(ses): Consistently name console methods

gibson042:
Ref #2643
Ref 5e8fd33e7c9bf23b12640022a5ca4b11c29d0136

## Description

Unify the inconsistent approaches to creating [virtual] console functions with appropriate "name" properties.

### Security Considerations

None.

### Scaling Considerations

n/a

n/a

### Testing Considerations

I don't think these details are covered by testing, but it seems like an acceptable gap for now.

None.

### Upgrade Considerations

None.


kriskowal (approved):
Looks like a refactor to me. Consider `refactor(ses):` for the commit message. I don’t see it, but if you did incidentally fix a missing `.name`, `feat` or `fix` seem okay, but it would be neat to have a test.

PR #2664, opened 2024-12-18
fix(compartment-mapper/node): support anonymous entrypoints

boneskull:
An "anonymous" entrypoint is one where the closest `package.json` does not contain a `name` field. It is not unusual for CJS packages to create a subdirectory containing `package.json` like this:

```json
{"type": "module"}
```

The effect is that any `.js` file in this subdir will be treated as an ECMAScript module by Node.js.

A problem arises when a module in this subdir is provided as an entrypoint to `mapNodeModules()`. In this case, the "root" package descriptor _has no name_. Ultimately, the resulting shape of the `CompartmentMapDescriptor` fails `assertCompartmentMap()`, presumably due to an `undefined` value (or `"undefined"` string!) where one was not expected.

To avoid the failure, we can assign a string value to the root package descriptor's `name` field.

This is very much a corner case and was discovered by running arbitrary packages through `mapNodeModules()` (specifically, `@babel/runtime`).

- Also added a test for this
- Also added a broad/shallow test case for `mapNodeModules()` since one did not exist

* * *

Questions at time of commit:

- Does the name need to be more unique to avoid collisions with other packages in the dependency tree?
- Severe enough to pull in `node:crypto`?
- Can there ever be more than one "anonymous" package?


kriskowal:
I could also imagine solving this problem with an alternate implementation of `mapNodeModules` and `languageHook` that does not consult `package.json` for dependencies at all and instead follows the Node.js search algorithm closely, and note that would have the drawback that it would, like Node.js, only reveal missing dependency declarations after they have been published and consumed elsewhere.

boneskull:
> A negative side-effect of this change is that modules under the anonymous package.json would not be able to find any of their dependencies. In Node.js, this is not a problem because Node.js does not consult package.json, but for the Compartment Mapper, this would be fatal.

If Node.js didn't consult `package.json` then these `{"type": "module"}` manifests wouldn't work at all 😄

boneskull:
@kriskowal 

But anyhow, the more correct solution is what you suggested: "inherit" from the parent `package.json`.  There seem to be some helper functions.

> I could also imagine solving this problem with an alternate implementation of mapNodeModules and languageHook that does not consult package.json for dependencies at all and instead follows the Node.js search algorithm closely, and note that would have the drawback that it would, like Node.js, only reveal missing dependency declarations after they have been published and consumed elsewhere.

This is Lavamoat, so we wish to (reasonably) maximize out-of-box compat with Node.js.

@naugtur This involves some of the stuff we were talking about today.  I thought you might have _opinions_ on the API.

naugtur:
> I would like to consider a similarly surgical change that would probably have the same intended effect: We can alter the search() function to skip any `package.json` that lacks a `name` property, such that it would fall through to the true package root. These `{"type": "module"}` `package.json`s just alter the interpretation of `.js` extensions in a subtree of the containing package.

This sounds like it could work.

> We might need to do further work to ensure that the `.js` extension for the inner and outer “packages” get interpreted consistently for their respective subtrees. I think that would entail a moduleTypeHook for importHook that closes over the same package descriptor memo as mapNodeModules, allowing that memo to grow in response to importHook as needed.

If moduleTypeHook can be avoided, I'd like that. In node.js loaders the module type is part of the result of calling the equivalent of importHook (part of the return value - the record in our case) and a responsibility of the importHook to provide. Are we not implementing similar logic?

boneskull:
This is tricky.

On the one hand, it _looks like_ we should be able to treat the parent and any "anonymous" package in its scope as a single Compartment, and then configure `types` (`LanguageForModuleSpecifier`) on a file-by-file (folder by folder?) basis to tell it which files are ESM and which are CJS. At least, I _think_ that's what that's for?  This would make the presence of `.js` in `LanguageForExtension` invalid.

On the other, the anonymous package can specify its own dependencies.  Node.js doesn't care; if it's in `node_modules`, it works. But does a dependency in an anonymous package mean its parent should _also_ have access?  If the anon pkg _doesn't_ have its own compartment, is it unreachable from other packages? 

This is further complicated by import maps. I will not think about that right now.

kriskowal:
This is not a case where I would choose to pursue a perfect emulation of Node.js. In my opinion, it’s sufficient to consider a `package.json` that exists solely to qualify the module type for modules within its parent directory a special directive to the containing package. We need a fancier mechanism for tracking language for extension under a prefix tree in each compartment descriptor. We’ll need a similar fancier mechanism to support wildcards/patterns in package imports and exports, but I digress.

naugtur:
> but I digress.

Disagree - I think we'd benefit from designing a solution that can accomodate both. 

As for the logic of considering package.json
A package.json that consists solely of the module type specification (and is lacking name most importantly) we don't take it into account as a package descriptor at all. We threat it as interesting only in case of figuring out the module types. 
That should be consistent with Node and others.

boneskull:
Update: it appears that's all it's using `package.json` for.  However...

The tactics Node.js uses to determine module type are _exceedingly complex_ and I cannot recommend attempting to mimic it in `@lavamoat/node` and _especially_ not in `mapNodeModules()`.

To that end: instead of supporting anonymous entrypoints _per se_, we support a new hook such as:

```ts
type GetModuleTypeHook = (filepath: string) => Promise<'module' | 'commonjs' | undefined>
```

...which might be the best way to support what `@lavamoat/node` wants to do.  There'd be a default "naïve" implementation for when the hook is not provided, and then some fallback value if it _is_ provided and resolved with `undefined`. 

> [!NOTE]
> There's also a notion of a "WASM module" in Node.js which should probably just invoke an exit module; but if supported, `wasm` would be a potential value returned from `GetModuleTypeHook`.

How `@lavamoat/node` would presumably ask Node.js to determine the module format is something I haven't yet thought about.

But in terms of maximizing ecosystem compatibility, a piecemeal approach to supporting edge cases in `mapNodeModules()` seems like taking on too much complexity. A heavy maintenance burden

kriskowal:
> ```ts
> type GetModuleTypeHook = (filepath: string) => Promise<'module' | 'commonjs' | undefined>
> ```

The approach sounds fine to me. Let’s call it `languageHook(location: string) => Promise<Language>`, where Language is `cjs`, `mjs`, arbitrary string, or `undefined` to fall through to the default.

PR #2665, opened 2024-12-19
chore: add typescript declarations to .gitignore

boneskull:
When `npm`-`link`-ing Endo against external projects, the types are not resolvable without first building the declaration files (unlike how types are resolved within the project itself).  This requires compiling declaration files in the appropriate Endo workspaces and re-running as needed.

Previously, all generated declaration files (and their sourcemaps) were recognized as unstaged new files by Git, which is incredibly annoying when trying to see what changes have been made to Endo sources.

**This change adds declaration files, their sourcemaps and any build cache files (`*.tsbuildinfo`) to Git's ignorelist.**

## Escape Hatch

Sometimes it's helpful to keep declarations under version control.  To that end, any _new_ declaration file that needs to be tracked should match:

```js
/^.+\.types\.d\.[mc]?ts$/
```

Example: `foo.types.d.ts`. This will avoid the need to force-add the file to Git.

## Notes

- To avoid the need to rename files (e.g., the various `types.js` files amongst the workspaces), this pattern was **not** chosen:
    ```js
    /^types.*\.d\.[mc]?ts$/
    ```
- If the need arises to track a directory full of declaration files, another convention can be established and appended to the ignorefile.
- All `clean` scripts will need to be updated, since `git clean` needs another flag (`-X`) to delete _ignored_ files.

### What I'm Dealing With

An example of my workflow is something like this:

```sh
npm link -w @endo/compartment-mapper \
  -w ses \
  -w @endo/env-options \
  -w @endo/cjs-module-analyzer \
  -w @endo/static-module-record \
  -w @endo/zip \
  -w @endo/evasive-transform
cd /path/to/lavamoat
npm link @endo/compartment-mapper \
  ses \
  @endo/env-options \
  @endo/cjs-module-analyzer \
  @endo/static-module-record \
  @endo/zip \
  @endo/evasive-transform
```

At this point, my build will fail because the type declarations I need are missing. So I need to:

```sh
cd /path/to/endo
npm run prepack -w @endo/compartment-mapper -w ses
```

This will generate the declarations I need, but `git status` results in:

```text
...
?? packages/compartment-mapper/src/policy.d.ts
?? packages/compartment-mapper/src/policy.d.ts.map
?? packages/compartment-mapper/src/powers.d.ts
?? packages/compartment-mapper/src/powers.d.ts.map
?? packages/compartment-mapper/src/search.d.ts
?? packages/compartment-mapper/src/search.d.ts.map
?? packages/compartment-mapper/src/types/compartment-map-schema.d.ts
?? packages/compartment-mapper/src/types/compartment-map-schema.d.ts.map
?? packages/compartment-mapper/src/types/external.d.ts
?? packages/compartment-mapper/src/types/external.d.ts.map
?? packages/compartment-mapper/src/types/internal.d.ts
?? packages/compartment-mapper/src/types/internal.d.ts.map
?? packages/compartment-mapper/src/types/node-powers.d.ts
?? packages/compartment-mapper/src/types/node-powers.d.ts.map
?? packages/compartment-mapper/src/types/policy-schema.d.ts
?? packages/compartment-mapper/src/types/policy-schema.d.ts.map
?? packages/compartment-mapper/src/types/policy.d.ts
?? packages/compartment-mapper/src/types/policy.d.ts.map
?? packages/compartment-mapper/src/types/powers.d.ts
?? packages/compartment-mapper/src/types/powers.d.ts.map
?? packages/compartment-mapper/src/types/typescript.d.ts
?? packages/compartment-mapper/src/types/typescript.d.ts.map
?? packages/compartment-mapper/src/url.d.ts
?? packages/compartment-mapper/src/url.d.ts.map
?? packages/compartment-mapper/tsconfig.build.tsbuildinfo
```

...which is annoying if I have actual changes in Endo!

## Script changes

I updated `clean` and `postpack` scripts _en masse_:

- Now that declaration files are ignored, `git` needs the `-X` flag to delete them
- Normalized all `postpack` scripts:
  - Updated with new patterns matching newly-ignored declaration, declaration maps, and `.tsbuildinfo` files
  - If a workspace has a `clean` script which was essentially identical to `postpack`, `postpack` just invokes `yarn clean`
  - Replace single quotes `'` with escaped double quotes `\"` for the sake of portability (it is ugly, but Endo may want to support a Windows-based development environment one day)
  - `ses`, in particular, has a `clean` script which differs substantially from `postpack`, so I left `clean` as-is.

## No change to packed files
 
I used my preferred diff tool to compare `new-pack.json`, generated by this PR:

```sh
npm pack -ws --json --dry-run > new-pack.json
```

...and `old-pack.json`, generated by the `master` branch:

```sh
npm pack -ws --json --dry-run > old-pack.json
```

The only differences were the `package.json` files having different sizes and hashes due to the changes to the `clean` and `postpack` scripts (as noted above)

* * *

_Description updated Jan 17 2025_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,3 +86,23 @@ api-docs
 !*.types.d.ts
 !*.types.d.mts
 !*.types.d.cts
+
+# legacy exceptions to the above rule
+!packages/bundle-source/src/exports.d.ts
+!packages/captp/src/ts-types.d.ts
+!packages/cli/test/_types.d.ts
+!packages/compartment-mapper/src/types-external.d.ts
+!packages/compartment-mapper/src/types.d.ts
+!packages/daemon/src/types.d.ts
+!packages/daemon/types.d.ts
+!packages/eventual-send/src/exports.d.ts
+!packages/eventual-send/src/types.d.ts
+!packages/exo/src/types.d.ts
+!packages/far/src/exports.d.ts
+!packages/lp32/types.d.ts
+!packages/pass-style/src/types.d.ts
+!packages/ses/src/reporting-types.d.ts
+!packages/ses/types.d.ts
+!packages/stream/types.d.ts
+!packages/trampoline/types.d.ts
+!packages/where/types.d.ts


boneskull:
(Yes, this change is basically just for me)

mhofman (changes requested):
We have a lot of manually authored `.d.ts` files so a blanket ignore of them is not ok. In general I wish we had a CI job to verify that any files in the repo are not currently excluded by the gitignore config, as it's a pain to manage files "force added"

mhofman:
> * When publishing, `npm`'s default behavior is to skip ignored files unless they are present in a `package.json`'s `files` array. I believe this is already configured correctly, but future packages must maintain the same strategy. I recommend carefully examining the output of `npm pack --dry-run` before the next publish to avoid regressions in the list of published artifacts.

I believe @kriskowal has some commands to verify no regression here that I'd like to see the output of before approving a merge.
Basically list all the files in packed packages before and after, and diff the lists.

boneskull:
@mhofman I feel like the concern may be overblown.

```
git ls-files '*.d.ts'
packages/bundle-source/src/exports.d.ts
packages/captp/src/ts-types.d.ts
packages/cli/test/_types.d.ts
packages/compartment-mapper/src/types-external.d.ts
packages/compartment-mapper/src/types.d.ts
packages/daemon/src/types.d.ts
packages/daemon/types.d.ts
packages/eventual-send/src/exports.d.ts
packages/eventual-send/src/types.d.ts
packages/exo/src/types.d.ts
packages/far/src/exports.d.ts
packages/lp32/types.d.ts
packages/pass-style/src/types.d.ts
packages/ses/src/reporting-types.d.ts
packages/ses/types.d.ts
packages/stream/types.d.ts
packages/trampoline/types.d.ts
packages/where/types.d.ts
```

We aren't really adding a ton of these on a regular basis.

boneskull:
This should surface everything to be published:

```sh
npm pack --workspaces --json --dry-run
```

So we can diff the output of that

turadg:
One way forward is to have a consistent module for hand-written types and make that an exclusion in gitignore.

boneskull:
Some tests are failing because `git clean -f *.d.ts` doesn't work anymore. 😄

boneskull:
@turadg Since I was touching the `postpack` scripts, I noticed that some of the `postpack` scripts are `git clean -f '*.d.ts*' '*.tsbuildinfo'` and others are `git clean -fX '*.d.ts*'`; should all of them be the former?

kriskowal (changes requested):

boneskull:
@turadg @kriskowal @mhofman This is ready for re-review; please see the updated PR description

.gitignore
```diff
@@ -74,3 +74,15 @@ dist/
 api-docs
 
 .aider*
+
+# ts build cache
+*.tsbuildinfo
+
+# ignore all declarations and mappings, but whitelist allow matching the "*.types.d{m,c,t}s"
```
kriskowal, line 81: suggest permit/allow list

packages/base64/package.json
```diff
@@ -33,7 +33,7 @@
   "scripts": {
     "build": "exit 0",
     "prepack": "tsc --build tsconfig.build.json",
-    "postpack": "git clean -f '*.d.ts*' '*.tsbuildinfo'",
+    "postpack": "git clean -fX \"*.d.ts*\" \"*.d.cts*\" \"*.d.mts*\" \"*.tsbuildinfo\"",
```
kriskowal, line 36: I suspect that this will work:
```suggestion
    "postpack": "git clean -fX \"*.d.{,c,m}ts*\" \"*.tsbuildinfo\"",
```

packages/errors/package.json
```diff
@@ -23,14 +23,14 @@
...
-    "clean": "git clean -f '*.d.ts*'",
+    "clean": "git clean -fX \"*.d.ts*\" \"*.d.cts*\" \"*.d.mts*\" \"*.tsbuildinfo\"",
     "lint": "yarn lint:types && yarn lint:eslint",
     "lint-check": "yarn lint",
     "lint-fix": "yarn lint:eslint --fix && yarn lint:types",
     "lint:eslint": "eslint '**/*.js'",
     "lint:types": "tsc",
     "prepack": "tsc --build tsconfig.build.json",
-    "postpack": "git clean -f '*.d.ts*' '*.tsbuildinfo'",
+    "postpack": "yarn clean",
```
kriskowal, line 33: This is probably fine, but I’m reminded that we should avoid using `yarn` in scripts that might be invoked under `npm install` outside our workspace. This is not among them, but it might be a good habit to start using `npm` (or eventually `node --run clean` here).

kriskowal (approved):
Looks like it will work to me.

packages/base64/package.json
boneskull: It does not

.gitignore
boneskull: old habits..

packages/base64/package.json
boneskull: (likewise this syntax is unsupported by `gitignore`)

boneskull:
@mhofman please approve or otherwise withdraw the block

packages/base64/package.json
kriskowal: 🐑,(bah)

.gitignore
```diff
@@ -74,3 +74,15 @@ dist/
...
+*.tsbuildinfo
+
+# ignore all declarations and mappings, but allow the "*.types.d{m,c,t}s"
+# convention below to make it easier to track new declaration files as needed
+*.d.ts*
+*.d.mts*
+*.d.cts*
+!*.types.d.ts
+!*.types.d.mts
+!*.types.d.cts
```
mhofman, line 88: According to the following command, there are still a bunch of existing manually authored files that would end up getting excluded

```
$ git ls-files --cached -i --exclude-standard | grep -v node_modules
packages/bundle-source/src/exports.d.ts
packages/captp/src/ts-types.d.ts
packages/cli/test/_types.d.ts
packages/compartment-mapper/src/types-external.d.ts
packages/compartment-mapper/src/types.d.ts
packages/daemon/src/types.d.ts
packages/daemon/types.d.ts
packages/eventual-send/src/exports.d.ts
packages/eventual-send/src/types.d.ts
packages/exo/src/types.d.ts
packages/far/src/exports.d.ts
packages/lp32/types.d.ts
packages/pass-style/src/types.d.ts
packages/ses/src/reporting-types.d.ts
packages/ses/types.d.ts
packages/stream/types.d.ts
packages/trampoline/types.d.ts
packages/where/types.d.ts
```

I don't think this is working as expected.

boneskull: err, yes. this does not affect existing files. if they are deleted then re-added without being renamed they will need to be force-added.

kriskowal: I agree with @mhofman that we want the above command to produce an empty list in CI. I’d like to add that as a step to lint in CI and I think it’s acceptable to make a policy decision on whether we would introduce any further files consistent with the historical naming conventions, and either explicitly permit them in `.gitignore` either in general or for the specific exceptions.

I think we should allow `types.d.ts`, `types-*.d.ts`, and `types/**/*.d.ts`. I think making explicit legacy exceptions for e.g., `packages/eventual-send/src/exports.d.ts` would help make the above report empty.

boneskull:
I have added explicit exceptions to the existing files matching the glob spec.

mhofman (approved):

.gitignore
```diff
@@ -74,3 +74,35 @@ dist/
...
+
+# grandfathered exceptions to the above rule
```
kriskowal:
```suggestion
# legacy exceptions to the above rule
```

PR #2666, opened 2024-12-19
refactor(pass-style): Avoid name convention conflict

erights:
closes: #XXXX
https://github.com/Agoric/agoric-sdk/pull/10751

## Description

We generally use names of the form `^([A-Z]\w*)I$`, i.e., identifiers beginning with a capital letter and ending with a lone capital "I", for interface guards. However there were some uses of the same form of name for typescript interfaces. Because interfaces and interface guards are adjacent concepts, this could be confusing. This PR fixes this for endo. The companion https://github.com/Agoric/agoric-sdk/pull/10751 fixes it for agoric-sdk. Because the endo names in question were not exported, there is clearly no dependency between these two PRs. 

### Security Considerations

Consistent naming conventions makes code more reviewable, which is good for security.

### Scaling Considerations
none

Consistent naming conventions helps documentation and its readers.

### Testing Considerations
none

Because these names are not exported, and are only type names erased prior to execution, none.

### Upgrade Considerations
Because these are only type names that are erased prior to execution, none.

### DIFF
--- a/packages/pass-style/src/types.d.ts
+++ b/packages/pass-style/src/types.d.ts
@@ -85,14 +85,14 @@ export type Passable<
 > = void | Primitive | Container<PC, E> | PC | E;
 
 export type Container<PC extends PassableCap, E extends Error> =
-  | CopyArrayCommon<PC, E>
-  | CopyRecordCommon<PC, E>
-  | CopyTaggedCommon<PC, E>;
-interface CopyArrayCommon<PC extends PassableCap, E extends Error>
+  | CopyArrayInterface<PC, E>
+  | CopyRecordInterface<PC, E>
+  | CopyTaggedInterface<PC, E>;
+interface CopyArrayInterface<PC extends PassableCap, E extends Error>
   extends CopyArray<Passable<PC, E>> {}
-interface CopyRecordCommon<PC extends PassableCap, E extends Error>
+interface CopyRecordInterface<PC extends PassableCap, E extends Error>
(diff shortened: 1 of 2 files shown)


turadg (commented):
[requesting lint rule](https://github.com/Agoric/agoric-sdk/pull/10751#pullrequestreview-2516441053)

packages/pass-style/src/types.d.ts
```diff
@@ -85,14 +85,14 @@ export type Passable<
 > = void | Primitive | Container<PC, E> | PC | E;
 
 export type Container<PC extends PassableCap, E extends Error> =
-  | CopyArrayI<PC, E>
-  | CopyRecordI<PC, E>
-  | CopyTaggedI<PC, E>;
-interface CopyArrayI<PC extends PassableCap, E extends Error>
+  | CopyArrayCommon<PC, E>
```
turadg, line 88: this really is an interface version of the CopyArray type.

Would you be okay with `CopyArrayInterface`?

erights:
> [requesting lint rule](https://github.com/Agoric/agoric-sdk/pull/10751#pullrequestreview-2516441053)

https://github.com/Agoric/agoric-sdk/pull/10751#issuecomment-2555998777 ;)

packages/pass-style/src/types.d.ts
erights: Yes. Done, thanks.

turadg (approved):
👌

PR #2667, opened 2024-12-20
fix(compartment-mapper): invert keys/values of renames in captureFromMap

boneskull:
<details>
BREAKING CHANGE: The `CaptureResult.compartmentRenames` record returned by `captureFromMap()` is now _new to old_; previously it was _old to new_.

This more accurately solves for the use case of `compartmentRenames`, because the main use case demands the record is inverted. We can plainly see this issue in `captureFromMap`'s test suite (which this change simplifies).

In a future change, [branded types](https://www.learningtypescript.com/articles/branded-types) could describe the distinction between key and value, which are currently both `string`.

> [!IMPORTANT]
>
> The only consumer of `captureFromMap` that I'm aware of is _me_, and _me_ doesn't need this to be SemVer major.
</details>

This change isolates the duplicated code in `archive-lite.js` and `capture-lite.js` into a new internal module, `digest.js`.

It provides `digestCompartmentMap()`, which is now consumed by both `archive-lite.js` and `capture-lite.js`.

Both `makeArchiveCompartmentMap()` and `captureFromMap()` now pass new properties from `digestCompartmentMap()` through to their consumers: `newToOldCompartmentMap` and `oldToNewCompartmentMap`.

**The `compartmentRenames` property is now soft-deprecated** (since its naming is ambiguous). The same property is now `newToOldCompartmentMap`; `compartmentRenames` should eventually be removed.

_Note_: The duplicated code drifted slightly in `archive-lite.js`; see e3b310d. These changes were kept and put into `digest.js`.


packages/compartment-mapper/src/capture-lite.js
```diff
@@ -250,7 +250,12 @@ const captureCompartmentMap = (compartmentMap, sources) => {
   return {
     captureCompartmentMap,
     captureSources,
-    compartmentRenames,
+    compartmentRenames: fromEntries(
+      entries(compartmentRenames).map(([oldName, newName]) => [
+        newName,
+        oldName,
+      ]),
+    ),
```
kriskowal, line 258: I think we consume the same data as `compartmentRenames` in `archive-lite.js` in the new to old mapping order. That at least suggests both the map and inverse map are useful and the mistake is to favor one with the good name.

I am now noticing that `capture-lite.js` contains copypasta of `archive-lite.js` that I would like not to drift apart so we can eventually merge them.

May I propose that we merge these first, then augment this result object with both the map and comap? We could take the opportunity to deprecate `compartmentRenames` in favor of a more clear dual:

- `newToOldCompartmentNames`
- `oldToNewCompartmentNames`

kriskowal (changes requested):

packages/compartment-mapper/src/capture-lite.js
boneskull: @kriskowal 

- `capture-lite` returns:
  ```ts
  {
    compartmentRenames: Record<string, string>,
    captureCompartmentMap: CompartmentMapDescriptor, 
    captureSources: Sources
  }
  ```
  and `archive-lite` returns:
  ```ts
  {
    archiveCompartmentMap: CompartmentMapDescriptor,
    archiveSources: Sources
  }
  ```
  IMO it'd make sense to call the common properties `compartmentMap` and `sources`, respectively (in addition to the two props you mention above).
- By deprecate you mean soft-deprecate? Emit warning? Throw? Document in `NEWS`?
- In your mind, does "merging" ultimately look like "delete one of these files", or remove the copied method, or keep both methods and alias one to the other, or..?
boneskull
- or would you rather I just push some code and we'll see?

kriskowal: By merge, I mean eliminate duplication in the implementation. That might mean extracting a utility module that both can depend upon, or if one depends on the entirety of the other, you can import the utilities from one or the other, making sure not to expose the utilities to the public package exports.

By deprecate, I mean jsdoc `@deprecated` so it shows up in lint but no runtime consequences.

boneskull:
@kriskowal I've applied the requested changes

kriskowal (approved):

boneskull:
will merge when 🟢

Issue #1583, opened 2023-05-10
Why modules must be instantiated uniquely per compartment, despite bundling


erights:
From https://github.com/Agoric/agoric-sdk/issues/7632#issuecomment-1540990371
> I don't understand what this means. Are CopySets not iterable?

CopySets are not iterable, and the reason makes very clear the high costs of the way we currently use Hardened JS aka SES. We are still operating under the normal JS practice of assuming that any one module or package might be loaded multiple times. 

Say a module x.js must both create correct Xs and validate that an alleged X is a correct X. In most programming languages, x.js could just check an alleged X to see if it was made by itself, i.e., x.js. Since x.js is careful to only make correct Xs, it knows that if it made the alleged X to be an X that it is a correct X. To do this requires some sort of branding, which most programming languages provide by some combination of static and dynamic techniques.

In JS we have two branding mechanisms that it seems we could use: WeakSets and classes-with-private-fields. For our purposes here they are equivalent, so let's only consider WeakSets. x.js could create and encapsulate a WeakSet in which it puts only the Xs that it creates. It then validates than an alleged X is a correct X simply by seeing if it is in the encapsulated WeakSet. In fact, we used to do that before we got bitten by the "multiple instances of x.js blindly co-existing" problem.

The culprit, as I understand it, is the happenstance of bundling decisions. Current bundling practice, as I understand it, is not careful to avoid duplicating the same x.js module as bundled into two different bundles loaded into the same environment. The unbundling process does nothing to prevent the code duplication from resulting in co-existing module instances. In fact, the multiple bundlings may be of different versions and subject to different linkage. This also explains why x.js cannot coordinate with "itself" purely by convention, such as placing a registered symbol on an X to say "I really am an X". There is no convention that one correct instance of x.js can use to talk to another that an attacker cannot use.

As a result, an X made by one x.js instance x1 is not registered in the WeakSet created by another x.js instance x2. When a correct X make by x1 is presented to x2 for validation, x2 rejects it. The uncontrolled blind duplication has resulted in a rejection that would normally indicate misbehavior, even though all the source code involved was operating correctly.

[...]

PR #2673, opened 2024-12-30
feat(no-trapping-shim)

erights:

Closes: #XXXX
Refs: #XXXX

## Description

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations


packages/no-trapping-shim/src/no-trapping-pony.js
```diff
@@ -0,0 +1,206 @@
...
+  }
+};
+ObjectPlus.prototype = OriginalObject.prototype;
+addExtras(ObjectPlus, OriginalObject, extraObjectMethods);
+export { ObjectPlus };
+
+const makeMetaHandler = handler =>
+  freeze({
+    get(_, trapName, _receiver) {
+      return freeze((target, ...rest) => {
```
mhofman, line 165: The overhead of creating a new closure for every trap invocation feels scary. If the goal is to catch future traps not yet specified, why not use a handler with well known traps defined in advance and a proxy as proto to dynamically define unknown traps?

packages/no-trapping-shim/src/no-trapping-pony.js
```diff
@@ -0,0 +1,206 @@
...
+        if (
+          isNoTrappingInternal(target, true) ||
+          handler[trapName] === undefined
+        ) {
+          return ReflectPlus[trapName](target, ...rest);
+        } else {
+          return handler[trapName](target, ...rest);
```
mhofman, line 172: Conformance: there should be a single `[[Get]]` for the trap on the user provided `handler`

packages/no-trapping-shim/src/no-trapping-pony.js
```diff
@@ -0,0 +1,206 @@
...
+  const safeHandler = makeSafeHandler(handler);
+  const proxy = new OriginalProxy(target, safeHandler);
+  proxyHandlerMap.set(proxy, [target, handler]);
+  return proxy;
+};
+ProxyPlus.revocable = (target, handler) => {
+  const safeHandler = makeSafeHandler(handler);
+  const { proxy, revoke } = OriginalProxy.revocable(target, safeHandler);
+  proxyHandlerMap.set(proxy, [target, handler]);
+  return { proxy, revoke };
```
mhofman, line 203: I thought in our previous discussions we arrived at the conclusion that a non trapping proxy effectively cannot be revoked. I guess this should be an open issue for the proposal.

packages/no-trapping-shim/src/no-trapping-pony.js
```diff
@@ -0,0 +1,206 @@
...
+ObjectPlus.prototype = OriginalObject.prototype;
+addExtras(ObjectPlus, OriginalObject, extraObjectMethods);
+export { ObjectPlus };
+
+const makeMetaHandler = handler =>
+  freeze({
+    get(_, trapName, _receiver) {
+      return freeze((target, ...rest) => {
+        if (
+          isNoTrappingInternal(target, true) ||
```
mhofman, line 167: Isn't setting the proxy as non-trapping effectively the same as removing all traps from the handler? Since we interject on the user provided handler, can't we actually do just that?

mhofman (commented):
I've only skimmed through quickly, but from what I can tell this is following closely (but not fully) the letter of the proposed spec: that no state is held on the proxy and checks are instead made on the shadow target. However I have a feeling that we can prove an equivalent approach would be to store state on proxy objects only, since by definition ordinary objects can't be trapping. This should reduce the cost of side tables necessary for the implementation.

packages/no-trapping-shim/CHANGELOG.md
```diff
@@ -0,0 +1,143 @@
+# Change Log
```
erights, line 1: @kriskowal This first commit is the pristine state produced from 
```sh
$ ./scripts/create-package.sh no-trapping-shim
```
See the voluminous `CHANGELOG.md` in that first commit, which I truncated in a subsequent commit. Is this a tooling bug, where the thing that updated `CHANGELOG.md` in general does not know about the one that serves as a template for new packages?

packages/no-trapping-shim/src/no-trapping-pony.js
erights: Given the second part of your suggestion "a proxy as proto to dynamically define unknown traps?", I don't need to do the first part "why not use a handler with well known traps defined in advance". Rather, I have the "proxy as proto" (aka the metaHandler) install overrides on the handlerPlus lazily, avoiding further uses of the double reflection.

So, the second part of your suggestion, done. ***PLEASE*** rereview this carefully, as it is much trickier than the previous non-lazy version.

Resolving.

packages/no-trapping-shim/src/no-trapping-pony.js
erights: Done

packages/no-trapping-shim/src/no-trapping-pony.js
erights: Actually, our previous conclusion is necessary in order for, for example, a copyRecord to have the data-only stability properties we need. So no need to make this an open issue.

Done.

packages/no-trapping-shim/src/no-trapping-pony.js
erights: Can't do it immediately, since a target becoming non-trapping does not immediately reveal all the proxies with that target that then need to become non-trapping.

Rather, in making the metaHandler lazily override the trapName on the handlerPlus, I also arranged for it to lazily re-override with `undefined` for the no-trapping case. So after the first time, handlerPlus[trapName] will be `undefined` and it will continue to do the right thing, but faster.

Done.

packages/no-trapping-shim/CHANGELOG.md
erights: @kriskowal should I file a bug on this? Do you expect to fix it by changing the tooling for updating `CHANGELOG.md` or by changing `create-package.sh`?

erights:
@kriskowal , at https://github.com/endojs/endo/actions/runs/12563852500/job/35026252842?pr=2673 I am getting lint errors that seem clearly unrelated to this PR. Any idea what happened?

erights:
@mhofman , all your suggestions addressed. PTAL. Thanks.

packages/no-trapping-shim/src/no-trapping-pony.js
mhofman: I thought about that but the main reason I suggested preinstalling traps is for performance reasons for engines that use a hidden class. I'm not sure how a proxy proto interacts with that, but in general adding properties to an object tends to make the object change shape.

packages/no-trapping-shim/src/no-trapping-pony.js
mhofman: Yeah I definitely considered suggesting either setting the object's trap as `undefined` or as the original Reflect function.

mhofman:
> I have a feeling that we can prove an equivalent approach would be to store state on proxy objects only, since by definition ordinary objects can't be trapping.

It was obviously too late for me to form a proper intuition there. The non-trapping trait is not proxy specific, but actually a regular object's trait about any proxy that may target them. Otherwise we would have a predicate that can ask "is this a proxy object behaving exotically". By defining the trait on objects in general, we maintain the transparency of proxies.

packages/no-trapping-shim/src/no-trapping-pony.js
erights: This first optimizations seems justified as highly likely to make a difference. Any further optimization is, IMO, premature until we have profile data indicating that it is worth the extra complexity cost.

As a shim for a proposal, I also want it to be highly readable, ideally as an executable spec. Unfortunately, even this first optimization seems to have lost that.

Leaving resolved.

erights: I'll just state for the record that I predict profiling will not flag a need for this additional optimization. The lazy creation should converge quickly, and once converged has no remaining difference from this further optimization.

packages/no-trapping-shim/src/no-trapping-pony.js
```diff
@@ -0,0 +1,278 @@
...
+ *   isNoTrapping: (target: any) => boolean,
+ *   suppressTrapping: (target: any) => boolean,
+ *   originalHandler: ProxyHandler<any>
+ * }}
+ */
+const makeHandlerPlus = originalHandler => ({
+  // @ts-expect-error TS does not know what this __proto__ is doing
+  __proto__: new OriginalProxy({}, metaHandler),
+  // relies on there never being a trap named `originalHandler`.
+  originalHandler,
```
mhofman, line 237: To avoid any potential collisions with trap methods, let's store the original handler as a unique symbol property on `handlerPlus`. After all that's what symbols were meant for :)

```js
const ORIGINAL_HANDLER = Symbol('OriginalHandler');

const makeHandlerPlus = originalHandler => ({
  __proto__: new OriginalProxy({}, metaHandler),
  [ORIGINAL_HANDLER]: originalHandler,
});
```

packages/no-trapping-shim/src/no-trapping-pony.js
```diff
@@ -0,0 +1,278 @@
...
+  }
+  const handlerPlus = makeHandlerPlus(handler);
+  const proxy = new OriginalProxy(target, handlerPlus);
+  proxyHandlerMap.set(proxy, [target, handler]);
+  return proxy;
+};
+// The `OriginalProxy` is both constructible (i.e., responsive to `new`) and
+// lacks a `prototype` property. The closest we can come to this is to set
+// `ProxyPlus.prototype` to `undefined`
+ProxyPlus.prototype = undefined;
```
mhofman, line 262: Not entirely true. Since we never care about the receiver in the case of a plain call, you could use `.bind(null)` which creates a function that lacks a `.prototype` but is constructible if the original is. You'll of course need to redefine the `.name`.

mhofman (commented):
I've only reviewed the implementation so far. Looks good overall, except I don't understand this `shouldThrow` condition.

Also a potential optimization: when adding a proxy to the `noTrappingSet`, we could eagerly redefine all own properties of `handlerPlus` to `undefined` (including the `originalHandler`). To make that easier we just need to save `handlerPlus` in the `proxyHandlerMap` instead of the original `handler`. If we do this, the original handler would become unreferenced when the proxy is found to be non-trapping, giving the nice property that it can be collected (this would be true with a native spec implementation).

erights:
Note to reviewers: In relayering this on #2679 but before addressing the unresolved conversations above, I did some squashing.

packages/no-trapping-shim/src/no-trapping-pony.js
```diff
@@ -0,0 +1,278 @@
...
+  if (trap === undefined) {
+    return false;
+  }
+  const result = apply(trap, handler, [target]);
+  const ofTarget = isNoTrappingInternal(target, shouldThrow);
+  if (result !== ofTarget) {
+    if (shouldThrow) {
+      throw TypeError(
+        `'isNoTrapping' proxy trap does not reflect 'isNoTrapping' of proxy target (which is '${ofTarget}')`,
+      );
```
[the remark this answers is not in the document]
erights: Done

packages/no-trapping-shim/src/no-trapping-pony.js
```diff
@@ -0,0 +1,278 @@
...
+    return isNoTrappingInternal(target, true);
+  },
+  suppressTrapping(target) {
+    if (isPrimitive(target)) {
+      return target;
+    }
+    if (suppressTrappingInternal(target, true)) {
+      return target;
+    }
+    throw TypeError('preventExtensions trap returned falsy');
```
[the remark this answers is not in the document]
erights: Done

packages/no-trapping-shim/src/no-trapping-pony.js
erights: Good idea. Done

packages/no-trapping-shim/src/no-trapping-pony.js
erights: Awesome, thanks! I completely missed that.

Done.

packages/non-trapping-shim/src/non-trapping-pony.js
```diff
@@ -0,0 +1,289 @@
...
+/**
+ * In the shim, `ProxyPlus` replaces the global `Proxy`.
+ *
+ * We use `bind` as the only way for user code to produce a
+ * constructible function (i.e., one that responds to `new`) without a
+ * `.prototype` property.
+ *
+ * @type {ProxyConstructor}
+ */
+const ProxyPlus = ProxyInternal.bind(undefined);
```
mhofman, line 272: You do need to redefine `.name` if you want to keep the original `"Proxy"` name instead `"bound Proxy"`

packages/no-trapping-shim/src/no-trapping-pony.js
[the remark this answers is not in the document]
mhofman: Maybe I missed something, but I still see this `shouldThrow` which I don't understand the purpose of. Why are we not throwing unconditionally here?

packages/no-trapping-shim/src/no-trapping-pony.js
mhofman: I wouldn't expect anyone to know this, it's a very obscure quirk of `.bind()`

mhofman (commented):
Reopened the `shouldThrow` conversation as I don't see anything changed

packages/no-trapping-shim/src/no-trapping-pony.js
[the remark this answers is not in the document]
erights: GOOD CATCH!

I am surprised I somehow closed this conversation. At that moment I must have thought I was looking at something else. I don't remember, but I never thought I had done this one.

THANKS!

packages/non-trapping-shim/src/non-trapping-pony.js
erights: Done

mhofman:
> In theory, old code could use `Proxy` in ways that this PR would break. But this is unlikely in general, and highly unlikely for any code in or depending on endo. Thus, practically, none.

As discussed in https://github.com/endojs/endo/pull/2675#pullrequestreview-2530812896, this turns out to not be so simple.

erights:
In a verbal discussion, @gibson042 and I concluded that the [eval twin problems](https://github.com/endojs/endo/issues/1583) with the ponyfill are fatal. So I just added a commit that stops exporting the ponyfill. Currently, we still have the ponyfill vs shim internally for modularity purposes. But the ponyfill should only be used for
- separate unit testing
- building the shim.

packages/no-trapping-shim/src/no-trapping-pony.js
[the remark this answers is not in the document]
erights: Done

packages/no-trapping-shim/CHANGELOG.md
erights: Filed as https://github.com/endojs/endo/issues/2697 , so resolving here.

erights:
> > In theory, old code could use `Proxy` in ways that this PR would break. But this is unlikely in general, and highly unlikely for any code in or depending on endo. Thus, practically, none.
> 
> As discussed in [#2675 (review)](https://github.com/endojs/endo/pull/2675#pullrequestreview-2530812896), this turns out to not be so simple.

Quite right. In discussion with @michaelfig , we arrived at the opt-in mechanism this PR now implements. I added updates to the PR comment to reflect this.

erights:
This has been idle long enough, and reconciling it with changes on master since then was noisy enough, that I'm about to squash. I doubt it will lose any context still in reviewers' heads.

PR #2674, opened 2024-12-30
typecheck tests and bump typescript-eslint

turadg:
_evergreen_

## Description

The `typescript-eslint` dep was behind so I tried to bump it. The major version breaking change required linted files to be included in a tsconfig. Files in `demo`, `scripts` and `test` were linted but not typechecked, so I added those to the tsconfigs. Then I did a bunch of work to get them to green.

Best to review by commit.

### Security Considerations

none

### Scaling Considerations

none

none

### Testing Considerations

CI

Some slight changes to type interfaces but no breaking runtime changes

### Upgrade Considerations

 ? necessary?


packages/bundle-source/demo/reexport-fortune-ts.js
boneskull, line 1: may exclude this dir from typechecking?

packages/bundle-source/test/endo-script-format.test.js
```diff
@@ -9,6 +9,7 @@ import bundleSource from '../src/index.js';
  * @param {string} entry
  * @param {Options} options
  */
+// @ts-expect-error 'Options' could be instantiated with a different subtype of constraint 'Partial<any>'.
 const generate = async (entry, options = {}) => {
```
boneskull, line 13:
```suggestion
const generate = async (entry, options) => {
```

This should suppress the error, but then you'll need to supply an empty object somewhere below

packages/bundle-source/test/no-transforms.test.js
```diff
@@ -11,6 +11,7 @@ test('no-transforms applies no transforms', async t => {
   const entryPath = url.fileURLToPath(
     new URL(`../demo/circular/a.js`, import.meta.url),
   );
+  // @ts-expect-error Property 'endoZipBase64' does not exist on type '{ moduleFormat: "endoScript"; source: string; } | { moduleFormat: "endoZipBase64"; endoZipBase64: string; endoZipBase64Sha512: string; } | { moduleFormat: "nestedEvaluate"; source: string; sourceMap: string; } | { ...; }'.
```
boneskull, line 14: urgh, bad return type

packages/common/test/object-meta-map.test.js
```diff
@@ -11,6 +11,7 @@ test('test objectMetaMap', async t => {
         ? undefined
         : {
             ...desc,
+            // @ts-expect-error desc.value possibly undefined
```
boneskull, line 14: I'd probably throw a `t.assert.ok(desc.value)` before this statement

packages/ses/test/tame-date.test.js
```diff
@@ -16,7 +16,9 @@ test('lockdown start Date is powerful', t => {
 });
 
 test('lockdown Date.prototype.constructor is powerless', t => {
-  const SharedDate = Date.prototype.constructor;
+  const SharedDate = /** @type {DateConstructor} */ (
```
boneskull, line 19: Is `Date` not `Date.prototype.constructor`? cc @erights

packages/ses/tsconfig.json
```diff
@@ -2,12 +2,15 @@
   "extends": "../../tsconfig.eslint-base.json",
   "compilerOptions": {
     "allowJs": true,
+    "skipLibCheck": true,
```
boneskull, line 5: `skipLibCheck: false` probably also why linting is slow

packages/ses/types.d.ts
```diff
@@ -135,10 +135,12 @@ export type ModuleMapHook = (
   moduleSpecifier: string,
 ) => ModuleDescriptor | undefined;
 export type ImportHook = (moduleSpecifier: string) => Promise<ModuleDescriptor>;
-export type ImportNowHook = (moduleSpecifier: string) => ModuleDescriptor;
+export type ImportNowHook = (
```
boneskull, line 138: Is this true? Can it return `undefined`?

packages/zip/test/zip.test.js
```diff
@@ -19,6 +19,7 @@ test('zip round trip', async t => {
 
   const reader = new ZipReader(writer.snapshot());
   const text = textDecoder.decode(reader.read('hello/hello.txt'));
+  // @ts-expect-error undefined if file not found
```
boneskull, line 22: would probably assert the result is truthy instead

boneskull (approved):
many comments and questions but nothing blockworthy

packages/eslint-plugin/lib/configs/internal.js
```diff
@@ -28,12 +33,14 @@ dynamicConfig.overrides.push({
...
-// Downgrade restrict-plus-operands to a warning for test files
-// until we have time to clean them up.
 dynamicConfig.overrides.push({
   files: ['**/test/**/*.{js,ts}'],
   rules: {
+    // Downgrade restrict-plus-operands to a warning for test files
+    // until we have time to clean them up.
     '@typescript-eslint/restrict-plus-operands': 'warn',
+    // XXX override for RESM concession below
+    '@endo/no-optional-chaining': 'off',
```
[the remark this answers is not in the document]
turadg: The "RESM concession below" is:
```
    // Agoric still uses Endo dependencies under an emulation of ESM we call RESM
    // because it is invoked with `node -r esm`.
    // RESM does not support ?? nor ?. operators, so we must avoid them expressly.
```
Endo tests are not exported so optional-chaining is fine in tests.

packages/eventual-send/test/types.test.js
```diff
@@ -43,7 +45,7 @@ const Far = (iface, value) => {
 /**
  * Check the performance of the legacy ERef type.
  *
- * @param {import('./prepare-test-env-ava').ExecutionContext<unknown>} t
+ * @param {import('ava').ExecutionContext<unknown>} t
```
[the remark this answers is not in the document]
turadg: That's a preference, not a repo requirement. It's better but out of scope of this PR.

packages/ses/test/_harden-mockery.js
```diff
@@ -1,8 +1,10 @@
 export const mockHardened = new WeakSet();
+/** @type {import("../types.js").Harden & {isFake?: true}} */
```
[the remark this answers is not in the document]
turadg: [maybe later](https://github.com/endojs/endo/pull/2674#discussion_r1935815024)

packages/compartment-mapper/test/scaffold.js
```diff
@@ -88,7 +89,7 @@ export function scaffold(
   assertFixture,
   fixtureAssertionCount,
   {
-    onError,
+    onError = /** @type {(t, {error, title})} */ (undefined),
```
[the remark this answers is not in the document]
turadg: good point. I made it accurate instead

packages/eventual-send/test/types.test.js
```diff
@@ -6,7 +6,8 @@ import { E } from './_get-hp.js';
 /**
  * Mock a Remotable maker.
  *
- * @template L,R
+ * @template {Record<string, any>} L
+ * @template {Record<string, any>} R
```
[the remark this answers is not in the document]
turadg: nope, props != remoteMethods

packages/import-bundle/demo/source-maps/RUNME.js
```diff
@@ -15,9 +15,7 @@ lockdown();
 
 const start = url.fileURLToPath(new URL('start.js', import.meta.url));
 
-const bundle = await bundleSource(start, {
-  cacheSourceMaps: true,
-});
+const bundle = await bundleSource(start, {});
```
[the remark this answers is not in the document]
turadg: no overload matches

packages/bundle-source/test/no-transforms.test.js
kriskowal: I think we already do, and specifically that passing `moduleFormat: 'endoZipBase64'` is supposed to obviate this check. The template for `bundleSource` is supposed to infer which overload is called based on that option and narrow the return type.

packages/captp/test/worker.js
```diff
@@ -1,10 +1,13 @@
 import '@endo/init/pre-remoting.js';
 import '@endo/init/debug.js';
 
-import { parentPort } from 'worker_threads';
+import { parentPort as maybeParentPort } from 'worker_threads';
 import { Fail } from '@endo/errors';
 import { makeGuest, makeHost } from './traplib.js';
 
+if (!maybeParentPort) throw new Error('null parentPort');
```
kriskowal, line 8: I guess but do not know that this is safe in the face of port `0` as opposed to `undefined`. `0` is a special port for `listen`, but `connect` won’t ever provide an address with a port of `0` unless there’s an error.

packages/compartment-mapper/test/policy.test.js
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
kriskowal, line 1: These can no longer be inferred from directory pattern matching? I do like that they’re evident, but also they are numerous.

packages/env-options/src/env-options.js
```diff
@@ -51,12 +51,11 @@ export const makeEnvironmentCaptor = (aGlobal, dropNames = false) => {
...
    *
-   * @template {string} [T=string]
    * @param {string} optionName
-   * @param {T} defaultSetting
-   * @param {T[]} [optOtherValues]
+   * @param {string} defaultSetting
+   * @param {string[]} [optOtherValues]
    * If provided, the option value must be included or match `defaultSetting`.
-   * @returns {T}
+   * @returns {string}
```
kriskowal, line 58: I don’t recall why this was a template type before, but there was probably a reason, @erights?

packages/eslint-plugin/lib/configs/internal.js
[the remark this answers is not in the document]
kriskowal: @boneskull we have very recently been liberated from this constraint. Thanks for spotting.

packages/exo/test/non-enumerable-methods.test.js
```diff
@@ -42,7 +42,7 @@ test('test defineExoClass', t => {
   t.throws(() => upCounter.incr(-3), {
     message: 'In "incr" method of (UpCounter): arg 0?: -3 - Must be >= 0',
   });
-  // @ts-expect-error bad arg
+  // FIXME typedef should catch bad arg
```
kriskowal, line 45: Attn @erights, I think we can tolerate a regression in tsc catching a case here. Maybe we can downgrade FIXME to a note that we would not be surprised if ts-expect-error became necessary in a future tsc upgrade.

packages/import-bundle/demo/source-maps/RUNME.js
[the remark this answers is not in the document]
kriskowal: I think the purposes of this file is to validate that source maps get cached. Let’s keep the code correct at runtime and mark the error as expected, maybe file an issue to fix the signature of `bundleSource`.

packages/module-source/test/_native.js
```diff
@@ -1,3 +1,4 @@
 /* global globalThis */
 export const NativeModuleSource = globalThis.ModuleSource;
+// @ts-expect-error XXX typedef
```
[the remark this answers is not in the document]
kriskowal: Not in this case. I don’t think we have an analogous `/// <reference type="@endo/module-source/shim.js">` but perhaps we should.

packages/ses/test/_harden-mockery.js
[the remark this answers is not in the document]
kriskowal: I’d like to converge on this preference opportunistically. It is not a requirement for this change.

packages/ses/test/tame-date.test.js
kriskowal: No, `SharedDate` is attenuated to avoid a timing side-channel by default, as is `Date.now()` and similarly `Math.random()` for a deterministic PRNG snooping side-channel. The signature is the same.

kriskowal (approved):
This is largely mechanical and I don’t want a big diff like this to stay on the lam for long. Please at least address the runtime behavior regression in RUNME.js with a type exception for now. Filed https://github.com/endojs/endo/issues/2834.

PR #2675, opened 2024-12-31
feat(ses,pass-style): use no-trapping integrity level for safety

erights:

Closes: #XXXX
Refs: #XXXX

## Description

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations


packages/pass-style/src/passStyle-helpers.js
```diff
@@ -165,7 +166,7 @@ const makeCheckTagRecord = checkProto => {
       (isObject(tagRecord) ||
         (!!check &&
           CX(check)`A non-object cannot be a tagRecord: ${tagRecord}`)) &&
-      (isFrozen(tagRecord) ||
+      (isNoTrapping(tagRecord) ||
         (!!check && CX(check)`A tagRecord must be frozen: ${tagRecord}`)) &&
```
erights, line 170: Changing this to

```suggestion
        (!!check && CX(check)`A tagRecord must be no-trapping: ${tagRecord}`)) &&
```

is technically more correct. But will be understood much less often.

packages/pass-style/src/passStyleOf.js
```diff
@@ -177,7 +182,7 @@ const makePassStyleOf = passStyleHelpers => {
           return 'remotable';
         }
         case 'function': {
-          isFrozen(inner) ||
+          isNoTrapping(inner) ||
             Fail`Cannot pass non-frozen objects like ${inner}. Use harden()`;
```
erights, line 186: Changing to

```suggestion
            Fail`Cannot pass trapping objects like ${inner}. Use harden()`;
```

would technically be more correct. But will less often be understood.

erights:
@michaelfig Adding you as an additional reviewer. Please look at least at the changes to eventual-send and captp. Thanks!

packages/captp/src/captp.js
```diff
@@ -58,12 +58,12 @@ const reverseSlot = slot => {
  * @property {(slot: CapTPSlot, value: any) => void} markAsExported
  * @property {(slot: CapTPSlot) => void} deleteExport
  * @property {() => void} didDisconnect
- 
+
```
michaelfig, line 61:
```suggestion
  *
```

packages/captp/src/captp.js
```diff
@@ -58,12 +58,12 @@ const reverseSlot = slot => {
...
  * @typedef {object} MakeCapTPImportExportTablesOptions
  * @property {boolean} gcImports
  * @property {(slot: CapTPSlot) => void} releaseSlot
  * @property {(slot: CapTPSlot) => RemoteKit} makeRemoteKit
- 
+
```
michaelfig, line 66:
```suggestion
  *
```

packages/eventual-send/src/E.js
```diff
@@ -171,6 +180,13 @@ const makeEGetProxyHandler = (x, HandledPromise) =>
  * @param {HandledPromiseConstructor} HandledPromise
  */
 const makeE = HandledPromise => {
+  // Note the use of `freeze` rather than `harden` below. This is because
+  // `harden` now implies no-trapping, and we depend on proxies with these
+  // almost-empty targets to remain trapping for traps `get`, `apply`, and `set`
+  // which can still be interesting even when the target is frozen.
+  // `get` and `has`, if not naming an own property, are still general traps,
+  // which we rely on. `apply`, surprisingly perhaps, is free to ignore the
+  // target's call behavior and just do its own thing instead.
```
michaelfig, line 189:
```suggestion
  // Note the use of `freeze` of the Proxy targets below. This is because
  // `harden` now implies non-trapping, and we depend on proxies with these
  // almost-empty targets to remain trapping for traps `get`, `apply`, and `set`
  // which can still be interesting even when the target is frozen.
  // `get` and `has`, if not naming an own property, are still general traps,
  // which we rely on. `apply`, surprisingly perhaps, is free to ignore the
  // target's call behavior and just do its own thing instead.
```

michaelfig (commented):
Your changes to eventual-send and captp LGTM.

Is it too late to change all occurrences of "no-trapping" to "non-trapping"?  That would be more consistent with other terms in the integrity levels, like "non-frozen".

packages/captp/src/captp.js
erights: Done

packages/captp/src/captp.js
erights: Done

erights:
> Your changes to eventual-send and captp LGTM.
> 
> Is it too late to change all occurrences of "no-trapping" to "non-trapping"? That would be more consistent with other terms in the integrity levels, like "non-frozen".

Done.

packages/eventual-send/src/E.js
erights: Explanation moved and reworked at https://github.com/endojs/endo/pull/2679/files#diff-96bb1b0ba7133e217b7a5114fe5fd34af40a92b43c7b5911a4886ce6554f0dd6 preparing-for-stabilize.md

Please have a look and review there. Resolving here.

mhofman (commented):
So I have 2 concerns after looking at this PR and its adoption in agoric-sdk (https://github.com/Agoric/agoric-sdk/pull/10795):
- Given the early nature of the non-trapping proposal, I would like SES and endo packages to avoid taking a hard dependency on it. Can we instead feature detect and use non-trapping only if available? Then we can move the shim application to `@endo/init`?
- The changes to `harden` (and to some extend `passStyleOf`) are presenting a potential upgrade hazard. We generally consider `harden` to be part of the platform, which means it may get upgraded independently of the code using it. The problem is that we have some code with `harden` usages that are no longer valid after this change: any code that was hardening proxies or their target. That means we will not be able to use the same user code with updated XS and/or lockdown bundles that include this change. I am really not sure how we work around this one, besides accepting this breaking change / incompatibility, and require that platforms like agoric-sdk use "repaired" lockdown bundles to skip/undo this change if it needs to load incompatible user code.

packages/pass-style/src/passStyleOf.js
```diff
@@ -143,14 +149,17 @@ const makePassStyleOf = passStyleHelpers => {
...
+            if (!isFrozen(inner)) {
+              throw assert.fail(
+                // TypedArrays get special treatment in harden()
+                // and a corresponding special error message here.
+                isTypedArray(inner)
+                  ? X`Cannot pass mutable typed arrays like ${inner}.`
+                  : X`Cannot pass non-frozen objects like ${inner}. Use harden()`,
+              );
+            }
+            throw Fail`Cannot pass non-trapping objects like ${inner}`;
```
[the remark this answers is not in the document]
erights: Done

packages/pass-style/test/passStyleOf.test.js
```diff
@@ -13,7 +13,7 @@ const harden = /** @type {import('ses').Harden & { isFake?: boolean }} */ (
   global.harden
 );
 
-const { getPrototypeOf, defineProperty, freeze } = Object;
+const { getPrototypeOf, defineProperty, freeze, suppressTrapping } = Object;
```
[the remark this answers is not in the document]
erights: Done

erights:
Converting to draft while we wait for draft https://github.com/endojs/endo/pull/2684 to settle down

packages/pass-style/test/passStyleOf.test.js
```diff
@@ -141,8 +117,8 @@ test('some passStyleOf rejections', t => {
  * For testing purposes, makes a *non-frozen* TagRecord-like object with
  * non-enumerable PASS_STYLE and Symbol.toStringTag properties.
  * A valid Remotable must inherit from a valid TagRecord.
- * - Before stabilize/suppressTrapping, a valid TagRecord must be frozen.
- * - After stabilize/suppressTrapping, a valid TagRecord must also be
+ * - Before stabilize/hardenOrSuppressTrapping, a valid TagRecord must be frozen.
+ * - After stabilize/hardenOrSuppressTrapping, a valid TagRecord must also be
```
erights, line 121: typo

packages/ses/src/commons.js
```diff
@@ -14,6 +14,8 @@
 /* global globalThis */
 /* eslint-disable no-restricted-globals */
 
+import '@endo/non-trapping-shim/shim.js';
+
```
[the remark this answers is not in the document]
erights: I do ***not*** want to move this switch to `@endo/init` because the decision by other importers whether to start ses via `@endo/init` or otherwise is uncorrelated with whether they should be subject to this shim.

In discussion with @michaelfig , we decided to make the shim itself opt-in via an env-option. The env-option defaults to `'disabled'`, so by default there should be no compat issue. See latest updates to https://github.com/endojs/endo/pull/2673. Resolving.

packages/ses/src/make-hardener.js
```diff
@@ -182,8 +187,17 @@ export const makeHardener = () => {
         // Also throws if the object is an ArrayBuffer or any TypedArray.
         if (isTypedArray(obj)) {
           freezeTypedArray(obj);
+          if (isFrozen(obj)) {
```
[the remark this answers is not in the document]
erights: If this is worth doing, can it be postponed to a later PR?

packages/ses/package.json
```diff
@@ -85,7 +85,8 @@
     "postpack": "git clean -f '*.d.ts*' '*.tsbuildinfo'"
   },
   "dependencies": {
-    "@endo/env-options": "workspace:^"
+    "@endo/env-options": "workspace:^",
+    "@endo/non-trapping-shim": "^0.1.0"
```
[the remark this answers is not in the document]
erights: Done

packages/ses/src/make-hardener.js
```diff
@@ -238,8 +252,6 @@ export const makeHardener = () => {
                     // NOTE: Calls getter during harden, which seems dangerous.
                     // But we're only calling the problematic getter whose
                     // hazards we think we understand.
-                    // @ts-expect-error TS should know FERAL_STACK_GETTER
-                    // cannot be `undefined` here.
```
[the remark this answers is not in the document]
erights: https://github.com/endojs/endo/pull/2673/commits/c9b61ee4209eef0271d191586307c3855ec46950 from https://github.com/endojs/endo/pull/2673 has now changed this to @ts-ignore to avoid inconsistent typing results. See https://github.com/endojs/endo/issues/2698

Resolving.

yarn.lock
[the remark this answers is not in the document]
erights: The ses deps are fixed as you suggest, and the changes to yarn.lock are milder. PTAL.

erights:
> So I have 2 concerns after looking at this PR and its adoption in agoric-sdk (https://github.com/Agoric/agoric-sdk/pull/10795):
> 
> * Given the early nature of the non-trapping proposal, I would like SES and endo packages to avoid taking a hard dependency on it. Can we instead feature detect and use non-trapping only if available? Then we can move the shim application to `@endo/init`?
> * The changes to `harden` (and to some extend `passStyleOf`) are presenting a potential upgrade hazard. We generally consider `harden` to be part of the platform, which means it may get upgraded independently of the code using it. The problem is that we have some code with `harden` usages that are no longer valid after this change: any code that was hardening proxies or their target. That means we will not be able to use the same user code with updated XS and/or lockdown bundles that include this change. I am really not sure how we work around this one, besides accepting this breaking change / incompatibility, and require that platforms like agoric-sdk use "repaired" lockdown bundles to skip/undo this change if it needs to load incompatible user code.

The shim is now opt-in, i.e., disabled by default. It has no effect unless explicitly opting in to the new behavior via env-options `SES_NON_TRAPPING_SHIM=enabled`

packages/ses/src/make-hardener.js
```diff
@@ -128,6 +129,10 @@ const freezeTypedArray = array => {
  * @returns {Harden}
  */
 export const makeHardener = () => {
+  // TODO Get the native hardener to suppressTrapping at each step,
+  // rather than freeze. Until then, we cannot use it, which is *expensive*!
+  // TODO Comment out the following to skip the native hardener.
```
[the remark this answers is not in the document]
erights: The native hardener is now only bypassed if the shim is opted in to. Otherwise, the native hardener will continue to be used.

erights:
> Converting to draft while we wait for draft #2684 to settle down

No longer any need to wait on that.

erights:
> Converting to draft while we wait for draft #2684 to settle down

No longer any need to wait on that. This PR is again Ready for Review. PTAL

PR #2677, opened 2025-01-01
fix(ses): dont use native harden

erights:
Was staged on #2675 , but changed to standalone to isolate problems.

Closes: #XXXX
Refs: #XXXX

## Description

Old: Differs from #2675 only in that is doesn't use the native hardener if there, and therefore not on XS, because that one does not (yet?) support suppressTrapping.

This is expected to be too expensive for production, so currently this PR is only to see what breaks under CI, and why.

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations


erights:
Closing in favor of https://github.com/endojs/endo/pull/2675 , which incorporates the ideas from this PR

PR #2679, opened 2025-01-02
refactor: prepare for non-trapping integrity trait

erights:
Closes: #XXXX
Refs: https://github.com/Agoric/agoric-sdk/pull/10795

## Description

Prepare for anticipated introduction and use of the non-trapping integrity trait as explained at https://github.com/endojs/endo/blob/b12eb434b6672f0ceae41be55aac7f24c4562b7b/packages/ses/docs/preparing-for-stabilize.md

These preparations must work now, before these traits are introduced, and should continue to work after these traits are introduced and used.

### Security Considerations

Some things that had been deeply frozen automatically by `harden` are now manually frozen by explicit calls to `freeze`. We need to review these carefully to ensure that nothing has inadvertently be left unfrozen as a result of the changes in this PR.

Some proxies will become unhardenable, but they will still be hardenable as of now, so mistaken hardenings will not be detected.

### Scaling Considerations

For this PR by itself, none. Using the shim-based implementation of the non-trapping trait will have scaling consequences: https://github.com/endojs/endo/pull/2675

https://github.com/endojs/endo/blob/b12eb434b6672f0ceae41be55aac7f24c4562b7b/packages/ses/docs/preparing-for-stabilize.md will need to be reflected in developer docs.

### Testing Considerations

Since this PR by itself should be a pure refactor with no observable changes, there is nothing to test at this stage. The testing burden will come with https://github.com/endojs/endo/pull/2675 to see how adequate these preparations were.

The point. This changes to coding patterns that should be compat both with the current status quo and with https://github.com/endojs/endo/pull/2675

### Upgrade Considerations

As a pure refactor, none.

### DIFF
--- a/packages/eventual-send/src/E.js
+++ b/packages/eventual-send/src/E.js
@@ -2,7 +2,7 @@ import { trackTurns } from './track-turns.js';
 import { makeMessageBreakpointTester } from './message-breakpoints.js';
 
 const { details: X, quote: q, Fail, error: makeError } = assert;
-const { assign, create } = Object;
+const { assign, create, freeze } = Object;
 
 /**
  * @import { HandledPromiseConstructor } from './types.js';
@@ -171,6 +171,16 @@ const makeEGetProxyHandler = (x, HandledPromise) =>
  * @param {HandledPromiseConstructor} HandledPromise
  */
 const makeE = HandledPromise => {
+  /**
+   * `freeze` but not `harden` the proxy target so it remains trapping.
+   * @see https://github.com/endojs/endo/blob/master/packages/ses/docs/preparing-for-stabilize.md
+   */
+  const funcTarget = freeze(() => {});
+  /**
+   * `freeze` but not `harden` the proxy target so it remains trapping.
+   * @see https://github.com/endojs/endo/blob/master/packages/ses/docs/preparing-for-stabilize.md
+   */
+  const objTarget = freeze(create(null));
   return harden(
     assign(
       /**
@@ -183,7 +193,7 @@ const makeE = HandledPromise => {
        * @returns {ECallableOrMethods<RemoteFunctions<T>>} method/function call proxy
        */
       // @ts-expect-error XXX typedef
-      x => harden(new Proxy(() => {}, makeEProxyHandler(x, HandledPromise))),
+      x => new Proxy(funcTarget, makeEProxyHandler(x, HandledPromise)),
       {
         /**
          * E.get(x) returns a proxy on which you can get arbitrary properties.
@@ -196,11 +206,8 @@ const makeE = HandledPromise => {
          * @returns {EGetters<LocalRecord<T>>} property get proxy
          * @readonly
          */
-        get: x =>
-          // @ts-expect-error XXX typedef
-          harden(
-            new Proxy(create(null), makeEGetProxyHandler(x, HandledPromise)),
-          ),
+        // @ts-expect-error XXX typedef
+        get: x => new Proxy(objTarget, makeEGetProxyHandler(x, HandledPromise)),
 
         /**
          * E.resolve(x) converts x to a handled promise. It is
@@ -224,9 +231,7 @@ const makeE = HandledPromise => {
          */
         sendOnly: x =>
           // @ts-expect-error XXX typedef
-          harden(
-            new Proxy(() => {}, makeESendOnlyProxyHandler(x, HandledPromise)),
-          ),
+          new Proxy(funcTarget, makeESendOnlyProxyHandler(x, HandledPromise)),
 
         /**
          * E.when(x, res, rej) is equivalent to
--- a/packages/eventual-send/src/E.js
+++ b/packages/eventual-send/src/E.js
@@ -2,7 +2,7 @@ import { trackTurns } from './track-turns.js';
 import { makeMessageBreakpointTester } from './message-breakpoints.js';
 
 const { details: X, quote: q, Fail, error: makeError } = assert;
-const { assign, create, freeze } = Object;
+const { assign, freeze } = Object;
 
 /**
  * @import { HandledPromiseConstructor } from './types.js';
@@ -169,35 +169,20 @@ const makeEGetProxyHandler = (x, HandledPromise) =>
 
 /**
  * `freeze` but not `harden` the proxy target so it remains trapping.
- * This is safe to share between proxy instances because they are encapsulated
- * within the proxy.
- * - Before stabilize/suppressTrapping, this is safe
- *   because they are already frozen, and so they cannot be damaged by the
- *   proxies that encapsulate them.
- * - After stabilize/suppressTrapping, this is safe because the only damage
- *   that could be done would be by stabilize/suppressTrapping. These proxies
- *   do not explicitly provide such a trap, and thus will use the default
- *   behavior which is to refuse to be made non-trapping.
+ * Thus, it should not be shared outside this module.
  *
  * @see https://github.com/endojs/endo/blob/master/packages/ses/docs/preparing-for-stabilize.md
  */
 const funcTarget = freeze(() => {});
 
 /**
+/**
  * `freeze` but not `harden` the proxy target so it remains trapping.
- * This is safe to share between proxy instances because they are encapsulated
- * within the proxy.
- * - Before stabilize/suppressTrapping, this is safe
- *   because they are already frozen, and so they cannot be damaged by the
- *   proxies that encapsulate them.
- * - After stabilize/suppressTrapping, this is safe because the only damage
- *   that could be done would be by stabilize/suppressTrapping. These proxies
- *   do not explicitly provide such a trap, and thus will use the default
- *   behavior which is to refuse to be made non-trapping.
+ * Thus, it should not be shared outside this module.
  *
  * @see https://github.com/endojs/endo/blob/master/packages/ses/docs/preparing-for-stabilize.md
  */
-const objTarget = freeze(create(null));
+const objTarget = freeze({ __proto__: null });
 
 /**
  * @param {HandledPromiseConstructor} HandledPromise
--- a/packages/eventual-send/src/E.js
+++ b/packages/eventual-send/src/E.js
@@ -168,19 +168,41 @@ const makeEGetProxyHandler = (x, HandledPromise) =>
   });
 
 /**
+ * `freeze` but not `harden` the proxy target so it remains trapping.
+ * This is safe to share between proxy instances because they are encapsulated
+ * within the proxy.
+ * - Before stabilize/suppressTrapping, this is safe
+ *   because they are already frozen, and so they cannot be damaged by the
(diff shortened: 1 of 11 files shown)


packages/eventual-send/src/handled-promise.js
```diff
@@ -309,6 +309,9 @@ export const makeHandledPromise = () => {
         if (proxyOpts) {
           const {
             handler: proxyHandler,
+            // The proxy target can be frozen but should not be hardened
+            // so it remains trapping.
+            // See https://github.com/tc39/proposal-stabilize
```
erights, line 314: Where should this be documented?

packages/ses/docs/preparing-for-stabilize.md
```diff
@@ -0,0 +1,24 @@
...
+- The `get`, `set`, and `has` traps applied to a non-own property name are still general traps that can have useful trapping behavior.
+- The `apply` trap can ignore the target's call behavior and just do its own thing.
+
+However, to prepare for these changes, we need to avoid hardening both such proxies and their targets. We need to avoid hardening their target because this will bypass the traps. We need to avoid hardening the proxy because such proxies will *by default* refuse to be made non-trapping, and thus refuse to be hardened.
+
+## How passable objects should prepare
+
+Although we think of `passStyleOf` as requiring its input to be hardened, `passStyleOf` instead checked that each relevant object is frozen. Manually freezing all objects reachable from a root object had been equivalent to hardening that root object. With these changes, even such manual transitive freezing will not make an object passable. To prepare for these changes, use `harden` explicitly instead.
```
mhofman, line 24: Oh yeah I didn't think of that. That will potentially create a decent amount of churn on code that was manually freezing. It would be so much easier if frozen regular object would automatically pass trapping checks, but that would effectively transform non-trapping into a "is proxy" predicate.

packages/pass-style/test/passStyleOf.test.js
```diff
@@ -193,16 +193,22 @@ test('passStyleOf testing remotables', t => {
...
-  /** @type {any} UNTIL https://github.com/microsoft/TypeScript/issues/38385 */
-  const farObj2 = Object.freeze({
-    __proto__: tagRecord2,
-  });
+  /**
+   * TODO In order to run this test before we have explicit support for a
+   * non-trapping integrity trait, we have to `freeze` here but not `harden`.
+   * However, once we do have that support, and `passStyleOf` checks that
+   * its argument is also non-trapping, we still need to avoid `harden`
+   * because that would also hardden `__proto__`. So we will need to
```
mhofman, line 205:
```suggestion
   * because that would also harden `__proto__`. So we will need to
```

packages/pass-style/test/passStyleOf.test.js
```diff
@@ -193,16 +193,22 @@ test('passStyleOf testing remotables', t => {
...
+   * explicitly make this non-trapping, which we cannot yet express.
+   * @see https://github.com/endojs/endo/blob/master/packages/ses/docs/preparing-for-stabilize.md
```
mhofman, line 207: Ok I was confused at first, but this is a test that explicitly verifies that an object with a non-hardened proto is not passable. Can we make that more clear in the prelude of this comment?

packages/pass-style/test/passStyleOf.test.js
```diff
@@ -212,39 +218,27 @@ test('passStyleOf testing remotables', t => {
...
 
-  const tagRecord3 = Object.freeze(
-    makeTagishRecord('Alleged: both manually frozen'),
-  );
+  const tagRecord3 = harden(makeTagishRecord('Alleged: both manually frozen'));
   /** @type {any} UNTIL https://github.com/microsoft/TypeScript/issues/38385 */
-  const farObj3 = Object.freeze({
-    __proto__: tagRecord3,
-  });
+  const farObj3 = harden({ __proto__: tagRecord3 });
```
mhofman, line 223: I think this should get a TODO to be switched to both manually "non-trapping" instead of using `harden`.

packages/ses/src/strict-scope-terminator.js
```diff
@@ -21,7 +27,7 @@ const { Fail, quote: q } = assert;
  * create one and share it between all Proxy handlers.
  */
 export const alwaysThrowHandler = new Proxy(
-  immutableObject,
+  onlyFrozenObject,
   freeze({
     get(_shadow, prop) {
       Fail`Please report unexpected scope handler trap: ${q(String(prop))}`;
```
mhofman, line 33: So thinking about this, unless we do something wrong inside ses itself, we should never trigger this for the `isNonTrapping` trap, right? And even if we did, it would be a safe failure.

mhofman (commented):
Reading this and thinking through the implications, I have concerns we may be painting ourselves in a corner. Our recommendation is that one should harden values before passing them around. That implies we will never be able to use a `Proxy` as a remotable value. Is that something we're ok with? I know you have reservations with using `O` style objects as presences, but I'm a little worried about forever closing the door to this possibility.

packages/eventual-send/src/E.js
```diff
@@ -171,6 +171,16 @@ const makeEGetProxyHandler = (x, HandledPromise) =>
...
+  /**
+   * `freeze` but not `harden` the proxy target so it remains trapping.
+   * @see https://github.com/endojs/endo/blob/master/packages/ses/docs/preparing-for-stabilize.md
+   */
+  const funcTarget = freeze(() => {});
+  /**
+   * `freeze` but not `harden` the proxy target so it remains trapping.
+   * @see https://github.com/endojs/endo/blob/master/packages/ses/docs/preparing-for-stabilize.md
+   */
+  const objTarget = freeze(create(null));
```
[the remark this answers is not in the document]
erights: Done. By virtue of that safety, also lifted them to top level.

packages/pass-style/test/passStyleOf.test.js
erights: Done

packages/pass-style/test/passStyleOf.test.js
erights: Done

packages/pass-style/test/passStyleOf.test.js
erights: Done

packages/ses/src/strict-scope-terminator.js
erights: Seems so. So no suggest change, right? Resolving.

erights:
ping

erights:
Reviewers: I claim that this PR by itself prior to the introduction or use of the non-trapping or stabilize integrity traits is a ***pure*** refactor, i.e., it should not produce any observable changes. Please review with that claim in mind. If you do spot something that looks like an observable change, please let me know. Thanks!

packages/pass-style/test/passStyleOf.test.js
```diff
@@ -13,7 +13,7 @@ const harden = /** @type {import('ses').Harden & { isFake?: boolean }} */ (
   global.harden
 );
 
-const { getPrototypeOf, defineProperty } = Object;
+const { getPrototypeOf, defineProperty, freeze } = Object;
```
michaelfig, line 16: I get dizzy with all the newly-introduced TODOs.  Can we instead just alias `suppressTrapping` to `harden` once per module?  If you want to leave a TODO, maybe write it once per module in that place?

Here's a suggestion, but I am not attached to how you actually code this if you like the idea behind it:

```suggestion
const { getPrototypeOf, defineProperty, freeze, suppressTrapping = harden } = Object;
```

packages/pass-style/test/passStyleOf.test.js
```diff
@@ -193,16 +197,27 @@ test('passStyleOf testing remotables', t => {
 
   const tagRecord1 = harden(makeTagishRecord('Alleged: manually constructed'));
   /** @type {any} UNTIL https://github.com/microsoft/TypeScript/issues/38385 */
-  const farObj1 = harden({
-    __proto__: tagRecord1,
-  });
+  // TODO Once we have support for explicit non-trapping, this should change
+  // from `harden` to `suppressTrapping` or `stabilize`.
+  const farObj1 = harden({ __proto__: tagRecord1 });
```
michaelfig, line 202: Please remove the `TODO` noise?

```suggestion
  const farObj1 = suppressTrapping({ __proto__: tagRecord1 });
```

... and so on for every `suppressTrapping` TODO.

michaelfig (commented):
Please consider my suggestion for reducing some TODOs.

packages/pass-style/test/passStyleOf.test.js
erights: Much better, thanks!

Done

packages/pass-style/test/passStyleOf.test.js
erights: Done

packages/pass-style/test/passStyleOf.test.js
```diff
@@ -13,7 +13,35 @@ const harden = /** @type {import('ses').Harden & { isFake?: boolean }} */ (
...
+ * do not yet depend on the final name.
+ *
+ * TODO Once we do have support for an explicit `suppressTrapping` operation,
+ * we should import that instead, and if necessary rename all uses to that
+ * operation's final name.
+ */
+const hardenToBeSuppressTrapping = harden;
+
+/**
+ * Local alias of `harden` to eventually be switched to whatever applies
```
michaelfig, line 32:
```suggestion
 * Local alias of `freeze` to eventually be switched to whatever applies
```

packages/pass-style/test/passStyleOf.test.js
```diff
@@ -387,7 +410,22 @@ test('remotables - safety from the gibson042 attack', t => {
...
+   *
+   * TODO In order to run this test before we have explicit support for a
+   * non-trapping integrity trait, we have to `freeze` here but not `harden`.
+   * However, once we do have that support, and `passStyleOf` checks that
+   * its argument is also non-trapping, we still need to avoid `harden`
+   * because that would also harden `__proto__`. So we will need to
+   * explicitly make this non-trapping, which we cannot yet express.
+   * @see https://github.com/endojs/endo/blob/master/packages/ses/docs/preparing-for-stabilize.md
+   */
+  const makeInput = () => freeze({ __proto__: mercurialProto });
```
michaelfig, line 428:
```suggestion
  const makeInput = () => freezeToBeSuppressTrapping({ __proto__: mercurialProto });
```

michaelfig (approved):
LGTM after these two suggestions, which I expect you'll address somehow before merging.

packages/pass-style/test/passStyleOf.test.js
erights: Done

packages/pass-style/test/passStyleOf.test.js
erights: Good catch! I missed that one. Done

Issue #2660, opened 2024-12-11
Flakey browser tests (playwright)


leotm:
## Describe the bug

not a required job, but should deal with it some point
- https://github.com/endojs/endo/pull/2583#issuecomment-2408156067
- https://github.com/endojs/endo/pull/2563#issuecomment-2416818991
- https://github.com/endojs/endo/pull/2655#pullrequestreview-2496396182
- https://github.com/endojs/endo/pull/2659
- etc

```console
Run npx playwright install --with-deps
  npx playwright install --with-deps
  shell: /usr/bin/bash -e {0}
Installing dependencies...
Switching to root user to install dependencies...
Get:1 file:/etc/apt/apt-mirrors.txt Mirrorlist [14[2](https://github.com/endojs/endo/actions/runs/12281686710/job/34271115469?pr=2655#step:9:2) B]
Hit:2 http://azure.archive.ubuntu.com/ubuntu noble InRelease
Get:[3](https://github.com/endojs/endo/actions/runs/12281686710/job/34271115469?pr=2655#step:9:3) http://azure.archive.ubuntu.com/ubuntu noble-updates InRelease [126 kB]
Hit:6 https://packages.microsoft.com/repos/azure-cli noble InRelease
Get:[4](https://github.com/endojs/endo/actions/runs/12281686710/job/34271115469?pr=2655#step:9:5) http://azure.archive.ubuntu.com/ubuntu noble-backports InRelease [126 kB]
Get:[5](https://github.com/endojs/endo/actions/runs/12281686710/job/34271115469?pr=2655#step:9:6) http://azure.archive.ubuntu.com/ubuntu noble-security InRelease [12[6](https://github.com/endojs/endo/actions/runs/12281686710/job/34271115469?pr=2655#step:9:7) kB]
Hit:7 https://packages.microsoft.com/ubuntu/24.04/prod noble InRelease
Get:8 http://azure.archive.ubuntu.com/ubuntu noble-updates/main amd64 Packages [[7](https://github.com/endojs/endo/actions/runs/12281686710/job/34271115469?pr=2655#step:9:8)01 kB]
Get:9 http://azure.archive.ubuntu.com/ubuntu noble-updates/main Translation-en [162 kB]
Get:10 http://azure.archive.ubuntu.com/ubuntu noble-updates/main amd64 Components [132 kB]
Get:11 http://azure.archive.ubuntu.com/ubuntu noble-updates/universe amd64 Packages [727 kB]
[...]

PR #2686, opened 2025-01-10
Fix browser tests

naugtur:
Browser tests are failing to install browsers for playwright since ubuntu version was updated in the runners. I'll try upgrading versions of playwright and related, or downgrade the os image for now.

### DIFF
--- a/browser-test/package-lock.json
+++ b/browser-test/package-lock.json
@@ -8,23 +8,24 @@
       "name": "browser-test",
       "license": "Apache-2.0",
       "devDependencies": {
-        "@playwright/test": "^1.41.2",
+        "@playwright/test": "^1.49.1",
         "@types/node": "^20.11.16"
       }
     },
     "node_modules/@playwright/test": {
-      "version": "1.41.2",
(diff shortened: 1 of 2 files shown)


kriskowal (approved):
Thanks. I’m not gravely concerned about the prominence of the error. We didn’t fail to notice. And, given that this shifts underneath us and isn’t grave enough to block other work from continuing, I’m inclined to keep the check optional.

leotm:
- fixes https://github.com/endojs/endo/issues/2660
- (supersedes) closes https://github.com/endojs/endo/pull/2587

leotm (approved):
- fixes https://github.com/endojs/endo/issues/2660 (updated description)
- (supersedes) closes https://github.com/endojs/endo/pull/2587

naugtur:
@kriskowal I'd like to put some pressure on making it non-optional because the fix was trivial and its coverage of important SES compatibility is going to increase:
https://github.com/endojs/endo/pull/2687

I've added coverage for CSP and also a convenient display of new intrinsics SES is deleting in each browser.

PR #2693, opened 2025-01-22
fix(ses): update permits for stage 2.7.4 proposals

erights:
Closes: #XXXX
Refs: https://github.com/tc39/proposals

## Description

The `permits.js` file should stay reasonably up to date with https://github.com/tc39/proposals stage 4, 3, and 2.7 proposals. Those it is not yet ready to list, it should at least include in a TODO comment. For non-ancient proposals, each such entry should cite the proposal url.

### Security Considerations

The entire reason for the permits mechanism is so that elements added to the primordials are not enabled for Hardened JS until the Hardened JS stewards have vetted them. This is mostly by having vetted the specs during the proposal process, ***assuming*** that the implementations follow the spec correctly enough for security purposes. When we later find that an implementation has not, this is a separate security concern to be dealt with then, probably by repair, but conceivably by withdrawing the permit.

***Reviewers***: Therefore, when reviewing each newly permitted entry, please think about whether it may threaten any security property of Hardened JS.

### Scaling Considerations

none

The closer we stay to permitting the JS that is shipping, and that developers have come to expect, the fewer surprises we need to explain.
### Testing Considerations

Not sure what would help. Suggestions welcome.

Code that was feature testing for one of these new features might find one present that had previously been absent. If both branches of the code are correct, well, that is why it feature tests, so all is happy. However, the newly taken branch may have been previously untested.

### Upgrade Considerations

See the discussion above about feature testing. Such code, even when correct on both branches, may fail to deterministically replay a previous run done without the feature.

### DIFF
--- a/packages/ses/src/permits.js
+++ b/packages/ses/src/permits.js
@@ -98,10 +98,6 @@ export const universalPropertyNames = {
   // TODO ShadowRealm
   // ShadowRealm: 'ShadowRealm',
 
-  // https://github.com/tc39/proposal-source-phase-imports
-  // TODO %AbstractModuleSource%, ModuleSource?
-  // ModuleSource: 'ModuleSource',
-
   // *** Other Properties of the Global Object
 
   JSON: 'JSON',
--- a/packages/ses/src/permits.js
+++ b/packages/ses/src/permits.js
@@ -98,6 +98,10 @@ export const universalPropertyNames = {
   // TODO ShadowRealm
   // ShadowRealm: 'ShadowRealm',
 
+  // https://github.com/tc39/proposal-source-phase-imports
+  // TODO %AbstractModuleSource%, ModuleSource?
+  // ModuleSource: 'ModuleSource',
+
   // *** Other Properties of the Global Object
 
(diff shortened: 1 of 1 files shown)


packages/ses/src/permits.js
```diff
@@ -89,6 +89,19 @@ export const universalPropertyNames = {
...
+  // DisposableStack: 'DisposableStack',
+  // AsyncDisposableStack: 'AsyncDisposableStack',
+
+  // https://tc39.es/proposal-shadowrealm/
+  // TODO ShadowRealm
+  // ShadowRealm: 'ShadowRealm',
+
+  // https://github.com/tc39/proposal-source-phase-imports
+  // TODO %AbstractModuleSource%, ModuleSource?
+  // ModuleSource: 'ModuleSource',
```
kriskowal, line 103: `ModuleSource` already appears below.

kriskowal (approved):

packages/ses/src/permits.js
erights: Good. Gone.

PR #2696, opened 2025-01-23
chore: Release 2025-01-23

kriskowal:
# ses v1.11.0

- Adds support for dynamic `import` in conjunction with an update to
  `@endo/module-source`.

- Specifying the long-discontinued `mathTaming` or `dateTaming` options logs a
  warning.

Incubating: Please do not rely on these features as they are under development
and subject to breaking changes that will not be signaled by semver.

- Adds support for an XS-specific variant of the SES shim that is triggered
  with the `xs` package export condition.
  This version of SES preserves all the features of `Compartment` provided
  uniquely by the SES shim, but with the `__native__` constructor option,
  loses support for importing precompiled module records and gains support
  for native `ModuleSource`.

# @endo/module-source v1.2.0

- Supports dynamic `import` within a `ModuleSource` in conjunction with
  a related change in `ses`.
  For example, `await import(specifier)` can now call through to the
  surrounding compartment's `importHook` to load and evaluate further modules.
- Provides an XS-specific variant of `@endo/module-source` that adapts the
  native `ModuleSource` instead of entraining Babel.

# @endo/compartment-mapper v1.5.0

- `mapNodeModules` and all functions that use it now tolerate the absence of
  expected packages.
  These packages are now omitted from the generated package skeleton map.
  So, loading a physically missing module now occurs during the load phase
  instead of the mapping phase.
- Adds a `strict` option to all functions that `mapNodeModules` to restore old
  behavior, which produces an error early if, for example, a non-optional
  peer dependency is missing.
  Peer dependencies are strictly required unless `peerDependenciesMeta` has an
  object with a truthy `optional` entry.
  Correct interpretation of `peerDependencies` is not distributed evenly, so
  this behavior is no longer the default.

Incubating: Please do not rely on these features as they are under development
and subject to breaking changes that will not be signaled by semver.

- The module `@endo/compartment-mapper/import-archive-parsers.js` does not
  support modules in archives in their original ESM (`mjs`) or CommonJS (`cjs`)
  formats because they entrain Babel and a full JavaScript lexer that are
  not suitable for use in all environments, specifically XS.
  This version introduces an elective
  `@endo/compartment-mapper/import-archive-all-parsers.js` that has all of the
  precompiled module parsers (`pre-cjs-json` and `pre-mjs-json`) that Endo's
  bundler currently produces by default and additionally parsers for original
  sources (`mjs`, `cjs`).
  Also, provided the `xs` package condition,
  `@endo/compartment-mapper/import-archive-parsers.js` now falls through to the
  native `ModuleSource` and safely includes `mjs` and `cjs` without entraining
  Babel, but is only supported in conjunction with the `__native__` option
  for `Compartment`, `importArchive`, `parseArchive`, and `importBundle`.
  With the `node` package condition (present by default when running ESM on
  `node`), `@endo/compartment-mapper/import-archive-parsers.js` also now
  includes `mjs` and `cjs` by entraining Babel, which performs adequately on
  that platform.
- Adds a `__native__: true` option to all paths to import, that indicates that
  the application will fall through to the native implementation of
  Compartment, currently only available on XS, which lacks support for
  precompiled module sources (as exist in many archived applications,
  particularly Agoric smart contract bundles) and instead supports loading
  modules from original sources (which is not possible at runtime on XS).

### DIFF
--- a/packages/eslint-plugin/CHANGELOG.md
+++ b/packages/eslint-plugin/CHANGELOG.md
@@ -3,6 +3,15 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [2.3.0](https://github.com/endojs/endo/compare/@endo/eslint-plugin@2.2.3...@endo/eslint-plugin@2.3.0) (2025-01-24)
+
+
+### Features
+
+* **ses:** Add XS variant of shim ([f6c8456](https://github.com/endojs/endo/commit/f6c84566bb6a698709dc3474726000f07b94f3db))
+
+
+
 ### [2.2.3](https://github.com/endojs/endo/compare/@endo/eslint-plugin@2.2.2...@endo/eslint-plugin@2.2.3) (2024-11-13)
 
 **Note:** Version bump only for package @endo/eslint-plugin
--- a/packages/eventual-send/CHANGELOG.md
+++ b/packages/eventual-send/CHANGELOG.md
@@ -3,6 +3,15 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [1.3.0](https://github.com/endojs/endo/compare/@endo/eventual-send@1.2.8...@endo/eventual-send@1.3.0) (2025-01-24)
+
+
+### Features
+
+* EReturn type ([5aefe10](https://github.com/endojs/endo/commit/5aefe1032be33c70c097663a86f240c857dcd2a1))
+
+
+
 ### [1.2.8](https://github.com/endojs/endo/compare/@endo/eventual-send@1.2.7...@endo/eventual-send@1.2.8) (2024-11-13)
 
 **Note:** Version bump only for package @endo/eventual-send
--- a/packages/bundle-source/CHANGELOG.md
+++ b/packages/bundle-source/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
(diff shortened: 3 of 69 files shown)


packages/eventual-send/CHANGELOG.md
```diff
@@ -3,6 +3,15 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [1.3.0](https://github.com/endojs/endo/compare/@endo/eventual-send@1.2.8...@endo/eventual-send@1.3.0) (2025-01-23)
+
+
+### Features
+
+* EReturn type ([5aefe10](https://github.com/endojs/endo/commit/5aefe1032be33c70c097663a86f240c857dcd2a1))
```
mhofman, line 11: Surprised that we labelled a type change with a `feat` causing a new minor version

packages/eslint-plugin/CHANGELOG.md
```diff
@@ -3,6 +3,15 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [2.3.0](https://github.com/endojs/endo/compare/@endo/eslint-plugin@2.2.3...@endo/eslint-plugin@2.3.0) (2025-01-23)
+
+
+### Features
+
+* **ses:** Add XS variant of shim ([f6c8456](https://github.com/endojs/endo/commit/f6c84566bb6a698709dc3474726000f07b94f3db))
```
mhofman, line 11: That PR probably shouldn't have been squashed to avoid a minor version bump of eslint-plugin

mhofman (approved):
Seems mechanical enough, even though the mechanics provide some slightly surprising results

packages/eslint-plugin/CHANGELOG.md
kriskowal: In this case, I simply didn’t have the foresight to separate the eslint change into its own conventional commit. I’ll have to watch for that in the future.

PR #2701, opened 2025-01-27
fix(pass-style): fix #2700 ignore more safe async_hook extra properties

### LINKED ISSUE #2700: In vscode debug session, node async_hooks promise keys appear too late
## Describe the bug

Node's async_hooks apparently adds symbol-named own properties to `Promise.prototype` and to (some?) individual promises. For `passStyleOf` to judge a promise passable, ideally it should not have any own properties. However, to accommodate the own properties added by Node's async_hooks to an individual promise, `checkPromiseOwnKeys` allows symbol-named own properties if they also already appear on `Promise.prototype` and the properties are visibly safe enough(*).

However, under the vscode debugger, entered using vscode JavaScript Debug Terminal, I have several times run into the 
```
Must not have any own properties: ["[Symbol(async_id_symbol)]","[Symbol(trigger_async_id_symbol)]"]
```
error, apparently because async_hooks added these to an individual promise instance but not to `Promise.prototype`.

(*) This should test that the property is a non-configurable non-writable own data property with a visibly harmless value, such as a primitive. However, currently the test seems weaker than that and should be tightened.

## Steps to reproduce

Check out a current endo and `yarn && yarn build`

In vscode, set a breakpoint at https://github.com/endojs/endo/blob/895ea0a4ef4f930cc7b1fa1286107bec7af5b904/packages/ses/src/error/assert.js#L373 

```js
  // The next line is a particularly fruitful place to put a breakpoint.
  return error;
```

In a vscode JavaScript Debug Terminal, run

```sh
$ cd packages/cli
$ yarn test test/demo/index.test.js
```

If the debugger then runs until it stops in the state captured below, you've reproduced the problem.

(Note: Currently speculative that this happens on endo master. Currently I'm on a draft branch with changes that should make no difference. But not sure.)

## Expected behavior

Not hitting this error

## Platform environment

 - What OS are you using?
   - MacOS 15.1.1 (24B91)
 - What version of Node.js?
   - v18.19.0
 - Is there anything special/unusual about your platform?
   - not that I can think of, that could possibly be relevant.
 - What version of Endo are you using? (run `git describe --tags --always`)
   - @endo/bundle-source@3.5.1-13-geb96f7feb

## Additional context

I am currently working on a local draft of the `markm-flip-non-trapping-shim-default` branch. I will check later if this also happens on master.

## Screenshots

```sh
cli(markm-flip-non-trapping-shim-default)$ yarn test test/demo/index.test.js
```

```
Error: "[Promise]" - Must not have any own properties: ["[Symbol(async_id_symbol)]","[Symbol(trigger_async_id_symbol)]"]
  at makeError (file:///Users/markmiller/src/ongithub/endojs/endo/packages/ses/src/error/assert.js:359:61)
  at fail (file:///Users/markmiller/src/ongithub/endojs/endo/packages/ses/src/error/assert.js:491:20)
  at baseAssert (file:///Users/markmiller/src/ongithub/endojs/endo/packages/ses/src/error/assert.js:513:13)
  at assertChecker (file:///Users/markmiller/src/ongithub/endojs/endo/packages/pass-style/src/passStyle-helpers.js:78:3)
  at reject (file:///Users/markmiller/src/ongithub/endojs/endo/packages/pass-style/src/passStyle-helpers.js:93:34)
  at checkPromiseOwnKeys (file:///Users/markmiller/src/ongithub/endojs/endo/packages/pass-style/src/safe-promise.js:42:6)
```

```js
const checkPromiseOwnKeys = (pr, check) => {
  const keys = ownKeys(pr);

  if (keys.length === 0) {
    return true;
  }

  /**
   * This excludes those symbol-named own properties that are also found on
   * `Promise.prototype`, so that overrides of these properties can be
   * explicitly tolerated if they pass the `checkSafeOwnKey` check below.
   * In particular, we wish to tolerate
   *   * An overriding `toStringTag` non-enumerable data property
   *     with a string value.
   *   * Those own properties that might be added by Node's async_hooks.
   */
  const unknownKeys = keys.filter(
  );

  if (unknownKeys.length !== 0) {
    return CX(
      check,
    )`${pr} - Must not have any own properties: ${q(unknownKeys)}`;
  }
  // ...
};
```

In the vscode Debug Console when looking at the `checkPromiseOwnKeys` stack frame

```
Object.getOwnPropertyDescriptors(Promise.prototype)
{constructor: {…}, then: {…}, catch: {…}, finally: {…}, Symbol(Symbol.toStringTag): {…}}
  catch = {value: ƒ, writable: false, enumerable: false, configurable: false}
  constructor = {get: ƒ, set: ƒ, enumerable: false, configurable: false}
  finally = {value: ƒ, writable: false, enumerable: false, configurable: false}
  Symbol(Symbol.toStringTag) = {value: 'Promise', writable: false, enumerable: false, configurable: false}
  then = {value: ƒ, writable: false, enumerable: false, configurable: false}
  toString = ƒ [prop]() {\n            return value;\n          }
  valueOf = ƒ [prop]() {\n            return value;\n          }
```

```
Object.getOwnPropertyDescriptors(pr)
{Symbol(async_id_symbol): {…}, Symbol(trigger_async_id_symbol): {…}}
  Symbol(async_id_symbol) = {value: 15589, writable: false, enumerable: true, configurable: false}
  Symbol(trigger_async_id_symbol) = {value: 15579, writable: false, enumerable: true, configurable: false}
  toString = ƒ [prop]() {\n            return value;\n          }
  valueOf = ƒ [prop]() {\n            return value;\n          }
```

erights:
Closes: #2700 
Refs: #2708

## Description

#2700 explains how `passStyleOf`'s criteria for judging a promise to be passable happens to cause some failures when run under the vscode debugger. This both loosens the criteria to accept those properties that async_hooks might add to an individual promise instance, and tightens the criteria to enforce that these properties are symbol-named non-configurable non-writable own data properties with values that are obviously safe, such as primitives.

Because `@@toStringTag` would normally satisfy this more general criteria, and doing so would guarantee it safe enough for all known purposes, this PR no longer makes a special case for it. By folding it into the more general check above, we no longer enforce that this property is non-enumerable, nor that the value of this property is a string.

This is an alternate approach to #2708 

### Security Considerations

The tightened criteria above enhance safety. But the loosenings above endanger safety, and so must be carefully considered. This loosening allows properties with any symbol name, whether or not async_hooks would ever add them. Further, it loosens the restrictions on the `@@toStringTag`-named property in ways that code might conceivably find confusing.

***Reviewers***, please form your own judgement about whether such loosening actually matters. I will tighten back up whatever you request.

### Scaling Considerations

none

none for normal programmers.

For use of a debugger, this PR only removes an unpleasant surprise that would otherwise need to be explained.

### Testing Considerations

none for non-debugger-based testing.

Both the tightenings and the loosenings are conceivably compat breaks, but are highly unlikely to matter in practice.
### Upgrade Considerations

none.

### DIFF
--- a/packages/pass-style/test/safe-promise.test.js
+++ b/packages/pass-style/test/safe-promise.test.js
@@ -18,7 +18,8 @@ test('safe promise loophole', t => {
     const p2 = Promise.resolve('p2');
     p2.silly = 'silly own property';
     t.throws(() => passStyleOf(harden(p2)), {
-      message: '"[Promise]" - Must not have any own properties: ["silly"]',
+      message:
+        '"[Promise]" - Must not have any string-named own properties: ["silly"]',
     });
     t.is(p2[toStringTag], 'Promise');
     t.is(`${p2}`, '[object Promise]');
@@ -39,9 +40,7 @@ test('safe promise loophole', t => {
     defineProperty(p3, toStringTag, {
       value: 3,
     });
-    t.throws(() => passStyleOf(harden(p3)), {
-      message: 'Own @@toStringTag value must be a string: 3',
-    });
+    t.is(passStyleOf(harden(p3)), 'promise');
   }
 
   {
@@ -50,10 +49,7 @@ test('safe promise loophole', t => {
       value: 'Promise p4',
       enumerable: true,
     });
-    t.throws(() => passStyleOf(harden(p4)), {
-      message:
-        'Own @@toStringTag must not be enumerable: {"configurable":false,"enumerable":true,"value":"Promise p4","writable":false}',
-    });
+    t.is(passStyleOf(harden(p4)), 'promise');
 
     const p5 = Promise.resolve('p5');
     defineProperty(p5, toStringTag, {
--- a/packages/pass-style/src/safe-promise.js
+++ b/packages/pass-style/src/safe-promise.js
@@ -2,15 +2,86 @@
 
 import { isPromise } from '@endo/promise-kit';
 import { q } from '@endo/errors';
-import { assertChecker, hasOwnPropertyOf, CX } from './passStyle-helpers.js';
+import {
+  assertChecker,
+  hasOwnPropertyOf,
+  CX,
+  isObject,
+} from './passStyle-helpers.js';
 
 /** @import {Checker} from './types.js' */
 
 const { isFrozen, getPrototypeOf, getOwnPropertyDescriptor } = Object;
 const { ownKeys } = Reflect;
-const { toStringTag } = Symbol;
 
 /**
+ * Explicitly tolerate symbol-named non-configurable non-writable data
+ * property whose value is obviously harmless, such as a primitive value.
+ *
+ * The motivations are to tolerate `@@toStringTag` and those properties
+ * that might be added by Node's async_hooks. Thus, beyond primitives, the
+ * only values that must be tolerated are those safe values that might be
+ * added by async_hooks.
+ *
+ * At the time of this writing, Node's async_hooks contains the
+ * following code, which we need to tolerate if safe:
+ *
+ * ```js
+ * function destroyTracking(promise, parent) {
+ *   trackPromise(promise, parent);
+ *   const asyncId = promise[async_id_symbol];
+ *   const destroyed = { destroyed: false };
+ *   promise[destroyedSymbol] = destroyed;
+ *   registerDestroyHook(promise, asyncId, destroyed);
+ * }
+ * ```
+ *
+ * @param {object} obj
+ * @param {string|symbol} key
+ * @param {Checker} check
+ */
+const checkSafeOwnKeyOf = (obj, key, check) => {
+  const desc = getOwnPropertyDescriptor(obj, key);
+  assert(desc);
+  const quoteKey = q(String(key));
+  if (!hasOwnPropertyOf(desc, 'value')) {
+    return CX(
+      check,
+    )`Own ${quoteKey} must be a data property, not an accessor: ${obj}`;
+  }
+  const { value, writable, configurable } = desc;
+  if (writable) {
+    return CX(check)`Own ${quoteKey} must not be writable: ${obj}`;
+  }
+  if (configurable) {
+    return CX(check)`Own ${quoteKey} must not be configurable: ${obj}`;
+  }
+  if (!isObject(value)) {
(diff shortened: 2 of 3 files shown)


packages/pass-style/test/safe-promise.test.js
```diff
@@ -39,9 +40,7 @@ test('safe promise loophole', t => {
     defineProperty(p3, toStringTag, {
       value: 3,
     });
-    t.throws(() => passStyleOf(harden(p3)), {
-      message: 'Own @@toStringTag value must be a string: 3',
-    });
+    t.is(passStyleOf(harden(p3)), 'promise');
```
erights, line 43: The change to this test case demonstrates the loosening of the `@@toStringTag` tests

packages/pass-style/test/safe-promise.test.js
```diff
@@ -50,10 +49,7 @@ test('safe promise loophole', t => {
       value: 'Promise p4',
       enumerable: true,
     });
-    t.throws(() => passStyleOf(harden(p4)), {
-      message:
-        'Own @@toStringTag must not be enumerable: {"configurable":false,"enumerable":true,"value":"Promise p4","writable":false}',
-    });
+    t.is(passStyleOf(harden(p4)), 'promise');
```
erights, line 52: The change to this test case demonstrates the loosening of the `@@toStringTag` tests

erights:
Because the observable change of #2700 that this PR allegedly fixes is only observed interactively when running the vscode debugger, this PR does not add a test case verifying the fix. Rather, I manually verified that this PR fixes #2700 by running the steps it documents. That run got well past the point where the previous problem was encountered, but then causes #2702 , which I suspect and hope is unrelated.

mhofman:
I'm gonna have to look at this closer because the async hooks taming was supposed to handling these things

mhofman (changes requested):
I believe the orignal checks are fine, it's just that they rely on `@endo/init` to have applied the node async_hooks patch that tames things.

See https://github.com/endojs/endo/issues/2700#issuecomment-2626231686

kriskowal:
@mhofman Please confirm we can close this, since we made async hook patches opt-in?

erights:
IMO it can be closed as obsolete, but I'll defer to @mhofman

PR #2705, opened 2025-01-27
feat(compartment-mapper): Make bundle using named evaluator

kriskowal:
Refs: #2444 

## Description

Toward better preservation of formatting through the ModuleSource and censorship evasion transforms (#2444), we ran into difficulty in composition with Rollup and source map transformations. The simplest route forward is to replace the Rollup implementation behind `getExport` and `nestedEvaluate` bundle formats with Endo’s own bundler. To that end, the `nestedEvaluate` format needs to be able to evaluate each module functor with the _named evaluate_ function instead of evaluating them inline. This change introduces a `useNamedEvaluate` option to `makeBundle`, which facilitates the `nestedEvaluate` bundle format.

### Security Considerations

The `nestedEvaluate` bundle format runtime in `importBundle` is obliged to provide a confined evaluator and also confine the evaluation of the bundle proper, as the intent is to confine the bundle.

### Scaling Considerations

None.

In a subsequent change, where we carve out the implementation of `nestedEvaluate`, there must be notes about the change in behavior this implementation will cause over the prior Rollup.

### Testing Considerations

This change includes a pair of tests to cover the bundler behavior and also to verify that the resulting evaluation preserves line numbers in error stacks.

This is purely additive at this layer, but does not perfectly emulate the behavior of Rollup.

### Upgrade Considerations

When this change becomes incorporated in `bundleSource`, compensating changes are required in the contemporary versions of Agoric SDK kernel components. These changes have already landed. https://github.com/Agoric/agoric-sdk/pull/10878 https://github.com/Agoric/agoric-sdk/pull/10877

### DIFF
--- a/packages/compartment-mapper/test/bundle.test.js
+++ b/packages/compartment-mapper/test/bundle.test.js
@@ -69,6 +69,54 @@ test('bundles work', async t => {
   t.deepEqual(log, expectedLog);
 });
 
+test('using named evaluate bundles work', async t => {
+  const bundle = await makeBundle(read, fixture, {
+    useNamedEvaluate: 'nestedEvaluate',
+  });
+  const log = [];
+  const print = entry => {
+    log.push(entry);
+  };
+  const compartment = new Compartment({
+    globals: {
+      print,
+      nestedEvaluate(source) {
+        return compartment.evaluate(source);
+      },
+    },
+    __options__: true,
+  });
+  compartment.evaluate(bundle);
+  t.deepEqual(log, expectedLog);
+});
+
+test('using named evaluate bundles preserve error line numbers', async t => {
+  const bundle = await makeBundle(read, fixture, {
+    useNamedEvaluate: 'nestedEvaluate',
+  });
+  const log = [];
+  const print = entry => {
+    log.push(entry);
+  };
+  const compartment = new Compartment({
+    globals: {
+      print,
+      nestedEvaluate(source) {
+        return compartment.evaluate(source);
+      },
+    },
+    __options__: true,
+  });
+  const { raise } = compartment.evaluate(bundle);
+  let error = null;
+  try {
+    raise();
+  } catch (_error) {
+    error = _error;
+  }
+  t.assert(error.stack.includes(':4:'));
+});
+
 test('equivalent archive behaves the same as bundle', async t => {
   const log = [];
   const print = entry => {
--- a/packages/compartment-mapper/NEWS.md
+++ b/packages/compartment-mapper/NEWS.md
@@ -1,5 +1,11 @@
 User-visible changes to `@endo/compartment-mapper`:
 
+# Next release
+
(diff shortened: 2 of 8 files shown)


kriskowal:
attn @naugtur since you have a hand in the maintenance of the compartment-mapper/bundle.js feature.

packages/compartment-mapper/test/bundle.test.js
```diff
@@ -69,6 +69,54 @@ test('bundles work', async t => {
   t.deepEqual(log, expectedLog);
 });
 
+test('using named evaluate bundles work', async t => {
```
gibson042, line 72: These tests are very helpful in understanding the new functionality, but documentation would also be appreciated.

packages/compartment-mapper/test/bundle.test.js
```diff
@@ -69,6 +69,54 @@ test('bundles work', async t => {
...
+  const log = [];
+  const print = entry => {
+    log.push(entry);
+  };
+  const compartment = new Compartment({
+    globals: {
+      print,
+      nestedEvaluate(source) {
+        return compartment.evaluate(source);
+      },
```
gibson042, line 85: Assuming this is the expected use pattern, are there concerns about the endowment being retrieved and possibly exploited by source code? Although the options for addressing that are limited and messy.

gibson042 (approved):
This works, although I find myself wondering if `${useNamedEvaluate}(${JSON.stringify(cjsFunctor)})` is _too_ general. Are there anticipated use cases for something other than recursive evaluation?

packages/compartment-mapper/test/bundle.test.js
kriskowal: Wherein the protagonist discovers that there’s no documentation of makeBundle.

kriskowal:
These are good questions.

> This works, although I find myself wondering if `${useNamedEvaluate}(${JSON.stringify(cjsFunctor)})` is _too_ general. Are there anticipated use cases for something other than recursive evaluation?

> Assuming this is the expected use pattern, are there concerns about the endowment being retrieved and possibly exploited by source code? Although the options for addressing that are limited and messy.

It might be too general. Expected values would be `"eval"` or _like_ `"evaluate"` if you use a compartment and endow it with an `evaluate` function. We aren’t constraining the name of the evaluator here, and maybe could and should. In subsequent changes, I might generalize `makeBundle` so it can curry an options bag instead of grabbing the named evaluate out of lexical scope. But, for the SES dist bundles, it’s important that the completion value of the the bundle is discardable. For `nestedEvaluate`, on the other hand, it’s useful to be able to inject a `sourceUrlPrefix` at runtime (coming soon). It might be that we can limit `useNamedEvaluate` to the case where the options bag gets curried, and then the name becomes irrelevant: it’s injected.

I have your stamp, but I will hold this open and stack subsequent, non-orthogonal changes so we can revisit.

kriskowal:
Subsumed into https://github.com/endojs/endo/pull/2707

PR #2707, opened 2025-01-30
feat(compartment-mapper): Make script or functor bundles

kriskowal:
Refs: #2444 

## Description

Toward better preserving the format of original source through the `ModuleSource` and censorship evasion transforms, we upgraded Babel and use a new code generator. The new code generator does not compose with Rollup. However, in the intervening years, we have reïmplemented every part of Rollup we find desirable and better in keeping with our integrity requirements. So, we are poised to reïmplement the `nestedEvaluate` and `getExport` formats using our own implementation. To that end, this change closes the feature parity gap needed to undertake that refactor.

We refactor `makeBundle` into `makeScript` and `makeFuctor`. We effectively rename `makeBundle` to `makeScript` for clarity. The `script` form surfaces as the `endoScript` bundler format and we expect `functor` to surface in an `endoFunctor` format, in addition to replacing the implementation of `nestedEvaluate` and `getExport`.

Both of these functions are extended to accept compile time `useEvaluate`, `sourceUrlPrefix`, and `format`, to exit to CommonJS `require`. We leave open the possibility of an `esm` or `mjs` format that would exit to `import` for host modules.

The new `makeFunctor` is analogous but isn’t suitable for `<script>` tags and instead allows the user to supply runtime options, including `require` for CommonJS format, `evaluate` if compiled with `useEvaluate` to override indirect `eval`, and `sourceUrlPrefix` which also in combination with `useEvaluate` overrides the compile time option by the same name.

Captures `sourceDirname` for each packaged compartment for improved `sourceURL` generation. So, if the directory name does not match the package name, the source URL will be more likely to united with the original sources if they are open in the developer’s IDE.

Moves bundle `use strict` pragma into evaluated function expression bodies for better composition with `eval`.

Improve error message for misconfigured `require` option.

### Security Considerations

No impact.

### Scaling Considerations

No impact.

The new bundling features necessitate new API reference documentation, both those generated from TypeScript, and our hand-crafted README.

### Testing Considerations

This change includes coverage for all essential combinations of the new bundler options, including compile time and runtime options and cases where the latter overrides the former.

Documentation included for was that the new features do not have parity with Rollup. This change is purely additive, but the upcoming changes to `@endo/bundle-source` will require a major version bump for not being comfortably equivalent to former behavior. We have confirmed that the new behavior produces a passing CI run in composition with usage patterns in Agoric SDK, with a couple preparatory changes to make the transition.

### Upgrade Considerations

These changes will impact the formation of the Agoric SwingSet XSnap Supervisor Lockdown (bootstrap) script. They should not produce observable differences in behavior.


packages/compartment-mapper/src/import-hook.js
```diff
@@ -720,6 +712,7 @@ export function makeImportNowHookMaker(
           packageLocation,
           packageSources,
           readPowers,
+          archiveOnly,
```
kriskowal, line 715: I did not need to reveal a new option to `makeBundle` in order to indicate whether a `sourceURL` comment should be generated at the time a module is parsed off disk, because we already had `archiveOnly`, which is not orthogonal. I did get caught surprised that we _do_ need to inject a `sourceURL` at parse time in order to preserve stack traces for `importLocation`, since that doesn’t go through the archiver or bundler.

packages/compartment-mapper/src/parse-cjs.js
```diff
@@ -41,7 +41,7 @@ export const parseCjs = (
    */
   const execute = (moduleEnvironmentRecord, compartment, resolvedImports) => {
     const functor = compartment.evaluate(
-      `(function (require, exports, module, __filename, __dirname) { ${source} //*/\n})\n//# sourceURL=${location}`,
+      `(function (require, exports, module, __filename, __dirname) { 'use strict'; ${source} })\n`,
```
kriskowal, line 44: Moving the pragma into each individual functor and the generated linker was needed to make bundles suitable inlined anywhere an expression is accepted, like `(${source})`.

packages/compartment-mapper/README.md
```diff
@@ -134,6 +134,82 @@ object with an `import({ globals?, modules? })` method.
...
+  This can make the line numbers in stack traces more useful.
+- `curryRuntimeOptions` (`boolean`, default `false`):
+  By default, the completion value of a bundle is the exports namespace of
+  the entry module after all the modules have executed.
+  This is suitable for shims like `ses`, where the completion value is
+  discarded and any runtime options must be communicated through the
+  global environment, like `process.env` or the presence of a `require` function.
+  Currying runtime options changes the completion value to a function that receives
+  runtime options and then returns the exports namespace of the entry module.
+  Runtime options include `evaluate`, `require`, and `sourceUrlPrefix`.
```
gibson042, line 198: Formatting suggestions:
```suggestion
- `sourceUrlPrefix` (string, default `""`):
  Specifies a prefix to occur on each module's `sourceURL` comment, as injected
  at runtime.
  Should generally end with `/` if non-empty.
  This can improve stack traces.
  Can be overridden with the `sourceUrlPrefix` curried runtime option with
  `curryRuntimeOptions`.
- `format` (`"cjs"` or `undefined`, default `undefined`):
  By default, `makeBundle` generates a bundle that can be evaluated in any
  context.
  By specifying `"cjs"`, the bundle can assume there is a host CommonJS
  `require` function available for resolving modules that exit the bundle.
  The default is `require` on `globalThis`.
  The `require` function can be overridden with a curried runtime option.
- `useEvaluate` (boolean, default `false`):
  By default, a bundle has an array of functions for each bundled module.
  With `useEvaluate`, the bundle will instead capture a string for
  each module function and use an evaluator function to rehydrate them.
  The default evaluator is `eval` on `globalThis`, but can be overridden
  with the `evaluate` curried runtime option, with `curryRuntimeOptions`.
  This can make the line numbers in stack traces more useful.
- `curryRuntimeOptions` (boolean, default `false`):
  By default, the completion value of a bundle is the exports namespace of
  the entry module after all the modules have executed.
  This is suitable for shims like `ses`, where the completion value is
  discarded and any runtime options must be communicated through the
  global environment, like `process.env` or the presence of a `require` function.
  Currying runtime options changes the completion value to a function that receives
  runtime options and then returns the exports namespace of the entry module.
  Runtime options include `evaluate`, `require`, and `sourceUrlPrefix`.
```

packages/compartment-mapper/README.md
```diff
@@ -134,6 +134,82 @@ object with an `import({ globals?, modules? })` method.
...
+  with the `evaluate` curried runtime option, with `curryRuntimeOptions`.
+  This can make the line numbers in stack traces more useful.
+- `curryRuntimeOptions` (`boolean`, default `false`):
+  By default, the completion value of a bundle is the exports namespace of
+  the entry module after all the modules have executed.
+  This is suitable for shims like `ses`, where the completion value is
+  discarded and any runtime options must be communicated through the
+  global environment, like `process.env` or the presence of a `require` function.
+  Currying runtime options changes the completion value to a function that receives
+  runtime options and then returns the exports namespace of the entry module.
```
gibson042, line 197:
```suggestion
  Currying runtime options changes the completion value to a function that receives
  a runtime options object and returns the exports namespace of the entry module.
```

gibson042, line 197: I don't think "currying" is technically correct here, nor does it communicate to me what's really going on (the indirection from "bundle represents a [non-function] module namespace" to "bundle represents a function that returns a module namespace"). I don't have a better name suggestion, but I wonder if the indirection might be better represented by a sibling function distinct from `makeBundle` rather than an option for it (and I really an wondering; the answer might be "no").

packages/compartment-mapper/README.md
```diff
@@ -134,6 +134,82 @@ object with an `import({ globals?, modules? })` method.
...
+  Runtime options include `evaluate`, `require`, and `sourceUrlPrefix`.
+
+Curried runtime options:
+
+- `evaluate`: for bundles made with `useEvaluate` or `useNamedEvaluate`,
+  specifies a function to use to evaluate each module.
+  If provided, overrides the evaluator named by `useNamedEvaluate`.
```
gibson042, line 204: These are the only mention of `namedEvaluate`. Are they typos, or indications of missing documentation?

gibson042 (commented):
Initial feedback on the documentation. I have not yet reviewed any code changes.

packages/compartment-mapper/test/bundle.test.js
```diff
@@ -69,6 +71,272 @@ test('bundles work', async t => {
   t.deepEqual(log, expectedLog);
 });
 
+test('eval preserves error line numbers for bundles', async t => {
```
gibson042, line 74: I appreciate the coverage of these new tests.

packages/compartment-mapper/src/types/external.ts
```diff
@@ -131,6 +131,41 @@ export type SyncArchiveLiteOptions = SyncOrAsyncArchiveOptions &
...
+   * a host module.
+   * Specifying `cjs` makes `require` available for modules outside the bundle
+   * (exits to the import graph).
+   */
+  format?: 'cjs';
+  /**
+   * Evaluates individual module functors in-place so stack traces represent
+   * original source locations better.
+   */
+  useEvaluate?: boolean;
```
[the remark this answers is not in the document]
kriskowal: We would not `useEvaluate` for the SES bundle, or any bundle expected to function with a no-unsafe-eval Content-Security-Policy.

packages/compartment-mapper/src/types/external.ts
```diff
@@ -194,6 +244,20 @@ export type SyncModuleTransformsOption = {
...
+   * Whether to eschew preparing modules for execution in favor of defering
+   * linkage to modules that exit the archive to the eventual host environment.
+   * By default, host modules must be available in the current environment.
+   * Also, archives do not capture the sourceURL in applicable compiled sources,
+   * to ensure that archvies and bundles are generated consistently from
+   * the same physical sources, regardless of the project's file system
+   * prefix.
+   * Should be false for runtime and true for making archives and bundles.
+   */
+  archiveOnly?: boolean;
```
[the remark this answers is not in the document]
kriskowal: Rewrote.

packages/compartment-mapper/src/types/external.ts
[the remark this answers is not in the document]
kriskowal: Documented.

packages/compartment-mapper/README.md
kriskowal: Removed.

packages/compartment-mapper/README.md
kriskowal: Per out-of-band conversation, we found that we can avoid the option name by splitting `makeBundle` into `makeScript` and `makeFunctor`, where the latter provides an opportunity to apply runtime options.

kriskowal: Obviated.

packages/compartment-mapper/README.md
```diff
@@ -134,6 +134,82 @@ object with an `import({ globals?, modules? })` method.
...
+  `curryRuntimeOptions`.
+- `format` (`"cjs"` or `undefined`, default `undefined`):
+  By default, `makeBundle` generates a bundle that can be evaluated in any
+  context.
+  By specifying `"cjs"`, the bundle can assume there is a host CommonJS
+  `require` function available for resolving modules that exit the bundle.
+  The default is `require` on `globalThis`.
+  The `require` function can be overridden with a curried runtime option.
+- `useEvaluate` (`boolean`, default `false`):
+  By default, a bundle has an array of functions for each bundled module.
```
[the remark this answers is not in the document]
kriskowal: Rewrote.

packages/compartment-mapper/src/bundle-lite.js
```diff
@@ -0,0 +1,612 @@
...
+ * a function expression followed by a comma delimiter that will be evaluated in
+ * a lexical scope with no free variables except the globals.
+ * In the generated bundle runtime, the function will receive an environment
+ * record: a record mapping every name of the corresponding module's internal
+ * namespace to a "cell" it can use to get, set, or observe the linked
+ * variable.
+ * @property {() => string} getCells Produces a JavaScript string consisting of
+ * a JavaScript object and a trailing comma.
+ * The string is evaluated in a lexical context with a `cell` maker, the `cells`
+ * array of every module's internal environment record.
```
gibson042, line 32:
```suggestion
 * The string is evaluated in a lexical context with a `cell` maker and the `cells`
 * array of every module's internal environment record.
```
?

packages/compartment-mapper/src/bundle-lite.js
```diff
@@ -0,0 +1,612 @@
...
+ * @property {() => string} getFunctorCall Produces a JavaScript string may
+ * be a statement that calls this module's functor with the calling convention
+ * appropriate for its language, injecting whatever cells it needs to link to
+ * other module namespaces.
```
gibson042, line 36: I think this needs a grammar pass.

packages/compartment-mapper/src/bundle-lite.js
```diff
@@ -0,0 +1,612 @@
...
+ * @property {() => string} getReexportsWiring Produces a JavaScript string
+ * that may include statements that bind the cells reexported by this module.
```
gibson042, line 38: These descriptions are good, but I think would further benefit from simple examples and/or a simple explanation demonstrating how the receiver of a BundleKit can tie together output from its methods.

packages/compartment-mapper/src/bundle-lite.js
```diff
@@ -0,0 +1,612 @@
...
+});
+
+/**
+ * @param {Record<string, CompartmentDescriptor>} compartmentDescriptors
+ * @param {Record<string, CompartmentSources>} compartmentSources
+ * @param {string} entryCompartmentName
+ * @param {string} entryModuleSpecifier
+ * @param {Array<string>} exitModuleSpecifiers
+ */
+const sortedModules = (
```
gibson042, line 117: Can you add a description for `sortedModules`? If I'm reading correctly, it's basically a bottom-up topological traversal of the module graph (i.e., a generalization of post-order [tree traversal](https://en.wikipedia.org/wiki/Tree_traversal#Depth-first_search) to handle non-tree graphs) with a side table at each node for alias lookup?

gibson042 (approved):
:heart:
LGTM after minor tweaks.

packages/compartment-mapper/src/bundle-lite.js
```diff
@@ -0,0 +1,612 @@
...
+      },
+      set() {
+        throw new TypeError('Non-writable export');
+      },
+      observe(observer) {
+        observer(Reflect.get(namespace, name));
+      },
+      enumerable: true,
+    }])
+  );
```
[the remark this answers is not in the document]
kriskowal: Given that this is used to generate the SES bundle, yes. I’ll add a commit to shore this up.

packages/compartment-mapper/src/bundle-lite.js
kriskowal: Got it in one.

packages/compartment-mapper/src/bundle-lite.js
kriskowal: Adding JSON example for simplicity.

packages/compartment-mapper/src/bundle-lite.js
kriskowal: Reworking.

packages/compartment-mapper/README.md
```diff
@@ -134,6 +134,170 @@ object with an `import({ globals?, modules? })` method.
...
+  ["(function () { /* ... */ })", "bundle/dependency.js"],
+  // 2. bundle ./dependent.js
+  ["(function () { /* ... */ })", "bundle/dependent.js"],
+  ])(/* runtime options */)
+  ```
+
+  With `useEvaluate`, the bundle will instead capture a string for
+  each module function and use an evaluator function to rehydrate them.
+  This can make the file locations and line numbers in stack traces more
+  useful.
```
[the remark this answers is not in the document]
kriskowal: It can’t. The default is indirect `eval`. Reworking.

PR #2708, opened 2025-02-02
fix(daemon,cli): fix #2700 use endo/init to prepare async hooks

### LINKED ISSUE #2700: In vscode debug session, node async_hooks promise keys appear too late

erights:

Closes: #XXXX
Refs: #XXXX

## Description

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations


packages/cli/src/endo.js
```diff
@@ -3,6 +3,9 @@
 /* eslint-disable no-await-in-loop, no-bitwise, no-throw-literal */
 
 // Establish a perimeter:
+// TODO should this
+//   import '@endo/init';
+// rather than the following three lines?
```
kriskowal, line 8: I think so.

packages/daemon/test/endo.test.js
```diff
@@ -2,6 +2,9 @@
 /* global process */
 
 // Establish a perimeter:
+// TODO should this
+//   import '@endo/init/debug.js';
+// rather than the following three lines?
```
kriskowal, line 7: Also, would not hurt.

kriskowal (approved):
Please resolve TODOs and take out of draft.

erights:
@kriskowal , reassigned to you. Thanks!

Issue #2113, opened 2024-03-03
Must compare strings by codepoint instead of codeunit


erights:
## Describe the bug

Currently, our `compareRank` and `compareKeys` applied to strings compares them in sort order using JavaScript's `<` operator as applied to strings. Unfortunately, JS `<` compares strings according to their lexicographic UTF16 code unit order. (This is preserved on XS by lexicographic comparison of bytes in a CESU-8 encoding of the string.)

However, with our agreement, ocapn has standardized on using the (much more semantically sensible!) Unicode lexicographic codepoint order (which would be preserved by a proper UTF-8 encoding of well formed strings. See https://github.com/endojs/endo/issues/1739 ).

Indeed https://github.com/endojs/endo/pull/2008 "solves" the immediate problem regarding `compareRank` and `compareKeys`. However, there are plenty of places where we still sort strings by their implicit sort order. Worse, we don't know how much data we have already stored on chain organized by the wrong sort order. Nor do we have any practical plans for how to find or fix that data. This needs design.

PR #2710, opened 2025-02-04
feat(patterns): M.containerHas(el,n) to support want patterns

erights:
Closes: #XXXX
Refs: #2002 #2008 #2113 #1739 https://github.com/Agoric/agoric-sdk/pull/10952

## Description

This PR adds a new `M.containerHas(elementPatt, positiveBigint)` matcher, and exported `containerHasSplit` function. This is motivated to support https://github.com/Agoric/agoric-sdk/pull/10952 , which introduces a minimal form of want pattern in terms of `M.containerHas`.

- [x] Actually merging this must happen only after we've decided either to move forward with #2008 or to give up on it. Once a decision is made, and even before it is acted on, then this PR can move forward. (Any decision to move forward or not with #2008 should also consider changing the default of the feature flag introduced by #2002 .)

### Security Considerations

none

### Scaling Considerations

Might help due to early termination of the split operations, which https://github.com/Agoric/agoric-sdk/pull/10952 uses for `AmountMath.isGTE`.

Already doc-documents `M.containerHas` in the types.js file for `M`. That's probably good enough for this PR. The interesting documentation will be explaining want patterns in https://github.com/Agoric/agoric-sdk/pull/10952

### Testing Considerations

Added tests for `M.containerHas`

The reason to postpone merging this PR until decisions are made on #2008 is that this PR will further expose `rankOrder` in the API, amplifying the danger that changing the string order will cause surprising observable changes.

### Upgrade Considerations

This PR itself does not introduce any BREAKING changes or Upgrade issues.

### DIFF
--- a/packages/patterns/src/patterns/patternMatchers.js
+++ b/packages/patterns/src/patterns/patternMatchers.js
@@ -1,5 +1,11 @@
+/* eslint-disable @endo/no-optional-chaining */
 // @ts-nocheck So many errors that the suppressions hamper readability.
 // TODO parameterize MatchHelper which will solve most of them
+import { q, b, X, Fail, makeError, annotateError } from '@endo/errors';
+import { identChecker } from '@endo/common/ident-checker.js';
+import { applyLabelingError } from '@endo/common/apply-labeling-error.js';
+import { fromUniqueEntries } from '@endo/common/from-unique-entries.js';
+import { listDifference } from '@endo/common/list-difference.js';
 import {
   assertChecker,
   Far,
@@ -8,6 +14,8 @@ import {
   passStyleOf,
   hasOwnPropertyOf,
   nameForPassableSymbol,
+} from '@endo/pass-style';
+import {
   compareRank,
   getPassStyleCover,
   intersectRankCovers,
@@ -15,12 +23,7 @@ import {
   recordNames,
   recordValues,
 } from '@endo/marshal';
-import { identChecker } from '@endo/common/ident-checker.js';
-import { applyLabelingError } from '@endo/common/apply-labeling-error.js';
-import { fromUniqueEntries } from '@endo/common/from-unique-entries.js';
-import { listDifference } from '@endo/common/list-difference.js';
 
-import { q, b, X, Fail, makeError, annotateError } from '@endo/errors';
 import { keyEQ, keyGT, keyGTE, keyLT, keyLTE } from '../keys/compareKeys.js';
 import {
   assertKey,
@@ -33,12 +36,14 @@ import {
   checkCopyBag,
   getCopyMapEntryArray,
   makeCopyMap,
+  makeCopySet,
+  makeCopyBag,
 } from '../keys/checkKey.js';
 import { generateCollectionPairEntries } from '../keys/keycollection-operators.js';
 
 /**
- * @import {Checker, CopyRecord, CopyTagged, Passable} from '@endo/pass-style'
- * @import {ArgGuard, AwaitArgGuard, CheckPattern, GetRankCover, InterfaceGuard, MatcherNamespace, MethodGuard, MethodGuardMaker, Pattern, RawGuard, SyncValueGuard, Kind, Limits, AllLimits, Key, DefaultGuardType} from '../types.js'
+ * @import {Checker, CopyArray, CopyRecord, CopyTagged, Passable} from '@endo/pass-style'
+ * @import {CopySet, CopyBag, ArgGuard, AwaitArgGuard, CheckPattern, GetRankCover, InterfaceGuard, MatcherNamespace, MethodGuard, MethodGuardMaker, Pattern, RawGuard, SyncValueGuard, Kind, Limits, AllLimits, Key, DefaultGuardType} from '../types.js'
  * @import {MatchHelper, PatternKit} from './types.js'
  */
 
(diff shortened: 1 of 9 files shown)


packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1258,6 +1259,177 @@ const makePatternKit = () => {
...
+  };
+
+  /**
+   * @param {CopyArray<[Key, bigint]>} pairs
+   * @param {Pattern} elementPatt
+   * @param {bigint} bound Must be >= 1n
+   * @param {CopyArray<[Key, bigint]>} [inResults]
+   * @param {CopyArray<[Key, bigint]>} [outResults]
+   * @returns {boolean}
+   * @param {Checker} [check]
```
Chris-Hibbert, line 1315: nit: swap order so @returns is last.

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1258,6 +1259,177 @@ const makePatternKit = () => {
...
+  ) => {
+    let count = 0n;
+    // Since this feature is motivated by ERTP's use on
+    // semi-fungible (`copyBag`) amounts,
+    // their arrays store their elements in decending lexicographic order.
+    // But this function has to make some choice amoung equally good minimal
+    // results. It is more intuitive for the choice to be the first `bound`
+    // matching elements in ascending lexicigraphic order, rather than
+    // decending. Thus we iterate `pairs` in reverse order.
+    for (let i = pairs.length - 1; i >= 0n; i -= 1) {
```
Chris-Hibbert, line 1333: nit:

```suggestion
    for (let i = pairs.length - 1; i >= 0; i -= 1) {
```

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1258,6 +1259,177 @@ const makePatternKit = () => {
...
+      const [element, num] = pairs[i];
+      const numRest = bound - count;
+      if (numRest >= 1n) {
+        if (matches(element, elementPatt)) {
+          if (num <= numRest) {
+            count += num;
+            inResults?.push([element, num]);
+          } else {
+            const numIn = num - numRest;
+            count += numIn;
```
Chris-Hibbert, line 1343: Shouldn't `inResults` get `numRest`, and outResults get `num - numRest`?

If this isn't clear, I can try to construct a test case. I'd start by trying to remove 2 'b's from a collection that included at least 3.

erights: I'd love a test case, thanks!

Chris-Hibbert: Added commit `abfc1f7` with a test and a proposed fix.

packages/patterns/src/patterns/patternMatchers.js

packages/patterns/src/patterns/patternMatchers.js

packages/patterns/src/patterns/patternMatchers.js
erights: Accidentally wiped that out, but it is now restored. Done by you, Thanks!

Chris-Hibbert (approved):
I agree with @gibson042's suggestions on naming. Additionally, I wouldn't expect there to be many calls, so a long name is fine here.

erights:
I went with `containerHas`, so first bullet done.

erights:
As for the second, my whole approach with the pair of PRs is to get unstuck by adding minimal complexity needed to get enough expressiveness to cover our expected actual needs. We can always expand the patterns we accept in these roles later if
- we find we need it
- the need is worth paying the extra complexity cost.

erights:
I am ready to squash and merge. But in the OP (and at Show&Tell) I said

> Actually merging this must happen only after we've decided either to move forward with https://github.com/endojs/endo/pull/2008 or to give up on it. Once a decision is made, and even before it is acted on, then this PR can move forward. (Any decision to move forward or not with https://github.com/endojs/endo/pull/2008 should also consider changing the default of the feature flag introduced by https://github.com/endojs/endo/pull/2002 .)

I'd like to appeal this to my reviewers. The existence of this new feature will probably not cause the problem yet. The use of it by agoric-sdk will cause the problem. So I'd like to proceed so it gets in the next endo release, enabling a smoother experience of local testing and debugging of https://github.com/Agoric/agoric-sdk/pull/10952

***Reviewers, waddaya think? I'll wait a day for at least one yes and no objections.***

gibson042:
> We can always expand the patterns we accept in these roles later

AFAICT, such an expansion would be awkward unless it takes the form of an entirely new name, because e.g. `M.containerHas([["foo"], ["bar"]])` is already valid in this PR ("_has an element `[["foo"], ["bar"]]`_") and therefore could not be reinterpreted to mean "_has an element 'foo' and an element 'bar'_". So if we anticipate potential introduction of the latter behavior, we should not burn its likely name on the former.

erights:
I was not thinking of either a new name nor a new pattern. Rather, using `M.containerHas` as a part of a larger pattern. This would require `AmountMath` to cope with more complicated patterns, though still hopefully a narrow subset of all patterns. 

That's what I meant by "We can always expand the patterns we accept in these roles later". The "we accept" is more precisely "AmountMath accepts". `M.containerHas` already accepts any pattern as the elementPattern. I don't think the expressiveness of `M` needs to expand at all.

turadg:
>smoother experience of local testing and debugging of https://github.com/Agoric/agoric-sdk/pull/10952

Given that locally one can have second checkouts of the two repos and `yarn link` them, I don't think the motivation given warrants expediting this into Endo master.

(And for remote testing https://github.com/Agoric/agoric-sdk/pull/10952 is already using `#endo-branch`)

erights:
Thanks @turadg , I will stick with the original commitment. Since it is otherwise ready to go, I'm putting it in draft to protect me from myself ;)

erights:
We have now decided to move forward with #2008 (thanks @mhofman @gibson042 ), so this PR can now be merged.

erights:
> Meta-commentary:
> 
> * I think `M.has` is too general a name for the narrow behavior that this actually consists of. Some alternate possibilities:
>   
>   * `M.{has,contains,includes}Element`
>   * `M.{container,collection,bag}{Has,Contains,Includes}`
>   * `M.{container,collection,bag}Satisfies`

@kriskowal suggests `contains` which I like too. But it doesn't fit with @gibson042 's suggested naming choices above. (FWIW, I still like `has` best.) @kriskowal also says he's willing to hold this upcoming release for this renaming.

So how do y'all @gibson042 @kriskowal @turadg @Chris-Hibbert @mhofman feel about `contains` vs `containerHas`? This is probably the last chance we have to painlessly rename this.

gibson042:
I still think the name needs either a prefix or a suffix indicating its restriction to collection membership testing. Bare `has` seems like it should encompass CopyRecord property presence à la JavaScript `Object.has`, and bare `contains` seems like it should encompass substring testing.

erights:
Thanks @gibson042 .

Ok @kriskowal , we're going with `containerHas`. No need to revise the contributions from this PR.

kriskowal:
> I still think the name needs either a prefix or a suffix indicating its restriction to collection membership testing. Bare `has` seems like it should encompass CopyRecord property presence à la JavaScript `Object.has`, and bare `contains` seems like it should encompass substring testing.

I agree that `has` is not a viable name for on the precedent of `Object.has`. I also agree that `contains` at least implies a multi-element argument if not specifically a `string` on the precedent of `String.prototype.contains`. I think `includes` is viable on the precedent of `Array.prototype.includes`, which has a single-element RHS and I believe is the analogous set-theory verb.

At this point, I’m not holding up the release on continued discussion, but do let me know if this is a compelling argument.

gibson042:
I think I could live with a single-element `includes`, especially if we had something pre-reserved for a multi-element generalization (should we decide to want that in the future). Would you have a proposal for that?

kriskowal:
> I think I could live with a single-element `includes`, especially if we had something pre-reserved for a multi-element generalization (should we decide to want that in the future). Would you have a proposal for that?

For multi-element, I would favor `contains`, on the precedent of String.

erights:
The pattern we're talking about has a single subpattern (the elementPattern) and a positive nat count, but most certainly is not about a single member of the collection. In fact, if the count is >= 2 then is cannot be satisfied by a single member of the collection. That's why the earlier suggestion `M.{has,contains,includes}Element` made no sense to me. Whatever it is, it is plural in general, not singular.

Is it unclear what the pattern means?

erights:
I mention all this because `includes` makes the same mistake as `M.{has,contains,includes}Element`

erights:
In any case, I agree this is not about to converge, so we should go ahead with `containerHas`. We will regret it, but we can stand it.

gibson042:
My above use of "single-element" and "multi-element" should instead be read as "single-pattern" and "multi-pattern". As noted in [the original comment](https://github.com/endojs/endo/pull/2710#issuecomment-2731175970), I do anticipate eventual introduction of multi-pattern composites to express more intricate wants.

erights:
Is there a string contains or array contains? I could not find them. Where?

gibson042:
ECMAScript String and Array/%TypedArray% `includes` were introduced as fallback choices for what was [originally `contains`](https://web.archive.org/web/20141210155030/https://people.mozilla.org/~jorendorff/es6-draft.html#sec-string.prototype.contains), the [renaming](https://web.archive.org/web/20141212232513/https://people.mozilla.org/~jorendorff/es6-draft.html#sec-string.prototype.includes) (_and deviation from [Python](https://docs.python.org/3/reference/datamodel.html#object.__contains__) et al._) being forced by MooTools as recorded in November 2014 ([topic](https://github.com/tc39/notes/blob/main/meetings/2014-11/nov-18.md#44-arrayprototypecontains-breaks-mootools) and [conclusion](https://github.com/tc39/notes/blob/main/meetings/2014-11/nov-18.md#51--44-arrayprototypecontains-and-stringprototypecontains)). `String.prototype.includes` shipped in [2015 ES6](https://262.ecma-international.org/6.0/index.html#sec-string.prototype.includes) and `{Array,%TypedArray%}.prototype.includes` in [2016 ES7](https://262.ecma-international.org/7.0/index.html#sec-array.prototype.includes).

Map/Set/WeakMap/WeakSet `has(key)` was an initial method on the respective prototypes in 2015 ES6 (_followed by `Object.hasOwn` as a static analogue of `Object.prototype.hasOwnProperty` in https://github.com/tc39/ecma262/commit/077317b1529602407f25eab3e0e96fd5ae09c32d for the 2022 edition_). And Set-only `isSubsetOf`/`isSupersetOf` were added (_alongside other methods concerned with multiple elements_) by [Set Methods for JavaScript](https://github.com/tc39/proposal-set-methods) just last year.

----

But in Endo, a pattern function doesn't have the context of a prototype and so ought to indicate by name the domain of inputs for which it is applicable and corresponding predicate behavior over them—and they generally succeed at that! (_`splitArray` and `splitRecord` being the outliers as indicated by our own internal use._) This one is tricky because a) it generalizes across CopyArray/CopySet/CopyBag but not CopyRecord or CopyMap, b) it is limited to a single pattern in a way that a plausible successor won't be, and c) due to the wide open domain of its first parameter, future introduction of a successor would almost certainly take the form of a new method rather than domain expansion. And in the world anticipated by point c, the name for the single-pattern method should not suggest **behavior** of the multi-pattern method, which is my biggest objection to an unqualified name here... even `M.containerHas` seems like it might be the method to reach for if one has a subset or subbag constraint and knows that such a method exists, so the latter would need a name that does an even better job of suggesting its general multi-pattern nature (which is made difficult by point a already generalizing from sets to bags, stretching otherwise obvious use of "superset" for such a purpose).

PR #2716, opened 2025-02-11
refactor(pass-style): faster `passStyleOf`

erights:
Closes: #XXXX
Refs: https://github.com/Agoric/agoric-sdk/pull/10975 https://github.com/Agoric/agoric-sdk/pull/10982 

## Description

Using the benchmark tool from #10975 by @muhammadahmadasifbhatti as modified by #10982 to produce flamegraphs (interactive in vscode) I iterated on that benchmark test which already had very specific simple examples exercising  `passStyleOf`. Together with the snippet at https://github.com/Agoric/agoric-sdk/blob/acbb5da3c5a52bab8db319fd696aed70146f9a89/.github/actions/restore-node/action.yml#L156 (which @gibson042 brought to my attention)

```sh
$ scripts/get-packed-versions.sh ~/endo | scripts/resolve-versions.sh
$ yarn install && yarn build
$ cd packages/benchmark-test
$ yarn bench
```
and then clicking on the latest *.cpuprofile file, assuming you've already installed the https://marketplace.visualstudio.com/items?itemName=ms-vscode.vscode-js-profile-flame vscode extension.

I was able to iterate and tinker on possible improvements, to see what made how much of a difference. This PR has the improvements to `passStyleOf` that I and my reviewers have came up with so far.

@gibson042 suggested the main technique used: to avoid redundant checking by breaking up `assertValid` so we could still running the checks of `canBeValid` twice.

### Security Considerations

I claim that this PR is a pure refactoring. If true, none. ***Reviewers***, please review with a skeptical eye. `passStyleOf` is security critical, so any observable difference might lead to an opportunity for an adversary.

### Scaling Considerations

Even though I iterated on a specialized benchmark exercising `passStyleOf`, I believe I only made changes that would also be an improvement for the typical cases.

If this is a pure refactor, none.

### Testing Considerations

If this is a pure refactor, none. In fact, this PR did not need to change any tests, providing some weak evidence that it is indeed a pure refactor.

If this is a pure refactor, none.

### Upgrade Considerations

If this is a pure refactor, none.

### DIFF
--- a/packages/pass-style/src/copyArray.js
+++ b/packages/pass-style/src/copyArray.js
@@ -8,25 +8,17 @@ const { ownKeys } = Reflect;
 const { isArray, prototype: arrayPrototype } = Array;
 
 /**
- * @param {unknown} candidate
- * @param {import('./types.js').Checker} [check]
- * @returns {boolean}
- */
-const canBeValid = (candidate, check = undefined) =>
-  isArray(candidate) ||
-  (!!check && check(false, X`Array expected: ${candidate}`));
-
-/**
  *
  * @type {import('./internal-types.js').PassStyleHelper}
  */
 export const CopyArrayHelper = harden({
   styleName: 'copyArray',
 
-  canBeValid,
+  canBeValid: (candidate, check = undefined) =>
+    isArray(candidate) ||
+    (!!check && check(false, X`Array expected: ${candidate}`)),
 
-  assertValid: (candidate, passStyleOfRecur) => {
-    canBeValid(candidate, assertChecker);
+  assertRestValid: (candidate, passStyleOfRecur) => {
     getPrototypeOf(candidate) === arrayPrototype ||
       assert.fail(X`Malformed array: ${candidate}`, TypeError);
     // Since we're already ensured candidate is an array, it should not be
--- a/packages/pass-style/src/copyRecord.js
+++ b/packages/pass-style/src/copyRecord.js
@@ -64,9 +64,7 @@ export const CopyRecordHelper = harden({
     );
   },
 
-  assertValid: (candidate, passStyleOfRecur) => {
-    checkObjectPrototype(candidate, assertChecker);
-
+  assertRestValid: (candidate, passStyleOfRecur) => {
     // Validate that each own property is appropriate, data/enumerable,
     // and has a recursively passable associated value.
     for (const name of ownKeys(candidate)) {
@@ -76,7 +74,6 @@ export const CopyRecordHelper = harden({
         true,
         assertChecker,
       );
-      checkPropertyCanBeValid(candidate, name, value, assertChecker);
       passStyleOfRecur(value);
     }
   },
--- a/packages/pass-style/src/copyRecord.js
+++ b/packages/pass-style/src/copyRecord.js
@@ -65,8 +65,9 @@ export const CopyRecordHelper = harden({
   },
 
   assertRestValid: (candidate, passStyleOfRecur) => {
-    // Validate that each own property is appropriate, data/enumerable,
-    // and has a recursively passable associated value.
+    // Validate that each own property has a recursively passable associated
+    // value (we already know from canBeValid that the other constraints are
+    // satisfied).
     for (const name of ownKeys(candidate)) {
       const { value } = getOwnDataDescriptor(
         candidate,
--- a/packages/pass-style/src/passStyleOf.js
+++ b/packages/pass-style/src/passStyleOf.js
@@ -4,7 +4,12 @@
 
 import { isPromise } from '@endo/promise-kit';
 import { X, Fail, q, annotateError, makeError } from '@endo/errors';
-import { isObject, isTypedArray, PASS_STYLE } from './passStyle-helpers.js';
+import {
+  isObject,
+  isTypedArray,
+  PASS_STYLE,
+  assertChecker,
+} from './passStyle-helpers.js';
 
 import { CopyArrayHelper } from './copyArray.js';
 import { CopyRecordHelper } from './copyRecord.js';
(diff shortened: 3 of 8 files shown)


packages/ses/src/make-hardener.js
```diff
@@ -73,6 +75,10 @@ assert(getTypedArrayToStringTag);
  * @param {unknown} object
  */
 export const isTypedArray = object => {
+  if (!(object instanceof TypedArray)) {
+    // quicker reject for typical case
+    return false;
+  }
```
gibson042, line 81: This gives false negatives for inputs like `Object.setPrototypeOf(Uint8Array.from([]), null)`, which is acceptable for some applications of the exported function but possibly not for others. I don't think this change should be included.

packages/pass-style/src/copyRecord.js
```diff
@@ -76,7 +74,6 @@ export const CopyRecordHelper = harden({
         true,
         assertChecker,
       );
-      checkPropertyCanBeValid(candidate, name, value, assertChecker);
```
gibson042, line 79: :+1:

packages/pass-style/src/copyArray.js
```diff
@@ -7,26 +7,18 @@ const { getPrototypeOf } = Object;
...
   styleName: 'copyArray',
 
-  canBeValid,
+  canBeValid: (candidate, check = undefined) =>
+    isArray(candidate) ||
+    (!!check && check(false, X`Array expected: ${candidate}`)),
 
-  assertValid: (candidate, passStyleOfRecur) => {
-    canBeValid(candidate, assertChecker);
+  assertRestValid: (candidate, passStyleOfRecur) => {
```
gibson042, line 21: :+1:

packages/pass-style/src/copyRecord.js
```diff
@@ -64,9 +64,7 @@ export const CopyRecordHelper = harden({
     );
   },
 
-  assertValid: (candidate, passStyleOfRecur) => {
-    checkObjectPrototype(candidate, assertChecker);
-
+  assertRestValid: (candidate, passStyleOfRecur) => {
     // Validate that each own property is appropriate, data/enumerable,
     // and has a recursively passable associated value.
```
gibson042, line 69:
```suggestion
    // Validate that each own property has a recursively passable associated
    // value (we already know from canBeValid that the other constraints are
    // satisfied).
```

packages/pass-style/src/passStyleOf.js
```diff
@@ -63,6 +68,21 @@ const makeHelperTable = passStyleHelpers => {
...
+ * which checks both conditions in the right order.
+ *
+ * @param {PassStyleHelper} helper
+ * @param {any} candidate
+ * @param {(val: any) => PassStyle} passStyleOfRecur
+ * @returns {void}
+ */
+const assertValid = (helper, candidate, passStyleOfRecur) => {
+  helper.canBeValid(candidate, assertChecker);
+  helper.assertRestValid(candidate, passStyleOfRecur);
```
gibson042, line 83: I really like the switch to this pattern, but I wonder if we can come up with a better name than `assertRestValid`. Maybe `assertActuallyValid`? or `assertCompleteValidity`?

packages/ses/src/make-hardener.js
erights: Well, damn. I agree with your diagnosis. But this early reject turned out to make a *huge* difference. How else could we get the benefit? I suppose we could put a try/catch around the freeze, and fall back to this only in the error case. Waddaya think?

packages/pass-style/src/passStyleOf.js
erights: I am open to a new term. But the overriding issue is a name that indicates the it isn't providing assurance by itself, but only if it follows another check. Both `assertActuallyValid` and `assertCompleteValidity` are likely to be misread this way. The second misreading is amusing: "complete" as a verb or an adjective?

packages/pass-style/src/copyRecord.js
erights: Done

packages/ses/src/make-hardener.js
erights: I did the try technique instead. PTAL. 

I am about to benchmark it so I can say how the three cases compare!

erights: Current master before this PR

image
image

With this PR, but before I took @gibson042 's suggestion of removing the `instanceof`
image
image

With this PR after taking @gibson042 's suggestion, and using the try/catch trick instead
image
image

erights: To my great surprise, I conclude from this that both `harden` fixes were worse, which makes no sense to me. But that's why we measure. I will back out the harden fixes.

`passStyleOf` fix looks great! The new one takes about 76% of the original. I'll measure again after reverting the harden changes.

erights: After reverting the `harden` change

image image

erights: Scratch the 76%. The `passStyleOf` results are still significant enough to be worth it.

Altogether, I'm surprised at how much random variance I get. I would have thought 10,000 iterations was enough.

erights: Since I fully reversed the `harden` change, resolving.

erights:
Ready for review!

packages/pass-style/src/passStyleOf.js
gibson042: That's true. Most accurate would be something like `assertConditionallyValid`, which is unfortunately a mouthful. We're really doing material condition in an unusual decomposition, where `assertRestValid` is "A ⇒ B" and `canBeValid` is "A", such that the combination in temporal order is "A ∧ (A ⇒ B)" = "B".

gibson042 (approved):

packages/pass-style/src/passStyleOf.js
erights: Good. Resolving for now. But unresolve if you have a new idea. Thx.

PR #2720, opened 2025-02-13
feat(bench-test): Benchmark Testing

muhammadahmadasifbhatti:
Note: This PR is migrated from the [`agoric-sdk`](https://github.com/Agoric/agoric-sdk/pull/10975). All the comments from [Mark's reviewed PR](https://github.com/Agoric/agoric-sdk/pull/10982) are addressed. 

### Description
This is PR is just a minimalistic POC to create a small tool for the benchmark testing. 

### How to run locally
Run the command `yarn test` in the `packages/benchmark-test` folder.

### Pre-reqs of running `yarn test`
Run the command `yarn install-engines` to ensure installation of V8 and XS engines in `$HOME/.esvu` for this package's `eshost`.


muhammadahmadasifbhatti:
Currently, I am working the following changes but planing them for the follow-up PRs.
* `yarn test` should run all the files in the `/test` folder. 
* `yarn test filename` should run the test from that specific file.
* See any system level libraries that provide time in ~ns.

packages/benchmark/src/benchmark.js
```diff
@@ -0,0 +1,38 @@
...
+
+  console.log(`${name} | Average time: ${avgTime}ms`);
+  t.assert(
+    avgTime < expedtedTime,
+    `Expected ${avgTime} to be less than ${expedtedTime}`,
+  );
+}
+
+async function test(name, fn) {
+  await null;
```
Muneeb147, line 20: Why this await null?

packages/benchmark/README.md
```diff
@@ -0,0 +1 @@
+
```
Muneeb147, line 1: Please add Readme for this package. It can be the details we've added in the description of PR

muhammadahmadasifbhatti: Sure.

packages/benchmark/src/benchmark.js
erights: This is to enforce that the "synchronous prelude" (a neat concept discovered by @gibson042 ) is statically, not dynamically, distinct from the rest of the code of an asynchronous function. The reason is that the synchronous prelude executes on top of its caller's call stack, so when it hits this first `await`, it synchronously returns a promise for its eventual outcome to the caller. Everything after the synchronous prelude happens in later turns with an empty call stack above it, i.e., at the bottom of the stack.

Without this rule, whether a particular bit of code in the asynchronous function executes with or without a call stack underneath it would be data dependent, creating a debugging nightmare. If the code is tested under only one of these scenarios, it may fail under the other for reasons that the test author did not anticipate.

The irony for the code in question here is that this rationale does not apply to the test framework code itself, only to the production code being tested. Nevertheless, unless there is a reason not to, it is better to stick with this convention than not. Every line of code we have is implicitly an expository example for the next line of code written by someone else. Best to practice everywhere the habits we want everyone to learn.

@gibson042 , do you know where we've documented the rational for the synchronous prelude and the resulting `await` safety rule?

packages/benchmark/src/benchmark.js
```diff
@@ -0,0 +1,38 @@
+const performance = { now: () => Date.now() };
```
erights, line 1: I though `performance.now()` was microseconds. Sorry for the confusion. Now that I understand it is nanoseconds, my corresponding suggestion would be

```suggestion
const performance = { now: () => Date.now() * 1_000_000 };
```

to normalize to nanoseconds. But from your comments, it sounds like you agree we should normalize to nanoseconds, and that you're doing this somewhere. Where?

packages/benchmark/src/benchmark.js
```diff
@@ -0,0 +1,38 @@
...
+    console.log('Running test: ', name);
+    await fn({ assert, truthy });
+    console.log(`✅ Passed`);
+  } catch (err) {
+    console.log(`❌ Failed: ${err.message}`);
+  }
+}
+
+function assert(condition, message = 'Assertion failed') {
+  if (!condition) throw new Error(message);
```
erights, line 31: minor
```suggestion
  if (!condition) throw Error(message);
```

packages/benchmark/test/index.test.js
```diff
@@ -0,0 +1,58 @@
+import '@endo/init';
+import { passStyleOf } from '@endo/pass-style';
+import { test, benchmark } from '../src/benchmark.js';
```
erights, line 3: Yeah, `src/` is perhaps a better choice than `tools/`. No change suggested.

packages/benchmark/package.json
```diff
@@ -0,0 +1,73 @@
...
+  "scripts": {
+    "build": "exit 0",
+    "lint": "yarn lint:types && yarn lint:eslint",
+    "lint-check": "yarn lint",
+    "lint-fix": "yarn lint:eslint --fix && yarn lint:types",
+    "lint:eslint": "eslint '**/*.js'",
+    "lint:types": "tsc",
+    "postpack": "git clean -f '*.d.ts*'",
+    "prepack": "tsc --build tsconfig.build.json",
+    "test": "yarn rollup -c && eshost -h v8,xs dist/bundle.js",
```
erights, line 34: Good. I think I understand now why `v8,xs` is a better choice, and how your README instructions make this work. (Though I haven't tried it myself yet.) No change suggested.

packages/benchmark/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "bench": "yarn rollup -c && node --jitless --cpu-prof dist/bundle.js"
```
erights, line 35: I am on the fence about whether my suggested name here should be changed to emphasize this is a node-only test whose purpose is to generate a `.cpuprofile` to get a flamegraph. Or whether we should leave it simple in expectation of generalizing the script to be cross platform using eshost. No change suggested.

packages/benchmark/package.json
```diff
@@ -0,0 +1,73 @@
...
+  },
+  "devDependencies": {
+    "@endo/lockdown": "workspace:^",
+    "@endo/ses-ava": "workspace:^",
```
erights, line 39: Are you using ses-ava?

erights (approved):
LGTM, thanks!!!!

packages/benchmark/src/benchmark.js
muhammadahmadasifbhatti: Thanks for reminding.

muhammadahmadasifbhatti:
@erights I have addressed the comments. I think now I can tag @kriskowal to have a final look. 
Nevertheless, I am done with another small feature that when you use this tool from CLI it can take filename as the input. That PR will be ready soon and you will be able to review that on Monday. 

I also have Richard's idea in mind to have an executable like `ava` but before that I'll write a `js` script that'll have an entry point in the `package.json`.

erights:
Hi @muhammadahmadasifbhatti , I resolved all the conversations above that are done, or that were "No change suggested" that you thumbs-uped. That turned out to be all of them ;). I also approved the workflows and see we still have CI errors. The lint error looks like the one that would be fixed by doing a `yarn format` in the top level `endo` directory.

muhammadahmadasifbhatti:
> Hi @muhammadahmadasifbhatti , I resolved all the conversations above that are done, or that were "No change suggested" that you thumbs-uped. That turned out to be all of them ;). I also approved the workflows and see we still have CI errors. The lint error looks like the one that would be fixed by doing a `yarn format` in the top level `endo` directory.

Thanks!
I just ran the `yarn format` command and pushed the changes. Hopefully, CI passes this time.

erights:
In the last CI run, it looks like you were also getting

> command not found: eshost

errors. I'm not sure what you're supposed to do so this is available in CI. @gibson042 do you know? Since (I think) we're already using eshost for other tests, why wasn't it already available?

@kriskowal , what permission is @muhammadahmadasifbhatti missing that needed me to approve the workflow for CI to run? Whatever that is, could you add it? Thanks.

muhammadahmadasifbhatti:
> In the last CI run, it looks like you were also getting
> 
> > command not found: eshost
> 
> errors. I'm not sure what you're supposed to do so this is available in CI. @gibson042 do you know? Since (I think) we're already using eshost for other tests, why wasn't it already available?
> 
> @kriskowal , what permission is @muhammadahmadasifbhatti missing that needed me to approve the workflow for CI to run? Whatever that is, could you add it? Thanks.

For the above mentioned error, I think we might need to update the `yaml` file of the `workflow` to install `eshost` globally if that is not case already. 

I am also getting another error which seems to be a `linting` error. 
```
This expression is not callable.
  Type 'typeof import("/Users/muhammadahmad/Desktop/agoric-repos/endo/packages/benchmark/node_modules/@rollup/plugin-commonjs/types/index")' has no call signatures
 ```
VSCode is also squiggly here but the code is executing perfectly in both the engines. I will post the update in case I find anything.

File: packages/benchmark/src/benchmark.js [hunk 9]
```diff
@@ -0,0 +1,38 @@
+const performance = { now: () => Date.now() * 1_000_000 };
+
+async function benchmark(name, t, fn, expedtedTime, iterations = 10000) {
+  await null;
+  const start = performance.now();
+  for (let i = 0; i < iterations; i += 1) {
+    await fn();
```
erights, line 7: @michaelfig @gibson042 , do you agree that the lint complaint about the `await` on this line 7 is incorrect and a bug in our lint rule, since the earlier top-level `await null;` on line 4 should have closed the synchronous prelude?

erights: Huh. Or maybe the "no await inside a loop" is a pre-existing lint setting separate from any of our `await` safety rules? In which case we should turn off this pre-existing lint rule globally, i.e., everywhere our own `await` safety rules are enforced? Attn @kriskowal @mhofman @turadg

turadg: https://eslint.org/docs/latest/rules/no-await-in-loop is built into Eslint. It's not for safety so I don't think the Jessie rule affects whether to enable it.

btw, I don't see any linting output in CI because of [this error](https://github.com/endojs/endo/actions/runs/13358283307/job/37304093134):
```
Run yarn lint
Checking formatting...
All matched files use Prettier code style!
Error: rollup.config.js(12,13): error TS23[4](https://github.com/endojs/endo/actions/runs/13358283307/job/37304093134#step:7:5)9: This expression is not callable.
  Type 'typeof import("/home/runner/work/endo/endo/packages/benchmark/node_modules/@rollup/plugin-node-resolve/types/index")' has no call signatures.
```

muhammadahmadasifbhatti: I am of this opinion that we should disable `no-await-in-loop` rule for this line. Reason being, we actually need to check the performance of the function execution separately.

@turadg the issue you mentioned in the code snippet above is being resolved but the workflow has not been ran after the update.

michaelfig: @turadg writes:
> It's not for safety so I don't think the Jessie rule affects whether to enable it.

Correct.  I consider it a reminder for folks to use `for-await-of` instead if that makes the code more straightforward. 

@muhammadahmadasifbhatti writes:
> I am of this opinion that we should disable `no-await-in-loop` rule for this line.

I agree with that opinion.

muhammadahmadasifbhatti:
`test` job is failing and the reason is: `eshost` not found. I have some solutions in mind

* Update the `.github/workflows/ci.yml` and install `eshost` globally.
* Install `eshost` in the `devDependencies` and then use `yarn eshost`. 
* Just use `yarn dlx` as it will temporarily installs and runs a package.

Calling out the expert opinion. I would go for the 1st one as it is the simplest solution.

turadg:
> we might be needing `eshost` in other `yarn scripts`

True, but that's not a good reason to install it globally. The downsides of a global install is isn't automatically in all environments and it may be the wrong version. This is why we have package managers 😀 

But I suggest a fourth option:
- 'devDependencies' so it installs a particular version AND "npm exec eshost" so it runs independently of the package manager

muhammadahmadasifbhatti:
Fixed the CI. 
Another quick follow-up PR is coming where I add the `--verbose` flag. Right now, the output is verbose by default and there is a bunch of info(output) that a developer does not want to see. By adding the flag, I will make it optional.

Muneeb147 (approved):

packages/benchmark/src/benchmark.js
gibson042: I guess https://github.com/endojs/Jessie/blob/main/packages/eslint-plugin/lib/rules/safe-await-separator.js , although the comments at https://github.com/Agoric/test-agoric-sdk/blob/d76e2ca5d6c4286eea046abefd816fcaf067b41f/.eslintrc.cjs#L106-L119 are more informative.

gibson042:
FWIW, I consider dependency upon `eshost` to be inappropriate here... it makes sense for `esbench`, but what _this_ application really needs is an entry point configured with a collection of generic commands which it can invoke with a path to an input file—and for Agoric's purposes, the command for XS should wrap xsnap rather than xst and the command for V8 should just be `node` (maybe with options for self-identification as a benchmark subprocess).

But we're still iterating, and if it is delivering utility with `eshost` then full speed ahead and don't sweat the details.

packages/benchmark/src/benchmark.js
```diff
@@ -0,0 +1,39 @@
+const performance = { now: () => Date.now() * 1_000_000 };
```
gibson042, line 1: This is not representative of [`performance.now`](https://nodejs.org/api/perf_hooks.html#performancenow), which returns _milliseconds_.
```suggestion
const performance = { now: () => Date.now() };
```

muhammadahmadasifbhatti:
> FWIW, I consider dependency upon `eshost` to be inappropriate here... it makes sense for `esbench`, but what _this_ application really needs is an entry point configured with a collection of generic commands which it can invoke with a path to an input file—and for Agoric's purposes, the command for XS should wrap xsnap rather than xst and the command for V8 should just be `node` (maybe with options for self-identification as a benchmark subprocess).
> 
> But we're still iterating, and if it is delivering utility with `eshost` then full speed ahead and don't sweat the details.

@gibson042 thanks for the input. I've noted your opinion and added it in the follow-up list. 
As you mentioned, and also, what I have in my mind; we are inclined towards `ship and iterate`. Current focus is to provide the bare minimum usability. The changes you suggested might be in the next PR.

packages/benchmark/src/benchmark.js
muhammadahmadasifbhatti: Umm you are right. The name and function are conflicting. 
We just want to measure the time in nano-seconds. So, if anything's gonna change it's the name of the function.

muhammadahmadasifbhatti: Changed the name.

muhammadahmadasifbhatti:
@gibson042 just leaving a reminder to have a final look at the PR.

muhammadahmadasifbhatti:
I have created all the follow-up tickets on the SRE board.[ Here is the link to them](https://github.com/orgs/Agoric/projects/19/views/99?filterQuery=label%3ASRE+-status%3A%22Code+Complete%22+assignee%3Amuhammadahmadasifbhatti+sprint%3A%22Feb+7%22&visibleFields=%5B%22Title%22%2C%22Assignees%22%2C%22Status%22%2C132120334%2C53042410%2C52629898%2C127697407%2C58903131%2C%22Repository%22%2C%22Labels%22%5D).

@gibson042 please feel free to leave more suggestions and feedbacks. I would highly appreciate if you can review it today.

packages/benchmark/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "build": "exit 0",
+    "lint": "yarn lint:types && yarn lint:eslint",
+    "lint-check": "yarn lint",
+    "lint-fix": "yarn lint:eslint --fix && yarn lint:types",
+    "lint:eslint": "eslint '**/*.js'",
+    "lint:types": "tsc",
+    "postpack": "git clean -f '*.d.ts*'",
+    "prepack": "tsc --build tsconfig.build.json",
+    "test": "./run-tests.sh",
+    "bench": "yarn rollup -c && node --jitless --cpu-prof dist/bundle.js"
```
gibson042, line 35: I find it very strange for `yarn bench` to run `node --cpu-prof` against the hard-coded test/index.test.js file.
```suggestion
    "cpu-prof-test": "yarn rollup -c && node --jitless --cpu-prof dist/bundle.js"
```

packages/benchmark/run-tests.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/sh
+
+set -e 
+
+echo $(yarn --version)
```
gibson042, line 5:
```suggestion
echo "yarn version: $(yarn --version)"
```

packages/benchmark/run-tests.sh
```diff
@@ -0,0 +1,48 @@
...
+are_engines_installed() {
+    [ -d "$HOME/.esvu/engines/xs" ] && [ -d "$HOME/.esvu/engines/v8" ]
+}
+
+if are_engines_installed; then
+    echo "Engines already installed. Skipping installation."
+else
+    echo "Installing engines..."
+    INSTALL_OUTPU_XS=$(yarn dlx esvu install xs 2>&1) || INSTALL_STATUS_XS=$?
+    INSTALL_OUTPU_V8=$(yarn dlx esvu install v8 2>&1) || INSTALL_STATUS_V8=$?
```
gibson042, line 16: I'm not comfortable with `yarn test` modifying $HOME.
```suggestion
    echo "xs and/or v8 not found in $HOME/.esvu/engines; please esvu install them."
    # The bash "command not found" exit status code is 127.
    # https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html#Exit-Status-1
    exit 127
```

packages/benchmark/run-tests.sh
```diff
@@ -0,0 +1,48 @@
...
+if [ -n "$INSTALL_STATUS_XS" ] || [ -n "$INSTALL_STATUS_V8" ]; then 
+    if are_engines_installed; then
+        echo "Engines installed successfully despite esvu error."
+    else
+        echo "Error installing XS or V8:"
+        echo "$INSTALL_OUTPU_XS"
+        echo "$INSTALL_OUTPU_V8"
+        exit 1
+    fi
+fi
```
gibson042, line 28: I'm not comfortable with `yarn test` modifying $HOME.
```suggestion
if are_engines_installed; then
    echo "Engines already installed. Skipping installation."
else
    echo "xs and/or v8 not found in $HOME/.esvu/engines; please esvu install them."
    # The bash "command not found" exit status code is 127.
    # https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html#Exit-Status-1
    exit 127
fi
```

packages/benchmark/run-tests.sh
```diff
@@ -0,0 +1,48 @@
...
+
+
+ls -la "$HOME/.esvu/engines"
+ls -la "$HOME/.esvu/engines/xs"
+ls -la "$HOME/.esvu/engines/v8"
+
+
+chmod +x "$HOME/.esvu/engines/xs/xst"
+chmod +x "$HOME/.esvu/engines/v8/d8"
+
```
gibson042, line 38: What purpose is served by these commands?
```suggestion
```

packages/benchmark/run-tests.sh
```diff
@@ -0,0 +1,48 @@
...
+yarn eshost --add "xs" xs "$HOME/.esvu/engines/xs/xst"
+yarn eshost --add "v8" d8 "$HOME/.esvu/engines/v8/d8"
+
+yarn eshost --list
```
gibson042, line 42:
```suggestion
yarn eshost -h xs,v8 -se true >/dev/null 2>&1 || ESHOST_STATUS=$?
if [ "${ESHOST_STATUS:-0}" != 0 ]; then
    echo "xs and/or v8 not known to eshost; please eshost --add them per README.md."
    exit $ESHOST_STATUS
fi
```

packages/benchmark/run-tests.sh
```diff
@@ -0,0 +1,48 @@
...
+
+echo "Running eshost..."
+
+yarn rollup -c
+
+yarn eshost -h xs,v8 dist/bundle.js
```
gibson042, line 48:
```suggestion
yarn rollup -c

yarn eshost -h xs,v8 dist/bundle.js
```

gibson042 (approved):
LGTM after addressing a few notes.

--------

_[also, capturing comments from elsewhere (without them blocking this PR]_

I think we want to drop the nesting; there's no need for a `test` function.
```js
import { benchmark } from '@endo/benchmark';

// Each benchmark will be run with each element of the cross product of all arguments:
// * length 10, fillWith 'x'
// * length 100, fillWith 'x'
// * length 1000, fillWith 'x'
//
// This supports investigation of both scaling properties and of asymmetric input
// (cache hit vs. miss, match vs. not, accept vs. reject, small vs. large input, etc.).
// But such configuration is optional; when absent, each benchmark is observed without variations.
benchmark.setArgumentDomains({
  length: [10, 100, 1000],
  fillWith: ['x'],
});

// This assumes that values from `setArgumentDomains` are provided as
// separate arguments. Another alternative is providing a single record with
// named properties, albeit with the minor extra costs and boilerplate
// associated with destructuring (e.g., `(b, { length, fillWith }) => {…}`).
benchmark('Array(length).fill', (b, length, fillWith) => {
  // To support measuring fast code that would be distorted by otherwise
  // unnecessary function calls, `N` tells the function under test how
  // many times to exercise itself. Each callback body should therefore
  // have this for-loop shape, and the framework might even detect
  // incorrect callbacks by initializing `N` to a self-replacing getter
  // (e.g., `{ recordGet(); defineProperty(b, 'N', { value: N }); return N; }`).
  for (let i = b.N; i >= 0; i--) {
    b.result = Array(length).fill(fillWith);
  }
});

// `benchmark.withTransform` is *roughly* analogous to ava `test.macro`.
// Transformation functionality is important for e.g. measuring
// `passStyleOf(bigObject)` without inappropriately including the time to
// construct `bigObject`.
const fromFiller = benchmark.withTransform((b, length, fillWith) => {
  // This transformation of length → { length } and fillWith → () => fillWith
  // takes place outside of measurement, and therefore avoids distorting it.
  return [{ length }, () => fillWith];
});
benchmark('Array.from', fromFiller, (b, lengthObj, fillFn) => {
  for (let i = b.N; i >= 0; i--) {
    b.result = Array.from(lengthObj, fillFn);
  }
});

// or with the record approach:
// ```js
// const fromFiller = benchmark.withTransform((b, { length, fillWith }) => {
//   return { lengthObj: { length }, fillFn: () => fillWith };
// });
// benchmark('Array.from', fromFiller, (b, { lengthObj, fillFn }) => {…});
// ```

// or further dropping the "with" formal pattern:
// ```js
// const fromFiller = (b, { length, fillWith }) => {
//   b.context = { lengthObj: { length }, fillFn: () => fillWith };
// };
// benchmark('Array.from', fromFiller, b => {
//   const { lengthObj, fillFn } = b.context;
//   for (let i = b.N; i >= 0; i--) {
//     b.result = Array.from(lengthObj, fillFn);
//   }
// });
// ```

File: packages/benchmark/README.md [hunk 18]
```diff
@@ -0,0 +1,17 @@
...
+Run the command `yarn test` in the `packages/benchmark` folder.
+
+## Pre-reqs of running `yarn test`
+### Install [`esvu`](https://www.npmjs.com/package/esvu) globally 
+Select the `v8` and `xs` from CLI so that it can download the binaries of those engines. 
+
+### Install [`eshost-cli`](https://www.npmjs.com/package/eshost) globally.
+Run the following commands to configure the binaries of engines with `eshost`.
+* `eshost --add "v8" d8 $ESHOST_PATH_V8`
+* `eshost --add "xs" xs $ESHOST_PATH_XS`
```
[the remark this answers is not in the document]
muhammadahmadasifbhatti: I copied from the `eshost` with the same motivation that sometimes people change the directories manually. So, instead of a hardcode path just using a variable. 
But, I get the idea. I think it should be more straight forward. Changing it.

packages/benchmark/run-tests.sh
muhammadahmadasifbhatti: Again, I did this for the CI.

muhammadahmadasifbhatti:
For [this comment](https://github.com/endojs/endo/pull/2720#pullrequestreview-2637241990).
By looking at the code you suggested, I guess this will change the whole scope of the PR. What do you think? The idea is to ship and iterate and think this way we will be blocked on this another couple of days. 

One more thing; Although, we both agree that `benchmark` should not be nested but we might need some other opinions on this. So, we can ask this question in the `endo` channel and then work on it.

packages/benchmark/run-tests.sh
gibson042: Changing ci.yaml is _entirely_ appropriate; this adds a new requirement on the host environment just like the one currently fulfilled by setup-node.

gibson042: And for local purposes, a package.json script for use like `yarn run install-engines` would make the consequences more clear.

packages/benchmark/run-tests.sh
gibson042: > The first three are the logs for the debugging purposes. I still wanted to have them in case something goes off after merging the PR in the CI.

But they won't serve that purpose without either a preceding echo or a `set -x`.

> In the last two lines, we are changing the permissions of engine binaries so that we can execute them to run our bundled.

Aren't they already marked executable when installed? I've never had to chmod them locally.

packages/benchmark/run-tests.sh
gibson042: But it shouldn't degrade the local experience. Something like this would still accomplish CI visibility:
```
yarn eshost -h xs,v8 -se true >/dev/null 2>&1 || ESHOST_STATUS=$?
if [ "${ESHOST_STATUS:-0}" != 0 ]; then
    echo "xs and/or v8 not known to eshost; please eshost --add them per README.md."
    echo "# eshost --list"
    yarn eshost --list
    exit $ESHOST_STATUS
fi
```

gibson042:
> For [this comment](https://github.com/endojs/endo/pull/2720#pullrequestreview-2637241990). By looking at the code you suggested, I guess this will change the whole scope of the PR. What do you think? The idea is to ship and iterate and think this way we will be blocked on this another couple of days.

Right, I hoped the preface would make that clear. That comment is not requesting changes for this PR, just documenting more publicly where I want things to end up.

packages/benchmark/run-tests.sh
muhammadahmadasifbhatti: I am still having second thoughts about this change. Although, what you are saying is right, but we know that there are follow-up tasks aligned. This code might not even exist in the next PR.  So, we have to reverse the `ci.yaml` again? 

Another reason is `ci.yaml` is being run for every package in the `package/` directory. So, it will be unnecessary side effects to install `esvu` and `eshost` globally. 

@turadg had a similar idea in his mind to not install `esvu` and `eshost` globally. [Link to the comment](https://github.com/endojs/endo/pull/2720#issuecomment-2664210550).


I am more inclined towards having different script (as you mentioned in the second comment).
> And for local purposes, a package.json script for use like yarn run install-engines would make the consequences more clear.

* `yarn install-engines` which will do all the `esvu` and `eshost` related stuff.
* `yarn test:bench` to run the benchmark tests. In this case bundle the code and run the tests in `index.test.js`
* `yarn test` to run `yarn install-engines` and `yarn test:bench`. Might be vague in names but open for suggestions.

packages/benchmark/run-tests.sh
muhammadahmadasifbhatti: > But they won't serve that purpose without either a preceding echo or a set -x
I am able to see the logs both on local and in ci. Nevertheless, pushing an `echo` because that seems right.
image

> Aren't they already marked executable when installed? I've never had to chmod them locally.
Nope, I had to mark change the `mod` locally. Also, at first I push the changes without changing the `mod` and got an error on the `CI` i.e. Either the file does not exist or it is not executable.

packages/benchmark/run-tests.sh
muhammadahmadasifbhatti: I think we are not on the same page here. 
Let's suppose I accept the suggested changes. Now, the `CI` will run `yarn test` and it we will immediately faill because of error `xs and/or v8 not known to eshost; please eshost --add them per README.md.`

This will work only if we update the `ci.yaml` to run these commands (of adding the engines) for us.

packages/benchmark/run-tests.sh
muhammadahmadasifbhatti: But, even with different scripts `yarn` will be modifying the `$HOME` directory.

gibson042: > I am still having second thoughts about this change. Although, what you are saying is right, but we know that there are follow-up tasks aligned. This code might not even exist in the next PR. So, we have to reverse the `ci.yaml` again?

Yes. You can add a comment as a reminder.

> Another reason is `ci.yaml` is being run for every package in the `package/` directory. So, it will be unnecessary side effects to install `esvu` and `eshost` globally.

That's not what I'm seeing at https://github.com/endojs/endo/blob/master/.github/workflows/ci.yml. Only the [job "test"](https://github.com/endojs/endo/blob/master/.github/workflows/ci.yml#L58-L95) runs `yarn test` at package scope, and it does so after some setup steps. So what I'm asking for is insertion of esvu engine installation and eshost setup before its `yarn build` step, which will run exactly as often as the rest of those steps (i.e., once for each item of the (`[node-version](node-version: [18.x, 20.x, 22.4.1])`, `platform: [ubuntu-latest]`) `matrix` cross product).

> I am more inclined towards having different script (as you mentioned in the second comment).

Having the separate package.json script is good, and the CI job step could even use it to reduce duplication. But the step on its own does not resolve this issue, because CI _must_ have the engines but local `yarn test` _must not_ install them.

> > And for local purposes, a package.json script for use like yarn run install-engines would make the consequences more clear.
> 
> * `yarn install-engines` which will do all the `esvu` and `eshost` related stuff.
> * `yarn test:bench` to run the benchmark tests. In this case bundle the code and run the tests in `index.test.js`
> * `yarn test` to run `yarn install-engines` and `yarn test:bench`. Might be vague in names but open for suggestions.

No, I'm still objecting to `yarn test` modifying $HOME.

packages/benchmark/run-tests.sh
gibson042: > This will work only if we update the `ci.yaml` to run these commands (of adding the engines) for us.

Yes, that's exactly what I'm requesting. `yarn test` on its own requires the engines to already exist and fails with an informative message when they don't. ci.yaml makes sure that the engines are installed before invoking `yarn test`.

packages/benchmark/run-tests.sh
```diff
@@ -0,0 +1,46 @@
...
+echo $(ls -la "$HOME/.esvu/engines/xs")
+echo $(ls -la "$HOME/.esvu/engines/v8")
+
+
+chmod +x "$HOME/.esvu/engines/xs/xst"
+chmod +x "$HOME/.esvu/engines/v8/d8"
+
+yarn eshost --add "xs" xs "$HOME/.esvu/engines/xs/xst"
+yarn eshost --add "v8" d8 "$HOME/.esvu/engines/v8/d8"
+
```
gibson042, line 41: None of this belongs outside of a "setup host environment" container.
```suggestion
```

packages/benchmark/run-tests.sh
```diff
@@ -0,0 +1,46 @@
...
+        echo "$INSTALL_OUTPU_XS"
+        echo "$INSTALL_OUTPU_V8"
+        exit 1
+    fi
+fi
+
+
+echo $(ls -la "$HOME/.esvu/engines")
+echo $(ls -la "$HOME/.esvu/engines/xs")
+echo $(ls -la "$HOME/.esvu/engines/v8")
```
gibson042, line 33: I think you misunderstood my "echo" suggestion; the point was to indicate _which_ directory is being listed.
```suggestion
for d in "$HOME/.esvu/engines" "$HOME/.esvu/engines/xs" "$HOME/.esvu/engines/v8"; do
    echo "# $d"
    ls -la "$d"
done
```

gibson042 (changes requested):

muhammadahmadasifbhatti:
@gibson042 thanks for the detailed review. I am installing the `ehost` and `esvu` globally. 
How?
* Have a script(in the root directory of `endo`) named `install-global-dependencies.sh`. 
* `install-global-dependencies.sh` will install `eshost` and `esvu` globally. It will also install the necessary engines i.e. `v8` and `xs` add them in the `eshost` config.
* Change the `ci.yaml` to run `install-global-dependencies.sh` before the `yarn build` step. 
* Finally, `run-tests.sh` will only be bundling the code and running them in both engines (`v8`, and `xs`) using `eshost`.

Just putting it out here, in case you have any other reservations please do comment. Meanwhile, I am working on this change.

gibson042:
> Suggesting the following changes.
> 
> * Have a script(in the benchmark directory) named `install-engines.sh`.
> * `iinstall-engines.sh` will install `eshost` and `esvu`. It will also install the necessary engines i.e. `v8` and `xs` add them in the `eshost` config.
> * Change the `ci.yaml` to run `yarn workspace benchmark run install-engines` before the `yarn build` step.
> * Finally, `run-tests.sh` will only be bundling the code and running them in both engines (`v8`, and `xs`) using `eshost`.

Thanks, that approach works for me. And to belatedly explain why it matters: making unrequested changes to a user's global environment is not consistent with the disciplined reputation we are building, and so shouldn't be part of our publically-committed code even while we're still working on it.

muhammadahmadasifbhatti:
> > Suggesting the following changes.
> 
> > 
> 
> > * Have a script(in the benchmark directory) named `install-engines.sh`.
> 
> > * `iinstall-engines.sh` will install `eshost` and `esvu`. It will also install the necessary engines i.e. `v8` and `xs` add them in the `eshost` config.
> 
> > * Change the `ci.yaml` to run `yarn workspace benchmark run install-engines` before the `yarn build` step.
> 
> > * Finally, `run-tests.sh` will only be bundling the code and running them in both engines (`v8`, and `xs`) using `eshost`.
> 
> 
> 
> Thanks, that approach works for me. And to belatedly explain why it matters: making unrequested changes to a user's global environment is not consistent with the disciplined reputation we are building, and so shouldn't be part of our publically-committed code even while we're still working on it.

@gibson042 Thanks for the explaination. Will keep that in mind. 
I have made the suggested changes, you can review the PR now.

gibson042 (approved):
Thanks @muhammadahmadasifbhatti.

PR #2655, opened 2024-12-05
fix(ses): `removeUnpermittedIntrinsics` on Hermes

leotm:
Follow-up to
- https://github.com/endojs/endo/pull/2624

Progresses
- https://github.com/endojs/endo/issues/1891

---

## Description

Fix `removeUnpermittedIntrinsics` (previously `whitelistIntrinsics`) on Hermes

```console
SES Removing unpermitted intrinsics
  Removing intrinsics.Promise.caller
  failed to delete intrinsics.Promise.caller TypeError: Property is not configurable
Uncaught TypeError: Property is not configurable
```

after starting again on the problem (now with https://github.com/endojs/endo/pull/2624 addressing `completePrototypes`), it boils down to tolerating two more undeletable non-standard properties (`.caller` and `.arguments`) on the remaining Hermes functions after attempting to remove them
edit: and `.prototype` on some properties

so we can simply extend `cauterizeProperty` after dealing with prototypes to deal with these additional properties

repro
1.  `gh pr checkout 2334` https://github.com/endojs/endo/pull/2334
2. enable Hermes VM in `packages/ses/scripts/hermes-test.sh`
3. comment `// assertDirectEvalAvailable()` in `packages/ses/src/lockdown.js`
4. `yarn build:hermes`
5. `yarn test:hermes`

<details>
  <summary>Hermes VM output (before)</summary>

```console
➜  endo git:(ses-hermes) ✗ yarn build:hermes && yarn test:hermes
-- Building 'hermes' version of SES --
Bundle size: 448125 bytes
Copied ./types.d.ts to ./dist/types.d.cts
Concatenating: dist/ses-hermes.cjs + test/_hermes-smoke.js
Generated: test/_hermes-smoke-dist.js
Executing: test/_hermes-smoke-dist.js on Hermes compiler
test/_hermes-smoke-dist.js:13007:27: warning: the variable "Compartment" was not declared in function "?anon_0_?anon_0_testCompartmentHooks"
  const compartment = new Compartment({}, {}, { resolveHook, importHook });
                          ^~~~~~~~~~~
test/_hermes-smoke-dist.js:13015:3: warning: the variable "assert" was not declared in function "?anon_0_?anon_0_testCompartmentHooks"
  assert(module);
  ^~~~~~
test/_hermes-smoke-dist.js:3838:12: warning: the variable "AggregateError" was not declared in anonymous function " 119#"
if( typeof AggregateError!==  'undefined') {
           ^~~~~~~~~~~~~~
[...]

PR #2723, opened 2025-02-14
feat(ses): update lockdown init and enable Hermes VM

leotm:
Closes: https://github.com/endojs/endo/issues/1891 (tracker), tested on https://github.com/endojs/endo/pull/2334, [Endo Sync: 2025-01-29](https://youtu.be/IyNMk5rC7CU?si=0KSHEGf2QtfsKjNW&t=264)

Previous iterations: https://github.com/endojs/endo/pull/2723#issuecomment-2759145021

TODO
- [x] add new lockdown option
  - [x] <s>legacyHermesTaming: safe (default), unsafe</s>
  - [x] hostEvaluators: '_legacy' (default), 'all' (regular), 'none' (strict csp), 'no-direct' (hermes)
  - [x] fail on safe-eval (default) and no-direct (hermes) early
    - i.e. require unsafe-eval and no-direct on hermes (since _with_ not supported)
  - [x] add assertions (on non _legacy options) and types
  - [x] update [ses/error-codes/SES_DIRECT_EVAL.md](https://github.com/endojs/endo/blob/master/packages/ses/error-codes/SES_DIRECT_EVAL.md)
  - [x] update [lockdown.md](https://github.com/endojs/endo/blob/master/packages/ses/docs/lockdown.md)
  - [x] refactor assertDirectEvalAvailable to probeHostEvaluators
    - [x] return results, now used for further assertions
    - [x] test Function() constructor and eval() separately then together
  - [x] ensure backwards compatibility (lockdown({}) - no option provided)
    - before: worked with a strict CSP (assertDirectEvalAvailable swallows strict CSP error then skips SES_DIRECT_EVAL error)
    - after: 'all' as default (breaking change) with a strict CSP will error, 'none' is now required
    - endo sync: <s>4th option undefined</s>, warn (with new error reporter) to use the new lockdown option
    - <s>since getEnvironmentOption default must be a string ('all'), warn instead of SES_DIRECT_EVAL error with a strict CSP</s>
    - default to '_legacy' under-the-hood if lockdown({}), warn to use new option; SES_DIRECT_EVAL error with a normal CSP, also when set to 'all'
    - document TODOs to replace '_legacy' with 'all' as default in future (the breaking change)
  - <s>disable compartment-shim when bundling ses for hermes</s>
  - [x] <s>disable: globalThis Compartment and testCompartmentHooks</s>
  - [x] disable compartmentInstance.evaluate instead of deleting compartment
    - makeCompartmentConstructor setGlobalObjectEvaluators safeEvaluate is the default (no evalTaming yet)
    - CompartmentPrototype.evaluate returns compartmentEvaluate safeEvaluate is the default (no evalTaming)
    - endo sync: throw error on usage, if env lacks direct eval support
- [x] test on hermes
  - [x] evalTaming: no-eval/safe-eval/unsafe-eval
  - [x] hostEvaluators: all/none/no-direct
  - [x] post-lockdown: eval(42), eval('42'), Function('return 42')()
- [x] check prev test branch (below)
- [x] fix typos
- [x] update smoke test with Hermes eval assertions
- [x] enable Hermes VM in CI
- [x] https://github.com/endojs/endo/pull/2739 merged 🎉 resolve conflicts
- endo sync
  - lockdown options enum values kebab-case <s>vs camelCase</s>
  - could make no-direct a new evalTaming option instead, however this would mean no longer being able to choose unsafe-eval or no-eval, one would have to be the assumed default
  - could use no-eval instead of unsafe-eval (example added in docs)
  - dynamic no-direct instead of via lockdown option? no since we don't know the user intention, whether to throw SES_DIRECT_EVAL or allow SES to init
  - why not error on safe-eval? did this [originally](https://github.com/endojs/endo/pull/2723#discussion_r1958551028), but better to fail earlier when there are incompatible lockdown options (safe-eval requires direct eval)
  - no-compartment-eval instead of no-direct? maybe not since we also now allow SES to init w/o direct eval
  - `--disallow-code-generation-from-strings` useful node flag to emulate a strict csp runtime in tests
- [x] re-review

Follow-up: new Compartment() fails on removeUnpermittedIntrinsics at
  - `Tolerating undeletable intrinsics.%CompartmentPrototype%.importNow.prototype === undefined`
  - `Uncaught TypeError: property is not configurable`

test branch https://github.com/endojs/endo/tree/ses-hermes-p2 (from https://github.com/endojs/endo/pull/2334)
- `yarn build:hermes` bundle ses for hermes
- `yarn test:hermes` run ses/test/_hermes-smoke.js 

Hermes eval behaviour on `bin/hermesc` (standalone compiler) and `bin/hermes` (vm, eshost)

```js
// Hermes direct eval cannot access or modify variables within its closure scope (sloppy and indirect)

eval = globalThis.eval; // remains direct eval, compiler: "error: invalid assignment left-hand side"

// calling globalThis.eval(...) uses indirect eval

// calling eval on Hermes VM emits: "warning: Direct call to eval(), but lexical scope is not supported."

print(eval(1 + 1));
print(eval("1 + 1"));

function a(eval, TEST) {
  // compiler: "error: cannot declare 'eval'"
  eval("TEST=true");
  return TEST;
}
print(a(eval, false)); // vm: false

function b(eval) {
  // compiler: "error: cannot declare 'eval'"
  var TEST = false;
  eval("TEST=true");
  return TEST;
}
print(b(eval)); // vm: false

function c() {
  var TEST = false;
  eval("TEST=true");
  return TEST;
}
print(c()); // vm: false

function d() {
  var TEST2 = 42;
  return eval("TEST2");
}
print(d()); // ReferenceError: Property 'TEST2' doesn't exist

// Note: eshost only runs bin/hermes (vm), not bin/hermesc (compiler)
```


packages/ses/src/lockdown.js
```diff
@@ -408,6 +382,12 @@ export const repairIntrinsics = (options = {}) => {
     markVirtualizedNativeFunction,
   });
 
+  if (legacyHermesTaming === 'unsafe') {
```
leotm, line 385:
```suggestion
  if (legacyHermesTaming === 'safe') {
```

i.e. disable Compartment by default

packages/ses/src/make-eval-function.js
```diff
@@ -19,7 +23,30 @@ export const makeEvalFunction = safeEvaluate => {
         // rule. Track.
         return source;
       }
-      return safeEvaluate(source);
+      if (legacyHermesTaming === 'unsafe') {
+        throw TypeError(
+          `Legacy Hermes unsupported eval() with strings arguments cannot be tamed safe under legacyHermesTaming ${legacyHermesTaming}
+  See: https://github.com/facebook/hermes/issues/1056
+  See: https://github.com/endojs/endo/issues/1561
```
leotm, line 30:
```suggestion
  See: https://github.com/endojs/endo/discussions/1944
```

packages/ses/src/make-eval-function.js
```diff
@@ -19,7 +23,30 @@ export const makeEvalFunction = safeEvaluate => {
         // rule. Track.
         return source;
       }
-      return safeEvaluate(source);
+      if (legacyHermesTaming === 'unsafe') {
+        throw TypeError(
+          `Legacy Hermes unsupported eval() with strings arguments cannot be tamed safe under legacyHermesTaming ${legacyHermesTaming}
```
leotm, line 28:
```suggestion
          `Legacy Hermes unsupported eval() called with strings arguments cannot be tamed safe under legacyHermesTaming ${legacyHermesTaming}
```

packages/ses/src/make-eval-function.js
```diff
@@ -19,7 +23,30 @@ export const makeEvalFunction = safeEvaluate => {
...
+      // } catch (e) {
+      //   // throw Error(e); // Uncaught Error: SyntaxError: 2:5:invalid statement encountered.
+      //   throw TypeError(
+      //     `legacy Hermes unsupported eval() with strings arguments cannot be tamed safe under legacyHermesTaming ${legacyHermesTaming}
+      // see: https://github.com/facebook/hermes/issues/1056
+      // see: https://github.com/endojs/endo/issues/1561
+      // did you mean evalTaming: 'unsafeEval'?`,
+      //   );
+      // }
+      // Disabling safeEvaluate is not enough, since returning the source string is not evaluating it.
```
leotm, line 48: delete, move to comment

packages/ses/src/commons.js
```diff
@@ -380,3 +380,73 @@ export const FERAL_STACK_GETTER = feralStackGetter;
...
+  if (!allowed && !evaluatorsBlocked) {
+    // See https://github.com/endojs/endo/blob/master/packages/ses/error-codes/SES_DIRECT_EVAL.md
+    // throw TypeError(
+    //   `SES cannot initialize unless 'eval' is the original intrinsic 'eval', suitable for direct-eval (dynamically scoped eval) (SES_DIRECT_EVAL)`,
+    // );
+    // TODO (hermes): (legacyHermesTaming === 'safe')
+    //     throw TypeError(
+    //       `SES cannot initialize unless 'eval' is the original intrinsic 'eval', suitable for direct-eval (dynamically scoped eval) (SES_DIRECT_EVAL)
+    // See: https://github.com/facebook/hermes/issues/957`,
+    //     );
```
leotm, line 418: *unsafe

packages/ses/src/lockdown.js
```diff
@@ -100,38 +99,6 @@ const safeHarden = makeHardener();
 // only ever need to be called once and that simplifying lockdown will improve
 // the quality of audits.
 
-const assertDirectEvalAvailable = () => {
```
naugtur, line 103: There's no good reason to move it if it's only used in this file anyway.

packages/ses/src/lockdown.js
```diff
@@ -199,6 +169,10 @@ export const repairIntrinsics = (options = {}) => {
     ...extraOptions
   } = options;
 
+  legacyHermesTaming === 'safe' ||
```
naugtur, line 172: I disagree with the name here fitting the naming convention of other options. 
We need an option to indicate whether the environment is expected to have the direct eval behavior.

Option A
`directEvalMissingHost` a boolean 
Option B
`hostEvaluator` one of: `none` for CSP blocking all eval, `all` for the regular environments, `no-direct` for what we need for hermes

In both options, we want the `assertDirectEval` function to verify our assumptions, which makes the B option more appealing to me

packages/ses/src/lockdown.js
```diff
@@ -408,6 +382,12 @@ export const repairIntrinsics = (options = {}) => {
     markVirtualizedNativeFunction,
   });
 
+  if (legacyHermesTaming === 'unsafe') {
+    globalThis.testCompartmentHooks = undefined;
+    // @ts-ignore Compartment does exist on globalThis
+    delete globalThis.Compartment;
```
naugtur, line 388: getting rid of Compartment makes sense for Hermes for now, but eventually we will want the `compartmentInstance.globalThis` to work correctly, so I'd rather make the Compartment's code eval capabilities go.

Either remove or, I think preferably, replace with a function that throws a descriptive error, the following:
- `c.evaluate` instance function
- Compartment constructor accepting importHook and related
I think that could be achieved by checking the option and throwing an error in make safe evaluator somewhere.

packages/ses/src/make-eval-function.js
```diff
@@ -19,7 +23,30 @@ export const makeEvalFunction = safeEvaluate => {
         // rule. Track.
         return source;
       }
-      return safeEvaluate(source);
```
naugtur, line 22: If the condition and throwing gets implemented in the safeEvaluate function, it'll cover more cases from one place I think. (as mentioned in https://github.com/endojs/endo/pull/2723/files#r1961688463)

packages/ses/src/make-function-constructor.js
```diff
@@ -12,9 +12,10 @@ const { Fail } = assert;
 /*
  * makeFunctionConstructor()
  * A safe version of the native Function which relies on
- * the safety of safeEvaluate for confinement.
+ * the safety of safeEvaluate for confinement, unless noEval
+ * is specified (then a TypeError is thrown).
  */
-export const makeFunctionConstructor = safeEvaluate => {
+export const makeFunctionConstructor = evaluator => {
```
naugtur, line 18: AFAIR our conclusion was, as long as we don't want compartmentalization, the indirect eval used for tamed eval and tamed Function would transparently work and we don't need to throw an error when that's used.

packages/ses/src/make-eval-function.js
leotm: also needs to throw on makeFunctionConstructor to cover all cases

packages/ses/src/global-object.js
```diff
@@ -146,14 +146,16 @@ export const setGlobalObjectMutableProperties = (
...
  * @param {(object) => void} markVirtualizedNativeFunction
+ * @param {string} [legacyHermesTaming]
  */
 export const setGlobalObjectEvaluators = (
   globalObject,
   evaluator,
   markVirtualizedNativeFunction,
+  legacyHermesTaming,
 ) => {
   {
```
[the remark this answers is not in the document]
leotm: yes unless refactored
`const f = freeze(makeEvalFunction(evaluator, legacyHermesTaming));`
`const f = freeze(makeFunctionConstructor(evaluator));`

packages/ses/src/lockdown.js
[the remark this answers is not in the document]
leotm: i see so currently deleting compartment breaks hermes on webpack (repack)? 😅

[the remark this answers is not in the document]
leotm: and only disabling compartmentInstance.evaluate would be more ideal?
to still be able to construct instances (with fresh tamed copies of compartmentInstance.globalThis)

packages/ses/src/lockdown.js
```diff
@@ -242,7 +249,13 @@ export const repairIntrinsics = (options = {}) => {
   // trace retained:
   priorRepairIntrinsics.stack;
 
-  assertDirectEvalAvailable();
+  if (legacyHermesTaming === 'safe') {
+    assertDirectEvalAvailable();
+  } else if (legacyHermesTaming === 'unsafe') {
+    warn(
+      `SES initializing with an unoriginal intrinsic 'eval', not suitable for direct-eval (dynamically scoped eval) (SES_DIRECT_EVAL)`,
```
leotm, line 256:
```suggestion
  } else if (legacyHermesTaming === 'unsafe') {
    // See https://github.com/facebook/hermes/issues/957
    warn(
      `SES initializing with an unoriginal intrinsic 'eval', not suitable for direct-eval (dynamically scoped eval) (SES_DIRECT_EVAL)`,
    );
```

packages/ses/src/lockdown.js
leotm: noting possible hermes scenarios (with Option B)

evalTaming: 'safeEval' (default)
- [hermes] assertDirectEvalAvailable throw TypeError SES_DIRECT_EVAL

evalTaming: 'noEval'
- [all] assertDirectEvalAvailable throw TypeError SES_NO_EVAL
- (creating a compartment still possible in theory)

evalTaming: 'unsafeEval'
hostEvaluator: 'all' (default)
- [hermes] assertDirectEvalAvailable throw TypeError SES_DIRECT_EVAL

evalTaming: 'unsafeEval'
hostEvaluator: 'no-direct' 
- don't throw TypeError SES_DIRECT_EVAL in assertDirectEvalAvailable
- create compartments, throw TypeError on compartment.evaluate

evalTaming: 'safeEval'
hostEvaluator: 'no-direct'
- don't throw TypeError SES_DIRECT_EVAL in assertDirectEvalAvailable
- create compartments, throw TypeError on compartment.evaluate
- throw TypeError on eval string args and function constructors (makeSafeEvaluator)
  - (better than Hermes' ambiguous SyntaxError, no clue its failing on `with` statement)

packages/ses/src/make-function-constructor.js
leotm: this was more a refactor since the arg name and JSDoc both assume safeEvaluate which is incorrect (it could be noEvalEvaluate)

leotm: regardless of compartmentalization, the (hermes) indirect eval used for tamed eval (string args only) and tamed Function doesn't transparently work, since safeEval eventually calls makeSafeEvaluator, which throws Hermes' ambiguous SyntaxError, so i think it's better to throw an error that safeEval requires an engine that supports `with`

packages/ses/src/make-eval-function.js
leotm: throw earlier instead

packages/ses/src/make-eval-function.js
```diff
@@ -19,7 +23,15 @@ export const makeEvalFunction = safeEvaluate => {
...
+      if (legacyHermesTaming === 'unsafe') {
+        throw TypeError(
+          `Legacy Hermes unsupported eval() called with string arguments cannot be tamed safe under legacyHermesTaming ${legacyHermesTaming}
+  See: https://github.com/facebook/hermes/issues/1056
+  See: https://github.com/endojs/endo/discussions/1944
+Did you mean evalTaming: 'unsafeEval'?`,
+        );
+      }
+      return evaluator(source);
```
leotm, line 34: this is better than `Uncaught SyntaxError: 2:5:invalid statement encountered`

leotm:
```suggestion
      try {
        evaluator(source);
      } catch (e) {
        if (
          e.name === 'SyntaxError' &&
          e.message.includes('invalid statement encountered')
        ) {
          throw TypeError(
            `Legacy Hermes unsupported eval() with strings arguments cannot be tamed safe under legacyHermesTaming ${legacyHermesTaming}
  See: https://github.com/facebook/hermes/issues/1056
  See: https://github.com/endojs/endo/discussions/1944
Did you mean evalTaming: 'unsafeEval'?`,
          );
        }
      }
```

leotm: try/catch preferred to support safeEval on hosts supporting `with` statement

packages/ses/src/make-eval-function.js
leotm: consider throwing earlier instead at global-object.js (src/lockdown.js too early)

leotm: under safeEval we want run makeSafeEvaluator on no-direct-eval hosts that support the with statement
(e.g. yet-another-es-engine, static hermes [with statement](https://github.com/facebook/hermes/pull/1571) backported to legacy hermes)
i.e. try makeSafeEvaluator on string, runs transparently if `with` supported, otherwise throw TypeError

packages/ses/src/make-eval-function.js
leotm: else re throw original error, in case yet-another-es-engine fails on evaluateFactory

packages/ses/src/lockdown.js
```diff
@@ -208,6 +244,9 @@ export const repairIntrinsics = (options = {}) => {
     evalTaming === 'noEval' ||
     Fail`lockdown(): non supported option evalTaming: ${q(evalTaming)}`;
 
+  (evalTaming === 'safeEval' && hostEvaluators === 'no-direct') ||
+    Fail`lockdown(): option evalTaming: ${q(evalTaming)} is incompatible with hostEvaluators: ${q(hostEvaluators)}`;
```
leotm, line 248:
```suggestion
  evalTaming === 'safeEval' &&
    hostEvaluators === 'no-direct' &&
    Fail`lockdown(): option evalTaming: ${q(evalTaming)} is incompatible with hostEvaluators: ${q(hostEvaluators)}`;
```

packages/ses/src/lockdown.js
```diff
@@ -115,20 +118,43 @@ const assertDirectEvalAvailable = () => {
...
+    case 'no-direct':
+      assert(
+        evalAllowed === false &&
+          functionAllowed === true &&
+          evaluatorsSuccessful === true,
+        `"hostEvaluators" was set to "no-direct", but ${evalAllowed === true ? 'direct eval is functional' : 'evaluators are not allowed'} (SES_DIRECT_EVAL)`,
+      );
+      break;
+    default:
+      throw TypeError('Invalid hostEvaluators option');
```
leotm, line 157:
```suggestion
      throw TypeError(`Invalid hostEvaluators option ${hostEvaluators}`);
```

leotm:
```suggestion
      Fail`Invalid hostEvaluators option ${hostEvaluators}`;
```

leotm: or remove since never reached

we reach here first

https://github.com/endojs/endo/blob/13989af9fd1206f95659a7fd779b73ad12c243fb/packages/ses/src/lockdown.js#L233-L236

packages/ses/src/lockdown.js
```diff
@@ -242,7 +252,33 @@ export const repairIntrinsics = (options = {}) => {
...
+    assert(
+      !directEvalAllowed && csp,
+      "hostEvaluators' was set to 'none', but the CSP is allowing APIs to execute (SES_DIRECT_EVAL)",
+    );
+
+  hostEvaluators === 'no-direct' &&
+    assert(
+      directEvalAllowed === false &&
+        functionAllowed === true &&
+        evalAllowed === true,
```
leotm, line 279:
```suggestion
      !directEvalAllowed && evalAllowed && functionAllowed,
```

packages/ses/src/lockdown.js
```diff
@@ -242,7 +252,33 @@ export const repairIntrinsics = (options = {}) => {
...
+  const { directEvalAllowed, functionAllowed, evalAllowed } =
+    assertDirectEval();
+
+  // A Content Security Policy containing either a default-src or a script-src directive.
+  const csp = !evalAllowed && !functionAllowed;
+
+  // Assert a regular environment where eval() and the Function() constructor are allowed to execute and direct eval() is not sloppy.
+  if (hostEvaluators === 'all' && !directEvalAllowed && !csp) {
+    throw TypeError(
+      "SES cannot initialize unless 'eval' is the original intrinsic 'eval', suitable for direct-eval (dynamically scoped eval) (SES_DIRECT_EVAL)",
```
leotm, line 264: restore docs URL comment

packages/ses/src/lockdown.js
```diff
@@ -189,6 +188,10 @@ export const repairIntrinsics = (options = {}) => {
       /** @param {string} debugName */
       debugName => debugName !== '',
     ),
+    hostEvaluators = /** @type { 'all' | 'none' | 'no-direct'  } */ (
+      // TODO: Breaking change, ensure backwards compatibility under CSP.
+      getenv('LOCKDOWN_HOST_EVALUATORS', 'all')
```
leotm, line 193: `undefined` value not currently supported by `getEnvironmentOption`

leotm:
https://github.com/endojs/endo/blob/f94be7be55488ad9bbdeb1030be2e323ed17354d/packages/ses/src/lockdown.js#L276-L279

breaking CI test cov

```console
  ✘ [fail]: error › permit-removal-warnings › permit removal warnings

    [
  -   'warn',
  +   'groupCollapsed',
  -   'Please now use the new lockdown option { hostEvaluators : \'<all|none|no-direct>\' }',
  +   'SES Removing unpermitted intrinsics',
    ]

  ✘ [fail]: error › permit-removal-warnings-node › node reporting to stderr with indented group

  - 'Please now use the new lockdown option { hostEvaluators : \'<all|none|no-direct>\' }'
  + 'SES Removing unpermitted intrinsics'
```

update tests

packages/ses/docs/lockdown.md
leotm, line 1: mention this new option is backwards-compatible with users with a strict CSP, but displays a warning

packages/ses/src/lockdown.js
```diff
@@ -208,6 +223,17 @@ export const repairIntrinsics = (options = {}) => {
...
+  // TODO: Remove '_legacy' when breaking change introduced (hostEvaluators: 'all' as default), where strict CSPs will require 'none'.
+  hostEvaluators === '_legacy' ||
+    hostEvaluators === 'all' ||
+    hostEvaluators === 'none' ||
+    hostEvaluators === 'no-direct' ||
+    Fail`lockdown(): non supported option hostEvaluators: ${q(hostEvaluators)}`;
+
+  evalTaming === 'safeEval' &&
+    hostEvaluators === 'no-direct' &&
+    Fail`lockdown(): option evalTaming: ${q(evalTaming)} is incompatible with hostEvaluators: ${q(hostEvaluators)}`;
```
leotm, line 235: document this in lockdown.md

packages/ses/src/compartment.js
```diff
@@ -326,6 +332,18 @@ export const makeCompartmentConstructor = (
...
+
+    if (hostEvaluators === 'no-direct') {
+      evaluator = () => {
+        throw TypeError(
+          'Compartment evaluation not supported without direct eval.',
+        );
+      };
+    } else {
+      evaluator = safeEvaluate;
+    }
```
leotm, line 345: document this in lockdown.md

leotm: and not yet visible in userland until `new Compartment()` compatible on Hermes under lockdown (in [tracker](https://github.com/endojs/endo/issues/1891))
`removeUnpermittedIntrinsics` > `intrinsics.%CompartmentPrototype%.importNow.prototype === undefined` > `Uncaught TypeError: property is not configurable` (similar to https://github.com/endojs/endo/pull/2655)

leotm:
flakey test https://github.com/endojs/endo/actions/runs/13784820806/job/38550261268?pr=2723

```
  ✘ [fail]: endo › indirect cancellation via worker Rejected promise returned by test
```

leotm:
consider expanding current evalTaming test for Hermes case

https://github.com/endojs/endo/blob/088b7464b380b4a4a1b43576d0771b4bd517bcd8/packages/ses/test/evalTaming-default.test.js#L1-L21

packages/ses/docs/lockdown.md
```diff
@@ -30,6 +30,7 @@ Each option is explained in its own section below.
 | `reporting`                      | `'platform'`     | `'console'` `'none'`                   | where to report warnings ([details](#reporting-options))
 | `unhandledRejectionTrapping`     | `'report'`       | `'none'`                               | handling of finalized unhandled rejections ([details](#unhandledrejectiontrapping-options)) |
 | `evalTaming`                     | `'safeEval'`     | `'unsafeEval'` `'noEval'`              | `eval` and `Function` of the start compartment ([details](#evaltaming-options)) |
+| `hostEvaluators`                 | `'all'`          | `'none'` `'no-direct'`                 | handling of sloppy and indirect eval ([details](#hostevaluators-options)) |
```
naugtur, line 33: not true AFAIR - current default is undefined with legacy behavior.

leotm: `undefined` is not a currently supported `getEnvironmentOption`

https://github.com/endojs/endo/blob/a7954e962f50145be98ea0aca556b83532f684e1/packages/env-options/src/env-options.js#L65

so the current `_legacy` default represents that, not surfaced in docs since it's under-the-hood

(equivalent to `all`, but doesn't error with a strict CSP to make it backwards-compatible)

packages/ses/error-codes/SES_DIRECT_EVAL.md
```diff
@@ -3,7 +3,19 @@
...
 If you see this error, something running before `ses` initialized, most likely
 another instance of `ses`, has replaced `eval` with something else.
+
+If you're running under an environment that doesn't support direct eval (Hermes), try setting `hostEvaluators` to `no-direct`.
+
+If you're running under CSP, try setting `hostEvaluators` to `none`.
+
+# _hostEvaluators_ was set to _none_, but evaluators are not blocked (`SES_DIRECT_EVAL`)
+
+It seems your CSP allows execution of `eval()`, try setting `hostEvaluators` to `all` or `no-direct`.
```
naugtur, line 17:
```suggestion
You indicated that you expected evaluators to be blocked (eg. by Content Security Policy) but they actually work. It's either a misunderstanding of `hostEvaluators` option or whatever method was used to block evaluators no longer works.
```

packages/ses/src/lockdown.js
```diff
@@ -100,10 +100,30 @@ const safeHarden = makeHardener();
...
+    // host, or is outright forbidden by a Content Security Policy.
+    functionAllowed = false;
+  }
+
+  let evalAllowed;
+  try {
+    evalAllowed = FERAL_EVAL('true');
+  } catch (_error) {
+    // We reach here if eval() has not been implemented by the host, or is outright
+    // forbidden by a Content Security Policy.
```
naugtur, line 118:
```suggestion
    // We reach here if eval() has not been implemented by the host or is forbidden,
    // but Function was not forbidden. 
```

packages/ses/src/lockdown.js
```diff
@@ -242,7 +272,35 @@ export const repairIntrinsics = (options = {}) => {
...
 
-  assertDirectEvalAvailable();
+  const { functionAllowed, evalAllowed, directEvalAllowed } =
+    probeHostEvaluators();
+
+  // This could be a strict Content Security Policy containing either a `default-src` or a `script-src` directive, or an ES host with broken APIs.
+  const noEvaluators = !evalAllowed && !functionAllowed; // eval() itself and the Function() constructor are not allowed to execute.
+
+  hostEvaluators === 'all' &&
+    noEvaluators &&
```
naugtur, line 282: more precisely, if we want to cover a situation where just one of the evaluators we check for is broken:
```suggestion
    !(evalAllowed && functionAllowed) &&
```

packages/ses/src/lockdown.js
```diff
@@ -242,7 +272,35 @@ export const repairIntrinsics = (options = {}) => {
...
+    Fail`'hostEvaluators' was set to 'all', but the Function() constructor and eval() are not allowed to execute (SES_DIRECT_EVAL)`;
+
+  hostEvaluators === 'none' &&
+    !noEvaluators &&
+    Fail`'hostEvaluators' was set to 'none', but the Function() constructor and eval() are allowed to execute (SES_DIRECT_EVAL)`;
+
+  hostEvaluators === 'no-direct' &&
+    directEvalAllowed &&
+    Fail`'hostEvaluators' was set to 'no-direct', but ${directEvalAllowed === true ? 'direct eval is functional' : 'the Function() constructor and eval() are not allowed to execute'} (SES_DIRECT_EVAL)`;
```
naugtur, line 291:
```suggestion
  hostEvaluators === 'no-direct' &&
    directEvalAllowed &&
    Fail`'hostEvaluators' was set to 'no-direct', but direct eval is available (SES_DIRECT_EVAL)`;
    
  hostEvaluators === 'no-direct' &&
    noEvaluators &&
    Fail`'hostEvaluators' was set to 'no-direct', but all evaluators seem blocked (SES_DIRECT_EVAL)`;
```

packages/ses/src/lockdown.js
```diff
@@ -115,21 +135,12 @@ const assertDirectEvalAvailable = () => {
...
       delete globalThis.SES_changed;
     }
-  } catch (_error) {
-    // We reach here if eval is outright forbidden by a Content Security Policy.
-    // We allow this for SES usage that delegates the responsibility to isolate
-    // guest code to production code generation.
-    allowed = true;
-  }
-  if (!allowed) {
-    // See https://github.com/endojs/endo/blob/master/packages/ses/error-codes/SES_DIRECT_EVAL.md
```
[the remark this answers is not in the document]
leotm: [déjà vu](https://github.com/endojs/endo/pull/2723#discussion_r1974147489) (; preserved here

https://github.com/endojs/endo/blob/1e5c83f06a4c1accce9e78426b9aa811fd1b4804/packages/ses/src/lockdown.js#L295-L303

leotm:
flakey test https://github.com/endojs/endo/actions/runs/13815628699/job/38647946438?pr=2723

```console
yarn workspace @endo/benchmark run install-engines
  shell: /usr/bin/bash -e {0}
yarn version: 4.5.1
Installing engines...
Error installing XS or V8:
➤ YN0000: · Yarn 4.5.1
➤ YN0000: ┌ Resolution step
Resolution step
➤ YN0000: └ Completed in 0s 720ms
➤ YN0000: ┌ Fetch step
Fetch step
➤ YN0000: └ Completed in 0s 220ms
➤ YN0000: ┌ Link step
Link step
➤ YN0000: └ Completed in 1s 28ms
➤ YN0000: · Done in 2s 2ms

esvu ❯ version 1.2.16
XS ❯ Checking version...
esvu ✖ TypeError: body.find is not a function
    at XSInstaller.resolveVersion (/tmp/xfs-4191b526/dlx-1965/node_modules/esvu/src/engines/xs.js:[3](https://github.com/endojs/endo/actions/runs/13815628699/job/38647946438?pr=2723#step:6:3)7:19)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async XSInstaller.install (/tmp/xfs-[4](https://github.com/endojs/endo/actions/runs/13815628699/job/38647946438?pr=2723#step:6:5)191b526/dlx-1965/node_modules/esvu/src/installer.js:4[5](https://github.com/endojs/endo/actions/runs/13815628699/job/38647946438?pr=2723#step:6:6):21)
    at async installEngine (/tmp/xfs-4191b52[6](https://github.com/endojs/endo/actions/runs/13815628699/job/38647946438?pr=2723#step:6:7)/dlx-1965/node_modules/esvu/src/bin.js:121:3)
    at async main (/tmp/xfs-4191b526/dlx-1965/node_modules/esvu/src/bin.js:168:[7](https://github.com/endojs/endo/actions/runs/13815628699/job/38647946438?pr=2723#step:6:8))
esvu ✖ Some engines were not installed.
➤ YN0000: · Yarn 4.5.1
➤ YN0000: ┌ Resolution step
Resolution step
➤ YN0000: └ Completed in 0s 712ms
➤ YN0000: ┌ Fetch step
Fetch step
➤ YN0000: └ Completed
➤ YN0000: ┌ Link step
Link step
➤ YN0000: └ Completed in 1s
➤ YN0000: · Done in 1s [8](https://github.com/endojs/endo/actions/runs/13815628699/job/38647946438?pr=2723#step:6:9)04ms

esvu ❯ version 1.2.16
V8 ❯ Checking version...
V8 ❯ Installing version 13.6.46
V8 ❯ Downloading https://storage.googleapis.com/chromium-v8/official/canary/v8-linux64-rel-13.6.46.zip
V8 ❯ Extracting /tmp/esvu-87ab78401[9](https://github.com/endojs/endo/actions/runs/13815628699/job/38647946438?pr=2723#step:6:10)4352020570b4679677fb0d.zip
V8 ❯ Installing /tmp/esvu-87ab7840194352020570b4679677fb0d.zip-extracted
V8 ❯ Testing...
V8 ✔ Installed with bin entries: v8
esvu ✖ Some engines were not installed.
Error: Process completed with exit code 1.
```

https://github.com/endojs/endo/blob/0def3106d33492082aa3705a223350c9a8ebbcbe/packages/benchmark/install-engines.sh#L12-L18

perhaps it could be worth falling back on [jsvu](https://github.com/GoogleChromeLabs/jsvu) when esvu fails @muhammadahmadasifbhatti

packages/ses/src/lockdown.js
leotm: i've left out `but Function was not forbidden` since it may or may not be forbidden

https://github.com/endojs/endo/blob/25407a9ce4bab4eaab8e87cf8288d35af4ba81a6/packages/ses/src/lockdown.js#L104-L121

(Function now tested first in its own try/catch, then eval tested separately after in its own try/catch)

erights:
Unless there is a surprise, I request that https://github.com/endojs/endo/pull/2739 get merged before this one, and then this one adjust to that. Thanks.

packages/ses/src/lockdown.js
```diff
@@ -218,17 +243,21 @@ export const repairIntrinsics = (options = {}) => {
...
   if (mathTaming !== undefined) {
-    // eslint-disable-next-line no-console
     warn(
       `SES The 'mathTaming' option is deprecated and does nothing. In the future specifying it will be an error.`,
     );
   }
+  if (hostEvaluators === '_legacy') {
+    // TODO: Remove when 'all' introduced as the new default option (breaking change).
+    warn(
+      `SES Please now use the 'hostEvaluators' option. In the future not specifying 'none' will error under strict CSP.`,
```
leotm, line 258: possibly worth mentioning since 'all' will be the new default, but currently straight to the point

packages/ses/docs/lockdown.md
```diff
@@ -617,6 +619,70 @@ LOCKDOWN_EVAL_TAMING=no-eval
 LOCKDOWN_EVAL_TAMING=unsafe-eval
 ```
 
+## `hostEvaluators` Options
+
+**Background**: Hermes is a JavaScript engine that does not yet support direct `eval()` nor the `with` statement. The SES `evalTaming` default option `"safeEval"` uses multiple nested `with` statements to create a restricted scope chain, so on Hermes we must run under the `"unsafeEval"` option. However SES cannot initialize unless 'eval' is the original intrinsic 'eval', suitable for direct-eval (dynamically scoped eval), which is where we introduce the `hostEvaluators` option `"no-direct"`.
```
leotm, line 624: update

packages/ses/docs/lockdown.md
```diff
@@ -617,6 +619,70 @@ LOCKDOWN_EVAL_TAMING=no-eval
...
+Further
+
+* `'all'`: asserts evaluators are allowed to execute
+* `'none'`: asserts evaluators are *not* allowed to execute
+* `'no-direct'`: asserts direct-eval is not available
+
+Hermes example
+
+```js
+lockdown({ evalTaming: 'unsafeEval', hostEvaluators: 'no-direct' }); // Recommended on Hermes
```
leotm, line 645: update

packages/ses/test/_hermes-smoke.js
```diff
@@ -6,7 +8,17 @@
  * Test calling SES lockdown.
  */
 const testLockdown = () => {
-  lockdown();
+  lockdown({ evalTaming: 'unsafeEval', hostEvaluators: 'no-direct' });
```
leotm, line 11: update

packages/ses/docs/lockdown.md
leotm, line 645: add `no-eval` example

packages/ses/src/make-eval-function.js
```diff
@@ -1,11 +1,12 @@
 /**
  * makeEvalFunction()
  * A safe version of the native eval function which relies on
- * the safety of safeEvaluate for confinement.
+ * the safety of safeEvaluate for confinement, unless noEval
```
leotm, line 4: update

packages/ses/src/make-function-constructor.js
```diff
@@ -12,9 +12,10 @@ const { Fail } = assert;
 /*
  * makeFunctionConstructor()
  * A safe version of the native Function which relies on
- * the safety of safeEvaluate for confinement.
+ * the safety of safeEvaluate for confinement, unless noEval
```
leotm, line 15: update

packages/ses/src/lockdown.js
leotm, line 1: update 6 safeeval occurences

packages/ses/docs/lockdown.md
```diff
@@ -617,6 +619,70 @@ LOCKDOWN_EVAL_TAMING=no-eval
...
+```
+
+However, attempting `"safeEval"` with `"no-direct"` we fail early, due to Hermes' lack of `with` statement
+
+```js
+lockdown({ evalTaming: 'safeEval', hostEvaluators: 'no-direct' }); // Fail: `lockdown(): option evalTaming: safeEval is incompatible with hostEvaluators: no-direct`
```
leotm, line 651:
```suggestion
lockdown({ evalTaming: 'safe-eval', hostEvaluators: 'no-direct' }); // Fail: `lockdown(): option { evalTaming: 'safe-eval' } is incompatible with { hostEvaluators: 'no-direct' }`
```

packages/ses/docs/lockdown.md
```diff
@@ -617,6 +619,70 @@ LOCKDOWN_EVAL_TAMING=no-eval
...
+**Background**: Hermes is a JavaScript engine that does not yet support direct `eval()` nor the `with` statement. The SES `evalTaming` default option `"safeEval"` uses multiple nested `with` statements to create a restricted scope chain, so on Hermes we must run under the `"unsafeEval"` option. However SES cannot initialize unless 'eval' is the original intrinsic 'eval', suitable for direct-eval (dynamically scoped eval), which is where we introduce the `hostEvaluators` option `"no-direct"`.
+
+```js
+lockdown(); // Warning: `SES Please now use the 'hostEvaluators' option. In the future not specifying 'none' will error under strict CSP.`
+// or
+lockdown({ hostEvaluators: 'all' }); // SES fails to initialize if direct-eval is not the original intrinsic 'eval'
+// vs
+lockdown({ hostEvaluators: 'none' }); // SES initializes when evaluators are not allowed to execute (e.g. a strict CSP)
+// vs
+lockdown({ hostEvaluators: 'no-direct' }); // SES initializes without direct-eval (e.g. on Hermes)
```
leotm, line 633:
```suggestion
lockdown({ hostEvaluators: 'no-direct' }); // SES initializes if direct-eval is not the original intrinsic 'eval'
```

packages/ses/docs/lockdown.md
```diff
@@ -617,6 +619,70 @@ LOCKDOWN_EVAL_TAMING=no-eval
...
 ```
 
+## `hostEvaluators` Options
+
+**Background**: Hermes is a JavaScript engine that does not yet support direct `eval()` nor the `with` statement. The SES `evalTaming` default option `"safeEval"` uses multiple nested `with` statements to create a restricted scope chain, so on Hermes we must run under the `"unsafeEval"` option. However SES cannot initialize unless 'eval' is the original intrinsic 'eval', suitable for direct-eval (dynamically scoped eval), which is where we introduce the `hostEvaluators` option `"no-direct"`.
+
+```js
+lockdown(); // Warning: `SES Please now use the 'hostEvaluators' option. In the future not specifying 'none' will error under strict CSP.`
+// or
+lockdown({ hostEvaluators: 'all' }); // SES fails to initialize if direct-eval is not the original intrinsic 'eval'
```
leotm, line 629:
```suggestion
lockdown({ hostEvaluators: 'all' }); // SES fails to initialize if direct-eval is not the original intrinsic 'eval' or evaluators are not allowed to execute
```

packages/ses/docs/lockdown.md
```diff
@@ -617,6 +619,70 @@ LOCKDOWN_EVAL_TAMING=no-eval
 LOCKDOWN_EVAL_TAMING=unsafe-eval
 ```
 
+## `hostEvaluators` Options
+
+**Background**: Hermes is a JavaScript engine that does not yet support direct `eval()` nor the `with` statement. The SES `evalTaming` default option `"safeEval"` uses multiple nested `with` statements to create a restricted scope chain, so on Hermes we must run under the `"unsafeEval"` option. However SES cannot initialize unless 'eval' is the original intrinsic 'eval', suitable for direct-eval (dynamically scoped eval), which is where we introduce the `hostEvaluators` option `"no-direct"`.
+
+```js
+lockdown(); // Warning: `SES Please now use the 'hostEvaluators' option. In the future not specifying 'none' will error under strict CSP.`
```
leotm, line 627:
```suggestion
lockdown(); // hostEvaluators defaults to 'all' and warns `SES Please now use the 'hostEvaluators' option. In the future not specifying 'none' will error under strict CSP.` if not set
```

packages/ses/src/lockdown.js
```diff
@@ -211,6 +225,17 @@ export const repairIntrinsics = (options = {}) => {
     evalTaming === 'noEval' || // deprecated
     Fail`lockdown(): non supported option evalTaming: ${q(evalTaming)}`;
 
+  // TODO: Remove '_legacy' when breaking change introduced (hostEvaluators: 'all' as default), where strict CSPs will require 'none'.
+  hostEvaluators === '_legacy' ||
```
leotm, line 229:
```suggestion
  hostEvaluators === '_legacy' || // deprecated
```

packages/ses/types.d.ts
```diff
@@ -47,6 +47,7 @@ export interface RepairOptions {
   overrideTaming?: 'moderate' | 'min' | 'severe';
   overrideDebug?: Array<string>;
   domainTaming?: 'safe' | 'unsafe';
+  hostEvaluators?: 'all' | 'none' | 'unsafe';
```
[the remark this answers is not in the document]
leotm: or

```suggestion
  hostEvaluators?:
    | 'all'
    | 'none'
    | 'unsafe'
    // deprecated
    | '_legacy';
```

packages/ses/test/_hermes-smoke.js
```diff
@@ -6,7 +8,17 @@
...
 
 testLockdown();
+
+assert(typeof eval === 'function', 'eval is not a function');
+assert(
+  eval.toString() === 'function eval() { [native code] }',
+  'eval is not a native code function',
+);
+assert(eval(42) === 42, 'eval is not functional');
+assert(eval('42') === 42, 'eval is not functional');
```
leotm, line 22:
```suggestion
assert(eval('42') === 42, 'eval called with string argument is not functional');
```

packages/ses/test/error/permit-removal-warnings-node.test.js
```diff
@@ -38,6 +38,11 @@ test('node reporting to stderr with indented group', async t => {
   const stderrText = new TextDecoder().decode(stderrBytes);
   const stderrLines = stderrText.trim().split('\n');
 
+  // Group label for removing unpermitted intrinsics
```
leotm, line 41: invalid comment

packages/ses/error-codes/SES_DIRECT_EVAL.md
```diff
@@ -3,7 +3,19 @@
...
+
+If you're running under CSP, try setting `hostEvaluators` to `none`.
+
+# _hostEvaluators_ was set to _none_, but evaluators are not blocked (`SES_DIRECT_EVAL`)
+
+You indicated that you expected evaluators to be blocked (eg. by Content Security Policy) but they actually work. It's either a misunderstanding of `hostEvaluators` option or whatever method was used to block evaluators no longer works.
+
+# `"hostEvaluators" was set to "no-direct", but direct eval is functional
+
+If evaluators are working, including direct eval, it seems you've upgraded to a version of host environment that now supports them (future Hermes). Or you're running code intended for Hermes in a different host.
```
leotm, line 21:
```suggestion
# 'hostEvaluators' was set to 'no-direct', but direct eval is available (SES_DIRECT_EVAL)

If evaluators are working, including direct eval, it seems you've upgraded to a version of 
the host environment that now supports them (future Hermes). Or you're running code intended
for Hermes in a different host.
```

packages/ses/error-codes/SES_DIRECT_EVAL.md
```diff
@@ -3,7 +3,19 @@
...
 If you see this error, something running before `ses` initialized, most likely
 another instance of `ses`, has replaced `eval` with something else.
+
+If you're running under an environment that doesn't support direct eval (Hermes), try setting `hostEvaluators` to `no-direct`.
+
+If you're running under CSP, try setting `hostEvaluators` to `none`.
+
+# _hostEvaluators_ was set to _none_, but evaluators are not blocked (`SES_DIRECT_EVAL`)
+
+You indicated that you expected evaluators to be blocked (eg. by Content Security Policy) but they actually work. It's either a misunderstanding of `hostEvaluators` option or whatever method was used to block evaluators no longer works.
```
leotm, line 17:
```suggestion
# 'hostEvaluators' was set to 'none', but the Function() constructor or eval() are allowed to execute (SES_DIRECT_EVAL)

You indicated that you expected evaluators to be blocked (e.g. by Content Security Policy),
but they actually work. It's either a misunderstanding of `hostEvaluators` option or whatever
method was used to block evaluators no longer works.
```

packages/ses/error-codes/SES_DIRECT_EVAL.md
leotm, line 1: add error code
`# 'hostEvaluators' was set to 'all', but the Function() constructor and eval() are not allowed to execute (SES_DIRECT_EVAL)`

packages/ses/test/_hermes-smoke.js
```diff
@@ -6,7 +8,17 @@
...
+
+assert(typeof eval === 'function', 'eval is not a function');
+assert(
+  eval.toString() === 'function eval() { [native code] }',
+  'eval is not a native code function',
+);
+assert(eval(42) === 42, 'eval is not functional');
+assert(eval('42') === 42, 'eval called with string argument is not functional');
+// eslint-disable-next-line no-new-func
+assert(Function('return 42')() === 42, 'eval is not functional');
```
leotm, line 24:
```suggestion
assert(
  // eslint-disable-next-line no-new-func
  Function('return 42')() === 42,
  'Function constructor is not functional',
);
```

leotm:
> Unless there is a surprise, I request that #2739 get merged before this one, and then this one adjust to that. Thanks.

thank you for #2739 ^ i've rebased and adjusted this one, ready for any feedback

packages/ses/types.d.ts
```diff
@@ -42,16 +42,12 @@ export interface RepairOptions {
     // deprecated
     | 'safeEval'
     | 'unsafeEval'
-    | 'noEval';
+    | 'noEval'
+    | 'unsafe-no-direct';
```
erights, line 46: Put this above the deprecated line, with its kabob-case siblings. Check the other occurrences for similar implicit classification.

erights:
Please update the PR title

packages/ses/NEWS.md
```diff
@@ -1,4 +1,6 @@
-User-visible changes in `ses`:
+<!-- markdownlint-disable MD041 -->
```
erights, line 1: What's this?

packages/ses/NEWS.md
```diff
@@ -19,6 +21,9 @@ User-visible changes in `ses`:
   This is a divergence from the expected behavior of a native Hardened
   JavaScript implementation, like XS.
 
+Also the new `unsafe-no-direct` option has been added to to initialize SES on
+hosts with no direct eval available (e.g. Hermes).
```
erights, line 25:
```suggestion
Also the new `unsafe-no-direct` option has been added to initialize SES on
hosts with no direct eval available (e.g. Hermes), but where you want to preserve their builtin unsafe indirect `eval` function and other evaluators (`Function`) on the global of the start compartment.

When there is no unsafe indirect `eval` or `Function`, or when they have been disabled anyway by CSP, or when you do not wish to preserve it on the start compartment, use `'no-eval'` instead.
```

This is assuming we actually need to preserve the Hermes `eval` at all. Do we have an actual use case, or is this still just hypothetical? If hypothetical, then I propose you hold `unsafe-no-direct` in a draft PR and postpone it until you need it. In the meantime, all Hermes use should just use `no-eval`.

packages/ses/docs/lockdown.md
```diff
@@ -30,6 +30,7 @@ Each option is explained in its own section below.
 | `reporting`                      | `'platform'`     | `'console'` `'none'`                   | where to report warnings ([details](#reporting-options))
 | `unhandledRejectionTrapping`     | `'report'`       | `'none'`                               | handling of finalized unhandled rejections ([details](#unhandledrejectiontrapping-options)) |
 | `evalTaming`                     | `'safe-eval'`    | `'unsafe-eval'` `'no-eval'`            | `eval` and `Function` of the start compartment ([details](#evaltaming-options)) |
+| `hostEvaluators`                 | `'all'`          | `'none'` `'no-direct'`                 | handling of sloppy and indirect eval ([details](#hostevaluators-options)) |
```
erights, line 33:
```suggestion
```

packages/ses/docs/lockdown.md
```diff
@@ -51,6 +52,7 @@ for threading environment variables into a JavaScript program.
 | `reporting`                      | `LOCKDOWN_REPORTING`                         |                       |
 | `unhandledRejectionTrapping`     | `LOCKDOWN_UNHANDLED_REJECTION_TRAPPING`      |                       |
 | `evalTaming`                     | `LOCKDOWN_EVAL_TAMING`                       |                       |
+| `hostEvaluators`                 | `LOCKDOWN_HOST_EVALUATORS`                   |                       |
```
erights, line 55:
```suggestion
```

packages/ses/docs/lockdown.md
```diff
@@ -617,6 +619,70 @@ LOCKDOWN_EVAL_TAMING=no-eval
...
+Once Hermes engine supports direct eval and the `with` statement, `evalTaming: 'safe-eval'` will be possible.
+Currently there is an open feature request and open pull request targeting Static Hermes.
+
+* <https://github.com/facebook/hermes/issues/1056>
+  * <https://github.com/facebook/hermes/pull/1571>
+
+There is also an open alternate idea to sandbox `Compartment` *without* the `with` statement.
+
+* <https://github.com/endojs/endo/discussions/1944>
+
```
erights, line 685:
```suggestion
```

packages/ses/src/compartment.js
```diff
@@ -326,6 +332,18 @@ export const makeCompartmentConstructor = (
       sloppyGlobalsMode: false,
     });
 
+    let evaluator;
+
+    if (evalTaming === 'unsafe-no-direct') {
+      evaluator = () => {
+        throw TypeError(
```
erights, line 339: Does this take care of all three evaluators: `eval`, `Function`, `Compartment` (or just Compartment `evaluate`)?

`'no-eval'` should have the same effect on Compartments as `'unsafe-no-direct'`. Why is it a separate code path? Are there any differences observable from within a Compartment?

packages/ses/src/global-object.js
```diff
@@ -75,6 +75,8 @@ export const setGlobalObjectConstantProperties = globalObject => {
  * @param {Function} args.makeCompartmentConstructor
  * @param {(object) => void} args.markVirtualizedNativeFunction
  * @param {Compartment} [args.parentCompartment]
+ * @param {string} [args.evalTaming]
+ * @param args.evalTaming
```
erights, line 79:
```suggestion
```

packages/ses/src/lockdown.js
```diff
@@ -100,10 +100,32 @@ const safeHarden = makeHardener();
...
+const probeHostEvaluators = () => {
+  let functionAllowed;
   try {
-    allowed = FERAL_FUNCTION(
+    functionAllowed = FERAL_FUNCTION('return true')();
+  } catch (_error) {
+    // We reach here if the Function() constructor is outright forbidden by a
+    // strict Content Security Policy (containing either a `default-src` or a
+    // `script-src` directive), not been implemented in the host, or the host
+    // is configured to throw an exception instead of `new Function`.
```
erights, line 111: Shouldn't `'no-eval'` also disallow these?

packages/ses/src/lockdown.js
```diff
@@ -115,21 +137,12 @@ const assertDirectEvalAvailable = () => {
     // and indirect, which generally creates a new global.
     // We are going to throw an exception for failing to initialize SES, but
     // good neighbors clean up.
-    if (!allowed) {
+    if (!directEvalAllowed) {
       delete globalThis.SES_changed;
```
erights, line 141: What's `SES_changed`?

packages/ses/src/lockdown.js
naugtur: This is where we check what the capabilities of the host are. 
`no-eval` should have an effect after this check. 

If the above is not correct, we have to look into running this earlier or changing the logic of asserting on the results of `probeHostEvaluators` a bit.

packages/ses/NEWS.md
leotm: it fixes the markdownlint warning

Screenshot 2025-03-26 at 4 23 25 pm

leotm: we don't have makedownlint (or cSpell) running in CI (yet)

packages/ses/src/lockdown.js
leotm: the global var created as a side effect when the host (Hermes) only has indirect eval available (same as [before](https://github.com/endojs/endo/blob/f27ab4d422d7e367036c2d75c27e2eea97db361c/packages/ses/src/lockdown.js#L103-L133))

packages/ses/NEWS.md
leotm: yes `unsafe-no-direct` being the _full job_ if required, but you're right `no-eval` can be used for now in the smoke test (and `eval` assertions removed), use cases will become apparent soon (after harden on React Native native solved)

leotm: - https://github.com/endojs/endo/pull/2748

packages/ses/src/lockdown.js
```diff
@@ -245,7 +257,28 @@ export const repairIntrinsics = (options = {}) => {
...
+    directEvalAllowed === false &&
+    Fail`'evalTaming' was set to 'safe-eval', but direct eval is not available (SES_DIRECT_EVAL)`;
+
+  evalTaming === 'unsafe-no-direct' &&
+    directEvalAllowed === true &&
+    Fail`'evalTaming' was set to 'unsafe-no-direct', but direct eval is available (SES_DIRECT_EVAL)`;
+
+  if (
+    directEvalAllowed === false &&
+    (evalTaming === 'safe-eval' || evalTaming === 'unsafe-eval') &&
```
leotm, line 274:
```suggestion
    evalTaming === 'safe-eval' &&
```

packages/ses/NEWS.md
```diff
@@ -19,6 +21,9 @@ User-visible changes in `ses`:
   This is a divergence from the expected behavior of a native Hardened
   JavaScript implementation, like XS.
 
+Also the `unsafe-eval` option has been updated to allow SES initialize on
+hosts with no direct eval available (e.g. Hermes) for a successful lockdown.
```
leotm, line 25:
```diff
Also the `unsafe-eval` and `no-eval` options now allow SES initialize on
hosts with no direct eval available (e.g. Hermes) for a successful lockdown.
```

or simpler now only `safe-eval` can throw SES_DIRECT_EVAL preventing SES init

packages/ses/NEWS.md
```diff
@@ -19,6 +21,9 @@ User-visible changes in `ses`:
   This is a divergence from the expected behavior of a native Hardened
   JavaScript implementation, like XS.
 
+Also now only `safe-eval` can throw error `SES_DIRECT_EVAL`. This allows SES to
+initialize with `unsafe-eval` or `no-eval` on hosts with no direct eval available (e.g. Hermes) for a successful lockdown.
```
leotm, line 25: move under new `# v1.13.0 (2025-XX-XX)`

leotm:
```
# v1.13.0 (2025-XX-XX)

The `evalTaming:` option `safe-eval` now can only throw error `SES_DIRECT_EVAL`.
This allows SES to initialize with `unsafe-eval` or `no-eval` on hosts with no
direct eval available (e.g. Hermes) for a successful lockdown.
```

erights:
@Jack-Works , this PR is now proceeding under the assumption that it would be ok with you to change the behavior of `'unsafe-eval'` slightly: On a platform (like Hermes) without any direct eval, it would be ok to move the error that compartment-evaluators-are-not-possible from initialization time to the time that someone tries to use a compartment evaluator. This would allow non-compartment use of unsafe-eval even on those platforms. On platforms with direct eval, this change should have no observable difference.

Would that change in `'unsafe-eval'` behavior be ok with you? If not, could you clarify your motivating use case for `'unsafe-eval'`? Thanks.

Jack-Works:
> @Jack-Works , this PR is now proceeding under the assumption that it would be ok with you to change the behavior of `'unsafe-eval'` slightly: On a platform (like Hermes) without any direct eval, it would be ok to move the error that compartment-evaluators-are-not-possible from initialization time to the time that someone tries to use a compartment evaluator. This would allow non-compartment use of unsafe evaluators even on those platforms. On platforms with direct eval, this change should have no observable difference.
> 
> Would that change in `'unsafe-eval'` behavior be ok with you? If not, could you clarify your motivating use case for `'unsafe-eval'`? Thanks.

Yes, it's ok with me. I'm only using this on platforms with direct eval available

packages/ses/NEWS.md
```diff
@@ -1,4 +1,12 @@
-User-visible changes in `ses`:
+<!-- markdownlint-disable MD041 -->
+
+User-visible changes in `ses`
+
+# v1.13.0 (2025-XX-XX)
+
+The `evalTaming:` option `safe-eval` now can only throw error `SES_DIRECT_EVAL`.
+This allows SES to initialize with `unsafe-eval` or `no-eval` on hosts with no
+direct eval available (e.g. Hermes) for a successful lockdown.
```
leotm, line 9:
```suggestion
- The `evalTaming:` option `safe-eval` now can only throw error `SES_DIRECT_EVAL`.
  This allows SES to initialize with `unsafe-eval` or `no-eval` on hosts with no
  direct eval available (e.g. Hermes) for a successful lockdown.
```

leotm:
🐪

```console
Run yarn install --immutable
  yarn install --immutable
  shell: /usr/bin/bash -e {0}
➤ YN0000: Yarn detected that the current workflow is executed from a public pull request. For safety the hardened mode has been enabled.
➤ YN0000: It will prevent malicious lockfile manipulations, in exchange for a slower install time. You can opt-out if necessary; check our documentation for more details.

➤ YN0000: · Yarn 4.7.0
➤ YN0000: ┌ Resolution step
Resolution step
➤ YN0035: camelcase@npm:6.3.0: camelcase@npm:5.3.1: camelcase@npm:[2](https://github.com/endojs/endo/actions/runs/14199755892/job/39783579496?pr=2723#step:5:2).1.1: The remote server failed to provide the requested resource
➤ YN00[3](https://github.com/endojs/endo/actions/runs/14199755892/job/39783579496?pr=2723#step:5:3)5:   Response Code: 403 (Forbidden)
➤ YN0035:   Request Method: GET
➤ YN0035:   Request URL: https://registry.yarnpkg.com/camelcase
➤ YN0000: └ Completed in [4](https://github.com/endojs/endo/actions/runs/14199755892/job/39783579496?pr=2723#step:5:5)s 116ms
➤ YN0000: · Failed with errors in 4s 130ms
Error: Process completed with exit code 1.
```

leotm:
@erights i've rebased the 9 commits into 6 and ready for merge, unless any further feedback

packages/ses/NEWS.md
```diff
@@ -137,13 +145,17 @@ and subject to breaking changes that will not be signaled by semver.
...
   lockdown options setting,
+
   ```js
   errorTaming: 'unsafe'
   ```
+
   or setting the environment variable
+
   ```sh
-  $ export LOCKDOWN_ERROR_TAMING=unsafe
```
erights, line 145: Was the `$` removal manual or by the .md fixer thing?

I do mildly prefer keeping the `$` prefix so readers can easily this is shell and not js.

packages/ses/docs/lockdown.md
```diff
@@ -608,6 +608,31 @@ lockdown({ evalTaming: 'no-eval' }); // disallowing calling eval like there is a
 lockdown({ evalTaming: 'unsafe-eval' });
 ```
 
+Also `'unsafe-eval'` and `no-eval` allow us to initialize SES when no direct eval is available.
```
erights, line 611:
```suggestion
Also `'unsafe-eval'` and `'no-eval'` allow us to initialize SES when no direct eval is available.
```

erights (approved):
LGTM, thanks!

packages/ses/NEWS.md
leotm: yep .md fixer https://github.com/DavidAnson/markdownlint/blob/v0.37.4/doc/md014.md

> Rationale: It is easier to copy/paste and less noisy if the dollar signs are omitted when they are not needed. See https://cirosantilli.com/markdown-style-guide#dollar-signs-in-shell-code for more information.

leotm: > Because some commands do not produce output, it is not a violation if _some_ commands do not have output:

```
$ mkdir test
mkdir: created directory 'test'
$ ls test
```

seems a bit silly that _some_ doesn't include _one_ command with no output 😄

packages/ses/NEWS.md
```diff
@@ -1,4 +1,12 @@
-User-visible changes in `ses`:
+<!-- markdownlint-disable MD041 -->
```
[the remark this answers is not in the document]
leotm: sorry i resolved https://github.com/endojs/endo/pull/2723#pullrequestreview-2714957686 too early, exactly we should add a [script](https://github.com/LavaMoat/lavamoat.github.io/blob/main/package.json#L17) (or include it in `yarn format`) then ensure it's run in CI ([example](https://github.com/LavaMoat/lavamoat.github.io/commit/eb0a4c7260b202a5ff9447f309521e2e55424130))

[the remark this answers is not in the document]
leotm: reverted formatting changes in the end to keep this PR relevant/minimal and not step on any toes

all .md docs in the codebase better updated when addressing the filed issue then reviewed separetely

leotm:
nb: next release (ses@1.13.0) `ses-hermes.cjs` shim will be consumed by

- https://github.com/LavaMoat/LavaMoat/pull/1438

to plug into Metro for bundling React Native apps

leotm:
RE https://github.com/endojs/endo/pull/2334#pullrequestreview-2376027183  @kriskowal

> Please add a draft of release notes to NEWS.md. I roll these up to a blog on hardenedjs.org now and this would likely be a headline.

and RE https://github.com/endojs/endo/pull/2108 @erights

> None. Nothing Breaking certainly. Nothing newsworthy yet. But when the ses-shim as a whole is know to support Hermes, that will be newsworthy ;)

happy to expand current NEWS.md update in follow-up a PR to be more newsworthy (;

which i imagine would land on https://github.com/endojs/hardenedjs.org/tree/main/src/content/docs/blog

i foresee the blog post could detail how SES can now be used with Metro to run Hardened JS on React Native Hermes (or JSC) apps, like we're doing in https://github.com/LavaMoat/LavaMoat/pull/1438

packages/ses/docs/lockdown.md
erights: @leotm , I don't see the single quotes around no-eval

PR #2730, opened 2025-02-26
feat(common): Improve objectMap typing

gibson042:
## Description

Inspired by https://github.com/Agoric/agoric-sdk/pull/11040 , try to type `objectMap` more precisely. I'm interested to see what this breaks.

### DIFF
--- a/packages/common/object-map.js
+++ b/packages/common/object-map.js
@@ -1,4 +1,30 @@
-const { entries, fromEntries } = Object;
+/**
+ * @typedef {<O extends Record<string, unknown>>(
+ *   obj: O,
+ * ) => { [K in keyof O]: K extends string ? [K, O[K]] : never }[keyof O][]} TypedEntries
+ */
+export const typedEntries = /** @type {TypedEntries} */ (Object.entries);
+
+/**
+ * @typedef {<
+ *   const Entries extends ReadonlyArray<readonly [PropertyKey, unknown]>,
+ * >(
+ *   entries: Entries,
+ * ) => { [Entry in Entries[number] as Entry[0]]: Entry[1] }} FromTypedEntries
+ */
+export const fromTypedEntries = /** @type {FromTypedEntries} */ (
+  Object.fromEntries
+);
+
+/**
+ * @typedef {<A extends unknown[], V>(
+ *   arr: A,
+ *   mapper: <K extends number>(el: A[K], idx: K, arr: A) => V,
+ * ) => V[]} TypedMap
+ */
+export const typedMap = /** @type {TypedMap} */ (
+  Function.prototype.call.bind(Array.prototype.map)
+);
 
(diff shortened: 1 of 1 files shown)


packages/common/object-map.js
```diff
@@ -1,4 +1,30 @@
...
+ * ) => { [K in keyof O]: K extends string ? [K, O[K]] : never }[keyof O][]} TypedEntries
+ */
+export const typedEntries = /** @type {TypedEntries} */ (Object.entries);
+
+/**
+ * @typedef {<
+ *   const Entries extends ReadonlyArray<readonly [PropertyKey, unknown]>,
+ * >(
+ *   entries: Entries,
+ * ) => { [Entry in Entries[number] as Entry[0]]: Entry[1] }} FromTypedEntries
```
mhofman, line 13: TIL [Key Remapping via `as`](https://www.typescriptlang.org/docs/handbook/2/mapped-types.html#key-remapping-via-as)

packages/common/object-map.js
```diff
@@ -33,17 +59,18 @@ const { entries, fromEntries } = Object;
...
-  const mapEnts = ents.map(
-    ([k, v]) => /** @type {[keyof O, R]} */ ([k, mapFn(v, k)]),
-  );
-  return /** @type {Record<keyof O, R>} */ (harden(fromEntries(mapEnts)));
+  const oldEntries = typedEntries(original);
+  /** @type {<K extends keyof O>(entry: [K, O[K]]) => [K, R]} */
+  const mapEntry = ([k, v]) => [k, mapFn(v, k)];
+  const newEntries = typedMap(oldEntries, mapEntry);
+  const newObj = fromTypedEntries(newEntries);
+  return /** @type {any} */ (harden(newObj));
```
mhofman, line 74: Where did it stick here?

mhofman (approved):
It all looks reasonable, but tsc is often unreasonable so without tests I'm not 100% sure it does work.

packages/common/object-map.js
gibson042: I'm not sure; it seems to have just gotten super confused:
```
Type '{ [Entry in [keyof O, R] as Entry[0]]: Entry[1]; }' is not assignable to type '{ [K in keyof O]: K extends string ? R : never; }'.
  Type 'keyof O' is not assignable to type '[keyof O, R]'.
    Type 'string | number | symbol' is not assignable to type '[keyof O, R]'.
      Type 'string' is not assignable to type '[keyof O, R]'.(2322)
```

mhofman: Looks like the type mapping failing ...

mhofman: Possibly https://github.com/microsoft/TypeScript/issues/49653#issuecomment-1172537589 ?

PR #2733, opened 2025-03-04
feat: Show memory usage in the Benchmark tests

muhammadahmadasifbhatti:

Closes: [Show Memory usage in the benchmark tests](https://github.com/Agoric/product-tasks/issues/250)

## Description

We need to show the `memory` usage before and after each test. Since, we are currently using binaries of the engines and not the runtimes the `process` object won't be available. So, I am creating a new script in the `package.json` (`test:node`). This will run the bundles file using node.

Also, I will be writing another function named `logMemoryUsage`. It will check if the process object is available it will call the `process.memoryUsage` function.

### Security Considerations

N/A

### Scaling Considerations

N/A

N/A

### Testing Considerations
This PR in itself is addition to our new testing tool.

N/A

### Upgrade Considerations

N/A

### DIFF
--- a/packages/benchmark/package.json
+++ b/packages/benchmark/package.json
@@ -33,7 +33,8 @@
     "prepack": "tsc --build tsconfig.build.json",
     "install-engines": "./install-engines.sh",
     "test": "./run-tests.sh",
-    "cpu-prof-test": "yarn rollup -c && node --jitless --cpu-prof dist/bundle.js"
+    "cpu-prof-test": "yarn rollup -c && node --jitless --cpu-prof dist/bundle.js",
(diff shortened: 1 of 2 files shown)


naugtur:
This is very unlikely to produce reliable measurements without interacting with GC. What is the goal here? Are you after the allocation spike or leak detection? Are you concerned about memory consumption of a specific usecase?

See
- work in progress:
https://github.com/RafaelGSS/bench-node/pull/33
- tiny and less statistically precise, but usable now:
https://github.com/naugtur/i-can-bench

- leak warning that should work in modern node.js
https://github.com/naugtur/memory-leak-warning

PR #2737, opened 2025-03-11
feat(ci): Divide lint/docs and add linear-history and no-fixups

kriskowal:
These are CI workflow improvements.

I’ve copied no-fixups and linear-history from Agoric SDK, but removed their predicates on Mergify labels. We will simply see failures for these jobs for any change that is under review, so CI may have a couple failed jobs for otherwise ready changes.

I’ve also divided the lint job into separate lint and doc generation jobs in hopes that this takes some time off of CI. We should consolidate type checking so we do not duplicate effort for every package, but that remains a separate ambition.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -398,3 +398,76 @@ jobs:
         run: |
           PATH=$PATH:$GITHUB_WORKSPACE/bin
           yarn test:xs
+
+  linear-history:
+    runs-on: ubuntu-latest
+    if: >-
+      github.event_name == 'pull_request' &&
+      github.event.pull_request.draft == false &&
+      github.event.pull_request.base.ref == 'master' &&
+      !contains(github.event.pull_request.labels.*.name, 'bypass:linear-history')
+    steps:
+      - uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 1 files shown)


mhofman (commented):
The problem of removing the mergify label conditions is that these tests will become very noisy for PR authors (they'll get an email for CI test failure every time they push to a PR failing these checks). I'd recommend at least matching an extra condition of having an approval, but that makes it more complicated (see below).

Which brings me to the second issue: reusing the existing `ci.yml` workflow, and its triggers. The conditions for the new CI checks still rely on labels which can technically change anytime, not on pull request synchronization. And since I'm advocating a new condition on approval, we need a trigger for that too. I would recommend moving these tests to their own workflow file, with extra `pull-request` trigger types for label conditions (see [agoric-sdk triggers](https://github.com/Agoric/agoric-sdk/blob/8dee73d5b6da4a9671f91427f080d331df0d21e6/.github/workflows/mergify-ready.yml#L4-L14)), and look into also using a [`pull_request_review`](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_review) trigger?

PR #961, opened 2021-12-18
feat: add evalTaming option

Jack-Works:
This PR adds a new option `evalTaming` which defaults to `"safeEval"` (current behavior).

I introduced a new behavior `"unsafeEval"` and `"noEval"`.

This option is for the environment that does not allow `eval` in the production mode but allows it in the development mode for webpack `eval-source-map`.
The current behavior will replace the `eval` constructor and cause the webpack bundle to fail to load.

I have documented this new option and warned any users do not use it unless they have the same situation as us.

PR #2739, opened 2025-03-12
fix(ses): lockdown options should be kebob-case

erights:
Closes: #XXXX
Refs: #961 #2690 #2723

## Description

#961 deviated from our general convention that lockdown option values be kebob-case, instead adding `evalTaming:` option values `safeEval`, `unsafeEval`, `noEval`. (I approved #961 at the time, apparently without noticing this discrepancy.) This PR fixes those to be `safe-eval`, `unsafe-eval`, and `no-eval`. But to avoid breaking any old usage, this PR ensure the only names continue to work for now, but always commented in the code as "deprecated".

This PR does only that. Other changes to the relevant lockdown option or relevant lockdown options machinery are left to #2723 or #2690 respectively. I request that this PR go first, with those others adjusting to this one.

### Security Considerations

none
### Scaling Considerations

non

This PR simply changes the documentation to use the new names without ever mentioning the deprecated old names. That seems like an appropriate simplification for the docs.

### Testing Considerations

With a bit of duplication and renaming, we now test the new names and the old deprecated names.

To avoid breaking any old usage, this PR ensure the only names continue to work for now, but always commented in the code as "deprecated". It would be very nice to eventually be able to retire the deprecated names, but I find it hard to imagine how we'd test that it is safe to do so.

### Upgrade Considerations

Nothing BREAKING, since the old deprecated names continue to work.

### DIFF
--- a/packages/ses/src/eval-scope.js
+++ b/packages/ses/src/eval-scope.js
@@ -27,9 +27,9 @@ const { Fail } = assert;
 //   prepareStack(depth, () => {
 //     (eval)('');
 //   });
-//   const unsafeEval = (eval);
-//   const safeEval = (eval);
-//   const realGlobal = unsafeEval('globalThis');
+//   const unsafe-eval = (eval);
+//   const safe-eval = (eval);
+//   const realGlobal = unsafe-eval('globalThis');
 //
 // To protect against that case, we also delete `eval` from the `evalScope` in
 // a `finally` block surrounding the call to the safe evaluator.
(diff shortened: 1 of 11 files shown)


packages/ses/src/eval-scope.js
```diff
@@ -27,9 +27,9 @@ const { Fail } = assert;
 //   prepareStack(depth, () => {
 //     (eval)('');
 //   });
-//   const unsafeEval = (eval);
-//   const safeEval = (eval);
-//   const realGlobal = unsafeEval('globalThis');
+//   const unsafe-eval = (eval);
+//   const safe-eval = (eval);
+//   const realGlobal = unsafe-eval('globalThis');
```
mhofman, line 32: Overzealous replace

mhofman (approved):
Besides one extra rename, looks fine

PR #2743, opened 2025-03-19
fix(ses): Limit scope proxy exposure to discernably owned properties …

### LINKED ISSUE #1305: Revisit behavior of `has` proxy trap in safe evaluator’s terminal scope
The `has` trap in the safe evaluator’s terminal scope proxy (before and after #1293) leaks the presence or absence of properties on the realm’s intrinsic global object. The current behavior might be the optimal compromise, but we should revisit the possibility of making `has` always return `true` to fully blot out the parent scope. @kumavis’s preliminary investigation suggests that this would break too much existing usage. We should isolate the cases that this change would break and evaluate next steps.

kriskowal:
…of host globalThis

Closes #1305

## Description

This change circles back to an old observation about the scope proxy, where we deliberately allow guest code to sense which properties of the host globalThis exist in order to provide a higher fidelity behavior for `typeof` style feature detection. This change reflects a shift in our preference to avoid this information leak at the expense of ecosystem compatibility. We have come to believe closing the leak is worth the expense, especially given that the work-around of checking `globalThis` properties explicitly has become a norm.

### Security Considerations

This change increases confinement in exchange for a minor loss of ecosystem compatibility.

### Scaling Considerations

None.

Forthcoming.

### Testing Considerations

This change will be validated with an integration PR with Agoric SDK to expose any extant compatibility risks.

This change may uncover existing code that relied on `typeof` for feature detection, which is not strictly compatible. We are treating these considerations as bug fixes.

### Upgrade Considerations

These changes will become observable to vats upon restart after this software arrives in an Agoric chain software update.

### DIFF
--- a/packages/ses/test/make-safe-evaluator.test.js
+++ b/packages/ses/test/make-safe-evaluator.test.js
@@ -4,17 +4,29 @@ import test from 'ava';
 import { makeSafeEvaluator } from '../src/make-safe-evaluator.js';
 
 test('safeEvaluate - default (non-sloppy, no moduleLexicals)', t => {
-  t.plan(6);
+  t.plan(7);
 
   const globalObject = { abc: 123 };
   const { safeEvaluate: evaluate } = makeSafeEvaluator({ globalObject });
 
   t.is(evaluate('typeof def'), 'undefined', 'typeof non declared global');
 
-  t.throws(
-    () => evaluate('def'),
-    { instanceOf: ReferenceError },
-    'non declared global cause a reference error',
+  // Expected to be undefined in sloppy mode.
+  t.is(
+    evaluate(
+      'def',
+      undefined,
+      'non-declared global is undefined in sloppy mode',
+    ),
+  );
+  // Known deviation from fidelity to Hardened JavaScript:
+  // In strict mode, an undeclared lexical name should produce a ReferenceError.
+  t.is(
+    evaluate(
+      '(() => { "use strict"; return def; })()',
+      undefined,
+      'non-declared global is undefined in strict mode',
+    ),
   );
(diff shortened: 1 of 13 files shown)


kriskowal:
Integration with Agoric SDK https://github.com/Agoric/agoric-sdk/pull/11133

erights (approved):
LGTM

packages/ses/test/strict-scope-terminator.test.js
```diff
@@ -1,13 +1,13 @@
 import test from 'ava';
 import { strictScopeTerminator } from '../src/strict-scope-terminator.js';
 
-test('strictScopeTerminator/get - always has start compartment properties but they are undefined', t => {
+test('strictScopeTerminator/get - does not have start compartment properties and they are undefined', t => {
```
mhofman, line 4:
```suggestion
test('strictScopeTerminator/get - has all properties and they are undefined', t => {
```

packages/ses/test/make-safe-evaluator.test.js
```diff
@@ -4,17 +4,29 @@ import test from 'ava';
 import { makeSafeEvaluator } from '../src/make-safe-evaluator.js';
 
 test('safeEvaluate - default (non-sloppy, no moduleLexicals)', t => {
-  t.plan(6);
+  //t.plan(6);
```
mhofman, line 7:
```suggestion
  // t.plan(6);
```

Or restore assertion with corrected value?

mhofman (approved):
LGTM once the 2 nits are addressed.

Snapp949 (commented):
Ty

PR #2744, opened 2025-03-20
chore: Release 2025-03-19

kriskowal:
# `ses` v1.12.0

- The `evalTaming:` option values are renamed:

  - from `'safeEval'`, `'unsafeEval'`, and `'noEval'`
  - to `'safe-eval'`, `'unsafe-eval'`, and `'no-eval'`

  in order to follow the convention that lockdown option values use kebob-case rather than camelCase. To avoid breaking old programs during the transition, the old names are deprecated, but continue to work for now.

- The value of expressions like `typeof unlikelyGlobal` is now `undefined` instead of producing a `ReferenceError` because it proves impossible to do so without revealing what properties exist on the host `globalThis` to compartmentalized code.

# `@endo/patterns` v1.5.0

- New pattern: `M.containerHas(elementPatt, bound = 1n)` motivated to support want patterns in Zoe, to pull out only `bound` number of elements that match `elementPatt`. `bound` must be a positive bigint.

- Closely related, `@endo/patterns` now exports `containerHasSplit` to support ERTP's use of `M.containerHas` on non-fungible (`set`, `copySet`) and semifungible (`copyBag`) assets, respectively. See https://github.com/Agoric/agoric-sdk/pull/10952 .

# `@endo/import-bundle` v1.4.0

- Adds support for `test` format bundles, which simply return a promise for an object that resembles a module exports namespace with the objects specified on the symbol-named property @exports, which is deliberately not JSON serializable or passable.
- Adds a `typedImportBundle<ExpectedExportsNamespace>` function with a proper type signature, to provide a narrower signature than `any` without disrupting existing usage.

# `@endo/bundle-source` v4.0.0

- Replaces the implementation for the `nestedEvaluate` and `getExport` formats with one based on Endo's Compartment Mapper instead of Rollup, in order to obviate the need to reconcile source map transforms between Rollup and the underlying Babel generator.  As a consequence, we no longer generate a source map for the bundle, but Babel ensures that we preserve line and column numbers between the original source and the bundled source.

# `@endo/compartment-mapper` v1.6.0

- Accommodates CommonJS modules that use `defineProperty` on `exports`.

- Divides the role of `makeBundle` into `makeScript` and `makeFunctor`.  The new `makeScript` replaces `makeBundle` without breaking changes, producing a JavaScript string that is suitable as a `<script>` tag in a web page.
- The new `makeFunctor` produces a JavaScript string that, when evaluated, produces a partially applied function, so the caller can provide runtime options.
- Both `makeScript` and `makeFunctor` now accept `format`, `useEvaluate` and `sourceUrlPrefix` options.
- The functor produced by `makeFunctor` now accepts `evaluate`, `require`, and `sourceUrlPrefix` runtime options.
- Both `makeScript` and `makeFunctor` now accept a `format` option.  Specifiying the `"cjs"` format allows the bundle to exit to the host's CommonJS `require` for host modules.
- Adds `sourceDirname` to compartment descriptors in the compartment maps generated by `mapNodeModules` and uses these to provide better source URL comments for bundles generated by `makeScript` and `makeFunctor`, by default.

These changes collectively allow us to replace the implementation of `nestedEvaluate` and `getExports` formats in `@endo/bundle-source`, including the preservation of useful line numbers and file names in stack traces.

- `mapNodeModules`, `importLocation` and `loadLocation` now accept a `log` option for users to define a custom logging function. As of this writing, _only `mapNodeModules`_ will potentially call this function if provided.  Expansion of log messaging and support for the `log` option in more APIs is expected _in the future_.

# `@endo/evasive-transform` v1.4.0

- Adds a `sourceMap` option so that the generated sourcemap can project back to the original source code without `unmapLoc`.
- Removes support for sourcemap `unmapLoc` because it is not used by contemporary Endo packages.  The option is now ignored.

### DIFF
--- a/packages/ses/NEWS.md
+++ b/packages/ses/NEWS.md
@@ -1,14 +1,23 @@
 User-visible changes in `ses`:
 
-# Next release
+# v1.12.0 (2025-03-11)
 
-The `evalTaming:` option values are renamed
-- from `'safeEval'`, `'unsafeEval'`, and `'noEval'`
-- to `'safe-eval'`, `'unsafe-eval'`, and `'no-eval'`
+- The `evalTaming:` option values are renamed:
 
-in order to follow the convention that lockdown option values use kebob-case
-rather than camelCase. To avoid breaking old programs during the transition,
-the old names are deprecated, but continue to work for now.
+  - from `'safeEval'`, `'unsafeEval'`, and `'noEval'`
+  - to `'safe-eval'`, `'unsafe-eval'`, and `'no-eval'`
+
+  in order to follow the convention that lockdown option values use kebob-case
+  rather than camelCase. To avoid breaking old programs during the transition,
+  the old names are deprecated, but continue to work for now.
+
+- Evaluating a non-lexical name that is also absent on the global object of a
+  compartment no longer throws a `ReferenceError` and instead produces
+  `undefined` because it proves impossible to do so without revealing what
+  properties exist on the host `globalThis` to compartmentalized code with a
+  shim.
+  This is a divergence from the expected behavior of a native Hardened
+  JavaScript implementation, like XS.
 
 # v1.11.0 (2025-01-23)
 
--- a/packages/benchmark/CHANGELOG.md
+++ b/packages/benchmark/CHANGELOG.md
@@ -1,0 +1,13 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
+See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
+
+### 0.1.1 (2025-03-24)
+
+
+### Bug Fixes
+
+* add echo to run-test.sh ([0f9bd62](https://github.com/endojs/endo/commit/0f9bd62f853758ea70d11e67ec333924d2c7da3a))
+* **camelCase:** Change  to ([b5a9bf0](https://github.com/endojs/endo/commit/b5a9bf0944c6d96f630e7a1dd5c83f40d371a988))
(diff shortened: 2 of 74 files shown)


packages/ses/NEWS.md
```diff
@@ -1,14 +1,20 @@
...
+  - to `'safe-eval'`, `'unsafe-eval'`, and `'no-eval'`
+
+  in order to follow the convention that lockdown option values use kebob-case
+  rather than camelCase. To avoid breaking old programs during the transition,
+  the old names are deprecated, but continue to work for now.
+
+- The value of expressions like `typeof unlikelyGlobal` is now `undefined`
+  instead of producing a `ReferenceError` because it proves impossible to
+  do so without revealing what properties exist on the host `globalThis`
+  to compartmentalized code.
```
mhofman, line 17: `typeof unlikelyGlobal` was already undefined before.

it's more about the case `try { unlikelyGlobal; } catch(e) { console.error(e); }` that would no longer print an error.

mhofman (approved):
The News for SES needs some tweaking. Besides that, looks mechanical.

mhofman:
We may have an [integration failure with agoric-sdk](https://github.com/Agoric/agoric-sdk/actions/runs/13963271849/job/39088378624#step:10:7905) due to unsupported syntax by esm.

```
SyntaxError#2: Invalid or unexpected token
          inResults?.push(element);
```

mhofman (commented):
lint is unhappy

Issue #2285, opened 2024-05-15
"Checker" parameters would be more ergonomic as "Rejectors"


gibson042:
## What is the Problem Being Solved?

#2284 demonstrated that there is a relatively small but still significant performance cost to creating `reject = !!check && ((T, ...subs) => check(false, X(T, ...subs)))` functions that will only be used in failure paths, and so moved that definition into a new `CX` function and updated many sites to use that:
```diff
-    reject && reject`…`
+    !!check && CX(check)`…`
```

However, this has drawbacks of its own with respect to comprehensibility and readability, especially when prettier introduces line breaks like
```js
    (
      !!check &&
      CX(
        check,
      )`…`
    )
```

## Description of the Design

I think we'd be better off replacing `(cond: boolean, details?: Details | undefined) => boolean` "[Checker](https://github.com/endojs/endo/blob/fa20921dc8bd05084258e4783f93fe016d6e162a/packages/pass-style/src/types.d.ts#L184-L203)" parameters with optional `(template: TemplateStringsArray | string[], ...args: any[]) => false` "Rejector" parameters, recovering the old local pattern but without local function creation.

And per https://github.com/endojs/endo/issues/2285#issuecomment-2115922362 , we'll also want to rename internal functions for indicating the progress of this transition.

```diff
-const checkSomething = (candidate, check = undefined) => {
+const confirmSomething = (candidate, reject = false) => {
   return (
     isOk(candidate) ||
-    (!!check && CX(check)`…`)
+    (reject && reject`…`)
   );
 };
```

And at the highest possible level, we'd replace use of `x => x` checkers with `undefined` rejectors and [`assertChecker`](https://github.com/endojs/endo/blob/5bdb77d2689fef284475176f52e81b166ad788f7/packages/pass-style/src/passStyle-helpers.js#L76-L79) with `assert.Fail`.
```diff
-    checkSomething(candidate, assertChecker)
+    confirmSomething(candidate, Fail)
```

## Security Considerations

None that I can see.

## Scaling Considerations

Negligible, but the new pattern might be _slightly_ faster.

## Test Plan

n/a, this is behavior-preserving refactoring.

## Compatibility Considerations

`assertChecker` is exported by the pass-style package, and used by at least the patterns package, which also makes use of checkers so should be updated similarly. But I *think* all such use is internal for supporting {is,assert}$Type package exports (e.g., `isKey` and `assertKey`).

## Upgrade Considerations

[...]

PR #2746, opened 2025-03-22
refactor(pass-style): minor types gardening

erights:
Closes: #XXXX
Refs: #XXXX

## Description

`@endo/pass-style` was redundantly defining and exporting its own `Checker` type. This PR changes it to simply re-export the `Checker` definition `'endo/common/ident-checker.js'`. This is a pure refactor with no observable effect.

### Security Considerations

none
### Scaling Considerations

none

none
### Testing Considerations

includes a type test

none
### Upgrade Considerations

none


packages/pass-style/src/types.d.ts
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable no-use-before-define */
+import type { Checker as _Checker } from '@endo/common/ident-checker.js';
```
erights, line 2: Is this, together with line 208 below `export type Checker = _Checker;`, really the best way in type script to import-and-reexport a type from elsewhere?

kriskowal: I defer to @turadg. I don’t recall whether this was a specific limitation for ambient types.

kriskowal (changes requested):
Good enough in my book, but it would be lovely to have a `types.test-d.ts` to validate these type signatures get exported as intended and so we don’t regress. There are some examples of `test-d.ts` elsewhere in the repo.

packages/pass-style/src/types.d.ts
turadg: There's a simpler way. I just pushed it.

turadg (approved):

gibson042:
Possibly relevant: https://github.com/endojs/endo/pull/2851#issuecomment-2960167182 (cf. #2285)

packages/pass-style/src/types.d.ts
erights: Thanks @turadg , resolving

erights:
> Good enough in my book, but it would be lovely to have a `types.test-d.ts` to validate these type signatures get exported as intended and so we don’t regress. There are some examples of `test-d.ts` elsewhere in the repo.

Done, thanks to @turadg . PTAL

kriskowal (approved):

PR #2751, opened 2025-03-29
wip(syrup): syrup reader, writer, and codecs for efficient parsing

kumavis:
replaced by https://github.com/endojs/endo/pull/2766

wip improvement to syrup -- working towards allowing a future consumer (eg OCapN) to lazily parse syrup messages without reifying all encountered values

see test file https://github.com/endojs/endo/blob/7632803791dc6160c39c1af33963f52e0f2730ec/packages/syrup/test/parse.test.js


packages/syrup/src/symbol.js
```diff
@@ -0,0 +1,3 @@
+// To be used as keys, syrup symbols must be javascript symbols.
+// To avoid an otherwise meaningful symbol name, we prefix it with 'syrup:'.
+export const SyrupSymbolFor = (name) => Symbol.for(`syrup:${name}`);
```
kriskowal, line 3: This unfortunately registers every symbol encountered in a global table that isn’t necessarily garbage collected.

My intuition is that we’ll need to either wrap _every_ value in a wrapper object like `{type: 'symbol', symbol: 'hello'}` or just wrap all non-primitive values, including `list`, `struct`, and `record`, like `{type: 'record', record: [1, 2, 3]}` and `{type: 'list', list: [1, 2, 3]}`.

kriskowal: This is also why we’re now calling Syrup symbols “Selector”. They do not necessarily correspond to implementation-language symbols.

kumavis: i made them prefixed symbols so that they can be keys on dictionaries (represented as js objects) though I don't create them when they are Record labels, as used in ocapn messages.

could create a custom clearable symbol registry though `Symbol('hi') !== Symbol.for('hi')`

as well as other options

kumavis:
separate branch with `symbol` -> `selector` rename https://github.com/endojs/endo/compare/kumavis-syrup-lazy-parse...kumavis-syrup-lazy-parse-selector?expand=1

kumavis:
replaced by https://github.com/endojs/endo/pull/2766

PR #2753, opened 2025-04-02
feat(bundle-source,import-bundle): Thread importHook option to endoZipBase64 moduleFormat.

grypez:
## Description

This PR allows `bundleSource` and `importBundle` to bundle files and import bundled files respectively in the case that
1. the source file depends on external dependencies (such as 'fs' and 'path') and
2. the bundle is created using the endoZipBase64 moduleFormat option

This functionality already existed in `@endo/compartment-mapper`, and this PR simply plumbs the importHook option to the bundleSource and importBundle APIs.

### Security Considerations

Any designs which depend upon code bundled with the endoZipBase64 moduleFormat to have no external dependencies will be broken. To my knowledge none exist.

### Scaling Considerations

None

The bundleSource documentation does not have a use case showing the importHook option.

### Testing Considerations

See included tests.

None

### Upgrade Considerations

Bundling with endoZipBase64 is now more permissive, so existing importBundle calls may fail in a way they couldn't before.

### DIFF
--- a/packages/import-bundle/src/index.js
+++ b/packages/import-bundle/src/index.js
@@ -32,6 +32,7 @@ export async function importBundle(bundle, options = {}, powers = {}) {
     inescapableTransforms = [],
     inescapableGlobalProperties = {},
     expectedSha512 = undefined,
+    importHook = undefined,
   } = options;
   const {
     computeSha512 = undefined,
@@ -74,6 +75,7 @@ export async function importBundle(bundle, options = {}, powers = {}) {
       expectedSha512,
       computeSourceLocation,
       computeSourceMapLocation,
+      importHook,
     });
     // Call import by property to bypass SES censoring for dynamic import.
     // eslint-disable-next-line dot-notation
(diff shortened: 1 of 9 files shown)


packages/import-bundle/src/index.js
```diff
@@ -32,6 +32,7 @@ export async function importBundle(bundle, options = {}, powers = {}) {
     inescapableTransforms = [],
     inescapableGlobalProperties = {},
     expectedSha512 = undefined,
+    importHook = () => null,
```
kriskowal, line 35: Should be okay to simply omit, and obviate a closure allocation when it’s unused.
```suggestion
    importHook = undefined,
```

packages/import-bundle/test/import-bundle.test.js
```diff
@@ -19,13 +19,13 @@ function transform1(src) {
     .replace('two foot wide', 'sixty feet wide');
 }
 
-async function testBundle1(t, b1, mode, ew) {
-  const ns1 = await importBundle(b1, { endowments: ew });
+async function testBundle1(t, b1, mode, ew, opt = {}) {
```
kriskowal, line 22: Style: please avoid abbreviation.
```suggestion
async function testBundle1(t, b1, mode, ew, options = {}) {
```

If you find yourself doing an initialism or an acronym, treat it as a word. We say `XmlHttpRequest` ’round these here parts.

kriskowal (approved):
Thank you, style nits only.

Issue #2754, opened 2025-04-02
compartment-mapper: dynamic require of "node:"-prefixed builtins fails


boneskull:
## Describe the bug

While I was careful to add a test case for "dynamic require of a builtin module", that test case only covers a bare module specifier (`cluster`).

Changing `cluster` to `node:cluster` fails the test:

```
  Rejected promise returned by test. Reason:

  TypeError {
    message: 'Failed to load module "node:cluster" in package "file:///path/to/compartment-mapper/test/fixtures-dynamic/node_modules/hooked-app/" (1 underlying failures: The URL must be of scheme file',
  }

  TypeError: Failed to load module "node:cluster" in package "file:///path/to/compartment-mapper/test/fixtures-dynamic/node_modules/hooked-app/" (1 underlying failures: The URL must be of scheme file
      at throwAggregateError (file:///path/to/ses/src/module-load.js:557:11)
      at loadNow (file:///path/to/ses/src/module-load.js:654:3)
      at Compartment.importNow (file:///path/to/ses/src/compartment.js:187:5)
      at require (file:///path/to/compartment-mapper/src/parse-cjs-shared-export-wrapper.js:149:33)
      at Proxy.eval (eval at <anonymous> (eval at makeEvaluate (file:///path/to/ses/src/make-evaluate.js:92:27)), <anonymous>:5:1)
      at Object.execute (file:///path/to/compartment-mapper/src/parse-cjs.js:56:13)
      at execute (file:///path/to/ses/src/module-instance.js:101:24)
      at compartmentImportNow (file:///path/to/ses/src/compartment.js:103:3)
      at file:///path/to/ses/src/compartment.js:160:27
      at async file:///path/to/compartment-mapper/test/dynamic-require.test.js:303:25
```

This is reproducible outside of the `@endo/compartment-mapper`'s test environment (I found it via `@lavamoat/node`'s testing environment; lol).

## Steps to reproduce

1. Modify `compartment-mapper/test/fixtures-dynamic/node_modules/hooked-app/index.js`:

    ```diff
    diff --git a/packages/compartment-mapper/test/fixtures-dynamic/node_modules/hooked-app/index.js b/packages/compartment-mapper/test/fixtures-dynamic/node_modules/hooked-app/index.js
    index f64a83d77acdc2d1b5b5b9ffe254108d2c0fa567..972b76820199e55ef0c0624417708c4a432a9cb4 100644
[...]

PR #2755, opened 2025-04-03
fix dynamic require of node:-namespaced builtins

boneskull:
Closes: #2754

## Description

This fixes a couple of bugs, actually:

- `inferExports` no longer throws when encountering the plainly-invalid-but-first-discovered-in-the-wild export alias of `./`. This alias is now ignored.
- `importNowHook` now handles dynamically-required builtins without considering them to first be packages, which a) is faster and b) avoids a potential thrown exception when a `FileUrlToPathFn` encounters a string that appears to have a scheme. (like `node:fs`). This was the root cause of the issue in #2754.

And some tangents:

- While fixing my `inferExports` issue, I refactored type signatures to use `PackageDescriptor` instead of `object`. This required defining a few more fields in `PackageDescriptor` which `infer-exports.js` considers.
- `WILDCARD_POLICY_VALUE` is now interally-exported from `policy-format.js` which should probably be used instead of `'any'` in the context of policy (wherever possible). It's now used in the test suite for dynamic requires.

And with shame:

- I had refactored the type signature for `mapParsers` to have the signature that I had originally intended. I and didn't feel like it deserved an entire PR. _This is completely unrelated to anything else here_

### Security Considerations

n/a

### Scaling Considerations

It should be faster for `@endo/compartment-mapper` to crunch dynamically-required builtins.

I don't think this needs any documentation changes.

### Testing Considerations

I've added coverage for #2754, but did not add coverage for anything else. I should probably add coverage for the change to `inferExports`.

Backwards compatible


packages/compartment-mapper/src/import-hook.js
```diff
@@ -695,6 +696,40 @@ export function makeImportNowHookMaker(
...
+            packageLocation,
+          );
+          // It'd be nice to check the policy before importing it, but we can only throw a policy error if the
+          // hook returns something. Otherwise, we need to fall back to the 'cannot find' error below.
+          enforceModulePolicy(moduleSpecifier, compartmentDescriptor, {
+            exit: true,
+            errorHint: `Blocked in loading. ${q(
+              moduleSpecifier,
+            )} was not in the compartment map and an attempt was made to load it as a builtin`,
+          });
```
boneskull, line 715: - [ ] wrap in `if (exitRecord)`

packages/compartment-mapper/src/import-hook.js
```diff
@@ -695,6 +696,40 @@ export function makeImportNowHookMaker(
...
+
+          // should this happen prior to policy enforcement?
+          // (and also in importHook?)
+          if (!exitRecord) {
+            throw new Error(
+              `Could not import external module: ${q(moduleSpecifier)}`,
+            );
+          }
+
```
boneskull, line 724: - [ ] remove

packages/compartment-mapper/src/infer-exports.js
```diff
@@ -81,7 +84,9 @@ function* interpretExports(name, exports, conditions) {
     );
   }
   for (const [key, value] of entries(exports)) {
-    if (key.startsWith('./') || key === '.') {
+    // "./" is explicitly an invalid key, but that doesn't
+    // stop people from trying to use it.
+    if (key !== './' && (key.startsWith('./') || key === '.')) {
```
kriskowal, line 89: I think this might be more true. `./` is also presumably an invalid name for a condition.
```suggestion
        if (key === './') {
           continue; // or no-op
        } else if (key.startsWith('./') || key === '.') {
```

packages/compartment-mapper/test/fixtures-dynamic/node_modules/hooked-app-namespaced/index.js
```diff
@@ -0,0 +1,5 @@
+exports.isOk = require('hooked').isOk;
+
+const builtin = 'node:cluster';
+
+require(builtin);
```
kriskowal, line 5: Missing newline. Windows?

kriskowal (approved):

PR #2762, opened 2025-04-08
feat(compartment-mapper): support treating top-level return in CJS as a way to export from module

naugtur:
## Description
One of the rarely used ways to export from a CJS module is to `return` a reference instead of setting `module.exports` to that reference. This PR adds support for that behavior and a tiny fixture in tests to ensure it works in all scenarios.

### Security Considerations

N/A

### Scaling Considerations

N/A

It's part of CJS and also something very few people know. Documenting it would lead to more confusion than not.

### Testing Considerations

Covered with a case in existing test coverage.

Increases ecosystem compatibility by the few unfortunate packages that use this way of exporting.

### Upgrade Considerations

None


packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
```diff
@@ -184,8 +184,9 @@ export const wrap = ({
 
   freeze(require);
 
-  const afterExecute = () => {
-    const finalExports = module.exports; // in case it's a getter, only call it once
+  const afterExecute = returnValue => {
+    const finalExports =
+      returnValue !== undefined ? returnValue : module.exports; // in case it's a getter, only call it once
```
boneskull, line 189: Wait. Is this right? 

Executing `node -e console.log(require("/path/to/exports-return.js"))` reports `{}` regardless of what was returned. Change it to anything else; it's always `{}`.

It appears that we need to assert `exports-return.js` returns an empty object. So... not `{magicNumber: 42}` at all.

It looks as if the `originalExports` value comes from the AST?  But the parsing cannot determine whether or not a `return` comes _before_ or _after_ an assignment to `exports` except in the most useless of trivial cases. That `return` can be anywhere in module scope.

Then, it seems like we _cannot statically determine_ the value of `module.exports` because it should always be treated as dynamic.

Maybe what needs to happen is:

1. Ignore whatever `originalExports` is, assuming we don't need it for something else
2. Ignore the return value of `functor`
3. Remove the conditional `exportsHaveBeenOverwritten` and just assume they have.

cc @kriskowal who will likely find this abhorrent

boneskull (changes requested):
I think this is busted 😄

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
boneskull: Might need to tweak the parser to flag a module's exports as "lol idk"

boneskull:
We will also need to enable the `allowReturnOutsideFunction` option for calls to `@babel/parser` (if `sourceType` is `script`).  This change would need to happen both in `@endo/module-source` _and_ `@endo/evasive-transform`.

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
naugtur: I implemented this because bundlers had it claiming it to be node.js compatibility AFAIR. Now I see Node doesn't have that behavior and am confused. Will report back

PR #2763, opened 2025-04-09
feat(evasive-transform)!: allow returns outside functions in CommonJS sources

boneskull:
## Description

`return` statements outside of functions are valid in CommonJS (sloppy-mode only? not sure), but Babel doesn't allow this by default. Unfortunately, _people do this_ and We need to support it.

This behavior is enabled by passing a flag into Babel when parsing. This flag is _only_ enabled when the `sourceType` option (for `evadeCensor` and `evadeCensorSycn`) is set to `script`.

To discourage use of non-portable syntax, we now restrict the possible values of `sourceType` to `script` and `module` only (which is the **BREAKING CHANGE**).

The types were changed to reflect this. While updating the types, I noted that they were _wrong_, so I fixed them using overloads. Note that `@endo/evasive-transform` does not ship declarations (yet).

### Security Considerations

No

### Scaling Considerations

No

I have updated `README.md` with a little bit of information, but there are no such comprehensive API docs for this package.

### Testing Considerations

I've added tests to cover the new behavior.

If someone is using this package and supplying a `sourceType` option of `unambiguous` (which I _believe_ is Babel's default behavior), they will get a _TypeScript error_. We will not throw an exception, since this is a valid option for Babel, after all—but we do not claim to support it. This will not affect any existing consumers using `sourceType` of `script`, because if they were parsing sources that had bare `return` statements, it would have broken already.

The new (correct) return types of the public APIs _may_ cause TypeScript type errors if they were manually fixing the types themselves.

### Upgrade Considerations

If using an explicit `sourceType` of `unambiguous`, omit it. It is encouraged to use `module` or `script` if possible.


boneskull:
I'm calling it breaking (and I've covered my bases) but others may disagree.

kriskowal (approved):

packages/evasive-transform/src/parse-ast.js
```diff
@@ -38,6 +36,7 @@ export function parseAst(source, opts = {}) {
   return parseBabel(source, {
     tokens: true,
     createParenthesizedExpressions: true,
+    allowReturnOutsideFunction: opts.sourceType === 'script',
```
naugtur, line 39:
```suggestion
    allowReturnOutsideFunction: opts.sourceType !== 'module',
```
^ this would cover cases of sourceType not being set/known

boneskull: paraphrasing @kriskowal, it is only for `script` because we do not want to be toxic enablers of non-portable syntax

PR #2765, opened 2025-04-10
fix(compartment-mapper): specifier of module descriptor returned from findRedirect fixed

boneskull:
## Description

`findRedirect`, which is part of `importNowHook`, is called when we encounter a dynamic require where the specifier is an absolute path.

Previous to this change, the `RedirectStaticModuleInterface` (a `ModuleDescriptor`) could contain an absolute path. Absolute paths are not generally assumed by `ses` or `@endo/compartment-mapper`. In the situation where the module referenced in the `RedirectStaticModuleInterface` itself imports _another_ package (via mutual recursion & trampolining), these assumptions were violated and we made an attempt to call `resolve()` (from the `node-module-specifier` module) using its absolute path as the referrer, which is explicitly disallowed by `resolve()`; it throws an exception.

To that end, we need to ensure that whatever `ModuleDescriptor` we return from `importNowHook` with is one where its `specifier` is relative to its package location.

The fact that this nominally "worked" against our tests in both `@endo/compartment-mapper` _and_ `@lavamoat/node` was due to a) chance, and b) lack of a test case against a fixture representing this configuration.

In addition, I removed the `packageLocation` parameter from `findRedirect`'s options bucket since it was serving no practical purpose (as noticed by @naugtur).

### Security Considerations

None that I am aware of

### Scaling Considerations

no

This is a bugfix

### Testing Considerations

I need to add two more tests against fixtures before I publish this:

1. [x] A fixture which is similar to the configuration mentioned above
2. [x] A fixture where the dynamically-required file is in a different compartment (which should be subject to policy enforcement)

None that I am aware of

### Upgrade Considerations

Not breaking. Probably not worth a note in `NEWS.md`

### DIFF
--- a/packages/compartment-mapper/src/import-hook.js
+++ b/packages/compartment-mapper/src/import-hook.js
@@ -101,7 +101,9 @@ const nodejsConventionSearchSuffixes = [
  * @returns {string} Redirect static module interface; the `specifier` prop of this value _must_ be a relative path with leading `./`
  */
 const relativeSpecifier = (moduleSpecifierLocation, location) => {
-  return `./${moduleSpecifierLocation.replace(location, '')}`;
+  assert(moduleSpecifierLocation.startsWith(location), 
+    `Module specifier location "${moduleSpecifierLocation}" does not start with compartment location "${location}"`);
+  return `./${moduleSpecifierLocation.substring(location.length)}`;
 };
 
 /**
--- a/packages/compartment-mapper/src/import-hook.js
+++ b/packages/compartment-mapper/src/import-hook.js
@@ -15,7 +15,8 @@
  *   ImportHook,
  *   ImportNowHook,
  *   RedirectStaticModuleInterface,
- *   StaticModuleType
+ *   StaticModuleType,
+ VirtualModuleSource
  * } from 'ses'
  * @import {
  *   CompartmentDescriptor,
@@ -702,19 +703,13 @@ export function makeImportNowHookMaker(
 
     /** @type {ImportNowHook} */
     const importNowHook = moduleSpecifier => {
-      // many dynamically-required specifiers will be absolute paths owing to use of `require.resolve()` and `path.resolve()`
-      if (isAbsolute(moduleSpecifier)) {
-        const record = findRedirect({
-          compartmentDescriptor,
-          compartmentDescriptors,
-          compartments,
-          absoluteModuleSpecifier: moduleSpecifier,
-        });
-        if (record) {
-          return record;
-        }
-        // if there is no record found this way, we will search for it instead of considering it to be an exit module
-      } else if (moduleSpecifier !== '.' && !moduleSpecifier.startsWith('./')) {
+      /**
+       * Attempt to load the moduleSpecifier via an {@link ExitModuleImportNowHook}, if one exists.
+       *
+       * If it doesn't exist, then throw an exception.
+       * @returns {VirtualModuleSource}
+       */
+      const tryExitModuleImportNowHook = () => {
         if (exitModuleImportNowHook) {
           // This hook is responsible for ensuring that the moduleSpecifier
           // actually refers to an exit module.
@@ -733,14 +728,53 @@ export function makeImportNowHookMaker(
             });
             return exitRecord;
           }
+          throw Error(
+            `Cannot find external module ${q(
+              moduleSpecifier,
+            )} in package at ${packageLocation}`,
+          );
+        } else {
+          throw Error(
+            `Cannot find external module ${q(
+              moduleSpecifier,
+            )} from package at ${packageLocation}; try providing an importNowHook`,
+          );
         }
-        throw Error(
-          `Cannot find external module ${q(
+      };
+
+      // many dynamically-required specifiers will be absolute paths owing to use of `require.resolve()` and `path.resolve()`
+      if (isAbsolute(moduleSpecifier)) {
+        const record = findRedirect({
+          compartmentDescriptor,
+          compartmentDescriptors,
+          compartments,
+          absoluteModuleSpecifier: moduleSpecifier,
+        });
+        if (record) {
+          return record;
+        }
+
+        // if and only if the module specifier is within the compartment can we
+        // make it a relative specifier. the following conditional avoids a try/catch
+        // since `relativeSpecifier` will throw if this condition is not met
+        if (
+          isLocationWithinCompartment(
             moduleSpecifier,
-          )} in package ${packageLocation}`,
-        );
+            compartmentDescriptor.location,
+          )
+        ) {
+          moduleSpecifier = relativeSpecifier(
+            moduleSpecifier,
+            compartmentDescriptor.location,
+          );
+        }
+      } else if (moduleSpecifier !== '.' && !moduleSpecifier.startsWith('./')) {
+        // could be a builtin, which means we should not bother bouncing on the trampoline to find it.
+        return tryExitModuleImportNowHook();
       }
 
+      // we might have an absolute path here, but it might be within the compartment, so
+      // we will try to find it.
       const candidates = nominateCandidates(moduleSpecifier, searchSuffixes);
 
       const record = syncTrampoline(
@@ -770,11 +804,8 @@ export function makeImportNowHookMaker(
         return record;
       }
 
-      throw new Error(
-        `Could not import module: ${q(
-          moduleSpecifier,
-        )}; try providing an importNowHook`,
-      );
+      // at this point, we haven't found the module by guessing, so we'll try the importer-of-last-resort
+      return tryExitModuleImportNowHook();
     };
 
     return importNowHook;
--- a/packages/compartment-mapper/src/import-hook.js
+++ b/packages/compartment-mapper/src/import-hook.js
@@ -16,7 +16,7 @@
  *   ImportNowHook,
  *   RedirectStaticModuleInterface,
  *   StaticModuleType,
- VirtualModuleSource
+ *   VirtualModuleSource
  * } from 'ses'
  * @import {
  *   CompartmentDescriptor,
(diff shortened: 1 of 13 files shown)


packages/compartment-mapper/src/import-hook.js
```diff
@@ -108,12 +108,21 @@ const findRedirect = ({
...
+  /**
+   * Computes the relative path to a module from its compartment location (including a leading `./`)
+   *
+   * @param {string} moduleSpecifierLocation Absolute path (not URL) to a module specifier
+   * @param {string} location Compartment location
+   * @returns {string} Redirect static module interface; the `specifier` prop of this value _must_ be a relative path with leading `./`
+   */
+  const relativeSpecifier = (moduleSpecifierLocation, location) => {
+    const relativePath = moduleSpecifierLocation.replace(location, '');
+    return relativePath.startsWith('.') ? relativePath : `./${relativePath}`;
```
naugtur, line 121: how could a result of `.replace(location, '')` ever produce a leading dot?

packages/compartment-mapper/test/dynamic-require.test.js
```diff
@@ -418,3 +418,45 @@ test('sync module transforms work without dynamic require support', async t => {
 
   t.is(transformCount, 2);
 });
+
+test('dynamic require of missing module falls through to importNowHook', async t => {
```
naugtur, line 422: could you point out which test is covering the situation where findRedirect finds the redirect?

packages/compartment-mapper/src/import-hook.js
```diff
@@ -692,7 +701,6 @@ export function makeImportNowHookMaker(
           compartmentDescriptors,
           compartments,
           absoluteModuleSpecifier: moduleSpecifier,
-          packageLocation,
```
naugtur, line 695: I still believe we should not continue with an absolute path if findRedirect returns undefined. Any new findings to counter that since we talked?

packages/compartment-mapper/test/dynamic-require.test.js
boneskull: any test which contains a fixture which is absolute-requiring something is.  ... which is most of them.

packages/compartment-mapper/src/import-hook.js
boneskull: hm, right. I think I was worried about somehow getting a `../` somewhere. but I didn't think it through.

packages/compartment-mapper/src/import-hook.js
boneskull: It looks like `findRedirect` fails to find absolute paths to compartment-relative modules which were not already mentioned.  For example, if we try to `require(require.resolve('./good.js'))`—and there are no existing `ModuleDescriptor`s for a `./good.js` in the compartment descriptor—then `findRedirect` will return `undefined`. We need to let it bounce on the trampoline to find the file.  If the file still can't be found, then we fall through to the hook-of-last-resort (exit hook).

That said, maybe it would be prudent to rewrite the path as relative to the compartment's location before we continue (though it "works" as-is).

boneskull:
OK, this is ready

packages/compartment-mapper/src/import-hook.js
```diff
@@ -692,11 +700,17 @@ export function makeImportNowHookMaker(
...
         if (record) {
           return record;
         }
+
+        // make the module specifier relative to location, since we cannot be sure
+        // any of the subsequent code will accept absolute paths without protest
+        moduleSpecifier = relativeSpecifier(
+          new URL(moduleSpecifier, 'file:').href,
+          compartmentDescriptor.location,
+        );
```
naugtur, line 713: While the current implementation of relativeSpecifier makes sense if we've already found a matching compartment and just need to remove the part of the path, it doesn't make sense when we don't know whether the compartment matches.

```js

console.log(relativeSpecifier(
  new URL('/a/b/c.js', 'file:').href,
  '/something/else',
))
```
```
./file:///a/b/c.js
```

boneskull: OK, so then we need to leave as-is.

packages/compartment-mapper/test/dynamic-require.test.js
boneskull: specifically, [this test](https://github.com/endojs/endo/blob/41c4494e5db51b00606491c49d31ff216a19f1df/packages/compartment-mapper/test/dynamic-require.test.js#L276-L350) covers the same sort of configuration as we encountered:

1. `hooked-app` statically requires package `hooked`
2. `hooked` dynamically requires package `dynamic` by absolute path
3. `dynamic` dynamically requires package `is-ok` by absolute path
4. `is-ok` statically requires package `is-not-ok`

packages/compartment-mapper/src/import-hook.js
boneskull: > That said, maybe it would be prudent to rewrite the path as relative to the compartment's location before we continue (though it "works" as-is).

I did this and it was wrong as pointed out by @naugtur.  Reverted it

boneskull:
I've had to change the `importNowHook` logic somewhat considerably due to #2755.  There are now two paths to a potential call to an `ExitModuleImportNowHook`: 

1. If the module specifier was absolute _and_ not found via guesswork (`chooseModuleDescriptor()`)
2. If the module specifier is likely a builtin (or something else weird)

This preserves support for `node:`-prefixed builtins while still enabling a fallback to the `ExitModuleImportNowHook` if a module specifier w/ an absolute path could not be found otherwise.

[This test](https://github.com/endojs/endo/blob/0744b25cee9ad48e922bc3fe293f85b1b6672127/packages/compartment-mapper/test/dynamic-require.test.js#L592-L631) covers the situation where we have a dynamic require of a module requested by absolute path which is found successfully via `chooseModuleDescriptor` (`good.js`), and another which cannot be found this way (`missing.js`) because it is missing; this invokes the `ExitModuleImportNowHook`.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -15,7 +15,8 @@
  *   ImportHook,
  *   ImportNowHook,
  *   RedirectStaticModuleInterface,
- *   StaticModuleType
+ *   StaticModuleType,
+ VirtualModuleSource
```
naugtur, line 19: this looks odd. Did the IDE add it for you? Is it supposed to be here?

packages/compartment-mapper/src/import-hook.js
```diff
@@ -95,6 +93,17 @@ const nodejsConventionSearchSuffixes = [
...
+/**
+ * Computes the relative path to a module from its compartment location (including a leading `./`)
+ *
+ * @param {string} moduleSpecifierLocation Absolute path (not URL) to a module specifier
+ * @param {string} location Compartment location
+ * @returns {string} Redirect static module interface; the `specifier` prop of this value _must_ be a relative path with leading `./`
+ */
+const relativeSpecifier = (moduleSpecifierLocation, location) => {
+  return `./${moduleSpecifierLocation.replace(location, '')}`;
+};
```
naugtur, line 105: I still cringe at the possibility of returning `./${moduleSpecifierLocation}`

Here's a version that should be equaly performant and more strict.

```suggestion
const relativeSpecifier = (moduleSpecifierLocation, location) => {
  assert(moduleSpecifierLocation.startsWith(location), 
    `Module specifier location "${moduleSpecifierLocation}" does not start with compartment location "${location}"`);
  return `./${moduleSpecifierLocation.substring(location.length)}`;
};
```

packages/compartment-mapper/src/import-hook.js
```diff
@@ -686,20 +692,13 @@ export function makeImportNowHookMaker(
...
-        }
-        // if there is no record found this way, we will search for it instead of considering it to be an exit module
-      } else if (moduleSpecifier !== '.' && !moduleSpecifier.startsWith('./')) {
+      /**
+       * Attempt to load the moduleSpecifier via an {@link ExitModuleImportNowHook}, if one exists.
+       *
+       * If it doesn't exist, then throw an exception.
+       * @returns {VirtualModuleSource}
+       */
+      const tryExitModuleImportNowHook = () => {
```
naugtur, line 701: (could) I'm pretty sure there's nothing stopping us from not nesting this one within importNowHook

packages/compartment-mapper/src/import-hook.js
```diff
@@ -718,14 +717,38 @@ export function makeImportNowHookMaker(
...
+      } else if (moduleSpecifier !== '.' && !moduleSpecifier.startsWith('./')) {
+        // could be a builtin, which means we should not bother bouncing on the trampoline to find it.
+        return tryExitModuleImportNowHook();
       }
 
+      // we might have an absolute path here, but it might be within the compartment, so
+      // we will try to find it.
       const candidates = nominateCandidates(moduleSpecifier, searchSuffixes);
 
       const record = syncTrampoline(
```
naugtur, line 754: AFAIR going into `chooseModuleDescriptor` with an absolute path can result in an error that got us to find this issue in the first place. Some of the nested functionality is not ready to handle absolute paths. If the path is within the current compartment, we should be able to convert it to relative.

packages/compartment-mapper/src/import-hook.js
naugtur: (nit, not sure) I'm thinking the name could more explicitly remind us that the "cannot find" error is here
```suggestion
      const importExitModuleOrError = () => {
```

packages/compartment-mapper/src/import-hook.js
boneskull: it sure did!

packages/compartment-mapper/src/import-hook.js
boneskull: so your recommendation would be to convert it to relative if the path is in the current compartment?

naugtur: if and only if, yes.

naugtur (approved):

packages/compartment-mapper/src/import-hook.js
boneskull: this is done.

PR #2766, opened 2025-04-15
Introduce initial "ocapn" package

kumavis:
- Introduces new `ocapn` package with initial codecs for OCapN messages and Passable types
- subsumes `syrup` package into `ocapn` as we don't intend to support generic syrup parsing


packages/ocapn/test/syrup/reader.test.js
kriskowal, line 1: Not clear why this is marked as binary.

packages/ocapn/test/_table.js
kriskowal, line 1: It’s worth making a note that this framework works in string-space and that means there are some intentional limitations and choices. You can definitely carry binary in strings if you only use the uint8 half of the uint16 and convert to ArrayBuffer with set and not TextEncoder.

packages/ocapn/test/_table.js
```diff
@@ -0,0 +1,523 @@
+const sym = s => `${s.length}'${s}`;
+const str = s => `${s.length}"${s}`;
```
kriskowal, line 2: Consider making the expected behavior for `>0xff` words explicit, either forbidding or tolerating correctly. That would look like:

```suggestion
const str = s => {
  const b = new TextEncoder().encode(s);
  return `${b.byteLength}"${String.fromCharCode(...b)}`;
};
```

Same for sym/sel

packages/ocapn/test/_table.js
```diff
@@ -0,0 +1,523 @@
+const sym = s => `${s.length}'${s}`;
```
kriskowal, line 1: Consider `sel`

packages/ocapn/test/_table.js
```diff
@@ -0,0 +1,523 @@
...
+const str = s => `${s.length}"${s}`;
+const bts = s => `${s.length}:${s}`;
```
kriskowal, line 3: Consider `assert(s.every(b => b <= 0xff))`

packages/ocapn/test/_table.js
```diff
@@ -0,0 +1,523 @@
...
+const bool = b => (b ? 't' : 'f');
+// eslint-disable-next-line @endo/restrict-comparison-operands
+const int = i => `${Math.floor(Math.abs(i))}${i < 0 ? '-' : '+'}`;
+const list = items => `[${items.join('')}]`;
+const makeNode = (transport, address, hints) => {
+  return `<10'ocapn-node${sym(transport)}${bts(address)}${bool(hints)}>`;
```
kriskowal, line 9: Rec sym/sel
```suggestion
  return `<${sym('capn-node')}${sym(transport)}${bts(address)}${bool(hints)}>`;
```

kriskowal (approved):

packages/ocapn/test/syrup/reader.test.js
kumavis: oh strange -- there's a comment that includes syrup text that includes binary encoded float64

if its problematic I can remove the comment

https://github.com/endojs/endo/blob/f140ab61991d2fb22ee7ebaa8b92bc221563a8ce/packages/ocapn/test/syrup/reader.test.js#L120

packages/ocapn/test/_table.js
kumavis: this is redundant with https://github.com/endojs/endo/pull/2766#discussion_r2047425192 correct?

packages/ocapn/test/syrup/reader.test.js
kumavis: removed binary values from comment

packages/ocapn/test/_table.js
kumavis: determined correct

PR #2769, opened 2025-04-19
fix(compartment-mapper): fix crawling

boneskull:
The `node_modules`-crawler in `@endo/compartment-mapper` is currently unstable. When there are multiple paths to a module, the "shortest path" to that module remains the first path that was found.

The first path that is found is non-deterministic.  As a result, a `CompartmentDescriptor.path` value can differ from run to run.

This fixes the problem: when a package location in the graph is encountered for the _second_ time (and every time thereafter), we recursively (and its dependency tree, if that too needs updating) check if it still has the shortest path, and update paths as appropriate within the graph.

A fixture in `packages/compartment-mapper/test/fixtures-shortest-path` shows an example of the structure necessary to trigger the original problem.  Previous to the change, this test would typically fail within 20 iterations, thus exhibiting non-determinism.

### DIFF
--- a/packages/compartment-mapper/src/node-modules.js
+++ b/packages/compartment-mapper/src/node-modules.js
@@ -126,6 +126,83 @@ const readDescriptorWithMemo = async (memo, maybeRead, packageLocation) => {
 };
 
 /**
+ * Compares `logicalPath` to the current best logical path in `preferredPackageLogicalPathMap` for `packageLocation`.
+ *
+ * If no current best path exists, it returns `logicalPath`.
+ *
+ * @template {string[]} T
+ * @template {string[]} U
+ * @param {T} logicalPath
+ * @param {string} packageLocation
+ * @param {Map<string, U>} preferredPackageLogicalPathMap
+ * @returns {T|U}
+ */
+const currentBestLogicalPath = (
+  logicalPath,
+  packageLocation,
+  preferredPackageLogicalPathMap,
+) => {
+  const theCurrentBest = preferredPackageLogicalPathMap.get(packageLocation);
+  if (theCurrentBest === undefined) {
+    return logicalPath;
+  }
+  return pathCompare(logicalPath, theCurrentBest) < 0
+    ? logicalPath
+    : theCurrentBest;
+};
+
+/**
+ * Updates the shortest paths in a subgraph of `graph` starting with `packageLocation`.
+ *
+ * This should be called upon the second (and each subsequent) visit to a graph node.
+ *
+ * @param {Graph} graph Graph
+ * @param {string} packageLocation Location of the package to start with
+ * @param {string[]} logicalPath Current path parts of the same package
+ * @param {Map<string, string[]>} [preferredPackageLogicalPathMap] Mapping of shortest known paths for each package location
+ * @returns {void}
+ */
+const updateShortestPaths = (
+  graph,
+  packageLocation,
+  logicalPath,
+  preferredPackageLogicalPathMap = new Map(),
+) => {
+  const node = graph[packageLocation];
+  if (!node) {
+    throw new ReferenceError(
+      `Cannot find package at ${packageLocation} in graph`,
+    );
+  }
+
+  const bestLogicalPath = currentBestLogicalPath(
+    logicalPath,
+    packageLocation,
+    preferredPackageLogicalPathMap,
+  );
+
+  if (bestLogicalPath === logicalPath) {
+    preferredPackageLogicalPathMap.set(packageLocation, bestLogicalPath);
+
+    for (const name of keys(node.dependencyLocations).sort()) {
+      const packageLocation = node.dependencyLocations[name];
+      if (!packageLocation) {
+        // "should never happen"
+        throw new ReferenceError(
+          `Expected graph node ${q(node.name)} to contain a dependency location for ${q(name)}`,
+        );
+      }
+      updateShortestPaths(graph, packageLocation, [...logicalPath, name]);
+    }
+  }
+
+  if (node.path !== bestLogicalPath) {
+    node.path = bestLogicalPath;
+  }
+
+  return undefined;
+};
+/**
  * `findPackage` behaves as Node.js to find third-party modules by searching
  * parent to ancestor directories for a `node_modules` directory that contains
  * the name.
@@ -293,6 +370,13 @@ const graphPackage = async (
   } = {},
 ) => {
   if (graph[packageLocation] !== undefined) {
+    updateShortestPaths(
+      graph,
+      packageLocation,
+      logicalPath,
+      preferredPackageLogicalPathMap,
+    );
+
     // Returning the promise here would create a causal cycle and stall recursion.
     return undefined;
   }
@@ -321,6 +405,7 @@ const graphPackage = async (
     optionalDependencies = {},
     devDependencies = {},
   } = packageDescriptor;
+  /** @type {Record<string, string>} */
   const allDependencies = {};
   for (const [name, descriptor] of Object.entries(
     commonDependencyDescriptors,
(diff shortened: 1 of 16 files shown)


packages/compartment-mapper/test/map-node-modules.test.js
```diff
@@ -0,0 +1,47 @@
...
+    // but return type of `t.assert()` is incorrect; it should use the `asserts` keyword.
+
+    if (expectedPath) {
+      t.deepEqual(
+        /** @type {CompartmentDescriptor} */ (compartmentDescriptor).path,
+        expectedPath,
+        `compartment descriptor should have had path: ${expectedPath.join('>')}`,
+      );
+    } else {
+      expectedPath = /** @type {CompartmentDescriptor} */ (
```
boneskull, line 42: note to self:

this is one of `['pippo', 'gambadilegno', 'goofy']` or `['paperino', 'topolino', 'goofy']` _but_ it should always be one and never the other.  and no, I don't actually know which one it should be.

boneskull: **update**: it should be the latter, and it now always is. 😄

packages/compartment-mapper/test/map-node-modules.test.js
```diff
@@ -0,0 +1,47 @@
...
+import url from 'node:url';
+import test from 'ava';
+import { mapNodeModules } from '../src/node-modules.js';
+import { makeReadPowers } from '../src/node-powers.js';
+
+/**
+ * @import {CompartmentDescriptor} from '../src/types.js'
+ */
+
+// since we cannot dynamically define tests in AVA (🤦‍♂️), this test will run in a loop.
```
kriskowal, line 13: Consider `test.macro`.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -125,6 +125,55 @@ const readDescriptorWithMemo = async (memo, maybeRead, packageLocation) => {
...
 
+/**
+ *
+ * @param {Graph} graph
+ * @param {string} packageLocation
+ * @param {string[]} logicalPath
+ * @param {Map<string, string[]>} [preferredPackageLogicalPathMap]
+ * @returns {Promise<void>}
+ */
+const updateShortestPaths = async (
```
kriskowal, line 136: My intuition is that this function should be synchronous since it only needs to visit the nodes that have already been discovered and assigned potentially-longer paths.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -125,6 +125,55 @@ const readDescriptorWithMemo = async (memo, maybeRead, packageLocation) => {
...
+        .map(async name => {
+          const packageLocation = node.dependencyLocations[name];
+          if (packageLocation) {
+            return updateShortestPaths(graph, packageLocation, [
+              ...logicalPath,
+              name,
+            ]);
+          }
+        }),
+    );
```
kriskowal, line 169: Indeed, I don’t see any promises in this traversal. Could be a simple loop.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -293,6 +342,13 @@ const graphPackage = async (
   } = {},
 ) => {
   if (graph[packageLocation] !== undefined) {
+    await updateShortestPaths(
+      graph,
+      packageLocation,
+      logicalPath,
+      preferredPackageLogicalPathMap,
+    );
```
kriskowal, line 350: This probably obviates code farther down, run immediately before calling this function. https://github.com/endojs/endo/pull/2769/files#diff-047437f94608177a76b8636b72612ed583b0ae5a6d5cb619b851e7830c4425d8R575-R579

kriskowal (commented):
This looks like the right shape. It needs to be synchronous to be correct, since we need to hold a lock on discovery of further packages while we shorten the graph. And it looks like we can trivially make that routine synchronous.

packages/compartment-mapper/test/map-node-modules.test.js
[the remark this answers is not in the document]
boneskull: @kriskowal Unsure what you're seeing that a macro buys us.  I could write a macro, but I'd still have to explicitly repeat calling test _n_ times:

```js
test('iteration 1', myMacro);
test('iteration 2, myMacro);
test('iteration 3', myMacro);
test('iteration 4', myMacro);
```

boneskull:
- [ ] make `updateShortestPaths` synchronous
- [ ] use a custom read power in test with a random delay
- [ ] do whatever [this comment](https://github.com/endojs/endo/pull/2769#discussion_r2057078143) is referring to

packages/compartment-mapper/src/node-modules.js
boneskull: @kriskowal I tried removing that bit, but it seems to be necessary.  I think because  `updateShortestPaths()` will only be called if we encounter the package two or more times.

kriskowal: Can it trigger `updateShortestPaths` instead to reduce duplication?

packages/compartment-mapper/test/fixtures-shortest-path/README.md
kriskowal, line 1: ❤️

packages/compartment-mapper/test/map-node-modules.test.js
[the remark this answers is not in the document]
kriskowal: Not a blocking consideration, but looping would be fine.

```js
for (let i = 0; i < 10; i += 1) test(`iteration ${i}`, myMacro, i);
```

packages/compartment-mapper/test/map-node-modules.test.js
```diff
@@ -0,0 +1,63 @@
...
+  const readPowers = makeReadPowers({ fs, url });
+  const { maybeRead } = readPowers;
+
+  // inserts a random delay before the read
+  readPowers.maybeRead = async specifier => {
+    return new Promise(resolve => {
+      const delay = Math.random() * 50;
+      return setTimeout(resolve, delay);
+    }).then(() => maybeRead(specifier));
+  };
```
kriskowal, line 28: As a matter of style, you’ve got some options that would simplify this. The `async` is unnecessary if you return a `new Promise` explicitly, and you can also boil it down to an arrow without return or braces.

But, on the other hand, you could factor out a `delay(ms) => Promise` utility function, keeping the `async`, like

```js
async specifier => {
  await delay(ms);
  return maybeRead(specifier);
};
```

kriskowal (approved):

packages/compartment-mapper/src/node-modules.js
boneskull: No, but I extracted what I could into another function to reduce duplication.

packages/compartment-mapper/test/map-node-modules.test.js
boneskull: I've changed this. 😈

packages/compartment-mapper/test/map-node-modules.test.js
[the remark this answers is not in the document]
boneskull: if you could do that, then I could do it without a macro. 😉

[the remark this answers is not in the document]
boneskull: trying this a second time, it does actually work as I thought it should. I don't know why it was failing to detect the tests before. ¯\\_(ツ)_\/¯

scripts/mermaid-deptree.cjs
```diff
@@ -0,0 +1,108 @@
...
+ * Converts a dependency graph into a Mermaid graph definition.
+ *
+ * @param {Record<string, string[]>} graph - The dependency graph.
+ * @returns {string} - A Mermaid graph definition.
+ */
+function generateMermaidGraph(graph) {
+  const lines = ['graph TD'];
+  for (const [packageName, dependencies] of Object.entries(graph)) {
+    dependencies.forEach(dependency => {
+      lines.push(`    ${packageName} --> ${dependency}`);
```
naugtur, line 72: packageName could contain a character that breaks mermaid.
tell copilot to use indexOf on keys as source of IDs

see: https://github.com/naugtur/trust-graph/blob/main/src/visualize-trust.js#L142

PR #2773, opened 2025-04-24
feat(pass-stlye): export make-tagged

kumavis:
`@endo/pass-style` exports `makeTagged` via
```js
import { makeTagged } from '@endo/pass-style/make-tagged.js';
```
 As requested in https://github.com/endojs/endo/pull/2772#discussion_r2059334845

### DIFF
--- a/packages/pass-style/make-tagged.js
+++ b/packages/pass-style/make-tagged.js
@@ -1,0 +1,1 @@
+export { makeTagged } from './src/makeTagged.js';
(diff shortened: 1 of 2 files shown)


kumavis:
determined there was no "bundle size savings" but making this export

erights (commented):
LGTM in meeting @kriskowal 's request. But since there is no bundle-size savings, delegating final approval to @kriskowal .

@kriskowal , since there is no bundle size savings, and you still interested in this separation?

PR #2774, opened 2025-04-25
feat(pass-style): add selector

kumavis:


mhofman:
What's a selector?

kriskowal:
> What's a selector?

OCapN Selector corresponds to Syrup symbol and Scheme symbol. Because we can’t use JavaScript symbols (risk of registry stuffing attacks), MarkM insisted we give this a distinguishing name. It follows that it should have its own passStyle, which should replace symbols entirely in the long run.

https://github.com/ocapn/ocapn/blob/main/draft-specifications/Model.md#selector

packages/pass-style/src/types.d.ts
```diff
@@ -22,7 +22,7 @@ export type PrimitiveStyle =
   | 'string'
   | 'symbol';
 
-export type ContainerStyle = 'copyRecord' | 'copyArray' | 'tagged';
+export type ContainerStyle = 'copyRecord' | 'copyArray' | 'tagged' | 'selector';
```
kriskowal, line 25: I don’t believe selector qualifies as a container. It’s more like a string. But, the reïfication of a selector is an object, so that might be the distinguishing feature for containers. I do not know.

packages/pass-style/src/types.d.ts
```diff
@@ -31,7 +31,7 @@ export type PassStyle =
   | 'error'
   | 'promise';
 
-export type TaggedOrRemotable = 'tagged' | 'remotable';
+export type TaggedOrRemotable = 'tagged' | 'selector' | 'remotable';
```
kriskowal, line 34: Selector definitely does not belong in this union.

kriskowal (commented):
I defer to @erights and @mhofman in this package. I have questions about whether Selector (née Symbol) fits in the established categories. It represents an Atom on the wire, but one we are obliged to reïfy as an Object.

packages/pass-style/src/types.d.ts
kumavis: its a container for a string! yeah ok was mostly just following 'tagged' around the code base

packages/pass-style/src/types.d.ts
kumavis: its a container for a string! yeah ok was mostly just following 'tagged' around the code base

kumavis:
> we can’t use JavaScript symbols (risk of registry stuffing attacks)

is the concern here only from using `Symbol.for`? would making our own symbol registry as a WeakRefMap (map to weakly held values) resolve the attack vector?

erights:
Seeing this thread, I realize I gave you bad advise. Selector is like Tagged in some ways, but is like Symbol in other ways. I'll make a PR staged on this one as my corrected review suggestions. Hold on.

> would making our own symbol registry as a WeakRefMap (map to weakly held values) resolve the attack vector?

Would be an interesting experiment, but I don't expect it to be a better solution than what this PR is attempting.

erights:
https://github.com/endojs/endo/pull/2777

Thanks for the great idea, and sorry for how it took for me to get it.

packages/pass-style/src/types.d.ts
gibson042: It's definitely not a container. Our best documentation is at https://endojs.github.io/endo/types/_endo_pass_style.Passable.html , and we could describe Selector as primitive if we're taking a data model perspective, or otherwise a new special case (because, unlike the existing primitives, `Object.is` will differentiate identical values).

kumavis:
closed for https://github.com/endojs/endo/pull/2777

PR #2775, opened 2025-04-25
fix(compartment-mapper): add missing arg to recursive call

boneskull:
This fixes a bug introduced in #2769 which can lead to a "max call stack exceeded" exception.

And another that will potentially overwrite the entry node with a path if a cycle is encountered.


kriskowal (approved):

boneskull:
Added another test fixture proving the entry `Node`'s `path` is not overwritten when we have a cycle.

PR #2777, opened 2025-04-26
fix(pass-style)! flip which symbols are passable

erights:
Closes: #XXXX
Refs: https://github.com/ocapn/ocapn/issues/165 https://github.com/endojs/endo/pull/2772 https://github.com/endojs/endo/pull/2774 #2778 #2779 https://github.com/tc39/proposal-symbol-predicates https://github.com/Agoric/agoric-sdk/pull/11338

## Description

@kriskowal and @kumavis just had a great idea! Instead of simply avoiding symbols altogether because of gc worries, switch from treating only registered and well known symbols as passable to treating only unregistered symbols as passable. Then, treat two unregistered symbols as equal within our distributed object semantics iff they have the same `description` string. 

This does what I thought impossible: it recapitulates at the user level for unregistered symbols the implementation technique that I had recommended engines do internally for registered symbols --- forget any registry, and just treat two registered symbols are equal if they have the same contents. Just like JS already does for strings.

Most well known symbols are unregistered. I think they all must be unregistered, but at least on Node, `Symbol.dispose` and `Symbol.asyncDispose` are mistakenly registered symbols.

- [ ] file bugs that `Symbol.dispose` and `Symbol.asyncDispose` are mistakenly registered symbols.

In any case, for all well-known symbols that are unregistered, these are passable with no special case. But remember that they unserialize to fresh symbol with distinct JS identities.

### Security Considerations

By always unserializing to fresh symbols, we avoid coinciding with the identity of pre-existing JS symbols. This helps separate concerns in ways helpful to security.

### Scaling Considerations

This fixes the gc problem we had with registered symbols without introducing a separate `Selector` object type. This scales better.

This will make harmonization with ocapn smoother.

### Testing Considerations

So far, the change has only required changes that were testing how symbols pass. We have not seen any other breakage yet. But the real test will be integration with agoric-sdk.

I marked this PR with `!` because it is a compat break. To ease the transition I also added a `PASS_STYLE_LEGACY_ASYNC_ITERATOR_SYMBOL` environment variable which defaults to `'disabled'`. But if set to `'enabled'`, it tries to be somewhat accommodating to legacy uses of `Symbol.asyncIterator` until we can retire those.

- [x] This change is fundamentally incompatible with remotables having symbol-named methods. This PR should enhance `passStyleOf` to reject any such objects rather than classifying them as remotables.

### Upgrade Considerations

If we currently have any old-style passable symbols in durable storage, or if there are elsewhere persistent samples of serialized data containing old-style passable symbols, we may be in trouble. We need to check. If we are in trouble, we'll need an upgrade strategy to get across this transition.

### DIFF
--- a/packages/exo/test/heap-classes.test.js
+++ b/packages/exo/test/heap-classes.test.js
@@ -3,6 +3,7 @@
 import test from '@endo/ses-ava/prepare-endo.js';
 
 import { getInterfaceMethodKeys, M } from '@endo/patterns';
+import { testFullOrderEQ } from '@endo/marshal/tools/ava-full-order-eq.js';
 import {
   GET_INTERFACE_GUARD,
   defineExoClass,
@@ -81,22 +82,21 @@ test('test defineExoClass', t => {
   t.deepEqual(upCounter[GET_INTERFACE_GUARD]?.(), UpCounterI);
   t.deepEqual(getInterfaceMethodKeys(UpCounterI), ['incr']);
 
-  const symbolic = Symbol.for('symbolic');
   const FooI = M.interface('Foo', {
     m: M.call().returns(),
-    [symbolic]: M.call(M.boolean()).returns(),
+    m2: M.call(M.boolean()).returns(),
   });
-  t.deepEqual(getInterfaceMethodKeys(FooI), ['m', Symbol.for('symbolic')]);
+  testFullOrderEQ(t, getInterfaceMethodKeys(FooI), ['m', 'm2']);
   const makeFoo = defineExoClass('Foo', FooI, () => ({}), {
     m() {},
-    [symbolic]() {},
+    m2() {},
   });
   const foo = makeFoo();
   t.deepEqual(foo[GET_INTERFACE_GUARD]?.(), FooI);
   // @ts-expect-error intentional for test
-  t.throws(() => foo[symbolic]('invalid arg'), {
+  t.throws(() => foo.m2('invalid arg'), {
     message:
-      'In "[Symbol(symbolic)]" method of (Foo): arg 0: string "invalid arg" - Must be a boolean',
+      'In "m2" method of (Foo): arg 0: string "invalid arg" - Must be a boolean',
   });
 });
 
--- a/packages/exo/test/non-enumerable-methods.test.js
+++ b/packages/exo/test/non-enumerable-methods.test.js
@@ -1,6 +1,7 @@
 import test from '@endo/ses-ava/prepare-endo.js';
 
 import { objectMetaMap } from '@endo/common/object-meta-map.js';
+import { testFullOrderEQ } from '@endo/marshal/tools/ava-full-order-eq.js';
 import { getInterfaceMethodKeys, M } from '@endo/patterns';
 import { defineExoClass } from '../src/exo-makers.js';
 import { GET_INTERFACE_GUARD } from '../src/get-interface.js';
@@ -49,25 +50,24 @@ test('test defineExoClass', t => {
   t.deepEqual(upCounter[GET_INTERFACE_GUARD](), UpCounterI);
   t.deepEqual(getInterfaceMethodKeys(UpCounterI), ['incr']);
 
-  const symbolic = Symbol.for('symbolic');
   const FooI = M.interface('Foo', {
     m: M.call().returns(),
-    [symbolic]: M.call(M.boolean()).returns(),
+    m2: M.call(M.boolean()).returns(),
   });
-  t.deepEqual(getInterfaceMethodKeys(FooI), ['m', Symbol.for('symbolic')]);
+  testFullOrderEQ(t, getInterfaceMethodKeys(FooI), ['m', 'm2']);
   const makeFoo = defineExoClass(
     'Foo',
     FooI,
     () => ({}),
     denumerate({
       m() {},
-      [symbolic]() {},
+      m2() {},
     }),
   );
   const foo = makeFoo();
   t.deepEqual(foo[GET_INTERFACE_GUARD](), FooI);
-  t.throws(() => foo[symbolic]('invalid arg'), {
+  t.throws(() => foo.m2('invalid arg'), {
     message:
-      'In "[Symbol(symbolic)]" method of (Foo): arg 0: string "invalid arg" - Must be a boolean',
+      'In "m2" method of (Foo): arg 0: string "invalid arg" - Must be a boolean',
   });
 });
--- a/packages/marshal/src/marshal-justin.js
+++ b/packages/marshal/src/marshal-justin.js
@@ -4,6 +4,7 @@ import { Nat } from '@endo/nat';
 import {
   getErrorConstructor,
   isObject,
+  nameForPassableSymbol,
   passableSymbolForName,
 } from '@endo/pass-style';
 import { q, X, Fail } from '@endo/errors';
@@ -115,8 +116,6 @@ const makeNoIndenter = () => {
 
 const identPattern = /^[a-zA-Z]\w*$/;
 harden(identPattern);
-const AtAtPrefixPattern = /^@@(.*)$/;
-harden(AtAtPrefixPattern);
 
 /**
  * @param {Encoding} encoding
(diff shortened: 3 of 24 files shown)


packages/marshal/src/marshal-justin.js
```diff
@@ -315,24 +314,16 @@ const decodeToJustin = (encoding, shouldIndent = false, slots = []) => {
           return out.next(`${BigInt(digits)}n`);
         }
         case '@@asyncIterator': {
-          // TODO deprecated. Eventually remove.
+          // TODO deprecated. Eventually remove. TODO now
           return out.next('Symbol.asyncIterator');
         }
```
erights, line 319: remove

packages/exo/test/heap-classes.test.js
```diff
@@ -81,12 +82,23 @@ test('test defineExoClass', t => {
...
   const FooI = M.interface('Foo', {
     m: M.call().returns(),
     [symbolic]: M.call(M.boolean()).returns(),
   });
-  t.deepEqual(getInterfaceMethodKeys(FooI), ['m', Symbol.for('symbolic')]);
+  // Cannot use `t.deepEqual` because it does not recognize that two
+  // unregistered symbols with the same `description` are the same in
+  // our distributed object semantics. Unfortunately, `compareRank` is
+  // too imprecise. We'd like to also test `keyEQ`, but that would violate
+  // our package layering.
```
erights, line 94: For all these, see if we could use @gibson042 's ordering test instead,

erights:
So far at least, it is very satisfying that I only needed to change two non-test files, only one non-trivially, and everything else other than passable symbol tests just worked.

mhofman:
I don't understand how unregistered symbols with fresh identities can be used in anything meaningful. How do you look up that symbol on an object ? Do you have to iterate over all properties and check the description? That seems irrational.

erights:
The ocapn Selector is a primitive forgeable type, like strings and numbers. Because it is primitive, it would be nice to map it only JS primitives, i.e., such that `Object(x) !== x`. For ocapn purposes, these are indeed useless for naming symbol-named properties. That's why I say above

> This change is fundamentally incompatible with remotables having symbol-named methods. This PR should enhance passStyleOf to reject any such objects rather than classifying them as remotables.

Except for this case, which we will fix, there is other possibility for passable objects to have symbol-named properties.

erights:
IOW, neither @kriskowal nor I (and possibly also not @kumavis ) liked mapping an ocapn primitive value onto a JS object.

OTOH, we have no choice but to map ocapn ByteArray to restricted Immutable ArrayBuffers, so this is a distinction that has to give anyway. I still think it is cleaner not to do it for ocapn Selectors.

packages/marshal/src/marshal-justin.js
erights: Done

packages/exo/test/heap-classes.test.js
erights: I was confused. The code I had in mind is `orderInvariants`. But now that I looked at it, it is not relevant. Resolving.

kriskowal:
In OCapN as with status quo for `pass-style`, OCapN Symbols are not eligible as field names on OCapN Struct / Endo Record, but are eligible as arguments to function calls. OCapN protocol convention is that the first argument is an OCapN Symbol/Selector for method invocation. For OCapN, the first argument symbol’s description will carry the corresponding JavaScript method name for an Exo receiver. We are only admitting Symbols into OCapN at all because the Scheme side of the house also uses Symbols for enums, for which a JavaScript unforegeable symbol is sufficient.

So, in short, symbols are not terribly useful for Endo’s purposes, but using Symbols to denote the same passable values in OCapN, JavaScript, and Guile Scheme will be less confusing than alternatives.

erights:
> for which a JavaScript unforegeable symbol is sufficient

In our use of it, as in the OCapN semantics, it is forgeable in our distributed object semantics. At the JS level, they happen to be unforgeable, but we need to steer clear of that unforgeability.

packages/exo/test/non-enumerable-methods.test.js
```diff
@@ -49,12 +50,23 @@ test('test defineExoClass', t => {
...
+  const symbolic = Symbol('symbolic');
   const FooI = M.interface('Foo', {
     m: M.call().returns(),
     [symbolic]: M.call(M.boolean()).returns(),
   });
-  t.deepEqual(getInterfaceMethodKeys(FooI), ['m', Symbol.for('symbolic')]);
+  // Cannot use `t.deepEqual` because it does not recognize that two
+  // unregistered symbols with the same `description` are the same in
+  // our distributed object semantics. Unfortunately, `compareRank` is
+  // too imprecise. We'd like to also test `keyEQ`, but that would violate
```
kumavis, line 61: > `compareRank` is too imprecise.

could you explain more here?

packages/marshal/test/_marshal-test-data.js
```diff
@@ -49,12 +49,10 @@ export const roundTripPairs = harden([
   [9007199254740993n, { '@qclass': 'bigint', digits: '9007199254740993' }],
 
   // Well known symbols
-  [Symbol.asyncIterator, { '@qclass': 'symbol', name: '@@asyncIterator' }],
-  [Symbol.match, { '@qclass': 'symbol', name: '@@match' }],
-  // Registered symbols
-  [Symbol.for('foo'), { '@qclass': 'symbol', name: 'foo' }],
-  // Registered symbol hilbert hotel
-  [Symbol.for('@@foo'), { '@qclass': 'symbol', name: '@@@@foo' }],
+  [Symbol.asyncIterator, { '@qclass': 'symbol', name: 'Symbol.asyncIterator' }],
```
kumavis, line 52: Does this mean well known symbols are still passable?

packages/pass-style/src/symbol.js
```diff
@@ -1,117 +1,70 @@
...
+
+/**
+ * An adapter to help transition to flip which symbols are passable.
+ * See `ses-utils.js` in `@agoric/internal` at
+ * https://github.com/Agoric/agoric-sdk/pull/11338
+ *
+ * @param {string} name
+ * @returns {symbol}
+ */
+export const unpassableSymbolForName = name => Symbol(name);
```
kumavis, line 70: name is confusing

kumavis (commented):
asked a few clarifying questions

packages/exo/test/non-enumerable-methods.test.js
erights: Thanks. Comment on these usage sites deleted, and instead all these call the new `makeFullOrderComparatorKit`, which
- uses `fullCompare` rather than `rankCompare` to get better precision. The difference is that `fullCompare` will only judge two remotables as equal if they are the same remotable, whereas `rankCompare` considers all remotable to have the same rank. Really, that's the relevant additional precision we would have gotten from `keyEQ` anyway.

The outliers are promises and errors. In our distributed object semantics, neither of these have a meaningful equality.
- `rankCompare` and `fullCompare` both treat all promises as having the same rank, and treat all errors as having the same rank.
- Neither promises nor errors are Keys, so `keyEQ` rejects both.

Does this answer you question clearly?

packages/marshal/test/_marshal-test-data.js
erights: Good question! 

- To ease the transition, we are currently making a temporary special case for `Symbol.asyncIterator` both for endo here, and for agoric-sdk at https://github.com/Agoric/agoric-sdk/pull/11338

- Aside from that temporary special case, we treat all unregistered symbols as passable, whether well-known or not. However, when we unserialize such symbols, we create fresh ones with the same description, as we do for all unregistered symbols. Thus they lose their JS equality to the well-known symbols.

packages/pass-style/src/symbol.js
erights: Suggestions?

erights: Huh. Now I'm confused. Hold on. More soon.

erights: The right hand side should have been `Symbol.for(name)`. Fixed.

erights:
***Reviewers***, does anyone mind if I squash differential PR #2779 and then merge it into this one #2777 ?

erights:
Putting into Draft until we have time to discuss this approach.

kriskowal:
I believe we have tentative consensus that we want to move toward modeling CapTP symbols as JavaScript objects with passStyle `"symbol"` and toStringTag of the symbol string. This is reduces the risk of incompatibility with our history because Ava `t.deepEqual` will continue to report that pass-equal symbols are equal. It also eliminates the risk of symbol registry stuffing.

So, the representation of CapTP symbols would bi-map:
- `@@asyncIterator` <-> `Symbol.asyncIterator`
- other <-> `{ @passStyle: 'symbol', @@toStringTag: other }`

For OCapN, we would bi-map method names to avoid collisions with loaded JS object property names:
- `'@@asyncIterator` <-> `Symbol.asyncIterator`
- `'then` <-> `Symbol.for('then')
- `'constructor` <-> `Symbol.for('constructor')`

I recommend that we close this PR and reconstruct the change for the new consensus. For expedience please reopen if you disagree.

PR #2782, opened 2025-04-30
Harden before Lockdown

kriskowal:
Refs: #1686

## Reviewer instructions

This change introduces two packages, with implementation factored out of `ses` and _lightly_ modified. To highlight those modifications, the commits are individually reviewable. In separate commits, I have copied certain modules to new locations verbatim, then applied changes in place. For example, the `ses/commons.js` effectively moves out to another package, but parts stay in `ses` and other parts move to `harden/commons.js` since they are not otherwise shared. The `harden` tests are duplicated and lightly edited, since the intended behavior is similar before and after lockdown.

## Description

This change introduces an `@endo/harden` package that provides the implementation of `harden` for `ses` but also a relaxed implementation of `harden` suitable before `lockdown`, for modules that should interoperate between HardenedJS and non-HardenedJS. The expectation is that we will use `@endo/harden` in most of Endo, like `marshal` and `pass-style`, to make CapTP usable with or without Lockdown.

### Security Considerations

The `harden` provided pre-lockdown differs only in that it does not visit prototypes. The `harden` and `intrisnics` packages remain inside the SES TCB and are linted according to the same stricter restrictions for access to free variables and use of polymorphic calls.
Hardening before and after lockdown should converge on the same state.

### Scaling Considerations

Hardening before and after lockdown should converge on the same state, but redundantly.
Hardening before lockdown and after lockdown use distinct tables for tracking prior work.

The new modules have minimal README documentation. Further documentation for the `harden` package in particular should follow.

### Testing Considerations

Tests preserved and copied with the appropriate variation for pre-lockdown.

Compatibility is preserved.

The `harden` and `intrinsics` packages may be versioned on an independent cadence of `ses` since they are not directly exposed. That is, `intrinsics` may make breaking changes to its API without breaking `ses`.

The `intrinsics` package may or may not be a suitable foundation for a future `get-intrinsics` shim which we expect to be entangled with the internals of the SES `permits`. Future designs will determine how permits and the get intrinsics shims get packaged and it is our hope that this structure does not impede progress in that direction.

### Upgrade Considerations

None.


kriskowal:
I should note that I am not hoping for timely review of these changes.

packages/harden/CHANGELOG.md
```diff
@@ -0,0 +1,4 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
```
erights, line 3: What is the scope of "project"? Do you mean "package"?

packages/harden/README.md
```diff
@@ -0,0 +1,39 @@
+# harden
+
+For common usage, this package provides an implementation of `harden` that will
+freeze an object and its transitive properties, making them shallowly
+immutable.
+If used in composition with Lockdown in [HardenedJS](https://hardenedjs.org),
```
erights, line 6:
```suggestion
If used in composition with `lockdown` in [HardenedJS](https://hardenedjs.org),
```

packages/harden/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+export const behavior = () => {
+  return harden({});
+};
+harden(behavior);
+```
+
+In this example, we create a hadened module.
+That is, the module's surface and any value reachable by interacting with its
+behaviors transitively produces values that sare safe to share with guest code.
```
erights, line 25:
```suggestion
behaviors transitively produces values that are safe to share with guest code.
```

packages/harden/README.md
```diff
@@ -0,0 +1,39 @@
+# harden
+
+For common usage, this package provides an implementation of `harden` that will
+freeze an object and its transitive properties, making them shallowly
```
erights, line 4:
```suggestion
freeze an object and its transitive own properties, making them shallowly
```

Additional pedantic detail that can be omitted in a high level summary, but needs to be explained somewhere. It does a "transitive reflective" walk of properties. The difference is that avoids invoking getters. Rather it proceeds to walk both getters and setters as objects.

packages/harden/src/commons.js
```diff
@@ -0,0 +1,57 @@
+import { TypeError, freeze, getOwnPropertyDescriptor } from '@endo/intrinsics';
+
+const er1StackDesc = getOwnPropertyDescriptor(Error('er1'), 'stack');
+const er2StackDesc = getOwnPropertyDescriptor(TypeError('er2'), 'stack');
+
+let feralStackGetter;
+let feralStackSetter;
+if (er1StackDesc && er2StackDesc && er1StackDesc.get) {
+  // We should only encounter this case on v8 because of its problematic
+  // error own stack accessor behavior.
```
erights, line 10: We should cite Jordan's recent Error stack proposal.

packages/harden/src/commons.js
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * If on a v8 with the problematic error own stack accessor behavior,
+ * `FERAL_STACK_GETTER` will be the shared getter of all those accessors
+ * and `FERAL_STACK_SETTER` will be the shared setter. On any platform
+ * without this problem, `FERAL_STACK_GETTER` and `FERAL_STACK_SETTER` are
+ * both `undefined`.
+ *
+ * @type {(() => any) | undefined}
+ */
+export const FERAL_STACK_GETTER = feralStackGetter;
```
erights, line 46: Nice that you pulled these out here. However, should the ses error handling also be pulled out into a separate low level module, they may need to coordinate here. Not sure.

packages/harden/src/make-hardener.js
```diff
@@ -0,0 +1,285 @@
...
+  }
+};
+
+/**
+ * @import {Harden} from '../types.js'
+ */
+
+// Obtain the string tag accessor of of TypedArray so we can indirectly use the
+// TypedArray brand check it employs.
+const typedArrayToStringTag = getOwnPropertyDescriptor(
```
erights, line 65: Separately, I have been thinking that once Immutable ArrayBuffers are in the language and adequately deployed, we should retire harden's special case for TypedArrays. If someone wants to harden a TypedArray, they should start with a frozen TypedArray.

erights (approved):
> I should note that I am not hoping for timely review of these changes.

Well, you got one anyway. LGTM!

kriskowal:
Agreement out-of-band from @mhofman: There is a hazard that an object `harden`ed before `lockdown` might not be fully hardened under a `lockdown` regime. There are sophisticated and subtle to maintain this invariant in an application, but one simple way: if `lockdown` sense that a `harden` was used before `lockdown`, throw an error. There should only be two kinds of applications: those that use `harden` and never call `lockdown` and other applications that call `lockdown` before ever calling `harden`.

So, I can introduce a coarse mechanism to maintain the partition of lockdown and no-lockdown hardening (that can be frustrated by eval twins, but eval twins are simply frustrating). It involves a single bit of module scope state for which I hope to be forgiven.

mhofman:
I also want to change the harden semantics to only verify that the prototype has been previously (or concurrently) hardened under lockdown. In non lockdown it would not check the prototype. This is a breaking change and I would prefer if it happened at the same time or before to avoid having to invite a future breaking change of the standalone harden (unless thus reifies as a 3 configuration state for the prototype traversal option)

kriskowal:
I think that change can be made additive, even with the current framing. I don’t expose the `makeHarden` function directly to any consumer. If you import `@endo/harden`, you get a singleton with the pre-lockdown configuration, but which falls thru to `globalThis.harden` if present. I exposed a `makeHarden` variant for `ses` to import at `@endo/harden/make-harden.js`. This could be parameterized further, or a new method name could be added, for the new behavior. I take it that we will also want a lockdown option to migrate to the new behavior gracefully.

packages/harden/CHANGELOG.md
kriskowal: Yes. I’ll fix uniformly in follow-up. This boilerplate is in every package.

packages/harden/src/make-hardener.js
```diff
@@ -271,5 +307,31 @@ export const makeHardener = () => {
...
+      harden(root) {
+        // Use a native hardener if possible.
+        if (typeof globalThis.harden === 'function') {
+          const globalHarden = globalThis.harden;
+          return globalHarden(root);
+        } else {
+          return innerHarden(root);
+        }
+      },
+    }))(harden));
```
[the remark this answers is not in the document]
gibson042: But I must admit, this `harden` being vulnerable to future globalThis manipulation seems to fly in the face of our goals. What if it were instead a one-time sensitivity?
```suggestion
    let innerHarden = harden;
    harden = root => {
      // Capture the global hardener (or lack thereof) on first invocation, and
      // use it if present.
      if (mode === undefined) {
        mode = ascendPrototypeChains;
        const globalHarden = globalThis.harden;
        if (typeof globalHarden === 'function' && globalHarden !== harden) {
          innerHarden = globalHarden;
        }
      }
      return innerHarden(root);
    };
```
or
```suggestion
    let innerHarden = harden;
    let globalHarden;
    harden = root => {
      // Capture the global hardener (or lack thereof) on first invocation,
      // require it to remain constant, and use it if present.
      if (mode === undefined) {
        mode = ascendPrototypeChains;
        globalHarden = globalThis.harden;
        if (typeof globalHarden === 'function' && globalHarden !== harden) {
          innerHarden = globalHarden;
        }
      }
      if (globalThis.harden !== globalHarden) {
        throw new TypeError('a global hardener must not be replaced');
      }
      return innerHarden(root);
    };
```

naugtur (changes requested):
please release a version of SES before merging this

packages/intrinsics/src/intrinsics.js
```diff
@@ -0,0 +1,302 @@
...
+  if (result !== object) {
+    // See https://github.com/endojs/endo/blob/master/packages/ses/error-codes/SES_DEFINE_PROPERTY_FAILED_SILENTLY.md
+    throw TypeError(
+      `Please report that the original defineProperty silently failed to set ${stringifyJson(
+        String(prop),
+      )}. (SES_DEFINE_PROPERTY_FAILED_SILENTLY)`,
+    );
+  }
+  return result;
+};
```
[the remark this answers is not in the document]
erights: On this one, I'm suspicious. So we should test before we assume that they "resolved" it because they actually fixed it. A one-time manual test would be fine.

packages/intrinsics/src/intrinsics.js
```diff
@@ -0,0 +1,302 @@
...
+
+/**
+ * @type { &
+ *   ((that: string, searchValue: { [Symbol.replace](string: string, replaceValue: string): string; }, replaceValue: string) => string) &
+ *   ((that: string, searchValue: { [Symbol.replace](string: string, replacer: (substring: string, ...args: any[]) => string): string; }, replacer: (substring: string, ...args: any[]) => string) => string)
+ * }
+ */
+export const stringReplace = /** @type {any} */ (
+  uncurryThis(stringPrototype.replace)
+);
```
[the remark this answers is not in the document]
kriskowal: I’m choosing to walk past this since this is a refactor, and I don’t think we need a more expressive type since this is currently internal to `ses` and can be relaxed without breaking later.

packages/intrinsics/package.json
```diff
@@ -0,0 +1,81 @@
+{
+  "name": "@endo/intrinsics",
```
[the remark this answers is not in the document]
kriskowal: In order to apply this feedback and not create a rebase nightmare for future self, I’m going to reconstruct this PR as if `@endo/ses-primordials` were my first choice. I’m making the name more specific so as to leave a void for a package with a broader charter than serving the overlapping needs of `ses` and `harden`.

kriskowal:
Toward reviving this stale branch, I had a conversation out-of-band with @mhofman that uncovered a hazard we want to address before merging this change as designed. That is, if a hardened module initializes before lockdown, it may be in a state where its prototypes are unfrozen until the first instance is hardened.

This is related to an orthogonal hazard of having modules that explicitly support both shallow hardening before lockdown and deep hardening after lockdown. We do not believe we can in good conscience relax post-lockdown hardening so that it throws an exception if it encounters an unhardened prototype instead of implicitly capturing it in the closure of to-be-hardened objects. This would create a hazard where a module uses shallow pre-lockdown harden for its own partial defense but is incompatible with a post-lockdown mode because it neglected to pre-harden its prototypes. As a condition of proceeding with this design @mhofman proposes that we add a lockdown mode that issues warnings when harden encounters an unhardened prototype, in order to softly encourage code to pre-harden prototypes and eliminate a window between creation and instantiation where an attacker can manipulate an exported prototype.

We currently have a mechanism in place that causes a pre-lockdown harden to throw an exception if it is called again post-lockdown, but should add a mechanism that causes repairIntrinsics to throw an exception if it is called after a pre-lockdown harden was called. (It occurs to me as I summarize this, that the latter mechanism may obviate the former.)

This new mechanism has some requirements:
- Older and newer versions of both `@endo/harden` and `ses` need to tolerate meeting in the wild.
- Multiple instances of `@endo/harden` must be able to coëxist in an application that never locks down. So, we can’t simply have every implementation call through to a global implementation, as that would invite cyclic recursion. We might be able to instead invite the winner of the race to detect that they won the race and use their implementation instead of the global.

We also wish to encourage code going forward to adopt `@endo/harden` as a veneer over the post-lockdown harden. To that end, in a distant future, the point of coördination between `@endo/harden` and lockdown may not be `globalThis.harden` but something else, tentatively the existence of a `Object.@@harden` property.

It occurs to me as I summarize that the simplest formulation of this problem is that, between eval twins of the pre-lockdown harden and the installation of a post-lockdown harden, we can and should only ever use one implementation of these. For pre-lockdown hardens, this is not a race to initialization, but a race to first-use, with the additional requirement that lockdown should fail if it loses the race.

@mhofman proposes a mechanism that will universally cause lockdown to fail if repairIntrinsics occurs after any pre-lockdown harden gets used: the installation of a non-writable non-configurable property on an intrinsic, e.g., Object.@@harden.

# Proposed change

So, the proposed design is to have pre-lockdown hardens race to install a Object.@@harden and then defer to Object.@@harden going forward. They will need to fall back to globalThis.harden if it is present in order to function with versions of ses that predate the decision to install harden at Object.@@harden.

As a courtesy to developers, we add explicit code to repairIntrinsics that detects Object.@@harden and throws a more informative error if they are using a more recent version of ses than this change.

erights (changes requested):
Just requesting changes in order to turn off my approval in light of upcoming big changes. I want to be sure to review again after these big changes but before merge. I do not currently have any changes in mind to request.

kriskowal:
As much for my own edification, with propositions labeled for reference:

1. I start from the (urgent) premise that it be possible to join an ocap network without lockdown or oblige the application author to coordinate the order of installation of shims. To that end,
2. Suppose that we partition all valid programs into those that use a pre-lockdown `harden` and those that use a post-lockdown `harden`, and any application that tries to do both fails when they try to `lockdown`.
3. To ensure A, suppose that co-tenant `@endo/harden` implementations race to install non-writable, non-configurable `Object[Symbol.for('harden')]` upon the first invocation of `harden`. Any instance of `@endo/harden` is eligible to win the race and determines the semantics (and holder of the memo), and the losers silently tolerate losing the race. Whereas, if `lockdown` loses the race, it throws an exception.
4. We accept this is not the usual strategy for shims, which lately tend to encourage overwriting what was found, because there is a possibility applications will rely on the behavior of the shim and break if a future version of the language behaves differently than the shim predicted. We also accept that this kind of shimming is weirdly compatible with HardenedJS because it tolerates a previously locked-down environment.
5. The `@endo/harden` module would fail upon first attempt to `harden` inside a `Compartment` in composition with a version of SES that does not install `Object[Symbol.for('harden')]` and where `globalThis.harden` was not endowed. No such versions of `ses` currently exist, as compartments receive `harden` by default. No future version of `ses` would break because it will introduce `Object[Symbol.for('harden')]` which cannot be denied by the creator of a `Compartment`. It is a shared intrinsic that cannot be omitted from a `Compartment` except under extreme duress.
6. To ensure that `@endo/harden` works with versions of SES that predate the decision to install `Object[Symbol.for('harden')]`, it will also sense (upon first invocation) that it lost the race if it instead finds `globalThis.harden` is installed, and will fall-through to that instead.
7. There would be no versions of `@endo/harden` that predate the choice to install `Object[Symbol.for('harden')]`, so no mutual compatibility concerns among versions of `@endo/harden`.
8. Using the `"exports"` directive in the `@endo/harden` `package.json`, we can let the user of an application or the creator of a bundle elect a mode of `harden` that it will use if it wins the race. So, we can choose a default mode and elective alternate modes. Electing an alternative mode would look like `bundle-source -C shallow-harden`
9. The mode most useful for bundling applications that can presume that they are executed in a locked-down environment would simply omit its own implementation of `harden` and use the one installed at `globalThis.harden` or `Object[Symbol.for('harden')]`. That would look like `bundle-source -C lockded-down`. So, bundles that migrate from assuming the existence of a post-lockdown `globalThis.harden` can migrate to use `@endo/harden`, become more portable, without incurring a significant bundle size increase.
10. For every mode of `@endo/harden`, the mode variant only applies to the pre-lockdown behavior.
11. Ideally, a library that is built to use `@endo/harden` does not need to separately test its behavior pre- and post-lockdown. If it works with pre-lockdown, it should not break post-lockdown.
12. All post-lockdown `harden` options must guarantee that the transitive closure of both prototypes and properties are frozen, because calling `harden` should relieve the reviewer of any concern of reachable transitive mutability, even for code that might be run pre-lockdown.
13. Regardless of whether `harden` closes over prototypes when hardening instances, a library that does not immediately and explicitly harden a prototype will be vulnerable to mutation in coordination with other linked modules until the first instance is hardened, which may be considerably later, so we ideally provide some signal or encouragement to opportunistically `harden`.

The dimensions to the design of a `harden` implementation are:
- **pre** vs **post**
- **harden extent** the extent that `harden` freezes
- **throwing extent** the extent in which `harden` will throw if it finds an unfrozen object
- **warning extent** the extent in which `harden` will warn if it finds an unfrozen object
- for each extent, whether it close over:
  - **surface** transitive closure over properties of an object
  - **volume** transitive closure over both properties and prototypes of an object
  - **global** `globalThis`
  - **undeniable intrinsics** intrinsics that can’t be denied because they’re reachable with syntax, like `AsyncFunction.prototype`.
  - **shared intrinsics** the shared intrinsics, a set that may grow version over version
  - **special intrinsics** intrinsics that are reachable by calling methods of *shared intrinsics*, a set that may grow version over version

`harden` variants of interest.
- **fake harden** which returns the subject unaltered
- **assuring harden** for which the **harden extent** is the **volume** of the subject (only current behavior)
- **throwing harden** for which the **harden extent** is the **surface** of the subject and otherwise the **throwing extent** is the **volume** of the subject.
- **warning harden** for which the **harden extent** is the **surface** and otherwise the **warning extent** is the **volume**
- **harden until intrinsics** for which the **harden extent** is the **volume** of the subject except for the **volume** of **shared intrinsics**
- **harden until global** for which the **harden extent** is the **volume** of the subject except for the **volume** of **global** and **special intrinsics**

Discussion of variants:
- **assuring harden** would not be eligible before `lockdown` because it hardens shared intrinsics in a way that would interfere with `lockdown`, and also can interfere with shims of anything captured by a hardened instance. It’s technically at peace with `lockdown` because (2) ensures that using `harden` before `lockdown` foreswears ever calling `lockdown`. But, aggressively hardening prototypes down to the intrinsics is inconsistent with the values of a pre-lockdown application.
- **warning harden** pre-lockdown entrains complications about reporting warnings, because we will want knobs to mute or redirect those warnings, which will have to be on `process.env`. This isn’t a deal-breaker.
- **warning harden** post-lockdown runs afoul of (12) because it may proceed after `harden` for a subject with unfrozen objects in its volume.
- A library designed to work pre-lockdown with any eligible flavor of `harden` except **throwing harden** may fail to work post-lockdown with a **throwing harden** post-lockdown if the author failed to harden a prototype of a hardened instance.
- Moving to use **throwing harden** would be a breaking change for some existing libraries and would need to be opted-in with a `lockdown` option.
- Neither **harden until intrinsics** or **harden until global** are eligible post-lockdown because (12) they leave shared mutable objects in their wake. These are strictly pre-lockdown variants.
- Both **harden until intrinsics** and **harden until global** close over an extent that can get captured in a bundled version of `@endo/harden`. This is fine for post-lockdown usage because the bundled version will be ignored. This may be fine pre-lockdown because hardening an instance won’t accidentally freeze something that might be shimmed, particularly not objects that hide behind method invocation, unless you consider the case of shimming an object exported by another module by pre-emptively importing it.

Tangentially, the regarding whether `@endo/harden` should prefer `Object[Symbol.for('harden')]` over `globalThis.harden`:
- **Favor Object** because `Object` will always be frozen post-lockdown and `globalThis` is not necessarily. This is a weakly preferable since Compartments are not generally co-tenant and co-tenant Compartments **must** freeze `globalThis`.
- **Favor globalThis** because this gives individual compartments the option of receiving from without or installing from within a specific alternative `globalThis.harden`.

My tentative preference is:
- `ses` post-lockdown provides **assuring harden** by default
- `ses` post-lockdown provides **throwing harden** if configured so folks can opt into early discovery of unhardened prototypes. (Recall that we can’t do **warning harden** post-lockdown.)
- `@endo/harden` races to provide a **fake harden** implementation
- `@endo/harden` with `-C hardened` presumes `lockdown` and throws upon initialization if it doesn’t find `harden` in the environment
- `@endo/harden` with other `-C` conditions to race to install stronger pre-lockdown harden like **throwing harden** or **warning harden**, maybe suppressing warnings for objects reachable from the volume of global or known shared intrinsics.

kriskowal:
Closing with intention to instead pursue https://github.com/endojs/endo/pull/2976

PR #2783, opened 2025-04-30
feat(ses): exclude immutable-arraybuffer shim on Hermes

leotm:
Unblock
- https://github.com/endojs/endo/pull/2400
  - https://github.com/endojs/endo/actions/runs/14745659058/job/41392366659?pr=2400

(hacky solution prior to endo sync discussion)

### DIFF
--- a/packages/ses/scripts/hermes-transforms.js
+++ b/packages/ses/scripts/hermes-transforms.js
@@ -61,11 +61,21 @@ const asyncGeneratorDestroyer = {
   FunctionDeclaration: destroyAsyncGenerators,
 };
 
+// Remove on Hermes since classes and private fields are unsupported on Hermes
+const removeImmutableArrayBufferShim = {
+  ImportDeclaration(path) {
+    if (path.node.source.value === '@endo/immutable-arraybuffer/shim.js') {
+      path.remove();
+    }
+  },
+};
+
 export const hermesTransforms = {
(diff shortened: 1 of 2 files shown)


erights (commented):
Would it make more sense to stage this on #2400 ? Would the change to `get-anonymous-intrinsics.js` then disappear from this as a differential PR?

Even with it shrunk to just the change to hermes-transforms.js, I am unfamiliar with `ImportDeclaration(path)` and so defer to @kriskowal for final approval.

FWIW, if it means what I think it means, LGTM for now. But we would still need to get the shim actually working well enough on Hermes if we want endo-on-Hermes to participate in ocapn, yes?

leotm:
> Would it make more sense to stage this on #2400 ?

was thinking exactly that, i'll add the commit now to unblock which fixes the Hermes CI over there

leotm:
> Would the change to `get-anonymous-intrinsics.js` then disappear from this as a differential PR?

oops 😅 yep not going to claim your credit, all then disappears except the condition i've added which is correct
(the condition is needed or Hermes will try `ab.transferToImmutable()` which isn't a function cos we've transformed away the import)

leotm:
> FWIW, if it means what I think it means, LGTM for now. But we would still need to get the Immutable ArrayBuffer shim actually working well enough on Hermes if we want endo-on-Hermes to participate in ocapn, yes?

yes this is the next move ^

leotm:
staged to
- https://github.com/endojs/endo/pull/2400

PR #2784, opened 2025-04-30
refactor: Deprecate trivialComparator in favor of type-specific `compareNumerics`

gibson042:
Ref #2113

## Description

We want to discourage comparison of strings by UTF-16 code unit as implemented by relational operators.

### Security Considerations

None.

### Scaling Considerations

Includes a trivial (and separable) improvement to `sortByRank`.

None beyond `@deprecated`.

### Testing Considerations

Covered by existing tests.

n/a

### Upgrade Considerations

I don't think a NEWS.md update is warranted, but could be persuaded otherwise.

### DIFF
--- a/packages/marshal/src/rankOrder.js
+++ b/packages/marshal/src/rankOrder.js
@@ -308,6 +308,8 @@ harden(assertRankSorted);
  * @returns {T[]}
  */
 export const sortByRank = (passables, compare) => {
+  /** @type {T[]} mutable for in-place sorting, but with hardened elements */
+  let unsorted;
   if (Array.isArray(passables)) {
     harden(passables);
     // Calling isRankSorted gives it a chance to get memoized for
@@ -316,9 +318,10 @@ export const sortByRank = (passables, compare) => {
     if (isRankSorted(passables, compare)) {
       return passables;
     }
+    unsorted = [...passables];
+  } else {
+    unsorted = Array.from(passables, harden);
   }
-  const unsorted = [...passables];
-  unsorted.forEach(harden);
   const sorted = unsorted.sort(compare);
   // For reverse comparison, move `undefined` values from the end to the start.
   // Note that passStylePrefixes (@see {@link ./encodePassable.js}) MUST NOT
--- a/packages/marshal/src/rankOrder.js
+++ b/packages/marshal/src/rankOrder.js
@@ -8,7 +8,7 @@ import {
 
 /**
  * @import {Passable, PassStyle} from '@endo/pass-style'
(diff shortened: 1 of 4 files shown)


packages/marshal/src/rankOrder.js
```diff
@@ -44,9 +44,25 @@ const { entries, fromEntries, setPrototypeOf, is } = Object;
...
+
+/**
+ * @template {number | bigint} T
+ * @param {T} left
+ * @param {T} right
+ * @returns {RankComparison}
+ */
+export const compareNumerics = (left, right) =>
+  // eslint-disable-next-line no-nested-ternary, @endo/restrict-comparison-operands
+  left < right ? -1 : left > right ? 1 : 0;
```
erights, line 65: What about `NaN`?

packages/marshal/src/rankOrder.js
```diff
@@ -55,15 +71,20 @@ export const trivialComparator = (left, right) =>
...
+        return aPrefixes < bPrefixes ? -1 : aPrefixes > bPrefixes ? 1 : 0;
       })
       .map(([passStyle, prefixes], index) => {
-        // Cover all strings that start with any character in `prefixes`,
-        // verifying that it is sorted so that is
+        // Verify that `prefixes` is sorted, and cover all strings that start
+        // with any of its characters, i.e.
         // all s such that prefixes.at(0) ≤ s < successor(prefixes.at(-1)).
-        prefixes === [...prefixes].sort().join('') ||
+        prefixes === prefixes.split('').sort().join('') ||
```
erights, line 87: Is `prefixes` an array of strings or a string?

erights (approved):
LGTM

packages/marshal/src/rankOrder.js
gibson042: It's a string from [`passStylePrefixes`](https://github.com/endojs/endo/blob/db9e48f24fd3b532c8478e5cfc0cf9720fc13fcb/packages/marshal/src/encodePassable.js#L867-L884), usually one character like "s" or "?", but two characters for copyArray "[^" (which also covers "\" and "]") and bigint "np" (which also covers "o").

packages/marshal/src/rankOrder.js
gibson042: Hmm, good point. The current implementation of `trivialComparator` returns 1 for any comparison involving NaN, which is obviously broken but doesn't come up in either the Endo or agoric-sdk repositories. Still, we shouldn't leave that gap, so I've reverted the changes to `trivialComparator` and updated `compareNumerics` to support `compareRank`.

PR #2787, opened 2025-05-01
feat(path-compare): create @endo/path-compare

boneskull:
## Description

This extracts the `pathCompare` (~~and `stringCompare`~~ `stringCompare` is present, but not exposed as a public API) functions from `@endo/compartment-mapper` so that they may be used directly by the ecosystem.

Ref: #2789 for use of this package

### Security Considerations

none

### Scaling Considerations

n/a

n/a

### Testing Considerations

"We" wrote new tests and added them.

n/a

### Upgrade Considerations

n/a

### DIFF
--- a/packages/path-compare/README.md
+++ b/packages/path-compare/README.md
@@ -1,0 +1,85 @@
+# @endo/path-compare
+
+> [Shortlex](https://en.wikipedia.org/wiki/Shortlex_order)-based comparison of string arrays
+
+## Overview
+
+This is used by [@endo/compartment-mapper][] when crawling a `node_modules` directory to find the shortest _path_ to any given transitive dependency from the entry package.
+
+### Example
+
+Given a dependency tree like this:
+
+```mermaid
+graph TD
+  entry --> foo
+  foo --> bar
+  bar --> baz
+```
+
+The computed _path_ from `entry` to `baz` as `['foo', 'bar', 'baz']`.
+
+If we have another dependency tree like this:
+
+```mermaid
+graph TD
+  entry --> foo
+  foo --> a
+  a --> b
+  b --> baz
+  foo --> bar
+  bar --> baz
+```
+
+The computed _path_ from `entry` to `baz` is again `['foo', 'bar', 'baz']`, because that has fewer elements than `['foo', 'a', 'b', 'baz']`.
+
+If we have yet another dependency tree like this:
+
+```mermaid
+graph TD
+  entry --> foo
+  foo --> bar
+  foo --> quux
+  bar --> baz
+  quux --> baz
+```
+
+The computed _path_ from `entry` to `baz` is again `['foo', 'bar', 'baz']`. This is because while `['foo', 'quux', 'baz']` is another path with the same array length, the algorithm _must_ choose one, and so chooses the one with the fewest cumulative characters.
+
+And if we have a tree like this:
+
+```mermaid
+graph TD
+  entry --> foo
+  foo --> spam
+  foo --> quux
+  spam --> baz
+  quux --> baz
+```
+
+The computed path from `entry` to `baz` is `['foo', 'quux', 'baz']`. Both the array length and cumulative character counts are the same. `['foo', 'quux', 'baz']` is lexicographically "before" `['foo', 'spam', 'baz']` (at the first code unit of each array's second element).
+
+> Note: the "lexicographic" comparison uses the UTF-16 code unit order, and thus may be surprising.
+
+## Usage
+
+This package exports `pathCompare`.
+
+### `pathCompare(a?: string[], b?: string[]) => number`
+
+Compares two arrays of strings (if provided) and returns a numeric comparison result: 0 if they are equal, -1 if `a` is "less than" `b`, and 1 if `a` is "greater than" `b` (where "less than" means having fewer elements, or the same element count but a shorter cumulative length, or the same element count and cumulative length but prior when comparing each element lexicographically by UTF-16 code unit).  
+
+```js
+import { pathCompare } from '@endo/path-compare';
+
+const path1 = ['foo', 'bar', 'baz'];
+const path2 = ['foo', 'quux', 'baz'];
+
+const result = pathCompare(path1, path2); // -1
+```
+
+## License
+
+Copyright (c) 2021 Endo Project Contributors. Licensed Apache-2.0
+
+[@endo/compartment-mapper]: https://github.com/endojs/endo/tree/main/packages/compartment-mapper
--- a/packages/path-compare/SECURITY.md
+++ b/packages/path-compare/SECURITY.md
@@ -1,0 +1,38 @@
+# Security Policy
+
+## Supported Versions
+
+The SES package and associated Endo packages are still undergoing development and security review, and all
+users are encouraged to use the latest version available. Security fixes will
+be made for the most recent branch only.
+
+## Coordinated Vulnerability Disclosure of Security Bugs
+
+SES stands for fearless cooperation, and strong security requires strong collaboration with security researchers. If you believe that you have found a security sensitive bug that should not be disclosed until a fix has been made available, we encourage you to report it. To report a bug in HardenedJS, you have several options that include:
+
+* Reporting the issue to the [Agoric HackerOne vulnerability rewards program](https://hackerone.com/agoric).
+
+* Sending an email to security at (@) agoric.com., encrypted or unencrypted. To encrypt, please use  @Warner’s personal GPG key  [A476E2E6 11880C98 5B3C3A39 0386E81B 11CAA07A](http://www.lothar.com/warner-gpg.html)  .
+
+* Sending a message on Keybase to `@agoric_security`, or sharing code and other log files via Keybase’s encrypted file system. ((_keybase_private/agoric_security,$YOURNAME).
+
+* It is important to be able to provide steps that reproduce the issue and demonstrate its impact with a Proof of Concept example in an initial bug report. Before reporting a bug, a reporter may want to have another trusted individual reproduce the issue.
+
+* A bug reporter can expect acknowledgment of a potential vulnerability reported through  [security@agoric.com](mailto:security@agoric.com)  within one business day of submitting a report. If an acknowledgement of an issue is not received within this time frame, especially during a weekend or holiday period, please reach out again. Any issues reported to the HackerOne program will be acknowledged within the time frames posted on the program page.
+	* The bug triage team and Agoric code maintainers are primarily located in the San Francisco Bay Area with business hours in  [Pacific Time](https://www.timeanddate.com/worldclock/usa/san-francisco) .
+
+* For the safety and security of those who depend on the code, bug reporters should avoid publicly sharing the details of a security bug on Twitter, Discord, Telegram, or in public Github issues during the coordination process.
+
+* Once a vulnerability report has been received and triaged:
+	* Agoric code maintainers will confirm whether it is valid, and will provide updates to the reporter on validity of the report.
+	* It may take up to 72 hours for an issue to be validated, especially if reported during holidays or on weekends.
+
+* When the Agoric team has verified an issue, remediation steps and patch release timeline information will be shared with the reporter.
+	* Complexity, severity, impact, and likelihood of exploitation are all vital factors that determine the amount of time required to remediate an issue and distribute a software patch.
+	* If an issue is Critical or High Severity, Agoric code maintainers will release a security advisory to notify impacted parties to prepare for an emergency patch.
+	* While the current industry standard for vulnerability coordination resolution is 90 days, Agoric code maintainers will strive to release a patch as quickly as possible.
+
+When a bug patch is included in a software release, the Agoric code maintainers will:
+	* Confirm the version and date of the software release with the reporter.
+	* Provide information about the security issue that the software release resolves.
+	* Credit the bug reporter for discovery by adding thanks in release notes, securing a CVE designation, or adding the researcher’s name to a Hall of Fame.
--- a/packages/path-compare/package.json
+++ b/packages/path-compare/package.json
@@ -1,0 +1,72 @@
+{
+  "name": "@endo/path-compare",
+  "version": "1.0.0",
+  "description": "Compare string arrays",
+  "keywords": [
+    "string",
+    "array",
+    "compare",
+    "comparison"
+  ],
+  "author": "Endo contributors",
+  "license": "Apache-2.0",
+  "homepage": "https://github.com/endojs/endo/tree/master/packages/path-compare#readme",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/endojs/endo.git",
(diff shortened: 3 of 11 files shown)


packages/path-compare/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+const str1 = 'foo';
+const str2 = 'bar';
+
+const result = stringCompare(str1, str2); // 1
+```
+
+## License
+
+Copyright (c) 2021 Endo Project Contributors. Licensed Apache-2.0
```
boneskull, line 86: 2021 was when this function first appeared, from what I can tell. 😜

packages/path-compare/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+const path1 = ['foo', 'bar', 'baz'];
+const path2 = ['foo', 'quux', 'baz'];
+
+const result = pathCompare(path1, path2); // -1
+```
+
+### `stringCompare(a: string, b: string) => number`
+
+Compares two strings using UTF-16 code unit order and returns a numeric comparison result: 0 if they are equal, -1 if `a` is "less than" `b`, and 1 if `a` is "greater than" `b`.
```
kriskowal, line 73: Not blocking:

I forgot that we’re doing UTF-16 compare. I can’t think of a reason why this would be a problem for our bundler, but maybe @gibson042 can.

kriskowal (commented):
Looks great.

Consider adding a commit that integrates this into `compartment-mapper` to prove that this works in its integration. You might be able to preserve continuity of `blame` if you create some commits that strategically `git mv` the original files, followed by a commit that makes changes in place, to make the differences more evident.

packages/path-compare/src/index.js
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * Compares two string arrays and returns a comparison result.
+ *
+ * The algorithm is as follows:
+ *
+ * 1. _Check if either value is `undefined`._ If both are `undefined`, return `0`. If `a` is `undefined`, return `1`. If `b` is `undefined`, return `-1`.
+ * 2. _Check the lengths of the arrays._ If they are different, return the difference.
+ * 3. _Check the cumulative lengths of the arrays_ using the count of UTF-16 units in each string. If they are different, return the difference.
+ * 4. _Check the individual elements of the arrays_ via lexical comparison. If they are different, return the difference.
+ * 5. _If all elements are the same_ ("deep equality"), return `0`.
```
gibson042, line 43:
```suggestion
 * 1. _Check if either value is `undefined`._ If both are `undefined`, return
      `0`. If `a` is `undefined`, return `1`. If `b` is `undefined`, return
      `-1`.
 * 2. _Check the lengths of the arrays._ If they are different, return the
      difference.
 * 3. _Check the cumulative lengths of the arrays_ using the count of UTF-16
      code units in each string. If they are different, return the difference.
 * 4. _Check the individual elements of the arrays_ via lexical comparison. If
      they are different, return the difference.
 * 5. _If all elements are the same_ ("deep equality"), return `0`.
```

packages/path-compare/test/index.test.js
```diff
@@ -0,0 +1,58 @@
...
+  t.true(pathCompare(['a', 'b'], ['a']) > 0);
+});
+
+test('pathCompare - returns negative for smaller cumulative length', t => {
+  t.true(pathCompare(['a', 'b'], ['aa', 'bb']) < 0);
+});
+
+test('pathCompare - returns positive for larger cumulative length', t => {
+  t.true(pathCompare(['aa', 'bb'], ['a', 'b']) > 0);
+});
```
gibson042, line 34: I'd like these tests to cover smaller/larger cumulative length where the naïve element-by-element approach would favor the opposite result, e.g. `["bb", "aa"]` before `["a", "bbbb"]`.

gibson042 (changes requested):
I'm fine with comparing lexicographically by UTF-16 code unit, but there's some other nuance here that merits explicit documentation.

packages/path-compare/README.md
naugtur: Different consideration: Is the ordering being pedantically correct more important to us than performance? A regular string compare after length check should be fine - we need the ordering to be stable, not perfectly correct, I think.

packages/path-compare/test/index.test.js
```diff
@@ -0,0 +1,58 @@
+import test from '@endo/ses-ava/prepare-endo.js';
+import { pathCompare, stringCompare } from '../src/index.js';
```
naugtur, line 2: We could use a test demonstrating the utf-16-ness of the comparison

packages/path-compare/src/index.js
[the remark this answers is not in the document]
kriskowal: We do have the liberty to change behavior here. The contract is to produce consistent results within a version, not consistency with previously constructed bundles. Hashes of generated bundles can change in even patch releases without a break of contract. I don’t think we would break anything by omitting step 3, and I agree that it would be less surprising.

packages/path-compare/README.md
```diff
@@ -0,0 +1,88 @@
...
+  quux --> baz
+```
+
+The computed path to `baz` from `entry` is `['foo', 'quux', 'baz']`. Both the array length and cumulative character counts are the same. However, `['foo', 'quux', 'baz']` is lexicographically "before" `['foo', 'spam', 'baz']`.
+
+> Note: the "lexicographic" comparison uses the UTF-16 code unit order, and thus may be surprising.
+
+## Installation
+
+(the usual suspects)
```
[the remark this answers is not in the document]
boneskull: removed

packages/path-compare/README.md
boneskull: For our purposes, the fact that it does what it does is not really important; all we care about is that someone could eyeball two strings and reasonably agree with the outcome—and that the outcome is consistent.  It just _so happens_  to use UTF-16 code units.  

> Of course, if the _only_ means to accomplish this _is_ UTF-16 code unit comparison (I don't know), then it does not _so happen_ to be implemented this way. 

In other words: the function body could use given a different strategy for string comparison as long as it is a) reasonable and b) consistent across any JS environment.

Some options:

- Do not mention that a UTF-16 comparison is happening ("implementation detail")
- Alias `stringCompare` to `stringCompareByCodeUnit` or some other mouthful
- Rename to `universalStringCompare` or something that communicates that, for example, _locale_ has no bearing on the result
- Just rename it as suggested

@gibson042 @kriskowal What do we think of the above?

packages/path-compare/README.md
```diff
@@ -0,0 +1,88 @@
...
+Given a dependency tree like this:
+
+```mermaid
+graph TD
+  entry --> foo
+  foo --> bar
+  bar --> baz
+```
+
+The computed _path_ to `baz` from `entry` as `['foo', 'bar', 'baz']`.
```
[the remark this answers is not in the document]
boneskull: done

packages/path-compare/src/index.js
```diff
@@ -0,0 +1,93 @@
...
+  // Undefined is not preferred
+  if (a === undefined && b === undefined) {
+    return 0;
+  }
+  if (a === undefined) {
+    return 1;
+  }
+  if (b === undefined) {
+    return -1;
+  }
```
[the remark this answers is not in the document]
boneskull: going to leave this as-is, as I do not want to change the algorithm itself. this can be done in another PR

packages/path-compare/src/index.js
```diff
@@ -0,0 +1,93 @@
...
+
+  // Prefer the shortest dependency path.
+  if (a.length !== b.length) {
+    return a.length - b.length;
+  }
+
+  // Otherwise, favor the shortest cumulative length.
+  const aSum = a.reduce(cumulativeLength, 0);
+  const bSum = b.reduce(cumulativeLength, 0);
```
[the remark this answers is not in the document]
boneskull: same here

packages/path-compare/src/index.js
boneskull: formatting changes should be applied by tooling 😄

boneskull:
@kriskowal Use of this module is in PR #2789 (targeting this branch)

boneskull:
@gibson042 @kriskowal I've applied many of the requested changes, but have _thus far_ declined to:

- Modify the algorithm per @gibson042's suggestions
- Rename `stringCompare` (see [question here](https://github.com/endojs/endo/pull/2787#discussion_r2074218407))

packages/path-compare/src/index.js
boneskull: (I did this manually. something like [prettier-plugin-jsdoc](https://www.npmjs.com/package/prettier-plugin-jsdoc) would be helpful)

packages/path-compare/README.md
gibson042: My first preference is still not exporting this function at all, with a distant second being renaming to indicate the UTF-16 code unit detail.

packages/path-compare/src/index.js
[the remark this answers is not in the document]
gibson042: Works for me.

packages/path-compare/src/index.js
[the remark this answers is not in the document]
gibson042: Eh, ok.

packages/path-compare/README.md
naugtur: I'm also in favor of treating it as implementation detail (also, other places where it's used are unrelated to path compare).

packages/path-compare/src/index.js
```diff
@@ -0,0 +1,98 @@
...
+
+const { stringify: q } = JSON;
+
+/**
+ * String comparison function based on UTF-16 code units
+ *
+ * @type {CompareFn<string>}
+ */
+// eslint-disable-next-line no-nested-ternary
+export const stringCompare = (a, b) => (a === b ? 0 : a < b ? -1 : 1);
```
kriskowal, line 22: Feel free to keep this export since it can be masked by the top-level thunk. That leaves you free to import it in tests.

kriskowal (changes requested):
Small changes. I believe we have a clear plan to get this merged.

packages/path-compare/package.json
```diff
@@ -0,0 +1,75 @@
...
+  "main": "./index.js",
+  "module": "./index.js",
+  "exports": {
+    ".": "./index.js",
+    "./package.json": "./package.json"
+  },
+  "scripts": {
+    "build": "exit 0",
+    "lint": "yarn lint:types && yarn lint:eslint",
+    "lint-check": "yarn lint",
```
gibson042: #2824
```suggestion
```

packages/path-compare/package.json
```diff
@@ -0,0 +1,75 @@
...
+    "lint-fix": "yarn lint:eslint --fix && yarn lint:types",
+    "lint:eslint": "eslint '**/*.js'",
+    "lint:types": "tsc",
+    "postpack": "git clean -f '*.d.ts*'",
```
gibson042: #2824
```suggestion
    "postpack": "git clean -f '*.d.ts*' '*.tsbuildinfo'",
```

packages/path-compare/SECURITY.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Supported Versions
+
+The SES package and associated Endo packages are still undergoing development and security review, and all
+users are encouraged to use the latest version available. Security fixes will
+be made for the most recent branch only.
+
+## Coordinated Vulnerability Disclosure of Security Bugs
+
+SES stands for fearless cooperation, and strong security requires strong collaboration with security researchers. If you believe that you have found a security sensitive bug that should not be disclosed until a fix has been made available, we encourage you to report it. To report a bug in HardenedJS, you have several options that include:
```
gibson042: #2814
```suggestion
SES stands for fearless cooperation, and strong security requires strong collaboration with security researchers. If you believe that you have found a security sensitive bug that should not be disclosed until a fix has been made available, we encourage you to report it. To report a bug in HardenedJS, you have several options that include:
```

packages/path-compare/SECURITY.md
```diff
@@ -0,0 +1,38 @@
...
+
+The SES package and associated Endo packages are still undergoing development and security review, and all
+users are encouraged to use the latest version available. Security fixes will
+be made for the most recent branch only.
+
+## Coordinated Vulnerability Disclosure of Security Bugs
+
+SES stands for fearless cooperation, and strong security requires strong collaboration with security researchers. If you believe that you have found a security sensitive bug that should not be disclosed until a fix has been made available, we encourage you to report it. To report a bug in HardenedJS, you have several options that include: 
+
+* Reporting the issue to the [Agoric HackerOne vulnerability rewards program](https://hackerone.com/agoric).
```
gibson042: #2814
```suggestion
* Reporting the issue to the [Agoric HackerOne vulnerability rewards program](https://hackerone.com/agoric).
```

packages/path-compare/src/index.js
```diff
@@ -0,0 +1,98 @@
...
+ * Comparison function for two values of the same type `T`.
+ *
+ * Can be used with `Array.prototype.sort` and other similar contexts
+ *
+ * @template T The type of the values to compare
+ * @callback CompareFn
+ * @param {T} a First value
+ * @param {T} b Second value
+ * @returns {number} Negative integer if `a < b`; positive integer if `a > b`;
+ * `0` or `NaN` if equal
```
gibson042:
```suggestion
 * `0` if equal
```
(other packages in this project interpret `NaN` to mean "incomparable")

packages/path-compare/src/index.js
```diff
@@ -0,0 +1,93 @@
...
+
+  // Otherwise, compare terms lexically.
+  // This loop guarantees that if any pair of terms is different, including the
+  // case where one is a prefix of the other, we will return a non-zero value.
+  for (let i = 0; i < a.length; i += 1) {
+    const comparison = stringCompare(a[i], b[i]);
+    if (comparison !== 0) {
+      return comparison;
+    }
+  }
```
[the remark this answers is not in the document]
gibson042: Reopening unapplied comment-update suggestion.

packages/path-compare/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+If we have another dependency tree like this:
```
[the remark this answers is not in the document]
gibson042: I'd still like the README to include https://github.com/endojs/endo/pull/2787/files#r2072160421 , for completeness.

gibson042 (approved):
This looks ready after minor tweaks, but 🤞 for the followups as well.

packages/path-compare/package.json
boneskull:
```suggestion
    "postpack": "git clean -fX \"*.d.ts*\" \"*.d.cts*\" \"*.d.mts*\" \"*.tsbuildinfo\"",
```

Should be the above now (#2665)

boneskull:
@gibson042 OK, I think I got all yr suggestions in.

PR #2789, opened 2025-05-01
chore(compartment-mapper): use @endo/path-compare

boneskull:
This removes the code copied from `@endo/compartment-mapper` to create `@endo/path-compare` and uses `@endo/path-compare` instead.

Ref: #2787

### DIFF
--- a/packages/compartment-mapper/src/digest.js
+++ b/packages/compartment-mapper/src/digest.js
@@ -16,11 +16,8 @@
  * } from './types.js'
  */
 
-import {
-  assertCompartmentMap,
-  pathCompare,
-  stringCompare,
-} from './compartment-map.js';
+import { pathCompare } from '@endo/path-compare';
+import { assertCompartmentMap, stringCompare } from './compartment-map.js';
 
 const { create, fromEntries, entries, keys } = Object;
 
--- a/packages/compartment-mapper/package.json
+++ b/packages/compartment-mapper/package.json
@@ -62,6 +62,7 @@
   "dependencies": {
     "@endo/cjs-module-analyzer": "workspace:^",
     "@endo/module-source": "workspace:^",
+    "@endo/path-compare": "workspace:^",
     "@endo/trampoline": "workspace:^",
     "@endo/zip": "workspace:^",
     "ses": "workspace:^"
--- a/packages/compartment-mapper/src/compartment-map.js
+++ b/packages/compartment-mapper/src/compartment-map.js
@@ -15,55 +15,6 @@ const q = JSON.stringify;
 export const stringCompare = (a, b) => (a === b ? 0 : a < b ? -1 : 1);
 
 /**
- * @param {number} length
- * @param {string} term
- */
-const cumulativeLength = (length, term) => {
-  return length + term.length;
-};
-
-/**
- * @param {Array<string> | undefined} a
- * @param {Array<string> | undefined} b
- */
-export const pathCompare = (a, b) => {
-  // Undefined is not preferred
-  if (a === undefined && b === undefined) {
-    return 0;
-  }
-  if (a === undefined) {
-    return 1;
-  }
-  if (b === undefined) {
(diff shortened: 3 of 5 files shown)


packages/compartment-mapper/src/digest.js
```diff
@@ -16,11 +16,8 @@
  * } from './types.js'
  */
 
-import {
-  assertCompartmentMap,
-  pathCompare,
-  stringCompare,
-} from './compartment-map.js';
+import { pathCompare, stringCompare } from '@endo/path-compare';
```
gibson042, line 19: I don't think this import justifies exporting `stringCompare` from package "path-compare"—it should either be (re)defined in this file or imported from some other package's `stringCompareByCodeUnit` (or similar) export.

gibson042 (changes requested):
Commentary mirroring that of #2787.

packages/compartment-mapper/src/digest.js
naugtur: The use of stringCompare here vs in pathCompare are two different cases and there's no need for reuse TBF. We loose nothing if they diverge. I'd inline the string comparison in https://github.com/endojs/endo/pull/2789/files#diff-2eb3d852a38c5ca4c0667b1dd108d40a3a4ea03194f368c4032f8536273792b4R76 and avoid the export. 
This is not even our main usecase for pathCompare. (we want it to be stable and shareable with others for the sake of the comparisons we do in node-modules.js)

boneskull: We are exporting `stringCompare` or we are duplicating it, as it's reused [here](https://github.com/endojs/endo/blob/e0534d7c0596dd96526512fb9be0676b5754858e/packages/compartment-mapper/src/digest.js#L76)).

gibson042: Right, and I'm saying that package "path-compare" should not export a function for string comparison. Instead, one of the following alternatives should be followed:
* packages/compartment-mapper/src/digest.js defines its own string comparison function, presumably duplicating the simple one in "path-compare"
* both compartment-mapper and path-compare import a function like `stringCompareByCodeUnit` or `compareStringsByCodeUnit` from package "internal"

boneskull: We shouldn't duplicate it, since `digest.js` expects the same implementation; they should not diverge.

But I am not really happy to create an entirely new package for `stringCompare`, either.

boneskull: I'll dupe it for now. 

Upon further inspection of this use in `digest.js`, this code looks kind of sketchy because (cc @naugtur) we (`@lavamoat/node`) were under the impression that `label` (which we are overwriting) was not used for anything.  Endo needs to either a) avoid touching `label` in `digest.js` or otherwise prove that the rename map is consistent even if somebody happened to go in and overwrite labels.  😇 

At any rate, I agree it looks as if `stringCompare` is not critical here. It should probably be removed entirely and replaced with a lambda or something.

naugtur: duplicating sounds good. 

The part in digest that looks at labels is there to prevent a situation where two things have the same label (which is possible when two copies of the same version of the package somehow exist in the dep tree) and it solves it by adding `-n{index}`
What `@lavamoat/node` is doing is improving the situation and it's no longer a concern, this mechanism should not end up getting invoked if we replace labels with canonicalNames earlier.

kriskowal:
This is just a poke. It looks like we have a clear plan for changes needed to merge.

packages/compartment-mapper/src/digest.js
boneskull: @naugtur It still makes me itchy, so I'm going to avoid mutating it entirely (I did this already)

boneskull:
This can be merged after #2787

kriskowal (approved):

PR #2790, opened 2025-05-02
chore(ci): workflow hygiene

boneskull:
- ~~Pinned all actions to avoid certain classes of vulnerabilities.~~
  - Pinned deps are best maintained automatically by something like Renovate, but given `actions/setup-node` was still at v3, I don't pinning will necessarily incur more maintenance burden if we choose to maintain it manually.
  - See https://www.wiz.io/blog/github-action-tj-actions-changed-files-supply-chain-attack-cve-2025-30066
- Attempt to run against Node.js v22.x in various places (I'll revert this if stuff is still broken)
- Remove invalid syntax
  - `fail-fast` cannot be used if there is no matrix
  - `path` is an invalid input to `actions/setup-node`

I did not implement the following changes, but I would recommend doing so:

- The `cover` and `viable-release` jobs probably don't need to be run more than once:
  - `cover` shouldn't differ between Node.js versions, and the cost/benefit isn't there
  - `viable-release`'s behavior should only depend on the version of `yarn`
- We can extract some of the boilerplate setup into a local action.
- Everything that can run on Node.js 22.x probably should, but I am unsure if we want to incur that cost.
- Jobs only using one Node.js version and configuration do not need a matrix at all (e.g., `test-xs` unless we add 22.x).

Please provide feedback!

**UPDATE May 22 2025**: I have reverted pinning of GH action versions to avoid conflicts with #2810 and #2818 (the latter of which will do so automatically).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,20 +20,18 @@ jobs:
     # begin macro
 
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
 
       # without this, setup-node errors on mismatched yarn versions
       - run: corepack enable
 
       - name: Use Node.js
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
-          node-version: 20.x
+          node-version: 22.x
           cache: yarn
 
       - name: Install dependencies
@@ -64,7 +62,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [18.x, 20.x, 22.4.1]
+        node-version: [18.x, 20.x, 22.x]
         platform: [ubuntu-latest]
         # windows-latest exhibited flakey tests that are not yet worth the
         # trouble to investigate, and blocked us from upgrading yarn from 1 to
@@ -72,13 +70,13 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
 
       # without this, setup-node errors on mismatched yarn versions
       - run: corepack enable
 
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
           node-version: ${{ matrix.node-version }}
           cache: yarn
@@ -121,13 +119,13 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
 
       # without this, setup-node errors on mismatched yarn versions
       - run: corepack enable
 
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
           node-version: ${{ matrix.node-version }}
           cache: yarn
@@ -158,13 +156,13 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
 
       # without this, setup-node errors on mismatched yarn versions
       - run: corepack enable
 
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
           node-version: ${{ matrix.node-version }}
           cache: yarn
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -20,20 +20,18 @@ jobs:
     # begin macro
 
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
```
turadg, line 26: > Pinned all actions to avoid certain classes of vulnerabilities.

What vulnerabilities are you expecting to prevent?

.github/workflows/ci.yml
```diff
@@ -64,21 +62,21 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [18.x, 20.x, 22.4.1]
+        node-version: [18.x, 20.x, 22.x]
```
turadg, line 65: +1 to including v22. @kriskowal since Node 18 EOLed a couple days ago, are you okay with removing it?

.github/workflows/ci.yml
naugtur: the ability of the action authors to publish an update with malicious code running with full access to the CI and ability to silently undermine correctness and security of what's being built.

turadg: if Github makes an update to `actions/checkout` that is malicious, I expect they would halt all the runners before it affected us. The looks like noise to me.

mhofman:
> * Pinned all actions to avoid certain classes of vulnerabilities.

Let's only pin 3rd party actions. GH maintained ones should be out of scope for this class of vulnerability

.github/workflows/ci.yml
mhofman: > * Pinned all actions to avoid certain classes of vulnerabilities.

Let's only pin 3rd party actions. GH maintained ones should be out of scope for this class of vulnerability

mhofman:
> * Everything that can run on Node.js 22.x probably should, but I am unsure if we want to incur that cost.

FYI in agoric-sdk we [recently switched](https://github.com/Agoric/agoric-sdk/pull/11167) to having "semantic" version names for the matrix input, so that bumping node versions doesn't branch required checks to be updated. We are actually considering [further changing this](https://github.com/Agoric/agoric-sdk/issues/11344).

.github/workflows/ci.yml
boneskull: I think that kind of misses the point...

FWIW, GitHub itself recommends you pin SHAs.

.github/workflows/ci.yml
boneskull: That'd be a breaking change...

.github/workflows/ci.yml
turadg: > GitHub is a third party

I don't think that page GitHub is referring to itself as a third party. Even if it were, what you cite also says,
>Although pinning to a commit SHA is the most secure option, specifying a tag is more convenient and is widely used. If you’d like to specify a tag, then be sure that you trust the action's creators. The ‘Verified creator’ badge on GitHub Marketplace is a useful signal, as it indicates that the action was written by a team whose identity has been verified by GitHub. Note that there is risk to this approach even if you trust the author, because a tag can be moved or deleted if a bad actor gains access to the repository storing the action.

`v4` is a tag and thus the status quo satisfies the advice you cite.

boneskull: @turadg 

> Note that there is risk to this approach even if you trust the author, because a tag can be moved or deleted if a bad actor gains access to the repository storing the action.

boneskull: I don't worry GitHub will become a bad actor, but "third party" is strictly a relative term. There's no way GitHub is _not_ a third party to me or you _unless_ you work for GitHub.

.github/workflows/ci.yml
boneskull: FWIW, if we drop Node.js v18, then all of LavaMoat is stuck until we can drop v18.

.github/workflows/ci.yml
turadg: yeah I read the text I quoted. I'm done with this debate.

mhofman: This is about a GH workflow, running on GH infra. By definition endo is trusting that infra, and whatever latest deployments there may be of it. When I said 3rd party, I meant to include GH as a 1st party since that's where the root is.

While I agree the possibility of a branch in a `actions` org repo being updated maliciously is higher than the actions workflow infra itself being updated maliciously, I still consider both extremely unlikely, and not worth defensing against.

I will not die on that hill however.

boneskull: I am dying on this hill. 😄

boneskull: (we have some [nascent tooling](https://github.com/LavaMoat/LavaMoat/tree/main/packages/git-safe-dependencies) which checks this sort of thing as well, but it needs docs explaining why it does what it does. @naugtur 😉)

.github/workflows/ci.yml
turadg: When can LavaMoat drop v18?

mhofman: I endo can continue to support Node 18 for longer than EOL. There probably should be a policy about this that endo maintainers can reference.

.github/workflows/ci.yml
mhofman: As long as I'm not the one maintaining this, I'm ok with shas here.

leotm: RE maintaining, do we plan on introducing Renovate to endo?

since i don't think dependabot can PR yml SHA updates (i could be wrong)

boneskull: IMO, we should, but others may disagree. I'll throw something on the meeting agenda.

.github/workflows/ci.yml
boneskull: @turadg Not anytime in the near future

boneskull: @mhofman @turadg I've created an agenda item for the next meeting to discuss such a policy.

kriskowal:
Regarding pinning to hashes, unless we undertake the diligence of auditing every update to our supply-chain, we must chose between two kinds of vulnerability: the kind where there is a vulnerability in our pinned version and we stay on it too long for lack of attention, or a vulnerability is introduced in a version and goes unaddressed for lack of the maintainer’s action. I find neither of these options attractive, but I think we are more likely to benefit from an action getting updated to mitigate a vulnerability by its maintainer than we are likely to benefit from standing still, unless we commit to regularly auditing the action and its supply chain.

turadg (approved):

boneskull:
I should have squashed that; my bad.

PR #2791, opened 2025-05-02
fix(pass-style): rename passable "Primitive" to ocapn "Atom"

erights:
Closes: #XXXX
Refs: #XXXX

## Description

For describing Passables that correspond to OCapN Atoms, the terms around "primitive" because increasingly awkward because not all of these are (byteArray) or will be (symbol) represented by JS language-level primitives. Instead, they are/will be represented by JS objects. Thus, we adopt the OCapN name `Atom` instead. We still export the "Primitive*" type names, but deprecate them to remove them once we're past this transition.

We also fixed a few remaining bugs where we were using the JS-level `isObject` to partition Passable types between Atom and non-Atom. We now use `isAtom` instead.

### Security Considerations

Aside from the naming clarity and the bug fixes, none.

### Scaling Considerations

none

It will now be more natural to delegate much of the `@endo/pass-style` documentation to OCapN
### Testing Considerations

- [ ] Should add test cases that fail before these bugs were fixed.

We might need to preserve a few more exported names while deprecating them, to avoid breaking clients of endo like agoric-sdk

### DIFF
--- a/packages/marshal/src/types.test-d.ts
+++ b/packages/marshal/src/types.test-d.ts
@@ -1,18 +1,14 @@
-import { expectType, expectNotType } from 'tsd';
+import { expectType } from 'tsd';
 
-import {
-  Far,
-  type PrimitiveStyle,
-  type RemotableObject,
-} from '@endo/pass-style';
+import { Far, type AtomStyle, type RemotableObject } from '@endo/pass-style';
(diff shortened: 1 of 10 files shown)


gibson042 (approved):
Seems like an improvement to me. :+1:

PR #2792, opened 2025-05-05
fix! only string named methods

erights:

Closes: #XXXX
Refs: #XXXX

## Description

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations


packages/marshal/test/marshal-capdata.test.js
```diff
@@ -68,9 +68,9 @@ test('serialize static data', t => {
   t.deepEqual(ser(-0), { body: '0', slots: [] });
   t.deepEqual(ser(-0), ser(0));
   // unregistered symbols
-  t.throws(() => ser(Symbol('sym2')), {
-    // An anonymous symbol is not Passable
-    message: /Only registered symbols or well-known symbols are passable:/,
+  t.throws(() => ser(passableSymbolForName('sym2')), {
+    message:
+      'Only unregistered symbols are passable: "[unpassableSymbolForName(sym2)]"',
```
erights, line 73: Revert error message change in this PR

packages/pass-style/src/types.d.ts
```diff
@@ -1,6 +1,32 @@
...
+ *
+ * Remotable has a prototype chain in which the penultimate object has own
+ * `{ [PASS_STYLE]: "remotable", [Symbol.toStringTag]: $iface }`, where both
+ * $tag and $iface must be strings, and the latter must either
+ * be "Remotable" or start with "Alleged: " or "DebugName: ".
+ *
+ * Selector has its own
+ * `{ [PASS_STYLE]: "selector", [Symbol.toStringTag]: description }`.
+ */
+export type PassStyleMarker = 'tagged' | 'remotable' | 'symbol';
```
erights, line 15: Not for this PR

erights: But do keep the refactoring rename from `TaggedOrRemotable` to `PassStyleMarker`

packages/pass-style/test/safe-promise.test.js
```diff
@@ -33,7 +33,7 @@ test('safe promise loophole', t => {
       {
         // Override mistake
         message:
-          "Cannot assign to read only property 'Symbol(Symbol.toStringTag)' of object '[object Promise]'",
+          "Cannot assign to read only property 'unpassableSymbolForName(Symbol.toStringTag)' of object '[object Promise]'",
```
erights, line 36: Error message wrong

packages/patterns/src/keys/compareKeys.js
```diff
@@ -58,7 +57,7 @@ harden(bagCompare);
  * A unique local value that is guaranteed to not exist in any inbound data
  * structure (which would not be the case if we used `Symbol.for`).
  */
-const ABSENT = Symbol('absent');
+const ABSENT = unpassableSymbolForName('absent');
```
erights, line 60: Revert to Symbol.for

packages/patterns/src/patterns/getGuardPayloads.js
```diff
@@ -203,10 +203,6 @@ const LegacyInterfaceGuardShape = M.splitRecord(
   {
     defaultGuards: M.or(M.undefined(), 'passable', 'raw'),
     sloppy: M.boolean(),
-    // There is no need to accommodate LegacyMethodGuardShape in
-    // this position, since `symbolMethodGuards happened
-    // after https://github.com/endojs/endo/pull/1712
-    symbolMethodGuards: M.mapOf(M.symbol(), MethodGuardShape),
```
erights, line 209: Need to figure out how to retire this with some kind of deprecation dance instead.

packages/patterns/test/copyMap.test.js
```diff
@@ -72,7 +77,7 @@ test('backwards-compatible static shape', t => {
     Object.defineProperties(
       {},
       {
-        [Symbol.for('passStyle')]: { value: 'tagged' },
+        [passableSymbolForName('passStyle')]: { value: 'tagged' },
```
erights, line 80: Revert to Symbol.for, or better, import from main (only?) definition.

packages/patterns/test/copyMap.test.js
```diff
@@ -127,7 +127,10 @@ test('getCopyMapEntries', t => {
   ]);
   t.deepEqual([...getCopyMapEntries(m)], getCopyMapEntryArray(m));
   const entriesIterable = getCopyMapEntries(m);
-  t.is(passStyleOf(entriesIterable), 'remotable');
+  t.throws(() => passStyleOf(entriesIterable), {
+    message:
+      'Remotables can only have string-named methods: "unpassableSymbolForName(Symbol.iterator)"',
```
erights, line 132: fix error message

packages/marshal/test/marshal-capdata.test.js
erights: Done in #2793

packages/pass-style/src/types.d.ts
erights: Done in #2793

packages/pass-style/test/safe-promise.test.js
erights: Done in #2793

packages/patterns/src/keys/compareKeys.js
erights: Reverted to `Symbol` in #2793 . Done

packages/patterns/test/copyMap.test.js
erights: Done in #2793

packages/patterns/test/copyMap.test.js
erights: Done

packages/exo/src/exo-tools.js
```diff
@@ -359,36 +363,20 @@ export const defendPrototype = (
...
       interfaceName,
       methodGuards: mg,
-      symbolMethodGuards,
       sloppy,
       defaultGuards: dg = sloppy ? 'passable' : defaultGuards,
     } = getInterfaceGuardPayload(interfaceGuard);
     methodGuards = harden({
       ...mg,
-      ...(symbolMethodGuards &&
-        fromEntries(getCopyMapEntries(symbolMethodGuards))),
```
erights, line 391: Need deprecation dance

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -2214,7 +2213,6 @@ export const InterfaceGuardPayloadShape = M.splitRecord(
   {
     defaultGuards: M.or(M.undefined(), 'passable', 'raw'),
     sloppy: M.boolean(),
-    symbolMethodGuards: M.mapOf(M.symbol(), MethodGuardShape),
```
erights, line 2217: Need deprecation dance

packages/patterns/src/types.js
```diff
@@ -541,8 +541,6 @@ export {};
  *   interfaceName: string,
  *   methodGuards:
  *     Omit<T, symbol> & Partial<{ [K in Extract<keyof T, symbol>]: never }>,
- *   symbolMethodGuards?:
```
erights, line 544: Need deprecation dance

packages/pass-style/test/passStyleOf.test.js
```diff
@@ -292,6 +292,7 @@ test('passStyleOf testing remotables', t => {
   }
   const farBaseProto7 = FarBaseClass7.prototype;
   t.is(getPrototypeOf(farBaseProto7), Object.prototype);
+  // TODO Fails during construction because 'constructor' is not filtered out
```
erights, line 295: This TODO is done. Remove

erights:
Putting this back into Draft until I make some changes:
- Split the PR into two staged PRs.
  1. change all remotable creation to avoid symbol-named methods
  2. enforce that remotables cannot have symbol-named methods
    -  Special carve-out to grandfather in `Symbol.asyncIterator` with its current wire representation, while enforcing that a remotable cannot have both a method named with the string `@@asyncIterator`.

packages/daemon/src/reader-ref.js
```diff
@@ -60,9 +60,6 @@ export const makeIteratorRef = iterable => {
       }
       return harden({ done: true, value: undefined });
     },
-    [Symbol.asyncIterator]() {
-      return this;
-    },
```
erights, line 65: @kriskowal , is this a non-test case? Do we need to preserve it?

packages/pass-style/src/passStyleOf.js
```diff
@@ -215,7 +215,8 @@ const makePassStyleOf = passStyleHelpers => {
   return harden(passStyleOf);
 };
 
-export const PassStyleOfEndowmentSymbol = Symbol.for('@endo passStyleOf');
+export const PassStyleOfEndowmentSymbol =
```
erights, line 218: Revert this to `Symbol.for` since it has nothing to do with passability and needs to be a registered symbol in any case.

packages/marshal/src/marshal-justin.js
```diff
@@ -315,10 +315,6 @@ const decodeToJustin = (encoding, shouldIndent = false, slots = []) => {
           assert.typeof(digits, 'string');
           return out.next(`${BigInt(digits)}n`);
         }
-        case '@@asyncIterator': {
-          // TODO deprecated. Eventually remove.
-          return out.next('Symbol.asyncIterator');
-        }
```
erights, line 321: Revive, i.e., revert the deletion.

packages/exo/src/exo-tools.js
```diff
@@ -386,14 +372,11 @@ export const defendPrototype = (
     const {
       interfaceName,
       methodGuards: mg,
-      symbolMethodGuards,
```
erights, line 389: When getting rid of `symbolMethodGuards`, need to instead accommodate `@@asyncIterator` as a reserved method guard name. That's weird.

Need deprecation dance

packages/pass-style/src/remotable.js
```diff
@@ -38,39 +47,48 @@ export const canBeMethod = func =>
...
+ * Is `key` a possible method name of a remotable? To qualify, `key` must be
+ * a string other than `then` or `constructor`.
+ *
+ * TODO How we bind an OCapN symbol "then" or "constructor" to JS remotables
+ * is still TBD. Probably some form of Hilbert hotel.
  *
  * @param {any} key
  * @returns {key is RemotableMethodName}
  */
 const canBeMethodName = key =>
```
erights, line 59: Do the whole check/is/assert pattern, and consolidate new error message in `checkMethodName`.

What should this do with `Symbol.asyncIterator` and `@@asyncIterator`?

packages/pass-style/src/remotable.js
```diff
@@ -38,39 +47,48 @@ export const canBeMethod = func =>
...
-export const getRemotableMethodNames = behaviorMethods =>
-  getMethodNames(behaviorMethods);
+export const getRemotableMethodNames = behaviorMethods => {
+  const result = getMethodNames(behaviorMethods).filter(
+    name => name !== 'constructor',
+  );
+  for (const name of result) {
+    typeof name === 'string' ||
+      Fail`${q(name)} must be a string, not ${q(typeof name)}`;
+    name !== 'then' || Fail`${q(name)} may not be a method name`;
```
erights, line 86: Delegate more to `assertMethodName`

packages/pass-style/src/remotable.js
```diff
@@ -272,9 +290,13 @@ export const RemotableHelper = harden({
...
                 )} in ${candidate}`)) &&
-              (key !== PASS_STYLE ||
+              (canBeMethodName(key) ||
+                // @ts-expect-error nothing wrong comparing string to symbol
+                key === 'constructor' ||
                 (!!check &&
-                  CX(check)`A pass-by-remote cannot shadow ${q(PASS_STYLE)}`))))
+                  CX(
+                    check,
+                  )`Remotables can only have string-named methods other than "then" and "constuctor": ${String(key)}`))))
```
erights, line 299: Delegate more to `assertMethodName`

packages/pass-style/src/symbol.js
```diff
@@ -53,8 +53,8 @@ harden(assertPassableSymbol);
  * encoding. We identify a well known symbol such as `Symbol.iterator`
  * by prefixing the property name with `"@@"`, such as `"@@iterator"`.
  * For registered symbols whose name happens to begin with `"@@"`, such
- * as `Symbol.for('@@iterator')` or `Symbol.for('@@foo')`, we identify
- * them by prefixing them with an extra `"@@"`, such as
+ * as `passableSymbolForName('@@iterator')` or `passableSymbolForName('@@foo')`,
+ * we identify them by prefixing them with an extra `"@@"`, such as
  * `"@@@@iterator"` or `"@@@@foo"`. (This is the Hilbert Hotel encoding
  * technique.)
```
erights, line 59: Fix.

packages/pass-style/src/passStyleOf.js
erights: Done

packages/marshal/src/marshal-justin.js
erights: Done

packages/exo/src/exo-tools.js
erights: Reverted in this PR

packages/pass-style/src/symbol.js
erights: Reverted in this PR

packages/pass-style/src/remotable.js
erights: Not in this PR

kriskowal:
Where I would like to land with this change is a _bijection_ where all valid method names have one and only one corresponding JavaScript reification.

I think that means:

| wire | javascript |
| -- | -- |
| `@@asyncIterator` | `Symbol.asyncIterator` |
| `constructor` | `Symbol.for('constructor')` |
| `then` | `Symbol.for('then')` |
| other string | other string |

My preference is to avoid `Symbol.iterator` entirely since the signature over the wire differs from the local signature, and assuming it’s not actually present anywhere. I am pretty sure we have no risk of breaking for other symbols, but defer to @mhofman about the course for maintaining the chain.

kriskowal (changes requested):
Requesting changes to clear out my review inbox. Positions are already stated and I agree with the broad strokes of the proposed plan to converge on the simplest practically-non-breaking single category for method names.

Ideally, method names are only strings in JavaScript memory (and OCapN symbols on the wire; for CapTP, method names are not directly passable data). But, this ideal conflicts with safety, since the method names `constructor`, `prototype`, and `then` create attack surface.

Ideally, all method names are valid on the wire. We might be able to cross-over certain string names like `constructor`, `prototype`, and `then` to `Symbol.for('constructor')`, `Symbol.for('prototype')`, and `Symbol.for('then') when declaring remotables or exos. I find it unlikely that any extant exos use these method names. I do not think we need to validate this change against chain transcripts because it should be equally satisfying to statically analyze existing makeExo.

Ideally we would just drop support for `Symbol.asyncIterator` entirely. To do so, we must validate the change against chain transcript replay. To avoid the cost of validation against chain transcript relay, we have the option of making a (permanent) legacy exception for `Symbol.asyncIterator` by similarly swapping the method name `"@@asyncIterator"` for `Symbol.asyncIterator` when converting the string on the wire to an in-memory method name and _vice versa_. Honestly, I find this wart tolerable.

erights:
> since the method names `constructor`, `prototype`, and `then` create attack surface.

Yes for `constructor` and `then`. What's the problem with a remotable method named `prototype`? (I admit it is weird, but I don't see a problem.)

PR #2794, opened 2025-05-05
chore(benchmark): Add eslint devDependencies

gibson042:
I had issues without this:
```console
$ yarn lint
command not found: eslint
```


kriskowal:
I’m in favor of adding these back, in general, but vaguely recall an initiative underway to lint only at the top of the repo. Does that ring a bell @turadg?

turadg:
> I’m in favor of adding these back, in general, but vaguely recall an initiative underway to lint only at the top of the repo. Does that ring a bell @turadg?

I've wanted to but haven't pushed for it. What I did change was that whitespace formatting is only at the top (being a repo-wide policy), but each package still controls its own `lint` command.

It's regrettable to have to repeat these deps in each package. I think a better solution is to depend on `@endo/eslint-plugin` which is the thing requiring these others. But there are trade-offs in that approach as well and it's probably not worth tackling at this time. So I'm approving.

turadg (approved):

PR #2795, opened 2025-05-06
chore: update LICENSE files

### LINKED ISSUE #2788: fix all the LICENSE files
Nearly every copy (if not all of them!) of the Apache-2.0 license (`packages/*/LICENSE`) contains:

```text
   Copyright [yyyy] [name of copyright owner]
```

...this should probably be fixed. 

- Use whatever year `endojs/endo` was created, or whatever year the specific workspace was created.
- "Endo Contributors" may be the only feasible owner

boneskull:
Fixes #2788

### DIFF
--- a/packages/ses/LICENSE
+++ b/packages/ses/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2018 Endo Contributors
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
--- a/scripts/set-license-text.sh
+++ b/scripts/set-license-text.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+
+# This script tries to figure out what year a package was created and updates
+# its `LICENSE` file accordingly (including copyright owner)
+
+set -euo pipefail
+
+declare -A years
+for pkg in ./packages/*; do
+  year=$(git --no-pager log --diff-filter=A --follow --format=%ad --date format:%Y --reverse -- "${pkg}" | head -n1) || true
+  years[$pkg]=$year
+done
+
+for i in "${!years[@]}"; do
+  license="${i}/LICENSE"
+  year="${years[$i]}"
+  if [ -f "$license" ]; then
+    sed -i '' -e "s/\[yyyy\]\ \[name\ of\ copyright\ owner\]/${year} Endo Contributors/g" "$license"
+  fi
+done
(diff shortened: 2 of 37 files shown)


scripts/set-license-text.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
boneskull, line 1: this is because macOS

scripts/set-license-text.sh
boneskull, line 1: @kriskowal drop your illest science on my bash-fu, sensei

scripts/set-license-text.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# This script tries to figure out what year a package was created and updates
+# its `LICENSE` file accordingly (including copyright owner)
+
+set -euo pipefail
+
+declare -A years
+for pkg in ./packages/*; do
+  year=$(git --no-pager log --diff-filter=A --follow --format=%ad --date format:%Y --reverse -- "${pkg}" | head -n1) || true
```
boneskull, line 10: `|| true` due to `pipefail`. `pipefail` causes the script to fail _but_ doesn't give me any information about what happened?

scripts/set-license-text.sh
```diff
@@ -0,0 +1,20 @@
...
+  years[$pkg]=$year
+done
+
+for i in "${!years[@]}"; do
+  license="${i}/LICENSE"
+  year="${years[$i]}"
+  if [ -f "$license" ]; then
+    sed -i '' -e "s/\[yyyy\]/${year}/g; s/\[name\ of\ copyright\ owner\]/Endo Contributors/g" "$license"
```
boneskull, line 18: I suppose I could concatenate these two now that I think about it

packages/ses/LICENSE
boneskull, line 1: this was the only one I set manually

scripts/set-license-text.sh
kriskowal: I’m fine with this. I assume `head` exiting is causing `git` to write to a closed file. I’m not sure there’s a clean solution for this except to unnecessarily absorb the whole log into a buffer. I like using `git hash-object -w --stdin` for slurping, but I believe there’s something you can pipe inline. The buffering conceit presumes that `git cat-file blob $HASH` exits gracefully if it fails to write to stdout, unlike `git log`.

scripts/set-license-text.sh
kriskowal, line 18: I would be delighted to have this in `scripts/create-package.sh` but I can wait until the next package gets created to make that update.

kriskowal (approved):

scripts/set-license-text.sh
boneskull: Might as well do it now, because whoever makes the next package may not be one of us.

I combined this into a single search/replace operation.

PR #2797, opened 2025-05-07
refactor(pass-style): avoid symbol-named methods

erights:
Staged on #2793 

Closes: #XXXX
Refs: #2793 #2792

## Description

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations


packages/daemon/src/reader-ref.js
```diff
@@ -60,9 +60,6 @@ export const makeIteratorRef = iterable => {
       }
       return harden({ done: true, value: undefined });
     },
-    [Symbol.asyncIterator]() {
-      return this;
-    },
```
erights, line 65: @kriskowal is this a non-test use? Do we need to preserve it?

packages/pass-style/src/iter-helpers.js
```diff
@@ -12,8 +12,14 @@ import { Far } from './make-far.js';
...
-  /** @type {Iterable<U>} */
-  Far('mapped iterable', {
+  /**
+   * An `Iterable` should not be a `Remotable` because it has a symbol-named
+   * method. That's why we use `harden` rather than, say, `Far`. However,
+   * an `Iterator` can be a `Remotable`.
+   *
+   * @type {Iterable<U>}
+   */
+  harden({
```
erights, line 22: This is used by code in agoric-sdk, but it doesn't seem like anything relies on these Iterables being remotable. 

TODO test in agoric-sdk.

packages/pass-style/src/passStyle-helpers.js
```diff
@@ -48,23 +49,6 @@ harden(isTypedArray);
...
- *
- * TODO HAZARD Because we check this on the way to hardening a remotable,
- * we cannot yet check that `func` is hardened. However, without
- * doing so, it's inheritance might change after the `PASS_STYLE`
- * check below.
- *
- * @param {any} func
- * @returns {boolean}
- */
-export const canBeMethod = func =>
```
erights, line 64: Just moved to remotable.js, as that is the least surprising home.

packages/patterns/src/keys/checkKey.js
```diff
@@ -415,7 +415,7 @@ export const getCopyMapEntries = m => {
     payload: { keys, values },
   } = m;
   const { length } = /** @type {Array} */ (keys);
-  return Far('CopyMap entries iterable', {
+  return harden({
```
erights, line 418: TODO test in agoric-sdk.

Not sure why `getCopyMapEntries` alone returns an Iterable whereas its siblings all return arrays. Probably that is what should really be fixed, though hopefully not in this PR.

We make the returned Iterable non-passable. Hopefully that is consistent with all uses in agoric-sdk. Otherwise, hopefully all uses are compat with this returning an array, which would be passable.

packages/daemon/src/reader-ref.js
kriskowal: I think this is vestigial and safe to remove. I don’t see any `E(ref)[Symbol.asyncIterator]()`.

packages/patterns/src/keys/checkKey.js
```diff
@@ -392,48 +391,14 @@ harden(getCopyMapValues);
  * @template {Key} K
  * @template {Passable} V
  * @param {CopyMap<K,V>} m
- * @returns {Array<[K,V]>}
- */
-export const getCopyMapEntryArray = m => {
```
erights, line 397: TODO test in agoric-sdk

I renamed this method with its types to `getCopyMapEntries` which is more inline with similar names for other copy collections. I deleted the old `getCopyMapEntries` which returned an iterable, because nothing in endo or agoric-sdk seems sensitive to the difference between this being only an Iterable vs it being an array that also happens to be an Iterable. The name `getCopyMapEntryArray` was only used within this package and was never exported beyond this package, so no need to deprecate. We simply removed it.

packages/patterns/src/keys/checkKey.js
erights: Done. Changes to returning an array, and removed the other name. Nothing was sensitive to whether this returns an only-Iterable vs an array which also happens to be an Iterable. The other name was never exported from this package, so could be safely removed. Done.

erights:
This is now Ready for Review. PTAL

packages/pass-style/src/types.d.ts
```diff
@@ -162,6 +162,12 @@ export type RemotableObject<I extends InterfaceSpec = string> = PassStyled<
   I
 >;
 
+/**
+ * Abstract remotable method names to its own type in preparation of restricting
+ * it in a later PR.
+ */
+export type RemotableMethodName = string | symbol;
```
kriskowal, line 169: I believe we’re in agreement that this will be narrowed to strings except "@@asyncIterator" and specifically Symbol.asyncIterator, somehow, at least at runtime if not here.

kriskowal (changes requested):
As this is not a pure refactor, I would like to fully separate the changes that have zero risk from the changes that have non-zero risk.

The changes under `daemon` have zero risk.

Changing `PropertyKey` to `RemotableMethodName` that is effectively equivalent to `PropertyKey` has almost negligible risk. I don’t expect to find any `number` keys in Exo. But, I will still want to see an integration PR pass bound to this.

Changing `getCopyMapEntries` and removing `getCopyMapEntryArray` is risky. I might be inclined to just not do that in order to avoid that risk, but I would like to be able to bisect that risk with a separate commit, maybe even a separate PR, and I need a nod from @mhofman that it won’t affect the chain.

packages/pass-style/src/types.d.ts
erights: Yes. Resolving.

erights:
> As this is not a pure refactor, I would like to fully separate the changes that have zero risk from the changes that have non-zero risk.

Done. The zero-risk changes moved to #2840. I then rebased this one on #2840 .

> 
> The changes under `daemon` have zero risk.

Moved to #2840

> 
> Changing `PropertyKey` to `RemotableMethodName` that is effectively equivalent to `PropertyKey` has almost negligible risk. I don’t expect to find any `number` keys in Exo. But, I will still want to see an integration PR pass bound to this.

The `RemotableMethodName` in #2840 is just an alias for `PropertyKey`, in keeping with the zero-risk refactor goal.

The more substantive change remains in this PR.

> 
> Changing `getCopyMapEntries` and removing `getCopyMapEntryArray` is risky. I might be inclined to just not do that in order to avoid that risk, but I would like to be able to bisect that risk with a separate commit, maybe even a separate PR, and I need a nod from @mhofman that it won’t affect the chain.

Fully omitted from #2840 . Remains here.

packages/pass-style/src/types.d.ts
erights: While we're agreed, I'm leaving this further narrowing to a later PR since this one is still not trying to enforce anything.

PR #2798, opened 2025-05-09
fix: suggestions for PR #2785

erights:

Closes: #XXXX
Refs: #XXXX

## Description

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations

### DIFF
--- a/packages/immutable-arraybuffer/index.js
+++ b/packages/immutable-arraybuffer/index.js
@@ -142,6 +142,13 @@ const {
 
 setPrototypeOf(immutableArrayBufferPrototype, arrayBufferPrototype);
 
+/**
+ * Transfer the contents to a new Immutable ArrayBuffer
+ *
+ * @param {ArrayBuffer} buffer The original buffer.
+ * @param {number} [newLength] The start index.
+ * @returns {ArrayBuffer}
+ */
 export const transferBufferToImmutable = (buffer, newLength = undefined) => {
   if (newLength !== undefined) {
     if (transfer) {
@@ -160,7 +167,8 @@ export const transferBufferToImmutable = (buffer, newLength = undefined) => {
       }
     }
   }
-  return new ImmutableArrayBufferInternal(buffer);
+  const result = new ImmutableArrayBufferInternal(buffer);
+  return /** @type {ArrayBuffer} */ (/** @type {unknown} */ (result));
 };
 
 export const isBufferImmutable = buffer => {
@@ -177,6 +185,14 @@ export const isBufferImmutable = buffer => {
   }
 };
 
+/**
+ * Enforces that `arrayBuffer` is a genuine `ArrayBuffer` exotic object.
+ *
+ * @param {ArrayBuffer} buffer
+ * @param {number} [start]
+ * @param {number} [end]
+ * @returns {ArrayBuffer}
+ */
 const sliceBuffer = (buffer, start = undefined, end = undefined) => {
   try {
     // @ts-expect-error We know it is really there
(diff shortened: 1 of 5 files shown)


packages/immutable-arraybuffer/index.js
```diff
@@ -142,6 +142,13 @@ const {
 
 setPrototypeOf(immutableArrayBufferPrototype, arrayBufferPrototype);
 
+/**
+ * Transfer the contents to a new Immutable ArrayBuffer
+ *
+ * @param {ArrayBuffer} buffer The original buffer.
+ * @param {number} [newLength] The start index.
+ * @returns {ArrayBuffer}
+ */
```
erights, line 151: I retrofitted your improved typing into the main shim, so the two parallel each other using the best of both. Thanks for these improved types.

packages/immutable-arraybuffer/index.js
```diff
@@ -160,7 +167,8 @@ export const transferBufferToImmutable = (buffer, newLength = undefined) => {
       }
     }
   }
-  return new ImmutableArrayBufferInternal(buffer);
+  const result = new ImmutableArrayBufferInternal(buffer);
+  return /** @type {ArrayBuffer} */ (/** @type {unknown} */ (result));
```
erights, line 171: I added this cast to both, so that the external typing is always simply `ArrayBuffer`. With these cast here, most messy types were no longer needed.

packages/immutable-arraybuffer/shim-hermes.js
```diff
@@ -28,7 +29,7 @@ if ('transfer' in arrayBufferPrototype) {
   );
 }
 
-if ('transferToImmutable' in arrayBufferPrototype) {
+if ('sliceToImmutable' in arrayBufferPrototype) {
```
erights, line 32: For hermes the warning would be wrong because the hermes shim would not overwrite `transferToImmutable`.

I changed both to `sliceToImmutable` because it works for both. For the main shim, `sliceToImmutable` is just as good and maintains the parallelism.

packages/immutable-arraybuffer/src/limited-pony-for-hermes.js
```diff
@@ -95,10 +75,24 @@ ImmutableArrayBufferInternal.prototype = {
...
+/**
+ * Emulates the encapsulated `ImmutableArrayBufferInternal` class constructor
+ * from the original except this function takes the `realBuffer` which its
+ * result encapsulates. Security demands that this result has exclusive access
+ * to the `realBuffer` it is given, which its callers must ensure.
+ *
+ * @param {ArrayBuffer} realBuffer
+ * @returns {ArrayBuffer}
+ */
+const makeImmutableArrayBufferInternal = realBuffer => {
```
erights, line 88: Since this instance creator is purely internal, there was no need for it to emulate a class. I just made it an array function that doesn't mention `this`. As a result, I was able to figure out how to fix the types.

packages/ses/src/get-anonymous-intrinsics.js
```diff
@@ -168,17 +168,9 @@ export const getAnonymousIntrinsics = () => {
...
-    // eslint-disable-next-line @endo/no-polymorphic-call
-    iab = ab.transferToImmutable(); // Note: Unavailable on Hermes
-  } else {
-    // @ts-expect-error TODO How do I add sliceToImmutable to ArrayBuffer type?
-    // eslint-disable-next-line @endo/no-polymorphic-call
-    iab = ab.sliceToImmutable();
-  }
+  // @ts-expect-error TODO How do I add sliceToImmutable to ArrayBuffer type?
+  // eslint-disable-next-line @endo/no-polymorphic-call
+  const iab = ab.sliceToImmutable();
```
erights, line 173: No reason for the condition. `sliceToImmutable` works fine for both.

erights (commented):
I explain the rationale for some of these differential changes as self-review comments on this differential PR.

packages/ses/src/get-anonymous-intrinsics.js
leotm: was thinking exactly this, nice

PR #2799, opened 2025-05-12
feat(marshal): passable to quasi-quoted Justin expr

erights:
Closes: #XXXX
Refs: https://github.com/Agoric/agoric-sdk/pull/5401

## Description

We often need to see the deep contents of a Passable. Indeed, this was kicked off by a slack thread about wanting to quote
```js
M.splitRecord(
    harden({
      assetKind: M.or('nat', 'set', 'copySet', 'copyBag'),
      decimalPlaces: M.and(M.gte(-100), M.lte(100)),
    }),
  )
```

This PR provides a `qp` function meaning "quote passable". You can use it where you'd use `q`. It produces a quasi-quoted Justin expression of the form in which the passable would be passed. For example, the `qp` of the above pattern is

```js
`makeTagged("match:splitRecord",  [
  {
    assetKind: makeTagged("match:or", [
      "nat",
      "set",
      "copySet",
      "copyBag",
    ]),
    decimalPlaces: makeTagged("match:and", [
      makeTagged("match:gte", -100),
      makeTagged("match:lte", 100),
    ]),
  },
])`
```

### Security Considerations

Unlike `q`, `qp` does not lazily postpone rendering the expression into Justin. `q` can do this easily because it is inside the `'ses'` assert.js module and can use the rights amplification available only there. `qp` depends on `'@endo/marshal'` which is several levels up from `'ses'`. Rather than expose these rights amplification powers (or several other kludges that failed), I just made `qp` non-lazy. Thus there are no security implications.

### Scaling Considerations

Because `qp` is not lazy, simply constructing an error with `Fail` or `makeError(X...)` will pay all the cost of rendering to Justin, whether or not that error is ever used. 

Because Justin is a subset of HardenedJS, the Justin expressions produced by `qp` can be used in documentation if needed. Though see Compatibility Considerations below.

### Testing Considerations

- [ ] Only minimal tests written for `qp`. We should also write a test that loops through all the `jsonJustinPairs`, just as other Justin tests do.

Using Justin with symbols exposes in the rendered string how Justin currently represents passable symbols. That changes in #2793. Whichever of these gets merged first, the other will have to reconcile merge conflicts. And any saved Justin strings containing that old Justin representation will need to be updated or marked stale.

### Upgrade Considerations

Other than the Compatibility Considerations above, none.

### DIFF
--- a/packages/marshal/src/marshal-justin.js
+++ b/packages/marshal/src/marshal-justin.js
@@ -1,12 +1,12 @@
 /// <reference types="ses"/>
 
+import { q, X, Fail } from '@endo/errors';
 import { Nat } from '@endo/nat';
 import {
   getErrorConstructor,
   isObject,
   passableSymbolForName,
 } from '@endo/pass-style';
-import { q, X, Fail } from '@endo/errors';
 import { QCLASS } from './encodeToCapData.js';
 import { makeMarshal } from './marshal.js';
 
@@ -468,14 +468,40 @@ export const passableAsJustin = (passable, shouldIndent = true) => {
 };
 harden(passableAsJustin);
 
+// The example below is the `patt1` test case from `qp-on-pattern.test.js`.
+// Please co-maintain the following doc-comment and that test module.
 /**
- * qp for quote passable as a quasi-quoted Justin expression.
+ * `qp` for quote passable as a quasi-quoted Justin expression.
  *
- * Modelled on `quote` from `assert.js` in `'ses'`. But uses Justin
- * instead of `bestEffortStringify`. The full name of this would have been
- * `quotePassable`. But since the `quote` from `assert.js` is always used via
- * its rename to `q`, we're skipping the rename and just
- * naming this variable `qp`.
+ * Both `q` from `@endo/errors` and this `qp` from `@endo/marshal` can
+ * be used together with `Fail`, `X`, etc from `@endo/errors` to mark
+ * a substitution value to be both
+ * - visually quoted in some useful manner
+ * - unredacted
+ *
+ * Differences:
+ * - given a pattern `M.and(M.gte(-100), M.lte(100))`,
+ *   ```js
+ *   `${q(patt)}`
+ *   ```
+ *   produces `"[match:and]"`, whereas
+ *   ```js
+ *   `${qp(patt)}`
+ *   ```
+ *   produces quasi-quotes Justin of what would be passed:
+ *   ```js
+ *   `makeTagged("match:and", [
+ *     makeTagged("match:gte", -100),
+ *     makeTagged("match:lte", 100),
+ *   ])`
+ *   ```
+ * - `q` is lazy, minimizing the cost for using it in an error that's never
+ *   logged. Unfortunately, due to layering constraints, `qp` is not
+ *   lazy, always rendering to quasi-quoted Justin immediately.
+ *
+ * Since Justin is a subset of HardenedJS, neither the name `qp` nor the
+ * rendered form need to make clear that the rendered form is in Justin rather
+ * than HardenedJS.
  *
  * @param {Passable} payload
  * @returns {StringablePayload}
--- a/packages/marshal/src/marshal-justin.js
+++ b/packages/marshal/src/marshal-justin.js
@@ -8,8 +8,13 @@ import {
 } from '@endo/pass-style';
 import { q, X, Fail } from '@endo/errors';
 import { QCLASS } from './encodeToCapData.js';
+import { makeMarshal } from './marshal.js';
 
-/** @import {Encoding} from './types.js' */
+/**
+ * @import {StringablePayload} from 'ses';
+ * @import {Passable} from '@endo/pass-style';
+ * @import {Encoding} from './types.js';
+ */
 
 const { ownKeys } = Reflect;
 const { isArray } = Array;
@@ -48,7 +53,7 @@ const makeYesIndenter = () => {
     },
     line,
     next: token => {
-      if (needSpace && token !== ',') {
+      if (needSpace && token !== ',' && token !== ')') {
         strings.push(' ');
       }
       needSpace = true;
@@ -336,7 +341,8 @@ const decodeToJustin = (encoding, shouldIndent = false, slots = []) => {
         }
         case 'tagged': {
           const { tag, payload } = rawTree;
-          out.next(`makeTagged(${quote(tag)},`);
+          out.next(`makeTagged(${quote(tag)}`);
+          out.next(',');
           decode(payload);
           return out.next(')');
         }
@@ -445,3 +451,34 @@ const decodeToJustin = (encoding, shouldIndent = false, slots = []) => {
 };
 harden(decodeToJustin);
 export { decodeToJustin };
+
+/**
+ * @param {Passable} passable
+ * @param {boolean} [shouldIndent]
+ * @returns {string}
+ */
+export const passableAsJustin = (passable, shouldIndent = true) => {
+  let slotCount = 0;
+  // eslint-disable-next-line no-plusplus
+  const convertValToSlot = val => `s${slotCount++}`;
+  const { toCapData } = makeMarshal(convertValToSlot);
+  const { body, slots } = toCapData(passable);
+  const encoded = JSON.parse(body);
+  return decodeToJustin(encoded, shouldIndent, slots);
+};
+harden(passableAsJustin);
+
+/**
+ * qp for quote passable as a quasi-quoted Justin expression.
+ *
+ * Modelled on `quote` from `assert.js` in `'ses'`. But uses Justin
+ * instead of `bestEffortStringify`. The full name of this would have been
+ * `quotePassable`. But since the `quote` from `assert.js` is always used via
+ * its rename to `q`, we're skipping the rename and just
+ * naming this variable `qp`.
+ *
+ * @param {Passable} payload
+ * @returns {StringablePayload}
+ */
+export const qp = payload => `\`${passableAsJustin(harden(payload), true)}\``;
+harden(qp);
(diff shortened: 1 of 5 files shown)


packages/marshal/src/marshal-justin.js
```diff
@@ -445,3 +451,34 @@ const decodeToJustin = (encoding, shouldIndent = false, slots = []) => {
...
+harden(passableAsJustin);
+
+/**
+ * qp for quote passable as a quasi-quoted Justin expression.
+ *
+ * Modelled on `quote` from `assert.js` in `'ses'`. But uses Justin
+ * instead of `bestEffortStringify`. The full name of this would have been
+ * `quotePassable`. But since the `quote` from `assert.js` is always used via
+ * its rename to `q`, we're skipping the rename and just
+ * naming this variable `qp`.
```
dckc, line 478: When somebody looks up `qp`, I think it's less likely that they'll want to know this sort of information about how it's built and more likely that they'll want to know how to use it. An `@example` such as the one in the PR description would be nice.

Please be sure to note that it's not lazy.

I suggest moving these notes about how it's implemented to `//` comments.

packages/patterns/test/qp-on-pattern.test.js
```diff
@@ -0,0 +1,38 @@
...
+
+// This is a continuation of a test case in marshal-justin.test.js,
+// but is placed here because it uses patterns which don't exist at
+// the `@endo/marshal` layer.
+//
+// This was the original example in a slack thread looking for a more
+// legible way to render patterns. Patterns are Passables, and all
+// Passables can kinda be rendered in Justin. Though Remotables and
+// promises will be rendered only in terms of slot numbers.
+test('qp on a pattern', t => {
```
dckc, line 14: this is another case where I wish for updoc/doctest, so that the same example can serve as an executable test and documentation.

dckc (commented):
some notes on docs

The code looks plausible. I'd have to study it (and the surrounding context) quite a bit more to be confident that it's correct.

packages/marshal/src/marshal-justin.js
erights: Much better, thanks!

Done

packages/marshal/src/marshal-justin.js
```diff
@@ -445,3 +451,60 @@ const decodeToJustin = (encoding, shouldIndent = false, slots = []) => {
...
+ * @returns {string}
+ */
+export const passableAsJustin = (passable, shouldIndent = true) => {
+  let slotCount = 0;
+  // eslint-disable-next-line no-plusplus
+  const convertValToSlot = val => `s${slotCount++}`;
+  const { toCapData } = makeMarshal(convertValToSlot);
+  const { body, slots } = toCapData(passable);
+  const encoded = JSON.parse(body);
+  return decodeToJustin(encoded, shouldIndent, slots);
```
dckc, line 467: I'm familiar with `decodeToJustin` from its use in the `agoric follow` command:
https://github.com/Agoric/agoric-sdk/blob/777e6a4cefd5c533aab3aeb224f95d5e1edbbd03/packages/agoric-cli/src/follow.js#L76

for example:

```console
$ agoric follow -B https://devnet.agoric.net/network-config :published.agoricNames.brand
[
  [
    "BLD",
    slotToVal("board0566","Alleged: BLD brand"),
  ],
  [
    "IST",
    slotToVal("board0257","Alleged: IST brand"),
  ],
  [
    "Invitation",
    slotToVal("board0074","Alleged: Zoe Invitation brand"),
  ],
...
]
```



In fact, my first reaction to this PR was: we already have `decodeToJustin`; why do we need this?

But now I see that the `makeMarshal` here is simpler than [the makeMarshal in agoric follow](
https://github.com/Agoric/agoric-sdk/blob/777e6a4cefd5c533aab3aeb224f95d5e1edbbd03/packages/agoric-cli/src/follow.js#L19). @michaelfig why does `makeCapDataToQclass` need to keep tables in both directions? Why don't we need that here?

I suspect that there's a test we should add to clarify how these are different.

At a minimum, I'd like to see a test that includes a few remotables.

packages/marshal/src/marshal-justin.js
```diff
@@ -445,3 +451,60 @@ const decodeToJustin = (encoding, shouldIndent = false, slots = []) => {
...
+
+/**
+ * @param {Passable} passable
+ * @param {boolean} [shouldIndent]
+ * @returns {string}
+ */
+export const passableAsJustin = (passable, shouldIndent = true) => {
+  let slotCount = 0;
+  // eslint-disable-next-line no-plusplus
+  const convertValToSlot = val => `s${slotCount++}`;
```
dckc, line 463: consider:

```suggestion
  const convertValToSlot = val => `s${(slotCount += 1)}`;
```

not exactly the same... using `+=` is more like `++slotCount`. But close enough?

erights: All else being equal, I would prefer that. But I'd to start counting at zero, because that would reflect what the slot indexes would be. And I don't want to initialize to `-1` because that just moves the surprise elsewhere that's not better.

erights: Resolving. But if you object, please unresolve.

packages/marshal/src/marshal-justin.js
erights: The test case I added to marshal does include two remotables and a promise: https://github.com/endojs/endo/blob/032533a6de48a9e00c21caa1f154856ca89d139b/packages/marshal/test/marshal-justin.test.js#L85

Unlike the other tests, this one is not intended to be a good example, but rather just to probe an edge case.

erights: If this test satisfies, please resolve this conversation. Thanks.

packages/marshal/src/marshal-justin.js
dckc: I don't feel all that strongly about it, but... what's the justification for this exception to `no-plusplus`? I suggest including it in the comment.

Initializing to `-1` doesn't seem that surprising to me.

packages/marshal/src/marshal-justin.js
michaelfig: `marshal-justin.js`'s `passableAsJustin` accepts a passable and fabricates the slot ids out of thin air (`s${number}`),  where that ID is basically meaningless.  It's unique for every remotable traversed by `toCapData(passable)`, so it can't be used to correlate identity by comparing the slotIDs of two remotables.

The reason for the tables in `capDataToQclass` is to preserve all the semantics of the input capData.  The information described by the `slotIDs` and `ifaces` of the input capData is passed through to the output Justin `makeSlot` calls.

It may very well be that `passableAsJustin` can't recover meaningful `slotIDs`, but @erights, I would encourage you to consider keeping a short-lived WeakMap to reuse an existing `slotID` if the same `remotable` (or even `promise`) appears elsewhere in the input passable.

I don't really remember what the second map was used for.  Humph.

michaelfig: Oh.  nvm.  Does `@endo/marshal` already track remotable identity already?  I thought that was up to `valToSlot`.

dckc (approved):
I'm satisfied.

I leave it to you to judge whether to wait for a review from @kriskowal

packages/marshal/src/marshal-justin.js
erights: I merged this PR before adding the requested comment. #2801 would correct.

packages/marshal/src/marshal-justin.js
```diff
@@ -445,3 +451,60 @@ const decodeToJustin = (encoding, shouldIndent = false, slots = []) => {
...
+  const { body, slots } = toCapData(passable);
+  const encoded = JSON.parse(body);
+  return decodeToJustin(encoded, shouldIndent, slots);
+};
+harden(passableAsJustin);
+
+// The example below is the `patt1` test case from `qp-on-pattern.test.js`.
+// Please co-maintain the following doc-comment and that test module.
+/**
+ * `qp` for quote passable as a quasi-quoted Justin expression.
```
dckc, line 474: Now that this landed, I'd expect it to appear in the list of function on https://endojs.github.io/endo/modules/_endo_marshal.html . But I don't see it. Anybody know why not?

erights: I do not.

mhofman:
> ### Security Considerations
> Unlike `q`, `qp` does not lazily postpone rendering the expression into Justin. `q` can do this easily because it is inside the `'ses'` assert.js module and can use the rights amplification available only there. `qp` depends on `'@endo/marshal'` which is several levels up from `'ses'`. Rather than expose these rights amplification powers (or several other kludges that failed), I just made `qp` non-lazy. Thus there are no security implications.
> 
> ### Scaling Considerations
> Because `qp` is not lazy, simply constructing an error with `Fail` or `makeError(X...)` will pay all the cost of rendering to Justin, whether or not that error is ever used.


My understanding is that `ses`'s `quote` does not attempt to prevent trapping by the quoted object when lazily evaluated.

Looking at `bestEffortStringify` is there a reason it couldn't look for an own `toString` property and invoke that if it exists?

In that case, `qp` could simply create an object that does what this PR does when its `toString` method is invoked, to recover the lazy aspect.

erights:
That's a good idea. I actually tried to do something similar and gave up because, for some reason, it did not occur to me to modify `bestEffortStringify` itself. Without mod, I noticed that it looked for a `Symbol.toString`-named property, and I tried to shoehorn the behavior into a getter for that property. Doing this with a simple non-accessor `toString` seems good.

If I do it, it'll be in PR separate from #2801 . Or, if anyone else wants to try, I'll be happy to review.

mhofman:
> Without mod, I noticed that it looked for a `Symbol.toString`-named property, and I tried to shoehorn the behavior into a getter for that property.

Yeah a `{get [Symbol.toStringTag]() { return qp(payload) }}` would work too, but be a little more hacky.

PR #2801, opened 2025-05-15
doc(marshal): adds missing comment from #2799 review

erights:
Closes: #XXXX
Refs: https://github.com/endojs/endo/pull/2799#discussion_r2085652080

## Description

At https://github.com/endojs/endo/pull/2799#discussion_r2085652080 @dckc suggested
>  I suggest including it in the comment.

I intended to. But somehow I resolved it before actually doing this, causing me to miss it when evaluating if I should just merge-and-squash #2799. This PR corrects that omission.

Minor internal comment only. It should not have any other implications.

### DIFF
--- a/packages/marshal/src/marshal-justin.js
+++ b/packages/marshal/src/marshal-justin.js
@@ -459,6 +459,9 @@ export { decodeToJustin };
  */
 export const passableAsJustin = (passable, shouldIndent = true) => {
   let slotCount = 0;
+  // Using post-increment below only so that the indexes start at zero
+  // and the `slotCount` variable can be initialized to `0` rather than
+  // `-1`.
   // eslint-disable-next-line no-plusplus
   const convertValToSlot = val => `s${slotCount++}`;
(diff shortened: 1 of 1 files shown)


mhofman:
Curious about your thoughts on https://github.com/endojs/endo/pull/2799#issuecomment-2885163254 as well if you're following up.

erights:
> Curious about your thoughts on [#2799 (comment)](https://github.com/endojs/endo/pull/2799#issuecomment-2885163254) as well if you're following up.

At https://github.com/endojs/endo/pull/2799#issuecomment-2885223290

> That's a good idea. I actually tried to do something similar and gave up because, for some reason, it did not occur to me to modify `bestEffortStringify` itself. Without mod, I noticed that it looked for a `Symbol.toString`-named property, and I tried to shoehorn the behavior into a getter for that property. Doing this with a simple non-accessor `toString` seems good.
>
> If I do it, it'll be in PR separate from #2801 . Or, if anyone else wants to try, I'll be happy to review.

dckc (approved):

Issue #2094, opened 2024-02-21
Module transform cannot export value extracted from default


kriskowal:
## Describe the bug

Endo’s bundler cannot transform an ESM where an exported value was previously obtained from extraction of an object, where the default value was previously extracted. Or rather: this module can’t be bundled:

```js
const { x, y = x } = z;
export { x, y };
```

## Steps to reproduce

1. In `packages/module-source` run the following failing tests:

```js
test.failing('should support export of defaulted extraction', t => {
  new ModuleSource(`
    const { x, y = x } = globalThis;
    export { x, y };
  `)
});

test.failing('should support export const object with default', t => {
  const _ = new ModuleSource(`
    export const {
      x = undefined,
    } = globalThis;
  `);
  t.pass();
});
```

## Expected behavior

Above test should pass and effect consistent behavior to other spec-compliant implementations.

PR #2802, opened 2025-05-15
fix(module-source): allow defaults in destructuring assignments

boneskull:
Closes: #2094

## Description

Modifies `babel-plugin.js` to allow for defaults when destructuring.

The problem noted in #2094 was affecting _all_ such assignments, not just those which are exported.

I'm _extremely suspicious_ of this change because it was so easy that it _must_ be wrong.

### Security Considerations

If there is a security-related reason for not implementing this, I don't know what it is.

### Scaling Considerations

no

n/a

### Testing Considerations

Could throw this in a test fixture, I suppose.

Well, it allows for parsing code containing destructuring assignments with defaults, so take that as you will.

### Upgrade Considerations

May want a `NEWS.md` entry? I'm honestly surprised somebody isn't complaining about this already.

### DIFF
--- a/packages/module-source/src/babel-plugin.js
+++ b/packages/module-source/src/babel-plugin.js
@@ -31,6 +31,8 @@ const collectPatternIdentifiers = (path, pattern) => {
         // Non-elided pattern.
         return collectPatternIdentifiers(path, pat);
       });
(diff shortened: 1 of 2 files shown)


kriskowal (approved):
Huzzah.

Issue #2096, opened 2024-02-21
Implement an isPassable that does not rely on try/catch.


erights:
@agoric/base-zone has an implementation of `isPassable` that emulates the proper behavior using a try/catch to turn the thrown failure of `passStyleOf` into a `return false;`.

As of https://github.com/endojs/endo/pull/2042 , that is duplicated into @endo/pass-style , as a first step of migrating `isPassable` from @agoric/base-zone into @endo/pass-style. Once synced, we should also delete the one in @agoric/base-zone and change its users to import from @endo/pass-style instead.

`passStyleOf` should be rewritten into the form of our other `checkFoo` functions to take a `Checker` argument, so it can be used to create either an `assertPassable` that throws or an `isPassable` that never throws but rather returns false iff `assertPassable` would have thrown. `passStyleOf` would then be like `assertPassable` except that the success case also returns a string.

This will require a significant refactoring of `passStyleOf`, which is already performance critical, so we'll need to be careful.

PR #2803, opened 2025-05-16
fix(pass-style): isAtom much accept non-passables

erights:

Closes: #XXXX
Refs: #XXXX

## Description

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations


packages/pass-style/src/typeGuards.js
```diff
@@ -107,16 +107,17 @@ export const assertRemotable = (
...
-  // If that happens, we should call it here instead. As it is now,
-  // any non-passable will be rejected by `passStyleOf` with a thrown
-  // error.
-  const passStyle = passStyleOf(val);
+  let passStyle;
+  try {
+    passStyle = passStyleOf(val);
+  } catch (err) {
+    return check(false, X`Not even Passable: ${q(err)}: ${val}`);
+  }
```
kriskowal, line 120: At least on V8, this will be considerably faster if we can avoid the try/catch. Probably considerably more understandable if `passStyleOf` simply did not throw, or if there were a variant that did not throw.

gibson042: Hmm, I also see opportunities for this and other functions to use `canBeValid` and other fast checks to avoid the full `passStyleOf` traversal. Let's leave this for now and open an issue.

gibson042 (approved):
The JSDoc for `checkAtom` and/or `isAtom` should probably mention that they don't throw for non-passable input (and for that matter, the other functions in this file should mention that they do, even if we're willing to change that in the future).

packages/pass-style/src/typeGuards.js
erights: Yeah, that's been a longstanding TODO. Not for this PR.

Resolving.

packages/pass-style/src/typeGuards.js
```diff
@@ -107,16 +107,17 @@ export const assertRemotable = (
...
-  // TODO There is not yet a checkPassable, but perhaps there should be.
-  // If that happens, we should call it here instead. As it is now,
-  // any non-passable will be rejected by `passStyleOf` with a thrown
-  // error.
-  const passStyle = passStyleOf(val);
+  let passStyle;
+  try {
+    passStyle = passStyleOf(val);
+  } catch (err) {
+    return check(false, X`Not even Passable: ${q(err)}: ${val}`);
```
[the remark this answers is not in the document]
erights: I don't understand the change to the error message. This is the case where `val` is not even passable, and so there is no `passStyle`.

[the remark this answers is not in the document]
erights: Just added the `check !== identChecker &&` but kept the original error message. Resolving.

erights:
>  Let's leave this for now and open an issue.

https://github.com/endojs/endo/issues/2096

PR #2804, opened 2025-05-16
fix(ses): avoid depth-first loading sequence in loadNow to prevent getting stuck in cycles

naugtur:
## Description

It took a while, but we've encountered  case that would first go into a dynamic require (therefore trigger compartment-mapper using importNow) and proceed to import a codebase with a cycle in cjs modules.

The minimal implementation of enqueueJob that we used in importNow was causing the synchronous case for `memoizedLoadWithErrorAnnotation` to become depth-first and getting into cycles before anything got successfully memoized.

The new `syncJobQueue` guarantees a breadth-first order. 

### Security Considerations

none

### Scaling Considerations

`current` and `next` in `syncJobQueue` get swapped and each layer of the module dependency tree gets processed as a separate step, which should be reasonably memory efficient. I could eliminate recursion from `drainQueue` altogether, but that seems a premature optimization.

It's a fix

### Testing Considerations

The testcase is minimal and gets to stack limit fairly quickly when failing. 

I noticed signatures of `import` and `importNow` differ. I don't recall us making that decision, but `await import` gives us `{ namespace }` while `importNow` gives us `namespace` directly. Is that some future-proofing I don't remember?

### Upgrade Considerations

N/A

### DIFF
--- a/packages/ses/src/module-load.js
+++ b/packages/ses/src/module-load.js
@@ -507,11 +507,9 @@ const memoizedLoadWithErrorAnnotation = (
   return moduleLoading;
 };
 
-const asyncJobQueue = () => {
+const asyncJobQueue = (errors = []) => {
   /** @type {Set<Promise<undefined>>} */
   const pendingJobs = new Set();
-  /** @type {Array<Error>} */
-  const errors = [];
 
   /**
    * Enqueues a job that starts immediately but won't be awaited until drainQueue is called.
@@ -530,8 +528,6 @@ const asyncJobQueue = () => {
   };
   /**
    * Sequentially awaits pending jobs and returns an array of errors
-   *
-   * @returns {Promise<Array<Error>>}
    */
   const drainQueue = async () => {
     await null;
@@ -539,9 +535,33 @@ const asyncJobQueue = () => {
       // eslint-disable-next-line no-await-in-loop
       await job;
     }
-    return errors;
   };
-  return { enqueueJob, drainQueue };
+  return { enqueueJob, drainQueue, errors };
+};
+
+const syncJobQueue = (errors = []) => {
+  let current = [];
+  let next = [];
+  const enqueueJob = (func, args) => {
+    arrayPush(next, [func, args]);
+  };
+  const drainQueue = () => {
+    // Attention: load bearing flow order. Calling another enqueued function in the
+    // synchronous usecase must happen after the one that enqueued it has finished.
+    // Jobs enqueued in one pass do not interleave with jobs resulting from them.
+    // It's necessary for efficient memoization and to break cycles.
+    for (const [func, args] of current) {
+      try {
+        func(...args);
+      } catch (error) {
+        arrayPush(errors, error);
+      }
+    }
+    current = next;
+    next = [];
+    if (current.length > 0) drainQueue();
+  };
+  return { enqueueJob, drainQueue, errors };
 };
 
 /**
@@ -588,7 +608,7 @@ export const load = async (
   /** @type {Map<object, Map<string, Promise<Record<any, any>>>>} */
   const moduleLoads = new Map();
 
-  const { enqueueJob, drainQueue } = asyncJobQueue();
+  const { enqueueJob, drainQueue, errors } = asyncJobQueue();
 
   enqueueJob(memoizedLoadWithErrorAnnotation, [
     compartmentPrivateFields,
@@ -600,8 +620,7 @@ export const load = async (
     moduleLoads,
   ]);
 
-  // Drain pending jobs queue and throw an aggregate error
-  const errors = await drainQueue();
+  await drainQueue();
 
   throwAggregateError({
     errors,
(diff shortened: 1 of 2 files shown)


packages/ses/test/import-cjs.test.js
```diff
@@ -607,6 +607,65 @@ test('synchronous loading via importNowHook', async t => {
...
+  const compartment = new Compartment({
+    resolveHook: resolveNode,
+    importHook: async a => importNowHookWithCyclicDep(a),
+    importNowHook: importNowHookWithCyclicDep,
+    __options__: true,
+  });
+
+  let namespace;
+  t.notThrows(() => {
+    namespace = compartment.importNow('./a.cjs');
```
naugtur, line 662: The different shape returned from `import` and `importNow` got me a little worried. Is that deliberate? @kriskowal

packages/ses/test/import-cjs.test.js
```diff
@@ -607,6 +607,65 @@ test('synchronous loading via importNowHook', async t => {
   compartment.importNow('./main.js');
 });
 
+const importNowHookWithCyclicDep = specifier => {
```
kriskowal, line 610: Please avoid abbreviation. `importNowHookWithCycle` or `importNowHookWithCyclicDependency`.

packages/ses/test/import-cjs.test.js
kriskowal: The shape difference is deliberate, but we are converging on this one with the `__noNamespaceBox__` option on the `Compartment` constructor. The namespace box made it possible to avoid calling the `then` method of a module namespace object when it passes through `import`, which is analogous to dynamic `import`. Module Harmony decided that we can’t solve this problem for dynamic import so there is not much to gain from solving it for async import on modules. Synchronous `importNow` never had this concern.

kriskowal (approved):

packages/ses/test/import-cjs.test.js
naugtur: makes sense. Thanks

boneskull:
@kriskowal IMO we need more coverage of dynamic requires. 

It's tempting to write a harness that builds CJS test fixtures using both static and dynamic require to attempt to future-proof new fixtures... but will probably end up duplicating fixtures and search/replace.

I'll make a ticket for this, anyway.

packages/ses/src/module-load.js
```diff
@@ -507,11 +507,9 @@ const memoizedLoadWithErrorAnnotation = (
   return moduleLoading;
 };
 
-const asyncJobQueue = () => {
+const asyncJobQueue = (errors = []) => {
```
boneskull, line 510: Here are a bunch of types you can use

```suggestion
/**
 * @template {readonly any[]} [Args=unknown[]]
 * @callback SyncJob
 * @param {Args} args
 * @returns {void}
 */

/**
 * @template {readonly any[]} [Args=unknown[]]
 * @callback AsyncJob
 * @param {Args} args
 * @returns {Promise<void>}
 */

/**
 * @template {readonly any[]} [Args=unknown[]]
 * @callback EnqueueSyncJobFn
 * @param {SyncJob<Args>} func
 * @param {Args} [args]
 * @returns {void}
 */
 
/**
 * @template {readonly any[]} [Args=unknown[]]
 * @callback EnqueueAsyncJobFn
 * @param {AsyncJob<Args>} func
 * @param {Args} [args]
 * @returns {void}
 */

/**
 * @typedef SyncJobQueueResult
 * @property {EnqueueSyncJobFn} enqueueJob
 * @property {() => void)} drainQueue
 * @property {Error[]} errors
 */

/**
 * @typedef AsyncJobQueueResult
 * @property {EnqueueAsyncJobFn} enqueueJob
 * @property {() => Promise<void>} drainQueue
 * @property {Error[]} errors
 */

/**
 * @param {Error[]} errors 
 * @returns {AsyncJobQueueResult}
 */
const asyncJobQueue = (errors = []) => {
```

and for `syncJobQueue`,

```js
/**
 * @param {Error[]} errors 
 * @returns {SyncJobQueueResult}
 */
const syncJobQueue = (errors = []) => {
```

It doesn't look like SES compiles declarations from `.ts` files, so in `@typedef`s they go.

When defining `enqueueJob` in both of `syncJobQueue` and `asyncJobQueue`, prepend `/** @type {EnqueueSyncJobFn} */` and `/** @type {EnqueueAsyncJobFn} */`, respectively (there is already a docstring for the `enqueueJob` function in `asyncJobQueue` which can be changed or replaced).

packages/ses/test/import-cjs.test.js
```diff
@@ -607,6 +607,65 @@ test('synchronous loading via importNowHook', async t => {
...
+  throw Error(`Cannot load module for specifier ${specifier}`);
+};
+
+test('import handles a cycle in CommonJS modules', async t => {
+  t.plan(4);
+  const compartment = new Compartment({
+    resolveHook: resolveNode,
+    importHook: async a => importNowHookWithCyclicDep(a),
+    importNowHook: importNowHookWithCyclicDep,
+    __options__: true,
```
boneskull, line 638: Out of curiosity, what does this do?

boneskull (approved):

boneskull:
my merge finger is itchy

packages/ses/test/import-cjs.test.js
naugtur: It's the new compartment options syntax. Compartment used to accept 3 objects in the original signature. This is a gradual migration thing.

naugtur:
> my merge finger is itchy

Was gonna merge it right away but you added the suggestion to add types just before I ran out of workweek ;)

PR #2806, opened 2025-05-17
fix(compartment-mapper): handle invalid peer dependencies

boneskull:
## Description

_May be controversial!_

This adds dependency names found in `peerDependenciesMeta` which are otherwise unspecified in other dependency fields to the list of all dependencies so that a `CompartmentDescriptor` will be created if the package exists.

~~We will prefer any spec already provided in a dependency field instead of the default (which is `*` for lack of a better value).~~

~~A variation on this PR would be to use the version from `devDependencies` if present (regardless of the `dev` flag) and then fall back to `*`. _Most_ packages tend to include `peerDependencies` in `devDependencies`.~~

> [!IMPORTANT]
>
> I noticed that `graphPackage()` was storing the specification for each dependency but this data was unused. I changed the data structures to avoid doing this.

### Security Considerations

n/a

### Scaling Considerations

no

I don't think this needs any documentation changes.

### Testing Considerations

If someone wants it, I can add a test for this. I'd rather pretend it doesn't exist.

no

### Upgrade Considerations

I can't see how this would break anyone who isn't already broken.

### DIFF
--- a/packages/compartment-mapper/src/node-modules.js
+++ b/packages/compartment-mapper/src/node-modules.js
@@ -403,7 +403,20 @@ const graphPackage = async (
   const dependencyLocations = {};
   /** @type {ReturnType<typeof gatherDependency>[]} */
   const children = [];
+
+  /**
+   * A set containing dependency names which are considered "optional"
+   */
   const optionals = new Set();
+
+  /**
+   * Contains the names of _all_ dependencies
+   *
+   * @type {Set<string>}
+   */
+  const allDependencies = new Set();
+
+  // these are fields from package.json containing dependencies
   const {
     dependencies = {},
     peerDependencies = {},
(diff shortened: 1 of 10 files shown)


boneskull:
I should mention why this pattern exists:

npm (post-wide-usage) and pre-v6.x did not install `peerDependencies` by default.  npm v7 introduced `peerDependenciesMeta` and also introduced the behavior of automatically installing `peerDependencies`.  While it's pointless to add `peerDependenciesMeta` (maybe it works in yarn or smth) _without_ `peerDependencies`, there's no notion of an _optional_ peer dep prior to npm v7.  nvm v6 was shipped w/ Node.js v10.x thru v14.x and npm v8 in LTS versions thereafter.

kriskowal (changes requested):
Sounds good to me. Please add a test.

kriskowal (approved):

PR #2808, opened 2025-05-19
chore(ses): types for import job queues

naugtur:
A sad attempt at putting types where the expressive power of JavaScript is being used.

Getting `args` as a single argument array and then spreading into a function is hard for TypeScript to believe.

I have a strong preference in favor of https://github.com/endojs/endo/pull/2811 for what we get out of it.

### DIFF
--- a/packages/ses/src/module-load.js
+++ b/packages/ses/src/module-load.js
@@ -507,11 +507,9 @@ const memoizedLoadWithErrorAnnotation = (
   return moduleLoading;
 };
 
-const asyncJobQueue = () => {
+const asyncJobQueue = (errors = []) => {
   /** @type {Set<Promise<undefined>>} */
   const pendingJobs = new Set();
-  /** @type {Array<Error>} */
-  const errors = [];
 
   /**
    * Enqueues a job that starts immediately but won't be awaited until drainQueue is called.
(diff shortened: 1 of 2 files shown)


packages/ses/src/module-load.js
```diff
@@ -507,18 +507,72 @@ const memoizedLoadWithErrorAnnotation = (
...
   /**
    * Enqueues a job that starts immediately but won't be awaited until drainQueue is called.
    *
-   * @template {any[]} T
-   * @param {(...args: T)=>Promise<*>} func
-   * @param {T} args
+   * @type {EnqueueAsyncJobFn}
    */
   const enqueueJob = (func, args) => {
+    if (!isArray(args)) {
```
naugtur, line 573: Why do I need to suffer runtime performance hits to convince TS an array is still an array?

packages/ses/src/module-load.js
```diff
@@ -507,18 +507,72 @@ const memoizedLoadWithErrorAnnotation = (
...
+ * @param {SyncJob<Args>} func
+ * @param {readonly [...Args]} args // the only way to make the types match
+ * @returns {void}
+ */
+
+/**
+ * @template {readonly any[]} [Args=unknown[]]
+ * @callback EnqueueAsyncJobFn
+ * @param {AsyncJob<Args>} func
+ * @param {readonly [...Args]} args // the only way to make the types match
```
naugtur, line 536: Note that despite using templates, if I mess up the arguments array ([here](https://github.com/endojs/endo/pull/2808/files#diff-f9990051733b98f01f6918095f0f2d4acaa3885a93f71dbafeeefc83ec2b8137R678)) to not match the type expected by `func` I still get no warning :(

PR #2809, opened 2025-05-19
chore: update deprecated github action and pin versions, add validation

naugtur:
## Description

A long deprecated (and now deleted) version of actions/deploy-pages was used. 

I ran `@lavamoat/git-safe-dependencies` and fixed errors.

Proposing it as a lint step.

Additionally, setting up a Renovate github bot to update the pinned versions could be useful.

### Security Considerations

Pinned versions are not prone to changes from authors, but the repository is only using actions from `actions/*` repositories, which means if github folks push an update to a relevant major, they update automatically, so I'm opening this up to discussion whether we want to pin or not.

`@lavamoat/git-safe-dependencies` verifies that the pinned commit ID belongs to the original repository not a fork.

### Upgrade Considerations

there was a breaking change to deploy-pages, but documentation indicates it should not affect us.

### DIFF
--- a/.github/workflows/browser-test.yml
+++ b/.github/workflows/browser-test.yml
@@ -26,8 +26,8 @@ jobs:
     steps:
       - name: Put unstable chrome where playwright would look for it
         run: mv /opt/google/chrome /opt/google/chrome-unstable
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v3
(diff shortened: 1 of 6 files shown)


boneskull:
@naugtur See #2790. I've also added an agenda item for Weds meeting to consider using Renovate

naugtur:
> @naugtur See #2790. I've also added an agenda item for Weds meeting to consider using Renovate

Should I move the validation there?

PR #2811, opened 2025-05-19
chore(ses): add typings to job queues to ensure arguments match

naugtur:
Much less types but better ROI than https://github.com/endojs/endo/pull/2808

Internals of job queues are not typed precisely, but what this gives us and previous didn't is TS complaining when 
`enqueueJob(memoizedLoadWithErrorAnnotation, [ wrong, argz, go, here ]`

### DIFF
--- a/packages/ses/src/module-load.js
+++ b/packages/ses/src/module-load.js
@@ -514,9 +514,10 @@ const asyncJobQueue = (errors = []) => {
   /**
    * Enqueues a job that starts immediately but won't be awaited until drainQueue is called.
    *
-   * @template {any[]} T
-   * @param {(...args: T)=>Promise<*>} func
-   * @param {T} args
+   * @template {(...args: any[]) => Promise<void>} F
+   * @param {F} func - An async function to execute
+   * @param {Parameters<F>} args - Arguments to pass to the function
+   * @returns {void}
    */
   const enqueueJob = (func, args) => {
     setAdd(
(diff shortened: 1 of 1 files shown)


packages/ses/src/module-load.js
```diff
@@ -514,9 +514,10 @@ const asyncJobQueue = (errors = []) => {
   /**
    * Enqueues a job that starts immediately but won't be awaited until drainQueue is called.
    *
-   * @template {any[]} T
-   * @param {(...args: T)=>Promise<*>} func
-   * @param {T} args
+   * @template {(...args: any[]) => Promise<any>} F
```
boneskull, line 517:
```suggestion
   * @template {(...args: any[]) => Promise<void>} F
```

packages/ses/src/module-load.js
```diff
@@ -542,6 +543,15 @@ const asyncJobQueue = (errors = []) => {
 const syncJobQueue = (errors = []) => {
   let current = [];
   let next = [];
+
+  /**
+   * Enqueues a job
+   *
+   * @template {(...args: any[]) => any} F
```
boneskull, line 550:
```suggestion
   * @template {(...args: any[]) => void} F
```

boneskull (approved):

packages/ses/src/module-load.js
boneskull: my reasoning is that we do not want to encourage the use of a return value in any of these "jobs" because it will be ignored

PR #2812, opened 2025-05-19
chore(skel): Improve the state of nascent packages

gibson042:
## Description

Document package creation in CONTRIBUTING.md and update files in package "skel" so scripts/create-package.sh creates a directory that requires minimal boilerplate editing.

### Security Considerations

n/a

### Scaling Considerations

n/a

This could be extended to cover new-package _publishing_ as well, but I'll have to lean on @kriskowal for those details.

### Testing Considerations

Verified manually to create a new package for an upcoming PR.

n/a

### Upgrade Considerations

n/a

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,6 +11,12 @@ Endo is a yarn workspaces repository. Running yarn in the root will install and 
 Note: running yarn `--ignore-scripts` will not complete the setup of SES.
 Note: Endo uses `lerna` only for releasing. `lerna bootstrap` is unlikely to work.
 
+## Creating a new package
+
+Run <code>[scripts/create-package.sh](./scripts/create-package.sh) $name</code>,
+then update the resulting README.md, package.json (specifically `description`),
+index.js, and index.test.js files.
+
 ## Rebuilding `ses`
 
 Changes to `ses` require a `yarn build` to be reflected in any dependency where `import 'ses';` appears. Use `yarn build` under `packages/ses` to refresh the build.
--- a/scripts/create-package.sh
+++ b/scripts/create-package.sh
@@ -45,7 +45,7 @@ NEWPKGJSONHASH=$(
   } + . + {
     name: "@endo/\($name)",
     version: "0.1.0",
-    homepage: (.homepage // "https://github.com/endojs/endo/tree/master/packages/\($name)#readme"),
+    homepage: "https://github.com/endojs/endo/tree/master/packages/\($name)#readme",
     repository: (.repository + { directory: "packages/\($name)" }),
     scripts: ((.scripts // {}) | to_entries | sort_by(.key) | from_entries),
     dependencies: ((.dependencies // {}) | to_entries | sort_by(.key) | from_entries),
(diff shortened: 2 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -11,6 +11,12 @@ Endo is a yarn workspaces repository. Running yarn in the root will install and
 Note: running yarn `--ignore-scripts` will not complete the setup of SES.
 Note: Endo uses `lerna` only for releasing. `lerna bootstrap` is unlikely to work.
 
+## Creating a new package
+
+Run <code>[scripts/create-package.sh](./scripts/create-package.sh) $name</code>,
+then update the resulting README.md, package.json (specifically `description`),
+index.js, and index.test.js files.
```
boneskull, line 18:
```suggestion
then update the resulting `README.md`, `package.json` (specifically `description`),
`index.js`, and `index.test.js` files.
```

scripts/create-package.sh
```diff
@@ -55,5 +55,19 @@ NEWPKGJSONHASH=$(
...
-sed -i '' -e "s/\[yyyy\]\ \[name\ of\ copyright\ owner\]/$(date '+%Y') Endo Contributors/g" "packages/$NAME/LICENSE"
+# update other files in place
+cd "packages/$NAME"
+BSD_SED="$(sed --help 2>&1 | sed 2q | grep -qe '-i ' && echo 1 || true)"
+function sedi() {
+  if [ -n "$BSD_SED" ]; then
+    sed -i '' "$@"
+  else
+    sed -i "$@"
+  fi
```
boneskull, line 66: I thought `sed -i ''` did the same thing in both versions, but it was required in BSD?

scripts/create-package.sh
```diff
@@ -55,5 +55,19 @@ NEWPKGJSONHASH=$(
...
+}
+# CHANGELOG.md: remove `skel` content
+sedi -ne '/###/q; p' CHANGELOG.md
```
boneskull, line 69: for my own edification, what does `q` do and what does `p` do?

boneskull (approved):
couple questions and a suggestion; no blockers

scripts/create-package.sh
gibson042: With GNU sed, [any argument to `-i` is optional](https://www.gnu.org/software/sed/manual/sed.html#:~:text=%2Di%5BSUFFIX,place%5B%3DSUFFIX%5D) and must therefore immediately follow it with no intervening whitespace (e.g., the `''` in `sed -i ''` is a _positional_ argument unrelated to `-i`). With BSD sed, [`-i` _requires_ an argument](https://man.openbsd.org/sed#i) and therefore claims the rest of the current token if non-empty or otherwise the following token (e.g., the `''` in `sed -i ''` is associated with `-i`). So while e.g. `sed -i.bak -e ...` means "edit in place with `.bak`-suffixed backups" for both, I'm not aware of any uniform invocation that means "edit in place with no backups"—hence this shell function.

scripts/create-package.sh
gibson042: `q` is "quit", `p` is "print", and `-n` disables automatic printing of the pattern space at end-of-script (cf. [How sed Works](https://www.gnu.org/software/sed/manual/sed.html#Execution-Cycle)). So this invocation edits CHANGELOG.md in place, quitting on the first line containing "###" without printing it but printing every line before that—i.e., truncating the copied contents of [skel/CHANGELOG.md](https://github.com/endojs/endo/blob/master/packages/skel/CHANGELOG.md?plain=1) right before the first "###".

CONTRIBUTING.md
gibson042: Formatting file names as code generally seems a bit off to me. We can revisit later, but for now I'll be leaving this as-is.

scripts/create-package.sh
boneskull: I see.  `sed -i.bak -e s/foo/bar/ $file && rm $file.bak` seems portable 😄

PR #2815, opened 2025-05-20
feat(panic): ponyfill `panic` from "Don't Remember Panicking" proposal

erights:
Closes: #XXXX
Refs: https://github.com/Agoric/agoric-sdk/pull/11173 https://github.com/tc39/proposal-oom-fails-fast 

## Description

Adds `@endo/panic` as an imperfect ponyfill of the `panic` function from the [Don't Remember Panicking](https://github.com/tc39/proposal-oom-fails-fast) tc39 draft proposal. Please see the `README.md` in `@endo/panic` for example use and details.

### Security Considerations

Felipe Natal reported a security weakness in Agoric's handling of alleged all-or-nothing transactions, especially in the context of transfer or erights. While we do not know of any practical way for an outside party to trigger a throw during the execution of the "commit phase" (see the README.md), if such an error could be induced in the absence of a reliable `panic` operation, execution might proceed with the transfer partly done, causing either the disappearance of erights or their duplication, both of which would violate conservation.

This `panic` ponyfill is imperfect because it cannot reliably and portably immediately exit on all platforms. However, it provides a hook that `@agoric/swingset/liveslots` will use so that `panic` in a vat running on SwingSet will reliably immediately exit, preserving transactional integrity for that environment.

But in general, the hazard will still remain.

### Scaling Considerations

Under SwingSet, we'd like a vat `panic` to immediately terminate the vat incarnation but only suspend the vat, so that the vat can be upgraded from a state that still had transactional integrity to recover and continue correct execution.

We should rely on the draft tc39 proposal to provide most of the needed documentation for the "real" `panic`.

The README.md of `@endo/panic` does provide adequate documentation of how the ponyfill differs from the "real" one.

### Testing Considerations

It's hard to write an ava test that tests immediate exit, since (in the same agent) there is no longer an ava to observe that the `exit` happened. Instead, this PR finds adequate indirect measures to test whether this ponyfill does what it is supposed to do -- as described in its README.

Just introducing the ponyfill should have no compat considerations. Later PRs that start using it, such as draft https://github.com/Agoric/agoric-sdk/pull/11173, may well have their own compat issues. Looking forward, as the tc39 proposal evolves, we'd like to evolve the ponyfill to best match it. Once the proposal settles down, we'd like to add a shim to the ponyfill to emulate the settled proposal.

### Upgrade Considerations

To support vat upgrade under liveslots, as noted, liveslots will need to react to any fatal termination including `panic` by terminating only the vat incarnation, suspending the vat, enabling the vat to upgrade and continue from a consistent state in a new incarnation. Liveslots does not yet have the support needed for such vat suspension.


packages/ses-ava/test/panic.test.js
```diff
@@ -0,0 +1,73 @@
+/* global globalThis */
+
+// `@endo/panic` is `ses` and `@endo/ses-ava` in package layering, so its test
```
erights, line 3:
```suggestion
// `@endo/panic` is below `ses` and `@endo/ses-ava` in package layering, so its test
```

packages/panic/index.js
```diff
@@ -0,0 +1,59 @@
...
+
+  if (typeof globalThis[PanicEndowmentSymbol] === 'function') {
+    /** @type {never} */ (globalThis[PanicEndowmentSymbol](err));
+  } else if (
+    globalThis.process &&
+    typeof globalThis.process.abort === 'function'
+  ) {
+    globalThis.process.abort();
+  } else {
+    // TODO Once Moddable provides a host `panic` function, use that here.
```
kriskowal, line 51: I assume we would test for globalThis.panic first.

packages/panic/index.js
```diff
@@ -0,0 +1,59 @@
...
+  }
+  throw lastResortError;
```
kriskowal, line 53: Or `for(;;)` in some environments?

erights:
Note to self: Preserve first two commits so https://github.com/endojs/endo/issues/2816 makes sense.

erights:
@kriskowal , why can't I add @phoddie as a reviewer?

@phoddie, please consider yourself a reviewer anyway. I look forward to your comments!

packages/panic/package.json
```diff
@@ -0,0 +1,69 @@
+{
+  "name": "@endo/panic",
+  "version": "0.1.0",
+  "private": true,
```
kriskowal, line 4: I’m fine with merging now since this is marked as not-to-be-published, and I like waiting until we have a concrete use to test in integration before committing to an API/behavior. If you’ve got one in mind now, let’s flip this bit.

kriskowal (approved):
We might consider holding this

packages/panic/index.js
erights: @phoddie, if you do provide us with a host `panic` function before the proposal advances in tc39, where do you expect to put it?

@kriskowal , yes, from wherever @phoddie suggests. Fortunately, this can wait for a later PR since XS does not yet provide one.

packages/panic/index.js
erights: From the README:

> (As noted in the proposal, a higher fidelity emulation could, as a last resort, go into an infinite loop. But the consequences of this are too painful for both manual and CI testing. Besides, on some engines (browsers), in violation of the current JS spec, resume execution of user-code within the agent after the "infinite" loop exceeds a timeout. So even this strategy would not be safe on such engines.)

The last resort is what happens if the user has forgone the first two opportunities to advise the ponyfill. If the user does neither, how should the ponyfill decide between throwing vs an infinite loop?

phoddie:
> @phoddie, if you do provide us with a host panic function before the proposal advances in tc39, where do you expect to put it?

My feeling is that we should have xsnap provide this `globalThis.panic`. xsnap is a host so it can (and already does) add globals. (If you have a global you'd prefer to hang it off, that's do-able too)

It doesn't need to be in the engine here. The operation is essentially the same as the meter overflowing – exiting all the way out – so that can be a model for implementing here. If and when this advances in TC39, we can look at moving that into XS.

packages/panic/package.json
erights: I do! Repairing the ERTP safety issue reported by Felipe Natal. Draft at https://github.com/Agoric/agoric-sdk/pull/11173 I will fix once agoric-sdk is synced with an endo with this ponyfill.

kriskowal:
> @kriskowal , why can't I add @phoddie as a reviewer?

Evidently, @phoddie was not in the Committers role. That appears to be the current convention. I’ve made the requisite benedictions and conscripted them for the review.

kriskowal:
> > @phoddie, if you do provide us with a host panic function before the proposal advances in tc39, where do you expect to put it?
> 
> My feeling is that we should have xsnap provide this `globalThis.panic`. xsnap is a host so it can (and already does) add globals. (If you have a global you'd prefer to hang it off, that's do-able too)

That sounds right to me too. This would be a binding for `fxAbort`?

packages/panic/index.js
erights: Done, for `globalThis.panic`. See thread starting at https://github.com/endojs/endo/pull/2815#issuecomment-2896059277

packages/panic/index.js
erights: Resolving with this unchanged. If there is another test the ponyfill could do to decide, it could get added in a later PR.

packages/panic/package.json
erights: Aha! I missed this. Adding to https://github.com/endojs/endo/issues/2816#issuecomment-2896090497

erights: Done. In light of #11173 I changed it to `false`.

PR #2818, opened 2025-05-21
chore: add Renovate config

boneskull:
This adds a Renovate config file. It is not boilerplate; I ended up tailoring it specifically to `endojs/endo`.

It uses JSON5 because the other option is JSON, and I find comments to be very helpful for Renovate configs.


boneskull:
@kriskowal I imagine Dependabot is disabled in the repo settings, but to make [vulnerability alerts](https://docs.renovatebot.com/configuration-options/#vulnerabilityalerts) work, it needs to be configured properly. I don't know if that means the `.github/dependabot.yml` needs to be removed.

boneskull:
(Note to self: this will likely mean conflicts with #2790, so I should probably revert the GHA action pinning)

.github/renovate.json5
```diff
@@ -0,0 +1,62 @@
+{
+  $schema: 'https://docs.renovatebot.com/renovate-schema.json',
+  description: 'Renovate config for endojs/endo',
+  extends: [
+    // only pin dev deps
```
kriskowal, line 5: I think we’re in accord on why we shouldn’t pin deps. We’re also in agreement that pinning devDeps doesn’t expose dependent packages to Eval Twin hazards the way pinning deps would. I gather that this shifts the calculus for pinning devDeps, returning the question to whether to *passively* rely on the latest version before breaking changes to *actively* relying on the latest version after deliberately merging a Renovation.

I stress that this only makes sense if we intend to promptly and thoroughly review Renovations, all the way down to the edges, and I do not believe we have enough people involved to do that work. My preference is to passively rely on the latest versions in the non-breaking range and use automation to at least explore the option of bumping major versions for us.

PR #2822, opened 2025-05-22
feat(cache-map): Promote makeLRUCacheMap into a new package

gibson042:
## Description

* Introduce `@endo/cache-map`.
* Remove "LRU" eviction policy constraint (but keep that as the actual current implementation).
* Support opting in to strongly-held keys (to be used in [agoric-sdk](https://github.com/Agoric/agoric-sdk/tree/master) for caching string-keyed board mappings).
* Add a `getMetrics()` facet for evaluating cache efficiency (currently rudimentary).

### Security Considerations

This package sits beneath SES, and so must be defensive without relying upon `harden`/`assert`/etc.

### Scaling Considerations

The implementation remains a doubly-linked ring with permanent head and LRU eviction policy. All operations are O(1), mutating up to five nodes for promotion/insertion/removal (the target node and its former and new neighbors) but minimizing that when possible (e.g., bailing out early in `moveCellAfter` when the target is already in the right position). Cells are created up to the specified capacity and never destroyed, although their single-entry weak maps are necessarily replaced when evicting a tail in `set`. Cells are currently created lazily, although that could be made eager for better memory locality (probably with a boolean `lazy` option for opting out). All in all, it's probably about as good as an object-based implementation can be (and a bit better than the `makeLRUCacheMap` from which it is derived).

Includes a README.md with explanation and example usage.

### Testing Considerations

Existing tests have been ported and expanded. A future improvement could add GC assertions.

This continues to support SES, just now as a package that can additionally be used elsewhere.

### Upgrade Considerations

None known; this is a new package.


packages/cache-map/types.d.ts
```diff
@@ -0,0 +1 @@
+export * from './src/cachemap.js';
```
gibson042, line 1: Am I doing this right?

packages/cache-map/src/cachemap.js
```diff
@@ -0,0 +1,322 @@
...
+  cell.data.delete(oldKey);
+};
+
+const zeroMetrics = freeze({
+  totalQueryCount: 0,
+  totalHitCount: 0,
+  // TODO?
+  // * method-specific counts
+  // * liveTouchStats/evictedTouchStats { count, sum, mean, min, max }
+  //   * p50/p90/p95/p99 via Ben-Haim/Tom-Tov streaming histograms
```
gibson042, line 147: A former colleague referred to count/sum/mean/min/max/p50/p90/p95/p99/p99.9 as the "big ten" metrics[^1].

[^1]: I believe [this](https://www.oracle.com/cloud/streaming/faq/#monitoring) is the result of that influence.

kriskowal: They’re pretty ubiquitous, to be sure.

packages/cache-map/src/cachemap.js
```diff
@@ -0,0 +1,321 @@
...
+ * @param {CacheMapCell<K, V>} prev
+ * @param {number} id
+ * @param {WeakMapAPI<K, V>} data
+ * @returns {CacheMapCell<K, V>}
+ */
+const appendNewCell = (prev, id, data) => {
+  const next = prev?.next;
+  const cell = { id, next, prev, data };
+  if (prev) prev.next = cell;
+  if (next) next.prev = cell;
```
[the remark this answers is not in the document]
gibson042: Hmm, I think this was left over from refactoring. Fixed.

gibson042:
> Please change the version number to 1.0.0. When this lands, I can do a one-off publish so Agoric integration CI will continue without interruption.

Done. @kriskowal this should be ready to merge, which I'd like to leave in your hands for that initial publish.

packages/cache-map/test/index.test.js
```diff
@@ -139,6 +139,96 @@ test('makeCacheMapKit (strong)', testMakeCacheMapKit, {
   keys: [{}, Symbol('unique symbol'), 'string key'],
 });
 
+test('makeCacheMapKit(1)', t => {
```
mhofman, line 142: Nit: I would love to see a deletion test in the 1 capacity case too.

packages/cache-map/test/index.test.js
```diff
@@ -139,6 +139,96 @@ test('makeCacheMapKit (strong)', testMakeCacheMapKit, {
...
+
+  // Overwrite key1.
+  cache.set(key1, 'xxx');
+  expectMetrics.totalQueryCount += 1;
+  expectMetrics.totalHitCount += 1;
+  assertNoEntry(key2);
+  assertEntry(key1, 'xxx');
+});
+
+test('makeCacheMapKit(2)', t => {
```
mhofman, line 191: I don't think capacity 2 is any more special than 3+, but I'd love to see `0` as from what I can tell it's supported, but should have constant cache miss behavior.

mhofman: I missed that the test is titled `2` but it actually tests `0`.

gibson042: Fixed.

packages/cache-map/test/index.test.js
gibson042: Added.

mhofman (approved):

packages/cache-map/src/cachemap.js
```diff
@@ -0,0 +1,321 @@
...
+      cell.data = makeMap();
+      return;
+    }
+    // Map instances can always be cleared for reuse.
+    if (size === 0) return;
+    if (size !== 1) {
+      throw Error('internal: Unexpected non-singular cell data entry count');
+    }
+    const cellData = /** @type {Map<K, V>} */ (cell.data);
+    oldKey = /** @type {K} */ (cellData.keys().next().value);
```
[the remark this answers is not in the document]
gibson042: I originally had something like that, and it took me a minute to remember why I switched away so I added a comment explaining the paranoia. But I definitely want to use `delete(oldKey)` rather than `clear()`, primarily to use the bare minimum of the non-weak Map API—to which end, I've now made explicit that SingleEntryMap extends WeakMapAPI by at most `size` and `keys`.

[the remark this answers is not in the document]
mhofman: Thinking more of it, I would prefer if the minimum WeakMap-ish API was an optional `clear()` instead of a `keys()`. The former is technically implementable by a something that keeps weak keys, where the latter isn't.

Was there a reason to prefer `keys` over `clear` ?

packages/cache-map/types.d.ts
kriskowal: I’ve commandeered the branch and aligned this with the new policy.

packages/cache-map/package.json
```diff
@@ -0,0 +1,145 @@
+{
+  "name": "@endo/cache-map",
+  "version": "1.0.0",
+  "description": "bounded-size caches having WeakMap-compatible methods",
+  "keywords": [],
```
boneskull: might wanna add smth here

packages/cache-map/package.json
```diff
@@ -0,0 +1,145 @@
...
+          "@endo/no-polymorphic-call": "off"
+        }
+      }
+    ]
+  },
+  "ava": {
+    "files": [
+      "test/**/*.test.*"
+    ],
+    "timeout": "2m"
```
boneskull, line 87: I realize this is probably the default in `skel`, but I imagine this could be significantly reduced 😄

packages/cache-map/types.d.ts
kriskowal: And it transpires that the `types.d.ts` was doing a better job of exporting the types than the implicit `index.d.ts` so I have gone back to @gibson042’s original solution in hopes that we can find a better resolution eventually.

packages/cache-map/test/index.test.js
```diff
@@ -0,0 +1,264 @@
+// @ts-check
```
boneskull:
```suggestion
```

not needed; it's already configured to be typechecked

packages/cache-map/test/index.test.js
```diff
@@ -0,0 +1,264 @@
...
+import { makeCacheMapKit } from '../index.js';
+
+/**
+ * @param {import('ava').ExecutionContext} t
+ * @param {object} config
+ * @param {any} [config.makeMap]
+ * @param {string} config.expectTag
+ * @param {[unknown, unknown, unknown]} config.keys
+ * @param {unknown[]} [config.badKeys]
+ */
```
boneskull: This should be a docstring on the function definition on L15, not on the macro itself. 
   
suggestions:

- Might be nicer to just use `any` instead of `unknown` to avoid extra type assertions
- Use `@import {x} from 'y'` instead of `import(y).x`
- `ExecutionContext` is generic on a context object, so you might want to use `@template [T=unknown]` / `ExecutionContext<T>`
- There are a _lot_ of assertions here; consider using `t.plan()`

packages/cache-map/src/cachemap.js
```diff
@@ -0,0 +1,323 @@
...
+ * @template T
+ * @typedef {T extends object ? { -readonly [K in keyof T]: T[K] } : never} WritableDeep
+ *   Intentionally limited to local needs; refer to
+ *   https://github.com/sindresorhus/type-fest if insufficient.
+ */
+
+/**
+ * @template T
+ * @param {T} value
+ * @param {<U,>(name: string, value: U) => U} [reviver]
```
boneskull, line 29: What's this extra comma do?

Is the reviver really just an identity fn?

packages/cache-map/src/cachemap.js
```diff
@@ -0,0 +1,323 @@
...
+  deepCopyJsonable(value, (_name, subvalue) => freeze(subvalue));
+
+/**
+ * A cache of bounded size, implementing the WeakMap interface but holding keys
+ * strongly if created with a non-weak `makeMap` option of
+ * {@link makeCacheMapKit}.
+ *
+ * @template K
+ * @template V
+ * @typedef {Pick<Map<K, V>, Exclude<keyof WeakMap<WeakKey, *>, 'set'>> & {set: (key: K, value: V) => WeakMapAPI<K, V>}} WeakMapAPI
```
boneskull, line 51: Where is `WeakKey` defined?

packages/cache-map/src/cachemap.js
```diff
@@ -0,0 +1,323 @@
...
+    }
+  })(options.makeMap ?? WeakMap);
+  const tag =
+    /** @type {any} */ (makeMap()).size === undefined
+      ? 'WeakCacheMap'
+      : 'CacheMap';
+
+  /** @type {WeakMapAPI<K, CacheMapCell<K, V>>} */
+  const keyToCell = makeMap();
+  // @ts-expect-error this sentinel head is special
```
boneskull, line 208: make a type for it?

packages/cache-map/src/cachemap.js
```diff
@@ -0,0 +1,323 @@
...
+    if (cell) keyToCell.set(key, cell);
+
+    // eslint-disable-next-line no-use-before-define
+    return implementation;
+  };
+  freeze(set);
+
+  // "delete" is a keyword.
+  const { delete: deleteEntry } = {
+    /** @type {(key: K) => boolean} */
```
boneskull: if the API is built within `implementation` instead of piecemeal beforehand, we could avoid some type assertions.  or, alternatively, use the properties of `WeakMapAPI` (e.g., `WeakMapAPI['delete']`) to avoid typing things twice

packages/cache-map/src/cachemap.js
```diff
@@ -132,6 +132,9 @@ const resetCell = (cell, oldKey, makeMap) => {
     if (size !== 1) {
       throw Error('internal: Unexpected non-singular cell data entry count');
     }
+    // Manually run the Iterator interface to ensure reading oldKey from an
+    // IteratorResult (rather than defaulting to undefined if the iteration is
+    // empty).
```
mhofman: Wouldn't the manual usage of the iterator result also get us `undefined` if the iteration is empty, as it's the likely value when we have a `done: true` result/

packages/cache-map/src/cachemap.js
mhofman: In the stdlib. 

[Base ES5](https://github.com/microsoft/TypeScript/blob/b504a1eed45e35b5f54694a1e0a09f35d0a5663c/src/lib/es5.d.ts#L1672) has `object` and [ES2023 extends](https://github.com/microsoft/TypeScript/blob/b504a1eed45e35b5f54694a1e0a09f35d0a5663c/src/lib/es2023.collection.d.ts#L2) that with `symbol` (Symbol as WeakMap key proposal)

packages/cache-map/src/cachemap.js
gibson042: Building `implementation` in one shot is awkward with the need to manually freeze, but I did like the idea to DRY out method signatures by referencing WeakMapAPI.

packages/cache-map/src/cachemap.js
gibson042: I'd rather not; this is a singular exception.

packages/cache-map/src/cachemap.js
gibson042: The comma is necessary in some contexts to differentiate type annotations from valid code, and I tend to use it even when not strictly necessary. As for the reviver callback itself, this function requires it to have identity _typing_ but not necessarily identity _behavior_ (and in fact `deepCopyAndFreezeJsonable` passes a freeze function that satisfies the former but not the latter).

packages/cache-map/package.json
gibson042: I'm open to ideas.

packages/cache-map/src/cachemap.js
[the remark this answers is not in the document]
gibson042: Okay, I'm convinced. Done.

packages/cache-map/test/index.test.js
gibson042: I took the middle two bullet-point suggestions, but moving the docstring produced a lot of errors that I'd rather not deal with.

gibson042:
@kriskowal this should again be ready to merge, possibly pending re-review from @boneskull.

packages/cache-map/package.json
boneskull: "cache", "lru", "weakmap", etc. 😄

packages/cache-map/test/index.test.js
boneskull: I'll take a look after merge

boneskull (approved):

packages/cache-map/package.json
```diff
@@ -0,0 +1,148 @@
...
+    "lib",
+    "src",
+    "tools"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "eslintConfig": {
+    "extends": [
+      "plugin:@endo/internal"
```
kriskowal:
```suggestion
      "plugin:@endo/ses"
```

packages/cache-map/package.json
```diff
@@ -0,0 +1,148 @@
...
+        "Date",
+        "Error",
+        "EvalError",
+        "Float32Array",
+        "Float64Array",
+        "Function",
+        "HandledPromise",
+        "Int16Array",
+        "Int32Array",
+        "Int8Array",
```
kriskowal: etc, can be removed

Issue #2819, opened 2025-05-22
compartment-mapper: fix optional dependency handling


boneskull:
In order to handle optional dependencies, `importNowHook` needs to mimic what `importHook` does w/r/t `deferError` such that exceptions occur at time of execution instead of during module loading.

PR #2826, opened 2025-05-23
fix(compartment-mapper): handle missing optional packages in dynamically-required modules

boneskull:
Closes: #2819

## Description

This adds support for deferred errors to the `importNowHook` implementation. In practice, that means _dynamically-required packages may now attempt to load missing optional dependencies_ without blowing up at load time.

To avoid algorithm divergence, I created a factory function `makeDeferError()` which returns a `deferError()` function that both `makeImportHook()` and `makeImportNowHook()` use. `makeImportNowHook` now requires a `shouldDeferError` option, like `makeImportHook`.

### Security Considerations

n/a

### Scaling Considerations

n/a

n/a

### Testing Considerations

I added a test

I don't believe `makeImportNowHook` is a public API, so there should be no compatibility issues.

### Upgrade Considerations

This is just a bugfix.

### DIFF
--- a/packages/compartment-mapper/src/import-hook.js
+++ b/packages/compartment-mapper/src/import-hook.js
@@ -33,7 +33,10 @@
  *   ParseResult,
  *   ReadFn,
  *   ReadPowers,
- *   ReadNowPowers
+ *   ReadNowPowers,
+ *   StrictlyRequiredFn,
+ *   CompartmentSources,
+ *   DeferErrorFn
  * } from './types.js'
  */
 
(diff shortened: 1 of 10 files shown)


packages/compartment-mapper/src/import-hook.js
```diff
@@ -33,7 +33,10 @@
  *   ParseResult,
  *   ReadFn,
  *   ReadPowers,
- *   ReadNowPowers
+ *   ReadNowPowers,
+ StrictlyRequiredFn,
+ CompartmentSources,
+ DeferErrorFn
```
kriskowal, line 39: formatting

packages/compartment-mapper/src/types/internal.ts
```diff
@@ -141,14 +141,15 @@ export type ChooseModuleDescriptorParams = {
    * Should be false for archives and bundles, but true for runtime.
    */
   sourceMapHook?: SourceMapHook;
-  /**
-   * Function returning a set of module names (scoped to the compartment) whose
-   * parser is not using heuristics to determine imports.
-   */
```
kriskowal, line 147: Do we deem this doc redundant? Are IDEs passing thru to the doc on StrictlyRequiredFn?

kriskowal (approved):
Please fixup nit and consider my question about the lost doc.

packages/compartment-mapper/src/types/internal.ts
boneskull: We can either keep the description in two places or not.  I figured we probably shouldn't.  This description will not be shown in the context of the options bag, no; only when you view info for `StrictlyRequiredFn`.

boneskull: I don't think there's any way to reference the `StrictlyRequiredFn` description other than using the `@inheritDoc` tag for TypeDoc; the TS language server does not understand this tag.

boneskull: ...and given this is an internal type, we won't get any TypeDoc output for it _anyway_, so I think it's moot.

Issue #2820, opened 2025-05-22
ses: opt-out of error aggregation


boneskull:
Aggregation of errors during module loading invalidations a common pattern in CJS w/r/t optional deps:

```js
try {
  require('foo')
} catch (err) {
  if (err.code === 'ERR_MODULE_NOT_FOUND') {
    // recover
  }
  throw err
}
```

However, aggregation is otherwise useful for identifying _all_ module loading issues (not just the first encountered).  

An option should be added to `ses` (where?) to opt-out of error aggregation during `Compartment.import` / `Compartment.importNow` (for consistency, the flag should affect both).

PR #2832, opened 2025-05-28
feat(ses): add noAggregateLoadErrors flag

boneskull:
Refs: #2820

## Description

This adds the `noAggregateLoadErrors` flag to the `Compartment` constructor in its `CompartmentOptions`. The default value of this flag is `false`, which retains the current behavior.

If `true`, this will cause the first error encountered during load (`load` or `loadNow`) to be thrown or rejected.

> ~~A second PR (or subsequent commit in this PR) will thread this flag through `@endo/compartment-mapper` where it can more easily be provided (e.g., to `importLocation()`, `loadFromMap()`, etc.).~~
> This won't be strictly necessary for `@lavamoat/node`'s use-case, so I'd rather just kick the can down the road until it's needed.

### Security Considerations

none that I'm aware of

### Scaling Considerations

n/a

The `aggregateLoadErrors` flag should be documented w/ the rest of `CompartmentOptions`.

### Testing Considerations

This has tests which cover both `load` and `loadNow` cases.

non-breaking

### Upgrade Considerations

### DIFF
--- a/packages/ses/NEWS.md
+++ b/packages/ses/NEWS.md
@@ -14,8 +14,8 @@ User-visible changes in `ses`
   - Unlike `ArrayBuffer` and `SharedArrayBuffer` this shim's ArrayBuffer-like object cannot be transfered or cloned between JS threads.
   - Unlike `ArrayBuffer` and `SharedArrayBuffer`, this shim's ArrayBuffer-like object cannot be used as the backing store of TypeArrays or DataViews.
   - The shim depends on the platform providing either `structuredClone` or `Array.prototype.transfer`. Node <= 16 and provides neither, causing the shim to fail to initialize, and therefore SES to fail to initialize on such platforms.
-  - Current Hermes has even stronger constraints, lacking `structuredClone`, `transfer`, private fields, and even `class` syntax. This requires other coping strategies. See https://github.com/endojs/endo/pull/2785
-  - Even after the upcoming `transferToImmutable` proposal is implemented by the platform, the current code will still replace it with the shim implementation, in accord with shim best practices. See https://github.com/endojs/endo/pull/2311#discussion_r1632607527 . It will require a later manual step to delete the shim or have it avoid overriting a platform implementation, after manual analysis of the compat implications.
+  - Current Hermes has even stronger constraints, lacking `structuredClone`, `transfer`, private fields, and even `class` syntax. This requires other coping strategies. See <https://github.com/endojs/endo/pull/2785>
+  - Even after the upcoming `transferToImmutable` proposal is implemented by the platform, the current code will still replace it with the shim implementation, in accord with shim best practices. See <https://github.com/endojs/endo/pull/2311#discussion_r1632607527> . It will require a later manual step to delete the shim or have it avoid overriting a platform implementation, after manual analysis of the compat implications.
 
 - The [evalTaming](https://github.com/endojs/endo/blob/master/packages/ses/docs/lockdown.md#evaltaming-options)
   option `'safe-eval'` now can only throw error `SES_DIRECT_EVAL`.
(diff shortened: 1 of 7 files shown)


packages/ses/src/module-load.js
```diff
@@ -597,18 +615,24 @@ const throwAggregateError = ({ errors, errorPrefix }) => {
...
 const preferAsync = (asyncImpl, _syncImpl) => asyncImpl;
 
-/*
+/**
  * `load` asynchronously gathers the module records for a module and its
  * transitive dependencies.
  * The module records refer to each other by a reference to the dependency's
  * compartment and the specifier of the module within its own compartment.
  * This graph is then ready to be synchronously linked and executed.
+ * @param {WeakMap<Compartment, any>} compartmentPrivateFields
```
boneskull, line 624: `any` means "TODO: type this properly"

packages/ses/src/module-load.js
```diff
@@ -523,7 +528,11 @@ const asyncJobQueue = (errors = []) => {
     setAdd(
       pendingJobs,
       promiseThen(func(...args), noop, error => {
-        arrayPush(errors, error);
+        if (aggregateErrors) {
+          arrayPush(errors, error);
+        } else {
+          throw error;
+        }
```
naugtur, line 535: While this implementation is correct AFAICT, I would like to analyze the difference in control flow between this and putting the error in `errors` and shutting down the queue. Is there any work that would be disrupted less/later in that case? 

I don't see any specific differences right now, but I see the potential for things happening in different order between the two.

packages/ses/src/module-load.js
```diff
@@ -564,7 +578,11 @@ const syncJobQueue = (errors = []) => {
       try {
         func(...args);
       } catch (error) {
-        arrayPush(errors, error);
+        if (aggregateErrors) {
+          arrayPush(errors, error);
+        } else {
+          throw error;
+        }
```
naugtur, line 585: (could) if we had 2 implementations of drainQueue conditional on aggregateErrors, we could skip the try-catch entirely and make it more performant and less complex in debugging. Is it worth having more code for?

packages/ses/test/import-now-hook.test.js
```diff
@@ -465,3 +466,75 @@ test('module map hook precedes import now hook', t => {
...
+      return new ModuleSource(
+        `
+        import meaning from './meaning.mjs';
+        t.is(meaning, 42);
+      `,
+        'https://example.com/main.js',
+      );
+    }
+    throw Error(`Cannot load module for specifier ${specifier}`);
+  };
```
naugtur: Would be nice to extract this outside the test cases to clearly state both use identical `importNowHook` and the behavior change is elsewhere.

packages/ses/src/module-load.js
boneskull: Well, it would be different because we would not throw the original error; we would throw the aggregate-ish `TypeError` (why isn't this an `AggregateError`, anyway?) instead.

packages/ses/src/module-load.js
boneskull: This is a thing that I naturally consider doing but don't because I am worried people will hate it.

I couldn't tell you how an engine would optimize either case.  I feel like throwing errors negatively impacts any potential optimization, so maybe if one of these functions did not ever (intentionally) throw, it would be faster.  But we're buried deep in closures, so. 🤷 

If it is an optmization, my hunch is that it won't be especially impactful.

boneskull:
@kriskowal I've resolved conflicts.

packages/ses/NEWS.md
```diff
@@ -35,6 +35,14 @@ User-visible changes in `ses`
 
   Also `ses/hermes` can now be hooked into bundlers such as Metro to run Hardened JS.
 
+- The `Compartment` constructor now accepts a `boolean` option, `aggregateLoadErrors`, to control how module-loading errors are reported.
+
+  By default, its value is `true`, which retains the previous behavior (it causes all relevant errors to be collected and rejected or thrown in a single exception from `compartment.import()` or `compartment.importNow()`, respectively).
```
kriskowal: There’s a pretty strong precedent in the language demanding that boolean options have a default of `false`, like property descriptors, which continues through `__noNamespaceBox__`. Please consider an affirmative option name like `throwFirstLoadError`.

boneskull: Ah, yeah, I can swing that.

PR #2840, opened 2025-06-03
refactor(pass-style): refactor-only avoid symbol-named methods

erights:
Extracted from #2797 

Closes: #XXXX
Refs: #2793 #2792

## Description

This PR is the extraction from #2797 of only the safe pure refactor elements of #2797 as suggested by @kriskowal at https://github.com/endojs/endo/pull/2797#pullrequestreview-2865819668 . #2797 is then rebased on this one, representing only the elements whose refactor-purity was questionable.

This PR should be a pure refactor, to avoid creating remotables with symbol-named methods when there is clearly no possible compat hazard. Non-remotable objects with symbol-named methods are fine.  But see Compatibility Considerations below.

### Security Considerations

none
### Scaling Considerations

none

none
### Testing Considerations

Most of the remotable objects with symbol-named methods were in tests. These can clearly be fixed without any compat hazard, so I count these as pure refactors. However, in removing the symbol-ness of some of these tests, I may have accidentally removed the purposes of the tests. They might now be fully redundant with other tests and should be removed. If anyone does investigate this, that cleanup could be in a later PR.

none. But see Compatibility Considerations of #2797 

### Upgrade Considerations

none

### DIFF
--- a/packages/pass-style/src/remotable.js
+++ b/packages/pass-style/src/remotable.js
@@ -36,10 +36,10 @@ export const canBeMethod = func =>
   typeof func === 'function' && !(PASS_STYLE in func);
 harden(canBeMethod);
 
+// TODO https://github.com/endojs/endo/issues/2884
+// Abstract out canBeMethodName so later PRs agree on method name restrictions.
+
 /**
- * Abstract out this test so
- * (TODO) a later PR can restrict possible method names
- *
  * @param {any} key
  * @returns {key is RemotableMethodName}
  */
--- a/packages/pass-style/src/types.d.ts
+++ b/packages/pass-style/src/types.d.ts
@@ -177,8 +177,11 @@ export type RemotableObject<I extends InterfaceSpec = string> = PassStyled<
 
 /**
  * Abstract remotable method names to its own type in preparation of
- * (TODO) restricting it in a later PR.
+ * restricting it in a later PR.
  */
+// TODO https://github.com/endojs/endo/issues/2884#issuecomment-3063896482
+// Restrict this type, and to use this restricted type rather than `PropertyKey`
+// to type method names of Remotables.
 // export type RemotableMethodName = string | symbol;
 export type RemotableMethodName = PropertyKey;
 


packages/pass-style/src/remotable.js
```diff
@@ -13,12 +13,56 @@ import {
...
+ * @param {any} func
+ * @returns {func is CallableFunction}
+ */
+export const canBeMethod = func =>
+  typeof func === 'function' && !(PASS_STYLE in func);
+harden(canBeMethod);
+
+/**
+ * Abstract out this test so
+ * (TODO) a later PR can restrict possible method names
```
kriskowal: TODO should not appear in docstring. TODO should embed tracking issue URL.

packages/pass-style/src/remotable.js
```diff
@@ -13,12 +13,56 @@ import {
...
+ *
+ * @param {any} key
+ * @returns {key is RemotableMethodName}
+ */
+const canBeMethodName = key =>
+  // typeof key === 'string' || typeof key === 'symbol';
+  typeof key === 'string' || typeof key === 'symbol' || typeof key === 'number';
```
kriskowal, line 48: This is a pure refactor so no change recommended, but do you agree we’re moving in the direction toward:

```js
const canBeMethodName = key => {
  if (typeof key === 'string') {
    return key !== '@@asyncIterator' &&
      key !== 'constructor' &&
      key !== 'then';
  } else if (typeof key === 'symbol') {
    return key === Symbol.asyncIterator ||
      key === Symbol.for('constructor') ||
      key === Symbol.for('then');
  } else {
    return false;
  }
}
```

I’ve left out `number` because I’m not sure it’s needed and might interfere with the invariant that `methodName1 === methodName2` for every pair of method names that converge on a single representation on the wire.

erights: Yes. Resolving.

packages/pass-style/src/remotable.js
erights: Done

PR #2841, opened 2025-06-03
fix(marshal): fix consequence of typo

erights:
## Description

The typo was that in a NEWS.md file, I said "Nest release" instead of "Next release". The consequence is that the release that happened since then missed this one, so it is too late to fix it simply by revising to "Next release". Instead, I manually revised it to what I imagine the release process would have renamed it to.


kriskowal (approved):

kriskowal:
Updated release notes https://github.com/endojs/endo/releases/tag/ses%401.13.0

PR #2842, opened 2025-06-04
chore(ses): add missing types to main compartment implementation

boneskull:
## Description

This adds some types to `src/compartment.js` which were missing; specifically it creates a type for the bag of private fields associated with a `Compartment` via the `privateFields` `WeakMap`.

I've attempted to infer the types, but I am probably wrong in a place or two. I do not know which of these private fields are _required_ vs which are _optional_ (except for a couple obvious ones).

I also removed a triple-slash directive that was no longer necessary (afaict).

### Security Considerations

n/a

### Scaling Considerations

n/a

n/a

### Testing Considerations

n/a

This would break compatibility with TS < v4.0 due to the use of [labeled tuples](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#labeled-tuple-elements), but I don't believe this is a concern.

### Upgrade Considerations

n/a


packages/ses/src/compartment.js
```diff
@@ -106,27 +138,20 @@ const compartmentImportNow = (compartment, specifier) => {
...
-  /**
-   * @param {string} source is a JavaScript program grammar construction.
-   * @param {object} [options]
-   * @param {Array<import('./lockdown-shim').Transform>} [options.transforms]
-   * @param {boolean} [options.sloppyGlobalsMode]
-   * @param {object} [options.__moduleShimLexicals__]
-   * @param {boolean} [options.__evadeHtmlCommentTest__]
-   * @param {boolean} [options.__evadeImportExpressionTest__]
-   * @param {boolean} [options.__rejectSomeDirectEvalExpressions__]
-   */
```
boneskull, line 129: This was duplicated in `types.d.ts` and I figured it was better to use that as the source of truth.

packages/ses/types.d.ts
```diff
@@ -167,6 +167,9 @@ export interface CompartmentOptions {
   modules?: Map<string, ModuleDescriptor>;
   __shimTransforms__?: Array<Transform>;
   __noNamespaceBox__?: boolean;
+  loadHook?: (specifier: string) => Promise<ModuleDescriptor>;
+  loadNowHook?: (specifier: string) => ModuleDescriptor;
+  __native__?: boolean;
```
boneskull: These are only used in `src-xs/compartment.js` (which I otherwise did not bother typing)

packages/ses/types.d.ts
```diff
@@ -167,6 +167,9 @@ export interface CompartmentOptions {
   modules?: Map<string, ModuleDescriptor>;
   __shimTransforms__?: Array<Transform>;
   __noNamespaceBox__?: boolean;
+  loadHook?: (specifier: string) => Promise<ModuleDescriptor>;
+  loadNowHook?: (specifier: string) => ModuleDescriptor;
```
kriskowal: If it’s possible to mark these as deprecated, we can.

kriskowal (approved):
Much obliged.

PR #2846, opened 2025-06-06
compartment-mapper: fix dynamic require of relative paths

boneskull:
## Description

This fixes a situation in which a dynamically-required _relative_ path (specifically, starting with `../`) could not be resolved. It is converted to an absolute path, which is subequently resolved the same way `importNowHook` resolves all other absolute paths.

This behavior is _only_ applicable to dynamic requires; it is not supported by static imports of any kind.

Additionally, I simplified a couple types (`FileURLToPathFn` and `PathToFileURLFn`). This is not a breaking change, since `node:url`'s `fileURLToPath` and `pathToFileURL` functions _extend_ the new types.

### Security Considerations

n/a

### Scaling Considerations

n/a

n/a

### Testing Considerations

I've added a couple test cases. I did not add a test to assert this fails for static imports, since I expect such a test already exists.

This enables dynamic requires of ancestor relative paths.

### Upgrade Considerations

Kind of an edge case as most dynamic requires use absolute paths; probably not worth announcing.

### DIFF
--- a/packages/compartment-mapper/src/import-hook.js
+++ b/packages/compartment-mapper/src/import-hook.js
@@ -194,7 +194,8 @@ const findRedirect = ({
       // go up a directory
       const parentLocation = new URL('../', someLocation).href;
 
-      // afaict this behavior is consistent across both windows and posix
+      // afaict this behavior is consistent across both windows and posix:
+      // if this is true, we hit the filesystem root
       if (parentLocation === someLocation) {
         throw new Error(
           `Could not import unknown module: ${q(absoluteModuleSpecifier)}`,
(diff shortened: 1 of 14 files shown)


packages/compartment-mapper/src/import-hook.js
```diff
@@ -338,6 +339,20 @@ function* chooseModuleDescriptor(
...
+    const { fileURLToPath } = unpackReadPowers(readPowers);
+    if (
+      typeof fileURLToPath === 'function' &&
+      !isLocationWithinCompartment(moduleLocation, packageLocation)
+    ) {
+      return {
+        specifier: fileURLToPath(moduleLocation),
+        compartment: compartments[packageLocation],
+      };
+    }
```
naugtur, line 354: (could)
This returns a record that gets processed by the absolute path logic to check which compartment contains the absolute path, but we've already done the work here to find the right compartment. It's now only a matter of relativizing the specifier using a function that might already be in scope for this line.

naugtur (approved):

kriskowal (commented):
Approved, posthumously.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -338,6 +339,20 @@ function* chooseModuleDescriptor(
     // name, they are usable as URL's.
     const moduleLocation = resolveLocation(candidateSpecifier, packageLocation);
 
+    // check if moduleLocation is within the compartment if it is not, then
+    // create a redirect record using absolute path which will be resolved by
+    // findRedirect
+    const { fileURLToPath } = unpackReadPowers(readPowers);
```
kriskowal, line 345: We could avoid doing this work for every importHook invocation by moving it up to the maker.

PR #2848, opened 2025-06-10
feat(daemon): PINS restored on startup

kriskowal:
The pet dæmon will revive all of the objects in the PINS special directory whenever it restarts. This is a more general case of NETS, which must also be revived on restart so that they can establish listener sockets, but are more deeply integrated in peer names. The test included illustrates the usage through API.


rekmarks (approved):
LGTM!

PR #2849, opened 2025-06-10
Improve eventual types

mhofman:
Refs: https://github.com/Agoric/agoric-sdk/pull/11454

## Description

Narrows some eventual-send types. In particular, ensure that the result of an eventual call is a promise.

Add an `EResult` type which in the future can represent the return type of an eventual send call where remotables are mapped as necessary to their "brand" type only to avoid the caller from relying on what may otherwise look like a local remotable instead of a presence.

This mapping is implemented in the `EAwaitedResult` type for explicit usages, but not yet used in the type of `EResult` itself as it breaks too many consumers of `E` that rely on the currently "incorrect" typing.

### Security Considerations

None

### Scaling Considerations

None

TBD

### Testing Considerations

Integration tested in agoric-sdk: https://github.com/Agoric/agoric-sdk/pull/12065
In a perfect world, I'd like to add some type tests of cases I've encountered.

None

### Upgrade Considerations

Type only changes, which we usually don't consider breaking.

### DIFF
--- a/packages/eventual-send/src/E.js
+++ b/packages/eventual-send/src/E.js
@@ -322,17 +322,17 @@ export default makeE;
  *
  * @template T
  * @typedef {(
- *   0 extends (1 & T)
- *     ? T
- *     : T extends RemotableBrand<infer L, infer P>
- *       ? (P | RemotableBrand<L, P>)
- *       : T extends PromiseLike<infer U>
- *         ? Promise<EAwaitedResult<Awaited<T>>>
+ *   0 extends (1 & T)                                        // If T is any
+ *     ? T                                                    // Propagate the any type through the result
+ *     : T extends RemotableBrand<infer L, infer P>           // If we have a Remotable
+ *       ? (P | RemotableBrand<L, P>)                         // map it to its "maybe remote" form (primary behavior or remotable presence)
+ *       : T extends PromiseLike<infer U>                     // If T is a promise
+ *         ? Promise<EAwaitedResult<Awaited<T>>>              // map its resolution
  *         : T extends (null | undefined | string | number | boolean | symbol | bigint | Callable) // Intersections of these types with objects are not mapped
- *           ? T
- *           : T extends object
- *             ? { [P in keyof T]: EAwaitedResult<T[P]>; }
- *             : T
+ *           ? T                                              // primitives and non-remotable functions are passed-through
+ *           : T extends object                               //
+ *             ? { [P in keyof T]: EAwaitedResult<T[P]>; }    // other objects are considered copy data and properties mapped
+ *             : T                                            // in case anything wasn't covered, fallback to pass-through
  * )} EAwaitedResult
  */
 


mhofman:
This is integration tested in https://github.com/Agoric/agoric-sdk/pull/11454, I don't have the bandwidth to write proper type tests here right now.

packages/eventual-send/src/E.js
```diff
@@ -302,26 +302,90 @@ export default makeE;
  * `Promise<T>` or `T` itself.
  */
 
+/**
+ * A type that doesn't assume its parameter is local, but is satisfied with both
+ * local and remote references. It accepts both near and marshalled references
```
erights:
```suggestion
 * local and remote references. It accepts both near and unserialized (imported) references
```

packages/eventual-send/src/E.js
```diff
@@ -302,26 +302,90 @@ export default makeE;
...
+ * that were returned from `Remotable` or `Far`.
+ *
+ * @template Primary The type of the primary reference.
+ * @template [Local=DataOnly<Primary>] The local properties of the object.
+ * @typedef {Primary | RemotableBrand<Local, Primary>} Remote
```
erights: If this mean "local or remote" as it seems to, "Remote" seems a strange name. Remotable already seems to mean what you want, with the difference being that your Remote includes non-remotable local objects that also implement the Primary API. Is that what you intend?

packages/eventual-send/src/E.js
```diff
@@ -302,26 +302,90 @@ export default makeE;
...
+ */
+
+/**
+ * A type that accepts either resolved or promised references that may be either
+ * near or marshalled. @see {ERef} and @see {Remote}.
+ *
+ * @template Primary The type of the primary reference.
+ * @template [Local=DataOnly<Primary>] The local properties of the object.
+ * @typedef {ERef<Remote<Primary, Local>>} ERemote
```
erights: How does this differ from `PassableCap`, `PassByReference`, or your `RemotableObject | RemotableBrand` pattern?

erights (changes requested):
Having types `ERemote` `PassableCap`, `PassByRef`, or your `RemotableObject | RemotableBrand` pattern as ways to say essentially the same thing makes me feel helpless. When next I need to write code, I don't know that I can do better than guess which type I should use. Not good. When a change would make me scared to try to write more code, I gotta request changes.

If we did accept all these, what I would do in practice is scatter `@ts-ignore` all over the place so I could still write code. Also not good.

mhofman:
> Having types `ERemote` `PassableCap`, `PassByRef`, or your `RemotableObject | RemotableBrand` pattern as ways to say essentially the same thing makes me feel helpless. When next I need to write code, I don't know that I can do better than guess which type I should use. Not good. When a change would make me scared to try to write more code, I gotta request changes.

Let's talk offline, there is obviously a misunderstanding about the scope of this change. `Remote` is already all over the place in agoric-sdk. `ERemote` is a trivial helper on top of it. I'm ok removing these from endo if needed, but they're the types that are most likely to be used by actual users (not maintainers). I agree the naming is not optimal, but agoric-sdk is kinda forcing our hand here.

mhofman:
I removed the `Remote` and `ERemote` helpers from endo as it's not clear those will be what end users are expected to use, and we shouldn't force endo to suffer the imperfect naming choices of agoric-sdk.

@erights PTAL

packages/captp/src/loopback.js
```diff
@@ -92,13 +92,14 @@ export const makeLoopback = (ourId, nearOptions, farOptions) => {
...
      * @param {T} x
-     * @returns {Promise<T>}
+     * @returns {Promise<EResult<T>>}
      */
     async x => {
       lastNonce += 1;
       const myNonce = lastNonce;
       const val = await x;
       nonceToRef.set(myNonce, harden(val));
+      // @ts-expect-error Type 'T | Awaited<T>' is not assignable to type 'EResult<T>'
```
erights, line 102: Minor. Doesn't affect actual rendering, but eases copy-pasting text to/from doc-comments
```suggestion
      // @ts-expect-error Type `T | Awaited<T>` is not assignable to type `EResult<T>`
```

packages/eventual-send/src/E.js
erights, line 278: Is this in preparation for the "auxiliary data" that we do not implement and probably never will?

packages/eventual-send/src/E.js
```diff
@@ -304,24 +304,69 @@ export default makeE;
...
  * @template {(...args: any[]) => any} T
  * @typedef {T extends (...args: any[]) => infer R ? Awaited<R> : never} EReturn
  */
 
 /**
- * @template {import('./types.js').Callable} T
+ * An eventual value where remotable objects are recursively mapped to Remote types
+ *
+ * @template T
+ * @typedef {Awaited<T>} EResult
```
erights, line 317: How does `Awaited<T>` differ from `EResult<T>`?

packages/eventual-send/src/E.js
```diff
@@ -304,24 +304,69 @@ export default makeE;
...
+ */
+
+/**
+ * Experimental type mapping remotable objects to Remote types
```
erights, line 321: In what sense is this experimental? If it is experimental, does that mean production code should not rely on it? In that case, why is it in this PR?

erights (commented):
Just comments so far. More coming...

packages/eventual-send/src/E.js
erights: Ah. The PR comment says

> Add an `EResult` type which in the future can represent the return type of an eventual send call mapped as necessary to avoid having the caller rely on the locality of the result.

which explains why it is not redundant to introduce something that starts identical. However, I did not understand the rest of that sentence. Rephrase?

packages/eventual-send/src/E.js
erights: With `Remote` and `ERemote` gone, resolving.

packages/eventual-send/src/E.js
erights: With `Remote` and `ERemote` gone, resolving.

packages/eventual-send/src/E.js
mhofman: Out of scope for this PR, this type was there before.

packages/eventual-send/src/E.js
mhofman: It's implemented but not used, because too many things break in agoric-sdk if we try to change `EResult` to use it.

But technically one could manually use it for their own experiments, so it is defined and exported. Id' rather not lose it completely.

packages/captp/src/loopback.js
mhofman: Suggestion declined, as this is a literal copy of the error that tsc spits out.

packages/eventual-send/src/E.js
mhofman: Clarified as:

> Add an `EResult` type which in the future can represent the return type of an eventual send call where remotables are mapped as necessary to their "brand" type only to avoid the caller from relying on what may otherwise look like a local remotable instead of a presence.
> 
> This mapping is implemented in the `EAwaitedResult` type for explicit usages, but not yet used in the type of `EResult` itself as it breaks too many consumers of `E` that rely on the currently "incorrect" typing.

packages/captp/src/loopback.js
erights: Ah. Resolving.

packages/eventual-send/src/E.js
erights: Ok, resolving. But what do you think it means?

packages/eventual-send/src/E.js
erights: Thanks! Resolving.

packages/eventual-send/src/E.js
erights: Uh, ok. Reluctantly resolving. My own practice is to keep such experiments in draft PRs on the side.

packages/eventual-send/src/E.js
mhofman: I believe @michaelfig can better speak to this.

packages/eventual-send/src/E.js
erights: Actually, can you also explain this (briefly?) in the doc-comment here? Thx.

packages/eventual-send/src/E.js
```diff
@@ -304,24 +304,69 @@ export default makeE;
...
+ *
+ * @template T
  * @typedef {(
- *   ReturnType<T> extends PromiseLike<infer U>                       // if function returns a promise
- *     ? T                                                            // return the function
- *     : (...args: Parameters<T>) => Promise<EReturn<T>>  // make it return a promise
+ *   0 extends (1 & T)
```
erights: I have no idea what this line is about. Also below. Whatever this is doing, I have not seen this trick before.

mhofman: This is detecting an `any` type, and bailing if it is. Otherwise `any` causes both branches of a ternary to be taken, and the result is the union of both sides.

erights: WOW!

Worth an inline `//` comment!

packages/eventual-send/src/E.js
```diff
@@ -304,24 +304,69 @@ export default makeE;
...
+ *     ? T
+ *     : T extends RemotableBrand<infer L, infer P>
+ *       ? (P | RemotableBrand<L, P>)
+ *       : T extends PromiseLike<infer U>
+ *         ? Promise<EAwaitedResult<Awaited<T>>>
```
erights: I thought we could not define recursive types in jsdoc-comments. Glad to see we can!

packages/eventual-send/src/E.js
```diff
@@ -304,24 +304,69 @@ export default makeE;
...
+ *         : T extends (null | undefined | string | number | boolean | symbol | bigint | Callable) // Intersections of these types with objects are not mapped
```
erights, line 331: Is a so-called `FarFunction` a `Callable` and a remotable? Is this branch consistent with that?

packages/eventual-send/src/E.js
erights: Could we give it a name so we don't need to comment each occurrence?

erights (approved):
LGTM. Thanks for bearing with me ;)

kriskowal (approved):

packages/eventual-send/src/E.js
mhofman: Being a type ternary, unfortunately we cannot make a helper for this. This trick is commented elsewhere in this file, but I will add a comment at this location as well.

packages/eventual-send/src/E.js
mhofman: A far function would be a remotable, and thus match the `RemotableBrand` case earlier. This is the case where we're supporting eventual-send with results containing a pure local (non remotable) functions, and simply passing them through at the type level. Added a comment to clarify

kriskowal (approved):
Consider adding an entry in `NEWS.md` for `eventual-send` and/or `far` under `# Next release`.

PR #2850, opened 2025-06-10
types(pass-style): accept RemotableBrand as PassableCap

mhofman:
Refs: https://github.com/Agoric/agoric-sdk/pull/11454

## Description

In some cases, a remotable object may have a `RemotableBrand` type but not a `RemotableObject` type. This is because the former can easily be "tagged" onto an existing type (as a nominal type without public properties), but the latter would cause type checking issues (as a structural type).

This change enables pass-style and pattern types to accept `RemotableBrand` wherever a `RemotableObject` is accepted as a remotable passable cap.

### Security Considerations

None, types only.

### Scaling Considerations

None

None

### Testing Considerations

Manually integration tested against agoric-sdk.

This may break some  erroneous type checks.

### Upgrade Considerations

Types only are usually considered non breaking.

### DIFF
--- a/packages/pass-style/src/types.d.ts
+++ b/packages/pass-style/src/types.d.ts
@@ -1,3 +1,4 @@
+import type { RemotableBrand } from '@endo/eventual-send';
 /* eslint-disable no-use-before-define */
 import { PASS_STYLE } from './passStyle-helpers.js';
 
@@ -65,7 +66,9 @@ export type PassByCopy = Atom | Error | CopyArray | CopyRecord | CopyTagged;
 
 export type PassByRef =
   | RemotableObject
+  | RemotableBrand<any, any>
   | Promise<RemotableObject>
+  | Promise<RemotableBrand<any, any>>
   | Promise<PassByCopy>;
 
 /**
@@ -165,7 +168,10 @@ export type RemotableObject<I extends InterfaceSpec = string> = PassStyled<
 /**
  * The authority-bearing leaves of a Passable's pass-by-copy superstructure.
  */
-export type PassableCap = Promise<any> | RemotableObject;
+export type PassableCap =
+  | Promise<any>
+  | RemotableObject
+  | RemotableBrand<any, any>;
 
 /**
  * A Passable sequence of Passable values.
--- a/packages/pass-style/src/deeplyFulfilled.js
+++ b/packages/pass-style/src/deeplyFulfilled.js
@@ -7,6 +7,7 @@ import { makeTagged } from './makeTagged.js';
 import { isAtom } from './typeGuards.js';
 
 /**
+ * @import {RemotableBrand} from '@endo/eventual-send';
  * @import {Passable, ByteArray, CopyRecord, CopyArray, CopyTagged, RemotableObject} from '@endo/pass-style'
  */
 
@@ -47,9 +48,11 @@ const { fromEntries } = Object;
  * @template T
  * @typedef {T extends PromiseLike<any>
  *     ? Awaited<T>
- *     : T extends {}
- *       ? Simplify<DeeplyAwaitedObject<T>>
- *       : Awaited<T>} DeeplyAwaited
+ *     : T extends (RemotableBrand<any, any> | RemotableObject)
+ *       ? T
+ *       : T extends {}
+ *         ? Simplify<DeeplyAwaitedObject<T>>
+ *         : Awaited<T>} DeeplyAwaited
  */
 
 /**
--- a/packages/patterns/src/types.js
+++ b/packages/patterns/src/types.js
@@ -4,6 +4,7 @@ export {};
 
 // NB: as of TS 5.5 nightly, TS thinks RankCover and Checker "is declared but never read" but they are
 /**
+ * @import {RemotableBrand} from '@endo/eventual-send';
  * @import {Checker, CopyArray, CopyRecord, CopyTagged, Passable, PassStyle, Atom, RemotableObject} from '@endo/pass-style';
  * @import {PartialCompare, PartialComparison, RankCompare, RankCover} from '@endo/marshal';
  */
@@ -14,7 +15,7 @@ export {};
  */
 
 /**
- * @typedef {Exclude<Passable<RemotableObject, never>, Error | Promise>} Key
+ * @typedef {Exclude<Passable<RemotableObject | RemotableBrand<any, any>, never>, Error | Promise>} Key
  *
  * Keys are Passable arbitrarily-nested pass-by-copy containers
  * (CopyArray, CopyRecord, CopySet, CopyBag, CopyMap) in which every
@@ -69,7 +70,7 @@ export {};
  */
 
 /**
- * @typedef {Atom | RemotableObject} ScalarKey
+ * @typedef {Atom | RemotableObject | RemotableBrand<any, any>} ScalarKey
  */
 
 /**


mhofman:
@michaelfig @erights PTAL I'd like to land this to get moving on some agoric-sdk type tightening.

packages/pass-style/src/types.d.ts
```diff
@@ -78,7 +79,9 @@ export type PassByCopy = Atom | Error | CopyArray | CopyRecord | CopyTagged;
 
 export type PassByRef =
```
erights, line 80: Conceptually is this the same as `PassableCap`? Do we need both? When should we use one vs the other? Does `PassableCap` need the same elaboration?

packages/pass-style/src/types.d.ts
```diff
@@ -78,7 +79,9 @@ export type PassByCopy = Atom | Error | CopyArray | CopyRecord | CopyTagged;
...
   | RemotableObject
+  | RemotableBrand<any, any>
```
erights, line 82: What does `RemotableBrand<any, any>` cover that `RemotableObject` does not?

I understand the old definition here as corresponding to what this means dynamically. Looking at the new definition, all I infer is "Oh that's probably some magical type stuff that's besides the point of describing the dynamic meaning of `PassByRef`"

packages/pass-style/src/types.d.ts
```diff
@@ -188,7 +191,10 @@ export type RemotableMethodName = PropertyKey;
 /**
  * The authority-bearing leaves of a Passable's pass-by-copy superstructure.
  */
-export type PassableCap = Promise<any> | RemotableObject;
+export type PassableCap =
+  | Promise<any>
+  | RemotableObject
+  | RemotableBrand<any, any>;
```
erights, line 200: Ok, I see you do make a similar elaboration of `PassableCap`. Same comments apply regarding meaning.

For both before and after, why is this distinct from `PassByRef`?

packages/patterns/src/types.ts
```diff
@@ -72,11 +73,11 @@ export type { FullCompare } from '@endo/marshal';
  *    contain more than one key.
  */
 export type Key = Exclude<
```
erights, line 75: For both before and after, what is `Exclude` doing that the limited parameterization of `Passable` is not doing?

packages/patterns/src/types.ts
```diff
@@ -72,11 +73,11 @@ export type { FullCompare } from '@endo/marshal';
...
-  Passable<RemotableObject, never>,
+  Passable<RemotableObject | RemotableBrand<any, any>, never>,
```
erights, line 76: If most places where we say `RemotableObject` we need to add `| RemotableBrand<any, any>`, can we just have one type that covers both? Can that type be the new `RemotableObject`, in which case most of these other changes can be reverted?

packages/patterns/src/types.ts
```diff
@@ -779,9 +780,7 @@ export type CopyTaggedMethodGuard = CopyTagged<
  */
 export type MethodGuard = CopyTagged<'guard:methodGuard', MethodGuardPayload>;
 
-export type AwaitArgGuardPayload = {
-  argGuard: Pattern;
-};
+export type AwaitArgGuardPayload = { argGuard: Pattern };
```
erights: Not that it affects the review, but I am curious. Before and after looks like the same sequence of tokens (except for an optional semicolon). Why did `yarn format` change its mind?

packages/patterns/src/types.ts
erights: Ok, I see you do explain it in the PR comment:

> In some cases, a remotable object may have a RemotableBrand type but not a RemotableObject type. This is because the former can easily be "tagged" onto an existing type (as a nominal type without public properties), but the latter would cause type checking issues (as a structural type).
>
> This change enables pass-style and pattern types to accept RemotableBrand wherever a RemotableObject is accepted as a remotable passable cap.

But after reading twice, I don't understand it. Still, same question: Can we instead just extend the meaning of `RemotableObject` to cover enough to maintain its intended meaning?

IOW, does the explanation above mean that the current `RemotableObject` definition is a bug; that it is too narrow to express its intended meaning?

mhofman: in TS terms:
- `RemotableObject` is the structural way of identifying a remotable. it has a property "pass-style symbol" with a value of "remotable"
- `RemotableBrand` is the orthogonal nominal way of identifying a remotable (which carries the remote type information). Think of it as the equivalent of a private field, but at the type level, with value bring the remote type info

For historical reasons not all remotable objects have the combination of both these types. The types definitions also live at different layers in our stack (RemotableBrand is from the eventual send layer, RemotableObject from the pass-style layer)

packages/patterns/src/types.ts
mhofman: Not sure tbh, I will remove formatting changes

packages/pass-style/src/types.d.ts
mhofman: I am not interested in renaming or cleaning up existing definitions in this PR. It does seem to narrow the nested Promise result types.

I have some attempts at properly recursing promise type narrowing in PassableCaps, maybe some day I'll be able to land it in a way that doesn't make agoric-sdk explode.

mhofman:
@erights I added the `RemotableBrand` vs `RemotableObject` clarifications we discussed. PTAL

packages/eventual-send/src/types.d.ts
```diff
@@ -19,8 +19,16 @@ export type * from './track-turns.js';
 
 export type Callable = (...args: any[]) => any;
 
+// TODO: We likely want to merge the RemotableObject symbol shape into this
+// type to reduce confusion about why both exist, but all kinds of things blow
+// up when trying to do that.
```
erights: Thanks. This seems worth an issue, which you can cite here.

packages/marshal/src/encodeToSmallcaps.js
```diff
@@ -18,7 +18,7 @@ import {
 } from '@endo/pass-style';
 import { X, Fail, q } from '@endo/errors';
 
-/** @import {Passable, Remotable} from '@endo/pass-style' */
+/** @import {Passable, RemotableObject as Remotable} from '@endo/pass-style' */
```
erights: Why "as" rather than just a local search-and-replace? We should just stop using `Remotable` as a type name, even if it is a safely local alias as you've done here.

packages/eventual-send/src/types.d.ts
```diff
@@ -19,8 +19,16 @@ export type * from './track-turns.js';
...
+
 /**
  * Nominal type to carry the local and remote interfaces of a Remotable.
+ *
+ * Note: this type does not currently include the structural aspect of the
+ * {@link RemotableObject} type, and as such is not suitable to represent a
+ * "remotable" for APIs that expect an object with a pass-style symbol.
  */
 export declare class RemotableBrand<Local, Remote> {
```
erights, line 33: Since you're deprecating `RemotableObject` in favor of `RemotableBrand`, please make the type parameters of `RemotableBrand` optional. Which is something I'd ask for anyway.

packages/pass-style/src/types.d.ts
erights: Since you're deprecating `RemotableObject` in favor of `RemotableBrand`, resolving.

packages/pass-style/src/types.d.ts
erights: Ok, resolving. But please file an issue. (Ok to file after this PR is merged)

packages/pass-style/src/types.d.ts
erights: Per above, resolving.

packages/pass-style/src/types.d.ts
erights: Actually, unresolving until we agree on whether or not we should at least deprecate `PassByRef`

packages/patterns/src/types.ts
erights: Now that you're deprecating `RemotableObject` in favor of `RemotableBrand` and planning on enhancing `RemotableBrand` so it subsumes `RemotableObject`, once you do so enhance `RemotableBrand`, then we can drop the `RemotableObject |` disjunct. Yes?

Can we do that in this PR?

erights (approved):

packages/marshal/src/encodeToSmallcaps.js
mhofman: A couple other files in the same package already used this style so I adopted it, happy to change the other files to a non aliased version.

packages/eventual-send/src/types.d.ts
mhofman: The problem is that `RemotableBrand`, which is already extensively used, should require a mandatory `Remote` type. And switching the order of the Local and Remote type parameter would be breaking. I'm not sure there is a solution to be honest.

packages/patterns/src/types.ts
mhofman: `RemotableObject` is still in massive use so cannot be removed, and merging the type definitions is a much bigger scope and higher risk of breakage, so out of scope.

packages/pass-style/src/types.d.ts
mhofman: `PassByRef` is different and stricter than `PassableCap`. I don't think it should be deprecated. There is some cleanup we could do, possibly making `PassableCap` stricter or in fact make `Passable` a more customizable type. There has been attempts before, one of them at https://github.com/endojs/endo/pull/2421. I suppose we don't have a proper issue that documents the problem.

packages/marshal/src/encodeToSmallcaps.js
erights: Yes please!

packages/pass-style/src/types.d.ts
erights: Ok, reluctantly resolving. Such an issue would be appreciated though.

packages/patterns/src/types.ts
erights: Ok, not in this PR, so resolving. But when we try this, do you expect it to work?

packages/patterns/src/types.ts
mhofman: Yeah I'm not sure, but I'd like to avoid more drive by changes.

packages/patterns/src/types.ts
mhofman: I'm not super hopeful at this point, having tried and failed multiple times.

packages/eventual-send/src/types.d.ts
erights: > The problem is that `RemotableBrand`, which is already extensively used, should require a mandatory `Remote` type.

Why? And if so, why not just make it the default? Layering?

> And switching the order of the Local and Remote type parameter would be breaking.

Is this relevant if we make all type parameters optional as I'm suggesting?

packages/patterns/src/types.ts
erights: ok. resolving

packages/eventual-send/src/types.d.ts
mhofman: `RemotableBrand` is currently only used to carry the behavior "remote" type. Defaulting to `any` would be too easy to spread `any` throughout, which I do not want. The only time `any` is acceptable is when saying "I want to accept any object that has this brand on", which for the very few times this is needed, can be expressed with an explicit type.

I reverting the question, why do you want default values? IMO there is no default value for `Remote` that is appropriate.

erights: What about defaulting to `object` or `{}`?

> I reverting the question, why do you want default values? IMO there is no default value for Remote that is appropriate.

Ok, now that you ask, I'm not sure. I'd have to look at existing mentions of `RemotableObject` and see if there's an easy, obvious, correct value for at least one of the type parameters there.

IOW, it only makes sense to deprecate `RemotableObject` to `RemotableBrand` with a mandatory type parameter if it would hypothetically make sense to give `RemotableObject` a mandatory type parameter, transition costs aside.

erights: In any case, this is enough to resolve.

packages/eventual-send/src/types.d.ts
mhofman: https://github.com/endojs/endo/issues/2979

PR #2854, opened 2025-06-11
refactor(path-compare): Improve documentation and reduce implementation code size

gibson042:
Refs: #2787

## Description

Applies suggestions from https://github.com/endojs/endo/pull/2787#pullrequestreview-2812907861 .

### DIFF
--- a/packages/path-compare/src/index.js
+++ b/packages/path-compare/src/index.js
@@ -22,16 +22,6 @@ const { stringify: q } = JSON;
 export const stringCompare = (a, b) => (a === b ? 0 : a < b ? -1 : 1);
 
 /**
- * Reducer to calculate the cumulative length of a string array
- * @param {number} length Current length
- * @param {string} value Some string
- * @returns {number} New length
- */
-const cumulativeLength = (length, value) => {
-  return length + value.length;
-};
-
-/**
  * Compares two string arrays and returns a comparison result.
  *
  * The algorithm is as follows:
@@ -50,15 +40,10 @@ const cumulativeLength = (length, value) => {
  * @type {CompareFn<string[]|undefined>}
  */
 export const pathCompare = (a, b) => {
-  // Undefined is not preferred
-  if (a === undefined && b === undefined) {
-    return 0;
-  }
-  if (a === undefined) {
-    return 1;
-  }
-  if (b === undefined) {
-    return -1;
+  // Undefined compares greater than anything else.
+  if (a === undefined || b === undefined) {
+    // eslint-disable-next-line no-nested-ternary
+    return a === b ? 0 : a === undefined ? 1 : -1;
   }
 
   // Prefer the shortest dependency path.
@@ -67,10 +52,10 @@ export const pathCompare = (a, b) => {
   }
 
   // Otherwise, favor the shortest cumulative length.
-  const aSum = a.reduce(cumulativeLength, 0);
-  const bSum = b.reduce(cumulativeLength, 0);
-  if (aSum !== bSum) {
-    return aSum - bSum;
+  const aStringLength = a.join('').length;
+  const bStringLength = b.join('').length;
+  if (aStringLength !== bStringLength) {
+    return aStringLength - bStringLength;
   }
 
   // sanity check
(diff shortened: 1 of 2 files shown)


packages/path-compare/src/index.js
```diff
@@ -67,10 +52,10 @@ export const pathCompare = (a, b) => {
   }
 
   // Otherwise, favor the shortest cumulative length.
-  const aSum = a.reduce(cumulativeLength, 0);
-  const bSum = b.reduce(cumulativeLength, 0);
-  if (aSum !== bSum) {
-    return aSum - bSum;
+  const aStringLength = a.join('').length;
+  const bStringLength = b.join('').length;
```
kriskowal, line 56: I do not feel strongly about this: I agree that the `reduce` is not ideal, but I prefer a solution that improves both legibility and is lest wasteful with allocations. To that end, I’d prefer to keep this as-is or go the other direction: replace `cumulativeLength`’s internals with a C-style for loop and call that once on each of these lines.

gibson042: I understand the concern, but in practice I just can't see those allocations at all... I think the internal allocation implied by `strings.join('').length` doesn't actually happen in V8, as observed by comparing output from the following:
```console
$ esbench.mjs -h V8 \
  -s 'const arr = Array.from({ length: 1000 }, (_, i) => `element${i}`);' \
  join:'ret = []; for(let i = 0; i < arr.length; i++) ret.push( arr.slice(i).join("") );' \
  --dump \
  | node --heap-prof-dir="$(pwd)" --heap-prof --heap-prof-interval 100
join sync 0.04 ops/ms (24.5e+6 ns/op) after 123 1-count observations

$ esbench.mjs -h V8 \
  -s 'const arr = Array.from({ length: 1000 }, (_, i) => `element${i}`);' \
  join:'ret = []; for(let i = 0; i < arr.length; i++) ret.push( arr.slice(i).join("").length );' \
  --dump \
  | node --heap-prof-dir="$(pwd)" --heap-prof --heap-prof-interval 100
join sync 0.04 ops/ms (24.5e+6 ns/op) after 123 1-count observations
```

The first captures output from `join` and so requires string allocations (`takeObservation` claims 5 MB, dominated by `join`), while the second does not (`takeObservation` claims just 18 KB, a third of which is `Array.from`).

packages/path-compare/src/index.js
```diff
@@ -50,15 +40,10 @@ const cumulativeLength = (length, value) => {
...
-  }
-  if (a === undefined) {
-    return 1;
-  }
-  if (b === undefined) {
-    return -1;
+  // Undefined compares greater than anything else.
+  if (a === undefined || b === undefined) {
+    // eslint-disable-next-line no-nested-ternary
+    return a === b ? 0 : a === undefined ? 1 : -1;
```
boneskull, line 46: [This rule](https://eslint.org/docs/latest/rules/no-nested-ternary) exists because:

> Nesting ternary expressions can make code more difficult to understand.

It's hard for me to disagree with that.  

If we're going to disable it on a one-off basis (and again on L21 above), we should have a justification for doing so.  Otherwise, it may be better to leave it for now, then  propose we disable the rule because we feel it isn't buying us anything.

Ideally, I'd like a rule that demands an extra comment for each one-off disabled rule (much like [ban-ts-comment](https://typescript-eslint.io/rules/ban-ts-comment/) configured with `'ts-expect-error': 'allow-with-description'`). 😄  I bet someone solved this already...

packages/path-compare/src/index.js
boneskull: Is it really worth optimizing?  Given where `pathCompare` is consumed, the bottleneck is I/O.

boneskull (approved):
No blockers

packages/path-compare/src/index.js
gibson042: What I think we really want is a maximum ternary nesting depth of 1 rather than 0, to allow for precisely this kind of use in comparison functions (`$equal ? 0 : $aLessThanB ? -1 : 1`).

PR #2855, opened 2025-06-12
fix(immutable-arraybuffer): unify shim to work on more platforms

erights:
Closes: #XXXX
Refs: #2785 #2399

## Description

The Immutable ArrayBuffers shim can only emulate `transferToImmutable` on platforms with either `structuredClone` or `ArrayBuffer.prototype.transfer`. Otherwise, the shim would refuse to initialize. Without Immutable ArrayBuffers, we cannot initialize `ses`, and `pass-style` cannot implement the new `byteArray` Passable type, which `marshal` and `patterns` now depends on.

Unfortunately, it turns out that some platforms we want to support do not have `structuredClone` or `ArrayBuffer.prototype.transfer`. The first we encountered was Hermes, that also does not have classes with private fields. #2785 makes all these work on Hermes, by supporting `sliceToImmutable` instead, and using WeakMaps rather than private fields. On the theory that Hermes was a one-off, #2785 sniffed to see if it was on Hermes, only only then used the alternate shim.

In turns out Hermes was not a one-off. Some version of JSC of concern (TODO which ones?) also have neither `structuredClone` or `ArrayBuffer.prototype.transfer`. So this PR unifies the two shim strategies into one which uses the WeakMap technique, which works everywhere, and emulating `sliceToImmutable`, which we can do everywhere. It still sniffs for `structuredClone` or `ArrayBuffer.prototype.transfer`, and only emulates `transferToImmutable` if one is present. Thus, we sniff only for features not platforms, and emulate `sliceToImmutable` on all platforms.

Other PRs (TODO: which ones?) already ensure that `pass-style`, `marshal`, and `patterns` are fine with the absence of `transferToImmutable`.

### Security Considerations

A singe piece of code with small feature detects is simpler, and thus more secure, than maintaining two alternate pieces of code, each for a different set of platforms.

### Scaling Considerations

On most platforms, a class with private fields is probably more efficient than a WeakMap. This PR sacrifices that optimization in exchange for unification.

Less to document.

### Testing Considerations

The "unified" tests are just the union of the old tests, and so have two problems
- [ ] redundancy, which needs to be removed
- [ ] contains unconditional `transferToImmutable` tests that won't work on some platforms. These specific tests need to be made conditional, probably using the same `test.skip` trick some tests use for `Array.prototype.transfer`.

The goal is to make the shim compat with more platforms. On platforms that worked before this PR, this PR should be a pure refactor, and so without any compat hazards.

### Upgrade Considerations


leotm (approved):
Tested working (React Native runtime) on both legacy RN JSC (`ses.cjs`) and Hermes (`ses-hermes.cjs`) ^ better unified approach than my prev initial thoughts of introducing another flavour SES shim, less maintenance and 1 Hermes transform fewer

packages/immutable-arraybuffer/src/immutable-arraybuffer-pony.js
```diff
@@ -0,0 +1,215 @@
...
+  // @ts-ignore At the time of this writing, the `ArrayBuffer` type built
+  // into TypeScript does not know about the recent standard `transfer` method.
+  // Indeed, the `transfer` method is absent from Node <= 20.
+  transfer: transferMaybe,
+} = arrayBufferPrototype;
+// Capture structuredClone before it could be scuttled.
+const { structuredClone: structuredCloneMaybe } = globalThis;
+
+/**
+ * Enforces that `realBuffer` is a genuine `ArrayBuffer` exotic object.
```
gibson042:
```suggestion
 * Copy a range of values from a genuine ArrayBuffer exotic object into a new
 * ArrayBuffer.
```

packages/immutable-arraybuffer/src/immutable-arraybuffer-pony.js
```diff
@@ -0,0 +1,215 @@
...
+ * original `arrayBuffer` has been moved into the new one, and the original
+ * `arrayBuffer` has been detached. We can only do this emulation on platforms
+ * that support `structureClose` or `ArrayBuffer.prototype.transfer`.
+ * On other platforms, we can still emulate `sliceToImmutable` but not
+ * `arrayBufferTransferMaybe`, and therefore not
+ * `ArrayBuffer.prototype.transferToImmutable`. Currently, these other platforms
+ * are
+ * - Hermes
+ * - Node <= 16
+ * - Apparently some versions of JSC that are still of concern.
```
gibson042:
```suggestion
 * Move the contents of a genuine ArrayBuffer exotic object into a new fresh
 * ArrayBuffer and detach the original source.
 * We can only do this on platforms that support `structuredClone` or
 * `ArrayBuffer.prototype.transfer`.
 * On other platforms, we can still emulate
 * `ArrayBuffer.prototoype.sliceToImmutable`, but not
 * `ArrayBuffer.prototype.transferToImmutable`.
 * Currently, these known-deficient platforms are
 * - Hermes
 * - Node.js <= 16
 * - Apparently some versions of JavaScriptCore that are still of concern.
```

packages/immutable-arraybuffer/src/immutable-arraybuffer-pony.js
```diff
@@ -0,0 +1,215 @@
...
+}
+
+/**
+ * If we could use classes with private fields everywhere, this would have
+ * been a `this.#buffer` private field on an `ImmutableArrayBufferInternal`
+ * class. But we cannot do so on Hermes. So, instead, we
+ * emulate the `this.#buffer` private field, including its use as a brand check.
+ * Maps from all and only emulated Immutable ArrayBuffers to real ArrayBuffers.
+ */
+const buffers = new WeakMap();
```
gibson042: Post-hoc prototype method invocations on `buffers` are not safe in code that sits underneath SES. The `commons` approach captures such methods and converts them into functions, but for this narrow case we can also just patch up the singleton object:
```suggestion
/**
 * If we could use classes with private fields everywhere, this would have
 * been a `this.#buffer` private field on an `ImmutableArrayBufferInternal`
 * class. But we cannot do so on Hermes. So, instead, we
 * emulate the `this.#buffer` private field, including its use as a brand check.
 * Maps from all and only emulated Immutable ArrayBuffers to real ArrayBuffers.
 *
 * @type {Pick<WeakMap<ArrayBuffer, ArrayBuffer>, 'get' | 'has' | 'set'>}
 */
const buffers = new WeakMap();
// Avoid post-hoc prototype lookups.
for (methodName of ['get', 'has', 'set']) {
  defineProperty(buffers, methodName, { value: buffers[methodName] });
}
```

packages/immutable-arraybuffer/src/immutable-arraybuffer-pony.js
```diff
@@ -0,0 +1,215 @@
...
+const getBuffer = immuAB => {
+  const result = buffers.get(immuAB);
+  if (result) {
+    return result;
+  }
+  throw TypeError('Not an emulated Immutable ArrayBuffer');
+};
+
+// Omits `constructor` so `Array.prototype.constructor` is inherited.
```
gibson042, line 102:
```suggestion
// Omits `constructor` so `ArrayBuffer.prototype.constructor` is inherited.
```

packages/immutable-arraybuffer/src/immutable-arraybuffer-pony.js
```diff
@@ -0,0 +1,215 @@
...
+  get byteLength() {
+    return getBuffer(this).byteLength;
+  },
+  get detached() {
+    getBuffer(this); // shim brand check
+    return false;
+  },
+  get maxByteLength() {
+    // Not underlying maxByteLength, which is irrelevant
+    return getBuffer(this).byteLength;
```
gibson042:
```suggestion
    return apply(arrayBufferByteLength, getBuffer(this), []);
  },
  get detached() {
    getBuffer(this); // shim brand check
    return false;
  },
  get maxByteLength() {
    // Not underlying maxByteLength, which is irrelevant
    return apply(arrayBufferByteLength, getBuffer(this), []);
```

packages/immutable-arraybuffer/src/immutable-arraybuffer-pony.js
```diff
@@ -0,0 +1,215 @@
...
+  slice,
+  // TODO used to be a-ts-expect-error, but my local IDE's TS server
+  // seems to use a more recent definition of the `ArrayBuffer` type.
+  // @ts-ignore At the time of this writing, the `ArrayBuffer` type built
+  // into TypeScript does not know about the recent standard `transfer` method.
+  // Indeed, the `transfer` method is absent from Node <= 20.
+  transfer: transferMaybe,
+} = arrayBufferPrototype;
+// Capture structuredClone before it could be scuttled.
+const { structuredClone: structuredCloneMaybe } = globalThis;
```
gibson042:
```suggestion
const { ArrayBuffer, Object, Reflect, TypeError, Uint8Array } = globalThis;

// Capture structuredClone before it can be scuttled.
const { structuredClone: structuredCloneMaybe } = globalThis;

const { defineProperty } = Object;
const { apply, ownKeys } = Reflect;

const { prototype: arrayBufferPrototype } = ArrayBuffer;
const {
  slice,
  // TODO used to be a-ts-expect-error, but my local IDE's TS server
  // seems to use a more recent definition of the `ArrayBuffer` type.
  // @ts-ignore At the time of this writing, the `ArrayBuffer` type built
  // into TypeScript does not know about the recent standard `transfer` method.
  // Indeed, the `transfer` method is absent from Node <= 20.
  transfer: transferMaybe,
} = arrayBufferPrototype;
const { get: arrayBufferByteLength } =
  Object.getOwnPropertyDescriptor(arrayBufferPrototype, 'byteLength');
  
const { prototype: uint8ArrayPrototype } = Uint8Array;
const { set: uint8ArraySet } = uint8ArrayPrototype;
const { get: uint8ArrayBuffer } =
  Object.getOwnPropertyDescriptor(uint8ArrayPrototype, 'buffer');
```

packages/immutable-arraybuffer/src/immutable-arraybuffer-pony.js
```diff
@@ -0,0 +1,215 @@
...
+      buffer = arrayBufferTransferMaybe(buffer);
+      const oldLength = buffer.byteLength;
+      // eslint-disable-next-line @endo/restrict-comparison-operands
+      if (newLength <= oldLength) {
+        buffer = arrayBufferSlice(buffer, 0, newLength);
+      } else {
+        const oldTA = new Uint8Array(buffer);
+        const newTA = new Uint8Array(newLength);
+        newTA.set(oldTA);
+        buffer = newTA.buffer;
```
gibson042:
```suggestion
        apply(uint8ArraySet, newTA, [oldTA]);
        buffer = apply(uint8ArrayBuffer, newTA, []);
```

packages/immutable-arraybuffer/src/immutable-arraybuffer-shim.js
```diff
@@ -0,0 +1,103 @@
...
+   * @returns {ArrayBuffer} The sliced immutable ArrayBuffer.
+   */
+  sliceToImmutable(start = undefined, end = undefined) {
+    return sliceBufferToImmutable(this, start, end);
+  },
+
+  get immutable() {
+    return isBufferImmutable(this);
+  },
+};
```
gibson042, line 52: I think it's cleaner to avoid `moreMethods`.
```suggestion

  ...(tbtiMaybe
    ? {
        /**
         * Transfer the contents to a new Immutable ArrayBuffer
         *
         * @this {ArrayBuffer} buffer The original buffer.
         * @param {number} [newLength] The start index.
         * @returns {ArrayBuffer} The sliced immutable ArrayBuffer.
         */
        transferToImmutable(newLength = undefined) {
          return tbtiMaybe(this, newLength);
        },
      }
    : {}),
};
```

packages/immutable-arraybuffer/src/immutable-arraybuffer-shim.js
```diff
@@ -0,0 +1,103 @@
...
+  // we're about to overwrite a prior installation, we simply issue this
+  // warning and continue.
+  //
+  // TODO, if the primordials are frozen after the prior implementation, such as
+  // by `lockdown`, then this precludes overwriting as expected. However, for
+  // this case, the following warning text will be confusing.
+  console.warn(
+    'About to overwrite a prior implementation of "sliceToImmutable"',
+  );
+}
```
gibson042:
```suggestion
// Modern shim practice frowns on conditional installation, at least for
// proposals prior to stage 3. This is so changes to the proposal since
// an old shim was distributed don't need to worry about the proposal
// breaking old code depending on the old shim. Thus, if we detect that
// we're about to overwrite a prior installation, we simply issue this
// warning and continue.
//
// TODO, if the primordials are frozen after the prior implementation, such as
// by `lockdown`, then this precludes overwriting as expected. However, for
// this case, the following warning text will be confusing.
const overwrites = ownKeys(arrayBufferMethods).filter(
  key => key in arrayBufferPrototype,
);
if (overwrites.length > 0) {
  console.warn(
    `About to overwrite ArrayBuffer.prototype properties ${JSON.stringify(overwrites)}`,
  );
}
```

packages/immutable-arraybuffer/src/immutable-arraybuffer-pony.js
```diff
@@ -0,0 +1,215 @@
...
+const { apply, ownKeys } = Reflect;
+const { prototype: arrayBufferPrototype } = ArrayBuffer;
+
+const {
+  slice,
+  // TODO used to be a-ts-expect-error, but my local IDE's TS server
+  // seems to use a more recent definition of the `ArrayBuffer` type.
+  // @ts-ignore At the time of this writing, the `ArrayBuffer` type built
+  // into TypeScript does not know about the recent standard `transfer` method.
+  // Indeed, the `transfer` method is absent from Node <= 20.
```
kriskowal: What’s TO be DOne? Should be done or link an issue to track.

packages/immutable-arraybuffer/src/immutable-arraybuffer-shim.js
```diff
@@ -0,0 +1,103 @@
...
+   * @param {number} [end] The end index.
+   * @returns {ArrayBuffer} The sliced immutable ArrayBuffer.
+   */
+  sliceToImmutable(start = undefined, end = undefined) {
+    return sliceBufferToImmutable(this, start, end);
+  },
+
+  get immutable() {
+    return isBufferImmutable(this);
+  },
```
boneskull, line 36:
```suggestion
  /**
   * @this {ArrayBuffer}
   */
  get immutable() {
    return isBufferImmutable(this);
  },
```

It doesn't look like we're augmenting the lib-provided `ArrayBuffer` type anywhere to include the new members; maybe we should (at least until they make it into lib).  Doesn't need to be part of this PR though.  

If we did that, we could instead do away with the `@this` tags and use:

```js
/**
 * @type {Pick<ArrayBuffer, 'immutable' | 'sliceToImmutable'> & ThisType<ArrayBuffer>}
 */
const arrayBufferMethods = {
```

Something similar (using `ThisType`) is also possible if we created static types for `immutable` and `sliceToImmutable` somewhere else—even if they are not strictly members of `ArrayBuffer`.

packages/immutable-arraybuffer/src/immutable-arraybuffer-pony.js
```diff
@@ -0,0 +1,215 @@
...
+const makeImmutableArrayBufferInternal = realBuffer => {
+  const result = { __proto__: ImmutableArrayBufferInternalPrototype };
+  buffers.set(result, realBuffer);
+  return /** @type {ArrayBuffer} */ (/** @type {unknown} */ (result));
+};
+// Since `makeImmutableArrayBufferInternal` MUST not escape,
+// this `freeze` is just belt-and-suspenders.
+Object.freeze(makeImmutableArrayBufferInternal);
+
+export const isBufferImmutable = buffer => buffers.has(buffer);
```
boneskull, line 186:
```suggestion
/**
 * @param {ArrayBuffer} buffer
 * @returns {boolean}
 */
export const isBufferImmutable = buffer => buffers.has(buffer);
```

boneskull (commented):
I am too ignorant to approve this. I will, however, test it with LavaMoat once it's shipped. 😄

packages/immutable-arraybuffer/src/immutable-arraybuffer-pony.js
erights: Good point!

kriskowal: We should adopt the eslint rule for no-polymorphic etc. If we do, this will have to be slightly different: `weakmapGet` & al at call sites, with uncurryThis.

packages/immutable-arraybuffer/src/immutable-arraybuffer-pony.js
erights: Seems to no longer be necessary. Gone.

packages/immutable-arraybuffer/src/immutable-arraybuffer-pony.js
```diff
@@ -0,0 +1,241 @@
...
+   * See https://github.com/endojs/endo/tree/master/packages/immutable-arraybuffer#purposeful-violation
+   */
+  [Symbol.toStringTag]: 'ImmutableArrayBuffer',
+};
+
+// Better fidelity emulation of a class prototype
+for (const key of ownKeys(ImmutableArrayBufferInternalPrototype)) {
+  defineProperty(ImmutableArrayBufferInternalPrototype, key, {
+    enumerable: false,
+  });
```
kriskowal, line 157: Today I learned that `defineProperty` can edit a descriptor, not just replace.

packages/immutable-arraybuffer/package.json
```diff
@@ -25,7 +25,6 @@
   "exports": {
     ".": "./index.js",
     "./shim.js": "./shim.js",
-    "./shim-hermes.js": "./shim-hermes.js",
```
kriskowal, line 28: I trust that @leotm’s stamp here indicates that this is not a breaking change in any practical sense since the target customer does not yet depend on this export. Please confirm.

packages/immutable-arraybuffer/src/immutable-arraybuffer-shim.js
```diff
@@ -0,0 +1,81 @@
+import {
+  isBufferImmutable,
+  sliceBufferToImmutable,
+  optTransferBufferToImmutable,
+} from './immutable-arraybuffer-pony.js';
+
+// Even though the exported one is not a live binding, TS doesn't know that,
+// so it cannot do it's normal flow-based inference. By making a using a local
```
kriskowal:
```suggestion
// so it cannot do its normal flow-based inference. By making a using a local
```

packages/immutable-arraybuffer/src/immutable-arraybuffer-shim.js
```diff
@@ -0,0 +1,81 @@
...
+// copy, no problem.
+const optTBTI = optTransferBufferToImmutable;
```
kriskowal: I think I prefer using the inscrutable name for the temporary and using the scrutable name throughout the ordinary usage, for transparency and despite the verbosity.

packages/immutable-arraybuffer/src/immutable-arraybuffer-shim.js
```diff
@@ -0,0 +1,81 @@
...
+//
+// TODO, if the primordials are frozen after the prior implementation, such as
+// by `lockdown`, then this precludes overwriting as expected. However, for
+// this case, the following warning text will be confusing.
+const overwrites = ownKeys(arrayBufferMethods).filter(
+  key => key in arrayBufferPrototype,
+);
+if (overwrites.length > 0) {
+  console.warn(
+    `About to overwrite ArrayBuffer.prototype properties ${JSON.stringify(overwrites)}`,
```
kriskowal: I imagine this warning will be perplexing. Ideally, we could call the reader to action, like if the underlying platform supports a complete implementation of immutable ArrayBuffer, we can advise them to either withdraw `@endo/immutable-arraybuffer/shim.js` from their setup, or relayer their usage of `ses` in terms of its constituent shims so that is possible. We might be able to speculatively say “consider upgrading `ses` to a more recent major version, as we expect a future major release to withdraw the immutable-arraybuffer shim, possibly as soon as 2026.”

It’s also a bit of a shame that we don’t benefit from the lockdown reporting (mute) option, here, but perhaps it can’t be helped. We at least send to stderr on Node.js and color the log appropriately in DevTools with `console.warn`.

Attn, @naugtur please signal if muting this warning will be a sticking point.

kriskowal (approved):
With notes.

kriskowal:
Needs NEWS

```
diff --git a/packages/immutable-arraybuffer/NEWS.md b/packages/immutable-arraybuffer/NEWS.md
index d12ca09d3f..dc1a584999 100644
--- a/packages/immutable-arraybuffer/NEWS.md
+++ b/packages/immutable-arraybuffer/NEWS.md
@@ -7,6 +7,11 @@ User visible changes in `@endo/immutable-arraybuffer`:
   `@endo/immutable-arraybuffer/shim.js` does not interfere with this module's
   designed behavior.
 
+- Removes `@endo/immutable-arraybufer/shim-hermes.js` and absorbs the necessary
+  features into `@endo/immutable-arraybuffer/shim.js`.
+  We are not qualifying this as a breaking change since the feature did not
+  exist long enough to become relied upon.
+
 # 1.0.0 (2025-05-07)
 
 First release.
```

packages/immutable-arraybuffer/src/immutable-arraybuffer-shim.js
erights: Done

packages/immutable-arraybuffer/src/immutable-arraybuffer-shim.js
erights: Good idea! Done.

packages/immutable-arraybuffer/package.json
erights: (awaiting @leotm 's confirmation)

packages/immutable-arraybuffer/src/immutable-arraybuffer-shim.js
erights: Alternatively, since the proposal has already reached 2.7 and we're confident enough that it'll get through the rest of the process, we could instead ***not*** overwrite if we detect that these properties are already there. Basically, feature detect for a native implementation, and if it seems to be there, ***assume*** it is compat enough.

(awaiting feedback from @naugtur and @kriskowal )

erights:
> Removes `@endo/immutable-arraybufer/shim-hermes.js` and absorbs the necessary features into `@endo/immutable-arraybuffer/shim.js`. We are not qualifying this as a breaking change since the feature did not exist long enough to become relied upon.

Done, thanks.

packages/immutable-arraybuffer/src/immutable-arraybuffer-pony.js
```diff
@@ -0,0 +1,241 @@
...
+  structuredClone: optStructuredClone,
+} = globalThis;
+
+const { defineProperty, getPrototypeOf } = Object;
+const { apply, ownKeys } = Reflect;
+
+const { prototype: arrayBufferPrototype } = ArrayBuffer;
+const { slice, transfer: optTransfer } = arrayBufferPrototype;
+// @ts-expect-error TS doesn't know it'll be there
+const { get: arrayBufferByteLength } = Object.getOwnPropertyDescriptor(
```
kriskowal: I don’t expect there to be a meaningful difference in behavior, but you might capture `getOwnPropertyDescriptor` to be consistent with `defineProperty` and `getPrototypeOf`.

erights: Done. Pick up a few other such cases as well.

kriskowal (approved):

packages/immutable-arraybuffer/src/immutable-arraybuffer-pony.js
erights: I agree, but not in this PR. Resolving.

packages/immutable-arraybuffer/src/immutable-arraybuffer-shim.js
naugtur: From the perspective of "normal" hosts, I leave the decision to you.
As for Hermes and React Native, the issue may not be easily observable to the user.
At the time SES has to run repairIntrinsics before a lot of other code runs, the `console` is not wired up to the output that ends up being visible for the app developer. They won't see the warning unless it's being emitted after ReactNative core initializes.

If Hermes comes up with its own implementation if immutable arraybuffer, it's likely to be native and aiming for performance. Compliance with the standard is not guaranteed either. 
**I'd say from that point of view, not overwriting might be preferable for real-life use-cases** and potentially risky (the native implementation could have nonstandard fields that we'd need to repair)

Hermes people were discussing mutability of arraybuffers here https://github.com/facebook/hermes/issues/1578  (in case it adds perspective)

packages/immutable-arraybuffer/package.json
naugtur: I can confirm we have not released anything that would be damaged by this change.

erights: Thanks. Resolving.

packages/immutable-arraybuffer/src/immutable-arraybuffer-shim.js
erights: Thanks. Resolving.

PR #2856, opened 2025-06-12
feat(compartment-mapper): support mapping of undiscoverable packages

boneskull:
## Description

This adds the `additionalPackageLocations` and `additionalPackageDetails` options to `mapNodeModules()`. The former is a mapping of module location to one or more additional module locations. The latter is a data structure `mapNodeModules` will populate, which can then be provided to `captureFromMap` (which is in turn provided to `makeImportHookMaker()`.

This is intended to be used when dynamic requires prevent `mapNodeModules` & `captureFromMap` from otherwise creating a complete `CompartmentMapDescriptor` due to its ignorance of what would be dynamically required. Archival is not a use-case.

Also:

- Fixed signature of `chooseModuleDescriptor`
- Tweaked ESLint config to ignore my new fixture
- Add options type for `translateGraph`
- Fixed deprecation notice for `compartmentMapForNodeModules`; it was previously on the options object, but it should only be on the export.
- `graphPackages` now accepts an optional `Graph` object to be able to re-use the same graph between runs

### Questions for Reviewers

- Naming is hard and I was not sure what to name the new options. The `AdditionalPackageDetails` type looks like the existing `PackageDetails` type (with a new field) thus influencing the naming—but they are otherwise unrelated and because of this I did not extend `PackageDetails`.
- I am monkeying with `imports` in a `RecordModuleDescriptor` object in order to influence the module loader. Is there a better way? The things I stuff in `imports` are also _extremely dubious._

### Security Considerations

n/a

These changes do not meddle with systems loading untrusted code, afaik.

### Scaling Considerations

Use of the new option is not free.

n/a

### Testing Considerations

I did add some tests, but could probably use more.

No.

### Upgrade Considerations

This warrants an entry in `NEWS.md` since it is part of a public API.

### DIFF
--- a/packages/compartment-mapper/src/import-hook.js
+++ b/packages/compartment-mapper/src/import-hook.js
@@ -14,6 +14,7 @@
  * @import {
  *   ImportHook,
  *   ImportNowHook,
+ *   RecordModuleDescriptor,
  *   RedirectStaticModuleInterface,
  *   StaticModuleType,
  *   VirtualModuleSource
@@ -279,7 +280,7 @@ const nominateCandidates = (moduleSpecifier, searchSuffixes) => {
  * @param {ChooseModuleDescriptorParams} options Options/context
  * @param {Operators} operators Operators
  * @returns {Generator<ChooseModuleDescriptorYieldables,
- * StaticModuleType|undefined, Awaited<ChooseModuleDescriptorYieldables>>}
+ * RecordModuleDescriptor|undefined, Awaited<ChooseModuleDescriptorYieldables>>}
  * Generator
  */
 function* chooseModuleDescriptor(
@@ -381,7 +382,7 @@ function* chooseModuleDescriptor(
           compartment: packageLocation,
         };
       }
-      /** @type {StaticModuleType} */
+      /** @type {RecordModuleDescriptor} */
       const record = {
         record: concreteRecord,
         specifier: candidateSpecifier,
@@ -495,6 +496,7 @@ export const makeImportHookMaker = (
     entryCompartmentName,
     entryModuleSpecifier,
     importHook: exitModuleImportHook = undefined,
+    additionalPackageDetails = {},
   },
 ) => {
   // Set of specifiers for modules (scoped to compartment) whose parser is not
@@ -547,7 +549,6 @@ export const makeImportHookMaker = (
 
       // for lint rule
       await null;
-
       // All importHook errors must be deferred if coming from loading dependencies
       // identified by a parser that discovers imports heuristically.
       try {
@@ -621,6 +622,23 @@ export const makeImportHookMaker = (
         );
 
         if (record) {
+          if (additionalPackageDetails[compartmentDescriptor.location]) {
+            for (const {
+              moduleSpecifier: additionalPackageModuleSpecifier,
+              packageLocation: additionalPackageLocation,
+            } of additionalPackageDetails[compartmentDescriptor.location]) {
+              if (compartmentDescriptors[additionalPackageLocation]) {
+                // XXX: this "works" but that's probably a conincidence!
+                record.record?.imports?.push(
+                  `${compartmentDescriptors[additionalPackageLocation].name}${additionalPackageModuleSpecifier.substring(1)}`,
+                );
+              } else {
+                throw new Error(
+                  `Cannot find compartment descriptor for additional package location ${q(additionalPackageLocation)}`,
+                );
+              }
+            }
+          }
           return record;
         }
 
--- a/packages/compartment-mapper/src/node-modules.js
+++ b/packages/compartment-mapper/src/node-modules.js
@@ -28,7 +28,7 @@
  *   ReadFn,
  *   ReadPowers,
  *   SomePackagePolicy,
- *   SomePolicy,
+ *   AdditionalPackageDetails,
  * } from './types.js'
  * @import {
  *   Graph,
@@ -39,6 +39,7 @@
  *   GraphPackageOptions,
  *   GraphPackagesOptions,
  *   PackageDetails,
+ *   TranslateGraphOptions,
  * } from './types/node-modules.js'
  */
 
@@ -686,7 +687,7 @@ const graphPackages = async (
   commonDependencies,
   languageOptions,
   strict,
-  { log = noop } = {},
+  { log = noop, graph = create(null) } = {},
 ) => {
   const memo = create(null);
   /**
@@ -730,7 +731,6 @@ const graphPackages = async (
     };
   }
 
-  const graph = create(null);
   await graphPackage(
     packageDescriptor.name,
     readDescriptor,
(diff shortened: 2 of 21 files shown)


packages/compartment-mapper/src/import-hook.js
```diff
@@ -621,6 +622,23 @@ export const makeImportHookMaker = (
...
+          if (additionalPackageDetails[compartmentDescriptor.location]) {
+            for (const {
+              moduleSpecifier: additionalPackageModuleSpecifier,
+              packageLocation: additionalPackageLocation,
+            } of additionalPackageDetails[compartmentDescriptor.location]) {
+              if (compartmentDescriptors[additionalPackageLocation]) {
+                // XXX: this "works" but that's probably a conincidence!
+                record.record?.imports?.push(
+                  `${compartmentDescriptors[additionalPackageLocation].name}${additionalPackageModuleSpecifier.substring(1)}`,
+                );
```
boneskull: ⬆️ 🚨 ⬆️

boneskull: I'm not sure how this would play out at runtime, either.  At runtime, what we'll get is (probably) an absolute path to this thing; not `packageName/file.js`

packages/compartment-mapper/src/import-hook.js
```diff
@@ -621,6 +622,23 @@ export const makeImportHookMaker = (
...
+              } else {
+                throw new Error(
```
boneskull: the only way for this to happen would be to provide a broken `CompartmentMapDescriptor` in the first place. IDK if it's worth worrying about.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -761,19 +761,36 @@ const graphPackages = async (
...
+    if (node === undefined) {
+      throw new Error(
+        `Cannot find package at ${additionalPackageLocation} in graph`,
+      );
+    }
+    for (const {
+      packageLocation,
+      packageDescriptor: { name },
+    } of moduleDetails) {
+      node.dependencyLocations[name] ??= packageLocation;
```
boneskull: this seems to be sufficient for associating a `Node` with another `Node`, but maybe there's a safer way to do so?

packages/compartment-mapper/src/node-modules.js
```diff
@@ -1042,12 +1080,34 @@ export const compartmentMapForNodeModules = async (
...
+      const moduleDescriptor = values(
+        compartmentMap.compartments[additionalPackageLocation].modules,
+      ).find(({ compartment }) => compartment === packageLocation);
+      if (!moduleDescriptor) {
+        throw new Error(
+          `Cannot find module descriptor for ${packageLocation} in compartment descriptor ${additionalPackageLocation}`,
+        );
+      }
+      moduleDescriptor.retained = true;
+      compartmentMap.compartments[packageLocation].retained = true;
```
boneskull: I'm not sure these are actually needed.

- [x] @boneskull: check assumptions

packages/compartment-mapper/src/node-modules.js
```diff
@@ -1075,12 +1142,54 @@ export const mapNodeModules = async (
     parseLocatedJson(packageDescriptorText, packageDescriptorLocation)
   );
 
-  return compartmentMapForNodeModules(
+  await Promise.all(
```
boneskull, line 1311: afaict there is no memoization happening here that would make it so I shouldn't just spray promises everywhere

packages/compartment-mapper/test/capture-lite.test.js
```diff
@@ -87,3 +91,53 @@ test('captureFromMap() should round-trip sources based on parsers', async t => {
...
+  const entryModuleLocation = new URL(
+    'fixtures-additional-modules/node_modules/goofy/index.js',
+    import.meta.url,
+  ).href;
+
+  const additionalModuleLocations = {
+    [entryModuleLocation]: [
+      new URL('fixtures-additional-modules/config.js', import.meta.url).href,
+    ],
+  };
```
boneskull: the fixture here mimics something like webpack.  we have a tool (`goofy`) and that tool dynamically loads a config file (`config.js`) from our application root.

packages/compartment-mapper/src/import-hook.js
naugtur: I'm expecting an explosion of edge cases here, starting with there being a bunch of record types and only one (or two?) have the structure of `record.record` so it's a little specific to what you've been testing with.
Unless no other record shape needs to be affected by this. 

I don't like the concept of stuffing imports on the records. It's obviously working if added early enough, but there might be a more gentle way.

---

I'd rather include the information necessary for this to work in the compartmentMap itself. What's preventing that? 

```js
  const compartmentMap = await mapNodeModules(readPowers, entryModuleLocation, {
    additionalModuleLocations,
    // ~~additionalPackageDetails~~
  });
```

If we managed to do that, the step to manually add imports here could end up being unnecessary. 

---


A single map of referrer-target should be enough to feed graphPackages with information to add missing links to the list it gets from packageDescriptor
Calling load on each of the targets in the same place where we call load on attenuators would suffice to keep them retained.

I'm not 100% sure the links are even necessary. If we had multiple entries, made graphPackages accept multiple (or call it many times if that works) and then proceed to load each of them, it should be enough to reach them through importNow flow.

packages/compartment-mapper/src/node-modules.js
naugtur: manually setting `retained` should not be necessary and if it is, it suggests a flaw in the design. 
Causing them to be retained should make other things easier too.

naugtur (commented):
Review in progress, but I think there might be a less (opinion) invasive way to serve the same need. 

That being said, I want to encourage others to review because this exists and my hypothetical other way doesn't (at least until I try)

packages/compartment-mapper/src/import-hook.js
naugtur: Here's a more precise wording of what I mean in a language that works better than English:
https://github.com/endojs/endo/pull/2864

boneskull:
@naugtur Regarding setting the `retained` flag:

AFAICT it _is_ needed, since we do not know what will happen with the compartment descriptor map after it exits `mapNodeModules()`.

If it is provided to `loadFromMap()`, then you're right; setting it shouldn't be necessary.

If it is provided to `captureFromMap()`, then it _is_ necessary because no execution occurs and there's no other way to tell the "digest" function that we should keep these compartments around.

Whether or not to set this bit could potentially be a flag to `mapNodeModules`, I suppose.

The latest change will actually set this flag recursively. It should be possible to narrow the set by determining which compartment descriptors are _only_ referenced by the additional modules and other like descriptors.  I haven't done that yet, but I think it makes sense to do since the additional module may reference the entry point (so we've just flagged the whole compartment map).

An alternative design (which may be a better idea, but idk) would involve fiddling with the digest function to force it to keep certain compartments.

packages/compartment-mapper/src/import-hook.js
boneskull: I've nixed this bit

packages/compartment-mapper/test/capture-lite.test.js
boneskull: This test fixture was dropped in lieu of the one created in the target branch.

boneskull:
working on sorting out the merge conflict

PR #2860, opened 2025-06-16
refactor(pass-style): possibly faster isObject

erights:

Closes: #XXXX
Refs: #XXXX

## Description

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations


packages/pass-style/src/passStyle-helpers.js
```diff
@@ -31,7 +31,8 @@ export const hasOwnPropertyOf = (obj, prop) =>
 harden(hasOwnPropertyOf);
 
 // TODO try typing this; `=> val is {} too narrow, implies no properties
-export const isObject = val => Object(val) === val;
+export const isObject = val =>
+  val && (typeof val === 'object' || typeof val === 'function');
```
gibson042:
```suggestion
  !!val && (typeof val === 'object' || typeof val === 'function');
```

packages/pass-style/src/passStyle-helpers.js
```diff
@@ -31,7 +31,8 @@ export const hasOwnPropertyOf = (obj, prop) =>
 harden(hasOwnPropertyOf);
 
 // TODO try typing this; `=> val is {} too narrow, implies no properties
```
gibson042: It might be time to address this TODO, either accepting that implication of no [known] properties is appropriate:
```suggestion
/** @type {(val: unknown) => val is (Function | {})} */
```
or allowing lookup of arbitrary properties:
```suggestion
/** @type {(val: unknown) => val is (Function | Record<string | symbol, unknown>)} */
```

packages/pass-style/src/passStyle-helpers.js
```diff
@@ -31,7 +31,8 @@ export const hasOwnPropertyOf = (obj, prop) =>
...
-export const isObject = val => Object(val) === val;
+export const isObject = val =>
```
gibson042, line 48: As noted elsewhere, the `isObject` name fails to communicate expected classification of inputs that are functions... this predicate might be better named as something like `canOwnProperties` (_and some `isObject` call sites might **actually** want to use e.g. `export const hasObjectType = val => !!val && typeof val === 'object';` rather than treating functions like property bags_). Maybe we should introduce the less ambiguous name and deprecate `isObject` as an alias.

gibson042 (approved):
This change has minor impact in V8, but `Object(value)` must be very slow in XS because the speedup there is significant.

packages/pass-style/src/passStyle-helpers.js
erights: Good point. I'm going with a new `isPrimitive` as the replacement for the deprecated `isObject`.

erights: Corroborating your point about the potentially confusing terminology, https://github.com/Agoric/agoric-sdk/blob/51ee2e06758e4e8a474cbee7b2f2645cd055bf36/packages/agoric-cli/src/lib/format.js#L73 defines its own `isObject` that excludes functions.
```js
const isObject = x => typeof x === 'object' && x !== null;
```

packages/module-source/test/fixtures/exportheavy.js
```diff
@@ -43,7 +44,7 @@ export {
 export * from './src/types.js';
 
 
-const { details: X, Fail } = assert;
+export { X, Fail } from '@endo/errors';
```
erights, line 47: This is a pure-refactor drive-by

packages/pass-style/src/passStyle-helpers.js
erights: Done!

packages/pass-style/src/passStyleOf.js
```diff
@@ -359,11 +359,11 @@ export const toThrowable = specimen => {
...
-  // Note that this step will fail if `specimen` would be a passable container
-  // except that it contains non-passable errors that could be converted.
-  // This will need to be fixed to do the TODO above.
-  const passStyle = passStyleOf(specimen);
-  if (isObject(specimen)) {
+  if (!isPrimitive(specimen)) {
+    // Note that this step will fail if `specimen` would be a passable container
+    // except that it contains non-passable errors that could be converted.
+    // This will need to be fixed to do the TODO above.
+    const passStyle = /** @type {PassStyle} */ (passStyleOf(specimen));
```
erights, line 366: Moving the call to `passStyleOf` inside the conditional is a drive-by. But typescript's imprecision about the negation of an `is` result type provoked the extra cast.

packages/pass-style/src/types.d.ts
```diff
@@ -25,11 +25,24 @@ export type AtomStyle =
   | 'byteArray'
   | 'symbol';
 
+/**
+ * The actual JS primitive types.
+ */
+export type JSPrimitive =
```
erights, line 31: Made it easier to define the type of `isPrimitive`.

packages/ses/src/permits-intrinsics.js
```diff
@@ -131,8 +131,8 @@ export default function removeUnpermittedIntrinsics(
    * Validate the object's [[prototype]] against a permit.
    */
   function visitPrototype(path, obj, protoName) {
-    if (!isObject(obj)) {
-      throw TypeError(`Object expected: ${path}, ${obj}, ${protoName}`);
+    if (isPrimitive(obj)) {
+      throw TypeError(`Object expected: ${path}, ${String(obj)}, ${protoName}`);
```
erights, line 135: The better typing of `isPrimitive` (thanks @gibson042) provoked typescript into reporting this genuine bug as a type error. So this bug fix is more than a drive-by.

packages/pass-style/src/passStyle-helpers.js
erights: Done

packages/pass-style/src/passStyle-helpers.js
erights: The second. Done.

erights:
> limit their exposure to corruption that might follow from a poorly implemented so-called vetted shim that runs after ses initialization but before calling lockdown

Also, to be a proper shim of the (to be) proposed standard HardenedJS of what happens if code running in this interval does strange things. Presumably, XS is already is conformance to what we expect to propose, so this also increases the similarity of the ses and XS implementations of HardenedJS.

PR #2861, opened 2025-06-17
fix(pass-style): Tolerate non-existence of immutable ArrayBuffer

kriskowal:
The most recent releases of ses and pass-style introduce support for immutable ArrayBuffer with a pass-style of “byteArray” in anticipation of the upcoming JavaScript standard. Although versions of pass-style and ses that were published together work together, pass-style on older versions of ses on platforms that predate immutable ArrayBuffer (virtually all), pass-style does not tolerate the non-existence of immutable `ArrayBuffer.prototype.transferToImmutable` whether shim or native. This change makes pass-style suitable on all previously supported platforms.

https://github.com/endojs/endo/releases/tag/ses%401.13.0

### DIFF
--- a/packages/pass-style/src/byteArray.js
+++ b/packages/pass-style/src/byteArray.js
@@ -7,21 +7,45 @@ import { X, Fail } from '@endo/errors';
 const { getPrototypeOf, getOwnPropertyDescriptor } = Object;
 const { ownKeys, apply } = Reflect;
 
-// @ts-expect-error TODO How do I add it to the ArrayBuffer type?
-const AnImmutableArrayBuffer = new ArrayBuffer(0).transferToImmutable();
+// Detects the presence of a immutable ArrayBuffer support in the underlying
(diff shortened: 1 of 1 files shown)


erights:
> coæval

I don't get it. ?

packages/pass-style/src/byteArray.js
```diff
@@ -36,7 +60,7 @@ export const ByteArrayHelper = harden({
     (!!check && check(false, X`Immutable ArrayBuffer expected: ${candidate}`)),
 
   assertRestValid: (candidate, _passStyleOfRecur) => {
-    getPrototypeOf(candidate) === ImmutableArrayBufferPrototype ||
+    getPrototypeOf(candidate) === immutableArrayBufferPrototype ||
```
erights, line 63: I stared and stared to figure out what's different until I noticed the casing. Your casing is better, thanks.

erights (approved):
LGTM, thanks!

PR #2862, opened 2025-06-17
chore(pass-style): Patch release v1.6.1

kriskowal:
Releases patch in #2861

### DIFF
--- a/packages/pass-style/NEWS.md
+++ b/packages/pass-style/NEWS.md
@@ -1,5 +1,19 @@
 User-visible changes in `@endo/pass-style`:
 
+# v1.6.1 (2024-06-17)
+
+- Fixes so that the package initializes on platforms that lack
+  `ArrayBuffer.prototype.transferToImmutable`.
+
+# v1.6.0 (2024-06-02)
+
+**BROKEN BUT PATCHED** in 1.6.1, this version introduced a dependence on the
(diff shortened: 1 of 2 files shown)


packages/pass-style/NEWS.md
```diff
@@ -1,5 +1,19 @@
...
+  `ArrayBuffer.prototype.transferToImmutable`.
+
+# v1.6.0 (2024-06-02)
+
+**BROKEN BUT PATCHED** in 1.6.1, this version introduced a dependence on the
+underlying platform supporting `ArrayBuffer.prototype.transferToImmutable`.
+The patch restores the ability to use `pass-style` on older platforms without
+the immutable `ArrayBuffer` shim (as entrained by `ses`).
+
+- Introduces support for `byteArray`.
```
michaelfig, line 15: For the future, I think this text could be worded to avoid discerning the referents for "this version" and "the patch".

```suggestion
**v1.6.0 is BROKEN BUT PATCHED in v1.6.1**

v1.6.0 accidentally introduced a reliance on the underlying platform
to support `ArrayBuffer.prototype.transferToImmutable`.  The patch
in the newer v1.6.1 restores the ability to use `@endo/pass-style`
even on older platforms without `transferToImmutable`.

- Introduces support for `byteArray`.
```

PR #2864, opened 2025-06-18
PoC multiple entries in compartment-mapper

naugtur:
This is not meant to be merged as-is but used in a discussion first.

I believe adding an option to have additional entrypoints to the endo app solves a few issues with inflexibility of compartment-mapping when applied to lavamoat.
This is a change that feels more native to how compartment-mapper behaves than stuffing imports and setting retained to allow dynamically importing from locations indicated by the user.

The test without policy works fine and without abusing compartmentMapper beyond what's clearly indicated (the mock isAdditionalEntry function where I'm not sure if I even want to drill down that information or just allow everything)

As of the 3rd commit here, policy enforcement on the dynamically loaded entry  doesn't work because it can't be identified with the contents of compartmentMap. I might find a way around that, if not - the input to importLocation would need to change to also contain information about the linkage to put in place (which feels like an annoyance to the end-user, even in lavamoat - we might be able to help them with that)

---

Added 4th commit with an implementation of a dynamic policy lookup at runtime. 
- What I like about it: eliminates the need to resolve items within the package that's being imported from ahead of time to put them in modules
- What I don't like about it: compartmentDescriptor.path was not used for runtime policy lookups before and there was less damage a confused contributor could do to policy enforcement. (could be mitigated by introducing a brand-check on compartmentDescriptor.path values. I'll start by freezing them)

### DIFF
--- a/packages/compartment-mapper/src/import-hook.js
+++ b/packages/compartment-mapper/src/import-hook.js
@@ -46,6 +46,7 @@ import {
   attenuateModuleHook,
   ATTENUATORS_COMPARTMENT,
   enforceModulePolicy,
+  enforcePackagePolicyDynamic,
 } from './policy.js';
 import { unpackReadPowers } from './powers.js';
 
@@ -205,11 +206,14 @@ const findRedirect = ({
       }
       // verify policy and if allowed, load the entry even without a link between compartments
       const isAdditionalEntry = c => true; // good enough approximation for PoC
-      if (isAdditionalEntry(someCompartmentDescriptor)) {
-        enforceModulePolicy(
-          someCompartmentDescriptor.path[
-            someCompartmentDescriptor.path.length - 1
-          ],
+      if (
+        has(someCompartmentDescriptor, 'path') &&
+        someCompartmentDescriptor.path &&
+        isAdditionalEntry(someCompartmentDescriptor)
+      ) {
+        enforcePackagePolicyDynamic(
+          absoluteModuleSpecifier,
+          someCompartmentDescriptor.path,
           compartmentDescriptor,
           {
             errorHint: `Blocked in import hook. ${q(absoluteModuleSpecifier)} is part of the compartment map and resolves to ${location}`,
--- a/packages/compartment-mapper/src/node-modules.js
+++ b/packages/compartment-mapper/src/node-modules.js
@@ -55,7 +55,7 @@ import {
 import { unpackReadPowers } from './powers.js';
 import { search, searchDescriptor } from './search.js';
 
-const { assign, create, keys, values, entries } = Object;
+const { assign, create, keys, values, entries, freeze } = Object;
 
 const decoder = new TextDecoder();
 
@@ -516,7 +516,7 @@ const graphPackage = async (
 
   assign(result, {
     name,
-    path: logicalPath,
+    path: freeze(logicalPath),
     label: `${name}${version ? `-v${version}` : ''}`,
     sourceDirname,
     explicitExports: exportsDescriptor !== undefined,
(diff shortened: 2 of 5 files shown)


packages/compartment-mapper/src/import-hook.js
```diff
@@ -171,14 +181,36 @@ const findRedirect = ({
...
+        );
+        return {
+          specifier: relativeSpecifier(moduleSpecifierLocation, location),
+          compartment: compartments[location],
+        };
+      }
+      // verify policy and if allowed, load the entry even without a link between compartments
+      const isAdditionalEntry = c => true; // good enough approximation for PoC
+      if (isAdditionalEntry(someCompartmentDescriptor)) {
+        enforceModulePolicy(
```
naugtur: This doesn't work yet

packages/compartment-mapper/src/import-lite.js
```diff
@@ -227,6 +228,11 @@ export const loadFromMap = async (readPowers, compartmentMap, options = {}) => {
 
     await pendingJobsPromise;
 
+    if(otherEntrypoints) {
+      await Promise.all(otherEntrypoints.map(entry => compartment.load(entry)))
```
naugtur, line 232: This is all we need to get them retained etc.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -1048,6 +1026,93 @@ export const compartmentMapForNodeModules = async (
...
+  const preferredPackageLogicalPathMap = new Map();
+
+  const firstEntry = entriesToProcess[0];
+
+  for (const entrypoint of entriesToProcess) {
+    const { packageLocation, packageDescriptor, moduleSpecifier } = entrypoint;
+    const logicalPath =
+      entrypoint === firstEntry
+        ? []
+        : [`$external:${packageDescriptor.name || packageLocation}`];
```
naugtur, line 997: Not super happy with this yet. The discrepancy between item in path and package name might be getting in the way. 
I've considered overriding `packageDescriptor.name` for compartment map. Not sure what consequences that would have.

Anyway, the two must be true:
1. only the main entrypoint has an empty path
2. paths of other entrypoints must not collide with packages in dependencies

packages/compartment-mapper/src/node-modules.js
```diff
@@ -1048,6 +1026,93 @@ export const compartmentMapForNodeModules = async (
...
+  const conditions = new Set(conditionsOption || []);
+
+  // dev is only set for the entry package, and implied by the development
+  // condition.
+  // The dev option is deprecated in favor of using conditions, since that
+  // covers more intentional behaviors of the development mode.
+
+  const memo = create(null);
+  const graph = create(null);
+  const preferredPackageLogicalPathMap = new Map();
```
naugtur, line 988: Sharing the 3 makes any number of entries produce a shared graph

packages/compartment-mapper/src/node-modules.js
```diff
@@ -975,22 +953,19 @@ const makeLanguageOptions = ({
...
- * @param {string} packageLocation
+ * @param {Array<{packageLocation:string, packageDescriptor:PackageDescriptor, moduleSpecifier:string}>} entriesToProcess
  * @param {Set<string>} conditionsOption
- * @param {PackageDescriptor} packageDescriptor
- * @param {string} moduleSpecifier
  * @param {CompartmentMapForNodeModulesOptions} [options]
  * @returns {Promise<CompartmentMapDescriptor>}
  * @deprecated Use {@link mapNodeModules} instead.
  */
 export const compartmentMapForNodeModules = async (
```
naugtur, line 962: Yes, this is a breaking change to a once-public deprecated function. We'd have to make a facade that translates to this.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -171,14 +182,39 @@ const findRedirect = ({
...
+            errorHint: `Blocked in import hook. ${q(absoluteModuleSpecifier)} is part of the compartment map and resolves to ${location}`,
+          },
+        );
+        return {
+          specifier: relativeSpecifier(moduleSpecifierLocation, location),
+          compartment: compartments[location],
+        };
+      }
+      // verify policy and if allowed, load the entry even without a link between compartments
+      const isAdditionalEntry = c => true; // good enough approximation for PoC
```
naugtur, line 208: The ways planets seem to have aligned, it'd be best for this to actually limit to just the specified entrypoints.

PR #2865, opened 2025-06-19
feat(pola-io): least-authority file, net, cmd access

dckc:
## Description

access to files, net, cmd with attentuation and such. see README for more.

### Security Considerations

DRAFT until:

 - [ ] in the interest of ZERO dependencies, uses `freeze()` rather than `harden()`. expected to be compatible with HardenedJS / ses. **what hazards does this expose us to?**
   - [x] includes unit tests that the API surface is frozen
 - [ ] by default, doesn't enforce downward-only file traversal

### Scaling Considerations

n/a/

 - [ ] `dyn()` part of file API needs more docs

### Testing Considerations

 - [x] some basic unit tests
 - [x] API surface is frozen
 - [ ] see CONTRIBUTING for more things to demonstrate by way of tests, if not features

I think it overlaps with some `readPowers` stuff in compartment mapper?

migrating from agoric-sdk:
 - https://github.com/Agoric/agoric-sdk/pull/11486

### Upgrade Considerations

 - [ ] not sure what the `exports` should be
   - I removed `exports` from the standard `create-package.sh` output.


packages/pola-io/src/index.js
```diff
@@ -0,0 +1,3 @@
+export * from './file.js';
+export * from './cmd.js';
+export * from './net.js';
```
kriskowal, line 3: What do you think of omitting the `index.js` “barrel module” entirely to encourage importing the narrowest necessary set of modules? PoLA works for reducing heap/bundle sizes and startup times.

packages/pola-io/src/file.js
```diff
@@ -0,0 +1,118 @@
...
+      /** @param {string} [suffix] */
+      basename: suffix => pathio.basename(there, suffix),
+      existsSync: () => fsio.existsSync(there),
+      /** @returns {Promise<FileRd[]>} */
+      readdir: () =>
+        fspio
+          .readdir(there)
+          .then(files => files.map(segment => self.join(segment))),
+    };
+    return freeze(self);
```
kriskowal, line 80: I gather we’re doing a shallow freeze here for portability with and without `harden` in the environment. That leaves the underlying functions unfrozen. Maybe https://github.com/endojs/endo/pull/2782 will help.

packages/pola-io/package.json
```diff
@@ -0,0 +1,71 @@
...
+    "url": "git+https://github.com/endojs/endo.git",
+    "directory": "packages/pola-io"
+  },
+  "bugs": {
+    "url": "https://github.com/endojs/endo/issues"
+  },
+  "type": "module",
+  "main": "./src/index.js",
+  "types": "./src/types.ts",
+  "$exports": "TODO",
```
kriskowal, line 21: I think we should omit `"main"` and `index.js` entirely.

Our current stance on `"exports"` is that all module specifiers should match file corresponding file paths, like `"./package.json": "./package.json"` because that ensures that any tool that hasn’t caught up with Node.js 19, when `"exports"` was introduced, will still be able to import the referent. This rule creates some friction with the `"./src"` subdirectory, which should not appear in the exports namespace of the package. That is, I would prefer to establish usage like:

```js
import { makeCommand } from "@endo/pola-io/command.js"`
```


```js
// not this
import { makeCommand } from "@endo/pola-io/src/command.js"`
```

To do that, we have the option of moving files up from `src` or creating a “thunk” for each like:

```js
// command.js
export { makeCommand } from './src/command.js';
```

I like being explicit about which exports of the proxied module are exported, since that gives `./src/command.js` the liberty to export internal functions just for the purposes of unit tests, without making them public API surface.

I do not have a strong preference either way and you will see both in Endo.

Ideally, we enough time will pass that we can rely on all of our tooling to support `"exports"`. That might be today already, but I am not sure. In the long term, I would be delighted if we could make the public usage for a package like this look more like:

```js
import { makeCommand } from '@endo/pola-io/command';
```

Noting the absence of the `.js` extension. But, that is such a small wart, I’ve not spent any energy evaluating whether this is the right time to begin that shift in stance.

kriskowal (commented):
Preliminarily, I think we should have this and more like it.

You mention that there are overlapping concerns with the `powers.js` conventions in Compartment Mapper (and elsewhere). The substance of those conventions is that the web/node/xs portable portion of a package should be the primary functionality exported by a package, and the package *may* provide convenient adapters for constructing those powers from the material laying around in Node.js, the web, or whereever. This package, like `@endo/stream-node`, adapts Node.js APIs to a more portable, netural, and ocappy interface. So, I’d expect to import the powers types from `@endo/pola-io` and have `-node-` or `-web-` and `-powers.js` somewhere in the module specifier to indicate which platform the power maker adapts. I chose to make `make*Powers` functions pluck methods in-situ from Node.js modules for ease-of-use, but without obviating the option of letting the caller pluck the methods themselves.

I think `fetch` highlights a need for a category of `powers.js` that can be made from material that is laying around both on the web and node, but I also would be content with using `@endo/pola-io/web-powers.js` to make web powers even if I’m using the `fetch` as implemented in Node.js.

I think there’s an opportunity to build on `@endo/stream` and `@endo/stream-node` with this package, but not necessarily in the first cut. I believe you’ve left the requisite void for that feature to fill, in the way that the `fetch` API produces a `Readable` that in turn has `text()`, `json()`, but also a `body` `ReadableStream`, we have the option of adding a streaming read method to a file cell later. Ideally, we share an abstraction between readable cells on the web, on a file system, in a zip file.

The pet daemon has an Exo for a readable cell of its CAS. That overlaps too, but this is beneath the Exo layer. We should just have a coherent approach. The Readable cell in Dæmon happens to have `text()` and `json()` like the web’s fetch/ReadableStream, and also `sha512()` since that’s laying around. I think Exo-layer streaming is still an unsettled design issue, but at *this* layer definitely should be `stream()` returning an AsyncIterator or AsyncIterable of Uint8Array. At the Exo layer, that’ll be hardened immutable ArrayBuffers over some other streaming protocol that may resemble iterator protocol, but maybe not.

I am hoping that I can convince you to avoid abbreviation in the public names exported by this package, like `makeCommand` over `makeCmd`, `makeReadableFile` and `makeMutableFile` over `makeFileRd` and `makeFileRW`.

packages/pola-io/src/index.js
dckc: no strong feelings either way

dckc: an `@endo/pola-io/kitchen-sink` export might be nice.

packages/pola-io/src/file.js
dckc: > That leaves the underlying functions unfrozen.

I have only the vaguest notion of what that means.

Is it straightforward to demonstrate by example how one can detect that this is the case?
to exploit it?

dckc:
> avoid abbreviation in the public names exported by this package, like makeCommand over makeCmd

yes, dictionary words are good.

packages/pola-io/src/file.js
kriskowal: `Object.isFrozen(self.readdir) === false`.

packages/pola-io/src/index.js
kriskowal: Yeah, I think I would be okay with an export as pithy as `@endo/pola-io/all.js`. Seems like enough to discourage over-dependency without denying convenience outright. Though, I’m sure folks will expect `@endo/pola-io` to be a barrel export if a barrel export exists.

packages/pola-io/src/cmd.js
```diff
@@ -0,0 +1,65 @@
...
+  // @ts-expect-error undefined is filtered out
+  const skipUndef = Object.entries(record).filter(([_k, v]) => v !== undefined);
+  return skipUndef.flatMap(([key, value]) => {
+    if (Array.isArray(value)) {
+      return value.flatMap(v => [`--${key}`, v]);
+    }
+    return [`--${key}`, value];
+  });
+};
+freeze(flags);
```
gibson042, line 65: I'm recommending extraction of this POLA-independent helper, cf. https://github.com/Agoric/agoric-sdk/pull/11576#discussion_r2190734206
```suggestion
```

dckc: It's part of an attenuation mechanism: For example, I can take a `agd` command runner and attenuate it so that `--from` is always set to `agoric1xyz....`.

IOU tests / examples.

erights:
> DRAFT until

This PR looks like it still addresses an important need. What are our expectations for this PR? Since all the changes are only in the new package, I imagine that it would be easy to reconcile with current master?

dckc:
> > DRAFT until
> 
> This PR looks like it still addresses an important need.

:)

> What are our expectations for this PR?

I'd like it to land in due course. I managed to get it into agoric-sdk as part of ymax work:

 - https://github.com/Agoric/agoric-sdk/pull/11488

And that's meeting my needs for now. I'd be happy for somebody to carry this forward.

> Since all the changes are only in the new package, I imagine that it would be easy to reconcile with current master?

yes.

I suppose the main thing is [Object.freeze vs. harden](#discussion_r2164765290) above.

PR #2867, opened 2025-06-24
refactor: Consolidate ses TCB eslint configuration

kriskowal:
## Description

This sequence of changes consolidates the eslint rules we apply to SES into an endo eslint-plugin configuration we can extend to cover the packages that participate in ses initialization, the Trusted Compute Base of SES. 

### Security Considerations

These packages all strive to limit their exposure to corruption that might follow from a poorly implemented so-called vetted shim that runs after ses initialization but before calling lockdown. This change brings more lint rules to bear on these ejected packages and raises those packages up toward the required bar.

### Scaling Considerations

None.

None.

### Testing Considerations

Covered by lint. Some changes motivated by discovery in lint.

None, except to the extent we rely on vetted shims to not become dependent on their ability to subvert ses, which we consider a defect.

### Upgrade Considerations

None.


gibson042 (approved):

erights:
> limit their exposure to corruption that might follow from a poorly implemented so-called vetted shim that runs after ses initialization but before calling lockdown

Also, to be a proper shim of the (to be) proposed standard HardenedJS of what happens if code running in this interval does strange things. Presumably, XS is already is conformance to what we expect to propose, so this also increases the similarity of the ses and XS implementations of HardenedJS.

erights (approved):
LGTM, thanks!

PR #2868, opened 2025-06-24
fix(patterns): use `qp` for nested pattern error messages

erights:
Closes: #XXXX
Refs: #2799 

## Description

#2799 added a new `q`-like quoter named `qp`, for "quote passable", that renders any value into Justin source text. This was specifically motivated for displaying patterns within error messages in a readable and informative manner.

The `@endo/patterns` package itself formulates many error messages to explain why a specimen did not match a pattern. Some of these message include quoted forms of nested patterns. But those were still quotes with `q`, leading @dckc to report still seeing error messages like

```
Error {
    message: '... - Must match one of ["[match:splitRecord]","[match:splitRecord]"]',
  }
```

This PR changes the quoting of nested errors to use `qp` instead, so a `[match:splitRecord]` might instead render as
```js
`makeTagged("match:splitRecord", [
  {
    assetKind: makeTagged("match:or", [
      "nat",
      "set",
      "copySet",
      "copyBag",
    ]),
    decimalPlaces: makeTagged("match:and", [
      makeTagged("match:gte", -100),
      makeTagged("match:lte", 100),
    ]),
  },
])`
```

This is more verbose. But error messages are designed to carry the critical clues to help find a bug. For nested patterns, this extra information is worth the verbosity.

### Security Considerations

We were already quoting nested patterns with `q`, unredacting them, under the assumption that specimens may carry secrets worth protecting from callers, but patterns typically do not. Switching these from `q` to `qp` is consistent with that assumption, but actually exposes info that `q` happened not to expose. When unredacting happened not to expose secrets it said were ok to expose, it is easy for developers not to notice the secrets that could have been shown but weren't. After this PR, these unredacted secrets will be exposed to callers.

### Scaling Considerations

Where `q` is lazy, `qp` is eager, rendering its argument up front into Justin even if the containing error is never created or thrown. Thus, we must take care to only use `qp` when we're already committed to creating and throwing the error. I believe all such changes in this PR do so, and so do not add overhead to the happy path.

I don't know what documentation we provide explaining the interpretation of error messages. If we have any explain pattern match error messages, this PR might need us to change those explanations.

### Testing Considerations

Using `qp` to formulate the error messages definitely makes it more awkward to test is the error message is as expected. You can see this in the tests this PR needed to change. Fortunately, as shown in these error message tests, you can use `${q(qp(...))}` to formulate the golden in the test.

Because this PR changes the error messages, as pointed out above, golden tests of error message text may no longer match, ***causing old working tests to start failing***. However, this is not ***technically*** a compat break, because we take the stance that the content of error messages is not normative, so the error messages can become more informative over time with a version bump. Thus, I use `:` rather than `!` to recommend that this PR not cause a version bump. (See https://github.com/endojs/endo/pull/2868#pullrequestreview-2955029970 below)

### Upgrade Considerations

None.

### DIFF
--- a/packages/patterns/NEWS.md
+++ b/packages/patterns/NEWS.md
@@ -1,5 +1,9 @@
 User-visible changes in `@endo/patterns`:
 
+# Next release
+
+- In errors explaining why a specimen does not match a pattern, sometimes the error message contains a quoted form of a nested pattern. This quoting was done with `q`, producing an uninformative rendering of these nested patterns. Not this quoting is done with `qp`, which renders these nested patterns into readable Justin source code.
+
 # v1.5.0 (2025-03-11)
 
 - New pattern: `M.containerHas(elementPatt, bound = 1n)` motivated to support want patterns in Zoe, to pull out only `bound` number of elements that match `elementPatt`. `bound` must be a positive bigint.
--- a/packages/patterns/NEWS.md
+++ b/packages/patterns/NEWS.md
@@ -2,7 +2,7 @@ User-visible changes in `@endo/patterns`:
 
 # Next release
 
-- In errors explaining why a specimen does not match a pattern, sometimes the error message contains a quoted form of a nested pattern. This quoting was done with `q`, producing an uninformative rendering of these nested patterns. Not this quoting is done with `qp`, which renders these nested patterns into readable Justin source code.
+- In errors explaining why a specimen does not match a pattern, sometimes the error message contains a quoted form of a nested pattern. This quoting was done with `q`, producing an uninformative rendering of these nested patterns. Now this quoting is done with `qp`, which renders these nested patterns into readable Justin source code.
 
 # v1.5.0 (2025-03-11)
 
--- a/packages/patterns/NEWS.md
+++ b/packages/patterns/NEWS.md
@@ -2,7 +2,7 @@ User-visible changes in `@endo/patterns`:
 
 # Next release
 
-- In errors explaining why a specimen does not match a pattern, sometimes the error message contains a quoted form of a nested pattern. This quoting was done with `q`, producing an uninformative rendering of these nested patterns. Now this quoting is done with `qp`, which renders these nested patterns into readable [Justin](https://github.com/agoric-labs/jessica/blob/master/lib/quasi-justin.js.ts) source code.
+- In errors explaining why a specimen does not match a pattern, sometimes the error message contains a quoted form of a nested pattern. This quoting was done with `q`, producing an uninformative rendering of these nested patterns. Now this quoting is done with `qp`, which renders these nested patterns into readable [Justin](https://github.com/endojs/Jessie/blob/main/packages/parse/src/quasi-justin.js) source code.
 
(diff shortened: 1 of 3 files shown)


dckc (commented):
works for me, but I defer to more active maintainers (@kriskowal ) to approve

for golden tests, I find ava snapshots pretty handy

packages/patterns/NEWS.md
```diff
@@ -1,5 +1,9 @@
 User-visible changes in `@endo/patterns`:
 
+# Next release
+
+- In errors explaining why a specimen does not match a pattern, sometimes the error message contains a quoted form of a nested pattern. This quoting was done with `q`, producing an uninformative rendering of these nested patterns. Now this quoting is done with `qp`, which renders these nested patterns into readable Justin source code.
```
kriskowal: Make Justin famous with a link to justin.org or the nearest approximation thereof.

kriskowal (changes requested):
Please remove the `!` from the commit message and PR description. If this is going to cause practical consequences for Agoric CI, please add a commit to `endo-integration-master` to reconcile or relax the overconstrained tests instead. You will likely need to verify this one with an endo-branch pinned in an Agoric SDK PR to validate that, but I prefer that over the eval twins we’ll inevitably have to iron out if we ramp the major version for error message changes, which I don’t consider part of our contract.

Also, please take care to remove the checklist from the end of the PR description so it doesn’t get preserved in the merge commit message.

erights:
> for golden tests, I find ava snapshots pretty handy

Thanks. I will keep that in mind next time I'm adding or maintaining tests with golden tests. Thanks!

But it doesn't sound like you're requesting that shift in this PR. I agree that it isn't needed in this PR, so not doing it.

packages/patterns/NEWS.md
erights: Wow, I cannot find an *explanation* of the Justin language anywhere. Anyone know where to look? Attn @dckc @michaelfig

erights: The best I could find is https://github.com/agoric-labs/jessica/blob/master/lib/quasi-justin.js.ts , which is unsatisfying, but will do for now unless someone has a better suggestion. Resolving.

michaelfig: There's a better (more maintained, and evolving) reference: https://github.com/endojs/Jessie/blob/main/packages/parse/src/quasi-justin.js

erights: Done, thanks!

erights:
> If this is going to cause practical consequences for Agoric CI, please add a commit to endo-integration-master

In accord with our conversation, both https://github.com/Agoric/agoric-sdk/pull/7937 and are rebased on agoric-sdk master. The only difference between them is the `#endo-branch`. Both in CI now. I decided not to cherry pick anything, since we agreed that if they have the same CI errors and the errors are unrelated to this PR, then we're clear. Otherwise, I'll create an agoric-sdk PR that fixes the relevant differences revealed by that experiment.

Watching.

kriskowal (approved):

PR #2870, opened 2025-06-26
refactor(base64): minor improvements

erights:
Closes: #XXXX
Refs: #2844 

## Description

https://github.com/endojs/endo/pull/2844 currently depends on `@endo/base64` for base64 encoding and decoding. @kriskowal asked that it be changed to use a hex encoding instead, for which changes to `@endo/base64` would be irrelevant. So I'm capturing in this PR the minor improvement I made to that package while working on #2844 .

### Security Considerations

exporting mutable data (`monodu64`) from a module is a security hazard. It is not exported from the package, so this is only a hazard, not a vulnerability. In this case there was zero reason not to freeze `monodu64`, so this PR does, leaving us one less security hazard to worry about.

### Scaling Considerations

none

none
### Testing Considerations

Aside from `monodu64`, the other changes to this PR is to expand the test in ways I noticed while working on #2844 .

none

### Upgrade Considerations

none

### DIFF
--- a/packages/base64/src/common.js
+++ b/packages/base64/src/common.js
@@ -1,5 +1,7 @@
 // @ts-check
 
+const { freeze } = Object;
+
 export const padding = '=';
 
 export const alphabet64 =
@@ -17,3 +19,4 @@ for (let i = 0; i < alphabet64.length; i += 1) {
   const c = alphabet64[i];
   monodu64[c] = i;
 }
+freeze(monodu64);
--- a/packages/base64/test/main.test.js
+++ b/packages/base64/test/main.test.js
@@ -3,17 +3,22 @@ import { atob as origAtob, btoa as origBtoa } from './_capture-atob-btoa.js';
 import { encodeBase64, decodeBase64, atob, btoa } from '../index.js';
 
 /**
- * @param {string} string Only uses the low byte of each UTF16 code unit, which
- * is ok as long as it is used only for this purpose for a local test, and not
- * exported.
+ * `asciiString` must consist only of characters with code points between
+ * 0 and 255 (or `oxff`), i.e., characters whose code points fit into one byte.
+ * `asciiStringToUint8Array` converts that to a Uint8Array of the same size,
+ * consisting of those code points in order.
+ *
+ * @param {string} asciiString
  * @returns {Uint8Array}
  */
-const encode8BitString = string => {
-  const data = new Uint8Array(string.length);
-  for (let i = 0; i < string.length; i += 1) {
-    const byte = string.charCodeAt(i);
+const asciiStringToUint8Array = asciiString => {
+  const data = new Uint8Array(asciiString.length);
+  for (let i = 0; i < asciiString.length; i += 1) {
+    const byte = asciiString.charCodeAt(i);
     if (byte > 0xff) {
-      throw Error(`invalid character at index ${i}: U+${byte.toString(16).padStart(4, '0')}`);
+      throw Error(
+        `invalid character at index ${i}: U+${byte.toString(16).padStart(4, '0')}`,
+      );
     }
     data[i] = byte;
   }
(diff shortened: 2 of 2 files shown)


packages/base64/src/common.js
```diff
@@ -17,3 +19,4 @@ for (let i = 0; i < alphabet64.length; i += 1) {
   const c = alphabet64[i];
   monodu64[c] = i;
 }
+freeze(monodu64);
```
erights, line 22: I know it is not exported from the package. But it is exported from the module and its potential shared mutability is a hazard I'd like to avoid. @kriskowal , `@endo/base64` does depend on `'ses'`, so if you prefer, I can switch it to `harden`. Please let me know.

kriskowal: Harden is fine.

kriskowal (approved):

packages/base64/test/main.test.js
```diff
@@ -3,20 +3,28 @@ import { atob as origAtob, btoa as origBtoa } from './_capture-atob-btoa.js';
...
+ * exported.
  * @returns {Uint8Array}
  */
-function stringToBytes(string) {
+const stringToUint8Array = string => {
   const data = new Uint8Array(string.length);
   for (let i = 0; i < string.length; i += 1) {
     data[i] = string.charCodeAt(i);
   }
   return data;
```
gibson042: If we're renaming this anyway, we could do a better job of indicating behavior.
```suggestion
const encode8BitString = string => {
  const data = new Uint8Array(string.length);
  for (let i = 0; i < string.length; i += 1) {
    const byte = string.charCodeAt(i);
    if (byte > 0xff) {
      throw Error(`invalid character at index ${i}: U+${byte.toString(16).padStart(4, '0')}`);
    }
    data[i] = byte;
  }
  return data;
```

packages/base64/test/main.test.js
```diff
@@ -3,20 +3,28 @@ import { atob as origAtob, btoa as origBtoa } from './_capture-atob-btoa.js';
...
-}
+/**
+ * @param {Uint8Array} data
+ * @returns {string} Interpreting each 8-bit value as an 8-bit UTF-16 code
+ * unit. Since this cannot include any UTF-16 surrogates, this is equivalent
+ * to interpreting each 8-bit value as an 8-bit ascii code point. This
+ * may be unexpected, and so is ok as long as it is used only for this purpose
+ * for a local test, and not exported.
+ */
+const unit8ArrayToString = data => String.fromCharCode(...data);
```
gibson042: There's a typo here ("unit8" vs. "uint8"). But aligning with the previous suggestion:
```suggestion
const decodeTo8BitString = data => String.fromCharCode(...data);
```

kriskowal: At the risk of circling the bike shed, if we switch to `encode` and `decode`, with the precedent of `decodeBase64` and `encodeBase64` right here, I would suggest `decodeBytes`. I suggest bytes over `8Bit` or `Uint8s`. I suggest omitting `String` because it’s omitted in `decodeBase64`. Although `Ascii` and `Eascii` are both valid in their own right, I assert they’re not better than `Bytes` when standing next to `Base64` or `Base16` albeit `Hex`.

packages/base64/src/common.js
erights: $ yarn lint
src/common.js:20:1 - error TS2304: Cannot find name 'harden'.

20 harden(monodu64);
   ~~~~~~


Found 1 error in src/common.js:20

erights: I was wrong. @endo/base64 does not depend on `'ses'`. I'll stick with `freeze`. Resolving.

packages/base64/test/main.test.js
erights: Typo done. Still contemplating naming. Leaving unresolved until that's done. Fortunately, the names are only non-exported and within a test file, so ok if not great. But still...

erights: Went with `asciiStringToUint8Array` and `unt8ArrayToAsciiString`. Done

PR #2871, opened 2025-06-27
fix(marshal): marshal ByteArrays to hex

erights:
Closes: #XXXX
Refs: #XXXX

## Description

Provides for serialization and unserialization of ByteArrays (passable frozen Immutable ArrayBuffers) to our various textual encodings (capData, smallcaps, justin).

It does not yet add any encoding for ByteArrays to `encodePassable`. Attn @gibson042 

Alternative to https://github.com/endojs/endo/pull/2844 using hex instead of base64

### Security Considerations

Assuming this PR is correct, none

### Scaling Considerations

Hex is a terribly inefficient encoding of binary, but that's the price of using a textual format.

Further, with https://github.com/tc39/proposal-arraybuffer-base64 not yet being part of the platform, Doing this conversion in JS vs native is horribly inefficient, especially on XS.

Hopefully, most of the time when we actually want to marshal large ByteArrays, we'll have switched to a binary format that needs no translation.

Just an extension of the passable types that marshal should accept.

### Testing Considerations

Includes tests

If a system post this PR sends a ByteArray through marshal to a receiver that is pre this PR, the receiver will error since this is a case it does not know about.

### Upgrade Considerations

When storing a ByteArray to durable storage, you'll actually be storing this hex-based encoding. That durable storage can only be understood post this PR, which is a problem if there is a co-existence of code in one vat, some of which has bundled pre this PR and some post this PR.

### DIFF
--- a/packages/marshal/NEWS.md
+++ b/packages/marshal/NEWS.md
@@ -2,6 +2,7 @@ User-visible changes in `@endo/marshal`:
 
 # 1.8.0 (2025-07-11)
 
+- `@endo/pass-style` had already grown a new `Passable` data type, [`ByteArray`](https://github.com/endojs/endo/pull/2843), which is a frozen [Immutable ArrayBuffer](https://github.com/tc39/proposal-immutable-arraybuffer/). `@endo/marshal` can now serialize and deserialize these to the capData, smallcaps, and Justin formats using a textual hex encoding. The `encodePassable` format does not yet support ByteArrays.
 - Introduces an environment variable config option `ENDO_RANK_STRINGS` to change the rank ordering of strings from the current (incorrect) ordering by UTF-16 code unit used by JavaScript's `<` and `.sort()` operations to (correct and OCapN-conformant) ordering by Unicode code point. It currently defaults to "utf16-code-unit-order", matching the previously-unconditional behavior.
 
 # v1.7.0 (2025-06-02)
--- a/packages/marshal/src/encodeToCapData.js
+++ b/packages/marshal/src/encodeToCapData.js
@@ -15,6 +15,8 @@ import {
   assertPassableSymbol,
   nameForPassableSymbol,
   passableSymbolForName,
+  byteArrayToHex,
+  hexToByteArray,
 } from '@endo/pass-style';
 import { X, Fail, q } from '@endo/errors';
 
@@ -195,8 +197,10 @@ export const makeEncodeToCapData = (encodeOptions = {}) => {
         return passable.map(encodeToCapDataRecur);
       }
       case 'byteArray': {
-        // TODO implement
-        throw Fail`marsal of byteArray not yet implemented: ${passable}`;
+        return {
+          [QCLASS]: 'byteArray',
+          data: byteArrayToHex(passable),
+        };
       }
       case 'tagged': {
         return {
@@ -368,6 +372,10 @@ export const makeDecodeFromCapData = (decodeOptions = {}) => {
           const { tag, payload } = jsonEncoded;
           return makeTagged(tag, decodeFromCapData(payload));
         }
+        case 'byteArray': {
+          const { data } = jsonEncoded;
+          return hexToByteArray(data);
+        }
         case 'slot': {
           // See note above about how the current encoding cannot reliably
           // distinguish which we should call, so in the non-default case
@@ -442,3 +450,4 @@ export const makeDecodeFromCapData = (decodeOptions = {}) => {
   };
   return harden(decodeFromCapData);
 };
+harden(makeDecodeFromCapData);
--- a/packages/marshal/src/rankOrder.js
+++ b/packages/marshal/src/rankOrder.js
@@ -359,11 +359,27 @@ export const { comparator: compareRank, antiComparator: compareAntiRank } =
   makeComparatorKit();
 
 /**
+ * Like `compareRank` and `compareAntiRank` and unlike `fullCompare`,
+ * `compareRankRemotablesTied` and `compareAntiRankRemotablesTied`
+ * considers all remotables tied for the same rank.
+ * Unlike `compareRank` and `compareAntiRank`,
+ * `compareRankRemotablesTied` and `compareAntiRankRemotablesTied`
+ * do not short circuit on encounting remotables.
+ */
+export const {
+  comparator: compareRankRemotablesTied,
+  antiComparator: compareAntiRankRemotablesTied,
+} = makeComparatorKit((_x, _y) => 0);
+
+/**
  * @param {Passable[]} passables
- * @param {RankCompare} compare
+ * @param {RankCompare} [compare]
  * @returns {boolean}
  */
-export const isRankSorted = (passables, compare) => {
+export const isRankSorted = (
+  passables,
+  compare = compareRankRemotablesTied,
+) => {
   const subMemoOfSorted = memoOfSorted.get(compare);
   assert(subMemoOfSorted !== undefined);
   if (subMemoOfSorted.has(passables)) {
@@ -382,9 +398,9 @@ harden(isRankSorted);
 
 /**
  * @param {Passable[]} sorted
- * @param {RankCompare} compare
+ * @param {RankCompare} [compare]
  */
-export const assertRankSorted = (sorted, compare) =>
+export const assertRankSorted = (sorted, compare = compareRankRemotablesTied) =>
   isRankSorted(sorted, compare) ||
   // TODO assert on bug could lead to infinite recursion. Fix.
   // eslint-disable-next-line no-use-before-define
@@ -394,10 +410,10 @@ harden(assertRankSorted);
 /**
  * @template {Passable} T
  * @param {Iterable<T>} passables
- * @param {RankCompare} compare
+ * @param {RankCompare} [compare]
  * @returns {T[]}
  */
-export const sortByRank = (passables, compare) => {
+export const sortByRank = (passables, compare = compareRankRemotablesTied) => {
   /** @type {T[]} mutable for in-place sorting, but with hardened elements */
   let unsorted;
   if (Array.isArray(passables)) {
@@ -438,12 +454,17 @@ harden(sortByRank);
  * https://en.wikipedia.org/wiki/Binary_search_algorithm#Procedure_for_finding_the_leftmost_element
  *
  * @param {Passable[]} sorted
- * @param {RankCompare} compare
  * @param {Passable} key
- * @param {("leftMost" | "rightMost")=} bias
+ * @param {RankCompare} [compare]
+ * @param {"leftMost" | "rightMost"} [bias]
  * @returns {number}
  */
-const rankSearch = (sorted, compare, key, bias = 'leftMost') => {
+const rankSearch = (
+  sorted,
+  key,
+  compare = compareRankRemotablesTied,
+  bias = 'leftMost',
+) => {
   assertRankSorted(sorted, compare);
   let left = 0;
   let right = sorted.length;
@@ -460,16 +481,26 @@ const rankSearch = (sorted, compare, key, bias = 'leftMost') => {
   return bias === 'leftMost' ? left : right - 1;
 };
 
+// TODO https://github.com/endojs/endo/issues/2883#issuecomment-3063809592
+// Some (all?) rank cover operations take a compare parameter.
+// These should all now be optional, moved to the end, and default to
+// `compareRankRemotablesTied`.
+
 /**
  * @param {Passable[]} sorted
- * @param {RankCompare} compare
  * @param {RankCover} rankCover
+ * @param {RankCompare} [compare] which rank comparison function to use.
+ * Default to `compareRankRemotablesTied`.
  * @returns {IndexCover}
  */
-export const getIndexCover = (sorted, compare, [leftKey, rightKey]) => {
+export const getIndexCover = (
+  sorted,
+  [leftKey, rightKey],
+  compare = compareRankRemotablesTied,
+) => {
   assertRankSorted(sorted, compare);
-  const leftIndex = rankSearch(sorted, compare, leftKey, 'leftMost');
-  const rightIndex = rankSearch(sorted, compare, rightKey, 'rightMost');
+  const leftIndex = rankSearch(sorted, leftKey, compare, 'leftMost');
+  const rightIndex = rankSearch(sorted, rightKey, compare, 'rightMost');
   return [leftIndex, rightIndex];
 };
 harden(getIndexCover);
--- a/packages/marshal/test/encodePassable.test.js
+++ b/packages/marshal/test/encodePassable.test.js
@@ -451,12 +451,14 @@ const testOrderInvariants = test.macro(async (t, pickEncode) => {
   // Ensure at least one ava assertion.
   t.pass();
 });
-test(
+// Failing because we now need to encode byteArrays
+test.failing(
   'original passable encoding corresponds to rankOrder',
   testOrderInvariants,
   pickLegacy,
 );
-test(
+// Failing because we now need to encode byteArrays
+test.failing(
   'small passable encoding corresponds to rankOrder',
   testOrderInvariants,
   pickCompact,
--- a/packages/marshal/test/rankOrder.test.js
+++ b/packages/marshal/test/rankOrder.test.js
@@ -15,6 +15,7 @@ import {
   getPassStyleCover,
   getIndexCover,
   assertRankSorted,
+  compareRankRemotablesTied,
 } from '../src/rankOrder.js';
 import { unsortedSample, sortedSample } from '../tools/marshal-test-data.js';
 
@@ -28,6 +29,8 @@ test('compareRank is reflexive', async t => {
   );
 });
 
+// TODO https://github.com/endojs/endo/issues/2883
+// Should these tests use `compareRank` or `compareRankRemotablesTied`?
 test('compareRank totally orders ranks', async t => {
   await fc.assert(
     fc.property(arbPassable, arbPassable, (a, b) => {
@@ -45,6 +48,8 @@ test('compareRank totally orders ranks', async t => {
   );
 });
 
+// TODO https://github.com/endojs/endo/issues/2883
+// Should these tests use `compareRank` or `compareRankRemotablesTied`?
 test('compareRank is transitive', async t => {
   await fc.assert(
     fc.property(
@@ -105,11 +110,11 @@ test('compareRank is transitive', async t => {
 });
 
 test('compare and sort by rank', t => {
-  assertRankSorted(sortedSample, compareRank);
-  t.false(isRankSorted(unsortedSample, compareRank));
-  const sorted = sortByRank(unsortedSample, compareRank);
+  assertRankSorted(sortedSample);
+  t.false(isRankSorted(unsortedSample));
+  const sorted = sortByRank(unsortedSample);
   t.is(
-    compareRank(sorted, sortedSample),
+    compareRankRemotablesTied(sorted, sortedSample),
     0,
     `Not sorted as expected: ${q(sorted)}`,
   );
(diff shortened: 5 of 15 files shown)


packages/marshal/src/encodeToCapData.js
```diff
@@ -442,3 +450,4 @@ export const makeDecodeFromCapData = (decodeOptions = {}) => {
   };
   return harden(decodeFromCapData);
 };
+harden(makeDecodeFromCapData);
```
erights, line 453: drive by, for several of these

packages/marshal/src/rankOrder.js
```diff
@@ -33,8 +33,8 @@ const { entries, fromEntries, setPrototypeOf, is } = Object;
 /**
  * This is the equality comparison used by JavaScript's Map and Set
  * abstractions, where NaN is the same as NaN and -0 is the same as
- * 0. Marshal serializes -0 as zero, so the semantics of our distributed
- * object system does not distinguish 0 from -0.
+ * 0. Marshal still serializes -0 as zero, so the semantics of our distributed
+ * object system does not yet distinguish 0 from -0.
```
erights: drive by

packages/marshal/src/rankOrder.js
```diff
@@ -74,6 +74,7 @@ export const compareNumerics = (left, right) => {
   assert(NumberIsNaN(left));
   return 1;
 };
+harden(compareNumerics);
```
erights, line 77: drive by

packages/marshal/src/rankOrder.js
```diff
@@ -292,19 +293,41 @@ export const makeComparatorKit = (compareRemotables = (_x, _y) => NaN) => {
...
 /**
  * @param {RankCompare} comparator
  * @returns {RankCompare=}
  */
 export const comparatorMirrorImage = comparator =>
   comparatorMirrorImages.get(comparator);
+harden(comparatorMirrorImage);
+
+export const { comparator: compareRank, antiComparator: compareAntiRank } =
+  makeComparatorKit();
```
erights: Many drive by's to this file. Could be broken off into a separate PR if reviewers request.

packages/marshal/test/encodePassable.test.js
```diff
@@ -451,12 +451,14 @@ const testOrderInvariants = test.macro(async (t, pickEncode) => {
...
-test(
+// Failing because we now need to encode byteArrays
+test.failing(
   'original passable encoding corresponds to rankOrder',
   testOrderInvariants,
   pickLegacy,
 );
-test(
+// Failing because we now need to encode byteArrays
+test.failing(
```
erights, line 461: Attn @gibson042

packages/marshal/NEWS.md
```diff
@@ -2,6 +2,7 @@ User-visible changes in `@endo/marshal`:
 
 # Next release
 
+- `@endo/pass-style` had already grown a new `Passable` data type, [`ByteArray`](https://github.com/endojs/endo/pull/2843), which is essentially a frozen [Immutable ArrayBuffer](https://github.com/tc39/proposal-immutable-arraybuffer/). `@endo/marshal` can now serialize and deserialize these to the capData, smallcaps, and Justin formats using a textual hex encoding. The `encodePassable` format does not yet support ByteArrays.
```
kriskowal: Does this explanation less accurate without the word “essentially”?
```suggestion
- `@endo/pass-style` had already grown a new `Passable` data type, [`ByteArray`](https://github.com/endojs/endo/pull/2843), which is a frozen [Immutable ArrayBuffer](https://github.com/tc39/proposal-immutable-arraybuffer/). `@endo/marshal` can now serialize and deserialize these to the capData, smallcaps, and Justin formats using a textual hex encoding. The `encodePassable` format does not yet support ByteArrays.
```

packages/marshal/src/rankOrder.js
kriskowal: As long as the commit is separate, we won’t lose them if we have to revert a commit later.

packages/marshal/src/rankOrder.js
```diff
@@ -455,14 +483,20 @@ const rankSearch = (sorted, compare, key, bias = 'leftMost') => {
 
 /**
  * @param {Passable[]} sorted
- * @param {RankCompare} compare
  * @param {RankCover} rankCover
+ * @param {RankCompare} [compare] TODO Should also move `compare` parameter to
```
kriskowal: TODO will appear in documentation. Consider moving it out. Consider filing an issue and embedding a link.

packages/marshal/test/rankOrder.test.js
```diff
@@ -28,6 +29,7 @@ test('compareRank is reflexive', async t => {
   );
 });
 
+// TODO Should these tests use `compareRank` or `compareRankRemotablesTied`?
```
kriskowal: Is it our position that `compareRankRemotablesTied` is closer to the spirit intended by the name `compareRank` and that we’re using a distinguished name to avoid a breaking change? My understanding of rank comparison is that equivalent values in all ranks are tied, and that should have carried for remotables. I assume that, had we gotten this right the first time, “compare” would be a total order and “compareRank” would be a partial order for all applicable ranks. Am I right that there is a hint here that the in-band representation for promises and targets should not have the slot number, but imply the slot number by order of appearance in OCapN?

packages/pass-style/src/byteArray.js
```diff
@@ -71,3 +72,150 @@ export const ByteArrayHelper = harden({
...
+ * https://github.com/tc39/proposal-arraybuffer-base64/blob/main/playground/polyfill-core.mjs
+ *
+ * @param {Uint8Array} data
+ * @returns {string} hex encoded string
+ */
+export const uint8ArrayToHex = data => {
+  const len = data.length;
+  const out = [];
+  for (let i = 0; i < len; i += 1) {
+    out.push(data[i].toString(16).padStart(2, '0'));
```
kriskowal: It would not be awful to precompute or even embed a 256 entry table.

kriskowal (approved):

packages/marshal/src/rankOrder.js
```diff
@@ -344,19 +345,41 @@ export const makeComparatorKit = (compareRemotables = (_x, _y) => NaN) => {
...
+export const { comparator: compareRank, antiComparator: compareAntiRank } =
+  makeComparatorKit();
+
+/**
+ * Like `compareRank` and `compareAntiRank` and unlike fullCompare,
+ * `compareRankRemotablesTied` and `compareAntiRankRemotablesTied`
+ * considers all remotables tied for the same rank.
+ * Unlike `compareRank` and `compareAntiRank`
+ * `compareRankRemotablesTied` and `compareAntiRankRemotablesTied`
+ * do not short circuit on encounting remotables.
```
gibson042:
```suggestion
 * Unlike `compareRank` and `compareAntiRank`,
 * `compareRankRemotablesTied` and `compareAntiRankRemotablesTied`
 * do not short circuit on encounting remotables.
```

packages/marshal/test/encodePassable.test.js
gibson042: Hmm, I'm of the opinion that this should not land without reënabling these tests, either by supporting byteArrays in `encodePassable` or by filtering them out of the inputs here.

packages/pass-style/src/byteArray.js
```diff
@@ -71,3 +72,150 @@ export const ByteArrayHelper = harden({
...
+// /////////////////////////////////////////////////////////////////////////////
+
+/**
+ * Returns a Uint8Array reflecting the current contents of `byteArray`.
+ *
+ * @param {ArrayBuffer} byteArray i.e., a hardened Immutable ArrayBuffer whose
+ * `passStyle` is `'byteArray'`. NOTE: `encodeByteArrayToBase64` does not
+ * enforce that `byteArray` is a ByteArray, but rather, will also accept a
+ * normal `ArrayBuffer`. In that case, the returned `Uint8Array` would
+ * not be freezable, and would be a read-write view onto `byteArray`.
```
gibson042: I see no such function as `encodeByteArrayToBase64`, but I think it's intended as a self-reference.
```suggestion
 * @param {ArrayBuffer} byteArray i.e., a hardened Immutable ArrayBuffer whose
 * `passStyle` is `'byteArray'`. NOTE: A normal (mutable) ArrayBuffer is also
 * accepted, in which case the returned `Uint8Array` is a non-freezeable
 * read-write view of `byteArray`.
```

packages/pass-style/src/byteArray.js
gibson042: This should probably just be copied from agoric-sdk: https://github.com/Agoric/agoric-sdk/blob/master/packages/internal/src/hex.js

packages/pass-style/src/byteArray.js
```diff
@@ -71,3 +72,150 @@ export const ByteArrayHelper = harden({
...
+    u8[i / 2] = parseInt(hexits, 16);
+  }
+  // By special dispensation, we can harden Uint8Arrays even though we cannot
+  // (yet) lock down its indexed properties.
+  return harden(u8);
+};
+harden(hexToUint8Array);
+
+/**
+ * Like `uint8ArrayToHex` but starting from a presumably a frozen
```
gibson042:
```suggestion
 * Like `uint8ArrayToHex` but starting from a presumably frozen
```

packages/marshal/NEWS.md
erights: > Does this explanation less accurate without the word “essentially”?

Yeah, a bit. But the "essentially" without further words doesn't really help. The missing criteria is that the Immutable ArrayBuffer must also not have any extraneous own properties, and (shim aside) must inherit directly from `ArrayBuffer.prototype`.

I don't want to say all this here though, as it is the wrong place. Leaving this unresolved until this is fixed.

packages/marshal/test/rankOrder.test.js
erights: > Is it our position that `compareRankRemotablesTied` is closer to the spirit intended by the name `compareRank` and that we’re using a distinguished name to avoid a breaking change? 

Yeah. In fact, `compareRankRemotablesTied` restores the default semantics we had before https://github.com/endojs/endo/pull/2597 . I remember looking at #2597 at the time and thinking it was ok for the short circuiting behavior to be the new default. But in answer to your question, while I agree we need the short circuiting option, I no longer think it should have been the default. If we didn't have the compat worry, I would restore the `compareRankRemotablesTied` as the default behavior and call the #2597 behavior something else. Attn @gibson042 

> My understanding of rank comparison is that equivalent values in all ranks are tied, and that should have carried for remotables.

Not sure what you mean by "equivalent" above. The `compareRankRemotablesTied` behavior, which is also the pre-#2597 default behavior, is simply to consider all remotables to have the same rank, even though they are not equivalent. The #2597 default behavior does still consider all remotables to have the same rank. But it also short circuits further comparisons so that `[r1, 0]` and `[r2, "x"]` also have the same rank. This is coherent and needed, but I don't know any math name for such an ordering. 

> I assume that, had we gotten this right the first time, “compare” would be a total order and “compareRank” would be a partial order for all applicable ranks.

What "compare"?

All the rank comparisons (short-circuiting, `compareRankRemotablesTied`, and the "full order" comparators made by `makeFullOrderComparatorKit`) are all full orders among ranks, not partial orders. The "among ranks" qualifier is why we call these "rank orderings" (whose confusing mathematical name is "pre-order"). In a full order among ranks, for all x,y, compare(x,y) says either that 
- x's rank is below y's rank
- x and y have the same rank
- x's rank is above y's rank

In a partial order such as `keyCompare` we have the additional case
- x and y are incomparable.

For example, 
- compareRank(0, -0) === 0 // 0 and -0 have the same rank
- compareRank(3, NaN) === -1 // NaN is "above" all other numbers
- keyCompare(0, -0) === 0 // 0 and -0 have the same magnitude
- keyCompare(3, NaN) === NaN // The magnitudes of 3 and NaN are incomparable

> Am I right that there is a hint here that the in-band representation for promises and targets should not have the slot number, but imply the slot number by order of appearance in OCapN?

- For `encodePassable`, absolutely, because body ordering must reflect rank ordering.
- For others, we have up on body ordering. But it would still be nice for body equality to reflect same-rank. In that case, we'd still need this positional representation.

For both, we still need the body to have the same rep for 0 and -0 and to distinguish them outside the body, such as yet another slot entry.

packages/pass-style/src/byteArray.js
erights: Done

packages/marshal/NEWS.md
erights: I think omitting "essentially" here without further note is fine. This isn't a place where it is confusing to state the truth but not the full truth.

packages/marshal/test/rankOrder.test.js
erights: Given the compat constraints, I don't see a better option, so Resolving.

packages/marshal/src/rankOrder.js
erights: Done

packages/pass-style/src/byteArray.js
```diff
@@ -71,3 +73,82 @@ export const ByteArrayHelper = harden({
...
+export const uint8ArrayToByteArray = uint8Array =>
+  // If we're using a native implementation of the full Immutable ArrayBuffer
+  // proposal, and if `uint8Array`'s backing store is such a genuine
+  // Immutable ArrayBuffer, and if the native implementation does the obvious
+  // optimization of `immutableArrayBuffer.sliceToImmutable(0, length)`,
+  // then this conversion is zero-copy.
+  // Otherwise, it pays the copy cost implicit in `sliceToImmutable`.
+  //
+  // @ts-expect-error How can shim add to `ArrayBuffer` ts type?
+  harden(uint8Array.buffer.sliceToImmutable(0, uint8Array.length));
```
gibson042: I think this should account for non-zero buffer offsets (e.g., we expect the ArrayBuffer instance returned from `uint8ArrayToByteArray(new Uint8Array(Uint8Array.of(0, 1, 2, 3, 4, 5, 6).buffer, 1, 4))` to have data [0x01, 0x02, 0x03, 0x04]).
```suggestion
export const uint8ArrayToByteArray = uint8Array => {
  const { byteOffset, length } = uint8Array;
  // If we're using a native implementation of the full Immutable ArrayBuffer
  // proposal, and if `uint8Array`'s backing store is such a genuine
  // Immutable ArrayBuffer, and if the native implementation does the obvious
  // optimization of a non-expanding `sliceToImmutable`, then this conversion is
  // zero-copy. Otherwise, it pays the copy cost implicit in `sliceToImmutable`.
  //
  // @ts-expect-error How can the shim update the `ArrayBuffer` TS type?
  const byteArray = uint8Array.buffer.sliceToImmutable(
    byteOffset,
    byteOffset + length,
  );
  return harden(byteArray);
};
```

gibson042 (changes requested):

packages/marshal/src/rankOrder.js
erights: Done

packages/pass-style/src/byteArray.js
erights: Well, not so done. After the switch, tests that used to succeed are now failing. This might be because the old code was incorrect and the tests followed along, or that the new `hex.js` code is incorrect. I'll track this down.

erights: Unresolving

kriskowal: I would like to avoid entraining a dependency on the `BufferConstructor` type, if it can be helped.

packages/pass-style/src/hex.js
```diff
@@ -0,0 +1,206 @@
...
+harden(makeBufferishHexCodec);
+
+/**
+ * Export a hex codec that can work with standard JS engines, but takes
+ * advantage of optimizations on some platforms (like Node.js's Buffer API).
+ */
+export const { uint8ArrayToHex, encodeHex, hexToUint8Array, decodeHex } =
+  typeof Buffer === 'undefined'
+    ? makePortableHexCodec()
+    : makeBufferishHexCodec(Buffer);
```
gibson042: Should this file also anticipate [proposal-arraybuffer-base64](https://github.com/tc39/proposal-arraybuffer-base64), which is currently at Stage 3?

<details>

```suggestion
/**
 * @param {Pick<HexCodec, 'uint8ArrayToHex' | 'hexToUint8Array'>} hexCodec
 * @returns {HexCodec}
 */
const addDeprecatedFunctions = hexCodec => {
  const { uint8ArrayToHex, hexToUint8Array } = hexCodec;

  /**
   * @deprecated Use `uint8ArrayToHex` instead
   * @type {HexCodec['uint8ArrayToHex']}
   */
  const encodeHex = u8 => uint8ArrayToHex(u8);

  /**
   * @deprecated Use `hexToUint8Array` instead
   * @type {HexCodec['hexToUint8Array']}
   */
  const decodeHex = hex => hexToUint8Array(hex);
  
  return harden({
    uint8ArrayToHex,
    encodeHex,
    hexToUint8Array,
    decodeHex,
  });
};

/**
 * Create the most performant hex codec from the available global environment
 * (possibly leveraging platform-specific optimizations such as the Node.js
 * Buffer).
 *
 * @returns {Pick<HexCodec, 'uint8ArrayToHex' | 'hexToUint8Array'>}
 */
const makeBestHexCodec = () => {
  if (Uint8Array.fromHex) {
    return harden({
      uint8ArrayToHex: u8 => u8.toHex(),
      hexToUint8Array: hex => Uint8Array.fromHex(hex);
    });
  }
  return typeof Buffer === 'function'
    ? makeBufferishHexCodec(Buffer)
    : makePortableHexCodec();
};

export const { uint8ArrayToHex, encodeHex, hexToUint8Array, decodeHex } =
  addDeprecatedFunctions(makeBestHexCodec());
```

</details>

packages/pass-style/src/hex.js
```diff
@@ -0,0 +1,206 @@
...
+  const hexToUint8Array = hex => {
+    const buf = Bufferish.from(hex, 'hex');
+
+    // Coerce to Uint8Array to avoid leaking the abstraction.
+    const u8a = new Uint8Array(
+      buf.buffer,
+      buf.byteOffset,
+      buf.byteLength / Uint8Array.BYTES_PER_ELEMENT,
+    );
+    return u8a;
```
gibson042: I just noticed that the `makePortableHexCodec` `hexToUint8Array` hardens its output but the `makeBufferishHexCodec` analog does not, which is presumably an unintentional difference.

packages/pass-style/src/byteArray.js
erights: Even though the dependency is gated by a feature test?

packages/pass-style/src/hex.js
erights: Good catch. Yes, unintentional.

packages/marshal/src/rankOrder.js
```diff
@@ -358,12 +358,28 @@ harden(comparatorMirrorImage);
 export const { comparator: compareRank, antiComparator: compareAntiRank } =
   makeComparatorKit();
 
+/**
+ * Like `compareRank` and `compareAntiRank` and unlike `fullCompare`,
+ * `compareRankRemotablesTied` and `compareAntiRankRemotablesTied`
```
erights, line 363: This change not propagated to #3226

packages/marshal/tools/marshal-test-data.js
```diff
@@ -1,11 +1,19 @@
-import { makeTagged, passableSymbolForName } from '@endo/pass-style';
+import {
```
erights, line 1: Changes to this file not propagated to #3226

packages/pass-style/src/hex.js
```diff
@@ -0,0 +1,210 @@
+// /* global Buffer */
```
erights, line 1: Changes to this file not propagated to #3226, or at least, not obviously.

packages/pass-style/index.js
```diff
@@ -26,6 +26,15 @@ export {
...
 } from './src/string.js';
 
+export { uint8ArrayToHex, hexToUint8Array } from './src/hex.js';
+
+export {
+  byteArrayToUint8Array,
+  uint8ArrayToByteArray,
+  byteArrayToHex,
+  hexToByteArray,
+} from './src/byteArray.js';
```
erights, line 36: exports 6 whereas #3226 exports 4. Does #3226 drop two correctly?

erights:
Consider closing in favor of #3226 and https://github.com/endojs/endo-but-for-bots/pull/73 in anticipation of it being an endo PR.

PR #2872, opened 2025-06-30
fix(compartment-mapper): guarantee stable paths

boneskull:
## Description

Previously, it was trivially possible to generate a dependency tree which caused
`mapNodeModules` to return incorrect shortest values of
`CompartmentDescriptor.path`. An example such tree (which is used in test
fixtures):

```mermaid
graph TD
    app --> paperino
    app --> pippo
    paperino --> topolino
    pippo --> gambadilegno
    topolino --> gambadilegno
    gambadilegno --> goofy
```

In the above tree, there are two paths from `app` to `goofy`; `['app', 'pippo',
'gambadilegno', 'goofy']` and `['app', 'paperino', 'topolino', 'gambadilegno',
'goofy']`. The shortest path from `app` to `goofy` is the former, but
`mapNodeModules` would consistently return the `latter` due to its depth-first
nature and real-time calculation of optimal paths. In addition, it was possible
for larger dependency trees to be completely non-deterministic due to
asynchronous I/O operations.

The current strategy seemed too fragile & error-prone to fix _in situ_, so I
tore it out. The result is this change, which fixes `@endo/compartment-mapper`
so that `mapNodeModules` will return a `CompartmentMapDescriptor` with
**stable** and **correct shortest values** for `path`.

This was done by implementing a graph data structure which maps nodes (packages)
via edges (dependee -> dependency) using an edge weighting function based only
on the dependency name (which is analogous to order of comparisons done in the
`@endo/path-compare` algorithm [I think]).

> The new graph structure used is _not_ the same as the `Graph` type already
> used in `node-modules.js`; this is just a flat Record of package locations to
> data (and is arguably a misnomer).

When the filesystem crawl is complete—and prior to translation into a
`CompartmentMapDescriptor`—we calculate the shortest path from the entry to each
other member of the graph (excepting the attenuator node, if it exists), and use
this information to set the `path` value for the data structure entering
`translateGraph`.

### Security Considerations

n/a

### Scaling Considerations

This potentially has negative performance implications, but the tradeoff is the
resulting `CompartmentMapDescriptor` is now deterministic and correct (for some
definition of "correct").

n/a

### Testing Considerations

This adds some unit tests for the new graph data structure, the shortest-path
algorithm, and updates existing shortest-path tests to use the "problem"
dependency tree shown above.

Manual testing with a large dependency tree (1500+ compartments) shows that the
resulting `CompartmentMapDescriptor` is stable and correct (but has not been
exhaustively tested).

- `CompartmentMapDescriptor`s created via `mapNodeModules()` _prior_ to this
  change should be considered invalid for purposes of reproduction or comparison
  against the resulting `CompartmentMapDescriptor` now generated
  `mapNodeModules()`. This suggests to me that archives _could_ break if their
  data ever passes thru `mapNodeModules()` again; the resulting
  `CompartmentMapDescriptor` is likely to be different in all but the most
  simple cases. I don't know if this is a use-case; if it isn't, then this is a
  non-issue.

- Since the `path` value may change, the _canonical names_ (as used in policy)
  will also change, and so any static policies generated from the resulting
  `CompartmentMapDescriptor`s _prior_ to this change will be invalid against the
  new `CompartmentMapDescriptor`s. I don't know if this is a use-case for
  anything other than `@lavamoat/node` which I do not care about breaking due to
  it being alpha software.

### Upgrade Considerations

Given the [compatibility considerations](#compatibility-considerations) above,
it may be considered a breaking change. If so, I can change the commit message
as appropriate. `NEWS.md` should also be updated if so. Otherwise, this is just
a bugfix and maybe not worth a `NEWS.md` entry unless we can show significant
performance degredation.


packages/compartment-mapper/src/logical-path.js
```diff
@@ -0,0 +1,343 @@
...
+  assert(
+    nodeList.length >= 2,
+    `The path from ${source} to ${target} should have a least two nodes`,
+  );
+
+  return /** @type {[T, T, ...T[]]} */ (nodeList);
+};
+
+/**
+ * @template [T=string] The type of nodes in the graph
```
boneskull: The generic is not strictly necessary because we do not use anything but the default. It enables reuse of this data structure for other purposes, though. 🤷

packages/compartment-mapper/test/generic-graph.test.js
```diff
@@ -0,0 +1,168 @@
...
+const calculatePackageWeight = packageName => {
+  let totalCodeValue = packageName.length * 65536; // each character contributes 65536
+  for (let i = 0; i < packageName.length; i += 1) {
+    totalCodeValue += packageName.charCodeAt(i);
+  }
+  return totalCodeValue;
+};
+
+/**
+ * Creates a graph with the following structure:
```
kriskowal:
```suggestion
 * Creates a graph from a tree with the following structure:
```

packages/compartment-mapper/src/generic-graph.js
```diff
@@ -0,0 +1,351 @@
...
+ */
+const getPath = ({ predecessors }, source, target) => {
+  /** @type {T[]} */
+  const nodeList = [];
+
+  let node = target;
+
+  while (predecessors.has(node)) {
+    const currentNode = /** @type {T} */ (predecessors.get(node));
+    nodeList.unshift(node);
```
kriskowal: Unshift is O(length) on Array and can be avoided with push() if you can tolerate an in-place reverse() at the end.

boneskull: Ooh, OK. I was wondering about that.

packages/compartment-mapper/test/project-fixture.js
```diff
@@ -0,0 +1,301 @@
...
+ * @template [Context=unknown]
+ * @param {{log: LogFn}|ExecutionContext<Context>} logger
+ * @param {CompartmentMapDescriptor} compartmentMap
+ * @returns {void}
+ */
+export const dumpCompartmentMap = (logger, compartmentMap) => {
+  const compartmentMapForInspect = {
+    ...compartmentMap,
+    /** @type {CustomInspectFunction} */
+    [inspect.custom]: () => ({
```
boneskull, line 239: I'd consider embedding this stuff into the "real" data structure (or even creating a `CompartmentDescriptor` class) but since this is a node-ism we'd need to wrap it in a power. Not worth it.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -339,6 +319,37 @@ const inferParsers = (descriptor, location, languageOptions) => {
...
+ * @param {string} packageName - Name of package to calculate weight for.
+ * @returns {number} Numeric weight
+ */
+const calculatePackageWeight = packageName => {
+  let totalCodeValue = packageName.length * 65536; // each character contributes 65536
+  for (let i = 0; i < packageName.length; i += 1) {
+    totalCodeValue += packageName.charCodeAt(i);
+  }
+  return totalCodeValue;
+};
```
gibson042, line 351: To provide much-belated commentary, this function is not correct—the value of each successive code unit in two strings can differ by up to 0xFFFF, which can overcome a single-character length difference in just two characters (_e.g., the weight of \<FF8F FF8F> "ﾏﾏ" is 2 × 65536 + 0xFF8F + 0xFF8F = 261_918, which is **greater** than the length of the **longer** \<0041 0041 0041> "AAA" at 3 × 65536 + 0x41 + 0x41 + 0x41 = 196_803_). Further, ignoring the position of each code unit means that anagrams are incorrectly given equal weight (_e.g., both "endo" and "node" are weighted at 4 × 65536 + 0x64 + 0x65 + 0x6E + 0x6F = 262_566_), potentially thwarting the stability sought by this PR.

Honestly, this approach seems like an abuse of graph edge costs ("weights")... we want e.g. "foo" → "A" → "bar" to be preferred over "foo" → "B" → "bar", although it's less clear to me that "foo" → "B" → "C" → "bar" should be preferred over "foo" → "ABC" → "bar". But even if that is fine, achieving it correctly seems to require non-scalar definitions for both weights and distances, in which case you could weight "A" as [1, 65], "B", as [1, 66], "C" as [1, 67], "ABC" as [3, 65, 66, 67], and "B" → "C" as [1, 66, 1, 67] (_comparing lexicographically and defining the weight of a path by concatenation_), but having gone that far it seems reasonable to just define costs using strings directly and operate on a corresponding **non**-generic Graph.
```patch
 const dijkstra = (graph, source) => {
   const { nodes } = graph;
   /** @type {TraversalContext<T>} */
   const context = {
-    distances: new Map(),
+    paths: new Map(),
     predecessors: new Map(),
     queue: new Set(),
   };
-  const { queue, distances } = context;
+  const { queue, paths } = context;
 
-  for (const node of nodes) {
-    distances.set(node, Infinity);
-  }
-
-  assert(
-    distances.get(source) === Infinity,
-    `Source ${q(source)} is not in the graph`,
-  );
-
-  distances.set(source, 0);
-
   for (const node of nodes) {
     queue.add(node);
   }
+  assert(queue.has(source), `Source ${q(source)} is not in the graph`);
+  paths.set(source, []);
 
   while (queue.size !== 0) {
     const node = extractMin(context);
     if (node === undefined) {
       return context;
     }
     const adjacent = graph.adjacent(node);
     if (adjacent) {
       for (const edge of adjacent) {
         relax(graph, context, node, edge);
       }
     }
   }
   return context;
 };
```
```patch
 /**
- * Remove the node with the minimum weight from the priority queue.
+ * Remove the lowest-cost node from the priority queue.
  *
  * Performs linear search.
  * @template [T=string]
  * @param {TraversalContext<T>} tracks
  * @returns {T|undefined}
  */
-const extractMin = ({ distances, queue }) => {
-  let min = Infinity;
+const extractMin = ({ paths, queue }) => {
+  /** @type {T[]|undefined} */
+  let minPath = undefined;

   /** @type {T|undefined} */
   let minNode;
 
   queue.forEach(node => {
-    const nodeWeight = distances.get(node) ?? Infinity;
+    const path = paths.get(node);
+    if (!path) return;
 
-    if (nodeWeight < min) {
+    if (!minPath || isLowerCost(path, minPath)) {
-      min = nodeWeight;
+      minPath = path;
       minNode = node;
     }
   });
```
```patch
 /**
- * Update context with the new distance to the target node if the distance
- * through the source node is shorter than the current distance.
+ * Update context to include the current lowest-cost path to a target node
+ * reachable by a single edge from a source node.
  *
  * @template [T=string]
  * @param {GenericGraph<T>} graph
  * @param {TraversalContext<NoInfer<T>>} context
  * @param {NoInfer<T>} source
  * @param {NoInfer<T>} target
  */
-const relax = (graph, { distances, predecessors }, source, target) => {
+const relax = (graph, { paths, predecessors }, source, target) => {
-  const number = graph.getEdgeWeight(source, target);
-
-  const distanceSource = distances.get(source);
-  const distanceTarget = distances.get(target);
-
-  assert(
-    distanceSource !== undefined,
-    `Missing distance for source ${q(source)}`,
-  );
-  assert(
-    distanceTarget !== undefined,
-    `Missing distance for target ${q(target)} target`,
-  );
+  const pathSource = paths.get(source);
+  assert(pathSource, `Missing path to source ${q(source)}`);
+
+  const pathTarget = paths.get(target);
+  const newPath = [...pathSource, target];
 
-  if (distanceTarget > distanceSource + number) {
+  if (!pathTarget || isLowerCost(newPath, pathTarget)) {
-    distances.set(target, distanceSource + number);
+    paths.set(target, newPath);
     predecessors.set(target, source);
   }
 };
```
```js
/**
 * @template [T=string]
 * @param {T[]} a
 * @param {T[]} b
 */
const isLowerCost = (a, b) => pathCompare(a, b) < 0;
```

PR #2873, opened 2025-06-30
feat(marshal,patterns): env-option to config string order

erights:
closes: #2113 
refs: #2002  https://github.com/Agoric/agoric-sdk/pull/10299 https://github.com/Agoric/agoric-sdk/pull/10165/files#diff-b51392c299617fe43392dd0bcf11b9ed89b619cd4e5051fcffc9bfedef1c4d15 https://github.com/endojs/endo/pull/2008 #2875 

## Description

Based on https://github.com/endojs/endo/pull/2008 and alternative to it. Some comments on #2008 likely apply to this PR as well.

- [ ] In particular, many places that currently sort strings using the native JS comparison will need to instead delegate to `compareRank` so that they use whatever order `compareRank` is currently using.

Unlike #2008, this PR introduces an environment variable config option `ENDO_RANK_STRINGS`, defaulting off for now, to change the rank ordering of strings from the current (incorrect) ordering by UTF-16 code unit used by JavaScript's `<` and `.sort()` operations to (correct and OCapN conformant) ordering by Unicode code point. Thus, for now, when this default is not overridden, there is no observable change.

`@endo/patterns` provides a `compareKeys` partial order that delegates some ordering, including strings, to the rank ordering provided by `@endo/marshal`. So when the `ENDO_RANK_STRINGS` default is not overridden, then `compareKeys` also follows the (incorrect) UTF-16 code unit order. But when it is overridden, then `compareKeys` also follows the (correct) Unicode code-point order.

The considerations below are under the assumption that the default `ENDO_RANK_STRINGS` is not overridden, in which case this PR is a pure refactor. #2875 is staged on this PR and changes the default. See the documentation in that PR for the considerations of switching to unicode code point order. 

### Security Considerations

When `ENDO_RANK_STRINGS` is not overridden, none. Otherwise, as noted in #2875 :

> The fact that the string ordering is closer to the Unicode semantics of the strings probably minimizes some surprises in ways that help security. OTOH, this difference from JS native string ordering probably causes other surprises that hurt security. Altogether, we do not expect much effect.

### Scaling Considerations

When `ENDO_RANK_STRINGS` is not overridden, none. Otherwise, as noted in #2875 :

> As a comparison written in JS, will be slower that the JS native string comparison. On XS at least, we expect to have a native code point comparison function available eventually. Altogether, given either a native implementation or an aggressive jit, we do not expect much effect.

When `ENDO_RANK_STRINGS` is not overridden, none. Otherwise, as noted in #2875 :

> Most developers will not care. But it needs to be explained somewhere carefully so that developers that do care can easily find out. 

### Testing Considerations

This PR includes tests for each setting of `ENDO_RANK_STRINGS`. Each of those tests explicitly sets the option to the setting they test, and so are insensitive to the changing of default between this PR and #2875 .

When `ENDO_RANK_STRINGS` is not overridden, none. Otherwise, as noted in #2875 :

>  - These string ordering changes brings Endo into conformance with any string ordering components of the OCapN standard.
>  - To accommodate these change, you may need to adapt applications that relied on rank-order or key-order being the same as JS native order. You may need to re-sort any data that had previously been rank sorted using the prior `compareRank` function. You may need to revisit any use of patterns like `M.gte(string)` expressing inequalities over strings.
 
But these issues only occur when encountering strings that would sort differently under the two orderings. To help us test for that, this PR also includes an `'error-if-order-choice-matters'` setting for `ENDO_RANK_STRINGS` that might help us identify problems cases we did not detect by other means. Note that `'error-if-order-choice-matters'` is *only* for such testing. It is not a setting to be used in production.

### Upgrade Considerations

When `ENDO_RANK_STRINGS` is not overridden, none. Otherwise, as noted in #2875 :

> If we currently have any persistent data, especially on chain, sorted according to JS native order (by UTF-16 code unit), then we cannot accept this PR until we have a plan to re-sort that data, or somehow continue to live with mis-sorted old data. (Historical note: This is how Oracle came to permanently rely on UTF-16 code unit order, because of the impracticality of resorting all that data.)


packages/marshal/src/rankOrder.js
```diff
@@ -1,5 +1,6 @@
-import { getTag, passStyleOf, nameForPassableSymbol } from '@endo/pass-style';
+import { getEnvironmentOption as getenv } from '@endo/env-options';
```
kriskowal, line 1: It would be good to have a test of `@endo/env-options` inside a `Compartment`. I think that, in addition to endowing `process.env`, you have to also at least endow `assert`, which might be undesirable coupling, but fine in all our environments.

erights:
> I'd rather not rely on a link to something mutable for **security considerations**; I took the liberty of inlining them (in addition to the link).
> 
> I'd like to do the same for **upgrade considerations**, but it seems to imply that "we cannot accept this PR". Please inline the upgrade considerations, adjusting for the context of this PR.

Done. I put all of these quotes from the future with markdown quotes, which should be adequate adjustment.

packages/marshal/src/rankOrder.js
erights: Because of layering constraints, the tests of env-options within a `Compartment` is at https://github.com/endojs/endo/blob/master/packages/ses-ava/test/env-options.test.js . However, rather than test an import statement importing from `@endo/env-options` within a compartment, these test cases use `makeEnvironmentCaptor` to avoid having to set up a per compartment import environment.

We should indeed have such tests for compartment code importing from `@endo/env-options`. But this need not be in this PR. In the interests of unblocking you, I'll file an issue instead.

erights: https://github.com/endojs/endo/issues/2879

erights:
> To clarify why environment variables are a reasonable approach to module parameterization, I suggest adding a test to show it in action in the preferred way: the top level test script doesn't consult the unix environment; rather, it makes a Compartment with a global `process.env` with the flag set, and sorts some stuff inside that compartment.

In the interest of unblocking @kriskowal 's sync efforts, and in light of https://github.com/endojs/endo/pull/2873#discussion_r2191085939, I filed https://github.com/endojs/endo/issues/2879 to postpone this issue.

packages/marshal/src/rankOrder.js
erights: Resolving

PR #2877, opened 2025-07-03
chore: speed up linting

boneskull:
Previously, linting took this long (M4 mac):

```
yarn lint:workspaces:eslint  88.01s user 8.69s system 165% cpu 58.326 total
```

With this change:

```
yarn lint:eslint  18.74s user 2.13s system 166% cpu 12.545 total
```

## Description

This PR tweaks ESLint and TypeScript config to allow the entire monorepo to be linted from the monorepo root:

```shell
eslint .
```

Summary of changes:

- Add a root `.eslintignore` file, since it's kind of a lot to throw in `package.json`
- Configure the workspace root as the ESLint root, which prevents ESLint from attempting to traverse past it looking for configuration
- Move some small `.eslintignore` files into `package.json`-based ESLint config
- Add TS configuration files to workspaces missing them, since tseslint expects them to be there
- Removed redundant `allowJs: true` and `checkJs: true` in various TS config files (including `skel/tsconfig.build.json`; did not check _all_ of them)
- Removed unused `eslintIgnore` field from `lockdown/package.json`
- Migrated `eslintIgnore` field to `eslintConfig.ignorePatterns` in `init/package.json`
- Enable lint of the ESLint plugin 😄
- Shuffle scripts around; `yarn lint:eslint` now uses the new strategy; `yarn lint:workspaces:eslint` uses the old strategy (if we want to keep it)

---

The second commit applies lint fixes to previously-unlinted files.


packages/bundle-source/test/explicit.test.js
```diff
@@ -1,4 +1,3 @@
-import { resolve as pathResolve } from 'path';
```
kriskowal: The changes in this file obviate the file. I believe this is because `pathResolve` is not properly threaded through the `bundleSource` type. I would favor a ts-ignore over deleting. Or, at least, commenting out the offending lines with a note that we need to fix the type signature.

boneskull: Yah, I should have done that. my bad

boneskull:
@kriskowal 

> Does the story for linting a package in isolation change?

Each package can be linted in isolation.  This is what the `lint:workspaces:eslint` script does.

Similarly, any file or subdirectory can be linted via:

```sh
npx eslint <glob>
```

It'd make sense _to me_ to keep all common dev deps in the workspace root `package.json`, but IDK how well Yarn supports that strategy (so I didn't try to change it).  If it works the way it should, practically speaking, there won't be any change to how tools get run (because they all wind up in `/node_modules/.bin` anyhow).  It _does_ mean that a workspace cannot be

packages/bundle-source/test/explicit.test.js
boneskull: FWIW, removing `pathResolve` doesn't cause the test to fail, so the test is of dubious value anyhow...

kriskowal: I agree. It only verifies that injecting the power doesn’t break. A test for the presence of the feature would have to be framed in the negative: if you inject a malformed power, it should fail.

kriskowal (approved):
Let’s roll forward. Further improvements may follow.

boneskull:
@kriskowal FWIW, if we _do_ end up moving `devDependencies` into the workspace root, then any tool installed this way can always be executed from any workspace or subdirectory via `npx` or even `npm exec`.  There may be a `yarn` equivalent, but those two commands shouldn't conflict with Yarn in any way.

Issue #2876, opened 2025-07-02
compartment-mapper: revisit findRedirect (dynamic requires)


boneskull:
## What is the Problem Being Solved?

We currently support two use-cases where a compartment is allowed to use a dynamic require with an aboslute-path specifier to load a module in a different compartment:

1. The module lives in a compartment which the requesting compartment already has access to
2. The module lives in a compartment which has access to the requesting compartment

1. is pretty obvious, but the reason for 2. is to support the `node-gyp-build` use case:

    ```js
    const path = require('node:path');
    const myFoo = require('node-gyp-build')(path.join(__dirname, 'foo.node'));
    ```

In both of these cases, the compartment of the requested module already has a reference to `node-gyp-build` in its own compartment map.

**I've identified a third use-case** which will need addressing:

_The module lives in a compartment whose ancestor has access to the requesting compartment._

The use-case comes from Webpack. Typically, a user has a `webpack.config.js` file.  In this config file, you may declaratively specify _loaders_ to process certain filetypes.  Instead of actually importing these loaders in your config file, you provide just the _package name(s)_ in the Webpack configuration (similarly to how ESLint-pre-v9 config files work).

During execution, when Webpack encounters a file it needs to hand to a configured loader, it imports the loader (likely via dynamic require). Unlike `node-gyp-build`, `webpack` _itself_ does not perform the import; _another_ module, `loader-runner`, does.  This introduces a level of indirection that we do not currently account for.  Nowhere does `webpack.config.js`'s compartment know about the loaders, because they are not imported anywhere, even if they are direct dev dependencies of its package.

## Description of the Design

"The module lives in a compartment whose ancestor has access to the requesting compartment" is just my initial impression of what we need.  

I don't know of any way offhand to make this work _other_ than referencing package policy; I can certainly tell policy that `loader-runner` can access a module in, e.g., `babel-loader`.  

This may be the most feasible way forward.

Another strategy could be to leverage policy to make certain compartments "cascade" into the scopes of child compartments, though this sounds like both a morass (cycles!) and a slouchy security posture.

Other than that, I'm open to ideas because I am out of them.

## Security Considerations

IMO, this shouldn't be any worse than the currently-supported use-cases.

## Scaling Considerations

n/a

## Test Plan

Happy path / sad path, fixtures, etc.

## Compatibility Considerations

n/a

## Upgrade Considerations

If the implementation and security implications feel skeevy, then the behavior should be opt-in, and `NEWS.md` should contain an entry for API changes.

PR #2878, opened 2025-07-03
feat(compartment-mapper): allow loading of ancestor compartments via absolute path

boneskull:
Closes: #2876

## Description

- This adds `enforcePackagePolicyByPath()` which accepts two (2) `CompartmentDescriptor`s (_A_ and _B)_ and determines if _B_ is allowed access to _A_ via package policy.  `findRedirect()`, called by `importNowHook()`, now uses this if it cannot enforce policy any other way.  This solves the issue described in [#2876](https://github.com/boneskull/endo/issues/2876), which is that a package needs to dynamically require a module from an otherwise-unrelated-except-by-policy `CompartmentDescriptor`. This _cannot_ happen if policy is not defined.

    Slightly changed the existing error hint to differentiate this case.

- Tangential to this change, I refactored some policy-related things while I was debugging:
    - Adds a reusable function, `policyEnforcementFailureMessage()`, to generate helpful error messages when policy enforcement fails
    - Narrows `PolicyOption` type (to have field of type `SomePolicy` instead of `any`)
    - Loosens `FullAttenuationDefinition` type, wherein `params` is actually an optional field per the extant test suites
    - `policy-format.js`:
      - Moves `generateCanonicalName()` here
      - Updates `PackageNamingKit.name` for use in this context
      - Adds many more type guards and assertions for validation. Some of these are likely useful elsewhere in the codebase, but I have not exported any new type guards.
      - Adds `or()`, `and()` and `not()` which can be used to compose type guards in a type-safe way (e.g., `or([isString, isArray])`)
      - Moved `ATTENUATORS_COMPARTMENT` here to avoid cycle
    - Added some helpers to `typescript.ts`
    - Update snapshots to reflect slight changes to error messaging

- Updated the `mermaid-deptree` script to take non-production-dependencies into account

### Security Considerations

See #2876.  This opens up what can be dynamically required via absolute path by any given `Compartment`, but _only_ works if package policy is defined—unlike current policy enforcement.

### Scaling Considerations

n/a

n/a

### Testing Considerations

I've added a comprehensive test fixture and updated existing cases to assert behavior around allowed policy, disallowed policy, and omitted policy.

n/a

### Upgrade Considerations

This updates the `PolicyOption` external-facing type which has potential to affect consumers; it is more strict about the shape of the `policy` object provided to `policy`-accepting public APIs. Mitigating this is that runtime policy validation would throw an exception if the policy was invalid--so probably not that big of a deal.


boneskull:
Is this a feature or a bug fix?  I don't even know anymore

scripts/mermaid-deptree.cjs
```diff
@@ -42,7 +42,11 @@ async function buildDependencyGraph(entry, nodeModulesPath) {
         await fs.readFile(packageJsonPath, 'utf8'),
       );
       const packageName = packageJson.name || path.basename(directory);
-      const dependencies = Object.keys(packageJson.dependencies || {});
+      const dependencies = Object.keys({
+        ...packageJson.peerDependencies,
+        ...packageJson.devDependencies,
+        ...packageJson.dependencies,
```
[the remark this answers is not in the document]
boneskull: This is wrong.  🤦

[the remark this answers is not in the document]
kriskowal: Provably false, indeed.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -189,6 +190,21 @@ const findRedirect = ({
         };
       }
 
+      if (someCompartmentDescriptor.path && compartmentDescriptor.policy) {
```
naugtur: We should make sure that if policy is truthy and path is not we don't end up allowing something by default. 
I'd limit the if statement to just policy and react to path missing inside the if block (with a `continue` statement if that's appropriate, or with an error. Not sure which one that would be right now.

packages/compartment-mapper/src/types/policy.ts
```diff
@@ -6,12 +6,12 @@
 
 /* eslint-disable no-use-before-define */
 
-import type { SomeObject } from './typescript.js';
+import type { LiteralUnion, SomeObject } from './typescript.js';
```
naugtur, line 9: I've seen it work without the extension before. Not sure what it depends on for that, but can be done and maybe works by default by now.

boneskull: @naugtur I'm not sure what this is in reference to.

PR #2880, opened 2025-07-09
refactor: `hasOwn` rather than `hasOwnPropertyOf` or `objectHasOwnProperty`

erights:
Closes: #XXXX
Refs: #1627 

## Description

Replaces #1627 

Starting with Node 16.9, `Object.hasOwn` is supported and does what our hand rolled `hasOwnPropertyOf` or `objectHasOwnProperty` did. This PR changes all uses to use `hasOwn` instead. However, for the sake of compat, it only deprecates the old exports. It does not delete them.

Reviewers, `objectHasOwnProperty` is not exported from ses. Should I delete it instead of deprecating it?

### Security Considerations

none
### Scaling Considerations

none

one less thing to document
### Testing Considerations

none

none, as long as all platforms we still support already provide `Object.hasOwn`.

### Upgrade Considerations

none

### DIFF
--- a/packages/exo/src/exo-tools.js
+++ b/packages/exo/src/exo-tools.js
@@ -1,5 +1,5 @@
 import { getMethodNames } from '@endo/eventual-send/utils.js';
-import { hasOwnPropertyOf, toThrowable } from '@endo/pass-style';
+import { toThrowable } from '@endo/pass-style';
 import { E, Far } from '@endo/far';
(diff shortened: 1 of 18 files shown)


kriskowal:
Just noting that Object.hasOwn became ubiquitous on the web in 2022 https://caniuse.com/?search=Object.hasOwn

PR #2881, opened 2025-07-09
chore: Clean up after #2873

gibson042:
Refs: #2873

## Description

* Differentiate string rank order tests by name
* DRY out data and functions for string rank order tests
* Accommodate implementation-defined order of `sort(comparator)` comparisons
* Clarify in NEWS.md that ENDO_RANK_STRINGS is not limited to on/off

### Security Considerations

n/a

### Scaling Considerations

n/a

n/a

### Testing Considerations

See above

n/a

### Upgrade Considerations

n/a

### DIFF
--- a/packages/marshal/test/string-rank-both-order.test.js
+++ b/packages/marshal/test/string-rank-both-order.test.js
@@ -17,7 +17,7 @@ import { encodePassable } from './encodePassable-for-testing.js';
  */
 const sorted = (strings, comp) => [...strings].sort(comp);
 
-test('unicode code point order', t => {
+test('string ranking by code point/UTF-16 code unit agreement', t => {
   // Test case from
   // https://icu-project.org/docs/papers/utf16_code_point_order.html
   const str0 = '\u{ff61}';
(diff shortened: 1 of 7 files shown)


packages/marshal/test/string-rank-both-order.test.js
```diff
@@ -2,57 +2,34 @@ import '../tools/prepare-error-if-order-choice-matters.js';
...
- *
- * @param {string[]} strings
- * @param {(
- *   left: string,
- *   right: string
- * ) => import('../src/types.js').RankComparison} comp
- * @returns {string[]}
- */
-const sorted = (strings, comp) => [...strings].sort(comp);
+const { loneSurrogate$bmpHigh, bmpHigh, surrogatePair } = multiplanarStrings;
```
kriskowal, line 13: What’s the reasoning behind the `$` in this name?

kriskowal (approved):

packages/marshal/test/string-rank-both-order.test.js
kriskowal: Out of band, @gibson042 explained to me that this is effectively a lower precedence delimiter that indicates that the referent is a string comprised of lone surrogate and high BMP code points. I won’t take this as a precedent for naming in general, but a one-off convention that works for this particular case.

kriskowal: Or: read `$` as concat.

gibson042: It separates descriptions of consecutive code points, and unlike `_` (the only other ASCII option) does not trigger linter complaints about violating camelCase conventions.

PR #2888, opened 2025-07-12
chore: Release 2025-07-11

kriskowal:
# `@endo/immutable-arraybuffer` 1.1.2

- Removes `@endo/immutable-arraybufer/shim-hermes.js` and absorbs the necessary features into `@endo/immutable-arraybuffer/shim.js`. We are not qualifying this as a breaking change since the feature did not exist long enough to become relied upon.

# `@endo/marshal` 1.8.0

- Introduces an environment variable config option `ENDO_RANK_STRINGS` to change the rank ordering of strings from the current (incorrect) ordering by UTF-16 code unit used by JavaScript's `<` and `.sort()` operations to (correct and OCapN-conformant) ordering by Unicode code point. It currently defaults to "utf16-code-unit-order", matching the previously-unconditional behavior.

# `@endo/pass-style` 1.6.3

- The exported function name `isObject` is ambiguous. It is unclear whether it includes functions or not. (It does.) To avoid this confusion, we're deprecating `isObject` and suggesting to use the new export `isPrimitive` instead, that has the opposite answer. IOW, for all `x`, `isObject(x) === !isPrimitive(x)`

# `@endo/patterns` 1.7.0

- `@endo/marshal` introduces an environment variable config option `ENDO_RANK_STRINGS` to change the rank ordering of strings from the current (incorrect) ordering by UTF-16 code unit used by JavaScript's `<` and `.sort()` operations to (correct and OCapN-conformant) ordering by Unicode code point. It currently defaults to "utf16-code-unit-order", matching the previously-unconditional behavior.
  - `@endo/patterns` provides a `compareKeys` partial order that delegates some ordering, including strings, to the rank ordering provided by `@endo/marshal`. So when the `ENDO_RANK_STRINGS` default is not overridden, then `compareKeys` also follows the (incorrect) UTF-16 code unit order. But when it is overridden, then `compareKeys` also follows the (correct) Unicode code-point order.
- In errors explaining why a specimen does not match a pattern, sometimes the error message contains a quoted form of a nested pattern. This quoting was done with `q`, producing an uninformative rendering of these nested patterns. Now this quoting is done with `qp`, which renders these nested patterns into readable [Justin](https://github.com/endojs/Jessie/blob/main/packages/parse/src/quasi-justin.js) source code.

### DIFF
--- a/packages/base64/CHANGELOG.md
+++ b/packages/base64/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+### [1.0.12](https://github.com/endojs/endo/compare/@endo/base64@1.0.11...@endo/base64@1.0.12) (2025-07-12)
+
+**Note:** Version bump only for package @endo/base64
+
+
+
+
+
 ### [1.0.11](https://github.com/endojs/endo/compare/@endo/base64@1.0.10...@endo/base64@1.0.11) (2025-06-17)
 
 **Note:** Version bump only for package @endo/base64
--- a/packages/base64/package.json
+++ b/packages/base64/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@endo/base64",
-  "version": "1.0.11",
+  "version": "1.0.12",
   "description": "Transcodes base64",
   "keywords": [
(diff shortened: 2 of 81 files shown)


erights (approved):
LGTM!

gibson042 (approved):

PR #2889, opened 2025-07-12
chore: Update release instructions for release notes and such

kriskowal:
This change documents some minor changes that have accumulated to our release process since we began posting release notes and a step was obviated by the last Lerna upgrade.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,6 +73,18 @@ https://github.com/endojs/endo/labels/next-release
   +# 0.5.1 (2021-08-12)
   ```
 
+  Also, capture these into a release notes document, where each heading
+  is the name of the package and the version number instead of the
+  version number and release date, like so:
+
+  ```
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -73,6 +73,18 @@ https://github.com/endojs/endo/labels/next-release
...
+  is the name of the package and the version number instead of the
+  version number and release date, like so:
+
+  ```
+  # `ses` 1.0.0
+  ```
+
+  In the release notes document, we do not manually wrap long lines.
+  All paragraphs should be joined into long lines, since Github uses a
+  different flavor of Markdown for descriptions and release notes.
```
boneskull, line 86: What is this "release notes document"? Where does it live?

kriskowal: This is a document that we made as we’re preparing the release and then post on https://github.com/endojs/endo/releases and use for the release PR description

PR #2892, opened 2025-07-15
fix(pass-style): revert precision of deprecated isObject type

erights:
Closes: #XXXX
Refs: https://github.com/endojs/endo/pull/2860

## Description

#2860 deprecated `isObject` as confusing and promoted use of `isPrimitive` instead. However, it also made the type of `isObject` more precise, which accidentally broken a bunch of client code in agoric-sdk (and possibly elsewhere) that relied on the type being less precise. Since `isObject` is now deprecated and all clients should eventually switch to `!isPrimitive` anyway, it costs nothing to lose this extra type precision of `isObject`.

This PR restores the old imprecise type.

### Security Considerations

none
### Scaling Considerations

none

none
### Testing Considerations

none

The point. This PR restores the compatibility that #2860 accidentally broke.
### Upgrade Considerations

none

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,6 +73,18 @@ https://github.com/endojs/endo/labels/next-release
   +# 0.5.1 (2021-08-12)
   ```
 
+  Also, capture these into a release notes document, where each heading
+  is the name of the package and the version number instead of the
+  version number and release date, like so:
+
+  ```
+  # `ses` 1.0.0
+  ```
+
+  In the release notes document, we do not manually wrap long lines.
+  All paragraphs should be joined into long lines, since Github uses a
+  different flavor of Markdown for descriptions and release notes.
+
 * Commit the results.
 
   ```sh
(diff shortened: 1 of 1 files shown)


packages/pass-style/src/passStyle-helpers.js
```diff
@@ -42,9 +42,12 @@ export const isPrimitive = val =>
   !val || (typeof val !== 'object' && typeof val !== 'function');
 harden(isPrimitive);
 
+// NOTE: Do not make this type more precise because it breaks only clients
+// that rely on it being less precise.
 /**
  * @deprecated use `!isPrimitive` instead
- * @type {(val: unknown) => val is (Function | Record<string | symbol, unknown>)}
```
mhofman, line 47: TBH, this was the wrong type in the first place.

Instead of `Function` it should be `Callable`, aka `(...args: any[]) => any`.

Instead of `Record` it likely should just be `object`, but that has issues in non-"typescript strict mode".

mhofman (approved):

kriskowal (commented):
Thank you.

Issue #2886, opened 2025-07-11
compartment-mapper: captureFromMap incompatible with dynamic requires


boneskull:
I'm not sure how I missed this before, but I can note that there is no test of dynamic require which attempts the operations described below.

`mapNodeModules` creates a `CompartmentMapDescriptor` based on the state of a dependency tree.  When passing this `CompartmentMapDescriptor` into `captureFromMap`—to generate a minimal, absolute-path-less policy—any packages which are _only_ required dynamically will be tossed out of the resulting digested  `CompartmentMapDescriptor`; `captureFromMap` only _loads_ modules. Dynamic imports, by definition, are excluded from this operation.

If we use the resulting digested `CompartmentMapDescriptor` to generate policy (via `@lavamoat/node`), packages will be missing and their sources will not be inspected for globals, builtins, etc. (via `lavamoat-core`).  Passing the resulting policy _back_ into `importLocation()` results in many policy enforcement failures because the policy is insufficient.

It may simply be that `captureFromMap()` is not fit for our purpose.  Ultimately, I don't think there _is_ a minimal policy when dynamic requires are involved. We could not know which packages were actually "used"  until _after_ runtime; even then, that minimal policy is not necessarily minimal for the _next_ execution.

If true, then `captureFromMap()`'s scope may need to be explicitly reduced to the archival use-case.

As such I'm not labeling this issue as `bug`.  There may or may not be "action items" here, but I wanted to bring this up in case someone wanted to discuss it or had any _bright ideas._

PR #2893, opened 2025-07-15
feat(compartment-mapper): add retainAll flag to captureFromMap

boneskull:
Closes: #2886
Ref: #2894 (future implementation may make _this_ PR unnecessary)

## Description

This adds a boolean flag to `captureFromMap()` which instructs `digestCompartmentMap()` to ignore the `retained` flag on `CompartmentDescriptor`s and `ModuleDescriptor`s, which ensures no "unused" `CompartmentDescriptor` is dropped from the mapping.

Considered creating a new public function for this, but it did not seem worth the bother. Still, it would be feasible to:

1. Create a new function `captureAllFromMap()`
2. Extract common logic from `captureFromMap()` into a new module-scoped function
3. Call that function from both `captureFromMap()` and `captureAllFromMap()`
4. `captureAllFromMap()` would pass `retainAll: true` to the common function which would hand it off to `digestCompartmentMap()`

### Security Considerations

n/a

### Scaling Considerations

Use of this will result in a large compartment map, which may bloat whatever you're using it for.

n/a

### Testing Considerations

Added fixture and tests for the specific behavior with and without the flag.

This is backwards-compatible.

### Upgrade Considerations

Update `NEWS.md`

### DIFF
--- a/packages/compartment-mapper/src/capture-lite.js
+++ b/packages/compartment-mapper/src/capture-lite.js
@@ -38,6 +38,7 @@
  *   CompartmentMapDescriptor,
  *   ReadFn,
  *   ReadPowers,
+ *   RetainAllOptions,
  *   Sources,
  * } from './types.js'
  */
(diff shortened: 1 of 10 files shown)


boneskull:
Closing in lieu of #2915

PR #2895, opened 2025-07-16
feat(ses): Add assert.makeError and deprecate assert.error as an alias

gibson042:
...as the culmination of general cleanup.

## Description

* Make `assert` an arrow function, resolving a TODO
* Rename StringablePayload to Stringable
* Rename HiddenDetails to DetailsParts
* Improve parameter/variable names
* Improve type documentation and examples
* Add `assert.makeError` and deprecate `assert.error` as an alias
* Add an `@endo/errors` export of `assert.details` with its own name

### Security Considerations

None known.

### Scaling Considerations

n/a

If accepted, [agoric-sdk use of `assert.error`](https://github.com/search?q=repo%3AAgoric%2Fagoric-sdk+%22assert.error%22&type=code) should be replaced with `assert.makeError`.

### Testing Considerations

Existing coverage is sufficient.

`assert.error` is deprecated, but retained as an alias of `assert.makeError` to support existing code.

### Upgrade Considerations

NEWS.md was updated to document the change.

### DIFF
--- a/packages/errors/index.js
+++ b/packages/errors/index.js
@@ -54,9 +54,9 @@ if (missing.length > 0) {
 // and also updates the names of the utility functions.
 const {
   bare,
-  details: redacted,
+  details,
   error,
-  Fail: throwRedacted,
+  Fail,
   makeAssert: _omittedMakeAssert,
   makeError,
   note,
@@ -81,19 +81,21 @@ const bestMakeError = makeError || error;
 export {
   // assertions
   assert,
-  // related utilities that aren't assertions
+  // non-assertion utilities that appear as properties of `assert`
   bareOrQuote as bare,
   bestMakeError as makeError,
+  details,
   note,
   quote,
-  redacted,
-  throwRedacted,
-  // conventional abbreviations and aliases
+  Fail,
+  // conventional abbreviations
   bareOrQuote as b,
+  details as X,
   quote as q,
-  redacted as X,
-  throwRedacted as Fail,
+  // other aliases
   note as annotateError,
+  details as redacted,
+  Fail as throwRedacted,
 };
 
 /**
@@ -116,7 +118,7 @@ export {
  * @returns {T}
  */
 export const hideAndHardenFunction = func => {
-  typeof func === 'function' || throwRedacted`${func} must be a function`;
+  typeof func === 'function' || Fail`${func} must be a function`;
   const { name } = func;
   defineProperty(func, 'name', {
     // Use `String` in case `name` is a symbol.
--- a/packages/errors/index.js
+++ b/packages/errors/index.js
@@ -21,9 +21,8 @@ if (globalAssert === undefined) {
   );
 }
 
-const missing = /** @type {const} */ ([
+const missing = [
   'typeof',
-  'error',
   'fail',
   'equal',
   'string',
@@ -31,12 +30,17 @@ const missing = /** @type {const} */ ([
   'details',
   'Fail',
   'quote',
-  // As of 2024-02, the Agoric chain's bootstrap vat runs with a version of SES that
-  // predates addition of the 'bare' method, so we must tolerate its absence and fall
-  // back to quote behavior in that environment (see below).
+  // As of 2025-07, the Agoric chain's bootstrap vat runs with a version of SES
+  // that predates addition of the 'bare' and 'makeError' methods, so we must
+  // tolerate their absence and fall back to other behavior in that environment
+  // (see below).
   // 'bare',
+  // 'makeError',
   'makeAssert',
-]).filter(name => globalAssert[name] === undefined);
+].filter(name => globalAssert[name] === undefined);
+if (globalAssert.makeError === undefined && globalAssert.error === undefined) {
+  missing.push('makeError');
+}
 if (missing.length > 0) {
   throw Error(
     `Cannot initialize @endo/errors, missing globalThis.assert methods ${missing.join(
@@ -51,9 +55,10 @@ if (missing.length > 0) {
 const {
   bare,
   details: redacted,
-  error: makeError,
+  error,
   Fail: throwRedacted,
   makeAssert: _omittedMakeAssert,
+  makeError,
   note,
   quote,
   ...assertions
@@ -64,10 +69,11 @@ const assert = (value, optDetails, errContructor, options) =>
   globalAssert(value, optDetails, errContructor, options);
 Object.assign(assert, assertions);
 
-// As of 2024-02, the Agoric chain's bootstrap vat runs with a version of SES
-// that predates the addition of the 'bare' method, so we must fall back to
-// quote behavior for that environment.
+// As of 2025-07, the Agoric chain's bootstrap vat runs with a version of SES
+// that predates the addition of the 'bare' and 'makeError' methods, so we must
+// fall back to 'quote' for the former and 'error' for the latter.
 const bareOrQuote = bare || quote;
+const bestMakeError = makeError || error;
 
 // XXX module exports fail if these aren't in scope
 /** @import {AssertMakeErrorOptions, Details, GenericErrorConstructor} from 'ses' */
@@ -77,7 +83,7 @@ export {
   assert,
   // related utilities that aren't assertions
   bareOrQuote as bare,
-  makeError,
+  bestMakeError as makeError,
   note,
   quote,
   redacted,
--- a/packages/ses/src/error/assert.js
+++ b/packages/ses/src/error/assert.js
@@ -47,8 +47,8 @@ import './internal-types.js';
 import { makeNoteLogArgsArrayKit } from './note-log-args.js';
 
 /**
- * @import {BaseAssert, Assert, AssertionFunctions, AssertionUtilities, StringablePayload, DetailsToken, MakeAssert} from '../../types.js'
- * @import {LogArgs, NoteCallback, LoggedErrorHandler} from "./internal-types.js";
+ * @import {BaseAssert, Assert, AssertionFunctions, AssertionUtilities, Stringable, DetailsToken, MakeAssert} from '../../types.js';
+ * @import {LogArgs, NoteCallback, LoggedErrorHandler} from './internal-types.js';
  */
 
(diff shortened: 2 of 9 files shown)


packages/ses/types.d.ts
```diff
@@ -403,43 +391,26 @@ export interface AssertionUtilities {
...
+   * The normal convention is to locally rename `details` to `X` like
+   * `const { details: X, quote: q, Fail } = assert;`.
    * However, note that in most cases it is preferable to instead use the `Fail`
-   * template literal tag (which has the same input signature as `details`
-   * but automatically creates and throws an error):
+   * template literal tag (which has the same input signature but automatically
+   * creates and throws an error):
    * ```js
-   * sky.isBlue() || Fail`${sky.color} should be "blue"`;
+   * sky.isBlue() || assert.Fail`${sky.color} should be "blue"`;
```
erights: Weird that here you change my `Fail` to `assert.Fail` whereas on line 463/428 below, you just use `Fail` unqualified in a similar example. I prefer `Fail` to `assert.Fail` whenever possible.

```suggestion
   * sky.isBlue() || Fail`${sky.color} should be "blue"`;
```

packages/ses/types.d.ts
```diff
@@ -403,43 +391,26 @@ export interface AssertionUtilities {
...
    *
-   * The details template tag returns a `DetailsToken` object that can print
-   * itself with the formatted message in two ways.
-   * It will report full details to the console, but
-   * mask embedded substitution values with their typeof information in the thrown error
-   * to prevent revealing secrets up the exceptional path. In the example
-   * above, the thrown error may reveal only that `sky.color` is a string,
-   * whereas the same diagnostic printed to the console reveals that the
-   * sky was green. This masking can be disabled for an individual substitution value
-   * using `quote`.
```
erights, line 443: This concrete "green" example text deserves to go somewhere. Things moved around enough that perhaps I missed it, but AFAICT you seem to have dropped it.

(Extra credit if we can find a way to slip in an allusion to the Hume/Goldman "grue" and "bleen", but only if it does not detract from the explanation for normal users ;) .)

packages/ses/types.d.ts
```diff
@@ -449,68 +420,54 @@ export interface AssertionUtilities {
...
-   * sky.color === expectedColor || Fail`${sky.color} should be ${q(expectedColor)}`;
+   * actual === expected || Fail`${actual} should be ${q(expected)}`;
    * ```
    */
-  quote(
-    /** What to declassify */
-    payload: any,
-    spaces?: string | number,
-  ): /** The declassified and quoted payload */ StringablePayload;
+  quote(value: any, spaces?: string | number): Stringable;
```
erights: Huh. Neither the old or new text say what the optional `spaces` parameter means, and I no longer remember. Would be nice to fix, but feel free not to.

packages/errors/index.js
```diff
@@ -49,9 +53,10 @@ if (missing.length > 0) {
 const {
   bare,
   details: redacted,
```
erights:
```suggestion
  details: X,
```

packages/errors/index.js
```diff
@@ -49,9 +53,10 @@ if (missing.length > 0) {
...
-  error: makeError,
+  error,
   Fail: throwRedacted,
```
erights:
```suggestion
  Fail,
```

packages/errors/index.js
```diff
@@ -75,7 +81,7 @@ export {
   assert,
   // related utilities that aren't assertions
   bareOrQuote as bare,
-  makeError,
+  bestMakeError as makeError,
   note,
   quote,
   redacted,
```
erights:
```suggestion
  X as redacted, // and deprecate `redacted` if possible
  Fail as throwRedacted, // and deprecate `throwRedacted` if possible
  Fail,
  ...
```

packages/errors/index.js
erights: Or better yet, normalize `assert.X` as the replacement for a deprecated `assert.details`, like this PR does for `assert.makeError` vs `assert.error`.

Need not be in this PR but I'd at least like your opinion ;)

packages/ses/src/error/assert.js
```diff
@@ -74,39 +78,40 @@ const canBeBare = freeze(/^[\w:-]( ?[\w:-])*$/);
...
 };
 freeze(bare);
 
 // /////////////////////////////////////////////////////////////////////////////
 
 /**
- * @typedef {object} HiddenDetails
- *
- * Captures the arguments passed to the `details` template string tag.
+ * @typedef {{ template: TemplateStringsArray | string[], args: any[] }} DetailsParts
```
erights, line 96: I'm ok with the rename if the doc-comment here mentions that a `DetailParts` may hold redacted secrets, and so must not leak. I originally named it `Hidden*` to suggest that such care should be taken.

packages/ses/src/error/assert.js
```diff
@@ -196,7 +198,11 @@ freeze(unredactedDetails);
 export { unredactedDetails };
```
erights, line 201: Surprised to see that I exported `unredactedDetails`. Does not seem to be imported anywhere. If we cannot yet kill the (likely unused) export, could we at least deprecate it?

In any case, feel free not to address this quirk in this PR.

erights:
(I think the two CI errors were flakes, so restarted them. We'll see.)

packages/ses/types.d.ts
kriskowal: It’s the third arg to JSON.stringify and I forgot we threaded it. It’s not a great idea to thread it.

packages/ses/types.d.ts
```diff
@@ -187,6 +187,10 @@ export interface EvaluateOptions {
   __rejectSomeDirectEvalExpressions__?: boolean;
 }
 
+interface Stringable {
+  toString(): string;
+}
```
kriskowal, line 194: Let’s keep a deprecated alias for `StringPayload` so we don’t break anything.

packages/ses/types.d.ts
gibson042: It's now the ```actual === expected || Fail`${actual} should be ${q(expected)}`;``` example for `quote`, but if you prefer the "sky" example I could switch that over.

packages/errors/index.js
gibson042: I agree that the introduction of name `redacted` (and for that matter, likewise `throwRedacted`) is odd. This file exports both `assert` and properties thereof, the latter with both full names and "conventional abbreviations and aliases", and `details` is the only property that is exported but not with its literal name. I'm pushing a commit that adds it.

packages/ses/types.d.ts
```diff
@@ -307,61 +313,54 @@ export type GenericErrorConstructor =
...
 export interface AssertionFunctions extends BaseAssert {
   typeof: AssertTypeof;
 
   /**
-   * The `assert.equal` method
-   *
-   * Assert that two values must be `Object.is`.
+   * Assert that two values are the same as observed by `Object.is`.
    */
   equal<T>(
```
erights, line 350: These additional helpers seemed worth it before we came up with the `|| Fail...` and `|| reject && reject...` patterns. Now that we have those, I wonder if we should deprecate these helpers. They only subtract value compared to the new patterns. 

Doesn't need to happen in this PR but I'd like your opinion.

packages/ses/types.d.ts
erights: I see you did add the sky example, so Resolving. Thanks.

packages/ses/types.d.ts
```diff
@@ -449,70 +427,69 @@ export interface AssertionUtilities {
...
+   * `const { quote: q, Fail } = assert;`):
    * ```js
-   * sky.color === expectedColor || Fail`${sky.color} should be ${q(expectedColor)}`;
+   * actual === expected || Fail`${actual} should be ${q(expected)}`;
    * ```
+   *
+   * The optional `space` parameter matches that of `JSON.stringify`, and is
+   * used to request insertion of non-semantic line feeds, indentation, and
+   * separating spaces in the output for improving readability of objects and
+   * arrays.
```
erights, line 483: Thanks

packages/ses/types.d.ts
```diff
@@ -522,35 +499,25 @@ export interface AssertionUtilities {
    * by inline replacement similar to that of `bestEffortStringify` for producing
    * rendered messages like `(an object) was tagged "[Unsafe bare string]"`).
    */
-  bare(
-    /** What to declassify */
-    payload: any,
-    spaces?: string | number,
-  ): /** The declassified payload without quotes (beware confusion hazard) */
-  StringablePayload;
+  bare(text: string, spaces?: string | number): Stringable;
```
erights, line 504: Why is the first argument changed from `any` to `string`? Doesn't `bare` have essentially the same behavior as `quote` but without the quoting? If it does only take a string, how can the second parameter mean anything?

packages/ses/types.d.ts
gibson042: Hmm, whatever benefit they have comes from the message being optional (e.g., `assert.equal(brand1, brand2)`, `assert.typeof(marshalled, 'string')`) and relying upon stack information to pinpoint the specific issue, and in fact providing a message with the helpers seems like an anti-pattern that should be replaced with ```$condition || Fail`…`;```.

So the question is whether or not auto-generated messages are sufficiently useful to stave off deprecation. And given the simplicity of doing so manually (e.g., ```Object.is(brand1, brand2) || Fail`brands ${brand1} and ${brand2} must match`;```, ```typeof marshalled === 'string' || Fail`marshalled ${marshalled} must be a string`;```), I think I support deprecation.

packages/ses/types.d.ts
gibson042: _Intended_ use of `bare` is with strings, and this type annotation will prompt a warning if it is used with input not known to be a string. But at _runtime_, we decided that it must also handle non-string input by falling back to `quote`, which will also receive the `spaces` argument. But we should probably deprecate or drop that parameter at the next opportunity, because I don't think `b(value, spaces)` should ever pass code review.

PR #2901, opened 2025-07-17
refactor: Embrace default chaining

kriskowal:
Follow-up upon #1514 completion.

We left several notes to our future selves to embrace default chaining when we could.

### DIFF
--- a/packages/compartment-mapper/src/bundle-lite.js
+++ b/packages/compartment-mapper/src/bundle-lite.js
@@ -420,19 +420,22 @@ export const makeFunctorFromMap = async (
     } else {
       module.indexedImports = Object.fromEntries(
         Object.entries(module.resolvedImports).map(([importSpecifier, key]) => {
-          // UNTIL https://github.com/endojs/endo/issues/1514
-          // Prefer: key = aliases.get(key) ?? key;
           const alias = aliases.get(key);
-          if (alias != null) {
-            key = alias;
-          }
-          const module = modulesByKey[key];
+          const module = modulesByKey[alias ?? key];
           if (module === undefined) {
-            throw new Error(
-              `Unable to locate module for key ${q(key)} import specifier ${q(
-                importSpecifier,
-              )}`,
-            );
+            if (alias === undefined) {
+              throw new Error(
+                `Unable to locate module for key ${q(key)} import specifier ${q(
+                  importSpecifier,
+                )}`,
+              );
+            } else {
+              throw new Error(
+                `Unable to locate module for key ${q(alias)} (via alias ${q(key)}) import specifier ${q(
+                  importSpecifier,
+                )}`,
+              );
+            }
           }
           const { index } = module;
           return [importSpecifier, index];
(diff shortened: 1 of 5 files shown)


packages/compartment-mapper/src/bundle-lite.js
```diff
@@ -425,13 +425,7 @@ export const makeFunctorFromMap = async (
...
-          // Prefer: key = aliases.get(key) ?? key;
-          const alias = aliases.get(key);
-          if (alias != null) {
-            key = alias;
-          }
-          const module = modulesByKey[key];
+          const module = modulesByKey[aliases.get(key) ?? key];
           if (module === undefined) {
             throw new Error(
               `Unable to locate module for key ${q(key)} import specifier ${q(
```
erights: Not quite equivalent because the before code might use the value of `alias` in the error message and the after code will only ever use `key`. Aside from syntax, which is the better behavior?

packages/compartment-mapper/src/bundle.js
```diff
@@ -432,13 +432,8 @@ export const makeFunctorFromMap = async (
...
-          const alias = aliases.get(key);
-          if (alias != null) {
-            key = alias;
-          }
-          const module = modulesByKey[key];
+          key = aliases.get(key) ?? key;
+          const module = modulesByKey[key] ?? modulesByKey[importSpecifier];
           if (module === undefined) {
             throw new Error(
               `Unable to locate module for key ${q(key)} import specifier ${q(
```
erights: Same question re error message

packages/compartment-mapper/src/bundle.js
```diff
@@ -432,13 +432,8 @@ export const makeFunctorFromMap = async (
...
         Object.entries(module.resolvedImports).map(([importSpecifier, key]) => {
-          // UNTIL https://github.com/endojs/endo/issues/1514
-          // Prefer: key = aliases.get(key) ?? key;
-          const alias = aliases.get(key);
-          if (alias != null) {
-            key = alias;
-          }
-          const module = modulesByKey[key];
+          key = aliases.get(key) ?? key;
+          const module = modulesByKey[key] ?? modulesByKey[importSpecifier];
```
erights: I see how line 435 is equivalent, modulo the error message question below. But line 436 does not seem to be a refactor of anything from the before code. Is this still a pure refactor?

packages/compartment-mapper/src/bundle-lite.js
kriskowal: Neither. Posting an update that varies the error depending on whether an alias was followed to provide a better debugging experience.

kriskowal (commented):
Will my apologies for overwriting the commit rather than presenting a diff, I’ve posted an update with a more nuanced take for better error messages.

And also rediscovered that there is duplicate code that should have disappeared in factoring bundle-lite out of bundle, which I will address in another change.

erights (approved):
LGTM

PR #2905, opened 2025-07-22
fix(module-source): fix re-exported names in ModuleSource

boneskull:
## Description

This seems very subtly wrong. `Array.prototype.flatMap()` wants its callback to return an array. The values returned by this operation will return different values depending on the count of exported items (mod 2, in fact):

```js
const reexports = {
  foo: [['bar', 'bar'], ['baz', 'baz']],
  spam: [['quux', 'quux']]
};

// current
const a = Object.values(reexports).flatMap(([_, exportName]) => exportName); // ['baz', 'baz', undefined]

// fixed
const b = Object.values(reexports).flat().map(([_, exportName]) => exportName); // ['bar', 'baz', 'quux']
```

### Security Considerations

n/a

### Scaling Considerations

This notices many more re-exports, which may cause more memory pressure.

I'm not sure.

### Testing Considerations

I added a test which demonstrates the correct value of `ModuleSource.exports`.

It's a bugfix, but it might surface other problems!

### Upgrade Considerations

Should probably update `NEWS.md`.

### DIFF
--- a/packages/module-source/src/module-source.js
+++ b/packages/module-source/src/module-source.js
@@ -90,7 +90,9 @@ export function ModuleSource(source, opts = {}) {
     [
       ...keys(liveExportMap),
       ...keys(fixedExportMap),
-      ...values(reexportMap).flatMap(([_, exportName]) => exportName),
+      ...values(reexportMap)
+        .flat()
+        .map(([_, exportName]) => exportName),
     ].sort(),
   );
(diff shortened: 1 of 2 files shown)


naugtur (approved):
it should have been 
flatMap - flats first, maps next
mapFlat - what current flatmap is doing
flatMapFlat - you know what to expect, right?

PR #2909, opened 2025-07-24
fix(compartment-mapper): do not blast named default exports in CJS

boneskull:
## Description

This fixes a situation where a module which uses `Object.defineProperty` to define its own `default` property. Previously, the CJS parser wrapper erroneously overwrote the property with the value of `default.default` (which, practically, is often `undefined`).

Example:

```js
const a = (exports.a = {});

Object.defineProperty(exports, 'default', {
  get: () => a,
  enumerable: true,
});
```

In the above, `a` would be exported, but `default` would be lost.

This more accurately mimics Node.js' behavior.

The problem was originally detected in `@babel/core` which uses this pattern in its CJS distfiles.

* * *

- Also adds types for the scaffolding harness. 

### Security Considerations

n/a

### Scaling Considerations

n/a

n/a

### Testing Considerations

Test added.

Improves ecosystem compat.

### Upgrade Considerations

n/a

### DIFF
--- a/packages/compartment-mapper/test/scaffold.js
+++ b/packages/compartment-mapper/test/scaffold.js
@@ -1,6 +1,6 @@
-// @ts-nocheck
 import 'ses';
 import fs from 'fs';
+
 import crypto from 'crypto';
 import url from 'url';
 import { ZipReader, ZipWriter } from '@endo/zip';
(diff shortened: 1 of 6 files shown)


packages/compartment-mapper/test/scaffold.js
```diff
@@ -560,22 +586,25 @@ export function scaffold(
...
-        languageForExtension,
-        commonjsLanguageForExtension,
-        moduleLanguageForExtension,
-        workspaceLanguageForExtension,
-        workspaceCommonjsLanguageForExtension,
-        workspaceModuleLanguageForExtension,
-        ...additionalOptions,
-      });
+      const expectedSha512 = await hashLocation(
+        /** @type {HashPowers} */ (readPowers),
```
boneskull, line 590: This is to note that the type signature for `makeReadPowers` is insufficient; the return type depends on the parameters.  If, for example, a `crypto: CryptoInterface` object is provided, then the return type should extend `HashPowers`.  If it contains both `fs: FsInterface` and `url: UrlInterface`, then the return type should extend `ReadNowPowers`.  The return type always extends `MaybeReadPowers`.  And combinations thereof

PR #2910, opened 2025-07-25
fix(ses): re-exported names available in imported modules

boneskull:
## Description

This fixes a situation in which module _A_ re-exports name _X_ from module _B_, and then imports module _C_, which in turn imports _X_ from _A_.

Originally encountered in the source code for `terser`.

**This change needs a close reading,** as I'm not confident in its correctness.

### Security Considerations

n/a

### Scaling Considerations

Incurs another loop through update records.

n/a

### Testing Considerations

Test added

Improve ecosystem compat

### Upgrade Considerations

n/a


boneskull:
It looks like I broke something, but not sure what.

boneskull:
OK, this looks better.

boneskull (commented):
This fix absolutely works, but it is also incorrect.

packages/ses/src/module-instance.js
```diff
@@ -412,24 +412,24 @@ export const makeModuleInstance = (
...
+            get() {
+              return value;
+            },
+            set: undefined,
+            enumerable: true,
+            configurable: false,
+          };
+        }
       }
     }
```
naugtur, line 438: I now understand why this issue exists and why moving the for loop into the other for loop helps.

```
instance.execute(); // bottom up cycle tolerant
```
is not fully cycle tolerant.

Here's what happens when the test case is run. inside and outside refers to the state of the candidateAll inside and outside the for loop.

```
 imports called { moduleSpecifier: './index.js', depth: 1 }
   before instance execute ./a.js
     imports called { moduleSpecifier: './a.js', depth: 2 }
     inside skipped, zero loop
     outside { moduleSpecifier: './a.js', depth: 2 } [["default",false]]
   after instance execute ./a.js
   inside { moduleSpecifier: './index.js', specifier: './a.js', depth: 1 } [["default",false],["a",null]]
   before instance execute ./c.js
     imports called { moduleSpecifier: './c.js', depth: 2 }
     before instance execute ./index.js
     after instance execute ./index.js
     inside { moduleSpecifier: './c.js', specifier: './index.js', depth: 2 } [["default",false]]
     outside { moduleSpecifier: './c.js', depth: 2 } [["default",false]]
   after instance execute ./c.js
   inside { moduleSpecifier: './index.js', specifier: './c.js', depth: 1 } [["default",false],["a",null]]
   outside { moduleSpecifier: './index.js', depth: 1 } [["default",false],["a",null]]
```

Note that the only chance we get to set the value for `'a'` before c.js causes `instance.execute()` cll for `index.js` is inside the loop.

----

At this point, calling the same logic to populate exportsProps in the loop over reexportsMap fixes the testcase without rearranging the loop scope.

Gotta add more exports to the index.js in test to see whether that's sufficient or whether I've found an explanation why the change in this PR is correct.

naugtur: regular exports work as expected in a cycle - expanded the test to verify that they do.

PR #2911, opened 2025-07-27
feat(errors): Rejectors (as suggested by #2285)

erights:
Closes: #2285
Refs: #2851 #2917 #2918

## Description

Follows approximately @gibson042 's suggestions at #2285, with some minor differences I'd like @gibson042 's feedback on:

- I changed the template type in the `Rejector` type to end with `=> never` rather than `=> false`.
- I added a `{false}` type disjunct to `Rejector`
- made the `reject` parameter to the `confirmFoo` functions non-optional, since it would also be supplied by the `isFoo` or `assertFoo` functions. Typically, only the latter two would be exported. The only reason I can see to make `confirmFoo`'s `reject` parameter optional is to be able to export and use it directly as the predicate, which would be an abstraction leak.

As stated in #2285, the purpose is to migrate existing uses of `Checker` to use `Rejector` instead. The first real test for that will be to redo #2851 to use Rejectors instead of Checkers, while also giving `passStyleOf` a variant which can avoid erroring on non-passables. And then to see how it compares with the old ones on speed. Likewise to see if @endo/patterns speeds up.

### Security Considerations

Since this PR itself is only additive, if it is correct, none. If so, then the redo of #2851 hopefully should improve security since code using Rejectors should overall be simpler and easier to review for correctness that code using Checkers.

### Scaling Considerations

Most of the point. We hope that code using Rejectors will be faster than current code using Checkers. We should measure the redo of #2851 to find out.

Most code uses only `isFoo` or `assertFoo` functions, whose behavior should be identical. Only users of definers of the old `checkFoo` or the new `confirmFoo` functions need to understand the difference. Not sure where these are or should be documented.

### Testing Considerations

The test included in this PR also demonstrate the `confirm/is/assert` pattern that Rejectors are designed to support.

For this PR by itself, since it is only additive, none. Further, the design of Rejectors is intended to enable code to switch over without causing external compat considerations. The redo of #2851 will be a good first test.

For any old `checkFoo` code using Checkers rather than Rejectors, if those `checkFoo` functions were exported, we can deprecate and continue to export them during the transition to further avoid compat issues.

### Upgrade Considerations

None of this causes any change to data that is communicated or stored. Given that the `isFoo` and `assertFoo` functions are behaviorally identical and that any exported `checkFoo` functions continue to be exported (as deprecated), there should be no upgrade concerns.

- [ ] NEWS.md

### DIFF
--- a/packages/errors/rejector.js
+++ b/packages/errors/rejector.js
@@ -1,0 +1,111 @@
+import { Fail } from './index.js';
+
+// TODO `RejectOptPrefixPair` should be a circular type, which TS, surprisingly
+// to me, seems unhappy about. The `any` within it should be another
+// `RejectOptPrefixPair`
+/**
+ * @typedef {string|number|bigint} RejectPrefix
+ * @typedef {RejectPrefix|undefined} RejectOptPrefix
+ * @typedef {undefined | [any, RejectPrefix]} RejectOptPrefixPair
+ */
+
+/**
+ * @param {RejectOptPrefixPair} prePre
+ * @param {RejectPrefix} postPre
+ */
+const makeReject = (prePre, postPre) => {
+  /**
+   * @param {TemplateStringsArray | string[]} template
+   * @param {...any} args
+   * @returns {never}
+   */
+  const reject = ([...template], ...args) => {
+    let [pre, post] = [prePre, postPre];
+    for (;;) {
+      switch (typeof post) {
+        case 'string': {
+          break;
+        }
+        case 'number':
+        case 'bigint': {
+          // based on throw-labeled.js
+          post = `[${post}]`;
+          break;
+        }
+        default: {
+          post = `${post}`;
+          break;
+        }
+      }
+      template[0] = `${post}: ${template[0]}`;
+      if (pre === undefined) {
+        break;
+      }
+      [pre, post] = pre;
+    }
+    throw Fail(template, ...args);
+  };
+  /**
+   * @param {RejectPrefix} prefix
+   * @returns {typeof reject}
+   */
+  reject.nestRejectMethod = prefix => makeReject([prePre, postPre], prefix);
+  return harden(reject);
+};
+
+/**
+ * Given a Rejector `reject`, returns a Rejector just like it, except that
+ * if the returned Rejector is invoked to create and throw an error,
+ * the error message will begin with all the prefixes already accumulated by
+ * the paren `reject`, followed by `prefix` and `': '` (colon space),
+ * followed by the normal `Fail` formulated error message.
+ *
+ * This is recursive of course, so it the returned Rejector is used as an
+ * argument in another call to `nestReject`, then that prefix
+ * (with its own `': '`) would appear between this prefix and `Fail`
+ * formulated error message.
+ *
+ * As a convenience, if the prefix is a number or a bigint, it is first
+ * converted to a string within square brackets, to be suggestive of an index.
+ * If this is not your intention, then convert it to a string first
+ * yourself.
+ *
+ * @param {RejectOptPrefix} prefix
+ * @param {Rejector} reject
+ * @returns {Rejector}
+ */
+export const nestReject = (prefix, reject) => {
+  if (reject === false) {
+    return false;
+  }
+  if (prefix === undefined) {
+    return reject;
+  }
+  if (reject === Fail) {
+    return makeReject(undefined, prefix);
+  }
+  // @ts-expect-error purposely omitted nestRejectMethod from type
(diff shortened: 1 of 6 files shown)


packages/common/rejector.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * @import {AssertionUtilities} from 'ses';
+ */
+
+/**
+ * @typedef {false | AssertionUtilities['Fail']} Rejector;
```
erights: I don't understand lint error
```
packages/common/rejector.js:6:3 - [error] Failed to find JSDoc tag for Rejector after parsing comment, please file a bug report
```
Please help. Thanks.

erights: The `false |` lets me make the `reject` parameter non-optional.

Separately, I saw no need to weaken the template return type from `=> never` to `=> false`. By not weakening it, the type system may (or may not) help us remember to stick to the pattern 
```js
reject && reject`...`
```
rather that just
```js
reject`...`
```

erights:
@gibson042 I further revised this to also subsume `applyLabelingError`. Instead, one `confirm*` function would pass a prefixed `reject` to its nested `confirm*` function. See the test case.

AFAICT, this add no overhead to normal use, and beats both my 'applyLabelingError` and your alternative with an explicit stack for the nested use where no rejection actually happens. PTAL with both uses in mind. Thanks!

packages/common/test/rejector.test.js
```diff
@@ -20,7 +21,8 @@ const confirmString = (candidate, reject) =>
  * @returns {candidate is 'foo'}
  */
 const confirmFoo = (candidate, reject) =>
-  (confirmString(candidate, reject) && candidate === 'foo') ||
+  (confirmString(candidate, nestReject('str: ', reject)) &&
```
erights: This `nestReject` does the same work as `applyLabelingError`

(explanation for reviewers. no change suggested)

erights:
See https://github.com/endojs/endo/pull/2914 for more examples of use

packages/common/rejector.js
```diff
@@ -0,0 +1,72 @@
...
+  };
+  /**
+   * @param {string|number} prefix
+   * @returns {typeof reject}
+   */
+  reject.pre = prefix => makeReject([prePre, postPre], prefix);
+  return harden(reject);
+};
+
+export const Reject = makeReject(undefined, '');
```
erights: This and the exported `pre` message appearing in the type are unnecessary. Instead, all `Reject` uses should just using `Fail` instead, as originally planned. Have `nestReject` feature test on the presence of the `pre` message, which can now have a different name. The `pre` need not appear in the exported type.

erights: Done

erights:
Moved from @endo/common to @endo/errors, which is where someone would think to look for it.

I kept it as a distinct export, following our new conventions, but it is so small I am unconvinced it is worth it. Attn @kriskowal

packages/errors/rejector.js
```diff
@@ -0,0 +1,72 @@
...
+      }
+      [prePre, postPre] = prePre;
+    }
+    throw Fail(template, ...args);
+  };
+  /**
+   * @param {string|number} prefix
+   * @returns {typeof reject}
+   */
+  reject.pre = prefix => makeReject([prePre, postPre], prefix);
```
erights: This method name is visible. Either it should not be (best) or it should be a good name. Since no one outside this package will normal need to be aware of it, there's no penalty for a long name.

packages/errors/rejector.js
```diff
@@ -0,0 +1,72 @@
...
+ * ```
+ * If `cond` is truthy, that is the value of the expression.
+ * Else if `reject` is false, it is the value
+ * Otherwise, invoke `reject` just like you would invoke `Fail`, with the
+ * same template arguments. This throws the same kind of Error object that
+ * `Fail` would throw, except the accumulated prefixes are preprended to
+ * the error message.
+ *
+ * Because a `reject` variable of type `Rejector` may be false, the `pre`
+ * method is not typically how a `confirm*` function would pass a
```
erights: doc-comment should not mention pre method since it is no longer in the type.

packages/common/rejector.js
erights: Problem went away.

packages/errors/rejector.js
erights: done

packages/errors/rejector.js
erights: done

erights:
Closing in favor of #2923

PR #2913, opened 2025-07-30
fix(compartment-mapper): fix handling of module field in package.json

boneskull:
## Description

This changes how package exports are inferred from within `mapNodeModules`, such that a `.` external alias is always present for `package.json`s without an `exports` field.

### Testing Considerations

Needs tests!

Improves ecosystem compatibility.


boneskull:
I don't think I need this.

PR #2914, opened 2025-07-31
refactor: use confirmFoo pattern

erights:

Closes: #XXXX
Refs: #XXXX

## Description

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations

### DIFF
--- a/packages/common/ident-checker.js
+++ b/packages/common/ident-checker.js
@@ -1,7 +1,5 @@
-// TODO Complete migration of Checker type from @endo/pass-style to @endo/common
-// by having @endo/pass-style, and everyone else who needs it, import it from
-// `@endo/common/ident-checker.js`.
 /**
+ * @deprecated Use `Rejector` instead
  * @callback Checker
  * Internal to a useful pattern for writing checking logic
  * (a "checkFoo" function) that can be used to implement a predicate
@@ -30,6 +28,7 @@
  * identity function, but is typed as a `Checker` to indicate its
  * intended use.
  *
+ * @deprecated use `Reject` instead
  * @type {Checker}
  */
 export const identChecker = (cond, _details) => cond;
--- a/packages/pass-style/src/passStyle-helpers.js
+++ b/packages/pass-style/src/passStyle-helpers.js
@@ -3,7 +3,8 @@
 import { X, q } from '@endo/errors';
 
 /**
- * @import {Checker, PassStyle, JSPrimitive} from './types.js';
+ * @import {Rejector} from '@endo/common/rejector.js';
+ * @import {JSPrimitive, PassStyle} from './types.js';
  */
 
 const { isArray } = Array;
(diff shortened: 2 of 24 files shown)


packages/common/ident-checker.js
```diff
@@ -30,6 +28,7 @@
  * identity function, but is typed as a `Checker` to indicate its
  * intended use.
  *
+ * @deprecated use `Reject` instead
```
erights, line 31: This one should be "use `false` instead, with explanation

packages/pass-style/src/passStyle-helpers.js
```diff
@@ -70,76 +71,41 @@ harden(isTypedArray);
...
 
-/**
- * Below we have a series of predicate functions and their (curried) assertion
- * functions. The semantics of the assertion function is just to assert that
- * the corresponding predicate function would have returned true. But it
- * reproduces the internal tests so failures can give a better error message.
- *
- * @type {Checker}
- */
-export const assertChecker = (cond, details) => {
```
erights, line 81: Should restore but deprecate it

packages/pass-style/src/passStyle-helpers.js
```diff
@@ -70,76 +71,41 @@ harden(isTypedArray);
...
-/**
- * Returns a template literal tag function to fail the provided Checker with details.
- * The name must be short for ergonomic inline use as in:
- * ```
- * return checkCondition(...) || (!!check && CX(check)`...`);
- * ```
- *
- * @param {Checker} check
- */
-export const CX = check => {
```
erights, line 96: We could restore and deprecate, but I hope we don't need to

packages/pass-style/src/passStyle-helpers.js
```diff
@@ -162,31 +128,32 @@ const makeCheckTagRecord = checkProto => {
...
-      checkProto(tagRecord, getPrototypeOf(tagRecord), check)
+      (typeof confirmOwnDataDescriptor(
+        tagRecord,
+        Symbol.toStringTag,
+        false,
+        reject,
+      ).value === 'string' ||
+        (reject &&
+          reject`A [Symbol.toStringTag]-named property must be a string: ${tagRecord}`)) &&
+      checkProto(tagRecord, getPrototypeOf(tagRecord), reject)
```
erights, line 156:
```suggestion
      confirmProto(tagRecord, getPrototypeOf(tagRecord), reject)
```

packages/pass-style/src/passStyle-helpers.js
```diff
@@ -195,8 +162,7 @@ const makeCheckTagRecord = checkProto => {
 export const checkTagRecord = makeCheckTagRecord(
   (val, proto, check) =>
     proto === objectPrototype ||
-    (!!check &&
-      check(false, X`A tagRecord must inherit from Object.prototype: ${val}`)),
+    (!!check && check`A tagRecord must inherit from Object.prototype: ${val}`),
```
erights, line 165: How did I miss this? What else might I have missed?

erights: worse. It looks like I made a bad change

packages/pass-style/src/types.d.ts
```diff
@@ -228,4 +228,8 @@ export type CopyTagged<
  */
 export type InterfaceSpec = string;
 
+/**
+ * Consider this export deprecated.
+ * Import `Checker` directly from `'@endo/common/ident-checker.js'` instead.
+ */
```
erights, line 234: Is this comment still needed, or is the type itself now deprecate on its own?

packages/common/ident-checker.js
erights: In #2911 done

packages/pass-style/src/passStyle-helpers.js
erights: In #2917 done

packages/pass-style/src/passStyle-helpers.js
erights: Not exported from package, so we don't need to. Whew!

packages/pass-style/src/passStyle-helpers.js
erights: done

packages/pass-style/src/passStyle-helpers.js
erights: done

packages/pass-style/src/types.d.ts
erights: Transferred the question to #2917

PR #2915, opened 2025-07-31
feat(compartment-mapper): add "forceLoad" option to captureFromMap()

boneskull:
Prior art:

Ref: #2856
Ref: #2864
Ref: #2876

## Description

This adds option `forceLoad` to `captureFromMap()`. This is an optional `string[]` of keys of the `CompartmentMapDescriptor` also provided to `captureFromMap()`.

After loading the entry Compartment and attenuators Compartment, the Compartments having names (practically speaking, these will be _locations_) mentioned in `forceLoad` will then be loaded _if they were not already_.

This option can be used to support dynamic requires and imports which would otherwise be omitted from the captured `CompartmentMapDescriptor` via digestion.

> - Refactored `capture-lite.js`; stuffed all of the Compartment-loading business into a function.
> - Added test and fixture that shows how a `Compartment` which would otherwise be omitted from the captured Compartment Map is included when used with `forceLoad`.

### Security Considerations

Just the usual disclaimers about being careful what you put in the compartment map.

### Scaling Considerations

There's a tradeoff: use of `forceLoad` has performance implications, since it causes (potentially _many_) module sources to be read and analyzed which would not otherwise be.

### Testing Considerations

There are probably some conditionals I've left uncovered.

Backwards-compatible

### Upgrade Considerations

This is a user-facing change which should be described in `NEWS.md`.


packages/compartment-mapper/src/capture-lite.js
```diff
@@ -79,12 +85,160 @@ const captureCompartmentMap = (compartmentMap, sources) => {
...
+      if (!compartment) {
+        throw new ReferenceError(
+          `No compartment found for ${q(compartmentName)}`,
+        );
+      }
+      const compartmentOwnModuleDescriptor =
+        compartmentDescriptor.modules[compartmentDescriptor.name];
+
+      if (!compartmentOwnModuleDescriptor?.module) {
+        throw new Error(
```
boneskull: Is a `CompartmentDescriptor` _not having a `ModuleDescriptor` referencing itself_ considered malformed?

packages/compartment-mapper/src/capture-lite.js
```diff
@@ -79,12 +85,158 @@ const captureCompartmentMap = (compartmentMap, sources) => {
...
+ * laden with useful metadata. This, for example, could be used for automatic
+ * policy generation.
+ *
+ * @param {ReadFn | ReadPowers} readPowers Powers
  * @param {CompartmentMapDescriptor} compartmentMap
  * @param {CaptureLiteOptions} [options]
  * @returns {Promise<CaptureResult>}
  */
-export const captureFromMap = async (powers, compartmentMap, options = {}) => {
+export const captureFromMap = async (
```
boneskull, line 249: besides the refactors, I renamed:

- parameter `powers` ➡️ `readPowers`
- internal name of option `Compartment` ➡️ `CompartmentOption`
- module-scoped `defaultCompartment` ➡️ `DefaultCompartment`
- internal entry Compartment `compartment` ➡️ `entryCompartment`
- internal `Record<string, Compartment>` post-linking `compartments` ➡️ `linkedCompartments`

packages/compartment-mapper/src/policy.js
```diff
@@ -15,6 +15,7 @@
  *   Attenuator,
  *   SomePolicy,
  *   PolicyEnforcementField,
+ *   SomePackagePolicy,
```
boneskull, line 18: Without this change, the generated `policy.d.ts` has no reference to `SomePackagePolicy` and is thus broken.  I would expect `lint:types` to report this error, but it doesn't and I don't know why.

boneskull: ref: https://github.com/endojs/endo/issues/2921

naugtur:
With outside-facing compartment map transforms the use of a marker in compartment map indicating what to load looks appealing again.
Instead of loading entry by default, the approach would be to load everything marked as a linkingStartNode, which would be true for entry by default. 

As a result adding missing links and linking start nodes based on policy overrides from lavamoat could happen in one transform provided by lavamoat.

boneskull:
@naugtur Yes, I may have mentioned above that I considered a similar design. I'm happy to go that way if we feel it's a better fit.  Still, there will need to be an option _somewhere_ because how else is `mapNodeModules` going to know which `CompartmentDescriptor` to flag for loading?

packages/compartment-mapper/src/capture-lite.js
```diff
@@ -36,25 +36,31 @@
...
 import { detectAttenuators } from './policy.js';
 import { unpackReadPowers } from './powers.js';
-import { digestCompartmentMap } from './digest.js';
 
-const { freeze, assign, create } = Object;
+const { freeze, assign, create, keys } = Object;
+const { stringify: q } = JSON;
 
-const defaultCompartment = Compartment;
+const DefaultCompartment = Compartment;
```
kriskowal: It might be good to say `globalThis.Compartment` on the RHS so it’s not a reference error, in cases where it’s intended to be injected.

packages/compartment-mapper/src/capture-lite.js
```diff
@@ -79,12 +85,161 @@ const captureCompartmentMap = (compartmentMap, sources) => {
...
+     * This index increments in the order in which compartments finish
+     * loading—_not_ the order in which they began loading.
+     */
+    let loadedCompartmentIndex = 0;
+    await Promise.all(
+      compartmentsToLoad.map(
+        async ([compartmentName, compartment, moduleSpecifier]) => {
+          await compartment.load(moduleSpecifier);
+          log(
+            `Force-loaded Compartment: ${q(compartmentName)} (${(loadedCompartmentIndex += 1)}/${compartmentsToLoadCount})`,
```
kriskowal: Style, please move the increment expression into a statement. Not many people know whether postincrement expressions evaluate to the pre or post value.

packages/compartment-mapper/src/capture-lite.js
```diff
@@ -79,12 +85,161 @@ const captureCompartmentMap = (compartmentMap, sources) => {
...
+   * Will not load the "attenuators" `Compartment`, nor will it load any
+   * `Compartment` having a non-empty value in `sources` (since it is presumed
+   * it has already been loaded).
+   *
+   * @param {Record<string, Compartment>} compartments
+   * @returns {Promise<void>} Resolves when all appropriate compartments are
+   * loaded.
+   */
+  const forceLoadCompartments = async compartments => {
+    const compartmentsToLoad = forceLoad.reduce((acc, compartmentName) => {
```
kriskowal: Consider using an ordinary loop. Consider naming the accumulator after what it holds or how those will be used.

packages/compartment-mapper/src/capture-lite.js
```diff
@@ -79,12 +85,161 @@ const captureCompartmentMap = (compartmentMap, sources) => {
...
+      }
+
+      const compartment = compartments[compartmentName];
+      if (!compartment) {
+        throw new ReferenceError(
+          `No compartment found for ${q(compartmentName)}`,
+        );
+      }
+      const compartmentOwnModuleDescriptor =
+        compartmentDescriptor.modules[compartmentDescriptor.name];
```
kriskowal: This seems like a fragile invariant, that the modules map will contain an entry that matches the name property of the `package.json`. I think that’s true now because of the reflexive imports feature, but I think the name `'.'` is a more reliable canonical name for the “main” module of a package, since it’s always the same, if present.

I see that the subsequent check accounts for packages that don’t have a main export. Should this be framed more generally in terms of the module locations to load?

packages/compartment-mapper/src/capture-lite.js
boneskull: `globalThis` being undefined per ESLint seems like a misconfiguration.  SES also does not assign its globals to `globalThis`, which incurs a TS error.  I just wanted to note this.

Regardless, your suggestion is correct, so I will just squelch the TS error for now.

packages/compartment-mapper/src/capture-lite.js
```diff
@@ -79,12 +89,168 @@ const captureCompartmentMap = (compartmentMap, sources) => {
...
+    if (policy) {
+      // retain all attenuators.
+      await Promise.all(
+        detectAttenuators(policy).map(attenuatorSpecifier =>
+          attenuatorsCompartment.load(attenuatorSpecifier),
+        ),
+      );
+    }
+
+    await preloader(linkedCompartments);
```
naugtur, line 230: This assumes we'd be preloading packages through their default entry based on a canonical name.
I provides for a nice design.
But the goal is to be able to get hints from LavaMoat config to include certain dependencies and I'm not sure this provides for the case where a package has multiple independent exports.

naugtur: nevermind, I saw the `entry` option now.

packages/compartment-mapper/src/capture-lite.js
```diff
@@ -128,25 +301,27 @@ export const captureFromMap = async (powers, compartmentMap, options = {}) => {
...
-        attenuatorsCompartment.load(attenuatorSpecifier),
-      ),
-    );
-  }
+
+  await loadCompartments(
+    linkedCompartments,
+    entryCompartment,
+    attenuatorsCompartment,
+  );
```
naugtur, line 324: This is only used in capture-lite. Would an application work at runtime when its policy generation situation depended on this? I'm assuming it'd go and do a dynamic import for what's not been preloaded. But wanted to point at it and say I'm not sure if using it only in one place is enough.

boneskull: It would go and do a dynamic import.  This is basically here to account for dynamic imports in a load-only situation.

naugtur (approved):

PR #2917, opened 2025-08-01
refactor(pass-style): use confirm/reject pattern in pass-style

erights:

Closes: #XXXX
Refs: #XXXX

## Description

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations

### DIFF
--- a/packages/pass-style/src/internal-types.js
+++ b/packages/pass-style/src/internal-types.js
@@ -1,8 +1,9 @@
 export {};
 
-/** @import {Checker} from './types.js' */
-/** @import {PassStyle} from './types.js' */
-/** @import {PassStyleOf} from './types.js' */
+/**
+ * @import {Rejector} from '@endo/errors/rejector.js';
+ * @import {PassStyle} from './types.js';
+ */
 
 /**
  * The PassStyleHelper are only used to make a `passStyleOf` function.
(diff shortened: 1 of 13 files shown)


packages/pass-style/src/types.d.ts
```diff
@@ -228,4 +228,8 @@ export type CopyTagged<
  */
 export type InterfaceSpec = string;
 
+/**
+ * Consider this export deprecated.
+ * Import `Checker` directly from `'@endo/common/ident-checker.js'` instead.
+ */
```
erights, line 234: Is this comment still needed, or is the type itself now deprecated on its own?

packages/pass-style/src/internal-types.js
```diff
@@ -17,11 +18,12 @@ export {};
...
+ * @property {(candidate: any, reject: Rejector) => boolean} confirmCanBeValid
+ * If `confirmCanBeValid` returns true, then the candidate would
  * definitely not be valid for any of the other helpers.
  * `assertRestValid` still needs to be called to determine if it
- * actually is valid, but only after the `canBeValid` check has passed.
+ * actually is valid, but only after the `confirmCanBeValid` check has passed.
+ *
  * @property {(candidate: any,
  *             passStyleOfRecur: (val: any) => PassStyle
  *            ) => void} assertRestValid
```
erights, line 29: In a later PR, we should make this a confirm method as well, enabling us to test passsibility, or to call a "predicate" variant of `passStyleOf` that returns `undefined` rather than throwing for a non-passible argument.

No change suggested.

gibson042: Agreed.

packages/pass-style/src/passStyle-helpers.js
```diff
@@ -84,62 +86,41 @@ export const assertChecker = (cond, details) => {
...
- * return checkCondition(...) || (!!check && CX(check)`...`);
- * ```
- *
- * @param {Checker} check
- */
-export const CX = check => {
-  const reject = (T, ...subs) => check(false, X(T, ...subs));
-  return reject;
-};
-harden(CX);
```
gibson042, line 100: 🥳

packages/pass-style/src/types.d.ts
gibson042: I don't think it hurts to have this comment.

gibson042 (approved):
Nice!

PR #2918, opened 2025-08-02
refactor(patterns): use confirm/reject pattern in @endo/patterns

erights:

Closes: #XXXX
Refs: #XXXX

## Description

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations

### DIFF
--- a/packages/patterns/src/keys/checkKey.js
+++ b/packages/patterns/src/keys/checkKey.js
@@ -1,39 +1,32 @@
 /// <reference types="ses"/>
 
-import { X, q, Fail } from '@endo/errors';
-import { identChecker } from '@endo/common/ident-checker.js';
-import {
-  assertChecker,
-  Far,
-  getTag,
-  makeTagged,
-  passStyleOf,
-  isAtom,
-} from '@endo/pass-style';
+import { Fail, q } from '@endo/errors';
+import { Far, getTag, makeTagged, passStyleOf, isAtom } from '@endo/pass-style';
 import {
   compareAntiRank,
   makeFullOrderComparatorKit,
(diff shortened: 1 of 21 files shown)


packages/patterns/src/keys/checkKey.js
```diff
@@ -81,34 +74,34 @@ export const checkKey = (val, check) => {
...
   if (result) {
     // Don't cache the undefined cases, so that if it is tried again
-    // with `assertChecker` it'll throw a diagnostic again
+    // with `Fail` it'll throw a diagnostic again
     // @ts-expect-error narrowed
     keyMemo.add(val);
   }
-  // Note that we do not memoize a negative judgement, so that if it is tried
+  // Note that we must not memoize a negative judgement, so that if it is tried
   // again with a checker, it will still produce a useful diagnostic.
```
erights:
```suggestion
  // again with `Fail`, it will still produce a useful diagnostic.
```

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -583,28 +593,31 @@ const makePatternKit = () => {
...
+  const matches = (specimen, patt) => confirmMatches(specimen, patt, false);
 
   /**
    * Returning normally indicates success. Match failure is indicated by
    * throwing.
    *
    * @param {any} specimen
    * @param {Pattern} patt
-   * @param {string|number} [label]
+   * @param {string|number} [label] // really?
```
erights: Remove ironic comment. This is correct.

erights:
Closing in favor of #2922

PR #2922, opened 2025-08-06
refactor(patterns): use confirm/reject style in @endo/patterns

erights:

Closes: #XXXX
Refs: #XXXX

## Description

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations


packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -407,47 +404,48 @@ const makePatternKit = () => {
...
    * @returns {boolean}
    */
-  const checkMatchesInternal = (specimen, patt, check) => {
-    // Worth being a bit verbose and repetitive in order to optimize
-    const patternKind = kindOf(patt, check);
+  const confirmMatchesInternal = (specimen, patt, reject) => {
+    // If `patt` does not have a kind (or is not a pattern)
+    // then, even if `reject === false`, we should throw about
+    // `patt` anyway.
+    const patternKind = confirmKindOf(patt, reject || Fail);
```
gibson042: I wonder if such validation should bypass `reject` entirely.
```suggestion
    const patternKind = confirmKindOf(patt, Fail);
```

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -566,45 +568,58 @@ const makePatternKit = () => {
...
 
+  /**
+   * @param {any} specimen
+   * @param {Pattern} pattern
+   * @param {string} prefix
+   * @param {Rejector} reject
+   * @returns {boolean}
+   */
+  const confirmNestedMatches = (specimen, pattern, prefix, reject) =>
+    applyLabelingError(confirmMatches, [specimen, pattern, reject], prefix);
```
gibson042, line 604: I like that this provides more context in the event of an error.

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -943,24 +957,23 @@ const makePatternKit = () => {
...
-        check(
-          /** @type {bigint} */ (specimen) >= 0n,
-          X`${specimen} - Must be non-negative`,
-        ) &&
-        checkDecimalDigitsLimit(specimen, decimalDigitsLimit, check)
+        confirmKind(specimen, 'bigint', reject) &&
+        // eslint-disable-next-line @endo/restrict-comparison-operands
+        (specimen >= 0n ||
+          (reject && reject`${specimen} - Must be non-negative`)) &&
+        confirmDecimalDigitsLimit(specimen, decimalDigitsLimit, reject)
```
gibson042: An alternative to suppressing the eslint checks:
```js
      const { decimalDigitsLimit } = limit(limits);
      const typedSpecimen = /** @type {bigint} */ (specimen);
      return (
        confirmKind(specimen, 'bigint', reject) &&
        (typedSpecimen >= 0n ||
          (reject && reject`${typedSpecimen} - Must be non-negative`)) &&
        confirmDecimalDigitsLimit(typedSpecimen, decimalDigitsLimit, reject)
      );
```

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -970,26 +983,23 @@ const makePatternKit = () => {
...
+        confirmKind(specimen, 'string', reject) &&
           // eslint-disable-next-line @endo/restrict-comparison-operands
-          (/** @type {string} */ (specimen).length <= stringLengthLimit ||
-          check(
-            false,
-            X`string ${specimen} must not be bigger than ${stringLengthLimit}`,
-          ))
+          (specimen.length <= stringLengthLimit ||
+          reject && reject`string ${specimen} must not be bigger than ${stringLengthLimit}`)
       );
```
gibson042: Likewise here:
```js
      const { stringLengthLimit } = limit(limits);
      const typedSpecimen = /** @type {string} */ (specimen);
      return (
        confirmKind(specimen, 'string', reject) &&
        (typedSpecimen.length <= stringLengthLimit ||
          (reject && reject`string ${typedSpecimen} must not be bigger than ${stringLengthLimit}`))
      );
```

packages/patterns/src/patterns/patternMatchers.js

packages/patterns/src/patterns/patternMatchers.js

packages/patterns/src/patterns/types.js
```diff
@@ -39,14 +40,14 @@ export {};
...
- * ) => boolean} checkMatches
+ * ) => boolean} confirmMatches
  * @property {(specimen: any, patt: Pattern) => boolean} matches
  * @property {(specimen: any, patt: Pattern, label?: string|number) => void} mustMatch
  * @property {(patt: Pattern) => void} assertPattern
  * @property {(patt: any) => boolean} isPattern
  * @property {GetRankCover} getRankCover
  * @property {MatcherNamespace} M
- * @property {(specimen: Passable, check?: Checker) => Kind | undefined} kindOf
+ * @property {(specimen: Passable, reject: Rejector) => Kind | undefined} kindOf
```
mhofman, line 52: This change is erroneous. The original type seems wrong too, but the `kindOf` in the PatternKit does not have a second argument.

erights: Thanks for catching! Already merged, so if you would like to correct, I'm happy to review. Thanks!

mhofman: Working on it, trying to get a type assertion that would catch this

erights: Ah. The old `kindOf` has been split in the new `confirmKindOf` and `kindOf`. What's above is mistakenly trying to type what is now `confirmKindOf`, which is not actually exposed.

mhofman: Ugh the whole `patternMatchers.js ` file is `//@ts-nocheck` which is why we didn't get an error. Fantastic...

mhofman: https://github.com/endojs/endo/pull/2970

PR #2928, opened 2025-08-06
feat(errors): self censoring stack frames

erights:
Closes: #XXXX
Refs: #XXXX

## Description

Introduces `hideAndHardenFunction` for functions to opt out of having their frames be visible in abbreviated stacks. See changes to `lockdown.md` for more.

```js
import { hideAndHardenFunction } from '@endo/errors';

export const foo() = {...stuff...};
hideAndHardenFunction(foo);

console.log(foo.name); // '__HIDE_foo'
```

If a function `foo` is first frozen with `hideAndHardenFunction(foo)` rather than `freeze(foo)` or `harden(foo)`, then `foo.name` is changed from `'foo'` to `'__HIDE_foo'`. 

When `stackFiltering: 'concise'` or `stackFiltering: 'omit-frames'`, then (currently only on v8), the stack frames for function whose `name` begins with `'__HIDE_'` are omitted from the stacks reported by our causal console. See 

### Security Considerations

By allowing functions to opt into this themselves, a sneaky programmer could use this to hide their attack functions from appearing in frames of abbreviated stacks. However, more security conscious scenarios should consider `stackFiltering: 'verbose'` or `stackFiltering: 'shorten-paths'` anyway, which do not drop frames. Nevertheless, this security hazard is real.

### Scaling Considerations

none

`hideAndHardenFunction` has a good doc-comment, so presumably that will appear in docs generated from doc-comments.

### Testing Considerations

As always, test that are about what should appear in stacks are hard to write regression tests for, since their details depend on, for example, platform and precise line and column numbers. This would make for goldens that are *far* to fragile. Instead, I manually tested these, and I captured the results of manual tests of `deep-send.test.js` in `lockdown.md`.

Production code should never depend on the contents of error messages or error stacks. However, tests can. We know of goldens for error messages, but none of error stacks. This PR only affects error stacks, so we do not expect any compat problems even with golden tests.

### Upgrade Considerations

none


packages/eventual-send/test/deep-send.test.js
```diff
@@ -36,5 +36,4 @@ const testDeepStacksE = test.macro({
   },
 });
 
-test(testDeepStacksE, 'console.log', _t => console.log.bind(console));
```
erights, line 39: This extra test dates from when we didn't know how well `t.log` works.

packages/eventual-send/test/deep-stacks.test.js
```diff
@@ -25,5 +25,4 @@ const testDeepStacksWhen = test.macro({
   },
 });
 
-test(testDeepStacksWhen, 'console.log', _t => console.log.bind(console));
```
erights, line 28: This extra test dates from when we didn't know how well `t.log` works.

erights:
Moved deep-send.test.js and deep-stacks.test.js from eventual-send to errors in order to avoid introducing a devDependency cycle.

packages/errors/test/deep-send.test.js
```diff
@@ -0,0 +1,45 @@
+// This file is not really useful as an
```
erights, line 1: Not really a new file. Rather, migrated from eventual-send/test, but somehow github didn't notice.

packages/errors/test/deep-stacks.test.js
```diff
@@ -0,0 +1,21 @@
+// This file is not really useful as an
```
erights, line 1: Not really a new file. Rather, migrated from eventual-send/test, but somehow github didn't notice.

kriskowal (approved):
For me, this change evokes the [Grand Galactic Inquisitor](https://venturebrothers.fandom.com/wiki/The_Grand_Galactic_Inquisitor) from Venture Brothers (parody of Jonny Quest), a 12 foot alien robot that constantly, loudly demands that you ignore him by shouting “IGNORE ME”. For this reason alone, I would be tempted to name `hideAndHardenFunction` to `ignoreMe` and prefix all such functions `__IGNORE_ME_`, but although I would suggest finding a more concise name than `hideAndHardenFunction`, and a prefix that better communicates the intent of `__HIDE_` to a casual debugger who is unaware of this subsystem, I cannot seriously recommend “ignore me”. I would not object, though 😉

This otherwise looks good to me.

Issue #2894, opened 2025-07-15
compartment-mapper: Compartment Map Transforms


boneskull:
## What is the Problem Being Solved?

Currently, `mapNodeModules()` accepts a policy, which—if present—it uses to _omit_ `ModuleDescriptor`s from the resulting `CompartmentDescriptor.module`.  Specifically, if a _package policy_ exists for the package being crawled, it will only add `ModuleDescriptor` entries for those packages (which are dependencies) mentioned in the policy.  If no package policy is found, then all dependencies are added.

### Compartment Map Transform

I'm proposing a new concept: the _Compartment Map Transform_.

The Compartment Map Transform accepts a `CompartmentMapDescriptor` and options (including policy), performs operations on the `CompartmentMapDescriptor`, then returns it (probably easiest to actually mutate it in-place since the object is rather deep).  Async transforms can be supported unless there's a specific reason not to.

`mapNodeModules()` would accept a new option, `compartmentTransforms` of type `CompartmentMapTransformFn[]`. Before resolving, `mapNodeModules()` would apply each of these _in order_, returning the final result.

For backwards compatibility, the default value of `compartmentTransforms` should be an array containing a new transform which mimics the current behavior (considering package policy) by _omitting_ values from the `CompartmentMapDescriptor.modules` objects.  We can call this function `enforcePolicyCompartmentMapTransform` for lack of a better idea.  This function will be included in the public API so that users can concatenate it with other transforms.  `translateGraph()` would change such that it is no longer in the business of checking policy.

### Optional New Transform

While we can omit `ModuleDescriptor`s based on policy, it'd be nice to be able to _create_ `ModuleDescriptor`s based on policy! Currently, `ModuleDescriptor`s are not created without an associated entry in the dependency graph.

In the case of dynamic imports/requires, a situation arises wherein a package _A_ instructed by package _B_ to load package _C_. There is _no other relationship_ between package _A_ and package _C_; there is no dependency relationship described in the package descriptor(s).

> Example: a legacy ESLint config, where ESLint (_B_) loads an `.eslintrc` file (_A_) which is configured to use a the `@endo/ses` plugin (_C_).

I'd like to create a new transform (and maybe even provide it from `@endo/compartment-mapper`) which would create `ModuleDescriptor` references in the `CompartmentMapDescriptor` based on the presence of a canonical name (mapping to an extant `CompartmentDescriptor`) within package policy.

[...]

PR #2929, opened 2025-08-07
feat(compartment-mapper)!: create Compartment Map Transforms

boneskull:
Closes: #2894

## Description

**BREAKING CHANGE: The `CompartmentDescriptor.label` property now represents a _canonical name_. `CompartmentDescriptor.path` has been removed. Filenames in created archives are no longer percent-encoded. Compartments no longer have implicit access to ancestors during dynamic requires.  Access to the root compartment via policy is now supported using the canonical name `$root$`. Many type changes.**

This is a whole-ass heap of changes because the Compartment Map Transform implementation is based on the breaking changes to compartment maps. 😢  It may be possible to split this into two PRs (the Compartment Map Transform implementation following one containing the breaking changes), but it would involve reverting changes to `mapNodeModules()` and many tests.
 
* * *

At a high level, this provides an API for consumers to make arbitrary changes to `CompartmentMapDescriptor`s via plugin-like Compartment Map Transform functions.

### Compartment Map Transforms

These transforms are applied by `mapNodeModules()` immediately after creation of the initial `CompartmentMapDescriptor`. Two transforms are built-in, and are always executed:

- `enforcePolicyTransform` - when a policy is provided, this transform is responsible for removing `ModuleDescriptor`s from the `CompartmentDescriptor`s in the map which are not explicitly allowed by policy. This used to be `mapNodeModules()`' responsibility.
- `createReferencesByPolicyTransform` - This is new behavior and can be thought of as the _inverse_ of `enforcePolicyTransform`; this transform _adds_ `ModuleDescriptor`s and `ScopeDescriptor`s to `CompartmentDescriptor`s based on policy.  This solves quite a bit of our policy-related problems at once and means we can revert some of my hacks for dynamic requires (no more implicit access to "related" compartments; no more `CompartmentDescriptor.compartments` property).
  
Each `CompartmentMapTransformFn` receives arguments containing a `CompartmentMapTransformContext` object. This is a public API containing helper functions. See the type definitions for more information on what's included.

### Type Overhaul

This overhauls the types for `CompartmentMapDescriptor` such that the `CompartmentMapDescriptor` returned by `captureFromMap()` is _not_ the same type as the `CompartmentMapDescriptor` returned by `mapNodeModules()` (which has always been true, but was not previously expressed).  In addition, the `ModuleDescriptor` and `ModuleSource` types have been narrowed. Because `ModuleDescriptor` conflicts with `ses`' `ModuleDescriptor`, this type has been renamed to `ModuleDescriptorConfiguration` because `ModuleDescriptorDescriptor` smells like Java and @kriskowal is allergic to abbreviations.

**I am open to any suggestions for naming these new types**; there are a lot of them and others probably better understand the intent behind them.

### Label is now a Canonical Name

This changes the `CompartmentDescriptor.label` property to be a _canonical name_ which matches that found in policy.  Since the canonical name is now owned by the `CompartmentDescriptor` object itself, we can drop the `path` property and simplify several other policy-related utility functions.

This had a specific impact on filenames in archives; while `foo-v1.2.3` would be used verbatim as a "directory" name, `bar>foo` became `bar%3Efoo` which ultimately meant it could not be successfully matched to a `Compartment`.  To solve this, I changed how the filepath is computed to match how the filepath is retrieved.  Previously, computation involved passing a name and module specifier thru `URL`, which caused the percent-encoding.  Now, we just make a `/` sandwich from the two--which happens to be the same strategy as how filenames were computed for retrieval. 

> _Recommendation_: This should be a function that both creation & retrieval call; I think they were simply out-of-sync.

### More Logging

The `log` option is now handled by other public APIs. When running tests, you can set env var `SCAFFOLD_LOGGING=1` to cause log messages to be echoed thru AVA's `t.log()`. I added a dev dep on `@endo/env-options` to do this, but I probably didn't need to.

### Using `assert.quote()` 

Replaced `JSON.stringify` with `assert.quote` where appropriate (which is most places).

### Common Type Guards

Many public APIs want to check what kind of `ModuleDescriptorConfiguration` or `ModuleSource` they are working with.  These checks are now type-safe via guards found in `src/guards.js`.

### Security Considerations

If you use policy, the `CompartmentMapDescriptor` returned by `mapNodeModules()` will more accurately reflect what you meant.  Access to the entry compartment by other compartments may be bad; use `$root$` with caution.

### Scaling Considerations

`mapNodeModules()` is likely slower, but not _an order of magnitude_ slower.

Lots of user-facing changes need documentation

### Testing Considerations

Tests were harmed during the conception of this PR.

Improves ecosystem compatibility for running tools such as Webpack and ESLint which often want to touch files in entry Compartment.

### Upgrade Considerations

This PR includes **BrEaKiNg ChAnGeS**.  I am not sure if archives created previously will still be loadable; if policy is involved, things will probably break.

### DIFF
--- a/packages/compartment-mapper/src/types/compartment-map-schema.ts
+++ b/packages/compartment-mapper/src/types/compartment-map-schema.ts
@@ -13,7 +13,7 @@ import type {
   ENTRY_COMPARTMENT,
 } from '../policy-format.js';
 import type { CanonicalName } from './canonical-name.js';
-import type { FileUrlString, PackageDescriptor } from './external.js';
+import type { FileUrlString } from './external.js';
 import type { SomePackagePolicy } from './policy-schema.js';
 import type { LiteralUnion } from './typescript.js';
 
@@ -104,8 +104,6 @@ export interface PackageCompartmentDescriptor
   scopes: Record<string, ScopeDescriptor<FileUrlString>>;
 
   sourceDirname: string;
-
-  retained?: never;
 }
 
 /**
@@ -154,7 +152,7 @@ export type CompartmentDescriptorWithPolicy<
  * A compartment descriptor digested by `digestCompartmentMap()`
  */
 export interface DigestedCompartmentDescriptor
-  extends CompartmentDescriptor<SourceModuleDescriptorConfiguration> {
+  extends CompartmentDescriptor<ModuleDescriptorConfiguration> {
   path: never;
   retained: never;
   scopes: never;


boneskull:
Needs some snapshot fixes.

packages/compartment-mapper/src/digest.js
```diff
@@ -8,19 +8,33 @@
...
+ *   DigestedCompartmentMapDescriptor,
+ *   DigestedCompartmentDescriptor,
  * } from './types.js'
  */
 
-import { pathCompare } from '@endo/path-compare';
-import { assertCompartmentMap, stringCompare } from './compartment-map.js';
+import {
+  assertDigestedCompartmentMap,
+  assertFileCompartmentMap,
```
naugtur: is this unused?

packages/compartment-mapper/src/digest.js
```diff
@@ -175,7 +198,7 @@ const renameSources = (sources, compartmentRenames) => {
 };
```
naugtur, line 198: (barely actionable) I'm a bit lost in the layers between import/archive/bundle top level APIs and then the -lite bits and then the digest. It's not intuitive to me what divides responsibility into those layers. That might be on me, I'll put some more effort into understanding it but I consider it signal that it could be made easier to reason about (or diagrammed?)

packages/compartment-mapper/src/import-archive-lite.js
```diff
@@ -161,22 +165,22 @@ const makeArchiveImportHookMaker = (
           )} in archive ${q(archiveLocation)}`,
         );
       }
-      if (module.deferredError !== undefined) {
+      if (isErrorModuleDescriptorConfiguration(module)) {
```
naugtur, line 168: these are definitely a step in the right direction. It'd be great to have a relationship diagram of all the descriptor related types handy at all times too.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -993,8 +1093,10 @@ export const compartmentMapForNodeModules = async (
     dev = false,
     commonDependencies = {},
     policy,
+    policyOverride,
```
naugtur, line 1095: (should) I disagree with node-modules being aware of the concept of policyOverride. I'm not sure if I like Endo being aware of it at all. Might be just a naming issue, I'm not sure what it is yet. I couldn't find a single example of them being passed in too. IT seems like it's a dedicated pass-through for lavamoat to put things in the context of transform, which can be achieved by `maketransform: ({policyOverride}) => (stuff) => {}`

packages/compartment-mapper/src/node-modules.js
```diff
@@ -1028,51 +1130,43 @@ export const compartmentMapForNodeModules = async (
...
     conditions,
-    policy,
+    { policy, log },
+  );
+
+  // always run our transforms first
+  const transforms = [
+    ...defaultCompartmentMapTransforms,
+    ...compartmentMapTransforms,
+  ];
```
naugtur, line 1158: transforms should be outside of node-modules. everything they rely on (and themselves) gets here by prop drilling. Might as well be applied on the outside. 
 One way to do that:
   - add a compartmentMapBuilder
   - make it call mapNodeModules
   - give it the responsibility for validation
   - give it the responsibility for transforms
   - potentially also give it the responsibility for adding package policy (in a transform or not) so that node-modules becomes unaware of what policy even is.

packages/compartment-mapper/src/compartment-map-transforms/index.js
```diff
@@ -0,0 +1,136 @@
...
+) => {
+  await null;
+  assert(optionsForTransforms !== undefined, 'optionsForTransforms expected');
+
+  const context = makeCompartmentMapTransformContext(
+    compartmentMap,
+    canonicalNameMap,
+    optionsForTransforms,
+  );
+
```
naugtur, line 117: (not sure about any of this, but posting anyway) each transform iterates over the whole compartmentMap on its own. They contain the unnecessary flexibility of running their own iteration. I know you said they have effects that require global coordination, but I'd like to challenge that. Let's expore the possibility that a transform consisting of `init()` or a constructor and `transformCompartment` where init creates a context object that's available to transformCompartment (either via scope or by being passed around as first arg) would be enough for all reasonable purposes. We could then iterate compartments once and for each apply the changes. 
  Both existing transforms are making changes in a sequence without depending on reading the entire map first, which leads me to a conclusion that they don't depend on owning their iteration.
  Even if we don't apply them in one iteration, taking the concern of skipping ATTENUATORS_COMPARTMENT away from transforms and letting them have a separate method for transforming the entry compartment would simplify (IMHO) reading them.

packages/compartment-mapper/src/compartment-map-transforms/transforms.js
```diff
@@ -0,0 +1,351 @@
...
+ *
+ * @param {unknown} value
+ * @returns {value is PropertyPolicy}
+ */
+const isPropertyPolicy = value =>
+  !!value &&
+  typeof value === 'object' &&
+  !isArray(value) &&
+  values(value).every(item => typeof item === 'boolean');
+
```
naugtur, line 51: (nit) Split each transform into a separate file, aggregate builtin transforms in index.js

packages/compartment-mapper/src/compartment-map-transforms/transforms.js
```diff
@@ -0,0 +1,351 @@
...
+   * @param {FileUrlString} compartmentDescriptorName - The name of the compartment
+   * descriptor being updated.
+   * @param {string} canonicalName - The canonical name of the compartment
+   * descriptor.
+   * @param {string} policyCanonicalName - The canonical name of the policy
+   * compartment descriptor.
+   * @throws {ReferenceError} If the policy compartment descriptor cannot be
+   * found.
+   */
+  const updateCompartmentDescriptor = (
```
naugtur: (nit) this whole function could be extracted from the local context and bound to the stuff it needs so that it'd be easier to understand what it relies on.

packages/compartment-mapper/src/compartment-map-transforms/transforms.js
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    assert(
+      compartmentDescriptor,
+      `No CompartmentDescriptor for name ${q(compartmentDescriptorName)}`,
+    );
+
+    const packagePolicy = getPackagePolicy(
+      compartmentDescriptor,
+      policy ?? policyOverride,
```
naugtur, line 319: this seems to be the only use of this varaible that gets drilled down from afar. I really don't see the point of it being passed through all those layers and the name makes this treatment here incorrect unless the only purpose of it is to add missing stuff to compartment map, which I know from wider context is the case. That being said, this approach seems incorrect to me. if we're using policyOveride for the lavamoat usecase here, it should either already be merged into policy before it becomes endo policy or passed into a custom transform that lavamoat provides. SO either this gets extracted and goes into lavamoat to serve only lavamoat purposes or we make sure it only relies on endo policy and not on a concept that doesn't really exist in endo. (I accept this might be a misunderstanding stemming form policyOverride meaning something else)

packages/compartment-mapper/src/types/compartment-map-schema.ts
```diff
@@ -8,80 +8,230 @@
...
+  | 'import-hook'
+  | 'digest'
+  | 'node-modules';
+
+export interface BaseModuleDescriptorConfiguration {
+  deferredError?: string;
+
+  retained?: true;
+
+  createdBy?: ModuleDescriptorConfigurationCreator;
```
naugtur: This field being consistently populated relies on good will now. Let's either wrap the object in something that will figure out who is creating the descriptor and automate adding this field, or (seems easier) make the field mandatory so that implementers are forced to add it. Force its existence in validation too, so even badly typing the transform will not prevent it from being enforced. (nit) it'd be nice if the field name was more indicative of it being an internal parameter mostly used for debugging/understanding what's going on.

packages/compartment-mapper/src/digest.js
boneskull: yes, good catch

packages/compartment-mapper/src/digest.js
boneskull: Might be helpful to look at Kris' PR that split these up.  It attempts to explain this in the docblocks at the top of the files, too, I think.

packages/compartment-mapper/src/node-modules.js
boneskull: This is a carve-out, yes.  I hesitated to add this and the "create" transform.  These could both live in `@lavamoat/node` instead (with consequences; see below)

packages/compartment-mapper/src/node-modules.js
boneskull: I see that point.  Originally I had thought these transforms could be applicable to all manner of `CompartmentMapDescriptor`s, which set me off on trying to determine what those _are_, exactly (and you see the result here).  But at the time I believed (I am not sure any longer) that the `CompartmentMapTransformContext` API would necessarily differ based on the subtype of `CompartmentMapDescriptor`.  And since `mapNodeModules()` is the only place that uses a `PackageCompartmentMapDescriptor` and presumably transforms only work on `PackageCompartmentMapDescriptor`s, I found it appropriate to stuff in there.

Furthermore, the `enforcePolicyTransform` _should_ be run by `mapNodeModules()` or we've introduced another breaking change.  Might be worth it?

I will go back and validate my assumptions.  I think since I added the canonical name as the `label` prop, it would be feasible to extract into its own module and possibly work on any `CompartmentMapDescriptor`.

**EDIT**: FWIW _not_ including the "creation" transform in Endo breaks a bunch of stuff that used to work w/r/t dynamic requires. Might be fine, but it's just more ruin.

packages/compartment-mapper/src/compartment-map-transforms/index.js
boneskull: You're suggesting, instead, a `CompartmentDescriptorTransform`, right?  Where iteration is owned by Endo, yes?

At any rate, such a transform _must_ have access to the entire compartment map.  There's no other way to create relationships between two arbitrary compartment descriptors without it.  We'd need to supply the name of the entry compartment in this init context (otherwise it'd need to iterate looking for the compartment descriptor having the `ENTRY_COMPARTMENT` `label`) as well.

They don't need to touch `ATTENUATORS_COMPARTMENT`, no.  But they do need to be able to monkey with the entry compartment.

> Recall these can be async, which may negatively impact performance if we were to iterate this way; there would be at minimum _transformer-count_ * _compartment-count_  async frames added and then popped off the stack because these things should not run concurrently

Ultimately I am not sure that the damage that can be done to a `CompartmentDescriptor`  is much less worse than the damage that could be done to a `CompartmentMapDescriptor`.  

If we _really_ wanted to constrain things and ensure referential integrity, then the `CompartmentMapTransformContext` should have methods for modifying the compartment map (how many, exactly?); it should otherwise be read-only. And all of _that_ points to `CompartmentMapDescriptor` being a _class_ and owning responsibility for maintaining its own integrity.

I'm not actually opposed to any of that, since I think managing a compartment map containing string references is a footgun & prone to errors.  IMO, It'd be better to have a properly serializable/deserializable class maintaining its own object references.

packages/compartment-mapper/src/compartment-map-transforms/transforms.js
boneskull: ~~Yeah, that's probably good here. I am so used to pulling closures out to avoid re-creating them over and over, but that's not happening here.~~

packages/compartment-mapper/src/compartment-map-transforms/transforms.js
boneskull: I needed a way to affect the "create" transform w/o affecting the "enforce" transform.  This would be for the policy gen case.  We'd use the translated lavamoat policy override to tell the "create" transform to make our associations from it.  The `policy` here will be `undefined` (which also causes the "enforce" transform to bail).  No packagepolicy is applied to `CompartmentDescriptor.policy` in this case.

We can absolutely remove `policyOverride` if we rip out the "create" transform.  Removing it _does_ have a negative impact on the functionality of `captureFromMap` w/r/t dynamic imports/requires—which might not necessarily be bad, since the intent of `captureFromMap` is kind of fuzzy nowadays.

At runtime, `policy` will be preferred over `policyOverride` if both are provided (there are tests for the interaction between the two options; you must have missed them).

packages/compartment-mapper/src/types/compartment-map-schema.ts
boneskull: I didn't make it mandatory because I was concerned that we wouldn't want it to be in an archive. 

I can come up with some convention for this; `retained` should also be renamed, as it's an internal parameter which is transient within the digestion process.

packages/compartment-mapper/src/compartment-map-transforms/index.js
boneskull: ...but at some point this PR needs to stop expanding, I suppose.

packages/compartment-mapper/src/compartment-map-transforms/transforms.js
boneskull: Wait, I think I get what you mean now.

packages/compartment-mapper/src/node-modules.js
naugtur: There might be a misunderstanding here. I wanted transforms to happen outside node-modules.js but not much further away. The work of the module ends with the transforms, so moving them outside didn't seem that big of a deal. 

>  the enforcePolicyTransform should be run by mapNodeModules() or we've introduced another breaking change

how?

packages/compartment-mapper/src/compartment-map-transforms/index.js
naugtur: That would be too compelx. POJO representation makes debugging easier and leaves more flexibility for future decisions. A class would cement the design and be costly to change later. I was thinking of limiting the flexibility a little, not entirely; mostly in order to hide some complexity from transforms.

packages/compartment-mapper/src/compartment-map-transforms/transforms.js
naugtur: Something is not right with the design here and I don't think there has to be impact on functionality to fix the design.
Maybe the transforms are not independent afterall and need to go together (eliminating the need for a separate policyOverride)
Maybe the logic of applying transforms needs to change to acomodate the create transform working with dynamic requires even if passed in by lavamoat.

At this point I'm convinced a better design exists and we are close, no need to turn everything upside down to get to it.

Issue #2908, opened 2025-07-24
`overrideDebug` reports do not render Errors correctly, just `{}`


michaelfig:
## Describe the bug

The `overrideDebug` reports no longer output error messages and stack traces.  This makes debugging code with many modules impossible to find where something trips over the override mistake.

## Steps to reproduce

1. Go to the top of the `endo` checkout
2. `yarn install`
3. Run `sh ./repro.sh`, whose contents are:
```sh
#! /bin/sh
DEBUG_LOCKDOWN_OPTIONS='{"errorTaming":"unsafe","stackFiltering":"verbose","overrideTaming":"severe","overrideDebug":["constructor"]}'

cat <<EOF > ./override-mistake.mjs
import '@endo/init';

const o = new Object();
o.constructor = 42;
EOF
LOCKDOWN_OPTIONS=$DEBUG_LOCKDOWN_OPTIONS node ./override-mistake.mjs
```
4. Observe output:
```console
$ ./repro.sh
'@endo/lockdown' sniffed and found a 'LOCKDOWN_OPTIONS' environment variable

SES Enabling property overrides
  {}
$
```

Note that the `Error`, whose stack trace would have given valuable debugging information, is rendered just as `{}`.

## Expected behavior

Render an error indicating the `constructor` override mistake instead of just `{}`.

## Platform environment

 - What OS are you using? macOS Sequoia 15.5
 - What version of Node.js? v20.18.3
 - Is there anything special/unusual about your platform? No, not really.
 - What version of Endo are you using? `@endo/import-bundle@1.5.2-9-g0927be926`

## Additional context

n/a

## Screenshots

Console output as above, not pretty pictures.

erights:
@mfig, when I follow your repro instructions on Node 20 I get exactly the buggy symptoms you report. On Node 18, I get the full diagnostic we expect:

```
'@endo/lockdown' sniffed and found a 'LOCKDOWN_OPTIONS' environment variable

SES Enabling property overrides
  [TypeError <Object <Object <[Object: null prototype] {}>>>: Override property constructor
  at set constructor (file:///Users/markmiller/src/ongithub/endojs/endo/packages/ses/src/enable-property-overrides.js:116:22)
  at file:///Users/markmiller/src/ongithub/endojs/endo/override-mistake.mjs:4:15
  at ModuleJob.run (node:internal/modules/esm/module_job:195:25)
  at async ModuleLoader.import (node:internal/modules/esm/loader:336:24)
[...]

Issue #945, opened 2021-12-01
Separate console taming and causal console


mhofman:
Currently the causal console (`consoleTaming: "safe"`) does a lot of magic. In particular it replaces error objects with the decorated error name, and separately calls `console.debug` with the stack and related errors and information (from `assert.note`), all as strings.

`consoleTaming: "unsafe"` is not appropriate because, some properties of the start console may reveal powers (e.g. the `stdout` stream in node). Furthermore because of #636, the override of `Error.prototype.constructor` causes node's inspect to fail and print confusing error info (#944).

I propose to separate console taming from the causal console and error restoring mechanisms. Console taming would wrap all initial console methods to leave a safe console object in compartments (as long as we trust the initial methods to not mutate the arguments). Console augmenting would have different levels:
- None: passes argument through without changing anything or making extra console calls. A virtualized console aware of SES can use the `getStackString` and other note inspection powers to detail and serialize the console calls as it wishes. E.g. a plugin for sentry.
- Minimal: recreate a new error object with a `stack` own property containing the untamed stack, an `constructor` own property to satisfy node's inspect, and potentially a `cause` own property to convey some of the causal capabilities of SES. We could also imagine adding a `note` or `details` own property with the details from `assert.note`. It would not do anything else beside substituting the tamed error objects for detailed ones.
- Causal: The current causal console logic which serializes errors and their details as multiple console calls.

mhofman:
Also the causal console should be updated to natively support `AggregateError` and `.cause`

erights:
I like the shape of this plan. Your idea to synthesize an unredacted error object from the redacted error + side info is superb. You should take it farther. The synthesized error should have an unredacted message.

mhofman:
Related:
- #731

Issue #944, opened 2021-11-30
bare Error instance does not log correctly


kumavis:
ses: (`master` at time of writing [ac1cc0](https://github.com/endojs/endo/tree/ac1cc06f3665c94ace2f551a07072cc094706a62)) 
node: v14.17.6 
```js
require('ses/lockdown');
lockdown({ errorTaming: 'unsafe', stackFiltering: 'verbose', consoleTaming: 'unsafe', domainTaming: 'unsafe' })
console.error(new Error('boom'));
```

logs only
```
{}
```

error stack and message are intact

kumavis:
with `overideTaming: 'min'`
```
[Error: boom
  at file:///home/xyz/Development/endo/packages/ses/abc.js:10:13
  at ModuleJob.run (internal/modules/esm/module_job.js:170:25)
  at async Loader.import (internal/modules/esm/loader.js:178:24)
  at async Object.loadESM (internal/process/esm_loader.js:68:5)]
```

with `overideTaming: 'moderate'`**(default)**
```
{}
```

with `overideTaming: 'severe'`
```
[Error <Object <[Object: null prototype] {}>>: boom
  at file:///home/xyz/Development/endo/packages/ses/abc.js:10:13
  at ModuleJob.run (internal/modules/esm/module_job.js:170:25)
  at async Loader.import (internal/modules/esm/loader.js:178:24)
  at async Object.loadESM (internal/process/esm_loader.js:68:5)]
```

kumavis:
```js
lockdown({ overrideTaming, errorTaming: 'unsafe', stackFiltering: 'verbose', consoleTaming: 'unsafe' });
console.log(Object.getOwnPropertyDescriptors(Error.prototype));
```


- min:
```
{
  constructor: {
    value: [Function: Error],
    writable: false,
    enumerable: false,
    configurable: false
  },
  name: {
    get: [Function: getter],
    set: [Function: setter],
    enumerable: false,
    configurable: false
  },
  message: {
    value: '',
    writable: false,
    enumerable: false,
    configurable: false
  },
  toString: {
    value: [Function: toString],
    writable: false,
    enumerable: false,
    configurable: false
  }
}
```

- moderate (default):
```
{
  constructor: { get: {}, set: {}, enumerable: false, configurable: false },
  name: { get: {}, set: {}, enumerable: false, configurable: false },
  message: { get: {}, set: {}, enumerable: false, configurable: false },
[...]

Issue #636, opened 2021-03-25
Where the Node console gets confused is `constructor` is an accessor


erights:
Following in the vscode debugger, I found where the Node inspector gets confused if a `constructor` is an accessor rather than data property. In something`/internal/util/inspect.js` I see

```js
    const descriptor = ObjectGetOwnPropertyDescriptor(obj, 'constructor');
    if (descriptor !== undefined &&
        typeof descriptor.value === 'function' &&
        descriptor.value.name !== '' &&
        tmp instanceof descriptor.value) {
```

@bmeck @michaelfig @kumavis @kriskowal  Where do I find the actual source for this, so I can file a PR to fix this to work with constructors than have been made accessors in order to work around the override mistake?

@kumavis this is why `{errorTaming: 'unsafe', consoleTaming: 'unsafe'}` causes the built-in node console to print those errors as `{}`. Try changing to `{errorTaming: 'unsafe', consoleTaming: 'unsafe', overrideTaming: 'min'}` and you'll see your stack traces. Try changing to 
```js
{
  errorTaming: 'unsafe', 
  consoleTaming: 'unsafe', 
  overrideTaming: 'min',
  stackFiltering: 'verbose',
}
```
and you'll see the full original stack traces.

Btw, I *strongly* recommend against `{consoleTaming: 'unsafe'}`. Is there a reason you want to turn that on?

erights:
See #944

erights:
See https://github.com/endojs/endo/issues/2908

erights:
The snippet above has grown into the code at 

https://github.com/nodejs/node/blob/fc3f19ef932444e8de5d69fe3fb5033b2e0ec2e4/lib/internal/util/inspect.js#L627

If the prototypes whose constructors were made into accessors were added to `wellKnownPrototypes`, our problems my well disappear.

kriskowal:
@erights Can we close this issue?

erights:
I don't think so. At least not until we see if anything has changed on the node side. If not, then we should reexamine if we can do anything about it on our side.

erights:
> [@erights](https://github.com/erights) Can we close this issue?

Due to https://github.com/nodejs/node/pull/59456 I happily close this one. Thanks for the reminder.

PR #2934, opened 2025-08-12
fix(ses): to workaround #2908 use current console

erights:
Closes: #2908
Refs: #636 #944 #945 https://github.com/nodejs/node/pull/59456

## Description

Due to timing, reporter.js runs before lockdown replaces the system console with our causal console. Thus, when `makeReporter` makes a reporter that uses some global console, it uses the platform's original global console rather than the causal console. I don't know if this was purposeful or good enough at the time, so I don't know whether this PR is a workaround or a proper fix. In any case, it fixes the case of immediate interest by using the current global console to report rather than the original one.

### Security Considerations

@gibson042 checked below and found that the `console` property of the global of the start compartments remains mutable after lockdown. This is a security hazard, but is consistent with our stance on the privileges available in the start compartment. In any case, this hazard is not affected by this PR.

### Scaling Considerations

none

none
### Testing Considerations

Hard to write a regression test to ensure that a stack is output since there is no portable agreement on what stacks look like. But not impossible. 

Manually running the test case from #2908 , we get the correct outputs at https://github.com/endojs/endo/issues/2908#issuecomment-3180527514 . 

Running that test case from #2908 on Node 22 and 24, we find that the bug does not manifest there either. The problem is ***only*** on Node 20. I don't understand how that can be. In any case, this PR does fix the behavior across all these versions.

If someone suppresses the `console` replacement with `consoleTaming: 'unsafe'`, then this PR in its current state fails to workaround #2908 

If some code depends on reporters using the original `console` rather than the current one, that code might break or misbehave.

### Upgrade Considerations

none

### DIFF
--- a/packages/ses/src/reporting.js
+++ b/packages/ses/src/reporting.js
@@ -1,5 +1,3 @@
-/* eslint-disable @endo/no-polymorphic-call */
-// import { functionBind, globalThis } from './commons.js';
 import { globalThis } from './commons.js';
 import { assert } from './error/assert.js';
 
@@ -7,8 +5,9 @@ import { assert } from './error/assert.js';
  * @import {Reporter, GroupReporter} from './reporting-types.js'
  */
 
+/* eslint-disable @endo/no-polymorphic-call */
 /**
- * As a workaround of https://github.com/endojs/endo/issues/2908,
+ * To address https://github.com/endojs/endo/issues/2908,
  * the `consoleReporter` uses the current `console` rather
  * than the original one.
  *
@@ -36,6 +35,7 @@ const consoleReporter = {
       }
     : undefined),
 };
+/* eslint-enable @endo/no-polymorphic-call */
 
 /**
  * Creates a suitable reporter for internal errors and warnings out of the
--- a/packages/ses/src/reporting.js
+++ b/packages/ses/src/reporting.js
@@ -49,24 +49,23 @@ const mute = () => {};
  * @param {"platform" | "console" | "none"} reporting
  */
 export const chooseReporter = reporting => {
+  // // On Node.js, we send all feedback to stderr, regardless of purported level.
+  // As a workaround of https://github.com/endojs/endo/issues/2908,
+  // the `consoleError` function uses the current `console` rather than the
+  // original one.
+  // eslint-disable-next-line @endo/no-polymorphic-call
+  const consoleError = (...args) => globalThis.console.error(...args);
+
   if (reporting === 'none') {
     return makeReportPrinter(mute);
   }
   if (
     reporting === 'console' ||
     globalThis.window === globalThis ||
-    globalThis.importScripts !== undefined
+    globalThis.importScripts !== undefined ||
+    globalThis.console !== undefined
   ) {
-    return console;
-  }
-  if (globalThis.console !== undefined) {
-    // // On Node.js, we send all feedback to stderr, regardless of purported level.
-    // As a workaround of https://github.com/endojs/endo/issues/2908,
-    // the `print` function uses the current `console` rather than the
-    // original one.
-    // eslint-disable-next-line @endo/no-polymorphic-call
-    const print = (...args) => globalThis.console.error(...args);
-    return makeReportPrinter(print);
+    return makeReportPrinter(consoleError);
   }
   if (globalThis.print !== undefined) {
     return makeReportPrinter(globalThis.print);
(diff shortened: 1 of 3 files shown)


packages/ses/src/reporting.js
```diff
@@ -59,10 +60,13 @@ export const chooseReporter = reporting => {
     return console;
```
erights: If #2934 fixes the problem for reporting = "platform", then returning `console` here should cause the problem #2908 to reappear when using reporting = "console". If we decide to merge this PR into master, that inconsistency becomes an issue.

packages/ses/src/reporting.js
```diff
@@ -59,10 +60,13 @@ export const chooseReporter = reporting => {
...
   }
   if (globalThis.console !== undefined) {
-    // On Node.js, we send all feedback to stderr, regardless of purported level.
-    const console = globalThis.console;
-    const error = functionBind(console.error, console);
-    return makeReportPrinter(error);
+    // // On Node.js, we send all feedback to stderr, regardless of purported level.
+    // As a workaround of https://github.com/endojs/endo/issues/2908,
+    // the `print` function uses the current `console` rather than the
+    // original one.
```
erights: If we merge this PR into master, we should remove comments about the technique being a workaround.

packages/ses/src/reporting.js
```diff
@@ -1,4 +1,5 @@
-import { functionBind, globalThis } from './commons.js';
+// import { functionBind, globalThis } from './commons.js';
```
erights: If we merge this PR into master, delete this useless comment in that case.

erights:
https://github.com/endojs/endo/issues/945#issuecomment-985165983 may be as simple and much less fragile.

packages/ses/src/reporting.js
erights: Fixed

packages/ses/src/reporting.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @endo/no-polymorphic-call */
```
michaelfig: Can you add a comment why this lint rule is disabled for the whole file instead of just individual `// eslint-disable-next-line .. directives?

michaelfig (approved):

packages/ses/src/reporting.js
erights: Done

packages/ses/src/reporting.js
erights: Done

packages/ses/src/reporting.js
erights: Better, I moved it to disable a block around `consoleReporter`.

Done

erights:
See also https://github.com/nodejs/node/pull/59456

gibson042 (commented):
I think it might be better if the output from `chooseReporter` included a method for accepting a replacement printer, for use after the SES console is instantiated.

packages/ses/src/reporting.js
```diff
@@ -51,18 +83,13 @@ export const chooseReporter = reporting => {
...
   }
-  if (
-    reporting === 'console' ||
-    globalThis.window === globalThis ||
-    globalThis.importScripts !== undefined
-  ) {
-    return console;
-  }
   if (globalThis.console !== undefined) {
+    if (reporting === 'console') {
```
erights: add back other disjuncts

packages/ses/src/reporting.js
```diff
@@ -51,18 +83,13 @@ export const chooseReporter = reporting => {
...
+      return consoleReporter;
+    }
+    assert(reporting === 'platform');
     // On Node.js, we send all feedback to stderr, regardless of purported level.
-    const console = globalThis.console;
-    const error = functionBind(console.error, console);
-    return makeReportPrinter(error);
+    return makeReportPrinter(consoleReporter.error);
```
erights, line 101: explain cleverness

erights:
> I think it might be better if the output from `chooseReporter` included a ~method~ facet for accepting a replacement printer, for use after the SES console is instantiated.

We discussed this and the conclusion was: perhaps, but not in this PR

packages/ses/src/reporting.js
erights: Done

packages/ses/src/reporting.js
erights: Done

PR #2935, opened 2025-08-12
chore: Upgrade typescript 5.9

turadg:
Upgrade TypeScript to 5.9,
https://devblogs.microsoft.com/typescript/announcing-typescript-5-9/

Bring code into conformance and fix a testing bug it revealed


turadg:
@kriskowal this is breaking on some [AVA hackery in scaffold.js](https://github.com/endojs/endo/blob/master/packages/compartment-mapper/test/scaffold.js#L129-L144):
```
  strict › strict / loadLocation

  Test was expected to fail, but succeeded, you should stop marking the test as failing
```

kriskowal (approved):

Issue #2937, opened 2025-08-14
`this` typing for `makeExo` methods is incorrect


gibson042:
_(originally posted by @dckc in https://github.com/Agoric/agoric-sdk/pull/11748#discussion_r2276846402)_

## Describe the bug

Typing for `this` inside a `makeExo` method incorrectly lacks a "self" property.

## Steps to reproduce

<details><summary>source code</summary>

```js
  const publicFacet = zone.exo('postalSvc publicFacet', publicI, {
    /** @param {string} addr */
    async getDepositFacet(addr) {
      trace('getDepositFacet', addr);
      const depositFacet = await E(namesByAddress).lookup(
        addr,
        depositFacetKey,
      );
      trace('getDepositFacet', addr, depositFacet);
      return depositFacet;
    },
    /**
     * @param {string} addr
     * @param {Payment} pmt
     */
    async deliverPayment(addr, pmt) {
      // @ts-expect-error XXX TS2339: Property 'self' does not exist...???
      const depositFacetP = this.self.getDepositFacet(addr);
      await E(depositFacetP).receive(pmt);
    },
    //…
  });
```

</details>

> strange! the tests pass, so I know `.self` is how exos work, but now I get [lint errors in ci](https://github.com/Agoric/agoric-sdk/actions/runs/16954692547/job/48054643968?pr=11748#step:4:144)!
> 
> ```
> Error: src/postal-service.contract.js(57,34): error TS2339: Property 'self' does not exist on type '{ getDepositFacet(addr: string): Promise<any>; deliverPayment(addr: string, pmt: Payment): Promise<void>; deliverPrize(addr: string, prize: unknown, label?: string): Promise<...>; }'.
> 144
> Error: src/postal-service.contract.js(72,18): error TS2339: Property 'self' does not exist on type '{ getDepositFacet(addr: string): Promise<any>; deliverPayment(addr: string, pmt: Payment): Promise<void>; deliverPrize(addr: string, prize: unknown, label?: string): Promise<...>; }'.
> ```

## Expected behavior

`self` is known to the type system.

PR #2938, opened 2025-08-14
fix(exo): Correct the typing for `makeExo` methods

gibson042:
Fixes: #2937
agoric-sdk integration PR: https://github.com/Agoric/agoric-sdk/pull/11756

## Description

Use `ExoClassMethods` for `makeExo`, in alignment with the implementation internally leveraging `defineExoClass`.

### Security Considerations

None known.

### Scaling Considerations

n/a

Nothing specific.

### Testing Considerations

Should be covered by existing tests, but I'm open to ideas for more coverage.

I suspect that internal method calls for non-class exos are probably rare, which is why this went undetected. But if anything is affected, it will probably invalidate undesirable `@ts-expect-error` comments.

### DIFF
--- a/packages/exo/src/exo-makers.js
+++ b/packages/exo/src/exo-makers.js
@@ -32,6 +32,7 @@ const makeSelf = (proto, instanceCount) => {
   return harden(self);
 };
 
+/** @type {Record<PropertyKey, never>} */
 const emptyRecord = harden({});
 
 /**
@@ -40,8 +41,6 @@ const emptyRecord = harden({});
  * state record of the (virtual/durable) instances of the kind/exoClass
  * should be empty, and that the returned maker function should have zero
  * parameters.
- *
- * @returns {{}}
  */
 export const initEmpty = () => emptyRecord;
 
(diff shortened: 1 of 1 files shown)


packages/exo/src/exo-makers.js
```diff
@@ -218,14 +218,14 @@ export const defineExoClassKit = (
...
+ * @template {Methods} M
  * @param {string} tag
  * @param {import('@endo/patterns').InterfaceGuard<{
- *   [M in keyof T]: import('@endo/patterns').MethodGuard
+ *   [K in keyof M]: import('@endo/patterns').MethodGuard
  * }> | undefined} interfaceGuard CAVEAT: static typing does not yet support `callWhen` transformation
- * @param {T} methods
- * @param {FarClassOptions<import('./types.js').ClassContext<{},T>>} [options]
- * @returns {Guarded<T>}
+ * @param {ExoClassMethods<M, () => Record<PropertyKey, never>>} methods
```
turadg: `never` because of `initEmpty`, why not the literal?

Consider documenting in this file. 
```suggestion
 * @param {ExoClassMethods<M, () => {}>} methods
```

Either way, please include in the JSdoc that there is no state with `makeExo`. Also good to include the "singleton" term.

kriskowal (approved):

PR #2939, opened 2025-08-15
feat(patterns): `oldM` makes legacy guards for ancient runtimes

erights:
Closes: #XXXX
Refs: #1712 , https://github.com/Agoric/agoric-sdk/issues/8826 , #2038 , https://github.com/Agoric/agoric-sdk/issues/8826 , https://github.com/Agoric/agoric-sdk/pull/11748/commits/907249ab35f9046c0a45cb90b272e36abda29774 , https://github.com/Agoric/agoric-sdk/issues/9247 , https://github.com/Agoric/agoric-sdk/pull/11757

## Description

#1712 changed the representation of guards because the previous representation was unsound. Guards could not reliably be distinguished from copyRecords. 

https://github.com/Agoric/agoric-sdk/issues/8826 However, when we did so, we did not anticipate the cross-version linkage incompatibilities this created for Agoric's on chain use of endo.

#2038 fixed https://github.com/Agoric/agoric-sdk/issues/8826 by allowing old guards (as would be expressed by old contract bundles) to be linked with new guard consumers, by making those consumers tolerant of the old guards despite the unsoundness.

Recently @dckc encountered a similar error whose stack trace is
```js
2025-08-14T19:24:54.631Z SwingSet: vat: v1: CORE_EVAL failed: (RangeError#13)
2025-08-14T19:24:54.632Z SwingSet: vat: v1: RangeError#13: Expected undefined is same as Interface
2025-08-14T19:24:54.632Z SwingSet: vat: v1: RangeError: Expected (an undefined) is same as (a string)
 at makeError (/bundled-source/.../node_modules/ses/src/error/assert.js:243)
 at fail (/bundled-source/.../node_modules/ses/src/error/assert.js:357)
 at equal (/bundled-source/.../node_modules/ses/src/error/assert.js:390)
 at defendPrototype (/bundled-source/.../node_modules/@endo/exo/src/exo-tools.js:238)
 at defineKindInternal (/bundled-source/.../packages/swingset-liveslots/src/virtualObjectManager.js:957)
 at defineDurableKind (/bundled-source/.../packages/swingset-liveslots/src/virtualObjectManager.js:1155)
 at delegatePortfolioContract (.../portfolio-deploy/src/portfolio-control.core.js:47)
 at ()
 at ()
```

which we suspect is about new guards (produced by new core-eval bundles to be evaluated) being run in core-eval, which runs on an ancient version of endo, and is not easily upgraded. This is the "other way around" of https://github.com/Agoric/agoric-sdk/issues/9247 that we thought we'd never encounter in production, and thus could avoid supporting. However, this core-eval scenario above seems to make it unavoidable so long as we continue evaluating new code in an ancient core-eval.

So this PR supports that "other way around". @endo/patterns will now support an `oldM` that is like `M`, except that the guard making methods make the guards as we made them before #1712. Fortunately, current endo remains tolerant of those old guards (#2038), so using `oldM` to make old guards in new code should not break any other new code consumers. But, because of the unsoundness, you should only use `oldM` when you're in this scenario.

https://github.com/Agoric/agoric-sdk/pull/11757 is an attempt to pin down what mismatch we're actually seeing, since it may not simply be Dan's core-eval scenario. Besides, given the stack trace above, I have not been able to find a version of endo in which exo-tools.js line 238 contains anything that could call `equal` directly or indirectly. So we should not yet be confident we've diagnosed the problem that this PR would fix.

### Security Considerations

The old guards remain unsound, since they are just copyRecords. The hazard is that an old guard may be interpreted as a simple data copyRecord. Thus, we still need to eventually retire all uses and mentions of the old guards, sometime after we've retired or upgraded core-eval.

### Scaling Considerations

None

We need to explain when to use `oldM` (core eval scripts) and when not to (everything else). Programmers that never write such scripts only need to be warned away from using `oldM`.

`oldM` is born `@deprecated`, which should help.

### Testing Considerations

The existing legacy-guard-tolerance.test.js already tests our tolerance to these old guards, and to various mixtures of old and new guards. To do that testing, it needed code to make the old guard records it tests. That making code was too low level to use. This PR replaces those makers with `oldM`, which both makes the tests more readable and tests `oldM`.

But this really needs to be tested with a real chain running an ancient core-eval to see if it fixes the motivating problem. I'll need help with that.

The point. This PR enables new code that creates guards to be linked with old code (pre #1712) that consumes guards.

### Upgrade Considerations

Hopefully the old guards this PR supports never get stored to durable storage, in which case this PR does not create any new upgrade problems. This is another reason to use it only with core-eval, that has no access to durable state.


erights:
My @kriskowal simulation in my head just told me that `oldM` should be provided by a separate package export, so those that don't need it don't pay for it.

erights:
Closing as not needed. 

Because core-eval cannot be ungraded, there's no need for it to use a durable zone, as opposed to heap or virtual. As long as things are small, there's no need for virtual zone. If exos made for core-eval simply used the heap zone, then they would make the exo only using the endo code their bundled with, which is presumably the same one they use during development.

It is only the use of durable or virtual zones that causes the code to ask the liveSlots pinned by core-eval to make the exos. So shifting to the heap zone is a strictly better solution than use of the `oldM` invented by this PR.

Issue #2941, opened 2025-08-18
SES error censorship issues


naugtur:
## Describe the bug

In some circumstances I've seen SES error censorship produce an error with a falsy value as message and a stack containing only references to SES concole implementation. Other times it corrupts the error message.

All examples I found to date seem related to `errorTrapping` implementation

I'll collect ways to reproduce it in this issue. 

## 1. broken functionality

1.1. when opening a html file via `file://` protocol with invalid syntax in a script, error gets censored to nothing. `errorTrapping: "none"` prevents the issue.

[test1.tar.gz](https://github.com/user-attachments/files/21831816/test1.tar.gz)

1.2. When a multiline message is thrown in Node.js, SES will print all lines of the message, except first, twice.
(does not happen in the browser)
[test2.tar.gz](https://github.com/user-attachments/files/21832043/test2.tar.gz)

## 2. bad UX

2.1. The `Error("",{ cause: e })` output can be considered correct, but differs between `errorTrapping` default, `errorTrapping: "none"` and no lockdown at all.

[test3.tar.gz](https://github.com/user-attachments/files/21832288/test3.tar.gz)

2.2. console output of unhandled rejections is not compatible with browsers

```
async function que() {
  throw Error('karramba')
}
que()
```

2.2.1. FireFox: the original uncaught error in promise is reported first, the SES version after a few seconds

<img width="691" height="478" alt="Image" src="https://github.com/user-attachments/assets/1a8fe378-378b-478d-be7a-cf61e0f214da" />

<img width="690" height="365" alt="Image" src="https://github.com/user-attachments/assets/597ae068-671a-4b13-a9d2-e0f1b19b9936" />

 2.2.2. Both FireFox and Chromium: 3 console lines each are an error with a misleading stacktrace.

<img width="750" height="225" alt="Image" src="https://github.com/user-attachments/assets/3c6978a5-daa9-45e8-bdff-ddfb825883b5" />

<img width="739" height="452" alt="Image" src="https://github.com/user-attachments/assets/b7d20b34-ee6e-4423-973b-a62dc420a873" />

PR #2943, opened 2025-08-18
fix(ses): errorTrapping loses errors

erights:

Closes: #XXXX
Refs: #2941 

## Description

For #2941 test1, reporting the event when there is no event.error, in the browser, after clicking on the outline-expansion triangle, looks like

image

which is still much less informative that the platform's default error logging behavior, but probably the best we can do given the platform's support for user-level error trapping.

### Security Considerations

We were not previously exposing the event itself. Here we're only exposing it to the console, under our standard assumption that only very privileged programs or users can read the console output. So we seem to be safe under that assumption.

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations


erights:
Closing in favor of #2945

PR #2944, opened 2025-08-20
test(ses): repro 2941 errorTrapping bugs

erights:
Just a baseline repro, so future PRs that fix these bugs can illustrate what's different.

Closes: #XXXX
Refs: #2941 #2945 

## Description

Just a baseline repro of #2941 , so a future PR (such as #2945) that fix this bug can illustrate what's different.

### Security Considerations

Adds only test cases without any changes to the system being tested, so none.

### Scaling Considerations

none
### Documentation Considerations

none
### Testing Considerations

The point. The added test cases are a baseline repro to be fixed in an upcoming PR #2945 
### Compatibility Considerations

none
### Upgrade Considerations

none

PR #2945, opened 2025-08-20
fix(ses): fix 2943 error trapping issues

erights:
Staged on #2944

Closes: #2941
Refs: #XXXX

## Description

In browsers, if `errorTrapping:` is set to anything other than `'none'`, then `globalWindow.addEventListener('error', event => {...});`
is used to log a diagnostic that we wish contain the reason from an unhandled rejection, where the reason should be `event.error`.

However, in at least some browsers, `event.error` is absent, `null`, or `undefined`. We don't care which. In that case, all we've got is the event, so we log that instead.

### Security Considerations

`errorTrapping` only provides a fixed menu, with the repair doing the privileged registration of a closure. This mechanism does not give post-lockdown user code any non-logging abilities it does not otherwise have. And the logging abilities are consistent with the log being write-only. 

With this PR, an event object does get logged, which may contain info we would have liked to redact for normal user code. But this extra info only appears in the log, from which we never redact info anyway.

### Scaling Considerations

none

This PR does add an explanation to SES_UNCAUGHT_EXCEPTION.md, which is where it is most likely to be discovered by those who encounter this.

### Testing Considerations

The relevant test were already introduced by #2944 , so this PR could illustrate the fix by the diff between it and #2944.

none. It is hard to imagine anything was depending on the diagnostic being inadvertently empty.

### Upgrade Considerations

none


packages/ses/error-codes/SES_UNCAUGHT_EXCEPTION.md
```diff
@@ -6,6 +6,8 @@ This behavior is configured by calling `lockdown` with the [errorTrapping][] opt
 
 Otherwise, when thrown exceptions have no `catch` handler, the SES shim will print `SES_UNCAUGHT_EXCEPTION: exception...` to the error console.  In that case, `exception...` is the thrown exception, augmented with causal error information when possible.
 
+On some platforms, the default errorTrapping behavior reports the error, but the platform's user trapping mechanism reports only an event with no error. (See https://github.com/endojs/endo/issues/2941 test1.) In that case, non-default `errorTrapping` reports the event about the error, rather than the error itself. This often has much less relevant information.
+
```
erights: rewrite

erights: Done

erights:
This is now Ready for Review

kriskowal (approved):

PR #2946, opened 2025-08-20
feat: Add immutable `ArrayBuffer` shim types

rekmarks:
## Description

Adds types for the immutable `ArrayBuffer` shim via the `./shim.js` export path of `@endo/immutable-arraybuffer`. Also imports and thereby exposes these types via the `ses` shim, which entrains the immutable `ArrayBuffer` shim.

### Testing Considerations

Added type tests asserting that the expected properties have been added to `ArrayBuffer` in `@endo/immutable-arraybuffer` and `ses`.


kriskowal (approved):
Huzzah.

PR #2947, opened 2025-08-30
chore: Lint for kebab-case

kriskowal:
I have a pretty strong preference for not having to guess about case conventions in file names. This will make exceptions to the rule evident in CI and in PR diffs. The code that checks is not exact, but close enough for now. A better checker would require fewer explicit exemptions and would disallow many cases where `_` is tolerated currently.


erights (changes requested):
> Could we exempt whole directories, so we don't need to exempt test262 tests individually? Since they are not under our control anyway?

Looking at the PR again, I'm escalating this to a "request changes". Introducing a 9,775 line source file that actually conveys only a tiny bit of information is bad. If this is automatically generated, as I'm sure at least partly it is, then the thing to review is the auto-generation code, not its impossible-to-review output. Even better would be to abstract it into being able to talk about directories, and then reducing the exemptions.txt file down to something manually reviewable.

PR #2950, opened 2025-09-10
Update README.md

danfinlay:
Added an explanation to first time repository setup that would've saved me a little research as a non-lerna-native.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,18 @@ security-sensitive bugs privately.
 For non-security bugs, please use the [regular Issues
 page](https://github.com/Agoric/SES-shim/issues).
 
+## Contributing
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -104,6 +104,18 @@ security-sensitive bugs privately.
 For non-security bugs, please use the [regular Issues
 page](https://github.com/Agoric/SES-shim/issues).
 
+## Contributing
+
+This repository uses lerna as a monorepo, to allow the included packages to depend on each other.
```
kriskowal, line 109: I think the cross-package linking is a `yarn` feature and the tooling for publishing is facilitated by `lerna`, but the boundary is vague in my mind.

kriskowal: And maybe link `CONTRIBUTING.md`.

PR #2952, opened 2025-09-12
fix(ses): fix #2951 stronger sniffing for v8

### LINKED ISSUE #2951: ses v8 sniff test is too weak to exclude very-non-v8 platforms
At 
https://github.com/endojs/endo/blob/5885418223025cb6c8d67edfe953a2175be9ce2a/packages/ses/src/error/tame-error-constructor.js#L41-L42
we see that the ses error taming detects if it is on v8 by sniffing the existing of only `captureStackTrace`. That was accurate at the time, but since then other platforms have implemented `captureStackTrace` without `prepareStackTrace`, and so will be tamed as if it were v8. But if they don't implement `prepareStackTrace`, then we need to tame them as not-v8.

We cannot simply fix the sniff test to check for the existence of `prepareStackTrace` too, because it does not exist on the initial state of the v8 error constructor. We need some more elaborate sniff test that might involve setting a `prepareStackTrace` and seeing if it gets called.

This issue is now urgent, not simply because other platforms have implemented `captureStackTrace`, but because https://github.com/tc39/proposal-error-capturestacktrace is a live standards track proposal that looks good.

erights:
Closes: #2951 
Refs: #1798 , https://github.com/tc39/proposal-error-capturestacktrace , https://github.com/tc39/proposal-error-stacks , https://github.com/tc39/proposal-error-stack-accessor

## Description

(Description mostly from Copilot's overview below, which is quite good!)

This PR's error taming implements stronger platform detection for v8, addressing issue #2951. The change improves the reliability of v8 detection by testing the `prepareStackTrace` mechanism, in addition to checking for the presence of `captureStackTrace`.

This is needed because some non-v8 platforms have implemented `captureStackTrace` without `prepareStackTrace`. The https://github.com/tc39/proposal-error-capturestacktrace proposal would make it standard, eventually causing all conformant JS engines to implement it.

A verbally reported bug motivated this, even though that bug itself may be a false alarm. The report raised the possibility that our error-taming of treating the Hermes platform as if it was v8, and then misbehaving because it does not act like v8. Whether or not this actually happens on Hermes, it is plausible and would happen elsewhere with ses prior to this PR.

### Security Considerations

If non-v8 engines start implementing the `prepareStackTrace` mechanism, we may need to tighten the sniff test once again. ***Such is the nature of sniff tests***. Since we do not yet know the specifics of such possible additions, it does not yet make sense to tighten the sniff test, unless I'm missing some we should sniff instead.

During such a transition period, it is possible that the falsly-triggered v8 error taming could misbehave, including in ways that may be exploitable.

### Scaling Considerations

none

This is a bug fix and does not need documentation beyond the issue, this PR, and the inline comments added by this PR.

### Testing Considerations

- [ ] I did not add automated tests, because I am unclear on how automated tests would test this. ***Reviewers***: suggestions welcome.

- [ ] What I did do is manually step through the logic in the vscode debugger while running on Node. The sniff test code worked as expected there. If we do rely on such manual testing, I should at least test on some non-v8 platform that does implement `captureStackTrace`.

See "Security Considerations" above.

### Upgrade Considerations

none

### DIFF
--- a/packages/ses/src/error/tame-error-constructor.js
+++ b/packages/ses/src/error/tame-error-constructor.js
@@ -3,6 +3,8 @@ import {
   apply,
   construct,
   defineProperties,
+  getPrototypeOf,
+  uncurryThis,
   setPrototypeOf,
   getOwnPropertyDescriptor,
   defineProperty,
@@ -43,6 +45,7 @@ export default function tameErrorConstructor(
     prepareStackTrace: originalPrepareStackTrace,
   } = FERAL_ERROR;
   let platform = 'unknown';
+  let callSiteToStringFallback;
   if (typeof originalCaptureStackTrace === 'function') {
     // we might be on v8
     if (typeof originalPrepareStackTrace === 'function') {
@@ -65,6 +68,27 @@ export default function tameErrorConstructor(
             // error stack logic is close enough that we can treat it
             // like v8.
             platform = 'v8';
+
+            if (`${sst[0]}` === '[object CallSite]') {
+              const csProto = getPrototypeOf(sst[0]);
+
+              const [
+                getFunctionName,
+                getMethodName,
+                getFileName,
+                getLineNumber,
+                getColumnNumber,
+              ] = [
+                uncurryThis(csProto.getFunctionName),
+                uncurryThis(csProto.getMethodName),
+                uncurryThis(csProto.getFileName),
+                uncurryThis(csProto.getLineNumber),
+                uncurryThis(csProto.getColumnNumber),
+              ];
+
+              callSiteToStringFallback = callSite =>
+                `${getFunctionName(callSite) || getMethodName(callSite)} (${getFileName(callSite)}:${getLineNumber(callSite)}:${getColumnNumber(callSite)})`;
+            }
           }
         };
         const sacrificialError = new FERAL_ERROR('just for testing');
@@ -247,6 +271,7 @@ export default function tameErrorConstructor(
       InitialError,
       errorTaming,
       stackFiltering,
+      callSiteToStringFallback,
     );
   } else if (errorTaming === 'unsafe' || errorTaming === 'unsafe-debug') {
     // v8 has too much magic around their 'stack' own property for it to
--- a/packages/ses/src/error/tame-v8-error-constructor.js
+++ b/packages/ses/src/error/tame-v8-error-constructor.js
@@ -200,6 +200,7 @@ export const tameV8ErrorConstructor = (
   InitialError,
   errorTaming,
   stackFiltering,
+  callSiteToStringFallback,
 ) => {
   if (errorTaming === 'unsafe-debug') {
     throw TypeError(
@@ -232,6 +233,10 @@ export const tameV8ErrorConstructor = (
 
   const callSiteStringifier = callSite => {
     let callSiteString = `${callSite}`;
+    if (callSiteToStringFallback) {
+      callSiteString = callSiteToStringFallback(callSite);
+    }
+
     if (shortenPaths) {
       callSiteString = shortenCallSiteString(callSiteString);
     }


packages/ses/src/error/tame-error-constructor.js
```diff
@@ -37,9 +38,42 @@ export default function tameErrorConstructor(
...
+      // This case should not occur on v8 or any other platform.
+      // But if it does, we assume we're on v8, or a platform whose
+      // error stack logic is close enough that we can treat it
+      // like v8.
+      platform = 'v8';
+    } else {
+      try {
+        FERAL_ERROR.prepareStackTrace = (error, sst) => {
+          // eslint-disable-next-line no-use-before-define
+          if (error === sacrificialError && isArray(sst)) {
```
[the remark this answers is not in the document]
erights: We know this case cannot have TDZ issues. I would rather not even construct the `sacrificialError` until the test `prepareStackTrace` has been installed on `FERAL_ERROR`. Resolving.

packages/ses/src/error/tame-error-constructor.js
```diff
@@ -37,9 +38,42 @@ export default function tameErrorConstructor(
...
+            // `prepareStackTrace` property. To test whether the current
+            // platform implements it,
+            // we must set our own and then see if `error.stack` trigger it.
+            // If so, then we assume we're on v8, or a platform whose
+            // error stack logic is close enough that we can treat it
+            // like v8.
+            platform = 'v8';
+          }
+        };
+        const sacrificialError = new FERAL_ERROR('just for testing');
```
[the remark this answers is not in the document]
erights: Good idea! With a slight reprase, done.

packages/ses/scripts/hermes-test.js
```diff
@@ -0,0 +1,159 @@
...
+
+  const testContent = readFileSync(testFile, 'utf8');
+  // Combine the code in a way that line numbers are preserved and therefore helpful
+  const combined = `${testContent}
+;;;;
+${hermesPrefixSES}
+;;;;
+
+var TEST;
+function test(name, fn) {
```
naugtur, line 96: Thanks to the test going in a function I can start the concatenation with it, so if errors are thrown, the line numbers will be correct. SES gets concatenated after but the test runs synchronously after SES still.

packages/ses/src/error/tame-error-constructor.js
```diff
@@ -37,9 +40,66 @@ export default function tameErrorConstructor(
...
+            // of the original error constructor, `FERAL_ERROR`, has no
+            // `prepareStackTrace` property. To test whether the current
+            // platform implements it,
+            // we must set our own and then see if `error.stack` triggers it.
+            // If so, then we assume we're on v8, or a platform whose
+            // error stack logic is close enough that we can treat it
+            // like v8.
+            platform = 'v8';
+
+            if (`${sst[0]}` === '[object CallSite]') {
```
naugtur, line 72: Here we detect Hermes providing a broken toString

packages/ses/src/error/tame-error-constructor.js
```diff
@@ -37,9 +40,66 @@ export default function tameErrorConstructor(
...
+                getColumnNumber,
+              ] = [
+                uncurryThis(csProto.getFunctionName),
+                uncurryThis(csProto.getMethodName),
+                uncurryThis(csProto.getFileName),
+                uncurryThis(csProto.getLineNumber),
+                uncurryThis(csProto.getColumnNumber),
+              ];
+
+              callSiteToStringFallback = callSite =>
```
naugtur, line 89: Instead of creating it and passing around, we could instead repair the original, but I'm afraid someone else might be doing similar feature detection

packages/ses/test/hermes/error-taming-unsafe.js
```diff
@@ -0,0 +1,25 @@
+// tip: use this to trip up platform detection
+// Error.captureStackTrace = null
+
+/* global test */
+test('error taming unsafe', () => {
```
naugtur, line 5: This is the one that fails without my toStringFallback

packages/ses/test/hermes/known-issue-callsite.js
```diff
@@ -0,0 +1,92 @@
+/* global test */
+test('known issue: CallSite implementation', () => {
```
naugtur, line 2: full test documenting broken callsite implementation

packages/ses/test/hermes/known-issue-promise.js
```diff
@@ -0,0 +1,36 @@
+/* global test */
+/* global repairIntrinsics */
+test('knonw issue: Hermes Promise is non-standard', () => {
```
naugtur, line 3: This test documents the broken promise implementation and verifies that the fix we're applying in lavamoat still works.

Are we interested in adding the fix to ses-hermes? I think it'd make sense - Promise constructors will throw often if not fixed.

erights:
> Get ready, this will hurt.

Thanks for the warning. It did indeed hurt :/

packages/ses/src/error/tame-error-constructor.js
```diff
@@ -37,9 +40,66 @@ export default function tameErrorConstructor(
...
+  let platform = 'unknown';
+  let callSiteToStringFallback;
+  if (typeof originalCaptureStackTrace === 'function') {
+    // we might be on v8
+    if (typeof originalPrepareStackTrace === 'function') {
+      // This case should not occur on v8 or any other platform.
+      // But if it does, we assume we're on v8, or a platform whose
+      // error stack logic is close enough that we can treat it
+      // like v8.
+      platform = 'v8';
```
mhofman, line 56: I believe it does happen in Node.js according to the realm init logic I'm seeing: https://github.com/nodejs/node/blob/b8870c4a61f9c4c4490e43472e98655b00055359/lib/internal/bootstrap/realm.js#L444-L470

FYI that logic shows that it's not v8 that calls `Error.prepareStackTrace` but the embedder that sets a callback (in this case `setPrepareStackTraceCallback`), which [itself does the forwarding](https://github.com/nodejs/node/blob/b8870c4a61f9c4c4490e43472e98655b00055359/lib/internal/errors.js#L116-L156) to `Error.prepareStackTrace` (or some other fallback)

erights:
> potential fallback for repareStackTrace CallSite toString method being broken

Btw, given the topic, I first read this as "repairStackTrace" and became unreasonably hopeful for a moment ;) .

mhofman (commented):
Given the new modular design of `prepareStackTrace` in Node.js introduced in https://github.com/nodejs/node/pull/50827, I think we should at least restore (or not touch?) the original `prepareStackTrace` if the `errorTaming` mode is `'unsafe'`.

Also, since this is so related, maybe we could attempt to leverage the original `prepareStackTrace` to build the stack string (see https://github.com/endojs/endo/issues/1798), but happy to have this as a follow up PR.

(@mhofman , restoring your original. Somehow when I thought I was replying to it, I accidentally revised it instead. In any case, I found a better place for the reply, above.)

packages/ses/test/hermes/known-issue-promise.js
erights: Where can I find this fix?

packages/ses/src/error/tame-error-constructor.js
erights: Huh. Testing just now, it was added sometime between Node v18.19.0 exclusive and Node v20.18.3 inclusive.

packages/ses/test/hermes/known-issue-promise.js
naugtur: The fix is demonstrated here (capturing all truthy underscore-prefixed fields from Promise and preserving them till after repair to bring them back) 
It's used in `@lavamoat/react-native-lockdown` but I think if we claim SES works under Hermes, we should consider adding this exception to the hermes build by default. 


Another awful detail - the field that is exposed and must exist or the Promie implementation falls apart is containing a no-op function that the implementation reuses in a few places, which could have been a local scope function. Other fields, null by default, allow instrumenting all promises, so good riddance.

erights:
@naugtur , ping?

erights:
@naugtur , even though I handed this off to you, I took the liberty of rebasing on master

PR #2954, opened 2025-09-16
docs(marshal): expand smallcaps readability invariants

erights:

Closes: #XXXX
Refs: #XXXX

## Description

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations


erights:
@kriskowal , I only changed one *.md file and all my CI tests are read. Any again on retry. Help!

packages/marshal/docs/smallcaps-cheatsheet.md
```diff
@@ -11,7 +11,7 @@ An example-based summary of the Smallcaps encoding on the OCapN [Abstract Syntax
 | number           | Float64       | `Infinity`<br>`-Infinity`<br>`NaN`<br>`-0`<br>`7.1` | `"#Infinity"`<br>`"#-Infinity"`<br>`"#NaN"`<br>`"#-0"` // unimplemented<br>`7.1` |
 | string           | String        | `'#foo'`<br>`'foo'`   | `"!#foo"` // special strings<br>`"foo"` // other strings |
 | byteArray        | ByteArray     | `buf.toImmutable()`   | // undecided & unimplemented |
-| selector         | Selector      | `makeSelector('foo')` | `"%foo"` // converting from symbol |
+| passable symbols | Symbol        | `passableSymbolForName('foo')` | `"%foo"` // in transition |
 | copyArray        | List          | `[a,b]`               | `[<a>,<b>]`          |
 | copyRecord       | Struct        | `{x:a,y:b}`           | `{<x>:<a>,<y>:<b>}`  |
```
mhofman: what pass-styles are allowed for `x` and `y`? is it only string and thus the `<x>` and `<y>` are simply the encoded strings ?

packages/marshal/docs/smallcaps-cheatsheet.md
```diff
@@ -20,14 +20,20 @@ An example-based summary of the Smallcaps encoding on the OCapN [Abstract Syntax
...
 * The ByteArray encoding is not yet designed or implemented.
 
-Every JSON encoding with no special strings anywhere decodes to itself.
+## Readability Invariants
+
+For every JSON encoding with no special strings, the JSON and smallcaps decodings are the same.
+
+For every value with no special strings that round trips through JSON, the JSON and smallcaps encodings are the same.
+
+In other words, for these simple values, ***you can ignore the differences between smallcaps and JSON***.
```
mhofman, line 39: I still wish we defined smallcaps as a transformation of passable data into JS objects such that a plain `JSON.stringify` produces the string encoding described above, and that parsing that string with `JSON.parse` is guaranteed to produce a value that is "deep-equal" to the output of the smallcaps transformation.

packages/marshal/docs/smallcaps-cheatsheet.md
erights: Yes. Resolving.

packages/marshal/docs/smallcaps-cheatsheet.md
erights: Easy enough to extract that from the internals of @endo/marshal. I agree it would be good to do. But ok not to do in this PR?

packages/marshal/docs/smallcaps-cheatsheet.md
```diff
@@ -20,14 +20,20 @@ An example-based summary of the Smallcaps encoding on the OCapN [Abstract Syntax
...
 * Errors can also carry an optional `errorId` string property.
 * We expect to expand the optional error properties over time.
 * The ByteArray encoding is not yet designed or implemented.
 
-Every JSON encoding with no special strings anywhere decodes to itself.
+## Readability Invariants
+
+For every JSON encoding with no special strings, the JSON and smallcaps decodings are the same.
+
+For every value with no special strings that round trips through JSON, the JSON and smallcaps encodings are the same.
```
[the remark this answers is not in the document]
erights: The two cases are distinguished based on one starting from an encoding and the other starting from the value. Your "parallel" structure is so parallel it describes only one case, redundantly. Resolving as invalid.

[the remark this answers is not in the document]
erights: The two cases are distinguished based on one starting from an encoding and the other starting from the value. Your "parallel" structure is so parallel it describes only one case, redundantly. Resolving as invalid.

packages/marshal/docs/smallcaps-cheatsheet.md
mhofman: of course!

packages/marshal/docs/smallcaps-cheatsheet.md
mhofman: Not in scope for this PR, but it would be nice to clarify that in this doc. I ended up finding the information in the ocapn spec, but besides that, most of the doc is pretty standalone.

packages/marshal/docs/smallcaps-cheatsheet.md
```diff
@@ -20,14 +20,20 @@ An example-based summary of the Smallcaps encoding on the OCapN [Abstract Syntax
...
 * To be passable, arrays, records, and errors must also be hardened.
 * Structs [can only have string-named properties](https://github.com/endojs/endo/blob/master/packages/pass-style/doc/copyRecord-guarantees.md).
 * Errors can also carry an optional `errorId` string property.
 * We expect to expand the optional error properties over time.
 * The ByteArray encoding is not yet designed or implemented.
 
-Every JSON encoding with no special strings anywhere decodes to itself.
+## Readability Invariants
+
+For every JSON encoding with no special strings, the JSON and smallcaps decodings are the same.
```
mhofman, line 35: I think copilot had a point here. What does it mean for a "JSON encoding" to be with a special string? Special string is in the realm of the value to be encoded (or that was decoded).

packages/marshal/docs/smallcaps-cheatsheet.md
mhofman: I thought we established that the ocapn specification does not define a canonical encoding. Isn't it allowed for the encoding to be different yet equivalent, and as such a small caps implementation is not required to follow the exact semantics of json encoding (e.g. in struct field order). Furthermore there are many ways to encode strings, and there are indentation settings in `JSON.serialize` that would similarly produce different encodings.

mhofman (commented):
An actual review of the invariants as requested

packages/marshal/docs/smallcaps-cheatsheet.md
erights: Unresolving this one because it should be simple enough to do. I'll consider it as a drive-by.

packages/marshal/docs/smallcaps-cheatsheet.md
erights: I don't understand. I'm not referring to any concrete encoding mandated by ocapn. Indeed, ocapn itself is likely to mandate something completely different. I am only speaking here about smallcaps vs JSON, and about the smallcaps concrete encoding of the ocapn abstract-syntax/data-model.

packages/marshal/docs/smallcaps-cheatsheet.md
erights: In any JSON encoding, there is a concrete lexical element for "string" that is always quoted. https://www.json.org/json-en.html

packages/marshal/docs/smallcaps-cheatsheet.md
erights: Does this question make sense?

packages/marshal/docs/smallcaps-cheatsheet.md
erights: @mhofman , Does my reply make sense?

packages/marshal/docs/smallcaps-cheatsheet.md
erights: Addressed by an additional bullet in the list below. Done.

erights:
PTAL, thanks.

packages/marshal/docs/smallcaps-cheatsheet.md
gibson042: Suggestion: demonstrate key encoding and ordering by example.
```suggestion
| copyRecord       | Struct        | `{a}`<br>`{a,"#a":a,b}` | `{"a":<a>}` // only string keys<br>`{"!#a":<a>,"a":<a>,"b":<b>}` // keys sorted |
```

packages/marshal/docs/smallcaps-cheatsheet.md
```diff
@@ -20,14 +20,21 @@ An example-based summary of the Smallcaps encoding on the OCapN [Abstract Syntax
 | error            | Error         | `TypeError(msg)`      | `{"#error":<msg>,"name":"TypeError"}` |
 
 * The `-0` encoding is defined as above, but not yet implemented in JS.
-* In JS, selectors in transition from symbols to their own representation
+* In JS, passable symbols are in transition from JavaScript symbols to their own representation
 * The number after `"$"` or `"&"` is an index into a separate slots array.
```
gibson042:
```suggestion
* The number after `"$"` or `"&"` for remotable/Target or promise/Promise is an index into a separate slots array.
```

packages/marshal/docs/smallcaps-cheatsheet.md
```diff
@@ -20,14 +20,21 @@ An example-based summary of the Smallcaps encoding on the OCapN [Abstract Syntax
...
-* Special strings begin with any of the `!"#$%&'()*+,-` characters.
+* ***Special strings*** begin with any of the `!"#$%&'()*+,-` characters.
 * `<expr>` is nested encoding of `expr`.
+* Since `x` and `y` can only be strings, `<x>` and `<y>` are the nested encodings of those strings.
```
gibson042: This isn't necessary if we refactor copyRecord as suggested above.
```suggestion
```

packages/marshal/docs/smallcaps-cheatsheet.md
```diff
@@ -20,14 +20,21 @@ An example-based summary of the Smallcaps encoding on the OCapN [Abstract Syntax
...
 * Errors can also carry an optional `errorId` string property.
 * We expect to expand the optional error properties over time.
 * The ByteArray encoding is not yet designed or implemented.
 
-Every JSON encoding with no special strings anywhere decodes to itself.
+## Readability Invariants
+
+For every JSON encoding with no special strings, the JSON and smallcaps decodings are the same.
+
+For every value with no special strings that round trips through JSON, the JSON and smallcaps encodings are the same.
```
gibson042: I'm not sure how to interpret this claim. Does this rephrasing capture its intent?
```suggestion
If a value `v` round-trips through `JSON.parse(JSON.stringify(v))` and contains no special strings, then the smallcaps encoding of `v` is identical to `JSON.stringify(v)`.
```
(where "_round-trips_" is understood in a deep-equal sense and excludes -0 because `JSON.stringify(-0)` is "0", and "_contains no special strings_" includes top-level values, deep values, and both top-level and deep property names)

gibson042 (approved):

packages/marshal/docs/smallcaps-cheatsheet.md
erights: Done

packages/marshal/docs/smallcaps-cheatsheet.md
erights: Done

packages/marshal/docs/smallcaps-cheatsheet.md
erights: Gone

packages/marshal/docs/smallcaps-cheatsheet.md
erights: Done.

packages/marshal/docs/smallcaps-cheatsheet.md
erights: @mhofman I took @gibson042 's suggestion below. Resolving.

packages/marshal/docs/smallcaps-cheatsheet.md
erights: @mhofman I took @gibson042 's suggestion below. Resolving.

PR #2955, opened 2025-09-16
refactor(patterns): Deprecate `sloppy` interface guard option

rekmarks:
Refs: #1831, https://github.com/Agoric/agoric-sdk/pull/11931

## Description

We wish to visibly deprecate the `sloppy` option of `@endo/patterns` interface guards. To avoid an explosion of `@typedef` blocks, we migrate the `@endo/patterns` types to TypeScript and then add the minimally required `@deprecated` tags.

### Security Considerations

N/A

### Scaling Considerations

N/A

N/A

### Testing Considerations

This modifies the implementation of the `@endo/patterns` types, but should not modify them. Since the number of lines changed is larged, we should attempt to convince ourselves that the types are in fact the same.

N/A

### Upgrade Considerations

Deprecates but does not remove the `sloppy` option. `NEWS.md` is updated.


tsconfig.build.json
```diff
@@ -4,6 +4,7 @@
         "./tsconfig-build-options.json"
     ],
     "compilerOptions": {
+        "allowImportingTsExtensions": true,
```
rekmarks: I'm not sure if this is controversial @turadg. I enabled it because ESLint was complaining about unresolved imports in `patterns` following the TypeScript migration.

turadg: I'm on my phone but without looking over the whole PR I wonder if the imports would work with .js instead. Usually TS can figure out the correct module. We'd like to avoid this option in this repo because it makes it easier to accidentally ship modules that have TS syntax.

rekmarks: Based on the number of CI failures due to this, I think I will have to walk it back anyway.

mhofman: I'm not sure what approach endo follows, but [this](https://github.com/Agoric/agoric-sdk/blob/master/docs/typescript.md#ts-modules) is what has been working for us on the agoric-sdk side to author types in `.ts` without the baggage that comes with a mandatory build step during authorship (build step is still required for publishing)

rekmarks: Never mind!

rekmarks: Ah, your comments didn't show up until a refresh. tl;dr: I walked it back to `.js`. Not sure why I got the ESLint errors in the first place but they're gone now.

packages/patterns/package.json
```diff
@@ -18,7 +18,10 @@
   "main": "./index.js",
   "module": "./index.js",
   "exports": {
-    ".": "./index.js",
+    ".": {
+      "types": "./types.d.ts",
+      "default": "./index.js"
+    },
```
mhofman: I still prefer to have an `export * from './types.js'` in `index.js` (with `types.js` being a stub `export {}`)

packages/patterns/types.d.ts
```diff
@@ -0,0 +1,2 @@
+export * from './src/types.js';
+export * from './index.js';
```
mhofman: This should be removed if switching to `index.js` re-exporting `types.js`

packages/patterns/test/types.test-d.ts
```diff
@@ -2,7 +2,7 @@ import type { Passable } from '@endo/pass-style';
 import { expectNotType, expectType } from 'tsd';
 import { isKey, isScalarKey } from '../src/keys/checkKey.js';
 import { M } from '../src/patterns/patternMatchers.js';
-import type { Key, ScalarKey } from '../src/types.js';
+import type { Key, ScalarKey } from '../src/types.ts';
```
mhofman: FYI I prefer types tests to import from the entrypoint like a dependency would.

packages/patterns/index.js
```diff
@@ -75,6 +75,9 @@ export {
   getInterfaceMethodKeys,
 } from './src/patterns/getGuardPayloads.js';
 
+// eslint-disable-next-line import/export
+export * from './src/types.js';
```
mhofman:
```suggestion
export * from './types.js';
```

packages/patterns/src/types.js
mhofman, line 1: This file should be removed.

turadg (approved):
LGTM, thanks

packages/patterns/test/types.test-d.ts
```diff
@@ -2,7 +2,7 @@ import type { Passable } from '@endo/pass-style';
 import { expectNotType, expectType } from 'tsd';
 import { isKey, isScalarKey } from '../src/keys/checkKey.js';
 import { M } from '../src/patterns/patternMatchers.js';
```
mhofman: Nit: not strictly in scope of this PR, but I believe you can import the runtime exports of the package from the root of the package too. Also I think it's legal to import from `@endo/patterns` to go through the package.json exports resolution mechanism, verifying that everything is correctly exported. (in this case the `package.json` is pretty straightforward tho)

mhofman (approved):
I have not reviewed the actual change of the types.js -> types.ts.

PR #2958, opened 2025-09-17
feat(ocapn-noise): Initial draft for OCapN Noise Protocol cryptography

kriskowal:
This change covers an implementation of a Noise Protocol suitable for an OCapN netlayer that authenticates and preserves confidentiality for a session between a pair of peers, including the mechanisms necessary for public and private keys.

The selected protocol variant is XX-x25519-ChaChaPoly1309-Blake2 with ed25519 key signing, which presumes that the initiator knows the ed25519 signature verification key of the responder. In OCapN, that key is expressed by the locator. The protocol also allows the responder to discover the public key of the initiator, which is useful for resolving crossed hellos and preserving the identity of targets throughout an OCapN network implementing 3-party-handoff. The ed25519 keys are conveniently small and easy to capture in an OCapN locator. Elliptic curves are a heck of a thing.

The intention for this netlayer is for it to support multiple transport protocols, message delimiters, and potentially even multiple concrete encodings. As such, this implementation is incomplete because it doesn’t do encoding negotiation in the three-way-handshake yet, but the scaffold is in place.

This change also leaves the high level API unwritten. My intention is for the OCapN package to accept a netlayer implementation and that interface is not yet implemented here. I am assuming that we will implement a listener as an async iterator over accepted, negotiated connections with their public key, where each connection is itself a pair of async iterators over Uint8Array ranges. There is some room for avoiding buffer copies out of the WasmMemory and using the iterator protocol to release messages.

This implementation assumes that the overhead of a small Wasm VM for each connection will be acceptable, so does not attempt to manage allocation and tracking for multiple concurrent sessions, leaving that concern to JavaScript. The intent is to later introduce separate exports for Node.js and web implementations, with a default implementation that throws "not implemented", for environments where the mechanism for importing the WebAssembly implementation is less certain. In time, I hope we can use `import source` across multiple platforms.


rust/ocapn_noise/build.sh
rekmarks, line 1: This lil' guy could use a shebang, and needs a `chmod +x`.

rust/ocapn_noise/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+                assert!(hs.get_next_message_overhead() == 96);
+                match hs.write_message(&[], &mut BUFFER[64..64 + 96]) {
+                    Err(_) => {
+                        return 1;
+                    }
+                    Ok(()) => (),
+                };
+                return 0;
+            }
+        }
```
rekmarks: We replace the `match` here with the more succinct `.unwrap()`. Ditto for other instances where we set and then immediately use `HS`:
```suggestion
        let hs = HS.as_mut().unwrap();
        assert!(hs.get_next_message_overhead() == 96);
        match hs.write_message(&[], &mut BUFFER[64..64 + 96]) {
            Err(_) => {
                return 1;
            }
            Ok(()) => (),
        };
        return 0;
```

rust/ocapn_noise/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+    fn genkey_callback(private: *const u8, public: *const u8);
+    fn buffer_callback(buffer: *const u8);
+}
+
+unsafe extern "Rust" {
+    fn __getrandom_v03_custom(
+        _dest: *mut u8,
+        _len: usize,
+    ) -> Result<(), getrandom::Error>;
+}
```
rekmarks, line 86: I imagine there's a good reason for this, but why are we delegating the randomness fetching to JS?

rust/ocapn_noise/Cargo.toml
rekmarks, line 1: By also adding a `Cargo.toml` with the following content to the repository root, my tooling can learn of the existence of this file and vastly improve my (and perhaps your) development experience:

```toml
[workspace]
members = ["rust/ocapn_noise"]
resolver = "2"

```

rekmarks (commented):
Neat!

rust/ocapn_noise/src/lib.rs
kriskowal: The WASM VM is a blank slate (in the ocap sense!) and has to get entropy from outside. This is usually rendered invisible because folks usually use a higher level binding mechanism that does this automatically for you. But, the bindings got in the way and weren’t helping.

packages/ocapn-noise/gen/ocapn-noise.wasm
rekmarks, line 1: Should we be committing build output?

rust/ocapn_noise/src/lib.rs
```diff
@@ -0,0 +1,424 @@
...
+//            local (initiator for SYN and ACK, resonder for SYNACK):
+// 32   @0      PRIVATE_CRYPT_KEY
+// 32   @32     PUBLIC_CRYPT_KEY
+// 32   @64     SIGNING_KEY
+//
+//            SYN
+// 98   @128    SYN_PAYLOAD:
+// 32   @128      INITIATOR_VERIFYING_KEY
+// 64   @160      INITIATOR_SIGNATURE of initiator's PUBLIC_CRYPT_KEY
+// 1    @224      FIRST_ENCONDING supported
```
rekmarks:
```suggestion
// 1    @224      FIRST_ENCODING supported
```

rust/ocapn_noise/src/lib.rs
```diff
@@ -0,0 +1,424 @@
...
+use ed25519_dalek::{Signature, Signer, SigningKey, Verifier, VerifyingKey};
+use noise_protocol::patterns::noise_xx;
+use noise_protocol::{CipherState, HandshakeState, U8Array, DH};
+use noise_rust_crypto::{Blake2s, ChaCha20Poly1305, X25519};
+use rand_core::OsRng;
+
+// These bindings are deliberately not thread safe.
+// The intention is that the JavaScript driver will create an instance
+// of this virtual machine for each OCapN session.
+const SIZE: usize = 65535;
```
rekmarks, line 12: Just to satisfy my own curiosity, why `2**16`? OCapN messages not expected to ever exceed this size?

rekmarks: Oh, this is from Noise?

kriskowal: Correct. This is a limit set by Noise Protocol, precisely at 2**16-1.

packages/ocapn-noise/gen/ocapn-noise.wasm
kriskowal: I think so, because this enables JavaScript developers to proceed without a Rust to Wasm toolchain. But, we might revisit if we switch to more language-neutral tools as a baseline.

erights: For the current state of the world, I cannot disagree. But anytime we do so, it would (eventually) be good if 

- the build output were a deterministic result of the inputs and pinned versions of the tools
- we explained how to verify that the committed output is genuine by pointing at build instructions, or explaining them if needed.
- we encourage any noticed discrepancy to be publicly reported, as well as offering a bounty if the discrepancy is independently reproducible.

kriskowal: To that end, this change does include the Rust `Cargo.lock` which is analogous to `yarn.lock`. I believe Rust also automatically uses the compiler toolchain version that’s in the lockfile. (Go does something similar.) It also has a `build.sh` and any resulting differences to this file will be evident with `git diff`.

I could introduce Rust tooling in a CI job that automatically detects divergence. And, I can add a note to the `rust/ocapn_noise/README.md`.

rust/ocapn_noise/src/lib.rs
kriskowal: Good catch. That was fishy.

packages/ocapn-noise/gen/ocapn-noise.wasm
erights: Those would both be awesome!

Btw, the same applies to us anytime we checkin a bundle into our source tree.

PR #2959, opened 2025-09-17
refactor(ses): Simplify overrideTaming enablements.js

gibson042:
Define `moderateEnablements` as an explicit extension of `minEnablements`, inheriting the treatment of %IteratorPrototype%.


erights (approved):
LGTM, thanks!

PR #2962, opened 2025-09-18
feat(ocapn): Minimum viable OCapN

kumavis:
This PR is intended to land a work-in-progress implementation of the client, and is not ready to be published.

The implementation is a single vat single process (no comms vat).

Implements Passables
Remote calls + Pipelining
Third Party Handoffs
Eager promise subscription (lazy promise subscription to be explored later)
A tcp-testing-only netlayer (no tor netlayer yet)
This does not yet implement GC messages.

this was mostly built against this test suite https://github.com/ocapn/ocapn-test-suite
which lags behind the spec a bit. I expect dramatic spec changes to come, so I don't mind being behind spec at this point.

There is some implementation weirdness around use of pass-style due to a need for an ImmutableArrayBuffer. This implementation currently still uses mutable Uint8Arrays and thus some homespun passStyleOf checks. The intention is to replace these with exclusively real pass-style before publishing.

This PR replaces https://github.com/endojs/endo/pull/2781 with a rebase on `master`. See it for previous review and discussion. No other changes.

This is intended to be merged as is but is not yet deemed ready to be published

### DIFF
--- a/packages/ocapn/package.json
+++ b/packages/ocapn/package.json
@@ -35,7 +35,10 @@
     "test:xs": "exit 0"
   },
   "dependencies": {
+    "@endo/errors": "workspace:^",
+    "@endo/eventual-send": "workspace:^",
     "@endo/init": "workspace:^",
+    "@endo/marshal": "workspace:^",
     "@endo/pass-style": "workspace:^",
     "@endo/promise-kit": "workspace:^",
     "@noble/curves": "^1.9.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -695,8 +695,11 @@ __metadata:
   version: 0.0.0-use.local
   resolution: "@endo/ocapn@workspace:packages/ocapn"
   dependencies:
+    "@endo/errors": "workspace:^"
+    "@endo/eventual-send": "workspace:^"
     "@endo/init": "workspace:^"
     "@endo/lockdown": "workspace:^"
+    "@endo/marshal": "workspace:^"
     "@endo/pass-style": "workspace:^"
     "@endo/promise-kit": "workspace:^"
     "@endo/ses-ava": "workspace:^"


kumavis:
see previous PR https://github.com/endojs/endo/pull/2781 for review and discussion

packages/ocapn/src/captp/captp-engine.js
```diff
@@ -0,0 +1,535 @@
...
+    /**
+     * @param {CapTPSlot} slotID
+     */
+    slotID => {
+      releaseSlot(slotID);
+    },
+    { weakValues: gcImports },
+  );
+
+  let lastExportID = 0;
```
kriskowal, line 75: I have documented feelings about identity document initialism https://kriskowal.com/naming-things/#identifer

packages/ocapn/src/captp/captp-engine.js
```diff
@@ -0,0 +1,535 @@
...
+
+  /**
+   * Called when we have encountered a new value that needs to be assigned a slot.
+   *
+   * @param {any} val
+   * @returns {CapTPSlot}
```
kriskowal, line 81: I also have case convention feelings https://kriskowal.com/naming-things/#pick-one-case-convention

I recommend `CaptpSlot` or `CapSlot` or `Slot`.

packages/ocapn/src/client/index.js
```diff
@@ -0,0 +1,586 @@
...
+ * @typedef {import('./types.js').Client} Client
+ * @typedef {import('./types.js').NetLayer} NetLayer
+ * @typedef {import('./types.js').SelfIdentity} SelfIdentity
+ * @typedef {import('./types.js').Logger} Logger
+ * @typedef {import('./types.js').LocationId} LocationId
+ * @typedef {import('./types.js').PendingSession} PendingSession
+ * @typedef {import('./types.js').SessionManager} SessionManager
+ * @typedef {import('./ocapn.js').GrantTracker} GrantTracker
+ * @typedef {import('./ocapn.js').Ocapn} Ocapn
+ */
```
kriskowal, line 20: Can use `/** import {} from "" */` notation now, which is more transparent to the language server since it doesn’t create a new type.

packages/ocapn/src/client/types.js
```diff
@@ -0,0 +1,98 @@
+// @ts-check
```
kriskowal, line 1: Elsewhere, we’re moving in the direction of writing all-types files as `.ts`. This is not a hard line but one we can move incrementally.

packages/ocapn/src/cryptography.js
```diff
@@ -0,0 +1,154 @@
...
+ * @typedef {object} OcapnKeyPair
+ * @property {OcapnPublicKey} publicKey
+ * @property {(msg: Uint8Array) => OcapnSignature} sign
+ */
+
+/**
+ * @param {OcapnSignature} sig
+ * @returns {Uint8Array}
+ */
+export const ocapNSignatureToBytes = sig => {
```
kriskowal, line 32:
```suggestion
export const ocapnSignatureToBytes = sig => {
```

packages/ocapn/src/cryptography.js
```diff
@@ -0,0 +1,154 @@
...
+  // Double SHA256 hash of the public key
+  return sha256(sha256(publicKeyEncoded));
+};
+
+/**
+ * @param {Uint8Array} peerIdOne
+ * @param {Uint8Array} peerIdTwo
+ * @returns {Uint8Array}
+ */
+export const makeSessionId = (peerIdOne, peerIdTwo) => {
```
kriskowal, line 126: As much as I loathe “id”, I will accept it in general if it’s consistently capitalized like you have here: `Id`.

kriskowal (approved):
I’m reapproving this with light review and some comments that can be addressed now or later given that we want this to grow fast and evolve. We will likely be relying on the rigor of our test suite.

There’s code in here for finalization, which I would like to be sure works somehow.

PR #2972, opened 2025-09-26
refactor: Linting dependencies

kriskowal:
This is a minor refactor that restores linting dependencies to three packages, such that `yarn lint` will function again in each workspace.

### DIFF
--- a/packages/captp/package.json
+++ b/packages/captp/package.json
@@ -54,6 +54,11 @@
     "@endo/ses-ava": "workspace:^",
     "ava": "catalog:testing",
     "c8": "catalog:testing",
+    "eslint": "^8.57.0",
+    "eslint-config-airbnb-base": "^15.0.0",
+    "eslint-config-prettier": "^9.1.0",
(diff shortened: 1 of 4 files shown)


erights:
Is create-package.sh already set up (via skel) so that it makes new packages with all this set up correctly?

kriskowal:
> Is create-package.sh already set up (via skel) so that it makes new packages with all this set up correctly?

It took a moment and learned that, of all these `eslint-*` dependencies, we only actually need `eslint` and the existing `eslintConfig` that appeals to `@endo/eslint-plugin` for all the other stuff. So, I’ve posted an update to this change that does just that, and also made sure `packages/skel` got the current best configuration.

kriskowal:
For the record, I used this script to convince myself it worked:

```
yarn workspaces foreach -W run lint ; say done linting
```

There’s a possibility we could update our `lint` scripts to use `yarn run -T eslint`. My investigation into this run into a wall of inscrutable type errors, which exceeds the time box for now. I’m merging and saving that for later, possibly with a tip from @boneskull

PR #2973, opened 2025-09-26
refactor: Subsume lint into yarn catalog for testing

kriskowal:
In order to make more use of the deduplicative effect of yarn catalogs, this makes the catalog the source of truth for versions of linting packages.


turadg (approved):
Linting can be but isn't usually considered "testing". I suppose the only reason to have multiple catalogs is if they have different versions. Anticipating that dev deps tend to move together I suggest renaming to "dev-common".

PR #2976, opened 2025-10-01
feat(@endo/harden): Lightweight pre-lockdown hardened module support

kriskowal:
Refs: #2983
Refs: https://github.com/endojs/endo/issues/1686
Refs: [#2782](https://github.com/endojs/endo/pull/2782#issuecomment-3354015178)

## Description

This very small change introduces an `@endo/harden` package that provides a fake harden by default, but in a way that ensures that it is never used in composition with `lockdown`.
Description of the feature included in the PR documents.

While this avoids the need to eject the `harden` internals of `ses` into another package, along with the complications that entrained in #2782, we may elect to provide additional build conditions for `@endo/harden` that provide greater degrees of immutability to applications that opt-in, either for their own safety or for earlier detection of inadequate prototype hardening.

### Security Considerations

Currently, hardened modules can only be used after lockdown. This change introduces a new era where some hardened modules can be used before lockdown if the application forswears ever using lockdown in the same realm. If a hardened module is initialized before lockdown in error, this approach ensures that the developer gets a useful diagnostic with a stack pointed at the offending module.

### Scaling Considerations

Modules that adopt `@endo/harden` will be slightly larger. The overhead can be reduced considerably by using the `bundle-source -C hardened` condition, in which case the additional module will only be a few lines and any additional excess is in the package linkage.

This new feature is documented in the ses NEWS and the new harden module’s README.

### Testing Considerations

This module required multiple modules to exercise the various destructive behaviors of lockdown relative to when `harden` gets called.

Hardened modules were not previously usable before lockdown. This change commits us to a course where modules can be used before lockdown. Also, lockdown on previous hardened modules was already inadequate to the task of identifying gaps where the hardening of a module was not rigorous enough to the extent that prototypes were frozen hardened before the first instance was hardened. This is not a regression but does not create progress on that front. For progress toward ensuring rigorous hardening, I propose future changes that introduce new `lockdown` modes and `@endo/harden` build conditions that create the conditions for opt-in detection of inadequately hardened prototype chains.

The provision of `harden.isFake` is intended to compose well with other libraries that sense this property as an indication that they’re in a less-immutable environment.

### Upgrade Considerations

This change creates some coordination risks between versions of harden captured in bundles with divergent behavior from host environments that provide harden. I believe all of the matrix of possible interactions is accounted for.

- New bundles may entrain `@endo/harden` which will defer to `globalThis.harden` on platforms using older SES. There are some possible challenges with bundle size changing, but the changes should be minimal, and the build condition mitigation provided is not likely to make a big difference.
- Old hardened modules will continue to use `globalThis.harden` directly.
- New versions of SES will provide an additional `Object[@harden]` intrinsic that new code may come to rely upon. This new code will not be portable to older SES platforms. However, using a symbol and providing `@endo/harden` as a typed facade should discourage the creation of incompatible code.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -462,6 +462,19 @@ __metadata:
   languageName: unknown
   linkType: soft
 
+"@endo/harden@workspace:packages/harden":
+  version: 0.0.0-use.local
+  resolution: "@endo/harden@workspace:packages/harden"
+  dependencies:
+    ava: "catalog:dev"
+    c8: "catalog:dev"
+    eslint: "catalog:dev"
+    ses: "workspace:^"
+    tsd: "catalog:dev"
+    typescript: "npm:~5.9.2"
+  languageName: unknown
+  linkType: soft
+
 "@endo/immutable-arraybuffer@workspace:^, @endo/immutable-arraybuffer@workspace:packages/immutable-arraybuffer":
   version: 0.0.0-use.local
   resolution: "@endo/immutable-arraybuffer@workspace:packages/immutable-arraybuffer"


packages/harden/test/freezes-after-lockdown.test.js
```diff
@@ -0,0 +1,13 @@
...
+import harden from '../index.js';
+
+// This test is framed as a stand-alone module because calling `lockdown` has
+// side-effects on the realm.
+
+test('does not freeze if not locked down', async t => {
+  lockdown();
+  const object = {};
+  harden(object);
+  t.assert(Object.isFrozen(object));
```
michaelfig: I'm not entirely sure if what you're trying to assert is that `object` is actually hardened, or that the `Object.isFrozen` function reports `true` even though it may possibly have been monkey-patched by various modes of lockdown.

Please be more explicit by modifying this suggestion to mesh with what you're actually testing:

```suggestion
  const { isFrozen: isFrozenPreLockdown } = Object;
  lockdown();
  const { isFrozen: isFrozenPostLockdown } = Object;
  const object = {};
  harden(object);
  t.assert(isFrozenPreLockdown(object));
  t.assert(isFrozenPostLockdown(object));
```

packages/harden/package.json
```diff
@@ -0,0 +1,72 @@
...
+    "url": "https://github.com/endojs/endo/issues"
+  },
+  "type": "module",
+  "main": "./index.js",
+  "module": "./index.js",
+  "exports": {
+    ".": {
+      "hardened": "./hardened.js",
+      "default": "./index.js"
+    },
```
michaelfig, line 24: I like the use of a build condition!

packages/harden/README.md
```diff
@@ -0,0 +1,102 @@
...
+can be used either way.
+
+```js
+import { harden } from '@endo/harden';
+
+export const myFunction = () => {};
+harden(myFunction);
+
+export const MyConstructor = function MyConstructor() {};
+harden(MyConstructor.prototype);
```
michaelfig: I may be off-base, but this line seems a little misleading, IIUC the latest `harden` design.  Under HardenedJS, isn't it supposed to at the very least recurse over own properties (i.e. `Reflect.ownKeys(obj)`), but not prototypes (i.e. `Reflect.getPrototypeOf(obj)`)?

With that in mind, `harden(MyConstructor.prototype)` seems completely redundant, since `.prototype` is an own property of `MyConstructor`, since it will be hardened upon calling `harden(MyConstructor)`.

Reading further, the proposed pre-lockdown `harden` is a no-op.  If it's never recursive (not even under HardenedJS), it just seems like an attractive nuisance.

packages/harden/README.md
```diff
@@ -0,0 +1,102 @@
...
+harden(MyConstructor);
+```
+
+By avoiding the export of hoisted `function` and `var` declarations and by
+immediately calling `harden` on any exposed function (or prototype thereof!) we
+leave no window of opportunity for another module to alter our exports.
+Depending on whether the objects our functions return to be shared by multiple
+parties (like memoized objects), a hardened module author should consider
+hardening those objects too.
```
michaelfig:
```suggestion
If a function's return value is meant to be shared by multiple parties (such as memoized objects), a hardened module author should harden the value before the function returns it (`return harden(value);`).
```

packages/ses/NEWS.md
```diff
@@ -4,6 +4,15 @@ User-visible changes in `ses`
...
   the API already exported from `@endo/errors`.
+- The `lockdown` and `repairIntrinsics` functions will now sense if a
+  "hardened module" (one using the new `@endo/harden`) was initialized before
+  `lockdown`, causing `lockdown` to fail with an informative error indicating
+  that hardened modules were not properly initialized because they executed
+  before lockdown.
+  They sense this condition because `@endo/harden` will install its own
+  `Object[Symbol.for('harden')]` if `harden` gets used before `lockdown`.
+- Adds `Object[Symbol.for('harden')]`, a variation on `globalThis.harden` that
+  cannot be denied by endowment of an alternative `harden` to compartments.
```
michaelfig, line 15: Ahhhh.  Now it makes sense!  You may want to call this out in the `packages/harden/README.md`.

michaelfig (commented):
Looking good!  I was curious, so I took the time to review.

packages/harden/README.md
kriskowal: I think you are right that this is not a good example. I may need to appeal to @mhofman for a good example of a case where a defensively hardened module accidentally leaves something mutable exposed until the first instance is hardened. I’m likely confused and it’s possible that such a case is only possible with a speculatively different combination of alternative harden behaviors.

packages/harden/index.js
```diff
@@ -0,0 +1,101 @@
+/* global globalThis */
+
+const symbolForHarden = Symbol.for('harden');
+
+const makeFakeHarden = () => {
+  const harden = o => o;
+  harden.isFake = true;
+  harden.lockdownError = new Error(
+    'Cannot lockdown (repairIntrinsics) because @endo/harden used before lockdown on this stack',
```
erights, line 9: What does "on this stack" mean?

erights (changes requested):
Based on a partial review of #2978 , what I want from this PR is:

- Have the normal noses testing config not redact, so ***ALMOST ALL*** the changes to golden error message tests can revert.
- Abstract `harden.isFake || Object.isFrozen(...)` into a reusable abstraction, and use it.

Together, these would restore the brevity and readability of most existing tests.

kriskowal:
Notably, the new approach relies on https://github.com/endojs/endo/pull/3005 so tests using SES-AVA do not observe different error messages.

PR #2978, opened 2025-10-03
feat: Post-lockdown support in all hardened modules

kriskowal:
Closes: #1686

## Description

Many of Endo’s modules would be useful before lockdown but, until #2976, we did not have a strategy for implementing a module that could use `harden` without obligating the environment to be locked-down first. New strategies emerged with the insight that we can have `@endo/harden` “race” to provide a fake `harden` upon first invocation, in a way that prevents `lockdown` from succeeding, so partitioning the supported applications strictly between those that use `lockdown` and therefor only use a reliable `harden` after `lockdown`, or those that use a fake `harden` and _never_ call `lockdown`.

Consequently, we can now use `@endo/harden` throughout all Endo packages that use `harden` to enforce the defense of their integrity, so a single implementation can work with or without `lockdown`. This change validates #2976 by performing all necessary ramified relaxations necessary to make the Endo stack work without `lockdown`, without compromising its post-lockdown integrity. This change also makes some adjustments to `@endo/ses-ava` to allow a each package’s test suite to be used in a no-lockdown or post-lockdown configuration, and the changes necessary to adopt this feature in all improved packages.

These relaxations generally take two forms:

1. We do not have a mechanism in place that allows `@endo/ses-ava` to unredact error messages before lockdown, so many tests need to generalize their error message pattern matching to tolerate redaction. This is a potentially reversible commit, so it is separated.
2. Some pass-style invariants presume a functional `harden` and needed to relax in the presence of `harden.isFake`. These relaxations were previously accomplished with `lockdown` in `unsafe-fast` `hardenTaming` mode, which cheated by lobotomizing the MOP. With these relaxations, `unsafe-fast` may become unnecessary.

### Security Considerations

This change merits both close design and implementation review and is emphatically not small. To ease review, the changes are in individually reviewable commits to make obvious which changes are mechanical or closely related.

Particularly, the changes that relax `isFrozen(x)` to `!harden.isFake || isFrozen(x)` rely on `harden.isFake` being inviolable in a `lockdown` environment. To that end, it should be noted that `@endo/harden` selects to reflect either `globalThis.harden` or a new `Object[Symbol.for('harden')]`, preferring the former, and the former can be deliberately corrupted by the creator of a `Compartment` by replacing `globalThis.harden`, or be corrupted by any module sharing a compartment with a mutable `globalThis`. `importBundle` by default creates compartments with a frozen `globalThis` _but can be configured not to_.

It is possible to reverse the precedence of `globalThis` vs `Object`, but the hazard of providing a corrupted `globalThis.harden.isFake` is very small and being able to configure `globalThis.harden`, even a fake one, on a compartment-by-compartment basis may be useful.

- [ ] This is in any case a design consideration this change puts before its reviewers.

### Scaling Considerations

This change does slightly increase the linkage size for Endo modules, and benefits from `-C hardened` as a build condition to slightly mitigate the increase in size. Omitting the fake implementation of `@endo/harden` is unlikely to make bundles much smaller. The previous attempt at this feature, #2782, provided a much larger pre-lockdown implementation of `harden`.

The news for the affect modules should be sufficient for now to signal that these packages can be used without `lockdown`.

- [ ] We have long needed to provide an article with guidance on how to make “hardened modules”, modules that defend their interface against supply chain attacks, and with this change, we can add a footnote that this defensive posture does not come at the expense of usability without `lockdown`.

### Testing Considerations

This change mostly concerns the relaxation of tests so that they pass with or without `lockdown`. The machinery to do dual mode testing is included.

This change is intended to be backward compatible.

### Upgrade Considerations

Per #2976, this change accounts for new modules on old versions of SES, which is the primary consideration for upgrade. 

- [ ] This change must be accompanied by a passing integration test in Agoric SDK.


packages/captp/test/_ava-noses.config.js
```diff
@@ -0,0 +1,5 @@
+export default {
+  require: ['ses', '@endo/eventual-send/shim.js'],
```
michaelfig, line 2: I'm a little confused by why the `_ava-noses.config.js` files `require: ['ses', ...]`.
Is it really `_ava-nolockdown.config.js`?  That would be clearer.  Even though noses are cute, it's a misnomer, and clarity is worth throwing away an opportunity for humour.

If you found an apt abbreviation of `lockdown` to replace `ses` (like maybe `hjs` for Hardened JS), I'd suggest using it consistently throughout this PR.

packages/import-bundle/NEWS.md
```diff
@@ -1,5 +1,17 @@
 User-visible changes to `@endo/import-bundle`:
 
+# Next release
+
+- Relaxes dependence on a global, post-lockdown `harden` function by taking a
+  dependency on the new `@endo/harden` package.
+  Consequently, bundles will now entrain a `harden` implementation that is
+  superfluous if the bundled program is guaranteed to run in a post-lockdown
+  HardenedJS environment.
+  To compensate, use `bundle-source` with `-C hardened` or the analgous feature
```
michaelfig, line 10: I didn't notice if there were more NEWS.md files with this typo:

```suggestion
  To compensate, use `bundle-source` with `-C hardened` or the analogous feature
```

erights:
@kriskowal , just asking for a sanity check that the "tred" (transitive reduction) of the import graph 

packages-graph-tred

and the full import graph 

packages-graph

are what you intend. I myself don't see anything suspicious. Just want your sanity check.

packages/base64/NEWS.md
```diff
@@ -1,5 +1,19 @@
...
+  dependency on the new `@endo/harden` package.
+  Consequently, bundles will now entrain a `harden` implementation that is
+  superfluous if the bundled program is guaranteed to run in a post-lockdown
+  HardenedJS environment.
+  To compensate, use `bundle-source` with `-C hardened` or the analgous feature
+  for packaging conditions with your preferred bundler tool.
+  This will hollow out `@endo/harden` and defer exclusively to the global
+  `harden`.
+
+# Forgotten release
```
erights: What does "Forgotten release" mean?

packages/base64/test/_ava-noses.config.js
```diff
@@ -0,0 +1,5 @@
+export default {
```
erights, line 1: What does "-noses." mean?

erights: Aha! no ses. Nevermind. But someone else may likewise not immediately see this.

Resolving.

packages/pass-style/src/remotable.js
```diff
@@ -1,5 +1,4 @@
-/// <reference types="ses"/>
-
+import harden from '@endo/harden';
```
michaelfig, line 1: For the record, I'm really happy to see the `/// <reference types="ses"/>` lines disappear!

packages/base64/test/_ava-noses.config.js
erights: Unresolving. Surprised to see a `require: ['ses', ...` if this is the "no ses" config. I probably still misunderstand what you mean by "no ses".

packages/ses-ava/NEWS.md
```diff
@@ -1,5 +1,48 @@
...
+      "test": "yarn test:ses && yarn test:noses",
+      "test:ses": "ava --config test/_ava-ses.config.js",
+      "test:noses": "ava --config test/_ava-noses.config.js"
+    }
+  }
+  ```
+  The contents of `test/_ava-ses.config.js` pull in the new prepare endo config
+  module.
+  The only difference between this module and `@endo/ses-ava/prepare-endo.js`
+  is that it does not export a `default`.
```
michaelfig, line 25: Which doesn't export a `default`?  The former or the latter?

packages/base64/test/_ava-noses.config.js
erights: Further confused by what it mean to "require" ses, eventual-send, or (in _ava-ses.config.js) ses-ava when this package does not "depend" on these. Btw, I noticed this by looking at the generated graphs.

packages/stream-node/test/_ava-noses.config.js
michaelfig, line 1: After reading ses-ava's README.md, I submit these names to tickle your inspiration:

- `_ava-noses.config.js` -> `_ava-shims.config.js`
- `_ava-ses.config.js` -> `_ava-lockdown.config.js`

packages/common/test/object-meta-map.test.js
```diff
@@ -17,18 +19,21 @@ test('test objectMetaMap', async t => {
           },
     null,
   );
+  const expectedWritable = !!harden.isFake;
+  const expectedConfigurable = !!harden.isFake;
```
erights, line 23: cute

packages/errors/test/deep-send.test.js
```diff
@@ -9,7 +9,6 @@
 // `errorTaming` is subject to variation by environment variables. If no
 // LOCKDOWN_ERROR_TAMING is set, then it defaults to `'safe'`, whereas
 // `commit.js` sets it to `'unsafe'`.
-import '@endo/lockdown/commit.js';
```
erights, line 12: TODO(erights): I need to understand this change in this file and its two siblings below

packages/eventual-send/package.json
```diff
@@ -5,7 +5,7 @@
   "type": "module",
   "main": "src/no-shim.js",
   "scripts": {
-    "test": "ava",
+    "test": "yarn test:ses && yarn test:noses",
```
erights, line 8: Interesting!

packages/exo/test/exo-wobbly-point.test.js
```diff
@@ -131,11 +132,13 @@ test('ExoPoint instances', t => {
...
   t.throws(() => negPt.getX(), {
-    message: 'In "getX" method of (ExoPoint): result: -3 - Must be >= 0',
+    message:
+      /^In "getX" method of \(ExoPoint\): result: (\(a number\)|-3) - Must be >= (\(a number\)|0)$/,
   });
   // `self` calls are guarded
   t.throws(() => `${negPt}`, {
-    message: 'In "getX" method of (ExoPoint): result: -3 - Must be >= 0',
+    message:
+      /^In "getX" method of \(ExoPoint\): result: (\(a number\)|-3) - Must be >= (\(a number\)|0)$/,
```
erights: Is there any way we could provide a reusable abstraction so we don't need to make each occurrence less readable?

Actually, now that I think about it, how does the redacted case occur? If we're not calling lockdown, what installs the redaction mechanism?

packages/harden/NEWS.md
```diff
@@ -1 +1,3 @@
 User-visible changes in `@endo/harden`:
+
+# Next release
```
erights, line 3: Empty?

packages/import-bundle/test/compartment-wrapper.test.js
```diff
@@ -15,13 +14,16 @@ function check(t, c, n) {
...
   const Con = Object.getPrototypeOf(c.globalThis.Compartment).constructor;
-  t.throws(
-    () => new Con(),
-    {
-      message: 'Function.prototype.constructor is not a valid constructor.',
-    },
-    `${n} .constructor is tamed`,
-  );
+  // Constructor taming only applies in locked-down environments
+  if (Object.isFrozen(Object.prototype)) {
```
erights, line 18: Why not test `harden.isFake`?

packages/marshal/test/marshal-capdata.test.js
```diff
@@ -12,11 +13,6 @@ const {
   getPrototypeOf,
 } = Object;
 
-const harden = /** @type {import('ses').Harden & { isFake?: boolean }} */ (
-  // eslint-disable-next-line no-undef
-  global.harden
-);
```
erights, line 18: What was this about?

Now that it'll be gone and I understand why, my question doesn't matter. But I am curious.

packages/marshal/test/marshal-justin.test.js
```diff
@@ -101,24 +102,28 @@ test('qp for quote passable as a quasi-quoted Justin expression', t => {
...
-    // starts the slot count over again. Otherwise, `qp` would be a
-    // communications channel. Since the rendered string is only for
-    // diagnostic value and does not show the slot content anyway,
-    // this is ok.
-    // TODO maybe better would be to show the slot contents using an outer
-    // `bestEffortsStringify` on the slots array. But that would be a
-    // terribly confusing mix of notation. We simply can't show them in
-    // Justin other than by slots.
-    `xxx ${JSON.stringify(`\`[
+  if (Object.isFrozen(Object.prototype)) {
```
erights, line 105: Why not test `harden.isFake`?

packages/marshal/test/marshal-smallcaps.test.js
```diff
@@ -74,8 +70,8 @@ test('smallcaps serialize static data', t => {
   });
 
   const cd = ser(harden([1, 2]));
-  t.is(isFrozen(cd), true);
-  t.is(isFrozen(cd.slots), true);
+  t.is(harden.isFake || isFrozen(cd), true);
```
erights, line 73: Can we make a reusable abstraction for `harden.isFake || isFrozen(...)` to restore brevity? I'm a bit on the fence about this, but more in favor than not.

packages/marshal/test/string-rank-both-order.test.js
```diff
@@ -20,6 +21,8 @@ test('string ranking by code point/UTF-16 code unit agreement', t => {
       t.log(msg);
       if (!msg.startsWith('Comparisons differed: ')) return false;
       if (!msg.endsWith('-1 vs 1') && !msg.endsWith('1 vs -1')) return false;
+      // ses-ava is not currently able to unredact error messages before lockdown.
```
erights: TODO(erights): I think this is what I need to understand better.

packages/marshal/test/string-rank-both-order.test.js
```diff
@@ -20,6 +21,8 @@ test('string ranking by code point/UTF-16 code unit agreement', t => {
...
+      if (!Object.isFrozen(Object.prototype)) return true;
```
erights: test `harden.isFake`?

packages/ocapn/test/handoffs.test.js
```diff
@@ -5,8 +5,6 @@
  * @import { OcapnLocation } from '../src/codecs/components.js'
  */
 
-import '@endo/ses-ava/prepare-endo.js';
```
erights, line 8: Why only removal only here?

packages/pass-style/src/error.js
```diff
@@ -108,18 +107,23 @@ export const confirmRecursivelyPassableErrorPropertyDesc = (
     );
   }
   if (!hasOwn(desc, 'value')) {
-    return (
-      reject &&
-      reject`Passable Error ${q(
-        propName,
-      )} own property must be a data property: ${desc}`
-    );
+    // Without lockdown, Error.prototype.stack is typically an accessor property.
```
erights: What do you mean by "typically"? On v8 without lockdown, it is an own accessor property.

packages/pass-style/src/passStyle-helpers.js
```diff
@@ -157,6 +156,7 @@ const makeConfirmTagRecord = confirmProto => {
       (!isPrimitive(tagRecord) ||
         (reject && reject`A non-object cannot be a tagRecord: ${tagRecord}`)) &&
       (isFrozen(tagRecord) ||
+        harden.isFake ||
```
erights, line 159:
```suggestion
      (harden.isFake ||
        isFrozen(tagRecord) ||
```
because that's the order you use almost every else. Please normalize everywhere this appears.

packages/pass-style/src/safe-promise.js
```diff
@@ -134,7 +133,9 @@ const confirmPromiseOwnKeys = (pr, reject) => {
  */
 const confirmSafePromise = (pr, reject) => {
   return (
-    (isFrozen(pr) || (reject && reject`${pr} - Must be frozen`)) &&
+    (isFrozen(pr) ||
+      harden.isFake ||
```
erights, line 137: again

packages/pass-style/test/atom.test.js
```diff
@@ -1,26 +1,27 @@
...
+import harden from '@endo/harden';
 import { Far } from '../src/make-far.js';
 import { isAtom, assertAtom } from '../src/typeGuards.js';
 
 test('isAtom test', t => {
   const alice = Far('Alice', {});
   t.false(isAtom(alice));
   t.throws(() => assertAtom(alice), {
-    message: 'A "remotable" cannot be an atom: "[Alleged: Alice]"',
+    message: /A "remotable" cannot be an atom: /,
```
erights:
```suggestion
    message: /^A "remotable" cannot be an atom: /,
```

packages/pass-style/test/atom.test.js
```diff
@@ -1,26 +1,27 @@
...
   });
 
   const p = Promise.resolve();
   t.false(isAtom(p));
   t.throws(() => assertAtom(p), {
     message:
-      'Not even Passable: "[Error: Cannot pass non-frozen objects like \\"[Promise]\\". Use harden()]": "[Promise]"',
+      /A "promise" cannot be an atom: \(an object\)|Not even Passable: "\[Error: Cannot pass non-frozen objects like (\\"\[Promise\]\\"|\(an object\))\. Use harden\(\)\]": ("\[Promise\]"|\(an object\))/,
```
erights: What's going on here?

packages/pass-style/test/errors.test.js
```diff
@@ -48,17 +51,17 @@ test('toPassableError, toThrowable', t => {
   // Since then, we changed `makeError` to make reasonable effort
   // to return a passable error by default. But also added the
   // `sanitize: false` option to suppress that.
-  t.false(Object.isFrozen(e));
+  t.false(Object.isFrozen(e) && !harden.isFake);
```
erights: I almost missed that this does not demorganize into the not of the normal idiom. Why not?

Given the non-equivalence, I don't know that it is worth normalizing this one.

packages/pass-style/test/errors.test.js
```diff
@@ -48,17 +51,17 @@ test('toPassableError, toThrowable', t => {
...
   t.false(isPassable(e));
 
   // toPassableError hardens, and then checks whether the hardened argument
   // is a passable error.
   const e2 = toPassableError(e);
 
-  t.true(Object.isFrozen(e));
+  t.true(Object.isFrozen(e) || harden.isFake);
```
erights, line 60: normalize (or abstract)

packages/pass-style/test/errors.test.js
```diff
@@ -72,7 +75,8 @@ test('toPassableError, toThrowable', t => {
   // a throwable singleton copyArray containing a toThrowable(e), i.e.,
   // an error like e2.
   t.throws(() => toThrowable(notYetCoercable), {
-    message: 'Passable Error has extra unpassed property "foo"',
+    message:
+      /Passable Error "stack" own property must be a data property: \(an object\)|Passable Error has extra unpassed property "foo"/,
```
erights: Wow

packages/pass-style/test/passStyleOf.test.js
```diff
@@ -348,9 +356,24 @@ test('passStyleOf testing remotables', t => {
     'null-proto-tagRecord grandproto is rejected',
   );
 
-  t.throws(() => passStyleOf(Object.prototype), {
-    message: 'cannot serialize Remotables with accessors like "toString" in {}',
-  });
+  t.throws(
```
erights: Why the deeper change?

packages/pass-style/test/safe-promise.test.js
```diff
@@ -34,7 +36,8 @@ test('safe promise loophole', t => {
       {
         // Override mistake
         message:
-          "Cannot assign to read only property 'Symbol(Symbol.toStringTag)' of object '[object Promise]'",
+          // Note: #<Promise> without SES. [object Promise] with SES.
```
erights: Wild!

packages/ses-ava/prepare-endo-config.js
```diff
@@ -0,0 +1,3 @@
+// This indirection is necessary because this module exports a function named default
+// and ava calls said function if it exists when used in an ava config.
```
erights, line 2: What?

packages/base64/NEWS.md
kriskowal: It means that the notes below were not attributed to a release and I have not yet investigated which one. This lapsed because there wasn’t a `# Next release` header to alert me to the need to update `NEWS.md`.

packages/base64/test/_ava-noses.config.js
kriskowal: No SES. It’s a misnoser, though, so I should go through this holistically renaming these to something else, like `lockdown` and `lockup`. No, that’s not it. Open to suggestions. The critical difference is that `noses` means lockdown was not called, but shims including `ses` may be entrained if necessary, for `assert` and `HandledPromise` for example.

packages/errors/test/deep-send.test.js
kriskowal: The cumulative effect is that `lockdown` is getting imported in the Ava config instead of here, so that this code works regardless of the Ava configuration.

packages/ocapn/test/handoffs.test.js
kriskowal: The behavior moves to the Ava configuration so it can be used regardless of whether lockdown is applied.

Issue #2983, opened 2025-10-07
Decoupling of lockdown and Endo’s hardened modules


kriskowal:
## What is the Problem Being Solved?

Endo provides many “hardened” modules. These are modules that use `harden` defensively and, because `ses` reveals `globalThis.harden` only after `lockdown`, these hardened modules can only run in a locked-down environment.

These modules, particularly `@endo/marshal`, are useful in environments in which the difficulty exceeds the value of locking down, like web frontend properties where the value of lockdown is limited to cases where they also elect to use LavaMoat to mitigate supply chain attacks. The constraints of the lockdown environment interact with complications introduced by popular frontend frameworks.

This issue subsumes terminating CapTP without lockdown, but referenced for the useful design discussion: https://github.com/endojs/endo/issues/1686

Option: Refactor SES such that `@endo/harden` exports a version that can be used before lockdown but doesn’t visit prototypes
- https://github.com/endojs/endo/pull/2782
- https://github.com/endojs/endo/pull/2830

Option: Provide an `@endo/harden` that is a no-op by default with build-condition and lockdown-option opt-in for alternatives
- https://github.com/endojs/endo/pull/2976
- https://github.com/endojs/endo/pull/2978

Option: Provide a `@endo/harden` which is superficial or freezes the “surface” of objects by default, with opt-in for alternatives.
- https://github.com/endojs/endo/pull/3008

```ibis
? What should the *default* no-lockdown behavior of `harden` be when imported from `@endo/harden`, such that any other alternative is opt-in with `-C` conditions. Regardless of which option is the default, `bundle-source -C hardened` will provide a minimal-overhead implementation of `@endo/harden` that presumes that it will run in a locked-down environment, or throws upon initialization.
  : No freezing, a.k.a., pinky-swear immutability of pass-style.
    + Fast
    + Small (very little overhead to bundle sizes)
      + And therefore much less likely to grow existing bundles beyond size limits by default.
    - No enforcement of purported immutability of pass-style.
    - Either rejection by Endo patterns, or similar weakening of `matches`/`mustMatch`
  : Freezing the transitive closure of properties.
    + Safe to the extent that own properties are immutable
      - But unsafe to the extent that prototype properties that are not overshadowed are mutable
    - Entrains a copy or variant of the SES hardening machinery in bundles, which is likely to cause bundles to grow beyond their size limit.
      + Which can be mitigated with an opt-out to the no-freeze or expect global harden variant with a `-C condition`
[...]

PR #2987, opened 2025-10-09
feat(ses-ava): Multiplex configurations

kriskowal:
Refs: #2983 

## Preface

There are a number of ways an Endo environment can be configured. We consider it table stakes that being able to move between these configurations should have as little friction as possible. Code written that works in one configuration should work in others.

The important configuration is the strict, default post-lockdown environment. This is where most of our test coverage occurs today. However, we also support a lockdown mode `hardenTaming: unsafe`, through `@endo/init/unsafe-fast.js`, which lets single-tenant vats like the Agoric chain supervisor run fast and rely on pinky-swear immutability of hardened data and interfaces. We wish to have test coverage for this mode, and to that end, we wish to run our existing test suites in multiple modes.

## Description

In order to better cover multiple environment configurations, this change introduces a `ses-ava` command and machinery to let us reuse tests regardless of whether `lockdown` is called. Then, adopts this feature in every applicable Endo package, in every configuration for which tests currently pass.

The new `ses-ava` command passes most flags through to `ava`, and runs `ava` once for each of the configurations in `package.json`, including the default `ava` if present, then other configuration files named in a `sesAvaConfigs` section. The `ses-ava` command then goes on to add `--only-config-${x}` and `--no-config-${x}` to include or exclude configurations in a test run, to help facilitate isolation of test failures.

In this change, we update every applicable workspace’s `package.json` and provide a bank of common AVA configuration files at the root of the repository for each workspace to use. Without modification, we adopt all the configurations that are currently working, with the expectation that we will support 2 or 3 such configurations for most packages going forward.

### Security Considerations

None.

### Scaling Considerations

Tests will run longer.

See ses-ava/README and ses-ava/NEWS.

### Testing Considerations

In entire.

This will increase test coverage for supported scenarios.

### Upgrade Considerations

None.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -235,6 +235,7 @@ __metadata:
     "@endo/compartment-mapper": "workspace:^"
     "@endo/errors": "workspace:^"
     "@endo/init": "workspace:^"
+    "@endo/ses-ava": "workspace:^"
     "@endo/zip": "workspace:^"
     ava: "catalog:dev"
     babel-eslint: "npm:^10.1.0"
@@ -373,7 +374,7 @@ __metadata:
     c8: "catalog:dev"
     eslint: "catalog:dev"
     ses: "workspace:^"
-    ses0_18_3: "npm:ses@0.18.3"
+    ses0.18.3: "npm:ses@0.18.3"
     tsd: "catalog:dev"
     typescript: "npm:~5.9.2"
   languageName: unknown
@@ -562,6 +563,7 @@ __metadata:
     "@endo/init": "workspace:^"
     "@endo/ses-ava": "workspace:^"
     ava: "catalog:dev"
+    c8: "catalog:dev"
     eslint: "catalog:dev"
     ses: "workspace:^"
     typescript: "npm:~5.9.2"
@@ -591,6 +593,7 @@ __metadata:
   resolution: "@endo/nat@workspace:packages/nat"
   dependencies:
     "@endo/compartment-mapper": "workspace:^"
+    "@endo/ses-ava": "workspace:^"
     ava: "catalog:dev"
     c8: "catalog:dev"
     eslint: "catalog:dev"
@@ -605,6 +608,7 @@ __metadata:
   dependencies:
     "@endo/init": "workspace:^"
     "@endo/promise-kit": "workspace:^"
+    "@endo/ses-ava": "workspace:^"
     "@endo/stream": "workspace:^"
     ava: "catalog:dev"
     c8: "catalog:dev"
@@ -707,6 +711,7 @@ __metadata:
     "@endo/ses-ava": "workspace:^"
     "@fast-check/ava": "catalog:dev"
     ava: "catalog:dev"
+    c8: "catalog:dev"
     eslint: "catalog:dev"
     typescript: "npm:~5.9.2"
   languageName: unknown
@@ -738,6 +743,8 @@ __metadata:
     typescript: "npm:~5.9.2"
   peerDependencies:
     ava: ^5.3.0 || ^6.1.2
+  bin:
+    ses-ava: bin/ses-ava
   languageName: unknown
   linkType: soft
 
@@ -9816,6 +9823,7 @@ __metadata:
   version: 0.0.0-use.local
   resolution: "root-workspace-0b6124@workspace:."
   dependencies:
+    "@endo/ses-ava": "workspace:^"
     "@jessie.js/eslint-plugin": "npm:^0.4.2"
     "@lavamoat/allow-scripts": "npm:^3.3.5"
     "@lavamoat/preinstall-always-fail": "npm:^2.1.1"
@@ -10009,7 +10017,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"ses0_18_3@npm:ses@0.18.3":
+"ses0.18.3@npm:ses@0.18.3":
   version: 0.18.3
   resolution: "ses@npm:0.18.3"
   checksum: 10c0/c07df71cf297f31e862efbc2cc46ee724e85f73cc67002b9b4fa47089fc89d147a56e579b8012fb1534419c77afafbb98bc5808fae82096951479d0bd6dfb377


packages/ses-ava/README.md
```diff
@@ -30,7 +30,67 @@ dependency, bundlers might bundle your code with all of Ava.
...
+For example, the `lockdown` configuration might look like:
+
+```js
+export default {
+  require: ['@endo/ses-ava/prepare-endo-config.js'],
+};
+```
+
+This relies on `@endo/ses-ava/prepare-endo-config.js` to initialize an
+Endo envrionment, including the SES shims and Eventual Send shim, and also
```
gibson042:
```suggestion
Endo environment, including the SES shims and Eventual Send shim, and also
```

packages/ses-ava/test.js
```diff
@@ -0,0 +1,25 @@
...
+
+// A smelly but convenient use for a mutable export binding.
+// The alternative is to create another layer of wrapping around the test
+// function tree, which in this case, would be onerous.
+// eslint-disable-next-line import/no-mutable-exports
+let test = rawTest;
+
+export const register = newTest => {
+  test = newTest;
+};
```
gibson042: I think `register` is a bit too generic.
```suggestion
/**
 * @template [Context=unknown]
 * @param {import('ava').TestFn<Context>} newTestFn
 */
export const replaceAvaTestFn = newTestFn => {
  test = newTestFn;
};
```

packages/ses-ava/README.md
```diff
@@ -30,7 +30,67 @@ dependency, bundlers might bundle your code with all of Ava.
...
-## Compat note
+# Supporting multiple configurations
+
+SES-Ava also provides a command line tool, `ses-ava`, that can run Ava with
+multiple configurations in a single command, intercepting flags to filter
+for interesting configurations.
+The `ses-ava` command consumes the `"ava"` and (new) `"avaConfigs"` properties
+in `package.json` to discover and name the supported configurations and infer
+flags like `--only-configname` and `--no-configname` for each, where the `"ava"`
+configuration is the `default`, if present.
```
gibson042: We should rename "avaConfigs" to something that makes clear that the functionality is _not_ being provided by ava itself.
```suggestion
The `ses-ava` command consumes the `"ava"` and (new) `"sesAvaConfigs"` properties
in `package.json` to discover and name the supported configurations and infer
flags like `--only-configname` and `--no-configname` for each, where the `"ava"`
configuration is the `default`, if present.
```
Also note that such a wrapper is probably useful for more than just ses-ava... should we use "multiAvaConfigs" in anticipation of possible generalization?

packages/ses-ava/README.md
```diff
@@ -30,7 +30,67 @@ dependency, bundlers might bundle your code with all of Ava.
...
+  "avaConfigs": {
+    "lockdown": "test/_ava-lockdown.config.mjs",
+    "unsafe": "test/_ava-lockdown-unsafe.config.mjs",
+  }
+}
+```
+
+With this configuration, `ses-ava ...args --no-lockdown` and `ses-ava ...args
+--only-unsafe` would both just run the `unsafe` configuration.
+Using `ses-ava` under `c8` allows all configurations to cover used code.
```
gibson042: What controls which arguments are interpreted by `ses-ava` vs. passed along? Argument hockey against a moving target gets dicey.

packages/ses-ava/src/command.js
```diff
@@ -0,0 +1,66 @@
...
+    }
+    const noKey = noFlags.get(arg);
+    const onlyKey = onlyFlags.get(arg);
+    if (noKey) {
+      no.add(noKey);
+    } else if (onlyKey) {
+      only.add(onlyKey);
+    } else {
+      args.push(arg);
+    }
```
gibson042, line 97: Yeah, this dynamic sensitivity can make for awful debugging, and hinders future evolution. Suggestion: just claim everything that starts with `-` up to the first `--`.
```js
  // Claim all options up to the first `--` (if any), but pass through every
  // positional argument.
  const args = [];
  const argsIterator = process.argv.slice(2)[Symbol.iterator]();
  for (const arg of argsIterator) {
    if (arg === '--') {
      args.push(...argsIterator);
      break;
    } else if (!arg.startsWith('-')) {
      args.push(arg);
      continue;
    }
    if (arg.startsWith('--no-')) {
      no.add(arg.slice('--no-'.length));
    } else if (arg.startsWith('--only-')) {
      only.add(arg.slice('--only-'.length));
    } else {
      throw Error(`Unsupported argument: ${arg}`);
    }
  }
  for (const key of no) {
    if (Object.hasOwn(avaConfigs, key)) continue;
    throw Error(`Cannot find excluded configuration: ${key}`);
  }
  for (const key of only) {
    if (Object.hasOwn(avaConfigs, key)) continue;
    throw Error(`Cannot find exclusive configuration: ${key}`);
  }
```

packages/ses-ava/src/command.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+  const configs = (only.size > 0 ? only : all).difference(no);
```
gibson042: `--only-foo` with `--no-foo` should probably be rejected, but if not then letting `--no-foo` win seems odd.

packages/ses-ava/src/command.js
```diff
@@ -0,0 +1,66 @@
...
+  for (const config of configs) {
```
gibson042, line 138:
```suggestion
  if (!configs.size) throw Error(`No configurations selected!`);
  for (const config of configs) {
```

packages/ses-ava/src/command.js
kriskowal: Downside: You then need to `yarn test -- -m case`.

kriskowal: Also downside: You cannot `yarn test -m 'case name' --only-lockdown` and iterate on the final argument.

packages/ses-ava/README.md
kriskowal: I considered `sesAvaConfigs` to avoid name-squatting in Ava’s domain, so I think that’s two votes. We can update our packages if `ava` upstream’s the feature.

packages/ses-ava/README.md
kriskowal: I’m open to other options. My intuition is consistent with yours until I considered the ergonomics. I’m convinced, given the absence of better alternatives, that we should just use a more elaborate pair of prefixes to avoid collisions.

packages/ses-ava/src/command.js
kriskowal: Rewrote with safeties.

packages/ses-ava/README.md
kriskowal: I went with fully virtualizing the `ava` flags. The new code intercepts the common flags and passes them down. `--` allows for pass-thru of arbitrary additional flags, but isn’t necessary in any of the common cases. I went with `--no-config-` and `--only-config-` since they don’t collide, are unlikely to collide, and we can compensate for collisions later without breaking anything.

packages/ses-ava/README.md
```diff
@@ -30,7 +30,74 @@ dependency, bundlers might bundle your code with all of Ava.
...
+
+This relies on SES-Ava  to initialize an
+Endo environment, including the SES shims and Eventual Send shim, and also
+register the SES-Ava wrapped `test` declarator, which can unredact error
+messages produced by the Assert shim from SES.
+If the test doesn't import `@endo/ses-ava/test.js`, requiring
+`@endo/ses-ava/prepare-endo-config.js` ensures the environment is fully
+initialized.
+In the root of the Endo repository, look the the `ava-*.config.mjs` modules
+for example configurations.
```
gibson042:
```suggestion
In the root of the Endo repository, look the at the `ava-*.config.mjs` modules
for example configurations.
```

packages/ses-ava/README.md
```diff
@@ -30,7 +30,74 @@ dependency, bundlers might bundle your code with all of Ava.
 
 SES-Ava rhymes with Nineveh.
```
gibson042: Nit (throughout the documentation): https://github.com/avajs/ava/tree/main#how-is-the-name-written-and-pronounced

```suggestion
SES-AVA rhymes with Nineveh.
```

packages/ses-ava/src/command.js
```diff
@@ -0,0 +1,155 @@
+/* global process */
+/* eslint-disable no-await-in-loop, no-continue, no-labels, no-unreachable-loop */
+
+/* The ses-ava command allows a single package to run the same test suite with
+ * multiple configurations named in package.json under `sesAvaConfigs`.
+ * Each of these configurations requires a separate Ava config file like
+ * test/_ava.special.config.js and a name.
+ * The ses-ava command will by default run all tests in every mode but allows
+ * the user to pass --only-* and --no-* at any argument position for any of
+ * the named configurations to filter.
```
gibson042, line 10: Suggestion: rather than dynamically defining flags from configuration, use repeatable argument-bearing options.
```suggestion
 * The ses-ava command will by default run all tests in every mode but allows
 * the user to pass --only <name> and --exclude <name> (or shorthands -o <name>
 * and -x <name>) at any argument position for any of the named configurations
 * to filter.
```
(and if accepted, remember to also update the test descriptions and documentation in e.g. README.md)

packages/ses-ava/src/command.js
```diff
@@ -0,0 +1,155 @@
...
+  '-s',
+  '--serial',
+  '-u',
+  '--update-snapshots',
+  '--no-worker-threads',
+  '--verbose',
+  '--no-verbose',
+]);
+
+const doublePassThrough = new Set(['-m', '--match', '--node-arguments']);
```
gibson042: Composite suggestions (format short/long aliases on the same line; include options for `ava debug`; rename these two variables):
```suggestion
// We pass these flags directly to ava.
const passThroughFlags = new Set([
  ...['-s', '--serial'],
  ...['-t', '--tap'],
  ...['-T', '--timeout'],
  ...['-u', '--update-snapshots'],
  ...['-v', '--verbose'],
  '--no-worker-threads',
  '--no-verbose',
  // Specific to ava debug.
  '--break',
]);

// We read a single argument for each of these options and pass the pair
// directly to ava.
const passThroughArgOptions = new Set([
  ...['-m', '--match'],
  '--node-arguments',
  // Specific to ava debug.
  '--host',
  '--port',
]);
```

packages/ses-ava/src/command.js
```diff
@@ -0,0 +1,155 @@
...
+  for (const arg of argsIterator) {
+    if (arg === '--') {
+      passThroughArgs.push(...argsIterator);
+      break;
+    }
+    const noKey = noFlags.get(arg);
+    const onlyKey = onlyFlags.get(arg);
+    if (singlePassThrough.has(arg)) {
+      passThroughArgs.push(arg);
+    } else if (doublePassThrough.has(arg)) {
```
gibson042:
```suggestion
    } else if (doublePassThrough.has(arg)) {
      // Note that we don't support GNU-style `--name=value` syntax.
```

packages/ses-ava/src/command.js
```diff
@@ -0,0 +1,155 @@
...
+      foundNextArg: {
+        for (const nextArg of argsIterator) {
+          passThroughArgs.push(arg, nextArg);
+          break foundNextArg;
+        }
+        throw new Error(`Expected argument after ${arg}`);
+      }
```
gibson042: This seems like overkill in comparison to just calling `next()`.
```suggestion
      const { done, value } = argsIterator.next();
      if (done) throw new Error(`Expected argument after ${arg}`);
      passThroughArgs.push(arg, value);
```

packages/ses-ava/src/command.js
```diff
@@ -0,0 +1,155 @@
...
+    ];
+    const child = popen.spawn('ava', avaArgs, {
+      stdio: ['inherit', 'inherit', 'inherit'],
+    });
+    await new Promise((resolve, reject) => {
+      child.on('exit', code => {
+        process.exitCode ||= typeof code === 'number' ? code : 1;
+        if (failFast) {
+          process.exit();
+        }
```
gibson042:
```suggestion
        if (failFast && process.exitCode) {
          process.exit();
        }
```

packages/ses-ava/src/command.js
```diff
@@ -0,0 +1,155 @@
...
+      });
+      child.on('error', error => {
+        reject(error);
+      });
+    });
+
+    // Debug will only apply to the first matching configuration.
+    if (debug) {
+      break;
+    }
```
gibson042: I think this will be misleading, and also frustrating in cases where I cared about debugging against multiple configurations (or just not the first one). The child processes are run serially, so we might as well just apply debug to all of them.
```suggestion
```

packages/ses-ava/package.json
```diff
@@ -76,6 +82,11 @@
     ],
     "timeout": "2m"
   },
+  "sesAvaConfigs": {
+    "one": "test/_ava1.config.js",
+    "two": "test/_ava2.config.js",
+    "raw": "test/_ses-ava-is-ava.config.js"
+  },
```
gibson042, line 89: These config names seem unnecessarily generic; can we make the naming (and thus comprehension of how this file expects testing to proceed) more clear? A possibility:
```suggestion
  "sesAvaConfigs": {
    "custom-env": "test/_custom-env.ava-config.js",
    "must-ignore": "test/_must-ignore.ava-config.js",
    "raw": "test/_raw.ava-config.js"
  },
```

packages/ses-ava/src/command.js
```diff
@@ -0,0 +1,155 @@
...
+    const configFile = sesAvaConfigs[config];
+    if (configFile !== undefined && !fs.existsSync(configFile)) {
+      throw new Error(
+        `ses-ava config ${config} file does not exist: ${sesAvaConfigs[config]}`,
+      );
+    }
+  }
+
+  // Execute configurations serially.
+  for (const config of configs) {
```
gibson042, line 138: Suggestion: tack on our own detail.

```suggestion
  for (const config of configs) {
    console.warn(`[ses-ava] config:`, config);
```
(or similar)

packages/ses-ava/src/command.js
kriskowal: I am sure I will regret this, but in the interest of using time efficiently, I’m going to merge without addressing this one, though it is absolutely a good idea.

gibson042: Captured as #2998.

gibson042: Eh, I just went ahead and submitted a fix: #2999

PR #2988, opened 2025-10-09
feat(compartment-mapper)!: new hooks & type overhaul

boneskull:
Refs: #2929 

This PR supersedes #2929 which should now be considered abandoned.

This PR targets #2915, if you didn't notice.

## Description

This PR is "complete" and ready for review.  It makes a significant overhaul to types and adds new options to several public APIs.

### Description of changes to Types

- The types around `CompartmentMapDescriptor` have been vastly expanded to reflect the different "flavors" of `CompartmentMapDescriptor`, `CompartmentDescriptor`, `ModuleDescriptor` (now `ModuleDescriptorConfiguration` to differentiate it from `ses`' `ModuleDescriptor`), and the formatting of the compartment names (keys).  
- `CompartmentDescriptor.label` is now a canonical name.
- `CompartmentDescriptor.path` is removed
- `CompartmentDescriptor.compartments` is removed
- Added many type guards for validation of these differing types
  - Validation of different types of compartment map
  - Stricter validation
  - Validation of all types of compartment descriptors, module descriptor configurations, etc.

### New Default Behavior in `mapNodeModules()`

Currently, `mapNodeModules()` avoids adding a `ModuleDescriptor` (`ModuleDescriptorConfiguration`) to a nascent `CompartmentDescriptor`'s `modules` prop if policy disallows it.  This PR changes the behavior to consider policy & excise the _dependency_ from the `Node` itself _before_ the (dependency) `Graph` is translated into a `CompartmentDescriptor`.  This has implications around what sort of errors are thrown when Compartment A attempts to load a Compartment B to which it has no access (see note about policy tests below). 

### Other Notes

- Since `CompartmentDescriptor.compartments` has been removed, that means implicitly allowing dynamic requires of parent/ancestor Compartments from a given Compartment via absolute path _is no longer supported_.  That's fine, since:
  - This should be explicit in policy _anyway_
  - `@lavamoat/node` was the only user of this feature and no longer needs it
- The entry compartment now has a canonical name of `$root$`.  It _is_ valid within `PackagePolicy`; i.e. some other package can be allowed to access the `$root$` compartment.
- Many tests in `test/policy.test.ts` needed to change because of how `mapNodeModules`' `packageDependencies` hook works.  Instead of rejecting module descriptors (`ModuleDescriptorConfiguration`s) based on policy, we reject entire _dependencies_ before they can be "digested" into module descriptors.  This results in `ScopeDescriptor`s not being populated, in addition to `ModuleDescriptorConfiguration`s, and causes downstream effects. Different exceptions are thrown at different times, but the _intent_ of these tests doesn't deviate from "ensure this naughty behavior is not allowed."

### Security Considerations

None that I'm aware of.

### Scaling Considerations

Any consumer providing a resource-intensive blocking hook to `@endo/compartment-mapper` gets the performance they deserve.

This is a breaking change to the contents of the archives as well as a breaking change to the types. It is not a breaking change to the archive _format_, but it will contain different values (instead of `packageName-v<version>` the compartment descriptors will be keyed on the canonical name).

### Testing Considerations

- [ ] There are some unit tests but we need more integration tests.

The breaking changes mentioned above, but it also solves like ten different problems `@lavamoat/node` was having trying to get things to work.  It significantly improves ecosystem compatibility _and_ will improve the performance of `@lavamoat/node`.

### Upgrade Considerations

- [x] Include `*BREAKING*:` in the commit message with migration instructions for any breaking change.

* * * 

This commit introduces a comprehensive universal hook system for `@endo/compartment-mapper`,
enabling extensible customization of module mapping, bundling, and policy enforcement.

BREAKING CHANGES: `CompartmentDescriptor` objects no longer have a `path` property and the `label` property is now a canonical name. Types have changed dramatically. Enhanced validation of `CompartmentMapDescriptor` objects.

### `mapNodeModules()`

- `packageDataHook`: Called once before `translateGraph` with data about all packages found while crawling `node_modules`. Receives a read-only `Map<CanonicalName, PackageData>` where each entry contains:
  - `name`: Package name
  - `packageDescriptor`: The package.json contents
  - `location`: File URL to the package
  - `canonicalName`: The canonical name used in policy

- `packageDependenciesHook`: Called for each package's dependencies during graph translation, allowing dynamic modification of the dependency graph. Can add or remove dependencies from packages based on policy or other criteria.

- `unknownCanonicalNameHook`: Called when policy references canonical names that don't exist in the dependency graph, with optional suggestions for typos/similar names.

### `makeImportHookMaker()`
- `moduleSourceHook()`: Called when module source objects are created (either from local files containing `bytes`, exit modules, or "error-type" sources)

### `captureFromMap()`
- `packageConnectionsHook`: Called during digest; surfaces "connections" between compartment descriptors

#### CompartmentDescriptor Interface Restructure:
- **BREAKING**: Removed `path` property - compartment descriptors no longer track dependency paths
- **BREAKING**: `label` property type changed from `string` to `CanonicalName<U>` - labels are now type-safe canonical names
- **BREAKING**: Made `CompartmentDescriptor` generic with `<T extends ModuleDescriptorConfiguration, U extends string>` for better type safety
- Added `location: string` as a required property for all compartment descriptors
- Added optional `sourceDirname` as found in the sources

#### CompartmentMapDescriptor Genericization (Is That A Word? No):
- **BREAKING**: `CompartmentMapDescriptor` is now generic: `CompartmentMapDescriptor<T, Name, EntryName>`
- **BREAKING**: `compartments` property changed from `Record<string, CompartmentDescriptor>` to `CompartmentDescriptors<T, Name>`
- **BREAKING**: `entry` property type changed from `EntryDescriptor` to `EntryDescriptor<EntryName>`
- New specialized types: `PackageCompartmentMapDescriptor`, `FileCompartmentMapDescriptor`, and `DigestedCompartmentMapDescriptor`

#### Enhanced Module Descriptor System:
- `ModuleDescriptor` is now `ModuleDescriptorConfiguration` to differentiate between it & the `ModuleDescriptor` from `ses`
- Added `ModuleDescriptorConfigurationCreator` enum tracking module creation source (`'link' | 'transform' | 'import-hook' | 'digest' | 'node-modules'`)
- Enhanced `BaseModuleDescriptorConfiguration` with `__createdBy` property for debugging
- Added `ErrorModuleDescriptorConfiguration` for deferred error handling
- Improved type discrimination with `ModuleDescriptorConfigurationKindToType` utility type

#### `canonical-name.ts` - Canonical Name Type System:
- **NEW**: Type-level canonical name validation using template literal types
- `CanonicalName<S>` - validates npm package name chains separated by '>' (e.g., "foo>bar", "@scope/pkg>dep")
- `ScopedPackageName` and `UnscopedPackageName` for npm package name validation
- `SplitOnGt<S>` and `AllValidPackageNames<Parts>` for compile-time canonical name parsing
- `IsCanonicalName<S>` predicate type for conditional type logic
- 
#### Compartment Descriptor Validation:
- `DigestedCompartmentDescriptor` - restricted descriptor for archived compartment maps
- Properties marked as `never` for digested descriptors: `path`, `retained`, `scopes`, `parsers`, `types`, `__createdBy`, `sourceDirname`
- `CompartmentDescriptorWithPolicy<T>` - enforces policy presence where required

#### Module Configuration Type Safety:
- `ModuleDescriptorConfigurationKind` union for module type discrimination
- Type-safe module configuration creators with `__createdBy` tracking

#### Policy Integration:
- Enhanced policy types in `PackageCompartmentDescriptor` with canonical name constraints
- `LiteralUnion` usage for special canonical names (`ATTENUATORS_COMPARTMENT`, `ENTRY_COMPARTMENT`)
- Policy-aware compartment descriptor types with enhanced validation

#### `typescript.ts` Enhancements:
- Enhanced `LiteralUnion<L, B>` for better literal type preservation
- Type utilities supporting the new generic compartment map architecture
- Moved `Simplify` from tests here (because it's useful dammit)

### Policy Validation:
- Unknown canonical name detection with suggestion system
- Cross-reference policy resources against actual compartment contents
- Detailed error reporting with path information for policy issues
- Hook integration for custom policy validation logic

### DIFF
--- a/packages/compartment-mapper/src/hooks.md
+++ b/packages/compartment-mapper/src/hooks.md
@@ -1,0 +1,124 @@
+# Review of compartment-mapper hooks
+
+
+
+| Hook Name                 | Description                                                                           |
+|-------------------------- | --- |
+| `packageDataHook`         | Receives all found package descriptors data before graph translation.                                   |
+| `packageDependenciesHook` | Allows dynamic mutation of dependencies during node_modules graph translation.              |
+| `unknownCanonicalNameHook`| Called when the policy references unknown canonical names, can suggest typos/similar names. |
+| `moduleSourceHook`        | Invoked when a module source is created.                                             |
+| `packageConnectionsHook`  | Surfaces connections during digest. (ignored in archiving)                                               |
+
+
+
+[Type declarations for the hooks](./types/external.ts)
+
+```mermaid
+graph TB
+
+  exports((public exports))
+  exports -.- compartmentMapForNodeModules
+  exports -.- mapNodeModules
+  exports -.- loadLocation
+  exports -.- importLocation
+  exports -.- captureFromMap
+  exports -.- importFromMap
+  exports -.- loadFromMap
+  exports -.- digestCompartmentMap
+  exports -.- makeFunctor
+  exports -.- makeScript
+
+
+subgraph "import-hook.js"
+  moduleSourceHook{{moduleSourceHook}} -.- note5["for all module sources read"]
+  makeDeferError --> moduleSourceHook
+  makeImportHookMaker --> makeDeferError
+  makeImportNowHookMaker --> makeDeferError
+  makeImportNowHookMaker -- via importNowHook --> chooseModuleDescriptor --> executeLocalModuleSourceHook--> moduleSourceHook
+  makeImportHookMaker --via importHook exitModule logic--> moduleSourceHook
+  makeImportHookMaker --via importHook call--> chooseModuleDescriptor
+end
+
+subgraph "node-modules.js"
+  compartmentMapForNodeModules --> packageDataHook{{packageDataHook}} -.- note0["called once"]
+  compartmentMapForNodeModules --> unknownCanonicalNameHook{{unknownCanonicalNameHook}} -.- note1["for all issues from policy;<br>after defaultUnknownCanonicalNameHandler"]
+  compartmentMapForNodeModules --> translateGraph --> packageDependenciesHook{{packageDependenciesHook}} -.-note3["for all locatons in graph<br>after defaultPackageDependenciesFilter"]
+  mapNodeModules --> compartmentMapForNodeModules
+
+end
+
+subgraph "digest.js"
+  digestCompartmentMap --> translateCompartmentMap --> packageConnectionsHook{{packageConnectionsHook}} -.- note4["for all retained compartments"]
+end
+
+subgraph "bundle.js"
+  makeFunctor -- options:can include hooks ----------> mapNodeModules
+  makeScript -- options:can include hooks ----------> mapNodeModules
+end
+
+
+subgraph "import-lite.js"
+  importFromMap --> loadFromMap ---> makeImportHookMaker
+  loadFromMap ----> makeImportNowHookMaker
+end
+
+
+subgraph "capture-lite.js"
+  captureFromMap -----> makeImportHookMaker
+  captureFromMap --> captureCompartmentMap --> digestCompartmentMap
+end
+
+
+subgraph "import.js"
+  loadLocation ----------> mapNodeModules
+  importLocation --> loadLocation
+  loadLocation --> loadFromMap
+end
+
+
+%% STYLING
+classDef note fill:#999, stroke:#ccb
+class note0,note1,note2,note3,note4,note5 note
+
+```
+
+
+
+<details>
+<summary>Bundle and Archive bits of the diagram that don't use hooks</summary>
+
+These are calling the functions accepting hooks but don't pass them
+
+> [TODO]
+> copy-paste this to the main diagram whenever the connections are made.
+
+```mermaid
+
+graph TB
+
+subgraph "bundle.js"
+  makeFunctor -- options:transparently ----------> mapNodeModules
+  makeScript -- options:transparently ----------> mapNodeModules
+  makeFunctorFromMap  -- no moduleSourceHook ----x makeImportHookMaker
+end
+
+subgraph "bundle-lite.js"
+makeFunctorFromMap2 --no moduleSourceHook -----x makeImportHookMaker
+end
+
+subgraph "archive-lite.js"
+digestFromMap -- no moduleSourceHook -----x makeImportHookMaker
+makeArchiveCompartmentMap --no packageConnectionsHook----x digestCompartmentMap
+end
+
+subgraph "archive.js"
+archive(("multiple <br> methods")) --no hooks passed-----------x mapNodeModules
+end
+
+```
+
+</details>
+
+
+


packages/compartment-mapper/src/policy.js
```diff
@@ -115,60 +120,63 @@ export const detectAttenuators = policy => {
...
-    return policy.entry;
-  }
-  const canonicalName = generateCanonicalName(namingKit);
-  if (canonicalName === ATTENUATORS_COMPARTMENT) {
-    return { defaultAttenuator: policy.defaultAttenuator, packages: 'any' };
+export const makePackagePolicy = (
+  label,
+  { policy, packagePolicy = undefined } = {},
+) => {
+  if (packagePolicy !== undefined) {
```
naugtur: how would a package policy be already known but still go through this function? What is the point of doing it here instead of checking at call location?

packages/compartment-mapper/src/policy.js
```diff
@@ -237,8 +245,10 @@ export const makeDeferredAttenuatorsProvider = (
       throw Error(`No attenuators specified in policy`);
     };
   } else {
-    defaultAttenuator =
-      compartmentDescriptors[ATTENUATORS_COMPARTMENT].policy.defaultAttenuator;
+    // TODO: the attenuators compartment must always have a non-empty policy; maybe throw if it doesn't
```
naugtur: sounds good

packages/compartment-mapper/src/node-modules.js
```diff
@@ -50,21 +65,75 @@ import { GenericGraph, makeShortestPath } from './generic-graph.js';
...
+ * @param {CanonicalName} params.canonicalName
+ * @param {string} params.message
+ * @param {LogFn} params.log
+ */
+const defaultUnknownCanonicalNameHandler = ({
+  canonicalName,
+  message,
+  log,
+}) => {
+  log(`WARN: Invalid resource ${q(canonicalName)} in policy: ${message}`);
```
naugtur, line 100: This kind of things, at least in my experience with webpack, worked better if aggregated and displayed as a single warning. In a debug mode of sorts they could be used to generate proposed policy fixes.

packages/compartment-mapper/src/link.js
```diff
@@ -47,17 +57,17 @@ const { allSettled } = Promise;
...
 const promiseAllSettled = allSettled.bind(Promise);
 
-const defaultCompartment = Compartment;
-
-// q, as in quote, for strings in error messages.
-const q = JSON.stringify;
+const DefaultCompartment = Compartment;
 
 /**
+ * TODO: can we just use `Object.hasOwn` instead?
```
naugtur, line 63: It's a new addition to SES AFAIR

packages/compartment-mapper/src/node-modules.js
```diff
@@ -613,13 +695,25 @@ const gatherDependency = async (
...
   if (dependency === undefined) {
     // allow the dependency to be missing if optional
     if (optional || !strict) {
       return;
     }
     throw Error(`Cannot find dependency ${name} for ${packageLocation}`);
   }
+
+  // #region packageDescriptor hook
+  if (packageDescriptorHook) {
```
naugtur: I'm not sure this is useful, but it could be useful if it allowed altering the descriptor (eg. in case it contain something that was not supported or needed a fix.

packages/compartment-mapper/src/capture-lite.js
```diff
@@ -263,6 +273,8 @@ export const captureFromMap = async (
     Compartment: CompartmentOption = DefaultCompartment,
     log = noop,
     preload = [],
+    packageConnectionsHook,
```
naugtur, line 277: `packageDependenciesHook` and `packageConnectionsHook` are similar

I looked at both of these hooks and I think `packageConnectionsHook` could be replaced with `packageDependenciesHook` and the main difference is that one of them allows returning a mutated set. 
If we could offer the same in digest, we'd have a single hook for both.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -504,9 +554,19 @@ const makeDeferError = (
...
       deferredError: error.message,
     };
+    packageSources[specifier] = moduleSource;
+
+    if (moduleSourceHook) {
+      moduleSourceHook({
+        moduleSource: { error: moduleSource.deferredError },
+        canonicalName,
+        log,
+      });
```
naugtur: (should) 

I'm not sure I understand the point of calling the hook for a faux module source that only represents a deferred error. There's no point in scanning it and the only value this could have is if there was a point in knowing the error is there, waiting to potentially be invoked.

deferred error exists because our lexer was configured to catch all reasonably looking invocations of `require(specifier)` even if it's a local scope function named `require`. In that case, the actual require function will never be called for that specifier and the error will never be thrown. 

In that case - we're passing garbage to the hook as far as I'm concerned.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -751,9 +836,9 @@ export function makeImportNowHookMaker(
         if (exitRecord) {
           // It'd be nice to check the policy before importing it, but we can only throw a policy error if the
           // hook returns something. Otherwise, we need to fall back to the 'cannot find' error below.
-          enforceModulePolicy(moduleSpecifier, compartmentDescriptor, {
+          enforcePolicyByModule(moduleSpecifier, compartmentDescriptor, {
```
naugtur, line 819: (should) Verify whether it's intentional that only async importHook flow calls `moduleSourceHook` for the exit modules it finds and here it's not being called.

Also, I'm not sure if there is a point in calling it for exit modules. What would be the usecase?

packages/compartment-mapper/src/hooks.md
```diff
@@ -0,0 +1,124 @@
...
+
+end
+
+subgraph "digest.js"
+  digestCompartmentMap --> translateCompartmentMap --> packageConnectionsHook{{packageConnectionsHook}} -.- note4["for all retained compartments"]
+end
+
+subgraph "bundle.js"
+  makeFunctor -- options:can include hooks ----------> mapNodeModules
+  makeScript -- options:can include hooks ----------> mapNodeModules
```
naugtur, line 57: (nit) Typesscript would complain, but it seems `MapNodeModulesOptions` would be successfully passed through these.

packages/compartment-mapper/src/bundle.js
naugtur, line 676: (nit) transparently passes options to mapNodeModules, therefore allowing hooks if you ignore typescript squiggles.

naugtur, line 691: (nit) transparently passes options to mapNodeModules, therefore allowing hooks if you ignore typescript squiggles.

packages/compartment-mapper/src/capture-lite.js
boneskull: The data from `packageDependenciesHook` and `packageConnectionsHook` come from two different types of compartment map, and the latter happens post-digest. I would not want to do this.

packages/compartment-mapper/src/hooks.md
boneskull: TypeScript doesn't complain because `BundleOptions` is a superset of `MapNodeModulesOptions`

packages/compartment-mapper/src/import-hook.js
boneskull: ~~OK, I probably don't even have coverage for this anyway.  Wasn't sure if I should do this or not; I'll remove it.~~

boneskull: Actually, I am listening for this on the LavaMoat side and will abort generation if it occurs.  Would you or would you not agree that if this "error" module source is encountered, it would be something that needs addressing?  If it does ever happen, it means we will not be able to generate a complete policy (and may be missing more than can reasonably be user-defined in overrides), even if we do ignore & continue.

packages/compartment-mapper/src/import-hook.js
boneskull: Well, we only use this at time of generation. Generation _only_ calls `importHook` and never `importNowHook`.  So that's why.  

We aren't using this on the LavaMoat side, but I'm not sure why we wouldn't, just semantically speaking.  These are all `ModuleSource` objects whether they represent source on disk, a builtin, or an error.  🤷

packages/compartment-mapper/src/node-modules.js
boneskull: This sounds like more work 😄

I could do this, but I think even reporting this at all is going out on a limb (in terms of the overall user-friendliness of `@endo/compartment-mapper`).  I would not be necessarily opposed to expanding this and implementing other friendly improvements, but I don't think they need to be in this PR.

packages/compartment-mapper/src/policy.js
boneskull: Actually... hm. This is leftover from something else.  It must be called in `defaultPackageDependenciesFilter`, but that _used_ to add the `packagePolicy` to the `Node` directly, but I think I undid that change because it was kind of a naughty side-effect.

I can remove the option.  It might be worthwhile to cache the result, otherwise we compute more times than necessary.

I also wanna rename `defaultPackageDependenciesFilter` because it is not a "default" that can be overridden.

packages/compartment-mapper/src/bundle.js
boneskull: You shouldn't be getting any squiggles

boneskull: ditto

packages/compartment-mapper/src/hooks.md
naugtur: I missed that. saw it was extending ArchiveOptions

packages/compartment-mapper/src/import-hook.js
naugtur: This will make for annoying documentation: "this hook is triggered whenever (...) except for this one case (...)" 
Unless there's a reason calling it there would be invalid, I'd go for it for consistency.

packages/compartment-mapper/src/import-hook.js
naugtur: deferred error modules are a natural consequence of not being able to detect what is and what is not _the_ require function early enough. aborting policy generation on finding those is incorrect. 
They exist only to avoid throwing errors at link time about not found packages when the string argument that doesn't match a known package name is passed to a function that happens to be called `require` (but we do want to still throw them at runtime when cjs stuff resolves)

I object to exposing them to the hook.

packages/compartment-mapper/src/capture-lite.js
naugtur: I'm not enjoying there being two flows that are somewhat similar but ultimately distinct. But the solution to that is ultimately not forcing the unification fo the hook, but finding  way to split regular flow into steps so that capture was no longer necessary a all.
Can leave this be if you're certain unifying the 2 hooks wouldn't be feasible.
Note that I'm not fond of explaining the details and differences between the two different types of compartment maps to the end user :see_no_evil:

packages/compartment-mapper/test/dynamic-require.test.js
naugtur, line 67: (nit) Why is this doing anything instead of just throwing?

naugtur: (nit) it's used in a few spots, so it could become a factory that returns an importNowHook and the count getter.

packages/compartment-mapper/src/capture-lite.js
boneskull: Yeah, I mean, ideally the end user wouldn't need to touch a compartment map.

packages/compartment-mapper/src/bundle-lite.js
```diff
@@ -188,28 +193,18 @@ const sortedModules = (
...
-      );
-      assert(
-        sourceDirname !== undefined,
-        `No sourceDirname for ${moduleSpecifier} in ${compartmentName}`,
-      );
-      if (deferredError) {
+      if (isErrorModuleSource(source)) {
         throw Error(
-          `Cannot bundle: encountered deferredError ${deferredError}`,
+          `Cannot bundle: encountered deferredError ${source.deferredError}`,
```
naugtur, line 198: I think this is incorrect. deferredErrors are deferred because we don't know if they're actual errors until runtime. They should not be a reason to fail anything prior to that. It's the whole purpose of their existence (as opposed to throwing right away)

I know it's not being introduced in this PR, but let's at least file an issue for it so it's there when @kriskowal  comes back to the bundler.

packages/compartment-mapper/src/types/compartment-map-schema.ts
```diff
@@ -8,80 +8,231 @@
...
+}
+
+export interface ErrorModuleDescriptorConfiguration
+  extends BaseModuleDescriptorConfiguration {
+  deferredError: string;
+}
+
+/**
+ * For every module explicitly mentioned in an `exports` field of a
+ * `package.json`, there is a corresponding `CompartmentModuleDescriptorConfiguration`.
```
naugtur: (could) 
ModuleDescriptorConfiguration and CompartmentModuleDescriptorConfiguration have the same exact description. 
It'd be nice to add the differentiating factor to the description - in what cases or why one is used and not the other.

packages/compartment-mapper/src/types/compartment-map-schema.ts
```diff
@@ -8,80 +8,231 @@
...
+   */
+  compartment: LiteralUnion<typeof ATTENUATORS_COMPARTMENT, FileUrlString>;
+  /**
+   * The module name within {@link CompartmentDescriptor.modules} of the
+   * `CompartmentDescriptor` referred to by {@link compartment}
+   */
+  module: string;
+}
+
+export interface ExitModuleDescriptorConfiguration
```
naugtur: (could) add a description

packages/compartment-mapper/src/types/compartment-map-schema.ts
```diff
@@ -8,80 +8,231 @@
...
+export type CompartmentDescriptorWithPolicy<
+  T extends ModuleDescriptorConfiguration = ModuleDescriptorConfiguration,
+> = Omit<CompartmentDescriptor<T>, 'policy'> & { policy: SomePackagePolicy };
+
+/**
+ * A compartment descriptor digested by `digestCompartmentMap()`
+ */
+export interface DigestedCompartmentDescriptor
+  extends CompartmentDescriptor<ModuleDescriptorConfiguration> {
+  path: never;
```
naugtur, line 158: Wait, didn't we remove path from the parent type?
(Oh no, now you now I'm reading types)

packages/compartment-mapper/src/compartment-map.js
```diff
@@ -26,432 +56,904 @@ function* enumerate(iterable) {
...
+ * @returns {asserts allegedModule is ModuleDescriptorConfiguration}
+ */
+
+/**
+ * @param {unknown} allegedModule
+ * @param {string} keypath
+ * @param {string} url
+ * @param {ModuleDescriptorConfigurationKind[]} kinds
+ */
+function assertModuleConfiguration(allegedModule, keypath, url, kinds) {
```
naugtur, line 316: What reasons do we have to provide a custom list of kinds? I see digested module and file provide subsets, but it's unclear to me why some of the items are there while others aren't 
(could) This function would be a nice place to document that maybe.

packages/compartment-mapper/src/policy.js
```diff
@@ -115,60 +120,58 @@ export const detectAttenuators = policy => {
...
- * Returns the policy applicable to the canonicalName of the package
- *
+ * Generates the {@link SomePackagePolicy} value to be used in {@link CompartmentDescriptor.policy}
  * @overload
- * @param {PackageNamingKit} namingKit - a key in the policy resources spec is derived from these
- * @param {SomePolicy} policy - user supplied policy
- * @returns {SomePackagePolicy} packagePolicy if policy was specified
+ * @param {CanonicalName | typeof ATTENUATORS_COMPARTMENT} label
+ * @param {object} options Options
+ * @param {SomePolicy} options.policy User-supplied policy
```
naugtur: If the intention behind this was to ensure that if the `options` arg gets provided, it has to contain `policy` at least, I don't see it working when I type `makePackagePolicy('asdf', {})`

packages/compartment-mapper/src/policy.js
```diff
@@ -115,60 +120,58 @@ export const detectAttenuators = policy => {
...
+        packages: WILDCARD_POLICY_VALUE,
+      };
+    } else if (label === ENTRY_COMPARTMENT) {
+      packagePolicy = policy.entry;
+    } else if (label) {
+      packagePolicy = policy.resources?.[label];
+    }
+    return packagePolicy ?? create(null);
   }
+  return undefined;
```
naugtur, line 165: (nit) I liked the early return pattern

packages/compartment-mapper/src/policy.js
```diff
@@ -115,60 +120,58 @@ export const detectAttenuators = policy => {
...
+      packagePolicy = {
+        defaultAttenuator: policy.defaultAttenuator,
+        packages: WILDCARD_POLICY_VALUE,
+      };
+    } else if (label === ENTRY_COMPARTMENT) {
+      packagePolicy = policy.entry;
+    } else if (label) {
+      packagePolicy = policy.resources?.[label];
+    }
+    return packagePolicy ?? create(null);
```
naugtur, line 163: (attention) This is a change in logic. Previously if policy.entry was undefined the function would return undefined.
The difference in meaning is therefore - previously when no policy was set for entry it could do everything as if there was no policy, now it's constrained.

Is this deliberate? Am I wrong? Does spotting this keep me from being replaced by AI for one more month?

packages/compartment-mapper/src/node-modules.js
```diff
@@ -759,19 +844,101 @@ const graphPackages = async (
...
+    const packageDependenciesHookInput = {
+      canonicalName: label,
+      dependencies: new Set(dependencies),
+      log,
+    };
+
+    // Call default filter first if policy exists
+    let packageDependenciesHookResult;
+    if (policy) {
+      packageDependenciesHookResult = prePackageDependenciesFilter(
```
naugtur, line 882: (nit) nothing in prePackageDependenciesFilter name says it depends on policy. This if statement should be inside of it, but that's a total nitpick.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -759,19 +844,101 @@ const graphPackages = async (
...
+          return newDependencyLocations;
+        } catch {
+          log(
+            `WARNING: packageDependencies hook returned invalid value ${q(
+              packageDependenciesHookResult,
+            )}; using original dependencies`,
+          );
+        }
+      } else {
+        dependencyLocations = create(null);
```
naugtur, line 935: if packageDependenciesHook didn't return dependencies, shouldn't we return the original dependencyLocations instead? If the intention was to erase them, it'd be by explicitly setting dependencies field to an empty object (or even undefined, but if the hook returns undefined that's very implicit)

packages/compartment-mapper/src/bundle-lite.js
boneskull: What is incorrect? The actual change or the behavior itself?

packages/compartment-mapper/src/types/compartment-map-schema.ts
boneskull: Yes, this can be removed

packages/compartment-mapper/src/compartment-map.js
boneskull: The types are mirroring the system as it's written, so that's why. 😄   Certain module descriptor configuration types are allowed in certain types of compartment descriptors, and others are not; the compartment descriptor subtypes document this.

packages/compartment-mapper/src/policy.js
boneskull: The intent here is that if `options` is provided and `options.policy` is provided, you are guaranteed a `SomePackagePolicy` return type.

If you do not provide `options` or do not provide `options.policy`, we cannot guarantee that, and you may get `undefined`.

packages/compartment-mapper/src/policy.js
boneskull: Yeah, I think you're right.  I'll fix that.  But _only_ for `entry`.  I'm...not entirely sure that's _correct_, as it implies:

```js
{
  builtins: 'any',
  globals: 'any',
  packages: 'any'
}
```

...which may be more appropriate as the default `PackagePolicy` if `entry` is `undefined`.

packages/compartment-mapper/src/node-modules.js
boneskull: I'll note this in its docstring.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -759,19 +844,101 @@ const graphPackages = async (
...
+      const userResult = packageDependenciesHook(packageDependenciesHookInput);
+      // If user hook also returned a result, use it (overrides default)
+      if (userResult?.dependencies) {
+        packageDependenciesHookResult = userResult;
+      }
+    }
+
+    // if "dependencies" are in here, then something changed the list.
+    if (packageDependenciesHookResult?.dependencies) {
+      const { size } = packageDependenciesHookResult.dependencies;
```
boneskull: This should probably not be a destructuring assignment.

packages/compartment-mapper/src/node-modules.js
boneskull: If packageDependenciesHook didn't return dependencies, we don't get to this branch, because of the conditional on L899.

naugtur: now I'm even more confused. So when _do_ we get to this branch?

packages/compartment-mapper/src/policy.js
naugtur: It's about policy conventions in general, where by default the entry to your program has access to all endowments. 
no policy for entry means everything. no policy for anything else means nothing.

I'm ok with deliberately changing to consistently nothing means nothing, but that'd be a breaking change at this point.

packages/compartment-mapper/src/compartment-map.js
naugtur: internal DX feedback - I didn't get the context when I needed it. (skill issue) ;)

packages/compartment-mapper/src/bundle-lite.js
naugtur: The behavior itself seems wrong to me. It's effectively banning a module like this:

```js
require('somevaliddependency')

function a() {
  const require = (x)=> {
    if (!x) { throw Error('x is required')
  }
  
  require('not a package but a string')
}
```

packages/compartment-mapper/src/import-hook.js
naugtur: This is now my only blocker to this whole PR.
Do away with the hook for deferredError (and update the diagram in hooks.md or lmk) and it's a LGTM:shipit::rocket:

packages/compartment-mapper/test/make-import-hook-maker.test.js
```diff
@@ -0,0 +1,156 @@
...
+  const options = {
+    moduleSourceHook,
+    parserForLanguage: defaultParserForLanguage,
+  };
+
+  const args = await argsFn(
+    readPowers,
+    /** @type {string} */ (testFixture.entrypoint),
+    options,
+  );
```
naugtur, line 78: (nit) feels like having a single function `doTheLoading` that takes the argsFn arguments and puts things together in one pass would be easier to follow. And the testcase code would resembl the expected usage pattern of the `mapNodeModules` and `captureFromMap` more closely.

packages/compartment-mapper/src/hooks.md
```diff
@@ -0,0 +1,115 @@
...
+| `packageDataHook`          | Receives all found package descriptors data before graph translation.                       |
+| `packageDependenciesHook`  | Allows dynamic mutation of dependencies during node_modules graph translation.              |
+| `unknownCanonicalNameHook` | Called when the policy references unknown canonical names, can suggest typos/similar names. |
+| `moduleSourceHook`         | Invoked when a module source is created.                                                    |
+| `packageConnectionsHook`   | Surfaces connections during digest. (ignored in archiving)                                  |
+
+[Type declarations for the hooks](./types/external.ts)
+
+```mermaid
+graph TB
```
naugtur, line 18:
```suggestion

graph TB
```

A newline here is needed for some MD renderers 🙈

packages/compartment-mapper/src/node-modules.js
boneskull: We get to the branch when the `packageDependenciesHook` returns a `dependencies` `Set`. If it returns an empty `dependencies` `Set`, then we will remove all dependencies. This would be the only way to do so, if needed.

boneskull:
@naugtur I've considered/addressed your non-nitpicky comments (and some of the nitpicky ones)

packages/compartment-mapper/src/policy.js
```diff
@@ -166,6 +155,7 @@ export const makePackagePolicy = (label, { policy } = {}) => {
       };
     } else if (label === ENTRY_COMPARTMENT) {
       packagePolicy = policy.entry;
+      return packagePolicy;
```
naugtur, line 158: (could) add a comment explaining why this is here and what it means.
```suggestion
      // no policy set for entry compartment should allow eveything and not return an empty policy object for enforcement by default, so we need to skip defaulting to an empty object below.
      return packagePolicy;
```

packages/compartment-mapper/src/types/compartment-map-schema.ts
naugtur:
```suggestion
```

packages/compartment-mapper/src/import-hook.js
```diff
@@ -504,9 +554,19 @@ const makeDeferError = (
...
     };
+    packageSources[specifier] = moduleSource;
+
+    if (moduleSourceHook) {
+      moduleSourceHook({
+        moduleSource: { error: moduleSource.deferredError },
+        canonicalName,
+        log,
+      });
+    }
```
naugtur:
```suggestion
```

packages/compartment-mapper/src/import-hook.js
boneskull: We _are_ linking at time of policy generation.  If this error is encountered during policy generation (and I'm not sure how to trigger that, based on your description), would it not fail at runtime anyway?  Is your argument that we should just generate the policy anyhow?

Regardless. I'm not going to die on this hill.

packages/compartment-mapper/src/import-hook.js
```diff
@@ -472,17 +518,21 @@ function* chooseModuleDescriptor(
...
+ * @param {import('./types.js').LogFn} log
  * @returns {DeferErrorFn}
  */
 const makeDeferError = (
   strictlyRequiredForCompartment,
   packageLocation,
   packageSources,
+  moduleSourceHook,
+  canonicalName,
+  log,
```
naugtur: These seem unused now, but somehow linter did not complani

packages/compartment-mapper/src/bundle-lite.js
naugtur: please add an issue to look into this and mention @kriskowal in it too.

naugtur:
> @naugtur For some reason I thought that doc was outdated and reflected the old system.

It was only outdated by the commit that removed a hook call in deferError

PR #2989, opened 2025-10-10
feat(harden): Surface harden before lockdown

kriskowal:
Refs: #2983

## Description

This change introduces an `@endo/harden` package that has three modes, as documented in the new README.

**post-lockdown** All of these modes adopt the behavior of the global or intrinsic `harden` after `lockdown`. All of these modes cause `lockdown` to fail if `harden` is used before `lockdown`.

**presume post-lockdown** In order to minimize bundle size increase, the `-C hardened` bundler condition can omit a user code implementation of `harden` and rely on the global environment, failing if the global environment lacks a `harden` global or intrinsic.

**shallow** The default behavior before lockdown is to transitively freeze own properties and to not traverse prototype chains. This makes the mode suitable for non-lockdown environments that tend to make uncoordinated changes to shared intrinsics, but preserves a sense of hardening for shallow immutability.

**unsafe** In order to use Hardened Modules without the performance penalty of even a shallow freeze, electing `-C harden:unsafe` will provide a fake `harden` implementation that freezes nothing.

### Security Considerations

Each of these modes provides different security footing.

### Scaling Considerations

Build conditions are provided to compensate for difficulties scaling, both for performance and bundle size increases consequent on adopting `@endo/harden` for a hardened module.

The new package’s README is comprehensive.

### Testing Considerations

All modes are covered. The shortfall in coverage mostly pertains to shimmed behavior like `hasOwn`.

Adopting `@endo/harden` will, by itself, increase bundle sizes. We consider it acceptable to obligate applications with sensitive bundle sizes to also adopt `-C hardened` in their build procedures to mitigate the increase.

This overall effect of this change is to increase the portability of Hardened modules with frontend and testing environments that require uncoordinated modification to shared intrinsics.

### Upgrade Considerations

None.

### DIFF
--- a/packages/harden/CHANGELOG.md
+++ b/packages/harden/CHANGELOG.md
@@ -1,0 +1,6 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
+See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
+
+
--- a/packages/harden/LICENSE
+++ b/packages/harden/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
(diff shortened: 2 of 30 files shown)


kriskowal:
Note to self: when I pick this up again, @erights and I are in agreement to use the term “surface” rather than “shallow”, as opposed to “volume”.

kriskowal:
Superseded https://github.com/endojs/endo/pull/3008

PR #2990, opened 2025-10-10
fix(pass-style): Relax passable error criteria in hardenTaming unsafe mode.

kriskowal:
This change increases test coverage over lockdown with hardenTaming unsafe. The increase in test coverage revealed the need to relax certain cases, but moreof adjust `passStyleOf` to repair unfrozen V8 error objects in `unsafe` `hardenTaming` mode.

### DIFF
--- a/packages/pass-style/NEWS.md
+++ b/packages/pass-style/NEWS.md
@@ -3,6 +3,10 @@ User-visible changes in `@endo/pass-style`:
 # Next release
 
 - deprecates `assertChecker`. Use `Fail` in the confirm/reject pattern instead, as supported by `@endo/errors/rejector.js`.
+- Enables `passStyleOf` to make errors passable as a side-effect when SES locks
+  down with `hardenTaming` set to `unsafe`, which impacts errors on V8 starting
+  with Node.js 21, and similar engines, that own a `stack` getter and setter
+  that would otherwise be repaired as a side-effect of `harden`.
 
 # 1.6.3 (2025-07-11)
 


packages/pass-style/src/error.js
```diff
@@ -108,18 +108,24 @@ export const confirmRecursivelyPassableErrorPropertyDesc = (
...
-    return (
-      reject &&
-      reject`Passable Error ${q(
-        propName,
-      )} own property must be a data property: ${desc}`
-    );
+    // Error.prototype.stack is own accessor in V8, but nowhere else to our
+    // knowledge (2025).
+    // We relax validation for "stack" only when harden.isFake.
+    if (!(harden.isFake && propName === 'stack')) {
```
mhofman: I fail to see how unsafe harden is related to error stacks. I think what we possibly need to do is allow known "platform own stack accessors" in pass-style. It would be nice to only allow these if there was no lockdown, or if lockdown didn't tame errors.

kriskowal: I’ll think more next week about sensing the effects of error taming variants. I’ll probe your mind next week about “platform own stack accessors” because my understanding is that expectations about the identity or uniqueness of error own get and set may be evolving in a way that will make it harder to sense whether they are the platform’s intrinsic get and set. I fear we may truly need to leave a pinhole for `stack` in pass-style, generally, and simply be careful to discriminate them with `isError` and time we destructure all properties of an object.

packages/common/test/object-meta-map.test.js
```diff
@@ -4,6 +4,9 @@ import { objectMetaMap } from '../object-meta-map.js';
 
 const { getOwnPropertyDescriptors, getPrototypeOf } = Object;
 
+// @ts-expect-error isFake is a secret.
```
erights: WAT? If it is a secret, it is a very poorly kept secret!

(Please avoid ever seeming to make a security claim that we're not actually making.)

packages/exo/test/heap-classes.test.js
```diff
@@ -9,6 +9,9 @@ import {
   makeExo,
 } from '../index.js';
 
+// @ts-expect-error isFake is a secret property.
```
erights: again

packages/exo/test/heap-classes.test.js
```diff
@@ -280,37 +283,41 @@ test('raw guards', t => {
...
+      t.true(hardenIsFake || Object.isFrozen(obj));
       return { ...obj };
     },
     passthrough(obj) {
-      t.is(Object.isFrozen(obj), false);
+      t.false(Object.isFrozen(obj));
       return obj;
     },
     tortuous(hardA, softB, hardC, optHardD, optSoftE = {}) {
+      // Recall that isFrozen lies with hardenTaming: unsafe
```
erights, line 304: Aren't you fixing that? We should at least vehemently deprecate the existing unsafe-fast.

packages/pass-style/src/error.js
```diff
@@ -8,7 +8,96 @@ import { Fail, q, hideAndHardenFunction } from '@endo/errors';
...
+  hasOwn,
+  entries,
+  freeze,
+} = Object;
+
+const { apply } = Reflect;
+
+const hardenIsFake = () => {
+  // We do not trust isFrozen because lockdown with unsafe hardenTaming replaces
+  // isFrozen with a version that is in cahoots with fake harden.
```
erights, line 25: This is the unsafe-fast case that we're at least deprecating and hopefully removing soon, yes?

packages/pass-style/src/error.js
```diff
@@ -8,7 +8,96 @@ import { Fail, q, hideAndHardenFunction } from '@endo/errors';
...
+  return desc?.writable === true;
+};
+
+const makeRepairError = () => {
+  if (!hardenIsFake()) {
+    return undefined;
+  }
+
+  const er1StackDesc = getOwnPropertyDescriptor(Error('er1'), 'stack');
+  const er2StackDesc = getOwnPropertyDescriptor(TypeError('er2'), 'stack');
```
erights: Just curious: Do you have a particular case in mind that you're defending against with the "redundant" check?

packages/pass-style/src/error.js
```diff
@@ -8,7 +8,96 @@ import { Fail, q, hideAndHardenFunction } from '@endo/errors';
...
+
+  if (
+    er1StackDesc === undefined ||
+    er2StackDesc === undefined ||
+    er1StackDesc.get === undefined
+  ) {
```
erights, line 96: The fact that you wrote `||` means I'm confused what the purpose of the "redundant" check is. I expected something more like

```suggestion
  if (
    (er1StackDesc === undefined &&
    er2StackDesc === undefined) ||
    er1StackDesc.get === undefined
  ) {
```

packages/pass-style/src/error.js
```diff
@@ -8,7 +8,96 @@ import { Fail, q, hideAndHardenFunction } from '@endo/errors';
...
+    return undefined;
+  }
+
+  // We should only encounter this case on v8 because of its problematic
```
erights: And the JS engine MetaMask is using?

packages/pass-style/src/error.js
```diff
@@ -8,7 +8,96 @@ import { Fail, q, hideAndHardenFunction } from '@endo/errors';
...
+  // error own stack accessor behavior.
+  // Note that FF/SpiderMonkey, Moddable/XS, and the error stack proposal
+  // all inherit a stack accessor property from Error.prototype, which is
+  // great. That case needs no heroics to secure.
+  if (
+    // In the v8 case as we understand it, all errors have an own stack
+    // accessor property, but within the same realm, all these accessor
+    // properties have the same getter and have the same setter.
+    // This is therefore the case that we repair.
```
erights: Should also mention somewhere that the upcoming `Error.captureStackTrace` proposal may result in more own stack accessors, on more platforms, that may or may not use the same getter and setter. If we don't think those are a problem, we should explain somewhere why not. But not necessarily here.

packages/pass-style/src/error.js
```diff
@@ -8,7 +8,96 @@ import { Fail, q, hideAndHardenFunction } from '@endo/errors';
...
+    typeof er1StackDesc.get !== 'function' ||
+    er1StackDesc.get !== er2StackDesc.get ||
```
erights: Aha! Now I understand

packages/pass-style/src/error.js
```diff
@@ -8,7 +8,96 @@ import { Fail, q, hideAndHardenFunction } from '@endo/errors';
...
+  ) {
+    // See https://github.com/endojs/endo/blob/master/packages/ses/error-codes/SES_UNEXPECTED_ERROR_OWN_STACK_ACCESSOR.md
+    throw TypeError(
+      'Unexpected Error own stack accessor functions (PASS_STYLE_UNEXPECTED_ERROR_OWN_STACK_ACCESSOR)',
+    );
+  }
+
+  // Otherwise, we have own stack accessor properties that are outside
+  // our expectations, that therefore need to be understood better
+  // before we know how to repair them.
```
erights: Should this comment be moved to immediately above the above `throw`?

packages/pass-style/src/error.js
```diff
@@ -178,6 +267,9 @@ export const confirmRecursivelyPassableError = (
       reject`Passable Error must inherit from an error class .prototype: ${candidate}`
     );
   }
+  if (repairError !== undefined) {
+    repairError(candidate);
+  }
```
erights, line 342: Comment here would be good about why we're ok with this special case side effect in what should be a pure query function.

packages/common/test/object-meta-map.test.js
kriskowal: This is a pattern copied from elsewhere. It is not a secret in this sense. It is a secret in the sense that we do not make its existence evident in the type, in order to avoid burdening common users with undue knowledge of its existence. Recommendation welcome.

packages/pass-style/src/error.js
kriskowal: I don’t. This is copied from SES.

packages/common/test/object-meta-map.test.js
kriskowal: How about, “Existence of harden.isFake is not communicated by the types.”?

packages/exo/test/heap-classes.test.js
kriskowal: Yes, I expect this to change to accommodate the new mode. We may need to expose the real `isFrozen` somewhere to make this test work in every mode.

packages/exo/test/heap-classes.test.js
```diff
@@ -280,37 +283,41 @@ test('raw guards', t => {
       return 'hello';
     },
     rawIn(obj) {
-      t.is(Object.isFrozen(obj), false);
+      t.false(Object.isFrozen(obj));
       return obj;
     },
     rawOut(obj) {
-      t.is(Object.isFrozen(obj), true);
+      t.true(hardenIsFake || Object.isFrozen(obj));
```
mhofman: Curious why this is the only `Object.isFrozen` case where we have to condition on `hardenIsFake`

packages/pass-style/src/error.js
```diff
@@ -8,7 +8,103 @@ import { Fail, q, hideAndHardenFunction } from '@endo/errors';
...
+  const desc = getOwnPropertyDescriptor(subject, 'x');
+  return desc?.writable === true;
+};
+
+const makeRepairError = () => {
+  if (!hardenIsFake()) {
+    return undefined;
+  }
+
+  const er1StackDesc = getOwnPropertyDescriptor(Error('er1'), 'stack');
```
mhofman: I'm curious, why don't we build an error by syntax instead for one of the candidates, to be resilient against global `Error` override?

packages/pass-style/src/error.js
```diff
@@ -8,7 +8,103 @@ import { Fail, q, hideAndHardenFunction } from '@endo/errors';
...
+    // them.
+    // See https://github.com/endojs/endo/blob/master/packages/ses/error-codes/SES_UNEXPECTED_ERROR_OWN_STACK_ACCESSOR.md
+    throw TypeError(
+      'Unexpected Error own stack accessor functions (PASS_STYLE_UNEXPECTED_ERROR_OWN_STACK_ACCESSOR)',
+    );
+  }
+
+  // We should otherwise only encounter this case on V8 and possibly immitators
+  // like FaceBook's Hermes because of its problematic error own stack accessor
+  // behavior.
```
mhofman: Should we mention the capability leak that we're trying to protect against?

packages/pass-style/src/error.js
```diff
@@ -8,7 +8,103 @@ import { Fail, q, hideAndHardenFunction } from '@endo/errors';
...
+      });
+    }
+  };
+  harden(repairError);
+
+  return repairError;
+};
+harden(makeRepairError);
+
+const repairError = makeRepairError();
```
mhofman: I'm wondering if it might not be too early to make this. Any way to delay until we `makePassStyleOf` ?

packages/pass-style/src/error.js
```diff
@@ -178,6 +274,16 @@ export const confirmRecursivelyPassableError = (
       reject`Passable Error must inherit from an error class .prototype: ${candidate}`
     );
   }
+  if (repairError !== undefined) {
+    // This point is unreachable unless the candidate is mutable and the
```
mhofman, line 334: How do we know the object is mutable here? I mean it's possible that fake harden didn't freeze but that someone actually explicitly froze it.

packages/pass-style/test/errors.test.js
mhofman, line 1: I would like to see a test where an error is manually frozen (without using possibly fake harden), and that all cases show that the error is not passable under v8.

mhofman (commented):
Looks good, but would love to see an explicit test for the can't patch error case (either by harden or by passStyleOf).

Also wondering about how lazy the `makeRepairError` should be.

packages/pass-style/src/error.js
kriskowal: This is largely copied from SES where the constraints are admittedly different. In SES, we are explicitly vulnerable to scripts that run earlier.

For sake of rigor, thinking aloud: passStyleOf can and often does run in a Compartment, and the globalThis is typically (always with importBundle) but not necessarily frozen. So, passStyleOf can be deceived by the creator of the compartment in the base case, but all compartments are vulnerable to their creator. Any multi-tenant compartment should have a frozen `globalThis`, which obviates the concern of interference from within the compartment. So, I think this approach is defensible.

However, a belt-and-suspenders defense would also be welcome, in my opinion. I know half a dozen ways to get a TypeError instance reliably by syntax, but don’t know any way to get a plain `Error` instance from syntax alone. Maybe we can rely on @gibson042’s Encyclopedia 262.

packages/pass-style/src/error.js
kriskowal: It is possible. The threading is not trivial, and I could turn it around in a day.

packages/exo/test/heap-classes.test.js
kriskowal: I count 9 total. 5 affirmative. 4 negative. Perhaps you mean something else?

kriskowal (commented):
Pushed a commit to explicitly test frozen-not-hardened errors are unfixable on pathological-V8 (but fine on prior V8 versions we still test).

packages/pass-style/test/errors.test.js
```diff
@@ -80,6 +83,29 @@ test('toPassableError, toThrowable', t => {
...
+    t.throws(() => passStyleOf(e1), {
+      message: /^Passable Error "stack" own property must be a data property:/,
+    });
+  } else {
+    t.is(passStyleOf(e1), 'error');
+  }
+});
+
+test('passStyleOf hardened (albeit fake hardened) error adapts to pathological V8', t => {
+  const e = harden(new Error('that which is frozen but not hardened'));
```
mhofman: Nit: mismatching message

packages/exo/test/heap-classes.test.js
mhofman: I see some `Object.freeze` calls above in the same test that don't get the same treatment, and arguably I didn't look closely at what the test does, I just saw those lines modified, but not get an extra `hardenIsFake` condition, and was just wondering why.

packages/pass-style/src/error.js
kriskowal: I’ve posted a follow-up commit that introduces `makeTypeError` with a diatribe about wearing a belt and suspenders on overalls. ;-)

mhofman: I do not see such a commit

kriskowal:
I believe I’ve addressed all feedback. Commits added, none changed, for incremental review.

packages/pass-style/src/error.js
```diff
@@ -28,13 +28,67 @@ const hardenIsFake = () => {
...
+ * As such, it is irrelevant that the globalThis is mutable and also holds all
+ * of the host's authority.
+ *
+ * Pass-style may be loaded into a guest compartment, and the globalThis of the
+ * compartment may or may not be frozen.
+ * We typically, as with importBundle, run every Node.js package in a dedicated
+ * compartment with a gratuitiously frozen globalThis.
+ * In this configuration, we can rely on globalThis.Error and
+ * globalThis.TypeError to correspond to the realm's intrinsics, either because
+ * the Comaprtment arranged for a frozen globalThis or because the pass-style
```
mhofman:
```suggestion
 * the Compartment arranged for a frozen globalThis or because the pass-style
```

packages/pass-style/src/error.js
```diff
@@ -8,7 +8,157 @@ import { Fail, q, hideAndHardenFunction } from '@endo/errors';
...
+    return undefined;
+  }
+
+  const er1StackDesc = getOwnPropertyDescriptor(Error('er1'), 'stack');
+  const er2StackDesc = getOwnPropertyDescriptor(makeTypeError(), 'stack');
+
+  if (
+    er1StackDesc === undefined ||
+    er2StackDesc === undefined ||
+    er1StackDesc.get === undefined
```
mhofman: Can we inverse the role of `er1` and `er2` so that we only bypass repair if the error constructed syntactically doesn't have an accessor?

My concern is with a global `Error` constructor that would "touch" the `stack` and transform it into a data property while errors created by the engine directly would still require repairing. Of course it's safe to forget to repair such errors, but I'd rather find out early without such an error trying to be passed.

packages/pass-style/src/error.js
kriskowal: https://github.com/endojs/endo/pull/2990/commits/714703cae3dbf91c8747df9997a325a7cb723a01

packages/pass-style/src/passStyleOf.js
```diff
@@ -38,6 +39,8 @@ import { assertPassableString } from './string.js';
 const { ownKeys } = Reflect;
 const { isFrozen, getOwnPropertyDescriptors, values } = Object;
 
+const repairError = makeRepairError();
```
mhofman: So this still causes `makeRepairError` to still run at import of `@endo/pass-style`, but from what I gather, for the use of the `toPassableError` export, which unlike `passStyleOf`, is always exported anew.

I think the pass-style package has grown into a weird monster of instantiable and instantiated exports, both exposed on the same entry-point. Because of this I'm tempted to ask to revert my previous request for a lazy `repairError`, as it's pointless with the current structure of this package. Sorry about the back and forth, I forgot how unwieldy this whole thing is.

packages/pass-style/NEWS.md
```diff
@@ -3,6 +3,10 @@ User-visible changes in `@endo/pass-style`:
 # Next release
 
 - deprecates `assertChecker`. Use `Fail` in the confirm/reject pattern instead, as supported by `@endo/errors/rejector.js`.
+- Enables `passStyleOf` to make errors passable as a side-effect when SES locks
+  down with `hardenTaming` set to `unsafe`, which impacts errors on V8 starting
+  with Node.js 20, and similar engines, that own a `stack` getter and setter
```
mhofman: Node 20 actually isn't affected. Only 21+

packages/pass-style/src/error.js
```diff
@@ -87,22 +87,22 @@ export const makeRepairError = () => {
...
+  const typeErrorStackDesc = getOwnPropertyDescriptor(makeTypeError(), 'stack');
+  const errorStackDesc = getOwnPropertyDescriptor(Error('obligatory'), 'stack');
 
   if (
-    er1StackDesc === undefined ||
-    er2StackDesc === undefined ||
-    er1StackDesc.get === undefined
+    typeErrorStackDesc === undefined ||
+    typeErrorStackDesc.get === undefined ||
+    errorStackDesc === undefined
```
mhofman: Actually thinking of it, I would prefer to move the `errorStackDesc === undefined` into the next `if` statement.

packages/pass-style/src/error.js
mhofman: As discussed, trying to delay was not feasible and a bad idea.

packages/exo/test/heap-classes.test.js
```diff
@@ -231,14 +234,14 @@ const PassableGreeterI = M.interface(
...
-      t.is(Object.isFrozen(immutabe), true);
+      t.true(Object.isFrozen(immutabe));
       return 'hello';
     },
   });
 
   const mutable = {};
   t.is(greeter.sayHello(mutable), 'hello', `passableGreeter can sayHello`);
-  t.is(Object.isFrozen(mutable), true, `mutable is frozen`);
+  t.true(Object.isFrozen(mutable), `mutable is frozen`);
```
kriskowal, line 244: This case relies on `harden` and `isFrozen` to be in cahoots. This passes with unsafe harden taming because isFrozen is lying. The safe harden taming passes because harden, called as a side-effect of passing `mutable` to an Exo, is effective. This will continue to work without modification when we introduce surface hardening. If we obviate unsafe harden taming by making an unsafe mode of `@endo/harden` which doesn’t collude with `isFrozen`, this test will need to be adjusted to take that into account.

packages/exo/test/heap-classes.test.js
```diff
@@ -280,37 +283,41 @@ test('raw guards', t => {
       return 'hello';
     },
     rawIn(obj) {
-      t.is(Object.isFrozen(obj), false);
+      t.false(Object.isFrozen(obj));
```
kriskowal: This test passes because the `M.raw` guard is bypassing the `harden` otherwise implied. It’s always passed an unfrozen/unhardened object in the tests.

packages/exo/test/heap-classes.test.js
kriskowal: Your intuition was onto something, anyway. I went through each of these cases again and found 4 that did not require the relaxation for `harden.isFake`, left over from when I relaxed these tests to accommodate the unsafe mode of `@endo/harden` where `isFrozen` is not rigged. Will have to revisit if we go so far.

packages/exo/test/heap-classes.test.js
```diff
@@ -280,37 +283,41 @@ test('raw guards', t => {
...
       return obj;
     },
     rawOut(obj) {
```
kriskowal, line 291: `rawOut` has an `M.any()` guard, so the argument is implicitly hardened, albeit fake hardened.  The relaxation is unnecessary in this case and I will restore the case.

packages/exo/test/heap-classes.test.js
```diff
@@ -280,37 +283,41 @@ test('raw guards', t => {
...
-      t.is(Object.isFrozen(softB), false);
+      t.true(hardenIsFake || Object.isFrozen(hardA));
+      t.false(!hardenIsFake && Object.isFrozen(softB));
       softB.b = 2;
-      t.is(Object.isFrozen(hardC), true);
-      t.is(Object.isFrozen(optHardD), true);
-      t.is(Object.isFrozen(optSoftE), false);
+      t.true(Object.isFrozen(hardC));
+      t.true(Object.isFrozen(optHardD));
+      t.false(!hardenIsFake && Object.isFrozen(optSoftE));
```
kriskowal: This case requires a relaxation because isFrozen is misreporting a soft object has having been frozen. It’s either a soft object that was passed through a raw optional arg, or defaulted to a soft object.

Issue #2992, opened 2025-10-15
Override mistake in Iterator


naugtur:
## Describe the bug

Packages in the ecosystem override the iterator methods on custom objects with `Iterator.prototype` as their prototype.

```
Cannot assign to read only property \'Symbol(Symbol.iterator)\' of object \'[object Object]\'
```

Example: https://socket.dev/npm/package/@rive-app/canvas/files/2.31.6/rive.js#L4191

## Steps to reproduce

tests included

## Expected behavior

override taming moderate and severe should prevent override mistake on `Symbol.iterator` assignment

## Platform environment

all

## Additional context

discovered in https://github.com/MetaMask/metamask-extension/pull/36398

PR #2993, opened 2025-10-15
feat(ses): handle iterator override

naugtur:

Closes: #2992

## Description

Adds an enablement to cover for override mistake on Iterator's `Symbol.iterator` key.

Additionally, added symbols to override test setup in general.

### Security Considerations

None I'm aware of

### Scaling Considerations

no

### Testing Considerations

tests added

compatibility improved

### Upgrade Considerations

Not a breaking change

### DIFF
--- a/packages/ses/src/enablements.js
+++ b/packages/ses/src/enablements.js
@@ -106,6 +106,10 @@ export const moderateEnablements = {
     [iteratorSymbol]: true, // set by mobx generated code (old TS compiler?)
   },
 
+  '%IteratorPrototype%': {
+    [iteratorSymbol]: true, // is sometimes used in custom iterators and generators implementations eg. @rive-app/canvas
+  },
+
   // Function.prototype has no 'prototype' property to enable.
   // Function instances have their own 'name' and 'length' properties
   // which are configurable and non-writable. Thus, they are already
--- a/packages/ses/test/_override-tester.js
+++ b/packages/ses/test/_override-tester.js
@@ -1,7 +1,7 @@
 import {
   getPrototypeOf,
   getOwnPropertyDescriptor,
-  getOwnPropertyNames,
+  ownKeys,
 } from '../src/commons.js';
 
 function getValue(obj, name) {
@@ -11,37 +11,43 @@ function getValue(obj, name) {
 
 export function overrideTester(t, type, obj, allowed = []) {
   const proto = getPrototypeOf(obj);
-  for (const name of getOwnPropertyNames(proto)) {
-    if (name === '__proto__') {
-      t.notThrows(() => {
-        obj[name] = 1;
-      }, `Should not throw when setting property ${name} of ${type} instance`);
-      t.not(
-        getValue(obj, name),
-        1,
-        `Should not allow setting property ${name} of ${type} instance`,
+  for (const key of ownKeys(proto)) {
+    if (key === '__proto__') {
+      t.notThrows(
+        () => {
+          obj[key] = 1;
+        },
+        `Should not throw when setting property ${String(key)} of ${type} instance`,
       );
-    } else if (allowed.includes(name)) {
-      t.notThrows(() => {
-        obj[name] = 1;
-      }, `Should not throw when setting property ${name} of ${type} instance`);
-      t.is(
-        getValue(obj, name),
+      t.not(
+        getValue(obj, key),
         1,
-        `Should allow setting property ${name} of ${type} instance`,
+        `Should not allow setting property ${String(key)} of ${type} instance`,
+      );
+    } else if (allowed.includes(key)) {
+      t.notThrows(
+        () => {
+          obj[key] = 1;
+        },
+        `Should not throw when setting property ${String(key)} of ${type} instance`,
+      );
+      t.is(
+        getValue(obj, key),
+        1,
+        `Should allow setting property ${String(key)} of ${type} instance`,
       );
     } else {
       t.throws(
         () => {
-          obj[name] = 1;
+          obj[key] = 1;
         },
         undefined,
-        `Should throw when setting property ${name} of ${type} instance`,
+        `Should throw when setting property ${String(key)} of ${type} instance`,
       );
       t.not(
-        getValue(obj, name),
+        getValue(obj, key),
         1,
-        `Should not allow setting property ${name} of ${type} instance`,
+        `Should not allow setting property ${String(key)} of ${type} instance`,
       );
     }
   }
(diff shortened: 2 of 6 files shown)


packages/ses/test/_override-tester.js
```diff
@@ -4,14 +4,20 @@ import {
   getOwnPropertyNames,
 } from '../src/commons.js';
 
+const { getOwnPropertySymbols } = Object;
```
naugtur: Should this go into commons if it's only ever used in tests?

packages/ses/src/enablements.js
```diff
@@ -106,6 +106,11 @@ export const moderateEnablements = {
     [iteratorSymbol]: true, // set by mobx generated code (old TS compiler?)
   },
 
+  '%IteratorPrototype%': {
+    toString: true,
```
naugtur: should toString be listed here?

erights: Only if Iterator.prototype has an *own* toString method, which it does not in Node 22.16. But someone should check the spec.

packages/ses/src/enablements.js
```diff
@@ -106,6 +106,11 @@ export const moderateEnablements = {
...
+    [iteratorSymbol]: true, // set by mobx generated code (old TS compiler?)
```
erights: Does this comment reflect the concrete case you found? If not, please revise.

packages/ses/test/_override-tester.js
erights: On a case by case basis. For `getOwnPropertySymbols`, seems fine to put into commons.

packages/ses/test/_override-tester.js
```diff
@@ -4,14 +4,20 @@ import {
...
+
 function getValue(obj, name) {
   const desc = getOwnPropertyDescriptor(obj, name);
   return desc && desc.value;
 }
 
+function getOwn(obj) {
+  return [...getOwnPropertyNames(obj), ...getOwnPropertySymbols(obj)];
+}
```
erights: How does this differ from `Reflect.ownKeys`?

packages/ses/test/_override-tester.js
```diff
@@ -4,14 +4,20 @@ import {
...
+
 export function overrideTester(t, type, obj, allowed = []) {
   const proto = getPrototypeOf(obj);
-  for (const name of getOwnPropertyNames(proto)) {
+  for (const name of getOwn(proto)) {
```
erights:
```suggestion
  for (const key of ownKeys(proto)) {
```

packages/ses/test/_override-tester.js
```diff
@@ -24,24 +30,24 @@ export function overrideTester(t, type, obj, allowed = []) {
     } else if (allowed.includes(name)) {
       t.notThrows(() => {
         obj[name] = 1;
-      }, `Should not throw when setting property ${name} of ${type} instance`);
+      }, `Should not throw when setting property ${name.toString()} of ${type} instance`);
```
erights:
```suggestion
      }, `Should not throw when setting property ${String(key)} of ${type} instance`);
```

And likewise below

packages/ses/test/enable-property-overrides-default.test.js
```diff
@@ -12,7 +12,12 @@ test('enablePropertyOverrides - on', t => {
   overrideTester(t, 'Object', {}, ['toString', 'valueOf']);
   // We allow 'length' *not* because it is in enablements; it is not;
   // but because each array instance has its own.
-  overrideTester(t, 'Array', [], ['toString', 'length', 'push', 'concat']);
+  overrideTester(
+    t,
+    'Array',
+    [],
+    ['toString', 'length', 'push', 'concat', Symbol.iterator],
```
erights, line 19: Huh. This PR is not the one that adds the enablement for `Array.prototype[Symbol.iterator]`. So how did this test pass before you added it to the test?

packages/ses/test/enable-property-overrides-severe-debug.test.js
```diff
@@ -16,7 +16,12 @@ test('enablePropertyOverrides - on, with debug', t => {
 
   // We allow 'length' *not* because it is in enablements; it is not;
   // but because each array instance has its own.
-  overrideTester(t, 'Array', [], ['toString', 'length', 'push', 'concat']);
+  overrideTester(
+    t,
+    'Array',
+    [],
+    ['toString', 'length', 'push', 'concat', Symbol.iterator],
```
erights, line 23: Likewise

packages/ses/test/_override-tester.js
naugtur: Good point. It doesn't. I had an implementation that consisted of two loops - one over names, another over symbols and this is the result of a refactor.

packages/ses/test/enable-property-overrides-default.test.js
naugtur: The tests stopped passing once I got overrideTester to also look at symbol keys.

packages/ses/test/_override-tester.js
```diff
@@ -11,37 +11,40 @@ function getValue(obj, name) {
...
       t.not(
-        getValue(obj, name),
+        getValue(obj, key),
         1,
-        `Should not allow setting property ${name} of ${type} instance`,
+        `Should not allow setting property ${key} of ${type} instance`,
+      );
+    } else if (allowed.includes(key)) {
+      t.notThrows(
+        () => {
```
naugtur, line 29: prettier changed its mind on newlines, I didn't do it.

packages/ses/src/enablements.js
naugtur: nevermind, it was already covered in min

packages/ses/src/enablements.js
naugtur: fixed

packages/ses/test/_override-tester.js
```diff
@@ -11,37 +11,40 @@ function getValue(obj, name) {
...
   const proto = getPrototypeOf(obj);
-  for (const name of getOwnPropertyNames(proto)) {
-    if (name === '__proto__') {
+  for (const key of ownKeys(proto)) {
+    if (key === '__proto__') {
       t.notThrows(() => {
-        obj[name] = 1;
-      }, `Should not throw when setting property ${name} of ${type} instance`);
+        obj[key] = 1;
+      }, `Should not throw when setting property ${key} of ${type} instance`);
```
erights:
```suggestion
      }, `Should not throw when setting property ${String(key)} of ${type} instance`);
```

packages/ses/test/_override-tester.js
```diff
@@ -11,37 +11,40 @@ function getValue(obj, name) {
...
       t.not(
-        getValue(obj, name),
+        getValue(obj, key),
         1,
-        `Should not allow setting property ${name} of ${type} instance`,
+        `Should not allow setting property ${key} of ${type} instance`,
```
erights:
```suggestion
        `Should not allow setting property ${String(key)} of ${type} instance`,
```

packages/ses/test/_override-tester.js
naugtur: Fun fact, these were literally obscured by a "helpful" popup in the IDE when I was adding missing `String()` calls previously

PR #3000, opened 2025-10-30
wip: Function constructor evaluator

naugtur:
ported the code from an old experiment within LavaMoat webpack plugin
It's still very broken in tests, but for what seems like minor features missing :sweat_smile: 

How many things (other than most tests) rely on `evaluate` returning the value of the last statement in code?

**Intentions**

The goal here is to put together a PoC that can later be extracted into an evaluator factory that is used under Hermes (and potentially other cases where it's tempting and globalThis can be made non-extensible - eg. Node.js)

I'm not intending to replace the original entirely, but allow to opt-in to it. 
I'm not implementing the opt-in part because I want to minimize work that's necessary to test the PoC and compare with the original implementation.

### DIFF
--- a/packages/ses/src/make-evaluate.js
+++ b/packages/ses/src/make-evaluate.js
@@ -1,24 +1,28 @@
 // @ts-check
 
-import { FERAL_FUNCTION, arrayJoin, apply } from './commons.js';
-import { getScopeConstants } from './scope-constants.js';
+import {
+  FERAL_FUNCTION,
+  apply,
+  getOwnPropertyNames,
+  // preventExtensions,
+  arrayJoin,
+  Set,
+  globalThis,
+} from './commons.js';
 
-/**
- * buildOptimizer()
- * Given an array of identifiers, the optimizer returns a `const` declaration
- * destructuring `this.${name}`.
- *
- * @param {Array<string>} constants
- * @param {string} name
- */
-function buildOptimizer(constants, name) {
-  // No need to build an optimizer when there are no constants.
-  if (constants.length === 0) return '';
-  // Use 'this' to avoid going through the scope proxy, which is unnecessary
-  // since the optimizer only needs references to the safe global.
-  // Destructure the constants from the target scope object.
-  return `const {${arrayJoin(constants, ',')}} = this.${name};`;
-}
+// // This should be necessary for sealing the environment, but for now it fails tests
+// let tried = false;
+// const attemptMakingGlobalThisNonExtensible = () => {
+//   if (tried) {
+//     return;
+//   }
+//   try {
+//     preventExtensions(globalThis);
+//   } catch (e) {
+//     /* empty */
+//   }
+//   tried = true;
+// };
 
 /**
  * makeEvaluate()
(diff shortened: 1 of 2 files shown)


packages/ses/src/make-evaluate.js
```diff
@@ -31,80 +35,38 @@ function buildOptimizer(constants, name) {
...
 
-  return apply(evaluateFactory, context, []);
+    return apply(
+      FERAL_FUNCTION(
+        `{${SCOPE_TERMINATOR_KEYS}}`,
+        `{${MODULE_LEXICAL_KEYS}}`,
+        `return (function () {
+          'use strict';
+          ${code}
+        }).bind(this)();`,
```
naugtur, line 66: simply prepending `'use strict';` to `code` sadly doesn't work because of the destructuring of arguments

`SyntaxError: Illegal 'use strict' directive in function with non-simple parameter list `

packages/ses/src/make-evaluate.js
```diff
@@ -31,80 +35,38 @@ function buildOptimizer(constants, name) {
...
-              return eval(arguments[0]);
-            };
-          }
-        }
-      }
-    }
-  `);
+    context.evalScope.eval; // required by make-safe-evaluator.
 
-  return apply(evaluateFactory, context, []);
```
mhofman, line 109: I think we haven't used the evaluate factory as a standalone thing in a while. I guess this removes the concept completely.

packages/ses/src/make-evaluate.js
```diff
@@ -31,80 +35,38 @@ function buildOptimizer(constants, name) {
...
+    return apply(
+      FERAL_FUNCTION(
+        `{${SCOPE_TERMINATOR_KEYS}}`,
+        `{${MODULE_LEXICAL_KEYS}}`,
+        `return (function () {
+          'use strict';
+          ${code}
```
mhofman, line 65: This means that `code` be allowed to utter `new.target` (`undefined`), which may be an acceptable fidelity concern.

I think `super` would still be illegal?

packages/ses/src/make-evaluate.js
```diff
@@ -31,80 +35,38 @@ function buildOptimizer(constants, name) {
...
-        }
-      }
-    }
-  `);
+    context.evalScope.eval; // required by make-safe-evaluator.
 
-  return apply(evaluateFactory, context, []);
+    return apply(
+      FERAL_FUNCTION(
+        `{${SCOPE_TERMINATOR_KEYS}}`,
```
mhofman, line 61: I'm confused, how does it work for changing keys of the global object? Either the code changing a pre-existing global binding `foo = 42`, or someone changing an existing property of the global object and expecting the global binding to be updated. `globalThis.bar = 42; console.log(bar)`.

naugtur:
@mhofman Clarification of intentions added to the description

packages/ses/src/make-evaluate.js
naugtur: For new keys, `globalThis` is intended to be made non-extensible (but it was breaking tests even further so I commented it out)

as for updating existing fields... I forgot about that issue, but I have a vague memory of a solution for that existing. Did I hallucinate that?

packages/ses/src/make-evaluate.js
naugtur: Is that a concern on a dummy empty function?

Was `eval` the thing that prevented these issues before?

packages/ses/src/make-evaluate.js
naugtur: that's not a problem. We can split into steps by adding some complexity if need be.

packages/ses/src/make-evaluate.js
mhofman: Actually I think because of our [direct eval](https://tc39.es/ecma262/#sec-performeval) in a function scope, `new.target` being allowed is an existing fidelity concern, however undocumented I believe.

packages/ses/src/make-evaluate.js
mhofman: Not a concern, just an observation. I added re-used of evaluator functions a while ago, so re-use of factory is pretty much not needed anymore.

packages/ses/src/make-evaluate.js
mhofman: There is currently no way to hook into scope binding access (without `this`). Someday we may have [`let` decorators](https://github.com/tc39/proposal-decorators/blob/master/EXTENSIONS.md#let-decorators) that would do that. The [refs proposal](https://github.com/rbuckton/proposal-refs) might also be able to do something here. Both of those would still require a non-extensible global object, but at least the bindings could be mutable (mutated either by the evaluated code, or reflect mutation of the source)

naugtur: We're talking about propagating updates between multiple evaluations within one compartment here I assume. So if we only ever had one, that would not be a problem. It'd be hard to pull off, but theoretically possible with a bundler.

Otherwise, manual propagation of individual globals seems to be the only option. 
A code snippet that reads specified local variables and passes them on to a callback on demand could work.

mhofman: > We're talking about propagating updates between multiple evaluations within one compartment here I assume.

Or within a single evaluation too. If `bar` exist on the global object and is writable/configurable, code would expect that doing `globalThis.bar = 42` is immediately visible in calls to `console.log(bar)`.

> It'd be hard to pull off, but theoretically possible with a bundler.

If we're considering code rewrites, I assume the easiest is to transform all binding references that do not resolve to a surrounding scope into some kind of call access. E.g. `console.log(bar)` -> `console.log(__getGlobal__('bar'))` and `bar = 42;` -> `__setGlobal__('bar', 42);`.

PR #3002, opened 2025-11-01
refactor(ses-ava): Remove use of ambient process from main

gibson042:
Refs: #2987

## Description

Update the signature of `main` to (args: string[]) => Promise\<ExitCode>, removing access to ambient `process.argv` and `process.exit`.

### DIFF
--- a/packages/ses-ava/bin/ses-ava.cjs
+++ b/packages/ses-ava/bin/ses-ava.cjs
@@ -1,8 +1,13 @@
 #!/usr/bin/env node
 (async () => {
   const { main } = await import('../src/command.js');
-  await main(process.argv.slice(2));
-})().catch(error => {
-  console.error(error);
-  process.exitCode = 1;
-});
+  return main(process.argv.slice(2));
+})().then(
+  exitCode => {
+    process.exitCode ||= exitCode;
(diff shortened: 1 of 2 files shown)


packages/ses-ava/bin/ses-ava.cjs
```diff
@@ -1,8 +1,13 @@
...
+  return main(process.argv.slice(2));
+})().then(
+  exitCode => {
+    process.exitCode ||= exitCode;
+  },
+  error => {
+    console.error(error);
+    process.exitCode ||= 1;
+  },
+);
```
kriskowal, line 13: Aside: there’s also a Francophone Jurisprudence variant of this pattern, where you set `process.exitCode = 1` early and then set it to `0` only if main exits cleanly, such that a bare call to `process.exit()` without a specified code will default to failure, among other unusual exit conditions that should not be treated as success. There are disadvantages to both approaches, but I’m not recalling all them.

gibson042: This is more common in our codebase, and I'm taking the position that invocation of `process.exit` is a code smell. We can revisit en masse in the future.

PR #3005, opened 2025-11-04
fix(ses): (assert-shim): No error redaction without lockdown

kriskowal:
## Description

In preparation to introduce a new `@endo/harden` package, toward #2983 decoupling harden from lockdown, it became evident that redacting errors without lockdown left a system in an undebuggable state, since lockdown is necessary to install an unredacter. This was an oversight easily fixed. The default `assert` introduced by initialization of `ses` should not redact, and `lockdown` should replace this `assert` with one that does redact along with its paired unredacting `console` and platform hooks for unredacting uncaught errors and unhandled rejections.

This is important because it will allow multiplexed tests to see the same exception messages before and after lockdown when using `ses-ava`.

### Security Considerations

This relaxation pertains only to systems that run without lockdown and for which SES provides no security claims.

### Scaling Considerations

None.

None.

### Testing Considerations

None.

None.

### Upgrade Considerations

None.

### DIFF
--- a/packages/ses/NEWS.md
+++ b/packages/ses/NEWS.md
@@ -4,6 +4,14 @@ User-visible changes in `ses`
 
 - Adds `assert.makeError` and deprecates `assert.error` as an alias, matching
   the API already exported from `@endo/errors`.
+- Before this version, the `assert` left in global scope before `lockdown`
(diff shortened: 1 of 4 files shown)


erights (approved):
LGTM.

Can we test this? If so, please add a test. Thanks!

PR #3008, opened 2025-11-07
feat(harden): Introduce @endo/harden

kriskowal:
Closes: #2978 

## Description

This change introduces an `@endo/harden` package that allows packages to be written for use in a JS or a HardenedJS environment without modification. The `@endo/harden` module provides a behavior that depends on the environment and packaging conditions.

Without any packaging conditions, in a HardenedJS environment, `@endo/harden` provides the environment’s “volume freezing” `harden`, meaning that it freezes the closure over both dimensions: transitive properties and prototypes.

Also without any packaging conditions, if the environment does not provide a `harden`, `@endo/harden` provides a “surface freezing” `harden`, meaning that it freezes the closure over only the one dimension: properties. This provides a modicum of immutability without interfering with shims or other mutations to shared, intrinsic prototypes.

With the `hardened` condition (`node -C hardened`, `bundle-source -C hardened`), `@endo/harden` will not retain an implementation of `harden` and will assert that `harden` existed as `Object[Symbol.for('harden')]` or `globalThis.harden` in the environment and vend out that `harden`. This is useful to minimize the size of bundles that can safely presume that they will run in a HardenedJS environment.

With the `noop-harden` condition (`node -C noop-harden`), `@endo/harden` will provide a version of `harden` that returns its argument unaltered.

With these new modes, we expect to deprecate the `lockdown` option for `"unsafe"` `hardenTaming` which goes further and replaces `isExtensible`, `isFrozen`, and `isSealed` with versions that misreport `true` for extensible, unfrozen, or unsealed objects respectively. We hope that the new default behavior of surface hardening will suffice, but we leave the `noop-harden` condition as an option since that should have performance parity with unsafe harden taming for environments that need it.

As a side-effect, every kind of `harden` will install itself on first use at `Object[Symbol.for('harden')]` as a non-configurable property such that the first `@endo/harden` implementation used wins the race to define the hardening behavior of the realm. SES will install the same property at time of `lockdown`, but if it loses the race, will throw an exception indicating that the realm cannot be locked down because of unsafe usage of `harden` before `lockdown`, and render up the stack of the first use for diagnostic purposes.

### Security Considerations

The `@endo/harden` provides a new mode of usage that is less safe than `lockdown` for environments in which `lockdown` is not practical. We do not expect safety to regress in lockdown environments as a consequence.

This change strengthens one safety guarantee: going forward, hardened modules using `@endo/harden` will not be vulnerable to hosts that endow a compartment with a weakened version of `harden`, because `@endo/harden` always favors the `Object[Symbol.for('harden')]` enshrined on a shared intrinsic hardened by `lockdown`.

### Scaling Considerations

Adopting `@endo/harden` will increase the size of bundles, and since this change adopts `@endo/harden` throughout the Endo stack, this bundle size increase may become problematic for systems close to their bundle size limits. We provide the bundler condition `hardened` to mitigate this problem.

- [ ] This change comes with documentation in README and NEWS for all impacted packages, including advice to adopt the `hardened` bundle condition to mitigate the bundle size increase.

### Testing Considerations

This change adds configurations to `sesAvaConfigs` where adopting `@endo/harden` allows those packages to be used in more configurations. The salient configuration Endo with shims installed only, without calling Lockdown, in the cases where packages continue to rely on Assert or Eventual Send. We hope in time to test in the Base configuration, without any shims. Some packages are able to adopt the No-op mode of harden and are accordingly tested in that mode.

This change is additive apart from the expected increase in bundle size, for which we provide a mitigation.

### Upgrade Considerations

None.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -441,6 +441,7 @@ __metadata:
     "@endo/base64": "workspace:^"
     "@endo/compartment-mapper": "workspace:^"
     "@endo/evasive-transform": "workspace:^"
+    "@endo/harden": "workspace:^"
     "@endo/init": "workspace:^"
     "@endo/lockdown": "workspace:^"
     "@endo/promise-kit": "workspace:^"
@@ -476,6 +477,7 @@ __metadata:
   dependencies:
     "@endo/errors": "workspace:^"
     "@endo/eventual-send": "workspace:^"
+    "@endo/harden": "workspace:^"
     "@endo/init": "workspace:^"
     "@endo/marshal": "workspace:^"
     "@endo/nat": "workspace:^"
@@ -496,6 +498,7 @@ __metadata:
     "@endo/bundle-source": "workspace:^"
     "@endo/compartment-mapper": "workspace:^"
     "@endo/errors": "workspace:^"
+    "@endo/harden": "workspace:^"
     "@endo/init": "workspace:^"
     "@endo/ses-ava": "workspace:^"
     "@endo/zip": "workspace:^"
@@ -529,6 +532,7 @@ __metadata:
     "@endo/eventual-send": "workspace:^"
     "@endo/exo": "workspace:^"
     "@endo/far": "workspace:^"
+    "@endo/harden": "workspace:^"
     "@endo/import-bundle": "workspace:^"
     "@endo/init": "workspace:^"
     "@endo/lockdown": "workspace:^"
@@ -555,6 +559,7 @@ __metadata:
   dependencies:
     "@endo/errors": "workspace:^"
     "@endo/eventual-send": "workspace:^"
+    "@endo/harden": "workspace:^"
     "@endo/lockdown": "workspace:^"
     "@endo/promise-kit": "workspace:^"
     "@endo/ses-ava": "workspace:^"
@@ -572,6 +577,7 @@ __metadata:
   dependencies:
     "@endo/cjs-module-analyzer": "workspace:^"
     "@endo/env-options": "workspace:^"
+    "@endo/init": "workspace:^"
     "@endo/module-source": "workspace:^"
     "@endo/path-compare": "workspace:^"
     "@endo/trampoline": "workspace:^"
@@ -596,6 +602,7 @@ __metadata:
     "@endo/eventual-send": "workspace:^"
     "@endo/exo": "workspace:^"
     "@endo/far": "workspace:^"
+    "@endo/harden": "workspace:^"
     "@endo/import-bundle": "workspace:^"
     "@endo/init": "workspace:^"
     "@endo/lockdown": "workspace:^"
@@ -631,6 +638,7 @@ __metadata:
   resolution: "@endo/errors@workspace:packages/errors"
   dependencies:
     "@endo/eventual-send": "workspace:^"
+    "@endo/harden": "workspace:^"
     "@endo/lockdown": "workspace:^"
     "@endo/ses-ava": "workspace:^"
     ava: "catalog:dev"
@@ -680,6 +688,7 @@ __metadata:
   resolution: "@endo/eventual-send@workspace:packages/eventual-send"
   dependencies:
     "@endo/env-options": "workspace:^"
+    "@endo/harden": "workspace:^"
     "@endo/lockdown": "workspace:^"
     ava: "catalog:dev"
     c8: "catalog:dev"
@@ -698,6 +707,7 @@ __metadata:
     "@endo/errors": "workspace:^"
     "@endo/eventual-send": "workspace:^"
     "@endo/far": "workspace:^"
+    "@endo/harden": "workspace:^"
     "@endo/init": "workspace:^"
     "@endo/marshal": "workspace:^"
     "@endo/pass-style": "workspace:^"
@@ -716,12 +726,26 @@ __metadata:
   dependencies:
     "@endo/errors": "workspace:^"
     "@endo/eventual-send": "workspace:^"
+    "@endo/harden": "workspace:^"
     "@endo/init": "workspace:^"
     "@endo/pass-style": "workspace:^"
     "@endo/ses-ava": "workspace:^"
     ava: "catalog:dev"
     c8: "catalog:dev"
     eslint: "catalog:dev"
+    typescript: "npm:~5.9.2"
+  languageName: unknown
+  linkType: soft
+
+"@endo/harden@workspace:^, @endo/harden@workspace:packages/harden":
+  version: 0.0.0-use.local
+  resolution: "@endo/harden@workspace:packages/harden"
+  dependencies:
+    ava: "catalog:dev"
+    c8: "catalog:dev"
+    eslint: "catalog:dev"
+    ses: "workspace:^"
+    tsd: "catalog:dev"
     typescript: "npm:~5.9.2"
   languageName: unknown
   linkType: soft
@@ -746,6 +770,7 @@ __metadata:
     "@endo/bundle-source": "workspace:^"
     "@endo/compartment-mapper": "workspace:^"
     "@endo/errors": "workspace:^"
+    "@endo/harden": "workspace:^"
     "@endo/init": "workspace:^"
     "@endo/ses-ava": "workspace:^"
     "@endo/where": "workspace:^"
@@ -762,8 +787,8 @@ __metadata:
   resolution: "@endo/init@workspace:packages/init"
   dependencies:
     "@endo/base64": "workspace:^"
-    "@endo/compartment-mapper": "workspace:^"
     "@endo/eventual-send": "workspace:^"
+    "@endo/harden": "workspace:^"
     "@endo/lockdown": "workspace:^"
     "@endo/promise-kit": "workspace:^"
     ava: "catalog:dev"
@@ -786,6 +811,7 @@ __metadata:
   resolution: "@endo/lp32@workspace:packages/lp32"
   dependencies:
     "@endo/errors": "workspace:^"
+    "@endo/harden": "workspace:^"
     "@endo/init": "workspace:^"
     "@endo/ses-ava": "workspace:^"
     "@endo/stream": "workspace:^"
@@ -805,6 +831,7 @@ __metadata:
     "@endo/env-options": "workspace:^"
     "@endo/errors": "workspace:^"
     "@endo/eventual-send": "workspace:^"
+    "@endo/harden": "workspace:^"
     "@endo/init": "workspace:^"
     "@endo/lockdown": "workspace:^"
     "@endo/nat": "workspace:^"
@@ -823,6 +850,7 @@ __metadata:
   version: 0.0.0-use.local
   resolution: "@endo/memoize@workspace:packages/memoize"
   dependencies:
+    "@endo/harden": "workspace:^"
     "@endo/init": "workspace:^"
     "@endo/ses-ava": "workspace:^"
     ava: "catalog:dev"
@@ -841,6 +869,7 @@ __metadata:
     "@babel/parser": "npm:~7.28.3"
     "@babel/traverse": "npm:~7.28.3"
     "@babel/types": "npm:~7.28.2"
+    "@endo/harden": "workspace:^"
     ava: "catalog:dev"
     babel-eslint: "npm:^10.1.0"
     benchmark: "npm:^2.1.4"
@@ -869,6 +898,7 @@ __metadata:
   version: 0.0.0-use.local
   resolution: "@endo/netstring@workspace:packages/netstring"
   dependencies:
+    "@endo/harden": "workspace:^"
     "@endo/init": "workspace:^"
     "@endo/promise-kit": "workspace:^"
     "@endo/ses-ava": "workspace:^"
@@ -901,6 +931,7 @@ __metadata:
   dependencies:
     "@endo/errors": "workspace:^"
     "@endo/eventual-send": "workspace:^"
+    "@endo/harden": "workspace:^"
     "@endo/immutable-arraybuffer": "workspace:^"
     "@endo/init": "workspace:^"
     "@endo/lockdown": "workspace:^"
@@ -939,6 +970,7 @@ __metadata:
     "@endo/env-options": "workspace:^"
     "@endo/errors": "workspace:^"
     "@endo/eventual-send": "workspace:^"
+    "@endo/harden": "workspace:^"
     "@endo/init": "workspace:^"
     "@endo/promise-kit": "workspace:^"
     "@endo/ses-ava": "workspace:^"
@@ -969,6 +1001,7 @@ __metadata:
     "@endo/common": "workspace:^"
     "@endo/errors": "workspace:^"
     "@endo/eventual-send": "workspace:^"
+    "@endo/harden": "workspace:^"
     "@endo/init": "workspace:^"
     "@endo/marshal": "workspace:^"
     "@endo/pass-style": "workspace:^"
@@ -986,6 +1019,7 @@ __metadata:
   version: 0.0.0-use.local
   resolution: "@endo/promise-kit@workspace:packages/promise-kit"
   dependencies:
+    "@endo/harden": "workspace:^"
     ava: "catalog:dev"
     c8: "catalog:dev"
     eslint: "catalog:dev"
@@ -999,6 +1033,7 @@ __metadata:
   resolution: "@endo/ses-ava@workspace:packages/ses-ava"
   dependencies:
     "@endo/env-options": "workspace:^"
+    "@endo/harden": "workspace:^"
     "@endo/init": "workspace:^"
     "@endo/panic": "workspace:^"
     ava: "catalog:dev"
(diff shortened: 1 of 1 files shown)


packages/harden/make-hardener.js
```diff
@@ -0,0 +1,471 @@
+// Adapted from SES/Caja - Copyright (C) 2011 Google Inc.
```
erights, line 1: Why does this PR only add this file here, rather than (re)moving it from ses/src? What are the salient differences from the one in ses/src?

packages/ocapn/src/captp/captp-engine.js
```diff
@@ -0,0 +1,539 @@
+/** @import {RemoteKit, Settler} from '@endo/eventual-send' */
```
erights: Why does this PR contain this file as a new file?

packages/module-source/src/transform-analyze.js
```diff
@@ -183,7 +183,7 @@ export const makeModuleTransformer = (babel, importer) => {
           ? maybeSource.slice(0, -1)
           : maybeSource;
 
-      return harden({ ...ss, endowments, src: actualSource });
+      return { ...ss, endowments, src: actualSource };
```
erights, line 186: Why remove this `harden` call, or any similar ones?

packages/module-source/NEWS.md
```diff
@@ -1,5 +1,9 @@
 User-visible changes in `@endo/module-source`:
 
+# Next release
+
+- Removes dependence on global `harden`.
```
erights: But why? This PR makes it possible to use `harden` with almost zero cost.

packages/harden/make-hardener.js
kriskowal: I’ve set up the commits for this PR to make these changes evident.

- Straight copy from SES: https://github.com/endojs/endo/pull/3008/changes/8ef4cb8c23fee4e5b525047ecd4279ed6c8b8a17
- Relieve dependence on SES internal assert https://github.com/endojs/endo/pull/3008/changes/8ef4cb8c23fee4e5b525047ecd4279ed6c8b8a17
- Add traversePrototypes option https://github.com/endojs/endo/pull/3008/changes/b17cb9f4645256331ee544c976e7aa75175ce378
- Copy entirety of commons inline into this module without modification https://github.com/endojs/endo/pull/3008/changes/00e76cc4692bbb2bf85a146f22c00311dfba9a13
- Mechanically remove `exports` for commons https://github.com/endojs/endo/pull/3008/changes/0be9b4195768c068b059a4e01c394405c5b4a330
- Inline the Harden type to relive a dependence on SES types https://github.com/endojs/endo/pull/3008/changes/3ab5053fa8a1ee23f9046d3bc46013f0910bfb7a
- Collect unused Commons https://github.com/endojs/endo/pull/3008/changes/d1fea29d52ee984352cb01ed31f67e09cdec7445

packages/module-source/NEWS.md
kriskowal: This was not previously a rigorously hardened module. The constructor was not hardened, so this `harden` was a paltry nod. But, you are correct that we could make it a hardened module, now.

packages/module-source/src/transform-analyze.js
kriskowal: I have appended commits that make ModuleSource a hardened module with `@endo/harden`. There is a tiny caveat: some existing code imported ModuleSource before calling lockdown, which worked because it wasn’t attempting to harden itself. I’ve fixed the relevant code inside but there’s a risk that this will require accommodation downstream. I think the risk is acceptable.

kriskowal:
I ran into trouble using `@endo/harden` in `ModuleSource` because we require `@endo/module-source/shim.js` to be imported before `lockdown`, and using `harden` before `lockdown` breaks `lockdown`.

The solution was to perform a transitive surface freeze of ModuleSource constructors, prototypes, and instances manually.

.changeset/hardened-lockdown-sense.md
```diff
@@ -0,0 +1,9 @@
+---
+'ses': minor
+---
+
+- `lockdown` and `repairIntrinsics` now detect if a hardened module (using
```
erights: It is possible to misunderstand this as detecting whether a module imported `@endo/harden` as opposed to detecting whether it called the imported `harden`.

packages/eslint-plugin/lib/configs/internal.js
```diff
@@ -78,6 +78,9 @@ module.exports = {
     'bundles/**',
     'coverage/**',
     'dist/**',
+    '**/*.d.ts*',
+    '**/*.d.cts*',
+    '**/*.d.mts*',
```
erights, line 83: The `*` at the end seems overly broad. Would

```suggestion
    '**/*.d.ts',
    '**/*.d.cts',
    '**/*.d.mts',
```

work?

packages/module-source/src/transform-analyze.js
erights: Good. Resolving.

packages/ocapn/src/client/index.js
```diff
@@ -351,6 +352,9 @@ export const makeClient = ({
    */
   const handleConnectionClose = (connection, reason) => {
     logger.info(`handleConnectionClose called`, { reason });
+    if (!connection.isDestroyed) {
+      connection.end();
+    }
```
erights: Is this a drive-by, or is it somehow related to `harden`?

packages/harden/README.md
```diff
@@ -0,0 +1,156 @@
+# harden
+
+Hardened modules are modules that make their exports resist tampering by other
```
erights: This is an overview, so if we do not find a compact and correct way to fix this, no change requested. But it is a lot more than just the exports. I do not have a compact correct suggestion.

packages/harden/README.md
```diff
@@ -0,0 +1,156 @@
...
+modules that import them, making them less suceptible to supply chain attack.
+In [HardenedJS](https://hardenedjs.org), the global `harden` function
+transitively freezes an object and all of the objects that are reachable by
+walking up chains of properties and prototypes.
```
erights:
```suggestion
walking chains of properties and prototypes.
```
"up" only applies to prototypes.

packages/harden/README.md
```diff
@@ -0,0 +1,156 @@
...
+All the primordials like `Array.prototype` and `Object` are frozen in
+this environment, which gives your module a place to stand toward its own
+defense.
+Then, with [LavaMoat](https://github.com/lavamoat/lavamoat), each package
+is credibly isolated and only receives the subset of globals and host modules
+it needs to function.
+That is, we can enforce [Principle of Least
+Authority](https://en.wikipedia.org/wiki/Principle_of_least_privilege).
```
erights, line 17: Not now, and probably not in this PR, but I hope we find a better link to something that does not confuse POLA with "Principle of least privilege". Or maybe eventually revise the wikipedia article to actually explain the difference. Sigh.

packages/harden/README.md
```diff
@@ -0,0 +1,156 @@
...
+But, that leaves the module to use `harden` to freeze all its exports and
+anything it returns that might be shared by other packages that use it.
```
erights, line 19: Aha! This fixes my earlier concern.

packages/harden/README.md
```diff
@@ -0,0 +1,156 @@
...
+an exception if `harden` is not present.
+
+```
+bundle-source -C hardened entry.js > entry.json
+```
+
+# Without HardenedJS
+
+Libraries that use `@endo/harden` can be used without HardenedJS and the
+exported `harden` only freezes the transitive owned properties of the object
```
erights:
```suggestion
exported `harden` only freezes the transitive own properties of the object
```

erights: This phrasing does not clearly imply that the object itself is frozen, i.e., that it is non-extensible.

packages/harden/README.md
```diff
@@ -0,0 +1,156 @@
...
+does not interfere with uncoordinated alteration of the realm intrinsics, on
+which some testing and frontend user interface frameworks rely.
+
+To opt out of any safety guarantees and to avoid the computation cost of
+transitively hardening own properties, use the `-C harden:unsafe` build
+condition with tools like `node` and Endo's `bundle-source`.
+
+# Multiple instances
+
+The first instance of `@endo/harden` will determine the behavior of any
```
erights: Not the first call?

packages/harden/README.md
```diff
@@ -0,0 +1,156 @@
...
+subsequent instance of `@endo/harden` that initializes later, regardless of
+differences in behavior.
+In a mutable, pre-lockdown JavaScript environment, it does this by behaving
+somewhat like a shim.
+A side-effect of the _first use_ of `harden` is that it installs its flavor of
```
erights: "_first use_" sounds like first call, not first import or instantiation.

packages/harden/README.md
```diff
@@ -0,0 +1,156 @@
...
+`harden` at `Object[Symbol.for('harden')]` and all subsequent initializations
+just adopt that behavior.
+This property is how `lockdown` senses that it should fail.
+
+# With _or_ Without _not_ Both
+
+Hardened modules calling `harden` should be fine at any time in an application
+that never uses HardenedJS, calling `lockdown`.
```
erights: The comma confuses me. What if an app imports, for example `'ses'` and does not call `lockdown`? What if an app imports `'ses'` and calls `repairIntrinsics` but not `hardenIntrinsics`? I am not saying this needs to be explained here. But if not, how about linking to where it is explained?

packages/harden/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+However, initializing a hardened module before setting up a HardenedJS
+environment (before calling `lockdown`) and then proceeding on the assumption
+that it's hardened after `lockdown` would leave the apparently-hardened module
+vulnerable.
+
+So, `@endo/harden` arranges for `lockdown()` to throw an exception with
+a _helpful_ stack if `harden` gets called before `lockdown`.
```
erights: I love this point!

packages/harden/README.md
```diff
@@ -0,0 +1,156 @@
...
+`@endo/lockdown`, `@endo/init`, or by convention, modules with names like
+`prepare-*`.
+
+# Configurability of Compartment harden
+
+The `harden` exported by `@endo/harden` prefers `Object[Symbol.for('harden')]`
+over `globalThis.harden` since the former is an intrinsic that cannot be
+overridden by an endowment.
+Any code that relies on `globalThis.harden` being endowed with a different
+behavior than `Object[Symbol.for('harden')]` cannot substitute `@endo/harden`.
```
erights: What does "cannot substitute `@endo/harden`" mean?

packages/harden/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+# isFake
```
erights:
```suggestion
# isFake (deprecated)
```

packages/harden/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+Using `lockdown` with the `"unsafe"` `hardenTaming` option creates an environment
```
erights:
```suggestion
Using `lockdown` with the deprecated `"unsafe"` `hardenTaming` option creates an environment
```

packages/harden/README.md
```diff
@@ -0,0 +1,156 @@
...
+and its compatriots, which may not be indicated by empirical behavior of `harden`.
+
+For example, `Object.isFrozen({})` when `harden.isFake` and more clearly
+conveys the reason a test might be invalidated by `unsafe` `hardenTaming`.
+Testing that the outcome of `Object.isFrozen({})` is the same as the outcome of
+`Object.isFrozen(object)` for an object that should not be frozen makes a test
+work just as well between `safe` and `unsafe` `hardenTaming`.
+
+The module `@endo/harden/is-noop.js` provides `hardenIsNoop(harden)` to
+indicate that `harden` is a no-op, regardless of `hardenTaming`.
```
erights:
```suggestion
detect whether `harden` is a no-op, regardless of `hardenTaming`.
```

packages/pass-style/src/make-far.js
```diff
@@ -97,14 +96,10 @@ export const Remotable = (
...
-  // to succeed anyway. Faking `harden` is only correctness preserving
-  // if the code in question contains no bugs that the real `harden` would
-  // have caught.
-  // @ts-ignore `isFake` purposely not in the type
-  harden.isFake ||
-    // Ensure that the remotable isn't already frozen.
-    !isFrozen(remotable) ||
+  // Ensure that the remotable isn't already frozen.
+  // Recall that isFrozen always returns true when using lockdown with unsafe
+  // hardenTaming.
```
erights:
```suggestion
  // Recall that isFrozen always returns true when using lockdown with
  // hardenTaming set to the deprecated `'unsafe'` option.
```

packages/harden/noop.js
```diff
@@ -0,0 +1,20 @@
+/* This implementation of harden first senses and provides the implementation
+ * of harden present in the global object tor shared intrinsics.
+ * Failing to find an existing implementation, this provides and installs a
+ * fake version that does nothing.
+ * This version can be selected with the package.json condition
+ * "harden:unsafe", as with node -C harden:unsafe or Endo's bundle-source -C
+ * harden:unsafe.
```
erights:
```suggestion
 * "harden:unsafe", as with "node -C harden:unsafe or Endo's bundle-source -C
 * harden:unsafe".
```

or backquotes

packages/harden/surface.js
```diff
@@ -0,0 +1,17 @@
+/* This implementation of harden first senses and provides the implementation
```
erights: Why repeat `index.js`? Can we DRY?

packages/harden/make-hardener.js
```diff
@@ -51,7 +51,13 @@ import {
...
   isError,
 } from './commons.js';
-import { assert } from './error/assert.js';
+
+/** @type {(condition: any) => asserts condition} */
+const assert = condition => {
+  if (!condition) {
+    throw new TypeError('assertion failed');
+  }
+};
```
erights, line 249: I'm surprised that you don't need diagnostic info for the `message` of the thrown error. But if you don't, good to omit. No change suggested.

packages/harden/make-hardener.js
```diff
@@ -0,0 +1,471 @@
...
+ * for such a trivial case outside commons.js, we provide a utility function.
+ *
+ * @param {any} value
+ */
+const isError = value => value instanceof FERAL_ERROR;
+
+// ////////////////// FERAL_STACK_GETTER FERAL_STACK_SETTER ////////////////////
+
+// The error repair mechanism is very similar to code in ses/src/commons.js
+// and these implementations should be kept in sync.
```
erights, line 174: Please add a similar comment there.

packages/harden/make-hardener.js
```diff
@@ -0,0 +1,471 @@
...
+                  stackDesc.get === FERAL_STACK_GETTER &&
+                  stackDesc.configurable
+                ) {
+                  // Can only repair if it is configurable. Otherwise, leave
+                  // unrepaired, in which case it will not be judged passable,
+                  // avoiding a safety problem.
+                  defineProperty(obj, 'stack', {
+                    // NOTE: Calls getter during harden, which seems dangerous.
+                    // But we're only calling the problematic getter whose
+                    // hazards we think we understand.
```
erights, line 436: Oh this brings back so many unpleasant memories. Sigh.

No change suggested!

packages/harden/make-hardener.js
erights: THANKS! This was amazingly helpful.

packages/eslint-plugin/lib/configs/internal.js
kriskowal: TypeScript generates some dross with extensions that match the broader pattern.

packages/ocapn/src/client/index.js
kriskowal: This is churn that got inadvertently captured in a rebase. I’ve deleted it.

packages/harden/README.md
kriskowal: Change `exports` to `interfaces` and added a sentence that clarifies the greater extent.

packages/harden/README.md
kriskowal: I’ve changed this to:

```
Hardened modules calling `harden` should be fine at any time in an application
that never calls `lockdown` or `repairIntrinsics`.
```

Which, I think is much clearer.

packages/harden/README.md
kriskowal: What I mean is more clear if I remove that fragment and do not elaborate.

The problem is that code can either rely on `harden` to exist in scope or they can use `@endo/harden`. If they rely on `harden` in scope, their behavior can be a function of what version of `harden` was endowed to the compartment. If they use `@endo/harden`, they will use the realm intrinsic regardless.

packages/harden/surface.js
kriskowal: Can DRY. An artifact of refactoring.

packages/harden/make-hardener.js
kriskowal: So it is written.

PR #3009, opened 2025-11-20
test(ses): cover the allowNextEvalToBeUnsafe failure after stack overflow

naugtur:
## Description

Just a test covering an edge-case that's rare but known but often hard to recall.

### DIFF
--- a/packages/ses/src/eval-scope.js
+++ b/packages/ses/src/eval-scope.js
@@ -79,6 +79,7 @@ export const makeEvalScopeKit = () => {
     eval: {
       value: props => {
         const { eval: evalProp } = props;
+        delete evalScope.eval;
         defineProperties(evalScope, { eval: evalProp });
       },
       enumerable: false,
@@ -88,7 +89,7 @@ export const makeEvalScopeKit = () => {
 
   const evalScopeKit = {
     evalScope,
-    allowNextEvalToBeUnsafe(evaluate) {
+    allowNextEvalToBeUnsafe(evaluate, globalObject) {
       const { revoked } = evalScopeKit;
       if (revoked !== null) {
         Fail`a handler did not reset allowNextEvalToBeUnsafe ${revoked.err}`;
@@ -98,7 +99,7 @@ export const makeEvalScopeKit = () => {
       // its return value.
       defineProperties(evalScope, allowEvalProperties);
       try {
-        apply(evaluate, null, [oneTimeEvalProperties]);
+        apply(evaluate, globalObject, [oneTimeEvalProperties]);
       } catch (err) {
         // We didn't manage to reveal eval, so clean up the powerless revealer
         delete evalScope.eval;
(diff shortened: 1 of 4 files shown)


packages/ses/test/make-safe-evaluator.test.js
```diff
@@ -115,3 +115,31 @@ test('safeEvaluate - transforms - rewrite source', t => {
...
+    globalObject,
+    moduleLexicals,
+  });
+
+  let depth = 0;
+  function stackBuildUp() {
+    depth += 1;
+    evaluate(`1+1`);
+    // didn't fail yet? keep digging.
+    return stackBuildUp();
```
mhofman, line 134: I would recommend avoiding tail recursion here as that is an explicitly allowed optional optimization.

packages/ses/test/make-safe-evaluator.test.js
```diff
@@ -115,3 +115,31 @@ test('safeEvaluate - transforms - rewrite source', t => {
...
+  }
+
+  t.throws(stackBuildUp, {
+    instanceOf: RangeError,
+    message: /Maximum call stack size exceeded/,
+  });
+  t.is(depth > 1, true, 'stack overflow occurred after many recursions');
+  t.throws(() => evaluate('1+1'), {
+    message: /a handler did not reset allowNextEvalToBeUnsafe \(a RangeError\)/,
+  });
```
mhofman: Interesting, I thought we couldn't actually trigger this case. We may have broken the stack overflow protection at some point.

mhofman (commented):
I'm surprised you managed to build a test reproducing this. I'll try to look how we actually end up in this case.

packages/ses/src/eval-scope.js
```diff
@@ -69,17 +75,35 @@ export const makeEvalScopeKit = () => {
     },
   });
 
+  const allowEvalProperties = freeze({
+    eval: {
+      value: props => {
+        const { eval: evalProp } = props;
+        delete evalScope.eval;
+        defineProperties(evalScope, { eval: evalProp });
+      },
```
naugtur, line 84: Took me a good 10 minutes of staring at this to figure out why it works.
I suggest we add a lengthy comment to explain it 😅 
Something like this maybe:

```suggestion
  // This conforms to the shape of evalScope properties, but instead of evaluating what arrives as first 
  // argument, it accepts a new shape to replace eval with. 
  // As a result, when `apply(evaluate, globalObject, [oneTimeEvalProperties]);` is called below,  
  // the same exact stack depth is reached and the same optimizations are applied to it before the  
  // FERAL_EVAL is put in place, making it impossible for a stack overflow error to occur in the process 
  // of getting eval for the main call to the evaluator if it didn't occur in revealing eval.
  const allowEvalProperties = freeze({
    eval: {
      value: props => {
        const { eval: evalProp } = props;
        delete evalScope.eval;
        defineProperties(evalScope, { eval: evalProp });
      },
```

PR #3013, opened 2025-12-01
OCapN: improve spec conformance

kumavis:
added CI test against python server (temporarily points to [this commit](https://github.com/ocapn/ocapn-test-suite/pull/24))

also see https://github.com/ocapn/ocapn-test-suite/issues/21

Outstanding:
- needs means of sending `op:untag`
- rename of Selector to Symbol
- flakey test observed in python test! https://github.com/endojs/endo/actions/runs/20027716477/job/57428931009

### DIFF
--- a/packages/ocapn/src/client/index.js
+++ b/packages/ocapn/src/client/index.js
@@ -156,6 +156,7 @@ const compareSessionKeysForCrossedHellos = (
  * @param {any} message
  * @param {string} captpVersion
  * @param {boolean} enableImportCollection
+ * @param {boolean} debugMode
  */
 const handleSessionHandshakeMessage = (
   debugLabel,
@@ -169,6 +170,7 @@ const handleSessionHandshakeMessage = (
   message,
   captpVersion,
   enableImportCollection,
+  debugMode,
 ) => {
   logger.info(`handling handshake message of type ${message.type}`);
   switch (message.type) {
@@ -261,6 +263,7 @@ const handleSessionHandshakeMessage = (
         sturdyRefTracker,
         debugLabel,
         enableImportCollection,
+        debugMode,
       );
       const session = makeSession({
         id: sessionId,
@@ -302,6 +305,7 @@ const handleSessionHandshakeMessage = (
  * @param {Uint8Array} data
  * @param {string} captpVersion
  * @param {boolean} enableImportCollection
+ * @param {boolean} debugMode
  */
 const handleHandshakeMessageData = (
   debugLabel,
@@ -315,6 +319,7 @@ const handleHandshakeMessageData = (
   data,
   captpVersion,
   enableImportCollection,
+  debugMode,
 ) => {
   try {
     const syrupReader = makeSyrupReader(data);
@@ -347,6 +352,7 @@ const handleHandshakeMessageData = (
           message,
           captpVersion,
           enableImportCollection,
+          debugMode,
         );
       } else {
         logger.info(
@@ -496,6 +502,7 @@ const makeSessionManager = () => {
  * @param {Map<string, any>} [options.giftTable]
  * @param {string} [options.captpVersion] - For testing: override the CapTP version sent in handshakes
  * @param {boolean} [options.enableImportCollection] - If true, imports are tracked with WeakRefs and GC'd when unreachable. Default: true.
+ * @param {boolean} [options.debugMode] - If true, exposes `debug` object on Ocapn instances with internal APIs for testing. Default: false.
  * @returns {Client}
  */
 export const makeClient = ({
@@ -505,6 +512,7 @@ export const makeClient = ({
   giftTable = new Map(),
   captpVersion = '1.0',
   enableImportCollection = true,
+  debugMode = false,
 } = {}) => {
   /** @type {Map<string, NetLayer>} */
   const netlayers = new Map();
@@ -605,6 +613,7 @@ export const makeClient = ({
           data,
           captpVersion,
           enableImportCollection,
+          debugMode,
         );
       }
     },
--- a/packages/ocapn/src/client/ocapn.js
+++ b/packages/ocapn/src/client/ocapn.js
@@ -593,16 +593,20 @@ const makeBootstrapObject = (
 };
 
 /**
+ * @typedef {object} OcapnDebug
+ * @property {OcapnTable} ocapnTable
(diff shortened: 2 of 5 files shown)


packages/ocapn/src/client/ocapn.js
```diff
@@ -1171,6 +1171,13 @@ export const makeOcapn = (
         const syrupMessage = decodeSyrup(problematicBytes);
         logger.error(`Message decode error:`);
         console.dir(syrupMessage, { depth: null });
+        console.log(
+          JSON.stringify(
+            syrupMessage,
+            (key, value) => (typeof value === 'bigint' ? `${value}n` : value),
+            2,
+          ),
+        );
```
kriskowal: Consider commenting out the diagnostics, apart from the logger, since these should be silent by default. If you keep the `console`, consider using `console.error` instead of `console.log` so it doesn’t get inadvertently interleaved with `stdout`. I don’t know whether `console.dir` can interfere with `stdout`. If you keep these, consider using `@endo/marshal` for rendering passable data as “Justin”, which is an unambiguous representation.

Consider:
```js
import { passableAsJustin } from @endo/marshal';

console.error(passableAsJustin(syrupMessage));
```

packages/ocapn/test/client.test.js
```diff
@@ -637,3 +637,82 @@ testWithErrorUnwrapping(
     shutdownBoth();
   },
 );
+
+testWithErrorUnwrapping(
```
kriskowal, line 641: Apropos of this, tho not directly related, OCapN errors are not yet fully specified. They’re likely to be a `message` and a Struct, but with the limitation that the Struct must not transitively contain any references, just data.

packages/ocapn/test/client.test.js
```diff
@@ -637,3 +637,82 @@ testWithErrorUnwrapping(
...
+          instanceOf: Error,
+          message: /Cannot apply functions to values with pass-style sturdyref/,
+        },
+        'Delivering to a promise that resolves to a sturdyref should throw',
+      );
+
+      t.truthy(error, 'Error should be thrown');
+
+      // Wait a bit to ensure no disconnection happens
+      await new Promise(resolve => setTimeout(resolve, 100));
```
kriskowal, line 701: I noticed this. I don’t like it but don’t see a way around it. It doesn’t seem to be racy, just that a slow test runner might be less likely to catch a disconnect.

packages/ocapn/test/client.test.js
kriskowal: https://github.com/ocapn/ocapn/pull/248

packages/ocapn/src/client/index.js
```diff
@@ -653,6 +649,19 @@ export const makeClient = ({
...
+     * @returns {Promise<any>}
+     */
+    enlivenSturdyRef(sturdyRef) {
+      return enlivenSturdyRef(
+        sturdyRef,
+        location => client.provideSession(location),
+        isSelfLocation,
+        swissnumTable,
+      );
+    },
```
kriskowal, line 684: Adequate, for now. For the record, we want the client to have methods for “locator from sturdyref” (inspect coupon), “sturdyref from locator” (issue coupon), and the “redeem” capability of a coupon should be inherent to the coupon.

Where coupon a the name for sturdyref I’m entertaining.

packages/ocapn/src/client/sturdyrefs.js
```diff
@@ -6,9 +6,16 @@
  */
 
 import { E } from '@endo/eventual-send';
+import { makeTagged } from '@endo/pass-style';
 import { decodeSwissnum } from './util.js';
 
 /**
+ * @import { CopyTagged } from '@endo/pass-style'
+ * @typedef {CopyTagged<'ocapn-sturdyref', undefined>} SturdyRef
+ * A SturdyRef is reified as a tagged value with tag 'ocapn-sturdyref' and undefined payload.
```
kriskowal: suggest “tentatively reified”

packages/ocapn/src/client/sturdyrefs.js
```diff
@@ -98,6 +43,41 @@ export const getSturdyRefDetails = sturdyRef => {
...
 
+/**
+ * Enliven a SturdyRef by fetching the actual object
+ * @param {SturdyRef} sturdyRef
+ * @param {(location: OcapnLocation) => Promise<Session>} provideSession
+ * @param {(location: OcapnLocation) => boolean} isSelfLocation
+ * @param {Map<string, any>} swissnumTable
+ * @returns {Promise<any>}
+ */
+export const enlivenSturdyRef = async (
```
kriskowal, line 54: Great.

packages/ocapn/test/codecs/passable.test.js
```diff
@@ -1,83 +1,167 @@
...
-      ),
-    ),
+    name: 'byte array',
     value: uint8ArrayToImmutableArrayBuffer(
       new Uint8Array([0x68, 0x65, 0x6c, 0x6c, 0x6f]),
     ),
   },
   {
-    syrup: `${sel('hello')}`,
+    name: 'selector',
```
kriskowal, line 56: Will need a note/issue to track follow-up to go back to Symbol, since we got consensus on that. Not in scope for your change.

.github/workflows/ci.yml
```diff
@@ -382,3 +382,73 @@ jobs:
...
+        uses: actions/checkout@v4
+        with:
+          path: endo
+
+      - name: Checkout OCapN Test Suite
+        uses: actions/checkout@v4
+        with:
+          # https://github.com/ocapn/ocapn-test-suite/pull/24
+          repository: kumavis/ocapn-test-suite
+          ref: 08d4a3a132f301a54ab2d86f02e237efc9bd3093
```
kriskowal, line 402: In the future, I expect that we will be making many changes to the test suite in advance of acceptance and standardization and I’ll make a point to create an endojs/ocapn-test-suite fork for that later.

packages/ocapn/test/client.test.js
kumavis: this is just for nice logging of deeply nested .cause chains

packages/ocapn/test/client.test.js
kumavis: open to suggestions

Issue #1837, opened 2023-10-25
A valid JS code is considered HTML comment


nazarhussain:
## Describe the bug

I am trying out a big bundled JS code in SES and faced with following error: 

```
TypeError#1: Failed to load module "./html_comment_code.js" in package "test" (1 underlying failures: Possible HTML comment rejected at .../html_comment_code.js:5. (SES_HTML_COMMENT_REJECTED)
```
The file contains valid JS code, apparently it seems that SES is considering a specific JS expression as HTML comment. 

## Steps to reproduce

```js
// html_comment_code.js
// That code is extracted from the big bundled package, so sorry for the variables names. 

const _0n$6 = 0;

export function pow2(x, power, modulo) {
  let res = x;
  while (power-- > _0n$6) {
    res *= res;
    res %= modulo;
  }
  return res;
}

```

Here is the code which I am trying to execute above in SES: 

```js
import "ses";
import url from "node:url";
import bundleSource from "@endo/bundle-source";
import {importBundle} from "@endo/import-bundle";
import {StaticModuleRecord} from "@endo/static-module-record";

lockdown();

const resolveHook = (spec, referrer) => {
  new URL(spec, referrer).toString();
};

const importHook = async (moduleSpecifier) => {
    const fullPath = import.meta.resolve(moduleSpecifier);

    if (fullPath.startsWith("file://")) {
      const sourceBundlePath = url.fileURLToPath(fullPath);
      const sourceBundleP = await bundleSource(sourceBundlePath);

      // To check if we had a valid bundle 
      const bundle = await importBundle(sourceBundleP, {endowments: {console}});
      console.log(bundle);

      return new StaticModuleRecord(sourceBundleP, moduleSpecifier);
    }
};

const c = new Compartment(
  {
    print: harden(console.log),
  },
  {},
  {
    name: "test",
    resolveHook,
    importHook,
  }
);

await c.import("./html_comment_code.js");

```

## Expected behavior

This code is simple JS expression and should work well in SES. 

## Platform environment

  System:
    OS: macOS 14.0
    CPU: (8) x64 Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
    Memory: 645.75 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.7.0 - ~/.asdf/installs/nodejs/20.7.0/bin/node
[...]

PR #3026, opened 2025-12-17
feat(evasive-transform): add meaning-preserving evasive transform for imports in strings

naugtur:
## Description

With compartment-mapper introducing support for dynamic imports, the old regexp-based ways in which we evaded import censorship in SES are no longer viable and AST based transform can be the only meaning-preserving one.

### Open questions

- [x] what about `import (` - with whitespace? We could add complexity to the transform or assume they're rare enough and let them hit censorship.
  - will implement as `split` by regexp unless somebody stops me
- [x] Should it be opt-in or opt-out? 
  - opt-out by unanimous opinion from people present in Endo meeting
  - [x] implement
- [x] Does a step in the process between compartment-mapper and SES exist where the supported dynamic import call has been changed to its implementation and the only occurences of `import(` that are left need not be preserved? Could we use a simple regexp-replace there for a cheaper way to get the same result?
  - not a welcome option to create a potential for invalid programs in ModuleSource constructor
- [x] Do we even support `eval('var a = import("somemodule")')` or intend to ?

### Security Considerations

- Could the new transform be targeted to unlock malicious meaning in an otherwise harmless code? (it's meaning-preserving and AST based, so it seems unlikely)

### Scaling Considerations

AST transform is slower than regexp, but for cases where `import(` doesn't exist in a string, it could be faster than a regexp search o the entire string.

- needs documenting after we decide opt-in or out
- [x] naming things is hard

### Testing Considerations

- added a testcase. 
- [x] once we figure out opt-in/out a more end2end test in compartment-mapper would be nice.
- more template string tests could help, the transform for them is more complex than I'd like, with room for off-by-one errors.

with the transform being meaning-preserving, this should not be breaking for anyone. Also, considering opt-in


packages/evasive-transform/src/transform-ast.js
```diff
@@ -16,6 +16,126 @@ const traverse = /** @type {typeof import('@babel/traverse')['default']} */ (
...
+          right: expressions[i],
+        };
+      }
+      p.replaceWith(expr);
+    }
+  }
+
+  // Handle template literals (multiline strings)
+  // `import(` -> `imp${}ort(`
+  if (type === 'TemplateLiteral') {
```
naugtur: skip tagged TemplateLiterals here

packages/evasive-transform/test/evade-censor.test.js
```diff
@@ -91,3 +91,36 @@ test('evadeCensor() - successful source transform w/ source map, source URL & un
...
+
+// import in a string will triger censorship in SES
+const evadeThat = `
+    // HTML comment <!-- should be evaded -->
+    // import comment ...import('some-module');
+    const result = eval("...import('some-module'); await import(\\"other\\");");
+    const result2 = eval('...import(\\'some-module\\'); await import("other");');
+    const multilineimport = \`
+        await import('some-module');
+    \`;
```
naugtur, line 156:
```
String.dedent` import("") `
```

packages/evasive-transform/src/transform-ast.js
naugtur: done

packages/evasive-transform/src/generate.js
```diff
@@ -96,6 +98,11 @@ export const generate = (ast, options) => {
     source,
   );
 
+  if (options && options.preventHtmlCommentRegression) {
+    // Workaround for generator creating `-->` from `-- >` in code generated from the AST
+    result.code = result.code.replace(/-->/g, '-- >');
+  }
```
naugtur: @kriskowal @mhofman is this controversial at all?

packages/evasive-transform/src/transform-strings.js
```diff
@@ -0,0 +1,171 @@
+const evadeRegexp = /import\s*\(|<!--|-->/g;
```
naugtur: Note this now provides multiple evasions in one go

packages/evasive-transform/src/index.js
```diff
@@ -21,6 +21,8 @@ import { generate } from './generate.js';
  * @property {boolean} [elideComments] - Replace comments with an ellipsis but preserve interior newlines.
  * @property {import('./parse-ast.js').SourceType} [sourceType] - Module source type
  * @property {boolean} [useLocationUnmap] - deprecated, vestigial
+ * @property {boolean} [noStringTransform] - if true, will skip evasive transforms on string and template literals.
+ * @property {boolean} [preventHtmlCommentRegression] - If true, will prevent `-->` from appearing in generated code.
```
naugtur: @kriskowal For the record, I'm not proud of these names, but got nothing better at the moment.

packages/evasive-transform/test/snapshots/evade-censor.test.js.md
```diff
@@ -170,3 +170,53 @@ Generated by [AVA](https://avajs.dev).
...
+        const result2 = eval("...im"+"port('some-module'); await im"+"port(\\"other\\");");␊
+        const multilineimport =\`␊
+            console.log(${a})␊
+            await im${""}port('some-module');␊
+            console.log(${b})␊
+        \`;␊
+    ␊
+        const taggedtemplate = String.dedent\`␊
+            await import('some-module');␊
+        \`;␊
```
gibson042: Why is there this difference between tagged vs. untagged template literal contents?

And as an aside, I think this test would be better if the input and expected output appeared in the same file (i.e., not using `t.snapshot`). It's valuable for high volume, but this use seems to come down on the wrong side of the tradeoff.

packages/evasive-transform/src/generate.js
gibson042: Is there a test case covering that this won't affect the contents of strings in the output?

packages/evasive-transform/test/snapshots/evade-censor.test.js.md
mhofman: Because the tag function would see an extra split and empty string value in it. Where without a tag the effects of having an extra hole filled with an empty string are unobservable.

gibson042: I know it's more code to get right, but if we're making this change, we should be thorough about it, e.g.:
```js
        const taggedtemplate = makeEvadedTemplateApplier(String.dedent)`
            await ${IMPORT}('some-module');
        `;
```
where `IMPORT` is a unique global sentinel such as `Symbol('import')` and `makeEvadedTemplateApplier` is a global function like
```js
const makeEvadedTemplateApplier =
  fn =>
  (T0, ...A0) => {
    const argCount = A0.length;
    const T1 = getTemplateMapping(T0) || [];
    const args = [T1];
    if (!T1.length) {
      const raw0 = T0.raw;
      const raw1 = [];
      for (let i = 0, j = 0; i <= argCount; ++i, ++j) {
        T1[j] = T0[i];
        raw1[j] = raw0[i];
        while (i < argCount && A0[i] === IMPORT) {
          i++;
          T1[j] += `import${T0[i]}`;
          raw1[j] += `import${raw0[i]}`;
        }
      }
      defineProperty(T1, 'raw', {
        ...{ writable: false, enumerable: false, configurable: false },
        value: freeze(raw1),
      });
      setTemplateMapping(T0, freeze(T1));
    }
    for (let i = 0, j = 0; i < argCount; ++i) {
      if (A0[i] !== IMPORT) args[++j] = A0[i];
    }
    return apply(fn, undefined, args);
  };
```

(which AFAIK is unobservable, until/unless https://github.com/tc39/proposal-array-is-template-object advances to Stage 4)

kriskowal: We would have to put all that runtime in a single line of code, and it’s a lot. Might have to do some hard trade-off calculus. I’m in favor of making monotonic progress until we hit a cost cliff.

gibson042: I don't know to what extent minification is on the table for injected helpers like the above, but if we are comfortable with it then the cost need be no more than 482 bytes, and only necessary when the source text being transformed has one or more tagged template literals with static contents that would be affected.

packages/evasive-transform/src/generate.js
naugtur: I'm not sure what you mean here. if a string contains an html coment sequence, it'll get disconnected. So there should be nothing left in strings to affect. Is that what you mean? Or what kind of fixture would the test need?

packages/evasive-transform/test/snapshots/evade-censor.test.js.md
naugtur: It's about the potential issues with sourcemaps or having to find a place to put it. Also, the level of complexity I'm not comfortable with introducing. 
It's a complex fix for a very unlikely problem. And if a module contains a tagged template string with an import statement in it, it's likely problematic in many other ways. 

HTML in tagged template strings is somewhat common, but as a virtual dom component, where HTML comments don't make sense.
I fail to find an example of tagged template strings containing hardcoded javascript with imports in it.

I'd like to avoid shipping a tagged template string transform for now.

packages/evasive-transform/src/generate.js
gibson042: Code like `const htmlCommentClose = "-->";` (in which the sequence appears as string contents) must not be affected.

naugtur: OH, it would be affected.
It's not meaning-preserving, that's why I made it opt-in. 
We've been doing that same transform in LavaMoat for a long time and it didn't bite us (this particular one at least) so I thought we could put it in at the much better time. 
Maybe a scarier name is gonna make this feature more palatable.

Anyway, I'll extract it into a separate PR to proceed with the part we most care about.

naugtur: Ok, I've replaced this with a potentially cheaper trick that introduces an empty comment inbetween `a--/**/>b`

Now the only controversy is the new regexp evasion. @gibson042 check it out :)

packages/evasive-transform/src/transform-code.js
```diff
@@ -0,0 +1,215 @@
...
+/**
+ * Transforms RegExp literals containing "import" to use a character class
+ * to break the pattern detection.
+ *
+ * `/import(/` -> `/im[p]ort(/`
+ *
+ * @param {import('@babel/traverse').NodePath} p
+ * @returns {void}
+ */
+export const evadeRegexpLiteral = p => {
```
naugtur, line 191: @gibson042 @mhofman this is a new addition to this PR

it's meaning preserving unless the regexp is listing characters to match, including an open brace like this: `/[import(]+/` - which as regular expressions go seems unlikely enough. Do we want to avoid that with one extra test to see if a ( exists between matching `[]` ?

packages/evasive-transform/test/snapshots/evade-censor.test.js.md
```diff
@@ -170,3 +170,81 @@ Generated by [AVA](https://avajs.dev).
...
+> Snapshot 1
+
+    `␊
+        /* HTML comment <!=- should be evaded -=>*/␊
+        var HTMLstring ="<!"+"-- should be evaded --"+">";␊
+        var HTMLtString =\`<!${""}-- should be evaded --${""}>\`;␊
+        /* import comment ...IMPORT('some-module');*/␊
+        const result = eval("...im"+"port('some-module'); await im"+"port(\\"other\\");");␊
+        const result2 = eval("...im"+"port('some-module'); await im"+"port(\\"other\\");");␊
+        const multilineimport =\`␊
```
gibson042: There's strange loss of the space between `=` and ``` \` ``` here.

packages/evasive-transform/src/index.js
```diff
@@ -21,6 +21,7 @@ import { generate } from './generate.js';
  * @property {boolean} [elideComments] - Replace comments with an ellipsis but preserve interior newlines.
  * @property {import('./parse-ast.js').SourceType} [sourceType] - Module source type
  * @property {boolean} [useLocationUnmap] - deprecated, vestigial
+ * @property {boolean} [noCodeTransforms] - if true, will skip evasive transforms on string and template literals.
```
gibson042: Naming clarity suggestion:
```suggestion
 * @property {boolean} [onlyComments] - if true, will limit transformation to
     comment contents, preserving code positions within each line
```

packages/evasive-transform/src/transform-ast.js
```diff
@@ -57,6 +67,16 @@ export function transformAst(ast, { elideComments = false } = {}) {
...
       (innerComments || []).forEach(node => transformComment(node));
       (trailingComments || []).forEach(node => transformComment(node));
+      if (!noCodeTransforms) {
+        evadeStrings(p);
+        evadeTemplates(p);
+
+        // Prevent `-->` from appearing in output (HTML comment end marker)
+        evadeDecrementGreater(p);
+
+        // evade import ( in RegExp literals
```
gibson042:
```suggestion
        // evade `import(` in RegExp literals
```

packages/evasive-transform/src/transform-code.js
erights: We worked out the safe regexp escaping issues in arguing about `RegExp.escape`. I forgot when I approved `im[p]ort`. Should be `im\x70ort`.

```js
> RegExp.escape('p')
'\\x70'
> /(im\x70ort)/.exec('ximporty')
[ 'import', 'import', index: 1, input: 'ximporty', groups: undefined ]
> /([im\x70ort(]+)/.exec('qiopz')
[ 'iop', 'iop', index: 1, input: 'qiopz', groups: undefined ]
```

erights: Glad to see that long argument was worth it!

packages/evasive-transform/test/snapshots/evade-censor.test.js.md
erights: https://github.com/tc39/proposal-array-is-template-object/issues/21

packages/evasive-transform/src/transform-code.js
```diff
@@ -0,0 +1,215 @@
...
+      lastIndex = nextSplitIndex;
+    }
+
+    // Add final chunk
+    expr = addStringToExpressions(expr, value.substring(lastIndex));
+
+    evadeRegexp.lastIndex = 0; // Reset for next use
+
+    p.replaceWith(expr);
+  }
```
gibson042: I'd very much prefer _not_ having code that messes with `lastIndex`.
```suggestion
  // Break up problematic substrings, e.g. `"import("` -> `"im"+"port("`.
  /** @type {import('@babel/types').Expression | undefined} */
  let expr;
  let lastIndex = 0;
  for (const match of value.matchAll(evadeRegexp)) {
    const index = match.index + 2;
    const part = value.substring(lastIndex, index);
    expr = !expr
      ? { type: 'StringLiteral', value: part }
      : addStringToExpressions(expr, part);
    lastIndex = index;
  }
  if (expr) {
    expr = addStringToExpressions(expr, value.substring(lastIndex));
    p.replaceWith(expr);
  }
```

packages/evasive-transform/src/transform-code.js
```diff
@@ -0,0 +1,215 @@
...
+};
+
+export const evadeTemplates = p => {
+  /** @type {import('@babel/types').Node} */
+  const { type } = p.node;
+
+  // Handle template literals (multiline strings)
+  // `import(` -> `im${}port(`
+  // The transform is only meaning-preserving if not part of a TaggedTemplateExpression, so these need to be excluded until a motivating case shows up. It should be possible to wrap the tag with a function that omits expressions we insert, but that's a lot of work to do preemptively.
```
gibson042:
```suggestion
  // The transform is only meaning-preserving if not part of a TaggedTemplateExpression, so these need to be excluded until a motivating case shows up. It should be possible to wrap the tag with a function that omits expressions we insert, but that's a lot of work to do preemptively.
  // https://github.com/endojs/endo/pull/3026#discussion_r2632507228
```

packages/evasive-transform/src/transform-code.js
```diff
@@ -0,0 +1,215 @@
...
+    if (evadeRegexp.test(quasi.value.raw)) {
+      needsTransform = true;
+      evadeRegexp.lastIndex = 0;
+      break;
+    }
+  }
+
+  if (!needsTransform) {
+    return;
+  }
```
gibson042:
```suggestion
  // Check if any quasi needs transformation
  if (!quasis.some(quasi => quasi.value.raw.match(evadeRegexp))) return;
```

packages/evasive-transform/src/transform-code.js
```diff
@@ -0,0 +1,215 @@
...
+        lastIndex = nextSplitIndex;
+      }
+
+      // Add final chunk of this quasi
+      addQuasi(quasiValue.substring(lastIndex));
+
+      evadeRegexp.lastIndex = 0;
+    } else {
+      newQuasis.push(quasi);
+    }
```
gibson042, line 158: Similar simplification as above (but not tested):
```suggestion
    // We're not currently preserving raw vs. cooked literal data.
    const quasiValue = quasi.value.raw;

    let lastIndex = 0;
    for (const match of quasiValue.matchAll(evadeRegexp)) {
      const index = match.index + 2;
      const raw = quasiValue.substring(lastIndex, index);
      if (lastIndex === 0) {
        // Literal text up to our first cut point.
        newQuasis.push({
          type: 'TemplateElement',
          value: { raw, cooked: raw },
          tail: false,
        });
      } else {
        addQuasi(raw);
      }
      lastIndex = index;
    }
    if (lastIndex !== 0) {
      addQuasi(quasiValue.substring(lastIndex));
    } else {
      newQuasis.push(quasi);
    }
```

packages/evasive-transform/src/transform-code.js
gibson042: I second this `p` → `\x70` suggestion; transformation should not (must not?) change the semantics of any regular expresson. And it might as well cover the rest of `evadeRegexp` as well, using the same approach.

packages/evasive-transform/src/transform-code.js
```diff
@@ -0,0 +1,215 @@
...
+    p.node.type === 'BinaryExpression' &&
+    p.node.operator === '>' &&
+    p.node.left.type === 'UpdateExpression' &&
+    p.node.left.operator === '--' &&
+    !p.node.left.trailingComments?.length
+  ) {
+    // Add an empty block comment to force a space between -- and >
+    p.node.left.trailingComments = [{ type: 'CommentBlock', value: '' }];
+  }
+};
```
gibson042, line 228: Comments can be stripped by downstream tools, recreating the problem—`$x-->$y` → `[$x--][0]>$y` or `(0,$x--)>$y` is better IMO (cf. https://github.com/endojs/endo/issues/1837#issuecomment-2136074644 ).

packages/evasive-transform/src/transform-code.js
naugtur: Great! I'll use that.

packages/evasive-transform/test/snapshots/evade-censor.test.js.md
mhofman: I sure would like to avoid that complexity as well.

packages/evasive-transform/src/transform-code.js
naugtur: But evasive transforms are applied to code last by design, before feeding code to SES. I don't know of a usecase for evasive transforms where the resulting string gets modified again.

We could still switch to it. 
The only argument against is it could impact optimizations/inlining by pushing code across a complexity threshold or not matching a pattern someone hardcoded as optimizable. Is that a concern? 
Should I go to the issue and comment there?

packages/evasive-transform/test/snapshots/evade-censor.test.js.md
naugtur: I think it's just babel. Even in whitespace-preserving mode this space is at the end of a line in some meaning, so it's probably not represented anywhere.
Are we worried?

Also, impressive you noticed it

packages/evasive-transform/test/snapshots/evade-censor.test.js.md
naugtur: I tentatively volunteered to co-champion it. Feed me your concerns.

packages/evasive-transform/src/transform-code.js
naugtur: done

Added a test that confirms the regex actually work so that we don't make an off-by-one error in escaping backslashes.

packages/evasive-transform/src/transform-code.js
naugtur: I know `test` is multiple times faster than `match`, especially in V8, so I put in the extra effort to use `test` - my initial implementation was splitting by regexp.

Your version is significantly shorter and that's even appealing to me more than the lastIndex concern, but I feel like I should check what the perf impact would be.

packages/evasive-transform/test/snapshots/evade-censor.test.js.md
gibson042: Not worried, just confused.

packages/evasive-transform/src/transform-code.js
gibson042: I was under the impression that this code is mostly used at build time—how important do we consider its performance to be?

At any rate, it looks to me like the `matchAll` implementation is slower, but only very slightly so for inputs that need no evasion (probably the dominant case by far) and adding something like 50% for inputs that need lots of evasion. Using `push(node)` in place of `expr = addStringToExpressions(expr, nodeValue)` for simplicity:
```
byTest_zero  [100] sync 586.26 ops/ms (1.71e+3 ns/op) after 115 15360-count observations
byMatch_zero [100] sync 552.41 ops/ms (1.81e+3 ns/op) after 108 15360-count observations
byTest_one   [100] sync 384.76 ops/ms (2.60e+3 ns/op) after 76 15360-count observations
byMatch_one  [100] sync 269.74 ops/ms (3.71e+3 ns/op) after 106 7680-count observations
byTest_many  [100] sync 348.98 ops/ms (2.87e+3 ns/op) after 137 7680-count observations
byMatch_many [100] sync 214.68 ops/ms (4.66e+3 ns/op) after 84 7680-count observations

byTest_zero  [500] sync 145.71 ops/ms (6.86e+3 ns/op) after 102 4307-count observations
byMatch_zero [500] sync 143.08 ops/ms (6.99e+3 ns/op) after 101 4267-count observations
byTest_one   [500] sync 128.86 ops/ms (7.76e+3 ns/op) after 106 3658-count observations
byMatch_one  [500] sync 114.12 ops/ms (8.76e+3 ns/op) after 106 3246-count observations
byTest_many  [500] sync 83.18 ops/ms (12.0e+3 ns/op) after 112 2237-count observations
byMatch_many [500] sync 57.54 ops/ms (17.4e+3 ns/op) after 93 1874-count observations

byTest_zero  [1000] sync 74.32 ops/ms (13.5e+3 ns/op) after 101 2228-count observations
byMatch_zero [1000] sync 74.5 ops/ms (13.4e+3 ns/op) after 98 2286-count observations
byTest_one   [1000] sync 69.48 ops/ms (14.4e+3 ns/op) after 99 2111-count observations
byMatch_one  [1000] sync 63.9 ops/ms (15.7e+3 ns/op) after 60 3246-count observations
byTest_many  [1000] sync 43.69 ops/ms (22.9e+3 ns/op) after 59 2237-count observations
byMatch_many [1000] sync 30.44 ops/ms (32.9e+3 ns/op) after 96 953-count observations
```
<details>

```console
$ cat setup.js
const L = 1000;
const noReplacements = Array.from({ length: L }, (_, i) =>
  Array.from({ length: scale }, (_, j) => `IMPORT${j}(${i});`)
    .join("\n")
    .toLowerCase(),
);
const oneReplacement = noReplacements.map(s =>
  s.replaceAll("import42(", "import(").toUpperCase().toLowerCase(),
);
const manyReplacements = noReplacements.map(s =>
  s
    .replaceAll(/import[0-9]+8\(/g, "import(")
    .toUpperCase()
    .toLowerCase(),
);
const evadeByTest = (
  (evadeRegexp = /import\s*\(|<!--|-->/g) =>
  value => {
    if (!evadeRegexp.test(value)) return;
    const parts = [];
    evadeRegexp.lastIndex = 0;
    let match = evadeRegexp.exec(value);
    let lastIndex = match.index + 2;
    parts.push({ type: "StringLiteral", value: value.substring(0, lastIndex) });
    while ((match = evadeRegexp.exec(value)) !== null) {
      const nextSplitIndex = match.index + 2;
      parts.push({ type: "StringLiteral", value: value.substring(lastIndex, nextSplitIndex) });
      lastIndex = nextSplitIndex;
    }
    parts.push({ type: "StringLiteral", value: value.substring(lastIndex) });
    evadeRegexp.lastIndex = 0;
    return parts;
  }
)();
const evadeByMatch = (
  (evadeRegexp = /import\s*\(|<!--|-->/g) =>
  value => {
    if (value.search(evadeRegexp) === -1) return;
    const parts = [];
    let lastIndex = 0;
    for (const match of value.matchAll(evadeRegexp)) {
      const index = match.index + 2;
      const part = value.substring(lastIndex, index);
      parts.push({ type: "StringLiteral", value: part });
      lastIndex = index;
    }
    if (parts.length === 0) return;
    parts.push({ type: "StringLiteral", value: value.substring(lastIndex) });
    return parts;
  }
)();
for (const [name, fn] of Object.entries({ evadeByTest, evadeByMatch })) {
  const input = `import(foo); import(bar);`,
    want = `im ‖ port(foo); im ‖ port(bar);`,
    got = fn(input)
      .map(part => part.value)
      .join(" ‖ ");
  if (got !== want)
    throw Error(`${name}: wanted ${JSON.stringify(want)}, got ${JSON.stringify(got)}`);
}
let i = L - 1;

$ esbench.mjs -h V8 -b3 -a scale=100,500,1000 -s "$(cat setup.js)" \
   byTest_zero:'result = evadeByTest(noReplacements[++i % L]);' \
  byMatch_zero:'result = evadeByMatch(noReplacements[++i % L]);' \
   byTest_one:'result = evadeByTest(oneReplacement[++i % L]);' \
  byMatch_one:'result = evadeByMatch(oneReplacement[++i % L]);' \
   byTest_many:'result = evadeByTest(manyReplacements[++i % L]);' \
  byMatch_many:'result = evadeByMatch(manyReplacements[++i % L]);'
```

</details>

> my initial implementation was splitting by regexp

`split` with a full-capture pattern is my favorite approach, but sadly it's slower by about two orders of magnitude.

packages/evasive-transform/src/transform-code.js
gibson042: > But evasive transforms are applied to code last by design, before feeding code to SES. I don't know of a usecase for evasive transforms where the resulting string gets modified again.

I think we had one at Agoric, but my memory is fuzzy.

> We could still switch to it. The only argument against is it could impact optimizations/inlining by pushing code across a complexity threshold or not matching a pattern someone hardcoded as optimizable. Is that a concern?

I certainly don't consider that a concern.

> Should I go to the issue and comment there?

Feel free, but we're all aware that `[$x--][0]>$y` is slightly more expensive than `$x--/**/>$y` (it's just unlikely for that extra cost to matter in realistic code).

packages/evasive-transform/src/transform-code.js
naugtur: This code is meant to predominantly be used at runtime. It processes the code that compartment-mapper feeds into SES.

Thanks for doing all the benchmarking. I was going to work on it now and saw you already did.
As infrequent as the replacements are, 50% doesn't concern me. I'll take it.

packages/evasive-transform/src/transform-code.js
naugtur: We've had a conversation at the Endo meeting that concluded in "if we doublecheck that there's no config to define a transform that runs after this one in Endo it's fine to use a comment"

Since you were not present, it's definitely not final.

I am concerned about the optimization and inlining

packages/evasive-transform/test/snapshots/evade-censor.test.js.md
gibson042: The primary concern is taking away the ability to locally virtualize tagged template literals, which is a steep price to pay for what seems to be [negligible benefit](https://github.com/tc39/proposal-array-is-template-object/issues/22).

packages/evasive-transform/src/transform-code.js
gibson042: Since I'm in a microbenchmarking mood, I ran the numbers—`[$x--][0]>$y` seems to take 25% to 70% longer than `$x--/**/>$y` in browser engines, but both forms are about as fast as anything can be so performance is *not* the relevant consideration here. However, there appears to be negligible overhead (if any) for `(0,$x--)>$y`:
```
#### V8
comment sync 33.9e+3 ops/ms (29.5e+0 ns/op) after 414 245760-count observations
comma   sync 33.7e+3 ops/ms (29.6e+0 ns/op) after 412 245760-count observations
array   sync 24.7e+3 ops/ms (40.4e+0 ns/op) after 278 267429-count observations (1 restart)

#### SpiderMonkey
comment sync 6977.19 ops/ms (143e+0 ns/op) after 137 153600-count observations
comma   sync 6927.05 ops/ms (144e+0 ns/op) after 64 329143-count observations
array   sync 5445.73 ops/ms (184e+0 ns/op) after 107 153600-count observations

#### JavaScriptCore
comment sync 7629.27 ops/ms (131e+0 ns/op) after 94 245760-count observations
comma   sync 7685.11 ops/ms (130e+0 ns/op) after 94 245760-count observations
array   sync 4557.14 ops/ms (219e+0 ns/op) after 112 122880-count observations
```
<details>

```sh
for engine in V8 SpiderMonkey JavaScriptCore; do
  esbench.mjs -h "$engine" -b3 \
    -s '
      const length = 10000, numbers = Array.from({ length }, (_, i) => i);
      let i = length - 1;
    ' \
    comment:'result = numbers[++i % length]; result = result--/**/>result;' \
    comma:'result = numbers[++i % length]; result = (0,result--)>result;' \
    array:'result = numbers[++i % length]; result = [result--][0]>result;'
done
```

</details>

So I'd advocate for the comma expression transformation, although this is not a hill worth dying on and I don't have a strong objection to comment-injection either.

erights: I like the comma transform. What are the advantages of comment injection over comma transform?

naugtur: /addressed in meeting. update pushed

naugtur:
All finished.

@gibson042 @kriskowal @boneskull

packages/evasive-transform/README.md
```diff
@@ -2,9 +2,40 @@
...
+This package provides a function which transforms source code which would otherwise be rejected outright by SES.
+The transform is meaning-preserving.
+
+It covers sequences resembling HTML comments inside of:
+- comments
+- strings
+- template strings (but not tagged template strings)
+- code itself, eg `if (a-->b)`
+
+It covers sequences resembling dynamic import use inside of:
```
gibson042:
```suggestion
It covers sequences resembling dynamic `import` inside of:
```

packages/evasive-transform/README.md
```diff
@@ -2,9 +2,40 @@
...
 
-This package provides a function which transforms comments contained in source code which would otherwise be rejected outright by SES.
+This package provides a function which transforms source code which would otherwise be rejected outright by SES.
+The transform is meaning-preserving.
+
+It covers sequences resembling HTML comments inside of:
+- comments
+- strings
+- template strings (but not tagged template strings)
+- code itself, eg `if (a-->b)`
```
gibson042: We should also evade `-->` to e.g. `--\x3E` inside regular expression literals, e.g.
```markdown
It covers sequences resembling HTML comments and dynamic `import` inside of:
- comments
- strings
- template strings (but not tagged template strings)
- regular expression literals

and additionally covers sequences resembling HTML comments inside of code itself (e.g., `if (a-->b)`).
```

packages/evasive-transform/README.md
gibson042: ``` `@property {ІᛖРΟᏒТ('foo').Bar} bar` ``` seems to now be inaccurate.
```suggestion
 * The resulting file will now contain `@property {IMPORT('foo').Bar} bar`, which SES will allow (and TypeScript no longer understands, but that should be fine for the use-case). 
```

.changeset/stale-pots-whisper.md
```diff
@@ -0,0 +1,5 @@
+---
+"@endo/evasive-transform": minor
+---
+
+feat(evasive-transform): add meaning-preserving evasive transform for imports and sequences resembling HTML comments in code
```
gibson042: Aren't the changes more sweeping than this?

packages/evasive-transform/src/transform-ast.js
```diff
@@ -57,6 +67,14 @@ export function transformAst(ast, { elideComments = false } = {}) {
       }
       (innerComments || []).forEach(node => transformComment(node));
       (trailingComments || []).forEach(node => transformComment(node));
+      if (!onlyComments) {
+        evadeStrings(p);
+        evadeTemplates(p);
+        // evade `import(` in RegExp literals
+        evadeRegexpLiteral(p);
+        // Prevent `-->` from appearing in output (HTML comment end marker)
+        evadeDecrementGreater(p);
```
gibson042, line 74:
```suggestion
        evadeStrings(p);
        evadeTemplates(p);
        evadeRegexpLiteral(p);
        evadeDecrementGreater(p);
```

packages/evasive-transform/src/transform-code.js
```diff
@@ -0,0 +1,180 @@
...
+  if (p.node.type !== 'RegExpLiteral') {
+    return;
+  }
+  /** @type {import('@babel/types').RegExpLiteral} */
+  const node = p.node;
+  const { pattern } = node;
+
+  if (pattern.match(importRegexp)) {
+    node.pattern = pattern.replace(importRegexp, 'im\\x70ort$1');
+  }
```
gibson042: Suggestion for also evading HTML-like comments in regular expression literals:
```suggestion
  if (pattern.match(evadeRegexp)) {
    node.pattern = pattern.replace(
      evadeRegexp,
      s => regexpReplacements[s[0]] + s.substring(1),
    );
  }
```
where `regexpReplacements` is `{ i: '\\x69', '<': '\\x3C', '-': '\\x2D' }`

(and corresponding test coverage as well)

packages/evasive-transform/test/snapshots/evade-censor.test.js.md
gibson042: Turns out that this lost space stems from generation against our locationless synthetic AST nodes, and it's fairly easy to fix:
```patch
diff --git i/packages/evasive-transform/src/transform-code.js w/packages/evasive-transform/src/trans
form-code.js
index 1ef096d9d..f186e7ee5 100644
--- i/packages/evasive-transform/src/transform-code.js
+++ w/packages/evasive-transform/src/transform-code.js
@@ -1,6 +1,32 @@
 const evadeRegexp = /import\s*\(|<!--|-->/g;
 const importRegexp = /import(\s*\()/g;
 
+/**
+ * Copy the location from one AST node to another (round-tripping through JSON
+ * to sever references), updating the target's end position as if it had zero
+ * length.
+ *
+ * @param {import('@babel/types').Node} target
+ * @param {import('@babel/types').Node} src
+ */
+const adoptStartFrom = (target, src) => {
+  try {
+    const srcLoc = src.loc;
+    if (!srcLoc) return;
+    const loc = /** @type {typeof srcLoc} */ (
+      JSON.parse(JSON.stringify(srcLoc))
+    );
+    const start = loc?.start;
+    target.loc = loc;
+    if (!start) return;
+    target.loc.end = /** @type {typeof start} */ (
+      JSON.parse(JSON.stringify(start))
+    );
+  } catch (_err) {
+    // Ignore errors; this is purely opportunistic.
+  }
+};
+
 /**
  * Creates a BinaryExpression adding two expressions
  *
@@ -36,6 +62,7 @@ export const evadeStrings = p => {
     expr = !expr
       ? { type: 'StringLiteral', value: part }
       : addStringToExpressions(expr, part);
+    if (lastIndex === 0) adoptStartFrom(expr, p.node);
     lastIndex = index;
   }
   if (expr) {
@@ -128,11 +155,14 @@ export const evadeTemplates = p => {
     newQuasis[newQuasis.length - 1].tail = true;
   }
 
-  p.replaceWith({
+  /** @type {import('@babel/types').Node} */
+  const replacement = {
     type: 'TemplateLiteral',
     quasis: newQuasis,
     expressions: newExpressions,
-  });
+  };
+  adoptStartFrom(replacement, p.node);
+  p.replaceWith(replacement);
 };
 
 /**
```

.changeset/stale-pots-whisper.md
naugtur: Not really. Feature-wise, the only thing not described here is the opt-out option.

Do we want the changesets to be more elaborate?

packages/evasive-transform/test/snapshots/evade-censor.test.js.md
naugtur: I'd like to put this in a separate PR

packages/evasive-transform/README.md
naugtur: This must have been inaccurate for a while now.

packages/evasive-transform/src/transform-code.js
naugtur: The inconsistency in letter choice between `import` replacements is a tiny bit annoying now, but I like the trick of only worrying about the first character

.changeset/stale-pots-whisper.md
gibson042: Ah, I was parsing it wrong.
```suggestion
feat(evasive-transform): add meaning-preserving transformation of expressions and literals containing content that would otherwise be rejected by SES for looking like dynamic import or HTML-like comments
```

packages/evasive-transform/src/transform-code.js
naugtur: I ended up adding a test so we don't spend any more time confused by the pre/post decrement

packages/evasive-transform/README.md
```diff
@@ -2,9 +2,35 @@
...
+
+### Options
+
+Both `evadeCensor` and `evadeCensorSync` accept a source string as the first argument and an options object as the second argument:
+
+| Option | Type | Description |
+|--------|------|-------------|
+| `sourceUrl` | `string` | The URL or filename of the source file. Used for source map generation and error messages. |
+| `sourceMap` | `string \| object` | Optional. An existing source map (as JSON string or object) to be updated with the transform's mappings. |
+| `sourceType` | `'script' \| 'module'` | Optional. Specifies whether the source is a CommonJS script (`'script'`) or an ES module (`'module'`). When provided, it helps the parser handle the code correctly. |
```
erights, line 26: Which does this default to if omitted?

packages/evasive-transform/README.md
```diff
@@ -2,9 +2,35 @@
...
+| `elideComments` | `boolean` | Optional. If `true`, removes comment contents while preserving newlines. Defaults to `false`. |
```
erights, line 27: I hope it also preserves columns. Yes?
```suggestion
| `elideComments` | `boolean` | Optional. If `true`, removes comment contents while preserving newlines and columns. Defaults to `false`. |
```

packages/evasive-transform/README.md
```diff
@@ -2,9 +2,35 @@
...
+| `onlyComments` | `boolean` | Optional. If `true`, limits transformation to comment contents only, leaving code unchanged. Defaults to `false`. |
```
erights, line 28: What's the motivation of this option?

packages/evasive-transform/test/snapshots/evade-censor.test.js.md
gibson042: I'd like it in this PR so there's never a `master` state where code transformation loses whitespace, but if you do defer it then make sure to open an issue.

packages/evasive-transform/test/snapshots/evade-censor.test.js.md
```diff
@@ -170,3 +170,84 @@ Generated by [AVA](https://avajs.dev).
...
+
+## evadeCensor() - actual evasions in ESM
+
+> Snapshot 1
+
+    `␊
+        /* HTML comment <!=- should be evaded -=>*/␊
+        var HTMLstring ="<!"+"-- should be evaded --"+">";␊
+        var HTMLtString =\`<!${""}-- should be evaded --${""}>\`;␊
+        /* import comment ...IMPORT('some-module');*/␊
```
gibson042, line 182: @erights @michaelfig @kriskowal   `//…` comments are converted to `/*…*/` comments, but the reason seems to be a mystery. @boneskull asked in #1812, and [that answer](https://github.com/endojs/endo/pull/1812#issuecomment-1756461619) was "_I do not remember being aware that we changed comments this way, so I also do not remember why_". It doesn't exactly cause problems because the start sequences are equal-length, although forcing the conversion at all does run against our efforts to preserve input as much as possible. Should we contemplate improving the behavior, or is it best to let this lie?

erights: We should improve this behavior. But it is low urgency and low importance. Still, we should indeed minimize gratuitous changes.

erights: If the question is: should we do it in this PR, depends on level of effort. If easy and quick, sure.

packages/evasive-transform/README.md
naugtur: That's a very good question.
I didn't touch that part, just documented the existence of the option.

The option is being used to decide whether return statements should be allowed
```
allowReturnOutsideFunction: opts.sourceType === 'script',
```
And the way it's written here, it defaults to "not `script`" 

But the `sourceType` itself is being passed to babel, where it's defined as defaulting to "script" 

```
/**
     * Indicate the mode the code should be parsed in.
     * Can be one of "script", "commonjs", "module", or "unambiguous". Defaults to "script".
     * "unambiguous" will make @babel/parser attempt to guess, based on the presence
     * of ES6 import or export statements.
     * Files with ES6 imports and exports are considered "module" and are otherwise "script".
     *
     * Use "commonjs" to parse code that is intended to be run in a CommonJS environment such as Node.js.
     */
    sourceType?: SourceType;
```

Let's create an issue about that and fix it separately. We might need to talk about how to avoid it being a breaking change.

packages/evasive-transform/README.md
naugtur: The way it's been documented in code (and I stole from there) there doesn't seem to be a promise of preserving columns there. 
Which makes sense when 
```
// a line of comment
```
turns into 
```
// 
```
and also when internal newlines in a multiline comment are kept but the content is gone.

But it seems like it's not replacing inline comments with whitespace to match.

packages/evasive-transform/README.md
naugtur: It's an opt-in to the behavior from before this PR. 
This package used to only transform the contents of the comments and none of the code.

packages/evasive-transform/test/snapshots/evade-censor.test.js.md
naugtur: This PR did not touch the comments transforms and I have a slight preference to keep the scope of it as such. I only added more coverage and more visible test results, which surfaced the issue.

Happy to start a separate branch to work on the fixes for comments and white space issues. They're also low priority to me, but if we find a fix that's not costly in performance or contributor confusion, I'm all for it.

packages/evasive-transform/test/snapshots/evade-censor.test.js.md
naugtur: I'm not sure I understand why target.loc.end is set to start, but I need this merged, so I'll take it.

The first JSON pass should be enough to flatten everything, so I'm changing the second one into a shallow copy with a spread operator.

naugtur: Applied and updated the tests.

packages/evasive-transform/src/transform-code.js
```diff
@@ -0,0 +1,216 @@
...
+  try {
+    const srcLoc = src.loc;
+    if (!srcLoc) return;
+    const loc = /** @type {typeof srcLoc} */ (
+      JSON.parse(JSON.stringify(srcLoc))
+    );
+    const start = loc?.start;
+    target.loc = loc;
+    if (!start) return;
+    target.loc.end = /** @type {typeof start} */ ({ ...start });
```
gibson042:
```suggestion
    const loc = /** @type {typeof srcLoc} */ (
      JSON.parse(JSON.stringify(srcLoc))
    );
    const start = loc?.start;
    target.loc = loc;
    // Text of the new node is likely shorter than text of the old (e.g.,
    // "import(<url>)" -> "im"), and in such cases we don't ever want rendering
    // of the new node to claim too much real estate so we future-proof by
    // making it appear to be zero-width and trusting in recovery of the actual
    // location immediately afterwards.
    if (start) target.loc.end = /** @type {typeof start} */ ({ ...start });
```

packages/evasive-transform/src/transform-code.js
```diff
@@ -0,0 +1,216 @@
...
+
+export const evadeStrings = p => {
+  /** @type {import('@babel/types').Node} */
+  const { type } = p.node;
+  if (type !== 'StringLiteral') {
+    return;
+  }
+  /** @type {import('@babel/types').StringLiteral} */
+  const { value } = p.node;
+  // Break up problematic substrings, e.g. `"import("` -> `"im"+"port("`.
```
gibson042: Let's promote these comments into proper documentation.
```suggestion
/**
 * Break up problematic substrings into concatenation expressions, e.g.
 * `"import("` -> `"im"+"port("`.
 *
 * @param {import('@babel/traverse').NodePath} p
 */
export const evadeStrings = p => {
  const { node } = p;
  if (node.type !== 'StringLiteral') {
    return;
  }
  const { value } = node;
```

packages/evasive-transform/src/transform-code.js
```diff
@@ -0,0 +1,216 @@
...
+  // The transform is only meaning-preserving if not part of a TaggedTemplateExpression, so these need to be excluded until a motivating case shows up. It should be possible to wrap the tag with a function that omits expressions we insert, but that's a lot of work to do preemptively.
+  // https://github.com/endojs/endo/pull/3026#discussion_r2632507228
+  if (
+    type !== 'TemplateLiteral' ||
+    p.parent.type === 'TaggedTemplateExpression'
+  ) {
+    return;
+  }
+  /** @type {import('@babel/types').TemplateLiteral} */
+  const node = p.node;
```
gibson042:
```suggestion
/**
 * Break up problematic substrings in template literals with empty-string
 * expressions, e.g. `import(` -> `im${''}port(`.
 *
 * @param {import('@babel/traverse').NodePath} p
 */
export const evadeTemplates = p => {
  const { node } = p;

  // The transform is only meaning-preserving if not part of a
  // TaggedTemplateExpression, so these need to be excluded until a motivating
  // case shows up. It should be possible to wrap the tag with a function that
  // omits expressions we insert, but that's a lot of work to do preemptively.
  // https://github.com/endojs/endo/pull/3026#discussion_r2632507228
  if (
    node.type !== 'TemplateLiteral' ||
    p.parent.type === 'TaggedTemplateExpression'
  ) {
    return;
  }
```

packages/evasive-transform/src/transform-code.js
```diff
@@ -0,0 +1,216 @@
...
+ *
+ * @param {import('@babel/traverse').NodePath} p
+ * @returns {void}
+ */
+export const evadeRegexpLiteral = p => {
+  if (p.node.type !== 'RegExpLiteral') {
+    return;
+  }
+  /** @type {import('@babel/types').RegExpLiteral} */
+  const node = p.node;
```
gibson042:
```suggestion
  const { node } = p;
  if (node.type !== 'RegExpLiteral') {
    return;
  }
```

packages/evasive-transform/src/transform-code.js
```diff
@@ -0,0 +1,216 @@
...
+  ) {
+    return;
+  }
+  /** @type {import('@babel/types').TemplateLiteral} */
+  const node = p.node;
+
+  const { quasis } = node;
+
+  // Check if any quasi needs transformation
+  if (!quasis.some(quasi => quasi.value.raw.match(evadeRegexp))) return;
```
gibson042: Since we do care about performance, it's worth noting that `search` is faster than `match` while still being `lastIndex`-safe.
```suggestion
  if (!quasis.some(quasi => quasi.value.raw.search(evadeRegexp) !== -1)) return;
```

packages/evasive-transform/src/transform-code.js
```diff
@@ -0,0 +1,216 @@
...
+ */
+export const evadeRegexpLiteral = p => {
+  if (p.node.type !== 'RegExpLiteral') {
+    return;
+  }
+  /** @type {import('@babel/types').RegExpLiteral} */
+  const node = p.node;
+  const { pattern } = node;
+
+  if (pattern.match(evadeRegexp)) {
```
gibson042:
```suggestion
  if (pattern.search(evadeRegexp) !== -1) {
```

packages/evasive-transform/src/transform-code.js
```diff
@@ -0,0 +1,216 @@
...
+    p.node.operator === '>' &&
+    p.node.left.type === 'UpdateExpression' &&
+    p.node.left.operator === '--'
+  ) {
+    // Wrap the UpdateExpression in a SequenceExpression: (0, x--)
+    p.node.left = {
+      type: 'SequenceExpression',
+      expressions: [{ type: 'NumericLiteral', value: 0 }, p.node.left],
+    };
+  }
```
gibson042, line 227:
```suggestion
  const { node } = p;
  if (
    node.type === 'BinaryExpression' &&
    node.operator === '>' &&
    node.left.type === 'UpdateExpression' &&
    node.left.operator === '--' &&
    !node.left.prefix
  ) {
    // Wrap the UpdateExpression in a SequenceExpression: (0, x--)
    node.left = {
      type: 'SequenceExpression',
      expressions: [{ type: 'NumericLiteral', value: 0 }, node.left],
    };
  }
```
(we should also test that _prefix decrements_ as in `--x>y` are not modified)

packages/evasive-transform/test/snapshots/evade-censor.test.js.md
gibson042: > I'm not sure I understand why target.loc.end is set to start, but I need this merged, so I'll take it.

The short answer is "future-proofing"—I don't want generation to skip forward too far if it starts paying attention to `end`. Suggested explanatory comment text at https://github.com/endojs/endo/pull/3026#discussion_r2833745432

> The first JSON pass should be enough to flatten everything, so I'm changing the second one into a shallow copy with a spread operator.

Sure, that was me being overcautious about possible future introduction of new non-primitive properties.

packages/evasive-transform/src/transform-code.js
naugtur: good catch!

packages/evasive-transform/src/transform-code.js
naugtur: typescript's ability to follow this was relying on some of the redundancy and now lint is failing. I'll bring some of the redundancy back

PR #3031, opened 2025-12-29
Ocapn: improved handling of remote promises + etc

kumavis:
### core
- [x] Refactor to use HandledPromise handler `returnedP` as answerPromise
- [x] Refactor to correctly set sendOnly handler hooks
- [x] Rejection of remote promises on disconnect
### etc
- [x] Fix setting of Error cause
- [x] Test against python gc tests

---

> [!NOTE]
> Modernizes OCapN messaging and lifecycle semantics with clearer promise handling and accurate refcounting.
> 
> - Refactor client/ref-kit to use the HandledPromise-returned promise as the registered `answer`, add explicit `apply*SendOnly` pathways, and pipeline using internal vs external promises; `fulfill/break` now use send-only
> - Introduce precise refcount APIs: `recordSentSlot`/`recordReceivedSlot` and per-message counting; remove implicit increments from lookups; adjust tables to drop/GC correctly and reject pending settlers on `destroy(reason)`
> - Propagate disconnect reasons: `ocapn.abort(reason: Error)` tears down connection and rejects all pending remote answers; local answers are preserved on the responder
> - Codecs/descriptors: wrap errors with `cause`, `DescAnswer` reads via `getLocalAnswerValue`
> - CI/test: update Python test-suite ref, capture server logs; expose GC for test server and add GC utilities; add comprehensive unit/integration tests for pipelining, GC (export/answer), E vs E.sendOnly, disconnect behavior
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ab639bd3b88eb7ef563980f3e9a36ef73ad4d665. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -435,16 +435,24 @@ jobs:
       - name: Start OCapN test server
         working-directory: endo
         run: |
-          yarn workspace @endo/ocapn run test:python &
+          yarn workspace @endo/ocapn run test:python > /tmp/ocapn-server.log 2>&1 &
           echo $! > /tmp/ocapn-server.pid
           # Wait for server to be ready
           sleep 5
+          # Show initial server output
+          cat /tmp/ocapn-server.log
 
       - name: Run Python test suite
         working-directory: ocapn-test-suite
         run: |
           source ./venv/bin/activate
           python ./test_runner.py 'ocapn://test.tcp-testing-only?host=127.0.0.1&port=22046'
+
+      - name: Show OCapN server logs on failure
+        if: failure()
+        run: |
+          echo "=== OCapN Server Logs ==="
+          cat /tmp/ocapn-server.log || echo "No log file found"
 
       - name: Stop OCapN test server
         if: always()


packages/ocapn/test/_gc-util.js
```diff
@@ -0,0 +1,115 @@
...
+ */
+const getEngineGC = () => {
+  if (typeof globalThis.gc === 'function') {
+    return globalThis.gc;
+  }
+
+  // Try to expose GC using v8 flags
+  v8.setFlagsFromString('--expose_gc');
+  const gc = vm.runInNewContext('gc');
+  v8.setFlagsFromString('--no-expose_gc');
```
kriskowal, line 20: Cute. Also: This Should Not Be.

packages/ocapn/test/_gc-util.js
```diff
@@ -0,0 +1,115 @@
...
+ * @returns {Promise<void>}
+ */
+const gcAndFinalize = async () => {
+  await new Promise(setImmediate);
+  await new Promise(setImmediate);
+  engineGC();
+  await new Promise(setImmediate);
+  await new Promise(setImmediate);
+  await new Promise(setImmediate);
+};
```
kriskowal, line 49: Criminal.

packages/ocapn/test/_gc-util.js
```diff
@@ -0,0 +1,115 @@
...
+ */
+export const waitForSentinelGc = async (timeoutMs = 5000) => {
+  await undefined;
+  /** @type {object | null} */
+  let sentinel = {};
+  const collected = new Promise(resolve => {
+    sentinelRegistry.register(sentinel, () => resolve(undefined));
+  });
+  // Make sentinel unreachable
+  sentinel = null;
```
kriskowal, line 68: I’m sure this works well enough, but with generational GC, I don’t think that there’s an implication that one object’s collection implies that every unreachable object was collected. It might be close enough for V8, but also, if it were inherently good enough for V8, I would not expect to have to call `gc` more than once to ensure all unreachable objects were collected.

kriskowal (approved):

packages/ocapn/test/_gc-util.js
kumavis: precedent https://github.com/endojs/endo/blob/f1b9e5224353f88ca5a0d905c986b1da4b108433/packages/captp/test/gc-and-finalize.js#L67-L94

packages/ocapn/test/_gc-util.js
kumavis: precedent https://github.com/endojs/endo/blob/f1b9e5224353f88ca5a0d905c986b1da4b108433/packages/promise-kit/test/promise-kit.test.js#L12-L20

kumavis: precedent https://github.com/endojs/endo/blob/f1b9e5224353f88ca5a0d905c986b1da4b108433/packages/captp/test/engine-gc.js

packages/ocapn/test/_gc-util.js
kumavis: precedent https://github.com/endojs/endo/blob/f1b9e5224353f88ca5a0d905c986b1da4b108433/packages/promise-kit/test/promise-kit.test.js#L34-L63

kumavis: though here we could just use the background gc scheduling using in the ocapn python test

PR #3032, opened 2025-12-31
Cancellation primitive

kriskowal:
## Description

This PR introduces a new `@endo/cancel` package that provides `Promise<never>`-based cancellation tokens with synchronous observation capability, designed for use with Endo and CapTP.

### Core API

- **`makeCancelKit(parentCancelled?)`** - Creates a `{ cancelled, cancel }` pair where `cancelled` is a `Promise<never>` with a synchronous `.cancelled` getter. Supports hierarchical cancellation via optional parent token.

### Operators

- **`allMap(values, fn, parentCancelled)`** - Maps over values with cancellation propagation; cancels all on first rejection
- **`anyMap(values, fn, parentCancelled?)`** - Races cancellable jobs; cancels remaining after first success
- **`delay(ms, parentCancelled)`** - Cancellable delay using ambient `setTimeout`
- **`makeDelay(setTimeout, clearTimeout)`** - Factory for `delay` with injectable timer functions (from `delay-lite`)

### Web API Integration

- **`toAbortSignal(cancelled)`** - Converts `Cancelled` → `AbortSignal` for use with `fetch` and other web APIs
- **`fromAbortSignal(signal)`** - Converts `AbortSignal` → `Cancelled` for integrating web cancellation sources

### Adoption in Daemon and CLI

This PR also refactors `@endo/daemon` and `@endo/cli` to use `makeCancelKit` instead of the previous pattern of using `makePromiseKit` for cancellation. This demonstrates the ergonomic improvements and validates the design against real-world usage.

### Pass-Style Relaxation

**This PR includes a surgical change to `@endo/pass-style` that allows promises with a `cancelled` getter to be passable.**

Previously, `safe-promise.js` rejected any promise with string-named own properties. This change adds a specific exception for `cancelled`:

```js
// packages/pass-style/src/safe-promise.js
if (key === 'cancelled') {
  // Allow a non-enumerable `cancelled` getter for cancellation tokens.
  // This property is local-only and will not be passed over CapTP.
  const cancelledDesc = getOwnPropertyDescriptor(pr, 'cancelled');
  assert(cancelledDesc !== undefined);
  return (
    (hasOwn(cancelledDesc, 'get') ||
      (reject &&
        reject`Own 'cancelled' must be an accessor property, not a data property: ${q(cancelledDesc)}`)) &&
    (!cancelledDesc.enumerable ||
      (reject &&
        reject`Own 'cancelled' must not be enumerable: ${q(cancelledDesc)}`))
  );
}
```

**Constraints enforced:**
1. Must be an **accessor property** (getter), not a data property
2. Must be **non-enumerable**

**Why this is safe:**
- When promises are serialized over CapTP/marshal, only a slot reference is passed (e.g., `"&0"`)
- The `cancelled` property is naturally omitted during serialization—it never crosses the wire
- The getter provides local-only synchronous observation, which is the intended design

**Rationale:**
- Enables `Cancelled` tokens (promises with `.cancelled` getter) to be passed as arguments to remote methods
- The remote side receives a normal promise that rejects on cancellation
- The local side retains synchronous observability for performance-critical checks

### Security Considerations

- All exports are hardened with `harden()`
- The synchronous `.cancelled` getter is intentionally non-passable over CapTP—only the promise behavior crosses boundaries
- The pass-style relaxation is narrowly scoped: only a non-enumerable `cancelled` getter is allowed, no other own properties

### Scaling Considerations

None.

Provides README and DESIGN.

### Testing Considerations

- Tests with coverage of all functions in `@endo/cancel`
- All 112 daemon tests pass
- All 10 CLI tests pass

- Designed to integrate with existing CapTP infrastructure
- Web API integration enables gradual adoption alongside `AbortController`
- The pass-style change is backward compatible—existing promises without `cancelled` continue to work unchanged

### Upgrade Considerations

Packages using the `makePromiseKit` pattern for cancellation may optionally migrate to `makeCancelKit` for improved ergonomics and the synchronous `.cancelled` getter.

### DIFF
--- a/packages/cancel/DESIGN.md
+++ b/packages/cancel/DESIGN.md
@@ -12,7 +12,8 @@ However, promises do not provide any mechanism for synchronously observing
 their state locally. There are some conditions where synchronous observation
 of cancellation can help a utility avoid unnecessary work.
 
-To that end, we extend the `Promise<never>` to have a `cancelled` getter.
+To that end, `makeCancelKit` returns an `isCancelled` function alongside the
+promise, providing a separate synchronous observation channel.
 
 ## Core API: `makeCancelKit`
 
@@ -20,8 +21,8 @@ The `Promise.withResolvers` API produces a `promise`, `resolve`, and `reject`.
 We envision the eventual addition of `Promise.withCanceller` that returns
 instead `cancelled` and `cancel`.
 
-The `cancelled` is a promise with a `cancelled` own getter that returns:
-- `undefined` - cancellation has not been requested
+The `isCancelled` function returns:
+- `false` - cancellation has not been requested
 - `true` - cancellation has been requested
 
 This package anticipates this eventual evolution of the language and exports
@@ -30,10 +31,10 @@ This package anticipates this eventual evolution of the language and exports
 ```js
 import { makeCancelKit } from '@endo/cancel';
 
-const { cancelled, cancel } = makeCancelKit();
+const { cancelled, cancel, isCancelled } = makeCancelKit();
 
 // Synchronous check
-if (cancelled.cancelled) {
+if (isCancelled()) {
   return; // Skip unnecessary work
 }
 
@@ -48,13 +49,15 @@ cancel(Error('User requested abort'));
 
 ### Hierarchical Cancellation
 
-`makeCancelKit` accepts an optional `parentCancelled` token, enabling
-hierarchical cancellation patterns. When a parent token is cancelled,
-all child tokens automatically cancel with the same reason:
+`makeCancelKit` accepts an optional `parentCancelled` token and an optional
+`parentIsCancelled` function, enabling hierarchical cancellation patterns.
+When a parent token is cancelled, all child tokens automatically cancel with
+the same reason. If `parentIsCancelled` is provided and returns true at
+creation time, the child is synchronously cancelled:
 
 ```js
-const { cancelled: parentCancelled, cancel: cancelParent } = makeCancelKit();
-const { cancelled: childCancelled } = makeCancelKit(parentCancelled);
+const { cancelled: parentCancelled, cancel: cancelParent, isCancelled: parentIsCancelled } = makeCancelKit();
+const { cancelled: childCancelled, isCancelled: childIsCancelled } = makeCancelKit(parentCancelled, parentIsCancelled);
 
 cancelParent(Error('Operation aborted'));
 // childCancelled is now also cancelled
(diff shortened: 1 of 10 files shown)


kriskowal:
Aside, I noticed an opportunity to use makePromiseKit, but that exercise is dashed because that produces a hardened promise, and so we cannot introduce a new `cancelled` property.

packages/pass-style/src/safe-promise.js
rekmarks, line 1: Seems like this warrants a couple of test cases in `test/safe-promise.test.js`.

packages/cancel/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+Creates a cancellation kit containing a cancellation token and a cancel function.
+
+```js
+import { makeCancelKit } from '@endo/cancel';
+
+const { cancelled, cancel } = makeCancelKit();
+
+// Check synchronously if cancelled
+console.log(cancelled.cancelled); // undefined
```
rekmarks: Not `false`?

packages/cancel/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+If a parent cancellation token is provided, cancellation automatically propagates
+from the parent to the child:
+
+```js
+const { cancelled: parentCancelled, cancel: cancelParent } = makeCancelKit();
+const { cancelled: childCancelled } = makeCancelKit(parentCancelled);
+
+cancelParent(Error('Parent cancelled'));
+// childCancelled is now also cancelled
```
rekmarks, line 59:
```suggestion
console.log(childCancelled.cancelled) // true
```

packages/cancel/README.md
```diff
@@ -0,0 +1,284 @@
...
+};
+
+// Callback signature for allMap and anyMap
+type CancellableCallback<T, R> = (
+  value: T,
+  index: number,
+  cancelled: Cancelled
+) => R | Promise<R>;
+```
+
```
rekmarks, line 275:
```suggestion
```

packages/cancel/DESIGN.md
rekmarks, line 1: This mainly seems to restate the readme. Delete?

rekmarks (approved):
It's good! Some non-blocking feedback.

kriskowal:
Having discussed this change with @erights out-of-band, I’ve posted an amendment (to be absorbed before merging) that removes the `cancelled` accessor for promises, making them passable without weakening the passable invariants _locally_. In the new mode, when you receive a `cancelled` promise through the eventual-send boundary, the expectation is that you would recover synchronous observability by using `makeCancelKit`, which produces both a dependent `cancelled` promise and `isCancelled()` function that caches the cancelled state for synchronous checks going forward. I’ve also added `isCancelled` to the abort signal adapters.

PR #3033, opened 2026-01-01
feat(ocapn): CBOR alternative encoding

kriskowal:
## Description

This change introduces an experimental CBOR codec as an alternative to Syrup for OCapN message serialization. The implementation enables OCapN messages to be parsed by any RFC 8949 compliant CBOR decoder, allowing more of the ecosystem's existing tools to come to bear and making the wire format strictly a compact, machine-readable format.

### New: CBOR Codec Implementation

- RFC 8949 compliant canonical encoding
- All integers encoded as bignums (Tags 2/3) for arbitrary precision
- Float64 values with canonical NaN representation
- Symbols via Tag 280
- Records via Tag 27 (generic array)
- Tagged values via Tag 55799 (self-described CBOR)
- BMP-only string validation (rejects surrogate code points)

### New: CBOR Diagnostic Notation Codec

A text-based codec for human-readable CBOR representation:
- `src/cbor/diagnostic/encode.js` - CBOR bytes → diagnostic string
- `src/cbor/diagnostic/decode.js` - Diagnostic string → JavaScript values
- `src/cbor/diagnostic/util.js` - Hex conversion and comparison helpers

This enables writing test cases in readable form and debugging encoding issues.

### New: Codec Interface Abstraction

`src/codec-interface.js` defines abstract `OcapnReader` and `OcapnWriter` interfaces that both Syrup and CBOR implement. This enables:
- Swapping codecs without changing application code
- Type-safe codec injection
- Future codec negotiation in netlayers

### Codec Architecture: Dependency Injection

Applications import only the codec they need:

```js
// For CBOR
import { makeCborWriter, makeCborReader } from '@endo/ocapn/cbor/index.js';

// For Syrup
import { makeSyrupWriter, makeSyrupReader } from '@endo/ocapn/syrup/index.js';
```

This replaces a hypothetical central factory, ensuring unused implementations are tree-shaken.

### Test Infrastructure Enhancements

- **Dual-codec testing**: All codec tests now run with both Syrup and CBOR
- **AVA macros**: Table-driven interop tests using AVA macro pattern
- **Interop tests**: Validates CBOR output parses with the `cbor` npm package
- **Snapshot updates**: All codec snapshots now include both Syrup and CBOR variants

### Documentation

- `docs/cbor-encoding.md` - Complete specification covering:
  - All OCapN type mappings to CBOR
  - Canonical encoding rules
  - CapTP operation and descriptor formats
  - Passable data and slot reference design
- `docs/codec-usage.md` - Usage patterns for both codecs
- `src/cbor/README.md` - Implementation overview
- `src/syrup/README.md` - Updated with codec ID

### Security Considerations

- **Canonical encoding**: CBOR output is deterministic, enabling signature verification without re-serialization
- **BMP string validation**: Rejects strings with surrogate code points (outside Basic Multilingual Plane), consistent with Syrup
- **Tag 24 for signed envelopes**: The `desc:sig-envelope` design wraps signed content in embedded CBOR (Tag 24) to enable signature verification against bytewise representation

### Scaling Considerations

- 65535-byte message limit consistent with Noise Protocol framing
- Slot references separated from passable body enables forwarding without re-serialization

The `docs/cbor-encoding.md` specification is comprehensive and self-contained. Users wanting to implement an OCapN codec in another language should be able to do so from this document alone.

### Testing Considerations

Existing Syrup tests parameterized to cover CBOR.

Inludes tests that verify that cross-reference legibility of generated CBOR messages with an off-the-shelf, generic CBOR implementation.

Round-trip and canonicalization tests.

None.

### Upgrade Considerations

None.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -650,6 +650,7 @@ __metadata:
     "@noble/hashes": "npm:^1.8.0"
     ava: "catalog:dev"
     c8: "catalog:dev"
+    cbor: "npm:^9.0.2"
     eslint: "catalog:dev"
     tsd: "catalog:dev"
     typescript: "npm:~5.9.2"
@@ -3476,6 +3477,15 @@ __metadata:
   dependencies:
     nofilter: "npm:^3.0.2"
   checksum: 10c0/49b59036c340ab0c6f4fa39aaf37ed6cb2bec6d54ec27b45a03f5df0fcd5767594b0abb5cbf44d69bdd8593d6a2e131c3e7017c511bacf05f5aa4ff2af82d07d
+  languageName: node
+  linkType: hard
+
+"cbor@npm:^9.0.2":
+  version: 9.0.2
+  resolution: "cbor@npm:9.0.2"
+  dependencies:
+    nofilter: "npm:^3.1.0"
+  checksum: 10c0/709d4378067e663107b3d63a02d123a7b33e28946b4c5cc40c102f2f0ba13b072a79adc4369bb87a4e743399fce45deec30463fc84d363ab7cb39192d0fe5f30
   languageName: node
   linkType: hard
 
@@ -8170,7 +8180,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"nofilter@npm:^3.0.2":
+"nofilter@npm:^3.0.2, nofilter@npm:^3.1.0":
   version: 3.1.0
   resolution: "nofilter@npm:3.1.0"
   checksum: 10c0/92459f3864a067b347032263f0b536223cbfc98153913b5dce350cb39c8470bc1813366e41993f22c33cc6400c0f392aa324a4b51e24c22040635c1cdb046499


packages/ocapn/docs/cbor-encoding.md
```diff
@@ -0,0 +1,843 @@
...
+| Boolean | Simple values 20/21 | `false`, `true` |
+| Integer | Tag 2/3 + byte string | `2(h'01')` = 1 |
+| Float64 | Float64 (major 7) | `1.5` |
+| String | Text string (major 3) | `"hello"` |
+| ByteArray | Byte string (major 2) | `h'deadbeef'` |
+| Symbol | Tag 280 + text string | `280("method")` |
+| List | Array (major 4) | `[1, 2, 3]` |
+| Struct | Map (major 5) | `{"a": 1, "b": 2}` |
+| Record | Tag 27 + array | `27(["op:deliver", ...])` |
+| Tagged | Tag 55799 + array | `55799(["decimal", "3.14"])` |
```
gibson042: [Self-Described CBOR](https://www.rfc-editor.org/rfc/rfc8949#name-self-described-cbor) defines tag number 55799 to "_not impart any special semantics on the data item that it encloses; that is, the semantics of the tag content enclosed in tag number 55799 is exactly identical to the semantics of the tag content itself_", which makes it a poor choice for representing OCapN Tagged values. I think instead we would want to be consistent with https://github.com/ocapn/ocapn/pull/247 and represent them as Records.
```suggestion
| Tagged | Record encapsulating tag string and value | `27(["desc:tag", "decimal", "3.14"])` |
```

packages/ocapn/docs/cbor-encoding.md
```diff
@@ -0,0 +1,843 @@
...
+| Error | Record with label "desc:error" | `27(["desc:error", "message"])` |
+| Target (in-band) | Record marker | `27([280("target")])` |
+| Promise (in-band) | Record marker | `27([280("promise")])` |
+| Error (in-band) | Record with message | `27([280("error"), "TypeError"])` |
```
gibson042: [CBOR tag number 27](https://cbor.schmorp.de/generic-object) suggests (if not requires) that the first element be a string ("_The typename [in array [typename, constructargs...]] is usually a class name, or another string that indicates the name of the type_"), and I think we should honor that expectation.
```suggestion
| Target (in-band) | Record marker | `27(["target"])` |
| Promise (in-band) | Record marker | `27(["promise"])` |
| Error (in-band) | Record with message | `27(["error", "TypeError"])` |
```

packages/ocapn/docs/cbor-encoding.md
```diff
@@ -0,0 +1,843 @@
...
+| Major Type | Value | Usage in OCapN |
+|------------|-------|----------------|
+| 0 | Unsigned int | Length prefixes only (not for OCapN integers) |
+| 1 | Negative int | Length prefixes only (not for OCapN integers) |
+| 2 | Byte string | ByteArray values |
+| 3 | Text string | String values |
+| 4 | Array | List values |
+| 5 | Map | Struct values |
+| 6 | Tag | Symbols (280), Records (27), Bignums (2/3), Tagged (55799) |
+| 7 | Float/Simple | Float64, Boolean, Null, Undefined |
```
gibson042, line 53: This encoding prohibits indefinite-length values, right?

packages/ocapn/docs/cbor-encoding.md
```diff
@@ -0,0 +1,843 @@
...
+
+| Value | Encoding | Diagnostic |
+|-------|----------|------------|
+| 0 | `C2 40` | `2(h'')` |
+| 1 | `C2 41 01` | `2(h'01')` |
+| 255 | `C2 41 FF` | `2(h'ff')` |
+| 256 | `C2 42 01 00` | `2(h'0100')` |
+| -1 | `C3 40` | `3(h'')` |
+| -2 | `C3 41 01` | `3(h'01')` |
+| -256 | `C3 41 FF` | `3(h'ff')` |
```
gibson042:
```suggestion
The byte string uses minimal encoding: no leading zero bytes, even for numeric
value zero (which is represented using an empty byte string).

**Examples**:

| Value | Encoding | Diagnostic |
|-------|----------|------------|
| 0 | `C2 40` | `2(h'')` |
| 1 | `C2 41 01` | `2(h'01')` |
| 255 | `C2 41 FF` | `2(h'ff')` |
| 256 | `C2 42 01 00` | `2(h'0100')` |
| -1 | `C3 40` | `3(h'')` |
| -2 | `C3 41 01` | `3(h'01')` |
| -256 | `C3 41 FF` | `3(h'ff')` |
```

packages/ocapn/docs/cbor-encoding.md
```diff
@@ -0,0 +1,843 @@
...
+
+```
+0x60-0x7B <length info> <UTF-8 bytes>
+```
+
+- Major type 3 (text string)
+- Length encoded per CBOR rules:
+  - 0-23: Single byte `0x60 | length`
+  - 24-255: `0x78` + 1 byte length
+  - 256-65535: `0x79` + 2 byte big-endian length
```
gibson042:
`````suggestion
### String

UTF-8 encoded text string of definite byte length. Note that UTF-8 encoding
precludes surrogate code points (U+D800–U+DFFF).

```
0x60-0x7B <length info> <UTF-8 bytes>
```

- Major type 3 (text string)
- Length encoded per CBOR rules:
  - 0-23: Single byte `0x60 | length`
  - 24-255: `0x78` then 1 byte length
  - 256-65535: `0x79` then 2 byte big-endian length
`````

File: packages/ocapn/docs/cbor-encoding.md [hunk 6]
```diff
@@ -0,0 +1,843 @@
...
+
+| Value | Encoding | Diagnostic |
+|-------|----------|------------|
+| "" | `60` | `""` |
+| "a" | `61 61` | `"a"` |
+| "hello" | `65 68 65 6C 6C 6F` | `"hello"` |
+
+### ByteArray
+
+Raw byte sequence.
```
gibson042:
```suggestion
### ByteArray

Raw byte sequence of definite length.
```

packages/ocapn/docs/cbor-encoding.md
```diff
@@ -0,0 +1,843 @@
...
+> ensures consistent encoding regardless of magnitude. CBOR's compact integer
+> encoding could be added as an optimization in a future version.
+
+**Positive integers** (n >= 0): Tag 2 + byte string of big-endian magnitude
+
+```
+0xC2 <bytestring of n>
+```
+
+**Negative integers** (n < 0): Tag 3 + byte string of big-endian (-1 - n)
```
gibson042:
`````suggestion
**Positive integers** (n >= 0): Tag 2 then byte string of big-endian magnitude n

```
0xC2 <bytestring of n>
```

**Negative integers** (n < 0): Tag 3 then byte string of big-endian (-1 - n)
`````

File: packages/ocapn/docs/cbor-encoding.md [hunk 8]
```diff
@@ -0,0 +1,843 @@
...
+| Value | Encoding | Diagnostic |
+|-------|----------|------------|
+| {} | `A0` | `{}` |
+| {"a": 1} | `A1 61 61 C2 41 01` | `{"a": 2(h'01')}` |
+| {"a": 1, "b": 2} | `A2 61 61 C2 41 01 61 62 C2 41 02` | `{"a": 2(h'01'), "b": 2(h'02')}` |
+
+### Record
+
+Tag 27 wrapping an array. The first element is the record label (a string or
+symbol), followed by the record fields.
```
gibson042: As noted above, I'd prefer the first element to always be a string. But while OCapN does have symbols (which I'd also prefer to drop), I can see an argument for using them here. Even in that case, though, I think Record should be defined to constrain the first element to a single type.

packages/ocapn/docs/cbor-encoding.md
```diff
@@ -0,0 +1,843 @@
...
+
+```
+0xD8 0x1B <array>
+```
+
+- Bytes 0-1: Tag 27 (`0xD8` = tag with 1-byte number, `0x1B` = 27)
+- Remaining: Array with label as first element
+
+> **Tag number 27**: This is the "Self-Described CBOR Sequence" tag, commonly
+> used for record-like structures in CBOR-based protocols.
```
gibson042:
```suggestion
> **Tag number 27**: This is the "Serialised language-independent object with
> type name and constructor arguments" tag, used for record-like structures in
> CBOR-based protocols.
```

packages/ocapn/docs/cbor-encoding.md
```diff
@@ -0,0 +1,843 @@
...
+
+**Diagnostic**: `27([280("desc:import-object"), 2(h'05')])`
+
+### Tagged (OCapN Tagged Values)
+
+Tag 55799 wrapping an array of [tag-name, payload].
+
+> **Tag number 55799**: This is the CBOR ["Self-Described CBOR"][SelfDescribe]
+> magic number defined in RFC 8949 §3.4.6. Using it here provides a clear marker
+> for OCapN tagged values while remaining valid CBOR.
```
gibson042: As noted above, I don't think CBOR tag number 55799 is right here, and would instead suggest a specialization of Record.

packages/ocapn/docs/cbor-encoding.md
```diff
@@ -0,0 +1,843 @@
...
+
+**Diagnostic**: `27([280("desc:error"), "Error message here"])`
+
+## Passable Data and Slots
+
+OCapN messages contain "passable" data that may include references to remote
+objects (targets), promises, and errors. To enable efficient message forwarding
+through intermediaries without re-serialization, references are encoded using
+**in-band markers** within an **embedded CBOR body**, with **parallel arrays**
+mapping each marker to its CapTP table position.
```
gibson042, line 361: So references don't bear any identifying metadata such as a slot index, and the idea is instead to use straightforward counting with repeats as necessary, such that targets `[remotable1, remotable2]` vs. `[remotable1, remotable1]` have the same in-band representation but target arrays that vary like e.g.
```diff
-[27(["desc:export", 1), 27(["desc:export", 2)]
+[27(["desc:export", 1), 27(["desc:export", 1)]`
```
?

packages/ocapn/docs/cbor-encoding.md
```diff
@@ -0,0 +1,843 @@
...
+
+1. Fast-forward past the body without parsing nested structures
+2. Remap slot positions in the parallel arrays
+3. Forward the body bytes unchanged
+
+### Embedded CBOR (Tag 24)
+
+The body of passable data is encoded as **Tag 24** (Encoded CBOR data item)
+wrapping a byte string. This standard CBOR tag indicates that the byte string
+contains valid CBOR data.
```
gibson042, line 378: What is the structure of the envelope containing this body and some number of reference arrays? Is it itself a CBOR array or map, are there CBOR tags relevent to that level, etc.?

packages/ocapn/docs/codec-usage.md
```diff
@@ -0,0 +1,184 @@
...
+// Or import both from the index
+import { makeSyrupWriter, makeSyrupReader } from '@endo/ocapn/syrup/index.js';
+```
+
+#### Writing with Syrup
+
+```js
+import { makeSyrupWriter } from '@endo/ocapn/syrup/encode.js';
+
+const writer = makeSyrupWriter({ name: 'my-message' });
```
gibson042: What are the semantics of the `name` option?

packages/ocapn/docs/codec-usage.md
```diff
@@ -0,0 +1,184 @@
...
+
+```js
+import { makeSyrupReader } from '@endo/ocapn/syrup/decode.js';
+
+const reader = makeSyrupReader(bytes, { name: 'my-message' });
+reader.enterRecord();
+const label = reader.readSelectorAsString(); // 'op:deliver'
+const num = reader.readInteger(); // 42n
+const str = reader.readString(); // 'hello'
+reader.exitRecord();
```
gibson042: How does a consumer detect that the data is a Record for which it is appropriate to call `enterRecord` vs. some other kind of value and corresponding different method, and likewise for all other typed reads? Similarly, how does the consumer know how many values are in a Record or List or Struct?

packages/ocapn/docs/cbor-encoding.md
kriskowal: should be merely `[27(["desc:export"]), 27(["desc:export"])]` in-band.

packages/ocapn/docs/codec-usage.md
kriskowal: This appears in logs, only.

packages/ocapn/docs/codec-usage.md
kriskowal: The codec interface is a generalization to straddle Syrup and CBOR, for the time being. To that end, `enter*` take the length prefixes computed up front, which Syrup ignores. The exit is expected when these are exhausted. For CBOR, exiting a record is a no-op.

packages/ocapn/docs/cbor-encoding.md
gibson042: Right, that covers "same in-band representation"... I'm asking about what the (presumably out-of-band) target arrays would look like for `[remotable1, remotable2]` vs. `[remotable1, remotable1]`.

packages/ocapn/docs/codec-usage.md
gibson042: I think that's answering a question I'm not asking... the above code shows `enterRecord()`, then `readSelectorAsString()`, then `readInteger()`, then `readString()`, then `exitRecord()`, which would presumably fail (or at least misbehave) if the input data were not a \<symbol, integer, string> Record. But by what means would the author of that code know to expect such input? I don't want to be painted into a situation where things only work if one has an externally-supplied schema.

packages/ocapn/docs/cbor-encoding.md
```diff
@@ -0,0 +1,843 @@
...
+Within the embedded CBOR body, references to targets, promises, and errors are
+encoded as simple records containing a sequential index (0-based). These markers
+are placeholders that reference entries in parallel arrays outside the body.
+
+#### Target Marker
+
+A reference to a remote object (target) in the CapTP tables.
+
+```
+27([280("target")])
```
dckc: why a `target` string here? why not a CBOR tag number?

(likewise promise and error markers)

kriskowal: Using a CBOR tag number would require a trip to IANA. 27 is a CBOR tag number we’re using to denote a Syrup record, which is analogous to a single-tier Cap’n Web array, e.g., `["date",ms]`. CBOR tag number 280 indicates a symbol. So, this is a lot like the existing `<desc:import-object>` family of descriptors, but we’re collapsing all the targets, promises, and errors to place-holders and all other distinctions moving to slots, as is consistent with Endo’s marshaling. We could add back a gratuitous slot index to each of these, which would make it easier for a human to parse the CBOR diagnostic format.

packages/ocapn/docs/cbor-encoding.md
kriskowal: The body is a CBOR encoded array. There is no additional tag, as writ.

packages/ocapn/docs/cbor-encoding.md
dckc: > Using a CBOR tag number would require a trip to IANA.

'nuff said.

packages/ocapn/docs/codec-usage.md
```diff
@@ -0,0 +1,257 @@
...
+the type of the next value without consuming it:
+
+```js
+const reader = makeCborReader(bytes, { name: 'unknown structure' });
+
+const hint = reader.peekTypeHint();
+switch (hint) {
+  case 'number-prefix':
+    // Could be integer or float - check further or read as appropriate
+    const num = reader.readInteger(); // or readFloat64()
```
gibson042, line 210: Wouldn't `readInteger()` misbehave if the value is actually floating point, and `readFloat64()` misbehave if the value is actually an integer? I see no value in conflating the two distinct types via a single value from `peekTypeHint`.

packages/ocapn/docs/cbor-encoding.md
```diff
@@ -0,0 +1,877 @@
...
+| ByteArray | Byte string (major 2) | `h'deadbeef'` |
+| Symbol | Tag 280 + text string | `280("method")` |
+| List | Array (major 4) | `[1, 2, 3]` |
+| Struct | Map (major 5) | `{"a": 1, "b": 2}` |
+| Record | Tag 27 + array | `27(["deliver", ...])` |
+| Tagged | Record encapsulating tag string and value | `27(["tag", "decimal", "3.14"])` |
+| Error | Record with string label | `27(["error", "message"])` |
+| Target (in-band) | Record marker | `27(["target"])` |
+| Promise (in-band) | Record marker | `27(["promise"])` |
+| Error (in-band) | Record with message | `27(["error", "TypeError"])` |
```
gibson042, line 39: Why does Error appear twice?

packages/ocapn/docs/cbor-encoding.md
```diff
@@ -0,0 +1,877 @@
...
+| Embedded CBOR | Tag 24 + byte string | `24(h'...')` |
+
+## CBOR Major Types Used
+
+| Major Type | Value | Usage in OCapN |
+|------------|-------|----------------|
+| 0 | Unsigned int | Length prefixes only (not for OCapN integers) |
+| 1 | Negative int | Length prefixes only (not for OCapN integers) |
```
gibson042, line 47: "not for OCapN integers"?

packages/ocapn/docs/cbor-encoding.md
```diff
@@ -0,0 +1,877 @@
...
+
+```
+D8 1B                     # Tag 27 (Record)
+  82                      # Array of 2 elements
+    65                    # Text string, 5 bytes
+      65 72 72 6F 72      # "error"
+    <message string>
+```
+
+**Diagnostic**: `27(["error", "Error message here"])`
```
gibson042, line 353: This seems inconsistent with Quick Reference—is the element after "error" a message or a type?

kriskowal: My intention is that the second value in the array is a message and that it be carried in-band, while the respective identifier be carried out-of-band. There should actually be a third element with an OCapN Struct with additional data.

packages/ocapn/docs/codec-usage.md
kriskowal: These methods assert the type matches the expected type by direct observation. This is an artifact of our desire to, in many case, avoid reïfying certain values, and in particular, reïfying symbols as strings when it suits us (often).

kriskowal: The explanation is sloppy. Working on that.

PR #3034, opened 2026-01-01
feat(ocapn-noise): Send expected responder public key in clear text for routing

kriskowal:
This change prepares the experimental OCapN Noise Protocol cryptography so that a relay can see the expected responder’s public key in plain text and use that to route. This will likely limit the use of this cryptography to connection hints specifically denoting relays, and create an obligation to communicate to the user that this kind of connection hint reveals the key of the recipient, and Tor’s right there if you need to hide that information better.

Regardless, I expect most OCapN clients will stand for multiple public keys, one for each profile, and this feature is necessary to complete a handshake on such a peer.

### DIFF
--- a/packages/ocapn-noise/src/bindings.js
+++ b/packages/ocapn-noise/src/bindings.js
@@ -17,13 +17,13 @@
 /**
  * @typedef {object} InitiatorResult
  * @property {SigningKeys} signingKeys - The generated or provided signing keys
- * @property {(syn: Uint8Array) => InitiatorWriteSynResult} initiatorWriteSyn - Function to write SYN message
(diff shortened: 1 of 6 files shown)


packages/ocapn-noise/src/bindings.js
```diff
@@ -270,36 +273,78 @@ export const makeOcapnSessionCryptography = ({
...
+        ),
+      );
+
     let code = wasmInstance.exports.responder_read_syn();
-    if (code === 1) {
+    if (code === 7) {
+      throw new Error(
+        'OCapN Noise Protocol SYN intended for different responder',
+      );
+    } else if (code === 1) {
```
rekmarks: nit: Numerically out of order

PR #3035, opened 2026-01-02
POC broken alt cancellation 3032

erights:

Closes: #XXXX
Refs: #XXXX

## Description

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations

### DIFF
--- a/packages/cancel/src/cancel-kit.js
+++ b/packages/cancel/src/cancel-kit.js
@@ -4,6 +4,67 @@
  * @import { Cancelled, Cancel, CancelKit } from './types.js'
  */
 
+const { apply } = Reflect;
+
+/**
+ * @type {WeakSet<Promise>}
+ */
+const knownCancelledSet = new WeakSet();
+harden(knownCancelledSet);
+
+const thenMethod = Promise.prototype.then;
+
+// TODO more accurate typing of `when`. It is intended to emulate E.when.
+/**
+ * Since this package has no dependencies, it likely needs to work reliably in
+ * a non-ses environment too.
+ *
+ * @template {any} V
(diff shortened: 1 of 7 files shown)


packages/cancel/src/cancel-kit.js
```diff
@@ -20,47 +81,39 @@
...
-    },
-    enumerable: false,
-    configurable: false,
-  });
+  // Propagate cancellation from parent if provided
+  if (parentCancelled) {
+    if (isKnownCancelled(parentCancelled)) {
+      knownCancelledSet.add(cancelled);
+    }
+  }
```
erights, line 98: This is an orthogonal observable change from #3032 . I claim it is an improvement, in that if the parent is already known to be cancelled, then the child is born cancelled. You could adopt this observable change by itself, without the rest of the PR.

Btw, this observable change did not affect any tests.

packages/cancel/src/cancel-kit.js
```diff
@@ -4,6 +4,67 @@
  * @import { Cancelled, Cancel, CancelKit } from './types.js'
  */
 
+const { apply } = Reflect;
+
+/**
+ * @type {WeakSet<Promise>}
+ */
+const knownCancelledSet = new WeakSet();
```
erights, line 12: Turns out that `isKnownCancelled`'s use of this piece of static mutable state does make it a communication channel. This SECURITY BUG seems inherent in this approach, which is why this is a failed experiment. See below.

kriskowal:
Let’s close. I think we’re in agreement to pursue `{cancelled: promise, isCancelled() -> boolean}` in #3032

PR #3036, opened 2026-01-02
feat(exo-stream): Introduce Exo streams

kriskowal:
## Description

This change introduces a new `@endo/exo-stream` package that provides a CapTP abstraction for bridging async iterator streams. The package offers a stable client API designed to allow protocol evolution without modifying clients.

The PR also refactors `@endo/daemon` and `@endo/cli` to use the new `exo-stream` package, removing the old `reader-ref.js` and `ref-reader.js` modules from daemon.

### New Package: `@endo/exo-stream`

Four exported functions provide symmetric conversions between local iterators and remote passable streams:

| Function | Purpose |
|----------|---------|
| `streamIterator` | Convert local `AsyncIterable<Passable>` → remote `PassableStream` |
| `iterateStream` | Convert remote `PassableStream` → local `AsyncIterableIterator<Passable>` |
| `streamBytesIterator` | Convert local `AsyncIterable<Uint8Array>` → remote `PassableBytesStream` |
| `iterateBytesStream` | Convert remote `PassableBytesStream` → local `AsyncIterableIterator<Uint8Array>` |

The byte stream variants automatically encode/decode via base64 for CapTP transport. This is a temporary measure, pending support for byte arrays in CapTP, based on existing support in pass-style for immutable ArrayBuffer that simply hasn’t been fully realized.

### Refactored Modules

**`@endo/daemon`:**
- `src/directory.js`, `src/host.js`, `src/guest.js` → use `streamIterator`
- `src/daemon.js` → uses `iterateBytesStream`
- `src/daemon-node-powers.js` → uses `streamBytesIterator`
- Removed `src/reader-ref.js`, `src/ref-reader.js`, and their re-exports

**`@endo/cli`:**
- `src/commands/store.js`, `make.js`, `install.js`, `bundle.js` → use `streamBytesIterator`
- `src/commands/list.js`, `inbox.js`, `follow.js` → use `iterateStream`
- `src/commands/cat.js` → uses `iterateBytesStream`
- `demo/cat.js` → uses `iterateStream`

### Security Considerations

No new security concerns. The package uses existing hardened patterns (`makeExo`, `harden`) and InterfaceGuards from `@endo/patterns`. The abstraction maintains the same security boundaries as the previous implementation.

### Scaling Considerations

This change implements an improved protocol, no longer relying on `next` invocations. We transit async chains of nodes directly. The initiator provides the synchronizer queue and the responder returns the acknowledgement queue.

- The new package includes a README with usage examples
- `DESIGN.md` provides detailed implementation specifications
- `PROTOCOL.md` provides a draft proposal for the corresponding over-CapTP protocol, as a basis for collaborating on over-CapTP protocols with OCapN community constituents.

### Testing Considerations

- Unit tests in `packages/exo-stream/test/index.test.js` cover all four functions
- Tests include empty streams, single items, large streams (1000 items), and round-trip scenarios
- Existing daemon and cli tests should continue to pass

**Breaking for internal consumers only:**
- `makeIteratorRef`, `makeRefIterator`, `makeReaderRef`, `makeRefReader` are no longer exported from `@endo/daemon`
- These were only used internally within daemon and cli, which have been refactored

### Upgrade Considerations

No upgrade steps needed. The refactoring is internal and the wire protocol remains compatible.


erights (commented):
I love this use of patterns. What they were meant for; why they're passable.

For files like `interfaces.js`, I've generally used the name `typeGuards.js`. I still prefer it for reasons I'll explain.

erights:
> Looking at the file, given that it’s a module that exports a bank of `M.interface`, I find it hard to consider naming it anything else. If we’d named these `M.typeGuard`, I think it would be an easier case.

The argument from consistency seems strong enough. If we decide to go with `interfaces.js` we should rename all the other `typeGuards.js` to `interfaces.js`. In general these files contain both patterns and guards, neither of which are literally types, so none of these names is literally correct. However, they are best understood as paralleling the type structure, and should be easily discoverable by those looking for type-ish stuff.

Let's discuss. We might come up with a name that is both better and more correct than these choices.

packages/daemon/src/daemon.js
```diff
@@ -847,13 +847,19 @@ const makeDaemonCore = async (
...
       async () => {
         await null;
         const values = {
           formulaNumber: await randomHex512(),
-          contentSha512: await contentStore.store(makeRefReader(readerRef)),
+          contentSha512: await contentStore.store(
+            // Use a higher string length limit to accommodate large payloads
+            // like bundles. 10MB base64 ~= 7.5MB binary.
+            await iterateBytesStream(streamRef, {
+              stringLengthLimit: 10_000_000,
```
jcorbin, line 1593: what's 0.5 M(i)B between friends... I guess using base-10 is current year syntactic inertia, but the old ways of `10 * 1024 * 1024` are Still Right There ;-)

packages/daemon/test/endo.test.js
```diff
@@ -421,7 +415,9 @@ test('store with name', async t => {
 
   {
     const { host } = await makeHost(config, cancelled);
-    const readerRef = makeReaderRef([new TextEncoder().encode('hello\n')]);
+    const readerRef = streamBytesIterator([
+      new TextEncoder().encode('hello\n'),
```
jcorbin, line 438: that's far less than 10MB of test data there... is there a case in here with enough juice to even stream more than one chunk? let alone saturate that buffer size above...

packages/exo-stream/DESIGN.md
```diff
@@ -0,0 +1,134 @@
...
+## Protocol Overview
+
+The exo-stream protocol provides non-lossy streams of passable data, with control flow,
+such that the producer does not overwhelm the consumer.
+
+The exo-stream protocol uses bidirectional promise chains for streaming over CapTP.
+By default, the protocol is fully synchronized and just as chatty as naive protocols,
+suitable for deliberate synchronization. With a buffer value in excess of 1, promise
+chain nodes propagate via CapTP before the event loop yields to I/O, keeping the
+responder busy while the initiator processes values:
```
jcorbin: so your first sentence leads in with "producer" and "consumer", which yes good expected...

but then pretty much immediately drop them and say "initiator" and "responder" below

can we at least add an explicit sentence like "Below we will calling the consumer **Initiator** and the producer **Responder**" ... also maybe say "the initiator consumes values" rather than "processes" which rhymes too much with "produces" ;-)

packages/exo-stream/DESIGN.md
```diff
@@ -0,0 +1,134 @@
...
+- **Acknowledgement chain**: Promise chain from responder to initiator.
+
+## Protocol Flow
+
+```mermaid
+sequenceDiagram
+    participant I as Initiator
+    participant R as Responder
+
+    Note over I: Create synchronize chain<br/>(hold resolver locally)
```
jcorbin: s/synchronize/synchronization/ ?

packages/exo-stream/DESIGN.md
```diff
@@ -0,0 +1,134 @@
...
+    Note over I: Pre-resolve buffer synchronizes
+
+    I->>R: stream(synHead)
+    
+    Note over R: Create acknowledge chain<br/>Start pump loop
+
+    R-->>I: ackHead (promise)
```
jcorbin: why `<Space>` after `ackHead` unlike `stream` above?

packages/exo-stream/DESIGN.md
```diff
@@ -0,0 +1,134 @@
...
+## Terminology
+
+- **Initiator**: The side that starts streaming (creates synchronization chain, iterates remotely)
+- **Responder**: The side that provides values (wraps a local async iterator)
+- **Synchronization chain**: Promise chain from initiator to responder.
+- **Acknowledgement chain**: Promise chain from responder to initiator.
+
+## Protocol Flow
+
+```mermaid
```
jcorbin, line 96: what is a mermaid, is that supposed to get rendered somehow? how can I see it during review?

packages/exo-stream/PROTOCOL.md
```diff
@@ -0,0 +1,300 @@
...
+A stream capability exposes the following interface:
+
+```typescript
+interface PassableStream {
+  stream(synchronizes: ERef<StreamNode>): Promise<StreamNode>;
+  yieldPattern(): Pattern | undefined;
+  returnPattern(): Pattern | undefined;
+}
+
+// Where ERef<T> = T | PromiseLike<T>
```
jcorbin: Where Pattern = ???

packages/exo-stream/PROTOCOL.md
```diff
@@ -0,0 +1,300 @@
...
+    // Iteration loop
+    while true:
+        let node = await nodePromise
+        let value = node.value
+        
+        if node.promise === null:
+            // Stream ended - this is the return value
+            return value
+        
+        yield value
```
jcorbin: just say `node.value` and drop the temporary ; not only is it shorter pseudo code, but it's more type stable if ever used as an implementation template:
- above `value : Y|R` until the promise-null-check collapses it to `Y`
- and most progress type chasers are so conservative, as to not narrow down the aliased type of value then in this yield site

packages/exo-stream/PROTOCOL.md
```diff
@@ -0,0 +1,300 @@
...
+        while true:
+            let synNode = await synPromise
+            
+            if synNode.promise === null:
+                // Initiator signaled close
+                ackResolve({value: undefined, promise: null})
+                return
+            
+            // Produce next value, passing synchronize value to iterator
+            let result = await iterator.next(synNode.value)
```
jcorbin: let alone symmetry with above, see here you correctly access a type stable alias for value, rather than an unstable local binding...

packages/exo-stream/PROTOCOL.md
```diff
@@ -0,0 +1,300 @@
...
+### Normal Completion
+
+When the iterator is exhausted:
+
+1. Responder receives `{done: true, value}` from `iterator.next()`
+2. Responder resolves ack node with `{value, promise: null}`
+3. Initiator receives node with `promise === null`
+4. The `value` is the iterator's return value
+
+### Early Close (Initiator-Initiated)
```
jcorbin: seems like we're missing a 4th case here for Responder-Initiated early close?

packages/exo-stream/stream-iterator.js
```diff
@@ -0,0 +1,235 @@
...
+/**
+ * A type that can be iterated asynchronously.
+ * Mirrors Stream<TRead, TWrite, TReadReturn, TWriteReturn> template parameters.
+ * The `@endo/stream` Stream type is also accepted since it extends AsyncIterator.
+ *
+ * @template TRead
+ * @template [TWrite=undefined]
+ * @template [TReadReturn=unknown]
+ * @template [TWriteReturn=unknown]
+ * @typedef {AsyncIterable<TRead, TReadReturn, TWrite> | Iterable<TRead, TReadReturn, TWrite> | AsyncIterator<TRead, TReadReturn, TWrite> | Iterator<TRead, TReadReturn, TWrite>} SomehowAsyncIterable
```
jcorbin: minor nit: I find large unions like this much more readable if we line wrap and dangle the `|` ; is that counter code style here? still getting my bearings... trying to recall if you need an extra `( ... )` wrapper inside the `{ ... }` to get away with that...

packages/exo-stream/stream-bytes-iterator.js
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * Returns the iterator for the given iterable object.
+ * Supports both synchronous and asynchronous iterables.
+ *
+ * @template T The item type of the iterable.
+ * @param {SomehowAsyncIterable<T>} iterable The iterable object.
+ * @returns {AsyncIterator<T> | Iterator<T>}
+ */
+const asyncIterate = iterable => {
```
jcorbin: why is this utility in the bytes module, rather than alongside where its main argument type is declared in `stream-iterator.js`?

packages/exo-stream/stream-bytes-iterator.js
```diff
@@ -0,0 +1,154 @@
...
+          const ackNode = freeze({
+            value: result.value,
+            promise: result.done ? null : promise,
+          });
+          ackResolve(ackNode);
+          ackResolve = resolve;
+
+          if (result.done) {
+            break;
+          }
```
jcorbin: consider fully disentangling the branches here like:
```javascript

          if (result.done) {
            ackResolve(freeze({
              value: result.value,
              promise: null,
            }));
            break;
          }

          const { promise, resolve } = makePromiseKit();
          ackResolve(freeze({
            value: result.value,
            promise: promise,
          }));
          ackResolve = resolve;
```

reintroduce `ackNode` tmp to taste

in addition to type collapsing the ackNode value literals in both worlds, it'll also avoid making a dangling promise when we break

altho there's a chance then that this will allow the .catch branch below to double-resolve the ack promise? does that get silently ignored, or is problem?

packages/exo-stream/stream-iterator.js
```diff
@@ -0,0 +1,235 @@
...
+          const ackNode = freeze({
+            value: result.value,
+            promise: result.done ? null : promise,
+          });
+          ackResolve(ackNode);
+          ackResolve = resolve;
+
+          if (result.done) {
+            break;
+          }
```
jcorbin: similar thought about disentangling the ack side dualism as above

jcorbin (commented):
idk how all this works yet, but here's some demi-outsider thoughts trying to cog it

packages/exo-stream/DESIGN.md
kriskowal: I pushed a change that clarifies initiator and responder. I picked these words because streams are symmetric, and can be used both for reader and writer streams, depending on whether data flows from initiator to responder (writer) or from responder to initiator (reader). This mirrors the duality of the generator protocol.

packages/exo-stream/PROTOCOL.md
kriskowal: Added a reference to `@endo/pattern`.

packages/exo-stream/stream-bytes-iterator.js
kriskowal: Factored out.

packages/exo-stream/DESIGN.md
grypez: On github you can use the rich diff toggle at the top of the file to see the rendered markdown.

Screenshot 2026-01-12 at 7 28 54 AM

packages/exo-stream/stream-bytes-iterator.js
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * The reader uses bidirectional promise chains for flow control:
+ * - Initiator sends synchronizations via the synchronization chain to induce production
+ * - Responder sends acknowledgements (base64 strings) via the acknowledgement chain
+ *
+ * @param {SomehowAsyncIterable<Uint8Array>} bytesIterator
+ * @param {StreamBytesIteratorOptions} [options]
+ * @returns {PassableBytesReader}
+ */
+export const streamBytesIterator = (bytesIterator, options = {}) => {
```
grypez: In response to your comment about a possible `E.reader(ref)`: perhaps this should be a method of `Far`, i.e. `Far.writer(somehowAsyncIterable)` to obviate that the result is passable. Or similarly `makeExoStreamBytesIterator` would surface invocations in global searches for `makeExo`.

packages/exo-stream/MIGRATION.md
```diff
@@ -0,0 +1,132 @@
...
+- ✅ `src/commands/follow.js` - Uses `iterateStream`
+- ✅ `src/commands/inbox.js` - Uses `iterateStream`
+- ✅ `src/commands/install.js` - Uses `streamBytesIterator`
+- ✅ `src/commands/list.js` - Uses `iterateStream`
+- ✅ `src/commands/make.js` - Uses `streamBytesIterator`
+- ✅ `src/commands/store.js` - Uses `streamBytesIterator`
+- ✅ `demo/cat.js` - Uses `iterateStream`
+
+### Tests
+- ✅ `daemon/test/endo.test.js` - Uses `iterateStream` and `iterateBytesStream`
```
grypez: This information, though nicely summarized here for review, appears in the commit history.

```suggestion
```

Perhaps the entire file can be removed.

Aside: Do github PRs have "files for review but not commit"?

grypez (commented):
Ergonomically as good as we could hope for without enshrining into `E`. We're still early.

packages/exo-stream/MIGRATION.md
kriskowal: The PR description is the only place in a PR for content that’s not committed. In this case, I’m leaving this file because there are branches in flight that have not been migrated and completing those migrations will benefit from this file being in place. There will be a time it’s not useful, soon.

packages/exo-stream/PROTOCOL.md
erights: What does "progress type chasers" mean?

CONTRIBUTING.md
```diff
@@ -76,11 +76,63 @@ guides in the `docs/` directory, and package-specific documentation.
 should use long lines (paragraphs joined without manual wrapping), as GitHub
 uses a different Markdown flavor for those contexts.
 
+### Mermaid Diagrams
+
+Use 2-space indentation for Mermaid diagrams:
```
erights: Could we show the example below within the "```mermaid" .. "```" brackets that would be used? I don't know how to escape it so it works within a "```" block.

Separately, below the quoted example, could you also use the same block unquoted below it to show what that mermaid diagram looks like?

CONTRIBUTING.md
```diff
@@ -76,11 +76,63 @@ guides in the `docs/` directory, and package-specific documentation.
...
 ## Rebuilding `ses`
 
 Changes to `ses` require a `yarn build` to be reflected in any dependency where `import 'ses';` appears. Use `yarn build` under `packages/ses` to refresh the build.
 Everything else is wired up thanks to workspaces, so no need to run installs in other packages.
 
+# Code Style
+
+## JSDoc Type Imports
+
+Prefer `@import` JSDoc tags over inline dynamic imports in type declarations:
```
erights, line 119:
```suggestion
Prefer `@import` JSDoc tags over inline imports in type declarations:
```
"dynamic" confuses me.

packages/exo-stream/CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
+See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
```
erights: Is this still true in the era of changesets?

packages/exo-stream/DESIGN.md
```diff
@@ -0,0 +1,246 @@
+# Design Document for `@endo/exo-stream`
+
+## Protocol Overview
+
+The exo-stream protocol provides non-lossy streams of passable data, with control flow,
+such that the producer does not overwhelm the consumer.
+
+The exo-stream protocol uses bidirectional promise chains for streaming over CapTP.
```
erights, line 8: Currently, when should we say "CapTP" vs "ocapn"?

packages/exo-stream/DESIGN.md
```diff
@@ -0,0 +1,246 @@
...
+By default, the protocol is fully synchronized (buffer=0) and just as chatty as naive
+protocols, suitable for deliberate synchronization. With a buffer value in excess of 0,
```
erights: I think I can guess what you mean by "chatty" here, but I'm unsure.

packages/exo-stream/DESIGN.md
```diff
@@ -0,0 +1,246 @@
...
+
+- **Initiator**: The side that starts streaming (creates synchronization chain,
+  iterates remotely)
+- **Responder**: The side that provides values (wraps a local async iterator),
+- **Synchronization chain**: Promise chain from initiator to responder,
+- **Acknowledgement chain**: Promise chain from responder to initiator.
+
+Streams come in the **Reader** and **Writer** flavors that vary only
+in usage, because the protocol is symmetric.
+
```
erights, line 34: Great place for a pair of diagrams showing each of these scenarios.

packages/exo-stream/DESIGN.md
```diff
@@ -0,0 +1,246 @@
...
+  initiator calls `return(value)` to close early, the final syn node carries that
+  argument value. If the responder is backed by a JavaScript iterator with a
+  `return(value)` method, it forwards the argument and uses the iterator’s
+  returned value as the terminal ack; otherwise it terminates with the original
+  argument value.
+- We leave a void in the terminology for configurations where neither or both
+  parties send data.
+  **Duplex** passable streams are best modeled with a pair of unentangled
+  reader and writer, even if they share a duplex connection for purposes of
+  transport.
```
erights, line 57: Good!

erights: Just curious: when might someone want the "neither" case?

packages/exo-stream/DESIGN.md
erights: Not a comment on the PR:

How/why did that screenshot disappear after it was uploaded to github?

erights: Nevermind. It reappeared.

packages/exo-stream/DESIGN.md
erights: OMG I see the extensive diagram below. Maybe my additional diagram suggestions here are not needed.

erights:
> OP above:
> 
> > based on existing support in pass-style for immutable ArrayBuffer that simply hasn’t been fully realized
> 
> Please cite #2871 ?

Hypothetically, if #2871 were merged first, could we get rid of these separate "Bytes" streams support?

packages/exo-stream/DESIGN.md
```diff
@@ -0,0 +1,246 @@
...
+
+3. **Transmission over CapTP**:
+   - Synchronizes flow: initiator → responder (base64 strings via synchronize promise chain)
+   - Acknowledges flow: responder → initiator (flow control via acknowledge promise chain)
+   - CapTP transmits resolved nodes opportunistically
+
+## Why E.get() Pipelining Works
+
+When an Exo method returns a Promise, `E()` normally "flattens" it—waiting for
+the Promise to resolve.
```
erights: No it doesn't. If it did, we wouldn't get promise pipelining. `E` fully supports promise pipelining of remotables and sends to remotables. Your statement here is ***only*** about promises for composite copy-data.

packages/exo-stream/DESIGN.md
erights: Is a "promise chain" different from a "promise stream" or a "stream of promises"? If not, what is a chain? Do we need this additional concept?

packages/exo-stream/DESIGN.md
```diff
@@ -0,0 +1,246 @@
...
+```javascript
+// We pipeline through the promise:
+let nodePromise = E(streamRef).stream(synHead);  // No await yet
+// ...later in next():
+const value = await E.get(node).value;           // Pipeline property access
+nodePromise = node.promise;                      // Get next node
+```
+
+This allows the initiator to immediately start resolving synchronization nodes
+without waiting for the acknowledge chain to resolve first, avoiding datalock.
```
erights, line 232: Likewise, can we globally replace "chain" with "stream"?

erights: Oh! Is a stream always a pair of chains? That makes sense. If so, nevermind.

erights: Resolving.

packages/exo-stream/DESIGN.md
erights: Oh! Is a stream always a pair of chains? That makes sense. If so, nevermind. Resolving.

packages/exo-stream/NEWS.md
```diff
@@ -0,0 +1 @@
+User-visible changes in `@endo/exo-stream`:
```
erights, line 1: NEWS.md in the changeset era?

packages/exo-stream/README.md
```diff
@@ -0,0 +1,349 @@
...
+`iterateBytesReader()`, and `iterateWriter()` instead of
+`iterateBytesWriter()`.
+
+## Design
+
+See [DESIGN.md](./DESIGN.md) for design documentation.
+
+## Protocol Specification
+
+See [PROTOCOL.md](./PROTOCOL.md) for the formal protocol specification,
```
erights: Does `PROTOCOL.md` already exist? Any idea why I'm not finding it?

packages/exo-stream/DESIGN.md
erights: Resolving

CONTRIBUTING.md
gibson042: > Could we show the example below within the "`mermaid" .. "`" brackets that would be used? I don't know how to escape it so it works within a "```" block.

Inline code spans support increasing the enclosing backticks and optional inner whitespace, so e.g. <tt>\`\`\`\` mermaid\`\`\` \`\`\`\`</tt> renders as ```` mermaid``` ````.

A similar trick applies to code blocks, e.g.
```````
`````
```js
import '@endo/init';
```
`````
```````
renders as
`````
```js
import '@endo/init';
```
`````

packages/exo-stream/CHANGELOG.md
kriskowal: Rebasing now that master has changesets. NEWS.md entries move to a changeset. This gets rescaffolded.

packages/exo-stream/DESIGN.md
kriskowal: I’m inclined to say CapTP even if it is OCapN’s CapTP since this is on the passable abstraction and not a specific protocol. But, (e.g., OCapN) is probably useful someday.

packages/exo-stream/DESIGN.md
kriskowal: I’ll try another phrasing. I received “chatty” from the folks I worked with at FastSoft and assumed it was broadly known to mean the opposite of pipelining, where the sender and receiver wait for the other to finish “speaking” before sending another message.

packages/exo-stream/DESIGN.md
kriskowal: I can’t imagine.

kriskowal:
> > OP above:
> > > based on existing support in pass-style for immutable ArrayBuffer that simply hasn’t been fully realized
> > 
> > 
> > Please cite #2871 ?
> 
> Hypothetically, if #2871 were merged first, could we get rid of these separate "Bytes" streams support?

Yes, I believe so.

packages/exo-stream/README.md
kriskowal: I deleted it because it was largely redundant with DESIGN.md. I’m sweeping this reference up in the next push.

packages/exo-stream/DESIGN.md
kriskowal: Rephrasing.

PR #3037, opened 2026-01-04
feat(ses): Increase defense against RegExp.prototype poisoning

gibson042:
## Description

I noticed in passing use of RegExp.prototype methods that contain further lookups, and also a set-like array where an actual set would be better.

### Security Considerations

The issue doesn't seem to be exploitable, but better safe than sorry.

### Scaling Considerations

The third commit provides a minor speedup to compartment initialization.

n/a

### Testing Considerations

Existing test coverage continues to suffice.

None known.

### Upgrade Considerations

None known.

### DIFF
--- a/packages/ses/src/commons.js
+++ b/packages/ses/src/commons.js
@@ -188,11 +188,6 @@ if (!('hasOwn' in Object)) {
 }
 
 export const { hasOwn } = Object;
-
-/**
- * @deprecated Use `hasOwn` instead
- */
-export const objectHasOwnProperty = hasOwn;
 //
 export const arrayFilter = uncurryThis(arrayPrototype.filter);
 export const arrayForEach = uncurryThis(arrayPrototype.forEach);
@@ -234,10 +229,9 @@ export const setHas = uncurryThis(setPrototype.has);
 export const iterateSet = uncurryThis(setPrototype[iteratorSymbol]);
 //
 /**
- * @deprecated `regexpTest` is vulnerable to RegExp.prototype poisioning; use
- * `regexpExec` instead.
+ * `regexpExec` is provided in exclusion of `regexpTest`, which would be
+ * vulnerable to RegExp.prototype poisoning.
  */
-export const regexpTest = uncurryThis(regexpPrototype.test);
 export const regexpExec = uncurryThis(regexpPrototype.exec);
 /**
  * @type { &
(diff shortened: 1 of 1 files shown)


packages/ses/src/scope-constants.js
```diff
@@ -4,15 +4,18 @@ import {
...
+  setHas,
 } from './commons.js';
 
 /**
- * keywords
- * In JavaScript you cannot use these reserved words as variables.
- * See 11.6.1 Identifier Names
+ * reservedNames
+ * In JavaScript you cannot use reserved words as variable names (except for
+ * "eval", which is speically reserved to prevent shadowing).
```
kriskowal:
```suggestion
 * "eval", which is specially reserved to prevent shadowing).
```

packages/ses/src/transforms.js
```diff
@@ -170,7 +172,7 @@ export const rejectImportExpressions = src => {
  */
 export const evadeImportExpressionTest = src => {
   const replaceFn = (_, p1, p2) => `${p1}__import__${p2}`;
-  return stringReplace(src, importPattern, replaceFn);
+  return regexpReplace(importPattern, src, replaceFn);
```
kriskowal, line 175: Under maintenance, may be useful to clarify why we prefer this method. I think it’s just that the protocol is simpler. The pattern is known to be a regex, so this skips a step to sense whether it’s a string.

packages/ses/src/scope-constants.js
```diff
@@ -89,24 +95,12 @@ const identifierPattern = /^[a-zA-Z_$][\w$]*$/;
...
-export const isValidIdentifierName = name => {
-  // Ensure we have a valid identifier. We use regexpTest rather than
-  // /../.test() to guard against the case where RegExp has been poisoned.
-  return (
-    name !== 'eval' &&
-    !arrayIncludes(keywords, name) &&
-    regexpTest(identifierPattern, name)
-  );
-};
+export const isValidIdentifierName = name =>
```
kriskowal, line 102: Might be good to preserve the reasoning for avoiding regexpTest and polymorphic test.

packages/ses/src/commons.js
```diff
@@ -244,6 +272,8 @@ export const generatorNext = uncurryThis(generatorPrototype.next);
 export const generatorThrow = uncurryThis(generatorPrototype.throw);
 
 /**
+ * @deprecated `stringReplace` is vulnerable to RegExp.prototype poisoning; use
+ * `regexpReplace` instead (and `sealRegexp` on its regular expressions).
```
kriskowal: I see. Perhaps this is sufficient to capture the reasoning. Since this is internal, is it an option to just delete it?

gibson042: In principle yes, but in practice we still have `objectHasOwnProperty` as well. Maybe such deprecated exports should be removed in the next significant release?

kriskowal: We can safely delete anything unused in `commons.js` because they are not part of the public API. Deprecating `objectHasOwnProperty` rather than deleting it was, I believe, an oversight.

PR #3039, opened 2026-01-05
docs: Documentation for message passing

kriskowal:
This change introduces a graduated tutorial to message passing and fleshes out the READMEs for the packages that pertain to message passing including pass-style, patterns, and exos. These will all be visible on endojs.org. The tutorial will be prominent.

My preferred method of reviewing a chunk of documentation of this magnitude is to hand this off to the next editor and ask them to append their edits and create a new PR, to supersede this one. It will be impractical to do a round-trip for all the suggested changes. I would like to pass this potato to @erights first.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,32 @@ jobs:
       - name: build API docs
         run: yarn docs
 
+  docs-markdown:
+    # begin macro
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      # without this, setup-node errors on mismatched yarn versions
+      - run: corepack enable
+
+      - name: Use Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: 22.x
+          cache: yarn
+
+      - name: Install dependencies
+        run: yarn install --immutable
+
+      # end macro
+
+      - name: Run yarn build
+        run: yarn build
+
       # build the API docs in markdown for agoric/documentation repo to verify it works
       - name: build API docs in markdown
         run: yarn docs:markdown-for-agoric-documentation-repo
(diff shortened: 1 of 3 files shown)


docs/message-passing.md
```diff
@@ -0,0 +1,1340 @@
...
+in exchange, obviates any possiblity of deadlock, even in composition of many,
+independently designed compoennts, utterly unaware of each other's
+synchronization models.
+This is the linchpin of fearless cooperation among components in a
+loosely-coordinated, distributed, multi-agent system, which is to say, any
+system at sufficient scale.
+
+### The Pass Styles Taxonomy
+
+Every passable value falls into one of eight categories:
```
erights, line 69: Starting from here, the github markdown preview shows lots of inappropriate underbars. However, from looking at the source here, I have no idea why. If you know how to fix this, please do. It is very annoying.

erights: Actually, it starts much earlier. But I still cannot figure out why.

Image

erights: Nevermind. It is an artifact of the rendered diff view. Just viewing the file without the diff view makes it go away. Resolving.

CONTRIBUTING.md
```diff
@@ -18,6 +18,31 @@ then update the resulting README.md, package.json (specifically setting
...
+Each package has a specific role in enabling safe message passing.
+Together, they form the foundation of distributed computing.
+```
+
+This convention applies to all documentation files including README.md files,
+guides in the `docs/` directory, and package-specific documentation.
+
+**Exception:** Release notes in pull request descriptions and GitHub releases
+should use long lines (paragraphs joined without manual wrapping), as GitHub
+uses a different Markdown flavor for those contexts.
```
erights, line 44: Huh. Different how? Is there a name for that markdown flavor?

kriskowal: This is the sole difference that I’m aware of and the flavor doesn’t appear to have a name: merely “Github pull request and release notes flavor Markdown”.

PR #3042, opened 2026-01-07
docs(memoize): Fill in README

kriskowal:


packages/memoize/README.md
```diff
@@ -1 +1,76 @@
...
+
+- **Caching**: Results are cached per argument identity using a `WeakMap`.
+  Subsequent calls with the same argument return the cached result.
+- **Throws are not memoized**: If the wrapped function throws, the exception
+  propagates and no result is cached.
+  The next call with the same argument will invoke the function again.
+- **Rejected promises are memoized**: If the wrapped function returns a
+  rejected promise, that promise is cached like any other return value.
+- **Invalid keys throw**: If an argument is not a valid `WeakMap` key
+  (e.g., a primitive string or number), the memoized function throws after
```
erights, line 53: "after" is weird, so I reread the code to see if it is correct, and whether it would be worth it to change to "before"

- it is correct
- IMO it is worth changing, but not in this PR of course. Perhaps note in the README that we may change the "after".

erights: Still in draft, but will be https://github.com/endojs/endo/pull/3056

kriskowal: Added a note.

PR #3044, opened 2026-01-08
docs(lp32): Elaborate README

kriskowal:
This change documents the lp32 package. It also addresses some minor inconsistencies in the implementation. The byte order for native messages is not configurable.

### DIFF
--- a/packages/lp32/README.md
+++ b/packages/lp32/README.md
@@ -1,10 +1,136 @@
 # `@endo/lp32`
 
-Also known as the web extension "Native Host Message" protocol, this package
-implements async iterator streams for reading and writing with 32-bit
-host-byte-order length-prefix message envelopes for binary data, represented
-with Uint8Arrays.
+Length-prefixed message streams using 32-bit host byte order framing,
+implemented as async iterators.
 
-These streams are "hardened" and depend on Hardened JavaScript.
-Most JavaScript environments can be locked down with the
-[SES shim](../ses/README.md).
+## Overview
+
+This package implements the binary message framing protocol used by
+[WebExtension Native Messaging][native-messaging].
+Each message is prefixed with a 32-bit unsigned integer indicating the message
+length in bytes, using host byte order.
+
+The protocol is simple:
+- A 4-byte length prefix (uint32, host byte order)
+- Followed by the message payload of that length
+
+For example, a 5-byte message `hello` is transmitted as:
+```
+[0x05, 0x00, 0x00, 0x00] [h, e, l, l, o]
+```
+
+This package provides hardened async iterator streams for reading and writing
+these length-prefixed messages, represented as `Uint8Array`s.
+
+[native-messaging]: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Native_messaging
+
+## Usage
+
+### Reading Messages
+
+Use `makeLp32Reader` to wrap an async iterable of byte chunks and produce
+an async iterable of complete messages:
+
+```js
+import { makeLp32Reader } from '@endo/lp32';
+
+const reader = makeLp32Reader(byteStream, {
+  name: '<my-stream>',        // optional, for error messages
+  maxMessageLength: 1024 * 1024, // optional, defaults to 1MB
+});
+
+for await (const message of reader) {
+  // message is a Uint8Array containing one complete message
+  console.log(new TextDecoder().decode(message));
+}
+```
+
+### Writing Messages
+
+Use `makeLp32Writer` to wrap an output stream and automatically frame
+messages with length prefixes:
+
+```js
+import { makeLp32Writer } from '@endo/lp32';
+
+const writer = makeLp32Writer(outputStream, {
+  name: '<my-writer>',
+  maxMessageLength: 1024 * 1024,
+});
+
+const encoder = new TextEncoder();
+await writer.next(encoder.encode('hello'));
+await writer.next(encoder.encode('world'));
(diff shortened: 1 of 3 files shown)


packages/lp32/README.md
```diff
@@ -1,10 +1,136 @@
...
+Each message is prefixed with a 32-bit unsigned integer indicating the message
+length in bytes, using host byte order.
+
+The protocol is simple:
+- A 4-byte length prefix (uint32, host byte order)
+- Followed by the message payload of that length
+
+For example, a 5-byte message `hello` is transmitted as:
+```
+[0x05, 0x00, 0x00, 0x00] [h, e, l, l, o]
```
erights, line 19: If this example is specific to a particular host endianness, then you should give an example of each right here and say which is which. For most programmer, including me, our ability to think about endianness has gone stale.

erights: BTW, I find it incredibly bizarre that a communications format would use host-dependent endianness. How would communication between hosts of different endiannesses work?

(At least we can be relieved that the world no longer needs to worry about VAX endianness any longer.)

packages/lp32/reader.js
```diff
@@ -44,7 +42,7 @@ async function* makeLp32Iterator(
 
     let drained = false;
     while (!drained && length >= 4) {
-      const messageLength = data.getUint32(0, littleEndian);
+      const messageLength = data.getUint32(0, hostIsLittleEndian);
```
erights, line 45: right here. What happens if the length we read was written on a host with a different endianness?

packages/lp32/README.md
```diff
@@ -1,10 +1,136 @@
...
+```js
+import { makeLp32Writer } from '@endo/lp32';
+
+const writer = makeLp32Writer(outputStream, {
+  name: '<my-writer>',
+  maxMessageLength: 1024 * 1024,
+});
+
+const encoder = new TextEncoder();
+await writer.next(encoder.encode('hello'));
```
erights, line 62: Good that we no longer need to seed the generator push-side protocol with a bogus first call to `next()`. 

Noting that @ljharb is about to withdraw `function.sent`, pending approval from @hax, so we could never repair this anyway if `writer` were a generator`.

packages/lp32/README.md
```diff
@@ -1,10 +1,136 @@
...
+
+Use `makeLp32Reader` to wrap an async iterable of byte chunks and produce
+an async iterable of complete messages:
+
+```js
+import { makeLp32Reader } from '@endo/lp32';
+
+const reader = makeLp32Reader(byteStream, {
+  name: '<my-stream>',        // optional, for error messages
+  maxMessageLength: 1024 * 1024, // optional, defaults to 1MB
```
erights, line 39: What happens if the max is exceeded?

packages/lp32/README.md
```diff
@@ -1,10 +1,136 @@
...
+const writer = makeLp32Writer(output);
+const reader = makeLp32Reader(input);
+
+const encoder = new TextEncoder();
+const decoder = new TextDecoder();
+
+// Producer
+await writer.next(encoder.encode('message 1'));
+await writer.next(encoder.encode('message 2'));
+await writer.return();
```
erights, line 83: Are these `await`s for flow control? I notice there is an `initialCapacity` and a `maxMessageLength` but no `maxCapacity`. So what does the flow control wait for?

erights (approved):
LGTM, thanks!

packages/lp32/README.md
kriskowal: I believe the designers of this protocol intended it to be specifically used only for same-host communication between a web browser and a child process, so discounted the upside of network order (cross-host portability) in favor of the upside of host order (no translation). For such a length prefix, I would probably have gone another way. In any case, we have this in anticipation of needing it to connect a web extension to the daemon, which we’ve prototyped but have no concrete plan for.

packages/lp32/reader.js
kriskowal: It would fail in a strange way. Probably would block for a very long time, waiting for sufficient data to fill a 4MB buffer, then produce a protocol error. Likely never.

As such, this is not a suitable protocol for cross-host comms and only intended for browser to app comms.

packages/lp32/README.md
kriskowal: Indeed, using the iterator protocol directly, we don’t suffer the parity mismatch of generators. This comes up in the draft exo-stream change as well.

packages/lp32/README.md
kriskowal: The awaits are for flow control. The reader must call `next` for the producer to advance. There’s a facility in `@endo/stream` for pre-acking so that these pipeline within pre-arranged bounds, but the default is that they’re synchronized.

packages/lp32/README.md
erights: Good. Resolving.

packages/lp32/reader.js
erights: got it. Resolving.

packages/lp32/README.md
erights: By "but the default is that they’re synchronized." do you mean there is effectively a zero-sized buffer, i.e., that this is a rendezvous protocol by default? This question should not delay merging, especially since this PR only documents the status quo. But I would like to discuss sometime. Low priority compared to everything else going on right now though.

kriskowal: Yes, the default is effectively a zero-sized buffer.

PR #3045, opened 2026-01-08
docs(ocapn): Flesh out README

kriskowal:
As advertised.

PR #3047, opened 2026-01-09
docs: Populate READMEs

kriskowal:


packages/cli/README.md
```diff
@@ -1,5 +1,187 @@
-# Endo CLI
+# `@endo/cli`
```
gibson042, line 1: If we're finally committing to this convention, this PR should update [skel](https://github.com/endojs/endo/tree/master/packages/skel) and/or [create-package.sh](https://github.com/endojs/endo/blob/master/scripts/create-package.sh) accordingly.

packages/cli/README.md
```diff
@@ -1,5 +1,187 @@
...
+endo where state   # State directory
+endo where sock    # Unix socket / named pipe
+endo where log     # Log file
+endo where cache   # Cache directory
+endo where run     # PID file directory
+```
+
+## License
+
+[Apache-2.0](./LICENSE)
```
gibson042, line 193: Likewise this.

packages/cli/README.md
```diff
@@ -1,5 +1,187 @@
...
+
+```sh
+endo ls|list                    # List known names
+endo ls|list some-directory     # List names in a directory
+endo ls|list path.to.directory  # List names in a deeply nested directory
+endo show my-value              # Print a value
+endo cat my-blob                # Dump blob contents
+endo rm|remove old-name         # Forget a name
+endo mv|move old-name new-name  # Rename
+endo cp|copy src-name dst-name  # Duplicate a name
```
jcorbin: I had to do a double take to realize that the `|` in `ls|list` was meant as a BNF-style lexical alternate, since the `sh` lang context otherwise primed me to think you were trying to pipe-into some `list` command

packages/cli/README.md
```diff
@@ -1,5 +1,187 @@
...
+```
+
+Check inbox and handle messages:
+
+```sh
+endo inbox                  # List received messages
+endo inbox --follow         # Follow messages as they arrive
+endo adopt 1 value-name     # Adopt a value from message #1
+endo dismiss 1              # Delete message #1
+```
```
jcorbin, line 150: is this is supposed to be the "alice" host inbox from above? how do we know?

packages/cli/README.md
```diff
@@ -1,5 +1,187 @@
...
+
+Request and grant capabilities:
+
+```sh
+endo request "I need access to the database" --to HOST
+endo resolve 1 database-ref  # Grant request #1 with a named value
+endo reject 1 "Not authorized"
+```
```
jcorbin: These are interleaving alice/bob viewpoints implicitly?

packages/netstring/README.md
```diff
@@ -1,19 +1,150 @@
...
+
+## Hardened JavaScript
+
+This package depends on Hardened JavaScript.
+The environment must be locked down before use, typically via `@endo/init`.
+
+## References
+
+[Netstring][]
+D. J. Bernstein
```
jcorbin, line 143: <3 djb ... it's been far too long since I was in the sendmail vs qmail trenches ... world could really do with some more [djbtools](https://cr.yp.to/daemontools.html) in current year

packages/cli/README.md
kriskowal: Updated skel README to use `[package]` placeholders that create-package.sh expects.

packages/cli/README.md
kriskowal: Good point! Changed to parenthetical notation: `endo list  # (or: endo ls)`

packages/cli/README.md
kriskowal: Fixed\! Added explicit `--as bob` to clarify whose inbox we're viewing, with a header note "(bob's perspective)".

packages/cli/README.md
kriskowal: Fixed! Added comments clarifying "Bob requests..." and "Host reviews inbox..." to make the perspective switches explicit.

erights (commented):
The changes to the lp32 package look very similar to those in #3044 . I already reviewed that one. Does that review cover this version of those three files?

erights (commented):
same q again for #3045 which I already reviewed.

packages/stream-node/README.md
```diff
@@ -1,3 +1,91 @@
...
+# `@endo/stream-node`
 
-This package provides `makeNodeReader` and `makeNodeWriter` adapters that adapt Node.js Reader and Writer to Endo's async iterable streams.
+Adapters for Node.js streams to Endo's async iterable streams.
+
+## Overview
+
+This package provides functions to adapt Node.js `Readable` and `Writable`
+streams to Endo's stream interface, which models streams as hardened async
+iterators of `Uint8Array` chunks.
```
erights: Since writer users pass an argument to `next`, I wondered whether it is still ok to call these "async iterators". Indeed it is. The TS def seem to be

```ts
interface AsyncIterator<T, TReturn = any, TNext = any> {
    // NOTE: 'next' is defined using a tuple to ensure we report the correct assignability errors in all places.
    next(...[value]: [] | [TNext]): Promise<IteratorResult<T, TReturn>>;
    return?(value?: TReturn | PromiseLike<TReturn>): Promise<IteratorResult<T, TReturn>>;
    throw?(e?: any): Promise<IteratorResult<T, TReturn>>;
}
```
and
```ts

/**
 * Describes a user-defined {@link AsyncIterator} that is also async iterable.
 */
interface AsyncIterableIterator<T, TReturn = any, TNext = any> extends AsyncIterator<T, TReturn, TNext> {
    [Symbol.asyncIterator](): AsyncIterableIterator<T, TReturn, TNext>;
}
```

Given those, do we really need our own
```ts
// Stream is nearly identical to AsyncGenerator and AsyncGenerator should
// probably be identical to this definition of Stream.
// Stream does not make the mistake of conflating the read and write return
// types.
export interface Stream<
  TRead,
  TWrite = undefined,
  TReadReturn = undefined,
  TWriteReturn = undefined,
> {
  next(value: TWrite): Promise<IteratorResult<TRead, TReadReturn>>;
  return(value: TWriteReturn): Promise<IteratorResult<TRead, TReadReturn>>;
  throw(error: Error): Promise<IteratorResult<TRead, TReadReturn>>;
  [Symbol.asyncIterator](): Stream<TRead, TWrite, TReadReturn, TWriteReturn>;
}
```
?

Ah. The comment explains:

```
// Stream does not make the mistake of conflating the read and write return
// types.
```

That just seems to be a but in the TS type. Should we report it? In any case, I see the problem. No change suggested.

Curious though if the comment should say `AsyncIterableIterator` rather than `AsyncGenerator`.

packages/stream-node/README.md
```diff
@@ -1,3 +1,91 @@
...
+A complete example that pipes stdin to stdout:
+
+```js
+import '@endo/init';
+import { pump } from '@endo/stream';
+import { makeNodeReader, makeNodeWriter } from '@endo/stream-node';
+
+const reader = makeNodeReader(process.stdin);
+const writer = makeNodeWriter(process.stdout);
+await pump(writer, reader);
```
erights, line 54: The example is only self contained if you also understand `pump`. Brief explanation here?

Btw, since we're only using iterators and not generators, `prime` is now obsolete. Can we kill it? (though not in this PR of course)

packages/stream-node/README.md
```diff
@@ -1,3 +1,91 @@
...
+
+## API
+
+### `makeNodeReader(input)`
+
+Adapts a Node.js `Readable` stream to an Endo reader.
+
+- Converts Node.js `Buffer` chunks to `Uint8Array`
+- The stream must not be in object mode
+- The stream must not have an encoding set
```
erights: I'm scared to ask what these are ;) . In any case, this part of the doc is for people who presumably know something about node streams, so no change suggested.

packages/stream-node/README.md
```diff
@@ -1,3 +1,91 @@
...
+**Parameters:**
+- `input` - A Node.js `Readable` stream
+
+**Returns:** A `Reader<Uint8Array>` async iterator
+
+### `makeNodeWriter(output)`
+
+Adapts a Node.js `Writable` stream to an Endo writer.
+
+- Respects back-pressure via the `drain` event
```
erights: Cool! I look forward to understanding this.

Perhaps for all of these, how about we link to the relevant Node Stream docs at the top of this README.md ?

erights (commented):
Just comments so far. More coming.

packages/netstring/README.md
```diff
@@ -1,19 +1,150 @@
...
+### Reading netstrings
+
+`makeNetstringReader` decodes a stream of bytes into individual messages:
+
+```js
+import { makeNetstringReader } from '@endo/netstring';
+
+const reader = makeNetstringReader(byteStream, {
+  name: '<my-stream>',       // optional, for error messages
+  maxMessageLength: 1000000, // optional, defaults to 999999999
```
erights: In the comment, please put either commas of underbars to separate each triple of digits. I'd also appreciate such underbars in the code for long numbers.

erights: Curious why the max here is `99...` but the max for another one (stream node?) was `100...`

packages/netstring/README.md
```diff
@@ -1,19 +1,150 @@
...
+
+const encoder = new TextEncoder();
+await writer.next(encoder.encode('hello'));
+await writer.next(encoder.encode('world'));
+await writer.return();
+```
+
+### Writing chunked messages
+
+Messages can be passed as arrays of chunks to avoid pre-concatenation:
```
erights, line 54: Given that the writer itself does concat these, do you really want to call them "chunks" given your other use of "chunks" in streams?

packages/netstring/README.md
```diff
@@ -1,19 +1,150 @@
-# netstring
+# `@endo/netstring`
 
-This is an implementation of asynchronous streams framed as [Netstrings][].
-A netstring is a binary protocol for length-prefixed frames,
-using decimal strings as variable width integers.
-For example, the frame `5:hello,` corresponds to the message `hello`,
-where `5` is the length of `hello` in bytes.
+Async iterator streams for [Netstring][] protocol.
```
erights, line 3: Somewhere you should point out that several of the `make*Reader` and `make*Writer` pairs follow the same conventions. User will make their own pairs and you should encourage them to follow the pattern.

packages/cli/README.md
```diff
@@ -1,5 +1,187 @@
...
+Command line interface for the Endo daemon.
+
+## Overview
+
+The `endo` command provides a user interface for managing the Endo daemon,
+a persistent host for guest programs running in Hardened JavaScript.
+Through the CLI, users can run confined programs, manage named values, send
+messages, and control the daemon lifecycle.
+
+See [`@endo/daemon`](../daemon/README.md) for more about the underlying daemon.
```
erights, line 12: Do you expect to similarly flesh out daemon/README.md ? Without knowing it yet, my only suggestion is to turn some mentions into links.

packages/cli/README.md
```diff
@@ -1,5 +1,187 @@
...
+yarn install
+cd packages/cli
+yarn start        # Start the daemon
+yarn endo --help  # Run CLI commands
+```
+
+Or invoke the CLI directly:
+
+```sh
+node packages/cli/bin/endo --help
```
erights, line 31: What is the scope of the daemon? On a multi-account system (i.e., allegedly "multi-user", like POSIX), if each account follows these instructions, does each get their own daemon? If not, Bob have the ability to stop the daemon Alice started?

packages/cli/README.md
erights: I still see `|` notation. Did you forget to push?

packages/cli/README.md
```diff
@@ -1,5 +1,187 @@
...
+```
+
+## Running Programs
+
+Run a confined program:
+
+```sh
+endo run program.js
+endo run program.js --powers NONE    # No special powers (default)
+endo run program.js --powers AGENT   # All of the primary user's agency
```
erights: What is an "agent"? What is an "agency"? Do you intend "agency" to simply be the plural of "agent"? If so, I find that weird. Google says the plural is "agents", as I expected.

erights: What is the "primary user"? This is the only mention in this doc.

erights: Is this an ACL check? Is the right to act as primary user sharable? transferable? deniable to programs that run as me?

kriskowal: This is “agency” in the second sense,
> the capacity, condition, or state of acting or of exerting power
> — Merriam Webster

The `AGENT` powers are effectively limited only by the host operating system and the user that started the daemon, since they include creation of unconfined programs.

The name AGENT is left over from the last round of development and clearly suffers from semantic diffusion. We’ll need to change it. The pool of alternatives is mucky. My best idea today is PROFILE because there are to be other profiles like profile-for-that-app, my-alternate-profile, profile-for-friend, &c. So, SELF would be your handle and PROFILE would be the combination of powers and mailbox that the user’s alter-egos, apps, peers, and LLM agents hold.

(Hah! Or, we could use ID for your handle and EGO for your powers. They have the virtue of brevity, and the pun.)

For the purposes of this documentation, I’ll clarify that AGENT designates all the powers available to the user that started the Dæmon.

packages/stream-node/README.md
kriskowal: Node.js conflates all sorts of streams. The same implementation does ringbuffers, transcoders, and object streams. It goes to tortuous lengths to use the same implementation for all these cases. Without an encoding, the unit of transmission is a chunk of bytes. With an encoding, the unit of transmission is a string, and it introduces a transcoder.

packages/netstring/README.md
kriskowal: In this case, the max is dictated by the maximum number of ASCII decimal digits that it will read before committing to read the body.

packages/netstring/README.md
kriskowal: I think these usages of “chunk” are spiritually the same, undifferentiated ranges of bytes that would be equivalent if they traded some bytes with their neighbors. The important distinction is “message”, which cannot. The boundary between messages is significant. In this case, Mathieu had added an optimization that allows a message to be communicated as an array of chunks.

packages/cli/README.md
kriskowal: The daemon currently only supports one user, who can have multiple personas (“hosts”), and multiple “guests”. The single user corresponds to the user who runs the daemon. On POSIX, this is ideally a dedicated user for the daemon, but there are other options. The daemon is going to eventually need to listen on a socket for CapTP (with encryption) and ideally on a single socket. We could either go with a relay that dispatches to local, user-specific daemons, or a single daemon owned by a “super user” where all POSIX users are guests. Or, just lean into the emerging convention that POSIX systems are not multi-user in practice and frequently run in single-user containers.

packages/cli/README.md
```diff
@@ -1,5 +1,193 @@
...
+
+This package is not yet published to npm.
+To use it locally, clone the Endo repository and run from the workspace:
+
+```sh
+git clone https://github.com/endojs/endo.git
+cd endo
+yarn install
+cd packages/cli
+yarn start        # Start the daemon
```
erights, line 24: Apparently should be

```suggestion
yarn endo start   # Start the daemon
```

packages/cli/README.md
```diff
@@ -1,5 +1,193 @@
...
+```sh
+node packages/cli/bin/endo --help
+```
+
+## Daemon Lifecycle
+
+Start, stop, and manage the Endo daemon:
+
+```sh
+endo start    # Start the daemon
```
erights, line 39: Not a comment on this PR. Rather, I'm using the opportunity to review this file by trying everything out.

If I've already started the daemon with `endo --start` and I `endo --start` again I get the error

```
endo start
(Error#1)
Error#1: listen EADDRINUSE: address already in use /Users/markmiller/Library/Application Support/Endo/captp0.sock
  at ChildProcess.<anonymous> (packages/daemon/index.js:132:18)
  at ChildProcess.emit (node:events:518:28)
```

I understand why this is correct, but it is not of much help. Can the `endo --start` understand that the thing at that socket is what it would have put at that socket? If it is, would it make sense for a redundant `endo --start` to give a warning but otherwise do nothing with a zero exit code so scripts continue? Or is this a case where we want scripts to halt with an error?

If we do error, I'd at least like to see the address it collided on so I can start my bug hunt.

packages/cli/README.md
```diff
@@ -1,5 +1,193 @@
...
+```sh
+git clone https://github.com/endojs/endo.git
+cd endo
+yarn install
+cd packages/cli
+yarn start        # Start the daemon
+yarn endo --help  # Run CLI commands
+```
+
+Or invoke the CLI directly:
```
erights, line 28: I made my own shell shortcut using my old bad bash habits
```bash
endo() {
    "${HOME}/src/ongithub/endojs/endo/packages/cli/bin/endo" "$@"
}
```

Since the examples below do (as they should) presume some such shorthand, how would you suggest people set this up? Does it make sense to suggest that here?

packages/cli/README.md
erights: Weirder, `endo restart` while the daemon is running displays an error but exits with a zero exit code, with the daemon still running

Image

erights: I've got `endo log -f` going in another window, which is great. I also appreciate `endo ping`

packages/cli/README.md
```diff
@@ -1,5 +1,193 @@
...
+node packages/cli/bin/endo --help
+```
+
+## Daemon Lifecycle
+
+Start, stop, and manage the Endo daemon:
+
+```sh
+endo start    # Start the daemon
+endo stop     # Stop the daemon
```
erights, line 40: Image

packages/cli/README.md
```diff
@@ -1,5 +1,193 @@
...
+endo restart  # Restart the daemon
```
erights, line 41: Both `endo start` and `endo restart` only work when there's no daemon running. They both cause the daemon to run. What's the difference between the two?

packages/cli/README.md
```diff
@@ -1,5 +1,193 @@
...
+endo purge  # Erase all persistent state (requires confirmation)
+```
+
+## Running Programs
+
+Run a confined program:
+
+```sh
+endo run program.js
+endo run program.js --powers NONE   # No special powers (default)
```
erights, line 59:
```suggestion
endo run program.js                               # No special powers
endo run program.js --powers NONE   # No special powers (default)
```
I realize this is implied by the "(default)" on the next line, but on this it is worth being explicit.

packages/cli/README.md
```diff
@@ -1,5 +1,193 @@
...
+endo run program.js --powers AGENT  # All of the primary user's agency
+endo run program.js --powers ENDO   # All of the power of the daemon
```
erights, line 61: Interesting

packages/cli/README.md
```diff
@@ -1,5 +1,193 @@
...
+endo clean  # Erase ephemeral state (logs, sockets)
+endo purge  # Erase all persistent state (requires confirmation)
+```
+
+## Running Programs
+
+Run a confined program:
+
+```sh
+endo run program.js
```
erights, line 58: When `endo run` runs a js program that immediately terminates with an error, the `endo run` still terminates with a zero exit code. This is probably the best decision so no change suggested. I'm just noting since at first it surprised me.

Image

erights: My little test file is
```js
/* global globalThis */
console.log(Reflect.ownKeys(globalThis).length);
```

It seems to have run fine, gives me a normal looking output, and exists with a zero error code. Vut it also reports a `namespace.main is not a function` error, which my test code does not mention. Is this a known problem?

Image

erights: I notice that `endo run` after en `endo stop` will start the daemon. I'm confident that's good and appreciate the console notice, so no change suggested. But noting that it surprised me.

Image

packages/netstring/reader.js
```diff
@@ -14,7 +14,7 @@ const NINE = '9'.charCodeAt(0);
  */
 async function* makeNetstringIterator(
   input,
-  { name = '<unknown>', maxMessageLength = 999999999 } = {},
```
erights, line 17: Thanks

packages/cli/README.md
erights: More mysteries. If I follow `--powers` with nonsensical `foo` it acts like the others. 

Image

But if I say `Foo` it more errors first. Here, I only see the errors from the sender's (`endo run`) pov. Made we wonder whether the errors from the receiver's pov (daemon) is logged. Then I remembered I have an `endo log -f` running, and it does complete the picture. Yay!

Image

packages/cli/README.md
erights: So NONE, AGENT, ENDO are each supposed to provide different powers. But they all seem to have 72 globals in the `globalThis` seen by the program. The globals beyond what `Compartment` provides by default are
```
Base64
E
Far
M
TextDecoder
TextEncoder
URL
assert
atob
btoa
console
makeExo
```
all of which seem powerless enough and fine. (Btw, `Base64` is not in https://github.com/endojs/endo/blob/b024b06c7b801e8683a8930715558cfbedfd653d/packages/cli/src/commands/run.js#L12 . Is it provided by the unbundling? No objection. Just curious.)

Where are the differential powers found? Differences in import namespace?

packages/cli/README.md
```diff
@@ -1,5 +1,193 @@
...
+endo run program.js --powers NONE   # No special powers (default)
+endo run program.js --powers AGENT  # All of the primary user's agency
+endo run program.js --powers ENDO   # All of the power of the daemon
+```
+
+Make a persistent worker wherein programs can run:
```
erights, line 64: What does "persistent worker" mean?

packages/cli/README.md
erights: Another thing I find surprising is that the `--powers <something>` option appears after the filename even though it is an option to the `endo run` command, not an argument given to the program being run. Indeed, `endo help` summarizes the `endo run` command line as
```
  run [options] [<file>] [<args>...]                         runs a program (runlet)
```
which makes sense to me, but contradicts this placement of `--powers`

erights: I like the *let terms including runlet. Since we do use these terms in `endo help`, should we also use them in this README?

packages/cli/README.md
```diff
@@ -1,5 +1,193 @@
...
+
+```sh
+endo bundle ./app --name my-bundle
+endo run --bundle my-bundle
+```
+
+## Managing Names
+
+Endo uses petnames to refer to values.
+Each agent has its own namespace of names.
```
erights, line 81: This is the second mention of "agent". The previous mention was

> Endo run program.js --powers AGENT  # All of the primary user's agency

Is this the agent you mean?

packages/cli/README.md
```diff
@@ -1,5 +1,193 @@
...
+```
+
+## Running Programs
+
+Run a confined program:
+
+```sh
+endo run program.js
+endo run program.js --powers NONE   # No special powers (default)
+endo run program.js --powers AGENT  # All of the primary user's agency
```
erights, line 60: When is a user not a primary user?

packages/cli/README.md
```diff
@@ -1,5 +1,193 @@
...
+  --worker my-worker \
+  --name doubled
+```
+
+## Agents and Messages
+
+Endo supports multiple agents (personas, profiles) that can exchange messages
+and capabilities.
+These personas can be held by people, confined programs, bots, or AI agents.
+People can have more than one.
```
erights, line 127: Nevermind the question about "agent" above. This establishes that it covers a lot more than the primary user.

packages/cli/README.md
```diff
@@ -1,5 +1,193 @@
...
+
+Create agents:
+
+```sh
+endo mkhost alice  # Create a host agent (unlimited local authority)
```
erights, line 132: Is 

```suggestion
endo mkhost alice  # Create a host agent (All of the primary user's authority)
```

correct? It borrow language from the earlier comment about `--powers AGENT`

packages/cli/README.md
erights: Below you use "authority". Can we just use it here and avoid seeming to introduce a new concept?

```suggestion
endo run program.js --powers AGENT  # All of the primary user's authority
```

packages/cli/README.md
```diff
@@ -1,5 +1,193 @@
...
+endo mkdir my-directory
+```
+
+## Evaluating Expressions
+
+Evaluate JavaScript in a worker:
+
+```sh
+endo eval '1 + 1' --name two
+endo eval 'a + b' a b --name sum  # With named values as arguments
```
erights, line 116: Further indulging the pedantic nitpicker side of my brain, I tried these just as written, and was surprised to see `NaN` rather than an error. So I tried `endo eval 'a'` to diagnose. It seems evaluating an unknown petname produces `undefined` rather than throwing. Is this what we want?


Image

erights: Trying out instead an expanded form of this example that a reader might think is correct at this point:

Image

packages/cli/README.md
```diff
@@ -1,5 +1,193 @@
...
+
+## Managing Names
+
+Endo uses petnames to refer to values.
+Each agent has its own namespace of names.
+A path of names can traverse into nested namespaces.
+
+```sh
+endo list                    # List known names (or: endo ls)
+endo list some-directory     # List names in a directory
```
erights, line 86: At this point, the term "directory" will suggest to many people a file system directory. Above you say "nested namespaces". Can we just say "namespace" everywhere we currently say "directory"?

packages/cli/README.md
```diff
@@ -1,5 +1,193 @@
...
+```sh
+endo mkhost alice  # Create a host agent (unlimited local authority)
+endo mkguest bob   # Create a guest agent (limited authority)
+```
+
+Send messages with embedded references (alice the host sends to bob the guest):
+
+```sh
+endo send bob "Here is @my-value for you" --as alice
+endo send bob "Take @this-thing:your-name" --as alice  # bob sees it as "your-name"
```
erights, line 140: Is the embedded petname syntax here supposed to converge with the Familiar?

packages/cli/README.md
```diff
@@ -1,5 +1,193 @@
...
+```
+
+Check inbox and handle messages (bob the guest receives from alice):
+
+```sh
+endo inbox --as bob               # List bob's received messages
+endo inbox --follow --as bob      # Follow messages as they arrive
+endo adopt 1 value-name --as bob  # Adopt a value from message #1
```
erights, line 148: Which value from message #1 ?

Btw, I had a good talk with @warner that has me questioning whether we should send message-scoped names at all, since they are a new concept, vs an edge name rooted in the sender, which is also a new concept at this point, but will be a necessary concept anyway?

packages/cli/README.md
erights: > with a header note "(bob's perspective)".

I do not see this note. Forgot to push?

packages/cli/README.md
```diff
@@ -1,5 +1,193 @@
...
+
+Request and grant capabilities:
+
+```sh
+# bob (guest) requests a capability from alice (host)
+endo request "I need access to the database" --as bob --to alice
+
+# alice (host) reviews inbox and grants or rejects
+endo inbox --as alice                      # See bob's request
+endo resolve 1 database-ref --as alice     # Grant request #1 with a named value
```
erights, line 160: In message passing terms, does this mean "return this value in response to this request" ?

Are we trying to obscure the message passing perspective on this interaction?

packages/cli/README.md
```diff
@@ -1,5 +1,193 @@
...
+endo reject 1 "Not authorized" --as alice  # Or reject it
+```
+
+Act as a different agent:
+
+```sh
+endo list --as alice
+endo list --as bob
+```
```
erights, line 169: Should this block be moved above the examples that use `--as`?

packages/cli/README.md
```diff
@@ -1,5 +1,193 @@
-# Endo CLI
+# `@endo/cli`
 
-The Endo command line is a user interface for managing the Endo application
-runner (daemon).
-This includes managing the lifecycle of the daemon process.
+Command line interface for the Endo daemon.
+
+## Overview
```
erights, line 5: I notice another useful overview file at demo/README.md, which I will read next. But I discovered it by accident. Should this file link to that one?

erights (approved):
I got very picky and pedantic with my comments on cli/README.md. Some are just recording my thoughts as I went through trying things while reading the doc. Some are questions I would like to understand. I don't consider any of these things crucial to fix before merging. So LGTM and use your judgement.

Finally, I am more and more curious about the relationship between this cli "language" and the "language" of the Familiar. As I was going through the Familiar earlier, I imagined how each element would desugar to JS. Is it more relevant to imagine how it would desugar to this cli "language"? Should they be equally expressive and directly map to one another?

packages/cli/README.md
erights: Indeed, the last section https://github.com/endojs/endo/tree/kriskowal-docs-readmes/packages/cli/demo#running-a-confined-script clears up several of the confusions I had trying to understand this one.

erights: Yeah, I wish I had read that one first. Could these usefully be merged into one document?

packages/cli/README.md
kriskowal: This is correct.

packages/cli/README.md
kriskowal: Oh, I had never seen this behavior, because this emerges from the new SES change where we denied guest code ability to sense properties of the true global in exchange for `undefined` instead of `ReferenceError`.

packages/cli/README.md
kriskowal: This is a tricky naming problem. Currently, name hub, directory, and pet store have a lot of common methods in their interfaces and I don’t consider them to have settled into stone.

* **name hub** is intended to be a shape that anything can implement to take advantage of certain Familiar verbs.
* **guest** and **host** are both **agent** and implement both **name hub** *and* **mailbox**
* **directory** is an implementation of a **name hub** that doesn’t do anything except hold names it it. These can be constructed with `mkdir` at the CLI and `makeDirectory` from JS.
* **pet store** is the storage mechanism underlying these and various other KV retention componnents.
* **pet sitter** is a layer kumavis factored out for a **pet store** with some additional indelible keys.

Names are subject to discussion we get more experience.

* **name hub** https://github.com/endojs/endo/blob/master/packages/daemon/src/types.d.ts#L527-L546
* **directory** https://github.com/endojs/endo/blob/master/packages/daemon/src/types.d.ts#L548-L550
* **pet store** https://github.com/endojs/endo/blob/master/packages/daemon/src/types.d.ts#L484-L509

packages/cli/README.md
kriskowal: The intention is for the user experience to converge. That is, the same keystrokes should have the same effect in the Familiar. However, the Familiar is at liberty to solve the Hotel problems in different ways. Namely, it does not simply parse this text, but typing `@` puts you in a pet-name completion context. Pressing `:` advances you to the link-name context. Pressing SP or TAB in either of those contexts should complete the token/chip/card. Pressing Esc should leave an `@` literal.

And, this is subject to user feedback. The intention is for it to be as fluent as chat and there are other idioms we can bring to bear. For example, we could instead ask the user to select text and ⌘K to convert the text to a link and  prompt the user to designate a capability.

packages/cli/README.md
kriskowal: Edge names and name hubs are a useful concept, but put the user in a position where they are not just interacting with the recipient but to all their name hub’s subscribers. Messages are more intimate.

packages/cli/README.md
kriskowal: This does indeed mean as you say. I am adding a `reply` verb for messages in general but it’s not equivalent. Replying to a request does not necessarily resolve the request with any particular of the named powers in the response. Having a separate verb for this kind of message gives the Familiar UI a hint about how to present the request.

I am expecting that we will do a lot of iteration on the discriminated union of message types.

packages/cli/README.md
erights: Good answer at demo/README.md:

`alias endo=$PWD/packages/cli/bin/endo`

erights: Oh. Except that it is unsuitable for placement into a `.bashrc` file

packages/cli/README.md
```diff
@@ -1,5 +1,193 @@
...
+Endo uses petnames to refer to values.
+Each agent has its own namespace of names.
+A path of names can traverse into nested namespaces.
+
+```sh
+endo list                    # List known names (or: endo ls)
+endo list some-directory     # List names in a directory
+endo list path.to.directory  # List names in a deeply nested directory
+endo show my-value           # Print a value
+endo cat my-blob             # Dump blob contents
```
erights, line 89: What is a blob? No mention elsewhere in this file or in demo/README.md

kriskowal: The usual meaning: Binary Large OBject.

When the daemon makes a formula from some text, it captures it in an immutable content-address-store for future reference. This is where the code for an `evaluate` formula goes or the bundle for an `importBundle` goes.

erights: Given that we're visibly in a JS-based system, will people assume we're talking about web Blobs?

Issue #2897, opened 2025-07-16
Compartment map support for subpath wildcard pattern replacement in exports (and imports)


kriskowal:
## What is the Problem Being Solved?

The compartment mapper currently supports `package.json` with `"exports"` but does not support wildcard pattern replacement, as in `"./x/*/y/*/z": "./src/x/*/y/*z.js"`.

## Description of the Design

Supporting this feature requires us to communicate patterns through the compartment-map JSON format, so we’ll need to introduce a `"patterns"` array of `from` `to` duples or pairwise arrays. This will allow the runtime to create a prefix tree that generally translates a module specifier.

There is a possibility that this might subsume the `"scopes"` feature.

## Security Considerations

None.

## Scaling Considerations

None.

## Test Plan

We will need fixtures that include patterns with at least two wildcards. We will need unit tests that cover the cases where the number of wildcards do not match between the key and value.

## Compatibility Considerations

This will increase our ecosystem compatibility, particularly enabling us to avoid barrel modules in Protobuf generated packages like `@agoric/cosmic-proto` more fluently.

We must make sure that the `patterns` get erased from a `compartment-map.json` in a bundle, because the current Agoric chain runtime will refuse any bundle that contains unrecognized properties. This will require all *used* patterns to be replaced with concrete module aliases.

## Upgrade Considerations

None.

PR #3048, opened 2026-01-09
feat(compartment-mapper): Import and export subpath patterns

kriskowal:
Closes: #2897

## Description

Adds support for Node.js subpath pattern replacement in `package.json` `exports` and `imports` fields to the compartment-mapper. This enables patterns like:

```json
{
  "exports": {
    "./features/*.js": "./src/features/*.js"
  },
  "imports": {
    "#internal/*.js": "./lib/*.js"
  }
}
```

The implementation matches Node.js semantics where `*` is a string replacement token that matches any substring, including across `/` path separators. Patterns are matched by specificity: the pattern with the longest matching prefix wins, and exact entries always take precedence over wildcard patterns.

### Security Considerations

Pattern-matched module imports go through the same policy enforcement (`enforcePolicyByModule`) as other module resolutions. Cross-compartment patterns (from dependency exports) resolve within the dependency's compartment, maintaining compartment isolation. The prefix/suffix string matching approach avoids regex-based matching, eliminating potential ReDoS concerns.

### Scaling Considerations

Pattern matching is O(p) where p is the number of wildcard patterns, sorted by prefix length for specificity. Exact entries are checked first via a `Map` lookup. Results are cached via write-back to `moduleDescriptors`, so each specifier is matched at most once.

Users can now use wildcard patterns in their `package.json` `exports` and `imports` fields when using compartment-mapper. The patterns follow Node.js subpath pattern semantics:
- `*` matches any substring, including across `/` separators
- Exact entries take precedence over wildcard patterns
- Longer prefix wins when multiple patterns match
- Null targets (`null` values) exclude subpaths from resolution
- Conditional pattern values are resolved through the same condition-matching rules as non-pattern exports
- Globstar (`**`) patterns are silently ignored (matching Node.js behavior)
- Array fallback values are deliberately not supported (see design doc for rationale)
- Pattern and replacement must have matching wildcard counts

### Testing Considerations

- 13 unit tests cover the pattern matcher: exact match, wildcard matching, cross-`/` matching, specificity ordering, `#`-imports patterns, null-target exclusion, error cases (globstar rejection, wildcard count mismatch), and various input formats (tuples, `PatternDescriptor` array, record object)
- Scaffold integration tests verify the full pipeline through `loadLocation`, `importLocation`, `mapNodeModules`, `makeArchive`, `parseArchive`, `writeArchive`, `loadArchive`, `importArchive`
- Node.js parity tests run the same fixtures under plain Node.js, ensuring behavioral equivalence by construction. Assertions are shared via `_subpath-patterns-assertions.js`.
- User-specified conditions tested via ses-ava with `--conditions=blue-moon` to verify the correct branch is selected by both Node.js and the Compartment Mapper
- Policy enforcement tested: pattern-matched imports allowed when package is permitted, rejected when not
- Archive stripping tested: inspects `compartment-map.json` in archive and asserts no compartment has a `patterns` property
- Tests verify that patterns work correctly through archiving (patterns are expanded to concrete module entries)

This is a purely additive feature. Existing packages without wildcard patterns are unaffected. The `patterns` field is only added to compartment descriptors when patterns exist (using conditional spread), so existing snapshot tests pass without modification.

### Upgrade Considerations

No breaking changes. Packages can start using wildcard patterns in `exports`/`imports` immediately after upgrading to a version with this feature.

### DIFF
--- a/packages/compartment-mapper/test/subpath-patterns-node-parity.test.js
+++ b/packages/compartment-mapper/test/subpath-patterns-node-parity.test.js
@@ -107,8 +107,9 @@ test('exports edge cases - node parity', async t => {
 });
 
 test('non-object exports field is rejected by Node.js', async t => {
-  // Node.js ignores the invalid numeric exports field and falls back to
-  // looking for index.js, which does not exist.
+  // Node.js rejects the invalid numeric exports field. The error code varies
+  // by version: ERR_PACKAGE_PATH_NOT_EXPORTED on 18/20, ERR_MODULE_NOT_FOUND
+  // on 22+. We just verify it throws.
   await t.throwsAsync(
     () =>
       import(
@@ -117,9 +118,6 @@ test('non-object exports field is rejected by Node.js', async t => {
           import.meta.url,
         ).href
       ),
-    {
-      code: 'ERR_MODULE_NOT_FOUND',
-    },
   );
 });
 


packages/compartment-mapper/src/types/node-modules.ts
```diff
@@ -89,6 +90,11 @@ export interface PackageDescriptor {
    * TODO: Update with proper type when this field is handled.
    */
   exports?: unknown;
+  /**
+   * Package imports field for self-referencing subpath patterns.
+   * Keys must start with '#'.
+   */
+  imports?: unknown;
```
boneskull: (suggestion)

Steal this code for the type defs of `exports` and `imports`: https://github.com/sindresorhus/type-fest/blob/850b33c4dd292e0ff8cff039ee167d69be324fce/source/package-json.d.ts#L227-L248

packages/compartment-mapper/src/types/pattern-replacement.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+export type SubpathReplacer = (specifier: string) => string | null;
+
+/**
+ * Input format for pattern mappings - either an array of tuples
+ * or a record object.
+ */
+export type SubpathMapping =
+  | Array<[pattern: string, replacement: string]>
+  | Record<string, string>;
```
boneskull, line 31: 👍 for named tuple elements

packages/compartment-mapper/src/infer-exports.js
```diff
@@ -105,6 +107,46 @@ function* interpretExports(name, exports, conditions) {
   }
 }
 
+/**
+ * Interprets the `imports` field from a package.json file.
+ * The imports field provides self-referencing subpath patterns that
+ * can be used to create private internal mappings.
+ *
+ * @param {string} _name - the name of the package (unused, but kept for consistency)
```
boneskull: Consistency with what?

packages/compartment-mapper/src/infer-exports.js
```diff
@@ -105,6 +107,46 @@ function* interpretExports(name, exports, conditions) {
...
+ */
+function* interpretImports(_name, imports, conditions) {
+  if (Object(imports) !== imports) {
+    throw Error(
+      `Cannot interpret package.json imports property, must be object, got ${imports}`,
+    );
+  }
+  for (const [key, value] of entries(imports)) {
+    // imports keys must start with '#'
+    if (!key.startsWith('#')) {
```
boneskull, line 140: (suggestion) warn

packages/compartment-mapper/src/infer-exports.js
```diff
@@ -218,3 +260,86 @@ export const inferExportsAndAliases = (
...
 };
+
+/**
+ * Determines if a key or value contains a wildcard pattern.
+ *
+ * @param {string} key
+ * @param {string} value
+ * @returns {boolean}
+ */
+const hasWildcard = (key, value) => key.includes('*') || value.includes('*');
```
boneskull: Isn't it also important to know which of these two values contains the wildcard(s)?

packages/compartment-mapper/src/link.js
```diff
@@ -113,6 +114,19 @@ const makeModuleMapHook = (
   moduleDescriptors,
   scopeDescriptors,
 ) => {
+  // Build prefix-tree-based pattern matcher once per compartment if patterns exist.
+  // @ts-expect-error patterns may exist on PackageCompartmentDescriptor
```
boneskull: (should) add the field to `PackageCompartmentDescriptor`

packages/compartment-mapper/src/link.js
```diff
@@ -113,6 +114,19 @@ const makeModuleMapHook = (
...
+  const { patterns } = compartmentDescriptor;
+  const matchPattern =
+    patterns && patterns.length > 0
+      ? makeMultiSubpathReplacer(
+          patterns.map((/** @type {{ from: string, to: string }} */ p) => [
```
boneskull: ...which will also eliminate the need for this type assertion

packages/compartment-mapper/src/link.js
```diff
@@ -162,6 +176,40 @@ const makeModuleMapHook = (
...
 
+    // Check patterns for wildcard matches (before scopes).
+    // Patterns resolve within the same compartment only.
+    if (matchPattern) {
+      const resolvedPath = matchPattern(moduleSpecifier);
+      if (resolvedPath !== null) {
+        // Policy enforcement for pattern-matched modules
+        enforcePolicyByModule(moduleSpecifier, compartmentDescriptor, {
+          exit: false,
+          errorHint: `Pattern matched: ${q(moduleSpecifier)} -> ${q(resolvedPath)}`,
```
boneskull: (suggestion) make a bit more verbose

packages/compartment-mapper/src/pattern-replacement.js
```diff
@@ -0,0 +1,314 @@
...
+const WILDCARD = '*';
+
+/**
+ * Path separator; Win32-style paths unsupported
+ */
+const PATH_SEP = '/';
+
+/**
+ * Greedy magic globstar; unsupported
+ */
```
boneskull: I can't remember if Node actually supports this or not.  If it does, what are the technical barriers to implementation (if any)?

packages/compartment-mapper/src/pattern-replacement.js
```diff
@@ -0,0 +1,314 @@
...
+const GLOBSTAR = '**';
+
+/**
+ * Checks if a segment contains a wildcard.
+ *
+ * @param {string} segment
+ * @returns {boolean}
+ */
+const hasWildcard = segment => segment.includes(WILDCARD);
```
boneskull: (could) Refactor and share with `infer-exports.js`

packages/compartment-mapper/src/pattern-replacement.js
```diff
@@ -0,0 +1,314 @@
...
+ * @param {string} specifierSegment - Actual segment to match
+ * @returns {string | null} The captured wildcard value, or null if no match
+ */
+const matchSegment = (patternSegment, specifierSegment) => {
+  if (!hasWildcard(patternSegment)) {
+    // Exact match required
+    return patternSegment === specifierSegment ? '' : null;
+  }
+
+  const wildcardIndex = patternSegment.indexOf(WILDCARD);
```
boneskull: (could) if we're able to assert here, I'd probably want to assert that `wildcardIndex` is a non-negative integer

packages/compartment-mapper/src/pattern-replacement.js
```diff
@@ -0,0 +1,314 @@
...
+  const prefixTree = new PathPrefixTree();
+
+  const mappingEntries = isArray(mapping) ? mapping : entries(mapping);
+  for (const [pattern, replacement] of mappingEntries) {
+    if (pattern.includes(GLOBSTAR)) {
+      throw new TypeError(
+        `Globstar (**) patterns are not supported in pattern: "${pattern}"`,
+      );
+    }
+    if (replacement.includes(GLOBSTAR)) {
```
boneskull: (could) `hasGlobstar()`

packages/compartment-mapper/test/pattern-replacement.test.js
```diff
@@ -0,0 +1,211 @@
...
+  makeMultiSubpathReplacer,
+  assertMatchingWildcardCount,
+} from '../src/pattern-replacement.js';
+
+// PathPrefixTreeNode tests
+
+test('PathPrefixTreeNode - creates empty node', t => {
+  const node = new PathPrefixTreeNode();
+  t.is(node.value, null);
+  t.deepEqual(Object.keys(node.children), []);
```
boneskull: (could) refactor to `t.like(node, {value: null, children: []})` assuming it works the way I think it should

packages/compartment-mapper/test/pattern-replacement.test.js
```diff
@@ -0,0 +1,211 @@
...
+// assertMatchingWildcardCount tests
+
+test('assertMatchingWildcardCount - passes for matching counts', t => {
+  t.notThrows(() => assertMatchingWildcardCount('./foo', './bar'));
+  t.notThrows(() => assertMatchingWildcardCount('./*/foo', './*/bar'));
+  t.notThrows(() => assertMatchingWildcardCount('./*/foo/*', './*/bar/*'));
+});
+
+test('assertMatchingWildcardCount - throws for mismatched counts', t => {
+  const error = t.throws(() => assertMatchingWildcardCount('./*/a/*', './*'));
```
boneskull: (complaining) `t.throws()` is weaksauce.

```ts
expect(
  () => assertMatchingWildcardCount('./*/a/*', './*'),
  'to throw',
  {
    message: expect.it(
      'to match', /wildcard count mismatch/i,
      'and', 'to match', /2/,
      'and', 'to match', /1/
    )
  }
);
```

(imperative) feel the power of BUPKIS

packages/compartment-mapper/src/pattern-replacement.js
kriskowal: I’ve added parity tests that illustrate that the Node.js behavior for `*` is analogous to globstar, and also that Node.js only supports one wildcard per pattern. Now we know.

packages/compartment-mapper/designs/subpath-pattern-replacement.md
```diff
@@ -0,0 +1,271 @@
...
+**Array fallback values.**
+Node.js allows array values in exports as fallback lists, where each
+entry is tried in order and the first file that exists on disk is used.
+Pattern resolution in the compartment-mapper is a pure string operation
+with no filesystem access.
+Array fallbacks would require threading read powers through the pattern
+matcher and changing the `SubpathReplacer` signature.
+Node.js documentation discourages array fallbacks.
+If a pattern value is an array, `interpretExports` yields all elements
+as separate entries and the first match wins without fallback probing.
```
boneskull, line 201: This seems to imply that we _don't_ have 1:1 parity with Node.js, is this correct?

packages/compartment-mapper/src/types/compartment-map-schema.ts
```diff
@@ -105,6 +106,13 @@ export interface PackageCompartmentDescriptor
 
   scopes: Record<string, ScopeDescriptor<PackageCompartmentDescriptorName>>;
 
+  /**
+   * Wildcard patterns for dynamic module resolution within this compartment.
+   * `*` matches any substring including `/` (Node.js semantics).
+   * Stripped during digest/archiving - expanded patterns become concrete module entries.
+   */
+  patterns?: PatternDescriptor[];
```
boneskull:
```suggestion
  patterns?: Array<PatternDescriptor>;
```

I'm pretty sure we're almost exclusively using this form.

packages/compartment-mapper/src/types/node-modules.ts
```diff
@@ -13,9 +13,44 @@ import type {
...
 import type { LiteralUnion } from './typescript.js';
 import { ATTENUATORS_COMPARTMENT } from '../policy-format.js';
 
+/**
+ * A mapping of conditions to their resolved exports.
+ * Each condition key (e.g., "import", "require", "node", "default")
+ * maps to an {@link Exports} value.
+ */
+export type ExportConditions = {
+  // eslint-disable-next-line no-use-before-define
```
boneskull, line 28: We should disable this rule entirely in `**/*.ts` as ESLint doesn't understand type scoping.  TS sources should use [tseslint's implementation](https://typescript-eslint.io/rules/no-use-before-define/) instead.

boneskull: This and the next two types might also want an attribution to `type-fest`, since it appears identical 😄

packages/compartment-mapper/src/types/compartment-map-schema.ts
```diff
@@ -158,6 +166,7 @@ export interface DigestedCompartmentDescriptor
   path: never;
   retained: never;
   scopes: never;
+  patterns: never;
```
boneskull: This appears redundant, since this field does not appear in `CompartmentMapDescriptor`.  Its presence in a `DigestedCompartmentDescriptor` should cause an error, even without the above addition.
```suggestion
```

packages/compartment-mapper/src/types/pattern-replacement.ts
```diff
@@ -0,0 +1,51 @@
...
+  /** Source pattern with wildcard, e.g., "./lib/*.js" */
+  from: string;
+  /** Target pattern with wildcard, e.g., "./*.js". Null means exclusion. */
+  to: string | null;
+  /**
+   * Optional compartment name where the resolved module lives.
+   * When absent, the pattern resolves within the owning compartment.
+   * Set when propagating export patterns from a dependency package.
+   */
+  compartment?: string;
```
boneskull, line 69: I could have sworn we had [exactOptionalPropertyTypes](https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes) enabled.  Maybe not.

packages/compartment-mapper/src/infer-exports.js
```diff
@@ -115,6 +122,54 @@ function* interpretExports(name, exports, conditions, types) {
...
+ * @param {string} _name - the name of the package (unused, kept for signature
+ * parity with `interpretExports` and `interpretBrowserField`)
+ * @param {Imports} imports - the `imports` field from a package.json.
+ * @param {Set<string>} conditions - build conditions about the target environment
+ * @param {LogFn} log
+ * @yields {[string, string | null]}
+ * @returns {Generator<[string, string | null]>}
+ */
+function* interpretImports(_name, imports, conditions, log) {
+  if (Object(imports) !== imports) {
```
boneskull:
```suggestion
  if (Object(imports) !== imports || isArray(imports)) {
```

packages/compartment-mapper/src/infer-exports.js
```diff
@@ -115,6 +122,54 @@ function* interpretExports(name, exports, conditions, types) {
...
+          if (target === null) {
+            yield [key, null];
+          } else if (typeof target === 'string') {
+            yield [key, relativize(target)];
+          }
+          // Take only the first matching condition
+          break;
+        }
+      }
+    }
```
boneskull, line 165: (suggestion) Exhausive `else` clause which logs a warning or throws

packages/compartment-mapper/src/infer-exports.js
```diff
@@ -228,3 +283,154 @@ export const inferExportsAndAliases = (
...
+
+/**
+ * Determines if a key or value contains a wildcard pattern.
+ *
+ * @param {string} key
+ * @param {string | null} value
+ * @returns {boolean}
+ */
+const hasWildcard = (key, value) =>
+  key.includes('*') || (value !== null && value.includes('*'));
```
boneskull:
```suggestion
  key.includes('*') || (value?.includes('*'));
```

Welcome to 2020

packages/compartment-mapper/src/infer-exports.js
```diff
@@ -228,3 +283,154 @@ export const inferExportsAndAliases = (
...
+ * @param {string} key
+ * @param {string} value
+ * @param {LogFn} log
+ * @returns {boolean}
+ */
+const validateWildcardPattern = (key, value, log) => {
+  if (key.includes('**') || value.includes('**')) {
+    log(`Ignoring globstar pattern "${key}": "${value}"`);
+    return false;
+  }
```
boneskull: If this isn't actually valid in Node.js, is it overly-defensive?

packages/compartment-mapper/src/infer-exports.js
```diff
@@ -228,3 +283,154 @@ export const inferExportsAndAliases = (
...
+ *
+ * @param {PackageDescriptor} descriptor
+ * @param {Node['externalAliases']} externalAliases
+ * @param {Node['internalAliases']} internalAliases
+ * @param {PatternDescriptor[]} patterns - array to populate with wildcard patterns
+ * @param {Set<string>} conditions
+ * @param {Record<string, string>} types
+ * @param {LogFn} log
+ */
+export const inferExportsAliasesAndPatterns = (
```
boneskull, line 285: (suggest) split this up a bit. it's pretty gnarly.  

Do we have coverage of all these conditions?

packages/compartment-mapper/src/link.js
```diff
@@ -113,6 +115,16 @@ const makeModuleMapHook = (
   moduleDescriptors,
   scopeDescriptors,
 ) => {
+  // Build pattern matcher once per compartment if patterns exist.
+  const { patterns } = /** @type {Partial<PackageCompartmentDescriptor>} */ (
```
boneskull, line 118: why the type assertion here?

packages/compartment-mapper/src/link.js
```diff
@@ -162,6 +174,57 @@ const makeModuleMapHook = (
...
+          match;
+
+        // Null result means the specifier is explicitly excluded.
+        if (resolvedPath === null) {
+          throw Error(
+            `Cannot find module ${q(moduleSpecifier)} — excluded by null target pattern in ${q(compartmentName)}`,
+          );
+        }
+        const targetCompartmentName =
+          /** @type {typeof compartmentName} */
```
boneskull: Why type assertion?

packages/compartment-mapper/src/link.js
```diff
@@ -162,6 +174,57 @@ const makeModuleMapHook = (
...
+
+        // Write back to moduleDescriptors for caching, archival, and
+        // policy enforcement. The write-back must precede the policy
+        // check because enforcePolicyByModule verifies the specifier
+        // exists in compartmentDescriptor.modules (the same object).
+        moduleDescriptors[moduleSpecifier] = {
+          retained: true,
+          compartment:
+            /** @type {FileUrlString} */
+            (targetCompartmentName),
```
boneskull: There's an `assertFileUrlString()` somewhere that will avoid this sort of assertion

packages/compartment-mapper/src/link.js
```diff
@@ -162,6 +174,57 @@ const makeModuleMapHook = (
...
+          exit: false,
+          errorHint: `Pattern matched: ${q(moduleSpecifier)} -> ${q(resolvedPath)}`,
+        });
+
+        const targetCompartment = compartments[targetCompartmentName];
+        if (targetCompartment === undefined) {
+          throw Error(
+            `Cannot import from missing compartment ${q(targetCompartmentName)}`,
+          );
+        }
```
boneskull, line 215: Is this possible?  I can't recall if a) the compartment map validation even gets run before this, and if so, b) does it cross-reference the `CompartmentModuleConfiguration` objects?

packages/compartment-mapper/src/pattern-replacement.js
```diff
@@ -0,0 +1,186 @@
...
+};
+
+/**
+ * @typedef {object} ResolvedPattern
+ * @property {string} prefix - The part of the pattern before `*`
+ * @property {string} suffix - The part of the pattern after `*`
+ * @property {string | null} replacementPrefix - The part of the replacement before `*`, or null for exclusions
+ * @property {string | null} replacementSuffix - The part of the replacement after `*`, or null for exclusions
+ * @property {string} [compartment] - Optional compartment for cross-compartment patterns
+ */
```
boneskull, line 48: (suggest) move this into a `.ts` file somewhere

packages/compartment-mapper/src/pattern-replacement.js
```diff
@@ -0,0 +1,186 @@
...
+      },
+    );
+  } else {
+    normalizedEntries = entries(mapping).map(
+      /**
+       * @param {[string, string]} entry
+       * @returns {[string, string | null, string | undefined]}
+       */
+      ([pattern, replacement]) => [pattern, replacement, undefined],
+    );
```
boneskull: maybe slightly more concise. nbd

```suggestion
    normalizedEntries = /** @type {typeof normalizedEntries} */(mapping.map(
      entry => {
        if (isArray(entry)) {
          // [pattern, replacement] tuple
          return [entry[0], entry[1], undefined];
        }
        // PatternDescriptor { from, to, compartment? }
        return [entry.from, entry.to, entry.compartment];
      },
    ));
  } else {
    normalizedEntries = /** @type {typeof normalizedEntries} */(entries(mapping).map(
      ([pattern, replacement]) => [pattern, replacement, undefined],
    ));
```

packages/compartment-mapper/src/pattern-replacement.js
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  for (const [pattern, replacement, compartment] of normalizedEntries) {
+    if (pattern.includes(GLOBSTAR)) {
+      throw new TypeError(
```
boneskull: The pedant in me thinks this isn't a `TypeError` because `pattern` is a string. 😄

packages/compartment-mapper/src/pattern-replacement.js
```diff
@@ -0,0 +1,186 @@
...
+        const prefix = pattern.slice(0, wildcardIndex);
+        const suffix = pattern.slice(wildcardIndex + 1);
+        wildcardEntries.push({
+          prefix,
+          suffix,
+          replacementPrefix: null,
+          replacementSuffix: null,
+          compartment,
+        });
+      }
```
boneskull: (could) split out into function, and L138-L150 as well

packages/compartment-mapper/src/pattern-replacement.js
```diff
@@ -0,0 +1,186 @@
...
+        replacementPrefix,
+        replacementSuffix,
+        compartment,
+      });
+    }
+  }
+
+  // Sort wildcard entries by prefix length descending for specificity.
+  // Node.js selects the pattern with the longest matching prefix.
+  wildcardEntries.sort((a, b) => b.prefix.length - a.prefix.length);
```
boneskull: I'm gonna guess `wildcardEntries` doesn't get big enough to start worrying about incurring another loop over it.

packages/compartment-mapper/src/pattern-replacement.js
```diff
@@ -0,0 +1,186 @@
...
+      ) {
+        // Null replacement means this path is explicitly excluded.
+        if (entry.replacementPrefix === null) {
+          return { result: null, compartment: entry.compartment };
+        }
+        const captured = specifier.slice(
+          entry.prefix.length,
+          specifier.length - entry.suffix.length,
+        );
+        const result = `${entry.replacementPrefix}${captured}${entry.replacementSuffix}`;
```
boneskull, line 191: do we have some validation code that ensures there isn't anything fishy in `exports`/`imports` like `/etc/pwd`

packages/compartment-mapper/test/subpath-patterns-node-parity.test.js
```diff
@@ -0,0 +1,93 @@
...
+  const main = await import(
+    new URL('multi-star-lib/src/main.js', fixtureDir).href
+  );
+  t.is(main.main, 'main');
+
+  // The multi-star subpath pattern should NOT resolve.
+  await t.throwsAsync(
+    () => import(new URL('app/multi-star-import.js', fixtureDir).href),
+    {
+      code: 'ERR_PACKAGE_PATH_NOT_EXPORTED',
```
boneskull, line 68: does this actually work?? I didn't think AVA knew about `code`.

CONTRIBUTING.md
```diff
@@ -51,6 +51,12 @@ then update the resulting README.md, package.json (specifically setting
 `description` and [if appropriate] removing `"private": false`), index.js, and
 index.test.js files.
 
+### Coding Style
+
+- Prefer `/** @import */` over dynamic `import()` in JSDoc type annotations.
+  Use a top-level `/** @import {Foo} from 'bar' */` comment instead of inline
+  `{import('bar').Foo}` in `@param`, `@type`, or `@returns` tags.
```
boneskull, line 58: 👍

packages/compartment-mapper/src/types/pattern-replacement.ts
turadg: No but worth it,
- https://github.com/endojs/endo/pull/3165

packages/compartment-mapper/designs/subpath-pattern-replacement.md
kriskowal: That is correct. I did not attempt to do fallbacks.

CONTRIBUTING.md
turadg: Btw Agoric-sdk has a lint rule for this with autofixer.

packages/compartment-mapper/src/infer-exports.js
kriskowal: This was silly. I’ve removed the argument.

packages/compartment-mapper/test/subpath-patterns-node-parity.test.js
kriskowal: Ava doesn’t specifically know about `code`, but can check arbitrary props like this, in addition to `message`.

packages/compartment-mapper/src/pattern-replacement.js
kriskowal: I’m adding a test. The pattern replacements are normalized to have a ./ prefix, so absolute paths are not expressible.

packages/compartment-mapper/src/pattern-replacement.js
kriskowal: This has been removed.

packages/compartment-mapper/src/link.js
kriskowal: Yes, but only with a hand-edited `compartment-map.json` in an archive, which we should account for.

packages/compartment-mapper/src/infer-exports.js
kriskowal: This is overly defensive and I’m removing it and related checks.

PR #3053, opened 2026-01-16
fix(patterns): Account for the compactOrdered "~" prefix in covering-range bounds

### LINKED ISSUE #3046: Rank cover should support iterating over partial tuple prefix

Assume I have a `SetStore<[user: string, nonce: bigint, deadline: bigint]>`, I want to be able to call `store.keys([user, nonce, M.bigint()])` to check if there are any entries in the store for that user+nonce regardless of the deadline.

## Description of the Design

@gibson042 @erights know what to do

## Security Considerations

None

## Scaling Considerations

This is to enable efficient iteration of a known prefix of keys in a collection

## Test Plan

TBD

### LINKED ISSUE #3051: patterns: `kindOf` incorrectly looks at properties inherited from Object.prototype
## Describe the bug

`kindOf` does not adequately guard against `Object.prototype` property reads.

## Steps to reproduce

```js
import '@endo/init';
import { kindOf } from '@endo/patterns';
import { makeTagged } from '@endo/pass-style';

const goodTagged = makeTagged('foo', undefined);
console.log(kindOf(goodTagged));
// => undefined

const badTagged = makeTagged('valueOf', undefined);
console.log(kindOf(badTagged));
// => TypeError: matchHelper.checkIsWellFormed is not a function
```

## Expected behavior

Object.prototype properties are not confused as helpers.

### LINKED ISSUE #3052: patterns: `getRankCover(kindMatcher, …)` incorrectly treats "copyBag" as a pass style
## Describe the bug

The match:kind `getRankCover` has an incorrect list of known tags.

## Steps to reproduce

```js
import '@endo/init';
import { makePassableKit } from '@endo/marshal';
import { M, getRankCover } from '@endo/patterns';

const { encodePassable } = makePassableKit();

const goodMatcher = M.kind('copySet');
console.log(getRankCover(goodMatcher));
// => [ ':', ';' ]

const badMatcher = M.kind('copyBag');
console.log(getRankCover(badMatcher));
// => TypeError: Cannot read properties of undefined (reading 'cover')

const unknownMatcher = M.kind('foo');
console.log(getRankCover(unknownMatcher));
// => TypeError: Cannot read properties of undefined (reading 'cover')
```

## Expected behavior

`getRankCover` treats copyBag matchers the same as copySet matchers. I also think it should emit a range covering all taggeds for an unknown kind matcher, but (for accommodating the possibility of introducing new pass styles in the future) a case could also be made for emitting a range that covers all values, or for preserving the current behavior of throwing an exception. But as noted elsewhere, I'm not comfortable with mixing known pass styles and known tag names into a single "kind" concept at all.

gibson042:
Fixes #3046
Fixes #3051
Fixes #3052

## Description

* Update functions for computing covering bounds to account for the "~" prefix of format "compactOrdered".
* Improve `getRankCover` implementations and output, particularly for non-Key CopyArrays that have an all-Keys prefix.
* Add property-based testing for `getRankCover`.
* Fix bugs uncovered by that testing.

### Security Considerations

None known; the bugs do not seem to be exploitable in the hardened environment relied upon by `@endo/patterns`.

### Scaling Considerations

n/a

Functions that assume the absence of an encoding prefix have been marked as deprecated in both JSDoc comments and in NEWS.md.

### Testing Considerations

New and improved.

The included fix for #3052 results in the range corresponding with an unknown kind matcher covering all values (à la `M.any()`), which is inefficient but supports future addition of new pass styles. It could instead be changed to throw.

### Upgrade Considerations

These improvements should be harmless to include, and AFAICT will have only minor effects on agoric-sdk because that system does not even use compactOrdered, let alone support composite collection keys in durable storage.


.changeset/new-lizards-switch.md
```diff
@@ -0,0 +1,7 @@
+---
+'@endo/marshal': minor
+---
+
+- `FullRankCover` is deprecated. Instead, use `provideStaticRanks(encodePassable)['*'].cover`.
```
kriskowal, line 6: Just aesthetically, I don’t think we need to resort to `'*'` for the wildcard and could simply spell out `any`, on the assumption that pass styles will not grow much and certainly will never need to include `any`.

kriskowal (commented):
We need a top-level document, in `docs`, to be surfaced on endojs.org, that covers the topic of rank cover. I defer to @erights for your final stamp on these changes. I found no defects.

.changeset/new-lizards-switch.md
```diff
@@ -0,0 +1,7 @@
+---
+'@endo/marshal': minor
```
kriskowal, line 2: There’s a commit with the `!` indicator on it, which suggests this should be `major`. If it’s not breaking anything in practice, I do not object to minor, provided we take off the indication on the conventional commit.

packages/pass-style/tools/arb-passable.js
```diff
@@ -42,121 +93,196 @@ export const makeArbitraries = (fc, exclusions = []) => {
...
+  const recoverableMap = (arb, mapper) =>
+    arb.map(
+      recoverableInput => {
+        const result = mapper(recoverableInput);
+        recoverabilityCache.set(result, recoverableInput);
+        return result;
+      },
+      v => {
+        const found = recoverabilityCache.get(/** @type {any} */ (v));
+        if (found !== undefined || recoverabilityCache.has(found)) return found;
```
kriskowal, line 135: Is `undefined` in the range of values that can be cached? Are we checking it up front as an optimization, to avoid a redundant check for presence in the cache?

kriskowal (commented):
Relaying germane feedback from the bot:

* It might be worth noting in the changeset that empty covers are no longer supported, although they were admittedly nonsense before.

Comments below.

.changeset/new-lizards-switch.md
gibson042: I prefer use of a non-identifier key for emphasizing the special nature of this non-pass-style-based cover.

.changeset/new-lizards-switch.md
gibson042: Sure, reduced to minor.

packages/pass-style/tools/arb-passable.js
gibson042: > Is `undefined` in the range of values that can be cached? Are we checking it up front as an optimization, to avoid a redundant check for presence in the cache?

Yes, and yes. Input generation is a hot path and I went out of my way to avoid inefficiency.

erights:
> effectively precluding the addition of any new pass style in the future

@gibson042 , I don't understand. Explain please?

gibson042:
> > effectively precluding the addition of any new pass style in the future
> 
> @gibson042 , I don't understand. Explain please?

This relates to the range cover associated with a `match:kind` matcher in file packages/patterns/src/patterns/patternMatchers.js...
* if the payload corresponds with a known pass style, the range covers that pass style specifically
* else if the payload corresponds with a specifically enumerated Tagged kind, the range covers all Taggeds
* else the payload cannot be understood, and the range covers "everything" (_subject to the relevant prefix for `serializeBodyFormat` and the passStylePrefixes reservation of `|`_)

That state emerged fully formed in https://github.com/endojs/endo/pull/3053/changes/559167b8248619dfb22d20b84e647963d78f7b5f , so it looks like I did take the future-compatible path in a rebase along the way. I'll update the PR description accordingly.

gibson042:
This PR has been rebased and is once again ready for review. #3073 (adding `M.choose`) is now explicitly stacked upon it.

.changeset/new-lizards-switch.md
erights: I actually don't understand what the `['*']` means here. Both square brackets (meaning computed) and quotes (meaning non-computed)?

erights: Separately, why deprecate `FullRankCover`? Is a motivation stated anywhere?

erights: Oh! You meant the `*` literally. I assumed you meant "any string goes here". With a literal `*`, the quotes and square brackets make sense. Now that I understand better, question withdrawn too. Resolving

.changeset/new-lizards-switch.md
```diff
@@ -0,0 +1,10 @@
...
+'@endo/patterns': minor
+---
+
+- `FullRankCover` is deprecated. Instead, use `provideStaticRanks(encodePassable)['*'].cover`.
+- `getPassStyleCover(passStyle)` is deprecated. Instead, use `provideStaticRanks(encodePassable)[p
+assStyle].cover`.
```
erights:
```suggestion
- `getPassStyleCover(passStyle)` is deprecated. Instead, use `provideStaticRanks(encodePassable)[passStyle].cover`.
```

.changeset/new-lizards-switch.md
```diff
@@ -0,0 +1,10 @@
...
+- `getRankCover` now works with format "compactOrdered", and produces tighter bounds for arrays and `M.nat()` (which has an implicit lower bound at 0).
+- `intersectRankCovers` and `unionRankCovers` now reject a non-empty list of covers (not that an empty list was ever semantically valid).
```
erights:
```suggestion
- `intersectRankCovers` and `unionRankCovers` now reject an empty list of covers (not that an empty list was ever semantically valid).
```

kriskowal (approved):

PR #3056, opened 2026-01-20
fix(memoize): cannot recur through memo with same arg

erights:
Closes: #XXXX
Refs: #3042 

## Description

Makes the memoize package more robust in two ways:
- if the `arg` is not WeakKey compatible, it now errors before calling the underlying function, not after.
- if the underlying function recurs by calling its own memoization with the same `args`, that causes an error. Otherwise it could have caused an infinite regress.

### Security Considerations

Removes a possibly exploitable vulnerability. Though, even if not fixed, this potential vulnerability is highly unlikely to be exploitable.

### Scaling Considerations

none.

I changed the readme to say "before" rather than "after". But I do not document anywhere the recursion issue as it is too obscure to call anyone's attention to. Rather, it avoids the need to document the (previously unnoticed) problem. Likewise, I did not update NEWS.md

### Testing Considerations

This PR tests these changes well.

In theory, but highly highly unlikely it would actually break anything.

### Upgrade Considerations

none

### DIFF
--- a/packages/memoize/README.md
+++ b/packages/memoize/README.md
@@ -50,11 +50,8 @@ result1 === result2; // true
 - **Rejected promises are memoized**: If the wrapped function returns a
   rejected promise, that promise is cached like any other return value.
 - **Invalid keys throw**: If an argument is not a valid `WeakMap` key
-  (e.g., a primitive string or number), the memoized function throws after
+  (e.g., a primitive string or number), the memoized function throws before
   invoking the wrapped function.
-  The implementers wish to reserve the option of changing this to throw
-  *before* invoking the wrapped function.
-  Please refrain from depending upon this detail.
 
 ## Memoization Safety
 


packages/memoize/src/memoize.js
```diff
@@ -17,9 +22,23 @@ export const memoize = fn => {
   const memo = new WeakMap();
   const memoFn = arg => {
     if (memo.has(arg)) {
-      return memo.get(arg);
+      const memoedResult = memo.get(arg);
+      if (Object.is(memoedResult, encapsulatedPumpkin)) {
```
gibson042: `encapsulatedPumpkin` is known to be an object.
```suggestion
      if (memoedResult === encapsulatedPumpkin) {
```

packages/memoize/src/memoize.js
```diff
@@ -17,9 +22,23 @@ export const memoize = fn => {
...
+        throw new TypeError('no recursion through memoization with same arg');
+      }
+      return memoedResult;
+    }
+    // This both prevents recursion through memoization,
+    // and errors early on a non-weak-key-compat arg, rather than calling `fn`.
+    memo.set(arg, encapsulatedPumpkin);
```
gibson042: Possible typing improvements:
```suggestion
 * @template {WeakKey} A
 * @template R
 * @param {(arg: A) => R} fn
 * @returns {(arg: A) => R}
 */
export const memoize = fn => {
  /** @type {WeakMap<A, R>} */
  const memo = new WeakMap();
  /** @type {(arg: A) => R} */
  const memoFn = arg => {
    if (memo.has(arg)) {
      const memoedResult = memo.get(arg);
      if (memoedResult === encapsulatedPumpkin) {
        throw new TypeError('no recursion through memoization with same arg');
      }
      return memoedResult;
    }
    // This both prevents recursion through memoization,
    // and errors early on a non-weak-key-compat arg, rather than calling `fn`.
    memo.set(arg, /** @type {R} */ (encapsulatedPumpkin));
```

packages/memoize/src/memoize.js
erights: Done, thanks.

packages/memoize/src/memoize.js
erights: Required one more type adjustment to make TS happy. Done, thanks.

PR #3057, opened 2026-01-26
fix(memoize): I missed another place doc-ing old behavior

erights:
Closes: #XXXX
Refs: #3056 #3042

## Description

#3056 made memoize more robust, and tried to fix the docs to match. But it missed one fixed by this PR.

@gibson042 noticed more stale documentation in that same file and suggested fixes.

### Security Considerations

none
### Scaling Considerations

none

the point. (I wonder how soon we can depend on our AI tools to notice such inconsistencies on their own)

### Testing Considerations

none

none
### Upgrade Considerations

none

### DIFF
--- a/packages/memoize/docs/memoize.md
+++ b/packages/memoize/docs/memoize.md
@@ -36,9 +36,11 @@ export const memoize = fn => {
 };
 harden(memoize);
 ```
+
 By "contingent safety", we mean the safety guarantees that follow given that
 certain requirements are met. Before we examine these, let's first understand
 the non-contingent semantics of this code.
+
 ## Base semantics
 
 ***Hardened JS***: Even the supposedly non-contingent base semantics does
@@ -54,12 +56,14 @@ Given a function `fn`, the call `memoize(fn)` returns a `fn`-like function, `mem
 The one-arg function `memoFn` is a memoizing form of `fn`
 as a one-argument function. When `memoFn(arg)` is called the first time for any
 given `arg`, it calls `fn(arg)`.
+
 If `arg` is a valid `WeakMap` key and `fn(arg)` returns a result
 rather than throwing, then the mapping from `arg` to this result is memoized in
 `memoFn`'s encapsulated `WeakMap`, and `result` is returned.
 All further calls `memoFn(arg)` with the same
 `memoFn` and the same `arg` will return the same memoized result
 without calling `fn`.
+
 Otherwise:
    * If `arg` is not a valid WeakMap key, then `memoFn(arg)` throws without any
      effect.


packages/memoize/docs/memoize.md
```diff
@@ -46,9 +46,8 @@ without calling `fn`.
 Otherwise:
    * If `fn(arg)` throws, then `memoFn(arg)` throws without any further effect
      beyond that performed by the `fn(arg)` call.
-   * If `fn(arg)` returns a `result`, but `arg` is not a valid WeakMap key,
-     then `memoFn(arg)` throws without any further effect
-     beyond that performed by the `fn(arg)` call.
+   * If `arg` is not a valid WeakMap key,
+     then `memoFn(arg)` throws without any effect.
```
gibson042: I think this change only holds if the preceding code is also updated:
`````suggestion
```js
/**
 * @template {WeakKey} A
 * @template R
 * @param {(arg: A) => R} fn
 * @returns {(arg: A) => R}
 */
  const memo = new WeakMap();
  const memoFn = arg => {
    if (memo.has(arg)) {
      const memoedResult = memo.get(arg);
      if (memoedResult === encapsulatedPumpkin) {
        throw new TypeError('no recursion through memoization with same arg');
      }
      return memoedResult;
    }
    memo.set(arg, encapsulatedPumpkin);
    let result;
    try {
      result = fn(arg);
    } catch (e) {
      memo.delete(arg);
      throw e;
    }
    memo.set(arg, result);
    return result;
  };
  return harden(memoFn);
};
harden(memoize);
```

By "contingent safety", we mean the safety guarantees that follow given that
certain requirements are met. Before we examine these, let's first understand
the non-contingent semantics of this code.

## Base semantics

Given a function `fn`, the call `memoize(fn)` returns a `fn`-like function, `memoFn`.
The one-arg function `memoFn` is a memoizing form of `fn`
as a one-argument function. When `memoFn(arg)` is called the first time for any
given `arg`, it calls `fn(arg)`.

If `arg` is a valid `WeakMap` key and `fn(arg)` returns a result
rather than throwing, then the mapping from `arg` to this result is memoized in
`memoFn`'s encapsulated `WeakMap`, and `result` is returned.
All further calls `memoFn(arg)` with the same
`memoFn` and the same `arg` will return the same memoized result
without calling `fn`.

Otherwise:
   * If `arg` is not a valid WeakMap key, then `memoFn(arg)` throws without any
     effect.
   * If `arg` is  a valid WeakMap key, but `fn(arg)` throws, then `memoFn(arg)`
     propagates the error without any further effect.
`````

It might also be worth mentioning the dependence upon a hardened or at least sufficiently primordial environment, with `memo = new WeakMap()` and `memo.has`/`memo.get`/`memo.set`/`memo.delete` all presuming ECMA-262 behavior rather than anything that might have been introduced by other code.

erights: Done

PR #3058, opened 2026-01-26
feat(nat,marshal,ocapn): help support apps script limitations

erights:
Closes: #XXXX
Refs: #3008  , #1686 , #1582
Refs: https://github.com/dckc/inter-fun/blob/main/gapp/unmarshal.js
Refs: https://www.google.com/search?q=what+version+of+ecmascript+does+apps+script+support&oq=what+version+of+ecmascript+does+apps+script+support&gs_lcrp=EgZjaHJvbWUyBggAEEUYOTIGCAEQRRg7MgYIAhBFGDsyBggDEEUYOzIGCAQQLhhA0gEHODg4ajBqMagCALACAA&sourceid=chrome&ie=UTF-8 

## Description

One of the reasons @dckc reimplemented marshal as https://github.com/dckc/inter-fun/blob/main/gapp/unmarshal.js is to be able to use it in Apps Script. After marshal adapts to #3008 , it should be much easier to create a marshal with far fewer dependencies that should be adequate for these purposes. However, at least one annoying problem would prevent that.

 https://www.google.com/search?q=what+version+of+ecmascript+does+apps+script+support&oq=what+version+of+ecmascript+does+apps+script+support&gs_lcrp=EgZjaHJvbWUyBggAEEUYOTIGCAEQRRg7MgYIAhBFGDsyBggDEEUYOzIGCAQQLhhA0gEHODg4ajBqMagCALACAA&sourceid=chrome&ie=UTF-8
 
at one point said 

> Literal syntax limitation: The shortcut syntax for `BigInt` literals (e.g., `10n`) is not supported by the script editor’s parser, and will cause a syntax error. You must use the `BigInt()` constructor with a string argument instead (e.g., `BigInt("10"))`.

Actually, when a number is accurate, we can use that instead of a string.

Endo is not in general trying for compat with Apps Script. But packages that will have minimal dependencies after adapting to https://github.com/endojs/endo/pull/3008 might, such as `@endo/marshal` and `@endo/ocapn`. This PR readies such packages for that by avoiding the bigint literals that would prevent that.

### Security Considerations
none
### Scaling Considerations

none

none
### Testing Considerations

none

the point. After we adapt to #3008, this PR will help enable some packages (marshal, ocapn) to run under Apps Script despite the limitations quoted above. There may be other problems, but at least this PR eliminates one known problem.

### Upgrade Considerations
none.

### DIFF
--- a/packages/nat/src/index.js
+++ b/packages/nat/src/index.js
@@ -16,6 +16,46 @@
 // @ts-check
 
 /**
+ * Regarding Google Apps Script limitations,
+ * https://www.google.com/search?q=what+version+of+ecmascript+does+apps+script+support&oq=what+version+of+ecmascript+does+apps+script+support&gs_lcrp=EgZjaHJvbWUyBggAEEUYOTIGCAEQRRg7MgYIAhBFGDsyBggDEEUYOzIGCAQQLhhA0gEHODg4ajBqMagCALACAA&sourceid=chrome&ie=UTF-8
+ * at one point said
+ * > Literal syntax limitation: The shortcut syntax for `BigInt` literals
+ * > (e.g., `10n`) is not supported by the script editor’s parser,
+ * > and will cause a syntax error. You must use the `BigInt()` constructor
+ * > with a string argument instead (e.g., `BigInt("10"))`.
+ * Actually, when a number is accurate, we can use that instead of a string.
+ *
+ * Endo is not in general trying for compat with Apps Script. But packages that
+ * will have minimal dependencies after adapting to
+ * https://github.com/endojs/endo/pull/3008
+ * might, such as `@endo/marshal` and `@endo/ocapn`.
+ */
+export const ZERO_N = BigInt(0);
+/**
+ * Regarding Google Apps Script limitations,
+ * https://www.google.com/search?q=what+version+of+ecmascript+does+apps+script+support&oq=what+version+of+ecmascript+does+apps+script+support&gs_lcrp=EgZjaHJvbWUyBggAEEUYOTIGCAEQRRg7MgYIAhBFGDsyBggDEEUYOzIGCAQQLhhA0gEHODg4ajBqMagCALACAA&sourceid=chrome&ie=UTF-8
+ * at one point said
+ * > Literal syntax limitation: The shortcut syntax for `BigInt` literals
+ * > (e.g., `10n`) is not supported by the script editor’s parser,
+ * > and will cause a syntax error. You must use the `BigInt()` constructor
+ * > with a string argument instead (e.g., `BigInt("10"))`.
+ * Actually, when a number is accurate, we can use that instead of a string.
+ *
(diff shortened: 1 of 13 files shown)


packages/nat/src/index.js
```diff
@@ -15,6 +15,46 @@
...
+ * > and will cause a syntax error. You must use the `BigInt()` constructor
+ * > with a string argument instead (e.g., `BigInt("10"))`.
+ * Actually, when a number is accurate, we can use that instead of a string.
+ *
+ * Endo is not in general trying for compat with Apps Script. But packages that
+ * will have minimal dependencies after adapting to
+ * https://github.com/endojs/endo/pull/3008
+ * might, such as `@endo/marshal` and `@endo/ocapn`.
+ */
+export const ONE_N = BigInt(1);
```
gibson042, line 49: Given the quoted text, should this be `BigInt('1')`?

packages/nat/src/index.js
```diff
@@ -15,6 +15,46 @@
...
+ * > and will cause a syntax error. You must use the `BigInt()` constructor
+ * > with a string argument instead (e.g., `BigInt("10"))`.
+ * Actually, when a number is accurate, we can use that instead of a string.
+ *
+ * Endo is not in general trying for compat with Apps Script. But packages that
+ * will have minimal dependencies after adapting to
+ * https://github.com/endojs/endo/pull/3008
+ * might, such as `@endo/marshal` and `@endo/ocapn`.
+ */
+export const ZERO_N = BigInt(0);
```
gibson042, line 33: Rather than repeating this large comment, we could define one constant in terms of the other, e.g. `export const ZERO_N = ONE_N - ONE_N;`.
```suggestion
```

gibson042 (approved):
This is good, but I think we should also have an eslint rule to forbid use of bigint literals—basically an inversion of [`prefer-bigint-literals`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-bigint-literals.md). I'd be ok with an issue if you don't want to include that here, though.

packages/nat/src/index.js
erights: I'm trying to provide meaningful hover info for both names. I tried your suggestion, and the derived one, as I expected, no longer showed a useful hover.

I also tried
```
/**
 * bug doc-comment
 */
export const [ZERO_N, ONE_N] = [BigInt(0), BigInt(1)];
```
which I thought would work, but it also broke the hover text. Leaving as is.

erights: Resolving.

packages/nat/src/index.js
erights: I do add to the quote

> Actually, when a number is accurate, we can use that instead of a string.

which IMO is adequate, and better code than what the comment itself suggests. So I'm not changing this either. Resolving.

PR #3059, opened 2026-01-28
refactor(daemon): Use branded types for validated strings

kriskowal:
## Description

This PR introduces branded string types for identifiers and names in the `@endo/daemon` package, clarifying that some strings have been pre-validated and can be relied upon to be in the valid range of pet names, special names, or either.

### Security Considerations

No new security considerations. The branded types improve type safety but don't change runtime behavior beyond ensuring that validation has occurred. Some care needed to ensure string types are cast only after runtime validation.

### Scaling Considerations

None.

The branded types make the codebase more self-documenting. Parameter types now clearly indicate what kind of string is expected (e.g., `FormulaNumber` vs `NodeNumber` vs `PetName`).

### Testing Considerations

Existing tests pass. New tests added for:
- `storeBlob` with subdirectory paths
- `storeBlob` requiring a name parameter
- `storeBlob` with empty path rejection

No breaking changes to the public API. Internal type changes are transparent to callers since public methods accept `string` and validate internally.

### Upgrade Considerations

None.

### DIFF
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -12,9 +12,16 @@ import { makeMailboxMaker } from './mail.js';
 import { makeGuestMaker } from './guest.js';
 import { makeHostMaker } from './host.js';
 import { makeRemoteControlProvider } from './remote-control.js';
-import { assertPetName } from './pet-name.js';
+import { assertPetName, assertName } from './pet-name.js';
 import { makeContextMaker } from './context.js';
-import { assertValidNumber, parseId, formatId } from './formula-identifier.js';
+import {
+  assertValidNumber,
+  assertValidId,
+  assertFormulaNumber,
+  assertNodeNumber,
+  parseId,
+  formatId,
+} from './formula-identifier.js';
 import { makeSerialJobs } from './serial-jobs.js';
 import { makeWeakMultimap } from './multimap.js';
 import { makeLoopbackNetwork } from './networks/loopback.js';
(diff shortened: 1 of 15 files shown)


packages/daemon/src/daemon.js
kumavis, line 111: prefer branded return type?

packages/daemon/src/daemon.js
```diff
@@ -63,7 +69,10 @@ const delay = async (ms, cancelled) => {
  */
 const makeInspector = (type, number, record) =>
   makeExo(`Inspector (${type} ${number})`, InspectorInterface, {
-    lookup: async petName => {
+    lookup: async petNameOrPath => {
+      const petName = Array.isArray(petNameOrPath)
+        ? petNameOrPath[0]
```
kumavis: expected to ignore the rest of the path here?

packages/daemon/src/daemon.js
```diff
@@ -823,17 +831,21 @@ const makeDaemonCore = async (
   };
 
   /**
-   * @param {string} nodeId
+   * @param {NodeNumber} nodeNumber
    * @returns {Promise<string>}
```
kumavis: prefer branded return type?

packages/daemon/src/host.js
```diff
@@ -116,41 +127,42 @@ export const makeHostMaker = ({
...
     };
 
     /**
-     * @param {string[]} workerNamePath
+     * @param {string | string[]} workerNamePath
      */
     const provideWorker = async workerNamePath => {
-      const workerId = await E(directory).identify(...workerNamePath);
+      const namePath = namePathFrom(workerNamePath);
+      assertNamePath(namePath);
```
kumavis, line 188: seems redundant / footgun to assertNamePath after namePathFrom. maybe can do inside namePathFrom

packages/daemon/src/types.d.ts
```diff
@@ -3,6 +3,40 @@ import type { ERef } from '@endo/eventual-send';
...
 
+// Branded string types for pet names and special names
+declare const PetNameBrand: unique symbol;
+declare const SpecialNameBrand: unique symbol;
+declare const FormulaNumberBrand: unique symbol;
+declare const NodeNumberBrand: unique symbol;
+declare const FormulaIdentifierBrand: unique symbol;
+
+/** A validated pet name (lowercase, e.g., 'worker', 'my-app') */
+export type PetName = string & { [PetNameBrand]: true };
```
kumavis, line 14: nice. ive always done this with `_brand` string keys and a unique value, but this is better.

packages/daemon/src/types.d.ts
```diff
@@ -3,6 +3,40 @@ import type { ERef } from '@endo/eventual-send';
...
+export type FormulaNumber = string & { [FormulaNumberBrand]: true };
+
+/** A 128-character hex string identifying a node */
+export type NodeNumber = string & { [NodeNumberBrand]: true };
+
+/** A full formula identifier in the format {FormulaNumber}:{NodeNumber} */
+export type FormulaIdentifier = string & { [FormulaIdentifierBrand]: true };
+
+/** Either a pet name or a special name */
+export type Name = PetName | SpecialName;
```
kumavis, line 29: generally dislike single-word exported identifiers, but ya know `Name`ing things is hard

kumavis (approved):
theres a lot of inline casting, usually i find that a motivation to put it behind some util function that does the casting and maybe the assertion that it matches the format.

but generally looks like an improvement

packages/daemon/src/types.d.ts
kriskowal: This approach has an eval-twin problem, but my reasoning is that `daemon` should not accommodate eval twins.

PR #3061, opened 2026-01-28
chore(pass-style): make `CopyArray` compatible with `ReadonlyArray`

### LINKED ISSUE #3060: Arrays typed as `readonly` aren't Passable
## Describe the bug

ReadonlyArrays aren't Passable because they don't conform to CopyArrayInterface.  CopyArrays are hardened, thus the mutating methods will always fail, and marking them as `readonly` only changes the typing to omit the mutating methods.

## Steps to reproduce

1. Create `repro.ts` with the following contents:
```ts
import type { Passable } from '@endo/pass-style';
export default ['ibc'] as Readonly<[string]> satisfies Passable;
```

2. Load in IDE or compile with `tsc`, and see error:

```
Type 'readonly [string]' does not satisfy the expected type 'Passable'.
  Type 'readonly [string]' is not assignable to type 'CopyArrayInterface<PassableCap, Error>'.ts(1360)
```

## Expected behavior

No error.

## Platform environment

 - What OS are you using? macOS
 - What version of Node.js? v22.21.0
 - Is there anything special/unusual about your platform? not especially
 - What version of Endo are you using? @endo/import-bundle@1.5.2-303-g31f81e8ac`

michaelfig:
Closes: #3060

## Description

This PR updates the CopyArray<T> type definition from Array<T> to ReadonlyArray<T> to better reflect that CopyArrays are hardened and immutable at runtime. This allows readonly arrays and tuples to be accepted as Passable, resolving issue https://github.com/endojs/endo/issues/3060.

### Security Considerations

Types now reflect that `@endo/pass-style`'s Passable type doesn't require `CopyArrays` to be mutable.

### Scaling Considerations

n/a

If you typed a mutable array as a `CopyArray`, Typescript will show errors if you try to mutate it directly, or call methods that would mutate it.  You can achieve the original behavior by typing your array as `Passable[]`.

### Testing Considerations

Type tests have been added.

Makes for better compatibility.

### Upgrade Considerations

No runtime change, typing only.

### DIFF
--- a/packages/pass-style/src/types.d.ts
+++ b/packages/pass-style/src/types.d.ts
@@ -116,11 +116,11 @@ export type Container<PC extends PassableCap, E extends Error> =
   | CopyArrayInterface<PC, E>
   | CopyRecordInterface<PC, E>
   | CopyTaggedInterface<PC, E>;
-interface CopyArrayInterface<PC extends PassableCap, E extends Error>
+export interface CopyArrayInterface<PC extends PassableCap, E extends Error>
   extends CopyArray<Passable<PC, E>> {}
-interface CopyRecordInterface<PC extends PassableCap, E extends Error>
+export interface CopyRecordInterface<PC extends PassableCap, E extends Error>
   extends CopyRecord<Passable<PC, E>> {}
-interface CopyTaggedInterface<PC extends PassableCap, E extends Error>
+export interface CopyTaggedInterface<PC extends PassableCap, E extends Error>
   extends CopyTagged<string, Passable<PC, E>> {}
 
 export type PassStyleOf = {
@@ -134,7 +134,7 @@ export type PassStyleOf = {
   (p: Promise<any>): 'promise';
   (p: Error): 'error';
   (p: CopyTagged): 'tagged';
-  (p: any[]): 'copyArray';
+  (p: readonly any[]): 'copyArray';
   (p: Iterable<any>): 'remotable';
   (p: Iterator<any, any, undefined>): 'remotable';
   <T extends PassStyled<PassStyleMarker, any>>(p: T): ExtractStyle<T>;
@@ -200,9 +200,14 @@ export type PassableCap =
   | RemotableBrand<any, any>;
 
 /**
+ * Types you would get from Awaited<PassableCap>
+ */
+export type AwaitedPassableCap = RemotableObject | RemotableBrand<any, any>;
+
+/**
  * A Passable sequence of Passable values.
  */
-export type CopyArray<T extends Passable = any> = Array<T>;
+export type CopyArray<T extends Passable = any> = ReadonlyArray<T>;
 
 /**
  * A hardened immutable ArrayBuffer.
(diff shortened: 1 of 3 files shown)


michaelfig:
@copilot, can you specify a few examples of exactly where it is in "the patterns package that uses incorrect type annotations for mutable accumulators"?

packages/pass-style/src/types.d.ts
```diff
@@ -204,6 +207,11 @@ export type PassableCap =
  */
 export type CopyArray<T extends Passable = any> = Array<T>;
 
+/**
+ * A Passable and Readonly sequence of Passable values.
+ */
+export type CopyReadonlyArray<T extends Passable = any> = ReadonlyArray<T>;
```
gibson042: Why introduce a new type? What breaks if CopyArray is instead updated to use ReadonlyArray rather than Array?

michaelfig: Thanks for asking!

The breaking change is that there are some uses in the wild of `CopyArray` as the type of a mutable array, and then the code proceeds to mutate the array.  If `CopyArray = ReadonlyArray`, the code fails to typecheck.

At this stage in the game, I'm ~very reluctant~ somewhat willing to introduce breaking changes.

michaelfig: @copilot, for this `endojs/endo` repository, in `*.js` files containing JSDoc TypeScript types and `*.ts` TypeScript files, can you please identify all arrays that are declared to be of type `CopyArray` that are later mutated?

Use shell commands with intermediary files whenever possible to save your work and avoid consuming too much of your agent's context space.

michaelfig: Besides the highly enthusiastic affirmations of my AI sycophant, I also see that a mutable `CopyArray` type is useful to enforce that all its elements are Passable.

michaelfig: Having said all that, the friendly bots scoured `Agoric/agoric-sdk` and `endojs/endo`, and only `@endo/patterns/src/patterns/patternMatchers.js` uses a mutable `CopyArray` in its internals.  What do you see as the best way forward?

gibson042: #3065 will fix the typing that makes it look like patternMatchers.js is attempting CopyArray mutation, and any mutability associated with Passable is itself a bug whose footprint I'd rather reduce than expand. So my preference is to land #3065 and then update this PR to tighten CopyArray without introducing a new type.

kriskowal:
@gibson042 @turadg Can we confirm this is superseded by https://github.com/endojs/endo/pull/3172 and close, or follow-up?

Related https://github.com/endojs/endo/pull/3063

michaelfig:
> @gibson042 @turadg Can we confirm this is superseded by #3172 and close, or follow-up?

#3172 does indeed contain the following:

```js
export type CopyArray<T extends Passable = any> = readonly T[];
```

However, I don't see the equivalent of [this PR's widening of PassStyleOf](https://github.com/endojs/endo/pull/3061/changes#diff-76b810d2ec9accf2da723b1f2124ca94003d004e4c4e2a4ecccc438d1fcbc200R137)

which could be more succinctly expressed as:

```js
  (p: CopyArray): 'copyArray';
```

I don't have a round tuit on me right now, so either this languishes until I get one, or if somebody with an interest in pass-style wants to take it on, please do!

PR #3065, opened 2026-01-29
fix(patterns): Update `containerHasSplit(copyArray, ...)` to return copyArray results

### LINKED ISSUE #3064: `containerHasSplit(copyArray, …)` returns mutable in/out arrays
## Describe the bug

CopyArray, CopySet, and CopyBag are all immutable, and `containerHasSplit` returns results with similar immutability for the latter two but not the first.

## Steps to reproduce

```js
import '@endo/init';
import { M, makeCopySet, makeCopyBag, containerHasSplit } from '@endo/patterns';

const input = ['foo', 'bar', 'baz', 'qux'];
const bound = 3n;
for (const [kind, [makeSpecimen, makeElementFromKey]] of Object.entries({
  copyArray: [harden],
  copySet: [makeCopySet],
  copyBag: [makeCopyBag, k => [k, 2n]],
})) {
  const specimen = makeSpecimen(input.map(makeElementFromKey || (k => k)));
  const [keep] = containerHasSplit(specimen, M.any(), bound, assert.Fail, true);
  console.log(
    `containerHasSplit(${kind}, …)`,
    Object.isFrozen(keep)
      ? 'output array is frozen'
      : 'output array is mutable',
    ':',
    keep,
  );
}
```

## Expected behavior

All outputs are hardened:
```
containerHasSplit(copyArray, …) output array is frozen : [ 'qux', 'baz', 'bar' ]
containerHasSplit(copySet, …) output array is frozen : Object [copySet] { payload: [ 'foo', 'baz', 'bar' ] }
containerHasSplit(copyBag, …) output array is frozen : Object [copyBag] { payload: [ [ 'baz', 1n ], [ 'bar', 2n ] ] }
```

## Actual behavior

Array outputs are not hardened:
```diff
-containerHasSplit(copyArray, …) output array is frozen : [ 'qux', 'baz', 'bar' ]
+containerHasSplit(copyArray, …) output array is mutable : [ 'qux', 'baz', 'bar' ]
 containerHasSplit(copySet, …) output array is frozen : Object [copySet] { payload: [ 'foo', 'baz', 'bar' ] }
 containerHasSplit(copyBag, …) output array is frozen : Object [copyBag] { payload: [ [ 'baz', 1n ], [ 'bar', 2n ] ] }

### PR DESCRIPTION
**Pull request by gibson042**
Fixes #3064

## Description

* adds test coverage for `containerHasSplit`
* hardens copyArray results from `containerHasSplit` so they actually have pass style "copyArray"
* improves typing to avoid attempted mutation of a CopyArray (which is defined to be immutable)
* refactors some code supporting `containerHasSplit`

### Security Considerations

None known.

### Scaling Considerations

n/a

### Documentation Considerations

n/a

### Testing Considerations

✔

### Compatibility Considerations

CopyArrays are already claimed by documentation to be immutable, so this shouldn't affect anything (and doesn't AFAICT).

### Upgrade Considerations

I think this fix falls below the level of inclusion in NEWS.md, but could be persuaded out of that position.

### DIFF
--- a/.changeset/short-horses-buy.md
+++ b/.changeset/short-horses-buy.md
@@ -1,0 +1,6 @@
+---
+'@endo/patterns': patch
+---
+
+- `containerHasSplit` now hardens its output(s) when working with copyArrays,
+  ensuring that each output is itself a copyArray instance.

### CODE REVIEW

File: packages/patterns/test/containerHasSplit.test.js [hunk 1]
```diff
@@ -0,0 +1,223 @@
...
+    expectRejected: [2, 'foo', 1],
+  });
+
+  test('split first two matches from copyArray', testContainerHasSplit, {
+    specimen,
+    pattern: M.string(),
+    bound: 2n,
+    expectAccepted: ['bar', 'foo'],
+    expectRejected: [2, 1],
+  });
```
gibson042, line 97: @erights Whether or not this was intentional, we should reconsider the direction for CopyArray scanning. It makes sense for CopySet and CopyBag to start at the end because of their ordering constraints, but I think CopyArray scanning should start at the beginning.

erights: That makes sense. But I won't get to it anytime soon, so putting this in Draft until then. Unless you wanna give it a try?

gibson042: This PR is ready for review now to hopefully merge soon (and in particular ahead of #3061), and isn't changing that behavior. But since you're amenable to it, I've opened #3067 as a followup.

packages/patterns/test/containerHasSplit.test.js
```diff
@@ -0,0 +1,223 @@
...
+{
+  const specimen = harden([1, 'foo', 2, 'bar']);
+
+  test('split first match from copyArray', testContainerHasSplit, {
+    specimen,
+    pattern: M.string(),
+    bound: 1n,
+    expectAccepted: ['bar'],
+    expectRejected: [2, 'foo', 1],
+  });
```
michaelfig, line 89: Is `bound: 0n` a relevant edge case to test?

gibson042: No, the function is documented to require a minimum `bound` of 1n: https://github.com/endojs/endo/blob/3b533de28517a14233ac80331342ef62fe880bef/packages/patterns/src/patterns/patternMatchers.js#L1426

But I have added https://github.com/endojs/endo/pull/3065/commits/1d6b76e996b12172cfa6a764035780107da1a4c3 for hopefully propagating such documentation to https://docs.endojs.org/variables/_endo_patterns.containerHasSplit.html .

erights:
> I'll go out on a bit of a limb, and assert that this PR is strictly an improvement.
> 
> @gibson042, if you would like @erights to review before merging, please bring it to his attention.

Hi @michaelfig , that brought it back to my attention. Looking now. Thanks.

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1346,22 +1348,21 @@ const makePatternKit = () => {
...
-        }
-      } else if (outResults === undefined) {
-        break;
-      } else {
+      if (inCount >= bound) {
+        if (!outResults) break;
+        outResults.push(element);
+        continue;
+      }
+      if (matches(element, elementPatt)) {
```
erights: Minor, but I prefer

```suggestion
      } else if (matches(element, elementPatt)) {
```

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1370,8 +1371,8 @@ const makePatternKit = () => {
    * @param {Pattern} elementPatt
    * @param {bigint} bound Must be >= 1n
    * @param {Rejector} reject
-   * @param {CopyArray<[Key, bigint]>} [inResults]
-   * @param {CopyArray<[Key, bigint]>} [outResults]
+   * @param {Array<[Key, bigint]>} [inResults]
+   * @param {Array<[Key, bigint]>} [outResults]
```
erights: Minor, but I prefer

```suggestion
   * @param {[Key, bigint][]} [inResults]
   * @param {[Key, bigint][]} [outResults]
```
especially given your change on lines 1329-1330 above

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1392,44 +1393,51 @@ const makePatternKit = () => {
...
-      } else if (outResults === undefined) {
-        break;
-      } else {
+      const stillNeeds = bound - inCount;
+      if (stillNeeds <= 0n) {
+        if (!outResults) break;
+        outResults.push([element, num]);
+        continue;
+      }
+      if (matches(element, elementPatt)) {
```
erights: Minor preference again

```suggestion
      } else if (matches(element, elementPatt)) {
```

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1392,44 +1393,51 @@ const makePatternKit = () => {
...
+        const isPartial = num > stillNeeds;
+        const numTake = isPartial ? stillNeeds : num;
+        inCount += num;
+        if (inResults) inResults.push([element, numTake]);
+        if (isPartial && outResults) outResults.push([element, num - numTake]);
```
erights, line 1404: Just checking: This is all a pure refactor with no observable diff, yes?

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1392,44 +1393,51 @@ const makePatternKit = () => {
...
+      (reject && reject`Has only ${q(inCount)} matches, but needs ${q(bound)}`)
     );
   };
 
   /**
+   * Confirms that `specimen` contains at least `bound` instances of an element
+   * matched by `elementPatt`, optionally returning those bounded matches and/or
+   * their complement as specified by `needInResults` and `needOutResults`
+   * (ensuring for CopyBags that at most one Key is split across both, but
+   * otherwise making no guarantee regarding the order in which elements are
```
erights, line 1420: Really? That seems like a surprising under-specification.

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1392,44 +1393,51 @@ const makePatternKit = () => {
...
+   * considered beyond a best-effort attempt to align with intuition).
+   * If the specimen does not contain enough matching instances, this function
+   * terminates as directed by `reject` (i.e., either returning `false` or
+   * throwing an exception).
```
erights:
```suggestion
   * throwing an error).
```
JS has no exceptions

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1392,44 +1393,51 @@ const makePatternKit = () => {
...
    * @param {bigint} bound Must be >= 1n
    * @param {Rejector} reject
-   * @param {boolean} [needInResults]
-   * @param {boolean} [needOutResults]
-   * @returns {[Container | undefined, Container | undefined] | false}
+   * @param {boolean} [needInResults] collect and return matches inside a
+   *   container of the same shape as `specimen`
+   * @param {boolean} [needOutResults] collect and return rejects inside a
+   *   container of the same shape as `specimen`
+   * @returns {[matches: Container | undefined, discards: Container | undefined] | false}
```
erights, line 1435: Good. I always forget that we can name the elements of a tuple type.

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1442,56 +1450,46 @@ const makePatternKit = () => {
...
             reject,
             inResults,
             outResults,
-          )
-        ) {
-          // check logic already performed by confirmContainerHasSplit
-          return false;
-        }
-        return [inResults, outResults];
+          ) && [harden(inResults), harden(outResults)]
```
erights: Now that you brought the missing hardens to my attention, how about

```suggestion
          ) && harden([inResults, outResults])
```

and likewise hardening the other returned pairs below?

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1522,14 +1520,7 @@ const makePatternKit = () => {
...
       }
-      return !!confirmContainerHasSplit(
-        specimen,
-        elementPatt,
-        bound,
-        reject,
-        false,
-        false,
-      );
+      return !!containerHasSplit(specimen, elementPatt, bound, reject);
```
erights, line 1522: If it takes a `reject` parameter and follows the normal `reject` convention, shouldn't it be named `confirmFoo`?

```suggestion
      return !!confirmContainerHasSplit(specimen, elementPatt, bound, reject);
```

erights (approved):
LGTM, thanks!

packages/patterns/src/patterns/patternMatchers.js
gibson042: Perhaps, but it is externally exposed as `containerHasSplit` and things are generally more convenient if that name is consistent all the way down.

packages/patterns/src/patterns/patternMatchers.js
gibson042: Correct; https://github.com/endojs/endo/pull/3065/changes/2903ae98786db44334e8e059ec7951f136ae28db rearranges the blocks and then https://github.com/endojs/endo/pull/3065/changes/d309dc17e85d9ae92c3a96b221a993dd80231116 renames some variables.

PR #3067, opened 2026-01-30
feat(patterns): For `containerHasSplit(copyArray, ...)`, scan from the start

gibson042:
Ref https://github.com/endojs/endo/pull/3065#discussion_r2743684826

## Description

Avoids returning result arrays with elements in reversed order.

### Security Considerations

None known.

### Scaling Considerations

n/a

Ignored.

### Testing Considerations

Updated tests from #3065.

This _is_ a behavior change, but I don't think `containerHasSplit` even has use yet, nor do I expect reliance upon the counterintuitive behavior.

### Upgrade Considerations

I think this fix falls below the level of inclusion in NEWS.md, but could be persuaded out of that position.

### DIFF
--- a/packages/patterns/src/patterns/patternMatchers.js
+++ b/packages/patterns/src/patterns/patternMatchers.js
@@ -1366,7 +1366,7 @@ const makePatternKit = () => {
   };
 
   /**
-   * @param {CopyArray<[Key, bigint]>} pairs
+   * @param {CopyArray<[Key, bigint]>} pairs in descending lexicographic order
    * @param {Pattern} elementPatt
    * @param {bigint} bound Must be >= 1n
    * @param {Rejector} reject
@@ -1383,13 +1383,9 @@ const makePatternKit = () => {
     outResults = undefined,
   ) => {
     let inCount = 0n;
-    // Since this feature is motivated by ERTP's use on
-    // semi-fungible (`copyBag`) amounts,
-    // their arrays store their elements in decending lexicographic order.
-    // But this function has to make some choice amoung equally good minimal
-    // results. It is more intuitive for the choice to be the first `bound`
-    // matching elements in ascending lexicigraphic order, rather than
-    // decending. Thus we iterate `pairs` in reverse order.
+    // To produce intuitive results with CopyBag payloads (which are ordered by
+    // descending lexicographic Key order), we iterate by reverse array index
+    // and therefore consider elements in *ascending* lexicographic Key order.
     for (let i = pairs.length - 1; i >= 0; i -= 1) {
       const [element, num] = pairs[i];
       const stillNeeds = bound - inCount;


packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1326,27 +1326,26 @@ const makePatternKit = () => {
...
+    direction,
   ) => {
     let inCount = 0n;
-    // Since this feature is motivated by ERTP's use on
-    // non-fungible (`set`, `copySet`) amounts,
-    // their arrays store their elements in decending lexicographic order.
-    // But this function has to make some choice amoung equally good minimal
-    // results. It is more intuitive for the choice to be the first `bound`
-    // matching elements in ascending lexicigraphic order, rather than
-    // decending. Thus we iterate `elements` in reverse order.
```
erights, line 1349: I think a transform of this comment is still worth saying somewhere.

erights (commented):
Be aware that a Set amount value necessarily lists its elements in descending order, just like the payload of a CopySet amount value. See

https://github.com/Agoric/agoric-sdk/blob/b5638874e8aa65b633fd348139fa4980f15670ff/packages/ERTP/src/mathHelpers/setMathHelpers.js#L26

```js
 list = coerceToElements(list);
```

and 

https://github.com/endojs/endo/blob/b41c0809e25f573ff01da077b35e0cdc7109c77a/packages/patterns/src/keys/copySet.js#L94

How does that affect the desirability or relevance of this change? (Not rhetorical. I genuinely don't know yet.)

packages/patterns/src/patterns/patternMatchers.js
gibson042: It's now in the `confirmElementsHasSplit` doc comment above:
> direction -1 for picking from the end (which gives intuitive results with descending lexicographic CopySet payloads); 1 for picking from the start (which gives intuitive results with other arrays)

packages/patterns/src/patterns/patternMatchers.js
gibson042, line 1435: https://github.com/endojs/endo/pull/3065#discussion_r2825409367
> Really? That [_ensuring for CopyBags that at most one Key is split across both, but otherwise making no guarantee regarding the order in which elements are considered beyond a best-effort attempt to align with intuition_] seems like a surprising under-specification.

@erights If you would like to strengthen the guarantee to actually include the order in which elements are considered, I think this would be the right PR for it. But I'm happy to preserve flexibility for the future by keeping it relatively weak.

gibson042: Done: https://github.com/endojs/endo/pull/3067/changes/e1c92b3790c35d1b0a1d4cfcaa56591e757c3b5e

packages/patterns/src/patterns/patternMatchers.js
erights: Good. Resolving.

packages/patterns/src/patterns/patternMatchers.js
erights: Thanks. Resolving.

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1327,27 +1327,26 @@ const makePatternKit = () => {
    * @param {Pattern} elementPatt
    * @param {bigint} bound Must be >= 1n
    * @param {Rejector} reject
-   * @param {T[]} [inResults]
-   * @param {T[]} [outResults]
+   * @param {T[] | undefined} inResults
+   * @param {T[] | undefined} outResults
+   * @param {-1 | 1} direction -1 for picking from the end (which gives
+   * intuitive results with descending lexicographic CopySet payloads); 1 for
```
erights:
```suggestion
   * intuitive results with descending lexicographic CopySet or CopyBag payloads); 1 for
```

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1416,9 +1415,11 @@ const makePatternKit = () => {
    * Confirms that `specimen` contains at least `bound` instances of an element
    * matched by `elementPatt`, optionally returning those bounded matches and/or
    * their complement as specified by `needInResults` and `needOutResults`
-   * (ensuring for CopyBags that at most one Key is split across both, but
-   * otherwise making no guarantee regarding the order in which elements are
-   * considered beyond a best-effort attempt to align with intuition).
+   * (considering CopyArray elements by ascending index and CopySet/CopyBag
+   * elements in lexicographic order).
+   * Note that CopyBag elements can be split; when only some of the count
+   * associated with a single Key is necessary to bring instantaneous matches up
```
erights: "instantaneous"?

packages/patterns/src/patterns/patternMatchers.js
erights, line 1486:
```suggestion
    outResults,
    -1,
  ) &&
```

erights: or am I missing something?

gibson042: This is a call to `pairsHasSplit`, which is analogous to `confirmElementsHasSplit` but lacks a `direction` parameter because it's only applicable to CopyBag payloads and so _always_ iterates by descending array index. There's coverage in packages/patterns/test/containerHasSplit.test.js , but I've just updated src/ comments as well.

PR #3068, opened 2026-01-30
chore(pass-style): only `Promise<Passable>` is passable

michaelfig:
## Description

This PR strengthens TypeScript type safety for passable promises by requiring them to accept only `Promise<Passable>` rather than `Promise<any>`. This prevents promises that will likely fulfill to non-passable values from being accepted at compile time.

**Changes:**
- Added `AwaitedPassable` type, which only includes the non-Promise branches of the existing `Passable` union type.
- Updated `Passable` type to be `AwaitedPassable | Promise<AwaitedPassable>` to break infinite promise expansion.
- Simplified `PassByRef` type to use a single `Promise<Passable>` instead of multiple specific promise types
- Added `AwaitedPassableCap` type, which only includes the non-Promise branches of the existing `PassableCap` union type.
- Updated `PassableCap` type to be `AwaitedPassableCap | Promise<Passable>` instead of `... | Promise<any>`.
- Added test annotations to validate the new type constraints

### Security Considerations

Guardrails only.  By design, the TypeScript programming language provides pragmatic but unsound types.

### Scaling Considerations

n/a

Documented in `NEWS.md`, but should be intuitive enough, and can always be `@ts-expect-error`red if it gets in the way.

### Testing Considerations

CI, and `types.test-d.ts` linting.

Should not break naïve code, only code that is really doing something dodgy.

### Upgrade Considerations

Types only.  No effect on runtime execution.

### DIFF
--- a/packages/patterns/src/types.ts
+++ b/packages/patterns/src/types.ts
@@ -10,7 +10,7 @@ import type {
   Passable,
   PassStyle,
   Atom,
-  RemotableObject,
+  AwaitedPassableCap,
 } from '@endo/pass-style';
 import type {
   PartialCompare,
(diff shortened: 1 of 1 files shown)


turadg:
I'd love to see this tightened up. But per what Mathieu pointed out, I'm requesting that the test plan include a green integration branch with agoric-sdk. (IMO that can use ts-expect-error and other escape hatches so we can correct Endo and catch in in agoric-sdk over time.)

PR #3070, opened 2026-02-02
use changesets

boneskull:
## Overview

**This PR adopts [Changesets](https://github.com/changesets/changesets) for versioning and releasing.**

Changesets are a solution to the problem of versioning and updating release notes.  Unlike other solutions which generate release notes from [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/), Changesets do not prescribe a commit message format and provide more granular control over the contents of release notes.  Instead, it keeps information about versions and release notes in a temporary file-based "database".  These files are committed to version control, then "flushed" upon release.

### Adding Changesets

To add a Changeset to your PR, you'll interface with [`@changesets/cli`](https://npm.im/@changesets/cli), which will ask you to choose whether the Changeset represents a patch, minor, or major version bump.  It will also ask you to provide a changelog entry.  That entry replaces our current convention for updating a `NEWS.md` file.

Once complete, you'll be reminded to commit the Changeset (which will be a unique `.md` file in `.changeset/`), which you'll then push up to your PR.

> [!NOTE]
>
> - Changesets don't need their own commit. 
> - Not all PRs need changesets--only those which impact the end user.
> - [`changeset-bot`](https://github.com/changesets/bot) will automatically detect Changesets (or the lack thereof) and comment on any PR containing them with info about the version bump and the contents of the Changelog entry.

### Releasing

The [Changesets Release Action](https://github.com/changesets/action) runs after every push to `master`. It creates and maintains a PR which uses the Changeset files it finds to bump packages and update `CHANGELOG.md` files.  Once merged, this PR will create appropriate tags for all updated packages & GitHub Releases for each tag.

### Publishing

To publish to npm, a maintainer will pull down the changes, then publish via `lerna publish from-package` or `laverna` (depending; see "MAYBE TODO" below).

## TODO

An incomplete list of TODOs:

- [x] Strongly recommend installing [`changeset-bot`](https://github.com/changesets/bot) to automatically nag contributors to create Changesets and display information about Changesets it detects. This requires it to be added to the Endo org, however.
- [x] Update [`CONTRIBUTING.md`](https://github.com/endojs/endo/blob/master/CONTRIBUTING.md) to reflect the new processes.
- [x] We likely need to enable **Allow GitHub Actions to create and approve pull requests** in [the Actions settings](/endojs/endo/settings/actions)

### MAYBE TODO

This work _could_ be in this PR, or in a subsequent PR.

- [ ] Replace `lerna` with [`@lavamoat/laverna`](https://github.com/LavaMoat/LavaMoat/tree/main/packages/laverna).  Since `lerna` will no longer handle versioning, it would only provide `lerna publish from-package` ... which happens to be exactly the thing `@lavamoat/laverna` provides.  If we do this in a separate PR, we'll likely incur another change to the publishing workflow (even if slight). I could go either way on this.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,5 +31,7 @@ jobs:
       - name: Install Dependencies
         run: yarn
 
-      - name: Create Release Pull Request
+      - name: Process Changesets
+        # This action will create/update Changesets' Release PR *or* create a
+        # GitHub Release and tags if its Release PR was just merged
         uses: changesets/action@c48e67d110a68bc90ccf1098e9646092baacaa87 # v1
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,6 @@
-# Initial setup
+# Contributing to Endo
+
+## Initial setup
 
 ```sh
 git clone git@github.com:endojs/endo.git
@@ -6,12 +8,13 @@ cd endo
 yarn
 ```
 
-Endo is a yarn workspaces repository. Running yarn in the root will install and hoist most dependencies up to the root `node_modules`.
+Endo is a yarn workspaces repository. Running yarn in the root will install and
+hoist most dependencies up to the root `node_modules`.
 
 Note: running yarn `--ignore-scripts` will not complete the setup of SES.
 Note: Endo uses `lerna` only for releasing. `lerna bootstrap` is unlikely to work.
 
-## Action pinning
+### Action pinning
 
 GitHub Actions are pinned to commit SHAs.
 Run `node scripts/update-action-pins.mjs` to refresh patch/minor pins.
@@ -23,21 +26,21 @@ If no version comment exists, it infers the latest tag for that action.
 CI enforces pinning with `node scripts/update-action-pins.mjs --check-pins`.
 If this check fails, run the updater and commit the resulting changes.
 
-## Creating a new package
+### Creating a new package
 
 Run <code>[scripts/create-package.sh](./scripts/create-package.sh) $name</code>,
 then update the resulting README.md, package.json (specifically setting
 `description` and [if appropriate] removing `"private": false`), index.js, and
 index.test.js files.
 
-## Markdown Style Guide
+### Markdown Style Guide
 
 When writing Markdown documentation:
 
-* Wrap lines at 80 to 100 columns for readability in terminal editors.
-* Start each sentence on a new line.
+- Wrap lines at 80 to 100 columns for readability in terminal editors.
+- Start each sentence on a new line.
   This ensures changes in one sentence do not cascade into the next in diffs.
-* Starting sentences on new lines also obviates any question of whether to use
+- Starting sentences on new lines also obviates any question of whether to use
   one or two spaces after a period.
 
 Example:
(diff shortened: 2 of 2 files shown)


.changeset/config.json
```diff
@@ -0,0 +1,21 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@3.1.2/schema.json",
+  "access": "public",
+  "baseBranch": "master",
+  "bumpVersionsWithWorkspaceProtocolOnly": true,
```
boneskull, line 5: This assumes we're using those odious `workspace:^` version entries everywhere we should.

.changeset/config.json
```diff
@@ -0,0 +1,21 @@
...
+  "changelog": [
+    "@changesets/changelog-github",
+    {
+      "repo": "endojs/endo"
+    }
+  ],
```
boneskull, line 11: If we don't like this, it's pretty trivial to write our own changelog-generating plugin.

.changeset/config.json
```diff
@@ -0,0 +1,21 @@
...
+  "commit": false,
```
boneskull, line 12: This means "don't automatically commit after adding a Changeset".  If enabled, we can specify what the commit message should look like.  I don't have an opinion about this.

.changeset/config.json
```diff
@@ -0,0 +1,21 @@
...
+  "fixed": [],
+  "ignore": [],
+  "linked": [],
+  "privatePackages": {
+    "tag": true,
+    "version": true
+  },
```
boneskull, line 19: we have a lot of private packages.  I figure we want to version and tag them and add release notes like any other (why not?).

.changeset/config.json
```diff
@@ -0,0 +1,21 @@
...
+  "updateInternalDependencies": "patch"
```
boneskull, line 20: see [docs](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#updateinternaldependencies)

.changeset/config.json
```diff
@@ -0,0 +1,21 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@3.1.2/schema.json",
+  "access": "public",
+  "baseBranch": "master",
+  "bumpVersionsWithWorkspaceProtocolOnly": true,
+  "changelog": [
+    "@changesets/changelog-github",
```
boneskull, line 7: this changelog-generator plugin adds references to GitHub PRs, commits, etc. the default generator does not.

.github/workflows/release.yml
```diff
@@ -0,0 +1,35 @@
...
+on:
+  push:
+    branches:
+      - master
+
+concurrency: ${{ github.workflow }}-${{ github.ref }}
+
+permissions:
+  contents: write
+  pull-requests: write
```
boneskull, line 12: I _think_ this is sufficient.

.github/workflows/release.yml
```diff
@@ -0,0 +1,35 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - master
```
kriskowal, line 6: Does this mean we’re going to always have a release PR ready to go?

.github/workflows/release.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Setup Node.js 20
+        uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6
+        with:
+          node-version: 22.x
+          cache: yarn
+
+      - name: Install Dependencies
+        run: yarn
+
+      - name: Create Release Pull Request
```
kriskowal: Create or create and update?

.github/workflows/release.yml
boneskull: Not always, but as long as there are changesets sitting in `.changeset/`.

.github/workflows/release.yml
boneskull: Since this action can do one of:

- Creates a Release PR
- Updates a Release PR
- Creates a GitHub Release + tags

I've renamed the step to something more generic and added a comment explaining what it does.

PR #3073, opened 2026-02-03
feat(patterns): Add `M.choose`

### LINKED ISSUE #3072: discriminated union patterns with better error messages than M.or()
_(originally posted by @dckc at https://github.com/Agoric/agoric-sdk/issues/6115)_

We can write a pattern for a discriminated union using `M.or()`, but when a match fails, the diagnostic isn't specialized to any branch of the union.

## Description of the Design

Where we can now write...

```js
M.or(
  {kind: 'animal', eats: FoodShape},
  {kind: 'vegetable', lifetime: M.number()},
  ...
);
```

if we could write...

```js
M.choose('kind', [
  {kind: 'animal', eats: FoodShape},
  {kind: 'vegetable', lifetime: M.number()},
  ...
])
```

Then perhaps a diagnostic for `{kind: 'animal', lifetime: 3}` could say "property eats missing" rather than just "didn't match".

gibson042:
Fixes #3072
Extends #3053 for avoiding known fast-check failures

## Description

Add support for `M.choose(keyName, { keyValue1: restPatt, keyValue2: restPatt, … })`. This differs from the `M.choose(keyName, [{ [keyName]: keyValue1, … }, { [keyName]: keyValue2, … }, …])` suggestion of #3072 in a handful of ways:
* Using a record rather than an array avoids repeating the discriminator key name and inherently avoids non-disjoint sub-patterns.
* Matching is more efficient, checking at most one sub-pattern after finding it by direct lookup.
* Error messages are either directly associated with the literal discriminator value used to select the sub-pattern, or else indicate failure to identify one.
* The sub-patterns theirselves are not required to be records; e.g. `{ keyValueN: M.recordOf(M.string(), M.any() }` is just fine.
* The specimen can only match if it is a CopyRecord, because otherwise it can't have the discriminator field.
* There is no escape hatch to recover looser key matching in which e.g. the discriminator value is itself matched by non-static pattern (but since we don't have e.g. starts-with or regular expression patterns anyway, this seems like not much of a loss).
* ~~`discriminated` more accurately communicates these constraints than `choose`, and if we ever want to relax them, that would be the time for a more generic name.~~ https://github.com/endojs/endo/pull/3073#issuecomment-3862248456

For reference, Zod calls this [`discriminatedUnion`](https://zod.dev/api?id=discriminated-unions) and its API is similar to the original suggestion—defined from a string key name and an array of disjoint object patterns, each of which must define the corresponding key with a value that is either a literal or a string enumeration and does not match any value that would also be matched by another pattern in the array, which I think means it's actually slightly weaker than what we have here (_because our sub-patterns are not limited to object matchers and can be e.g. `M.or(…)`/`M.splitRecord(…)`/etc. for covering non-trivial combinations of fields_).

### Security Considerations

None known.

### Scaling Considerations

As mentioned above, this is faster than `M.or` because it never considers more than one sub-pattern.

Hopefully covered by JSDoc.

### Testing Considerations

Covered by new unit tests.

I don't think matchers can ever expand beyond their initial shape, so if this is accepted then it will always have its current limitations (specifically, `match:choose` matchers cannot grow support for non-static key matching because matchers theirselves are passable and the shape introduced here has no way to express it—but future changes _could_ still update `M.choose(…)` to return instances of a new type).

### Upgrade Considerations

Changes are purely additive, so I think the only possible upgrade issue would be if new code passes a `match:choose` instance to old code that tries to match against it. I'm not aware of any such old code that cannot itself be upgraded.


packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1636,6 +1647,35 @@ const makePatternKit = () => {
...
+      if (typeof keyValue !== 'string' || !hasOwn(patts, keyValue)) {
+        return (
+          reject &&
+          reject`${specimen} - Must have discriminator key ${q(keyName)} with value in ${q(ownKeys(patts))}`
+        );
+      }
+      const label = `{${q(keyName)}: ${q(keyValue)}}`;
+      const { [keyName]: _, ...rest } = specimen;
+      const subPatt = patts[keyValue];
+      return confirmNestedMatches(freeze(rest), subPatt, label, reject);
```
dckc: `rest` is constructed by destructuring. The Jessie [Must freeze API Surface Before Use](https://github.com/endojs/Jessie?tab=readme-ov-file#must-freeze-api-surface-before-use) don't seem to capture this case.

dckc (commented):
the feature looks nice,
but the upgrade/compatibility considerations make my head hurt

packages/patterns/src/patterns/patternMatchers.js
erights: The `specimen` must already be passable and therefore hardened. Given that, only the top level of `rest` is not frozen. Everything deeper is frozen. Therefore, the `freeze` here means the code does uphold the purpose  "Must freeze API Surface Before Use". The lack is on the Jessie side, where we don't have any syntactic machinery to recognize that this code conforms.

packages/patterns/NEWS.md
```diff
@@ -3,6 +3,7 @@ User-visible changes in `@endo/patterns`:
 # Next release
 
 - The `sloppy` option for `@endo/patterns` interface guards is deprecated. Use `defaultGuards` instead.
+- PatternMatchers now includes a `M.discriminated(keyName, subPatternsRecord)` function to match a CopyRecord against a sub-pattern selected by the value of a specified discriminator key (i.e., for matching values of a discriminated union type). Behaviorally, the new matchers are slightly weaker than those from `M.or(...patterns)`, but more efficient, and they produce more precise error messages upon match failure. Note that the sub-patterns apply to a derived CopyRecord that lacks the discriminator property, so e.g. `M.discriminated('flavor', { original: M.and({ flavor: 'original' }, M.any()) })` does not match anything, while `M.discriminated('flavor', { original: M.any() })` matches any CopyRecord with a "flavor" property whose value is "original".
```
erights:
`````suggestion
- PatternMatchers now includes a `M.discriminated(keyName, subPatternsRecord)` function to match a CopyRecord against a sub-pattern selected by the value of a specified discriminator key (i.e., for matching values of a discriminated union type). Behaviorally, the new matchers are slightly weaker than those from `M.or(...patterns)`, but more efficient, and they produce more precise error messages upon match failure. Note that the sub-patterns apply to a derived CopyRecord that lacks the discriminator property, so e.g. 
  ```js
  M.discriminated('flavor', { 
    original: M.and({ flavor: 'original' }, M.any()),
  })
  ```
  does not match anything, while 
  ```js
  M.discriminated('flavor', { 
    original: M.any(), 
  })
  ``` 
  matches any CopyRecord with a "flavor" property whose value is "original".
`````

erights (approved):
> `discriminated` more accurately communicates these constraints than `choose`, and if we ever want to relax them, that would be the time for a more generic name.

I technically agree with the premise but disagree on the conclusion. "discriminated" imposes more pain from its verbosity than it saves from its slightly more precise meaning.

I like "choose". But if we want something else, please something short. Once this exists, if the name is short, it may be used a lot.

The original "choose" suggestion was never implemented, so there's no conflict using "choose".

packages/patterns/src/patterns/patternMatchers.js
dckc: yes, this code conforms.

My comment aims to say: the Jessie rules don't seem to say that you need a `harden(...)` here.

gibson042:
For reference, Zod calls this [`discriminatedUnion`](https://zod.dev/api?id=discriminated-unions) and its API is similar to the original suggestion—defined from a string key name and an array of disjoint object patterns, each of which must define the corresponding key with a value that is either a literal or a string enumeration and does not match any value that would also be matched by another pattern in the array, which I think means it's actually slightly weaker than what we have here (_because our sub-patterns are not limited to object matchers and can be e.g. `M.or(…)` for covering non-trivial combinations of fields_).

dckc:
In https://github.com/endojs/endo-but-for-bots/issues/220 a bot wants @kriskowal to
> confirm @dckc's post-approval comment isn't blocking and click merge.

I'm not sure which comment that refers to, but I'm pretty sure it isn't blocking.

PR #3083, opened 2026-02-12
bundle-source maintenance

turadg:
_incidental_

## Description

There was some confusion recently about the API of `BundleCache`. This adds docs for it.

It also adds tests coverage, fixes up types and some runtime cleanup.

The new tests fail under Node 18 because it defaults to CJS for `.js` files. ~Since that version is past EOL this also removes it from CI matrix. And while searching it unpins `node-version` from jobs which aren't meant to test release compatibility (and should always work with latest default Node version).~ This provides a package.json and defers CI changes to https://github.com/endojs/endo/pull/3084

### Security Considerations

none

### Scaling Considerations

none

This should go out in a release so it needs a changeset. Patch release I suppose.

### Testing Considerations

CI

no, just docs and types and some small bug fixes

### Upgrade Considerations

will address in a changeset file

### DIFF
--- a/.changeset/olive-socks-join.md
+++ b/.changeset/olive-socks-join.md
@@ -1,0 +1,5 @@
+---
+'@endo/bundle-source': patch
+---
+
+Fix bundle cache corner cases, improve cache-root validation, and clarify CLI docs for `endoScript` bundle format.


packages/bundle-source/test/tool-command.test.js
```diff
@@ -43,8 +43,42 @@ const shellOut = () =>
...
+
 test('bundle-source command is concurrency safe', async t => {
   const concurrentJobs = Array.from({ length: 5 }).map(() => shellOut());
   await Promise.all(concurrentJobs);
   t.pass();
 });
+
+test('bundle-source usage includes endoScript format', async t => {
+  const error = await t.throwsAsync(() => shellOutInvalidFormat());
+  t.regex(error.message, /-f,--format .*endoScript/);
```
michaelfig: There's a more succinct way:

```suggestion
  await t.throwsAsync(() => shellOutInvalidFormat(), { message: /-f,--format .*endoScript/ });
```

packages/bundle-source/test/cache.test.js
```diff
@@ -1,18 +1,96 @@
...
+
+  await cache.add(entry, targetName, t.log);
+  await fs.writeFile(path.join(dest, `bundle-${targetName}-js-meta.json`), '{');
+
+  const error = await t.throwsAsync(() =>
+    cache.validateOrAdd(entry, targetName, t.log),
+  );
+  t.true(error instanceof SyntaxError);
+  t.regex(error.message, /Cannot parse JSON from cache-test-malformed-meta/);
+});
```
michaelfig, line 76: More idiomatic:
```suggestion
  await t.throwsAsync(() =>
    cache.validateOrAdd(entry, targetName, t.log),
    {
      instanceOf: SyntaxError,
      message: /Cannot parse JSON from cache-test-malformed-meta/,
    },
  );
```

packages/bundle-source/cache.js
```diff
@@ -365,13 +379,27 @@ export const makeBundleCache = (wr, cwd, readPowers, opts) => {
       return found.bundle;
     }
     const todo = makePromiseKit();
+    // This promise may be rejected before any concurrent caller awaits it.
+    todo.promise.catch(() => {});
```
michaelfig, line 386: I'm curious about the rationale, but this at least should have a comment:
```suggestion
    // This promise may be rejected before any concurrent caller awaits it.
    // HardenedJS only logs a warning if the promise is never awaited.
    // We silence the promise completely because we want to ensure the
    // rejection is never logged, with or without HardenedJS.
    todo.promise.catch(() => {});
```

PR #3087, opened 2026-02-13
feat(daemon): Reaper

kriskowal:
## Description

This change introduces a garbage collector to the daemon.  The user level requirement is that, if a user deletes every pet name in their local user agent that ultimately retains a formula, we must revoke the corresponding live reference.  This is sufficient for purposes of security, but creates an incentive for developers on the platform to create elaborate accommodations for the sudden revocation of live references and the reestablishment of a fresh live reference from "sturdy" paths.  To mitigate this, the daemon instead elects to obligate developers to tolerate a restart if a worker is contaminated with a live reference to a revoked capability.  So, in this change, we also cancel (deincarnate) any value for which the formula has been collected, and any value that transitively dependended upon a collected formula for its construction, and furthermore reincarnate that value if it is still retained from PINS.

**Formula collection** is a deterministic garbage collector for the formula graph. Formulas that become unreachable from the root are cancelled, their persistence files deleted, and any CapTP connections or workers still holding references are disconnected or terminated.

The collector uses a union-find structure to group formulas that share identity (agent+handle, promise+resolver) so they are collected atomically.  A refcount over the group-level dependency graph drives collection.  Pet store writes and removals maintain dynamic edges. Collection runs inside `formulaGraphJobs` and is triggered in the `finally` block of every host and guest command, ensuring the graph is consistent before any command returns.

New modules extracted from `daemon.js`:
- **`graph.js`** — `makeFormulaGraph`: union-find, dependency tracking, root set, dirty flag.
- **`residence.js`** — `makeResidenceTracker`: CapTP export/import hooks that track which connections retain which formulas, and `disconnectRetainersHolding` to sever connections to collected values.

Pet store path construction and deletion were consolidated into `pet-store.js`, removing duplication with `daemon-node-powers.js`.

### Security Considerations

Formula collection introduces a new authority boundary: when a formula is collected, its live value is cancelled and its CapTP connection is severed.  The error message intentionally avoids leaking formula internals, stating only that a formula "became unreachable by any pet name path and was collected." Workers hosting collected values are terminated to prevent use-after-free of stale references.

### Scaling Considerations

Collection is O(formulas + edges) per pass but runs only when the graph is dirty. The `dirty` flag avoids redundant passes. Transient roots prevent premature collection during command execution without holding the formula graph lock.

Not yet addressed. This feature will obligate the user interface to provide a tutorial on revocation, for which is not yet sufficient support to write.  A user will need to be able to inspect local retention paths to a petnamed capability.

### Testing Considerations

Nine new tests in `packages/daemon/test/endo.test.js`:
- Collects formulas after pet name removal
- Terminates worker retaining collected values
- Terminates worker retaining derived value after dependency collection
- Recreates counter after collection resets state
- PINS values survive collection
- PINS values reincarnate after cancellation
- Facet group (agent + handle) collects atomically
- Unnamed eval results are collected
- Reply and threading tests (reply links to parent, metadata naming)

No breaking changes. The collector is always active — there is no feature flag. Existing formulas that are reachable from roots are never collected.  The `endo reply` command is additive.

### Upgrade Considerations

No migration needed. Existing daemon state is seeded into the formula graph on boot. Unreachable formulas left over from prior sessions will be collected on the first command after upgrade.

### DIFF
--- a/packages/daemon/src/daemon.js
+++ b/packages/daemon/src/daemon.js
@@ -2131,7 +2131,6 @@ const makeDaemonCore = async (
     // Wait for any in-flight graph operation (formulation, collection)
     // to finish before cancelling.
     await formulaGraphJobs.enqueue();
-    const formula = formulaForId.get(id);
     const controller = provideController(id);
     console.log('Cancelled:');
     return controller.context.cancel(reason);
@@ -2140,7 +2139,7 @@ const makeDaemonCore = async (
   /** @type {DaemonCore['formulateReadableBlob']} */
   const formulateReadableBlob = async (readerRef, deferredTasks) => {
     return /** @type {FormulateResult<FarRef<EndoReadable>>} */ (
-      await withFormulaGraphLock(async () => {
+      withFormulaGraphLock(async () => {
         await null;
         const formulaNumber = /** @type {FormulaNumber} */ (
           await randomHex512()
@@ -2180,7 +2179,7 @@ const makeDaemonCore = async (
     deferredTasks,
   ) => {
     return /** @type {FormulateResult<Invitation>} */ (
-      await withFormulaGraphLock(async () => {
+      withFormulaGraphLock(async () => {
         const invitationNumber = /** @type {FormulaNumber} */ (
           await randomHex512()
         );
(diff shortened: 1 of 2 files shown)


packages/daemon/src/residence.js
```diff
@@ -0,0 +1,174 @@
...
+ * @param {(ref: unknown) => FormulaIdentifier | undefined} args.getLocalIdForRef
+ * @param {(id: FormulaIdentifier) => Formula | undefined} args.getFormula
+ * @param {(workerId: string, reason: Error) => void} args.terminateWorker
+ */
+export const makeResidenceTracker = ({
+  getLocalIdForRef,
+  getFormula,
+  terminateWorker,
+}) => {
+  let nextRetainerId = 0;
```
rekmarks: I realize we only use this in one place, but I'd replace this with a `getNextRetainerId()` that encapsulates incrementing the id.

PR #3099, opened 2026-02-19
perf(bundle-source): cut multi-entry agoric bundling time and add detailed profiling

turadg:
Refs: (none)

## Description

This PR speeds up `@endo/bundle-source` for multi-entrypoint workloads (notably `agoric-sdk`) and improves profiling visibility for further tuning.

Primary changes:
- Add richer end-to-end profiling and trace merge summaries:
  - `packages/bundle-source/tools/profile-agoric-bundling.mts`
  - `packages/bundle-source/tools/trace-merge.js`
- Add import/parse/transform instrumentation in compartment-mapper and evasive-transform.
- Optimize node-modules graph finalization path (`finalizeGraph`) and canonical/read behavior:
  - `packages/compartment-mapper/src/generic-graph.js`
  - `packages/compartment-mapper/src/node-modules.js`
  - `packages/compartment-mapper/src/node-powers.js`
  - `packages/compartment-mapper/src/import-hook.js`
  - `packages/compartment-mapper/src/map-parser.js`
- Add cross-bundle parser reuse in archive parsing:
  - `packages/compartment-mapper/src/parse-archive-mjs.js`
- Improve archive/zip path and add zip microbenchmark tooling:
  - `packages/compartment-mapper/src/archive-lite.js`
  - `packages/zip/src/writer.js`
  - `packages/zip/src/format-writer.js`
  - `packages/zip/tools/benchmark-writer.mjs`
- Add/read cache behavior in bundle-source zip bundling path:
  - `packages/bundle-source/src/zip-base64.js`

Measured result from `yarn workspace @endo/bundle-source profile:agoric-bundling`:
- Baseline: `/var/folders/v2/309w97f90q5d0r47d3np62v40000gn/T/profile-agoric-bundling-2026-02-19T21-21-58-060Z`
  - `bundleSource.total`: `13180.267ms`
- Latest: `/var/folders/v2/309w97f90q5d0r47d3np62v40000gn/T/profile-agoric-bundling-2026-02-19T23-00-54-856Z`
  - `bundleSource.total`: `7538.092ms`

Total win: **-42.8%** (`-5642.175ms`)

Optimization contribution checkpoints:
- Node-modules/read-path optimizations (`21:21:58` -> `21:33:22`):
  - `13180.267ms -> 10021.350ms` (**-23.97%**)
- Zip path tuning and profiling (`21:33:22` -> `21:53:12`):
  - `10021.350ms -> 9884.006ms` (**-1.37%**)
- Cross-bundle parse reuse (`21:53:12` -> `23:00:54`):
  - `9884.006ms -> 7538.092ms` (**-23.74%**)
  - cache stats (latest): `3887` hits / `357` misses (~91.6% hit rate)

### Security Considerations

No new external runtime dependency is introduced.

Changes are internal performance/instrumentation improvements in bundling and zip writing code paths. New behavior is primarily caching and profiling spans. Cache scope remains process-local and only stores parser output for identical transformed source/module URL combinations.

### Scaling Considerations

This PR is intended to reduce CPU and wall-clock time for repeated bundling of overlapping dependency graphs.

Tradeoffs:
- Added in-process caches increase transient memory usage.
- Cache in `parse-archive-mjs` is bounded by an entry cap and resets when exceeding the cap.

Net effect on profiled workload is significantly lower bundling latency.

No end-user API change.

Operationally useful additions:
- New `@endo/zip` benchmark script: `yarn workspace @endo/zip bench:writer`
- Enhanced profiling summaries for bundle-source tooling.

No migration instructions needed.

### Testing Considerations

Executed tests include:
- `yarn workspace @endo/zip test`
- `yarn workspace @endo/compartment-mapper test test/map-node-modules.test.js`
- `yarn workspace @endo/compartment-mapper test test/node-powers.test.js`
- `yarn workspace @endo/compartment-mapper test test/parse-archive-mjs.test.js`
- `yarn workspace @endo/compartment-mapper test test/bundle.test.js` (known expected failure remains)
- `yarn workspace @endo/bundle-source test test/profiling.test.js`
- `yarn workspace @endo/bundle-source test test/comment.test.js`
- Multiple runs of `yarn workspace @endo/bundle-source profile:agoric-bundling`

No intentional breaking changes.

Behavior should remain compatible; changes are performance-oriented and profiling-oriented with conservative cache bypass conditions where side effects may exist.

### Upgrade Considerations

No special upgrade action required.

No `*BREAKING*` change in commit message and no `NEWS.md` entry added, since this is internal perf and tooling behavior.


packages/bundle-source/src/profile.js
```diff
@@ -0,0 +1,145 @@
...
+/**
+ * @param {string | undefined} value
+ * @returns {boolean}
+ */
+const parseBoolean = value => {
+  if (value === undefined) {
+    return false;
+  }
+  return truthy.has(value.toLowerCase());
+};
```
gibson042: Suggestion:
```suggestion
const parseBoolean = value => truthy.has(value?.toLowerCase());
```

packages/bundle-source/src/profile.js
```diff
@@ -0,0 +1,145 @@
...
+  moduleFormat,
+  startFilename,
+  pid = process.pid,
+  env = process.env,
+  profile = undefined,
+}) => {
+  const enabled =
+    profile?.enabled !== undefined
+      ? profile.enabled
+      : parseBoolean(env.ENDO_BUNDLE_SOURCE_PROFILE);
```
gibson042: Suggestion:
```suggestion
  const enabled =
    profile?.enabled ?? parseBoolean(env.ENDO_BUNDLE_SOURCE_PROFILE);
```

packages/bundle-source/src/profile.js
```diff
@@ -0,0 +1,145 @@
...
+  const logToStderr = parseBoolean(env.ENDO_BUNDLE_SOURCE_PROFILE_STDERR);
+
+  if (!enabled) {
+    const noop = () => {};
```
gibson042: Is there any reason not to define `noop` at top level?

packages/bundle-source/src/profile.js
```diff
@@ -0,0 +1,145 @@
...
+
+  /** @type {Array<Record<string, unknown>>} */
+  const traceEvents = [];
+  const zeroMs = performance.now();
+
+  /**
+   * @param {number} ms
+   * @returns {number}
+   */
+  const toMicros = ms => Math.round(ms * 1000);
```
gibson042: This also seems like a top-level function.

packages/bundle-source/src/profile.js
```diff
@@ -0,0 +1,145 @@
...
+            }
+          : {
+              name,
+              cat: 'bundle-source',
+              ph: 'X',
+              ts: toMicros(startMs),
+              dur: toMicros(endMs - startMs),
+              pid,
+              tid: 0,
+            },
```
gibson042, line 125: It looks like this is trying to conform with an external specification, in which case that should be linked to and ideally also given specific typing. But if not, let's strive for comprehensibility instead of super-abbreviated field names. Or if raw size is the predominant concern, then we'll want something array-based (cf. SenML).

packages/bundle-source/src/zip-base64.js
```diff
@@ -13,10 +13,54 @@ import { encodeBase64 } from '@endo/base64';
...
+const cachedReads = new Map();
+/** @type {Map<string, Promise<Uint8Array | undefined>>} */
+const pendingReads = new Map();
+let cachedReadBytes = 0;
+
+/**
+ * @param {string} location
+ * @param {Uint8Array | undefined} bytes
+ */
+const cacheReadValue = (location, bytes) => {
```
gibson042: This function is right on the threshold, but I think probably sufficiently non-obvious to necessitate a description.

packages/bundle-source/src/zip-base64.js
```diff
@@ -13,10 +13,54 @@ import { encodeBase64 } from '@endo/base64';
...
+  if (prior !== undefined) {
+    cachedReadBytes -= prior.length;
+  }
+
+  cachedReads.set(location, bytes);
+  if (bytes !== undefined) {
+    cachedReadBytes += bytes.length;
+  }
+
+  while (cachedReadBytes > readCacheMaxBytes && cachedReads.size > 0) {
```
gibson042: Given the nature and management of `cachedReadBytes`, this guard on `cachedReads.size` seems pointless.
```suggestion
  while (cachedReadBytes > readCacheMaxBytes) {
```

packages/bundle-source/src/zip-base64.js
```diff
@@ -13,10 +13,54 @@ import { encodeBase64 } from '@endo/base64';
...
+    const oldestKey = cachedReads.keys().next().value;
+    if (oldestKey === undefined) {
+      break;
+    }
+    const value = cachedReads.get(oldestKey);
+    cachedReads.delete(oldestKey);
+    if (value !== undefined) {
+      cachedReadBytes -= value.length;
+    }
+  }
```
gibson042: Simpler:
```suggestion
  if (cachedReadBytes <= readCacheMaxBytes) return;

  for (const [oldestKey, value] of cachedReads) {
    cachedReads.delete(oldestKey);
    if (value !== undefined) cachedReadBytes -= value.length;
    if (cachedReadBytes <= readCacheMaxBytes) return;
  }
```

packages/bundle-source/src/zip-base64.js
```diff
@@ -50,57 +95,166 @@ export async function bundleZipBase64(
...
-    {
-      parserForLanguage,
-      moduleTransforms,
+  let phaseStatus = 'ok';
+  let phaseError;
+  try {
+    const maybeRead = async location => {
+      if (readCacheMaxBytes === 0) {
+        return powers.maybeRead(location);
+      }
```
gibson042: This bypass of `pendingReads` looks inappropriate.
```suggestion
```

packages/bundle-source/tools/trace-merge.js
gibson042, line 1: There's so much going on in these tools/ files and so little documentation; I'm skipping them in this pass.

packages/bundle-source/README.md
```diff
@@ -119,6 +119,64 @@ map for every physical module.
 It is not yet quite clever enough to collect source maps for sources that do
 not exist.
 
+## Profiling
```
gibson042, line 123: 👍

packages/bundle-source/README.md
```diff
@@ -119,6 +119,64 @@ map for every physical module.
...
+Environment variables:
+- `ENDO_BUNDLE_SOURCE_PROFILE`: enable profiling when truthy (`1`, `true`, `yes`, `on`)
+- `ENDO_BUNDLE_SOURCE_PROFILE_DIR`: output directory for generated trace files
+- `ENDO_BUNDLE_SOURCE_PROFILE_FILE`: explicit output file for a single run
+- `ENDO_BUNDLE_SOURCE_PROFILE_STDERR`: if truthy, prints each generated trace path to stderr
+
+Merge and summarize many profile traces:
+
+```console
+yarn workspace @endo/bundle-source trace:merge -- /tmp/bs-profiles
```
gibson042, line 161: Is this actually specific to bundle-source, or e.g. sufficiently general for arbitrary trace composition?

packages/bundle-source/README.md
```diff
@@ -119,6 +119,64 @@ map for every physical module.
...
+This generates:
+- `merged.trace.json` for trace viewers.
+- `summary.json` with aggregate span statistics.
+- `summary.md` with a top spans table by total duration.
+
+Profile bundling all `source-spec-registry.js` entries from an `agoric-sdk`
+checkout using the current checkout's `bundle-source`:
+
+```console
+yarn workspace @endo/bundle-source profile:agoric-bundling -- \
```
gibson042, line 175: Same question about this tool.

packages/compartment-mapper/src/archive-lite.js
```diff
@@ -70,22 +70,54 @@ const { assign, create, freeze, keys } = Object;
...
+            maybeWrite &&
+            typeof maybeWrite === 'object' &&
+            'then' in maybeWrite &&
+            typeof maybeWrite.then === 'function'
+          ) {
+            asyncWriteCount += 1;
+            // Preserve deterministic write order for truly async writers.
+            // eslint-disable-next-line no-await-in-loop
+            await maybeWrite;
+          }
```
gibson042: This is a pretty big increase in complexity to save some `await`s and track a presumably uninteresting count. Is it actually worthwhile?

packages/compartment-mapper/src/generic-graph.js
```diff
@@ -335,6 +335,55 @@ const dijkstra = (graph, source, target) => {
...
+  assert(
+    distances.get(source) === Infinity,
+    `Source ${q(source)} is not in the graph`,
+  );
+
+  distances.set(source, 0);
+
+  for (const node of nodes) {
+    queue.add(node);
+  }
```
gibson042, line 276:
```suggestion
  const context = {
    // XXX once we raise the Node.js bar: `nodes.values().map(...)`
    distances: new Map([...nodes].map(node => [node, Infinity])),
    predecessors: new Map(),
    queue: new Set(nodes),
  };
  const { queue, distances } = context;

  distances.set(source, 0);
```

packages/compartment-mapper/src/import-hook.js
```diff
@@ -579,9 +641,33 @@ export const makeImportHookMaker = (
...
+    profileStartSpan = undefined,
     log = noop,
   },
 ) => {
+  const { maybeRead } = unpackReadPowers(readPowers);
+  /** @type {Map<string, Promise<Uint8Array|undefined>>} */
+  const maybeReadCache = new Map();
+  /**
+   * Cache both hits and misses for module reads during a mapping run.
+   * This avoids repeated filesystem probes for the same candidate path.
```
gibson042: This seems like a good idea, but the function body doesn't implement it AFAICT (on error from `maybeRead`, it deletes the entry from `maybeReadCache`).

packages/compartment-mapper/src/node-powers.js
```diff
@@ -69,40 +69,61 @@ const fakeIsAbsolute = () => false;
...
  * @param {PathInterface} [args.path]
+ * @param {number} [args.maxConcurrentReads]
  * @returns {MaybeReadPowers<FileUrlString>}
  */
 const makeReadPowersSloppy = ({
   fs,
   url = undefined,
   crypto = undefined,
   path = undefined,
+  maxConcurrentReads = 32,
```
gibson042, line 81: 👍

packages/compartment-mapper/src/parse-archive-mjs.js
```diff
@@ -20,19 +24,68 @@ export const parseArchiveMjs = (
...
+  const { sourceMap, sourceMapHook, profileStartSpan } = options;
+  const canUseCache = sourceMapHook === undefined;
   const source = textDecoder.decode(bytes);
+  const sourceMapKey =
+    sourceMap === undefined
+      ? ''
+      : typeof sourceMap === 'string'
+        ? sourceMap
+        : JSON.stringify(sourceMap);
+  const cacheKey = `${source}\n//# sourceMappingURL=${sourceMapKey}`;
```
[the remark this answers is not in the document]
gibson042: It does seem like a very unwieldy cache key; can we somehow do better?

packages/compartment-mapper/src/parse-archive-mjs.js
```diff
@@ -20,19 +24,68 @@ export const parseArchiveMjs = (
...
+      parseArchiveMjsCache.set(sourceUrl, byLocation);
+    }
+    if (!byLocation.has(cacheKey)) {
+      parseArchiveMjsCacheEntries += 1;
+      if (parseArchiveMjsCacheEntries > MAX_PARSE_ARCHIVE_MJS_CACHE_ENTRIES) {
+        parseArchiveMjsCache.clear();
+        parseArchiveMjsCacheEntries = 0;
+        byLocation = new Map();
+        parseArchiveMjsCache.set(sourceUrl, byLocation);
+      }
```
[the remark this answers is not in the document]
gibson042: I agree with copilot, must we clear the _entire_ cache rather than e.g. evicting all entries under the oldest key?
```suggestion
      parseArchiveMjsCacheEntries += 1;
      if (parseArchiveMjsCacheEntries > MAX_PARSE_ARCHIVE_MJS_CACHE_ENTRIES) {
        // Find the oldest key in parseArchiveMjsCache and evict all of its
        // sub-entries, either by deleting its entry or (if we need it here)
        // by clearing is associated map.
        const oldest = parseArchiveMjsCache.entries().next().value;
        const oldestByLocation = oldest?.[1];
        const size = oldestByLocation?.size;
        assert(size > 0);
        if (byLocation === oldestByLocation) {
          byLocation.clear();
        } else {
          parseArchiveMjsCache.delete(oldest[0]);
        }
        parseArchiveMjsCacheEntries -= size;
      }
```

packages/evasive-transform/src/index.js
```diff
@@ -12,6 +12,54 @@ import { transformAst } from './transform-ast.js';
...
+    return true;
+  }
+  // Fast path: if none of the risky comment payload tokens appear anywhere in
+  // the source, the transform cannot change semantics-relevant content.
+  return (
+    source.includes('import(') ||
+    source.includes('<!--') ||
+    source.includes('-->')
+  );
+};
```
gibson042, line 30: This seems very likely to conflict with #3026; if changes to evasive-transform are on the table then I think this PR should be blocked on that one and can therefore be rebased on top of it. For example, this triple-scan would probably be faster if it leveraged the introduction of `evadeRegexp`:
```suggestion
/**
 * Efficiently classify source text as might-need-transformation vs.
 * definitely-does-not, to avoid the cost of parsing/traversal/transformation/
 * reserialization whenever possible. The former category includes all source
 * text when `elideComments` is true, but otherwise includes only source text
 * that contains a tranformable substring.
 *
 * @param {string} source
 * @param {boolean} elideComments
 * @returns {boolean}
 */
const shouldRunTransform = (source, elideComments) =>
  elideComments || source.search(evadeRegexp) !== -1;
```

packages/compartment-mapper/src/parse-archive-mjs.js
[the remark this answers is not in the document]
turadg: Addressed in the fixup commits. I replaced the concatenated cache key with a structured nested cache keyed as `sourceUrl -> source -> sourceMapKey`, so we no longer build a synthetic `${source}\n//# sourceMappingURL=...` string or depend on a delimiter that could collide with source contents.

This should handle the collision concern and make the key structure more explicit. It still uses the full source text as the source-level key; if you’d prefer reducing key size too, I can switch that level to a content hash.

PR #3100, opened 2026-02-21
feat(daemon): HTTP and WebSocket gateway

kriskowal:
## Description

This sets the stage for Endo Familiar, an Electron app that manages the daemon and serves local HTTP web applications through a custom `localhttp://` protocol handler. The unified gateway eliminates the need for each weblet to claim a dedicated port, allowing Electron to route requests to weblets by hostname instead.

Adds a unified HTTP/WebSocket gateway server to the Endo daemon, listening on `ENDO_ADDR` (default `127.0.0.1:8920`). Weblets can register on the shared server using hostname-based isolation or opt into a dedicated port with path-based isolation. Gateway clients connect over WebSocket and call `fetch(token)` to retrieve formula values via CapTP.

Extracts `trackConnection`, `openCapTPSession`, and `makeWebletFar` helpers to deduplicate the gateway and weblet connection paths. Installs `@types/ws` to replace `@ts-ignore` directives.

### Security Considerations

Only loopback connections (`127.0.0.1`, `::1`, `::ffff:127.0.0.1`) are accepted. Non-local connections are rejected with a 1008 close code. This restriction will be revisited when CapTP is replaced with OCapN over Noise Protocol.

### Scaling Considerations

None. One HTTP server replaces per-weblet servers when no dedicated port is requested.

The daemon README now documents the `ENDO_ADDR` environment variable and its usage with `endo start`.

### Testing Considerations

New `gateway.test.js` covers: gateway info page, WebSocket `fetch(token)`, weblet on the unified server, and weblet on a dedicated port.

`getPort()` is renamed to `getAddress()` and now returns a URL string instead of a port number. Callers must be updated.

### Upgrade Considerations

None.

### DIFF
--- a/packages/daemon/src/web-server-node.js
+++ b/packages/daemon/src/web-server-node.js
@@ -3,16 +3,17 @@ import fs from 'node:fs';
 import http from 'node:http';
 import { fileURLToPath } from 'url';
 
-/** @import { HttpRespond, FarContext } from './types.js' */
+/** @import { HttpRespond, HttpConnect, FarContext, EndoBootstrap } from './types.js' */
 
-// @ts-ignore We cannot use a synthetic default export in practice here (circa Node.js 16)
 import * as ws from 'ws';
 
-import { mapWriter, mapReader } from '@endo/stream';
+import { makePromiseKit } from '@endo/promise-kit';
+import { makePipe, mapWriter, mapReader } from '@endo/stream';
 import { E, Far } from '@endo/far';
 import { makeBundle } from '@endo/compartment-mapper/bundle.js';
+import { makeExo } from '@endo/exo';
+import { M } from '@endo/patterns';
 
-import { q } from '@endo/errors';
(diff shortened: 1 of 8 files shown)


packages/daemon/src/web-server-node.js
```diff
@@ -150,45 +448,65 @@ export const make = async (_powers, context) => {
...
   };
 
-  return Far('WebletService', { makeWeblet });
+  return Far('WebletService', {
+    makeWeblet,
+    async getAddress() {
+      const addr = await started;
+      // NOTE this was prior behavior, but we may want to consider passing `actual.address` out, since config might be `0.0.0.0`?
+      const actual = `http://${addr.config.address}:${addr.actual.port}`;
+      return actual;
```
jcorbin, line 512: @kriskowal you attention here please, as this is a new thing I introduced in my review update ; also rhymes with the usage site where we log-listening above

PR #3102, opened 2026-02-26
chore(ci): create custom CHANGELOG generator

boneskull:
This is a custom `CHANGELOG.md` generator for Endo.

It mimics the behavior of [conventional-changelog-conventionalcommits](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-conventionalcommits) and is based on [@changesets/changelog-github](https://github.com/changesets/changesets/tree/main/packages/changelog-github). It differs from the latter in two ways:

  - The version header is a link to a "diff" of the changes between the previous and current version
  - The version header contains a date in the format `YYYY-MM-DD`

_Note_: This generator does **not** run when the "Version Packages" PR is merged; as such, the date shown may not be the date of the release itself.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -4,7 +4,7 @@
   "baseBranch": "master",
   "bumpVersionsWithWorkspaceProtocolOnly": true,
   "changelog": [
-    "@changesets/changelog-github",
+    "./endo-changelog-generator.js",
     {
       "repo": "endojs/endo"
     }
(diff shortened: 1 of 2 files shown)


boneskull:
@kriskowal my 2c:

I _do_ like the links to the diff, links to the commits, links to the PRs, etc. As someone who often engages in release archaeology, these are helpful. Attribution is nice, but I have no strong opinion about it. 

You explained you like the format of `NEWS.md` which is quite minimal in comparison.

Can we come to a compromise between these two?

boneskull:
Whatever we want to do here will need to be built on top of the _other_ custom changelog generator as added in #3330

PR #3105, opened 2026-02-27
feat(ses):  add `toTemporalInstant` method to DatePrototype permits

jordan-enev:
## Description

Add `toTemporalInstant` method to DatePrototype permits. This method has been included in the latest browser APIs.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toTemporalInstant
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal

### Security Considerations

> No

### Scaling Considerations

> No

> No

### Testing Considerations

> No

> No

### Upgrade Considerations

> No

### DIFF
--- a/packages/ses/src/permits.js
+++ b/packages/ses/src/permits.js
@@ -847,6 +847,7 @@ export const permitted = {
     toLocaleString: fn,
     toLocaleTimeString: fn,
     toString: fn,
+    toTemporalInstant: fn,
     toTimeString: fn,
     toUTCString: fn,
     valueOf: fn,


erights (changes requested):
Hi @jordan-enev , thanks for the PR. We just discussed it internally and decided we do not want to add `toTemporalInstant` to permits.js before we add the rest of safe Temporal(*). We do expect to add all of safe Temporal, but it'll take attention we don't currently have at the moment.

(*) We hope and expect that "safe Temporal" is "all of Temporal except Temporal.Now".

I'm requesting changes rather than closing in case there's more to discuss. If not, feel free to either close this, or make it a PR that also adds all of allegedly safe Temporal to permits.js. Then, this PR would become the natural place to discuss what safe Temporal is, and to approve it. Just be prepared for that approval to be delayed until we have (ahem) time.

Thanks!

PR #3107, opened 2026-02-28
refactor(memoize): depend on harden, not ses

erights:
Closes: #XXXX
Refs: #3108

## Description

`@endo/memoize` is a tiny package that had depended on `'ses'` only for `harden`. It is a great example of why we factored `harden` out into an importable `@endo/harden`. The PR changes `@endo/memoize` to depend directly on `endo@harden` rather than `ses`.

### Security Considerations

@endo/harden enables non-protecting hardens, but only when protection is off in general. @endo/memoize only provides its strong guarantees when @endo/harden is fully protecting. This is all consistent, so I would say:

No weakening of security.

### Scaling Considerations

A bit. It means other packages can depend on @endo/memoize without bringing in all of ses.

none

### Testing Considerations

- [x] Should do the multi-ses-ava mode testing that @kriskowal set up.
  Addendum: happy to find that it already happens without my needing to do anything. Thanks @kriskowal !

none

### Upgrade Considerations

none

### DIFF
--- a/packages/memoize/index.js
+++ b/packages/memoize/index.js
@@ -1,3 +1,1 @@
-/// <reference types="ses"/>
-
 export * from './src/memoize.js';


turadg (changes requested):
This isn't valid as long as we have:
https://github.com/endojs/endo/blob/d1d96256f47c5209dfce3f3d52d3f222f266121a/packages/memoize/index.js#L1

That line makes "ses" a dependency of the package, so that any consumer can resolve its types. If you can remove the line above and everything passes, then I can approve.

PR #3108, opened 2026-02-28
refactor(promise-kit): depend on harden, not ses

erights:
Closes: #XXXX
Refs: #3107

## Description

`@endo/promise-kit` is a tiny package that had depended on `'ses'` only for `harden`. It is a great example of why we factored `harden` out into an importable `@endo/harden`. The PR changes `@endo/promise-kit` to depend directly on `endo@harden` rather than `ses`.

### Security Considerations

none

### Scaling Considerations

A bit. It means other packages can depend on @endo/promise-kit without bringing in all of ses.

none

### Testing Considerations

- [x] Should do the multi-ses-ava mode testing that @kriskowal set up.
  Addendum: happy to find that it already happens without my needing to do anything. Thanks @kriskowal !

none

### Upgrade Considerations

none

### DIFF
--- a/.changeset/fancy-pots-enjoy.md
+++ b/.changeset/fancy-pots-enjoy.md
@@ -1,0 +1,5 @@
+---
+'@endo/promise-kit': patch
+---
+
+`@endo/promise-kit` no longer depends on `ses`, just `@endo/harden`


turadg (approved):
Worth merging but I think a more comprehensive PR is in order: https://github.com/endojs/endo/pull/3127

PR #3110, opened 2026-02-28
refactor(error-console-internal): for use only by ses and @endo/errors

erights:
Staged on #3107

Closes: #XXXX
Refs: #XXXX

## Description

We have wanted to unbundle assert and console from ses, so, for example, others could use @endo/errors without pulling in ses. But untangling assert from console would take more work than I'm willing to spend right now. The ses/src/errors directory contains both assert and console entangled. Fortunately, I took a look at all imports in this directory, and I saw very little dependence on the rest of ses. So this PR will 
- [x] start by moving ses/src/error to error-console-internal/src.
- [x] refactor ses to depend on error-console-internal instead of its ses/src/error directory.
- [x] refactor @endo/errors to depend error-console-internal instead of ses

The package name uses "internal" because it is not meant to be imported by anything other than ses and @endo/errors. 

- [x] ~Ideally @endo/error-console-internal would not be published to npm, though I will need help with that.~
  Addendum: https://github.com/endojs/endo/pull/3110#issuecomment-3982096134 convinced me to give up on that.

### Security Considerations

none that I know of yet, but I'll keep my eyes open.

### Scaling Considerations

The point. It would allow others to import from @endo/errors without pulling in all of ses.

none

### Testing Considerations

My original plan was also to move ses/test/error to error-console-internal/test, but that is some trouble and adds no value. I'm leaving that where it is.

none that I know of yet, but I'll keep my eyes open.

### Upgrade Considerations

none that I know of yet, but I'll keep my eyes open.


packages/check-bundle/lite.js
```diff
@@ -70,7 +70,7 @@ export const checkBundle = async (
     assert.typeof(
       endoZipBase64Sha512,
       'string',
-      X`checkBundle cannot bundle without the property 'endoZipBase64Sha512', which must be a string, got ${typeof endoZipBase64Sha512}`,
+      X`checkBundle cannot bundle without the property 'endoZipBase64Sha512', which must be a string, got ${q(typeof endoZipBase64Sha512)}`,
```
erights, line 73: This seems correct. Why did it work before unquoted?

erights: Because @endo/errors is no longer pulling in ses, which changes the default redaction policy ?

Attn @kriskowal 

In any case, we're getting more errors from this same problem. I'll leave them be until we understand these redaction defaults better.

kriskowal: The mostly likely cause for a change is that assertions are not redacted unless and until lockdown gets called.

packages/error-console-internal/src/assert.js
```diff
@@ -192,14 +192,13 @@ freeze(redactedDetails);
  *
  * @type {AssertionUtilities['details']}
```
boneskull, line 193:
```suggestion
 * @type {Details}
```

packages/error-console-internal/src/assert.js
```diff
@@ -48,7 +48,7 @@ import './internal-types.js';
 import { makeNoteLogArgsArrayKit } from './note-log-args.js';
 
 /**
- * @import {BaseAssert, Assert, AssertionFunctions, AssertionUtilities, DeprecatedAssertionUtilities, Stringable, DetailsToken, MakeAssert} from '../../types.js';
+ * @import {BaseAssert, Assert, AssertionFunctions, AssertionUtilities, DeprecatedAssertionUtilities, Stringable, DetailsToken, MakeAssert} from '../types.js';
```
boneskull, line 51:
```suggestion
 * @import {BaseAssert, Assert, AssertionFunctions, AssertionUtilities, DeprecatedAssertionUtilities, Stringable, DetailsToken, MakeAssert, Details} from '../types.js';
```

boneskull (approved):
This must be the reason for adding `@endo/errors` to `@endo/compartment-mapper`.

Looks like mostly just moving things around.

Needs proper `README.md`; otherwise LGTM.

PR #3115, opened 2026-03-05
fix(compartment-mapper): remove unused type from ModuleSourceHookModuleSource

boneskull:
- Removes unused `ModuleSourceHook`'s "error" module source type.
- Adds tests for specific behavior around exit modules
- Fix a JSDoc import

### DIFF
--- a/packages/compartment-mapper/src/import-hook.js
+++ b/packages/compartment-mapper/src/import-hook.js
@@ -41,7 +41,8 @@
  *   CompartmentModuleConfiguration,
  *   LogOptions,
  *   CanonicalName,
- *   LocalModuleSource
+ *   LocalModuleSource,
+ *   ModuleSourceHook,
  * } from './types.js'
  */
 
@@ -262,7 +263,7 @@ const nominateCandidates = (moduleSpecifier, searchSuffixes) => {
  *
  * Preprocesses the fields for the hook.
  *
- * @param {import('./types.js').ModuleSourceHook | undefined} moduleSourceHook Hook function
+ * @param {ModuleSourceHook | undefined} moduleSourceHook Hook function
  * @param {LocalModuleSource} moduleSource Original `LocalModuleSource` object
  * @param {CanonicalName} canonicalName Canonical name of the compartment/package
  * @param {LogOptions} options Options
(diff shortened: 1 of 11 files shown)


packages/compartment-mapper/src/types/external.ts
```diff
@@ -98,24 +98,40 @@ export type PackageDependenciesHook = (params: {
...
-  | {
-      location: FileUrlString;
-      language: Language;
-      bytes: Uint8Array;
-      imports?: string[] | undefined;
-      exports?: string[] | undefined;
-      reexports?: string[] | undefined;
-      sha512?: string | undefined;
-    }
-  | { error: string }
```
kriskowal, line 110: Error seems to be missing. Is that deliberate?

naugtur: that's the unused one. AFAIR, it represents a deferred error and it is not being exposed via the hook (because it's not actually a module but an explicit representation of a lack of one in case someone attempts to require it)

packages/compartment-mapper/src/import-hook.js
```diff
@@ -663,7 +664,7 @@ export const makeImportHookMaker = (
                   moduleSource: {
                     exit: moduleSpecifier,
                   },
-                  canonicalName: compartmentDescriptor.label,
+                  canonicalName: moduleSpecifier,
```
naugtur: I'm not sure this is correct.

My understanding of what the facts are  (please correct me if I'm wrong)
- this is only reachable for exit modules
- endo has a less specific meaning for exit modules - they're modules that were not provided in compartment map. In lavamoat they should be limited to node builtins and native modules
- the compartmentDescriptor here is the parent compartment in which the exit module was requested
- moduleSpecifier happens to be compatible with canonicalName but formally is not one 

With that, I'd rather have canonicalName be undefined for exit modules or remain the parent label, so that values passed to the hook as canonicalName always have a matching compartment.

packages/compartment-mapper/src/types/external.ts
boneskull: (yes, it is deliberate)

packages/compartment-mapper/src/import-hook.js
boneskull: I had considered `undefined` but this breaks the API in a way I didn't want to deal with.  

FWIW, if `canonicalName` is the "parent" `canonicalName` (i.e. I revert this change), then we're just right back where we started (sans type changes).  I am fine with that.

PR #3121, opened 2026-03-08
docs: Design for daemon formula persistence

kriskowal:
This change introduces a design rationale for formula persistence.

### DIFF
--- a/designs/daemon-persistence.md
+++ b/designs/daemon-persistence.md
@@ -1,0 +1,516 @@
+# Formula Persistence (Pass by Construction)
+
+## Abstract
+
+Formula Persistence is a persistence strategy for the Endo Daemon's petname
+database and object capability system. It occupies a distinctive position in the
+design space between orthogonal persistence with masked partition and revival (as
+in the Waterken server) and exposed partition and revival at every reference (as
+in Mark Miller's E language). Rather than persisting live heap state or requiring programs to
+defensively recover broken reference chains, Formula Persistence records *how to
+reconstruct* a capability from a graph of formulas, discarding and rebuilding
+entire cohorts of live references when partition occurs.
+
+## The Problem: Persistence in Distributed Capability Systems
+
+### Two entangled dimensions
+
+Distributed capability systems must choose along two entangled dimensions: how
+to present partition and revival to application programs, and how to persist
+state across restarts and upgrades. These dimensions are entangled — choices
+about partition visibility constrain persistence strategy and vice versa — but
+they are not the same axis. Formula Persistence is distinctive in making a
+different choice on each.
+
+### Masked partition and revival (Waterken model)
+
+At one end of the spectrum, the system masks partition and revival. All
+loss of connectivity is treated as temporary. The program never observes a broken
+reference; it simply waits. Waterken achieves this by combining Joe-E (a
+capability-safe subset of Java) as its local object-capability language with its
+own live crypto-cap protocol in the CapTP family.
+
+This enables orthogonal persistence: the system can snapshot the heap and restart
+from that snapshot, possibly on a different host, without the program's
+knowledge. Because the program never distinguishes between "slow" and
+"partitioned," the system can be made deterministic over all communicating
+programs.
+
+**Advantages:**
+- Simpler programming model: no defensive code against partition
+- Determinism across communicating programs
+- Transparent relocation
+
+**Disadvantages:**
+- Sacrifices availability: a single partitioned dependency stalls all dependents
+- Entangled distributed heaps require distributed garbage collection
+- Differences in incentives among participants necessitate market-based
+  approaches to garbage collection (see "The market-sweep algorithms" in Drexler
+  and Miller, "Incentive Engineering for Computational Resource Management,"
+  1988)
+- Upgrading programs in flight is difficult; the heap snapshot encodes
+  assumptions about program behavior that the upgrade may violate
+
+### Exposed partition and revival per-reference (E model)
+
+At the other end, partition and revival are exposed at every individual reference. A
+program must be written so that any dereference or message send to a potentially
+remote reference might fail due to partition. Recovery requires reconstructing
+the chain of computation that led to the broken reference, after partition heals.
+
+**Advantages:**
+- Simpler runtime implementation
+- Does not sacrifice availability to the extent of the Waterken model
+- No obligation to retain "offline capabilities" (sturdyrefs and out-of-band
+  URL-like references) indefinitely. Both are necessarily weak references.
+  Sturdyrefs are like out-of-band references but can participate in "distributed
+  confinement" without revealing cryptographic material to a confined program
+  with parts running on multiple peers.
+
+**Disadvantages:**
+- More complex programming model: every dependent computation must handle
+  mid-process recovery
+- Programs must reconstruct chains of computation defensively
+
+### The common substrate: URL-like references
+
+Both models share the notion of a URL or URL-like reference (sturdy reference,
+locator) that weakly retains a capability on a peer and can be redeemed for a
+live reference. In the Waterken model, these must be persisted indefinitely, or
+all dependent distributed processes are silently corrupted (they continue waiting
+for references that will never return). In E, sturdy references and locators
+serve as the basis for restoring connectivity after partition heals.
+
+In both models, petname systems are expected to be built *on top of* these
+reference mechanisms.
+
+## Formula Persistence: Inverting the Relationship
+
+Formula Persistence inverts the relationship between petnames and the underlying
+reference mechanism. The petname database is not built on top of sturdy
+references; instead, it *is* the persistence system.
+
+### The petname graph as persistence root
+
+The petname database is a graph database mapping human-readable, self-assigned
+names to capabilities. It maps a tree of named paths to locators. As in E, a
+locator serves to re-establish connectivity — but here, to the latest
+*incarnation* of a capability, via a level of indirection through the formula
+graph. The locators refer to nodes in a graph of formulas.
+
+### Formulas as reconstruction recipes
+
+Each formula describes:
+- How to arrive at a live reference for a capability
+- How to construct the capability's dependencies
+
+A formula is not a snapshot of state. It is a recipe for producing state. The
+system persists *construction*, not *content*.
+
+### Destruction by cohort, reconstruction on demand
+
+A capability is a member of a cohort: itself and the live references for its
+transitive dependencies. The system responds to partition with **destruction by
+cohort**: when any reference in the cohort becomes partitioned, the entire
+subgraph of dependent live references is collectively destroyed. The system then
+offers **reconstruction on demand**: the affected capabilities may later be
+reincarnated from their formulas when partition heals and a consumer requests
+them.
+
+This is the "pass by construction" property: rather than attempting to patch a
+partially broken graph of live references, the system destroys the affected
+cohort and rebuilds from formulas on demand.
+
+When a cohort is destroyed, the system provides a window for the hosting process
+to shut down gracefully — flushing external storage, releasing resources. During
+this window, partition of individual references becomes observable. However, the
+moment the daemon has committed to the disincarnation of a reference, the
+incarnation lives only in limbo — it cannot be reached through the daemon. Its
+severance is, from that point on, inconsequential.
+
+Rather than obligate the code to react to partition, we automate reconstruction.
+We require the code to manually persist anything that might need to survive
+reconstruction, and provide formula-based storage mechanisms to that end. We
+find this burden tolerable given that every system in which software can be
+upgraded necessarily has the same obligation.
+
+### The live reference graph and the formula graph
+
+The live reference graph passes through the heaps of distributed processes. It
+contains a mix of references that correspond to formulas and references that do
+not. Both kinds — formula-backed and ephemeral — can pass over CapTP among
+processes.
+
+Ephemeral references are bounded by sessions and can suffer partition.
+Formula-backed references can be reconstructed after partition, but only because
+someone previously arranged for the formula to exist.
+
+Creating a formula is not a matter of mere message passing. It requires appealing
+to a user agent — typically by proposing code that can construct the live
+reference, then persisting that code along with its petnamed dependencies. This
+is a deliberate act of policy, not an automatic consequence of holding a
+reference.
+
+Consequently, Formula Persistence does not entirely avoid the problem of
+programming explicitly against partition. Programs must still be aware that
+ephemeral references can vanish. What it offers is a more ergonomic way to do so:
+the formula graph provides a declarative substrate for recovery, and the petname
+database provides a human-legible map of what has been made durable and why.
+
+This is very much a hybrid approach. It sits between systems that mask partition
+and revival entirely and systems that expose them at every reference.
+
+### Graph structure and garbage collection
+
+The formula graph and the ephemeral reference graph have complementary structural
+properties:
+
+- The **formula graph is acyclic**. It can be collected with simple, local
+  reference counting. No distributed garbage collection protocol is needed.
+- The **ephemeral reference graph may be cyclic**. It is scoped to sessions.
+  Cyclic relationships are permitted in the ephemeral layer.
+
+The formula graph is acyclic across peers, but admits limited cycles among
+certain groups of formulas that must present unique, unforgeable identifiers to
+the network while being constructed as facets of a shared underlying capability.
+These include promise-and-resolver pairs and agent handle pairs.
+
+### Mitigating heap bloat
+
+Because values incarnated from formula can be reincarnated on demand, the user
+can observe that a connected peer is imposing an undue burden to retain ephemeral
+state and intervene:
+
+- The user can force the offending worker to restart, discarding its ephemeral
+  heap. Any formula-backed values will be reconstructed on demand.
+- The operating system can obligate a worker to restart as a matter of
+  out-of-memory mitigation, with the same consequence.
+- If neither is sufficient, the user can manually sever the offending peer.
+
+Heap bloat in the ephemeral layer is mitigated by interventions one layer down,
+in the formula graph. The formula graph provides the floor from which the system
+can always recover.
+
+### Timely revocation through local reachability
+
+The acyclicity of the formula graph enables timely destruction and severance of
+capabilities the user has revoked.
+
+Once a reference is locally unreachable in the petname graph, the corresponding
+live reference can be made immediately unreachable and gracefully destroyed.
+All heaps that refer to it and all CapTP sessions that retain it can be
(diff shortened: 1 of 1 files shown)


designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+state across restarts and upgrades. These dimensions are entangled — choices
+about partition visibility constrain persistence strategy and vice versa — but
+they are not the same axis. Formula Persistence is distinctive in making a
+different choice on each.
+
+### Masked partition and revival (Waterken model)
+
+At one end of the spectrum, the system masks partition and revival. All
+loss of connectivity is treated as temporary. The program never observes a broken
+reference; it simply waits. Waterken achieves this by combining Joe-E (a
```
erights, line 29: Some problems with initial phrasing:
- promises still reject/break, so "never observes a broken reference" is wrong. "never observes a reference broken due to partition" would be correct.
- "it simply waits" what simply waits? The Waterken computation model is as non-blocking as E or Endo -- communicating event loops. Programs never wait in the conventional sense of blocking.

```suggestion
loss of connectivity is treated as temporary. A message sent during loss of connectivity will still be delivered exactly once, after connectivity is reestablished. Waterken achieves this by combining Joe-E (a
```

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+capability-safe subset of Java) as its local object-capability language with its
+own live crypto-cap protocol in the CapTP family.
+
+This enables orthogonal persistence: the system can snapshot the heap and restart
+from that snapshot, possibly on a different host, without the program's
+knowledge. Because the program never distinguishes between "slow" and
+"partitioned," the system can be made deterministic over all communicating
+programs.
```
erights, line 37: There is still arrival order non-determinism, which is fundamental. If vatA and vatB both send a message to vatC at the "same time", they will arrive in some order, but the order is not determined by prior distributed semantic state.

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+
+**Advantages:**
+- Simpler programming model: no defensive code against partition
+- Determinism across communicating programs
+- Transparent relocation
+
+**Disadvantages:**
+- Sacrifices availability: a single partitioned dependency stalls all dependents
+- Entangled distributed heaps require distributed garbage collection
```
erights, line 46: Waterken did not do distributed gc at all. It simply had all distributed references leak forever, i.e., as long as those vats lived.

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+- Differences in incentives among participants necessitate market-based
+  approaches to garbage collection (see "The market-sweep algorithms" in Drexler
+  and Miller, "Incentive Engineering for Computational Resource Management,"
+  1988)
```
erights, line 50: Neither Waterken nor any other system has yet actually implemented market-based gc. Check back in another 40 years ;)

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+- Upgrading programs in flight is difficult; the heap snapshot encodes
+  assumptions about program behavior that the upgrade may violate
```
erights, line 52: Yeah, Tyler has a completely diff perspective on upgrade: https://waterken.sourceforge.net/upgrade/ . I don't know if anyone has actually tried this. It is interesting.

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+
+### Exposed partition and revival per-reference (E model)
+
+At the other end, partition and revival are exposed at every individual reference. A
+program must be written so that any dereference or message send to a potentially
```
erights, line 57:
```suggestion
program must be written so that any dereference or message send of a potentially
```

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+remote reference might fail due to partition. Recovery requires reconstructing
+the chain of computation that led to the broken reference, after partition heals.
```
erights, line 59: > Recovery requires reconstructing the chain of computation

Recovery from sturdy refs need not, and often does not, reconstruct the original chain of computation that produced the sturdy ref.

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+
+**Advantages:**
+- Simpler runtime implementation
+- Does not sacrifice availability to the extent of the Waterken model
```
erights, line 63: Tyler argues that E's "react to partition" translates in Waterken to "react to timeout". Tyler also points out what's different: E's partition atomically breaks all references that are multiplexed over the partitioned collection. But I think this detail is below the level you're trying to explain.

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+- No obligation to retain "offline capabilities" (sturdyrefs and out-of-band
+  URL-like references) indefinitely. Both are necessarily weak references.
```
erights, line 65: In some version of E, SturdyRefs had timeouts. Not clear what "obligation" means when nothing can force the counter-party to retain, but I certainly consider the host to be obligated to retain the reference until the timeout (or deadline) expires. IIRC, modern E dropped the timeouts. Does that mean the host is obligated to retain forever? I'm unsure how to answer the question.

This comparison with "weak references" is interesting. I never thought to describe it this way, but I think it is valid.

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+  Sturdyrefs are like out-of-band references but can participate in "distributed
+  confinement" without revealing cryptographic material to a confined program
+  with parts running on multiple peers.
```
erights, line 68: A way to think of Waterken in E terms is that all Waterken references are sturdy. Just like in E, computation denied the ability to convert between opaque references are bits could be confined in Waterken. I'm not sure about this, but it is definitely consistent with the Waterken model.

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+
+**Disadvantages:**
+- More complex programming model: every dependent computation must handle
+  mid-process recovery
```
erights, line 72: What does "mid-process" mean?

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+- Programs must reconstruct chains of computation defensively
```
erights, line 73: Is this a distinct point?

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+
+### The common substrate: URL-like references
+
+Both models share the notion of a URL or URL-like reference (sturdy reference,
+locator) that weakly retains a capability on a peer and can be redeemed for a
+live reference. In the Waterken model, these must be persisted indefinitely, or
```
erights, line 79: Waterken had no separate notion of a live reference. All references are sturdy. Because partitions are masked, you just send messages on these sturdy references and they are still delivered exactly once.

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+**Advantages:**
+- Simpler runtime implementation
+- Does not sacrifice availability to the extent of the Waterken model
+- No obligation to retain "offline capabilities" (sturdyrefs and out-of-band
+  URL-like references) indefinitely. Both are necessarily weak references.
+  Sturdyrefs are like out-of-band references but can participate in "distributed
+  confinement" without revealing cryptographic material to a confined program
+  with parts running on multiple peers.
+
+**Disadvantages:**
```
erights, line 70: In E, a partition drops messages in flight, without E itself providing any bookkeeping to ascertain after the fact which messages were lost. Thus, in E, messages are delivered at most once. This creates way more application complexity than an exactly-once guarantee.

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+- More complex programming model: every dependent computation must handle
+  mid-process recovery
+- Programs must reconstruct chains of computation defensively
+
+### The common substrate: URL-like references
+
+Both models share the notion of a URL or URL-like reference (sturdy reference,
+locator) that weakly retains a capability on a peer and can be redeemed for a
+live reference. In the Waterken model, these must be persisted indefinitely, or
+all dependent distributed processes are silently corrupted (they continue waiting
```
erights, line 80: Corruption is loss of integrity. The case you're describing is at most loss of availability.

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+for references that will never return). In E, sturdy references and locators
```
erights, line 81: References don't return. They resolve/forward/settle/fulfill/break, or not.

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+  manually persisted storage is lost across incarnations.
+
+## Formula Persistence as a Choice within Endo
+
+The Endo Daemon implements Formula Persistence, but Formula Persistence is not
+intrinsic to Endo. Endo provides a shared model for passable values (data and
+capabilities), patterns, and message passing. Other systems built on the same
+Endo components make different choices along the entangled dimensions:
+
+- The choice of **CapTP** determines message ordering.
```
erights, line 328: We've converged on the data model of ocapn whose order is point-to-point fifo, with the assumption that we will add an "after" operation eventually to express stronger orders. But in any case, the CapTP choices in Endo should not determine any message order other than point-to-point fifo (+ "after"), yes?

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+- The choice of a **Network** determines the range within which pass-invariant
+  equality can be relied upon.
+- The **Daemon** chooses Formula Persistence.
+
+For example, the Agoric chain uses Endo components with orthogonal persistence to
+ensure that all honest validators produce the same deterministic computation,
+independent of whether they crashed and restarted or simply continued. Formula
```
erights, line 335: Agoric's use of Endo follows the Waterken model

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+locator) that weakly retains a capability on a peer and can be redeemed for a
+live reference. In the Waterken model, these must be persisted indefinitely, or
+all dependent distributed processes are silently corrupted (they continue waiting
+for references that will never return). In E, sturdy references and locators
+serve as the basis for restoring connectivity after partition heals.
+
+In both models, petname systems are expected to be built *on top of* these
+reference mechanisms.
+
+## Formula Persistence: Inverting the Relationship
```
erights, line 87: Need to see a concrete example early with an actual concrete formula.

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+properties:
+
+- The **formula graph is acyclic**. It can be collected with simple, local
+  reference counting. No distributed garbage collection protocol is needed.
+- The **ephemeral reference graph may be cyclic**. It is scoped to sessions.
+  Cyclic relationships are permitted in the ephemeral layer.
+
+The formula graph is acyclic across peers, but admits limited cycles among
+certain groups of formulas that must present unique, unforgeable identifiers to
+the network while being constructed as facets of a shared underlying capability.
```
erights, line 175: runon sentence. Too hard for me to parse even though I'm already more oriented than your readers.

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+that were already granted — "harassing the user" — erodes trust and makes the
+system unusable in practice.
+
+Formula-based retention of policy enables this. The formula graph encodes not
+just how to reconstruct a capability but the fact that it was authorized. When an
+ephemeral session is re-established, the system can re-derive the authorized
+capabilities from their formulas, restoring the user's prior policy decisions
+without requiring the user to re-confirm them.
+
+### Revocation by withdrawal of construction
```
erights, line 279: Is this selective revocation? If not, it fails to express the main use case for revocation: revoking the access I gave to Bob without revoking the access I gave to Carol.

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+This cascades into the disincarnation of the corresponding live reference and
+anything that depends upon it for its own construction.
+
+Because the formula graph is acyclic and locally managed, this revocation is
+immediate and requires no distributed protocol (as described above in "Timely
+revocation through local reachability").
+
+## Why Not Orthogonal Persistence?
+
+### The upgrade problem dissolves the distinction
```
erights, line 294: I might put this as "whether on not persistence is orthogonal, upgrade cannot be orthogonal"

erights: I do not suggest you go into any depth on the Agoric upgrade model. But just noting:

Agoric's current support for upgrade treats messages sent during an upgrade, under some circumstances, as "deliver at most once", creating all the application defensiveness burden Agoric tried to avoid by adopting the Waterken persistence and communication model. Instead, Agoric's next refinement of its upgrade model, based on durable promises, will preserve the "deliver exactly once" guarantee even across upgrades.

However, I do suggest that for formula persistence you do discuss what the message delivery guarantees are across distributed traumas.

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+load, no replay log to process. The formulas are evaluated lazily as
+capabilities are demanded.
+
+### What is sacrificed
+
+- **Determinism:** Reconstruction from formula may produce observably different
+  results from one incarnation to the next (e.g., if a dependency's behavior
+  has changed).
+- **Ephemeral state:** Heap state that is not captured in a formula or in
+  manually persisted storage is lost across incarnations.
```
erights, line 319: This is true for Agoric as well, since incarnations are upgrade boundaries, not crash recovery boundaries.

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+### Formulas as reconstruction recipes
+
+Each formula describes:
+- How to arrive at a live reference for a capability
+- How to construct the capability's dependencies
+
+A formula is not a snapshot of state. It is a recipe for producing state. The
+system persists *construction*, not *content*.
+
+### Destruction by cohort, reconstruction on demand
```
erights, line 110: I don't get cohorts yet. Probably will once there's an example with a concrete formula?

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+
+## Position in the Design Space
+
+| Property | Waterken (masked partition/revival) | E (exposed partition/revival) | Formula Persistence |
+|---|---|---|---|
+| Partition and revival | Masked | Exposed per-reference | Exposed per-cohort |
+| Persistence mechanism | Orthogonal | Manual + sturdy refs | Formula graph |
+| Programming model | Simple (no partition code) | Defensive (per-reference) | Moderate (cohort-aware) |
+| Restart cost | Snapshot restore | Reference re-establishment | Formula evaluation (lazy) |
+| Upgrade story | Difficult (heap assumptions) | Natural (references re-resolve) | Natural (formulas re-evaluate) |
```
erights, line 404: For E, "natural" is way too generous. It is correct wrt recovering connectivity. But E's persistence was manual, so upgrade was no *additional* burden. It was just the same manual revival from what was manually stored. But this burden of manual storage and revival was awkward and "unnatural".

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+| Retention: live references | Indefinite (partition masked) | Distributed acyclic GC | Scoped to cohort |
```
erights, line 405: In E, live refs also only retained up to partition.

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+| Retention: durable references | Indefinite (web-keys) | Weak (sturdyrefs) | Local reference counting (formula graph) |
```
erights, line 406: On second thought, for E, "Weak" is too weak. It is part of the application semantics, rather than the platform semantics, whether a host must honor a sturdy ref. It has to be a host/app choice. Otherwise it is impossible to write any correct distributed system. For example, an ERTP issuer must retain all purses for which there may be an outstanding sturdy ref.

"Weak" implies that it is not an app choice, but rather a platform choice based on strong reachability. Yes, one can build ERTP's obligations on top by arranging permanent strong reachability. But this goes against the connotations of "weak".

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+| Availability | Sacrificed for consistency | Maintained per-reference | Maintained per-cohort |
```
erights, line 407: For E, I don't understand "availability" "Maintained per-reference".

By using timeouts where E uses partition, Waterken still provides extremely high availability.

I also don't yet understand "per cohort" because I don't yet understand "cohort". Will revisit once I do.

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+### Market-based distributed garbage collection
+
+Drexler and Miller (1988) propose market-based mechanisms for computational
+resource management, including a distributed garbage collection algorithm able to
+collect unreferenced loops that cross trust boundaries through decentralized
+market negotiations. This addresses the problem that orthogonal persistence with
+entangled distributed heaps entrains an obligation for distributed garbage
+collection, and that differences in incentives among participants necessitate
+market-based solutions. Formula Persistence sidesteps this obligation by keeping
+the formula graph acyclic and locally reference-counted.
```
erights, line 464: For formula persistence, when is a host required to preserve state on behalf of remote clients? This answer cannot be "none" or "never". If there is such an obligation, as there always must be in any significant distributed app, then you're not sidestepping the need for incentives. You're ignoring the incentives, just as every other distributed system has always done. Including, ironically, Agoric.

designs/daemon-persistence.md
```diff
@@ -0,0 +1,516 @@
...
+distributed, secure, persistent, and ubiquitous computational fabric — enabling
+mutually suspicious parties to cooperate safely through the exchange of rights.
+This work is an ancestor of the Endo/Agoric platform on which Formula Persistence
+is built.
+
+- <http://papers.agoric.com/papers/distributed-electronic-rights-in-javascript/abstract/>
+
+### Petname systems
+
+Marc Stiegler's petname systems describe a naming architecture with three
```
erights, line 483: Grumble. AFAIK Marc everywhere acknowledges that I invented petname systems. But he's certainly done much much more to explain, implement, and promulgate them than I have.

My first and only petname paper: https://erights.org/elib/capability/pnml.html
Has the notion of embedded cards with different terminology. Has paths. Worth a look when you have time ;)

Having gotten all that off my chest, please keep "Marc Stiegler's petname systems". No change suggested.

erights: The Acknowledgements at the bottom of the PNML paper is a nice historical record of proper credit.

PR #3130, opened 2026-03-17
feat(ses): add support for "code" prop in SES-managed Errors

boneskull:
- Adds a new `string` option to `AssertMakeErrorOptions`, `code`
- If a `string`, will be an non-enumerable property of the `Error` instance returned by `makeError`.
- To avoid mutating global types, a new type `SesError` is introduced which is simply `Error & {code?: string}`.

Note that `code` is expected to be a `string` here, but only by convention; the language will not enforce this (as per the below proposal).

Ref: https://github.com/tc39-transfer/proposal-error-code-property

### DIFF
--- a/packages/ses/src/error/assert.js
+++ b/packages/ses/src/error/assert.js
@@ -295,7 +295,7 @@ const tagError = (err, optErrorName = err.name) => {
  *     such as `stack` on v8 (Chrome, Brave, Edge?)
  *   - `sanitizeError` will freeze the error, preventing any correct engine from
  *     adding or
- *     altering any of the error's own properties `sanitizeError` is done.
+ *     altering any of the error's own properties once `sanitizeError` is done.
  *
  * However, `sanitizeError` will not, for example, `harden`
  * (i.e., deeply freeze)
@@ -315,10 +315,16 @@ export const sanitizeError = error => {
     errors: _errorsDesc = undefined,
     cause: _causeDesc = undefined,
     stack: _stackDesc = undefined,
+    code: codeDesc = undefined,
     ...restDescs
   } = descs;
 
   const restNames = ownKeys(restDescs);
+
+  // the spec allows any value, but we drop 'code' if it's not a string
+  if (codeDesc?.value !== undefined && typeof codeDesc.value !== 'string') {
+    arrayPush(restNames, 'code');
+  }
   if (restNames.length >= 1) {
     for (const name of restNames) {
       delete error[name];
@@ -350,6 +356,7 @@ const makeError = (
     cause = undefined,
     errors = undefined,
     sanitize = true,
+    code = undefined,
   } = {},
 ) => {
   // The first two parameters above cannot be inferred unless this is rewritten
@@ -357,7 +364,8 @@ const makeError = (
   // that we at least have type-safety within the function body.
   //
   // Note that due to the overload of AssertionUtilities['makeError'], strict
-  // mode will complain if default parameters are provided in the method
+  // Note that due to the overload of AssertionUtilities['makeError'], Typescript's so-called "strict
+  // mode" will complain if default parameters are provided in the method
   // signature. The below workaround (optDetails -> details; errConstructor ->
   // errCtor) is functionally equivalent but allows us to use type assertions to
   // workaround the strict mode issue.
@@ -397,6 +405,14 @@ const makeError = (
         configurable: true,
       });
     }
+  }
+  if (code !== undefined) {
+    defineProperty(error, 'code', {
+      value: code,
+      writable: true,
+      enumerable: false,
+      configurable: true,
+    });
   }
   weakmapSet(hiddenMessageLogArgs, error, getLogArgs(hiddenDetails));
   if (errorName !== undefined) {
--- a/packages/ses/src/error/assert.js
+++ b/packages/ses/src/error/assert.js
@@ -368,7 +368,7 @@ const makeError = (
   // mode" will complain if default parameters are provided in the method
   // signature. The below workaround (optDetails -> details; errConstructor ->
   // errCtor) is functionally equivalent but allows us to use type assertions to
-  // workaround the strict mode issue.
+  // workaround the issue with TypeScript's so-called "strict mode".
   let details = /** @type {Details} */ (
     optDetails ?? redactedDetails`Assert failed`
   );
--- a/packages/ses/types.d.ts
+++ b/packages/ses/types.d.ts
@@ -242,6 +242,11 @@ export interface AssertMakeErrorOptions {
    * {@link sanitizeError}.
    */
   sanitize?: boolean;
+
+  /**
+   * The error code to assign to the error.
+   */
+  code?: string;
 }
 
 // TODO inline overloading
--- a/packages/ses/test/error/assert-log.test.js
+++ b/packages/ses/test/error/assert-log.test.js
@@ -339,6 +339,21 @@ test('makeError named', t => {
   );
 });
 
+test('makeError code', t => {
+  const err = makeError(X`<${'bar'},${q('baz')}>`, URIError, {
+    code: 'FOO_ERR',
+  });
+  t.is(/** @type {Error & {code: string}} */ (err).code, 'FOO_ERR');
+});
+
+test('makeError code dropped (non-string)', t => {
+  const err = makeError(X`<${'bar'},${q('baz')}>`, URIError, {
+    // @ts-expect-error intentional bad input
+    code: 123,
+  });
+  t.is(/** @type {Error & {code?: unknown}} */ (err).code, undefined);
+});
+
 test('assert.quote', t => {
   throwsAndLogs(t, () => Fail`<${'bar'},${q('baz')}>`, /<\(a string\),"baz">/, [
     ['log', 'Caught', Error],
--- a/packages/ses/types.test-d.ts
+++ b/packages/ses/types.test-d.ts
@@ -1,6 +1,6 @@
 /* eslint-disable @endo/no-polymorphic-call, import/no-extraneous-dependencies, no-restricted-globals, no-underscore-dangle */
-import { expectType } from 'tsd';
 import type { Assert } from 'ses';
+import { expectAssignable, expectType } from 'tsd';
 
 // Lockdown
 
@@ -221,6 +221,18 @@ expectType<Error>(makeError(X`details are ${stringable}`, TypeError));
 expectType<Error>(
   makeError(X`details are ${stringable}`, TypeError, {
     errorName: 'Nom de plum',
+  }),
+);
+
+expectType<AggregateError>(
+  makeError(X`details are ${stringable}`, AggregateError, {
+    errors: [new Error('error 1'), new Error('error 2')],
+  }),
+);
+
+expectAssignable<Error & { code?: string }>(
+  makeError(X`details are ${stringable}`, Error, {
+    code: '123',
   }),
 );
 


packages/ses/types.d.ts
```diff
@@ -384,6 +389,11 @@ export interface AssertionFunctions extends BaseAssert {
   ): never;
 }
 
+/**
+ * An `Error` with a `code` property.
+ */
+export type SesError = Error & { code?: string };
```
kriskowal: Consider `CodedError` (imprecise) or `OptionallyCodedError` (ew) or `MaybeCodedError` (ew but less so) or `ErrorConCodé` (facetious) or `CodableError` (getting there, leaves air for `CodedError` to require the code) since this isn’t particularly intrinsic to SES.

```ts
type CodedError = Error & { code: string };
type CodableError = Error & {code?: string };
```

Could be that we overload these functions so that `CodedError` is implied by presence of `code` option.

packages/ses/types.d.ts
```diff
@@ -305,14 +310,19 @@ export type GenericErrorConstructor =
   | ErrorConstructor
   | AggregateErrorConstructor;
 
+/**
+ * An `Error` with a `code` property.
+ */
+export type SesError = Error & { code?: string };
```
boneskull: Open to naming suggestions

kriskowal (commented):
Needs changeset and consensus on names.

packages/ses/types.d.ts
boneskull: (We could have used module augmentation to extend the global `Error` interface with a `code?: string` prop, but that would affect all downstream consumers of `ses`, which would be inconsiderate)

packages/ses/types.d.ts
boneskull: "Codeträger?" "Fehlercodetragendesding?" "Codebesitzend" really rolls off the tongue

boneskull: Going w/ `CodableError`.  🐟

packages/ses/src/error/assert.js
erights:
```suggestion
 *     altering any of the error's own properties once `sanitizeError` is done.
```

This was my mistake, but might as well add as a drive-by to this PR is you don't mind.

packages/ses/src/error/assert.js
```diff
@@ -378,6 +380,14 @@ const makeError = (
       });
     }
   }
+  if (code !== undefined && typeof code === 'string') {
```
erights: Here, there's no good answer for "what do to if `code` is not a string?". Rather, I think that here, following the proposal, you should not filter by the value of `code`. Rather, put this value test above in `sanitizeError`. There, if you see a non-string code, delete it and add it to `droppedDetails` so that it is hidden but shows up as a diagnostic on console output.

packages/ses/types.d.ts
erights: Under any name, I don't like introducing a new type for this purpose if possible.

What about `cause` and `errors`? Do they have this problem too? If not, can we treat `code` the same way? Once we support `SuppressedError` (hopefully soon) we will also need to support the further instance properties `error` and `suppressed`.

erights (changes requested):

packages/ses/types.d.ts
boneskull: They don't have the same problem, because `Error.cause` and `AggregateError.errors` are in the TS lib.  

Our options are:

1. Create a new type (as this PR does)
2. Augment the global `Error` type with a `code` prop
  This is hazardous because it will affect downstream consumers; anyone pulling in `ses` will have a global `Error` with a `code` prop on it. If we do not publish the augmentation, types will work in a development environment, but will be broken downstream. 
3. Fudge it with `// @ts-expect-error` everywhere `code` is referenced (gross)

I'm not sure what TS' policy is on adopting proposals (if they have one), but I think it's probably too early to ask.

* * *

That being said, `code` will/should/might end up in the TS lib eventually, at which point we can refactor away the custom type and/or alias it to `Error`.

mhofman:
FYI, I [found myself](https://github.com/Agoric/ymax-web/pull/534) in need of creating errors with a `.code`, so I overrode and created extra helpers around `@endo/errors`

In particular I'm adding an `XError` helper to create Error using template literals like `Fail` does but without throwing (a wrapper around `makeError` but without the constructor and details cruft), and make both this `XError` and `Fail` chainable to define the options that `makeError` accepts (credit to @gibson042 for the idea).

<details>

<summary>errors.ts:</summary>

```ts
import { makeError as originalMakeError, X } from "@endo/errors";

export * from "@endo/errors";

type Details = NonNullable<Parameters<typeof originalMakeError>[0]>;
type GenericErrorConstructor = NonNullable<
  Parameters<typeof originalMakeError>[1]
>;
type AssertMakeErrorOptions = NonNullable<
  Parameters<typeof originalMakeError>[2]
> & {
  code?: string;
};

type ChainedErrorMaker<T> = ((
  template: TemplateStringsArray | string[],
  ...args: any[]
) => T) & {
  code(code: string): ChainedErrorMaker<T>;
  instance(ctor: GenericErrorConstructor): ChainedErrorMaker<T>;
  options(options: AssertMakeErrorOptions): ChainedErrorMaker<T>;
};

const assertFailedDetails = X`Check failed`;

export const makeError = (
  details: Details = assertFailedDetails,
  errConstructor: GenericErrorConstructor | undefined = undefined,
  options: AssertMakeErrorOptions | undefined = undefined,
): Error => {
  const { code, sanitize = false, ...makeErrorOptions } = options ?? {};
  const hasCode = code !== undefined;
  const error = originalMakeError(details, errConstructor, {
    ...makeErrorOptions,
    // `code` and `sanitize` are mutually exclusive options, ignore sanitize
    sanitize: !hasCode && sanitize,
  });
  if (code !== undefined) {
    Object.defineProperty(error, "code", {
      value: code,
      enumerable: false,
      configurable: true,
      writable: true,
    });
  }
  return error;
};

export const fail = (
  /** The details of what was asserted */
  details: Details = assertFailedDetails,
  /** An optional alternate error constructor to use */
  errConstructor: GenericErrorConstructor | undefined = undefined,
  options: AssertMakeErrorOptions | undefined = undefined,
): never => {
  throw makeError(details, errConstructor, options);
};

const makeChainedHelper = <T>(
  fn: (...args: Parameters<typeof makeError>) => T,
) => {
  const name = fn.name && fn.name.charAt(0).toUpperCase() + fn.name.slice(1);
  const maker = (
    options: AssertMakeErrorOptions = {},
    constructor?: GenericErrorConstructor | undefined,
  ): ChainedErrorMaker<T> => {
    // XXX: Consider defaulting to carrying over the code of any error passed in values
    // with the caveat that such value may be classified (quoted or bare)
    const Fail: ChainedErrorMaker<T> = (template, ...values) =>
      fn(X(template, ...values), constructor, options);
    Fail.code = (code: string) => maker({ ...options, code }, constructor);
    Fail.instance = (ctor: GenericErrorConstructor) => maker(options, ctor);
    Fail.options = (opts: AssertMakeErrorOptions) =>
      maker({ ...options, ...opts }, constructor);
    Object.defineProperty(Fail, "name", { value: name });
    return Fail;
  };
  return maker();
};

export const Fail = makeChainedHelper(fail);

export const XError = makeChainedHelper(makeError);
```
</details>

mhofman:
> @mhofman Wait does this mean I can rewrite `@endo/compartment-mapper` in TS?? 😄

When https://github.com/endojs/endo/pull/3137 lands, yes

boneskull:
@erights 

> Under any name, I don't like introducing a new type for this purpose if possible.

> What about cause and errors? Do they have this problem too? If not, can we treat code the same way? Once we support SuppressedError (hopefully soon) we will also need to support the further instance properties error and suppressed.

`cause` is already part of TS' lib as the sole prop of `ErrorOptions`. `errors`, likewise, is in TS' lib as a prop of `AggregateError`.  Given `code` being in stage 1, I wouldn't expect it to make it into TS' lib until it's further down the proposal stages.

To recap, there are two ways that I'm aware of to declare a `code` prop:

1. **Create a new type with the `code` prop.**
  When `code` lands in TS' lib, we can deprecate/remove the type however we see fit.
2. **Globally augment `Error` with a `code` prop.**
  The problem with this is that it would affect downstream consumers; when they `import 'ses'`, all of their `Error`s will be affected by our augmentation. Worst case, it would conflict with a downstream augmentation of `Error`.  A library like `ses` should _not_ do this, but it's fair play in an app.

While I think 1 is the only feasible way forward, it would be considered a temporary solution, albeit indefinitely so. 😄

erights:
For anyone that uses `code` right now in normal JS, they would face the same dilemma, that it is not yet in the TS type. I don't think we need to be ahead of the normal JS experience in this. They would need to cope in the same manner for both, `@ts-expect-error` or `@ts-ignore`, so we do achieve parity. I prefer this to either (1) or (2) above.

Frankly, if I had to choose between (1) and (2), I would be inclined towards (2). But I agree that neither is correct.

erights:
> What I'd really like to see here is a `makeError` which is generic on the error constructor type (so that it would return a value of that type), but the signature is such that it is not possible--the default parameter of `globalThis.Error` creates a type error when it's generic, and removing the default value is invalid syntax (since the previous message/details parameter _also_ has a default value).
> 
> https://github.com/endojs/endo/blob/e6d46c72775ab7dd22a165fb15ab580f85b7b4f8/packages/ses/src/error/assert.js#L341-L352
> 
> The workaround is to just wrap `makeError` in another function.

I don't understand this. The declared type of `makeError` is already in terms of

```js
export type GenericErrorConstructor =
  | ErrorConstructor
  | AggregateErrorConstructor;
```

Unpleasantly, we will need to expand this to include `SuppressedError` as well, but I don't think that bears on this PR.

packages/ses/types.d.ts
erights: Ah. You already enumerated that option earlier with (3). Yes, it is gross. But it achieves parity with normal JS use, so I consider it acceptable. And more acceptable than (1) or (2).

erights:
> > See also DRAFT PRs #2223 and #2429 . Like them, we need to coordinate this change with @endo/pass-style and @endo/marshal so that `.code` is actually passed. Doesn't need to happen in this PR, but that is my preference since it should be a coordinated change. Please borrow as much as you like from those other draft PRs.
> 
> @erights What should I be looking at there? I'm unfamiliar with both of those packages and any discussions around either PR.
>
> I'm operating under the assumption that this PR only satisfies the (fairly minimal) requirements of `@endo/compartment-mapper` so that it can be consumed sooner rather than later.
> 
> I'd expect we'd extend the implementation and/or adopt in other packages when ready. Are you concerned about shipping this in `ses` as-is? If so, would you be amenable to _moving_ the implementation into `@endo/compartment-mapper` temporarily so it can be trialed in that context, and then re-integrating with `ses` in the future?

No, we should keep it in ses where it would eventually land anyway. Moving it would be more unpleasant.

My concern is that endo as a batteries included distributed secure ocap variant of JS violates the principle of least surprise badly if we locally support some error properties (like `cause`, `errors`, `error`, `suppressed`, and `code`) at the SES level but not the pass-style or marshal level. In particular, to just a frozen error to not be passable when it has only properties we support, with the kinds of values we accept, is terrible.

I agree that can be a later PR that depends on this PR. However, I would not want to see an endo release containing this PR and not the other one. Thus, even once I'm otherwise ready to approve this PR I'd like to postpone merging until some reconciliation with pass-style and marshal is close.

erights:
I just looked at all your uses of `CodableError`. For all of these I find (3) acceptable. The pain is not worth the complexity of (1) or the bad practice of (2).

Other than that and https://github.com/endojs/endo/pull/3130#issuecomment-4724772297 , I have now looked at the entire PR and am ready to approve it after those are dealt with.

erights:
For the follow-on PR needed to merge this one, @kriskowal and I discussed what the semantics would be for an Error to be Passable, and to round trip through @endo/marshal.

- The only special cases for Error property names would be `name`, `message`, and `stack`. We would seek agreement with ocapn to standardize on these special cases.
  - `name` must be a string and will be used as the high-order bit of error classification. `code` will in practice be the low order bit, but it doesn't need a special case. Also, this is the only property that is not necessarily and own property of the error instance, but rather inherited from its prototype. If an unexpected `name` is unmarshalled, marshal should create a new Error subclass with that `name` and memoize the association within that marshaller instance.
  - `message` must be a string. The contents of the `message` string are only for diagnostic purposes. The contents of the `message` string is not something any non-diagnostic code should make a decision on. Thus, an improvement to the content of these strings does not warrant even a minor version bump. It is everyone else's responsibility not to break under changes to this string. This is the extent to which it is a special case, which might only need to be a non-normative note in ocapn.
  - We reserve the name `stack` because it should not be a transmitted property value. This is because the association between `stack` and callstacks on errors in JS is too strong to consider breaking. OTOH, actual callstack information reveals more information than the ocap model allows without special meta-level powers (like `getStackString` in the start compartment only. Or like the behavior of the causal console.) Once all engines implement the Error stack proposal (now at stage 3), we will be able to redact the stack locally. The redacted stack should locally be an empty string. The actual unredacted callstack information should be logged somewhere, so that the participants in a distributed system can choose which callstack logs they wish to share with whom.
- Symbol-named properties
  - In light of [Immutable ArrayBuffer README Purposeful Violation in PR](https://github.com/endojs/endo-but-for-bots/blob/82b8fa90f8dd308d3a2ddf9bc6b42f838e336653/packages/immutable-arraybuffer/README.md#purposeful-violation), we need a special local carve out allowing the existence of an own `Symbol.toStringTag` property whose value is a string. This is never transmitted.
  - No other symbol-named properties. 
- No accessor properties aside from the inherited `Error.prototype.stack` accessor from the Error stacks proposal.
- We need to make a decision about enumerability, but I am not sure what that decision should be.
- All other allowed properties must be string named, and their values must only be throwables, recursively. A throwable is a tree of passable copy-data and passable errors whose leaves are only primitives or passable errors. All the current special cases, `cause`, `errors`, `error`, `suppressed`, and `code` fall in this category and do not need any special case handling. If tc39 later adds a new error subclass with a new name and new property names, then a site that knows this can still round trip those errors through a site that has never hear of it.
- When we locally log the association between an error and its unredacted info (callstack, unredacted error message, and notes), we must also include something in the error descriptor sent over ocapn that is adequate to look up that log entry is one has that log. This is not yet a solved problem. ***@mhofman , have you explained your "notes" proposal anywhere?***

erights:
We need to uniquely identify each passed error so we can locally log its association with the unredacted info (callstack, unredacted error message, notes). @mhofman 's notes proposal associates the error instance with a clist entry much as we do for remote promises and targets.

Open question: Given that, does it still make sense to say, in the distributed object semantics, that error instances have no comparable identity, like promises don't. Or should we say this unique lookup info (whether clist based on not) is a comparable identity, like remotables/targets.

erights:
The points above + the relevant prior PRs and issues are probably enough for me to get @kriscendobot to draft an endo-but-for-bots PR for the needed first steps. I am content to further kick the error-id can down the road for now. ***@kriskowal , does this make sense to you? Do the points I captured above accurately reflect your understanding of what we agreed on?*** Should we get @kriscendobot started on this?

boneskull:
@erights 

### `Error.code`

> For anyone that uses code right now in normal JS, they would face the same dilemma, that it is not yet in the TS type. I don't think we need to be ahead of the normal JS experience in this. They would need to cope in the same manner for both, @ts-expect-error or @ts-ignore, so we do achieve parity. I prefer this to either (1) or (2) above.

I would never expect a `code` prop on an `Error`/`AggregateError` (at least until the proposal lands).  If I _am_ expecting a `code` prop on some `Error`-like object project, I _am_ going to be defining a `CodableError`-like type and will use a type assertion or type guard instead of `@ts-expect-error`/`@ts-ignore`. 

> e.g.: most errors thrown out of Node.js have a `code` prop, and such errors are of type `ErrnoException` per `@types/node`. When I want to reference that prop, I use a type assertion or guard.

I think what I'm trying to say here is if there's a `code` in my object, it is not strictly an `Error`.

### `CodableError`

In this PR (as written), `CodableError` is exported, but it needn't be.  If we kept it internal, we would also be able to leverage type assertions.  _But_ I'm not seeing a significant benefit to defining `CodableError` as an internal-only type in `ses` _while we're not actually consuming the field anywhere_.  If we _do_ end up consuming `code` in `ses`, then leaning on `@ts-*` directives will suck.

If we _don't_ export a type from `ses`, projects in Endo wanting to use `code` are then  incentivized to create their own `CodableError`-like types—I suspect others would also want to avoid `@ts-*` directives.  I think I was trying to avoid creating this incentive.

That said, I'm fine with simply laying the groundwork in `ses` for `code` to work as we expect at runtime.  

### `makeError` signature

> > What I'd really like to see here is a `makeError` which is generic on the error constructor type (so that it would return a value of that type), but the signature is such that it is not possible--the default parameter of `globalThis.Error` creates a type error when it's generic, and removing the default value is invalid syntax (since the previous message/details parameter _also_ has a default value).
> > https://github.com/endojs/endo/blob/e6d46c72775ab7dd22a165fb15ab580f85b7b4f8/packages/ses/src/error/assert.js#L341-L352
> > 
> > The workaround is to just wrap `makeError` in another function.
> 
> I don't understand this. The declared type of `makeError` is already in terms of
> 
> ```js
> export type GenericErrorConstructor =
>   | ErrorConstructor
>   | AggregateErrorConstructor;
> ```
> 
> Unpleasantly, we will need to expand this to include `SuppressedError` as well, but I don't think that bears on this PR.

What I was saying was that `makeError` returning a union type such as `GenericErrorConstructor` is not narrow enough for my sensibilities—mainly due to the discrepancy between `Error` and `AggregateError`.  I was suggesting something like this would be more kind to the consumer:

```ts
const makeError<T extends GenericErrorConstructor> = (
  optDetails: Details, 
  errConstructor: T, 
  options?: AssertMakeErrorOptions
): InstanceType<T> {
   /// ...
}
```

> I note the type of `AssertionUtilities.makeError` returns an `Error`, which is plainly wrong (since it can be an `AssertionError`). Returning `Error | AggregateError` is technically correct, but awkward for consumers.

What currently prevents changing the signature of `makeError` to the above is the _default value_ of `errConstructor` (`Error`).  TS will complain about this, since `T` satisfies `GenericErrorConstructor` but may not _be_ `ErrorConstructor`. We'd have to remove the default value of the second parameter (`errConstructor`) for it to work, which is impossible because the first parameter (`optDetails`) _also_ has a default value (and of course an expected parameter cannot follow an optional parameter).

I think we could create an overload for `AssertionUtilities.makeError` using the above signature. This would fix the following problem:

```ts
// ✅ err is type Error; this currently works as expected
const err = makeError('foo');

// ❌ err is type AggregateError; this currently does not work (aggErr is Error)
const aggErr = makeError('foo', AggregateError);
```

At any rate, I think it's worth doing (and afaict only needs type changes), but it's orthogonal to this PR.  

**UPDATE**: See https://github.com/endojs/endo/pull/3309

### Next Steps

> I just looked at all your uses of `CodableError`. For all of these I find (3) acceptable. The pain is not worth the complexity of (1) or the bad practice of (2).
> 
> Other than that and [(earlier comment)](https://github.com/endojs/endo/pull/3130#issuecomment-4724772297) , I have now looked at the entire PR and am ready to approve it after those are dealt with.

Confirming I understand: in order for this to be merged, I need to:

1. Remove the `CodableError` type
2. Wait for a follow-up PR which satisfies [your design here](https://github.com/endojs/endo/pull/3130#issuecomment-4725020402)

Is that correct?

erights:
> ```ts
> const makeError<T extends GenericErrorConstructor> = (
> ```

would be fine.

> I think we could create an overload for `AssertionUtilities.makeError` using the above signature. 

An overload would be fine too.

> Confirming I understand: in order for this to be merged, I need to:
> 
> 1. Remove the `CodableError` type
> 2. Wait for a follow-up PR which satisfies [your design here](https://github.com/endojs/endo/pull/3130#issuecomment-4725020402)
> 
> Is that correct?

Yes. Better, after sleeping on it, I am ok seeing this merged before we implement that design, even if that means we ship a release with one and not the other. The resulting problem is minor enough that we can solve it in documentation for now.

erights:
Some of the CI errors seem clearly unrelated to your PR so I hit "Rerun failed jobs". We'll see if it helps.

packages/ses/src/error/assert.js
erights:
```suggestion
// Note that due to the overload of AssertionUtilities['makeError'], Typescript's so-called "strict
// mode" will complain if default parameters are provided in the method
```

erights:
```suggestion
  // workaround the issue with TypeScript's so-called "strict mode".
```

erights: For TypeScript to call their thing "strict mode" is horribly confusing since JS has long had a "strict mode" that is totally unrelated. We should only use the unqualified "strict mode" when referring to JS's strict mode.

boneskull:
@erights I think this is ready now.  I removed the new type and added a check in `types.test-d.ts`.  I also added another check for changes in #3309 showing that `makeError` can now return an `AggregateError`.

erights (approved):
Thanks for killing `CodableError`.

LGTM!

boneskull:
Just realized this probably needs a changeset. ugh

boneskull:
Ref: #3316

PR #3131, opened 2026-03-19
evasive-transform - class method evasion and extensibility

naugtur:
## Description

- add extensibility - so it's possible to externally add a transform visitor to be applied on the AST, since it was already parsed and it's best to not parse again if someone wants to add more transforms
- add a simple class field transform I found I need

### Security Considerations

none

### Scaling Considerations

none

will need to document the new field

### Testing Considerations

tests updated

 all good 

### Upgrade Considerations

none


packages/evasive-transform/src/transform-code.js
```diff
@@ -226,3 +226,19 @@ export const evadeDecrementGreater = p => {
...
+/**
+ * @param {import('@babel/traverse').NodePath} p
+ */
+export const evadeClassMethod = p => {
+  const { node } = p;
+  // find class definitions with a method named import and turn it into `['import']()` syntax in class definition
+  if (
+    p.isClassMethod() &&
+    node.key.type === 'Identifier' &&
+    node.key.name === 'import'
```
naugtur: I know we could also match eval for completeness, but eval was not handled by evasive-transforms yet and maybe doesn't have to?

packages/evasive-transform/src/transform-ast.js
```diff
@@ -71,7 +73,11 @@ export function transformAst(
         evadeStrings(p);
         evadeTemplates(p);
         evadeRegexpLiteral(p);
+        evadeClassMethod(p);
         evadeDecrementGreater(p);
+        if (customVisitor) {
+          customVisitor(p);
+        }
```
gibson042: IIRC, we now allow optional chaining.
```suggestion
        customVisitor?.(p);
```

packages/evasive-transform/test/evade-censor.test.js
```diff
@@ -157,6 +157,10 @@ const evadeThat = `
 
     const re = /import (.*)/g;
     const reHtmlAlike = /<!--.*-->/g;
+
+    class Foo {
+      import() { return 'yes'; }
+    }
```
gibson042, line 166: Adding a static method named `import`:
```suggestion
    class Foo {
      static import() { return 'static method'; }
      import() { return 'instance method'; }
    }
```

packages/evasive-transform/test/evade-censor.test.js
```diff
@@ -318,6 +322,14 @@ test('evadeCensor() - templates - only last quasi has pattern', async t => {
   t.is(code, 'const x = `safe text ${a}im${""}port(`');
 });
 
+test('evadeCensor() - class method named import is evaded', async t => {
+  const { code } = evadeCensorSync(`class Foo { import() { return 'yes'; } }`, {
+    sourceType: 'module',
+  });
```
gibson042:
```suggestion
  const { code } = evadeCensorSync(
    `class Foo {
      static import() { return 'static method'; }
      import() { return 'instance method'; }
    }`,
    {
      sourceType: 'module',
    },
  );
```

packages/evasive-transform/src/transform-code.js
```diff
@@ -226,3 +226,19 @@ export const evadeDecrementGreater = p => {
     };
   }
 };
+
+/**
+ * @param {import('@babel/traverse').NodePath} p
+ */
+export const evadeClassMethod = p => {
```
gibson042: I don't think it makes sense to transform class methods without also transforming object methods.
```suggestion
export const evadeMethod = p => {
```

gibson042 (commented):
This looks good to me, but I think we should subject class methods and object methods to equal treatment (and regardless, should have test coverage for both instance and static methods).

boneskull (approved):
LGTM, but I like @gibson042's suggestions

packages/evasive-transform/src/transform-ast.js
naugtur: I don't like the fact that it's harder to spot that way, but ok

packages/evasive-transform/src/transform-code.js
naugtur: I mean, we should probably prevent 
```js
class Foo {
  eval(x) {}
}
```
from getting mangled by the string replacement we use all over the place, because
```js
class Foo {
  (0,eval)(x) {}
}
```
is a syntax error.

I'll just commit that here and see if anybody opposes it.

packages/evasive-transform/src/transform-code.js
naugtur: you tricked me, it was not replacing object methods after just the name change! :D

packages/evasive-transform/test/snapshots/evade-censor.test.js.md
```diff
@@ -221,6 +232,17 @@ Generated by [AVA](https://avajs.dev).
...
+        class Foo {␊
+          static["import"](){return'static method'; }␊
+          ["import"](){return'instance method';}␊
+          ["import"](x){throw Error(x+' not found')}␊
+          ["eval"](x){return'I refuse to run: '+x;}␊
+        }␊
+    ␊
+        const notAClass = {␊
+          ["import"](){return'this is fine';}␊
+        };␊
```
naugtur, line 245: This is what this change is doing now

boneskull:
@naugtur needs conflict resolution

naugtur:
> @naugtur needs conflict resolution

I'm disappointed diff didn't figure it out.

.changeset/sour-cherries-fetch.md
```diff
@@ -0,0 +1,6 @@
+---
+"@endo/evasive-transform": minor
+---
+
+- Add `customVisitor` option - a visitor function to be called on each node of the Babel traverse, in addition to the standard transforms in evasive-transform. Receives the same path argument as a normal Babel visitor.
+- Add an easion for a class or object method being named `import` or `eval`
```
boneskull:
```suggestion
- Add an evasion for a class or object method being named `import` or `eval`
```

Issue #2392, opened 2024-07-31
patterns: Constructing a pattern should also create a static type


gibson042:
## What is the Problem Being Solved?

Independently constructing runtime patterns with `M` matcher methods and defining static types with `@typedef` is cumbersome and error-prone, leading us to implement manual validation even when `M` would be useful (cf. https://github.com/Agoric/agoric-sdk/pull/9169#discussion_r1690588997). It should instead be possible to extract static types from patterns, e.g.
```js
import { mustMatch } from '@endo/patterns';

const BundleShape = M.or(…);

/**
 * @param {TypeFromPattern<BundleShape>} bundle
 */
const evaluateBundle = bundle => {
  mustMatch(bundle, BundleShape);
  const { moduleFormat } = bundle;
  …
};

evaluateBundle({}); // static error
```

## Description of the Design

Annotate Matcher functions to return annotated values, and define a utility type for reading such annotations into types.

## Security Considerations

None AFAIK

## Scaling Considerations

None AFAIK

## Test Plan

We should take examples from existing endo and agoric-sdk code.

## Compatibility Considerations

None AFAIK

## Upgrade Considerations

Nothing out of the ordinary, just the usual caveat that functionality is not incorporated into mainnet vats until they are upgraded.

gibson042:
[...]

PR #3133, opened 2026-03-22
infer TypeScript types from pattern guards

turadg:
closes: #2392 
supports: https://github.com/Agoric/agoric-sdk/issues/6160

## Summary

Add comprehensive TypeScript type inference from pattern guards (`M.*` matchers) to exo methods and interfaces. When `makeExo`, `defineExoClass`, and `defineExoClassKit` are called with typed `InterfaceGuard` values, method signatures are now inferred from the guard at compile time.

### Core Features
- **`TypeFromPattern<P>`** — infer static types from any pattern matcher (string, bigint, remotable, arrays, records, etc.)
- **`TypeFromMethodGuard<G>`** — infer function signatures from `M.call()` or `M.callWhen()` guards
- **`TypeFromInterfaceGuard<G>`** — infer method records from interface guard definitions
- **`M.remotable<typeof Guard>()`** — facet-isolated return types with guard polymorphism
- **Per-facet `ThisType`** in `defineExoClassKit` — `this.facets` and `this.state` properly typed

### Breaking Changes
- `defineExoClassKit` now requires facet method types to match their guard signatures (compile-time check)
- `makeExo` and `defineExoClass` enforce method signatures against guard at compile time
- Single-facet exos have `this.self` (not `this.facets`); multi-facet kits have `this.facets` (not `this.self`)

### Type System Architecture
- New `types-index.d.ts` / `types-index.js` convention for re-exporting values with enhanced type signatures
- `.ts` files (e.g. `type-from-pattern.ts`) contain type definitions only; no runtime code
- `JSDoc` `@import` for type-only imports in `.js` files

### Documentation
- **`AGENTS.md`** — TypeScript conventions for monorepo contributors
- Enhanced JSDoc in `src/types.d.ts` explaining `ClassContext` vs `KitContext`
- Comprehensive type test suite in `types.test-d.ts` covering all inference cases

### Tests
- Full runtime test coverage for kit facet isolation and `this` context
- Type-level tests (tsd) for guard-driven method inference
- Tests for `M.callWhen`, `M.await`, `M.promise`, and async method returns

### DIFF
--- a/packages/patterns/src/patterns/patternMatchers.js
+++ b/packages/patterns/src/patterns/patternMatchers.js
@@ -924,7 +924,8 @@ const makePatternKit = () => {
 
   /** @type {MatchHelper<string>} */
   const matchKindHelper = Far('match:kind helper', {
-    confirmMatches: confirmKind,
+    confirmMatches: (specimen, kind, reject) =>
+      confirmKind(specimen, /** @type {Kind} */ (kind), reject),
 
     confirmIsWellFormed: (allegedKeyKind, reject) =>
       typeof allegedKeyKind === 'string' ||
@@ -1505,7 +1506,7 @@ const makePatternKit = () => {
    * throwing an error).
    *
    * @typedef {CopyArray | CopySet | CopyBag} Container
-   * @param {Container} specimen
+   * @param {Passable} specimen
    * @param {Pattern} elementPatt
    * @param {bigint} bound Must be >= 1n
    * @param {Rejector} reject
@@ -1581,28 +1582,22 @@ const makePatternKit = () => {
 
   /** @type {MatchHelper<[Pattern, bigint, Limits?]>} */
   const matchContainerHasHelper = Far('M.containerHas helper', {
-    /**
-     * @param {CopyArray | CopySet | CopyBag} specimen
-     * @param {[Pattern, bigint, Limits?]} payload
-     * @param {Rejector} reject
-     */
     confirmMatches: (
       specimen,
       [elementPatt, bound, limits = undefined],
       reject,
     ) => {
-      const kind = confirmKindOf(specimen, reject);
+      confirmKindOf(specimen, reject);
       const { decimalDigitsLimit } = limit(limits);
-      if (
-        !applyLabelingError(
-          confirmDecimalDigitsLimit,
-          [bound, decimalDigitsLimit, reject],
-          `${kind} matches`,
-        )
-      ) {
+      if (!confirmDecimalDigitsLimit(bound, decimalDigitsLimit, reject)) {
         return false;
       }
-      return !!containerHasSplit(specimen, elementPatt, bound, reject);
+      return !!containerHasSplit(
+        /** @type {Container} */ (specimen),
+        elementPatt,
+        bound,
+        reject,
+      );
     },
 
     confirmIsWellFormed: (payload, reject) =>
@@ -1745,16 +1740,22 @@ const makePatternKit = () => {
     },
 
     /**
-     * @param {Array} splitArray
+     * @param {Passable} splitArray
      * @param {Rejector} reject
      */
     confirmIsWellFormed: (splitArray, reject) => {
-      if (
-        passStyleOf(splitArray) === 'copyArray' &&
-        (splitArray.length >= 1 || splitArray.length <= 3)
-      ) {
+      if (passStyleOf(splitArray) === 'copyArray') {
+        const typedSplitArray = /** @type {CopyArray<Passable>} */ (splitArray);
+        if (!(typedSplitArray.length >= 1 && typedSplitArray.length <= 3)) {
+          return (
+            reject &&
+            reject`Must be an array of a requiredPatt array, an optional optionalPatt array, and an optional restPatt: ${q(
+              splitArray,
+            )}`
+          );
+        }
         const [requiredPatt, optionalPatt = undefined, restPatt = undefined] =
-          splitArray;
+          typedSplitArray;
         if (
           isPattern(requiredPatt) &&
           passStyleOf(requiredPatt) === 'copyArray' &&
@@ -1861,16 +1862,22 @@ const makePatternKit = () => {
     },
 
     /**
-     * @param {Array} splitArray
+     * @param {Passable} splitArray
      * @param {Rejector} reject
      */
     confirmIsWellFormed: (splitArray, reject) => {
-      if (
-        passStyleOf(splitArray) === 'copyArray' &&
-        (splitArray.length >= 1 || splitArray.length <= 3)
-      ) {
+      if (passStyleOf(splitArray) === 'copyArray') {
+        const typedSplitArray = /** @type {CopyArray<Passable>} */ (splitArray);
+        if (!(typedSplitArray.length >= 1 && typedSplitArray.length <= 3)) {
+          return (
+            reject &&
+            reject`Must be an array of a requiredPatt record, an optional optionalPatt record, and an optional restPatt: ${q(
+              splitArray,
+            )}`
+          );
+        }
         const [requiredPatt, optionalPatt = undefined, restPatt = undefined] =
-          splitArray;
+          typedSplitArray;
         if (
           isPattern(requiredPatt) &&
           passStyleOf(requiredPatt) === 'copyRecord' &&
--- a/packages/common/object-meta-map.js
+++ b/packages/common/object-meta-map.js
@@ -27,9 +27,9 @@ const { ownKeys } = Reflect;
  *
  * @template {Record<PropertyKey, any>} O
  * @param {O} original
- * @param {(
- *   desc: TypedPropertyDescriptor<O[keyof O]>,
- *   key: keyof O
+ * @param {<K extends keyof O>(
+ *   desc: TypedPropertyDescriptor<O[K]>,
+ *   key: K
  * ) => (PropertyDescriptor | undefined)} metaMapFn
  * @param {any} [proto]
  * @returns {any}
@@ -41,11 +41,17 @@ export const objectMetaMap = (
 ) => {
   const descs = getOwnPropertyDescriptors(original);
   const keys = ownKeys(original);
+  /**
+   * Preserve the per-key descriptor type when calling `metaMapFn`.
+   *
+   * @template {keyof O} K
+   * @param {K} key
+   * @returns {[K, PropertyDescriptor | undefined]}
+   */
+  const mapDesc = key => [key, metaMapFn(descs[key], key)];
 
   const descEntries = /** @type {[PropertyKey,PropertyDescriptor][]} */ (
-    keys
-      .map(key => [key, metaMapFn(descs[key], key)])
-      .filter(([_key, optDesc]) => optDesc !== undefined)
+    keys.map(mapDesc).filter(([_key, optDesc]) => optDesc !== undefined)
   );
   return harden(create(proto, fromEntries(descEntries)));
 };
--- a/packages/marshal/src/encodePassable.js
+++ b/packages/marshal/src/encodePassable.js
@@ -568,7 +568,7 @@ const makeEncodeError = (unsafeEncodeError, verifyEncoding) => {
 
 /**
  * @param {(str: string) => string} encodeStringSuffix
- * @param {(arr: unknown[], encodeRecur: (p: Passable) => string) => string} encodeArray
+ * @param {(arr: Passable[], encodeRecur: (p: Passable) => string) => string} encodeArray
  * @param {Required<EncodeOptions> & {verifyEncoding?: (encoded: string, label: string) => void}} options
  * @returns {(p: Passable) => string}
  */
--- a/packages/patterns/src/keys/compareKeys.js
+++ b/packages/patterns/src/keys/compareKeys.js
@@ -16,7 +16,8 @@ import {
 } from './checkKey.js';
 import { makeCompareCollection } from './keycollection-operators.js';
 
-/** @import {CopySet, Key, KeyCompare} from '../types.js' */
+/** @import {CopyBag, CopyMap, CopySet, Key, KeyCollection, KeyCompare} from '../types.js' */
+/** @import {Passable} from '@endo/pass-style' */
 
 /**
  * CopySet X is smaller than CopySet Y iff all of these conditions hold:
@@ -26,8 +27,14 @@ import { makeCompareCollection } from './keycollection-operators.js';
  * X is equivalent to Y iff the condition 1 holds but condition 2 does not.
  */
 export const setCompare = makeCompareCollection(
-  /** @type {<K extends Key>(s: CopySet<K>) => Array<[K, 1]>} */ (
-    s => harden(getCopySetKeys(s).map(key => [key, 1]))
+  /** @type {(collection: KeyCollection) => Array<[Key, 1]>} */ (
+    collection =>
+      harden(
+        getCopySetKeys(/** @type {CopySet} */ (collection)).map(key => [
+          key,
+          1,
+        ]),
+      )
   ),
   0,
   compareNumerics,
@@ -43,7 +50,9 @@ harden(setCompare);
  * X is equivalent to Y iff the condition 1 holds but condition 2 does not.
  */
 export const bagCompare = makeCompareCollection(
-  getCopyBagEntries,
+  /** @type {(collection: KeyCollection) => Array<[Key, bigint]>} */ (
+    collection => getCopyBagEntries(/** @type {CopyBag} */ (collection))
+  ),
   0n,
   compareNumerics,
 );
@@ -71,7 +80,10 @@ const ABSENT = Symbol('absent');
  */
 // eslint-disable-next-line no-underscore-dangle
 const _mapCompare = makeCompareCollection(
-  getCopyMapEntryArray,
+  /** @type {(collection: KeyCollection) => Array<[Key, Passable]>} */ (
+    collection =>
+      getCopyMapEntryArray(/** @type {CopyMap<Key, Passable>} */ (collection))
+  ),
   ABSENT,
   (leftValue, rightValue) => {
     if (leftValue === ABSENT && rightValue === ABSENT) {
@@ -82,7 +94,10 @@ const _mapCompare = makeCompareCollection(
       return 1;
     } else {
       // eslint-disable-next-line no-use-before-define
-      return compareKeys(leftValue, rightValue);
+      return compareKeys(
+        /** @type {Key} */ (leftValue),
+        /** @type {Key} */ (rightValue),
+      );
     }
   },
 );
--- a/packages/patterns/src/keys/keycollection-operators.js
+++ b/packages/patterns/src/keys/keycollection-operators.js
@@ -10,7 +10,7 @@ import { makeIterator } from '@endo/common/make-iterator.js';
 import { makeArrayIterator } from '@endo/common/make-array-iterator.js';
 
 /**
- * @import {FullCompare, RankCompare} from '@endo/marshal'
+ * @import {FullCompare, PartialCompare, RankCompare} from '@endo/marshal'
  * @import {Key, KeyCompare, KeyComparison, KeyCollection} from '../types.js'
  */
 
@@ -183,7 +183,7 @@ harden(generateCollectionPairEntries);
  * @template [V=unknown]
  * @param {(collection: C) => Array<[Key, V]>} getEntries
  * @param {V} absentValue
- * @param {KeyCompare} compareValues
+ * @param {PartialCompare<V>} compareValues
  * @returns {(left: C, right: C) => KeyComparison}
  */
 export const makeCompareCollection = (getEntries, absentValue, compareValues) =>
--- a/packages/patterns/src/patterns/types.ts
+++ b/packages/patterns/src/patterns/types.ts
@@ -49,7 +49,7 @@ export type MatchHelper<P extends Passable = Passable> = {
 export type PatternKit = {
   confirmMatches: (
     specimen: any,
-    patt: Passable,
+    patt: Pattern,
     reject: Rejector,
     label?: string | number,
   ) => boolean;


packages/patterns/src/types.ts
```diff
@@ -272,98 +278,124 @@ export type PatternMatchers = {
...
+  ) => MatcherOf<'string', T>;
 
   /**
    * Matches any registered or well-known symbol,
    * subject to limits.
    */
-  symbol: (limits?: Limits) => Matcher;
+  symbol: <T extends symbol = symbol>(
+    limits?: Limits,
+  ) => MatcherOf<'symbol', T>;
```
erights, line 330: No change suggested. 

Just noting that I'm glad to be reminded of this, since we're in progress with changing what a passable symbol is.

packages/patterns/src/types.ts
```diff
@@ -272,98 +278,124 @@ export type PatternMatchers = {
...
-  remotable: (label?: string) => Matcher;
+   *
+   * For facet-isolated return types in exo kits, pass an InterfaceGuard
+   * as the type parameter:
+   * ```ts
+   * const PublicI = M.interface('Public', {
+   *   getData: M.call().returns(M.string()),
+   * });
+   * const AdminI = M.interface('Admin', {
+   *   getPublic: M.call().returns(M.remotable<typeof PublicI>('Public')),
```
erights, line 374: This is awesome as it expresses what anyone using `M.remotable` expects to express but still keeps a clear and locally visible separation of what types are enforced/sound and which are not.

packages/patterns/src/types.ts
```diff
@@ -272,98 +278,124 @@ export type PatternMatchers = {
...
+  map: (limits?: Limits) => MatcherOf<'kind', 'copyMap'>;
 
   /**
    * Matches a far object or its remote presence.
    * The optional `label` is purely for diagnostic purposes and does not
    * add any constraints.
-   */
-  remotable: (label?: string) => Matcher;
+   *
+   * For facet-isolated return types in exo kits, pass an InterfaceGuard
```
erights, line 367: What are "facet-isolated return types" ?

packages/patterns/src/types.ts
erights: Curious: When using the type parameter, do you expect to also include the string parameter? I ask because I made the mistake of giving `remotable` an unenforced string parameter but not `M.promise`. Your type parameterization of these looks similar, which is great. But I wonder whether we should better align them by deprecating the `remotable` string parameter, or by adding a `promise` string parameter.

In my tentative opinion, I think both should have a string parameter, as it is useful in runtime diagnostics.

packages/patterns/src/types.ts
```diff
@@ -384,58 +416,64 @@ export type PatternMatchers = {
   /**
    * Matches any value that compareKeys reports as less than rightOperand.
    */
-  lt: (rightOperand: Key) => Matcher;
+  lt: (rightOperand: Key) => MatcherOf<'lt'>;
```
erights: For all of these, does


```suggestion
  lt: (rightOperand: Key) => MatcherOf<'lt', Key>;
```

make sense? If so, what about

```suggestion
  lt: <K extends Key = Key>(rightOperand: K) => MatcherOf<'lt', K>;
```

?

packages/patterns/src/types.ts
```diff
@@ -454,14 +496,18 @@ export type PatternMatchers = {
     elementPatt?: Pattern,
```
erights, line 495: No change suggested.

Just checking my understanding: The reason you did not try to parameterize the `Pattern` is because it would need to be overloaded, and an overloaded type parameter would not be very useful. Yes?

packages/patterns/src/types.ts
```diff
@@ -475,11 +521,11 @@ export type PatternMatchers = {
...
    */
-  splitArray: (
-    required: Pattern[],
-    optional?: Pattern[],
+  splitArray: <Req extends Pattern[], Opt extends Pattern[] = []>(
+    required: [...Req],
+    optional?: [...Opt],
     rest?: Pattern,
-  ) => Matcher;
+  ) => MatcherOf<'splitArray', [Req, Opt]>;
```
erights: The `rest` is part of the payload. Should it be included in the tuple?

Likewise below.

packages/patterns/src/types.ts
```diff
@@ -475,11 +521,11 @@ export type PatternMatchers = {
    * Any array elements beyond the summed length of `required` and `optional`
    * are collected and matched against `rest`.
    */
-  splitArray: (
-    required: Pattern[],
-    optional?: Pattern[],
+  splitArray: <Req extends Pattern[], Opt extends Pattern[] = []>(
```
erights: Curious why you did not give `Req` a default? Likewise below.

packages/patterns/src/types.ts
```diff
@@ -538,12 +587,16 @@ export type PatternMatchers = {
...
+    subPatt: P,
+  ) => MatcherOf<'or', [P, MatcherOf<'promise'>]>;
 
   /**
    * Matches any Passable that is matched by `subPatt` or is the exact value `undefined`.
    */
-  opt: (subPatt: Pattern) => Pattern;
+  opt: <P extends Pattern>(
+    subPatt: P,
+  ) => MatcherOf<'or', [P, MatcherOf<'kind', 'undefined'>]>;
```
erights, line 605: I don't understand the `MatcherOf<'kind', 'undefined'>`

packages/patterns/src/types.ts
```diff
@@ -621,13 +674,17 @@ export type GuardMakers = {
    * Guard a synchronous call. Arguments not guarded by `M.raw()` are
    * automatically hardened and must be at least Passable.
    */
-  call: (...argPatterns: SyncValueGuard[]) => MethodGuardMaker;
+  call: <A extends SyncValueGuard[]>(
```
erights, line 683: Should the return type somehow appear? Likewise for `callWhen`.

packages/patterns/src/patterns/getGuardPayloads.js
erights, line 44: Are we there yet?

packages/patterns/src/type-from-pattern.ts
```diff
@@ -0,0 +1,380 @@
...
+ * ```ts
+ * const shape = M.splitRecord({ name: M.string(), age: M.nat() });
+ * type Person = TypeFromPattern<typeof shape>;
+ * // { name: string; age: bigint }
+ * ```
+ */
+export type TypeFromPattern<P> =
+  P extends CopyTagged<`match:${infer K}`, infer Payload>
+    ? TFDispatch<K, Payload>
+    : P extends readonly [infer H, ...infer T]
```
erights, line 43: What is special about the zeroth element of a CopyArray?

erights: What about empty CopyArrays?

packages/patterns/src/type-from-pattern.ts
```diff
@@ -0,0 +1,380 @@
...
+
+/** Recursively map a tuple of patterns to a tuple of inferred types. */
+type TFTuple<T extends readonly any[]> =
+  T extends readonly [infer H, ...infer R]
+    ? [TypeFromPattern<H>, ...TFTuple<R>]
+    : [];
+
+/**
+ * Leaf matcher lookup table.
+ * These matchers return their Payload directly or a fixed type, with no
```
erights: Please keep static concepts distinct from dynamic runtime concepts


```suggestion
 * These matcher types return their Payload type directly or a fixed type, with no
```

packages/patterns/src/type-from-pattern.ts
```diff
@@ -0,0 +1,380 @@
...
+  string: Payload;
+  number: Payload;
+  bigint: Payload;
+  nat: Payload;
+  symbol: Payload;
+  scalar: ScalarKey;
+  key: Key;
+  pattern: Pattern;
+  not: Passable;
+  // Comparison matchers — can only narrow to Key
```
erights: Am I confused or should this be "can only widen to Key" ?

packages/patterns/src/type-from-pattern.ts
```diff
@@ -0,0 +1,380 @@
...
+type TFKindMap = {
+  boolean: boolean;
+  undefined: undefined;
+  null: null;
+  error: Error;
+  number: number;
+  bigint: bigint;
+  string: string;
+  symbol: symbol;
+  byteArray: ArrayBuffer;
```
erights: No change suggested.

Just noting that we expect to change the type of `byteArray` to `Uint8Array` (or possibly `readonly Uint8Array` ?).

packages/patterns/src/type-from-pattern.ts
```diff
@@ -0,0 +1,380 @@
...
+/** The `M` pattern-matcher namespace value. */
+// eslint-disable-next-line import/export
+export declare const M: MatcherNamespace;
+
+/**
+ * Namespace merged with the `M` value export so users can write
+ * `M.infer<typeof pattern>`, analogous to Zod's `z.infer<typeof schema>`.
+ */
+// eslint-disable-next-line @typescript-eslint/no-namespace, no-redeclare, import/export
+export namespace M {
```
erights, line 374: I remember when we had both a `Remotable` type and a `Remotable` runtime value. The collision kept causing annoyance so we renamed the type to `RemotableObject`. I thought that since then we try to avoid such collisions. Why is having an `M` type ok?

packages/patterns/src/type-from-pattern.ts
```diff
@@ -0,0 +1,380 @@
...
+ * @param patt - The pattern to match against.
+ *
+ * @example
+ * ```ts
+ * if (matches(value, M.string())) {
+ *   value; // narrowed to string
+ * }
+ * ```
+ */
+export declare function matches<P extends Pattern>(
```
erights, line 406: This is declaring the type of the runtime `matches` function, right? This is a really cool payoff (likewise with `mustMatch` below).

I am surprised you can separate a function's type declaration from the function's declaration in this way. But even if we can, is it a good idea?

packages/patterns/src/type-from-pattern.ts
erights: I also see that we're using `M` as a type parameter in many places. The collision of these with the runtime `M` seems unproblematic. 

What about the collision of `M` type parameters with this exported namespace type `M` ?

packages/exo/types-index.d.ts
```diff
@@ -0,0 +1,136 @@
...
+  FacetName,
+} from './src/types.js';
+
+/**
+ * Create a singleton exo object whose methods are type-checked against
+ * the InterfaceGuard at compile time.
+ *
+ * When a typed InterfaceGuard is provided (built with `M.call(...).returns(...)`),
+ * the `methods` parameter must satisfy the inferred method signatures.
+ * The return type preserves the concrete method types from `methods`.
```
erights: This sentence leaves me puzzle about whether the return type in inferred from the concrete method or whether it is inferred from the guard and enforced on the concrete method. (I hope the latter)

packages/exo/types-index.d.ts
```diff
@@ -0,0 +1,136 @@
...
+ *   bar: M.call(M.string()).returns(M.nat()),
+ * });
+ * const foo = makeExo('Foo', FooI, {
+ *   bar(name) { // name: string
+ *     return 0n; // must return bigint
+ *   },
+ * });
+ * ```
+ */
+// TS limitation: When the guard is `InterfaceGuard<any>` (e.g., from
```
erights: This seems to me to be a fundamental limitation, not a TS limitation. How could a hypothetical better type system surmount this limitation?

packages/exo/types-index.d.ts
```diff
@@ -0,0 +1,136 @@
...
+// sloppy mode with `defaultGuards: 'passable'`), TypeFromInterfaceGuard
+// produces a broad index-signature type, and M is inferred from the
+// methods parameter.  Guard-driven enforcement only works when the
+// guard carries specific MethodGuard type parameters.
+//
+// TS limitation: Excess property checking does not apply in generic
+// contexts, so passing extra methods not in the guard is not a
+// compile-time error.  If TS adds exact types for generics, this
+// could be tightened.
```
erights: It could be tightened only if your types are cognizant of `defaultGuards`. If `defaultGuards: 'passable'`, then you must allow "extra" methods.

packages/exo/types-index.d.ts
```diff
@@ -0,0 +1,136 @@
...
+>(
+  tag: string,
+  interfaceGuard: G,
+  methods: M & ThisType<{ self: Guarded<M>; state: {} }>,
+  options?: FarClassOptions<ClassContext<{}, M>>,
+): Guarded<M>;
+
+export declare function makeExo<M extends Methods>(
+  tag: string,
+  interfaceGuard: undefined,
```
erights, line 71: Does `makeExo(tag, undefined, methods)` have the same type as 
```js
makeExo(
  tag,
  M.interface(something, undefined, { defaultGuards: 'passable' },
  methods);
```
? 
I ask because this is the runtime meaning of omitting the interfaceGuard argument from `makeExo`. Likewise for all similar overloads below.

packages/patterns/src/types.ts
turadg: In a `defineExoClassKit`, when facet A's method should return facet B's object, you want the TS return type to be the full B-facet interface, not just bare `RemotableObject`. By writing `M.remotable<typeof PublicI>()` in the guard, TypeFromPattern (via TFRemotable) walks the InterfaceGuard<MG> and resolves a methods-record type. That's what "facet-isolated" means — each facet's return type is independently typed to its own interface shape.

packages/patterns/src/type-from-pattern.ts
turadg: > What is special about the zeroth element of a CopyArray?

This is to handle tuples. The `P extends readonly [infer H, ...infer T]` branch in `TypeFromPattern` handles TypeScript tuple literal types, not CopyArray runtime values. When a caller writes `[M.string(), M.nat()]` (a literal tuple as a pattern), TS infers the type as a tuple `[MatcherOf<'string'>, MatcherOf<'nat'>]`, and this branch maps it element-by-element to `[string, bigint]`. It's not about CopyArray the data structure — actual M.arrayOf(...) patterns come in as CopyTagged and are handled by the `match:arrayOf` branch.

> What about empty CopyArrays?

An empty tuple `[]` doesn't match `[infer H, ...infer T]` (needs at least one element), so it falls through to the identity `P` branch, returning `[]`. That's correct — an empty pattern tuple matches nothing extra.

packages/patterns/src/type-from-pattern.ts
turadg: >I remember when we had both a Remotable type and a Remotable runtime value. The collision kept causing annoyance so we renamed the type to RemotableObject

IIRC, the problem there was they were different shapes. Here, `M` is the namespace for both.

> Why is having an M type ok?

TypeScript allows **value + namespace declaration merging**: `export declare const M: MatcherNamespace` (value) and `export namespace M { type infer<P> = ... }` (namespace) co-exist cleanly. The namespace adds only type-level members (`type infer<P>`), not runtime-accessible properties. This is different from the Remotable/RemotableObject case where both a class (type + value) and another runtime value shared the same name, causing confusion about which was meant at runtime.

>What about the collision of `M` type parameters with this exported namespace type `M` ? 

TypeScript resolves identifiers by scope. An `M extends InterfaceGuard` type parameter shadows the namespace `M` within that generic's scope. Since the namespace `M` is only used at the type level and only for `M.infer<...>`, and type parameters named `M` appear in different scopes (each generic function), there is no runtime collision and the TS compiler correctly distinguishes them.

Worth noting in a comment?

packages/patterns/src/type-from-pattern.ts
turadg: > This is declaring the type of the runtime matches function, right?

Right.

> is it a good idea?

Good question. Claude says: it's the standard pattern for augmenting JS runtime exports with richer TS types than JSDoc can express. The declare function in a .ts / .d.ts file overlays the type seen by consumers without emitting duplicate code. The runtime implementation stays in patternMatchers.js. Many major libraries (Zod, Effect) use this approach. "The key invariant to maintain: the declared signature must be a subtype of the runtime behavior."

And while that invariant would be easier to maintain as a single .ts file, it's a big lift right now because `patternMatcher.js` has `ts-nocheck`. I think this is the right approach for the scope of this PR and a separate PR could refactor the .js/.ts and build steps to make that work.

Meanwhile there is one improvement I'll make:
```ts
// In types-index.ts — verify runtime impl is assignable to base signature
import { matches as _m } from './patternMatchers.js';
// Widened (non-predicate) base check — if sig drifts this line will error
const _matchesCompat: (specimen: unknown, patt: Pattern) => boolean = _m;
```

This is a lightweight guard that catches any future signature divergence at the types-index boundary, since `types-index.js` is not `@ts-nocheck`-ed. The declare function overlay in `type-from-pattern.ts` is then a safe, intentional narrowing (adding the type predicate) on top of a verified base.

packages/patterns/src/types.ts
turadg: `opt(P`) desugars to `M.or(P, M.undefined())`. `M.undefined()` returns `MatcherOf<'kind', 'undefined'>`. In `TFKindMap`, key `'undefined'` maps to the JS type `undefined`. So `TypeFromPattern` for the whole or produces `TypeFromPattern<P> | undefined`. The `MatcherOf<'kind', 'undefined'>` is just `M.undefined()` written in its fully-expanded internal form.

What code comment would help to include? Note that the above was generated by Claude so that's one option for answering the question in the future.

packages/patterns/src/types.ts
turadg: The return guard is captured when `.returns(RG)` is called — `MethodGuardMaker<CK, A>.returns<RG>(...)` → `MethodGuard<CK, A, OptArgs, RG>`. At `M.call(...)` time we only have the arg guards; there is no return guard yet. Fully-typed inference happens only once `.returns()` is finalized.

packages/patterns/src/patterns/getGuardPayloads.js
turadg: Not TMK. This PR is purely additive TS types.

packages/patterns/src/types.ts
turadg: Agreed, both should have a label string param for runtime diagnostics.

We could add `label?: string` to `M.promise<T>(label?: string)` signature to match `M.remotable<T>(label?: string)`. Do you want that in this PR?

packages/patterns/src/types.ts
turadg: The first suggestion `MatcherOf<'lt', Key>` (explicit Payload = Key) is accurate and we can adopt it for clarity. 

The second `<K extends Key = Key>` would be incorrect semantics. The comparison matchers don't narrow the matched value to the operand's type — `M.lt(5n)` matches any Key that is less than 5n, which could include strings in the compareKeys ordering,

I'll update lt/lte/eq/neq/gte/gt to MatcherOf<'lt', Key>.

packages/patterns/src/types.ts
turadg: Good catch!
I'll:
- Add optional 3rd element: `MatcherOf<'splitArray', [Req, Opt, Rest?]> `(or use `MatcherOf<'splitArray', [Req, Opt] | [Req, Opt, Rest]>`)
- Update `TFSplitArray` to accept and spread the rest type
- Same for `splitRecord` with `rest`

packages/patterns/src/types.ts
turadg: Oversight. I'll add `Req extends Pattern[] = Pattern[]` default to avoid callers needing to specify it explicitly.

packages/patterns/src/type-from-pattern.ts
turadg: Good catch! I'll revise to,
```suggestion
  // Comparison matchers — can only infer as broad as Key (operand type is not preserved)
```

packages/exo/types-index.d.ts
turadg: You're absolutely right. With Claude's help to see why: when `defaultGuards: 'passable'`, the guard accepts any method, so `InterfaceGuard<any>` is the correct representation.

I'll reword: "Expected behavior: when the guard uses `defaultGuards: 'passable'`, `TypeFromInterfaceGuard` produces a broad index-signature, and guard-driven enforcement is intentionally disabled."

packages/exo/types-index.d.ts
turadg: True. I'll add a TODO to make the strict overload (`MakeInterfaceGuardStrict`) infer a type that rejects extra methods, while the sloppy overload (`MakeInterfaceGuardSloppy`) allows them.

packages/exo/types-index.d.ts
turadg: They're semantically equivalent at runtime, but the typed overload with `defaultGuards: 'passable'` allows weaker static checking. A follow-up could collapse these into one via better overload.

packages/patterns/src/types.ts
erights: Yes please. If so, would that be the only runtime change in this PR?

packages/patterns/src/types.ts
erights: > any Key that is less than 5n, which could include strings in the compareKeys ordering

Good point! I missed that.

packages/patterns/src/types.ts
turadg: there's now one other: https://github.com/endojs/endo/pull/3133#discussion_r2983584780

I'll plan to include that change too.

packages/patterns/src/types.ts
erights: Should Opt's default also be `Pattern[]` rather than `[]`?

packages/patterns/src/types.ts
erights: > What code comment would help to include?

Your explanation above would be fine.

packages/patterns/src/type-from-pattern.ts
erights: > Worth noting in a comment?

please.

packages/exo/types-index.d.ts
erights: > A follow-up could collapse these into one via better overload.

"follow-up" meaning separate PR? Is it much trouble to include in this PR?

.changeset/infer-pattern-types.md
```diff
@@ -0,0 +1,17 @@
...
+
+- `TypeFromPattern<P>` — infer static types from any pattern matcher
+- `TypeFromMethodGuard<G>` — infer function signatures from `M.call()` / `M.callWhen()` guards
+- `TypeFromInterfaceGuard<G>` — infer method records from interface guard definitions
+- `M.remotable<typeof Guard>()` — facet-isolated return types in exo kits
+- `M.infer<typeof pattern>` — namespace shorthand analogous to `z.infer`
+- `matches` and `mustMatch` now narrow the specimen type via type predicates
+- `makeExo`, `defineExoClass`, and `defineExoClassKit` enforce method signatures against guards at compile time
+
+These are compile-time type changes only; there are no runtime behavioral changes.
```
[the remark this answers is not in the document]
erights: For both the endo and agoric-sdk repos, we take the stance that prose content of error messages are not semantic changes, and do not require a version bump. Even more benign than static-only changes. Even though it may break tests with such prose as goldens. If it changes any non-test runtime behavior beyond the change to the prose itself, that would be a bug in those clients.

packages/exo/test/heap-classes.test.js
```diff
@@ -141,8 +141,10 @@ test('test defineExoClassKit', t => {
     message:
       'In "decr" method of (Counter down): arg 0?: string "foo" - Must be a number',
   });
-  // @ts-expect-error bad arg
-  t.throws(() => upCounter.decr(3), {
+  // TS limitation: Guarded<M> extends Methods which has an index signature
+  // (Record<PropertyKey, CallableFunction>), so upCounter.decr is not a
+  // type error even though 'decr' is only on the down facet.
```
erights: I'm surprised by this limitation. Is there any reasonable way around it?

packages/patterns/src/type-from-pattern.ts
```diff
@@ -96,7 +95,7 @@ type TFKindMap = {
   bigint: bigint;
   string: string;
   symbol: symbol;
-  byteArray: ArrayBuffer;
+  byteArray: ArrayBuffer; // TODO: update to Uint8Array when @endo/pass-style changes the byteArray type
```
erights, line 99: Thanks

packages/exo/test/heap-classes.test.js
turadg: Claude came up with one: `StripIndexSignature`. Incoming

packages/patterns/src/types.ts
turadg: Right, for the plural "you" because I'm leaning on Claude here. :) 

`containerHas` checks whether *at least N elements* match the pattern — it doesn't constrain the *container's own type*. Compare with `M.arrayOf(P)` where the pattern determines the element type of the whole array (`Array<TypeFromPattern<P>>`). For `containerHas`, knowing that
  "some elements match `M.string()`" doesn't tell you the container is an `Array<string>` — it could be a mixed array, a CopySet, or a CopyBag. So  parameterizing the pattern wouldn't give us a useful narrowing of the container type, just noise on the type signature.

packages/patterns/src/types.ts
turadg: No, defaults are intentionally asymmetric: Req defaults to the widest type (any patterns), while Opt and Rest default to the narrowest.

Changing `Req extends Pattern[] = Pattern[]` to `Req extends Pattern[] = []` would break calls like `M.splitArray([M.string(), M.nat()])` when TypeScript can't infer Req from the argument. The default Pattern[] means "any array of patterns", which is the correct upper-bound fallback. Defaulting to [] would mean "no required elements" — a different semantic that would produce wrong types when the default kicks in.

In contrast, `Opt extends Pattern[] = []` defaults to empty because optional elements genuinely default to "none" when the second argument is omitted. And `Rest extends Pattern = never` defaults to `never` because no rest pattern means "no rest matching".

I've added an inline comment.

packages/patterns/src/type-from-pattern.ts
turadg: I'm going to punt on this and follow up soon with a PR that unifies into on .ts file

packages/exo/types-index.d.ts
turadg: Not too much. After trying this branch HEAD in agoric-sdk I see it's necessary.

packages/exo/test/exo-class-js-class.test.js
```diff
@@ -18,6 +17,9 @@ const DoublerI = M.interface('Doubler', {
   double: M.call(M.lte(10)).returns(M.number()),
 });
 
+// @ts-expect-error XXX makeExo expects a string index signature for the methods
+// argument, but we are passing a class prototype which is an object with method
+// names as keys.
 const doubler = makeExo('doubler', DoublerI, DoublerBehaviorClass.prototype);
```
boneskull:
```suggestion
const doubler = makeExo('doubler', DoublerI, Object.create(DoublerBehaviorClass.prototype));
```

packages/exo/test/exo-class-js-class.test.js
```diff
@@ -39,14 +41,17 @@ test('exo doubler using js classes', t => {
 class DoubleAdderBehaviorClass extends DoublerBehaviorClass {
   doubleAddSelfCall(x) {
     const {
+      // @ts-expect-error no `state`
       state: { y },
+      // @ts-expect-error no `self`
```
boneskull:
```suggestion
  /**
   * @param {number} x
   * @this {import('../src/types.js').ClassContext<{ y: number }, { double: (x: number) => number }>}
   */
  doubleAddSelfCall(x) {
    const {
      state: { y },
```

packages/exo/test/exo-class-js-class.test.js
```diff
@@ -39,14 +41,17 @@ test('exo doubler using js classes', t => {
...
       self,
     } = this;
     return self.double(x) + y;
   }
 
   doubleAddSuperCall(x) {
     const {
+      // @ts-expect-error no `state`
```
boneskull:
```suggestion
  /**
   * @param {number} x
   * @this {import('../src/types.js').ClassContext<{ y: number }, { double: (x: number) => number }>}
   */
  doubleAddSuperCall(x) {
    const {
```

packages/exo/test/exo-class-js-class.test.js
```diff
@@ -63,6 +68,7 @@ const makeDoubleAdder = defineExoClass(
   'doubleAdderClass',
   DoubleAdderI,
   y => ({ y }),
+  // @ts-expect-error no `state`
   DoubleAdderBehaviorClass.prototype,
```
boneskull:
```suggestion
  Object.create(DoubleAdderBehaviorClass.prototype),
```

packages/exo/test/heap-classes.test.js
```diff
@@ -36,6 +37,7 @@ test('callWhen-guarded method called without optional array argument', async t =
       t.is(arr, undefined);
     },
   });
+  // @ts-expect-error TS infers foo(arr) as required from the impl, but guard makes it optional at runtime
```
boneskull: only because it's not told otherwise. maybe stick this above `async foo`?

```js
    /**
     * @param {any[]} [arr]
     */
```

packages/exo/test/heap-classes.test.js
```diff
@@ -324,9 +321,11 @@ test('raw guards', t => {
   t.is(Object.isFrozen({}), Object.isFrozen(greeter2.passthrough({})));
 
   t.true(Object.isFrozen(greeter2.tortuous({}, {}, {}, {}, {})));
+  // @ts-expect-error TS infers 4 required params from impl, guard makes last 2 optional at runtime
```
boneskull: it looks like many of these `@ts-expect-error`s can be avoided by adding docstrings to the `greeter2` definition

packages/exo/src/types.d.ts
```diff
@@ -12,10 +12,36 @@ export type MatchConfig = {
...
+ * - `this.state` — the per-instance state record returned by `init()`
+ *   (empty `{}` for `makeExo` which has no `init`).
+ * - `this.self` — the exo instance itself (the object whose methods you're
+ *   implementing). Useful for passing "yourself" to other code.
+ *
+ * **Not available on kits.** Multi-facet exos use {@link KitContext} instead,
+ * which provides `this.facets` (the record of all facet instances in the
+ * cohort) rather than `this.self`.
+ */
 export type ClassContext<S = any, M extends Methods = any> = {
```
boneskull, line 34: The comment suggests that type arg `S` is a "record", but it can be any value (including `undefined`).  From this, I'm not sure whether or not it's supposed to be an object (likewise with the `state` prop).  Maybe reword to be more clear about the type (and below for `KitContext`)?

From the type definition of `makeExo`, we're expecting the `state` prop to be a non-nullable value, and also demanding `S` is non-nullable:

```ts
export declare function makeExo<
  G extends InterfaceGuard,
  M extends TypeFromInterfaceGuard<G> & Methods,
>(
  tag: string,
  interfaceGuard: G,
  methods: M & ThisType<{ self: Guarded<M>; state: {} }>,
  options?: FarClassOptions<ClassContext<{}, M>>,
): Guarded<M>;
```

Is `makeExo` intended to restrict `S` in `ClassContext<S>`?  Might also be worth mentioning in the docstring.

packages/exo/test/heap-classes.test.js
turadg: I'm testing the impl types coming from the guard, but good point there's a deficiency here. I'll see what I can do.

packages/exo/types-index.d.ts
```diff
@@ -0,0 +1,145 @@
...
+  methods: M & ThisType<{ self: Guarded<M>; state: {} }>,
+  options?: FarClassOptions<ClassContext<{}, M>>,
+): Guarded<M>;
+
+/**
+ * Define an exo class whose methods are type-checked against the InterfaceGuard.
+ */
+export declare function defineExoClass<
+  G extends InterfaceGuard,
+  I extends (...args: any[]) => any,
```
boneskull: If `I` is intended to be _literally any function_, it should be typed as `(...args: readonly any[]) => any`; otherwise it will exclude functions typed as accepting read-only rest params. 

(repeat below for various shades of `I`)

packages/patterns/src/patterns/getGuardPayloads.js
```diff
@@ -69,10 +69,9 @@ const LegacyAwaitArgGuardShape = harden({
  */
 export const getAwaitArgGuardPayload = awaitArgGuard => {
   if (matches(awaitArgGuard, LegacyAwaitArgGuardShape)) {
-    // @ts-expect-error Legacy adaptor can be ill typed
-    const { klass: _, ...payload } = awaitArgGuard;
-    // @ts-expect-error Legacy adaptor can be ill typed
-    return payload;
+    const legacy = /** @type {any} */ (awaitArgGuard);
+    const { klass: _, ...payload } = legacy;
+    return /** @type {AwaitArgGuardPayload} */ (payload);
```
boneskull, line 74: Seems like `matches` should be a type guard, but it ain't.

packages/patterns/test/containerHasSplit.test.js
```diff
@@ -39,14 +39,9 @@ const testContainerHasSplit = test.macro((t, config) => {
...
-    specimen,
-    pattern,
-    bound,
-    Fail,
-    true,
-    true,
-  );
+  const split = containerHasSplit(specimen, pattern, bound, Fail, true, true);
+  assert(split);
+  const [accepted, rejected] = split;
```
boneskull, line 44: could avoid the code change this way

```suggestion
  const [accepted, rejected] = /* @type {Exclude<typeof split, false>} */(split);
```

packages/exo/src/types.d.ts
turadg: yes, `makeExo` is intentionally more restrictive than ClassContext itself. `ClassContext<S>` is a generic type  used across all exo forms — `S` defaults to any so it works as a base type. But `makeExo` has no `init` function, so there's no user-defined state.
   It hardcodes `S = {}` (empty object), meaning `this.state` exists but has no accessible properties. `defineExoClass` is the form that gets a real `S = ReturnType<I>` from the init function. The ClassContext type is a container — the concrete `S` comes from whichever API instantiates it.

packages/patterns/test/containerHasSplit.test.js
turadg: yeah. an earlier version just made it `any` since it's validated below. I went with the least annotated

packages/exo/types-index.d.ts
turadg: Good catch! Added

packages/patterns/src/patterns/getGuardPayloads.js
turadg: It is, just not in the `.js`. That's what the `declare function matches` overlay in `type-from-pattern.ts` does (the type predicate `specimen is TypeFromPattern<P>`). But `getGuardPayloads.js` imports matches directly from `./patternMatchers.js`, bypassing the typed overlay in `types-index.d.ts`.

After this lands I'm going to convert all this package's modules to .ts, which will clean up the indirection. I'm saving that for a separate PR because it gets into the build system.

packages/patterns/test/patterns.test.js
```diff
@@ -913,10 +913,12 @@ test('collection contents rankOrder tie insensitivity', t => {
     [r1, 1],
     [r2, 2],
   ]);
-  const map2 = makeCopyMap([
-    [r2, M.gte(2)],
-    [r1, M.lte(1)],
-  ]);
+  const map2 = makeCopyMap(
```
boneskull, line 916: If `makeCopyMap` had an overload that allowed wider types, then you wouldn't need the type assertion here. e.g.:

```js
/**
 * @template {Key} K
 * @template {Passable} V
 * @overload
 * @param {Iterable<[K, V]>} entries
 * @returns {CopyMap<K,V>}
 */

/**
 * @template {Key} K
 * @overload
 * @param {Iterable<[K, Passable]>} entries
 * @returns {CopyMap<K, Passable>}
 */

/**
 * @template {Key} K
 * @template {Passable} V
 * @param {Iterable<[K, V]>} entries
 */
export const makeCopyMap = entries => {
```

Where `V` is is the same type for all entries, the first overload applies, and we get a more narrowly-typed return value.  But for something like the call site here--where we have a union, but it isn't inferred as such due to `Iterable`--the second overload applies.

If the type was `[K, V][]` instead of `Iterable<K, V>`, we wouldn't have this problem. Alas!

packages/exo/test/types.test-d.ts
```diff
@@ -304,6 +307,200 @@ import type { Guarded, GuardedKit } from '../src/types.js';
...
+
+// M.callWhen with mixed args: M.await + plain pattern
+{
+  const mg = M.callWhen(M.await(M.nat()), M.string()).returns(M.boolean());
+  type Fn = TypeFromMethodGuard<typeof mg>;
+  // first arg awaited (bigint), second passed as-is (string)
+  expectType<(n: bigint, s: string) => Promise<boolean>>(null as unknown as Fn);
+}
+
+// M.callWhen with an exo as awaited argument (M.await(M.remotable<typeof I>()))
```
erights: excellent

packages/patterns/src/patterns/patternMatchers.js
erights, line 374: Tentatively suggesting a change, not requiring it

Would it work to change this to
```suggestion
   * @returns {patt is Pattern}
```
?

If so, could you change
```js
/** @type {WeakSet<Pattern & object>} */
const patternMemo = new WeakSet();
```
above to
```js
/** @type {WeakSet<Pattern>} */
const patternMemo = new WeakSet();
```
?

The reason I ask is that we only put Patterns in `patternMemo`

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -169,7 +168,7 @@ const makePatternKit = () => {
    * Otherwise result undefined.
```
erights: Otherwise result is undefined
or
Otherwise returns undefined

packages/patterns/src/patterns/patternMatchers.js
```diff
@@ -1,4 +1,3 @@
-// @ts-nocheck So many errors that the suppressions hamper readability.
```
erights, line 1: Thanks!

packages/exo/types-index.d.ts
```diff
@@ -69,7 +69,7 @@ export declare function makeExo<M extends Methods>(
  */
 export declare function defineExoClass<
   G extends InterfaceGuard,
-  I extends (...args: any[]) => any,
+  I extends (...args: readonly any[]) => any,
```
erights, line 81: No change suggested.

Why do we say `readonly` in some places where it is correct, but omit it in plenty of other places where it would be correct?

packages/exo/test/types-typed-guarded.test-d.ts
```diff
@@ -70,3 +67,34 @@ import type { GuardedMethods } from '../src/types.js';
   expectType<(step?: bigint) => bigint>(null as unknown as CM['incr']);
   expectType<() => bigint>(null as unknown as CM['read']);
 }
+
+// Limitation: widening a guard's param type in the implementation is NOT
+// caught at compile time, even with strictFunctionTypes.  TypeScript only
+// applies strict (contravariant) checking to arrow-syntax function types
+// `(x: T) => U`, not to method shorthand `method(x: T): U` — methods
```
erights: No change suggested.

Wow. I had no idea. Any idea why?

packages/exo/test/exo-class-js-class.test.js
erights: My own esthetics mildly favor
```js
{ __proto__: DoublerBehaviorClass.prototype }
```
over
```js
Object.create(DoublerBehaviorClass.prototype));
```

since the first depends only on syntax(*), not on the current binding of `Object.create`. Remember, the `__proto__` in this position is a context-dependent keyword, not a property name.

packages/exo/test/exo-class-js-class.test.js
erights: Please avoid inline `import` type expressions. Instead

```js
/**
 * @import {ClassContext} from '../src/types.js;
 */

...

    * @this {ClassContext...
```

And of course everywhere else this applies

packages/exo/types-index.d.ts
turadg: Here we do it so `I` function slot wide allows functions that have `readonly` params. Pointed out here https://github.com/endojs/endo/pull/3133#discussion_r2984563382

There may be improvements yet around this aspect but I opt to defer them and first get integration with agoric-sdk.

packages/exo/test/types-typed-guarded.test-d.ts
turadg: https://github.com/microsoft/TypeScript/wiki/FAQ#why-are-function-parameters-bivariant

turadg: If you want, in a later PR we could add an opt-in for function strictness:

```ts
export type StrictifyMethods<M> = {
  [K in keyof M]:
    M[K] extends (...args: infer A) => infer R
      ? (...args: A) => R
      : never;
};

export const defineExoMethods = <G extends InterfaceGuard<any>>(
  _guard: G,
  methods: StrictifyMethods<TypeFromInterfaceGuard<G>>,
): TypeFromInterfaceGuard<G> => methods as TypeFromInterfaceGuard<G>;
```

Then when defining an Exo,
```ts
const exo = makeExo(
  'Iface',
  guard,
  // extra code to opt into strictness
  defineExoMethods(guard, {
    foo(x: 'a') {},
  }),
);
```

packages/patterns/src/patterns/patternMatchers.js
turadg: But not all Patterns are valid weak keys. Another option is,
```ts
/** @type {WeakSet<Exclude<Pattern, undefined | null | string | number | bigint | boolean| void>>} */
```

but I think the `& object` is better.

erights: ah. i misunderstood the purpose.

packages/exo/test/types-typed-guarded.test-d.ts
erights: Let's remember the possibility, but not do it until we encounter an actual need.

PR #3137, opened 2026-03-25
feat: support .ts runtime modules via erasable type syntax

turadg:
Refs: #2864

## Description

Adds infrastructure for authoring Endo package source in TypeScript using [erasable type syntax](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-8.html#the---erasablesyntaxonly-option) (types removable without changing runtime semantics — no `enum`/`namespace`/`const enum`/parameter properties). Published packages continue to ship `.js`.

Type stripping is handled per context:

- **Dev/test runtime:** Node.js ≥22 native type stripping (repo minimum is Node 22; CI runs 22.x/24.x). No loader required.
- **Bundling (build time):** [`amaro`](https://github.com/nodejs/amaro) `transformSync` strip-only — e.g. the `@endo/benchmark` rollup config, matching `@endo/bundle-source`.
- **Publishing (pack time):** [`ts-node-pack`](https://github.com/turadg/ts-node-pack) erases `.ts`→`.js` and rewrites relative `.ts` specifiers (`yarn pack:all`).
- **`@endo/cli` / `@endo/daemon` production runtime:** load `.ts` via `amaro/strip`.

### Migrations (proof of approach)

- `@endo/eventual-send`: `E.js` → `E.ts`
- `@endo/patterns`: `patternMatchers.js` → `patternMatchers.ts` (helper types encapsulated; standalone `types.ts` removed)

### Authoring conventions (see `docs/typescript.md`, `AGENTS.md`)

- **Package entrypoints stay `.js`** — Node.js will not resolve `.ts` under `node_modules`, so a `.js` entrypoint re-exports the `.ts` implementation by relative import.
- **Runtime `.ts` imports must be relative** (same package / same-repo). Enforced by a new `@endo/no-nonrelative-ts-import` lint rule in the `internal` config.
- **`.ts` entrypoints are intentionally out of scope** for now (would need cross-package `.ts` resolution + rewrite machinery).

### Security Considerations

No new authorities or trust boundaries. Type erasure is a build/load-time transformation; runtime semantics are unchanged.

### Testing Considerations

Existing suites pass against the migrated `.ts` via native Node stripping. New unit tests cover the `@endo/no-nonrelative-ts-import` rule.

Published packages still ship `.js` with identical exports. Repo minimum Node is 22 (native stripping of `.ts` source at dev/test time). No breaking changes for downstream consumers of the published `.js`.

### Upgrade Considerations

None — published artifacts are unchanged.

### Known follow-ups

- Publishing `.ts` source + `.d.ts.map` so "Go to definition" lands on the `.ts` source (deferred; coupled to `ts-node-pack` emit).


docs/typescript.md
```diff
@@ -0,0 +1,186 @@
+# TypeScript
+
+This repo uses a "type-stripping" approach to TypeScript. Source files are
+authored in `.ts` with only erasable type syntax (no `enum`, `namespace`,
+`const enum`, or parameter properties). This lets the same source work:
+
+- **At development time** — imported directly as `.ts` (via a type-stripping
+  loader or Node's built-in `--experimental-strip-types`)
```
mhofman, line 8: Make it clear this is only for same package (or same repo) relative imports.

docs/typescript.md
```diff
@@ -0,0 +1,186 @@
...
+- `enum`, `namespace`, `const enum`, parameter properties — **not allowed**
+  (these require emit-time transformation, not just erasure)
+
+See the
+[ts-blank-space unsupported syntax list](https://github.com/bloomberg/ts-blank-space/blob/main/docs/unsupported_syntax.md)
+for the full set of restrictions.
+
+### Import specifiers
+
+In `.ts` source files, import other `.ts` files using `.ts` specifiers:
```
mhofman: Clarify that for runtime imports, `.ts` should only be used for relative (same package / same repo) imports.

docs/typescript.md
mhofman, line 1: We need to include some guidance for  package entrypoints authored in `.ts`. Since Node.js will not allow importing non-relative `.ts` specifiers (well more precisely `.ts` from under `node_modules`, a package entrypoint should be a `.js` file in source. However that `.js` file can simply contain `export * from './src/index.ts'` (or wherever the ts implementation is, as long as it's not the same file name).

Alternatively, if we want to use `.ts` entrypoints, we must have linting rules that only allow cross-package `.ts` if the target is in the same repo, and that rewrites processes these and only these (not arbitrary packages under `node_modules`)

docs/typescript.md
```diff
@@ -0,0 +1,186 @@
...
+```js
+export {};
+```
+
+**types-index.d.ts**:
+```ts
+export type * from './src/types.js';
+```
+
+This pattern is being replaced by direct `.ts` exports as packages migrate.
```
mhofman: "direct `.ts` exports" is a bit misleading, when `.ts` files cannot be entrypoints.

mhofman (changes requested):
I only reviewed the docs and `package.json`.

We're lacking instructions (and likely lint rules) enforcing that only imports from non `node_modules` can use `.ts` extensions to satisfy Node.js semantics. While cross-package `.ts` imports will work within the same repo, I am concerned about blurring the lines in authoring rules (and complexity of enforcement) if we rely on same repo cross package imports to be rewritten.

Also since we're now allowing and encouraging `.ts` source, these files should be included in the package so that "Navigate to source" continues working (the `.d.ts.map` would point at the `.ts` source file)

packages/daemon/src/daemon-node-powers.js
```diff
@@ -1,4 +1,5 @@
 // @ts-check
+/* global process */
```
boneskull, line 2: Would it be prudent to actually configure ESLint for the `node` env?

packages/daemon/test/_ava.config.mjs
```diff
@@ -0,0 +1,6 @@
+export default {
```
boneskull: Are we moving AVA configs into standalone files from `package.json`?

scripts/packing/build-ts-to-js.mjs
```diff
@@ -0,0 +1,137 @@
...
+  if (tsFiles.length === 0) {
+    console.log('No .ts files found in src/');
+    return;
+  }
+
+  console.log(`Found ${tsFiles.length} .ts file(s) to transform`);
+
+  let hasChanges = false;
+
+  for (const tsPath of tsFiles) {
```
boneskull: What prevents this from running in parallel?

scripts/packing/build-ts-to-js.mjs
```diff
@@ -0,0 +1,137 @@
...
+      await fs.writeFile(jsPath, jsContent);
+      console.log(`  ${relPath} -> ${relPath.replace(/\.ts$/, '.js')}`);
+    }
+  }
+
+  if (checkMode && hasChanges) {
+    console.error(
+      '\nRun "scripts/packing/build-ts-to-js.mjs" to update generated files',
+    );
+    process.exit(1);
```
boneskull:
```suggestion
    process.exitCode = 1;
    return;
```

scripts/packing/build-ts-to-js.mjs
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  if (!checkMode) {
+    console.log('Done!');
+  }
+}
+
+main().catch(err => {
+  console.error(err);
+  process.exit(1);
```
boneskull:
```suggestion
  process.exitCode = 1;
```

generally you only want to use `process.exit()` with sync code.

scripts/packing/postpack-package.mjs
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env node
```
boneskull: I _really_ don't like all of the complexity involved in this build process.  It's going to suck if something breaks in the middle of it.

Off the top of my head, given our constraints (grr), is that maybe we should do all of this work in a transient directory somewhere.

1. Generate `.d.ts` files; move/copy them into a transient dir structure
2. Suck in all the files we need for publish (`package.json`, `README.md`, `LICENSE`, whatever's in `package.json`'s `files`), strip types & rename extensions as necessary, and poop out the result into the dir structure alongside the declarations. A Gulp-esque pipeline
3. Pack / publish from the transient dir.

Then we shouldn't need to cleanup anything. 

> Rant:
>
> We should be using incremental TS builds. I've always been grouchy about the need to remove `.d.ts` files.  Will stop harping on this for now.

.changeset/build-ts-support.md
```diff
@@ -0,0 +1,10 @@
+---
+'@endo/eventual-send': minor
+'@endo/patterns': minor
+'@endo/exo': patch
+---
+
+Add build infrastructure for `.ts` source modules using erasable type syntax
```
boneskull: I don't understand why we need `ts-blank-space` and can't use Node.js' facilities.

boneskull (commented):
From my limited experience with type-stripping, I am loathe to accept the complexity. This complexity is further compounded by our constraints and historical choices.

There's too much "bespoke build system" here. As much as I'd _love_ to write TS instead of JS, I'm unhappy with this direction.

I'm not going to block, because I assume I'm in the minority here.   

I hope to be pleasantly surprised, but for now I will grit my teeth.

scripts/packing/postpack-package.mjs
turadg: I agree. Last week I made a ts-node-pack CLI that does that and would make this PR much simpler. My kid was sick and is on vacation this week so I had to pause it but I'll try to get that out and redo this using that.

It would be a trusted dep in the build system but a very light one easy to audit.

packages/daemon/src/daemon-node-powers.js
turadg: Eslint has deprecated their eslint-env directive. We could configure by file path overrides but we don't have a convention for what modules will run (only) in Node or not.

scripts/packing/postpack-package.mjs
mhofman: If building the output in a temp dir works I'm ok with that. I remember some issue with tsc outDir when generating d.ts, but maybe I'm confusing things.

.changeset/build-ts-support.md
naugtur: strong +1 to that. an exact copy of the functionality ships with node.js now and is available as a function (see my link in the first comment for inspiration)

mhofman: Older versions of node don't support type stripping. I believe this is there until the minimum supported node version does.

packages/daemon/test/_ava.config.mjs
turadg: Not generally. This was just so the Ava config could have a comment. Darn package.json!

scripts/packing/build-ts-to-js.mjs
turadg: Nothing. Obsolete now tho

scripts/packing/postpack-package.mjs
turadg: It took more work than I expected but it seems to work.

.changeset/build-ts-support.md
turadg: Node also only strips on loading. It doesn't have a way to strip as a transform.

packages/daemon/src/daemon-node-powers.js
boneskull: Why do we need a convention?  Either a pkg is designed to run in Node.js or not, and it can be trivially configured via `eslintConfig` in its `package.json`.  

Of course, ESLint 9+ means you pull in [globals](https://npm.im/globals), but that's another story.

.changeset/build-ts-support.md
boneskull: @turadg It does, though: 

https://nodejs.org/api/module.html#modulestriptypescripttypescode-options

(scheduled to be flagged as stable imminently, iirc)

boneskull: That being said it seems premature to me to leverage type-stripping if we cannot use native facilities.

turadg: Ok, not before v22.13.0 and Endo currently supports 18+

turadg:
Reworked since your review (@mhofman) — thanks for the feedback. The approach changed and the branch was rebased, so your inline comments may show against outdated lines. Summary of how each point is addressed:

- **`.ts` entrypoints** — punted for now. Entrypoints stay `.js` and re-export the `.ts` implementation by relative import (e.g. `eventual-send/src/no-shim.js` → `./E.ts`). Documented under "Package entrypoints stay `.js`" in `docs/typescript.md`, and in `AGENTS.md`.
- **Relative-only `.ts` imports** — now documented *and* lint-enforced: the new `@endo/no-nonrelative-ts-import` rule (enabled in the `internal` config) rejects `.ts`/`.mts`/`.cts` specifiers on non-relative paths, so a cross-package `.ts` import (which Node can't resolve under `node_modules`) can't slip in. The branch currently has zero cross-package `.ts` imports.
- **Docs clarity** (your line 8 / 27 / 128 comments) — the "Import specifiers" section now states relative-only explicitly with a cross-package warning, and the misleading "direct `.ts` exports" wording is fixed.
- **Stripping approach** — `ts-blank-space` is gone. Dev/test relies on Node ≥22 native stripping; bundling uses `amaro` strip-only (matching `bundle-source`); packing uses `ts-node-pack`. The old in-repo `scripts/packing/` prepack/postpack scripts are gone.
- **Node bump** — the `ocapn-guile-interop` job moved 20.x → 22.x (20 is EOL), so every CI lane that loads `.ts` has native stripping.

Still open, flagged as a follow-up rather than expanding this PR: publishing `.ts` source + `.d.ts.map` so "Go to definition" lands on the `.ts` source. Happy to pull it in here if you'd prefer it in-scope.

PR #3152, opened 2026-03-31
refactor(compartment-mapper): use @endo/errors

erights:
Closes: #XXXX
Refs: https://github.com/endojs/endo/commit/3fcb5b117eccb326c6c81339ae6a293a6bcaa9d4#diff-2ce5fb201c10cb24e075359132ab825315f08bdd168d4c8e9ff464b87d3e74ffR4-R7 , #3110 

## Description

@boneskull reminded me of https://github.com/endojs/endo/commit/3fcb5b117eccb326c6c81339ae6a293a6bcaa9d4#diff-2ce5fb201c10cb24e075359132ab825315f08bdd168d4c8e9ff464b87d3e74ffR4-R7

This PR moves from the `assert` style towards the 
```js
    expr || Fail`...`
```
style, which is more efficient in the typical non-erroneous case, and is more idiomatic for Endo. It does this at the price of introducing a new dependency from `@endo/compartment-mapper` to `@endo/error` . If it is important for `@endo/compartment-mapper` not to transitively depend on `ses`, then this PR should probably wait on #3110 or a successor that accomplishes the same goal.

### Security Considerations

None

### Scaling Considerations

If this PR waits for #3110 or equivalent, then none. Otherwise, the transitive dependency on `ses` may be significant. OTOH, `@endo/compartment-mapper`'s use of `ses`'s `assert` shows that it is already transitively dependent, so probably makes no difference either way.

We want to encourage the `... || Fail...` style over the `assert` style, so good to get rid of violations of our own style preferences in our own code.

### Testing Considerations

All existing tests pass without modification, which should be good enough for this refactoring PR.

Some error messages that were formed with untagged template literals are now formed with `Fail`. Untagged template literals don't redact, which seems to be intentional here. To preserve that intention, this PR annotations all `Fail` interpolations that did not already have a `q(...)` with a `b(...)`.

### Upgrade Considerations

none

### DIFF
--- a/packages/compartment-mapper/src/compartment-map.js
+++ b/packages/compartment-mapper/src/compartment-map.js
@@ -1,5 +1,6 @@
 /* Validates a compartment map against its schema. */
 
+import { Fail, q, b } from '@endo/errors';
 import {
   assertPackagePolicy,
   ATTENUATORS_COMPARTMENT,
@@ -31,17 +32,13 @@ import {
  *   DigestedCompartmentDescriptor} from './types.js'
  */
 
-// TODO convert to the new `||` assert style.
-// Deferred because this file pervasively uses simple template strings rather than
-// template strings tagged with `assert.details` (aka `X`), and uses
-// this definition of `q` rather than `assert.quote`
-const q = JSON.stringify;
 const { keys, entries } = Object;
 const { isArray } = Array;
 
-/** @type {(a: string, b: string) => number} */
-// eslint-disable-next-line no-nested-ternary
-export const stringCompare = (a, b) => (a === b ? 0 : a < b ? -1 : 1);
+/** @type {(left: string, right: string) => number} */
+export const stringCompare = (left, right) =>
+  // eslint-disable-next-line no-nested-ternary
+  left === right ? 0 : left < right ? -1 : 1;
 
 /**
  * @template T
@@ -83,12 +80,8 @@ function* enumerate(iterable) {
  * @returns {asserts value is string}
  */
 const assertString = (value, pathOrMessage, url) => {
-  const keypath = pathOrMessage;
-  assert.typeof(
-    value,
-    'string',
-    `${keypath} in ${q(url)} must be a string; got ${q(value)}`,
-  );
+  typeof value === 'string' ||
+    Fail`${b(pathOrMessage)} in ${q(url)} must be a string; got ${q(value)}`;
 };
 
 /**
@@ -97,7 +90,7 @@ const assertString = (value, pathOrMessage, url) => {
  * @param {unknown} allegedLabel
  * @param {string} keypath
  * @param {string} url
- * @returns {asserts alleged is string}
+ * @returns {asserts allegedLabel is string}
  */
 const assertLabel = (allegedLabel, keypath, url) => {
   assertString(allegedLabel, keypath, url);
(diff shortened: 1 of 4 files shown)


packages/compartment-mapper/src/compartment-map.js
```diff
@@ -1,5 +1,6 @@
 /* Validates a compartment map against its schema. */
 
+import { Fail, q, b } from '@endo/errors';
```
boneskull, line 3: I don't understand why we're pulling in `@endo/errors` if all these symbols are already available in `globalThis`. I'm missing some context.

packages/compartment-mapper/src/compartment-map.js
```diff
@@ -83,12 +79,8 @@ function* enumerate(iterable) {
...
  */
 const assertString = (value, pathOrMessage, url) => {
-  const keypath = pathOrMessage;
-  assert.typeof(
-    value,
-    'string',
-    `${keypath} in ${q(url)} must be a string; got ${q(value)}`,
-  );
+  typeof value === 'string' ||
+    Fail`${b(pathOrMessage)} in ${q(url)} must be a string; got ${q(value)}`;
```
boneskull, line 84: As I noted in Slack, I am not fond of `Fail` generally for two reasons:

1. It cannot support `Error.code`
2. When used outside of an `asserts` type guard, TS cannot infer that `Fail` terminates unless `throw` is used.  This makes correct usage of `Fail` rather tricky.

We could mitigate 1. by introducing a `Fail.code(str)`-like API, but that's very much a separate PR & discussion.

I have a branch where I've replaced all thrown errors (including usage of `Fail`, `throw new Error()`, & `throw Error()`) with a wrapper using `makeError()`, built on top of #3130. While this would be fine to merge as-is, I would want to replace the `Fail` calls (regardless of whether `Fail.code()` becomes a thing).

packages/compartment-mapper/src/compartment-map.js
```diff
@@ -97,7 +89,7 @@ const assertString = (value, pathOrMessage, url) => {
  * @param {unknown} allegedLabel
  * @param {string} keypath
  * @param {string} url
- * @returns {asserts alleged is string}
+ * @returns {asserts allegedLabel is string}
```
boneskull, line 93: 👍

packages/compartment-mapper/src/compartment-map.js
```diff
@@ -107,12 +99,10 @@ const assertLabel = (allegedLabel, keypath, url) => {
...
-  assert(
-    /^(?:@[a-z][a-z0-9-.]*\/)?[a-z][a-z0-9-.]*(?:>(?:@[a-z][a-z0-9-.]*\/)?[a-z][a-z0-9-.]*)*$/.test(
-      allegedLabel,
-    ),
-    `${keypath} must be a canonical name in ${q(url)}; got ${q(allegedLabel)}`,
-  );
+  /^(?:@[a-z][a-z0-9-.]*\/)?[a-z][a-z0-9-.]*(?:>(?:@[a-z][a-z0-9-.]*\/)?[a-z][a-z0-9-.]*)*$/.test(
+    allegedLabel,
+  ) ||
+    Fail`${b(keypath)} must be a canonical name in ${q(url)}; got ${q(allegedLabel)}`;
```
boneskull, line 106: I am a little unclear on where we want to use `b`/`bare` vs `q`/`quote`; obviously the latter is for strings, but from what I can tell, `keypath` is also a string.

boneskull (approved):
So while I don't have a major problem with landing this, I'll likely change it later once we land support for `Error.code`; I want all exceptions thrown (directly) out of `@endo/compartment-mapper` to have a proper error code.

Please let the record show I didn't ask for an implementation. 😉 Mostly I was curious about whether I needed to use `details`/`X` anywhere, and how. I should have been more specific. That said, I don't even know when to use `b`/`bare`, so I'm hoping you can enlighten me.

The reason behind the original ask was that I'm also looking into partitioning the assertions here into this module for `CompartmentMapDescriptor`-specific assertions and another for general, reusable assertions (e.g., `assertPlainObject`, `assertFileUrlString`), with an eye towards enabling `PackageCompartmentMapDescriptor` validation before returning from `mapNodeModules`. I saw the TODO and didn't want to propagate the same problem to a new module.

packages/compartment-mapper/src/compartment-map.js
kriskowal: We’re moving toward importing these global facilities from a package mostly because it’s easier to thread types, but in the long run, `@endo/errors` will become usable without first importing `ses`.

kriskowal (approved):

PR #3153, opened 2026-03-31
fix(ses): plug implicit NaN side-channel

erights:
Closes: #XXXX
Refs: #XXXX

- [x] TODO cite Nikita's NaN side-channel demonstration.

## Description

The JavaScript language can leak the bit encoding of a NaN via shared TypedArray views of an common ArrayBuffer. Although the JavaScript language has only one NaN value, the underlying IEEE 754 double-precision floating-point representation has many different bit patterns that represent NaN. This can be exploited as a side-channel to leak information. This actually happens on some platforms such as v8.

@ChALkeR explains at https://github.com/tc39/ecma262/pull/758#issuecomment-3919093669 the behavior of this side-channel on v8. At https://junk.rray.org/poc/nani.html he demonstrates it, and it indeed even worse than I expected.

To plug this side-channel, we make two coordinated changes.
  * We stop listing the `Float*Array` constructors as universal globals. This prevents them from being implicitly endowed to created compartments, because they are not harmless. However, we still keep them on the start compartment (the original global), consider them intrinsics, and still repair and harden them on `lockdown()`. Thus, they can be explicitly endowed to child compartments at the price of enabling code in that compartment to read the side-channel.
  * On `lockdown()`, we repair the `DataView.prototype.setFloat*` methods so that they only write canonical NaNs into the underlying ArrayBuffer.

  The `@endo.marshal` package's `encodePassable` encodings need to obtain the bit representation of floating point values. It had used `Float64Array` for that. However, sometimes the `@endo/marshal` package is evaluated in a created compartment that would now lack that constructor. (This reevaluation typically occurs when bundling bundles in that package.) So instead, `encodePassable` now uses the `DataView` methods which are now safe.

### Security Considerations

The point. This NaN side-channel in ses had been implicit, easily forgotten, and hard to plug in user code.

### Scaling Considerations

Not really.

Somewhere we need to document that the `Float*Array` constructors are not available in created compartments by default, and why.

### Testing Considerations

New tests included.

One test newly skipped because it always hangs for me locally, which is annoying. ***Reviewers***, feel free to ask me to unskip it.

The absence of the `Float*Array` constructors in created compartments is potentially a compat break. In fact our need to repair `encodePassable` demonstrates such a compat break.

### Upgrade Considerations

~none.~

EDIT 2026-06-16: we must publish this as a major version since it will break existing applications relying on the implicit availability of Float64Array in Compartments (copied from https://github.com/endojs/endo/pull/3153#pullrequestreview-4065021176)


packages/eslint-plugin/lib/configs/ses.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
   extends: ['plugin:@endo/internal'],
   rules: {
     'no-restricted-globals': [
-      'error',
+      'error', // TODO why is this here?
```
erights: @kriskowal , this came from https://github.com/endojs/endo/pull/2867 . But I don't understand it. AFAICT there is no such global.

packages/ses/src/commons.js
```diff
@@ -26,11 +26,16 @@ export const {
   ArrayBuffer,
   Date,
   FinalizationRegistry,
-  Float32Array,
+  // Renamed to FERAL_* because it enables the NaN side-channel
+  Float32Array: FERAL_FLOAT32_ARRAY,
+  Float64Array: FERAL_FLOAT64_ARRAY,
+  BigUint64Array,
+  DataView,
```
erights: Refactored since then. Check which are still relevant.

packages/ses/src/permits.js
```diff
@@ -144,6 +141,18 @@ export const initialGlobalPropertyNames = {
...
   Math: '%InitialMath%',
 
+  // We move these from universalPropertyNames because the NaN side channel
+  // means that they are not quite harmless.
+  // We move them to initialGlobalPropertyNames so that they'll still be
+  // included in the primordials, repaired, and hardened. Thus, they
+  // can be endowed into compartments without hazard beyond the
+  // NaN side channel.
+  //
+  // // https://github.com/tc39/proposal-float16array
```
erights, line 151:
```suggestion
  // https://github.com/tc39/proposal-float16array
```

packages/ses/src/tame-nan-sidechannel.js
```diff
@@ -0,0 +1,97 @@
...
+ * Separately, we do not include the `Float*Array` constructors on the
+ * list of universal safe globals. Thus, constructed compartments do not
+ * get these by default.
+ *
+ * These replacement `setFloat*` methods canonicalize NaN, but they
+ * do not canonicalize `-0` to `0`. If callers wish to do so, they should do
+ * it themselves before calling these
+ */
+export const tameNaNSideChannel = () => {
+  defineProperties(dataViewPrototype, getOwnPropertyDescriptors(methods));
```
erights: These should not be enumerable.

packages/eslint-plugin/lib/configs/ses.js
erights: @kriskowal , I saw no reason to keep it so I deleted it. Done. Please let me know if you object.

packages/ses/src/commons.js
erights: Deleted Float32Array and BigUint64Array. Done.

packages/ses/src/tame-nan-sidechannel.js
erights: Done.

packages/eslint-plugin/lib/configs/ses.js
erights: My mistake. This `'error'` is not the name of a global. It is a severity level. Restored.

.changeset/moody-wolves-grow.md
```diff
@@ -0,0 +1,16 @@
+---
+'ses': minor
```
kriskowal: major

kriskowal (approved):
Approved with the caveat that we must publish this as a major version since it will break existing applications relying on the implicit availability of Float64Array in Compartments.

.changeset/moody-wolves-grow.md
erights: Done

packages/ses/src/tame-nan-sidechannel.js
```diff
@@ -0,0 +1,103 @@
...
+        littleEndian,
+      ]);
+    } else {
+      return apply(FERAL_SET_FLOAT64, this, [byteOffset, value, littleEndian]);
+    }
+  },
+};
+
+/**
+ * Replaces the dangerous `setFloat*` methods on `DataView.prototype`
```
gibson042, line 83: @erights should we also update the getFloat* methods to something like `const value = apply(…); return is(value, NaN) ? NaN : value`? Not all ArrayBuffers will have been populated by SES code, and if anything it is probably more common to receive contents from the outside world.

erights: Wouldn't help. Either way, get produces a an opaque NaN. Only the setting side pierces the opacity.

PR #3154, opened 2026-04-01
support dynamic imports in CJS

boneskull:
## REVIEWERS

I acknowledge mauling the CJS lexer to transform `import()` calls on-the-fly is an egregious violation of the single-responsibility principle and is thus a _bad idea._  That said, I want to replace the lexer with an AST parser (done similarly to ESM sources), perhaps soon.  

### feat(compartment-mapper): support dynamic import() in CommonJS

This change enables use of (`await`) `import(specifier)` in CommonJS sources.

To make this work, `parse-cjs-shared-export-wrapper.js` provides an `importFn` to the functor, which invokes `compartment.import()`.  The CJS module source is transformed on-the-fly by `@endo/cjs-module-analyzer` which evades SES.

Note that dynamically-imported _non-exit_ modules are still unsupported by `loadLocation`/`importLocation` due a lack of support for pre-loaded compartments (`_preload`). Currently, only `captureForMap()` supports this option. Without it, dynamically-imported modules will be unknown to Endo unless they were statically loaded via other means.  Exit modules can be loaded dynamically _if_ the caller provides an `importHook`.

Added tests covering dynamic imports in _both_ ESM and CJS.

### feat(cjs-module-analyzer): detect dynamic imports; add on-the-fly evasion

This modifies the `analyzeCommonJS` function to return an object with two new properties:

1. `imports`: list of specifiers imported via `import()` calls
2. `source`: transformed source for SES evasion, replacing `import()` with `$h_import()`

If `$h_import` is detected in the original source, it will throw an exception. There are better ways to handle this, of course.

Updated `README.md` and added missing MIT license / attribution.

### DIFF
--- a/packages/cjs-module-analyzer/LICENSE-MIT
+++ b/packages/cjs-module-analyzer/LICENSE-MIT
@@ -1,0 +1,10 @@
+MIT License
+-----------
+
+Copyright (C) 2018-2020 Guy Bedford
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+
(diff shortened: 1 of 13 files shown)


boneskull:

- branch: `master`
  - **[support dynamic imports in CJS #3154](https://github.com/endojs/endo/pull/3154)** - branch: `boneskull/cjs-imports` *(this PR)*

---

boneskull:
Closing in lieu of #3220

PR #3158, opened 2026-04-04
feat(parser-pipeline): create @endo/parser-pipeline

boneskull:
Introduces `@endo/parser-pipeline`, a new package that eliminates redundant Babel AST parsing when multiple consumers need to analyze or transform the same JavaScript module source.

The core problem: tools built on `@endo/compartment-mapper` (such as LavaMoat) have historically parsed each module two or three times — once for import/export analysis, once for evasive transforms, and once for policy-relevant globals analysis. This package composes those passes into a single parse-traverse-generate cycle.

**`createParsers(config?)`** is the primary entry point. It accepts a single flat configuration object that combines pipeline options (`analyzerFactories`, `transformFactories`, per-language `mjs`/`cjs`/`mts` overrides, lifecycle hooks) with worker-pool options (`workerScript`, `workerData`, `maxWorkers`, `idleTimeout`). It returns `{ sync, async }` parser maps that are drop-in replacements for `parserForLanguage` in `@endo/compartment-mapper`. The module-source analysis step is handled implicitly by the pipeline; consumers only supply user-defined analyzer and transform factories.

Async-only consumers (e.g. policy generation) need only supply the worker/pool options and lifecycle hooks — they do not need to pass factory configs that only run inside the worker.

**`runPipelineInWorker(port, config)`** powers the async path. It accepts the same pre-merge `PipelineConfig` shape as `createParsers`, performing the merge internally. Consumer-provided worker scripts call this to listen for parse tasks dispatched by the worker pool, run the full pipeline in a worker thread, and post results back. The worker pool (`WorkerParserPool`) manages spawning, queuing, idle timeouts, and unref'd workers so the process can exit cleanly once all in-flight dispatches settle.


boneskull:

- branch: `master`
  - **[feat(parser-pipeline): create @endo/parser-pipeline #3158](https://github.com/endojs/endo/pull/3158)** - branch: `boneskull/ast-service` *(this PR)*

---

packages/parser-pipeline/src/worker-runner.js
```diff
@@ -0,0 +1,132 @@
...
+      }
+
+      const ast = parseBabel(source, {
+        sourceType: 'module',
+        tokens: true,
+        createParenthesizedExpressions: true,
+        allowReturnOutsideFunction: true,
+        errorRecovery: true,
+      });
+
```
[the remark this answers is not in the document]
boneskull: Fixed in a9fdc6108. Added `ast.errors` check after parsing.

packages/parser-pipeline/src/worker-runner.js
```diff
@@ -0,0 +1,132 @@
...
+      /** @type {WorkerResultMessage} */
+      const result = {
+        type: 'result',
+        id,
+        record,
+        bytes: resultBytes,
+        analyzerResults,
+      };
+
+      port.postMessage(result);
```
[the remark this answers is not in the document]
boneskull: Fixed in a9fdc6108. Now transferring `resultBytes.buffer` in the transfer list.

packages/parser-pipeline/src/worker-runner.js
```diff
@@ -0,0 +1,132 @@
...
+ * composed pipeline for each one.
+ *
+ * This is the main entry point for consumer-provided worker scripts. The
+ * worker script imports its own visitor modules, builds a
+ * {@link WorkerPipelineConfig}, and calls this function.
+ *
+ * @param {MessagePort} port The worker's `parentPort`.
+ * @param {WorkerPipelineConfig} config Pipeline configuration with visitor
+ *   factories.
+ * @returns {{removePipelineListener: () => void}} An object with a `stop` method to stop listening for messages.
```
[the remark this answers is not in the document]
boneskull: Fixed in a9fdc6108. Updated JSDoc to say `removePipelineListener`.

packages/parser-pipeline/src/worker-pool.js
```diff
@@ -0,0 +1,360 @@
...
+    return new Promise((resolve, reject) => {
+      const worker = this.#getWorker();
+      if (!worker) {
+        this.#log('Worker pool is at capacity, queuing task', message);
+        this.#queue.push({ message, resolve, reject });
+        return;
+      }
+      this.#log('Worker pool is dispatching task', message);
+      this.#pending.set(id, { worker, resolve, reject });
+      worker.postMessage(message);
```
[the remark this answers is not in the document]
boneskull: Fixed in a9fdc6108. Wrapped `worker.postMessage()` in try/catch that cleans up and rejects.

packages/compartment-mapper/src/map-parser.js
```diff
@@ -187,128 +140,146 @@ const makeExtensionParser = (
...
     );
-    if ('then' in result && typeof result.then === 'function') {
-      throw new TypeError(
-        'Sync parser cannot return a Thenable; ensure parser is actually synchronous',
-      );
-    }
-    return result;
-  };
   syncParser.isSyncParser = true;
+  return syncParser;
```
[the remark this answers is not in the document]
boneskull: Fixed in a9fdc6108. Restored the explicit thenable check in the sync parser wrapper.

packages/parser-pipeline/src/composed-parser.js
```diff
@@ -0,0 +1,230 @@
...
+ * @returns {ParserImplementation} An object with a `parse` method compatible with
+ *   compartment-mapper's `parserForLanguage` entries.
+ */
+
+/**
+ * Creates a `parserForLanguage`-compatible parser that composes multiple Babel
+ * visitor passes into a single parse-traverse-generate cycle.
+ *
+ * The returned object has a `parse` method matching compartment-mapper's
+ * `ParseFn` signature. For each module source:
```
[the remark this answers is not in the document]
boneskull: Fixed in a9fdc6108. Removed the duplicate JSDoc block, keeping only the overload + implementation docs.

packages/parser-pipeline/README.md
```diff
@@ -0,0 +1,154 @@
...
+import { isMainThread, parentPort, workerData } from 'node:worker_threads';
+
+if (!parentPort || isMainThread) {
+  throw new Error('This module must be run as a worker thread');
+}
+
+// runWorkerPipeline registers a listener for messages on parentPort;
+// removePipelineListener removes it, if needed.
+const { removePipelineListener } = runWorkerPipeline(parentPort, {
+  createAnalyzerPasses: (location, specifier) => [
```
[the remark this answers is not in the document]
boneskull: Fixed in a9fdc6108. Updated to `runPipelineInWorker`.

packages/parser-pipeline/README.md
```diff
@@ -0,0 +1,154 @@
...
+  analyzerFactories, // read-only visitor factories (run first)
+  transformFactories, // mutating visitor factories (run after analyzers)
+  sourcePreprocessor, // optional string transform before Babel parse
+  onModuleComplete, // optional callback after each module completes
+});
+
+// parser satisfies ParserImplementation with synchronous: true
+```
+
+The `recordBuilder` is a required first argument. It is a synchronous function whichreceives the generated code and module location. It must return a `FinalStaticModuleType` (e.g., a SES `ModuleSource`-compatible record).
```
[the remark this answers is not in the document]
boneskull: Fixed in a9fdc6108.

packages/parser-pipeline/package.json
```diff
@@ -0,0 +1,79 @@
...
+  "license": "Apache-2.0",
+  "main": "./index.js",
+  "name": "@endo/parser-pipeline",
+  "publishConfig": {
+    "access": "public"
+  },
+  "repository": {
+    "directory": "packages/module-analyzer",
+    "type": "git",
+    "url": "git+https://github.com/endojs/endo.git"
```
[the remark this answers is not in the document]
boneskull: Fixed in a9fdc6108. Both `homepage` and `repository.directory` now point to `packages/parser-pipeline`.

packages/parser-pipeline/src/worker-parser.js
```diff
@@ -0,0 +1,74 @@
...
+ * returns the serializable results. This allows multiple modules to be parsed
+ * in parallel across worker threads.
+ *
+ * The returned parser has `synchronous: false`, which causes
+ * compartment-mapper's `makeMapParsers` to use the async trampoline. The async
+ * trampoline implicitly awaits the Promise returned by `parse()`.
+ *
+ * @param {string | URL} workerScript Path to the consumer-provided worker
+ *   entry point. The worker script should call `runPipelineInWorker` from
+ *   `@endo/parser-pipeline/worker`.
```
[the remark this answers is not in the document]
boneskull: Fixed in a9fdc6108. Updated to `@endo/parser-pipeline/worker-runner.js`.

boneskull:
Lint job will fail until #3325 is merged, but I didn't want to restack on top of that PR.

packages/parser-pipeline/src/external.types.js
```diff
@@ -0,0 +1 @@
+export {};
```
naugtur, line 1: I'm not exactly sure why the empty types are needed here while there's a whole src/types folder with the actual stuff.

packages/parser-pipeline/test/mts-parser.test.js
```diff
@@ -0,0 +1,314 @@
...
+    {},
+  );
+
+  t.is(result.parser, 'mts');
+  t.truthy(result.record);
+  t.truthy(result.bytes);
+
+  const out = textDecoder.decode(result.bytes);
+  // Type annotations must be gone from the generated code.
+  t.false(out.includes(': number'), 'type annotation stripped');
```
naugtur: (could) this is whitespace sensitive. I'd pass it to `new Function` and see if it errors out.

packages/parser-pipeline/test/worker-pool.test.js
```diff
@@ -0,0 +1,363 @@
...
+test.serial('WorkerParserPool.terminate rejects pending tasks', async t => {
+  const pool = new WorkerParserPool(workerScript, { maxWorkers: 1 });
+  const promise = pool.dispatch(
+    encode('export const x = 1;'),
+    'x',
+    'file:///x.js',
+    'file:///',
+  );
+  await pool.terminate();
+  await t.throwsAsync(promise, { message: /terminated/i });
```
naugtur, line 134: (nit) 
```suggestion
  const testResult = t.throwsAsync(promise, { message: /terminated/i });
  await pool.terminate();
  return testResult
```
This would be more accurate, but less appealing to linters.

packages/parser-pipeline/src/parsers.js
```diff
@@ -0,0 +1,356 @@
...
+      freeze(
+        fromEntries(
+          entries(finalConfig).map(([language, langConfig]) => {
+            const lang = /** @type {PipelineLanguage} */ (language);
+            return [lang, createSyncParserFromConfig(lang, langConfig)];
+          }),
+        ),
+      )
+    );
+    return syncMap;
```
naugtur, line 311: (nit) 
```suggestion
    if (!syncMap) {
      syncMap = /** @type {Record<PipelineLanguage, ParserImplementation>} */ (
        freeze(
          fromEntries(
            entries(finalConfig).map(([language, langConfig]) => {
              const lang = /** @type {PipelineLanguage} */ (language);
              return [lang, createSyncParserFromConfig(lang, langConfig)];
            }),
          ),
        )
      );
    }
    return syncMap;
```
---
> Also, the casting from `language` to `lang` should be unnecessary if `FinalPipelineConfig`was specified to have keys that explicitly, not matter-o-factly match `PipelineLanguage`

Actually, no, typescript forces return type of Object.entries to have a `string` for key regardless. I tried it :(

packages/parser-pipeline/src/pipeline-config.js
```diff
@@ -0,0 +1,163 @@
...
+          const userExtras = perLanguageExtras[language] ?? {};
+          return [
+            language,
+            mergePerLanguageConfig(
+              sharedFields,
+              userExtras,
+              langDef.defaultFinalizeRecord,
+            ),
+          ];
+        }),
```
naugtur, line 143: It's great that we can have the shared fields propagate to every definition transparently. 
From what I see in the code you can also override the shared fields for a specific language. Is that supporting a specific usecase we already know, or is it motivated by good API design only?

packages/parser-pipeline/src/pipeline-config.js
```diff
@@ -0,0 +1,163 @@
...
+      ? {
+          babelGeneratorOptions: pick(
+            langBabelGeneratorOptions,
+            sharedBabelGeneratorOptions,
+          ),
+        }
+      : {}),
+    ...(pick(langLog, sharedLog) ? { log: pick(langLog, sharedLog) } : {}),
+    finalizeRecord: langFinalizeRecord ?? defaultFinalize,
+  };
```
naugtur, line 94: TS seems fine with this much more human readable version:

```suggestion
 /**
   * @param {Record<string, any>} obj
   * @returns {Record<string, any>}
   */
  const unlessEmpty = obj => {
    return fromEntries(entries(obj).filter(([_, v]) => v !== undefined));
  };

  return {
    visitorFactories: /** @type {VisitorPassFactory<unknown>[]} */ ([
      ...new Set([
        ...(sharedVisitorFactories ?? []),
        ...(langVisitorFactories ?? []),
      ]),
    ]),
    ...unlessEmpty({
      onModuleComplete: langOnModuleComplete ?? sharedOnModuleComplete,
      onModuleStart: langOnModuleStart ?? sharedOnModuleStart,
      onParseError: langOnParseError ?? sharedOnParseError,
      babelParserOptions: langBabelParserOptions ?? sharedBabelParserOptions,
      babelGeneratorOptions: langBabelGeneratorOptions ?? sharedBabelGeneratorOptions,
      log: langLog ?? sharedLog,
      finalizeRecord: langFinalizeRecord ?? defaultFinalize,
    }),
  };
```

packages/parser-pipeline/src/parsers.js
```diff
@@ -0,0 +1,356 @@
...
+  let asyncPool;
+
+  /** @returns {Record<PipelineLanguage, AsyncParserImplementation>} */
+  const getAsyncParsers = () => {
+    if (asyncMap) {
+      return asyncMap;
+    }
+    if (!workerScript) {
+      throw new Error(
+        'createParsers: workerScript option required to create async parser',
```
naugtur, line 327: This seems to be the only reason that parsers are lazily initialized. I don't see any performance gains worth the complexity. Is that correct or am I missing something?

packages/parser-pipeline/src/worker-pool.js
```diff
@@ -0,0 +1,487 @@
...
+ * A pool of worker threads that parse modules via the composed pipeline.
+ *
+ * Each worker runs a consumer-provided script that imports its own visitor
+ * modules. The pool dispatches parse tasks and routes responses back to the
+ * caller.
+ *
+ * @template {readonly any[]} [TVisitorResults=unknown[]] - The type of the visitor results array returned by the worker.
+ * @group Worker Parser
+ */
+export class WorkerParserPool {
```
naugtur, line 79: Are we assuming there can be multiple pools in one process?

packages/parser-pipeline/src/worker-pool.js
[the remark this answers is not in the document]
naugtur: github claims that commit doesn't belong to any branch. 

Some questions: 
1. Are new workers automatically spawned after one was terminated after an error? (seems like worker-pool.js:303 is that mechanism)
2. If a worker hangs instead of erroring (ignores the postMessage or fails to respond) do we have a way to recover or is that worker stuck?
3. It seems retrying if postMessage failed and worker was terminated is left for the user of the pipeline package. An error wrapping the actual error with unified code would help distinguish situations where it's worth retrying vs situations where it's an objective, likely reproducible, error.

packages/parser-pipeline/src/worker-pool.js
```diff
@@ -0,0 +1,487 @@
...
+      this.#pending.delete(msg.id);
+      task.resolve(msg);
+      this.#returnWorker(worker);
+    } else if (msg.type === 'error') {
+      this.#pending.delete(msg.id);
+      task.reject(msg.error);
+      this.#returnWorker(worker);
+    } else {
+      this.#log('Unexpected message:', msg);
+    }
```
naugtur: (could) I don't see value in allowing unexpected messages. They should be errors, not noise.

```suggestion
    this.#pending.delete(msg.id);
    if (msg.type === 'result')
      task.resolve(msg);
    } else if (msg.type === 'error') {
      task.reject(msg.error);
    } else {
      this.#log('Unexpected message:', msg);
      task.reject('Unexpected message ' + msg.type);
    }
     this.#returnWorker(worker);
```

packages/parser-pipeline/src/worker-pool.js
```diff
@@ -0,0 +1,487 @@
...
+
+    const timeout = this.#timeouts.get(worker);
+    if (timeout) {
+      clearTimeout(timeout);
+      this.#timeouts.delete(worker);
+    }
+
+    worker.removeAllListeners();
+    this.#all.delete(worker);
+    worker.removeAllListeners();
```
naugtur: Is this repetition deliberate? Are we expecting listeners added as a result of synchronously calling `this.#all.delete`?

packages/parser-pipeline/src/worker-pool.js
```diff
@@ -0,0 +1,487 @@
...
+    this.#pendingTerminations.push(
+      new Promise((resolve, reject) => {
+        worker.terminate().then(resolve, reject);
+      }).then(
```
naugtur: What happened here? Is this because worker.terminate() could error synchronously? 

```suggestion
      Promise.try(worker.terminate).then(
```

naugtur (commented):
I've read most of the implementation in src now. Sending a new batch of comments. Nothing fundamental I think.

packages/parser-pipeline/test/sync-parser.test.js
```diff
@@ -0,0 +1,274 @@
...
+  const fixturePath = fileURLToPath(fixtureUrl);
+  const fixtureBytes = readFileSync(fixturePath);
+  const fixtureFileUrl = fixtureUrl.href;
+
+  /** @type {ParseError[]} */
+  let capturedErrors = [];
+  let parseErrorCallCount = 0;
+
+  const { sync } = createParsers({
+    onParseError: errors => {
```
naugtur, line 142: When I create parsers, can I decide to rethrow the error here to make `sync.mjs.parse` explicitly throw and not accept the weirdness?

packages/parser-pipeline/src/external.types.js
boneskull: IIRC we do the same thing in `compartment-mapper`.  We are exporting this file from the `index.js` for the types only via `export * from 'src/external.types.js'`, which only practically affects the generated declarations.  If the file didn't exist or didn't have an `export` statement, this would fail at runtime.

packages/parser-pipeline/test/worker-pool.test.js
boneskull: Don't understand how it'd be more accurate.

packages/parser-pipeline/src/worker-pool.js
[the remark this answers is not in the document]
boneskull: That commit SHA got rebased away.

1. A new worker would be spawned if there's something in the queue or a new task is ready and there's no worker available in the pool.
2. If a worker hangs, it will timeout, cause an error, get terminated and removed from the pool.
3. Right, currently the error is punted to the consumer. There are at least two error modes: 
    1. Parsing fails (probably an unrecoverable syntax error). Should cause an "error" message to be posted to the main thread, which will cause the current task to be rejected. These are unlikely to be transient, so automatically retrying is probably futile.
    2. Parsing fails with errors ("Babel-recoverable"). `onParseError` is called with these, but otherwise parsing continues.
    3. The worker emits an `error`. I think typically this is a "failed to spawn worker" type of thing, which could represent a variety of root causes. Will also cause the current task to be rejected. Could be retried.

What do you think of handling these modes?  Some _n_ retries for type 3? What of syntax errors? Currently we're calling `onParseError` for "Babel-recoverable" syntax errors and throwing if the error is not "Babel-recoverable".

@naugtur

packages/parser-pipeline/test/mts-parser.test.js
boneskull: I'll think on this one.

packages/parser-pipeline/src/parsers.js
boneskull: Yeah, this is because `Object.entries` returns all enumerable entries with `string` keys, regardless of what the type of that object is in TS' cloud cuckoo land.

packages/parser-pipeline/src/pipeline-config.js
boneskull: Just my brilliant API design.

packages/parser-pipeline/src/parsers.js
boneskull: Well, a bunch of objects and functions would get created and then GC'd.  I guess I don't feel the additional complexity is so complex that we shouldn't.

packages/parser-pipeline/src/worker-pool.js
boneskull: There's nothing stopping a consumer from doing this, at any rate.  I felt a class would be less finicky than a singleton.

packages/parser-pipeline/src/worker-pool.js
boneskull: Yeah, I agree.

- [x] Reject on unexpected messages

packages/parser-pipeline/src/worker-pool.js
boneskull: That looks like a bad merge, good catch

packages/parser-pipeline/src/worker-pool.js
boneskull: That's a good question and I'd have to look at the `worker_threads` source to answer you.  I'm guessing that it probably cannot synchronously throw, and that if it could we'd have a bug in Node.js.

Looks like it can be simplified.

Regardless, no `Promise.try` in Node.js v22.23.2.

packages/parser-pipeline/test/sync-parser.test.js
boneskull: Yes.

packages/parser-pipeline/src/worker-pool.js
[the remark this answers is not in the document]
naugtur: I generally don't like the babel recoverable case. But it's fine to let it fail at runtime after.

The failure being recoverable is highly unlikely IMHO, so a single retry when an error indicates something died for no reason, would make sense to me. If implemented in this package, I'd make it a recommended opt-in

packages/parser-pipeline/test/worker-pool.test.js
naugtur: ah, nevermind

packages/parser-pipeline/src/parsers.js
naugtur: just checking my understanding

packages/parser-pipeline/test/sync-parser.test.js
naugtur: neat. worth documenting explicitly.

packages/parser-pipeline/src/pipeline-config.js
boneskull: - [ ] do this

naugtur (approved):

packages/parser-pipeline/src/worker-pool.js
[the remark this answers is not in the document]
boneskull: @naugtur I'm not going to merge this until we've settled on what we're going to do here.

After fiddling with it some more, I'm not seeing an obvious situation in which a module containing a recoverable parse error _cannot_ fail at time of execution.

The only situations (I can think of) in which you'd want to let this slide:

1. When the module is not necessarily executed at runtime. Say a conditional dynamic require.  The only way for such a module to be parsed _in the first place_ would be if the user explicitly told us to do so (e.g., via `additionalLocations` and/or `_preload`).
2. If runtime invokes some sort of custom loader that mitigates the problem.

To that end, I think we should probably do this:

1. By default, disable error recovery and throw on all parsing errors.
2. _If and only if_ `onParseError` is supplied, we would enable error recovery. If we have recoverable errors, call `onParseError` with them.

> This makes "unconditionally throwing from `onParseError`" pointless.

The "fail fast" strategy has always served me pretty well; if we know something is going to fail later, we might as well fail now and save the user some time. `onParseError` just supplies an escape hatch.

packages/parser-pipeline/test/mts-parser.test.js
boneskull: implemented

packages/parser-pipeline/src/worker-pool.js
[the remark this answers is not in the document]
naugtur: The consequences of a parse error (nom atter if recoverable) are contained anyway - `if (false) { require('broken-module') }` will not take down an app nor its policy generation. 
As long as that is true, fail fast is the right option.

PR #3171, opened 2026-04-08
adopt TS "strict"

turadg:
_incidental_

## Description

Adopts the `strict` option, default in TS 6.

Solve fallout. 

### Security Considerations

none

### Scaling Considerations

none

none

### Testing Considerations

CI

no runtime changes

### Upgrade Considerations

release notes included


packages/daemon/src/remote-control.js
```diff
@@ -19,9 +19,9 @@ export const makeRemoteControlProvider = localNodeId => {
     // We do not have a pending outbound connection attempt.
     /**
      * @type {(
-     *   remoteGateway: Promise<EndoGateway>,
+     *   remoteGateway: import('@endo/eventual-send').ERef<EndoGateway>,
      *   cancel: (error: Error) => void | Promise<void>,
-     *   cancelled: Promise<never>,
```
kriskowal, line 24: Cancelled promises should be `Promise<never>`. The intention is to express that the promise will never be fulfilled and that the consumer can expect awaiting it to throw. Not sure why this needs a change.

packages/ses/test/_package/test.js
```diff
@@ -1,6 +1,6 @@
 // node -r esm test.js
 
-import 'ses';
+import '../../index.js';
```
kriskowal, line 3: What tool doesn’t recognize reflexive imports?

kriskowal (dismissed):
This is large, and largely fine, but I do not want to degrade `Promise<never>` to `Promise<unknown>` because the former states correctly that the promise will never be fulfilled and must reject, therefore awaiting it necessarily throws.

I would like to know whether that’s a blocker or incoherent with the new TypeScript or difficult to maintain.

boneskull:
I would _prefer_ if we actually set `strict: false` here globally, which would make this PR much easier to review and merge. Individual packages can override for now, which I think we're doing anyway?

packages/exo/test/types-typed-guarded.test-d.ts
```diff
@@ -100,7 +100,10 @@ const CounterI = M.interface('Counter', {
     },
   );
   const counter = makeCounter(0n);
-  counter.incr(1); // Runtime guard will reject this, but TypeScript does not error on the impl.
+  counter.incr(1n);
+  // @ts-expect-error 'number' not compatible with the guard's type 'bigint'
+  counter.incr(1);
+  // The above only errors with strictFunctionTypes, so before we had that enabled we also did this check, which is still valid:
```
turadg: I'm not certain which strictness flag, actually.

packages/patterns/src/keys/checkKey.js
```diff
@@ -67,17 +68,17 @@ export const confirmKey = (val, reject) => {
   if (isAtom(val)) {
     return true;
   }
-  // @ts-expect-error narrowed
-  if (keyMemo.has(val)) {
+  if (typeof val === 'object' && val !== null && keyMemo.has(val)) {
```
turadg: this is a runtime code change that shouldn't affect behavior

packages/patterns/src/keys/checkKey.js
```diff
@@ -67,17 +68,17 @@ export const confirmKey = (val, reject) => {
...
     return true;
   }
   // eslint-disable-next-line no-use-before-define
   const result = confirmKeyInternal(val, reject);
   if (result) {
     // Don't cache the undefined cases, so that if it is tried again
     // with `Fail` it'll throw a diagnostic again
-    // @ts-expect-error narrowed
-    keyMemo.add(val);
+    if (typeof val === 'object' && val !== null) {
```
turadg: this is a runtime code change that shouldn't affect behavior

packages/ses/src/error/tame-error-constructor.js
```diff
@@ -43,10 +43,12 @@ export default function tameErrorConstructor(
 
   const makeErrorConstructor = (_ = {}) => {
     // eslint-disable-next-line no-shadow
+    /** @this {ErrorConstructor} */
+    /** @param {...any} rest */
     const ResultError = function Error(...rest) {
       let error;
       if (new.target === undefined) {
-        error = apply(FERAL_ERROR, this, rest);
+        error = apply(FERAL_ERROR, undefined, rest);
```
mhofman: This is a semantic difference

turadg: I missed that, thanks. Claude's take:

Correct — in any spec-compliant engine there's no observable difference. Per ES2015+ §20.5.1.1, `Error()` called as a function ignores its `this` binding and creates a fresh object, so `apply(FERAL_ERROR, this, rest)` and `apply(FERAL_ERROR, undefined, rest)` produce identical Error objects. I reverted it not because it changes runtime behavior but to keep the shim a faithful forwarder (symmetric with `construct(FERAL_ERROR, rest, new.target)` in the `new`-branch) and avoid  narrowing semantics in case of future spec/host nuance. Happy to leave it as `undefined` if you'd rather not carry the dead  forwarding.

kriskowal (approved):
The copilot  reviews have some good points. I’m leaving you with a stamp so you can merge when ready.

packages/patterns/src/keys/checkKey.js
mhofman: I'm not sure we don't or would never allow functions as passable value. Actually pretty sure some of our presences are proxies with function behavior.

packages/patterns/src/keys/checkKey.js
mhofman: Same, functions would now be filtered. I don't believe that's correct.

mhofman (changes requested):
Thanks for pointing out some of the runtime changes. I actually believe those are incorrect.

Would like eyes from @gibson042

packages/patterns/src/keys/checkKey.js
gibson042: Indeed, there are valid Keys with typeof "function" and pass-style "remotable":
```js
import '@endo/init';
import { Far } from '@endo/far';
import { passStyleOf } from '@endo/pass-style';
import { M, matches } from '@endo/patterns';

const farFunction = Far('fn', () => {});
typeof farFunction; // => "function"
passStyleOf(farFunction); // => "remotable"
matches(farFunction, M.key()); // => true
```

packages/captp/src/captp.js
```diff
@@ -224,6 +224,7 @@ export const makeCapTP = (
 
   /** @type {any} */
   let unplug = false;
+  /** @type {(reason?: any, returnIt?: boolean) => Promise<any>} */
```
gibson042: Can we avoid this `any`?
```suggestion
  /** @type {(reason?: any, returnIt?: boolean) => Promise<undefined> | Promise<never>} */
```

packages/compartment-mapper/src/map-parser.js
```diff
@@ -181,9 +181,10 @@ const makeExtensionParser = (
...
         ));
       } catch (err) {
+        const error = /** @type {Error} */ (err);
         throw Error(
-          `Error transforming ${q(language)} source in ${q(location)}: ${err.message}`,
-          { cause: err },
+          `Error transforming ${q(language)} source in ${q(location)}: ${error.message}`,
+          { cause: error },
         );
       }
```
gibson042, line 189: Suggested idiom:
```suggestion
      } catch (err) {
        const cause = /** @type {Error} */ (err);
        throw Error(
          `Error transforming ${q(language)} source in ${q(location)}: ${cause.message}`,
          { cause },
        );
      }
```

packages/compartment-mapper/test/bundle.test.js
```diff
@@ -109,11 +109,14 @@ test('makeScript with useEvaluate preserves error stack trace line numbers', asy
...
   }
-  t.assert(error.stack.includes(':4:'));
-  t.assert(error.stack.includes('bundle/main.js'));
-  t.false(error.stack.includes('file:/.*main.js'));
+  t.assert(error);
+  const thrown1 = /** @type {Error} */ (error);
+  const errorStack1 = `${thrown1.stack}`;
+  t.assert(errorStack1.includes(':4:'));
+  t.assert(errorStack1.includes('bundle/main.js'));
+  t.false(errorStack1.includes('file:/.*main.js'));
```
gibson042: Suggested simplication (to be used rather than introducing errorStack1, errorStack2, …, errorStack6):
```suggestion
  /** @type {null | Error} */
  let error = null;
  try {
    raise();
  } catch (_error) {
    error = /** @type {Error} */ (_error);
  }
  t.assert(error);
  const stack = /** @type {string} */ (error?.stack);
  t.true(stack.includes(':4:'));
  t.true(stack.includes('bundle/main.js'));
  t.false(stack.includes('file:/.*main.js'));
```

packages/compartment-mapper/test/bundle.test.js
```diff
@@ -138,12 +141,15 @@ test('makeScript with useEvaluate and sourceUrlPrefix preserves source URLs in s
...
-  t.assert(error.stack.includes(':4:'));
-  t.assert(error.stack.includes('bundled-sources/.../bundle/main.js'));
-  t.assert(!error.stack.includes('file:/.*bundle/main.js'));
+  t.assert(error);
+  const thrown2 = /** @type {Error} */ (error);
+  const errorStack2 = `${thrown2.stack}`;
+  t.log(errorStack2);
+  t.assert(errorStack2.includes(':4:'));
+  t.assert(errorStack2.includes('bundled-sources/.../bundle/main.js'));
+  t.assert(!errorStack2.includes('file:/.*bundle/main.js'));
```
gibson042: Suggested simplification:
```suggestion
  /** @type {null | Error} */
  let error = null;
  try {
    raise();
  } catch (_error) {
    error = /** @type {Error} */ (_error);
  }
  t.assert(error);
  const stack = /** @type {string} */ (error?.stack);
  t.log(stack);
  t.true(stack.includes(':4:'));
  t.true(stack.includes('bundled-sources/.../bundle/main.js'));
  t.false(stack.includes('file:/.*bundle/main.js'));
```
etc.

packages/compartment-mapper/test/dynamic-require.test.js
gibson042: Defining `cause` with a fallback is a bit misleading.
```suggestion
    const error = /** @type {Error & { cause?: Error }} */ (err);
    t.regex(error.message, /Could not require pantsFolder "jorts-folder"/);
    t.assert(error.cause);
    t.regex(
      error.cause.message,
      new RegExp(
        `Importing "jorts-folder" in resource "jorts-folder" in "pantspack>pantspack-folder-runner" was not allowed by "packages" policy: ${JSON.stringify(policy.resources['pantspack>pantspack-folder-runner'].packages)}`,
      ),
    );
```

packages/compartment-mapper/test/scaffold.js
```diff
@@ -173,7 +173,8 @@ export function scaffold(
         );
       } catch (error) {
         if (onError) {
-          return onError(t, { error, title });
+          const typedError = /** @type {Error} */ (error);
+          return onError(t, { error: typedError, title });
```
gibson042:
```suggestion
          return onError(t, { error: /** @type {Error} */ (error), title });
```

packages/compartment-mapper/test/source-url-rewrite.test.js
```diff
@@ -50,12 +50,14 @@ test('rewrite source url', async t => {
...
 
   t.assert(error);
-  t.log(error.stack);
+  const thrown = /** @type {Error} */ (error);
+  const errorStack = `${thrown.stack}`;
+  t.log(errorStack);
   t.assert(
-    error.stack.includes('/compartment-mapper/test/fixtures-stack/index.js:3:'),
+    errorStack.includes('/compartment-mapper/test/fixtures-stack/index.js:3:'),
   );
```
gibson042:
```suggestion
  /** @type {Error | undefined} */
  let error;
  try {
    await app.import();
  } catch (_error) {
    error = /** @type {Error} */ (_error);
  }

  t.assert(error);
  const stack = /** @type {string} */ (error?.stack);
  t.log(stack);
  t.true(stack.includes('/compartment-mapper/test/fixtures-stack/index.js:3:'));
```

packages/compartment-mapper/test/stack.test.js
```diff
@@ -26,13 +26,15 @@ test('archive stack trace source', async t => {
...
-  t.log(error.stack);
+  const thrownArchive = /** @type {Error} */ (error);
+  const archiveStack = `${thrownArchive.stack}`;
+  t.log(archiveStack);
   t.assert(
-    error.stack.includes(
+    archiveStack.includes(
       '.../compartment-mapper/test/fixtures-stack/index.js:3:',
     ),
   );
```
gibson042, line 38:
```suggestion
  /** @type {Error | undefined} */
  let error;
  try {
    await app.import();
  } catch (_error) {
    error = /** @type {Error} */ (_error);
  }

  t.assert(error);
  const stack = /** @type {string} */ (error?.stack);
  t.log(stack);
  t.true(
    stack.includes('.../compartment-mapper/test/fixtures-stack/index.js:3:'),
  );
```

packages/compartment-mapper/test/stack.test.js
```diff
@@ -46,13 +48,15 @@ test('disk stack trace source', async t => {
...
-  t.log(error.stack);
+  const thrownDisk = /** @type {Error} */ (error);
+  const diskStack = `${thrownDisk.stack}`;
+  t.log(diskStack);
   t.assert(
-    error.stack.includes(
+    diskStack.includes(
       '/packages/compartment-mapper/test/fixtures-stack/index.js:3:',
     ),
   );
```
gibson042, line 60:
```suggestion
  /** @type {Error | undefined} */
  let error;
  try {
    await importLocation(readPowers, fixtureLocation);
  } catch (_error) {
    error = /** @type {Error} */ (_error);
  }

  t.assert(error);
  const stack = /** @type {string} */ (error?.stack);
  t.log(stack);
  t.true(
    stack.includes(
      '/packages/compartment-mapper/test/fixtures-stack/index.js:3:',
    ),
  );
```

packages/ses-ava/test/panic.test.js
```diff
@@ -70,7 +70,8 @@ test('panic last resort', t => {
     panic(Error('testing panic using last resort'));
   } catch (err) {
     catchHappened = true;
-    t.true(Object.hasOwn(err, PanicEndowmentSymbol));
+    const thrown = /** @type {object} */ (err);
+    t.true(Object.hasOwn(thrown, PanicEndowmentSymbol));
     t.is(err, lastResortError);
```
gibson042:
```suggestion
    t.true(Object.hasOwn(thrown, PanicEndowmentSymbol));
    t.is(thrown, lastResortError);
```

packages/zip/src/buffer-reader.js
```diff
@@ -15,8 +15,12 @@ const q = JSON.stringify;
...
 const privateFields = new WeakMap();
 
-/** @type {(bufferReader: BufferReader) => BufferReaderState} */
-const privateFieldsGet = privateFields.get.bind(privateFields);
+/**
+ * @param {BufferReader} bufferReader
+ * @returns {BufferReaderState}
+ */
+const privateFieldsGet = bufferReader =>
+  /** @type {BufferReaderState} */ (privateFields.get(bufferReader));
```
gibson042: This is technically a behavioral difference that might actually matter.
```suggestion
const privateFieldsGet =
  /** @type {(bufferReader: BufferReader) => BufferReaderState} */ (
    privateFields.get.bind(privateFields)
  );
```

packages/daemon/src/daemon.js
```diff
@@ -310,8 +310,11 @@ const makeDaemonCore = async (
   /** @type {Map<FormulaIdentifier, object>} */
   const refForId = new Map();
 
-  /** @type {WeakMultimap<Record<string | symbol, unknown>, FormulaIdentifier>['get']} */
-  const getIdForRef = ref => idForRef.get(ref);
+  /** @type {DaemonCore['getIdForRef']} */
+  const getIdForRef = ref =>
+    typeof ref === 'object' && ref !== null
+      ? idForRef.get(/** @type {Record<string | symbol, unknown>} */ (ref))
+      : undefined;
```
gibson042: We never directly use `typeof ref === 'object' && ref !== null`, `isPrimitive` exists to more clearly communicate intent (and returns `false` for functions). But in this case, I prefer the version with no direct filtering.
```suggestion
  /** @type {DaemonCore['getIdForRef']} */
  const getIdForRef = ref => idForRef.get(/** @type {any} */ (ref));
```

gibson042 (dismissed):

packages/ses/test/_package/test.js
turadg: the reflexive import trips TS6's strict mode because ses's exports map routes ESM to `types.d.ts` and CJS to `dist/types.d.cts`.  Both define `Compartment` as a global, and the duplicate-identifier error surfaces even under `@ts-nocheck` because TS still loads the imported module's declarations. Relative  '../../index.js' sidesteps the package-exports path entirely.

packages/captp/src/captp.js
turadg: Good. Though It had to be `Promise<void>` to match the `Promise.resolve()`

turadg:
@gibson042 this has fixups now with 12 of your suggestions. Nine landed verbatim:

  - `map-parser.js:189` — `const cause` + `{ cause }` idiom
  - `bundle.test.js:119` — collapse `errorStack1` to `stack`
  - `bundle.test.js:152` — collapse `errorStack2` to `stack` (and extended to `errorStack3..6` per your "etc.")
  - `source-url-rewrite.test.js:62`
  - `stack.test.js:40`
  - `stack.test.js:62`
  - `scaffold.js:177` — inline `Error` cast in `onError` call
  - `buffer-reader.js:23` — `.bind(privateFields)` + outer cast
  - `daemon.js:317` — `idForRef.get(/** @type {any} */ (ref))`

  Three needed small adjustments to typecheck under strict — happy to switch to your literal form if you'd rather:

  | Thread | Your suggestion | What landed | Why |
  |---|---|---|---|
  | `captp.js:227` | `Promise<undefined> \| Promise<never>` | `Promise<void> \| Promise<never>` | Your type forced changing `Promise.resolve()` (typed `Promise<void>`) at the call site. `Promise<void>` matches exactly with no runtime change while still dropping `any`. |
  | `dynamic-require.test.js:802` | `Error & { cause?: Error }` then `t.assert(error.cause); error.cause.message` | `Error & { cause: Error }` then `t.assert(error.cause); error.cause.message` | TS doesn't propagate type-predicate narrowing through property access chains, so `error.cause.message` stayed flagged as possibly-undefined under `cause?:`.  The non-optional cast pairs with `t.assert` for runtime safety. |
  | `panic.test.js:75` | `Object.hasOwn(thrown, PanicEndowmentSymbol)` | `Object.hasOwn(/** @type {object} */ (thrown), PanicEndowmentSymbol)` | TS's `Object.hasOwn(o: object, …)`  rejects `unknown`, which is what catch variables are under strict. The cast keeps your catch-variable rename without resurrecting a `const thrown = …` local. |

packages/bundle-source/cache.js
```diff
@@ -105,17 +105,18 @@ export const makeBundleCache = (wr, cwd, readPowers, opts) => {
...
         conditions: sortedConditions,
-      },
-      {
+      }),
+      /** @type {{ read?: ReadFn, canonical?: CanonicalFn, externals?: string[] }} */ ({
         ...readPowers,
         read: loggedRead,
-      },
+        canonical: /** @type {CanonicalFn} */ (readPowers.canonical),
+      }),
```
mhofman: Unfortunate to have to change runtime code to make the type system happy here, but in practice this won't have an actual impact so I'm letting it go.

I will also note that the claimed type `{ read?: ReadFn, canonical?: CanonicalFn, externals?: string[] }` is not the actual type this would have, which is `{ ...typeof readPowers; read: typeof loggedRead; canonical: CanonicalFn; }` (note the non optional `read` and `canonical` properties).

`loggerRead` is defined, so claiming it may be missing is technically compatible, but I'm not sure why do that.

I haven't traced it but if `readPowers.canonical` can actually be missing, then the type is wrong. If it's guaranteed to be there, then I'm also confused by the type is claiming it may be missing.

kriskowal (approved):

packages/daemon/test/deferred-tasks.test.js
```diff
@@ -6,12 +6,15 @@ test('execute', async t => {
   const results = [];
   tasks.push(async () => {
     results.push(1);
+    return undefined;
```
mhofman, line 9: I dislike this kind of hack to make `Promise<undefined>` happy, but this being a test, I'm at peace with it.

packages/init/src/node-async-local-storage-patch.js
```diff
@@ -8,8 +8,11 @@ const { apply: ReflectApply } = Reflect;
 /** @type {WeakMap<AsyncLocalStorageInternal, WeakMap>} */
 const resourceStoreMaps = new WeakMap();
 
-/** @type {(key: AsyncLocalStorageInternal) => WeakMap} */
-const getStoreMap = resourceStoreMaps.get.bind(resourceStoreMaps);
+/**
+ * @param {AsyncLocalStorageInternal} key
+ * @returns {WeakMap}
+ */
+const getStoreMap = key => /** @type {WeakMap} */ (resourceStoreMaps.get(key));
```
mhofman, line 15: There are subtle debugging differences here, but given this patch is now inactive in most circumstances, I'm also ok with this change.

packages/patterns/src/keys/checkKey.js
```diff
@@ -54,8 +54,10 @@ hideAndHardenFunction(assertScalarKey);
 
 // ////////////////////////////// Keys /////////////////////////////////////////
 
-// @ts-expect-error Key does not satisfy WeakKey
-/** @type {WeakSet<Key>} */
+// Non-atom Keys (objects and callable remotables) are memoized. Atom keys
+// are primitives handled by the early return above and cannot inhabit a
+// WeakSet.
+/** @type {WeakSet<object>} */
```
mhofman: I'm curious if `WeakSet<Exclude<Key, Atom>>` would have worked.

mhofman (approved):
No runtime changes to be concerned about, and all the type changes seem sane.

packages/patterns/src/keys/checkKey.js
turadg: `WeakSet<Exclude<Key, Atom | void>>` does. I'll use that.

It's necessary because `Passable<PC, E> = void | Atom | Container<PC, E> | PC | E`. TS treats `void` as "ignore what was returned" whereas `undefined` is literally that that's the return value.

packages/bundle-source/cache.js
turadg: Good catches. I pushed a follow-up. This cast was doing two unrelated jobs (bypass excess-property checks for the spread, and assert  the parameter shape) and ended up describing neither accurately. Tightened it to:  

```js
  /** @type {{ read: ReadFn, canonical: CanonicalFn }} */ ({
    ...readPowers,
    read: loggedRead,
  }),
```

  - Dropped the `read?` / `canonical?` markers — both are always defined in this expression.
  - Dropped the phantom `externals?: string[]` — never set here.
  - Removed the redundant inner `/** @type {CanonicalFn} */ (readPowers.canonical)` and replaced it with a comment explaining why a
  cast is needed at all: `readPowers.canonical` is `CanonicalFn<FileUrlString>` (compartment-mapper's branded URL string) and
  `bundleSource`'s declared `CanonicalFn` is `(string) => Promise<string>`. Same runtime, contravariant input mismatch — the cast
  bridges the two packages' URL typings. Probably worth a follow-up to align them, but out of scope here.

packages/patterns/src/keys/checkKey.js
mhofman: 👍

Unfortunate that `isAtom` doesn't actually help narrowing the type

PR #3178, opened 2026-04-13
AI Safety: Rewrite daemon message streaming design around send/act/edit

kriskowal:
Replace the stream-formula/promise-chain approach with a simpler three-verb model: send creates messages, act broadcasts ephemeral status indicators, and edit revises existing messages via an append-only update ledger. Streaming becomes emergent from sequential edits rather than a first-class protocol concept.

### DIFF
--- a/designs/README.md
+++ b/designs/README.md
@@ -30,7 +30,7 @@
 | [daemon-checkin-checkout](daemon-checkin-checkout.md) | 2026-03-17 | 2026-03-17 | Not Started |
 | [daemon-capability-filesystem](daemon-capability-filesystem.md) | 2026-02-15 | 2026-02-24 | Not Started |
 | [daemon-content-store-gc](daemon-content-store-gc.md) | 2026-03-20 | 2026-03-20 | Not Started |
-| [daemon-message-streaming](daemon-message-streaming.md) | 2026-03-26 | 2026-03-26 | Draft |
+| [daemon-message-streaming](daemon-message-streaming.md) | 2026-03-26 | 2026-04-13 | Not Started |
 | [daemon-mount](daemon-mount.md) | 2026-03-20 | 2026-03-20 | In Progress |
 | [platform-fs](platform-fs.md) | 2026-03-18 | 2026-03-18 | In Progress |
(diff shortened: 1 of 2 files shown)


jcorbin:
So the `MessagedEdit` looks good, no notes ( other than my own iceberg of confusion around "why do messages have a strings list? what even are edge names?" &C )

But the `Action` is triggering various shades of "aint-it-chief":
1. actions need be full blown free form messages, like an ephemeral sub-reply
2. a pending message needs to be able to have multiple statuses, and each status needs to be able to be dismissed, and then maybe later re-added
3. from a presentation stand point "badges" aint it, they are ephemeral small messages like:
  - "thinking..."
  - maybe "thinking: okay, so here, is my actual train of thought, when you opt in to seeing it, should appear to 'stream', but then collapse or dismiss once the next round kicks, or at least dismiss once final response settles, or collapse, or ..."
  - "Calling file read <path> tool..."
  - "Calling file read <path> tool... done." ( maybe with result dump opt-in for debugging )
  - "Calling sub-agent <name>..."

So on balance, how about we nix actions as a separate thing, and just use message edits instead:
- what if `Package` and `MessageUpdate` both had an additional `status: Array<string>` field
- to motivate perhaps, here is a decently representative "Must Be This Tall To Provide Insight" scenario:
  1. `status: ["thinking..."]`
  2. `status: ["thinking: okay so maybe I need to use Tool A? or Tool Z? Let's see..."]`
  3. `status: ["Calling tool A...", "Calling tool B...", "Calling tool C..."]`
  4. `status: ["Calling tool A...", "Called tool B: result note", "Calling tool B..."]`
  5. `status: ["thinking: darn those tool calls did not work, and I keep making this error, let's Try Harder :tm:"]`
  6. `status: ["Calling tool python with shell script omg so much code"]`
  7. `status: ["thinking: wow that worked, I'm so good at writing code, now what should I tell the user? Let me see..."]`
  8. `strings: ["Okay, so I tried so hard, and got so far, and in the end I wrote a 30 line python script, and won this prize: 42"], status: ["typing..."]`

kriskowal:
Nixing actions and keeping edits sgtm.

PR #3179, opened 2026-04-14
disable exactOptionalPropertyTypes

turadg:
_incidental_

## Description

Disables `exactOptionalPropertyTypes` introduced in,
- https://github.com/endojs/endo/pull/3165

It leads into other parts like typedoc build and agoric-sdk. Not worth the additional integration cost.

The leak wasn't caught because of TS config. So this also tightens that up a bit and fixes errors.

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations

### DIFF
--- a/packages/daemon/test/multimap.test.js
+++ b/packages/daemon/test/multimap.test.js
@@ -6,10 +6,12 @@ import {
   makeWeakMultimap,
 } from '../src/multimap.js';
 
-[
+/** @type {Array<[() => import('../src/types.js').Multimap<any, any>, string]>} */
+const multimapCases = [
   [makeMultimap, 'multimap'],
   [makeWeakMultimap, 'weak multimap'],
-].forEach(([multimapConstructor, mapName]) => {
+];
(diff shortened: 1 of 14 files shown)


boneskull:
There seems to be a lot here other than disabling it in a `tsconfig.json`...

packages/daemon/test/multimap.test.js
```diff
@@ -6,10 +6,12 @@ import {
   makeWeakMultimap,
 } from '../src/multimap.js';
 
-[
+/** @type {Array<[() => import('../src/types.js').Multimap<any, any>, string]>} */
```
boneskull, line 9:
```suggestion
/** @import {Multimap} from '../src/types.js' */
/** @type {Array<[() => Multimap<any, any>, string]>} */
```

boneskull (approved):
ship it

packages/daemon/test/multimap.test.js
turadg: I'll leave that for a lint autofix once we port the lint rule from agoric-sdk

PR #3180, opened 2026-04-14
bump TypeScript to 6

turadg:
_evergreen_

## Description

Upgrade to TypeScript 6. Disable its new "strict" default, then fix new errors. Also: 
- gitignores Claude files
- fixes bundleSource's return type to be conditional on the format argument.
- Speeds up repo-wide TS checks by enabling TS incremental compilation

### Security Considerations

none

### Scaling Considerations

none

none

### Testing Considerations

CI

Possibly some type changes for the consumer, not part of semver

### Upgrade Considerations

changeset for the BundleSource improvement


kriskowal (approved):

package.json
```diff
@@ -20,7 +20,7 @@
     "@lavamoat/allow-scripts": "^3.3.5",
     "@lavamoat/preinstall-always-fail": "^2.1.1",
     "@octokit/core": "^3.4.0",
-    "@types/node": "^20.17.24",
+    "@types/node": "^20.19.39",
```
boneskull, line 23: (could) 

We might want to normalize the version of this pkg across the workspaces where it's used.  Is that what `catalog:dev` does?

packages/bundle-source/cache.js
```diff
@@ -362,17 +362,24 @@ export const makeBundleCache = (wr, cwd, readPowers, opts) => {
...
-   * @param {BundleCacheOperationOptions} [options]
-   * @returns {Promise<unknown>}
+   * @param {Opts} [options]
+   * @returns {Promise<
+   *   Opts extends { format: infer F }
+   *     ? F extends ModuleFormat
+   *       ? import('./src/types.js').BundleSourceResult<F>
+   *       : import('./src/types.js').BundleSourceResult<'endoZipBase64'>
+   *     : import('./src/types.js').BundleSourceResult<'endoZipBase64'>
+   * >}
```
boneskull: (could) 

```ts
/** @import {BundleSourceResult} from './src/types.js' */
```

packages/bundle-source/cache.js
```diff
@@ -362,17 +362,24 @@ export const makeBundleCache = (wr, cwd, readPowers, opts) => {
...
    */
   const load = async (
     rootPath,
     targetName = readPowers.basename(rootPath, '.js'),
     log = defaultLog,
-    options = {},
+    options = undefined,
```
boneskull, line 376: I've seen this idiom elsewhere. Why do we do it?

tsconfig.eslint-base.json
```diff
@@ -4,11 +4,14 @@
...
+    "incremental": true,
     "noEmit": true,
-    "downlevelIteration": true,
+    "strict": false,
     "strictNullChecks": true,
     "verbatimModuleSyntax": true,
-    "moduleResolution": "NodeNext"
+    "moduleResolution": "NodeNext",
+    "noImplicitAny": false,
+    "types": ["node"],
```
boneskull, line 14: Do we want to do this?  I'd probably lean towards only providing these types for packages which run exclusively under Node.js.

tsconfig.eslint-base.json
```diff
@@ -4,11 +4,14 @@
     "target": "esnext",
     "module": "NodeNext",
     "checkJs": true,
+    "incremental": true,
```
boneskull, line 7: Does this do anything if the projects aren't composite?

boneskull (approved):
no major complaints

packages/bundle-source/cache.js
turadg: Pre-Typescript it was a way to say "this is optional". Without an annotation, TS interprets it as the domain of the parameter. In this case it was to pacify ESlint that complained about optional params being last.

tsconfig.eslint-base.json
turadg: That's a bigger lift. This is just restoring how it worked before TS 6

tsconfig.eslint-base.json
turadg: I think it still caches the buildinfo per package

package.json
turadg: If we used it in multiple packages, yes using catalog would be good. But for now this one suffices.

Issue #2671, opened 2024-12-26
Relax compartment map validation for new, non-behavioral properties


kriskowal:
## What is the Problem Being Solved?

We cannot currently (2024) add new properties to a compartment map because the compartment map validator does not tolerate _any_ unknown properties. This is a deliberate conservative policy to protect the Agoric blockchain from attacks where the behavior of a contract (an application that has been archived with Endo’s compartment mapper) might vary depending on the version of the contract runtime depending on the presence or absence of a property.

We intend to relax this policy, such that the runtime tolerates any new properties that begin with an underscore, like `_sourceSha512` or `_sourceDirname`. These example properties in particular will allow us to improve the debugging experience, generating better `//# sourceMapURL` or `//# sourceURL` comments in the runtime to better converge with source files in a developer IDE.

Any other property addition would require version negotiation to ensure the runtime can effect the intended behavior. For these, we would likely introduce a simple monotonically increasing number `version` non-optional property.

## Description of the Design

We can pre-process compartment maps to omit any properties with the underscore prefix before passing it to either the checker or runtime. This design has the virtue of ensuring that the tolerated properties have no impact on behavior. Some care may be needed to avoid degrading the performance of the initial JSON parse or creation of intermediate object garbage.

## Security Considerations

Described above.

## Scaling Considerations

Should not impact scale, except insofar as that the runtime might gradually expand to cover all or a range of previous versions.

## Test Plan

Cover runtime behavior and validation of compartment maps with or without unknown properties.

## Compatibility Considerations

Described above.

## Upgrade Considerations

The limitation on new properties will continue after the implementation of this feature until all Agoric blockchains have been upgraded to include this change.

PR #3182, opened 2026-04-15
fix: Gracefully handle informative compartment map properties

kriskowal:
Refs: #2671

## Description

This change consists of two adjustments.

1. The compartment mapper will strip all informative compartment map properties, like `__createdBy`, on the basis that they have a prefix `_`.
2. The bundle checker will ignore informative compartment map properties, like `__createdBy`, on the basis that they have a prefix `_`.

The first change addresses a backward incompatibility in the introduction of the `__createdBy` property without removing this instrument entirely. We only remove it when generating bundles. This is analogous to many other properties left behind in this transformation.

The second change will, in the long run, obviate the first. When this change lands in a chain software update for all Agoric chains, new bundles will be able to entrain diagnostic properties, bearing the `_` prefix safely.

### Security Considerations

These changes create a normative invariant that no future property added to the `compartment-map.json` may use the `_` underscore prefix to designate an intended change in behavior. Otherwise, new bundles running on old chains will have different behavior.

### Scaling Considerations

None.

None.

### Testing Considerations

This introduces a test to ensure that we do not introduce new properties to bundled compartment-map.json, at least until we deliberately decide to, in which case the test will need to be deliberately removed.

This restores compatibility with earlier versions of `@endo/check-bundle` as deployed on Agoric chains.

### Upgrade Considerations

None.

### DIFF
--- a/packages/compartment-mapper/src/compartment-map.js
+++ b/packages/compartment-mapper/src/compartment-map.js
@@ -182,9 +182,7 @@ const getModuleConfigurationSpecificProperties = allegedModule => {
     ...other
   } = allegedModule;
   return /** @type {Omit<T, keyof BaseModuleConfiguration>} */ (
-    Object.fromEntries(
-      entries(other).filter(([key]) => !key.startsWith('_')),
-    )
+    Object.fromEntries(entries(other).filter(([key]) => !key.startsWith('_')))
   );
 };
 
--- a/packages/compartment-mapper/src/compartment-map.js
+++ b/packages/compartment-mapper/src/compartment-map.js
@@ -177,12 +177,15 @@ const assertConditions = (conditions, url) => {
  */
 const getModuleConfigurationSpecificProperties = allegedModule => {
   const {
-    __createdBy: _createdBy,
     retained: _retained,
     deferredError: _deferredError,
     ...other
   } = allegedModule;
-  return other;
+  return /** @type {Omit<T, keyof BaseModuleConfiguration>} */ (
+    Object.fromEntries(
+      entries(other).filter(([key]) => !key.startsWith('_')),
+    )
+  );
 };
 
 /**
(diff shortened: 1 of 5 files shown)


packages/compartment-mapper/src/compartment-map.js
```diff
@@ -177,12 +177,13 @@ const assertConditions = (conditions, url) => {
...
   const {
-    __createdBy: _createdBy,
     retained: _retained,
     deferredError: _deferredError,
     ...other
   } = allegedModule;
-  return other;
+  return Object.fromEntries(
+    entries(other).filter(([key]) => !key.startsWith('_')),
+  );
```
boneskull: Mostly a note to myself, but perhaps would be helpful paraphrased in a `TODO` comment:

Currently, `mapNodeModules()` does not execute `assertPackageCompartmentMap()`.  When it _does_ (I already have a branch languishing which does this), this will end up stripping the `__createdBy` prop, which I'd like to keep.

This means I'll need to refactor `getModuleConfigurationSpecificProperties()` so that it is `CompartmentDescriptor`-aware.  The change introduced here should run in the context of a `FileCompartmentDescriptor`, but not a `PackageCompartmentDescriptor`.

Further, the type of a `ModuleConfiguration` within a `FileCompartmentDescriptor` may want to explicitly omit any `_`-prefixed keys, which is another can of worms.

Given this, I need to think about whether some sort of opt-in "debug mode" would be a better alternative (such as storing object-specific metadata in some sort of `WeakMap` and exposing it to the user, avoiding the need to pollute objects with garbage like `__createdBy`).

Anyway, this is _my_ problem; not yours.

turadg (approved):
Shippable. Consider unit tests targeted on the specification that leading underscore properties are omitted.

boneskull (approved):

packages/compartment-mapper/src/compartment-map.js
boneskull: Oh, and, you could dereference `fromEntries` from `Object` like the other static methods near the top of the file.

PR #3186, opened 2026-04-16
fix(compartment-mapper,bundle-source): async parser support

boneskull:
This fixes the half-implemented type defs and support for async parsers by introducing `AsyncParserImplementation`.

We needed sync-and-async-specific parse-mappers in `map-parser.js` and some type guards.

A sync-parser-specific `SyncParserForLanguage` is introduced as well.

_Bonus: added some tests for `import()` in ESM to `@endo/compartment-mapper`._

### DIFF
--- a/packages/compartment-mapper/src/map-parser.js
+++ b/packages/compartment-mapper/src/map-parser.js
@@ -17,6 +17,9 @@
  *   AsyncParseFn,
  *   ParseResult,
  *   ParserForLanguage,
+ *   SyncParserForLanguage,
+ *   ParserImplementation,
+ *   AsyncParserImplementation,
  *   SyncModuleTransform,
  *   SyncModuleTransforms
  * } from './types.js';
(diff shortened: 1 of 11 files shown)


boneskull:

- branch: `master`
  - **[fix(compartment-mapper,bundle-source): async parser support #3186](https://github.com/endojs/endo/pull/3186)** - branch: `boneskull/async-parser` *(this PR)*
    - [feat(evasive-transform): expose makeEvasiveTransformVisitor #3217](https://github.com/endojs/endo/pull/3217) - branch: `boneskull/evasive-pipeline2`
      - [fix(compartment-mapper): policy, parser type fixes #3247](https://github.com/endojs/endo/pull/3247) - branch: `boneskull/fix-module-source-types2`
        - [feat(module-source): fix types, expose analyzer.js subpath #3218](https://github.com/endojs/endo/pull/3218) - branch: `boneskull/module-source-pipeline2`
          - [feat(module-source): expose AST-based CJS parser #3220](https://github.com/endojs/endo/pull/3220) - branch: `boneskull/cjs-ast-parser2`

---

packages/compartment-mapper/src/map-parser.js
naugtur: (should) reuse these from commons

packages/compartment-mapper/src/map-parser.js
```diff
@@ -49,87 +46,55 @@ const has = (object, key) => apply(hasOwnProperty, object, [key]);
...
+  languageForModuleSpecifier,
+) => {
+  const extension = parseExtension(location);
+  if (
+    !extensionImpliesLanguage(extension) &&
+    has(languageForModuleSpecifier, specifier)
+  ) {
+    return languageForModuleSpecifier[specifier];
+  }
+  return languageForExtension[extension] || extension;
```
naugtur: (untrusted input) result of resolveLanguage is trusted later on as if it was an internal key, but it's input from untrusted source. `index.toString` or `index.__proto__` are possible filenames

packages/compartment-mapper/src/map-parser.js
```diff
@@ -49,87 +46,55 @@ const has = (object, key) => apply(hasOwnProperty, object, [key]);
...
+};
 
 /**
- * Produces a `parser` that parses the content of a module according to the
- * corresponding module language, given the extension of the module specifier
- * and the configuration of the containing compartment. We do not yet support
- * import assertions and we do not have a mechanism for validating the MIME type
- * of the module content against the language implied by the extension or file
- * name.
+ * Produces a sync `ParseFn` for use when all parsers and transforms are
```
naugtur, line 173: a lot of commentary is gone from the file. It expressed what's not implemented yet and such. Should stay there.

packages/compartment-mapper/src/map-parser.js
```diff
@@ -174,9 +125,6 @@ const makeExtensionParser = (
           location,
           packageLocation,
           {
-            // At time of writing, sourceMap is always undefined, but keeping
-            // it here is more resilient if the surrounding if block becomes a
-            // loop for multi-step transforms.
             sourceMap,
```
naugtur: is the comment no longer true? Do we have sourceMap support here?

packages/compartment-mapper/src/map-parser.js
```diff
@@ -188,21 +136,19 @@ const makeExtensionParser = (
...
       }
     }
-    if (!has(parserForLanguage, language)) {
+    if (!has(parsers, language)) {
       throw Error(
         `Cannot parse module ${specifier} at ${location}, no parser configured for the language ${language}`,
       );
     }
-    const { parse } = parserForLanguage[language];
+    const { parse } = parsers[language];
```
naugtur: (nit) I liked the redundant but explicit parserForLanguage name for the map.

packages/compartment-mapper/src/map-parser.js
```diff
@@ -220,96 +166,123 @@ const makeExtensionParser = (
...
+ * be asynchronous.
+ *
+ * @param {Record<string, string>} languageForExtension
+ * @param {Record<string, string>} languageForModuleSpecifier
+ * @param {ParserForLanguage} parsers
+ * @param {ModuleTransforms} moduleTransforms
+ * @param {SyncModuleTransforms} syncModuleTransforms
+ * @returns {AsyncParseFn}
+ */
+const makeAsyncExtensionParser = (
```
naugtur: (nit) make*ExtensionParser - they're not parsing extensions tho. `makeAsyncParserForExtension`?

packages/compartment-mapper/src/map-parser.js
```diff
@@ -220,96 +166,123 @@ const makeExtensionParser = (
...
   /**
-   * @type {AsyncParseFn}
+   * @param {Uint8Array} bytes
+   * @param {string} specifier
+   * @param {string} location
+   * @param {string} packageLocation
+   * @param {*} options
+   * @returns {Generator<ReturnType<ModuleTransform> | ReturnType<SyncModuleTransform>, ParseResult | Promise<ParseResult>, any>}
    */
+  function* getParserGenerator(
```
naugtur: why put it on a trampoline if it only exists in an async implementation's scope? I doubt it's more efficient that way. Wait, there's 2 of these. they're functionally identical but the other one is typed to optionally return a promise, and that seems wrong (either wrong types or missing yield to remove the promise involvement)

naugtur:
This is where I'd put things: https://github.com/endojs/endo/commit/46f1fe400bf94eb26c99a9fd5d904f611e84f20d 
(warning - pseudocode in types)

packages/compartment-mapper/src/map-parser.js
boneskull: oops

packages/compartment-mapper/src/map-parser.js
boneskull: can be another PR

packages/compartment-mapper/src/map-parser.js
boneskull: `index`--the `Record` of transform functions--is _also_ untrusted input, so I'm not sure what you're proposing we do.

packages/compartment-mapper/src/map-parser.js
boneskull: it's intentional...

boneskull:
> This is where I'd put things: https://github.com/endojs/endo/commit/46f1fe400bf94eb26c99a9fd5d904f611e84f20d
> (warning - pseudocode in types)

Types _are_ the problem here, though.

The problem is typing `getParserGenerator` properly; we need a `getParserGenerator` that returns a sync parser if we expect a sync parser, and an async parser _or_ sync parser if we don't.  And that depends on whether or not we hand it off to the sync or async trampoline.  One way to ensure this is to provide a narrowly-typed `getParserGenerator` based on whether we expect it to return a sync parser or not necessarily—and that's most easily done with a factory function; now we have a factory for sync and a factory for async.

These two factories could be combined into a single polymorphic factory function w/ overloads and inner conditionals, but that looks like adding even more complexity to me.

So that type-fantasy matches reality, I threw in a guard or two to check we're _actually_ working with sync parsers when we expect to be.

boneskull:
(The alternative is just fudging the types and crossing fingers)

naugtur:
Trampolines are in use here to make it possible for code to be reused between sync and async. Having two copies of the same exact function just to hang different types off of defeats the purpose.

The types for the generator are only consumed by trampolines and don't matter for correctness of what we do here. 

If there is an unavoidable divergence in the two implementations, removed have to be removed

packages/compartment-mapper/src/map-parser.js
naugtur: Totally different category. Extension is untrusted input for the user of Endo, it's coming from untrusted packages. We need to either use a fixed listy of what is available and throw early here, or very carefully handle the return value to not match prototype fields. The former would be my choice

boneskull:
@naugtur 

> Trampolines are in use here to make it possible for code to be reused between sync and async. Having two copies of the same exact function just to hang different types off of defeats the purpose.

I see what you're saying here.  I'll take another look at it.

boneskull:
@naugtur OK, I've reverted most of my changes to `map-parser.js` and added a type guard.  The types are still not where I'd like them to be, but there's little we could do other than add a type assertion:

```js
/**
 * @typedef GetSyncParserGeneratorFn
 * @param {ParserGeneratorConfig} config
 * @param {Uint8Array} bytes
 * @param {string} specifier
 * @param {string} location
 * @param {*} options
 * @returns {Generator<ReturnType<SyncModuleTransform>, ParseResult, any>}
 */

const result = syncTrampoline(
  /** @type {GetSyncParserGeneratorFn} */(getParserGenerator),
  config,
  bytes,
  specifier,
  location,
  packageLocation,
  options,
);
```

I don't view the above as particularly valuable, so I didn't do it.

packages/compartment-mapper/src/map-parser.js
boneskull: Where is it being used as if it is trusted?  `has` omits unowned properties.

boneskull: I'm _pretty sure_ this is not a problem in practice, but I've added some extra defense anyway:

```js
const resolveLanguage = (
  specifier,
  location,
  languageForExtension,
  languageForModuleSpecifier,
) => {
  const extension = parseExtension(location);
  if (
    !extensionImpliesLanguage(extension) &&
    has(languageForModuleSpecifier, specifier)
  ) {
    return languageForModuleSpecifier[specifier];
  }
  // new check
  if (has(languageForExtension, extension)) {
    return languageForExtension[extension];
  }
  return extension;
};
```

So `resolveLanguage` cannot return anything untoward on `languageForExtension`.  Further:

```js
const validateLanguageForExtension = (
  languageForExtension,
  parserForLanguage,
) => {
  const languageForExtensionEntries = [];
  const problems = [];
  for (const [extension, language] of entries(languageForExtension)) {
    if (has(parserForLanguage, language)) {
      languageForExtensionEntries.push([extension, language]);
    } else {
      problems.push(`${q(language)} for extension ${q(extension)}`);
    }
  }
  if (problems.length > 0) {
    throw Error(`No parser available for language: ${problems.join(', ')}`);
  }
  // null-proto object
  return assign(create(null), fromEntries(languageForExtensionEntries));
};
```

Now `languageForExtension` shouldn't even _have_ such properties.

naugtur:
looks good, I need one more pass tomorrow, but no comments yet.

naugtur (approved):

PR #3189, opened 2026-04-17
Revert "fix(pass-style): retype PASS_STYLE as string literal for decl-emit

mhofman:
This partially reverts commit c05c9a884fd6f2f0888bd954ea8295af7020146f.

Refs: #3184

## Description

It's unclear to me whether this is an agoric-sdk bug, or a TS bug, but let's not propagate a hack to the world just yet

### DIFF
--- a/.changeset/pass-style-decl-emit-2.md
+++ b/.changeset/pass-style-decl-emit-2.md
@@ -1,5 +1,5 @@
 ---
-'@endo/pass-style': minor
+'@endo/pass-style': patch
 ---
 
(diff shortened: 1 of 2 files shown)


.changeset/pass-style-decl-emit-2.md
```diff
@@ -0,0 +1,7 @@
+---
+'@endo/pass-style': minor
```
mhofman: Actually I think that type only changes qualify as patch.

```suggestion
'@endo/pass-style': patch
```

turadg:
Closing per our Endo meeting conclusion that the hack is sufficiently encapsulated by Endo's utility types

PR #3204, opened 2026-04-22
fix(daemon)!: Detect truncation of Unix domain socket paths

### LINKED ISSUE #3203: endo daemon fails to detect Unix domain socket path truncation
## Describe the bug

On some platforms, an overly-long Unix domain socket path is silently truncated and the [`server.listen`](https://nodejs.org/docs/latest/api/net.html#serverlisten) callback in [`servePath`](https://github.com/endojs/endo-carbon-villa/blob/377047757442411b1f14ae68d3d7beda4f32d10d/packages/daemon/src/daemon-node-powers.js#L100) is invoked without an error—generally resulting in a running server connected to a misplaced UDS (which therefore won't receive messages) that doesn't get cleaned up (which results in subsequent start attempts to encounter EADDRINUSE).

## Steps to reproduce

Checkout endo to a long path name, `cd` to packages/daemon, run `yarn test test/endo.test.js`, and observe the EADDRINUSE errors.

## Expected behavior

We can't prevent the socket path truncation, but we _can_ detect it and report the failure.

## Platform environment

Observed in CI runs for Linux and macOS.

gibson042:
Fixes #3203

## Description

Use [`access`](https://nodejs.org/docs/latest/api/fs.html#fspromisesaccesspath-mode) to verify existence of a Unix domain socket at the requested path after it should have been created by [`server.listen`](https://nodejs.org/docs/latest/api/net.html#serverlisten), and reject the server-is-ready promise if such verification fails.

Also includes related improvements to test files and GitHub Actions CI.

### Security Considerations

None known.

### Scaling Considerations

n/a

Covered by an existing warning which is being extended to include path truncation.

### Testing Considerations

Just a new unit test.

This is a breaking change because some functions now require an `access` power from node:fs/promises. We could preserve backwards compatibility by instead making the new power optional, but I'd rather maximize our ability to detect what is otherwise a silent failure.

### Upgrade Considerations

The daemon is not used in production AFAIK, but direct imports may need an update to provide the required `access` function.

### DIFF
--- a/.changeset/lucky-views-train.md
+++ b/.changeset/lucky-views-train.md
@@ -1,0 +1,5 @@
+---
+'@endo/daemon': major
+---
+
+Detect truncation of Unix domain socket paths
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,16 +87,36 @@ jobs:
           node-version: ${{ matrix.node-version }}
           cache: yarn
 
+      # Some tests (e.g., in package "daemon") create Unix domain sockets, and
+      # their full absolute paths MUST remain short.
+      - name: Move working directory
+        id: move-wd
+        run: |
+          ORIGINAL_WORKDIR="$(pwd)"
+          echo "ORIGINAL_WORKDIR=$ORIGINAL_WORKDIR" >> $GITHUB_ENV
+          cd "$RUNNER_TEMP"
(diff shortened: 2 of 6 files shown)


erights:
Do you understand the CI failures?

gibson042:
> Do you understand the CI failures?

Yes; should be fixed now.

kriskowal (approved):

PR #3208, opened 2026-04-25
feat(hex): Hexadecimal transcoder

kriskowal:
## Description

This is a hardened ponyfill for hexadecimal encoding, which we already use in OCapN and will need for byte array transcoding in our marshal packages.

### Security Considerations

None.

### Scaling Considerations

None.

Just the README.

### Testing Considerations

Basic testing included.

None.

### Upgrade Considerations

None.


packages/hex/src/decode.js
```diff
@@ -0,0 +1,103 @@
...
+ * Pure-JavaScript hex decoder, exported for benchmarking and for
+ * environments where the native TC39 `Uint8Array.fromHex` intrinsic
+ * (proposal-arraybuffer-base64) is unavailable or has been removed.
+ * See `decodeHex` below for the dispatched default.
+ *
+ * Accepts both upper- and lowercase input.  Throws on odd-length input
+ * and on any character outside `[0-9a-fA-F]`.
+ *
+ * Computes nibble values directly from character codes rather than
+ * indexing a lookup table.  On Node 22, this is roughly 2.5 to 3 times
```
gibson042: What about on XS? Generally speaking, I think we tend to care much more about that because slowdowns hurt much more in an already-slow engine, especially for functionality that will be provided natively by even our oldest supported Node.js in relatively short order.

packages/hex/src/decode.js
```diff
@@ -0,0 +1,103 @@
...
+ * faster than the table-based decoder for ~1 MiB inputs and avoids any
+ * module-scope mutable data; see `test/decode.bench.js`.
+ *
+ * @param {string} string
+ * @param {string} [name] Name of the string for error diagnostics.
+ * @returns {Uint8Array}
+ */
+export const jsDecodeHex = (string, name = '<unknown>') => {
```
gibson042, line 32: The `name` parameter takes this beyond a ponyfill. Also, we have a number of such ad hoc error labeling/annotating approaches and I wonder if a dedicated effort to come up with something uniform would bear fruit. If nothing else, we could start to capture requirements.

packages/hex/src/encode.js
```diff
@@ -0,0 +1,52 @@
...
+ * Encodes a Uint8Array as a lowercase hex string.
+ *
+ * Dispatches to the native `Uint8Array.prototype.toHex` intrinsic when
+ * available (stage-4 TC39 proposal-arraybuffer-base64).  Otherwise
+ * falls through to the pure-JavaScript polyfill.
+ *
+ * @type {typeof jsEncodeHex}
+ */
+export const encodeHex =
+  nativeToHex !== undefined ? bytes => nativeToHex.call(bytes) : jsEncodeHex;
```
gibson042: We should continue to prefer use of a captured `Reflect.apply` over `.call`, even where the latter is assumed to be primordial.
```suggestion
  nativeToHex !== undefined ? bytes => apply(nativeToHex, bytes, []) : jsEncodeHex;
```

packages/hex/test/decode.bench.js
```diff
@@ -0,0 +1,185 @@
...
+
+/**
+ * @param {number} size
+ * @param {number} iters
+ */
+const runSize = (size, iters) => {
+  const bytes = makeBytes(size);
+  const hex = encode(bytes);
+
+  // Sanity check both decoders agree on clean input.
```
gibson042, line 262: I wonder if this should cover _multiple_ alternate table-based implementations, e.g. Map\<Character, number> and/or Map\<CharacterPair, number> and/or null-prototype object analogs?

packages/hex/package.json
```diff
@@ -0,0 +1,82 @@
+{
+  "name": "@endo/hex",
+  "version": "1.0.0",
+  "description": "Transcodes hexadecimal",
+  "keywords": [
+    "hex",
```
gibson042, line 6:
```suggestion
    "hex",
    "hexadecimal",
```

gibson042 (approved):
I think there's also an existing implementation hanging around in some package that could presumably now be replaced with a dependency on this package.

packages/bundle-source/cache.js
```diff
@@ -452,7 +453,12 @@ export const makeNodeBundleCache = async (
   }
 
   const readPowers = {
-    ...makeReadPowers({ fs, url, crypto }),
+    ...makeReadPowers({
+      fs,
+      url,
+      crypto,
+      computeSha512: makeComputeSha512(crypto),
```
kriskowal: My preference is to subsume the responsibility of using `makeComputeSha512` into `makeReadPowers`. The invariant is that it should be easy to just give this function Node.js modules and it will tease out the platform-neutral API from them.

packages/compartment-mapper/package.json
```diff
@@ -45,6 +45,7 @@
     "./script-lite.js": "./script-lite.js",
     "./node-powers.js": "./node-powers.js",
     "./node-modules.js": "./node-modules.js",
+    "./sha512-hex.js": "./sha512-hex.js",
```
kriskowal: Let’s not expose this.

PR #3194, opened 2026-04-18
ocapn: guile spritely interop test + goblin-chat tui client

kumavis:
This adds a CI test against Spritely's Guile implementation

It also adds a new package `goblin-chat` including a tui for working with https://codeberg.org/spritely/goblin-chat/ built on ocapn

PR #3209, opened 2026-04-25
ocapn: guile spritely interop test + goblin-chat tui client

kumavis:
pending https://github.com/endojs/endo/pull/3207 (this to be rebased on top)

## Summary

This adds a CI test against Spritely's Guile implementation.

It also adds a new package `goblin-chat` including a TUI for working with https://codeberg.org/spritely/goblin-chat/ built on OCapN.

⚠️ This adds a new workflow that builds and tests against Guile Goblins (typical runtime is 3m20s). This workflow could be change to be gated behind PR changes to the `ocapn` package.

## Test plan

- [x] CI: `ocapn-guile-interop` workflow exercises bilateral chatroom interop against guile-goblins via `guix shell`
- [x] CI: `@endo/ocapn` and `@endo/goblin-chat` package tests pass (incl. websocket netlayer tests and the all-JS interop self-test)
- [x] Manual: `goblin-chat` TUI hosts a room, prints a sturdyref URI, and a second client can join it

## Notes for reviewers

- Final tree is identical to #3194; this PR is a history cleanup, not a content change. Consider closing #3194 in favor of this one.
- Intermediate commits 1 and 2 are not yarn-install-clean on their own (the lockfile lives in commit 3 with the bulk of the new deps); only the final tree is verified.

### DIFF
--- a/packages/goblin-chat/src/animated-logo.js
+++ b/packages/goblin-chat/src/animated-logo.js
@@ -1,0 +1,343 @@
+// @ts-check
+/* eslint-disable no-bitwise, no-continue, import/no-unresolved */
+/* global performance, setInterval, clearInterval */
+
+/**
+ * Animated ASCII logo for the goblin-chat TUI.
+ *
+ * Renders the "OCAPN CHAT" figlet as a 2D glyph map and shades each
+ * non-space cell every frame using a small lighting model:
+ *
+ *   - **Diagonal sweep**: a Gaussian band of bright light travels
+ *     across the logo along a vector that itself slowly rotates, so
+ *     successive sweeps approach from slightly different angles.
+ *   - **Counter sweep**: a fainter, warmer band crosses the opposite
+ *     way at a different speed. The two never quite line up, so the
+ *     interference pattern never visibly repeats.
+ *   - **Volumetric density**: each cell's "matter density" is the
+ *     fraction of non-space neighbours in a 5×3 stencil, computed
+ *     once at module load. Dense regions glow more, mimicking
+ *     subsurface scattering — letter bodies fluoresce, decorative
+ *     thin strokes stay cooler.
+ *   - **Plasma shimmer**: three layered sin/cos terms with
+ *     incommensurate frequencies add a low-amplitude wandering
+ *     perturbation. Keeps the surface alive even when the sweeps
+ *     are far away.
+ *   - **Specular desaturation**: peaks of the sweep bleach toward
+ *     white instead of just brightening the hue, so the highlight
+ *     reads as a hot reflection rather than a colour shift.
+ *   - **Breath**: a long-period (~9s) global brightness oscillation
+ *     gives the whole sign a slow lung-like rise and fall.
+ *   - **Sparkle**: a deterministic per-cell hash gates a brief
+ *     overshoot when both the sweep and the sin-locked sparkle clock
+ *     line up. Most cells never sparkle; a handful pop occasionally.
+ *
+ * Output is 24-bit colour (`\u001B[38;2;R;G;Bm`). Each row is built
+ * as one string with the per-cell escapes embedded, then handed to a
+ * single `<Text>` node. Spaces emit no escape so the row stays sparse.
+ *
+ * Falls back to a plain "goblin-chat" line when the terminal is too
+ * narrow to fit the glyph (avoids ugly mid-line wraps).
+ */
+
+import React, { useEffect, useRef, useState } from 'react';
+import { Box, Text } from 'ink';
+
+const h = React.createElement;
+
+// The ASCII art. `String.raw` preserves backslashes verbatim so the
+// source matches the rendered glyphs character-for-character. Backticks
+// (which would terminate the template) are interpolated as `${BT}`.
+const BT = '`';
+const LOGO_LINES = [
+  String.raw`   )\.-.      .-./(     /(,-.   .')      .'(   )\  )\          )\.-.       .'(     /${BT}-.   .-,.-.,-. `,
+  String.raw` ,' ,-,_)   ,'     )  ,' _   ) ( /       \  ) (  \, /        ,' ,-,_)  ,') \  )  ,' _  \  ) ,, ,. ( `,
+  String.raw`(  .   __  (  .-, (  (  '-' (   ))       ) (   ) \ (        (  .   _  (  '-' (  (  '-' (  \( |(  )/ `,
+  String.raw` ) '._\ _)  ) '._\ )  )  _   )  )'._.-.  \  ) ( ( \ \        ) '..' )  ) .-.  )  )   _  )    ) \    `,
+  String.raw`(  ,   (   (  ,   (  (  '-' /  (       )  ) \  ${BT}.)/  )      (  ,   (  (  ,  ) \ (  ,' ) \    \ (    `,
+  String.raw` )/'._.'    )/ ._.'   )/._.'    )/,__.'    )/     '.(        )/'._.'   )/    )/  )/    )/     )/    `,
+];
+
+const LOGO_WIDTH = Math.max(...LOGO_LINES.map(l => l.length));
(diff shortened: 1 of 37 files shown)


packages/goblin-chat/src/animated-logo.js
```diff
@@ -0,0 +1,341 @@
...
+
+// The ASCII art. Carefully transcribed — single-quoted strings with
+// `\\` for backslash, `\'` for apostrophe; backticks pass through.
+const LOGO_LINES = [
+  "   )\\.-.      .-./(     /(,-.   .')      .'(   )\\  )\\          )\\.-.       .'(     /`-.   .-,.-.,-. ",
+  " ,' ,-,_)   ,'     )  ,' _   ) ( /       \\  ) (  \\, /        ,' ,-,_)  ,') \\  )  ,' _  \\  ) ,, ,. ( ",
+  "(  .   __  (  .-, (  (  '-' (   ))       ) (   ) \\ (        (  .   _  (  '-' (  (  '-' (  \\( |(  )/ ",
+  " ) '._\\ _)  ) '._\\ )  )  _   )  )'._.-.  \\  ) ( ( \\ \\        ) '..' )  ) .-.  )  )   _  )    ) \\    ",
+  "(  ,   (   (  ,   (  (  '-' /  (       )  ) \\  `.)/  )      (  ,   (  (  ,  ) \\ (  ,' ) \\    \\ (    ",
+  " )/'._.'    )/ ._.'   )/._.'    )/,__.'    )/     '.(        )/'._.'   )/    )/  )/    )/     )/    ",
```
kriskowal: Might be more legible with a `String.raw` template literal.

packages/goblin-chat/src/animated-logo.js
```diff
@@ -0,0 +1,341 @@
+// @ts-check
```
kriskowal, line 1: Gratuitous (laudable)!

packages/goblin-chat/src/clipboard.js
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * @returns {Array<[string, string[]]>}
+ */
+const platformCandidates = () => {
+  if (process.platform === 'darwin') {
+    return [['pbcopy', []]];
+  }
+  if (process.platform === 'win32') {
+    return [['clip', []]];
```
kriskowal, line 88: Today I learned that the gospel arrived in Windows. I tried to make one of these in win32 API in college, but it wasn’t possible at the time. It became possible with .Net and this is the only program I wrote in C♯.

packages/goblin-chat/src/clipboard.js
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * Copy `text` to the system clipboard. The promise always resolves —
+ * callers should branch on the returned `ok` flag. The first candidate
+ * that succeeds wins; if none does, the last error is returned so the
+ * caller can surface it (e.g. into a log panel).
+ *
+ * @param {string} text
+ * @returns {Promise<{ ok: true } | { ok: false, error: unknown }>}
+ */
+export const copyToClipboard = async text => {
```
kriskowal, line 109: There are a variety of these on `npm` and I do not mind rolling our own. I would not object to ejecting this to a package and making it a “hardened module” so we can encourage folks to level up their supply chain attack resistance with LavaMoat.

packages/goblin-chat/src/host-room.js
```diff
@@ -0,0 +1,300 @@
...
+ *      rooms in persistent history should treat self-hosted URIs as
+ *      transient (see the `transient` flag on `useGoblinChat.joinRoom`).
+ *
+ * The module also keeps a process-wide `hostRegistry` so the entrypoint
+ * can shut every host cleanly on process exit / SIGINT / SIGTERM. The
+ * websocket server holds the event loop open, so missing this teardown
+ * means a clean Ctrl+C from the menu wouldn't actually exit.
+ */
+
+import { Buffer } from 'node:buffer';
```
kriskowal: Mild preference to proactively decrease coupling to Node.js by not using Buffer. Maybe we can add base64url support to the existing `@endo/base64`.

packages/goblin-chat/src/host-room.js
```diff
@@ -0,0 +1,300 @@
...
+};
+
+/**
+ * Build the user-facing sturdyref URI for a hosted chatroom.
+ *
+ * @param {OcapnLocation} location
+ * @param {string} swissStr
+ * @returns {string}
+ */
+const buildSturdyRefUri = (location, swissStr) => {
```
kriskowal: I’m increasingly dubious about calling locators sturdy refs. I would prefer `formatLocator` for this, to be consistent with similar functions in the Dæmon.

packages/goblin-chat/src/interop-driver.js
```diff
@@ -0,0 +1,163 @@
...
+        sentLocalMessageAck = true;
+        log(`[${name}] sent message, ack = ${JSON.stringify(ack)}`);
+        finishIfReady();
+      })
+      .catch(err => {
+        rejectDone(err);
+      });
+  };
+
+  const observer = Far('interop-observer', {
```
kriskowal: Prefer makeExo with well-defined interface guards.

packages/goblin-chat/src/state-store.js
```diff
@@ -0,0 +1,251 @@
...
+  const recordJoin = (uri, displayName) => {
+    const now = new Date().toISOString();
+    const existing = state.recentRooms.find(r => r.uri === uri);
+    /** @type {RecentRoom} */
+    const updated = {
+      uri,
+      lastJoinedAt: now,
+      ...((displayName ?? existing?.displayName)
+        ? { displayName: displayName ?? existing?.displayName }
+        : {}),
```
kriskowal: This can be simplified considerably, but may require relaxing a TypeScript rule. My preference is to relax the TS rule so this can be a simple destructure to a maybe-undefined/maybe-absent property.

packages/goblin-chat/src/uri-parse.js
```diff
@@ -0,0 +1,130 @@
...
+ *
+ * The swissnum, when present, is base64url(no-padding) of the raw
+ * swissnum bytes per the OCapN spec (`draft-specifications/Locators.md`)
+ * and the Spritely Goblins reference implementation
+ * (`goblins/ocapn/ids.scm`). No other encodings are accepted.
+ *
+ * @param {string} uri
+ * @returns {ParsedOcapnUri}
+ */
+export const parseOcapnUri = uri => {
```
kriskowal: Prefer `parseLocator` to go with `formatLocator`. This avoids the debate (and the burden to remember) whether we like `Uri` more than `Url`.

packages/goblin-chat/src/uri-parse.js
```diff
@@ -0,0 +1,130 @@
...
+  const trimmed = uri.trim();
+  const match = trimmed.match(/^ocapn:\/\/([^/?#]+)(\/[^?#]*)?(\?[^#]*)?$/);
```
kriskowal: I like to rely on the `URL` parser because URLs are considerably more sophisticated than readily expressible in a regular expression.

packages/goblin-chat/test/interop-self.test.js
```diff
@@ -0,0 +1,112 @@
...
+ * helper end-to-end against the real OCapN websocket transport (not
+ * just an in-process Far reference).
+ */
+
+import '@endo/init';
+
+import test from '@endo/ses-ava/test.js';
+
+import { makeClient, swissnumFromBytes } from '@endo/ocapn';
+import { makeWebSocketNetLayer } from '@endo/ocapn/src/netlayers/websocket.js';
```
kriskowal: Prefer exporting this explicitly as `@endo/ocapn/netlayer/ws`. We should not reach arbitrarily into `src`, and we should have an `"exports"` that prevents it. I’m surprised this works.

kriskowal (approved):
Feedback for your discretion.

PR #3214, opened 2026-04-28
fix(ses): fix bug #3202 from PR #3153

### LINKED ISSUE #3202: ses@2.0.0: setFloat16 / setFloat32 write zero bytes for NaN (tame-nan-sidechannel uses 64-bit canonical pattern for 16/32-bit writes)
Note: Claude prepared this issue for me; i discovered the failures by upgrading ses to v2 in es-abstract.

## Summary

In `ses@2.0.0`, `DataView.prototype.setFloat32(offset, NaN, …)` and `setFloat16(offset, NaN, …)` silently write all-zero bytes instead of a canonical NaN bit pattern. `setFloat64(offset, NaN, …)` is unaffected. This is a regression introduced by the new unconditional `tameNaNSideChannel` in `lockdown()` (there is no config option to disable it).

## Reproducer

```js
require('ses');
lockdown({ errorTaming: 'unsafe' });
const dv = new DataView(new ArrayBuffer(8));
dv.setFloat32(0, NaN, true);
console.log(Array.from(new Uint8Array(dv.buffer, 0, 4))); // [0, 0, 0, 0] — should be a NaN pattern, e.g. [0, 0, 192, 127]
console.log(dv.getFloat32(0, true));                       // 0 — should be NaN
```

Observed output with `ses@2.0.0` on Node:

```
Float32 NaN bytes: [ 0, 0, 0, 0 ]
Float32 readback: 0
Float64 NaN bytes: [ 0, 0, 0, 0, 0, 0, 248, 127 ]
Float64 readback: NaN
```

`setFloat64` produces the correct canonical NaN bytes `[0,0,0,0,0,0,248,127]` and reads back as `NaN`.

## Root cause

`packages/ses/src/tame-nan-sidechannel.js` defines:

```js
const canonicalNaN = 0x7ff8000000000000n; // 64-bit pattern
```

and the `setFloat32` wrapper does:

```js
return apply(setUint32, this, [byteOffset, Number(canonicalNaN), littleEndian]);
```

`Number(0x7ff8000000000000n)` is `9221120237041091000`, which is far outside the Uint32 range. `ToUint32(9221120237041091000)` is `0`, so four zero bytes get written. The same bug applies to the `setFloat16` branch via `setUint16`.

```
Number(0x7ff8000000000000n) === 9221120237041091000
Number(0x7ff8000000000000n) >>> 0 === 0            // setUint32 result
Number(0x7ff8000000000000n) & 0xffff === 0         // setUint16 result
```

`setFloat64` is unaffected because it passes the BigInt directly to `setBigUint64`.

## Expected fix

Use the appropriately-sized canonical NaN bit pattern for each precision rather than `Number`-coercing the 64-bit pattern:

- Float32 canonical quiet NaN: `0x7fc00000` → pass to `setUint32`
- Float16 canonical quiet NaN: `0x7e00` → pass to `setUint16`

## Impact

Any code that relies on `setFloat16` / `setFloat32` preserving NaN (round-tripping, writing canonical NaN bytes, etc.) silently produces zero values after `lockdown()`. Discovered while upgrading `es-abstract` from `ses@1` to `ses@2`; 107 tests in its `test:ses` suite fail for this single root cause.

## Workaround

None via configuration — `tameNaNSideChannel()` is called unconditionally from `lockdown.js`. Users must pin `ses@^1`.

erights:
Closes: #3202
Refs: #3153

## Description

Previously (#3153), I only implemented the float64 case correctly and I only tested the float64 case at all.

#3202 points out that the bit encoding of the 32 bit canonical NaN is (of course!) 32 bits. And the bit encoding of the 16 bit canonical NaN is 16 bits. This PR fixes this as suggested in #3202, breaks out one test file for each concrete size, and adds the reproducer test from #3202 as adapted to each. 

Note: Where the #3202 reproducer says `should be a NaN pattern, e.g. [0, 0, 192, 127]` it seems it should actually be `[0, 0, 248, 127]`. See https://github.com/endojs/endo/issues/3202#issuecomment-4331905523

### Security Considerations

The NaN side channel leak was plugged by #3153, i.e., it fixed the security problem. It just broke functionality while doing so. This PR leaves the side channel leak identically plugged, but also restores the functionality.

### Scaling Considerations

none

none. It is just a bug fix for a bug starting in #3153.

### Testing Considerations

Better testing, including as suggested in #3202 

The bug was a compat break since it observably did the wrong thing. This PR repairs the compat break.

### Upgrade Considerations

none.


erights:
@ljharb, I cannot add you as a reviewer, but I would still appreciate your review.

packages/ses/src/tame-nan-sidechannel.js
```diff
@@ -24,7 +24,11 @@ const {
 // that this is the canonical NaN for web standards.
 // Casual googling stongly suggests that this is also the cosmWasm
 // canonical NaN. But I have not yet found an authoritative page stating this.
-const canonicalNaN = 0x7ff8000000000000n;
+const canonicalNaN64Encoding = 0x7ff8000000000000n;
```
ljharb, line 45: i'm not sure what these encodings are - I usually think of NaNs as an array of 4 bytes. There are many, but `[0, 0, 192, 127]` is the one I see most often on my Mac. (endianness matters, ofc)

packages/ses/src/tame-nan-sidechannel.js
```diff
@@ -24,7 +24,11 @@ const {
...
+// eslint-disable-next-line no-bitwise
+const canonicalNaN32Encoding = Number(canonicalNaN64Encoding >> 32n);
+// eslint-disable-next-line no-bitwise
+const canonicalNaN16Encoding = canonicalNaN32Encoding >>> 16;
```
gibson042: These don't seem right to me... as noted in [Web IDL](https://webidl.spec.whatwg.org/#js-unrestricted-double), 0x7ff8000000000000 is chosen as canonical because it is the lowest-valued binary64 quiet NaN. But since binary32 and binary16 have smaller exponent fields, equivalent reasoning would instead yield
```suggestion
// By analogous construction, define canonical binary32 and binary16 NaNs as the
// respective lowest-valued quiet NaNs (sign bit 0 and a trailing significand
// field in which only the most significant bit [`is_quiet`] is set).
const canonicalNaN32Encoding = 0x7fc00000;
const canonicalNaN16Encoding = 0x7e00;
```

gibson042 (changes requested):

packages/ses/src/tame-nan-sidechannel.js
gibson042: This is a bigint whose big-endian uint64 bit pattern corresponds with the big-endian bit pattern for the Web IDL canonical binary64 NaN.

The other constants are numbers whose big-endian {uint32,uint16} bit patterns correspond with the big-endian bit pattern for what this PR is choosing as canonical {binary32,binary16} NaNs (respectively).

ljharb: ok, but that's not the bit pattern that a literal `NaN` produces, so i'm not sure why it'd be "canonical"?

packages/ses/src/tame-nan-sidechannel.js
erights: Hi @gibson042 , I understand the analogy, but I disagree with the conclusion. `getFloat*` always widens to a float64 `number` since those are the only first class float values in JS. Granted, when `getFloat*` returns a NaN, which NaN is not visible. But IMO the corresponding `getUint*` should therefore return just a shift of the 64bit canonical NaN encoding because anything else is too confusing. As long as that does not violate any other correctness property of course, but I don't see a conflict.

I see you requested changes. If the above argument does not sway you, we need to discuss. Thanks.

packages/ses/src/tame-nan-sidechannel.js
erights: > ok, but that's not the bit pattern that a literal NaN produces,

What does this mean? Are you speaking about the spec or about implementations?

> so i'm not sure why it'd be "canonical"?

@gibson042 's citation of WebIDL is adequate. IIRC I also found it in other relevant places such as cosmwasm (a deterministic wasm for blockchain).

erights: > i'm not sure what these encodings are - I usually think of NaNs as an array of 4 bytes. There are many, but `[0, 0, 192, 127]` is the one I see most often on my Mac. (endianness matters, ofc)

Just verified that for the WebIDL canonical NaN which is what this PR uses, `[0, 0, 248, 127]` is correct. So I'm sticking with that too.

ljharb: Because if i put a NaN into a Float32Array, and then get the bytes back out - `new Uint8Array(new Float32Array([NaN]).buffer)` - i get `[0, 0, 192, 127]`. "Canonical" is defined by "what implementations do", not "what a document somewhere says" :-)

packages/ses/src/tame-nan-sidechannel.js
ljharb: It "should" do whatever it already _does_ do, regardless of whether it's confusing or not.

gibson042: So IIUC, this comes down to a question of what should be returned by e.g. `view.getUint16(offset)` after `view.setFloat16(offset, NaN)`, where the current state of the PR returns a number corresponding to the [16-bit] prefix of the ecosystem-canonical binary64 NaN and my suggestion is to instead return a number corresponding to an n-bit NaN constructed by reasoning equivalent to that of the 64-bit NaN.

This is sufficiently esoteric that it will probably never matter either way, but it's probably worth noting that the prefix-based approach will **not** hold if DataView is ever extended to support binary128 and/or binary256, where the leading bits of a NaN must be *{0,1}111_1111_1111_1111* = *0x{7,F}FFF* and *{0,1}111_1111_1111_1111_1111* = *0x{7,F}FFFF* (respectively). So if that day comes, the truncation invariant will be broken anyway.

But to reïterate, this will probably never matter. I still prefer the analogous-construction approach, although that preference does not rise to the level of a blocking concern if you still want to adopt the truncation approach. I'm removing the "changes requested" status.

gibson042 (approved):

packages/ses/src/tame-nan-sidechannel.js
gibson042: @ljharb it depends upon _which_ NaN you start with, and the whole point of this PR is to remove such sensitivity.
```console
$ eshost -se '
  [NaN, Infinity - Infinity, 0 / 0, Math.acos(2), Math.log(-1), Math.sqrt(-1)]
  .map(nan => (new Uint8Array(new Float32Array([nan]).buffer)).join(", "))
  .join("\n")
'
#### engine262, LibJS, Moddable XS
0, 0, 192, 127
0, 0, 192, 127
0, 0, 192, 127
0, 0, 192, 127
0, 0, 192, 127
0, 0, 192, 127

#### GraalJS
0, 0, 192, 127
0, 0, 192, 255
0, 0, 192, 255
0, 0, 192, 255
0, 0, 192, 255
0, 0, 192, 255

#### Hermes, QuickJS
0, 0, 192, 127
0, 0, 192, 255
0, 0, 192, 255
0, 0, 192, 127
0, 0, 192, 127
0, 0, 192, 255

#### JavaScriptCore
0, 0, 192, 127
0, 0, 192, 255
0, 0, 192, 255
0, 0, 192, 127
0, 0, 192, 255
0, 0, 192, 255

#### SpiderMonkey
0, 0, 192, 127
0, 0, 192, 255
0, 0, 192, 127
0, 0, 192, 255
0, 0, 192, 255
0, 0, 192, 255

#### V8
0, 0, 192, 127
0, 0, 192, 255
0, 0, 192, 127
0, 0, 224, 127
0, 0, 224, 127
0, 0, 192, 255
```

ljharb: Right - I'm suggesting that the canonical one to start with should be the literal `NaN`. Anything else is a NaN produced by math, which isn't the canonical one to me.

gibson042: Well, that also varies by the element size into which it's being written.
```console
$ eshost -se '
  [Float64Array, Float32Array]
  .map(Ctor =>
    Array.from(new Uint8Array(new Ctor([NaN]).buffer), octet =>
      octet.toString(16).padStart(2, "0"),
    ).join(""),
  )
  .join("\n")
'
#### engine262, GraalJS, Hermes, JavaScriptCore, LibJS, Moddable XS, QuickJS, SpiderMonkey, V8
000000000000f87f
0000c07f

$ eshost -se 'globalThis.Float16Array && Array.from(
    (new Uint8Array(new Float16Array([NaN]).buffer)),
    octet => octet.toString(16).padStart(2, "0"),
  ).join("")
'
#### engine262, GraalJS, Hermes, LibJS
undefined

#### JavaScriptCore, Moddable XS, SpiderMonkey, V8
007e

#### QuickJS
017c
```

Interestingly, only QuickJS deviates from the [analogous-construction approach](https://github.com/endojs/endo/pull/3214#discussion_r3155852021) that I'm recommending, and even then only for 16-bit elements (0x7fc00000 is the lowest-valued binary32 quiet NaN, and 0x7e00 is the lowest-valued binary16 quiet NaN—0x7c01 is a _signaling_ NaN).

ljharb: right - the bit size certainly dictates which pattern is the canonical one, but "from a literal NaN" imo remains the correct input to use.

packages/ses/src/tame-nan-sidechannel.js
gibson042, line 236: [I just realized](https://github.com/endojs/endo/pull/3214#discussion_r3157706795) that #3153 presumed binary16 support, which is not yet present in at least one implementation that _we_ support ([Hermes v0.16](https://github.com/facebook/hermes/blob/release-v0.16/include/hermes/VM/TypedArrays.def)[^1]), so we should avoid `defineProperty` in that case.
```suggestion
export const tameNaNSideChannel = () => {
  for (const [name, method] of entries(methods)) {
    // setFloat16 is not available in Hermes as of v0.16.
    if (!method) continue;
    defineProperty(dataViewPrototype, name, {
      // Since we're redefining properties that already exist, by omitting the
      // other descriptor attributes here, they are unchanged.
      value: method,
    });
  }
};
```

[^1]: although [it is coming](https://github.com/facebook/hermes/commit/5da1ba6d06e1dafe610f32ad028ba1fdb4731008)

erights: It is also not present on earlier versions of Node that we still support. So yes.

packages/ses/src/tame-nan-sidechannel.js
erights: What are the results across platforms for float64 and (where supported) float16?

gibson042: All implementations agree that the literal `NaN` gets represented as binary64 **0x7ff8000000000000**, binary32 **0x7fc00000**, and (with the almost surely unintentional exception of QuickJS) binary16 **0x7e00**. As for the representation of NaN values with _non_-literal provenance, https://github.com/endojs/endo/pull/3214#discussion_r3157617380 shows some tremendous intra- and inter-implementation diversity in binary32, which is echoed in the other sizes as well:

<details><summary>binary64</summary>

```console
$ eshost -se '
  [NaN, Infinity - Infinity, 0 / 0, Math.acos(2), Math.log(-1), Math.sqrt(-1)]
  .map(nan => (new Uint8Array(new Float64Array([nan]).buffer)).join(", "))
  .join("\n")
'
#### engine262, LibJS, Moddable XS
0, 0, 0, 0, 0, 0, 248, 127
0, 0, 0, 0, 0, 0, 248, 127
0, 0, 0, 0, 0, 0, 248, 127
0, 0, 0, 0, 0, 0, 248, 127
0, 0, 0, 0, 0, 0, 248, 127
0, 0, 0, 0, 0, 0, 248, 127

#### GraalJS
0, 0, 0, 0, 0, 0, 248, 127
0, 0, 0, 0, 0, 0, 248, 255
0, 0, 0, 0, 0, 0, 248, 255
0, 0, 0, 0, 0, 0, 248, 255
0, 0, 0, 0, 0, 0, 248, 255
0, 0, 0, 0, 0, 0, 248, 255

#### Hermes, QuickJS
0, 0, 0, 0, 0, 0, 248, 127
0, 0, 0, 0, 0, 0, 248, 255
0, 0, 0, 0, 0, 0, 248, 255
0, 0, 0, 0, 0, 0, 248, 127
0, 0, 0, 0, 0, 0, 248, 127
0, 0, 0, 0, 0, 0, 248, 255

#### JavaScriptCore
0, 0, 0, 0, 0, 0, 248, 127
0, 0, 0, 0, 0, 0, 248, 255
0, 0, 0, 0, 0, 0, 248, 255
0, 0, 0, 0, 0, 0, 248, 127
0, 0, 0, 0, 0, 0, 248, 255
0, 0, 0, 0, 0, 0, 248, 255

#### SpiderMonkey
0, 0, 0, 0, 0, 0, 248, 127
0, 0, 0, 0, 0, 0, 248, 255
0, 0, 0, 0, 0, 0, 248, 127
0, 0, 0, 0, 0, 0, 248, 255
0, 0, 0, 0, 0, 0, 248, 255
0, 0, 0, 0, 0, 0, 248, 255

#### V8
0, 0, 0, 0, 0, 0, 248, 127
0, 0, 0, 0, 0, 0, 248, 255
0, 0, 0, 0, 0, 0, 248, 127
0, 0, 0, 0, 0, 0, 252, 127
0, 0, 0, 0, 0, 0, 252, 127
0, 0, 0, 0, 0, 0, 248, 255
```

</details>

<details><summary>binary16</summary>

```console
$ eshost -se 'globalThis.Float16Array && 
  [NaN, Infinity - Infinity, 0 / 0, Math.acos(2), Math.log(-1), Math.sqrt(-1)]
  .map(nan => (new Uint8Array(new Float16Array([nan]).buffer)).join(", "))
  .join("\n")
'
#### engine262, GraalJS, Hermes, LibJS
undefined

#### JavaScriptCore
0, 126
0, 254
0, 254
0, 126
0, 254
0, 254

#### Moddable XS
0, 126
0, 126
0, 126
0, 126
0, 126
0, 126

#### QuickJS
1, 124
1, 252
1, 252
1, 124
1, 124
1, 252

#### SpiderMonkey
0, 126
0, 254
0, 126
0, 254
0, 254
0, 254

#### V8
0, 126
0, 254
0, 126
0, 126
0, 126
0, 254
```

</details>

packages/ses/src/tame-nan-sidechannel.js
```diff
@@ -38,7 +42,7 @@ const methods = {
     if (is(value, NaN)) {
```
gibson042:
```suggestion
    // Start by converting byteOffset and value to numbers (in that order).
    // https://tc39.es/ecma262/multipage/structured-data.html#sec-setviewvalue
    byteOffset = Number(byteOffset);
    value = Number(value);
    if (is(value, NaN)) {
```

packages/ses/src/tame-nan-sidechannel.js
```diff
@@ -54,7 +58,7 @@ const methods = {
...
```
gibson042:
```suggestion
    // Start by converting byteOffset and value to numbers (in that order).
    // https://tc39.es/ecma262/multipage/structured-data.html#sec-setviewvalue
    byteOffset = Number(byteOffset);
    value = Number(value);
    if (is(value, NaN)) {
```

packages/ses/src/tame-nan-sidechannel.js
erights: I find @ljharb 's (paraphrase) "Do what all implementations do for ***literal*** `NaN`." compelling, since 
- all implementations (but current QuickJS) seem to agree
- these values are the same as @gibson042 's "follow the WHATWG rational for lowest quiet NaN".

Done.

packages/ses/src/tame-nan-sidechannel.js
erights: I changed my mind to your position anyway https://github.com/endojs/endo/pull/3214#discussion_r3165218870

Done.

packages/ses/src/tame-nan-sidechannel.js
erights: Your suggested test wasn't quite right, but I implemented the test you have in mind and test it.

Done.

PR #3216, opened 2026-04-29
feat(base64): Fall through to native Uint8Array

kriskowal:
## Description

Makes `@endo/base64`'s `encodeBase64` and `decodeBase64` dispatch to the native
TC39 `Uint8Array` base64 intrinsics (proposal-arraybuffer-base64, Stage 4) when
available, falling through to the legacy `globalThis.Base64` XS binding and
finally to the existing pure-JavaScript polyfill.
Also hardens the package's named exports while preserving the pre-lockdown
shim path that `@endo/init/pre.js` relies on.

This branch also adds `CLAUDE.md` with guiding principles distilled from
review feedback during the AI-assisted sessions on this work.

### Security Considerations

- Native intrinsics are captured at module load via `Reflect.apply`, with
  `Reflect.apply` itself destructured at module load, so a tampered
  `Function.prototype.call` cannot redirect the dispatched invocation.
- The native options bag is `Object.freeze`'d (not `harden`'d), since
  `src/decode.js` is on the pre-lockdown shim path.
  Importing `@endo/harden` here would install its fallback `harden` before SES
  `lockdown()` and break it.
- Hardening lives in `index.js`, not in the shim path
  (`shim.js` -> `atob.js` / `btoa.js`).
  `@endo/init/pre.js` continues to install `globalThis.atob` /
  `globalThis.btoa` before `lockdown()` without poisoning it.
- `decodeBase64` re-runs the polyfill on any native error to surface the
  polyfill's diagnostic.
  Native error message text is implementation-defined and does not embed the
  caller-supplied `name` or the failing offset.

### Scaling Considerations

The native intrinsic is materially faster than the pure-JS polyfill on V8 and
orders of magnitude faster on XSnap.
Existing call sites pick up the speedup without any change.

Two changesets ship with this PR:

- `@endo/base64: minor` — exports are now frozen.
  Read-only consumers are unaffected; consumers that previously assigned to or
  extended `encodeBase64`, `decodeBase64`, `atob`, or `btoa` will see a
  `TypeError` under SES.
- `@endo/base64: patch` — native dispatch with no API change.
  `decodeBase64` error messages (including any caller-supplied `name` and the
  failing offset) are unchanged.

### Testing Considerations

- `forced-polyfill.test.js` exercises the polyfill directly (via the
  package-internal `./src/encode.js` / `./src/decode.js` paths) and
  cross-checks dispatched output against polyfill output, so CI continues to
  cover the polyfill on runtimes that ship the native intrinsic.
- Existing `@endo/base64` tests pass unchanged.
- `@endo/bundle-source` sanity tests pass under both
  `stackFiltering: 'concise'` and `stackFiltering: 'verbose'`.

- Public API unchanged.
- Polyfill error-message text contract preserved on the native path through
  the polyfill-rerun-on-throw fallback.
- `@endo/base64` adds a runtime dependency on `@endo/harden`.
  The shim path (`@endo/base64/shim.js`) remains free of `@endo/harden` so it
  is still safe to load pre-lockdown.

### Upgrade Considerations

No production upgrade work is required for typical consumers.
Consumers that mutate `@endo/base64`'s exports (an unusual pattern) must stop
or clone first; the failure mode is a clear `TypeError` under SES, not a
silent change.

### DIFF
--- a/packages/base64/index.js
+++ b/packages/base64/index.js
@@ -1,3 +1,13 @@
+// Re-exports the package's named bindings.  Each source module
+// (`./src/encode.js`, `./src/decode.js`, `./atob.js`, `./btoa.js`)
+// applies `Object.freeze` to its export at module-evaluation time, so
+// the bindings are hardened on both the public path through this
+// module and on the pre-lockdown shim path that `@endo/init/pre.js`
+// uses (`@endo/base64/shim.js` -> `./atob.js` / `./btoa.js`).  Using
+// `Object.freeze` rather than `@endo/harden` keeps the shim path free
+// of any module that would install a fallback `harden` before SES
+// `lockdown()` freezes the well-known properties of `globalThis`.
+
 export { encodeBase64 } from './src/encode.js';
 export { decodeBase64 } from './src/decode.js';
 export { btoa } from './btoa.js';
--- a/packages/base64/src/decode.js
+++ b/packages/base64/src/decode.js
@@ -4,17 +4,30 @@
 
 import { monodu64, padding } from './common.js';
 
+// Capture `Reflect.apply` once at module load, before any consumer
+// can tamper with `Function.prototype.call`.
+// Using `Reflect.apply` for the native dispatch ensures a tampered
+// `Function.prototype.call` cannot redirect the dispatched native
+// intrinsic invocation.
+const { apply } = Reflect;
+
 /**
- * Decodes a Base64 string into bytes, as specified in
- * https://tools.ietf.org/html/rfc4648#section-4
+ * Pure-JavaScript base64 decoder, exported for benchmarking and for
+ * environments where the native TC39 `Uint8Array.fromBase64` intrinsic
+ * (proposal-arraybuffer-base64) is unavailable or has been removed.
+ * See `decodeBase64` below for the dispatched default.
  *
- * XSnap is a JavaScript engine based on Moddable/XS.
- * The algorithm below is orders of magnitude too slow on this VM, but it
- * arranges a native binding on the global object.
- * We use that if it is available instead.
+ * XSnap, a JavaScript engine based on Moddable/XS, runs this polyfill
+ * orders of magnitude slower than V8.
+ * Older XS builds expose a native `globalThis.Base64` binding that the
+ * dispatched `decodeBase64` uses ahead of the polyfill on legacy
+ * Agoric chain XS; going forward, XS is expected to favor the standard
+ * `Uint8Array.fromBase64`.
  *
  * This function is exported from this *file* for use in benchmarking,
  * but is not part of the *module*'s public API.
+ *
+ * @see https://datatracker.ietf.org/doc/html/rfc4648#section-4
  *
  * @param {string} string Base64-encoded string
  * @param {string} [name] The name of the string as it will appear in error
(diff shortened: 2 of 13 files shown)


packages/base64/src/decode.js
```diff
@@ -65,20 +78,93 @@ export const jsDecodeBase64 = (string, name = '<unknown>') => {
...
+// load, before any caller can reach `decodeBase64` and before SES
+// lockdown freezes `Uint8Array`.
+// Post-lockdown mutation cannot redirect the dispatched binding.
+const fromBase64 = /** @type {any} */ (Uint8Array).fromBase64;
+const nativeFromBase64 =
+  typeof fromBase64 === 'function'
+    ? /** @type {(input: string, options?: object) => Uint8Array} */ (
+        fromBase64
+      )
+    : undefined;
```
gibson042: This aligns with analogous code in package "hex", and I wonder if both shouldn't be simplified.
```suggestion
/** @type {typeof Uint8Array.fromBase64 | undefined} */
const nativeFromBase64 = Uint8Array.fromBase64;
```

packages/base64/src/decode.js
```diff
@@ -65,20 +78,93 @@ export const jsDecodeBase64 = (string, name = '<unknown>') => {
...
+  alphabet: 'base64',
+});
+
+/** @type {typeof jsDecodeBase64} */
+const nativeDecodeBase64 = (string, name) => {
+  try {
+    return apply(/** @type {any} */ (nativeFromBase64), Uint8Array, [
+      string,
+      nativeFromBase64Options,
+    ]);
```
gibson042:
```suggestion
    return apply(
      /** @type {typeof Uint8Array.fromBase64} */ (nativeFromBase64),
      Uint8Array,
      [string, nativeFromBase64Options],
    );
```

packages/base64/src/decode.js
```diff
@@ -65,20 +78,93 @@ export const jsDecodeBase64 = (string, name = '<unknown>') => {
...
     }
     return new Uint8Array(decoded);
   };
 
-/** @type {typeof jsDecodeBase64} */
-export const decodeBase64 =
+/** @type {typeof jsDecodeBase64 | undefined} */
+const xsDecodeBase64 =
   globalThis.Base64 !== undefined
     ? adaptDecoder(globalThis.Base64.decode)
```
gibson042, line 142: Suggestion: ensure a name for this function 
```suggestion
    ? Object.defineProperty(adaptDecoder(globalThis.Base64.decode), 'name', {
        value: 'xsDecodeBase64',
      })
```

packages/base64/src/decode.js
```diff
@@ -65,20 +78,93 @@ export const jsDecodeBase64 = (string, name = '<unknown>') => {
...
+ * diagnostic that embeds `name` and the failing offset; native
+ * error messages are implementation-defined and do neither.
+ *
+ * @type {typeof jsDecodeBase64}
+ */
+export const decodeBase64 = (() => {
+  if (nativeFromBase64 !== undefined) return nativeDecodeBase64;
+  if (xsDecodeBase64 !== undefined) return xsDecodeBase64;
+  return jsDecodeBase64;
+})();
```
gibson042, line 164: Another question also applicable to package "hex": do we care about the `name` property of exported functions? The export from packages/hex/src/decode.js is sometimes named "" and sometimes named "jsDecodeHex", while this export is sometimes named "nativeDecodeBase64", sometimes named "jsDecodeBase64", and sometimes named "".

packages/base64/src/decode.js
```diff
@@ -65,20 +78,93 @@ export const jsDecodeBase64 = (string, name = '<unknown>') => {
   return data.subarray(0, j);
 };
```
gibson042, line 79:
```suggestion
};
Object.freeze(jsDecodeBase64);
```

packages/base64/src/decode.js
gibson042, line 164:
```suggestion
})();
Object.freeze(decodeBase64);
```

packages/base64/src/encode.js
```diff
@@ -62,11 +76,43 @@ export const jsEncodeBase64 = data => {
   return string;
 };
```
gibson042, line 77:
```suggestion
};
Object.freeze(jsEncodeBase64);
```

packages/base64/src/encode.js
```diff
@@ -62,11 +76,43 @@ export const jsEncodeBase64 = data => {
...
+
+// Legacy XSnap path.
+// Older Moddable/XS builds ship a native `globalThis.Base64.encode`
+// that predates the TC39 intrinsic; we dispatch to it on Agoric chain
+// XS.
+// Going forward, XS will favor the standard
+// `Uint8Array.prototype.toBase64` and this branch will retire.
+/** @type {typeof jsEncodeBase64 | undefined} */
+const xsEncodeBase64 =
+  globalThis.Base64 !== undefined ? globalThis.Base64.encode : undefined;
```
gibson042: Hygiene nit: this should probably wrap `Base64.encode`.
```suggestion
const xsEncodeBase64 = (() => {
  if (globalThis.Base64 === undefined) return undefined;
  const { encode } = globalThis.Base64;
  // eslint-disable-next-line no-shadow
  const xsEncodeBase64 = data => apply(encode, undefined, [data]);
  return xsEncodeBase64;
})();

```

packages/base64/src/encode.js
```diff
@@ -62,11 +76,43 @@ export const jsEncodeBase64 = data => {
...
  *
  * @type {typeof jsEncodeBase64}
  */
-export const encodeBase64 =
-  globalThis.Base64 !== undefined ? globalThis.Base64.encode : jsEncodeBase64;
+export const encodeBase64 = (() => {
+  if (nativeToBase64 !== undefined) return nativeEncodeBase64;
+  if (xsEncodeBase64 !== undefined) return xsEncodeBase64;
+  return jsEncodeBase64;
+})();
```
gibson042, line 125:
```suggestion
})();
Object.freeze(encodeBase64);
```

packages/base64/index.js
```diff
@@ -1,4 +1,28 @@
-export { encodeBase64 } from './src/encode.js';
-export { decodeBase64 } from './src/decode.js';
-export { btoa } from './btoa.js';
-export { atob } from './atob.js';
+// Hardens the package's named exports without disturbing the
+// pre-lockdown shim path: `@endo/init/pre.js` imports
+// `@endo/base64/shim.js` (which loads `./atob.js` and `./btoa.js`,
+// not this module) so that `globalThis.atob` and `globalThis.btoa`
+// can be installed before SES `lockdown()` freezes the global.
```
gibson042: Uh, what? `lockdown()` freezes the global?
```suggestion
// can be installed before SES `lockdown()` freezes the well-known properties of
// globalThis.
```

packages/base64/index.js
```diff
@@ -1,4 +1,28 @@
...
+// Hardening lives here, not in the shim path, so consumers that go
+// through `@endo/base64` get hardened bindings while consumers that
+// only use the shim do not pull `@endo/harden` into their
+// pre-lockdown import chain.
```
gibson042: I think we should consistently harden-via-freeze in the source files (as suggested there).

gibson042 (approved):
All comments about decode.js apply analogously to encode.js.

packages/base64/src/decode.js
kriskowal: I don’t think we care enough to suffer the code we would need to care.

PR #3218, opened 2026-04-30
feat(module-source): fix types, expose analyzer.js subpath

boneskull:
Introduces a low-level analysis primitive `analyzeModule` which is exported from the `analyzer.js` subpath.

- Adds `src/types/external.ts` as a single re-export index for the public types (`SourceMapHook`, `SourceMapHookDetails`, `SourceMapObject`, `ModuleSourceOptions`) and the new analyzer types. `src/external.types.d.ts` now delegates to it.
- Adds `src/types/analyzer.ts` with `AnalysisContext<T>`, `ModuleAnalysisContext`, `AnalysisOptions`, and `VisitorPlugin` — typed by `tsc` rather than inferred from JSDoc.
- Adds `src/shim.types.d.ts` for the global `ModuleSource` augmentation used by the SES shim.
- `tsconfig.json` is tightened: `stripInternal` removes `@internal` declarations from emitted `.d.ts` files.
- Update `typedoc.json` to also consider the additional entry points

- We no longer need to provide custom `@babel/types` exports to these functions. So, `makeModulePlugins` and `makeCjsModulePlugins` no longer return `PluginFactory` functions (`({ types }) => ({ visitor })`). Both now return plain `{ analyzePlugin: { visitor }, transformPlugin: { visitor } }` objects with `@babel/types` imported at module scope. `transform-source.js` is updated accordingly.
- Benchmark and test fixture updated for the new API shape.

- Adds `./analyzer.js` as a new package export. The `analyzeModule(options?)` function creates a fresh per-parse context with `analyzePass` / `transformPass` (plain `Visitor` objects) and `buildRecord(code, location)`. The same function powers the `ModuleSource` constructor internally. This will be consumed by the forthcoming `@endo/parser-pipeline`.


boneskull:

- branch: `master`
  - **[feat(module-source): fix types, expose analyzer.js subpath #3218](https://github.com/endojs/endo/pull/3218)** - branch: `boneskull/module-source-pipeline2` *(this PR)*
    - [feat(module-source): expose AST-based CJS parser #3220](https://github.com/endojs/endo/pull/3220) - branch: `boneskull/cjs-ast-parser2`

---

packages/module-source/src/transform-analyze.js
```diff
@@ -1,133 +1,176 @@
...
-      exportAlls: freeze(sourceOptions.exportAlls),
-      imports: freeze(sourceOptions.imports),
-      liveExportMap: freeze(sourceOptions.liveExportMap),
-      fixedExportMap: freeze(sourceOptions.fixedExportMap),
-      reexportMap: freeze(sourceOptions.reexportMap),
-      needsImport: sourceOptions.dynamicImport.present,
-      needsImportMeta: sourceOptions.importMeta.present,
-      functorSource,
-    });
-    return moduleAnalysis;
```
naugtur, line 120: transformation and analysis used to be executed together to avoid parsing twice. I'm assuming that concern is now gone because the refactor is unifying all parsing to one pass.

If this comment is incorrect, it makes for a review blocker, requesting explanation.

packages/module-source/src/module-source.js
```diff
@@ -33,28 +35,6 @@ const freeze = /** @type {<T>(v: T) => T} */ (Object.freeze);
 
 const analyzeModule = makeModuleAnalyzer();
```
naugtur: this being called the same as the main export from `analyzer.js` is hurting my ability to create an efficient mental model of what's going on here.

Maybe this could help: 

```suggestion
const analyzeModuleSource = makeModuleSourceAnalyzer();
```

packages/module-source/src/analyzer.js
```diff
@@ -0,0 +1,48 @@
...
+import { createSourceOptions } from './source-options.js';
+import { buildFunctorSource, buildModuleRecord } from './functor.js';
+
+/**
+ * Creates a fresh ESM analysis context for a single module parse.
+ *
+ * @param {AnalysisOptions} [options]
+ * @returns {ModuleAnalysisContext}
+ */
+export const analyzeModule = (options = {}) => {
```
naugtur:
```suggestion
export const makeAnalyzerContext = (options = {}) => {
```

packages/module-source/src/module-source.js
```diff
@@ -75,45 +55,18 @@ export function ModuleSource(source, opts = {}) {
...
-    freeze(reexports);
-  }
-  this.__liveExportMap__ = freeze(liveExportMap);
-  this.__reexportMap__ = freeze(reexportMap);
-  this.__fixedExportMap__ = freeze(fixedExportMap);
-  this.__needsImport__ = needsImport;
-  this.__needsImportMeta__ = needsImportMeta;
+  // analyzeModule now returns a frozen PrecompiledModuleSource-shaped record
+  // via buildModuleRecord(), so we copy its properties directly.
+  const record = analyzeModule(source, opts);
```
naugtur, line 60: (nit) `record` is already frozen into oblivion. the copying and freezing seems redundant. Could we pass an optional empty record object to the analyzer so it can create fields there and make it default to an empty object, but in this case pass the `this` reference?

packages/module-source/src/functor.js
```diff
@@ -0,0 +1,126 @@
...
+ * Computes the derived arrays (`imports`, `exports`, `reexports`) from the
+ * raw maps in `sourceOptions` and assembles a {@link ModuleSourceRecord}.
+ *
+ * @param {Record<string, any>} sourceOptions The mutable state bag populated
+ *   by `makeModulePlugins`.
+ * @param {string} functorSource The functor source string from
+ *   {@link buildFunctorSource}.
+ * @returns {ModuleSourceRecord}
+ */
+export const buildModuleRecord = (sourceOptions, functorSource) => {
```
naugtur, line 101: nit followup:
```suggestion
export const buildModuleRecord = (sourceOptions, functorSource, emptyRecord = {}) => {
```

packages/module-source/src/functor.js
```diff
@@ -0,0 +1,126 @@
...
+    freeze(entry);
+  }
+  for (const reexports of values(sourceOptions.reexportMap)) {
+    for (const pair of reexports) {
+      freeze(pair);
+    }
+    freeze(reexports);
+  }
+
+  return freeze({
```
naugtur, line 115: more nit folloup

```suggestion
  return freeze(assign(emptyRecord, {
```

packages/module-source/src/babel-plugin.js
```diff
@@ -1,8 +1,15 @@
 /* eslint max-lines: 0 */
 
+import * as t from '@babel/types';
```
naugtur, line 3: I can only see formatting and type changes in this file.

packages/module-source/src/transform-analyze.js
boneskull: I don't know what this is in reference to

packages/module-source/src/module-source.js
boneskull: I'd like to push back on this just because this convention (unfortunately used _often_ within Endo) makes things much more difficult for me to reason about.

packages/module-source/src/babel-plugin.js
boneskull: yes, because I had to make an equivalent plugin for CJS, I figured I might as well attempt to address _this_ file too.  I only had some success.

packages/module-source/src/module-source.js
naugtur: okay

packages/module-source/src/functor.js
```diff
@@ -0,0 +1,126 @@
...
+      if (cvname) {
+        src = `Object.defineProperty(${cvname},'name',{value:${js(vname)}});`;
+      }
+      const hDeclId = isOnce ? h.HIDDEN_ONCE : h.HIDDEN_LIVE;
+      src += `${hDeclId}.${vname}(${cvname || ''});`;
+      return src;
+    })
+    .join('');
+
+  let functorSource = `\
```
naugtur, line 77: Preserve the original comment
```suggestion
  // The outer function destructures the module calling convention's internal
  // variables into hidden lexical variables.
  // The inner function binds `this` to `undefined` and overshadows the
  // evaluator's `arguments` with a completely empty `arguments` object.
  // There is no avoiding the overshadowing of `globalThis.arguments` if it
  // exists in this emulation of ESM since the evaluator binds `arguments` as
  // well.
  // Relies on the evaluator to ensure these functions are strict.
  let functorSource = `\
```

naugtur (approved):

PR #3220, opened 2026-04-30
feat(module-source): expose AST-based CJS parser

boneskull:
### feat(compartment-mapper): expose Babel-based CJS parser w/ dynamic import support

Expose Babel-based CJS parser, `parse-cjs-babel`. Expose shared functionality for wrapping CJS functors with `__dirname`, `__filename`, etc. Add support for dynamic `import()` (`parse-cjs-babel` only).

### feat(module-source): expose AST-based CJS parser

Exposes AST-based parser for CJS, as well as an `analyzeCjs` function from the `analyzer.js` subpath export.

`CjsModuleSource` is also exported from the main entry point.


boneskull:

- branch: `master`
  - **[feat(module-source): expose AST-based CJS parser #3220](https://github.com/endojs/endo/pull/3220)** - branch: `boneskull/cjs-ast-parser2` *(this PR)*

---

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
```diff
@@ -80,22 +81,64 @@ export const getModulePaths = (readPowers, location) => {
...
+      (importSpecifier.startsWith('./') || importSpecifier.startsWith('../'))
+    ) {
+      try {
+        return resolve(importSpecifier, referrerSpecifier);
+      } catch (_) {
+        // Specifier escapes the package root (e.g. `../../sibling-pkg`).
+        // Fall through so the importNowHook resolves it against packageLocation.
+      }
+    }
+    return importSpecifier;
```
naugtur, line 141: @boneskull  This was already in a diff on a previous item in the PR stack, I think something did not rebase correctly

boneskull: @naugtur I can't find it anywhere else.  It may have erroneously been in another branch, but I moved it here.

packages/compartment-mapper/src/parse-cjs-babel.js
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * Drop-in replacement for {@link parse-cjs.js}. Consumers opt in via the
+ * pre-built parser map:
+ *
+ * ```js
+ * import { parserForLanguageWithCjsBabel } from '@endo/compartment-mapper/import-parsers.js';
+ *
+ * await importLocation(readPowers, entryUrl, {
+ *   parserForLanguage: parserForLanguageWithCjsBabel,
+ * });
```
naugtur, line 13: I think the only way to anticipate what the impact could be is to run my old e2e tests for cjs on both and compare. Booking time to do that tomorrow. 
I'm expecting this to be better than node.js at detecting exports, which should be mostly backwards-compatible.

packages/compartment-mapper/src/import-parsers.js
```diff
@@ -22,3 +25,11 @@ export const defaultParserForLanguage = Object.freeze(
...
   }),
 );
+
+/** @satisfies {Readonly<ParserForLanguage>} */
+export const parserForLanguageWithCjsBabel = freeze(
+  /** @type {const} */ ({
+    ...defaultParserForLanguage,
+    cjs: parserCjsBabel,
+  }),
+);
```
naugtur, line 35: (just a thought) It'd be very convenient, at least in initial testing, but maybe in general, if the defaults would get the new cjs parser based on an env variable.

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
```diff
@@ -80,22 +81,64 @@ export const getModulePaths = (readPowers, location) => {
...
+   * Specifiers that traverse above the package root (e.g. `../../other-pkg`)
+   * are passed through unchanged so the importNowHook can handle them by
+   * resolving against `packageLocation` instead — which is the pre-existing
+   * behavior for cross-package ancestor traversal.
+   *
+   * @param {string} importSpecifier
+   * @returns {string}
+   */
+  const canonicalize = importSpecifier => {
+    if (
```
naugtur, line 130: force all specifiers to string to avoid `require({ toString: ()=>{} })` causing vulnerabilities or just regular mayhem


```suggestion
    importSpecifier = `${importSpecifier}`
    if (
```

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
naugtur, line 187: force all specifiers to string to avoid `require({ toString: ()=>{} })` causing vulnerabilities or just regular mayhem


```suggestion
  const require = importSpecifier => {
    importSpecifier = `${importSpecifier}`
```

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
```diff
@@ -204,6 +247,16 @@ export const wrap = ({
 
   freeze(require);
 
+  /** @param {string} importSpecifier */
+  const importFn = async importSpecifier => {
```
naugtur, line 253: force all specifiers to string to avoid `require({ toString: ()=>{} })` causing vulnerabilities or just regular mayhem


```suggestion
  const importFn = async importSpecifier => {
    importSpecifier = `${importSpecifier}`
```

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
naugtur: yes, I know it's been like this for a while and was not vulnerable, but the complexity is increasing and we should do it everywhere we accept selectors

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/dynamic-import/index.js
```diff
@@ -0,0 +1,3 @@
+module.exports.dynamicImport = async () => {
+  return import('node:fs');
```
naugtur, line 2: Where is calling this tested? I can't find it.

naugtur (commented):
bookmark: reviewed the compartment-mapper part.

packages/compartment-mapper/src/parse-cjs-babel.js
naugtur: update: ran the matrix on the cjs implementation from this branch and it does not differ from default implementation.

https://github.com/endojs/endo-e2e-tests/tree/main/matrix


❌ - missing  
☑ - own property, but falsy  
✔️ - truthy  

|  | **endo** | **endo_nu** | **node-v12** | **node-v12** | **node-v16** | **node-v20** | **node-v24** | **bun** | **esbuild** | **parcel** | **rollup** | **tsc** | **tscInterop** | **webpack** |
|  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |
| [1.cjs](#file-1cjs) even | ✔️ | ✔️ | ❌ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ |
| [1.cjs](#file-1cjs) default | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ❌ | ✔️ | ❌ | ✔️ | ✔️ |
| [1.cjs](#file-1cjs) default.even | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ❌ | ✔️ | ❌ | ✔️ | ✔️ |
| [2.cjs](#file-2cjs) even | ✔️ | ✔️ | ❌ | ❌ | ❌ | ❌ | ❌ | [💥](#bun-error) | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ |
| [2.cjs](#file-2cjs) default | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ❌ | ✔️ | ❌ | ✔️ | ✔️ |
| [2.cjs](#file-2cjs) default.even | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ❌ | ✔️ | ❌ | ✔️ | ✔️ |
| [3.cjs](#file-3cjs) even | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | [💥](#bun-error) | ✔️ | ✔️ | ❌ | ✔️ | ❌ | ✔️ |
| [3.cjs](#file-3cjs) default | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ❌ | ✔️ | ❌ | ✔️ | ✔️ |
| [3.cjs](#file-3cjs) default.even | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ❌ | ✔️ | ❌ | ✔️ | ✔️ |
| [4.cjs](#file-4cjs) even | ✔️ | ✔️ | ❌ | ❌ | ❌ | ❌ | ❌ | [💥](#bun-error) | ✔️ | ✔️ | ❌ | ✔️ | ❌ | ✔️ |
| [4.cjs](#file-4cjs) default | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ❌ | ✔️ | ❌ | ✔️ | ✔️ |
| [4.cjs](#file-4cjs) default.even | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ❌ | ✔️ | ❌ | ✔️ | ✔️ |
| [5.cjs](#file-5cjs) even | ✔️ | ✔️ | ❌ | ❌ | ❌ | ❌ | ❌ | [💥](#bun-error) | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ |
| [5.cjs](#file-5cjs) default | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ❌ | ✔️ | ❌ | ✔️ | ✔️ |
| [5.cjs](#file-5cjs) default.even | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ❌ | ✔️ | ❌ | ✔️ | ✔️ |
| [6.cjs](#file-6cjs) even | ✔️ | ✔️ | ❌ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ❌ |
| [6.cjs](#file-6cjs) default | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ |
| [6.cjs](#file-6cjs) default.even | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ |
| [7.cjs](#file-7cjs) even | ☑ | ☑ | ❌ | ☑ | ☑ | ☑ | ☑ | [💥](#bun-error) | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ |
| [7.cjs](#file-7cjs) default | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ❌ | ❌ | ❌ | ✔️ | ✔️ |
| [7.cjs](#file-7cjs) default.even | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | [💥](#bun-error) | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ |
| [8.cjs](#file-8cjs) even | ✔️ | ✔️ | ❌ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ |
| [8.cjs](#file-8cjs) default | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ |
| [8.cjs](#file-8cjs) default.even | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ☑ | ✔️ | ☑ | ✔️ | ✔️ |
| [9.cjs](#file-9cjs) even | ✔️ | ✔️ | ❌ | ❌ | ❌ | ❌ | ❌ | [💥](#bun-error) | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ |
| [9.cjs](#file-9cjs) default | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ❌ | ✔️ | ❌ | ❌ | ✔️ |
| [9.cjs](#file-9cjs) default.even | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ❌ | ✔️ | ❌ | ❌ | ✔️ |
| [10.cjs](#file-10cjs) even | ✔️ | ✔️ | ❌ | ❌ | ❌ | ❌ | ❌ | [💥](#bun-error) | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ |
| [10.cjs](#file-10cjs) default | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ❌ | ✔️ | ❌ | ❌ | ✔️ |
| [10.cjs](#file-10cjs) default.even | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ❌ | ✔️ | ❌ | ❌ | ✔️ |
| [11.cjs](#file-11cjs) even | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | [💥](#bun-error) | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ |
| [11.cjs](#file-11cjs) default | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ |
| [11.cjs](#file-11cjs) default.even | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | [💥](#bun-error) | ❌ | ✔️ | ✔️ | ✔️ | ✔️ | ❌ |
| [12.cjs](#file-12cjs) even | ✔️ | ✔️ | ❌ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ |
| [12.cjs](#file-12cjs) default | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ❌ | ✔️ | ❌ | ❌ | ✔️ |
| [12.cjs](#file-12cjs) default.even | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ❌ | ✔️ | ❌ | ❌ | ✔️ |
| [13.cjs](#file-13cjs) even | ✔️ | ✔️ | ❌ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ |
| [13.cjs](#file-13cjs) default | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ❌ | ✔️ | ❌ | ✔️ | ✔️ |
| [13.cjs](#file-13cjs) default.even | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | ✔️ | [💥](#bun-error) | ✔️ | ❌ | ✔️ | ❌ | ✔️ | ✔️ |
|  | **endo** | **endo_nu** | **node-v12** | **node-v12** | **node-v16** | **node-v20** | **node-v24** | **bun** | **esbuild** | **parcel** | **rollup** | **tsc** | **tscInterop** | **webpack** |

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/dynamic-import/index.js
boneskull: Here: https://github.com/endojs/endo/blob/959360a2e531c86b86ca2d13a7dc180e90689e5d/packages/compartment-mapper/test/cjs-compat.test.js#L143-L171

`import('node:fs')` triggers the custom exit module import hook in this test, which returns a dummy object (instead of the real `node:fs`).

packages/compartment-mapper/src/parse-cjs-babel.js
naugtur: Ran the entire import/require e2e and it's identical between the two implementations.
I find that surprising, because I expected _some_ difference at least, but it's good signal.

packages/compartment-mapper/src/policy.js
```diff
@@ -260,10 +258,16 @@ export const makeDeferredAttenuatorsProvider = (
...
+      const result = await attenuatorsCompartment.import(
         /** @type {string} */ (attenuatorSpecifier),
       );
-      return namespace;
+      // A compartment constructed with `__noNamespaceBox__: true` returns the
+      // bare namespace from `import` rather than a boxed `{ namespace }`.
+      return /** @type {Attenuator} */ (
+        // eslint-disable-next-line no-underscore-dangle
+        attenuatorsCompartment.__noNamespaceBox__ ? result : result.namespace
+      );
```
boneskull, line 271: @naugtur This and a few other changes are fixes for the assumptions that `compartment-mapper` was making about the shape of the resolved value from `compartment.import()`; `compartment-mapper` had assumed here (and in a couple other places) that `__noNamespaceBox__` was always `false`.  

These fixes are tangential to this PR (as well as the changes to `ses` to expose the getter), and I could conceivably extract them into their own PR.

I've updated some tests (and the `scaffold` harness) to exercise the flag and prove that the fixes worked.

packages/compartment-mapper/test/scaffold.js
```diff
@@ -52,16 +52,17 @@ const getLogger = t => {
 };
 
 /**
+ * @param {typeof Compartment} [CompartmentCtor]
  * @returns {{getCompartments: () => Array<Compartment>, Compartment: typeof Compartment}}
  */
-const compartmentInstrumentationFactory = () => {
+const compartmentInstrumentationFactory = (CompartmentCtor = Compartment) => {
```
boneskull, line 58: @naugtur This now uses the same `Compartment` as provided by the `scaffold` caller to create `InstrumentedCompartment`.

packages/compartment-mapper/test/test.types.d.ts
```diff
@@ -189,7 +191,9 @@ export interface ScaffoldOptions extends LoadLocationOptions {
   knownArchiveFailure?: boolean;
   onError?: ScaffoldOnErrorFn;
   addGlobals?: object;
-  additionalOptions?: object;
+  additionalOptions?: LoadLocationOptions &
+    MapNodeModulesOptions &
+    ArchiveOptions;
```
boneskull, line 196: I'm not sure how accurate this is, but it's close enough.

packages/compartment-mapper/src/parse-cjs-shared-export-wrapper.js
```diff
@@ -204,6 +249,19 @@ export const wrap = ({
...
+    importSpecifier = `${importSpecifier}`;
+    const specifier = has(resolvedImports, importSpecifier)
+      ? resolvedImports[importSpecifier]
+      : canonicalize(importSpecifier);
+    const ns = await compartment.import(specifier);
+
+    // eslint-disable-next-line no-underscore-dangle
+    return compartment.__noNamespaceBox__ ? ns : ns.namespace;
```
boneskull, line 261: namespace box fix for `import()`

.changeset/legal-crabs-lay.md
boneskull, line 1: this is funny and I am twelve

packages/module-source/src/cjs-babel-plugin.js
```diff
@@ -0,0 +1,781 @@
...
+              isRequireCall(part.arguments[0])
+            ) {
+              const specifier = getStringCallArg(part.arguments[0]);
+              if (specifier !== null) {
+                requires.push(specifier);
+                reexports.add(specifier);
+              }
+            }
+          }
+        }
```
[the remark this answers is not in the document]
boneskull: This is invalid due to babel's traversal order

packages/compartment-mapper/src/policy.js
naugtur: ok to leave the changes here.

packages/module-source/src/cjs-babel-plugin.js
```diff
@@ -0,0 +1,788 @@
...
+  if (node.type === 'StringLiteral') return node.value;
+  return null;
+};
+
+/**
+ * @param {CallExpression} node
+ * @returns {string | null}
+ */
+const getStringCallArg = node => {
+  if (node.arguments.length !== 1) return null;
```
naugtur, line 98: (must) This is used to extract argument from dynamic import. 
`import` can have a second argument and this function will return null in such a case. 
```js
await import("./foo.json", { with: { type: "json" } })
```

packages/module-source/src/cjs-babel-plugin.js
```diff
@@ -0,0 +1,788 @@
...
+          const specifier = getStringCallArg(node);
+          if (specifier !== null) {
+            importsArr.push(specifier);
+          }
+          return;
+        }
+
+        if (
+          node.callee.type === 'Identifier' &&
+          node.callee.name === '_interopRequireWildcard' &&
```
naugtur: (must) document what this is for and why knowledge about it is here with a comment block. 

`_interopRequireWildcard` is documented as a helper to transpile `import *` but is it not delegating to `require` anyway? Why do we need to understand it?

naugtur: There are other conventions that differ per bundler and babel is losing market share. Why do we cater to babel specifics here? Or am I misunderstanding something?
Would this be treated as a dynamic require otherwise? What heuristic are we going to use to decide which bundler/transpiler idiosincracies to include here?

packages/module-source/src/cjs-babel-plugin.js
```diff
@@ -0,0 +1,788 @@
...
+          if (specifier !== null) {
+            requires.push(specifier);
+          }
+          return;
+        }
+
+        const calleeName = getCalleeName(node);
+
+        if (
+          (calleeName === '__export' || calleeName === '__exportStar') &&
```
naugtur, line 833: This in turn is typescript specific, which I can appreciate as informative, but I wonder if we're not overstepping. The reexports are going to be available on default (or via require) just fine, and all we won't get is named exports from a typescript-transpiled-to-cjs when imported into esm. 

This is a compatibility improvement for sure, but aren't we letting our users create code that won't work anywhere else? 
Do we now have to track output across versions of typescript so that if a package updates their typescript version in a patch it doesn't turn into a breaking change when used in Endo?

packages/module-source/src/cjs-babel-plugin.js
```diff
@@ -0,0 +1,788 @@
...
+    imports: importsArr,
+    unsafeGetters,
+    dynamicImport,
+    starExportMap,
+  } = options;
+
+  const analyzePlugin = {
+    /** @type {Visitor} */
+    visitor: {
+      CallExpression(path) {
```
naugtur, line 802: (should) It'd be perfect to have a block comment (or a few) in every visitor implementation with examples of what it's looking for.

packages/module-source/test/cjs-compat.test.js
```diff
@@ -0,0 +1,952 @@
...
+        get: function () {
+          return dynamic();
+        }
+      });
+    }
+  `);
+
+  // Both defineProperty calls for 'a' are visited. The second one has an
+  // unsafe getter, so 'a' ends up in unsafeGetters and gets filtered out.
+  t.false(exports.includes('a'));
```
naugtur, line 116: wait, but why?
Oh, so it's not going to be treated as a named export, but will be available as a field lookup on `default` or the object returned by `require`, right?

packages/module-source/src/cjs-babel-plugin.js
naugtur: Additionally, should we make sure that `__export` exists in the AST and is the function we're expecting if we're doing this?

packages/module-source/test/cjs-compat.test.js
```diff
@@ -0,0 +1,952 @@
...
+  `);
+  t.true(exports.includes('__esModule'));
+  t.true(reexports.includes('external'));
+  t.true(reexports.includes('external2'));
+  t.true(reexports.includes('external001'));
+  t.true(reexports.includes('external003'));
+  t.true(reexports.includes('external002'));
+  t.true(reexports.includes('external004'));
+  t.true(reexports.includes('external3'));
+  t.true(reexports.includes('external4'));
```
naugtur, line 431: Wait, how do we know these are reexports, if all we see is `Object.defineProperty(exports, key,` ?

This has a likelyhood of false positives and false negatives, or am I missing something?

What's the motivation for catering to Babel this much while Node.js is entirely incapable of that?

packages/module-source/src/cjs-babel-plugin.js
```diff
@@ -0,0 +1,788 @@
...
+};
+
+/**
+ * Detects the Babel-compiled star-reexport pattern.
+ *
+ * @param {CallExpression} node
+ * @param {Record<string, string>} starExportMap
+ * @returns {string | null}
+ */
+const matchBabelReexportPattern = (node, starExportMap) => {
```
naugtur: Seems expensive (in runtime perf and maintenance cost) for a feature that no other runtime has or will have.

packages/module-source/src/cjs-functor.js
```diff
@@ -0,0 +1,80 @@
...
+  sourceUrl,
+) => {
+  const needsImport = sourceOptions.dynamicImport.present;
+  const params = needsImport
+    ? `require, exports, module, __filename, __dirname, ${h.HIDDEN_IMPORT}`
+    : 'require, exports, module, __filename, __dirname';
+
+  let functorSource = `(function (${params}) { 'use strict'; ${scriptSource} //*/\n})\n`;
+  if (sourceUrl) {
+    functorSource += `//# sourceURL=${sourceUrl}\n`;
```
naugtur, line 48: This is new. is sourceUrl remaining unchanged from original source fole or are we processing it?

What are the odds that someone could make our processing remove or alter a fragment of the sourceUrl in a way that introduces one of the 4 possible newline characters so that the remainder of the string becomes code again and can introduce an excess of closing curlies to escape with blocks?

packages/module-source/src/cjs-babel-plugin.js
```diff
@@ -0,0 +1,788 @@
...
+            return;
+          }
+
+          if (right.type === 'ObjectExpression') {
+            // Reassigning module.exports discards any previously-detected
+            // re-exports (e.g. an earlier `module.exports = require('x')`),
+            // matching the character-level lexer and Node's runtime behavior.
+            // This visitor also fires on the nested `module.exports = { ... }`
+            // inside an esbuild `0 && (...)` hint, so that case is covered too.
+            reexports.clear();
```
naugtur, line 928: One of the popular bundlers (rolldown AFAIR) produces a fake unreachable segment to expose named exports to the lexer. It's likely going to break some assumptions here

example:
https://npmx.dev/package-code/@tanstack/query-core/v/5.101.2/build%2Flegacy%2Findex.cjs

```
// Annotate the CommonJS export names for ESM import in node:
0 && (module.exports = {
  CancelledError,
  InfiniteQueryObserver,
  Mutation,
  MutationCache,
```

We've already had issues with that because the names are unused identifiers (because they're unreachable anyway) and LAvaMoat policy generation recognized them as globals.

naugtur: I pushed a little test case covering this and luckily it catches the lexer-bait and produces the same result. 
I do wonder if more complexity in the test case would break something still.

packages/module-source/src/cjs-babel-plugin.js
naugtur: CC @kriskowal for the ROI and risk conversations

packages/module-source/src/cjs-babel-plugin.js
boneskull: I'm confused; this code is meant to consider this particular use-case.

packages/compartment-mapper/src/import-parsers.js
kriskowal: I’m in favor of using env-options for a feature flag.

packages/compartment-mapper/src/parse-cjs-babel.js
kriskowal: We certainly could tease out some adversarial examples of imports and exports that the lexer would miss but the parser would not.

.changeset/famous-seals-pick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@endo/compartment-mapper': minor
+---
+
+Expose Babel-based CJS parser, `parse-cjs-babel`. Expose shared functionality for wrapping CJS functors with `__dirname`, `__filename`, etc. Add support for dynamic `import()` (`parse-cjs-babel` only).
```
kriskowal: I’m happy to test this in prod with a flag or config like this. Should we plan to promote the new behavior to default in a future release, provided this shakes out? That would entail deprecating the -babel.js entrypoint eventually, leaving a tiny scar.

.changeset/legal-crabs-lay.md
```diff
@@ -0,0 +1,5 @@
+---
+'ses': minor
+---
+
+Exposes `__noNamespaceBox__` getter on `Compartment`.
```
kriskowal, line 5: …that reveals how the compartment was configured originally. `__noNamespaceBox__` and `__options__` flags both imply the new, default behavior. Absence of the property implies `false`.

kriskowal (approved):
This looks ready to me.

.changeset/famous-seals-pick.md
boneskull: I feel like it should be opt-in, given standalone usage without `@endo/parser-pipeline` (#3158) is theoretically less performant.

boneskull:
@kriskowal 

> I’m in favor of using env-options for a feature flag.

Is this something you'd want to use (LavaMoat wouldn't use it)?

packages/module-source/src/cjs-babel-plugin.js
naugtur: yup. Bad comment placement. I pushed a test to cover this. It does consider the lexer trick correctly. 
I was more worried about the interaction with all the ways it detects reexports. 

There's a bunch of stuff that could break here though.
1. we're prone to producing named exports that don't exist
2. reexports can be cleared for no good reason. (which may not be that big of a deal, because Node.js doesn't consider reexports at all. That being said, the experience might be flaky - a refactor that moves function declarations around would be enough to cause changes in what our parser sees)

Both assertions below fail. Overwriting `module.exports` being detected in a hoisted function clears the reexports but all of the named exports remain. The exports assertion describes what Node would do. I'm not saying we must do the same. 

```js
test('overwrite clearing reexports backfires', t => {
  const { exports, reexports } = analyze(`
    prepopulate();
    const that = 'thing';
    module.exports = {
      ...require('./types.cjs'),
      that
    }

    function prepopulate() {
      module.exports = {
        notMuch:1
      }
    }
  `);
  t.deepEqual(exports, ['that', 'default']);
  t.deepEqual([...reexports], ['./types.cjs']);
});
```

naugtur:
> @kriskowal
> 
> > I’m in favor of using env-options for a feature flag.
> 
> Is this something you'd want to use (LavaMoat wouldn't use it)?

It'd be convenient in e2e testing. I don't see applications beyond that off the top of my head.

packages/module-source/test/cjs-compat.test.js
boneskull: The lexer and parser output is identical for this example. See the newly-added `cjs-parity.test.js` and `_corpus.js`

packages/module-source/src/cjs-babel-plugin.js
boneskull: The parser now outputs the same `exports` and `reexports` as the lexer for this example.

packages/module-source/test/cjs-compat.test.js
```diff
@@ -632,20 +629,20 @@ test('Literal exports example', t => {
...
-      // This WILL be detected as an export
+      // This WILL be detected as an export; parsing stops here
       e: require('d'),
 
-      // DIVERGENCE: The AST parser correctly sees all object keys, unlike the
-      // lexer which stops at the first non-identifier value expression.
+      // 'f' is NOT detected: the character-level lexer stops parsing the
+      // object literal after the first property whose value is a non-identifier
+      // expression (here require('d')).  collectObjectExports now replicates
+      // this stop-at-first-hard-property semantics.
```
naugtur, line 639: (should) That might be a step too far in compatibility with the lexer. We're adding complexity to the parser for a worse result. And the better output from the parser would be less flaky than the lexer situation in this case  rearranging exports is a breaking change under lexer but doesn't have to be under parser. I would prefer not bailing out here.

packages/module-source/test/cjs-parity.test.js
```diff
@@ -0,0 +1,45 @@
...
+import { analyzeCommonJS } from '@endo/cjs-module-analyzer';
+import { CjsModuleSource } from '../src/cjs-module-source.js';
+import { corpus } from './_corpus.js';
+
+for (const { name, source, failing } of corpus) {
+  const title = `parity: ${name}${failing ? ` (reason: ${failing})` : ''}`;
+
+  const testFn = failing ? test.failing : test;
+
+  testFn(title, t => {
```
naugtur, line 25: re https://github.com/endojs/endo/pull/3220/changes#r3594861726
For the sake of parity testing we could add a list of additional exports that parser should detect to the _corpus entries

packages/compartment-mapper/test/fixtures-cjs-compat/node_modules/nonidentifier-exports/index.js
```diff
@@ -0,0 +1,25 @@
+// Fixture for lexer/parser divergence on non-identifier export names.
+//
+// The character-level lexer (`@endo/cjs-module-analyzer`, the `default` CJS
+// parser) emits export names verbatim, including strings that are not valid
+// JavaScript identifiers. The Babel analyzer (`CjsModuleSource`, the `babel`
+// CJS parser) filters these via `isValidExportName`, because they cannot be
+// used as static ESM export bindings.
+//
+// This module is fully executable: every value is defined, so the divergence
+// is purely in *analysis* (which export names are detected), not execution.
```
naugtur, line 10: good stuff

naugtur (commented):
I'm very close to approving this. I have one new concern about the latest changes and two marked as (must) that we need to discuss first.

packages/module-source/test/cjs-compat.test.js
boneskull: I'd prefer to leave it as-is until we get hit by a situation in which we should have done this. 😜

boneskull:
## TODO

I need to address these:

- [x] https://github.com/endojs/endo/pull/3220#discussion_r3490739763
- [x] https://github.com/endojs/endo/pull/3220#discussion_r3490999161
- [x] https://github.com/endojs/endo/pull/3220#pullrequestreview-4590728629
- [x] https://github.com/endojs/endo/pull/3220#pullrequestreview-4713059391
- [x] Do not inhibit exports/reexports named using invalid identifiers, as it seems there's no good reason to do this.

packages/module-source/test/_corpus.js
```diff
@@ -610,4 +610,27 @@ export const corpus = [
       exports['b'].b;
     `,
   },
+
+  {
+    // A property whose value is not an identifier token (`notMuch: 1`) must not
+    // be recorded, and it halts the literal parse: the lexer bails on the first
+    // "hard" value, so `notMuch` is never emitted and neither is `theOther`.
```
naugtur, line 620: For the same reasons I pushed back against very specific ways of detecting exports that depend on certain known output shapes from bundlers, I think bailing out on "hard" value is not beneficial. The reasons boil down to stability of named exports list across non-breaking versions of a package. 
 
- If a package updates its bundler without changing any of its logic, the complete list of its named exports should preferably stay the same
- If  package updates its linter and it sorts the fields in exports, the complete list of its named exports should preferably stay the same

The latter is why I think not bailing out like lexer does is preferable. The lexer bails out only because the value could be a string and not being a parser it has no way of continuing. We can continue and therefore provide a more stable end result.

boneskull: My rationale for leaving it as-is boils down to this:

- Any Node.js user who would be impacted by such a change will find that their app is broken when running under Node.js directly.  By diverging, Endo/LavaMoat could _hide breaking changes_ from the consumer, potentially impacting any _downstream_ consumers.
- Maintaining exceptions to the behavior is a cognitive burden; "parity" is easy to grasp.
- From what I can determine, this is something of a corner case in the wild. Recall that this behavior was _unknown to me_ before you told me about it, and I've been engaged w/ Node.js core & the ecosystem for years.

That being said, if one of our downstream users gets bitten by this and wants different behavior, I'm happy to entertain it.

packages/module-source/src/cjs-babel-plugin.js
boneskull: I've determined this special-case was unnecessary (and actually introduced duplicate exported names!) after subsequent fixes, so I've removed it.

packages/module-source/src/cjs-functor.js
boneskull: This behavior is cargo-culted from the ESM functor in `functor.js`.  `sourceUrl` doesn't seem to be a user-provided string at all, and is rather computed from the `CompartmentDescriptor`, as far as I can tell from what's happening in `compartment-mapper`.  Regardless, I've added a helper to ensure the value is a valid URL.

packages/module-source/test/cjs-parity.test.js
boneskull: Don't think this is currently needed, as we already have coverage of the specific divergences in `cjs-compat.test.js`.  The parity test is strictly scoped to comparison of exact output.

To recap, there are two:

1. The lexer allows invalid syntax in at least one case; we cannot parse this.
2. The parser allows non-identifier names for named exports.

packages/module-source/test/_corpus.js
naugtur: I accept leaving it as-is. Don't think we'll ever get a report of this since the same faulty behavior happens in Node. I was hoping to move the state of the art forward by showing it can be done. (especially since people don't mind the parsing needed to strip types, performance may no longer be an argument for using a lexer instead of a full parse)

Anyway, ok to keep as-is.

naugtur (approved):
LGTM. 

Maybe enough changed to warrant a rerun of endo-e2e-tests on this?

PR #3221, opened 2026-04-30
feat(compartment-mapper): expose findUnknownCanonicalNames

boneskull:
Expose `findUnknownCanonicalNames()` from new export `@endo/compartment-mapper/policy.js`.

Should be used _after_ `captureFromMap` when called without policy, since unknown canonical names are not emitted by the `UnknownCanonicalNameHook` unless a policy is present. Practically, `@lavamoat/node` can provide it with canonical names from its own generated policy or policy overrides.

* * * 

**UPDATE** Jul 16 2026

Alternatively, we could thread an option through so that `mapNodeModules()` can receive an arbitrary list of canonical names which could potentially be reported by `UnknownCanonicalNameHook`.  

> [!NOTE]
> In the case where `@lavamoat/node` already has a policy (execution), we do not need `findUnknownCanonicalNames()` because `UnknownCanonicalNameHook` reports them as expected.  At the time of calling `mapNodeModules()` during our policy generation, we may have policy overrides (and implicitly a list of canonical names).

Another idea would be to just keep `findUnknownCanonicalNames()` out of this package entirely (since it doesn't absolutely need to live here!) and move the function into `@lavamoat/node`, since it's _yet another random function_ exported from `@endo/compartment-mapper`.

### DIFF
--- a/packages/compartment-mapper/src/node-modules.js
+++ b/packages/compartment-mapper/src/node-modules.js
@@ -23,7 +23,11 @@ import {
   ENTRY_COMPARTMENT,
   generateCanonicalName,
 } from './policy-format.js';
-import { dependencyAllowedByPolicy, makePackagePolicy } from './policy.js';
+import {
+  dependencyAllowedByPolicy,
+  makePackagePolicy,
+  findUnknownCanonicalNames,
+} from './policy.js';
 import { unpackReadPowers } from './powers.js';
 import { search, searchDescriptor } from './search.js';
 import { GenericGraph, makeShortestPath } from './generic-graph.js';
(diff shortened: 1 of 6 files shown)


boneskull:

- branch: `master`
  - **[feat(compartment-mapper): expose findUnknownCanonicalNames #3221](https://github.com/endojs/endo/pull/3221)** - branch: `boneskull/export-find-canonical-names2` *(this PR)*

---

naugtur:
Conceptually, would it be valid to use `findUnknownCanonicalNames` instead of `UnknownCanonicalNameHook` for the case where policy is set and the hook would normally handle that?

packages/compartment-mapper/src/node-modules.js
```diff
@@ -1269,16 +1273,16 @@ const finalizeGraph = (
...
  * @param {Set<CanonicalName>} canonicalNames Set of all known canonical names
  * @param {SomePolicy} policy Policy to validate
+ * @param {Set<CanonicalName>} unknownSet Set of unknown canonical names from {@link findUnknownCanonicalNames}
  * @returns {Array<{canonicalName: CanonicalName, message: string, path:
- * string[], suggestion?: CanonicalName}>} Array of issue objects, or `undefined` if no issues were
- * found
+ * string[], suggestion?: CanonicalName}>} Array of issue objects
  */
-const validatePolicyResources = (canonicalNames, policy) => {
+const validatePolicyResources = (canonicalNames, policy, unknownSet) => {
```
naugtur: (could) Passing a known list of unknown things to complain about in validation is a bit awkward. 
Could `validatePolicyResources` call `findUnknownCanonicalNames`? It already gets all needed data as arguments.

packages/compartment-mapper/src/node-modules.js
```diff
@@ -1300,7 +1304,7 @@ const validatePolicyResources = (canonicalNames, policy) => {
   for (const [resourceName, resourcePolicy] of entries(
     policy.resources ?? {},
   )) {
-    if (!canonicalNames.has(resourceName)) {
+    if (unknownSet.has(resourceName)) {
```
naugtur: (should) I don't see how using `findUnknownCanonicalNames` here is an improvement - it does the same pass of iterating over all resources and their package policies a second time before we go and iterate here again. 

`findUnknownCanonicalNames` is just not coming up with the nice error messages. 

I'd expect `findUnknownCanonicalNames` to produce enough output for `validatePolicyResources` to only need to iterate the results of `findUnknownCanonicalNames`, call `findSuggestion` and add to issues.

boneskull:
@naugtur 

> Conceptually, would it be valid to use findUnknownCanonicalNames instead of UnknownCanonicalNameHook for the case where policy is set and the hook would normally handle that?

I just pushed an update, but I think the answer is yes.  Basically, if you have policy before calling `mapNodeModules`, you can use the hook. If you don't have policy, then you can call `findUnknownCanonicalNames` whenever you _do_ have a policy.

packages/compartment-mapper/src/node-modules.js
boneskull: This has been refactored away

packages/compartment-mapper/src/node-modules.js
boneskull: This has also been refactored away

boneskull:
@naugtur `UnknownCanonicalNameHook` now uses `findUnknownCanonicalNames`. The original `findUnknownCanonicalNames` was canned and `findUnknownCanonicalNames` is now just the `validatePolicyResources` fn as used by `UnknownCanonicalNameHook`.  This has downstream implications, of course, since I changed the signature to return more information.

Just to cover all my bases, I also added logic to check `entry.packages` (which doesn't apply to LavaMoat policies, obviously) for unknown canonical names.

packages/compartment-mapper/src/policy.js
```diff
@@ -638,3 +638,105 @@ export const attenuateModuleHook = async (
...
+   * Finds a suggestion for `badName` if it is a suffix of any canonical name in
+   * `canonicalNames`.
+   *
+   * @param {string} badName Unknown canonical name
+   * @returns {CanonicalName | undefined} Canonical name with a suffix of
+   * `>${badName}`, or `undefined` if none found
+   */
+  const findSuggestion = badName => {
+    for (const canonicalName of canonicalNames) {
+      if (canonicalName.endsWith(`>${badName}`)) {
```
naugtur, line 662: Note: this sort of relies on canonicalNames order resulting from their finding being breadth-first, which at first glance you might consider problematic, but if canonical name search was depth-first it'd be multiple orders of magnitude less performant (we know that from experience)

packages/compartment-mapper/src/policy.js
```diff
@@ -638,3 +638,105 @@ export const attenuateModuleHook = async (
...
+        const issueMessage = `Resource ${q(entryPackageName)} from entry policy was not found`;
+        const suggestion = findSuggestion(entryPackageName);
+        const issue = {
+          canonicalName: entryPackageName,
+          message: issueMessage,
+          path: ['entry', 'packages', entryPackageName],
+        };
+        if (suggestion) {
+          issue.suggestion = suggestion;
+        }
```
naugtur, line 688: (nit) decorating issues with suggestions could happen in a separate pass after all 3 loops

naugtur (approved):
much improved.

naugtur:
CI failing on lockfile mismatch, accidentally committed?

PR #3222, opened 2026-04-30
feat(xorshift): Factor out prng used for testing

kriskowal:
## Description

Adds a new `@endo/xorshift` package that hosts the xorshift128+ PRNG previously duplicated under `packages/ocapn/test/_xorshift.js` and `packages/hex/test/_xorshift.js`, then rewires both consumers to import from it and deletes the duplicates.

`makeXorShift(seed)` returns `{ random, int }`. `random` is a plain function, and `int(lo, hi)` is a sibling method on the returned object.

### Security Considerations

None for production. `@endo/xorshift` is documented as **not** cryptographically secure and is only consumed by test/bench files (`devDependencies`). The package itself runs `harden` on both the factory and the returned generator to prevent prototype tampering when imported after lockdown.

### Scaling Considerations

Test-only code; no production hot-path impact. The rejection-sampling loop in `int(lo, hi)` runs an expected `≤ 2` draws per call for any `range ≤ 2^53`, so worst-case fuzz/bench overhead is unchanged from the helpers it replaces.

- New `packages/xorshift/README.md` documents the seed format, returned shape, and the not-cryptographically-secure caveat.
- A changeset bumps `@endo/xorshift` (minor — first publish), `@endo/hex` (patch), and `@endo/ocapn` (patch).
- No downstream user migration: the deleted `_xorshift.js` files were internal test helpers, never exported.

### Testing Considerations

The new package ships its own ses-ava multi-config suite (lockdown / unsafe / shims-only) so portability across SES configurations is covered. The pre-existing hex benches and ocapn fuzz tests continue to pass against the new package, which exercises it under each consumer's actual usage pattern.

No public-API breakage. The two deleted `_xorshift.js` helpers were never exported from their respective packages. The new `makeXorShift` shape (`{ random, int }`, `Uint8Array` seed) is internally consistent across both consumers; the seed format differs from the legacy 4-number array form used by the old helpers, but both call sites are migrated in the same PR.

### Upgrade Considerations

None — no production deployments are affected. Live systems do not import `@endo/xorshift`.

### DIFF
--- a/.changeset/byte-array-hex-codecs.md
+++ b/.changeset/byte-array-hex-codecs.md
@@ -1,0 +1,23 @@
+---
+'@endo/pass-style': minor
+'@endo/marshal': minor
+---
+
+Immutable `ArrayBuffer` (passStyle `'byteArray'`) is now serializable through
+the capdata, smallcaps, and encode-passable codecs.
+
+- `@endo/pass-style` gains `byteArrayToHex`, `hexToByteArray`,
+  `byteArrayToUint8Array`, and `uint8ArrayToByteArray` helpers, all routed
+  through `@endo/hex`.
+- `@endo/marshal`:
+  - **capdata**: byteArray encodes as `{"@qclass":"byteArray","data":"<hex>"}`.
+  - **smallcaps**: byteArray encodes as `"*<hex>"`.  The `*` prefix is now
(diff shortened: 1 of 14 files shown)


kriskowal:
Closing in favor of #3232

PR #3223, opened 2026-04-30
chore(ses): Improve the fidelity of DataView.prototype.setFloat* replacements

gibson042:
Ref #3214

## Description

* Start each DataView.prototype.setFloat* replacement with the RequireInternalSlot required by [SetViewValue](https://tc39.es/ecma262/multipage/structured-data.html#sec-setviewvalue).
* Simplify the manual `toNumber` to just `Math.max(v)`, which is a [minimal wrapper over ToNumber](https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-math.max) (also fixes failure of the old implementation to error on input like `{ valueOf: () => 0n }`, which is not caught by a `typeof v === 'bigint'` check but nevertheless ToPrimitive-coerces to a bigint that still must be rejected with a TypeError).
* Tweak the ToIndex emulation for efficiency (especially when the setFloat* byteOffset is 0) and for a less tailored error message when the byteOffset is invalid (closer to the messages produced by implementations).
* Export the additional built-ins needed by packages/ses/src/tame-nan-sidechannel.js from packages/ses/src/commons.js .

### Security Considerations

Performing RequireInternalSlot as required by the spec means that code will run the same steps in Hardened JS as it would in a primordial environment.

### Scaling Considerations

n/a

No additional documentation is warranted for this refinement.

### Testing Considerations

No tests were added; the right place for this level of coverage is test262.

Performing RequireInternalSlot as required by the spec means that code will run the same steps in Hardened JS as it would in a primordial environment.

### Upgrade Considerations

No additional considerations beyond those of #3214 and #3153 before it.

### DIFF
--- a/packages/ses/src/tame-nan-sidechannel.js
+++ b/packages/ses/src/tame-nan-sidechannel.js
@@ -1,23 +1,22 @@
 import {
-  Object,
-  DataView,
-  Reflect,
-  Number,
-  TypeError,
   RangeError,
-  Math,
+  dataViewPrototype,
+  MAX_SAFE_INTEGER,
+  apply,
+  defineProperty,
+  entries,
+  getOwnPropertyDescriptor,
   hasOwn,
+  is,
+  max,
+  trunc,
+  uncurryThis,
 } from './commons.js';
-
-const { is, defineProperty, entries } = Object;
-const { apply } = Reflect;
-const { prototype: dataViewPrototype } = DataView;
-const { trunc } = Math;
 
 /**
  * These `FERAL*` methods open up the NaN side-channel on some platforms,
  * like v8. Thus, we need to encapsulate them and replace them with wrappers
- * that canonicaize NaNs.
+ * that canonicalize NaNs.
  */
 const {
   setFloat16: FERAL_SET_FLOAT16,
@@ -28,6 +27,11 @@ const {
   setUint32,
   setBigUint64,
 } = dataViewPrototype;
+
+const dataViewGetBuffer = uncurryThis(
+  // @ts-expect-error we know it is there on all conforming platforms
+  getOwnPropertyDescriptor(dataViewPrototype, 'buffer').get,
+);
 
 /**
  * See https://webidl.spec.whatwg.org/#js-unrestricted-double which implies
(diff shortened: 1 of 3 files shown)


packages/ses/src/tame-nan-sidechannel.js
```diff
@@ -79,6 +83,19 @@ const canonicalNaN32Encoding = 0x7fc00000;
...
+/**
+ * Perform `RequireInternalSlot(obj, [[DataView]])` as needed in
+ * SetViewValue.
+ * https://tc39.es/ecma262/multipage/structured-data.html#sec-setviewvalue
+ * https://tc39.es/ecma262/#sec-get-dataview.prototype.buffer
+ *
+ * @param {unknown} obj
+ * @returns {void}
+ */
+const requireDataView = obj => {
```
gibson042, line 95: I guess we could also name this `assertDataView`.

erights (approved):
I understand and agree with all the improvements. LGTM, thanks!!

packages/ses/src/tame-nan-sidechannel.js
```diff
@@ -79,6 +83,19 @@ const canonicalNaN32Encoding = 0x7fc00000;
...
 const canonicalNaN16Encoding = 0x7e00;
 
+/**
+ * Perform `RequireInternalSlot(obj, [[DataView]])` as needed in
+ * SetViewValue.
+ * https://tc39.es/ecma262/multipage/structured-data.html#sec-setviewvalue
+ * https://tc39.es/ecma262/#sec-get-dataview.prototype.buffer
+ *
+ * @param {unknown} obj
+ * @returns {void}
```
mhofman, line 93: I think we could have made this a type assertion

```suggestion
 * @returns {asserts obj is DataView}
```

packages/ses/src/tame-nan-sidechannel.js
```diff
@@ -90,44 +107,41 @@ const toIndex = v => {
   // @ts-expect-error Math.trunc uses the internal `ToNumber` to coerce its
   // argument, whatever it is, to an integer number.
   const n = trunc(v);
-  if (is(n, NaN) || is(n, -0)) {
+  if (n === 0 || is(n, NaN)) {
```
mhofman, line 110: curious, why the flip in order?

PR #3224, opened 2026-04-30
fix(ses): Further harden internal use of regular expressions

gibson042:
## Description

As discovered by @ChALkeR, `_regexpConstructor` was accidentally set to undefined rather than the intended `RegExp.prototype.constructor`.

### Security Considerations

A hardened environment preserves primordial prototype properties, but package "ses" still strives for very defensive code, which this aligns with.

### Scaling Considerations

n/a

Nothing beyond the included JSDoc.

### Testing Considerations

Just the current tests.

None.

### Upgrade Considerations

None.

### DIFF
--- a/packages/ses/src/transforms.js
+++ b/packages/ses/src/transforms.js
@@ -3,9 +3,10 @@
 import {
   FERAL_REG_EXP,
   SyntaxError,
+  freezeRegexp,
   regexpReplace,
+  regexpSearch,
   sealRegexp,
-  stringSearch,
   stringSlice,
   stringSplit,
   freeze,
@@ -21,7 +22,7 @@ import { getSourceURL } from './get-source-url.js';
  * @returns {number}
  */
 function getLineNumber(src, pattern) {
-  const index = stringSearch(src, pattern);
+  const index = regexpSearch(pattern, src);
   if (index < 0) {
     return -1;
   }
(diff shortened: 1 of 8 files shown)


packages/ses/src/transforms.js
```diff
@@ -177,9 +178,8 @@ export const evadeImportExpressionTest = src => {
 
 // /////////////////////////////////////////////////////////////////////////////
 
-const someDirectEvalPattern = new FERAL_REG_EXP(
-  '(^|[^.])\\beval(\\s*\\()',
-  'g',
```
kriskowal, line 182: Will we miss the `g`?

kriskowal (approved):

packages/ses/src/transforms.js
mhofman: Was there a vulnerability here before where the regexp would remember the last position and possibly miss the eval if it was before that position ?

mhofman (commented):
Change looks fine, one comment about previous behavior that was fixed.

mhofman (approved):

packages/ses/src/transforms.js
gibson042: No to both questions. This regular expression was only used with `stringSearch`, which (via [RegExp.prototype [ %Symbol.search% ]](https://tc39.es/ecma262/multipage/text-processing.html#sec-regexp.prototype-%symbol.search%)) doesn't care about flag "g" and always starts from `lastIndex` 0. This refactoring just allows it to be frozen rather than merely sealed, which isn't possible with flag "g" because that causes [RegExpBuiltinExec](https://tc39.es/ecma262/multipage/text-processing.html#sec-regexpbuiltinexec) to set `lastIndex` upon finding a match.

PR #3229, opened 2026-05-01
chore(cli): remove unreachable branch in run command (#3081)

### LINKED ISSUE #3081: Dead code in cli run command?
https://github.com/endojs/endo/blob/b024b06c7b801e8683a8930715558cfbedfd653d/packages/cli/src/commands/run.js#L81

Unless I'm missing something obvious, the then branch of this if can never be taken, because it happens in the else branch of an outer 

https://github.com/endojs/endo/blob/b024b06c7b801e8683a8930715558cfbedfd653d/packages/cli/src/commands/run.js#L59

kriskowal:
Closes #3081 

The inner `if (importPath !== undefined)` block sat inside the `else` branch of an outer `if (importPath !== undefined)`, so it could never be reached. The dead block re-emitted the same "Must specify either --bundle or --UNCONFINED, not both" error already produced by the live arm of the outer conditional, so deleting it is a pure dead-code removal with no observable behavior change.

### DIFF
--- a/packages/cli/src/commands/run.js
+++ b/packages/cli/src/commands/run.js
@@ -79,13 +79,6 @@ export const run = async ({
         /** @type {any} */
         let bundle;
         if (bundleName !== undefined) {
-          if (importPath !== undefined) {
-            console.error(
-              'Must specify either --bundle or --UNCONFINED, not both',
-            );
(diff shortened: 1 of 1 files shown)


erights (approved):
LGTM thanks.

PR #3232, opened 2026-05-02
feat(chacha12): Consolidate PRNG for fuzzing

kriskowal:
## Description

Adds two new workspace packages and rewires the existing
deterministic-fuzz consumers in this repo to use them.

### `@endo/random`

A source-agnostic library of random sampling functions.  Each
sampler accepts a `RandomSource` as its first argument:

```ts
type RandomSource = (out: Uint8Array) => void;
```

The shape mirrors `crypto.getRandomValues` (minus the return value),
so the canonical browser/Node entropy source and a
`@endo/chacha12`-backed source returned by `makeChaCha12(seed)` are
both directly usable wherever a `RandomSource` is expected.

Names follow the TC39
[proposal-random-functions](https://tc39.es/proposal-random-functions/)
translation `Random.method` -> `randomMethod`.  Each sampler is its
own subpath export so consumers can import only what they use:

| Path | Exports |
| --- | --- |
| `@endo/random` | all of the below |
| `@endo/random/random.js` | `random` |
| `@endo/random/int.js` | `randomInt` |
| `@endo/random/uint.js` | `randomUint8`, `randomUint16`, `randomUint24`, `randomUint32`, `randomUint53` |
| `@endo/random/fast-check.js` | `adaptToPureRandomGenerator`, `adaptFromPureRandomGenerator`, `makeRandomTypeFromSeed` |
| `@endo/random/seeds.js` | `bobsCoffee64` |

`@endo/random/fast-check.js` carries bidirectional adapters between
`RandomSource` and the `pure-rand` v5 `RandomGenerator` interface
that `fast-check` consumes via the `randomType` parameter, plus a
`makeRandomTypeFromSeed` builder.  It imports nothing from
`pure-rand` or `fast-check`; it depends only on those interfaces
being structurally compatible.

`@endo/random/seeds.js` exports `bobsCoffee64`, a 32-byte seed
(`b0b5c0ffeefacade` repeated four times) shared across Endo
deterministic fuzz suites so a single seed change propagates.

### `@endo/chacha12`

A pure-JavaScript ChaCha12 keystream.  The package's sole public
entry point is `makeChaCha12(key)`, which returns a function
`(out: Uint8Array) => void` that fills `out` with the next bytes of
the keystream.  That shape matches `crypto.getRandomValues` and
`@endo/random`'s `RandomSource` exactly, so the same function serves
either ecosystem and the two packages compose without an adapter.

ChaCha12 is the 12-round variant of Daniel J. Bernstein's ChaCha
family.  The block function is identical to ChaCha20 modulo the
loop count (6 double-rounds vs 10).  The reduced round count trades
cryptographic safety margin for speed.  The package is positioned
and documented as a PRNG, not a cipher recommendation.

### Consumer rewires

The `@endo/hex` encode/decode benchmarks and the `@endo/ocapn`
syrup and passable fuzz tests now draw deterministic bytes from
`@endo/chacha12` directly, seeded from `@endo/random/seeds.js`'s
`bobsCoffee64`.  This removes the duplicated `_xorshift.js` helpers
that previously lived in `packages/hex/test/` and
`packages/ocapn/test/`.

Most critical files to review:

- `packages/random/src/int.js` — the range-aware rejection-sampling
  staircase for `randomInt`.  Each tier picks the smallest draw
  width that covers the requested range with reject fraction
  below 0.5, capping the worst-case reject rate across the entire
  safe-integer domain.
- `packages/random/src/read-uint.js` — fixed-width little-endian
  unsigned readers backed by per-width scratch buffers (sharing a
  single 8-byte buffer measured ~2.5x slower under V8's
  element-kind specialization).  Each reader zeroes its scratch
  prefix after the read.
- `packages/random/src/random.js` — `random()` returns
  `randomUint53(source) * 2 ** -53`, deterministic across runs and
  engines.
- `packages/random/fast-check.js` — `pure-rand` v5 `RandomGenerator`
  adapters and the `randomType` builder.  Documents the `clone()`
  alias semantics for serial keystreams without a snapshot facility.
- `packages/random/seeds.js` — canonical `bobsCoffee64` seed.
- `packages/random/random.types.d.ts` — `RandomSource` and
  `PureRandomGenerator` type definitions.
- `packages/chacha12/src/chacha12.js` — the block function, state
  layout, and the IETF / RFC 7539 nonce convention.
- `packages/chacha12/test/chacha12.test.js` — TC1, TC4, TC8
  keystream vectors from
  `draft-strombergson-chacha-test-vectors-01`, with a layout note
  documenting the DJB 8-byte-IV vs IETF 12-byte-nonce
  reconciliation at counter = 0.
- `packages/random/test/random.bench.js` — the cross-source
  microbenchmark; lives in `@endo/random` rather than
  `@endo/chacha12` because it drives both keystreams through the
  same samplers (and because chacha12 cannot devDep on random
  without a cycle — random already devDeps on chacha12).
- `packages/hex/test/{decode,encode}.bench.js`,
  `packages/ocapn/test/{codecs/passable-fuzz,syrup/fuzz}.test.js` —
  consumer-side rewires.

### Security Considerations

`@endo/chacha12` is positioned and documented as a PRNG, not a
cipher.  The README explicitly steers cipher use cases to ChaCha20
or another 20-round implementation; the `cipher` package keyword
has been intentionally omitted.  ChaCha12 reduces the cryptographic
safety margin relative to ChaCha20 (12 rounds vs 20); for
deterministic test fixtures, property-based testing, and fuzzing —
the consumer use cases in this PR — the extra speed is the right
tradeoff, but downstream users should not adopt this package for
confidentiality or for key derivation from caller-supplied seeds.

Neither package introduces new authority surface.  `makeChaCha12`
is a pure deterministic function of its 32-byte key; the returned
function is hardened.  `@endo/random`'s samplers are pure functions
over their `source` argument, with module state limited to small
per-width scratch buffers that are zeroed after each read so no
random bytes linger across calls.  No I/O, no entropy source, no
global state.

The hex/ocapn rewires add a workspace `devDependency` only — no
runtime dependency change for consumers of those packages.

### Scaling Considerations

`@endo/chacha12/BENCH.md` reports the cross-source microbenchmark
(`packages/random/test/random.bench.js`).  Headline numbers on the
test bed (Node 22.22.2, AMD Ryzen AI MAX+, x64, median of 10 runs):

| Workload | xorshift128+ | ChaCha20 | ChaCha12 | C12/C20 |
| --- | ---: | ---: | ---: | ---: |
| Bulk bytes (MB/s) | 696 | 190 | 281 | 1.48x |
| `random()` (ns/call) | 17.2 | 45.4 | 35.1 | 1.29x |
| `randomInt(0, 99)` (ns/call) | 11.1 | 15.6 | 14.3 | 1.10x |

The bulk-bytes ratio is below the naive 20/12 = 1.67x because
per-block fixed costs are identical between ChaCha20 and ChaCha12.
The sampler workloads narrow the gap further because the per-call
envelope (range-aware staircase, mask-and-reject) is shared across
all sources and amortizes the keystream difference.

No new on-chain storage, message exchange, or persistent resource
consumption.

Self-contained docs ship with each new package:

- `packages/random/README.md` — `RandomSource` interface, subpath
  exports, determinism guarantees, scratch-buffer rationale.
- `packages/chacha12/README.md` — `makeChaCha12` API, ChaCha12 vs
  ChaCha20 framing, keystream-length bound, verification.
- `packages/chacha12/BENCH.md` — full benchmark report.
- `packages/random/SECURITY.md`, `packages/chacha12/SECURITY.md` —
  standard Endo security policy.
- `.changeset/endo-chacha12.md` — `@endo/random: minor`,
  `@endo/chacha12: minor`, `@endo/hex: patch`, `@endo/ocapn: patch`.

No existing data or deployment is affected — both packages are new,
and the hex/ocapn changes are confined to test/benchmark helpers.
No upgrade instructions are needed for downstream users of
`@endo/hex` or `@endo/ocapn`.

### Testing Considerations

`packages/random/test/`:

- `random.test.js` — `random(source)` range invariants,
  determinism, and a pinned 4-element golden float vector.
- `int.test.js` — closed-interval invariants, endpoint coverage,
  bounds and unsafe-range error paths, determinism, and a
  draw-width sweep that exercises every tier of the
  rejection-sampling staircase (1-byte through 53-bit slow path).
- `pure-rand.test.js` — `adaptToPureRandomGenerator` /
  `adaptFromPureRandomGenerator` round-trip equivalence with a
  fresh twin, plus distribution-shape sanity.
- `fast-check.test.js` — drives `fc.assert` with a
  ChaCha12-backed `randomType` to confirm the adapter satisfies
  `fast-check`'s contract end-to-end.
- `random.bench.js` — local microbenchmarks against an inlined
  ChaCha20 reference (`_chacha20.js`) and an `xorshift128+`
  baseline (`_xorshift.js`).

`packages/chacha12/test/chacha12.test.js` — three published
ChaCha12 keystream vectors (TC1, TC4, TC8) from
`draft-strombergson-chacha-test-vectors-01`, plus block-boundary
equivalence (a 192-byte single pull vs piecewise 7/57/64/32/32
pulls from a freshly-seeded twin) and `RandomSource`-shape
conformance.

The `@endo/hex` benchmarks and `@endo/ocapn` fuzz tests now run
against the new chacha12 source; existing assertions are preserved.
No new CI jobs required.

- `@endo/random` and `@endo/chacha12` are new packages — no prior
  usage to break.
- `@endo/hex` and `@endo/ocapn` change only test and benchmark
  internals; their published runtime surface is unchanged.  The
  duplicated `_xorshift.js` helpers were never exported.

### Upgrade Considerations

Pure additive change for production systems:

- No runtime dependency added to `@endo/hex` or `@endo/ocapn`
  consumers (the `@endo/chacha12` and `@endo/random` links are
  `devDependencies`).
- No data migration, no on-chain implications.
- Not a breaking change.


.changeset/endo-chacha12.md
```diff
@@ -0,0 +1,26 @@
...
+family.
+The block function is identical to ChaCha20 modulo the round count
+(6 double-rounds vs 10), so the implementation, API, and harden
+discipline mirror the sibling 20-round implementation.
+ChaCha12 trades cryptographic margin for throughput; for
+deterministic test fixtures, property-based testing, and fuzzing
+the extra speed is generally the right tradeoff.
+
+The keystream is verified against three published test vectors
+from `draft-strombergson-chacha-test-vectors` (TC1, TC4, TC8).
```
gibson042: Might as well make this a hyperlink to https://datatracker.ietf.org/doc/html/draft-strombergson-chacha-test-vectors-01 .

packages/chacha12/BENCH.md
```diff
@@ -0,0 +1,114 @@
...
+## Test bed
+
+| Field    | Value                                          |
+| -------- | ---------------------------------------------- |
+| CPU      | AMD Ryzen AI MAX+ 395 w/ Radeon 8060S (32 vCPU) |
+| RAM      | 128 GiB                                        |
+| OS       | Linux 6.14 (Ubuntu 24.04)                      |
+| Node     | 22.22.2                                        |
+| Arch     | x64                                            |
+| Harness  | `packages/chacha12/test/random.bench.js`       |
```
gibson042: This _near_-alignment... 😐

packages/chacha12/README.md
```diff
@@ -0,0 +1,95 @@
...
+The block function is identical to
+[ChaCha20](https://datatracker.ietf.org/doc/html/rfc8439) modulo the
+loop count: 6 double-rounds (= 12 rounds) instead of 10 (= 20).
+The reduced round count trades cryptographic safety margin for
+speed.
+
+This package is the ChaCha12 sibling of `@endo/chacha20` (the
+20-round variant) and `@endo/xorshift` (a non-cryptographic PRNG);
+choose among them by package name based on the throughput vs margin
+tradeoff a given consumer needs.
```
gibson042: Do all three packages cross-reference the other two? It seems valuable for them to implement and maintain a consistent API, and I'm wondering whether that and other commonalities should rely upon something like importing from an `@endo/prng` package.

packages/chacha12/README.md
```diff
@@ -0,0 +1,95 @@
...
+throughput-vs-margin knob), not a cipher recommendation.
+
+ChaCha12 (like ChaCha20) **must not be used to derive cryptographic
+keys** when the seed is caller-supplied.
+This package is a PRNG, not a key-derivation function.
+
+The PRNG method names align with the
+[TC39 proposal-random-functions](https://tc39.es/proposal-random-functions/)
+(Stage 1) so consumers can swap in a `Random.Seeded` instance later
+without changing call sites.
```
gibson042: Given that the proposal is at stage 1, this claim is too strong. I'd prefer to pare down the API to the point where such a swap becomes possible in the future without breaking changes here regardless of dramatic evolution in the proposal. Even if we need all four abilities (draw sub-unitary non-negative float, draw integer from closed interval, draw fresh Uint8Array, fill Uint8Array), perhaps it makes sense to start with just a block stream here and a more ergonomic translating wrapper in the previously-suggested common package?
```ts
import { BLOCK_SIZE, makeChaCha12 } from '@endo/chacha12';
import { decodeHex } from '@endo/hex';
import { makeRandomFromBlockFiller } from '@endo/prng';

const { pullBlock }: { pullBlock: (out: Uint8Array) => void } = makeChaCha12(
  decodeHex('b0b5c0ffeefacade'.repeat(4)),
);

// `makeRandomFromBlockFiller` makes a proposal-random-functions Random from a
// `pullBlock`-shaped Uint8Array-filling function.
const prng: {
  random: () => number;
  int: (lo: number, hi: number) => number;
  bytes: (n: number) => Uint8Array;
  fillBytes: (buf: Uint8Array, start?: number, end?: number) => Uint8Array;
} = makeRandomFromBlockFiller(pullBlock, { blockSize: BLOCK_SIZE });

assert(0 <= prng.random() && prng.random() < 1);
assert(Math.abs(prng.int(10, 20) - 15) <= 5);
```

packages/chacha12/README.md
```diff
@@ -0,0 +1,95 @@
...
+All returned objects pass through `harden` from `@endo/harden`.
+
+## Determinism
+
+- `random()` reads 8 keystream bytes, masks the top 11 bits, and
+  divides by `2 ** 53`.
+  Each respective returned value from streams with the same seed is
+  equal across runs and engines because the float is computed by
+  dividing a 53-bit integer by an exact power of two, with no
+  engine-dependent rounding.
```
gibson042:
```suggestion
- `random()` ensures that each respective returned value from streams with the
  same seed is equal across runs and engines by internally constructing a 53-bit
  integer and dividing it by `2 ** 53` (which avoids engine-dependent rounding).
```

packages/chacha12/README.md
```diff
@@ -0,0 +1,95 @@
...
+- `int(lo, hi)` uses rejection sampling on the 53-bit float to
+  eliminate modulo bias.
```
gibson042: This doesn't belong in a Determinism section.
```suggestion
```

packages/chacha12/README.md
gibson042: And for bonus points, we should also be able to represent a ChaCha12 as a [pure-rand](https://www.npmjs.com/package/pure-rand) [RandomGenerator](https://github.com/dubzzz/pure-rand/blob/b50c95b380fc0a625cb9797d94a566212fce4a78/src/types/RandomGenerator.ts) for compatibility with [fast-check `randomType`](https://fast-check.dev/docs/api/interfaces/Parameters/#randomtype).
```ts
import fc from 'fast-check';
import type { RandomType as FastCheckRandomType } from 'fast-check';
import type { RandomGenerator } from 'pure-rand/types/RandomGenerator';
import {
  BLOCK_SIZE,
  makeChaCha12Kit,
  adaptToPureRandomGenerator,
} from '@endo/chacha12';

/*
const adaptToPureRandomGenerator = ({
  getState,
  pullBlock,
}: ChaCha12Kit): RandomGenerator => {
  const buf8 = new Uint8Array(BLOCK_SIZE);
  const buf32 = new Int32Array(buf8.buffer);
  let prevBuf32Idx = 0;
  return {
    clone: () =>
      adaptToPureRandomGenerator(makeChaCha12Kit({ state: getState() })),
    next: () => {
      if (prevBuf32Idx === 0) {
        // Refill and reset.
        pullBlock(buf8);
        prevBuf32Idx = buf32.length;
      }
      prevBuf32Idx -= 1;
      return buf32[prevBuf32Idx];
    },
    getState: () => getState(),
  };
};
*/

const randomGenerator: RandomGenerator = adaptToPureRandomGenerator(
  makeChaCha12Kit({ seed: decodeHex('b0b5c0ffeefacade'.repeat(4)) }),
);

// Our fast-check adapter.
const chaCha12RandomType: FastCheckRandomType = (int32Seed: number) => {
  const chaChaSeed = new Int32Array(8).fill(int32Seed);
  return adaptToPureRandomGenerator(
    makeChaCha12Kit({ seed: new Uint8Array(chaChaSeed.buffer) }),
  );
};

// A deterministically failing PBT fed by ChaCha12.
fc.assert(
  fc.property(fc.integer(), n => n >= 0),
  { randomType: chaCha12RandomType, seed: 0xdeadbeef },
);
```

packages/hex/test/decode.bench.js
```diff
@@ -187,16 +184,43 @@ const pairMapTableDecode = (string, name = '<unknown>') => {
...
-  const bytes = new Uint8Array(size);
-  const prng = new XorShift(defaultSeed);
-  for (let i = 0; i < size; i += 1) {
-    bytes[i] = Math.floor(prng.random() * 256);
-  }
-  return bytes;
-};
+// Deterministic PRNG, same seed bytes as other Endo fuzz tests
+// repeated to fill the 32-byte ChaCha12 key.
+const defaultSeed = Uint8Array.of(
```
gibson042: Maybe define this once for import from e.g. packages/hex/tools/benchmarking-seed.js?

packages/chacha12/src/chacha12.js
```diff
@@ -0,0 +1,197 @@
...
+  //   state[13..15] = nonce (zero in this PRNG)
+  const baseState = new Uint32Array(16);
+  baseState[0] = C0;
+  baseState[1] = C1;
+  baseState[2] = C2;
+  baseState[3] = C3;
+  for (let i = 0; i < 8; i += 1) {
+    baseState[4 + i] = readU32LE(key, i * 4);
+  }
+  // counter = 0, nonce = 0 (already zeroed by Uint32Array).
```
gibson042:
```suggestion
  const baseState = chacha12State(key);
```

packages/chacha12/src/chacha12.js
```diff
@@ -0,0 +1,197 @@
...
+  state[1] = C1;
+  state[2] = C2;
+  state[3] = C3;
+  for (let i = 0; i < 8; i += 1) state[4 + i] = readU32LE(key, i * 4);
+  state[12] = counter >>> 0;
+  state[13] = readU32LE(nonce, 0);
+  state[14] = readU32LE(nonce, 4);
+  state[15] = readU32LE(nonce, 8);
+  return state;
+};
```
gibson042, line 187: Tweaking for convenient use by `makeChaCha12Source`:
```suggestion
/**
 * Builds a 16-word ChaCha12 state from a 32-byte key, optional 12-byte
 * nonce, and optional 32-bit counter.  Exported for ChaCha12 test
 * vectors.
 *
 * @param {Uint8Array} key 32 bytes
 * @param {Uint8Array} [nonce] 12 bytes
 * @param {number} [counter] unsigned 32-bit
 * @returns {Uint32Array}
 */
export const chacha12State = (key, nonce = undefined, counter = 0) => {
  if (!(key instanceof Uint8Array) || key.length !== 32) {
    throw TypeError('chacha12 key must be 32 bytes');
  }
  if (nonce && (!(nonce instanceof Uint8Array) || nonce.length !== 12)) {
    throw TypeError('chacha12 nonce must be 12 bytes');
  }
  const state = new Uint32Array(16);
  state[0] = C0;
  state[1] = C1;
  state[2] = C2;
  state[3] = C3;
  for (let i = 0; i < 8; i += 1) state[4 + i] = readU32LE(key, i * 4);
  state[12] = counter >>> 0;
  if (nonce) {
    state[13] = readU32LE(nonce, 0);
    state[14] = readU32LE(nonce, 4);
    state[15] = readU32LE(nonce, 8);
  }
  return state;
};
```

packages/chacha12/src/random.js
```diff
@@ -0,0 +1,253 @@
...
+      i += n;
+    }
+  };
+
+  /**
+   * Pulls 8 keystream bytes, masks the high 11 bits to 0, and
+   * divides the resulting 53-bit integer by `2 ** 53`.  The result
+   * is a float in `[0, 1)`, deterministic for a given seed across
+   * engines.
+   */
```
gibson042:
```suggestion
  /**
   * Produces a float in `[0, 1)` that is deterministic for a given seed
   * across all engines by constructing an unsigned 53-bit integer and
   * dividing it by `2 ** 53`.
   */
```

packages/chacha12/src/random.js
```diff
@@ -0,0 +1,253 @@
...
+  // Read a 32-bit unsigned little-endian integer from the keystream.
+  // Used by `int()` for ranges that fit in 32 bits, avoiding the
+  // float round-trip through `random()`.
+  const readU32 = () => {
+    const b0 = readByte();
+    const b1 = readByte();
+    const b2 = readByte();
+    const b3 = readByte();
+    return (b0 | (b1 << 8) | (b2 << 16) | (b3 << 24)) >>> 0;
+  };
```
gibson042:
```suggestion
  /** Read a 16-bit unsigned little-endian integer from the keystream. */
  const readU16 = () => {
    const b0 = readByte();
    const b1 = readByte();
    return (b0 | (b1 << 8)) >>> 0;
  };

  /** Read a 24-bit unsigned little-endian integer from the keystream. */
  const readU24 = () => {
    const b0 = readByte();
    const b1 = readByte();
    const b2 = readByte();
    return (b0 | (b1 << 8) | (b2 << 16)) >>> 0;
  };

  /** Read a 32-bit unsigned little-endian integer from the keystream. */
  const readU32 = () => {
    const b0 = readByte();
    const b1 = readByte();
    const b2 = readByte();
    const b3 = readByte();
    return (b0 | (b1 << 8) | (b2 << 16) | (b3 << 24)) >>> 0;
  };

  /** Read a 53-bit unsigned little-endian integer from the keystream. */
  const readU53 = () => {
    // Assemble the integer from two 32-bit halves, masking off the top
    // 11 bits of the upper half to leave 21 high bits + 32 low bits =
    // 53 bits.
    const lo = readU32();
    const hi24 = readU24();
    const hi21 = hi24 & 0x1fffff;
    return hi21 * 4294967296 + lo;
  };
```

packages/chacha12/src/random.js
```diff
@@ -0,0 +1,253 @@
...
+    const b6 = readByte();
+    const b7 = readByte();
+    const lo = (b0 | (b1 << 8) | (b2 << 16) | (b3 << 24)) >>> 0;
+    // Mask top 11 bits of the high word to 0, keeping 21 bits.
+    const hi = ((b4 | (b5 << 8) | (b6 << 16) | (b7 << 24)) >>> 0) & 0x1fffff;
+    // hi * 2 ** 32 + lo, divided by 2 ** 53:
+    //   = hi * 2 ** -21  +  lo * 2 ** -53
+    // Computed as a single multiply by POW2_M53 of the 53-bit int.
+    return (hi * 4294967296 + lo) * POW2_M53;
+  };
```
gibson042:
```suggestion
  const random = () => readU53() * POW2_M53;
```

packages/chacha12/src/random.js
```diff
@@ -0,0 +1,253 @@
...
+    // `[limit53, 2 ** 53)` and return `lo + (u % range)`.
+    const limit53 = Math.floor(9007199254740992 / range) * range;
+    for (;;) {
+      const lo32 = readU32();
+      const hi21 = readU32() & 0x1fffff;
+      const u = hi21 * 4294967296 + lo32; // exact 53-bit non-negative int.
+      if (u < limit53) {
+        return lo + (u % range);
+      }
+    }
```
gibson042: Let's eliminate some pathologies (e.g., `int(0, 2**31)` should not throw away almost half of the draws because its upper acceptable limit ends up at 2**31 + 1).
```suggestion
    // Draw the fewest bytes necessary to either match the requested range or
    // cover at least two repetitions of it, discarding draws in the range
    // beyond the last full repetition to avoid modulo bias while keeping the
    // discard fraction acceptably low.
    const { draw, limit } = (() => {
      // Read one byte to cover a range of exactly 256 or up to 128.
      if (range === 0x100 || range <= 0x80) {
        return { draw: readByte, limit: 0x100 - (0x100 % range) };
      }
      // Read two bytes to cover a range of exactly 65536 or up to 32768.
      if (range === 0x10000 || range <= 0x8000) {
        return { draw: readU16, limit: 0x10000 - (0x10000 % range) };
      }
      // ...and so on.
      if (range === 0x1000000 || range <= 0x800000) {
        return { draw: readU24, limit: 0x1000000 - (0x1000000 % range) };
      }
      if (range === 0x100000000 || range <= 0x80000000) {
        return { draw: readU32, limit: 0x100000000 - (0x100000000 % range) };
      }
      // The final limit is a 53-bit integer, which will discard just under half
      // of all possible draws when the range is just slightly above 2**52.
      return {
        draw: readU53,
        limit: Math.floor(9007199254740992 / range) * range,
      };
    })();
    for (;;) {
      const n = draw();
      if (n < limit) {
        return lo + (n % range);
      }
    }
```

packages/chacha12/src/random.js
```diff
@@ -0,0 +1,253 @@
...
+ */
+export const assertSeed = seed => {
+  if (!(seed instanceof Uint8Array)) {
+    throw TypeError('seed must be a Uint8Array');
+  }
+  if (seed.length !== 32) {
+    throw TypeError(
+      `seed must be 32 bytes (got ${seed.length}); ChaCha12 keys are 256 bits`,
+    );
+  }
```
gibson042: Suggestion: align with patterns in chacha12.js.
```suggestion
  if (!(seed instanceof Uint8Array) || seed.length !== 32) {
    throw TypeError('ChaCha12 seed must be a 32-byte Uint8Array');
  }
```

packages/chacha12/README.md
gibson042: Hmm, the other two packages don't even exist! Comment stands if there are plans to introduce them, but otherwise:
```suggestion
```

gibson042 (commented):
Some thoughts and suggestions on design and implementation.

I'd really like to see a more isolated core that can be used to construct multiple interfaces such as [proposal-random-functions](https://tc39.es/proposal-random-functions/) Random and [pure-rand](https://www.npmjs.com/package/pure-rand) RandomGenerator.

packages/random/random.types.d.ts
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * A `RandomSource` is a function that fills a `Uint8Array` with random
+ * bytes.  The shape mirrors `crypto.getRandomValues` (minus the
+ * return value), so the canonical browser/Node entropy source and a
+ * `@endo/chacha12`-backed source returned by `makeChaCha12(seed)` are
+ * both directly usable wherever a `RandomSource` is expected.
+ *
+ * Implementations MUST mutate the supplied buffer in place; they
+ * MUST NOT retain the buffer reference after the call returns.
```
gibson042:
```suggestion
 * Implementations MUST set every element of the supplied `Uint8Array`
 * and MUST NOT retain any reference after the call returns.
```

packages/random/random.types.d.ts
```diff
@@ -0,0 +1,36 @@
...
+ * without state-snapshot support the adapter returns an alias whose
+ * state is shared with the original.
+ */
+export interface PureRandomGenerator {
+  next(): [number, PureRandomGenerator];
+  unsafeNext(): number;
+  clone(): PureRandomGenerator;
+  min(): number;
+  max(): number;
+}
```
gibson042: Why pure-rand v5? The [v8 interface](https://github.com/dubzzz/pure-rand/blob/871ee3969f643df7463fa6483421b5e2e1c412b8/src/types/RandomGenerator.ts) is much better, and even preferred by fast-check: https://github.com/dubzzz/fast-check/blob/0858ecd84595c1245b8d26ca86e797643c2ad4e4/packages/fast-check/src/random/generator/RandomGenerator.ts

packages/random/fast-check.js
```diff
@@ -0,0 +1,131 @@
...
+ * Builds a `fast-check`-compatible `randomType` function from a
+ * factory that turns a 32-byte seed into a `RandomSource`.
+ * `fast-check` calls the returned function with a 32-bit signed
+ * integer seed and expects back a `pure-rand`-shaped
+ * `RandomGenerator`.  The integer seed is broadcast across a 32-byte
+ * buffer (compatible with ChaCha-style key inputs).
+ *
+ * @param {(seed: Uint8Array) => RandomSource} makeSourceFromSeed
+ * @returns {(int32Seed: number) => PureRandomGenerator}
+ */
```
gibson042: This should probably link to https://fast-check.dev/docs/api/interfaces/Parameters/#randomtype .

packages/random/test/int.test.js
```diff
@@ -0,0 +1,105 @@
...
+  }
+});
+
+// Range-aware rejection sampling: each tier of the draw-width
+// staircase covers a specific range domain.  This test exercises the
+// boundaries: single-byte (range <= 128 and exact 256), two-byte (up
+// to 32768 and exact 65536), three-byte (up to 8388608), four-byte
+// (up to 2^31 and exact 2^32), and the 53-bit slow path (range >
+// 2^32).  Each branch produces values in the requested closed
+// interval.
```
gibson042, line 90: Do we also want testing that verifies the expected count of read bytes for ranges of particular size?

packages/random/test/fast-check.test.js
```diff
@@ -0,0 +1,25 @@
...
+  fc.assert(
+    fc.property(fc.integer(), () => true),
+    /** @type {any} */ ({
+      randomType: chaCha12RandomType,
+      seed: 0xdeadbeef,
+      numRuns: 100,
+    }),
+  );
+  t.pass();
+});
```
gibson042: Likewise here; can we extend this to verify that the value passed as `randomType` is called with the provided seed and that bytes are actually read from the value it returns?

gibson042 (commented):
The packages/random/$noun.js files that just `export { … } from './src/$noun.js';` clutter up the package root; I understand why we don't want to expose the src/ files directly, but perhaps we should establish a convention of storing such files in an exports/ sibling directory.

Aside from that, I think the pure-rand v8 RandomSource interface would be a much better interop target than the v5 one.

Neither of these are blocking concerns, and the rest of the comments won't need another round of review.

kriskowal:
> The packages/random/$noun.js files that just `export { … } from './src/$noun.js';` clutter up the package root; I understand why we don't want to expose the src/ files directly, but perhaps we should establish a convention of storing such files in an exports/ sibling directory.
> 
> Aside from that, I think the pure-rand v8 RandomSource interface would be a much better interop target than the v5 one.
> 
> Neither of these are blocking concerns, and the rest of the comments won't need another round of review.

There’s a deeper reason. Not all tools and older versions of Node.js don’t recognize the exports directive in package.json, so mirroring key to value is necessary as a fallback. We could just put these thunks in src/ otherwise.

kriskowal:
Another note: I pulled out the fastcheck adapters. The depth of that conversation led me to understand that there are at least two random number adapters we would want depending on the fastcheck version and that warranted a dedicated package or or packages if we sought to use our prng instead of one that ships with fastcheck.

I also missed a couple notes in the review and will finish revising Monday.

gibson042:
> > The packages/random/$noun.js files that just `export { … } from './src/$noun.js';` clutter up the package root; I understand why we don't want to expose the src/ files directly, but perhaps we should establish a convention of storing such files in an exports/ sibling directory.
> 
> There’s a deeper reason. Not all tools and older versions of Node.js don’t recognize the exports directive in package.json, so mirroring key to value is necessary as a fallback. We could just put these thunks in src/ otherwise.

What software specifically are you concerned about here? The files in question are ECMAScript modules that use `import` statements, which are [supported in Node.js without a CLI flag since version 12.7.0](https://nodejs.org/api/esm.html#modules-ecmascript-modules:~:text=Loading%20ECMAScript%20modules%20no%20longer%20requires%20a%20command%2Dline%20flag), while [package.json `exports` are supported since Node.js version 12.20.0](https://nodejs.org/api/packages.html#exports:~:text=v14.13.0%2C%20v12.20.0-,Add%20support%20for%20%22exports%22%20patterns,-.). Is the gap between Node.js versions 12.7.0 and 12.20.0 really relevant? Just how far back does Endo support extend?


> > Aside from that, I think the pure-rand v8 RandomSource interface would be a much better interop target than the v5 one.
> 
> Another note: I pulled out the fastcheck adapters. The depth of that conversation led me to understand that there are at least two random number adapters we would want depending on the fastcheck version and that warranted a dedicated package or or packages if we sought to use our prng instead of one that ships with fastcheck.

I don't think we need to support anything other than the pure-rand v8 RandomSource interface, which is preferred by the latest version of fast-check and is the only one for which I have a use case (because I have seen Node.js crashes in #3053 that I suspect were ultimately caused by the PRNG internal to fast-check). And by "support" I mean that it must be possible to write source text as shown in https://github.com/endojs/endo/pull/3232#discussion_r3177216582 :
```ts
import fc from 'fast-check'; // ^4.6.0
import type { RandomType as FastCheckRandomType } from 'fast-check'; // ^4.6.0
import type {
  RandomGenerator as PureRandGenerator,
} from 'pure-rand/types/RandomGenerator'; // ^8.4.0
import { $EXPORT as makeChaCha12Source } from '@endo/chacha12;
import { adaptToChaCha12Seed, adaptToPureRandV8Generator } from './adapters.js';

// Our fast-check adapter.
const chaCha12RandomType: FastCheckRandomType = (int32Seed: number) => {
  const randomGenerator: PureRandGenerator = adaptToPureRandV8Generator(
    makeChaCha12Source({ seed: adaptToChaCha12Seed(int32Seed) }),
  );
  return randomGenerator;
};

// A deterministically failing PBT fed by ChaCha12.
fc.assert(
  fc.property(fc.integer(), n => n >= 0),
  { randomType: chaCha12RandomType, seed: 0xdeadbeef },
);
```

And the ability to write and use those `adaptToChaCha12Seed` and `adaptToPureRandV8Generator` functions should be covered by tests even if the package doesn't export anything related to them. The reason why this matters is that the pure-rand RandomGenerator interface includes a `getState` method that requires `@endo/chacha12` to export more than just a function whose return value has the same shape as `Crypto.getRandomValues`.

kriskowal:
The last hold out for exports support is eslint. I prompted a report that reveals there may be a workaround by using a forked plugin. https://github.com/endojs/endo-but-for-bots/issues/218

packages/chacha12/test/chacha12.test.js
```diff
@@ -0,0 +1,218 @@
+// @ts-check
+
+import test from '@endo/ses-ava/test.js';
+
+import { chacha12Block, chacha12State, makeChaCha12 } from '../src/chacha12.js';
+
+// Inline hex helpers: `@endo/hex` already depends on `@endo/chacha12`
```
kriskowal, line 14: Could break the cycle with a/the hex-test package.

.changeset/chacha12-next-getstate.md
```diff
@@ -0,0 +1,57 @@
...
+[Upstream review feedback on PR #3232](https://github.com/endojs/endo/pull/3232#issuecomment-4421637048)
+asked for `@endo/chacha12` to expose enough of its keystream state
+to support a `pure-rand` v8 / `fast-check` v4 adapter.  The closure
+returned by the previous shape encapsulated `baseState`, `counter`,
+`offset`, and `block` with no accessor, so an out-of-package
+adapter could not snapshot or clone the keystream.
+
+The new sibling package `@endo/chacha12-fast-check-test` (private)
+exercises the surface end-to-end against `fast-check@4`'s
+`randomType` parameter.
```
kriskowal: Please omit gratuitous process comment.

.changeset/endo-chacha12.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
+The package also ships:
+
+- `@endo/random/fast-check.js` -- bidirectional adapters between
+  `RandomSource` and `pure-rand`'s `RandomGenerator` (the contract
+  `fast-check` uses via the `randomType` parameter), plus a
+  `randomType` builder.  Imports nothing from `pure-rand` or
+  `fast-check`; depends only on those interfaces being structurally
+  compatible.
```
kriskowal: No longer true.

.changeset/endo-chacha12.md
```diff
@@ -0,0 +1,55 @@
...
+  seed, shared across Endo deterministic fuzz suites.
+
+Add `@endo/chacha12`: a pure-JavaScript ChaCha12 keystream.  The
+factory `makeChaCha12(key)` returns a `(out: Uint8Array) => void`
+function that conforms to `@endo/random`'s `RandomSource` and to
+`crypto.getRandomValues`-style ergonomics; pass it directly to the
+samplers.  The keystream is cross-checked against three published
+ChaCha12 test vectors from
+[`draft-strombergson-chacha-test-vectors-01`](https://datatracker.ietf.org/doc/html/draft-strombergson-chacha-test-vectors-01)
+(TC1, TC4, TC8).
```
kriskowal: The interface documented here is stale. We changed to an interface consistent with pure-rand-v8, structurally, with a fillRandomBytes function that satisfies RandomSource, like crypto.getRandomBytes.

.changeset/endo-chacha12.md
```diff
@@ -0,0 +1,55 @@
...
+
+Rewires the `@endo/hex` encode/decode benchmarks and the
+`@endo/ocapn` syrup and passable fuzz tests to draw deterministic
+bytes from `@endo/chacha12` directly, using the shared seed from
+`@endo/random/seeds.js`.
+
```
kriskowal: Omit as not interesting to package authors updating dependencies.

.changeset/chacha12-next-getstate.md
kriskowal, line 1: We have only made one change in this release cycle, when this PR merges. Please consolidate these changeset.

packages/ocapn/test/codecs/passable-fuzz.test.js
```diff
@@ -120,12 +122,10 @@ function fuzzyPassable(budget, random) {
...
-// Chris Hibbert really wanted the default i to be Bob's Coffee Façade,
-// which is conveniently exactly 64 bits long.
-const defaultSeed = [0xb0b5c0ff, 0xeefacade, 0xb0b5c0ff, 0xeefacade];
-
-const prng = new XorShift(defaultSeed);
-const random = () => prng.random();
+// Default seed shared across the hex/ocapn fuzz suites; see
+// `@endo/random/seeds.js`.
+const source = makeChaCha12(bobsCoffee64).fillRandomBytes;
+const randomNumber = () => randomFloat(source);
```
kriskowal: This was better when named `random`, unless there was a shadowing issue that I don’t see.

packages/ocapn/test/syrup/fuzz.test.js
```diff
@@ -94,19 +96,17 @@ function fuzzySyrupable(budget, random) {
...
-// Chris Hibbert really wanted the default i to be Bob's Coffee Façade,
-// which is conveniently exactly 64 bits long.
-const defaultSeed = [0xb0b5c0ff, 0xeefacade, 0xb0b5c0ff, 0xeefacade];
-
-const prng = new XorShift(defaultSeed);
-const random = () => prng.random();
+// Default seed shared across the hex/ocapn fuzz suites; see
+// `@endo/random/seeds.js`.
+const source = makeChaCha12(bobsCoffee64).fillRandomBytes;
+const randomNumber = () => randomFloat(source);
```
kriskowal: Ditto. Gratuitous rename.

packages/random/src/random.js
```diff
@@ -0,0 +1,26 @@
+// @ts-check
+
+import harden from '@endo/harden';
+
+import { randomUint53 } from './uint.js';
+
+// 1 / 2 ** 53.  Multiplying a 53-bit non-negative integer by this
+// produces a float in [0, 1) using deterministic integer arithmetic,
+// so the same seed produces the same float across runs and engines.
+const POW2_M53 = 1.1102230246251565e-16; // = 2 ** -53
```
kriskowal: Let’s add an assertion to the test suite to make sure this equivalence sticks.

packages/random/test/_make-source.js
```diff
@@ -0,0 +1,25 @@
...
+  return seed;
+})();
+
+/** @param {Uint8Array} seed */
+export const makeSource = seed => makeChaCha12(seed).fillRandomBytes;
+
+export const seedA = seedAll;
+export const seedB = seedRev;
+
+export const cloneSeed = s => Uint8Array.from(s);
```
kriskowal, line 25: Why do we clone the seed? Does makeChaCha12 retain the seed and mutate it? It should not.

packages/chacha12/src/chacha12.js
```diff
@@ -0,0 +1,400 @@
...
+ * @param {Uint32Array} baseState 16 u32 words, ownership transferred
+ *   to the generator (must not be retained by the caller).
+ * @param {number} initialCounter
+ * @param {number} initialOffset
+ * @param {Uint8Array} initialBlock 64 bytes, ownership transferred
+ *   (must not be retained by the caller).
+ * @returns {ChaCha12Generator}
+ */
+const makeGenerator = (
+  baseState,
```
kriskowal, line 227: Confirmed that we are indeed retaining this. That’s a foot-gun. Please allocate your own copy.

packages/chacha12/test/fill-random-bytes.bench.js
```diff
@@ -0,0 +1,345 @@
...
+
+// Engine-portable nanosecond timer.
+const hasHrtime =
+  typeof globalThis.process === 'object' &&
+  globalThis.process !== null &&
+  typeof globalThis.process.hrtime === 'function' &&
+  typeof globalThis.process.hrtime.bigint === 'function';
+const nowNs = hasHrtime
+  ? () => Number(globalThis.process.hrtime.bigint())
+  : () => Date.now() * 1_000_000;
```
kriskowal: We can, frankly, presume that Ava tests are running under Node.js 20+.

packages/chacha12/test/fill-random-bytes.bench.js
```diff
@@ -0,0 +1,345 @@
...
+    out[i + k] = block[offset + k];
+    k += 1;
+  }
+};
+
+// `copyWithin` only operates within a single TypedArray, so to copy
+// from `block` to `out` we route through `set` after a subarray view.
+// This is essentially the same as `copySet`; included as a sanity
+// probe.
+const copyCopywin = (out, i, block, offset, n) => {
```
kriskowal: I would call this `copyWithin`

packages/random/CHANGELOG.md
kriskowal, line 1: This needs to be reframed for this package and mostly deleted.

packages/random/README.md
```diff
@@ -0,0 +1,98 @@
...
+sampler arguments.
+
+Names follow the TC39
+[proposal-random-functions](https://tc39.es/proposal-random-functions/)
+(Stage 1) translation `Random.method` -> `randomMethod`.
+
+| TC39 proposal | `@endo/random` |
+| --- | --- |
+| `Random.random()` | `random(source)` |
+| `Random.int(lo, hi)` | `randomInt(source, lo, hi)` |
```
kriskowal: Use prettier on this. Leave a note to do this in general for all new or altered Markdown documents going forward.

packages/random/test/random.test.js
```diff
@@ -0,0 +1,78 @@
...
+// the constant in `src/random.js` is what its accompanying comment
+// claims it is.
+test('random() multiplies randomUint53 by exactly 2 ** -53', t => {
+  /** @param {Uint8Array} out */
+  const oneSource = out => {
+    for (let i = 0; i < out.length; i += 1) out[i] = 0;
+    out[0] = 1;
+  };
+  t.is(random(oneSource), 2 ** -53);
+});
```
gibson042: This strikes me as overly brittle; it presumes a particular pattern of consumption that is not required (i.e., little-endian consumption of the buffer that retains its first octet and interprets it as the least significant of the value to be scaled down). Less brittle would be something like this:
```suggestion
// We expect random() to achieve a uniform distribution by scaling down
// randomUint53() output by exactly `2 ** -53`, but don't want to overly
// constrain _how_ the former maps octets into integers (consuming 8 vs. 7, big
// vs. little endian, etc.). So we provide a source that sets every bit to
// force maximum output, and if we ever need even more assurance could also use
// sources that set every bit of four contiguous octets after a prefix of 0 to 4
// fully-clear octets (0xFFFFFFFF_00..., 0x00FFFFFF_FF00...,
// 0x0000FFFF_FFFF00..., 0x000000FF_FFFFFF00..., 0x00000000_FFFFFFFF_00...),
// exactly one of which should drive random() output to be `2 ** -21 - 2 ** -53`
// (i.e., the one in which the set octets correspond exactly with the least
// significant 32 bits for randomUint53).
test('random() multiplies randomUint53 by exactly 2 ** -53', t => {
  /** @param {Uint8Array} out */
  const maxSource = out => {
    for (let i = 0; i < out.length; i += 1) out[i] = 0xFF;
  };
  t.is(random(maxSource), 1 - 2 ** -53);
});
```

gibson042: But on the other hand, this brittleness could be construed as a feature that prevents refactors from accidentally slipping in breaking changes. In which case I would highlight _that_ in the explanatory comments (but also still do the set-all-bits analog as well).

kriskowal: Agreed. This was emitted in order to provide assurance that the max float int constant matched the value prescribed in the comment, but I like narrowing this fence. I’m going to propose both all set, none set, 52 bits set, and 53 bits set.

.changeset/endo-chacha12.md
```diff
@@ -0,0 +1,30 @@
...
+import { random } from '@endo/random/random.js';
+import { randomInt } from '@endo/random/int.js';
+```
+
+The package also ships `@endo/random/seeds.js`, exporting the canonical `bobsCoffee32` 32-byte seed shared across Endo deterministic fuzz suites.
+
+Add `@endo/chacha12`: a pure-JavaScript ChaCha12 keystream.
+The factory `makeChaCha12(key)` returns a `ChaCha12Generator` record `{ next, getState, clone, fillRandomBytes }`.
+The `fillRandomBytes` method has the shape `(out: Uint8Array) => void`, conforming to `@endo/random`'s `RandomSource` and to `crypto.getRandomValues`-style ergonomics; it can be passed directly to the samplers.
+The remaining methods expose the keystream's internal state (snapshot via `getState`, independent copy via `clone`, signed-int32 pull via `next`) so the generator satisfies `pure-rand` v8's `RandomGenerator` interface structurally, driving `fast-check` v4's `randomType` parameter without a separate adapter.
```
gibson042:
```suggestion
The remaining methods expose the keystream's internal state (snapshot via `getState`, independent copy via `clone`, signed-int32 pull via `next`) so the generator satisfies the [`pure-rand` v8 `RandomGenerator` interface](https://github.com/dubzzz/pure-rand/blob/v8.0.0/src/types/RandomGenerator.ts) structurally and can be used directly as a [`fast-check` v4 `randomType` parameter](https://fast-check.dev/docs/api/interfaces/Parameters/#randomtype).
```

packages/chacha12-fast-check-test/test/_random-type.js
```diff
@@ -0,0 +1,65 @@
+// @ts-check
+/* eslint no-bitwise: ["off"] */
+
+// Integration-test package only.  The adapter exported here exists
+// for the test in `test/fast-check.test.js`; it is intentionally
+// not published as a reusable library surface.  See
+// `packages/random-fast-check/` (designed in
+// `designs/random-pure-rand-v8-interface.md`) for the
```
gibson042: Neither `packages/random-fast-check/` nor `designs/random-pure-rand-v8-interface.md` seem to exist.

packages/chacha12-fast-check-test/test/_random-type.js
```diff
@@ -0,0 +1,65 @@
...
+ * @property {(out: Uint8Array) => void} fillRandomBytes
+ */
+
+/**
+ * Builds a fast-check `randomType` callback from `@endo/chacha12`.
+ * The returned callback takes a 32-bit signed seed (the shape
+ * `fast-check` invokes its `randomType` with) and produces a
+ * `ChaCha12Generator` whose `next` / `clone` / `getState` surface is
+ * structurally compatible with `pure-rand` v8's `RandomGenerator`
+ * (and therefore with `fast-check@4`'s `randomType` parameter).
```
gibson042: Consistency:
```suggestion
 * structurally compatible with `pure-rand@8`'s `RandomGenerator`
 * (and therefore with `fast-check@4`'s `randomType` parameter).
```

packages/chacha12-fast-check-test/test/_random-type.js
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * The seed is broadcast across the full 32-byte ChaCha12 key by
+ * little-endian writes into 8 successive u32 slots.  This is the
+ * same broadcast pattern proposed for `@endo/random-fast-check`'s
+ * `makeRandomTypeFromSeedV8` builder; we restate it inline here to
+ * keep this integration-test package free of any fast-check adapter
+ * production code.
```
gibson042: I don't see an `@endo/random-fast-check`.

packages/chacha12-fast-check-test/test/fast-check.test.js
```diff
@@ -0,0 +1,171 @@
...
+      { randomType, seed: 12_345, numRuns: 50 },
+    );
+    return seen;
+  };
+
+  const runA = collect();
+  const runB = collect();
+  t.deepEqual(runA, runB);
+  // And not trivial: the generator actually walked the space.
+  t.true(new Set(runA).size > 1);
```
gibson042:
```suggestion
  t.true(runA.length > 1);
```

packages/chacha12-fast-check-test/test/fast-check.test.js
```diff
@@ -0,0 +1,171 @@
...
+  // shrink toward 101 (or some small value > 100) deterministically
+  // given the same seed and randomType.
+  const randomType = makeChaCha12RandomType();
+  const findCounterexample = () => {
+    const r = fc.check(
+      fc.property(fc.integer({ min: 0, max: 1000 }), n => n <= 100),
+      { randomType, seed: 7, numRuns: 200 },
+    );
+    return r.failed ? r.counterexample : null;
+  };
```
gibson042, line 101:
```suggestion
  const findCounterexample = () => {
    const distinctResults = new Set();
    const r = fc.check(
      fc.property(fc.integer({ min: 0, max: 1000 }), n => {
        const pass = n <= 100;
        distinctResults.add(pass);
        return pass;
      }),
      { randomType, seed: 7, numRuns: 200 },
    );
    t.deepEqual(
      [...distinctResults].sort(),
      [false, true],
      'fast-check explored the domain',
    );
    return r.failed ? r.counterexample : null;
  };
```

packages/chacha12/src/chacha12.js
```diff
@@ -0,0 +1,400 @@
...
+// stream cipher.  The block function is otherwise identical to
+// ChaCha20 (same quarter round, same state layout, same "expand
+// 32-byte k" constants, same little-endian conventions, same final
+// state add).  The only difference is the loop count: 6
+// double-rounds in ChaCha12 (= 12 rounds), 10 in ChaCha20.
+//
+// `makeChaCha12(key)` returns a `ChaCha12Generator` record with
+// `next`, `getState`, `clone`, and `fillRandomBytes` methods.  The
+// shape was chosen to align with `pure-rand` v8's `RandomGenerator`
+// interface (the contract `fast-check@4` uses to drive property-based
```
gibson042:
```suggestion
// shape was chosen to align with `pure-rand@8`'s `RandomGenerator`
// interface (the contract `fast-check@4` uses to drive property-based
```

gibson042 (approved):
LGTM after addressing (or explicitly choosing not to address) some final comments and suggestions.

PR #3239, opened 2026-05-05
chore(docs): typedoc configuration fixes

boneskull:
I was having a terrible time building the docs; excluded packages don't get excluded, some packages build using arbitrary (unknown) TS configuration causing errors, etc.

This _pragmatic_ change does three (3) things:

1. Suppresses BS errors
2. Improves performance
3. More accurate package-specific docs

## Suppress Bogus Errors

Traditionally in Endo, if the docs build uniquely fails CI, the problem is _TypeDoc-specific_ and relates to TS misconfiguration (or potentially TypeDoc bugs).  Errors reported do not typically reflect _actual type errors_.  Any type-checking error will fail the docs build outright.

I also feel comfortable arguing that the responsibility of type-checking the packages does _not_ lie with TypeDoc.

This is accomplished by:

1. Extending the base `typedoc.base.json` from `packages/*/typedoc.json`
2. Updating `typedoc.base.json` to include `skipErrorChecking: true`

## Faster Build

By providing a `typedoc.json` for each package which refers to `tsconfig.build.js`, tests are omitted from consideration. Further, skipping error checking also speeds things up.

## Better Docs

By providing a `typedoc.json` for each package, we can specify an explicit list of entrypoints for TypeDoc to consider. While this is typically not necessary in a composite TS project—where TypeDoc can more easily discover entrypoints—it is useful in Endo's case, since the entry points are `.js` sources instead of generated declarations.  These `entryPoints` fields should match the sources as specified in `exports` for each package (with an exception for `ses`, which was unmodified).


boneskull:

- branch: `master`
  - **[chore(docs): typedoc configuration fixes #3239](https://github.com/endojs/endo/pull/3239)** - branch: `boneskull/typedoc-fix` *(this PR)*

---

boneskull:
This whole thing was due to Yarn.

PR #3241, opened 2026-05-06
fix(module-source): Fix AST traversal and hidden variable censor error #1596

### LINKED ISSUE #1596: source-transformation error
I wrote a tool that uses `makeArchive` to deduce the full set of imported source files from a given entry point (to help answer the question "if I change file X, how many bundles/deployed things do I have to upgrade/re-deploy?"). It works fine on everything in agoric-sdk that we're currently bundling, of course, but I also want to ask about non-bundled things, like https://github.com/Agoric/agoric-sdk/blob/master/packages/cosmic-swingset/src/entrypoint.js (the starting point for the JS portion of the Agoric chain executable). I hacked around this code's desire to import e.g. Node.js builtin modules, by adding a fake `modules: { fs: null, process: null, .. }` record, to provide exits for those non-source modules. That works pretty well.

But, for at least two source modules, I got an error during `makeArchive`. One was when pointed at https://github.com/Agoric/agoric-sdk/blob/master/packages/SwingSet/src/index.js , which ends with some type-related imports:
```js
export * from './types-external.js';

export * as shape from './typeGuards.js';
```

If I comment those out, or if I point `makeArchive` directly at `types-external.js` or `typeGuards.js`, it's fine. But if I point it at `index.js`, I get:
```
err (TypeError#1)
TypeError#1: Failed to load module "./src/index.js" in package "file:///home/node/trees/agoric-sdk/packages/SwingSet/" (1 underlying failures: Error transforming source in <unknown>: Cannot read properties of undefined (reading 'name')
```

Likewise, I have problems with https://github.com/Agoric/agoric-sdk/blob/master/packages/cosmic-swingset/src/anylogger-agoric.js , which gives me:
```
err (TypeError#1)
TypeError#1: Failed to load module "./src/anylogger-agoric.js" in package "file:///home/node/trees/agoric-sdk/packages/cosmic-swingset/" (1 underlying failures: Error transforming source in <unknown>: Cannot read properties of undefined (reading 'buildError')
```

I know this error is coming from https://github.com/endojs/endo/blob/master/packages/static-module-record/src/transform-analyze.js#L38 , which means it's coming from a call to the transformer created by `const transformSource = makeTransformSource(makeModulePlugins, babel);`, but I don't know how to investigate further.

The core of my tool looks like:
```js
const builtinsToIgnore = [
  'buffer',
  ..
];
async function findContents(entryfn) {
  const moduleLocation = url.pathToFileURL(path.resolve(entryfn));
  const modules = {};
  builtinsToIgnore.forEach(name => modules[name] = null); // value is unused
  const options = { modules };
  const archiveBytes = await makeArchive(readPowers, moduleLocation, options);
  return extractMemberNames(archiveBytes);
}
```

cc @kriskowal

kriskowal:
Closes #1596

## Description

Fixes the two `TypeError` crashes in `@endo/module-source`'s analyzer reported in #1596. Both fixes live in `packages/module-source/src/`.

**Crash 1 (`reading 'name'`).** The export-specifier loop destructured `local` from each specifier and read `local.name`. `ExportNamespaceSpecifier` nodes (`export * as ns from 'src'`) have no `local` property, so the analyzer threw `TypeError: Cannot read properties of undefined (reading 'name')` whenever the source contained the Stage 4 namespace re-export shorthand. The fix substitutes the already-computed `importFrom` variable, which is `'*'` for the namespace form and `local.name` for every other specifier type (`01380110b2`).

**Crash 2 (`reading 'buildError'`).** The analyzer calls `path.buildCodeFrameError(...)` from several visitors to surface SyntaxErrors for unrecognized constructs. Internally `buildCodeFrameError` reads `this.hub.buildError(...)`, but `traverse(ast, visitor)` initializes child paths with `hub === undefined`. The first such visitor to fire therefore raised the cryptic `TypeError: Cannot read properties of undefined (reading 'buildError')` instead of the intended diagnostic, masking the analyzer's actual complaint. The fix constructs a synthetic `parentPath` carrying a Babel `Hub` and threads it as the fifth argument to each `traverse` call so child paths inherit it (`e9631f5759`).

### Security Considerations

No impact. Both changes are defensive fixes inside the static analyzer; they remove a crash path and do not introduce new authority, new APIs, or new boundaries.

### Scaling Considerations

No impact. The Hub-supplying wrapper is a single allocation per `traverse` call (of which the analyzer makes a fixed number per module). The `importFrom` substitution is a variable swap inside an existing loop iteration.

No user-facing documentation change. The Stage 4 `export * as ns from 'src'` form now works in `@endo/module-source` consumers without any opt-in. Sources that previously triggered the cryptic `buildError` TypeError will now surface the intended `SyntaxError` with a code frame.

### Testing Considerations

`packages/module-source` ships 53 tests (49 baseline plus four regression assertions added by this PR):

- Two new tests cover the namespace re-export form, both in isolation against the analyzer and end-to-end through SES.
- Two pre-existing reserved-identifier tests were sharpened to assert `instanceOf: SyntaxError`, a regex on the inner message, and a `t.notRegex` against the buildError-undefined `TypeError`.

Reverting either fix in isolation causes the corresponding new assertions to fail closed with the original crash, confirming both are load-bearing.

Strictly additive. Sources that previously parsed continue to parse identically; sources that previously crashed with `TypeError` now either parse (the namespace re-export case) or surface the intended `SyntaxError` (the Hub-less case). No prior usage pattern is broken.

### Upgrade Considerations

None.


packages/module-source/src/transform-source.js
```diff
@@ -9,6 +9,40 @@ const visitorFromPlugin = plugin => plugin({ types: babelTypes }).visitor;
...
+ * Threading a hub through a wrapper `parentPath` causes child paths to
+ * inherit it (see `NodePath.get`: `if (!hub && parentPath) hub = parentPath.hub;`),
+ * so error reporting from `path.buildCodeFrameError` produces the intended
+ * `SyntaxError` instead.
+ *
+ * The wrapper holds the parsed `File` as its sole child, keyed `'container'`,
+ * which mirrors how Babel's own compiler driver wires up the root path.
+ *
+ * @param {object} ast - the parsed Babel `File` node.
+ * @returns {object} a `NodePath` whose child paths inherit a `Hub`.
```
boneskull:
```suggestion
 * @param {File} ast - the parsed Babel `File` node.
 * @returns {NodePath} a `NodePath` whose child paths inherit a `Hub`.
```

packages/module-source/test/module-source.test.js
```diff
@@ -441,16 +441,47 @@ export default async function (arg) { return arg; }
...
+  const err = t.throws(
+    () => {
+      const _ = new ModuleSource(
+        `const $h\u034f_import = 123; $h\u034f_import`,
+      );
+    },
+    { instanceOf: SyntaxError, message: /is reserved/ },
+  );
+  // Ensure the Hub-backed diagnostic flowed through, not the raw TypeError
+  // from `undefined.buildError`.
```
boneskull: might be better expressed as a custom assertion message to `t.notRegex`

boneskull (approved):

PR #3243, opened 2026-05-06
fix(compartment-mapper,bundle-source): Improve error diagnostic for nameless packages

### LINKED ISSUE #1845: Bundler: better diagnostic if package.json lacks a `name`
## Describe the bug

If one bundles an entrypoint with a `package.json` that does not have a `name`, the compartment mapper fails with an uninformative diagnostic. The bundler should produce a better error if it finds a `package.json` without a `name`.

## Steps to reproduce

1. Create a package.json without a name.
2. Create an entry.js module.
3. Attempt to bundle with `bundle-source` provided by `@endo/bundle-source`

## Expected behavior

The bundler should fail with an error diagnostic referring the the `package.json` that did not have a name, possibly suggesting one based on the parent directory name. The suggestion cannot be reliably inferred because of workspaces.

## Platform environment

All

kriskowal:
Closes: #1845

## Description

When a `package.json` lacks a `name` field, the compartment mapper previously produced cryptic downstream failures: a bundler error of the form `Cannot import name X (has ...)`, or a compartment with `name: undefined` whose location was not reported anywhere. This change surfaces a precise diagnostic at the point the offending descriptor is adopted as the root of a compartment, blaming the file by URL and suggesting a fix.

The new error reads:

```
package.json at "file:///.../no-name-pkg/package.json" must have a "name" field; consider naming it after the parent directory
```

A new helper `assertPackageDescriptorHasName(packageDescriptor, packageDescriptorLocation)` lives in `packages/compartment-mapper/src/node-modules.js` and is invoked at the three sites that adopt a descriptor as the root of a compartment: `mapNodeModules` for the entry package's `package.json`, `graphPackages` as a defensive double-check on the entry package, and `findPackage` for every dependency discovered under `node_modules/`.

`assertPackageDescriptor` is unchanged structurally and still validates plain-object shape without requiring `name`, so `readDescriptorUpwards` can continue to walk ancestor `package.json` files that legitimately exist only to declare `type: "module"` for sub-folder module resolution. The existing `nested-pkg` test continues to pass.

A new fixture under `packages/compartment-mapper/test/fixtures-no-name/` and three regression tests in `packages/compartment-mapper/test/no-name.test.js` cover the entry-package, dependency, and bundler paths. Reverting the new check causes all three to fail with the original cryptic outputs, confirming they are load-bearing.

A second commit names the previously-anonymous `packages/bundle-source/demo/endo/package.json` fixture (`name: "endo-demo"`, `private: true`) so that the new diagnostic does not fire on `bundle-source`'s own `marshal-failure.test.js`.

### Security Considerations

No impact.

### Scaling Considerations

No impact.

No impact. The error message itself is the user-facing documentation of the new requirement.

### Testing Considerations

The new fixture and `no-name.test.js` cover the three call sites that adopt a descriptor. The unrelated `bundle-source` demo fixture was renamed in a separate commit to keep the two concerns reviewable independently.

A `package.json` that participates in the compartment graph and lacks a `name` now produces an error rather than a silent miscompile. Any consumer that relied on the previous undefined-name behavior will see a clear diagnostic and can add the missing field. Ancestor `package.json` files that exist only to declare `type: "module"` are unaffected.

### Upgrade Considerations

No impact.

### DIFF
--- a/packages/compartment-mapper/src/node-modules.js
+++ b/packages/compartment-mapper/src/node-modules.js
@@ -199,12 +199,13 @@ export const basename = location => {
 };
 
 /**
- * Asserts that the given value is a `PackageDescriptor`.
+ * Asserts that the given value is a plain-object `PackageDescriptor`.
  *
- * TODO: This only validates that the value is a plain object. As mentioned in
- * {@link PackageDescriptor}, `name` is currently a required field, but in the
- * real world this is not so. We _do_ make assumptions about the shape of a
- * `PackageDescriptor`, but it may not be worth eagerly validating further.
+ * Note: this does _not_ enforce the presence of a `name`. Callers that adopt
+ * a descriptor as the root of a compartment must additionally call
+ * {@link assertPackageDescriptorHasName} so that the diagnostic blames the
+ * correct `package.json`.
+ *
  * @param {unknown} allegedPackageDescriptor
  * @returns {asserts allegedPackageDescriptor is PackageDescriptor}
  */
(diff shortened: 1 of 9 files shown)


packages/compartment-mapper/src/node-modules.js
```diff
@@ -216,6 +217,33 @@ const assertPackageDescriptor = allegedPackageDescriptor => {
...
+ * @param {PackageDescriptor} packageDescriptor
+ * @param {string} packageDescriptorLocation - URL of the `package.json`
+ * file that produced this descriptor, used to attribute errors.
+ * @returns {void}
+ */
+const assertPackageDescriptorHasName = (
+  packageDescriptor,
+  packageDescriptorLocation,
+) => {
+  const { name } = packageDescriptor;
```
boneskull, line 239: Have we already asserted that `packageDescriptor` is a non-`null` `object` by the time we hit this?

boneskull (approved):

PR #3246, opened 2026-05-06
fix(ses) exported namespace consistency between VirtualModuleInstance and ModuleInstance

naugtur:
TBD

Closes: #XXXX
Refs: #XXXX

## Description

### Security Considerations

### Scaling Considerations

### Testing Considerations

### Upgrade Considerations

> For any breaking change, add a changeset (`yarn changeset`) with a `major` bump and migration instructions in the description. See the "Using Changesets" section in `CONTRIBUTING.md`.


naugtur:
A few cjs-compat tests are failing - their fixtures are documented with comments that Node.js is not supporting finding the named export. With this change we no longer find it either. I think we should drop these tests 

One set of attenuator tests is failing because it's now missing the default - should probably be introduced manually.

The test finding inconsistencies in live bindings between endo and node should probably be turned into a snapshot test and issues to fix compatibility if possible. Fixing one of the inconsistencies feels like it could be a side effect of this fix, but I didn't hit that target with my current attempt.

---
Failing tests:

> cjs-compat › fixtures-cjs-compat  
> cjs-compat › fixtures-cjs-compat-exit-module 

`Error: Expected 2.cjs to to have a named export called even`
- tests needs to be adjusted


> dynamic-require › dynamic require of ancestor
- not sure, needs debugging

> import-live-bindings-interop › import mutability compared with node.js
- probably best to convert to a snapshot test. deepEqual is easier to read for now.


> policy › policy - *  

`missing "default" in snapshots of module namespaces`
- "default" needs to be added to the list of expected exports in the implementation instead of being implicitly added always (or test needs to be updated - not 100% sure if the default should be there TBH)

packages/compartment-mapper/test/snapshots/import-live-bindings-interop.test.js.md
```diff
@@ -0,0 +1,47 @@
...
+    [
+      '    ESM importing ESM/defaultExport: both STATIC',
+      '    ESM importing ESM/namedLet: both LIVE',
+      '    ESM importing ESM/namedVar: both LIVE',
+      '    ESM importing ESM/namedConstValue: both LIVE',
+      '    ESM importing ESM/namespaceDefault: both STATIC',
+      '    ESM importing ESM/namespaceNamedLet: both LIVE',
+      '    ESM importing ESM/namespaceNamedVar: both LIVE',
+      '    ESM importing ESM/namespaceNamedConstValue: both LIVE',
+      '[!] ESM importing CJS (exports.*)/namedExport_namedA: node=STATIC endo=LIVE ',
```
kriskowal, line 20: This is a nice report.

packages/compartment-mapper/test/fixtures-import-mutability/index.js
```diff
@@ -0,0 +1,39 @@
...
+
+import {
+  getResults as cjsFromCjs,
+  getResultsOverwrite as cjsFromCjsOw,
+} from './consumer-cjs-from-cjs.cjs';
+import { getResults as cjsFromEsm } from './consumer-cjs-from-esm.cjs';
+
+export const getSummary = async () => {
+  // Wait for mutations (50ms)
+  await new Promise(resolve => setTimeout(resolve, 100));
```
kriskowal, line 16: We can almost certainly solve this coordination problem without timers. Consider injecting promise and resolver pairs into the environment instead and coordinating on those.

packages/compartment-mapper/test/snapshots/import-live-bindings-interop.test.js.md
```diff
@@ -0,0 +1,47 @@
...
+      '    CJS importing CJS (exports.*)/destructured_namedB: both STATIC',
+      '    CJS importing CJS (exports.*)/destructured_default: both STATIC',
+      '    CJS importing CJS (module.exports = {...})/wholeObject_namedA: both LIVE',
+      '    CJS importing CJS (module.exports = {...})/wholeObject_namedB: both LIVE',
+      '    CJS importing CJS (module.exports = {...})/wholeObject_default: both LIVE',
+      '    CJS importing CJS (module.exports = {...})/destructured_namedA: both STATIC',
+      '    CJS importing CJS (module.exports = {...})/destructured_namedB: both STATIC',
+      '    CJS importing CJS (module.exports = {...})/destructured_default: both STATIC',
+      '[!] CJS importing ESM (require())/namespace_namedLet: node=LIVE endo=STATIC ',
+      '[!] CJS importing ESM (require())/namespace_namedVar: node=LIVE endo=STATIC ',
```
kriskowal, line 45: Interesting, but defensible.

kriskowal (approved):

PR #3248, opened 2026-05-08
feat(ocapn): add op:flush for safe promise shortening

kumavis:
## Summary

Implements the receiver-initiated `op:flush` / `op:flush-done` handshake for OCapN's promise-shortening proposal. The goal is to preserve per-reference FIFO order when a peer is about to short-circuit an exported promise (typically as part of a third-party handoff).

## Protocol mechanics

- Bob, holding a remote reference to Alice's promise `p1` at export-position `N`, sends `<op:flush N>`.
- Alice, on receipt:
  - mints a fresh local promise `p'` (and its resolver `r'`),
  - swaps the value at her export-slot `p+N` for `p'` in place, preserving the slot's wire refcount,
  - replies with `<op:flush-done N>`.
- Bob's `flushExport()` promise resolves on the ack. Because messages on the connection are FIFO, by the time `op:flush-done` arrives Bob has already received every previously-sent message Alice had targeting `p1`, which is the precondition the proposal needs before continuing with the handoff.
- Subsequent deliveries Bob sends for position `N` queue on the new (unresolved) `p'` on Alice's side. They naturally buffer until something — typically the shortened reference — fulfills `p'`, at which point the buffered traffic flushes.

This implements [the latest design at ocapn/ocapn#11](https://github.com/ocapn/ocapn/issues/11). The full integration with the third-party handoff flow (auto-flushing during a 3PH and resolving `p'` from the handoff result) is intentionally left for a follow-up; this PR delivers the protocol primitive that integration will depend on.

## Changes

- `src/codecs/operations.js`: new `OpFlushCodec` and `OpFlushDoneCodec`, both with a single `position: NonNegativeInteger` field, added to `OcapnMessageUnionCodec`.
- `src/captp/pairwise.js`: new `replaceExportValue(slot, newValue)` for in-place value replacement preserving refcount.
- `src/client/ocapn.js`:
  - `op:flush` handler — performs the swap and sends `op:flush-done`.
  - `op:flush-done` handler — resolves the matching pending settler.
  - `flushExport(remoteValue)` exposed via `_debug` for tests.
  - `abort()` rejects all pending and flushed-promise settlers.

## Test plan

- [x] `yarn test test/codecs/operations.test.js` — new snapshot entries for `op:flush` (`<8'op:flush7+>`) and `op:flush-done` (`<13'op:flush-done7+>`) generated via `yarn test:update-snapshots`.
- [x] `yarn test test/flush.test.js` — four end-to-end tests covering: swap-and-ack happy path, double-flush rejected, non-promise reference rejected, untracked value rejected.
- [x] `yarn test` — full ocapn suite (288 tests) passes.
- [x] `yarn lint` — clean.

https://claude.ai/code/session_013hZeUaQrKVRoZQjpd1vcs4

---
_Generated by [Claude Code](https://claude.ai/code/session_013hZeUaQrKVRoZQjpd1vcs4)_


kumavis:
During development we noticed that `getInfoForVal` will put un-exported objects in the export table, which may be unexpected

PR #3252, opened 2026-05-12
ci: add zizmor workflow audit and harden workflows to pass at pedantic persona

kumavis:
## Summary

Introduces [`zizmor`](https://docs.zizmor.sh/) as a CI check that statically audits every workflow under `.github/workflows/**` on each PR and on pushes to `master`, then ratchets the existing workflows up to pass cleanly under the `pedantic` persona at `min-severity: low`. The intent is to convert the workflow-hardening conventions this repo already follows (no `pull_request_target`, no PR-context OIDC, SHA-pinned actions) from aspirational into enforceable, so configuration cannot quietly drift into the patterns that enabled the TanStack May 2026 npm supply-chain compromise ([postmortem](https://tanstack.com/blog/npm-supply-chain-compromise-postmortem)).

The attack chain the audit gate is calibrated against:

1. **`pull_request_target` "Pwn Request"** — untrusted PR code runs with base-repo write tokens.
2. **Cache poisoning across the fork ↔ base trust boundary** — PR-side cache writes consumed by privileged jobs.
3. **In-memory OIDC token extraction (`/proc/<runner>/mem`)** — untrusted code in a job with `id-token: write` exfiltrates the token and publishes to npm directly.

zizmor's `dangerous-triggers`, `excessive-permissions`, `template-injection`, `cache-poisoning`, `unpinned-uses`, `artipacked`, `dependabot-cooldown`, and `stale-action-refs` audits cover the lever points in that chain.

## Configuration choices for the new check

- **Trigger**: `pull_request` (any) and `push` to `master`. The audit is sub-second, so it always runs rather than gating on `paths:`. Always-on makes it eligible to become a required status check via branch protection.
- **Severity / persona**: `persona: pedantic` + `min-severity: low`. Persona surfaces "code smell" findings beyond `regular` (most importantly the workflow-level-vs-job-level scoping of `id-token: write`); the low severity floor future-proofs against any future audits that land at that tier without requiring a follow-up.
- **Output**: inline PR annotations (`annotations: true`). `advanced-security` (SARIF upload to Code Scanning) is left disabled until GitHub Advanced Security is enabled on the repo and `security-events: write` can be granted to this job.
- **Pinning**: `zizmorcore/zizmor-action` is SHA-pinned with a `# v0.5.3` comment, consistent with the repo convention and the `scripts/update-action-pins.mjs --check-pins` enforcement.

## What got hardened on the way to "clean at pedantic"

All landed in this branch; each is its own commit. Grouped by audit class:

### High-severity `excessive-permissions` (the TanStack-class lever)

- `ci.yml`: dropped unnecessary workflow-level `actions: write` (`actions/cache` doesn't need it). `64e75ef2`
- `typedoc-gh-pages.yml`: moved `pages: write` from workflow level to the `deploy` job. `01e1dd14`
- `typedoc-gh-pages.yml`: moved `id-token: write` from workflow level to the `deploy` job. `5284a490`

### Medium `artipacked` (checkout credential persistence)

- 18 read-only `actions/checkout` steps gained `persist-credentials: false`. `bab98d8e`
- The one deliberately-persistent checkout (`release.yml`, which feeds `RELEASE_TOKEN` to `changesets/action`) got an inline `# zizmor: ignore[artipacked]` with a rationale comment. `555fa19d`

### Medium `dependabot-cooldown` and `stale-action-refs` (online audits)

- `dependabot.yml`: 7-day `cooldown.default-days` on both `npm` and `github-actions` ecosystems, matching `update-action-pins.mjs`'s default `--min-age-days=5`. `a77b7e55`
- Four upstream-bumped action tags refreshed to current SHAs (all >= 5 days old): `actions/setup-node@v6`, `dorny/paths-filter@v3`, `changesets/action@v1`, `peter-evans/create-pull-request@v8`. `93f57b34`

### Pedantic-tier scoping and hygiene

- `browser-test.yml`: dropped unnecessary `actions: write`. `e1c01bd8`
- `release.yml`: moved `contents: write` + `pull-requests: write` from workflow level to the `release` job; workflow now defaults to `permissions: {}`. `f7f92fc6`
- `update-action-pins{,-major}.yml`: same — writes scoped to the `update` job, workflow defaults to `permissions: {}`. `1ad1f00f`
- `ci.yml`: routed `${{ matrix.moddable-version }}` through an `env:` var instead of interpolating into `run:` strings (defense-in-depth against `template-injection`). `d90681b9`
- `release.yml`: made `concurrency.cancel-in-progress: false` explicit (mid-release cancellation could leave tags pushed without their matching GitHub release). `15466daa`
- Inline rationale comments on every job-level permission grant so a reviewer can tell at a glance which step consumes each grant. `5b5f441d`

### Gate ratchets

- `min-severity: high` (initial). `73bc9d3e`
- `min-severity: medium`. `0837f45e`
- `min-severity: low`. `0a5cedaf`
- `persona: pedantic`. `cb563931`

## Verification

Local `zizmor --persona pedantic --min-severity low .github/workflows .github/dependabot.yml` reports **zero findings** (9 ignored by the `# zizmor: ignore[...]` directive in `release.yml`, 26 suppressed by zizmor's default rules).

## Suggested follow-ups (not in this PR)

These were on the original list; they're orthogonal enough to belong in their own PRs:

1. **`CODEOWNERS` for `.github/workflows/`** — routes every future workflow edit through a security-aware reviewer regardless of what zizmor says.
2. **Extend `scripts/update-action-pins.mjs --check-pins`** with the anti-pattern assertions (`no pull_request_target`, `no workflow-level id-token: write`, `no ${{ github.event.* }}` in `run:` blocks). Defense-in-depth: if zizmor is ever disabled or its upstream compromised, the home-grown check still fires.
3. **Investigate why `update-action-pins` cron missed the four stale pins** that this PR refreshed manually. They were >= 5 days old at the most recent Monday run, so the age gate isn't the cause — the cron may be silently failing or unable to push the chore branch.
4. **`persona: auditor`** — adds `secrets-outside-env`, which would require wrapping the `release` job in a GitHub deployment Environment with required reviewers. Real behavioral change rather than a configuration tweak; worth its own discussion.
5. **Admin-side, not in-repo**: make the new `zizmor` job a required status check, set default `GITHUB_TOKEN` workflow permissions to read-only, and require approval for all outside-collaborator workflow runs.

## Test plan

- [x] Local `zizmor --persona pedantic --min-severity low` reports zero findings against the final tree.
- [x] All third-party actions remain SHA-pinned; `scripts/update-action-pins.mjs --check-pins` accepts the new pins on `zizmorcore/zizmor-action` and the four refreshed actions.
- [ ] CI on this PR runs the new `zizmor` job and exits zero.
- [ ] Every existing CI job still runs and exits zero (i.e., the permission-scoping moves didn't strand any job below the permissions it actually needs).
- [ ] After merge, the next `update-action-pins` cron run produces no diff (i.e., we already landed everything its weekly bump would have proposed).

https://claude.ai/code/session_013UFBZh9WsjfZtDHiLWFNv9

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,6 +21,8 @@ updates:
         update-types:
           - 'minor'
           - 'patch'
+    cooldown:
+      default-days: 7
 
   - package-ecosystem: github-actions
     directory: /
@@ -28,3 +30,5 @@ updates:
       interval: weekly
     commit-message:
       prefix: chore
+    cooldown:
+      default-days: 7
--- a/.github/workflows/browser-test.yml
+++ b/.github/workflows/browser-test.yml
@@ -9,7 +9,6 @@ on:
 
 permissions:
   contents: read
-  actions: write
 
 # Cancel in-progress runs for the same branch when a new run is started
 concurrency:
--- a/.github/workflows/browser-test.yml
+++ b/.github/workflows/browser-test.yml
@@ -39,7 +39,7 @@ jobs:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
           persist-credentials: false
-      - uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
         with:
           node-version: 18
       - name: Provide yarn # yarn is missing in the Selenium image
--- a/.github/workflows/browser-test.yml
+++ b/.github/workflows/browser-test.yml
@@ -37,6 +37,8 @@ jobs:
       - name: Put unstable chrome where playwright would look for it
         run: mv /opt/google/chrome /opt/google/chrome-unstable
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          persist-credentials: false
       - uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6
         with:
           node-version: 18
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -423,8 +423,10 @@ jobs:
       - name: Choose XS download or build
         if: steps.restore-xs.outputs.cache-hit != 'true'
         id: check-release
+        env:
+          MODDABLE_VERSION: ${{ matrix.moddable-version }}
         run: |
-          if curl -f -s -o /dev/null -I -L https://api.github.com/repos/Moddable-OpenSource/moddable/releases/tags/${{ matrix.moddable-version }}
+          if curl -f -s -o /dev/null -I -L "https://api.github.com/repos/Moddable-OpenSource/moddable/releases/tags/${MODDABLE_VERSION}"
           then
             echo release=download
           else
@@ -433,8 +435,10 @@ jobs:
 
       - name: Download XS
         if: steps.check-release.outputs.release == 'download'
+        env:
+          MODDABLE_VERSION: ${{ matrix.moddable-version }}
         run: |
-          wget "https://github.com/Moddable-OpenSource/moddable/releases/download/${{ matrix.moddable-version }}/xst-lin64.zip"
+          wget "https://github.com/Moddable-OpenSource/moddable/releases/download/${MODDABLE_VERSION}/xst-lin64.zip"
           unzip xst-lin64.zip -d bin
           mkdir -p bin
           chmod 755 bin/xst
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
       - run: corepack enable
 
       - name: Use Node.js
-        uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
         with:
           node-version: 22.x
           cache: yarn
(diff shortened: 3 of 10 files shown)


.github/dependabot.yml
```diff
@@ -21,10 +21,14 @@ updates:
         update-types:
           - 'minor'
           - 'patch'
+    cooldown:
+      default-days: 7
```
kumavis, line 25: team should review this one

kriskowal: This is my thinking in general, though over in the bot garden where chaos reigns, I gave the bots more discretion:
- If there isn’t a known vuln on the new version, wait seven days
- If there is a known vuln on the current version, hazard the new version regardless of how old it is
And it’s free to research vulns on its own.

I think that’s safer than a hard 7 day window, but we’re not set up for that here.

kriskowal (approved):

.github/dependabot.yml
kumavis: Per the [Dependabot cooldown docs](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown) and the [security-updates docs](https://docs.github.com/en/code-security/concepts/supply-chain-security/about-dependabot-security-updates), the two-tier policy you describe is actually what this config already produces:

- Regular version updates respect `cooldown.default-days: 7`, so new releases without a known vuln wait the week.
- Dependabot **security updates** (the PRs created automatically when an installed version has a matching GitHub Advisory Database entry) are a separate flow that ignores cooldown entirely — quoting the docs, *"cooldown only applies to version updates, not security updates."* So if there's a known vuln on the version we currently resolve to, Dependabot opens the bump PR regardless of how fresh the fix is.

That covers (1) and (2) of your list. The "free to research vulns on its own" piece is essentially the Advisory Database driving the security-updates flow; we don't have the more aggressive bot-garden-style scanning you're describing, but for the threat model this PR is about (the Shai-Hulud-class "publish, then immediately get consumed" pattern), the 7-day-with-advisory-bypass behavior already captures the spirit of what you wanted.

If it'd help, I can add a short comment in `dependabot.yml` next to the `cooldown:` block spelling out that security advisories bypass it — that way the next person reading the config doesn't have to remember the docs. Happy to leave it alone if the docs reference is enough.

---
_Generated by [Claude Code](https://claude.ai/code)_

kumavis (commented):
this is human approving the bot work

kumavis:
I'm have merged this, if this becomes a problem for someone, please let me know (ideally on signal)

PR #3255, opened 2026-05-12
feat(eslint-plugin): alias eslint-plugin-import to import-x@4

kriskowal:
## Description

Alias `eslint-plugin-import` to the actively-maintained `eslint-plugin-import-x` soft fork via a Yarn package alias in the `dev` catalog: `eslint-plugin-import: 'npm:eslint-plugin-import-x@4.16.2'`. The package on disk is still named `eslint-plugin-import`, so ESLint continues to register its rules under the existing `import/*` namespace and every `import/*` rule reference (in configs, eslintrc files, and source-file suppression comments) keeps working unchanged. The original `eslint-plugin-import` is unmaintained; import-x is its soft fork with identical rule semantics. The version is exact-pinned in the catalog so a malicious patch release does not reach us through `yarn install` without review.

The import-x implementation ships its own `unrs-resolver` which natively honours the `package.json` `exports` field, so the explicit `import/resolver` settings block previously declared in the `imports` config is no longer required and has been removed.

The seven `package.json` devDependencies that listed `eslint-plugin-import: ^2.31.0` now read `catalog:dev` and pick up the alias. `@endo/eslint-plugin` itself takes a runtime dependency on `eslint-plugin-import` (via the catalog) so downstream consumers extending `plugin:@endo/imports` resolve the plugin automatically.

### Security Considerations

The migration trades the maintainer of `eslint-plugin-import` for the maintainer of `eslint-plugin-import-x`. The latter is a single-maintainer project, post-CVE-2025-54313, on OIDC trusted publishing with SLSA provenance. The residual single-maintainer concentration risk is mitigated by the exact-version pin in the catalog (no caret), the maintainer's post-incident migration to OIDC trusted publishing, and SLSA provenance on the npm artifact (verifiable with `npm audit signatures`).

### Scaling Considerations

Dependency-graph reduction: `eslint-plugin-import-x` reports its own dependency tree at 16 packages versus `eslint-plugin-import`'s 117, almost all of that reduction from replacing `tsconfig-paths` + `typescript` + `resolve` with `unrs-resolver` + `get-tsconfig`. Fewer transitive packages reduce the supply-chain surface per dev-dependency install.

A changeset entry at `@endo/eslint-plugin: minor` records the aliasing for downstream consumers. Because the alias keeps the rule namespace as `import/*`, downstream consumers who reference `import/*` rules through `plugin:@endo/imports`, `plugin:@endo/strict`, or `plugin:@endo/internal` need no change.

### Testing Considerations

`yarn lint` from the root reports 0 errors (2 pre-existing JSDoc warnings, none import-related). `yarn lint:prettier` is clean.

No consumer-visible rule namespace change. The `import/*` rule references that downstream consumers extend continue to resolve. The aliasing is invisible at the ESLint configuration surface.

### Upgrade Considerations

Downstream packages that vendor or extend `@endo/eslint-plugin` configs require no migration. No `import/*` to `import-x/*` rename is needed; the alias preserves the existing namespace.

### DIFF
--- a/.changeset/migrate-eslint-plugin-import-x.md
+++ b/.changeset/migrate-eslint-plugin-import-x.md
@@ -1,0 +1,8 @@
+---
+'@endo/eslint-plugin': minor
+---
+
+Migrate the bundled `@endo/imports` ESLint config off the unmaintained `eslint-plugin-import` and onto the actively-maintained `eslint-plugin-import-x` soft fork.
+This is done via a Yarn package alias (`eslint-plugin-import: 'npm:eslint-plugin-import-x@4.16.2'` in the `dev` catalog), so the package on disk is still named `eslint-plugin-import` and ESLint continues to register its rules under the existing `import/*` namespace.
+The import-x implementation ships its own `unrs-resolver`, which natively honours the `package.json` `exports` field, so the explicit `import/resolver` settings block is no longer required and has been removed.
+Downstream consumers do not need to rename any `import/*` rule references; existing `eslintrc` snippets continue to work.
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -7,7 +7,12 @@ catalogs:
     eslint-config-airbnb-base: ^15.0.0
     eslint-config-prettier: ^9.1.0
     eslint-plugin-eslint-comments: ^3.2.0
-    eslint-plugin-import: ^2.31.0
+    # Aliased to eslint-plugin-import-x so that the package on disk is
+    # named `eslint-plugin-import` (and ESLint registers it under the
+    # `import/*` namespace), while the implementation is the
+    # actively-maintained import-x soft fork. The pinned-exact version
+    # tracks security advisories on import-x.
+    eslint-plugin-import: 'npm:eslint-plugin-import-x@4.16.2'
     tsd: ^0.32.0
     typescript: ~6.0.2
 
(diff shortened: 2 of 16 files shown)


.changeset/migrate-eslint-plugin-import-x.md
```diff
@@ -0,0 +1,8 @@
+---
+'@endo/eslint-plugin': minor
+---
+
+Migrate the bundled `@endo/imports` ESLint config from `eslint-plugin-import` to `eslint-plugin-import-x`.
+The replacement plugin is a soft fork with identical rule semantics under a new namespace prefix (`import-x/*` instead of `import/*`).
+Its bundled `unrs-resolver` natively honours the `package.json` `exports` field.
+Downstream configs that referenced `import/*` rule names through `plugin:@endo/imports`, `plugin:@endo/strict`, or `plugin:@endo/internal` must rename those references to `import-x/*`.
```
turadg: Fortunately, this isn't true. The new can be aliased to the old.

```
// eslint.config.js
import importX from 'eslint-plugin-import-x';

export default [
  {
    plugins: {
      import: importX,
    },

    rules: {
      'import/no-cycle': 'error',
      'import/no-unresolved': 'error',
    },
  },
];
```

That's with flat config. I think that's prevalent now but for older configs the package name could be mapped:
```
{
  "devDependencies": {
    "eslint-plugin-import": "npm:eslint-plugin-import-x@^4"
  }
}
```

Please try the aliasing and confirm it works by reverting the suppression edits in this PR.

turadg (changes requested):

turadg (approved):
🚀

PR #3256, opened 2026-05-13
feat(syrup-frame): add @endo/syrup-frame package and opt-in syrup framing for OCapN TCP-for-testing

kriskowal:
## Description

Adds a new workspace package `@endo/syrup-frame`: a sibling of `@endo/netstring` whose framing omits the trailing `,` separator, so each framed payload on the wire is literally a Syrup byte-string record (`<length>:<payload>`). The exported surface is symmetric with `@endo/netstring`: `makeSyrupsReader` and `makeSyrupsWriter`, including the `chunked` zero-copy writer mode.

The OCapN TCP-for-testing netlayer gains an opt-in `framing` option. The default `'none'` preserves the current wire format used by the `ocapn/ocapn-test-suite` Python `testing_only_tcp` netlayer (a single syrup-encoded record per write, no length prefix). Passing `framing: 'syrups'` wraps each message in the new framing so the transport becomes robust to TCP chunk boundaries that split a single OCapN message, and two peers that both opt in share a single length-prefixed primitive with the syrup payload format itself.

### Security Considerations

The package parses untrusted byte streams. The reader rejects malformed prefixes, oversize length declarations (configurable via `maxMessageLength`, default 999_999_999), missing colon, empty prefix, and trailing characters after a complete frame. No new authority is introduced beyond the `Uint8Array` reader/writer pair the package exports. All exported factories, the inner generator, and the returned writer object are `harden()`-ed. The opt-in TCP netlayer framing does not alter the authority boundary of the netlayer; it changes only how bytes are grouped on the wire.

### Scaling Considerations

`maxMessageLength` bounds reader memory at the message granularity. The chunked writer mode avoids a prefix-plus-payload buffer copy. The netlayer framing wrapper adds one length-prefix per message and no other steady-state cost.

The new package's `README.md` describes the grammar, the asymmetry against `@endo/netstring`, the OCapN TCP-for-testing scope, the API, and the options. Initial-release notes ship in the package changeset. The TCP-netlayer framing option ships with its own changeset describing the opt-in. No deployed-data migration is required.

### Testing Considerations

The new package carries reader, writer, chunked-write, concurrent-write, byte-by-byte fragmentation, trailing-comma rejection, and `writer.throw` forwarding tests, run under both `lockdown` and `noop-harden` configurations. An additional netlayer test exercises a syrups-framed TCP handshake end to end against the JS netlayer.

New package; no backwards-compatibility concerns. `@endo/netstring` is unmodified; existing callers are unaffected. The Endo daemon's `tcp+netstring+json+captp0` transport remains netstring-compliant. The TCP-for-testing netlayer's default framing is unchanged; the new behavior is opt-in.

### Upgrade Considerations

No upgrade action required. Initial release of a new package and an additive, opt-in option on an existing test-only netlayer.

### DIFF
--- a/.changeset/ocapn-tcp-syrup-framing.md
+++ b/.changeset/ocapn-tcp-syrup-framing.md
@@ -1,0 +1,5 @@
+---
+'@endo/ocapn': minor
+---
+
+- Add a `framing` option to `makeTcpNetLayer` (`@endo/ocapn/netlayer/tcp-testing`). The default is `'syrup'`, which wraps each message in the `<length>:<payload>` framing implemented by `@endo/syrup-frame`. This is the framing the OCapN TCP-for-testing netlayer is moving toward (cf. the 2025-12-09 OCapN plenary, https://github.com/ocapn/ocapn/blob/main/meeting-minutes/2025-12-09.md), and is robust to TCP chunk boundaries that split a single OCapN message. Pass `framing: 'none'` to interoperate with the existing `ocapn/ocapn-test-suite` Python `testing_only_tcp` netlayer, which writes a syrup-encoded record with `sendall` and reads one back with `syrup.syrup_read` (no length prefix on the wire). The `'none'` option exists only for that suite's sake and goes away once the suite either adopts syrup framing or is retired.
--- a/.changeset/syrup-initial.md
+++ b/.changeset/syrup-initial.md
@@ -1,0 +1,18 @@
+---
+'@endo/syrup-frame': patch
+---
+
+- New package `@endo/syrup-frame`: a sibling of `@endo/netstring` that
+  drops the trailing `,` separator, so each framed payload on the wire
+  is literally a Syrup byte-string record (`<length>:<payload>`).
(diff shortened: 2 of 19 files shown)


packages/syrups/package.json
```diff
@@ -0,0 +1,83 @@
+{
+  "name": "@endo/syrups",
```
kumavis: name is awkward, id expect generic syrup codecs by that name. "syrup-frame" or something is more specific

packages/syrups/reader.js
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * @param {Iterable<Uint8Array> | AsyncIterable<Uint8Array>} input
+ * @param {object} opts
+ * @param {string} [opts.name]
+ * @param {number} [opts.maxMessageLength]
+ */
+async function* makeSyrupsIterator(
+  input,
+  { name = '<unknown>', maxMessageLength = 999999999 } = {},
```
kumavis: prefer underscore infixes for large number literal.

kumavis (approved):

PR #3258, opened 2026-05-14
chore: enforce general package uniformity across workspace

kriskowal:
## Description

Adds a general package-uniformity check that runs in CI, plus per-package fixes so the check is green on day one. The check uses `packages/skel/` as the single source of truth for every workspace package and fails closed on any drift.

The script (`scripts/check-package-uniformity.sh`, wired into the existing `lint` job in `.github/workflows/ci.yml` as a step named `Check package uniformity`) follows the style of the sibling `scripts/check-packages.sh` and enforces:

- `SECURITY.md` is byte-identical to `packages/skel/SECURITY.md`.
- `LICENSE` matches `packages/skel/LICENSE` modulo the copyright line. The copyright line must match either the skel placeholder (`Copyright [yyyy] [name of copyright owner]`) or the filled form (`Copyright <YYYY> Endo Contributors`), preserving the existing `scripts/set-license-text.sh` convention of stamping the package's creation year into its LICENSE.
- `package.json` field shape: `author`, `license`, `type`, `repository.type`, `repository.url`, `bugs.url` match skel; `repository.directory` is `packages/<dir>`; `name` ends with `/<dir>` (or equals `<dir>` for `ses`); `publishConfig.access` is `public` for non-private packages; `description` is non-empty and not equal to skel's placeholder.

The script ships a small `EXCEPTIONS` allowlist for legitimate historical deviations, with one initial entry: `packages/eslint-plugin` is permitted to ship without a `.type` field because it remains a CommonJS ESLint v8 plugin.

The remaining five commits land the per-package fixes the new check surfaces: aligning three `SECURITY.md` files to the skel template; adding `LICENSE` to five packages that shipped without one (`eslint-plugin`, `far`, `init`, `lockdown`, `stream-types-test`); aligning `repository`/`bugs` fields on four packages and recording the `eslint-plugin` `.type` exception; filling in descriptions for `ocapn` and `ocapn-noise`; and aligning the `.author` label on the four SES-heritage packages (`lockdown`, `module-source`, `ses`, `ses-ava`) from `Agoric` to `Endo contributors`.

The SES-heritage `.author` change is the one substantive metadata change in the series. It is intentionally its own commit so it can be reverted in isolation if the maintainer prefers to retain `Agoric` as the author label on those four packages; the equivalent alternative is to add per-package entries to the `EXCEPTIONS` allowlist in the check script.

### Security Considerations

`SECURITY.md` is the published vulnerability-disclosure entry point; uniform copies remove ambiguity about which contact path is current. The check itself is a read-only bash script that hashes files and queries `package.json` fields via `jq`; it introduces no new authority and runs in CI only.

### Scaling Considerations

The check runs once per CI invocation and scales linearly with the number of packages. Negligible.

The script header enumerates every check and its rationale, so future readers can audit what is enforced without running the script. `SECURITY.md` itself is user-facing documentation; making all copies identical means downstream readers see one consistent disclosure policy regardless of which package's npm page they land on.

### Testing Considerations

Manually exercised the script in three states: all-uniform (exit 0), one `SECURITY.md` hash drifted (exit 1, names the package), one `.description` field nulled (exit 1, names the package). The new CI step exercises it on every PR.

The metadata fields touched (`author`, `repository.url`, `bugs.url`, `description`) surface on npm package pages and in published `package.json` files. The substantive change is the SES-heritage `author` flip from `Agoric` to `Endo contributors` on four packages; everything else is filling in missing or visibly-broken values. The body of `SECURITY.md` and the Apache-2.0 license text itself are unchanged. No API surface is affected.

### Upgrade Considerations

When the canonical `SECURITY.md` or `LICENSE` body needs to change, edit the file in `packages/skel/` and propagate with a one-liner across `packages/*/`; the CI check then enforces the new uniform state on every subsequent PR. Adding a new uniformity rule is a one-block addition in the script. Granting a new legitimate exception is one line in the `EXCEPTIONS` array with a comment naming the package and the reason.

Out of scope: migrating `packages/eslint-plugin` to ESM (named explicitly in the `EXCEPTIONS` allowlist as a follow-up), and aligning the `homepage` field, which currently splits 24 / 24 across two URL patterns and warrants a separate scope decision before the check enforces it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,6 +62,9 @@ jobs:
       - name: Check composite tsconfig files are up to date
         run: yarn build:types:check
 
+      - name: Check package uniformity
+        run: node scripts/check-package-uniformity.mjs
+
       # build the API docs to verify it works
       - name: build API docs
         run: yarn docs
--- a/packages/bundle-source/package.json
+++ b/packages/bundle-source/package.json
@@ -56,7 +56,7 @@
   ],
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/endojs/endo",
+    "url": "git+https://github.com/endojs/endo.git",
     "directory": "packages/bundle-source"
   },
   "author": "Endo contributors",
--- a/packages/bytes/SECURITY.md
+++ b/packages/bytes/SECURITY.md
@@ -21,7 +21,7 @@ SES stands for fearless cooperation, and strong security requires strong collabo
 * A bug reporter can expect acknowledgment of a potential vulnerability reported through  [security@agoric.com](mailto:security@agoric.com)  within one business day of submitting a report. If an acknowledgement of an issue is not received within this time frame, especially during a weekend or holiday period, please reach out again. Any issues reported to the HackerOne program will be acknowledged within the time frames posted on the program page.
 	* The bug triage team and Agoric code maintainers are primarily located in the San Francisco Bay Area with business hours in  [Pacific Time](https://www.timeanddate.com/worldclock/usa/san-francisco) .
 
-* For the safety and security of those who depend on the code, bug reporters should avoid publicly sharing the details of a security bug on Twitter, Discord, Telegram, or in public GitHub issues during the coordination process.
+* For the safety and security of those who depend on the code, bug reporters should avoid publicly sharing the details of a security bug on Twitter, Discord, Telegram, or in public Github issues during the coordination process.
 
 * Once a vulnerability report has been received and triaged:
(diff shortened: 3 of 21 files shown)


boneskull:
@kriskowal 

I haven't touched it a few years, but I created [sync-monorepo-packages](https://www.npmjs.com/package/sync-monorepo-packages) to solve some of these issues.  While it doesn't do the _checking_, it does do the synchronization.

If you think this'd be useful, I'll work some necromancy and cut a new release.

boneskull (approved):
I'm not going to block on it, but I'd prefer any new scripts are written in JS.  We have `zx` at our disposal...

PR #3259, opened 2026-05-14
chore: add composite typescript build configuration

boneskull:
This adds a new TS build configuration, allowing incremental builds of the entirety of Endo's types at once.  This is useful for a development environment where we are linking Endo against another project (e.g. LavaMoat or agoric-sdk).  It eliminates the need to run potentially multiple discrete per-workspace TS builds.

This configuration is entirely _opt-in_ and does not conflict with the current `prepack` scheme for generating declaration files.

## Usage

> If you have leftover `.d.ts` files from running `prepack`, run `yarn clean` from the workspace root first.

From the workspace root, execute:

```sh
yarn build:types
```

This will build the declaration files of all workspaces.  Running `yarn build:types` again will quickly exit, as there's no work to be done.

Use `yarn build:types:watch` for "watch" mode.  The benefit here is that modifying a type in any given workspace will only rebuild the minimum declarations necessary.

## TypeScript Configuration Files

The `tsconfig.composite.json` files in the workspace root and within each workspace are _automatically generated_ by a script, `scripts/generate-composite-tsconfigs.mjs`.  This is because for a composite configuration to work correctly, each workspace must have references to each _other_ workspace it depends upon.  These references typically correspond to the intra-project `dependencies` of any given workspace.  For example, `@endo/hex` depends upon `@endo/harden`.  This means `@endo/hex`'s `tsconfig.composite.json` must contain a reference to `@endo/harden`'s composite configuration:

```json
{
  "references": [{"path": "../harden/tsconfig.composite.json"}]
}
```

In smaller monorepos, maintaining references is easy, but Endo has roughly 50 workspaces with varying degrees of interdependence.  In a perfect world, TypeScript would be able to derive references from `package.json` files (much like Yarn can create a topological graph of workspace dependencies). Alas.

However, the `scripts/generate-composite-tsconfigs.mjs` leverages this functionality of Yarn to output data about the relationships between workspaces. This data does not discern between _`dependencies`_ and _`devDependencies`_, however, which means we still must filter the result; our composite build _does not care_ about `devDependencies`, because those are not involved in the emission of declarations (e.g., `test/` is skipped, and this is the only place `@endo/ses-ava` ever shows up).

The script generates each `tsconfig.composite.json`, containing all necessary references, as well as the root `tsconfig.composite.json`, which contains all workspaces (those which generate declarations, anyway).

This script should be run when adding or removing a `workspace:^` dependency.

> Open to ideas about how to automate this; typically I'd use `husky`/`lint-staged` and just regenerate the configs when a `package.json` is updated.

This script has tests!

## CI

In CI, we assert the composite configs are up-to-date by running `yarn build:types:check`.

> I wouldn't be opposed to omitting this check from CI, since it makes work for people.

## Notes

- The generation script does cycle checking. It probably doesn't need to, since Yarn should complain about that.
- The `tsconfig.json` files in most workspaces needed updating; you'll have noticed this was an issue if you tried to rebuild the declarations in any given project; the `.d.ts` files would be considered part of the build, and `tsc` would refuse to overwrite them. The main reason is that we emit `.d.ts` files alongside the sources (which is a tradeoff).
- I've updated each `tsconfig.json` in each workspace to be specific about which files should be considered sources, which deviates from the "boilerplate" `tsconfig.json` we had previously.  In practice, the only effect of this is that it enables is the rebuilding of types. This may require future manual upkeep of each `tsconfig.json`, but I consider explicitly listing sources to be good hygiene.
- I added a `scripts/.eslintrc.json` because ESLint was not configured properly for `scripts/`.
- I fixed the `clean` script, which now just runs `postpack` in all workspaces. `lerna` does not understand `workspace:^`.
- The `build:types:clean` script is like `clean`, but faster if used in tandem with `build:types`.
- This also seems to fix the `build-ts` script, but I'm unsure if it is still useful.
- Added missing `prepack`/`postpack` scripts in `@endo/trampoline`.


packages/bundle-source/tsconfig.json
```diff
@@ -2,11 +2,10 @@
   "extends": "../../tsconfig.eslint-base.json",
   "include": [
     "*.js",
-    "*.ts",
-    "src",
-    "test"
-  ],
-  "exclude": [
-    "demo/"
```
boneskull, line 10: This looks like an oversight, but it didn't work in practice anyway, since `tsconfig.build.json` contains an `exclude` field which overrides it.

packages/compartment-mapper/tsconfig.json
```diff
@@ -2,13 +2,13 @@
...
     "src/**/*.js",
-    "src/**/*.ts",
+    "src/types/*.ts",
+    "!src/types/*.d.ts",
+    "src/types-external.d.ts",
+    "src/types.d.ts",
     "test",
-  ],
-  "exclude": [
-    "**/*.d.ts"
```
boneskull, line 12: Because `exclude` in `tsconfig.build.json` overrides this, it wasn't working anyway.  Fortunately, we can use negated glob patterns to the same effect.

scripts/generate-composite-tsconfigs.mjs
```diff
@@ -0,0 +1,340 @@
+#!/usr/bin/env node
```
boneskull, line 1: Logic is mostly clanker-generated here, but I did screw with it quite a bit.

scripts/generate-composite-tsconfigs.mjs
```diff
@@ -0,0 +1,340 @@
...
+
+/**
+ * Serialise a config object to a JSON string in the style used by this repo
+ * (2-space indent, trailing newline).
+ *
+ * @param {object} config
+ * @returns {string}
+ */
+export const serialise = config => {
+  return `// DO NOT EDIT! THIS FILE IS AUTO-GENERATED BY scripts/generate-composite-tsconfigs.mjs\n${JSON.stringify(config, null, 2)}\n`;
```
turadg: the exclamation is severe.

```suggestion
  return `// DO NOT EDIT. This file is  auto-generated by scripts/generate-composite-tsconfigs.mjs\n${JSON.stringify(config, null, 2)}\n`;
```

even better, pull that comment into a const for,
```suggestion
  return `${WARNING}\n${JSON.stringify(config, null, 2)}\n`;
```

AGENTS.md
```diff
@@ -85,6 +86,45 @@ Never mix `self` and `facets` in the same context type.
...
+```
+
+**When to regenerate:** run `yarn build:types:gen` after adding, removing, or
+changing the runtime `dependencies`/`peerDependencies`/`optionalDependencies`
+of any workspace.
+
+The generator (`scripts/generate-composite-tsconfigs.mjs`) reads
+`yarn workspaces list` output and each package's `package.json`.
+
+CI will fail if the generated files drift from what the generator would produce.
```
turadg, line 114: good, let's prevent drift

CONTRIBUTING.md
```diff
@@ -82,15 +89,47 @@ guides in the `docs/` directory, and package-specific documentation.
 should use long lines (paragraphs joined without manual wrapping), as GitHub
 uses a different Markdown flavor for those contexts.
 
+## TypeScript declarations
+
+Each package generates `.d.ts` declaration files via `tsc --build
+tsconfig.build.json` during `prepack`.
```
turadg: https://github.com/endojs/endo/pull/3260 changes that so if that lands first this will need new languuge, or if this lands first that will need to revise this file. Robustly,

```suggestion
TypeScript `.d.ts` declaration files are generated as part of publishing.
```

package.json
```diff
@@ -44,7 +44,7 @@
     "zx": "^8.1.8"
   },
   "scripts": {
-    "clean": "lerna clean",
+    "clean": "yarn workspaces foreach --all run postpack",
```
turadg, line 47: I'd like to get away from having build files in the source tree but if this change helps your workflow it seems sufficiently isolated to allow opt-in.

turadg (approved):

kriskowal:
This carries implications for https://github.com/endojs/endo/pull/3258, so let’s land this first and adjust the job that will make sure these files stay correct.

package.json
boneskull: @turadg Unclear if this comment is in reference to the `clean` fix, but this particular change is tangential; `lerna clean` was fully broken due to its lack of support for `workspace:*` versions.

scripts/generate-composite-tsconfigs.mjs
boneskull: I prefer to keep it severe, but I'll pull it out into a variable.

PR #3260, opened 2026-05-14
Pack workspaces via ts-node-pack, out of the source tree

turadg:
Refs: https://github.com/endojs/endo/pull/3137#discussion_r3018302496

## Description

Replaces the in-place `prepack`/`postpack` mutation pipeline with a tarball-native packing flow built on [ts-node-pack][tnp]. Driven from the repo root by `yarn pack:all` (which writes every public workspace's `.tgz` into `dist/`) and `yarn release:npm` (which wraps that and `npm publish`-es each tarball, replacing `lerna publish from-package`).

[tnp]: https://github.com/turadg/ts-node-pack

The most critical files to review:

- `scripts/pack-all.mjs`, `scripts/release-npm.mjs` — the new publish entry points.
- `scripts/smoketest-publishing.sh` — CI's end-to-end check: spins up Verdaccio, runs `release:npm` against it, installs a representative subset of the published tarballs into a fresh consumer and exercises them under SES lockdown. This replaces the previous `yarn workspaces foreach exec npm pack` step (which could not survive the refactor) and also subsumes the "Prepack packages" type-resolution check.
- `packages/*/package.json` — 41 packages drop their `prepack`/`postpack` hooks (see "Simplification to package.jsons" below).
- Three latent bugs are fixed in their own commits: `bundle-source` (dangling `./exported.js` export pointing at a file that has never existed in any published tarball), `module-source` and `ses` (both with `xs` conditional exports pointing into `src-xs/`, which was missing from the `files` list — see "Value of ts-node-pack's validation" below).

### Why building outside the source tree

The legacy flow mutated the source tree during `npm pack`:

1. `prepack`: `git clean -fX && tsc --build tsconfig.build.json` — emit `.d.ts` files alongside the `.ts` / `.js` sources so `npm pack` can scoop them up.
2. `npm pack` — zip the (now-decorated) source tree.
3. `postpack`: `git clean -fX` — wipe the generated artifacts back out.

This works, but it has real costs:

- **The working copy is not safe to edit during a pack.** A `git status` mid-pack is unrecognizable; an interrupted pack leaves the tree dirty; tools watching the filesystem (IDEs, file watchers, ESLint daemons) react to artifacts that will exist for seconds.
- **`git clean -fX` is a load-bearing footgun.** It removes everything in `.gitignore` to keep the build deterministic. That's correct for CI but can delete an unsuspecting developer's local notes, cached test data, or scratch files. We've inherited bug reports about this from agoric-sdk.
- **It's stateful and order-dependent.** The `postpack` *has* to run to clean up, which means failures mid-pack require manual recovery. `npm pack` doesn't strictly guarantee `postpack` runs on all failure paths.
- **It hides `npm pack`'s "what would be published" answer.** You can only inspect the contents of the tarball by either letting the lifecycle run to completion or untarring the result.

ts-node-pack copies the `npm-packlist` enumeration into a temp staging directory, runs `tsc` (and type-stripping) *there*, and runs `npm pack` *there*. The source working copy is never touched. Failures leave the source tree exactly as it was. `yarn pack:all` is idempotent and side-effect-free outside of `dist/`.

### Simplification to package.jsons

41 packages drop this scripts pair:

```json
"prepack": "git clean -fX -e node_modules/ && tsc --build tsconfig.build.json",
"postpack": "git clean -fX -e node_modules/",
```

Three close variants are also dropped: `errors` (`postpack: yarn clean`), `harden` (prepack without the `git clean` prefix, postpack scoped to `*.d.ts*`/`*.tsbuildinfo`), and the bare-postpack-only forms in `eslint-plugin` and `lockdown`. `ses` keeps its postpack — that one cleans the rollup-built `dist/` produced by its own `prepare` script, unrelated to the tsc declaration flow.

Net: 82 lines removed from package manifests, and every public workspace now has the same packing story (none) rather than a per-package incantation. New packages don't need to wire anything up — `pack-all.mjs` enumerates all public workspaces from `yarn workspaces list` and drives ts-node-pack uniformly.

### Value of the validation ts-node-pack does

Stock `npm pack` doesn't validate that `main`/`module`/`types`/`exports` actually point at files in the tarball. ts-node-pack does, strictly, after rewriting the package.json. Adding it to the publish path immediately surfaced three latent registry bugs that have been quietly shipping for years:

- **`@endo/bundle-source`** declared `"./exported.js": "./exported.js"` since 2022. No `exported.js` has ever existed at that path — not in the source tree, not in the published `@endo/bundle-source@4.3.0` tarball on npm (verified by pulling it). The old prepack pipeline didn't generate it either; it was dangling from day one. Consumers importing `@endo/bundle-source/exported.js` would have hit `ERR_MODULE_NOT_FOUND`, which is presumably why no consumer in this monorepo uses it. The fix wires the public subpath to `./src/exports.js`, which does exist and which `src/exports.js`'s top comment explicitly describes as "a dummy to use exports.d.ts and satisfy runtime imports".
- **`@endo/module-source`** and **`ses`** both declared `xs` conditional exports pointing into `src-xs/`, but `src-xs` was not in either package's `files` list. Both currently-published tarballs on npm have zero `src-xs/*` files. XS-engine consumers (agoric-sdk's blockchain VM substrate) resolving the `xs` condition would have hit `ERR_MODULE_NOT_FOUND`. The fix is one line per package — add `src-xs` to `files`.

These bugs have been latent because the V8/SpiderMonkey/Hermes resolution path never picks the `xs` condition, and nobody in the monorepo imports the broken `./exported.js` subpath. ts-node-pack's validator is exactly the kind of guardrail that makes that class of bug impossible to ship again.

### Security Considerations

No change to the trust boundary of any package. The publish flow now runs `npm publish` per tarball under the same npm auth context as `lerna publish from-package` did. The smoketest uses a disposable Verdaccio with no uplink — by construction it cannot leak a package to the real registry. `HOME` is rerooted to a per-run tmpdir so the smoketest cannot stomp on the developer's `~/.npmrc` or leave auth tokens behind.

### Scaling Considerations

`pack-all.mjs` packs workspaces sequentially. CPU cost per workspace is dominated by `tsc` declaration emit, which is unchanged from the legacy flow. Wall-clock parity in local testing. No on-chain impact.

`docs/typescript.md` rewritten to describe the ts-node-pack flow rather than the deleted `scripts/packing/` pipeline.

`CONTRIBUTING.md` local-QA section updated: the previous recipes (`yarn workspaces foreach exec npm pack` and `yarn lerna run prepack`) depended on the per-package prepack hooks and are now replaced by `yarn pack:all` and `yarn smoketest:publish`, which exercise the same checks against the new flow.

Downstream consumers see no change in published tarball contents *for packages that were not affected by the three latent bugs*. The three fixed packages will start shipping files that previously dangled — for `bundle-source/exported.js`, the subpath becomes resolvable for the first time; for `module-source` and `ses`, the `xs` condition becomes resolvable for the first time. None of this is a breaking change; it's a strict improvement.

### Testing Considerations

CI gains `smoketest-publish`, a full end-to-end check that:

1. Boots Verdaccio with no uplink on a free port.
2. Runs `yarn release:npm` against it (which calls `pack:all` and `npm publish` per tarball).
3. Installs a representative subset (`ses`, `@endo/eventual-send`, `@endo/marshal`, `@endo/captp`, `@endo/bundle-source`) into a throwaway consumer.
4. Imports them under SES `lockdown()` and exercises a few runtime paths.

This is strictly stronger than the previous `yarn workspaces foreach exec npm pack` check, because it proves `workspace:` specifier resolution and cross-package type/runtime resolution work *from the published tarball state*, not from whatever source tree happens to be on disk.

`yarn release:npm` now invokes `scripts/release-npm.mjs` instead of `lerna publish from-package`. Behavior is equivalent: it `npm publish`-es each tarball in `dist/` and honors `--tag`. No change to release semantics.

Plain per-package `yarn pack` / `npm pack` will no longer emit declarations (the `prepack` hooks that did so are gone). This is a behavior change for anyone who invoked `npm pack` directly on a single package — they should use `yarn pack:all` instead, which produces equivalent tarballs through the centralized flow. Nothing in CI relied on the per-package form.

### Upgrade Considerations

No live-system upgrade implications. This affects the publish path, not the runtime artifact.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mhofman:
Haven't started reviewing yet, but besides the fixes, how can we verify whether the published packages have no differences?

If there are no large expected differences, I would like to see a diff of the packages before and after.

yarn.lock
```diff
@@ -11765,6 +11799,18 @@ __metadata:
   languageName: node
   linkType: hard
 
+"ts-node-pack@npm:^0.3.1":
+  version: 0.3.1
+  resolution: "ts-node-pack@npm:0.3.1"
+  dependencies:
+    amaro: "npm:^0.3.2"
+    npm-packlist: "npm:^10.0.4"
```
mhofman: I suspect `ts-node-pack` should declare a peer dependency for `tsc` to be compatible with more strict package managers.

https://github.com/turadg/ts-node-pack/issues/1

scripts/files.sh
```diff
@@ -5,9 +5,15 @@
...
-  # shellcheck disable=SC2097,SC2098,SC2016
-  PKG="$PKG" "$TAR" xf "$PKG/package.tgz" --to-command='echo "$PKG/${TAR_FILENAME#package/}"'
-done < <(npm query '.workspace:not([private])' | jq -r '.[].location') |
-sort
+REPO_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
+DIST="$REPO_ROOT/dist"
+
+# Build every workspace's tarball into dist/ via pack-all (which uses
+# ts-node-pack for .ts packages and yarn pack for the rest).
+yarn pack:all 1>&2
```
mhofman: Nit: this comment re-iterates what the commit's message says:
>   - scripts/pack-all.mjs builds every public workspace into dist/, using
>     ts-node-pack for packages with .ts sources and 'yarn pack' for the
>     pure JS+JSDoc majority.

Clearly `pack-all.mjs` is only using `ts-node-pack` for every package (which is what I'd expect anyway since we do need to build d.ts for JSDoc too).

I suspect an earlier version had that behavior ?

scripts/release-npm.mjs
```diff
@@ -0,0 +1,65 @@
...
+  throw new Error(`pack-all did not produce ${distDir}`);
+}
+const tarballs = readdirSync(distDir)
+  .filter(name => name.endsWith('.tgz'))
+  .map(name => path.join(distDir, name))
+  .sort();
+
+if (tarballs.length === 0) {
+  throw new Error('release:npm: no tarballs to publish');
+}
```
mhofman, line 62: Should we check that the number of tarballs match the number of packages to be published?

scripts/release-npm.mjs
```diff
@@ -0,0 +1,65 @@
...
+  });
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const repoRoot = path.resolve(__dirname, '..');
+const distDir = path.join(repoRoot, 'dist');
+
+const tagFlag = process.argv.indexOf('--tag');
+const tag = tagFlag >= 0 ? process.argv[tagFlag + 1] : process.env.npm_config_tag;
+
+// Always re-pack so the tarballs match HEAD. pack-all.mjs cleans dist/ first.
```
mhofman: I don't see any explicit logic in pack-all.mjs to clean dist/. I suppose `ts-node-pack` does it, but maybe `pack-all.mjs` should explicitly do it too?

mhofman (commented):
I like the idea of leveraging a new external tool to build packages from source following our type support approach.

I have some likely unfounded concerns that we may end up publishing leftover or partially built output from `dist/`.

Regarding devex, I'm not convinced excluding the source `.ts` is the right approach. Is there really a problem with build tools getting confused with `.ts` source alongside `.js` of the same name? At runtime, an attached debugger would not be able to show the ts source, which is usually more readable than the js with blank space from stripped types. See https://github.com/turadg/ts-node-pack/issues/3

yarn.lock
turadg: I'll look into that. Not a blocker.

scripts/files.sh
turadg: Good catch. Revised.

turadg: Good catch — that comment was stale from an earlier draft where pack-all did branch between ts-node-pack and yarn pack. The current code unconditionally drives ts-node-pack for every workspace (which, as you note, is what we need anyway because every package emits .d.ts from JSDoc). Removed the misleading line.

scripts/release-npm.mjs
turadg: Good safety. Added: release-npm.mjs now re-runs `yarn workspaces list --json --no-private` and refuses to publish if the tarball count doesn't equal the public-workspace count. Catches an interrupted pack-all, a workspace flipped to private between pack and publish, or a stray manual deletion from dist/ — all of which would silently under-publish today.

scripts/release-npm.mjs
turadg: It is explicit — pack-all.mjs line ~46: `if (existsSync(distDir)) rmSync(distDir, { recursive: true, force: true });` runs before any tarball is written. I also added a header comment in this push spelling out the freshness guarantee so it's not hiding.

scripts/release-npm.mjs
turadg: Good idea. Done.

scripts/release-npm.mjs
turadg: Can't hurt. Done.

mhofman (approved):
Assuming the output of `compare-pack` is expected (I cannot run locally right now), this LGTM.

As we discussed we can postpone the discussion on including original `.ts` source file, and tweaks to the `ts-node-pack` pakcage.

turadg:
> Assuming the output of `compare-pack` is expected

Thanks for the poke on that.  One real issue surfaced and is fixed: ts-node-pack <0.3.2 was leaking absolute publisher paths into .d.ts.map sources. Bumped to ^0.3.2 in this branch.

Other differences:
  - CHANGELOG.md dropped from all 41 packages — intentional under the ts-node-pack files policy.
  - endo-compartment-mapper: src/types/*.ts → src/types/*.js — type-strip, as expected.
  - ses: +15 files (dist/* bundles, src-xs/*) — these are shipped correctly now; the legacy yarn pack flow was missing them because ses relies on prepare + postpack: git clean -fX, which yarn pack doesn't fully exercise. The new flow is more faithful to what npm consumers actually need.

PR #3263, opened 2026-05-15
chore(eslint-plugin): require underscore-delimited groups in numeric literals

kriskowal:
## Description

Adds `eslint-plugin-unicorn`'s `numeric-separators-style` rule to the `@endo/internal` ESLint preset, enforcing underscore-delimited groupings on numeric literals. Decimal numbers of five or more digits group every three digits (`1_234_567`); hexadecimal, binary, and octal literals group by the rule's conventional widths (hex in twos, binary and octal in fours). The rule is autofixable, so `eslint --fix` applied the migration across the codebase in a separate commit. A final commit reflows seven autofix-touched files that Prettier wanted to re-wrap once the separators pushed a few expressions past the print-width threshold.

### Security Considerations

None. Numeric separators are a syntactic transformation that does not change the runtime value of any literal.

### Scaling Considerations

None.

The `@endo/eslint-plugin` changeset notes that consumers of `plugin:@endo/internal` will see lint errors on numeric literals that violate the rule and must add `eslint-plugin-unicorn` to their devDependencies to satisfy the new peer dependency. `eslint --fix` rewrites violations automatically, so the migration cost downstream is one autofix run per consumer.

### Testing Considerations

`yarn lint` is the test surface. The pre-PR run on this branch reports zero ESLint errors and zero Prettier warnings on the files this migration touches. Two pre-existing ESLint warnings on `master` remain unaddressed (in `evasive-transform/src/index.js` and `ses/src/compartment.js`); they predate this change and are out of scope here.

The rule strips separators from decimal literals below the minimum-digits threshold (`1_000` becomes `1000`). This matches the rule's defaults and is consistent with the convention that separators communicate magnitude only when the number is long enough to benefit from grouping.

### Upgrade Considerations

None for production systems. Downstream consumers of `@endo/eslint-plugin` will need to install `eslint-plugin-unicorn@^56.0.1` (a new peer dependency of the plugin); the published changeset describes the requirement.


packages/bundle-source/cache.js
```diff
@@ -448,7 +448,7 @@ export const makeNodeBundleCache = async (
   ]);
 
   if (nonce === undefined) {
-    nonce = crypto.randomInt(0xffff_ffff);
+    nonce = crypto.randomInt(0xff_ff_ff_ff);
```
turadg: I find the pairs harder to read that quadruplets. Consider groups of four.

packages/cjs-module-analyzer/index.js
```diff
@@ -1289,7 +1289,7 @@ function isIdentifierChar(code, astral) {
   if (code < 91) return true;
   if (code < 97) return code === 95;
   if (code < 123) return true;
-  if (code <= 0xffff)
+  if (code <= 0xff_ff)
```
turadg: Another case in point of four-segment reading better

packages/cli/src/commands/log.js
```diff
@@ -25,7 +25,7 @@ const delay = async (ms, cancelled) => {
 export const log = async ({ follow, ping }) =>
   withInterrupt(async ({ cancelled }) => {
     await null;
-    const logCheckIntervalMs = ping !== undefined ? Number(ping) : 5_000;
+    const logCheckIntervalMs = ping !== undefined ? Number(ping) : 5000;
```
turadg, line 28: Why prefer no underscore for thousands?

packages/cli/test/number-parse.test.js
```diff
@@ -4,10 +4,10 @@ import { parseBigint } from '../src/number-parse.js';
 test('returns a bigint when input is a valid integer string', t => {
   t.is(parseBigint('42'), 42n);
   t.is(parseBigint('0'), 0n);
-  t.is(parseBigint('   9007199254740993   '), 9007199254740993n);
+  t.is(parseBigint('   9007199254740993   '), 9_007_199_254_740_993n);
   t.is(
     parseBigint('123456789012345678901234567890'),
-    123456789012345678901234567890n,
+    123_456_789_012_345_678_901_234_567_890n,
```
turadg: This is worse. The reader wants to verify the numeral sequence matches the above. Consider suppressing the rule on this file

packages/hex/test/decode.bench.js
```diff
@@ -188,7 +188,9 @@ const pairMapTableDecode = (string, name = '<unknown>') => {
 const tableDecode = arrayTableDecode;
 
 // Deterministic PRNG, same seed shape as other Endo fuzz tests.
-const defaultSeed = [0xb0b5c0ff, 0xeefacade, 0xb0b5c0ff, 0xeefacade];
+const defaultSeed = [
+  0xb0_b5_c0_ff, 0xee_fa_ca_de, 0xb0_b5_c0_ff, 0xee_fa_ca_de,
```
turadg: These visually break the mnemonic. Please suppress this line or file

packages/hex/test/encode.bench.js
```diff
@@ -163,7 +163,9 @@ const nativeToHex =
   typeof toHex === 'function' ? /** @type {() => string} */ (toHex) : undefined;
 
 // Deterministic PRNG, same seed shape as other Endo fuzz tests.
-const defaultSeed = [0xb0b5c0ff, 0xeefacade, 0xb0b5c0ff, 0xeefacade];
+const defaultSeed = [
+  0xb0_b5_c0_ff, 0xee_fa_ca_de, 0xb0_b5_c0_ff, 0xee_fa_ca_de,
```
turadg: Ditto re mnemonic

packages/marshal/src/encodePassable.js
```diff
@@ -146,9 +146,9 @@ const decodeBinary64 = (encoded, skip = 0) => {
   encoded.charAt(skip) === 'f' || Fail`Encoded number expected: ${encoded}`;
   let bits = BigInt(`0x${getSuffix(encoded, skip + 1)}`);
   if (encoded.charAt(skip + 1) < '8') {
-    bits ^= 0xffffffffffffffffn;
+    bits ^= 0xff_ff_ff_ff_ff_ff_ff_ffn;
```
turadg: What's salient here is the length and this is harder to count than groups of four.

packages/marshal/tools/marshal-test-data.js
```diff
@@ -115,7 +115,7 @@ export const roundTripPairs = harden([
   [-Infinity, { '@qclass': '-Infinity' }],
   [4n, { '@qclass': 'bigint', digits: '4' }],
   // Does not fit into a number
-  [9007199254740993n, { '@qclass': 'bigint', digits: '9007199254740993' }],
+  [9_007_199_254_740_993n, { '@qclass': 'bigint', digits: '9007199254740993' }],
```
turadg, line 120: Ditto the comparison point

packages/ocapn/src/syrup/encode.js
```diff
@@ -10,8 +10,8 @@ const defaultCapacity = 256;
 
 // For strings, U+D800–U+DFFF are invalid.
 // See https://github.com/ocapn/ocapn/blob/main/draft-specifications/Model.md#string
-const INVALID_STRING_CHARS_START = 0xd800;
-const INVALID_STRING_CHARS_END = 0xdfff;
+const INVALID_STRING_CHARS_START = 0xd8_00;
```
turadg: Another case for fours: the comment above is in fours. The literal should match

turadg (commented):
I'm not opposed to the rule but I think the config needs adjusting. Consider omitting the rule for test files and removing the hex requirement

packages/eslint-plugin/lib/configs/internal.js
```diff
@@ -39,10 +39,20 @@ dynamicConfig.overrides.push({
...
     'import/no-unresolved': ['error', { ignore: ['ava'] }],
+    'unicorn/numeric-separators-style': [
+      'error',
+      {
+        onlyIfContainsSeparator: false,
+        number: { minimumDigits: 5, groupLength: 3 },
+        binary: { minimumDigits: 0, groupLength: 4 },
+        octal: { minimumDigits: 0, groupLength: 4 },
+        hexadecimal: { minimumDigits: 0, groupLength: 4 },
+      },
```
kriskowal, line 54: @turadg The bulk of your feedback is addressed here, and with some one-off exceptions where the rules are disabled to make other representations more obviously equivalent.

packages/ses/src/compartment.js
```diff
@@ -337,7 +337,10 @@ export const makeCompartmentConstructor = (
   /**
    * @param {...CompartmentOptionsArgs|LegacyCompartmentOptionsArgs} args
    */
-  /** @this {Compartment} */
+  /**
+   * @param {...any} args
```
turadg: driveby but no objection

turadg (approved):

packages/ses/src/compartment.js
kriskowal: Thank you for catching this.

PR #3226, opened 2026-04-30
feat(marshal,pass-style): admit immutable ArrayBuffer through codecs

kriskowal:
## Description

Admits the immutable `ArrayBuffer` (passStyle `'byteArray'`) through the
capdata, smallcaps, and encode-passable codecs in `@endo/marshal`, and
adds the supporting hex/byteArray helpers in `@endo/pass-style`.

- `@endo/pass-style` gains `byteArrayToHex`, `hexToByteArray`,
  `byteArrayToUint8Array`, and `uint8ArrayToByteArray`, all routed
  through `@endo/hex`.
- `@endo/marshal`:
  - **capdata**: `byteArray` encodes as
    `{"@qclass":"byteArray","data":"<hex>"}`.
  - **smallcaps**: `byteArray` encodes as `"*<hex>"`; the `*` prefix is
    now reserved for byteArray.
  - **encode-passable**: `byteArray` encodes as
    `a<encodeBigInt(byteLength)>:<hex>`. The Elias-delta length prefix
    gives shortlex ordering matching `compareRank` with no arbitrary
    size cap, and every body character is safe inside both
    `legacyOrdered` and `compactOrdered` array framings.
  - **marshal-justin**: renders `byteArray` as `hexToByteArray("<hex>")`.

Syrup already supported this value through its bytestring primitive; CBOR
remains out of scope until `packages/ocapn/src/cbor` lands.

Most critical files to review:
- `packages/marshal/src/encodePassable.js` (ordering + framing safety)
- `packages/marshal/src/encodeToSmallcaps.js` (new reserved prefix)
- `packages/marshal/src/encodeToCapData.js` (new `@qclass`)
- `packages/pass-style/src/byteArray.js` (helper surface and validation)

### Security Considerations

The new encodings introduce new vocabulary that decoders must validate:
the smallcaps `*` prefix, the capdata `@qclass: "byteArray"`, and the
encode-passable `a` prefix. Decoders reject malformed hex and
length-mismatched payloads, and the resulting `ArrayBuffer` is hardened
(immutable) before exposure, preserving the pass-style invariant that
peers cannot mutate shared bytes. No new authority is introduced; this
admits existing immutable-`ArrayBuffer` values through codecs that
previously rejected them.

### Scaling Considerations

Hex encoding doubles payload size relative to raw bytes for capdata and
smallcaps. The encode-passable form uses an Elias-delta-encoded
`byteLength` prefix, so there is no arbitrary size cap and the cost is
[...]

Issue #2883, opened 2025-07-11
When to use `compareRank` vs `compareRankRemotablesTied` from #2871


erights:
## What is the Problem Being Solved?

#2871 introduces the name `compareRankRemotablesTied` as the name for what used to be default semantics for `compareRank` we had before https://github.com/endojs/endo/pull/2597 . I remember looking at #2597 at the time and thinking it was ok for the short circuiting behavior to be the new default. We do need the short circuiting option, I no longer think it should have been the default. If we didn't have the compat worry, I would restore the `compareRankRemotablesTied` as the default behavior and call the #2597 behavior something else.

See https://github.com/endojs/endo/pull/2871/files#r2201446324 for more.

Now that we have both options, we need to figure out which calls to `compareRank` should really be calls to `compareRankRemotablesTied`

PR #3265, opened 2026-05-15
refactor(marshal): compareRankRemotablesTied for rank-cover ops

kriskowal:
Refs: #2871
Refs: #2883
Refs: #3226

## Description

Splits the rank-comparison refactor out from #2871 so that the codec-admission core of that PR can land in #3226 without depending on this orthogonal change.

Introduces `compareRankRemotablesTied` (and its anti-rank twin), which behaves like `compareRank` but treats all remotables as tied for the same rank rather than short-circuiting on the first remotable encountered. Makes this the default `compare` for the rank-cover operations:

- `isRankSorted`, `assertRankSorted`, `sortByRank`, `getIndexCover` (commit 1)
- `unionRankCovers`, `intersectRankCovers` (commit 2)

`compare` becomes optional and is moved to the trailing position on each. `getIndexCover`'s argument order shifts from `(sorted, compare, rankCover)` to `(sorted, rankCover, compare?)`; the only callers are within the Endo monorepo and are updated. `unionRankCovers` and `intersectRankCovers` likewise shift from `(compare, covers)` to `(covers, compare?)`, and `@endo/patterns`'s `patternMatchers.js` (the only caller) is updated. `compareRankRemotablesTied` and `compareAntiRankRemotablesTied` are re-exported from `@endo/marshal`'s `index.js`.

The first commit is salvaged from #2871 with original authorship preserved (Mark S. Miller).

### Security Considerations

No change to trust boundaries, no new authorities. The comparator change is internal to marshal's rank ordering; callers that pass an explicit `compare` keep their existing behavior.

### Scaling Considerations

The tied comparator visits every remotable rather than short-circuiting on the first, so rank-cover operations over inputs containing many remotables do more work per call than the prior `compareRank` default. This matches the intent of the operations (they want a stable cover, not an early exit), and the prior behavior is still available by passing `compareRank` explicitly.

The argument-order changes on `getIndexCover`, `unionRankCovers`, and `intersectRankCovers` are caller-visible. No internal callers remain on the old signatures. External callers, if any, would need to swap argument order; this is captured in the changeset.

### Testing Considerations

`packages/marshal/test/rankOrder.test.js` is extended to:

- Exercise the new optional/default-arg behavior of all six rank-cover operations.
- Resolve two pre-existing TODOs that asked for the `totally orders ranks` and `is transitive` property tests to run against both `compareRank` and `compareRankRemotablesTied`. Both are total preorders, so both properties hold for both, and we now parameterize the tests to exercise each.

Argument-order shifts on three exported functions (`getIndexCover`, `unionRankCovers`, `intersectRankCovers`) and one new default comparator on six. Captured in a changeset. The `compareRankRemotablesTied` and `compareAntiRankRemotablesTied` exports are additions; nothing is removed.

### Upgrade Considerations

No live-system upgrade concerns. The changeset covers the package-level migration for external consumers.


erights (approved):
Assuming it is the same as https://github.com/endojs/endo-but-for-bots/pull/73 which I already reviewed, LGTM. Thanks!

PR #3266, opened 2026-05-15
Claude container

kumavis:
experiment: qemu claude container backed by endo fs object


kumavis:
replaced by https://github.com/endojs/endo-but-for-bots/pull/325

Issue #13, opened 2020-02-13
rename this repo to `ses-shim` ?


warner:
We're planning to retire https://github.com/Agoric/SES , in favor of this repo. Obviously we need to remove the `-beta` from the name. @erights @dtribble any objections to naming it `https://github.com/Agoric/ses-shim` ?

(note that we've used that name for this repo in the past, before renaming it to SES-beta, so if you visit that URL, you'll be redirected back here)

erights:
Once it is ready for others to use for production, yes, I think it should be renamed `ses-shim`. Since we're already using the name `SES-beta` to signal unreadiness, I think we should not rename until then.

warner:
Ah, good point. To unblock the swingset-on-new-SES effort (https://github.com/Agoric/agoric-sdk/issues/477), we need to publish new SES to NPM, but that doesn't mean the repository itself has to be named anything other than `SES-beta`.

The NPM release will probaby be named `SES-0.7.0`. The last old-SES release (made just today) was `SES-0.6.5`. We can reserve the 0.6.x numberspace for old-SES, and deliver 0.7.x or 0.x (x >= 7) until we're ready to declare it 1.0.

jfparadis:
Great.
1. I like `ses-shim`: it is the SES repo as in `git clone <something>ses-shim.git`
2. I like the `npm` package `ses`. It is also the `ses` path in the monorepo `./packages/ses`
That was the intention to avoid confusion between the two.

Let's do this for version:
- This SES is too big of a change to be called a minor release and `0.8.0` doesn't cut it. It's a complete API change, and a minor just doesn't capture that. We need to abide to the semantic:
> Changes that break backward compatibility: Major release
https://docs.npmjs.com/about-semantic-versioning
- Remove `-beta` from the git repo and add it to the version: `1.0.0-beta.0`. The semantic of npm versioning allow for a `prerelease`. The .0 indicates the first beta. When we have a new beta, we increment the .0 to .1 and so on.
https://docs.npmjs.com/cli/dist-tag
- The readme will capture the npm version and display it (it is configures to show the current version)

This is a very good read:
https://medium.com/@kevinkreuzer/publishing-a-beta-or-alpha-version-to-npm-46035b630dd7

michaelfig:
Maybe we can keep the `-beta` in the git repo name for now, but still use something like `2.0.0-beta.0` for the NPM version?

dtribble:
[...]

PR #3274, opened 2026-05-21
fix(eslint-plugin): harden-exports handles destructuring patterns (#2390)

### LINKED ISSUE #2390: eslint-plugin harden-exports rule mishandles many `export`s
_(Originally posted by @gibson042 in https://github.com/endojs/endo/pull/2387#discussion_r1695807510 )_

Non-comprehensive list:
* `export const { propName: exportName } = objWithPropName` (_**`prop.key.name`** of `propName: exportName` pulls propName, whereas we should pull **`prop.value.name`** to  ensure the presence of `harden(exportName)`_) and
* `export const { name, ...rest } = obj` (object rest) and
* `export const { name, parent: { subName } }` (nested patterns) and
* `export const [ exportName1, … ] = array` (_`declaration.id.type === 'ArrayPattern'` is just skipped entirely_)

The linked comment proposes a fix in the form of a `pushDeclaredNames` helper that handles default assignment and arbitrarily nested destructuring patterns, to be used in combination with post hoc verification of the style already employed for "_should be followed by a call to 'harden'_" complaints (i.e., collect all declarations in one phase and validate their conformance in a second phase).

kriskowal:
Closes: #2390

## Description

The `harden-exports` ESLint rule now handles every binding-pattern shape that can appear on the left-hand side of an `export const … = …` declaration. The previous implementation mishandled several shapes: `export const { propName: exportName } = obj` pulled `propName` (the source property) instead of `exportName` (the binding); `export const { name, ...rest } = obj` ignored the rest binding; `export const { name, parent: { subName } } = obj` ignored nested patterns; and `export const [ a, b ] = arr` was skipped entirely because the rule only branched on `ObjectPattern`.

A recursive `pushDeclaredNames(pattern, names)` helper now walks every binding-pattern shape: `Identifier`; `ObjectPattern` shorthand and alias properties (recursing into `prop.value`, not `prop.key`); object rest (`{ a, ...rest }`); `ArrayPattern` elements including sparse holes; array rest (`[ head, ...tail ]`); `AssignmentPattern` defaults (`[ a = 1 ]`, `{ a: b = 1 }`); and arbitrary nesting of all of the above. The default branch of the helper returns a recognized flag and surfaces an `unknownBindingPattern` report so future ECMAScript binding patterns do not silently slip past the rule. Report messages moved into `meta.messages` with `messageId`/`data` to match sibling-rule conventions.

### Security Considerations

No new authority. The rule's surface is lint-time analysis of source files in the repos that adopt the rule; it neither grants nor restricts runtime capability.

### Scaling Considerations

The recursive helper visits each binding node once. Asymptotic cost is unchanged from the prior ad-hoc traversal; nested patterns that the prior code skipped now incur their own constant per-node cost.

No documentation surface changes. A `patch` changeset accompanies the change.

### Testing Considerations

Every binding shape listed above is exercised by paired `valid` (with the proper `harden(...)` call) and `invalid` fixtures in `packages/eslint-plugin/test/harden-exports.test.js`, including a TypeScript-annotated `{ name, ...rest }: { name: string; [k: string]: unknown } = obj` case under both testers. Coverage was verified per the contributor checklist by temporarily removing the `RestElement` branch inside `ObjectPattern`; the rest-binding fixtures failed as expected before the handling was restored.

The rule will now flag previously-uncaught `export const` destructuring shapes that lack a matching `harden(...)` call. Downstream consumers may see new lint errors at the existing rule severity; the fix is to add the missing `harden(...)` call for the binding name(s) the rule now recognizes. The alias-misnaming case is a bug fix: code that satisfied the prior rule by hardening the source property name was hardening the wrong identifier and will now report; the fix is to harden the binding name instead.

### Upgrade Considerations

No data or wire-format change. Consumers upgrading the eslint plugin run `yarn lint` and address any newly-surfaced reports per the Compatibility note above.

### DIFF
--- a/packages/eslint-plugin/lib/rules/harden-exports.js
+++ b/packages/eslint-plugin/lib/rules/harden-exports.js
@@ -11,6 +11,85 @@
  */
 
 /**
+ * Recursively collect the names introduced by a destructuring or identifier
+ * binding pattern.
+ *
+ * Handles all binding pattern shapes that may appear on the left-hand side of
+ * any `export <kind> ... = ...` declaration where `<kind>` is `const`, `let`,
+ * or `var` (the rule visits every `ExportNamedDeclaration` carrying a
+ * `VariableDeclaration`, not just `const`):
+ *
+ * - Identifier: `export const a = ...`.
+ * - ObjectPattern properties: `export const { a } = ...` and
+ *   `export const { propName: aliasName } = ...`. We recurse into prop.value
+ *   (the binding target) rather than prop.key (the source property name);
+ *   in shorthand they are the same node, but with an alias they differ.
+ * - ObjectPattern rest: `export const { ...rest } = ...`.
+ * - ArrayPattern elements: `export const [ a, b ] = ...`. Skips null elements
+ *   that represent sparse holes (`[ , a ]`).
+ * - AssignmentPattern: defaults like `export const [ a = 1 ] = ...` or
+ *   `export const { a: b = 1 } = ...`. The bound name lives in node.left.
+ * - RestElement: `export const [ a, ...rest ] = ...` and object rest above.
+ *
+ * Returns true if every encountered sub-pattern was a recognized shape, false
+ * if any unknown pattern type was traversed (so the caller may decide whether
+ * to surface a report rather than silently produce a possibly incomplete name
+ * list).
+ * @param {ESTree.Pattern | null} pattern
+ * @param {string[]} names
+ * @returns {boolean}
+ */
+const pushDeclaredNames = (pattern, names) => {
+  if (pattern === null) {
+    // Sparse array hole, e.g., `const [ , a ] = ...`.
+    return true;
+  }
+  switch (pattern.type) {
+    case 'Identifier': {
+      names.push(pattern.name);
+      return true;
+    }
+    case 'ObjectPattern': {
+      let ok = true;
+      for (const prop of pattern.properties) {
+        if (prop.type === 'RestElement') {
+          ok = pushDeclaredNames(prop.argument, names) && ok;
+        } else if (prop.type === 'Property') {
+          // For `{ propName: aliasName }`, prop.value is the binding target
+          // (`aliasName`); for shorthand `{ name }`, prop.value === prop.key.
+          // ESTree narrows ObjectPattern's properties to Property|RestElement,
+          // so prop.value is always a Pattern here.
+          ok = pushDeclaredNames(prop.value, names) && ok;
+        } else {
+          ok = false;
+        }
+      }
+      return ok;
+    }
+    case 'ArrayPattern': {
+      let ok = true;
+      for (const element of pattern.elements) {
+        ok = pushDeclaredNames(element, names) && ok;
+      }
+      return ok;
+    }
+    case 'AssignmentPattern': {
+      // The default value lives in node.right; the binding is in node.left.
+      return pushDeclaredNames(pattern.left, names);
+    }
+    case 'RestElement': {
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin/lib/rules/harden-exports.js
```diff
@@ -92,10 +184,14 @@ module.exports = {
           }
 
           if (missingHardenCalls.length > 0) {
-            const noun = missingHardenCalls.length === 1 ? 'export' : 'exports';
+            const messageId =
+              missingHardenCalls.length === 1
+                ? 'missingHardenCallSingle'
+                : 'missingHardenCallMultiple';
```
turadg: I'm not sure the plurality declension is worth the code logic. Consider simplifying.

turadg:
Cross-referencing the Copilot review with a separate Claude review I ran. Claude's review prompt was:

> I'm reviewing this PR against master. are there any test cases that would improve coverage? I'm particularly interested in cases that might seem like they should have `harden` but don't need it

Two follow-up commits on this branch address everything below that's marked _Fixed_ / _Test added_.

## Issues considered

| # | Topic | Claude | Copilot | Status |
|---|---|---|---|---|
| 1 | Re-exports `export { foo } from 'mod'` push a name with no local binding (false positive) | ✓ | | Fixed |
| 2 | Local rename `export { foo as bar }` pushes `bar` instead of `foo` (false positive / ReferenceError) | ✓ | | Fixed |
| 3 | `export { foo as default }` pushes literal `'default'` (would generate a syntax-error fix) | ✓ | | Fixed (same change as #2) |
| 4 | TS `export type { Foo }` specifier form (false positive) | ✓ | | Fixed |
| 5 | TS inline `export { type Foo, value }` modifier (false positive on `Foo`) | ✓ | | Fixed |
| 6 | `export default …` silently ignored — coverage gap | ✓ | | Test added (valid) |
| 7 | `export * from 'mod'` silently ignored — coverage gap | ✓ | | Test added (valid) |
| 8 | `export class Foo {}` silently ignored — design call | ✓ | | Not changed |
| 9 | `export const a = harden({…})` still requires a second top-level `harden(a)` — design call | ✓ | | Not changed |
| 10 | Near-miss `harden(a, b)` (multi-arg) doesn't satisfy matcher — untested | ✓ | | Test added (invalid) |
| 11 | Near-miss `harden.foo(a)` (member-expression callee) — untested | ✓ | | Test added (invalid) |
| 12 | Near-miss `harden(obj.a)` (non-Identifier arg) — untested | ✓ | | Test added (invalid) |
| 13 | Near-miss `harden(...args)` (SpreadElement arg) — untested | ✓ | | Mentioned, not added |
| 14 | Near-miss conditional `if (cond) harden(a);` (not a top-level statement) — untested | ✓ | | Mentioned, not added |
| 15 | Docstring says `export const`, but rule also fires for `export let` / `export var` | | ✓ | Fixed (rule narrowed to `const`) |
| 16 | When `pushDeclaredNames` returns false, rule reports `unknownBindingPattern` *and* emits a missing-harden report + autofix from a possibly-incomplete name list | | ✓ | Fixed (names buffered per-declaration; dropped on unrecognized shape) |

## Notes

- No overlap between the two reviews. Copilot's two comments were about the rule as it stood; Claude's pass focused on specifier-form false positives and uncovered shapes.
- `#8` (`export class`) and `#9` (`harden(...)`-on-RHS) are behavior calls left for follow-up if desired.
- `#13` and `#14` are easy to add if you want belt-and-suspenders coverage of the harden-call matcher — just say the word.

turadg (approved):

PR #3276, opened 2026-05-21
fix(ses): Cyclic star export with renaming reexport (issue #59)

### LINKED ISSUE #59: Unresolved exports when resolving bindings through "star"
Modules can be visited more than once when resolving bindings through "star" exports as long as the exportName is different each time.

**Steps to reproduce**
In `make-importer`, replace `moddir/index.js` with:
```js
import { x } from './mod1.js';
console.log(x);
```
Replace `moddir/mod1.js` with:
```js
export * from './mod2.js';
```
Replace `moddir/mod2.js` with:
```js
export { y as x } from './mod1.js';
export var y = 45;
```

Then execute:
```sh
$ node -r esm test/test-moddir.js 
```
Actual result:
```
$ node  -r esm test/test-moddir.js 
TAP version 13
# moddir index.js
not ok 1 unexpected exception
  ---
    operator: notEqual
    expected: |-
      [SyntaxError: The requested module 'file://~/projects/make-importer/test/moddir/mod1.js' does not provide an export named 'y']
    actual: |-
      [SyntaxError: The requested module 'file://~/projects/make-importer/test/moddir/mod1.js' does not provide an export named 'y']
    at: Test._387‍.r.test (~/projects/make-importer/test/test-moddir.js:61:7)
    stack: |-
      SyntaxError: The requested module 'file://~/projects/make-importer/test/moddir/mod1.js' does not provide an export named 'y'
          at instance.initialize.then (~/pr
```
Expected result, same as node's native import:
```sh
$ node -r esm test/moddir/index.js
45
```

kriskowal:
Closes: #59

## Description

When a module re-exports `*` from another module, and that other module re-exports a binding from the first under a different exported name (`export { y as x } from './mod1.js'`), the linker visited the first module while its star-imported notifier for `y` had not yet been wired. The synchronous wireUp at the cycle's back-edge then passed `undefined` as the upstream notifier, which manifested as `TypeError: notify is not a function`. The original 2019 issue described the failure as `SyntaxError: ... does not provide an export named 'y'`; the surface symptom evolved as the linker matured, but the underlying defect is the same.

`wireUpExportNotifier` now installs a deferred forwarding notifier for re-exports whose upstream notifier is not yet present. The forwarding queues subscribers until the upstream resolves, then drains them through. By the time any module downstream subscribes to the re-export, the upstream module has completed its candidate-all walk and the upstream notifier exists, so the chain converges.

### Security Considerations

None. The change is internal to the SES module linker: a previously unreachable subscriber path that threw on a missing notifier now installs a forwarding notifier and drains it once the upstream resolves. No new authority is exposed and no trust boundary moves.

### Scaling Considerations

Negligible. The forwarding notifier allocates a small queue per re-export whose upstream is not yet wired, and the queue is drained synchronously the first time the upstream resolves. The fix runs once per qualifying re-export at link time, not per use, and does not affect steady-state import or evaluation cost.

No user-facing documentation change. The behavior the fix restores is the behavior the module spec already prescribes; the previous behavior was a defect in the linker, not a documented capability.

### Testing Considerations

Two regression surfaces accompany the fix. SES's `import-gauntlet` adds `cyclic star export with renaming reexport (issue #59)` exercising the exact reproducer from the issue against the SES linker directly. A compartment-mapper test port drives a three-module fixture (`star-reexporter` re-exports `*` from `export-renamer`; `export-renamer` re-exports `y as x` from `star-reexporter`) through `loadLocation`, `importLocation`, the archive round-trip pair, and `makeArchiveFromMap`. A Node.js parity test imports the same fixture under plain Node.js (no SES, no compartment mapper) and asserts identical expected values from a shared assertions module, pinning the compartment mapper's behavior to Node.js's reference behavior so the expected values are defined in one place.

Reverting the `wireUpExportNotifier` change while keeping either test surface reproduces `TypeError: notify is not a function` (nine of the eleven compartment-mapper import-path variants fail; the two archive-integrity variants pass because they do not import the fixture; the Node.js parity test is unaffected). Re-applying the fix restores all twelve passing tests.

The fix turns a previously-throwing import shape into one that resolves to the binding the module spec prescribes. No code that worked before this change is broken by it. Downstream code that depended on the `TypeError` (or, historically, the `SyntaxError`) as a signal of an unrelated condition is the only conceivable casualty; we are not aware of any such consumer.

### Upgrade Considerations

No upgrade ceremony. A consumer who has been working around the defect by avoiding the cyclic-rename shape can drop the workaround on adoption; no other action is required. A patch-level changeset accompanies the fix.

### DIFF
--- a/.changeset/fix-ses-star-export-cycle-rename.md
+++ b/.changeset/fix-ses-star-export-cycle-rename.md
@@ -1,0 +1,7 @@
+---
+'ses': patch
+---
+
+Fix a star-export cycle defect where a module reached more than once via `export *` and a renaming reexport with a different exported name (`export { y as x } from ...`) raised a spurious `SyntaxError: ... does not provide an export named 'X'` (latterly `TypeError: notify is not a function`).
+The reexport wire-up now installs a deferred forwarding notifier that resolves through the upstream's notifier table on first subscription, so cyclic star-export fixed-points converge.
+Resolves endojs/endo#59.
--- a/packages/compartment-mapper/test/fixtures-cycle-rename/node_modules/app/main.js
+++ b/packages/compartment-mapper/test/fixtures-cycle-rename/node_modules/app/main.js
@@ -1,0 +1,7 @@
+import { x } from './star-reexporter.js';
+import * as ns1 from './star-reexporter.js';
+import * as ns2 from './export-renamer.js';
+
+export const captured = x;
+export const namespace1 = { x: ns1.x, y: ns1.y };
+export const namespace2 = { x: ns2.x, y: ns2.y };
--- a/packages/ses/src/module-instance.js
+++ b/packages/ses/src/module-instance.js
@@ -23,6 +23,7 @@ import {
   assign,
 } from './commons.js';
 import { compartmentEvaluate } from './compartment-evaluate.js';
+import { makeNotifierWithResolver } from './notifier-with-resolver.js';
 
 const { quote: q } = assert;
 
@@ -373,26 +374,18 @@ export const makeModuleInstance = (
       // that name may be wired only later, in its own candidate-all walk,
       // after this module's `imports()` returns. Install a notifier that
       // queues subscribers until the upstream resolves, then forwards them
-      // through.
-      const pendingUpdaters = [];
-      let resolvedUpstreamNotify;
+      // through. `makeNotifierWithResolver` is the synchronous variant of
+      // `Promise.withResolvers` that captures this pattern; each `notify`
+      // call lazily attempts to resolve against the upstream's notifiers.
+      const { notify: queueOrForward, resolve: resolveUpstream } =
+        makeNotifierWithResolver();
       notify = update => {
-        if (resolvedUpstreamNotify !== undefined) {
-          resolvedUpstreamNotify(update);
-          return;
-        }
         const upstreamInstance = mapGet(importedInstances, deferredSpecifier);
         const upstreamNotify = upstreamInstance.notifiers[deferredImportName];
-        if (upstreamNotify === undefined) {
-          arrayPush(pendingUpdaters, update);
-          return;
+        if (upstreamNotify !== undefined) {
+          resolveUpstream(upstreamNotify);
         }
-        resolvedUpstreamNotify = upstreamNotify;
-        for (const pending of pendingUpdaters) {
-          upstreamNotify(pending);
-        }
-        pendingUpdaters.length = 0;
-        upstreamNotify(update);
+        queueOrForward(update);
       };
     }
     notifiers[exportName] = notify;
--- a/packages/ses/src/module-instance.js
+++ b/packages/ses/src/module-instance.js
@@ -354,23 +354,61 @@ export const makeModuleInstance = (
   };
   notifiers['*'] = notifyStar;
 
-  const wireUpExportNotifier = (exportName, notify) => {
-    if (!notifiers[exportName] && notify !== false) {
-      notifiers[exportName] = notify;
-
-      // exported live binding state
-      let value;
-      const update = newValue => (value = newValue);
-      notify(update);
-      exportsProps[exportName] = {
-        get() {
-          return value;
-        },
-        set: undefined,
-        enumerable: true,
-        configurable: false,
+  const wireUpExportNotifier = (
+    exportName,
+    notify,
+    deferredSpecifier,
+    deferredImportName,
+  ) => {
+    if (notifiers[exportName] || notify === false) {
+      return;
+    }
+    if (notify === undefined) {
+      if (deferredSpecifier === undefined) {
+        return;
+      }
+      // The upstream module did not yet expose a notifier for
+      // `deferredImportName` when this re-export was wired. This is the
+      // star-export cycle of endojs/endo#59: the upstream's notifier for
+      // that name may be wired only later, in its own candidate-all walk,
+      // after this module's `imports()` returns. Install a notifier that
+      // queues subscribers until the upstream resolves, then forwards them
+      // through.
+      const pendingUpdaters = [];
+      let resolvedUpstreamNotify;
+      notify = update => {
+        if (resolvedUpstreamNotify !== undefined) {
+          resolvedUpstreamNotify(update);
+          return;
+        }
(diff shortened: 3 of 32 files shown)


packages/ses/src/module-instance.js
```diff
@@ -354,23 +354,61 @@ export const makeModuleInstance = (
...
+      // The upstream module did not yet expose a notifier for
+      // `deferredImportName` when this re-export was wired. This is the
+      // star-export cycle of endojs/endo#59: the upstream's notifier for
+      // that name may be wired only later, in its own candidate-all walk,
+      // after this module's `imports()` returns. Install a notifier that
+      // queues subscribers until the upstream resolves, then forwards them
+      // through.
+      const pendingUpdaters = [];
+      let resolvedUpstreamNotify;
+      notify = update => {
```
naugtur, line 382: Is a situation possible where all calls to the deferring notify happen before `upstreamNotify` can be obtained? I was trying to come up with when that could happen and I thought of unused live bindings.

packages/compartment-mapper/test/fixtures-cycle-rename/node_modules/app/main.js
```diff
@@ -0,0 +1,7 @@
+import { x } from './star-reexporter.js';
+import * as ns1 from './star-reexporter.js';
+import * as ns2 from './export-renamer.js';
```
naugtur, line 3: I see the change is only in `makeModuleInstance`. Would it still work if the cycle was longer and a cjs module was involved in it?

Would you be so kind to ask a clanker for a fixture for that?

packages/ses/src/module-instance.js
```diff
@@ -354,23 +354,61 @@ export const makeModuleInstance = (
...
+  const wireUpExportNotifier = (
+    exportName,
+    notify,
+    deferredSpecifier,
+    deferredImportName,
+  ) => {
+    if (notifiers[exportName] || notify === false) {
+      return;
+    }
+    if (notify === undefined) {
```
naugtur, line 367: I know this is vague, but it'd feel nicer design-wise to create all notifiers ahead of time and give them an ability to forward to or pull from another notifier, in which case the reexport * would be a matter of connecting notifiers in a loop. 

If we could use this as an opportunity to design a fancier notifier primitive to share between the two implementations `makeModuleInstance` and `makeVirtualModuleInstance` we'd avoid some of the risk of interop failure

packages/ses/src/module-instance.js
kriskowal: I (loosely) have added tests that cover the unused live binding case. This was possible with an uninitialized `export var`.

packages/compartment-mapper/test/fixtures-cycle-rename/node_modules/app/main.js
kriskowal: We (dually) have added fixtures that cover {esm,cjs}x{esm,cjs}, with {node,endo} parity validation.

packages/ses/src/module-instance.js
kriskowal: I managed to convince my inanimate engineering partner to extract a notifier primitive, which is spiritually similar to Promise.withResolvers but with synchronous propagation. I didn’t find a way to integrate it in virtual module instances, which don’t support live bindings by design.

.changeset/fix-ses-star-export-cycle-rename.md
boneskull, line 1: The PR description mentions "no user-facing changes" yet the changeset here implies a user-facing change (by its existence).  Just wanted to point out the possible incongruity.

packages/ses/src/notifier-with-resolver.js
```diff
@@ -0,0 +1,56 @@
+import { arrayPush } from './commons.js';
```
boneskull, line 1: (nit) add a `@module` docstring

packages/ses/src/notifier-with-resolver.js
```diff
@@ -0,0 +1,56 @@
...
+ * once `resolve` is called, queued updaters are replayed to the target
+ * notifier and subsequent `notify(update)` calls forward directly through.
+ *
+ * `resolve` is one-shot: the first call settles the resolver to its target,
+ * and subsequent calls are no-ops. (This matches `Promise.withResolvers`
+ * semantics: a promise settles once.) Callers that need to discover the
+ * target lazily may safely call `resolve` again on each `notify`; only the
+ * first call has effect.
+ *
+ * Used by `module-instance.js` `wireUpExportNotifier` to resolve the
```
boneskull, line 19: (nit) this sort of comment is brittle. maybe makes more sense to reference `{@link makeNotifierWithResolver}` within the docstring for `wireUpExportNotifier` instead.

packages/ses/src/notifier-with-resolver.js
```diff
@@ -0,0 +1,56 @@
...
+  const pendingUpdaters = [];
+  /** @type {((update: (value: any) => void) => void) | undefined} */
+  let resolvedTargetNotify;
+
+  const notify = update => {
+    if (resolvedTargetNotify !== undefined) {
+      resolvedTargetNotify(update);
+      return;
+    }
+    arrayPush(pendingUpdaters, update);
```
boneskull, line 41: (style opinion) I'd probably change this (and `resolve` below) so the conditional uses strict equality (`resolvedTargetNotify === undefined`) and/or use an `else` instead of an early return.

packages/ses/test/import-cjs.test.js
```diff
@@ -667,6 +667,87 @@ test('importNow handles a cycle in CommonJS modules', t => {
...
+  t.plan(3);
+
+  const resolveHook = resolveNode;
+  const importHook = async specifier => {
+    if (specifier === './star-reexporter.cjs') {
+      // CommonJS analog of `export * from './export-renamer.mjs'`: capture
+      // the renamer's properties by assignment. Because both are required
+      // from each other in a cycle, the values read here are whatever the
+      // renamer had set on its `exports` by the re-entry instant.
+      return CjsModuleSource(
```
boneskull, line 700: note to self: #3220 creates a "real" `CjsModuleSource` in `@endo/module-source` which this test should use instead of the mock.

boneskull (approved):

PR #3284, opened 2026-05-26
feat(ses): add four new permitted intrinsics shipping in V8 v14.6

boneskull:
- `Map.prototype.getOrInsert`
- `Map.prototype.getOrInsertComputed`
- `WeakMap.prototype.getOrInsert`
- `WeakMap.prototype.getOrInsertComputed`

`Date.prototype.toTemporalInstant` is marked as `false` to omit it from the warning that `lockdown` would otherwise emit.

Ref: https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V26.md#2026-05-05-version-2600-current-rafaelgss

PR #3285, opened 2026-05-27
feat(ses): permit Temporal

erights:
Closes: #XXXX
Refs: #3284 

## Description

Permits `Temporal` in a way analogous to how we permit `Date` or `Math`
- The original powerful one is still available on the start compartment (i.e. the realm global)
- All constructed compartments are endowed by default with a shared safe one.

Shared safe version by omission
- For `Math`, the shared safe one omits `Math.random`.
- For `Date`, the shared safe one omits `Date.now` and the no-arg behavior of the `Date` constructor.
- For `Temporal`, the shared safe one omits `Temporal.Now`.

### Security Considerations

The original intention is that the shared safe one omit all dynamic non-determinism, since dynamic non-determinism enable non-overt (side and covert) channels. In general, we are not trying to suppress static non-determinism, but we want to be aware of all sources of static non-determinism since those enable fingerprinting.

The implementation in this PR follows the original implementation plan which was motivated by the original intention. However, to my surprise, it does not quite achieve the original intention.
- Locale is not specified to be stable during the lifetime of an agent. Instead, if the system locale changes, the locale-sensitive parts of JS can reflect that change, and does on some engines. However
  - With the default `localeTaming: 'safe'` setting of the lockdown option, it was already the case that `*Locale*` methods, like `toLocaleString`, would be aliased to the non-locale-sensitive sibling method, like `toString`. Without needing any explicit change, this PR should cause that default aliasing to apply to `Temporal` as well. Likewise, the `localeTaming: 'unsafe'` setting should suppress the aliasing both for the existing methods and for this PR's permitting of `Temporal` as well.
  - Date.prototype.toString()` may currently reveal locale, timezone, and UTC offset (daylight saving time) as well. This should probably be fixed, at least under the default `localeTaming: 'safe'` setting.
- Timezone and UTC offset is not specified to be stable during the lifetime of an agent. Instead, if the system timezone and/or UTC offset changes, the timezone and UTC offset parts of JS can reflect that change, and does on some agents. However,
  - `Temporal` only reveals timezone and UTC offset from `Temporal.Now`. With `Temporal.Now` suppressed in the shared safe `Temporal`, changes to timezone or UTC offset should be unobservable as well. So even though, to my surprise, this is a source of (very slow) dynamic non-determinism, it is already suppressed along with normal dynamic non-determinism (the ever-changing current time).
  - See above caveats about `Date.toString()`.
- Which CLDR-like table is in use is still only static non-determinism. This can help fingerprinting, but cannot be suppressed at reasonable cost, so we elect not to. Instead, we should just remain aware of its contribution to fingerprinting. This is analogous to our stance on the error messages of engine-thrown errors.

It is worth noting that even though locale, timezone, and UTC offset are technically sources of dynamic non-determinism, they change so slowly that they are not a practical enabler of non-overt (side and covert) channels.

### Scaling Considerations

none

by permitting `Temporal` we do not need to explain its absence. However, the above Security Considerations should be re-explained somewhere findable.

### Testing Considerations

- [x] needs tests

Makes HardenedJS much more compatible with JS.

### Upgrade Considerations

none.


erights:
The attached `permit-temporal.js` file was a helper I wrote to help me make the final changes to `permits.js`. But it still required ad hoc adjustments, so it isn't the whole story. It is also terrible code, which is fine for run once code, but not production code. So it should not be checked in. Instead, just recording it here for now.

[permit-temporal.js](https://github.com/user-attachments/files/28295129/permit-temporal.js)

boneskull (commented):
Not familiar enough with `Temporal` to feel comfortable approving, but don't see anything obviously untoward here.

packages/evasive-transform/src/index.js
```diff
@@ -23,7 +23,7 @@ import { generate } from './generate.js';
  * @property {import('./parse-ast.js').SourceType | undefined} [sourceType] - Module source type
  * @property {boolean | undefined} [onlyComments] - if true, will limit transformation to
 comment contents, preserving code positions within each line
-* @property {(path: import('@babel/traverse').NodePath) => void} [customVisitor] - A visitor function to be called on each node, in addition to the standard transforms. Receives the same path argument as a normal Babel visitor.
+ * @property {(path: import('@babel/traverse').NodePath) => void} [customVisitor] - A visitor function to be called on each node, in addition to the standard transforms. Receives the same path argument as a normal Babel visitor.
```
erights, line 26: driveby: fix lint warning

erights:
Ready for review

@boneskull , a lot has changed since you last looked. PTAL

kriskowal (approved):

PR #3286, opened 2026-05-27
chore: fix clean script

boneskull:
This fixes the clean script so that instead of attempting to use the now-generally-unused `postpack` script, it instead runs what _was_ typically the `postpack` script in each workspace.


boneskull:

- branch: `master`
  - **[chore: fix clean script #3286](https://github.com/endojs/endo/pull/3286)** - branch: `boneskull/fix-clean-script` *(this PR)*

---

package.json
```diff
@@ -44,7 +44,7 @@
     "zx": "^8.1.8"
   },
   "scripts": {
-    "clean": "yarn workspaces foreach --all run postpack",
+    "clean": "yarn workspaces foreach --all exec git clean -fX -e node_modules/",
```
turadg: no need for `yarn workspaces` since  this is a git command.

```suggestion
    "clean": "clean -fX -e node_modules",
```

assuming it will exclude nested node_modules too, which I'm not certain of. please confirm either way.

turadg (approved):

package.json
boneskull: Yeah, IDK either. I'll confirm

boneskull: This works. I'll do that instead.

PR #3292, opened 2026-05-29
chore: dependency maintenance (typescript-eslint, ava 8, fast-check 4, prettier)

turadg:
Refs: #2327

## Description

A grab-bag of dependency maintenance and the lint/type cleanups required to make those bumps build. No runtime/API behavior changes; the substantive commits are dependency hygiene plus the source adjustments their new versions demanded.

**Dependency bumps & hygiene**
- `chore(deps): bump prettier to 3.8.3` + `style: yarn format` — reformat under the new prettier.
- `chore(deps): bunx taze --write` — routine in-range bumps.
- `chore(deps): bump testing deps (ava, fast-check, tsd)`, then `chore(deps): yarn dedupe` — collapse duplicate resolutions.
- `chore(deps): remove unused @octokit/core` — dead dep added in 2021 (`b93478221`) only to silence a peer-warning from since-removed tooling; nothing imports it. Removing it beats bumping it 3.x→7.x (which would force ESM-only Octokit + Node 20 engines) for code that doesn't exist.
- `chore(deps): bump typescript-eslint` + `fix(eslint-plugin): use version ranges for typescript-eslint deps` — the family had been pinned to exact `8.39.1` in `58ef1d9ec` (PR #2327, the pnpm-linker migration) to stabilize typechecks under the strict linker. Exact pins in the **published** `@endo/eslint-plugin` are contagious to consumers, so this restores caret ranges there while the (private) root carries `^8.60.0`; the lockfile dedupes everything to 8.60.0.

**Source changes the bumps required**
- `fix(pass-style): migrate arb-passable to fast-check 4 API` — fast-check 4 removed `fullUnicodeString()`; replaced with `string({ unit: 'binary' })` and dropped a now-unneeded `@ts-expect-error`.
- `fix(ses-ava): align ava to a single deduped version` — the catalog moved to ava 8 while ses-ava/root still resolved ava 6, producing two incompatible virtual `Assertions` types under `tsc --build`. ses-ava now declares a permissive `^6 || ^7 || ^8` direct dep (not `catalog:dev`, which would force consumers onto ava 8; not a peer, which breaks the bin under the pnpm linker).

**CI / docs**
- `ci: remove obsolete test-async-hooks`
- `docs: promise-hook Node version compatibility`

### Security Considerations

None. Net dependency surface *shrinks* (removed `@octokit/core` and its transitive subtree). No new authorities or trust boundaries.

### Scaling Considerations

None.

`fix(ses-ava)` widens the published ava range to `^6 || ^7 || ^8`, so consumers may stay on ava 6/7 rather than being forced to 8 — a compatibility *improvement*, no migration required.

### Testing Considerations

No new functionality; existing unit tests cover the changed source. The fast-check 4 and ava 8 migrations are validated by the existing pass-style / ses-ava suites under CI. Worth confirming `@endo/eslint-plugin`'s pack typechecks remain stable now that it resolves typescript-eslint 8.60.0 (the stabilization the original exact-pin guarded).

No breaking changes to published APIs. Published dependency ranges are *widened* (ava in ses-ava; typescript-eslint in eslint-plugin), which only loosens consumer constraints. No changeset needed (no `major` bumps).


yarn.lock
```diff
@@ -2149,7 +2082,14 @@ __metadata:
...
 
-"@types/estree@npm:*, @types/estree@npm:1.0.8, @types/estree@npm:^1.0.0":
+"@types/estree@npm:*, @types/estree@npm:^1.0.0, @types/estree@npm:^1.0.6":
+  version: 1.0.9
+  resolution: "@types/estree@npm:1.0.9"
+  checksum: 10c0/3ad3286ca2988cd550dafb8f2ad599c8474868e954fa601a36655bdfefd8039f7c714b8c1c7f2ae219ffbd58bd4660e66fa7479a0120fc02d4777057d4865387
+  languageName: node
+  linkType: hard
+
+"@types/estree@npm:1.0.8":
```
mhofman, line 2092: I suppose case of a sequence of yarn install and depupe not coalescing on the minimal set of deps

.changeset/pass-style-fast-check-4.md
```diff
@@ -0,0 +1,8 @@
+---
+'@endo/pass-style': patch
+---
+
+Update the `tools/arb-passable` arbitrary for the fast-check 4 API:
+`fullUnicodeString()` was removed, so it now uses the documented
+equivalent `string({ unit: 'binary' })`. Consumers using this arbitrary
+should depend on fast-check 4.
```
mhofman, line 8: Ok this is actually a little confusing. We're bumping to `@fast-check/ava` to `3.0.1`, but that depends on `fast-check: `3.0.0^ || 4.0.0^`, so we ended up on the 4 side (even though we previously were on a resolved `fast-check: 3.1.1`).

I'm surprised yarn wouldn't have tried to keep the version we were on.

.changeset/ses-ava-ava-range.md
```diff
@@ -0,0 +1,8 @@
+---
+'@endo/ses-ava': patch
+---
+
+Widen the declared `ava` dependency range to `^6 || ^7 || ^8` so
+consumers can stay on ava 6 or 7 rather than being forced onto ava 8.
```
mhofman, line 6: Given our own experience here with the fast-check package, I suspect our consumers will just end up getting bumped unless they take a manual resolution action.

mhofman (commented):
Seems like some worthwhile cleanups

packages/pass-style/tools/arb-passable.js
```diff
@@ -18,7 +18,7 @@ export const exampleCarol = Far('carol', {});
  * @param {Array<'byteArray'>} [exclusions]
  */
 export const makeArbitraries = (fc, exclusions = []) => {
-  const arbString = fc.oneof(fc.string(), fc.fullUnicodeString());
+  const arbString = fc.oneof(fc.string(), fc.string({ unit: 'binary' }));
```
mhofman, line 21: I would want @gibson042 to double check the semantics here.

turadg:
I've removed this from the repo config:
Screenshot 2026-05-29 at 2 54 23 PM

packages/pass-style/tools/arb-passable.js
gibson042: Confirmed; I've actually already run into this in Endo.

gibson042 (approved):

PR #3300, opened 2026-06-08
chore(shellcheck): add yarn shellcheck script and CI workflow

kriskowal:
## Description

Adds a `yarn shellcheck` gate that lint-checks every tracked shell script in the repository, and brings the existing scripts into compliance with it.

The wrapper (`scripts/shellcheck.sh`) enumerates `git ls-files '*.sh'` and runs `shellcheck -S warning` over the result, forwarding any extra arguments to the underlying checker. The file list is streamed to shellcheck through a git blob (`git hash-object -w --stdin`, then `git cat-file blob`) rather than an argument vector or shell variable, so the gate stays clear of `ARG_MAX` on hosts with many tracked scripts. It exits 0 when the list is empty, so a fork or branch with no shell scripts stays green.

A companion CI workflow runs the gate on push to `master` and on any pull request whose diff touches a `.sh` file; the workflow's path filter means a pull request that changes no shell scripts does not trigger it at all. shellcheck ships preinstalled on the `ubuntu-latest` runner, so the workflow needs no install step.

The remainder of the change fixes the warnings shellcheck reports at `-S warning` across the scripts that already exist in the tree, so the gate is green from the first run. The fixes are mechanical and do not change runtime behavior: missing shebangs, `|| exit` on `cd` invocations whose failure should abort, whitespace-safe iteration over `find` output, `read -r`, and unambiguous `CDPATH=''`.

### Security Considerations

The workflow runs with `contents: read` and `persist-credentials: false`, and pins its actions to commit SHAs in keeping with the repository's existing posture. The gate executes shellcheck over the repository's own tracked scripts and grants no new authority to those scripts.

### Scaling Considerations

The workflow only runs when a pull request's diff includes a shell script, so it adds no CI time to the common case. When it does run, it is a single shellcheck pass over a small file set.

None. The `yarn shellcheck` script is self-describing and the workflow is internal to CI.

### Testing Considerations

The gate is its own test: it passes against the current tree because this change also fixes the pre-existing warnings. New shell scripts (or edits to existing ones) that introduce a warning at `-S warning` will fail the gate in CI.

No usage patterns change. The script edits preserve existing behavior; the new gate is additive.

### Upgrade Considerations

None. This affects developer tooling and CI only, with no runtime or on-chain impact.


.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,44 @@
+name: Shell script lint
+
+# Run shellcheck against every .sh file tracked in the repo. The
+# workflow is gated on path-filter `**/*.sh`, so a PR that touches no
+# shell scripts does not trigger any runs at all (the "skip when no
+# .sh in the diff from the base" behavior).
```
turadg: It's a very cheap check. Why not include it in the lint job?

package.json
```diff
@@ -55,6 +55,7 @@
     "lint:fix": "yarn lint:eslint -- --fix",
     "lint:workspaces:fix": "yarn workspaces foreach --all run lint-fix",
     "lint:prettier": "prettier --check .github packages",
+    "shellcheck": "scripts/shellcheck.sh",
```
turadg: I would expect this to run as part of `yarn lint`.

In general there is little point in having a package.json entry that just redirects to a `scripts` file of the same name.

kriskowal: I find it helps with discoverability and also sets CWD to the project root, if you’re not already there.

.github/workflows/shellcheck.yml
kriskowal: I think, though I prefer to keep the `node --run shellcheck` separate, rolling shellcheck into the lint CI job would be more economical use of action runners. I’ll drive that direction.

gibson042:
> The file list is streamed to shellcheck through a git blob (`git hash-object -w --stdin`, then `git cat-file blob`) rather than an argument vector or shell variable, so the gate stays clear of `ARG_MAX` on hosts with many tracked scripts.

I think this is overly cute. If the `ubuntu-latest` runner includes shellcheck, then _surely_ it also includes `xargs`!

scripts/shellcheck.sh
```diff
@@ -0,0 +1,34 @@
...
+# The list is enumerated via `git ls-files '*.sh'` so untracked scratch
+# scripts and worktree noise are ignored.
+
+set -eu
+
+# Stash the file list as a git blob and stream it back out, rather than
+# holding it in a shell variable. This keeps us clear of ARG_MAX limits
+# on hosts with a large number of tracked .sh files, and lets the body
+# branch on emptiness without an oversize positional-argument list.
+HASH=$(git ls-files -z '*.sh' | tr '\0' '\n' | git hash-object -w --stdin)
```
gibson042: I take it back; this is not unreasonable.

scripts/shellcheck.sh
```diff
@@ -0,0 +1,34 @@
...
+
+if [ -z "$(git cat-file blob "$HASH")" ]; then
```
gibson042: We don't need to load the whole list into a string here, though.
```suggestion
if [ "$(git cat-file -s "$HASH")" -eq 0 ]; then
```

scripts/shellcheck.sh
```diff
@@ -0,0 +1,34 @@
...
+  exit 0
+fi
+
+# shellcheck disable=SC2086
+# Word splitting on the blob's contents is intentional: each line is a
+# separate path and tracked .sh filenames in this repo do not contain
+# whitespace. `git ls-files` would happily emit NULs for whitespace-bearing
+# names; if such a name lands here, switch to xargs -0 with a portability
+# shim.
+git cat-file blob "$HASH" | xargs shellcheck -S warning "$@"
```
gibson042:
```suggestion
git cat-file blob "$HASH" | tr '\n' '\0' | xargs -0 shellcheck -S warning "$@"
```

gibson042 (approved):
I agree with @turadg; this should run as part of repo-level linting.

package.json
turadg: Why would one need to discover "shellcheck" independent of "lint"? It's very much a linking concern.

At the least please name this "lint:shellcheck"

package.json
turadg: Please include lint:shellcheck here. Then you don't need to change the CI workflow. You could still put the command presence check there in a setup step

turadg (changes requested):

kriskowal:
@turadg @gibson042 Addressed your feedback. The new change is much more surgical.

PR #3287, opened 2026-05-27
fix(patterns): bump version for publishing

boneskull:
This bumps the version to v2.0.0 to align it with what is actually in the registry; v2.0.0 is already published, but our `package.json` lags behind.  Version v1.9.1 could not be published.

Once this is merged, the next release will bump the version to v2.0.1.

PR #3301, opened 2026-06-12
fix(release): Idempotent publish in face of prior partial success

kriskowal:
This change just alters `yarn release:npm` such that it will skip packages that were previously published. It also includes some errant version changes I made in the process of discovering that this was a problem.


boneskull (approved):
What of #3287?

boneskull:
cc @turadg ; this seems to address some issues I complained about earlier

PR #3306, opened 2026-06-17
fix(ses): Prevent Node.js console logging from invoking custom inspect methods

gibson042:
Closes: TBD

## Description

Reported by @ChALkeR 

Updates `consoleTaming: 'safe'` to suppress invocation of custom inspect methods by delegating to an internal console instance constructed with appropriate `inspectOptions` configuration rather than to the global console (which defaults to `customInspect: true`).
Further reading:
* https://nodejs.org/docs/latest/api/util.html#utilinspectobject-options
* https://nodejs.org/docs/latest/api/util.html#custom-inspection-functions-on-objects
* https://nodejs.org/docs/latest/api/util.html#utilinspectcustom

### Security Considerations

We could also go further and suppress the callbacks associated with [console format specifiers](https://console.spec.whatwg.org/#formatting-specifiers) by prepending an empty string to (string, unknown, ...unknown[]) arguments (such that `console.log("%d", { valueOf: () => 42 })` is effectively transformed into `console.log("", "%d", { valueOf: () => 42 })` and logs ` %d { valueOf: [Function: valueOf] }` rather than `42`).

### Scaling Considerations

n/a

Updates [lockdown.md](https://github.com/endojs/endo/blob/master/docs/lockdown.md) and [reference.md](https://github.com/endojs/endo/blob/master/docs/reference.md).

### Testing Considerations

Covered in packages/ses/test/error/tame-console*test.js.

We never intended to support the Node.js functionality.

### Upgrade Considerations

This is a patch fix that introduces no compatibility concerns of its own.

### DIFF
--- a/.changeset/lemon-states-do.md
+++ b/.changeset/lemon-states-do.md
@@ -1,0 +1,6 @@
+---
+'ses': patch
+---
+
+- `consoleTaming: 'safe'` prevents Node.js console logging from invoking custom
+  inspect methods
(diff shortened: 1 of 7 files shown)


erights:
@gibson042 , I revised the PR comment to add "Reported by @ChALkeR" at the top, paralleling #3307 . I believe he reported both.

erights (approved):
LGTM, thanks!

PR #3307, opened 2026-06-17
fix(ses): Sanitize console format strings

gibson042:
Closes: #TBD

## Description

Reported by @ChALkeR 

The console format specifier `%c` is for consuming a CSS style string and applying that style to the rendering of the remaining arguments. Node.js and browsers both parse `%c` the same way and have it consume one argument.
- Browsers actually make use of the CSS, leading to security problems (e.g., https://issues.chromium.org/40056332 ).
- Node.js ignores the `%c` and its corresponding argument, which is allowed by the [WHATWG Console specification](https://console.spec.whatwg.org/).

To avoid the CSS security problems on all platforms, under the defaut `consoleTaming: 'safe'`, we now sanitize out the `%c` and corresponding argument, emulating the allowed current Node.js behavior on all platforms. This fixes this CSS vulnerability while maintaining compatibility with the specification.

We also treat unknown specifiers in a future-proof manner. For example, `%j` is unspecified by the WHATWG spec, but it is implemented by Node.js to render its corresponding argument as JSON. Being unspecified, `%j` would just print literally without consuming an argument. But on Node.js, because it consumes an argument, all the remaining arguments are shifted from their standard treatment.

### Security Considerations

The point. @ChALkeR found that the browser's processing of the `%c` CSS string can cause it to fetch resources from the network, breaking confinement and the write-only nature of `console`. We can only widely and safely share `console` if it is write-only to everyone that cannot read the console log output.

### Scaling Considerations

None

Not really. After this change, the `ses` console still conforms to the WHATWG spec, including the allowed behavior of ignoring `%c` and its argument as Node.js does.

### Testing Considerations

Two tests. The `sanitize-by-method.test.js` test also caught some previously undiscovered bugs in the ses console implementation.

With this PR, the ses console conforms to the WHATWG spec, and so is compatible with any code relying only on the spec. The two minor compat concerns with previous behavior on the ses console on various platforms:
- Even on browsers, `%c` will not be ignored.
- Even on Node.js, `%j` will be treated as not a specified specifier, shifting the treatment of remaining arguments from what Node does.

### Upgrade Considerations

None.

### DIFF
--- a/.changeset/rich-zoos-make.md
+++ b/.changeset/rich-zoos-make.md
@@ -1,0 +1,9 @@
+---
+'ses': patch
+---
+
+The console format specifier `%c` is for consuming a CSS style string and applying that style to the rendering of the remaining arguments. Node.js and browsers both parse `%c` the same way and have it consume one argument.
+- Browsers actually make use of the CSS, leading to security problems (e.g., https://issues.chromium.org/40056332 ).
+- Node.js ignores the `%c` and its corresponding argument, which is allowed by the [WHATWG Console specification](https://console.spec.whatwg.org/).
+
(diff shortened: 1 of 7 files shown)


erights (approved):
LGTM, thanks!

